PR #2127, opened 2016-07-17
replace py27-flake8 with py34-flake8

jrobeson:
We definitely want py34-flake8 to succeed, so we don't end up with
`unicode` and other py27 specific stuff. With that in mind, we don't need
py27-flake8 anymore either.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,14 +17,13 @@ matrix:
           env: {TOX_ENV: py35-test}
         # - python: pypy
         #   - env: {TOX_ENV: pypy-test}
-        - python: 2.7
-          env: {TOX_ENV: py27-flake8}
         - python: 3.4
           env: {TOX_ENV: py34-flake8}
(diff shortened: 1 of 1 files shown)


jackwilsdon:
Aren't we trying to support both 2.7 and 3.x currently? Is there some code that fails for flake8 under 2.7 but passes under 3.x or visa versa?

jrobeson:
there's nothing that I've seen that will pass in 3.x that fails in 2.7 that wouldn't also fail the test suite.

jackwilsdon:
Is there a reason we don't need `py27-flake8` anymore then? I don't see any reason we shouldn't keep it until we stop supporting 2.7 completely.

jrobeson:
I was just trying to be nice to travis by keeping our build matrix small, since it didn't seem to provide any value.  I certainly didn't remove it from our tox config, for folks running tests locally.

PR #2128, opened 2016-07-18
Switch from pathlib to pathlib2

jrobeson:
Quoting the pathlib2 documentation:

> The old pathlib module on bitbucket is in bugfix-only mode. The goal of pathlib2 is to provide a backport of standard pathlib module which tracks the standard library module, so all the newest features of the standard pathlib can be used also on older Python versions.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -41,7 +41,12 @@ And there are a few bug fixes too:
 The last release, 1.3.19, also erroneously reported its version as "1.3.18"
 when you typed ``beet version``. This has been corrected.
 
+????
+
+* The pathlib module dependency was replaced by `pathlib2`_
+
 .. _six: https://pythonhosted.org/six/
+.. _pathlib2: https://pypi.python.org/pypi/pathlib2/
 
 
 1.3.19 (June 25, 2016)
--- a/beetsplug/thumbnails.py
+++ b/beetsplug/thumbnails.py
@@ -25,7 +25,7 @@ from hashlib import md5
 import os
 import shutil
 from itertools import chain
-from pathlib import PurePosixPath
+from pathlib2 import PurePosixPath
 import ctypes
 import ctypes.util
 
--- a/docs/plugins/thumbnails.rst
+++ b/docs/plugins/thumbnails.rst
@@ -7,7 +7,7 @@ Nautilus or Thunar, and is therefore POSIX-only.
 
 To use the ``thumbnails`` plugin, enable it (see :doc:`/plugins/index`) as well
 as the :doc:`/plugins/fetchart`.  You'll need 2 additional python packages:
-`pyxdg` and `pathlib`.
+`pyxdg` and `pathlib2`.
 
 ``thumbnails`` needs to resize the covers, and therefore requires either
 `ImageMagick`_ or `Pillow`_.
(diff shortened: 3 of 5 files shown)


docs/changelog.rst
```diff
@@ -41,7 +41,12 @@ And there are a few bug fixes too:
 The last release, 1.3.19, also erroneously reported its version as "1.3.18"
 when you typed ``beet version``. This has been corrected.
 
+????
+
```
jrobeson: I wasn't sure what heading to put this under :)

sampsyo: Under "Fixes" is probably fine for this sort of thing. And we should probably note that the dependency is for the `thumbnails` plugin, not a global dependency for beets.

sampsyo:
This seems good! Is there any sense in doing something like:

```py
try:
    import pathlib2 as pathlib
except ImportError:
    import pathlib
```

for Python 2/3 compatibility? That would have the side benefit of not disrupting existing installations.

jrobeson:
Exclusively using the same module would (hopefully) give us the same behavior on 2.7 and 3.4+, rather than trying to track down issues in one place for 3.4 and one place  for 2.7.

sampsyo:
I'm not too worried about differences between the stdlib module and the one from PyPI—those seem like they'd be pretty rare, given that the two are supposed to track one another closely.

About disrupting installations: I meant that we could avoid bothering people who use the `thumbnails` plugging and have a working setup today (i.e., they ran `pip install pathlib`) and making them switch to a new dependency (with `pip uninstall pathlib && pip install pathlib2`).

jrobeson:
but then we'll have to make sure that we're helping debug against the correct version of pathlib, and it'd be easy to forget to ask which one they have.  I certainly will do it that way if you insist, but I'd need more convincing to agree with it.

sampsyo:
Well, if we just say that, on Python 2, it's always pathlib2, and on Python 3, it's always the standard library, it seems pretty easy to distinguish. It would be a shame to depend on a duplicate of something Python ships with standard!

jrobeson:
I had forgotten for a moment that we're targeting 3.4, not 3.3. If we were targeting 3.3 that wouldn't be so simple

jrobeson:
Oh, but that doesn't take into account your original issue about forcing people to change. What you said, suggests that we're still going to do that.

sampsyo:
Yes, we could do the one thing (use the stdlib module) without the other (avoiding upgrade disruption).

Your point about bytestring paths in pathlib elsewhere is worth considering, too—should we use pathlib2 on Python versions that don't support that?

jrobeson:
that would be all python versions in the wild.

sampsyo:
Well, maybe we should cross this bridge when we *do* want those features then. While the library is only used for this specific purpose, and we don't need the new features, does switching gain us anything?

Of course, when we switch the whole codebase over, I totally agree we'll want a version that can use bytestring paths. (Unless we also want to switch to surrogate-escaped Unicode strings at the same time.)

jrobeson:
i suppose we can hold off for a bit. I was just looking for small things to do while waiting for your glorious return!

sampsyo:
Cool; thanks for making me take a look! Things are starting to settle down, and I should be back in the game soon.

PR #2137, opened 2016-07-22
Fix patcher to always be correctly cleaned up

### LINKED ISSUE #2135: test_nonexistent_file (test_library.FilesizeTest) fails
### Problem
test_nonexistent_file (test_library.FilesizeTest) fails when run via `python setup.py test`, but not when run via `tox -e py27-test`.

```
    self.assertEqual(item.filesize, 0)
AssertionError: 987 != 0
```

### Setup

* OS: Debian Jessie
* Python version: 2.7.9
* beets version: master (e7428972ca88e688dcef579f4a3d363a9ba8bca1)

JesseWeinstein:
fixes #2135 .

### DIFF
--- a/test/test_ui.py
+++ b/test/test_ui.py
@@ -24,7 +24,7 @@ import subprocess
 import platform
 from copy import deepcopy
 
-from mock import patch
+from mock import patch, Mock
 from test import _common
 from test._common import unittest
 from test.helper import capture_stdout, has_program, TestHelper, control_stdin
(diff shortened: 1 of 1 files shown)


JesseWeinstein:
While this is quite trivial, it's also not urgent, and I'm still new to beets development, so I'll let someone else press the Merge button.

sampsyo:
Wow; thank you! This looks incredibly hard to pin down. Makes me wish (once again) for a more failsafe way to ensure test harnesses are cleaned up…

PR #2141, opened 2016-07-24
(Mostly) remove patch dot start

JesseWeinstein:
This removes most of the uses of `mock.patch.start()` from the tests. It is good to remove these, because unlike the other two forms of `mock.patch` (annotation and context manager) the `start()` method requires explicit cleanup.

It is probably possible to remove it from the two remaining files (`test_plugins.py` and `test_query.py`) but it's trickier, and I thought it worth applying these changes in the meantime.

Even though this only touches test code, I'd appreciate someone else hitting the Merge button.

### DIFF
--- a/test/test_edit.py
+++ b/test/test_edit.py
@@ -138,7 +138,7 @@ class EditCommandTest(unittest.TestCase, TestHelper, EditMixin):
         self.assertTrue(all(i.title.startswith(title_starts_with)
                             for i in self.lib.items()))
 
-    def test_title_edit_discard(self, m):
+    def test_title_edit_discard(self, mock_write):
         """Edit title for all items in the library, then discard changes."""
         # Edit track titles.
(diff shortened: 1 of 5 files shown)


sampsyo:
Hey; thank you for doing this! All these changes look good from here. Maybe we should add a note about sticking with this style to the wiki page about the tests…

Looks Travis sees one more error in a `test_play` test: https://travis-ci.org/beetbox/beets/jobs/147038572

After which I'll hit the big green button.

JesseWeinstein:
Yep, I'm working on figuring that out. I've added a warning to the wiki page already.

sampsyo:
Great! Thanks for the wiki addition. This looks like it's passing now—is it ready from your perspective?

JesseWeinstein:
Yep, I'm happy with it.

sampsyo:
Great; thank you! I gave this another quick review and everything looks good.

PR #2143, opened 2016-07-26
Add BPM data with acousticbrainz

Jell-E:
The bpm plugin is a very slow way to add bpm data to songs, using the acousticbrainz plugin is much faster and can doesn't need a new plugin.


sampsyo:
Good idea! Could you also please add to the documentation and the changelog?

sampsyo:
Awesome! Thank you again. This is merged. :sparkles:

PR #2144, opened 2016-07-26
fix bug where $args would be included in the command output

tweitzel:
if $args was set in config but -A or --args was left off the beet play command, $args would appear in the command. Fixed and test included thanks to Wise Tutelage of jrobeson

### DIFF
--- a/beetsplug/play.py
+++ b/beetsplug/play.py
@@ -91,8 +91,7 @@ class PlayPlugin(BeetsPlugin):
             else:
                 command_str = u"{} {}".format(command_str, opts.args)
         else:
-            if ARGS_MARKER in command_str:
-                command_str = command_str.replace(" " + ARGS_MARKER, "")
+            command_str = command_str.replace(" " + ARGS_MARKER, "")
 
         # Perform search by album and add folders rather than tracks to
         # playlist.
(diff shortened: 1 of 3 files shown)


beetsplug/play.py
```diff
@@ -90,6 +90,9 @@ def play_music(self, lib, opts, args):
                 command_str = command_str.replace(ARGS_MARKER, opts.args)
             else:
                 command_str = u"{} {}".format(command_str, opts.args)
+        else:
+            if ARGS_MARKER in command_str:
```
sampsyo: This condition shouldn't be necessary. `.replace` will just silently leave the string unchanged if it doesn't contain the "needle."

sampsyo: Also, could you please add a short comment here that explains why this is necessary? Posterity will thank you. :smiley:

sampsyo:
Thanks for noticing this! The fix looks great from here.

Could you please add a note to the changelog explaining what went wrong?

tweitzel:
Removed that if clause, added changelog note. Waiting for one of the CI tests to fail for some reason because I'm not that lucky.

sampsyo:
It's your lucky day. :four_leaf_clover: Thank you again!

PR #2146, opened 2016-07-31
install imagemagick on appveyor (via cholatey)

jrobeson:


jrobeson:
This installs IM 7 by default, but most Linux distributions still only use a 6.x version. We could force a 6.x, but if we can make it work with 7.x, then we can avoid conflicting with the `convert` utility already included in Windows.

sampsyo:
Great! I think we should merge this. It looks like IM 7 is mostly backwards compatible on the command line (according to [this document](http://www.imagemagick.org/script/porting.php#cli)), and it doesn't bother me to be testing against two different versions on the two different platforms.

PR #2151, opened 2016-08-01
pass the tests on Windows

jrobeson:


jrobeson:
i'm not really sure i'm gonna be able to go anywhere with this. the windows machine I have access to is far too slow.  I wonder if there's anybody else with a windows machine who can help out.

sampsyo:
I second that! I don't really have access to a Windows machine these days, so we could really use a little help here from someone who has one sitting on their desk.

beetsplug/edit.py
```diff
@@ -243,9 +244,9 @@ def edit_objects(self, objs, fields):
         old_data = [flatten(o, fields) for o in objs]
 
         # Set up a temporary file with the initial data for editing.
-        new = NamedTemporaryFile(mode='w', suffix='.yaml', delete=False)
+        new = NamedTemporaryFile(suffix='.yaml', delete=False)
         old_str = dump(old_data)
-        new.write(old_str)
+        new.write(old_str.encode('utf-8'))
```
jrobeson: breaks python2 :(

jrobeson: specifically 
`UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 7: ordinal not in range(128)`

I guess I'm a little confused as to what's expected here! If you can give me a hand on this one, then we're down to less than 10 total failures!

sampsyo: As I understand it, `yaml.dump` is inconsistent across Python versions? That is, on Python 2, it produces UTF-8-encoded bytes, and on Python 3, it produces unicode. Seems like a strange decision, but OK I guess.

Because the `NamedTemporaryFile` call appears to open the file in binary mode on both versions, you're right to want to produce bytes here. I think we just need something like:

```py
old_str if PY2 else old_str.encode('utf8')
```

to get bytes on both versions.

jrobeson: i didn't even suspect to look at yaml. thanks!

test/rsrc/convert_stub.py
```diff
@@ -20,8 +20,8 @@ def convert(in_file, out_file, tag):
     # On Windows, use Unicode paths. (The test harness gives them to us
     # as UTF-8 bytes.)
     if platform.system() == 'Windows':
-        in_file = in_file.decode('utf8')
-        out_file = out_file.decode('utf8')
+        in_file = in_file.encode('latin-1').decode('utf-8')
+        out_file = out_file.encode('latin-1').decode('utf-8')
```
jrobeson: so obviously we don't wanna do latin-1 here, but the equivalent of `arg_encoding`.  It does break py3 though.

test/test_util.py
```diff
@@ -115,7 +115,7 @@ def popen_fail(*args, **kwargs):
         with self.assertRaises(subprocess.CalledProcessError) as exc_context:
             util.command_output([b"taga", b"\xc3\xa9"])
         self.assertEqual(exc_context.exception.returncode, 1)
-        self.assertEqual(exc_context.exception.cmd, b"taga \xc3\xa9")
+        self.assertEqual(exc_context.exception.cmd, 'taga \xc3\xa9')
```
jrobeson: this breaks Linux on py3 (but not Windows)

jrobeson:
All the windows tests pass, so now I just need a review and some suggestions to get this done.

test/test_util.py
jrobeson: Check the travis output to see.

sampsyo: Right; that makes sense—because on Unixes, the command is now a surrogate-escaped Unicode string, and on Windows, commands are just plain Unicode. Such are the hazards of trying to pretend arguments are Unicode on Unix. :cry:

test/rsrc/convert_stub.py
```diff
@@ -20,8 +27,11 @@ def convert(in_file, out_file, tag):
     # On Windows, use Unicode paths. (The test harness gives them to us
     # as UTF-8 bytes.)
     if platform.system() == 'Windows':
-        in_file = in_file.decode('utf8')
-        out_file = out_file.decode('utf8')
+        if not six.PY2:
+            in_file = in_file.encode(arg_encoding())
+            out_file = out_file.encode(arg_encoding())
+        in_file = in_file.decode('utf-8')
+        out_file = out_file.decode('utf-8')
```
sampsyo: Is there a reason we need an encode followed by a decode on Windows Python 3? Could we instead just pass through the Unicode string?

beets/util/__init__.py
```diff
@@ -737,10 +737,29 @@ def cpu_count():
...
 
 
+def convert_command_args(args):
+    """Convert command arguments to bytes on Python 2 and strings on
+    Python 3."""
+    encoding = arg_encoding()
+    for i, elem in enumerate(args):
+        if six.PY2:
+            if isinstance(elem, six.text_type):
+                args[i] = elem.encode(encoding)
```
sampsyo: Modifying an input array in place is somewhat unexpected. This function should probably return a new array (or be implemented as a generator, with `yield`).

beets/util/__init__.py
```diff
@@ -737,10 +737,29 @@ def cpu_count():
...
+        else:
+            if isinstance(elem, bytes):
+                args[i] = elem.decode(encoding)
+            else:
+                errors = 'surrogateescape'
+                args[i] = elem.encode(encoding, errors).decode(encoding)
```
sampsyo: Can you explain why this needs to encode and then immediately decode again? I don't quite see it…

beets/util/__init__.py
```diff
@@ -737,10 +737,29 @@ def cpu_count():
...
+    """Convert command arguments to bytes on Python 2 and strings on
+    Python 3."""
+    encoding = arg_encoding()
+    for i, elem in enumerate(args):
+        if six.PY2:
+            if isinstance(elem, six.text_type):
+                args[i] = elem.encode(encoding)
+        else:
+            if isinstance(elem, bytes):
+                args[i] = elem.decode(encoding)
```
sampsyo: This seems like it will fail with an error. Shouldn't it use `surrogateescape`?

test/rsrc/convert_stub.py
jrobeson: if you mean no encode/decode at all, then no. That was the very first thing i tried. My first iteration removed the platform check, and it couldn't find the files.  It passed the escape strings right along, and `open` didn't like that.

beets/util/__init__.py
jrobeson: I was just copying the similar code from the test helper. I I will redo that portion as you suggest.

test/rsrc/convert_stub.py
jrobeson: or rather, windows didn't like it.

jrobeson: You can see it here: https://ci.appveyor.com/project/beetbox/beets/build/824/job/226mw7hfg9vdrkto#L64

beets/util/__init__.py
jrobeson: using a generator works on py3, but not 2. the as subprocess seems to only deal with lists, not iterables.

beets/util/__init__.py
jrobeson: yeah. it's good now. not sure how i overlooked it.

beets/util/__init__.py
jrobeson: nope, it's gone!

beets/util/__init__.py
```diff
@@ -737,10 +737,26 @@ def cpu_count():
...
+            if isinstance(arg, six.text_type):
+                arg = arg.encode(arg_encoding())
+        else:
+            if isinstance(arg, bytes):
+                arg = arg.decode(arg_encoding(), 'surrogateescape')
+        return arg
+
+    return [convert(a) for a in args]
+
+
```
jrobeson: is this better?

jrobeson: could be named better if you have any suggestions

test/test_util.py
jrobeson: any suggestions? I can't always tell when we should do python version checks in cases like this.

test/rsrc/convert_stub.py
sampsyo: Whoa; that's crazy. I still don't completely understand why that was happening, but I guess I'll have to live with that…

beets/util/__init__.py
sampsyo: Yes, this looks great. The name seems fine to me, but it seems like the docstring could perhaps use a tiny bit more detail: for example, "On Python 2, this produces byte strings; on Python 3, it enforce surrogate-escaped Unicode strings."

test/test_util.py
sampsyo: I'm actually not quite sure. It would be one thing if this were a straightforward "can't compare bytes and unicode" issue, but something a bit subtler is going on here with the encoding and re-decoding so that we're actually getting different mojibake. I'll have to ponder this—maybe it would (paradoxically) be more straightforward to just use bytes on Unix, even on Python 3? (That might also be a terrible idea!)

jrobeson: It depends on what our goal is for paths. If we plan on moving to pathlib (or otherwise surrogateescaped strings), then I think we should continue treating them as strings here as well. 

if we do that, then we can simplify all sorts of code while keeping our complexity localized.

sampsyo: Yes, I can see that—conforming to the way things are going with Python 3 makes sense. We'll just have to sort out what's going wrong with the escaping for us here.

jrobeson: Here's why:
```
env LANG="en_US" python3
Python 3.4.5 (default, Jul  2 2016, 04:09:35) 
[GCC 6.1.1 20160621 (Red Hat 6.1.1-3)] on linux
>>> print('taga \xc3\xa9')
taga é
```

```
env LANG="en_US.utf-8" python3
Python 3.4.5 (default, Jul  2 2016, 04:09:35) 
[GCC 6.1.1 20160621 (Red Hat 6.1.1-3)] on linux
>>> print('taga \xc3\xa9')
taga Ã©
```
since the default windows cmd codepage on appveyor is mostly equivalent to en_US.iso-8859-1 (aka en_US), you can see why it worked there.

beets/util/__init__.py
jrobeson: I don't think we wanna talk about unicode strings related to python 3. They aren't unicode strings, they are just strings. Other than that though, I'm going to use the wording.

sampsyo: Sure. Sometimes I like to use "text strings" to avoid any ambiguity.

jrobeson: if you're still here, can you cancel https://ci.appveyor.com/project/beetbox/beets/build/842  ?  they broke the default behaviour of `cinst` so the build is hung.

jrobeson: I should mention that I don't have the privileges to do so myself, otherwise I would. Either that I or just can't see any button.

jrobeson:
and now it all passes.

PR #2154, opened 2016-08-04
Move ui._arg_encoding to util.arg_encoding

jrobeson:
We use `_arg_encoding()` for more than UI concerns, so I think it
should go there.

The method is now marked public (the underscore prefix was removed)

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -20,7 +20,6 @@ CLI commands are implemented in the ui.commands module.
 
 from __future__ import division, absolute_import, print_function
 
-import locale
 import optparse
 import textwrap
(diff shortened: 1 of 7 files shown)


sampsyo:
Good call! We could even consider dropping the leading underscore—it'a clearly not very "private" anymore.

jrobeson:
will do!

jrobeson:
should API changes like that be marked in the changelog? If so, what should we call the header?

sampsyo:
IMO, no—usually, I don't bother putting internal refactoring in the changelog.

jrobeson:
i updated the title, and now it's good to go.

PR #2156, opened 2016-08-07
add flake8-coding to our tox config

jrobeson:
Now we can enforce that all py files have the coding magic comment with
utf-8.

https://pypi.python.org/pypi/flake8-coding


jrobeson:
We'd want the reverse (`no-accept-encodings`) when we go all py3

sampsyo:
Looks great! Let's do it.

I'm happy to learn that Python 3's default encoding is UTF-8.

PR #2159, opened 2016-08-07
Read and write edit plugin yaml as utf-8

jrobeson:
I've detached this from #2151 to see how it goes separately.

### DIFF
--- a/beetsplug/edit.py
+++ b/beetsplug/edit.py
@@ -24,6 +24,7 @@ from beets.dbcore import types
 from beets.importer import action
 from beets.ui.commands import _do_query, PromptChoice
 from copy import deepcopy
+import codecs
 import subprocess
 import yaml
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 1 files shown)


jrobeson:
insert meme image of "I don't know what I'm doing" dog.

How we can turn UnicodeWarning into errors for all the tests. I tried stuff that led to `ImportError` happening (trying to find in `__init__.py` in `/path/to/beets` vs /path/to/beets/beets`.

jrobeson:
interesting. I didn't expect it to succeed on appveyor, but I guess other stuff was breaking it.

jrobeson:
I'm pretty sure the lack of the magic coding string in the test was breaking it before.

sampsyo:
This looks great; I didn't know `tempfile` could be given an encoding. I say merge away!

I don't exactly know how to turn only `UnicodeWarning`s into errors in the tests. The Python warning settings are arcane and fiddly and hard to get right. We want something like this: http://stackoverflow.com/a/17211698/39182

But using the `-W` command-line option, here: https://docs.python.org/2/using/cmdline.html#cmdoption-W

Instead of a statement in the Python code.

jrobeson:
i tried the `-W` and it broke in `ImportError`. I saw stuff that suggested fiddling with the nose setup to turn off path adjustment and setting the path manually, but I got scared and ran away.

PR #2173, opened 2016-08-23
Include extra bits in the source distribution

stefanor:
We got [a request](https://bugs.debian.org/775811) to include the zsh completion, in our Debian beets package. To do this, it needs to be shipped in the source tarball.


sampsyo:
Looks great; thanks!

Issue #2172, opened 2016-08-23
Path query tests fail when SQLITE_LIKE_DOESNT_MATCH_BLOBS


stefanor:
Related to #1587.

sqlite 3.10 has an option `SQLITE_LIKE_DOESNT_MATCH_BLOBS` [docs](https://sqlite.org/compile.html#like_doesnt_match_blobs) that breaks the path handling logic in beets. It's a performance optimisation, because LIKE doesn't really seem applicable to blobs, and has historically had bugs. Debian (and therefore Ubuntu) are building sqlite with this option as of 3.10.0-1, so you can see this in stretch and xenial.

`test_query` failures resulting from it:

```
======================================================================
FAIL: test_case_sensitivity (test.test_query.PathQueryTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/build/beets-TC7EW2/beets-1.3.19/test/test_query.py", line 538, in test_case_sensitivity
    self.assert_items_matched(results, [u'path item', u'caps path'])
  File "/build/beets-TC7EW2/beets-1.3.19/test/test_query.py", line 81, in assert_items_matched
    self.assertEqual([i.title for i in results], titles)
AssertionError: Lists differ: [] != [u'path item', u'caps path']

Second list contains 2 additional elements.
First extra element 0:
u'path item'

- []
+ [u'path item', u'caps path']

======================================================================
FAIL: test_escape_backslash (test.test_query.PathQueryTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/build/beets-TC7EW2/beets-1.3.19/test/test_query.py", line 524, in test_escape_backslash
    self.assert_items_matched(results, [u'with backslash'])
  File "/build/beets-TC7EW2/beets-1.3.19/test/test_query.py", line 81, in assert_items_matched
    self.assertEqual([i.title for i in results], titles)
AssertionError: Lists differ: [] != [u'with backslash']

Second list contains 1 additional elements.
First extra element 0:
u'with backslash'

- []
+ [u'with backslash']

======================================================================
FAIL: test_escape_percent (test.test_query.PathQueryTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/build/beets-TC7EW2/beets-1.3.19/test/test_query.py", line 514, in test_escape_percent
    self.assert_items_matched(results, [u'with percent'])
  File "/build/beets-TC7EW2/beets-1.3.19/test/test_query.py", line 81, in assert_items_matched
[...]

Issue #2058, opened 2016-06-20
functemplate AST code needs rewritten for python 3


jrobeson:
The AST code definitely can't work in python3, since the API is different.

sampsyo:
Thanks for the reminder; I'll look into this.

jrobeson:
@sampsyo : Can you look into what it would take to get this done?  I tried my hand at it, but failed.

According to the grammar, it should just be some additional `None`, and `[]` for the kwargs_defaults, and a few other things in `FunctionDef`, but I'm having trouble understanding how `Return` is supposed to work.   I can't seem to find a straightforward example anywhere.  Everybody wants to parse, but few want to create.

jrobeson:
So, i tried again, and this time, succeeded! I finally found some examples via http://nullege.com

sampsyo:
Looks like we're good here!

PR #1596, opened 2015-09-08
Bugfix for cas insensitive searches for a file path

mried:
This will fix #1587 by using the case-lowering of the database instead of Python and SQLite.

PR #2174, opened 2016-08-26
Replace LIKE in PathQuery with SUBSTR matching

jrobeson:
This is to work around the issue pointed out in #2172 where
`BLOB` types can't be used with `LIKE` when sqlite is compiled with
[-DSQLITE_LIKE_DOESNT_MATCH_BLOBS](https://sqlite.org/compile.html#like_doesnt_match_blobs)

NOTE: this prevents such queries from working with (future) indexes, so this will
make it more difficult for us to speed up beets.

We should probably find a way to migrate away from using the blob type for paths

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -57,9 +57,6 @@ class PathQuery(dbcore.FieldQuery):
     and case-sensitive otherwise.
     """
 
-    escape_re = re.compile(br'[\\_%]')
-    escape_char = b'\\'
-
     def __init__(self, field, pattern, fast=True, case_sensitive=None):
         """Create a path query. `pattern` must be a path, either to a
         file or a directory.
@@ -108,20 +105,17 @@ class PathQuery(dbcore.FieldQuery):
         return (path == self.file_path) or path.startswith(self.dir_path)
 
     def col_clause(self):
-        if self.case_sensitive:
-            file_blob = BLOB_TYPE(self.file_path)
-            dir_blob = BLOB_TYPE(self.dir_path)
-            return '({0} = ?) || (substr({0}, 1, ?) = ?)'.format(self.field), \
-                   (file_blob, len(dir_blob), dir_blob)
+        file_blob = BLOB_TYPE(self.file_path)
+        dir_blob = BLOB_TYPE(self.dir_path)
 
-        escape = lambda m: self.escape_char + m.group(0)
-        dir_pattern = self.escape_re.sub(escape, self.dir_path)
-        dir_blob = BLOB_TYPE(dir_pattern + b'%')
-        file_pattern = self.escape_re.sub(escape, self.file_path)
-        file_blob = BLOB_TYPE(file_pattern)
-        return '({0} LIKE ? ESCAPE ?) || ({0} LIKE ? ESCAPE ?)'.format(
-            self.field), (file_blob, self.escape_char, dir_blob,
-                          self.escape_char)
+        if self.case_sensitive:
+            query_part = '({0} = ?) || (substr({0}, 1, ?) = ?)'
+        else:
+            query_part = '(UPPER({0}) = UPPER(?)) || \
+                         (substr(UPPER({0}), 1, ?) = UPPER(?))'
+
+        return query_part.format(self.field), \
+            (file_blob, len(dir_blob), dir_blob)
 
 
 # Library-specific field types.


jrobeson:
Fedora doesn't build SQLite with this option, so I rebuilt the package myself with it.  It broke in the same way as mentioned in #2172 before I wrote the patch.

jrobeson:
@sampsyo : i'm afraid that as per https://github.com/beetbox/beets/pull/1596 blob types for paths are a lost cause.

sampsyo:
Oh no; sorry for the duplicate work, @jrobeson. Alternatives include fashioning our own case normalization and storing pre-normalized paths.

I'm concerned that storing surrogate escaped Unicode in the database might not work—Python uses the invalid codepoints U+DC80..U+DCFF for surrogates, and SQLite would be within its rights to reject them. We'd need to check.

jrobeson:
I'll take a look.   but what do you think about the current fix?

sampsyo:
Ah, right---it looks like all the tests pass, but I would have expected the problem from https://github.com/beetbox/beets/pull/1596#issuecomment-138680936 here. Specifically, won't `UPPER({0}) = UPPER(?)` crash if SQLite can't decode either of the blobs?

jrobeson:
ah, I saw somwhere that mentioned that sqlite only supported  utf-8/utf-16 text anyways, so I thought we'd already have run into trouble with that sort of thing by now in other areas. But here's what it actually says:

https://www.sqlite.org/version3.html
> SQLite is not particular about the text it receives and is more than happy to process text strings that are not normalized or even well-formed UTF-8 or UTF-16. Thus, programmers who want to store IS08859 data can do so using the UTF-8 interfaces. As long as no attempts are made to use a UTF-16 collating sequence or SQL function, the byte sequence of the text will not be modified in any way.

jrobeson:
However, the db encoding is always going to be one of utf-8, utf-16-be, or utf-16-le . There aren't any other choices.

jrobeson:
so , imo.. we should probably always set it to utf-8, so our databases are portable.

jrobeson:
on case folding: We are definitely going to have a provide a case folding implementation!

jrobeson:
So, i guess there is no quick fix to this is there?

sampsyo:
I suppose not, unless we can somehow prove my concern is unfounded (i.e., UPPER on a blob can never fail). I haven't been able to ascertain that from the docs.

:cry:

jrobeson:
```python
import sqlite3
conn = sqlite3.connect('example.db')
c = conn.cursor()
c.execute('CREATE TABLE foo (bar text)')
latin1 = b'caf\xe9'.decode('latin1')
c.execute('INSERT INTO foo VALUES (?)', (latin1,))
big5 = b'OK\r\n\xcf\xb5\xcd\xb3\xd0\xc5\xcf\xa2\xa3\xba\r\n\xb2\xa8\xcc\xd8\xc2\xca\xa3\xba115200\r\n\xc4\xbf\xb1\xea\xb6\xd4\xb7\xbd\xb5\xd8\xd6\xb7\xa3\xba 0xFF,0xFF,0xFF,0xFF,0xFF\r\n\xb1\xbe\xb5\xd8\xbd\xd3\xca\xd5\xb5\xd8\xd6\xb70\xa3\xba0xFF,0xFF,0x'.decode('big5')

all = c.execute('SELECT UPPER(bar) FROM foo').fetchall()
print(all)

[('CAFé',), ('OK\r\n炵苀陓洘ㄩ\r\n疏杻薹ㄩ115200\r\n醴梓勤源華硊ㄩ 0XFF,0XFF,0XFF,0XFF,0XFF\r\n掛華諉彶華硊0ㄩ0XFF,0XFF,0X',)]
```

I got the big5 bytes from this usb serial dongle output (via pyserial)

sampsyo:
Huh, that's not bad! Looks like it replaced errors with ㄩ?

jrobeson:
I'm not sure, but i don't think it's sqlte doing it as 

```python
print(big5)
print(big5)
OK
炵苀陓洘ㄩ
疏杻薹ㄩ115200
醴梓勤源華硊ㄩ 0xFF,0xFF,0xFF,0xFF,0xFF
掛華諉彶華硊0ㄩ0xFF,0xFF,0x

```

jrobeson:
I might have this all wrong, but I don't think we're ever going to get decent case folding unless we make sure all our input is the same encoding, and we provide case folding.

jrobeson:
I didn't clarify that i ran that all on python3, not 2.

jrobeson:
OH, and here's a translated example of the output

 ```
 Information about the system: 
 Baud Rate: 115200 
 Location Address: 0x11, 0x22, 0x33, 0x44, 0x55 
 Get endpoint address 00:00 x11, 0x22, 0x33, 0x44, 0x55 
```

**Comment by jrobeson** [comment]
but you are right about doing it on blobs! I created a new table with blob type and inserted the bytes directly instead of text.

`sqlite3.OperationalError: Could not decode to UTF-8 column 'UPPER(quux)' with text 'CAF�'`

**Comment by sampsyo** [comment]
Ah, I got it: looks like we're out of luck in that case. I concur that there's no easy fix here, tragically.

**Comment by jrobeson** [comment]
https://www.mercurial-scm.org/wiki/EncodingStrategy   . Something I found today.

**Comment by sampsyo** [comment]
Thanks! This USA useful summary of the situation. Operatively, they conclude:

> there exists no general, portable solution to the makefile problem.

(i.e., the file name encoding problem). They do the pragmatic thing and err on the side of Unix-like compatibility.

**Comment by jrobeson** [comment]
USA?

**Comment by jrobeson** [comment]
So, a solution for now could be a custom function. so we can handle it on the python side.

**Comment by sampsyo** [comment]
Oops! "Is a." Relying heavily on autocorrect to type atm...

Yeah, that would definitely do it.

**Comment by jrobeson** [comment]
So this does seem to work:

```python
import sqlite3
conn = sqlite3.connect('example.db')
conn.create_function('bytelower', 1, lambda b: b.lower())
c = conn.cursor()
c.execute('CREATE TABLE foo (bar blob)')
c.execute('INSERT INTO foo VALUES (?)', (b'caf\xe9',))
all = c.execute('SELECT BYTELOWER(bar) FROM foo').fetchall()
print(all)
[(b'caf\xe9',)]
```

So where do you want it?

sampsyo:
Cool! I'm glad it works.

Installing custom functions is tricky to modularize. I suppose it should go in the Library class's setup code (not dbcore's) unless you have a better idea.

jrobeson:
fixed in #2058

PR #2175, opened 2016-08-26
mpdupdate: Use `$MPD_HOST` environment variable.

tarruda:
This variable is not standardized, but is already used by the mpc which is the official command-line client. It would be nice if beets read it too, to avoid duplicate configuration for mpc users.


nathdwek:
Much appreciated!
Could you please amend your commit in order to:
* Link to somewhere supporting the usage of this variable in the commit message (for example http://linux.die.net/man/1/mpc).
* Similarly to how mpc explains it, explicitly mention in the doc the precedence of the variable.

tarruda:
Done

nathdwek:
Thank you very much!

PR #2176, opened 2016-08-29
Compare case insensitive paths with SUBSTR and BYTELOWER

jrobeson:
Create a custom `bytelower` sqlite function so we can compare bytestrings in a semi
case insensitive fashion.  This is to work around sqlite builds are that compiled
with `-DSQLITE_LIKE_DOESNT_MATCH_BLOBS`.

See https://github.com/beetbox/beets/issues/2172 for details.

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -57,9 +57,6 @@ class PathQuery(dbcore.FieldQuery):
     and case-sensitive otherwise.
     """
 
-    escape_re = re.compile(br'[\\_%]')
-    escape_char = b'\\'
-
     def __init__(self, field, pattern, fast=True, case_sensitive=None):
         """Create a path query. `pattern` must be a path, either to a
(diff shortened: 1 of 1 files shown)


jrobeson:
so while it does work, I'm not yet usre if connections are the keeper of functions, or the db as a whole. If it's the db as a whole, then the connection should probably explicitly be added to `connection[0]` or be moved into dbcore

jrobeson:
so.. it breaks on py2 with

`AttributeError: 'buffer' object has no attribute 'lower'` :(

sampsyo:
:confused: I suppose we need to convert the buffer to str, call lower, and convert back.

jrobeson:
done.

sampsyo:
Nice work! :sparkles: Since the tests pass, I say merge it. Thank you!!

PR #2179, opened 2016-08-29
Correct a typo

stefanor:
Spotted by [lintian](https://lintian.debian.org/full/python-apps-team@lists.alioth.debian.org.html#beets_1.3.19-1) which runs spellchecks on man pages.


sampsyo:
Looks good; thanks!

PR #2180, opened 2016-08-30
PY3 compatibility for the functemplate AST compiler

### LINKED ISSUE #2058: functemplate AST code needs rewritten for python 3

jrobeson:
Highlights:
  * Conditional Python 3 compatble `ast.FunctionDef` and `ast.Call` calls
  * Python 2 and 3 compatible string types for `ex_rvalue`
  * Only encode strings to `utf-8` on Python2, as they are `unicode` type there. 
  * Enable template compilation on Python 3

### DIFF
--- a/beets/util/functemplate.py
+++ b/beets/util/functemplate.py
@@ -534,8 +534,7 @@ class Template(object):
     def __init__(self, template):
         self.expr = _parse(template)
         self.original = template
-        if six.PY2:  # FIXME Compiler is broken on Python 3.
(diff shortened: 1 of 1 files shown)


jrobeson:
I definitely wanna redo the commit messages before merging.

sampsyo:
Amazing! Sorry I was unresponsive on this, but these fixes look just right from here.

PR #2118, opened 2016-07-07
Add gstreamer deps to travis config

jrobeson:
[appveyor skip]

PR #2183, opened 2016-09-01
add site-packages tests to the travis build matrix

jrobeson:
Now we can test any plugin that depends on Gstreamer. It was previously impossible to do so, since `pygobject` (a gstreamer dep) is not installable via pip. 

I discovered (after totally missing it it in the quodlibet `.travis.yml` for months)  that travis offers `with_site_package` variants of the 2 pythons (one 2.7, one 3.x) that come with the distribution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,11 @@ matrix:
     include:
         - python: 2.7
           env: {TOX_ENV: py27-cov, COVERAGE: 1}
+        - python: 2.7_with_system_site_packages
+          env: {TOX_ENV: py27-test}
         - python: 3.4
+          env: {TOX_ENV: py34-test}
+        - python: 3.4_with_system_site_packages
           env: {TOX_ENV: py34-test}
         - python: 3.5
           env: {TOX_ENV: py35-test}
@@ -51,7 +55,13 @@ install:
     - travis_retry pip install tox sphinx
     - travis_retry tox -e $TOX_ENV --notest
 
-script: tox -e $TOX_ENV
+script:
+    # prevents "libdc1394 error: Failed to initialize libdc1394" errors
+    - sudo ln -s /dev/null /dev/raw1394
+    - if [[ $TRAVIS_PYTHON_VERSION == *_site_packages ]]; then SITE_PACKAGES=--sitepackages; fi
+    # pip in trusty breaks on packages prefixed with "_". See https://github.com/pypa/pip/issues/3681
+    - if [[ $TRAVIS_PYTHON_VERSION == 3.4_with_system_site_packages ]]; then sudo rm -rf /usr/lib/python3/dist-packages/_lxc-0.1.egg-info; fi
+    - tox -e $TOX_ENV $SITE_PACKAGES
 
 # Report coverage to codecov.io.
 before_install:


sampsyo:
Is the idea to catch interactions with packages Travis has pre-installed?

jrobeson:
Two reasons:
  * catch issues when beets is installed as a distro package (although 14.04 is getting a little old for this)
  * make our tests work with pygobject dependent stuff (like gstreamer)

jrobeson:
see the discussion here: https://github.com/beetbox/beets/pull/2118

jrobeson:
that did remind me i forgot to add those gstreamer deps, which is part of the reason this didn't handle gstreamer correctly.

sampsyo:
Ah, the pygobject thing definitely makes sense; good point. I'd recommend documenting this thoroughly in YAML comments for posterity.

Might it also make sense to call that environment variable TOX_ARGS?

jrobeson:
I will, whenever i make it work!

sampsyo:
Awesome! :sparkles:

jrobeson:
I got an ubuntu 14.04 vm and realized that this is not going to work. the version of mutagen and six are too old.  The six problem will disappear eventually, but not the mutagen issue.

jrobeson:
ah, to be clear, 14.04 ships with mutagen 1.22.

jrobeson:
I didn't realize how advanced our mutagen dependency was. We can't even run against the currently packaged version of mutagen in Ubuntu 16.10.  I guess I'll just keep the focus on making gobject work.

sampsyo:
Sounds good. Yeah, we tend to track mutagen pretty closely---I've reasoned that users installing beets from pip can also install its dependencies from pip.

jrobeson:
If we track mutagen so closely, we might never be able to be packaged for various distros.

lazka:
I always have a recent mutagen in the quodlibet PPA, if needed: https://launchpad.net/~lazka/+archive/ubuntu/ppa

jrobeson:
i finally got some semblence of this working! It does show that we're broken (at least in the tests) on gstreamer backend of replaygain on py3!   I'm not yet sure what's causing it though

jrobeson:
and we have a segfault.. "test_cli_saves_album_gain_to_file (test.test_replaygain.ReplayGainGstCliTest) ... Segmentation fault (core dumped)
"

jrobeson:
and segfault fixed!  so this is all good now. EXCEPT THE UGLINESS IN `.travis.yml`!

jrobeson:
So, now that this works, we should discuss whether we want tox envs for site package testing.

jrobeson:
@sampsyo: you mentioned that you wanted me to rename the `SITE_PACKAGES` variable, but now since it's internalized to the test step and can only mean one thing, then what would you suggest. i don't see `TOX_ARGS` as being a good choice.  If we integrate it more into tox, then the variable might be irrelevant anyways.

jrobeson:
looking at the history I seem to have finally completed @geigerzaehler's dream here! https://github.com/beetbox/beets/commit/4f844dfb925efd6fcc7ca9e6df09da627b2704fc

sampsyo:
Wow! Just two and a half years later!

That's some amazing detective work. Thank you for tracking down those obscure bugs.

I totally agree that my earlier suggestion is now invalid; a special-purpose "site packages" flag makes sense if we need these one-off hacks to make that mode work.

Personally, I don't see a great value to testing with the system-provided Python packages from Trusty, *except for GStreamer*, the one thing we can't seem to install inside Tox. But that's worth it by itself.

jrobeson:
that's all we're doing atm.. we just get the other junk by association. I wonder if it'd be better to try to hack into tox and symlink it, but we'd only want to do that on systems we can guarantee it exists, or make it run a shell script to detect it or something?

jrobeson:
so maybe it'd be better to just say if folks are hacking on gstreamer dependent things to use the `--sitepackage` arg to tox vs going all out on this?

If so, then is this ready to merge?

sampsyo:
Agreed; no need to go crazy making this reusable on dev machines. So yes, merge away!

PR #1484, opened 2015-05-31
[WIP] Port CLI to Click

untitaker:
cont. #1446

Issue #1448, opened 2015-05-03
TypeError on windows 8.1 when using fetchart plugin


noc7c9:
```
Traceback (most recent call last):
  File "C:\Python27\Scripts\beet-script.py", line 9, in <module>
    load_entry_point('beets==1.3.13', 'console_scripts', 'beet')()
  File "C:\Python27\lib\site-packages\beets\ui\__init__.py", line 1104, in main
    _raw_main(args)
  File "C:\Python27\lib\site-packages\beets\ui\__init__.py", line 1094, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "C:\Python27\lib\site-packages\beets\ui\commands.py", line 893, in import_func
    import_files(lib, paths, query)
  File "C:\Python27\lib\site-packages\beets\ui\commands.py", line 870, in import_files
    session.run()
  File "C:\Python27\lib\site-packages\beets\importer.py", line 316, in run
    pl.run_parallel(QUEUE_SIZE)
  File "C:\Python27\lib\site-packages\beets\util\pipeline.py", line 301, in run
    out = self.coro.send(msg)
  File "C:\Python27\lib\site-packages\beets\util\pipeline.py", line 183, in coro
    func(*(args + (task,)))
  File "C:\Python27\lib\site-packages\beets\importer.py", line 1343, in plugin_stage
    func(session, task)
  File "C:\Python27\lib\site-packages\beets\plugins.py", line 123, in wrapper
    return func(*args, **kwargs)
  File "C:\Python27\lib\site-packages\beetsplug\fetchart.py", line 348, in fetch_art
    path = self.art_for_album(task.album, task.paths, local)
  File "C:\Python27\lib\site-packages\beetsplug\fetchart.py", line 443, in art_for_album
    if self._is_valid_image_candidate(candidate):
  File "C:\Python27\lib\site-packages\beetsplug\fetchart.py", line 412, in _is_valid_image_candidate
    size = ArtResizer.shared.get_size(candidate)
  File "C:\Python27\lib\site-packages\beets\util\artresizer.py", line 153, in shared
    cls._instance = cls()
  File "C:\Python27\lib\site-packages\beets\util\artresizer.py", line 165, in __init__
    self.method = self._check_method()
  File "C:\Python27\lib\site-packages\beets\util\artresizer.py", line 219, in _check_method
    version = has_PIL()
  File "C:\Python27\lib\site-packages\beets\util\artresizer.py", line 249, in has_PIL
[...]

PR #2190, opened 2016-09-08
[WIP] Click Take 2

jrobeson:
This is a rebased version of #1484.  It's still massively in progress.

I might squash it down even more, but I'm still trying to get a sense of what we're doing here.

I hope it didn't lose anything too valuable in the process.

TODO: add a checklist HERE!


beets/ui/__init__.py
```diff
@@ -1145,9 +1157,8 @@ def _configure(options):
     # Add any additional config files specified with --config. This
     # special handling lets specified plugins get loaded before we
     # finish parsing the command line.
-    if getattr(options, 'config', None) is not None:
-        config_path = options.config
-        del options.config
+    config_path = options.pop('config', None)
+    if config_path is not None:
```
jrobeson: @sampsyo : what do you think about the confit patch that made it possible? Should we apply it to master?

beets/ui/commands.py
```diff
@@ -1512,8 +1511,7 @@ def move_func(lib, opts, args):
         if not os.path.isdir(dest):
             raise ui.UserError(u'no such directory: %s' % dest)
 
-    move_items(lib, dest, decargs(args), opts.copy, opts.album, opts.pretend,
-               opts.timid)
```
jrobeson: was timid removed here on purpose?

jrobeson:
I could push all this rebased stuff back into #1448 if folks would prefer. I'm not sure why i didn't ask about that first :)

sampsyo:
Yes, it would be nice to keep this stuff in the click branch if you don't mind—that way, everyone can edit and see their changes reflected in the PR.

beets/ui/commands.py
sampsyo: I don't think so—maybe timid was added after click branched?

beets/ui/__init__.py
sampsyo: Sorry for being slow, but which patch exactly?

beets/util/confit.py
```diff
@@ -248,7 +248,11 @@ def set_args(self, namespace):
         like argparse or optparse, onto this view's value.
         """
         args = {}
-        for key, value in namespace.__dict__.items():
+        if isinstance(namespace, dict):
+            items = namespace.items()
+        else:
+            items = namespace.__dict__.items()
+        for key, value in items:
```
jrobeson: this is the patch.  seems to let you use it like `foo.bar` as well as `foo['bar']`.

jrobeson:
so you're cool with the squashings I've made so far?

jrobeson:
BTW: https://github.com/blog/2247-improving-collaboration-with-forks. I still agree with it being hosted in beets proper though.

jrobeson:
I still plan on squashing it more though, so that means it will constantly break on pull . so I'd like to clean it up a bit more.

sampsyo:
Yes, looks fine so far.

test/test_convert.py
```diff
@@ -197,8 +192,7 @@ def test_embed_album_art(self):
         with open(os.path.join(image_path), 'rb') as f:
             image_data = f.read()
 
-        with control_stdin('y'):
-            self.run_convert()
+        self.run_convert(input='y')
```
jrobeson: I'm thinkin that we should make these sort of changes in master if this is how we want the API to behave.  It shouldn't be hard to emulate it right? alternatively, we make attempt to make click act like the original.

beets/util/confit.py
jrobeson: So do you want this to be in confit ? If so, I'd like to get that in there.

sampsyo: Yes, let's add this. It would be nice to update the docstring, though.

beets/ui/__init__.py
jrobeson: i totally misunderstood where this came from, so you weren't being slow. :(

jrobeson:
ok, i tried, but I don't think there's a way to get a consistent story out of this without too much more time. I think I'll have to squash it up more and then extract it.

jrobeson:
so, i massively squashed and cleaned this up.  I'm not sure if you'll find it acceptable now, but if you do, then I'll commit it to the main click branch.

It shouldn't hopefully take too much more work now, but I'm gonna need help.

There's so many callbacks passing to callbacks that I'm having trouble figuring out stuff like this atm:

```sh
  File "/home/johnny/src/beets/beetsplug/info.py", line 207, in run
    fmt = ui.decargs([opts.format])[0] if opts.format else None
AttributeError: 'Values' object has no attribute 'format'
```

It seems to ok when it gets into the `LegacyOptionParser`, but lost after that.

I think there are a few things like that, then it seems like it would pass the tests.

on py3.5: `FAILED (SKIP=21, errors=24, failures=13)`

jrobeson:
so the `Values` not having `format` seems to have to do with the fact that `-f` is marked with `expose_value=False`.   Although it still doesn't catch the value even it the toggle flipped.

jrobeson:
@sampsyo : can i merge this back into click?

jrobeson:
@sampsyo ?

beets/ui/__init__.py
```diff
@@ -939,10 +939,12 @@ def beet(ctx, **options):
     # Configure the MusicBrainz API.
     mb.configure()
 
-    # Open the library database and set it on the context.
-    lib = _open_library(config)
-    plugins.send('library_opened', lib=lib)
-    beets_ctx.lib = lib
+    # Open the library database and set it on the context
+    # if we didn't already get it from the test suite
```
sampsyo: This sentence needs a period.

sampsyo:
Yes! Looks good to me.

PR #2191, opened 2016-09-09
use a Click style Result object for run_with_output

jrobeson:
This is like http://click.pocoo.org/6/testing/#basic-testing

Currently only `output` is implemented. Others can be added as needed.

This should make a ton of our tests align with click's `invoke` method output

### DIFF
--- a/test/helper.py
+++ b/test/helper.py
@@ -40,6 +40,7 @@ import shutil
 import subprocess
 from tempfile import mkdtemp, mkstemp
 from contextlib import contextmanager
+from collections import namedtuple
 from six import StringIO
 from enum import Enum
 
@@ -447,7 +448,10 @@ class TestHelper(object):
     def run_with_output(self, *args):
         with capture_stdout() as out:
             self.run_command(*args)
-        return util.text_string(out.getvalue())
+
+        result = namedtuple('Result', ['output'])
+        result.output = util.text_string(out.getvalue())
+        return result
 
     # Safe file operations
 
--- a/test/test_info.py
+++ b/test/test_info.py
@@ -45,7 +45,8 @@ class InfoTest(unittest.TestCase, TestHelper):
         mediafile.composer = None
         mediafile.save()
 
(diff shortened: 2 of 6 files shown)


jrobeson:
If we adopt this, then I plan on moving on to changing the `with capture_stdout() as output`  type calls too.

sampsyo:
Seems fine!  I don't have a strong feeling either way about whether we move to the result tuple before the click transition, but centralizing these standard-output capturing seems like a great idea for maintainability.

jrobeson:
I wouldn't wanna merge it just because it's like click and that we're attempting to transition there.  I wanna merge it if you think the click testing style is good in and of itself. If so, I'd like to adopt it either way.

sampsyo:
Aha. Most of the changes seem quite reasonable, but the single-element tuple return type is a little odd out of context. That's the only change that seems somewhat weird without all of quick.

jrobeson:
I plan on adding the raised exception too, and the exit code if possible.

sampsyo:
Yep, that makes sense, especially if tests use that information.

jrobeson:
I'm having a little trouble figuring out what we're trying to achieve by tests like https://github.com/beetbox/beets/blob/master/test/test_ui.py#L55  compared to using `run_with_output` and ls/list directly.

There's no real space between `list_items` the actual list command as seen here:  https://github.com/beetbox/beets/blob/master/beets/ui/commands.py

In that case, we already have quite a few ls tests test_ui.

jrobeson:
I guess the problem is that we tend to format our output and print it to stdout at the same time.

jrobeson:
i totally forgot about the whole other can of worms with DummyIO, as it's gonna be a similar problem.

sampsyo:
Those direct calls to the command functions are sometimes nice: they are narrower unit tests that avoid exercising irrelevant code like command line parsing.

jtpavlock:
Is this still relevant?

PR #2193, opened 2016-09-10
Use -k for each arg in duplicates plugin

jrobeson:
This was extracted from the click branch. I'm not sure how we could make this BC, but if you have any ideas, then let me know.

Also, please suggest any wording changes.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -18,6 +18,12 @@ to know:
 * Please don't package beets as a Python 3 application *yet*, even though some
   things may appear to work under Python 3.4 and later.
 
+UI Change:
+
+* :doc:`/plugins/duplicates` no longer accepts multiple arguments in the form
+  of ``-k title albumartist album``. Each argument must be prefixed with
+  ``-k``. Example: ``-k title -k albumartist -k album``
+
 The are a couple of small new features:
 
 * :doc:`/plugins/mpdupdate`, :doc:`/plugins/mpdstats`: When the ``host`` option
--- a/docs/plugins/duplicates.rst
+++ b/docs/plugins/duplicates.rst
@@ -28,7 +28,7 @@ duplicates themselves via command-line switches ::
   -d, --delete          delete items from library and disk
   -F, --full            show all versions of duplicate tracks or albums
   -s, --strict          report duplicates only if all attributes are set
-  -k, --keys            report duplicates based on keys
+  -k, --key            report duplicates based on keys (can be used multiple times)
   -M, --merge           merge duplicate items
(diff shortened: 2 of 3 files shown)


docs/plugins/duplicates.rst
```diff
@@ -46,8 +46,8 @@ file. The available options mirror the command-line options:
 - **checksum**: Use an arbitrary command to compute a checksum
   of items. This overrides the ``keys`` option the first time it is run;
   however, because it caches the resulting checksum as ``flexattrs`` in the
-  database, you can use ``--keys=name_of_the_checksumming_program
-  any_other_keys`` (or set configuration ``keys`` option) the second time
+  database, you can use ``--key=name_of_the_checksumming_program
+  --key=any_other_keys`` (or set configuration ``keys`` option) the second time
```
sampsyo: This isn't your fault, but the parenthetical should probably read "or set the keys configuration option".

sampsyo:
I like this change; it not only makes the code simpler but also makes the interface more predictable in my opinion. The only concern, as you noted, is BC—and I'm inclined to say that we should just go ahead with the change, since I can't imagine there are too many scripts out there relying on the old behavior.

docs/plugins/duplicates.rst
jrobeson: so other than this, it's good to go?

docs/changelog.rst
```diff
@@ -18,6 +18,12 @@ to know:
 * Please don't package beets as a Python 3 application *yet*, even though some
   things may appear to work under Python 3.4 and later.
 
+UI Change:
```
jrobeson: this header doesn't match up with the others. Other headers are sentences.

jrobeson:
I hope you like it, as you're the author :)

docs/plugins/duplicates.rst
sampsyo: Yep!

docs/changelog.rst
sampsyo: Good point. I usually massage the release notes quite a bit before doing the final release anyway, so don't worry about it too much.

sampsyo:
Ha! I didn't mean to give myself a compliment. I just have a terrible memory, apparently...

docs/plugins/duplicates.rst
jrobeson: i'm going to change it after i merge it, as it's unrelated to the PR.

sampsyo: Good idea. Thanks for cleaning up someone else's writing.

PR #2198, opened 2016-09-17
Add python 3.6-dev to Travis

jrobeson:
This may or may not be available in the installed pyenv yet.


jrobeson:
@sampsyo : codecov is reporting in comments again.

PR #2204, opened 2016-09-22
stretch InfoGroup's search quota

ftiantian:


jrobeson:
If you're trying to update attachments, then it'd be best if you rebased it on top of master instead of merging.

sampsyo:
Looks like a misfire.

PR #2206, opened 2016-09-23
Issue2205: search for lyrics using part of song title preceding ':'

Kraymer:


jrobeson:
Can you add a changelog entry too?

beetsplug/lyrics.py
```diff
@@ -144,29 +144,32 @@ def search_pairs(item):
...
     # Remove any featuring artists from the title
     pattern = r"(.*?) {0}".format(plugins.feat_tokens(for_artist=False))
-    for title in titles[:]:
-        match = re.search(pattern, title, re.IGNORECASE)
-        if match:
-            titles.append(match.group(1))
+    for title in list(titles):
+        titles.add(strip_part(title, pattern))
+
+    # Remove part of the title string after colon ':'
```
sampsyo: This comment could use a tiny bit more detail: in particular, that this helps with songs with subtitles.

sampsyo:
Nice! That was fast! This looks great to me. And I like the  refactoring too.

beetsplug/lyrics.py
Kraymer: done

Kraymer:
I've changed the implementation slightly, do you mind having another look on it ?

beetsplug/lyrics.py
```diff
@@ -144,29 +145,37 @@ def search_pairs(item):
...
-            titles.append(match.group(1))
+    for title in list(titles):
+        titles.append(strip_part(title, pattern))
+
+    # Remove part of the title string after colon ':' for songs with subtitles
+    titles.append(strip_part(title, r"(.+?)\s*:.*"))
+
+    # Yield artist and title obtained from item first
+    titles = list(OrderedDict.fromkeys([title] + titles))
+    artists = list(OrderedDict.fromkeys([artist] + artists))
```
sampsyo: Is the idea with the `OrderedDict` to remove duplicates while preserving the order? That makes sense, but a couple of improvements come to mind:

* First, the original `title` and `artist` are already first in the list (from the initialization, above).
* We could also just avoid inserting duplicates in the first place. Specifically, `strip_part` could return None instead of the original string when the pattern doesn't match, and we could avoid appending that to the list (or just filter out Nones from the list).

Kraymer: I moved the burden to update the alternatives lists to the function that performs the `re.search` and think it works better that way : it avoids the *return None+filter* processing and makes caller code simpler.

sampsyo: Nice! This is even better. Using a list of patterns makes this declarative, which avoids repetition and a whole vector of potential bugs. :sparkles:

Thanks! I say merge when ready.

PR #2209, opened 2016-09-25
Remove deprecated top level colors config option

jrobeson:
It was deprecated in https://github.com/beetbox/beets/commit/d3fce35481f09f28561dc2438baa7ec5245b7870 on 2015-01-26.

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -1121,14 +1121,6 @@ def _configure(options):
     else:
         log.set_global_level(logging.INFO)
 
-    # Ensure compatibility with old (top-level) color configuration.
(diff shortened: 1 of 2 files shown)


jrobeson:
If you're cool with this, I'd also like to do the same for the deprecated list_format stuff too from https://github.com/beetbox/beets/commit/f14f47f059459118a5d3f9123c5f72a948ac3a4c

jrobeson:
I think being deprecated for over a year is pretty good.

sampsyo:
Removing these sounds fine. They mostly stuck around because they weren't costing us anything, but if either or both of these is getting in the way of Click (for example), that's a good reason to remove them.

PR #2211, opened 2016-09-26
Proposal: Add specificity option to lastgenre plugin

heylookltsme:
(Apologies in advance for my terrible, terrible python. I really wanted this feature, so I hacked my way through it. Please provide guidance on the better ways to do things!)

### What Is This?
I love the lastgenre plugin, but after running it on my library, I found myself wishing that it would prefer the _most specific_ tag available instead of simply defaulting to the _most popular_ tag, which is often the most generic. 

Here's an example. I have both "folk" and "americana" on my whitelist and in my canonicalization tree, where "americana" is a leaf within "folk." 

If I run lastgenre for the artist "The Be Good Tanyas", the genres that last.fm returns are: 
```
[u'folk', u'albums i own', u'female vocalists', u'alt-country', u'country', u'singer-songwriter', u'americana', u'canadian', u'2003', u'00s']
```

You see that both `folk` and `americana` are present. Now, the default behavior of lastgenre is to just use `folk`, as it's the most popular. However, since `americana` is present in my whitelist, I'd rather use that as the genre as it's more specific and thus more useful for categorization. 

Note, I **do not** want to remove `folk` from my whitelist because there are cases where a more specific tag is not available. `Folk` is still a useful genre to me; I just want to be able to use more specific genres _when_ they're available. 

Does all that make sense? 

### What I Changed
So to achieve this, I added a `specificity` option (defaults to `False`, and I'm definitely open to suggestions on the name) which, when set to `True` will sort genres by their depth in the genre tree from most to least deep before determining which genre(s) to use. That's pretty much it. 

Like I said at the top, I don't really know any python so I'm positive there are much more elegant ways to do what I did. Please school me! 

Also, I'm new to contributing to the project, so please let me know if there's anything else I need to do. Cheers!

### DIFF
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -200,8 +200,8 @@ class LastGenrePlugin(plugins.BeetsPlugin):
                 tags_all += parents
                 # Stop if we have enough tags already, unless we need to find
                 # the most specific tag (instead of the most popular).
-                if (not self.config['prefer_specific']
-                        and len(tags_all) >= count):
+                if (not self.config['prefer_specific'] and
+                        len(tags_all) >= count):
                     break
             tags = tags_all
 
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -109,6 +109,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
             'force': True,
             'auto': True,
             'separator': u', ',
+            'specificity': False,
         })
 
         self.setup()
@@ -158,6 +159,27 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         elif source == 'artist':
             return 'artist',
 
+    def _get_depth(self, tag):
+        """Find the depth of a tag in the genres tree.
+        """
+        depth = None
+        for key, value in enumerate(self.c14n_branches):
+            if tag in value:
+                depth = value.index(tag)
+                break
+        return depth
+
+    def _sort_by_depth(self, tags):
+        """Given a list of tags, sort the tags by their depths in the
+        genre tree.
+        """
+        tags_by_depth = []
+        for key, value in enumerate(tags):
+            depth = self._get_depth(value)
+            tags_by_depth.append({'tag': value, 'depth': depth})
+        tags_by_depth = sorted(tags_by_depth, key=lambda k: k['depth'], reverse=True)
+        return [i['tag'] for i in tags_by_depth]
+
     def _resolve_genres(self, tags):
         """Given a list of strings, return a genre by joining them into a
         single string and (optionally) canonicalizing each.
@@ -179,11 +201,17 @@ class LastGenrePlugin(plugins.BeetsPlugin):
                     parents = [find_parents(tag, self.c14n_branches)[-1]]
 
                 tags_all += parents
-                if len(tags_all) >= count:
+                # Don't break here if we need to sort by specificity, which happens with
+                # the full tag list below
+                if not self.config['specificity'] and len(tags_all) >= count:
                     break
             tags = tags_all
 
         tags = deduplicate(tags)
+
+        # sort the tags by specificity
+        if self.config['specificity']:
+            tags = self._sort_by_depth(tags)
 
         # c14n only adds allowed genres but we may have had forbidden genres in
         # the original tags list
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -173,12 +173,9 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         """Given a list of tags, sort the tags by their depths in the
         genre tree.
         """
-        tags_by_depth = []
-        for key, value in enumerate(tags):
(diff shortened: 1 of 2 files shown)


sampsyo:
Nice! The motivation here definitely makes sense, and your Python is 100% fine. (You should see the style on some of the PRs we get around here…)

Here are some suggestions:

* About the name: We generally use verb- or adjective-like names for flags like this. What about something like `prefer_specific` or even just `specific`?
* Please add a description to the documentation (i.e., `lastgenre.rst`) about the new option. A bullet point in the changelog would also be helpful.
* Travis has a couple of style suggestions for you: https://travis-ci.org/beetbox/beets/jobs/162901157 Specifically, lines should be kept to 79 characters.

beetsplug/lastgenre/__init__.py
```diff
@@ -158,6 +159,27 @@ def sources(self):
...
+    def _sort_by_depth(self, tags):
+        """Given a list of tags, sort the tags by their depths in the
+        genre tree.
+        """
+        tags_by_depth = []
+        for key, value in enumerate(tags):
+            depth = self._get_depth(value)
+            tags_by_depth.append({'tag': value, 'depth': depth})
+        tags_by_depth = sorted(tags_by_depth, key=lambda k: k['depth'], reverse=True)
+        return [i['tag'] for i in tags_by_depth]
```
sampsyo: One way to do this sort that might be slightly shorter to write would be to use tuples. For example, something like this:

```py
depth_tag_pairs = [self._get_depth(t), t for t in tags]
depth_tag_pairs.sort(reverse=True)
return [p[0] for p in depth_tag_pairs]
```

That will automatically sort by the first element in the tuple first.

beetsplug/lastgenre/__init__.py
```diff
@@ -179,12 +201,18 @@ def _resolve_genres(self, tags):
                     parents = [find_parents(tag, self.c14n_branches)[-1]]
 
                 tags_all += parents
-                if len(tags_all) >= count:
+                # Don't break here if we need to sort by specificity, which happens with
+                # the full tag list below
```
sampsyo: It can be helpful to phrase comments in a positive way. For example, consider "Stop if we have enough tags already, unless we need to find the most specific tag (instead of the most popular)."

beetsplug/lastgenre/__init__.py
```diff
@@ -179,12 +201,18 @@ def _resolve_genres(self, tags):
...
+                if not self.config['specificity'] and len(tags_all) >= count:
                     break
             tags = tags_all
 
         tags = deduplicate(tags)
 
+        # sort the tags by specificity
```
sampsyo: I know the comment right below violates this :angry:, but we usually like to use normal English style for comments: i.e., start with a capital letter and end with a period.

heylookltsme:
Thanks for the feedback @sampsyo! Updated!

beetsplug/lastgenre/__init__.py
```diff
@@ -179,12 +198,19 @@ def _resolve_genres(self, tags):
                     parents = [find_parents(tag, self.c14n_branches)[-1]]
 
                 tags_all += parents
-                if len(tags_all) >= count:
+                # Stop if we have enough tags already, unless we need to find
+                # the most specific tag (instead of the most popular).
+                if (not self.config['prefer_specific'] and
+                    len(tags_all) >= count):
```
sampsyo: I think Travis's flake8 checker is unhappy about the alignment here: https://travis-ci.org/beetbox/beets/jobs/162943469

I'm honestly a little bewildered by this message, and a little angry at the tool for being so confusing, but I think PEP-8 might want you to indent the line one more set of 4 spaces, according to this issue: https://github.com/PyCQA/pycodestyle/issues/386

This error is weird, though—in the future, I might look into disabling it.

sampsyo:
The docs look great! :sparkles: Thanks!

beetsplug/lastgenre/__init__.py
heylookltsme: ¯\_(ツ)_/¯

heylookltsme: Hmmm travis still seems unhappy :(

beetsplug/lastgenre/__init__.py
```diff
@@ -179,12 +198,19 @@ def _resolve_genres(self, tags):
                     parents = [find_parents(tag, self.c14n_branches)[-1]]
 
                 tags_all += parents
-                if len(tags_all) >= count:
+                # Stop if we have enough tags already, unless we need to find
+                # the most specific tag (instead of the most popular).
+                if (not self.config['prefer_specific']
+                        and len(tags_all) >= count):
```
Kraymer: Move the `and` on the line above and travis should be satisfied.

heylookltsme: Ah, thanks @Kraymer!

sampsyo: Yes; thanks, @Kraymer! I can't really explain why flake8 would prefer that, but OK…

sampsyo:
Woohoo! This looks great. Thank you again for the contribution.

PR #2212, opened 2016-09-26
Add mp3gain/accgain to appveyor.yml

jrobeson:


jrobeson:
they do get installed correctly, but they don't seem to work.

jtpavlock:
Is this still relevant? I'm trying to understand the context for the change.

sampsyo:
I think we can leave this aside for now, especially since we're hoping to move to Actions.

Issue #2166, opened 2016-08-10
Template doesn't format at all when contains a comma 


miblodelcarpio:
### Problem

Including a comma in a path yields incorrect path.

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv ls -f 'Music/Pajo, David/%the{$albumartist_sort}/$releasedate. $album%aunique{}/$disc_and_track. $title' Pajo
```

Led to this problem:

```
Music/Pajo, David/%the{$albumartist_sort}/$releasedate. $album%aunique{}/$disc_and_track. $title
```

Simply removing the comma in `Pajo, David` yields the expected result.

### Setup

* OS: Arch Linux
* Python version: 2.7.12-1
* beets version: 1.3.18
* Turning off plugins made problem go away (yes/no): No

My configuration (output of `beet config`) is:

```yaml
paths:
    default: Music/%the{$albumartist_sort}/$releasedate. $album%aunique{}/$disc_and_track. $title
    comp: Music/%the{$albumartist_sort}/$releasedate. $album%aunique{}/$disc_and_track. $artist - $title
    albumartist:Aerial M: Music/Pajo, David/%the{$albumartist_sort}/$releasedate. $album%aunique{}/$disc_and_track. $title
    albumartist:Papa M: Music/Pajo, David/%the{$albumartist_sort}/$releasedate. $album%aunique{}/$disc_and_track. $title
    albumartist:Pajo: Music/Pajo, David/%the{$albumartist_sort}/$releasedate. $album%aunique{}/$disc_and_track. $title
    albumtype:soundtrack: Soundtrack/Game/%the{$albumartist_sort}/$releasedate. $album%aunique{}/$disc_and_track. $title
    albumtype:soundtrack comp: Soundtrack/Game/%the{$albumartist_sort}/$releasedate. $album%aunique{}/$disc_and_track. $artist - $title
    albumstatus:bootleg: Music/%the{$albumartist_sort}/Bootlegs/$album%aunique{}/$disc_and_track. $title
item_fields:
    disc_and_track: u'%02i' % (track) if disctotal == 1 else u'%i-%02i' % (disc, track)
    releasedate: u'%02i-%02i-%02i' % (year, month, day) if day else u'%02i-%02i' % (year, month) if month else u'%02i' % (year)
asciify_paths: yes
library: ~/.config/beets/library.db

replace:
    '[\\/]': _
    ^\.: _
    \s+$: ''
    ^\s+: ''

[...]

PR #2213, opened 2016-10-02
Revise comma handling on templates

diego-plan9:
A first take for treating commas (`ARG_SEP`) as a regular character during the template parsing, except when it is used inside a function's argument (ie. `%foo{bar,baz}`), as discussed on #2166.

I have added a couple of tests checking for the usage of commas *before* other elements, as it seems that the problem was that the parser stopped processing the rest of the string whenever an unescaped comma was found. The test suite already includes other tests that deal with separators, (`a $, b`, `a , b`, `%foo{bar,baz}`, `%foo{bar$,baz}`, ...) but I'd be happy to add more if needed.

The solution basically splits the handling of the special characters (`special_chars,` `special_char_re`, and the newly promoted `escapable_chars` and `terminator_chars`) in two cases: one for regular parsing that does not treats commas in a special way, and one just for `parse_argument_list()` which does consider commas a special character (ie. in the same way it has been until these changes). I opted for subclassing, albeit I'm not sure it's entirely justified - it could easily be moved to a lower abstraction level, ie. `__init__` or `parse_expression` if preferred.

### DIFF
--- a/beets/util/functemplate.py
+++ b/beets/util/functemplate.py
@@ -484,7 +484,7 @@ class Parser(object):
         expressions = []
 
         while self.pos < len(self.string):
-            subparser = Parser(self.string[self.pos:])
+            subparser = ArgumentsParser(self.string[self.pos:])
             subparser.parse_expression()
 
             # Extract and advance past the parsed expression.
@@ -511,6 +511,18 @@ class Parser(object):
         ident = re.match(r'\w*', remainder).group(0)
         self.pos += len(ident)
         return ident
+
+
+class ArgumentsParser(Parser):
+    """``Parser`` that considers ``ARG_SEP`` to be a special character.
+    """
+    # Common parsing resources.
+    special_chars = (SYMBOL_DELIM, FUNC_DELIM, GROUP_OPEN, GROUP_CLOSE,
+                     ARG_SEP, ESCAPE_CHAR)
+    special_char_re = re.compile(r'[%s]|$' %
+                                 u''.join(re.escape(c) for c in special_chars))
+    escapable_chars = (SYMBOL_DELIM, FUNC_DELIM, GROUP_CLOSE, ARG_SEP)
+    terminator_chars = (GROUP_CLOSE, ARG_SEP)
 
 
 def _parse(template):
--- a/beets/util/functemplate.py
+++ b/beets/util/functemplate.py
@@ -318,9 +318,11 @@ class Parser(object):
 
     # Common parsing resources.
     special_chars = (SYMBOL_DELIM, FUNC_DELIM, GROUP_OPEN, GROUP_CLOSE,
-                     ARG_SEP, ESCAPE_CHAR)
+                     ESCAPE_CHAR)
     special_char_re = re.compile(r'[%s]|$' %
                                  u''.join(re.escape(c) for c in special_chars))
+    escapable_chars = (SYMBOL_DELIM, FUNC_DELIM, GROUP_CLOSE, ARG_SEP)
+    terminator_chars = (GROUP_CLOSE)
 
     def parse_expression(self):
         """Parse a template expression starting at ``pos``. Resulting
@@ -348,14 +350,13 @@ class Parser(object):
                 # The last character can never begin a structure, so we
                 # just interpret it as a literal character (unless it
                 # terminates the expression, as with , and }).
-                if char not in (GROUP_CLOSE, ARG_SEP):
+                if char not in self.terminator_chars:
                     text_parts.append(char)
                     self.pos += 1
                 break
 
             next_char = self.string[self.pos + 1]
-            if char == ESCAPE_CHAR and next_char in \
-                    (SYMBOL_DELIM, FUNC_DELIM, GROUP_CLOSE, ARG_SEP):
+            if char == ESCAPE_CHAR and next_char in self.escapable_chars:
                 # An escaped special character ($$, $}, etc.). Note that
                 # ${ is not an escape sequence: this is ambiguous with
                 # the start of a symbol and it's not necessary (just
@@ -375,7 +376,7 @@ class Parser(object):
             elif char == FUNC_DELIM:
                 # Parse a function call.
                 self.parse_call()
-            elif char in (GROUP_CLOSE, ARG_SEP):
+            elif char in self.terminator_chars:
                 # Template terminated.
                 break
             elif char == GROUP_OPEN:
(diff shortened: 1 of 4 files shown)


beets/util/functemplate.py
```diff
@@ -318,9 +318,11 @@ def __init__(self, string):
 
     # Common parsing resources.
     special_chars = (SYMBOL_DELIM, FUNC_DELIM, GROUP_OPEN, GROUP_CLOSE,
-                     ARG_SEP, ESCAPE_CHAR)
+                     ESCAPE_CHAR)
     special_char_re = re.compile(r'[%s]|$' %
                                  u''.join(re.escape(c) for c in special_chars))
+    escapable_chars = (SYMBOL_DELIM, FUNC_DELIM, GROUP_CLOSE, ARG_SEP)
```
diego-plan9: I left `ARG_SEP` deliberately in the list of escapable chars in order to allow both escaped and unescaped commands outside a function (`-f 'foo, $bar'` would work exactly the same as `-f 'some$, $thing'`).

I'm undecided about it, though: on one hand, it keeps the syntax backwards compatible-ish, but on the other hand introduces some ambiguity.

sampsyo: Yes, I agree here—I think it's more predictable to make $, work the same way—as an escape—both inside and outside of function arguments.

beets/util/functemplate.py
```diff
@@ -318,9 +318,11 @@ def __init__(self, string):
...
+    terminator_chars = (GROUP_CLOSE)
```
sampsyo: I think you want `(GROUP_CLOSE,)` here to make it a tuple. Otherwise this is just a one-character string, which happens to work in this case!

beets/util/functemplate.py
```diff
@@ -512,6 +513,18 @@ def _parse_ident(self):
         return ident
 
 
+class ArgumentsParser(Parser):
+    """``Parser`` that considers ``ARG_SEP`` to be a special character.
```
sampsyo: Perhaps a more descriptive docstring would say "a parser used inside function arguments," i.e., state the purpose first before getting into the implementation details.

beets/util/functemplate.py
diego-plan9: Yep, that was the idea - thanks for catching it!

beets/util/functemplate.py
```diff
@@ -512,6 +513,18 @@ def _parse_ident(self):
...
+    """
+    # Common parsing resources.
+    special_chars = (SYMBOL_DELIM, FUNC_DELIM, GROUP_OPEN, GROUP_CLOSE,
+                     ARG_SEP, ESCAPE_CHAR)
+    special_char_re = re.compile(r'[%s]|$' %
+                                 u''.join(re.escape(c) for c in special_chars))
+    escapable_chars = (SYMBOL_DELIM, FUNC_DELIM, GROUP_CLOSE, ARG_SEP)
+    terminator_chars = (GROUP_CLOSE, ARG_SEP)
```
sampsyo: I like this subclassing approach, but I'm a little sad that it required us to copy n' paste the definitions from the base class. This seems like it could get us into trouble if we eventually change one set of lists and forget to update the other one.

It's a tough call, but maybe just an `in_argument` flag on `Parser` would be simpler?

beets/util/functemplate.py
```diff
@@ -322,24 +328,34 @@ def __init__(self, string):
...
         """
+        # Append comma (ARG_SEP) to the list of special characters only when
+        # parsing function arguments.
+        extra_special_chars = ()
+        special_char_re = self.special_char_re
+        if self.in_argument:
+            extra_special_chars = (ARG_SEP,)
+            special_char_re = re.compile(
+                r'[%s]|$' % u''.join(re.escape(c) for c in
+                                     self.special_chars + extra_special_chars))
```
diego-plan9: Take 2, using an `in_argument` flag! This block introduces an amount of extra-cruft on `parse_expression`, but I wasn't too comfortable directly overwriting the class variables with instance variables on the constructor, ie:

```python
def __init__(self, string, in_argument=False):
    ...
    if in_argument:
        self.special_chars = ...
        self.special_char_re = ...
```

It might be quite a minor concern, as the Parsers that are used for list arguments only get a single call to parse_expression (and none to the other methods) in practice, so I'd be up for another refactoring if you think the trade-off make sense.

sampsyo: Yes; this looks great! The small bit of extra cruft is a little bit annoying, but I agree this is the right direction. It's preferable, as you note, to overriding those class variables.

diego-plan9:
Thanks for the quick review, it's my pleasure to squash bugs! I'm wondering if I could trouble you to revise the Path Formats\Syntax Details documentation before merging, as it will probably result in a clearer explanation of the situation?

sampsyo:
I'd be happy to, but I don't think I can push to your fork to update this PR? I can also just add the docs post-merge, but I'd say something like this:

> Commas are used as argument separators in function calls. Inside of a function's argument, use `$,` to get a literal `,` character. Outside of any function argument, escaping is not necessary: `,` by itself will produce `,` in the output.

diego-plan9:
> I'd be happy to, but I don't think I can push to your fork to update this PR?

Hmmm, I think it should be possible since the recent-ish github changes (I have an "_Allow edit from mantainers_" box checked right after the "_Lock conversation_" and was prompted when creating the pull request), but honestly haven't tried out that feature.

Nevertheless, I adjusted the documentation based on your suggestion, plus included a fix for a formatting issue one the `%first{}` section that I noticed while revising generated html file that seems to be caused by a [limitation on reST inline markup](http://www.sphinx-doc.org/en/stable/rest.html#inline-markup):

> content may not start or end with whitespace: * text* is wrong,

sampsyo:
Aha; thanks for fixing that!

And for committing the docs. I honestly can't quite figure out what's going on with the "allow edit from maintainers" checkbox—AFAICT, I still don't have ordinary push permissions to your repository, but maybe there's some kind of invisible exception that requires some incantation I don't know?

Anyway, this looks great! Please merge whenever you like.

diego-plan9:
Thanks! As for the permissions issue, it seems that they are "per-branch"  (in this case, just for the `template-comma-behaviour` branch), so there might be some magic involved - or alternatively, I might have missed something on my end, which is always an option! I'll look into it for future pull requests.

Merging!

sampsyo:
Yay! Thanks again. :sparkles:

I'll look into that new feature more closely in future PRs too.

PR #1691, opened 2015-11-03
Fix discogs disc field is set to zero

kooimens:
Fix for #587. Disc field is only zero when there is only one medium, so this will do the trick. I wasn't able to reproduce the real problem within the code. This is just a small workaround.

PR #1351, opened 2015-03-10
Fix discogs reauth bug by avoiding empty queries

scztt:
This fixes #1345 - or, at least, the secondary issue where the discogs plugin repeatedly tried to authenticate. This is sort of a bug in the Discogs API, since 401 is probably not an appropriate error code for a malformed query.... But this fix sidesteps getting that error for empty queries.

Issue #1345, opened 2015-03-05
Discogs crashes beets 1.3.11


sammcj:
```
beets version 1.3.11
plugins: fromfilename, duplicates, lastgenre, mbsync, lastimport, fetchart, ftintitle, discogs
Python 2.7.9
OSX 10.10
```

```
Traceback (most recent call last):
  File "/usr/local/bin/beet", line 9, in <module>
    load_entry_point('beets==1.3.11', 'console_scripts', 'beet')()
  File "/usr/local/lib/python2.7/site-packages/beets/ui/__init__.py", line 953, in main
    _raw_main(args)
  File "/usr/local/lib/python2.7/site-packages/beets/ui/__init__.py", line 943, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "/usr/local/lib/python2.7/site-packages/beets/ui/commands.py", line 876, in import_func
    import_files(lib, paths, query)
  File "/usr/local/lib/python2.7/site-packages/beets/ui/commands.py", line 853, in import_files
    session.run()
  File "/usr/local/lib/python2.7/site-packages/beets/importer.py", line 316, in run
    pl.run_parallel(QUEUE_SIZE)
  File "/usr/local/lib/python2.7/site-packages/beets/util/pipeline.py", line 301, in run
    out = self.coro.send(msg)
  File "/usr/local/lib/python2.7/site-packages/beets/util/pipeline.py", line 183, in coro
    func(*(args + (task,)))
  File "/usr/local/lib/python2.7/site-packages/beets/importer.py", line 1236, in lookup_candidates
    task.lookup_candidates()
  File "/usr/local/lib/python2.7/site-packages/beets/importer.py", line 567, in lookup_candidates
    autotag.tag_album(self.items)
  File "/usr/local/lib/python2.7/site-packages/beets/autotag/match.py", line 433, in tag_album
    search_album, va_likely)
  File "/usr/local/lib/python2.7/site-packages/beets/autotag/hooks.py", line 565, in album_candidates
    out.extend(plugins.candidates(items, artist, album, va_likely))
  File "/usr/local/lib/python2.7/site-packages/beets/plugins.py", line 344, in candidates
    out.extend(plugin.candidates(items, artist, album, va_likely))
  File "/usr/local/lib/python2.7/site-packages/beetsplug/discogs.py", line 141, in candidates
    self.reset_auth()
[...]

Issue #587, opened 2014-03-08
discogs: disc field is set to zero


kooimens:
Releases coming from MB with only a single disc are tagged in a correct way (to my view) like: 1/1. However, when using the discogs plugin for releases from discogs the disc field of a single disc album is set to 0/1.

What is exactly going wrong here? Is it just a mistake within the plugin or is it the discogs database serving the wrong 0/1 disc tag? And, is there a workaround to avoid this and have it tagged in the correct way like MB does as 1/1?

PR #2222, opened 2016-10-07
Several discogs fixes

diego-plan9:
A round of fixes for the *discogs* plugin, with different degrees of elegance! This is hopefully a work in progress pull request, as there seems to be a number of small issues related to the plugin that can probably be handled during this round.

While working on the fixes, my main problem was that there seems to be no existing unit tests, which combined with my lack of familiarity with the discogs API makes things rather inconvenient - and as a result, the current revision might break some existing cases. In order to alleviate this, I include an initial unit test file for discogs, heavily inspired on `test_mb.py` - and also heavily lacking in features, as only tests meant to test the fixes (and still not as thoroughly as they should) are included. I'd really welcome if someone more used to discogs, either as a user or a collaborator, would double-check or provide some help!

Some general notes:
* discogs_client includes [some API dumps](https://github.com/discogs/discogs_client/tree/master/discogs_client/tests/res) that might be useful - I was tempted to make us of their approach at some point.
* in general, I tried to preserve the existing structure of the discogs plugin, and treat the new additions as refinements of the general cases rather than going for a bigger refactor. There seems to be a number of fields on discogs API that could be useful for several purposes (for example, there is a [format.quantity](https://www.discogs.com/help/submission-guidelines-release-format.html#Format) that probably would be useful for determining beets' medium fields; or an additional `type_` field on tracks might help identifying index tracks robustly) - but I'm not that familiar to try that approach yet.
* the helper methods on the unit test (`_make_release`) are very incomplete - the idea was to try to mimic the minimal amount of fields required by the discogs plugin and the fixes.


beetsplug/discogs.py
```diff
@@ -315,30 +339,95 @@ def get_tracks(self, tracklist):
 
         return tracks
 
+    def coalesce_tracks(self, raw_tracklist):
```
diego-plan9: This is still not working on one example on the initial issue:

https://www.discogs.com/Various-Defqon1-Weekend-Festival/release/7168134

It seems that particular release has the `sub_tracks` defined *inside* the track:
```json
"tracklist" = [
        ...
	{
			"duration": "3:06",
			"position": "",
			"type_": "index",
			"sub_tracks": [
				{
					"extraartists": [...],
					"title": "Eat This",
					"type_": "track",
					"artists": [...],
					"duration": "",
					"position": "2-20A"
				},
				{
					"extraartists": [...],
					"title": "God! What The Hell!",
					"type_": "track",
					"artists": [...],
					"duration": "",
					"position": "2-20B"
				}
			],
			"title": "Eat This / God! What The Hell!"
		},
```

as opposed to the other samples mentioned, where the subtracks are just a regular `track` inside `tracklist`. I'm not sure if the first sample is the exception, or if both are pretty common on discogs?

beetsplug/discogs.py
```diff
@@ -315,30 +339,95 @@ def get_tracks(self, tracklist):
...
     def get_track_index(self, position):
-        """Returns the medium and medium index for a discogs track position.
+        """Returns the medium, medium index and subtrack index for a discogs
+        track position.
         """
+        # TODO: revise comment
         # medium_index is a number at the end of position. medium is everything
         # else. E.g. (A)(1), (Side A, Track )(1), (A)(), ()(1), etc.
-        match = re.match(r'^(.*?)(\d*)$', position.upper())
+        match = re.match(r'^(.*?)(\d*?)(\.[\w]+|[A-Z]+)?$', position.upper())
```
diego-plan9: This was derived from the information at 12.2.9 on the [discogs guidelines](https://www.discogs.com/help/submission-guidelines-release-trk.html#Position) - however, 12.2.2 worries me a bit:

> 12.2.2. You can enter the positions from release, or using the Discogs standard positions (see below). The positions from the release are preferred.

Some extra checks for non-standard positions would probably be needed!

beetsplug/discogs.py
```diff
@@ -234,6 +235,10 @@ def get_album_info(self, result):
             catalogno = None
         country = result.data.get('country')
         media = result.data['formats'][0]['name']
+        # Explicitely set the `media` for the tracks, since it is expected by
```
sampsyo: Spelling: "Explicitly"

beetsplug/discogs.py
sampsyo: Oh wow, that's pretty crazy! Perhaps we should flatten these subtrack lists in a subsequent round of changes.

diego-plan9:
Thanks - I noticed the CI failing right after pushing but felt it was a bit premature to modify the tox configuration: anyway, I just included the modifications (and also the dependency at  `setup.py`, which I believe would be needed as well?) in e7f2a11.

I'm wondering if it would be reasonable to expect some extra input or testing from someone more familiar with discogs (either as a regular user or collaborator) in the short future? I don't mind having this pull request turn into a long-lived one, and I'd be equally happy merging the current changes after doing a bit more testing myself. My main concern is being reasonably sure that these changes do not break existing functionality of the discogs plugin, which is hard to ascertain at this point!

Regarding the "_sub_tracks-defined-inside-track_" scenario: I'm tempted to try to include that case on this pull request, as it seems easy to identify and would make the fix a bit more complete - I will look into it!

tox.ini
```diff
@@ -23,6 +23,7 @@ deps =
     jellyfish
     python-mpd2
     coverage
+    discogs_client
```
sampsyo: This should probably be a hyphen, not an underscore, for consistency—but it looks like pip is happy with either one? :thinking:

diego-plan9:
Thanks for the quick feedback, as usual!

I have added a few more tests, and in particular a `try...catch` wrapping the function that attempts to merge the subtracks. It's a bit ugly, and I couldn't yet find a case where an Exception was actually raised, but I included in order to have a little more peace of mind (at least until further tests are included) and revert to parsing the original tracklist if something goes wrong.

The case where `sub_tracks` are defined inside a `track` is also included in the latest revision, as it seemed easy to detect and integrate with the existing changes. I could not find too many examples, other than:
https://www.discogs.com/Various-Defqon1-Weekend-Festival/release/483738
and had no luck trying to find the official discogs documentation in regards to the `sub_tracks` field - but hopefully they will appear in time.

Also, while going through the open issue list, I noticed #415 and couldn't resist the urge of including a fix on this round as well. It does not tackle the underlying issue, but at least should achieve the goal of including `medium_total` until an eventual bigger refactor.

And also, I agree with your plan: I'll ping the mailing list, and make some more tests during the next days before merging!

diego-plan9:
While looking for other discogs issues, I also noticed a couple of pull request/issues that might be already fixed or rather easy to include:
* #1351: it seems to be no longer required after the closing of  #1345?
* #587: a solution is proposed on #1691 that looks conceptually spot on (`medium_count` is increased whenever there is a change of medium, but *not* for the first medium), but seems to have been stalled for a while?

sampsyo:
Looking good. A catch-all `except` is OK temporarily, but as you note, it would be nice to remove it when we're more confident in the code.

About the other issues:

- In general, I like an approach with smaller PRs that attack specific problems. That way, you can see which chunks of work were conceptually related (and people get to use enhancements earlier on). But since the other changes depend on this unmerged PR, it's fine to fold them in if you're motivated!
- Yeah, #1691 could be fine! I was just a little nervous about it because we didn't have any tests, and I couldn't immediately see why the fix was correct. But now that we have tests, that could be less of a concern! I'd be happiest if we found a failing test for the current behavior and then fixed it, TDD-style.
- At first glance, I agree about #1351: it might be "expired."

diego-plan9:
After having done a few more tests, I'm following your advice and finally merging - nevertheless, I'll be keeping an eye on new discogs-related issues on the repository in the near future, as I still have some doubt about obscure cases, and please do ping me explicitly if I happen to miss any that might be related to these changes.

And I fully agree with the "*one PR per problem*" policy - I deviated in this case due to the fixes being rather small and ended up being tightly coupled to the new tests, but will make sure to be follow the guideline more closely on the future :+1: . As a matter of fact, I'll most likely start a new pull request soon working on #1691 (basically adding tests, etc).

sampsyo:
Great; thank you! Aside from the improvements themselves, I will sleep much better tonight knowing we have tests for the discogs plugin landed. :sparkles:

Issue #2229, opened 2016-10-18
beet update for selective fields


dangmai:
Right now `beet update` doesn't have an option for updating only selective fields yet, and I think it'd be a good addition.

My use case for that is the following: I've manually updated the music metadata out of band for the Grouping tag using an external media player. However that media player also updates other tags that it's not supposed to, and now I want to update beet's backend with only the new Grouping information.

If this is something desirable, I can open a PR and start working on those. I'm thinking of adding a `-f` flags to specify which fields to update, and they can be comma separated.

sampsyo:
Yes! This sounds great. The request even sounds familiar—although I can't find another ticket that mentions the idea.

An `-f` flag to the `update` command sounds reasonable. But I'd slightly prefer a version where you just provide the `-f` flag multiple times over a comma-separated list.

dangmai:
Yup multiple `-f` is definitely better, I'll implement it that way.

Is this the ticket you're thinking about? https://groups.google.com/forum/#!topic/beets-users/o0hibIx3FSI

sampsyo:
Yep, that was exactly it. Thanks for digging through the mailing list.

PR #2231, opened 2016-10-21
Selective field updates

dangmai:
PR to fix https://github.com/beetbox/beets/issues/2229

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -342,7 +342,7 @@ class Model(object):
 
     # Database interaction (CRUD methods).
 
-    def store(self):
+    def store(self, fields_to_store=None):
         """Save the object's metadata into the library database.
         """
         self._check_db()
@@ -350,7 +350,9 @@ class Model(object):
         # Build assignments for query.
         assignments = []
         subvars = []
-        for key in self._fields:
+        if fields_to_store is None:
+            fields_to_store = self._fields
+        for key in fields_to_store:
             if key != 'id' and key in self._dirty:
                 self._dirty.remove(key)
                 assignments.append(key + '=?')
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -342,17 +342,19 @@ class Model(object):
 
     # Database interaction (CRUD methods).
 
-    def store(self, fields_to_store=None):
+    def store(self, fields=None):
         """Save the object's metadata into the library database.
+        :param fields: the fields to be stored. If not specified, all fields
+        will be.
         """
+        if fields is None:
+            fields = self._fields
         self._check_db()
 
         # Build assignments for query.
         assignments = []
         subvars = []
-        if fields_to_store is None:
-            fields_to_store = self._fields
-        for key in fields_to_store:
+        for key in fields:
             if key != 'id' and key in self._dirty:
                 self._dirty.remove(key)
                 assignments.append(key + '=?')
--- a/beets/library.py
+++ b/beets/library.py
@@ -730,7 +730,7 @@ class Item(LibModel):
         self._db._memotable = {}
 
     def move(self, copy=False, link=False, basedir=None, with_album=True,
-             fields=None):
+             store=True):
         """Move the item to its designated location within the library
         directory (provided by destination()). Subdirectories are
         created as needed. If the operation succeeds, the item's path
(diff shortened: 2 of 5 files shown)


beets/dbcore/db.py
```diff
@@ -342,15 +342,17 @@ def __delattr__(self, key):
 
     # Database interaction (CRUD methods).
 
-    def store(self):
+    def store(self, fields_to_store=None):
         """Save the object's metadata into the library database.
         """
```
sampsyo: This is just cosmetic, but I would prefer to use (a) a simpler parameter name like `fields` and (b) documentation in the docstring here to explain exactly what it does.

beets/dbcore/db.py
```diff
@@ -342,15 +342,17 @@ def __delattr__(self, key):
...
         self._check_db()
 
         # Build assignments for query.
         assignments = []
         subvars = []
-        for key in self._fields:
+        if fields_to_store is None:
+            fields_to_store = self._fields
```
sampsyo: This is again just cosmetic, but I like to put these default-value `if` blocks to the very top of the function to make it clear what the default values are.

beets/library.py
```diff
@@ -323,8 +323,8 @@ def _template_funcs(self):
         funcs.update(plugins.template_funcs())
         return funcs
 
-    def store(self):
-        super(LibModel, self).store()
+    def store(self, fields_to_store=None):
+        super(LibModel, self).store(fields_to_store)
```
sampsyo: (The parameter name should be consistent with the above, if it changes.)

beets/library.py
```diff
@@ -729,7 +729,8 @@ def remove(self, delete=False, with_album=True):
 
         self._db._memotable = {}
 
-    def move(self, copy=False, link=False, basedir=None, with_album=True):
+    def move(self, copy=False, link=False, basedir=None, with_album=True,
+             fields_to_store=None):
```
sampsyo: I'm not sure `move` needs the new parameter—it's a little unintuitive for clients, and the only reason we store here is to update the paths for files.

beets/library.py
```diff
@@ -1000,7 +1001,7 @@ def move_art(self, copy=False, link=False):
             util.prune_dirs(os.path.dirname(old_art),
                             self._db.directory)
 
-    def move(self, copy=False, link=False, basedir=None):
+    def move(self, copy=False, link=False, basedir=None, fields_to_store=None):
```
sampsyo: (As above.)

beets/library.py
```diff
@@ -1108,7 +1110,7 @@ def set_art(self, path, copy=True):
 
         plugins.send('art_set', album=self)
 
-    def store(self):
+    def store(self, fields_to_store=None):
         """Update the database with the album information. The album's
         tracks are also updated.
         """
```
sampsyo: (This probably also needs the name change and docstring expansion.)

beets/ui/commands.py
```diff
@@ -1081,11 +1081,16 @@ def list_func(lib, opts, args):
 
 # update: Update library contents according to on-disk tags.
 
-def update_items(lib, query, album, move, pretend):
+def update_items(lib, query, album, move, pretend, fields):
     """For all the items matched by the query, update the library to
     reflect the item's embedded tags.
```
sampsyo: `fields` needs a mention in the docstring.

beets/ui/commands.py
```diff
@@ -1081,11 +1081,16 @@ def list_func(lib, opts, args):
...
     """
     with lib.transaction():
+        if move and fields is not None and 'path' not in fields:
+            # Special case: if an item needs to be moved, the path field has to
+            # updated; otherwise the new path will not be reflected in the
+            # database
```
sampsyo: Missing a period in this comment.

beets/ui/commands.py
```diff
@@ -1124,24 +1129,25 @@ def update_items(lib, query, album, move, pretend):
...
+                item,
+                fields=fields or library.Item._media_fields)
 
             # Save changes.
             if not pretend:
                 if changed:
                     # Move the item if it's in the library.
                     if move and lib.directory in ancestry(item.path):
-                        item.move()
+                        item.move(fields_to_store=fields)
```
sampsyo: Because we're *about* to call `item.store`, maybe the right thing to do here is just to disable the redundant store to the database inside `item.move`.

dangmai:
I've pushed a fix for the small cosmetic changes you asked for. The `fields` param to the `move` method is a little bit more tricky and I'll (hopefully) tackle it this weekend. My first thought is to specify a `store=True` argument to `move`, and if it's `False` we won't do the automatic storing in that method. Then we can manually handle the storing in the `update_items` instead. And since it has a default value of `True`, the other part of `beets` that depend on that method won't be affected.

Thoughts?

sampsyo:
Fantastic; thank you! I think the `store=True` argument to `move` makes the changes there much simpler. I'll merge this now.

Nice work! :sparkles:

PR #2233, opened 2016-10-22
Drop `reduce` imports

nathdwek:
Avoid using `reduce` when it can be elegantly avoided.

`ag '[^\.]reduce\(' -C 5`

````
  beets/autotag/match.py
  267-        log.debug(u'No album IDs found.')
  268-        return None
  269-
  270-    # If all album IDs are equal, look up the album.
  271:    if bool(reduce(lambda x, y: x if x == y else (), albumids)):
  272-        albumid = albumids[0]
  273-        log.debug(u'Searching for discovered album ID: {0}', albumid)
  274-        return hooks.album_for_mbid(albumid)
  275-    else:

  beets/dbcore/query.py
  380-    def __hash__(self):
  381-        """Since subqueries are mutable, this object should not be hashable.
  382-        However and for conveniences purposes, it can be hashed.
  383-        """
  384:        return reduce(mul, map(hash, self.subqueries), 1)
  385-
  386-
  387-class AnyFieldQuery(CollectionQuery):
  388-    """A query that matches if a given FieldQuery subclass matches in

  beetsplug/acousticbrainz.py
  60-    """
  61-
  62-    def get_value(*map_path):
  63-        try:
  64:            return reduce(operator.getitem, map_path, data)
  65-        except KeyError:
  66-            log.debug(u'Invalid Path: {}', map_path)
  67-
  68-    for item in items:
````

I don't think there is better way to do the hashing in `beets/dbcore/query.py`, and I did not take a look at `get_value` in `beetsplug/acousticbrainz.py` yet.

So far, I refactored the one in `beets/autotag/match.py`. I went in to get rid of `reduce`, but I think the rewrite is actually useful in terms or readability. Moreover, it explicitly uses lazy evaluation, while I don't think the old way did, even though I didn't benchmark. It also accommodates generators better for future improvements.

### DIFF
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -29,7 +29,6 @@ from beets import config
 from beets.util import plurality
 from beets.autotag import hooks
 from beets.util.enumeration import OrderedEnum
-from functools import reduce
 
 # Artist signals that indicate "various artists". These are used at the
 # album level to determine whether a given release is likely a VA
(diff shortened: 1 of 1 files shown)


nathdwek:
Mmmmh, there are some shenanigans going on with the `__next__` and `next` methods and the `next` function across python versions. I'm on it.

sampsyo:
This looks good; thank you! I do think the new version is more readable. And you're right that `next(e)` is the right way to do iteration compatibly.

I also wouldn't be opposed to getting rid of the other ones either. The `__hash__` method in `query.py` is particularly suspicious; the docstring suggests the whole method may not be necessary

nathdwek:
Oh, I was planning on submitting patches for the other ones in this PR. Regarding \__hash__` I read the docstring but did not push further in that direction, I will investigate it. The code itself however couldn't be more expressive or correct I think.

I was however going for a big refactor of acousticbrainz. I don't think I was actually going to get rid of the reduce usage it has (pretty elegant I thought), but the rest of the plugin is really really ugly:
* There is only a single "business" method that is very long.
* This method defines two inner functions
* There is a LOT of repetition going on to parse the JSON response.
* A lot of single lines and ternaries, not the good kind I feel.

Those two changes might end up not really belong to the drop-reduce topic, so all good then!
As usual, thanks for your being so responsive!!

sampsyo:
Oops! Sorry for jumping the gun. I'll be more circumspect in the future and let you work on the pull request for longer.

Yes, a refactoring of acousticbrainz would be great! A declarative version—just a list of fields to fetch and their paths—would make that plugin a lot cleaner. Thanks for taking a look.

PR #2243, opened 2016-10-30
Use HTTPS where possible in fetchart plugin

orf:
Lets not let the NSA intercept our album art.


orf:
http://albumart.org/ (and https) appears to be offline right now, so the tests will fail. Re-run when it's back?

Freso:
coverartarchive.org will fail with Python <2.7.9. See https://tickets.musicbrainz.org/browse/MBH-363 - I don't know if any of the other HTTPSified providers use SNI as well, but if so, they will also fail.

PR #2244, opened 2016-10-30
Use HTTPS in artresizer.py

orf:
https://images.weserv.nl/ works fine, lets use that as the default?

### DIFF
--- a/beets/util/artresizer.py
+++ b/beets/util/artresizer.py
@@ -32,7 +32,7 @@ PIL = 1
 IMAGEMAGICK = 2
 WEBPROXY = 3
 
-PROXY_URL = 'http://images.weserv.nl/'
+PROXY_URL = 'https://images.weserv.nl/'
 
 log = logging.getLogger('beets')
 


Freso:
If the server is using SNI it won't work on Python <2.7.9. There doesn't seem to be a good (= simple and 100% accurate) way to [find out whether it is](https://serverfault.com/questions/506177/how-can-i-detect-if-a-server-is-using-sni-for-https), except for actually trying to connect to using a client that doesn't understand SNI (e.g., Python <2.7.9).

jackwilsdon:
It looks like @Freso is correct:

```Text
$ pyenv local 2.7
$ python --version
Python 2.7
```

```Python
import requests
requests.get('https://images.weserv.nl/?url=avatars0.githubusercontent.com/u/15920313&shape=circle&w=200')
```

```Python
/Users/jack/.pyenv/versions/2.7/lib/python2.7/site-packages/requests/packages/urllib3/util/ssl_.py:318: SNIMissingWarning: An HTTPS request has been made, but the SNI (Subject Name Indication) extension to TLS is not available on this platform. This may cause the server to present an incorrect TLS certificate, which can cause validation failures. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#snimissingwarning.
  SNIMissingWarning
/Users/jack/.pyenv/versions/2.7/lib/python2.7/site-packages/requests/packages/urllib3/util/ssl_.py:122: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
  File "/Users/jack/.pyenv/versions/2.7/lib/python2.7/site-packages/requests/api.py", line 70, in get
    return request('get', url, params=params, **kwargs)
  File "/Users/jack/.pyenv/versions/2.7/lib/python2.7/site-packages/requests/api.py", line 56, in request
    return session.request(method=method, url=url, **kwargs)
  File "/Users/jack/.pyenv/versions/2.7/lib/python2.7/site-packages/requests/sessions.py", line 475, in request
    resp = self.send(prep, **send_kwargs)
  File "/Users/jack/.pyenv/versions/2.7/lib/python2.7/site-packages/requests/sessions.py", line 596, in send
    r = adapter.send(request, **kwargs)
  File "/Users/jack/.pyenv/versions/2.7/lib/python2.7/site-packages/requests/adapters.py", line 497, in send
    raise SSLError(e, request=request)
requests.exceptions.SSLError: [Errno 1] _ssl.c:507: error:14077438:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert internal error
```

The request works fine if I change `https` back to `http`. I'm led to believe we're heading towards dropping 2.7 support eventually, so I'm not sure if this should be merged yet or not.

sampsyo:
Thank you for confirming, @jackwilsdon! As discussed briefly over in #2248, we might need a version check for now.

Issue #2246, opened 2016-10-30
lyrics: Remove Genius backend


orf:
The function [get_lyrics](https://github.com/beetbox/beets/blob/master/beetsplug/lyrics.py#L291) looks up the lyrics from a server called `genius-api.com`

As far as I can tell this address is now defunct, I can't resolve it anywhere. Should it be removed?

sampsyo:
Yes; looks like that service has gone away forever. We should indeed remove it, which will also involve removing the backend from the configuration structure and from the docs.

PR #2248, opened 2016-10-30
Use the HTTPS musicbrainz endpoint

orf:


Freso:
Note that Python <2.7.9 will not work with this patch (Ubuntu Trusty's Python is 2.7.6). This might also be true for some of the other HTTPS PRs having been submitted.

See https://tickets.musicbrainz.org/browse/MBH-363

orf:
Ahh, that may require more work. Well, Python 2.7.9 was released 2 years ago, and it's only a point release. We could mandate only 2.7.9 and above?

The way I see it is you either explicitly need to support Py2 below 2.7.9, in which case you should add it to the tox runs, or mandate that you need >2.7.9, as there are already some HTTPS URLs within the code and at any moment the website owners could turn on SNI (so they should be tested).

sampsyo:
Hi! Thanks for opening all these HTTPS-related pull requests.

In this case, there are tests for the code that also need to be updated. As a policy, our tests do not make outside network connections—we mock all requests. So `test_art.py` mocks the requests to `albumart.org` and others; those mocks will need to be updated for the HTTPS URLs. Can you please also check the tests for your other PRs?

Relatedly, it would be nice to have evidence here in the PR that all of the endpoints we need actually work over HTTPS. Could you please include output demonstrating that each bit of functionality still works with each change?

Finally, there's the question of Python < 2.7.9. I think the simplest thing to do would be to use a version test where necessary—use HTTPS on recent Pythons and not on older ones. Or we could ask our users if they're on any distributions that still use Python < 2.7.9—I'm not sure where the Debian cycles currently stand on that, for example.

sampsyo:
Oops! The comment about `fetchart` was meant for #2243, of course.

orf:
Thanks for replying! Are you sure about the mocking? I had one test fail in the `fetchart` MR that seemed to be because of a network issue: https://travis-ci.org/beetbox/beets/jobs/171748821#L890

About the HTTPS tests, you're right that you need to verify that the endpoints actually work over HTTPS. I can include output to show that it works, but some tests relating to this would be good. What do you think about some endpoint tests that goes through and just ensures that each endpoint that we use resolves and is accessible over HTTPS (or HTTP if needed?). This would catch problems like #2246, but I'm not sure how to write those tests in a nice way as the URL's are often embedded in the code. 

I could do some refactoring and move all hosts to module-level definitions and collect all the endpoints by importing each of the plugins, then test them that way?

I could add a version test to use HTTPS on recent versions of Python (this would be easy with the refactoring idea above), but it would be somewhat tricky (some HTTPS endpoints work on < 2.7.9, some don't, and some don't even support HTTPS at all). Word on the street is you're moving to Py3 only in a couple of releases, perhaps just leave this as it is until then?

sampsyo:
Right; I think that's the problem in those tests: we set up the mock on an `http://` URL, so the change is letting the requests leak through to the outside world.

And sure—perhaps a separate battery of tests for contacting the outside world would be in order? I have enough experience with spurious failures due to network conditions that I would love to avoid running that kind of test on every push, but having them around seems helpful.

If we do add that kind of test, maybe the right thing would be to re-use the existing tests but disable the mocking?

PR #2249, opened 2016-10-30
[WIP]Use generators and decorators to match items:

nathdwek:
The easy parts.

So far it is relatively straight forward and it is easy to see that the code is improved as a consequence, but my end goal would be to propagate the changes upwards to `tag_album` in `beets/match.py` and refactor/simplify along the way

### DIFF
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -491,6 +491,12 @@ def sanitize_choices(choices, choices_all):
 
 
 def notify_info_yielded(event):
+    """Makes a generator send the event 'event' every time it yields.
+    This decorator is supposed to decorate a generator, but any function
+    returning an iterable should work.
+    Each yielded value is passed to plugins using the 'info' parameter of
+    'send'.
+    """
     def decorator(generator):
         def decorated(*args, **kwargs):
             for v in generator(*args, **kwargs):
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -490,7 +490,7 @@ def sanitize_choices(choices, choices_all):
     return res
 
 
-def notify_info_received(msg):
+def notify_info_yielded(msg):
     def decorator(generator):
         def decorated(*args, **kwargs):
             for v in generator(*args, **kwargs):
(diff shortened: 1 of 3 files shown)


beets/plugins.py
```diff
@@ -488,3 +488,13 @@ def sanitize_choices(choices, choices_all):
             if not (s in seen or seen.add(s)):
                 res.extend(list(others) if s == '*' else [s])
     return res
+
+
+def notify_info_received(msg):
```
sampsyo: The new decorator could probably use a docstring. Also, the functionality here doesn't seem specific to the `*_info_received` events—so perhaps the name should be more general too?

sampsyo:
Nifty! This does help a lot. I'm in favor. 👏

beets/plugins.py
nathdwek: It took me a very long time to choose a name for this decorator.
The core problem is that passing the yielded value as the info parameter is hardcoded, and that's basically what defines the decorator.
So I feel one way or the other the name should reflect this aspect. I agree that the 'received' part of the name only makes sense in context. To be honest I did not look for other places in the code where this exact decorator could be used.

The other option I thought of, which I think you'd maybe like better given your remark would be `yield_notify_info`. Once you read the code the name makes sense and is very accurate, but I rejected it because:
* In general, I think names that describe what a functions "technically" does are never really good
* If you don't read the code of the function, the name makes very little sense
* I couldn't choose an order and formulation that was concise enough (another problem in using function names which stick to the code, you end up with `do_a_then_b_return_c` which is uglyyyyy)

sampsyo: Got it! Thanks for clarifying. I agree that the behavioral name is better than the "operational" name—let's keep it as is.

nathdwek: Well I totally agree that it is not ideal as is though. Since this is a WIP I'll keep that in the back of my head as a continue examining other things.

nathdwek:
A quick grep shows that `item_candidates` and `album_candidates` in `beets/autotag/hooks.py` are the only other places where an event is sent using the `info` parameter. They were next on my hit list anyway (called by `tag_album`) so I'll see how that goes.

beets/plugins.py
nathdwek: How about `notify_info_yielded`? Technically not accurate, but I like it.

sampsyo: Sure! That seems nice and transparent (especially with an accompanying docstring).

PR #2251, opened 2016-10-31
acousticbrainz: refactor plugin

nathdwek:
Personally I like this rewrite very much. However, the rewrite is sizeable, there is no test coverage and I do not even use this plugin, so we should think twice about it before merging.

Summary of changes:
* drop `reduce` and `operator` although they were not outrageously misused
* separate information scheme from logic
* simplify some functions


jrobeson:
can you split it into more ocmmits then? make it much easier to review.

nathdwek:
I probably should :). I just liked this solution very much and because there was no tests to look over my shoulder I wanted to get feedback as I go on (Also it was 3AM).

nathdwek:
I split it up in two commits, that's really the best I could do.
If everything goes well, the linter should not complain anymore.
I improved some things, and managed to test it on my setup: it seems to work.
Docstrings for the interesting bits are coming.

beetsplug/acousticbrainz.py
```diff
@@ -42,127 +42,127 @@ def commands(self):
...
+                return reduce(operator.getitem, map_path, data)
+            except KeyError:
+                log.debug(u'Invalid Path: {}', map_path)
+
+        for item in (item for item in items if item.mb_trackid):
+            self._log.info(u'getting data for: {}', item)
+            data = self._get_data(item.mb_trackid)
+            if data:
+                # Get each field and assign it on the item.
+                item.danceable = get_value(
```
nathdwek: I'm sorry I tried my best to convince git this whole patch did not change as much as git thinks it did, but to no avail.

beetsplug/acousticbrainz.py
```diff
@@ -42,127 +118,102 @@ def commands(self):
...
-def generate_url(mbid, level):
+                self._log.debug(u'Invalid Response: {}', res.text)
+                return {}
+
+        return data
+
+    def _fetch_info(self, items, write):
+        """Get data from AcousticBrainz for the items.
+        """
+        for item in (item for item in items if item.mb_trackid):
```
sampsyo: For some reason, this strikes me as slightly less idiomatic than:

```py
for item in items:
    if not item.mb_trackid:
        continue
```

but I definitely don't have a strong argument against it.

beetsplug/acousticbrainz.py
```diff
@@ -42,127 +118,102 @@ def commands(self):
...
+                    self._log.info(u'attribute {} of {} set to {}',
+                                   attr,
+                                   item,
+                                   val)
+                    setattr(item, attr, val)
+                item.store()
+                if write:
+                    item.try_write()
+
+    def _map_dict_to_scheme(self, dictionary, scheme):
```
sampsyo: It goes without saying, but these helpers (this one and the one below) could use some nice docstrings to explain how they fit together.

nathdwek: Yes I plan for all of these to have docstrings, but this one especially is hard to get right!

beetsplug/acousticbrainz.py
nathdwek: I did not like it either when I wrote it, I think I will go back on it.

nathdwek:
The bulk of the doc is in! Let me know what you think.

I'm continuing to do some rudimentary tests on my machine, but if anyone here uses this plugin extensively or knows its potential quirks, it would be nice to have some cross checks or results.

nathdwek:
I'm writing unit tests and there are remaining obvious blunders in my code. Hopefully I'll fix those and then get out an integration test as well.

nathdwek:
I added a test with realistic data for the business logic, and the validation data was generated using the plugin on master. So far so good. I think I just need to get an integration test out now.

nathdwek:
Two questions:
* How would you go about implementing an integration test? Are there dummy albums on acousticbrainz for this purpose (IIRC there are on musicbrainz, and we use them)? Is is absolutely necessary?
* I just realized `DefaultList` doesn't support slicing. I'd say YAGNI on this one. I will definitely add this to the doc, but maybe also we should throw an adequate `NotImplementedError`?

nathdwek:
I took a step towards simplifying the approach a bit by dropping defaultlist. I think I am going to merge this very soon.

nathdwek:
Here it comes.

PR #1545, opened 2015-07-21
Embedart: fails on some jpegs because imghdr doesn't recognize the mimetype

nathdwek:
This errors pops up
```
embedart: Embedding album art into Queen - Live at Wembley ’86

embedart: not embedding image of unsupported type: image/x-None
```
with [this cover](http://i.imgur.com/b8TPrsC.jpg) (found by beets-fetchart).

This is due to [this issue](http://bugs.python.org/issue16512). Fortunately, `imghdr` provides the list `imghdr.tests` to add extra mimetype tests. I added a test which identifies jpeg files based on their magic bytes, which is the method used by `file`.

This is my first time contributing to beets, so I think some code style guidance is probably needed (mainly regarding the modification of `imghdr.tests`, which should only happen once). I also absolutely need some help with adding the right tests. There is some testing code in this PR, which is very naively based on the rest of `test_mediafile.py`, but the test doesn't fail without the patch, and I think it actually never runs. I should probably also find/build a smaller cover which can only be recognized based on the magic bytes.

PR #2255, opened 2016-11-07
Rewrite handling of images extensions and mimetype

### LINKED ISSUE #2254: fetchart uses .jpg, extractart uses .jpeg
The extractart command in the embedart plugin uses the string output of imghdr.what as an extension. That's "jpeg." 
https://github.com/beetbox/beets/blob/db782a2404fa8a6827c10a6536b4a960d19af135/beets/art.py#L197

Meanwhile, over in fetchart, it seems to use the first entry in IMAGE_EXTENSIONS for mime type image/jpeg, which is .jpg.
https://github.com/beetbox/beets/blob/db782a2404fa8a6827c10a6536b4a960d19af135/beetsplug/fetchart.py#L44

To reproduce:

Import something with no art, but has jpeg art in a configured fetchart source.
beet fetchart album
beet embedart album
beet extractart album
now you have cover.jpg and cover.jpeg or sometimes cover.1.jpeg or sometimes cover.1.jpg... it's kind of a big mess and seems weirdly nondeterministic at times.

This is normally harmless, but sometimes results in excessive clutter when re-importing old things.

nathdwek:
* Extend #1545 to hopefully all other operations which rely on image type detection
* Fix #2254

The `IdentityUnlessDict` part is probably a bit overkill for this fix, but I like to play around with collections these days. Tell me what you think about it.

When we'll be able to merge #2251 (Hopefully), `DefaultList` would also end up in `beets.util.collections`.

Many thanks to @tweitzel for noticing this inconsistency.


sampsyo:
Awesome! Thank you for the fast fix. This looks basically ready to merge, and I don't mind having the new utility class. A couple of small suggestions:

* Please add a short changelog entry.
* What do you think of the name `IdentityFallbackDict`? Seems ever so slightly more logical to me.

sampsyo:
Looks great! I say merge whenever you like.

nathdwek:
I added a test and rebased. Will merge as soon as this passes.

PR #2258, opened 2016-11-10
sanitize playlist names

michaelbub:
Make e. g.
- name: $albumartist/$year-$album.m3u'
work when album data includes non fs-friendly characters

### DIFF
--- a/beetsplug/smartplaylist.py
+++ b/beetsplug/smartplaylist.py
@@ -20,7 +20,8 @@ from __future__ import division, absolute_import, print_function
 
 from beets.plugins import BeetsPlugin
 from beets import ui
-from beets.util import mkdirall, normpath, sanitize_path, syspath, bytestring_path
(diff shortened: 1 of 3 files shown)


sampsyo:
This is a good idea; thanks!

Can you please look into the test failures? Perhaps some tests need to be updated to stay in sync.

Then, would you mind adding a quick changelog entry?

beetsplug/smartplaylist.py
```diff
@@ -186,7 +187,7 @@ def update_playlists(self, lib):
             # As we allow tags in the m3u names, we'll need to iterate through
             # the items and generate the correct m3u file names.
             for item in items:
-                m3u_name = item.evaluate_template(name, True)
+                m3u_name = item.evaluate_template(sanitized_name, True)
```
sampsyo: Hmm; are you sure you want to be sanitizing the template instead of the result of filling in the template?

sampsyo:
Thank you! :sparkles: I've merged this.

PR #2268, opened 2016-11-16
Removed call to  .decode('utf-8') when printing flexible attributes

ggVGc:
I think this is because of the switch to python3. Fixes printing of flexible attributes, which threw an exception previously.


sampsyo:
Thank you for catching this!

PR #2269, opened 2016-11-16
drop IdentityFallbackDict and use get(key, key) instead

nathdwek:
Used by mediafile to determine the preferred extension of an image.
That solution looked cool for 2 days, but in hindsight it was pretty bad.

Also:
```python
#dict_default_benchmark.py
from timeit import timeit


class IDFallbackDict(dict):
    def __getitem__(self, key):
        try:
            return dict.__getitem__(self, key)
        except KeyError:
            return key

d1 = IDFallbackDict(jpeg='jpg')
d2 = {'jpeg': 'jpg'}

if __name__ == '__main__':
    print(timeit('_ = d1["jpeg"]', setup='from __main__ import d1'))
    print(timeit('_ = d1["png"]', setup='from __main__ import d1'))
    print(timeit('_ = d2.get("jpeg", "jpeg")', setup='from __main__ import d2'))
    print(timeit('_ = d2.get("png", "png")', setup='from __main__ import d2'))
```
```
$ python3 dict_default_benchmark.py
0.31737019100000907
0.5284962050009199
0.08122138600083417
0.0809994880000886
```


nathdwek:
Appveyor should not have failed. Everything should be alright.

sampsyo:
Ha! You're right; this was a bit of overkill in retrospect. Thanks for walking it back.

Issue #2264, opened 2016-11-13
embedart has invalid text encoding?


gboysko:
### Problem

When I try to `embedart` into an entire album, MP3Diag complains that the "text encoding is unsupported for APIC frame in ID3V2 tag". 

```sh
$ beet embedart grind
```

### Setup

* OS: Ubuntu 16.04
* Python version: 2.7.12
* beets version: 1.3.18
* Turning off plugins made problem go away (yes/no): N/A

My configuration (output of `beet config`) is:

```yaml
directory: /mnt/980425CD0425AF66/MP3s.new
import:
    move: yes
    write: yes
original_date: yes
asciify_paths: yes
id3v23: yes
plugins: edit embedart fetchart lastgenre scrub missing info
paths:
    default: $albumartist/$album%aunique{}/$track $title
replaygain:
    backend: gstreamer
```

gboysko:
Looking at `mid3v2` it seems to be using UTF-16 encoding. Looking at the output generated by Picard, I see that the encoding for the ID3V2.3 tag is UTF-16, however, the `APIC` tag is encoded with `LATIN1`. I wonder if UTF-16 encoding makes sense for image data.

sampsyo:
Interesting! Can you find any other references on the Web that dictate specific encodings for `APIC`? Or perhaps this is a special proclivity in MP3Diag?

gboysko:
I will look to see what I can find. The only reference I have found is http://id3.org/id3v2.3.0, but there is no indication that `APIC` should be encoding in any other format than `UTF-16`. I'll reach out to the creator of MP3Diag.

gboysko:
Is this the same problem as #899?

sampsyo:
Good point; it could be! Encoding in `id3v23` mode is a somewhat tricky business.

gboysko:
Here is another investigation showing problems with the EasyTag application:

https://mail.gnome.org/archives/easytag-list/2013-March/msg00008.html

Seems to be consistent with what I'm seeing.

What do we need to take this forward? Is this code in another module? Or is this controlled by beets?

sampsyo:
Does it seem like the right solution is to solve #899 entirely? That is, if we switch to non-Unicode encodings entirely for ID3v2.3, will this problem be solved?

If so, you can help by taking a look at `mediafile.py` (yes, in this repository) and thinking about how we might switch to use different encodings based on the `id3v23` flag on the `MediaFile` class.

gboysko:
[...]

PR #2270, opened 2016-11-18
mediafile: prefer latin-1 encoding for ID3 APIC descriptions. Fixes #899

### LINKED ISSUE #899: MediaFile: use older text encodings in ID3v2.3 mode
I am trying to create an auto-tagging configuration in which my tags are saved in ID3v2.3 (as 2.4 lacks compatibility with some players I use) and I like the cover art to be embedded in each music file. However, the cover art of the output files is not recognised by OS X 10.9.4 (i.e. in Finder) or iTunes.

Here is a simple configuration with which the problem occurs:

``` yaml
directory: /mnt/data/home/Music

plugins: mbsync fetchart embedart
per_disc_numbering: yes
id3v23: yes

import:
    copy: yes
    write: yes

paths:
    default: $albumartist/$album%aunique{}/$disc-$track $title
```

When I comment the `id3v23: yes` option the covers of the output files are correctly recognised in Mac.

The cover art of the ID3v2.3 output files is recognised in Windows, so it seems a Mac-specific issue. Strangely enough, the input files I used already have ID3v2.3 tags and embedded cover art and are correctly recognised in Mac. Below you have a diff of the ID3v2.3 tags (sorted by name) between an input and an output file taken with `mid3v2`:

``` diff
1c1,2
< APIC= (image/jpeg, 32205 bytes)

---
> APIC= (image/jpeg, 111083 bytes)
> COMM=iTunNORM='eng'= 00001700 00001700 00003981 00003981 00000000 00000000 00008187 00008187 00000000 00000000
3a5,6
> TBPM=0
> TCMP=0
5a9
> TLAN=eng
12a17,19
> TXXX=Album Artist Credit=The Beatles
> TXXX=ALBUMARTISTSORT=Beatles, The
> TXXX=Artist Credit=The Beatles
14a22
> TXXX=MusicBrainz Album Comment=UK mono
17c25
< TXXX=MusicBrainz Album Status=official

---
> TXXX=MusicBrainz Album Status=Official
27a36
> USLT=[unrepresentable data]
```

lazka:
iTunes has problems with everything but latin-1
Try to use latin-1 if possible and fall back to utf-16.

### DIFF
--- a/beets/mediafile.py
+++ b/beets/mediafile.py
@@ -926,9 +926,9 @@ class MP3ImageStorageStyle(ListStorageStyle, MP3StorageStyle):
         try:
             frame.desc.encode("latin-1")
         except UnicodeEncodeError:
-            frame.encoding = 1  # utf-16
+            frame.encoding = mutagen.id3.Encoding.UTF16
         else:
-            frame.encoding = 0  # latin-1 if possible
+            frame.encoding = mutagen.id3.Encoding.LATIN1
 
         frame.type = image.type_index
         return frame
--- a/beets/mediafile.py
+++ b/beets/mediafile.py
@@ -920,7 +920,16 @@ class MP3ImageStorageStyle(ListStorageStyle, MP3StorageStyle):
         frame.data = image.data
         frame.mime = image.mime_type
         frame.desc = image.desc or u''
-        frame.encoding = 3  # UTF-8 encoding of desc
+
+        # For compatibility with OS X/iTunes prefer latin-1 if possible.
+        # See issue #899
+        try:
+            frame.desc.encode("latin-1")
+        except UnicodeEncodeError:
+            frame.encoding = 1  # utf-16
+        else:
+            frame.encoding = 0  # latin-1 if possible
+
         frame.type = image.type_index
         return frame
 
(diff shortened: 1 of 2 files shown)


sampsyo:
Cool! This seems like a reasonable policy. And as far as I can tell, this is the right thing to do on v2.4, right? There's no sense in preferring UTF-8 on v2.4 and Latin-1 on v2.3?

lazka:
> And as far as I can tell, this is the right thing to do on v2.4, right? There's no sense in preferring UTF-8 on v2.4 and Latin-1 on v2.3?

See https://github.com/beetbox/beets/issues/899#issuecomment-62437773

utf-16 and latin-1 are the only working encodings in both cases.

An alternative strategy would be `utf-16 if desc else latin-1` (empty utf-16 doesn't work) but preferring latin-1 seems like a better choice given the problems described in #2264

gboysko:
This change looks great to me. My only suggestion would be to use the symbolic constants (mutagen.id3.Encoding.UTF-16, etc) instead of the numeric values (0, 1).

gboysko:
Thanks @lazka. Looks great! I'm eager to try it out as soon as it gets merged!

sampsyo:
Yes, thank you, @lazka! I've merged the fix.

PR #1843, opened 2016-01-28
Selective apply of functions to items

pkess:
as suggested in #1838 i implemented a more generic way to let the user select a list of items to apply an action for

@sampsyo implementation for modify and move command will follow if this approach meets your requirements

PR #2280, opened 2016-11-24
Extra documentation for #1843

pkess:
Here you can find some extra documentation for use of selective modify.
Please check (especially spell and grammar) and merge if it is ok

(addition for #1843)


sampsyo:
Awesome! Thanks for the addition!

I've fiddled with the changes a little bit. I hope I didn't overstep my bounds, but I thought the docs seemed more at home in the CLI docs than the FAQ---so I moved some of the text and eliminated the FAQ entry for now. We can bring it back if it does in fact become frequently-asked.

PR #755, opened 2014-05-09
FIX: web plugin should work behind web proxy too

EmPeWe:
With this modifications, the web plugin will work not only on base root (http://example.com), but also on path URLs like used with proxies e.g. mod_proxy (http://example.com/beets/)

PR #2287, opened 2016-11-26
support reverse proxy

WyseNynja:
I'm trying to get nginx serving beets at example.com/beets instead of localhost:port.

I came across https://github.com/beetbox/beets/pull/755 and a few other issues.

This is my attempt to solve it. I am in the process of testing.

I wasn't sure the best way to do the config.  It seems like this new key should be under the "web" section, but I'm not sure exactly how the config object works since I haven't dug into the code much yet.

### DIFF
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -123,3 +123,5 @@ match:
     required: []
     track_length_grace: 10
     track_length_max: 30
+
+reverse_proxy: no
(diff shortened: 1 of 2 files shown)


WyseNynja:
Hmm. This isn't working right.

I think I'm going to look at some of the solutions at http://stackoverflow.com/questions/18967441/add-a-prefix-to-all-flask-routes/36033627#36033627

sampsyo:
Cool! Thanks for looking into this. Let us know if we can help test.

WyseNynja:
Closing in favor of https://github.com/beetbox/beets/pull/2482#issuecomment-287667642

Issue #1547, opened 2015-07-22
Extra fields from MusicBrainz (missing fields)


bearcatsandor:
Yanno, for all that beets prides itself in being the tagger for OCD music lovers, i'm a little surprised at the info it doesn't pull from musicbrainz.

It doesn't pull performer, composer, lyricist, conductor, engineer or producer tags. In short, why not just pull ALL the tags that it can from musicbrainz (see here  https://picard.musicbrainz.org/docs/mappings/ ) and apply it to the file and database for searching.  Why wouldn't i want to find all the various tracks that Jon Coltrane plays on even when the album title or group doesn't bare his name?

I'm just learning Python now, so should i take a shot at this? Should it be a plugin, or should i not bother because it's on the roadmap?

sampsyo:
I'm for it! Specifically, this won't be too onerous if we take this tactic:

* Refactor the `AlbumInfo` and `TrackInfo` classes to just be "bags of tags" rather than having specific fields that need to be filled (and explicitly added to the code every time we think of a new field).
* Refactor the autotag machinery for applying metadata to just copy data from the Info object to the database object blindly (again to avoid needing to add per-field plumbing). This would be agnostic to whether the field is fixed or flexible.
* Now the `mb.py` backend is free to iterate as much as possible and add all the metadata it likes without disturbing the rest of the codebase.

See also #506.

Schweinepriester:
+1

ekjaker:
+1

anshuman73:
@sampsyo,
How exactly would you like me to proceed ?
I believe reading in the task description that we need to change the entire way of storing the metadata received from MusicBrainz.

sampsyo:
Hey again, @anshuman73! You're right; this one is a bit more involved than last time.

Let's start this way. Try prototyping replacements for the `AlbumInfo` and `TrackInfo` classes in `beets/autotag/hooks.py`. At first, these classes should work the same way as before, but they should get rid of the prescribed list of fields. In fact, it should work to set *any* field on it, much like a Python dictionary. This should work:

```py
>>> info = TrackInfo(some_field=12)
>>> info.some_other_field = 50
```

Then, once that's working, we'll want to revisit the `apply_metadata` function in `autotag/__init__.py` to take advantage of this new structure. We should be able to simplify it down to just copy over all the fields from `AlbumInfo` to `Album`, and the same for `TrackInfo` to `Item`, with no complicated logic.

Sound good? Let me know if you have any questions.

[...]

PR #2294, opened 2016-11-28
Fixed #2200 and a small typo

### LINKED ISSUE #2200: Parse secondary album types from MusicBrainz
### Problem

I've different radio plays. If i import one (e.g. https://musicbrainz.org/release-group/19e6209b-2ddc-30b8-9273-484bd075fe7b) to my library the ALBUMTYPE tag is always "other". The Musicbrainz site says "other + spokenword". I think the ALBUMTYPE should be "spokenword" instead. "other" is pretty useless.
### Setup
- OS: Fedora 24
- Python version: 2.7.12
- beets version: 1.3.19
- Turning off plugins made problem go away (yes/no): no

``` yaml
directory: /mnt/Dateien/Musik
plugins: inline discogs chroma copyartifacts info badfiles edit scrub missing mbsync
paths:
    albumtype:soundtrack: Soundtracks/$album/$disc_and_track $title
    albumtype:audiobook: Hörbuch/$album/$disc_and_track $title
    albumtype:spokenword: Hörspiel/$albumartist/$album/$disc_and_track $title
    singleton:true: Tracks/$artist - $title
    comp:true: Compilations/$album%aunique{}/$disc_and_track $title
    default: Albums/$albumartist/$album%aunique{}/$disc_and_track $title
import:
    languages: de en
    detail: no
match:
    preferred:
        countries: ['DE', 'US', 'GB|UK']
        media: ['CD', 'Digital Media|File']
copyartifacts:
    extensions: .cue .CUE .LOG  .log .txt .TXT .jpeg .JPG .jpg .png .PNG
    print_ignored: yes
scrub:
    auto: yes
item_fields:
edit:
    itemfields: album albumartist artist asin catalognum comments composer country day disc disc_and_track disctitle disctotal genre label language lyrics media month original_day original_month original_year title track tracktotal year    albumfields: album albumartist albumstatus albumtotal  
    albumtype asin catalognum country day disctotal genre label language month original_day original_month original_year year


```

anshuman73:
Added functionality to retrieve secondary album tags from Musicbrainz and add all tags as a comma separated string (with primary tag as first value)

Also fixed a small typo.

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -278,20 +278,19 @@ def album_info(release):
 
     # Considers all release types (both primary and secondary) and
     # stores them as a comma-separated string
-    all_types = []
+    if 'type' in release['release-group']:
+        reltype = release['release-group']['type']
+        if reltype:
+            info.albumtype = reltype.lower()
     if 'primary-type' in release['release-group']:
         rel_primarytype = release['release-group']['primary-type']
         if rel_primarytype:
-            all_types.append(rel_primarytype.lower())
+            log.debug('Primary Type (new data): ' + rel_primarytype.lower())
     if 'secondary-type-list' in release['release-group']:
-        all_types.extend([secondarytype.lower() for secondarytype in\
-            release['release-group']['secondary-type-list']])
-    if 'type' in release['release-group']:
-        rel_type = release['release-group']['type']
-        if rel_type and rel_type.lower() not in all_types:
-            all_types.append(rel_type)
-    if len(all_types):
-        info.albumtype = ', '.join(all_types).lower()
+        if release['release-group']['secondary-type-list']:
+            log.debug('Secondary Type(s) (new data): ' + ', '.join(
+                [secondarytype.lower() for secondarytype in \
+                    release['release-group']['secondary-type-list']]))
 
     # Release dates.
     release_date = release.get('date')
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -290,7 +290,7 @@ def album_info(release):
     if 'secondary-type-list' in release['release-group']:
         if release['release-group']['secondary-type-list']:
             log.debug('Secondary Type(s) (new data): ' + ', '.join(
-                [secondarytype.lower() for secondarytype in \
+                [secondarytype.lower() for secondarytype in
                     release['release-group']['secondary-type-list']]))
 
     # Release dates.
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -276,8 +276,9 @@ def album_info(release):
         disambig.append(release.get('disambiguation'))
     info.albumdisambig = u', '.join(disambig)
 
-    # Considers all release types (both primary and secondary) and
-    # stores them as a comma-separated string
+    # Retrieves the Release type.
+    # Considers all other release types(including primary and secondary) (new)
+    # and logs them.
     if 'type' in release['release-group']:
         reltype = release['release-group']['type']
         if reltype:
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -276,18 +276,22 @@ def album_info(release):
         disambig.append(release.get('disambiguation'))
     info.albumdisambig = u', '.join(disambig)
 
-    # Considers all release types (both primary and secondary) and stores as a comma-separated string
+    # Considers all release types (both primary and secondary) and
+    # stores them as a comma-separated string
     all_types = []
     if 'primary-type' in release['release-group']:
         rel_primarytype = release['release-group']['primary-type']
         if rel_primarytype:
             all_types.append(rel_primarytype.lower())
     if 'secondary-type-list' in release['release-group']:
-        for secondarytype in release['release-group']['secondary-type-list']:
-            all_types.append(secondarytype.lower())
-    if len(all_types) != 0:
-        all_types = ','.join(all_types)
-        info.albumtype = all_types.lower()
+        all_types.extend([secondarytype.lower() for secondarytype in\
+            release['release-group']['secondary-type-list']])
+    if 'type' in release['release-group']:
+        rel_type = release['release-group']['type']
+        if rel_type and rel_type.lower() not in all_types:
+            all_types.append(rel_type)
+    if len(all_types):
+        info.albumtype = ', '.join(all_types).lower()
 
     # Release dates.
     release_date = release.get('date')
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -276,11 +276,18 @@ def album_info(release):
         disambig.append(release.get('disambiguation'))
     info.albumdisambig = u', '.join(disambig)
 
-    # Release type not always populated.
-    if 'type' in release['release-group']:
-        reltype = release['release-group']['type']
-        if reltype:
-            info.albumtype = reltype.lower()
+    # Considers all release types (both primary and secondary) and stores as a comma-separated string
+    all_types = []
+    if 'primary-type' in release['release-group']:
+        rel_primarytype = release['release-group']['primary-type']
+        if rel_primarytype:
+            all_types.append(rel_primarytype.lower())
+    if 'secondary-type-list' in release['release-group']:
+        for secondarytype in release['release-group']['secondary-type-list']:
+            all_types.append(secondarytype.lower())
+    if len(all_types) != 0:
+        all_types = ','.join(all_types)
+        info.albumtype = all_types.lower()
 
     # Release dates.
     release_date = release.get('date')
--- a/beetsplug/mbcollection.py
+++ b/beetsplug/mbcollection.py
@@ -80,7 +80,7 @@ class MusicBrainzCollectionPlugin(BeetsPlugin):
             self.update_album_list([task.album])
 
     def update_album_list(self, album_list):
-        """Update the MusicBrainz colleciton from a list of Beets albums
+        """Update the MusicBrainz collection from a list of Beets albums
         """
         # Get the available collections.
         collections = mb_call(musicbrainzngs.get_collections)


anshuman73:
Ahh well, that's a lot of failing tests...

anshuman73:
Most tests (except for flake-8) are failing because of this - https://github.com/beetbox/beets/blob/master/test/test_mb.py#L131

This assertion fails due to the previous schema of the "release".

Possible fixes -
1) Update the schema in the test release object or
2) Add code where albumtype gets defined to type attribute if the primarytype and secondarytypes tags are not found.

beets/autotag/mb.py
```diff
@@ -276,11 +276,18 @@ def album_info(release):
...
-            info.albumtype = reltype.lower()
+    # Considers all release types (both primary and secondary) and stores as a comma-separated string
+    all_types = []
+    if 'primary-type' in release['release-group']:
+        rel_primarytype = release['release-group']['primary-type']
+        if rel_primarytype:
+            all_types.append(rel_primarytype.lower())
+    if 'secondary-type-list' in release['release-group']:
+        for secondarytype in release['release-group']['secondary-type-list']:
+            all_types.append(secondarytype.lower())
```
Kraymer: You can extend the list in one shot here :
all_types.extend([x.lower() for x in release['release-group']['secondary-type-list']])

beets/autotag/mb.py
```diff
@@ -276,11 +276,18 @@ def album_info(release):
...
-    # Release type not always populated.
-    if 'type' in release['release-group']:
-        reltype = release['release-group']['type']
-        if reltype:
-            info.albumtype = reltype.lower()
+    # Considers all release types (both primary and secondary) and stores as a comma-separated string
+    all_types = []
+    if 'primary-type' in release['release-group']:
+        rel_primarytype = release['release-group']['primary-type']
+        if rel_primarytype:
```
Kraymer: You check the string not emptiness in that case but not for secondarytype. Intentional ?

beets/autotag/mb.py
```diff
@@ -276,11 +276,18 @@ def album_info(release):
...
+            all_types.append(rel_primarytype.lower())
+    if 'secondary-type-list' in release['release-group']:
+        for secondarytype in release['release-group']['secondary-type-list']:
+            all_types.append(secondarytype.lower())
+    if len(all_types) != 0:
```
Kraymer: or just `if len(all_types):`

beets/autotag/mb.py
```diff
@@ -276,11 +276,18 @@ def album_info(release):
...
+        all_types = ','.join(all_types)
+        info.albumtype = all_types.lower()
```
Kraymer: these two lines could be done in one

Kraymer:
Welcome & thanks for tackling this issue!
At least the *docs* tox target stil passes ! ;-)

sampsyo:
Great! Thanks for getting started!

I have a couple of questions:

* Does the MB API no longer provide the plain "type" attribute? If it's been dropped, we should stop testing for it. If it does *sometimes* provide a "type" attribute, you're right: we'll need to try one style and then the other.
* To help with the style issues, you can try running flake8 locally. You can `pip install flake8` and then run `flake8 beets` or somesuch, or you can install Tox and then run `tox -e py27-flake8` or the like.

beets/autotag/mb.py
anshuman73: Hey !
Yes this was intentional
line 285 confirms if the key exists or not. The value for the key 'secondary-type-list' will always be a list, so if the key exists, it can either be an empty or full of elements, and in either way, the for loop handles the situation.

beets/autotag/mb.py
anshuman73: Yes, will do

Had another question though, should the joining string be ',' or ', ' (with a space) to follow the general standard everywhere of having a space after a comma in comma separated values .

anshuman73:
@sampsyo I guess the best place to know the answer to that is on the musicbrainz channel. I'll ask them there and correspond back to you here :)

anshuman73:
@sampsyo So I checked up with musicbrainz channel on irc. It mostly seems that the 'type' field is legacy (it isn't documented on http://wiki.musicbrainz.org/Release_Group/Type either)

Most of the testing confirms that type value is same as the secondarytypes value (http://dpaste.com/07H2C7T)

what do you suggest, should we keep type or replace it ?

sampsyo:
Awesome! Thanks for doing some investigation.

It sounds like the "type" field is deprecated. Since it's not completely clear yet how to get a 1-1 replacement with the new metadata fields, let's start by just adding some logging statements to record the new data. That way, we'll avoid disrupting anyone's workflow but we'll also be in a position to ask users what they want us to do with the new (and old) data.

Would you mind changing your current modifications to just use `log.debug` to record the new information (and leave the old information in place)? Then we'll call your GCI task complete.

beets/autotag/mb.py
```diff
@@ -276,11 +276,22 @@ def album_info(release):
...
-    # Release type not always populated.
+    # Considers all release types (both primary and secondary) and
+    # stores them as a comma-separated string
+    all_types = []
+    if 'primary-type' in release['release-group']:
+        rel_primarytype = release['release-group']['primary-type']
+        if rel_primarytype:
+            all_types.append(rel_primarytype.lower())
+    if 'secondary-type-list' in release['release-group']:
+        all_types.extend([secondarytype.lower() for secondarytype in\
```
Kraymer: backslash is redundant between bracket

beets/autotag/mb.py
```diff
@@ -276,11 +276,22 @@ def album_info(release):
...
+            release['release-group']['secondary-type-list']])
     if 'type' in release['release-group']:
-        reltype = release['release-group']['type']
-        if reltype:
-            info.albumtype = reltype.lower()
+        rel_type = release['release-group']['type']
+        if rel_type and rel_type.lower() not in all_types:
+            all_types.append(rel_type)
+    if len(all_types):
```
Kraymer: or just `if all_types:` :)

beets/autotag/mb.py
anshuman73: Ahh, any other way to avoid no. of characters > 79 error ?

beets/autotag/mb.py
anshuman73: sure :)

anshuman73:
@sampsyo , sorry for late reply

Also, I wonder how enabling logging feature will help to ask the user which data they want to use.
Freso did however suggested to me (later) to just go with the new replacements for the type (i.e the primarytype and secondarytype) field

beets/autotag/mb.py
anshuman73: But I think it's failing a flake-8 test due to this reason

sampsyo:
Ah, the idea is just to have the logging there so curious people can view what *would* be applied if we switched to the new data. Then, later, we can propagate these fields to the actual music.

The point is that people have configuration files out there that depend on `albumtype` being filled with certain values. We can consider adding new fields for the new data, but that's a bit to much work for the scope of this GCI task—so to make it easy, let's just log the data for now.

beets/autotag/mb.py
sampsyo: It's currently complaining that the slash is there when it doesn't need to be ("backslash is redundant").

anshuman73:
@sampsyo ,

I told this to Freso as well, I really don't mind the GCI task and everything. I'm here to help and make this better and in that way, learn a little as well. (And anyway, I'm still having problems claiming tasks, so I'm pretty sure I can keep on helping with this)

anshuman73:
That is ofc, if I'm able to and there aren't more pressing issues/features you have to work on

sampsyo:
Great! In any case, just for sane development practices, let's do the logging thing first. Then we can move on to inventing the new fields. In fact, the best way to attack the second stage will be through another ticket we put on GCI (#1547).

anshuman73:
Sure !
I'll get with the logging right away.
I haven't really used the log.debug feature before, so I may require help to get started ;)

sampsyo:
Great! You can see some other `log.debug` calls elsewhere in that `mb.py` file that might get you on the right track. You can try it out by running `beet` with the `-vv` flags to get very verbose output, including your new logging statements.

anshuman73:
Alright !
Also, should the log mention the primary type and secondary type separately or just log comma separated values including them both (current implementation)

sampsyo:
Let's log them separately. Thank you!

anshuman73:
Is there a a problem with the Appvoyer tests ?
Travis passes the same env tests that Appvoyer failed
Also Travis fails only because of Flake-8 :(

beets/autotag/mb.py
```diff
@@ -276,11 +276,22 @@ def album_info(release):
...
         if reltype:
             info.albumtype = reltype.lower()
+    if 'primary-type' in release['release-group']:
+        rel_primarytype = release['release-group']['primary-type']
+        if rel_primarytype:
+            log.debug('Primary Type (new data): ' + rel_primarytype.lower())
+    if 'secondary-type-list' in release['release-group']:
+        if release['release-group']['secondary-type-list']:
+            log.debug('Secondary Type(s) (new data): ' + ', '.join(
+                [secondarytype.lower() for secondarytype in \
```
sampsyo: This backslash is unnecessary. Just remove it and flake8 should be happy.

sampsyo:
Yes; thanks for pointing that out! We've broken some tests on Windows; I'll sort that out.

And thank you! This is looking good, and you're right that everything's passing except for flake8. As you can see from the log, the problem is another redundant backslash. Just remove that and everything should go green (and we can call the GCI task complete).

beets/autotag/mb.py
anshuman73: That may put up another flag for it being larger than 79 characters

anshuman73: I'll try nonetheless

anshuman73: oh wait
So I can remove the backslash and it still accepts it for being in the brackets !

Didn't know that. Oops. Sorry

anshuman73:
\o/ The Travis checks finally pass !

anshuman73:
I'll be closing this PR for the time we decide what to do with it so that the new commits don't get added to this one.

anshuman73:
Reopening this, as it was closed due to a stupid reason

sampsyo:
Sorry for the delay! I merged the logging code, which should bring us closer to a complete solution for storing multiple album types.

PR #2295, opened 2016-11-28
Fix rst syntax error in docs

musoke:


sampsyo:
Oops! Thanks for the fix. :sparkles:

PR #2297, opened 2016-12-01
Documentation improvement for convert plugin

pkess:
Hi all,

i just added some documentation for the convert plugin. It contains answers to questions that came up to me when using it for the first time some hours ago.

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -142,6 +142,11 @@ class ConvertPlugin(BeetsPlugin):
 
     def commands(self):
         cmd = ui.Subcommand('convert', help=u'convert to external location')
+        cmd.parser.usage = cmd.parser.usage + \
+            u' QUERY\n' + \
+            u'\n' + \
+            u'Convert tracks from your library to external location.\n' + \
+            u'Do not remove those tracks from library.'
         cmd.parser.add_option('-p', '--pretend', action='store_true',
                               help=u'show actions but do nothing')
         cmd.parser.add_option('-t', '--threads', action='store', type='int',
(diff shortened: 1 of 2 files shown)


docs/plugins/convert.rst
```diff
@@ -4,7 +4,9 @@ Convert Plugin
 The ``convert`` plugin lets you convert parts of your collection to a
 directory of your choice, transcoding audio and embedding album art along the
 way. It can transcode to and from any format using a configurable command
-line.
+line. This plugin is useful to convert existing tracks in your library
+to a path outside of your library. The tracks will not be removed from your
+library.
```
sampsyo: Well, that's not *entirely* correct: with the `-k` option, the convert plugin will keep the new, transcoded files instead of your original files. And when it's used in "auto" mode, the plugin converts newly-imported files on the fly.

Maybe a note like this would be better located below, under "Usage," which is where we explain what the command actually does?

beetsplug/convert.py
```diff
@@ -142,6 +142,11 @@ def __init__(self):
 
     def commands(self):
         cmd = ui.Subcommand('convert', help=u'convert to external location')
+        cmd.parser.usage = cmd.parser.usage + \
+            u' QUERY\n' + \
+            u'\n' + \
+            u'Convert tracks from your library to external location.\n' + \
+            u'Do not remove those tracks from library.'
```
sampsyo: This seems like a good addition, but it may be hard to preserve in the Click transition if we ever do it (c.f. #2190).

I also think the first sentence is a good synopsis but the second sentence might not be the most salient piece of information to include in help output. Can we stick to the first?

jtpavlock:
Closing as abandoned

Issue #2286, opened 2016-11-26
unidecode() needs a path escape


tweitzel:
### Alternate ticket title
Vaporwave artists are jerks

### Problem
Short version: ¢ gets decoded to 'C/' in ascii, which results in directories getting created mid-filename if using asciify_paths.

This album (track 21 of which is attached, as it causes the most weirdness) seems to be written solely to find bugs in beets, unicode, and filesystems. Notwithstanding the ¢ bug, track 25 got turned into a unicode comet emoji.

https://auto-7.bandcamp.com/album/greatest-hits
https://musicbrainz.org/release/c9dc947d-8fe1-4843-8ce9-7e9cdb56b64e

The database thinks this is the file:
/Users/xarph/Music/beets/hoesaAUTO/Greatest Hits/21 dollar menu (99C/ents + tax).mp3 - dollar menu (99¢ents + tax)

What happens instead is there's a file called 'ents + tax).mp3' inside a folder called '/Users/xarph/Music/beets/hoesaAUTO/Greatest Hits/21 dollar menu (99C'.
[UNICODE TOXIC WASTE HERE.zip](https://github.com/beetbox/beets/files/613985/UNICODE.TOXIC.WASTE.HERE.zip)

### Setup

* OS: Mac OS X 10.11.6
* Python version: 2.7.12
* beets version: 1.3.18
* Turning off plugins made problem go away (yes/no): no

My configuration (output of `beet config`) is:

```yaml
play:
    command: /usr/local/bin/mpv $args --display-tags='Artist,Album,Album_Artist,Genre,Title,Track,disc,date,icy-title,service_name' --vo null --no-audio-display --af=volume=replaygain-track --playlist
    warning_threshold: -2
    use_folders: no
    warning_treshold: 100
    raw: no
    relative_to:
scrub:
    auto: no
pluginpath: ~/work/whatlastgenre/plugin/beets/beetsplug
asciify_paths: yes
rewrite:
    "title \xA2": c
library: /Users/xarph/Music/beetslibrary.blb
original_date: yes
replaygain:
    backend: bs1770gain
    auto: yes
    overwrite: yes
    targetlevel: 89
    chunk_at: 5000
    method: replaygain

clutter: .DS_Store desktop.ini
importfeeds:
    formats: m3u_multi link
    dir: /Users/xarph/Music/beetsplaylist
    absolute_path: yes
badfiles:
    commands: {}

[...]

Issue #1541, opened 2015-07-12
asciify function not working properly


tomassvoboda:
Hello, my filenames are saved following this naming template: "$disc-$track - %asciify{$title}", but with the 11th song of the following release:
http://musicbrainz.org/release/433caf55-ee9d-4ee4-a33d-fce4ba3355c6 I get folder called "01-11 - 81" and in it file "2 Minutes.flac", instead of getting file "01-11 - 81_2 Minutes.flac", since underscore should be replacing all the slashes in filenames. (relevant part of my config: "replace:  '[\\/]': _")

tomassvoboda:
Sorry, another duplicate of https://github.com/sampsyo/beets/issues/1087

Issue #1159, opened 2014-12-21
Slashes not escaped when using asciify_paths


fortes:
This release http://musicbrainz.org/release/87981701-cc00-3c5c-af07-00050c115435 has a track with the unicode fraction character "½". When asciify_paths option is set, the `½` is converted to `1/2` and the `/` is never escaped.

PR #2303, opened 2016-12-03
Decompose unicode paths and unidecode each component. Fixes #1087 #1159 #1541 #2286 

### LINKED ISSUE #1087: In %asciify, asciify_paths, scrub path separators
Songs starting with "½" should be renamed like the following:

$artist/$album/01 - 1_2 song title.mp3

Actually they will be renamed like this:

$artist/$album/01 -1/2 sing title.mp3

So a new folder starting with "01" is created.

Suggestion:
We will probably need to special-case this to re-scrub path separators after asciification.

Also see:
https://groups.google.com/forum/#!topic/beets-users/cpY7zKRUGbA

Regards,
Matthias

tweitzel:
if asciify_paths on: breakdown incoming paths, run unidecode on each one, substitute os.path.sep into beets.config['path_sep_replace'], recompose paths with os.path.sep.


test/test_library.py
```diff
@@ -416,11 +416,14 @@ def test_unicode_extension_in_fragment(self):
 
     def test_asciify_and_replace(self):
         config['asciify_paths'] = True
-        self.lib.replacements = [(re.compile(u'"'), u'q')]
+        self.lib.replacements = [
+            (re.compile(u'"'), u'q'),
+            (re.compile(os.path.sep), config['path_sep_replace'].get())
```
sampsyo: Is this addition to the `replacements` necessary for this test? I don't quite see how it works…

It might also be nice to have a *separate* test just focused on the new functionality so we can see exactly what's being tested.

tweitzel:
The unicode_path_sep.mp3 was going to be used to make an end-to-end test of asciify_paths, but as a test newbie it was hard for me to figure out how to write the test, so I went with modifying the existing asciify test. If you can enlighten me as to what test frameworks are being used so I can look for documentation about them, I can try writing a more fully-featured test.

test/test_library.py
tweitzel: It was needed since the test just operates on a string, not a path. As you point out and I notice in hindsight, this test doesn't actually cover the whole code path that I modified, so it's back to the drawing board.

sampsyo:
Aha; thanks for explaining.

How about this test? Nice and simple, and I think it tests the change you've made.

```py
def test_asciify_character_expanding_to_slash(self):
    config['asciify_paths'] = True
    self.lib.directory = b'lib'
    self.lib.path_formats = [(u'default', u'$title')]
    self.i.title = u'ab\xa2d'
    self.assertEqual(self.i.destination(), np('lib/abC_d'))
```

tweitzel:
I'll throw it into my branch this week and finally get tests working
locally on my workstation, then update the PR. If this test is good, do
you still want the test .mp3 checked in?

sampsyo:
Cool; thanks for looking into it!

No—if this unit test suffices, there's no need for the new test MP3.

tweitzel:
I added the test you came up with off the top of your head because you're an old wizard from a kung fu movie and it passed all my tests; something was up with one of the travis tests, though.

sampsyo:
It me.

cthd

Anyway, something weird is up with Travis… I'm not sure what to make of it, but I don't think it's related. But actually, the Windows build on AppVeyor turned up something interesting: https://ci.appveyor.com/project/beetbox/beets/build/1141/job/mncrvbw3y1425e4e

In particular, the path still has backslashes where there were slashes introduced by "asciification." I think this means that the code should probably check for both `os.sep` and `os.path.altsep`, the latter of which is / on Windows and None on Unix. Here's where we do the same thing in dbcore: https://github.com/beetbox/beets/blob/ae5e55cad638dbde87af29bc21d0fb89996da995/beets/dbcore/db.py#L74-L77

Perhaps that should become a utility function?

tweitzel:
Yeah, the paths should all go through the same asciification process. The thing I'm not sure about (because I don't have quite an encyclopedic knowledge of the code), is at what stage the dbcore snippet comes into play - before or after the library asciify_paths bit? Is the output of dbcore used for subpath value, vice versa, or neither?

sampsyo:
Good question. It happens before. Here's the deal: dbcore is responsible for filling out templates and constructing paths in a general, abstract way; then the `library.py` stuff uses that to make beets-specific tweaks. In this case, the `asciify_paths` option is a post-processing step on the results produced by dbcore.

tweitzel:
It sounds to me like the question is: do we expect that paths are always going to be calculated based on asciify_paths as they move around inside beets, or have that applied only at the last minute as the library is about to write to the file system? Right now it looks like beets considers the output of dbcore as the path as objects are slung around by the longshoremen, and then at the last minute it gets asciified as part of library. If my understanding is right, we've got three options:

should asciify_paths be done in dbcore so everything is consistent all the time, which might open up more bugs later?
should asciify_paths become a utility function so other code paths and plugins that need it for some reason* can use it? (My vote)
should my existing pull be modified to include the os.path.altsep option and call it a day?

*I can see a plugin that exports to a fat32 usb stick or something using asciify_paths, but leaving the main library on a unicode-aware volume.

The other question is one of etiquette: barring option 3, do you want this pull request closed and a new one opened for the refactor, or keep working in this PR on my branch so context is kept?

sampsyo:
I'd agree with the longshoreman-slinging summary except for an important difference: I don't think anyone sees the direct output of dbcore's path formatting. For beets, *all* generated paths come from that `destination` method where asciiification is applied. So I don't think there's any danger that "raw" paths will leak out anywhere.

So yes, let's make it a utility function! In fact, we probably want to use it for the plain %asciify path function too, also in `library.py`: https://github.com/beetbox/beets/blob/master/beets/library.py#L1431
That's a strong argument for a reusable function here.

About PR etiquette, I don't have a strong opinion either way, but I generally find it easiest to keep on the same PR to continue the thread instead of switching to a new one.

tweitzel:
Aside from the one travis thing which I believe is fixed in a @jrobeson PR (per IRC), this looks good to go. Third time's a charm?

sampsyo:
Yep, that'd be #2310.

This looks great! Thank you for sticking with this. I'll merge this now. :sparkles: :rocket: All the folks watching the related tickets should get to thanking you profusely.

Issue #795, opened 2014-05-29
Allow play plugin to choose random albums


duailibe:
So, sometimes I don't feel like choosing what I want to listen and just want something random. Since we already have the great `random` plugin, I don't think duplicating features is the best way to go here.

I'm looking for some suggestions :-)

jubalh:
I am using:
```
beet random -n5 -p | xargs -d '\n' mpv
```
for similar purposes :-)

clach04:
* https://github.com/beetbox/beets/issues/795#issuecomment-177663532 - very neat!

I'm using:

    beet play $(beet random -n 1 -f '$title $artist')

This can/will generate more than one random track if there are dupes (or remixes).

Also:

    beet play "$(beet random -n 1 -p)"  # only works with one track, ensure handle spaces in path correctly

PR #2304, opened 2016-12-04
Add "play" option to "random" plugin

diomekes:
I copied some code from the random.py and added it to play.py to add two options to play plugin:
-r to randomize playlist items to play
-n to choose max number of random items to add (just like random plugin, default is 1)

The only problem I've encountered so far is that -n does not work without -r, because I did not add/edit any of the play plugin code.

I'm surprised I got it to work because I know no programming.

### DIFF
--- a/beetsplug/play.py
+++ b/beetsplug/play.py
@@ -18,20 +18,21 @@
 from __future__ import division, absolute_import, print_function
 
 from beets.plugins import BeetsPlugin
-from beets.ui import Subcommand, decargs, print_
+from beets.ui import Subcommand
 from beets import config
 from beets import ui
 from beets import util
+import contextlib
+import sys
+import cStringIO
 from os.path import relpath
 from tempfile import NamedTemporaryFile
-import random
-from operator import attrgetter
-from itertools import groupby
+from beetsplug import random
 
 # Indicate where arguments should be inserted into the command string.
 # If this is missing, they're placed at the end.
 ARGS_MARKER = '$args'
-
+_random = random.Random()
 
 class PlayPlugin(BeetsPlugin):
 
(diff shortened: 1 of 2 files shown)


sampsyo:
This is a good direction; thanks for taking the first step!

Before we consider merging this, we'll need to get rid of the duplicated code, which can become a maintenance nightmare down the road. This could work by just importing code from the random plugin module. Or, as #795 suggests, the "right way" to do this would be to add new query syntax that lets *any command at all* use random selection.

diomekes:
Does it not make more sense, then, for random to be an option to comands instead of a new query syntax?

sampsyo:
Right. Or rather, it's not that it doesn't make *sense* to extend commands—it's just that it would be *even better* to have query syntax. That way, all commands could take advantage of random selection "for free" without each one needing to be extended individually.

bearcatsandor:
Food for thought: It would be really cool to have a random by time function. Such that when you had 30 minutes before you raced out the door, you could give the randomizer a time and it would pick tracks that would fit within that time constraint.

sampsyo:
Great minds think alike, @bearcatsandor! I have been fantasizing about exactly the same idea lately, for when I have a meeting coming up and I need to fill the time with music.

For what it's worth, you can already randomize based on the length of an individual song already:

```
$ beet random length:120..240
```

but that's not exactly what you were proposing, of course.

bearcatsandor:
I'd love for it to be part of the random command, not just the play command, so that it could be piped to other commands. (i feel like i'm thread-jacking a bit now. Can this be spiraled off to it's own request?)

sampsyo:
Yes, good point—please do start a new thread for this (nifty) new feature request.

bearcatsandor:
https://github.com/beetbox/beets/issues/2305

sampsyo:
Here's one case where I frequently wish for random selection in queries: the web interface. It would be awesome to type `rand:5` or something into the search box to bring up a few songs from my library.

diomekes:
I've updated the play plugin with hopefully better code to reduce repetition.  If it is merged, the play plugin will require the random plugin to play random items.

I decided to add the functionality to the play plugin since it made more sense to have "play random" then "random play."  

I'm not sure if there's a better way to suppress stdout from an imported module's function, and what I currently have is the best I could come up with, but maybe there's a more elegant solution.

sampsyo:
OK, thanks for helping out! This does avoid the duplicated code from before. I'll continue to fiddle with this, but the ultimate solution may be to wrap your code up in query syntax.

diomekes:
I imagine a list needs to be created from the items returned by a query and then shuffled and shortened.
  I can't figure out how do that through a query prefix.

Doesn't the query run before the creation of a list of values that can be manipulated?  I think dbcore.query returns True or False on patterns and values, how can w randomize values matched in the same function?

sampsyo:
Yeah, it would be a little tricky—we'd need to design a new kind of query modifier that lets us take a random subset of the results that would be returned otherwise.

diomekes:
See #2350.

diomekes:
See #2350.

PR #2307, opened 2016-12-05
Check python version and enable https where it's possible

tigranl:


Freso:
Do we have a minimum support Python 3 version? (SNI wasn't included in Python 3.0.0 either, so the check might need to check for early versions of Py3 as well… Depending on minimum supported Py3 version.)

beets/autotag/mb.py
```diff
@@ -28,9 +28,14 @@
 from beets import util
 from beets import config
 import six
+import sys
 
 VARIOUS_ARTISTS_ID = '89ad4ac3-39f7-470e-963a-56509c546377'
-BASE_URL = 'http://musicbrainz.org/'
+
+if sys.version_info > (2, 7, 9):
```
Freso: The comparison should "greater than or equal to", since 2.7.9 also has support for it.

sampsyo:
My best guess from [this thread](https://github.com/kennethreitz/requests/issues/749) is that 3.2+ supports SNI. Fortunately, 3.2 is below our minimum required Py3 version.

Would you mind defining a `SNI_SUPPORTED` variable in the `beets.util` module so we can reuse it in other places? Other places where this might be needed are revealed by @orf's PRs from a couple weeks ago: https://github.com/beetbox/beets/pull/2247 https://github.com/beetbox/beets/pull/2243 https://github.com/beetbox/beets/pull/2245 https://github.com/beetbox/beets/pull/2244 https://github.com/beetbox/beets/pull/2248

sampsyo:
OMG I wish @codecov-io would stop commenting. :angry:

This is looking great so far! The last thing we'll need once all the endpoint changes are in place is to update the tests. Some tests use mocking to return fake responses to URL requests; these typically use hard-coded URLs. We'll need to change the tests to return the mocked response for both HTTP and HTTPS versions of each URL.

sampsyo:
I meant to suggest that `SNI_SUPPORTED` be a *boolean*, not just the current version number. Then the code in each module can look like this:

```py
if ui.SNI_SUPPORTED:
```

instead of this:

```py
if ui.SNI_SUPPORTED >= (2, 7, 9):
```

This variable should also probably reside in the `util` module, not the `ui` module.

tigranl:
`SNI_SUPPORTED` is defined in the `ui` module. Should I move it to the  `util` module?

sampsyo:
Yes, please.

tigranl:
Okay, so, what shall we do with tests?

sampsyo:
OK, let's take a look! The rough prescription is:

1. Look at the test failures (either locally on your machine or on Travis) to find the tests that need updating.
2. For each, look for the URL that's being mocked. Add a second mock call that uses HTTPS instead of HTTP so we catch both.
3. In the end, check that all the tests pass without network access—e.g., with the WiFi on your laptop disabled. If that works, we'll know that all the mocking does what we want it to do.

It would also be nice to ensure that all the endpoints *actually do support HTTPS*. That just involves running each piece of functionality to make sure it still works when contacting the server over HTTPS.

tigranl:
Hi, @sampsyo. Can you give me some comments on my work? I have never written unit tests before.

test/test_art.py
```diff
@@ -68,17 +68,21 @@ def mock_response(self, url, content_type='image/jpeg', file_type=None):
...
         self.source = fetchart.RemoteArtSource(logger, self.plugin.config)
         self.extra = {'maxwidth': 0}
         self.candidate = fetchart.Candidate(logger, url=self.URL)
+        self.candidate_https = fetchart.Candidate(logger, url=self.URL_HTTPS)
 
     def test_invalid_type_returns_none(self):
         self.mock_response(self.URL, 'image/watercolour')
+        self.mock_response(self.URL_HTTPS, 'image/watercolour')
         self.source.fetch_image(self.candidate, self.extra)
+        self.source.fetch_image(self.candidate_https, self.extra)
```
sampsyo: Since this test controls the URL on "both ends," i.e. the request and the response, I'm not sure it needs updating. Was it failing before?

test/test_art.py
```diff
@@ -157,13 +165,21 @@ class CombinedTest(FetchImageHelper, UseThePlugin):
     CAA_URL = 'http://coverartarchive.org/release/{0}/front' \
               .format(MBID)
 
+    AMAZON_URL_HTTPS = 'http://images.amazon.com/images/P/{0}.01.LZZZZZZZ.jpg' \
+                 .format(ASIN)
+    AAO_URL_HTTPS = 'http://www.albumart.org/index_detail.php?asin={0}' \
+                 .format(ASIN)
+    CAA_URL_HTTPS = 'http://coverartarchive.org/release/{0}/front' \
+                 .format(MBID)
```
sampsyo: Is there any way we can avoid copying and pasting here? For example, we could put most of the URLs (starting with `://`) into a variable and then concatenate on the `https` or `http` part separately.

test/test_art.py
```diff
@@ -176,38 +192,46 @@ def test_main_interface_returns_none_for_missing_asin_and_path(self):
...
         self.assertFalse(candidate.path.startswith(self.dpath))
 
     def test_main_interface_tries_amazon_before_aao(self):
         self.mock_response(self.AMAZON_URL)
+        self.mock_response(self.AMAZON_URL_HTTPS)
         album = _common.Bag(asin=self.ASIN)
         self.plugin.art_for_album(album, [self.dpath])
         self.assertEqual(len(responses.calls), 1)
         self.assertEqual(responses.calls[0].request.url, self.AMAZON_URL)
+        self.assertEqual(responses.calls[0].request.url, self.AMAZON_URL_HTTPS)
```
sampsyo: This seems to assert that the requested URL is equal to two *different* URLs, which probably isn't what you meant—instead, you probably want to write an assertion that says that the requested URL is *either* the HTTP one or the HTTPS one.

sampsyo: (The next couple of tests seem to have the same issue.)

tigranl:
Hi, @sampsyo. Sorry for delay. I found that amazon and albumart.org don't support https, so I removed https urls for them.

sampsyo:
No worries, and thank you!

It looks like Travis still has a few style/error reports: https://travis-ci.org/beetbox/beets/jobs/182870011#L848-L853

It also looks like there might be a typo in some URL, where there's "htts" instead of https? https://travis-ci.org/beetbox/beets/jobs/182870006#L913

tigranl:
It's strange, because I fixed typo in one of my previous commits https://github.com/beetbox/beets/pull/2307/commits/b65a7da8e2e6497ab1ef168e2073545875ec39d7

sampsyo:
Got can behave in mysterious ways. :ghost:

Those flake8 errors shouldn't be hard to address either.

tigranl:
Okay, my numerous commits look ridiculous. I should do something with it.

sampsyo:
It's cool; I don't have anything against long series of commits. But if you like cleaner histories, the terms to google are [git sqash].

jrobeson:
@sampsyo : I'd definitely prefer if we use "Squash and merge" button in cases like this. @tigranl : we can take take care of that on our end by pressing a button

sampsyo:
Yeah, that's cool too.

tigranl:
It would be great.

beetsplug/lastimport.py
```diff
@@ -17,13 +17,17 @@
...
 from beets import dbcore
 from beets import config
 from beets import plugins
 from beets.dbcore import types
 
-API_URL = 'http://ws.audioscrobbler.com/2.0/'
+if util.SNI_SUPPORTED:
+    API_URL = 'https://ws.audioscrobbler.com/2.0/'
+else:
+    API_URL = 'https://ws.audioscrobbler.com/2.0/'
```
Kraymer: https => http ?

tigranl: I think it's less readable.

sampsyo:
Merged with a few changes. Thank you! :sparkles:

PR #2309, opened 2016-12-06
try discogs without vendored urllib3 logger

jrobeson:
🚨 🚫 DO NOT MERGE 🚫 🚨


jrobeson:
this didn't achieve what we were trying to test. so it shall go.

PR #2310, opened 2016-12-07
install/update requests to satisfy sphinx

jrobeson:
The version of `requests` installed as a site package is too old to work with the version
of `sphinx` installed via `tox`, so we manually update it.


sampsyo:
Looks like it worked! :tada:

PR #2311, opened 2016-12-09
Remove requests/urllib3 logging override in discogs

jrobeson:
Discog plugin overrides the global logging for `requests` for a reason
that seems no longer valid.

I can't seem to reproduce the logging behavior mentioned in the
preceding comment, so I'm removing it.

If we do want to modify `requests` logging in the future, then we should
hook it into beets verbose logging.


sampsyo:
Great! Merged. Let's see if anybody gets spammed with logging…

Issue #1089, opened 2014-11-15
Do not consider [silence] as "missing tracks"


bkanuka:
Many albums have silent tracks, labelled as "[silence]" in musicbrainz. If these tracks are missing, it should be ignored.

sampsyo:
For `opatel99` in #metabrainz on IRC:

This should be fixed in a few different steps.

1. First, we'll need to *detect* silence tracks. This will happen in the `autotag.mb` module. The challenging part here is to decide exactly what criterion determines silence. Is there a specific MB ID we can use? Or do we just detect the title string "[silence]", as suggested above?
2. Next, we need to record the fact that the track is "ignorable" somewhere. This will require extending the `TrackInfo` class in the `autotag.hooks` module with a Boolean flag called `ignorable` or `inessential` or something. It will be important to make sure we don't break anything with this change by running the tests again after this step.
3. Finally, we need to tell the matching logic in `autotag.match` that it's OK to ignore TrackInfo objects with that flag set.

We can get into more detail on the specific steps when we come to them, but that's the high-level outline.

Freso:
As [I wrote on IRC](http://chatlogs.metabrainz.org/brainzbot/musicbrainz/msg/3452228/):
> The style guidelines also says that artist intent overrules all, so not all silent tracks are named "[silence]" and some artist out there may have made a track called "[silence]" that isn't silence.

I'm not saying it's a bad idea to filter on "[silence]" though, as that seems to be the most reliable thing to do detect silent tracks on MB regardless of reservations above - just as long as those reservations are kept in mind.

bkanuka:
@sampsyo I agree with most of what you said, however the original intent of my bug was not to detect silence tracks - I think they are already labeled as such in Musicbrainz.

To clarify, occasionally there is an album with CD tracks that are silent and unnamed (don't appear on the album art, but exist as tracks when played on a CD player). The Musicbrainz spec is to give these a title of `[silence]`.

The issue arises when you rip the CD and delete these silent tracks. Currently beets will say "missing tracks", but some may argue that there's no harm in these being missing - and will not want to "punish" or risk mis-tagging due to their absence.

Sorry if this was clear to everyone and I am beating a horse - just wanted to be on the same page. Merry Christmas or Festivus or whatever everyone :-)

sampsyo:
Yep, that's the goal. The first step is detecting the silence; then we can ignore those tracks. Thanks for checking back in!

tigranl:
[...]

PR #2315, opened 2016-12-12
[Silence] tracks

tigranl:


sampsyo:
So far so good! We'll also want to *use* the ignorable flag to avoid penalties (and set the flag according to silence, of course: #1089).

tigranl:
I guess, we need to find the number of `[silence]` tracks, and then subtract it from total amount of tracks. After that, we can use it in the for loop: https://github.com/tigranl/beets/blob/a39ab1057f4fefdc3d689be60954d8eed990bfb1/beets/autotag/match.py#L245

sampsyo:
Yep, pretty much exactly like that!

beets/autotag/mb.py
```diff
@@ -165,6 +170,7 @@ def track_info(recording, index=None, medium=None, medium_index=None,
         medium_total=medium_total,
         data_source=u'MusicBrainz',
         data_url=track_url(recording['id']),
+        ignorable=flag
```
sampsyo: Can you please add a trailing comma here?

beets/autotag/mb.py
```diff
@@ -156,6 +156,11 @@ def track_info(recording, index=None, medium=None, medium_index=None,
     ``medium_index``, the track's index on its medium; ``medium_total``,
     the number of tracks on the medium. Each number is a 1-based index.
     """
+    flag = False
+
+    if recording['title'] == '[silence]':
```
sampsyo: Can you please put the magic string `'[silence]'` into a constant at the top of the file? For example, the name `SILENCE_TITLE` would work. (It can go next to the other constants at the top of the file, such as `VARIOUS_ARTISTS_ID`.)

beets/autotag/mb.py
```diff
@@ -156,6 +156,11 @@ def track_info(recording, index=None, medium=None, medium_index=None,
     ``medium_index``, the track's index on its medium; ``medium_total``,
     the number of tracks on the medium. Each number is a 1-based index.
     """
+    flag = False
```
sampsyo: Please choose a more descriptive name for this variable—for example, `silence`.

beets/autotag/mb.py
```diff
@@ -156,6 +156,11 @@ def track_info(recording, index=None, medium=None, medium_index=None,
...
+
+    if recording['title'] == '[silence]':
+        flag = True
```
sampsyo: Instead of this pattern:

```py
some_variable = False
if some_boolean:
    some_variable = True
```

It's simpler to just write:

```py
some_variable = some_boolean
```

beets/autotag/match.py
```diff
@@ -241,8 +241,10 @@ def distance(items, album_info, mapping):
         dist.tracks[track] = track_distance(item, track, album_info.va)
         dist.add('tracks', dist.tracks[track].distance)
 
+    count_ign = len([x.ignorable for x in album_info.tracks if x.ignorable])
```
sampsyo: Two really tiny comments: first, this is a slightly clearer way to write this.

```py
sum([x.ignorable for x in album_info.tracks])
```

Second, can you please move this under the `# Missing tracks.` comment, because it's related to that code?

sampsyo: Hmm, maybe that bears further investigation? It works as expected here:

```py
>>> sum([True, False, False, True])
2
```

tigranl: @sampsyo that's what I was talking about: https://travis-ci.org/beetbox/beets/jobs/184632489#L859

sampsyo: OK, can you please look into it to find out why this is happening? It looks like some of the values are None, suggesting that the TrackInfo objects had `ignorable` set to `None` somewhere.

tigranl: The easiest solution is to set `ignorable` to 0.

sampsyo: Or `False`, instead of 0, so that it always has the same type (`bool`).

sampsyo:
OK, looks good! I'll approve the GCI task now. If you're interested in seeing this to the next step, that would be putting this through its paces and observing the effect on the UI, as described above:

> Also, have you tried the code out? That is, before we merge, it would be nice to try importing something that will exercise this code. You can do that, for example, by running beet import -t on any album at all, choosing "enter Id", and then pasting in the ID for the Nine Inch Nails album above that has all the silence tracks.
> 
> In particular, I think we might also want to do one extra thing to avoid displaying "missing tracks" in the UI. This will probably work by adjusting the extra_tracks list constructed in assign_items.

tigranl:
I actually run it. It works as supposed, but tracks aren't in the right order.

Issue #2305, opened 2016-12-05
random:  Return random tracks by time constraint


bearcatsandor:
It would be really cool to have a random by time function. Such that when you had 30 minutes before you raced out the door, you could give the random command a time and it would pick tracks that would fit within that time constraint.

Then the resulting tracks could be piped to other commands, and utilized for the "play" command as well.

One could do:

beet random -t 37 
to return 37 minutes of random tracks 

beet random -t 45 "Metallica" 
to return 40 minutes of random Metallica 

beet random -t 120 title:cats
 to return 2 hours of random tracks with the word "cats" in the title.

If one didn't have 2 hours worth of tracks with the word "cats" in them, it would just return as much as it can.

Random could be set up to return a defined number of tracks as well, but that's another Issue.

diomekes:
I've made changes to the random plugin to implement this.  I won't open a pull request because it has other changes that I made for #2304, but here's a pastebin of it:

https://ptpb.pw/tJ4N

bearcatsandor:
That looks awesome and better than i'd dreamed! Thank you!!!

sampsyo:
Nice work, @diomekes! This looks awesome. I'd love to merge the new feature separately if you're amenable to splitting up the patches.

diomekes:
Yes, I can do that.  I'll add the code to the current random plugin (which does not define several functions, just the main one, so I'll add the code under that function).

Can I run the tests that github runs on pull-requests on my end (either off-line or through github) so I can fix small issues before requesting a pull?

sampsyo:
Yes, awesome! Thank you!

Here are some tips on running tests: https://github.com/beetbox/beets/wiki/Testing

And yes, please do add some docs (even just a draft is good).

diomekes:
OK, I'm almost done with the plugin code.  Quick question: Should I make -e and -t mutually exclusive or should I add a if opts.time into the if opts.equal_chance?  Here are pastebins to see the two options:
https://ptpb.pw/_238
https://ptpb.pw/Z09i

sampsyo:
Huh—interesting question. Can you explain intuitively how things work if -e and -t are used together?

diomekes:
I'll try, if I understand -e correctly.

-e chooses a song by first making a list of artists and choosing a random artist from that list (instead of a list of songs, and choosing a random song).  That way, it doesn't matter how many songs from each artist you have, all artists are treated equal and once it chooses all the songs from one artist, that artist is deleted from the list.

[...]

PR #2322, opened 2016-12-15
random: add --time option, update docs

diomekes:
Adds feature #2305

### DIFF
--- a/beetsplug/random.py
+++ b/beetsplug/random.py
@@ -42,6 +42,8 @@ def random_item(lib, opts, args):
             objs = []
 
             for item in objs_shuffled:
+                if opts.album:
+                    item.length = sum(a.length for a in item.items())
                 if (total_time + item.length) <= time_sec:
                     objs.append(item)
                     total_time += item.length
@@ -63,6 +65,8 @@ def random_item(lib, opts, args):
                 if not objs_by_artists:
                     break
 
+                if opts.album:
+                    item.length = sum(a.length for a in item.items())
                 if (total_time + item.length) <= time_sec:
                     artist = random.choice(list(objs_by_artists.keys()))
                     objs_from_artist = objs_by_artists[artist]
--- a/beetsplug/random.py
+++ b/beetsplug/random.py
@@ -32,6 +32,23 @@ def random_item(lib, opts, args):
     else:
         objs = list(lib.items(query))
 
+    if opts.time:
+        total_time = 0.0
+        time_sec = (opts.time * 60)
+        objs_shuffled = objs
+        random.shuffle(objs_shuffled)
+
+        if not opts.equal_chance:
+            objs = []
+
+            for item in objs_shuffled:
+                if (total_time + item.length) <= time_sec:
+                    objs.append(item)
+                    total_time += item.length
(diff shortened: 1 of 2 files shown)


diomekes:
I just realized that this does not work with -a.  I need to figure out how to list the individual tracks of an album in order to get the total length.

sampsyo:
Ah, tricky!

Someone was asking on IRC about adding a pseudo-field to calculate exactly that. But in the mean time, you can use something like `sum(a.length for a in album.tracks())` to get the total time for an album.

diomekes:
It's pretty slow with -a, maybe someone know how to optimize it.  Otherwise, it seems it all works.

bearcatsandor:
It seems fast enough for me with -a.  I love it! Are there still plans to integrate it with the 'beet play' command?

diomekes:
I have #2304 open that adds --random to the play plugin, but if this gets merged, I'll close that one and open a new one because I had made some changes that didn't need to be made.

bearcatsandor:
awesomeness.

diomekes:
Are there any other steps I need to take in order to get this ready to be merged?

sampsyo:
Nope; I was the bottleneck here, sorry! I've merged, and I've also done some substantial refactoring. Can you please take a look to check that I didn't mess anything up too badly?

Also, one thing came up when I was looking over the code: I noticed that your strategy will often include the shortest songs in a person's library. That is, if I have a 2-second song somewhere, then it's quite common that there will be a 2-second shortfall when selecting other songs, so that short song is very likely to be picked at the end.

I don't have any great ideas for how to resolve that, but I thought I'd mention it.

diomekes:
Those changes to the code look great, thanks for taking the time to clean it up.  My simple tests show everything is all good.

As far as the short song issue, I though of that too, but couldn't think of any way around it.  A user-based solution might be to use queries like length to filter out a known short song that keeps being used, but with a big library there usually are several short songs that would fill the gap, and maybe not be repeated as often.

ArchangeGabriel:
Another option could be to have an option of a minimum duration for a song to be considered. For instance, I might not want any song less than 1 min to be added to the playlists: I have no way to check this right now, but I, for one, expect all such songs of my collection to have no interest out of context. ;)

sampsyo:
I like the minimum-length heuristic. It might be worth noting that the problem of finding a playlist of music that comes closest to a given maximum time (without going over) is exactly the [bin packing problem](https://en.wikipedia.org/wiki/Bin_packing_problem), which is NP-complete. That implies that it might be tricky to get a great solution here, although the addition of randomness makes the question a little more complicated.

sampsyo:
Yes, good point—that query does work out of the box for avoiding short songs, @diomekes.

I've also spent a little while thinking about the algorithmic problem (randomly sampling a bin packing solution) and I still don't see a clearly better solution than the current greedy algorithm. Maybe one small improvement, however, would be to shuffle the resulting list before returning it. This way, at least the short tracks won't always appear at the end of the playlist?

diomekes:
Yes, the list could be shuffled a second time after going through the timed selection process.

PR #2323, opened 2016-12-16
convert plugin: fix braces in filenames causing tracebacks in "pretend" mode

kierdavis:
Previously "pretend" mode (a.k.a. dry run mode) passed the command to be
printed directly to `_log.info`, whose first argument is technically a
format string. Thus the command string was parsed for replacement fields, such
as `{foo}`, which could cause the format evaluation to fail if the filenames
contained in the command contained valid (or partially valid) replacement fields.

This fix simply inserts an argument `'{0}'` to the call to `_log.info`, which is a
format string that simply evaluates to the second argument to `_log.info` (the
command string). By doing this, the command string is not parsed for replacement
fields.

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -202,7 +202,7 @@ class ConvertPlugin(BeetsPlugin):
                 encode_cmd.append(args[i].encode(util.arg_encoding()))
 
         if pretend:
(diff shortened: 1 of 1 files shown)


sampsyo:
Ah, this makes perfect sense! Thank you for catching this. :sparkles:

Issue #2324, opened 2016-12-16
convert: symlink/hardlink instead of copying files that don't need transcoding


kierdavis:
My music workflow involves creating a copy of my library in MP3 format, so that it can be uploaded to Google Play Music. The convert plugin does this brilliantly; however, since most (but not all) of my library is in MP3 format to begin with, this results in copying a lot of files. It looks like it should be fairly straightforward to allow convert to symlink or (when supported) hardlink files that don't need to be transcoded, instead of copying them. This would make the conversion process faster and the final result would generally occupy less disk space.

sampsyo:
Sounds like a good idea; thanks!

Vexatos:
It might be a neat idea to allow specifying a custom copy command, like
```yaml
convert:
  copy_command: ln $source $dest
```
This would allow any sort of links, including hardlinks and reflinks, if the user knows what they are doing.

kierdavis:
Resolved by #3348

wisp3rwind:
For the record, there's also https://github.com/geigerzaehler/beets-alternatives

PR #2326, opened 2016-12-16
convert: add option to symlink instead of copying

kierdavis:
As proposed in #2324.

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -105,6 +105,7 @@ class ConvertPlugin(BeetsPlugin):
         self.config.add({
             u'dest': None,
             u'pretend': False,
+            u'link': False,
             u'threads': util.cpu_count(),
             u'format': u'mp3',
             u'formats': {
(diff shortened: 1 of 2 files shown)


beetsplug/convert.py
```diff
@@ -156,6 +157,12 @@ def commands(self):
                               help=u'set the target format of the tracks')
         cmd.parser.add_option('-y', '--yes', action='store_true', dest='yes',
                               help=u'do not ask for confirmation')
+        cmd.parser.add_option('-l', '--link', action='store_true', dest='link',
+                              help=u'symlink files that do not need \
+                                     transcoding, instead of copying them; \
+                                     this may cause files in your library to \
+                                     be modified, if options such as "embed" \
+                                     are enabled')
```
sampsyo: The in-line help string should usually be shorter than this—just one phrase is fine. We can leave the full detail for the documentation.

docs/plugins/convert.rst
```diff
@@ -86,6 +93,13 @@ file. The available options are:
 - **threads**: The number of threads to use for parallel encoding.
   By default, the plugin will detect the number of processors available and use
   them all.
+- **link**: By default, files that do not need to be transcoded will be copied
+  to their destination. This option creates symbolic links instead. Note that
+  options such as ``embed`` that modify the output files after the transcoding
+  step will cause the original files to be modified as well if ``link`` is
+  enabled. For this reason, it is highly recommended not use to ``link`` and
+  ``embed`` at the same time.
+  Default: ``false``.
```
sampsyo: There's some identical text between this and the -l option. For maintainability, could you please just leave one copy and refer to one from the other? For example, you might just say "see the description of the `link` option below" or something.

kierdavis:
@sampsyo Thanks for the quick feedback. I did think it felt quite lengthy when I wrote it, but I wanted to ensure users are aware that embed (which is on by default) and link used together may cause the original files to be modified. Maybe a better solution would be to add an "are you sure?" prompt in the code when this combination of options is detected; what do you think?

sampsyo:
Sure! Or here are two other ideas:

- Just a short phrase like "(don't use with `embed`, probably)" as a warning there.
- The plugin could just refuse to run with both flags enabled—I doubt anyone will actually want that behavior.

Issue #2274, opened 2016-11-21
The zero plugin needs a manually triggered mode


nathdwek:
As of now, the `zero` plugin is only triggered at import time, which means that if a user wants to use the plugin down the road or change its configuration, he has to re-import his whole library.

As is the case with `fetchart`, `embedart`, `convert`, ... I propose that the plugin provides a `zero` (extra points for creativity) command to compliment the import hooks.

sampsyo:
Yes, good call.

sampsyo:
Fixed in #2329. Thank you, @SJoshBrown!

Issue #1286, opened 2015-01-31
default query that affects entire library is dangerous


randybias:
Maybe this is subjective, but I don't think so... Basically, the default query being '' is extremely dangerous.  A simple typo on the command line can result in something quite bad such as:

beet modify genre="Electro"

Which is the same as:

beet modify genre="Electro" ""

It should take a lot more deliberateness to do a library-wide change like that.  And in fact, something like "beet info" does the right thing in that "beet info" does nothing and you have to issue:

beet info ' '

To get "all".  Which, by the way, is a second argument for making this change.  Probably we want consistency across the various command line arguments as much as possible.

Related, arguably, the query for all should be "*", but maybe that's a whole separate ticket.

sampsyo:
Thanks for bringing this up. I'd like to hear others' opinions about this. Personally, I think `beet ls` should list the entire library, but I would be interested in hearing other stories.

randybias:
I think "beet ls" should list the whole library too.  Maybe that's an exception.  But none of the write operations should default to 'all'.

sampsyo:
It's difficult to see where the line should be drawn. `beet lyrics` should probably fetch all the lyrics, for example, and I'm very accustomed to `beet move` being a global operation by default.

randybias:
Really?  My library will be over 10000 at some point. I don't want lyrics to run as a default on that. Or anything else really. Right now my test set is 300-ish but once I do a full import I will be in the low 1000s. Once I add my non-DJ collection it will be close to 10k.

I can't be the only one with that big a music library. =

randybias:
I feel pretty strongly about this one but I would be happy even if just beet modify didn't have this behavior. There isn't really any reason to run it against all by default. =

sampsyo:
Another option might be to add a confirmation option that confirms when the query arguments are empty. Again, I'd love to hear general opinions on this if anyone's watching the thread.

Kraymer:
I use quasi exclusively the *import* command so this problematic is kinda remote for me but I know that the rare times I used the *beet ls* command, the fact that ``beet ls`` lists the whole library seems counter intuitive.

NB: but I'm a big fan of the `*` character , as made obvious by #1178 where I vouched for it against my better judgment, so ... just my 2c

Kraymer:
After fiddling a bit with *ls* and *rm* command, I have no opinion which is better : * or nothing to consider all files.

> My library will be over 10000 at some point. I don't want lyrics to run as a default on that. 

[...]

PR #2329, opened 2016-12-18
Add manually triggered mode for zero plugin

SJoshBrown:
I'm fairly new to the process so I hope I've done this correctly. Please let me know if I need to make any changes to get this merged up.

This is for Issue #2274

### DIFF
--- a/beetsplug/zero.py
+++ b/beetsplug/zero.py
@@ -23,7 +23,7 @@ import re
 from beets.plugins import BeetsPlugin
 from beets.mediafile import MediaFile
 from beets.importer import action
-from beets.ui import Subcommand, decargs
+from beets.ui import Subcommand, decargs, input_yn
 from beets.util import confit
 
 __author__ = 'baobab@heresiarch.info'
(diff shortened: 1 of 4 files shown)


SJoshBrown:
I've taken another pass at this and factored out some of the shared code. I also added a few more tests to increase the coverage.

nathdwek:
I'm interested in this feature being added for consistency's sake, so I took a look at the whole plugin.
There is only one important thing I think should be added before going through with this, that is adding safeguards in manual mode when the query is empty, because of the nature of the operation (see #1286 )

However, I think the original plugin is a mess in terms of style and structure, cf my review. Maybe this is the occasion to revisit that?

sampsyo:
Indeed, the plugin as it was was a bit of a mess. Thanks for looking into it, @nathdwek!

nathdwek:
I opened #2345 . Some unexpected issues in the testing department, but would you be able to rewrite on top of that when that's fixed? We can also do it the other way around if it's more convenient for you?

SJoshBrown:
I'm happy to wait until it gets some cleaning up. Then I can come back and rewrite this

nathdwek:
I'll see tomorrow if I can sort this test thing out with a fresh mind. But my absolute priority is not getting in your way!

sampsyo:
Either way seems to work from my perspective—I'd also be happy to merge this new feature first and then do the refactoring afterward.

SJoshBrown:
@sampsyo  @nathdwek, I think this should be ready to go now. Let me know if there are any other changes we should make

sampsyo:
All merged up. :sparkles: Thank you for taking care of this!

PR #2333, opened 2016-12-20
Add Composer, Lyricist and Arranger tags

ibmibmibm:
MusicBrainz provids composer, lyricist and arranger infomations related
to individual recordings. This commit adds query parameters to fetch them, and
write down to media files.

Tagging mapping is implemented according MusicBrainz Picard's data:
https://picard.musicbrainz.org/docs/mappings/

Signed-off-by: Shen-Ta Hsieh <ibmibmibm.tw@gmail.com>


ibmibmibm:
This pr implement some fields of #1547

beets/autotag/__init__.py
```diff
@@ -44,6 +44,13 @@ def apply_item_metadata(item, track_info):
         item.mb_artistid = track_info.artist_id
     if track_info.data_source:
         item.data_source = track_info.data_source
+
+    if track_info.lyricist is not None:
+        item.lyricist = track_info.lyricist
+    if track_info.composer is not None:
+        item.composer = track_info.composer
+    if track_info.arranger is not None:
+        item.arranger = track_info.arranger
```
sampsyo: Can you please add a blank line between this stanza and the comment below?

ibmibmibm: Added.

ibmibmibm:
As reminder, standalong recording's relation fetching needs [this pull request](https://github.com/alastair/python-musicbrainzngs/pull/213), so we'll need a newer version of musicbrainzngs not yet released.

sampsyo:
Ah, got it—thanks. So it sounds like we have two options:
1. Don't merge this until there's a new release of the client library.
2. Add some version checking so we just don't get the data on the current version of the library.

ibmibmibm:
I've add a commit that automatically detects old version of musicbrainzngs, and disable this feature.

sampsyo:
Wonderful; thank you! This feature detection appears to work perfectly. This is all merged up. :sparkles:

dosoe:
Hi! I'm using the 'composer' tag, but it doesn't always seem to work. As an example, I have a few tracks from https://musicbrainz.org/release/a334fe44-27a5-43dd-b1da-369ddad5cde7 where beet info doesn't even give me a composer tag, and beet info -l gives me an empty composer track. However on MB they all have a work and a composer linked to them.

ibmibmibm:
Did you upgrade your musicbrainzngs?

dosoe:
I have musicbrainzngs version 0.6.2 on an arch linux, I don't have the git version.

ibmibmibm:
You'll need git version of musicbrainzngs while 0.7 is not released yet.

dosoe:
I have an issue about the 'arranger' tag: In this pull request the 'arranger' tag is fetched as a tag of a recording. On MB there is another 'arranger' tag that is a tag of a work. Should we merge these two tags (fetch both in the tag 'arranger' and the corresponding 'arranger_sort') or should there be a 'recording_arranger' and a 'work_arranger'? The second one would be more rigorous but we might then end up with stuff like 'recording_arranger_sort' and 'recording_arranger_disambig' which starts to be ridiculously long. 
I have actually never encountered an occurence of 'arranger' as a recording tag, but I'm specialized on classical music, where 'arranger' is always related to a work, maybe other styles have it (I would like to check it up or link to examples, but MB is currently down).

dosoe:
Also, usually, for operas, there is another term for lyricist (librettist) which has his separate term in MB. Should I create a new tag for that or should I just fetch him in the lyricist tag?

sampsyo:
Some of these questions overlap with MusicBrainz style questions—maybe it would be useful to get input from #musicbrainz on IRC?

dosoe:
Ok after some discussion on MB IRC channel: work-arrangers are mostly in the context of classical music, when a specific arrangement is played by different people (as Fresco said:  However, since classicial music has a different tradition with regards to several "groups" performing the same arrangement, it makes more sense to make a new Work for these, so we can be sure that we're saying "these people are actually really playing the same arrangement, and not just the same Work that happens to have been arranged by the same )
recording-arrangers are more common in 'modern' music where one group makes his version of a popular piece/song. 
The distinction between lyricist and librettist is that librettist is basically reserved to stage-works (opera, musical and oratorio-like works) and lyricist gets the rest. It can get confusing because inside a musical there can be songs that are not written by the librettist of the musical, they have their own... lyricist. 

It seems pretty clear to anyone that work-arranger and recording arranger can be merged if necessary (maybe we can put them both into the 'arranger' tag since he gets written on the file but also keep separate tags for them if it's not too cumbersome). 

For librettist and lyricist there seems to be less consensus. I will continue to ask.

eziobandara:
> MusicBrainz provids composer, lyricist and arranger infomations related
> to individual recordings. This commit adds query parameters to fetch them, and
> write down to media files.
> 
> Tagging mapping is implemented according MusicBrainz Picard's data:
> https://picard.musicbrainz.org/docs/mappings/
> 
> Signed-off-by: Shen-Ta Hsieh [ibmibmibm.tw@gmail.com](mailto:ibmibmibm.tw@gmail.com)

I've no programming knowledge. Can you please help me with adding lyricist to my mp3s?

dosoe:
Hi! Just check out https://beets.readthedocs.io/en/stable/guides/main.html .

Issue #2253, opened 2016-11-02
acousticbrainz: Submit analyses upstream to AB


Freso:
Originally planned for being part of https://github.com/beetbox/beets/issues/1629 (which was resolved with https://github.com/beetbox/beets/pull/1787), the feature for the "acousticbrainz" plugin to run the analysis locally and submit results back to AcousticBrainz never made it in. It would still be a neat thing to have though. :)

sampsyo:
Yes; thanks for opening a reminder ticket!

inytar:
I'd like to pick this u if possible.

sampsyo:
Wonderful! Let's do it. This task has two parts: running the analysis program to get data and submitting the results to the web service API. Both start with information gathering: you'll want to look into the details for how to go both things and discuss your findings here. Sound good?

inytar:
Sound perfect! Here's my research:
Analysis:
- An extractor program is needed for the data analysis (http://acousticbrainz.org/download bottom left).
- AB would prefer if you use their static build of the extractor (http://acousticbrainz.org/faq second question).
- The extractor only works on: mp3 ogg oga flac mp4 m4a m4r m4b m4p aac wma asf mpc wv spx tta 3g2 aif aiff ape (http://acousticbrainz.org/contribute)
- The extractor will output a json file containing the `lowlevel` analysis, `metadata` and `rhythm`.

Uploading:
- Uploading should be done by using POST to the url `https://acousticbrainz.org/api/v1/<MBID>/low-level`.
- MBID is called recording id in different places, but all code uses the `musicbrainz_trackid`.
- The output file from the extractor and the sha1sum of the extractor program set to `metadata: version: essentia_build_sha` are send to the server (if you do not add the sha1sum the server responds with the error `{  "message": "Key 'metadata : version : essentia_build_sha' was not found in submitted data."}`)
- Content type header should be set to `application\json` else the server will respond with the error: `{  "message": "Cannot parse JSON document: No JSON object could be decoded"}`.

sampsyo:
Really awesome; thank you for looking into it! This all checks out, and it sounds like it should make the implementation relatively straightforward. We can just run the extractor agains the files and use `requests` to hit the AB API.

The only tricky part is that we'll need to compute the hash of the extractor executable, which we can do in a way similar to this: https://github.com/MTG/acousticbrainz-client/blob/c135a557c61984effcdb9d7a0c2686a64829b217/abz/config.py#L115-L118

Not too hard!

[...]

PR #2340, opened 2016-12-23
Implement the basic AcousticBrainz Submit plugin

inytar:
The basic implementation of the AcousticBrainz Submit plugin as mentioned in #2253. Missing documentation and tests. Especially on the test I'd love knowing what you would like to see.

### DIFF
--- a/beets/mediafile.py
+++ b/beets/mediafile.py
@@ -761,7 +761,6 @@ class MP3PeopleStorageStyle(MP3StorageStyle):
 
     def store(self, mutagen_file, value):
         frames = mutagen_file.tags.getall(self.key)
-        print(frames)
 
         # Try modifying in place.
         found = False
@@ -778,7 +777,6 @@ class MP3PeopleStorageStyle(MP3StorageStyle):
                 encoding=mutagen.id3.Encoding.UTF8,
                 people=[[self.involvement, value]]
             )
-            print(frame)
             mutagen_file.tags.add(frame)
 
     def fetch(self, mutagen_file):


sampsyo:
This looks awesome! I'd love to give it a try.

Something seems to have gone wrong with the PR view, and I'm not really sure why. First of all, GitHub seems to say that *I* authored this commit, which is clearly not true—do you know why it's doing that? And there are some strange, spurious changes to the changelog that don't seem to be part of this change—perhaps because this branched off of the wrong commit?

Yeah, docs and tests would be very useful here. The docs should be pretty sraightforward: they just need to state the dependency and tell you how to invoke the plugin-provided command. For the tests, maybe the right thing to do would be to mock the `command_output` function to show that `_get_analysis` invokes the command with the right arguments. Testing the interaction with the web service seems less essential.

beetsplug/absubmit.py
```diff
@@ -0,0 +1,159 @@
...
+        super(AcousticBrainzSubmitPlugin, self).__init__()
+
+        self.config.add({'extractor': u''})
+
+        self.extractor = self.config['extractor'].as_str()
+        if self.extractor:
+            self.extractor = util.normpath(self.extractor)
+            # Expicit path to extractor
+            if not os.path.isfile(self.extractor):
+                raise FatalABSubmitError(
```
sampsyo: This should probably be a `beets.ui.UserError`, right? That leads to a nicely-formatted user-visible error message.

beetsplug/absubmit.py
```diff
@@ -0,0 +1,159 @@
...
+                    u'extractor command does not exist: {0}'.
+                    format(self.extractor)
+                )
+        else:
+            # Implicit path to extractor, search for it in path
+            # TODO how to check for on Windows?
```
sampsyo: Does anything in particular need to be different on Windows?

beetsplug/absubmit.py
```diff
@@ -0,0 +1,159 @@
...
+            self.extractor = 'streaming_extractor_music'
+            try:
+                call([self.extractor])
+            except OSError:
+                raise FatalABSubmitError(
```
sampsyo: (As above.)

beetsplug/absubmit.py
```diff
@@ -0,0 +1,159 @@
...
+                    u'no extractor command found: install "{0}"'.
+                    format(self.extractor)
+                )
+            except AnalysisABSubmitError:
+                # Extractor found, will exit with an error if not called with
+                # the correct amount of arguments.
+                pass
+            # Get the executable needed to calculate the sha1 hash.
+            self.extractor = distutils.spawn.find_executable(self.extractor)
```
sampsyo: Is this `distutils` call necessary? It's usually fine just to use the the binary name by itself.

beetsplug/absubmit.py
```diff
@@ -0,0 +1,159 @@
...
+            help=u'calculate and submit AcousticBrainz analysis'
+        )
+        cmd.func = self.command
+        return [cmd]
+
+    def command(self, lib, opts, args):
+        # Get items from arguments
+        items = lib.items(ui.decargs(args))
+        # Get no_submit option.
+        # TODO get a should submit option from the command line.
```
sampsyo: Ah, so the plan is to just analyze unless the user specifies `-s` or somesuch?

beetsplug/absubmit.py
```diff
@@ -0,0 +1,159 @@
...
+        tmp_file, filename = tempfile.mkstemp(suffix='.json')
+        try:
+            # Close the file, so the extractor can overwrite it.
+            call([self.extractor, util.syspath(item.path), filename])
+            with open(filename) as tmp_file:
+                analysis = json.loads(tmp_file.read())
+            # Calculate extractor hash.
+            m = hashlib.sha1()
+            with open(self.extractor, 'rb') as extractor:
+                m.update(extractor.read())
```
sampsyo: It seems like we could do this once, at startup, and save the hash instead of recalculating it every time we execute the analysis.

beetsplug/absubmit.py
```diff
@@ -0,0 +1,159 @@
...
+        if item['format'].lower() not in self.supported_formats:
+            self._log.info('Not analysing {}, file not in '
+                           'supported format.', item)
+            return None
+
+        # Temporary file to save extractor output to.
+        tmp_file, filename = tempfile.mkstemp(suffix='.json')
+        try:
+            # Close the file, so the extractor can overwrite it.
+            call([self.extractor, util.syspath(item.path), filename])
```
sampsyo: Can we get the extractor tool to print the result JSON to stdout? If so, we can just use the result of `call` directly. This would avoid the need for the `tempfile` dance.

beetsplug/absubmit.py
```diff
@@ -0,0 +1,159 @@
...
+    def _submit_data(self, item, data):
+        mbid = item['mb_trackid']
+        headers = {'Content-Type': 'application/json'}
+        response = requests.post(self.base_url.format(mbid=mbid),
+                                 json=data, headers=headers)
+        # Test that request was successful and raise an error on failure.
+        if response.status_code != 200:
+            try:
+                message = response.json()['message']
+            except Exception as e:
```
sampsyo: To avoid unexpected effects, is usually better to list the specific set of exceptions here. Here, I think that would be `ValueError` for invalid JSON and `KeyError` for missing `message`.

beetsplug/absubmit.py
inytar: Personally I don't know, I do not have a Windows system to test..

beetsplug/absubmit.py
inytar: No, the reason of using this tempfile construction is that the extractor tool can only output to a given file and not to std out.

beetsplug/absubmit.py
inytar: Yes, this was originally planned, but I didn't find it very usefull, but forgot to remove the comment.

beetsplug/absubmit.py
inytar: I use the distutils to get the exact executable (especially in the implicit method) as we need that to calculate the executable hash.

beetsplug/absubmit.py
inytar: Yes, moving that shouldn't be a problem. Would you want it at init, or when the submit function is called?

beetsplug/absubmit.py
inytar: No, problem I can change the errors to UserError, this was a copy of how I understood the replaygain plugin did things.

beetsplug/absubmit.py
inytar: I'm also not very sure if we should raise an error here, or just skip uploading the file.

inytar:
Thanks for all the advice! I updated the pull request to hopefully fix the strange problem with authorization/ect. I'll also look into the comments, documentation and tests.
Some things I'm wondering:
- Do we want to raise an error if analysis or submitting a file fails?
- Do we want to save that a file has been analyzed and submitted? The analysis takes a long time and should not change if done multiple times and as far as I can see the the server code ignores any upload that is exactly the same as one already in the system.

beetsplug/absubmit.py
sampsyo: Got it; thanks for clarifying.

beetsplug/absubmit.py
sampsyo: Weird! That's too bad, but thanks for checking.

PR #2342, opened 2016-12-23
Implement the basic AcousticBrainz Submit plugin

inytar:
This is a new upload for #2340 as that had some strange problems with the pull request and trying to fix it closed the pull request.

Copy of the last comment in that pull request:

> Thanks for all the advice! I updated the pull request to hopefully fix the strange problem with authorization/ect. I'll also look into the comments, documentation and tests.
Some things I'm wondering:
>
> - Do we want to raise an error if analysis or submitting a file fails?
> - Do we want to save that a file has been analyzed and submitted? The analysis takes a long time and should not change if done multiple times and as far as I can see the the server code ignores any upload that is exactly the same as one already in the system.

### DIFF
--- a/beetsplug/absubmit.py
+++ b/beetsplug/absubmit.py
@@ -80,6 +80,12 @@ class AcousticBrainzSubmitPlugin(plugins.BeetsPlugin):
             # Get the executable needed to calculate the sha1 hash.
             self.extractor = distutils.spawn.find_executable(self.extractor)
 
+        # Calculate extractor hash.
+        self.extractor_sha = hashlib.sha1()
+        with open(self.extractor, 'rb') as extractor:
+            self.extractor_sha.update(extractor.read())
+        self.extractor_sha = self.extractor_sha.hexdigest()
+
     supported_formats = {'mp3', 'ogg', 'oga', 'flac', 'mp4', 'm4a', 'm4r',
                          'm4b', 'm4p', 'aac', 'wma', 'asf', 'mpc', 'wv',
                          'spx', 'tta', '3g2', 'aif', 'aiff', 'ape'}
@@ -124,13 +130,9 @@ class AcousticBrainzSubmitPlugin(plugins.BeetsPlugin):
             call([self.extractor, util.syspath(item.path), filename])
             with open(filename) as tmp_file:
                 analysis = json.loads(tmp_file.read())
-            # Calculate extractor hash.
-            m = hashlib.sha1()
-            with open(self.extractor, 'rb') as extractor:
-                m.update(extractor.read())
             # Add the hash to the output.
             analysis['metadata']['version']['essentia_build_sha'] = \
-                m.hexdigest()
+                self.extractor_sha
             return analysis
         finally:
             try:
--- a/beetsplug/absubmit.py
+++ b/beetsplug/absubmit.py
@@ -1,0 +1,159 @@
+"""Calculate acoustic information and submit to AcousticBrainz.
+"""
+
+from __future__ import division, absolute_import, print_function
+
+import hashlib
+import json
+import os
+import subprocess
+import tempfile
+
+import distutils
+import requests
+
+from beets import plugins
+from beets import util
+from beets import ui
+
+
+class ABSubmitError(Exception):
+    """Base exception for all excpetions this plugin can raise."""
+
+
+class FatalABSubmitError(ABSubmitError):
+    """Raised if the plugin is not able to start."""
+
+
+class AnalysisABSubmitError(ABSubmitError):
+    """Raised if analysis of file fails."""
+
+
+class SubmitABSubmitError(ABSubmitError):
+    """Raised if submitting data fails."""
+
+
+def call(args):
+    """Execute the command and return its output.
+
+    Raise a AnalysisABSubmitError on failure.
+    """
+    try:
+        return util.command_output(args)
+    except subprocess.CalledProcessError as e:
+        raise AnalysisABSubmitError(
+            u"{0} exited with status {1}".format(args[0], e.returncode)
+        )
+
+
+class AcousticBrainzSubmitPlugin(plugins.BeetsPlugin):
+
+    def __init__(self):
+        super(AcousticBrainzSubmitPlugin, self).__init__()
+
+        self.config.add({'extractor': u''})
+
+        self.extractor = self.config['extractor'].as_str()
+        if self.extractor:
+            self.extractor = util.normpath(self.extractor)
+            # Expicit path to extractor
+            if not os.path.isfile(self.extractor):
+                raise FatalABSubmitError(
+                    u'extractor command does not exist: {0}'.
+                    format(self.extractor)
+                )
+        else:
+            # Implicit path to extractor, search for it in path
+            # TODO how to check for on Windows?
+            self.extractor = 'streaming_extractor_music'
+            try:
+                call([self.extractor])
+            except OSError:
+                raise FatalABSubmitError(
+                    u'no extractor command found: install "{0}"'.
+                    format(self.extractor)
+                )
+            except AnalysisABSubmitError:
+                # Extractor found, will exit with an error if not called with
+                # the correct amount of arguments.
+                pass
+            # Get the executable needed to calculate the sha1 hash.
+            self.extractor = distutils.spawn.find_executable(self.extractor)
+
+    supported_formats = {'mp3', 'ogg', 'oga', 'flac', 'mp4', 'm4a', 'm4r',
+                         'm4b', 'm4p', 'aac', 'wma', 'asf', 'mpc', 'wv',
+                         'spx', 'tta', '3g2', 'aif', 'aiff', 'ape'}
+
+    base_url = 'https://acousticbrainz.org/api/v1/{mbid}/low-level'
+
+    def commands(self):
+        cmd = ui.Subcommand(
+            'absubmit',
+            help=u'calculate and submit AcousticBrainz analysis'
+        )
+        cmd.func = self.command
+        return [cmd]
+
+    def command(self, lib, opts, args):
+        # Get items from arguments
+        items = lib.items(ui.decargs(args))
+        # Get no_submit option.
+        # TODO get a should submit option from the command line.
+        for item in items:
+            analysis = self._get_analysis(item)
+            if analysis:
+                self._submit_data(item, analysis)
+
+    def _get_analysis(self, item):
+        mbid = item['mb_trackid']
+        # If file has no mbid skip it.
+        if not mbid:
+            self._log.info('Not analysing {}, missing '
+                           'musicbrainz track id.', item)
+            return None
+        # If file format is not supported skip it.
+        if item['format'].lower() not in self.supported_formats:
+            self._log.info('Not analysing {}, file not in '
+                           'supported format.', item)
+            return None
+
+        # Temporary file to save extractor output to.
+        tmp_file, filename = tempfile.mkstemp(suffix='.json')
+        try:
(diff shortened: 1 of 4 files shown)


jrobeson:
this failure here https://travis-ci.org/beetbox/beets/jobs/186448711  is the same I'm seeing in my PR. something has changed in either mutagen or the underlying build env.

jrobeson:
it seems to be mutagen. I filed https://github.com/beetbox/beets/issues/2343

sampsyo:
Thanks for opening the new PR! This one seems to have cleared up the weird problems.

Some answers to your questions from the other thread:

> Do we want to raise an error if analysis or submitting a file fails?

I think the appropriate thing to do is to use `self._log.error` to report the error, but no need to stop the whole process. This is usually nice because it means that if things go wrong for one specific file, or if the network connection drops out intermittently, progress isn't lost.

> Do we want to save that a file has been analyzed and submitted? The analysis takes a long time and should not change if done multiple times and as far as I can see the the server code ignores any upload that is exactly the same as one already in the system.

Good idea! If you don't mind, let's add this in a subsequent PR—we can merge the first version and then add this "don't submit twice" feature. It should be pretty easy—we can use an `abz_submitted` flexible attribute or somesuch, or even store the full analysis results—but we can get the basic functionality working first.

inytar:
I've updated this pull request, hopefully worked all the comments from the last pull request and added documentation.

sampsyo:
Wonderful; thank you! :sparkles:

It looks like Travis is reminding us that the new file needs the standard header, complete with license text and the `coding` comment at the top.

beetsplug/absubmit.py
```diff
@@ -0,0 +1,159 @@
...
+            self.extractor = util.normpath(self.extractor)
+            # Expicit path to extractor
+            if not os.path.isfile(self.extractor):
+                raise ui.UserError(
+                    u'Extractor command does not exist: {0}.'.
+                    format(self.extractor)
+                )
+        else:
+            # Implicit path to extractor, search for it in path
+            # TODO how to check for on Windows?
```
sampsyo: Is this comment still necessary? Any more questions about Windows?

sampsyo:
Or! There's some argument to be made here that we should just add this functionality to the `acousticbrainz` plugin, which has the same requirements (i.e., just requests). That would be neat and tidy!

docs/plugins/absubmit.rst
```diff
@@ -0,0 +1,31 @@
+AcousticBrainz Submit Plugin
+============================
+
+The `absubmit` plugin uses the `streaming_extractor_music`_ program to analyse an audio file and calculate different acoustic properties of the audio, the plugin then uploads this metadata to the AcousticBrainz server. The plugin does this when calling the ``beet absumbit [QUERY]`` command or on importing if the `auto` configuration options is set to ``yes``.
+
+Installation
+------------
+
+The `absubmit` plugin requires the the `streaming_extractor_music`_ program to run. Its source can be found on `github`_, and while it is possible to compile the extractor from source, AcousticBrainz would prefer if you used thier binary (see the AcousticBrainz `FAQ`_).
```
sampsyo: Capitalize "GitHub"

docs/plugins/absubmit.rst
```diff
@@ -0,0 +1,31 @@
+AcousticBrainz Submit Plugin
+============================
+
+The `absubmit` plugin uses the `streaming_extractor_music`_ program to analyse an audio file and calculate different acoustic properties of the audio, the plugin then uploads this metadata to the AcousticBrainz server. The plugin does this when calling the ``beet absumbit [QUERY]`` command or on importing if the `auto` configuration options is set to ``yes``.
```
sampsyo: The first comma here should actually end the sentence (and begin a new one).

> …acoustic properties of the audio. The plugin then uploads…

docs/plugins/absubmit.rst
```diff
@@ -0,0 +1,31 @@
...
+Installation
+------------
+
+The `absubmit` plugin requires the the `streaming_extractor_music`_ program to run. Its source can be found on `github`_, and while it is possible to compile the extractor from source, AcousticBrainz would prefer if you used thier binary (see the AcousticBrainz `FAQ`_).
+
+The `absubmit` also plugin requires `requests`_, which you can install using `pip_` by typing:
+
+    pip install requests
+
+After installing both the extractor binary and request you can enable the plgin ``absubmit`` in your configuration (see :ref:`using-plugins`).
```
sampsyo: plgin -> plugin

SusannaMaria:
Have added force option and check if acousticbrainz tag already set, similar to the acousticbrainz plugin pull request from myself
https://github.com/SusannaMaria/beets/commit/c474142fb4033851f44ce2cc8d425f469f378096

SusannaMaria:
There is some problem with removing the temp-json file in windows10.

       File "C:\Python27\lib\site-packages\beets-1.4.3-py2.7.egg\beetsplug\absubmit.py", line 149, in _get_analysis
    os.remove(filename)
       WindowsError: [Error 32] Der Prozess kann nicht auf die Datei zugreifen, da sie von einem anderen Prozess verwendet wird: 'c:\\users\\susanna\\appdata\\local\\temp\\tmpem70wo.json'

Sorry the the german error text, it says the file cannot deleted because ifs used from a different process.

Will check with https://docs.python.org/3/library/tempfile.html
it should solve this issue, maybe its completely windows related.

docs/plugins/absubmit.rst
```diff
@@ -0,0 +1,31 @@
+AcousticBrainz Submit Plugin
```
Kraymer: options is => option is

docs/plugins/absubmit.rst
Kraymer: the the => the

docs/plugins/absubmit.rst
Kraymer: and request => and requests

docs/plugins/absubmit.rst
```diff
@@ -0,0 +1,31 @@
...
+The `absubmit` also plugin requires `requests`_, which you can install using `pip_` by typing:
+
+    pip install requests
+
+After installing both the extractor binary and request you can enable the plgin ``absubmit`` in your configuration (see :ref:`using-plugins`).
+
+Configuration
+-------------
+
+To configute the plugin, make a ``absubmit:`` section in your configuration file. The available options are:
```
Kraymer: configute => configure

sampsyo:
Thanks, @SusannaMaria—perhaps we're doing something wrong when closing the new temporary file?

SusannaMaria:
Have to check tomorrow, just finished some changes because of force-option on @inytar fork (made a pull request on his fork) https://github.com/inytar/beets/pull/1/commits/b8fa8dcc5ea59e34063f36beacca2c957638b59c

inytar:
Sorry for the long wait, was away for a couple of days. I've fixed the above mentioned comments.
@sampsyo would you like to submit the plugin as is, or prefer it being merged with the acousticbrainz plugin? The reasons I originally did not do this are: this plugin needs a new dependency that is not needed for the original acousticbrainz plugin; this plugin does not save anything to file, but just upoads it to acousticbrainz. 
@SusannaMaria where you albe to fix the temp file problem in your branch?

docs/plugins/absubmit.rst
```diff
@@ -0,0 +1,31 @@
+AcousticBrainz Submit Plugin
+============================
+
+The `absubmit` plugin uses the `streaming_extractor_music`_ program to analyse an audio file and calculate different acoustic properties of the audio. The plugin then uploads this metadata to the AcousticBrainz server. The plugin does this when calling the ``beet absumbit [QUERY]`` command or on importing if the `auto` configuration option is set to ``yes``.
+
+Installation
+------------
+
+The `absubmit` plugin requires the `streaming_extractor_music`_ program to run. Its source can be found on `GitHub`_, and while it is possible to compile the extractor from source, AcousticBrainz would prefer if you used thier binary (see the AcousticBrainz `FAQ`_).
```
sampsyo: thier -> their

sampsyo:
No problem at all! There's no rush here. :smiley:

Sure—on further reflection, I can see the argument for keeping these as separate plugins. For the other one, the documentation doesn't even need to mention that you need to install the analysis program.

It looks like Travis is still complaining about the coding line at the top of the file—and indeed, the new Python file needs the entire standard header comment. Other than that, I think this is ready to merge.

sampsyo:
This is all merged up! :sparkles: Thank you for all your effort on this, @inytar.

sampsyo:
@tux-00 Sure; it's a good point to mention that music needs MBIDs for submission. Would you mind adding that to the docs with a PR? It's just a matter of hitting the "edit" button on the appropriate page on GitHub.

And we clearly shouldn't say "we recommend Picard" because beets of course does this perfectly well. Under ordinary circumstances, users' music will have already had MBIDs added by the import process.

tux-00:
No problem but just to make sure that I understand : 
- What plugin generates the MBID of a track ? absubmit ?

I'm a little bit confused with the names of the beets MusicBrainz plugins. For example the plugin "MusicBrainz Submit Plugin" does not seem to submit anything. It's possible that I don't understand something, I'm new with MusicBrainz !

sampsyo:
The importer (with no plugin at all) adds track MBIDs.

And that `mbsumbit` plugin generates text that you can paste into a submission form. That's the best we can do, unfortunately, because MusicBrainz doesn't offer a programmatic API for submission.

And no worries at all! I'm happy to answer questions, even if they seem basic.

tux-00:
When trying to run beets from sources, if the plugin doesn't find the the streaming_extractor_music binary beets raises this error : 
```
  File "/usr/bin/beet", line 9, in <module>
    load_entry_point('beets==1.4.3', 'console_scripts', 'beet')()
  File "/home/tux/git/beets/beets/ui/__init__.py", line 1209, in main
    _raw_main(args)
  File "/home/tux/git/beets/beets/ui/__init__.py", line 1192, in _raw_main
  File "/home/tux/git/beets/beets/ui/__init__.py", line 1089, in _setup
    plugins = _load_plugins(config)
  File "/home/tux/git/beets/beets/ui/__init__.py", line 1075, in _load_plugins
    plugins.send("pluginload")
  File "/home/tux/git/beets/beets/plugins.py", line 451, in send
    for handler in event_handlers()[event]:
  File "/home/tux/git/beets/beets/plugins.py", line 434, in event_handlers
    for plugin in find_plugins():
  File "/home/tux/git/beets/beets/plugins.py", line 288, in find_plugins
    _instances[cls] = cls()
  File "/home/tux/git/beets/beetsplug/absubmit.py", line 84, in __init__
    self.extractor = distutils.spawn.find_executable(self.extractor)
AttributeError: 'module' object has no attribute 'spawn'
```

So if I understand, if the importer cannot find the musicbrainz track id I cannot use the acousticbrainz to tag my tracks.  
Is that right ?

sampsyo:
Hmm; that actually looks like an unrelated bug—I don't know why `distutils.spawn` can't be found, but this doesn't have to do with MBIDs.

sampsyo:
In fact, could you please open a separate bug with that traceback and full details about your setup? Looks like we need to investigate more closely.

tux-00:
@sampsyo yes I will ! :)

tux-00:
It's strange, I can't reproduce the distutils error.. Now I get a good error reporting : 
`error: Extractor command does not exist: /usr/bin/test.`

I remember that I have played with pip to install beets from sources, maybe something happened and reboot solve the problem. This is the only reason I've found !

sampsyo:
Weird! Well, let us know if it comes back.

PR #2345, opened 2016-12-24
Cleanup zero plugin

nathdwek:
The testing is proving to be a real PITA.

I expected to have to rewrite it since it was originally written based on the internals rather than the interface so the cleanup would have consequences. However, what I did not expect is that for some reason there seem to be some state carried from test to test despite `setUp` and `tearDown`. Here are my current thoughts:

* Minor gripe, but `add_item_fixture()` is not *that* practical, because you have to call `item.write()` immediately after if you want the tags to be written to the mediafile fixture. This is a bit annoying, especially when you are trying to test a plugin which is triggered by a write event.

* It seems to me that despite `unload_plugins()`, plugins which did register a listener do not get nuked because they remain registered. This is my best guess at why several instances of `ZeroPlugin` try to take care of a single write event.

For those reasons, the tests do not pass for now. I only rewrote the three or four first tests because it was kinda frustrating...

### DIFF
--- a/beetsplug/zero.py
+++ b/beetsplug/zero.py
@@ -16,22 +16,20 @@
 """ Clears tag fields in media files."""
 
 from __future__ import division, absolute_import, print_function
+import six
 
 import re
+
 from beets.plugins import BeetsPlugin
 from beets.mediafile import MediaFile
 from beets.importer import action
 from beets.util import confit
-import six
 
 __author__ = 'baobab@heresiarch.info'
 __version__ = '0.10'
 
 
 class ZeroPlugin(BeetsPlugin):
-
-    _instance = None
-
     def __init__(self):
         super(ZeroPlugin, self).__init__()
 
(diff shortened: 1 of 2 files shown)


SJoshBrown:
I struggled with tests failing due to state being carried over as well. I added a call to remove_mediafile_fixtures() that seemed to clear things up. But I'm sure that is not the optimal solution.

sampsyo:
That's quite strange—don't really see why the plugins are failing to get "unregistered." It seems like, if this were happening every time, we'd have much bigger problems. I'll take a look…

sampsyo:
Shockingly, it looks like you're absolutely right and listeners aren't removed at all. This is an artifact of a long-ago, misguided decision to put the listeners on the plugin class instead of the instance. What a terrible idea.

As a short-term fix, it looks like the de facto standard is to manually remove the listeners. For example, `test_edit.py` has:

```py
EditPlugin.listeners = None
```

in a `tearDown` method.

Ugh, the plugin management stuff is in general such a mess—it's long overdue for a "manager class" style of refactoring.

nathdwek:
Whoop whoop, looking good now even though it's a workaround. Okay for you to merge, @sampsyo ?

sampsyo:
Perfect! Thank you for making this plugin (and the tests) far more manageable. :sparkles: :rocket:

Issue #1948, opened 2016-04-18
Random sort option


strifeknot:
I'd like to be able to produce randomly sorted tracks from the Smart Playlist plugin (I'm not a big fan of using random tracks from within mpd - especially with mpd-sima) - I could run:
`beet list ... | sort -R > playlist.m3u`
daily off a systemd timer but was wondering if random+ might be considered as a native sorting option.

sampsyo:
Hey, yeah, this kind of thing would be good! We've discussed routes toward this functionality in #795—could you perhaps add your thoughts there?

Issue #750, opened 2014-05-06
Pagination support in dbcore


PierreRust:
The creation of this issue has been discussed in #736.
To implement pagination properly in the HTTP API, we need to add support for pagination in dbCore. Ideally the pagination mechanism should also be usable with the command-line interface.

Ideas about pagination have been discussed in #718. It seems were are basically 4 options : 
- the naive approach : using start/offset parameters in query. While this is trivial to implement it is definitely not very efficient.
- use a seek-able id in the query ( i.e. "gives me the next items starting with this one") : this is the approach [recommended by sqlite](http://www.sqlite.org/cvstrac/wiki?p=ScrollingCursor), we have to make a small prototype to make sure it's doable with complex queries.
- when performing a query, store temporarily the result and return a token to the client, which can be used to walk through the ResultSet. The difficult part here is to maintain, and discard at some point, the ResultSets
- last solution : only return id for complex queries. This looks nice but I'm afraid it might not play well with the current query mechanism, which fetches full rows anyway.

pjones:
This would be especially useful for the Smart Playlist and Random plug-ins.  Right now there's no way to limit the number of tracks generated by the Smart Playlist plug-in.

jonthomas83:
Sorry guys, I'm not sure if I'm barking up the wrong tree here, and please tell me to get lost, but from a UI perspective, pagination or pager navigation isn't the most user friendly interface to have to use - we've done a lot of usability testing and found that they're not as straightforward as we'd like to think. You'd be surprised at the types of feedback, which I can share if you need me to.

A lot of the stuff mentioned here goes right over my head, but after reading https://github.com/sampsyo/beets/issues/738, it seems like pagers/paged navigation methods will play a part in the proposed UI implementations. Can we reconsider their use and look at alternative options before making a decision - like I said, in terms of UI. Many thanks.

sampsyo:
Yes! I absolutely agree that it's not a foregone conclusion that the interface should paginate. But it's still useful for the API to admit pagination for efficiency reasons—independent of the interface.

jonthomas83:
That's good news, as long as we're on the same "page" ;-)

FWIW, our guys are still developing paging on the API too, makes sense.

semohr:
I’m not sure if this issue is still relevant, but I just came across it.
[...]

PR #2350, opened 2016-12-28
play: add --random option

diomekes:
This adds feature request #795 and part of #2305.

It's the same as #2304, but I closed that one due to updates to the random plugin (and because I didn't branch my master fork to make changes, you live you learn).

As discussed in #2304, #795 and #1948, the best option would be to add a "random" query prefix, but that's too advanced for me (at least for right now) to develop, so I did this instead.

### DIFF
--- a/beetsplug/play.py
+++ b/beetsplug/play.py
@@ -25,6 +25,7 @@ from beets import ui
 from beets import util
 from os.path import relpath
 from tempfile import NamedTemporaryFile
+from beetsplug import random
 import subprocess
 
 # Indicate where arguments should be inserted into the command string.
@@ -81,6 +82,29 @@ class PlayPlugin(BeetsPlugin):
             action='store',
             help=u'add additional arguments to the command',
         )
+        play_command.parser.add_option(
+            u'-r', u'--random',
+            action='store_true',
+            help=u'play random item',
+        )
+        play_command.parser.add_option(
+            u'-n', u'--number',
+            action='store',
+            type="int",
+            help=u'with -r: number of random items to choose',
+            default=1,
+        )
+        play_command.parser.add_option(
+            u'-e', u'--equal-chance',
+            action='store_true',
+            help=u'with -r: each artist has the same chance',
+        )
+        play_command.parser.add_option(
+            u'-t', u'--time',
+            action='store',
+            type="float",
+            help=u'with -r: total time in minutes of random items to choose',
+        )
         play_command.func = self._play_command
         return [play_command]
 
@@ -95,9 +119,15 @@ class PlayPlugin(BeetsPlugin):
         # Perform search by album and add folders rather than tracks to
         # playlist.
         if opts.album:
-            selection = lib.albums(ui.decargs(args))
-            paths = []
 
+            # Send query to random plugin.
+            if opts.random:
+                selection = random.random_func(lib, opts, args,
+                                               print_list=False)
+            else:
+                selection = lib.albums(ui.decargs(args))
+
+            paths = []
             sort = lib.get_default_album_sort()
             for album in selection:
                 if use_folders:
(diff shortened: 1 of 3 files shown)


nathdwek:
First and foremost, thank you for your work and commitment to this feature, and welcome to beets!

I'd like to see this feature implemented very much, but to be honest I don't like it being done this way. I think we should be bite the bullet and implement randomness at the query level.

This would be useful for many other operations, right out of the box (the latest example being #2352 )! Furthermore, as I mentioned in #2304, I don't find the use of `with nostdout()` very elegant here, along with the three imports.

That being said, I perfectly understand that you would not take on such a project, and I am perfectly open to diverging opinions on this.

sampsyo:
Agreed—thank you for following up with this after #2322!

Let's explore what it would take to make this part of the query syntax. The tricky thing here is that a query like `beatles rand:5` or something would currently be parsed as an `AndQuery` of an ordinary `SubstringQuery` query and a new `RandomQuery` or somesuch. The "and" part doesn't make much sense for random sampling.

Instead, what we want is more like our current sort operators. They modify the whole query that they appear in, so `beatles year+` does *not* parse to an `AndQuery`. Maybe we need a generic mechanism for "query modifiers" like sorting?

diomekes:
I agree with the `nostdout()` being ugly, so I "fixed" it (I didn't see your comment on #2304).

If we are able to make into a query (I'm willing to help, I just don't have any experience), would we still have access to the --equal-chance and --time options provided by `random`?

I'm guessing it would involve multiple queries under the proposed `RandomQuery`.

sampsyo:
Yeah. We'd have to come up with good syntax for it. Here are some (probably bad) ideas:

* `rand.t:5` for 5 minutes
* `rand.e:5` for 5 objects with equal chance
* `randtime:5`
* or if you like ASCII, `5~` for 5 random objects and `5t~` or something for the time version

diomekes:
I think we do need a mechanism for modifiers like sorting (are there other types of modifiers already?).  Sorting requires a `field` and `+` or `-`.  Shuffling a query, instead of sorting, could require a `?`.  For example:

`ls beatles ?` would return a random beatles item (if 1 is the default, or something before `?` could be a requirement).
`ls beatles 10?` would return ten.
`ls beatles 0:30?` would return 30 minutes.

I have no idea how to implement this, I was just looking at `query.py` and `queryparse.py`, so I don't know if this is something a plugin can do or if it has to be in the core.

sampsyo:
Right! No, sorting is the only thing currently like this, so the query handling is currently hard-coded to only support sorting. We'd need to refactor things so that sorting, instead of being a special snowflake, is just a specific instance of a more general category.

The `?` syntax is definitely suggestive! And it's probably safe for most shells too—at least if they're configured to leave wildcards intact when they don't match any files.

diomekes:
After further thought, it seems like we need to do two things:

1. A query modifier that randomizes the query, maybe `?` as I suggested earlier.
2. Give beets the ability to output either a certain number of results or a list that stays within a time limit.

The first one seems like it could be a part of beets current sorting with `+` or `-`, excluding the `field` requirement.

The second one, as far as I can tell, isn't a part of beets yet, although it can be done with `random`.  I don't think a plugin should do it though, for the same reasons we're trying to de-plugin `random`'s functionality.  Unless, of course, it's a plugin that modifies the query syntax with a prefix, something like `beet ls 5:beatles`.

Has there been any discussion, as far as you know @sampsyo, of limiting beets' output?

sampsyo:
Good point! Randomizing and truncating could indeed be decoupled.

The closest thing here is #750, which we come up with in the context of the web interface. But the basic functionality is the same: limiting the number of results returned from a query.

jtpavlock:
@diomekes any interest in continuing this?

diomekes:
@jtpavlock Yes, but it's beyond my abilities.  I did try to figure it out way back when, but just couldn't find out how to do it the right way (making randomness a query instead of just baked into the "play" plugin).

jtpavlock:
I completely understand, it's not exactly what you signed up for when you opened this PR 😄 

I also appreciate that you've generated a worthwhile discussion to the future implementation of this feature. 

I'm going to close this since it seems we'd prefer the implementation to be a much bigger and different idea than what you originally intended/were able to work on. If you'd like to continue in the discussion or have any further ideas, I suggest moving to #795. And if you ever want to come back and attempt to tackle this, feel free to open another pull request!

PR #2352, opened 2016-12-28
Mosaic plugin

SusannaMaria:
It's really fun to learn python within beets. :-)
Some months ago I wrote a simple java-application which created a mosaic of my cover art from beet lib. I decided yesterday to implement a very first plugin which covers this feature.

Here is the result fetching my whole lib, took ~1 minute. I resized the result with irfanview because it is 50 MB.

mosaic

I will continue on the develop add documentation and meaningful configuration ... and code-documentation _*blush*_
Right now it's only tested on windows10 with python2.7


beetsplug/mosaic.py
```diff
@@ -0,0 +1,105 @@
...
+		return [cmd]
+
+	def _generate_montage(self, lib, albums, output_fn):
+		fullwidth=0;
+		fullheight=0;
+
+		covers = list();
+
+		for album in albums:
+			self._log.info(u'#{}#', album.artpath)
```
nathdwek: Here `album.artpath` is displayed before checking if it is not `None`

beetsplug/mosaic.py
```diff
@@ -0,0 +1,105 @@
...
+			
+			if not album.artpath:
+				continue
+
+			if not os.path.exists(album.artpath):
+				continue
+
+			covers.append(album.artpath)
```
nathdwek: Maybe this could be more elegantly written as

```python
if album.artpath and os.path.exists(album.artpath):
    #cover list and logging
else:
    #is some form of logging needed here?
```

beetsplug/mosaic.py
```diff
@@ -0,0 +1,105 @@
...
+	def __init__(self):
+		super(MosaicCoverArtPlugin, self).__init__()
+		self.config.add({'maxwidth': 300})
+		self.maxwidth = self.config['maxwidth'].get(int)
+
+ 	def commands(self):
+		cmd = ui.Subcommand('mosaic', help=u"create mosaic from coverart")
+
+		def func(lib, opts, args):
+			self._generate_montage(lib, lib.albums(ui.decargs(args)), u'mos.png')
```
nathdwek: Should the mosaic filename be chosen by the user, at least optionally? It makes sense from a ux point of view.
Also, this would make the code more sensible, as for now `_generate_montage` has the `output_fn` parameter which is always hardcoded to `u'mos.png'`. That made me scratch my head a little bit.

`embedart` for example could be an inspiration for this. Moreover, it  also does not let the user choose the extension (here I suppose you only generate `png` for convenience reasons), so you could use the exact same convention and ux.

nathdwek:
This looks awesome and very pretty! I'll give it a run on linux asap.

Looking forward to seeing the docs and finished ux! Do you think this plugin is fit for some testing? Is there some business logic to isolate?

beetsplug/mosaic.py
```diff
@@ -0,0 +1,105 @@
+# -*- coding: utf-8 -*-
+# This file is part of beets.
+# Copyright 2015-2016, Ohm Patel.
```
sampsyo: You might want to put your own name here. :smiley:

beetsplug/mosaic.py
```diff
@@ -0,0 +1,105 @@
...
+from beets.util.artresizer import ArtResizer
+from beets import config
+from beets import art
+from PIL import Image
+
+import math
+
+
+class MosaicCoverArtPlugin(BeetsPlugin):
+	col_size = 4
```
sampsyo: It looks like `col_size` might be unused?

beetsplug/mosaic.py
```diff
@@ -0,0 +1,105 @@
...
+		cmd = ui.Subcommand('mosaic', help=u"create mosaic from coverart")
+
+		def func(lib, opts, args):
+			self._generate_montage(lib, lib.albums(ui.decargs(args)), u'mos.png')
+		cmd.func = func
+		return [cmd]
+
+	def _generate_montage(self, lib, albums, output_fn):
+		fullwidth=0;
+		fullheight=0;
```
sampsyo: Python doesn't need semicolons. :duck:

nathdwek:
A mosaic
:1st_place_medal:

It took a bit of bulldozing, but still, congrats. One observation: it does not find the cover file for a several albums which in fact do have a cover art.

Some more minor comments:
* What to do if the number of album is prime?
* How to handle albums which are not square?

bearcatsandor:
Couldn't one use this with bpd or the play command to create wallpapers and statuses that change as you change things?

nathdwek:
BTW the code with the minimal changes for a working plugin, which I used to do the little tests is at nathdwek@d5f1d5bfb86b8d0c1a9250b03da79d4b0bda425b . @SusannaMaria you could maybe pull from there to get going?

SusannaMaria:
Thanks guys for the reply. I really want to follow the pep rules, and for sure I want to consider the issues related to ux.

SusannaMaria:
@nathdwek 
1. prime: My Idea was to have a squre mosaic and if it's not fiting because it is a prime just add a new row containing the tail of root calculation.
2. Right now I have no creative solution for this, in the beginning I would just crop to squre
3. The idea is to use the order which comes out of the query, do we have already a feature to randomize the query output?

nathdwek:
I think the two first points are perfectly reasonable, my idea would be sort of a bonus if anything.
As for the third point, we don't have exactly that yet, but it is absolutely the right approach! Shame on me for thinking about implementing this at the plugin level :disappointed: .

SusannaMaria:
Some ideas regarding config:
     mosaic:
             geometry: 100x100+3+3
             tile: square # or e.g. x3 3x
             label: $album
             background: ffffff
             result_filename: mosaic.png 
             show_mosaic: yes

SusannaMaria:
I added a blending of a picture to the whole mosaic
mos

in this example I used
tool

Maybe it's interesting to have something like a watermark in it. Maybe the github-user icon.

mried:
I've implemented the same thing (beside a lot of other things) some time ago in my arttools plugin with the `artcollage` command. You can find it here: https://github.com/mried/beetsplug. But don't expect to much: There's no watermark and has only a very simple configuration (output filename and tile size). I'm using it to create such a mosaic for every directory containing subdirectories. So I've got a mosaic for the whole library, for each genre and artist.

sampsyo:
Very cool! It looks like this is passing our style checker now too. Woohoo!

I think all this needs is some plugin documentation before it's mergeable.

sampsyo:
Cool! There's actually an easy way to handle command-line arguments that override defaults from the configuration file. Here are the relevant docs:
http://confuse.readthedocs.io/en/latest/#command-line-options

Calling something like `self.config.set_args(opts)` will overlay the configuration with values from the command line. Then you can just access the configuration as normal, with no conditional, to get the merged values.

nathdwek:
Is it possible that there are some miscalculation of the mosaic size? It seems that my bigger mosaics are systematically missing what seems like a last row?

SusannaMaria:
Added a creation of dummy cover if the artpath is  not given. I need to add a font therefor I moved the plugin into a subfolder and changed the manifest.in. 

Guess I need some days more to check for bugs and documentation...

sampsyo:
Hmm… including assets authored by other people can cause weird licensing headaches, especially for the people who package beets for OS package managers. Maybe we can make the plugin automatically download FreeSans when it first needs it? (Or use an installed font if one is available?)

SusannaMaria:
@sampsyo yes I understood, i have commented the font code out. Because I want to have a feature to label the cover however it's better to plan a solution which cause no issues. I will start with documentation and code comments tomorrow. Hope I get the bug with command-line fixed.

SusannaMaria:
That a weirdo bug.

       from PIL import Image

cause to fail all checks.

sampsyo:
OK, sounds good!

That's weird about the test failure. I don't see your plugin is getting imported during the tests (since there are no tests yet). If we do want to add tests, we'll of course have to instruct the test harness to install PIL/Pillow. But that shouldn't be a problem yet… I'll have to look a little closer. Sorry about that!

beetsplug/mosaic.py
```diff
@@ -0,0 +1,249 @@
...
+from __future__ import division, absolute_import, print_function
+
+import os.path
+
+from beets.plugins import BeetsPlugin
+from beets import ui
+
+from PIL import Image
+import math
+from parse import *
```
sampsyo: It's usually best to avoid "star imports," which make it difficult to tell where names come from.

beetsplug/mosaic.py
```diff
@@ -0,0 +1,249 @@
...
+                    'background'].get(str)
+            else:
+                background = opts.background
+
+            if self.config['geometry']:
+                geometry = opts.geometry or self.config[
+                    'geometry'].get(str)
+            else:
+                geometry = opts.background
+            albums = lib.albums(ui.decargs(args))
```
sampsyo: Since you've already used `self.config.set_args`, you don't need all these if/else blocks for fallbacks. For example, this suffices just fine:

```py
geometry = self.config['geometry'].get(str)
```

In fact, this is probably what you want:

```py
geometry = self.config['geometry'].as_str()
```

No need to test whether the value is present; just use it.

sampsyo: Of course, you don't even need to pass around stuff like `geometry` as a parameter—the code that needs it can just use `self.config['geometry'].as_str()` right there.

docs/plugins/mosaic.rst
```diff
@@ -0,0 +1,57 @@
+Mosaic Plugin
+=====================
+
+The ``mosaic`` plugin generates a montage of a mosiac from cover art.
+
+To use the ``mosaic`` plugin, first enable it in your configuration (see
+:ref:`using-plugins`). Then, install the `Pillow`_ library by typing::
+
+    pip install Pillow
```
sampsyo: Looks like this plugin also uses a library called `parse`, so you probably want to mention that here. (Or just use regular expressions instead.)

SusannaMaria:
Maybe you can give me some advise because of unit testing? Right now I have only one test implemented. which just generates a 2x2 mosaic with size 359x359. I compare the image size at the end.

Ideas
1. Load different Image types, should I store examples in the rsrc folder?
2. Robustness, different "crazy" options like negative geometry? (Have to implement a check for this)
3. A very big mosaic with 10.000 pieces?

SusannaMaria:
https://tympanus.net/Tutorials/ThumbnailGridExpandingPreview/
Thinking about going html5...as Extension of Web 🤔

sampsyo:
Very cool! I actually think a small test is probably fine in this case. If problems come up with strange parameters like very large numbers of albums, we can add tests for those too, but just getting the basic functionality covered is a great place to be for a new plugin.

The web thing is a neat idea too! Especially if there were still a way to flatten to a standalone image…

SusannaMaria:
Just added a second usecase and faced a problem with png-Watermark.
Coverage is at 88% 
beetsplug/mosaic/__init__.py     130     15    88%
@sampsyo thank you again for your great support and motivation. I will now refactor (again) and kick out the own directory in plugins. I'm using now a mechanism to download a ttf from google-font, will make it configurable.

Any change to put this plugin into master after QA and cleaning with static mosaic creation, or should I continue merging into the web-frontend as HTML5 with canvas?

sampsyo:
Very cool! Yes, I'd be happy to merge this into master whenever you think it's ready. Maybe we can hit the green button after your next round of refactoring?

SusannaMaria:
added option to download and use font: 
beet mosaic --geometry=150x150+1+1 -a 0.5 -w D:\temp\alpha.jpg -f https://github.com/google/fonts/raw/master/ofl/actor/Actor-Regular.ttf
or in config:
mosaic:
    font: https://github.com/google/fonts/raw/master/ofl/aladin/Aladin-Regular.ttf

sampsyo:
Nice. Would you mind mentioning the new font option in the documentation so people know how to use it?

Also, I’m not 100% sure what’s going on, but the GitHub pull request view is for some reason showing a bunch of unrelated changed. Maybe it’s a strange consequence of rebasing? I could dig a little deeper to find out what’s wrong, but maybe it’s clearer from your side.

SusannaMaria:
Because of the long break of 6 months I merged the master into the feature branch before I continued 3 days ago. Sorry because of the mess.

sampsyo:
Ah, no problem. It makes the diff display look weird, but I think it will still merge correctly on the command line.

SusannaMaria:
Okay that's it from my side for the mosaic plugin for now. Unfortunately I'm unable to understand the Problem with https://travis-ci.org/beetbox/beets/jobs/255377806
If I have fixed this problem, I really like to continue on the similarity plugin now. Guess it's best to use a conservative approach to store relations with Flexible Field Types. Lets see, but more or less I want to explore the best way to visualize the graph. HTML5 canvas or so

sampsyo:
Wow; that Travis failure is indeed really weird. I suspect it may be unrelated to your new plugin… I’ll look into it a bit more deeply. Thanks for all your effort on this! :sparkles:

SusannaMaria:
Maybe its the download of the font. I never asked the question, where I have to store the font permanently? 

            filename = fonturl[fonturl.rfind("/") + 1:]
            fontpath = os.path.join(os.path.dirname(__file__), filename)

so its stored at the same location of the plugin. Feeling guilty ...

SusannaMaria:
"The deceased were taken from the grave" ...
Come back to my music-lib and worked a little on tagging and cleaning up my music which I have bought since 2017. So I went back to the mosaic plugin which never made it.

I never was satisfied with the static generation of one file, but I have found pyqtgraph today, very neat lib.

grafik

Issue #2318, opened 2016-12-14
discogs: Match individual subtracks


welch:
### Problem
I'm trying to import this album:
https://www.discogs.com/Erich-Wolfgang-Korngold-Jerome-Moross-Franz-Waxman-Colin-Towns-featuring-The-Korngold-Quartet-Scott-/release/3647530

In the tracklist there are 0-length "index" items naming the work to which the next few tracks belong. These do not have track numbers. The tracks themselves are listed as "subtracks", and have the proper global track numbers (in the listing, and as ripped from my CD. track numbers and times agree between them). 

Beets does not see any of the track numbers that exist as "subtracks". It sees this album as containing a single track, #13, which it wants to renumber as #1 during import. 

Here's the output proposing the partial match for the album:
```
Correcting tags from:
    Jerome Moross - Classical Hollywood III
To:
    Erich Wolfgang Korngold / Jerome Moross / Franz Waxman / Colin Towns featuring The Korngold Quartet, Scott Woolley, The Mildred Pierce Orchestra - Classical Hollywood III
URL:
    https://www.discogs.com/Erich-Wolfgang-Korngold-Jerome-Moross-Franz-Waxman-Colin-Towns-featuring-The-Korngold-Quartet-Scott-/release/3647530
(Similarity: 36.6%) (unmatched tracks, artist, source, tracks) (Discogs, CD, 1992, US, Bay Cities)
CD: Quintet For Piano And Strings
 * American Suite For Orchestra (#13) -> American Suite For Orchestra (#1) (index)
Unmatched tracks (12):
 ! 01  (#1) (3:00)
 ! 02  (#2) (4:48)
 ! 03  (#3) (2:45)
 ! 04  (#4) (1:53)
 ! 05  (#5) (12:36)
 ! 06  (#6) (12:14)
 ! 07  (#7) (8:54)
 ! 08  (#8) (1:30)
 ! 09  (#9) (3:51)
 ! 10  (#10) (1:26)
 ! 11  (#11) (1:11)
 ! 12  (#12) (5:27)
```

(that item 13 it wants to renumber to 1 should really be 13)

### Setup

* OS: Mac OS 10.11.5
* Python version: 2.7.12 
* beets version:  1.3.18
* Turning off plugins made problem go away (yes/no): 

My configuration (output of `beet config`) is:

```yaml
duplicates:
    tiebreak:
        items: [bitrate, filesize]
    format: $artist - $album - $title .$format-$bitrate $path
    count: no
[...]

PR #2355, opened 2016-12-29
discogs: revise subtrack merging based on subindex

diego-plan9:
An attempt at #2318, by trying to make use of the subtrack positions (which in turn is based on the standard discogs positions at [guideline 12.2.9](https://www.discogs.com/help/doc/submission-guidelines-release-trk)) to identify in which cases subtracks should be merged into a single one, and in which cases should be kept as individual tracks.

As mentioned on https://github.com/beetbox/beets/issues/2318#issuecomment-269357217, after another look at the discogs guidelines and at a number of API responses, I can't seem to find another, more robust way of trying to guess which case should be applied for an individual release. While I believe this fix works properly on the examples mentioned on the issue (both the two release that have subtracks that should be kept as individual tracks, and the [megamix](https://www.discogs.com/Various-Defqon1-Weekend-Festival/release/483738) where subtracks should be merged) and it is a slight improvement, it is still not perfect -  in particular, it fails to correctly identify one of the examples on the guidelines:
https://www.discogs.com/King-Crimson-Lizard/release/1156598

as `get_track_index()` is unable to interpret the positions on side B (`Ba`, `Bb`, etc) as "logical divisions of track B", hence considering the album to have 4+4 tracks instead of 4+1. 

I'm quite certain there will be a number of other situations where the detection will not be accurate, in particular where non official positions are used or combinations of several mediums, annotations, etc - which leads me to believe that aiming for 100% correct detection is not possible using only this approach, unless I'm missing something! Perhaps instead of trying to further adjust the (already a bit convoluted) code to tune it for other cases, it would make sense to try to allow the user to specify somehow which "strategy" to apply when subtracks are found, bypassing the guess made by the plugin? If so - any preferences on the best way interface-wise to provide it (extra prompt/command line flag/configuration setting/etc)?

### DIFF
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -367,7 +367,7 @@ class DiscogsPlugin(BeetsPlugin):
                     tracklist[-1]['position'] = position
                 else:
                     # Promote the subtracks to real tracks, discarding the
-                    # index track, assumeing the subtracks are physical tracks.
+                    # index track, assuming the subtracks are physical tracks.
                     index_track = tracklist.pop()
                     # Fix artists when they are specified on the index track.
                     if index_track.get('artists'):
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -354,15 +354,27 @@ class DiscogsPlugin(BeetsPlugin):
             """Modify `tracklist` in place, merging a list of `subtracks` into
             a single track into `tracklist`."""
             # Calculate position based on first subtrack, without subindex.
-            idx, medium_idx, _ = self.get_track_index(subtracks[0]['position'])
+            idx, medium_idx, sub_idx = \
+                self.get_track_index(subtracks[0]['position'])
             position = '%s%s' % (idx or '', medium_idx or '')
 
-            if len(tracklist) > 1 and not tracklist[-1]['position']:
-                # Assume the previous index track contains the track title, and
-                # "convert" it to a real track. The only exception is if the
-                # index track is the only one on the tracklist, as it probably
-                # is a medium title.
-                tracklist[-1]['position'] = position
+            if tracklist and not tracklist[-1]['position']:
+                # Assume the previous index track contains the track title.
+                if sub_idx:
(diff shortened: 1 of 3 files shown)


beetsplug/discogs.py
```diff
@@ -354,15 +354,27 @@ def add_merged_subtracks(tracklist, subtracks):
...
+            if tracklist and not tracklist[-1]['position']:
+                # Assume the previous index track contains the track title.
+                if sub_idx:
+                    # "Convert" the track title to a real track, discarding the
+                    # subtracks assuming they are logical divisions of a
+                    # physical track (12.2.9 Subtracks).
+                    tracklist[-1]['position'] = position
+                else:
+                    # Promote the subtracks to real tracks, discarding the
+                    # index track, assumeing the subtracks are physical tracks.
```
diego-plan9: assumeing -> assuming will be corrected on the next commit :hammer:

sampsyo:
OK! In summary, it looks like the policy here is that the *existence* of a subindex dictates whether a given subtrack should be considered an independent track. That seems eminently reasonable to me.

I also totally agree with you that there is likely not to be any perfect solution here, and the best we can do is apply a reasonable policy that works most of the time. Later, if an exception to the policy makes sense, I think the right thing to do will be to add a configuration option. That will avoid bothering most users while letting advanced users control things precisely.

As usual, this looks great! Ready to merge (with changelog entry) as soon as the typo's fixed. :smiley:

diego-plan9:
Thanks for the prompt review and comments, as usual!

> OK! In summary, it looks like the policy here is that the existence of a subindex dictates whether a given subtrack should be considered an independent track. That seems eminently reasonable to me.

Yes, that would be the summary - with the remark that "*the existence of a subindex*" is determined on the plugin, and is not bullet-proof, in particular for non-standard discogs positions (and some edge cases with the standard ones).

+1 as well for waiting on the (likely) next round of discogs-related issues for introducing the configuration option - makes much sense!

PR #2358, opened 2016-12-30
Replace Python 3.4 with 3.6 on Appveyor

### LINKED ISSUE #2344: Remove python 3.4 when we add Python 3.6 to Appveyor
Appveyor takes forever to build since it only uses 1 worker at a time.  3.4 doesn't really offer us anything on Windows.

i opened a ticket https://github.com/appveyor/ci/issues/1237 to track python 3.6 implementation.

jrobeson:
Python 3.4 is quite old and offers little value on Windows.

Closes #2344

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,10 +12,10 @@ environment:
     matrix:
         - PYTHON: C:\Python27
           TOX_ENV: py27-test
-        - PYTHON: C:\Python34
-          TOX_ENV: py34-test
         - PYTHON: C:\Python35
           TOX_ENV: py35-test
+        - PYTHON: C:\Python36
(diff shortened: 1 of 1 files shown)


jrobeson:
looks like we've been bitten by this bug http://bugs.python.org/issue29097

sampsyo:
Wow; that's a pretty crazy bug. 😕 So I suppose we have to wait for 3.6.1, or skip those tests for now? (Occasionally, CI can seem like more of a headache than it's worth.)

sampsyo:
Or we can find a way to avoid running `datetime.fromtimestamp(0)` in `_to_epoch_time` by using some other method to subtract off the system epoch?

PR #2359, opened 2016-12-30
Temporarily allow empty checksum for unrar on appveyor

jrobeson:
Obviously it'd be best if we didn't have to do it, but I doubt we can
wait forever for the packagers to add it.


sampsyo:
Thanks for fixing that. It is indeed annoying.

Issue #2008, opened 2016-05-21
play: Add importer prompt to play new music


jnko:
When importing into my library sometimes a song has completely wrong tags and the candidates comming from musicbrainz are not always clear.

In this situation I wish I could just play this song to manually verify which it is.

Here's an example:

```
/home/injk/MusicLib/Compilations/The Voice Vol. 3 (1 items)
Finding tags for album "Eri esittäjiä - The Voice Vol. 3".
Candidates:
1. Various Artists - The Dome, Volume 46 (23.7%) (missing tracks, album, label, ...) (2xCD, 2008, DE, EMI Music Catalogue Marketing)
2. 3 Doors Down - It’s Not My Time (21.7%) (id, artist, album, ...) (CD, 2007, US, Universal Republic Records)
3. Jakob Ullmann - Voice, Books and FIRE 3 (21.5%) (id, artist, tracks, ...) (CD, 2008, DE, Edition RZ)
4. Nikhil Banerjee - The Genius Of Pandit Nikhil Banerjee - vol 3 (20.4%) (id, artist, album, ...) (CD, 1996, DE, Chhanda Dhara)
5. Jasper Carrott - Classic Carrott Vol 3 (17.5%) (id, artist, album, ...) (CD, 2004, GB, Sound Entertainment Limited)
6. Various Artists - Radio Contact Hits 3: The Daewoo Premium Mix (16.2%) (id, album, tracks, ...) (CD, 1998, BE, Arcade)
7. 3 Doors Down - 3 Doors Down (14.7%) (missing tracks, id, artist, ...) (CD, 2008, GB, Universal Republic Records)

# selection (default 1), Skip, Use as-is, as Tracks, Group albums,
Enter search, enter Id, aBort, eDit, edit Candidates?
```

In this situation I need to navigate to the import folder, search for the song, play it, close the player, switch back to beets window ect.

It would be nice if I could just play from selection for example like this:
Perhaps by using/extending the already existing play plugin?

```
# selection (default 1), Skip, Use as-is, as Tracks, Group albums,
Enter search, enter Id, aBort, eDit, edit Candidates, **Play**?
```

sampsyo:
Seems like a good idea! And not too complex if anyone's interested in jumping in.

PR #2360, opened 2016-12-31
play: Add prompt choice to importer

diomekes:
This is for #2008.

I hope my refactoring isn't way out of line.

### DIFF
--- a/beetsplug/play.py
+++ b/beetsplug/play.py
@@ -19,15 +19,33 @@ from __future__ import division, absolute_import, print_function
 
 from beets.plugins import BeetsPlugin
 from beets.ui import Subcommand
+from beets.ui.commands import PromptChoice
 from beets import config
 from beets import ui
 from beets import util
 from os.path import relpath
 from tempfile import NamedTemporaryFile
+import subprocess
 
 # Indicate where arguments should be inserted into the command string.
 # If this is missing, they're placed at the end.
 ARGS_MARKER = '$args'
+
+
+def play(command_str, paths, open_args, keep_open=False):
+    """Play items in paths with command_str and optional arguments. If
+    keep_open, return to beets, otherwise exit once command runs.
+    """
+    try:
+        if keep_open:
+            command = command_str.split()
+            command = command + open_args
+            subprocess.call(command)
+        else:
+            util.interactive_open(open_args, command_str)
+    except OSError as exc:
+        raise ui.UserError(
+            "Could not play the query: {0}".format(exc))
 
 
 class PlayPlugin(BeetsPlugin):
@@ -40,10 +58,13 @@ class PlayPlugin(BeetsPlugin):
             'use_folders': False,
             'relative_to': None,
             'raw': False,
-            # Backwards compatibility. See #1803 and line 74
+            # Backwards compatibility. See #1803 and line 155
             'warning_threshold': -2,
             'warning_treshold': 100,
         })
+
+        self.register_listener('before_choose_candidate',
+                               self.before_choose_candidate_listener)
 
     def commands(self):
         play_command = Subcommand(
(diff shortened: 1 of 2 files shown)


jrobeson:
I fixed the appveyor issue. Can you rebase this on masterand push again?

sampsyo:
This is a cool idea; thanks! Any chance you could add some documentation on the feature?

The refactoring also looks fine on first glance (I haven't done a full review yet). Anything in particular I should be looking for? (Or perhaps you can just give a brief overview of what needed changing?)

diomekes:
Yes, I forgot about adding documentation, I'll do that.

I just defined separate methods/functions so that the CLI and event listener could access them.  I did have to change the way the command is called for the importer prompt since the plugin originally would exit once the command was executed, but I needed it to return to beets.  All my tests had it working fine.

beetsplug/play.py
```diff
@@ -117,35 +111,76 @@ def play_music(self, lib, opts, args):
...
+                return u"{} {}".format(command_str, args)
+        else:
+            # Don't include the marker in the command.
+            return command_str.replace(" " + ARGS_MARKER, "")
+
+    def _playlist_or_paths(self, paths):
+        """Returns either the raw paths of items or a playlist of the items.
+        """
+        raw = config['play']['raw'].get(bool)
+        if raw:
```
sampsyo: This can be shortened to:

```py
if config['play']['raw']:
```

beetsplug/play.py
```diff
@@ -117,35 +111,76 @@ def play_music(self, lib, opts, args):
...
+            return paths
+        else:
+            return [self._create_tmp_playlist(paths)]
+
+    def _print_info(self, selection, command_str, open_args,
+                    item_type='track'):
+        """Prompts user whether to continue if playlist exceeds threshold.
```
sampsyo: I know this is minor, but docstrings should typically be written in the imperative voice—here, "Prompt" rather than "Prompts."

beetsplug/play.py
```diff
@@ -117,35 +111,76 @@ def play_music(self, lib, opts, args):
...
+                return True
 
+        # Print number of tracks or albums to be played, log command to be run.
         ui.print_(u'Playing {0} {1}.'.format(len(selection), item_type))
-        if raw:
-            open_args = paths
-        else:
-            open_args = [self._create_tmp_playlist(paths)]
-
         self._log.debug(u'executing command: {} {!r}', command_str, open_args)
```
sampsyo: Maybe these logging messages should go in the method that actually plays the music?

beetsplug/play.py
```diff
@@ -117,35 +111,76 @@ def play_music(self, lib, opts, args):
...
+        item_type += 's' if len(selection) > 1 else ''
         if warning_threshold and len(selection) > warning_threshold:
             ui.print_(ui.colorize(
                 'text_warning',
                 u'You are about to queue {0} {1}.'.format(
                     len(selection), item_type)))
 
             if ui.input_options((u'Continue', u'Abort')) == 'a':
-                return
+                return True
```
sampsyo: The method should explicitly `return False` at the end (when it doesn't return True).

beetsplug/play.py
sampsyo: Also, this docstring should probably explain what the return value means. Here, if I understand correctly, it returns true if the operation should be aborted (i.e., it returns false if the operation should actually happen, which is a little counterintuitive).

beetsplug/play.py
```diff
@@ -155,3 +190,21 @@ def _create_tmp_playlist(self, paths_list):
             m3u.write(item + b'\n')
         m3u.close()
         return m3u.name
+
+    def before_choose_candidate_listener(self, session, task):
```
sampsyo: Argh; I wish we had chosen a better name than `before_choose_candidate`. :scream: But that's hardly your fault.

docs/plugins/play.rst
```diff
@@ -29,6 +29,18 @@ would on the command-line)::
...
 command-line oriented, and you'll get its output in real time.
 
+Interactive Usage
+-----------------
+
+The `play` plugin can also be invoked during an import. If enabled, the plugin
+adds a `plaY` option to the prompt, so pressing `y` will execute the configured
+command and play the items currently being imported.
+
+Once you exit your configured player, you will be returned to the import
```
sampsyo: Perhaps more accurately, "once the configured command exists, …"

beetsplug/play.py
```diff
@@ -117,35 +111,76 @@ def play_music(self, lib, opts, args):
...
+        command_str = self._create_command_str(opts.args)
+
+        # If user aborts due to long playlist:
+        cancel = self._print_info(selection, command_str, open_args, item_type)
+
+        # Otherwise proceed with play command.
+        if not cancel:
+            play(command_str, paths, open_args)
+
+    def _create_command_str(self, args=None):
```
sampsyo: This can probably be called just `_command_str`.

beetsplug/play.py
```diff
@@ -19,17 +19,35 @@
...
 ARGS_MARKER = '$args'
 
 
+def play(command_str, paths, open_args, keep_open=False):
+    """Play items in paths with command_str and optional arguments. If
+    keep_open, return to beets, otherwise exit once command runs.
+    """
+    try:
+        if keep_open:
+            command = command_str.split()
```
sampsyo: I'd recommend using `beets.util.shlex_split` instead of `str.split` for more predictable command spitting behavior.

diomekes:
Thanks for the thorough review, I think I addressed all the issues and hopefully the warning_threshold configuration option handling is a little more sophisticated.

sampsyo:
This looks great; thank you! I don't know why the 3.6 test failed the most recent time around, but I'm going to merge this now. To simplify things, I may remove that old, backwards-compatible misspelling…

Issue #1831, opened 2016-01-23
Non-numeric track indices


awesomer:
When using the [P] option from https://github.com/beetbox/beets/issues/1689, I noticed this discrepancy :
1. Ride My Junk (Squarehead mix) - Waze & Odyssey (6:37)
2. Ride My Junk (Chamboche remix) - Waze & Odyssey (7:48)

vs. existing mb tracklist :

A. Ride My Junk (Squarehead mix) - Waze & Odyssey (?:??)
B. Ride My Junk (Chamboche remix) - Waze & Odyssey (?:??) 

http://musicbrainz.org/release/af714e5d-6cbb-497e-a495-868141eff541

I spoke briefly with sampsyo about it, he said that I should file this issue, to request support for non-numeric track identifiers :

1) in the database
2) in tags for formats which support them (ie, not mp3)

sampsyo:
Yep! Sounds reasonable. It's of course contingent on file formats supporting this kind of data. Another thorny issue is how we'd support both the common case (numbers stored as actual numbers, sorted numerically so 10 appears after 9) and the non-numeric case (stored as strings, sorted as strings, queried using substring search, etc.) without confusing the two.

diego-plan9:
Sounds like an useful addition indeed! Some preliminary look at the specifications (help would be appreciated in checking the rest of the formats):
* [ID3v2.3.0](http://id3.org/id3v2.3.0#TRCK) seems to specifically restrict it to numeric, or at most numeric-with-a-slash-in-between. 
(As a side note, playing around with mutagen allows to enter any string in the `TRCK`, saving the file without complaints. I'm not sure it's a good idea to differ from the standardand if other popular players/software would choke, though).

* FLAC does not seem to have a strict tags standard, although for the [suggested TRACKNUMBER field](https://xiph.org/vorbis/doc/v-comment.html), it states (a bit loosely):

> The track number of this piece if part of a specific larger collection or album

Leaving aside the database/ordering/confusion considerations raised by @sampsyo on the previous comments for a moment and going back to the `Print tracks` option, I'm wondering if the problem of *fully automatically* assigning vinyl track numbers during printing **when** the original files do not have vinyl track numbers is fully solvable: @awesomer, could you let me know if it is common that each *side* of a vinyl is stored on a separate folder, or for the files it to contain enough information to find out when each side starts?

[...]

PR #2363, opened 2017-01-02
Fetch non-numeric track indices from MusicBrainz and Discogs

lengtche:
This is a PR in reference to issue #1831.


sampsyo:
Great; thanks for getting the ball rolling here!

For the name of the field, maybe `alt_track` would suffice? Or, alternatively, `track_alt` to make the relationship with the current `track` field clear?

Also, to finish things up, would you mind:
* writing a short changelog entry that explains what's going on with that field
* cleaning up the whitespace changes
* taking a look at the [tests](https://travis-ci.org/beetbox/beets/jobs/188293344#L874), which look like they need updating

beets/autotag/mb.py
```diff
@@ -265,6 +265,7 @@ def album_info(release):
             )
             ti.disctitle = disctitle
             ti.media = format
+            ti.alt_track_no = track['number']
```
sampsyo: To complete my understanding, what does this `number` field usually hold? Is it typically a number, but sometimes something else?

lengtche: My understanding from looking at a handful of XML dumps is that `position` always holds an integer index of the track and the `number` holds whatever the user inputted when adding tracks. If you had an album on a double-sided CD, tape, vinyl, etc., then there's the potential of having the `number` being something like `A1`, or whatever the case may be.

So it seems to me that simply adding this data should have zero impact with sorting.

lengtche:
I can do what you requested.

I also like the idea of naming it `track_alt` for consistency.

beets/autotag/mb.py
sampsyo: Cool; that helps! Thanks for elaborating.

lengtche:
I believe I got everything done.

sampsyo:
Looks great; thanks!

I think the problem is related to adding a fixed attribute for the field, which isn't really necessary—I'm going to flip this back to a flexible attribute and see how things go.

sampsyo:
Looks like that worked! Thanks again for all your work; I'm merging now.

PR #2365, opened 2017-01-02
Absubmit add section in the documentation on skipped file by the plugin

inytar:
Add a section to the absubmit plugin documentation about files that are skipped by this plugin. As mentioned by @tux-00 in pull request #2342.


sampsyo:
Thanks. Some initial information was added in #2364 and b95c01fbea890334056f32f215a689aada1845fd, but this is even more detailed. Merging now…

Issue #2361, opened 2017-01-02
Similar artist e.g. from last.fm and storing in graph-db


SusannaMaria:
### Idea
I want to write a small plugin which collects "similar artists from artist" 
The first source can be last.fm: http://www.last.fm/de/api/show/artist.getSimilar
But I really want to go in direction of acoustic brainz and t-SNE

In the end, the data I will collect are weighted relations between artists and I'm not sure if sqlite directly s the best for doing this. But maybe with graphlite it can be interessting to stay on sqlite
https://github.com/eugene-eeo/graphlite 

But in the beginning to evaluate artist relation I will work with https://github.com/GraphAlchemist/GraphJSON
and
http://graphalchemist.github.io/Alchemy/#/examples

beets plugin -> lastfm -> export as GraphJSON file -> Import in Alchemy.js

sampsyo:
Seems very cool! It sounds like there are a couple of ideas for the data source: either getting the similarity data from Last.fm or computing it based on AcousticBrainz features. Do you have a plan for which you'd like to tackle first?

If the main goal is to export GraphJSON (rather than running graph queries directly on the data), I'm not certain we need a full-fledged graph database abstraction like GraphLite. It might make enough sense just to record plain adjacency lists in flexible attributes on the beets database schema as-is.

For a head start interacting with the Last.fm API, I might recommend stealing setup code from the `lastgenre` and `lastimport` plugins.

SusannaMaria:
Is it okay to use the API Key mentioned in the plugin.py for the simularity plugin?

Anyway I have created one at last.fm already, it you have concerns because I'm not a member of beets.
<img width="217" alt="clipboard01" src="https://cloud.githubusercontent.com/assets/14867328/21615708/fb5dcd68-d1de-11e6-974d-9977cd9eb777.png">

sampsyo:
Yep, `beets.plugins.LASTFM_KEY` is meant to be used by any plugin that needs it. Go right ahead.

And that name sounds great to me!

SusannaMaria:
![figure_1](https://cloud.githubusercontent.com/assets/14867328/21648997/63c5fa9e-d2a0-11e6-8c3e-1f58e3f066ff.png)

      beet similarity artist:Tool, artist:Mars Volta

SusannaMaria:
It seems that last.fm is not reliable to search beets artist with mbid tag. :-(
Found to much of last.fm artist without mbid tags. will try to search by name and query the user if more than one is found.

[...]

PR #2369, opened 2017-01-04
Similarity plugin

SusannaMaria:
see https://github.com/beetbox/beets/issues/2361

```
beet similarity -h
Usage: beet-script.py similarity [options]

Options:
  -h, --help            show this help message and exit
  -g FILE, --graphml=FILE
                        read and write Graph as gml-FILE.
  -s, --show            Show graph of relations.
```
figure_1


SusannaMaria:
Using https://networkx.github.io/documentation/networkx-1.10/reference/readwrite.gml.html
It's good supported by networks module and understood by gephi

beetsplug/similarity.py
```diff
@@ -0,0 +1,294 @@
+# -*- coding: utf-8 -*-
+# This file is part of beets.
+# Copyright 2016, Rafael Bodill http://github.com/rafi
```
sampsyo: You might want to use your name here. :smiley:

beetsplug/similarity.py
```diff
@@ -0,0 +1,294 @@
...
+from beets import plugins
+from beets.dbcore import types
+
+import networkx as nx
+import matplotlib.pyplot as plt
+import os.path
+
+_artistsOwned = list()
+_artistsForeign = list()
+_relations = list()
```
sampsyo: Do these need to be global variables, or would it work for them to be fields on the plugin class?

beetsplug/similarity.py
```diff
@@ -0,0 +1,294 @@
...
+
+REPLACE = {
+    u'\u2010': '-',
+}
+
+G = nx.Graph(day="Friday")
+
+
+class SimilarityPlugin(plugins.BeetsPlugin):
+    """List duplicate tracks or albums."""
```
sampsyo: This docstring (and the ones below) seem to be the same and also come from another plugin? They might need a little more elaboration.

beetsplug/similarity.py
```diff
@@ -0,0 +1,294 @@
...
+
+    def collect_artists(self, items):
+        """Collect artists from query."""
+        for item in items:
+            if item['mb_albumartistid']:
+                artistnode = ArtistNode(item['mb_albumartistid'],
+                                        item['albumartist'],
+                                        True)
+                if artistnode not in _artistsOwned:
+                    _artistsOwned.append(artistnode)
```
sampsyo: Perhaps this method could just return the owned artists instead of setting a global variable?

beetsplug/similarity.py
SusannaMaria: I put it as class members

beetsplug/similarity.py
SusannaMaria: :flushed: ... okay will fix it

beetsplug/similarity.py
SusannaMaria: Putting artistowned, artistforeign and relations as class members is okay?

SusannaMaria:
Ragarding computing similarity, it is definitly not enough to have just the similarity from last.fm. It's just one dimension of similarity. Maybe it is possible to combine t-SNE and last.fm in one plugin.

SusannaMaria:
Would it be interesting to have a feature that graphs are shown within the web-plugin using http://sigmajs.org/
I really want to check it with bigger graphs. The networkx feature to plot graph is too static.

sampsyo:
Yeah, I've been intrigued by sigma.js for other things—it looks great. One easy way to do it here would be to emit a short HTML file that can be opened in a browser to visualize a computed graph.

SusannaMaria:
Added D3-Visualization with json export
clipboard01

And now its possible to query hierarchically e.g. 
     beet similarity artist:tool -d 3 
this example means that beginning from the artist/band Tool it checks similar artists 3 levels down.

sampsyo:
Wow; this D3 output looks awesome! With that in place, is the networkx library still necessary? We could remove that code if it is less useful than D3.

SusannaMaria:
Yes it is possible to create the json directly from the structure. I will kick out the networkx :heart_eyes:
Maybe it is best to merge this afterwards with the web-plugin, to provide the json and html?

sampsyo:
Yay; great! That should make the plugin delightfully simple.

Maybe this would be a good candidate for an "extensible" web server, as proposed in #718?

SusannaMaria:
Okay lets do this, maybe I need some time. First question is, how to store, update the graph.

I have to provide a JSON for D3, so is it valid to store the graph as JSON inside SQLITE? Doing this in pure rdbms ... hm.
The graph (edges are Artists, vertices are relations to artist) combines artists from own collection and from "foreign" you don't have in your collection. The reference is the mb_id from owned and foreign. A usecase is also, to find new music out of relations.
It should be configurable how many related Artist (controlled by threshold and number) are searched and the depth of the graph. Idea is to have a "complete" graph of the collection, which can be updated by trigger, and a web frontend which provides a search and visualize of a "subset-graph" feature to the user.

Neigbourhoud analysis should be done outside the beet plugin, with a jquery/javascript plugin connected with the webfrontend right?!

sampsyo:
Ah, I think I see—I was originally thinking that networkx wouldn't be necessary at all for JSON output, but it looks like you're using it to generate the graph in memory and then dump it as JSON. That makes sense—if that's the easiest way, there's no need to remove the networkx dependency.

In general, I think re-generating the graph structure on every run of the plugin is fine. You're absolutely right that storing graph-structured information in the SQLite database is likely to be a pain.

SusannaMaria:
I'm starting playing with json serialization of objects. NetworkX is not necessary if I just want to create a json of graph, and I see right now no support from NetworkX for queries in a graph. 

I really want to store the graph structure, because of last.fm it takes long to process even a small query. but serialization into json looks fine right now, it can be reused by D3 to visualize und used to decode by beets.
So first thing for me is to find a nice and simple way to encode and decode the graph structure as json for beets.
http://stackoverflow.com/questions/3768895/how-to-make-a-class-json-serializable

sampsyo:
OK, that does make sense—storing the data to avoid re-fetching it is of course reasonable.

Would it make sense to store the data as standard GraphJSON on disk (e.g., in `~/.config/beets` along with our other data files)?

jrobeson:
Other than mac osx issues, couldn't we use the json sqlite extension? or is it not good enough?

SusannaMaria:
Just started again after long pause. 

The plugin will be only responsible to gain the similarity data from 
1. external sources (e.g. http://www.last.fm/de/api/show/artist.getSimilar) 
2. Calculate neighborhood of Bands using the acoustic fingerprint of already gained information from Acousticbrainz
The visualization will be done in the web-plugin using d3 or something completely different :-)

Right now the code is able to process last.fm.
e.g. 
       beet similarity tool

without other options this will collect all the similar bands within your collection recursively until it collects only unknown (from your own collection point of view) bands. 

Storage is done in json right now:
grafik
grafik
grafik

But it's planed to store it inside the sqlite database later if its okay with you guys.

SusannaMaria:
Have to test and bugfix right now, that collecting of data is stable. last.fm is awful slow and time should not be wasted. Really looking forward to explore t-SNE und work towards the visualization in 3D :-)

sampsyo:
Cool! Would it make sense to merge an initial version of the plugin, which can just fetch data, and add nifty features like visualization/clustering in future versions?

beetsplug/similarity/__init__.py
```diff
@@ -0,0 +1,459 @@
+# -*- coding: utf-8 -*-
+# This file is part of beets.
+# Copyright 2016, Susanna Maria Hepp http://github.com/SusannaMaria
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,e
```
jackwilsdon: Looks like a rogue `e` has slipped in here!

ribeaud:
How far are we here? I am seeing the PRs (https://github.com/beetbox/beets/pull/2369) but nothing happened since then? Would be interested...
Cheers,
christian

PR #2375, opened 2017-01-07
Change "items" to "albums" in config docs for sort_album

diomekes:
Just a small change for what looks like a typo in the docs.


sampsyo:
Yes; that didn't make much sense the way it was before! Thanks for catching this. :sparkles:

Issue #123, opened 2013-02-28
Playlists


ghost:
This issue was automatically migrated from Google Code.
_Original author: adrian.sampson (February 18, 2009 07:00:50)_
_Original issue: https://github.com/google-code-export/beets/issues/8_

Kernald:
I'm currently looking for a way to manage playlists with beets. I already use smartplaylists plugin, but there's something lacking with it: I can't specify any arbitrary order without relying on an existing field. Of course I could add a custom field named `playlist_rank` on an item, and order the playlist's query with this field, but what if the song is in two different playlists?

From an (my) user perspective, a simple way to manage playlists would require 5 functionalities:

- Create a new playlist
- Remove an existing playlist
- Display playlists
- Export playlists (automatic step whenever the database changes?) in different formats (m3u in a specific folder for MPD, for example)
- Edit playlists.

The edition part consists in multiple functionalities itself:

- Add a song
- Remove a song
- Move a song, which would be optional as it simply consists in a removal then addition of the said song.

sampsyo:
Yep, that seems ,Ike a great list of requirements.

If I could add one more: I've been fantasizing about having album playlists, not just track playlists. Perhaps even heterogeneous playlists that can contain a mixture of individual tracks and whole albums.

skorokithakis:
I'd like to plug my own format here: The [universal playlist format](https://universalplaylist.stavros.io/).

I'm working on a beets plugin that will use this, but it would be nice if it can get better support in core.

djhibee:
Hello,
I have created some shell [scripts](https://github.com/djhibee/playlists-manager) to maintain playlists  based on beets libraries. Some of the features requested above are implemented, like " When user A sends a playlist file (and just the playlist, not any of the music files) to user B, user B's music player should be able to find all the songs from the playlist that user B already owns, so they can be played."
 The main con is the performances for huge libraries but it still works. 
Unfortunately I did not wrote it in python as I wanted to play a bit with bash, so it needs to be migrated to include these as beets plugins.

skorokithakis:
@djhibee That's very interesting, I'm interested in playlist functionality and adding UPL support to it. I already have scripts to export to UPL, but importing is not working very well yet.

Holzhaus:
IMO beets should use standard M3U playlists instead of maintaining it's own in-database format. This would greatly simplify interaction with media players (MPD, etc.). Here's a list of features that would be nice:

[...]

PR #2380, opened 2017-01-10
[WIP] Add PlaylistQuery plugin

Xenopathic:
This plugin provides support for queries on M3U playlist files, giving some of the features requested in #123. In my workflow, this allows me to create and manage playlists with a music client (e.g. ncmpcpp) then leverage Beets to transform those tracks, e.g. exporting and converting them into a separate directory for consumption on another device.

It uses a flexible field type definition to add the `playlist:...` query pattern. I considered using prefixes, but that would make for strange behaviour when combined with a field selector. This approach adds a dynamic 'field' to items that allow for matching.

Two configuration parameters are present: `playlist_dir` is required, and defines the location of the playlist directory. `relative_to` defines how relative paths in M3U files are specified, as relative to the base music directory or relative to the playlist. It defaults to paths being relative to the base directory, since that is what ncmpcpp does.

TODO:
- [ ] Add documentation
- [ ] Write tests?

I wanted to get this PR open so people can start commenting on my code

### DIFF
--- a/beetsplug/playlistquery.py
+++ b/beetsplug/playlistquery.py
@@ -1,0 +1,70 @@
+# -*- coding: utf-8 -*-
+# This file is part of beets.
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+
+import os
+from beets.plugins import BeetsPlugin
+from beets.dbcore import FieldQuery, types
+from beets.util.confit import NotFoundError
+
+class PlaylistQuery(FieldQuery):
+    """Matches files listed by a playlist file.
+    """
+    relative_path = None
+    playlist_dir = None
+
+    def __init__(self, field, pattern, fast=True):
+        super(PlaylistQuery, self).__init__(field, pattern, fast)
+
+        playlist_file = pattern + '.m3u'
+        playlist_path = os.path.join(self.playlist_dir, playlist_file)
+
+        self.paths = []
+        with open(playlist_path, 'r') as f:
+            for line in f:
+                if line[0] == '#':
+                    # ignore comments, and extm3u extension
+                    continue
+                self.paths.append(os.path.normpath(
+                    os.path.join(self.relative_path, line.decode('utf-8').rstrip())
+                ))
+
+    def match(self, item):
+        return item.path.decode('utf-8') in self.paths
+
+
+class PlaylistType(types.String):
+    """Custom type for playlist query.
+    """
+    query = PlaylistQuery
+
+
+class PlaylistQueryPlugin(BeetsPlugin):
+    item_types = {'playlist': PlaylistType()}
+
+    def __init__(self):
+        super(PlaylistQueryPlugin, self).__init__()
+        self.register_listener('library_opened', self.library_opened)
+
+        PlaylistQuery.playlist_dir = self.config['playlist_dir'].as_filename()
+
+    def library_opened(self, lib):
+        try:
+            relative_to = self.config['relative_to'].as_choice(['base', 'playlist'])
+        except NotFoundError:
+            relative_to = 'base'
+
+        if relative_to == 'playlist':
+            PlaylistQuery.relative_path = PlaylistQuery.playlist_dir
+        else:
+            PlaylistQuery.relative_path = lib.directory


beetsplug/playlistquery.py
```diff
@@ -0,0 +1,70 @@
...
+    """Custom type for playlist query.
+    """
+    query = PlaylistQuery
+
+
+class PlaylistQueryPlugin(BeetsPlugin):
+    item_types = {'playlist': PlaylistType()}
+
+    def __init__(self):
+        super(PlaylistQueryPlugin, self).__init__()
```
sampsyo: You'll likely want to set defaults for the plugin's configuration here.

beetsplug/playlistquery.py
```diff
@@ -0,0 +1,70 @@
...
+        self.register_listener('library_opened', self.library_opened)
+
+        PlaylistQuery.playlist_dir = self.config['playlist_dir'].as_filename()
```
sampsyo: Using global state to keep track of this could be a little problematic for testing—the right thing to do here might be to look up the configuration option directly in the query class.

beetsplug/playlistquery.py
```diff
@@ -0,0 +1,70 @@
...
+    def library_opened(self, lib):
+        try:
+            relative_to = self.config['relative_to'].as_choice(['base', 'playlist'])
+        except NotFoundError:
+            relative_to = 'base'
+
+        if relative_to == 'playlist':
+            PlaylistQuery.relative_path = PlaylistQuery.playlist_dir
+        else:
+            PlaylistQuery.relative_path = lib.directory
```
sampsyo: Is there a difference between setting `playlist_dir` to be the same as `directory` vs. setting `relative_to` to `base`?

beetsplug/playlistquery.py
```diff
@@ -0,0 +1,70 @@
...
+from beets.util.confit import NotFoundError
+
+class PlaylistQuery(FieldQuery):
+    """Matches files listed by a playlist file.
+    """
+    relative_path = None
+    playlist_dir = None
+
+    def __init__(self, field, pattern, fast=True):
+        super(PlaylistQuery, self).__init__(field, pattern, fast)
```
sampsyo: I think you want `fast=False` here unless you implement a `clause` method (i.e., a pure-SQL way of executing the query).

sampsyo:
This is a nifty idea. I'm intrigued by the low-complexity approach of using a directory full of m3u files as the playlist store. Eventually, for the full set of #123 features, we *probably* want to keep playlists in the database itself. But for now, I'm curious about using this approach to prototype some of what we'd do eventually with that full-fledged implementation.

beetsplug/playlistquery.py
Xenopathic: Yes: `playlist_dir` affects where playlists are stored (and referenced from, so a `playlist:foobar` looks up `$playlist_dir/foobar.m3u`) while `relative_to` affects how relative paths inside the M3U file are handled. There is no formal M3U spec (https://en.wikipedia.org/wiki/M3U#File_format) so paths may be relative to the playlist file (`relative_to = playlist`) or relative to the base music directory (ncmpcpp does this, `relative_to = base`)

sampsyo: Aha, that makes sense! Thanks for clarifying.

beetsplug/playlistquery.py
Xenopathic: How will this work with access needed to `lib`? The Query objects don't get any context, so the best place I thought to put it was in the `library_opened` event on PlaylistQueryPlugin, which in turn requires some sort of global state passed to PlaylistQuery for instances to use. Unless there's an API that I haven't seen?

sampsyo: Hmm, yes, that is a problem. We haven't had a case where a query needed a reference to the database before—maybe we need to add that capability?

Xenopathic: At the moment a Query object is constructed with the field and pattern, which makes it very clean. I think it is unwise to expand that in the default case, since most of the time it will go unused. However, I can see potential for the Builder pattern here - have a Plugin object instantiate Query objects that it controls.

So currently, `construct_query_part()` in dbcore.queryparse directly constructs the selected `query_class`. Instead, it should know which Plugin object corresponds to each query class, then calls on that object to construct the desired class. The default implementation in BeetsPlugin would do what `construct_query_part()` does at the moment (i.e. check if we have a FieldQuery, and construct in two different ways) but can be overridden in custom plugins to do more complex things, passing in other information if desired.

sampsyo: Sure; good idea. Another relatively simple way would be to use a closure—that is, the plugin would just set `self.item_types` to reference a class that can access `lib` from the outer context.

I suppose one way to judge whether we've done it "right" is to ask whether the plugin would break if beets were ever to support multiple libraries for some reason—if there weren't a single, global `Library` object but queries would work on whatever library they're querying.

beetsplug/playlistquery.py
Xenopathic: I think the documentation could benefit from a small section on how to specify config defaults for plugins, it is currently missing so I had to look at other plugins for examples.

beetsplug/playlistquery.py
Xenopathic: I can't think of a good way to avoid using global state somewhere, without rewriting how the entire mechanism for registering queries works. Closures wouldn't work easily, since the PlaylistQuery class is defined inside PlaylistType, of which an instance belongs to PlaylistQueryPlugin, so the PlaylistQuery class inside the (now global) PlaylistType instance needs to be changed on `library_opened` events, taking you right back to mutable global state.

I don't think these two configuration variables as global state is necessarily a problem, though. If you consider PlaylistQueryPlugin and PlaylistQuery to be separate units of code, then each can be unit tested independently, where the global state isn't such a big deal.

terenc3:
@Xenopathic Any progress here?

PR #2384, opened 2017-01-12
Tiny update to clarify convert config options.

icefoxen:
Not sure why it wasn't explicitly explained before but it is now!


sampsyo:
Thanks! Yes, we could use an additional bullet for completeness. I think the idea here was that the next subsection would describe those options, but this is a little unclear. I'm going to keep that sentence and add "see the next section" to make this clearer.

sampsyo:
Added. Thanks again! :sparkles:

Issue #2177, opened 2016-08-29
Feature Request: API should return song/album paths


velvet-jones:
### Problem

We're writing a utility to synchronize music libraries and we could really use the ability to get the path to an album. Both for the web API and for the export plugin.
### Setup
- OS: Ubuntu 14.04.3
- Python version: 2.7.6
- beets version: 1.3.18
- Turning off plugins made problem go away (yes/no): N/A

My configuration (output of `beet config`) is: N/A

claman:
It's not perfect, but using the `info` command returns the path of the song in the first line

sampsyo:
Hi! I think someone recently added a path field to the web API, in https://github.com/beetbox/beets/pull/2050#issue-159936072. Can you check that out?

velvet-jones:
Thanks, we checked it and sure enough, there's a path - for individual tracks. Surely beets knows the path to each entire album in its database? I've been searching the code to try to figure out if the album path is possible to add to '/album'. Any ideas?

sampsyo:
There's an `album.path` field in beets, but it's actually just the dirname of the first track on the album. Just stripping the file name off of the item path should work.

irskep:
FYI I added paths to tracks in the web API in #2389. If that suffices, given @sampsyo's last comment, then this can probably be closed.

The history of this is a little confusing. #2050 added _relative_ paths originally. Then that was reverted due to #2182. Finally, I added it back behind a config flag in #2389. So people who need it can have it, and people who don't won't see a slowdown.

sampsyo:
Thanks for following up, @irskep.

PR #2389, opened 2017-01-14
A new 'beet web' endpoint and configuration option, with docs

irskep:
I am working on a web app which combines a beets library with the mpv player. Since mpv doesn't know about beets, but I am using mpv as the source of truth for the playback queue, I need to be able to turn a file path into a beets track for the purpose of showing the "now playing" info. Additionally, I need the API responses for tracks to contain their paths so I can enqueue the tracks in mpv.

To that end, I have added a new configuration option `exclude_paths_from_items`, which defaults to `true` (the current behavior) but which can be set to `false` to allow paths to make it into API responses. I figured you had a good reason for excluding paths in most cases.

I also added a new endpoint `/item/at_path/<path:path>`, which can be called like `/item/at_path/Users/irskep/Music/The Beatles/Abbey Road/01 Come Together.mp3` to get the beets track info for the file at that path. While you might think it's possible to just do a "path:" query at /item/query/, in practice it doesn't work for a couple reasons that are annoying to work around. I can elaborate on that if you like. There are about 3 layers of breakage.

I will add tests tomorrow. (It is after midnight for me!)

[You can see my project here.](https://github.com/irskep/summertunes) It is quite far along. I already wrote my own HTTP server to interface with beets in exactly the way I want, but I would rather just use the project's existing server and not maintain a separate codebase.

I am eager to know if you support this approach so I can make some decisions about my own project's implementation.

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -223,7 +223,6 @@ def item_query(queries):
 
 @app.route('/item/path/<path:path>')
 def item_at_path(path):
-    g.lib._connection().create_function('bytelower', 1, beets.library._sqlite_bytelower)
     query = beets.library.PathQuery('path', u'/' + path)
     item = g.lib.items(query).get()
     if item:
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -37,7 +37,7 @@ def _rep(obj, expand=False):
     out = dict(obj)
 
     if isinstance(obj, beets.library.Item):
-        if app.config['exclude_paths_from_items']:
+        if app.config.get('EXCLUDE_PATHS_FROM_ITEMS', True):
             del out['path']
         else:
             out['path'] = out['path'].decode('utf-8')
@@ -339,7 +339,7 @@ class WebPlugin(BeetsPlugin):
             # Normalizes json output
             app.config['JSONIFY_PRETTYPRINT_REGULAR'] = False
 
-            app.config['exclude_paths_from_items'] = (
+            app.config['EXCLUDE_PATHS_FROM_ITEMS'] = (
                 self.config.get('exclude_paths_from_items', True))
 
             # Enable CORS if required.
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -223,7 +223,7 @@ def item_query(queries):
 
 @app.route('/item/path/<path:path>')
 def item_at_path(path):
-    query = beets.library.PathQuery('path', u'/' + path)
+    query = beets.library.PathQuery('path', b'/' + path.encode('utf-8'))
     item = g.lib.items(query).get()
     if item:
         return flask.jsonify(_rep(item))
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -37,7 +37,10 @@ def _rep(obj, expand=False):
     out = dict(obj)
 
     if isinstance(obj, beets.library.Item):
-        del out['path']
+        if app.config['exclude_paths_from_items']:
+            del out['path']
+        else:
+            out['path'] = out['path'].decode('utf-8')
 
         # Get the size (in bytes) of the backing file. This is useful
         # for the Tomahawk resolver API.
@@ -309,6 +312,7 @@ class WebPlugin(BeetsPlugin):
             'host': u'127.0.0.1',
             'port': 8337,
             'cors': '',
+            'exclude_paths_from_items': True,
         })
 
     def commands(self):
@@ -326,6 +330,9 @@ class WebPlugin(BeetsPlugin):
             app.config['lib'] = lib
             # Normalizes json output
             app.config['JSONIFY_PRETTYPRINT_REGULAR'] = False
+
+            app.config['exclude_paths_from_items'] = (
+                self.config['exclude_paths_from_items'])
 
             # Enable CORS if required.
             if self.config['cors']:
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -38,7 +38,7 @@ def _rep(obj, expand=False):
 
     if isinstance(obj, beets.library.Item):
         if app.config.get('INCLUDE_PATHS', False):
-            out['path'] = out['path'].decode('utf-8')
+            out['path'] = util.displayable_path(out['path'])
         else:
             del out['path']
 
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -221,11 +221,14 @@ def item_query(queries):
     return g.lib.items(queries)
 
 
-@app.route('/item/at_path/<path:path>')
+@app.route('/item/path/<path:path>')
 def item_at_path(path):
-    try:
-        return flask.jsonify(_rep(beets.library.Item.from_path('/' + path)))
-    except beets.library.ReadError:
+    g.lib._connection().create_function('bytelower', 1, beets.library._sqlite_bytelower)
+    query = beets.library.PathQuery('path', u'/' + path)
+    item = g.lib.items(query).get()
+    if item:
+        return flask.jsonify(_rep(item))
+    else:
         return flask.abort(404)
 
 
@@ -339,8 +342,7 @@ class WebPlugin(BeetsPlugin):
             # Normalizes json output
             app.config['JSONIFY_PRETTYPRINT_REGULAR'] = False
 
-            app.config['INCLUDE_PATHS'] = (
-                self.config.get('include_paths', False))
+            app.config['INCLUDE_PATHS'] = self.config['include_paths']
 
             # Enable CORS if required.
             if self.config['cors']:
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -37,10 +37,10 @@ def _rep(obj, expand=False):
     out = dict(obj)
 
     if isinstance(obj, beets.library.Item):
-        if app.config.get('EXCLUDE_PATHS_FROM_ITEMS', True):
-            del out['path']
-        else:
+        if app.config.get('INCLUDE_PATHS', False):
             out['path'] = out['path'].decode('utf-8')
+        else:
+            del out['path']
 
         # Get the size (in bytes) of the backing file. This is useful
         # for the Tomahawk resolver API.
@@ -320,7 +320,7 @@ class WebPlugin(BeetsPlugin):
             'host': u'127.0.0.1',
             'port': 8337,
             'cors': '',
-            'exclude_paths_from_items': True,
+            'include_paths': False,
         })
 
     def commands(self):
@@ -339,8 +339,8 @@ class WebPlugin(BeetsPlugin):
             # Normalizes json output
             app.config['JSONIFY_PRETTYPRINT_REGULAR'] = False
 
-            app.config['EXCLUDE_PATHS_FROM_ITEMS'] = (
-                self.config.get('exclude_paths_from_items', True))
+            app.config['INCLUDE_PATHS'] = (
+                self.config.get('include_paths', False))
 
             # Enable CORS if required.
             if self.config['cors']:
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -340,7 +340,7 @@ class WebPlugin(BeetsPlugin):
             app.config['JSONIFY_PRETTYPRINT_REGULAR'] = False
 
             app.config['exclude_paths_from_items'] = (
-                self.config['exclude_paths_from_items'])
+                self.config.get('exclude_paths_from_items', True))
 
             # Enable CORS if required.
             if self.config['cors']:
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -221,6 +221,14 @@ def item_query(queries):
     return g.lib.items(queries)
 
 
+@app.route('/item/at_path/<path:path>')
+def item_at_path(path):
+    try:
+        return flask.jsonify(_rep(beets.library.Item.from_path('/' + path)))
+    except beets.library.ReadError:
+        return flask.abort(404)
+
+
 @app.route('/item/values/<string:key>')
 def item_unique_field_values(key):
     sort_key = flask.request.args.get('sort_key', key)
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -176,11 +176,16 @@ class QueryConverter(PathConverter):
         return ','.join(value)
 
 
+class EverythingConverter(PathConverter):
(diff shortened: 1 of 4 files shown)


sampsyo:
Cool; this looks like a really interesting project!

> While you might think it's possible to just do a "path:" query at /item/query/, in practice it doesn't work for a couple reasons that are annoying to work around. I can elaborate on that if you like. There are about 3 layers of breakage.

I would indeed like to know a little more about this. If something's broken, perhaps we should look into fixing it directly.

Anyway, adding a filesystem-oriented API to the web interface isn't a bad idea. In the medium term, I'd like to see something like that added to the proposed [AURA](https://github.com/beetbox/aura) API, which will replace our current ad hoc web plugin.

irskep:
Here are the issues I experienced:

1. It's not clear from the docs how to do a query that contains spaces. I think I eventually found something that worked but I now can't remember what it was.
2. The web plugin turns all slashes into spaces with no way to override this behavior, so trying to do a path query with the /item/query/ endpoint totally mangled the path.
3. Hacking around number 2, path queries just don't seem to work using `library.items()`. In the beets REST interface I wrote before I came to my senses, I ended up calling beets as a subprocess to get this to work: https://github.com/irskep/summertunes/blob/master/httbeets/httbeets.py#L95

irskep:
For AURA, it looks like the reference server is in JS. Will there also be a Python implementation that connects to beets?

test/test_plugin_mediafield.py
```diff
@@ -105,3 +106,7 @@ def test_overwrite_property(self):
             mediafile.MediaFile.add_field('artist', mediafile.MediaField())
         self.assertIn(u'property "artist" already exists',
                       six.text_type(cm.exception))
+
+
+def suite():
+    return unittest.TestLoader().loadTestsFromName(__name__)
```
irskep: Tests were failing without this, not sure why it wasn't there already

irskep:
Added a test for the config option, but I don't think I can test the new endpoint without including an actual test file since it reads it directly from the disk.

beetsplug/web/__init__.py
```diff
@@ -218,6 +221,14 @@ def item_query(queries):
     return g.lib.items(queries)
 
 
+@app.route('/item/at_path/<path:path>')
+def item_at_path(path):
+    try:
+        return flask.jsonify(_rep(beets.library.Item.from_path('/' + path)))
```
sampsyo: This looks like it will give the client access to any file on the host's filesystem—not just things in the beets database. Maybe it would make sense to construct a `PathQuery`, as in the `mpdstats` plugin, to look up an item by its path.

docs/plugins/web.rst
```diff
@@ -160,6 +162,13 @@ response includes all the items requested. If a track is not found it is silentl
 dropped from the response.
 
 
+``GET /item/by_path/...``
++++++++++++++++++++++
```
sampsyo: For symmetry with the endpoint called `query`, maybe this one should be called `path` instead of `by_path`?

sampsyo:
> - It's not clear from the docs how to do a query that contains spaces. I think I eventually found something that worked but I now can't remember what it was.
> - The web plugin turns all slashes into spaces with no way to override this behavior, so trying to do a path query with the /item/query/ endpoint totally mangled the path.

Hmm; both of these seem like they're flaws in the web plugin that we should fix! It's certainly annoying if you're not allowed to use slashes, even escaped slashes…

irskep:
The spaces-in-queries issue is one I also have on the command line or using `library.items()`.

As for the web plugin slash thing, yeah, that should probably get its own patch. But I also think a straightforward single file lookup by path endpoint is a good idea.

sampsyo:
> The spaces-in-queries issue is one I also have on the command line or using library.items().

On the command line, it's just a matter of using your shell escape syntax:

    beet ls 'foo bar'

or:

    beet ls foo\ bar

irskep:
If I try to construct a PathQuery, I get this exception:

```py
  File "/Users/steve/env/beets/lib/python3.5/site-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/steve/env/beets/lib/python3.5/site-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/steve/env/beets/lib/python3.5/site-packages/flask_cors/extension.py", line 161, in wrapped_function
    return cors_after_request(app.make_response(f(*args, **kwargs)))
  File "/Users/steve/env/beets/lib/python3.5/site-packages/flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/steve/env/beets/lib/python3.5/site-packages/flask/_compat.py", line 33, in reraise
    raise value
  File "/Users/steve/env/beets/lib/python3.5/site-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/steve/env/beets/lib/python3.5/site-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/steve/dev/3p/beets/beetsplug/web/__init__.py", line 227, in item_at_path
    item = g.lib.items(query).get()
  File "/Users/steve/dev/3p/beets/beets/library.py", line 1332, in items
  File "/Users/steve/dev/3p/beets/beets/library.py", line 1307, in _fetch
    model_cls, query, sort
  File "/Users/steve/dev/3p/beets/beets/dbcore/db.py", line 846, in _fetch
    rows = tx.query(sql, subvals)
  File "/Users/steve/dev/3p/beets/beets/dbcore/db.py", line 676, in query
sqlite3.OperationalError: no such function: BYTELOWER
```

I now remember that I had tried something similar and failed. I had a temporary workaround that did this on each request:

```py
g.lib._connection().create_function('bytelower', 1, beets.library._sqlite_bytelower)
```

But that's gross. I'm guessing the web plugin is not configuring its sqlite connection correctly but I don't understand how.

irskep:
On the other hand, the workaround does work! So no more direct file access required. I pushed my changes with the hack just so you can see what I had to do.

beetsplug/web/__init__.py
```diff
@@ -218,6 +221,17 @@ def item_query(queries):
     return g.lib.items(queries)
 
 
+@app.route('/item/path/<path:path>')
+def item_at_path(path):
+    g.lib._connection().create_function('bytelower', 1, beets.library._sqlite_bytelower)
```
irskep: Here is the hack that makes the bytelower function work in sqlite.

irskep:
Figured it out: `beets.library.Library` only adds the bytelower function on one thread's connection! I think the solution requires a tiny bit of refactoring to `beets.dbcore.Database` to make the connection factory overrideable, and then have `beets.library.Library` override it to add its custom function.

Let me know if you'd prefer another approach. This branch is getting more sophisticated than I expected but I really don't mind contributing. I used to work on a similarly large Python project (mrjob) so I know how important it is that contributors see the whole branch through.

irskep:
What pylintrc does this project use?

irskep:
I found and fixed the issue, see last commit.

irskep:
As a side note, I see that there is much discussion around changing the web API in beets. I just want to say, if you're worried about breaking backward compatibility for me specifically, don't worry about it for my case. I'm happy to update my project whenever the API overhaul lands. (As long as I can still do this path-based stuff!)

irskep:
Found the test/rsrc directory and added a full test for the `/item/path/` endpoint.

irskep:
I'm having trouble running the tests in Python 2.7, otherwise I would fix the problem. Not sure what's going on.

Here's what I get when I `python setup.py test`:

```py
  File "setup.py", line 142, in <module>
    'Programming Language :: Python :: 3.5',
  File "/Users/steve/.pyenv/versions/2.7/lib/python2.7/distutils/core.py", line 152, in setup
    dist.run_commands()
  File "/Users/steve/.pyenv/versions/2.7/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/Users/steve/.pyenv/versions/2.7/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/Users/steve/env/summertunes-2.7/lib/python2.7/site-packages/setuptools/command/test.py", line 211, in run
    self.run_tests()
  File "/Users/steve/env/summertunes-2.7/lib/python2.7/site-packages/setuptools/command/test.py", line 234, in run_tests
    **exit_kwarg
  File "/Users/steve/.pyenv/versions/2.7/lib/python2.7/unittest/main.py", line 94, in __init__
    self.parseArgs(argv)
  File "/Users/steve/.pyenv/versions/2.7/lib/python2.7/unittest/main.py", line 149, in parseArgs
    self.createTests()
  File "/Users/steve/.pyenv/versions/2.7/lib/python2.7/unittest/main.py", line 158, in createTests
    self.module)
  File "/Users/steve/.pyenv/versions/2.7/lib/python2.7/unittest/loader.py", line 128, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/Users/steve/.pyenv/versions/2.7/lib/python2.7/unittest/loader.py", line 113, in loadTestsFromName
    test = obj()
  File "/Users/steve/dev/3p/beets/test/testall.py", line 38, in suite
    s.addTest(__import__(modname).suite())
  File "/Users/steve/dev/3p/beets/test/test_web.py", line 12, in <module>
    from test import _common
  File "/Users/steve/dev/3p/beets/test/_common.py", line 33, in <module>
    from beets.ui import commands  # noqa: E402
  File "../beets/ui/commands.py", line 938, in <module>
    help=u"copy tracks into library directory (default)"
  File "/Users/steve/.pyenv/versions/2.7/lib/python2.7/optparse.py", line 1018, in add_option
    raise TypeError, "invalid arguments"
TypeError: invalid arguments
```

irskep:
OK, no idea why the tox tests are failing.

beets/library.py
```diff
@@ -1237,14 +1237,17 @@ def __init__(self, path='library.blb',
...
         self.path_formats = path_formats
         self.replacements = replacements
 
         self._memotable = {}  # Used for template substitution performance.
 
+    def _create_connection(self):
+        conn = super(Library, self)._create_connection()
+        conn.create_function('bytelower', 1, _sqlite_bytelower)
+        return conn
+
```
jrobeson: this was my fault for not understanding how create_function worked. I thought it went to all connection automatically :(

Can you submit this as a separate PR? It's a general bug fix.

sampsyo:
Great; thanks for all these changes!

About that `optparse` error: that's definitely strange, and probably not your fault. Does it still appear when using Nose to run the tests? You can check by getting with `pip install nose` and then running the command `nosetests` in the beets directory.

sampsyo:
I just pushed your fix for `test_plugin_mediafield.py` (thanks for catching this!!) but I can't seem to reproduce the test failure above, even with `python setup.py test`. Still investigating…

beets/library.py
irskep: Is that important even if this PR gets merged?

irskep:
I think my 2.7 tests are failing because I'm using Python 2.7.0, which broke optparse support for unicode. Installing a more recent version now.

irskep:
Yep, that was it.

The tests pass in 2.7 for me, so I still have no clue why they don't pass in Appveyor.

Fixed merge conflicts.

sampsyo:
Awesome; thanks. I'll take a look at the Windows failures as soon as I can—tomorrow, I hope.

irskep:
I suspect the tests are failing specifically on Windows because the path format is different. My web changes add a leading slash to the path because Flask strips it out.

I'm not yet sure how to deal with this because I don't know if Windows paths need to deal with drive letters or whatever.

irskep:
Tweaked my change to not need to add a leading slash to paths, by writing a custom PathConverter subclass with a more inclusive regex. UNIX users will have to add an extra slash to their paths, but Windows users will be able to write `/item/path/C:\\whatever`. If the tests pass, I'll update the docs.

irskep:
WOOHOO TESTS PASS

beetsplug/web/__init__.py
```diff
@@ -37,7 +37,10 @@ def _rep(obj, expand=False):
     out = dict(obj)
 
     if isinstance(obj, beets.library.Item):
-        del out['path']
+        if app.config.get('INCLUDE_PATHS', False):
+            out['path'] = out['path'].decode('utf-8')
```
sampsyo: You're right that everything needs to be converted to Unicode strings for JSON output. But, unfortunately, paths won't necessarily be encoded as UTF-8—on Unix, they can be arbitrary bytes. There are a few ways we might deal with this, such as base64-encoding the bytes in both directions. But one partial solution here would be to use `util.displayable_path`, which will at least avoid crashing here when the data contains un-decodeable bytes but will cause clients not to be able to find those paths.

beetsplug/web/__init__.py
```diff
@@ -218,6 +226,16 @@ def item_query(queries):
     return g.lib.items(queries)
 
 
+@app.route('/item/path/<everything:path>')
+def item_at_path(path):
+    query = beets.library.PathQuery('path', path.encode('utf-8'))
```
sampsyo: Nice! I like this solution—it gobbles up the entire remainder of the URL, which makes it safe to use slashes wherever necessary. Woohoo!

sampsyo (commented):
This is in great shape! There's one note in here about how to deal correctly with path encodings, but once that's sorted out this will be ready to merge.

beetsplug/web/__init__.py
```diff
@@ -37,7 +37,10 @@ def _rep(obj, expand=False):
     out = dict(obj)
 
     if isinstance(obj, beets.library.Item):
-        del out['path']
+        if app.config.get('INCLUDE_PATHS', False):
+            out['path'] = util.displayable_path(out['path'])
+        else:
+            del out['path']
```
irskep: Good catch on the encoding, I should have thought of that. Fixed now.

This project apparently does a really good job of Python 2/3 compatibility and encoding issues!

irskep:
Do you want me to clean up the commit history?

sampsyo:
No need, from my perspective! This looks great. Thanks again for working through all the details on this deceptively-subtle addition. :sparkles:

irskep:
Was I wrong not to use `obj.destination()` instead of `out['path']`? I've been doing some digging into the history of the web plugin and was surprised to see that paths had already been added, and then removed again.

Link to relevant diff in this PR: https://github.com/beetbox/beets/pull/2389/files#diff-61e3b07bd1b58062c4a48a7fd491550bR40

sampsyo:
Nope, I think you were right—using the actual filesystem path is right for your use case, I think, where you need direct access to the local file. And I think it's right for the other use cases mentioned in #2177 too.

PR #2390, opened 2017-01-15
beets.library.Library adds custom bytelower function to all connections, not just one

irskep:
Fixes the issue with multithreading I found when working on #2389.


sampsyo:
Thank you! :sparkles: Does this look right to you, @jrobeson?

jrobeson:
I think so, but I also got it wrong in the first place, because I thought the sqlite connection management was simpler than it was!

sampsyo:
Everything seems in order—thanks for the separate pull request, @irskep! :sparkles:

Issue #2393, opened 2017-01-16
Beets is crashing due to ascii encoding issues


fschoell:
### Problem

beets is crashing due to some ascii encoding issues when it should display the musicbrainz search result

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv import System\ of\ a\ Down\ -\ Hypnotize/
```

Led to this problem:

```
user configuration: /home/user/.config/beets/config.yaml
data directory: /home/user/.config/beets
plugin paths: 
Sending event: pluginload
library database: /home/user/.config/beets/library.db
library directory: /home/user/Music
Sending event: library_opened
Sending event: import_begin
Sending event: import_task_created
Sending event: import_task_start
Looking up: /some/folder/System of a Down - Hypnotize
Tagging System of a Down - Hypnotize
No album ID found.
Search terms: System of a Down - Hypnotize
Album might be VA: False
Sending event: albuminfo_received
Candidate: System of a Down - Hypnotize
Success. Distance: 0.01
Sending event: albuminfo_received
Candidate: System of a Down - Hypnotize
Success. Distance: 0.00
Sending event: albuminfo_received
Candidate: System of a Down - Hypnotize
Success. Distance: 0.01
Sending event: albuminfo_received
Candidate: System of a Down - Hypnotize
Success. Distance: 0.01
Sending event: albuminfo_received
Candidate: System of a Down - Hypnotize
Success. Distance: 0.01
Sending event: albuminfo_received
Candidate: System Of A Down - Hypnotize
Success. Distance: 0.04
Sending event: albuminfo_received
Candidate: System Of A Down - Hypnotize
Success. Distance: 0.04
Sending event: albuminfo_received
Candidate: System Of A Down - Hypnotize
Success. Distance: 0.04
Sending event: albuminfo_received
Candidate: System Of A Down - Hypnotize
Success. Distance: 0.04
Sending event: albuminfo_received
Candidate: System Of A Down - Hypnotize
Success. Distance: 0.04
Evaluating 10 candidates.

/some/folder/System of a Down - Hypnotize (12 items)
Sending event: before_choose_candidate
Tagging:
    System of a Down - Hypnotize
URL:
    https://musicbrainz.org/release/8a4034a9-7834-3b7e-a6f0-d0791e3731fb
(Similarity: 100.0%) (Vinyl, 2005, US)
Traceback (most recent call last):
  File "/usr/bin/beet", line 11, in <module>
[...]

PR #2398, opened 2017-01-19
On Python 3, use more flexible encoding to send bytes to stdout

sampsyo:
As explained in #2393, the last big issue in Python 3 is that `print` or `sys.stdout.write` can crash depending on the system's locale configuration. This change restores the Python 2 behavior, which avoids potential crashes by (a) using the `replace` error handler and (b) allowing the user to override the stdout encoding in the configuration.

Unfortunately, this breaks out test harnesses that capture stdout. One way to work around this might be to detect when we're in a test harness and just record the text strings directly.

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -136,16 +136,22 @@ def print_(*strings, **kwargs):
     txt += kwargs.get('end', u'\n')
 
     # Encode the string and write it to stdout.
-    txt = txt.encode(_out_encoding(), 'replace')
     if six.PY2:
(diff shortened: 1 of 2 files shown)


sampsyo:
Well, at least the tests pass! Does this seem crazy to you, @jrobeson?

jrobeson:
wouldn't the test harness break in the same way when run by the user with a `C` locale if that was done?

sampsyo:
Hmm, I don't quite see how... in the tests under Python 3, this change makes it so the locale encoding isn't used at all (everything just stays as `str`s).

PR #2399, opened 2017-01-19
Add bracket argument to aunique

### LINKED ISSUE #2397: remove [] from %aunique
%aunique in a path will by default add [] around the disambiguator.  If a disambiguator is found on one album but is empty on the other, then the album name will end with [].

There may be a reason to have the brackets as default, but it seems to me a better option for the user to set any characters wanted around the disambiguator string in the path format itself.

I can do a PR, but I figured it would be easier just to post here and discuss if necessary.  Here's the line of code:

https://github.com/beetbox/beets/blob/958ad430fc4854b2589ec3a2726b7ee269e5322e/beets/library.py#L1511

diomekes:
A little background as to why I felt this was needed:

If I have two albums that are disambiguated by `albumdisambig`, and one album has that field empty, then it would have an empty `[]`. In this case, I would only want the brackets gone.

Also, if I do something like `[$year%aunique{}]`, then I would have two sets of brackets, with the inner set possibly empty.  Plus, in this case, if the inner brackets are removed, then there is still a white space after the year and before the closing bracket.

So this change does a few things:

- It adds a third argument to %aunique{} that should be two characters to be used as brackets.
- It removes the blank space along with any brackets created from an empty disambiguator.
- It adds brackets when the fallback `album_id` is used as the disambiguator (I decided to add this for consistency, but maybe it was left out originally for a reason I'm not aware of).

This will close #2397.

Finally, I think adding the brackets to the `album_id ` fallback messed up the test for `library.py`.  I fixed it, but I don't know if I did it correctly.

And I'm totally fine if this request is considered unnecessary for whatever reason, especially if some of my changes look "ugly".

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -1455,7 +1455,7 @@ class DefaultTemplateFunctions(object):
         used. Both "keys" and "disam" should be given as
         whitespace-separated lists of field names, while "bracket" is a
         pair of characters to be used as brackets surrounding the
-        disambiguator or a white space to have no brackets.
+        disambiguator or empty to have no brackets.
         """
         # Fast paths: no album, no item or library, or memoized value.
         if not self.item or not self.lib:
--- a/beets/library.py
+++ b/beets/library.py
@@ -1469,13 +1469,14 @@ class DefaultTemplateFunctions(object):
 
         keys = keys or 'albumartist album'
         disam = disam or 'albumtype year label catalognum albumdisambig'
-        bracket = bracket or '[]'
+        if bracket is None:
+            bracket = '[]'
         keys = keys.split()
         disam = disam.split()
-        bracket = None if bracket == ' ' else list(bracket)
 
-        # Assign a left and right bracket from bracket list.
-        if bracket:
+        # Assign a left and right bracket or leave blank if argument is empty.
+        if len(bracket) == 2:
+            bracket = list(bracket)
             bracket_l = bracket[0]
             bracket_r = bracket[1]
         else:
--- a/beets/library.py
+++ b/beets/library.py
@@ -1453,7 +1453,9 @@ class DefaultTemplateFunctions(object):
         from "disam" is used in the string if one is sufficient to
         disambiguate the albums. Otherwise, a fallback opaque value is
         used. Both "keys" and "disam" should be given as
-        whitespace-separated lists of field names.
+        whitespace-separated lists of field names, while "bracket" is a
+        pair of characters to be used as brackets surrounding the
+        disambiguator or a white space to have no brackets.
         """
         # Fast paths: no album, no item or library, or memoized value.
         if not self.item or not self.lib:
@@ -1470,7 +1472,15 @@ class DefaultTemplateFunctions(object):
         bracket = bracket or '[]'
         keys = keys.split()
         disam = disam.split()
-        bracket = None if bracket is ' ' else list(bracket)
+        bracket = None if bracket == ' ' else list(bracket)
+
+        # Assign a left and right bracket from bracket list.
+        if bracket:
+            bracket_l = bracket[0]
+            bracket_r = bracket[1]
+        else:
+            bracket_l = u''
+            bracket_r = u''
 
         album = self.lib.get_album(self.item)
         if not album:
@@ -1504,18 +1514,17 @@ class DefaultTemplateFunctions(object):
 
         else:
             # No disambiguator distinguished all fields.
-            res = u' {1}{0}{2}'.format(album.id, bracket[0] if bracket else
-                                       u'', bracket[1] if bracket else u'')
+            res = u' {1}{0}{2}'.format(album.id, bracket_l, bracket_r)
             self.lib._memotable[memokey] = res
             return res
 
         # Flatten disambiguation value into a string.
         disam_value = album.formatted(True).get(disambiguator)
-        res = u' {1}{0}{2}'.format(disam_value, bracket[0] if bracket else u'',
-                                   bracket[1] if bracket else u'')
 
-        # Remove space and/or brackets if disambiguation value is empty
-        if bracket and res == u' ' + ''.join(bracket) or res.isspace():
+        # Return empty string if disambiguator is empty.
+        if disam_value:
+            res = u' {1}{0}{2}'.format(disam_value, bracket_l, bracket_r)
+        else:
             res = u''
 
         self.lib._memotable[memokey] = res
--- a/beets/library.py
+++ b/beets/library.py
@@ -1447,7 +1447,7 @@ class DefaultTemplateFunctions(object):
         cur_fmt = beets.config['time_format'].as_str()
         return time.strftime(fmt, time.strptime(s, cur_fmt))
 
-    def tmpl_aunique(self, keys=None, disam=None):
+    def tmpl_aunique(self, keys=None, disam=None, bracket=None):
         """Generate a string that is guaranteed to be unique among all
         albums in the library who share the same set of keys. A fields
         from "disam" is used in the string if one is sufficient to
@@ -1467,8 +1467,10 @@ class DefaultTemplateFunctions(object):
 
         keys = keys or 'albumartist album'
         disam = disam or 'albumtype year label catalognum albumdisambig'
+        bracket = bracket or '[]'
         keys = keys.split()
         disam = disam.split()
+        bracket = None if bracket is ' ' else list(bracket)
 
         album = self.lib.get_album(self.item)
         if not album:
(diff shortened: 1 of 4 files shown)


beets/library.py
```diff
@@ -1447,7 +1447,7 @@ def tmpl_time(s, fmt):
         cur_fmt = beets.config['time_format'].as_str()
         return time.strftime(fmt, time.strptime(s, cur_fmt))
 
-    def tmpl_aunique(self, keys=None, disam=None):
+    def tmpl_aunique(self, keys=None, disam=None, bracket=None):
```
sampsyo: Could you please add a description of `bracket` to the docstring?

beets/library.py
```diff
@@ -1467,8 +1467,10 @@ def tmpl_aunique(self, keys=None, disam=None):
 
         keys = keys or 'albumartist album'
         disam = disam or 'albumtype year label catalognum albumdisambig'
+        bracket = bracket or '[]'
         keys = keys.split()
         disam = disam.split()
+        bracket = None if bracket is ' ' else list(bracket)
```
sampsyo: - Is there a reason the condition is `bracket is ' '` instead of using the empty string `''`?
- `is` might work sometimes for comparing strings, but `==` is the right operator to use here.

```py
>>> 'a' is 'a'
True
>>> 'a' * 100 is 'a' * 100
False
>>> 'a' * 100 == 'a' * 100
True
```

beets/library.py
```diff
@@ -1502,13 +1504,20 @@ def tmpl_aunique(self, keys=None, disam=None):
...
+            res = u' {1}{0}{2}'.format(album.id, bracket[0] if bracket else
+                                       u'', bracket[1] if bracket else u'')
             self.lib._memotable[memokey] = res
             return res
 
         # Flatten disambiguation value into a string.
         disam_value = album.formatted(True).get(disambiguator)
-        res = u' [{0}]'.format(disam_value)
+        res = u' {1}{0}{2}'.format(disam_value, bracket[0] if bracket else u'',
+                                   bracket[1] if bracket else u'')
```
sampsyo: Maybe this and the prior line with bracket-formatting would be more legible if the two brackets were assigned into variables?

```py
if bracket:
    bracket_l = bracket[0]
    bracket_r = bracket[1]
else:
    bracket_l = u''
    bracket_r = u''
```

beets/library.py
```diff
@@ -1502,13 +1504,20 @@ def tmpl_aunique(self, keys=None, disam=None):
 
         else:
             # No disambiguator distinguished all fields.
-            res = u' {0}'.format(album.id)
+            res = u' {1}{0}{2}'.format(album.id, bracket[0] if bracket else
+                                       u'', bracket[1] if bracket else u'')
```
sampsyo: This is fine—I don't have a strong opinion about whether the brackets should appear here, but you're certainly right that it will look more consistent.

beets/library.py
```diff
@@ -1502,13 +1504,20 @@ def tmpl_aunique(self, keys=None, disam=None):
...
 
         # Flatten disambiguation value into a string.
         disam_value = album.formatted(True).get(disambiguator)
-        res = u' [{0}]'.format(disam_value)
+        res = u' {1}{0}{2}'.format(disam_value, bracket[0] if bracket else u'',
+                                   bracket[1] if bracket else u'')
+
+        # Remove space and/or brackets if disambiguation value is empty
+        if bracket and res == u' ' + ''.join(bracket) or res.isspace():
+            res = u''
```
sampsyo: Hmm… is there a cleaner way we can check for an empty disambiguation string? Perhaps, if this appears earlier, we can just use `if disam_value:` for the check.

docs/reference/pathformat.rst
```diff
@@ -112,21 +112,26 @@ will expand to "[2008]" for one album and "[2010]" for the other. The
...
+suffices, an arbitrary number is used to distinguish the two albums.
 
 The default identifiers are ``albumartist album`` and the default disambiguators
 are ``albumtype year label catalognum albumdisambig``. So you can get reasonable
 disambiguation behavior if you just use ``%aunique{}`` with no parameters in
 your path forms (as in the default path formats), but you can customize the
 disambiguation if, for example, you include the year by default in path formats.
 
+The default characters used as brackets are ``[]``. If a single blank space is
+used, then the disambiguator will not be surrounded by anything.
```
sampsyo: For full clarity, maybe something like this:

> The default characters used as brackets are `[]`. To change this, provide a third argument to the `%aunique` function consisting of two characters: the left and right brackets. Or, to turn off bracketing entirely, use a single blank space.

test/test_library.py
```diff
@@ -713,8 +713,8 @@ def test_use_fallback_numbers_when_identical(self):
         album2.year = 2001
         album2.store()
 
-        self._assert_dest(b'/base/foo 1/the title', self.i1)
-        self._assert_dest(b'/base/foo 2/the title', self.i2)
+        self._assert_dest(b'/base/foo [1]/the title', self.i1)
+        self._assert_dest(b'/base/foo [2]/the title', self.i2)
```
sampsyo: That looks like the right change to me!

beets/library.py
diomekes: The `bracket == ' '` is to check for a white space in the third argument for `%aunique` in order to remove any bracketing.

diomekes:
How do I add an empty field in the tests?  My efforts yielded `[0000]` for `year = None` and `[1]` for everything else.

beets/library.py
sampsyo: Right—I guess the question is whether the empty string might be a better "use nothing" indicator than a space.

test/test_library.py
```diff
@@ -730,6 +730,16 @@ def test_unique_sanitized(self):
         self._setf(u'foo%aunique{albumartist album,albumtype}/$title')
         self._assert_dest(b'/base/foo [foo_bar]/the title', self.i1)
 
+    def test_drop_empty_disam_string(self):
+        album1 = self.lib.get_album(self.i1)
+        album1.year = None
+        album1.store()
+        self._assert_dest(b'/base/foo/the title', self.i1)
```
sampsyo: Hmm; I suppose I assumed you had run into this with some music in your library. Maybe you can reuse the data from there? You could, for example, try setting some arbitrary field `foo` to be empty on one album and to have a known value on another, and then set up the `%aunique` invocation to use that as the only (or first) disambiguator.

diomekes: Yeah, that is what I ran into with my library.  I think adding the second album fixed it.

beets/library.py
```diff
@@ -1467,8 +1469,18 @@ def tmpl_aunique(self, keys=None, disam=None):
 
         keys = keys or 'albumartist album'
         disam = disam or 'albumtype year label catalognum albumdisambig'
+        bracket = bracket or '[]'
         keys = keys.split()
         disam = disam.split()
+        bracket = None if bracket == ' ' else list(bracket)
```
sampsyo: To avoid crashes, maybe this should check that `bracket` has length 2.

docs/reference/pathformat.rst
```diff
@@ -112,21 +112,28 @@ will expand to "[2008]" for one album and "[2010]" for the other. The
...
 The default identifiers are ``albumartist album`` and the default disambiguators
 are ``albumtype year label catalognum albumdisambig``. So you can get reasonable
 disambiguation behavior if you just use ``%aunique{}`` with no parameters in
 your path forms (as in the default path formats), but you can customize the
 disambiguation if, for example, you include the year by default in path formats.
 
+The default characters used as brackets are ``[]``. To change this, provide a
+third argument to the ``%aunique`` function consisting of two characters: the left
+and right brackets. Or, to turn off bracketing entirely, use a single blank
+space.
```
sampsyo: Great; this sounds good! But just to triple-check: is it necessary for the "null option" here to be a single space? That seems a little counterintuitive, but if there's a technical reason we can't use the empty string instead it's fine.

diomekes: Well, I set it up to default to `[]`.  I could change it to default to nothing, and that way we can get rid of the whole single space thing.  Is that what you mean?

sampsyo: Right—I meant that, to get no brackets, you would use `%aunique{foo,bar,}` instead of `%aunique{foo,bar, }`. To use the default square brackets, you would leave the comma off: `%aunique{foo,bar}`. In other words, the third argument to the function would be `''` for no brackets and `None` for the default.

diomekes: OK, I think I got it.  I was trying to distinguish `''` from `None` with `' '`, but I guess that's not necessary.

sampsyo:
Fantastic; thank you for finishing this up! It's all merged. :sparkles:

Issue #2400, opened 2017-01-20
test suite errors


antlarr:
### Problem

I'm trying to generate a beets package for openSUSE using python3 and got problems when running the test suite. Many tests fail with: "ValueError: underlying buffer has been detached" although that's not the ony problem happening.

The [test output log](https://github.com/beetbox/beets/files/720055/beets-test-output.txt) is large, so I'll copy here the relevant parts:

```
abuild@geeko.site:/home/abuild/beets> python3 setup.py test
running test
running egg_info
writing dependency_links to beets.egg-info/dependency_links.txt
writing entry points to beets.egg-info/entry_points.txt
writing top-level names to beets.egg-info/top_level.txt
writing beets.egg-info/PKG-INFO
writing requirements to beets.egg-info/requires.txt
reading manifest file 'beets.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
'test/rsrc/unicode\udce2\udc80\udc99d.mp3' not utf-8 encodable -- skipping
warning: no previously-included files matching '*.pyc' found under directory 'test/rsrc'
warning: no previously-included files matching '*.pyo' found under directory 'test/rsrc'
warning: no files found matching 'man/beet.1'
warning: no files found matching 'man/beetsconfig.5'
no previously-included directories found matching 'docs/_build'
warning: no previously-included files matching '.DS_Store' found anywhere in distribution
writing manifest file 'beets.egg-info/SOURCES.txt'
running build_ext
test_album_item (test_vfs.VFSTest) ... Sending event: database_change
Sending event: database_change
Sending event: database_change
Sending event: database_change
Sending event: database_change
Sending event: database_change
ok
test_singleton_item (test_vfs.VFSTest) ... Sending event: database_change
Sending event: database_change
Sending event: database_change
Sending event: database_change
Sending event: database_change
Sending event: database_change
ok
```

... I'll jump to where errors start happening ...

```
test_solo_pattern (test_autotag.StringDistanceTest) ... ok
test_different_artist (test_autotag.TrackDistanceTest) ... ok
test_different_title (test_autotag.TrackDistanceTest) ... ok
test_identical_tracks (test_autotag.TrackDistanceTest) ... ok
test_various_artists_tolerated (test_autotag.TrackDistanceTest) ... ok

======================================================================
ERROR: test_pretend_sync_from_itunes (test_metasync.MetaSyncTest)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #2403, opened 2017-01-21
Run python2 or python3 depending on what's being used

antlarr:
On a system with dependencies installed for python3 but not for python2,
we have to make sure python3 is used everywhere since 'python' might be running
the python2 interpreter.

This helps with some problems in #2400, but doesn't fix the issue completely.

### DIFF
--- a/test/test_convert.py
+++ b/test/test_convert.py
@@ -28,6 +28,15 @@ from beets.mediafile import MediaFile
 from beets import util
 
 
+def shell_quote(text):
+    if sys.version_info[0] < 3:
+        import pipes
+        return pipes.quote(text)
+    else:
+        import shlex
+        return shlex.quote(text)
+
+
 class TestHelper(helper.TestHelper):
 
     def tagged_copy_cmd(self, tag):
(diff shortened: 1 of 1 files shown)


sampsyo:
Interesting! I don't quite see how this causes the problem yet, but perhaps using `sys.executable` would be the right thing to use the same interpreter as the host?

jrobeson:
arg, i totally forget about that call :( glad somebody noticed.

antlarr:
Yes indeed, sys.executable is better. I added another commit to use that one instead.

About the how this causes the problem... In a system with python2 and python3 installed, with python being a symlink to python2 and dependencies like Unidecode being only available for python3, running python as test_convert.py did before caused all sort of "No module named 'whatever' " errors.

sampsyo:
Thanks; that all makes sense.

It looks like things are failing on Windows. I'll look into this now.

sampsyo:
Hmm; this is a little more subtle than it looked. Unless we do something drastic, we'll need to shell-escape `sys.executable`—otherwise, if the path has spaces or quotes in it, then the call will fail. It would probably work to use `shlex.quote`, but that's only available on 3.3+. On Windows, the backslashes in the path are causing the problem.

In `convert_stub.py`, is it only `six` and the `arg_encoding` we use that might not be present on the "wrong" Python? If so, we can certainly find a way to avoid those dependencies…

sampsyo:
Alright, here's an alternative—I just made that sub-invocation of `convert_stub.py` independent of any dependencies in #2408. Does that solve the same set of problems as this PR, @antlarr?

antlarr:
It seems shell escaping the python executable works on windows. Maybe you can squash the commits since it doesn't make much sense to have so many commits only for that :).

sampsyo:
Great! This is an indication that we diagnosed the problem correctly, but I think we still need real shell escaping… just putting single quotes around an argument will run into trouble if the string has quotes or backslashes in it. I mentioned `shlex.quote` above, which we can at least use on Python 3, and it looks like there might even be an alternative on Python 2! http://stackoverflow.com/questions/26790916/python-3-backward-compatability-shlex-quote-vs-pipes-quote

PR #2411, opened 2017-01-23
Add a plugin to update a Kodi music library

Pauligrinder:
I created one for an older version before, but it didn't work since the change to Python 3. So I created a new one that works.

### DIFF
--- a/beetsplug/kodiupdate.py
+++ b/beetsplug/kodiupdate.py
@@ -1,4 +1,17 @@
 # -*- coding: utf-8 -*-
+# This file is part of beets.
+# Copyright 2017, Pauli Kettunen.
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
 
 """Updates a Kodi library whenever the beets library is changed. This is based on the Plex Update plugin.
 
@@ -39,6 +52,7 @@ class KodiUpdate(BeetsPlugin):
             u'user': u'kodi',
             u'pwd': u'kodi'})
 
+        config['kodi']['pwd'].redact = True
         self.register_listener('database_change', self.listen_for_db_change)
 
     def listen_for_db_change(self, lib, model):
--- a/beetsplug/kodiupdate.py
+++ b/beetsplug/kodiupdate.py
@@ -12,8 +12,6 @@ Put something like the following in your config.yaml to configure:
 from __future__ import division, absolute_import, print_function
 
 import requests
-import json
-from requests.auth import HTTPBasicAuth
 from beets import config
 from beets.plugins import BeetsPlugin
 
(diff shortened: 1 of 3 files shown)


sampsyo:
Cool; looks good so far! We should include this alongside our other `*update` plugins. :sparkles:

It looks like this needs a few changes to be ready to merge:

* The code needs the standard license header at the top.
* Travis is reporting some style issues—you might try running `flake8` on the new code.
* We'll need documentation for the new plugin.

beetsplug/kodiupdate.py
```diff
@@ -0,0 +1,64 @@
...
+class KodiUpdate(BeetsPlugin):
+    def __init__(self):
+        super(KodiUpdate, self).__init__()
+
+        # Adding defaults.
+        config['kodi'].add({
+            u'host': u'localhost',
+            u'port': 8080,
+            u'user': u'kodi',
+            u'pwd': u'kodi'})
```
wordofglass: I think you should set
```python
config['kodi']['pwd'].redact = True
```
here to hide the password from `beet config` output.

Pauligrinder: Will do!

Pauligrinder:
Sorry for the delay! I'm not using github very actively so I didn't notice the comments :) I'll see what I can do about the travis build failing, it's probably something really stupid since I'm not that familiar with Python. I'll deal with the header and documentation once I get that build sorted.

sampsyo:
Awesome; thank you for looking into it!

For what it's worth, it's pretty easy to run the style checker locally. You can just run `pip install flake8` to get the tool and then type `flake8 beets beetsplug` to run the checker over the whole codebase.

Pauligrinder:
Hmm, I didn't know python is this picky with all sorts of things :) Nice learning experience.

Pauligrinder:
Yay, it works :) Here's a quick little documentation file for the plugin.
[kodiupdate.rst.txt](https://github.com/beetbox/beets/files/810210/kodiupdate.rst.txt)

Pauligrinder:
By the way! Since I'm an Android developer, I was thinking about making a beets remote control app. I don't know how useful that'd actually be, but it would be cool for sure ;)
Any ideas on what features such an app might have apart from browsing the library? Editing would be pretty convenient compared to using the cli at least!

Pauligrinder:
And now it's failing thanks to the documentation file. How do I make it not check that file?

sampsyo:
Hey, thanks again for the revisions! The right fix here for that docs issue, listed [here](https://travis-ci.org/beetbox/beets/jobs/206590531#L860):

    /home/travis/build/beetbox/beets/docs/plugins/kodiupdate.rst:: WARNING: document isn't included in any toctree

is to add the new document to the table of contents for plugin documentation. I'll do that now…

sampsyo:
OK, changes pushed. Let's see what the CI says!

An Android or other non-CLI frontend for beets is a really interesting idea, and it's something that we toss around from time to time. I even noticed [this effort](https://github.com/iverasp/androidbeets) on GitHub a while back. You're right that what we need is an API—and there's a larger effort at foot there too. Specifically, take a look at the [AURA](https://github.com/beetbox/aura) project, where we're trying to define what a standard API for beets would look like. Currently, it's read-only, but I'd love to hear your input about how to make it capable of supporting editing too!

Pauligrinder:
That looks like a pretty good spec already! 
I'd say it just needs PUT calls for /tracks/:id (or /albums/:id or /artists/:id, don't know if it's reasonable to change entire albums or artists though), taking in json data with something like { "artist":"some band", "title":"some song title", "genre" : "some genre", ...}.

sampsyo:
Cool! I'd be interested to collaborate on making this into a real API if you ever want to get started on that Android client.

sampsyo:
This plugin is all merged up! :sparkles: Thanks again for writing it.

Pauligrinder:
Awesome, glad I could contribute :)

Issue #773, opened 2014-05-16
mpdstats: mpd events are processed twice


rafi:
Running `beet -v  mpdstats` shows that mpd events are processed twice, for example:
```
mpdstats: playing /mnt/media/music/10 Her Majesty.mp3
Sending event: database_change
mpdstats: updated: last_played = 1400247006 [/mnt/media/music/10 Her Majesty.mp3]
mpdstats: playing /mnt/media/music/10 Her Majesty.mp3
Sending event: database_change
mpdstats: updated: last_played = 1400247006 [/mnt/media/music/10 Her Majesty.mp3]
Sending event: database_change
```

brunal:
This does not happen for all events, and depends on the sequence of commands executed. Example following. Music redacted for terseness (it's actually some awesome Grateful Dead).
```
# standard stuff
user configuration: /home/bru/.config/beets/config.yaml
data directory: /home/bru/.config/beets
Sending event: pluginload
library database: /home/bru/.beetsmusic.blb
library directory: /home/bru/Musique
Sending event: library_opened
mpdstats: connecting to localhost:6600
mpdstats: playing /home/bru/Musique/A.mp3
Sending event: database_change
mpdstats: updated: last_played = 1420711398 [/home/bru/Musique/A.mp3

# the song is now naturally over
Sending event: database_change
mpdstats: updated: play_count = 1 [/home/bru/Musique/A.mp3]
mpdstats: played /home/bru/Musique/A.mp3
Sending event: database_change
mpdstats: updated: rating = 0.6875 [/home/bru/Musique/A.mp3]
mpdstats: playing /home/bru/Musique/B.mp3
Sending event: database_change
mpdstats: updated: last_played = 1420711519 [/home/bru/Musique/B.mp3]
mpdstats: playing /home/bru/Musique/B.mp3
Sending event: database_change
mpdstats: updated: last_played = 1420711522 [/home/bru/Musique/B.mp3]
# it is happening twice indeed. Now I'm skipping to the next song

Sending event: database_change
mpdstats: updated: skip_count = 1 [/home/bru/Musique/B.mmp3]
mpdstats: skipped /home/bru/Musique/B.mp3
Sending event: database_change
mpdstats: updated: rating = 0.3125 [/home/bru/Musique/B.mp3]
mpdstats: playing /home/bru/Musique/C.mp3
Sending event: database_change
mpdstats: updated: last_played = 1420711578 [/home/bru/Musique/C.mp3]
[...]

PR #2412, opened 2017-01-23
mpdstats: More general approach to multiple on_play calls for the same song

artemutin:
This change is related to #773. On my machine, with mpd 20.2 and beets 1.4.4, redundant second on_play call with same song playing, happens only on start of playing queue. Afterwards, this event is triggering only once per song change, no matter if previous song was skipped or naturally finished. And current version of on_play in my case resulted in broken state transition and ruined rating and play counts 
So, I've made minor changes, which in my believe respects both cases.

### DIFF
--- a/beetsplug/mpdstats.py
+++ b/beetsplug/mpdstats.py
@@ -263,25 +263,27 @@ class MPDStats(object):
         played, duration = map(int, status['time'].split(':', 1))
         remaining = duration - played
 
-        if self.now_playing and self.now_playing['path'] != path:
-            skipped = self.handle_song_change(self.now_playing)
-            # mpd responds twice on a natural new song start
-            going_to_happen_twice = not skipped
-        else:
-            going_to_happen_twice = False
+        if self.now_playing:
+            if self.now_playing['path'] != path:
+                self.handle_song_change(self.now_playing)
+            else:  # in case we got mpd play event with same song playing multiple times
+                # assume low diff means redundant second play event after natural song start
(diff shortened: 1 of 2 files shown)


beetsplug/mpdstats.py
```diff
@@ -263,25 +263,27 @@ def on_play(self, status):
...
-                'beets_item': self.get_item(path),
-            }
-
-            self.update_item(self.now_playing['beets_item'],
-                             'last_played', value=int(time.time()))
+        if self.now_playing:
+            if self.now_playing['path'] != path:
+                self.handle_song_change(self.now_playing)
+            else:  # in case we got mpd play event with same song playing multiple times
+                # assume low diff means redundant second play event after natural song start
```
sampsyo: I know this is a tiny thing, but please use full sentences ("Starting with a capital letter, and ending with a period.") in comments.

sampsyo:
OK! I'll admit I don't fully understand the vagaries of the mpd interaction in this piece, but if you say it's more accurate this way, I'm willing to trust that. Thanks for looking into it! :sparkles:

Looks like a couple of long lines need to be wrapped: https://travis-ci.org/beetbox/beets/jobs/194469215

Also, would you mind writing a quick changelog entry? Then we can merge this.

artemutin:
I've made changes you asked for. Should I squash this two commits to one?

sampsyo:
No squash necessary. Thank you for contributing! :sparkles:

PR #2415, opened 2017-01-24
fetchart: Clean-up data structures

wordofglass:
When I restructured some of fetchart some time ago, I introduced the `extra` dictionary to pass around settings. In hindsight, this appears rather ugly and unnecessary, since it only duplicates attributes of the plugin object without any further processing. The `paths` that were included in this dictionary are the only exception. A separation between global settings and album-specific data seems sensible, though.

At least I couldn't remember any valid reason for the dictionary approach when looking at the code a few day ago..

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -69,7 +69,7 @@ class Candidate(object):
         self.match = match
         self.size = size
 
-    def _validate(self, extra):
+    def _validate(self, plugin):
         """Determine whether the candidate artwork is valid based on
         its dimensions (width and ratio).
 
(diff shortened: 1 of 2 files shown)


sampsyo:
This looks right to me! You're right that an intermediate data structure doesn't help too much here. This reminds me of the "threading-through" boilerplate we used to need to write before we switched to Confuse (aka Confit) to make configuration options globally accessible. It looks like the same problem is cropping up in a microcosm here.

I also did a light code review and didn't find any issues. Please feel free to hit the merge button whenever you're ready.

Issue #2410, opened 2017-01-23
Docs seem incorrect on how medium recommendation is determined


bartkl:
**EDIT:** *Reading this again I now doubt if it's not simply me who misunderstands. Maybe the docs are correct and my understanding of what it means if 'the importer suggests a match' and what recommendation leven is required for it. If I'm wrong about my report below, could someone perhaps elaborate on how this works? Thanks!*

The documentation on the medium recommendation threshold seems not to correspond with the code.

Quoting from the [docs](https://beets.readthedocs.io/en/v1.4.3/reference/config.html#autotagger-matching-options):
> The medium_rec_thresh and rec_gap_thresh options work similarly. When a match is **above** the medium recommendation threshold or the distance between it and the next-best match is above the gap threshold, the importer will suggest that match but not automatically confirm it. Otherwise, you’ll see a list of options to choose from.

I really felt like the mention of **above** I just emphasized in bald should be **below**, so I decided to check the code.

Using the search function on the repository, I am fairly confident that the sole piece of code to which this piece of documentation belongs is in the `_recommendation` function in `match.py`. The lines of code responsible for the medium recommendation are (as quoted from [match.py](https://github.com/beetbox/beets/blob/a414872430960546c8c5d768a73028e0066b11cf/beets/autotag/match.py#L290)):

```
elif min_dist <= config['match']['medium_rec_thresh'].as_number():
        # Medium recommendation level.
        rec = Recommendation.medium
    elif len(results) == 1:
        # Only a single candidate.
        rec = Recommendation.low
    elif results[1].distance - min_dist >= \
            config['match']['rec_gap_thresh'].as_number():
        # Gap between first two candidates is large.
        rec = Recommendation.low
```

So indeed, the docs should indeed say **below**, at least that's what I read from this.

But there's something else going on. The medium recommendation seems independent on the recommendation gap threshold (which is however used to determine a low recommendation as you can read here). So here too is a discrepancy with the segment from the docs quoted above.

Greetings!
Bart

sampsyo:
Hey, I think you're right! It looks like this should say "below." Directions are tricky when distance is the inverse of similarity...

Would you mind making a PR with the edit?

bartkl:
Haha, yes very understandable with that inverse relation.

How do you feel about my bottom remark about the medium recommendation being independent of the gap threshold?

I am a relative noob to Github: could you tell me in a brief sentence how such a pull request works? Or would it be clear to me if I read a tutorial about it?

Thanks

[...]

PR #2417, opened 2017-01-27
Update config.rst: see issue #2410

bartkl:
`When a match is above the *medium* recommendation threshold`
->
`When a match is below the *medium* recommendation threshold`

As discussed in issue #2410, which is fixed by this edit.


sampsyo:
Thanks for the fix! And yes, @Freso has it right about commit messages—this might be something to work on for the future.

bartkl:
Thanks for the feedback, I'll take it home!

Issue #1147, opened 2014-12-16
Support template functions in queries


djl:
I've written [a simple plugin](https://gist.github.com/djl/e72aac2fa57baec73817) which provides a template function for specifying time deltas. With this function you can do things like:

What's been added in the last day/week/month?

```
$ beet ls "added:%timedelta{1d}.."
$ beet ls "added:%timedelta{1w}.."
$ beet ls "added:%timedelta{1m}.."
```

Show me everything imported last month (relevant to the current month):

```
# the second argument is an optional strftime format
$ beet ls "added:%timedelta{1m,%Y-%m}"
```

Everything released this year (again, this will change based on the current date):

```
# the first argument is also optional and defaults to '0d' (today)
$ beet ls "year:%timedelta{,%Y}"
```

This is neat because it allows date-based queries which will update themselves over time.

Anyway, the problem with this approach is that template functions don't work in queries :(

Would it possible/useful to support add support for this?

(I also tried extending the query syntax to support the same functionality but ran into another issue. The new syntax, being a simple regex replace, can be used anywhere in a query so I can't easily subclass an existing Query class. I'd need to subclass at least two existing classes and from what I've seen through tests, two query classes can't use the same prefix.)

PR #2418, opened 2017-01-28
Relativedate

euri10:
this adds the ability to query with relative dates in the DateQuery

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -561,13 +561,13 @@ class Period(object):
             else:
                 m = 1
             if timespan == 'y':
-                date = datetime.now() + m*relativedelta(years=int(quantity))
+                date = datetime.now() + m * relativedelta(years=int(quantity))
             elif timespan == 'm':
-                date = datetime.now() + m*relativedelta(months=int(quantity))
+                date = datetime.now() + m * relativedelta(months=int(quantity))
             elif timespan == 'w':
-                date = datetime.now() + m*relativedelta(weeks=int(quantity))
+                date = datetime.now() + m * relativedelta(weeks=int(quantity))
             elif timespan == 'd':
-                date = datetime.now() + m*relativedelta(days=int(quantity))
+                date = datetime.now() + m * relativedelta(days=int(quantity))
 
             precision = 'relative'
         else:
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -551,10 +551,11 @@ class Period(object):
         """Parse a date and return a `Period` object or `None` if the
         string is empty.
         """
-        if re.match('@([+|-]?)(\d+)([y|m|w|d])', string) is not None:
-            sign = re.match('@([+|-]?)(\d+)([y|m|w|d])', string).group(1)
-            quantity = re.match('@([+|-]?)(\d+)([y|m|w|d])', string).group(2)
-            timespan = re.match('@([+|-]?)(\d+)([y|m|w|d])', string).group(3)
+        pattern_dq = '@([+|-]?)(\d+)([y|m|w|d])'
+        if re.match(pattern_dq, string) is not None:
+            sign = re.match(pattern_dq, string).group(1)
+            quantity = re.match(pattern_dq, string).group(2)
+            timespan = re.match(pattern_dq, string).group(3)
             if sign == '-':
                 m = -1
             else:
@@ -564,7 +565,7 @@ class Period(object):
             elif timespan == 'm':
                 date = datetime.now() + m * timedelta(days=int(quantity) * 30)
             elif timespan == 'w':
-                date = datetime.now() + timedelta(days=int(quantity) * 7)
+                date = datetime.now() + m * timedelta(days=int(quantity) * 7)
             elif timespan == 'd':
                 date = datetime.now() + m * timedelta(days=int(quantity))
             precision = 'relative'
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -554,22 +554,25 @@ class Period(object):
         relative.
         An absolute date has to be like one of the date_formats '%Y' or '%Y-%m'
         or '%Y-%m-%d'
-        A relative date begins by '@ 'and has to follow the pattern_dq format
-        '@([+|-]?)(\d+)([y|m|w|d])'
-        - '@' indicates it's a date relative to now()
-        - the optional '+' or '-' sign, which defaults to '+' will increment or
-        decrement now() by a certain quantity
-        - that quantity can be expressed in days, weeks, months or years
-        respectively 'd', 'w', 'm', 'y'
-        Please note that this relative calculation is rather approximate as it
-        makes the assumption of 30 days per month and 365 days per year
+        A relative date consists of three parts:
+        - a ``+`` or ``-`` sign is optional and defaults to ``+``. The ``+``
+        sign will add a time quantity to the current date while the ``-`` sign
+        will do the opposite
+        - a number follows and indicates the amount to add or substract
+        - a final letter ends and represents the amount in either days, weeks,
+        months or years (``d``, ``w``, ``m`` or ``y``)
+        Please note that this relative calculation makes the assumption of 30
+        days per month and 365 days per year.
         """
-        pattern_dq = '@([+|-]?)(\d+)([y|m|w|d])'
+
+        pattern_dq = '(?P<sign>[+|-]?)(?P<quantity>[0-9]+)(?P<timespan>[y|m|w|d])'  # noqa: E501
         match_dq = re.match(pattern_dq, string)
+        # test if the string matches the relative date pattern, add the parsed
+        # quantity to now in that case
         if match_dq is not None:
-            sign = match_dq.group(1)
-            quantity = match_dq.group(2)
-            timespan = match_dq.group(3)
+            sign = match_dq.group('sign')
+            quantity = match_dq.group('quantity')
+            timespan = match_dq.group('timespan')
             multiplier = -1 if sign == '-' else 1
             days = cls.relative[timespan]
             date = datetime.now() + multiplier * timedelta(
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -21,6 +21,7 @@ import re
 from operator import mul
 from beets import util
 from datetime import datetime, timedelta
+from dateutil.relativedelta import relativedelta
 import unicodedata
 from functools import reduce
 import six
@@ -533,8 +534,9 @@ class Period(object):
     instants of time during January 2014.
     """
 
-    precisions = ('year', 'month', 'day')
+    precisions = ('year', 'month', 'day', 'relative')
     date_formats = ('%Y', '%Y-%m', '%Y-%m-%d')
+    relative = ('y', 'm', 'w', 'd')
 
     def __init__(self, date, precision):
         """Create a period with the given date (a `datetime` object) and
@@ -550,19 +552,38 @@ class Period(object):
         """Parse a date and return a `Period` object or `None` if the
         string is empty.
         """
-        if not string:
-            return None
-        ordinal = string.count('-')
-        if ordinal >= len(cls.date_formats):
-            # Too many components.
-            return None
-        date_format = cls.date_formats[ordinal]
-        try:
-            date = datetime.strptime(string, date_format)
-        except ValueError:
-            # Parsing failed.
-            return None
-        precision = cls.precisions[ordinal]
+        if re.match('@([+|-]?)(\d+)([y|m|w|d])', string) is not None:
+            sign = re.match('@([+|-]?)(\d+)([y|m|w|d])', string).group(1)
+            quantity = re.match('@([+|-]?)(\d+)([y|m|w|d])', string).group(2)
+            timespan = re.match('@([+|-]?)(\d+)([y|m|w|d])', string).group(3)
+            if sign == '-':
+                m = -1
+            else:
+                m = 1
+            if timespan == 'y':
+                date = datetime.now() + m*relativedelta(years=int(quantity))
+            elif timespan == 'm':
+                date = datetime.now() + m*relativedelta(months=int(quantity))
+            elif timespan == 'w':
+                date = datetime.now() + m*relativedelta(weeks=int(quantity))
+            elif timespan == 'd':
+                date = datetime.now() + m*relativedelta(days=int(quantity))
+
+            precision = 'relative'
+        else:
+            if not string:
+                return None
+            ordinal = string.count('-')
+            if ordinal >= len(cls.date_formats):
+                # Too many components.
+                return None
+            date_format = cls.date_formats[ordinal]
+            try:
+                date = datetime.strptime(string, date_format)
+            except ValueError:
+                # Parsing failed.
+                return None
+            precision = cls.precisions[ordinal]
         return cls(date, precision)
 
     def open_right_endpoint(self):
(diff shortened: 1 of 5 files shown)


sampsyo:
Awesome; thank you for implementing this! I really like this idea.

I noticed that this implementation adds a new dependency for beets. Can you describe a little more detail about how you used `dateutil`? Maybe there's an easy way to use the stdlib for this purpose.

euri10:
well the way I thought it would be simple to implement is the following, I discover the library so maybe I'm wrong :)

DateQuery does in its dunder init  `start, end = _parse_periods(pattern)` that returns 2 Period objects that will ultimately be used in the query.

So I "hacked" the parse method of the Period object so that: if it matches the `@` its a relative date, if not it works as before.
If it's a relative date, it parses a sign, quantity and timespan and adds it to the datetime.now() 

```
        if re.match('@([+|-]?)(\d+)([y|m|w|d])', string) is not None:
            sign = re.match('@([+|-]?)(\d+)([y|m|w|d])', string).group(1)
            quantity = re.match('@([+|-]?)(\d+)([y|m|w|d])', string).group(2)
            timespan = re.match('@([+|-]?)(\d+)([y|m|w|d])', string).group(3)
            if sign == '-':
                m = -1
            else:
                m = 1
            if timespan == 'y':
                date = datetime.now() + m * relativedelta(years=int(quantity))
            elif timespan == 'm':
                date = datetime.now() + m * relativedelta(months=int(quantity))
            elif timespan == 'w':
                date = datetime.now() + m * relativedelta(weeks=int(quantity))
            elif timespan == 'd':
                date = datetime.now() + m * relativedelta(days=int(quantity))

            precision = 'relative'
```
So dateutil is used to easily add a given timespan / quantity to now(), is it feasible without it, I don't know, I can try to google it to check.

euri10:
I added a solution without dateutil, there might be some side effects because removing 30 days from now isnt strictly the same as removing one month, but maybe that's sufficient, no hard feelings.

sampsyo:
Great! It looks like it wasn't too tricky to use standard `datetime` functionality. Cool!

I'll do a full code review now.

beets/dbcore/query.py
```diff
@@ -550,19 +551,38 @@ def parse(cls, string):
...
-            # Too many components.
-            return None
-        date_format = cls.date_formats[ordinal]
-        try:
-            date = datetime.strptime(string, date_format)
-        except ValueError:
-            # Parsing failed.
-            return None
-        precision = cls.precisions[ordinal]
+        pattern_dq = '@([+|-]?)(\d+)([y|m|w|d])'
```
sampsyo: Some comments here (and below) would be *really* helpful. Specifically, at this point, it would be great to note that we are making a decision about whether the period seems to be relative or absolute.

beets/dbcore/query.py
```diff
@@ -550,19 +551,38 @@ def parse(cls, string):
...
+        if re.match(pattern_dq, string) is not None:
+            sign = re.match(pattern_dq, string).group(1)
+            quantity = re.match(pattern_dq, string).group(2)
+            timespan = re.match(pattern_dq, string).group(3)
```
sampsyo: You can save the result of `re.match(pattern_dq, string)` in a variable and reuse it four times.

beets/dbcore/query.py
```diff
@@ -550,19 +551,38 @@ def parse(cls, string):
...
+            if sign == '-':
+                m = -1
+            else:
+                m = 1
```
sampsyo: This is a case where an inline condition might be more readable: `m = -1 if sign == '-' else 1`. Also, a more descriptive variable name than `m` might be in order.

beets/dbcore/query.py
```diff
@@ -550,19 +551,38 @@ def parse(cls, string):
...
+            if timespan == 'y':
+                date = datetime.now() + m * timedelta(days=int(quantity) * 365)
+            elif timespan == 'm':
+                date = datetime.now() + m * timedelta(days=int(quantity) * 30)
+            elif timespan == 'w':
+                date = datetime.now() + m * timedelta(days=int(quantity) * 7)
+            elif timespan == 'd':
+                date = datetime.now() + m * timedelta(days=int(quantity))
```
sampsyo: A lookup table might make this code more maintainable. For example:

```py
UNIT_DAYS = {'y': 365, 'm': 30, 'w': 7, 'd': 1}
# ...
days = UNIT_DAYS[timespan]
date = datetime.now() + m * timedelta(days=int(quantity) * days)
```

docs/reference/query.rst
```diff
@@ -162,6 +162,20 @@ Dates are written separated by hyphens, like ``year-month-day``, but the month
...
 
+You can also use relative dates to the current time.
+Relative dates begin by the ``@`` character, followed by an optional ``+`` or
+``-`` sign that will increment or decrement now, followed by the time quantity
+that will be represented as an integer followed by either ``d`` for days, ``w``
+for weeks, ``m`` for months and finally ``y`` for year.
+
+Here is an example that finds all the albums added between now and last week::
+
+    $ beet ls -a 'added:@-1w..@0d'
```
sampsyo: Would `added:@-1w..` do the same?

beets/dbcore/query.py
euri10: Let me know what you think, I added it in the Period class doc

beets/dbcore/query.py
euri10: done

beets/dbcore/query.py
euri10: done

beets/dbcore/query.py
euri10: done, I used the relative tuple that already existed, just "upgraded" it to a dict

docs/reference/query.rst
euri10: it is the same, it's also tested if I didn't screw tests, fixed in doc

docs/reference/query.rst
```diff
@@ -162,6 +162,20 @@ Dates are written separated by hyphens, like ``year-month-day``, but the month
...
+Relative dates begin by the ``@`` character, followed by an optional ``+`` or
+``-`` sign that will increment or decrement now, followed by the time quantity
+that will be represented as an integer followed by either ``d`` for days, ``w``
+for weeks, ``m`` for months and finally ``y`` for year.
+
+Here is an example that finds all the albums added between now and last week::
+
+    $ beet ls -a 'added:@-1w..'
+
+Find all items added in a 2 weeks period 4  weeks ago::
```
Freso: A space too much.

docs/reference/query.rst
```diff
@@ -162,6 +162,20 @@ Dates are written separated by hyphens, like ``year-month-day``, but the month
...
+
+    $ beet ls -a 'added:@-6w..@-2w'
```
Freso: If I'm not misunderstanding the syntax, this should be `added:@-6w..@-4w`… ? The current example looks for things added "6 weeks ago .. 2 week ago", so using the description above the example, «Find all items added in a 4 weeks period 2 weeks ago».

Either the example or the description of the example should be fixed.

docs/reference/query.rst
```diff
@@ -186,6 +200,7 @@ Find all items with a file modification time between 2008-12-01 and
 
     $ beet ls 'mtime:2008-12-01..2008-12-02'
 
+
```
Freso: This seems/random/arbitrary and entirely unrelated to the PR… ?

setup.py
```diff
@@ -106,7 +106,7 @@ def build_manpages():
         'pyxdg',
         'pathlib',
         'python-mpd2',
-        'discogs-client'
+        'discogs-client',
```
Freso: This seems/random/arbitrary and entirely unrelated to the PR… ?

beets/dbcore/query.py
```diff
@@ -549,20 +550,45 @@ def __init__(self, date, precision):
...
+        or '%Y-%m-%d'
+        A relative date begins by '@ 'and has to follow the pattern_dq format
+        '@([+|-]?)(\d+)([y|m|w|d])'
+        - '@' indicates it's a date relative to now()
+        - the optional '+' or '-' sign, which defaults to '+' will increment or
+        decrement now() by a certain quantity
+        - that quantity can be expressed in days, weeks, months or years
+        respectively 'd', 'w', 'm', 'y'
+        Please note that this relative calculation is rather approximate as it
+        makes the assumption of 30 days per month and 365 days per year
```
Freso: These assumptions should probably be mentioned in the "human readable" docs (query.rst) as well. Most endusers will not read in-code documentation for details on the implementation.

docs/reference/query.rst
euri10: I hope the change I made satisfies your request, I'm a total newb on .rst format and sadly couldn't find a way in pycharm to get a wysiwyg, is there a way ?

docs/reference/query.rst
euri10: you're absolutely right, corrected

docs/reference/query.rst
euri10: indeed

setup.py
euri10: indeed 2

beets/dbcore/query.py
euri10: added

docs/reference/query.rst
```diff
@@ -163,18 +163,24 @@ and day are optional. If you leave out the day, for example, you will get
 matches for the whole month.
 
 You can also use relative dates to the current time.
-Relative dates begin by the ``@`` character, followed by an optional ``+`` or
-``-`` sign that will increment or decrement now, followed by the time quantity
-that will be represented as an integer followed by either ``d`` for days, ``w``
-for weeks, ``m`` for months and finally ``y`` for year.
+A relative date begins by ``@`` and has to follow the pattern_dq format
```
Freso: s/begins by/begins with an/

docs/reference/query.rst
```diff
@@ -163,18 +163,24 @@ and day are optional. If you leave out the day, for example, you will get
...
+``@([+|-]?)(\d+)([y|m|w|d])``
```
Freso: This is a very unwieldy piece of information. Maybe something more like "… and looks like ``@+3y``." E.g., give an example of the pattern instead of the regex definition of what a pattern has to adhere to («show, don't tell»). This is end-user documentation, not developer documentation. End users should not be expected to be able to read regular expressions. Besides, you explain (almost) each part of the pattern excellently below.

docs/reference/query.rst
```diff
@@ -163,18 +163,24 @@ and day are optional. If you leave out the day, for example, you will get
...
+- ``@`` indicates it's a date relative to now()
```
Freso: Use "now" instead of "now()". Again. This is end-user documentation, not developer documentation.

docs/reference/query.rst
```diff
@@ -163,18 +163,24 @@ and day are optional. If you leave out the day, for example, you will get
...
+- the optional ``+`` or ``-`` sign, which defaults to ``+`` will increment or
```
Freso: Maybe remove the leading "the "?

docs/reference/query.rst
```diff
@@ -163,18 +163,24 @@ and day are optional. If you leave out the day, for example, you will get
...
+decrement now() by a certain quantity
```
Freso: "increment or decrement now()" sounds heavily technical to me. Is there a way this can be described for someone less technically minded? "``+`` looks into the future, ``-`` looks into the past"? (Don't use this directly.)

docs/reference/query.rst
```diff
@@ -163,18 +163,24 @@ and day are optional. If you leave out the day, for example, you will get
...
+- that quantity can be expressed in days, weeks, months or years respectively
```
Freso: Maybe, just for completeness' sake (and in line with a previous comment), maybe add a bullet point for the numbers so each of 4 parts of the relative date is explained?

docs/reference/query.rst
```diff
@@ -163,18 +163,24 @@ and day are optional. If you leave out the day, for example, you will get
...
+'d', 'w', 'm', 'y'
```
Freso: Wrap these in ``'s.

docs/reference/query.rst
```diff
@@ -163,18 +163,24 @@ and day are optional. If you leave out the day, for example, you will get
...
+
+Please note that this relative calculation is rather approximate as it makes
+the assumption of 30 days per month and 365 days per year.
```
Freso: "Please note that the relative calculation always equates a month to 30 days and a year to 365 days, so it may sometimes be off the mark."

Or something like that. Idk. Your text still seems a bit too stiff and technical. Mine can certainly be improved upon as well. :)

(Feel free to disregard this comment. This tangentiates bikeshedding.)

beets/dbcore/query.py
irskep: Seems like these assumptions shouldn't be hard to change. You know exactly what the correct dates are as soon as you've parsed the query.

irskep: (i.e. it should just do the right thing and not require documenting assumptions)

beets/dbcore/query.py
irskep: Might try using named groups instead of positional. That should make this block more self-documenting.

But some inline "what is this trying to do" comments would still be helpful.

sampsyo:
This is all looking great; thank you for the continued iteration!

There's one last (important) design decision I wanted to ask about. Do you think the @ delimiter is necessary? That is, would this syntax be unambiguously parse-able?

```
$ beet ls added:-4w..+4w
```

as opposed to the current syntax:

```
$ beet ls added:@-4w..@+4w
```

Equivalently, I guess I'm asking whether the - and + characters are enough to determine that a period endpoint is relative. Does that make sense? It could make these queries a little easier to type.

euri10:
the `@` doesn't seem to be necessary, I put it in the first place because I was reading https://github.com/beetbox/beets/issues/1147
now I can remove it, put some named groups like @irskep suggested, and change the tests accordingly, seems all good.
let me know which one you prefer
```
        pattern_dq = '(?P<sign>[+|-]?)(?P<quantity>\d+)(?P<timespan>[y|m|w|d])'
        match_dq = re.match(pattern_dq, string)
        # test if the string matches the relative date pattern, add the 
        # parsed quantity to now in that case
        if match_dq is not None:
            sign = match_dq.group('sign')
            quantity = match_dq.group('quantity')
            timespan = match_dq.group('timespan')
            multiplier = -1 if sign == '-' else 1
            days = cls.relative[timespan]
            date = datetime.now() + multiplier * timedelta(
                days=int(quantity) * days)
            precision = 'relative'
```

sampsyo:
Oh, cool! I now see where the inspiration came from. I like your syntax with -1w better than my original @1w proposal. :smiley:

Those regex group names look good from here!

sampsyo:
Hi, @euri10—I'm really sorry for the long delay in getting around to a thorough review here! Things are looking good, but in the intervening time there's unfortunately been some changes to the same code, so we need to resolve some conflicts. (Specifically, we recently added more precise date-based queries that can include a time part.) Is there any chance you could help out by bringing the PR up to date?

euri10:
hello @sampsyo :)
I fixed the stuff, but I %$£@£" it up somehow ! I merged my relativedate branch into my master, and relativedate is the one asking for this PR, so I dunno what to do !
The build passes all tests (https://travis-ci.org/euri10/beets/builds/243157894)

Is there a way to change the PR somehow and make it point to my master branch instead ?

sampsyo:
Ah; it happens! The thing to do is just to open a new PR, which I've done at #2597.

sampsyo:
J/K; sorry! I had to pull your changes into a branch on the mothership so I could make my own changes. Please see #2598 instead. 😳 I'm adding you to the org so you can edit that branch too.

Issue #1999, opened 2016-05-11
embedart: Prompt for confirmation in command


taylorthurlow:
Most commands will prompt for changes when altering files as such:

```
$> beet modify -a album:"penguin prison" year=2011 month=10 day=18
Modifying 1 albums.
Penguin Prison - Penguin Prison
    month: 08 -> 10
    day: 00 -> 18
Really modify, move and write tags (Y/n)?
```

However, the current implementation of `embedart` has no such prompts - nor does it provide any sort of output as to which files/albums it is currently editing (at least usually). A simple confirmation prompt to confirm that the proposed changes are desired would be much more ideal:

```
$> beet embedart -f artwork.jpg album:"penguin prison"
Modifying 1 albums.
Penguin Prison - Penguin Prison
    Existing embedded artwork (300x294)
    Replacement artwork (800x800)
Really write artwork to files (Y/n)?
```

More than once I have fudged up the syntax or accidentally pressed enter when trying to insert a quote. It is a bit too eager to try and write the input artwork to every file in your library!

Thanks for `beets` by the way - it's absolutely invaluable.

sampsyo:
Sounds like a good idea.

taylorthurlow:
Per #2422 this looks like it's solved. Cheers to everyone involved.

PR #2422, opened 2017-01-31
Embedart plugin asks for confirmation before making changes to item’s…

Stunner:
… artwork. Addresses issue #1999

### DIFF
--- a/beetsplug/embedart.py
+++ b/beetsplug/embedart.py
@@ -27,7 +27,7 @@ from beets import config
 from beets import art
 
 
-def _confirmation(items, opts):
+def _confirmation(objs, opts):
     """Show the list of affected objects (items or albums) and confirm
     that the user wants to modify their artwork.
     """
@@ -42,11 +42,11 @@ def _confirmation(items, opts):
             fmt = u'$albumartist - $album - $title'
             istr = u'file'
         prompt = u'Modify artwork for %i %s%s (y/n)?' % \
-                 (len(items), istr, 's' if len(items) > 1 else '')
+                 (len(objs), istr, 's' if len(objs) > 1 else '')
 
-        # Show all the items.
-        for item in items:
-            print_(format(item, fmt))
+        # Show all the items or albums.
+        for obj in objs:
+            print_(format(obj, fmt))
 
         # Confirm with user.
         if not ui.input_yn(prompt, True):
--- a/beetsplug/embedart.py
+++ b/beetsplug/embedart.py
@@ -95,7 +95,7 @@ class EmbedCoverArtPlugin(BeetsPlugin):
                 # Confirm with user.
                 fmt = u'$albumartist - $album - $title'
                 prompt = u'Modify artwork for %i file%s (y/n)?' % \
-                             (len(items), 's' if len(items) > 1 else '')
+                         (len(items), 's' if len(items) > 1 else '')
                 if not _confirmation(items, opts, fmt, prompt):
                     return
 
--- a/beetsplug/embedart.py
+++ b/beetsplug/embedart.py
@@ -27,6 +27,33 @@ from beets import config
 from beets import art
 
 
+def _confirmation(items, opts):
+    """Show the list of affected objects (items or albums) and confirm
+    that the user wants to modify their artwork.
+    """
+    # Confirm artwork changes to library items.
+    if not opts.yes:
+        # Prepare confirmation with user.
+        print_()
+
+        fmt = u'$albumartist - $album'
+        istr = u'album'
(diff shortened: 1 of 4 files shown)


Stunner:
I realize that the tests need to be updated in order for this commit to pass the CI tests, however I am not familiar with that aspect in python. If you require me to modify/make tests for this change, could you kindly direct me to some documentation on how to get started with the testing suite you are using? Thanks.

beetsplug/embedart.py
```diff
@@ -60,22 +60,59 @@ def commands(self):
...
+                    # Prepare confirmation with user.
+                    print_()
+
+                    # Show all the items.
+                    for item in items:
+                        print_(format(item, fmt))
+
+                    # Confirm with user.
+                    if not ui.input_yn(prompt, True):
+                        return False
```
sampsyo: This looks like the same code as in `remove_items`. To avoid copying and pasting, we can move this code to a function and reuse it.

beetsplug/embedart.py
```diff
@@ -60,22 +60,59 @@ def commands(self):
...
             else:
-                for album in lib.albums(decargs(args)):
+                items = lib.albums(decargs(args))
+
+                # Confirm with user.
+                fmt = u'$albumartist - $album'
+                prompt = u'Modify artwork for %i album%s (y/n)?' % \
+                         (len(items), 's' if len(items) > 1 else '')
+                if not _confirmation(items, opts, fmt, prompt):
+                    return
```
sampsyo: This also looks like the same code twice, in both the `if` branch and the `else` branch, which means it's a good candidate for moving to a reusable function.

beetsplug/embedart.py
Stunner: There are similarities here, sure, but there is a subtle difference of return returning False instead of nothing. I don't know how to tidy this up any further. If you wouldn't mind doing it so that I can see how it should be done for next time that would be great!

beetsplug/embedart.py
Stunner: Funny you mention that here, but that is exactly why I created the _confirmation method. They are very similar, sure but there are subtle differences (items vs. albums for instance and differences in the prompt and format strings. Like the comment above, I really don't know how I would increase the reusability here.

Stunner:
Seems that the Travis CI test failed on 1 version of python due to an import problem. Let me know if I need to tweak anything in order to get it to pass.

sampsyo:
Hey! That's our PEP8 style checker. You can see its recommendations on Travis:
https://travis-ci.org/beetbox/beets/jobs/197270594#L848-L849

sampsyo:
Looks like the style checker passes now; thanks!

Would you mind also adding a changelog entry and some descriptive text to the documentation? The changelog is at `docs/changelog.rst` and the plugin's documentation is at `docs/plugins/embedart.rst`.

Stunner:
@sampsyo I have made some changes addressing your requested changes. Please review at your earliest convenience. Thanks.

sampsyo:
Sorry I've been slow to get to this! I want to make a couple of changes before merging that will deduplicate some similar code. I'll add those commits as soon as I can.

Stunner:
Not a problem. I am waiting on this to be merged before I tackle #2427. Or would you prefer that I tackle that in this PR as well?

sampsyo:
Hmm, good point—let's wait for the refactoring before tackling that one. But I swear I'll do it very soon. :smiley:

Stunner:
Cool. I'll standby for your changes before adding anything else. (Full Disclosure: there is a lot i'd like to help out with ;) )

sampsyo:
Hey, I'm really sorry it took me so long to get around to this! I've finally made a few minor changes to simplify things. Does this look OK? If so, we can merge this in time for the upcoming point release.

Stunner:
Not a problem! I totally understand, you are a busy prof. and am amazed that you are still doing such a stellar job keeping this project going strong. Learned a whole lot on the style you prefer for this project.

Looked at all the commits and they look good to me. Only exception is the simplification of the tests to all include the -y flag. From a comprehensive testing perspective, wouldn't it be wise to test both manual input (`self.io.addinput('y')`) as well as using the -y flag? That is why for one case I specified the -y flag in order to change it up, but maybe we should make `self.io.addinput('y')` the exception rather than the rule?

sampsyo:
Aww; thanks!

And sure; that's a good point! One test where we use the input style would make sense. Let's just make that a new test and keep it separate from the tests for other things.

sampsyo:
Looks perfect. Thanks for adding those tests.

Issue #2430, opened 2017-02-08
badfiles: Continue to other files if validator command is not found or exits with an error


karpinski:
The execution exits with an error if mp3val or flac validators aren't installed. On the other hand, for extensions other than mp3 or flac which aren't included in the plugin's config, information is logged that the validator is missing and the execution moves on to checking the rest of the files.

I think this behaviour should be consistent for all extensions, also in cases when a validator is specified in the plugin's config, but it's not actually installed.

I am happy to send a PR with a fix for this!

sampsyo:
Good call! Yes, this is definitely how it should work—it should use `self._log.error()` to produce an error message and just continue on instead of halting the whole process.

Thanks for looking into it!

mechanic9:
This issue should be closed

sampsyo:
Indeed; thanks!

Issue #1654, opened 2015-10-21
badfiles: silent mode


Kernald:
Running the `badfiles` plugin on an entire collection is pretty much unreadable for now. It would be great to have different levels of verbosity:
- display all
- display only bad files
- display nothing (just a return code).

sampsyo:
Sounds good. In fact, I'd argue that "show bad files only" should be the default mode.

Kolbasz12:
2 cents, if showing bad files only, does this mean files with "garbage at beginning/end of file"

This seems fairly common on some media so it could still be a lot of files.  I would argue to add option to fix files and then display only ones bad after a failed fix maybe.  I dont know, could be off topic for the OP

BeetsStudentTeam:
Hey guys, we would like to tackle this issue, if anyone is going to work on this soon let me know so we can look at other issues! We're planning to start working on this on Monday February 1st. Thanks! -Jake

fxthomas:
I was planning on doing this but I don't have much time to hack on beets at the moment. You're good to go!

SpirosChadoulos:
Greetings..I was thinking about tackling this issue but then I noticed tha some changes were made to the badfiles plugin. Does this mean this issue should be closed or shall I tackle it?

sampsyo:
Oh yeah, thanks! Looks like this was fixed in #2434.

PR #2432, opened 2017-02-08
A few changes to the badfiles plugin

karpinski:
Adding a verbosity option per request from #1654.

Process now continues to other files instead of stopping after a checker error (#2430).

The checker errors are logged only once for each file extension by default and for each file separately in the verbose mode. Let me know if this is fine or if you'd rather have just one of those.

### DIFF
--- a/beetsplug/badfiles.py
+++ b/beetsplug/badfiles.py
@@ -30,6 +30,16 @@ import sys
 import six
 
 
+class CheckerCommandException(Exception):
+    """Raised when running a Checker failed."""
+
+    def __init__(self, cmd, oserror):
+        self.checker = cmd[0]
+        self.path = cmd[-1]
+        self.errno = oserror.errno
+        self.msg = str(oserror)
+
+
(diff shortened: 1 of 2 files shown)


sampsyo:
Thanks for getting this started!

Is there any chance I could convince you to break this out into two separate PRs? The discussion here might get a little bit messy otherwise. It's possible we can resolve everything easily, of course, but if not, it might be nice to have two separate discussions.

For the verbosity mode: beets already has a global verbosity mode, which affects the logging level. Can we just use that judiciously instead of introducing a new flag? That would just mean sending everything to `debug` output except for the important messages.

For the new exception handling, throwing a new exception is fine. But it seems like an easier implementation here would just log a message and continue—that would be a bit simpler than building up a list of missing checkers. I know that will make the output more verbose, but I also think it will be slightly less surprising to use. Does that make sense?

karpinski:
Thanks for the feedback, @sampsyo! I've opened new PRs for those separate changes.

PR #2434, opened 2017-02-09
badfiles: displaying results only for corrupted files by default

karpinski:
Per request from #1654.

As a continuation to the discussion on the change: I think it would make sense to have a separate option for displaying all checker results (including the positive ones) with only bad files being displayed by default. The reason for having that, instead of logging the results for the uncorrupted files at the debug level, is that using this option the user would be able to see the results for all of the files without seeing a bunch of other debug messages, making the output less messy. The option could be named --all or --show-all, in order to be distinguished from the global --verbose.

This is just a suggestion, though. Let me know if you'd rather have the positive checker results be logged on the debug level, instead of introducing this new option.

### DIFF
--- a/beetsplug/badfiles.py
+++ b/beetsplug/badfiles.py
@@ -113,11 +113,16 @@ class BadFiles(BeetsPlugin):
                           .format(ui.colorize('text_warning', dpath), errors))
                 for line in output:
                     ui.print_(u"  {}".format(displayable_path(line)))
-            else:
+            elif opts.verbose:
                 ui.print_(u"{}: ok".format(ui.colorize('text_success', dpath)))
 
     def commands(self):
         bad_command = Subcommand('bad',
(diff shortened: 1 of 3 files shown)


jackwilsdon:
I like this idea, as `badfiles` creates quite a lot of unnecessary noise at the moment. Travis doesn't like this change as it looks like there's some left over whitespace:

    beetsplug/badfiles.py:123:33: W291 trailing whitespace

It might be worth documenting this in the changelog too, but it's up to @sampsyo in the end 😁.

sampsyo:
Sure; this looks good! I think just putting the "ok" lines behind the verbose flag will work well.

As @jackwilsdon says, a changelog entry would be great! With that and the whitespace fixed, this should be good to go.

karpinski:
Done :)

jackwilsdon:
Thank you! 🎉

PR #2445, opened 2017-02-19
Add option to hardlink when importing

jacobwgillespie:
This PR adds an option to hardlink files when importing, similar to the existing `link` option that enables symlinks.  Tests in the same style as the symlink tests are included.

```yaml
import:
    write: no
    hardlink: yes
```

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -478,16 +478,15 @@ def move(path, dest, replace=False):
 def link(path, dest, replace=False):
     """Create a symbolic link from path to `dest`. Raises an OSError if
     `dest` already exists, unless `replace` is True. Does nothing if
-    `path` == `dest`."""
-    if (samefile(path, dest)):
+    `path` == `dest`.
+    """
+    if samefile(path, dest):
         return
 
-    path = syspath(path)
-    dest = syspath(dest)
-    if os.path.exists(dest) and not replace:
+    if os.path.exists(syspath(dest)) and not replace:
         raise FilesystemError(u'file exists', 'rename', (path, dest))
     try:
-        os.symlink(path, dest)
+        os.symlink(syspath(path), syspath(dest))
     except NotImplementedError:
         # raised on python >= 3.2 and Windows versions before Vista
         raise FilesystemError(u'OS does not support symbolic links.'
@@ -504,16 +503,15 @@ def link(path, dest, replace=False):
 def hardlink(path, dest, replace=False):
     """Create a hard link from path to `dest`. Raises an OSError if
     `dest` already exists, unless `replace` is True. Does nothing if
-    `path` == `dest`."""
-    if (samefile(path, dest)):
+    `path` == `dest`.
+    """
+    if samefile(path, dest):
         return
 
-    path = syspath(path)
-    dest = syspath(dest)
-    if os.path.exists(dest) and not replace:
+    if os.path.exists(syspath(dest)) and not replace:
         raise FilesystemError(u'file exists', 'rename', (path, dest))
     try:
-        os.link(path, dest)
+        os.link(syspath(path), syspath(dest))
     except NotImplementedError:
         raise FilesystemError(u'OS does not support hard links.'
                               'link', (path, dest), traceback.format_exc())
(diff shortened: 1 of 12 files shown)


beets/util/__init__.py
```diff
@@ -500,6 +500,32 @@ def link(path, dest, replace=False):
...
+                              'link', (path, dest), traceback.format_exc())
+    except OSError as exc:
+        # TODO: Windows version checks can be removed for python 3
+        if hasattr('sys', 'getwindowsversion'):
+            if sys.getwindowsversion()[0] < 6:  # is before Vista
+                exc = u'OS does not support hard links.'
+        raise FilesystemError(exc, 'link', (path, dest),
+                              traceback.format_exc())
+
+
```
jrobeson: this part about windows version less than 6 needs to go. I don't think they are supported on windows at all, on any version.

jrobeson:
I think we also need to add a check to make sure both the source and the target are on the same device, otherwise the hardlink will fail.  Simply copying what symlinks do is not enough.

jrobeson:
Writing a proper test for this will probaby require creating a fake device using tmpfs.

jacobwgillespie:
I added a catch for the cross-device hardlink error so that the error's a bit prettier, however I'm not sure how to handle the test.  tmpfs would indeed work, but it doesn't exist on OS X.  Perhaps the implementation is simple enough at this point that we could get away with a cross-device hardlink test, we'd essentially be testing the Python implementation of `os.link`.

beets/util/__init__.py
```diff
@@ -500,6 +501,31 @@ def link(path, dest, replace=False):
                               traceback.format_exc())
 
 
+def hardlink(path, dest, replace=False):
+    """Create a hard link from path to `dest`. Raises an OSError if
+    `dest` already exists, unless `replace` is True. Does nothing if
+    `path` == `dest`."""
+    if (samefile(path, dest)):
```
sampsyo: No parentheses are necessary around the function call.

beets/util/__init__.py
```diff
@@ -500,6 +501,31 @@ def link(path, dest, replace=False):
                               traceback.format_exc())
 
 
+def hardlink(path, dest, replace=False):
+    """Create a hard link from path to `dest`. Raises an OSError if
+    `dest` already exists, unless `replace` is True. Does nothing if
+    `path` == `dest`."""
```
sampsyo: Please place the closing """ on a line by itself (that's PEP8 style).

beets/util/__init__.py
```diff
@@ -500,6 +501,31 @@ def link(path, dest, replace=False):
...
+    if (samefile(path, dest)):
+        return
+
+    path = syspath(path)
+    dest = syspath(dest)
```
sampsyo: It might seem like a shortcut, but doing this once and overwriting the variables will cause problems: namely, the exceptions below use `path` and `dest`, which will now be system paths. This will produce strange results on Windows, where a ridiculous prefix needs to be added for `syspath` results.

docs/reference/config.rst
```diff
@@ -433,15 +433,30 @@ link
...
+
+Either ``yes`` or ``no``, indicating whether to use hard links instead of
+moving or copying or symlinking files. (It conflicts with the ``move``,
+``copy``, and ``link`` options.) Defaults to ``no``.
+
+This option only works on platforms that support hardlinks: i.e., Unixes.
+It will fail on Windows.
+
+It's likely that you'll also want to set ``write`` to ``no`` if you use this
+option to preserve the metadata on the linked files.
```
sampsyo: Instead of copying and pasting the above two paragraphs, can you please make this a little more human by saying something like:

> As with symbolic links (see :ref:`link`, above), this will not work on Windows and you'll want to set ``write`` to ``no``.

jacobwgillespie:
Thanks for the feedback - most of your suggestions applied to the original `link` method, so I updated that as well.

sampsyo:
Ah, great! I hadn't realized that we had the same issues just above in the symlink function. Thanks for fixing that too.

sampsyo:
Thank you again! This is all merged up. :sparkles:

jacobwgillespie:
Sweet, thanks for the fast response 🎉

PR #2447, opened 2017-02-21
discogs: support simple auth.

robbat2:
The official OAuth authentication seems to have broken, so allow usage
of simple configuration instead.

See-Also: https://github.com/discogs/discogs_client/issues/78
Signed-off-by: Robin H. Johnson <robbat2@gentoo.org>

### DIFF
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -54,9 +54,11 @@ class DiscogsPlugin(BeetsPlugin):
             'apisecret': 'plxtUTqoCzwxZpqdPysCwGuBSmZNdZVy',
             'tokenfile': 'discogs_token.json',
             'source_weight': 0.5,
+            'user_token': None,
         })
         self.config['apikey'].redact = True
         self.config['apisecret'].redact = True
+        self.config['user_token'].redact = True
         self.discogs_client = None
         self.register_listener('import_begin', self.setup)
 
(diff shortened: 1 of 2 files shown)


sampsyo:
Oh no, that's bad news. Any chance you could add some documentation in `discogs.rst` to explain what's going on (and how to get a user token)?

robbat2:
Now includes documentation.

sampsyo:
OK, thanks! This looks good, and the documentation is helpful too.

For what it's worth, OAuth-based authentication is still working fine on my machine. I'm not sure what's going wrong on your setup, but it can't hurt to have an alternative authentication mechanism.

robbat2:
What version of `requests` and `oauthlib` do you have locally? It felt like something was broken there when I was debugging it, that a bunch of keys weren't being included in the POST body anymore.

sampsyo:
Looks like that was with discogs-client 2.0.2 and requests 2.12.4:

```
>>> discogs_client.__version__
'2.0.2'
>>> requests.__version__
'2.12.4'
```

That seemed a bit of out of date, but I checked and I think discogs-client is just misreporting its version—pip thinks it's up to date. (This was on Python 3.6, by the way.)

robbat2:
I have requests-2.12.5 & discogs-client-2.2.1 (from pypi, the tag is missing in github); wondering if it's a regression in discogs-client then.

sampsyo:
For what it's worth, I just double-checked, and it seems like I do have 2.2.1—this issue suggests that they released a wrong value in that `__version__` variable. https://github.com/discogs/discogs_client/issues/64

Issue #2272, opened 2016-11-19
web: View album art


claman:
I'm not sure if this has been brought up before, but I really like being able to view album art. Is there a way to add that to the web interface, or possibly through a plugin?

sampsyo:
Seems like a good idea! We'd really like to move to a new architecture for the web plugin so it could be extensible in this way (c.f. #718), but if you're interested, there's also no reason we couldn't hack it onto the current web interface.

claman:
I don't think that I know enough python to be of much help here, but I'll take a look soon

claman:
This seems like a good repo to pull ideas from: https://github.com/tkem/mopidy-local-images

Has development on https://github.com/jtemplet/beets-web stalled?

irskep:
To answer your last question about running code in the repo, I typically create a virtualenv, and then run code like this:

```sh
python setup.py install && beet <cmd>
```

It's a little slow but you always get the freshest version of your code.

claman:
So I was (finally) poking around the code a bit more, and it looks like flask is exposing the albumart if the item has it available, but I can't figure out how to access it.

I'm looking at these lines:
```
271 @app.route('/album/<int:album_id>/art')
272 def album_art(album_id):
273     album = g.lib.get_album(album_id)
274     if album.artpath:
275         return flask.send_file(album.artpath)
276     else:
277         return flask.abort(404)
```

sampsyo:
Just use an image tag like `<img src="/album/(id)/art">`, given the id of the album. That should do it.

PR #2458, opened 2017-02-26
web: View album art

claman:
WIP based on #2272

### DIFF
--- a/beetsplug/web/templates/index.html
+++ b/beetsplug/web/templates/index.html
@@ -80,6 +80,12 @@
                         <a target="_blank" href="http://musicbrainz.org/recording/<%= mb_trackid %>">view</a>
                     </dd>
                 <% } %>
+                <% if (art) { %>
+                    <dt>Art</dt>
+                    <dd>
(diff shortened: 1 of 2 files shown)


beetsplug/web/templates/index.html
```diff
@@ -80,6 +80,12 @@
                         <a target="_blank" href="http://musicbrainz.org/recording/<%= mb_trackid %>">view</a>
                     </dd>
                 <% } %>
+                <% if (art) { %>
```
claman: Unfortunately, this does not do what I think it should do

sampsyo: Hmm, you're right—not quite. The trouble is that `art` isn't an attribute on individual items because album art is associated per *album*. We'd need to check whether the item is associated with an album object that has an `artpath` property.

claman:
This is what currently comes up when I run `beet web`
web
I'm really not sure why the rest of the information isn't showing up

claman:
I'm going to go ahead and close this since it seems like other people are working on similar projects, and I don't really have time to make any meaningful progress.

Issue #2460, opened 2017-02-27
Documentation and beets.reg 


mskadu:
To start with, I am a windows user looking to install beets.

This is about 
1. contents of /master/extra/beets.reg 
2. And your excellent "Getting Started" guide, subsection "Installing for Windows" ([link](http://beets.readthedocs.io/en/v1.4.3/guides/main.html#installing-on-windows)). 

### Problem
The above-mentioned doc page mentions that I need Python 2.7. I spotted a related problem with contents the beets.reg file.

1. The path in beets.reg points to the default install path of Python 3.6 - This is in contrast with your doc page which says v2.7 is required
2. Anyone choosing to import and use the tool would run into issues since they will have Python v2.7 installed.  It might be worthwhile adding a friendly warning on the doc on the lines of "you must make sure the path within beets.reg needs to be reviewed and fixed to be in line with your own install".

Also, can I ask you to re-confirm that Python v2.7 is a strict requirement for Windows install? The section titled "Installing" starts by saying "Beets works on Python 2.7 and Python 3.4 or later."

Having said all of this, and given that this is my first interaction with the creator(s?) of the tool - kudos for the what looks like an exciting tool for managing my music files. I chuckled at the bit in your video intro which says its "for obsessive compulsive geeks who care about music". And special thanks for putting this up on github.

There's nothing better (to me) than a geek with a sense of humour :)

sampsyo:
Hey, thanks for pointing this out! Our recommended Python version is now indeed 3.6, and beets.reg does indeed assume that. Our docs were a tad out of date; I updated them now.

I also like the idea of adding a note about needing to edit the path if you have a different Python version/install directory. Is there a chance I can convince you to make the edit in `main.rst` and open a pull request? (This is as easy as hitting the "edit" button right here on GitHub.)

I'm also glad you liked the video. :smiley: It's due for an update, but we'll try to preserve the jokes as-is.

mskadu:
I have given it a shot. And have also made some slight 3.6 related text improvements, while I was at it. Do take a look and let me know.

sampsyo:
Addressed via #2461. Thank you!

PR #2461, opened 2017-02-28
Updates to main.rst - Python 3.6 and Windows

mskadu:
I have made some further changes to make it clearer that Python 3.6 is recommended and that v2.7 is the oldest tested version. And as discussed in #2460 I have made some Windows specific revisions aimed at making this friendlier for fellow community members :)

Feedback welcome!

### DIFF
--- a/docs/guides/main.rst
+++ b/docs/guides/main.rst
@@ -9,10 +9,9 @@ collection better.
 Installing
 ----------
 
-You will need Python.
-Beets works on `Python 2.7`_ and Python 3.4 or later.
+You will need `Python 3.6`_. But Beet also works with Python 2.7 and later.
 
-.. _Python 2.7: http://www.python.org/download/
+.. _Python 3.6: http://www.python.org/download/
 
 * **macOS** v10.7 (Lion) and later include Python 2.7 out of the box.
   You can opt for Python 3 by installing it via `Homebrew`_:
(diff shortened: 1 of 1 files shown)


docs/guides/main.rst
```diff
@@ -9,10 +9,9 @@ collection better.
 Installing
 ----------
 
-You will need Python.
-Beets works on `Python 2.7`_ and Python 3.4 or later.
+You will need `Python 3.6`_. But Beet also works with Python 2.7 and later.
```
sampsyo: Python 3.4 actually works fine. 3.6 is just what we recommend, especially for Windows.

docs/guides/main.rst
```diff
@@ -46,6 +45,8 @@ Beets works on `Python 2.7`_ and Python 3.4 or later.
 
 * On **NixOS**, run ``nix-env -i beets``.
 
+* On most versions of **Windows**, there is a `installer`_ available on the Python website.
```
sampsyo: In this list, we're pointing toward places where you can install *beets* from the package manager, not just Python.

docs/guides/main.rst
```diff
@@ -139,6 +146,11 @@ place to start::
     directory: ~/music
     library: ~/data/musiclibrary.db
 
+For Windows user, this would look like::
+
+    directory: "D:\\My Music Files"
+    library: "D:\\My Music Files\\Musiclibrary.db"
```
sampsyo: The quotes and double-backslashes aren't actually necessary—they are harmless, but I think we can recommend that people don't need to do that, thanks to YAML's elegant syntax.

sampsyo: Also, for what it's worth, you can still use the unixy-looking paths shown just above: `~/music` will work just fine and use a directory called "music" in your `%HOME%` directory.

mskadu: IMHO, unixy paths are not natural to dominantly Windows users. And translating Windows paths to Unix format is even less so. When I read the guide, that was the first question that popped up in my mind. 

May I suggest we add an example that shows how paths with spaces in them are used? Or if they can be used as-is, worth explicitly clarifying that in the text?

sampsyo: Right; you can use Windowsy paths too—I just wanted to point out that either works fine. And spaces work fine too; no need to handle them any differently.

PR #2464, opened 2017-03-01
Updated main.rst - Updates to seeing your music

mskadu:
Modified "Seeing your Music" section to add explanation of the ``-f`` param of the list command. I strongly believe that first time users would very much like to start by overriding the default format of displaying results.

### DIFF
--- a/docs/guides/main.rst
+++ b/docs/guides/main.rst
@@ -230,11 +230,14 @@ songs.  Thus::
     The Magnetic Fields - Distortion - Three-Way
     The Magnetic Fields - Distortion - California Girls
     The Magnetic Fields - Distortion - Old Fools
+    
     $ beet ls hissing gronlandic
     of Montreal - Hissing Fauna, Are You the Destroyer? - Gronlandic Edit
+    
     $ beet ls bird
     The Knife - The Knife - Bird
     The Mae Shi - Terrorbird - Revelation Six
+    
     $ beet ls album:bird
     The Mae Shi - Terrorbird - Revelation Six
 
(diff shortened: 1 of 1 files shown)


docs/guides/main.rst
```diff
@@ -230,11 +230,14 @@ songs.  Thus::
...
+    
     $ beet ls hissing gronlandic
     of Montreal - Hissing Fauna, Are You the Destroyer? - Gronlandic Edit
+    
     $ beet ls bird
     The Knife - The Knife - Bird
     The Mae Shi - Terrorbird - Revelation Six
+    
     $ beet ls album:bird
     The Mae Shi - Terrorbird - Revelation Six
```
sampsyo: The lack of spaces here is intentional—that's what the actual terminal output looks like.

docs/guides/main.rst
```diff
@@ -245,16 +248,33 @@ put the field before the term, separated by a : character. So ``album:bird``
...
+     
+    As you can see, Beets has replaced the fields specified by the ``$`` prefix 
+    (e.g. $format, $year) with values for each item in the results. A full 
+    list of fields available for use can be found by running ``beet fields``. If
+    you'd like beet to use this format as default (without having to use ``-f``)
+    simply add it to your config file like this::
+    
+       format_item: "[$format] $album ($year) - $artist - $title"
+    
+    And yes, the enclosing double-quotes are necessary.
```
sampsyo: I like the idea of describing `-f` here, which is very powerful! But I'm concerned that the `format_item` config option here is getting a little too detailed. The "getting started" guide is meant to be very high-level, and we need to avoid describing too much of beets's (vast) set of configuration options.

Is there any chance you could try simplifying this down to just describe the `-f` option?

Also, it looks like these paragraphs need to be "out-dented" to avoid being parsed as code by ReST.

mskadu: Can do. how do you "out-dent"? I am not familiar with the term, unfortunately.

sampsyo: Sorry, that was my creative phrasing—I just mean the opposite of *indent* (i.e., remove the whitespace on the left).

docs/guides/main.rst
mskadu: Makes sense. Let's keep it that way.

docs/guides/main.rst
sampsyo: Using git, you can push more commits to the same branch to update. I assume you can do the same thing in the browser by switching the view to the right branch?

mskadu:
Err, for this one I forked your repo. And then created a PR from changes to
the head of my fork. In any case, will try and let you know.

--
Mayuresh

mskadu:
Right, I like how easy it was to do this. Given I was working on my own fork of beets - I see no reason to make a separate branch for edits to a single file. I will though if it starts to get wider than a single file.

Hope that's OK with you. And if you do prefer it otherwise, you only need to say the word :)

sampsyo:
Great; thanks! No, this is definitely the right strategy—there's no need for a separate branch. Crafting the exact contents of a pull request is an art, not a science, but generally it's totally OK to group together multiple commits to multiple files. The criterion is that each PR should contain one coherent, logical set of changes that should be merged (or not) as a group. That's fuzzy, of course, but you're on the right track. :smiley:

sampsyo:
OK, this is merged! Thank you again! :sparkles:

Issue #2413, opened 2017-01-24
[Enhancement] Nfo files


OmgImAlexis:
It'd be great to see an option for NFO files being written for artists and albums which would mean the MBID can be saved as well as other details. This helps with things like KODI, Plex, etc. and would also help with beets and updating music libraries even if the database file is deleted.

For an example of the NFO files have a look at what KODI accepts. Most other platforms that use NFO files for music either use the same format as KODI or an extended format which KODI's as a base. 

http://kodi.wiki/view/NFO_files/Music
http://kodi.wiki/view/Music_tagging (go down to "6 NFO files")

akyag:
+1000000000000
And there is no other software which can do this for music. And maybe in the future artist artwork, thumbs, backgrounds from https://fanart.tv/music-fanart/  for ex. can be fetched too?

jrobeson:
@akyag: there's already a fanart.tv backend for the fetchart plugin, so that part is done

jrobeson:
@akyag : sorry, i thought we did have fanart.tv already, but this implies not so https://github.com/beetbox/beets/issues/889

sampsyo:
Ah, so these are little XML files containing metadata? Seems like a fine request.

tigranl:
Hi, @sampsyo! Where can we store .nfo files? I think I can implement this feature.

sampsyo:
Cool! I actually don't know where Kodi expects them to be... perhaps in the album directory (`album.path` in beets)?

tigranl:
Another one question. I assume it should be a plugin?

sampsyo:
Yeah, totally—integrations with specific other software count as narrow enough use cases that they should be plugins.

bajosoto:
Hi @tigranl. I'm also interested in implementing this feature. If you have already started something in a repo would you mind sharing so maybe we can collaborate on this? Also, do you think this kind of feature would be more useful as a CLI command, or as part of the importing process?

bajosoto:
And to your question, it's two kinds of file. artist.nfo goes in the artist folder and album.nfo in the album one. Also, according to the wiki, "Both album.nfo and artist.nfo files can contain the MusicBrainz tag or simply contain a URL to the correct page on the MusicBrainz website". Maybe this would be a much simpler approach than generating the xml structure?

tigranl:
@bajosoto I just started, I'll make a PR soon. I think it would be more appropriate as part of the importing process.

tigranl:
[...]

PR #2473, opened 2017-03-13
Add kodinfo plugin

bajosoto:
Here's the plugin to generate the .nfo files for kodi. It generates one per album imported and one for the artist. It generates one for each singleton as well, although I didn't find support for this anywhere in the Kodi wiki, but perhaps for Plex and other systems.
Is there anything else I need to add like documentation, style guide, add test cases or anything similar to it?
@sampsyo, please let me know your comments! :squirrel:

### DIFF
--- a/beetsplug/kodinfo.py
+++ b/beetsplug/kodinfo.py
@@ -30,7 +30,7 @@ class KodiNfo(BeetsPlugin):
         self.register_listener('album_imported', self.makeAlbumNfo)
         self.register_listener('item_imported', self.makeItemNfo)
 
-        def makeAlbumNfo(self, lib, album):
+    def makeAlbumNfo(self, lib, album):
         linkAlbum = 'https://musicbrainz.org/release/{0}'
         linkArtist = 'https://musicbrainz.org/artist/{0}'
         with open(album.path + '/album.nfo', 'w') as f:
--- a/beetsplug/kodinfo.py
+++ b/beetsplug/kodinfo.py
@@ -25,20 +25,20 @@ from beets import library
 
 
 class KodiNfo(BeetsPlugin):
-  def __init__(self):
-    super(KodiNfo, self).__init__()
-    self.register_listener('album_imported', self.makeAlbumNfo)
-    self.register_listener('item_imported', self.makeItemNfo)
+    def __init__(self):
+        super(KodiNfo, self).__init__()
+        self.register_listener('album_imported', self.makeAlbumNfo)
+        self.register_listener('item_imported', self.makeItemNfo)
 
-  def makeAlbumNfo(self, lib, album):
-    linkAlbum = 'https://musicbrainz.org/release/{0}'
-    linkArtist = 'https://musicbrainz.org/artist/{0}'
-    with open(album.path + '/album.nfo', 'w') as f:
-		f.write(linkAlbum.format(album.mb_albumid))
-    with open(album.path + '/../artist.nfo', 'w') as f:
-        f.write(linkArtist.format(album.mb_albumartistid))
+        def makeAlbumNfo(self, lib, album):
+        linkAlbum = 'https://musicbrainz.org/release/{0}'
+        linkArtist = 'https://musicbrainz.org/artist/{0}'
+        with open(album.path + '/album.nfo', 'w') as f:
+            f.write(linkAlbum.format(album.mb_albumid))
+        with open(album.path + '/../artist.nfo', 'w') as f:
+            f.write(linkArtist.format(album.mb_albumartistid))
 
-  def makeItemNfo(self, lib, item):
-    link = 'https://musicbrainz.org/recording/{0}'
-    with open(item.path + '/' + item.title + '.nfo/', 'w') as f:
-		f.write(link.format(item.mb_trackid))
+    def makeItemNfo(self, lib, item):
+        link = 'https://musicbrainz.org/recording/{0}'
+        with open(item.path + '/' + item.title + '.nfo/', 'w') as f:
+            f.write(link.format(item.mb_trackid))
--- a/beetsplug/kodinfo.py
+++ b/beetsplug/kodinfo.py
@@ -20,7 +20,8 @@ and other details.
 For singletons, it generates one .nfo for each track.
 """
 
-from beets.plugins import BeetsPlugin
+from beets.plugins import BeetsPlugin 
+
 
 class KodiNfo(BeetsPlugin):
     def __init__(self):
--- a/beetsplug/kodinfo.py
+++ b/beetsplug/kodinfo.py
@@ -1,4 +1,3 @@
-
 # -*- coding: utf-8 -*-
 # This file is part of beets.
 # Copyright 2017, Sergio Soto.
@@ -20,7 +19,14 @@ and other details.
 For singletons, it generates one .nfo for each track.
 """
 
+from __future__ import absolute_import, division, print_function
+
+import os
 from beets.plugins import BeetsPlugin
+
+LINK_ALBUM = 'https://musicbrainz.org/release/{0}'
+LINK_ARTIST = 'https://musicbrainz.org/artist/{0}'
+LINK_TRACK = 'https://musicbrainz.org/recording/{0}'
 
 
 class KodiNfo(BeetsPlugin):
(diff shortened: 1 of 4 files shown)


bajosoto:
Issue #2413

bajosoto:
Oh wow... Travis is really picky. I guess I should have ran flake8 locally. Sorry for all these mini commits.

sampsyo:
No problem! Thank you for coping with our picky style checker. It really helps with long-term maintainability. :smiley:

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,43 @@
...
+class KodiNfo(BeetsPlugin):
+    def __init__(self):
+        super(KodiNfo, self).__init__()
+        self.register_listener('album_imported', self.make_album_nfo)
+        self.register_listener('item_imported', self.make_item_nfo)
+
+    def make_album_nfo(self, lib, album):
+        linkalbum = 'https://musicbrainz.org/release/{0}'
+        linkartist = 'https://musicbrainz.org/artist/{0}'
+        with open(album.path + '/album.nfo', 'w') as f:
```
sampsyo: To generate paths, it's generally safer to use the functions in Python's `os.path` instead of plain string concatenation. For example, here, use `os.path.join(album.path, 'album.nfo')`. This helps with cross-platform compatibility, for instance.

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,43 @@
...
+            f.write(linkalbum.format(album.mb_albumid))
+        with open(album.path + '/../artist.nfo', 'w') as f:
```
sampsyo: And this one can be `os.path.join(os.path.dirname(album.path), 'artist.nfo')`.

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,43 @@
...
+            f.write(linkartist.format(album.mb_albumartistid))
+
+    def make_item_nfo(self, lib, item):
+        link = 'https://musicbrainz.org/recording/{0}'
+        with open(item.path + '/' + item.title + '.nfo/', 'w') as f:
```
sampsyo: This one can use `os.path` too. Also, is it supposed to end in a slash?

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,43 @@
...
+
+class KodiNfo(BeetsPlugin):
+    def __init__(self):
+        super(KodiNfo, self).__init__()
+        self.register_listener('album_imported', self.make_album_nfo)
+        self.register_listener('item_imported', self.make_item_nfo)
+
+    def make_album_nfo(self, lib, album):
+        linkalbum = 'https://musicbrainz.org/release/{0}'
+        linkartist = 'https://musicbrainz.org/artist/{0}'
```
sampsyo: Since these never change, it's good style to write them as *constants*. You can hoist them to the top of the module and spell them in all caps, like this:

```py
LINK_ALBUM = 'https://musicbrainz.org/release/{0}'
LINK_ARTIST = 'https://musicbrainz.org/artist/{0}'
```

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,43 @@
+
```
sampsyo: This extra blank line doesn't need to be here.

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,43 @@
...
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+
+"""Genertes artist.nfo and album.nfo after the import process
+These are used by Kodi and other platforms to save the MBID
+and other details.
+For singletons, it generates one .nfo for each track.
+"""
+
```
sampsyo: Can you please add our standard `__future__` imports?

```py
from __future__ import absolute_import, division, print_function
```

tigranl:
@sampsyo hi! Is it possible to pull out Album or Item object from Lib? See: https://gist.github.com/tigranl/79652e7dfac1d6216c1c0a2d0bbf696f.

sampsyo:
@tigranl Yes! You just need to issue a query to get the Album or Item you're interested in. You use `lib.items(<query>)` or `lib.albums(<query>)`. Here's an example from another plugin, where the query comes from the command-line arguments: https://github.com/beetbox/beets/blob/5ed2798a76daa2c547f04538dfdd2e699e487bba/beetsplug/zero.py#L81-L82

bajosoto:
Thank you for your feedback @sampsyo! I'm learning a few tricks from all this :D. I'll get the documentation page done in a few.

tigranl:
So, what about using `import` listener with only one function for both albums and singletons?

bajosoto:
That could be a possibility too @tigranl! Although from what I understand either album_imported or item_imported is called during the import. I'm not sure if it would be more complicated to iterate over the whole list of items/albums imported using the `import` listener since we would have to do a query for each one of them, rather than just creating the .nfo files on the fly as items are imported.

From the [documentation](http://beets.readthedocs.io/en/v1.4.3/dev/plugins.html#listen-for-events):
> item_imported: called with an Item object every time the importer adds a singleton to the library (not called for full-album imports). Parameters: lib, item

sampsyo:
The two seem more or less equivalent to me—as long as we minimize the code duplication between the two cases, two event handlers vs. one handler doesn't seem to make a huge difference.

docs/plugins/index.rst
```diff
@@ -66,6 +66,7 @@ like this::
    inline
    ipfs
    keyfinder
+   kodinfo
```
sampsyo: Nice! Could you also add this to `plugins/index.rst` with a human-readable summary of the plugin?

docs/plugins/kodinfo.rst
```diff
@@ -0,0 +1,31 @@
...
+
+.. _files: http://kodi.wiki/view/NFO_files/music
+.. _KODI: http://kodi.tv/
+.. _Plex: https://www.plex.tv
+.. _requests: http://docs.python-requests.org/en/latest/
+
+Configuration
+-------------
+
+The ``kodi:`` section has no configurable options.
```
sampsyo: I think it's OK to just leave this section off.

docs/plugins/kodinfo.rst
```diff
@@ -0,0 +1,31 @@
...
+
+
+A note on folder structure configuration
+----------------------------------------
+
+``kodinfo`` will only work for configurations where the artist path is the direct parent of the album 
+paths.
```
sampsyo: And this sentence can just go in the main section. It might also be nice to explain why this is: beets doesn't know what directory corresponds to an artist, so the plugin just takes the guess that it's the parent directory of where the album is located.

docs/plugins/index.rst
bajosoto: Done! Added the description down below in the interoperability section. I hope it's the correct one.

docs/plugins/kodinfo.rst
bajosoto: Removed :D

docs/plugins/kodinfo.rst
bajosoto: Done as well

bajosoto:
Docs updated. I'm also new to rst files so it's been nice to learn all these things while contributing. :D I'm always surprised by how fast you are at responding. You're doing an awesome job! ⭐️

bajosoto:
BTW I did test the .nfo files in Kodi and they worked. What I didn't test is singletons because a) I don't have a singleton track on my computer atm and b) From what I read in the documentation, I *believe* Kodi doesn't support single-file .nfo files for music (it does for movies), but probably some of the other programs that can read Kodi's .nfo files do.

sampsyo:
This looks great! Thanks for all the updates and for the docs! :sparkles:

It would also be great to hear from @tigranl—does everything look in order to you?

And good to know about Kodi not supporting the single-file case. If that's true, maybe we should remove the support for generating them from this plugin too?

tigranl:
Yeah, everything is fine. Well done, @bajosto!
It seems that Kodi doesn't actually support nfo for singletones, since there is no information about it on their site.

peace899:
New version with a choice of XML file or text file containg URL only and also update library after import of albums [here](https://github.com/peace899/beets/blob/master/beetsplug/kodinfo.py).

As suggested from #2569.

jtpavlock:
closed as migrated to #2569

PR #2477, opened 2017-03-16
Issue#2277

abigaildelarosa:
Hi @sampsyo, what do you think about adding just a paragraph with a briefly description?  and about the "standard programming".  Normally programmers have to re-edit their code because of the travis testing, I understand the travis importance. I think it will be nice, if it will be possible to add a link to explain programmers, how they should write their code in order to be approved at the first glance by travis. Is just if its possible...

### DIFF
--- a/beetsplug/kodinfo.py
+++ b/beetsplug/kodinfo.py
@@ -30,7 +30,7 @@ class KodiNfo(BeetsPlugin):
         self.register_listener('album_imported', self.makeAlbumNfo)
         self.register_listener('item_imported', self.makeItemNfo)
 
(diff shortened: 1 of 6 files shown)


sampsyo:
Hmm; interesting proposal. I sort of like how the current version of the guide lets people get up and running without worrying too much about the concepts… maybe this would be a good notion to add to the *end* of the guide as an invitation to explore in more depth? I think the abstract concepts might make a bit more sense after you've already gotten your hands dirty with the basics.

About coding conventions: we actually have some standard advice here on the wiki! https://github.com/beetbox/beets/wiki/Hacking#coding-conventions Perhaps we should expand that?

Issue #2471, opened 2017-03-10
absubmit: Do not filter on supported formats


bearcatsandor:
### Problem

When running `beet absubmit artist:"ace frehley" title:"rip it out"`  i get the error 
absubmit: Not analysing Ace Frehley - 1978 - Ace Frehley - Rip It Out - 3:39, file not in supported format.

However, i submited a bug about it to the essentia developers ( https://github.com/MTG/essentia/issues/569 ) and they claim that since essentia is based on ffmpeg that it should work just fine.

Can you tell me what is going on? I've placed a strace here: https://gist.github.com/4eb667a7eb81cdcfb16096b05240c651

Any ideas how i can fix this? The strace complains about my missing the files /home/bearcat/.config/beets/musiclibrary.blb-wal and /home/bearcat/.config/beets/musiclibrary.blb-journal but i've never seen mention of those files anywhere else.

### Setup

* OS: Linux jaguarundi 4.9.5-gentoo #1 SMP Sat Jan 21 16:29:49 MST 2017 x86_64 AMD FX(tm)-6300 Six-Core Processor AuthenticAMD GNU/Linux
* Python version: 2.7.13, 3.4.6, 3.5.6 and 3.6.0
* beets version: git
* Turning off plugins made problem go away (yes/no): no

My configuration (output of `beet config`) is:

```yaml
lyrics:
    bing_lang_from: []
    fallback: No lyrics found
    bing_client_secret: REDACTED
    google_API_key: REDACTED
    force: no
    google_engine_ID: REDACTED
    sources:
    - google
    - lyricwiki
    - lyrics.com
    - musixmatch
    auto: yes
    genius_api_key: REDACTED
    bing_lang_to:
duplicates:
    format: $albumartist - $album - $title
    count: yes
    album: no
    full: no
    keys: []
    move: ''
    tag: ''
    path: no
    copy: ''
    tiebreak: {}
    strict: no
    checksum: ''
    merge: no
    delete: no
play:
    command: gst-play-1.0 --gapless --volume=1.0
    use_folders: no
    raw: yes
    warning_threshold: 100
    relative_to:
library: ~/.config/beets/musiclibrary.blb

replace:
    '[\\/]': _
    ^\.: _
    '[\x00-\x1f]': _
    '[<>:"\?\*\|]': _
    \s+$: ''
    ' ': _
    \!: ''
original_date: yes
replaygain:
    backend: gstreamer
    auto: yes
    overwrite: yes
    targetlevel: 89

[...]

PR #2478, opened 2017-03-16
Remove supported formats check from absubmit

inytar:
This is my take on a solution for #2471. It removes the check for supported formats from the plugin and lowers the logging level when the extractor fails to warning.

Still wondering if we want to edit the log message to indicate that the reason for failing might be a wrong file format. The supported formats are mentioned in the documentation.


sampsyo:
This looks perfect; thank you! Would you mind adding a quick changelog entry?

I think the error message is probably fine… if it becomes a problem, we can consider adding the full output from the underlying tool to the debug output, which might help users diagnose when something goes wrong.

sampsyo:
All merged up! Thank you so much for your help with this, @inytar. :sparkles:

PR #2489, opened 2017-03-23
Allow to transcode a file on-the-fly and download it from the web ui

antlarr:
This commit adds two config options to the web plugin:
convert_extension and convert_command.

When convert_extension is defined (for example, as 'mp3'), a new
option "convert and download as mp3" is shown next to the download link
for files that are not already in mp3 format.

This also adds a config global variable to the javascript code, which
is filled with data obtained from the new /config route, which returns
configuration values useful in the UI (so far, only convert_extension).

The code to convert music to another format was mainly copied
from the convert plugin, and put in a new /item/<item_id>/converted_file
route. The convert command used by default is the same as in the convert
plugin too, but can be configured with the convert_command option.

To enable this, just define in the configuration:
```
web:
    convert_extension: mp3
```
And search for any non-mp3 file in the web plugin.

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -279,13 +279,13 @@ def item_converted_file(item_id):
                 u' '.join(ui.decargs(args)), exc
             )
         )
-    attachment_filename = os.path.basename(util.py3_path(item.path))
-    attachment_filename = attachment_filename[:attachment_filename.rfind('.')] + extension
+    attach_filename = os.path.basename(util.py3_path(item.path))
+    attach_filename = attach_filename[:attach_filename.rfind('.')] + extension
 
     response = flask.send_file(
         util.py3_path(dest),
         as_attachment=True,
-        attachment_filename=attachment_filename,
+        attachment_filename=attach_filename,
     )
     response.headers['Content-Length'] = os.path.getsize(dest)
     return response
@@ -430,9 +430,11 @@ class WebPlugin(BeetsPlugin):
 
             app.config['INCLUDE_PATHS'] = self.config['include_paths']
             if 'convert_extension' in self.config:
-                app.config['convert_extension'] = self.config['convert_extension'].as_str()
+                app.config['convert_extension'] = \
+                    self.config['convert_extension'].as_str()
             if 'convert_command' in self.config:
-                app.config['convert_command'] = self.config['convert_command'].as_str()
+                app.config['convert_command'] = \
+                    self.config['convert_command'].as_str()
 
             # Enable CORS if required.
             if self.config['cors']:
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -23,9 +23,14 @@ import beets.library
 import flask
 from flask import g
 from werkzeug.routing import BaseConverter, PathConverter
+from string import Template
 import os
 import json
 import base64
+import tempfile
+import shlex
+import six
+import subprocess
 
 
 # Utilities.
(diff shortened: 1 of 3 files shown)


antlarr:
Btw, as part of the tests I did for this I tried downloading (and converting) files with unicode characters in their filenames (like ／ or á). That fails due to a problem in flask, so I also submitted https://github.com/pallets/flask/pull/2223 to fix it properly and get correct filenames generated in all flask applications. If that PR is not accepted for some reason in flask, I'll submit another (not so nice) fix here. Note that the problem with unicode characters is not introduced by this PR, but also happens with the regular download link.

beetsplug/web/__init__.py
```diff
@@ -25,6 +25,11 @@
 from werkzeug.routing import BaseConverter, PathConverter
 import os
 import json
+import tempfile
+from string import Template
```
Freso: This should go with the other `from anything_but_spam import spam` imports, instead of the middle of the `import eggs` ones.

beetsplug/web/__init__.py
```diff
@@ -211,15 +216,75 @@ def all_items():
...
+              exc.output)
+        raise
+    except OSError as exc:
+        raise ui.UserError(
+            u"convert: couldn't invoke '{0}': {1}".format(
+                u' '.join(ui.decargs(args)), exc
+            )
+        )
+    attachment_filename = os.path.basename(util.py3_path(item.path))
+    attachment_filename = attachment_filename[:attachment_filename.rfind('.')] + extension
```
Freso: The line is too long, as per https://travis-ci.org/beetbox/beets/jobs/214352835

beetsplug/web/__init__.py
```diff
@@ -348,6 +423,10 @@ def func(lib, opts, args):
             app.config['JSONIFY_PRETTYPRINT_REGULAR'] = False
 
             app.config['INCLUDE_PATHS'] = self.config['include_paths']
+            if 'convert_extension' in self.config:
+                app.config['convert_extension'] = self.config['convert_extension'].as_str()
```
Freso: The line is too long, as per https://travis-ci.org/beetbox/beets/jobs/214352835

beetsplug/web/__init__.py
```diff
@@ -348,6 +423,10 @@ def func(lib, opts, args):
...
+            if 'convert_command' in self.config:
+                app.config['convert_command'] = self.config['convert_command'].as_str()
```
Freso: The line is too long, as per https://travis-ci.org/beetbox/beets/jobs/214352835

beetsplug/web/__init__.py
antlarr: Thanks for the review. I fixed all the isues you mentioned

sampsyo:
Wow; this is cool! Interestingly, we've talked about a flexible system for this kind of transcoding in the AURA spec, where the idea is that you could specify whatever file formats you're willing to accept: https://github.com/beetbox/aura/blob/master/docs/api.rst#audio-formats-and-quality

Is there any code that currently needs to be duplicated from the `convert` plugin? If so, should we move any of it to a library that we can share?

antlarr:
Doing a simple git rebase -i on master and not changing anything in the git-rebase-todo should just replay the changes but it gave some conflicts on commits unrelated to my branch. This gave me a hard time rebasing my branch so that it applied cleanly on master. Because of that, I decided to just branch from master again, cherry-pick my changes rebasing them at the same time and then force-push the changes to the origin branch of this pull request. That's the reason they appear as new commits.

sampsyo:
Hi, @antlarr—thanks for your patience while I'm digging myself out of a PR review backlog! I'm aware I have more significant changes from you to review too, but while I'm on my way there, here are a couple of comments:

- Eventually, we'll need to add documentation for the new options (and the URL scheme) to the web plugin.
- The transcoding logic here is pretty similar to the core logic for invoking FFmpeg in the `convert` plugin. To avoid code duplication, I think it would make sense to make a core module that both plugins can share to do this basic transcoding stuff. The idea is similar to the `beets.art` module we added some time back, which is shared among all the various plugins that need to embed album art. (The code originally lived in the `embedart` plugin, but it became necessary for other plugins to do the same thing as a subroutine.)
- I don't quite see what's going on with the `config` value on the JavaScript side—is it possible that's out of date?

jtpavlock:
@antlarr any interest in continuing this?

PR #2494, opened 2017-03-25
Fix import of multidisc releases with subdirectories

### LINKED ISSUE #2493: Import of a multidisc release is broken if it has subdirectories
### Problem

I have a directory structure like:
```
/music/artist/album/CD1/*.mp3
/music/artist/album/CD1/info/*.txt
/music/artist/album/CD1/covers/*.jpg
/music/artist/album/CD2/*.mp3
/music/artist/album/CD2/info/*.txt
/music/artist/album/CD2/covers/*.jpg
/music/artist/album/CD3/*.mp3
/music/artist/album/CD3/info/*.txt
/music/artist/album/CD3/covers/*.jpg
```

When importing such a directory structure with beets, the importer doesn't 
collapse the paths since it has subdirectories that "break" the pattern matching.
All the contents of CD1/CD2/CD3 directories should be colllapsed into the same release, as [the documentation says](http://beets.readthedocs.io/en/v1.3.17/guides/tagger.html#overview) :

```
First, directories that look like separate parts of a multi-disc album are tagged
together as a single release. If two adjacent albums have a common prefix,
followed by “disc,” “disk,” or “CD” and then a number, they are tagged together.
```

### Setup

* OS:  openSUSE Tumbleweed
* Python version: 3.6
* beets version: git master

antlarr:
Don't stop collapsing paths when iterating over subdirectories of
already collapsed paths.

Fixes #2493

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -1473,10 +1473,8 @@ def is_subdir_of_any_in_list(path, dirs):
     """Returns True if path os a subdirectory of any directory in dirs
     (a list). In other case, returns False.
     """
-    for d in dirs:
-        if path.startswith(d + b'/'):
-            return True
-    return False
+    ancestors = ancestry(path)
+    return any(d in ancestors for d in dirs)
 
 
 def albums_in_dir(path):
@@ -1498,10 +1496,9 @@ def albums_in_dir(path):
         # and add the current directory. If so, just add the directory
         # and move on to the next directory. If not, stop collapsing.
         if collapse_paths:
-            if (not collapse_pat and collapse_paths[0] in ancestry(root)) or \
+            if (is_subdir_of_any_in_list(root, collapse_paths)) or \
                     (collapse_pat and
-                     collapse_pat.match(os.path.basename(root))) or \
-                    is_subdir_of_any_in_list(root, collapse_paths):
+                     collapse_pat.match(os.path.basename(root))):
                 # Still collapsing.
                 collapse_paths.append(root)
                 collapse_items += items
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -1469,6 +1469,16 @@ MULTIDISC_MARKERS = (br'dis[ck]', br'cd')
 MULTIDISC_PAT_FMT = br'^(.*%s[\W_]*)\d'
 
 
+def is_subdir_of_any_in_list(path, dirs):
+    """Returns True if path os a subdirectory of any directory in dirs
+    (a list). In other case, returns False.
+    """
+    for d in dirs:
+        if path.startswith(d + b'/'):
+            return True
+    return False
+
+
 def albums_in_dir(path):
     """Recursively searches the given directory and returns an iterable
     of (paths, items) where paths is a list of directories and items is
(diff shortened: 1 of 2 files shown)


beets/importer.py
```diff
@@ -1469,6 +1469,16 @@ def group(item):
 MULTIDISC_PAT_FMT = br'^(.*%s[\W_]*)\d'
 
 
+def is_subdir_of_any_in_list(path, dirs):
+    """Returns True if path os a subdirectory of any directory in dirs
+    (a list). In other case, returns False.
+    """
+    for d in dirs:
+        if path.startswith(d + b'/'):
```
sampsyo: A safer way to do a subdirectory check is `path in ancestry(d)`.

beets/importer.py
```diff
@@ -1469,6 +1469,16 @@ def group(item):
...
+            return True
+    return False
```
sampsyo: A functional style might be slightly more readable:

```py
any(path in ancestry(d) for d in dirs)
```

beets/importer.py
```diff
@@ -1490,7 +1500,8 @@ def albums_in_dir(path):
         if collapse_paths:
             if (not collapse_pat and collapse_paths[0] in ancestry(root)) or \
                     (collapse_pat and
-                     collapse_pat.match(os.path.basename(root))):
+                     collapse_pat.match(os.path.basename(root))) or \
+                    is_subdir_of_any_in_list(root, collapse_paths):
```
sampsyo: Not having 100% thought through this, this check seems like it might be a bit redundant with `collapse_paths[0] in ancestry(root)` above. Do we still need that check if we add this new (seemingly broader) check?

beets/importer.py
antlarr: Ah, I didn't know about ancestry. I'll use it if you prefer that.

beets/importer.py
antlarr: That would be the other way around,
```
any(d in ancestry(path) for d in dirs)
```
```path``` would be ```/.../album/CD1/covers``` and dirs would be the list of already included paths to collapse, so it would include ```/.../album/CD1```

beets/importer.py
antlarr: Good point, I'm not completely sure about removing "not collapse_pat and", but indeed, if that's "removable", then it can be simplified. In my tests that seems to work fine at least.

sampsyo:
OK! I'm for merging this if we can convince ourselves it's safe… this will take some careful reasoning, but is there any case where the new condition (which checks whether the new path is a subdirectory of _any_ collapsed path) will cause unexpected problems? I don't have any examples at hand, but maybe we should sleep on it.

sampsyo:
Also, it looks like the Appveyor failure above is spurious? https://ci.appveyor.com/project/beetbox/beets/build/1639/job/xmcnha56lqul89km

I can't seem to find an option to re-run that.

antlarr:
I don't think it should cause problems since we're replacing the condition ```not collapse_pat and collapse_paths[0] in ancestry(root)``` with ```is_subdir_of_any_in_list(root, collapse_paths)``` which is true for all cases where the first condition is, and some more. I tested that the collapse stops when iterating over a directory which doesn't belong to the same multi CD release and not only because there was nothing else to collapse. Btw, I also tested with ```not collapse_pat and is_subdir_of_any_in_list(root, collapse_paths)``` but that was False for cases which should be True, so it didn't collapse my test case correctly.

About the Appveyor failure... yes, it seems a spurious problem. Just in case it helps, the python3.5 test in Appveyor runs fine and I'm running python3.6 here, so that should be fine too.

sampsyo:
OK, thanks! Could you please summarize what was fixed in the changelog entry and then this will be ready to go?

sampsyo:
Thanks again!

Issue #2490, opened 2017-03-23
Regression in play plugin: no CLI output on succesful play.


nathdwek:
It seems 8d972576 and friends have introduced a small regression in the play plugin: there is no output anymore on successful play indicating the number of tracks queued. This does not seem to be desired, as there is still this snippet left in the code:
```python
# Print number of tracks or albums to be played, log command to be run.
item_type += 's' if len(selection) > 1 else ''
ui.print_(u'Playing {0} {1}.'.format(len(selection), item_type))
log.debug(u'executing command: {} {!r}', command_str, open_args)
```

I suppose the weird `ui.print_` (as opposed to use the dedicated logger) lost in there is responsible for this.

@diomekes: would it be possible for you to parse this section real quick to see what was the intention here and if there is indeed a quick fix?

Repro:
```
beets version 1.4.4
Python version 3.5.2+
plugins: chroma, convert, discogs, duplicates, embedart, fetchart, info, missing, play, random, zero
```
Ubuntu 16.10

diomekes:
This is the output I get when exceeding the threshold:

`You are about to queue 704 track.`
`[C]ontinue, Abort?`
`Playing 704 tracks.`

@nathdwek, you don't get this output?

For what it's worth, that code you pasted was already part of the plugin, I just moved it to it's current position as part of the refactoring to add the importer prompt.

sampsyo:
That is odd—I don't quite see why `ui.print_` wouldn't show up. (It will, FWIW, go to stdout instead of stderr, but I doubt that's the problem.)

That "you are about to queue 704 track" grammar though… we should certainly do something about that.

nathdwek:
Woops: I fixed the content but not the title of the opening post.
* The warning certainly works!
* The plural issue was recently fixed in  920784a
* The only remaining issue is that when there is not warning, there is no output at all.

diomekes:
I'm still confused.  If my query plays 3 albums, it says:
`Playing 3 albums.`

If it plays 13 tracks, it says:
`Playing 13 tracks.`

Could you maybe post your command and the output, or lack of output?

nathdwek:
```
nathdwek@nathome:~[home]$ beet -vv play love in the dark
user configuration: /home/nathdwek/.config/beets/config.yaml
data directory: /home/nathdwek/.config/beets
plugin paths: 
Sending event: pluginload
library database: /home/nathdwek/.beets.db
library directory: /home/nathdwek/music/beetsdir
[...]

PR #2496, opened 2017-03-26
Flush the stdout buffer after printing a line

sampsyo:
Pointed out in #2490, this is a regression introduced by #2398. We need to flush the buffer to faithfully emulate the "real" print() function.

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -148,6 +148,7 @@ def print_(*strings, **kwargs):
         if hasattr(sys.stdout, 'buffer'):
             out = txt.encode(_out_encoding(), 'replace')
             sys.stdout.buffer.write(out)
+            sys.stdout.buffer.flush()
         else:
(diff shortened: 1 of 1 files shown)


jrobeson:
stupid buffer. :(     We might wanna go over the senf version of this to see if we're missing anything else.

sampsyo:
Good point. Looks like senf doesn't flush by default, oddly? I'm *pretty* sure the built-in `print` does…
https://github.com/quodlibet/senf/blob/681b71532e8acbf436c00a22e162fce9d9cb47dc/senf/_print.py#L125
https://github.com/quodlibet/senf/blob/681b71532e8acbf436c00a22e162fce9d9cb47dc/senf/_print.py#L62

sampsyo:
Just to confirm, sys.stdout on Python 3 does flush whenever there's a newline (which happens at the end of `print` calls by default):

```
>>> sys.stdout.line_buffering
True
```

PR #2500, opened 2017-04-01
Document using Python functions in a field

RollingStar:
I thought this part of the documentation was vague. I didn't know I could define *multiple* Python functions inline and reference them in a return statement until I tried it.

### DIFF
--- a/docs/plugins/inline.rst
+++ b/docs/plugins/inline.rst
@@ -48,6 +48,18 @@ a result for the value of the path field, like so::
 You might want to use the YAML syntax for "block literals," in which a leading
 ``|`` character indicates a multi-line block of text.
 
+You can also define multiple Python functions and use them in a custom field::
+
+    item_fields:
+       my_test_field: |
+            def sub_function_one():
+                return 'TEST-1234'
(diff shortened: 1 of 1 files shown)


sampsyo:
Thanks for pointing this out! I'm not sure we need the example, though—maybe it's just me, but I think it should be fairly straightforward to understand that, since you can write *any* Python code there, that would include functions. We can just say that: "you can write any Python code, including functions and classes and such."

Also, I'm in general not much of a fan of contrived examples. You'll notice that most of the examples we use in the docs—like the one just above there—are examples of real things you might actually want to do with beets. Contrived examples can confuse more than they help.

RollingStar:
Yeah, I also wasn't sure about the example used and how simple or complex to go. Maybe the beets documentation isn't the best place to learn Python syntax. Would what you wrote: "you can write any Python code, including functions and classes and such." fit the style of the beets documentation? I'm fine with it if you are.

sampsyo:
Yeah! I didn't necessarily mean for that to be the final text—just a starting point—but it's probably close enough.

jtpavlock:
@RollingStar any chance you'd be interested in finishing this up?

PR #2502, opened 2017-04-03
Use preferred countries when selecting a release event

### LINKED ISSUE #2501: Use preferred country to select release event within each release
When manually selecting a release during import (e.g. by using the MusicBrainz ID or via search) that contains multiple countries, it seems like preferred countries are ignored in favor of the first country in the list.

For example, in [this release](https://musicbrainz.org/release/f7304df7-2e3d-461a-b05f-442d330c1bf6) there are multiple countries. With the following config the preferred countries are ignored and Canada is selected instead:

    match:
      preferred:
        countries: ['XW', 'US', 'GB', 'XE']

The same also happens if you call `mbsync` and the country has been updated.

Should the importer also handle selecting preferred metadata when multiple options are available? Does it handle this already and I'm just doing it wrong?

djl:
Fixes #2501.

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -109,6 +109,22 @@ def _preferred_alias(aliases):
         return matches[0]
 
 
+def _preferred_release_event(release):
+    """Given a release, select and return the user's preferred release
+    event as a tuple of (country, release-date). Falls back to the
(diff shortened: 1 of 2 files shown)


sampsyo:
That was fast! This looks perfect; thank you. :sparkles:

Issue #505, opened 2014-01-25
Add support for multi-valued tags


ryanakca:
Please add support for multi-valued tags. For example, this could be used for the multi-valued occurrences of PERFORMER / COMPOSER / ENSEMBLE / etc. recommended by http://age.hobba.nl/audio/mirroredpages/ogg-tagging.html .

sampsyo:
Good summary; thanks!

The reason we haven't done 4 is because it will also make queries more complicated. Currently, there is a uniform way to construct queries for any fixed field. Having a separate table for the list-valued fields would mean that querying would look very different depending on the type of the field. It's also not 100% clear how this would work with flexible attributes, or how flexible attributes could be migrated to fixed attributes.

In summary: it could definitely be the best idea, but someone needs to do a detailed design to show exactly how this would work.

pdf:
I looked at extending some of the beets code some years ago now (though I don't normally write Python), but I gave up when I started looking at the DB stuff.  Making the entire database layer bespoke doesn't seem labor-efficient, or: _How I Learned to Stop Worrying and Love the ORM_.

The kludge in 3 would probably solve most people's desire here, and it's pretty trivial by comparison.

sampsyo:
As you noted, 3 has not been on the table mostly because querying is too weird. Queries are a core concept.

pdf:
That's only a problem if you care about querying these relationships - for me, I just want my tags right, without having to use a GUI (ie - Picard).  Other tickets have been created for the same reason.  I understand the desire to architect it elegantly though.

Boehemyth:
One thing I had started looking into (before I got a new job and got too busy) was moving beets over to ZODB instead of sqlite.  This would basically be a core overhaul, but it would really make sense in the long run.  ZODB is a object oriented database that would basically allow us to keep track of lists inside our records (and query them).  The problem is this will take a great deal of effort, but I think it is better than moving to a document oriented database like we've been talking about here.

sampsyo:
I looked at ZODB as an alternative too, and it has lots of great features. I was a little turned off by tapping into Zope ecosystem, for fear of too many dependencies and perceived lock-in—it would definitely be nice to be able to interoperate with non-Python tools. Still worth considering though.

Boehemyth:
>  I was a little turned off by tapping into Zope ecosystem, for fear of too many dependencies and perceived lock-in

[...]

PR #2503, opened 2017-04-04
Add multivalue genre tag support

antlarr:
These changes allow beets to store multiple genre tags for multivalues, with the intent of fixing #505 . Note that this is not ready to be merged yet, but it's a beginning of discussion on how to properly fix the issue.

Note that these commits store lists in album.genre and item.genre properties, while storing them as text strings in the database using a separator. Also, I checked that if a user has something like this as path definition:
```
paths:
    default: $albumartist/$genre/$album%aunique/$genre/$track - $title
```
then $genre is formatted as a string correctly . In my test case, creating a directory called "Rock;Blues" .

This can break some complex configurations. For example, I had this as album_fields definition to add a album_genre_for_path tag that would contain the most used genre in an album:

```
album_fields:
    album_genre_for_path: |
        if 'album_genre' in globals():
            return album_genre.title()
        if hasattr(items[0], 'album_genre'):
            return items[0].album_genre.title()
        genres = {}
        cnt, candidate = 0, ''
        for item in items:
            genres[item.genre] = genres.get(item.genre, 0) + 1
            if genres[item.genre] > cnt:
                cnt, candidate = genres[item.genre], item.genre.title()
        return candidate
```

When evaluating this tag, beet broke sine item.genre is no longer hashable. I think this is a step better than silently breaking the directory structure. I fixed it easily by replacing the code above with:

```
    album_genre_for_path: |
        if 'album_genre' in globals():
            return album_genre.title()
        if hasattr(items[0], 'album_genre'):
            return items[0].album_genre.title()
        genres = {}
        cnt, candidate = 0, ''
        for item in items:
            if isinstance(item.genre, list):
                values = item.genre
            else:
                values = [item.genre]
            for genre in values:
                genres[genre] = genres.get(genre, 0) + 1
                if genres[genre] > cnt:
                    cnt, candidate = genres[genre], genre.title()
        return candidate
```

Then I also tested what happens when reading a file written by another application with multiple genre tags:

```
Python 3.6.1 (default, Mar 23 2017, 13:04:44) [GCC] on linux
>>> import beets.mediafile
>>> mf=beets.mediafile.MediaFile('01 - Big Brother & the Holding Company -  Bye, Bye Baby.flac')
>>> mf.genre
['Rock', 'Album Rock']
```
Which is what I would have expected.


antlarr:
I tested this with flac and mp3 files (that is, with APE tags, where multiple genre tags are written and ID3 tags, where `\0` bytes are used to separate genres in a single ID3 tag). Everything works as expected due to `ListMediaField` (btw, congratulations to whoever implemented it and the `*ListStorageStyle` classes, they are really nicely done)

The remaining questions are:
- Should a global parameter be added to the configuration? If yes, should it be named `multivalue_separator` or ... ? (it would default to `';'` ? `'; '` ? `', '` ?
- What to use for queries in the `StringList` type? I think using `SubstringQuery` is probably "good enough", but that can be discussed as it could return some unexpected matches . I'm thinking on someone querying for `"ck;Bl"` for example, but on the other side, maybe someone would like to query for `"Rock;Blues"` to get items with both values (although that wouldn't match genres stored as `"Blues;Rock"` which should be considered ).

beets/dbcore/types.py
```diff
@@ -190,6 +190,38 @@ class String(Type):
...
+        return six.text_type(value)
+
+    def parse(self, string):
+        """Parse a (possibly human-written) string and return the
+        indicated value of this type.
+        """
+        if string is None:
+            return self.null
+        else:
+            return string.split(';')
```
sampsyo: We should probably make the value of list types like this a tuple rather than a Python list. That way, we avoid letting programmers mutate the list of genres. For example:

```
genres = item.genre
genres.append(foo)
item.store()
```

That wouldn't work, because the `Item` doesn't know you changed that copy of the list. We can capture this property by making the value immutable, i.e., a tuple.

beets/dbcore/types.py
```diff
@@ -190,6 +190,38 @@ class String(Type):
...
+    """A List of Unicode strings type, stored in a string
+    """
+    sql = u'TEXT'
+    query = query.SubstringQuery
+
+    def from_sql(self, sql_value):
+        return self.parse(sql_value)
+
+    def to_sql(self, value):
+        return self.format(value)
```
sampsyo: Interesting! We've never used `parse` and `format` to interact with the underlying SQL value, but I suppose this makes sense. But maybe null characters for the SQL storage would be a bit safer?

beets/dbcore/types.py
```diff
@@ -190,6 +190,38 @@ class String(Type):
...
+
+    def format(self, value):
+        """Given a value of this type, produce a Unicode string
+        representing the value. This is used in template evaluation.
+        """
+
+        if isinstance(value, list):
```
sampsyo: Any particular reason why this check is necessary? If the value is somehow taking on a string value, we should probably stop that! It will cause other problems—for example, `to_sql` is likely to do something unpredictable. As with the other types, we should enforce that the underlying value must always be a tuple of strings.

beets/dbcore/types.py
antlarr: hmm, the problem is that mutagen uses a list for the genre tag, so it might have problems if a tuple is used instead. Also, ListMediaField is also used as a base class of ImageListField, so using a tuple might break that too. I'll check both things.

beets/dbcore/types.py
antlarr: Hehe, indeed! Thanks for noticing. I'll try to change that

beets/dbcore/types.py
antlarr: Indeed, I just removed that check

beetsplug/info.py
```diff
@@ -98,8 +98,9 @@ def print_data(data, item=None, fmt=None):
     formatted = {}
     for key, value in data.items():
         if isinstance(value, list):
-            formatted[key] = u'; '.join(value)
-        if value is not None:
+            if value:
+                formatted[key] = u'; '.join(value)
+        elif value is not None:
```
sampsyo: This, for what it's worth, is one of the things that worries me about making arbitrary fields into lists: now client code will frequently need to have boilerplate for the form `if isinstance(value, list):`. This sort of thing puts additional weight on the side of having an explicit way to access lists, or some other foolproof conversion to produce a scalar value.

beets/mediafile.py
```diff
@@ -799,6 +799,9 @@ def fetch(self, mutagen_file):
             return []
 
     def store(self, mutagen_file, values):
+        if values == []:
```
sampsyo: I think the right PEP8y thing here might be `if not values:`.

beets/mediafile.py
```diff
@@ -582,7 +582,10 @@ def set(self, mutagen_file, value):
         """Set an individual value as the only value for the field using
         this style.
         """
-        self.set_list(mutagen_file, [value])
+        if value is None:
```
sampsyo: That's odd—I'm a little surprised that ListStorageStyle is allowed to return None from `get`. That seems like it makes things more complicated for clients in general. Maybe that's worth fixing?

antlarr:
Btw, I guess you'll doubt that getall('TIPL') can be replaced with get('TIPL') (mainly because I doubted about it too :) ). So here is the test I did with mutagen to check that it doesn't allow two TIPL frames:

```
>>> import mutagen
>>> m=mutagen.File('full.mp3')
>>> m.tags.get('TIPL')
>>> frame=mutagen.id3.Frames['TIPL'](encoding=mutagen.id3.Encoding.UTF8, people=[['aa','bb'],['cc','dd']])
>>> m.tags.add(frame)
>>> m.save()
```

Here, I checked with okteta (an hex editor) that the file included the added TIPL frame.

```
>>> frame2=mutagen.id3.Frames['TIPL'](encoding=mutagen.id3.Encoding.UTF8, people=[['ee','ff'],['gg','hh']])
>>> frame
TIPL(encoding=<Encoding.UTF8: 3>, people=[['aa', 'bb'], ['cc', 'dd']])
>>> frame2
TIPL(encoding=<Encoding.UTF8: 3>, people=[['ee', 'ff'], ['gg', 'hh']])
>>> m.tags.add(frame2)
>>> m.tags.getall('TIPL')
[TIPL(encoding=<Encoding.UTF8: 3>, people=[['ee', 'ff'], ['gg', 'hh']])]
>>> m.save()
```

and here, I checked again with an hex editor that the frame wasn't just added, but it replaced the previous TIPL frame.

beetsplug/info.py
antlarr: I understand what you mean. Note that list fields (genre, arranger ... ) now are always lists, the problem comes with code like that which tries to work with all fields at once. Would you prefer that I create a class that inherit from lists and reimplements __str__ so it returns '; '.join(self) ? That way code like that could be changed to:
```
if value:
    formatted[key] = str(value)
```
But still, that's not perfect neither.

beets/mediafile.py
antlarr: Perfect, I didn't know if you preferred that or preferred to compare with []. Probably after seeing some assertEqual with [] I thought you preferred to compare with []. I'll fix it

beets/mediafile.py
antlarr: In fact, it's ok. `get` is implemented as:
```
          try:
              return self.get_list(mutagen_file)[0]
          except IndexError:
              return None
```
So None is returned when get_list returns an empty lists, that is, when the tag doesn't exist in the file. This allows to identify when a tag doesn't exist and allows to remove a tag from a file. Do you think that it would be better to return an empty string and always interpret an empty value as a non-existant tag? (so setting ... say artist to '' would delete the tag entry)

antlarr:
I just noticed I had this written in a browser tab but forgot to click comment. I'll post it for the records, even if it's probably old now.

> Very cool! This is diff is surprisingly small—I expected this to be a far more disruptive change.
> 

Yes, I also was surprised, but most of the work was already implemented in ListMediaField, the *ListStorageStyle classes and mutagen.

> A couple of ideas come to mind for improving backwards compatibility, neither of which we necessarily have to use:
> 
> We could keep a field called genre that just gets the joined string, and add a new genres field for the list. That's what MediaFile previously did.

Yes, I saw that, but the genres tag is (afaik) not used by any other application, but genre is.

> We could somehow allow a string to be assigned into the genre field, which would then be treated as a single-element list.

Have a look at the test_write_genre_list_from_string test I added in test_mediafile.py to test fad58096046cb75e4fb3824a370d12509806ea14 :). I think that's even better since it allows assigning a string but automatically converts it to a list.

> The only thing that makes me somewhat uncomfortable here is the age-old slippery slope question. How many other fields would it make sense to transform into lists? Should artist be a list? albumartist? If we make too many changes like this, will we disrupt simpler code that just cares about a single value?

Hmm, I don't think that would be too much work. I can give it a try if you want. (This is now done)

beets/dbcore/query.py
```diff
@@ -227,6 +229,63 @@ def string_match(cls, pattern, value):
...
+        alias = self.get_unique_alias()
+        clause = alias + ".value like ? escape '\\'"
+        subvals = [search]
+        table = 'json_each({0}) as {1}'.format(self.field, alias)
+        return clause, subvals, (table,)
+
+    def negated_clause(self):
+        clause, subvals, table = self.col_clause()
+        _table = self.model_cls._table
+        primary_key = '%s.id' % (_table)
```
antlarr: @sampsyo Do you think it's worth it to add a get_primary_key_field to Model so instead of hardcoding `id` I call self.model_cls.get_primary_key_field() which would iterate over its fields and search which has a PRIMARY_ID type? I think that would be more elegant, but I'm not sure if it's worth the effort.

beets/dbcore/db.py
```diff
@@ -842,11 +882,15 @@ def _fetch(self, model_cls, query=None, sort=None):
         """
         query = query or TrueQuery()  # A null query.
         sort = sort or NullSort()  # Unsorted.
-        where, subvals = query.clause()
+        where, subvals, new_tables = query.clause()
```
antlarr: @sampsyo should I still allow Query objects to maybe return 2-tuples just in case any external plugin still uses the old API?  It's easy to check for unpack errors but it's a bit ugly, so I preferred to ask your opinion.

.travis.yml
```diff
@@ -47,6 +47,11 @@ addons:
             - python3-gst-1.0
             - unrar
 
+before_install:
+    - sudo add-apt-repository -y "deb http://archive.ubuntu.com/ubuntu xenial main"
+    - sudo apt-get -y update
+    - sudo apt-get install sqlite3 -y
```
sampsyo: Hmm… I suppose it's a somewhat troubling indication about the popularity of SQLite JSON support if we need to override the apt repository.

antlarr: Yep, that was a test, but anyway, I don't know about appveyor and I noticed it also failed, so I went back, removed that patch and instead added support for older sqlite versions, check the comment on a52ba556143cca87b561a85d448725d89e60ea0a and https://repology.org/metapackage/sqlite3/badges for a full distribution list and which sqlite3 version they have (remember, json support was introduced in 3.9.0)

sampsyo: Cool; thanks! I'm overdue for another full code review; I'll do that soon.

antlarr: Great, I understand this turned out to be much larger than expected (mostly because of Query related changes) but I think the result is worth it :)

arcresu:
@antlarr thanks for your work on this PR. It's been a while since it's had any attention, but it seems that multi-valued tags are still a feature that people would like to see!

Just a heads up that we're planning on splitting MediaFile out into a standalone Python module & git repository during the current beets release cycle, and this is clearly going to affect this PR. Your branch is anyway conflicting with master already, but after the split it'll mean that the changes will also need to be split into two separate sets. Are you still interested in working on this? If so it might be sensible to start with the MediaFile changes at beets/mediafile.

jtpavlock:
Closing as no response from the original author. There's still a lot of good info to be gained from this, but any further implementation ideas should be directed to #505. Also, I think it'd be ideal to first implement the basis for multi-value tags, and then add genres as an additional PR after.

PR #2508, opened 2017-04-11
OperationalError from SQlite that indicates a permissions problem.

Mary011196:
Fixes 1676.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -32,6 +32,10 @@ from beets.dbcore import types
 from .query import MatchQuery, NullSort, TrueQuery
 import six
 
+class AccessFileError(Exception):
+    """UI exception. Commands should throw this in order to display
+    nonrecoverable errors to the user.
+    """
 
 class FormattedMapping(collections.Mapping):
     """A `dict`-like formatted view of a model.
@@ -680,8 +684,11 @@ class Transaction(object):
         """Execute an SQL statement with substitution values and return
         the row ID of the last affected row.
         """
-        cursor = self.db._connection().execute(statement, subvals)
-        return cursor.lastrowid
+
+            cursor = self.db._connection().execute(statement, subvals)
+            raise AccessFileError("unable to open database file. It might be a permissions problem")
+            return cursor.lastrowid
+
 
     def script(self, statements):
         """Execute a string containing multiple SQL statements."""
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -36,7 +36,8 @@ import six
 class DBAccessError(Exception):
     """The SQLite database became inaccessible.
     This can happen when trying to read or write the
-    database when, for example, the database file is deleted or otherwise disappears.
+    database when, for example, the database file
+    is deleted or otherwise disappears.
     There is probably no way to recover from this error.
     """
 
@@ -692,10 +693,10 @@ class Transaction(object):
             cursor = self.db._connection().execute(statement, subvals)
             return cursor.lastrowid
         except sqlite3.OperationalError as e:
-            if e.args[0] in ('unable to open database file',
-            'attempt to write a readonly database file'):
-                raise DBAccessError("Unable to open database file." \
-                 "It might be a permissions problem.")
+            if e.args[0] in ("attempt to write a readonly database",
+                             "unable to open database file"):
+                raise DBAccessError('Unable to open database file.'
+                                    'It might be a permissions problem')
 
     def script(self, statements):
         """Execute a string containing multiple SQL statements."""
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -35,10 +35,10 @@ import six
 
 class DBAccessError(Exception):
     """The SQLite database became inaccessible.
-    This can happen when trying to read or write the
-    database when, for example, the database file
-    is deleted or otherwise disappears.
-    There is probably no way to recover from this error.
+
+    This can happen when trying to read or write the database when, for
+    example, the database file is deleted or otherwise disappears. There
+    is probably no way to recover from this error.
     """
 
 
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -693,10 +693,14 @@ class Transaction(object):
             cursor = self.db._connection().execute(statement, subvals)
             return cursor.lastrowid
         except sqlite3.OperationalError as e:
+            # In two specific cases, SQLite reports an error while accessing
+            # the underlying database file. We surface these exceptions as
+            # DBAccessError so the application can abort.
             if e.args[0] in ("attempt to write a readonly database",
                              "unable to open database file"):
-                raise DBAccessError('Unable to open database file.'
-                                    'It might be a permissions problem')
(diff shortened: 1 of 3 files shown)


beets/ui/__init__.py
```diff
@@ -1159,10 +1159,13 @@ def _open_library(config):
...
     except (sqlite3.OperationalError, sqlite3.DatabaseError):
+        message = ""
         log.debug(u'{}', traceback.format_exc())
-        raise UserError(u"database file {0} could not be opened".format(
+        if e.args[0] == "unable to open database file":
+            message = "It might be a permissions problem."
+        raise UserError(u"database file {0} could not be opened.%s".format(
             util.displayable_path(dbpath)
-        ))
+        )%message)
```
sampsyo: This handler already catches errors when we *open* the database. The bug report on this thread arises after the database is already opened, when it is accessed. So I think the new exception handling bit will need to go in the main function.

Mary011196: Hello,
I add this in main but it gives me the message from the first `UserError` excpetion that raised before. 
```
    except sqlite3.OperationalError as e:
        if e.args[0] == "unable to open database file":
            raise UserError(u"unable to open database file. It might be a permissions problem")
```
Is there a way to do it like `db_query.InvalidQueryError` in order to check the `mutate` function in db.py?

sampsyo: Oh, cool! Looks like you're almost there. There are actually two ways to go with this:

* Inside mutate, raise a `UserError` and let the top-level (`main`) exception handling print the message.
* Inside `main`, just catch the `OperationalError` directly and print the error message there. (But don't raise a new `UserError`.)

It's actually a difficult call to say what's better, but I like your idea of raising a new, specialized exception in `mutate`. In fact, maybe we can invent a new error exception class, sort of like `InvalidQueryError`, for filesystem access problems like this one?

Mary011196: Hello,
I' ve tried both ways but i have the same problem.  In the first way i just add in mutate ` raise UserError("unable to open database file. It might be a permissions problem")` and in the second way i wrote in `main` this:
```
    except sqlite3.OperationalError as e:
        if e.args[0] == "unable to open database file":
           print("unable to open database file. It might be a permissions problem")
```
I wrote also a new error exception class
```
class AccessFileError(Exception):
     """Permission Exception. Exception that Specifically inidcates the
     possibility of a permissions problem on database file.
     """
```
In order to work with this way i have to raise an `AccessFileError` in mutate and catch it in the `main`?

Thank you very much!

Mary011196: I think i fixed it and i tried it with all  ways.  But it works when i deny the write permission in my database file and the traceback that gives me is `attempt to write a readonly database `. Is this right??
Thank you

sampsyo: > In order to work with this way i have to raise an AccessFileError in mutate and catch it in the main?

Yes, exactly!

> But it works when i deny the write permission in my database file and the traceback that gives me is attempt to write a readonly database. Is this right??

It's somewhat hard to know exactly what causes specific kinds of errors in the SQLite library, which hides the precise details of what went wrong. But if you see that error when setting the database file to have permissions that prevent writing by your user, that does sound like something we should handle—perhaps in *addition* to the existing "unable to open database file" error?

Mary011196: But when it gives me the "unable to open datapase file" error it prints the message of UserError that raise the _open_library. Maybe the way i change my settings is wrong? 

Thank you

sampsyo: Hmm… can you try pushing your updated code so it appears here in the PR? Then I can take a look at how it's actually working.

beets/dbcore/db.py
```diff
@@ -680,9 +684,12 @@ def mutate(self, statement, subvals=()):
         """Execute an SQL statement with substitution values and return
         the row ID of the last affected row.
         """
+
         cursor = self.db._connection().execute(statement, subvals)
+        raise AccessFileError("unable to open database file. It might be a permissions problem")
```
sampsyo: This should probably be in an exception handler for OperationalError.

beets/dbcore/db.py
```diff
@@ -680,14 +684,17 @@ def mutate(self, statement, subvals=()):
...
+            return cursor.lastrowid
+        except sqlite3.OperationalError as e:
+            raise AccessFileError("unable to open database file. It might be a permissions problem")
+
 
     def script(self, statements):
         """Execute a string containing multiple SQL statements."""
         self.db._connection().executescript(statements)
 
-
```
sampsyo: I know it seems small, but we need this blank line here for consistency.

beets/dbcore/db.py
```diff
@@ -32,6 +32,10 @@
 from .query import MatchQuery, NullSort, TrueQuery
 import six
 
+class AccessFileError(Exception):
```
sampsyo: How about "DBAccessError" for clarity?

beets/dbcore/db.py
```diff
@@ -32,6 +32,10 @@
...
+    """UI exception. Commands should throw this in order to display
+    nonrecoverable errors to the user.
```
sampsyo: This looks like the description for UserError? Here's an alternative docstring:

> The SQLite database became inaccessible. This can happen when trying to read or write the database when, for example, the database file is deleted or otherwise disappears. There is probably no way to recover from this error.

beets/dbcore/db.py
```diff
@@ -680,14 +684,17 @@ def mutate(self, statement, subvals=()):
         """Execute an SQL statement with substitution values and return
         the row ID of the last affected row.
         """
-        cursor = self.db._connection().execute(statement, subvals)
-        return cursor.lastrowid
+        try:
+            cursor = self.db._connection().execute(statement, subvals)
+            return cursor.lastrowid
+        except sqlite3.OperationalError as e:
+            raise AccessFileError("unable to open database file. It might be a permissions problem")
```
sampsyo: Here is where the check for error strings like "attempt to write a readonly database" should occur. (For what it's worth, I did a little experiment—that's the message that appears if I try to access the database after deleting the file from my disk!)

Mary011196: So,  i have to add an if that check if e.args[0] is equals with "attempt to write a readonly database file" and another that checks if e.args[0] is equals with "unable to open database file"? I already tried it and when the message is "unable to open database file" it gives me again the message of UserError.

sampsyo: Yep! Or we can just have one `if` that checks whether it's either of those things:

```
if e.args[0] in ('string1', 'string2'):
```

I don't quite understand what you mean about the UserError… is some *other* code producing a UserError exception?

Mary011196: Hello,
I think i fixed it. Is it rigth?
Thank you!!

sampsyo:
This looks great! Thank you for your contribution! I just pushed a few small changes here; let's see what Travis says.

Mary011196:
Thank you too. I will continue with the next issue. Will my code be merged?

Thank you again!

sampsyo:
Looks like everything's in order. Merging now.

Issue #549, opened 2014-02-18
Add an entry to replace to lookout for paths that begin with "-"


dfc:
When I was looking at the replace config for another bug I realized that there was no entry for '^-'.  In light of the other defaults and the subsequent explanations it seems that looking out for pathnames that begin with a dash is a good candidate for inclusion. I looked and I did not find anything in my modest music collection but it does not mean its impossible. 

Nobody likes this:

```

dfc@ronin:/tmp/BeetsIsAwesome$ ls -l
total 8
drwxr-xr-x 2 dfc dfc 4096 Feb 17 22:35 Artist One
drwxr-xr-x 2 dfc dfc 4096 Feb 17 22:36 -Artist Two
dfc@ronin:/tmp/BeetsIsAwesome$ cd -Artist\ Two/
-bash: cd: -A: invalid option
cd: usage: cd [-L|[-P [-e]]] [dir]
dfc@ronin:/tmp/BeetsIsAwesome$

```

I realize that there are ways around this (`cd -- -ArtistTwo`) but it is minor inconvenience / interruption in normal command line work flow.  The default replace setting just needs an additional line;

```
'^-': _
```

PR #2509, opened 2017-04-12
Add default replace for - to _

mattmurch:
Also updated changelog and docs.

Resolves: #549

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -52,7 +52,7 @@ New features:
 
 Fixes:
 
-* :doc:`/beets/config_default`: Added default replace for - to _. :bug:`549`
+* :ref:`replace`: Added default replace for - to _. :bug:`549`
 * :doc:`/plugins/absubmit`: Do not filter for supported formats. :bug:`2471`
 * :doc:`/plugins/mpdupdate`: Fix Python 3 compatibility. :bug:`2381`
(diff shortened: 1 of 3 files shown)


sampsyo:
Thanks for the PR! But is #459 really what you meant to link to?

mattmurch:
Sorry, I meant #549. I'll push a fix in a second.

mattmurch:
Done

mattmurch:
There is a Travis CI failure in the line I introduced in the changelog. I wrote line 55 to mimic the surrounding code, but I do not understand what it is doing. Do you know what is happening?

docs/changelog.rst
```diff
@@ -52,6 +52,7 @@ New features:
 
 Fixes:
 
+* :doc:`/beets/config_default`: Added default replace for - to _. :bug:`549`
```
sampsyo: Aha; this is just a ReST syntax thing: the `:doc:` directive is for linking to other `*.rst` pages. We actually only really use this in the changelog for plugins, where the name of the ReST doc is the same as the name of the Python source file—here, you don't need to use a prefix. You might want to use ``:ref:`replace` `` inline in the sentence to link to the relevant config option in the configuration page, for example.

sampsyo:
Thanks! Merged, with a slightly updated changelog entry.

Issue #435, opened 2013-10-22
Move command should support copying files without modifying database


fortes:
Currently the `copy` flag on the `move` command will copy files _and_ change their paths in the database. It would be useful to have a version of this that copies without modifying the library.

BitterEarth:
Hey we would like to tackle this issue, if anyone is planning on working on this anytime soon let me know so we can find another issue. Thanks!

BeetsStudentTeam:
Hey guys! we would like to tackle this issue, if anyone is going to work on this soon let me know so we can look at other issues! We're planning to start working on this on Monday February 1st. Thanks! -Jake

SpirosChadoulos:
Greetings..I was thinking about tackling this issue but it is really old. Is it worth tackling it or not? Thanks a lot.

sampsyo:
Absolutely, if you're interested!

SpirosChadoulos:
Nice, could you explain me what you want me to do exactly?

sampsyo:
The idea would be to add a flag to the `move` command, with one of the names above, that works *like* the current -c (copy) option but slightly different. Where the current option does two things—it copies the files and also updates the library database—the new -e option would instead copy the file *but leave the database unchanged*. I think that will involve a call along the lines of `util.copy(item.path, item.destination(basedir=...))`.

SpirosChadoulos:
Should I do that in the commnad.py file or somewhere else?

sampsyo:
Right in commands.py is the right place.

SpirosChadoulos:
Could you explain to me briefly how the `move_items` function works in commands.py(line1453)? 
Thanks a lot.

sampsyo:
I'd be happy to explain more, but can I please ask you to be specific with your question? The comments we've put there are already my best effort at explaining what's going on. What part specifically are you wondering about?

SpirosChadoulos:
I see that the `move_items` function is where the move command is executed but I can distinguish where database interaction is happening.

sampsyo:
Aha! Good question. That's the `obj.remove` call, which invokes either `Item.remove` or `Album.remove`, which are defined in `library.py`.

SpirosChadoulos:
In line 1491 in command.py `obj.move` is called. `obj.remove` is called at line 1217. I am kind of confused with the code that is executed when the user uses the move command.

sampsyo:
[...]

PR #2510, opened 2017-04-13
New export command (issue #435)

SpirosChadoulos:
Could you help me with the errors that I am getting?

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1447,6 +1447,7 @@ modify_cmd.parser.add_option(
 modify_cmd.func = modify_func
 default_commands.append(modify_cmd)
 
+
 # move: Move/copy files to the library or a new base directory.
 
 def move_items(lib, dest, query, copy, album, pretend, export=False,
@@ -1479,9 +1480,10 @@ def move_items(lib, dest, query, copy, album, pretend, export=False,
             show_path_changes([(obj.path, obj.destination(basedir=dest))
                                for obj in objs])
     else:
-        #Copying files without modifying the database.
+        # Copying files without modifying the database.
         if export:
-            # export(file) function copies a file without modifying the database.
+            # export(file) function copies a file without modifying
+            # the database.
 
             def export(file):
                 # Create necessary ancestry for the copy.
@@ -1490,7 +1492,7 @@ def move_items(lib, dest, query, copy, album, pretend, export=False,
             if album:
                 for obj in objs:
                     for item in obj.items():
-                       export(item)
+                        export(item)
             else:
                 for obj in objs:
                     export(obj)
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1450,7 +1450,7 @@ default_commands.append(modify_cmd)
 
 # move: Move/copy files to the library or a new base directory.
 
-def move_items(lib, dest, query, copy, album, pretend, export, confirm=False):
+def move_items(lib, dest, query, copy, album, pretend, export=False, confirm=False):
     """Moves or copies items to a new base directory, given by dest. If
     dest is None, then the library's base directory is used, making the
     command "consolidate" files.
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1480,7 +1480,8 @@ def move_items(lib, dest, query, copy, album, pretend, export, confirm=False):
                                for obj in objs])
     else:
         if export:
-            util.copy(item.path, item.destination(basedir=dest))
+            for obj in objs:
+                util.copy(obj.path, obj.destination(basedir=dest))
         else:
             if confirm:
                 objs = ui.input_select_objects(
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1482,11 +1482,12 @@ def move_items(lib, dest, query, copy, album, pretend, export=False,
     else:
         if export:
             if album:
-                util.copy([(item.path, item.destination(basedir=dest))
-                                   for obj in objs for item in obj.items()])
+                for obj in objs:
+                    for item in obj.items():
+                        util.copy(item.path, item.destination(basedir=dest))
             else:
-                util.copy([(obj.path, obj.destination(basedir=dest))
-                                   for obj in objs])
+                for item in objs:
+                    util.copy(item.path, item.destination(basedir=dest))
         if confirm:
             objs = ui.input_select_objects(
                 u'Really %s' % act, objs,
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1447,13 +1447,6 @@ modify_cmd.parser.add_option(
 modify_cmd.func = modify_func
 default_commands.append(modify_cmd)
 
-#export(file) function copies a file without modifying the database.
-
-def export(file):
-    # Create necessary ancestry for the copy.
-    util.mkdirall(file.destination(basedir=dest))
-    util.copy(file.path, file.destination(basedir=dest))
-
 # move: Move/copy files to the library or a new base directory.
 
 def move_items(lib, dest, query, copy, album, pretend, export=False,
@@ -1488,6 +1481,12 @@ def move_items(lib, dest, query, copy, album, pretend, export=False,
     else:
         #Copying files without modifying the database.
         if export:
+            # export(file) function copies a file without modifying the database.
+
+            def export(file):
+                # Create necessary ancestry for the copy.
+                util.mkdirall(file.destination(basedir=dest))
+                util.copy(file.path, file.destination(basedir=dest))
             if album:
                 for obj in objs:
                     for item in obj.items():
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1480,30 +1480,19 @@ def move_items(lib, dest, query, copy, album, pretend, export=False, confirm=Fal
                                for obj in objs])
     else:
         if export:
-<<<<<<< HEAD
             for obj in objs:
                 util.copy(obj.path, obj.destination(basedir=dest))
         if confirm:
-=======
->>>>>>> 53618258faec970a0005b640d6811ad0aa04684a
             objs = ui.input_select_objects(
                 u'Really %s' % act, objs,
                 lambda o: show_path_changes(
                     [(o.path, o.destination(basedir=dest))]))
-            for obj in objs:
-                util.copy(obj.path, obj.destination(basedir=dest))
-        else:
-            if confirm:
-                objs = ui.input_select_objects(
-                    u'Really %s' % act, objs,
-                    lambda o: show_path_changes(
-                        [(o.path, o.destination(basedir=dest))]))
 
-                for obj in objs:
-                    log.debug(u'moving: {0}', util.displayable_path(obj.path))
+        for obj in objs:
+            log.debug(u'moving: {0}', util.displayable_path(obj.path))
 
-                    obj.move(copy, basedir=dest)
-                    obj.store()
+            obj.move(copy, basedir=dest)
+            obj.store()
 
 
 def move_func(lib, opts, args):
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1450,7 +1450,7 @@ default_commands.append(modify_cmd)
 
 # move: Move/copy files to the library or a new base directory.
 
-def move_items(lib, dest, query, copy, album, pretend, confirm=False, export):
+def move_items(lib, dest, query, copy, album, pretend, export, confirm=False):
     """Moves or copies items to a new base directory, given by dest. If
     dest is None, then the library's base directory is used, making the
     command "consolidate" files.
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1481,8 +1481,12 @@ def move_items(lib, dest, query, copy, album, pretend, export=False,
                                for obj in objs])
     else:
         if export:
-            for obj in objs:
-                util.copy(obj.path, obj.destination(basedir=dest))
+            if album:
+                util.copy([(item.path, item.destination(basedir=dest))
+                                   for obj in objs for item in obj.items()])
+            else:
+                util.copy([(obj.path, obj.destination(basedir=dest))
+                                   for obj in objs])
         if confirm:
             objs = ui.input_select_objects(
                 u'Really %s' % act, objs,
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1450,7 +1450,7 @@ default_commands.append(modify_cmd)
 
 # move: Move/copy files to the library or a new base directory.
 
-def move_items(lib, dest, query, copy, album, pretend, confirm=False):
+def move_items(lib, dest, query, copy, album, pretend, export=False, confirm=False):
     """Moves or copies items to a new base directory, given by dest. If
     dest is None, then the library's base directory is used, making the
     command "consolidate" files.
@@ -1479,6 +1479,9 @@ def move_items(lib, dest, query, copy, album, pretend, confirm=False):
             show_path_changes([(obj.path, obj.destination(basedir=dest))
                                for obj in objs])
     else:
+        if export:
+            for obj in objs:
+                util.copy(obj.path, obj.destination(basedir=dest))
         if confirm:
             objs = ui.input_select_objects(
                 u'Really %s' % act, objs,
@@ -1500,7 +1503,7 @@ def move_func(lib, opts, args):
             raise ui.UserError(u'no such directory: %s' % dest)
 
     move_items(lib, dest, decargs(args), opts.copy, opts.album, opts.pretend,
-               opts.timid)
+               opts.timid, opts.export)
 
 
 move_cmd = ui.Subcommand(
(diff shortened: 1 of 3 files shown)


beets/ui/commands.py
```diff
@@ -1450,7 +1450,7 @@ def modify_func(lib, opts, args):
 
 # move: Move/copy files to the library or a new base directory.
 
-def move_items(lib, dest, query, copy, album, pretend, confirm=False):
+def move_items(lib, dest, query, copy, album, pretend, confirm=False, export):
```
sampsyo: This:

> ERROR: Failure: SyntaxError (non-default argument follows default argument (commands.py, line 1453))

Means that export must either come before confirm or have a default value.

SpirosChadoulos: I still have an error at line 1483 with `item`

sampsyo: Can you please be a little more specific? I can't quite find the error you're talking about; maybe you can include the error text?

SpirosChadoulos: Unresolved reference 'item' less... (Ctrl+F1) 
This inspection detects names that should resolve but don't. Due to dynamic dispatch and duck typing, this is possible in a limited but useful number of cases. Top-level and class-level items are supported better than instance items.

beets/ui/commands.py
```diff
@@ -1479,17 +1479,20 @@ def move_items(lib, dest, query, copy, album, pretend, confirm=False):
             show_path_changes([(obj.path, obj.destination(basedir=dest))
                                for obj in objs])
     else:
-        if confirm:
-            objs = ui.input_select_objects(
-                u'Really %s' % act, objs,
-                lambda o: show_path_changes(
-                    [(o.path, o.destination(basedir=dest))]))
+        if export:
+            util.copy(item.path, item.destination(basedir=dest))
```
sampsyo: Is this the line you're referring to? If so, item is indeed an undefined variable. You will need to assign to it before using it.  Did you mean obj? But obj might be an item or an album, so you may need to check which it is.

SpirosChadoulos: this is the line I'm refering to `util.copy(item.path, item.destination(basedir=dest))`

SpirosChadoulos: @sampsyo how should I handle `item` ?

sampsyo: As I mentioned above, that variable is not defined. I suggested that you might have intended to use `obj` instead.

SpirosChadoulos: I used it and I got no errors but appveyor and travis failed..Could you help me with that?

sampsyo: This error:
https://travis-ci.org/beetbox/beets/jobs/222222532#L894

says that the tests are calling the `move_items` function without providing the new argument you've added. That gives you two options: you can either make the flag optional (provide a default value), or you can fix the tests to provide it every time they call the function. It would also be nice to have a test for this new functionality!

It might be helpful for you to set up the testing stuff locally so you can catch bugs during your development. There's more detail here:
https://github.com/beetbox/beets/wiki/Testing

But the short version is that you should install and run `nosetests`.

SpirosChadoulos: How can I provide a default value in order to make the flag optional?

sampsyo: It might be helpful to a little homework about how to define Python functions. There's lots of advice out there if you google "Python optional parameters". For example, these SO questions might help:
http://stackoverflow.com/q/43420535
http://stackoverflow.com/q/9539921

SpirosChadoulos: I provided a default value. Is this problem fixed now?

sampsyo: The usual way to answer that is to either run the tests yourself (instructions above) or wait for the CI results to come back and take a look at the output.

SpirosChadoulos: I took a look at the travis report but I can't completely understand what's the problem

sampsyo: If you need help understanding an error message, the thing to do is to paste a copy of it and try explaining what you think it means so far and which specific part is confusing you. Show that you've done some homework to try to understand it, and we'll be happy to help get you the rest of the way there!

SpirosChadoulos: File "/home/travis/build/beetbox/beets/test/test_ui.py", line 446, in test_move_album
    self.assertNotExists(self.itempath)
  File "/home/travis/build/beetbox/beets/test/_common.py", line 140, in assertNotExists
    u'file exists: {!r}'.format((path)))
AssertionError: True is not false : file exists: b'/tmp/tmpqr4bm0wx/testlibdir/srcfile'

This test asserted that the album didn't existed but it turned out that it existed?

sampsyo: Thanks! The next step is to look at the test code to see what it's doing, and reason about how your addition may have changed its behavior. In particular, this test is testing whether the `move` command actually moves files out of the way, and it's saying that the file that should be gone (i.e., moved way) is still there.

You can debug this by running the test locally and printing things out your code to see why the file didn't move away.

SpirosChadoulos: Okay I'll work on it.. Is the command I've used (`util.copy`) the right one for what we are trying to do?

sampsyo: Yep, absolutely!

SpirosChadoulos: Okay perfect. I will run the tests locally and try to fix all of them.

SpirosChadoulos: I should only care about the failures, not the errors right?

sampsyo: Errors are important too—that's when an unexpected exception is thrown.

SpirosChadoulos: I noticed something weird. AppVeyor and Travis say that I have 4 failures but when I run the command nosetests locally I get 3 failures and when I run the command python setup.py test locally I get 7 failures. How is that possible? Also in the local tests, no matter what I change in my `if export` statement the failures remain exactly the same.

sampsyo: That's odd! Which test do the two runs disagree on?

To help debug the failures, sometimes it can be helpful to try deleting the code you added and confirming that all the tests pass, as expected. Then you can add changes in one at a time to see when things break.

SpirosChadoulos: I deleted all the code I added to this file and the failures are still the same.. That's really weird..

SpirosChadoulos:
@sampsyo I fixed some issues and now my code passes the tests. Is everything okay?

sampsyo:
Yes; this looks great! Thank you for sorting out the various problems! :sparkles:

To make this ready to merge, we'll need at least an update to the docs (to describe the new option) and a changelog entry. It would also be super nice to have a unit test, which shouldn't be too hard—it can be pretty similar to a copy n' paste of one of the existing tests for the `move` command, but it should check that the item's path in the database doesn't change.

SpirosChadoulos:
Okay I'll do these the following days.

SpirosChadoulos:
Should I copy n' paste the MoveTest class (test_ui.py)??

sampsyo:
Yeah, that seems like a great place to start!

SpirosChadoulos:
@sampsyo I need some help with the errors I am getting because it is the first time I am writting a test for beets

sampsyo:
Cool! Happy to help, but to make this easy, would you mind pasting the specific errors you have questions about and give a little context? Then I can totally help with some answers.

SpirosChadoulos:
All my errors are about no such filed or directory errors

sampsyo:
OK! As usual, though, your part of the bargain is to lay out the problem clearly so your collaborators have a hope of being helpful. Specifically:

1. Copy and paste the error message.
2. Refer (perhaps link) to the part of the code where you think it's coming from.
3. Give your theories so far about what might be going wrong.
4. Ask specific questions about where you're confused.

SpirosChadoulos:
https://travis-ci.org/beetbox/beets/jobs/227346244#L887-L901

Here the error I get is AttributeError: no such field 'destination', but I can't figure out what this means and where this field is.
Also, I did what you told me and I copy n pasted the MoveTest in order to create the ExportTest and I added `export=True`. I noticed that the test methods that fail in the new class are those that go through the new code I added in commands.py(which you told me was okay). I can't figure out if the test methods need some tweaking or if the main code is the problem and the tests fail because the new feature isn't doing what it is supposed to do.

sampsyo:
Ah, I see—thanks for clarifying!

The problem here is that `obj` might be an Album object or it might be an Item object. On [this line](https://github.com/beetbox/beets/pull/2510/files#diff-37364792a6e3f85f399e29af135a4488R1485), your code assumes it's an Item—only Items have a `destination` method. You'll need an `if album:` condition, just like the one earlier in the method, to iterate over the album's items in album mode.

SpirosChadoulos:
Perfect, that was very helpfull! I will do exactly that.Thanks a lot for the help!!

SpirosChadoulos:
@sampsyo https://travis-ci.org/beetbox/beets/jobs/228798906#L887-L899
TypeError: copy() missing 1 required positional argument: 'dest'..Why is this happening? My copy function has a dest parameter. Didn't I did what you told me above the right way?

beets/ui/commands.py
```diff
@@ -1479,6 +1480,13 @@ def move_items(lib, dest, query, copy, album, pretend, confirm=False):
             show_path_changes([(obj.path, obj.destination(basedir=dest))
                                for obj in objs])
     else:
+        if export:
+            if album:
+                util.copy([(item.path, item.destination(basedir=dest))
+                                   for obj in objs for item in obj.items()])
+            else:
+                util.copy([(obj.path, obj.destination(basedir=dest))
+                                   for obj in objs])
```
sampsyo: Here, you're just passing one argument—a list, produced by a list comprehension—to the `copy` function. Maybe you mean something like:

```
for item in objs:
    util.copy(item.path, item.destination...)
```

and then again with `for item in obj.items()`.

SpirosChadoulos: Do you mean `objs.items()` ? Because `obj.items()` gives a Python error

SpirosChadoulos: https://travis-ci.org/beetbox/beets/jobs/228909615#L898
To which file is it refering to? Didn't I do what you explained right?

sampsyo:
OK, it sounds like you've gotten past the errors and are now finding real test failures! Can you please take a look at one of the specific tests you're working with and try to understand what it's trying to test? That might answer the central question: whether the test is wrong or the code is wrong.

SpirosChadoulos:
Let's take `test_move_item` (line 504). It tests if the file has reached it's new destination and if it is deleted from it's old destination, right? Isn't that what we want it to test?

SpirosChadoulos:
I replaced all `AssertNotExists` with `AssertExists` but the same file does not exist error remains

sampsyo:
Any theories as to why?

SpirosChadoulos:
Maybe I should change the paths in the assertions? `self.i.path` `self.itempath`

SpirosChadoulos:
@sampsyo 
https://travis-ci.org/beetbox/beets/jobs/229424482#L896-L897
I can't understand what this exact error is.

sampsyo:
Hmm; yes, that's not showing you the precise traceback. One way to test a theory is to run the test locally and add `print` statements around where you think the failure might be occurring. That can help you narrow down both where things are going wrong and why.

SpirosChadoulos:
I did some research and some testing and I think that the tests are okay and the problem is the `util.copy(item.path, item.destination(basedir=dest))` call, especially the 2 parameters the method gets.I also think that the problem is the destination path (2nd parameter) and not the first one. Any thoughts and suggestions?

sampsyo:
Hmm; what do you think the problem is there? That call looks (to me) like it copies the file from its original location to the path where it needs to go. Seems correct, I guess?

SpirosChadoulos:
But if this is the case why do we get this error? https://travis-ci.org/beetbox/beets/jobs/229424482#L983
Note that this error occurs just when `self._move()` is executed in our tests and not when the assertions happen.

sampsyo:
Good question. Since the error says this:

```
beets.util.FilesystemError: No such file or directory while copying /tmp/tmppqnmxzp9/testlibdir/srcfile to /tmp/tmppqnmxzp9/testotherdir/the artist/the album/02 full
```

that can actually mean either the source *or* the destination doesn't exist. I suspect it's the destination. Specifically, the destination path comes from here:

```
self.otherdir = os.path.join(self.temp_dir, b'testotherdir')
```

but that directory is never created!

The solution is probably to add code to the `export` functionality to create parent directories. For example, you can see that the `move` method on `Item` uses this code:

```
# Create necessary ancestry for the move.
util.mkdirall(dest)
```

to make sure that the destination directory will exist.

SpirosChadoulos:
I fixed it!! I think it works fine now. What should I do now in order to merge this pull request?

beets/ui/commands.py
```diff
@@ -1479,6 +1480,19 @@ def move_items(lib, dest, query, copy, album, pretend, confirm=False):
             show_path_changes([(obj.path, obj.destination(basedir=dest))
                                for obj in objs])
     else:
+        if export:
+
```
sampsyo: Could you please clean up this extra blank space, and instead add a brief comment about what's going on here?

beets/ui/commands.py
```diff
@@ -1479,6 +1480,19 @@ def move_items(lib, dest, query, copy, album, pretend, confirm=False):
...
+                for obj in objs:
+                    for item in obj.items():
+                        # Create necessary ancestry for the copy.
+                        util.mkdirall(item.destination(basedir=dest))
+                        util.copy(item.path, item.destination(basedir=dest))
+            else:
+                for obj in objs:
+                    # Create necessary ancestry for the copy.
+                    util.mkdirall(obj.destination(basedir=dest))
+                    util.copy(obj.path, obj.destination(basedir=dest))
```
sampsyo: Since these two lines are identical to the above, a "helper function" to encapsulate them would help with maintainability.

test/test_ui.py
```diff
@@ -822,37 +899,37 @@ def test_cli_config_file_overwrites_beetsdir_defaults(self):
...
+    #
+    #        with open(cli_config_path, 'w') as file:
+    #            file.write('anoption: value')
+    #
+    #        with open(cli_overwrite_config_path, 'w') as file:
+    #            file.write('anoption: overwrite')
+    #
+    #        self.run_command('--config', cli_config_path,
+    #                      '--config', cli_overwrite_config_path, 'test')
+    #        self.assertEqual(config['anoption'].get(), 'cli overwrite')
```
sampsyo: This change looks like it's just unrelated noise—would you mind reverting it?

test/test_ui.py
```diff
@@ -477,6 +475,85 @@ def test_pretend_move_album(self):
...
+        self.assertTrue(b'testlibdir' in self.i.path)
+        self.assertExists(self.i.path)
+        self.assertNotExists(self.itempath)
+
+    def test_copy_item(self):
+        self._move(copy=True)
+        self.i.load()
+        self.assertTrue(b'testlibdir' in self.i.path)
+        self.assertExists(self.i.path)
+        self.assertExists(self.itempath)
```
sampsyo: Export mode ignores the `copy` flag, so this test is probably redundant and can be removed. (The album versions with this flag are the same way.)

test/test_ui.py
```diff
@@ -477,6 +475,85 @@ def test_pretend_move_album(self):
...
+
+        # Alternate destination directory.
+        self.otherdir = os.path.join(self.temp_dir, b'testotherdir')
+
+    def _move(self, query=(), dest=None, copy=False, album=False,
+              pretend=False, export=True):
+        commands.move_items(self.lib, dest, query, copy, album,
+                            pretend, export)
+
+    def test_move_item(self):
```
sampsyo: `test_move_item` no longer really accurately describes what's going on here. Maybe you mean `test_export_item`? (And `move` -> `export` in other test names below?)

test/test_ui.py
```diff
@@ -477,6 +475,85 @@ def test_pretend_move_album(self):
...
+        self.assertTrue(b'testlibdir' in self.i.path)
+        self.assertExists(self.i.path)
+        self.assertExists(self.itempath)
+
+    def test_move_album(self):
+        self._move(album=True)
+        self.i.load()
+        self.assertTrue(b'testlibdir' in self.i.path)
+        self.assertExists(self.i.path)
+        self.assertNotExists(self.itempath)
```
sampsyo: The `move` flag is also ignored, right? So this test seems similarly redundant.

sampsyo (changes requested):
Yay yay yay! That's great! :tada: :sparkles:

I have a few code suggestions above, where the two themes are removing redundancy and thinking carefully about the naming of test cases.

There's one last thing we'll need in order to merge: documentation! Can you please add a brief description of the new flag to `docs/reference/cli.rst` under the `move` command?

beets/ui/commands.py
SpirosChadoulos: Do you mean lines 1489&1490 and 1494&1495?? They are not exactly identical, one uses `item` and the other uses `obj`. Do you want me to create a helper function that takes as a parameter either `item` or `obj`?

sampsyo: Yes, exactly.

SpirosChadoulos:
@sampsyo Are all of your change requests covered? Do you want anything else before I write the docs?

sampsyo:
Yes; this all looks great! I think the last missing piece is the docs.

SpirosChadoulos:
Ready to merge?

SpirosChadoulos:
@sampsyo Sorry for taging you but I thought you might have missed my comment..Are we ready to merge?

sampsyo:
Hi! Sorry, I did see your comment—things look good, and it's on my to-do list to do a complete review and merge.

SpirosChadoulos:
Okay thanks a lot!!

sampsyo:
Sorry I missed the May deadline! I finally had a moment to give this the attention it deserved. I simplified a few things, revised the tests, fixed some parameter-passing issues, and merged. :sparkles: Thank you for your help!

Issue #2513, opened 2017-04-13
Date-valued field queries allow non-date values


discopatrick:
### Problem

When querying on date-valued fields, such as added and mtime, non-date values (e.g. 'abcd') are allowed, and seem to output the full collection list. For example:

```sh
$ beet -vv list mtime:abcd
no user configuration found at /Users/patrick/.config/beets/config.yaml
data directory: /Users/patrick/.config/beets
plugin paths: 
Sending event: pluginload
library database: /Users/patrick/.config/beets/library.db
library directory: /Users/patrick/Music
Sending event: library_opened
... every item in my library ...
Sending event: cli_exit
```

I'm currently working on #2506 to extend the date parser, but I feel that this issue needs to be fixed first.

### Setup

* OS: OSX 10.11.6 El Capitan
* Python version: 2.7.13
* beets version: 1.4.4 (actually this commit: https://github.com/beetbox/beets/commit/45bed808d39fae2845ba4addbd71f31be2e485d4)
* Turning off plugins made problem go away (yes/no): No, no plugins active.

My configuration (output of `beet config`) is:

```yaml
{}
```

sampsyo:
Good call—this should probably raise an `InvalidQueryArgumentTypeError`.

discopatrick:
I'll start work on this.

sampsyo:
Fixed in #2517.

Issue #2506, opened 2017-04-10
Extend the query date parser with (optional) times


discopatrick:
I had a brief chat with @sampsyo about this on the beets irc channel. I just wanted to formally register myself as beginning this job here.

We can currently query based on date:
```
beet ls 'added:2008-12-01..2008-12-02'
```

The plan is to additionally allow querying on time:
```
beet ls 'added:2008-12-01 12:30:55..2008-12-02 19:15:22'
```

Without having looked at the code yet, I'm wondering whether the following are going to be problematic:
a) the space between date and time,
b) the colon symbols within the time.

If anyone can confirm this off the top of their head, and can advise on an approach, please do. Maybe it will work just fine. Maybe we'll need to change `2008-12-01 12:30:55` to `2008-12-01-12-30-55`.

Beets can also do shorthand date queries, like:
```
beet ls 'added:2008-12..2009-10-11'
```

I would also like to be able to do shorthand time queries, like this:
```
beet ls 'added:2008-12-01 12:..2008-12-02 19:15:'
```
i.e. list items added on or after 2008-12-01 12:00:00, but before 2008-12-02 19:16:00

I'll shout if I get stuck!

discopatrick:
Development on this feature is now underway: https://github.com/discopatrick/beets/commit/ba324df0d1f17e4ffb7de97a580a0fe33610097b

So far, it has seemed fairly logical to me that date queries including times should use the 'T' separator format, e.g.: `added:2000-01-01T12:30..2000-01-01T12:31`

However, I can't decide whether using the 'T' separator makes the query more readable or less readable than if we just used a space:

With minute precision:
`'added:2000-01-01 12:30..2000-01-01 12:31'` vs `added:2000-01-01T12:30..2000-01-01T12:31`

With hour precision:
`'added:2000-01-01 12..2000-01-01 13'` vs `added:2000-01-01T12..2000-01-01T13`

The 'T' separator makes it quite clear that what follows is a time value. This is most apparent when using hour precision - it's not immediately clear what the lone `12` and `13` stand for in the example using spaces. The 'T' also has the added benefit of keeping the value as one string of characters with no whitespace, and means you don't have to quote the query.

On the other hand, using a space separator keeps the format consistent with that which `beet list` outputs:
```bash
patrick$ ./beet list added:2017-04-13T11:43..2017-04-13T11:43
2017-04-13 11:43:08 - Danny Byrd - Supersized - Shock Out
[...]

PR #2517, opened 2017-04-14
Date value field validation

discopatrick:
This pull request is for issue #2513 (Date-valued field queries allow non-date values).
 
`Period.parse()` now raises a `InvalidQueryArgumentTypeError` when the string does not parse to a date - although I do wonder whether this is the correct error to raise, as I'm not comparing types here. I wonder whether a `ValueError`, or perhaps creating a new `InvalidQueryArgumentValueError` is more appropriate?

I have changed the way `Period.parse()` works - it used to first count the number of hyphens in the date to determine the precision (day, month, or year) and then try to parse the string against the appropriate date format. This did work fine. However, in anticipation of new feature #2506 (Extend the query date parser with (optional) times), where we will have '-', ':', and possibly ' ' or 'T' characters to count, I have opted for a different approach. I now test the string against each of the formats supported, from most precise to least precise, until a match is found (or not).

Tests have been updated to assert that the new errors are raised.

The output of a bad date query now reads quite nicely in the output:
```
$ beet list added:2017-01..2017-aa
invalid query: 'added:2017-01..2017-aa': '2017-aa' is not a valid datetime string
```

Finally, I have fixed a couple of older tests which were attempting to test `mtime` queries with an odd looking value of `0.0` and updated them to correct values.

`tox` passes all tests.

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -561,7 +561,7 @@ class Period(object):
             date = datetime.strptime(string, date_format)
         except ValueError:
             # Parsing failed.
-            raise InvalidQueryArgumentTypeError(string, datetime)
+            raise InvalidQueryArgumentTypeError(string, 'a valid datetime string')
         precision = cls.precisions[ordinal]
         return cls(date, precision)
 
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -561,7 +561,7 @@ class Period(object):
             date = datetime.strptime(string, date_format)
         except ValueError:
             # Parsing failed.
-            return None
+            raise InvalidQueryArgumentTypeError(string, datetime)
         precision = cls.precisions[ordinal]
         return cls(date, precision)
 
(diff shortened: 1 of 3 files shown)


sampsyo:
Yay! Thanks for getting this started!

> Period.parse() now raises a InvalidQueryArgumentTypeError when the string does not parse to a date - although I do wonder whether this is the correct error to raise, as I'm not comparing types here. I wonder whether a ValueError, or perhaps creating a new InvalidQueryArgumentValueError is more appropriate?

Sure. In fact, I think we only use this exception as more akin to `ValueError`, not `TypeError`. So perhaps we should just rename the exception.

> I have changed the way Period.parse() works -

Would you mind putting this refactoring into a separate pull request? Having one logical change at a time makes things a lot easier to review.

discopatrick:
Sure @sampsyo, I'm splitting into two pull requests now.

I will leave the renaming of the exception to yet another pull request, to keep things tidy.

sampsyo:
Awesome; thanks! :sparkles: I merged this one with a changelog entry.

PR #2518, opened 2017-04-15
Fix misuse of flags in re.sub() calls

jwilk:
The 4th argument of `re.sub()` is maximum number of substitutions, not flags.

Found using [pydiatra](https://github.com/jwilk/pydiatra).


sampsyo:
Great; thanks! Out of curiosity, were there any other red flags in your report from that tool?

jwilk:
> were there any other red flags in your report from that tool?

Apart from #2526, it was only complaining about [bare except](https://docs.python.org/2/howto/doanddont.html#except).

wordofglass:
There's already a PR about the bare excepts, I really should polish and merge that one.

PR #2521, opened 2017-04-18
add --yes argument to play command

tweitzel:
I am Super Lazy and bound a script to auto start shuffling my music when I log into my work machine in the morning and got tired of having to yes 'C' |  to it. So I added -f to beet play so I can bypass the warning_threshold at runtime.

You may disagree that this should be in the plugin; no hard feelings if so.

### DIFF
--- a/beetsplug/play.py
+++ b/beetsplug/play.py
@@ -82,9 +82,9 @@ class PlayPlugin(BeetsPlugin):
             help=u'add additional arguments to the command',
         )
         play_command.parser.add_option(
-            u'-f', u'--force',
+            u'-y', u'--yes',
             action="store_true",
-            help=u'disable the warning threshold',
+            help=u'skip the warning threshold',
         )
         play_command.func = self._play_command
         return [play_command]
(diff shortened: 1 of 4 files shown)


nathdwek:
In your particular use case, you could juste use a simple yaml file like so:

```
#conf.yaml
play:
     warning_threshold=0
```

and just run `beet -c conf.yaml play` to shuffle your entire library.

tweitzel:
An entire second config file just to turn off one feature seems like overkill to me.

beetsplug/play.py
```diff
@@ -125,7 +130,7 @@ def _play_command(self, lib, opts, args):
 
         # Check if the selection exceeds configured threshold. If True,
         # cancel, otherwise proceed with play command.
-        if not self._exceeds_threshold(selection, command_str, open_args,
+        if opts.force or not self._exceeds_threshold(selection, command_str, open_args,
```
jackwilsdon: It looks like this is longer than 80 characters now. Do you think you could wrap the `self._exceeds_threshold` call so that it's back under 80 (and to appease the Travis CI gods)?

Thanks!

jackwilsdon (changes requested):

sampsyo:
Yo! FWIW, this option is called "--yes" in other commands: https://github.com/beetbox/beets/blob/master/beetsplug/convert.py#L158

nathdwek:
@tweitzel I agree in principle, but I think a separate config file for scripts is not unheard of. It was just meant as a nicer stopgap solution while this PR gets through :)

jackwilsdon (approved):

nathdwek (approved):

sampsyo:
Perfect! Thank you!! :sparkles:

PR #2522, opened 2017-04-20
Date value field validation method

discopatrick:
As discussed in pull request #2517, here is a separate pull request for a refactoring of the method by which date queries are validated. This is in anticipation of adding three more datetime formats and precisions for hour, minute, and second, in new feature #2506.

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -533,8 +533,8 @@ class Period(object):
     instants of time during January 2014.
     """
 
-    precisions = ('day', 'month', 'year')
-    date_formats = ('%Y-%m-%d', '%Y-%m', '%Y')
+    precisions = ('year', 'month', 'day')
+    date_formats = ('%Y', '%Y-%m', '%Y-%m-%d')
 
     def __init__(self, date, precision):
         """Create a period with the given date (a `datetime` object) and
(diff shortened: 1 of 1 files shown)


discopatrick:
@sampsyo - I'm guessing it's better for you to create the changelog entry yourself once you're happy for a merge? Better for it to be created at the moment of merging to master, as only then will you know the correct number? Or should other developers write a changelog entry, but without a number?

beets/dbcore/query.py
```diff
@@ -553,16 +553,16 @@ def parse(cls, string):
...
-            # Parsing failed.
+        ordinal = 0
+        date = None
+        for date_format in cls.date_formats:
+            try:
+                date = datetime.strptime(string, date_format)
+                break
+            except ValueError:
+                # Parsing failed.
+                ordinal += 1
```
sampsyo: What do you think about using

```
for ordinal, date_format in enumerate(cls.date_formats):
```

instead of manually incrementing `ordinal`? If that can work, it might be a little more readable.

discopatrick: Good call. Declaring, looping over, and incrementing `ordinal` on separate lines is not as robust as doing it all on one line.

Come to think of it, the declaration of `ordinal` before the loop wasn't necessary in the first place because in Python loop variables are still in scope after the loop has finished.

sampsyo:
Great! This looks perfect. Thanks for refactoring a messy bit of code. :sparkles:

PR #2526, opened 2017-04-23
Don't hardcode errno constant

jwilk:
The value of `ENOENT` is architecture-dependent, so don't assume it's always 2.

Found using [pydiatra](https://github.com/jwilk/pydiatra).


sampsyo:
Good call; thanks again.

PR #2538, opened 2017-05-01
Tests to track whether lyrics websites are correctly fetched

Kraymer:
This PR goal is to check that the different backends (and every site registered as source for the Google backend) work as expected.

[Some old code](https://github.com/beetbox/beets/blob/master/test/test_lyrics.py#L197) was supposed to do that in the tests but was never executed as it required some files on the filesystem that were never there.
Because these tests don't mock the fetching part, they are more integration tests than unit tests and should not be executed by default. A `skipUnless` decorator is applied requiring an environment variable to be set explicitely for the tests to be executed.

Here are the results (we should take some actions concerning the state of the default backends) :

**Default plugin backends :** 
Only LyricsWiki and musixmatch pass the tests. 
- [ ] lyric.com urls structure has changed and can't be deduced anymore from artist and title, replaced by a magic number eg http://www.lyrics.com/lyric/2493457 
- [ ] after few attempts we get 404 from musixmatch backend (works ok with no user agent set)
- [ ] Genius backend `get_lyrics` method fails. cc @sadatay

**Google backend sources :**
Some websites have been removed from the domains considered by beets google custom search engine as they don't pass the tests.
The google backend searches now amongst 17 lyrics websites.

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -52,7 +52,7 @@ except ImportError:
 
 from beets import plugins
 from beets import ui
-
+import beets
 
 DIV_RE = re.compile(r'<(/?)div>?', re.I)
 COMMENT_RE = re.compile(r'<!--.*-->', re.S)
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -52,7 +52,6 @@ except ImportError:
 
 from beets import plugins
 from beets import ui
-import beets
 
 
 DIV_RE = re.compile(r'<(/?)div>?', re.I)
@@ -260,6 +259,7 @@ class MusiXmatch(SymbolsReplaced):
 
     def fetch(self, artist, title):
         url = self.build_url(artist, title)
+
         html = self.fetch_url(url)
         if not html:
             return
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -21,6 +21,7 @@ from __future__ import absolute_import, division, print_function
 import difflib
 import itertools
 import json
+import struct
 import re
 import requests
 import unicodedata
@@ -77,6 +78,11 @@ USER_AGENT = 'beets/{}'.format(beets.__version__)
 
 # Utilities.
 
+def unichar(i):
+    try:
+        return six.unichr(i)
+    except ValueError:
+        return struct.pack('i', i).decode('utf-32')
 
 def unescape(text):
     """Resolve &#xxx; HTML entities (and some others)."""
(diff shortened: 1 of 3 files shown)


beetsplug/lyrics.py
```diff
@@ -77,6 +78,11 @@ class HTMLParseError(Exception):
 
 # Utilities.
 
+def unichar(i):
+    try:
+        return six.unichr(i)
+    except ValueError:
+        return struct.pack('i', i).decode('utf-32')
```
sampsyo: Is there a particular reason why this is necessary? Are some files encoded with UTF-32 (which is a fairly rare format)?

beetsplug/lyrics.py
```diff
@@ -257,8 +263,8 @@ def fetch(self, artist, title):
         html = self.fetch_url(url)
         if not html:
             return
-        lyrics = extract_text_between(html,
-                                      '"body":', '"language":')
+        lyrics = extract_text_between(html, '<p class="mxm-lyrics__content" data-reactid="135">',
+            '</p>')
```
sampsyo: This extraction seems a little brittle… maybe we can find a way to avoid including the "data-reactid".

beetsplug/lyrics.py
Kraymer: I had such an exception occuring once in the `unichar` code with one of the page returned by the Google backend.

beetsplug/lyrics.py
Kraymer: Done.

Kraymer:
So it seems lyric.com url scheme is completely opaque now, ok to remove it from `sources` ?

sampsyo:
Ah, that is too bad about lyrics.com. Yes, let's remove it—but perhaps that should be a separate PR?

Kraymer:
OK to make this PR only add tests, others PR to fix them. In that vein, I reinstated the failing sources in `LyricsPluginSourcesTest`, which gives us :

~~~
$ export BEETS_TEST_LYRICS_SOURCES=1
$ nosetests -s -v test/test_lyrics.py
FAIL: Test default backends with songs known to exist in respective databases.
  File "/Users/flap/Dev/beets/test/test_lyrics.py", line 316, in test_backend_sources_ok
    self.assertFalse(errors)
AssertionError: ['LyricsCom', 'MusiXmatch', 'Genius'] is not false
~~~

beetsplug/lyrics.py
sampsyo: That sounds sensible—people could have invalid entities. But I guess I'm wondering whether UTF-32 is really the right thing to assume—perhaps we should just drop the character? (Eventually, we'll run into a situation where an entity is so crazy it can't even be decoded with UTF-32 packed as a multi-byte integer. :cry:)

Issue #1610, opened 2015-09-17
fetchart plugin documentation for itunes store is out of date


bengadbois:
https://github.com/sampsyo/beets/blob/master/docs/plugins/fetchart.rst#itunes-store

```
$ pip install python-itunes
Collecting python-itunes
  Could not find a version that satisfies the requirement python-itunes (from versions: )
No matching distribution found for python-itunes
```

You cannot ```pip install python-itunes``` anymore

Edit: appears to be a pip bug in recent versions, but the following works: ```https://pypi.python.org/packages/source/p/python-itunes/python-itunes.tar.gz ```

bengadbois:
```
$ pip install python-itunes
Collecting python-itunes
  Could not find a version that satisfies the requirement python-itunes (from versions: )
No matching distribution found for python-itunes
```
Perhaps this is a pip bug...?

Somasis:
That sounds more like an issue with `python-itunes` being python2 only.

```
somasis@cicero:~/ % pip --version
pip 7.1.2 from /usr/x86_64-pc-linux-gnu/lib/python3.4/site-packages (python 3.4)
somasis@cicero:~/ % pip2 --version
pip 7.1.2 from /usr/lib/python2.7/site-packages (python 2.7)
somasis@cicero:~/ % pip show python-itunes
somasis@cicero:~/ % pip2 show python-itunes
---
Metadata-Version: 1.0
Name: python-itunes
Version: 1.0
Summary: A simple python wrapper to access iTunes Store API
Home-page: https://github.com/ocelma/python-itunes
Author: Oscar Celma
Author-email: ocelma@bmat.com
License: http://www.gnu.org/copyleft/gpl.html
Location: /usr/x86_64-pc-linux-gnu/lib/python2.7/site-packages
Requires:
```

sampsyo:
Weirdly; I'm having the same trouble:

	$ pip2 install --user python-itunes11:11:43PM jasmine:~
	Collecting python-itunes
	  Could not find a version that satisfies the requirement python-itunes (from versions: )
	No matching distribution found for python-itunes

I'm actually not sure what that means. Any insight?

jackwilsdon:
It seems to be a known problem with python-itunes. Check out ocelma/python-itunes#9 for a bit more information. Currently you can do this to install python-itunes: `pip install -e git+https://github.com/ocelma/python-itunes.git#egg=itunes` (as suggested by @caxap), however it does feel a bit 'hacky'.

[...]

Issue #2371, opened 2017-01-06
fetchart  python3 compatibility when using itunes


PierreRust:
### Problem

This not exactly a problem on beets it-self, but when using fetchart on a python 3 installation it fails on itune import as the module used is compatible with python 3

Running this command:

```sh
 beet fetchart genre:Classical
```

Led to this problem:

```
** error loading plugin fetchart:
Traceback (most recent call last):
  File "/home/pierre/.env_beets/lib/python3.5/site-packages/beets/plugins.py", line 254, in load_plugins
    namespace = __import__(modname, None, None)
  File "/home/pierre/.env_beets/lib/python3.5/site-packages/beetsplug/fetchart.py", line 39, in <module>
    import itunes
  File "/home/pierre/.env_beets/lib/python3.5/site-packages/itunes/__init__.py", line 94
    except urllib2.HTTPError, e:
```

Of course, the best solution (and probably only solution, except patchin python-itunes) is not to use itune as a source at all. I could be enough to mention this in the documentation.

sampsyo:
Hmm; did that traceback get cut off? Looks like we're missing the all-important last line.

And yeah, on first glance it looks like that module is not py3-compatible. We should make it an optional import (if it's not already) and add a note to the docs.

jrobeson:
I didn't even know we imported another module. I took the list in `setup.py` for plugin deps as gospel. I guess that was wrong. :(   I guess it's time to do a search for all imports.

PR #2540, opened 2017-05-03
Remove mention of python-itunes

owcz:
The package is inactive and has been broken for months: https://github.com/beetbox/beets/issues/2371

Fixes https://github.com/beetbox/beets/issues/1610

Alternatively, could be rephrased to say it doesn't work with py3 (if that's the only condition), but I think that would cause even more confusion if there isn't a workaround

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -100,6 +100,11 @@ Fixes:
   ignored. Thanks to :user:`discopatrick`. :bug:`2513` :bug:`2517`
 * When the SQLite database stops being accessible, we now print a friendly
   error message. Thanks to :user:`Mary011196`. :bug:`1676` :bug:`2508`
+* :doc:`plugins/fetchart`'s iTunes Store artwork lookup no longer recommended
+  in documentation, as the unmaintained `python-itunes`_ is broken. Want
+  to adopt it? :bug:`2371` :bug:`1610`
+
+.. _python-itunes: https://github.com/ocelma/python-itunes
 
 
 1.4.3 (January 9, 2017)
--- a/docs/plugins/fetchart.rst
+++ b/docs/plugins/fetchart.rst
@@ -49,7 +49,7 @@ file. The available options are:
   (``enforce_ratio: 0.5%``). Default: ``no``.
 - **sources**: List of sources to search for images. An asterisk `*` expands
   to all available sources.
-  Default: ``filesystem coverart itunes amazon albumart``, i.e., everything but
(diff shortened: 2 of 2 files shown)


sampsyo:
Good call; thanks. Would you mind adding a changelog entry to this effect?

Kraymer:
Is keeping the iTunes Store section necessary ?

> iTunes Store
> There is currently no plugin to use the iTunes Store as an art source.

sampsyo:
Yep, good point—we can probably remove that section altogether. (Beyond that, this is totally ready to merge—anybody with commit access should feel free to close it up.) Thanks again!

owcz:
I recommend keeping the section (or at least a mention) because otherwise users (like me) will wonder where fetchart's iTunes support is. iTunes' artwork is some of the most reliable on the Internet—hopefully someone will fill in this functionality soon

Kraymer:
Gotcha, yet I think it's pretty clear in the first sentence that iTunes artwork is not available :

> By default, this plugin searches for art in the local filesystem as well as on the Cover Art Archive, the iTunes Store, Amazon, and AlbumArt.org, in that order. 

Have no strong opinion though and the commit to revert is `e7ea7ab5f` if @sampsyo prefers.

owcz:
As far as I can tell, as currently merged, https://github.com/beetbox/beets/blob/master/docs/plugins/fetchart.rst doesn't include that sentence or any other mention of iTunes

Kraymer:
It's the first sentence of that section :
https://github.com/beetbox/beets/blob/master/docs/plugins/fetchart.rst#album-art-sources

owcz:
Right, but not with "iTunes"—the point is that "iTunes" isn't mentioned at all on the page so someone looking for iTunes `fetchart` compatibility has no idea what to do next. Anyway, your call on how to resolve, though I suggest giving users (such as myself) a heads up within the documentation as that's where I went to look first (before finding that the plugin indeed no longer works)

sampsyo:
Thanks for editing the docs, @owcz, and for merging, @Kraymer.

Our usual MO in these situations is to advertise the change loudly and clearly in the changelog instead of in the documentation pages themselves. There's of course a trade-off, but keeping the core documentation changes free of historical cruft has been helpful in making them clear for new users, who just want to see what's currently there. I think our announcement in the changelog could still be louder and clearer, but we'll get there.

Kraymer:
@owcz in the meantime, if iTunes fetchart support is something you want back, you could open a dedicated feature request issue that would bring visibility

Issue #2532, opened 2017-04-30
web: Crash when serializing objects with bytes fields


ocelotsloth:
### Problem

For some reason the web interface is not working correctly. I can't actually get it to search for anything. When you type something into the search box nothing happens except for the following text being printed to the terminal.

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv web
```
Then searching for something in the web interface gives this problem:

```
~
➔ beet -vv web
 * Running on http://127.0.0.1:8337/ (Press CTRL+C to quit)
127.0.0.1 - - [30/Apr/2017 11:05:15] "GET / HTTP/1.1" 200 -
127.0.0.1 - - [30/Apr/2017 11:05:22] "GET /item/query/isabel HTTP/1.1" 200 -
Error on request:
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/werkzeug/serving.py", line 209, in run_wsgi
    execute(self.server.app)
  File "/usr/lib/python3.6/site-packages/werkzeug/serving.py", line 199, in execute
    for data in application_iter:
  File "/usr/lib/python3.6/site-packages/werkzeug/wsgi.py", line 704, in __next__
    return self._next()
  File "/usr/lib/python3.6/site-packages/werkzeug/wrappers.py", line 81, in _iter_encoded
    for item in iterable:
  File "/usr/lib/python3.6/site-packages/beetsplug/web/__init__.py", line 74, in json_generator
    yield json.dumps(_rep(item, expand=expand))
  File "/usr/lib/python3.6/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.6/json/encoder.py", line 180, in default
    o.__class__.__name__)
TypeError: Object of type 'bytes' is not JSON serializable
```

### Setup

* OS: Arch Linux
* Python version: Python 3.6
* beets version: 1.4.3
* Turning off plugins made problem go away (yes/no): no, the problem is with the web plugin

My configuration (output of `beet config`) is:

```yaml
directory: ~/Music
library: ~/Music/musicLibrary.db

import:
    move: yes

plugins: chroma web
acoustid:
    apikey: REDACTED
web:
    host: 127.0.0.1
[...]

PR #2542, opened 2017-05-03
web: #2532 - Decode bytes values to strings

ocelotsloth:
Looks to solve #2532 by decoding any bytes objects to strings.

I'm not as familiar with the iterable constructs in Python, but I think that I came up with a reasonable solution. I'm open to learning a better way to do this if there is one.

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -44,7 +44,8 @@ def _rep(obj, expand=False):
             del out['path']
 
         # Filter all bytes attributes and convert them to strings
-        for key in filter(lambda key: isinstance(out[key], bytes), out):
+        for key, value in out.items():
+            if isinstance(out[key], bytes):
                 out[key] = base64.b64encode(out[key]).decode('ascii')
 
         # Get the size (in bytes) of the backing file. This is useful
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -42,6 +42,10 @@ def _rep(obj, expand=False):
         else:
             del out['path']
 
+        # Filter all bytes attributes and convert them to strings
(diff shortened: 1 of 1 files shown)


beetsplug/web/__init__.py
```diff
@@ -42,6 +42,10 @@ def _rep(obj, expand=False):
         else:
             del out['path']
 
+        # Filter all bytes attributes and convert them to strings
+        for key in filter(lambda key: isinstance(out[key], bytes), out):
```
sampsyo: A somewhat more "idiomatic Python" way of doing this would be to use a plain `if`. For example:

```py
for key, value in out.items():
    if isinstance(value, bytes):
        out[key] = ...
```

beetsplug/web/__init__.py
```diff
@@ -42,6 +42,10 @@ def _rep(obj, expand=False):
...
+            out[key] = out[key].decode('utf-8')
```
sampsyo: Using `.decode('utf-8')` will assume that the binary data stored on the object happens to be a UTF-8-encoded string. That's not necessarily the case: the acoustid fingerprint, for example, is really a raw binary blob, not any representation of text at all.

Fortunately, there's a [standard library function](https://docs.python.org/2/library/base64.html#base64.b64encode) that can encode bytes as ASCII-safe bytes. Then you can use `.decode('ascii')` to get a text string to pass to the JSON encoder.

ocelotsloth:
Thanks for the comments @sampsyo!

I went ahead and made the modifications you suggested, and it definitely looks a bit simpler.

sampsyo:
Thank you for your help with this! :sparkles: I've merged with a changelog entry.

Issue #2548, opened 2017-05-03
Lyrics.com backend expect a wrong url scheme


Kraymer:
### Problem

Lyrics.com backend is broken.

```sh
$ export BEETS_TEST_LYRICS_SOURCES=1
$ nosetests -s -v test/test_lyrics.py
```

Led to this problem:

```
======================================================================
FAIL: Test default backends with songs known to exist in respective databases.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/flap/Dev/beets/test/test_lyrics.py", line 316, in test_backend_sources_ok
    self.assertFalse(errors)
AssertionError: ['LyricsCom', 'MusiXmatch', 'Genius'] is not false
-------------------- >> begin captured logging << --------------------
[...]
requests.packages.urllib3.connectionpool: DEBUG: Starting new HTTP connection (1): www.lyrics.com
requests.packages.urllib3.connectionpool: DEBUG: http://www.lyrics.com:80 "GET /the-beatles-lyrics-lady-madonna.html HTTP/1.1" 301 20
requests.packages.urllib3.connectionpool: DEBUG: http://www.lyrics.com:80 "GET /lyrics/the%20beatles%20by%20lady%20madonna HTTP/1.1" 200 10372
```

The song page *real* url is *http://www.lyrics.com/lyric/9853981* whereas the plugin try to fetch *http://www.lyrics.com/lyrics/the%20beatles%20by%20lady%20madonna*.

PR #2549, opened 2017-05-03
Drop lyrics.com backend

Kraymer:
Address #2548 by dropping the backend as I see no solution to get from (artist, title) to the magic number used in the url.


sampsyo:
Thanks for opening the separate PR! I merged and updated the changelog entry.

PR #2552, opened 2017-05-05
Recompress png file

ibmibmibm:


jackwilsdon:
```bash
-rw-r--r--@ 1 jack  staff  36607  5 May 19:49 before.png
-rw-r--r--@ 1 jack  staff  33782  5 May 19:49 after.png
```

Gotta save them 2,825 bytes!

Issue #2557, opened 2017-05-11
replaygain: R128_{TRACK,ALBUM}_GAIN in Opus


autrimpo:
The replaygain plugin using the bs1770gain backend with EBU method produces four REPLAYGAIN_ tags with Opus files. This is wrong, as it should only add two tags, R128_TRACK_GAIN and R128_ALBUM_GAIN, as described in the [RFC](https://tools.ietf.org/html/rfc7845). For example MPD [only looks for the track gain tag](https://github.com/MusicPlayerDaemon/MPD/blob/a3e28c2d1a1c85351ef5fe4e50ec75db6f4199c4/src/decoder/plugins/OpusTags.cxx#L44) and ignores any REPLAYGAIN_ tags (as it should). This obviously makes the replaygain plugin broken when used with Opus.

sampsyo:
Hi! That plugin in meant to work in a backend-agnostic way: it computes gain adjustments using any analysis tool and stores normalization results in a common format. I guess I don't quite understand why the Ogg Opus encapsulation standard's recommendation forbids other gain formats working also—is there a general reference for that idea aside from the MPD behavior?

If it make sense, we could consider using the R128 name too to maximize compatibility.

autrimpo:
To quote the linked RFC:

> To avoid confusion with multiple normalization schemes, an Opus
   comment header SHOULD NOT contain any of the REPLAYGAIN_TRACK_GAIN,
   REPLAYGAIN_TRACK_PEAK, REPLAYGAIN_ALBUM_GAIN, or
   REPLAYGAIN_ALBUM_PEAK tags, unless they are only to be used in some
   context where there is guaranteed to be no such confusion.
   [EBU-R128] normalization is preferred to the earlier REPLAYGAIN
   schemes because of its clear definition and adoption by industry.
   Peak normalizations are difficult to calculate reliably for lossy
   codecs because of variation in excursion heights due to decoder
   differences.  In the authors' investigations, they were not applied
   consistently or broadly enough to merit inclusion here.

There's also this message on the [mailing list](https://www.ietf.org/mail-archive/web/codec/current/msg03103.html):

>I can't remember if it's been said on the list before, but "true peak" is a meaningless concept for a lossy format with a non-bit-exact decoder specification, like Opus. The variation in the true peak calculation depending on your decoder implementation is theoretically unbounded. So you're guaranteed the value going into this tag is garbage.

[...]

PR #919, opened 2014-08-25
NULL fields and tag removal

geigerzaehler:
1. Deleting fields works by setting them to `NULL` in the database
2. The Zero plugin completely removes tags instead of storing the empty string (see #157).

PR #2560, opened 2017-05-13
RFC: replaygain: R128 support

autrimpo:
Following the discussion in #2557 I've decided to try implement a very WIP, 'proof of concept' support of R128 to the *replaygain* plugin.

As I'm not very familiar with the codebase nor Python I'm creating a PR already and would appreciate if someone more knowledgable could review the changes.

Things that I'd especially like comments on:

- I'm saving and restoring the backend because R128 is only supported by `bs1770` as far as I know. Is that needed? If yes, perhaps somehow detect and create the second backend in advance and switch between those two instead of recreating it every time.
- There's a lot of shared code between the branches right now, should it branch only for those 2 specific calls or perhaps even branch *inside* those calls?
- How to decide when to branch. My idea was a new plugin setting, a list of formats with a sensible default (so right now only **Opus**).
- The MediaType and db entries - they work, but are they correct?
- Removal of `REPLAYGAIN_` tags when tagging with *override* set - right now, my files are incorrectly tagged with these and I'd like the `R128_` to replace them. When/where should it be done, should it be configurable?
- Expanding on the 'branching on list' and the previous point - is tagging with both systems desirable and how should it be handled?

This only calculates track gain, album gain is still missing. But I've tested it on MPD and DeadBeef and the gain produced with these changes works as expected.


beets/mediafile.py
```diff
@@ -2005,6 +2005,15 @@ def update(self, dict):
         out_type=float,
     )
 
+    # EBU R128 fields.
+    r128_track_gain = MediaField(
+        StorageStyle(
+            u'R128_TRACK_GAIN',
+            float_places=0
```
sampsyo: You don't need `float_places` if you don't intend to store floating-point numbers. Is this field supposed to store an integer?

beets/mediafile.py
```diff
@@ -2005,6 +2005,15 @@ def update(self, dict):
...
+        ),
```
sampsyo: With just a `StorageStyle`, this will only support "free-form" tag formats. On MP3, MPEG-4, and WMA, it will crash unceremoniously. I think the right thing to do is to fill in logical names for these other formats, even if support isn't widespread yet.

beetsplug/replaygain.py
```diff
@@ -138,6 +144,17 @@ def compute_album_gain(self, album):
...
 
+    def compute_track_r128_gain(self, items):
+        """Computes the track gain of the given tracks, returns a list
+        of TrackGain objects.
+        """
+
+        method = self.method
+        self.method = '--ebu'
+        output = self.compute_gain(items, False)
+        self.method = method
```
sampsyo: Hmm, yes—mutating and restoring the object's state is a bit hacky. Maybe this decision should be made inside `compute_gain`? The idea would be to write a clear policy about when the "format-specific default" overrides the global default or the user's configuration.

beetsplug/replaygain.py
```diff
@@ -905,21 +928,55 @@ def handle_track(self, item, write):
...
+                    )
+
+                self.store_track_gain(item, track_gains[0])
+                if write:
+                    item.try_write()
+            except ReplayGainError as e:
+                self._log.info(u"ReplayGain error: {0}", e)
+            except FatalReplayGainError as e:
+                raise ui.UserError(
+                    u"Fatal replay gain error: {0}".format(e))
```
sampsyo: Hmm; I see what you mean about duplicate structure between the two branches. It does seem like a good idea to trim this down, but I don't have any non-obvious suggestions at the moment…

beetsplug/replaygain.py
```diff
@@ -905,21 +928,55 @@ def handle_track(self, item, write):
...
         self._log.info(u'analyzing {0}', item)
 
-        try:
-            track_gains = self.backend_instance.compute_track_gain([item])
-            if len(track_gains) != 1:
-                raise ReplayGainError(
-                    u"ReplayGain backend failed for track {0}".format(item)
+        self._log.debug(u'format: {0}', item.format)
+
+        if item.format == 'Opus':
```
sampsyo: Maybe some manner of global function to make this decision (e.g., `should_use_r128`) would be useful to encapsulate the policy here?

sampsyo:
> Removal of REPLAYGAIN_ tags when tagging with override set - right now, my files are incorrectly tagged with these and I'd like the R128_ to replace them. When/where should it be done, should it be configurable?

Good question. I don't have a strong feeling whether it should be configurable (maybe some people want both sets of tags for maximum compatibility? Or maybe we should force compliance with the RFC's recommendation, so adding R128 tags clobbers RG tags but not the other way around?). I imagine it should be done in the `store_track_gain`/`store_album_gain` functions (or rather, the new versions of the same).

beets/mediafile.py
autrimpo: I've added support for other formats, with names more or less correspoinding to their replaygain equivalents.

autrimpo:
I've cleaned up the code a little, the decision making is now done by a method and controlled by a setting (whitelist of formats). Unfortunately I can't delete the `REPLAYGAIN_` tags, only set their value to be empty. Any hints on how to do that would be appreciated.

sampsyo:
Sadly, actually deleting the tag is a little more difficult than it should be. (Real, general support for deleting tags is a long-standing debacle—see #919, for example.) The actual answer is that we need a way to invoke `del mediafile.rg_track_gain` inside the `Item.write()` method when the corresponding `Item` field is `None`. As pedantic as this sounds, can we perhaps address this in a separate PR? Adding this tag-deletion ability would be really useful in general.

beetsplug/replaygain.py
```diff
@@ -822,6 +823,9 @@ def __init__(self):
         if self.config['auto']:
             self.import_stages = [self.imported]
 
+        # Formats to use R128.
+        self.r128_whitelist = self.config['r128'].as_str_seq()
```
sampsyo: Cool! This configurable format list makes sense to me. (We'll need to add it to the documentation eventually—unless we want to leave it undocumented, if we think people shouldn't be fiddling with it.)

beetsplug/replaygain.py
```diff
@@ -830,9 +834,16 @@ def __init__(self):
             raise ui.UserError(
                 u'replaygain initialization failed: {0}'.format(e))
 
+        self.r128_backend_instance = ''
+
+    def should_use_r128(self, item):
+        return item.format in self.r128_whitelist
```
sampsyo: A docstring for this method might be useful.

beetsplug/replaygain.py
```diff
@@ -921,6 +950,23 @@ def handle_track(self, item, write):
...
+
+            try:
+                self.r128_backend_instance = self.backends[backend_name](
+                    self.config, self._log
+                )
+            except (ReplayGainError, FatalReplayGainError) as e:
+                raise ui.UserError(
+                    u'replaygain initialization failed: {0}'.format(e))
+
+            self.r128_backend_instance.method = '--ebu'
```
sampsyo: This method body looks like it might have gotten indented one step too far.

autrimpo:
I see, in that case I'll skip the tag removal functionality for now. I'll add the tests, and I of course plan to update the plugin documentation to reflect the final implementation.

autrimpo:
I've added album gain, so this can be considered a 'final draft'.
However, the ALAC and MP4 tests are failing with Python 3. I've tried to debug it but don't know what the problem is.

beets/mediafile.py
```diff
@@ -2005,6 +2005,62 @@ def update(self, dict):
...
+            index=0, desc=u'iTunR128',
+            id3_lang='eng'
+        ),
+        MP4StorageStyle(
+            '----:com.apple.iTunes:r128_track_gain'
+        ),
+        MP4SoundCheckStorageStyle(
+            '----:com.apple.iTunes:iTunR128',
+            index=0
+        ),
```
sampsyo: I'd recommend dropping the SoundCheck storage styles (this and the MP3 variant above). The SC format is included for the RG tags for compatibility with the iTunes-specific normalization metadata, which we should probably (eventually) break off as a *third* standard that deserves a third set of tags.

We can also cross our fingers and hope that this solves the crash on Python 3…

sampsyo:
Great! I think the error probably has to do with the SoundCheck conversion stuff, so we may be able to sidestep it altogether. In short, the MediaFile abstraction layer is trying to data stored as bytes in the MPEG-4 format as a Unicode string, which won't do.

Good point about heterogenous albums. Yeah, just logging an error and "giving up" on the album (raising a ReplayGainError) would be nice—and slightly better than silently assuming the format from the first track on the album.

autrimpo:
Alright, I've finally fixed the tests.

MP4 tags starting with `----:` are encoded as unicode. The *float* branch of `_safe_cast` handled the decoding fine but the *int* branch had no such logic, that's why `REPLAYGAIN_` passed fine but `R128_` struggled. I've also took the liberty to adjust the regex a little, to support a sign prefix. I think the spec allows both in the tag and `-` is actually very common, as with replaygain values.

If everything looks fine I'll rewrite the history to something prettier (I was thinking 3 commits - the `_safe_cast` fix, the tests and finally the implementation itself) and update the documentation (fourth commit).

beets/mediafile.py
```diff
@@ -154,10 +154,12 @@ def _safe_cast(out_type, val):
             return int(val)
         else:
             # Process any other type as a string.
-            if not isinstance(val, six.string_types):
+            if isinstance(val, bytes):
+                val = val.decode('utf-8', 'ignore')
+            else:
```
sampsyo: Cool; this looks right. Just to clarify what's going on, may I recommend `elif not isinstance(val, six.string_types)` here?

beets/mediafile.py
```diff
@@ -154,10 +154,12 @@ def _safe_cast(out_type, val):
...
                 val = six.text_type(val)
             # Get a number from the front of the string.
-            val = re.match(r'[0-9]*', val.strip()).group(0)
+            val = re.match(r'[\+-]?[0-9]*', val.strip()).group(0)
```
sampsyo: :+1: Good call.

beets/mediafile.py
```diff
@@ -2005,6 +2007,44 @@ def update(self, dict):
...
     )
 
+    # EBU R128 fields.
+    r128_track_gain = MediaField(
+        MP3DescStorageStyle(
+            u'R128_TRACK_GAIN'
+        ),
+        MP3DescStorageStyle(
+            u'r128_track_gain'
+        ),
```
sampsyo: It belatedly occurs to me to recommend that we don't try use capitalizations here. It was useful for RG, where there are apparently both styles in use, but here we're essentially "inventing a standard"—I don't see any prior art yet indicating one or the other. We can always add both later if someone complains.

The only problem is I'm not sure which to choose… maybe lower case?

autrimpo: Well, the RFC always refers to it in upper case, so I guess we should go with that for consistency.

autrimpo:
Cleaned up, updated the docs.

sampsyo:
Thank you for your patience! I've finally done a complete review and merged. This looks great. :sparkles:

autrimpo:
Glad to hear that!

PR #2569, opened 2017-05-18
beets2kodi plugin

peace899:
Based on the discussions [here](https://github.com/beetbox/beets/issues/2413).
I decided to try my hand with python and it's been interesting and came up with this plugin to generate xml format artist and album nfos.

Please check it out and provide feedback. The plugin is [here](https://github.com/peace899/beets/blob/master/beetsplug/kodinfo.py)
Thanks


Freso:
If what it does is create .nfo files for Kodi, then make a name like "kodinfo" or "nfo-for-kodi" would make more sense? "beets2kodi" leaves me very confused about what its purpose is anyway. :/ (I'm using both beets and Kodi too.)

peace899:
There's already a plugin called kodinfo which creates .nfo files with just the MB url for artist or album pages.
Will try to come up with a new name.

sampsyo:
Cool! Can you please help clear up one fundamental thing I don't understand? What's the relationship between these XML files and the textual, MBID-only files? Are they both called "NFO" files, confusingly? Do they serve the same purpose for Kodi, or different purposes?

If they're substantially related, maybe it even makes sense to have one plugin, called `kodinfo`, that just has a configuration option to switch between the two modes.

peace899:
Yes they are both called "NFO" files. They ultimately serve the same purpose of providing 'information' about your collection. This informattion can be album/artist rating, reviews, biographies and extra images.

The difference might be that the XML file points to information locally stored while the MBID only text file fetches remote data. I guess it's a matter of preference of the user, and I think it will be prudent to merge the two plugins.

Will get started on the travis issues, maybe over the weekend.

sampsyo:
Cool! It does sound like a single-plugin arrangement might work best.

peace899:
New version with name change. See #2473

docs/plugins/kodinfo.rst
```diff
@@ -0,0 +1,39 @@
+KodiNfo Plugin
+=================
+
+The plugin lets you create nfo files for Kodi music
+library and update it whenever you import an album into the library. 
+The plugin relies on the information provided by beets library and the audiodb
+(TADB). It only uses MusicBrainz album IDs, empty ID fields or discogs ID 
+are ignored.
```
sampsyo: Just saying this:

> It uses MusicBrainz IDs to look up metadata.

should suffice, I think.

docs/plugins/kodinfo.rst
```diff
@@ -0,0 +1,39 @@
+KodiNfo Plugin
+=================
+
+The plugin lets you create nfo files for Kodi music
+library and update it whenever you import an album into the library.
```
sampsyo: Is there some sort of link we can use here to explain more about what *.nfo files are and why you would want them?

docs/plugins/kodinfo.rst
```diff
@@ -0,0 +1,39 @@
...
+        host: localhost
+        port: 8080
+        user: kodi
+        pwd: kodi
+        nfo_format: xml
+
+    
+For the nfo_format key, choices are 'xml' or 'mbid_only_text'.
+The choice 'xml' produces XML type document, while the 'mbid_only_text'
+produces a text file containing the MusicBrainz url of the artist or album.
```
sampsyo: Similar to the above, can we explain a little bit more here (perhaps using links) why there are two different formats and how a user should think about choosing between them?

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,495 @@
...
+
+from beets import config
+import beets.library
+from beets.plugins import BeetsPlugin
+
+from pathlib import Path
+
+from lxml import etree as et
+
+import simplejson
```
sampsyo: Python now has a built-in `json` module that supplants `simplejson`.

docs/plugins/kodinfo.rst
```diff
@@ -0,0 +1,39 @@
...
+
+To use the ``kodinfo`` plugin you need  (requests, urllib.request, lxml, 
+simplejson, base64) modules.
```
sampsyo: `urllib`, `json`, and `base64` are Python standard library modules, so we don't need to mention them here. In addition, it's standard in the beets docs to mention dependencies you need to install from pip at the very top of the documentation along with some brief instructions (see `lyrics` or `lastgenre` for examples).

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,495 @@
...
+    def create_nfos(self, lib, album):
+        """Create nfos as per choice in config."""
+        try:
+            # Check if MBID is valid UUID as per MB recommendations
+            UUID(album.mb_albumid)
+            self._log.info(u'Album ID is valid MBID...creating .nfos')
+            albumid = 'mb_albumid:' + album.mb_albumid
+            mb_albumid = album.mb_albumid
+            mb_artistid = album.mb_albumartistid
+            nfo_format = '{0}'.format(config['kodi']['nfo_format'])
```
sampsyo: Any particular reason to use `'{0}'.format` instead of just saying `config['kodi']['nfo_format'].as_str()`?

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,495 @@
...
+            if nfo_format in 'mbid_only_text':
```
sampsyo: I recommend a shorter name like `text` for this configuration option—it could be a little less confusing.

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,495 @@
...
+            xml = et.tostring(
+                root,
+                pretty_print=True,
+                xml_declaration=True,
+                encoding='UTF-8',
+                standalone="yes").decode()
+            print(xml, file=open(artistnfo, 'w+'))
+
+
+def update_kodi(host, port, user, password):
```
sampsyo: Hmm… we already have a separate plugin to update the Kodi library (`kodiupdate`). Can we leave that responsibility to that plugin and just take care of the *.nfo file generation in this plugin?

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,495 @@
...
+from urllib.request import Request, urlopen
+from uuid import UUID
+
+from beets import config
+import beets.library
+from beets.plugins import BeetsPlugin
+
+from pathlib import Path
+
+from lxml import etree as et
```
sampsyo: Perhaps we can avoid depending on the lxml library when the user just wants plain-text files (no XML) by `import`ing the library conditionally.

docs/plugins/kodinfo.rst
peace899: Added info and links.

docs/plugins/kodinfo.rst
peace899: Yes, it makes sense. Changed to reflect that.

docs/plugins/kodinfo.rst
peace899: More info added in .rst file.

beetsplug/kodinfo.py
peace899: Done!

beetsplug/kodinfo.py
peace899: Removed and left with kodiupdate plugin.

beetsplug/kodinfo.py
peace899: Done. Changed to 'text'

beetsplug/kodinfo.py
peace899: Changed!

beetsplug/kodinfo.py
peace899: The changes have been effected. 
I guess I have to thank beets, for finally getting around to learning and most importantly understanding Python! Never touched python before trying to make this plugin. Back to learning!

peace899:
`There's one fundamental question I have with the design here, however: it looks like this plugin downloads data from an external metadata source (The Audio DB) just to write data into the *.nfo files. I think this would be better accomplished with two composable plugins: one to get the data from this database, if you think it's valuable, and store it in the beets database. (This way, other functionality—not just the *.nfo file writer—can enjoy the benefits of the enriched data.) Then, this plugin can stay simple and focused: it just generates the appropriate files from data you already have in your database; no database API lookups required. This kind of orthogonality, IMO, is what makes beets awesome—what do you think about keeping things simple here and delegating the extra metadata source to its own little plugin?`

Mhhh...this will take me some time to figure out but interisting in that it will let me use more databases like Discogs, Last.FM and MusicBrainz itself, since these databases contains some info required for the .nfo files.

docs/plugins/kodinfo.rst
peace899: I have removed the mentioning of standard modules as dependencies, and listed the required dependency module (lxml) and how to install it.

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,448 @@
...
+                "method": "Files.GetSources",
+                "params": {"media": "music"},
+                "id": 1}
+        json_data = json.dumps(data).encode('utf-8')
+        request = Request(url, json_data, headers)
+        result = json.load(urlopen(request))
+        _kodi_path = result['result']['sources'][0]['file']
+        return _kodi_path
+    except (requests.exceptions.RequestException, ValueError, TypeError):
+        return ''
```
sampsyo: It looks like the configuration for communicating with Kodi is just used to get Kodi's configured root path for music. Is that right? If so, might it be a little bit easier to just make *that* a configuration option? It's a little less elegant than auto-discovery, but there's also less chance of things going wrong.

sampsyo: Also, I guess I don't 100% understand why the root path is necessary. Is the idea that there might be music in your beets library at location `/path/A/B/C/foo.mp3`, but you have *another* copy that's in your Kodi library at `/path/X/Y/Z/foo.mp3`? A little more context here would help me greatly with understanding what's going on.

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,448 @@
...
+
+
+def album_path(albumid):
+    """Get album path."""
+    out_data = []
+    for album in lib.albums(albumid):
+        album_path = album.path.decode("utf-8")
+        root = str(config["directory"])
+        length = int(len(root) + 1)
+        kodi_lib_path = kodi_path() + album_path[length:]
```
sampsyo: Not every path in the beets database starts with the `directory` path; it's possible to have out-of-library files. So `album_path[length:]` could create a malformed path. (As above, it would be helpful to know why this "transplanting" of paths is necessary.)

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,448 @@
...
+            y = int(folder_length - indices[0])
+            artistpath_items = list(Path(albumpath).parts[-y:-1])
+            artist_path = os.path.join(root, *artistpath_items)
+            kodi_lib_path = kodi_path() + artist_path[int(len(root) + 1):]
+        out_data = artist_path, kodi_lib_path
+    return out_data
+
+
+def thumbs(tag, albumid):
+    """Name paths where art files reside."""
```
sampsyo: Can the docstring explain what `tag` is? (I suspect `albumid` is an MBID?)

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,448 @@
...
+            path = et.SubElement(root, 'path')
+            path.text = album_path(albumid)[i]
+
+        if album_info(albumid)[14] == '':
+            for i in range(len(thumbs('album', albumid))):
+                thumb = et.SubElement(root, 'thumb')
+                thumb.text = thumbs('album', albumid)[i]
+        else:
+            thumb = et.SubElement(root, 'thumb')
+            thumb.text = album_info(albumid)[14]
```
sampsyo: This looks like it calls `album_info` three times, which will fetch the same result from the metadata server three times. Maybe it would be wise to store the result and reuse it? (Or, this is part of why I'm suggesting that this part of the plugin be moved somewhere else—the data from the server can just be stored in the database.)

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,448 @@
...
+                                       (data2["strTheme"]) or '',
+                                       (comp),
+                                       (data2["strReview"]) or '',
+                                       (data[7]),
+                                       (rel_date),
+                                       (data[6]),
+                                       (data2["intScore"]) or '',
+                                       (date[0]),
+                                       (data2["strAlbumThumb"]) or ''))
+        return list(out_data.split(';'))
```
sampsyo: Formatting a string with semicolons and then splitting again on semicolons seems like it could cause trouble… is there any particular reason why this shouldn't just return a list? (Or, better yet, a dictionary with descriptive names for each field?)

beetsplug/kodinfo.py
```diff
@@ -0,0 +1,448 @@
...
+"""
+from __future__ import absolute_import, division, print_function
+
+import base64
+import json
+import os
+import time
+import requests
+import urllib.request
+from urllib.request import Request, urlopen
```
sampsyo: Does this library need to use both `requests` *and* `urllib`?

peace899:
Will look into the code quiries later, just busy with other projects.

**The kodi path confusion**:
beets and Kodi share the same music library. They can have the same paths but if you share your Kodi library with other computers or Kodi capable devices (which Kodi calls clients) in your network then it becomes imperative to use a file sharing protocol like nfs or samba. This is where the `kodi_lib_path` comes in, for example: 
An album can be in `/home/user/Music/Adele/25` and you import the directory` /home/user/Music ` to Kodi and you using nfs protocol, the `kodi_lib_path` of the same album will be 
`nfs://servername/Music/Adele/25`. This enables other clients to access the music. Hope it makes sense!

sampsyo:
Aha; I see! Thanks for explaining. And, just to clarify, the XML files need to have absolute paths in them? (I ask because using relative paths in those files would seem to sidestep the issue.)

beetsplug/kodinfo.py
peace899: > It looks like the configuration for communicating with Kodi is just used to get Kodi's configured root path for music. Is that right? If so, might it be a little bit easier to just make that a configuration option? It's a little less elegant than auto-discovery, but there's also less chance of things going wrong.

Still working on this, let me pick your brain a little how is this option:

```
def kodi_path(sources_xml):
    
    """ Get Kodi music library path or cifs/nfs url for shared libraries."""
    """ For config purposes, sources_xml = config['kodi']['sources_file']. """
    """ This discovery method uses the sources.xml file stored in the Kodi user
        directory. The user directory is usually in:
        ~/.kodi/userdata for Linux, %APPDATA%\kodi\userdata for Windows,
        /Users/<your_user_name>/Library/Application Support/Kodi/userdata/ for
        Mac, etc. Please see http://kodi.wiki/view/The_UserData_Folder for your
        system. """
        
    import xml.etree.ElementTree as ET
    
    tree = ET.parse(sources_xml)
    root = tree.getroot()

    default_source = tree.findtext('music/default')

    sources_dict = {}
    for source in tree.find('music'):
        if source.findtext('name'):
            sources_dict[source.findtext('name')] = source.findtext('path')

    if default_source:
        path = sources_dict[default_source]
    else:
        path = tree.findtext('music/source/path')
    return path

```

sampsyo: I see! This can automatically pick up the base directory by reading an XML, which seems somewhat simpler—but I was proposing something even simpler: just have the user put the base directory itself into the configuration.

peace899:
@sampsyo Would something like this work to create another 'plugin' to add flexible fields to AlbumInfo?

      
    def is_valid_uuid(mbid):
        """Check if MBID is valid UUID as per MB recommendations."""
        try:
            uuid_obj = UUID(mbid)
            return True
        except:
            return False

 
    def get_tadb_info(release_id, artist_id):
        """Collect artist and album information from theAudioDB.com."""

        artist_url = audiodb_url + "{0}/artist-mb.php?i={1}".format(
            audio_db_key, artist_id)

        album_url = audiodb_url + "{0}/album-mb.php?i={1}".format(
            audio_db_key, release_id)

        required_artist_keys = ["strStyle",
                                "strMood",
                                "intBornYear",
                                "strGenre",
                                "intFormedYear",
                                "strBiographyEN",
                                "intDiedYear",
                                "strDisbanded",
                                "strArtistThumb",
                                "strArtistFanart"]

        required_album_keys = ["strStyle",
                               "strMood",
                               "strTheme",
                               "strReview",
                               "intScore",
                               "strAlbumThumb"]

        try:
            response_ar = urlopen(artist_url)
            tadb_artist_info = json.load(response_ar)["artists"][0]
        
            response_al = urlopen(album_url)
            tadb_album_info = json.load(response_al)["album"][0]

            tadb_artist_info = {k: tadb_artist_info[k] for k in
                                required_artist_keys}

            tadb_album_info = {k: tadb_album_info[k] for k in
                               required_album_keys}

        except (ValueError, TypeError):
            """ Create NoneType or empty values for info dictionaries."""
            _none_list_album = [''] * len(required_album_keys)
            _none_list_artist = [''] * len(required_artist_keys)

            # Make dict values empty or NoneType
            tadb_album_info = dict(zip(required_album_keys, _none_list_album))
            tadb_artist_info = dict(zip(required_artist_keys, _none_list_artist))

        return tadb_album_info, tadb_artist_info


    class TADB(BeetsPlugin):
        """TADB Plugin."""

        def __init__(self):
            """Plugin docstring."""
            super(TADB, self).__init__()

            self.register_listener('album_imported', self.add_tadb_metadata)

        def add_tadb_metadata(self, lib, album):
            """Extend AlbumInfo with TADB metadata."""
        
            if is_valid_uuid(album.mb_albumid):
                # Validate mbid.
                self._log.info(u'Fetching and adding TheAudioDB.com info...')
            
                tadb_album_info, tadb_artist_info = get_tadb_info(
                    album.mb_releasegroupid,
                    album.mb_albumartistid)

                # Store artist info into library
                album.artist_bio = tadb_artist_info["strBiographyEN"]
                album.artist_genre = tadb_artist_info["strGenre"]
                album.artist_style = tadb_artist_info["strStyle"]
                album.artist_born_year = tadb_artist_info["intBornYear"]
                album.artist_formed_year = tadb_artist_info["intFormedYear"]
                album.artist_mood = tadb_artist_info["strMood"]
                album.artist_died_year = tadb_artist_info["intDiedYear"]
                album.artist_disbanded = tadb_artist_info["strDisbanded"]
                album.artist_thumb = tadb_artist_info["strArtistThumb"]
                album.artist_fanart = tadb_artist_info["strArtistFanart"]

                # Store album info into library
                album.album_review = tadb_album_info["strReview"]
                album.album_mood = tadb_album_info["strMood"]
                album.album_style = tadb_album_info["strStyle"]
                album.album_theme = tadb_album_info["strTheme"]
                album.album_rating = tadb_album_info["intScore"]
                album.album_thumb = tadb_album_info["strAlbumThumb"]
                album.store()

            else:
                self._log.info(u"Album ID is not valid MBID...can't use TADB")
                # Future use other APIs like Last.fm, WikiPedia or Discogs

sampsyo:
Yeah! Something like that would be a great way to avoid re-fetching data.

jtpavlock:
@peace899 looks like you've put a lot of awesome work into this! Any chance you'd be willing to continue it?

Issue #1881, opened 2016-02-13
Set field values on the import command line


t-8ch:
Usage: `beet import --explicit-field context=party --explicit-field category=disco some_folder`

I have different categories of music which I don't want to mix. For this is would be useful to be able to specify explicit metadata on import.

sampsyo:
Thanks! I'm in favor of this feature. I could have sworn we'd discussed this before, but I can't find any other issues about it (#565 discusses something *vaguely* similar). Perhaps it was on the mailing list?

t-8ch:
I never read the ML, so I have no clue whether it has been been proposed before. Maybe #488?

sampsyo:
Yeah, that's also related. Well, I'm glad we have an issue now!

acastaner:
I would totally love this feature too. I have thousands of artists and the way beets stores by default is great, but generates a ton of directories (one for each artist). This makes it impractical (for me anyway) to look for media when browsing, and causes long delays over network shares (CIFS sends all the folder names at once).

Being able to arbitrarily split the library some more (by genre or otherwise) allows to break this down more and I feel this is critical for people with a large library.

Kraymer:
> Being able to arbitrarily split the library some more (by genre or otherwise) allows to break this down more and I feel this is critical for people with a large library.

@acastaner any existing metadata (not just artist) can be used for path configuration, cf https://beets.readthedocs.org/en/v1.3.17/reference/config.html#path-format-configuration

[bucket plugin](https://beets.readthedocs.org/en/v1.3.17/plugins/bucket.html?highlight=buckets) can be of some use to resolve your problem of network delay due to huge folder listings.

acastaner:
Mostly my need (and what this feature request is about) is to set the field value upon import. This way it's not required to "pre-tag" the library we want to import, or manually tag after import.


For instance my beets is setup so that anything with the "savedir" field with value "metal" goes to the "metal_hard-rock" directory. But beets will act on this only through the mod command, not import. So I have to import everything then manually beet mod all the files. Not practical. If we could do it upon import then that issue is solved because the operation happens onthe specified directory not through filters/queries.


I hope I'm clear and making sense :-)





On Sun, Apr 17, 2016 at 1:40 PM -0700, "Fabrice Laporte" <notifications@github.com> wrote:





> Being able to arbitrarily split the library some more (by genre or otherwise) allows to break this down more and I feel this is critical for people with a large library.

[...]

PR #2581, opened 2017-06-01
Set field values on the import command line #1881

bartkl:
My first contribution, so be gentle ;).

- I've tested for both album and singleton import modes, using both as-is and autotagged import, and it seems to work fine.
- I extended the test-suite of the importer with `test_set_fields` functions in the `SingletonImportTest` and `ImportTest` classes.
- I added documentation in `cli.rst` and `config.rst`, and updated the changelog.
- `tox` seems to approve of my style and the tests (including mine) pass

I can imagine the `_store_dict` custom callback action I created should be moved elsewhere, but I do think this generic approach is better than a more helper function for the `importer_cmd` specifically.

I'd love to hear your comments!


beets/config_default.yaml
```diff
@@ -25,7 +25,8 @@ import:
     pretend: false
     search_ids: []
     duplicate_action: ask
-    bell: no  
+    bell: no
```
bartkl: removed a redundant space after 'bell: no'

beets/importer.py
```diff
@@ -530,6 +533,17 @@ def remove_duplicates(self, lib):
                 util.prune_dirs(os.path.dirname(item.path),
                                 lib.directory)
 
+    def set_fields(self):
```
sampsyo: A docstring here might be nice to have! (Perhaps even nicer than the summary in the overall "step sequence" in the overall class docstring.)

beets/importer.py
```diff
@@ -530,6 +533,17 @@ def remove_duplicates(self, lib):
...
+        set_fields_dict = config['import']['set_fields']
+        for field in set_fields_dict.keys():
```
sampsyo: A simpler way to iterate over Confit dicts:

```py
for key, view in config['import']['set_fields'].items():
    value = view.get()
```

beets/importer.py
```diff
@@ -530,6 +533,17 @@ def remove_duplicates(self, lib):
...
+            value = set_fields_dict[field].get()
+            log.debug(u'Set field {1}={2} for {0}',
+                      displayable_path(self.paths),
+                      field,
+                      value)
+            self.album[field] = value
+            self.album.store()
```
sampsyo: I recommend *not* including a `store` call here. Instead, we can just modify the values before they ever hit the database at all! Another comment below will elaborate…

beets/importer.py
```diff
@@ -1385,6 +1412,11 @@ def apply_choice(session, task):
 
     task.add(session.lib)
 
+    # If ``set_fields`` is set, set those fields to the
+    # configured values.
+    if config['import']['set_fields']:
+        task.set_fields()
```
sampsyo: I recommend moving this stanza above the call to `task.add`. This way, we'll modify the values on the Album before ever putting them in the database. This will be marginally more efficient and, more importantly, will be a little more predictable.

beets/importer.py
```diff
@@ -877,6 +891,19 @@ def choose_match(self, session):
...
+        """
+        set_fields_dict = config['import']['set_fields']
+        for field in set_fields_dict.keys():
+            value = set_fields_dict[field].get()
+            log.debug(u'Set field {1}={2} for {0}',
+                      displayable_path(self.paths),
+                      field,
+                      value)
+            self.item[field] = value
+            self.item.store()
```
sampsyo: (As above.)

beets/ui/commands.py
```diff
@@ -1017,6 +1018,12 @@ def import_func(lib, opts, args):
     metavar='ID',
     help=u'restrict matching to a specific metadata backend ID'
 )
+import_cmd.parser.add_option(
+    u'--set-field', dest='set_fields', action='callback',
```
sampsyo: This is a little bit of bikeshedding, but what do you think about the slightly simpler option name `--set`? In beets, the only thing you can really "set" is fields, so having "field" in the name feels a teensy bit redundant.

beets/ui/__init__.py
```diff
@@ -1060,6 +1060,32 @@ def parse_subcommand(self, args):
         suboptions, subargs = subcommand.parse_args(args)
         return subcommand, suboptions, subargs
 
+    @staticmethod
+    def _store_dict(option, opt_str, value, parser):
```
sampsyo: This is a nice way to parse a series of `foo=bar` options! Cool!

But I don't think this needs to be a static method of the `SubcommandsOptionParser` class, which is intended for a pretty specific purpose. Can we make it a plain, top-level function?

docs/reference/cli.rst
```diff
@@ -138,6 +138,17 @@ Optional command flags:
   searching for other candidates by using the ``--search-id SEARCH_ID`` option.
   Multiple IDs can be specified by simply repeating the option several times.
 
+* You can supply ``--set-field`` options with ``field=value`` pairs to assign to
+  those fields the specified values on import, in addition to such field/value
+  pairs defined in the ``importer.set_fields`` dictionary in the configuration
+  file. Make sure to use an option per field/value pair, like so:
+  ::
```
sampsyo: In ReST, if you just use :: at the end of your paragraph, you'll get a literal : there *and* mark the following indented block as code.

sampsyo (changes requested):
Thanks for your patience while I found the time to do a full review! This looks fantastic already; thank you for the careful and complete implementation. :sparkles: I've left a few suggestions scattered throughout the code.

beets/importer.py
bartkl: I actually used to have it structured that way, but it doesn't work. Without the call to `task.add` first, there's no `album` attribute defined on the task: `AttributeError: 'ImportTask' object has no attribute 'album'`.

Please let me know it you have another idea, because I do agree with you it would be better.

beets/ui/commands.py
bartkl: Agreed, and there's the `metavar` value which makes it sufficiently explicit. I'll change this.

beets/ui/__init__.py
bartkl: Had my doubts on where to place this. I'll follow your advice :).

docs/reference/cli.rst
bartkl: Thanks for the tip, I improved it.

beets/importer.py
sampsyo: Aha; got it! That makes sense. Let's leave it where it is—perhaps with a comment to clarify why it's there?—for this version at least.

bartkl:
Thanks for your feedback and kind words :).
I've corrected all the changes you've requested except for the ones involving removing the `store()` calls and moving the `set_fields` logic above the `task.add` call for reasons explained above. I'll commit the changes done so far, and I'll hear from you if it needs any extra work.

beets/importer.py
sampsyo: Scratch my previous comment in light of the ordering constraints, but I still recommend "un-indenting" this line so the album get stored only once to update all of its fields. (And the same below for items.)

beets/importer.py
bartkl: Added!

beets/importer.py
bartkl: Ah of course, way better indeed :), will do.

beets/importer.py
bartkl: I'll add a note.

bartkl:
I hope I'm not too much of a burden, but I don't get what goes wrong with the `py27-flake8` and python 3 tests. Could you help me out there?

And the merge conflict is probably not something I can or should resolve, right?

Thanks!

sampsyo:
Sure. Here are the results of the flake8 run (which you can get by typing `tox -e py27-flake8` or just installing flake8 and running `flake8 beets beetsplug test` to check everything):

```
beets/ui/__init__.py:774:1: E302 expected 2 blank lines, found 1
test/test_importer.py:561:24: E261 at least two spaces before inline comment
test/test_importer.py:722:25: E261 at least two spaces before inline comment
```

Those are some fairly banal style complaints: just about whitespace. In particular, you'll want to do this:

```py
print()  # Foo.
```

instead of:

```py
print() # Foo.
```

Tiny detail, I know!

I'm not seeing any Python 3 test failures on my end—is there a specific message you're looking at?

About merge conflicts: it's actually fairly straightforward for you to resolve this on your end. You can just type `git merge master` to bring everything up to date. You'll need to manually resolve the conflicts, but I think that will just be the changelog (sorry about that! Creating a new release meant monkeying with the changelog a bit).

beets/importer.py
bartkl: Thanks for the tip: done.

bartkl:
It seems to be fine :).

sampsyo:
Awesome!! Thank you for your hard work on this; I look forward to using this feature myself. Merging now…

Issue #2595, opened 2017-06-15
metasync does not escape file names fully when querying amarok


aranc23:
### Problem

The metasync plugin does not fully escape file names when querying amarok via dbus, 
in particular it does not escape double quotes which means any files with " in them causes a dbus error.

Running this command in verbose (`-vv`) mode:
```sh
[accx@a-lnx005:~/src/beets] (metasync-fixes)$ ./beet -vv metasync stuck out here
user configuration: /home/accx/.config/beets/config.yaml
data directory: /home/accx/.config/beets
plugin paths: 
Sending event: pluginload
library database: /home/accx/beets.db
library directory: /data/music
Sending event: library_opened
Traceback (most recent call last):
  File "./beet", line 23, in <module>
    beets.ui.main()
  File "/home/accx/src/beets/beets/ui/__init__.py", line 1256, in main
    _raw_main(args)
  File "/home/accx/src/beets/beets/ui/__init__.py", line 1243, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "/home/accx/src/beets/beetsplug/metasync/__init__.py", line 136, in func
    meta_source.sync_from_source(item)
  File "/home/accx/src/beets/beetsplug/metasync/amarok.py", line 74, in sync_from_source
    results = self.collection.Query(self.queryXML % escape(basename(path)))
  File "/home/accx/.local/lib/python2.7/site-packages/dbus/proxies.py", line 70, in __call__
    return self._proxy_method(*args, **keywords)
  File "/home/accx/.local/lib/python2.7/site-packages/dbus/proxies.py", line 145, in __call__
    **keywords)
  File "/home/accx/.local/lib/python2.7/site-packages/dbus/connection.py", line 651, in call_blocking
    message, timeout)
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.InvalidArgs: Invalid XML: <query version="1.0">                     <filters>                         <and><include field="filename" value="24 - Beck - Untitled ("Stuck Out Here...").ogg" /></and>                     </filters>                 </query>

```
If I change the template for the XML query to dbus to use single quotes, then queries will fail when the file name contains single quotes.

The call to escape(basename(path))) on line 76 doesn't escape " or ' and it should probably do so?

In a sane world my files wouldn't contain double or single quotes but some do...

### Setup

* OS: Fedora (not relevant)
* beets version 1.4.4
* Python version 2.7.13

My configuration (output of `beet config`) is:

```yaml
convert:
[...]

PR #2596, opened 2017-06-15
do not quote the value in xml, use quoteattr to add quotes as needed

aranc23:
This fixes issue #2595 by using quoteattr() instead of escape()


sampsyo:
Awesome; thank you. This looks like just the right fix.

PR #2598, opened 2017-06-15
Relative date queries (continuation of #2418)

sampsyo:
This adds relative dates to the query syntax for date-type fields. It's @euri10's work; I'm just opening the pull request to succeed #2418. 👻

And **this time**, I'm using a branch in the main repository.


beets/dbcore/query.py
```diff
@@ -593,10 +607,27 @@ def find_date_and_format(string):
...
+            quantity = match_dq.group('quantity')
+            timespan = match_dq.group('timespan')
+
+            # Add or subtract the given amount of time from the current
+            # date.
+            multiplier = -1 if sign == '-' else 1
+            days = cls.relative[timespan]
+            date = datetime.now() + \
+                timedelta(days=int(quantity) * days) * multiplier
+            string = date.strftime(cls.date_formats[5][0])
```
sampsyo: @euri10: Turning things back into a string seems a little odd. Can we just `return cls(date, something)` here?

beets/dbcore/query.py
```diff
@@ -562,6 +562,9 @@ class Period(object):
         ('%Y-%m-%dT%H:%M', '%Y-%m-%d %H:%M'),  # minute
         ('%Y-%m-%dT%H:%M:%S', '%Y-%m-%d %H:%M:%S')  # second
     )
+    relative = {'y': 365, 'm': 30, 'w': 7, 'd': 1}
```
sampsyo: A better name might be `relative_units`.

beets/dbcore/query.py
```diff
@@ -606,6 +637,8 @@ def open_right_endpoint(self):
         """
         precision = self.precision
         date = self.date
+        if 'relative' == self.precision:
+            return date
```
sampsyo: Can `self.precision` ever be "relative" in this version? Is this possibly out of date?

docs/reference/query.rst
```diff
@@ -164,6 +164,29 @@ Dates are written separated by hyphens, like ``year-month-day``, but the month
 and day are optional. If you leave out the day, for example, you will get
 matches for the whole month.
 
+You can also use relative dates to the current time.
+A relative date begins with an ``@``.
+It looks like ``@-3w``, ``@2m`` or ``@-4d`` which means the date 3 weeks ago,
```
sampsyo: Looks like we need to remove the @ here.

beets/dbcore/query.py
euri10: done, at first I didn't know if the ordinal was used anywhere, so turning things back into a string was a way to use the find_date_and_format function that already exists

beets/dbcore/query.py
euri10: indeed, no use

docs/reference/query.rst
euri10: yep, corrected

sampsyo:
Awesome; thank you! I've made a few minor refinements to the docs and merged. :sparkles: This is a small but awesome feature to have; I, for one, will be using it all the time. Nice work, and thank you for your patience! :rocket:

euri10:
Thanks to you for the amazing cli 😎.

I lacked time to look really into it, but the smartplaylist will need some
adaptation to handle correctly playlists containing relative dates.

The last time I tested some with songs added last week for instance (-1w),
the refresh wasn't handled appropriately.

Can't promise I will, but that would be cool to tackle that issue.

sampsyo:
Cool! Let me know if you need a hand looking into that issue.

Issue #2181, opened 2016-08-30
importer.py; asis and tagged comps


BevanBroun:
On an album where all tracks have albumartist as "Various Artist" but tracks have different artists, the importer doesnt recognise this as a compilation when "-A" (asis) is used.  

The behaviour makes sense when reading the code. I modified importer.py line 633 removing "i.albumartist or" to get the result I need.
Bevan@gonzilla>diff importer.py importer.comp_asis.py
633c633
## <                 [i.albumartist or i.artist for i in self.items]

> ```
>             [i.artist for i in self.items]
> ```

Bevan@gonzilla>beet -vv
user configuration: /cygdrive/d/dropbox/unix_home/.config/beets/config.yaml
data directory: /cygdrive/d/dropbox/unix_home/.config/beets
plugin paths:
Sending event: pluginload
lyrics: Disabling google source: no API key configured.
inline: adding item field multidisc
library database: /cygdrive/d/dropbox/unix_home/data/musiclibrary.blb
library directory: /d/music_beets
Sending event: library_opened
Usage:
  beet COMMAND [ARGS...]
  beet help COMMAND

Options:
  --format-item=FORMAT_ITEM
                        print with custom format
  --format-album=FORMAT_ALBUM
                        print with custom format
  -l LIBRARY, --library=LIBRARY
                        library database file to use
  -d DIRECTORY, --directory=DIRECTORY
                        destination music directory
  -v, --verbose         log more details (use twice for even more)
  -c CONFIG, --config=CONFIG
                        path to configuration file
  -h, --help            show this help message and exit

Commands:
  clearart          remove images from file metadata
  config            show or edit the user configuration
  embedart          embed image files into file metadata
  extractart        extract an image from file metadata
  fetchart          download album art
  fields            show fields available for queries and format strings
  help (?)          give detailed help on a specific sub-command
  import (imp, im)  import new music
  list (ls)         query the library
  lyrics            fetch song lyrics
  modify (mod)      change metadata fields
  move (mv)         move or copy items
  remove (rm)       remove matching items from the library
  stats             show statistics about the library or a query
  super             do something super
  update (upd, up)  update the library
  version           output version information
  write             write tag information to files
Sending event: cli_exit

Led to this problem:
Compilation album with correct tags for albumartist and artists.
### Setup

multiple OS
multiple pythons 2.6.8 and a 2.7. 
beets version 1.36 to 1.40

config:
directory: /d/music_beets
library: ~/data/musiclibrary.blb
[...]

Issue #1099, opened 2014-11-18
Compilation flag inference in as-is imports is inconsistent with update behavior


busybit:
When I import a compilation with "Use as is", and after that run "beet update", the comp: flag is set to false and the album is split by artist into many artist folders.

lfam:
> I'd be interested to hear from any other users about what they expect from "as-is" imports and the update command.

I just experienced this issue and it absolutely not what I expected to happen. As-is imports should stay as-are.

DirkReiners:
This just bit me... The current behavior is really non-intuitive and potentially very destructive (I now have hundreds of directories with one song each...).

My workaround was to manually set the compilation flag (using mid3v2 --TCMP 1 */*mp3) and update again. But I was lucky, the disassembled compilations ended up in one exclusive folder, making that possible. A preventative could be to run mid3v2 --TCMP 1 */Compilations/*/*mp3  */Compilations/*/*/*mp3 before running update, but that's pretty brutal, and you need to know about it before you do the update...

Is anybody working on this?

sampsyo:
No, I don't think anyone's working on it—any interest in picking it up?

So I can understand a little more, what do you think the right fix would be? We could, as two examples, just stop doing comp flag inference at all on as-is imports, or we could change the `update` command instead to do the same thing as `import -A`.

m42i:
As suggested in #2181  I'd refrain from guessing the album artist on importing.

PR #2602, opened 2017-06-18
Remove album artist guessing from importer (#1099)

m42i:
Also mentioned in #2181

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -631,23 +631,7 @@ class ImportTask(BaseImportTask):
         """
         changes = {}
 
-        if self.choice_flag == action.ASIS:
-            # Taking metadata "as-is". Guess whether this album is VA.
-            plur_albumartist, freq = util.plurality(
-                [i.albumartist or i.artist for i in self.items]
-            )
-            if freq == len(self.items) or \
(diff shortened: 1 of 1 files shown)


sampsyo:
Hi! Thanks for the contribution. As the test failures indicate, this is a tricky issue. I understand that your music has carefully-curated *missing* album artists, but in other cases where the tags are messier, our artist-guessing heuristic is pretty critical. Without values in an `albumartist` field, music collections can be pretty frustrating to use.

That said, we added these heuristics back in a time when we were transitioning to an album-level structure for such fields. It seemed particularly critical then, so it may be time to reevaluate that decision, especially because it belies the "as-is" nature of as-is imports.

In any case, we need to satisfy a few goals:

- Simplifying the behavior so it's as unsurprising as possible.
- Avoiding disruption of existing workflows.
- Matching the importer and `update` behavior.

This might mean adding an option for the new or old behavior, and asking some users what they expect to predict potential problems for others. Are you interested in chasing down possible solutions here?

jtpavlock:
closing as no response from the original author. 

@m42i if you decide you'd like to continue to work on this, please feel free to re-open the PR

PR #2626, opened 2017-07-12
Case sensitivity for german letters Ä, Ö and Ü

pkess:
Hi there,

i just discoverd a problem with unicode strings: The german letters ä, ö and ü are case-sensitive in queries. I was not able to fix the problem at the moment but i added a test case for this. Any ideas where to fix this?

### DIFF
--- a/test/test_ui.py
+++ b/test/test_ui.py
@@ -71,6 +71,19 @@ class ListTest(unittest.TestCase):
             out = out.decode(stdout.encoding)
         self.assertTrue(u'na\xefve' in out)
 
+    def test_list_unicode_query_upper_lower(self):
+        """Test if german letters Ä, Ö and Ü will discard case"""
+        self.item.title = u'Unrockbar'
+        self.item.artist = u'Die Ärzte'
+        self.item.store()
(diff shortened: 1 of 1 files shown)


sampsyo:
Hi! Thanks for pointing this out. The root cause is that we rely on the [SQLite LIKE operator](https://sqlite.org/lang_expr.html#like), which is case in-sensitive by default but only knows about ASCII characters for that purpose. The documentation linked there refers to an ICO extension to help resolve this, but it's not clear how widespread this is. This could use a little more digging, if you're up for it!

pprkut:
An alternative might be to do a lower case comparison, something like `WHERE lower(attr) LIKE lower("*foo*")` or some such. But no idea what the performance penalty might be for that.

sampsyo:
I believe that would have the same problem: the lower function https://sqlite.org/lang_corefunc.html#lower is ASCII-only by default; we can get a Unicode-aware version by enabling the ICO extension.

jtpavlock:
This could be another thing [supersqlite](https://github.com/plasticityai/supersqlite) helps with

PR #2628, opened 2017-07-14
print lyrics as HTML

anarcat:
Here's another crazy idea: how about printing lyrics in a usable
format? Here, `lyrics -p` on the whole library just dumps a
concatenation of all lyrics, which is not very useful.

I was thinking of improving the output a little, maybe by adding HTML
tags or some form of markup. this pull request does exactly that,
although rather crudely: the resulting output is probably not valid
HTML (it's just snippets) and embeds the lyrics in `<pre>` tags (as
opposed to trying to format them correctly).

the output is, of course, not an ebook in any way, although it could
be used to construct an ebook or other things.

comments? other ideas? suggestions?

i figured it was better to implement this directly in beets instead of
hacking at the sqlite database on the side...

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -726,7 +726,7 @@ class LyricsPlugin(plugins.BeetsPlugin):
             if self.rst is not None:
                 slug = re.sub(r'\W+', '-', unidecode(self.artist).lower())
                 path = os.path.join(directory, 'artists', slug + u'.rst')
-                with open(path, 'w') as output:
+                with open(path, 'wb') as output:
                     output.write(self.rst.encode('utf-8'))
                 self.rst = None
                 if item is None:
@@ -761,7 +761,7 @@ class LyricsPlugin(plugins.BeetsPlugin):
                 raise
         indexfile = os.path.join(directory, 'index.rst')
         if not os.path.exists(indexfile):
-            with open(indexfile, 'w') as output:
+            with open(indexfile, 'wb') as output:
                 output.write(u'''Lyrics
 ======
 
@@ -778,7 +778,7 @@ Artist index:
 ''')
         conffile = os.path.join(directory, 'conf.py')
         if not os.path.exists(conffile):
-            with open(conffile, 'w') as output:
+            with open(conffile, 'wb') as output:
                 output.write(u'''# -*- coding: utf-8 -*-
 master_doc = 'index'
 project = u'Lyrics'
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -598,7 +598,7 @@ class LyricsPlugin(plugins.BeetsPlugin):
                 "76V-uFL5jks5dNvcGCdarqFjDhP9c",
             'fallback': None,
             'force': False,
-            'skip': False,
+            'local': False,
             'sources': self.SOURCES,
         })
         self.config['bing_client_secret'].redact = True
@@ -673,9 +673,9 @@ class LyricsPlugin(plugins.BeetsPlugin):
             help=u'always re-download lyrics',
         )
         cmd.parser.add_option(
-            u'-s', u'--skip', dest='skip_fetched',
+            u'-l', u'--local', dest='local_only',
             action='store_true', default=False,
-            help=u'skip already fetched lyrics',
+            help=u'do not fetch missing lyrics',
         )
 
         def func(lib, opts, args):
@@ -686,7 +686,7 @@ class LyricsPlugin(plugins.BeetsPlugin):
             album = False
             output = sys.stdout
             for item in lib.items(ui.decargs(args)):
-                if not opts.skip_fetched and not self.config['skip']:
+                if not opts.local_only and not self.config['local']:
                     self.fetch_item_lyrics(
                         lib, item, write,
                         opts.force_refetch or self.config['force'],
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -797,11 +797,11 @@ class LyricsPlugin(plugins.BeetsPlugin):
                 raise
         indexfile = os.path.join(directory, 'index.rst')
         if not os.path.exists(indexfile):
-            with open(indexfile, 'wb') as output:
+            with open(indexfile, 'w') as output:
                 output.write(REST_INDEX_TEMPLATE)
         conffile = os.path.join(directory, 'conf.py')
         if not os.path.exists(conffile):
-            with open(conffile, 'wb') as output:
+            with open(conffile, 'w') as output:
                 output.write(REST_CONF_TEMPLATE)
 
     def imported(self, session, task):
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -760,7 +760,8 @@ class LyricsPlugin(plugins.BeetsPlugin):
         """
         if item is None or self.artist != item.artist:
             if self.rest is not None:
-                slug = re.sub(r'\W+', '-', unidecode(self.artist).lower())
+                slug = re.sub(r'\W+', '-',
+                              unidecode(self.artist.strip()).lower())
                 path = os.path.join(directory, 'artists', slug + u'.rst')
                 with open(path, 'wb') as output:
                     output.write(self.rest.encode('utf-8'))
@@ -769,14 +770,15 @@ class LyricsPlugin(plugins.BeetsPlugin):
                     return
             self.artist = item.artist
             self.rest = u"%s\n%s\n\n.. contents::\n   :local:\n\n" \
-                        % (self.artist, u'=' * len(self.artist))
+                        % (self.artist.strip(),
+                           u'=' * len(self.artist.strip()))
         if self.album != item.album:
             tmpalbum = self.album = item.album
             if self.album == '':
                 tmpalbum = u'Unknown album'
-            self.rest += u"%s\n%s\n\n" % (tmpalbum,
-                                          u'-' * len(tmpalbum))
-        title_str = u":index:`%s`" % item.title
+            self.rest += u"%s\n%s\n\n" % (tmpalbum.strip(),
+                                          u'-' * len(tmpalbum.strip()))
+        title_str = u":index:`%s`" % item.title.strip()
         block = u'| ' + item.lyrics.replace(u'\n', u'\n| ')
         self.rest += u"%s\n%s\n\n%s\n" % (title_str,
                                           u'~' * len(title_str),
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -595,6 +595,7 @@ class LyricsPlugin(plugins.BeetsPlugin):
                 "76V-uFL5jks5dNvcGCdarqFjDhP9c",
             'fallback': None,
             'force': False,
+            'skip': False,
             'sources': self.SOURCES,
         })
         self.config['bing_client_secret'].redact = True
@@ -663,18 +664,33 @@ class LyricsPlugin(plugins.BeetsPlugin):
             action='store_true', default=False,
             help=u'always re-download lyrics',
         )
+        cmd.parser.add_option(
+            u'-s', u'--skip', dest='skip_fetched',
+            action='store_true', default=False,
+            help=u'skip already fetched lyrics',
+        )
 
         def func(lib, opts, args):
             # The "write to files" option corresponds to the
             # import_write config value.
             write = ui.should_write()
+            artist = ''
+            album = ''
             for item in lib.items(ui.decargs(args)):
-                self.fetch_item_lyrics(
-                    lib, item, write,
-                    opts.force_refetch or self.config['force'],
-                )
+                if not opts.skip_fetched and not self.config['skip']:
+                    self.fetch_item_lyrics(
+                        lib, item, write,
+                        opts.force_refetch or self.config['force'],
+                    )
                 if opts.printlyr and item.lyrics:
-                    ui.print_(item.lyrics)
+                    if artist != item.artist:
+                        artist = item.artist
+                        ui.print_('<h1>' + artist + '</h1>')
+                    if album != item.album:
+                        album = item.album
+                        ui.print_('<h2>' + artist + '</h2>')
+                    ui.print_('<h3>' + item.title + '</h3>')
+                    ui.print_('<pre>' + item.lyrics + '</pre>')
 
         cmd.func = func
         return [cmd]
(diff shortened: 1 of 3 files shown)


anarcat:
this is a little crazy, i must admit, but it works. i can import the resulting output in Sigil and create a table of contents, then save that as an ePUB and it's read by my e-reader.

beets is still going through my library to get those lyrics, but a first dump of the database with the letters A and B done gave me a 840 pages ePUB.

so yeah, a little crazy... but i think it's worth exploring further.

anarcat:
i'm now at the optimization stage. with a full library, generating the ePUB takes a considerable amount of time, which is a blocker for me right now. i am considering making the patch actually write all the necessary files in a given output directory, which would allow me to write per-artist RST files, which would optimize the workflow a lot.

does that sound reasonable?

SusannaMaria:
Nice enhancement. My collection is not very cultivated in terms of lyrics but I tried your code, which looks nice. Some handwork has to be done, maybe in later versions we can add some automation using sphinx in beet directly. Would like to see if we can give the filename as option and doing the rst generation without any fetching of lyrics.

Tested on Windows10 x64 with Python3

grafik

anarcat:
> Some handwork has to be done, maybe in later versions we can add some automation using sphinx in beet directly.

That may be a little over the top, in my opinion. It would add additional dependencies into beets and i'm not sure the maintainers  here would like that. Furthermore, who knows what format the end user will want? HTML? PDF? ePUB? ePUB3? Let's just let each tool do its job. We can tell the user explicitly how to do theirs. :)

> Would like to see if we can give the filename as option [...]

Ask and you shall receive. ;) The latest commits pushed actually write RST files directly, in a given output directory, instead of outputting a long single RST document. that was necessary because the size of the file (and the resulting ePUB) was getting unwidely and hard to deal with by all parts of the toolchain. sphinx now takes much faster to parse the resulting documents (with progress information) and ePUB readers are much more responsive, as they need to open only smaller files now.

> [...] and doing the rst generation without any fetching of lyrics.

you should already be able to do that with the newly introduced `--skip / -s` option.

the only question that remains for me is whether the `conf.py` and `index.rst` files belong to the documentation or the program: should we automatically generate them or tell the user how to create the files? I tend towards the former, but right now the documentation is where those files are. i would think it better to create the files once, if they don't already exist, and tell the user how to use them interactively.

docs/plugins/lyrics.rst
```diff
@@ -84,10 +84,57 @@ lyrics will be added to the beets database and, if ``import.write`` is on,
 embedded into files' metadata.
 
 The ``-p`` option to the ``lyrics`` command makes it print lyrics out to the
-console so you can view the fetched (or previously-stored) lyrics.
+console so you can view the fetched (or previously-stored) lyrics. The
```
sampsyo: This discussion might deserve its own, separate section heading in the docs.

beetsplug/lyrics.py
```diff
@@ -659,22 +663,69 @@ def commands(self):
...
+        cmd.parser.add_option(
             u'-f', u'--force', dest='force_refetch',
             action='store_true', default=False,
             help=u'always re-download lyrics',
         )
+        cmd.parser.add_option(
+            u'-s', u'--skip', dest='skip_fetched',
+            action='store_true', default=False,
+            help=u'skip already fetched lyrics',
+        )
```
sampsyo: For some reason, the option name “skip” seems a little counterintuitive to me. (And “skip already fetched lyrics” is the opposite of what it does, right? It skips lyrics that have *not* been fetched, if I understand correctly.) Maybe another name like “local” or “print-only” or something would be clearer?

beetsplug/lyrics.py
```diff
@@ -659,22 +663,69 @@ def commands(self):
...
-                    ui.print_(item.lyrics)
+                if not opts.skip_fetched and not self.config['skip']:
+                    self.fetch_item_lyrics(
+                        lib, item, write,
+                        opts.force_refetch or self.config['force'],
+                    )
+                if item.lyrics:
+                    if opts.printlyr:
+                        ui.print_(item.lyrics)
+                    if opts.writerst:
```
sampsyo: Could we please consider moving the ReST stuff to its own function? Having it separated from the top-level workflow will make things easier to read.

beetsplug/lyrics.py
anarcat: that's true, the description is completely backwards. i was trying to make the opposite option of `--force`, so `--skip` made sense.  not sure what would be a better name... `-l` and `--local` seem better, but don't fit well with `--force`...

beetsplug/lyrics.py
anarcat: i'd be happy to. would it be a different `func` hook? or a different plugin? or just a different function? i'm not sure which level you're asking for here.

anarcat:
@sampsyo i'd love to hear your comments on whether the `conf.py` and `index.rst` files belong to the documentation or to the code. they could be auto-generated as well, which would streamline the process neatly.

docs/plugins/lyrics.rst
anarcat: makes sense.

beetsplug/lyrics.py
sampsyo: Yep, just a separate function (or method, really) will do nicely.

sampsyo:
Great! Here are my brief thoughts from the overall review comment:

> I don’t have a strong feeling about automatically generating the Sphinx config—producing an index file seems very easy to justify (it’s “just ReST”), and while the config file is more tool-specific, it could be convenient to have too.

anarcat:
ah yes, sorry i missed that. will add conf.py and index.rst generation as well then.

beetsplug/lyrics.py
anarcat: i can see moving it to a different section, but the problem is that we'd need to iterate over all the items another time - we wouldn't write the lyrics as soon as we download them, whereas we do for the `--print` function.

we need a complete loop access because of the way we open that rst file: one per artist means we need to close the file outside of the loop when we're done.

unless i missed something obvious here of course. been staring at that loop too long already. ;)

beetsplug/lyrics.py
anarcat: done: renamed to `--local`.

beetsplug/lyrics.py
anarcat: moved it to a different function. required moving more state into the class, but i think it works out okay. we still write as we go, i hope that's what you had in mind too.

docs/plugins/lyrics.rst
anarcat: done too.

anarcat:
@sampsyo alright, i believe all your concerns have been addressed.

@SusannaMaria the command now also generates the necessary `conf.py` and `index.rst` base structures so you should be able to just run the sphinx-build commands, which are shown when the rst generation completes, to generate your format of choice.

i think this is pretty much ready to sail now.

beetsplug/lyrics.py
```diff
@@ -695,40 +700,47 @@ def func(lib, opts, args):
...
+        """Write the item to an RST file
+
+        this will keep state (in the `rst` variable) in order to avoid
+        writing continuously to the same files
+        """
+        if item is None or self.artist != item.artist:
+            if self.rst is not None:
+                slug = re.sub(r'\W+', '-', unidecode(self.artist).lower())
+                path = os.path.join(directory, 'artists', slug + u'.rst')
+                with open(path, 'w') as output:
```
SusannaMaria: I got an error "TypeError: write() argument must be str, not bytes"
this fixed it: 
with open(path, 'w**b**') as output:

don't know if others face the same problem, win10x64

anarcat: ugh. typical python crap. let me guess: you're using Python 3?

isn't the `encode('utf-8')` idiom okay anymore?

anarcat: pushed a commit to add the `b`, not sure what's idiomatic anymore...

SusannaMaria: You are fast! Guess its okay to use utf-8.

beetsplug/lyrics.py
```diff
@@ -659,26 +671,133 @@ def commands(self):
             help=u'print lyrics to console',
         )
         cmd.parser.add_option(
+            u'-r', u'--write-rst', dest='writerst',
+            action='store', default='.', metavar='directory',
+            help=u'write lyrics to given directory as RST files',
```
sampsyo: I know this is totally bikeshedding, but reStructuredText is usually abbreviated reST or ReST rather than RST.

beetsplug/lyrics.py
```diff
@@ -659,26 +671,133 @@ def commands(self):
...
+            os.makedirs(os.path.join(directory, 'artists'))
+        except OSError as e:
+            if e.errno == errno.EEXIST:
+                pass
+            else:
+                raise
+        indexfile = os.path.join(directory, 'index.rst')
+        if not os.path.exists(indexfile):
+            with open(indexfile, 'wb') as output:
+                output.write(u'''Lyrics
```
sampsyo: For legibility, would you mind moving these long string literals to constants at the top of the file? I find that usually makes it easier to follow the code around the strings.

beetsplug/lyrics.py
```diff
@@ -659,26 +671,133 @@ def commands(self):
...
+.. toctree::
+   :maxdepth: 1
+   :glob:
+
+   artists/*
+''')
+        conffile = os.path.join(directory, 'conf.py')
+        if not os.path.exists(conffile):
+            with open(conffile, 'wb') as output:
+                output.write(u'''# -*- coding: utf-8 -*-
```
sampsyo: This (and the similar stanza above) actually breaks on Python 3 (and would break on Python 2 if this string were non-ASCII). When a file is opened in binary mode (with `b` in the flags to `open`), then you can only write raw bytes to the file, not Unicode strings like `u'''...'''`. Here's the error, FWIW:

```
  File "beet", line 23, in <module>
    beets.ui.main()
  File "/Users/asampson/Documents/code/beets/beets/ui/__init__.py", line 1256, in main
    _raw_main(args)
  File "/Users/asampson/Documents/code/beets/beets/ui/__init__.py", line 1243, in _raw_main
  File "/Users/asampson/Documents/code/beets/beetsplug/lyrics.py", line 694, in func
    self.writerst_indexes(opts.writerst)
  File "/Users/asampson/Documents/code/beets/beetsplug/lyrics.py", line 779, in writerst_indexes
    ''')
TypeError: a bytes-like object is required, not 'str'
```

The right resolution is probably to do an `.encode('utf8')` on the string before writing it.

beetsplug/lyrics.py
```diff
@@ -659,26 +671,133 @@ def commands(self):
...
+
+        this will keep state (in the `rst` variable) in order to avoid
+        writing continuously to the same files
+        """
+        if item is None or self.artist != item.artist:
+            if self.rst is not None:
+                slug = re.sub(r'\W+', '-', unidecode(self.artist).lower())
+                path = os.path.join(directory, 'artists', slug + u'.rst')
+                with open(path, 'wb') as output:
+                    output.write(self.rst.encode('utf-8'))
```
sampsyo: (Regarding the comment below: this works fine because you're now writing encoded bytes to a binary file handle.)

beetsplug/lyrics.py
```diff
@@ -659,26 +671,133 @@ def commands(self):
...
+                           u'&& make -C _build/latex all-pdf')
+                          % opts.writerst)
         cmd.func = func
         return [cmd]
 
+    def writerst(self, directory, item):
+        """Write the item to an RST file
+
+        this will keep state (in the `rst` variable) in order to avoid
+        writing continuously to the same files
```
sampsyo: Another incredibly tiny thing, but can I convince you to use full sentences (capital letters and periods) in docstrings and comments? That will help match the rest of the beets code.

beetsplug/lyrics.py
anarcat: fine with changing this

beetsplug/lyrics.py
anarcat: fine with this too, feel free to correct, otherwise i may take a day or two to catchup here too.

beetsplug/lyrics.py
anarcat: that makes sense.

beetsplug/lyrics.py
anarcat: ... or *not* use the `b` flag in this specific case?

anarcat: i removed the `b` flag and this works both in 2 and 3 now.

not sure if you want to remove the `encode()` call in the rest of the patch... but it would be tricky because we *do* need to write unicode there...

beetsplug/lyrics.py
anarcat: done.

beetsplug/lyrics.py
anarcat: right.

beetsplug/lyrics.py
anarcat: done, i believe.

beetsplug/lyrics.py
anarcat: changed. i changed *all* occurences of `rst` (except the `.rst` file extension), i hope this is what you meant. it has the nice side-effect of making the `self.rst` variable renamed to the appropriate `self.rest`.

anarcat:
how's that? all good? :)

beetsplug/lyrics.py
sampsyo: Yes; perfect!

beetsplug/lyrics.py
sampsyo: Yep, since those files are ASCII, this will work great on both Python 2 and 3. :tada:

sampsyo:
Awesome! Nice work on this. :sparkles: I've merged everything up. Woohoo!

anarcat:
whoohoo! awesome! :)

anarcat:
i noticed a few problems remaining with the code:

1. there is a duplication of code between the "slugification code" that i introduced to generate filenames and the already existing "slugify" function
2. a newline is missing after the lyrics block, which generates a warning for every song (!!) when the resulting rst is built
3, the last commit, which aimed at fixing some `:index:` directive that were not rendered correctly, doesn't seem to be working here.

i'll work on a new PR to fix those issues, sorry for the trouble.

i must say i'm not sure what to do about slug thing... the `slugify` function provided by beets does more than just create a slug: it removes stuff in parens, for example. it also uses NKFD normalization which doesn't actually work so well for asian characters, if i understand correctly. i would argue for adopting a simpler approach, but that involves changing that piece of code which introduces all sorts of other issues so i'm hesitant in doing so...

sampsyo:
Thanks for catching those! About `slugify` in particular: it looks like that is probably tuned specifically for getting Google results, which might result in slightly different decisions from a similar function for filename generation.

PR #2633, opened 2017-07-17
improvements and fixes to the lyrics RST rendering

anarcat:
a few issues were discovered in https://github.com/beetbox/beets/pull/2628#issuecomment-315792892:

 1. a newline is missing after the lyrics block, which generates a warning for every song (!!) when the resulting rst is built
 2. the last commit, which aimed at fixing some :index: directive that were not rendered correctly, doesn't seem to be working here, and may introduce overwrite between differently named artists that end up with the same "slug"

this introduces a new local function named `slug` which overlaps with the already existing `slugify` function. the code was taken in part from [this stack overflow discussion](https://stackoverflow.com/a/8366771/1174784) which explains how the `NKFD` approach used by the `slugify` function is actually incorrect. For example, `NKFD` will do this:

```
Mørdag -> mrdag
Æther -> ther
```

while unidecode will do that:

```
Mørdag -> mordag
Æther -> aether
```

we already depend on unidecode, so this is not an additional dependency. i'm hesitant in replacing the `slugify` function, however: it behaves very differently than my implementation and seems more complicated. it strips stuff in parens, for example, but also uses underscores instead of dashes to replace non-word characters.

i also wonder if we want to have unit tests for the lyrics stuff - is it mandatory for new things to get into beets?

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -236,6 +236,10 @@ def slug(text):
     2. shift everything to lowercase
     3. strip whitespace
     4. replace other non-word characters with dashes
+
+    This somewhat duplicates the :func:`Google.slugify` function but
+    slugify is not as generic as this one, which can be reused
+    elsewhere.
     """
     return re.sub(r'\W+', '-', unidecode(text).lower().strip())
 
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -774,11 +774,10 @@ class LyricsPlugin(plugins.BeetsPlugin):
                         % (self.artist,
                            u'=' * len(self.artist))
         if self.album != item.album:
-            tmpalbum = self.album = item.album
+            tmpalbum = self.album = item.album.strip()
             if self.album == '':
                 tmpalbum = u'Unknown album'
-            self.rest += u"%s\n%s\n\n" % (tmpalbum.strip(),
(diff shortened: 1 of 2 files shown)


sampsyo:
This looks great; thanks for catching all these things. Could I ask you the favor of moving the `slug` function to the top level, instead being a nested function, because it doesn't depend on any local variables? That's what we usually do with little "utility" functions like this.

Tests are always nice! We don't have a strict requirement, because that can sometimes slow things down too much, but they're definitely encouraged. Tests for the ReST output might be sort of annoyingly long, because entire output files might be required, but they certainly can't hurt.

anarcat:
fine with moving up the `slug` function, but what about the overlap with the existing `slugify` function? we don't care? :)

sampsyo:
:smiley: Fortunately, `slugify` is a method on the `Google` backend class, so at least there won't be a namespace clash.

anarcat:
done. i dodge the question of unit tests then, since they do seem a little of a burden right now. i scratched my itch, and just hope things will keep on working... it would be nice to have some tests for `slug`, at least, however...

i moved up the `slug` function so i guess this is done here.

anarcat:
added tests for `slug` at least, and found a small improvement to be done there as well...

sampsyo:
Looks perfect, and I like the slug output tweaks too. Nice work. :sparkles:

PR #2634, opened 2017-07-17
lyrics: detect MusixMatch blocking

### LINKED ISSUE #2632: lyrics: Detect when you get blocked by Musixmatch
### Problem

My IP seems to have been blocked by musixmatch. Some of the lyrics downloaded say:

```
We are human after all

Human After All
DAFT PUNK

We detected that your IP is blocked.
It looks like you’re a bot or more likely your network may have been blacklisted for some reason.
Please confirm you are a human by checking the box below.
```

It seems the problem is resolved now because I can't reproduce with a regular web browser. It would be nice, however, to detect such scenarios and not record the problematic lyrics. I am not sure how to do this - I don't know if a different status code gets sent, but maybe we can do a simple regex-based pattern-matching technique?

Thanks!

anarcat:
we just look for the bad string in the HTML. this has the downside
that we may consider songs that have those exact lyrics (you never
know, really) may trigger this warning as well and we would fail to
fetch those songs.

we also fail if lyrics contain another magic string that seems to come
up when you do fill in the CAPTCHA after being blocked.

for an example of error messages in songs, I suggest listening to the
excellent "Retry? Abort? Ignore?" from the band called "Martyr".

this is to fix #2632, obviously.

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -301,9 +301,19 @@ class MusiXmatch(SymbolsReplaced):
         html = self.fetch_url(url)
         if not html:
             return
+        if "We detected that your IP is blocked" in html:
+            self._log.warning(u'we are blocked at MusixMatch: url %s failed'
+                              % url)
+            return
         html_part = html.split('<p class="mxm-lyrics__content')[-1]
         lyrics = extract_text_between(html_part, '>', '</p>')
-        return lyrics.strip(',"').replace('\\n', '\n')
+        lyrics = lyrics.strip(',"').replace('\\n', '\n')
+        # another odd case: sometimes only that string remains, for
+        # missing songs. this seems to happen after being blocked
+        # above, when filling in the CAPTCHA.
+        if "Instant lyrics for all your music." in lyrics:
+            return
+        return lyrics
 
 
 class Genius(Backend):
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,6 +14,9 @@ Fixes:
 
 * :doc:`/plugins/replaygain`: Fix a regression in the previous release related
   to the new R128 tags. :bug:`2615` :bug:`2623`
+* :doc:`/plugins/lyrics`: The MusixMatch backend now detect and warns
+  the user when blocked on the server. Thanks to
+  :user:`anarcat`. :bug:`2634` :bug:`2632`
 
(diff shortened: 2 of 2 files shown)


anarcat:
i'm not sure about this. as I mention in the description, some bands actually use error messages in their lyrics now (!) so this might create false positives. especially for the latter check introduced...

anarcat:
i also wonder if the warning should not be a little more violent: should we just completely abort?

anarcat:
btw - the CAPTCHA solving doesn't work unless you carry over that cookie it sends you when you solve it. :( it also looks like the block is not permanent. i suspect there may be some rate-limiting going on here... for example:

```
lyrics: we are blocked at MusixMatch: url https://www.musixmatch.com/lyrics/Keny-Arkana/Entre-Les-Lignes-%3A-Une-Goutte-De-Plus failed
lyrics: lyrics not found: Keny Arkana - Entre ciment et belle étoile - Entre les lignes : Une goutte de plus
lyrics: lyrics already present: Keny Arkana - Entre ciment et belle étoile - Prière
lyrics: lyrics already present: Louis Armstrong - Louis Armstrong Plays W.C. Handy - St. Louis Blues
lyrics: lyrics already present: Louis Armstrong - Louis Armstrong Plays W.C. Handy - Yellow Dog Blues
lyrics: fetched lyrics: Louis Armstrong - Louis Armstrong Plays W.C. Handy - Loveless Love
```

now maybe the lyrics for that second entry were fetched elsewhere... but it seems to me an indicator of rate-limiting...

sampsyo:
Looks good. Could you please add a changelog entry? Then this should be good to go.

> i also wonder if the warning should not be a little more violent: should we just completely abort?

I'm good with this just being a warning—a fail-stop error seems a little aggressive, since other backends could work just fine.

anarcat:
changelog: done.

sampsyo:
Perfect! :sparkles:

sampsyo:
Also, I sent you an invitation to join the @beetbox org. Thanks for all your help recently, and welcome aboard! As a collaborator you can push directly, of course, but don’t hesitate to use a PR if you ever want a code review.

anarcat:
oh wow, thanks! i'll certainly send PRs for a while still, thank you so much for your confidence again, it's greatly appreciated!

PR #2643, opened 2017-07-28
reflink support

### LINKED ISSUE #2642: Copy on Write/reflink for supported filesystems
For the uninitiated, [this article explains what reflinks are](https://lwn.net/Articles/331808/).

### Problem

There are use cases where the user wants the data which they import using `beets im` to be untouched, unmoved. Currently, the only option is `copy`, which obviously makes a copy of the original to-be-imported file. This (obviously) effectively duplicates the total amount of data for the end-user, while actually beets (not considering transcoding options) only changes a few kB at most.

It would thus be beneficial to deduplicate said data, preferable *without* copying the file at all. Currently, `copy` followed by a (very CPU and disk intensive) deduplication action saves same space, but using reflinks on supported filesystems would increase both performance of beets (achieving the same speed as a hard link), while keeping the benefits from copy.

### Design decisions

I propose a tristate configuration option `reflink`, such as `cp --reflink` provides:

- no: do not use reflink (*default*)
- always: force reflink, failing when unable to use reflink
- auto: try using reflink, but fallback onto `copy` when fails

---

*I am currently working on this, expect a WIP PR soon.* Feedback before is obviously welcome :-)

rubdos:
~~Please refrain from merging this WIP PR :-)~~

Seems good to go now. C&c can begin! :D

Closes #2642

### DIFF
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -7,6 +7,7 @@ import:
     move: no
     link: no
     hardlink: no
+    reflink: no
     delete: no
     resume: ask
     incremental: no
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -221,7 +221,8 @@ class ImportSession(object):
             iconfig['incremental'] = False
 
         if iconfig['reflink']:
-            iconfig['reflink'] = iconfig['reflink'].as_choice(['auto', True, False])
+            iconfig['reflink'] = iconfig['reflink'] \
+                .as_choice(['auto', True, False])
 
         # Copy, move, reflink, link, and hardlink are mutually exclusive.
         if iconfig['move']:
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -220,19 +220,30 @@ class ImportSession(object):
             iconfig['resume'] = False
             iconfig['incremental'] = False
 
-        # Copy, move, link, and hardlink are mutually exclusive.
+        if iconfig['reflink']:
+            iconfig['reflink'] = iconfig['reflink'].as_choice(['auto', True, False])
+
+        # Copy, move, reflink, link, and hardlink are mutually exclusive.
         if iconfig['move']:
             iconfig['copy'] = False
             iconfig['link'] = False
             iconfig['hardlink'] = False
+            iconfig['reflink'] = False
         elif iconfig['link']:
             iconfig['copy'] = False
             iconfig['move'] = False
             iconfig['hardlink'] = False
+            iconfig['reflink'] = False
         elif iconfig['hardlink']:
             iconfig['copy'] = False
             iconfig['move'] = False
             iconfig['link'] = False
+            iconfig['reflink'] = False
+        elif iconfig['reflink']:
+            iconfig['copy'] = False
+            iconfig['move'] = False
+            iconfig['link'] = False
+            iconfig['hardlink'] = False
 
         # Only delete when copying.
         if not iconfig['copy']:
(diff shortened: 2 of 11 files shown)


rubdos:
That's step one: https://pypi.python.org/pypi/reflink/0.1.0

rubdos:
Good news:

```
% for i in {01..11}; do ~/go/bin/sharedextents [..]${i}* ~/Music/[..]${i}*; done
16220160 / 16351232 bytes (99.20%)
28049408 / 28573696 bytes (98.17%)
35676160 / 35807232 bytes (99.63%)
21233664 / 21364736 bytes (99.39%)
28274688 / 28405760 bytes (99.54%)
27779072 / 27910144 bytes (99.53%)
50233344 / 50364416 bytes (99.74%)
32276480 / 32407552 bytes (99.60%)
30986240 / 31117312 bytes (99.58%)
40529920 / 40660992 bytes (99.68%)
29716480 / 29847552 bytes (99.56%)
```

Seems like we save around 99.5% of disk space for those use cases.

rubdos:
Almost there. I'm adding AppVeyor to my own repo for Windows, and I'll fix the flake8 stuff asap! :D

rubdos:
[Wrote a blog post about the Python impl](https://www.rubdos.be/2017/07/30/about-reflinks.html)

rubdos:
Windows "works" (aka. `raise`s the not supported exception), Darwin should too, but I can't test that on GitLab without having a Mac server apparently.

Let's see whether Windows works here too :)

sampsyo:
Extremely cool. This new module came together very nicely. The Darwin backend does indeed raise the appropriate exception.

Since your nifty new module requires native code (and therefore a C compiler to install), can we make it an optional dependency for beets? This shouldn't be too hard—we can just put the `import` under a condition so that it's only needed when reflinking is actually enabled.

rubdos:
> Extremely cool. This new module came together very nicely. The Darwin backend does indeed raise the appropriate exception.

Good. Would you mind looking at the `darwin` branch over there? I *might* have implemented it for Darwin, but I can't test :')

> Since your nifty new module requires native code (and therefore a C compiler to install), can we make it an optional dependency for beets? This shouldn't be too hard—we can just put the import under a condition so that it's only needed when reflinking is actually enabled.

Of course. Just one question: 

>so that it's only needed when reflinking is actually enabled

Do you mean:

- when `config['reflink'] in ('auto', True)`, or
- something at install time?

sampsyo:
Totally; I’ll give this a try when I have a moment with a desktop machine.

And yes, I did mean the former, so putting the import inside the reflink function seems right. :sparkles:

sampsyo:
I tried out the module on macOS with APFS and it works too! Woohoo! :tada:

I did, however, notice that the module got a little confused when I tried to clone a file that doesn't exist:

```
>>> reflink.reflink('junk', 'junk2')
  File "/Users/asampson/Documents/code/pyreflink/reflink/reflink.py", line 33, in reflink
    backend.clone(oldpath, newpath)
  File "/Users/asampson/Documents/code/pyreflink/reflink/backend.py", line 39, in clone
    raise Exception("BUG: errno %s not implemented" % errno.errorcode[lib.errno])
Exception: BUG: errno ENOENT not implemented
```

It seems like `backend.py` should probably check for ENOENT and raise a standard Python `FileNotFoundError`.

rubdos:
> And yes, I did mean the former, so putting the import inside the reflink function seems right. ✨

Already done. :) 

> It seems like backend.py should probably check for ENOENT and raise a standard Python FileNotFoundError.

Working on that.

rubdos:
Hi! Would you find some time to review the macOS version again (and the rest in this PR for that matter)? Thanks :-)

rubdos:
I'm rebasing this thing on master; would have been cool you'd warned me about #2687 though (or waited until after this pr); this was quite a pain to rebase.

sampsyo:
I'm sorry for dropping the ball on this! It's in great shape, though.

I've created a new branch here and a new pull request, #2720, and added you as a @beetbox collaborator so we can work together. This is going to be awesome to have…

Issue #2609, opened 2017-06-22
zero: Support album art as an option in keep_fields


plittlefield:
This is a follow up to the issue - https://github.com/beetbox/beets/issues/2589 - and I have been asked to report it here as a 'bug' or 'request'.

Problem = The `zero` plugin is not able to keep, or is interferring with the `embedart` plugin.

Here is my beet `config` - https://pastebin.com/hQrNm0uC

This is the id3 tag for a song after beet import with `zero` plugin - https://pastebin.com/hrhvpzQ1

This is the command output - http://paste.ubuntu.com/24883079

* OS: Linux - Ubuntu 16.04.2 LTS
* Python version: 2.7.12
* beets version: 1.4.4

Thanks!

sampsyo:
Thanks! Yes, we should add special support for a key like `art` in the `keep_fields` line. In the mean time, the workaround is to use `fields`, i.e., the "negative" option—and just list all the fields *except* those you want to keep. That should preserve the album art.

nathdwek:
I don't really understand, it seems to me that in the last pastebin provided in the other thread, with `art` listed in `keep_fields`, there is indeed no art deletion going on.

Am I understanding this issue wrong?

sampsyo:
It’s true that the plugin doesn’t log anything about deleting the `art` field, but (unless I’m misunderstanding too!) we still end up removing the album art according to the second paste: https://pastebin.com/hrhvpzQ1

My best guess is that this comes from our having two different names in MediaFile for image data: `art` and `images`. But that really is just a guess...

PaulFinch:
working for me with "images" keyword

plittlefield:
I shall try this and report back.

plittlefield:
`zero:`
`keep_fields: title artist album year track genre images`

Boom, problem solved! Thank you so much for your work on this.

:-)

sampsyo:
Awesome! Is there any chance one of you (@plittlefield or @PaulFinch) might be able to add a note about this to the documentation for the `zero` plugin?

plittlefield:
Yes, would love to. Do I put something in this issue page or somewhere else?

sampsyo:
You can open a pull request by going to the `zero.rst` page here on GitHub and hitting the "edit" button.

plittlefield:
[...]

PR #2653, opened 2017-08-11
Update zero.rst

plittlefield:
See #2609.


sampsyo:
Thank you! That was really helpful. Merged with some text refinement.

PR #2654, opened 2017-08-13
new generic TrackInfo and AlbumInfo

### LINKED ISSUE #1547: Extra fields from MusicBrainz (missing fields)

gabrielhdt:
TrackInfo and AlbumInfo inheritate from class ItemInfo, which behaves
like a dict and is backward compatible regarding attributes management.
apply_metadata function has been updated consequently, favouring dict
like item access more than attributes. Apply_metadata isn't general yet.
#1547

### DIFF
--- a/beets/autotag/__init__.py
+++ b/beets/autotag/__init__.py
@@ -13,8 +13,7 @@
 # The above copyright notice and this permission notice shall be
 # included in all copies or substantial portions of the Software.
 
-"""Facilities for automatically determining files' correct metadata.
-"""
+"""Facilities for automatically determining files' correct metadata."""
 
 from __future__ import division, absolute_import, print_function
 
@@ -30,11 +29,21 @@ from .match import Recommendation  # noqa
 log = logging.getLogger('beets')
 
 
+INFO2ITEM = {
+    'track_id': 'mb_trackid',
+    'artist_id': 'mb_artistid',
+}
+"""Translates info attributes to item attributes"""
+
+
 # Additional utilities for the main interface.
 
 def apply_item_metadata(item, track_info):
     """Set an item's metadata from its matched TrackInfo object.
     """
+    for fld in track_info.keys():
+        if fld in INFO2ITEM.keys():
+            setattr(item, INFO2ITEM[fld], track_info[fld])
     item.artist = track_info.artist
     item.artist_sort = track_info.artist_sort
     item.artist_credit = track_info.artist_credit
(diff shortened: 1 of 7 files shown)


beets/autotag/hooks.py
```diff
@@ -67,63 +130,44 @@ class AlbumInfo(object):
...
+        else:
+            super(AlbumInfo, self).__init__(**kwargs)
+
+    def __hash__(self):
+        """Always a good idea to have a hash"""
+        return self['album_id']
+
+    def __reduce__(self):
+        """Used by :py:func:`copy.deepcopy`"""
+        return self['album_id']
```
dosoe: That's what I was looking for! How did you know this is the missing part for deepcopy?

beets/autotag/__init__.py
```diff
@@ -61,22 +61,70 @@ def apply_item_metadata(item, track_info):
...
+        'disctitle': 'disctitle',
+    }
+    # Maps the item attribute to the corresponding key to be tried
+    # first on TrackInfo and then on AlbumInfo
+    itemattr_to_torainfokey = {
+        'artist': 'artist',
+        'artist_sort': 'artist_sort',
+        'artist_credit': 'artist_credit',
+        'mb_artistid': 'artist_id',
+    }
```
dosoe: Why not just make a loop over all tags? Then when one adds a tag he doesn't need to add something here.

beets/autotag/hooks.py
gabrielhdt: I looked into copy.deepcopy ... you can see it on the pickle doc too:
[https://docs.python.org/2/library/pickle.html#object.__reduce__](url)

beets/autotag/__init__.py
gabrielhdt: What you suggest must be eventually done. For the moment I adapted the current process to behave in a 'generic-friendly' way. I will come back to `apply_metadata` after having edited the `track_info` function of `mb.py` to make it capable of copying tags from recording object to TrackInfo.

dosoe:
Thanks! Then I will just kill my own try (that failed on `deepcopy`) here on https://github.com/beetbox/beets/pull/2650

gabrielhdt:
It seems that perfect flexibility (i.e. dumping directly musicbrainz data to tags) isn't possible, due to the recursive architecture of musicbrainz relations (relations that have relations, which can be seen in musicbrainzngs dict as they are recursive) which can't be applied to the media tags.

Therefore, I chose to base the tagging process on the parsing of a mapping. The mapping reproduces the architecture of the musicbrainzngs dict and maps the musicbrainzngs dict key to the tag key (the ItemInfo attribute/key). Flexibility is improved as the mapping should eventually be stored in a json or yaml file, which will have to be updated to follow musicbrainz database schemes updates, one shouldn't need to edit python code to add an attribute, only update model.

For the moment, there's a bug in `mutate` of `db.py`, the variable `subvals` isn't properly formatted.

beets/autotag/__init__.py
```diff
@@ -61,22 +61,70 @@ def apply_item_metadata(item, track_info):
...
+        'disc': 'medium',
+        'mb_trackid': 'track_id',
+        'track_alt': 'track_alt',
+        'lyricist': 'lyricist',
+        'composer': 'composer',
+        'composer_sort': 'composer_sort',
+        'arranger': 'arranger',
+        'media': 'media',
+        'disctitle': 'disctitle',
+    }
```
dosoe: Hi! Thanks for all the work, it looks awesome. A question: will it be necessary for each new tag to add a corresponding line to the mapping?

gabrielhdt: Yes it will. But that shouldn't be a problem, since the only task will eventually be to add a line to a json/yaml file.

dosoe:
I'm wondering (I didn't try it out) how it works if there are several albumartists: what is currently done is that the names of all the artists are appenden in a string, does something similar happen here or does he just take the last one?

jtpavlock:
@gabrielhdt is there any interest in continuing this?

gabrielhdt:
Hi @jtpavlock, I don't really know, is it still relevant, and aren't there too many merge conflicts? On the part of the remaining code, if I remember correctly, we have to tinker with the database. I don't think I'll continue on this (even though I still use the software, so I'm still sort of interested). But I think that we should agree first on what we want: it has emerged from discussions with @dosoe that the needs differ in function of the people (which is normal), while I was more targeting for a system that mirrors musicbrainz' data on the files I have, @dosoe wanted to be able to add musics that should not be referenced into musicbrainz.

dosoe:
Hi! I implemented flexible tags in https://github.com/beetbox/beets/pull/3568 , now we just need to get the actual tags. I have a PR in progress that does this for performers (but could easily be extended to places and dates) here: https://github.com/beetbox/beets/pull/3589 . That one is still work in progress (and will need quite some changes based on comments).

gabrielhdt:
Ah, that's nice, thank you @dosoe! I guess this pull request is really out of date then.

dosoe:
Now that flexible tags have been implemented, it's much easier to do.

jtpavlock:
Awesome, in that case I'll close this in favor of #3568

Thanks for your work on this @gabrielhdt! Sorry we never got to see it all the way through.

Issue #2467, opened 2017-03-03
importfeeds: Configuration error due to handling of "dir" option on Python 3


Armael:
### Problem

Running `beet import /my/album` terminates with the following message: `configuration error: importfeeds.dir: must be a filename, not bytes`.

### Setup

* OS: Archlinux
* Python version: 3.6.0
* beets version: 1.4.4 (latest git)
* Turning off plugins made problem go away: yes

The output of `beet config` is:

```yaml
importfeeds:
    dir: !!binary |
        L3Bvb2wvbXVzaXF1ZQ==
    relative_to:
    formats: m3u
    m3u_name: imported.m3u
    absolute_path: no
directory: /pool/musique
library: /pool/musiclibrary.blb

plugins: fromfilename embedart fetchart lastgenre thumbnails importfeeds
embedart:
    ifempty: yes
    maxwidth: 0
    auto: yes
    compare_threshold: 0
    remove_art_file: no
fetchart:
    enforce_ratio: 5%
    auto: yes
    minwidth: 0
    maxwidth: 0
    cautious: no
    cover_names:
    - cover
    - front
    - art
    - album
    - folder
    sources:
    - filesystem
    - coverart
    - itunes
    - amazon
    - albumart
    google_key: REDACTED
    google_engine: 001442825323518660753:hrh5ch1gjzm
    fanarttv_key: REDACTED
    store_source: no
lastgenre:
    count: 5
    force: no
    whitelist: yes
    min_weight: 10
    fallback:
    canonical: no
    source: album
    auto: yes
    separator: ', '
    prefer_specific: no

import:
    log: /pool/beets_import.log
thumbnails:
    auto: yes
    force: no
    dolphin: no
```
The `~/.config/beets/config.yaml` itself is:

```yaml
directory: /pool/musique
library: /pool/musiclibrary.blb

plugins: fromfilename embedart fetchart lastgenre thumbnails importfeeds

embedart:
    ifempty: yes

fetchart:
    enforce_ratio: 5%

lastgenre:
    count: 5
    force: no

import:
#    incremental: yes                                                                                                                                                                                               
    log: /pool/beets_import.log

importfeeds:
    formats: m3u
```

jrobeson:
we should probably drop the Gio backend, but I was hoping to wait til we dropped py2 and could rely on the newer pathlib completely

sampsyo:
You're right that it would be nice to drop Gio, @jrobeson (although that applies to the other ticket, #2466).

[...]

PR #2658, opened 2017-08-18
Fix importfeeds plugin on Python 3.

djl:
Submitted as a pull request to run the tests because I can't get `tox` to work correctly.


sampsyo:
Thank you for putting this together! This is to address #2467, right?

If so, this looks much cleaner than it did before, and it should fix the problem. 👍 ✨

PR #2662, opened 2017-08-20
Fix kodiupdate URL and add better error reporting

natezb:
The most important aspect of this change is to correct the Kodi URL. The current URL, which has a trailing slash, produces a 404 error on my system (Kodi v17.1 on ArchLinux).

The existing error handling/logging missed this, so I added checks of the HTTP status code and the JSON response data for more detailed information.

### DIFF
--- a/beetsplug/kodiupdate.py
+++ b/beetsplug/kodiupdate.py
@@ -33,7 +33,7 @@ from beets.plugins import BeetsPlugin
 def update_kodi(host, port, user, password):
     """Sends request to the Kodi api to start a library refresh.
     """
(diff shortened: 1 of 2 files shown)


sampsyo:
Thank you for noticing and fixing this! The URL change seems harmless enough, even for other setups where the old URL happened to work. And the new error handling is a very good idea.

Would you mind adding a changelog entry? Then this will be ready to merge.

sampsyo:
Fantastic! :sparkles: Thank you again.

natezb:
Man, that was quick! Thanks

Issue #2664, opened 2017-08-21
`mtime` reset when assigning unchanged values to `Item` fields


wisp3rwind:
### Problem

In [`library.py`](https://github.com/beetbox/beets/blame/master/beets/library.py#L551)
``` python
class Item(LibModel):
    # ...
    def __setitem__(self, key, value):
        """Set the item's value for a standard field or a flexattr.
        """
        # Encode unicode paths and read buffers.
        if key == 'path':
            if isinstance(value, six.text_type):
                value = bytestring_path(value)
            elif isinstance(value, BLOB_TYPE):
                value = bytes(value)

        if key in MediaFile.fields():
            self.mtime = 0  # Reset mtime on dirty.

        super(Item, self).__setitem__(key, value)
```
the `mtime` will be reset (and consequently `Item._dirty` be set) irrespective of whether any field actually
changed. Most of the time, this is probably not a problem since the item would be dirtied anyway, and even if not, no real database corruption will occur. There is a particular situation where many unnecessary `mtime` updates and metadata writes to the mediafiles can be triggered, though:

Into the following `beet edit` session, the query selected a full album, but I only changed one song title:
```
beet -v edit txarango album:'som riu'                                                                                                                                                                      
...
Sending event: library_opened
edit: invoking editor command: ['/usr/bin/nvim', '/tmp/tmp9l8hedej.yaml']
Txarango - Som riu - Esperança
  title: Esperança -> Esperança [Not a remix]
continue [E]diting, Apply, Cancel? a
edit: saving changes to Txarango - Som riu - Esperança [Not a remix]
Sending event: write
Sending event: after_write
moving /.../Txarango/Som riu/01 - Esperança.mp3 to synchronize path
the: "Txarango" -> "Txarango"
Sending event: before_item_moved
Sending event: item_moved
Sending event: database_change
Sending event: database_change
Sending event: database_change
edit: saving changes to Txarango - Som riu - Músic de carrer
Sending event: write
Sending event: after_write
moving /.../Txarango/Som riu/02 - Músic de carrer.mp3 to synchronize path
the: "Txarango" -> "Txarango"
Sending event: before_item_moved
Sending event: item_moved
Sending event: database_change
Sending event: database_change
Sending event: database_change
edit: saving changes to Txarango - Som riu - La vuelta al mundo
Sending event: write
Sending event: after_write
moving /.../Txarango/Som riu/03 - La vuelta al mundo.mp3 to synchronize path
[...]

PR #2667, opened 2017-08-22
Mtime fixes

wordofglass:
As per #2664, the `Item.mtime` is reset for any assignment to `MediaFile`-backed fields, even if the `Item`s fields and the file's tags do not go out of sync (i.e. if the previous value is reassigned). This happens frequently in the `edit` plugin. While the effects are not devastating, it's certainly not ideal when using the plugin with coarse queries and only changing few items in the editor.

In fact, `test.test_edit` already had a method to check that these superfluous tag writes didn't happen. It didn't really check this, though, since the `mtime` for test fixtures would always be 0 to begin with. An "`mtime` reset" would therefore not `_dirty` the `Item`, and the plugin would in turn not write the tags out: In a real-world setting with non-zero `mtime`, these writes did happen.

Some additional details can be found in the commit messages. Thoughts?

**NOTE:** Forgot the changelog, will add it later.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -242,8 +242,9 @@ class Model(object):
         else:
             raise KeyError(key)
 
-    def __setitem__(self, key, value):
-        """Assign the value for a field.
+    def _setitem(self, key, value):
+        """Assign the value for a field, return whether new and old value
+        differ.
         """
         # Choose where to place the value.
         if key in self._fields:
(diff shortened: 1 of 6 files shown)


sampsyo:
Nice work! These test fixes look good as well. Fighting through the harnesses there wasn’t easy—you deserve some kind of a badge. 🏅 

This looks ready to merge to me.

wordofglass:
Ok. Would you usually write a changelog entry for this? As far as I can see, there's no user-visible change (well, file access times will be updated less frequently which might affect some rsync, but this is pretty minor IMO).

sampsyo:
In general, I do usually tend to write a changelog for even these esoteric things. I'm not sure anyone actually reads them, but it makes me happy to have a complete summary of what happened in a given release. :smiley:

Maybe something like this would work?

- Fixed a problem where "no-op" modifications would reset files' mtimes, resulting in unnecessary writes. This most prominently affected the :doc:\`/plugins/edit\` when saving the text file without making changes to some music.

wordofglass:
I couldn't have worded that better, I'm going to blatantly copy that to the changelog 😆

sampsyo:
Fantastic; thank you!

wordofglass:
Ok, resolving merge conflicts on Github is apparently not the best idea. Sorry for the extra commit...

PR #2668, opened 2017-08-24
Edit plugin logic: Regression from previous PR; incorrect diffs

wordofglass:
Contains two fixes:

- Turns out I introduced a regression in #2659 due to incorrectly assuming objects to have a valid `id` when `object._db` is set. Crashed beets when re-tagging albums (since the yaml file would contain `id: ''` which can obviously not be coerced to an integer)
- The plugin shows diffs against different initial states in first and subsequent edits, see the commit message for more detail.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -218,9 +218,12 @@ class Model(object):
             raise ValueError(u'{0} has no id'.format(type(self).__name__))
 
     def copy(self):
-        """Create a semi-deep copy of the item. In particular, the _db object
-        is not duplicated. A simple copy.deepcopy wouldn't work due to the
-        sqlite objects in _db.
+        """Create a copy of the model object.
+
+        The field values and other state is duplicated, but the new copy
+        remains associated with the same database as the old object.
+        (A simple `copy.deepcopy` will not work because it would try to
+        duplicate the SQLite connection.)
         """
         new = self.__class__()
         new._db = self._db
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -217,6 +217,18 @@ class Model(object):
         if need_id and not self.id:
             raise ValueError(u'{0} has no id'.format(type(self).__name__))
 
+    def copy(self):
+        """Create a semi-deep copy of the item. In particular, the _db object
+        is not duplicated. A simple copy.deepcopy wouldn't work due to the
+        sqlite objects in _db.
+        """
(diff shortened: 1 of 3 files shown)


wordofglass:
Ok, this still doesn't work because deepcopy doesn't like `sqlite3.Connection` objects. Do we have a method to safely copy `LibModel`s in some place?

wordofglass:
Should be fine now, I ended up implementing `Model.copy()`.

beetsplug/edit.py
```diff
@@ -302,9 +301,14 @@ def edit_objects(self, objs, fields):
                 elif choice == u'c':  # Cancel.
                     return False
                 elif choice == u'e':  # Keep editing.
-                    # Reset the temporary changes to the objects.
+                    # Reset the temporary changes to the objects. I we have a
+                    # deepcopy from above, use that, else reload from the
```
sampsyo: Perhaps just call it a "copy" here, since we're not using `deepcopy` anymore.

beets/dbcore/db.py
```diff
@@ -217,6 +217,18 @@ def _check_db(self, need_id=True):
         if need_id and not self.id:
             raise ValueError(u'{0} has no id'.format(type(self).__name__))
 
+    def copy(self):
+        """Create a semi-deep copy of the item. In particular, the _db object
+        is not duplicated. A simple copy.deepcopy wouldn't work due to the
+        sqlite objects in _db.
+        """
```
sampsyo: I know this is extremely low-level, but sometimes I like structuring docstrings so there is a quick summary at the top followed by more detail after a blank line. This can help when navigating a big codebase for the first time. For example, I might suggest:

```
"""Create a copy of the model object.

The field values and other state is duplicated, but the new copy
remains associated with the same database as the old object.
(A simple `copy.deepcopy` will not work because it would try to
duplicate the SQLite connection.)
"""
```

beetsplug/edit.py
wordofglass: Right, that's a leftover from before, thanks for spotting it!

beets/dbcore/db.py
wordofglass: :+1:

wordofglass:
Well, it could persist because the plugin didn't copy any items with a database connection, but let `ui.show_model_changes` retrieve those anew from the database. That's still what happens most of the time, the only exception is when re-importing an album.

PR #2683, opened 2017-09-05
Added a missing switch to the fetchart command

euri10:
I was trying to complete my album art collection today and found it not very friendly: maybe I'm using it the wrong way, there's always that possibility!

In the current state, issuing a beet fetchart command yields a gigantic list of already found stuff I really don't care about when your goal is just to fill the gaps in your collection,

that PR may solve the "issue"

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -824,9 +824,15 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
             action='store_true', default=False,
             help=u're-download art when already present'
         )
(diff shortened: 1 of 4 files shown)


sampsyo:
This looks great! Yes, that can be annoying.

What do you think about calling this something like "quiet" (`-q`) to convey that it really only affects the logging, not the actual art-fetching functionality?

Also, would you mind writing a quick changelog entry?

sampsyo:
Perfect. Thanks again! :sparkles: :rocket:

PR #2685, opened 2017-09-09
mbcollection: Support specifiying collection ID.

djl:
For those of us who have multiple MusicBrainz collections I've added support for specifying which collection you'd like to use.

In the future I'd like to add support for multiple collections updated via queries (in the vein of the smartplaylist plugin).


sampsyo:
This is perfect! Thanks for the refactoring—I like the new method too.

A “smart playlist” feature for collections could be cool too...

Issue #2682, opened 2017-09-04
The move function operation flags are error prone


zigarrre:
## Problem

As I'm in the process of implementing #2629 I noticed something else: The current move functions are somewhat ambiguous and error prone. According to the code only one of the operations (move, copy, link, hardlink) is possible at a time. So the operation should be passed as a single parameter instead of a set of flags.

Consider the following function call that makes no sense and will lead to errors but looks fine:
```
x.move(copy=True, link=True)
```

A good API should be hard to use incorrectly.

## Proposed solution

Define an enum or constants (of either integers or strings) describing the possible operations and change all move functions to accept the operation as a single operator.

For example a enum based solution could look like the following:
```
class MoveOperations(Enum):
    MOVE = 0
    COPY = 1
    LINK = 2
    HARDLINK = 3

def move(self, operation, ...)  # change all the move, move_item and move_art functions

x.move(MoveOperations.COPY)
```

If you like the idea just let me know what variant you like most (enum or module wide constants with integer or string values) and I will refactor the code and send you a pull request.

sampsyo:
Absolutely; this is a good refactoring idea. (The backstory here is essentially the same as the importer options: this made sense when the only alternative was "copy," but we've outlived that old design.)

It probably makes sense to do this first, before all of #2629, so the importer can use the same enum.

I do think an `Enum` class is probably the right thing. One tiny nit, though: I think the naming convention would suggest the singular `MoveOperation` instead of the plural `MoveOperations`.

zigarrre:
Doing this first and using the same enum for the events was my plan. I didn't notice till now that the importer could also use this refactor on some functions, good that you mentioned that.

I noticed that multiple methods only check the `copy` flag before attempting to prune the source directory (e.g. [here](https://github.com/beetbox/beets/blob/cb2cb009594aababfeacdb7dce30a270893e9a2e/beets/library.py#L804)). Is this a remnant from when move and copy where the only options? As far as I can tell pruning doesn't make sense when linking or hardlinking so I'll refactor that to `operation == MoveOperation.MOVE`.

sampsyo:
Yes, that's a good point—I think that is a holdover from when those were the only two options. Everything else should be considered "copy-like" for the purpose of pruning.

zigarrre:
OK.

[...]

PR #2687, opened 2017-09-11
Refactored move functions for clarity according to #2682

zigarrre:
The move functions in library.py and manipule_files in importer.py where
changed to use a single parameter for the file operation instead of
multiple boolean flags.

A typo in the documentation of the Album.move and Item.move functions
confusing True and False when describing the store parameter was fixed
as well.

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -37,7 +37,7 @@ from beets import dbcore
 from beets import plugins
 from beets import util
 from beets import config
-from beets.util import pipeline, sorted_walk, ancestry
+from beets.util import pipeline, sorted_walk, ancestry, MoveOperation
 from beets.util import syspath, normpath, displayable_path
 from enum import Enum
 from beets import mediafile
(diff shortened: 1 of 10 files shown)


beets/importer.py
```diff
@@ -675,28 +675,36 @@ def align_album_level_fields(self):
...
+            if operation is not None:
                 # In copy and link modes, treat re-imports specially:
                 # move in-library files. (Out-of-library files are
                 # copied/moved as usual).
                 old_path = item.path
-                if (copy or link or hardlink) and self.replaced_items[item] \
-                   and session.lib.directory in util.ancestry(old_path):
+                if (operation != MoveOperation.MOVE
+                        and self.replaced_items[item]
+                        and session.lib.directory in util.ancestry(old_path)):
```
sampsyo: flake8 has a (pretty annoying) suggestion here: it wants the `and`s to go on the line preceding, rather than the line following. Like this:

```py
                if (operation != MoveOperation.MOVE and
                        self.replaced_items[item] and
                        session.lib.directory in util.ancestry(old_path)):
```

zigarrre: I know about that flake8 warning, I did see it when running tox. Though I decided to ignore it as it is a bug. [[1]](https://github.com/PyCQA/pycodestyle/issues/513) [[2]](https://github.com/PyCQA/pycodestyle/issues/498) [[3]](https://github.com/PyCQA/pycodestyle/pull/499) [[4]](https://gitlab.com/pycqa/flake8/issues/139) 

PEP8 was updated to allow both stiles and states breaking before (like I did) even as the preferred stile for new code. [[5]](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator)

For those reasons I suggest we ignore this bogus warning by adding the following line to the flake8 configuration in `tox.ini`:
```
ignore = W503
````

sampsyo: Got it; thanks for digging! I agree: let’s ignore this warning. Would you mind adding that ignore flag to this PR?

zigarrre:
I did some more digging and found out a few more things. The W503 bug is already fixed as it's added to the default ignore list but you have a ignore option configured in `setup.cfg` and if that's the case the default ignores aren't respected anymore. This is in my opinion pretty stupid behavior but a flake8 thing I can't do anything about. That means though that if you specify a ignore list you have to list all default ignores as well. So I updated the list in `setup.cfg` and now the tests pass without problems.

On a side note: It might be worth thinking about cleaning up the tox and flake8 config. The multiple flake8 commands with different `--min-version` seem rather useless to me (especially as only the 2.7 one is run and `min-version=2.7` is specified in `setup.cfg` as well). Though this is the first time I'm using tox and flake8 so I might just not understand something here.

sampsyo:
Indeed; that is a little dumb. Thanks for investigating, and for adding the seemingly-redundant ignore to our config. I’ve merged the PR—it’s looking great!

It’s true that we’re a little more thorough than we probably need to be with our tox config for flake8; I’m not sure if the multiple versions for flake8 have ever caught anything that one run would not. However, my general philosophy is that having more tox configs that we really need is probably OK—as long as we keep our default set of environments (and the Travis runs) trimmed to a reasonable set. But if it’s getting unnecessarily confusing, I wouldn’t object to a rethink there.

PR #2692, opened 2017-09-17
mbcollection: Add support from removing albums.

djl:
Add a new ``mbcollection.remove`` configuration option (default: None) and a new ``-r`` (``--remove``) flag which removes albums from collections that are no longer present in the beets database.

The ``-r`` flag takes precedence over the ``remove`` configuration option.


sampsyo:
:tada: Thanks again!

PR #2695, opened 2017-09-22
Fix the --key option in the `duplicates` plugin.

Sjlver:
During a refactoring, the "dest='keys'" parameter was lost. This caused
the option to not have any effect.


sampsyo:
Oops! Thanks! Would you mind adding a changelog entry about this? Then we can merge it.

Sjlver:
Done.

sampsyo:
Awesome; thank you! :sparkles:

PR #2696, opened 2017-09-22
Update GStreamer installation instructions for Mac OS X.

Sjlver:
The GStreamer base plugins are required to run bpd; if they are not installed, bpd quits with

> error: Could not create playbin

This commit also deletes an unused link from the documentation.


sampsyo:
Looks great; thank you! Since we’re already installing two packages with one command, why don’t we just do all three at once? Would you mind combining those two `brew` commands into one?

Sjlver:
Done.

sampsyo:
:tada: Thank you! (And sorry for the noise; I fat-fingered it on my phone.)

PR #2702, opened 2017-09-26
[Play plugin]Respect relative paths when using albums

j000:
Now relative_to setting in play plugin is respected when searching by albums.

Small tweak, but my mpd likes it a lot :)


nathdwek:
This seems pretty straightforward, thank you!
Could you please make sure your code respects pep8? apparently there are some trailing white spaces.

sampsyo:
Many thanks to both of you! :sparkles: :fish: I know it’s tiny, but any chance we could get a short changlog entry?

PR #2712, opened 2017-10-15
Use multiprocessing to do parallel uploads of music to google-music

orf:
This improves the speed of uploads to google music. Currently it uploads sequentially, with this it will upload in parallel (depending on the number of cores).

We batch the files together because the google-music library can batch-detect songs that have already been uploaded, so it makes sense to do sequential uploads of 25 files.

### DIFF
--- a/beetsplug/gmusic.py
+++ b/beetsplug/gmusic.py
@@ -26,18 +26,19 @@ from gmusicapi import Musicmanager, Mobileclient
 from gmusicapi.exceptions import NotLoggedIn
 import gmusicapi.clients
 import multiprocessing
-
-# Global Musicmanager instance, used with multiprocessing pool upload
-_music_manager = None
+from multiprocessing.pool import ThreadPool
 
 
 class Gmusic(BeetsPlugin):
+    # Per-process Musicmanager instance, used with multiprocessing pool upload
+    _music_manager = None
+
     def __init__(self):
         super(Gmusic, self).__init__()
 
         self.m = self.create_music_manager()
         self.config.add({
-            'processes': multiprocessing.cpu_count()
+            'processes': str(multiprocessing.cpu_count())
         })
 
     def commands(self):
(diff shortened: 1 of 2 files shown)


beetsplug/gmusic.py
```diff
@@ -59,7 +58,9 @@ def upload(self, lib, opts, args):
         items = lib.items(ui.decargs(args))
         files = [x.path.decode('utf-8') for x in items]
         ui.print_(u'Uploading your files...')
-        self.m.upload(filepaths=files)
+        pool = multiprocessing.Pool(initializer=self._pool_init)
+        chunks = [tuple(files[i:i + 25]) for i in range(0, len(files), 25)]
```
jackwilsdon: It would be nice to have the chunk size configurable by the user in their `config.yaml` file.

orf: We could, but there isn't much value in changing it to be honest.

 Configuring the number of processes in the pool night be worth it though.

jackwilsdon: Yeah, that sounds like it would be useful.

orf:
I added a `processes` config option 👍

sampsyo:
Awesome; thank you! :sparkles: This seems really useful.

I have a couple of questions about the implementation:

* Did you try using threads instead of processes? That can work either with [multiprocessing.pool.ThreadPool](https://stackoverflow.com/a/3386632/39182) or using [the futures library](https://stackoverflow.com/a/11529742/39182). The reason I ask is that I suspect the Google Music API call for uploading probably releases the GIL—i.e., it spend most of its time doing socket IO.
* Using threads might make it unnecessary to use a global for the `_music_manager`, but barring that, is it possible to make that a class variable for clarity?
* It looks like our style checker is complaining about one line that's a little too long: https://travis-ci.org/beetbox/beets/jobs/288314583

orf:
I tried threads, but I found that beets became unresponsive to `ctrl+c`. With multiprocessing it worked fine.

I can make it a class variable and fix the failing line 👍

orf:
Using a threadpool is a much better solution, I'm running into https://github.com/simon-weber/gmusicapi/issues/543 and I don't have time to fix this (I've also moved away from google-music anyway!). If anyone else wants to pick this up please do.

sampsyo:
Awesome; thanks!! That issue in the library does look troublesome. We'll look into it.

jtpavlock:
Closing since @orf no longer intends to work on this. If anyone else would like to continue the work, feel free to open a new PR.

PR #2714, opened 2017-10-19
Update config.rst

Gustafosterberg:
Added command to edit the config.yaml file on windows "beet config -e" does not work.

### DIFF
--- a/docs/reference/config.rst
+++ b/docs/reference/config.rst
@@ -13,8 +13,7 @@ config -p`` to see the path on your system):
   Support/beets/config.yaml``.
 
 You can launch your text editor to create or update your configuration by
-typing ``beet config -e``. (See the :ref:`config-cmd` command for details.) It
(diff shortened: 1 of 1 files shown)


sampsyo:
Hello! Thank you for the suggestion. I have a couple of thoughts:

* It might be better just to fix the command! What's the problem you're experiencing? Let's address it directly.
* Or, if we add this to the docs, we should give a more complete path to file to edit. `notepad config.yaml` will just create a config file in the current directory.

Gustafosterberg:
Hi Adrian

1. The "beet config -e" command gives the error: "error: Could not edit configuration: [Errno 2] No such file or directory" when used in windows. 
2. You are correct it would be better to provide a more complete path, the complete command would be "notepad AppData\Roaming\beets\config.yaml"

jtpavlock:
Closing as abandoned

PR #2717, opened 2017-10-25
Made 'beet config -e' more likely to work on windows.

joelcma:
Addressing the same issue as in #2714 I made the  "beet config -e" command try opening the config file with notepad in addition to the current method of  telling them to add an EDITOR environment variable on windows machines. Reason being I think being met by setbacks first thing when getting into using a new program can be quite discouraging.

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1637,12 +1637,23 @@ def config_edit():
     """
     path = config.user_config_path()
     editor = util.editor_command()
+
     try:
         if not os.path.isfile(path):
             open(path, 'w+').close()
         util.interactive_open([path], editor)
+
     except OSError as exc:
+        """If it's a windows machine try this first"""
         message = u"Could not edit configuration: {0}".format(exc)
+        if os.name == 'nt':
+            message += u". Detected windows machine, attempting bootleg solution."
+            if(path):
+                confpath = path
+            else:
+                confpath = os.getenv('APPDATA') + "/beets/config.yaml"
+            os.system(u"notepad " + confpath)
+
         if not editor:
             message += u". Please set the EDITOR environment variable"
         raise ui.UserError(message)
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1644,15 +1644,15 @@ def config_edit():
         util.interactive_open([path], editor)
 
         except OSError as exc:
-        message = u"Could not edit configuration: {0}".format(exc)
+            message = u"Could not edit configuration: {0}".format(exc)
 
-        # If it's a windows machine, attempt to open with notepad
-        if os.name == 'nt':
-            util.interactive_open([path], u"notepad")
+            # If it's a windows machine, attempt to open with notepad
+            if os.name == 'nt':
+                util.interactive_open([path], u"notepad")
 
-        if not editor:
-            message += u". Please set the EDITOR environment variable"
-        raise ui.UserError(message)
+            if not editor:
+                message += u". Please set the EDITOR environment variable"
+            raise ui.UserError(message)
 
 config_cmd = ui.Subcommand(u'config',
                            help=u'show or edit the user configuration')
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1643,16 +1643,12 @@ def config_edit():
             open(path, 'w+').close()
         util.interactive_open([path], editor)
 
-    except OSError as exc:
-        """If it's a windows machine try this first"""
+        except OSError as exc:
         message = u"Could not edit configuration: {0}".format(exc)
+
+        # If it's a windows machine, attempt to open with notepad
         if os.name == 'nt':
-            message += u". Detected windows machine, attempting bootleg solution."
-            if(path):
-                confpath = path
-            else:
-                confpath = os.getenv('APPDATA') + "/beets/config.yaml"
-            os.system(u"notepad " + confpath)
+            util.interactive_open([path], u"notepad")
 
         if not editor:
             message += u". Please set the EDITOR environment variable"
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1643,16 +1643,16 @@ def config_edit():
             open(path, 'w+').close()
         util.interactive_open([path], editor)
 
-        except OSError as exc:
-            message = u"Could not edit configuration: {0}".format(exc)
+    except OSError as exc:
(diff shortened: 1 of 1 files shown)


beets/ui/commands.py
```diff
@@ -1637,12 +1637,23 @@ def config_edit():
...
     path = config.user_config_path()
     editor = util.editor_command()
+
     try:
         if not os.path.isfile(path):
             open(path, 'w+').close()
         util.interactive_open([path], editor)
+
     except OSError as exc:
+        """If it's a windows machine try this first"""
```
sampsyo: Docstrings only go at the tops of functions and classes; a normal comment is more appropriate here.

beets/ui/commands.py
```diff
@@ -1637,12 +1637,23 @@ def config_edit():
...
         message = u"Could not edit configuration: {0}".format(exc)
+        if os.name == 'nt':
+            message += u". Detected windows machine, attempting bootleg solution."
+            if(path):
```
sampsyo: When might `path` be empty? I *think* it should always have a value.

beets/ui/commands.py
```diff
@@ -1637,12 +1637,23 @@ def config_edit():
...
+                confpath = path
+            else:
+                confpath = os.getenv('APPDATA') + "/beets/config.yaml"
+            os.system(u"notepad " + confpath)
```
sampsyo: Can we please use `util.interactive_open` instead of `os.system`? One problem with this version, for example, would happen when there are spaces or other special characters in the path.

sampsyo (changes requested):
Wonderful! Thank you for getting this started! I agree that making this “just work” in most cases is the right tactic.

So I can understand the issue better, do you know what OSError is being raised on your machine? Is the problem that $EDITOR is not set? Or that it points to an executable that does not exist? If either of those is the problem, let’s specifically check for those situations to dispatch to the `notepad` fallback.

beets/ui/commands.py
Joelande94: It was just a precaution. But alright, removing it then 👍

beets/ui/commands.py
Joelande94: Ah, of course 👍

Joelande94:
The error is: "[Errno 2] No such file or directory" BUT the problem seems to be that EDITOR is not set and that is the case on 99% (guesstimate obviously but you get the point) of windows user's computers. In addition to that I don't even get the error saying I don't have it set, it just won't work (I have indeed checked my system variables). Of course you can just go ahead and set it IF you figure this out but I would not assume that 1 everyone knows how to and 2 that everyone has the patience for that.

sampsyo:
Got it! This is looking good!

To be clear, what's happening here is that `editor_command` is not finding an $EDITOR value, which is totally expected, and is falling back to the result of the `open_anything` function in `beets.util`. Take a look at that function—it guesses a default command for every platform. On Windows, we're currently guessing `start`. I don't know much about Windows, so I'm unsure whether this was a good idea—do you know whether `start` is a sensible default in the first place?

If we should be preemptively using `notepad` instead of `start`, then we can bypass the initial failure trying to use that.

Joelande94:
Simply changing base_cmd to 'notepad' in open_anything does indeed fix the problem (at least for me) but will this screw up something else? I am not aware of everything open_anything needs to open but 'notepad' isn't really quite as universal of a command as 'start' (as I'm sure you realize) so I imagine this will break something. Which is why I wanted it in the config_edit. But if you want the main change to be in "open_anything" I suggest making the following changes since running "start" on the path doesn't seem to be working but we still obviously don't want to mess open_anything up.

in util/init
```
def open_anything(type=None):
    """Return the system command that dispatches execution to the correct
    program.
    """
    sys_name = platform.system()
    if sys_name == 'Darwin':
        base_cmd = 'open'
    elif sys_name == 'Windows':
        if(type == "textedit"):
            base_cmd = 'notepad'
        else:
            base_cmd = 'start'
    else:  # Assume Unix
        base_cmd = 'xdg-open'
    return base_cmd
```

add "type=None" as a parameter for open_anything and add an if-else statement that if the type is "textedit" it should run notepad instead of start.

in util/init
```
def editor_command():
    """Get a command for opening a text file.

    Use the `EDITOR` environment variable by default. If it is not
    present, fall back to `open_anything()`, the platform-specific tool
    for opening files in general.
    """
    editor = os.environ.get('EDITOR')
    if editor:
        return editor
    return open_anything("textedit")
```

Add "textedit" to the fallback open_anything call.

Does this seem alright to you?

sampsyo:
Thanks for investigating! Here's the thing I'm still wondering about this: on Windows, is it possible to configure an alternative editor for text documents? That was the original intent with using `start` in the first place: if the user has Sublime Text, for example, it would be nice if the file opened in that. (That's what `open` on macOS and `xdg-open` on Linux, etc., do.)

I guess another way of putting this question is: does `start` *ever* work? For example, does typing `start foo.txt` on the command line manually work? If so, it would be useful to find out why it's not working inside the beets invocation.

Joelande94:
I'm afraid I don't know the answer to your first question. 
Yes, `start foo.txt` does normally work but it just doesn't seem to be working with os.execlp. You can try running the command manually in IDLE or whatever on a windows machine. [Computer says no](http://s2.quickmeme.com/img/ee/ee105be757bac0f5eb5d8b4aaff2686bef0e5a79f0df1b660927c3a4d7ecfdd6.jpg). With `notepad` instead of `start` it works just fine. [Here you have a screenshot showing that os.execlp does not work with `start` on windows and that it does work with `notepad`.](https://imgur.com/a/mAcha)

sampsyo:
Mysterious! After a little bit of googling, I found [a pointer to another option](http://grokbase.com/t/gg/golang-nuts/14b24p60rs/go-nuts-problem-in-running-the-windows-start-command-with-exec-cmd): we may need to invoke `cmd /c start` to get the right functionality in a subprocess. Can you try something like this?

```py
import subprocess
subprocess.check_call(["cmd", "/c", "start", "", "some_file.txt"])
```

Joelande94:
Now THAT worked **and** it actually opened it with notepad++ (even though I don't even have it set as a path variable right now, so not sure how that happened).

sampsyo:
OK, great! So maybe the right solution here is to change `open_anything` to return a list instead of a string, and on Windows, to use `["cmd", "/c", "start", ""]` instead of just `"start"`?

PR #2720, opened 2017-10-29
Add reflink option

sampsyo:
This is a continuation of #2643, by @rubdos.

Here are a few outstanding questions:

* How shall we document the need to install the `reflink` module?
* What should happen when it's not installed? Should a friendly error message remind you to install it, or should we warn and fall back to copying?
* Do we really need the `auto` option, or should `reflink: yes` automatically fall back to copying? (At the moment, I can't think of a situation where I'd want to use `yes`, i.e., fail with an error when reflinking is impossible.)

### DIFF
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -7,6 +7,7 @@ import:
     move: no
     link: no
     hardlink: no
+    reflink: no
     delete: no
     resume: ask
     incremental: no
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -220,19 +220,31 @@ class ImportSession(object):
             iconfig['resume'] = False
             iconfig['incremental'] = False
 
-        # Copy, move, link, and hardlink are mutually exclusive.
+        if iconfig['reflink']:
+            iconfig['reflink'] = iconfig['reflink'] \
+                .as_choice(['auto', True, False])
+
+        # Copy, move, reflink, link, and hardlink are mutually exclusive.
         if iconfig['move']:
             iconfig['copy'] = False
             iconfig['link'] = False
             iconfig['hardlink'] = False
+            iconfig['reflink'] = False
         elif iconfig['link']:
             iconfig['copy'] = False
             iconfig['move'] = False
             iconfig['hardlink'] = False
+            iconfig['reflink'] = False
         elif iconfig['hardlink']:
             iconfig['copy'] = False
             iconfig['move'] = False
             iconfig['link'] = False
+            iconfig['reflink'] = False
+        elif iconfig['reflink']:
+            iconfig['copy'] = False
+            iconfig['move'] = False
+            iconfig['link'] = False
+            iconfig['hardlink'] = False
 
         # Only delete when copying.
         if not iconfig['copy']:
@@ -676,7 +688,7 @@ class ImportTask(BaseImportTask):
             item.update(changes)
 
     def manipulate_files(self, operation=None, write=False, session=None):
-        """ Copy, move, link or hardlink (depending on `operation`) the files
+        """ Copy, move, link, hardlink or reflink (depending on `operation`) the files
         as well as write metadata.
 
         `operation` should be an instance of `util.MoveOperation`.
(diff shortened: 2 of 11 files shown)


sampsyo:
Also, it looks like were getting two spurious (unrelated) Travis failures? Hmm…

rubdos:
> How shall we document the need to install the reflink module?

Just in the documentation, I'd think. No real need to put that somewhere else beyond the configuration option.

> What should happen when it's not installed? Should a friendly error message remind you to install it, or should we warn and fall back to copying?

I'd say error out if it's not there, and `reflink: yes/auto` is set (so no error with `reflink: no`, otherwise, error out).

> Do we really need the auto option, or should reflink: yes automatically fall back to copying? (At the moment, I can't think of a situation where I'd want to use yes, i.e., fail with an error when reflinking is impossible.)

Yes. The use case is as follows: some people will want their collection to be ALWAYS reflinked, and be warned thoroughly when it's not possible to link, while others want it to reflink when possible (when e.g. data is already on the same partition), and not when it's elsewhere (e.g. importing from a USB key from a friend).

rubdos:
> Also, it looks like were getting two spurious (unrelated) Travis failures? Hmm…

On first sight, doesn't look like reflink code, or is it?

sampsyo:
OK, that all sounds good! Let's add a user-visible error message when the module isn't available.

Indeed, the Travis failures are incidental. In particular, I'm wrestling with a dependency suddenly becoming incompatible with old versions of Python 2.7 and Travis simultaneously deciding to revert to an old version of Python 2.7… :cry:

rubdos:
Good, let's get this merged. I rebased it on master, I'll now:

> [...] add a user-visible error message when the module isn't available.

rubdos:
> [...] add a user-visible error message when the module isn't available.

Okay, now I actually wonder what kind of error message this should be; should I throw an exception when the module cannot be imported, then catch that somewhere and print something nice? I tried reaching out over IRC, but didn't seem to get a reaction :-)

sampsyo:
Yeah, I think so! I'm not 100% sure about how the error should be exposed internally—perhaps a new exception just for this purpose?—but eventually, we have `ui.UserError` for displaying unrecoverable errors to the user.

rubdos:
> [...] add a user-visible error message when the module isn't available.

Okay, now I actually wonder what kind of error message this should be; should I throw an exception when the module cannot be imported, then catch that somewhere and print something nice? I tried reaching out over IRC, but didn't seem to get a reaction :-)

> Yeah, I think so! I'm not 100% sure about how the error should be exposed internally—perhaps a new exception just for this purpose?—but eventually, we have ui.UserError for displaying unrecoverable errors to the user.

I suppose inherting ui.UserError would be appropriate here?

rubdos:
Hmm, `beets.util` has a `HumanReadableException` too, which one should I take? :'-)

sampsyo:
It might be useful to see the evolving discussion in #2785, but I think raising a custom exception or a FilesystemError from the utility function itself is the right thing for `util`. Then, to actually show the problem to the user—if we want to stop the program altogether—higher levels in the stack should probably catch this at an opportune moment and raise a UserError.

Of course, if we wanted the same behavior as for trying to write to a read-only filesystem, for example, then just keeping it as a FilesystemError would be the right thing.

rubdos:
So, it's probably most useful to wait for #2785 to be resolved then.

fiatflux:
Since #2785 is closed now, and that appears to have been the stumbling block on this PR, what's the status on adding reflink support?

rubdos:
It's still on my TODO list. In fact:
screen shot 2018-11-29 at 09 52 16

reflink things are the only ones still on my TODO list, and I'll be doing both pretty soon when I start building out my home lab by the end of the year. So stay tuned, this is something I really want to have myself.

fiatflux:
> this is something I really want to have myself.

Wonderful, me too! Please let me know if you find something you'd like help with. It sounds like you have a specific plan and are more familiar with how the error handling should proceed, but feel free to kick some tasks over to me that are straightforward to someone new to this codebase.

rubdos:
(rebased)

PR #2722, opened 2017-10-30
Use 2.7.13 instead of 2.7_with_system_site_packages

hugovk:
`2.7_with_system_site_packages` uses a four-year-old Python 2.7.6.

This allows the latest version of pylast which supports Python 2.7.10+ as older versions don't have good HTTPS support.

Re: https://github.com/pylast/pylast/issues/201.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
     include:
         - python: 2.7.13
           env: {TOX_ENV: py27-cov, COVERAGE: 1}
-        - python: 2.7_with_system_site_packages
+        - python: 2.7.13
           env: {TOX_ENV: py27-test}
         - python: 3.4
           env: {TOX_ENV: py34-test}


hugovk:
https://github.com/pylast/pylast/issues/201#issuecomment-340360066:

> (I wonder if they have `2.7.14` too)

They do, but it's not initially installed and adds about 30 seconds to [download and extract](https://travis-ci.org/hugovk/beets/jobs/294733659#L424) so let's not bother.

2.7.13 https://travis-ci.org/hugovk/beets/builds/294717393
2.7.14 https://travis-ci.org/hugovk/beets/builds/294733658

sampsyo:
Ack! Thank you for catching that! :sparkles: :rocket: I should have noticed those other builds were referring to a different Python.

And, just for the record, I'm 100% fine with requiring a non-ancient 2.7 for beets.

PR #2725, opened 2017-10-31
Implement album merging for duplicates

### LINKED ISSUE #112: Album merging/completion
This issue was automatically migrated from Google Code.
_Original author: zent...@gmail.com (April 29, 2012 18:26:17)_
_Original issue: https://github.com/google-code-export/beets/issues/380_

udiboy1209:
Fixes #112

Would removing the album before running the merge task make sense? This code seems to work now
but I think it gives problems when reimporting library with existing duplicate albums. The tests don't fail though.

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -358,7 +358,7 @@ class ImportSession(object):
         """
         for path in paths:
             if path not in self._merged_items \
-                and path not in self._merged_dirs:
+               and path not in self._merged_dirs:
                 return False
         return True
 
@@ -367,7 +367,7 @@ class ImportSession(object):
         """
         self._merged_items.update(paths)
         dirs = set([os.path.dirname(path) if os.path.isfile(path) else path
-                        for path in paths])
+                    for path in paths])
         self._merged_dirs.update(dirs)
 
     def is_resuming(self, toppath):
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -1365,8 +1365,13 @@ def user_query(session, task):
                                 user_query(session))
 
     resolve_duplicates(session, task)
+
     if task.should_merge_duplicates:
+        # Create a new task for tagging the current items
+        # and duplicates together
         duplicate_items = task.duplicate_items(session.lib)
+
+        # duplicates would be reimported so make them look "fresh"
         _freshen_items(duplicate_items)
         duplicate_paths = [item.path for item in duplicate_items]
 
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -1038,6 +1038,7 @@ class ArchiveImportTask(SentinelImportTask):
         self.extracted = True
         self.toppath = extract_to
 
+
 class MergedImportTask(ImportTask):
     def __init__(self, toppath, paths, items):
         super(MergedImportTask, self).__init__(toppath, paths, items)
@@ -1367,9 +1368,8 @@ def user_query(session, task):
 
             duplicate_paths = [item.path for item in duplicate_items]
 
-            merged_task = MergedImportTask(None,
-                            task.paths + duplicate_paths,
-                            task.items + duplicate_items)
+            merged_task = MergedImportTask(None, task.paths + duplicate_paths,
+                                           task.items + duplicate_items)
 
             ipl = pipeline.Pipeline([
                 iter([merged_task]),
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -188,6 +188,8 @@ class ImportSession(object):
         self.paths = paths
         self.query = query
         self._is_resuming = dict()
+        self._merged_items = set()
+        self._merged_dirs = set()
 
         # Normalize the paths.
         if self.paths:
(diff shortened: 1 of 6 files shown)


beets/importer.py
```diff
@@ -1352,7 +1358,26 @@ def emitter(task):
         ])
         return pipeline.multiple(ipl.pull())
 
-    resolve_duplicates(session, task)
+    if type(task) != MergedImportTask:
```
sampsyo: I know this seems crazy, but just to check: is this condition actually necessary? That is, if we merge a bunch of albums into one, would it be OK to check again for duplicates of the newly-tagged merged album?

beets/importer.py
```diff
@@ -1352,7 +1358,26 @@ def emitter(task):
...
+        resolve_duplicates(session, task)
+        if task.should_merge_duplicates:
+            duplicate_items = task.duplicate_items(session.lib)
+            for item in duplicate_items:
+                item.id = None
+                item.album_id = None
```
sampsyo: This looks like the same rigamarole we have to do when re-importing (which is logical). Instead of duplicating the code to do this, maybe we need a helper function for this.

beets/importer.py
```diff
@@ -1352,7 +1358,26 @@ def emitter(task):
...
+            duplicate_paths = [item.path for item in duplicate_items]
+
+            merged_task = MergedImportTask(None, task.paths + duplicate_paths,
+                                           task.items + duplicate_items)
+
+            ipl = pipeline.Pipeline([
+                iter([merged_task]),
+                lookup_candidates(session),
+                user_query(session)
+            ])
```
sampsyo: Similarly, maybe a helper function would be useful here to encapsulate the temporary pipeline as used here and in `user_query`.

beets/importer.py
udiboy1209: I did this because the initially detected duplicates weren't removed, hence would again be marked in this MergedImportTask's duplicate list and give a second prompt to the user. Choosing "Remove" at this stage would be the correct way to go, which is basically what this condition does automatically. Also it helps avoid another pointless check for duplicates which we know exist.

Is there a way to remove those albums from the library, so that they don't get detected as duplicates again? like probably setting the album's id to `None` ?

sampsyo: Aha, got it. This makes me think that there might be a slightly deeper issue to address here: we probably shouldn't detect a duplicate when the "duplicate" album actually consists of the same songs as the newly-imported album! This should probably be true for any re-import situation, not just the new functionality you've added here.

In particular consider this check from `find_duplicates`:

https://github.com/beetbox/beets/blob/9c6910d1675926d9055508f4f67683cda86f9179/beets/importer.py#L637-L639

This says that we don't detect a duplicate if the set of songs is *exactly* the same between the old and new album. Perhaps this should be changed to a *subset* check instead. Does that seem reasonable?

udiboy1209: A subset check would be correct. I will make that change because the subset case would only occur in merge scenario. We also wouldn't need a separate MergeImportTask class and check too.

udiboy1209:
The problem now with re-importing duplicate albums from the library is that multiple sessions are create for each album, but if the user decides to merge the two in the first session, the second session would run into an error because the files would probably have moved. A similar error should error should occur when the user decides to remove the duplicates in the first session. What is done in that case?

sampsyo:
Hmm, I don't quite understand… what should I do to trigger a problem? For example, is it enough just to choose to merge and then "apply changes" in the temporary sub-pipeline that's created, or is there something more exotic that I need to do?

udiboy1209:
I'll explain better. There are two albums in the library , A and B which are duplicates (imported using the Keep both option)

Now when I try re-importing these two albums, two import sessions would be created for each album A and B. In the first session, when B is marked as duplicate, if I choose to Merge, tracks of B are re-imported with A into one album.

This works well and is the intended result, but now the second import session which was working on the original album B would fail. Same would happen if I choose to Remove the duplicates of A, and files of B are removed.

wordofglass:
(Disclaimer: I have read neither this PR entirely nor looked at the relevant code in detail)

Maybe, if you could detect this situation from the task that is processed second, you could make the `user_query` stage yield a `beets.pipeline.BUBBLE` (which means to abort the task and not execute any further pipeline stage) early (before showing any prompt). That would work since stages of the same type (e.g. `user_query`) are run strictly in sequence, so when the second one is run, the decision about merging the first has been made). One remaining point of failure that would need investigation might be any access to old database items that have been changed by the first task.

sampsyo:
Ah! I see; thanks for explaining, @udiboy1209, and for adding some thoughts, @wordofglass. In summary, the problem only happens when re-imports collide with merges (which is not an exotic scenario). When we merge with an old album, we don't *also* want to import the old album separately.

In a way, this is similar to the problem we used to have with recent duplicates. If you imported album X and then another copy of album X again immediately afterward, the first copy might not have reached the database yet, so no duplicate would be detected. We used to solve this with an extra data structure on the side to keep track of "recently imported" albums, but we've since replaced that with a strategy that just adds albums to the database earlier—before the files are copied and the metadata is updated.

@wordofglass's suggestion amounts to bringing back approximately the same strategy. The `user_query` stage can keep track of in-library albums that have been merged already and emit a bubble if they're encountered later on. Sounds perfect to me!

sampsyo:
Yay! This looks perfect from here. Are there any outstanding concerns, or shall we hit the merge button?

udiboy1209:
Great! Merge is good to go for me.

sampsyo:
Awesome! Thank you again for tackling this—it's been a very long-standing issue that many people have requested, and this is a very elegant solution. I hereby award you the Pull Request of the Month award. 🥇

sean-abbott:
Awesome, thank you so much! Just in time for me to merge my library with my partner's. :-)

PR #2726, opened 2017-11-01
Update fetchart.rst

translit:
Probably an error. Minwidth does not require image editors, does it?


sampsyo:
Yes, good point! That was a mistake.

PR #2728, opened 2017-11-01
Added to docs: admonish users to back up libraries.

andrewbenedictwallace:

PR #2730, opened 2017-11-01
Mention backups in the main guide

sampsyo:
As an alternative to #2728, this mentions the need to back up your music before importing stuff. This wording fits better with the style of the rest of the guide. I'd be interested to hear feedback on how to word this.

### DIFF
--- a/docs/guides/main.rst
+++ b/docs/guides/main.rst
@@ -182,6 +182,11 @@ here, including the directory and file naming scheme. See
 Importing Your Library
 ----------------------
 
+The next step is to import your music files into the beets library database.
+Because this can involve moving files around, data loss is always a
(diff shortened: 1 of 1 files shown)


docs/guides/main.rst
```diff
@@ -182,6 +182,11 @@ here, including the directory and file naming scheme. See
 Importing Your Library
 ----------------------
 
+The next step is to import your music files into the beets library database.
+Because this can involve moving files around, data loss is always a
```
Rovanion: I'd just add a mention of the fact that beets can also modify files.

"Because this can involve moving files around as well as modifying them, data loss is always a"

sampsyo: Indeed; thank you!

PR #2731, opened 2017-11-01
Modify task recommendation and candidates

### LINKED ISSUE #2441: Import: allow edit right after applying search results
Feature request. Often the information retrieved automatically by search engines during the import procedure needs some corrections (e.g. most CD reissues have the year of release as "year", instead of the year of the original recording): I'd like to make these fixes right away, using the "edit" command, instead of waiting for the import procedure to finish, and then run `beet edit ...` with the proper query to only match the album just imported.

Currently an "Edit" command is available in import mode, but it can be used alternatively to the Search command (i.e. starting from a blank template), it would be very convenient if "Edit" would be available right after applying the results of a search, using those as starting values in the editor.

sampsyo:
Should fix #2441.

The problem in that bug is that the `rec` and `candidates` produced by subsequent actions, like manual searching, were just kept in local variables instead of assigning them on the `ImportTask` object. That meant that plugins and such could not inspect these new candidates; they only saw the old ones (if they existed).

I'm not sure how wide-ranging the effects of this change will be, so some thorough testing here would be greatly appreciated!

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -691,7 +691,6 @@ class TerminalImportSession(importer.ImportSession):
             return action
 
         # Loop until we have a choice.
-        candidates, rec = task.candidates, task.rec
         while True:
             # Ask for a choice from the user. The result of
             # `choose_candidate` may be an `importer.action`, an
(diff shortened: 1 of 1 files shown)


Vrihub:
Ouch, it seems this patch didn't make it to 1.4.6 :(
FWIW I've been using it so far without any problems.

sampsyo:
Great! If it's working for you, I'll make a note to try out a few things and then merge it.

sampsyo:
Alright; I have finally merged this! Thanks for your patience.

PR #2733, opened 2017-11-03
Added source option in fetchart to fetch cover art using release group

samuelnilsson:
Added a coverartreleasegroup source in fetchart to fetch cover art using release group id instead of release id. Some users may prefer release group cover art instead of release specific ones. The possibility that a release group cover art is found is also higher than the possibility that release specific cover arts are found.

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -288,6 +288,8 @@ class RemoteArtSource(ArtSource):
 
 
 class CoverArtArchive(RemoteArtSource):
+    NAME = u"Cover Art Archive"
+
     if util.SNI_SUPPORTED:
         URL = 'https://coverartarchive.org/release/{mbid}/front'
         GROUP_URL = 'https://coverartarchive.org/release-group/{mbid}/front'
@@ -295,38 +297,18 @@ class CoverArtArchive(RemoteArtSource):
         URL = 'http://coverartarchive.org/release/{mbid}/front'
         GROUP_URL = 'http://coverartarchive.org/release-group/{mbid}/front'
 
-    def getReleaseCandidate(self, album):
-        return self._candidate(url=self.URL.format(mbid=album.mb_albumid),
-                               match=Candidate.MATCH_EXACT)
-
-    def getReleaseGroupCandidate(self, album):
-        return self._candidate(
-                url=self.GROUP_URL.format(mbid=album.mb_releasegroupid),
-                match=Candidate.MATCH_FALLBACK)
-
-
-class CAARelease(CoverArtArchive):
-    NAME = u"Cover Art Archive"
-
     def get(self, album, plugin, paths):
         """Return the Cover Art Archive and Cover Art Archive release group URLs
         using album MusicBrainz release ID and release group ID.
         """
-        if album.mb_albumid:
-            yield self.getReleaseCandidate(album)
+        use_release_group = self._config["use_release_group"]
+        if album.mb_albumid and not use_release_group:
+            yield self._candidate(url=self.URL.format(mbid=album.mb_albumid),
+                                  match=Candidate.MATCH_EXACT)
         if album.mb_releasegroupid:
-            yield self.getReleaseGroupCandidate(album)
-
-
-class CAAReleaseGroup(CoverArtArchive):
-    NAME = u"Cover Art Archive Release Group"
-
-    def get(self, album, plugin, paths):
-        """Return the Cover Art Archive release group URLs using album
-        MusicBrainz release group ID.
-        """
-        if album.mb_releasegroupid:
-            yield self.chooseReleaseGroupCandidate(album)
+            yield self._candidate(
+                    url=self.GROUP_URL.format(mbid=album.mb_releasegroupid),
+                    match=Candidate.MATCH_FALLBACK)
 
 
 class Amazon(RemoteArtSource):
@@ -691,13 +673,12 @@ class FileSystem(LocalArtSource):
 # Try each source in turn.
 
 SOURCES_ALL = [u'filesystem',
-               u'coverart', u'coverartreleasegroup', u'itunes', u'amazon',
-               u'albumart', u'wikipedia', u'google', u'fanarttv']
+               u'coverart', u'itunes', u'amazon', u'albumart', u'wikipedia',
+               u'google', u'fanarttv']
 
 ART_SOURCES = {
     u'filesystem': FileSystem,
-    u'coverart': CAARelease,
-    u'coverartreleasegroup': CAAReleaseGroup,
+    u'coverart': CoverArtArchive,
     u'itunes': ITunesStore,
     u'albumart': AlbumArtOrg,
     u'amazon': Amazon,
@@ -734,6 +715,7 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
             'google_engine': u'001442825323518660753:hrh5ch1gjzm',
             'fanarttv_key': None,
             'store_source': False,
+            'use_release_group': False
         })
         self.config['google_key'].redact = True
         self.config['fanarttv_key'].redact = True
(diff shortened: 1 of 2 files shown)


beetsplug/fetchart.py
```diff
@@ -310,6 +310,24 @@ def get(self, album, plugin, paths):
...
+        URL = 'http://coverartarchive.org/release-group/{mbid}/front'
+
+    def get(self, album, plugin, paths):
+        """Return the Cover Art Archive release group URLs using album
+        MusicBrainz release group ID.
+        """
+        if album.mb_releasegroupid:
+            yield self._candidate(
+                    url=self.URL.format(mbid=album.mb_releasegroupid),
+                    match=Candidate.MATCH_FALLBACK)
```
sampsyo: Would you mind doing a little refactoring to help avoid duplicating some code? For example, a base `CoverArtArchive` class and two subclasses, `CAARelease` and `CAAReleaseGroup`, might help share this code and some of the `SNI_SUPPORTED` logic for choosing the URL. That will make maintenance of this plugin easier in the future.

samuelnilsson: Thank you for the feedback! I made the changes you suggested. I was a little unsure where to write the documentation for the feature since I could not find any documentation on the other fetchart sources. Please, come with suggestions on improvements.

beetsplug/fetchart.py
```diff
@@ -288,26 +288,45 @@ def fetch_image(self, candidate, plugin):
...
-    NAME = u"Cover Art Archive"
-
     if util.SNI_SUPPORTED:
         URL = 'https://coverartarchive.org/release/{mbid}/front'
         GROUP_URL = 'https://coverartarchive.org/release-group/{mbid}/front'
     else:
         URL = 'http://coverartarchive.org/release/{mbid}/front'
         GROUP_URL = 'http://coverartarchive.org/release-group/{mbid}/front'
 
+    def getReleaseCandidate(self, album):
```
sampsyo: This is a tiny thing, but our (and [PEP 8](https://www.python.org/dev/peps/pep-0008/)'s) naming style for methods uses snake_case: i.e., `get_release_candidate`.

docs/plugins/fetchart.rst
```diff
@@ -173,6 +173,14 @@ default engine searches the entire web for cover art.
 Note that the Google custom search API is limited to 100 queries per day.
 After that, the fetchart plugin will fall back on other declared data sources.
 
+Cover art archive
+''''''''''''''''''''
+
+The coverart album art source fetch cover art for a specific release using the
+musicbrainz release id. To fetch cover art for the release group, use
+the coverartreleasegroup source which fetch cover art using the musicbrainz
+release group id.
```
sampsyo: Here, it would be nice to link to http://coverartarchive.org for more information about the source.

Also, here's an idea for alternative wording:

```
The `coverart` source fetches images from the `Cover Art Archive`_. The
default source prioritizes the MusicBrainz release ID: i.e., you get release-
specific art if it's available. To instead use only *release group* album art,
use the source named `coverartreleasegroup`. This will, for example, avoid
using the album art for a specific special edition of an album.

.. _Cover Art Archive: http://coverartarchive.org
```

sampsyo: Also, can you please mention the new name of the source in the "Configuration" section, where we list all the available sources?

beetsplug/fetchart.py
```diff
@@ -708,8 +728,8 @@ def __init__(self):
             'enforce_ratio': False,
             'cautious': False,
             'cover_names': ['cover', 'front', 'art', 'album', 'folder'],
-            'sources': ['filesystem',
-                        'coverart', 'itunes', 'amazon', 'albumart'],
+            'sources': ['filesystem', 'coverart', 'itunes', 'amazon',
+                        'albumart', 'coverartreleasegroup'],
```
sampsyo: This is really bikeshedding, but perhaps a more readable (and shorter) name would be nicer. For example, `coverart_rg` might emphasize that it's an alternative to `coverart` without getting too long about it.

wordofglass:
I really like the idea of matching by different criteria, but I'd like to propose a different way of implementing it: This is potentially a feature that is relevant to all sources. As such, it seems to me that using a new name and class for every criterion might blow up the amount of code unnecessarily, as @sampsyo already recognized. Also, conceptually, independent of the criterion, it is still the same `ArtSource`, and implementing the distinction as a conditional inside the `get` method might feel more natural.

An alternative configuration syntax might look like this
```yaml
fetchart:
    sources: filesystem coverart(release) itunes amazon albumart coverart(releasegroup)
```
where not giving the argument in parentheses would use the previous default. It might be easier to use the following syntax which would offload all of the parsing to YAML:
```yaml
fetchart:
    sources:
        - filesystem
        - coverart:  release
        - itunes
        - amazon
        - albumart
        - coverart: releasegroup
```
The plugin would then pass this to the sources via an additional argument to the `get` method
```diff
-def get(self, album, plugin, paths):
+def get(self, album, match_by, plugin, paths):
```

In the end, this would amount to trading a lower number of classes and not splitting the `get` method for a little more code in the plugin logic and somewhat more verbose configuration syntax, which I would personally prefer. On the other hand, changing the configuration syntax (although while retaining backwards compatibility) might be overkill as long as only one source is actually using it (plus I'm not sure if any other source even has backend support for another search criterion).

Thoughts? If you agree and noone else steps up, I could implement this myself.

sampsyo:
I like this more general approach! I also don't think other sources have variants like this, but this could even just simplify the code for the currently-proposed change.

If I can offer one refinement: instead of adding an argument to the `get` method, perhaps we should add arguments to the backend constructor. Here's where we construct the backend objects: https://github.com/beetbox/beets/blob/9c6910d1675926d9055508f4f67683cda86f9179/beetsplug/fetchart.py#L773-L774

We could pass the extra parameter then and avoid needing to keep track of it in the plugin logic.

samuelnilsson:
I like the more general approach too. I prefer the second suggested configuration syntax. However, is it possible to retain backwards compatibility when using that syntax? I guess it is unnecessary break backwards compatibility for such a minor change, since musicbrainz ids are only used by the coverart source.

sampsyo:
Yes! Good point. Using`coverart` by itself should still be supported.

samuelnilsson:
Another suggestion regarding the configuration syntax is to add a separate configuration key called for example use_release_group that defaults to false. When set to true, the plugin fetches cover art by using release group for the sources where it is possible (which is currently only the coverart source).

This way no additional source would need to be added, the sources configuration syntax would not need to be changed and backwards compatibility would be retained.

What do you think?

samuelnilsson:
I made an initial implementation of what I described above. If you have another idea of how the configuration should look like, that is general and backwards compatible, please come with suggestions.

wordofglass:
I think what I suggested can be implemented in a backwards-compatible manner quite easily by parsing `config['fetchart']['sources']` as follows:
- If it is a scalar, read it with`confit.as_str_seq()` as we do right now and create every source with a default matching criterion
- If it is a list, for each item check
  - if it is a scalar, create a source with that name and the default criterion
  - if it is a mapping (with a single key, maybe validate that?), take the key as name of the source to create and the value as the criterion.

The criteria would be passed to the constructors as suggested by @sampsyo (Thanks for that idea, seems a lot more elegant than my approach!).

With respect to your recent changes: If I'm reading them correctly, it is now less powerful than your previous code since a construction like `sources: coverart coverart_rg` is not possible anymore.

I'll try to come up with code for my idea (not before tomorrow evening, though), but feel free to write it yourself before.

samuelnilsson:
If it is possible to implement your suggested configuration while retaining backwards compatibility, I agree! Feel free to implement your suggestions. I think it is better if it is possible for you do it since it seems like you have more knowledge of the configuration parsing.

wisp3rwind:
Closing in favor of #2739

Issue #2734, opened 2017-11-03
import: using edit candidates prevents auto fetchart and embedart


Vrihub:
### Problem

If I import an album making some changes with the "Edit candidates" option, beets doesn't fetch and embed the album art automatically, as it is supposed to do in my configuration.

This is what happens when I import something without making changes via "Edit candidates":

```
...
Apply, More candidates, Skip, Use as-is, as Tracks, Group albums,
Enter search, enter Id, aBort, eDit, edit Candidates? a
Sending event: import_task_choice
Sending event: import_task_apply
0 of 1 items replaced
Sending event: database_change
Sending event: database_change
Sending event: database_change
Sending event: database_change
fetchart: trying source filesystem for album Wolfgang Amadeus Mozart - Eine Kleine Nachtmusik / Symphony no. 40 in G minor / Symphony no. 41 "Jupiter"
fetchart: trying source coverart for album Wolfgang Amadeus Mozart - Eine Kleine Nachtmusik / Symphony no. 40 in G minor / Symphony no. 41 "Jupiter"
fetchart: downloading image: https://coverartarchive.org/release/00185b27-bb54-40fb-b5cc-b26f246a6078/front
fetchart: downloaded art to: /tmp/tmpTD9lQ9.jpg
fetchart: using remote image /tmp/tmpTD9lQ9.jpg
Sending event: item_copied
Sending event: database_change
Sending event: database_change
Sending event: write
Sending event: after_write
Sending event: database_change
Sending event: import_task_files
Sending event: art_set
embedart: Resizing album art to 600 pixels wide
artresizer: ImageMagick resizing XXXXXXXXXXXXXXXXXXXXXXXXXXXXXX/wolfgang_amadeus_mozart_0000_eine_kleine_nachtmusik___symphony_no._40_in_g_minor___symphony_no._41__jupiter_/albumart.jpg to /tmp/tmpStol6V.jpg
embedart: Embedding album art into Wolfgang Amadeus Mozart - 0000 - Eine Kleine Nachtmusik / Symphony no. 40 in G minor / Symphony no. 41 "Jupiter"
embedart: embedding /tmp/tmpStol6V.jpg
Sending event: write
Sending event: after_write
Sending event: database_change
Sending event: album_imported
Sending event: import
Sending event: cli_exit
```

And here's what happens when I make some changes: it seems the fetchart and embedart plugins are not called at all:

```
...
Enter search, enter Id, aBort, eDit, edit Candidates? c
[...]

PR #2735, opened 2017-11-05
Tentative fix for issue 2734: also check for RETAG

Vrihub:
This is a tentative fix for issue #2734 , based on the suggestion in https://github.com/beetbox/beets/issues/2734#issuecomment-341888432

I've tested it and it fixes the problem for me, but I'm not sure if other changes are required for a complete solution, since my knowledge of the code base is very limited. Please review!

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -783,7 +783,8 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
             if task.choice_flag == importer.action.ASIS:
                 # For as-is imports, don't search Web sources for art.
                 local = True
-            elif task.choice_flag in (importer.action.APPLY, importer.action.RETAG):
+            elif task.choice_flag in (importer.action.APPLY,
+                importer.action.RETAG):
                 # Search everywhere for art.
                 local = False
             else:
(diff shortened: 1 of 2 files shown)


wordofglass:
Yep, that's exactly the code I meant to suggest.

`grep -r RETAG` shows that the `edit` plugin is the only piece of code using this choice and the following comment from `ìmporter.py` also suggests that the difference between `APPLY` and `RETAG` is really only the source of the metadata. Apart from that they both occur for the same kinds of import tasks and the same plugin actions should probably happen.

```python
action = Enum('action',
              ['SKIP', 'ASIS', 'TRACKS', 'APPLY', 'ALBUMS', 'RETAG'])
# The RETAG action represents "don't apply any match, but do record
# new metadata". It's not reachable via the standard command prompt but
# can be used by plugins.
```

Note that I'm not intimately familiar with the importer, either.

sampsyo:
This looks great; thank you!

It looks like Travis has pointed out a too-long line: https://travis-ci.org/beetbox/beets/jobs/297512858#L1095

Also, would you mind adding a changelog entry that summarizes the problem that was fixed?

Vrihub:
Uhm, no idea why AppVeyor fails here: https://ci.appveyor.com/project/beetbox/beets/build/2228

Also, no idea why all my commits here appear as by user "test" WTF?

sampsyo:
Thank you! :sparkles: I'm pretty sure that AppVeyor failure is spurious, so I'll merge this now. Woohoo!

PR #2739, opened 2017-11-13
Fetchart fine-grained source configuration (alternative to #2733)

wisp3rwind:
As per the discussion in #2733, this is my take on a revised fetchart configuration syntax. To give a few examples, all of the following are valid and should do the expected:

```yaml
fetchart:
    sources:
        - filesystem
        - coverart:  release
        - itunes
        - amazon
        - albumart
        - coverart: releasegroup
```
```yaml
fetchart:
    sources:
        - filesystem
        - coverart:  releasegroup
        - itunes
        - '*'
```
```yaml
fetchart:
    sources:
        - filesystem
        - coverart:  releasegroup release
        - itunes
        - '*'
```
and also any old configuration in the following style
```yaml
fetchart:
    sources: filesystem coverart itunes *
```
I'm not entirely satisfied with how to the parsing turned out, since it did require quite a bit of code. I think the whole thing should make for a pretty future-proof syntax at least.

@samuelnilsson: Would you mind testing this for your specific use-case? Thanks a lot!

Missing:
- [ ] ~~Tests~~
- [x] Docs
- [x] pep8

### DIFF
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -511,6 +511,44 @@ def sanitize_choices(choices, choices_all):
     return res
 
 
+def sanitize_pairs(pairs, pairs_all):
+    """Clean up a single-element mapping configuration attribute as returned
+    by `confit`'s `Pairs` template: keep only two-element tuples present in
+    pairs_all, remove duplicate elements, expand ('str', '*') and ('*', '*')
+    wildcards while keeping the original order. Note that ('*', '*') and
+    ('*', 'whatever') have the same effect.
+
+    For example,
+
+    >>> sanitize_pairs(
+    ...     [('foo', 'baz bar'), ('key', '*'), ('*', '*')],
+    ...     [('foo', 'bar'), ('foo', 'baz'), ('foo', 'foobar'),
+    ...      ('key', 'value')]
+    ...     )
+    [('foo', 'baz'), ('foo', 'bar'), ('key', 'value'), ('foo', 'foobar')]
+    """
+    pairs_all = list(pairs_all)
+    seen = set()
+    others = [x for x in pairs_all if x not in pairs]
+    res = []
+    for k, values in pairs:
+        for v in values.split():
+            x = (k, v)
+            if x in pairs_all:
+                if not x in seen:
+                    seen.add(x)
+                    res.append(x)
+            elif k == '*':
+                new = [o for o in others if o not in seen]
+                seen.update(new)
+                res.extend(new)
+            elif v == '*':
+                new = [o for o in others if o not in seen and o[0] == k]
+                seen.update(new)
+                res.extend(new)
+    return res
+
+
 def notify_info_yielded(event):
     """Makes a generator send the event 'event' every time it yields.
     This decorator is supposed to decorate a generator, but any function
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -535,7 +535,7 @@ def sanitize_pairs(pairs, pairs_all):
         for v in values.split():
             x = (k, v)
             if x in pairs_all:
-                if not x in seen:
+                if x not in seen:
                     seen.add(x)
                     res.append(x)
             elif k == '*':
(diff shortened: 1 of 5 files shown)


samuelnilsson:
Well done! I've tested this for my use-case and it seems to work well.

beets/plugins.py
```diff
@@ -478,9 +478,39 @@ def sanitize_choices(choices, choices_all):
...
+            elif s == '*':
+                res.extend(others)
+        seen.add(s)
+    return res
+
+
+def sanitize_pairs(pairs, pairs_all):
+    """Clean up a single-element mapping configuration attribute: keep only
+    choices elements present in pairs_all, remove duplicate elements, expand
+    ('str', '*') and ('*', '*') wildcards while the keeping original order.
```
sampsyo: Maybe a quick example would help describe what's going on here? This seems like a useful utility (and a useful configuration option format), and other plugins in the future might want to use it.

beets/util/confit.py
```diff
@@ -1242,30 +1248,68 @@ def __init__(self, split=True):
         super(StrSeq, self).__init__()
         self.split = split
 
+    def _convert_value(self, x, view):
```
sampsyo: Should this be a static method?

wordofglass: I think it's not easily turned into one since it uses `self.fail`. I guess if at a later point the function turns out to be useful elsewhere, the error handling could still be moved to an outer scope?

beets/plugins.py
wordofglass: An example might be appropriate, but where would I put it? In the [confuse docs](http://confuse.readthedocs.io/en/latest/#validation), an `as_pairs` method is actually documented (which is only very slightly different). There appeared to be no code for it, however (I did not check the history, it might have existed before). So either an inline comment, or a cherry-pick from the confuse repo and an addition to its documentation. I don't think I'm aware of the status of moving confit/confuse out of this repository, so I'm not quite sure which to prefer.

wordofglass:
I'll try to round this up next weekend. At least for the configuration parsing, I'd like to add tests before merging, I don't trust my code sufficiently to do so without.

samuelnilsson:
This has been open for a while. Is there anything I can do to help to get this done? Or do @wisp3rwind plan to implement the remaining changes?

sampsyo:
I’d love to merge this if anyone has time to round things out!

wisp3rwind:
Sorry for not finishing this off. The issue that kept me from doing this quickly was the question of where to put tests for the additions to confit/confuse. There are currently no tests for it in this repository but only in the confuse repo. However, IIRC, that is not quite up-to date with the state of confit in beets.

So I'd appreciate some advice (@sampsyo? sorry for offloading the responsibility) on how to proceed. Both syncing the confuse repository and importing all of the confuse tests into beets would be quite some work. The only easy option I can think of right now that does not fragment the testing infrastructure is not writing tests at all. Which as mentioned earlier I don't really feel comfortable about but could probably live with.

I have no idea of the general plans for confuse as a separate project, though, so I might be missing something. If that is unlikely to happen due to requiring distributor support, could embedding confuse into beets as a git submodule be an option? Assuming that this would play well with the test runner and with python package setup.

Otherwise, only documentation is missing which should quickly be written up.

sampsyo:
Aha, very good point. We would like to spin off the Confuse (a.k.a. Confit) project entirely someday, and you are right that you've stumbled across an incomplete implementation of that plan.

I suggest proceeding without Confuse-specific tests. Let's merge this, make sure everything seems to work, and then I'll take responsibility for syncing up with the external repository. Then tests can go there.

It would be awesome to finish off the separation sometime soon… maybe we can allocate a release for that in the future.

wisp3rwind:
Alright, I did a quick pass over the code and could spot no obvious mistake. Added some documentation, inline comments and changelog, rebased on master (force-push!).

Would be nice if someone native in english could skim the docs, otherwise, I guess I'm fine with merging.

sampsyo:
This looks great! I revised the documentation for the new stuff slightly and merged. :sparkles: Awesome work!

PR #2741, opened 2017-11-18
mediafile._safe_cast: be safer when converting to int

eigengrau:
The regex «[\+-]?[0-9]*» possibly matches a single minus/plus, which was then
passed on to int, raising a ValueError from within _safe_cast. The test suite
covered this for float, but not for int.

We now make sure we actually have a number after the sign by using a Kleene
plus.

### DIFF
--- a/beets/mediafile.py
+++ b/beets/mediafile.py
@@ -159,11 +159,8 @@ def _safe_cast(out_type, val):
             elif not isinstance(val, six.string_types):
                 val = six.text_type(val)
             # Get a number from the front of the string.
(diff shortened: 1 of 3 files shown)


sampsyo:
Looks great; thank you for catching this! Would you mind writing a changelog entry describing what could go wrong (i.e., a ValueError would occur when a number was just + or just -)?

eigengrau:
Thanks! Amended.

sampsyo:
Thanks again!! I've merged the fix.

eigengrau:
Thank you for making beets!

Issue #2740, opened 2017-11-17
hook: Crash when trying to access {item.track} in command


broddo:
I'm playing around with hook in the hope to use it to keep a lossy version of my flac lib. I came across this issue when trying to generate a new path based on the imported item.

Here's the relevant part of the config:

```
hook:
    hooks:
        - event: write
          command: echo "writing track number {item.track}"
```

And the error:

```
File "/usr/local/lib/python2.7/site-packages/beetsplug/hook.py", line 68, in convert_field
    converted = converted.decode(self._coding)
AttributeError: 'int' object has no attribute 'decode'
```

This on Freebsd

sampsyo:
Thank you! It looks like the plugin has a problem handing non-string fields. I’d guess that other numeric fields, like `year` or `disc`, would have the same problem.

jhermann:
Solved, can be closed.

sampsyo:
Indeed; thanks!

PR #2742, opened 2017-11-20
Addresses hook.py ability to handle non-string item properties

broddo:
This is a simple fix for #2740 that catches an AttributeError if the value returned by convert_field() doesn't have a decode() method.

### DIFF
--- a/beetsplug/hook.py
+++ b/beetsplug/hook.py
@@ -68,6 +68,8 @@ class CodingFormatter(string.Formatter):
             converted = converted.decode(self._coding)
         except UnicodeEncodeError:
             pass
+        except AttributeError:
+            converted = value    
 
         return converted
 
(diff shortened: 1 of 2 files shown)


sampsyo:
Thanks for getting this started! How about something like this to make it clear what types are involved?

```py
if isinstance(converted, bytes):
    return converted.decode(self._coding)
return converted
```

Would that cover the same set of cases?

broddo:
That certainly looks better than my approach of "break the rules, then ask for forgiveness" approach! I'll run some tests on the various item properties and get back to you.

broddo:
Yeah, that works just fine. A config file using this: 
```hook:
    hooks:
        - event: write
          command: echo "writing track {item.title}, number {item.track} and replay track gain of {item.rg_track_gain}"
```
results in an output like this:

```writing track Moonshine, number 12 and replay track gain of -9.07```

sampsyo:
Perfect! Thank you! Would you mind adding an entry to `changelog.rst` summarizing the fix?

Also, I know this is really small, but there's some extra whitespace that Travis is complaining about: https://travis-ci.org/beetbox/beets/jobs/305090699#L1099

sampsyo:
This looks perfect! :sparkles: Thank you again!

PR #2750, opened 2017-11-27
Add tags configuration to acousticbrainz plugin

woparry:
The acousticbrainz plugin pulls a lot of information from the database.

However, some people (me) might want to only apply some of the tags from acousticbrainz to our files, either because we want to avoid conflicts with tags from other sources, or because we know we won't use them and like to keep things neat and tidy.

Add a new tags configuration option to the plugin, which allows you to specify a list of tags to write to your files. The default is still to apply all available tags.

### DIFF
--- a/beetsplug/acousticbrainz.py
+++ b/beetsplug/acousticbrainz.py
@@ -110,6 +110,7 @@ class AcousticPlugin(plugins.BeetsPlugin):
         self.config.add({
             'auto': True,
             'force': False,
+            'tags': []
         })
 
         if self.config['auto']:
@@ -164,6 +165,7 @@ class AcousticPlugin(plugins.BeetsPlugin):
     def _fetch_info(self, items, write, force):
         """Fetch additional information from AcousticBrainz for the `item`s.
         """
+        tags = self.config['tags'].get()
         for item in items:
             # If we're not forcing re-downloading for all tracks, check
             # whether the data is already present. We use one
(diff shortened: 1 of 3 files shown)


sampsyo:
This looks great; thank you!

Travis has one small complaint: one line seems to be too long. https://travis-ci.org/beetbox/beets/jobs/307970001#L1096

Also, would you mind adding a changelog entry describing the new option?

beetsplug/acousticbrainz.py
```diff
@@ -164,6 +165,7 @@ def _get_data(self, mbid):
     def _fetch_info(self, items, write, force):
         """Fetch additional information from AcousticBrainz for the `item`s.
         """
+        tags = self.config['tags'].get()
```
sampsyo: One last suggestion here: instead of `.get()`, use `as_str_seq()`. That has a couple of benefits:

- It produces an error early if the user enters the wrong kind of data here (for example, a number) instead of causing an unpredictable Python error later.
- It lets the user enter tags as a whitespace-separated string as an optional alternative to using YAML list syntax, which can be a bit friendlier.

sampsyo:
Awesome; thank you! I left one last suggestion inline. And Travis is still, somewhat infuriatingly, complaining—the max line length is 79 according to PEP8:
https://travis-ci.org/beetbox/beets/jobs/308003861#L1094

sampsyo:
Perfect. Thank you again! I'll merge this now. :sparkles:

Issue #2732, opened 2017-11-02
Question: How to Configure Convert Plugin to Not Convert a Specific Format?


Stunner:
I currently have the configuration for the convert plugin is as follows:

```yaml
convert:
    auto: yes
    command: ffmpeg -i $source -y -vn -aq 2 $dest
    extension: mp3
    dest: /Music/beets/music
    tmpdir: /Music/beets/tmp_convert
```

Currently it converts anything that is non-`mp3` to `mp3` with the command specified in the config, which is fine. However, I want it to **ignore** any `m4a` files but convert anything else it finds to `mp3`. Is there an existing way for me to achieve this? (I have consulted [the docs](http://beets.readthedocs.io/en/v1.4.5/plugins/convert.html) and don't see any obvious way to do this.)

sampsyo:
At the moment, I don't think we have a configuration option quite like this. The idea is similar to the existing `never_convert_lossy_files` option but more general, and also somewhat similar to #1856. But perhaps we should treat this as a new feature request ticket for a simple format filter option? Or we could add an option that filters based on an arbitrary query, which would be a bit more flexible.

Stunner:
I was thinking an `ignore` setting may be intuitive and fit what I describe perfectly. Then I could merely add a comma-separated list of file extensions to ignore like so: `m4a, wma`. However I am intrigued by your notion of filtering based on a query. Could you expand on that?

sampsyo:
Yeah! Here's what a query-based `ignore` setting would look like:

```yaml
ignore: format:AAC, format:WMA
```

or:

```
ignore: path::\.(m4a|wma)$
```

or it could use any other attribute of the music to filter on.

About the semantics of the option: in explicit mode (i.e., when running the `beet convert` command), I assume that the goal would be to copy these files as-is without converting them—not to leave them behind entirely. In that case, `ignore` might not be the clearest name: perhaps something like `noconvert` or `preserve`?

Stunner:
Yeah I think `noconvert` would be the most clear/obvious. When you get a chance would you mind pointing me to where I should start digging through the code to begin implementing this? Some examples of similar commands would be helpful so that I can reference their implementations as an example. Thanks!

sampsyo:
Sure! The proposal is pretty similar to (kind of a generalization of) the `never_convert_lossy_files` option, which is implemented here: https://github.com/beetbox/beets/blob/master/beetsplug/convert.py#L95-L97

[...]

PR #2751, opened 2017-11-30
Convert Plugin: Added dont_convert Option

Stunner:
Addresses #2732. I changed the name from `noconvert` to a more intuitive `dont_convert` which I think falls better in-line with the other options of the convert plugin (i.e. `never_convert_lossy_files`).

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -31,6 +31,8 @@ from beets.plugins import BeetsPlugin
 from beets.util.confit import ConfigTypeError
 from beets import art
 from beets.util.artresizer import ArtResizer
+from beets.library import parse_query_string
+from beets.library import Item
 
 _fs_lock = threading.Lock()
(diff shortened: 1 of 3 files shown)


sampsyo:
Nice! This is a really straightforward, simple solution to the problem.

Would you mind adding a quick changelog entry about the feature?

Also, I have to admit that I kind of prefer the shorter name for the option. I’ve never really liked the verbosity of `never_etc_etc`, which is sort incongruous with the rest of our option names—maybe we should have worked harder on that. And the missing apostrophe in the contraction `dont` is a little more strange, in my opinion.

Other name options include `no_convert` and `copy_only`.

sampsyo:
This is perfect! :sparkles: Thanks for implementing this.

Stunner:
Thank you @sampsyo for being such an awesome and active maintainer of this wonderful project!

Issue #2752, opened 2017-12-05
mpdstats: using playlist to get current song may be a performance issue


ray66:
### Problem

MPDStats fails to update stats for all songs on a low performance system when a large playlist is active. In this case, getting the playlist in _on_play_ may take longer than the current song is playing and stats are not updated for the next song because _status['songid']_ is still pointing to the previous song when the playlist is finally received.

The playlist is just used to get the url of the current song. It would be more efficient do this with the _currentsong_ command which is faster and does create less load on the server.

### Setup

* HW: Raspberry Pi B+
* OS: Raspbian Stretch 
* Python version: 2.7.13
* beets version: 1.4.5
* Turning off plugins made problem go away (yes/no): n/a

My configuration (output of `beet config`) is:

```yaml
mpd:
    host: 127.0.0.1
    port: 6600
    rating: yes
    rating_mix: 0.75
    music_directory: /media/Medien/Musik
    password: REDACTED
smartplaylist:
    auto: no
    relative_to: /media/Medien/Musik/
    playlist_dir: /var/lib/mpd/playlists
    playlists: [{name: rating.m3u, query: 'rating:0.4..1.0'}, {name: not_played.m3u, query: '-play_count: artist:'}, {name: dylan.m3u, query: 'artist:Dylan'}]
mpdstats:
    host: 127.0.0.1
    port: 6600
    music_directory: /media/Medien/Musik
    rating: yes
    rating_mix: 0.75
library: ~/beets/musiclibrary.db

plugins: mpdstats smartplaylist lastgenre random
directory: /media/Medien/Musik/

import:
    copy: no
    write: no

```

sampsyo:
Good point! If a `currentsong` command will work, let's try switching to that. Would you be willing to try making a PR and testing out the change?

PR #2754, opened 2017-12-05
Improved the method to get the path of the current song. 

ray66:
Improves the performance issue described in issue #2752. Before, the complete playlist was fetched from the server. Now, the command "currentsong" is used for this purpose. This improves performance when a huge playlist is active.


sampsyo:
Awesome; thanks! It looks like the tests may need updating—the mocking seems to have assumed the previous implementation. Any chance you can give that a look next?

Issue #934, opened 2014-09-08
Option to clear all fields before applying a match


ecdlguy:
An option/plugin to use only metadata from online sources. 
1. Don't import any of the files original tag data to the beets-database, only use this information to obtain matches from the online sources.
2. Completely clear/delete the entire existing tag before writing the new one.

kooimens:
This would be nice!

tummychow:
I am very interested in implementing this, as the current metadata on my music collection is hot garbage. The genre tag is particularly problematic because musicbrainz doesn't set it, so I end up importing whatever genre was already tagged on the track. I'd much prefer beets to be the originator for all the tags that end up on my imported music.

I'd like to confirm my understanding of the problem before I work on a PR:

- the changes should be made in the apply_metadata function [here](https://github.com/beetbox/beets/blob/master/beets/autotag/__init__.py#L61) or somewhere in the callchain leading there ([ImportTask.apply_metadata](https://github.com/beetbox/beets/blob/master/beets/importer.py#L534) is around where I'm thinking).
- to achieve the desired effect, every [field](https://github.com/beetbox/beets/blob/master/beets/library.py#L424) in the Item should be zeroed before applying the matched metadata. This zeroing should only happen if the user actually chooses apply.
- a new configuration option (`import.from_scratch`) should be added to toggle the behavior globally, with a corresponding flag (`--from-scratch`) to toggle the behavior at runtime. Anyone want to bikeshed on the name of this?
- I need to add tests for all new functionality.

PR #2755, opened 2017-12-07
Implement from_scratch option

tummychow:
Okay well, I asked for advice in #934, but I accidentally implemented it before getting an answer. Whoops!

Feel free to criticize style, implementation details, naming, tests/lack thereof, etc.

### DIFF
--- a/docs/reference/config.rst
+++ b/docs/reference/config.rst
@@ -475,6 +475,15 @@ Either ``yes`` or ``no``, controlling whether imported directories are
 recorded and whether these recorded directories are skipped.  This
 corresponds to the ``-i`` flag to ``beet import``.
 
+.. _from_scratch:
+
+from_scratch
+~~~~~~~~~~~~
+
+Either ``yes`` or ``no`` (default), controlling whether existing metadata is
+discarded when a match is applied. This corresponds to the ``-from_scratch``
+flag to ``beet import``.
+
 quiet_fallback
 ~~~~~~~~~~~~~~
 
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -10,6 +10,7 @@ import:
     delete: no
     resume: ask
     incremental: no
+    from_scratch: no
     quiet_fallback: skip
     none_rec_action: ask
     timid: no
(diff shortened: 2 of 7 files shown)


tummychow:
Ping on this? @sampsyo ?

jackwilsdon:
That looks good to me, although I'll let @sampsyo have the final call 😂 👍

docs/reference/config.rst
```diff
@@ -475,6 +475,15 @@ Either ``yes`` or ``no``, controlling whether imported directories are
 recorded and whether these recorded directories are skipped.  This
 corresponds to the ``-i`` flag to ``beet import``.
 
+.. _from_scratch:
+
+from_scratch
+~~~~~~~~~~~~
+
+Either ``yes`` or ``no`` (default), controlling whether existing metadata is
+discarded when a match is applied. This corresponds to the ``-from_scratch``
```
sampsyo: Tiny typo: missing one dash in the option name.

sampsyo:
Sorry for the long delay in reviewing this! Thanks for waiting while it got to the top of my list.

This looks great: it's simple and it matches what users have been asking for. While we're here, I want to briefly take stock of all the ways we have for "clearing out metadata," which have grown steadily:

* The `scrub` plugin deletes all metadata and writes it back again, which removes **tags that beets doesn't know about** that show up in files' metadata.
* The `zero` plugin can selectively remove some metadata fields, or (in "whitelist" mode) only keep certain fields.
* This new feature is sort of like configuring `zero` to keep only the fields that beets fetches from remote servers.

Some users might want to configure *both* `scrub` and this new `from_scratch` option, or *also* enable `zero` to turn off some fields that come from remote sources.

It might be worth documenting this set of choices centrally somewhere, if anyone has a bright idea about where this should live.

sampsyo:
Merged! :sparkles: Thank you again for the clean, simple implementation and the readable documentation.

tummychow:
My pleasure, thank you for merging.

For myself, I actually use all three of scrub, zero, and from_scratch (zero for music that I import as-is, where from_scratch wouldn't apply). I personally think the detailed documentation on scrub/zero is sufficient for me to understand what they do and why I would want them, but my perspective might be skewed after thinking about and implementing this feature.

PR #2758, opened 2017-12-10
Duplicates plugin: tiebreak album vs items fix

cgevans:
At present, if tiebreak is specified to give keys for the order of priority in determining duplicates, it must be specified for both items and albums, or an error will be raised when the unspecified one is used.  This is particularly bad because the default order of priority appears to be impossible to specify in tiebreak, so if the user wants (as in my case) a tiebreak specification for items, and the default order for albums, this is impossible.

This small change makes it so that tiebreak is only used if it specifies keys for ordering the kind (track/album) being considered; otherwise, the default is used.

### DIFF
--- a/beetsplug/duplicates.py
+++ b/beetsplug/duplicates.py
@@ -253,13 +253,12 @@ class DuplicatesPlugin(BeetsPlugin):
         "completeness" (objects with more non-null fields come first)
         and Albums are ordered by their track count.
         """
-        if tiebreak:
-            kind = 'items' if all(isinstance(o, Item)
-                                  for o in objs) else 'albums'
+        kind = 'items' if all(isinstance(o, Item) for o in objs) else 'albums'
+
+        if tiebreak and kind in tiebreak.keys():
             key = lambda x: tuple(getattr(x, k) for k in tiebreak[kind])
         else:
-            kind = Item if all(isinstance(o, Item) for o in objs) else Album
-            if kind is Item:
+            if kind is 'items':
                 def truthy(v):
                     # Avoid a Unicode warning by avoiding comparison
                     # between a bytes object and the empty Unicode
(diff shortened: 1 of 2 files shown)


beetsplug/duplicates.py
```diff
@@ -253,13 +253,12 @@ def _order(self, objs, tiebreak=None):
...
-            kind = 'items' if all(isinstance(o, Item)
-                                  for o in objs) else 'albums'
+        kind = 'items' if all(isinstance(o, Item) for o in objs) else 'albums'
+
+        if tiebreak and kind in tiebreak.keys():
             key = lambda x: tuple(getattr(x, k) for k in tiebreak[kind])
         else:
-            kind = Item if all(isinstance(o, Item) for o in objs) else Album
-            if kind is Item:
+            if kind is 'items':
```
sampsyo: I think the proper comparison operator on strings is `==`, not `is`.

beetsplug/duplicates.py
```diff
@@ -253,13 +253,12 @@ def _order(self, objs, tiebreak=None):
...
         """
-        if tiebreak:
-            kind = 'items' if all(isinstance(o, Item)
-                                  for o in objs) else 'albums'
+        kind = 'items' if all(isinstance(o, Item) for o in objs) else 'albums'
+
+        if tiebreak and kind in tiebreak.keys():
             key = lambda x: tuple(getattr(x, k) for k in tiebreak[kind])
         else:
-            kind = Item if all(isinstance(o, Item) for o in objs) else Album
```
sampsyo: Line 269, below, assumes that `kind` is a Model class, not a string.

cgevans: Oops... that code is never run in my specific configuration, but line 269 just needs Item.all_keys(), so it's an easy change.

beetsplug/duplicates.py
cgevans: Yes, it is... oops.

cgevans:
Those issues should be fixed now, thanks.

sampsyo:
Perfect; thank you! :sparkles:

PR #2759, opened 2017-12-11
First tentative fix for issue 2738.

Vrihub:
Don't ignore anymore filenames such as "01.mp3", instead use the number
in the filename as track number.

### DIFF
--- a/beetsplug/fromfilename.py
+++ b/beetsplug/fromfilename.py
@@ -28,18 +28,13 @@ import six
 # Filename field extraction patterns.
 PATTERNS = [
     # Useful patterns.
-    r'^(?P<artist>.+)-(?P<title>.+)-(?P<tag>.*)$',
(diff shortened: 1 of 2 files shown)


samawsome:
#github

sampsyo:
Hi! Sorry for the delay in reviewing this, but it looks perfect!

Would you mind adding a brief changelog entry about the new capability? Then this should be ready to merge as-is, if it's been working for you. :sparkles:

Vrihub:
Ok, I've added a changelog entry (and fixed a conflict, since my branch was a bit old), did some more fixes to the regular expressions (see commit message 84cf336 ), fixed a line-too-long.
If you agree with the changes I think we should be ready to merge.

sampsyo:
Perfect! Thank you for the updates, and for the clean fix for this shortcoming. :sparkles: Woohoo!

Issue #2729, opened 2017-11-01
ImageMagick applies maxwidth to longest edge instead of to width


translit:
The [current invocation](https://github.com/beetbox/beets/blob/2120cf68c61649c22c14f20d83bd28d758720557/beets/util/artresizer.py#L96-L100) of ImageMagick applies the ```maxwidth``` parameter of ```fetchart``` to an image's longer edge instead of to its width. Possible solution suggested by Adrian [on the forum](https://discourse.beets.io/t/fetchart-taming-cover-art-resolution/206/12?u=dorade).

sampsyo:
Thank you for the summary! I'm marking this as a "bitesize" bug because I think it should be a good entry point for anyone looking to get started contributing to beets.

wordofglass:
Quoting the [ImageMagick documentation](http://www.imagemagick.org/Usage/resize/#shrink) linked to in the forum

>The Only Shrink Flag ('>' flag) is a special character in both UNIX Shell and in Window batch scripts, and you will need to escape that character (using backslash '\>' in shell, and '^>' in windows batch)

> As of IM v6.3.8-3 IM now has a new geometry option flag '^' which is used to resize the image based on the smallest fitting dimension.

and noting the [inline comment](https://github.com/beetbox/beets/blob/2120cf68c61649c22c14f20d83bd28d758720557/beets/util/artresizer.py#L91-L94)
```
# "-resize widthxheight>" shrinks images with dimension(s) larger
# than the corresponding width and/or height dimension(s). The >
# "only shrink" flag is prefixed by ^ escape char for Windows
# compatibility.
```
the fix might not be as obvious. The way I read this (without having done any testing/git blame myself)
- we do not want to drop the `>` flag since that would upscale small pictures
- the issue came about because the `^` flag was only introduced after our ImageMagick invocation was written. Unfortunately, that flag collides with the `^` escape character on Windows.

@translit, am I correct assuming that you ran into this on Linux and not Windows?

Unless there's another way to escape this syntax (quotes? I've really got no clue about Windows shell), it might be necessary to distinguish Unix-based systems and Windows when constructing the command line. Ideally, there would be a test for each of the architectures.

sampsyo:
Aha! Thanks for catching the potential pitfall about small images.

[...]

PR #2762, opened 2017-12-18
Remove unnecessary escape caret in ImageMagick convert argument

sekjun9878:
Hello,

I'm looking to get started contributing to beets with this simple pull request :)

Regarding the Windows caret (`^`) escaping issue, it seems that this is only an issue if we launch ImageMagick by cmd.exe. We instead launch using `CreateProcess()` on Windows.

In calling `CreateProcess()`, Python takes no special actions for neither `^` nor `>`, as referenced [here](https://docs.python.org/2/library/subprocess.html#converting-argument-sequence)

Because Windows uses strings, not arrays, to pass argv to processes, further command line parsing still happens inside the called process (i.e. escapes are still parsed even when called using CreateProcess()) and this behaviour seems dependent entirely on the compiler used. However, it seems that the caret (`^`) is not one of those escapes handled by the compiler, as referenced [here](http://daviddeley.com/autohotkey/parameters/parameters.htm#WINCRULESCHANGE). Specifically,

> **The caret character (^) is not recognized as an escape character or delimiter. The character is handled completely by the command-line parser in the operating system before being passed to the argv array in the program.**
> They are referring to the scenario discussed in sec. 6.2 below, where first the command line parser (cmd.exe) parses your command, handling such things as the escape character ^ , the redirection characters > & < , the pipe character | , the %  character which may identify environment variables that need to be expanded (e.g. %PROGRAMFILES%), etc. **The rules here describe how your C/C++ executable will parse the lpCommandLine that was passed to CreateProcess() by cmd.exe or whoever calls CreateProcess().**

Issue #2729

### DIFF
--- a/beets/util/artresizer.py
+++ b/beets/util/artresizer.py
@@ -88,14 +88,13 @@ def im_resize(maxwidth, path_in, path_out=None):
     log.debug(u'artresizer: ImageMagick resizing {0} to {1}',
               util.displayable_path(path_in), util.displayable_path(path_out))
 
-    # "-resize widthxheight>" shrinks images with dimension(s) larger
-    # than the corresponding width and/or height dimension(s). The >
-    # "only shrink" flag is prefixed by ^ escape char for Windows
-    # compatibility.
+    # "-resize WIDTHx>" shrinks images with the width larger
+    # than the given width while maintaining the aspect ratio
+    # with regards to the height.
     try:
         util.command_output([
             'convert', util.syspath(path_in, prefix=False),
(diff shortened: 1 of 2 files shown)


sampsyo:
Awesome; thanks! Based on the discussion in #2729, I think this is the right fix.

Could you please also add a brief changelog entry describing what was fixed, from the user’s perspective? Then we can merge this up.

sekjun9878:
@sampsyo Done!

sampsyo:
Perfect! Thank you! :sparkles: I've merged the changes.

PR #2775, opened 2017-12-30
Fixes #2688: Skip non-audio media from MusicBrainz

### LINKED ISSUE #2688: Don't try to match non-audio media fetched from MusicBrainz
Matching a folder containing all audio files of a MB release made of CD+DVD for instance fails since beets does take into account also the video files.
For instance (1 CD+ 1DVD):
https://musicbrainz.org/release/2775743e-2d9c-4b7f-9baa-1d24dc44c769
https://musicbrainz.org/release/f457c3e1-0d98-4904-9d9e-a085bc5b232b
I imagine the same happens for enhanced CDs with some additional video track.
In general I think the right thing to do is that all media files that are marked as video in MB should not be considered for matching. After all beets currently only deals with audio files.

nguillaumin:
Some releases have non-audio media, such as CD+DVD or CD+DVD-Video. Skip
these media when fetching album info as they will never match audio
tracks and will always report missing tracks.

I took the naive approach of cherry-picking a list of media suspected to
not contain audio from the MusicBrainz formats list:

https://musicbrainz.org/doc/Release/Format

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -36,6 +36,9 @@ if util.SNI_SUPPORTED:
 else:
     BASE_URL = 'http://musicbrainz.org/'
 
+NON_AUDIO_FORMATS = ['Data CD', 'DVD', 'DVD-Video', 'Blu-ray', 'HD-DVD', 'VCD',
+                     'SVCD', 'UMD', 'VHS']
+
 musicbrainzngs.set_useragent('beets', beets.__version__,
                              'http://beets.io/')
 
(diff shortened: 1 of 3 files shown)


nguillaumin:
This is the first time I do something in Python, so let me know if I did anything wrong, thanks!

sampsyo:
This looks exactly right! Thanks for including the tests, which look very reasonable, and even the changelog entry. Woohoo! :sparkles:

Congratulations on a successful first foray into Python. :snake: :tada:

nguillaumin:
Great thanks!

PR #2777, opened 2017-12-31
Fix fragile parsing of bs1770gain output

ylep:
bs1770gain is sorting the filenames internally, so in some corner cases it returns the results in a different order from the command-line arguments.

This patch uses the --xml option of bs1770gain (introduced in 0.4.6) in order to read the filenames reliably.


sampsyo:
Thanks; this looks great! Parsing the XML from the output seems more robust than our old regex parser anyway.

As with #2778, can you please add a changelog entry? It might also be useful to add a minimum bs1770gain version to the plugin's docs.

sampsyo:
Perfect! Thank you for these updates—and for finding the problem in the first place. :sparkles: Awesome work!

PR #2780, opened 2018-01-02
Added a setting to control which formats are ignored

nguillaumin:
This is related to #2688 where a list of hard-coded non-audio formats to
ignore has been added. Some users may want to rip the audio portion of
video tracks (e.g. DVD-Video) so it would be beneficial to let them
control exactly which formats to ignore.

I added a `ignored_formats` setting for that purpose and moved the
hard-coded list into the config. Test and documentation have been
updated accordingly.

Aside: I also clarified the changelog a bit regarding this change and
the related one for #1210.

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -277,7 +277,7 @@ def album_info(release):
         disctitle = medium.get('title')
         format = medium.get('format')
 
-        if format in config['match']['ignored_formats'].as_str_seq():
+        if format in config['match']['ignored_media'].as_str_seq():
             continue
 
         all_tracks = medium['track-list']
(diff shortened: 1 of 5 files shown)


sampsyo:
Looking good!

For consistency with the other current config options in the `match` section, what do you think about the name `ignored_media` instead? We have `match.preferred.media` already, for example.

AFAICT, the Travis failures above seem to be spurious (just Travis flakiness):
https://travis-ci.org/beetbox/beets/jobs/324249126

nguillaumin:
Sure makes sense, I like consistency too :wink: I just pushed another commit with the setting renamed.

sampsyo:
Looks perfect! Thank you again! :sparkles:

nguillaumin:
No worries, thank you for Beets!

PR #2782, opened 2018-01-04
New importer option incremental_skip_later, fixes #2773

### LINKED ISSUE #2773: When importer is in incremental mode, offer an option to avoid recording the album
It would be useful to have a "[L]ater" plugin that allows to non-permanently skip an import in incremental mode, in order to be processed again in subsequent import runs. That would basically be the "[S]kip" functionality, minus storing the path.

Typical use case would be, in a large volume import, to skip the difficult cases for the time being and to re-visit these later.

sekjun9878:
This pull request implements the configuration option `incremental_skip_later`, which changes the skip behaviour to bypass recording skipped items to the incremental list, allowing them to be revisited later on.

The configuration option defaults to current behaviour.

Issue #2773

### DIFF
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -10,6 +10,7 @@ import:
     delete: no
     resume: ask
     incremental: no
+    incremental_skip_later: no
     from_scratch: no
     quiet_fallback: skip
     none_rec_action: ask
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -577,7 +577,8 @@ class ImportTask(BaseImportTask):
         # Update progress.
         if session.want_resume:
             self.save_progress()
-        if session.config['incremental']:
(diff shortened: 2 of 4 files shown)


sampsyo:
Looks cool!

Since the name is already little long, what do you think about something more explicit, like `incremental_skip_record` (with the sense inverted), or `incremental_skip_later` (suggesting the use case, which is to come back and import those albums later)?

sekjun9878:
@sampsyo I've changed the config name. It's ready for review!

sampsyo:
Looks perfect! Thank you!

And, in general, thanks for all your contributions lately. I'm going to invite you to the beetbox organization, which should let you commit directly. Pull requests are still encouraged, of course, so we can review each other's code, but this will let you hit the merge button if you like. :tada:

Issue #2768, opened 2017-12-27
beet move IOError


vissieGP:
Hi. I'm trying to move my library to a new location.
Old location: /export/zfs/Music/Sorted
New Locasion: /export/zfs/Music/Sorted/Vissie 

So I ran the following command: 
beet move -p -d /export/zfs/Music/Sorted/Vissie/
That worked.
Then I run:
beet move -d /export/zfs/Music/Sorted/Vissie/
This gave me this error:
IOError: [Errno 2] No such file or directory: '/export/zfs/Music/Sorted/_/_.mp3'
I remembered that I had a file with issues that I fixed with ncmpcpp. So I was not too worried with this error.
1. Should the pretend not have given the same error? Why did everything seem to be fine with the pretend, but gave a fatal python error with the actual command?
2. Can the move not skip error and complete? This way it will not break have way into the process leaving one in the middle so to speak?

Seeing that I knew what was the issue (and as luck might have had it, that was the very first file that was moved!!) I continued. 
I ran:
beet update. 

That seemed to fix and update my database.
At some point, I got the following error:
IOError: [Errno 2] No such file or directory: '/export/zfs/Music/Sorted/Death Cab for Cutie/I Will Possess Your Heart/cover.1.jpg'

Not sure why, I just re-ran the update process. This time it finalized without any issues. I ran the same command again. Again, no error.

Thinking that I was good to go, I tried the move again, so I ran:
beet move -p -d /export/zfs/Music/Sorted/Vissie/
No errors this time. 
I ran:
beet move -d /export/zfs/Music/Sorted/Vissie/

This time it started to move my files, one by one. And sadly, I got this error:
Sending event: database_change
moving album art /export/zfs/Music/Sorted/Death Cab for Cutie/I Will Possess Your Heart/cover.1.jpg to /export/zfs/Music/Sorted/Vissie/Death Cab for Cutie/I Will Possess Your Heart/cover.jpg
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/beets/util/__init__.py", line 471, in move
    shutil.copyfile(path, dest)
  File "/usr/lib/python2.7/shutil.py", line 82, in copyfile
    with open(src, 'rb') as fsrc:
IOError: [Errno 2] No such file or directory: '/export/zfs/Music/Sorted/Death Cab for Cutie/I Will Possess Your Heart/cover.1.jpg'

[...]

Issue #2724, opened 2017-10-31
edit operation fails badly when albumart file missing


Vrihub:
### Problem

If I delete the albumart.jpg file from a directory in the beets tree and then I run an `edit` command that implies renaming the directory (e.g. because I changed a variable that is used in the directory name), beets exits badly and leaves me with two corrupted directories.

```
moving album art (oldpath) to (newpath)
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/beets-1.4.4-py2.7.egg/beets/util/__init__.py", line 471, in move
    shutil.copyfile(path, dest)
  File "/usr/lib/python2.7/shutil.py", line 82, in copyfile
    with open(src, 'rb') as fsrc:
IOError: [Errno 2] No such file or directory: (the albumart.jpg file that I had deleted)
```

Also, I'm left with two incomplete directories:
- The new directory, renamed as per my `edit` command, containing only the first file of the album
- The old directory, containing the remaining files

Suggested fix: the edit operation should ignore silently if the albumart file is missing.

### Setup

* OS: Linux
* Python version: 2.7.13
* beets version: 1.4.4

sampsyo:
Indeed! This is, I think, the same root issue as in #1926—that is, we should log an error instead of crashing when moving fails *in any command*. (I think we can probably address this with some centralized logic instead of changing every command that moves files.) Does that look right to you?

Vrihub:
> Does that look right to you?

Yes, I thought it was only an `edit` problem; sorry but I couldn't catch #1926 and  #2419 before opening this issue. So I think you can mark this as a duplicate and try to find a more general solution to the problem.

sampsyo:
OK! No problem at all. :sparkles:

Issue #2419, opened 2017-01-29
beet update: Do not stop when files (including album art) cannot be moved


azrdev:
### Problem

Running `beet up` fails for me with messages like
~~~
moving album art /srv/music/beets/ASP/Aus der Tiefe_ Der schwarze Schmetterling, Teil IV/cover.jpg to /srv/music/beets/ASP/Aus der Tiefe_ Der schwarze Schmetterling, Teil IV [5622]/cover.1.
jpg
Traceback (most recent call last):
  File "/srv/home/azrael/dev/beets/upstream/beets/util/__init__.py", line 470, in move
    shutil.copyfile(path, dest)
  File "/usr/lib64/python2.7/shutil.py", line 82, in copyfile
    with open(src, 'rb') as fsrc:
IOError: [Errno 2] No such file or directory: '/srv/music/beets/ASP/Aus der Tiefe_ Der schwarze Schmetterling, Teil IV/cover.jpg'
~~~

### Setup

* OS: Fedora 23
* Python version: 2.7.11
* beets version: v1.4.3-71-g0a13240 (git HEAD)
* Turning off plugins made problem go away (yes/no): 

My configuration (output of `beet config`) is:

```yaml
importfeeds:
    relative_to: /srv/music
    dir: /srv/music/lib/beets_importfeeds
    formats: m3u
    absolute_path: no
    m3u_name: imported.m3u

paths:
    singleton: $albumartist/$album%aunique{}/$track $title
    comp: $albumartist/$album%aunique{}/$track $title
play:
    command: bash -c 'cat $0 | mpc add'
    relative_to: /srv/music
    use_folders: no
    raw: no
    warning_threshold: 100
mpdstats:
    host: localhost
    port: 6600
    music_directory: /srv/music/
    rating: no
chroma:
    auto: no
convert: (...)
embedart:
    auto: no
    compare_threshold: 0
    ifempty: no
    remove_art_file: no
    maxwidth: 0
library: ~/.share/beets/library.blb

plugins: acousticbrainz embedart fetchart fromfilename info missing duplicates chroma info convert importfeeds play
directory: /srv/music/beets

import:
    move: yes
    log: ~/.share/beets/import.log
fetchart:
    auto: yes
    minwidth: 0
    sources:
    - filesystem
    - coverart
    - itunes
    - amazon
    - albumart
    google_engine: 001442825323518660753:hrh5ch1gjzm
    enforce_ratio: no
    cautious: no
    maxwidth: 0
[...]

PR #2036, opened 2016-06-04
add file check to move command

pszxzsd:
Currently missing files, including album art, cause move operations to error out midway, leaving things in a dirty state with half moved albums.

This patch adds a pre-check to the move command and aborts if any audio file to be moved is missing. Missing album art triggers a warning and is dereferenced in the library.

PR #2785, opened 2018-01-07
Catch and warn FilesystemError during Album/Item move()

msjyoo:
Hello,

This pull request fixes the old bug where filesystem errors during a batch move or update would halt the entire operation, leaving your files in an inconsistent state.

This patch implements the proposed fix of warning the user of the error, and continuing on with the rest of the files, such as a missing album art which is inconsequential in many cases and does not warrant a fatal halt of the program.

This patch also considers situations where a user may have forgotten to mount a music folder, taking care not to update library path references if a filesystem error has occurred. This prevents surprising behaviour leading to the potential loss of music files.

Issues #1926, #2036, #2768, #2419, #2724, and other duplicates.

Ready for comments and review.

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -29,7 +29,7 @@ from beets.mediafile import MediaFile, UnreadableFileError
 from beets import plugins
 from beets import util
 from beets.util import bytestring_path, syspath, normpath, samefile, \
-    MoveOperation
+    MoveOperation, FilesystemError
 from beets.util.functemplate import Template
 from beets import dbcore
 from beets.dbcore import types
@@ -701,27 +701,32 @@ class Item(LibModel):
         """
         if not util.samefile(self.path, dest):
             dest = util.unique_path(dest)
-        if operation == MoveOperation.MOVE:
-            plugins.send("before_item_moved", item=self, source=self.path,
-                         destination=dest)
-            util.move(self.path, dest)
-            plugins.send("item_moved", item=self, source=self.path,
-                         destination=dest)
-        elif operation == MoveOperation.COPY:
-            util.copy(self.path, dest)
-            plugins.send("item_copied", item=self, source=self.path,
-                         destination=dest)
-        elif operation == MoveOperation.LINK:
-            util.link(self.path, dest)
-            plugins.send("item_linked", item=self, source=self.path,
-                         destination=dest)
-        elif operation == MoveOperation.HARDLINK:
-            util.hardlink(self.path, dest)
-            plugins.send("item_hardlinked", item=self, source=self.path,
-                         destination=dest)
 
-        # Either copying or moving succeeded, so update the stored path.
-        self.path = dest
+        try:
+            if operation == MoveOperation.MOVE:
+                plugins.send("before_item_moved", item=self, source=self.path,
+                             destination=dest)
+                util.move(self.path, dest)
+                plugins.send("item_moved", item=self, source=self.path,
+                             destination=dest)
+            elif operation == MoveOperation.COPY:
+                util.copy(self.path, dest)
+                plugins.send("item_copied", item=self, source=self.path,
+                             destination=dest)
+            elif operation == MoveOperation.LINK:
+                util.link(self.path, dest)
+                plugins.send("item_linked", item=self, source=self.path,
+                             destination=dest)
+            elif operation == MoveOperation.HARDLINK:
+                util.hardlink(self.path, dest)
+                plugins.send("item_hardlinked", item=self, source=self.path,
+                             destination=dest)
+        except FilesystemError as exc:
+            raise exc
+        else:
+            # Do not update library reference if filesystem operation failed
+            # Either copying or moving succeeded, so update the stored path.
+            self.path = dest
 
     def current_mtime(self):
         """Returns the current mtime of the file, rounded to the nearest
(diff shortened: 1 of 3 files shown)


sekjun9878:
Implementation consideration diary for myself:

Originally, I tried adding a new set of Album / Item commands `try_move()` etc. I didn't want to unnecessarily change the exceptional-fail behaviour and there were prior examples of `write()` and `try_write()`. However, doing this failed to solve bugs in `edit` implicitly, since it would require changing every invocation of `move()` to `try_move()`. It was also impossible to implement `try_move()` for Albums cleanly, since catching exceptions outside of `move()` would result in the iteration breaking anyway.

I then decided to implement `try_move()` at a lower, filesystem operation level. This would solve the problem of refactoring every invocation, since `util.move()` is a function, not a class method, and thus more reliable to statically refactor using an IDE. This would also solve the `edit` bug implicitly, since only Album/Item `move()`'s internal logic needs to be modified. This meant that it was no longer possible for Album/Item `move()` API to generate exceptions, but it is a trade-off that never leaves your files in 2 separate places. However, this didn't allow for conditional update of library references, since that was done at a higher level, and checking `True`/`False` for every `move`/`copy`/`link` would be unreasonably complicated logic.

Then since I already decided to forgo the fail-hard API of Album/Item `move()`, I modified `move_file()` to conditionally update library references and float exceptions, and `move()` to handle the exceptions.

sekjun9878:
Looks like the build error is happening because `Unidecode==0.4.21` got updated to `Unidecode==1.0.22`

beets/library.py
```diff
@@ -701,27 +701,32 @@ def move_file(self, dest, operation=MoveOperation.MOVE):
...
+            elif operation == MoveOperation.LINK:
+                util.link(self.path, dest)
+                plugins.send("item_linked", item=self, source=self.path,
+                             destination=dest)
+            elif operation == MoveOperation.HARDLINK:
+                util.hardlink(self.path, dest)
+                plugins.send("item_hardlinked", item=self, source=self.path,
+                             destination=dest)
+        except FilesystemError as exc:
+            raise exc
```
sampsyo: Just `raise` will do here. (Re-raising the currently-handled exception is implicit.)

beets/library.py
```diff
@@ -701,27 +701,32 @@ def move_file(self, dest, operation=MoveOperation.MOVE):
         """
```
sampsyo: It might be nice to mention in the docstring that this is "fail-stop," and to mention the kinds of exceptions that can be thrown.

beets/library.py
```diff
@@ -701,27 +701,32 @@ def move_file(self, dest, operation=MoveOperation.MOVE):
...
+            elif operation == MoveOperation.HARDLINK:
+                util.hardlink(self.path, dest)
+                plugins.send("item_hardlinked", item=self, source=self.path,
+                             destination=dest)
+        except FilesystemError as exc:
+            raise exc
+        else:
+            # Do not update library reference if filesystem operation failed
+            # Either copying or moving succeeded, so update the stored path.
+            self.path = dest
```
sampsyo: Hmm, I think there may be something I'm not quite following here. Is there something that this new structure:

```
try:
  <the operation>
except:
  raise
else:
  <finalize>
```

accomplishes that the old `try`-less structure:

```
<the operation>
<finalize>
```

does not? It seems to me that, in both cases, the finalization step (self.path = dest) only gets executed when the move operation doesn't throw an exception. Is that right?

beets/util/__init__.py
```diff
@@ -94,6 +96,14 @@ def log(self, logger):
...
         logger.error(u'{0}: {1}', self.error_kind, self.args[0])
 
+    def log_as_warn(self, logger):
+        """Log to the provided `logger` a human-readable message as a
+        warning and a verbose traceback as a debug message.
+        """
+        if self.tb:
+            logger.debug(self.tb)
+        logger.warn(u'{0}: {1}', self.warn_kind, self.args[0])
+
```
sampsyo: I totally get the motivation here, but I also think it would be fine to log these as errors—the operation really did encounter a failure to do what the user asked to do, which I think rises above the level of a warning.

sekjun9878:
@sampsyo Thanks for the valuable feedback! The easiest way to preserve API compatibility would be adding a `soft=` named parameter that defaults to `False`, and we set the flag on places where users have reported abrupt aborts, such as `move_art`. This means some occurrences won't be fixed automatically, but we can always add more `soft=True` later on. This is for the *log error and truck on* solution.

The only sane way I can see for preventing duplicate album directories is for batch moves to copy all the files first, and THEN update the library references for all, and THEN delete and prune the original files. This way, a read error (i.e. No file exists) or a write error (i.e. file already exists / permissions) will not create duplicate library directories, but rather leave some stale files at the destination directory.

sampsyo:
Got it! Yes, a `soft` parameter might be the right thing. But just to check: a “standard” way you might expect this to work in Python would be that the client is responsible for using a `try` block to “truck on”:

```py
try:
    album.move()
except FilesystemError as exc:
    exc.log()
```

Of course, as you said above, catching exceptions externally for albums could be a problem. But this makes me wonder whether it’s possible to work around that problem elegantly—for example, maybe some cleanup for the item iteration could be put into a `finally` block to make sure it runs even when an exception occurs?

PR #2796, opened 2018-01-20
Remove reference to Tomahawk player in advanced.rst

adamhsparks:
Tomahawk is abandoned according to the GitHub README.md. Suggest removing the reference to using it with Beets library.


sampsyo:
So it goes. That’s awful news, but thanks for noticing!

PR #2798, opened 2018-01-27
Small performance optimization when fetching items from library.

pprkut:
Fetch flexible attributes once for all relevant items instead of once per item.
For queries with larger result sets this drastically cuts down the number of issued sqlite queries.

In the end, this didn't quite give the performance boost I'd hoped it would, but it feels still slightly faster for larger queries.
Unfortunately the hot spot seems to be somewhere else though :-/

One thing to note, since the list of IDs used in the WHERE..IN clause can be quite large (and reach sqlite's query parameter limit), I resorted to not passing them as parameters.
This should be fine, since these are not values originating from input but coming from the database from a previous query. Not sure how big of an issue this is nonetheless.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -565,6 +565,13 @@ class Results(object):
         a `Results` object a second time should be much faster than the
         first.
         """
+
+        # First fetch all flexible attributes for all items in the result.
+        # Doing the per-item filtering in python is faster than issuing
+        # one query per item to sqlite.
+        item_ids = [row['id'] for row in self._rows]
+        flex_attrs = self._get_flex_attrs(item_ids)
+
         index = 0  # Position in the materialized objects.
         while index < len(self._objects) or self._rows:
             # Are there previously-materialized objects to produce?
@@ -577,7 +584,10 @@ class Results(object):
             else:
                 while self._rows:
                     row = self._rows.pop(0)
-                    obj = self._make_model(row)
+                    if flex_attrs:
+                        obj = self._make_model(row, flex_attrs[row['id']])
+                    else:
+                        obj = self._make_model(row)
                     # If there is a slow-query predicate, ensurer that the
                     # object passes it.
                     if not self.query or self.query.match(obj):
@@ -599,20 +609,31 @@ class Results(object):
             # Objects are pre-sorted (i.e., by the database).
             return self._get_objects()
 
-    def _make_model(self, row):
-        # Get the flexible attributes for the object.
+    def _get_flex_attrs(self, ids):
+        # Get the flexible attributes for all ids.
         with self.db.transaction() as tx:
+            id_list = ', '.join(str(id) for id in ids)
             flex_rows = tx.query(
-                'SELECT * FROM {0} WHERE entity_id=?'.format(
-                    self.model_class._flex_table
-                ),
-                (row['id'],)
+                'SELECT * FROM {0} WHERE entity_id IN ({1})'.format(
+                    self.model_class._flex_table,
+                    id_list
+                )
             )
 
+        # Index flexible attributes by the entity id they belong to
+        flex_values = dict()
+        for row in flex_rows:
+            if row['entity_id'] not in flex_values:
+                flex_values[row['entity_id']] = dict()
+
+            flex_values[row['entity_id']][row['key']] = row['value']
+
+        return flex_values
+
+    def _make_model(self, row, flex_values={}):
         cols = dict(row)
         values = dict((k, v) for (k, v) in cols.items()
                       if not k[:4] == 'flex')
-        flex_values = dict((row['key'], row['value']) for row in flex_rows)
 
         # Construct the Python object
         obj = self.model_class._awaken(self.db, values, flex_values)
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -32,6 +32,7 @@ from beets.dbcore import types
 from .query import MatchQuery, NullSort, TrueQuery
 import six
 
+
 class DBAccessError(Exception):
     """The SQLite database became inaccessible.
 
@@ -567,9 +568,7 @@ class Results(object):
         first.
         """
 
-        # First fetch all flexible attributes for all items in the result.
-        # Doing the per-item filtering in python is faster than issuing
-        # one query per item to sqlite.
+        # Index flexible attributes by the item ID, so we have easier access
         flex_attrs = self._get_indexed_flex_attrs()
 
         index = 0  # Position in the materialized objects.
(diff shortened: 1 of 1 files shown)


beets/dbcore/db.py
```diff
@@ -594,20 +604,31 @@ def __iter__(self):
             # Objects are pre-sorted (i.e., by the database).
             return self._get_objects()
 
-    def _make_model(self, row):
-        # Get the flexible attributes for the object.
+    def _get_flex_attrs(self, ids):
+        # Get the flexible attributes for all ids.
```
sampsyo: Can we get a detailed docstring for this new method?

beets/dbcore/db.py
```diff
@@ -594,20 +604,31 @@ def __iter__(self):
...
         with self.db.transaction() as tx:
+            id_list = ', '.join(str(id) for id in ids)
             flex_rows = tx.query(
-                'SELECT * FROM {0} WHERE entity_id=?'.format(
-                    self.model_class._flex_table
-                ),
-                (row['id'],)
+                'SELECT * FROM {0} WHERE entity_id IN ({1})'.format(
+                    self.model_class._flex_table,
+                    id_list
```
sampsyo: Inlining values into a query generally makes one nervous, although clearly SQL injection should be impossible here because we're just inserting a list of numbers. I am a little concerned we could run into weird problems, however, if the query string gets really long: when doing a `beet ls` on an entire large, library, for example. We're now asking SQLite to parse an SQL string with possibly hundreds of thousands of numbers—we may even overflow some other limit on the length of the query.

Is there possibly some other way to pass the set of IDs into SQLite? With some more engineering, for example, we could endeavor to use a join to query for the flexible attributes at the same time as the original query.

Or, perhaps an `executemany` query would be a good middle ground?

beets/dbcore/db.py
```diff
@@ -594,20 +604,31 @@ def __iter__(self):
...
+        flex_values = dict()
+        for row in flex_rows:
+            if row['entity_id'] not in flex_values:
+                flex_values[row['entity_id']] = dict()
+
+            flex_values[row['entity_id']][row['key']] = row['value']
+
+        return flex_values
+
+    def _make_model(self, row, flex_values={}):
```
sampsyo: As above, a docstring would be nice here (even though we didn't have one before). It's also worth noting that the contract for this method is now different than it was: you need to have already fetched the flexattrs; previously the method did it for you.

beets/dbcore/db.py
pprkut: So, did a little digging.

The default value for sqlite's sql query length limit is 1,000,000 bytes. The default value for the maximum number of host query parameters (so, the sql injection save alternative) is 999.
If my calculations are correct, this should be good for around 160000 tracks in a collection. I currently have 11,000 tracks in my collection, so I'm far away from that limit. Not sure, do you think that value is high enough or should we consider larger collections?

If we go down that path we won't avoid running more than one query. But perhaps we can do it based on a count somehow (maybe 100,000 tracks since it's a nice round number), so that in 99.9% of all use cases there'd be only one query.

pprkut:
Finally had some time to rerun my performance tests under at least somewhat untainted circumstances.
I checked with `time beet ls &> /dev/null`, which for me is 11240 tracks. I have the db on a SSD, not sure that has a lot of impact, but mentioning it just to be aware.

Without the patch
```
real    0m9.766s
user    0m9.500s
sys     0m0.266s
```

With the patch
```
real    0m9.413s
user    0m9.210s
sys     0m0.203s
```

Both are fastest runs out of 5. Based on this I'd say on average it gives a 4% speedup.

beets/dbcore/db.py
pprkut: I considered `_make_model` a private method, so breaking the contract didn't seem like that big a deal. I can figure out a way to keep the current API in tact, but not sure it's worth it.

beets/dbcore/db.py
pprkut: I'll extend it for the next iteration of the patch :)

beets/dbcore/db.py
sampsyo: Good discovery! That is a pretty big number. It might be fine to just assume that never trips—one other question, though, is if using such a long query ever overwhelms the potential performance benefits for normal queries.

Overall, this gave me an idea for a somewhat larger refactoring here: maybe it would make sense to *semi-lazily* load the flexible attributes. That is, instead of loading them for every item as it's fetched from the database (fully lazy) or fetching them all up front (fully eager), maybe it makes sense to fetch them in "chunks," on demand. When you iterate through the result set, some reads will be slower than others because they will trigger the fetching of flexible attributes for the next "chunk." This could help make commands like `beet list` more responsive for very large queries while simultaneously solving the problem you set out to address here.

beets/dbcore/db.py
sampsyo: Yeah, absolutely—I agree it's no big deal to change the interface here. But just like the new method above, it would be nice to have a docstring about what it's supposed to do now so we can be aware of the difference in the future.

sampsyo:
Indeed! Thanks for doing the measurements—that's really helpful to know. And it helps point the way to the next bottleneck too (formatting the output).

pprkut:
Finally had time to do a revision on this. I moved the flex attr query to `_fetch()`, sort of like you already suggested. That way we still only query once, but we can construct a better subquery that no longer has to worry about a length limitation or SQL injection possibilities. I think this looks good now :-)

I did some further measurements, to get the actual raw performance of iterating the entire library (11832 items):

```python
start = time.time()
for item in lib.items(query):
    continue
end = time.time()
```

On master this takes `3.66s` average
Using a single flex attr query this goes down to `2.95s` average
I have another optimization potential that I'll look into next to get this further down to `1.41s`, but it'll need some more verification.

I also see some unit test failures in travis that I don't really understand (I don't get those locally), so if someone has some pointers on what to do about those I'd really appreciate it.

beets/dbcore/db.py
```diff
@@ -572,7 +577,10 @@ def _get_objects(self):
             else:
                 while self._rows:
                     row = self._rows.pop(0)
-                    obj = self._make_model(row)
+                    if flex_attrs:
+                        obj = self._make_model(row, flex_attrs[row['id']])
+                    else:
+                        obj = self._make_model(row)
```
sampsyo: This should perhaps be:

```
if row['id'] in flex_attrs:
```

or similar.

sampsyo: Or, I suppose, just a single line with no`if`/`else`:

```
obj = self._make_model(row, flex_attrs.get(row['id'], {}))
```

pprkut:
Hey! Just checking, this is good to go right? If so I'll go ahead and merge it

sampsyo:
Yes; sorry for not responding! This looks good to go. It's a small change but a very important one! I say merge away.

PR #2804, opened 2018-01-31
PlexUpdate: token redacted

Kovrinic:
Added Plex token redaction for plexupdate plugin.  For use with `beet config` sharing.


sampsyo:
Thank you! :sparkles: Merged with a changelog entry.

PR #2813, opened 2018-02-18
Use NullPaddedInt for R128 tags

autrimpo:


sampsyo:
Thank you! Yes, this seems like exactly the right fix. Would you mind adding a quick summary to the changelog?

sampsyo:
Perfect; thank you! ✨

djl:
I think this change may have caused beets to write (or rewrite) `r128_track_gain` and `r128_album_gain` tags on all of my files. This has made my usual `rsync` backup a little painful.

I'm going off memory here but I believe `beet write -p` wanted to change both fields from `0000 -> 0000` which to me looks like nothing should really have changed on the file themselves.

Has anyone else experienced this?

sampsyo:
Hmm, that's bad news—ideally, the change would only allow the tag to be missing. Perhaps the change is that beets now wants to take the "zero" tags on the files and remove them altogether?

autrimpo:
It does show `0000 -> 0000` when removing the tags, so I think they only got removed, no other changes should happen. Are you able to check what changes actually occured?

sampsyo:
No problem! For clarity, maybe it would be a good idea for the UI to display something like "none" instead of "0000" when the value is null (missing).

PR #2814, opened 2018-02-18
Early stage pipeline plugin functions

autrimpo:
This is a bug report together with a PR because I like to scratch my own itch :)
Something has been bugging me for quite some time after switching to Opus - some of the albums I've imported didn't have replaygain calculated, even though the replaygain plugin did definitely run. After lots of debugging the issue seems to be as follows:
1. User tries to import a FLAC album
2. Since the plugins don't run in a deterministic order (as far as I've noticed), ReplayGain is sometimes calculated on the FLAC files
3. Files get converted to Opus
4. There are ReplayGain values calculated from FLAC but Opus only takes R128
5. The imported files need to have `replaygain` calculated again, manually

I've added a (quite ugly) hack to always convert the file first, which from my testing seems to reliably solve the issue. If you have better idea how to solve this (I've considered dividing the plugins into groups by 'priority' but that felt like an overkill), I'll happily implement it.

### DIFF
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -81,6 +81,7 @@ class BeetsPlugin(object):
             self.template_fields = {}
         if not self.album_template_fields:
             self.album_template_fields = {}
+        self.early_import_stages = []
         self.import_stages = []
 
         self._log = log.getChild(self.name)
@@ -93,6 +94,17 @@ class BeetsPlugin(object):
         commands that should be added to beets' CLI.
         """
         return ()
+
+    def get_early_import_stages(self):
+        """Return a list of functions that should be called as importer
+        pipelines stages early in the pipeline.
+
+        The callables are wrapped versions of the functions in
+        `self.early_import_stages`. Wrapping provides some bookkeeping for the
+        plugin: specifically, the logging level is adjusted to WARNING.
+        """
+        return [self._set_log_level_and_params(logging.WARNING, import_stage)
+                for import_stage in self.early_import_stages]
 
     def get_import_stages(self):
         """Return a list of functions that should be called as importer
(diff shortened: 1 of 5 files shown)


sampsyo:
Got it! Thanks for explaining why it's nice for this to run first. I can see it being a good idea in general for `convert` to run before other operations.

It would be good to come up with a general way to prioritize import stages. We don't have to go crazy with anything too elaborate (such as a priority score), but how about just a mechanism for asking to be inserted at the beginning of the list? For example, we could add an `early_import_stages` alternative to `import_stages` that does this.

autrimpo:
So if I understand correctly, plugins would be either 'regular' or 'early' and 'early' ones would always run before 'regular' ones but other than that the order would be still non-deterministic? That's very similar to what I was trying with the priority groups, except I had three and this only has two. I'll have a go at it tomorrow.

sampsyo:
Yeah, you got it exactly. I could also see three groups, but two might suffice—I don't quite see a use case at the moment for running as late as possible.

autrimpo:
It can always be added later if the need arises.

autrimpo:
I've added functionality for early stage functions, decided it might be even better because plugins can now request to have part of their functions ran early and the rest at the normal time.

beets/plugins.py
```diff
@@ -94,6 +95,17 @@ def commands(self):
...
+    def get_early_import_stages(self):
+        """Return a list of functions that should be called as importer
+        pipelines stages early in the pipeline.
+
+        The callables are wrapped versions of the functions in
+        `self.early_import_stages`. Wrapping provides some bookkeeping for the
+        plugin: specifically, the logging level is adjusted to WARNING.
+        """
+        return [self._set_log_level_and_params(logging.WARNING, import_stage)
+                for import_stage in self.early_import_stages]
```
sampsyo: Perhaps the logging wrapper code (here and in the function below) can be factored out into a helper function?

autrimpo: Sure.

autrimpo:
Done, though I haven't written a test as I initially planned because unfortunately I don't know how.

sampsyo:
Awesome; thank you! :sparkles: :fish: :rose: Merged.

PR #2816, opened 2018-02-23
Make sure release events are selected in preferred order.

djl:
Previously, release events were selected based on the order that MusicBrainz listed them rather than the order defined in config.yaml.

Opening this as a pull request instead of pushing directly because this _may_* cause large metadata changes the next time someone runs `beet mbsync`. But it should _improve_ the metadata, making it align better with a user's preferred countries.

\* I ran this over my entire collection of around 4000 albums and it changed only eight.

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -118,8 +118,8 @@ def _preferred_release_event(release):
     """
     countries = config['match']['preferred']['countries'].as_str_seq()
 
-    for event in release.get('release-event-list', {}):
-        for country in countries:
+    for country in countries:
+        for event in release.get('release-event-list', {}):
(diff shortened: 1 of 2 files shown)


sampsyo:
Nice! This is an elegant solution to a subtle problem. Would you mind adding a changelog entry describing the problem? Then this looks ready to merge.

One tiny aside: I note that this will have slightly different performance (possibly a larger number of total loop iterations, if there are more releases than configured countries). But this is very unlikely to actually matter.

djl:
I always forget about changelog entries!

sampsyo:
Perfect. Thanks for adding that! :sparkles:

PR #2819, opened 2018-02-25
/fix keyfinder => put paths in quotationmarks

antonbaumann:
To avoid errors like this:

`keyfinder: execution failed: Command '/Applications/KeyFinder.app/Contents/MacOS/KeyFinder -f /Volumes/MUSIC/Music/Some Title With Spaces.mp3' returned non-zero exit status 1.`

You should put the paths in quotation marks.

### DIFF
--- a/beetsplug/keyfinder.py
+++ b/beetsplug/keyfinder.py
@@ -50,6 +50,9 @@ class KeyFinderPlugin(BeetsPlugin):
     def imported(self, session, task):
         self.find_key(task.items)
 
+    def set_quotation(s):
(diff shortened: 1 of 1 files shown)


sampsyo:
Hello! This doesn't do what you intend. Calls to `util.command_output` are not sent through the shell, so they're not subject to interpolation or parsing, and hence do not need escaping. With this change, we would be looking for a literal path like `'/usr/bin/cmd'`, i.e., in a top-level directory called `'` for a command called `cmd'`. (And the same for file paths.)

PR #2821, opened 2018-02-26
Add a support for supports_credentials

Konubinix:
If the web plugin is behind a credential based http server and is
accessed by another in-browser client in another domain, the
specification of CORS requires the server to indicate it supports
such credentials.

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -341,6 +341,7 @@ class WebPlugin(BeetsPlugin):
             'host': u'127.0.0.1',
             'port': 8337,
             'cors': '',
+            'cors_supports_credentials': False,
             'reverse_proxy': False,
             'include_paths': False,
         })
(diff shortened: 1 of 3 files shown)


Konubinix:
Done. Thank you for the answer.

A bit of context: I needed the feature when trying to prove that, using "progressive web app" standard, we could create an android application that is able to download musics from beet web in the cache for offline listening, without installing another native application.

You can try the POC in https://konubinix.eu/beetme.html. It needs a beetweb server running on https, with a valid certificate. And thanks to cors_supports_credentials, it also works when restricting the beetweb server access.

sampsyo:
Oh wow! This is very cool! Awesome work on that web app. Is the code for it on GitHub somewhere, out of curiosity?

Konubinix:
Sure :-). It was meant to be open sourced eventually.

You may find the code in here: https://github.com/Konubinix/beetme

Of course, as always, comments are welcome.

PR #2822, opened 2018-02-26
Fix jumping time in beets.js

waweic:
Round was used instead of floor, so the minute counter jumped after each 30 seconds

### DIFF
--- a/beetsplug/web/static/beets.js
+++ b/beetsplug/web/static/beets.js
@@ -4,7 +4,7 @@ var timeFormat = function(secs) {
         return '0:00';
     }
     secs = Math.round(secs);
-    var mins = '' + Math.round(secs / 60);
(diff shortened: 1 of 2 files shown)


waweic:
You are welcome, of course. I have never contributed to a changelog before and don't really know how to write an entry. How would you formulate that?

sampsyo:
Sure! Just add a bullet to the end of the current "Fixes" list at the top of `changelog.rst` saying, for example:

```rst
* :doc:`/plugins/web`: The time display in the web interface would incorrectly jump
  at the 30-second mark of every minute. Now, it correctly changes over at zero
  seconds. :bug:`2822`
```

waweic:
Thanks! If you don't mind, I will just take this for the changelog entry.

sampsyo:
Looks great; thank you! :sparkles:

PR #2823, opened 2018-02-26
Fix album_art() in __init__.py

waweic:
flask.send_file() expects a string, g.lib.get_album() returns bytes. Added decode() to album_art(). 

If g.lib.get_album() gets a non-existing id, it returns None. Python would throw an error in this case. Added check to prevent this.

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -285,8 +285,8 @@ def album_query(queries):
 @app.route('/album/<int:album_id>/art')
 def album_art(album_id):
     album = g.lib.get_album(album_id)
-    if album.artpath:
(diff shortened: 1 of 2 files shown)


sampsyo:
Got it—looks like this is two independent bugs.

For the additional `.decode`: this looks like a Python 3-specific change; do you know whether this still works under Python 2? That would be good to certify.

For a changelog entry, something close to the descriptions above would work well. :+1:

waweic:
Yes, they are independent in theory. I have just put them in one PR because it seemed more practical. It works both in Python 2.7 and Python 3.6. I'm going to write the Changelog entry tomorrow.

sampsyo:
Perfect. Thanks again!

PR #2824, opened 2018-02-27
Improve config.rst's introduction

jblang94:
This PR attempts to improve config.rst's introduction

### DIFF
--- a/docs/reference/config.rst
+++ b/docs/reference/config.rst
@@ -1,25 +1,23 @@
 Configuration
 =============
 
-Beets has an extensive configuration system that lets you customize nearly
-every aspect of its operation. To configure beets, you create a file called
-``config.yaml``. The location of the file depend on your platform (type ``beet
-config -p`` to see the path on your system):
+Beets has an extensive configuration system which lets you customize nearly
+every aspect of beets. To configure beets, a configuration file called
+``config.yaml`` must be created. The location of ``config.yaml`` depends on your operating system (OS).
 
-* On Unix-like OSes, write ``~/.config/beets/config.yaml``.
-* On Windows, use ``%APPDATA%\beets\config.yaml``. This is usually in a
-  directory like ``C:\Users\You\AppData\Roaming``.
-* On OS X, you can use either the Unix location or ``~/Library/Application
+* On a Unix OS, use ``~/.config/beets/config.yaml``.
+* On Windows, use ``%APPDATA%\beets\config.yaml``. ``%APPDATA%`` is usually an environment variable
+  which stores a directory like ``C:\Users\[Your Windows' username]\AppData\Roaming``.
+* On OS X, you can use either the Unix OS's location or ``~/Library/Application
   Support/beets/config.yaml``.
 
-You can launch your text editor to create or update your configuration by
-typing ``beet config -e``. (See the :ref:`config-cmd` command for details.) It
-is also possible to customize the location of the configuration file and even
-use multiple layers of configuration. See `Configuration Location`_, below.
+You can launch your text editor to create or update your ``config.yaml`` by
+executing ``beet config -e``. (See the :ref:`config-cmd` command for details.) The location of ``config.yaml`` 
+can be customized as described in `Configuration Location`_. Using beets, you can confirm 
+the ``config.yaml`` file's location by executing ``beet config -p``.
 
-The config file uses `YAML`_ syntax. You can use the full power of YAML, but
-most configuration options are simple key/value pairs. This means your config
-file will look like this::
+``config.yaml`` uses `YAML`_ syntax. Most of beets' configuration options are 
+key/value pairs which look like these examples::
 
     option: value
     another_option: foo
@@ -27,13 +25,13 @@ file will look like this::
         key: value
         foo: bar
 
-In YAML, you will need to use spaces (not tabs!) to indent some lines. If you
-have questions about more sophisticated syntax, take a look at the `YAML`_
+In YAML, you will need to use spaces (not tabs!) to indent lines. If you
+have questions about YAML's syntax, you can review the `YAML`_
 documentation.
 
 .. _YAML: http://yaml.org/
 
-The rest of this page enumerates the dizzying litany of configuration options
+The remainder of this document describes the numerous configuration options that are
 available in beets. You might also want to see an
 :ref:`example <config-example>`.
 
--- a/docs/reference/config.rst
+++ b/docs/reference/config.rst
@@ -3,18 +3,16 @@ Configuration
 
 Beets has an extensive configuration system which lets you customize nearly
 every aspect of beets. To configure beets, a configuration file called
-``config.yaml`` must be created. The location of ``config.yaml`` depends on your operating system (OS).
+``config.yaml`` must be created. The location of ``config.yaml`` depends on your platform. You can type ``beet config -p`` to see where ``config.yaml`` is on your system.
 
-* On a Unix OS, use ``~/.config/beets/config.yaml``.
+* On Unix-like OSes, use ``~/.config/beets/config.yaml``.
 * On Windows, use ``%APPDATA%\beets\config.yaml``. ``%APPDATA%`` is usually an environment variable
   which stores a directory like ``C:\Users\[Your Windows' username]\AppData\Roaming``.
 * On OS X, you can use either the Unix OS's location or ``~/Library/Application
   Support/beets/config.yaml``.
 
 You can launch your text editor to create or update your ``config.yaml`` by
-executing ``beet config -e``. (See the :ref:`config-cmd` command for details.) The location of ``config.yaml`` 
-can be customized as described in `Configuration Location`_. Using beets, you can confirm 
-the ``config.yaml`` file's location by executing ``beet config -p``.
+typing ``beet config -e`` (See the :ref:`config-cmd` command for details.). `Configuration Location`_ describes how you can customize ``config.yaml``'s location.
 
 ``config.yaml`` uses `YAML`_ syntax. Most of beets' configuration options are 
 key/value pairs which look like these examples::
@@ -31,7 +29,7 @@ documentation.
 
 .. _YAML: http://yaml.org/
 
-The remainder of this document describes the numerous configuration options that are
+The rest of this page describes the dizzying litany of configuration options that are
 available in beets. You might also want to see an
 :ref:`example <config-example>`.
 


jackwilsdon:
Is it still worth mentioning that you can use `beet config -p` to get the current configuration path somewhere?

jblang94:
No worries! I thought it made more sense to mention `beet config -p` after describing where `config.yaml` can be kept - made more sense to describe how you can _retrieve_ `config.yaml` 's location after _setting_ `config.yaml`'s location. 

If you'd like, I can change the wording to 
> "If you forget where your `config.yaml` file is, execute `beet config -p` to retrieve `config.yaml`'s location."

Would you like this wording change?

docs/reference/config.rst
```diff
@@ -1,39 +1,37 @@
...
-``config.yaml``. The location of the file depend on your platform (type ``beet
-config -p`` to see the path on your system):
-
-* On Unix-like OSes, write ``~/.config/beets/config.yaml``.
-* On Windows, use ``%APPDATA%\beets\config.yaml``. This is usually in a
-  directory like ``C:\Users\You\AppData\Roaming``.
-* On OS X, you can use either the Unix location or ``~/Library/Application
+Beets has an extensive configuration system which lets you customize nearly
+every aspect of beets. To configure beets, a configuration file called
+``config.yaml`` must be created. The location of ``config.yaml`` depends on your operating system (OS).
```
sampsyo: I actually kind of like mentioning `config -p` first because it tells the reader that they can skip the bullet points if they want. If they just care about where the file is on their system, they can just type the command—in fact, they can remember this for the future in case things change or they use a custom location. The bullet points are just for full details and most readers can just skip right over them.

I also marginally prefer “platform” to “operating system (OS).” It reads a little more naturally, I think, and avoids a jargony feeling.

docs/reference/config.rst
```diff
@@ -1,39 +1,37 @@
...
+
+* On a Unix OS, use ``~/.config/beets/config.yaml``.
```
sampsyo: I know it sounds silly, but “Unix-like OSes” is more accurate. Linux isn’t exactly Unix (even if it is a POSIX), and  Cygwin is even less of a Unix, but this advice applies there too.

docs/reference/config.rst
```diff
@@ -1,39 +1,37 @@
...
+  which stores a directory like ``C:\Users\[Your Windows' username]\AppData\Roaming``.
+* On OS X, you can use either the Unix OS's location or ``~/Library/Application
   Support/beets/config.yaml``.
 
-You can launch your text editor to create or update your configuration by
-typing ``beet config -e``. (See the :ref:`config-cmd` command for details.) It
-is also possible to customize the location of the configuration file and even
-use multiple layers of configuration. See `Configuration Location`_, below.
+You can launch your text editor to create or update your ``config.yaml`` by
+executing ``beet config -e``. (See the :ref:`config-cmd` command for details.) The location of ``config.yaml``
```
sampsyo: Any particular reason why you prefer “executing” to “typing” here? I used “typing” mostly because it sounds friendlier and less clinical. I like the effect it has of reminding me that I’m interacting with beets by typing commands at a prompt.

docs/reference/config.rst
```diff
@@ -1,39 +1,37 @@
...
+* On OS X, you can use either the Unix OS's location or ``~/Library/Application
   Support/beets/config.yaml``.
 
-You can launch your text editor to create or update your configuration by
-typing ``beet config -e``. (See the :ref:`config-cmd` command for details.) It
-is also possible to customize the location of the configuration file and even
-use multiple layers of configuration. See `Configuration Location`_, below.
+You can launch your text editor to create or update your ``config.yaml`` by
+executing ``beet config -e``. (See the :ref:`config-cmd` command for details.) The location of ``config.yaml`` 
+can be customized as described in `Configuration Location`_. Using beets, you can confirm
```
sampsyo: As a rule, I like to avoid the passive voice in the documentation wherever possible.

docs/reference/config.rst
```diff
@@ -1,39 +1,37 @@
...
-In YAML, you will need to use spaces (not tabs!) to indent some lines. If you
-have questions about more sophisticated syntax, take a look at the `YAML`_
+In YAML, you will need to use spaces (not tabs!) to indent lines. If you
+have questions about YAML's syntax, you can review the `YAML`_
 documentation.
 
 .. _YAML: http://yaml.org/
 
-The rest of this page enumerates the dizzying litany of configuration options
+The remainder of this document describes the numerous configuration options that are
```
sampsyo: Here’s another case where I prefer a friendlier, more whimsical diction to the technically more accurate version. While “rest” and “remainder” are synonyms, the former is less formal. Same with “page” vs. “document.” And finally, I like the flair of “dizzying litany” over the clinical “numerous”—beets really does have an overwhelming number of configuration options, and I like to poke fun at how nerdy and detailed this can seem at times.

docs/reference/config.rst
jblang94: I see! Thanks for clarifying

docs/reference/config.rst
jblang94: I prefer "executing" over "typing" because just "typing" a command won't actually "execute" the command. "executing" implies that the command is typed _and_ that the command performs its specific function. Are you still set on using "typing"?

docs/reference/config.rst
jblang94: Change to 
> `Configuration Location`_ describes how you can customize ``config.yaml``'s location.

?

docs/reference/config.rst
jblang94: I see, when I first read this phrase I did not know what "litany" meant and thought that it's not a commonly used term. 

When I looked up "litany" on Google, the first definition that popped up was 
> a series of petitions for use in church services or processions, usually recited by the clergy and responded to in a recurring formula by the people.

Is there an alternative that could be used other than "numerous"?

docs/reference/config.rst
jblang94: Ok I'll undo the "platform" change and move `config -p` back to its original location

docs/reference/config.rst
```diff
@@ -1,39 +1,37 @@
 Configuration
 =============
 
-Beets has an extensive configuration system that lets you customize nearly
-every aspect of its operation. To configure beets, you create a file called
-``config.yaml``. The location of the file depend on your platform (type ``beet
```
jblang94: "depend" should be "depends". I'll still fix this when I move ``config -p`` back here

docs/reference/config.rst
jblang94: I see. I'll undo the "platform" change and move config -p back to its original location.

docs/reference/config.rst
jblang94: "depend" should be "depends". I'll still fix this when I move config -p back here

docs/reference/config.rst
sampsyo: How about [definition #2 from Wiktionary](https://en.wiktionary.org/wiki/litany): “a prolonged or tedious list”? :smiley:

docs/reference/config.rst
sampsyo: Sure; something like that sounds reasonable!

docs/reference/config.rst
sampsyo: Sure. But to make the action take place as a human, you have to type it. :smiley: I guess I’m technically relying on some shared cultural context with the user here, where typing something means typing it at a prompt, which is how you get work done.

docs/reference/config.rst
jblang94: Interesting! But I still don't think that "litany" is the right word to use. "tedious" gives the connotation that a list is too long and boring. I still don't think that "litany" is the right word to use. "selection" is more direct.

sampsyo: It is direct, but I really do mean to use self-deprecating humor here—so in my mind, a “tedious” connotation, or at least “prolonged” (implying “too long”), is perfect!

docs/reference/config.rst
jblang94: I still think that "execute" is a better choice but I'll switch the phrasing back to "typing"

docs/reference/config.rst
jblang94: I see, I'll change it back to "dizzying litany" to give the document the voice you'd like it to have

jblang94:
@sampsyo thanks for your feedback, are there any other changes to be made? I applied the changes we discussed yesterday.

docs/reference/config.rst
```diff
@@ -1,39 +1,35 @@
...
-config -p`` to see the path on your system):
-
-* On Unix-like OSes, write ``~/.config/beets/config.yaml``.
-* On Windows, use ``%APPDATA%\beets\config.yaml``. This is usually in a
-  directory like ``C:\Users\You\AppData\Roaming``.
-* On OS X, you can use either the Unix location or ``~/Library/Application
+Beets has an extensive configuration system which lets you customize nearly
```
sampsyo: which -> that

docs/reference/config.rst
```diff
@@ -1,39 +1,35 @@
...
+every aspect of beets. To configure beets, a configuration file called
```
sampsyo: “Beets [...] beets” sounds redundant. Can we stick with the original final phrase?

docs/reference/config.rst
```diff
@@ -1,39 +1,35 @@
...
+``config.yaml`` must be created. The location of ``config.yaml`` depends on your platform. You can type ``beet config -p`` to see where ``config.yaml`` is on your system.
```
sampsyo: Passive voice.

docs/reference/config.rst
```diff
@@ -1,39 +1,35 @@
...
+  which stores a directory like ``C:\Users\[Your Windows' username]\AppData\Roaming``.
+* On OS X, you can use either the Unix OS's location or ``~/Library/Application
   Support/beets/config.yaml``.
 
-You can launch your text editor to create or update your configuration by
-typing ``beet config -e``. (See the :ref:`config-cmd` command for details.) It
-is also possible to customize the location of the configuration file and even
-use multiple layers of configuration. See `Configuration Location`_, below.
+You can launch your text editor to create or update your ``config.yaml`` by
+typing ``beet config -e`` (See the :ref:`config-cmd` command for details.). `Configuration Location`_ describes how you can customize ``config.yaml``'s location.
```
sampsyo: Usage: since the text in the parentheses is a full sentence, the period should go before the parentheses.

docs/reference/config.rst
```diff
@@ -1,39 +1,35 @@
...
 
-The config file uses `YAML`_ syntax. You can use the full power of YAML, but
-most configuration options are simple key/value pairs. This means your config
-file will look like this::
+``config.yaml`` uses `YAML`_ syntax. Most of beets' configuration options are 
+key/value pairs which look like these examples::
```
sampsyo: Any particular reason to use “like these examples” instead of “like this”? It’s wordier, and I think it’s already clear that these are examples.

docs/reference/config.rst
```diff
@@ -1,39 +1,35 @@
...
     option: value
     another_option: foo
     bigger_option:
         key: value
         foo: bar
 
-In YAML, you will need to use spaces (not tabs!) to indent some lines. If you
-have questions about more sophisticated syntax, take a look at the `YAML`_
+In YAML, you will need to use spaces (not tabs!) to indent lines. If you
+have questions about YAML's syntax, you can review the `YAML`_
```
sampsyo: As with other instances, I prefer a less-formal tone: “you can review” sounds a bit colder.

docs/reference/config.rst
```diff
@@ -1,39 +1,35 @@
...
 documentation.
 
 .. _YAML: http://yaml.org/
 
-The rest of this page enumerates the dizzying litany of configuration options
+The rest of this page describes the dizzying litany of configuration options that are
```
sampsyo: Maybe you would have predicted this, but I like “enumerates”—it’s a more descriptive choice here. Also, any particular reason why you want to add “that are”?

PR #2827, opened 2018-03-01
Fix unicode bug in web plugin

waweic:
Added Exception and unidecode to web plugin. Added Description to Changelog

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -225,12 +225,19 @@ def item_file(item_id):
         item_path = util.syspath(item.path)
     else:
         item_path = util.py3_path(item.path)
-    try:
-        os.path.basename(util.py3_path(item.path)).encode("latin-1", "strict") #Imitate http.server behaviour
-        response = flask.send_file(item_path,as_attachment=True)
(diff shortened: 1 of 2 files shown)


waweic:
Ok... Doesn't seem to work in Python 2.7, I will fix this later

waweic:
Now fixed and tested. Seems to work on both Python 2.7 and Python 3

beetsplug/web/__init__.py
```diff
@@ -225,10 +226,22 @@ def item_file(item_id):
...
 
+    try:
+        unicode_item_path = util.text_string(item.path)
+    except (UnicodeDecodeError, UnicodeEncodeError):
+        unicode_item_path = u"fallback" + os.path.splitext(item_path)[1]
+
+    try:
+        # Imitate http.server behaviour
+        os.path.basename(unicode_item_path).encode("latin-1", "strict")
+        safe_filename = os.path.basename(unicode_item_path)
```
sampsyo: This line should probably go in an `else` case for the try/catch block, which runs when an exception does not occur.

sampsyo:
Looks perfect. Thank you!

PR #2838, opened 2018-03-11
beet version: include platform info

jhermann:
This supports issue reporting, showing **all** the information that is asked for (thus the issue template got a hint, too).

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -19,10 +19,12 @@ Here's a link to the music files that trigger the bug (if relevant):
 
 ### Setup
 
-* OS: 
-* Python version: 
-* beets version: 
-* Turning off plugins made problem go away (yes/no): 
+Note that calling `beet version` shows this information (platform only in newer versions).
+
+* beets version:
+* Python version:
+* Platform:
+* Turning off plugins made problem go away (yes/no):
 
 My configuration (output of `beet config`) is:
 
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1324,8 +1324,21 @@ default_commands.append(stats_cmd)
 # version: Show current beets version.
 
 def show_version(lib, opts, args):
+    # Use the improved 'distro' package for platform info, if available
+    # (the stdlib platform module has simplistic logic, and 3.7 will drop it)
+    from platform import platform
+    os_id = platform()
+    if os.name == 'posix':
+        try:
+            import distro
+        except ImportError:
+            pass
+        else:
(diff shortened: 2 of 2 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -19,10 +19,12 @@ Here's a link to the music files that trigger the bug (if relevant):
 
 ### Setup
 
-* OS: 
-* Python version: 
-* beets version: 
-* Turning off plugins made problem go away (yes/no): 
+Note that calling `beet version` shows this information (platform only in newer versions).
```
sampsyo: If that’s the case, maybe this should just say “paste the output of `beet version` here”?

beets/ui/commands.py
```diff
@@ -1324,8 +1324,21 @@ def stats_func(lib, opts, args):
...
+        try:
+            import distro
+        except ImportError:
+            pass
+        else:
+            os_id = distro.name(pretty=True) or platform()
+
     print_(u'beets version %s' % beets.__version__)
     print_(u'Python version {}'.format(python_version()))
+    print_(u'Platform: {}'.format(os_id))
```
sampsyo: For simplicity, maybe this can go on one line? Like “Python version X, on Y”?

beets/ui/commands.py
```diff
@@ -1324,8 +1324,21 @@ def stats_func(lib, opts, args):
 # version: Show current beets version.
 
 def show_version(lib, opts, args):
+    # Use the improved 'distro' package for platform info, if available
+    # (the stdlib platform module has simplistic logic, and 3.7 will drop it)
+    from platform import platform
+    os_id = platform()
+    if os.name == 'posix':
+        try:
+            import distro
```
sampsyo: Do you have a sense of how likely it is for `distro` to be installed? Is there some fallback way we can get some evidence about what the platform is, even if it’s not as nice as this module’s output?

sampsyo (commented):
Thanks for the suggestion! I made a few comments inline.

Any chance you could elucidate a little about how this works by giving some sample outputs, etc.?

We’ll also need a changelog entry.

beets/ui/commands.py
jhermann: * Some people parse these things, so format changes need some thinking about
 * More useful would be "beet version @ install dir", "py x.y @ install dir", and the platform – pip does `pip 9.0.1 from ~/.local/lib/python3.4/site-packages (python 3.4)`
`

Daletxt:
cool

Rockstar43 (approved):

jtpavlock:
@jhermann do you have any interest in continuing this?

Issue #2848, opened 2018-03-20
[DOC] Mention terminal emulator for windows


SuperSandro2000:
To display beets correctly on Windows you need to run it in msys (eg Git Bash) or use an terminal emulator like http://cmder.net/. I think this should be mentioned in the Getting Started for Windows because otherwise beets is unusable.
Native:
![](https://i.imgur.com/NfjJyDH.png)
Cmder:
![](https://i.imgur.com/fr24tfO.png)

sampsyo:
See also the root cause: https://github.com/beetbox/beets/issues/2607

But I don't object to mentioning cmder in the docs. A pull request would be welcome.

07416:
@SuperSandro2000 Please don't use third-party hosts, both images deleted.

SuperSandro2000:
@07416 probably deleted them by accident. Short: Cmder is better.

Issue #2607, opened 2017-06-21
Beets output is displaying wrong on Windows 10 command-line


acastaner:
### Problem

When running beets in a Windows 10 command-line terminal (cmd.exe) the output is wrong. It seems like there's a character encoding error. This is what it looks like (look at the last line where beets prompts for Skipping, Replace, etc) : https://imgur.com/a/b9K9x

```sh
E:\Music\_trier>beet -vv
user configuration: C:\Users\Arnaud\AppData\Roaming\beets\config.yaml
data directory: C:\Users\Arnaud\AppData\Roaming\beets
plugin paths:
Sending event: pluginload
inline: adding item field initial
inline: adding item field disc_and_track
library database: C:\Users\Arnaud\AppData\Roaming\beets\library.db
library directory: E:\Music
Sending event: library_opened
Usage:
  beet COMMAND [ARGS...]
  beet help COMMAND

Options:
  --format-item=FORMAT_ITEM
                        print with custom format
  --format-album=FORMAT_ALBUM
                        print with custom format
  -l LIBRARY, --library=LIBRARY
                        library database file to use
  -d DIRECTORY, --directory=DIRECTORY
                        destination music directory
  -v, --verbose         log more details (use twice for even more)
  -c CONFIG, --config=CONFIG
                        path to configuration file
  -h, --help            show this help message and exit

Commands:
  acousticbrainz    fetch metadata from AcousticBrainz
  clearart          remove images from file metadata
  config            show or edit the user configuration
  embedart          embed image files into file metadata
  extractart        extract an image from file metadata
  fetchart          download album art
  fields            show fields available for queries and format strings
  fingerprint       generate fingerprints for items without them
  ftintitle         move featured artists to the title field
  help (?)          give detailed help on a specific sub-command
  import (imp, im)  import new music
  lastgenre         fetch genres
  list (ls)         query the library
  modify (mod)      change metadata fields
  move (mv)         move or copy items
  remove (rm)       remove matching items from the library
  replaygain        analyze for ReplayGain
  stats             show statistics about the library or a query
  submit            submit Acoustid fingerprints
  update (upd, up)  update the library
  version           output version information
  write             write tag information to files
Sending event: cli_exit

```

### Setup

* OS: Windows 10 Professional
* Python version: 3.6.1
* beets version: 1.4.5
* Turning off plugins made problem go away (yes/no): no

My configuration (output of `beet config`) is:

```yaml
E:\Music\_trier>beet config
directory: E:\Music

[...]

PR #2849, opened 2018-03-20
Mentioned #2607 in Getting started

SuperSandro2000:
Mention in the Getting Started Guide that Windows does not support ``Colorama's ANSI color replacement`` and two suggestions how to fix it.
After merge this https://github.com/beetbox/beets/issues/2848 can be closed.
Maybe this https://github.com/beetbox/beets/issues/2607 too as it likely will never be closed.


jtpavlock:
I think there are some updated solutions worth recommending in #2607

PR #2851, opened 2018-03-21
Fixes #2847

### LINKED ISSUE #2847: Opening config produces Errno 22
### Problem

### Setup

* OS: Windows 10
* beets version: 1.4.6

```yaml
{}
```

SuperSandro2000:


SuperSandro2000:
This needs test updates or code changes. Neither of them I know how to do.

SuperSandro2000:
Reverted the Unix changes.

sampsyo:
Thanks! I like the idea to avoid `exec` altogether on Windows; this should probably make things more predictable.

If we're not using `exec`, however, maybe the right thing to use here would be `subprocess`, which is less error prone than `system`. If we do that, it seems like we could use `shlex_split` on both platforms.

The AppVeyor output points toward the tests that need some additional examination:
https://ci.appveyor.com/project/beetbox/beets/build/2415/job/w43io2yinu90h50b#L122

SuperSandro2000:
Sounds reasonable. Including ``shlex_split`` should fix the test with raise error and the other one expects exec to be called so I think it needs to be updated.

SuperSandro2000:
Implemented ``subprocess``. We need to call ``cmd.exe /c "start targets"`` as python is not aware of cmd-based command. I didn't look at the tests yet.

test/test_config_command.py
```diff
@@ -126,9 +142,9 @@ def test_edit_invalid_config_file(self):
         config._materialized = False
 
         os.environ['EDITOR'] = 'myeditor'
-        with patch('os.execlp') as execlp:
+        with patch(func) as process:
```
SuperSandro2000: Todo: set func

SuperSandro2000:
I am not sure about the tests. Works but I think the behavior is different with ``os.exelp`` and ``subprocess.call``.

beets/util/__init__.py
```diff
@@ -855,7 +855,9 @@ def open_anything():
     if sys_name == 'Darwin':
         base_cmd = 'open'
     elif sys_name == 'Windows':
-        base_cmd = 'start'
+        # python is not aware of cmd commands
+        # Start cmd.exe with start as argument
+        base_cmd = 'cmd /c \"start {}\"'
```
sampsyo: Here's the last thing we need a way around: using interpolation will cause unpredictable problems here when the filename contains special characters (including spaces). Do you know whether it's possible to get the same effect while using proper argument lists?

For example, what I would *want* to work would be:

```py
subprocess.call('start', filename)
```

but, as you realized, that won't work because `start` is a cmd shell built-in. See [this SO answer](https://stackoverflow.com/questions/9137303/subprocess-popenstart-fails), for example.

Is there any clever way we can pass the filename argument as a proper shell argument without shell interpolation? As a last resort, we may need to escape the filename ourselves. :confused:

SuperSandro2000: ``subprocess.call(["cmd", "/c", "start", file])``
I got this working

jtpavlock:
@SuperSandro2000 do you still intend to continue this?

SuperSandro2000:
@jtpavlock It was dead for 2.5 years not from my side.

jtpavlock:
Well if you're still interested in getting it merged, it would need at the very least a relevant change log entry/docs and maybe we can get @sampsyo to give it another look over.

SuperSandro2000:
@jtpavlock Replaced by #3665

PR #2862, opened 2018-04-01
Plugins doc update

architek:
Sort plugins documentation within each section. Section "other plugins" kept as is.


sampsyo:
Good call; thanks! :sparkles:

PR #2864, opened 2018-04-02
Update beets.reg

bxnyintuc:


jackwilsdon:
I don't think you meant to put this here - not everyone has beets installed under your user :P

PR #2870, opened 2018-04-07
Artist credit config

### LINKED ISSUE #1249: Use Artist Credits for tag data, but "actual" artist data for filenaming
Currently beets will always normalise [artist credit](https://musicbrainz.org/doc/Artist_Credit) data to the current artist name. However, I want to see when playing the music when, e.g., [Orgi-E](https://musicbrainz.org/artist/345fe3da-b2cb-4ad4-a1a5-43afc903663d) was credited as [Klamfyr](https://musicbrainz.org/release/d09b3568-e9cc-4458-bcf7-0c215cca75ce), but I still like the normalisation for file tree organisation purposes. This should probably be an option though, as other people will likely want to always normalise the name (and others might want to not normalise the name in the path as well).

(Somewhat related morituri issues: thomasvs/morituri#80, thomasvs/morituri#48)

denisdefreyne:
Fixes #1249.

* [x] Write test cases that cover all the scenarios (but how?)

I based myself off the existing `original_date` and `per_disc_numbering` options.

**Questions:**

* What should we pick if there is an `artist`, no `artist_credit`, an `albumartist` and an `albumartist_credit`? I assume that `artist` is the best choice here; correct?

* How do I write a test that verifies all combinations of `artist`, `artist_credit`, `albumartist` and `albumartist_credit`?

### DIFF
--- a/beets/autotag/__init__.py
+++ b/beets/autotag/__init__.py
@@ -63,12 +63,19 @@ def apply_metadata(album_info, mapping):
     mapping from Items to TrackInfo objects.
     """
     for item, track_info in mapping.items():
-        # Album, artist, track count.
-        if track_info.artist:
-            item.artist = track_info.artist
+        # Artist or artist credit.
+        if config['artist_credit']:
+            item.artist = (track_info.artist_credit or
+                           track_info.artist or
+                           album_info.artist_credit or
+                           album_info.artist)
+            item.albumartist = (album_info.artist_credit or
+                                album_info.artist)
         else:
-            item.artist = album_info.artist
-        item.albumartist = album_info.artist
+            item.artist = (track_info.artist or album_info.artist)
+            item.albumartist = album_info.artist
+
+        # Album.
         item.album = album_info.album
 
         # Artist sort and credit names.
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -56,6 +56,7 @@ per_disc_numbering: no
 verbose: 0
 terminal_encoding:
 original_date: no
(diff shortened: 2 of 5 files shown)


sampsyo:
Nice work! This looks great already!

> What should we pick if there is an artist, no artist_credit, an albumartist and an albumartist_credit? I assume that artist is the best choice here; correct?

Yes, I believe this sounds right.

> How do I write a test that verifies all combinations of artist, artist_credit, albumartist and albumartist_credit?

Hmm… it's a little tricky, but I think the way to go is to modify `self.info` in the test method. For example, you can do `self.info.artist_credit = None` to simulate a missing value for that field. (Or perhaps your question was about how to sanely manage the 2^4=16 options for which values might be missing? Maybe the sensible thing would be to choose a few of the most salient questions and call that good.0

Would you also mind including a changelog entry?

ddfreyne:
@sampsyo Done. I added a handful of tests, and they should cover all reasonable cases.

I’d like to test this out more thoroughly on my own music library (I have backups, no worries, heh). Will let you know what the outcome is.

ddfreyne:
@sampsyo Works as intended.

Freso:
As far as I understand the code, this will just blanket replace `artist` with `artist_credit`, so `artist` used in file names will also use `artist_credit` with this setting set, correct? If so, this is not a fix for https://github.com/beetbox/beets/issues/1249

Freso:
Turns out my reading was correct, based on experimentation locally. I've reopened https://github.com/beetbox/beets/issues/1249 now.

PR #2881, opened 2018-04-15
Beets.io About Us page

dongentle2:
Adding an About us page to beets.io gives a new visitor a glance of what beets is all about in few lines . constructive criticism are welcomed .


sampsyo:
As with #2882, this is not in the right format and is not linked from anywhere else. The content also doesn’t add anything beyond what we already say about beets on the homepage. Again, I’m very curious what the motivation was here.

Issue #2887, opened 2018-04-19
discogs: Incorrect disc numbering on two-sided media such as vinyl


dbogdanov:
### Problem
There is an issue with annotating vinyl releases from Discogs that have multiple discs (sides A/B, C/D, etc.). 

For example, take [this release](https://www.discogs.com/Various-SW33/release/32654).

The playlist metadata on Discogs:
```
A - Damon Wild vs. Function - Friction
B1 - Damon Wild - Skyline
B2 - Norman - Fortnight
C - C D C Crew - Flightdance  
D1 - Chris Sattinger - Decoder Ring 
D2 - Function - CNTRL
```

Beet import (tracks for Vinyl 1 and Vinyl 2 are split incorrectly):
```
Tagging:
    Various Artists - SW33
URL:
    https://www.discogs.com/Various-SW33/release/32654
(Similarity: 76.5%) (id, source) (Discogs, 2xVinyl, 1997, US, Synewave)
Vinyl 1
 * Friction (#0)     -> Friction (#1-1)
 * Skyline (#0)      -> Skyline (#1-2)
 * Fortnight (#0)    -> Fortnight (#1-3)
 * Flightdance (#0)  -> Flightdance (#1-4)
Vinyl 2
 * Decoder Ring (#0) -> Decoder Ring (#2-1)
 * CNTRL (#0)        -> CNTRL (#2-2)
[A]pply, More candidates, Skip, Use as-is, as Tracks, Group albums,
Enter search, enter Id, aBort, plaY?
```

The output of beet info after importing:
```
beet info SW33 | grep -E "disc:|track:|tracktotal:|title:|artist:"
     albumartist: Various Artists
          artist: Damon Wild vs. Function
            disc: 1
           title: Friction
           track: 1
      tracktotal: 2
     albumartist: Various Artists
          artist: Damon Wild
            disc: 1
           title: Skyline
           track: 2
      tracktotal: 2
     albumartist: Various Artists
          artist: Norman
            disc: 1
           title: Fortnight
           track: 3
      tracktotal: 2
     albumartist: Various Artists
          artist: C D C Crew
            disc: 1
           title: Flightdance
           track: 4
      tracktotal: 2
     albumartist: Various Artists
          artist: Chris Sattinger
            disc: 2
           title: Decoder Ring
           track: 1
      tracktotal: 2
     albumartist: Various Artists
          artist: Function
            disc: 2
           title: CNTRL
           track: 2
      tracktotal: 2
```

[...]

Issue #2336, opened 2016-12-22
discogs: Track IDs not fetched


welch:
### Problem

While importing tracks correctly identified as 
https://www.discogs.com/The-Larry-Goldings-Trio-Moonbird/release/8913670
the import appears to succeed, yet track id's are not stored.
`beet info moonbird` gives this same set of id's for each track:

```
    mb_albumartistid: 354487
    mb_albumid: 8913670
    mb_artistid: 354487
    mb_releasegroupid:
    mb_trackid:
```
those are discogs id's, and no track has an mb_trackid, although the tracks
are given their proper titles.

This came to my attention because the `duplicates` command sees all these
tracks as duplicates of each other. I've done the drill of `beet rm moonbird`
and reimporting their directory, behavior is consistent.

### Setup

* OS: ubuntu 16.04
* Python version: 2.7.12
* beets version: 1.4.2
* Turning off plugins made problem go away (yes/no): uuuuuuhhhhh

My configuration (output of `beet config`) is:

```yaml

format_item: $artist - $album - $title .$format-$bitrate
duplicates:
    tiebreak:
        items: [added, bitrate, country]
        albums: [added]
    format: $artist - $album - $title .$format-$bitrate $path
    count: no
    full: no
    keys: []
    move: ''
    tag: ''
    path: no
    copy: ''
    album: no
    strict: no
    checksum: ''
    merge: no
    delete: no
library: /music/beets/musiclibrary.blb

ui:
    color: yes
    colors:
        text_success: green
        text_warning: purple
        text_error: red
        text_highlight: red
        text_highlight_minor: lightgray
        action_default: turquoise
        action: blue

plugins: acousticbrainz convert duplicates edit fetchart fromfilename info lastgenre play discogs
directory: /archive/music/

import:
    move: yes
    duplicate_action: remove
    log: /music/beets/log/flacs-2016-12-09.log
    ignore_hidden: no

match:
    strong_rec_thresh: 0.15
edit:
    itemfields: track title artist album
    albumfields: album albumartist
    ignore_fields: id path
acousticbrainz:
    auto: yes
discogs:
    tokenfile: discogs_token.json
    apikey: REDACTED
    apisecret: REDACTED
    source_weight: 0.5
lastgenre:
    count: 1
    source: album
    prefer_specific: no
    force: yes
    min_weight: 10
    auto: yes
    whitelist: yes
    separator: ', '
    fallback:
    canonical: no
play:
    warning_threshold: -2
    use_folders: no
[...]

PR #2891, opened 2018-04-27
Some fixes and updates for the Discogs plugin

dbogdanov:
Fix incorrect split of a tracklist by medium for the case of
two-sided mediums (#2887).

Following the discussion in #2887, the 'medium_total' value should
contain the number of tracks on the medium to which each particular
track belongs, not the total number of different mediums present on
a release.

Fix unit tests accordingly.

### DIFF
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -271,7 +271,8 @@ class DiscogsPlugin(BeetsPlugin):
         for track in tracks:
             track.media = media
             track.medium_total = mediums.count(track.medium)
-            # Invent our own track IDs as Discogs doesn't have them
+            # Discogs does not have track IDs. Invent our own IDs as proposed
+            # in #2336.
             track.track_id = str(album_id) + "-" + track.track_alt
 
         # Retrieve master release id (returns None if there isn't one)
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -347,30 +347,34 @@ class DiscogsPlugin(BeetsPlugin):
             # a 2-sided medium.
             if ''.join(m) in ascii_lowercase:
                 sides_per_medium = 2
-                side_count = 1  # Force for first item, where medium == None
 
         for track in tracks:
             # Handle special case where a different medium does not indicate a
             # new disc, when there is no medium_index and the ordinal of medium
             # is not sequential. For example, I, II, III, IV, V. Assume these
             # are the track index, not the medium.
+
+            # side_count is the number of mediums or medium sides (in the case 
+            # of two-sided mediums) that were seen before
             medium_is_index = track.medium and not track.medium_index and (
                 len(track.medium) != 1 or
-                ord(track.medium) - 64 != medium_count * sides_per_medium + side_count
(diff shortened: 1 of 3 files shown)


beetsplug/discogs.py
```diff
@@ -270,12 +270,17 @@ def get_album_info(self, result):
         # `autotag.apply_metadata`, and set `medium_total`.
         for track in tracks:
             track.media = media
-            track.medium_total = mediums
+            track.medium_total = mediums.count(track.medium)
```
sampsyo: This looks right! It seems like [collections.Counter](https://docs.python.org/2/library/collections.html#collections.Counter) could be a good data structure too, but this is a straightforward solution as well.

beetsplug/discogs.py
```diff
@@ -270,12 +270,17 @@ def get_album_info(self, result):
...
+            # Invent our own track IDs as Discogs doesn't have them
+            track.track_id = str(album_id) + "-" + track.track_alt
```
sampsyo: As proposed in https://github.com/beetbox/beets/issues/2336#issuecomment-384488651

beetsplug/discogs.py
```diff
@@ -351,7 +356,7 @@ def get_tracks(self, tracklist):
             # are the track index, not the medium.
             medium_is_index = track.medium and not track.medium_index and (
                 len(track.medium) != 1 or
-                ord(track.medium) - 64 != medium_count + 1
+                ord(track.medium) - 64 != medium_count * sides_per_medium + side_count
```
sampsyo: This looks right to me, but perhaps an additional comment (or intermediate variable name) would make the logic more immediately obvious for future maintainability.

beetsplug/discogs.py
dbogdanov: Should I add a full URL to that particular comment or only mention the issue number?

sampsyo: Just the issue number is fine.

dbogdanov:
All done

sampsyo:
Great! Although it looks like there are a couple of remaining complaints from the style checker:
https://travis-ci.org/beetbox/beets/jobs/372287861

beetsplug/discogs.py
```diff
@@ -342,30 +348,33 @@ def get_tracks(self, tracklist):
...
+                ord(track.medium) - 64 != side_count + 1
             )
 
             if not medium_is_index and medium != track.medium:
-                if side_count < (sides_per_medium - 1):
-                    # Increment side count: side changed, but not medium.
-                    side_count += 1
-                    medium = track.medium
+                side_count += 1
+                if sides_per_medium == 2:
```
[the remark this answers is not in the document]
dbogdanov: I was considering that, but if it could only be two then why not say that explicitly. It gives context and may help to understand the code.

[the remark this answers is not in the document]
sampsyo: Cool. Then maybe an `assert 1 <= sides_per_medium <= 2` would clarify the assumption?

sampsyo:
Awesome! Thank you for your help with this; merging now.

Issue #1122, opened 2014-11-29
Discogs: fetch original release year/date


ghost:
The Discogs plugin currently doesn't seem to fetch the original release year of an album, for instance [this](http://www.discogs.com/David-Bowie-The-Rise-And-Fall-Of-Ziggy-Stardust-And-The-Spiders-From-Mars/release/1557164) 1984 release of a 1972 album returns the 'year' as 1984 (which is correct) and the 'original_year' of 0000 (not correct, obviously :D)

dbogdanov:
I think the "main release" date would work well. In my experience it always coincides with the original release. No pagination necessary, and it's better then nothing.

dbogdanov:
> I think the "main release" date would work well. In my experience it always coincides with the original release. No pagination necessary, and it's better than nothing.

I can implement this in my pull request if you confirm that this is a good solution.

lbesnard:
Any suggestion as the cleanest way to re-import those albums already in the beet database and moving their path accordingly to the beet config
```
paths:                                                                                                                                                                                                                                    
      default: '%title{$genre}/%the{%title{$albumartist}}/$original_year - $album/%if{$multidisc,Disc $disc/}$track $title'
```
cheers !

sampsyo:
Hi, @lbesnard; that’s an FAQ: http://docs.beets.io/en/v1.4.7/faq.html#rename-my-files-according-to-a-new-path-format-configuration

lbesnard:
thanks a lot @sampsyo (and sorry for that)

lbesnard:
actually for reference (for someone else), the best way to re-import (which does the move automatically) for this specific issue was to do
```bash
beet import -L -a original_year:0000
```

henry-oberholtzer:
Closed as this was resolved by https://github.com/beetbox/beets/pull/2893

PR #2893, opened 2018-04-29
Fetch original_year from master releases for Discogs (#1122)

dbogdanov:
This adds an additional Discogs API call from inside get_album_info() via a very basic fetch method (it only returns year). This may be expanded in the future to fetch other fields from the master release (genres, styles, etc.) if necessary.

Assume that ``original_year`` equals to ``year`` for releases without a master release. If parsing the original year in the master release fails, the resulting value is kept as ``None``. 

I am not sure how to update tests, because the API call is inside ``get_album_info``. The functionality is so basic that possibly it is not necessary (and other API calls aren't tested as well).


sampsyo:
Looks great! This information will be useful to have.

I notice that we now have some similar-looking exception-handling code for the calls into the Discogs API. (The handlers for `DiscogsAPIError` and `CONNECTION_ERRORS` are very similar between the `Master` and `Release` fetching code, for example.) Would it make sense to refactor that code so it can be shared between the different resource retrievals?

dbogdanov:
May be, but it may also complicate the code. The code is similar for ``candidates()``, ``album_for_id()`` and ``get_master_year()`` methods. The return values, retry method and log messages are different though. If there will be more methods like this than yes.

sampsyo:
OK, cool. It's only worth pursuing if it can be done in a way that makes the code simpler, not more complex. Other than that, I think all we need is a changelog entry here.

sampsyo:
Perfect. Thanks again! :sparkles:

sampsyo:
...and thanks for all your contributions over the last week or two. I’ve invited you as an organization member, so you should now be able to commit directly. Of course, pull requests are always still an option as a way to get code reviews!

PR #2917, opened 2018-05-08
mbsync: Map tracks using release track MBIDs instead of recording MBIDs

### LINKED ISSUE #1234: mbsync: Tolerate MusicBrainz recording ID changes
mbsync uses the musicbrainz recording id to match local track information against a certain track in a musicbrainz release. However, there is no fallback matching defined, so once the recording id changes in musicbrainz the local track can no longer be synced unless the recording id is changed manually. But this behavior is also not very visible (I stumbled upon it by accident), so most users would not even be aware that the recording id is outdated.

jdetrey:
Hi! This fixes #1234, as it allows the `mbsync` plugin to follow recording MBID changes. It falls back to the old mechanism when items in the library do not yet have an `mb_releasetrackid` field.

### DIFF
--- a/beetsplug/mbsync.py
+++ b/beetsplug/mbsync.py
@@ -117,28 +117,35 @@ class MBSyncPlugin(BeetsPlugin):
                                album_formatted)
                 continue
 
-            # Map recording MBIDs to their information. Recordings can appear
-            # multiple times on a release, so each MBID maps to a list of
-            # TrackInfo objects.
+            # Map release track and recording MBIDs to their information.
+            # Recordings can appear multiple times on a release, so each MBID
+            # maps to a list of TrackInfo objects.
+            releasetrack_index = dict()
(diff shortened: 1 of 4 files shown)


jdetrey:
Hmm, the failure in check `py34-flake8` of the Travis CI seems to be linked to a version change in `pep8-naming` (0.6.1 was used, instead of 0.5.0 before). See #2918 for fix.

sampsyo:
Nice! This looks like the perfect fix—making it a cascade from release track ID to recording ID to disc/track number seems like exactly the right thing.

Would you mind adding a changelog entry? Then we can merge this up.

jdetrey:
Great, thanks for the feedback! I've just updated the changelog. We should be good to go, now!

sampsyo:
Perfect; thank you!

PR #2918, opened 2018-05-08
Fix coding style: `Shareable` is a metaclass.

jdetrey:
Release v0.6.1 of `pep8-naming` now detects that `Shareable` in `beets/util/artresizer.py` is a metaclass. The first argument of each class method should then named be `cls` instead of `self`.


sampsyo:
Good catch; thank you!

PR #2919, opened 2018-05-09
Add support for user and pass to the Musicbrainz Autotagger

belak:
I haven't had a chance to test this yet, but it's mostly copied from the mbcollection plugin. There are a few servers which use auth and this would allow their usage when autotagging.


sampsyo:
Thanks! I think this needs either defaults to be provided or a more careful mechanism for checking whether the options are defined.

Can you point to anywhere in the MB documentation where it talks about authenticated lookup requests for private servers?

jtpavlock:
closing as no response from author

Issue #2914, opened 2018-05-03
mbcollection: 'remove' option inadvertently removes merged releases


bjin:
### Problem

For an album 'A' which is later merged into another album 'B' on MusicBrainz, the MBID of A will be redirected to MBID of B.

If I tried to run `beet mbupdate -r` with a local collection containing album 'A', it will add album 'B' to my remote collection (redirected on the server side) first. But in the end, 'B' will inadvertently get removed (since `mbcollection` plugins believes 'B' is not newly added and should be removed).

#### Steps to reproduce:
1. Create a local database containing [this album](https://musicbrainz.org/release/d49a8839-f5f5-467d-a4ce-2d3d7e03908b).
2. run `beet mbupdate -r`

#### Expected Behavior:
[this album](https://musicbrainz.org/release/1b27bf95-17e9-48c2-82f6-087b7142b4d2) is added to remote collection instead of the one with original MBID.

#### Actual Behavior:
Nothing is added to remote collection. The expected album is added first but later get removed.

### Setup

* OS: Linux
* Python version: 3.6.5
* beets version: 1.4.6
* Turning off plugins made problem go away (yes/no): N/A

sampsyo:
Thanks for the clear description! That does sound like a real problem.

PR #2920, opened 2018-05-09
mbsync: fix updating album with invalid first track MBID

bjin:
MBID of recording could become invalid after merging. The existing
code always copies metadata from first track after updating. But for
albums with invalid track MBID that happens to be the first track,
MusicBrainz changes won't be applied to whole album, only whose
tracks with valid MBID. This is particularly annoying since those
changes are actually displayed for every `beet mbsync` run, but never
get applied.

Fix this issue by finding any track that get MusicBrainz updates, and
apply it to whole album.

Some additional comments:

1. While #2917 seems being able to solve `beet mbsync` issues in long term. It still can't handle existing collection with invalid track MBID. Since those tracks won't receive any updates (including `mb_release_trackid`), and becomes "zombie". This pull request won't solve the whole "track MBID missing" issue, just the most annoying part of it.
2. This issue actually caused #2914 (partially) for me. It stops merged album MBID from being updated during `beet mbsync` run. The issue of #2914 still exists, but with this pull request at least `beet mbsync` could become a work around.

### DIFF
--- a/beetsplug/mbsync.py
+++ b/beetsplug/mbsync.py
@@ -152,10 +152,13 @@ class MBSyncPlugin(BeetsPlugin):
             with lib.transaction():
                 autotag.apply_metadata(album_info, mapping)
                 changed = False
+                # Find any changed item to apply MusicBrainz changes to album.
+                any_changed_item = items[0]
                 for item in items:
                     item_changed = ui.show_model_changes(item)
                     changed |= item_changed
                     if item_changed:
+                        any_changed_item = item
                         apply_item_changes(lib, item, move, pretend, write)
 
                 if not changed:
@@ -165,7 +168,7 @@ class MBSyncPlugin(BeetsPlugin):
                 if not pretend:
                     # Update album structure to reflect an item in it.
                     for key in library.Album.item_keys:
-                        a[key] = items[0][key]
+                        a[key] = any_changed_item[key]
                     a.store()
 
                     # Move album art (and any inconsistent items).
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -115,6 +115,8 @@ Fixes:
   tracks. :bug:`2537`
 * In the ``mbsync`` plugin, support MusicBrainz recording ID changes, relying
   on release track IDs instead. Thanks to :user:`jdetrey`. :bug:`1234`
+* In the ``mbsync`` plugin, allow beets to update album if the first track
+  has a missing MusicBrainz recording ID.
 
 
 For developers:


jdetrey:
Hi! Thing is, with this change, if I'm not mistaken, tracks with an invalid `mb_trackid` will remain "zombie", right? It seems to me that this happens when the `candidates` list in [line 133](https://github.com/bjin/beets/blob/00c621d26c31c8da46c3ec30c5952e746d584a8d/beetsplug/mbsync.py#L133) of `mbsync.py` is empty: then `mapping` fails to map the corresponding `Item` to a `TrackInfo` object.

Note however that, when there are multiple candidates, `mbsync` falls back to identifying tracks based on medium and track numbers.

Maybe a solution when there is no available candidate for a given track could be to go through all elements of `album_info.tracks` and identify the missing track using medium and track numbers. See commit jdetrey/beets@1b929b5d0c39fd5453fb5712f45c71150b07e97b for a possible fix.

jdetrey:
By wrong mapping, do you mean that tracks have been renumbered in MusicBrainz, and that a mapping based on medium and track numbers only misses this renumbering and thus wrongly  identifies the tracks?

Wow, I didn't think this could happen: I would've thought disc and track numbers were not prone to frequent changes. But sure, adding an extra check based on `title` would definitely help avoiding such mixups (even though `mbsync` would then miss the tracks whose title has changed; but I guess that's probably the price to pay, here).

bjin:
Not just shuffling, some times tracks could also get [removed](https://musicbrainz.org/edit/51636413) or added.  MusicBrainz, after all, is created by human. Medium list could be wrong and be changed later in any way one can think of.

On the hand, leaving zombie track doesn't hurt much. It's not visible to user (after this PR merged), and won't hurt user later in another way later (unless we attempted to fix in an aggressive way and created wrong track mapping). So I would suggest the strict approach if we are going to fix the zombie tracks issue in one way or another.

jdetrey:
Yes, you're absolutely right. Leaving zombie tracks isn't that much of a problem indeed, especially since albums can still be updated (unless _all_ of their recording IDs have changed) thanks to your PR. Thanks!

sampsyo:
Sounds like a good idea to me; thanks!

Could you please add a changelog entry describing the change? Also, is there an easy way to write a test showcasing the problem?

bjin:
rebased and added changelog entry.

> Also, is there an easy way to write a test showcasing the problem?
Hmm, I don't think there is an easy way. However, the fix is actually pretty self explained. I also tested it while fixing zombie tracks from my collection. It works as expected.

PR #2927, opened 2018-05-14
Make lyrics plugin group songs by 'albumartist' rather than 'artist'

### LINKED ISSUE #2924: Lyrics plugin groups songs by 'artist' rather than 'albumartist'
### Problem

When you use the lyrics plugin to create rst files with the lyrics themselves, the resulting files are grouped by the items' `artist` field rather than `albumartist`, which would arguably be more reasonable. When an artist has many collaborations where for a song the two fields are different, it may lead to situations like this:

image

after a three line change ([commit link](https://github.com/Abrackadabra/beets/commit/62d5fe3692f0e93663da6c87818578a7ccef8d7c)):

image

Let me know if you want me to create a push request with this.

Abrackadabra:
Fixes #2924

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -772,7 +772,7 @@ class LyricsPlugin(plugins.BeetsPlugin):
         writing continuously to the same files.
         """
 
-        if item is None or slug(self.artist) != slug(item.artist):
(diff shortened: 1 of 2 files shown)


sampsyo:
This looks perfect. Thank you for the quick fix! Awesome work. :sparkles: :rocket:

PR #2936, opened 2018-05-29
Adds some Discogs interaction

rvvincelli:
Hi all,

This does not want to be a release branch - unfortunately I cannot get `tox` to work and I am more interested in showing something real to get info flowing from Discogs, automatically.

My use case: a library with well formatted filenames like `Artist - Song.mp3` (with a few other grammar alternatives), I want to automatically fill in tag info from Discogs. By setting the Discogs tagger, with different configuration, I never got it to work, so I thought a patch around and get something done.

The pipeline works as:
-pass the file through a cleaner that removes stopcharacters
-run a search with the [Discogs console](https://github.com/jesseward/discogs-cli)
-if an ID is found, `beet` it with that specific id

Find the following scripts as an implementation:
-[tagger](https://gist.github.com/rvvincelli/c04bd70fa470070ef1147c1c48402cbc) - run first
-[bucketer](https://gist.github.com/rvvincelli/139440cb1afb334ac40bab3649c4142d) - to organize the files in a `genre/style1/style2/style3` layout
-[linker](https://gist.github.com/rvvincelli/20ff0ec5d7d56e3ce235fb62f6c3d665) - to create handy symlinks from the source directory to the nested file structure

All should work from the tip of this branch but this is for the brave Discogs maniac only!

Questions are very welcome.

Happy tagging!


sampsyo:
Interesting project!

As you of course already intuited, we like to build beets as a set of small, reusable components that can be customized for any need rather than catering to any specific, integrated use case. The goal here would be that you can someday use the discogs plugin unmodified, and my combining that with a suitable paths configuration and possibly the `link` option to the importer, you can get what you want.

So I'd be curious to learn whether you think it's possible to incorporate any of these ideas into the discogs plugin as it currently stands. What do you think needs to change in that plugin for it to be a better match for your expectations?

rvvincelli:
Hi @sampsyo , first of all thank you for the efforts in beets because it is really very well made piece of software, and I can say I have found a home for all of my archiving and tagging needs :)

The discogs plugin as-is did not work for me, the main point being, if I analysed correctly, that the integration with the other plugin for feeding the filename - unfortunately I forgot how it is called - gets you a thing which is not smart enough.

As a use case, this is what one would expect from the discogs plugin: I want to be able to do a free-text search in Discogs with the result of some function on the filename. In the reasonable case this is a customized one, to be defined with sed-like rules, to cleanup the filename properly - see my tagger script example.

Also, it is important to be able to exclude the MusicBrainz source completely. IIRC this is possible with the scores in the config file but we must be sure.

All of what is shown in this attempt is integrable in the existing plugins or a couple of new ones may be created.

Thank you for the feedback and I cannot promise I am going to make it a real pullreq but if anybody wants to support I am here to help.

Thanks,

jtpavlock:
Closing as this PR seems to be abandoned and I'm not sure it's intended to be a final product ? Anyways, if @rvvincelli would like to turn this into a more concrete pull request or continue efforts, then feel free to re-open.

PR #2937, opened 2018-05-30
Update web plugin index.html

matlads:
use the 'double vertical bar' entity to represent the pause button. The current vertical bars are slightly ugly. :)

After some research I got the 'double vertical bar' thing here: http://graphemica.com/%E2%8F%B8

### DIFF
--- a/beetsplug/web/templates/index.html
+++ b/beetsplug/web/templates/index.html
@@ -20,7 +20,7 @@
 
                 <button class="disabled">&#9654;</button>
                 <button class="play">&#9654;</button>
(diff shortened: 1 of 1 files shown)


sampsyo:
Hmm… on my system (i.e., macOS), this displays as an emoji (a blue rounded rectangle). That doesn't quite seem right for the aesthetic. Is there maybe some way to force the display of a non-emoji version?

LuxAetern:
Seems to work in FF, but apparently not in all browsers...
https://stackoverflow.com/questions/32915485/how-to-prevent-unicode-characters-from-rendering-as-emoji-in-html-from-javascrip

Issue #2942, opened 2018-06-02
Field comparison problems


twrightsman:
### Problem

Generalized issue for the problem reported in https://github.com/beetbox/beets/issues/2929. `beets` has problems comparing certain fields like BPM for the commands `write` and `update`.
So far it seems that the comparison problems are with at least three fields:

1. BPM (`bpm: 96 -> 96.0649795532`)
2. bitrate
3. acoustid_fingerprint (the strings are the same but beets still reports them as different)

```
[...]

PR #2943, opened 2018-06-04
Field comparison problems

twrightsman:
Related to https://github.com/beetbox/beets/issues/2942. How do these fixes look?


twrightsman:
I'm not quite sure if `str()` is the proper way to convert from byte strings to text strings while maintaining backwards compatability with Python 2.

PR #2949, opened 2018-06-07
pre-resolve package path for lazy-load configuration

twrightsman:
The test `test_ui.ConfigTest.test_command_line_option_relative_to_working_dir` was failing on my system but, for some odd reason, not on Travis nor AppVeyor. The test fails because it changes working directories before the relative lazy-loaded default configuration path is resolved, later causing a `beets.util.confit.NotFoundError: musicbrainz.host not found` error. Unfortunately, the pkgutil loader returns `../beets/__init__.py` as the path to the default configuration source and 99% of the time this resolves because the working directory doesn't change.
I made a small change to the `LazyConfig` class to pre-resolve this path so that it can be lazy-loaded after a working directory change.

### DIFF
--- a/beets/util/confit.py
+++ b/beets/util/confit.py
@@ -794,6 +794,8 @@ class Configuration(RootView):
         super(Configuration, self).__init__([])
         self.appname = appname
         self.modname = modname
+        # Pre-resolve default source location
+        self._package_path = _package_path(appname)
 
         self._env_var = '{0}DIR'.format(self.appname.upper())
 
@@ -822,9 +824,8 @@ class Configuration(RootView):
         `modname` if it was given.
         """
         if self.modname:
-            pkg_path = _package_path(self.modname)
-            if pkg_path:
-                filename = os.path.join(pkg_path, DEFAULT_FILENAME)
+            if self._package_path:
+                filename = os.path.join(self._package_path, DEFAULT_FILENAME)
                 if os.path.isfile(filename):
                     self.add(ConfigSource(load_yaml(filename), filename, True))
 


twrightsman:
I would also argue that failing to load the default configuration should raise an error instead of being silently ignored. For example, changing the last part of `Configuration._add_default_source` from...
```
if os.path.isfile(filename):
  self.add(ConfigSource(load_yaml(filename), filename, True))
```
to
```
if os.path.isfile(filename):
  self.add(ConfigSource(load_yaml(filename), filename, True))
else:
  raise FileNotFoundError("Default configuration source could not be loaded from {}.".format(filename))
```

beets/util/confit.py
```diff
@@ -932,6 +932,15 @@ def __init__(self, appname, modname=None):
...
         self._lazy_suffix = []  # Calls to add().
+        self._package_path = _package_path(appname)  # Pre-resolve default source location
+
+    def _add_default_source(self):
+        if self.modname:
+            pkg_path = self._package_path
+            if pkg_path:
+                filename = os.path.join(pkg_path, DEFAULT_FILENAME)
+                if os.path.isfile(filename):
+                    self.add(ConfigSource(load_yaml(filename), filename, True))
```
sampsyo: Can we avoid copying most of the implementation of this method? One way, for example, would be to have the Configuration class also use a field called `_package_path`.

twrightsman:
Hi @sampsyo, do these new changes look better?

sampsyo:
Thanks for taking a look!

For simplicity, have you tried *unconditionally* running:

```py
self._package_path = _package_path(appname)
```

in `__init__` and the *unconditionally* using that field in `_add_default_source`? At first glance, it seems like moving the `_package_path` call to `__init__` would be equivalent and, at the same time, a little easier to read.

twrightsman:
You are absolutely correct, thanks for looking over my code and the suggestions! Third time's a charm?

sampsyo:
Awesome; thank you for taking care of this! :sparkles: Merged.

PR #2951, opened 2018-06-08
Add catalognumber to disambiguation string (when selecting candidates)

8h2a:
As per suggestion from: <https://discourse.beets.io/t/created-plugin-find-release-from-barcode/366/8>


sampsyo:
Looks great; thanks! I added a changelog entry and merged.

PR #2953, opened 2018-06-11
Display the artist credit when matching albums if the user prefers ar…

djl:
Display the artist credit when matching albums if the user prefers artist credits.

Opened as pull request in case this confuses new users (and so there's an issue number to put in the changelog).


sampsyo:
Looks good to me! Would you mind adding a quick changelog entry?

PR #2954, opened 2018-06-11
Translate README file in Korean.

tyg04308:
I translate README file to Korean language.


sampsyo:
Looks awesome!! (And I don't know why @codecov-io added a comment here…)

Do we have any other Korean speakers watching who can provide feedback?

simba328:
I'm Korean
This translation is good! I just want to point out a few things.
at line 31, 39, 48, 49, I recommend changing '니가' to '당신이'.
It 's just a matter of nuance.
at line 43, I think ''HTML5 Audio'를 지원하는' is better than '`HTML5 Audio`_ 를 사용하여' to make sense of the original text.

I think it is very great that the beets project gets more accessibility by providing Korean language README :)

tyg04308:
Thank you for your feedback!! I'm fix it now

sampsyo:
Fantastic! Thanks to both of you. :sparkles:

PR #2962, opened 2018-06-19
Add space-padding to missing tracks dialog

jams2:
beet

Hi all,
during a long session organising my library with Beets, I found readability was not great after a while, this might help.
Not sure how to write tests for this, happy to be pointed in the right direction, and happy to look at other dialogs if this is helpful.

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -379,14 +379,14 @@ def show_change(cur_artist, cur_album, match):
                len(match.extra_tracks) / len(match.info.tracks)
                ))
     for track_info in match.extra_tracks:
-        line = u' ! %s (#%s)' % (track_info.title, format_index(track_info))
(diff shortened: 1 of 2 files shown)


sampsyo:
This looks really great; thanks for looking into it!

Just for the sake of thoroughly considering the alternatives, how about this idea: instead of moving over the right-hand column by a fixed amount, we *could* move it over by the length of the longest track name. This is sort of like how we already lay out the “arrow” (`->`) in the metadata change display. This could make things more compact—although I’m not sure that’s better or worse than what you already have. I’d be interested in your thoughts, though!

jams2:
Yep, that might be a better option. I had a soft-limit of 80 characters in the back of my mind, but that's pretty arbitrary. I'll look into it.

jams2:
beet
@sampsyo made your suggested change, looks good I think!

jams2:
Happy to help out.

Issue #2979, opened 2018-07-10
beet web - ModuleNotFoundError on CORS import


rveachkc:
### Problem

Running beet web in python 3.6.6 virtual environment leads to ModuleNotFoundError when CORS is enabled due to import on [line 386 of beetsplug/web/__init__.py](https://github.com/beetbox/beets/blob/master/beetsplug/web/__init__.py#L386):

```python
from flask.ext.cors import CORS
```

It works if I change this line to:

```python
from flask_cors import CORS
```

Led to this problem:

```sh
(venv) user@host~/beets> beet -vv web
user configuration: /home/user/.config/beets/config.yaml
data directory: /home/user/.config/beets
plugin paths:
Sending event: pluginload
library database: /home/user/.config/beets/musiclibrary.db
library directory: /video/music_library/beets
Sending event: library_opened
web: Enabling CORS with origin: <hostname>
Traceback (most recent call last):
  File "/home/user/beets/venv/bin/beet", line 11, in <module>
    load_entry_point('beets==1.4.7', 'console_scripts', 'beet')()
  File "/home/user/beets/venv/lib/python3.6/site-packages/beets/ui/__init__.py", line 1256, in main
    _raw_main(args)
  File "/home/user/beets/venv/lib/python3.6/site-packages/beets/ui/__init__.py", line 1243, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "/home/user/beets/venv/lib/python3.6/site-packages/beetsplug/web/__init__.py", line 386, in func
    from flask.ext.cors import CORS
ModuleNotFoundError: No module named 'flask.ext'
```

### Setup

* OS: Ubuntu 16.04
* Python version: 3.6.6 virtual environment
* beets version: 1.4.7
* Turning off plugins made problem go away (yes/no): no

```
Package        Version
-------------- -------
beets          1.4.7
click          6.7
Flask          1.0.2
Flask-Cors     3.0.6
itsdangerous   0.24
jellyfish      0.6.1
Jinja2         2.10
MarkupSafe     1.0
munkres        1.0.12
musicbrainzngs 0.6
mutagen        1.40.0
pip            10.0.1
PyYAML         3.13
setuptools     39.0.1
six            1.11.0
Unidecode      1.0.22
Werkzeug       0.14.1
```

My configuration (output of `beet config`) is:

```yaml
plugins: web smartplaylist web
directory: /video/music_library/beets/
library: /home/user/.config/beets/musiclibrary.db
import:
[...]

PR #2980, opened 2018-07-10
changed flask cors import

rveachkc:
Per the [Flask Documentation](http://flask.pocoo.org/docs/1.0/upgrading/#extension-imports):

> ## Extension imports
> 
> Extension imports of the form flask.ext.foo are deprecated, you should use flask_foo.
> 
> The old form still works, but Flask will issue a flask.exthook.ExtDeprecationWarning for each extension you import the old way. We also provide a migration utility called flask-ext-migrate that is supposed to automatically rewrite your imports for this.

This changes the CORS import from the deprecated form to the preferred form.

Issue #2979


sampsyo:
Awesome; thank you! We should indeed be using the new package name. Nice work! :sparkles:

PR #2988, opened 2018-07-20
Add fallback for item access to album's attributes

### LINKED ISSUE #2797: Path config query does not work with album-level flexible attributes
### Problem

Path queries do not work with attributes that are set at album level. More specifically, I import an album setting at the same time a flexible attribute. I have various paths that are supposed to be modified by this attribute. Though the attribute is set at import time, the query at the path will not match, neither at import time, nor at a later time.

My configuration:
```

item_fields:
    initial: (albumartist_sort or artist_sort or albumartist or artist or '_')[0].upper()
    disc_count:  u'%02i.' % (disc)
    definite_year: u'%04i' % original_year or original_date[:3] or year or date[:3] or 0

album_fields:
    cond_release_category: |
            if albumstatus.lower()=='official':
                    if albumtype.lower() in ['album', 'compilation', 'ep']:
                            return None
                    elif supergenre=='Soundtrack' and albumtype.lower() in ['soundtrack']:
                            return None
                    else:
                            return albumtype.capitalize()
            else:
                    return albumstatus.capitalize()
    boxset_folder: |
            if boxset==1:
                    boxdisc_folder=u'%i. %s' % disc, disctitle
                    if boxdisc_year: boxdisc_folder=u'%s %s' % boxdisc_folder, boxdisc_year
                    return boxdisc_folder
            else:
                    return None

paths:
    default: $supergenre/%asciify{$initial}/$albumartist/%if{$cond_release_category,$cond_release_category/}[$definite_year] $album%aunique{albumartist album,albumdisambig label catalognum year}/$disc_and_track - $title
    boxset:1: $supergenre/%asciify{$initial}/$albumartist/%if{$cond_release_category,$cond_release_category/}[$definite_year] $album%aunique{albumartist album,albumdisambig label catalognum year}/$disc_count $disctitle%ifdef{boxdisc_year, ($boxdisc_year)}/$track - $title
    comp: $supergenre/Compilations/$album%aunique{}/$disc_and_track - $title
    supergenre:Soundtrack: Soundtrack/$album/%if{$disctotal>1,$disc.}$track $title
```

And this is the result:

```user@beethost:~$ beet ls supergenre:Soundtrack
user@beethost:~$ beet ls -a supergenre:Soundtrack
...
Michael Nyman - The Piano
...
user@beethost:~$ ls -l /music/Soundtrack/N/Michael\ Nyman/
total 0
drwxrwxr-x 2 user user 0 Jan 26 08:50 [1993] The Piano```

Note that the path follows the default template and that the album attribute expands correctly in the template.

### Setup
beets version 1.4.6
Python version 2.7.13

Relevant thread at discussion board:
https://discourse.beets.io/t/path-config-and-album-level-flexible-ttributes-at-import-time/278/3

### PR DESCRIPTION
**Pull request by FichteFoll**
Allows queries (especially for pathspecs) based on an album's flexattrs
while operating on items.

Fixes #2797.

I'll leave formulating the changelog up to you.

### CODE REVIEW

File: beets/dbcore/db.py [hunk 1]
```diff
@@ -330,19 +340,10 @@ def items(self):
...
 
-    def get(self, key, default=None):
-        """Get the value for a given key or `default` if it does not
-        exist.
-        """
-        if key in self:
-            return self[key]
-        else:
-            return default
-
```
sampsyo: This is really just a matter of aesthetics, but one alternative would be to have `get` just do the necessary exception handling. That is, `__getitem__` would continue to contain all the "real" logic and include the `raise` at the end. Then, `get` would change to contain:

```py
try:
    return self[key]
except KeyError:
    return default
```

That would help keep the docstring for `get` sensible, for example.

FichteFoll:
1. Not sure what you mean by this. Formatting works exactly like it did before. However, I missed that `self.model_keys` is fetched from `model.keys(True)`, which includes the album fallback. I'll override `self.model_keys` in `FormattedItemMapping`'s `__init__`.

2. I see. So, I could keep an Item-internal reference to an album and just run `load` on that before trying to access album attributes (or `keys()`). Is `load` smart enough not to update itself when there haven't been any changes to the database since the last fetch? If it doesn't, then this change would be more involved and requires some care to actually make the caching useful. (I was thinking about storing the same album reference in multiple items for example, but I'd need to call `load` for each item individually anyway because we can't be sure the database hasn't been updated in the meantime.) Otherwise, just keeping a lazy-loaded album attribute per item individually would probably already be an improvement.

3. I'm not sure I understand the problem you're describing because, as far as I'm aware, all of this is still possible. Attribute access on the item itself is prioritized over the album fallback, for their standard fields and even for flexattrs. `_setitem` was not modified, so you can also still set an item's field or flexattr to override an album's.

beets/dbcore/db.py
FichteFoll: It might have been a bit of premature optimization, but afaik exception handling in Python isn't exactly performant (creation of exception object, traceback, caching of frames etc.) and combining the code paths was pretty trivial. I also liked that I could use this to have a short implementation in `Item.get`.

The docstring issue doesn't bother me personally because it properly describes what `get` can do. It *does* have a different function signature than `dict.get`, which we need for the `with_album` parameter anyway, but it's still compatible. I think I prefer `Optionally` over `Alternatively`, however.

sampsyo:
> Not sure what you mean by this. Formatting works exactly like it did before. However, I missed that self.model_keys is fetched from model.keys(True), which includes the album fallback (and results in the item's formatter for a field being requested rather than the album's). I'll override self.model_keys in FormattedItemMapping's __init__.

Thanks for catching that!

What I'm worried about is not a direct conflict or anything—just that we're implementing the same logic ("fallback" between item and album attributes) twice. If evaluating the expression `item.field` already looks up `field` in `item`'s album, then ideally we would not need FormattedItemMapping—the plain old FormattedMapping from dbcore would do the trick.

But as you discovered, there's subtlety about which formatter gets used. Maybe there's an elegant way to provide a merged view without the duplication, but maybe this division of responsibilities is OK.

> I see. So, I could keep an Item-internal reference to an album and just run load on that before trying to access album attributes (or keys()). Is load smart enough not to update itself when there haven't been any changes to the database since the last fetch? If it doesn't, then this change would be more involved and requires some care to actually make the caching useful. (I was thinking about storing the same album reference in multiple items for example, but I'd need to call load for each item individually anyway because we can't be sure the database hasn't been updated in the meantime.) Otherwise, just keeping a lazy-loaded album attribute per item individually would probably already be an improvement.

No, `load` always loads the latest data. (Otherwise, we'd need some mechanism on the side for tracking when the database has changed—which likely would be no faster to check than just loading from the database.)

> I'm not sure I understand the problem you're describing because, as far as I'm aware, all of this is still possible. Attribute access on the item itself is prioritized over the album fallback, for their standard fields and even for flexattrs. _setitem was not modified, so you can also still set an item's field or flexattr to override an album's.

OK, good point! I had missed that existing values on items take precedence. That means, unless I'm mistaken, that item-level fixed attributes *always* take precedence—because it's impossible to remove them. Sounds good!

FichteFoll:
```
~ λ hyperfine "beet list" -m 2
Benchmark #1: beet list
  Time (mean ± σ):      7.207 s ±  0.018 s    [User: 5.438 s, System: 0.759 s]
  Range (min … max):    7.194 s …  7.220 s

~/code/beets ∃ hyperfine "python -m beets list" -m 2
Benchmark #1: python -m beets list
  Time (mean ± σ):     17.757 s ±  0.093 s    [User: 12.581 s, System: 2.135 s]
  Range (min … max):   17.691 s … 17.823 s
```

Well, not looking so bright. It's a >100% slowdown. This'd need some smart caching, probably. I do wonder why the difference is so hight, though. I mean, the ItemFormatter needed to access the item's album before as well. Maybe `keys` is run more often than I expected?

Also, I should probably add some documentation about this change.

sampsyo:
Hmm, that is a little worrisome. Let's dig a little deeper and see if we can't mitigate some of the effects. (Thanks for the tip about hyperfine, btw!)

FichteFoll:
(I just found out about hyperfine today as I browsed the fd Readme by the same author.)

This is probably the point where I would start to look into profiling as I'm still not too familiar with the code base and believe this would provide a good starting point. Have you ever done this in python and have some recommendations for tools or other tips? (I haven't.)

sampsyo:
I think that `cProfile`, in the standard library, is still probably the best profiler out there. One tip I do have, however, is that [SnakeViz](https://jiffyclub.github.io/snakeviz/) is a really nice browser-based GUI for viewing/navigating profile data.

FichteFoll:
Took a look back at this. I used [py-spy](https://github.com/benfred/py-spy) for some quick effort-less profiling and it was quite obvious that the majority of the time is being spent with database access in `get_album` (or rather the `album` property, as I changed it). Just uncommenting the `album.load()` code removes the entire performance impact, but it also means the albums we're trying to print could be outdated.

I considered the simplest solution forward to be what I suggested earlier:

1. Make the item cache its album field and provide access through a property. The album returned by this property is *read only* since it is, well, cached. I decided against preventive measures here and instead made the property "hidden" with an underscore and provided documentation.
2. Only load database model objects when they have changed by tracking a revision number that I added to the database and increase on each mutating transaction. Had to tweak this for a little while until it passed all tests, but I suppose this is fairly safe going forward now. I added a comment clarifying on the possibility of race conditions, but as long as the `_db_lock` is aquired, we are fine.

Let me know what you think.

Also, I wasn't sure if I should add a section regarding the API to the changelog. It would mention the fallback of item access on Item and that re-loading is now lazy, although the latter should be transparent.

FichteFoll:
I see. Yes, that would probably be useful for the aunique feature and might even warrant a proper implementation (i.e. a "public API"), but I'd rather not do this in here.

Regarding the changelog, I can do that. I deliberately made `_cached_album` an internal property (with the underscore) because I wasn't confident in exposing it. It is, after all, kind of a workaround, although it's the  best I could think of. But as long as it's private, it can be changed.

By the way, as you can see from the commit that references this PR, I started using this in production and haven't encountered any issues so far. I probably could have done that before as well, as I wasn't too concerned about the performance, but for this PR it was a must.

sampsyo:
OK, great. Since it’s a sensitive change, it might be wise to put out a call for testing so folks can try it out with funky configurations. I’ll post something to Discourse.

sampsyo:
Post added: https://discourse.beets.io/t/call-for-testers-better-queries-for-album-level-fields-a-performance-improvement/477

sampsyo:
(Thanks. I restarted that Travis job and everything’s fine.)

FichteFoll:
Any updates on this? Doesn't seem like the discourse thread attracted much attention.

FichteFoll:
Rebased to fix the merge conflict on the changelog. AppVeyor has some errors in the setup phase with chocolatey.

FichteFoll:
Still nobody using this, it seems. :disappointed:

Let me know when you intend to merge this, so I only need to fix the changelog conflict once (or you do it :man_shrugging: ).

FichteFoll:
Someone was asking for this a few days ago on IRC, but I missed them and couldn't point towards this PR.

Anyway, I've been using this branch for half a year now with exactly 0 issues so far. I don't use the entire feature set of beets, but importing and path styles based on album flexattrs, which is my primary use case, are just fine.

I'll try to remember making a new speed comparison since my library grew a bit over time, but I don't expect it to be much different compared to the last time.

kergoth:
FYI, I ran into a number of issues with this, mostly relating to types in the fallback, both in path format queries and in `beet ls`. See https://discourse.beets.io/t/ranges-not-working-in-beet-ls-with-album-fields-in-item-track-context/

arcresu:
I wasn't aware of this when I threw together the diff on the discourse thread @kergoth mentioned. I'll just reproduce it here:
```diff
diff --git a/beets/library.py b/beets/library.py
index 16db1e97..71b6db22 100644
--- a/beets/library.py
+++ b/beets/library.py
@@ -526,7 +526,17 @@ class Item(LibModel):
 
     @classmethod
     def _getters(cls):
-        getters = plugins.item_field_getters()
+        def atoi(f, ag):
+            def ig(i):
+                a = i.get_album()
+                if a:
+                    return ag(a)
+                else:
+                    return cls._type(f).null
+            return ig
+        getters = {f: atoi(f, g)
+                   for f, g in plugins.album_field_getters().items()}
+        getters.update(plugins.item_field_getters())
         getters['singleton'] = lambda i: i.album_id is None
         getters['filesize'] = Item.try_filesize  # In bytes.
         return getters
diff --git a/beets/ui/__init__.py b/beets/ui/__init__.py
index 327db6b0..c3adc72d 100644
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -1145,7 +1145,10 @@ def _setup(options, lib=None):
         plugins.send("library_opened", lib=lib)
 
     # Add types and queries defined by plugins.
-    library.Item._types.update(plugins.types(library.Item))
+    at = plugins.types(library.Album)
+    at.update(library.Item._types)
+    at.update(plugins.types(library.Item))
+    library.Item._types = at
     library.Album._types.update(plugins.types(library.Album))
     library.Item._queries.update(plugins.named_queries(library.Item))
     library.Album._queries.update(plugins.named_queries(library.Album))
```
This wasn't intended to be a final implementation, but my approach was a little bit different in that I thought the album-item relationship was something beets-specific and therefore should be reflected in `library.py` rather than dbcore. I used the existing getter mechanism. The `atoi` function takes an album-level getter and converts it into an item-level one that fetches the item's album and delegates to the original getter. Item-level properties still have precedence, as in this PR.

I did find that it was necessary to also change `Item._types` in order to get queries to work as intended since otherwise the album-level fields don't have type information when accessed on `Item`s.

Note that we recently picked up a helper for memoisation in another PR:
https://github.com/beetbox/beets/blob/909fd1eb272691d5cccaa0426db74871c770a8bb/beets/util/__init__.py#L1037-L1057

FichteFoll:
Thanks for the headsup. I suspect that the problem with ranges is related to me not updating the items' type information, as you did in your diff. I was entirely new to the code base before working on this, so I just never considered that to be relevant.

The `lazy_property` is similar so something I drafted earlier in the process but ended up scraping because of what I outlined in an earlier comment (https://github.com/beetbox/beets/pull/2988#issuecomment-406897912). The problem here is that the cached album is a snapshot of the database at whatever time it was first accessed, but the db may change during runtime and the lazy property will have no way to consider that fact.

I'll take a closer look at your getter approach when I find some time to work on this again.

(I'd like to mention that I cannot use beets without this feature anymore, so even if there is a huge update going on, I'll continue using my fork until I updated the PR for the changes.)

FichteFoll:
All right, I have an update to share.

I went over all the changes of the last months and rebased my PR onto current `master`, which implemented the mentioned `lazy_property` caching (in b9768e6c9b3591c5d512d32f8fc56e2cdd148cdd and 7043ed59474de417c240fa01664e460104b96c97). And indeed, this did improve `beet ls` quite a bit for me, as long as I wasn't accesing an album's attribute. With the method I used for this album fallback implementation, I actually get a worse performance for situations where album attributes are *not* accessed but better performance for when they are. Below is a table with my benchmark results:


commit | without album attr | with album attr
----------------|----|---
**Old `master`** (2018-11, b380a4c9a37640f93b02c3277af5e76d4e8567e9) | 9.0s | 14.8s
**Old PR** (2018-11, 29acdd676b670dba04ae0528c446a9509307e1be) | 9.5s | 12.0s
**Current `master`** (2019-06, be1daa9771d076f65789dce9dd7dbb7573b6d234) | 1.6s | 10.6s
**New PR** (2019-06, d7c9397e50a21fac7d5b659c9477ffc8c76ea397) | 6.0s | 7.1s

The testing commands were
```
time python -m beets ls --format '$artist - $album - $title' >/dev/null
time python -m beets ls --format '$artist - $album%ifdef{event, [$event]} - $title' >/dev/null
```
respectively, where `event` is an album flexattr. `hyperfine` didn't work properly for some reason.


After taking a couple brief looks at it, I'm not sure where the 1.6->6 slowdown comes from (that's 300%), but I wanted to wrap this up for today at least.

@arcresu's getter approach seems neat and more suimple, but has the downside of not being able to *not* get the album fallback, as it is used by the ipfs plugin at least. It will also need to make use of some caching mechanism because otherwise `Item.get_album()` is called on every wrapped album getter.

sampsyo:
Cool! Thank you for keeping this up to date! It does seem like there's sort of an intrinsic performance cost to querying album data by default. It's possible that a _really_ fast strategy for doing something like this will require further denormalization in the database, which could be fun but a somewhat complex change.

FichteFoll:
To be honest, the idea of my implementation was to not introduce album queries unless necessary, which would result in roughly the same performance for the non-album-attr-access case. The fact that this isn't the case currently seems to hint that there are some problems with my implementation, but I haven't figured out where yet.

Other than that, I'm still pretty confident in the change and, as I have mentioned, have been using this for almost a year now.

The only remaining question to me is whether I need to do something about carrying over type information of album attributes that don't exist for an item, because the type system isn't something I have dealt with really.

jtpavlock:
So what is outstanding that's preventing this from getting merged?

sampsyo:
Basically, the intrinsic performance cost (and secondarily the added complexity, but that's not a big deal). @FichteFoll has done a great job designing the implementation and benchmarking it, and there is a real performance cost that will affect all beets queries in all contexts that may be pretty hard to optimize away. Brilliant insights about how to do that would be welcome.

radusuciu:
Willing to test if still needed, though it seems like there are now merge conflicts..

FichteFoll:
It's primarily the changelog that becomes conflicted every now and then. I recently did a merge of `master` after my local version broke due to the Python 3.8 update that affected beets' AST usage. I'm still using the branch as my daily driver, FYI, but haven't found the time to dig into the performance cost since.

ctrueden:
This issue (#2797) also bit me while organizing my library. I have the same use case as @radusuciu: wanting to partition my directory structure based on flexible attributes set during import via the `--set` flag. For the moment I am using the same workaround with the inline plugin.

<details><summary>My `inline` and `paths` config (WIP)</summary>

```yaml
album_fields:
  topdir: |
    def value(f, otherwise):
      try: result = f()
      except: result = None
      return result if result else otherwise
    return value(lambda: category, 'Artists')
  subdir: |
    def value(f, otherwise):
      try: result = f()
      except: result = None
      return result if result else otherwise
    topdir = value(lambda: category, 'Artists')
    if topdir == 'Soundtracks':
      return value(lambda: franchise, '[Unknown]')
    return '[Various]' if comp else albumartist

item_fields:
  topdir: |
    def value(f, otherwise):
      try: result = f()
      except: result = None
      return result if result else otherwise
    return value(lambda: category, 'Artists')
  subdir: |
    def value(f, otherwise):
      try: result = f()
      except: result = None
      return result if result else otherwise
    topdir = value(lambda: category, 'Artists')
    if topdir == 'Soundtracks':
      return value(lambda: franchise, '[Unknown]')
    return artist
  disc_and_track: |
    if disctotal > 9:
      return u'%02i-%02i'% (disc, track)
    elif disctotal > 1:
      return u'%01i-%02i' % (disc, track)
    elif tracktotal > 99:
      return u'%03i' % (track)
    elif tracktotal > 9:
      return u'%02i' % (track)
    else:
      return u'%01i' % (track)

paths:
  # My album flexible attributes:
  # - avmedia: Comma-separated; e.g. Animation, TV, Video Games, Musicals, Movies
  # - nationality: e.g. German, Japanese, Korean
  # - franchise: e.g. Final Fantasy
  singleton: $topdir/%the{$subdir}/%the{$artist} - $title
  comp: $topdir/%the{$subdir}/($year) $album%aunique{}/($disc_and_track) $artist - $title
  default: $topdir/%the{$subdir}/($year) $album%aunique{}/($disc_and_track) $title
```

</details>

It works well, but there are downsides:

1. The performance of inline: anecdotally, `beet move` checks seem much slower with the workaround than without.
2. I don't see an obvious way to do varying levels of directory nesting with this approach. Every case has to fit into a `$topdir/$subdir` pattern. It's certainly doable to accommodate that, but if you wanted to have e.g. `Artists/A Band/(2020) Their Best Album` (i.e. two folders deep) in one case and `Soundtracks/Movies/My Favorite Movie/(2020) That Movie's Soundtrack` (i.e. three folders deep) in another, you'd have to set things up differently since template fields with path separators (`/` or `\`) are not interpolated and split (see [this forum thread](https://discourse.beets.io/t/returned-slashes-from-the-inline-plugin/1455) for discussion).

TL;DR: I would love to see this PR make it into 1.5.0!

@FichteFoll I'm relatively new to beets, but will try to make time to do some performance profiling in the next few days to see how it affects my own library.

ctrueden:
TL;DR I think the fix is as simple as:
```diff
diff --git a/beets/dbcore/db.py b/beets/dbcore/db.py
index 524ee0b0..b63b7ab1 100755
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -59,7 +59,7 @@ class FormattedMapping(Mapping):
     def __init__(self, model, for_path=False):
         self.for_path = for_path
         self.model = model
-        self.model_keys = model.keys(True)
+        self.model_keys = model.keys(computed=True, with_album=False)

     def __getitem__(self, key):
         if key in self.model_keys:
```
🙌 

## Explanation

Here is where the performance hit is happening:
```
  File ".../beets/beets/ui/commands.py", line 1076, in list_func
  File ".../beets/beets/ui/commands.py", line 1072, in list_items
    ui.print_(format(item, fmt))
  File ".../beets/beets/library.py", line 362, in __format__
    return self.evaluate_template(spec)
  File ".../beets/beets/dbcore/db.py", line 621, in evaluate_template
    return template.substitute(self.formatted(for_path),
  File ".../beets/beets/dbcore/db.py", line 611, in formatted
    return self._formatter(self, for_path)
  File ".../beets/beets/library.py", line 378, in __init__
    super(FormattedItemMapping, self).__init__(item, for_path)
  File ".../beets/beets/dbcore/db.py", line 62, in __init__
    self.model_keys = model.keys(True)
```

The `list_items` function iterates over the items in the query:
```python
for item in lib.items(query):
    ui.print_(format(item, fmt))
```

This invokes the library's formatter function on each item, which constructs a `FormattedItemMapping`:
```python
class FormattedItemMapping(dbcore.db.FormattedMapping):
    ...
    def __init__(self, item, for_path=False):
        super(FormattedItemMapping, self).__init__(item, for_path)
        # We treat album and item keys specially here,
        # so exclude transitive album keys from the model's keys.
        self.model_keys = item.keys(True) # <== EQUIVALENT TO item.keys(computed=True, with_album=True)
        self.item = item
```

And as you can see above, the `FormattedItemMapping` asks for the item keys including those from the album, which grabs them eagerly regardless of whether they will be needed for that template down the line—even though the comment suggests the intent was to exclude the album keys.

@FichteFoll Is the fix as simple as passing `with_album=False` here instead? It works in my tests!

## One other question

In my tests, I noticed that on master, flexible attributes _do_ fall back to album attributes successfully in the format string.

<details><summary>Example of flexible attributes falling back to albums</summary>

```
$ beet ls album:'Twin Peaks' -f '$album :: $title :: $franchise'
Twin Peaks (Music From the Limited Event Series) :: Lark :: Twin Peaks
Twin Peaks :: Twin Peaks Theme :: Twin Peaks
Twin Peaks :: Laura Palmer's Theme :: Twin Peaks
Twin Peaks :: Audrey's Dance :: Twin Peaks
...
Twin Peaks (Music From the Limited Event Series) :: Out of Sand :: Twin Peaks
Twin Peaks (Music From the Limited Event Series) :: Axolotl (Roadhouse mix) :: Twin Peaks
Twin Peaks (Limited Event Series Soundtrack) :: Threnody to the Victims of Hiroshima :: Twin Peaks
Twin Peaks (Music From the Limited Event Series) :: Sharp Dressed Man :: Twin Peaks
$ beet ls album:'Twin Peaks' -f '$album :: $title :: $franchise' | wc -l
84
$ echo 'select * from item_attributes where key="franchise" and value="Twin Peaks";' | sqlite3 ~/.config/beets/library.db
805891|49695|franchise|Twin Peaks
805895|49696|franchise|Twin Peaks
805899|49697|franchise|Twin Peaks
805945|49698|franchise|Twin Peaks
```

</details>

It seems to be only queries that don't fall back:
```
$ beet ls franchise:'Twin Peaks' | wc -l
4
```

Is that known/intended behavior? Certainly it's inconsistent. Finishing this PR would address that!

## Updated benchmarks with bug-fix

commit | without album attr | with album attr
-- | -- | --
master (769e424) | 16.19s | 41.56s
rebased PR (ctrueden/beets@0c7c586) | 26.44s | 38.16s
rebased PR + bug-fix (ctrueden/beets@b93b2ce) | 15.52s | 34.81s

sampsyo:
Wow!! That's truly amazing—excellent work digging into this! :tada: It would be really, really cool to include this.

To answer your second question: yes, that is indeed inconsistent, but it is also the intended behavior. Basically, the backstory is that we knew the templating fallback was useful, and it was straightforward to implement without too much performance impact, so we did it—but the query side of things was harder to do. But here we are?!

ctrueden:
@sampsyo Great! So would you say this latest patch series is ready for merge, then? If so, @FichteFoll, could you please force push the PR accordingly?

sampsyo:
Just to confirm, we want to merge the changes here _and_ your new patch together, right? So maybe the right thing is to open a new PR with all that together?

FichteFoll:
@ctrueden thank you a lot for looking into this. I will check out your over the weekend, most likely, and verify it in my environment as well as include your changes.

ctrueden:
@sampsyo wrote:
> we want to merge the changes here and your new patch together, right?

Right.

> So maybe the right thing is to open a new PR with all that together?

@FichteFoll can force-push the branch to update this PR. That would be cleaner IMHO than opening a new PR. :smile:

@FichteFoll wrote:
> I will check out your over the weekend

Great! :+1:

FichteFoll:
Good work, @ctrueden. This fixes the really bad performance for me as well. :tada: Glad it was only an oversight on my part and not an inherent flaw in the implementation. An alternative way would have been to skip the super contructor from `FormattedItemMapping` and reimplement it, but that didn't seem proper.

My results with 11171 entries in the database:

`master`
```
Benchmark #1: python -m  beets ls --format "$artist - $album - $title" >/dev/null
  Time (mean ± σ):      1.931 s ±  0.071 s    [User: 1.828 s, System: 0.094 s]
  Range (min … max):    1.881 s …  2.013 s    3 runs
  
Benchmark #1: python -m beets ls --format "$artist - $album%ifdef{event, [$event]} - $title" >/dev/null
  Time (mean ± σ):     12.289 s ±  0.291 s    [User: 7.841 s, System: 1.724 s]
  Range (min … max):   11.971 s … 12.542 s    3 runs
```

PR before the fix
```
Benchmark #1: python -m beets ls --format "$artist - $album - $title" >/dev/null
  Time (mean ± σ):     13.921 s ±  0.282 s    [User: 8.721 s, System: 2.090 s]
  Range (min … max):   13.624 s … 14.186 s    3 runs

Benchmark #1: python -m beets ls --format "$artist - $album%ifdef{event, [$event]} - $title" >/dev/null
  Time (mean ± σ):     13.929 s ±  0.769 s    [User: 9.457 s, System: 1.701 s]
  Range (min … max):   13.042 s … 14.416 s    3 runs
```

PR with fix
```
Benchmark #1: python -m  beets ls --format "$artist - $album - $title" >/dev/null
  Time (mean ± σ):      2.124 s ±  0.342 s    [User: 1.975 s, System: 0.132 s]
  Range (min … max):    1.908 s …  2.518 s    3 runs

Benchmark #1: python -m beets ls --format "$artist - $album%ifdef{event, [$event]} - $title" >/dev/null
  Time (mean ± σ):     14.250 s ±  0.469 s    [User: 9.232 s, System: 1.947 s]
  Range (min … max):   13.975 s … 14.792 s    3 runs

```

Still slightly slower, but much more manageable.

ctrueden:
Awesome, thanks @FichteFoll ! 👍 

@sampsyo Looks like this PR is good to go! 🏎️ 💨

ctrueden:
Anything else I can do to help move this forward?

ctrueden:
Just pinging one more time—seems a shame to let this languish now that the performance issue is resolved. I completely understand being too busy, but if there's anything the community can do to help ensure this can merge smoothly, let us know. E.g. if there is a manual testing process to help avoid regressions which others could run through on various platforms, to offload otherwise direct maintainer effort, let's write that up and do it!

FichteFoll:
Did a quick merge locally as I continue to use the PR as my main driver.

Linting fails because flake8-blind-except was [updated upstream](https://pypi.org/project/flake8-blind-except/) yesterday and introduced a new error code for `except Exception`, which is why the latest master passes. Considering how flake8 (pycodestyle) now has a check for bare `except:` statements itself, the plugin can probably be removed from the lint setup.

jackwilsdon:
@FichteFoll if you merge master in it should be resolved now :+1:

sampsyo (approved):
Huge apologies for the enormous, inexplicable delay on merging this PR. It's amazing work and didn't deserve to languish, but I'm going to hit the green button now!

My only remaining worry about this design is about concurrency: the new caching/revision-number scheme is obviously great, but because it can turn `model.load()` into a no-op, it runs the risk of missing updates from other threads and processes that would previously be visible. I can't at the moment think of a case where that kind of update is critical, however (perhaps the `web` plugin?), so I think it's time to unleash this change on the world and track down problems as they arise.

Thank you all for the longitudinal team effort. I'm seriously impressed at the clear thinking and solid engineering that went into this change. 👏

PR #2788, opened 2018-01-12
Added mstream to the docs

IrosTheBeggar:
as discussed here: https://github.com/beetbox/beets/issues/2700

PR #2991, opened 2018-07-25
Add mStream to the Docs

IrosTheBeggar:
I was going through my tickets and found one about adding mStream to the beets docs.  Figured I should finally follow up on that.  There was an old pull request that was mysteriously failing, so I decided to create a new one

https://github.com/beetbox/beets/pull/2788


sampsyo:
Awesome; thanks! This one seems to be green. :smiley: :sparkles:

Issue #2997, opened 2018-08-05
gmusic: Automatically Upload to Google Music on Import


shuaiscott:
### Feature Request

Uploading all my current music via Beets would result in a lot of duplicate entries. It'd be nice to have the functionality to automatically upload to Google Music when new items are added to the library

sampsyo:
Sounds good! Any hackers in the `gmusic` plugin are hereby invited to pitch in. 😇

shuaiscott:
Took a look at how this would've been implemented and decided to knock it out real quick 😄

jsetton:
@shuaiscott thanks for implementing this change. However, I noticed that the files uploaded during the import are missing some of the metadata updates made by other plugins such as album art images or featured artists added to title. Do you think it would be possible for the upload to happen after all the plugins interacting with the metadata are done?

shuaiscott:
Yeah, I think I noticed this too on my own auto-uploads. However, looking through the docs, it appears that plugins run all at the same time in parallel... meaning that while trying to get artwork and lyrics and etc, it will be uploading to Google Music at the same time. Might be best to just manually upload to Google after all, unless another post-processing plugin feature is added to beets.

shuaiscott:
@jsetton I just tried importing a new song with auto upload turned off. It too was missing some metadata and artwork. Maybe there's something going on with the Python library or with Google's upload service

jsetton:
@shuaiscott Related to the auto upload feature, based on my understanding, you can register the process to run after a specific [event](https://beets.readthedocs.io/en/v1.4.6/dev/plugins.html#listen-for-events). The downside is that it would be a blocking call compared to just having that process run asynchronously during the import stages.

Anyway, I have since moved away from gmusicapi and started to use a new [google-music api wrapper](https://github.com/thebigmunch/google-music) via a shell script.

shuaiscott:
Wow, I didn't realize that was built into plugins now. I'll reopen this and hopefully get this added to auto import after the whole import flow is run. 

```self.register_listener('import', self.autoimport)```

wisp3rwind:
Closing, since there's no more Google Play Music: https://github.com/beetbox/beets/issues/4089

PR #2998, opened 2018-08-05
Adding gmusic ability to upload on import

shuaiscott:
Implemented the import_stages def on gmusic and updated the docs

#2997

### DIFF
--- a/beetsplug/gmusic.py
+++ b/beetsplug/gmusic.py
@@ -74,7 +74,7 @@ class Gmusic(BeetsPlugin):
         self._log.info(u'Uploading files to Google Play Music...', files)
         self.m.upload(filepaths=files)
         self._log.info(u'Your files were successfully added to your '
-                  + 'Google Play Music library')
+                       + 'Google Play Music library')
 
     def getpaths(self, items):
         return [x.path for x in items]
--- a/beetsplug/gmusic.py
+++ b/beetsplug/gmusic.py
@@ -41,7 +41,7 @@ class Gmusic(BeetsPlugin):
         self.config.add({
             u'auto': False,
         })
-        if self.config['auto'].get(bool):
+        if self.config['auto']:
             self.import_stages = [self.autoupload]
 
     def commands(self):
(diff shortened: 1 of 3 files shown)


beetsplug/gmusic.py
```diff
@@ -38,6 +38,12 @@ def __init__(self):
         else:
             self.m.perform_oauth()
 
+        self.config.add({
+            u'auto': False,
+        })
+        if self.config['auto'].get(bool):
```
sampsyo: The `.get(bool)` bit is not actually necessary here—configuration views get coerced to Booleans automatically.

beetsplug/gmusic.py
```diff
@@ -62,6 +68,14 @@ def upload(self, lib, opts, args):
         self.m.upload(filepaths=files)
         ui.print_(u'Your files were successfully added to library')
 
+    def autoupload(self, session, task):
+        items = task.imported_items()
+        files = [x.path.decode('utf-8') for x in items]
```
sampsyo: Although I know it’s what the code above does too, this `decode` call is pretty worrisome and will probably break on some inevitable non-UTF-8 filenames. Can we please factor this out into a helper function so it’s possible to fix both at once?

beetsplug/gmusic.py
```diff
@@ -62,6 +68,14 @@ def upload(self, lib, opts, args):
...
+        ui.print_(u'Uploading your files to Google Play Music...')
+        self.m.upload(filepaths=files)
+        ui.print_(u'Your files were successfully added to your '
+                  + 'Google Play Music library')
```
sampsyo: Using `print_` during the import process is likely to interrupt some interactive prompts and such. It’s probably best to use [the plugin’s log](https://beets.readthedocs.io/en/v1.4.7/dev/plugins.html#logging) to emit this output while hiding it by default.

beetsplug/gmusic.py
```diff
@@ -57,11 +63,22 @@ def commands(self):
...
         self.m.upload(filepaths=files)
         ui.print_(u'Your files were successfully added to library')
 
+    def autoupload(self, session, task):
+        items = task.imported_items()
+        files = self.getpaths(items)
+        self._log.info(u'Uploading files to Google Play Music...', files)
+        self.m.upload(filepaths=files)
+        self._log.info(u'Your files were successfully added to your '
+                  + 'Google Play Music library')
```
sampsyo: Travis has a silly whitespace complaint about this line:
https://travis-ci.org/beetbox/beets/jobs/412848100

sampsyo:
Nice work; thank you!

PR #3002, opened 2018-08-13
Add uploader_id, uploader_name, and device_id for gmusicapi auth in gmusic plugin

thetarkus:
`gmusicapi` auth uses the MAC address to "register" in Google's service. Running `beet gmusic-upload` will fail if you use beets on multiple machines because too many MAC addresses will be registered. The fix it to be able to use the same MAC address (and `device_id` for searching) for all of the machines.

## uploader_id (optional)
Allow custom `uploader_id` to not reach maximum amount of registered devices. This field should be in the format of a MAC address.
From `gmusicapi` method docstring: [Link](https://github.com/simon-weber/gmusicapi/blob/1be1e0b461ae78ac5f1c504ab3387472e82e813e/gmusicapi/clients/musicmanager.py#L163)
```
There are hard limits on how many upload devices can be registered; refer to `Google's
docs <http://support.google.com/googleplay/bin/answer.py?hl=en&answer=1230356>`__. There
have been limits on deauthorizing devices in the past, so it's smart not to register
more devices than necessary.
```

## uploader_name (optional)
May not be necessary, but perhaps in the future.
From `gmusicapi` method docstring: [Link](https://github.com/simon-weber/gmusicapi/blob/1be1e0b461ae78ac5f1c504ab3387472e82e813e/gmusicapi/clients/musicmanager.py#L155)
```
This doesn't appear to be a part of authentication at all.
Registering with (id, name = X, Y) and logging in with
(id, name = X, Z) works, and does not change the server-stored
uploader_name.
```

## device_id (optional)
When running `beet gmusic-songs` from a different machine (with a different MAC address) sometimes you may get this error. The `device_id` is generated from the MAC address.
```
gmusicapi.exceptions.InvalidDeviceId: Invalid device_id XXXXXXXXA351.Your valid device IDs are:
* XXXXXXXXF92F
* XXXXXXXX43A5
```
You can then change `device_id` in your config to one of the valid device IDs that `gmusicapi` lists for you.

## config example
```yaml
plugins: gmusic

gmusic:
  email: musicgoogleaccount@gmail.com
  password: seekrit
  uploader_id: XX:XX:XX:XX:f7:97
  device_id: XXXXXXXX43A5
```

---

Let me know if this is acceptable and I will write something up for the gmusic docs.

### DIFF
--- a/docs/plugins/gmusic.rst
+++ b/docs/plugins/gmusic.rst
@@ -20,14 +20,7 @@ Then, you can enable the ``gmusic`` plugin in your configuration (see
 
 Usage
 -----
-
-To automatically upload all tracks to Google Play Music, add the ``auto: yes``
-parameter to your configuration file like the example below::
-
-    gmusic:
-        auto: yes
-        email: user@example.com
-        password: seekrit
+**Configuration required before use.**
 
 To upload tracks to Google Play Music, use the ``gmusic-upload`` command::
 
@@ -35,19 +28,7 @@ To upload tracks to Google Play Music, use the ``gmusic-upload`` command::
 
 If you don't include a query, the plugin will upload your entire collection.
 
-To query the songs in your collection, you will need to add your Google
-credentials to your beets configuration file. Put your Google username and
-password under a section called ``gmusic``, like so::
-
-    gmusic:
-        email: user@example.com
-        password: seekrit
-
-If you have enabled two-factor authentication in your Google account, you will
-need to set up and use an *application-specific password*. You can obtain one
-from your Google security settings page.
-
-Then, use the ``gmusic-songs`` command to list music::
+To list your music collection, use the ``gmusic-songs`` command::
 
     beet gmusic-songs [-at] [ARGS]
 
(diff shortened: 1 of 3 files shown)


sampsyo:
Looks cool! And it looks like you've thoroughly investigated what these parameters are for. Can you please add them to the documentation for the plugin? Feel free to leave ones that don't have an obvious use (i.e., uploader_name) undocumented for now.

docs/plugins/gmusic.rst
```diff
@@ -20,34 +20,15 @@ Then, you can enable the ``gmusic`` plugin in your configuration (see
...
 -----
-
-To automatically upload all tracks to Google Play Music, add the ``auto: yes``
-parameter to your configuration file like the example below::
-
-    gmusic:
-        auto: yes
-        email: user@example.com
-        password: seekrit
+**Configuration required before use.**
```
sampsyo: Hmm—I actually kind of liked having the configuration example up here, since people really will need to do this first. Any particular reason you moved it downward? Can we move it back here to stick with the logical flow of steps?

docs/plugins/gmusic.rst
```diff
@@ -59,3 +40,45 @@ example::
...
+- **uploader_id**: Unique id as a MAC address, eg ``'00:11:22:33:AA:BB'``.
+  This option should be set before the maximum number of authorized devices is
+  reached.  
+  If provided, use the same id for all future runs on this, and other, beets
+  installations as to not reach the maximum number of authorized devices.  
+  Default: device's MAC address.
+- **device_id**: Unique device ID for authorized devices.
+  This option only needs to be set if you receive an `InvalidDeviceId`
+  exception. Below the exception will be a list of valid device IDs.  
+  Default: none.
```
sampsyo: A proper list of config options is great! Let’s definitely keep this stuff here.

docs/plugins/gmusic.rst
thetarkus: You got it!

sampsyo:
Awesome. One last thing before we merge: could you please add a changelog entry to `changelog.rst`? Than we should be good to go!

thetarkus:
:call_me_hand:

sampsyo:
Nicely done.

Issue #2442, opened 2017-02-19
absubmit: Analyze music in parallel


nagisa:
### Problem

Have 8 cores capable of computation. Run beet absubmit. Wait 8 times more than strictly necessary.

### Setup

* OS: Arch
* Python version: 3.6
* beets version: 1.4.3
* Turning off plugins made problem go away (yes/no): not applicable

My configuration (output of `beet config`) is: not applicable

sampsyo:
Sounds good; thanks for the suggestion!

PR #3003, opened 2018-08-14
Parallelize absubmit

bemeurer:
Solves #2442

### DIFF
--- a/beetsplug/absubmit.py
+++ b/beetsplug/absubmit.py
@@ -26,8 +26,7 @@ import subprocess
 import tempfile
 import sys
 
-from concurrent import futures
-
+from multiprocessing.pool import ThreadPool
 from distutils.spawn import find_executable
 import requests
 
@@ -107,10 +106,15 @@ class AcousticBrainzSubmitPlugin(plugins.BeetsPlugin):
     def command(self, lib, opts, args):
         # Get items from arguments
         items = lib.items(ui.decargs(args))
-
-        with futures.ThreadPoolExecutor() as executor:
+        if sys.version_info[0] < 3:
             for item in items:
-                executor.submit(self.analyze_submit, (item))
+                self.analyze_submit(item)
+        else:
+            # Create threadpool
+            pool = ThreadPool()
+            pool.map(self.analyze_submit, items)
+            pool.close()
+            pool.join()
 
     def analyze_submit(self, item):
         analysis = self._get_analysis(item)
--- a/beetsplug/absubmit.py
+++ b/beetsplug/absubmit.py
@@ -25,7 +25,7 @@ import os
 import subprocess
 import tempfile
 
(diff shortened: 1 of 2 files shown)


sampsyo:
Looks like just what the doctor ordered! :sparkles: Would you mind adding a changelog entry?

bemeurer:
@sampsyo Almost! Trying to fix this right now
```Python
  File "./beet", line 23, in <module>
    beets.ui.main()
  File "/home/bemeurer/src/beets/beets/ui/__init__.py", line 1256, in main
    _raw_main(args)
  File "/home/bemeurer/src/beets/beets/ui/__init__.py", line 1243, in _raw_main
  File "/home/bemeurer/src/beets/beetsplug/absubmit.py", line 110, in command
    pool.map(self.analyze_submit, items)
  File "/usr/lib64/python3.6/multiprocessing/pool.py", line 266, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/usr/lib64/python3.6/multiprocessing/pool.py", line 644, in get
    raise self._value
  File "/usr/lib64/python3.6/multiprocessing/pool.py", line 424, in _handle_tasks
    put(task)
  File "/usr/lib64/python3.6/multiprocessing/connection.py", line 206, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
  File "/usr/lib64/python3.6/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
TypeError: can't pickle _thread._local objects
```

Any ideas?

sampsyo:
Hmm, not exactly. Maybe the problem is that multiprocessing is trying to pickle `self` to send it across the channel there, which could be causing problems? Maybe it’s possible to use `pdb` to find out what `obj` is here.

bemeurer:
Good idea, on it!

bemeurer:
@sampsyo Tested and ready for merging!

sampsyo:
Hmm; threads do seem like the right thing here, but I’m slightly concerned that `ThreadPool` isn’t documented in the stdlib. Does this work on Python 2? If not, maybe we’d be better off with [the thread pool in concurrent.futures](
https://docs.python.org/dev/library/concurrent.futures.html#threadpoolexecutor)?

bemeurer:
Oh, I don't have Python2 on my machine. Let me install it and test.

bemeurer:
It explodes with Python2

bemeurer:
@sampsyo I think I can make it works in parallel for both Python 2 and Python 3, but with slightly different techniques. I am giving it a shot right now :)

bemeurer:
@sampsyo I couldn't get it to work in Python 2, so I reverted to the nice original multithreding code for Python 3, and made it fallback to sequential on Python 2.

sampsyo:
Looks fantastic; thank you for your help!

PR #3004, opened 2018-08-15
gmusic plugin fixes and additions

thetarkus:
Currently, if the `gmusic` plugin is used, any command ran with beets will run the `gmusicapi` oauth authentication even if it has nothing related to `gmusic`.

**Example**, using command "ls" triggers the auth (every command does, including "web"):
bad gmusic

**Fixed example**, "ls" command does not trigger auth, but using one of the gmusic commands does:
bad gmusic


sampsyo:
Fantastic; thanks!

PR #3005, opened 2018-08-15
Improve error messages in case of sqlite errors

vvvrrooomm:
when beets is started and the database file is on e.g. a volume that is full beets quits with the message "database file could not be opened {path to file}". The original error message coming from sqlite is lost. the proposed patch attaches the original error as 'reason'. in the case of disk full it reads now "database file {path to file} could not be opened, reason: database or disk is full"

I have no idea how to create a test for it, because the test case either depends on a not-writeable database file or on a mock of sqlite

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -1195,7 +1195,7 @@ def _open_library(config):
         lib.get_item(0)  # Test database connection.
     except (sqlite3.OperationalError, sqlite3.DatabaseError) as db_error:
         log.debug(u'{}', traceback.format_exc())
-        raise UserError(u"database file {0} could not be opened, reason: {1}".format(
+        raise UserError(u"database file {0} cannot not be opened: {1}".format(
             util.displayable_path(dbpath),
             db_error
         ))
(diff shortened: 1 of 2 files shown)


sampsyo:
Hello! Can you please write something about what you changed and why? Some examples would be extremely useful.

vvvrrooomm:
when beets is started and the database file is on e.g. a volume that is full beets quits with the message  "database file could not be opened {path to file}". The original error message coming from sqlite is lost. the proposed patch attaches the original error as 'reason'. in the case of disk full it reads now "database file {path to file} could not be opened, reason: database or disk is full"

I have no idea how to create a test for it, because the test case either depends on a not-writeable database file or on a mock of sqlite

vvvrrooomm:
Also what it the problem with travis ci? I don't understand the error message

sampsyo:
That helps a lot; thanks!!

Travis is complaining about a line that’s too long and needs to be wrapped:
https://travis-ci.org/beetbox/beets/jobs/416552969#L1103

For grammar’s sake, can we please change the message to `database file {} cannot be opened: {}`, without the `, reason` part? That bit seems unnecessary and makes the style seem somewhat strange.

Could you also please include a changelog entry?

vvvrrooomm:
Ah, got it!. yeah I will fix the text and the ci error

sampsyo:
Looks great; thanks.

PR #3006, opened 2018-08-18
Parallelize the badfiles plugin

bemeurer:
Following my work on #3003 I figured I'd do the same thing for `beet bad`.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -75,6 +75,8 @@ New features:
   Thanks to :user:`wildthyme`.
 * :doc:`/plugins/discogs`: The plugin has rate limiting for the discogs API now.
   :bug:`3081`
+* The `badfiles` plugin now works in parallel (on Python 3 only).
+  Thanks to :user:`bemeurer`.
 
 Changes:
 
--- a/beetsplug/badfiles.py
+++ b/beetsplug/badfiles.py
@@ -20,7 +20,7 @@ from __future__ import division, absolute_import, print_function
 
 from beets.plugins import BeetsPlugin
 from beets.ui import Subcommand
-from beets.util import displayable_path, confit
+from beets.util import displayable_path, confit, par_map
 from beets import ui
 from subprocess import check_output, CalledProcessError, list2cmdline, STDOUT
 import shlex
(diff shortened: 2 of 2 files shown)


bemeurer:
I can try getting this in `util`, @sampsyo, but it's a little sad that we can't use multiprocessing because `self` has thread- local members. I think that should be the first step taken, to do this the right way everywhere possible.

I can try and investigate, any ideas why is it that `self` can't be shared across processes?

sampsyo:
No, I don’t have an immediate guess about what can’t be shared on `self`.

But I actually disagree that multiprocessing would be better than threading here. Since both cases work by shelling out to external programs, we get parallelism *outside* the Python VM—in other words, the GIL is released during the parallel section. So threading will work just fine to get the tasks to execute in parallel, and may even be faster because threads are faster to create and destroy.

bemeurer:
@sampsyo Sure, that's assuming that the slow path lies in the code executed outside of beets, and thus when the GIL is freed, which is true for `beet bad`, but not true for `beet absubmit`. 

After my previous PR got merged I started seeing what speed I would get, and was a little underwhelmed, I think this was due to GIL contention during the upload step. I _think_ this would be improved by multiprocessing, no? I think there are valid places within beets for both threads (badfiles) and processes (absubmit).

sampsyo:
Hmm, I’m not sure. Unless something’s going very wrong, most of the time for absubmit should be spent on invoking the analysis binary and sending the results over the network, both of which are OS calls and should release the GIL. If that’s not happening, maybe some profiling is in order?

bemeurer:
@sampsyo Do not merge this until #3153 has been merged (then the diff will make sense again, I have this branch rebased on top of that one)

bemeurer:
@sampsyo This is ready to be merged!

docs/changelog.rst
```diff
@@ -87,6 +89,8 @@ Changes:
   :bug:`3084`
 
 .. _Groups: https://groups.google.com/forum/#!searchin/beets-users/mbsync|sort:date/beets-users/iwCF6bNdh9A/i1xl4Gx8BQAJ
+=======
+  :bug:`2442`
```
sampsyo: Looks like this line might have gotten into the wrong place?

bemeurer: Oops, must've missed it during the rebase, fixed!

bemeurer:
@sampsyo What's the problem in master, I can take a look :)

sampsyo:
Oops; never mind! It must have been intermittent. There was a failure on AppVeyor that looked like it was coming from a dependency, but that seems to have cleared itself up. Thank you for coming back to this!!

Issue #3008, opened 2018-08-20
replaygain albumpeak on large collections is calculated wrong


vvvrrooomm:
### Problem

while working on parallelizing the replaygain command I found
beets/beetsplug/replaygain.py:202
``       for chunk in self.isplitter(items, self.chunk_at):
                i += 1
                returnchunk = self.compute_chunk_gain(chunk, is_album)
                albumgaintot += returnchunk[-1].gain
                albumpeaktot += returnchunk[-1].peak
                returnchunks = returnchunks + returnchunk[0:-1]
            return chunks.append(Gain(albumgaintot / i, albumpeaktot / i))
``
this averages the album gain and album peak but the peak is supposed to be the maximum, right?
It would have to be something like ``albumpeaktot = max(albumpeaktot ,returnchunk[-1].peak``
What about albumgain, I am not sure here?

jackwilsdon:
Do you think you could please fill in the [issue template](https://raw.githubusercontent.com/beetbox/beets/master/.github/ISSUE_TEMPLATE.md) so we can get a bit more information about your setup?

vvvrrooomm:
I don't think so. I don't have an album with more than 5000 titles. I was reasoning how to refactor the inner part of the for loop to reuse it while making the code run multiple replaygain encoders in parallel.
github ate the linebreaks in the code section, I fixed that

sampsyo:
Got it; yes, that does pretty much look like a bug! Would you mind opening a pull request to see how the tests do?

vvvrrooomm:
I can fix albumpeak but not albumgain. In the meantime I did a test:
``
replaygain: analysis finished: <bs1770gain>
  <album>
    <track total="2" number="1" file="tmplnruql7v&#x2E;ogg">
      <integrated lufs="-15.11" lu="-2.89" />
      <sample-peak spfs="0.30" factor="1.035347" />
    </track>
    <track total="2" number="2" file="tmpogjqseg&#x5F;&#x2E;ogg">
      <integrated lufs="-12.04" lu="-5.96" />
      <sample-peak spfs="0.29" factor="1.033405" />
    </track>
    <summary total="2">
      <integrated lufs="-12.97" lu="-5.03" />
      <sample-peak spfs="0.30" factor="1.035347" />
    </summary>
  </album>
</bs1770gain>

This clearly shows that albumpeak is the max but albumgain is not the average

sampsyo:
[...]

PR #3009, opened 2018-08-20
replaygain: albumpeak on large collections is calculated as average,not maximum (bug 3008)

vvvrrooomm:
this fixes  the first half of #3008 - the peak calculation, 
calculating the albumgain is still wrong

### DIFF
--- a/beetsplug/replaygain.py
+++ b/beetsplug/replaygain.py
@@ -181,7 +181,7 @@ class Bs1770gainBackend(Backend):
                 i += 1
                 returnchunk = self.compute_chunk_gain(chunk, is_album)
                 albumgaintot += returnchunk[-1].gain
-                albumpeaktot += returnchunk[-1].peak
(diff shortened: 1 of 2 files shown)


beetsplug/replaygain.py
```diff
@@ -181,7 +181,7 @@ def compute_gain(self, items, is_album):
                 i += 1
                 returnchunk = self.compute_chunk_gain(chunk, is_album)
                 albumgaintot += returnchunk[-1].gain
-                albumpeaktot += returnchunk[-1].peak
+                albumpeaktot = max(albumpeaktot, returnchunk[-1].peak)
                 returnchunks = returnchunks + returnchunk[0:-1]
             returnchunks.append(Gain(albumgaintot / i, albumpeaktot / i))
```
sampsyo: To compute a maximum, this last `/ i` should be deleted, right?

vvvrrooomm: darn.

sampsyo:
Thanks!!

PR #3011, opened 2018-08-20
convert works in parallel now (on Python 3 only)

vvvrrooomm:
similar to #3006 this patch parallelizes the convert step. Before the encoder used 1 of the 4 available cores with this patch all available cpu cores are used


bemeurer:
@vvvrrooomm Please checkout #3153!

jtpavlock:
closing as abandoned

Issue #3017, opened 2018-08-30
chroma: Bias toward looking up more relevant releases using date/country


Archer4499:
By default, the releases provided by the `chroma` plugin are effectively random, because it just presents the first five that are in the list supplied by Acoustid.

I would like to suggest a change that first sorts them according to the date of release and then their country depending on the `match` configuration options. The default config will still result in the same behaviour as before.

Archer4499@7a8f980

For example the following configuration:
```
match:
    preferred:
        countries: ['US', 'GB|UK']
        original_year: yes
```
will sort by date (oldest first) then by the preferred countries (US then GB/UK then others).

I find this useful because when I have just one or two songs from an artist I like tagging them with the original album that they were released in.

The main point of this being posted as an issue was to check if I should just make it sort by date by default (and still doing countries as an option), because I think it picking random releases isn't actually useful.
<br>
There is also possibly the separate issue of the [mb.py (Line 119)](https://github.com/beetbox/beets/blob/0c033bdaf10b78be04e6ebc8f998d9d3d5f70e3d/beets/autotag/mb.py#L119) file not using regex entries in the `countries` entry.

sampsyo:
Hi! This sounds like an interesting proposal. I think I might be misunderstanding one thing, though: in the importer interface, we typically sort releases by their similarity (i.e., the most "likely" match goes first). In what context are you talking about sorting stuff?

Archer4499:
The Acoustid lookup returns a list of recordings with sublists of releases which contain the MusicBrainz ID and a few other details including the date and country [Acoustid API](https://acoustid.org/webservice) (The request the plugin uses is recordings+releases).

The current behaviour (on line 190) uses the prefix function to just take the first five releases in the list from the Acoustid lookup then look them up in MusicBrainz and return those options, but since the order of the list seems to be semi random (possibly based on date submitted or something) it basically returns random albums that contain the song/s, which a lot of the time are best of or various artist albums.

This would sort the full list from Acoustid first (before the prefix and MusicBrainz lookups occur) to hopefully provide more useful options. And since the whole point of this plugin is to get the info from the fingerprint rather than the metadata in the file, I don't think there are many other ways of selecting relevant results. And beets still gives a few options based on any tags the file may have.

sampsyo:
Got it; thanks! That helps a lot. So the goal is to make some guesses about which releases are likely to be the most "productive" ones to look up. That sounds great!

[...]

PR #3020, opened 2018-08-31
chroma: Bias toward looking up more relevant releases using date/country

Archer4499:
Allow releases from Acoustid lookup to be sorted by date/country
#3017

### DIFF
--- a/beetsplug/chroma.py
+++ b/beetsplug/chroma.py
@@ -67,19 +67,20 @@ def releases_key(release, countries, original_year):
         year = date.get('year', 9999)
         month = date.get('month', 99)
         day = date.get('day', 99)
-        key = '{0:04d}{1:02d}{2:02d}'.format(year, month, day)
     else:
-        key = '99999999'
+        year = 9999
+        month = 99
+        day = 99
 
     # Uses index of preferred countries to sort
+    country_key = 99
     if release.get('country'):
         for i, country in enumerate(countries):
             if country.match(release['country']):
-                key += '{0:02d}'.format(i)
-                return key
+                country_key = i
+                break
 
-    key += '99'
-    return key
+    return (year, month, day, country_key)
 
 
 def acoustid_match(log, path):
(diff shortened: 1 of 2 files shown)


beetsplug/chroma.py
```diff
@@ -57,6 +59,29 @@ def prefix(it, count):
...
 
+def releases_key(release, countries, original_year):
+    """Used as a key to sort releases by date then preferred country
+    """
+    date = release.get('date')
+    if date and original_year:
+        year = date.get('year', 9999)
+        month = date.get('month', 99)
+        day = date.get('day', 99)
+        key = '{0:04d}{1:02d}{2:02d}'.format(year, month, day)
```
sampsyo: Any particular reason this produces a string instead of a tuple? Python will happily sort tuples indexwise:

```py
>>> sorted([(1,2,3), (1,1,1)])
[(1, 1, 1), (1, 2, 3)]
```

beetsplug/chroma.py
```diff
@@ -88,16 +113,24 @@ def acoustid_match(log, path):
         return None
     _acoustids[path] = result['id']
 
-    # Get recording and releases from the result.
+    # Get recording and releases from the result,
+    #  sorted by date and/or country depending on 'match'>'preferred' config.
```
sampsyo: We usually name config options using dot syntax. For example, "the match.preferred configuration option" here. Also, looks like this comment needs to get rewrapped.

beetsplug/chroma.py
```diff
@@ -88,16 +113,24 @@ def acoustid_match(log, path):
...
         return None
     recording_ids = []
-    release_ids = []
+    releases = []
     for recording in result['recordings']:
         recording_ids.append(recording['id'])
         if 'releases' in recording:
-            release_ids += [rel['id'] for rel in recording['releases']]
+            releases.extend(recording['releases'])
+    country_patterns = config['match']['preferred']['countries'].as_str_seq()
```
sampsyo: A comment here might be in order to introduce the reason for all this sorting business.

sampsyo:
Looks great; thanks!! ✨

PR #1638, opened 2015-10-07
Take the audio tracks in a data track into account.

jdetrey:
Hi,

On some releases, the data track of the medium (CD, usually) may contain audio/video files which MusicBrainz takes into account as actual tracks in the release. See for instance [this release](https://musicbrainz.org/release/9ce41d09-40e4-4d33-af0c-7fed1e558dba).

Commit [6ec04e6](https://github.com/jdetrey/python-musicbrainzngs/commit/6ec04e60fa7f0fade153a30f1c69e1614aa3a6a4) of [jdetrey/python-musicbrainzngs](https://github.com/jdetrey/python-musicbrainzngs) adds support for the corresponding `data-track-list` XML node (as per the [MusicBrainz schema](https://github.com/metabrainz/mmd-schema/blob/a43198f5ffdd6b726988481628aa08a05deccb6d/schema/musicbrainz_mmd-2.0.rng#L2051)). This is [pull request #180](https://github.com/alastair/python-musicbrainzngs/pull/180) of [alastair/python-musicbrainzngs](https://github.com/alastair/python-musicbrainzngs).

This patch merges the two track lists when importing a release from MusicBrainz.

Comments welcome :)

Cheers,
Jérémie.

PR #3021, opened 2018-09-02
Allow ignoring of audio data tracks on import

fortes:
https://github.com/beetbox/beets/pull/1638 by @jdetrey changed the importer to take data tracks into imports. This adds a configuration option to disable that feature.

I'm proposing adding this since it seems like video tracks still get included sometimes in `data-track-list`, such as in this release where there is a multimedia track: http://musicbrainz.org/ws/2/release/cbc97377-05f3-4011-a8fa-0aaa1ca9ffd1?inc=artists+artist-credits+recordings+aliases

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -281,7 +281,8 @@ def album_info(release):
             continue
 
         all_tracks = medium['track-list']
-        if 'data-track-list' in medium:
+        if ('data-track-list' in medium
+                and not config['match']['ignore_data_tracks']):
             all_tracks += medium['data-track-list']
         track_count = len(all_tracks)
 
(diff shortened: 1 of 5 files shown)


fortes:
@sampsyo I debating setting this to `yes` until I found the PR that just enabled the feature a few months ago, so I went conservative here in order to avoid breaking people's config.

I'd personally have this as `yes` by default, but can see the other side of the argument. 

Will merge now, I figure we can always change the default in a separate PR.

sampsyo:
OK, sounds good! I would not object to switching the default to yes—this is a new enough feature that I doubt many people are relying on it. (We probably should have put this behind an option to begin with.)

PR #3022, opened 2018-09-04
Make `ignore_data_tracks` `yes` by default

fortes:
Follow up for https://github.com/beetbox/beets/pull/3021

### DIFF
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -129,7 +129,7 @@ match:
     ignored: []
     required: []
     ignored_media: []
-    ignore_data_tracks: no
+    ignore_data_tracks: yes
     ignore_video_tracks: yes
     track_length_grace: 10
     track_length_max: 30
--- a/docs/reference/config.rst
+++ b/docs/reference/config.rst
@@ -815,10 +815,9 @@ ignore_data_tracks
 ~~~~~~~~~~~~~~~~~~~
 
 By default, audio files contained in data tracks within a release are included
-in the album's tracklist. If you do not want them to be included, set it to
-``yes``.
+in the album's tracklist. If you want them to be included, set it ``no``.
 
-Default: ``no``.
+Default: ``yes``.
 
 .. _ignore_video_tracks:
 
--- a/test/test_mb.py
+++ b/test/test_mb.py
@@ -371,7 +371,8 @@ class MBAlbumInfoTest(_common.TestCase):
         self.assertEqual(d.tracks[0].title, 'TITLE ONE')
         self.assertEqual(d.tracks[1].title, 'TITLE TWO')
 
-    def test_no_skip_audio_data_tracks(self):
+    def test_no_skip_audio_data_tracks_if_configured(self):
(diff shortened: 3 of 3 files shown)


sampsyo:
Hmm, looks like one test is failing—maybe it just needs its expectations updated?

fortes:
> Hmm, looks like one test is failing—maybe it just needs its expectations updated?

@sampsyo I feel like an idiot, but I'm not sure why those two tests are failing. There seems to be some interaction between the tests, because I can make unrelated tests fail depending on my changes within these data track tests. I kinda suck at python, so I'm not quite sure what I'm doing wrong here, any chance you can pull down the branch and take a quick look?

sampsyo:
Sure thing; I'll take a look tomorrow.

sampsyo:
Hmm… correct me if I'm wrong, but I think the failing test should logically fail. I only see one test failing (on my machine and on Travis), for starters:
https://travis-ci.org/beetbox/beets/jobs/424390152#L1102

That test looks like this:

```py
def test_no_skip_video_data_tracks_if_configured(self):
    config['match']['ignore_video_tracks'] = False
    tracks = [self._make_track('TITLE ONE', 'ID ONE', 100.0 * 1000.0),
              self._make_track('TITLE TWO', 'ID TWO', 200.0 * 1000.0)]
    data_tracks = [self._make_track('TITLE VIDEO', 'ID VIDEO',
                                    100.0 * 1000.0, False, True)]
    release = self._make_release(tracks=tracks, data_tracks=data_tracks)
    d = mb.album_info(release)
    self.assertEqual(len(d.tracks), 3)
```

In other words, it configures beets to *not* ignore video tracks. But data tracks are still ignored. Then it makes a release with a track that's *both video and data*. It asserts that this track should not be ignore, but in fact it should—because it's a data track! That's certainly what mb.py does:

```py
if ('data-track-list' in medium
        and not config['match']['ignore_data_tracks']):
    all_tracks += medium['data-track-list']
```

All data track lists are ignored when data tracks are ignored, regardless of whether they're video or audio. Does that make sense?

fortes:
@sampsyo Yup, you're right. Just pushed a fix. In my multitasking-addled mind I managed to misread the `skip_video` tests as `skip_audio`, and was looking at the wrong tests.

This does, however, bring up something I did not think about, this new setting overlaps with `ignore_video_tracks` -- should we be combining the config here?

sampsyo:
Great! I did the same thing a couple of times when reading the code.

Hmm—is it the case that *all* video tracks (in MB, at least) are also data tracks? That does seem like something that *could* be combined, but I actually don't have any creative proposals for how to do so—it still seems like you might want to allow data tracks but not video tracks, which requires both settings to exist. Hmm…

fortes:
Yeah, I don't know that there's a great answer. My quick take is that defaults matter, and ignoring video + data seems like a good default. Folks that want something different will have to change config.

sampsyo:
Yep, sounds right to me.

PR #3028, opened 2018-09-09
mbsync no longer queries MusicBrainz when the either the mb_albumid or mb_trackid field is invalid

arogl:
Attempt to stop excess queries to MusicBrainz when an invalid id exist

### DIFF
--- a/beetsplug/mbsync.py
+++ b/beetsplug/mbsync.py
@@ -82,18 +82,24 @@ class MBSyncPlugin(BeetsPlugin):
                                item_formatted)
                 continue
 
-            # Get the MusicBrainz recording info.
-            track_info = hooks.track_for_mbid(item.mb_trackid)
-            if not track_info:
-                self._log.info(u'Recording ID not found: {0} for track {0}',
-                               item.mb_trackid,
+            # Do we have a valid MusicBrainz TrackId
+            if re.match(r"(\d|\w){8}-(\d|\w){4}-(\d|\w){4}-(\d|\w){4}-(\d|\w){12}", item.mb_trackid):
+                # Get the MusicBrainz recording info.
+                track_info = hooks.track_for_mbid(item.mb_trackid)
+                if not track_info:
+                    self._log.info(u'Recording ID not found: {0} for track {0}',
+                                   item.mb_trackid,
+                                   item_formatted)
+                    continue
+
+                # Apply.
+                with lib.transaction():
+                    autotag.apply_item_metadata(item, track_info)
+                    apply_item_changes(lib, item, move, pretend, write)
+            else:
+                self._log.info(u'Skipping singleton with invalid mb_trackid: {0}',
                                item_formatted)
                 continue
-
-            # Apply.
-            with lib.transaction():
-                autotag.apply_item_metadata(item, track_info)
-                apply_item_changes(lib, item, move, pretend, write)
 
     def albums(self, lib, query, move, pretend, write):
         """Retrieve and apply info from the autotagger for albums matched by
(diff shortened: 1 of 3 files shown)


docs/changelog.rst
```diff
@@ -43,6 +43,13 @@ New features:
   disambiguation comment. A new ``releasegroupdisambig`` field has been added.
   :bug:`3024`
 
+Changes:
+
+* :doc:`/plugins/mbsync` no longer queries MusicBrainz when the either the
+  ``mb_albumid`` or ``mb_trackid`` field is invalid
+  Discussion here https://groups.google.com/forum/#!searchin/beets-users/mbsync|sort:date/beets-users/iwCF6bNdh9A/i1xl4Gx8BQAJ
```
sampsyo: Please use a proper ReST link here instead of inlining the URL.

beetsplug/mbsync.py
```diff
@@ -82,18 +84,29 @@ def singletons(self, lib, query, move, pretend, write):
...
 
-            # Get the MusicBrainz recording info.
-            track_info = hooks.track_for_mbid(item.mb_trackid)
-            if not track_info:
-                self._log.info(u'Recording ID not found: {0} for track {0}',
-                               item.mb_trackid,
-                               item_formatted)
-                continue
+            valid_trackid = re.match(
+                     r"(\d|\w){8}-(\d|\w){4}-(\d|\w){4}-(\d|\w){4}-(\d|\w){12}",
```
sampsyo: Perhaps this regex should go as a constant at the top of the file so it’s more easily identified.

beetsplug/mbsync.py
```diff
@@ -109,69 +122,86 @@ def albums(self, lib, query, move, pretend, write):
...
 
-            # Get the MusicBrainz album information.
-            album_info = hooks.album_for_mbid(a.mb_albumid)
-            if not album_info:
-                self._log.info(u'Release ID {0} not found for album {1}',
-                               a.mb_albumid,
-                               album_formatted)
-                continue
+            valid_albumid = re.match(
+                     r"(\d|\w){8}-(\d|\w){4}-(\d|\w){4}-(\d|\w){4}-(\d|\w){12}",
```
sampsyo: Also, deduplicating the regex so it doesn’t need to be copied & pasted would be awesome.

beetsplug/mbsync.py
```diff
@@ -82,18 +84,29 @@ def singletons(self, lib, query, move, pretend, write):
...
-                self._log.info(u'Recording ID not found: {0} for track {0}',
-                               item.mb_trackid,
-                               item_formatted)
-                continue
+            valid_trackid = re.match(
+                     r"(\d|\w){8}-(\d|\w){4}-(\d|\w){4}-(\d|\w){4}-(\d|\w){12}",
+                     item.mb_trackid)
+
+            # Do we have a valid MusicBrainz TrackId
+            if valid_trackid:
```
sampsyo: In order to mimic the style of the next check in this function, what do you think about using `continue` to about early instead of indenting the rest of the loop body under the `if` block? This makes it easy to see that the check aborts the process and more clearly associates the error message with its corresponding check.

arogl:
I have made the requested changes,

beetsplug/mbsync.py
```diff
@@ -22,6 +22,10 @@
 from beets.autotag import hooks
 from collections import defaultdict
 
+import re
+
+mb_regex = r"(\d|\w){8}-(\d|\w){4}-(\d|\w){4}-(\d|\w){4}-(\d|\w){12}"
```
sampsyo: Constant names should be capitalized. (Please see PEP8.)

beetsplug/mbsync.py
```diff
@@ -82,6 +86,13 @@ def singletons(self, lib, query, move, pretend, write):
                                item_formatted)
                 continue
 
+            valid_trackid = re.match(mb_regex, item.mb_trackid)
```
sampsyo: Please put this line after the comment below, which describes it.

beetsplug/mbsync.py
```diff
@@ -82,6 +86,13 @@ def singletons(self, lib, query, move, pretend, write):
...
+
+            # Do we have a valid MusicBrainz TrackId
+            if not valid_trackid:
+                self._log.info(u'Skipping singleton with invalid mb_trackid:' +
+                               ' {0}', item_formatted)
+                continue
```
sampsyo: A blank line for visual separation between “stanzas” is probably in order.

beetsplug/mbsync.py
```diff
@@ -109,6 +120,15 @@ def albums(self, lib, query, move, pretend, write):
 
             items = list(a.items())
 
+            valid_albumid = re.match(mb_regex, a.mb_albumid)
```
sampsyo: As above.

beetsplug/mbsync.py
```diff
@@ -109,6 +120,15 @@ def albums(self, lib, query, move, pretend, write):
...
+
+            # Do we have a valid MusicBrainz AlbumId
+
```
sampsyo: Please delete this blank line to visually associate the comment with the code it describes.

Also, did you mean to end this sentence with a question mark?

docs/changelog.rst
```diff
@@ -43,6 +43,15 @@ New features:
   disambiguation comment. A new ``releasegroupdisambig`` field has been added.
   :bug:`3024`
 
+Changes:
+
+* :doc:`/plugins/mbsync` no longer queries MusicBrainz when the either the
+  ``mb_albumid`` or ``mb_trackid`` field is invalid
+  Discussion on Google groups here_
```
sampsyo: To avoid “using up” the link word “here” for all time, how about this more descriptive link text?

> See also the [discussion on Google Groups].

where [] indicated where the link should go.

beetsplug/mbsync.py
arogl: Updated

beetsplug/mbsync.py
arogl: Updated

docs/changelog.rst
arogl: I think I got this one right

arogl:
@sampsyo I do not know what is causing the travis check to fail.

FAIL: test_multiple_replacements_parsed (test.test_ui.ConfigTest)

As said before if I run that test as a single test it completes. I'll try setting up a linux VM to test

nathdwek:
It seems like the test is comparing objects by reference instead of by value or through a proper `__eq__` method.

This explains why the result is not consistent. Depending on the platform and interpreter history are apparently sometimes reused on your side but not when Travis is running the tests.
Memory model is ID anyway so the test should be rewritten properly to check that the regexes are actually identical.

sampsyo:
Thank you for pointing that out, @nathdwek! I've (hopefully) fixed the underlying issue in the test by comparing the pattern strings instead of the pattern objects. See dc45119a8d1e843fe36ff0ffd031c3b35d48f27d. That should be more reliable.

PR #3030, opened 2018-09-09
move_art: Remove missing album art

projectgus:
Prevents a fatal file system error on "beet update" if the album art file has been moved or renamed
and the beets library hasn't been updated with the new path.

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -1030,6 +1030,12 @@ class Album(LibModel):
         if not old_art:
             return
 
+        if not os.path.exists(old_art):
+            log.error(u'removing reference to missing album art file {}',
+                      util.displayable_path(old_art))
+            self.artpath = None
+            return
+
         new_art = self.art_destination(old_art)
         if new_art == old_art:
             return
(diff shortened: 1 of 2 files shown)


sampsyo:
Interesting idea! This could help remove a common pain point. Have you tried this out fairly thoroughly? It seems innocuous enough, but this is a pretty sensitive path in common workflows, so it would be awesome to make sure we're not doing anything unexpected.

projectgus:
Hi @sampsyo ,

Good question! I've only tested this in my own library. A while ago I used a script to rename a bunch of cover files to folder.jpg (to keep Android happy), and after that `beet update` would crash with a filesystem error. After this change, the next `beet update` removed the stale covert art entries and I could continue.

I'm not sure that this is the best place in the code to be cleaning up this metadata in the library, but I couldn't find a better place...

An alternative change would be just to log that the file doesn't exist here and keep going (rather than quit with a fatal error). However I wasn't sure it's possible for a user to easily fix broken cover art links in another way, so clearing the entry at the same time seemed like a win/win. Happy to change it if you prefer, though.

sampsyo:
Alright; thanks! I think we should go for it. Could you please add a changelog entry?

Also, may I suggest this slightly more direct wording for the error log?

> removing reference to missing album art file {}

projectgus:
@sampsyo Sure thing, done and done!

sampsyo:
Awesome; thank you! Merged. :sparkles:

PR #3032, opened 2018-09-13
Actually run tests on Python 3.7

FichteFoll:
Well, only flake8 was being run.

Before: https://travis-ci.org/beetbox/beets/jobs/426861476#L1143


sampsyo:
Thanks!! :sparkles:

PR #3043, opened 2018-09-27
Make `%aunique` work with `-f/--format` options.

### LINKED ISSUE #3042: %aunique fails when listing albums
### Problem

```
$ beet ls -a -f '$albumartist - $album%aunique{}'
Weezer - Make Believe<no such field 'album_id'>
Weezer - Maladroit<no such field 'album_id'>
Weezer - Pinkerton<no such field 'album_id'>
Weezer - Raditude<no such field 'album_id'>
Weezer - Weezer<no such field 'album_id'>
Weezer - Weezer<no such field 'album_id'>
Weezer - Weezer<no such field 'album_id'>
```

Looks like `aunique` assumes that it'll only ever handle `Item`s (albums don't have an `album_id` field).

This seems like a simple fix. I should have a pull request ready today or tomorrow.

djl:
This should fix #3042. Let's see if Travis gives the thumbs up...

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -1395,6 +1395,8 @@ class Library(dbcore.Database):
         """
         if isinstance(item_or_id, int):
             album_id = item_or_id
+        elif isinstance(item_or_id, Album):
+            album_id = item_or_id.id
         else:
             album_id = item_or_id.album_id
         if album_id is None:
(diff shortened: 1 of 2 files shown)


beets/library.py
```diff
@@ -1395,6 +1395,8 @@ def get_album(self, item_or_id):
         """
         if isinstance(item_or_id, int):
             album_id = item_or_id
+        elif isinstance(item_or_id, Album):
+            album_id = item_or_id.id
```
sampsyo: If this function can now accept Album objects, maybe we should update the docstring to reflect that? (Also, I don’t quite see where the call is that makes this change necessary—maybe that call should be rewritten to just use the Album object it already has instead of going to the database to get another identical copy?)

djl: The call to `lib.get_album()` comes from [tmpl_aunique](https://github.com/beetbox/beets/blob/3b78d933cf46d5cad8f90ea6cdf38ada9e10846f/beets/library.py#L1531) where `self.item` can be either an Item or an Album.

I couldn't think of a clean way of refactoring that method. I suppose `lib.get_album()`, if given an Album, could just return the same object. That would prevent hitting the database twice.

sampsyo: How about using `lib.get_album(album_id)`? In your improved version of `tmpl_aunique`, we now have an `album_id` variable already. That won't prevent hitting the database twice, but it will avoid expanding the responsibilities of `get_album`.

djl: Good idea! Pushed a new commit.

sampsyo:
Awesome. Looks perfect. :tada: :rocket: :mouse:

PR #3044, opened 2018-09-27
Allow "deleting" fixed attributes via `beet modify`.

### LINKED ISSUE #2497: modify: Field deletion (foo!) should work for fixed attributes by setting them to null
### Problem

`foo!` works if foo is a flexible field, but it doesn't do anything if it's a fixed attribute.

Example from IRC Chat:

User not able to delete/reset artpath: to "None" as modify only results in artpath being the current working directory: 

```sh
# beet modify -a 2cellos artpath!
Modifying 1 albums.
No changes to make.
# beet modify -a 2cellos artpath=''
Modifying 1 albums. 
2CELLOS - 2Cellos
  artpath: /media/music/processed/2CELLOS/2011 - 2Cellos/cover.jpg -> /root  
Really modify, move and write tags? (Yes/no/select) n 
```

Led to this problem:

```
artpath: /media/music/processed/2CELLOS/2011 - 2Cellos/cover.jpg -> /root  
```
should be:
```
artpath: /media/music/processed/2CELLOS/2011 - 2Cellos/cover.jpg -> None
```
### Setup

* OS: Linux bedmoode 4.4.30-v7+ #919 SMP Tue Nov 1 16:57:28 GMT 2016 armv7l GNU/Linux 
* Python version: Python 3.4.2
* beets version: beets version 1.4.4

djl:
Deleting a fixed field isn't possible so we just reset the field back to it's "null" value.

Fixes #2497.

I spent a while trying to figure out why `beet modify -a foo artpath!` didn't work before discovering #2497. This *should* fix the problem.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -290,10 +290,11 @@ class Model(object):
         if key in self._values_flex:  # Flexible.
             del self._values_flex[key]
             self._dirty.add(key)  # Mark for dropping on store.
+        elif key in self._fields:  # Fixed
+            setattr(self, key, self._type(key).null)
+            self._dirty.add(key)  # Mark for dropping on store.
         elif key in self._getters():  # Computed.
             raise KeyError(u'computed field {0} cannot be deleted'.format(key))
-        elif key in self._fields:  # Fixed.
-            raise KeyError(u'fixed field {0} cannot be deleted'.format(key))
         else:
             raise KeyError(u'no such field {0}'.format(key))
 
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -292,7 +292,6 @@ class Model(object):
             self._dirty.add(key)  # Mark for dropping on store.
         elif key in self._fields:  # Fixed
             setattr(self, key, self._type(key).null)
-            self._dirty.add(key)  # Mark for dropping on store.
         elif key in self._getters():  # Computed.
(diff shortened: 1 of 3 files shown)


beets/dbcore/db.py
```diff
@@ -290,10 +290,11 @@ def __delitem__(self, key):
         if key in self._values_flex:  # Flexible.
             del self._values_flex[key]
             self._dirty.add(key)  # Mark for dropping on store.
+        elif key in self._fields:  # Fixed
+            setattr(self, key, self._type(key).null)
+            self._dirty.add(key)  # Mark for dropping on store.
```
sampsyo: This comment seems slightly inaccurate for this line: maybe “mark for updating” or “mark for setting to null.”

Actually, come to think of it, I’m not sure this dirtying line is necessary. We need it for flexible attributes because we change the underlying dict directly, but `setattr` should trigger the normal field modification mechanism, which automatically sets the dirty flag. Would you mind giving this a try without this line?

sampsyo (commented):
Great! I’ve wanted this for a while too; I’m glad the change was so simple. And thanks for updating the tests too!

I noticed one line that may not be necessary.

beets/dbcore/db.py
djl: You are correct. It wasn't needed. I've updated the PR.

sampsyo:
Awesome. Thanks for checking that out.

PR #3048, opened 2018-10-10
Subsonicupdate plugin: Added context path and fixed bug in port configuration

samuelnilsson:
This pull request suggests that solves the two following problems:

1. When specifying a port, the following error occurred:

`configuration error: subsonic.port: must be a string`

2. The subsonic configuration allows the user to specify a context path to append to the url, which is currently not supported by the plugin. This pull request contains a fix which suggests that an optional configuration `contextpath` is added.

### DIFF
--- a/beetsplug/subsonicupdate.py
+++ b/beetsplug/subsonicupdate.py
@@ -21,6 +21,7 @@ a "subsonic" section like the following:
         port: 4040 (default)
         user: <your username>
         pass: <your password>
+        contextpath: /subsonic
 """
 from __future__ import division, absolute_import, print_function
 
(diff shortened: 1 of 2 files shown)


sampsyo:
Great; thanks—this looks excellent. Would you mind adding a changeling entry describing the fixes? (Your descriptions in the issue comment will do nicely.)

samuelnilsson:
> Great; thanks—this looks excellent. Would you mind adding a changeling entry describing the fixes? (Your descriptions in the issue comment will do nicely.)

Great! I wouldn't mind doing that. Just wondering, should the changes really be added to the changelog since the plugin was added during the current in development release? The changelog for the current release already contains the following new feature:

`A new /plugins/subsonicupdate can automatically update your Subsonic library.`

sampsyo:
Argh; you're absolutely right! No, no need for an additional entry. I really need to cut a new release…

PR #3054, opened 2018-10-16
WIP: Remove `default` arg for `lastgenre --force` flag

### LINKED ISSUE #2704: lastgenre: `force` config option doesn’t work
### Problem

I want to write the popular genres from Last.fm in the metatags of my music.

I have the following configuration:

```
lastgenre:
    whitelist: no
    count: 5
    fallback: nogenre
    min_weight: 20
    source: track
    force: yes
```

Now I run "beets lastgenre"

Other than what I had expected that beets would now write the 5 most popular genres into each file, it just outputs with (keep) and doesn't change anything to the genre metatag...

What could be going wrong here and how can I fix this?

Thanks a lot for any help!

discopatrick:
To fix #2704

### DIFF
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -373,7 +373,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         lastgenre_cmd = ui.Subcommand('lastgenre', help=u'fetch genres')
         lastgenre_cmd.parser.add_option(
             u'-f', u'--force', dest='force',
-            action='store_true', default=False,
+            action='store_true',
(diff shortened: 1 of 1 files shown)


sampsyo:
Looks right to me! :smiley: :sparkles:

discopatrick:
@sampsyo I marked as WIP because this is not tested! But if you're confident to merge, please go ahead.

arcresu:
FWIW I was able to reproduce the original issue and tested this change at it looks like it does the trick.

sampsyo:
Yay!! Merged. :sparkles:

PR #3055, opened 2018-10-17
Replaygain backend r128gain

zsinskri:
Add a replaygain backend using the [r128gain](https://github.com/desbma/r128gain/) python module. 
It makes for an easier to install (as r128gain is on PyPI) alternative to the bs1770gain backend to create R128_* replaygain tags.


zsinskri:
It looks like #3056 will be merged. That will make this pull request rather uninteresting, as this backend uses ffmpeg (indirectly), too.
Closing. Or is this r128gain backend still useful?

PR #3056, opened 2018-10-18
Replaygain backend ffmpeg

zsinskri:
Add a replaygain backend using the [ffmpeg](https://ffmpeg.org) CLI-tool and its [ebur128](https://ffmpeg.org/ffmpeg-filters.html#ebur128-1) filter.
It is a alternative to the bs1770gain to create R128_* tags.

## comparision with r128gain backend
#3055 also uses ffmpeg, but with an intermediary python module ("r128gain"). This pull request replicates some of the work done in r128gain, but avoids the (beta-) dependency.

The album gain algorithm implemented in this backend - the mean of all track gains, also used by the bs1770gain backend - gives different results than the implementation in r128gain (calculating the gain for a concatenation of all tracks). Calculating the mean is obviously faster than rescanning the whole album, but might deliver worse results.

This backend also does not implement any kind of threading (a single track is scanned at a time), but r128gain does.

### DIFF
--- a/beetsplug/replaygain.py
+++ b/beetsplug/replaygain.py
@@ -531,7 +531,7 @@ class FfmpegBackend(Backend):
     def _parse_float(self, line):
         """Extract a float from a key value pair in `line`.
 
-        This format is expected: /[^:]:\s*value.*/, where `value` is
+        This format is expected: /[^:]:[[:space:]]*value.*/, where `value` is
         the float.
         """
         # extract value
--- a/beetsplug/replaygain.py
+++ b/beetsplug/replaygain.py
@@ -306,7 +306,7 @@ class Bs1770gainBackend(Backend):
 
 # ffmpeg backend
 class FfmpegBackend(Backend):
-    """A replaygain backend using ffmpegs ebur128 filter.
+    """A replaygain backend using ffmpeg's ebur128 filter.
     """
     def __init__(self, config, log):
         super(FfmpegBackend, self).__init__(config, log)
@@ -342,7 +342,7 @@ class FfmpegBackend(Backend):
             )
 
         # check that peak_method is valid
-        valid_peak_method = ("true", "sample")
+        valid_peak_method = "true", "sample"
         if self._peak_method not in valid_peak_method:
             raise ui.UserError(
                 u"Selected ReplayGain peak method {0} is not supported. "
@@ -437,8 +437,8 @@ class FfmpegBackend(Backend):
         ]
 
     def _analyse_item(self, item, count_blocks=True):
-        """Analyse item. Returns a Pair (Gain object, number of gating
-        blocks above threshold).
+        """Analyse item. Return a pair of a Gain object and the number
+        of gating blocks above the threshold.
 
         If `count_blocks` is False, the number of gating blocks returned
         will be 0.
@@ -459,7 +459,7 @@ class FfmpegBackend(Backend):
             line_peak = self._find_line(
                 output,
                 "  {0} peak:".format(self._peak_method.capitalize()).encode(),
-                start_line=(len(output) - 1), step_size=-1,
+                start_line=len(output) - 1, step_size=-1,
             )
             peak = self._parse_float(
                 output[self._find_line(
@@ -472,7 +472,7 @@ class FfmpegBackend(Backend):
 
         line_integrated_loudness = self._find_line(
             output, b"  Integrated loudness:",
-            start_line=(len(output) - 1), step_size=-1,
+            start_line=len(output) - 1, step_size=-1,
         )
         gain = self._parse_float(
             output[self._find_line(
@@ -529,9 +529,10 @@ class FfmpegBackend(Backend):
             )
 
     def _parse_float(self, line):
-        """Extract a float.
+        """Extract a float from a key value pair in `line`.
 
-        Extract a float from a key value pair in `line`.
+        This format is expected: /[^:]:\s*value.*/, where `value` is
+        the float.
         """
         # extract value
         value = line.split(b":", 1)
@@ -543,7 +544,7 @@ class FfmpegBackend(Backend):
         value = value[1].lstrip()
         # strip unit
         value = value.split(b" ", 1)[0]
-        # cast value to as_type
+        # cast value to float
         try:
             return float(value)
         except ValueError:
--- a/beetsplug/replaygain.py
+++ b/beetsplug/replaygain.py
@@ -18,6 +18,7 @@ from __future__ import division, absolute_import, print_function
 import subprocess
 import os
 import collections
+import math
 import sys
 import warnings
 import xml.parsers.expat
@@ -64,9 +65,22 @@ def call(args, **kwargs):
         raise ReplayGainError(u"argument encoding failed")
 
 
+def db_to_lufs(db):
+    """Convert db to LUFS.
+
+    According to https://wiki.hydrogenaud.io/index.php?title=
+      ReplayGain_2.0_specification#Reference_level
+    """
+    return db - 107
+
+
 # Backend base and plumbing classes.
 
+# gain: in LU to reference level
+# peak: part of full scale (FS is 1.0)
 Gain = collections.namedtuple("Gain", "gain peak")
+# album_gain: Gain object
+# track_gains: list of Gain objects
 AlbumGain = collections.namedtuple("AlbumGain", "album_gain track_gains")
 
 
@@ -81,11 +95,15 @@ class Backend(object):
         self._log = log
 
     def compute_track_gain(self, items):
+        """Computes the track gain of the given tracks, returns a list
+        of Gain objects.
+        """
         raise NotImplementedError()
 
     def compute_album_gain(self, items):
-        # TODO: implement album gain in terms of track gain of the
-        # individual tracks which can be used for any backend.
+        """Computes the album gain of the given album, returns an
+        AlbumGain object.
+        """
         raise NotImplementedError()
 
     def use_ebu_r128(self):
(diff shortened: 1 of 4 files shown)


zsinskri:
I'm not really happy with my changes to command_output: it feels wrong to just append both streams. Is there a better way to access stderr with the existing tooling in beets?

zsinskri:
I just found [this](https://github.com/beetbox/beets/issues/3008) conversation. It seems like using the mean to calculate the album gain is just plain wrong. I will look into this...

beets/util/__init__.py
```diff
@@ -829,6 +832,8 @@ def command_output(cmd, shell=False):
             cmd=' '.join(cmd),
             output=stdout + stderr,
         )
+    if include_stderr:
+        stdout += stderr
     return stdout
```
sampsyo: What about just returning `stdout, stderr` (when requested, or always) and letting the client pull out what they need?

beetsplug/replaygain.py
```diff
@@ -81,12 +81,28 @@ def __init__(self, config, log):
...
-        # individual tracks which can be used for any backend.
-        raise NotImplementedError()
+        """Computes the album gain of the given album, returns an
+        AlbumGain object.
+        """
+        # calculate track gains
+        tracks = self.compute_track_gain(album.items())
+
+        # calculate album gain
+        album_gain = sum(track.gain for track in tracks) / len(tracks)
```
sampsyo: As you mentioned in the PR thread, we probably need a more "interesting" average than the plain arithmetic mean here.

beetsplug/replaygain.py
```diff
@@ -81,13 +86,21 @@ def __init__(self, config, log):
         self._log = log
 
     def compute_track_gain(self, items):
+        """Computes the track gain of the given tracks, returns a list
+        of Gain objects.
+        """
         raise NotImplementedError()
 
     def compute_album_gain(self, album):
-        # TODO: implement album gain in terms of track gain of the
```
zsinskri: I have not restored this `TODO` after removing the wrong implementation as I do not think it is possible to implement a correct default implementation without creating a new item/file (a concatenation of all the album's items). E.g. when using the EBU R128 spec we don't know how many gating blocks in each track were over the gating threshold, and the album gain cannot be calculated from the track gains without that information.
One could use the length of the tracks as a proxy for the number of gating blocks. That might get reasonable results but would still be _wrong_.

Another option would be to implement a method computing the album gain from the track gains and weights (i.e. number of gating blocks above threshold) according to EBU R128. Thus, the main album gain calculation could be used by all backends that can provide some weights.
It should be really easy to move that calculation from `FffmpegBackend` to `Backend`, but I do not know if any other `Backend` could possibly make use of that?

beetsplug/replaygain.py
```diff
@@ -271,6 +284,210 @@ def end_element_handler(name):
...
+    def __init__(self, config, log):
+        super(FfmpegBackend, self).__init__(config, log)
+        config.add({
+            "peak": "true"
+        })
+        self._peak_method = config["peak"].as_str()
+        # convert db to LUFS according to:
+        # http://wiki.hydrogenaud.io/index.php?title=ReplayGain_
+        #   specification#Reference_level
+        self._target_level = config['targetlevel'].as_number() - 103
```
zsinskri: How exactly are the options `tragetlevel` and `r128` supposed to work together?
The only backend that already supports `r128`, `Bs1770gainBackend`, just completely ignores `targetlevel`. (That might be a bug?)

zsinskri: In `Bs1770gainBackend` initialising it as an EBU R128 Backend only sets the target level to -23 LUFS. So it is impossible to configure the target level of R128_* tags.
Maybe it would make sense to add a config `r128_targetlevel`, defaulting to `80` db (that corresponds to -23 LUFS according to http://wiki.hydrogenaud.io/index.php?title=ReplayGain_specification#Reference_level).
Then add a parameter `target_level` to `Backend.compute_track_gain` and `Backend.compute_album_gain` that is either `targetlevel` or `r128_targetlevel` depending on the filetype. That also makes the whole `ReplayGainPlugin.r128_backend_instance` obsolete, the same backend can just be called with different target levels.
I guess I should create a separate pull request implementing that change?

Also `Bs1770gainBackend`'s method configuration just changes the target level. Shouldn't that be controlled by `targetlevel`?

beetsplug/replaygain.py
sampsyo: OK, thanks for carefully investigating. I'ma also not aware of any other backend that can provide the necessary information about weighing for tracks.

beetsplug/replaygain.py
sampsyo: Yes, I like that proposed design a lot—with that change, it would be great to dispense with the distinction between different "backend instances" for classic RG and R128.

I also agree that it seems wrong for `Bs1770gainBackend` to unilaterally change the target level. It does seem like the configuration should control that.

sampsyo:
This is already looking great! Thank you for carefully considering all the details here.

If it seems appropriate to you, let's merge this PR and work on the refactoring you mentioned in a separate PR?

zsinskri:
Yes, I'm fine with merging.

zsinskri:
Have I missed something, that I should do to get this pull request merged, or is it just waiting for a review?

sampsyo:
It's just in my queue to review. Thanks for your patience; I'll get to it real soon!

zsinskri:
Just merged master and resolved conflicts, no new features in latest push.

zsinskri:
Testing this backend's accuracy (especially the custom album gain calculation) with an unscientific sample size of 1:

|| beets w/ ffmpeg | [r128gain](https://github.com/desbma/r128gain) (uses ffmpeg) | beets w/ bs1770gain |
|-|-|-|-|
| R128_TRACK_GAIN | -1536 | -1536 | -1526 |
| R128_ALBUM_GAIN | -1662 | -1664 | -1669 |

These are with #3314 merged (otherwise every backend will produce garbage).

r128gain calculates the album gain by truly concatenating all tracks and letting ffmpeg re-analyse all audio. We are only 0.0078125 LU (= (1664-1662)/pow(2,8)) off. Considering the difference between bs1770gain and r128gain is 3.5 times as large (or for the track gain 5 times as large) this seems quite good.

zsinskri:
Both ffmpeg and bs1770gain can calculate either the sample or true peak. At least to my understanding the sample peak is the highest stored value and thus really fast to compute, while the true peak is the maximum of the resulting waveform and takes somewhat longer.

The ffmpeg backend currently offers the `peak` configuration option, defaulting to true peak as the played waveform is what we actually care about. (Don't we? Is there any relevant standart?)

But the bs1770gain backend unconditionally chooses the sample peak (by using the `-p`, not `-t`,  CLI-Option). Should I change the default value of the `peak` option to `sample`? Why was the sample peak chosen for the bs1770gain backend?

I guess bs1770gain should also respect `peak` if we do not decide to remove that option. But that is something for a different PR.

sampsyo:
Huh, good point. I don’t know why that was chosen as the default behavior for the existing backend. It does seem like the backends should be consistent, but that true peak should be the default. Let’s make it do that—perhaps in a PR just for cleaning up the semantics of the “peak” config option?

zsinskri:
The `peak` config option is introduced by this PR as a ffmpeg specific option (as I did not know that other backends could support it too). 

> cleaning up the semantics of the “peak” config option?

If its semantics are off that should be fixed right here and only bs1770gain's support added in a separate PR to avoid merging suboptimal code. Or should we move the whole `peak` option to a new PR?

sampsyo:
Aha! Sorry, I misunderstood. No, I think it’s fine to keep it here, but I like your idea of making the other backend respect the option too (and use the same default).

beetsplug/replaygain.py
```diff
@@ -277,6 +299,246 @@ def end_element_handler(name):
...
+        super(FfmpegBackend, self).__init__(config, log)
+        config.add({
+            "peak": "true"
+        })
+        self._peak_method = config["peak"].as_str()
+        self._target_level = db_to_lufs(config['targetlevel'].as_number())
+        self._ffmpeg_path = "ffmpeg"
+
+        # check that ffmpeg is installed
+        # TODO: check if ffmpeg version is sufficient
```
zsinskri: Should we really check the ffmpeg version here? If so, how? Most other backends do not do this.

I have not tested any versions < libavfilter 7.16.100, but we can most likely support [libavfilter 6.67.100](https://git.videolan.org/?p=ffmpeg.git;a=commit;h=005d058f4230f3207ebcf1131df7426d4f57392f).

Before that ffmpeg had to be configured with the `--enable-libebur128` option. Should we maybe just raise a warning for ffmpeg versions < 6.67.100? That would allow users with that feature enabled to still use this backend.

If we decide to check the version: Are there any utility functions relevant to this in beets or the python stdlib? How should the libavfilter version be parsed and compared?

sampsyo: Good question. If this were super easy to do, I would probably say yes—it wouldn’t hurt to issue a reasonable warning if the backend is missing. But it actually seems pretty complicated to do. You’d have to get the stdout from an `ffmpeg -version`, as we do [in audioread](https://github.com/beetbox/audioread/blob/6a8e7283f380498d28f94a60631b65d3fa623be9/audioread/ffdec.py#L101), and then parse out the information we need, which could be somewhat brittle.

Do you know how widespread the required minimum version is? For example, it is always a useful barometer to check with version of Debian it was first included in.

zsinskri: I just checked an up-to-date Debian install
```sh
$ uname -a
Linux ares 4.9.0-9-amd64 #1 SMP Debian 4.9.168-1+deb9u3 (2019-06-16) x86_64 GNU/Linux
$ ffmpeg -version
ffmpeg version 3.2.14-1~deb9u1 Copyright (c) 2000-2019 the FFmpeg developers
built with gcc 6.3.0 (Debian 6.3.0-18+deb9u1) 20170516
configuration: --prefix=/usr --extra-version='1~deb9u1' --toolchain=hardened --libdir=/usr/lib/x86_64-linux-gnu --incdir=/usr/include/x86_64-linux-gnu --enable-gpl --disable-stripping --enable-avresample --enable-avisynth --enable-gnutls --enable-ladspa --enable-libass --enable-libbluray --enable-libbs2b --enable-libcaca --enable-libcdio --enable-libebur128 --enable-libflite --enable-libfontconfig --enable-libfreetype --enable-libfribidi --enable-libgme --enable-libgsm --enable-libmp3lame --enable-libopenjpeg --enable-libopenmpt --enable-libopus --enable-libpulse --enable-librubberband --enable-libshine --enable-libsnappy --enable-libsoxr --enable-libspeex --enable-libssh --enable-libtheora --enable-libtwolame --enable-libvorbis --enable-libvpx --enable-libwavpack --enable-libwebp --enable-libx265 --enable-libxvid --enable-libzmq --enable-libzvbi --enable-omx --enable-openal --enable-opengl --enable-sdl2 --enable-libdc1394 --enable-libiec61883 --enable-chromaprint --enable-frei0r --enable-libopencv --enable-libx264 --enable-shared
libavutil      55. 34.101 / 55. 34.101
libavcodec     57. 64.101 / 57. 64.101
libavformat    57. 56.101 / 57. 56.101
libavdevice    57.  1.100 / 57.  1.100
libavfilter     6. 65.100 /  6. 65.100
libavresample   3.  1.  0 /  3.  1.  0
libswscale      4.  2.100 /  4.  2.100
libswresample   2.  3.100 /  2.  3.100
libpostproc    54.  1.100 / 54.  1.100
```

It's libavfilter version is older than the included ebur128 library, but it was configured to use an external libebur128 library (`--enable-libebur128`). I only inspected the output visually, but it seems to be compatible.

To avoid a warning on compatible Debian systems we should check both the libavfilter version as well as the compile-time-configuration option.

zsinskri: This check is now implemented. Though it raises an Error, not a Warning  Mostly as I did not find an existing suitable Warning class. But since we check the `--enable-libebur128` option, too, I guess that is OK?

beetsplug/replaygain.py
zsinskri: Done by now, forgot to 'resolve conservation'.
We weigh the track gains with the number of gating blocks above the threshold.

beets/util/__init__.py
zsinskri: Done by now, forgot to 'resolve conservation'.
We return a `namedtuple` of `stdout` and `stderr`, allowing everyone to choose what they need while being explicit.

zsinskri:
Now everything is uptodate with `master` and even the version check has been added.

This branch should be ready to merge once again.
If there is anything I can do to help the review process of this merge request, please let me know.

beets/util/__init__.py
```diff
@@ -794,9 +798,15 @@ def convert(arg):
     return [convert(a) for a in args]
 
 
+# stdout and stderr as bytes
+CommandOutput = namedtuple("CommandOutput", ("stdout", "stderr"))
```
fortes: I don't know much about ffmpeg, etc -- but if you have the patience to split up this `CommandOutput` change into a separate PR it feels pretty low risk so I'll happily sign off so we can at least get part of this merged and make the rest easier to review?

zsinskri: That is a great suggestion, I'll get right to it!

zsinskri: Done with  #3329.

zsinskri:
I have rebased this branch onto #3329 which is now a perquisite for these changes (as per fortes great suggestion to split #3329 off), cleaning up some history in the process.

zsinskri:
Rebasing onto `master` again so that GitHub does not show changes from #3329 under "Files changed". This should make this PR easier to review.

beetsplug/replaygain.py
```diff
@@ -277,6 +299,264 @@ def end_element_handler(name):
...
+                    incompatible_ffmpeg = False
+        if incompatible_ffmpeg:
+            raise FatalReplayGainError(
+                u"Installed FFmpeg version does not support ReplayGain."
+                u"calculation. Either libavfilter version 6.67.100 or above or"
+                u"the --enable-libebur128 configuration option is required."
+            )
+
+        # check that peak_method is valid
+        valid_peak_method = ("true", "sample")
```
sampsyo: No parentheses are necessary here.

beetsplug/replaygain.py
```diff
@@ -277,6 +299,264 @@ def end_element_handler(name):
...
         return out
 
+    def use_ebu_r128(self):
+        """Set this Backend up to use EBU R128."""
+        self.method = '--ebu'
+
+
+# ffmpeg backend
+class FfmpegBackend(Backend):
+    """A replaygain backend using ffmpegs ebur128 filter.
```
sampsyo:
```suggestion
    """A replaygain backend using ffmpeg's ebur128 filter.
```

beetsplug/replaygain.py
```diff
@@ -277,6 +299,264 @@ def end_element_handler(name):
...
+            "-filter",
+            "ebur128=peak={0}".format(peak_method),
+            "-f",
+            "null",
+            "-",
+        ]
+
+    def _analyse_item(self, item, count_blocks=True):
+        """Analyse item. Returns a Pair (Gain object, number of gating
+        blocks above threshold).
```
sampsyo: > Analyse item. Return a pair of a Gain object and the number of gating blocks above the threshold.

beetsplug/replaygain.py
```diff
@@ -277,6 +299,264 @@ def end_element_handler(name):
...
+                    output, b"    Peak:",
+                    line_peak,
+                )]
+            )
+            # convert TPFS -> part of FS
+            peak = 10**(peak / 20)
+
+        line_integrated_loudness = self._find_line(
+            output, b"  Integrated loudness:",
+            start_line=(len(output) - 1), step_size=-1,
```
sampsyo: No parentheses are necessary here.

beetsplug/replaygain.py
```diff
@@ -277,6 +299,264 @@ def end_element_handler(name):
...
+                return i
+        raise ReplayGainError(
+            u"ffmpeg output: missing {0} after line {1}"
+            .format(repr(search), start_line)
+            )
+
+    def _parse_float(self, line):
+        """Extract a float.
+
+        Extract a float from a key value pair in `line`.
```
sampsyo: This extra sentence may not add much?

docs/plugins/replaygain.rst
```diff
@@ -10,10 +10,10 @@ playback levels.
...
 ------------
 
-This plugin can use one of three backends to compute the ReplayGain values:
-GStreamer, mp3gain (and its cousin, aacgain), Python Audio Tools. mp3gain
-can be easier to install but GStreamer and Audio Tools support more audio
-formats.
+This plugin can use one of many backends to compute the ReplayGain values:
+GStreamer, mp3gain (and its cousin, aacgain), Python Audio Tools or ffmpeg.
+mp3gain can be easier to install but GStreamer, Audio Tools and ffmpeg support
+more audio formats.
```
sampsyo: I would even say that ffmpeg belongs to the "easy to install" camp! It's available in nearly every package manager ever, and it usually comes as a complete package with everything included.

sampsyo:
This looks perfect! Thank you again for all your hard work on this—this new backend will be a _very_ useful alternative IMO. :tada:

beetsplug/replaygain.py
zsinskri: This has been implemented in #3065.

PR #3060, opened 2018-10-23
When selecting items to modify, use configured order

### LINKED ISSUE #3501: no error raised about malformed yml
It seems like that the loading of malformed yml files included in the configuration through the ìnclude`key does not raise the expected error.

### Problem
Whilst working on PR #3498 I purposefully included a malformed yml file just to test if the newly introduced warning would also catch the error about it:
```yml
include: 
    - badformat.yml
```
and with the 'badformat.yml' file:
```yml
this
    will
        not
            pass
        as
    ok
yml
```
I expected a confuse.ConfigReadError but no errors were raised.
This is what I got:
```bash
$ ./beet
configuration error: musicbrainz.host not found
```

More investigation is needed.

### Setup

* OS: macos 10.13.6 (17G11023)
* Python version: 3.7.6
* beets version: 1.4.9
* My configuration (output of `beet config`) is:  malformed :):):)

TaizoSimpson:
Closes #3501

Changes taken pretty much verbatim from the issue.  I haven't added tests or a changelog entry because this seemed pretty insignificant, but I'd be happy do do either if you see fit.

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1376,10 +1376,10 @@ def modify_items(lib, mods, dels, query, write, move, album, confirm):
     # objects.
     print_(u'Modifying {0} {1}s.'
            .format(len(objs), u'album' if album else u'item'))
-    changed = set()
+    changed = []
     for obj in objs:
         if print_and_modify(obj, mods, dels):
-            changed.add(obj)
+            changed.append(obj)
 
     # Still something to do?
     if not changed:


NoahTheDuke:
One thing to note is this doesn't remove duplicates anymore. You might want to change it to something like `if obj not in changed: changed.append(obj)`. This is an additional check, but it prevents issues later if for some reason duplicates are introduced accidentally.

sampsyo:
That's true. I could be wrong, but I _think_ that duplicates should be impossible here—in the result of a database query, we should only get distinct rows from the database.

TaizoSimpson:
Sounds good.  I've added a changelog entry and a check for duplicates.

PR #3061, opened 2018-10-23
docs: update the pip install instructions

pieqq:
Hello! I was reading Beets' install guide when I saw that it suggested using ``sudo`` if ``pip install beets`` didn't work.

Using ``sudo`` to install Python packages using pip is [not a good practice security-wise](https://askubuntu.com/questions/802544/is-sudo-pip-install-still-a-broken-practice). The proposed PR will suggest an alternative way using ``--user`` to make sure the package is installed locally.

Cheers!


pieqq:
Thanks for the suggestion! I reworded the part you indicated (and re-wrapped the text), and I modified the macOS section. I left some parts because I think it's informative, but removed the mentions to ``sudo``.

sampsyo:
Lovely; thank you!! :sparkles:

PR #3065, opened 2018-10-28
replaygain: target level refactor

zsinskri:
Remove REPLAYGAIN_* vs R128_* tag handling from the Backends, restrict it to the main plugin logic. That is archieved by passing the target loudness level to the backend with every album/track. It also necessitates that all Backends properly implement the target level.
As the different tag formats call for different default target levels this also seperates the configuration of those levels.

### DIFF
--- a/beetsplug/replaygain.py
+++ b/beetsplug/replaygain.py
@@ -21,6 +21,7 @@ import collections
 import math
 import sys
 import warnings
+import enum
 import xml.parsers.expat
 from six.moves import zip
 
@@ -74,6 +75,15 @@ def db_to_lufs(db):
     return db - 107
 
 
+def lufs_to_db(db):
+    """Convert LUFS to db.
+
+    According to https://wiki.hydrogenaud.io/index.php?title=
+      ReplayGain_2.0_specification#Reference_level
+    """
+    return db + 107
+
+
 # Backend base and plumbing classes.
 
 # gain: in LU to reference level
@@ -82,6 +92,12 @@ Gain = collections.namedtuple("Gain", "gain peak")
 # album_gain: Gain object
 # track_gains: list of Gain objects
 AlbumGain = collections.namedtuple("AlbumGain", "album_gain track_gains")
+
+
+class Peak(enum.Enum):
+    none = 0
+    true = 1
+    sample = 2
 
 
 class Backend(object):
@@ -94,21 +110,17 @@ class Backend(object):
         """
         self._log = log
 
-    def compute_track_gain(self, items):
+    def compute_track_gain(self, items, target_level, peak):
         """Computes the track gain of the given tracks, returns a list
         of Gain objects.
         """
         raise NotImplementedError()
 
-    def compute_album_gain(self, items):
+    def compute_album_gain(self, items, target_level, peak):
         """Computes the album gain of the given album, returns an
         AlbumGain object.
         """
         raise NotImplementedError()
-
-    def use_ebu_r128(self):
-        """Set this Backend up to use EBU R128."""
-        pass
 
 
 # bsg1770gain backend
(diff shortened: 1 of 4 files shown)


zsinskri:
The latest push merges master (or rather #3056, from which this branch was forked), resolving conflicts. It also fixes the `ReplayGainLdnsCliMalformedTest`, which caused unrelated tests to fail, due to missing cleanup.

zsinskri:
Is there any guideline on when to merge and when to rebase for this project? I will rework this branch's structure if a rebase is preferred.

zsinskri:
Rebased onto master.

zsinskri:
#3056 is merged. This is now ready for review.

beetsplug/replaygain.py
```diff
@@ -232,6 +267,12 @@ def compute_chunk_gain(self, items, is_album):
 
         self._log.debug(u'analysis finished: {0}', output)
         results = self.parse_tool_output(output, path_list, is_album)
+
+        if gain_adjustment != 0:
+            for i in range(len(results)):
+                orig = results[i]
+                results[i] = Gain(orig.gain + gain_adjustment, orig.peak)
```
sampsyo: Can be simplified a bit:

```py
for i, orig in enumerate(results):
    results[i] = Gain(orig.gain + gain_adjustment, orig.peak)
```

beetsplug/replaygain.py
```diff
@@ -232,6 +267,12 @@ def compute_chunk_gain(self, items, is_album):
 
         self._log.debug(u'analysis finished: {0}', output)
         results = self.parse_tool_output(output, path_list, is_album)
+
+        if gain_adjustment != 0:
```
sampsyo: Zero integers are falsy in Python:
```suggestion
        if gain_adjustment:
```

beetsplug/replaygain.py
```diff
@@ -436,29 +462,36 @@ def _construct_cmd(self, item, peak_method):
...
 
         If `count_blocks` is False, the number of gating blocks returned
         will be 0.
         """
+        target_level_lufs = db_to_lufs(target_level)
+        peak_method = {
+            Peak.none: "none",
+            Peak.true: "true",
+            Peak.sample: "sample",
+        }[peak]
```
sampsyo: Seems like `peak.name` would also work and avoid the lookup table?

beetsplug/replaygain.py
```diff
@@ -436,29 +462,36 @@ def _construct_cmd(self, item, peak_method):
...
+        cmd = self._construct_cmd(item, peak_method)
         self._log.debug(
             u'executing {0}', u' '.join(map(displayable_path, cmd))
         )
         output = call(cmd).stderr.splitlines()
 
         # parse output
 
-        if self._peak_method == "none":
+        if peak is Peak.none:
```
sampsyo: The [Python docs](https://docs.python.org/3/library/enum.html) suggest that enum values should be compared with equality (`==`), not identity (`is`).

docs/changelog.rst
```diff
@@ -42,6 +42,13 @@ New features:
   new ``discogs_albumid`` field from the Discogs API.
   Thanks to :user:`thedevilisinthedetails`.
   :bug:`465` :bug:`3322`
+* :doc:`plugins/replaygain`: ``r128_targetlevel`` is a new configuration option
+  for the ReplayGain plugin: It defines the reference volume for files using
+  ``R128_`` tags. ``targtelevel`` only configures the reference volume for
```
```suggestion
  ``R128_`` tags. ``targetlevel`` only configures the reference volume for
```

docs/plugins/replaygain.rst
```diff
@@ -120,6 +125,13 @@ This option only works with the "ffmpeg" backend:
 - **peak**: Either ``true`` (the default) or ``sample``. ``true`` is
   more accurate but slower.
 
+This option is deprecated:
+
+- **method**: The loudness scanning standard: either `replaygain` for
+  ReplayGain 2.0, `ebu` for EBU R128, or `atsc` for ATSC A/85. This dictates
+  the reference level: -18, -23, or -24 LUFS respectively. Only supported by
+  "bs1770gain" backend.
```
sampsyo: I think it's fine to just leave this one off, if we don't want anyone to use it anymore.

beetsplug/replaygain.py
sampsyo: Or perhaps better yet:

```py
results = [Gain(orig.gain + gain_adjustment, orig.peak) for orig in results]
```

sampsyo:
Awesome; thank you! This is a heroic effort that required lots of special domain expertise, so I wanted to offer extra appreciation for getting this right. :sparkles:

PR #3067, opened 2018-10-31
permissions: allow fixing permissions before writing tags

### LINKED ISSUE #3039: Don't import files with missing permissions on move+write
When importing files with moving and metadata writing enabled [import: \n write: yes \n move: yes], files always get moved and added to the library. However when the user has no write access to these files, the files get imported anyway, but the metadata can not be written.

For me this occurs every now and then when importing from an unpacked archive, that was created with different permissions.

Archlinux, python 3.7, beets 1.4.7
plugins: bandcamp, beatport, bucket, discogs, fetchart, inline, the

p3732:
Add write_import event before writing to files upon import. Use this event to enhance the permissions plugin to fix permissions before writing data. Closes #3039

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -737,6 +737,8 @@ class ImportTask(BaseImportTask):
                     # old paths.
                     item.move(operation)
 
+            plugins.send('write_import', lib=session.lib, item=item)
+
             if write and (self.apply or self.choice_flag == action.RETAG):
                 item.try_write()
 
--- a/beetsplug/permissions.py
+++ b/beetsplug/permissions.py
@@ -8,6 +8,7 @@ like the following in your config.yaml to configure:
     permissions:
             file: 644
             dir: 755
+            before_write: yes
 """
 import os
 from beets import config, util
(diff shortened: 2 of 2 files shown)


arcresu:
It looks like your changes cause a failure in the test suite. Would you mind looking into what needs to be changed there to fix the test?

p3732:
Yes, I noticed that too and took a quick look at it. However it seemed overwhelming at the time. I will try to have another look at it next month.

Moonbase59:
Probably caused by Linux `/tmp` usually having sticky bit permissions (`1664` instead of `0777` you were apparently going for).

p3732:
After having another go at this, I am fairly sure I can not fix the test for this plugin with the current test suite, since that apparently requires creating test files outside of /tmp. I couldn't find any other test that does this and it seemed messy to me.

So instead I just disabled the added feature in the test. I know that is not ideal, but since I am out of ideas, I hope it will be okay.

p3732:
The remaining failed tests result in this not having the current master merged in. If desired I can create a new MR based on current master though.

sampsyo:
Awesome; thanks for coming back around to this! Indeed, using /tmp for permissions tests may not be possible…

Anyway, yes, rebasing or merging would be awesome. You can also use this same PR—if you `git rebase` and then `git push -f`, this ticket should get updated.

p3732:
Should be all good now :)

sampsyo:
Cool; seems like this is in order. :smiley: The only things we need now are documentation for the new option in the plugin and a quick entry in the changelog.

p3732:
Okay, turns out I donked up with git and this does not even include the changes, so I am just gonna close this MR and open one from a new branch…

Issue #3070, opened 2018-11-01
beets absubmit: TypeError: the JSON object must be str, not 'bytes'


gorgobacka:
### Problem

Running beets absubmit produced the following error.

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv absubmit
user configuration: /home/noname/.config/beets/config.yaml
data directory: /home/noname/.config/beets
plugin paths: 
Sending event: pluginload
inline: adding item field multidisc
inline: adding item field bool_mp3
inline: adding item field custom_catalog
library database: /home/xxxxxxx/imported/musiclibrary.blb
library directory: /home/xxxxxxx/imported
Sending event: library_opened
Traceback (most recent call last):
  File "/usr/local/bin/beet", line 11, in <module>
    load_entry_point('beets==1.4.7', 'console_scripts', 'beet')()
  File "/usr/local/lib/python3.5/dist-packages/beets/ui/__init__.py", line 1256, in main
    _raw_main(args)
  File "/usr/local/lib/python3.5/dist-packages/beets/ui/__init__.py", line 1243, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "/usr/local/lib/python3.5/dist-packages/beetsplug/absubmit.py", line 108, in command
    analysis = self._get_analysis(item)
  File "/usr/local/lib/python3.5/dist-packages/beetsplug/absubmit.py", line 137, in _get_analysis
    analysis = json.load(tmp_file)
  File "/usr/lib/python3.5/json/__init__.py", line 268, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "/usr/lib/python3.5/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'
```

### Setup

* OS: Linux Mint
* Python version: 3.5.2
* beets version: 1.4.7
* Turning off plugins made problem go away (yes/no): no

It's do to this PR: #2673.

This change works for me:
```python
with open(filename, 'r') as tmp_file:
```

sampsyo:
Hi! That looks bad—can you please open a pull request with that change so we can try it out?

gorgobacka:
A newer version of Python fixed it.

PR #3071, opened 2018-11-02
Bugfix for absubmit (#3070)

gorgobacka:


sampsyo:
Thanks! To quote [the Python docs for the json module](
https://docs.python.org/3.7/library/json.htm):

> Changed in version 3.6: fp can now be a binary file. The input encoding should be UTF-8, UTF-16 or UTF-32.

So the reason you’re seeing this crash is that you’re using Python 3.5 (IIRC); later versions of Python (as well as Python 2.x) do not have this issue.

If we’re going to open the file in text mode (as this PR implements), we will need to choose a specific encoding—leaving it up to Python to pick the encoding runs the risk of it defaulting to ASCII, which will reintroduce the same bug that the original change was meant to fix. Can you please try forcing UTF-8 (if you’re interested in staying on 3.5) and making sure that still works for you?

gorgobacka:
I will close the PR, because it works with a newer version of Python. And there is no need to fix it.

Issue #3073, opened 2018-11-04
Question about pipeline.Queue init method


bartkl:
I'm playing around with websockets with the ultimate goal of building a web importer frontend for the beets importer process. Currently I'm trying to get a proof of concept to work using Flask-SocketIO with eventlet. For this it's necessary to monkey patch several stdlib modules to 'green' them so they're compatible with eventlet.

However, when I perform the monkey patch I get an error:
```
    queues = [CountedQueue(queue_size) for i in range(queue_count)]
  File "/opt/cucumba/venv/lib/python3.6/site-packages/beets/util/pipeline.py", line 95, in __init__
    queue.Queue.__init__(self, maxsize)
  File "/opt/cucumba/venv/lib/python3.6/site-packages/eventlet/green/Queue.py", line 15, in __init__
    super(Queue, self).__init__(maxsize)
TypeError: super(type, obj): obj must be an instance or subtype of type
```
After some analysis and trial and error, I got things to work by altering the init method of `CountedQueue` in pipeline.py:

```diff
95c95
<         queue.Queue.__init__(self, maxsize)
---
>         super(CountedQueue, self).__init__(maxsize)
```

That's nice, but I have a few questions still:
1. Is there any good reason why the `super()` call was not used in the first place, as seems general practice in the file (and community?)
2. I honestly still don't precisely understand the exact differences in the original call and my replaced `super()` call. Could someone elaborate on that?
3. I don't understand why the original situation leads to the specific error.
4. Is this adjustment okay for the pipeline (somewhere in the future)?

Trying to learn and contribute :). 

Thanks,
Bart

sampsyo:
Wow! This seems like an incredibly cool project! I'm really excited to see where you go with it.

1. I don't see any reason why I didn't use `super` there. This change of yours looks like a clear improvement. :smiley:
2. They should be exactly the same—except when you change the base class. That is, `super(CountedQueue, self)` gets a "version" of `self` with methods that look up in a different starting class. That does the same thing as calling methods explicitly on the base class as named.
3. The trouble is that "greening" the library changed the base class of `CountedQueue`, but the call to the super class's `__init__` method wasn't aware of that. So it ended up calling `__init__` in the "wrong" class.
4. Yes! Please go right ahead and open a pull request with this change by itself; we should merge it right away.

Thanks!!

sampsyo:
Sure; sounds good.

bartkl:
[...]

PR #1957, opened 2016-04-25
Allow force-quitting the pipeline

wisp3rwind:
The situation: I was running an import task, that apparently locked up, i.e. it didn't advance for hours, even though not all user queries were run. Replaygain was enabled, but for 20-ish albums, even that should not take as long. Worse, I couldn't quit beets per `KeyboardInterrupt` and as a result didn't end up with a traceback when I eventually `kill`-ed beets. Not very satisfying, because I couldn't find out where it actually got stuck.

(The actual culprit may have been related to a flaky USB connection to the external drive with the music.)

Therefore, this is my attempt at allowing to quit at any time, safeguarded by having to press `Ctrl-C` rapidly several times (and a warning about data loss). What do you think? Worth merging, or better off as a developer tool in a separate branch?

I think at least the first commit should go in, as it fixes a few catch-all `except:`s. According to `grep`, the only remaining catch-alls are now in `beets/util/bluelet.py`, which I dared not touch. Also, it is only used by the `bpd` plugin (again, says `grep`), and might already be properly handled (I did not read the `bluelet` code).

This tackles https://github.com/beetbox/beets/issues/803 , but does not actually solve the problem (Maybe I'll revive that issue).

Issue #803, opened 2014-06-02
Allow interrupting safe import tasks


Hawke:
Currently some long-running tasks cannot be interrupted by keyboard Ctrl+C. (e.g. 'chroma' plugin import scanning.

It would be nice to allow this to happen, assuming the task in question can be interrupted (e.g. chroma should be entirely read-only and should always be safe)

FichteFoll:
I suppose rewriting beets asynchronously is not an option for the foreseeable future? At least auto-matching albums during import should be interruptable.

PR #3075, opened 2018-11-05
altered `super` call in `CountedQueue` init method (see #3073)

bartkl:
The way how the `CountedQueue.__init__` called its parent class's init method gave problems when trying to green the `threading` library using Eventlet.

See #3073 for the original discussion and purpose.

### DIFF
--- a/beets/util/pipeline.py
+++ b/beets/util/pipeline.py
@@ -92,7 +92,7 @@ class CountedQueue(queue.Queue):
     finished with the queue.
     """
     def __init__(self, maxsize=0):
-        queue.Queue.__init__(self, maxsize)
+        super(CountedQueue, self).__init__(maxsize)
         self.nthreads = 0
         self.poisoned = False
 


bartkl:
Apparently, the Python 2.7.13 checks give this error:

``
`TypeError: super() argument 1 must be type, not classobj
```

I honeslty don't know why; the other classes call their super inits the same way.

Can someone help out? Thanks.

**Comment by NoahTheDuke** [comment]
[This](https://stackoverflow.com/a/1713052/3023252) is why! Queue is probably an old style class, so try changing it to the new style and see if that fixes the issue.

**Comment by sampsyo** [comment]
Aha! That would explain it—looks like `Queue` is what's called an "old-style class," an unfortunate wart of Python 2.x that has many problems, including a lack of support for `super`.

The easiest way to do this might be to use something like this:

```
if six.PY2:
    queue.Queue.__init__(self, maxsize)
else:
    super(CountedQueue, self).__init__(maxsize)
```

which will at least make your trick work on Python 3+.

wisp3rwind:
On a side note, I've recently been looking into cleaning up the pipeline code a little and revisiting the Queue invalidation in particular (in the context of #1957 and related issues). That being said, my current impression is that it's probably easiest to copy Python's Queue implementation to beets and modify it to our needs.

If I do indeed find the time to make a PR for this, it would also get rid of the new/old-style class problem.

sampsyo:
Interesting, @wisp3rwind! I'd be interested to hear more—reimplementing `threading.Queue` in our own code sounds like a pretty major step. Is there a way to summarize what you'd change about it?

wisp3rwind:
> Interesting, @wisp3rwind! I'd be interested to hear more—reimplementing `threading.Queue` in our own code sounds like a pretty major step. Is there a way to summarize what you'd change about it?

In fact, the code for it is not that complex, and it could be stripped down significantly. The "cleanup" for now is a local branch that removes some code duplication in `PipelineStage` and subclasses and those changes to the queues. The point with the latter was to get rid of the [hack](https://github.com/beetbox/beets/blob/b380a4c9a37640f93b02c3277af5e76d4e8567e9/beets/util/pipeline.py#L48) that is used to invalidate  queues on abort.

The larger picture is that I'd really like to implement [this proposal](https://github.com/beetbox/beets/issues/803#issuecomment-44906339) with the end goal of being more robust on both intentionally aborted and crashed imports. Where robust is supposed to mean that it's as unlikely as possible that incomplete imports remain in the library (where completeness includes all plugin stages having run), and if that cannot be guaranteed in some cases, that at least the affected import tasks/items are logged. Which in turn would require pretty tight control over queues and pipeline stages, which with the current code is difficult to say the least.

Motivation is, even if not as relevant to me anymore right now, that I was bitten by incomplete imports when I initially started using beets and had aborted several imports (due to the kernel bug that made the ntfs3g driver block forever). While for most things this is not so important (it's easy to run fetchart and replaygain manually to fix up the incomplete items, and there are incremental imports (which might not be granular enough, though)) other errors might go unnoticed. For example, I use the third-party copyartifacts plugin that only copies files at the very end of the import.

sampsyo:
Aha! Thanks for the detailed explanation. And thanks for looking into this thorny issue!

PR #3084, opened 2018-11-28
Export plugin can now export path from library.

domenn:
Exports path (file path) if specified in fields to export (with -i parameter). Only when exporting library fields, because file tags don't contain path. Info works as before. Path won't be displayed in the list even if user provides it in parameters.

### DIFF
--- a/beetsplug/info.py
+++ b/beetsplug/info.py
@@ -66,7 +66,6 @@ def library_data(lib, args):
 def library_data_emitter(item):
     def emitter():
         data = dict(item.formatted())
-        data.pop('path', None)  # path is fetched from item
 
         return data, item
     return emitter
@@ -184,6 +183,8 @@ class InfoPlugin(BeetsPlugin):
         included_keys = []
         for keys in opts.included_keys:
             included_keys.extend(keys.split(','))
+        # Drop path even if user provides it multiple times
+        included_keys = [k for k in included_keys if k != 'path']
         key_filter = make_key_filter(included_keys)
 
         first = True
@@ -239,4 +240,5 @@ def make_key_filter(include):
 
 
 def identity(val):
+    val.pop('path', None)  # path is fetched from item
     return val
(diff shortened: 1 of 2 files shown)


beetsplug/info.py
```diff
@@ -184,6 +183,7 @@ def run(self, lib, opts, args):
         included_keys = []
         for keys in opts.included_keys:
             included_keys.extend(keys.split(','))
+        included_keys = [k for k in included_keys if k != 'path'] # Drop path even if user provides it multiple times
```
jackwilsdon: [This line exceeds our line length limit of 79 characters](https://travis-ci.org/beetbox/beets/jobs/460820041#L1086), I'd suggest moving the comment onto the preceding line;

```suggestion
        # Drop path even if user provides it multiple times
        included_keys = [k for k in included_keys if k != 'path']
```

beetsplug/info.py
```diff
@@ -239,4 +239,5 @@ def filter_(data):
 
 
 def identity(val):
+    val.pop('path', None) # path is fetched from item
```
jackwilsdon: [We require two spaces before every inline comment](https://travis-ci.org/beetbox/beets/jobs/460820041#L1087);

```suggestion
    val.pop('path', None)  # path is fetched from item
```

domenn:
Ok. I see I need to better check the guidelines next time.
Do you like better if I submit changes as new commit or amend current one, to reduce number of commits for small changes like this one?

jackwilsdon:
I personally prefer amending the current commit (it leads to a cleaner history IMO), but it's up to you :+1:

domenn:
I agree. Not sure yet how this is usually done. So I asked

beetsplug/info.py
```diff
@@ -239,4 +240,5 @@ def filter_(data):
 
 
 def identity(val):
+    val.pop('path', None)  # path is fetched from item
```
sampsyo: It's a little odd to call this function `identity` when it's not really an identity function. :smiley: I'm going to fiddle around with this a little bit…

sampsyo:
OK, merged with a small refinement in the above commit. Let me know if anything looks wrong, please!

domenn:
Thanks. I will check within two days and let you know if there are problems.

PR #3085, opened 2018-11-28
Allow exiting object selection early

### LINKED ISSUE #3083: Selection interface: option to "skip remaining tracks" and take action
### Problem

I love the `modify` command's select mode, because it means you don't have to write super-specific queries - instead you can write a more general query and then simply type an `n` to skip those tracks you don't want to retag.

Unfortunately, there seems to be no way to "skip remaining tracks". For example, say I got through 50 out of 100 tracks that a query returned. I've answered `y` to most of them, but `n` to just a few. If I hit Ctrl-C at this point, I notice that the tags do not get written to the tracks I answered `y` to.

Perhaps we could add an `s` option that skips remaining tracks but still updates those that we answered `y` to. Alternatively, we could update tracks as-and-when we answer `y`, allowing us to Ctrl-C out of the process without losing work.

jackwilsdon:
Fixes #3083.

I've updated the CLI documentation to mention the new `q` option, but I'm not exactly a wordsmith 😁 let me know if you can think of some better phrasing!

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -408,9 +408,14 @@ def input_select_objects(prompt, objs, rep):
         out = []
         for obj in objs:
             rep(obj)
-            if input_yn(u'%s? (yes/no)' % prompt, True):
+            answer = input_options(
+                ('y', 'n', 'q'), True, u'%s? (yes/no/quit)' % prompt,
+                u'Enter Y or N:'
+            )
+            if answer == u'y':
                 out.append(obj)
-            print()  # go to a new line
+            elif answer == u'q':
+                return out
         return out
 
     else:  # No.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -46,6 +46,8 @@ New features:
   example, ``beet modify -a artist:beatles artpath!`` resets ``artpath``
   attribute from matching albums back to the default value.
   :bug:`2497`
+* Modify selection can now be applied early without selecting every item.
+  :bug:`3083`
 
 Changes:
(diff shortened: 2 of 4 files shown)


sampsyo:
Awesome! This implementation looks good, and I think the documentation text is perfect as well.

There's just one small thing that concerns me: whether the word "quit" alone is enough to suggest the right behavior to the user. I can imagine people seeing that for the first time and imagining that it will cancel the whole operation rather than applying to all the items that have been selected so far.

The thing is, that might be fine—because if someone intends to abort the whole process, they probably will do it on the first prompt, when the behavior is identical to "quit." And given that I don't have a great idea for an _alternative_ to "quit," maybe we should just roll with it! But I thought I'd get your thoughts on the matter before we hit the green button.

Other than that reservation, I think this is good to go!

sampsyo:
There's always something excessively verbose like "skip remaining," I guess! But you're right; "quit" is probably the best bet. :+1:

jackwilsdon:
What do you think about changing it to `cancel`?

```Text
$ beet modify title="New Title"
Modifying 1 items.
Artist - Album - Title
  title: Title -> New Title
Really modify, move and write tags? (Yes/no/select) s   

Artist - Album - Title
  title: Title -> New Title
Really modify, move and write tags? (yes/no/cancel) c
```

sampsyo:
Yes indeed, I think "quit" is a little better than "cancel." Merging! Thank you for your help on this (and for thinking through the wording with me)!! :sparkles: :rocket:

PR #3088, opened 2018-12-05
Add reflink option

### LINKED ISSUE #2642: Copy on Write/reflink for supported filesystems

rubdos:
(Would fix #2642) Rebased #2720 on master, thus supersedes #2720.

Outstanding discussion points:

- [ ] what to do with exception handling?

@sampsyo 
> It might be useful to see the evolving discussion in #2785, but I think raising a custom exception or a FilesystemError from the utility function itself is the right thing for util. Then, to actually show the problem to the user—if we want to stop the program altogether—higher levels in the stack should probably catch this at an opportune moment and raise a UserError.
>
> Of course, if we wanted the same behavior as for trying to write to a read-only filesystem, for example, then just keeping it as a FilesystemError would be the right thing.

I will do the exception thing the coming hours.

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -223,7 +223,8 @@ class ImportSession(object):
             iconfig['incremental'] = False
 
         if iconfig['reflink']:
-            iconfig['reflink'] = iconfig['reflink'].as_choice(['auto', True, False])
+            iconfig['reflink'] = iconfig['reflink'] \
+                .as_choice(['auto', True, False])
 
         # Copy, move, reflink, link, and hardlink are mutually exclusive.
         if iconfig['move']:
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -222,19 +222,30 @@ class ImportSession(object):
             iconfig['resume'] = False
             iconfig['incremental'] = False
 
-        # Copy, move, link, and hardlink are mutually exclusive.
+        if iconfig['reflink']:
+            iconfig['reflink'] = iconfig['reflink'].as_choice(['auto', True, False])
+
+        # Copy, move, reflink, link, and hardlink are mutually exclusive.
         if iconfig['move']:
             iconfig['copy'] = False
             iconfig['link'] = False
             iconfig['hardlink'] = False
+            iconfig['reflink'] = False
         elif iconfig['link']:
             iconfig['copy'] = False
             iconfig['move'] = False
             iconfig['hardlink'] = False
+            iconfig['reflink'] = False
         elif iconfig['hardlink']:
             iconfig['copy'] = False
             iconfig['move'] = False
             iconfig['link'] = False
+            iconfig['reflink'] = False
+        elif iconfig['reflink']:
+            iconfig['copy'] = False
+            iconfig['move'] = False
+            iconfig['link'] = False
+            iconfig['hardlink'] = False
 
         # Only delete when copying.
         if not iconfig['copy']:
@@ -707,7 +718,7 @@ class ImportTask(BaseImportTask):
             item.update(changes)
 
     def manipulate_files(self, operation=None, write=False, session=None):
-        """ Copy, move, link or hardlink (depending on `operation`) the files
+        """ Copy, move, link, hardlink or reflink (depending on `operation`) the files
         as well as write metadata.
 
         `operation` should be an instance of `util.MoveOperation`.
@@ -1536,6 +1547,8 @@ def manipulate_files(session, task):
             operation = MoveOperation.LINK
         elif session.config['hardlink']:
             operation = MoveOperation.HARDLINK
+        elif session.config['reflink']:
+            operation = MoveOperation.REFLINK
         else:
             operation = None
 
--- a/beets/library.py
+++ b/beets/library.py
@@ -756,11 +756,7 @@ class Item(LibModel):
             plugins.send("item_reflinked", item=self, source=self.path,
                          destination=dest)
         else:
-            plugins.send("before_item_moved", item=self, source=self.path,
-                         destination=dest)
-            util.move(self.path, dest)
-            plugins.send("item_moved", item=self, source=self.path,
-                         destination=dest)
+            assert False, 'unknown MoveOperation'
 
         # Either copying or moving succeeded, so update the stored path.
         self.path = dest
@@ -1106,7 +1102,7 @@ class Album(LibModel):
         elif operation == MoveOperation.REFLINK_AUTO:
             util.reflink(old_art, new_art, fallback=True)
         else:
-            util.move(old_art, new_art)
+            assert False, 'unknown MoveOperation'
         self.artpath = new_art
 
     def move(self, operation=MoveOperation.MOVE, basedir=None, store=True):
--- a/beets/library.py
+++ b/beets/library.py
@@ -747,6 +747,20 @@ class Item(LibModel):
             util.hardlink(self.path, dest)
             plugins.send("item_hardlinked", item=self, source=self.path,
                          destination=dest)
+        elif operation == MoveOperation.REFLINK:
+            util.reflink(self.path, dest, fallback=False)
+            plugins.send("item_reflinked", item=self, source=self.path,
+                         destination=dest)
+        elif operation == MoveOperation.REFLINK_AUTO:
+            util.reflink(self.path, dest, fallback=True)
+            plugins.send("item_reflinked", item=self, source=self.path,
+                         destination=dest)
+        else:
+            plugins.send("before_item_moved", item=self, source=self.path,
+                         destination=dest)
+            util.move(self.path, dest)
+            plugins.send("item_moved", item=self, source=self.path,
+                         destination=dest)
 
         # Either copying or moving succeeded, so update the stored path.
         self.path = dest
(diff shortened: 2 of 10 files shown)


jtpavlock:
This seems like a nice temporary solution until [python support is added upstream](https://bugs.python.org/issue37157). As for trying to continue this, I think a good first step is to rebase off master and fix any conflicts.

rubdos:
> This seems like a nice temporary solution until [python support is added upstream](https://bugs.python.org/issue37157).

Had no clue that Python was doing that. I added it to the README of `pyreflink`, thanks!

I'm still very interested in this (I'm quite sure I'm running this on my NAS atm), so if I should rebase this, I'll gladly do so.

jtpavlock:
> if I should rebase this, I'll gladly do so

Yes, please! We have a new CI, so it would be helpful to make sure everything still passes with a rebase. Also, fixing any merge conflicts as well. In the meantime, I'll try to give this another lookover. Was there anything outstanding you still wanted to fix or discuss?

rubdos:
Rebase pushed.

It's been a year and a half again, but I recall there was some error handling to discuss. I think some of the discussion was in #2720.

rubdos:
Looks like the CI found some flakey (get it?) stuff. Also fixed in a separate commit. Maybe it would be nice to squash some things together before merging, I'm not sure how keen you folks are on a clean git history.

setup.py
```diff
@@ -159,6 +160,7 @@ def build_manpages():
         'scrub': ['mutagen>=1.33'],
         'bpd': ['PyGObject'],
         'replaygain': ['PyGObject'],
+        'reflink': ['reflink'],
```
jtpavlock, line 163: I don't think this is needed since it's not an optional dependency for a plugin

docs/reference/config.rst
```diff
@@ -476,13 +476,34 @@ hardlink
...
+Defaults to ``no``.
+
+This kind of clone is only available on certain filesystems: for example,
+btrfs and APFS. For more details on filesystem support, see the `pyreflink`_
+documentation.
+
+The option is ignored if ``move`` is enabled (i.e., beets can move or
+copy files but it doesn't make sense to do both).
+
+.. _reflink clone: https://blogs.oracle.com/otn/save-disk-space-on-linux-by-cloning-files-on-btrfs-and-ocfs2
```
jtpavlock, line 504: Is this used anywhere?

beets/library.py
```diff
@@ -747,6 +747,20 @@ def move_file(self, dest, operation=MoveOperation.MOVE):
             util.hardlink(self.path, dest)
             plugins.send("item_hardlinked", item=self, source=self.path,
                          destination=dest)
+        elif operation == MoveOperation.REFLINK:
+            util.reflink(self.path, dest, fallback=False)
+            plugins.send("item_reflinked", item=self, source=self.path,
+                         destination=dest)
+        elif operation == MoveOperation.REFLINK_AUTO:
+            util.reflink(self.path, dest, fallback=True)
+            plugins.send("item_reflinked", item=self, source=self.path,
```
jtpavlock, line 756: Maybe only send if the reflink was successful? I am thinking of the case where it fails, and a copy is used instead. This signal would seem to indicate a reflink was created when it actually was not.

beets/library.py
```diff
@@ -747,6 +747,20 @@ def move_file(self, dest, operation=MoveOperation.MOVE):
...
+                         destination=dest)
+        else:
```
jtpavlock, line 758: Do you remember the reason the move stuff needed to be added? (this and line 1108)

I'm not too familiar with this module so I'm wondering how/if it pertains to the reflink logic.

setup.py
sampsyo: Because it’s a conditional import (inside the relevant function instead of at the top of a module), I actually think it’s OK to be an optional dependency.

jtpavlock: so then should it be removed from the normal dependencies?

https://github.com/beetbox/beets/pull/3088/files/99cd7e2de457af1b05a2974fee975b48f15518dc#diff-2eeaed663bd0d25b7e608891384b7298R126

sampsyo: Ah yes, good point.

rubdos: The line you refer to is in the `test` section, because it's required for some test cases, I think. Or am I looking at something else?

docs/reference/config.rst
rubdos: I've made "file clones" use that link in a new commit. If you'd rather have that link removed, that's fine too.

beets/library.py
rubdos: Hmm, I see why, but that will complicate things even more, I think. I notice most of the other operations have a `before_` signal too, and maybe it might mean `before_item_reflinked` would go before a `item_copied`.

Not sure how to proceed here.  `util.reflink` does not return whether it had to fall-back or not either, but that's an easy addition.

beets/library.py
rubdos: I have no clue any more! Maybe it's an rebase artefact of a previous attempt, where `move` was the fallback in those methods? If you, nor @sampsyo see why this is there, maybe it should just go.

On the other hand, are we sure that the `operation` variable is checked exhaustively?

beets/importer.py
```diff
@@ -222,19 +222,31 @@ def set_config(self, config):
             iconfig['resume'] = False
             iconfig['incremental'] = False
 
-        # Copy, move, link, and hardlink are mutually exclusive.
+        if iconfig['reflink']:
+            iconfig['reflink'] = iconfig['reflink'] \
+                .as_choice(['auto', True, False])
```
sampsyo, line 227: This looks a little odd—unless I'm mistaken, `iconfig` is just a flat dictionary rather than a Confuse view, right? So I'm not sure `.as_choice` is supported on it. (Maybe I'm missing something obvious?)

beets/library.py
sampsyo: Yeah, I think this is OK as is for now.

beets/library.py
sampsyo: Interesting! I'm similarly stumped. I say we replace this with:

```py
else:
    assert False, 'unknown MoveOperation'
```

…to check exhaustiveness.

beets/library.py
```diff
@@ -1087,6 +1101,12 @@ def move_art(self, operation=MoveOperation.MOVE):
             util.link(old_art, new_art)
         elif operation == MoveOperation.HARDLINK:
             util.hardlink(old_art, new_art)
+        elif operation == MoveOperation.REFLINK:
+            util.reflink(old_art, new_art, fallback=False)
+        elif operation == MoveOperation.REFLINK_AUTO:
+            util.reflink(old_art, new_art, fallback=True)
+        else:
+            util.move(old_art, new_art)
```
sampsyo, line 1109: As above, let's `assert False` here.

setup.py
sampsyo: Right again. Thanks, @rubdos.

jtpavlock: Ah okay, my mistake. In that case, maybe it should be noted in the docs that you need to install reflink? Either `python -m pip install beets[reflink]` or `python -m pip install reflink`

docs/reference/config.rst
jtpavlock: Looks good, just wanted to make sure you didn't forget to use it or something 😄

beets/library.py
jtpavlock: yeah, my initial thought was something like:

```
        elif operation == MoveOperation.REFLINK_AUTO:
            try:
                util.reflink(self.path, dest)
            except (NotImplementedError, pyreflink.ReflinkImpossibleError):
                self.move_file(self, dest, MoveOperation.COPY)
            else:
                plugins.send(
                    "item_reflinked", item=self, source=self.path, destination=dest
                )
```
removing the need for the fallback parameter

beets/importer.py
rubdos: I'm quite unsure here, I think I copied that from a few lines below, but then, this patch is kind of old. Three years or so :joy:

I don't really recall how this config method is supposed to work. FWIW, I've had
```yaml
import:
  reflink: auto
```
in my laptop's beets config on this branch for the last year or two, but then this config is also a transcoding setup, where reflink doesn't make sense at all.

beets/library.py
rubdos: asserts are handled in the new commit

setup.py
rubdos: Added in the config-docs!

sampsyo (approved):
Yay; this looks perfect! Thank you for the long-term effort, @rubdos.

PR #3089, opened 2018-12-05
Performance improvement: Type cast properties on demand

pprkut:
I analyzed where time is spent when querying the library and found that most of it is actually type casting properties after retrieving them from the database. However, this isn't really necessary. We hardly ever need access to all properties, so it makes much more sense to delay the actual casting until a property is accessed.

Taking my previous baseline of `3.66s` for master, doing the delayed type casting reduces this to `1.96s`, or down to `1.45s` when #2798 is also applied (a bit slower now than my first number with it now no longer being a poc).
Our of those remaining ~1.5s, about 0.5s is spent on the actual database queries, and the rest is object creation.

I'm open to ideas for further performance improvements here, but we get firmly into area of micro-optimizations (creating 11000 objects in 1s, we'd need to make every object creation 1/22000s faster to save another 0.5s). At least I didn't find any obvious candidates.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -177,6 +177,8 @@ class Model(object):
         """
         self._db = db
         self._dirty = set()
+        self._raw_values_fixed = {}
+        self._raw_values_flex = {}
         self._values_fixed = {}
         self._values_flex = {}
 
@@ -193,9 +195,9 @@ class Model(object):
         """
         obj = cls(db)
         for key, value in fixed_values.items():
-            obj._values_fixed[key] = cls._type(key).from_sql(value)
+            obj._raw_values_fixed[key] = value
         for key, value in flex_values.items():
-            obj._values_flex[key] = cls._type(key).from_sql(value)
+            obj._raw_values_flex[key] = value
         return obj
 
     def __repr__(self):
@@ -232,7 +234,9 @@ class Model(object):
         """
         new = self.__class__()
         new._db = self._db
+        new._raw_values_fixed = self._raw_values_fixed.copy()
         new._values_fixed = self._values_fixed.copy()
+        new._raw_values_flex = self._raw_values_flex.copy()
         new._values_flex = self._values_flex.copy()
         new._dirty = self._dirty.copy()
         return new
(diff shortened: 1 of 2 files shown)


sampsyo:
Nice!! This seems like a really good optimization target. Excellent work with that profiler. :smiley:

I have a small idea for how to make the code a little more centralized and perhaps more maintainable: define a little "dict-like" wrapper that is responsible for performing the lazy conversion. Here's a quick sketch I put together:

```py
class LazyConvertDict(object):
    def __init__(self, data, model_cls):
        self.data = data
        self.model_cls = model_cls
        self._converted = {}

    def _convert(key, value):
        return self.model_cls._type(key).from_sql(value)

    def __getitem__(self, key):
        if key in self._converted:
            return self._converted[key]
        elif key in self.data:
            value = self._convert(key, self.data[key])
            self._converted[key] = value
            return value
```

The model class could then just define `_values_fixed` and `_values_flex` to be instances of `LazyConvertDict` and access them normally—eliminating the need for the conversion handling inside of the lookup functions.

This would probably need a `__setitem__` method too, but does the rough idea make sense?

pprkut:
Uploaded a raw version that uses your suggested LazyConvertDict. Testing here looks fine. Performance for some reason seems to be even better with that version, not entirely sure why.
Please have a look. I'll perform cleanup, doc additions, test fixes within the next couple days :-)

beets/dbcore/db.py
```diff
@@ -88,6 +88,77 @@ def _get_formatted(self, model, key):
         return value
 
 
+class LazyConvertDict(object):
```
sampsyo: A docstring describing the functionality here would be nice!

beets/dbcore/db.py
```diff
@@ -177,8 +248,8 @@ def __init__(self, db=None, **values):
         """
         self._db = db
         self._dirty = set()
-        self._values_fixed = {}
-        self._values_flex = {}
+        self._values_fixed = LazyConvertDict(self);
+        self._values_flex = LazyConvertDict(self);
```
sampsyo: No semicolons necessary in Python. :smiley: :snake:

pprkut: Hah, force of habit ;-)

pprkut:
AFAICS everything's fixed up now :)

sampsyo:
OK, this is truly awesome!! We should be maintaining a trophy room somewhere that lists all the performance bottlenecks you have found and defeated. Thank you for all your help along these lines! :sparkles:

PR #3090, opened 2018-12-06
Kodi plugin equivalent of Scan-Item-To-Library

velvet-jones:
I've modified kodiupdate.py to allow adding two new configuration settings that support only scanning the album that was imported. Without the new configuration settings, the original behavior is retained (re-scan entire music library at exit).

I'm unsure about the decode('utf-8') when stripping off the last part of the path to the imported album. It works with my system when I encounter funky characters (Linux), but I'm not sure iit would work on Windows since the file system encoding there is not utf-8. Any ideas on how to process non-ascii characters in a cross-platform manner would be appreciated.

### DIFF
--- a/beetsplug/kodiupdate.py
+++ b/beetsplug/kodiupdate.py
@@ -25,19 +25,19 @@ to re-scan your entire library after importing one or more albums:
         pwd: secret
 
 Alternatively, you can choose to only scan each newly imported album directory.
-To do so, add all the config.yaml settings above, but also add 'source' and
-'library' settings that look something like this:
+To do so, add all the config.yaml settings above, but also add `source` and
+`library` settings that look something like this:
 
        source: nfs://myserver.local/music/library/
        library: /home/music/library/
 
-The value for 'source' should be the Kodi Music source found in
+The value for `source` should be the Kodi Music source found in
 .kodi/userdata/sources.xml.
 
-The value for 'library' should be the path to your beets library.
+The value for `library` should be the path to your beets library.
 
-After an album is imported, this plugin strips off the 'library' portion of
-the album path and appends the remaining portion to the 'source', then issues
+After an album is imported, this plugin strips off the `library` portion of
+the album path and appends the remaining portion to the `source`, then issues
 a Kodi update for that path.
 
 """
@@ -52,18 +52,18 @@ import os
 
 def update_kodi(host, port, user, password, path=None):
     """Sends request to the Kodi api to start a library refresh.
-       If 'path' is provided, only refresh that path.
+    If `path` is provided, only refresh that path.
     """
     url = "http://{0}:{1}/jsonrpc".format(host, port)
 
     """Content-Type: application/json is mandatory
-    according to the kodi jsonrpc documentation"""
+    according to the kodi jsonrpc documentation."""
 
     headers = {'Content-Type': 'application/json'}
 
     # Create the payload. Id seems to be mandatory.
     payload = {'jsonrpc': '2.0', 'method': 'AudioLibrary.Scan', 'id': 1}
-    if path is not None:
+    if path:
         payload['params'] = {'directory': path}
     r = requests.post(
         url,
@@ -89,17 +89,17 @@ class KodiUpdate(BeetsPlugin):
 
         config['kodi']['pwd'].redact = True
         if config['source'] == '':
-            # rescan entire library
+            # Re-scan the entire library.
             self.register_listener('database_change',
                                    self.listen_for_db_change)
         else:
-            # only rescan the path to this album
+            # Scan only the path to this album.
             self.register_listener('album_imported',
                                    self.album_imported)
 
     def listen_for_db_change(self, lib, model):
         """Listens for beets db change, waits for cli exit,
-        then registers the update"""
+        then registers the update."""
         self.register_listener('cli_exit', self.cli_exit)
 
     def album_imported(self, lib, album):
--- a/beetsplug/kodiupdate.py
+++ b/beetsplug/kodiupdate.py
@@ -25,17 +25,20 @@ to re-scan your entire library after importing one or more albums:
         pwd: secret
 
 Alternatively, you can choose to only scan each newly imported album directory.
-To do so, add all the config.yaml settings above, but also add 'source' and 'library'
-settings that look something like this:
+To do so, add all the config.yaml settings above, but also add 'source' and
+'library' settings that look something like this:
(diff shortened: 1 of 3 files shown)


velvet-jones:
Thanks for the feedback! Let me know if this needs any further changes.

docs/plugins/kodiupdate.rst
```diff
@@ -16,6 +17,27 @@ which looks like this::
         user: kodi
         pwd: kodi
 
+With the basic configuration settings above, the ``kodiupdate`` plugin
+will notify Kodi to scan your entire music library after the import
+operation has completed. For large music libraries you may prefer to have
```
sampsyo: It's not _quite_ the complete truth to say that the default configuration sends the request "after the import operation has completed"—it actually does this every time _anything_ changes in your library. For example, this happens after a `beet modify` where you update the metadata for a track or album.

docs/plugins/kodiupdate.rst
```diff
@@ -16,6 +17,27 @@ which looks like this::
...
+Kodi only scan the directory where each album was imported. To enable
+this behavior, add `source` and `library` settings, like this:
+
+    kodi:
+        host: localhost
+        port: 8080
+        user: kodi
+        pwd: kodi
+        source: nfs://myserver.local/music/library/
+        library: /home/music/library/
```
sampsyo: Could the plugin instead just use the global `directory` configuration value for beets? (Would this ever need to be different from that?)

docs/plugins/kodiupdate.rst
```diff
@@ -16,6 +17,27 @@ which looks like this::
         user: kodi
         pwd: kodi
 
+With the basic configuration settings above, the ``kodiupdate`` plugin
+will notify Kodi to scan your entire music library after the import
+operation has completed. For large music libraries you may prefer to have
+Kodi only scan the directory where each album was imported. To enable
+this behavior, add `source` and `library` settings, like this:
```
sampsyo: Can you add a brief description here of what exactly `source` is supposed to be? There's an example below, but a few words saying what it is in general (in terms Kodi users will understand) could be really useful.

beetsplug/kodiupdate.py
```diff
@@ -61,27 +83,51 @@ def __init__(self):
...
-
-    def update(self, lib):
-        """When the client exists try to send refresh request to Kodi server.
+        """Listens for beets db change, waits for cli exit,
+        then registers the update."""
+        self.register_listener('cli_exit', self.cli_exit)
+
+    def album_imported(self, lib, album):
+        source = config['kodi']['source'].get()
+        library = config['kodi']['library'].get()
```
sampsyo: Here's how you get the `directory` configuration value:

```py
from beets import config
config['directory'].as_filename()
```

docs/plugins/kodiupdate.rst
velvet-jones: You're right, the default behavior is more general, but it's also more time consuming on a large library. For certain operations such as an import, it's much faster to process the few albums imported than re-scan the entire library. Maybe I should also look at getting notifications for beets modify? (Is there such a notification?)

docs/plugins/kodiupdate.rst
velvet-jones: It's not well documented what the "library" config option *really* does, partly because I was trying to figure out how to describe it correctly. It has almost no bearing on the beets library path in a literal sense - it's used to remove the first part of the path from the album, which results in a string that is appended to the Kodi source. The beets library location could very well be different from the "library" configuration option. For example:

Kodi music source: nfs://192.168.3.5/home/public/music/
beets library: /home/public/music/library

After album import let's say the album path is "/home/public/music/library/Prince/Controversy".

In this scenario, "library" config should be "/home/public/music/". We strip that from the album path above, leaving: "library/Prince/Controversy". We then append the result to the Kodi music source to get "nfs://192.168.3.5/home/public/music/library/Prince/Controversy", which is a correct URL for Kodi.

The issue here is that the URL/path that Kodi works with may be fundamentally different from the beets library path.

docs/plugins/kodiupdate.rst
velvet-jones: I've modified the documentation a bit to try to explain how the two extra options work. Seems wordy and unwieldy. :)

docs/plugins/kodiupdate.rst
sampsyo: Hey, I'm really sorry for dropping the thread here a few months back! It sounds like we need to know _the Kodi equivalent_ of the beets `directory` path. Maybe we should call this `kodi_dir` or something similar to suggest that this needs to match a Kodi configuration parameter that determines _its_ base directory for music?

velvet-jones: No problem, been busy myself. 'kodi_dir' is fine, I just updated the code.

sampsyo:
Looking good!

Sorry if I've misapprehended something here, but I do note that the docs say that the `library` configuration option here should be set to the beets library directory. If that's really the case, can we use the root `config['library']` as a default? Then the user can override this if they really need to. Is it simple to explain when that overriding might be necessary?

velvet-jones:
The global 'directory' is the configuration option that might be duplicated here. If I change it to 'directory' under this plugin that might make it obvious that it overrides or at least is related to the global 'directory'. My concern initially may have been that the global 'directory' could be something like '~/Music'. But if it's guaranteed to be a fqpn, then I think you're right, we can make it optional for overriding.

In the simplest terms, if this plugin knows the exact path to the Kodi music directory and the exact path to the beets music directory, it can figure out how to pinpoint just this new album for scanning by Kodi.

sampsyo:
Ah, I see! The easiest way to get the beets library directory is with the `lib.directory` property of a `Library` object. That is indeed an absolute path (even if the string in the config file was originally a relative path or contained `~`). The `.as_filename()` getter for configuration options ensures that.

An easy way to do the fallback would be to check `'beets_dir' in self.config` and use it only if it's present.

But just to confirm: given that you know the absolute path to the beets base directory via `lib.directory`, is this config option actually necessary? Can you imagine ever using it? If not, maybe we can leave it out and add it later if someone complains.

jtpavlock:
@velvet-jones are you still interested in continuing this?

velvet-jones:
Nope, instead I log the beets activity and process that log separately.
Much more robust. /Bud

jtpavlock:
Sounds good, I'll close in that case then.

Issue #1270, opened 2015-01-26
"No matching release found..." on valid MusicBrainz Release IDs


asayler:
I have a handfull of albums that Beets fails to recognize, yet that seem to be available in teh MusicBrainz database. That alone isn't the point of this report. Instead, even when I manually specify the MusicBrianz release ID for these albums, Beets stills comes back with an error:

`No matching release found for N tracks.`

I've pasted soem of the ID's in question below, along with links to teh MusicBrainz page and a listing of the corresponding directory.

For reference, my beets config and version are:

```
$ beet --version
beets version 1.3.10
plugins: info, fetchart, copyartifacts, inline, replaygain, duplicates, lastgenre, embedart, importadded, scrub, mbsync, lyrics, chroma, missing
```

### Example 1: 823b4f5a-eae8-47fb-b896-f5ccdfaaf605

Link: https://musicbrainz.org/release/823b4f5a-eae8-47fb-b896-f5ccdfaaf605
Listing:
```
01-Stormtroopin'.flac
02-Free-For-All.flac
03-Cat Scratch Fever.flac
04-I Want to Tell You.flac
05-Take It or Leave It.flac
06-Need You Bad.flac
07-Wango Tango.flac
08-Snakeskin Cowboys.flac
09-Stranglehold.flac
10-Dog Eat Dog.flac
Super Hits.cue
Super Hits.log
Super Hits.m3u
```
Beets Output:
```
$ beet import
Ted Nugent/Super Hits (10 items)
Finding tags for album "Ted Nugent - Super Hits".
Candidates:
1. Ted Nugent - Collections (58.0%) (tracks, id, album) (CD, 2006, BMG)
2. Ted Nugent - Great Gonzos: The Best of Ted Nugent (41.2%) (tracks, id, album) (CD, 1981, US, Epic)
3. Ted Nugent - Great Gonzos: The Best of Ted Nugent (39.6%) (tracks, id, album, ...) (CD, 1992, XE, Epic)
4. Ted Nugent - Little Miss Dangerous (23.8%) (tracks, id, album) (CD, 1986, US, Atlantic)
# selection (default 1), Skip, Use as-is, as Tracks, Group albums,
Enter search, enter Id, aBort? I
Enter release ID: 823b4f5a-eae8-47fb-b896-f5ccdfaaf605
No matching release found for 10 tracks.
For help, see: http://beets.readthedocs.org/en/latest/faq.html#nomatch
[...]

PR #3095, opened 2018-12-17
Override match.require and match.ignored when searching for explicit IDs

Holzhaus:
Quick and dirty fix of the problem described in #1270.

### DIFF
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -537,7 +537,8 @@ class Distance(object):
 # Structures that compose all the information for a candidate match.
 
 AlbumMatch = namedtuple('AlbumMatch', ['distance', 'info', 'mapping',
-                                       'extra_items', 'extra_tracks'])
+                                       'extra_items', 'extra_tracks',
+                                       'problems'])
 
 TrackMatch = namedtuple('TrackMatch', ['distance', 'info'])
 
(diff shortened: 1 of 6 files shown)


sampsyo:
Nice! This seems like the right approach. It might be nice to keep a warning, however, when these occur—perhaps that wouldn't be too hard to do either?

Holzhaus:
Ok, now warnings are displayed to the user (using `text_highlight` color). Also, the recommendation is always set to none, so that user confirmation is necessary. That should mitigate accidental copy&pasting ;-)

Let me know if anything else is missing.

Holzhaus:
Whoops, forgot to update the testcases. Also, it's probably better to use a `set` instead of a `list` to keep track of problems.

Holzhaus:
@sampsyo Let me know if you want me to change something else.

sampsyo:
Hey, sorry for dropping the thread on this! This looks good overall—could you perhaps look into adding some docstrings to describe what `problems` is supposed to contain in `AlbumMatch`?

jtpavlock:
@Holzhaus do you intend to continue with this?

Holzhaus:
> @Holzhaus do you intend to continue with this?

Not anytime soon. Feel free to take over and use my branch as a starting point.

jtpavlock:
Thanks for the response. I'll close this for now and advise anyone that'd like to continue this work to re-open or open a new pr.

PR #3098, opened 2018-12-19
Ugly workaround for comparison problems with id3v23 tags

Holzhaus:
If you use the `write` command, some tags will be discarded by mutagen's
`update_to_id3v23()` method. These will then show up as changed fields
and cause a tag sync.

This is a small workaround, but there needs a to be a better way to do
this. Problems with this approach are:

1. It's ugly
2. This unecessarily affects ALL filetypes (e.g. FLAC/Ogg/etc. files) if
   the id3v23 option is set.
3. This still causes problems, e.g. if `month` is set but `day` is not,
   which will cause `update_to_id3v23()` to discard them both.

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1595,9 +1595,21 @@ def write_items(lib, query, pretend, force):
                       displayable_path(item.path), exc)
             continue
 
+        fields = library.Item._media_tag_fields
+        if beets.config['id3v23'].get(bool):
+            # Filter unsupported fields (ugly hack, affects ALL file types)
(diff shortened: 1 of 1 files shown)


Holzhaus:
I'd love have a better solution for this but this would probably need some bigger changes which I don't have the time for.

sampsyo:
Cool! This is a useful first step, in any case! Thank you!

If anyone else would like to take up the mantle, it would be even more awesome to figure out a way to put most of this into the MediaFile layer. This would let the command be sensitive only to fields that are supported by the actual file format.

jtpavlock:
Closing per @sampsyo's comment.

Issue #1632, opened 2015-10-02
embedart does not respect ignore


jhuldtgren:
if you have the following in your config.yaml

ignore: .AppleDouble ._* *~ .DS_Store 

When you run the importer it will ignore files like ._song.mp3 but import song.mp3, however embedart will not honor this and will try to use the bogus file ._album.jpg to embed as the art file instead of using the correct file album.jpg 

If you execute beet with verbose logging you'll even see it fail

Sending event:art_set
embedart: Embedding album art into Artist - Album
embedart: not embedding image of unsupported type: image/x-None
embedart: not embedding image of unsupported type: image/x-None
embedart: not embedding image of unsupported type: image/x-None
embedart: not embedding image of unsupported type: image/x-None

sampsyo:
Seems like a solid addition.

BeetsStudentTeam:
Hey guys, we would like to tackle this issue, if anyone is going to work on this soon let me know so we can look at other issues! We're planning to start working on this on Monday February 1st. Thanks! -Jake

untitaker:
@BeetsStudentTeam Since the latest update on this issue is from last October I don't think anybody will be working on this anytime soon, so feel free to start!

regagain:
This issue is fixed by #3099. It should be fine to close it. :blush:

PR #3099, opened 2018-12-20
Fetchart respect ignore settings

regagain:
See #1632.

# The code
Instead of iterating on all files in the directory, `sorted_walk()` is now being used so the files matching an ignore pattern and hidden files are skipped. The [importer](https://github.com/beetbox/beets/blob/master/beets/importer.py#L1611) does the same thing.

# The tests
I wrote a few tests covering both the `ignore` and `ignore_hidden` settings. I have also moved things around a bit in `test_fetchart.py` in order to avoid duplicate code.

# The documentation
No changes.

# The changelog entry
Added an entry under current version in development / fixes.

### DIFF
--- a/beetsplug/embedart.py
+++ b/beetsplug/embedart.py
@@ -189,7 +189,7 @@ class EmbedCoverArtPlugin(BeetsPlugin):
 
     def remove_artfile(self, album):
         """Possibly delete the album art file for an album (if the
-        appropriate configuration option is enabled.
+        appropriate configuration option is enabled).
         """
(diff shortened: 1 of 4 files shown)


regagain:
Ah, the hidden file test is failing on Windows. I will be looking into this.

regagain:
Fixed!

sampsyo:
Nice! This looks perfect! Thank you!! And double thanks for the thorough testing. :sparkles: :rocket: :rabbit:

PR #3109, opened 2019-01-04
Windows links and hardlinks

Kurchunk:
This adds python module `pywin32` to requirements and uses it in a couple of functions to create links. 
However, a few issues that need editing and testing:

- I deleted the Unix os.symlink code entirely - which means this only works in Windows. I'd like to just append the code so linking can work with Unix _and_ Windows.

- The pywin32 module may be unnecessary. I've read that a built-in module, [`cytypes`, could be used](https://stackoverflow.com/a/1447651/2664985), but I had trouble getting that to work.

- I think Windows only started using this type of linking and hardlinking in Vista. Notice the [requirements in MSDN](https://docs.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-createsymboliclinka)

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -505,7 +505,12 @@ def link(path, dest, replace=False):
         raise FilesystemError(u'file exists', 'rename', (path, dest))
 
     try:
-	symlink(syspath(path), syspath(dest))
+        if sys.platform == 'win32':
+            if ctypes.windll.kernel32.CreateSymbolicLinkW(syspath(dest),syspath(path),0) != 1:
+                raise ctypes.WinError()
+        else:
+            os.symlink(syspath(path), syspath(dest))
+
     except NotImplementedError:
         # raised on python >= 3.2 and Windows versions before Vista
         raise FilesystemError(u'OS does not support symbolic links.'
@@ -530,8 +535,12 @@ def hardlink(path, dest, replace=False):
     if os.path.exists(syspath(dest)) and not replace:
         raise FilesystemError(u'file exists', 'rename', (path, dest))
 
-    try: 
-	hrdlink(syspath(path), syspath(dest))
+    try:
+        if sys.platform == 'win32':
+            if ctypes.windll.kernel32.CreateHardLinkW(syspath(dest), syspath(path), 0) != 1:
+                raise ctypes.WinError()
+        else:
+            os.link(syspath(path), syspath(dest))
 	
     except NotImplementedError:
         raise FilesystemError(u'OS does not support hard links.'
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -32,6 +32,7 @@ from beets.util import hidden
 import six
 from unidecode import unidecode
 from enum import Enum
+import win32file
 
 
 MAX_FILENAME_LENGTH = 200
@@ -133,6 +134,11 @@ class MoveOperation(Enum):
     LINK = 2
     HARDLINK = 3
 
+def symlink(source, link_name):
+    win32file.CreateSymbolicLink(link_name,source)
+
+def hrdlink(source, link_name):
+    win32file.CreateHardLink(link_name,source)
 
 def normpath(path):
     """Provide the canonical form of the path suitable for storing in
@@ -497,8 +503,9 @@ def link(path, dest, replace=False):
 
     if os.path.exists(syspath(dest)) and not replace:
         raise FilesystemError(u'file exists', 'rename', (path, dest))
+
     try:
-        os.symlink(syspath(path), syspath(dest))
+	symlink(syspath(path), syspath(dest))
     except NotImplementedError:
         # raised on python >= 3.2 and Windows versions before Vista
         raise FilesystemError(u'OS does not support symbolic links.'
@@ -522,8 +529,10 @@ def hardlink(path, dest, replace=False):
 
     if os.path.exists(syspath(dest)) and not replace:
         raise FilesystemError(u'file exists', 'rename', (path, dest))
-    try:
-        os.link(syspath(path), syspath(dest))
+
+    try: 
+	hrdlink(syspath(path), syspath(dest))
+	
     except NotImplementedError:
         raise FilesystemError(u'OS does not support hard links.'
                               'link', (path, dest), traceback.format_exc())
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -32,7 +32,7 @@ from beets.util import hidden
 import six
 from unidecode import unidecode
 from enum import Enum
-import win32file
+import ctypes
 
 
 MAX_FILENAME_LENGTH = 200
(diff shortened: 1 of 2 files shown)


CiniqueL:
> 
> 
> [Here's a simpler example for calling `CreateSymbolicLink`](https://stackoverflow.com/a/1447651/2664985), looks like it can be used with `CreateHardLinkA` too.

That's actually the same page and the code I was trying to use, but when I used `ctypes` no links were created. I'm a novice Python programmer at best; so, I'm not sure what I did wrong.

jackwilsdon:
@sampsyo I've just remembered that you need to be an Administrator on Windows to create symlinks - is this an issue?

[Reference](https://security.stackexchange.com/questions/10194/why-do-you-have-to-be-an-admin-to-create-a-symlink-in-windows)

sampsyo:
Wow, that's weird. I guess we just need to be careful to allow this to fail somewhat gracefully, so non-administrators get a reasonable error message.

jackwilsdon:
This seems to work for me when symlinking directories;

```Python
import ctypes
ctypes.windll.kernel32.CreateSymbolicLinkW(r"C:\FakeWindows", r"C:\Windows", 1)
```

Note that I have to run IDLE as an Administrator, I can't just _be_ an administrator (i.e. I have to right click -> Run as Administrator). I think this would be a real pain with beets, and it should be made clear in the documentation if we're looking to do this.

I believe you can symlink files by passing `0` instead of `1` (see [`CreateSymbolicLinkA` on MSDN](https://docs.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-createsymboliclinka) for flags).

Another thing to note is that `CreateSymbolicLinkW` returns `0` if it's successful and _something other than `0`_ (the docs are not specific) in the case of an error.

Annoyingly it seems we can't get the actual error message, as even though `CreateSymbolicLinkW` returns a non-zero value for me when not running as an administrator, `ctypes.windll.kernel32.GetLastError()` returns `0` (which signifies "OK").

CiniqueL:
Ohh yeah. Good catch. Yeah, I forgot you have to have admin permissions for creating links.

CiniqueL:
> 
> 
> This seems to work for me when symlinking directories;
> 
> ```python
> import ctypes
> ctypes.windll.kernel32.CreateSymbolicLinkW(r"C:\FakeWindows", r"C:\Windows", 1)
> ```

I edited my code and tried that out, and it works. I put it in the two functions I made for softlinks and hardlinks.

```
def symlink(source, link_name):
    ctypes.windll.kernel32.CreateSymbolicLinkW(link_name,source,0)

def hrdlink(source, link_name):
    ctypes.windll.kernel32.CreateHardLinkW(link_name,source,0)
```

So, no need to use pywin32

jackwilsdon:
I'm not a fan of the changes suggested in the SO answer you linked, as it actually replaces `os.symlink`, which seems a bit sketchy to me.

We should be able to do something like this to support multiple platforms;

```Python
def link(path, dest, replace=False):
    """Create a symbolic link from path to `dest`. Raises an OSError if
    `dest` already exists, unless `replace` is True. Does nothing if
    `path` == `dest`.
    """
    if samefile(path, dest):
        return

        raise FilesystemError(u'file exists', 'rename', (path, dest))
    try:
        if sys.platform == 'win32':
            if ctypes.windll.kernel32.CreateSymbolicLinkW(syspath(link_name), syspath(source), 0) != 0:
                raise ctypes.WinError()
        else:
            os.symlink(syspath(source), syspath(link_name))
    except NotImplementedError:
        # raised on python >= 3.2 and Windows versions before Vista
        raise FilesystemError(u'OS does not support symbolic links.'
                              'link', (path, dest), traceback.format_exc())
    except OSError as exc:
        # TODO: Windows version checks can be removed for python 3
        if hasattr('sys', 'getwindowsversion'):
            if sys.getwindowsversion()[0] < 6:  # is before Vista
                exc = u'OS does not support symbolic links.'
        raise FilesystemError(exc, 'link', (path, dest),
                              traceback.format_exc())
```

CiniqueL:
Hmm, I'm having an issue witht the fix you suggested.

This is the code I used:

```python
def hardlink(path, dest, replace=False):
    """Create a hard link from path to `dest`. Raises an OSError if
    `dest` already exists, unless `replace` is True. Does nothing if
    `path` == `dest`.
    """
    if samefile(path, dest):
        return

        raise FilesystemError(u'file exists', 'rename', (path, dest))

    try:
        if sys.platform == 'win32':
            if ctypes.windll.kernel32.CreateHardLinkW(syspath(dest), syspath(path), 0) != 0:
                raise ctypes.WinError()
        else:
            os.link(syspath(path), syspath(dest))
	
    except NotImplementedError:
        raise FilesystemError(u'OS does not support hard links.'
                              'link', (path, dest), traceback.format_exc())
    except OSError as exc:
        if exc.errno == errno.EXDEV:
            raise FilesystemError(u'Cannot hard link across devices.'
                                  'link', (path, dest), traceback.format_exc())
        else:
            raise FilesystemError(exc, 'link', (path, dest),
                                  traceback.format_exc())
```

And did a beet import - which returns an error...

`Error: The system cannot find the file specified. during link of paths syspath(dest), syspath(path)`

**But** the link is created! However, it just stops after one file: `syspath(path)`
and if I run the import on the same folder again it moves to another track, creates the link and throws the same error on that file.

So, I tried changing the `!= 0 `to `!= 1` and it seems to be working... all the links are created, but I don't know how this would run on Linux or pre-Vista.

jackwilsdon:
I'm currently on Python 3.7.3 on a Windows 7 machine and it seems like `os.symlink` actually works when run as an administrator? This seems to go against the beets documentation found [here](https://beets.readthedocs.io/en/v1.3.17/reference/config.html#link) which mentions that the `link` option should fail (whereas I can confirm this works on my setup).

It appears that support for `os.symlink` on Windows was actually added in Python 3.2 - does this mean we can close this PR as we're dropping support for older Python versions anyway @sampsyo?

sampsyo:
Neat! Yeah, that sounds like a plan—let's change the docs to say exactly that. We already don't support Python 3.0–1, so using `os.symlink` seems preferable to constructing an OS call directly.

jtpavlock:
This seems close to completion. @Kurchunk any chance you'd be willing to see it through?

jackwilsdon:
Is this not just a documentation change now?

Kurchunk:
> This seems close to completion. @Kurchunk any chance you'd be willing to see it through?

Sorry, I've been way disconnected from this project for months and lost track of what's going on. 
I need a status update. How can I help?

sampsyo:
Indeed, I think @jackwilsdon summed it up well here: https://github.com/beetbox/beets/pull/3109#issuecomment-487403077

Basically, it seems like symlinking on Windows can "just work," under certain circumstances, even without this change. But the docs here need updating to reflect that fact:
https://beets.readthedocs.io/en/v1.3.17/reference/config.html#link

Issue #2694, opened 2017-09-22
Plugin: Spotify


dtvu:
### Problem

I was playing around with the Spotify plugin and I noticed that none of my tracks matched a Spotify ID. I wondered if this was potentially related to Spotify [removing unauthenticated calls to the web API](https://developer.spotify.com/news-stories/2017/01/27/removing-unauthenticated-calls-to-the-web-api/).

### Setup

* OS: Ubuntu 16.04
* Python version: Python 3.5.2
* beets version: 1.4.5
* Turning off plugins made problem go away (yes/no): no

sampsyo:
Hmm; interesting. Could you perhaps include some verbose output?

dtvu:
Sorry, I should have included that before: 
```
$ beet -vv
user configuration: /home/tony/.config/beets/config.yaml
data directory: /home/tony/.config/beets
plugin paths: 
Sending event: pluginload
library database: /home/tony/.config/beets/library.db
library directory: /media/archive/music
Sending event: library_opened
Usage: 
  beet COMMAND [ARGS...]
  beet help COMMAND

Options:
  --format-item=FORMAT_ITEM
                        print with custom format
  --format-album=FORMAT_ALBUM
                        print with custom format
  -l LIBRARY, --library=LIBRARY
                        library database file to use
  -d DIRECTORY, --directory=DIRECTORY
                        destination music directory
  -v, --verbose         log more details (use twice for even more)
  -c CONFIG, --config=CONFIG
                        path to configuration file
  -h, --help            show this help message and exit

Commands:
  clearart          remove images from file metadata
  config            show or edit the user configuration
  embedart          embed image files into file metadata
  extractart        extract an image from file metadata
  fetchart          download album art
  fields            show fields available for queries and format strings
  gmusic-songs      list of songs in Google Play Music library
  gmusic-upload     upload your tracks to Google Play Music
  help (?)          give detailed help on a specific sub-command
  import (imp, im)  import new music
  list (ls)         query the library
  mbsync            update metadata from musicbrainz
  modify (mod)      change metadata fields
  move (mv)         move or copy items
  remove (rm)       remove matching items from the library
  spotify           build a Spotify playlist
  stats             show statistics about the library or a query
  update (upd, up)  update the library
  version           output version information
  write             write tag information to files
Sending event: cli_exit
```

sampsyo:
[...]

PR #3123, opened 2019-01-20
Add Spotify OAuth support and autotagger integration

rhlahuja:
Addressing https://github.com/beetbox/beets/issues/2694, I've added OAuth support for the Spotify plugin using Client Credentials Flow. This requires a [registered application](https://developer.spotify.com/documentation/general/guides/app-settings/) and Client ID/Secret to be provided via `client_id` and `client_secret` under the `spotify` options in config.yaml.

I've also started working on adding Spotify support for the importer (incomplete `album_for_id` and `track_for_id` funcs included in this PR with some logic copied from the Discogs plugin), since I've come across more than a few releases that don't exist in MusicBrainz/Discogs/Beatport/Bandcamp but do in Spotify. However, I realize that the Spotify API doesn't provide certain fields like album `media`, `catalognum`, etc. like these other beets sources do. So, @sampsyo, do you think it's still worth extending the autotagger with Spotify matches, despite offering incomplete metadata?

### DIFF
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -104,6 +104,20 @@ class SpotifyPlugin(BeetsPlugin):
         return {'Authorization': 'Bearer {}'.format(self.access_token)}
 
     def _handle_response(self, request_type, url, params=None):
+        """
+        Send a request, reauthenticating if necessary.
+
+        :param request_type: Type of :class:`Request` constructor,
+            e.g. ``requests.get``, ``requests.post``, etc.
+        :type request_type: function
+        :param url: URL for the new :class:`Request` object.
+        :type url: str
+        :param params: (optional) list of tuples or bytes to send
+            in the query string for the :class:`Request`.
+        :type params: dict
+        :return: class:`Response <Response>` object
+        :rtype: requests.Response
+        """
         response = request_type(url, headers=self._auth_header, params=params)
         if response.status_code != 200:
             if u'token expired' in response.text:
@@ -122,6 +136,8 @@ class SpotifyPlugin(BeetsPlugin):
 
         :param url_type: Type of Spotify URL, either 'album' or 'track'
         :type url_type: str
+        :param id_: Spotify ID or URL
+        :type id_: str
         :return: Spotify ID
         :rtype: str
         """
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -2,7 +2,6 @@
 
 from __future__ import division, absolute_import, print_function
 
-import os
 import re
 import json
 import base64
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -16,7 +16,7 @@ from beets.autotag.hooks import AlbumInfo, TrackInfo
 
 
 class SpotifyPlugin(BeetsPlugin):
-    # URL for the Web API of Spotify
+    # Endpoints for the Spotify API
     # Documentation here: https://developer.spotify.com/web-api/search-item/
     oauth_token_url = 'https://accounts.spotify.com/api/token'
     base_url = 'https://api.spotify.com/v1/search'
@@ -54,7 +54,9 @@ class SpotifyPlugin(BeetsPlugin):
         self.setup()
 
     def setup(self):
-        """Retrieve previously saved OAuth token or generate a new one"""
+        """
+        Retrieve previously saved OAuth token or generate a new one
+        """
         try:
             with open(self.tokenfile) as f:
                 token_data = json.load(f)
@@ -113,8 +115,8 @@ class SpotifyPlugin(BeetsPlugin):
 
     def album_for_id(self, album_id):
         """
-        Fetches an album by its Spotify album ID or URL and returns an AlbumInfo object
-        or None if the album is not found.
+        Fetches an album by its Spotify album ID or URL and returns an
+        AlbumInfo object or None if the album is not found.
         """
         self._log.debug(u'Searching for album {}', album_id)
         match = re.search(self.id_regex.format('album'), album_id)
@@ -180,8 +182,8 @@ class SpotifyPlugin(BeetsPlugin):
 
     def track_for_id(self, track_id):
         """
-        Fetches a track by its Spotify track ID or URL and returns a TrackInfo object
-        or None if the track is not found.
+        Fetches a track by its Spotify track ID or URL and returns a
+        TrackInfo object or None if the track is not found.
         """
         self._log.debug(u'Searching for track {}', track_id)
         match = re.search(self.id_regex.format('track'), track_id)
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -24,7 +24,6 @@ class SpotifyPlugin(BeetsPlugin):
     album_url = 'https://api.spotify.com/v1/albums/'
     track_url = 'https://api.spotify.com/v1/tracks/'
     playlist_partial = 'spotify:trackset:Playlist:'
-    id_regex = r'(^|open\.spotify\.com/{}/)([0-9A-Za-z]{{22}})'
 
     def __init__(self):
         super(SpotifyPlugin, self).__init__()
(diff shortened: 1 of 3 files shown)


sampsyo:
Awesome; this is a great start!

We can register for a proper API key. For the tests, however, we should avoid needing to communicate with the actual Spotify API server—instead, the tests should mock the API’s responses so they can be run even without a connection to the Internet.

beetsplug/spotify.py
```diff
@@ -63,35 +328,35 @@ def parse_opts(self, opts):
...
         results = []
         failures = []
 
         items = lib.items(query)
 
         if not items:
-            self._log.debug(u'Your beets query returned no items, '
-                            u'skipping spotify')
+            self._log.debug(
+                u'Your beets query returned no items, ' u'skipping spotify'
```
sampsyo: This can be just one long string literal now. :smiley:

beetsplug/spotify.py
```diff
@@ -3,54 +3,319 @@
...
+                'client_id': 'N3dliiOOTBEEFqCH5NDDUmF5Eo8bl7AN',
+                'client_secret': '6DRS7k66h4643yQEbepPxOuxeVW0yZpk',
+                'tokenfile': 'spotify_token.json',
+                'source_weight': 0.5,
+                'user_token': '',
+            }
+        )
+        self.config['client_secret'].redact = True
+
+        """Path to the JSON file for storing the OAuth access token."""
```
sampsyo: Since this isn't at the top of a function, it's not a docstring—a comment would probably be more appropriate.

beetsplug/spotify.py
```diff
@@ -3,54 +3,319 @@
...
+        self.tokenfile = self.config['tokenfile'].get(
+            confit.Filename(in_app_dir=True)
+        )
+        self.setup()
+
+    def setup(self):
+        """
```
sampsyo: Our docstrings usually start directly following the `"""`, with no newline.

beetsplug/spotify.py
rhlahuja: Yea I wasn't sure if the [Sphinx autoattributes](http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoattribute) still requires docstrings, but looks like they support comments now.

rhlahuja:
I've added some documentation to spotify.rst regarding the new functionality, and happy to add more if necessary.

One thing I'm not quite sure about the proper implementation of the `album_distance` and `track_distance` functions. For now, I've simply copied these from the Beatport plugin, but let me know if we need to add anything else. I've also specified the Spotify ID for `AlbumInfo.album_id` and `TrackInfo.track_id` values which I think is correct behavior, but would like to confirm.

Thanks again for the swift replies all your work on this awesome project!

docs/plugins/spotify.rst
```diff
@@ -67,10 +106,20 @@ in config.yaml under the ``spotify:`` section:
   track/album/artist fields before sending them to Spotify.  Can be useful for
   changing certain abbreviations, like ft. -> feat.  See the examples below.
   Default: None.
+- **tokenfile**: Filename of the JSON file stored in the beets configuration
+  directory to use for caching the OAuth access token.
+  Default: ``spotify_token.json``.
```
sampsyo: Hmm… perhaps we shouldn't even mention this option in the docs. I don't expect anyone will ever need to change it, and omitting the unnecessary options should make the docs page a little easier to read. (FWIW, we don't document `tokenfile` for the other OAuth-based plugins we have.)

docs/plugins/spotify.rst
```diff
@@ -37,6 +58,24 @@ Command-line options include:
...
+        Bear Hands - Blue Lips / Ignoring the Truth / Back Seat Driver (Spirit Guide) / 2AM
+    URL:
+        https://open.spotify.com/album/2rFYTHFBLQN3AYlrymBPPA
+    (Similarity: 88.2%) (source, tracks) (Spotify, 2019, Spensive Sounds)
+     * Blue Lips (feat. Ursula Rose)   -> Blue Lips (feat. Ursula Rose) (source)
+     * Ignoring the Truth              -> Ignoring the Truth (source)
+     * Back Seat Driver (Spirit Guide) -> Back Seat Driver (Spirit Guide) (source)
+     * 2AM                             -> 2AM (source)
+    [A]pply, More candidates, Skip, Use as-is, as Tracks, Group albums,
+    Enter search, enter Id, aBort, eDit, edit Candidates, plaY?
```
sampsyo: I'm not sure having the whole candidate display is necessary here. Maybe the just the first two lines would be enough to get the idea across?

docs/plugins/spotify.rst
```diff
@@ -37,6 +58,24 @@ Command-line options include:
 * ``--show-failures`` or ``-f``: List the tracks that did not match a Spotify
   ID.
 
+A Spotify ID or URL may also be provided to the ``Enter release ID:``
+prompt during import::
```
sampsyo: Here's another opportunity for active voice: for example, "You can enter the URL for an album or song on Spotify at the `enter Id` prompt during import."

docs/plugins/spotify.rst
```diff
@@ -1,23 +1,44 @@
 Spotify Plugin
 ==============
 
-The ``spotify`` plugin generates `Spotify`_ playlists from tracks in your library. Using the `Spotify Web API`_, any tracks that can be matched with a Spotify ID are returned, and the results can be either pasted in to a playlist or opened directly in the Spotify app.
+The ``spotify`` plugin generates `Spotify`_ playlists from tracks in your
+library with the ``beet spotify`` command. Using the `Spotify Search API`_,
+any tracks that can be matched with a Spotify ID are returned, and the
+results can be either pasted in to a playlist or opened directly in the
+Spotify app.
```
sampsyo: Here's a slightly more direct way to say this without the passive voice: "Also, the plugin can use the Spotify Search API to provide metadata matches for the importer."

docs/plugins/spotify.rst
```diff
@@ -1,23 +1,44 @@
...
+
+Spotify URLs and IDs may also be provided to the ``Enter release ID:`` prompt
+during ``beet import`` to autotag music with data from the Spotify
+`Album`_ and `Track`_ APIs.
```
sampsyo: We can probably get away without mentioning this here—the "search by ID" functionality is a standard part of any plugin that extends the importer with a metadata backend, so the above description (and the detail below) probably covers this just fine.

docs/plugins/spotify.rst
rhlahuja: This part of the doc is actually referring to the `beet spotify` command which uses the Search API, but good point about using the active voice. I've consolidated this with the portion below and it now reads:

> The `spotify` plugin generates Spotify playlists from tracks in your
library with the `beet spotify` command using the Spotify Search API.
Also, the plugin can use the Spotify Album and Track APIs to provide
metadata matches for the importer.

Not sure if we even need to specifically mention that we're using the Search/Album/Track APIs, I can remove that if not.

rhlahuja:
Actually, I should also add some logic to populate `TrackInfo.medium_total`. Let's hold off on merging until I can get to that later today.

sampsyo:
Looking good! Just one last thing: do you want to include that API key in the default configuration? If so, perhaps we should mention that it's actually optional to register for your own Spotify app; you can just use the included key unless you really want to change it. If not, then maybe we should _not_ provide a default to force people to register with Spotify themselves.

rhlahuja:
Yea, that client ID/secret isn't actually valid (just randomly generated strings), so I agree we should either register a dedicated Spotify app and include those creds or remove them entirely and force people to register one themselves. I noticed the Discogs plugin sets a default client ID/secret which I'm assuming is for a dedicated app, so it seems like that's the preferred approach? Not sure if we would run into issues with rate limiting or anything else with many users sharing the same app. Which option do you prefer?

sampsyo:
Sounds good. Let's start by offering an "official" app key and revisit (i.e., reset the key and force people to register for their own apps) if it becomes a problem. Here are credentials for an app I just registered:

- client ID: `4e414367a1d14c75a5c5129a627fcab8`
- client secret: `f82bdc09b2254f1a8286815d02fd46dc`

sampsyo:
Looks perfect! I'll merge this now.

PR #3125, opened 2019-01-22
Add Spotify `candidates` and `item_candidates`

rhlahuja:
I realized I'd forgotten to implement the `candidates` and `item_candidates` functions in https://github.com/beetbox/beets/pull/3123 for the Spotify plugin, which I assume are used by the importer to identify potential matches when not provided manually. I'm using the Spotify Search API to retrieve these candidates, and have modularized that logic to make this easier to implement.

I noticed that the Beatport and Discogs plugins don't use any of the passed `Item` attributes when determining potential matches. Is this recommended behavior if that data can't be meaningfully used to refine the API queries?

### DIFF
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -261,7 +261,7 @@ class SpotifyPlugin(BeetsPlugin):
         for i, track_data in enumerate(album_data['tracks']['items']):
             if track_data['disc_number'] == track.medium:
                 medium_total += 1
-                if track_data['id'] == spotify_id:
+                if track_data['id'] == track.track_id:
                     track.index = i + 1
         track.medium_total = medium_total
         return track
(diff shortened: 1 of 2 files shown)


rhlahuja:
Gotcha, that's what I suspected. We could potentially apply some of those `Item` attributes as filters when searching Spotify, but a less restrictive search may actually be preferred since it should theoretically result in more candidate matches.

And if it's okay with you, before merging I'd like to add some tests for edge cases like where the album or artist contain a colon/quotes/spaces to ensure that we're still providing valid query string to the API.

sampsyo:
Sounds great. Let me know when you're happy with the tests.

rhlahuja:
In my preliminary testing, it seems that special characters don't actually require any specific handling/escaping. However, the Search API doesn't seem to interpret Unicode characters properly, e.g. querying for `artist:deadmau5 album:4×4` doesn't return [this album](https://musicbrainz.org/release-group/aa7ffa90-2581-4941-ac54-8f5abfd6cbc2), but `artist:deadmau5 album:4x4` does. This may be expected/intended behavior on Spotify's end, so I'm using `unidecode` (which I've seen used elsewhere in beets) on the query string to substitute Unicode characters with their ASCII equivalents and calling it a day.

Since tests around this conversion are outside the scope of this plugin, I think this is ready to merge unless you have other feedback. We can always revisit if special characters / Unicode handling later if it proves to be an issue.

sampsyo:
Very nice. I do think unidecode is the right thing if that’s the way the search engine works. Merging! Thanks again!! :sparkles:

Issue #3132, opened 2019-01-25
Data not updated correctly after re-import or mbsync


djl:
Apologies for the vague title, I wasn't sure how to concisely describe this issue so I'll show steps on how to recreate it:

1. Import an album with a non-empty disambiguation field. For example, [Paul Simon's "In the Blue Light"](https://musicbrainz.org/release/f09e7198-0b72-4521-906d-5589123b31b3)
2. Confirm the disambiguation string is stored in beets' database:
```
$ beet ls -a -f '$albumartist - $album ($albumdisambig)' paul simon in the blue light
Paul Simon - In the Blue Light (Mastered for iTunes)
```
3. Re-import this album with a release that doesn't have a disambiguation field (e.g. [this one](https://musicbrainz.org/release/5d3961f5-b4e0-4337-b5b6-a9bc4f2fc456)
4. Notice the disambiguation string hasn't been updated (or more accurately, it hasn't been removed):
```
$ beet ls -a -f '$albumartist - $album ($albumdisambig)' paul simon in the blue light
Paul Simon - In the Blue Light (Mastered for iTunes)
```

`beet mbsync` does not update the field either, leading to local data unexpectedly de-syncing from MusicBrainz.

(Note: I've demonstrated this with the `$albumdisambig` field but I believe it happens with any field.)

sampsyo:
It's true that we only update this field (as with several other "miscellaneous" metadata fields) if it has value:
https://github.com/beetbox/beets/blob/6d020e3d2ba675b1e4a21e8c1b753ec88134ca4f/beets/autotag/__init__.py#L145-L159

This trade-off means that we don't clobber existing metadata, in case it's actually meaningful. There could be an argument made that we should treat the disambiguation fields as "special" in that a missing value is actually significant and should overwrite an old string.

But in the mean time, you might be interested in the [from_scratch](https://beets.readthedocs.io/en/stable/reference/config.html#from-scratch) config option, which prevents the importer from preserving any old metadata.

djl:
> Perhaps these miscellaneous fields could be user-configurable with the current values being the defaults?

Opened a pull request to see if this is viable: #3133

sampsyo:
Cool! I’m glad that will work for now.

True—but on the other hand, a missing value might indicate that MB doesn’t know the catalog number (or whatever), meaning that it would be good to keep the existing value if it’s set.

Anyway, I see you’ve opened a PR while I as typing this—awesome! I’ll take a look at that.

[...]

PR #3133, opened 2019-01-25
First attempt at fixing #3132.

djl:
Add a new `no_clobber` config option that contains a list of "miscellaneous" metadata fields not to be overwritten by empty values.

I'm not entirely happy with the `no_clobber` name. Suggestions welcome.

### DIFF
--- a/beets/autotag/__init__.py
+++ b/beets/autotag/__init__.py
@@ -156,7 +156,7 @@ def apply_metadata(album_info, mapping):
                       'data_source',):
             # Don't overwrite fields with empty values unless the
             # field is explicitly allowed to be overwritten
-            clobber = field not in config['no_clobber'].get()
+            clobber = field not in config['no_clobber'].as_str_seq()
             value = getattr(album_info, field)
             if value is None and not clobber:
                 continue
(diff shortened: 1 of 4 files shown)


beets/autotag/__init__.py
```diff
@@ -154,9 +154,14 @@ def apply_metadata(album_info, mapping):
                       'albumdisambig',
                       'releasegroupdisambig',
                       'data_source',):
+            # Don't overwrite fields with empty values unless the
+            # field is explicitly allowed to be overwritten
+            clobber = field not in config['no_clobber'].get()
```
sampsyo: A type assertion here might be useful: `.as_str_seq()` instead of `.get()`.

djl: Good catch. Fixed.

sampsyo:
Hmm... I’m not sure what the best course of action is. Maybe it would be simplest to invert the sense of the option (overwrite_null?), leave it empty by default, and possibly leave it undocumented for now because of its hard-to-explain limitations?

djl:
Updated the code to handle the track-level fields and inverted the logic as asked. Also removed the documentation and changelog entries for now.

And it looks like Travis is having a bad day.

djl:
All green :+1:

sampsyo:
Awesome! This looks like it will be useful to build on, _and_ it made the code for that function simpler. Thanks for your help with this!! :sparkles:

PR #3135, opened 2019-01-30
Support ipfs add --nocopy in ipfs plugin

wildthyme:
This is a very minor change, but previously the ipfs plugin was making a separate copy of every file added. I left the default behaviour the same because using this requires setting an additional option in ipfs proper.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -64,7 +64,7 @@ New features:
   Thanks to :user:`rhlahuja`.
   :bug:`3123`
 * :doc:`/plugins/ipfs`: The plugin now supports a ``nocopy`` option which passes that flag to ipfs.
-  THanks to :user:`wildthyme`.
+  Thanks to :user:`wildthyme`.
 
 
 Changes:
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -63,6 +63,8 @@ New features:
   provider: you can match tracks and albums using the Spotify database.
   Thanks to :user:`rhlahuja`.
   :bug:`3123`
+* :doc:`/plugins/ipfs`: The plugin now supports a ``nocopy`` option which passes that flag to ipfs.
+  THanks to :user:`wildthyme`.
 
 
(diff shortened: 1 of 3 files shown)


docs/changelog.rst
```diff
@@ -63,6 +63,8 @@ New features:
   provider: you can match tracks and albums using the Spotify database.
   Thanks to :user:`rhlahuja`.
   :bug:`3123`
+* :doc:`/plugins/ipfs`: The plugin now supports a ``nocopy`` option which passes that flag to ipfs.
+  THanks to :user:`wildthyme`.
```
jackwilsdon:
```suggestion
  Thanks to :user:`wildthyme`.
```

sampsyo:
Nice! This looks good to me.

If I'm reading it right, the Travis failure here:
https://travis-ci.org/beetbox/beets/jobs/486649015

Is unrelated; it looks like yet another flake8 version bump we need to address.

jackwilsdon:
@sampsyo looks like it's due to `pep8-naming@0.8.0` (see [this issue](https://github.com/PyCQA/pep8-naming/issues/95)).

The easiest resolution for now is probably to [completely avoid 0.8.0](https://github.com/lpsinger/gwcelery/commit/a6b91c7a8325186fadf0909f9eaa1c28e28cfc77), as a version containing the [fix](https://github.com/PyCQA/pep8-naming/pull/97) has not been released yet.

sampsyo:
Good find!! Arg, that's frustrating—but you're correct that this is probably the right fix.

sampsyo:
Wow, that's quite a few independent issues. I suppose we should avoid this version, but maybe it would also make sense to `#noqa` those "ET" imports—those seem like a perfectly acceptable violation of a general policy.

jackwilsdon:
I'll open a PR to sort out as much as I can 👍we can discuss it more there

sampsyo:
Looks perfect; thanks again! :sparkles:

PR #3136, opened 2019-01-31
Fix linter issues

jackwilsdon:
As it turns out, there _is_ a whole bunch of crazy indentation in our code! No idea how it got there and how it wasn't caught before, but I've fixed the few places where the indentation was wrong.

I also changed the `is` comparison to be a `==` check, as we shouldn't be using `is` to compare strings.

And finally, I've locked `pep8-naming` to `0.7.x` so we can avoid the weird bugs that are in that version.

### DIFF
--- a/beets/art.py
+++ b/beets/art.py
@@ -60,8 +60,8 @@ def embed_item(log, item, imagepath, maxwidth=None, itempath=None,
             log.info(u'Image not similar; skipping.')
             return
     if ifempty and get_art(log, item):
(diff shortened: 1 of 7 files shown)


sampsyo:
Wow! I guess this is why we have style checkers—thank you for fixing all the wonky indentation.

Let's continue to monitor the changes in new versions of pep8-naming. When the underscore problem gets fixed, for example, that will go a long way to our being able to unlock the version.

Many, many thanks!! :sparkles:

Issue #3081, opened 2018-11-14
discogs: Rate limiting


arawde:
### Problem
I'm trying to run beets to import my music library, using the discogs plugin.

I run 
```sh
$ beet import /mnt/music
```
Which yields: 
```
Traceback (most recent call last):
  File "/usr/share/beets/beetsplug/discogs.py", line 226, in get_master_year
    year = result.fetch('year')
  File "/home/avr/.local/lib/python3.6/site-packages/discogs_client/models.py", line 245, in fetch
    self.refresh()
  File "/home/avr/.local/lib/python3.6/site-packages/discogs_client/models.py", line 211, in refresh
    data = self.client._get(self.data['resource_url'])
  File "/home/avr/.local/lib/python3.6/site-packages/discogs_client/client.py", line 123, in _get
    return self._request('GET', url)
  File "/home/avr/.local/lib/python3.6/site-packages/discogs_client/client.py", line 120, in _request
    raise HTTPError(body['message'], status_code)
discogs_client.exceptions.HTTPError: 429: You are making requests too quickly.
```
Some googling reveals that Discogs rate limits requests to about 3 per minute.

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv import /mnt/music
```
Causes the application to hang:
```
Candidate: Arctic Monkeys - Whatever People Say I Am, That’s What I’m Not (aa94ffcc-1c9b-39d6-953f-7aa803d8146c)
Computing track assignment...
...done.
Success. Distance: 0.01
discogs: Searching for master release 76279
discogs: Searching for master release 76279
discogs: Searching for master release 76279
```

### Setup

* OS: Ubuntu 18.10
* Python version:  3.6.7
* beets version: 1.4.7
* Turning off plugins made problem go away (yes/no): Problem is with plugin

My configuration (output of `beet config`) is:

```yaml
directory: /mnt/music
library: /mnt/music/beets.db
ignore: [0bleep, beats]
max_filename_length: 255
original_date: yes
per_disc_numbering: yes

ui:
    color: yes

import:
    copy: no
    move: yes
    timid: yes

paths:
    default: $albumartist/$album ($year) %aunique{}/$discn$track $title

plugins: inline edit discogs embedart fetchart replaygain scrub web
item_fields:
    discn: str(disc).lstrip('0') + '.' if disctotal > 1 else ''
discogs:
    source_weight: 0.0
    user_token: REDACTED
    apikey: REDACTED
    apisecret: REDACTED
[...]

PR #3138, opened 2019-02-07
keep discogs requests below rate limit

puhoy:
this is should fix issue #3081.

since the discogs_client doesnt expose the rate limit headers (yet) (see https://github.com/discogs/discogs_client/issues/82), i added some code to keep the request rate below 60 requests per minute for logged-in users, or 25 for not logged-in users, as described here: https://www.discogs.com/developers/#page:home,header:home-rate-limiting

### DIFF
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -73,6 +73,7 @@ class DiscogsPlugin(BeetsPlugin):
         # Try using a configured user token (bypassing OAuth login).
         user_token = self.config['user_token'].as_str()
         if user_token:
+            # rate limit for authenticated users is 60 per minute
             self.rate_limit_per_minute = 60
             self.discogs_client = Client(USER_AGENT, user_token=user_token)
             return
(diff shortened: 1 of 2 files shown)


beetsplug/discogs.py
```diff
@@ -222,8 +243,11 @@ def get_master_year(self, master_id):
         """
         self._log.debug(u'Searching for master release {0}', master_id)
         result = Master(self.discogs_client, {'id': master_id})
+
+        self.wait_for_rate_limiter()
         try:
             year = result.fetch('year')
+            self.last_request_timestamp = time.time()
```
sampsyo: Consider introducing a function `request_finished`?

beetsplug/discogs.py
```diff
@@ -222,8 +243,11 @@ def get_master_year(self, master_id):
         """
         self._log.debug(u'Searching for master release {0}', master_id)
         result = Master(self.discogs_client, {'id': master_id})
+
+        self.wait_for_rate_limiter()
```
sampsyo: Maybe a more semantic name like `request_start` would help explain why the call is here?

beetsplug/discogs.py
```diff
@@ -88,6 +91,20 @@ def setup(self, session=None):
         self.discogs_client = Client(USER_AGENT, c_key, c_secret,
                                      token, secret)
 
+    def _time_to_next_request(self):
+        seconds_between_requests = 60 / self.rate_limit_per_minute
+        seconds_since_last_request = time.time() - self.last_request_timestamp
+        seconds_to_wait = seconds_between_requests - seconds_since_last_request
+        if seconds_to_wait > 0:
```
sampsyo: There are two positivity checks: one here and one in `wait_for_rate_limiter`. Seems like we only need one of the two.

beetsplug/discogs.py
```diff
@@ -71,6 +73,7 @@ def setup(self, session=None):
         # Try using a configured user token (bypassing OAuth login).
         user_token = self.config['user_token'].as_str()
         if user_token:
+            self.rate_limit_per_minute = 60
```
sampsyo: Perhaps a comment here would help explain why the rate limit is different when we have a user token.

sampsyo:
Fantastic! Thank you again for your help. Nice work. :sparkles:

Issue #3139, opened 2019-02-09
Beets doesn't abort immediately when instructed, due to long running task in 'chroma' plugin


ssssam:
### Problem

Steps to reproduce:

1. Create a directory with lots of MP3 files (at least 10)
2. Run `beet import` on the directory.
3. Press CTRL+C.
4. Watch for a while as Beets fingerprints every file in the directory before exiting.

The same issue occurs when the 'aBort' option is selected from the autotag prompt.

I reproduced the issue using the following config file:

```
directory: /tmp/beets-testcase
library: /tmp/beets-testcase/db

plugins: chroma

import:
    timid: yes
```

### Cause

The issue is that the 'chroma' plugin fingerprints every file in the directory during a single `fingerprint_file()` task. This doesn't get aborted when the pipeline receives an exception, so the `Pipeline.run_parallel()` method blocks until all the files have been fingerprinted.

As soon as the task finishes all of this data is thrown away due to the exception, so we should really stop processing files as soon as possible here and allow Beets to exit quickly.

sampsyo:
Sounds good! This would be good behavior to have.

ssssam:
In fact, even without the 'chroma' plugin it seems like the `lookup_candidates()` stage could block for a while.

One way to fix this would be to abort Beets without allowing the threads to complete. This could introduce issues like data corruption in tasks which write to files, so it's probably a bad idea.

We could add periodical checks into long running tasks so that they exit if a certain flag is set. What about adding a `task.aborted` flag, and inserting blocks such as this in places:

```
if task.aborted:
    return
```

PR #3140, opened 2019-02-09
Abort quickly on user request

ssssam:
First attempt at fixing https://github.com/beetbox/beets/issues/3139

This branch greatly reduces the delay when I tell Beets to abort. It's not perfect, but it's survived my testing so far.

I haven't added any automated tests for this functionality. It's not immediately obvious how best to do that.

### DIFF
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -389,7 +389,7 @@ def _add_candidate(items, results, info):
 
 
 def tag_album(items, search_artist=None, search_album=None,
-              search_ids=[]):
+              search_ids=[], cancellable=None):
     """Return a tuple of the current artist name, the current album
     name, and a `Proposal` containing `AlbumMatch` candidates.
 
@@ -459,6 +459,9 @@ def tag_album(items, search_artist=None, search_album=None,
                                                         search_album,
                                                         va_likely):
             _add_candidate(items, candidates, matched_candidate)
+            if cancellable and cancellable.cancelled():
+                log.debug(u'Got cancellation flag. Aborting')
+                break
 
     log.debug(u'Evaluating {0} candidates.', len(candidates))
     # Sort and get the recommendation.
@@ -468,7 +471,7 @@ def tag_album(items, search_artist=None, search_album=None,
 
 
 def tag_item(item, search_artist=None, search_title=None,
-             search_ids=[]):
+             search_ids=[], cancellable=None):
     """Find metadata for a single track. Return a `Proposal` consisting
     of `TrackMatch` objects.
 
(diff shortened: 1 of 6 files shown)


sampsyo:
Thanks for quickly putting together a proposal! In spirit, something like this—setting and checking a synchronized "cancellation" flag to gently request that long-running stages finish up—might be one of the only ways to do this.

However, I am really sad about the amount of additional coupling and threading-through that this approach requires. One thing that makes me particularly sad, for example, is that `tag_album` and `tag_item` now have a `cancelable` parameter, which we will need to begin explaining as a value of a class that's defined somewhere else entirely—in the `pipeline` module. It would be really sad to couple this self-contained description of "work to do" with the library that orchestrates parallelism, which is a seemingly distinct concern.

I'll have to ponder this. This approach might be the only way, but it comes with unfortunate consequences. Some creativity will be required to come up with something that achieves the goals without the downsides—or else we'll need to convince ourselves that the additional complexity is essential and can't be avoided.

ssssam:
I understand your concern and I'll be happy if you can come up with a better way to do it!

Note that the Cancellable object is independent of the 'pipeline' module. It happens to be defined there at the moment, but it could equally be in its own module. 

I suspect that we could achieve a much nicer solution using the Python [asyncio](https://docs.python.org/3/library/asyncio.html) module. That already [supports task cancellation](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel) for example. I've not spent the time to fully learn how to use that module though, and it's also only available in Python 3.6 and later.

sampsyo:
Here’s one crazy idea: turn certain interruptible things into coroutines. For example, the fingerprinter function could `yield` between each file. Then we could add support to the pipeline module for handling `yield`s that don’t produce anything—they’d just be a no-op, except that they give the pipeline module a chance to check for the shutdown flag.

Not sure that would be better or worse!

ssssam:
> Here’s one crazy idea: turn certain interruptible things into coroutines. For example, the fingerprinter function could `yield` between each file. Then we could add support to the pipeline module for handling `yield`s that don’t produce anything—they’d just be a no-op, except that they give the pipeline module a chance to check for the shutdown flag.

I don't see how that could work for the 'chroma' plugin, because the slow task isn't triggered directly from the pipeline module but rather from `plugins.send()`. Currently the return value of that function is ignored almost everywhere, so if the chroma plugin returns a generator it'll never actually get executed.

sampsyo:
Ah, yes, good point—I was thinking of the plugins that add entire import stages to the pipeline, but chroma does not do that.

jtpavlock:
I believe [asyncio is also in 3.5](https://docs.python.org/3.5/library/asyncio.html), which will be the minimum supported python version started with beets v1.5.

ssssam:
I'm not going to pursue this further at the moment. I'd love to see a solution, especially a clean one using asyncio :)

PR #3144, opened 2019-02-14
Add id3v23 option to convert plugin

### LINKED ISSUE #3104: convert: Add id3v23 option
The `convert` plugin can be used to convert files, so that they can be copied to a media player or devices that does not have support for all file formats in the music library, e.g. if parts of my library are FLAC but my media player only supports MP3.

Similarily,  only reason to use ID3v2.3 is that certain mediaplayers and devices don't support ID3v2.4 tags. Thus, I'd be nice to use id3v24 for you local library but use the `convert` to not only convert audio data but also tag data to a format that my playback device understands.

Thus, it'd be nice to have an `id3v23` option in the `convert` plugins settings, to make sure that ID3v2.3 tags are written to all files that are copied/converted using this plugin.

Holzhaus:
This resolves #3104.

### DIFF
--- a/beets/art.py
+++ b/beets/art.py
@@ -51,7 +51,8 @@ def get_art(log, item):
 
 
 def embed_item(log, item, imagepath, maxwidth=None, itempath=None,
-               compare_threshold=0, ifempty=False, as_album=False):
+               compare_threshold=0, ifempty=False, as_album=False,
+               id3v23=None):
     """Embed an image into the item's media file.
     """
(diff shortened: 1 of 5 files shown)


sampsyo:
Looks good; thanks (and thanks for updating the docs!). Can you please add a quick changeling entry? Then we can merge this.

Holzhaus:
Done

sampsyo:
Looks great; thank you!!

Issue #3143, opened 2019-02-14
ImportError: cannot import name 'NoReturn'


DarkVirtue:
### Problem

Any invocation of beet immediately fails. This was working as recently as three weeks ago.

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv import Amaranthe
Traceback (most recent call last):
  File "/usr/local/bin/beet", line 7, in <module>
    from beets.ui import main
  File "/usr/local/lib/python3.5/dist-packages/beets/ui/__init__.py", line 42, in <module>
    from beets.autotag import mb
  File "/usr/local/lib/python3.5/dist-packages/beets/autotag/__init__.py", line 26, in <module>
    from .match import tag_item, tag_album, Proposal  # noqa
  File "/usr/local/lib/python3.5/dist-packages/beets/autotag/match.py", line 24, in <module>
    from munkres import Munkres
  File "/usr/local/lib/python3.5/dist-packages/munkres.py", line 20, in <module>
    from typing import Union, NewType, Sequence, Tuple, NoReturn, Optional, Callable
ImportError: cannot import name 'NoReturn'
```

### Setup

* OS: Debian 9
* Python version: 3.5.3
* beets version: 1.4.7 (I think; beet --version fails with the above error)
* Turning off plugins made problem go away (yes/no): no

My configuration (output of `beet config`) is:

```yaml
directory: /home/media/music
library: /var/lib/beets/data/musiclibrary.db
plugins: fetchart lyrics lastgenre multidisc missing scrub
import:
  move: yes
  language: en
paths:
  default: $albumartist_sort - $album ($year)/%if{$multidisc,Disc $disc%if{$disctitle, - $disctitle}/}/$track - $title
asciify_paths: yes
missing:
  format: $albumartist_sort - $album ($year)/%if{$multidisc,Disc $disc%if{$disctitle, - $disctitle}/}/$track - $title
  count: no
  total: no
```

ethanrjones97:
Seconding this bug report. 
```
Traceback (most recent call last):
  File "/home/erj/.local/bin/beet", line 7, in <module>
    from beets.ui import main
  File "/home/erj/.local/lib/python3.5/site-packages/beets/ui/__init__.py", line 42, in <module>
    from beets.autotag import mb
  File "/home/erj/.local/lib/python3.5/site-packages/beets/autotag/__init__.py", line 26, in <module>
    from .match import tag_item, tag_album, Proposal  # noqa
[...]

PR #3146, opened 2019-02-15
Lock munkres to 1.0.x

jackwilsdon:
Even though munkres 1.1.2 fixed the issue with Python 3.5 (#3143), it had already dropped support for Python <3.5.x in [1.1.0](https://github.com/bmc/munkres/blob/5badef0ae62ed5a8d4164fd358dda96163295c8f/CHANGELOG.md) (which beets still supports).

We are going to have to lock our version of munkres to 1.0.x until we drop support for 3.5 ourselves.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ setup(
     install_requires=[
         'six>=1.9',
         'mutagen>=1.33',
-        'munkres',
+        'munkres~=1.0.0',
         'unidecode',
         'musicbrainzngs>=0.4',
         'pyyaml',


sampsyo:
This will do nicely; thank you for tracking down the version information!

Freso:
This breaks beets on Arch Linux. If beets works with munkres 1.1.x, it shouldn't hard-fail when that's the version available, even if it shouldn't depend on munkres >1.0.x.

jackwilsdon:
We could add a conditional version dependency to require munkres 1.0.x on Python <3.5 and allow 1.1.x on >=3.5 if that would solve it @Freso?

Freso:
@jackwilsdon I replied at https://github.com/beetbox/beets/issues/3168#issuecomment-466993174

PR #3149, opened 2019-02-17
Allow overriding "fast" detection in FieldQuery subclasses

Holzhaus:
This is small fix that allows to override the `fast` detection for `FieldQuery` subclasses. This can potentially speed up the `playlist` query in #3145.

This also removes a duplicate `key.lower()` invocation.


Holzhaus:
Closed in favor of #3150.

PR #3150, opened 2019-02-17
Add support for "named queries" and use them in the playlist plugin

sampsyo:
As discussed in https://github.com/beetbox/beets/pull/3145#pullrequestreview-204523870, this adds a generic facility for defining "named" queries that are *not* field queries. This way, a query part like `playlist:foo` need not rely on a declaration of a field called `playlist` despite the fact that no one is ever supposed to put data in the `playlist` field. Instead, the query parser can recognize these non-field queries and use them directly.

This also should obviate the need for #3149, which was another way to "force" field queries to be fast, even when the underlying field did not exist. @Holzhaus, can you please take a look and let me know if this suffices to get the behavior you want?

This should also make it possible to support random queries like `rand:5`, as discussed in https://github.com/beetbox/beets/pull/2350.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -143,6 +143,11 @@ class Model(object):
     are subclasses of `Sort`.
     """
 
+    _queries = {}
+    """Named queries that use a field-like `name:value` syntax but which
+    do not relate to any specific field.
+    """
+
     _always_dirty = False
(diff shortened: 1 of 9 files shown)


Holzhaus:
@sampsyo Looks good to me!

sampsyo:
Wonderful! I'll merge now.

PR #3151, opened 2019-02-17
playlist: Add auto-update functionality and more tests

Holzhaus:
This is round two for the `playlist` plugin:

This improves the existing tests (and adds some more to ensure nothing breaks), and also adds the auto-update functionality that I described in issue #123:

If the `auto` is set, then whenever a database item is moved or removed, the playlists inside the `playlist_dir` are updated to reflect that change.

### DIFF
--- a/beetsplug/playlist.py
+++ b/beetsplug/playlist.py
@@ -99,13 +99,12 @@ class PlaylistPlugin(beets.plugins.BeetsPlugin):
             self.relative_to = beets.util.bytestring_path(
                 beets.config['directory'].as_filename())
         elif self.config['relative_to'].get() != 'playlist':
-            print(repr(self.config['relative_to'].get()))
             self.relative_to = beets.util.bytestring_path(
                 self.config['relative_to'].as_filename())
         else:
             self.relative_to = None
 
-        if self.config['auto'].get(bool):
+        if self.config['auto']:
             self.register_listener('item_moved', self.item_moved)
             self.register_listener('item_removed', self.item_removed)
             self.register_listener('cli_exit', self.cli_exit)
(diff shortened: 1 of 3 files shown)


Holzhaus:
"Trial and error" works surprisingly well when fixing issues with Windows without access to an installation btw ;-)

beetsplug/playlist.py
```diff
@@ -86,6 +87,100 @@ class PlaylistPlugin(beets.plugins.BeetsPlugin):
...
         })
+
+        self.playlist_dir = self.config['playlist_dir'].as_filename()
+        self.changes = {}
+
+        if self.config['relative_to'].get() == 'library':
+            self.relative_to = beets.util.bytestring_path(
+                beets.config['directory'].as_filename())
+        elif self.config['relative_to'].get() != 'playlist':
+            print(repr(self.config['relative_to'].get()))
```
sampsyo: A `print` left over from debugging, perhaps?

beetsplug/playlist.py
```diff
@@ -86,6 +87,100 @@ class PlaylistPlugin(beets.plugins.BeetsPlugin):
...
+            self.relative_to = beets.util.bytestring_path(
+                self.config['relative_to'].as_filename())
+        else:
+            self.relative_to = None
+
+        if self.config['auto'].get(bool):
```
sampsyo: FWIW, this conversion is done automatically, so this is equivalent:

```suggestion
        if self.config['auto']:
```

beetsplug/playlist.py
Holzhaus: Oops, removed it. Sorry.

beetsplug/playlist.py
Holzhaus: Ok, changed it.

sampsyo:
Awesome! Nice work again!!

PR #3153, opened 2019-02-19
Add parallel map utility

bemeurer:
Following #3003 and #3006, and the now long-forgotten promises I made, here is the parallel_map function making it into utils! This way we can now very easily apply transforms to the library in a parallel way (in Py3) while remaining compatible with Py2.

This also converts `absubmit` to use the new util. Expect updates to #3006 as well!

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -24,6 +24,7 @@ import re
 import shutil
 import fnmatch
 from collections import Counter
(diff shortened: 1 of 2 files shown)


sampsyo:
Awesome; thank you for factoring this out! I've merged this and committed some refinements to the docstring (and a new comment).

Issue #1579, opened 2015-08-25
fetchart: Handle exceptions in Google backend


randybias:
With a big library if you force a fetchart there is a good chance that a connection is reset and unfortunately, this case isn't handled with some basic retrying:

``` python
fetchart: The Chemical Brothers - The Private Psychedelic Reel: found album art
fetchart: The Chemical Brothers - We Are the Night: found album art
Traceback (most recent call last):
  File "/usr/local/bin/beet", line 9, in <module>
    load_entry_point('beets==1.3.15', 'console_scripts', 'beet')()
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/beets-1.3.15-py2.7.egg/beets/ui/__init__.py", line 1161, in main
    _raw_main(args)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/beets-1.3.15-py2.7.egg/beets/ui/__init__.py", line 1151, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/beets-1.3.15-py2.7.egg/beetsplug/fetchart.py", line 412, in func
    self.batch_fetch_art(lib, lib.albums(ui.decargs(args)), opts.force)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/beets-1.3.15-py2.7.egg/beetsplug/fetchart.py", line 513, in batch_fetch_art
    path = self.art_for_album(album, local_paths)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/beets-1.3.15-py2.7.egg/beetsplug/fetchart.py", line 487, in art_for_album
    for url in self._source_urls(album):
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/beets-1.3.15-py2.7.egg/beetsplug/fetchart.py", line 532, in _source_urls
    for url in urls:
[...]

PR #3155, opened 2019-02-19
fetchart: Add some error handling to prevent crashes

Holzhaus:
Today I had some network problems regarding dbpedia.org, which made
beets crash because a `requests.exceptions.ConnectionError` was raised
(`[Errno 113] No route to host`). This commits adds some error handling
around network requests to prevent further crashes in the future.


sampsyo:
This all looks grea; thanks! I know it’s a small change, but could you please add a quick changelog entry?

Holzhaus:
Done. I also mentioned #1579 in the changelog entry, since it should be fixed by this PR.

sampsyo:
Perfect; thank you!! Looks great.

Issue #3064, opened 2018-10-28
Beets not respecting my match: ignored: missing_tracks configuration


udiboy1209:
### Problem

I have mentioned in my config that `missing_tracks` should be ignored while matching. But the matching fails on an album which has 10 tracks out of which I am importing 5.
 
Running `beet import` in verbose (`-vv`) mode:

```sh
$ beet -vv import Doowhops\ and\ Hooligans/

user configuration: /home/udiboy/.config/beets/config.yaml
data directory: /home/udiboy/.config/beets
plugin paths: 
/home/udiboy/.local/lib/python2.7/site-packages/pylast/__init__.py:51: UserWarning: You are using pylast with Python 2. Pylast will soon be Python 3 only. More info: https://github.com/pylast/pylast/issues/265
  UserWarning,
Sending event: pluginload
library database: /home/udiboy/Music/.beetsdb
library directory: /home/udiboy/Music
Sending event: library_opened
Sending event: import_begin
Sending event: import_task_created
Sending event: import_task_start
Looking up: /home/udiboy/Downloads/Doowhops and Hooligans
Tagging Bruno Mars - Doo-Wops & Hooligans
No album ID found.
Search terms: Bruno Mars - Doo-Wops & Hooligans
Album might be VA: False
Searching for MusicBrainz releases with: {'release': u'doo-wops & hooligans', 'tracks': u'5', 'artist': u'bruno mars'}
Requesting MusicBrainz release e5b02753-c983-4ff9-81b6-a6cbd01a8d47
primary MB release type: album
Sending event: albuminfo_received
Candidate: Bruno Mars - Doo-Wops & Hooligans (e5b02753-c983-4ff9-81b6-a6cbd01a8d47)
Computing track assignment...
...done.
Ignored. Penalty: missing_tracks
Requesting MusicBrainz release 33030768-eed0-404a-bd5e-af6950546211
primary MB release type: album
Sending event: albuminfo_received
Candidate: Bruno Mars - Doo-Wops & Hooligans (33030768-eed0-404a-bd5e-af6950546211)
Computing track assignment...
...done.
Ignored. Penalty: missing_tracks
Requesting MusicBrainz release d6982d34-b422-4b37-8b49-025ae87b4c10
primary MB release type: album
Sending event: albuminfo_received
Candidate: Bruno Mars - Doo-Wops & Hooligans (d6982d34-b422-4b37-8b49-025ae87b4c10)
Computing track assignment...
...done.
[...]

PR #3179, opened 2019-03-07
Improve specifications in the matches.ignored setting

malan88:
The setting is quite confusing. After having difficulty getting the autotagger to accept an id for a specific release that I knew was correct, I investigated and count [this closed issue][0] that totally explains the problem.

To clarify the issue, I quoted Adrian from the issue and elaborated a bit, providing the solution. Unfortunately, the solution is undocumented. I'd like to document it, but I am unsure what the actual weighting system in `distance_weights` is. Adrian says 0.9 to ignore missing_tracks, which implies it's similar to the distance weighting in the other settings, but other settings in the default config are set to multiples of 1, implying it is not.

[0]: https://github.com/beetbox/beets/issues/3064

### DIFF
--- a/docs/reference/config.rst
+++ b/docs/reference/config.rst
@@ -802,13 +802,7 @@ the penalty name to the ``ignored`` setting::
 
 The available penalties are the same as those for the :ref:`max_rec` setting.
 
-Please note: the ignored setting will ignore any *matches* that have the specified penalties. In a sense, it maximizes the listed penalties. It will not ignore *the fact* that those matches have the penalties. For instance, with the above setting, it ignores all matches with missing tracks. It doesn't ignore the fact that there are missing tracks, which would sort of be the opposite.
-
-To ignore *the penalties themselves* in match selection, add something like the following to your matches settings:
-
-    match:
-        distance_weights:
-            missing_tracks: 0.9
+For example, setting ``ignored: missing_tracks`` will skip any album matches where your audio files are missing some of the tracks. The importer will not attempt to display these matches. It does not ignore the fact that the album is missing tracks, which would allow these matches to apply more easily. To do that, you'll want to adjust the penalty for missing tracks.
 
 .. _required:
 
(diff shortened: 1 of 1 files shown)


sampsyo:
How about this somewhat simpler wording?

> For example, setting `ignored: missing_tracks` will skip any album matches where your audio files are missing some of the tracks. The importer will not attempt to display these matches. It does not ignore the *fact* that the album is missing tracks, which would allow these matches to apply more easily. To do that, you'll want to adjust the penalty for missing tracks.

At which point we'd need to refer to another part of the documentation that says how to do that. Maybe we can make this change to the section now, and then in a separate PR, discuss how to document the `distance_weights`, which we have intentionally left undocumented in the past because it's pretty hard to explain what to do there.

malan88:
I'm fine with that. Would you like to make the change or would you like to? (this is my first PR for anything ever, so a little new to it).

sampsyo:
Congratulations on your first PR! :tada: :smiley:

I would love it if you could make the change. For what it's worth, I think I'm not alone among OSS maintainers in really enjoying a workflow where we recommend changes in comments, and contributors revise until we can merge the entire PR without modification. This way, we stay as transparent as possible about how we want changes to proceed, and contributors get more hands-on experience in following the guiding principles of the project.

malan88:
So, wasn't sure if I was supposed to directly edit the commit or just create a new one. I went with the latter [after googling][0]

[0]: https://stackoverflow.com/questions/16748115/how-to-modify-github-pull-request

sampsyo:
Excellent; thanks! Indeed, pushing a new commit to the same PR is the right thing to do (because it keeps the discussion organized).

malan88:
I guess I delete my branch now? That was much easier than I expected. Also: pretty cool. Very much makes me realize how much Eric Raymond was right that FOSS is all about do-ocracy.

sampsyo:
Yep, you're welcome to do so. Unless you want to keep it around as a keepsake. :smiley:

PR #3187, opened 2019-03-23
Clarify description of incremental_skip_later

jroitgrund:
Contrary to what the current docs say, `incremental_skip_later` should be set to `no` for beets  to revisit skipped directories: https://github.com/beetbox/beets/blob/3b78d933cf46d5cad8f90ea6cdf38ada9e10846f/beets/importer.py#L582-L586


sampsyo:
A solid improvement; thank you!!

PR #3189, opened 2019-03-27
beets update: dont move album art twice

jroitgrund:
In many cases https://github.com/beetbox/beets/issues/2419 was being hit not because users had moved the album art, but because:
* beets moves items in an album
* this moves the covert art since `with_album=True`, but doesn't update album.artpath in the DB because `store=False`
* beets moves the album itself
* this moves the cover art again from the old path, which no longer exists

https://github.com/beetbox/beets/pull/3030 hides this error in most cases, but without this fix beets will still lose track of your album art, even if it doesn't throw, as evidenced by the test.

note: to fail without the fix, the test needs to add a second item to the album. this is because only the *second* item move actually moves the album cover, because the album's path is guessed from its first item's path, and the first item is stored *after* being moved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,3 +89,4 @@ ENV/
 /.project
 /.pydevproject
 /.settings
+.vscode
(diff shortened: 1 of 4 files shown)


sampsyo:
Wow; this is incredibly subtle! Thank you for tracking it down—and for the clean, simple fix. This is truly awesome work.

Issue #800, opened 2014-06-02
BPD: Update to newer MPD protocol version


simonwjackson:
When running `ncmpcpp`, i get the following error message:

```
NCMPCPP: MPD < 0.16.0 is not supported
```

Can this be resolved by reporting a different version number to the clients? Or is there a fundamental difference in the code?

arcresu:
If anyone is still following this issue, this is an update to say that we're almost there! I have a pull request in progress (#3214) that is aiming to finish off the final missing parts of the MPD protocol support for 0.16. It's at a stage where it's now talking to ncmpcpp and basic things like changing songs seem to work.

Simple clients like mpdscribble, mpDris and mpc all seem to be working OK. The more complex clients like MPDroid, M.A.L.P. and ncmpcpp are all still broken to varying degrees. I've been keeping a list of problems in the description of the PR and gradually addressing them. If you'd like to help this effort it'd be great to test out some of your favourite MPD clients with the code in that PR and report back about any further issues you find.

bearcatsandor:
@arcresu Thanks for the hard work!

PR #3190, opened 2019-03-27
Add tests for BPD

arcresu:
I'd like to work on updating the MPD server version supported by BPD (#800) so as a first step I'm adding some tests that exercise BPD. I found an existing file that was testing parts of the BPD plugin, though the name is a bit unexpected (`test_player.py` rather than, say, `test_bpd.py`) so perhaps this should be renamed.

The test code adds a simple TCP client that can speak to an MPD server. It provides a small number of convenience functions for building requests and interpreting responses but it intended to be pretty dumb. This way we can treat BPD as a black box and test how it responds over the protocol rather than unit testing the functions in the plugin's module. The real BPD server is started in a separate thread, and we have to wait a small amount of time for it to spin up before attempting to connect to it. ~~Since starting the BPD server requires executing `import gi` this means the tests have to grow a dependency on the relevant GObject modules. Since I start the BPD server in the `setUp` method, we incur the wait-to-connect delay for each test. I set this to 0.1s and it's sufficient on my laptop but it might need to be longer on other hardware.~~ (The client now polls the server until the socket is bound so there's not much wasted delay, and the overall timeout is longer.)

I've written tests based on the current MPD protocol specification, which is several versions ahead of where BPD is at. For this reason I've marked several of them as expected failures for now. The various `test_implements_*` tests are just checking that BPD claims to support all of the commands in each category of the full MPD list. They can perhaps be removed after we support all the commands, but I'm just using them as a progress meter for now to highlight what's missing from the implementation.

My plan is to add more tests and update BPD in subsequent PRs. As well as new commands there are new behaviours like a new more powerful filter syntax and additional flags for old commands. For now I just want to get this stub for the test framework merged.

### DIFF
--- a/test/test_player.py
+++ b/test/test_player.py
@@ -216,7 +216,7 @@ class BPDTest(unittest.TestCase, TestHelper):
         self.lib.add_album([self.item1, self.item2])
 
         with open(os.path.join(self.temp_dir, b'bpd_mock.py'), 'wb') as f:
-            f.write(b'from unittest import mock\n')
+            f.write(b'import mock\n')
             f.write(b'import sys, imp\n')
             f.write(b'gstplayer = imp.new_module("beetsplug.bpg.gstplayer")\n')
             f.write(b'gstplayer._GstPlayer = mock.MagicMock(spec_set=["time",')
@@ -260,7 +260,7 @@ class BPDTest(unittest.TestCase, TestHelper):
         # Launch BPD in a new process:
         env = dict(os.environ.items())
         env['PYTHONPATH'] = ':'.join(
-                [self.temp_dir.decode('utf-8')] + sys.path[1:])
+                [self.temp_dir.decode('utf-8')] + sys.path[1:]).encode('utf-8')
         beet = os.path.join(os.path.dirname(__file__), '..', 'beet')
         server = subprocess.Popen([
             beet,
--- a/test/test_player.py
+++ b/test/test_player.py
@@ -20,6 +20,12 @@ from __future__ import division, absolute_import, print_function
 import unittest
 from beetsplug import bpd
 
+import multiprocessing as mp
+import socket
+import time
+
+from test.helper import TestHelper
+
 
 class CommandParseTest(unittest.TestCase):
     def test_no_args(self):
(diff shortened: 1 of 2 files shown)


arcresu:
On Travis PyGObject is complaining about GLib being too old. Based on [this commit](https://gitlab.gnome.org/GNOME/pygobject/commit/464889c874ccb0129d9ff9240094974f8030fa4b) is seems PyGObject 3.31.2 was the version where this last changed (and before that commit, the GLib version on Travis would be OK for PyGObject). On Appveyor the problem is about the cairo header being missing on Windows.

I could try and fix up the dependency configuration for the CIs to make this work (not sure what to do about cairo on Windows), but I think instead I'll change the test approach to use a dummy `gstplayer` module to avoid having this dependency at all. I'm not actually testing playback, so this should be fine.

arcresu:
The tests failed on Windows because it's attempting to pickle the whole BPD server as part of spawning the server process, and that includes some un-picklable objects (locks for the db). I tried running the server in a `threading.Thread` instead of a `multiprocessing.Process`, but that seemed to interfere with how bluelet works.

I suppose I need to be setting up the library in the process that's spawned to run the server rather than trying to pass it in. That would probably mean losing access to some of the `Library` test helpers though. I had a first attempt at rewriting to do this but ran into problems with the socket connections being refused or not closing properly. I'll try again when I next have time to look at this.

test/test_player.py
```diff
@@ -63,6 +89,305 @@ def test_backslash_in_arg(self):
...
+            if not hello.ok:
+                raise RuntimeError('Bad hello: {}'.format(hello.status))
+
+    def __del__(self):
+        self.sock.close()
+
+    def get_response(self):
+        ''' Wait for a full server response and wrap it in a helper class.
+        If the request was a batch request then this will return a list of
+        `MPCResponse`s, one for each processed subcommand. '''
```
sampsyo: In beets, we use [the PEP8 recommendations for docstrings](https://www.python.org/dev/peps/pep-0008/#documentation-strings), which include triple-double quotes, no spaces around the delimiters, and a newline before the closing delimiter:

```py
"""Like this.
"""
```

arcresu:
Thanks for the review! I've had another go now using `subprocess` and it's passing the tests on my machine.

The difficulty is that whatever mocking of the player is done, it has to happen in the new process rather than the test process. Maybe I'm missing an easier way of doing that, but my hacky solution was the best way I could think of after trying a few things. I can't see a good way to use `mock.patch` in this case. My solution for now is:
  1. During `setUp` I write python file in the test's `temp_dir` that does the necessary mocking and import mechanism hijacking before importing the real bpd plugin. (This could just be a real file in the `test/` directory since it doesn't need to be dynamically created.)
  1. I then configure beets to load this file as a plugin so that it gets executed.

The main inelegant bit is ensuring that the launched process has the correct python executable and paths. Looks like I didn't properly serialise the python path into the environment for Windows so I'll still have to sort that out.

sampsyo:
Aha; you're right—that is way more complicated than I made it seem at first. Sorry to lead you astray! This does seem to work—at least, most of the way—but I'm no longer convinced it's any easier than the "fork" approach with multiprocessing that you were proposing before.

To get the correct Python executable, you can use [sys.executable](https://docs.python.org/3/library/sys.html#sys.executable) (but you probably already tried that).

arcresu:
I think in the end I prefer the forking approach because I don't trust that I can launch a fresh process with exactly the same python environment otherwise. The detour wasn't wasted effort though since the server management is much more robust now.

I stumbled into an unexpected behaviour with paths on the BPD virtual filesystem under Windows. When listing directory contents with the `lsinfo` command, paths to directories are returned with a forward slash as the path separator, but paths to files are returned using a backslash. Attempting to add the backslash-separated paths to the BPD playlist causes an error. This might be the first bug flushed out by the new tests :)

As far as I'm concerned now that this is running it's ready to go in. I'll save the addition of new tests and changes in the BPD server itself for future PRs.

test/test_player.py
```diff
@@ -63,6 +89,380 @@ def test_backslash_in_arg(self):
...
+            '--library', self.config['library'].as_filename(),
+            '--directory', py3_path(self.libdir),
+            '--config', py3_path(config_file.name),
+            'bpd'
+        )
+        server = mp.Process(target=start_beets, args=args)
+        server.start()
+
+        # Wait until the socket is bound:
+        self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
sampsyo: I think this socket creation has to go *inside* the loop. At least on my system (macOS 10.14.4), if the connection fails *once* when the server is not ready, then subsequent attempts to call `self.sock.connect_ex` return error 22 (invalid argument).

arcresu: Ah, ok. It makes more sense that it works that way, thanks.

arcresu:
I didn't see that connection reset message on Linux, but I see from the CI logs that there's a similar error on Windows. I thought the client and its socket were being garbage collected (and so closed) before the server was killed but I guess I was wrong. I've tried a different approach where the "test socket" is passed into the client rather than having the client open a fresh one. I'm then explicitly closing that socket before killing the server process.

I've been mainly using pytest locally, but I'm seeing now that with nose there are warnings about `unittest.expectedFailure` not being supported. It also seems that nose completely fails to collect the tests dynamically created with my `implements()` helper (pytest sees them just fine). In both cases this should be a temporary thing while updating the protocol.

I'm really not sure why the Python 2.7 tests started failing on the CI with an ascii encoding exception. I don't see how the latest commit changes any string handling, and the tests run fine on my Linux machine under Python 2.7.16

arcresu:
It seems that the issue is something to do with the encoding of the source files. I can't see any problem with the files those commits touched and in fact rolling back to the exact commit that was previously passing CI is now failing CI. I can reproduce the error locally by adding `setenv = LANG=C` to `tox.ini`, and stop the error by making that `LANG=en_US.UTF-8`. That made no difference to either CI though. I think it must be something external to this MR responsible.

sampsyo:
Sorry about those test failures! I tracked that down to a Tox bug that was introduced just a couple of days ago:
https://github.com/tox-dev/tox/issues/1234

I changed the Travis and Appveyor configurations to avoid the version with the bug, so future commits should not run into this problem

sampsyo:
And you're right—it does look like nose just doesn't support `unittest.expectedFailure`, and that the issue is known: https://github.com/nose-devs/nose/issues/33

arcresu:
Ok thanks for following up about the tox issues. I rebased my branch and switched the `expectedFailure` decorators for `skip` so that there's no spurious output on nose as far as I can see besides bpd printing `Building directory tree...\n... done.` each test. I've already been working on another branch to improve bpd itself including converting that message to use the debug logger.

It might seem that the test file is a bit helper-heavy, but it means that the actual tests are short and readable and provide useful errors when they fail. As I improve the test coverage the ratio of test code to helper code will improve. E.g. this test asserts that all of the responses were OK up to the one that is expected to fail, and takes care of starting the bpd server, adding the item to its playlist (and checking that response), and stopping the server:

```python
    def test_cmd_playlistinfo(self):
        with self.run_bpd() as client:
            self._bpd_add(client, self.item1)
            responses = client.send_commands(
                    ('playlistinfo',),
                    ('playlistinfo', '0'),
                    ('playlistinfo', '200'))

        self._assert_failed(responses, bpd.ERROR_ARG, pos=2)
```

sampsyo:
Awesome. This will be incredibly valuable to have, as you say, for improving BPD and increasing its protocol coverage. Thank you for all the work you've already put into this. :sparkles:

PR #3195, opened 2019-03-31
the: log a debug only when text is changed

arcresu:
Previously the `the` plugin would log a debug message when the text _didn't_ get changed by the plugin, whereas I think what was intended was the opposite. With this change the logged messages show the actual transformations made by the plugin.


sampsyo:
Indeed; that totally looks wrong! Thanks for the quick fix!!

PR #3196, opened 2019-03-31
bpd: use plugin logger instead of global logger

arcresu:
There was a code comment indicating that there was a plan to use the plugin logger instead of the global one, so this PR makes that change. It means we drop the `--debug` flag from the `bpd` command since the logger is now controlled by the normal `beet -v` verbosity flag. Also fixes up a debug message that was previously `print`ed to stdout, and tweaks the format of the MPD protocol debug messages to make them a little easier to read/interpret. New look:
```
[...]
bpd: Building directory tree...
[...]
bpd: Finished building directory tree.
bpd: New client connected
bpd: server: OK MPD 0.13.0
bpd: client: command_list_ok_begin
bpd: client: status
bpd: client: currentsong
bpd: client: command_list_end
bpd: server: volume: 10
             repeat: 0
             random: 0
             consume: 0
             playlist: 0
             playlistlength: 0
             xfade: 0
             mixrampdb: 0
             mixrampdelay: nan
bpd: server: state: stop
bpd: server: list_OK
bpd: server: list_OK
bpd: server: OK
```

### DIFF
--- a/beetsplug/bpd/__init__.py
+++ b/beetsplug/bpd/__init__.py
@@ -729,7 +729,7 @@ class CommandList(list):
                 e.index = i  # Give the error the correct index.
                 raise e
 
-            # Otherwise, possibly send the output delimeter if we're in a
+            # Otherwise, possibly send the output delimiter if we're in a
             # verbose ("OK") command list.
             if self.verbose:
                 yield conn.send(RESP_CLIST_VERBOSE)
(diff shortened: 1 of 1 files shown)


sampsyo:
Awesome!! Yes, that has been a long-standing issue. :rocket: Thanks for cleaning this up!

PR #3197, opened 2019-03-31
Fix several deprecation warnings

arcresu:
While running the test suite with pytest in Python 3.7, it flagged a few distinct deprecation warnings. I see that the YAML one is handled in another PR, but this one addresses:

  * `log.warn()` [should be `log.warning`](https://bugs.python.org/issue13235)
  * Abstract base classes from `collections` were moved to `collections.abc` in Python 3.3 and [will be unavailable](https://docs.python.org/3/library/collections.abc.html) from their original location in Python 3.8.
  * Regex flags that aren't at the start of the pattern [have been deprecated](https://bugs.python.org/issue22493)
  * The `plistlib` module has a [changed interface](https://docs.python.org/3.7/library/plistlib.html#plistlib.readPlist) since Python 3.4

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -23,7 +23,6 @@ from collections import defaultdict
 import threading
 import sqlite3
 import contextlib
-import collections
 
 import beets
 from beets.util.functemplate import Template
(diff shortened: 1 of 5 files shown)


sampsyo:
These all look perfect! Thanks again!! :pancakes:

PR #3199, opened 2019-03-31
Provide an example for the languages list

thomas-mc-work:
It's hard to impossible to figure out the correct syntax for non python users.

### DIFF
--- a/docs/reference/config.rst
+++ b/docs/reference/config.rst
@@ -571,10 +571,12 @@ default is ``apply``.
 languages
 ~~~~~~~~~
 
-A list of locale names to search for preferred aliases (e.g. ``en jp es``. For example, setting
-this to "en" uses the transliterated artist name "Pyotr Ilyich Tchaikovsky"
+A list of locale names to search for preferred aliases. For example, setting
+this to ``en`` uses the transliterated artist name "Pyotr Ilyich Tchaikovsky"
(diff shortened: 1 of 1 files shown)


thomas-mc-work:
Or is `languages: de en` also allowed?

jackwilsdon:
You can probably do this (I've not checked);

```YAML
languages:
  - de
  - en
```

sampsyo:
Hi! Thanks for looking into this!

Yes, just `en de` will do fine (although both `[en, de]` and @jackwilsdon's YAML-style bulleted list will also work).

Could you revise this to use that, and to perhaps use a complete sentence (instead of a parenthetical) to give the example?

thomas-mc-work:
Like that?

sampsyo:
Yes, like that—but I also asked about trying to rephrase the example to use a complete sentence, not a parenthetical. (Also, you're missing a parenthesis, and a comma alway needs to follow "e.g.".) For example, you might put something like this after the current single-language example:

> You can use a space-separated list of language abbreviations, like `en jp es`, to specify a preference order.

thomas-mc-work:
Alright. Thank you for your clarifications! I hope this approach looks good to you. If not, please let me know.

sampsyo:
Perfect! Thank you!!

PR #3200, opened 2019-03-31
bpd: extend coverage of MPD protocol

arcresu:
Adds support for some of the "easy" MPD protocol features bpd is missing, as part of #800. I'll do more complicated features like the `idle` command separately, and unfortunately we need `idle` to be able to increment the protocol version string.

The MPD protocol specification document is not very precise and is intended to be read together with the MPD source code and an actual MPD server. Some of the clients I've seen tend to ignore the MPD version in the client hello message (or maybe at most issue a warning if it looks too old) and attempt to use the commands they want to use anyway, whereas others abort on an old version. For those reasons I think incremental updating of the protocol support should be OK, as long as we don't bump the protocol version string past a version of the protocol that we can actually support.

This adds/changes:
  * new `consume` mode, which erases songs from the playlist as they are played,
  * new `single` mode, which stops playback when the current song finishes,
  * `repeat` mode now repeats the entire playlist rather than a single track unless used in conjunction with `single` mode,
  * issuing `previous` when on track 0 now restarts track 0 instead of stopping bpd,
  * `status` gained some new fields, like the higher-precision times `elapsed:` and `duration:`,
  * improved reliability by sending protocol errors instead of crashing as before,
  * dummy support (i.e. the server accepts the commands but doesn't actually do anything) for `crossfade`, `mixramp*`, `replay_gain_*`, `volume`, and stored playlist commands (`listplaylists`, `save`, etc.).


arcresu:
Ok I think this might be enough for the first batch of bpd changes before this PR gets too long. The main feature in this PR that benefits current users right now is that bpd will crash less often.

After merging this bpd will be in a sort of hybrid state in terms of protocol support since since e.g. `repeat` is now using a more modern interpretation than our protocol version string indicates, which could potentially confuse some clients. However even then I don't think it would make bpd unusable. We really need `idle` to be able to start incrementing the protocol version, but that's a more complicated command since it requires implementing an event system and pushing messages to the client. I'll do that in another MR but just wanted to start with some easy ones.

The stored playlist features could be implemented for real easily enough, but since they're optional in MPD anyway I'm starting off with placeholders. Advanced playback features like crossfade and replaygain could potentially have real implementations instead of the current placeholders if we pick up an optional dependency on a more sophisticated audio player backend library. We can think about that after catching up with MPD though, if there's even a demand for it.

I've been testing locally with `mpc` and comparing behaviour with a real MPD version 0.21.4 server. Other clients like `ncmpcpp` refuse to talk to bpd since the protocol version is too old, so it's hard to do more robust real world testing until this effort is a little further along.

beetsplug/bpd/__init__.py
```diff
@@ -686,6 +758,13 @@ def run(self, conn):
                 for data in results:
                     yield conn.send(data)
 
+        except TypeError:
+            # The client provided too many arguments.
+            raise BPDError(ERROR_ARG,
+                           u'wrong number of arguments for "{}"'
+                           .format(self.name),
+                           self.name)
```
sampsyo: I know this is a little tricky, but this `TypeError` handler may be a little too broad. It could catch simple programming errors inside of command-handler functions, like `'a' + 2`, for example. It might be best to use our `inspect` helper to explicitly check the argument count of `func`.

beetsplug/bpd/__init__.py
```diff
@@ -939,11 +1018,19 @@ def cmd_status(self, conn):
         if self.current_index > -1:
             item = self.playlist[self.current_index]
 
-            yield u'bitrate: ' + six.text_type(item.bitrate / 1000)
-            # Missing 'audio'.
+            yield (
+                u'bitrate: ' + six.text_type(item.bitrate / 1000),
+                # TODO provide a real value samplerate:bits:channels 44100:24:2
+                u'audio: 0:0:0',
```
sampsyo: FWIW, this should actually be possible to provide: `item.samplerate`, `item.bitdepth`, and `item.channels` are all standard fields.

arcresu: Thanks, I was assuming this had to come from the player and didn't realise beets already knows all this!

beetsplug/bpd/__init__.py
arcresu: Thanks for insisting on this point- it lead me to fixing the deeper issue of why bpd was crashing at all when there was a catch-all `except` clause. The answer was a bug in the line that logged the backtrace in that clause. Now we get `bpd.ERROR_SYSTEM` on unexpected exceptions with the proper log message instead of the server crashing!

arcresu:
I've run into an issue with `beets.util.inspect` in that the Python 3 compatibility shim isn't actually compatible with the deprecated Python 2 function in the case of bound methods.

In Python 2 the wrapper is using:
```python
Python 2.7.16 (default, Mar  4 2019, 15:29:09) 
[GCC 8.3.0] on linux2
>>> import inspect
>>> class A:
...   def cmd_a(self, a, b, c): pass
... 
>>> an_a = A()
>>> inspect.getargspec(an_a.cmd_a)
ArgSpec(args=['self', 'a', 'b', 'c'], varargs=None, keywords=None, defaults=None)
```
Whereas in Python 3 it's using:
```python
Python 3.7.3 (default, Mar 26 2019, 07:25:18) 
[GCC 8.3.0] on linux
>>> import inspect
>>> class A:
...   def cmd_a(self, a, b, c): pass
... 
>>> an_a = A(); inspect.signature(an_a.cmd_a)
<Signature (a, b, c)>
>>> sig = inspect.signature(an_a.cmd_a)
>>> sig.parameters
mappingproxy(OrderedDict([('a', <Parameter "a">), ('b', <Parameter "b">), ('c', <Parameter "c">)]))
>>> inspect.getfullargspec(an_a.cmd_a)
FullArgSpec(args=['self', 'a', 'b', 'c'], varargs=None, varkw=None, defaults=None, kwonlyargs=[], kwonlydefaults=None, annotations={})
```
That is, in the shim the `self` parameter is missing from `ArgSpec.args` unlike the original, meaning we get different argument counts depending on the Python version. The [docs](https://docs.python.org/3/library/inspect.html) tell us that `inspect.getfullargspec` is mostly a drop-in replacement for the deprecated function, and indeed replacing the shim with that seems not to cause any issues.

beetsplug/bpd/__init__.py
arcresu: I think this bug has been there for a decade! https://github.com/beetbox/beets/commit/5ee460bcd5c05f6af0651dacbe47d67142018fa7#diff-88d200ed1c440335655e7351f6320344R494

sampsyo: Oh wow! Nice catch. That is extremely depressing—I suddenly feel very old. 👴

sampsyo:
Thank you for looking into the inspect deprecation! I'm really happy the stdlib authors un-deprecated that call. I like this so much better without the wrapper than with it. :tada:

beetsplug/bpd/__init__.py
```diff
@@ -542,12 +614,17 @@ def cmd_seekid(self, conn, track_id, pos):
...
+    # Debugging/testing commands that are not part of the MPD protocol.
+
     def cmd_profile(self, conn):
         """Memory profiling for debugging."""
         from guppy import hpy
         heap = hpy().heap()
         print(heap)
 
+    def cmd_crash_TypeError(self, conn):
+        'a' + 2
```
sampsyo: This looks a little silly, but I understand why it's there. Maybe it could use a quick comment?

beetsplug/bpd/__init__.py
```diff
@@ -670,9 +747,27 @@ def run(self, conn):
...
+                           u'unknown command "{}"'.format(self.name))
         func = getattr(conn.server, func_name)
 
+        if six.PY2:
+            # caution: the fields of the namedtuple are slightly different
+            argspec = inspect.getargspec(func)
+        else:
+            argspec = inspect.getfullargspec(func)
+
+        max_args = len(argspec.args) - 2
```
sampsyo: A short comment here could be useful:

> Check that `func` has the right number of parameters to accept `self.args` arguments.

beetsplug/bpd/__init__.py
arcresu: Done!

beetsplug/bpd/__init__.py
arcresu: Done!

sampsyo:
Wonderful! Thanks again for your attentiveness here! :sparkles:

PR #3205, opened 2019-04-04
bpd: support idle command and with it MPD 0.14

arcresu:
The `idle` command from a client signals that the connection should remain idle until an event occurs in the player, at which point the client should be notified. This PR adds support for the command by making the client's thread poll a list of events. The server can asynchronously populate the event list in response to requests from other clients.

This is the next step for #800 and gets us to MPD 0.14.

Details from the MPD protocol specification:
>Waits until there is a noteworthy change in one or more of MPD’s subsystems. As soon as there is one, it lists all changed systems in a line in the format `changed: SUBSYSTEM`.
>
>Change events accumulate, even while the connection is not in “idle” mode; no events gets lost while the client is doing something else with the connection. If an event had already occurred since the last call, the new idle command will return immediately.
>
>While a client is waiting for idle results, the server disables timeouts, allowing a client to wait for events as long as mpd runs. The idle command can be canceled by sending the command noidle (no other commands are allowed). MPD will then leave idle mode and print results immediately; might be empty at this time. If the optional SUBSYSTEMS argument is used, MPD will only send notifications when something changed in one of the specified subsytems.

Example excerpt of the log (in the new format introduced in this PR) from an mpDris session:
```
bpd: <[127.0.0.1:60926]: idle
bpd: <[127.0.0.1:60926]: noidle
bpd: >[127.0.0.1:60926]: OK
bpd: <[127.0.0.1:60926]: currentsong
bpd: >[127.0.0.1:60926]: OK
bpd: <[127.0.0.1:60926]: status
bpd: >[127.0.0.1:60926]: repeat: 0
bpd: >[127.0.0.1:60926]: random: 0
bpd: >[127.0.0.1:60926]: consume: 0
bpd: >[127.0.0.1:60926]: single: 0
bpd: >[127.0.0.1:60926]: playlist: 0
bpd: >[127.0.0.1:60926]: playlistlength: 0
bpd: >[127.0.0.1:60926]: mixrampdb: 0.0
bpd: >[127.0.0.1:60926]: volume: 10
bpd: >[127.0.0.1:60926]: state: stop
bpd: >[127.0.0.1:60926]: OK
bpd: <[127.0.0.1:60926]: idle
bpd: >[127.0.0.1:60926]: ACK [5@0] {} Got command while idle: close
bpd: *[127.0.0.1:60926]: disconnected
```
(NB: this shows that mpDris sends `close` when it's quit while in idle mode. Here bpd is replying with an error since it's not expecting that. It didn't matter since the client anyway quit, and the protocol says a client should just disconnect without using `close`, which is now deprecated.)

Remaining work:
- [x] ~~See if there's a way to use bluelet here without polling. The only alternative I can think of is to open a new socket we can `select` and use that to wait for the MPD events.~~
- [x] Support the full list of events specified by MPD.
- [x] Work out when events should be fired (the specification is a bit vague here so this will mean checking the MPD source).
- [x] Avoid sending duplicate events (use a `set` for the queued events).
- [x] Cleanup when clients disconnect while `idle`ing (currently the coroutine stays alive in that case).
- [x] Add tests.
- [x] Bump MPD protocol version.
- [x] Test on real clients besides `mpc`.

### DIFF
--- a/beetsplug/bpd/__init__.py
+++ b/beetsplug/bpd/__init__.py
@@ -27,7 +27,6 @@ import random
 import time
 import math
 import inspect
-import socket
 
 import beets
 from beets.plugins import BeetsPlugin
@@ -230,6 +229,14 @@ class BaseServer(object):
         bluelet.run(bluelet.server(self.host, self.port,
                                    Connection.handler(self)))
 
+    def dispatch_events(self):
+        """If any clients have idle events ready, send them.
+        """
+        # We need a copy of `self.connections` here since clients might
+        # disconnect once we try and send to them, changing `self.connections`.
+        for conn in list(self.connections):
+            yield bluelet.spawn(conn.send_notifications())
+
     def _send_event(self, event):
         """Notify subscribed connections of an event."""
         for conn in self.connections:
@@ -301,7 +308,7 @@ class BaseServer(object):
             if system not in SUBSYSTEMS:
                 raise BPDError(ERROR_ARG,
                                u'Unrecognised idle event: {}'.format(system))
-        raise BPDIdle(subsystems)  # put the connection to sleep
+        raise BPDIdle(subsystems)  # put the connection into idle mode
 
     def cmd_kill(self, conn):
         """Exits the server process."""
@@ -709,6 +716,7 @@ class Connection(object):
         self.authenticated = False
         self.address = u'{}:{}'.format(*sock.sock.getpeername())
         self.notifications = set()
+        self.idle_subscriptions = set()
 
     def send(self, lines):
         """Send lines, which which is either a single string or an
@@ -744,49 +752,24 @@ class Connection(object):
         self.server.disconnect(self)
         self.server._log.debug(u'*[{}]: disconnected', self.address)
 
-    def poll_notifications(self, subsystems):
-        """Sleep until we have some notifications from the subsystems given.
-        In order to promptly detect if the client has disconnected while
-        idling, try reading a single byte from the socket. According to the MPD
-        protocol the client can send the special command `noidle` to cancel
-        idle mode, otherwise we're expecting either a timeout or a zero-byte
-        reply. When we have notifications, send them to the client.
-        """
-        while True:
-            mpd_events = self.notifications.intersection(subsystems)
-            if mpd_events:
-                break
-            current_timeout = self.sock.sock.gettimeout()
-            try:
-                self.sock.sock.settimeout(0.01)
-                data = self.sock.sock.recv(1)
-                if data:  # Client sent data when it was meant to by idle.
-                    line = yield self.sock.readline()
-                    command = (data + line).rstrip()
-                    if command == b'noidle':
-                        self.server._log.debug(
-                                u'<[{}]: noidle'.format(self.address))
-                        break
-                    err = BPDError(
-                            ERROR_UNKNOWN,
-                            u'Got command while idle: {}'.format(
-                                command.decode('utf-8')))
-                    yield self.send(err.response())
-                    return
-                else:  # The socket has been closed.
-                    return
-            except socket.timeout:  # The socket is still alive.
-                pass
-            finally:
-                self.sock.sock.settimeout(current_timeout)
-            yield bluelet.sleep(0.02)
-        self.notifications = self.notifications.difference(subsystems)
-        for event in mpd_events:
-            yield self.send(u'changed: {}'.format(event))
-        yield self.send(RESP_OK)
-
     def notify(self, event):
+        """Queue up an event for sending to this client.
+        """
         self.notifications.add(event)
+
+    def send_notifications(self, force_close_idle=False):
+        """Send the client any queued events now.
+        """
+        pending = self.notifications.intersection(self.idle_subscriptions)
+        try:
+            for event in pending:
+                yield self.send(u'changed: {}'.format(event))
+            if pending or force_close_idle:
+                self.idle_subscriptions = set()
+                self.notifications = self.notifications.difference(pending)
+                yield self.send(RESP_OK)
+        except bluelet.SocketClosedError:
+            self.disconnect()  # Client disappeared.
 
     def run(self):
         """Send a greeting to the client and begin processing commands
@@ -813,6 +796,17 @@ class Connection(object):
             for l in line.split(NEWLINE):
                 self.server._log.debug('{}', session + l)
 
+            if self.idle_subscriptions:
+                # The connection is in idle mode.
+                if line == u'noidle':
+                    yield bluelet.call(self.send_notifications(True))
+                else:
+                    err = BPDError(ERROR_UNKNOWN,
+                                   u'Got command while idle: {}'.format(line))
+                    yield self.send(err.response())
+                    break
+                continue
+
             if clist is not None:
                 # Command list already opened.
                 if line == CLIST_END:
@@ -835,7 +829,10 @@ class Connection(object):
                     self.disconnect()  # Client explicitly closed.
                     return
                 except BPDIdle as e:
-                    yield bluelet.call(self.poll_notifications(e.subsystems))
+                    self.idle_subscriptions = e.subsystems
+                    self.server._log.debug('z[{}]: awaiting: {}', self.address,
+                                           ' '.join(e.subsystems))
+            yield bluelet.call(self.server.dispatch_events())
 
     @classmethod
     def handler(cls, server):
--- a/beetsplug/bpd/__init__.py
+++ b/beetsplug/bpd/__init__.py
@@ -767,7 +767,7 @@ class Command(object):
         # by the client (so we can raise ERROR_ARG instead of ERROR_SYSTEM).
         # Maximum accepted arguments: argspec includes "self" and "conn".
         max_args = len(argspec.args) - 2
-        # Minimum accepted arguments: some arguments might be optional/
+        # Minimum accepted arguments: some arguments might be optional.
         min_args = max_args
         if argspec.defaults:
             min_args -= len(argspec.defaults)
--- a/beetsplug/bpd/__init__.py
+++ b/beetsplug/bpd/__init__.py
@@ -187,8 +187,21 @@ class BaseServer(object):
         self.paused = False
         self.error = None
 
+        # Current connections
+        self.connections = set()
+
         # Object for random numbers generation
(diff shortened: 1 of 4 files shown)


arcresu:
I've tested this implementation of `idle` using mpc and mpDris. The latter of these is a realistic client. When interacting with bpd before this PR it repeatedly polls for status, but with these changes it recognises that `idle` is a supported command and makes use of it. Pausing a song with mpDris makes it issue a `noidle` followed by a `pause` command then it goes back into idle mode. It doesn't seem to fully trust the idle mode completely since it periodically breaks out of it and does a manual status request.

There are some subtleties about when events are fired by MPD that we aren't matching here. In practice it doesn't seem to matter since we're sending the events at least as often as MPD does, and redundant events shouldn't cause any problems. For example, the `database` event is supposed to fire only when `update` results in an actual change in the music database, but here we just fire it after every update unconditionally.

In terms of the implementation, the way that this command works doesn't seem to fit very elegantly with eventlet's coroutine model. I was initially hoping that I could avoid polling for events and instead rely on a waitable event. I couldn't find a nice way to do that, especially once I realised that the client is allowed to send `noidle` to cancel the idle mode, or that the client might disconnect while idle. The implementation in the end doesn't seem too messy.

The tests are also a bit hard to manage since they need threading to be able to deal with the blocking request. The tests I've added seem fairly robust, but there's always a chance that they could spuriously fail due to delays in the thread scheduler. If that becomes an issue we can add small delays to fix the tests, or skip them on certain platforms.

beetsplug/bpd/__init__.py
```diff
@@ -72,6 +73,13 @@
     u'close', u'commands', u'notcommands', u'password', u'ping',
 )
 
+SUBSYSTEMS = [
+    u'update', u'player', u'mixer', u'options', u'playlist', u'database',
+    # Related to unsupported commands:
+    # u'stored_playlist', u'output', u'subscription', u'sticker', u'message',
+    # u'partition',
+]
```
sampsyo: A quick comment here to say that subsystems are used for MPD's `idle` command could be helpful.

beetsplug/bpd/__init__.py
```diff
@@ -187,9 +205,28 @@ def __init__(self, host, port, password, log):
         self.paused = False
         self.error = None
 
+        # Current connections
+        self.connections = []
```
sampsyo: Perhaps this should be a set instead of a list (if the order doesn't matter)?

beetsplug/bpd/__init__.py
```diff
@@ -651,9 +719,10 @@ def send(self, lines):
         if isinstance(lines, six.string_types):
             lines = [lines]
         out = NEWLINE.join(lines) + NEWLINE
-        # Don't log trailing newline:
-        message = out[:-1].replace(u'\n', u'\n' + u' ' * 13)
-        self.server._log.debug('server: {}', message)
+        session = u'>[{}]: '.format(self.address)
+        for l in out.split(NEWLINE)[:-1]:
+            l = l.replace('{', '{{').replace('}', '}}')
+            self.server._log.debug(session + l)
```
sampsyo: Here's a way to avoid the need for those `replace`s:

```py
self.server._log.debug('{}', session + l)
```

beetsplug/bpd/__init__.py
```diff
@@ -670,24 +739,81 @@ def do_command(self, command):
...
+                err = BPDError(ERROR_UNKNOWN, u'No command given')
+                yield self.send(err.response())
+                self.disconnect()  # Client sent a blank line.
                 break
             line = line.decode('utf8')  # MPD protocol uses UTF-8.
-            message = line.replace(u'\n', u'\n' + u' ' * 13)
-            self.server._log.debug(u'client: {}', message)
+            for l in line.split(NEWLINE):
+                l = l.replace('{', '{{').replace('}', '}}')
+                self.server._log.debug(session + l)
```
sampsyo: As above: the `replace` dance is probably not necessary.

beetsplug/bpd/__init__.py
```diff
@@ -708,7 +834,10 @@ def run(self):
                 except BPDClose:
                     # Command indicates that the conn should close.
                     self.sock.close()
+                    self.disconnect()  # Client explicitly closed.
                     return
+                except BPDIdle as e:
+                    yield bluelet.call(self.poll_notifications(e.subsystems))
```
sampsyo: Sorry to be slow here, but I don't quite see why the `BPDIdle` exception is necessary. It seems like the idea is that we need to `bluelet.call` the `poll_notifications` method from here, instead of directly calling `poll_notifications` from within the `cmd_idle` handler—but I don't quite follow why *that* is necessary. Can you please explain a little more?

arcresu: I haven't fully wrapped by head around coroutines yet and I quite possibly missed a simpler implementation. I'll have another look because I might be overcomplicating things here.

The `idle` command is fundamentally different to the others though because whereas normally the server waits for a request and then sends a response, in this case it has to delay the response and start waiting on _either_ a new request from the same connection (`noidle`) or an event triggered by a request from a different connection/the player/etc..

The `Server.cmd_*` methods are all called from within `Connection.run` in these lines (where `func` is the bound method like `cmd_idle`):
```python
            args = [conn] + self.args
            results = func(*args)
            if results:
                for data in results:
                    yield conn.send(data)
```
I thought this meant that the command methods can only `yield` bytes to be sent over the connection, but aren't able to hook into the bluelet event loop in any other way or prevent the server from sending an immediate response. The method isn't called with `bluelet.run` so it can at most be a standard generator instead of a coroutine itself right? The new exception seemed necessary to break out of that flow.

beetsplug/bpd/__init__.py
arcresu: Much nicer!

beetsplug/bpd/__init__.py
sampsyo: You're very right! Yes, that answers my question—the `cmd_*` methods are all string generators, not ordinary Bluelet coroutines. Thanks for reminding me of the way my own code works! :smiley:

sampsyo:
To revise the above simplified/maybe-infeasible proposal, in light of the structure of `cmd_*` handlers you mentioned above:

Instead of calling `send` directly from the `notify` method, this could instead just queue up notifications (as it currently does). Instead, the `Connection.run` method, after running any command, will look through *all* the connections in the server and dispatch the notifications by calling `send`. This way, the handlers can remain "string generators" and leave IO to the `run` method.

arcresu:
Thanks for your pointers! I've had a go and managed to get most of the way there. The only remaining problem I can think of is that there's now no way for the player to trigger idle events, which is a necessary feature. That's because the new `Server.dispatch_events` coroutine is only run at the end of a request, but the player can trigger events independently of any requests.

When the player finishes a song it calls the `Server.play_finished()` callback, but this isn't bluelet-aware. Any ideas?

sampsyo:
Aha; that is a very good point! Of course, most of the purpose of doing this in the first place is to be able to send notifications from the player. :smiley:

Here is a somewhat off-the-wall idea that might work: have the player thread (from `finished_callback`, for example) send data _over a socket_ to the main plugin loop. That is, the BPD server would spawn a Bluelet "thread" that waits on a special-purpose socket, which only connects locally to a socket on the "other end" that the player writes to. In other words, use a socket for the inter-thread communication from the player to the server. Any chance that might be tractable?

arcresu:
I played around yesterday and tried two approaches successfully:
1) Spawning a bluelet thread that just calls `dispatch_events` every second. It's still polling, but it's less invasive than what I had initially.
2) Having the player talk to the server over a socket. The difference was that I reused the main socket and had it send a special message that was intercepted before the normal command handling.

I think you're right though that a separate socket would be neater so as not to pollute the real interface. We could even shift over the debugging commands like `profile` to the new socket too (locally I've also been using a simple `nickname` command that lets me assign a label to each open connection so that the log messages become a lot easier to read when there are several clients interacting with bpd at once).

arcresu:
Ok I've implemented a second socket server that I'm calling the control server. It delegates to `ctrl_*` methods on a new `ControlConnection` class, which are full coroutines instead of string generators. This provides a new home for debug commands. I reused as much of the infrastructure as possible so there's not much additional code complexity. I'm starting to think however that some of these classes ought to be moved into a separate module like `beetsplug.bpd.util` so that the actual protocol logic and plugin are easier to find in the main file.

I'll keep testing this out with my normal mpDris2/mpdscribble/mpc mixture, but it seems to be working as intended.

beetsplug/bpd/__init__.py
```diff
@@ -874,7 +1088,11 @@ def play_finished(self):
         """A callback invoked every time our player finishes a
         track.
         """
+        if not self.ctrl_sock:
+            self.ctrl_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            self.ctrl_sock.connect((self.ctrl_host, self.ctrl_port))
         self.cmd_next(None)
+        self.ctrl_sock.sendall(u'play_finished\n'.encode('utf-8'))
```
sampsyo: It might be nice to wrap up the `ctrl_sock` setup and the `sendall` call into a new `ctrl_send` method that conveniently does both.

sampsyo:
Looks perfect! Thank you for all your heroic work on this. Is it ready to merge, in your view?

arcresu:
Thanks for your reviews and ideas :) Yeah I've tried several MPD clients with the new implementation and pushed it around a bit with MPC and `netcat`, and it seems like everything is behaving well. I think this is good to go.

I had the crazy idea that this new control socket could be a way to disentangle the player from bpd. The MPD clients communicate with bpd on the normal socket which handles all the playlist and library stuff, and then bpd translates it into a very minimal protocol on the control socket. Besides cleaner code, it would allow us to control multiple players in a very natural way (each `ControlConnection` is either a player or a debug session). Multiple players is a prerequisite for supporting MPD "partitions", but this is a low priority for me since I don't fully understand the use case. Something to think about in the future anyway.

sampsyo:
That decoupling is a pretty cool idea! The notion could be that a lower-level MPD-like protocol is useful for direct playback control, whereas the higher-level MPD protocol is what you want for proper user interfaces. Seems worth exploring!

In any case, I'm excited about this initial version to support `idle`, so I'll merge this now. :tada: :ship:

PR #3206, opened 2019-04-06
random: move functionality into beets.random

arcresu:
This is a small part of #2321 that simply extracts the functionality of the `random` plugin into a new `beets.util` submodule. The motivation is to be able to share the code in the subsonic plugin, and also potentially in the bpd plugin.

I've added a single test that tests the `_equal_chance_permutation` function by checking the distributions of the index of a particular item when permuting a list of items. To make the test deterministic I added an optional `random_gen` parameter to the relevant functions that allows us to use a pre-seeded `random.Random` object (and in principle an alternative random generator). I don't think this needs a changelog entry since it's just internal rearranging.


sampsyo:
Nice! Thanks for doing this refactoring, which will certainly be useful.

Here's a somewhat bikesheddy organization nit: the modules under `util` currently really are general, non-beets-specific utility modules. Some of the randomization functionality here falls under that category, but the fact that these functions want to assume they have `Album` or `Item` objects (whose length they can obtain) means that this randomization is a little more beets-y than the other utility functions. In that sense, this new module is more similar to `beets.art`, which was refactored out of the `fetchart` and `embedart` plugins some time ago. So maybe this should live in `beets.random` instead of `beets.util.random`?

arcresu:
I hadn't quite understood the demarcation of `beets.util` but your reasoning makes perfect sense :) I suppose the utilities in `random` could be made slightly more generic so that they accept a callable instead of a field name, but there doesn't seem to be a point at the moment if the only use cases are for shuffling `Item`s.

sampsyo:
Excellent! Looks perfect.

PR #3208, opened 2019-04-07
inline: expose item in item_field

arcresu:
In order to be able to load the results of a beets query into a music player we have the `play` plugin. There's a bit of a problem if the music player is bpd however, since `play` gives the real filesystem paths to tracks, whereas bpd expects virtual paths.

There are a couple of ways to go about this, for example just manually recreating the virtual paths using other fields. I wanted a way of getting beets to tell my the paths directly, and it seemed like the `inline` plugin could help. To get the virtual path for an album I was able to do this:
```yaml
album_fields:
    bpd_path: |
      path = items[0].destination(fragment=True)
      return '/'.join(path.split('/')[:-1])
```
This uses `Item.destination` which is exactly what `beets.vfs` uses to construct the virtual path. However when it came to doing the same thing with `item_fields` I got stuck since only the fields are exposed and not the methods. The small change in this PR lets me do:
```yaml
item_fields:
    bpd_path: |
      return item.destination(fragment=True)
```

**Points against**:

- Maybe we don't want this since it's exposing more of the "internal" `Item` API to the configuration file.
- It's also adding a second way to get all of the fields of the item which is potentially confusing ("do I use `artist` or `item.artist`?").
- We could make the vfs structure configurable and/or expose a new field for the virtual path. This could be cleaner and more powerful anyway.
- If we don't do this, is it a problem that real `Item`s are exposed in `album_fields`? Should they be replaced with `namedtuple`s that only expose the fields?

### DIFF
--- a/beetsplug/inline.py
+++ b/beetsplug/inline.py
@@ -102,6 +102,8 @@ class InlinePlugin(BeetsPlugin):
             out = dict(obj)
             if album:
                 out['items'] = list(obj.items())
+            else:
+                out['item'] = obj
             return out
 
         if is_expr:


sampsyo:
There are some very good points here! It's true that this inclusion brings with it a certain amount of danger, which is already present in album-level inline fields via `items`. For example, you could use `item.remove(delete=True)` in an inline declaration, and then evaluating the field could delete files. As you observe, however, this is already a problem with `album_fields`, and it doesn't seem to have caused any widespread data loss. :smiley: It would be a somewhat Pythonic decision to say, "this would be a bad thing to do, but we can trust users not to do that." So I wouldn't be opposed to this change on principle.

About the interaction with `play` specifically: another possibly inelegant solution would be to expose `destination` as a "calculated field," like the current built-in fields $singleton and $filesize. The idea would basically be to make what you want a built-in feature rather than something accomplishable with the inline plugin. Again, it's not terribly elegant, but it would certainly be easy and wouldn't have the above downsides.

arcresu:
Thinking about this a bit more, I think this PR is probably the wrong direction. The other methods on `Item` pretty much all have side-effects and aren't just data. A calculated field for `destination` sounds like a better solution.

I'm tempted also to tweak `album_fields` so that it doesn't expose real `Item`s, for consistency with `item_fields`. If anyone was relying on this internal API then that just points to a useful field we aren't exposing through the intended API, like `destination`.

sampsyo:
OK, sounds good! I do like the idea of making the `items` exposure in `album_fields` a little safer by exposing a flattened data object instead of the Item objects themselves.

jtpavlock:
Closing since it seems this implementation has been decided against.

PR #3207, opened 2019-04-06
mpdstats: use currentsong instead of playlist

arcresu:
This is a copy of #2754, rebased and with the additional small change required to the test. Ive also tested it using bpd (with the changes from my `bpd-idle` branch) and it appears to be working well.

Closes #2754, closes #2752

PR #2707, opened 2017-10-07
mpdstats: update stats when switching from song to stream and when playing a song consecutively 

gdtwst:
Fixes an issue where mpdstats won't update play/skip counts of a song when MPD switches from the song to an internet stream. Also updates play counts for a song being played consecutively.

PR #3210, opened 2019-04-07
mpdstats: Re-add fixes from #2707

gdtwst:
#3207 seemed to have deleted lines related to #2707 which allowed mpdstats to update songs ratings when playing the same song consecutively as well as when switching from a song to an internet stream.  Re-added the lines and it seems to be working fine with the new currentsong method.

### DIFF
--- a/beetsplug/mpdstats.py
+++ b/beetsplug/mpdstats.py
@@ -256,10 +256,6 @@ class MPDStats(object):
         if not path:
             return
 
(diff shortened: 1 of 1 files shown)


arcresu:
Ah, sorry! Totally my bad for not noticing that unrelated lines were being touched when I rebased that commit. Thanks for the rapid PR!

Issue #3007, opened 2018-08-19
beets bpd crash


hrehfeld:
### Problem

beets bpd crashes when accessed with mpd client (while fetching albums for artist?).

```sh
$ beet -vv bpd
user configuration: /home/hrehfeld/.config/beets/config.yaml
data directory: /home/hrehfeld/.config/beets
plugin paths: 
Sending event: pluginload
library database: /home/hrehfeld/.local/beets.blb
library directory: /home/hrehfeld/music/2017
Sending event: library_opened
Building directory tree...
... done.
```

Now access with mpd client:
```sh
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/beetsplug/bpd/__init__.py", line 686, in run
    for data in results:
  File "/usr/lib/python3.7/site-packages/beetsplug/bpd/__init__.py", line 1060, in cmd_list
    rows = tx.query(statement, subvals)
  File "/usr/lib/python3.7/site-packages/beets/dbcore/db.py", line 709, in query
    cursor = self.db._connection().execute(statement, subvals)
sqlite3.OperationalError: near "ORDER": syntax error

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/beet", line 11, in <module>
    load_entry_point('beets==1.4.7', 'console_scripts', 'beet')()
  File "/usr/lib/python3.7/site-packages/beets/ui/__init__.py", line 1256, in main
    _raw_main(args)
  File "/usr/lib/python3.7/site-packages/beets/ui/__init__.py", line 1243, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "/usr/lib/python3.7/site-packages/beetsplug/bpd/__init__.py", line 1192, in func
    self.start_bpd(lib, host, int(port), password, volume, debug)
  File "/usr/lib/python3.7/site-packages/beetsplug/bpd/__init__.py", line 1168, in start_bpd
    server.run()
  File "/usr/lib/python3.7/site-packages/beetsplug/bpd/__init__.py", line 762, in run
    super(Server, self).run()
  File "/usr/lib/python3.7/site-packages/beetsplug/bpd/__init__.py", line 196, in run
    Connection.handler(self)))
  File "/usr/lib/python3.7/site-packages/beets/util/bluelet.py", line 380, in run
    exit_te.reraise()
  File "/usr/lib/python3.7/site-packages/beets/util/bluelet.py", line 204, in reraise
[...]

Issue #176, opened 2013-02-28
BPD: Persistent playlist


ghost:
This issue was automatically migrated from Google Code.
_Original author: voigh...@gmail.com (November 07, 2012 08:31:56)_
_Original issue: https://github.com/google-code-export/beets/issues/458_

PR #3214, opened 2019-04-13
bpd: support MPD 0.16 protocol and more clients

### LINKED ISSUE #800: BPD: Update to newer MPD protocol version

arcresu:
Now that `idle` has been implemented, we're almost all the way to MPD 0.16 support. There are a couple of small commands that need adding, but most of what we need is already in place. More importantly, claiming to support the newer protocol version means that more clients will talk to us; fixes #800.

- [x] `nextsong` and `nextsongid` fields in `status` output.
- [x] playlist range support in `playlistinfo`.
- [ ] playlist range support in `move`.
- [x] ~~stickers (which are basically flexible attrs).~~ Moved to #3290

I've expanded my local tests to include ncmpcpp and two Android clients (MPDroid and MALP). So far bpd breaks all three of these clients. Known compatibility issues are listed below.

#### ncmpcpp (Linux, libmpdclient)
- [x] ncmpcpp sends `noidle` when bpd doesn't expect it to be in the idle state.
- [x] ncmpcpp uses the deprecated `volume` command to change the volume and we ignore it.
- [x] ncmpcpp sometimes causes an error with errno 104 `ECONNRESET` when quitting.
- [x] ncmpcpp issues the unimplemented command `decoders` (fixed in #3222).
- [x] ~~It notices when playback is paused, but doesn't notice when it's resumed again, even if it was resumed using ncmpcpp itself! Issuing a different command (e.g. changing the volume) will prompt it to refresh the status and update the playback state.~~ This doesn't seem to happen any more. I'm not sure what fixed it.
- [x] ~~ncmpcpp expects stored playlists to be supported and when bpd sends an error it gets stuck in an infinite loop and the UI freezes.~~ Moved to #176.
- [ ] When it receives a `playlist` event and tries to update its playlist using `plchanges` it doesn't like the response returned by bpd and ncmpcpp's playlist becomes blank.

#### M.A.L.P. (Android)
- [x] I'm seeing the issue in #3007 when using MALP (fixed in #3215).
- [x] MALP would like to have extra tags, especially the MusicBrainz ones.

#### MPDroid (Android)
- [x] MPDroid uses extra idle events that we don't support yet (`idle "database" "mixer" "options" "output" "player" "playlist" "sticker" "update"`).
- [ ] MPDroid crashes at startup although there is no MPD error sent over the protocol. We must be giving it wrong output in the successful commands somehow.

#### mpDris2 (Linux)
- [x] mpDris2 seeks using fractional seconds and crashes when bpd doesn't accept that.
- [ ] mpDris2 (and another client but I've forgotten which) crash sometimes with an error that implies that bpd forgot the `state` field in a response to the `status` command. This doesn't seem to be possible though. Its backtrace starts with `KeyError: 'state'`

#### mpdscribble (Linux, libmpdclient)
- [x] ~~mpdscribble tries to register a client channel. It doesn't mind that it fails, but it would be nice to support it.~~ Moved to #3291
- [ ] I noticed it stopped scrobbling and there was an error in the log: `mpd error (7): expected more list_OK's`. I didn't manage to correlate it to the protocol log in bpd though so I'm not sure what happened.

#### Other clients
- [x] ~~beets' own mpdupdate plugin doesn't work with bpd since it expects output from the `update` command, but we never send any.~~ Moved to #3292

### DIFF
--- a/beets/util/bluelet.py
+++ b/beets/util/bluelet.py
@@ -346,6 +346,10 @@ def run(root_coro):
                             exc.args[0] == errno.EPIPE:
                         # Broken pipe. Remote host disconnected.
                         pass
+                    elif isinstance(exc.args, tuple) and \
+                            exc.args[0] == errno.ECONNRESET:
+                        # Connection was reset by peer.
+                        pass
                     else:
                         traceback.print_exc()
                     # Abort the coroutine.
--- a/beetsplug/bpd/__init__.py
+++ b/beetsplug/bpd/__init__.py
@@ -77,8 +77,8 @@ SAFE_COMMANDS = (
 SUBSYSTEMS = [
     u'update', u'player', u'mixer', u'options', u'playlist', u'database',
     # Related to unsupported commands:
-    # u'stored_playlist', u'output', u'subscription', u'sticker', u'message',
-    # u'partition',
+    u'stored_playlist', u'output', u'subscription', u'sticker', u'message',
+    u'partition',
 ]
 
 ITEM_KEYS_WRITABLE = set(MediaFile.fields()).intersection(Item._fields.keys())
--- a/beetsplug/bpd/__init__.py
+++ b/beetsplug/bpd/__init__.py
@@ -583,23 +583,25 @@ class BaseServer(object):
         """Indicates supported URL schemes. None by default."""
         pass
 
-    def cmd_playlistinfo(self, conn, index=-1):
+    def cmd_playlistinfo(self, conn, index=None):
         """Gives metadata information about the entire playlist or a
         single track, given by its index.
         """
-        index = cast_arg(int, index)
-        if index == -1:
+        if index is None:
             for track in self.playlist:
                 yield self._item_info(track)
         else:
+            indices = self._parse_range(index, accept_single_number=True)
             try:
-                track = self.playlist[index]
+                tracks = [self.playlist[i] for i in indices]
             except IndexError:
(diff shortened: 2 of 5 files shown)


arcresu:
I've been using this branch myself for some time now and it's mostly working OK. The main issues are because of parts of the MPD protocol we're missing, but I'd prefer to handle those additions separately.

The other issue I've seen is that after running fine for some time, clients connected to bpd will sometimes crash. I caught mpDris2 and ncmpcpp crashing at the same time with log output that suggested a malformed response from bpd. This is rare enough that I haven't been able to correlate it with the specific output in bpd's protocol traffic dump. Whenever this has happened I've only needed to restart the clients to get things back to normal, without having to restart bpd.

In the course of implementing this PR I encountered issues that went away when I implemented support for apparently unrelated protocol features. I guess the clients aren't robust to unexpected server behaviour since there's normally only one server implementation to test against. I'm hopeful that the issue I mentioned above will solve itself in the same way.

So I think I'd like to merge this now and keep working on the unchecked todo items during the release cycle. This should hopefully encourage some other people to test the changes too :)

arcresu:
Alright, I've opened a bunch of issues to keep track of missing features, and collected them in a new milestone. Let's get this in :tada:

Issue #3219, opened 2019-04-17
LastGenre for Singletons?


rain0r:
Is it possible to fetch genre tags for singletons after the import?

Currently, `beet lastgenre --force --source=track singleton:true` is not doing anything, but when I import new singletons, the plugin fetches genre tags.

sampsyo:
Can you include the verbose output?

sampsyo:
Aha, actually, belay that! Looking back at the code, it looks like the command _only_ matches albums, not individual tracks. That's an artificial limitation, of course, so it shouldn't be too hard to add a track-matching mode if anyone's interetsted.

rain0r:
... or just do the magic that happens when doing an `beet import -s`?

Could you point me to the corresponding location in the the code, please?

rain0r:
I think i found it.
https://github.com/beetbox/beets/pull/3220

PR #3220, opened 2019-04-17
Update __init__.py

rain0r:
Also fetch genres for single tracks via query.

### DIFF
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -14,6 +14,8 @@
 # included in all copies or substantial portions of the Software.
 
 from __future__ import division, absolute_import, print_function
+
+import ipdb
 import six
 
 """Gets genres for imported music based on Last.fm tags.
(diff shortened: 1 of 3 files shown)


sampsyo:
Hi! This is a nice start, but doing the same actions to *both* albums and items matching the *same* query is likely to do a bunch of redundant work in most cases. For example, the query `beatles` will match both some albums and the tracks on those albums, so the command would fetch the genres for those tracks twice.

I think the right thing to do is to put this behind a command-line option: one that switches between album and item mode. As you'll see in most other plugin-provided commands, there's an `-a` flag that switches to album mode. We could do the same thing, but make this the default—and require something like `-A` to switch back to item mode.

rain0r:
Added the parser options, what do you think?

sampsyo:
Cool! This arrangement, however, does make it possible for albums and tracks to be enabled _simultaneously_. It would be nice if this were a proper switch: i.e., track and album mode negated one another, and only one could be active at a time.

Also, can you please call the track mode "item" mode, for uniformity with other commands?

Finally, could you please look into the Travis failures, add documentation, and write a quick changeling entry?

rain0r:
> Cool! This arrangement, however, does make it possible for albums and tracks to be enabled _simultaneously_. It would be nice if this were a proper switch: i.e., track and album mode negated one another, and only one could be active at a time.
> 
> Also, can you please call the track mode "item" mode, for uniformity with other commands?
> 
> Finally, could you please look into the Travis failures, add documentation, and write a quick changelog entry?

Done

sampsyo:
Looking good! If you can bear with me, I have a few more suggestions:

- A clean way to do the "mutual excursion" thing would be to have the CLI flags control a single option variable. (This is the case in most other plugins and commands, if you take a look at that code.) To both option constructors, you can add `dest='album'`. Then, the `-A` option can use `action='store_false'` and the `-a` one can use `action='store_true'`. Then, the code for the command can just use `if opts.album:` and `else:` for the two cases.
- In the documentation, can you please explain how to use the new switch? That is, say that the command matches albums by default. If you want to look up genres for individual tracks (and singletons) instead, use `-A`.
- Please see the way that other plugin features are described in the changeling. We use ReST syntax like this: ``:doc:`/plugins/lastgenre`: Some change here.``.

rain0r:
No problem, I'm happy if someone has an eye on code quality ;-)

Now it works like this:
```
$ beet lastgenre -a
# albums
$ beet lastgenre -A
# items 
$ beet lastgenre 
# albums
```
Also updated the doc.

rain0r:
If this goes through, #3219 can also be closed.

sampsyo:
Awesome! Thank you for your persistent work on this. :fish: Merged!

JOJ0:
Renamed to a more self-explanatory name even though it's an older PR. Since I'm working on something related to this feature, I might reference it and appreciate a better name.

PR #3223, opened 2019-04-19
mpdupdate, mpdstats: use MPD_PORT env variable

arcresu:
Most MPD clients support the environment variables `MPD_HOST` and `MPD_PORT`. The two MPD client plugins in beets supported the former but not the latter. This change will allow the correct setting to be picked up automatically in more cases, while still allowing the port to be overridden in the config file.


sampsyo:
Looks perfect to me; thank you!! :sparkles:

PR #2061, opened 2016-06-20
Replace confit with confuse

jackwilsdon:
Use [sampsyo/confuse](https://github.com/sampsyo/confuse) (previously named confit) throughout beets instead of having our own copy of it.

This is part of #1966 that has been separated at the request of @jrobeson. Go there if you would like to see our discussion so far.

**Note**: [This PR should not be merged until we've notified packagers of the new dependency](https://github.com/beetbox/beets/pull/1966#issuecomment-225643360) (i.e. the 1st of August and 1.4.0).

PR #1966, opened 2016-04-28
Modularise beets

jackwilsdon:
This is a work-in-progress PR and should not be merged.

The goal of this PR is to modularise beets in such a way that the larger parts of beets can be re-used by external applications, as well as to reduce duplication (e.g. [`confit.py`](https://github.com/beetbox/beets/blob/91c8a30e55b38b507a763c3750a7dea94c5c1147/beets/util/confit.py) and [`sampsyo/confit`](https://github.com/sampsyo/confit/blob/7f5ab69e825060c38e9558ff623ed0669cb7c3ce/confit.py)).
### Current Progress
- [x] Use Confit from the `confuse` library (~~added in fb547a162dfe313ca0794c58d1da1d935e72fbaa, improved in 9568a1d8fc9762a53aacdf5d18caf822a8aeef5b~~ see #2061).
- [ ] Separate MediaFile ([`mediafile.py`](https://github.com/beetbox/beets/blob/91c8a30e55b38b507a763c3750a7dea94c5c1147/beets/mediafile.py)) into it's own library (and repository belonging to [`beetbox`](https://github.com/beetbox))
- [ ] Separate dbcore ([`dbcore`](https://github.com/beetbox/beets/tree/91c8a30e55b38b507a763c3750a7dea94c5c1147/beets/dbcore)) into it's own library (and repository belonging to [`beetbox`](https://github.com/beetbox))

This PR was inspired by @sampsyo's comment [here](https://github.com/beetbox/beets/pull/1961#issuecomment-215266618):

> The only reasons we haven't taken it as a dependency are the logistics: there's some friction for OS packagers to catalog a new dependency, and we'd need to stabilize the Confit API. The latter entails starting a Changelog, and because the Confit name is taken on PyPI, changing the module name to confuse.
> 
> It's definitely a worthwhile endeavor! It will just take a little more effort than deleting the file. Perhaps we should do the same thing for Confit, MediaFile, and dbcore simultaneously?

I created the [`modularise`](https://github.com/beetbox/beets/tree/modularise) branch so that anyone can contribute to this PR by creating another PR for the [`modularise`](https://github.com/beetbox/beets/tree/modularise) branch.

PR #3224, opened 2019-04-19
confit: replace with a re-export of beetbox/confuse

arcresu:
This re-exports all of the [confuse](https://github.com/sampsyo/confuse) module under `beets.util.confit` and patches in the minimum subset of commits to confit that have not yet been ported to confuse such that the test suite passes. A warning is issued when importing from the deprecated namespace.

Remaining work:
  - [x] Port the outstanding commits from confit to confuse (sampsyo/confuse#48, sampsyo/confuse#49, sampsyo/confuse#50).
  - [x] Make a new release of confuse.
  - [x] Update the minimum dependency of beets on confuse to that new release
  - [x] Remove the patches added here (including the change to flake8 config).
  - [x] Remove all uses of the deprecated namespace from beets itself (this should be left until later so as to avoid picking up many conflicts in this branch while the previous steps happen).
  - [x] Add changelog entry.
  - [x] Check if there are any irrelevant tests and documentation (including the wiki).
  - [ ] Notify downstream packagers about the new dependency (although I don't think that we need to block a beets release on this as discussed in https://github.com/beetbox/beets/pull/2061#issuecomment-480565229).

Supersedes and closes #2061. Part of #1966.


sampsyo:
Wow!! This is truly hard work, but very important despite being user-invisible. Thank you for taking this on! :horse:

It looks like we're pretty close to having the Confuse repository in sync. I think we should do this in the next release.

arcresu:
There are no currently-open PRs that would be affected by pulling out Confit (other than the older modularisation PR).

---
Here's the command to check in future (relies on having [PRs checked out locally](https://stackoverflow.com/a/43019825/759418)):
```
git for-each-ref --no-merged=upstream/master refs/remotes/upstream/pr/ --format='%(refname)' | while read r; do if ! git diff --quiet upstream/master...$r -- beets/util/confit.py; then echo ${r##*/}; fi; done | sort -n | sed -e's/^/- [ ] #/g'
```

sampsyo:
I'd love to address this next as part of the modularization plan! I've submitted Confuse v1.0.0 to PyPI (and tagged it on GitHub), so the next step seems to be to try depending that in our own setup.py.

arcresu:
Great! Thanks for taking care of yet another release :) I've updated the dependency on confuse and removed the backports. The tests pass locally for me and if it also looks good on the CI I'll make the additional changes needed to finish off this PR.

beets/ui/commands.py
```diff
@@ -39,7 +39,7 @@
 from beets import library
 from beets import config
 from beets import logging
-from beets.util.confit import _package_path
```
arcresu: This has been importing the internal `_package_path` function. Now that confuse is a separate library this seems like the wrong thing to do. Either that functionality should be exposed publicly in confuse or we should find another way to do this.

arcresu:
Ok, I think we're good to go! :tada:

beets/ui/commands.py
sampsyo: Indeed. The (sole) use of this could probably be easily replaced with a more standard `os.path.dirname(__file__)` sort of construction.

arcresu: Great, then I'll sort this small change out separately and we can merge this now!

PR #3225, opened 2019-04-19
Replace more instances of unsafe calls to yaml.load

sbraz:
Disclaimer: I don't use beets. But these changes should be simple enough that they don't break anything.


arcresu:
Thanks! I think this looks like a sensible change we needed to make :+1:

PR #3226, opened 2019-04-20
Add loadext plugin

### LINKED ISSUE #3160: String queries should be case-insensitive even for non-ASCII characters
### Problem

When searching for metadata written in English or, in other words, in latin script, `beet ls` performs case-insensitive search, which is very nice. With non-latin scripts (I've checked with cyrillic, since it is the only one in which I have tracks and which has the case concept),  however, case-insensitive search does not work.

Running this command:

```sh
$ beet ls сплин
```

outputs nothing, while this one:

```sh
$ beet ls Сплин
```

shows all the tagged files correctly.

I would expect this to work for any kind of scripts, probably based on unicode case transformations.

### Setup

* OS: Archlinux
* Python version: 3.7.2
* beets version: 1.4.7

```yaml
lyrics:
    bing_lang_from: []
    auto: yes
    bing_client_secret: REDACTED
    bing_lang_to:
    google_API_key: REDACTED
    google_engine_ID: REDACTED
    genius_api_key: REDACTED
    fallback:
    force: no
    local: no
    sources:
    - google
    - lyricwiki
    - musixmatch
    - genius
directory: ~/Music
library: ~/.local/share/beets/music.db
sort_item: artist+ year+ album+ disc+ track+
original_date: yes

ui:
    color: yes

import:
    move: no
    log: ~/.local/share/beets/import.log
    incremental: yes

paths:
    default: $albumartist/$year - $album%aunique{albumartist year album}/%if{$multidisc,Disc $disc/}$track - $title
    singleton: Singletons/$artist - $title
item_fields:
    multidisc: 'True if disctotal > 1 else False

        '

plugins: chroma fromfilename edit embedart fetchart lyrics zero bpd convert info export alternatives types inline web
chroma:
    auto: no
embedart:
    compare_threshold: 50
    maxwidth: 0
    auto: yes
    ifempty: no
    remove_art_file: no
fetchart:
    store_source: yes
    auto: yes
    minwidth: 0
    maxwidth: 0
    enforce_ratio: no
    cautious: no
    cover_names:
    - cover
    - front
    - art
    - album
    - folder
    sources:
    - filesystem
    - coverart
    - itunes
    - amazon
    - albumart
    google_key: REDACTED
    fanarttv_key: REDACTED
zero:
    fields: month day
    auto: yes
    keep_fields: []
    update_database: no
bpd:
    host: 127.0.0.1
    port: 6600
    password: REDACTED
    volume: 100
convert:
    format: opus
    dest:
    pretend: no
    threads: 12
    formats:
        aac:
            command: ffmpeg -i $source -y -vn -acodec aac -aq 1 $dest
            extension: m4a
        alac:
            command: ffmpeg -i $source -y -vn -acodec alac $dest
            extension: m4a
        flac: ffmpeg -i $source -y -vn -acodec flac $dest
        mp3: ffmpeg -i $source -y -vn -aq 2 $dest
        opus: ffmpeg -i $source -y -vn -acodec libopus -ab 96k $dest
        ogg: ffmpeg -i $source -y -vn -acodec libvorbis -aq 3 $dest
        wma: ffmpeg -i $source -y -vn -acodec wmav2 -vn $dest
    max_bitrate: 500
    auto: no
    tmpdir:
    quiet: no
    embed: yes

    paths: {}
    no_convert: ''
    copy_album_art: no
    album_art_maxwidth: 0
types:
    phone_export: bool
alternatives:
    phone:
        directory: ~/tmp/phone-music
        formats: opus mp3 ogg
        query: phone_export:true
export:
    default_format: json
    json:
        formatting:
            ensure_ascii: no
            indent: 4
            separators: [',', ': ']
            sort_keys: yes
edit:
    albumfields: album albumartist
    ignore_fields: id path
pathfields: {}
album_fields: {}
web:
    host: 127.0.0.1
    port: 8337
    cors: ''
    cors_supports_credentials: no
    reverse_proxy: no
    include_paths: no
```

jackwilsdon:
This is a rough draft at the moment of what a `loadext` plugin could look like for loading SQLite extensions. This plugin can be used to fix #3160 by allowing users to load the ICU plugin like so;

```
loadext:
  - libicu
```

Note that this will require the user to compile their own copy of `ext/icu` from SQLite sources.

## Points of Discussion
I've got a few points of discussion I'd like to bring up regarding this PR;

1. The elephant in the room;
 
    ```Python
    # TODO: Find a way of getting a connection without calling a private method.
    conn = lib._connection()
    ```

    This is certainly less than ideal, as we're reaching into beets internals from this plugin. There ~are two~ is one possible solution~s~ I can think of to this;

    * ~Use `SELECT load_extension(...)`~ This doesn't work properly, see [this thread](http://sqlite.1065341.n5.nabble.com/Error-on-loading-ICU-extension-td41219.html) regarding the issues.
    * Add a `load_extension` method to `dbcore.Database`.

2. Where should these extension paths be relative to? In this initial draft the paths are relative to wherever `beets` is running, and as such running beets in different folders breaks this unless you use an absolute path for the extension. I think using the beets config directory for relative paths is probably the best idea, but I'll leave this open to others to decide.

3. How in depth do we want to go into the documentation with this? This plugin seems like a somewhat advanced feature, as it requires the user to compile their own `ext/icu` extension from the SQLite sources.

4. How do we test this? Do we mock out the `load_extension` method and make sure it's called? Or do we compile the ICU extension on Travis and attempt to call one of the methods provided by it?

## Compiling the ICU extension
To compile the ICU extension, you need the following;

 * gcc
 * icu-devtools
 * libicu-dev
 * libicu60 (or any other version I guess?)
 * libsqlite3-dev

```Shell
$ wget https://sqlite.org/2019/sqlite-src-3280000.zip
$ unzip sqlite-src-3280000.zip
$ cd sqlite-src-3280000/ext/icu
$ gcc -shared -fPIC icu.c `icu-config --ldflags` -o libicu.so
$ cp libicu.so [wherever you want]
```

### DIFF
--- a/beetsplug/loadext.py
+++ b/beetsplug/loadext.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+# This file is part of beets.
+# Copyright 2019, Jack Wilsdon <jack.wilsdon@gmail.com>
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+
+"""Load SQLite extensions.
+"""
+
+from __future__ import division, absolute_import, print_function
+
+from beets.dbcore import Database
+from beets.plugins import BeetsPlugin
+import sqlite3
+
+
+class LoadExtPlugin(BeetsPlugin):
+    def __init__(self):
+        super(LoadExtPlugin, self).__init__()
+
+        if not Database.supports_extensions:
+            self._log.warn('loadext is enabled but the current SQLite '
+                           'installation does not support extensions')
+            return
+
+        self.register_listener('library_opened', self.library_opened)
+
+    def library_opened(self, lib):
+        for v in self.config:
+            ext = v.as_filename()
+
+            self._log.debug(u'loading extension {}', ext)
+
+            try:
+                lib.load_extension(ext)
+            except sqlite3.OperationalError as e:
+                self._log.error(u'failed to load extension {}: {}', ext, e)
--- a/docs/plugins/loadext.rst
+++ b/docs/plugins/loadext.rst
@@ -1,0 +1,53 @@
+Load Extension Plugin
+=====================
+
+Beets uses an SQLite database to store and query library information, which
+has support for extensions to extend its functionality. The ``loadext`` plugin
+lets you enable these SQLite extensions within beets.
+
+One of the primary uses of this within beets is with the `"ICU" extension`_,
+which adds support for case insensitive querying of non-ASCII characters.
+
+.. _"ICU" extension: https://www.sqlite.org/src/dir?ci=7461d2e120f21493&name=ext/icu
+
+Configuration
+-------------
+
+To configure the plugin, make a ``loadext`` section in your configuration
+file. The section must consist of a list of paths to extensions to load, which
+looks like this:
+
+.. code-block:: yaml
+
+    loadext:
+      - libicu
+
+If a relative path is specified, it is resolved relative to the beets
+configuration directory.
+
+If no file extension is specified, the default dynamic library extension for
+the current platform will be used.
+
+Building the ICU extension
+--------------------------
+This section is for **advanced** users only, and is not an in-depth guide on
+building the extension.
+
+To compile the ICU extension, you will need a few dependencies:
+
+ - gcc
+ - icu-devtools
+ - libicu
+ - libicu-dev
+ - libsqlite3-dev
+
+Here's roughly how to download, build and install the extension (although the
+specifics may vary from system to system):
+
+.. code-block:: shell
+
+    $ wget https://sqlite.org/2019/sqlite-src-3280000.zip
+    $ unzip sqlite-src-3280000.zip
+    $ cd sqlite-src-3280000/ext/icu
+    $ gcc -shared -fPIC icu.c `icu-config --ldflags` -o libicu.so
+    $ cp libicu.so ~/.config/beets
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -850,9 +850,13 @@ class Database(object):
     """A container for Model objects that wraps an SQLite database as
     the backend.
     """
+
     _models = ()
     """The Model subclasses representing tables in this database.
     """
+
+    supports_extensions = hasattr(sqlite3.Connection, 'enable_load_extension')
+    """Whether or not the current version of SQLite supports extensions"""
 
     def __init__(self, path, timeout=5.0):
         self.path = path
@@ -860,6 +864,7 @@ class Database(object):
 
         self._connections = {}
         self._tx_stacks = defaultdict(list)
+        self._extensions = []
 
         # A lock to protect the _connections and _tx_stacks maps, which
         # both map thread IDs to private resources.
@@ -909,6 +914,13 @@ class Database(object):
             py3_path(self.path), timeout=self.timeout
         )
 
+        if self.supports_extensions:
+            conn.enable_load_extension(True)
+
+            # Load any extension that are already loaded for other connections.
+            for path in self._extensions:
+                conn.load_extension(path)
+
         # Access SELECT results like dictionaries.
         conn.row_factory = sqlite3.Row
         return conn
@@ -936,6 +948,18 @@ class Database(object):
         with the underlying SQLite database.
         """
         return Transaction(self)
+
+    def load_extension(self, path):
+        """Load an SQLite extension into all open connections."""
+        if not self.supports_extensions:
+            raise ValueError(
+                    'this sqlite3 installation does not support extensions')
+
+        self._extensions.append(path)
+
+        # Load the extension into every open connection.
+        for conn in self._connections.values():
+            conn.load_extension(path)
 
     # Schema setup and migration.
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -94,6 +94,9 @@ New features:
   :bug:`3205` :bug:`800`
 * :doc:`/plugins/bpd`: MPD protocol command ``decoders`` is now supported.
   :bug:`3222`
+* The new :doc:`/plugins/loadext` allows loading of SQLite extensions, primarily
+  for use with the ICU SQLite extension for internationalization.
+  :bug:`3160` :bug:`3226`
 
 Changes:
 
--- a/docs/plugins/index.rst
+++ b/docs/plugins/index.rst
@@ -71,6 +71,7 @@ like this::
    kodiupdate
    lastgenre
    lastimport
+   loadext
    lyrics
    mbcollection
    mbsubmit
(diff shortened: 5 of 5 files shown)


sampsyo:
Awesome!! Thanks for getting this started!

Here are a few thoughts on your questions:

1. A `load_extension` method in `Database` sounds like just the right thing. :smiley:
2. I think relative to the beets configuration directory would be the most predictable. In fact, that's what Confuse's [as_filename](https://github.com/sampsyo/confuse/blob/35025399b9fb801ba178260c7df8a9512e98db61/confuse.py#L481) will do by default. It makes sense to recommend that users put their extensions in that directory.
3. Yeah, it does seem pretty advanced. I think a good place to start would be to leave it up to advanced users to accomplish for themselves, mostly—you could include your snippet about how to compile the ICU extension, for example, but surround it with language that says "this is for advanced users, but you will want to do _something_ like this, although the details will vary based on your system."
4. Perhaps this is heresy, but I think we can get away without tests for something this simple (and this hard to test)? Compiling the ICU extension on Travis sounds like a little bit of a nightmare, although it's possible…

jackwilsdon:
I've just realised that connections in beets are per-thread, which opens up an issue of the extension being loaded on one thread but not another. Would we be alright iterating through all of the open connections and loading the extension for all of them?

sampsyo:
Aha; I didn't realize connections were per thread! Maybe the right thing to do, then, would be for the `Database` to keep track of which plugins the application needs and to load them whenever it creates a new connection?

jackwilsdon:
Yeah that makes sense to me. sqlite3 itself doesn't do a connection per thread, but it seems our code is doing it instead.

https://github.com/beetbox/beets/blob/0f8a748ade88751a86108a1e19bb0f84c2da3c66/beets/dbcore/db.py#L885-L896

beetsplug/loadext.py
```diff
@@ -0,0 +1,40 @@
...
+
+class LoadExtPlugin(BeetsPlugin):
+    def __init__(self):
+        super(LoadExtPlugin, self).__init__()
+
+        self.register_listener('library_opened', self.library_opened)
+
+    def library_opened(self, lib):
+        for i in range(len(self.config.get(list))):
+            ext = self.config[i].as_filename()
```
jackwilsdon: Is there a better way to get a list of filenames? I'm not a fan of the `range(len(self.config.get(list)))`.

beetsplug/loadext.py
```diff
@@ -0,0 +1,40 @@
...
+
+            self._log.debug(u'loading extension {}', ext)
+
+            try:
+                lib.load_extension(ext)
+            except sqlite3.OperationalError as e:
```
jackwilsdon: Are we alright catching an SQLite-specific exception in a plugin? Or should a special type of exception be thrown by `dbcore.Database`?

beetsplug/loadext.py
sampsyo: Yep! This should work:

```py
for v in self.config:
    ext = v.as_filename()
```

beetsplug/loadext.py
sampsyo: Yeah, I think that's fine! There's no hiding the fact that this is an SQLite operation under the hood.

sampsyo:
It's true—the reason we do that is that SQLite doesn't allow connections to be shared across threads. See the docs for the [check_same_threads](https://docs.python.org/3/library/sqlite3.html#sqlite3.connect) option for some more details.

Anyway, this looks awesome!! Seems like this should totally work. :smiley:

jackwilsdon:
[Just to put another spanner in the works](https://docs.python.org/3/library/sqlite3.html#f1).

Oddly enough it's enabled on the versions of Python 2 and 3 I have installed from Homebrew on macOS, as well as on Python 2 and 3 I have installed from `apt` on Ubuntu.

My thoughts are now as follows:
1. Add a `supports_extensions` property to `dbcore.Database`, which is backed by `hasattr(sqlite3.Connection, 'enable_load_extension')`
2. Throw an exception when `dbcore.Database#load_extension` is called if `supports_extensions` is false
3. Check if `supports_extensions` is true when `loadext` starts, logging a warning (or error?) if `loadext` is enabled but the user's Python SQLite installation does not support extensions

What do you think @sampsyo?

sampsyo:
Wow; that's really too bad! So close, and yet so far.

All three of your proposals sound right to me. I think it's fine for `loadext` to warn and do nothing if it detects that extensions are not supported.

jackwilsdon:
Alright I think this is ready for a review now 😁 I'm not exactly great with documentation, so any feedback there is greatly appreciated!

docs/plugins/loadext.rst
```diff
@@ -0,0 +1,53 @@
+Load Extension Plugin
+=====================
+
+Beets uses an SQLite database to store and query library information, which
+has support for extensions to extend its functionality.
```
sampsyo: Maybe an additional sentence here:

> The `loadext` plugin lets you enable these SQLite extensions for beets.

docs/plugins/loadext.rst
```diff
@@ -0,0 +1,53 @@
...
+One of the primary uses of this within beets is with the `"ICU" extension`_,
+which adds support for case insensitive querying of non-ASCII characters.
+
+.. _"ICU" extension: https://www.sqlite.org/src/dir?ci=7461d2e120f21493&name=ext/icu
+
+Configuration
+-------------
+
+To configure the plugin, make a ``loadext`` section in your configuration
+file. The section must consist of a list of paths to extensions to load.
```
sampsyo: I recommend putting the example right here, instead of later down in the last section. You can introduce it by saying something like "The section should look like this:" and list the YAML right there. This makes the introduction mostly self-contained.

docs/plugins/loadext.rst
```diff
@@ -0,0 +1,53 @@
...
+
+To compile the ICU extension, you will need a few dependencies:
+
+ - gcc
+ - icu-devtools
+ - libicu
+ - libicu-dev
+ - libsqlite3-dev
+
+The following commands can be then be used to compile the ICU extension:
```
sampsyo: I recommend a slightly different wording:

> Here's roughly how to download, build, and install the extension (although the details might depend on your system):

docs/plugins/loadext.rst
```diff
@@ -0,0 +1,53 @@
...
+
+.. code-block:: shell
+
+    $ wget https://sqlite.org/2019/sqlite-src-3280000.zip
+    $ unzip sqlite-src-3280000.zip
+    $ cd sqlite-src-3280000/ext/icu
+    $ gcc -shared -fPIC icu.c `icu-config --ldflags` -o libicu.so
+    $ cp libicu.so [your beets config directory]
```
sampsyo: To make this a little more digestible, I'd probably just say:

```
$ cp libicu.so ~/.config/beets
```

Readers (advanced ones, anyway) will be able to sort out that they need to think about where to copy the file.

docs/plugins/loadext.rst
```diff
@@ -0,0 +1,53 @@
...
+-------------
+
+To configure the plugin, make a ``loadext`` section in your configuration
+file. The section must consist of a list of paths to extensions to load.
+
+If a relative path is specified, it is resolved relative to the beets
+configuration directory.
+
+No file extension is required, as SQLite will automatically use the correct
+extension for the current platform.
```
sampsyo: The correct _dynamic library_ extension for the current platform?

jackwilsdon:
Thanks for the feedback on the documentation! I think that's all corrected now :+1:

sampsyo:
Awesome work!! I'll hit the green button now. :smiley: :ship:

jackwilsdon:
Much appreciated! Thanks for all the help regarding the review :+1:

Moonbase59:
Note: On Ubuntu 14.04, using libsqlite 3.8.2-1ubuntu2.2, `SQLITE_DETERMINISTIC` isn’t defined so you’ll get a compilation error. You also need to `sudo apt-get install libicu-dev` before.

I did a quick one and added
```c
#ifndef SQLITE_DETERMINISTIC
#define SQLITE_DETERMINISTIC    0x800
#endif
```
near the top of `icu.c` (starting at line 50) so I could compile using
```bash
gcc -shared -fPIC icu.c `icu-config --ldflags` -o libicu.so
```

Afterwards, using loadext, case-insensitive search worked just fine, even on this old system.

talski:
I'm using Pop!_OS and icu-config seems to be deprecated and could not install it, so I have to run like this
```
gcc -shared -fPIC icu.c `pkg-config --libs icu-uc icu-io icu-i18n` -o libicu.so
```

PR #3227, opened 2019-04-20
docs: remove reference to BeetsPlugin.listen

### LINKED ISSUE #2885: Docs: description of listeners is outdated
### Problem
According to https://beets.readthedocs.io/en/v1.4.6/dev/plugins.html, `BeetsPlugin` should have a `listen` decorator. This is not the case on my system.

```py
In [1]: from beets.plugins import BeetsPlugin

In [2]: BeetsPlugin.listen
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-714d0503bfef> in <module>()
----> 1 BeetsPlugin.listen

AttributeError: type object 'BeetsPlugin' has no attribute 'listen'
```
Using `register_listener` works fine. Someone just removed the decorator without updating the docs.

### Setup

* OS: Arch Linux
* Python version: 3.6.4
* beets version: 1.4.6

arcresu:
This decorator was removed in 4578c4f0e1659daab8d25c79d1f466ae9838d8c7 and now `BeetsPlugin.register_listener` should be used instead. Fixes #2885.

### DIFF
--- a/docs/dev/plugins.rst
+++ b/docs/dev/plugins.rst
@@ -103,19 +103,18 @@ operation. For instance, a plugin could write a log message every time an album
 is successfully autotagged or update MPD's index whenever the database is
 changed.
 
-You can "listen" for events using the ``BeetsPlugin.listen`` decorator. Here's
+You can "listen" for events using ``BeetsPlugin.register_listener``. Here's
 an example::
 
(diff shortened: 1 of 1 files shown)


sampsyo:
Good catch! Thanks; I can't believe this stuck around for so long.

PR #3230, opened 2019-04-21
Allow FetchArt to work with single files

kontheocharis:
In its current state, the FetchArt plugin doesn't work with tracks that aren't in their album folders. This pull request enables an image to be recognised if it's in the same directory as the track, while not necessarily being in its own album folder.

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -683,10 +683,9 @@ class FileSystem(LocalArtSource):
         cover_names_str = b'|'.join(cover_names)
         cover_pat = br''.join([br"(\b|_)(", cover_names_str, br")(\b|_)"])
 
-        for _path in paths:
-            path = _path
+        for path in paths:
             if not os.path.isdir(syspath(path)):
-                path = os.path.dirname(_path)
+                path = os.path.dirname(path)
 
             # Find all files that look like images in the directory.
             images = []
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -683,9 +683,10 @@ class FileSystem(LocalArtSource):
         cover_names_str = b'|'.join(cover_names)
         cover_pat = br''.join([br"(\b|_)(", cover_names_str, br")(\b|_)"])
 
-        for path in paths:
+        for _path in paths:
+            path = _path
             if not os.path.isdir(syspath(path)):
-                continue
+                path = os.path.dirname(_path)
 
             # Find all files that look like images in the directory.
             images = []


beetsplug/fetchart.py
```diff
@@ -683,9 +683,10 @@ def get(self, album, plugin, paths):
         cover_names_str = b'|'.join(cover_names)
         cover_pat = br''.join([br"(\b|_)(", cover_names_str, br")(\b|_)"])
 
-        for path in paths:
+        for _path in paths:
+            path = _path
```
jackwilsdon: Is there a reason to do this? The following is simpler;

```Python
for path in paths:
    if not os.path.isdir(syspath(path)):
        path = os.path.dirname(path)
```

sampsyo:
Got it; looks cool! Could you please add a quick changelog entry? An example, notional directory path might help clarify what has changed (and what's possible now that wasn't before).

wisp3rwind:
> Got it; looks cool! Could you please add a quick changelog entry? An example, notional directory path might help clarify what has changed (and what's possible now that wasn't before).

I agree that an example use-case would be nice to have; the way I understand the effect of this change, its use is not fully clear to me; some thoughts on fetchart's behaviour:
- Note that the change only takes effect in import sessions (when called through `art_for_album(task.album, task.paths, local)`) and not when running `beet fetchart` (calling `art_for_album(album, local_paths)` with `local_paths = [album.path]`)
- I would assume that this change is mainly relevant in a `group_albums` scenario, which would also mean that there's a high risk of associating wrong covers to an album since `fetchart` will not prioritize images with some kind of ID/album name in their names. (That might not really be a change from the current behaviour, I don't really know what exactly `ImportTask.paths` will typically contain.)

sampsyo:
FWIW, `ImportTask.paths` typically contains the set of *containing* directories where a given album was found. This is *usually*, but not always, a single directory—the one containing the album being imported. But the autotagger can also coalesce multiple neighboring directories into a single album, in which case this list will have more than one element. The lists are generated by `ImportTaskFactory.paths`. For album import tasks, the paths here are always directories.

In singleton import tasks, the `paths` list is always one element long: it's the path to the single music file itself.

beetsplug/fetchart.py
kontheocharis: You are right indeed.

jtpavlock:
@kontheocharis are you still interested in working on this?

kontheocharis:
Hi, apologies for the complete inactivity. Unfortunately I am no longer using beets so I am not personally interested in pursuing this further.

jtpavlock:
No problem, thanks for checking back in with us! In that case, I'm going to close this for now, but if you ever feel the need to re-visit this, feel free to re-open or make a new pull request.

Issue #3216, opened 2019-04-15
m4a label tag should be in all caps 


slogsdon7:
The other day I noticed that a lot of my files were missing label info in the beets database, despite them being present on the files. After a little looking, I realized that the beets tag definition was cased differently than how its listed on MusicBrainz tag mapping page.

The code in question: 
https://github.com/beetbox/beets/blob/7b910c3fde50d502e63b890f1ffbaf3d59515c0e/beets/mediafile.py#L1736

On https://picard.musicbrainz.org/docs/mappings/ it is listed as:

`----:com.apple.iTunes:LABEL`

I ended up just replacing it on my machine for the quick fix(which worked, but probably not ideal for the official fix).

sampsyo:
Wow; you’re right! We should fix this, ideally in a way that preserves the ability to read the lower-case version.

slogsdon7:
> Wow; you’re right! We should fix this, ideally in a way that preserves the ability to read the lower-case version.

[This ](https://github.com/slogsdon7/beets/commit/fa0c690d3621b0741dfe0be4bd480a55a8e5a67b) is what I came up with last night to allow a StorageStyle to be set to read only. It works, but the write command doesn't work without the force option, and I'm not really sure the best way to handle that.

sampsyo:
Nice; looks good! Although I think it would also work to do a simple `if self.read_only:` check in the `set` method. Could you please open a PR?

(Also, I'm not sure what you mean about the write command—is this just preventing you from testing the change?)

slogsdon7:
> Nice; looks good! Although I think it would also work to do a simple `if self.read_only:` check in the `set` method. Could you please open a PR?
> 
> (Also, I'm not sure what you mean about the write command—is this just preventing you from testing the change?)

The write command issue ended up being more of an issue with the order of the StorageStyle tags in the MediaField constructor. The get method for fields returns on the first tag from the styles list it extracts from the file, which was the position my read-only style was in, so the corrected tag wasn't being written to because there were no apparent changes to the field that needed writing. Moving the read only fields so the other ones are read first seems to be working for the moment. 

I do think I'll end up refactoring the code so that the check is done in the set method instead since that feels more like the correct way of doing it. Thanks!

sampsyo:
Got it; that makes sense!

PR #3231, opened 2019-04-22
added read_only attribute to base StorageStyle class

slogsdon7:
Worked on as part of addressing #3216 in a backwards-compatible way. 

The way I did this was to modify the base StorageClass to take a named argument "read_only." If the argument evaluates too true, then the `store()` method on the instance is set to a lambda which simply returns None. 

Example usage: `MP4StorageStyle('----:com.apple.iTunes:Label', read_only=True)`.

For the moment, read only styles should be placed after the other styles of the same type in the argument list, as MediaField's get method returns on the first tag style it reads from the file.


arcresu:
This change looks good to me! The only issue is that one of your lines is too long and if failing the style check. Would you mind fixing that up? It might also be nice to add a test for this, but I won't insist :)

slogsdon7:
> This change looks good to me! The only issue is that one of your lines is too long and if failing the style check. Would you mind fixing that up? It might also be nice to add a test for this, but I won't insist :)

Done and done :). Gotta head out before these CI checks finish but if they catch anything else I'll fix them later today

arcresu:
The MediaFile split has happened now, so I'll close this in favour of beetbox/mediafile#9. Thanks for handling the migration!

PR #3236, opened 2019-04-23
Update artresizer's ImageMagick commands to use the magick binary

### LINKED ISSUE #2093: Use explicit paths to ImageMagick tools on Windows
### ImageMagick/artresizer fails on windows?

```
PS C:\Users\Kai\Desktop> beet -vv import '.\Space Probe Taurus - 15.rar'
…
fetchart: trying source coverart for album Space Probe Taurus - Mondo Satan
ImageMagick check `convert --version` failed: Command 'convert --version' returned non-zero exit status 4
artresizer: method is (3, 0)
…
```

```
PS C:\Users\Kai> convert --version
Version: ImageMagick 7.0.2-1 Q16 x64 2016-06-23 http://www.imagemagick.org
Copyright: Copyright (C) 1999-2015 ImageMagick Studio LLC
License: http://www.imagemagick.org/script/license.php
Visual C++: 180040629
Features: Cipher DPC Modules OpenMP
Delegates (built-in): bzlib cairo flif freetype jng jp2 jpeg lcms lqr openexr pangocairo png ps rsvg tiff webp xml zlib
```
### Setup
- OS: Windows 10
- Python version: `Python 2.7.11 (v2.7.11:6d1b6a68f775, Dec  5 2015, 20:32:19) [MSC v.1500 32 bit (Intel)] on win32`
- beets version: 1.3.19
- ~~Turning off plugins made problem go away (yes/no):~~

``` yaml
PS C:\Users\Kai\Desktop> beet config
fetchart: The `fetch_art.remote_priority` configuration option has been deprecated, see the documentation.
lyrics:
    bing_lang_from: []
    auto: yes
    force: no
    bing_client_secret: REDACTED
    google_API_key: REDACTED
    google_engine_ID: REDACTED
    sources:
    - google
    - lyricwiki
    - lyrics.com
    - musixmatch
    fallback:
    genius_api_key: REDACTED
    bing_lang_to:

paths:
    default: $albumartist/$album%aunique{}/%if{$multidisc,CD$disc/}$albumartist - $album%aunique{} - $track $title
item_fields:
    multidisc: 1 if disctotal > 1 else 0
per_disc_numbering: yes
fetchart:
    auto: yes
    minwidth: 750
    maxwidth: 2000
    enforce_ratio: yes
    remote_priority: yes
    sources: coverart itunes amazon albumart filesystem
    cautious: no
    store_source: no
    google_key: REDACTED
    fanarttv_key: REDACTED
    cover_names:
    - cover
    - front
    - art
    - album
    - folder
format_album: 'Musik: $albumartist - $album ($genre - $year) @ $bitrate kBit/s'
library: E:\beets\musiclibrary.db

replace:
    '[\\/]': _
    ^\.: ''
    '[\x00-\x1f]': _
    '[<>:"\?\*\|]': ''
    \.$: _
    \s+$: ''
    ^\s+: ''
ftintitle:
    auto: yes
    drop: no
    format: feat. {0}
zero:
    fields: comments
    update_database: no
    keep_fields: []
embedart:
    auto: yes
    remove_art_file: yes
    compare_threshold: 0
    ifempty: no
    maxwidth: 0

plugins: fetchart lastgenre embedart zero inline lyrics ftintitle discogs fromfilename edit
directory: F:\Audio\Mediathek\Musik

import:
    copy: yes
    write: yes
    resume: ask
    quiet_fallback: skip
    timid: no
    log: beetslog.txt
album_fields:
    bitrate: "total = 0\nfor item in items:\n    total += item.bitrate\nbitrate = total / len(items)\nreturn str(bitrate
)[:3]\n"
discogs:
    tokenfile: discogs_token.json
    apikey: REDACTED
    apisecret: REDACTED
    source_weight: 0.5
edit:
    albumfields: album albumartist
    ignore_fields: id path
lastgenre:
    count: 1
    source: album
    force: yes
    min_weight: 10
    auto: yes
    whitelist: yes
    separator: ', '
    fallback:
    canonical: no
pathfields: {}
```

ababyduck:
Updated artresizer's ImageMagick commands to use the magick binary added in ImageMagick 7.x, rather than the legacy utilities ('convert', 'identify', etc.) This resolves an issue where beets is failing to detect or use ImageMagick on Windows even when it is set correctly on the PATH, which in turn restores functionality to the fetchart and embedart plugins on Windows.

Closes #2093

### DIFF
--- a/beets/util/artresizer.py
+++ b/beets/util/artresizer.py
@@ -93,13 +93,12 @@ def im_resize(maxwidth, path_in, path_out=None):
     # than the given width while maintaining the aspect ratio
     # with regards to the height.
     try:
-        cmds = (['magick'], ['convert'])
-        cmd = cmds[0] if not ArtResizer.shared.im_legacy else cmds[1]
-        args = [util.syspath(path_in, prefix=False),
+        cmd = ArtResizer.shared.im_convert_cmd + \
+            [util.syspath(path_in, prefix=False),
                 '-resize', '{0}x>'.format(maxwidth),
                 util.syspath(path_out, prefix=False)]
 
-        util.command_output(cmd + args)
+        util.command_output(cmd)
     except subprocess.CalledProcessError:
         log.warning(u'artresizer: IM convert failed for {0}',
                     util.displayable_path(path_in))
@@ -124,12 +123,11 @@ def pil_getsize(path_in):
 
 
 def im_getsize(path_in):
-    cmds = (['magick', 'identify'], ['identify'])
-    cmd = cmds[0] if not ArtResizer.shared.im_legacy else cmds[1]
-    args = ['-format', '%w %h', util.syspath(path_in, prefix=False)]
-
     try:
-        out = util.command_output(cmd + args)
+        cmd = ArtResizer.shared.im_identify_cmd + \
+            ['-format', '%w %h', util.syspath(path_in, prefix=False)]
+
+        out = util.command_output(cmd)
     except subprocess.CalledProcessError as exc:
         log.warning(u'ImageMagick size query failed')
         log.debug(
@@ -180,6 +178,12 @@ class ArtResizer(six.with_metaclass(Shareable, object)):
 
         if self.method[0] == IMAGEMAGICK:
             self.im_legacy = self.method[2]
+            if self.im_legacy:
+                self.im_convert_cmd = ['convert']
+                self.im_identify_cmd = ['identify']
+            else:
+                self.im_convert_cmd = ['magick']
+                self.im_identify_cmd = ['magick', 'identify']
 
     def resize(self, maxwidth, path_in, path_out=None):
         """Manipulate an image file according to the method, returning a
@@ -246,30 +250,31 @@ def get_im_version():
     Try invoking ImageMagick's "magick". If "magick" is unavailable,
     as with older versions, fall back to "convert"
 
-    Our iterator `im_legacy` will be non-zero when the first command
-    fails, and will be returned in a tuple along with the version
+    Our iterator will be non-zero when the first command fails, and will
+    be returned in a tuple along with the version.
     """
-    cmds = ('magick', 'convert')
-    for im_legacy, cmd in enumerate(cmds):
+    cmd_names = (['magick'],
+                 ['convert'])
+    for i, cmd_name in enumerate(cmd_names):
 
         try:
-            out = util.command_output([cmd, '--version'])
+            cmd = cmd_name + ['--version']
+            out = util.command_output(cmd)
 
             if b'imagemagick' in out.lower():
                 pattern = br".+ (\d+)\.(\d+)\.(\d+).*"
                 match = re.search(pattern, out)
+                version = (int(match.group(1)),
+                           int(match.group(2)),
+                           int(match.group(3)))
+                legacy = bool(i)
                 if match:
-                    return ((int(match.group(1)),
-                            int(match.group(2)),
-                            int(match.group(3))),
-                            bool(im_legacy)
-                            )
+                    return (version, legacy)
 
         except (subprocess.CalledProcessError, OSError) as exc:
             log.debug(u'ImageMagick version check failed: {}', exc)
-            return None
 
-        return (0,)
+        return (0, None)
 
 
 def get_pil_version():
--- a/beets/util/artresizer.py
+++ b/beets/util/artresizer.py
@@ -93,7 +93,7 @@ def im_resize(maxwidth, path_in, path_out=None):
     # with regards to the height.
     try:
         util.command_output([
-            'convert', util.syspath(path_in, prefix=False),
+            'magick', util.syspath(path_in, prefix=False),
             '-resize', '{0}x>'.format(maxwidth),
             util.syspath(path_out, prefix=False),
         ])
@@ -121,7 +121,7 @@ def pil_getsize(path_in):
 
 
 def im_getsize(path_in):
-    cmd = ['identify', '-format', '%w %h',
+    cmd = ['magick', 'identify', '-format', '%w %h',
            util.syspath(path_in, prefix=False)]
     try:
         out = util.command_output(cmd)
@@ -235,7 +235,7 @@ def get_im_version():
     Try invoking ImageMagick's "convert".
     """
     try:
-        out = util.command_output(['convert', '--version'])
+        out = util.command_output(['magick', '-version'])
 
         if b'imagemagick' in out.lower():
             pattern = br".+ (\d+)\.(\d+)\.(\d+).*"
--- a/beets/util/artresizer.py
+++ b/beets/util/artresizer.py
@@ -232,9 +232,9 @@ class ArtResizer(six.with_metaclass(Shareable, object)):
     def _check_method():
         """Return a tuple indicating an available method and its version."""
         try:
-            version, im_legacy = get_im_version()
-            if version:
-                return IMAGEMAGICK, version, im_legacy
+            version, legacy = get_im_version()
+            if version > (0, 0, 0):
+                return IMAGEMAGICK, version, legacy
         except TypeError:
             pass
 
@@ -274,7 +274,7 @@ def get_im_version():
         except (subprocess.CalledProcessError, OSError) as exc:
             log.debug(u'ImageMagick version check failed: {}', exc)
 
-        return (0, None)
+        return ((0,), None)
 
 
 def get_pil_version():
--- a/beets/util/artresizer.py
+++ b/beets/util/artresizer.py
@@ -94,7 +94,7 @@ def im_resize(maxwidth, path_in, path_out=None):
     # with regards to the height.
     try:
         cmds = (['magick'], ['convert'])
-        cmd = cmds[0] if not im_legacy else cmds[1]
+        cmd = cmds[0] if not ArtResizer.shared.im_legacy else cmds[1]
         args = [util.syspath(path_in, prefix=False),
                 '-resize', '{0}x>'.format(maxwidth),
                 util.syspath(path_out, prefix=False)]
@@ -125,7 +125,7 @@ def pil_getsize(path_in):
 
 def im_getsize(path_in):
     cmds = (['magick', 'identify'], ['identify'])
-    cmd = cmds[0] if not im_legacy else cmds[1]
+    cmd = cmds[0] if not ArtResizer.shared.im_legacy else cmds[1]
     args = ['-format', '%w %h', util.syspath(path_in, prefix=False)]
 
     try:
@@ -178,6 +178,9 @@ class ArtResizer(six.with_metaclass(Shareable, object)):
         log.debug(u"artresizer: method is {0}", self.method)
         self.can_compare = self._can_compare()
 
+        if self.method[0] == IMAGEMAGICK:
+            self.im_legacy = self.method[2]
+
     def resize(self, maxwidth, path_in, path_out=None):
         """Manipulate an image file according to the method, returning a
         new path. For PIL or IMAGEMAGIC methods, resizes the image to a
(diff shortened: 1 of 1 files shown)


sampsyo:
Thanks! This seems like a good thing to do, but as a maintainer I'm contractually obligated to worry about backwards compatibility. How much do we need to worry about systems with only the older versions of IM, without the `magick` binary, installed? (How many CentOS or Debian versions back would you have to go for that to be the case?) If that's a real possibility, can we set this up to try `magick` and, if that doesn't exist, fall back to `convert`?

We do something similar in audioread to switch between `avconv` and `ffmpeg`, for example:
https://github.com/beetbox/audioread/blob/c8bedf7880f13a7b7488b108aaf245d648674818/audioread/ffdec.py#L82-L98

ababyduck:
You're right, it looks like the CentOS repos are still serving version 6.7 of ImageMagick. How would you feel about using that `popen_multiple()` function, maybe in `beets.util`, and just calling that from `artresizer`?

ababyduck:
How do you feel about this solution? 

Tested and working with both `fetchart` and `embedart` on:
- Windows 10, ImageMagick 7.0.8-41 with and without `magick`
- Ubuntu 18.04.1 LTS, ImageMagick 6.9.7-4 without `magick`, with and without `convert`

beets/util/artresizer.py
```diff
@@ -92,11 +93,13 @@ def im_resize(maxwidth, path_in, path_out=None):
     # than the given width while maintaining the aspect ratio
     # with regards to the height.
     try:
-        util.command_output([
-            'convert', util.syspath(path_in, prefix=False),
-            '-resize', '{0}x>'.format(maxwidth),
-            util.syspath(path_out, prefix=False),
-        ])
+        cmds = (['magick'], ['convert'])
```
sampsyo: Consider hoisting this out to a module-level constant, named something like `CONVERT_CMDS`.

beets/util/artresizer.py
```diff
@@ -92,11 +93,13 @@ def im_resize(maxwidth, path_in, path_out=None):
...
+        cmd = cmds[0] if not ArtResizer.shared.im_legacy else cmds[1]
```
sampsyo: Or, even better than the above: consider making the `ArtResizer` set a `convert_cmd` and `identify_cmd` field on itself, which chooses between these two options. Then the code that uses the command doesn't need to duplicate the logic to check `ArtResizer.shared.im_legacy` and switch between the two options.

beets/util/artresizer.py
```diff
@@ -121,10 +124,12 @@ def pil_getsize(path_in):
 
 
 def im_getsize(path_in):
-    cmd = ['identify', '-format', '%w %h',
-           util.syspath(path_in, prefix=False)]
+    cmds = (['magick', 'identify'], ['identify'])
+    cmd = cmds[0] if not ArtResizer.shared.im_legacy else cmds[1]
+    args = ['-format', '%w %h', util.syspath(path_in, prefix=False)]
```
sampsyo: I'd probably stick with the name `cmd` for the thing that includes both the executable name and the arguments:

```py
cmd = (the command) + ['-format', '%w %h', ...]
```

beets/util/artresizer.py
```diff
@@ -232,23 +243,33 @@ def _check_method():
...
+                            int(match.group(2)),
+                            int(match.group(3))),
+                            bool(im_legacy)
+                            )
+
+        except (subprocess.CalledProcessError, OSError) as exc:
+            log.debug(u'ImageMagick version check failed: {}', exc)
+            return None
+
+        return (0,)
```
sampsyo: The return type of this function could be a little cleaner and more regular: perhaps it should _always_ return a pair of a "legacy" flag (a Boolean) and a version. The return value here doesn't include the flag.

beets/util/artresizer.py
```diff
@@ -232,23 +243,33 @@ def _check_method():
...
+
+            if b'imagemagick' in out.lower():
+                pattern = br".+ (\d+)\.(\d+)\.(\d+).*"
+                match = re.search(pattern, out)
+                if match:
+                    return ((int(match.group(1)),
+                            int(match.group(2)),
+                            int(match.group(3))),
+                            bool(im_legacy)
+                            )
```
sampsyo: For legibility, something like this might be better:

```py
vers = (int(...), ...)
return vers, legacy
```

beets/util/artresizer.py
ababyduck: Went with your second suggestion here

beets/util/artresizer.py
ababyduck: Added in 48be3a7

beets/util/artresizer.py
ababyduck: Added in 48be3a7

beets/util/artresizer.py
ababyduck: Added in 48be3a7

beets/util/artresizer.py
ababyduck: Added in 48be3a7

ababyduck:
Thanks for the review! I've implemented the requested changes in 48be3a7.

beets/util/artresizer.py
ababyduck: I forgot a set of parentheses here and returned `(0, None)` rather than `((0,), None)`. After thinking about it some more though, I wonder if it should be reverted rather than fixed.

We know that if the version returns 0, all usable commands have failed and we can safely assume that ImageMagick is not installed or accessible. In this case, the legacy flag doesn't make sense since we a) have no installation to refer to as legacy or not, and b) can't make use of ImageMagick anyway. 

Since all existing code that checks the version refers to it using `self.method[1]`, we'd then have to change each of these to `self.method[1][0]` or else explicitly store the version in another variable. Doesn't seem worth it just to return a flag that will never be used.

I'm going to revert this specific change for now, but let me know your thoughts.

beets/util/artresizer.py
```diff
@@ -219,9 +231,12 @@ def _can_compare(self):
     @staticmethod
     def _check_method():
         """Return a tuple indicating an available method and its version."""
```
sampsyo: This docstring could use a little expansion to describe what the method now returns.

beets/util/artresizer.py
```diff
@@ -173,6 +176,15 @@ def __init__(self):
         log.debug(u"artresizer: method is {0}", self.method)
         self.can_compare = self._can_compare()
 
+        if self.method[0] == IMAGEMAGICK:
```
sampsyo: A comment here might be nice: something like,

> Use ImageMagick's `magick` binary when it's available. If it's not, fall back to the older, separate `convert` and `identify` commands.

beets/util/artresizer.py
```diff
@@ -232,23 +247,34 @@ def _check_method():
...
-
-    except (subprocess.CalledProcessError, OSError) as exc:
-        log.debug(u'ImageMagick check `convert --version` failed: {}', exc)
-        return None
+    cmd_names = (['magick'],
+                 ['convert'])
+    for i, cmd_name in enumerate(cmd_names):
+
+        try:
+            cmd = cmd_name + ['--version']
```
sampsyo: As above, the `cmd` assignment can go outside of the `try`.

beets/util/artresizer.py
```diff
@@ -232,23 +247,34 @@ def _check_method():
...
-                        int(match.group(2)),
-                        int(match.group(3)))
-            return (0,)
-
-    except (subprocess.CalledProcessError, OSError) as exc:
-        log.debug(u'ImageMagick check `convert --version` failed: {}', exc)
-        return None
+    cmd_names = (['magick'],
+                 ['convert'])
+    for i, cmd_name in enumerate(cmd_names):
```
sampsyo: I recommend something a little more obvious than using `enumerate`: something like this.

```py
for legacy, cmd_name in ((False, 'magick'), (True, 'convert')):
```

This avoids the need for a `bool(i)` conversion later on.

beets/util/artresizer.py
```diff
@@ -232,23 +247,34 @@ def _check_method():
...
+
+            if b'imagemagick' in out.lower():
+                pattern = br".+ (\d+)\.(\d+)\.(\d+).*"
+                match = re.search(pattern, out)
+                version = (int(match.group(1)),
+                           int(match.group(2)),
+                           int(match.group(3)))
+                legacy = bool(i)
+                if match:
+                    return (version, legacy)
```
sampsyo: No parentheses are necessary here:

```suggestion
                    return version, legacy
```

sampsyo: Also, in the old version of the code, we still return a version (which just happens to be zero). So we probably want to preserve that behavior:

```py
else:
    return (0,), legacy
```

beets/util/artresizer.py
```diff
@@ -232,23 +247,34 @@ def _check_method():
...
+        try:
+            cmd = cmd_name + ['--version']
+            out = util.command_output(cmd)
+
+            if b'imagemagick' in out.lower():
+                pattern = br".+ (\d+)\.(\d+)\.(\d+).*"
+                match = re.search(pattern, out)
+                version = (int(match.group(1)),
+                           int(match.group(2)),
+                           int(match.group(3)))
```
sampsyo: This will crash if `match` is None, i.e., the version isn't found. The `version` assignment should go under the `if` below.

beets/util/artresizer.py
```diff
@@ -232,23 +247,34 @@ def _check_method():
...
+                legacy = bool(i)
+                if match:
+                    return (version, legacy)
+
+        except (subprocess.CalledProcessError, OSError) as exc:
+            log.debug(u'ImageMagick version check failed: {}', exc)
+
+        return (0, None)
```
sampsyo: I think I finally see what's going on here: we want to abort and say that we couldn't find ImageMagick at all. If that's the case, I think this probably *should* just be `return None`, and the docstring should say "return a version/legacy-flag pair or None if the check fails".

beets/util/artresizer.py
```diff
@@ -219,9 +231,12 @@ def _can_compare(self):
     @staticmethod
     def _check_method():
         """Return a tuple indicating an available method and its version."""
-        version = get_im_version()
-        if version:
-            return IMAGEMAGICK, version
+        try:
+            version, im_legacy = get_im_version()
+            if version:
+                return IMAGEMAGICK, version, im_legacy
```
sampsyo: Following up on a comment below, the return value of `get_im_version` should probably be `None` (i.e., not a pair) when the command can't be found at all. This is a little annoying, but it means this code would look like:

```py
v = get_im_version()
if v:
    version, im_legacy = v
    ...
```

beets/util/artresizer.py
ababyduck: The bool conversion is actually unnecessary, since we really only need to evaluate `i` as non-zero, but I thought the intended purpose would be more clear as a bool. I added an explanation to the docstring on line 257 which I thought was reasonably clear, but your approach is perhaps more concise anyway.

beets/util/artresizer.py
ababyduck: The current try/except block already handles a None return, since trying to unpack None to two variables results in a TypeError, which we can `except` and `pass`. However, your approach is once again slightly shorter and more readable, so we'll go with that!

beets/util/artresizer.py
ababyduck: The bool conversion is actually unnecessary, since we really only need to evaluate `i` as non-zero, but I thought the intended purpose would be more clear as a bool. I added an explanation to the docstring on line 257 which I thought was reasonably clear, but your approach is perhaps more concise anyway.

ababyduck:
Looks like Appveyor's failing tests because it's having issues setting up unrar in the test environment. 👌

Latest round of suggestions has been implemented.

PR #608, opened 2014-03-13
Write tags only when changed

geigerzaehler:
This was previously discussed in #527 and some questions came up.

>  Unintentionally capturing modifications to other fields, such as when mgfile is assigned

It’s better to be on the safe side and perform unnecessary writes instead of missing necessary ones. Also
`mgfile` should never ever be changed.

> Missing changes when doing other kinds of modifications, like the delete operation used by the scrub plugin

`delete` is handled [here](https://github.com/geigerzaehler/beets/commit/b16eb61eb7232727646944ac2514db9976ec328f#diff-9f32ae8136422d861cc3215f657f1e1cR1219) and tested [here](https://github.com/geigerzaehler/beets/commit/b16eb61eb7232727646944ac2514db9976ec328f#diff-19d66b8d5e80f862f273212a6a1698e2R199)

> The potential inefficiency of reading each field twice on each write, especially for images

Files are not read twice since all tags are cached when the `mgfile` is created.

PR #3237, opened 2019-04-23
mediafile: replace with a re-export of beetbox/mediafile

arcresu:
This re-exports all of the [MediaFile](https://github.com/beetbox/mediafile) package under `beets.mediafile`. A warning is issued when importing from the deprecated namespace.

Remaining work:
  - [x] Port the outstanding commits from `beets.mediafile` to MediaFile (beetbox/mediafile#6).
  - [x] Make a new release of MediaFile.
  - [x] Update the minimum dependency of beets on MediaFile to that new release.
  - [x] Remove all uses of the deprecated namespace from beets itself (this should be left until later so as to avoid picking up many conflicts in this branch while the previous steps happen).
  - [ ] Delete MediaFile wiki page.
  - [x] Update docs (e.g. `dev/plugins`) and wiki.
  - [x] Add changelog entry.
  - [ ] Notify downstream packagers about the new dependency (although I don't think that we need to block a beets release on this as discussed in https://github.com/beetbox/beets/pull/2061#issuecomment-480565229).

Part of #1966.


arcresu:
The following open PRs touch `beets/mediafile.py` and will need to be (at least partially) migrated to the standalone mediafile. Merge conflicts will appear once this PR is merged.

- [ ] #608
- [ ] #1957
- [ ] #2503
- [ ] #2580
- [ ] #2619
- [ ] #2640
- [ ] #3231

arcresu:
I've started adding a new "mediafile" label to issues and PRs that should be completely moved to the standalone MediaFile. I'm not adding it to issues/PRs that affect MediaFile as part of a larger set of changes since these should be split rather than moved.

arcresu:
I've gone ahead and finished the MediaFile split so that it's ready to go. I think we should aim to get it merged as early as possible in the current release cycle.

For the list of affected open PRs above, most are outdated and are already conflicting anyway. I left comments on those to alert the original submitters in case they're able to let us know their plans. There are two active PRs that are affected. The author of one is currently splitting and resubmitting the PR, and the other is ~done and I'm happy to manually migrate those changes myself.

One other thing I noticed is that now Mutagen becomes a transient dependency. It's only used directly by the scrub plugin now. Are we ok with dropping the direct dependency on Mutagen in beets?

sampsyo:
Yay! I'm very excited about this, and—thanks to your effort ahead of time—the footprint of this change is now fairly small. It will be truly great to have this cleanly factored out.

Yep, I think we can indeed drop the direct Mutagen dependency; it's interesting that `scrub` is the only thing that really needs to skip the MediaFile abstraction level to do its work.

arcresu:
Of the three remaining PRs, one has removed the MediaFile changes and the other two are from branches that have been conflicting for a long time anyway. Shall we hit the button now?

sampsyo:
Yeah, sounds great! In the spirit of getting releases out the door more quickly, what do you think about pushing out the current release and its handful of fixes and then immediately merging this as the first thing in the next release? We can start the publicity campaign to notify packagers about the new dependency.

arcresu:
Alright, so this small release would mostly be for the imagemagick issue? That sounds like a good plan then. Some planning thoughts:

* The confit/confuse unbundling is also almost ready, mostly waiting on sampsyo/confuse#50 to go in followed by a release of confuse. Should we aim to do that in the same release (1.4.9 + 1) or do one at a time?
* The next bit of technical debt I want to tackle is reviving the effort to improve the plugin architecture. I looked through the stalled PR and can see some ways to break it up and make changes that maintain backward compatibility unlike the original approach. Once mediafile and confuse are done that'll be my next focus.

sampsyo:
Good point! I think doing the Confuse extraction in the same release would be a good idea. It could save people time to add two packages at once.

And I'm really excited that you're interested in helping refactor the plugin infrastructure! This I something I'd really like to clean up, so I'd be really happy to talk about how to approach this. Woohoo! :rocket:

sampsyo:
Done! :tada: This is the first change of v0.5.0. Thank you again for your work on getting this to happen!!

GuilhermeHideki:
I used grep to find the external plugins using mediafile:
-  `cd` the site-packages folder found with ` python -c "import site; print(site.getsitepackages())"`
- ` grep -rnw beetsplug -e 'mediafile'`

PR #3241, opened 2019-04-27
Use new GitHub issue template format

arcresu:
GitHub has been showing a warning that we're using an old style of the issue templates. The [new style](https://help.github.com/en/articles/about-issue-and-pull-request-templates) allows multiple templates for different issue types.

Taking inspiration from some [examples](https://github.com/stevemao/github-issue-templates/blob/457e1d0ae0f7812c0ac0c0909840f446edc242fd/system/ISSUE_TEMPLATE/) elsewhere, I converted the existing template to be for bug reports and added a new one for feature requests.

Just to add: there's a [template editor](https://help.github.com/en/articles/creating-issue-templates-for-your-repository) in GitHub but it seems you need more privileges than I have to use it. As such I don't know for sure that this will be formatted exactly correctly until it's merged.


.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -0,0 +1,15 @@
+---
+name: "\U0001F680 Feature request"
+about: Suggest a new idea for beets
+
+---
+
+### Use case
+
+I'm trying to use beet to...
```
sampsyo:
```suggestion
I'm trying to use beets to...
```

.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -0,0 +1,15 @@
...
+
+
+### Solution <!-- Describe how beets could be changed. -->
```
sampsyo: To expand on this instruction a little bit, how about this?

> Do you have a proposal for how beet should work? Try to be as specific as possible—for example, you could propose the name for a new command-line option or refer to the particular ID3 frame you wish were supported.

.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -0,0 +1,15 @@
...
+
+
+### Alternatives <!-- E.g. could an existing plugin do the same thing? -->
```
sampsyo: On the theme of more specificity, and phrasing things as questions:

> Have you tried using an existing plugin to do something similar? Is there any current feature that's close but not quite the same?

arcresu:
Thanks for the expanded placeholders. I agree my stubs didn't sound quite right, but I also was worried that people would get confused by the HTML comments and we'd start seeing blank issues with carefully typed but invisible details. I guess that sort of thing can be iterated on later.

sampsyo:
Sure; sounds great. I'll merge these now and we can see how they go.

Issue #1531, opened 2015-07-01
Finish disentangling dbcore from beets globals


sampsyo:
There are three places left in `beets.dbcore` that reference `beets.config` (find them with `ack config beets/dbcore`). The `dbcore` component is supposed to be independent of the rest of beets, so we need to remove these.

PR #3243, opened 2019-04-27
Stop using beets.config['sort_case_insensitive'] in beets.dbcore

jackwilsdon:
This PR removes the use of `beets.config['sort_case_insensitive']` from `beets.dbcore`, instead moving the configuration access into `library.py`.

There is still one more use of `beets.config` within `beets.dbcore`, which I am hoping to address separately.

Related: #1531

### DIFF
--- a/beets/dbcore/queryparse.py
+++ b/beets/dbcore/queryparse.py
@@ -20,7 +20,6 @@ from __future__ import division, absolute_import, print_function
 import re
 import itertools
 from . import query
-import beets
 
 PARSE_QUERY_PART_REGEX = re.compile(
     # Non-capturing optional segment for the keyword.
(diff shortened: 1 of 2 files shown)


sampsyo:
Cool; thanks! This is a solid refactoring. :smiley: I know this borders on bikeshedding, but did you consider inverting the sense for this option as it's exposed from dbcore? Perhaps `case_sensitive=False` would be a little less double-negative-y.

jackwilsdon:
Definitely not bikeshedding, I'm always open to this kind of discussion! I actually started off with it named `case_sensitive` but ended up switching it to match the `sort_case_insensitive` config option.

Looking around, it also seems we primarily use `case_insensitive`, even with it's double negativity.

https://github.com/beetbox/beets/search?q=case_insensitive

https://github.com/beetbox/beets/search?q=case_sensitive

jackwilsdon:
I'll go ahead and merge this :+1: cheers for the feedback!

PR #3245, opened 2019-04-28
Changelog tool

arcresu:
As I mentioned on Gitter, I like [how GitLab handles changelogs](https://docs.gitlab.com/ee/development/changelog.html) so here's my proposal for how we can do this in beets.

The goal is to keep the contribution barrier low while making it easier for maintainers to produce a high quality changelog. The implementation is to create a small YAML file in `docs/unreleased-changes` for each changelog entry that includes some structured information. PRs no longer have to touch `docs/changelog.rst` so other PRs or releases won't cause annoying merge conflicts.

### Helper script
There's a new `./extra/changelog.py` script that contributors can run to walk them through the process, autocompleting from the environment wherever possible. Options can be specified as flags to the helper for regular contributors who know what they're doing, and anything unspecified is filled in with a wizard-style prompt. The minimal invocation is `./extra/changelog.py <PR NUMBER>` and the questions are:

  * **Change type**: helps with where to put the entry in the final changelog. Default is `bugfix`.
  * **Related issues**: issues other than the PR to link to. Optional. We don't really need to know the current PR number separately to the related issues, but I wanted to make the PR number mandatory and enforcing that at the CLI level seemed appropriate.
  * **Title** (the entry itself). Defaults to the subject line of the most recent commit. I say that you can use ReST syntax but don't explain what that is. It's fine if they just put plain text. Hyperlinks are going to be awkward though since you can't enter line breaks.
  * **Author**: the GitHub username. Checks `GITHUB_USER` and git's `github.user` config which are both non-standard but are used by at least [some](https://github.com/jsmits/github-cli) other tools.
  * **Components**: parts of beets changes by the PR. Helps with where to include the changelog entry and forms the first part of the entry itself. We could deduce this if we could check the PR's diff but that needs an up-to-date upstream remote in the local repo and users probably won't have this set up.

I also added a pull request template that reminds contributors to add a changelog if necessary.

### Releases
I also had a go at implementing a new `./extra/release.py unreleased` command that gathers up the changes into a formatted changelog entry that looks similar to historical ones. The command only prints out the formatted changelog and leaves tasks like updating `docs/changelog.rst` and removing `docs/unreleased-changes/*.yaml` as manual steps (although this could of course be changed).

The specifics of how this works should be adjusted after some experience with real changelog entries.

### DIFF
--- a/extra/changelog.py
+++ b/extra/changelog.py
@@ -83,7 +83,7 @@ def guess_username():
     return os.getenv(
             'GITHUB_USER',
             subprocess.Popen(
-                ['git', 'config', '--global', 'github.user'],
+                ['git', 'config', 'github.user'],
                 stdout=subprocess.PIPE).stdout.read())
 
 
--- a/extra/changelog.py
+++ b/extra/changelog.py
@@ -27,6 +27,8 @@ COMPONENTS = CORE_COMPONENTS + PLUGINS
 
 
 class ChangelogEntry(object):
+    """Represents a single entry in the (unreleased) changelog.
+    """
     def __init__(self, title, pull_request, type, username=None,
                  components=None, related=None):
         self.title = title
@@ -37,6 +39,8 @@ class ChangelogEntry(object):
         self.related = related or []
 
     def as_dict(self):
+        """Collect data for serialisation.
+        """
         entry = {
             'title': self.title,
             'pull_request': self.pr,
@@ -49,14 +53,20 @@ class ChangelogEntry(object):
         return entry
 
     def write(self, path):
+        """Serialise the entry as YAML to the given path.
+        """
         with open(path, 'wb') as entry_file:
             yaml.safe_dump(self.as_dict(), entry_file, encoding='utf-8',
                            explicit_start=True)
 
     def print(self):
+        """Print out the serialised version of this entry.
+        """
         print(yaml.safe_dump(self.as_dict(), explicit_start=True))
 
     def affects(self):
+        """Sort the list of components and convert plugins to doc links.
+        """
         core = [x for x in sorted(self.components) if '/' not in x]
         plugins = [':doc:`/plugins/{}`'.format(x.split('/')[1])
                    for x in sorted(self.components)
@@ -64,8 +74,12 @@ class ChangelogEntry(object):
         return core + plugins
 
     def issues(self):
+        """Returns a list of relevant issues in ReST format.
+        """
         return [':bug:`{}`'.format(i) for i in [self.pr] + self.related]
 
+
+# Check the environment for default values we can suggest.
 
 def guess_name():
     log_entry = subprocess.Popen(['git', 'rev-parse', '--abbrev-ref', 'HEAD'],
@@ -88,6 +102,11 @@ def guess_username():
 
 
 def ask_value(name, default=None, convert=str):
+    """Prompt for a value and convert to the necessary type.
+    If a default value is provided, it can be accepted with a single key press.
+    The convert function is responsible for raising (any) exception if the
+    value typed by the user is invalid.
+    """
     prompt = '    {}>  '.format(name)
     if default:
         print("Default (press enter to accept): {}".format(default))
@@ -107,6 +126,8 @@ def ask_value(name, default=None, convert=str):
                 print(e)
                 print("That value wasn't valid!")
 
+
+# Validate entered data and convert it into the relevant types.
 
 def convert_components(components_string):
     components = components_string.split()
@@ -174,6 +195,8 @@ def parse_arguments():
 if __name__ == '__main__':
     args = parse_arguments()
 
+    # The filename for the YAML document is derived from the current git branch
+    # name, and can be overridden with a flag. Asks confirmation to clobber.
     name = convert_name(args.name or guess_name())
     path = os.path.join(CHANGES, name + '.yaml')
     if os.path.exists(path):
--- a/extra/changelog.py
+++ b/extra/changelog.py
@@ -1,0 +1,262 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+"""A tool for generating changelog entries.
+See more instructions at docs/unreleased-changes/README.md
+"""
+import yaml
+import os
+import sys
+import subprocess
+import argparse
+
+BASE = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+CHANGES = os.path.join(BASE, 'docs', 'unreleased-changes')
+
+# Types of changelog entries.
+TYPES = ['bugfix', 'feature', 'change', 'newplugin']
+
+# List of all built-in plugins.
+PLUGINS = ['plugin/' + os.path.splitext(x)[0]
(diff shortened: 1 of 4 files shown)


arcresu:
AppVeyor seems a little flaky recently. Nothing here is tested by the CI anyway.

This whole approach to changelogs might be overkill. I mean the separate files thing made sense for a huge project like GitLab CE, but beets might not be big enough to warrant that. We could alternatively just include a more thorough guide for writing changelog entries and live with the current situation with merge conflicts in `changelog.rst`.

If we go with this there's still scope to slightly simplify the questions (e.g. asking about the PR number and related issues at the same time, simplifying components to just be about plugins affected).

wisp3rwind:
>  Hyperlinks are going to be awkward though since you can't enter line breaks.

Would it be an option to open a temporary file using `$EDITOR` to enter the changelog body if a optional `-e`switch is given (or even do that by default, but not everyone might be happy about that)

> This whole approach to changelogs might be overkill.

> If we go with this there's still scope to slightly simplify the questions (e.g. asking about the PR number and related issues at the same time, simplifying components to just be about plugins affected, or not even asking about components at all).

While merge conflicts are certainly not as big an issue as in the case of Gitlab, I nevertheless believe this is worth experimenting with due to its potential to lower the burden of making a new release (in particular if the question about components is not dropped).

extra/changelog.py
```diff
@@ -0,0 +1,262 @@
...
+    log_entry = subprocess.Popen(['git', 'log', '--format=%s', '-1'],
+                                 stdout=subprocess.PIPE).stdout.read()
+    return log_entry.rstrip().decode('utf-8')
+
+
+def guess_username():
+    return os.getenv(
+            'GITHUB_USER',
+            subprocess.Popen(
+                ['git', 'config', '--global', 'github.user'],
```
wisp3rwind: The Gitlab changelog tool considers `git config user.name`, I think it would be nice to have this as an additional fallback (personally, I usually have that variable set, but not `github.user`). Is there a reason to pass `--global`? Without it, global settings will still be read, but can be overridden per-repository.

sampsyo:
Wow; this is very cool! I also don't _really_ know whether it will turn out to feel like overkill or not, but I would be interested to give it a try. In an abstract sense, I like the idea of separating ongoing, somewhat rough, just-the-facts change information from the more polished, readable version we produce to associate with a given release. The former has an opportunity to avoid merge conflicts, while the latter seems intrinsically conflict-prone.

Anyway, I'd be interested from hearing from other members of the community about whether this seems like an attractive option!

extra/changelog.py
jackwilsdon: I think `user.name` is often set to the developer's name as opposed to their GitHub username (e.g. mine is set to "Jack Wilsdon"), which isn't what is wanted here (the username ends up being appended to `https://github.com/` to create a link to their profile).

arcresu: Yes this was exactly the reason. I only wanted to collect the username rather than real name based on what was done in the past. I guess we could fall back on the real name and use a heuristic like "if it contains a space then don't expand it to a URL". It seemed simpler just to ask the user to type the username and let them know how to pre-fill it in future.

Yes it seems that the more official GitHub tools get the username from the auth token now so they don't tend to need a config item. I think it's more likely that people would set the env variable than use the git config anyway in our case. Agreed on the point about `--global`.

arcresu:
> > Hyperlinks are going to be awkward though since you can't enter line breaks.
> 
> Would it be an option to open a temporary file using `$EDITOR` to enter the changelog body if a optional `-e`switch is given (or even do that by default, but not everyone might be happy about that)

Yeah I considered that, but since the end result is just a small YAML file maybe it's better just to let people edit that by hand after the script fills it in, if they need less common things like line breaks or want to edit the text.

> While merge conflicts are certainly not as big an issue as in the case of Gitlab, I nevertheless believe this is worth experimenting with due to its potential to lower the burden of making a new release (in particular if the question about components is not dropped).

Do you have a feel for what would be a useful set of components to include? My list didn't have a lot of thought put into it. I included `dev` because I remember reading a request (that I can't find at the moment) for migration documentation for plugin developers, so this would at least help to highlight relevant changes. Not sure about `docs`, `data` and `import`.

arcresu:
I'm also unsure about some of the wording (particularly the checklist in the PR template). I don't want it to come across as patronising or to make the process feel too heavy.

extra/changelog.py
wisp3rwind: Alright, sorry about the fuss. To be honest, I hadn't really checked how that value is used.

wisp3rwind:
> Yeah I considered that, but since the end result is just a small YAML file maybe it's better just to let people edit that by hand after the script fills it in, if they need less common things like line breaks or want to edit the text.

Sounds reasonable :)

> Do you have a feel for what would be a useful set of components to include? My list didn't have a lot of thought put into it. I included `dev` because I remember reading a request (that I can't find at the moment) for migration documentation for plugin developers, so this would at least help to highlight relevant changes. Not sure about `docs`, `data` and `import`.

For reference, the 1.4.8 changelog was roughly structured as
```
- new core features
- new plugins
- improvements
  - existing plugins
  - performance
  - usability
  - crashes
  - compat
  - other fixes
- dev
- packager
```
So I guess in addition to `dev`, `packager` should also be a choice. Looking at recent changelogs, in some cases, these two categories were related to items in the improvements section while in other cases they were independent. So it might be useful to ask a _separate_ question about affected parties (`user`, `dev`, `packager`) where multiple values can be given. Whether the tool should then duplicate the entry to the `improvements` and `dev`/`packager` sections or leave that up to manual editing I'm not sure about.

Ultimately what the questions should look like heavily depends on how the changelog is to be structured, and I think previous changelogs vary quite a bit there. Looking at the structure above, the `component` could be replaced by a question about the kind of improvement (if `kind` is `improvement`, else that prompt would be skipped). However I don't have a strong opinion on what the structure should be, and maybe people who have actually ever edited any release notes can give better advice than me.

arcresu:
Thanks for your thoughts!

> For reference, the 1.4.8 changelog was roughly structured as
> 
> ```
> - new core features
> - new plugins
> - improvements
>   - existing plugins
>   - performance
>   - usability
>   - crashes
>   - compat
>   - other fixes
> - dev
> - packager
> ```

I broke up the last changelog that way because it was so long and it it seemed useful to group the changes. I don't think there's normally going to be enough entries to warrant that level of granularity though. We could start with a very simple list of sections and grow it later if we start to see a need.

> So I guess in addition to `dev`, `packager` should also be a choice. Looking at recent changelogs, in some cases, these two categories were related to items in the improvements section while in other cases they were independent. So it might be useful to ask a _separate_ question about affected parties (`user`, `dev`, `packager`) where multiple values can be given. Whether the tool should then duplicate the entry to the `improvements` and `dev`/`packager` sections or leave that up to manual editing I'm not sure about.

Normally there are at most one or two changes relevant to distro packagers, and they're normally made by frequent beets contributors, from what I've seen. You might be right that we could use an additional field for this information in the YAML file, but I would hesitate to add it as a question to the normal workflow.

jtpavlock:
I'm all for this!

PR #3251, opened 2019-05-05
#2579 Adding styles to discogs plugin

### LINKED ISSUE #2579: discogs: Fetch "style" field
Can beets please be enhanced to pull music style in addition to Genre. In my opinion discog have the most intelligent [music categorising](https://www.discogs.com/help/doc/submission-guidelines-release-genres-styles) system available. So if like me you wish to follow this it is impossible without styles. Discog only have [15 genres](https://reference.discogslabs.com/browse/genre) but [425 styles](https://reference.discogslabs.com/browse/styles).

As an example if we search [Bob Dylan - Fallen Angels](https://www.discogs.com/Bob-Dylan-Fallen-Angels/master/1002633) we see its associated with three genres:  **Rock**; **Blues**; **Pop** and one style: **Vocal**.  The [Vocal](https://reference.discogslabs.com/style/Vocal) style is a "sub-genre" of Pop.

To my knowledge the Discog API does not support probing this genre/style relationship structure unfortunately but simply provides the tags to each album.  So in my opinion it would be adequate for at least now to simply start recording album styles simply as a STYLE metatag against each track.  This is what currently manually do, so for our Fallen Angels example I would simply have the following ```
```metatags:
genre="Rock; Blues; Pop"
style="Pop"
```
Initially I would like the ability to automatic this process from Beets.

If there was is a way Beets is able to build on this in the future then even better.  I would for example to be able to see what albums in my collection have a style of Vocal and genres that include "Folk, World, & Country".

Thank you

How to store this information within Beets

thedevilisinthedetails:
Adding styles to discogs plugin. Closes #2579.

### DIFF
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -131,7 +131,6 @@ match:
         track_index: 1.0
         track_length: 2.0
         track_id: 5.0
-        style: 5.0
     preferred:
         countries: []
         media: []
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -131,6 +131,7 @@ match:
         track_index: 1.0
         track_length: 2.0
         track_id: 5.0
+        style: 5.0
     preferred:
         countries: []
         media: []
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -356,7 +356,6 @@ class DiscogsPlugin(BeetsPlugin):
         else:
             return self.config['separator'].as_str().join(sorted(style))
 
-
     def get_artist(self, artists):
         """Returns an artist string (all artists) and an artist_id (the main
         artist) for a list of discogs album or track artists.
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -304,9 +304,11 @@ class DiscogsPlugin(BeetsPlugin):
         country = result.data.get('country')
         data_url = result.data.get('uri')
         style = result.data.get('styles')
+        print('style', style)
         if style is None:
-            self._log.info('Style not Found')
-        elif len(style) == 0:
+            self._log.debug('Style not Found')
+        elif not style:
+            print('s', style)
             return style
         else:
             style = self.config['separator'].as_str().join(sorted(style))
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -306,7 +306,7 @@ class DiscogsPlugin(BeetsPlugin):
         style = result.data.get('styles')
         if style is None:
             self._log.debug('Style not Found')
-        elif len(style) == 0:
+        elif not style:
             return style
         else:
             style = self.config['separator'].as_str().join(sorted(style))
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -302,6 +302,13 @@ class DiscogsPlugin(BeetsPlugin):
         mediums = [t.medium for t in tracks]
         country = result.data.get('country')
         data_url = result.data.get('uri')
+        style = result.data.get('styles')
+        if style is None:
+            self._log.info('Style not Found')
+        elif len(style) == 0:
+            return style
+        else:
+            style = ' - '.join(sorted(style))
 
         # Extract information for the optional AlbumInfo fields that are
         # contained on nested discogs fields.
@@ -339,7 +346,8 @@ class DiscogsPlugin(BeetsPlugin):
                          day=None, label=label, mediums=len(set(mediums)),
                          artist_sort=None, releasegroup_id=master_id,
                          catalognum=catalogno, script=None, language=None,
-                         country=country, albumstatus=None, media=media,
+                         country=country, style=style,
+                         albumstatus=None, media=media,
                          albumdisambig=None, artist_credit=None,
                          original_year=original_year, original_month=None,
                          original_day=None, data_source='Discogs',
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -55,6 +55,7 @@ class DiscogsPlugin(BeetsPlugin):
             'tokenfile': 'discogs_token.json',
             'source_weight': 0.5,
             'user_token': '',
+            'separator': u', '
         })
         self.config['apikey'].redact = True
         self.config['apisecret'].redact = True
@@ -303,14 +304,12 @@ class DiscogsPlugin(BeetsPlugin):
         country = result.data.get('country')
         data_url = result.data.get('uri')
         style = result.data.get('styles')
-        print('style', style)
         if style is None:
             self._log.info('Style not Found')
-            return "Style not Defined"
         elif len(style) == 0:
             return style
         else:
-            style = ' - '.join(sorted(style))
+            style = self.config['separator'].as_str().join(sorted(style))
 
         # Extract information for the optional AlbumInfo fields that are
         # contained on nested discogs fields.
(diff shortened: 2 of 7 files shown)


beetsplug/discogs.py
```diff
@@ -302,6 +302,15 @@ def get_album_info(self, result):
         mediums = [t.medium for t in tracks]
         country = result.data.get('country')
         data_url = result.data.get('uri')
+        style = result.data.get('styles')
+        print('style', style)
```
sampsyo: Looks like some stray debugging statements got left in.

beetsplug/discogs.py
```diff
@@ -302,6 +302,15 @@ def get_album_info(self, result):
...
+        if style is None:
+            self._log.info('Style not Found')
+            return "Style not Defined"
```
sampsyo: I don't think you want to `return` anything in these cases.

beets/config_default.yaml
```diff
@@ -131,6 +131,7 @@ match:
         track_index: 1.0
         track_length: 2.0
         track_id: 5.0
+        style: 5.0
```
sampsyo: It doesn't look like style is used for matching, so this is probably unnecessary.

beetsplug/discogs.py
```diff
@@ -302,6 +303,13 @@ def get_album_info(self, result):
         mediums = [t.medium for t in tracks]
         country = result.data.get('country')
         data_url = result.data.get('uri')
+        style = result.data.get('styles')
+        if style is None:
+            self._log.info('Style not Found')
```
sampsyo: I think this should probably be a debug message (it's probably not urgent enough to interrupt the `import` output).

beetsplug/discogs.py
```diff
@@ -302,6 +303,13 @@ def get_album_info(self, result):
...
+        elif len(style) == 0:
```
sampsyo:
```suggestion
        elif not style:
```

beetsplug/discogs.py
```diff
@@ -302,6 +303,13 @@ def get_album_info(self, result):
...
+            return style
```
sampsyo: I expect you don't mean to return early from this function—perhaps you want `style = ''` or similar? (But the would also be the effect of the `else:` branch here.)

test/test_discogs.py
```diff
@@ -81,6 +84,8 @@ def _make_release_from_positions(self, positions):
         tracklist where tracks have the specified `positions`."""
         tracks = [self._make_track('TITLE%s' % i, position) for
                   (i, position) in enumerate(positions, start=1)]
+        release = self._make_release(tracks)
+        print('release', release)
```
sampsyo: Looks like this got left in by mistake.

test/test_discogs.py
```diff
@@ -89,6 +94,7 @@ def test_parse_media_for_tracks(self):
         release = self._make_release(tracks=tracks)
 
         d = DiscogsPlugin().get_album_info(release)
+        print('albumInfo', d.media)
```
sampsyo: Another stray debugging print.

tox.ini
```diff
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py27-test, py37-test, py38-test, py27-flake8, docs
+envlist = py27-test, py37-test, py27-flake8, docs
```
sampsyo: Not sure why this is showing up in the diff? `py38-test` is disabled by default on master.

beetsplug/discogs.py
thedevilisinthedetails: This statement is checking to see if there is one item in the array. Is what you suggested, "not style" making the same check?

sampsyo: Yes. (In Python, empty lists are falsy.)

beetsplug/discogs.py
thedevilisinthedetails: I want it to return the only one style pulled fro discogs. No formatting or join is needed in the case when there is only one style tag.

sampsyo: But—correct me if I'm wrong—this is in the context of the `get_album_info` function, which is supposed to return an `AlbumInfo` object. By returning early here, it's instead returning a string from this function. Am I misunderstanding something here?

thedevilisinthedetails: You're right. I can create a function outside of this function that creates the format for style(s)

beetsplug/discogs.py
```diff
@@ -339,12 +341,21 @@ def get_album_info(self, result):
...
                          albumdisambig=None, artist_credit=None,
                          original_year=original_year, original_month=None,
                          original_day=None, data_source='Discogs',
                          data_url=data_url)
 
+    def format_style(self, style):
+        if style is None:
+            self._log.debug('Style not Found')
+        elif not style:
+            return style
```
sampsyo: This seems to return `[]`, i.e., an empty list, instead of a string. I think it should be fine to remove this entire `elif` case: if the string is empty, then `'something'.join(sorted([]))` will return `''`, which is probably what we want.

thedevilisinthedetails: :D Think there was some confusion. When i was checking len == 0 it should have been len== 1. If there is only one style tag in the response then I want to just return that style tag as is with no formatting. So I guess I just needed to chage len == 0 to len == 1. Hope this clarifies what I am trying to achieve. If this makes sense I will make the change.

sampsyo: It does! But I still don't think this `elif` case is necessary for that—a single style in the list will still "do nothing" when joined. For example:

```py
>>> ', '.join(['foo'])
'foo'
```

…so the one-element list case is also handled by the general case.

thedevilisinthedetails: 👍 fixed

sampsyo:
Awesome; thank you!! Merged with a changelog entry.

thunderousleader10:
Hi did this feature ever get added to the plugin? If so how do I use it? Thanks

jordanbs:
Thank you for this feature! One issue to report:

It looks like the styles tag is only being written to the beets library, and not to the id3 tag itself. `beets write` has no impact.
```
% beet info -l album:Marius title:"Don't"
[..snip...]
style: Ambient, House, Techno

% eyeD3 "04 Don't.mp3"
04 Don't.mp3                                                                                                                                        [ 8.15 MB ]
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Time: 03:31	MPEG1, Layer III	[ 320 kb/s @ 44100 Hz - Joint stereo ]
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
ID3 v2.4:
title: Don't
artist: Batu
album: Marius EP
album artist: Batu
original release date: 2017
recording date: 2017
track: 4/4		genre: Electronic (id 52)
disc: 1/1
BPM: 0
Publisher/label: Hessle Audio
Unique File ID: [b'http://musicbrainz.org'] : 10446667-B2
Lyrics: [Description: ] [Lang: XXX]

UserTextFrame: [Description: R128_ALBUM_GAIN]
0
UserTextFrame: [Description: R128_TRACK_GAIN]
0
UserTextFrame: [Description: CATALOGNUMBER]
HES030
UserTextFrame: [Description: MusicBrainz Album Id]
10446667
UserTextFrame: [Description: MusicBrainz Artist Id]
2742670
UserTextFrame: [Description: MusicBrainz Album Type]
12", EP
UserTextFrame: [Description: MusicBrainz Album Artist Id]
2742670
UserTextFrame: [Description: MusicBrainz Album Release Country]
UK
UserTextFrame: [Description: MusicBrainz Release Group Id]
1228709
FRONT_COVER Image: [Size: 109541 bytes] [Type: image/jpeg]
Description: 
```

sampsyo:
Indeed; `style` is a flexible attribute that currently has no associated tag mapping in [MediaFile](https://github.com/beetbox/mediafile).

Issue #3045, opened 2018-09-30
Bucket plugin doesn't work for artist with 4 digits in their names


elnull:
Bucket plugin doesn't work for artist with 4 digits in their names.
With setup like this:

bucket:
    bucket_alpha: ['0']
    bucket_alpha_regex:
        '0': ^[0-9]

paths:
    default: %bucket{$albumartist}/${albumartist}/%if{$year,$year-}$album/$disc-$track--$title

artists 047 and 417.3 are placed in directories 0/047 and 0/417.3, but artists 1099 and 1900 are placed in directories 1099/1099 and 1900/1900.

sampsyo:
Hi! Does this persist even if you use `beet modify` to manually add a digit to one of those three-digit names, or remove a digit from one of the four-digit names?

elnull:
Please note, I changed resulting directories names in bug report.

I've tried:
beet modify 047 artist="2047" 
beet modify 1900 artist="19001"

Tag artist was changed but directory structure remains the same.

sampsyo:
Hmm, but your configuration uses albumartist, not artist. Does that have an effect?

elnull:
I think the trouble is in function:
    def _tmpl_bucket(self, text, field=None):
        if not field and len(text) == 4 and text.isdigit():
            field = 'year'
        if field == 'year':
            func = self.find_bucket_year
        else:
            func = self.find_bucket_alpha

It seems it moves all 4 digits text in year bucket.

sampsyo:
Got it! So it’s just assuming any 4-digit number is a year, not text. Is there a way to approach this in a backwards compatible manner?

elnull:
I think, field argument should be defined when function is calling but I don't know how to do it.

sampsyo:
Interesting... is there a way to describe abstractly what change you think would be warranted? I might be able to help pinpoint what actual lines of code need to change if you have specific suggestions about how the behavior needs to be modified.

sampsyo:
Got it. Have you tried just using `%bucket{$albumartist,text}` or something? This will invoke the `_tmpl_bucket` function with `"text"` as the value for the second parameter.

elnull:
After changing to `%bucket{$albumartist,text}` I have right directory structure (0/1900, 0/1099). Thank you.

sampsyo:
Great! Maybe we should mention this in the documentation? Any chance you'd be willing to make a quick pull request with a sentence about that?

elnull:
[...]

PR #3255, opened 2019-05-06
Don't evaluate bucket template argument as year if there's no year buckets

dagon666:
The bucket template_func makes the decision whether its argument is an
artist/album or a year bucket definition by doing a simple check on the
argument. If the argument is 4 letters long and it's all digits it is
assumed that the argument relates to a year span bucket. This condition
is way too simplistic and leads to incorrect behaviour. One example is
artists and bands which are actually named as a year i.e. band: "1349".
In order to disable incorrect argument evaluation, this change adds a
check if there are any year span definitions at all. In case there's
none, the template_func should default to an alpha evaluation.

Signed-off-by: Tomasz Wisniewski <tomasz.wisni3wski@gmail.com>

### DIFF
--- a/beetsplug/bucket.py
+++ b/beetsplug/bucket.py
@@ -235,7 +235,7 @@ class BucketPlugin(plugins.BeetsPlugin):
         return s[0].upper()
 
     def _tmpl_bucket(self, text, field=None):
-        if not field and len(text) == 4 and text.isdigit():
+        if not field and len(text) == 4 and text.isdigit() and len(self.year_spans):
             field = 'year'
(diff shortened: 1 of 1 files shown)


sampsyo:
Hi! Thanks for your contribution! This problem has previously been discussed here: https://github.com/beetbox/beets/issues/3045#issuecomment-426023850

Can you describe how people’s existing configurations would need to change to cope with this new behavior? Also, it looks like the tests need some attention.

jtpavlock:
Closing as no response from author

PR #3257, opened 2019-05-08
List included configurations in verbose output

### LINKED ISSUE #3254: In verbose mode, show information about included config files
When running in verbose mode, you get `user configuration: .config/beets/config.yaml`. It would be nice to have information about all included files with `config.html#include` directive. Example:

You have in config.yaml:

```
include: ['tokens.yaml', '~/external.yaml']
```

Running with verbose on would give:

`user configuration: .config/beets/config.yaml, .config/beets/tokens.yaml, ~/external.yaml`

jackwilsdon:
Fixes #3254.

---

I still need to add tests for this.

### DIFF
--- a/beets/__init__.py
+++ b/beets/__init__.py
@@ -27,12 +27,26 @@ class IncludeLazyConfig(confit.LazyConfig):
     """A version of Confit's LazyConfig that also merges in data from
     YAML files specified in an `include` setting.
     """
+    def __init__(self, *args, **kwargs):
+        super(IncludeLazyConfig, self).__init__(*args, **kwargs)
+
+        self._included_files = []
+
+    def user_config_paths(self):
+        """Points to a list of locations making up the user configuration.
+
+        The files may not exist.
+        """
+        return [self.user_config_path()] + self._included_files
+
     def read(self, user=True, defaults=True):
         super(IncludeLazyConfig, self).read(user, defaults)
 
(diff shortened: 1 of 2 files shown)


beets/__init__.py
```diff
@@ -27,12 +27,26 @@ class IncludeLazyConfig(confit.LazyConfig):
...
+        super(IncludeLazyConfig, self).__init__(*args, **kwargs)
+
+        self._included_files = []
+
+    def user_config_paths(self):
+        """Points to a list of locations making up the user configuration.
+
+        The files may not exist.
+        """
+        return [self.user_config_path()] + self._included_files
```
jackwilsdon: I'm not sure how to solve this, but if the file hasn't already been read then the included file list will be empty - should this be a note in the docstring or should this method actually call `read` if the config has not already been read?

sampsyo:
This is pretty tricky. One option would be to just print out `[p.as_filename() for p in config['include']]`, which duplicates a bit of logic but would sidestep the issue of laziness?

beets/__init__.py
jtpavlock: Stylistically, this method sounds like a getter, which may be better off served as an `@property`. As for your point, the latter makes sense, as I would expect the code to do whatever it needs to do to get my the `user_config_paths` when I call it.

Issue #2388, opened 2017-01-13
Performance: Common operations are slow.


icefoxen:
Just started using beets and I love it, but holy cats is it slow sometimes, even for very simple things like simple queries and listings.  I only have ~10k songs and it's still rather a pain.

So, I'd like to make it faster, and welcome advice on how to do so.  The first suspect is database calls.  Compare how long it takes to list items for even a small library, vs. just reading the information from the database:

```
$ time beet ls | wc  
1.01user 0.12system 0:03.15elapsed 36%CPU (0avgtext+0avgdata 29912maxresident)k
0inputs+0outputs (0major+9671minor)pagefaults 0swaps
    385    2725   15198
```

```
$ time sqlite3 musiclibrary.blb 'select * from items;' | wc
0.11user 0.00system 0:00.11elapsed 97%CPU (0avgtext+0avgdata 5456maxresident)k
0inputs+0outputs (0major+697minor)pagefaults 0swaps
  11437  193901 3919008
```

(Also I'm not sure why my library of 385 songs has 11k items in it; that's something else to investigate.)

Some profiling suggests that this is a likely candidate:

```
$ python3 -m cProfile -s cumtime /usr/local/bin/beet ls | less
...
         703753 function calls (689872 primitive calls) in 3.696 seconds
   Ordered by: internal time
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1159    2.377    0.002    2.377    0.002 {method 'execute' of 'sqlite3.Connection' objects}
      770    0.083    0.000    0.297    0.000 db.py:173(_awaken)
    34677    0.074    0.000    0.119    0.000 types.py:93(from_sql)
```

So yes, it does spend most of its time talking to the DB, and adding more items to the library seems to add 3 more `sqlite3.Connection.execute()` calls per library item.  So, the first goal is to do more intelligent batching of database calls.  If I can make ls faster, then at best it will make a lot of other things faster as well, and at worst I'll be able to start attacking other problems.

Other performance issues to look into: https://github.com/beetbox/beets/issues/2370  https://github.com/beetbox/beets/issues/2281  https://github.com/beetbox/beets/issues/1795

### Setup

* OS: Debian Stretch
* Python version: 3.5.3
* beets version: 1.4.3 (git master, commit 272aa8870386b171c132206829509b410cd11d18)

sampsyo:
Hi! Thanks for looking into this—we could *really* use some help with digging into performance issues exactly like this.

[...]

Issue #2030, opened 2016-06-01
Perfomance: reuse compiled format templates


sampsyo:
The question arose recently whether our funky AST compiler in the `functemplate` module is actually worth it. I wrote this compiler some time ago hoping for a dramatic speedup in `beet list` time, but I don't think I ever documented evidence either way about the speedup, and I had a vague inkling that it wasn't having the effect we wanted. Since it's giving @jrobeson trouble with the Python 3 port, I decided to investigate whether the code was worth keeping at all.

TL;DR: The compiler can yield a 1.6x speedup on my machine! But we're currently using it in a way that negates all the benefits.

I used `time beet ls > /dev/null` to measure the time it takes to list all the items in my 15265-track library on my really old Core 2 Duo server. The time was 41.5 ± 0.3 seconds (mean and standard error over 4 runs). Not great.

Then I tried disabling the compiler (the `functemplate` module also includes a straightforward interpreter). Shockingly, the time actually got *worse:* 37.8 ± 0.3 seconds.

I investigated and found that, as you might have guessed, the code is written to invoke the compiler every time---so instead of compiling the template once, beets was compiling it 15 thousand times. :cry:

I patched the `list` command as a stopgap to use the compiler correctly:

```diff
diff --git a/beets/ui/commands.py b/beets/ui/commands.py
index 867a473..7c5b454 100644
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1057,12 +1057,14 @@ def list_items(lib, query, album, fmt=''):
     """Print out items in lib matching query. If album, then search for
     albums instead of single items.
     """
+    from beets.util.functemplate import Template
+    tmpl = Template(fmt)
     if album:
         for album in lib.albums(query):
-            ui.print_(format(album, fmt))
+            ui.print_(album.evaluate_template(tmpl))
     else:
         for item in lib.items(query):
-            ui.print_(format(item, fmt))
+            ui.print_(item.evaluate_template(tmpl))
```

and got much better time: 25.5 ± 0.2 seconds. Hot dang!

So the conclusions are:

* The compiler actually does address a meaningful bottleneck!
* We're completely missing out on the potential benefits!
* We need to reuse compiled templates, at least for the `list` command if not for all commands that use the standard formats from the config file.

---

Here are the raw timings:

```sh
# Status quo.
$ time beet ls > /dev/null
beet ls > /dev/null  42.01s user 1.02s system 99% cpu 43.131 total
$ time beet ls > /dev/null
[...]

PR #3258, opened 2019-05-08
Improvements to query performance

SimonPersson:
As discussed in PR #2388, this PR makes gives some performance improvements for querying. On my database of around 14k songs, these commits make `beet ls > /dev/null` run in 6.4 seconds, down from 15.5.

The second commit confuses me. For some reason, inlining the call to `formatted()` makes the difference between 11 seconds and 6.4 seconds. I was confused for quite some time, thinking that `FormattedMapping` itself was doing something slow, but it turns out that the function call overhead was causing it?!

In this PR I have used `lru_cache` to cache the formatting templates. I would have preferred to explicitly create the template at the call site, but I can't seem to figure it out. The function `list_items`, in `ui/commands.py`, receives an argument `fmt`. I figured that this argument would contain the formatting string, but it appears to be empty. Whether I specify a formatting string on the command line or not does not seem to matter. Any pointers here would be great. In it's current state, I imagine that this PR would break Python2 support. I don't know if you intend to support Python 2, given that it will be deprecated in half a year now.

Finally, I'm not sure how to run the tests? I run the `tox` command, but it complains about some interpreters missing. Am I expected to have every version of Python installed on my computer for this to work? Sorry, I'm not really a pythonista :)

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -25,7 +25,7 @@ import sqlite3
 import contextlib
 
 import beets
-from beets.util.functemplate import Template
+from beets.util import functemplate
 from beets.util import py3_path
 from beets.dbcore import types
 from .query import MatchQuery, NullSort, TrueQuery
@@ -597,7 +597,7 @@ class Model(object):
         """
         # Perform substitution.
         if isinstance(template, six.string_types):
-            template = Template(template)
+            template = functemplate.template(template)
         return template.substitute(self.formatted(for_path),
                                    self._template_funcs())
 
--- a/beets/library.py
+++ b/beets/library.py
@@ -31,7 +31,7 @@ from beets import plugins
 from beets import util
 from beets.util import bytestring_path, syspath, normpath, samefile, \
     MoveOperation
-from beets.util.functemplate import Template
+from beets.util.functemplate import template, Template
 from beets import dbcore
 from beets.dbcore import types
 import beets
@@ -855,7 +855,7 @@ class Item(LibModel):
         if isinstance(path_format, Template):
             subpath_tmpl = path_format
         else:
-            subpath_tmpl = Template(path_format)
+            subpath_tmpl = template(path_format)
 
         # Evaluate the selected template.
         subpath = self.evaluate_template(subpath_tmpl, True)
@@ -1134,7 +1134,7 @@ class Album(LibModel):
         image = bytestring_path(image)
         item_dir = item_dir or self.item_dir()
 
-        filename_tmpl = Template(
+        filename_tmpl = template(
             beets.config['art_filename'].as_str())
         subpath = self.evaluate_template(filename_tmpl, True)
         if beets.config['asciify_paths']:
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -36,7 +36,7 @@ from beets import logging
 from beets import library
 from beets import plugins
 from beets import util
-from beets.util.functemplate import Template
+from beets.util.functemplate import template
 from beets import config
 from beets.util import confit, as_string
 from beets.autotag import mb
@@ -616,7 +616,7 @@ def get_path_formats(subview=None):
     subview = subview or config['paths']
     for query, view in subview.items():
         query = PF_KEY_QUERIES.get(query, query)  # Expand common queries.
-        path_formats.append((query, Template(view.as_str())))
+        path_formats.append((query, template(view.as_str())))
     return path_formats
 
 
--- a/beets/util/functemplate.py
+++ b/beets/util/functemplate.py
@@ -553,12 +553,21 @@ def _parse(template):
         parts.append(remainder)
     return Expression(parts)
 
-@functools.lru_cache(maxsize=128)
+
+# Decorator that enables lru_cache on py3, and no caching on py2.
+def cached(func):
+    if six.PY2:
+        # Sorry python2 users, no caching for you :(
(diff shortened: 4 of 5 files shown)


beets/dbcore/db.py
```diff
@@ -581,16 +581,6 @@ def add(self, db=None):
                     self._dirty.add(key)
             self.store()
 
-    # Formatting and templating.
-
-    _formatter = FormattedMapping
```
wisp3rwind: This change is not as innocuous as it seems: The `_formatter` attribute is used by the `beets.library.Item` subclass of `Model` which replaces `FormattedMapping` by `FormattedItemMapping`. I would strongly suspect that this is the reason for the performance improvement, not having an additional factory method wrapping it. Without reading the code, I don't know what the exact implications are, but this commit must surely be breaking some of beets functionality.

There might of course be room for optimization in `FormattedItemMapping`.

SimonPersson: Ah, thanks! I couldn't figure out why this would impact performance so much. I guess I'll investigate this further.

wisp3rwind:
Thanks for tackling this long-standing issue! However, I believe the overall role of templates in beets should be investigated somwhat further before deciding on a strategy for fixing it. In particular, @sampsyo made a few suggestions in [2030](https://github.com/beetbox/beets/issues/2030#issuecomment-240174354): There does not seem to be the obvious way to approach caching templates.
One additional point is that I think it might be good to keep the general code layout in mind: Templates are not really an integral part of `dbcore` (IIRC, there is the loose idea that it would be nice to make it completely independent of the rest of beets). Thus, it might be preferable not to entangle this two parts of beets even more deeply.

> In this PR I have used lru_cache to cache the formatting templates. I would have preferred to explicitly create the template at the call site, but I can't seem to figure it out. The function list_items, in ui/commands.py, receives an argument fmt. I figured that this argument would contain the formatting string, but it appears to be empty. Whether I specify a formatting string on the command line or not does not seem to matter. Any pointers here would be great. In it's current state, I imagine that this PR would break Python2 support. I don't know if you intend to support Python 2, given that it will be deprecated in half a year now.

You already mentioned #2030 in your comments in #2388: I think one of the two approaches I benchmarked is very close to what you intended to do. Note the comments about the default template.

I believe there are plans to drop Python 2 support in the near future,but as long as we still support it, this change must at least be disabled when on Python 2 (which might be an acceptable approach if Python 2 support is indeed removed soon-ish).

> Finally, I'm not sure how to run the tests? I run the tox command, but it complains about some interpreters missing. Am I expected to have every version of Python installed on my computer for this to work? Sorry, I'm not really a pythonista :)

Without any traceback, I cannot really tell what the issue is. Note that you can run individual test environments using `tox -e`.

SimonPersson:
> You already mentioned #2030 in your comments in #2388: I think one of the two approaches I benchmarked is very close to what you intended to do. Note the comments about the default template.

Thanks, I should have read the issue more carefully. So the suggestion to create a `functemplate.template()` function that is cached, and instantiates templates seems reasonable to me. I've pushed a commit with this change. I still used `lru_cache`, but if we agree that the approach is good I will look into how to make it python2 compatible.

I looked into the performance of `FormattedItemMapping`. The problem seems to be that the formatter fetches the related album for every item, even when it is not needed. One solution to this is to fetch the album lazily, only when we really need it. This makes the default `beet ls` fast, whereas `beet ls -f '$artpath'` runs a bit slower, as it needs to fetch the album. For most attributes, the album does not seem to be needed, so this should speed up "simple" of queries. I've pushed a commit with this change as well.

sampsyo:
Awesome work so far! This is super interesting. It seems like there are two main things here: LRU-caching the templates, and making `FormattedItemMapping` lazy. (The latter is a great insight—we didn't previously know this layer was costing so much.) Any chance I could convince you to split one of them off into a separate pull request for review (since they seem orthogonal)?

I don't see any big downsides to using `@lru_cache` (we can definitely figure out the Python 2 story). And I'd also be happy to dig into why you're seeing that mysterious behavior that was preventing explicit reuse…

SimonPersson:
@sampsyo - Sure, I created PR #3260 for the changes to `FormattedItemMapping`.

SimonPersson:
I fixed the flake8 errors, and disabled `lru_cache` on python2, so tests are passing now.

sampsyo:
I can confirm the performance impact on my machine! Here are three `time` runs _without_ the change:

```
beet ls > /dev/null  4.94s user 0.38s system 99% cpu 5.355 total
beet ls > /dev/null  4.90s user 0.36s system 99% cpu 5.288 total
beet ls > /dev/null  4.87s user 0.36s system 99% cpu 5.255 total
```

And _with_ it:

```
beet ls > /dev/null  6.99s user 0.41s system 99% cpu 7.424 total
beet ls > /dev/null  7.06s user 0.41s system 99% cpu 7.492 total
beet ls > /dev/null  7.09s user 0.43s system 99% cpu 7.589 total
```

So that's a change from about 7.0 seconds to about 4.9 seconds, or a 1.4x speedup. :tada:

sampsyo:
To answer your question:

> Finally, I'm not sure how to run the tests? I run the tox command, but it complains about some interpreters missing. Am I expected to have every version of Python installed on my computer for this to work? Sorry, I'm not really a pythonista :)

Yep, tox depends on Python interpreters for each version you want to test under. You can always choose a specific environment by typing something like `tox -e py37`.

Issue #3049, opened 2018-10-11
Updates "r128_album_gain: 000000 -> 000000" With or without replaygain plugin


K33k0:
### Problem

Running beets update leeds to the r128 tags been updated to 000000 even if its already at 000000. If I run the replaygain plugin manually this tag is updated correctly, only to be wiped again when beet update is run.

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv update
```

Led to this problem:

```
skipping /mnt/data/music/Pearl Jam/Vitalogy/06 Whipping.ogg because mtime is up to date (1537958068.0)
Pearl Jam - Ten - Alive
  r128_album_gain: 000000 -> 000000
Sending event: before_item_moved
Sending event: item_moved
Sending event: database_change
Pearl Jam - Ten - Why Go
  r128_album_gain: 000000 -> 000000
Sending event: before_item_moved
Sending event: item_moved
Sending event: database_change
```
(Happens to all my music)

### Setup

* OS: Ubuntu bionic
* Python version: 3.6.5
* beets version: 1.4.8
* Turning off plugins made problem go away (yes/no): No

My configuration (output of `beet config`) is:

```yaml
# Physical Location of your *Processed* music
directory: /mnt/data/music
# Where to store your beets database
library: /home/admin/.config/beets/musiclibrary.db

import:
  # write metadata to music files
  write: yes

  # move imported files from source to the music directory
  move: yes
  copy: no
  delete: no

  # use auto-tagging where possible
  # do not require confirmation on strong matches
  autotag: yes
  timid: no

  resume: ask
  incremental: no
  none_rec_action: ask
  log: /home/admin/logs/beets.log

  quiet: no
  quiet_fallback: skip
  default_action: apply


  singletons: no
  detail: no
  flat: no

#  duplicate_action: skip

# use the release-date of the original (first) release of an album?
original_date: yes

# on multi-disk releases, assign track numbers for the whole album.
# If "per disk", make sure tracknames do not collide ("paths" setting).
per_disc_numbering: no

# files matching these patterns are deleted from source after import
clutter: ["Thumbs.DB", ".DS_Store", "*.m3u", ".pls", "*.jpg"]

# files/directories matching one of these patterns are ignored during import
ignore: [".*", "*~", "System Volume Information"]

# Paths and filenames for music files
# relative to music directory
paths:
    default: $albumartist/$album%aunique{}/$track $title
    singleton: Non-Album/$artist/$title
[...]

PR #3261, opened 2019-05-09
Use NullPaddedInt for all r128_album_gain fields.

djl:
See [my comment](https://github.com/beetbox/beets/issues/3049#issuecomment-491009715) on #3049.


sampsyo:
This looks like exactly the thing to do. #2813 introduced NullPaddedInt for those fields, and it seems to have left out the album version.

djl:
I intentionally left out the changelog entry because the `NullPaddedInt` changes haven't yet made it into a release. This is just a small bugfix and should be covered by the that change's changelog entry (right?).

sampsyo:
Yep, sounds good.

PR #3267, opened 2019-05-13
Update setup.py to install to a directory rather than egg file

### LINKED ISSUE #3264: Data files are unreadable when beets is installed as an egg
Following up from [a related Discourse thread](https://discourse.beets.io/t/error-running-from-source/739?u=arcresu), we noticed that beets isn't happy when it's installed as an egg. This is currently the default way beets is installed if you run `setup.py install` (not sure if #2083 changes this story).

#### Problem
In egg mode, the whole beets and beetsplug directory trees remain bundled in a zipfile and are not extracted. This causes problems when we try and read data files (aka package resources)  like `config_default.yaml`, because we currently build up the path to such files relative to the executing module's source file and `open()` the resulting path. Obviously this doesn't work of the file is inside a zipfile.

* Symptom: `ConfigError`s like `configuration error: verbose not found`.
 Caused by the failure to open `config_default.yaml` (aside: it would have been clearer if we got a `confit.ConfigReadError` rather than missing default values, but the loading of the default config file is inside a `if os.path.isfile(filename)` guard so it just silently skips the file when it doesn't exist even though beets depends on the default values being set).
Needs to be fixed in confuse.

* Symptom: `FileNotFoundError`s related to files like `lastgenre/genres.txt`.
Caused by failure to open other data files needed by plugins.
Needs to be fixed in beets.

#### Solutions

* Explicitly state that we don't support running beets as an egg. This seems unfortunate if it's the result of `setup.py install`.
* Try and fix the way data files are accessed. There is [`pkg_resources`](https://setuptools.readthedocs.io/en/latest/pkg_resources.html) from `setuptools` that has helpers to transparently handle this. We could refactor these data file accesses into a helper for plugins.

arogl:
useful when run as
```
python setup.py install
```
Fixes #3264

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,6 +65,7 @@ setup(
     platforms='ALL',
     long_description=_read('README.rst'),
     test_suite='test.testall.suite',
+    zip_safe=False,
     include_package_data=True,  # Install plugin resources.
 
     packages=[


arcresu:
@jackwilsdon do you have any thoughts on this change to `setup.py`? It seems like the right thing to do to me, but I know you've looked at the packaging code before.

jackwilsdon:
Seems good to me @arcresu, all we're really doing with this change is telling setuptools that our package can't be run from a zip file — I can't see this causing any issues :+1:

PR #3269, opened 2019-05-17
Fix ImageMagick Detection

fortes:
The `return` statement was at the wrong indent level, so we never checked for the legacy `convert` executable. I ran into this with the 1.4.8 upgrade, getting warnings since `convert` no longer gets detected.

@ababyduck looks like you were last in this code, this look OK to you?

### DIFF
--- a/beets/util/artresizer.py
+++ b/beets/util/artresizer.py
@@ -278,7 +278,7 @@ def get_im_version():
                                int(match.group(3)))
                     return version, legacy
 
-        return None
+    return None
 
 
 def get_pil_version():


ababyduck:
Yikes, good catch!

The mistake I made in testing this was using WSL as a "legacy" environment. Even though I had Ubuntu running an older ImageMagick package, WSL apparently falls back to `magick` on the Windows path on the first iteration, when I thought it was simply the next iteration in `get_im_version()` trying `convert`.

sampsyo:
Thanks!! I've added a changelog entry.

PR #3272, opened 2019-05-25
add work, work-disambig and work_id tags

dosoe:
implements the first part of https://github.com/beetbox/beets/pull/2580 : changes to the core by adding  work, work-disambig and work_id tags. I didn't change the changelog, I don't really get how it works now. @sampsyo , @arcresu , how do I update the changelog now?


dosoe:
That should nail all changes to the core.

beets/autotag/mb.py
```diff
@@ -210,9 +210,18 @@ def track_info(recording, index=None, medium=None, medium_index=None,
...
+    work_disambig = []
     for work_relation in recording.get('work-relation-list', ()):
         if work_relation['type'] != 'performance':
             continue
+        work.append(work_relation['work']['title'])
+        work_id.append(work_relation['work']['id'])
+        if 'disambiguation' in work_relation['work']:
+            work_disambig.append(work_relation['work']['disambiguation'])
+        else:
+            work_disambig.append('')
```
sampsyo: Is this necessary? Seems like an empty list, in the case of no disambiguation, would be fine…

beets/autotag/mb.py
```diff
@@ -237,6 +246,11 @@ def track_info(recording, index=None, medium=None, medium_index=None,
                 arranger.append(artist_relation['artist']['name'])
     if arranger:
         info.arranger = u', '.join(arranger)
+    if work:
+        info.work = u', '.join(work)
+        info.work_id = u', '.join(work_id)
+        if all(dis for dis in work_disambig):
```
sampsyo: Under what circumstances would `dis` be false in this list? Would it be better to avoid inserting it into the list in the first place?

dosoe: Sorry, it was me not understanding correctly what ```join``` does.

dosoe: I think it was meant to not add an empty ```work_disambig``` field.

dosoe:
So should I wait for my PR until MediaFile is done moving?

dosoe:
If you don't mind, while I'm at it, I can also add the missing ```arranger_sort```and ```lyricist_sort``` and maybe their disambiguations.

beets/library.py
```diff
@@ -451,6 +451,9 @@ class Item(LibModel):
         'lyricist':             types.STRING,
         'composer':             types.STRING,
         'composer_sort':        types.STRING,
+        'work':                 types.STRING,
+        'work_id':              types.STRING,
```
sampsyo: Other MusicBrainz IDs (which are listed lower down in this file) are called `mb_trackid`, `mb_albumid`, etc. Can we please match that naming pattern?

beets/autotag/mb.py
```diff
@@ -57,6 +57,7 @@ def get_message(self):
             self._reasonstr(), self.verb, repr(self.query)
         )
 
+
```
sampsyo: Looks like a stray newline snuck in here?

beets/mediafile.py
```diff
@@ -1643,6 +1643,12 @@ def update(self, dict):
         StorageStyle('COMPOSERSORT'),
         ASFStorageStyle('WM/Composersortorder'),
     )
+    work = MediaField(
+        MP3DescStorageStyle(u'Work'),
+        MP4StorageStyle('----:com.apple.iTunes:Work'),
+        StorageStyle('WORK'),
+        ASFStorageStyle('beets/Work'),
+    )
```
sampsyo: Just to check: does Picard have a tag mapping for the work name? If so, we should do our best to match.

jackwilsdon: It looks like the iTunes one will want changing to `com.apple.iTunes:MusicBrainz Work Id`, and the "WORK" one will want changing to `MUSICBRAINZ_WORKID`, as these are the names [used by Picard](https://github.com/metabrainz/picard/blob/4983631f700c65786ea61b9e19c8a98a20205bb0/picard/formats/mp4.py#L95).

beets/library.py
dosoe: done

beets/autotag/mb.py
dosoe: done

beets/mediafile.py
dosoe: If I remember well I got the tag mapping from picard, but upon checking it turns out picard now has a tag for work title (Picard >=1.3) and MusicBrainz Work Id and a tag called Show Work & Movement (Picard>=2.1, only populated by some plugins). I'm updating, could you please check? I'm not 100% sure if it is correct.

dosoe: you find picards tag mapping here: https://picard.musicbrainz.org/docs/mappings/

dosoe: There is a tag conflict between Work and Grouping: in picard, if you take the option "Save iTunes compatible grouping and work" (since Picard>=2.1.0) then the ID3v2 tag for grouping becomes GRP1 and the tag for work TIT1, if it isn't enabled then the tags are TIT1 and  TXXX:WORK  respectively. I chose the first one, but you might disagree. I feel like if we have to choose, let's be ITunes compatible.

dosoe: This conflict and the subsequent change I made to the grouping tag seems to be the cause of the tests failing, an idea on how I could correct it? One would be to keep the old tags but I'm not really satisfied by this, do you have an opinion?

arcresu: Note that we already ported this particular change to MediaFile here: beetbox/mediafile#8

arcresu:
> So should I wait for my PR until MediaFile is done moving?

The MediaFile move is essentially complete, we're just delaying the final step to make sure we don't conflict with relevant open PRs like this one. It's currently identical to the copy of mediafile in beets, except that we added the MB work ID tag there already.

Besides the MB work ID, I'm not sure how related your other changes to MediaFile are to the main change here. In any case it would be ideal if you could submit those as a separate PR against [beetbox/mediafile](https://github.com/beetbox/mediafile) and remove all of the MediaFile changes from this PR.

beets/mediafile.py
```diff
@@ -1643,6 +1643,18 @@ def update(self, dict):
         StorageStyle('COMPOSERSORT'),
         ASFStorageStyle('WM/Composersortorder'),
     )
+    work = MediaField(
+        MP3StorageStyle('TIT1'),
```
arcresu: Isn't this going to cause some issues since the same storage tag was previously used for a different tag?

dosoe: Yes that's exactly the point raised here: 

> There is a tag conflict between Work and Grouping: in picard, if you take the option "Save iTunes compatible grouping and work" (since Picard>=2.1.0) then the ID3v2 tag for grouping becomes GRP1 and the tag for work TIT1, if it isn't enabled then the tags are TIT1 and TXXX:WORK respectively. I chose the first one, but you might disagree. I feel like if we have to choose, let's be ITunes compatible.

I think I will just not touch the mediafile and revert the changes.

dosoe:
The changes to mediafile have been reverted. Last thing to do before merge as far as I see it is to update the changelog, are there any subtleties regarding that? (AppVeyor fails because: ```The remote server returned an error: (503) Server Unavailable. Service Unavailable```)

dosoe:
So,

dosoe:
whoops closed accidentally...

dosoe:
Actually, I noticed that very few fields use the disambiguation, I wonder why?

beets/autotag/mb.py
dosoe: The point is that if there are several works, some have disambiguations and some don't, we want to keep track of which work has which disambiguation. I's a rather rare case, but in classical music, the name of the arrangement typically is in the disambiguation. Usually it is in the disambiguation of all works though. Maybe it would be a better thing to just integrate the disambiguation into the work title, but that might be too much... Actually, beets is not fetching the disambiguations for a while lot of things, first among them the artists/release artists, is there a reason for that?

sampsyo: I see! So in the case of a missing disambiguation string for a work, will we get something like `foo, , bar` in the final disambiguation string? I guess that's fine, but it looks a little weird…

> Actually, beets is not fetching the disambiguations for a while lot of things, first among them the artists/release artists, is there a reason for that?

The answer is pretty simple: nobody has ever asked for them. :smiley:

dosoe: Yes it's exactly that, and that's the best I could come up with... For works I feel that the disambiguations are needed, for example to distinguish all the different arrangements (even if one could argue that the 'arranger' tag is there for that, but then we would need an 'arranger_sort' tag) or the successive versions of a work (Mahler's symphonies or Bach's Passions have several versions, some works have versions for two pianos and others for orchestra or piano quintet like the piano quintet of Brahms). I just feel that for MB entities, the name, sort name and disambiguation belong together and fetching one without the others feels weird (especially name and sort name). 
Maybe we could assume that if a disambiguation affects one work, it will logically affect all the other works in the recording (as would be the case in all the examples I can think of) and don't need the sort of dirty meddling I did before. Maybe to make it even cleaner (but the code gets dirtier) we can organize it that there is only one disambiguation, we keep only the first disambiguation from the first work that has one.

arcresu:
It would be really nice if we had multi-valued fields (#505) but that seems like it's a long way from happening. MediaFile already supports list-type fields but I'm not very familiar with how that works. I feel like the ad-hoc serialisation of the lists in this PR might be a bit premature.

Could we simplify things by making the assumption that each track has only a single work? I know that it won't solve some of the more complicated cases like classical music, but there are other multi-valued fields like `artist` that we currently treat as single-valued fields. We can work to properly solve the handling of multi-valued fields in beets and fix all these fields consistently. In the short term, a single work for each track (e.g. the first one in MB) is still useful right?

dosoe:
Multi-valued fields would be perfect, do you have an idea on how far this is advanced? The whole point of this is to handle classical music, for other types of music you don't need the work (not that I know). Maybe we can keep only one disambiguation, but I would keep all the works.

dosoe:
Is it possible to store dicts in the database? This could solve our problems, and we would just need to handle the conversion from a dict to a tag. An artist/composer/... would have a dict with his name, sort name, credited name and disambiguation and the tag would be constructed from a list of dicts (a bit like it is done in musicbrainz-ngs.

sampsyo:
Multi-valued tags are sort of a long-term proposition—there are a lot of details to be worked out, so let's not depend on that happening to get this done. I think we should follow the behavior of other fields that also _could_ be multi-valued, such as artist, and just make do with either a joined string or picking the first. Going with the simplest possible solution for now, seeing how it pans out, and iterating over time seems like a good strategy.

arcresu:
Yep exactly, my suggestion was to only keep the first work for now because otherwise when we eventually come to solve multi-value tags properly we'll have to deal with different fields doing their own list serialisation.

I know that we'd like multiple works per track eventually, but from what I understood from the parentwork plugin that motivated this PR you only need to pick up a single work and then recurse up to a root work. I might have misunderstood something but it doesn't seem like you need more than one work per track to do that. Also, in this PR you've written the field descriptions as though they are all single-valued fields.

dosoe:
Sounds fine to me! Commit incoming.

arcresu:
Thanks for your work on this! I think this looks good to go now.

Multi-value tags is high on my list of tasks to tackle, but like Adrian said there's still quite a bit of design to sort out it, so it might not happen for a while yet.

sampsyo:
Merged with changelog! Thanks for your continued attention to this, @dosoe!

PR #3273, opened 2019-05-28
Add support for NO_COLOR.

djl:
This commit adds support for the [`NO_COLOR` environment variable](https://no-color.org/).

I was looking for a way to temporarily disable colors while piping the output of `beet mbsync -p` to a file. Rather than implement this as a top-level `--no-color` option, it was much easier to just add support for `NO_COLOR`.


sampsyo:
Awesome. I applaud the effort to make this a standard thing across lots of tools.

PR #3275, opened 2019-05-30
Make pathlib dependency in tests conditional

arcresu:
We already do this in the optional dependency section so this shouldn't cause any problems. It's only used by the `thumbnails` plugin.

I noticed that Debian was carrying a patch to remove this line since they install beets with Python 3 only and don't have a Py3 package for pathlib.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -115,10 +115,12 @@ setup(
         'rarfile',
         'responses',
         'pyxdg',
-        'pathlib',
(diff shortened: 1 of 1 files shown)


sampsyo:
Good call; thanks! I added a quick note to the changelog so the Debian people (and anyone else who's doing a similar thing) know they can remove that patch.

PR #3277, opened 2019-05-30
Setup path correctly in testall.py

arcresu:
I was unable to run the test suite using `testall.py` without removing this line. After this change it works when invoked from the top-level directory.

Our [Testing](https://github.com/beetbox/beets/wiki/Testing) wiki says one way to invoke tests is `cd test ; python testall.py`. I was unable to get this working:
  * From the `test` directory, calling `python testall.py` results in:
 ```
  File "testall.py", line 43, in <module>
    unittest.main(defaultTest='suite')
  File "/usr/lib/python3.7/unittest/main.py", line 100, in __init__
    self.parseArgs(argv)
  File "/usr/lib/python3.7/unittest/main.py", line 147, in parseArgs
    self.createTests()
  File "/usr/lib/python3.7/unittest/main.py", line 159, in createTests
    self.module)
  File "/usr/lib/python3.7/unittest/loader.py", line 220, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/lib/python3.7/unittest/loader.py", line 220, in <listcomp>
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/lib/python3.7/unittest/loader.py", line 205, in loadTestsFromName
    test = obj()
  File "testall.py", line 38, in suite
    s.addTest(__import__(modname).suite())
  File "<...>/beets/test/test_config_command.py", line 15, in <module>
    from test.helper import TestHelper
ModuleNotFoundError: No module named 'test.helper'
```
* From the top-level directory, `python test/testall.py` results in:
```
  File "test/testall.py", line 43, in <module>
    unittest.main(defaultTest='suite')
  File "/usr/lib/python3.7/unittest/main.py", line 100, in __init__
    self.parseArgs(argv)
  File "/usr/lib/python3.7/unittest/main.py", line 147, in parseArgs
    self.createTests()
  File "/usr/lib/python3.7/unittest/main.py", line 159, in createTests
    self.module)
  File "/usr/lib/python3.7/unittest/loader.py", line 220, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/lib/python3.7/unittest/loader.py", line 220, in <listcomp>
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/lib/python3.7/unittest/loader.py", line 205, in loadTestsFromName
    test = obj()
  File "test/testall.py", line 34, in suite
    for fname in os.listdir(pkgpath):
FileNotFoundError: [Errno 2] No such file or directory: 'test'
```

If I prevent it from changing directory from within `testall.py` then the second option works. If we accept this then we should also update that line on the wiki.

### DIFF
--- a/test/testall.py
+++ b/test/testall.py
@@ -22,16 +22,15 @@ import re
 import sys
 import unittest
 
-pkgpath = os.path.dirname(__file__) or '.'
-sys.path.append(pkgpath)
-os.chdir(pkgpath)
+pkgpath = os.path.dirname(os.path.dirname(os.path.realpath(__file__))) or '..'
+sys.path.insert(0, pkgpath)
 
 
 def suite():
     s = unittest.TestSuite()
     # Get the suite() of every module in this directory beginning with
     # "test_".
-    for fname in os.listdir(pkgpath):
+    for fname in os.listdir(os.path.join(pkgpath, 'test')):
         match = re.match(r'(test_\S+)\.py$', fname)
         if match:
             modname = match.group(1)

### CODE REVIEW

**Comment by sampsyo** [comment]
Yikes; this mechanism is pretty out of date! It seems to assume we're still relying on the old Python 2 relative import rules.

On my machine, oddly, I still get this error even with your change:

```
$ python3 test/testall.py
  File "test/testall.py", line 42, in <module>
    unittest.main(defaultTest='suite')
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/main.py", line 100, in __init__
    self.parseArgs(argv)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/main.py", line 147, in parseArgs
    self.createTests()
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/main.py", line 159, in createTests
    self.module)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/loader.py", line 220, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/loader.py", line 220, in <listcomp>
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/loader.py", line 205, in loadTestsFromName
    test = obj()
  File "test/testall.py", line 37, in suite
    s.addTest(__import__(modname).suite())
  File "/Users/asampson/Documents/code/beets/test/test_ui_importer.py", line 25, in <module>
    from test._common import DummyIO
ModuleNotFoundError: No module named 'test._common'
```

I thought the right fix might be to make sure we add the parent directory to `sys.path` instead of the test directory itself:

```py
pkgpath = os.path.dirname(__file__) or '.'
sys.path.append(os.path.dirname(os.path.abspath(pkgpath)))
```

but for reasons I don't quite understand, that doesn't work. I can confirm that `sys.path` has the `beets` directory on it, but the `test._common` import is still failing. I'll need to dig a little deeper…

arcresu:
There's definitely something strange here. I can't say I'm super familiar with the unittest ecosystem, but I imagine there must be a better test discovery mechanism by now that avoids the need for this sort of script in the first place. I'll take a look into that and see if we can simplify things a bit.

jackwilsdon:
Our CI just seems to run the tests through `nose` directly - is there a reason we can't just remove `testall.py` completely and document that users should run `nose` instead?

https://github.com/beetbox/beets/blob/1da34f26ff1017879e34dcb18efd6715847a753f/tox.ini#L45-L48

arcresu:
It looks like `testall.py` is [used by `setup.py`](https://github.com/beetbox/beets/blob/1da34f26ff1017879e34dcb18efd6715847a753f/setup.py#L67) to discover the tests.

jackwilsdon:
Interestingly enough it looks like running `python setup.py test` (which uses the discovered tests) actually works fine on my machine, however I still get the same `_common` error you guys are seeing when running `testall.py` directly.

sampsyo:
FWIW, the original reason I wrote that script (which is still semi-valid) is to have a way to run the tests even without a special test runner. The tests themselves only need `unittest` itself, so while nose certainly makes it more *fun* to use the tests, it's always possible to run them in a "standalone" way. Obviously not the most important reason in the world!

arcresu:
Yes I saw the same. I'm a bit confused about that.

In Debian they run the tests in two ways. The default python helpers for deb packaging invoke the suite as `setupy.py test`. Debian is even carrying a patch to `testall.py` to let them bypass a certain test file (it's not really relevant to us as upstream). Their CI system on the other hand calls `python -m unittest`. It's possible to reconfigure those tools to use a different runner, but I think we should make sure that `setup.py test` still runs the tests. I don't think we need to support a standalone runner script if there's a better way to do test discovery though.

As a separate issue Debian is running into some [flaky tests](https://bugs.debian.org/855092) on their build infrastructure that they're just disabling. I think that's worth fixing too. I _think_ that one of the issues they were having might be related to `setup.py test` first installing beets as a zipped egg and then running tests.

jackwilsdon:
So interestingly enough, here is my `sys.path` when running through `testall.py`;

```
['/tmp/beets/test', '/usr/lib/python36.zip', '/usr/lib/python3.6', '/usr/lib/python3.6/lib-dynload', '/tmp/beets/venv/lib/python3.6/site-packages', '.']
```

And through `nose`;

```
['/tmp/beets', '', '/usr/lib/python36.zip', '/usr/lib/python3.6', '/usr/lib/python3.6/lib-dynload', '/tmp/beets/venv/lib/python3.6/site-packages']
```

If we sort them, these are the differences;

```diff
--- testall.py
+++ nose
-"."
+""
-"/tmp/beets/test"
+"/tmp/beets"
```

So it looks like the main difference is `/tmp/beets` is missing from `sys.path`, and is instead replaced with `/tmp/beets/test` (note that `/tmp/beets` is where beets is checked out). We also have to _prepend_ `/tmp/beets` to `sys.path`, as otherwise Python's built-in `test` module takes over.

Changing the code to this seems to work for me;

```diff
diff --git a/test/testall.py b/test/testall.py
index 88eb7011..418b4a3c 100755
--- a/test/testall.py
+++ b/test/testall.py
@@ -22,16 +22,15 @@ import re
 import sys
 import unittest
 
-pkgpath = os.path.dirname(__file__) or '.'
-sys.path.append(pkgpath)
-os.chdir(pkgpath)
+pkgpath = os.path.dirname(os.path.dirname(os.path.realpath(__file__))) or '..'
+sys.path.insert(0, pkgpath)
 
 
 def suite():
     s = unittest.TestSuite()
     # Get the suite() of every module in this directory beginning with
     # "test_".
-    for fname in os.listdir(pkgpath):
+    for fname in os.listdir(os.path.join(pkgpath, 'test')):
         match = re.match(r'(test_\S+)\.py$', fname)
         if match:
             modname = match.group(1)
```

sampsyo:
I didn't know about `python -m unittest`! :flushed: That looks like a good alternative—maybe we should remove `testall.py` altogether and have `setup.py test` run `python -m unittest test/test_*.py` instead.

sampsyo:
Wow! That would do it. Nice detective work, @jackwilsdon.

arcresu:
The full command [run by the Debian configuration](https://sources.debian.org/src/beets/1.4.7-2/debian/tests/unittests/) is `python -m unittest discover -p 'test_*' 2>&1`

arcresu:
Thanks @jackwilsdon, I updated this PR to be your patch. It seems to do the job for me in a quick local test. Maybe we can merge this for now and then think about removing the script altogether as a separate change?

sampsyo:
Sounds good to me!

PR #3278, opened 2019-05-30
functemplate: Adapt ast syntax to PEP570 changes on python3.8

wisp3rwind:
Python 3.8 adds the required `posonlyargs` to `ast.arguments` (see the [Grammar](https://docs.python.org/3.8/library/ast.html?highlight=posonlyargs#abstract-grammar)), resulting in failures such as these: https://travis-ci.org/geigerzaehler/beets-alternatives/jobs/538912903.

There's some discussion about similar breaking API changes (namely in the inspect module) [here](https://github.com/python/cpython/pull/12701/files) and [here](https://bugs.python.org/issue36751),  which however does not mention the lower level ast syntax changes, so I'm assuming that the ast module is supposed to always reflect the grammer such that this is not to be considered a regression.

Does this warrant a changelog entry? Due to the werkzeug incompatibility, Python 3.8 still is not fully supported.

### DIFF
--- a/beets/util/functemplate.py
+++ b/beets/util/functemplate.py
@@ -118,29 +118,35 @@ def compile_func(arg_names, statements, name='_the_func', debug=False):
     bytecode of the compiled function.
     """
     if six.PY2:
-        func_def = ast.FunctionDef(
-            name=name.encode('utf-8'),
-            args=ast.arguments(
-                args=[ast.Name(n, ast.Param()) for n in arg_names],
-                vararg=None,
-                kwarg=None,
-                defaults=[ex_literal(None) for _ in arg_names],
-            ),
-            body=statements,
-            decorator_list=[],
+        name = name.encode('utf-8')
+        args = ast.arguments(
+            args=[ast.Name(n, ast.Param()) for n in arg_names],
+            vararg=None,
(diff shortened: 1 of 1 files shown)


sampsyo:
Thanks for pointing this out! It is sort of frustrating that these AST incompatibilities keep cropping up, but I think that's unavoidable.

For a small improvement in legibility/maintainability, what do you think about using a "feature test" instead of a version check to see whether the appropriate field needs to be included? The `ast.arguments` class has a `_fields` member that seems to list all its AST children. For example, on Python 3.7:

```py
>>> 'kwonlyargs' in ast.arguments._fields
True
>>> 'posonlyargs' in ast.arguments._fields
False
```

I think a short changelog entry might be a good idea, even if we need to mention that "real" Python 3.8 compatibility will need to wait for Werkzeug. Especially because Werkzeug is not a "core" dependency (it's only necessary for the web plugin).

sampsyo:
Gave that a try above.

wisp3rwind:
Thanks for finishing this up for the release :tada: Sorry for not replying, I was away from keyboard for a few days...

sampsyo:
No worries at all! Thanks again for your help.

Issue #266, opened 2013-05-06
Track when specific plugins last successfully updated an item's metadata


telejester:
In the course of a large import, remote services used by beets plugins may become unavailable and then reappear, causing errors and uneven results which the user is unable to find and repair.

Beets needs a way to know that a plugin has successfully operated on an asset so that the user may select the affected assets and resubmit then to the plugin for processing. 

Any solution should also consider the case where a plugin may be upgraded, which may create a need to select and resubmit assets which were already processed by an older version of the plugin.

In a similar vein, more than one plugin may attempt to modify a standard (i.e., not one that can be namespaced to a given plugin. Lyrics, for instance) metadata field. Which one wins? Can attempted clobbering be detected and presented to the user?

arcresu:
The second part of this suggestion is covered by #311. The common "metadata fetcher" plugin helpers can be solved as part of beets' planned plugin refactor (#607). I've renamed this issue to make the focus a little clearer.

PR #3279, opened 2019-05-31
add parentwork plugin

dosoe:
New try, closes #2580 .


beetsplug/parentwork.py
```diff
@@ -0,0 +1,197 @@
...
+from __future__ import division, absolute_import, print_function
+
+from beets import ui
+from beets.plugins import BeetsPlugin
+
+import musicbrainzngs
+
+
+def work_father(mb_workid, work_date=None):
+    """ This function finds the id of the father work given its id"""
```
sampsyo: The naming is a little unclear here—what exactly is a "father work" and how is it different from a "parent work"? Maybe something like "parent" vs. "ancestor" would be clearer? Also, it would be nice if the docstring described (in a sentence or two) exactly what's these things mean instead of just referencing the name.

Also, this function seems to return two things, not just one.

sampsyo: Also, FWIW, it's not necessary to say "This function does X…" It's usually more appropriate to just say "Do X." This series of examples might be helpful:
https://blog.dolphm.com/pep257-good-python-docstrings-by-example/

beetsplug/parentwork.py
```diff
@@ -0,0 +1,197 @@
...
+    work_date = None
+    while True:
+        (new_mb_workid, work_date) = work_father(mb_workid, work_date)
+        if not new_mb_workid:
+            return mb_workid, work_date
+        mb_workid = new_mb_workid
+    return mb_workid, work_date
+
+
+def find_parentwork(mb_workid):
```
sampsyo: Similarly, with the naming: `find_parentwork` sounds like it does the same thing as `work_parent`. :smiley: Naming some of the functions to use "id" might help.

beetsplug/parentwork.py
```diff
@@ -0,0 +1,197 @@
...
+class ParentWorkPlugin(BeetsPlugin):
+    def __init__(self):
+        super(ParentWorkPlugin, self).__init__()
+
+        self.config.add({
+            'auto': False,
+            'force': False,
+        })
+
+        self._command = ui.Subcommand(
```
sampsyo: It's probably more appropriate to declare the `ui.Subcommand` inside the `commands` method instead of making it a field. (It's only used there.)

beetsplug/parentwork.py
```diff
@@ -0,0 +1,197 @@
...
+        mb_parentworkid = work_info['work']['id']
+        if 'disambiguation' in work_info['work']:
+            parentwork_disambig = work_info['work']['disambiguation']
+            return [parentwork, mb_parentworkid, parent_composer,
+                    parent_composer_sort, parentwork_disambig]
+        else:
+            return [parentwork, mb_parentworkid, parent_composer,
+                    parent_composer_sort, None]
+
+    def find_work(self, item, force):
```
sampsyo: This could use a docstring describing which fields on `item` get updated.

beetsplug/parentwork.py
```diff
@@ -0,0 +1,197 @@
...
+
+        recording_id = item.mb_trackid
```
sampsyo: I think it would be OK (and a little clearer) to just use `item.mb_trackid` instead of putting it in a variable here.

beetsplug/parentwork.py
```diff
@@ -0,0 +1,197 @@
...
+        try:
+            item.parentwork
+            hasparent = True
+        except AttributeError:
+            hasparent = False
```
sampsyo: I recommend just using `hasattr(item, 'parentwork')`.

beetsplug/parentwork.py
```diff
@@ -0,0 +1,197 @@
...
+
+                found = True
+            except musicbrainzngs.musicbrainz.WebServiceError:
+                self._log.debug("Work unreachable")
+                found = False
+        elif hasparent:
+            self._log.debug("Work already in library, not necessary fetching")
+            return
+
+        if found:
```
sampsyo: I don't think you need `found` or this conditional—just put this right in the condition above after gathering the data. (Or perhaps more appropriately, after an "early return.")

beetsplug/parentwork.py
```diff
@@ -0,0 +1,197 @@
...
+                work_info, work_date = find_parentwork(item.mb_workid)
+                parent_info = self.get_info(item, work_info)
+                parentwork = parent_info[0]
+                mb_parentworkid = parent_info[1]
+                parent_composer = parent_info[2]
+                parent_composer_sort = parent_info[3]
+                parentwork_disambig = parent_info[4]
+
+                found = True
+            except musicbrainzngs.musicbrainz.WebServiceError:
```
sampsyo: A `try`/`except` should go only around the code that might raise the appropriate exception. Here, I think that's just `find_parentwork`?

beetsplug/parentwork.py
```diff
@@ -0,0 +1,197 @@
...
+        force_parent = self.config['force'].get(bool)
+
+        for item in task.imported_items():
+            self.find_work(item, force_parent)
+            item.store()
+
+    def get_info(self, item, work_info):
+        """Given the parentwork info dict, this function fetches
+        parent_composer, parent_composer_sort, parentwork,
+        parentwork_disambig, mb_workid and composer_ids"""
```
sampsyo: The return type here is a little awkward—perhaps consider returning a `dict` whose fields are exactly what needs to be assigned onto the item?

beetsplug/parentwork.py
```diff
@@ -0,0 +1,197 @@
...
+            item['parentwork'] = parentwork
+            if parentwork_disambig:
+                item['parentwork_disambig'] = parentwork_disambig
+            item['mb_parentworkid'] = mb_parentworkid
+            item['parent_composer'] = u''
+            item['parent_composer'] = u', '.join(parent_composer)
+            item['parent_composer_sort'] = u''
+            item['parent_composer_sort'] = u', '.join(parent_composer_sort)
+            if work_date:
+                item['work_date'] = work_date
```
sampsyo: Maybe there's a simpler way to do all these assignments—something along the lines of:

```py
for key, value in info.items():
    if value:
        item[key] = value
```

…especially if the return type of `get_info` is changed to a dict.

dosoe:
It's very fun, I'm correcting my students' informatics exams (basic python course) and I'm getting informatics teached at the same time. Thanks for the comments! I never got healthy programming practices teached (at least not on this level).

beetsplug/parentwork.py
dosoe: Strictly speaking, it is just find_father, but there I cannot use self._log.debug() as it is not in the ParentWorkPlugin class.

beetsplug/parentwork.py
dosoe: Thanks for the link! Yes, in the old PR I got requested to fetch the composition date, so I fetch it at each level until I find a date. It's not very satisfactory, I feel like this plugin is not really the place to do this, but it's possible.

beetsplug/parentwork.py
dosoe: I have no idea how to do this and how it works, I just copy-pasted it from another plugin. Isn't there a difference between ```_command``` and ```command```?

beetsplug/parentwork.py
dosoe: I don't like using 'ancestor', simply because 'ancestorwork' looks bad, I prefer 'parentwork'. If I clarify it in the docs, would it be fine?

beetsplug/parentwork.py
sampsyo: Nope, they're just different variable names. If you move this into the `commands()` function, you can just call it `command` instead of `self._command` (i.e., a local variable instead of a field).

beetsplug/parentwork.py
dosoe: Ok I clarified the docstring, this should be ready to merge now.

dosoe:
I don't understand. How can my PR change the test coverage of fetchart?

docs/plugins/parentwork.rst
```diff
@@ -0,0 +1,60 @@
...
+To fill in the parentwork tag and the associated parent** tags, in case there 
+are several works on the recording, it fills it with the results of the first 
+work and then appends the results of the second work only if they differ from 
+the ones already there. This is to care for cases of, for example, an opera 
+recording that contains several scenes of the opera: neither the parentwork 
+nor all the associated tags will be duplicated. 
+If there are several works linked to a recording, they all get a 
+disambiguation (empty as default) and if all disambiguations are empty, the 
+disambiguation field is left empty, else the disambiguation field can look 
+like ``,disambig,,`` (if there are four works and only the second has a
```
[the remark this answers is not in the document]
dosoe: yes you're right. It's fixed now.

sampsyo:
Arg, @codecov-io's output is pretty erratic and sometimes misleading. And we have configured it _not_ to post these pull request messages, but sometimes it just does it anyway, for some reason…

dosoe:
I have one more question: Works on MB change from time to time, and if the work changes, I would like the parentwork to be updated as well. A change in work happens only at a ```beet mbsync``` or maybe at a ```beet up``` if ```mb_workid``` gets implemented into ```mediafile```. Right now, it doesn't get updated, so the only solution is to run ```beet parentwork -f ``` again (with the ```--force``` option else it will only look for recordings with no parentwork) and this takes a huge time because it will recover all parentworks again (which is very slow). Could there be a way to apply ```beet parentwork --force``` only to items where the ```mb_workid``` got modified? Either by applying it only to recordings that got modified in the latest time (is there a way to list those? And I mean a modification in the database, not in the file itself like ```mtime```) or by a more in-depth harmonization between mbsync, up and parentwork.

dosoe:
One more thing, are there comments on the logging? It feels rather verbose right now, are you fine with that?

sampsyo:
Interesting! Here's one thing I'm uncertain about: which of these two scenarios are you most concerned about adapting to?

1. The user changes the track ID for an item. So you need to get a new work to match that.
2. The user does nothing, but the MusicBrainz database changes to associate a new work with an existing track. The user wants to synchronize these new changes with the file.

The thing is that "1" above seems less likely to be an issue (how often do users _really_ change the MBID for tracks they already have in their library?), but it's easier to deal with—there are a few ways we can consider to track when the MBID has gone out of sync with the work. "2" seems like much more of a real issue, but there's essentially nothing to be done about it—to know whether anything changed, you have to check with the MB server again, which essentially means doing `beet parentwork -f`. There is basically no way around this.

---

About logging: it's probably OK to be somewhat verbose in "debug" mode. But of course, brevity is the soul of wit. I'll take a look at the log messages now.

beetsplug/parentwork.py
```diff
@@ -0,0 +1,201 @@
...
+
+            parentwork_info['parent_composer'] = u', '.join(parent_composer)
+            parentwork_info['parent_composer_sort'] = u', '.join(
+                    parent_composer_sort)
+
+        if not composer_exists:
+            self._log.info(item.artist + ' - ' + item.title)
+            self._log.debug(
+                "no composer, add one at https://musicbrainz.org/work/" +
+                work_info['work']['id'])
```
sampsyo: It's odd that one of these lines goes to info and the other goes to debug. Here's a suggestion:

```py
self._log.debug(
    'no composer for {} - {}; add one at https://musicbrainz.org/work/{}',
    item.artist, item.title, work_info['work']['id'],
)
```

(Suitably wrapped, of course.)

beetsplug/parentwork.py
```diff
@@ -0,0 +1,201 @@
...
+        if hasattr(item, 'parentwork'):
+            hasparent = True
+        else:
+            hasparent = False
+        if not item.mb_workid:
+            self._log.info("No work attached, recording id: " +
+                           item.mb_trackid)
+            self._log.info(item.artist + ' - ' + item.title)
+            self._log.info("add one at https://musicbrainz.org" +
+                           "/recording/" + item.mb_trackid)
```
sampsyo: Collapsing this into one line, as above, might be nice. Especially since this goes to `info`.

beetsplug/parentwork.py
```diff
@@ -0,0 +1,201 @@
...
+            return
+        if force or (not hasparent):
+            try:
+                work_info, work_date = find_parentwork_info(item.mb_workid)
+            except musicbrainzngs.musicbrainz.WebServiceError:
+                self._log.debug("Work unreachable")
```
sampsyo: I suggest including some evidence about the error:

```py
except musicbrainzngs.musicbrainz.WebServiceError as e:
    self._log.debug("error fetching work: {}", e)
```

beetsplug/parentwork.py
```diff
@@ -0,0 +1,201 @@
...
+                return
+            parent_info = self.get_info(item, work_info)
+
+        elif hasparent:
+            self._log.debug("Work already in library, not necessary fetching")
```
sampsyo: Shorter wording: "work present; skipping". Including the artist/title here might be good too.

beetsplug/parentwork.py
```diff
@@ -0,0 +1,201 @@
...
+            return
+
+        self._log.debug("Finished searching work for: " +
+                        item.artist + ' - ' + item.title)
```
sampsyo: I recommend deleting this—it seems redundant with all the other messages that explain what happened with a given item.

beetsplug/parentwork.py
```diff
@@ -0,0 +1,201 @@
...
+        if hasattr(item, 'parentwork'):
```
sampsyo: It's not clear to me why this is the right condition to switch between these two messages. They both seem to print out the parent work.

arcresu:
Regarding the issue of updating the parent work when things change, this is an instance of the more general issue with metadata fetching covered in #266. We're planning to change how plugins work behind the scenes, and hopefully as part of that process we can solve that issue in a more satisfying way. That's a bigger issue that will take some time to get right though.

For example we could model `mbsync` as a cache of remote metadata, so it would be possible to sync just the least-recently updated items rather than the entire library. Derived metadata like the parent work could be "invalidated" if the dependent metadata (work ID) changes, triggering an immediate recalculation. This is just brainstorming rather than a concrete proposal.

sampsyo:
Another thought occurred to me about how to simplify the logging: you don't need to do this:

```py
self._log.info('processing {} - {}', item.artist, item.title)
```

Instead, you can just do this:

```py
self._log.info('processing {}', item)
```

and the `item` will get formatted in a sensible way automatically.

test/test_parentwork.py
```diff
@@ -0,0 +1,93 @@
...
+        command_output.return_value = u'32c8943f-1b27-3a23-8660-4567f4847c94'
+        self.run_command('parentwork')
+
+        item.load()
+        self.assertEqual(item['mb_parentworkid'], u'XXX')
+
+    # test different cases, still with Matthew Passion Ouverture or Mozart
+    # requiem
+
+    def test_father_work(self, command_output):
```
arcresu: Looks like you missed this test when changing the field name.
```suggestion
    def test_direct_parent_work(self, command_output):
```

beetsplug/parentwork.py
```diff
@@ -0,0 +1,201 @@
...
+            if artist['type'] == 'composer':
+                if 'end' in artist.keys():
+                    work_date = artist['end']
+
+    if 'work-relation-list' in work_info['work']:
+        for work_father in work_info['work']['work-relation-list']:
+            if work_father['type'] == 'parts' \
+                    and work_father.get('direction') == 'backward':
+                father_id = work_father['work']['id']
+                return father_id, work_date
```
arcresu: There are still some references to `father` here too.

docs/plugins/parentwork.rst
```diff
@@ -0,0 +1,46 @@
+Parentwork Plugin
+=================
+
+The ``parentwork`` plugin fetches the work title, parentwork title and 
+parentwork composer. 
+
+In the MusicBrainz database, a recording can be associated with a work. A 
+work can itself be associated with another work, for example one being part 
+of the other (what I call the father work). This plugin looks the work id 
+from the library and then looks up the father, then the father of the father
```
arcresu:
```suggestion
from the library and then looks up the direct parent, then _its_ direct parent
```

beetsplug/parentwork.py
```diff
@@ -0,0 +1,201 @@
...
+
+from beets import ui
+from beets.plugins import BeetsPlugin
+
+import musicbrainzngs
+
+
+def direct_parent_id(mb_workid, work_date=None):
+    """ Given a mb_workid, find the id one of the works the work is part of
+    and the first composition date it encounters."""
```
arcresu: For consistency with the beets coding style the final `"""` should go on its own line. There are a few places where this applies.

docs/plugins/parentwork.rst
```diff
@@ -0,0 +1,46 @@
+Parentwork Plugin
+=================
+
+The ``parentwork`` plugin fetches the work title, parentwork title and 
+parentwork composer.
```
arcresu: I know the plugin is spelled `parentwork`, but maybe we can use "parent work" when discussing the concept?

docs/plugins/parentwork.rst
```diff
@@ -0,0 +1,46 @@
...
+be, in this case, the whole symphony. 
+
+This plugin adds five tags: 
+
+- **parentwork**: The title of the parentwork.  
+- **mb_parentworkid**: The musicbrainz id of the parentwork. 
+- **parentwork_disambig**: The disambiguation of the parentwork title. 
+- **parent_composer**: The composer of the parentwork. 
+- **parent_composer_sort**: The sort name of the parentwork composer. 
+- **work_date**: THe composition date of the work, or the first parent work
```
arcresu:
```suggestion
- **work_date**: The composition date of the work, or the first parent work 
```

docs/plugins/parentwork.rst
```diff
@@ -0,0 +1,46 @@
+Parentwork Plugin
+=================
+
+The ``parentwork`` plugin fetches the work title, parentwork title and 
+parentwork composer. 
+
+In the MusicBrainz database, a recording can be associated with a work. A 
+work can itself be associated with another work, for example one being part 
+of the other (what I call the father work). This plugin looks the work id
```
arcresu:
```suggestion
of the other (what I call the direct parent work). This plugin looks the work ID 
```

dosoe:
Thanks for the detailed comments! I think I adressed them all. 
Concerning the metadata change, what I would like to have is that ```parentwork``` notices when ```mbsync``` (or anyone else) modifies the ```mb_workid``` tag and kicks in to update the parentwork tag (and all associated tags). 
While the ```mb_trackid``` almost never changes once in the library, the ```mb_workid``` can change (and I notice it changing from time to time), so what I'm looking for is a solution to 1. Is there already something implemented that could make it easier (maybe using ```register_listener``` from the ```BeetsPlugin``` class)?

dosoe:
I'm trying to read through https://beets.readthedocs.io/en/v1.4.9/dev/plugins.html#listen-for-events but I don't fully get it: I tried something like 
```
class ParentWorkPlugin(BeetsPlugin):
    def __init__(self):
        ...
        if self.config['auto']:
            self.register_listener('trackinfo_received', self.find_work2)
    def find_work2(self, item):
        self.find_work(self, item, True)
```
This should call ```find_work``` each time we get data from MB. However, maybe the event that I should listen to is ```database_change``` and not ```trackinfo_received``` , as the second one would trigger even if there is no change. For the first one, the parameters are ```lib``` and ```model``` and I don't know how to get an ```item``` out of them to feed to ```find_work``` . 
Additionally, as coded right now (with ```trackinfo_received```) a ```beet mbsync``` fails: 
```
  File "/usr/local/bin/beet", line 11, in <module>
  File "/usr/local/lib/python2.7/dist-packages/beets/ui/__init__.py", line 1266, in main
    _raw_main(args)
  File "/usr/local/lib/python2.7/dist-packages/beets/ui/__init__.py", line 1253, in _raw_main
  File "/usr/local/lib/python2.7/dist-packages/beetsplug/mbsync.py", line 75, in func
    self.singletons(lib, query, move, pretend, write)
  File "/usr/local/lib/python2.7/dist-packages/beetsplug/mbsync.py", line 96, in singletons
    track_info = hooks.track_for_mbid(item.mb_trackid)
  File "/usr/local/lib/python2.7/dist-packages/beets/autotag/hooks.py", line 574, in track_for_mbid
    plugins.send(u'trackinfo_received', info=track)
  File "/usr/local/lib/python2.7/dist-packages/beets/plugins.py", line 489, in send
    result = handler(**arguments)
  File "/usr/local/lib/python2.7/dist-packages/beets/plugins.py", line 149, in wrapper
    return func(*args, **kwargs)
TypeError: find_work2() takes exactly 2 arguments (1 given)
```

I'm puzzled, because I wrote ```find_work2``` specifically to have only one argument.

dosoe:
Hi! I found my error. I changed it to: 
```
class ParentWorkPlugin(BeetsPlugin):
    def __init__(self):
        ...
        if self.config['auto']:
            self.register_listener('database_change', self.find_work2)
    def find_work2(self, lib, model):
        self.find_work(model, True)
```

This however does not work entirely fine. A try with mbsync gives: 
```
Requesting MusicBrainz release 05909b27-18c5-47dc-bcd8-971a7f25ab79
primary MB release type: album
mbsync: applying changes to Johann Sebastian Bach; Richard Galliano - Galliano | Bach
Johann Sebastian Bach - Galliano | Bach - Art of Fugue, BWV 1080: Contrapunctus 1
  composer_sort:  -> Bach, Johann Sebastian
  work:  -> Die Kunst der Fuge, BWV 1080: I. Contrapunctus 1
  mb_workid:  -> 621651bc-7282-47ad-a401-d8be5a0988bd
  work_disambig:  -> catch all for arrangements
parentwork: no composer for Johann Sebastian Bach - Galliano | Bach - Art of Fugue, BWV 1080: Contrapunctus 1; add one at https://musicbrainz.org/work/621651bc-7282-47ad-a401-d8be5a0988bd
parentwork: Work fetched: Die Kunst der Fuge, BWV 1080: I. Contrapunctus 1 - Johann Sebastian Bach
Johann Sebastian Bach - Galliano | Bach - Art of Fugue, BWV 1080: Contrapunctus 1
  parentwork: Die Kunst der Fuge, BWV 1080: I. Contrapunctus 1
  parent_composer_sort: Bach, Johann Sebastian
  parentwork_disambig: catch all for arrangements
  parent_composer: Johann Sebastian Bach
  mb_parentworkid: 621651bc-7282-47ad-a401-d8be5a0988bd
  File "/usr/local/bin/beet", line 11, in <module>
  File "/usr/local/lib/python2.7/dist-packages/beets/ui/__init__.py", line 1266, in main
    _raw_main(args)
  File "/usr/local/lib/python2.7/dist-packages/beets/ui/__init__.py", line 1253, in _raw_main
  File "/usr/local/lib/python2.7/dist-packages/beetsplug/mbsync.py", line 76, in func
    self.albums(lib, query, move, pretend, write)
  File "/usr/local/lib/python2.7/dist-packages/beetsplug/mbsync.py", line 178, in albums
    apply_item_changes(lib, item, move, pretend, write)
  File "/usr/local/lib/python2.7/dist-packages/beetsplug/mbsync.py", line 40, in apply_item_changes
    item.store()
  File "/usr/local/lib/python2.7/dist-packages/beets/library.py", line 348, in store
    plugins.send('database_change', lib=self._db, model=self)
  File "/usr/local/lib/python2.7/dist-packages/beets/plugins.py", line 489, in send
    result = handler(**arguments)
  File "/usr/local/lib/python2.7/dist-packages/beets/plugins.py", line 143, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/beetsplug/parentwork.py", line 90, in find_work2
    self.find_work(model, True)
  File "/usr/local/lib/python2.7/dist-packages/beetsplug/parentwork.py", line 208, in find_work
    item.store()
  File "/usr/local/lib/python2.7/dist-packages/beets/library.py", line 348, in store
    plugins.send('database_change', lib=self._db, model=self)
  File "/usr/local/lib/python2.7/dist-packages/beets/plugins.py", line 489, in send
    result = handler(**arguments)
  File "/usr/local/lib/python2.7/dist-packages/beets/plugins.py", line 137, in wrapper
    assert self._log.level == logging.NOTSET
AssertionError
```
Do you know what the error is?

sampsyo:
Hi! Can we please leave this "automation" feature to a separate pull request? It would be awesome to close the book on the basic version before we start adding fancier features.

I was even going to suggest a simpler approach to this that might be a little less surprising to users. It seems little surprising that simple metadata changes with `beet modify`, which is typically a "local only" operation, could cause network requests and stuff. Instead, maybe the plugin could keep around a flexible attribute that says "here's the work ID that the current parent work is based off of." Then, when running for a second time, check whether that still matches the current work ID and re-fetch if so. This way, the heavy lifting only happens when users ask for it rather than truly automatically.

I don't exactly see the source of that error, but it might be caused by the presence of other plugins…

dosoe:
Ok, so I added a tag ```mb_workid_current``` which is there for this purpose, do you thing the name is appropriate? 
Also, I think that the plugin calling itself can be a real problem.

sampsyo:
Seems reasonable to me! But can we leave this enhancement to a separate pull request? (You could even base the new branch on this branch and create the PR now.)

dosoe:
Ok then this should be it!

beetsplug/parentwork.py
```diff
@@ -0,0 +1,200 @@
...
+                if write:
+                    item.try_write()
+        command = ui.Subcommand(
+            'parentwork',
+            help=u'Fetches parent works, composers and dates')
+
+        command.parser.add_option(
+            u'-f', u'--force', dest='force',
+            action='store_true', default=None,
+            help=u'Re-fetches all parent works')
```
sampsyo: Perhaps more specifically:

> re-fetch when parent work is already present

beetsplug/parentwork.py
```diff
@@ -0,0 +1,200 @@
...
+                item[key] = value
+
+        if work_date:
+            item['work_date'] = work_date
+        ui.show_model_changes(
+            item, fields=['parentwork', 'parentwork_disambig',
+                          'mb_parentworkid', 'parent_composer',
+                          'parent_composer_sort', 'work_date'])
+
+        item.store()
```
sampsyo: Delete this call to `item.store()`; the item is already stored in both places where this function is called.

dosoe: this might actually solve the problem I had with catching database changes, as the error was on the ```item.store()``` . But it's for another time I guess.

sampsyo:
Merged! Thank you for your long-term persistence in seeing this through!

PR #3282, opened 2019-06-01
Update Python 3 versions for CI

arcresu:
I noticed some deprecation messages in the Travis outout when it's running `flake8` using Python 3.4. Let's update that to a more modern version. There's probably more that can be updated in the CI config while we're at it.


arcresu:
Looks like just a flaky AppVeyor job unrelated to the changes here. I'll go ahead and merge.

sampsyo:
Great idea; thanks!

PR #3294, opened 2019-06-04
Add repology badge

arcresu:
This is just to advertise that beets is available in distros. The badge links to a list of distro packages for beets and the current versions they have available, which is useful for users and contributors.


arcresu:
Done! I was tossing up between putting it in the relevant section or up the top in the first place so I'm happy to move it.

Issue #3283, opened 2019-06-01
Audit the wiki pages


arcresu:
Some of our wiki pages are outdated. User manual or developer guide type information should move into the sphinx docs for better visibility and so that it's versioned with the code. I think it makes sense to use the wiki for longer-term planning or planning of big changes, with the target audience being beets contributors. Where possible smaller plans can go to issues on beets with appropriate labels, and pages planning many separate changes can be broken up. That would minimise duplication of information and keep discussions together. GitHub milestones can also be used like epics to group sub-issues.

Here are some thoughts so we can discuss:

. | Wiki page | Type | Proposed change
--|------------- | -------|----------------------
 :thought_balloon:  |  [Architecture](https://github.com/beetbox/beets/wiki/Architecture) | Dev reference |  Move to sphinx
:heavy_check_mark: | [Attachment Proposal](https://github.com/beetbox/beets/wiki/Attachment-Proposal) | Planning | None
:heavy_check_mark:  | [Attachments](https://github.com/beetbox/beets/wiki/Attachments) | Planning | None
:thought_balloon: | [Beets Web UI](https://github.com/beetbox/beets/wiki/Beets-Web-UI) | Planning | This is outdated. Move any unimplemented features to issues.
:thought_balloon: | [Concepts](https://github.com/beetbox/beets/wiki/Concepts) | User documentation | Move to sphinx #2277
:thought_balloon: | [Cookbook](https://github.com/beetbox/beets/wiki/Cookbook) | User documentation | Move to sphinx
:thought_balloon: | [Database structure](https://github.com/beetbox/beets/wiki/Database-structure) | Dev reference | Probably outdated. Delete or move to sphinx.
:thought_balloon: | [Hacking](https://github.com/beetbox/beets/wiki/Hacking) | Developer reference | Move to a new CONTRIBUTING file #2103
:thought_balloon: | [Issues](https://github.com/beetbox/beets/wiki/Issues) | Maintainer reference | We can probably delete this?
:thought_balloon: | [MacOSX](https://github.com/beetbox/beets/wiki/MacOSX) | User documentation | Probably not relevant any more? If it is then move to sphinx.
:thought_balloon: | [MediaFile](https://github.com/beetbox/beets/wiki/MediaFile) | Developer reference | Delete after confirming migration to beetbox/mediafile sphinx docs.
:heavy_check_mark:  | [Optimization](https://github.com/beetbox/beets/wiki/Optimization) | Planning | None
[...]

PR #3295, opened 2019-06-04
Expand library API docs

arcresu:
As part of #3283 I'd like to migrate the [Architecture](https://github.com/beetbox/beets/wiki/Architecture) wiki page into beets' built-in documentation. There was a lot of overlap with the existing dev/API page, so I've used the relevant text from the wiki page to supplement that existing documentation page.

This leave these sections from the wiki unaccounted for:

* Queries
* Modification times
* The importer
* Command line interface

Queries are related to the database API but maybe it'd be clearer to add a separate dev page and cross-reference it with the existing user manual page on queries.  The other additional sections could also go on their own pages but I'm not quite sure how to split it up yet.


docs/dev/api.rst
```diff
@@ -25,60 +40,178 @@ The Library Class
...
-.. autoclass:: Transaction
-    :members:
 
 Model Classes
 -------------
 
 The two model entities in beets libraries, :class:`Item` and :class:`Album`,
-share a base class, :class:`Model`, that provides common functionality and
+share a base class, :class:`LibModel`, that provides common functionality and
 ORM-like abstraction.
```
sampsyo: This is a detail that might not matter, but `dbcore.db.Model` provides that ORM-like abstraction. `LibModel` provides shared functionality between `Album` and `Item` that is more music-specific.

docs/dev/api.rst
```diff
@@ -25,60 +40,178 @@ The Library Class
...
 
-The fields model classes can be accessed using attributes (dots, as in
-``item.artist``) or items (brackets, as in ``item['artist']``). The
-:class:`Model` base class provides some methods that resemble `dict`
-objects.
```
sampsyo: I actually think this is nice to mention here—and perhaps not below, in the two places where we say something similar about the `Item` and `Album` classes. They really do work the same way, and it would be nice to emphasize that the main way to interact with beets data is by accessing these attributes.

docs/dev/api.rst
arcresu: I was trying to simplify the description but I agree this was imprecise.

docs/dev/api.rst
arcresu: Agreed. I've moved it back to the top and edited the wording slightly to emphasise that the behaviour is generic to models.

arcresu:
I've now incorporated the remaining text from the wiki page pretty much exactly how you suggested. The CLI and importer text didn't seem outdated, but could do with elaboration in future.

sampsyo:
Awesome work!! Thank you for putting this together!

PR #3296, opened 2019-06-04
Update a lot of URLs to use HTTPS

FichteFoll:
*All* URLs were checked manually, but only once per domain!

I mostly concerned myself with URLs in documentation rather than source
code because the latter may or may not have impactful changes, while the
former should be straight forward.

Changes in addition to simply adding an s:

- changed pip and pypi references as their location has changed
- MPoD (iOS app) url redirects to Regelian, so I replaced those
- updated homebrew references

Notable observations:

- beets.io does have HTTPS set up properly (via gh-pages)
- beatport.py uses the old HTTP url for beatport
- as does lyrics.py for lyrics.wikia.com
- https://tomahawk-player.org/ expired long ago, but the http page
  redirects to https regardless
- none of the sourceforge subdomains have https (in 2019!)

Regarding beets.io, I believe you will just need to select some checkbox near the CNAME input in the repo's settings.


arcresu:
Thanks for doing this! I'm happy to merge it once the CI passes (just to be sure that Travis' version of apt can cope with the HTTPS transport) :tada:

sampsyo:
This is truly astounding work. Thank you for doing it, @FichteFoll!

PR #3297, opened 2019-06-05
More https changes

FichteFoll:
Continuation of #3296. See commit messages for details.

This time I also changed/added a couple references in code.

Regarding beets.io, you'll probably need to update your [gh-pages settings](https://github.com/beetbox/beets/settings) to remove the custom domain name and then set it again so that Github will create a proper certificate for your domain. They support this for about a year now.

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -475,7 +475,7 @@ def human_seconds_short(interval):
 # Colorization.
 
 # ANSI terminal colorization code heavily inspired by pygments:
-# http://dev.pocoo.org/hg/pygments-main/file/b2deea5b5030/pygments/console.py
+# https://bitbucket.org/birkenfeld/pygments-main/src/default/pygments/console.py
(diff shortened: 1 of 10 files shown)


sampsyo:
Cool; thanks! Thanks especially much for tracking down the new URLs for source repositories that have moved around.

I've done the thing where I remove the custom domain from the Pages settings and re-add it. (I've also updated the DNS entries for beets.io to point to the new CDN servers.) The certificate still doesn't seem to be valid, but I'd be willing to believe this could take a while to set up, so let's see what happens.

FichteFoll:
Yes, it may take little while to be updated. I suppose if it doesn't work within a day or so, you could contact support.

FichteFoll:
Once that works, I can update all the references to `http://beets.io` as well.

FichteFoll:
Actually, it works already for me, so 🎉

sampsyo:
Oh yeah! It just started working for me too.

Issue #1935, opened 2016-04-12
Using beets for Video Library Organization?


addisonamiri:
I was wondering if there was any interest in adding video library support to beets. I really like the workflow of importing my media and playing it with `beet play` without the overhead of a full blown media player. I was curious what would be necessary for beets to implement video support. So far what I think would be required at a minimum is this:
- id3tag support for video formats
- new data sources for TV shows and movies
- a config option for a video directory

I don't really think this is in scope for the beets project and a lot of it won't carry over into video organizing but I can't seem to find a media organizer for videos that doesn't require a server or a gui (Plex, Kodi, Emby, etc) and most of those projects require manual editing of filenames in order for the lookup to succeed.

I was wondering what everyone thought about this functionality being in beets or another project similar to beets. I know I'd find it useful but I'm not too sure if it's worthy of being incorporated into the main project.
#### Desired Workflow

Ideally this would be the workflow I'm looking to achieve:
1.  Run `beet import The\ Princess\ Bride.mp4`.
   - Maybe a `--as-video` flag would be needed for an alternate import method.
   - Automatic movie detection could be done by looking at the length of the file maybe.
2.  Beets will search IMDB (or another metadata source) based on the name.
3.  Beets will identify the movie as Princess and the Bride.
4.  Beets will move/copy the file to the video directory specified in the config.
5.  Beets will rename the file, add all metadata desired, and finish the import.

Then after this is complete `beet play Princess Bride` would start playing that file with the configured media player.

amishb:
Filebot + ACM with some config is what you need

https://www.filebot.net/forums/viewtopic.php?t=215

sampsyo:
Thanks for starting the discussion! I've always thought expanding to cover other media types would be an interesting direction.

One major roadblock is that many video file types, last I checked, don't have standard metadata fields. And in any case, it's not common to find tagged video files even with non-standard formats, so we'd need to rely much more on filename heuristics. See also #1160.

I can see this going one of two ways:
- Adding "media kinds" to beets, so songs and videos coexist in the Items table with a flag to distinguish them.
- Start a sister project that uses much of beets' internals but focuses exclusively on videos. Much of the machinery is already reusable—namely, the `dbcore` module contains the "soul" of beets without knowing anything about music.

jackwilsdon:
Your second point is interesting, @sampsyo. Could we perhaps move `dbcore` into it's own repository and pip package?

sampsyo:
[...]

PR #3300, opened 2019-06-07
implement bulk_add for applying several SQLite operations per transaction

khimaros:
Initially inspired by #1935 and seemingly relevant to #2388

Based on my `snakeviz` benchmarks in https://github.com/beetbox/beets/issues/1935#issuecomment-433630128 and https://github.com/beetbox/beets/issues/1935#issuecomment-434074227, `dbcore` overhead for an import operation on 847 local files was reduced from **95% (153 seconds)** to **<10% (5 seconds)** .

I have exercised this patch with a dbcore frontend I built to organize, query, and de-duplicate arbitrary (including non-music) files. In order to realize these performance benefits in beets, library import will need to be modified to make use this method.

N.B. The degree of batching chosen for `bulk_add` is a trade-off between buffering delay, CPU time, and peak memory usage. I've found negligible memory usage for single transactions containing tens of thousands of model records.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -480,7 +480,7 @@ class Model(object):
 
     # Database interaction (CRUD methods).
 
-    def store(self, fields=None):
+    def store(self, fields=None, txn=None):
         """Save the object's metadata into the library database.
         :param fields: the fields to be stored. If not specified, all fields
         will be.
@@ -500,7 +500,7 @@ class Model(object):
                 subvars.append(value)
         assignments = ','.join(assignments)
 
-        with self._db.transaction() as tx:
+        with self._db.transaction(txn) as tx:
             # Main table update.
             if assignments:
                 query = 'UPDATE {0} SET {1} WHERE id=?'.format(
@@ -541,11 +541,11 @@ class Model(object):
         self.update(dict(stored_obj))
         self.clear_dirty()
 
-    def remove(self):
+    def remove(self, txn=None):
         """Remove the object's associated rows from the database.
         """
         self._check_db()
-        with self._db.transaction() as tx:
+        with self._db.transaction(txn) as tx:
             tx.mutate(
                 'DELETE FROM {0} WHERE id=?'.format(self._table),
                 (self.id,)
@@ -555,7 +555,7 @@ class Model(object):
                 (self.id,)
             )
 
-    def add(self, db=None):
+    def add(self, db=None, txn=None):
         """Add the object to the library database. This object must be
         associated with a database; you can provide one via the `db`
         parameter or use the currently associated database.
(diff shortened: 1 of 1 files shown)


sampsyo:
Cool! Thanks for looking into this!!

Here's one thing I don't quite understand about this profiling work: does the overhead come from literally creating SQLite transactions, or is it our in-Python cost of looking up the transactions? We built the transaction system in dbcore to avoid the creation of redundant transactions. So if you do this:

```py
with db.transaction():
    with db.transaction():
        db.mutate(...)
```

…only one SQLite transaction should actually be created. More relevantly, if you do this:

```py
with db.transaction():
    for item in items:
        db.add(...)
```

…then the transaction created inside the `add` call should approximately be a no-op. Of course, lookup up the transaction stack for the current thread is not free! So it's possible the overhead is coming from there.

But just to confirm: did you measure the performance difference between an unwrapped `add` loop, a transaction-wrapped `add` loop, and the new `bulk_add` that doesn't bother creating a transaction at all? The relative differences between these differences would be enlightening…

khimaros:
@sampsyo -- indeed, performance seems to be nearly identical with the method you describe here!

I sat on this patch for a very long time and it seems the landscape may have changed. For a bit of background, when we spoke about this last October the conclusion was:

> Yes, absolutely! A bulk insert would be a great way to do it. You could even imagine letting the bulk_add method accept an iterable (instead of just a list) to let new model objects get generated on the fly without materializing them all in memory.

It's possible that nested transactions were not handled gracefully back then.

If this PR doesn't add any value, let's just close it and I will switch my dbcore clients to the nested transaction solution. I'll leave the final judgment to you. If it does add value, it should have some tests. I have some staged in my local repo for the performance testing.

Thank you for continuing to build and maintain this excellent tool!

sampsyo:
Yeah! I read that comment of mine too with great interest. It's constantly surprising to me how hard it is for me to reconstruct my own thinking from a previous era. Reading my comments, it looks like I may have been imagining that the overhead came from SQLite execute calls, and that such a bulk insertion would use one big statement? Who can really say what I was thinking, I guess.

If things are just as fast with the explicit transaction approach, I say we probably don't need `bulk_add`. But it might be worth documenting this phenomenon somewhere, especially if other people are going to be as intrepid as you and write their own dbcore clients!

khimaros:
There is definitely no performance reason to favor `bulk_add` over nested transaction:

```
$ nosetests -v --with-timer test.test_dbcore_perf
test_dbcore_bulk_add (test.test_dbcore_perf.PerformanceTest) ... ok (34.5063s)
test_dbcore_single_transaction (test.test_dbcore_perf.PerformanceTest) ... ok (33.9157s)

[success] 50.43% test.test_dbcore_perf.PerformanceTest.test_dbcore_bulk_add: 34.5063s
[success] 49.57% test.test_dbcore_perf.PerformanceTest.test_dbcore_single_transaction: 33.9157s
Ran 2 tests in 68.423s
```

Closing this PR now, thanks for the feedback!

Issue #3303, opened 2019-06-07
Synology NAS: cannot concatenate 'str' and 'int' objects using fetchart plugin with Fanart.tv


musiczetetic:
### Problem

Running the import command gives an error on only some albums:

TypeError: cannot concatenate 'str' and 'int' objects

Then import stops

This only started after I updated beets, I cannot remember the version I was on before, it was about 2 years old.

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv import -t /volume1/Audio/Organising/XXXXXnewFORdukereplace/
```

Led to this problem:

```
user configuration: /root/.config/beets/config.yaml
data directory: /root/.config/beets
plugin paths:
Sending event: pluginload
library database: /volume1/Documents/BeetsDB/DukeboxBeeten.blb
library directory: /volume1/Audio/BeetifiedMusic/BeetenDiscoggedReplacements
Sending event: library_opened
Sending event: import_begin
Sending event: import_task_created
Sending event: import_task_start
Looking up: /volume1/Audio/Organising/XXXXXnewFORdukereplace/2011 - All Over Your Face
Tagging Garfunkel and Oates - All Over Your Face
Searching for discovered album ID: 7504a4b7-cf20-38fa-aad6-69173ad89601
Requesting MusicBrainz release 7504a4b7-cf20-38fa-aad6-69173ad89601
primary MB release type: album
Sending event: albuminfo_received
Candidate: Garfunkel and Oates - All Over Your Face (7504a4b7-cf20-38fa-aad6-69173ad89601)
Computing track assignment...
...done.
Success. Distance: 0.00
Album ID match recommendation is Recommendation.strong
Search terms: Garfunkel and Oates - All Over Your Face
Album might be VA: False
Searching for MusicBrainz releases with: {'release': u'all over your face', 'tracks': u'10', 'artist': u'garfunkel and oates'}
Requesting MusicBrainz release 7504a4b7-cf20-38fa-aad6-69173ad89601
Sending event: import_task_created
primary MB release type: album
Sending event: albuminfo_received
Candidate: Garfunkel and Oates - All Over Your Face (7504a4b7-cf20-38fa-aad6-69173ad89601)
Duplicate.
Sending event: import_task_created
Sending event: import_task_created
discogs: Searching for master release 511640
Sending event: import_task_created
Sending event: import_task_created
discogs: Searching for master release 511640
Sending event: import_task_created
Sending event: import_task_created
Sending event: import_task_created
[...]

PR #3304, opened 2019-06-07
Remove TypeError from string + int concat

GuilhermeHideki:
This removes the problem of concatenating string and int (#3303) but not the real problem (Discogs ids on mb* fields, I guess)

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -420,7 +420,7 @@ class FanartTV(RemoteArtSource):
 
         try:
             response = self.request(
-                self.API_ALBUMS + album.mb_releasegroupid,
+                self.API_ALBUMS + str(album.mb_releasegroupid),
                 headers={'api-key': self.PROJECT_KEY,
                          'client-key': self.client_key})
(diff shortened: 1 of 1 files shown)


sampsyo:
Cool; thanks! Two suggestions:

- It looks like a test failed because it's still trying to use concatenation. Maybe just using `self.API_ALBUMS + str(...)` would be simpler?
- I'm somewhat stumped by the fact that this field contains a non-string on your system! We have this declared in `beets.library` as a string-typed field: https://github.com/beetbox/beets/blob/e64ecc2f81b1f06dae29c6e09c84daf446c62b81/beets/library.py#L479
   so it should _always_ produce a string, even when someone erroneously assigns an integer into it. Any chance you could provide a copy of your library, or some fragment, that reproduces this?

GuilhermeHideki:
I found funny that the test were passing on my machine. Using `str()` is the best option because KISS and doesn't break the `API_ALBUMS` usage and we don't need to refactor tests. I used because was the same from other source (CAA).

PR #3307, opened 2019-06-11
docs: mention mpdstats needs to be running

arcresu:
Clarify that `mpdstats` is an MPD client and needs to be running all the time to collect statistics.

See https://discourse.beets.io/t/mpdstats-requirements/796


sampsyo:
Looks great; thanks!!

PR #3318, opened 2019-06-23
Absubmit improvements

SusannaMaria:
just wanted to have the possibility to skip items which have already Acousticbrainz data (which is now default). 
-force will go back to the old behavior
-dry will show only the items which would be processed

see
https://discourse.beets.io/t/acousticbrainz-for-big-libraries/378

### DIFF
--- a/beetsplug/absubmit.py
+++ b/beetsplug/absubmit.py
@@ -160,7 +160,7 @@ class AcousticBrainzSubmitPlugin(plugins.BeetsPlugin):
                     item=item, error=e
                 )
                 return None
-            with open(filename, 'rb') as tmp_file:
+            with open(filename, 'r') as tmp_file:
                 analysis = json.load(tmp_file)
             # Add the hash to the output.
             analysis['metadata']['version']['essentia_build_sha'] = \
--- a/beetsplug/absubmit.py
+++ b/beetsplug/absubmit.py
@@ -58,8 +58,10 @@ class AcousticBrainzSubmitPlugin(plugins.BeetsPlugin):
         self.config.add({
             'extractor': u'',
             'force': False,
-            'dry': False
+            'pretend': False
         })
+
+        self.PROBE_FIELD = 'mood_acoustic'
 
         self.extractor = self.config['extractor'].as_str()
         if self.extractor:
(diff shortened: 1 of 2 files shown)


sampsyo:
Looks great!! FWIW, we have used `--pretend` in the past as the name for this option instead of `--dry`. Would you be OK with using that name?

docs/plugins/absubmit.rst
```diff
@@ -34,12 +44,20 @@ will also skip music that the analysis tool doesn't support.
...
+- **auto**: Analyze every file on import. Otherwise, you need to use the
+  ``beet absubmit`` command explicitly.
   Default: ``no``
 - **extractor**: The absolute path to the `streaming_extractor_music`_ binary.
   Default: search for the program in your ``$PATH``
+- **force**: Analyze items and submit of AcousticBrainz data even for tracks
+  that already have it.
+  Default: ``no``.
+- **dry**: No analyze and submit of AcousticBrainz data but print out the
+  items which would be processed
```
sampsyo: “No” -> “Do not”. Also, missing a period here.

beetsplug/absubmit.py
```diff
@@ -113,12 +128,22 @@ def analyze_submit(self, item):
 
     def _get_analysis(self, item):
         mbid = item['mb_trackid']
+
         # If file has no mbid skip it.
+        if not self.opts.force_refetch and not self.config['force']:
+            mood_str = item.get('mood_acoustic', u'')
+            if mood_str:
+                return None
```
sampsyo: This simpler version would also suffice:

```
if not item.get('mood_acoustic'):
    return None
```

Maybe it would also make sense to move this "probe field" name to a module-level constant?

SusannaMaria:
> Looks great!! FWIW, we have used `--pretend` in the past as the name for this option instead of `--dry`. Would you be OK with using that name?
okay for sure

docs/plugins/absubmit.rst
SusannaMaria: Okay thx for the fast review. Your project is fun to work with.

sampsyo:
Awesome; thank you!! Merged with a few small tweaks. :sparkles:

PR #3319, opened 2019-06-28
Convert plugin skips previously-converted files

justinmayer:
One of the Convert plugin's greatest features is that it skips files that have already been converted. So one can just run `beet convert` willy-nilly without keeping track of which files have already been converted.

This adds one sentence to clarify the documentation, making it clear that one can run `beet convert` without fearing that previously-converted files will be needlessly re-converted.


sampsyo:
Looks wonderful; thank you for the addition! :sparkles:

justinmayer:
My pleasure, Adrian. Thank you for all the effort, time, and dedication you put into maintaining such a wonderful project! 💟

PR #3321, opened 2019-06-30
Fixes import problem when a leading slash is used in path names

mglukhovsky:
### Summary and resolution

This PR fixes a bug where path names that start with a leading slash will lead beets to try to copy/move files to the root directory on import.

This PR strips the leading path separator if it exists at the start of the subpath.

### Problem

If you use a path format like the following:
```
paths:
  default: $grouping/$artist - $title
  ...
```
...and the tag `$grouping` isn't present in the file, the subpath the importer sees will be just `/$artist - $title`. The subpath in this case starts with a leading slash.

When beets determines where to move the file (in the [`destination` function](https://github.com/beetbox/beets/blob/master/beets/library.py#L913), `os.path.join` uses two arguments: the library directory and the subpath for the specific file.

The second argument can't start with a leading slash, or `os.path.join()` will simply ignore the first argument, [as it's an absolute path](https://docs.python.org/3/library/os.path.html#os.path.join):
```
os.path.join('/library', '/Radiohead - Everything in Its Right Place.flac')`
# will return '/Radiohead - Everything in Its Right Place.flac`
```
This means the importer tries to place it in the root directory (`/`).

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -888,6 +888,9 @@ class Item(LibModel):
                 beets.config['path_sep_replace'].as_str()
             )
 
+        # If present, remove the leading slash to ensure this isn't an absolute path
+        subpath = subpath.lstrip(os.path.sep)
+
         maxlen = beets.config['max_filename_length'].get(int)
         if not maxlen:
             # When zero, try to determine from filesystem.


jackwilsdon:
This seems like you could end up with a confusing mixture of paths, with some containing a grouping and some not containing a grouping. It looks like you could instead do something like this to put things in an "Unknown" grouping:

```YAML
paths:
  default: %if{$grouping,$grouping,Unknown}/$artist - $title
```

sampsyo:
Hmm… could you please furnish us with a little more detail about the bug you're seeing? Paths are supposed to be sanitized so that things like this don't happen. And in my quick testing, that seems to work here:

```py
>>> import beets.library
>>> lib = beets.library.Library()
>>> item = beets.library.Item(title='hi')
>>> lib.add(item)
>>> lib.path_formats
(('default', '$artist/$album/$track $title'),)
>>> item.destination()
b'/Users/asampson/Music/__/00 hi'
```

As you can see there, the missing $artist and $album fields don't cause this file to get placed in the root directory—they get replaced with underscores.

However, it does look like the underlying path-splitting functionality (`components`) is behaving a little oddly in the cases of empty path components:

```py
>>> beets.util.components('foo/bar')
['foo', 'bar']
>>> beets.util.components('/foo')
['/', 'foo']
>>> beets.util.components('//foo/bar')
['//', 'foo', 'bar']
```

which could explain why this is unreliable. I think it might be nicer to split more reliably on path separators, and replace empty path components with `_` or something by default. This should probably be baked in and not easily overridable with `replace` because it's dangerous to omit—sort of like `path_sep_replace`.

mglukhovsky:
@jackwilsdon: Agreed, in this case your pathname will work (I also ended up using the `%if` operator to ensure that any items with the missing tag were placed in a predictable path.)

@sampyso: This was the result of a poorly formatted pathname in my configuration file:
```
# $grouping is missing in some tracks, so the destination is calculated as `/$artist - $title, placing it in the root directory`
pathname: $grouping/$artist - $title
```

I think replacing empty path components with a default like `_` is a good solution; it would keep generated `destination` as a relative path inside the library directory rather than presenting them as an absolute path. (I can take a look at updating this PR when I have a moment.)

sampsyo:
OK, cool! Thanks for taking a look. FWIW, the place where path sanitization stuff should happen is in the `beets.util` module, whereas this function in `Library` just does the abstract path generation. See `sanitize_path` and the related functions there.

jtpavlock:
Is there still interest in continuing this / getting it merged? Or is the solution @jackwilsdon presented sufficient to fix this.

Issue #604, opened 2014-03-11
Discogs (and other metadata sources) polluting mb_albumid


ghost:
The discogs plugin sets `album_id` of the `AlbumInfo` class to a discogs identifier, which `beets.autotag.apply_metadata` assigns to `mb_albumid`. The problem is that everywhere else it's assumed to be a MusicBrainz identifier.

dengste:
I just wanted to add that this is also true for the artist_id. If an album is identified via discogs, it will be tagged with discogs' artist-id. I stumbled upon this because when I import my library into Ampache, it will frequently have the same artist twice - one with the Musicbrainz id, and one with Discogs', and I have to manually fix that.

skupjoe:
It's also worth noting that the Dicogs data source does not supply mb_trackid information, so (by default) the Duplicates plugin will flag anything that shares null entries for this field as duplicate. Meaning, all Discogs-tagged tracks will be flagged as a duplicate.

The workaround is to additionally add a third parameter to the Duplicate plugin to lastly analyze Title.

nicolahinssen:
@sampsyo Asked me to leave my thoughts here. #2763 

My suggestion is to create the following fields for Discogs (TXXX is customizable so that wouldn't be a problem).

Old | New
--- | ---
TXXX:MusicBrainz Album Artist Id=_1560798_ | TXXX:Discogs Album Artist Id=_1560798_
TXXX:MusicBrainz Album Id=_3175577_ | TXXX:Discogs Album Id=_3175577_
TXXX:MusicBrainz Album Release Country=_Netherlands_ |  TXXX:Discogs Album Release Country=_Netherlands_
TXXX:MusicBrainz Album Type=_Album_ | TXXX:Discogs Album Type=_Album_
TXXX:MusicBrainz Artist Id=_1560798_ | TXXX:Discogs Artist Id=_1560798_

dbogdanov:
Agreed with @nicolahinssen on adding special fields for Discogs. This is a feature I would love to see. 

More generally, it will be great to store metadata from other sources (Bandcamp, Beatport, etc.) in the same way. The plugins retrieving the metadata should be able to define special fields themselves. Probably there won't be a consensus on naming of the fields, therefore this naming could be configurable by a user.

dbogdanov:
> The only trick is that they should also have some fields that are not just dict keys because some fields are required or need special handling—see the apply_metadata code for those fields.

[...]

PR #3322, opened 2019-06-30
465 discogs: Fetch a few more metadata fields

thedevilisinthedetails:


sampsyo:
Looks good at a high level so far! Can you please leave a comment with some background when you're ready for a review?

thedevilisinthedetails:
Sure. So I'm grabbing the additional fields (release date, genre and release id) mentioned in the issue and storing them in the db. For the release I just extracted it from the URI. Genre behaves the same as style (for consistency) and just grabbing the release date.

beetsplug/discogs.py
```diff
@@ -341,18 +344,25 @@ def get_album_info(self, result):
...
+                         data_url=data_url,
+                         discogs_release_id=discogs_release_id,
+                         released_date=released_date)
 
-    def format_style(self, style):
-        if style is None:
-            self._log.debug('Style not Found')
+    def format(self, classification):
+        if classification is None:
+            self._log.debug('Classification not Found')
```
sampsyo: This log message is now not very useful… maybe we can just remove it?

beetsplug/discogs.py
```diff
@@ -341,18 +344,25 @@ def get_album_info(self, result):
...
         else:
-            return self.config['separator'].as_str().join(sorted(style))
+            return self.config['separator'].as_str()\
+                .join(sorted(classification))
+
+    def extract_release_id(self, uri):
+        if uri:
+            return uri.split("/")[-1]
```
sampsyo: It's good style to explicitly `return None` if the function's intent is to sometimes return `None`.

thedevilisinthedetails:
addressed your comments @sampsyo

sampsyo:
Awesome; thank you!! Merged with a changelog entry. :sparkles:

Freso:
Based on the code, it doesn’t seem like this stops writing Discogs ID(s) as MusicBrainz ones? Any chance you could fix that as well? :pray: https://github.com/beetbox/beets/issues/604

thedevilisinthedetails:
> Based on my quick skimming of the code, it doesn’t seem like this stops writing Discogs ID(s) as MusicBrainz ones? Any chance you could fix that as well? 🙏 #604

You sure? When I import using discogs my files are tagged via discogs metadata values. I mean I see that correct values being added in my db

Issue #3309, opened 2019-06-11
Sporadic test failures in BPD tests


sampsyo:
Travis seems to fail surprisingly often in the BPD tests with this error:

```
[...]
  File "/home/travis/build/beetbox/beets/beets/util/bluelet.py", line 404, in __init__
    self.sock.bind((host, port))
OSError: [Errno 98] Address already in use
```

Here's an [example job](https://travis-ci.org/beetbox/beets/jobs/543462082#L1121), for instance.

The problem is that these tests, since #3190, use a socket to set up communication between a server and a client. But if two of these run at the same time, then the server runs into an error when trying to listen on that socket. I tried to mitigate this in a wacky way in https://github.com/beetbox/beets/commit/99778d9ece151736ea2447a70e3fe7efd146b6c8 by randomizing the port over 10,000 options, but failures of this kind are still happening pretty often on Travis.

A potential solution would be to catch this error and re-randomize a few times before giving up.

arcresu:
It seems really weird that there would still be collisions with the port randomised like that. It's not like any of the tests lasts for very long either, and the server should be killed reliably in the `finally` clause. Maybe there's something deterministic about the RNG in the CI environment? I'll take a look but I agree that catching the error is the most reliable way to deal with this flakiness.

zsinskri:
I have not yet looked at this thoroughly, so do not believe anything I say!

According to [this StackOverflow question](https://stackoverflow.com/questions/1365265/on-localhost-how-do-i-pick-a-free-port-number) we can just use port `0` to let the OS choose an available port. The assigned port can then be read as `sock.getsockname()[1]`.

To get that port number from the server process to the main process in order to create the clients, we could use a [multiprocessing.Value](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Value). Though to write to that `Value` after binding the socket and before waiting on it we probably have to mock _something_.

sampsyo:
Wow, that’s very cool! I didn’t know about that Unix feature. It’s at least worth giving it a try!

PR #3329, opened 2019-07-14
util.command_output: return stderr, too

zsinskri:
Return a namedtuple CommandOutput(stdout, stderr) instead of just stdout from
util.command_ouput, allowing separate access to stdout and stderr.

This change is required by the ffmpeg replaygain backend (GitHub
PullRequest #3056) as ffmpeg's ebur128 filter outputs only to stderr.

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -805,7 +805,8 @@ CommandOutput = namedtuple("CommandOutput", ("stdout", "stderr"))
 def command_output(cmd, shell=False):
     """Runs the command and returns its output after it has exited.
 
-    Returns a CommandOutput.
+    Returns a CommandOutput. The attributes ``stdout`` and ``stderr`` contain
+    byte strings of the respective output streams.
 
     ``cmd`` is a list of arguments starting with the command names. The
     arguments are bytes on Unix and strings on Windows.
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -24,7 +24,7 @@ import re
 import shutil
 import fnmatch
 import functools
-from collections import Counter
+from collections import Counter, namedtuple
 from multiprocessing.pool import ThreadPool
 import traceback
 import subprocess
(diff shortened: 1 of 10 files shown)


zsinskri:
I'm kind of lost with this [test failure](https://travis-ci.org/beetbox/beets/jobs/558592144). The exception seems to originate in `bpd` plugin. It tries to start some server on a used address (which of course fails) an then times out while waiting for that non-existing server.

In this branch neither the bpd plugin, nor the space of available addresses has been changed. My best guess is that this is an existing bug that surfaces now due to unhappy timing (and would disappear when rerunning the tests).

Any help?

zsinskri:
Also interestingly the same error occurred in #3056 after rebasing upon this branch, but in a different run.

Here `py27-cov` failed and `py34-test` passed while in #3056 only `py34-test` failed.

jackwilsdon:
Looks like an unstable test - I just re-ran the tests for this branch on Travis and it has passed now 😕

sampsyo:
Indeed; we do have problems with that test. There's more in #3309.

beets/util/__init__.py
```diff
@@ -794,9 +798,15 @@ def convert(arg):
...
 
 
+# stdout and stderr as bytes
+CommandOutput = namedtuple("CommandOutput", ("stdout", "stderr"))
+
+
 def command_output(cmd, shell=False):
     """Runs the command and returns its output after it has exited.
 
+    Returns a CommandOutput.
```
sampsyo: Might be nice to mention here, just for convenience, that you can access `stdout` and `stderr` in the result object, both of which are bytestrings.

zsinskri: Done.

One can also access the byte strings via deconstruction:
```python
stdout, stderr = CommandOutput(b"out", b"err")
```
But I don't think we have to mention that as attribute access is more explicit anyway and prevents one from messing up the order (i.e. `stderr, stdout = CommandOutput(...)`, after which `stdout` is stderr and `stderr` is stdout).

sampsyo: Yes, good point!

sampsyo:
Looks perfect. Thanks again for separating this out! Merged. :sparkles:

PR #3343, opened 2019-08-04
Fix handling of Spotify year-only release dates

rhlahuja:
I've had this fix in place locally for a while but forgot to submit a PR! This properly assigns the `AlbumInfo.year` attribute with year-only release dates returned by the Spotify Albums API and prevents a downstream `sqlite3.InterfaceError` when trying to insert e.g. `[2019]` in an INTEGER field.


sampsyo:
Looks right to me! Thank you!! :sparkles:

Issue #3346, opened 2019-08-17
Importer displays misleading info on album with track 0


IgniFerroque:
On importing this album:

https://musicbrainz.org/release/bbf6ecfb-cb4f-401c-9d05-2b6c22d1366d

I get this:

```/Volumes/Music/import/Guster/1997 - Goldfly (11 items)
Tagging:
    Guster - Goldfly
URL:
    https://musicbrainz.org/release/bbf6ecfb-cb4f-401c-9d05-2b6c22d1366d
(Similarity: 100.0%) (CD, 1997, US, Aware Records, AWA 107)
 * Melanie (#0) -> Melanie (#1)
```

Which suggests that that track will be tagged as track 1, but in fact it is correctly imported as track 0. I would expect the import to show no changes.

wisp3rwind:
Do you enable the `per_disc_numbering` option in your configuration? If yes, the following change in `ui/commands.py` might be enough to fix the problem:

``` diff
diff --git a/beets/ui/commands.py b/beets/ui/commands.py
index 53253c1d..5ad264a8 100755
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -241,7 +241,7 @@ def show_change(cur_artist, cur_album, match):
             if mediums and mediums > 1:
                 return u'{0}-{1}'.format(medium, medium_index)
             else:
-                return six.text_type(medium_index or index)
+                return six.text_type(medium_index if medium_index is not None else index)
         else:
             return six.text_type(index)
```

IgniFerroque:
I did enable that option:

my config:
```comp:~ me$ cat ~/.config/beets/config.yaml
directory: /Volumes/Music/BeetsMusic
library: /Volumes/Music/beetsMusic.db
import:
    move: no
per_disc_numbering: yes
paths:
    default: $albumartist/$album%aunique{}/%if{$multidisc,$disc-}$track $title
plugins: fetchart```

wisp3rwind:
Should be fixed on master now, if it isn't, please reopen this issue. Thanks for the report!

PR #3347, opened 2019-08-19
Correctly display track number 0 in show_change

wisp3rwind:
fixes issue #3346: When the `per_disc_numbering` option was set, the UI would
previously show a `#0 -> #1` change when actually the index would be set
to 0 (a valid index, such as for hidden tracks).  Now, properly
distinguish index 0 and None (i.e. not set)


sampsyo:
Nice; this looks great! Could you please add a quick changelog entry describing the fix?

PR #3348, opened 2019-08-20
convert: add option to symlink or hardlink instead of copying

Vexatos:
This is an updated version of #2326 and 7d26639e134051ac3a8a1e715fb324581b0727dc, resolving merge conflicts and addressing the comments on that PR. Based on proposal in #2324.
Credit goes to @kierdavis for the original PR.

Closes #2326 if merged.

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -307,45 +307,35 @@ class ConvertPlugin(BeetsPlugin):
                                    util.displayable_path(original))
                     util.move(item.path, original)
 
-            linked = False
-
             if should_transcode(item, fmt):
+                linked = False
                 try:
                     self.encode(command, original, converted, pretend)
                 except subprocess.CalledProcessError:
                     continue
             else:
-                if hardlink:
-                    if pretend:
-                        self._log.info(u'ln {0} {1}',
-                                       util.displayable_path(original),
-                                       util.displayable_path(converted))
-                    else:
-                        # No transcoding necessary.
-                        self._log.info(u'Hardlinking {0}',
-                                       util.displayable_path(item.path))
-                        util.hardlink(original, converted)
-                        linked = True
-                elif link:
-                    if pretend:
-                        self._log.info(u'ln -s {0} {1}',
-                                       util.displayable_path(original),
-                                       util.displayable_path(converted))
-                    else:
-                        # No transcoding necessary.
-                        self._log.info(u'Linking {0}',
-                                       util.displayable_path(item.path))
-                        util.link(original, converted)
-                        linked = True
+                linked = link or hardlink
+                if pretend:
+                    msg = 'ln' if hardlink else ('ln -s' if link else 'cp')
+
+                    self._log.info(u'{2} {0} {1}',
+                                   util.displayable_path(original),
+                                   util.displayable_path(converted),
+                                   msg)
                 else:
-                    if pretend:
-                        self._log.info(u'cp {0} {1}',
-                                       util.displayable_path(original),
-                                       util.displayable_path(converted))
+                    # No transcoding necessary.
+                    msg = 'Hardlinking' if hardlink \
+                        else ('Linking' if link else 'Copying')
+
+                    self._log.info(u'{1} {0}',
+                                   util.displayable_path(item.path),
+                                   msg)
+
+                    if hardlink:
+                        util.hardlink(original, converted)
+                    elif link:
+                        util.link(original, converted)
                     else:
-                        # No transcoding necessary.
-                        self._log.info(u'Copying {0}',
-                                       util.displayable_path(item.path))
                         util.copy(original, converted)
 
             if pretend:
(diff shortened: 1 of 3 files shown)


Vexatos:
Added `-H`  and `--hardlink` flags as well as the `hardlink` option to use hardlinks instead of symlinks.

beetsplug/convert.py
```diff
@@ -298,21 +307,46 @@ def convert_item(self, dest_dir, keep_new, path_formats, fmt,
...
+                    if pretend:
+                        self._log.info(u'ln -s {0} {1}',
+                                       util.displayable_path(original),
+                                       util.displayable_path(converted))
+                    else:
+                        # No transcoding necessary.
+                        self._log.info(u'Linking {0}',
+                                       util.displayable_path(item.path))
+                        util.link(original, converted)
+                        linked = True
```
sampsyo: Maybe we can hoist this assignment out and just write `linked = link or hardlink`?

sampsyo: Also, is there any way to factor out some of the logic here so there's less duplicated code? For example, perhaps we could assign to a string to switch between `'ln'` and `'ln -s'` but only keeping one copy of the `_log.info` line.

beetsplug/convert.py
```diff
@@ -399,15 +434,36 @@ def copy_album_art(self, album, dest_dir, path_formats, pretend=False):
...
+            elif link:
+                if pretend:
+                    self._log.info(u'ln -s {0} {1}',
+                                   util.displayable_path(album.artpath),
+                                   util.displayable_path(dest))
+                else:
+                    self._log.info(u'Linking cover art from {0} to {1}',
+                                   util.displayable_path(album.artpath),
+                                   util.displayable_path(dest))
+                    util.link(album.artpath, dest)
```
sampsyo: As above—it would be great to reduce the duplicated code between these branches.

beetsplug/convert.py
```diff
@@ -426,6 +482,21 @@ def convert_func(self, lib, opts, args):
...
+                hardlink = opts.hardlink
+            else:
+                hardlink = self.config['hardlink'].get(bool) and not link
+        else:
+            link = self.config['link'].get(bool)
+
+            if opts.hardlink is not None:
+                hardlink = opts.hardlink
+            else:
+                hardlink = self.config['hardlink'].get(bool)
```
sampsyo: Maybe something like this would be less repetitive?

```py
if opts.link is not None:
    link = True
    hardlink = False
elif opts.hardlink is not None:
    link = False
    hardlink = True
else:
    # use config
````

Or maybe we could use an enum with three values instead of two separate flags?

File: docs/changelog.rst [hunk 4]
```diff
@@ -11,6 +11,13 @@ New features:
   (the MBID), and ``work_disambig`` (the disambiguation string).
   Thanks to :user:`dosoe`.
   :bug:`2580` :bug:`3272`
+* :doc:`/plugins/convert`: Added new ``-l`` (``--link``) flag and ``link`` option
+  which symlinks files that do not need to be converted instead of copying them.
+  :bug:`2324`
+* :doc:`/plugins/convert`: Added new ``-H`` (``--hardlink``) flag and ``hardlink``
+  option which hardlinks files that do not need to be converted
+  instead of copying them.
+  :bug:`2324`
```
sampsyo: No need for two changelog entries; one general message will do fine.

docs/plugins/convert.rst
```diff
@@ -48,6 +48,12 @@ To test your configuration without taking any actions, use the ``--pretend``
 flag. The plugin will print out the commands it will run instead of executing
 them.
 
+By default, files that do not need to be transcoded will be copied to their
+destination. Passing the ``-l`` (``--link``) flag creates symbolic links
+instead, passing ``-H`` (``--hardlink``) creates hard links.
+Refer to the ``link`` and ``hardlink`` options below
+for potential issues with this.
```
sampsyo: Maybe it would be nice to mention exactly what those issues are—namely, just say that album art embedding won't work?

docs/plugins/convert.rst
```diff
@@ -93,6 +99,22 @@ file. The available options are:
...
+  Default: ``false``.
+- **hardlink**: By default, files that do not need to be transcoded will be
+  copied to their destination. This option creates hard links instead. Note that
+  options such as ``embed`` that modify the output files after the transcoding
+  step will cause the original files to be modified as well if ``hardlink`` is
+  enabled. For this reason, album-art embedding is disabled
+  for files that are linked.
+  This option overrides ``link``. Only works when converting to a directory
+  on the same filesystem as the library.
+  Default: ``false``.
```
sampsyo: No need to copy n' paste approximately the same description for both options—just say that the second one works similarly to the first, but it hard-links files instead.

beetsplug/convert.py
```diff
@@ -298,21 +307,46 @@ def convert_item(self, dest_dir, keep_new, path_formats, fmt,
...
+            linked = False
+
             if should_transcode(item, fmt):
                 try:
                     self.encode(command, original, converted, pretend)
                 except subprocess.CalledProcessError:
                     continue
             else:
-                if pretend:
-                    self._log.info(u'cp {0} {1}',
```
GuilhermeHideki: If you keep the `if pretend:` here, you should be able to change the `cp {0} {1}` to `{2} {0} {1}` or `{0} {1} {2}` and pass the command name

beetsplug/convert.py
GuilhermeHideki: maybe using the `MoveOperation` could be a good idea

beetsplug/convert.py
Vexatos: `link` and `hardlink` are set globally for the entire command. This way, copying album art still occurs for files that actually were converted, and is only skipped on those that are linked.

beetsplug/convert.py
Vexatos: This is a very good idea.

Vexatos:
Thank you for the suggestions. I reduced the amount of duplicate code now.

sampsyo:
Yay! Looks great! Thanks for tackling this.

kierdavis:
Thanks @Vexatos!

And sorry about abandoning this before; it got pushed down my todo list once my use case for it disappeared :(

Issue #1693, opened 2015-11-03
Apply `data_source` field to albums


l-t-k:
I like to check which albums in my library are from discogs. The use of `beet list data_source:discogs` returns all the tracks I like to see. However, when using `beet list -a data_source:discogs` the return is zero. It seems like the flexible attribute data_source is a item field and an album field.

sampsyo:
That's odd; it definitely should be applied to albums. Can you please check with `beet list` whether *any* of your music has an album-level `data_source` set?

kooimens:
I've tried several albums, including mb and discogs, none of them is having an album-level field.
If I configure this: 
`format_item: $album - $disc$track - $artist - $title - $data_source`. I get:
`The Year 2000 - 0101 - The O'Jays - The Year 2000 - Discogs`

`format_album: $albumartist - $album - $data_source` I get: 
The O'Jays - The Year 2000 - $data_source

sampsyo:
Thanks. I looked into it a little more—I should have remembered that we don't apply *anything* directly to album structures. (It's all just inferred from item-level data, and that doesn't include flexible attributes.)

This would be a useful thing to add.

BeetsStudentTeam:
Hey guys, we would like to tackle this issue, if anyone is going to work on this soon let me know so we can look at other issues! We're planning to start working on this on Monday February 1st. Thanks! -Jake

bartkl:
@jacob-elliott Have you begun working on this issue?
@sampsyo Is the idea to enable the inferring of the `data_source` field from item to album level, or more general to enable this kind of inferrence for _any_ flexible attribute? The latter sounds better to me, but then the next question would be: how to configure which flexible attributes are to be passed to the album level and which are not (config file?).

Just thinking out loud and depending on the complexity I might pick this up myself :).

Mary011196:
Hello, 
I would like to tackle this issue,is anyone that working on this?
Thank you

sampsyo:
Go for it, @Mary011196! Please let us know if more context would be useful.

Looks like I forgot to respond to @bartkl—oops! I'm not 100% sure about promoting arbitrary track-level fields to album-level fields. The generality is definitely a good idea, but because our metadata backends are fairly controlled, we can get the same generality by asking them to assign attributes to the `AlbumInfo` objects. That would remove the need to "guess" whether track-level data needs to be promoted.

Mary011196:
Hello,
[...]

PR #3350, opened 2019-08-22
Fix for issue #1693 (Apply `data_source` field to albums)

apitofme:
This adds the `data_source` field (flex-attribute) to Albums during an import, it copies the value from the  first Item's corresponding field.

Does this need a conditional test to make sure that the Item field exists and actually has a value first?

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -754,6 +754,7 @@ class ImportTask(BaseImportTask):
             self.record_replaced(lib)
             self.remove_replaced(lib)
             self.album = lib.add_album(self.imported_items())
+            self.album.set_parse('data_source', self.imported_items()[0].data_source)
             self.reimport_metadata(lib)
 
     def record_replaced(self, lib):
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -754,7 +754,8 @@ class ImportTask(BaseImportTask):
             self.record_replaced(lib)
             self.remove_replaced(lib)
             self.album = lib.add_album(self.imported_items())
-            self.album.set_parse('data_source', self.imported_items()[0].data_source)
+            if self.imported_items()[0].data_source:
+                self.album.data_source = self.imported_items()[0].data_source
             self.reimport_metadata(lib)
 
     def record_replaced(self, lib):
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -754,7 +754,7 @@ class ImportTask(BaseImportTask):
             self.record_replaced(lib)
             self.remove_replaced(lib)
             self.album = lib.add_album(self.imported_items())
-            if self.imported_items()[0].data_source:
+            if 'data_source' in self.imported_items()[0].keys():
                 self.album.data_source = self.imported_items()[0].data_source
             self.reimport_metadata(lib)
 
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -754,7 +754,7 @@ class ImportTask(BaseImportTask):
             self.record_replaced(lib)
             self.remove_replaced(lib)
             self.album = lib.add_album(self.imported_items())
-            if 'data_source' in self.imported_items()[0].keys():
+            if 'data_source' in self.imported_items()[0]:
                 self.album.data_source = self.imported_items()[0].data_source
             self.reimport_metadata(lib)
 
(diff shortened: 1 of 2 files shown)


beets/importer.py
```diff
@@ -754,6 +754,7 @@ def add(self, lib):
             self.record_replaced(lib)
             self.remove_replaced(lib)
             self.album = lib.add_album(self.imported_items())
+            self.album.set_parse('data_source', self.imported_items()[0].data_source)
```
sampsyo: How about `self.album.data_source = ...` instead of using the `set_parse` method?

sampsyo:
Thanks for getting this started! Indeed, it does need a check like that to avoid an error when the field is missing. You can see this by [running the tests](https://github.com/beetbox/beets/wiki/Testing) if you want!

apitofme:
> sampsyo commented 2 hours ago
> ... You can see this by running the tests if you want!

Just a 'curiosity' from running the tests... In addition to the extra Python modules required to run the tests (as listed in `setup.py`) I got errors reporting that `confuse` and `mediafile` were missing.
_(I thought they were installed as dependencies to beets?!)_

Anyway, other than that I just had to comment out the custom output templates in my config file.

There was also this error output:

```
ERROR: test_relative_to (test.test_play.PlayPluginTest)
  File "x:\development\python\venv\lib\site-packages\mock\mock.py", line 1330, in patched
    return func(*args, **keywargs)
  File "X:\Development\Python\beets\test\test_play.py", line 80, in test_relative_to
    path = os.path.relpath(self.item.path, b'/something')
  File "x:\development\python\venv\lib\ntpath.py", line 562, in relpath
    path_drive, start_drive))
ValueError: path is on mount b'C:', start on mount b'X:'
-------------------- >> begin captured logging << --------------------
```
which didn't seem important but does perhaps indicate that the test needs to be modified to allow for relative paths (?) starting from mount points other than drive `C:` ?

sampsyo:
> I got errors reporting that `confuse` and `mediafile` were missing.
> _(I thought they were installed as dependencies to beets?!)_

Yes, they should indeed be! Maybe a reinstall with `pip install -e .` would fix it?

> which didn't seem important but does perhaps indicate that the test needs to be modified to allow for relative paths (?) starting from mount points other than drive `C:` ?

Hmm; interesting! I don't quite see the problem yet, but if you're interested in helping out, perhaps you could try fixing this test in a separate PR! It is, however, suspicious that the test doesn't run into a problem on AppVeyor?

apitofme:
> Maybe a reinstall with pip install -e . would fix it?

I just installed them separately with `pip`

> It is, however, suspicious that the test doesn't run into a problem on AppVeyor?

It's probably only an issue with my particular set-up

...

Anyway I'm glad that this PR passes the tests now.
Thanks!

beets/importer.py
```diff
@@ -754,6 +754,8 @@ def add(self, lib):
             self.record_replaced(lib)
             self.remove_replaced(lib)
             self.album = lib.add_album(self.imported_items())
+            if 'data_source' in self.imported_items()[0].keys():
```
sampsyo: I think a simpler expression `'data_source' in self.imported_items()[0]` should work… any chance you could give that a quick try?

apitofme: No worries, all done
...I must admit I had wondered if it was possible that way, but with my inexperience and in my haste I neglected to test it first. Thanks again  :)

sampsyo:
Great! If this indeed seems to fix the problem in your tests, can you please add a quick changeling entry and we'll call this done? :sparkles:

apitofme:
Changelog updated - since the [original request](https://github.com/beetbox/beets/issues/1693) was tagged as a 'feature' I added it under _'New Features'_
... I assume/hope that's right (although it feels like a bug-fix!)

sampsyo:
Looks great; thank you! Just to double check, have you tried importing an album with this to confirm that the fix actually works and the album gets the data source field as intended? If so, then we're good to go!

apitofme:
I did indeed test it, though hardly extensively ... I only tested it with one album import  :/

Basically [as per the evidence I submitted](https://github.com/beetbox/beets/issues/1693#issuecomment-522323973) during the work building towards this PR, I would simply remove an existing album from beets (in this case _"How to Destroy Angels"_) and import it again. I would then perform a search, first for items then for the album, with the `data_source` field included in the output for comparison -- I did this each time I made changes to the code for this feature.

sampsyo:
Perfect! Thanks for confirming. And thanks for your work on this!

PR #3355, opened 2019-09-02
Add Deezer plugin

rhlahuja:
Add a [Deezer](https://www.deezer.com) plugin to allow tracks and albums to be matched using the [Deezer API](https://developers.deezer.com/api).

Introduce a base `MetadataSourcePlugin` class to define shared API import logic. This will be integrated into the Spotify plugin in a followup PR.

### DIFF
--- a/beets/autotag/__init__.py
+++ b/beets/autotag/__init__.py
@@ -22,13 +22,13 @@ from beets import logging
 from beets import config
 
 # Parts of external interface.
-from .hooks import (
+from .hooks import (  # noqa
     AlbumInfo,
     TrackInfo,
     AlbumMatch,
     TrackMatch,
     Distance,
-)  # noqa
+)
 from .match import tag_item, tag_album, Proposal  # noqa
 from .match import Recommendation  # noqa
 
--- a/beets/autotag/__init__.py
+++ b/beets/autotag/__init__.py
@@ -18,11 +18,21 @@
 
 from __future__ import division, absolute_import, print_function
 
+import re
+from abc import abstractmethod, abstractproperty
+
 from beets import logging
 from beets import config
+from beets.plugins import BeetsPlugin
 
 # Parts of external interface.
-from .hooks import AlbumInfo, TrackInfo, AlbumMatch, TrackMatch  # noqa
+from .hooks import (
+    AlbumInfo,
+    TrackInfo,
+    AlbumMatch,
+    TrackMatch,
+    Distance,
+)  # noqa
 from .match import tag_item, tag_album, Proposal  # noqa
 from .match import Recommendation  # noqa
 
@@ -31,6 +41,7 @@ log = logging.getLogger('beets')
 
 
 # Additional utilities for the main interface.
+
 
 def apply_item_metadata(item, track_info):
     """Set an item's metadata from its matched TrackInfo object.
@@ -72,14 +83,15 @@ def apply_metadata(album_info, mapping):
     for item, track_info in mapping.items():
         # Artist or artist credit.
         if config['artist_credit']:
-            item.artist = (track_info.artist_credit or
-                           track_info.artist or
-                           album_info.artist_credit or
-                           album_info.artist)
-            item.albumartist = (album_info.artist_credit or
-                                album_info.artist)
+            item.artist = (
+                track_info.artist_credit
+                or track_info.artist
+                or album_info.artist_credit
+                or album_info.artist
+            )
+            item.albumartist = album_info.artist_credit or album_info.artist
         else:
-            item.artist = (track_info.artist or album_info.artist)
+            item.artist = track_info.artist or album_info.artist
             item.albumartist = album_info.artist
 
         # Album.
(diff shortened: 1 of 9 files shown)


sampsyo:
Awesome! Looking pretty good so far.

I have just a couple of high-level questions:

- Since you said you repurposed some logic from your Spotify plugin, would it make sense to factor out some code that gets reused into a common library?
- It looks like there are also changes to the Spotify plugin here too—would you mind breaking those out into a separate PR?

rhlahuja:
- Yeah, I was going back and forth on this. Some helpers like `album_distance` and `track_distance` can easily be factored out, although some of the higher level functions like `track_for_id` and `album_for_id` may not quite be similar enough to easily modularize. I'll take another look at this later and refactor what makes sense. Is there a particular place you'd like to place the common library? Something like `beetsplug/api_utils.py`?
- Sure thing -- I'll include the aforementioned Spotify refactor in a separate PR.

sampsyo:
Cool; thanks for looking into it! Depending on the exact scope, something like `beets.autotag` or `beets.autotag.match` could be a good fit for these reusable bits.

rhlahuja:
@sampsyo I've factored out shared logic into a base `APIAutotaggerPlugin` class (better name suggestions welcome). I plan to revert all changes to `spotify.py` and `discogs.py` and include them in a separate PR, but I committed them all here for now to get a sense of the final state. Before I begin testing and documenting, what are your thoughts so far?

beets/autotag/__init__.py
```diff
@@ -197,3 +211,149 @@ def apply_metadata(album_info, mapping):
...
+    for albums or individual tracks.
+    """
+    dist = Distance()
+    if info.data_source == data_source:
+        dist.add('source', config['source_weight'].as_number())
+    return dist
+
+
+@six.add_metaclass(abc.ABCMeta)
+class APIAutotaggerPlugin(object):
```
sampsyo: In retrospect, perhaps the most natural home for this base class would be in the `beets.plugins` module?

In general, we have talked several times in beets's history about creating "template" base classes like this one to make it simpler to write certain kinds of plugins. So I'm very happy to see that you came to the same idea independently! It would be great to introduce more of these, so perhaps creating a new place to gather plugin templates together would be the thing to do.

sampsyo: Also, what do you think about the name `MetadataSourcePlugin`?

docs/changelog.rst
```diff
@@ -66,6 +66,10 @@ New features:
 * The 'data_source' field is now also applied as an album-level flexible
   attribute during imports, allowing for more refined album level searches.
   :bug:`3350` :bug:`1693`
+* :doc:`/plugins/deezer`:
+   * Added Deezer plugin as an import metadata provider: you can now match tracks
+     and albums using the `Deezer`_ database.
+     Thanks to :user:`rhlahuja`.
 
 Fixes:
```
sampsyo: We could even mention the new base class in a "for developers" section.

docs/changelog.rst
```diff
@@ -66,6 +66,10 @@ New features:
 * The 'data_source' field is now also applied as an album-level flexible
   attribute during imports, allowing for more refined album level searches.
   :bug:`3350` :bug:`1693`
+* :doc:`/plugins/deezer`:
+   * Added Deezer plugin as an import metadata provider: you can now match tracks
+     and albums using the `Deezer`_ database.
+     Thanks to :user:`rhlahuja`.
```
sampsyo: No need for a nested list here if it's just one bullet.

docs/plugins/deezer.rst
```diff
@@ -0,0 +1,38 @@
...
+
+.. _Deezer: https://www.deezer.com
+.. _Album: https://developers.deezer.com/api/album
+.. _Track: https://developers.deezer.com/api/track
+
+Why Use This Plugin?
+--------------------
+
+* You're a Beets user.
+* You want to autotag music with metadata from the Deezer API.
```
sampsyo: This section doesn't seem to add too much—both of these points are pretty well covered by your introductory sentence above. :smiley:

beets/autotag/__init__.py
rhlahuja: Awesome! I was unsure if this class is the right approach, but it's great to hear something like this was already being considered. And I agree, `beets.plugins.MetadataSourcePlugin` seems like a more natural place/name for this.

rhlahuja:
@sampsyo this is tested and ready for a final review. Once merged, I'll open another PR for reusing this shared logic in the Spotify and Discogs plugins.

beets/autotag/__init__.py
```diff
@@ -22,7 +22,13 @@
...
 
 # Parts of external interface.
-from .hooks import AlbumInfo, TrackInfo, AlbumMatch, TrackMatch  # noqa
+from .hooks import (  # noqa
+    AlbumInfo,
+    TrackInfo,
+    AlbumMatch,
+    TrackMatch,
+    Distance,
+)
```
sampsyo: I think this change can be reverted now because `Distance` is no longer used here.

sampsyo: Oops; I'm obviously wrong about that. Not changing anything!

Issue #3242, opened 2019-04-27
Make `none_rec_action` respect `timid`


RollingStar:
### Problem

https://github.com/beetbox/beets/blob/9abfc45adc5dd864840c9b72f35239c6204d39fa/beets/ui/commands.py#L493

When you have a none_rec and you set the none_rec_action to "asis", beets will import as-is even in timid mode. More predictable behavior would be having timid override the none_rec_action.

sampsyo:
Got it; yes, that does make sense! It does seem like `--timid` should override all "summary judgments."

RollingStar:
https://github.com/beetbox/beets/pull/3357

sampsyo:
Fixed in #3357.

PR #3357, opened 2019-09-03
 Make none_rec_action respect timid. #3242 

RollingStar:


beets/ui/commands.py
```diff
@@ -496,6 +496,9 @@ def _summary_judgment(rec):
             'asis': importer.action.ASIS,
             'ask': None,
         })
+        # prompt the user if timid is enabled
+        if config['import']['timid'] and action == importer.action.ASIS:
```
RollingStar: Is `importer.action.ASIS` the proper tool to use here? The as_choice function from beetbox/confuse complicates things.

sampsyo: Yep, looks right to me.

sampsyo:
Thanks for getting this started! I would have expected a somewhat stronger "override," though—i.e., the importer would ask you every time in timid mode, even if the recommendation is to "skip." That also has a somewhat simpler implementation—the first `if` in the `elif` cascade here would just say:

```py
if config['import']['timid']:
    return None
```

…and then proceed on to the other cases. Would that be too aggressive for your use case?

RollingStar:
I viewed timid as "prompt before changing anything", not "prompt before doing anything". [The documentation](https://beets.readthedocs.io/en/latest/reference/config.html#timid) and [commands](https://github.com/beetbox/beets/blob/96d61b8b9305957b1f4c438386ac97fc6a5c7571/beets/ui/commands.py#L1011) lean towards your approach:

>Either yes or no, controlling whether the importer runs in timid mode, in which it asks for confirmation on every autotagging match, even the ones that seem very close. Defaults to no. The -t command-line flag controls the same setting.

>help=u'always confirm all actions'

RollingStar:
Addressed your feedback with a new commit.

sampsyo:
Looks perfect; thank you!! :sparkles:

PR #3079, opened 2018-11-10
web: Improve UI with bootstrap 4

matlads:
* Upgrade jQuery, underscore, and backbone to the latest versions
* Introduce bootstrap 3.3.7 CSS
* Add a brand.png icon to use for the banner, and favicon
* Patch template to use the bootstrap class for a bootstrap look

PR #3359, opened 2019-09-05
web plugin: Improve UI with bootstrap 4

matlads:
These commits do the following: 

1. updated backbone, jquery, 
2. introduces howler (the audio library), and bootstrap css framework
3. uses the /stats, /album, and search API end points
4. introduces the concept of a queue for playing music from the collection
5. updates documentation accordingly


sampsyo:
Awesome; thank you for keeping up the good work! To confirm, this is a successor to #3079, right? @dbogdanov, did you want to take a look?

matlads:
> Awesome; thank you for keeping up the good work! To confirm, this is a successor to #3079, right? @dbogdanov, did you want to take a look?

Yes it is the successor to #3079.

eidoom:
@matlads This web UI is looking really nice! 

One thing I noticed: everything is working when I run it on http://127.0.0.1:8337/ but when I run it to be accessed over my local network as, say, http://192.168.0.2:8337/ then it is as though my library is empty. Artists, albums and stats are all empty and searches return no results. Is this something you have tested? And do others reproduce this behaviour?

jackwilsdon:
@eidoom have you tried setting `host` to `0.0.0.0`? 

https://beets.readthedocs.io/en/stable/plugins/web.html#configuration

eidoom:
> @eidoom have you tried setting `host` to `0.0.0.0`?
> 
> https://beets.readthedocs.io/en/stable/plugins/web.html#configuration

This gives the same behaviour for me: empty library on the network address and working on the localhost address.

beetsplug/web/static/js/beets.js
```diff
@@ -0,0 +1,562 @@
+/* globals _,$,Backbone,Marionette,Howl */
+(function () {
+    "use strict";
+
+    // Constants
+    var SETTINGS = {
+        API: "http://127.0.0.1:8337",
```
jackwilsdon: @eidoom looks like this here is the problem. This needs changing before merging @matlads, we should be able to use `window.location.origin` here instead.

```suggestion
        API: window.location.origin,
```

eidoom: This fixed the problem for me :) Thank you @jackwilsdon

dbogdanov:
This definitely looks much better than before! Works fine for me on localhost. This can be merged to foster further progress on top of it. All ideas for improvements can go to a new feature request issue.

PR #3361, opened 2019-09-09
Convert: create convert_lossy to replace never_convert_lossy_files

### LINKED ISSUE #3124: never_convert_lossy_files should flip to convert_lossy_files
### Problem

Convert: `never_convert_lossy_files` is one of the only config options I see in beets that uses negative wording. Instead of `never_convert_lossy_files: no`, the default could be `convert_lossy_files: yes`

RollingStar:
Fixes #3124

Would like some feedback on how I did things. Will update documentation + changelog once it looks good.

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -95,24 +95,22 @@ def should_transcode(self, item, fmt):
     conversion (i.e., its bitrate is high or it has the wrong format).
     """
     no_convert_queries = config['convert']['no_convert'].as_str_seq()
-    self._log.debug("asdf")
-    convert_lossy = config['convert']['never_convert_lossy_files'].as_str()
-
-    if convert_lossy:
-        convert_lossy = False
-        self._log.debug(u'Deprecated option "never_convert_lossy_files" is set.')
-        if config['convert']['convert_lossy'] != convert_lossy:
-            self._log.error(u'Opposite values in "convert_lossy" and "never_convert_lossy_files"')
-            # User can convert to lossy format after correcting the config.
-            return False
-    else:
-        convert_lossy = config['convert']['convert_lossy']
-
     if no_convert_queries:
         for query_string in no_convert_queries:
             query, _ = parse_query_string(query_string, Item)
             if query.match(item):
                 return False
+
+    try:
+        convert_lossy = not config['convert']['never_convert_lossy_files']
+        self._log.debug(u'Deprecated option "never_convert_lossy_files" is set.')
+    except:
+        convert_lossy = config['convert']['convert_lossy']
+    if str(convert_lossy) != str(config['convert']['convert_lossy']):
+        self._log.error(u'Opposite values in "convert_lossy" and "never_convert_lossy_files"')
+        # User can convert to lossy format after correcting the config.
+        # This is more reversible than the alternative.
+        return False
     if not convert_lossy and not (item.format.lower() in LOSSLESS_FORMATS):
         return False
     maxbr = config['convert']['max_bitrate'].get(int)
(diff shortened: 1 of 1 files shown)


beetsplug/convert.py
```diff
@@ -90,7 +90,7 @@ def get_format(fmt=None):
     return (command.encode('utf-8'), extension.encode('utf-8'))
 
 
-def should_transcode(item, fmt):
+def should_transcode(self, item, fmt):
```
RollingStar: Added `self` to allow logging calls. Not 100% sure of the ramifications of this.

beetsplug/convert.py
```diff
@@ -100,8 +100,18 @@ def should_transcode(item, fmt):
             query, _ = parse_query_string(query_string, Item)
             if query.match(item):
                 return False
-    if config['convert']['never_convert_lossy_files'] and \
-            not (item.format.lower() in LOSSLESS_FORMATS):
+
+    try:
+        convert_lossy = not config['convert']['never_convert_lossy_files']
```
RollingStar: This might be a little loose. Configs are handled through the `confuse` library? I couldn't find how it maps ex. "true" / "yes" / True (no quotes) in a config to boolean True / False.

beetsplug/convert.py
```diff
@@ -100,8 +100,18 @@ def should_transcode(item, fmt):
...
+        self._log.debug(u'Deprecated option "never_convert_lossy_files" is set.')
+    except:
+        convert_lossy = config['convert']['convert_lossy']
+    if str(convert_lossy) != str(config['convert']['convert_lossy']):
+        self._log.error(u'Opposite values in "convert_lossy" and "never_convert_lossy_files"')
+        # User can convert to lossy format after correcting the config.
+        # This is more reversible than the alternative.
+        return False
```
RollingStar: I assume this should throw an error rather than copy willy-nilly.

beetsplug/convert.py
```diff
@@ -100,8 +100,18 @@ def should_transcode(item, fmt):
             query, _ = parse_query_string(query_string, Item)
             if query.match(item):
                 return False
-    if config['convert']['never_convert_lossy_files'] and \
-            not (item.format.lower() in LOSSLESS_FORMATS):
+
+    try:
```
RollingStar: This is a try / except block, which apparently cause some controversy when used in Python. Elsewhere in `convert.py`, it [enumerates keys](https://github.com/beetbox/beets/blob/5f0d7c6d0a3f95846ad1d28f766aecdb34775580/beetsplug/convert.py#L79) and checks if certain keys exist. I couldn't get that working, but could try again if needed.

beetsplug/convert.py
sampsyo: FWIW, I think it's a little more common to just pass around the log rather than the entire `self` object. (And if we were going to do that, I might name it something like `plugin` instead of `self`, which could be misinterpreted.)

beetsplug/convert.py
sampsyo: Indeed, it's generally a good idea to avoid what's called a "bare except," where the `except:` line doesn't specify the type of exception that's expected. The problem is that it can silence exceptions you really want to know about, like a `NameError` if you misspell the name of a variable.

Using a try/except block is fine, but if you know what exception you're looking for, you can specify that—for example, `except KeyError:`.

beetsplug/convert.py
sampsyo: I think you can use `'never_convert_lossy_files' in config['convert']` to check whether the option has a setting.

beetsplug/convert.py
sampsyo: I think it would be fine to just let the new option override the old one. If it's set, the user presumably knows what they're doing—and I expect setting both to be very rare.

jtpavlock:
@RollingStar any interest in continuing this?

PR #3363, opened 2019-09-10
discogs: append 'Featuring' artists to 'artist'

### LINKED ISSUE #3353: discogs: add 'Featuring' artists to the 'artist' field
### Use case

Currently the Discogs importer throws away info about 'featuring' artists, as this is in the `credits` / `extraartists`, not `artists`, along with other roles like `Producer`. This is quite different from the behavior of the musicbrainz importer due to how MusicBrainz stores this information. I'd expect the artist to match up when importing the same album from two different importers, other than cosmetic differences due to differing database guidelines.

### Solution

For feature/behavior parity with the default MusicBrainz importer, I think we should append these to the artist field, i.e. ` feat. SomeArtist, AnotherArtist`.

kergoth:
The extraartists from the Discogs API do not include a 'join' field the
way artists are, so they're simply joined with ', ' as is typical for
artists, and are appended to the artist after 'feat. '.

With the discogs API, tracks don't always include artist, if it matches
the artist of the album, so we need to hold onto the track featuring
artists and append it to the albumartist to get the correct track
artist.

Close #3353 

In one example track, 'Everywhere I Go' from 'Starshipz and Rocketz' by 'G-Side', the artist changes from `G-Side` to `G-Side feat. Adrienne, Lance, Whosane`, see the album [Discogs Release](https://www.discogs.com/G-Side-Starshipz-And-Rocketz/release/1631636)

While I'm not 100% happy with injecting into TrackInfo, it was simpler than refactoring how tracks are passed around to the various methods, and makes a certain amount of sense. I'm also not a big fan of the way I duplicated the lines to clean up the artist name, but I'm not convinced splitting it into a method is worth the overhead it would add. Thoughts welcome.

### DIFF
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -292,7 +292,7 @@ class DiscogsPlugin(BeetsPlugin):
             self._log.warning(u"Release does not contain the required fields")
             return None
 
-        artist, artist_id = self.get_artist([a.data for a in result.artists])
+        artist, artist_id, _ = self.get_artist([a.data for a in result.artists], [a.data for a in result.credits] if result.data.get('credits') else None)
         album = re.sub(r' +', ' ', result.title)
         album_id = result.data['id']
         # Use `.data` to access the tracklist directly instead of the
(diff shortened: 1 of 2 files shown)


kergoth:
Fix forthcoming.

kergoth:
Setting aside for rework in a few days.

JOJ0:
This feature was implemented in https://github.com/beetbox/beets/pull/4515

PR #3364, opened 2019-09-10
library: show album id in empty album error

kergoth:
This makes it possible to recover from this case, as you can correct
whatever caused it, by either fixing album ids on tracks or removing the
empty album id.

I ran into this with joining albums and manual manipulation of the ids, which I expect most won't run into, but it's better to not get into a state that we can't easily recover from, so I think it's worthwhile to include the info needed to recover.

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -1119,7 +1119,7 @@ class Album(LibModel):
         """
         item = self.items().get()
         if not item:
-            raise ValueError(u'empty album')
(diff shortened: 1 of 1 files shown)


sampsyo:
Seems good to me! Indeed, this should not happen in "normal" uses of beets, but I don't see any harm in including more detail for advanced h4x0rs like yourself.

PR #3366, opened 2019-09-12
Fix #3358 

### LINKED ISSUE #3358: On some Linux systems, Beets default locale of utf-8 disagrees with subprocess
### Summary

First a description that should help to understand the problem, then I'll fill out the issue template.

Beets attempts to encode with the system default encoding when crafting strings for external programs, [for example here](https://github.com/beetbox/beets/blob/master/beetsplug/convert.py#L229). It does this with [`util.arg_encoding()`](https://github.com/beetbox/beets/blob/master/beets/util/__init__.py#L328):

    def arg_encoding():
        try:
            return locale.getdefaultlocale()[1] or 'utf-8'
        except ValueError:
            return 'utf-8'

On this Ubuntu install which didn't configure a locale, `locale.getdefaultlocale()[1]` is `None`. So Beets intelligently defaults to utf-8. However, it then passes that string to `subprocess`, which isn't so intelligent. It calls [Python's `fork_exec`](https://github.com/python/cpython/blob/3.6/Modules/_posixsubprocess.c#L557), which calls [`PyUnicode_FSConverter`](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_FSConverter) on each argument. When this attempts to encode the arguments, [according to the documentation](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_EncodeFSDefault), 

> If Py_FileSystemDefaultEncoding is not set, fall back to the locale encoding.

So what is the locale encoding? We can get it with `locale.getpreferredencoding()`. Unfortunately, the [documentation](https://docs.python.org/3/library/locale.html#locale.getpreferredencoding) doesn't say what the default encoding is, but we can check [the code](https://github.com/python/cpython/blob/3.6/Lib/locale.py#L610):

                res = getdefaultlocale()[1]
                if res is None:
                    # LANG not set, default conservatively to ASCII
                    res = 'ascii'
                return res

(Mournfully) "oh..."

### Problem

On a Linux with no locale set (e.g. minimal Ubuntu install), add a track with a non-ASCII filename to your database, then do 

    beet convert -d outdir

Result:

    UnicodeEncodeError: 'ascii' codec can't encode character '\u2019' in position 23: ordinal not in range(128)

The issue is [fixed](https://askubuntu.com/a/227513) with `locale-gen` and `dpkg-reconfigure locales`, followed by logging out and in again. 

Expected result: I'm actually not sure. Possible solutions:

1. Check if `locale.getpreferredencoding()` is an ASCII, and if so, coerce the filename to ASCII before passing it to `subprocess`. 
2. On commands that will involve `subprocess`, check if locale is undefined and error out early with an intelligent message for the user. 
3. Maybe there's someway to force `subprocess` to use UTF-8? I don't know of one, and I'm not sure if this might have some side effects. 

### Setup

* OS: Ubuntu 18.04.1 LTS
* Python version: 3.6.8
* beets version: 1.4.6-2

My configuration (output of `beet config`) is: n/a

afontenot:
I'm not 100% sure this solves the problem the correct way, but I wanted you to have a look at it.

Here's what I discovered. Passing `str` types to Python functions for file paths is actually the [recommended](https://docs.python.org/3/howto/unicode.html#unicode-filenames) way to do it. If you pass a string, it's true Python will encode it with the system locale (causing the bug I found), *but* I've discovered that as long as the string has the surrogates properly escaped, the file paths work. So even if my encoding is ASCII, `subprocess` has no complaints with paths that contain bytes that don't decode to ASCII, so long as you pass a string in that's got surrogate escaping for all non-ASCII characters.

Currently the process looks like this.

1. `convert.py` encodes args with the system's encoding (defaulting to UTF-8)
2. `util` takes the args and decodes them (defaulting to UTF-8) into `str` args with surrogate escaping.
3. `subprocess` gets the args and tries to encode them (defaulting to ASCII)

Note that `surrogateescape` is an error handler. When we decode `bytes` as UTF-8 with that handler (as in step 2), a ton of bytes outside of the ASCII range will get decoded without escaping because they don't raise errors. This is what causes the exception in `subprocess` to happen - it can't convert these characters to ASCII!

What I've discovered (this is the main part for you to double check) is that the `str` args handed to us at the beginning of step 1 are *already* escaped with the `arg_encoding()` locale! So the round-trip conversion is not only unnecessary, it's actually what breaks it in the case where there's no set locale.

I've checked that this works on systems with and without the locale defined. Note that if this is merged some other code can be deleted, e.g. in `convert_command_args`. I haven't done that yet until it's clear that this is the right approach, and to make the changes clearer.

If I'm right about how this works, the only change in behavior when a locale *is* defined is actually a slight bug fix. Suppose we're using an unusual encoding, like latin-1. In the case of any non-latin-1 characters in the args, they will be surrogate escaped for us already. Currently, we would then encode them to `bytes` with latin-1. [PEP-383](https://www.python.org/dev/peps/pep-0383/) tells us what will happen next:

> While providing a uniform API to non-decodable bytes, this interface has the limitation that chosen representation only "works" if the data get converted back to bytes with the surrogateescape error handler also. Encoding the data with the locale's encoding and the (default) strict error handler will raise an exception, encoding them with UTF-8 will produce non-sensical data.

So the `.encode()` call was missing a `surrogateescape` error handler! So this pull request probably fixes an unusual crash.

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -333,12 +333,7 @@ def arg_encoding():
     """Get the encoding for command-line arguments (and other OS
     locale-sensitive strings).
     """
-    try:
-        return locale.getdefaultlocale()[1] or 'utf-8'
-    except ValueError:
-        # Invalid locale environment variable setting. To avoid
-        # failing entirely for no good reason, assume UTF-8.
-        return 'utf-8'
+    return locale.getpreferredencoding()
 
 
 def _fsencoding():
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -826,7 +826,9 @@ def command_output(cmd, shell=False):
     This replaces `subprocess.check_output` which can have problems if lots of
     output is sent to stderr.
     """
-    cmd = convert_command_args(cmd)
+    # On Python 3, we now pass bytes directly to subprocess
+    if six.PY2:
+        cmd = convert_command_args(cmd)
 
     try:  # python >= 3.3
         devnull = subprocess.DEVNULL
@@ -843,6 +845,9 @@ def command_output(cmd, shell=False):
     )
     stdout, stderr = proc.communicate()
     if proc.returncode:
+        # convert bytes (CalledProcessError doesn't accept bytes)
+        if not six.PY2 and type(cmd[0]) == bytes:
+            cmd = [x.decode(arg_encoding()) for x in cmd]
         raise subprocess.CalledProcessError(
             returncode=proc.returncode,
             cmd=' '.join(cmd),


afontenot:
Interesting. I don't see why you *couldn't* do that, although I also don't see the point, given that you're decoding back to strings anyway before you pass to `subprocess`. It's also currently not true! The arguments are bytes on Windows too. The only difference is that on Windows the arguments are decoded [differently](https://github.com/beetbox/beets/blob/master/beetsplug/convert.py#L214) before we re-encode them to `bytes` and pass that to `command_output`. So if you really want it to work like that, it's got to be changed.

If you don't want to remove any of the encoding / decoding we're doing currently, check out [this commit](https://github.com/beetbox/beets/pull/3366/commits/47f88ffda3e6bfc69c561cb6472eca3b4c2127b2), since that's actually the way I was doing it originally before I noticed the steps could be simplified. Note that the `surrogateencoding` handler has to be added in that version, as I mentioned in the first comment on this pull request

The only concern is that this leaves a lot of encoding and decoding in place for no apparent reason. My description above actually left out two steps before step 1. First the file path is pulled from the database in bytes, then it's decoded to a string, then it's encoded to bytes again, then passed to `command_output`, which decodes it to a string again and passes it to `subprocess`, which encodes it again. So there are two encoding and decoding steps between the database and the OS. Surely if not both, at least one of these encoding / decoding pairs could be removed without changing the behavior? 

There are at least four options:

1. Don't remove any encoding / decoding. My first commit, see above.
2. Don't encode to `bytes` for `command_output`, and don't try to decode in `command_output`. This is what I'm doing now.
3. Don't do any encoding / decoding in `convert.py`, instead passing `bytes` from the database to `command_output`, where they will be decoded and then sent as strings to `subprocess` (which encodes them again). 
4. Don't do any encoding / decoding anywhere (except maybe for Windows); just pass `bytes` straight from the database directly to `subprocess`.

sampsyo:
> If you don't want to remove any of the encoding / decoding we're doing currently,

To clarify the motivation here, I'm not trying intrinsically to preserve a bunch of encoding/decoding. The idea is to keep the abstractions as simple and consistent as possible to avoid mistakes. Filename and argument encoding sadness is probably the most common source of bugs in beets by far, so it really helps to have a simple policy that we can apply everywhere. Even if that leads to inefficient repeated format changes, it's worth it if it makes the code more obviously correct. Put simply, I'm trying to prioritize maintainability for the entire beets codebase over just getting this one path to work in isolation.

> First the file path is pulled from the database in bytes, then it's decoded to a string, then it's encoded to bytes again, then passed to `command_output`, which decodes it to a string again and passes it to `subprocess`, which encodes it again.

Indeed; this is a useful summary. And I'd argue that the first decode/encode step is actually required—there's no way to get a correct implementation without it. The reason is that the conversion command is given to us as a text string and the paths are given as bytes, and we need to interpolate the latter into the former. I don't see how to do the interpolation correctly in the bytes world, so it has to happen in the string world. Then we have a string with surrogate-escaped path characters in it.

Anyway, we could make an argument that, on Python 3, properly surrogate-escaped Unicode strings should be allowed to go _straight_ to the OS from there. But the maintainability problem I'm worried about is that this requires that the `str` objects that get passed to APIs like `command_output` have to be a *very particular kind* of string to avoid crashes—they need to be pre-vetted strings that the client has somehow proven are free of characters that aren't representable in system's encoding. That's not true of arbitrary `str` objects—it's a pretty funky contract to put in the docstring, and it's not clear how client code would comply!

(As an aside, I think this is generally a design flaw with Python 3's OS APIs, which have the same problem. Everything works in simple cases, but seemingly-correct code starts to break when you use more complicated strings or run on a machine with unfortunate locale settings. Not too different from how things were in the bad old days of Python 2…)

Anyway, that's why I'm advocating for—at least on Unix—requiring that the input commands be bytes. This requires client code to handle the conversion and the errors it can produce themselves, rather than naively passing strings and creating "time bombs" that will turn into bugs somewhere down the line.

afontenot:
> Anyway, that's why I'm advocating for—at least on Unix—requiring that the input commands be bytes. This requires client code to handle the conversion and the errors it can produce themselves, rather than naively passing strings and creating "time bombs" that will turn into bugs somewhere down the line.

I see. So the issue is that you're thinking of `command_output` as an API that doesn't make assumptions about its arguments being properly encoded strings. I'm still not certain that I understand the value of this, since it's not necessarily more difficult to send correctly escaped strings than it is to send correctly encoded bytes (in fact, here, we screw both of them up), but I'll take your word for it.

In that case, the solution is to make sure that `command_output` escapes the args for the same encoding that `subprocess` will use to encode them. Since it's currently defaulting to decoding with UTF-8, characters not in the system's encoding aren't getting escaped, causing the exception I reported.

I wonder if it would also be better to use UTF-8 everywhere internally, except for the last step when when we create the strings sent to `subprocess`. Currently we're using `util.arg_encoding()` all over the place in `convert.py` when it doesn't seem to be really, since we only send bytes to `command_output` and that is *supposed* to escape our strings correctly for us.

At any rate, could you have a look at [this commit](https://github.com/beetbox/beets/pull/3366/commits/47f88ffda3e6bfc69c561cb6472eca3b4c2127b2) and merge that if you think it's fine? I can also fix up this pull request to do that if you prefer.

sampsyo:
> Edit: one more thought. If the `command_output` function assumes that its input args are correctly encoded bytes, why bother decoding them at all? `subprocess` accepts byte inputs, so just send everything directly from there!

Indeed; I think this might be the thing to do. I wish I had left myself a comment, because I don't currently recall why the function is decoding things before sending them to `subprocess`—just sending the raw bytes seems more natural and less prone to errors. Maybe we should give that a try and see if anything breaks?

jtpavlock:
@afontenot are you still interested in continuing work on this?

afontenot:
> @afontenot are you still interested in continuing work on this?

Yeah - life is a bit crazy for me at the moment, so it'll depend on how much time I have. I think my interest in fixing it last year kind of stalled out on what seemed to be a lack of consensus on the appropriate way to fix the problem. I certainly don't look at the beets codebase very often, so if I'm going to get back into this, maybe we could decide on what an acceptable path to go forward is? 

For example, is sending the bytes input to `command_output` directly to `subprocess` an acceptable fix for this issue, provided that no one runs into any major issues with it before the next release is cut?

afontenot:
Hey @sampsyo @jtpavlock I've completed a version of this that should do what we want here, passing bytes (on Python 3 at least) directly from `util.command_output` to `subprocess`. 

Interestingly, as of Python 3.7 (due to some major internal changes in Python) passing a string starts working again, see [here](https://vstinner.github.io/posix-locale.html). I think we're still safer passing bytes, though, and it's also mostly likely not working on pre-3.7 platforms. This does limit my ability to test the changes, just about every system I use regularly is already on Python 3.8. 

As I write this, it looks like some tests are failing. I believe they're checking the type of the passed value from `util.command_output`, so I'll need to fix them up for this pull request. I'll try to work that in over the next several days.

Let me know if you have any thoughts about this pull request.

sampsyo:
Cool; thanks for pushing this forward! Can you perhaps remind me why the change to `arg_encoding` is necessary & correct? Do we know that `getpreferredencoding` is the "right" encoding to use for CLI arguments, for example?

afontenot:
> Cool; thanks for pushing this forward! Can you perhaps remind me why the change to `arg_encoding` is necessary & correct? Do we know that `getpreferredencoding` is the "right" encoding to use for CLI arguments, for example?

You can definitely merge this without that change if you like, but I think it's a good improvement. The reasoning behind this is that we're decoding byte arguments to strings, and then when you pass the arguments to `subprocess`, it's encoding them back to bytes. The critical thing is that `subprocess` is using `getpreferredencoding` internally (whether explicitly or implicitly I don't recall), which means that if we use it when decoding to a string, we're *guaranteed* that that string will be encodable by `subprocess`. For example:

On a system without a locale set (prior to Python 3.7), `locale.getdefaultlocale()[1]` will be `None` and so the previous code will fall back to using utf-8. But `subprocess` will use `getpreferredencoding`, which (prior to Python 3.7) will return 'ascii'! That's what resulted in the original problem.

Note: it's probably possible to entirely remove the `arg_encoding` function, at least after some more testing with Python 2. If we *only* use that function to determine the encoding for arguments sent to `subprocess`, and we switch to *always* sending raw bytes to `subprocess`, there's no point for this function any more. I left it in because I took a more conservative route for now.

afontenot:
Well, that sucks. Looks like Python on Windows only got support for bytes arguments as of 3.8. That means that the possible solutions are:

1. Special case Windows: decode to string before passing to `subprocess` on both Windows and Python 2.

2. Go back to sending strings to `subprocess` on every platform, but keep the `getpreferredencoding` change. This will mean that we end up with ascii strings on pre 3.7 platforms that don't have a locale set. This might turn out to work okay, but we'd have to be extremely careful to make sure that surrogateescaping is handled consistently. 

Thoughts?

sampsyo:
Argh, that's pretty nasty! I don't have a strong opinion here and you might know better, but my instinct is leading me toward option 1, which preserves my intuition that "`str` is better on Windows; `bytes` is better on Unix" for such things…

Issue #3340, opened 2019-08-03
lyrics: Use MusicBrainz artist aliases if default query is not found


octos:
### Use case

I just noticed that the Lyrics plugin fails to fetch lyrics to [CHVRCHΞS](https://musicbrainz.org/artist/6a93afbb-257f-4166-b389-9f2a1e5c5df8) songs because the artist name contains a Greek uppercase Xi (Ξ), whereas musixmatch.com and lyrics.wikia.com URLs are presumably limited to ASCII characters.

```
lyrics: failed to fetch: http://lyrics.wikia.com/Chvrch%CE%BEs:Forever (404)
lyrics: failed to fetch: https://www.musixmatch.com/lyrics/Chvrch%CE%BEs/Forever (404)
```

### Solution

One solution would be for the Lyrics plugin to cycle through the artist’s [MuzicBrainz aliases](https://musicbrainz.org/artist/6a93afbb-257f-4166-b389-9f2a1e5c5df8/aliases) if it hits a 404 error.

### See also

https://discourse.beets.io/t/asciiized-filenames-for-fetching-lyrics/855

sampsyo:
Sounds good! It's worth noting that the plugin already does a few things like this for normalizing typography, but using the MB aliases would be a good idea too.

Moonbase59:
Now that _is_ an interesting example, since I also have a Chruches album. I mainly use Picard for tagging, and the have two interesting settings under _Options → Metadata_:

[x] Translate artist names to this locale when possible
    [locale dropdown, I selected _English_]
[ ] Use standardized artist names

When I select "Translate", I get "CHVRCHES", when I switch it off, it stays "CHVRCHΞS".
Selecting "Use standardized" seems to change nothing (i.e., it stays "CHVRCHΞS").

Btw, (Album) Artist Sort Order seems always to keep whatever the editor put in "Sort name".

Since Picard is also in Python, it might be worth spelunking in their code to find out how they do it.

hashhar:
I think the better answer is using the sort fields since MusicBrainz encourages (or maybe enforces, not sure) using only ASCII for the sort fields to improve compatibility.

If nobody is currently working on this, I can pick this up after I'm done with https://github.com/beetbox/beets/issues/3549.

[...]

PR #3380, opened 2019-09-30
add azlyrics to lyrics plugin

matlads:
Add azlyrics as a source backend for lyrics

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -404,6 +404,24 @@ class Genius(Backend):
             return self.lyrics_from_song_api_path(song_api_path)
 
 
+class Azlyrics(Backend):
+    URL_PATTERN = "https://www.azlyrics.com/lyrics/%s/%s.html"
+
+    def build_url(self, artist, title):
+        a = self._encode(artist.lower().replace(' ', ''))
+        t = self._encode(title.lower().replace(' ', ''))
+        return self.URL_PATTERN % (a, t)
+
+    def fetch(self, artist, title):
+        url = self.build_url(artist, title)
+        html = self.fetch_url(url)
+        if not html:
+            return
+        lyrics = extract_text_between(html, "<!-- Usage of azlyrics.com content by any third-party lyrics provider is prohibited by our licensing agreement. Sorry about that. -->", "</div")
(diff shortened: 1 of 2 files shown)


sampsyo:
Nice; looks good! I think Travis is complaining about a too-long line, which includes that HTML comment. Any chance there's a shorter way to match on that? Otherwise, you can pull it out into a multi-line string literal constant.

Also, would you mind adding a quick entry to the changelog about your addition?

Moonbase59:
Did you try some more exotic artists and titles? I think maybe they might need some conversion, don’t know how much beets would do beforehand.

Take "AC/DC" with "Thunderstruck" (→ "acdc", "thunderstruck").
AZLyrics URI: https://www.azlyrics.com/lyrics/acdc/thunderstruck.html

jackwilsdon:
It seems in certain cases that azlyrics switches to using IDs instead (see https://www.azlyrics.com/lyrics/hatters/810433.html) - I'm not sure how much we can do here except possibly look into using their search feature?

It seems like removing all non alphanumeric characters is the safest bet in regards to things such as "AC/DC".

Moonbase59:
See https://www.azlyrics.com/lyrics/massiv/ichhabemeinetraumewahrgemacht.html ("Ich Habe Meine Träume Wahr Gemacht").

Looks like they somehow ASCII-fy and then remove special characters, reverting to IDs for non-ASCII languages. Without knowing their code, probably not easy to hack.

sampsyo:
Indeed—this is sort of a cross-cutting concern across all the lyrics backends. See also #3340. Because it will never be possible to do this perfectly, I think it's fine to merge this as is (pending a line-length fix and a changelog entry) and incrementally improve the heuristic in the future.

jtpavlock:
Is there evidence this provides any worthwhile additions to the current lyrics sources?

jtpavlock:
A non-ideal solution could be to lookup the artist's page using the first letter of their (sorted) name and then using azlyric's alphabetized pages e.g. https://www.azlyrics.com/a.html for all artists starting the "A". Scrape the page for the matching artist, follow the link, then scrape that page for the matching song, follow the link, then finally scrape the lyrics.

Issue #3368, opened 2019-09-14
replaygain: Add a loudgain backend


Moonbase59:
### Use case

As the current developer of [loudgain](https://github.com/Moonbase59/loudgain), a ReplayGain 2.0 tagging tool for many file formats, I’d much like to see loudgain being supported by beets as an official ReplayGain backend.

### Solution

I assume integration shouldn’t be too complicated, since loudgain uses an enhanced version of the well-known _mp3gain_ commandline syntax.

Let me know how I could help getting this going.
Any testing results from beets users and feedback on loudgain’s [issue tracker](https://github.com/Moonbase59/loudgain/issues) welcome!

### More info

[loudgain on GitHub](https://github.com/Moonbase59/loudgain)
[loudgain in the Hydrogenaudio wiki](http://wiki.hydrogenaud.io/index.php?title=Loudgain)

sampsyo:
Looks cool! I am really not an expert in this space, so is there any chance you could expand a little on the advantages of your new tool vs. using FFmpeg's gain scanning module?

Moonbase59:
Good question, I never much used the FFmpeg gain scanner, albeit I do use the FFmpeg libraries to be able to easily read and decode many file types. In loudgain, RG2 calculation is done purely by using `libebur128` to get at correct gain and _true peak_ values, as well as (optional) dynamic range. It also allows clipping prevention according to the EBU R128 recommendation (at -1 dBTP), and other options like writing upper- or lowercase RG tags, this unfortunately still being a problem with many players (i.e. KODI requiring lower case, VLC requiring uppercase and so on).

Oh well, and it uses an enhanced version of the mp3gain commandline syntax, never changes audio data (it only writes tags), and handles a lot of commonly used file types, using the same piece of software and commandline options.

Don’t want to make this an ad—much more tech details can be found on loudgain’s GitHub page. :-)

sampsyo:
Got it. One thing you should know is that beets ReplayGain backends *do not* write anything to files; beets itself takes care of all the metadata. So we would not use the tagging features of loudgain, including format-specific casing and stripping existing metadata.

The ReplayGain plugin already has a `noclip` setting (for the mp3gain backend), so we could provide that option to the tool. For the FFmpeg backend, we also already have a true vs. sample peak config option.

[...]

PR #3381, opened 2019-09-30
Add extra replaygain fields from loudgain

Moonbase59:
As [discussed in forum](https://discourse.beets.io/t/better-replaygain-2-loudgain-support/947), here are the new MediaFile fields to make magic happen in beets.


jtpavlock:
I'm confused, is `loudgain` supported by beets, or is this PR attempting to preemptively add fields for an unsupported backend component?

sampsyo:
Indeed; as I mentioned recently in #3368, I've sort of lost the context here.

jtpavlock:
In that case I'm closing this for now. I think it would also make sense to couple this change with the actual implementation of `loudgain`

jtpavlock:
Looks like https://github.com/beetbox/mediafile/pull/19 is relevant, which provides the implementation of these fields, but still doesn't make sense unless we actually switch or add `loudgain` as a replay gain backend.

PR #3371, opened 2019-09-16
Integrate MetadataSourcePlugin

rhlahuja:
Refactor the Spotify and Discogs plugins to use `beets.plugins.MetadataSourcePlugin` base class logic (introduced in https://github.com/beetbox/beets/pull/3355). If preferred, I can create separate PRs for changes to the Discogs and Deezer plugins, instead of including them here.

PR #3385, opened 2019-10-01
docs/plugins/discogs.rst: Update plugin documentation

allesmi:
I have found that the documentation of DiscogsPlugin does not mention the 'source_weight' option. This PR adds that information:

Add Configuration section and describe the 'source_weight' option.

### DIFF
--- a/docs/plugins/beatport.rst
+++ b/docs/plugins/beatport.rst
@@ -36,6 +36,11 @@ also search for an id like so:
 
     beet import path/to/music/library --search-id id
 
+Configuration
+-------------
+
+This plugin can be configured like other metadata source plugins as described in :ref:`metadata-source-plugin-configuration`.
+
 .. _requests: https://docs.python-requests.org/en/latest/
 .. _requests_oauthlib: https://github.com/requests/requests-oauthlib
 .. _Beatport: https://beetport.com
--- a/docs/plugins/deezer.rst
+++ b/docs/plugins/deezer.rst
@@ -22,13 +22,4 @@ prompt during import::
 Configuration
 -------------
 
-Put these options in config.yaml under the ``deezer:`` section:
-
-- **source_weight**: Penalty applied to Deezer matches during import. Set to
-  0.0 to disable.
-  Default: ``0.5``.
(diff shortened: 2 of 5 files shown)


jackwilsdon:
Thanks for this change! It seems like this is missing from `docs/plugins/beatport.rst` too - would you mind adding it in there as well?

allesmi:
I can certainly add it to the documentation of the beatport plugin as well.
But in the future more plugins might follow and we risk that documentation
of plugins drifts apart.

The only mention of this option in the docs is in the SpotifyPlugin but it
should apply to every plugin that provides a source of metadata. Is there a
central place in the docs to mention this?

Issues #3355 and #3371 extracted a lot of code into the superclass
MetadataSourcePlugin. Documentation could follow and describe
'source_weight' in only one place in the docs. What do you think?

sampsyo:
Good question. It would be nice to centralize this information, especially because it's a pretty specialized thing that most people won't want to worry about—and if you do, then you'll worry about it _in concert_ between multiple plugins, not just by configuring a single plugin at a time.

I don't have a perfect idea about where that would be, however. Maybe on the main page about using plugins in general? I'd be open to suggestions.

allesmi:
How about a section under Using Plugins in docs/plugins/index.rst that describes configuration options common to all plugins that provide a metadata source. Then, for each individual plugin a section about configuration that starts with a reference to first section followed by additional configurations.

I think of something like this:

> ### Configuration
> 
> This plugin provides a source of metadata and shares configuration options as described in [Using MetadataSourcePlugins](link-to-metadata-source-plugins). In addition, it provides the following configuration options:
> - **Option A**: Description of option A
> - **Option B**: Description of option B.

sampsyo:
That's a great idea! Yeah, linking to the section seems very good.

allesmi:
I have added the description of the source_weight option to docs/plugins/index.rst and updated the plugin documentation with references back to that description for beatport, deezer, discogs, and spotify.

sampsyo:
This looks lovely! Thank you!!!

Issue #3387, opened 2019-10-01
Beatport plugin does not apply tags to file


temrix:
### Problem

Continuation of issue #2080 
The beatport plugin fetches genres, bpm and musical_key among others. These should be shown when querying using `beat ls ...` and written into the tags of the music files. However, `genre` is not applied, `bpm` is correctly applied and written to tags and `musical_key` is applied but not written to tags.

During writing this bug report I found that I did not pass the attribute `self.genre` to the hook. This explains why `genre` was not applied but not why `musical_key` is not written to tags.

I will issue a pull request resolving the first bug but I don't know why `musical_key_ is not written to the tags.

### Reproducing

Deleting with a tag editor `genre` and `bpm` beforehand.

First, we import the album in timid (`-t`) mode:

```sh
$ beet -vv import ~/music -t
```

Output:

```
user configuration: /home/temrix/.config/beets/config.yaml
data directory: /home/temrix/.config/beets
plugin paths: 
Sending event: pluginload
library database: /home/temrix/Dokumente/musiclibrary.db
library directory: /home/temrix/music
Sending event: library_opened
Sending event: import_begin
Sending event: import_task_created
Sending event: import_task_start
Looking up: /home/temrix/music/Supersillyus - Charade - 2016 - MP3
Tagging Supersillyus - Charade
No album ID found.
Search terms: Supersillyus - Charade
Album might be VA: False
Searching for MusicBrainz releases with: {'release': 'charade', 'artist': 'supersillyus', 'tracks': '6'}
Requesting MusicBrainz release 311af76c-3c77-4603-8b64-b4de17ee6c3a
primary MB release type: album
Sending event: albuminfo_received
Candidate: Supersillyus - Charade (311af76c-3c77-4603-8b64-b4de17ee6c3a)
Computing track assignment...
...done.
Success. Distance: 0.00
Requesting MusicBrainz release 8613ba08-2700-48bc-8a2c-349a81e37a31
primary MB release type: album
Sending event: albuminfo_received
Candidate: Supersillyus - Tesselations (8613ba08-2700-48bc-8a2c-349a81e37a31)
Computing track assignment...
...done.
Success. Distance: 0.70
Requesting MusicBrainz release 8ff531fd-198c-4188-9a60-0eff01591153
primary MB release type: album
Sending event: albuminfo_received
Candidate: Supersillyus - Interabang‽ (8ff531fd-198c-4188-9a60-0eff01591153)
Computing track assignment...
...done.
[...]

PR #3372, opened 2019-09-18
Fix #2080

temrix:
This fixes feature request #2080 .

I have added `bpm`, `musical_key` and `genre` in the beatport plugin. I needed to modify the hooks in the autotagger to accept those additional attributes. `genre` will be fetched for each track. I have also added a test file to test the plugin.

PR #3389, opened 2019-10-04
Add BPSyncPlugin

rhlahuja:
This is a proof-of-concept for a `bpsync` plugin, Beatport's equivalent of `mbsync`. It allows users to sync changes to Beatport metadata and "backfill" new fields recently added in https://github.com/beetbox/beets/pull/3372.

I'm considering abstracting this functionality into a `SyncMetadataSourcePlugin` base class (similar to `MetadataSourcePlugin` introduced in https://github.com/beetbox/beets/pull/3355) which can also be leveraged for Spotify/Deezer/MusicBrainz, although that requires some larger refactors of `BeatportPlugin` and `MBSyncPlugin`, perhaps better suited for a separate PR? 

@sampsyo, what do you think? Is it worth putting the finishing touches on this Beatport-specific plugin, or investing time in a `SyncMetadataSourcePlugin` to future-proof?

This PR also closes https://github.com/beetbox/beets/issues/3387 by properly assigning `genre` and renaming `musical_key` to `initial_key`.

Note: I had to run
```sql
UPDATE item_attributes SET key='initial_key' WHERE key='musical_key'
```
on my database to avoid raising ```AttributeError: 'TrackInfo' object has no attribute 'musical_key'``` when `autotag.apply_metadata(album_info, mapping)` is called.

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -1648,20 +1648,6 @@ class DefaultTemplateFunctions(object):
             return falseval
 
 
-def apply_item_changes(lib, item, move, pretend, write):
-    """Store, move and write the item according to the arguments.
-    """
-    if pretend:
-        return
-
-    # Move the item if it's in the library.
-    if move and lib.directory in util.ancestry(item.path):
-        item.move(with_album=False)
-
-    if write:
-        item.try_write()
-    item.store()
-
 # Get the name of tmpl_* functions in the above class.
 DefaultTemplateFunctions._func_names = \
     [s for s in dir(DefaultTemplateFunctions)
--- a/beets/library.py
+++ b/beets/library.py
@@ -1651,14 +1651,15 @@ class DefaultTemplateFunctions(object):
 def apply_item_changes(lib, item, move, pretend, write):
     """Store, move and write the item according to the arguments.
     """
-    if not pretend:
-        # Move the item if it's in the library.
-        if move and lib.directory in util.ancestry(item.path):
-            item.move(with_album=False)
+    if pretend:
+        return
+    # Move the item if it's in the library.
+    if move and lib.directory in util.ancestry(item.path):
+        item.move(with_album=False)
 
-        if write:
-            item.try_write()
-        item.store()
+    if write:
+        item.try_write()
+    item.store()
 
 # Get the name of tmpl_* functions in the above class.
 DefaultTemplateFunctions._func_names = \
--- a/beets/library.py
+++ b/beets/library.py
@@ -1648,6 +1648,18 @@ class DefaultTemplateFunctions(object):
             return falseval
 
 
+def apply_item_changes(lib, item, move, pretend, write):
+    """Store, move and write the item according to the arguments.
+    """
+    if not pretend:
+        # Move the item if it's in the library.
+        if move and lib.directory in util.ancestry(item.path):
+            item.move(with_album=False)
+
+        if write:
+            item.try_write()
+        item.store()
+
 # Get the name of tmpl_* functions in the above class.
 DefaultTemplateFunctions._func_names = \
     [s for s in dir(DefaultTemplateFunctions)
--- a/beets/library.py
+++ b/beets/library.py
@@ -1653,6 +1653,7 @@ def apply_item_changes(lib, item, move, pretend, write):
     """
     if pretend:
         return
+
     # Move the item if it's in the library.
     if move and lib.directory in util.ancestry(item.path):
         item.move(with_album=False)
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -26,7 +26,7 @@ from functools import wraps
 
 
 import beets
-from beets import logging
+from beets import library, logging, ui, util
 import mediafile
 import six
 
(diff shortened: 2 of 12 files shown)


sampsyo:
Nice!! This looks great. I do think it would be a good idea to land this self-contained version—and then consider factoring out another base class, if that makes sense, in a second pass.

rhlahuja:
@sampsyo, this is ready for a review whenever you're free!

beets/library.py
```diff
@@ -1648,6 +1648,20 @@ def tmpl_ifdef(self, field, trueval=u'', falseval=u''):
             return falseval
 
 
+def apply_item_changes(lib, item, move, pretend, write):
+    """Store, move and write the item according to the arguments.
+    """
```
sampsyo: This function seems useful, but it's a little out of place—for one thing, it's in between the definition of the default template functions and the little computation that gathers them up into a list. It's also mostly relevant to plugin actions, where this whole "pretend" notion has meaning, for example. What do you think about moving this to the `plugins` module?

This docstring could also use a little expansion to explain what "move" and "pretend" do, for example. I realize that you didn't write this docstring, but now might be a good opportunity to improve it. :smiley:

beets/plugins.py
```diff
@@ -649,6 +649,8 @@ def get_artist(artists, id_key='id', name_key='name'):
             if not artist_id:
                 artist_id = artist[id_key]
             name = artist[name_key]
+            # Strip disambiguation number.
+            name = re.sub(r' \(\d+\)$', '', name)
```
sampsyo: Maybe it's worth mentioning these normalization steps in the docstring?

beets/plugins.py
```diff
@@ -635,11 +635,11 @@ def get_artist(artists, id_key='id', name_key='name'):
 
         :param artists: Iterable of artist dicts returned by API.
         :type artists: list[dict]
-        :param id_key: Key corresponding to ``artist_id`` value.
-        :type id_key: str
-        :param name_key: Keys corresponding to values to concatenate
+        :param id_key: Key or index corresponding to ``artist_id`` value.
+        :type id_key: str or int
+        :param name_key: Key or index corresponding to values to concatenate
             for ``artist``.
```
sampsyo: Seems like this description could use a little more detail. What exactly is `artist`, for example? (I think it's an element of the `artists` list.) Furthermore, it sounds like the `artists` list can actually contain dicts _or_ lists—is that right?

docs/plugins/bpsync.rst
```diff
@@ -0,0 +1,37 @@
...
+=============
+
+This plugin provides the ``bpsync`` command, which lets you fetch metadata
+from Beatport for albums and tracks that already have Beatport IDs. This
+is useful for updating tags as they are fixed in the Beatport database, or
+when you change your mind about some config options that change how tags are
+written to files. If you have a music library that is already nicely tagged by
+a program that also uses Beatport, this can speed up the initial import if you
+just import "as-is" and then use ``bpsync`` to get up-to-date tags that are written
+to the files according to your beets configuration.
```
sampsyo: This last argument is somewhat less relevant for Beatport than it is for MusicBrainz. There are lots of other MB taggers out there (e.g., Picard), but I don't know of any other Beatport-based metadata taggers. Maybe it should reference that you might have downloaded your music from Beatport in the first place?

docs/plugins/bpsync.rst
```diff
@@ -0,0 +1,37 @@
+BPSync Plugin
+=============
+
+This plugin provides the ``bpsync`` command, which lets you fetch metadata
```
sampsyo: Instead of duplicating the text from the MBSync plugin's docs, let's just briefly summarize what this plugin does and link to the MBSync page (by just saying "this plugin works similarly to…").

docs/plugins/bpsync.rst
```diff
@@ -0,0 +1,37 @@
...
+  ``-p`` (``--pretend``) flag.
+* By default, files will be moved (renamed) according to their metadata if
+  they are inside your beets library directory. To disable this, use the
+  ``-M`` (``--nomove``) command-line option.
+* If you have the ``import.write`` configuration option enabled, then this
+  plugin will write new metadata to files' tags. To disable this, use the
+  ``-W`` (``--nowrite``) option.
+* To customize the output of unrecognized items, use the ``-f``
+  (``--format``) option. The default output is ``format_item`` or
+  ``format_album`` for items and albums, respectively.
```
sampsyo: This option is not actually supported, AFAICT.

beets/plugins.py
rhlahuja: Good catch, the artist list can indeed contain dicts or lists. Docstring expanded and updated accordingly.

beets/library.py
rhlahuja: Good call! Docstring improved and moved to `plugins.` I was also considering moving it to the `autotag` module alongside `apply_item_metadata` and `apply_metadata`, although not sure what makes more sense.

beets/plugins.py
```diff
@@ -694,8 +733,9 @@ def candidates(self, items, artist, album, va_likely):
         query_filters = {'album': album}
         if not va_likely:
             query_filters['artist'] = artist
-        albums = self._search_api(query_type='album', filters=query_filters)
-        return [self.album_for_id(album_id=a['id']) for a in albums]
+        results = self._search_api(query_type='album', filters=query_filters)
+        albums = [self.album_for_id(album_id=r['id']) for r in results]
+        return [a for a in albums if a is not None]
```
rhlahuja: This bit logic along with [this](https://github.com/beetbox/beets/pull/3389/files#diff-f4ce7fc73415332dc2333633555c3f06R87) early return were added to exclude "empty" (0-track) albums I encountered such as https://www.deezer.com/us/album/6979394.

PR #3390, opened 2019-10-04
discogs: restore strip of disambiguation number from artist

kergoth:
The integration of MetadataSourcePlugin removed the strip, as it's
discogs specific, but converted discogs to use it, which is a problem.
Restore our own get_artist which makes the per-artist change we need.

### DIFF
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -302,7 +302,7 @@ class DiscogsPlugin(BeetsPlugin):
             self._log.warning(u"Release does not contain the required fields")
             return None
 
-        artist, artist_id = MetadataSourcePlugin.get_artist(
+        artist, artist_id = self.get_artist(
             [a.data for a in result.artists]
         )
         album = re.sub(r' +', ' ', result.title)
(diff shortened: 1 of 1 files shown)


kergoth:
Without this fix, we get entries in the candidates list like this (note the `(2)` suffix:
```
4. Patrick Watson (2) - Wooden Arms (49.1%) (missing tracks, source, ...) (Discogs, CD, 2009, Canada, Secret City Records, SCR012CD)                                        
5. Patrick Watson (2) - Wooden Arms (49.1%) (missing tracks, source, ...) (Discogs, CD, 2009, Benelux, V2, VVNL20762)                                                       
6. Patrick Watson (2) - Wooden Arms (49.1%) (missing tracks, source, ...) (Discogs, CD, 2009, France, Tôt Ou Tard, 8345 10659 2)
```

kergoth:
I'm not sure how generally useful this get_artist method is going to be, but if we think it is, we might want to separate the 'get a list of artists' from the join, to let the caller manipulate the individual artists rather than only getting the merged string.

rhlahuja:
Hey @kergoth you're right, I realized I mistakenly removed that bit of logic from but have added it back in https://github.com/beetbox/beets/pull/3389 over [here](https://github.com/beetbox/beets/pull/3389/files#diff-8d92a0d1b800610ff039cb71cf5cc055R653). That PR also integrates it in the Beatport plugin, since the logic is identical.

I don't anticipate this to be an issue for metadata sources which don't include disambiguation numbers, unless we actually encounter an artist named e.g. `Artist (1)`. Although if we do, I agree returning a list of artists makes more sense for the API.

kergoth:
Ah, excellent, I didn’t notice that one. Thanks!

PR #3394, opened 2019-10-09
Delete original files, if converting and import->move is true

### LINKED ISSUE #2947: convert: In `auto` mode, respect `import.move` by deleting old files
The relevant parts of my config:
```
import:
    move: yes

plugins: convert

convert:
    auto: yes
    max_bitrate: 192
    format: mp3
```
If the imported mp3 file has a bitrate under 192 it gets moved unconverted as desired.
If it's higher than that the old file remains inside the source folder after the successful conversion and the new file get's copied to the destination folder. It acts like the copy flag. The output shows:

> convert: Encoding D:\Test\Test.mp3
> convert: Finished encoding D:\Test\Test.mp3

logan-arens:
Fixes #2947.

I don't think there's much to say about this PR, since I only added four lines of code, but basically:
If files would normally be moved on import, and the user has chosen to convert files instead, then the original files will be deleted once they are converted and imported.

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -30,6 +30,7 @@ from beets import ui, util, plugins, config
 from beets.plugins import BeetsPlugin
 from confuse import ConfigTypeError
 from beets import art
+from beets.util import syspath
 from beets.util.artresizer import ArtResizer
 from beets.library import parse_query_string
 from beets.library import Item
@@ -532,10 +533,15 @@ class ConvertPlugin(BeetsPlugin):
 
             # Change the newly-imported database entry to point to the
             # converted file.
+            source_path = item.path
             item.path = dest
             item.write()
             item.read()  # Load new audio information data.
             item.store()
+            
+            # If we would normally move files, let's delete the source
+            if config['import']['move']:
+                util.remove(syspath(source_path), False)
 
     def _cleanup(self, task, session):
         for path in task.old_paths:


sampsyo:
Thanks for getting this started! This seems like the right thing overall.

Does anyone think there should be a plugin config option to override this behavior? Or will everyone want it?

beetsplug/convert.py
```diff
@@ -532,11 +533,16 @@ def convert_on_import(self, lib, item):
 
             # Change the newly-imported database entry to point to the
             # converted file.
+            source_path = item.path
             item.path = dest
             item.write()
             item.read()  # Load new audio information data.
             item.store()
 
+            # If we would normally move files, let's delete the source
```
sampsyo:
```suggestion
            # If we would normally move files, let's delete the source.
```

logan-arens:
The Travis CI build failed due to an apt-get versioning error, is there anything I can do about that on my end?

sampsyo:
Totally not your fault! Looks like this was an intermittent Travis failure that has since cleared up.

Any chance you could add a changelog entry describing the new behavior?

sampsyo:
One last thing: I'm just slightly nervous about this change because it involves deleting files. Especially in a situation where `convert` is configured to transcode from a lossless format to a lossy format on import, it could be surprising for some users. They'd see that their FLACs have been permanently destroyed and replaced with MP3s, never to be recovered again. This is different from the typical `move` behavior, which does indeed make the old files vanish, but at least no actual audio data is ever lost.

So it would be useful to think carefully about how to do make this behavior clear and avoid accidental data loss…

logan-arens:
> So it would be useful to think carefully about how to do make this behavior clear and avoid accidental data loss…

Would it be better to have this as a separate config parameter? We could have convert->delete->yes to specify this directly.

sampsyo:
Possibly! Another reasonable option might be to make the config option decide whether to "agree" with `import.move` or to override it and never delete files.

jtpavlock:
Was there any follow-up to this?

jtpavlock:
Closing since the op deleted their account

PR #3395, opened 2019-10-09
Allow disabling plugins with $BEETS_DISABLED_PLUGINS.

djl:
My rationale for adding this is somewhat selfish.

I have a handful of custom plugins that perform some CPU or IO heavy tasks (e.g. kick off a full Madsonic rescan/cleanup or a filesystem snapshot). If I don't want these to run then I need to disable these plugins in my config and I'm far too lazy for that!

Originally I just added support for `BEETS_DISABLED_PLUGINS` to each of my plugins but thought this *might* be useful to other people, especially for testing beets itself.

### DIFF
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -18,6 +18,7 @@
 from __future__ import division, absolute_import, print_function
 
 import traceback
+import os
 import re
 import inspect
 import abc
@@ -268,7 +269,12 @@ def load_plugins(names=()):
     package in sys.path; the module indicated should contain the
     BeetsPlugin subclasses desired.
     """
+    disabled_plugins = os.environ.get('BEETS_DISABLED_PLUGINS', '').split(',')
     for name in names:
+        if name in disabled_plugins:
+            log.debug(u'Skipping disabled plugin {0}', name)
+            continue
+
         modname = '{0}.{1}'.format(PLUGIN_NAMESPACE, name)
         try:
             try:
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -80,6 +80,8 @@ New features:
 * :doc:`/plugins/beatport`: Fix assignment of `genre` and rename `musical_key`
   to `initial_key`.
   :bug:`3387`
+* Plugins can be disabled by setting the `BEETS_DISABLED_PLUGINS`
+  environment variable to a comma seperated list of plugins.
 
 Fixes:
 


sampsyo:
Not a bad idea! I can see this being useful to others. But just to check, have you considered having a custom "overlay" config file for this that has only a lightweight set of plugins? Then you can do `beet -c ~/beets-light.yaml` to switch to the faster mode.

djl:
That was my original solution before thinking of the environment variable. The problem with that approach is that it requires two nearly identical config files where the only difference is the plugin list. Changes to the "main" config file would need to duplicated/synced to the "overlay" file.

With the environment variable I have one config file and can easily have different plugins enabled just changing the list, like so:

    # bq == "beets quick"
    alias bq=BEETS_DISABLED_PLUGINS=mpdupdate,sonic,snapshots,rsync beet
    # bqq == "beets quite quick"
    alias bqq=BEETS_DISABLED_PLUGINS=mpdupdate beet

sampsyo:
Sounds about right, but I don't think you actually have to duplicate everything—the `-c` thing actually works by overriding _just_ what you define. So an auxiliary config file could _only_ contain a `plugins:` line, and all your other options would be preserved from your default config. (Unless I'm forgetting something subtle about how `-c` works…)

djl:
You're correct. I was mistaken about how the `-c` works but the problem of duplication still exists. Instead of a whole config file it's just `plugins:` section.

It's less duplication but more work for the end user. Now instead of a single config file that can be easily overridden on the fly, it's `N` config files which need to be kept in sync with the "main" config file (well, the `plugins:` line anyway).

`-c` would be great if someone wanted to disable *all* plugins except for a few. My use case is the opposite - I want to keep my ~40 plugins enabled and only disable a select few.

sampsyo:
Sounds reasonable! Thanks for talking though it. :smiley:

How about this: why don't we see if we can convince at least one other person that this would be a useful feature for their workflow? If we can find a second person who things "that sounds cool; I'd probably use it!" then let's add it. Maybe a post on the Discourse forum would help drum up such a person?

It also seems like this variable should probably appear in the CLI documentation, but I'm not entirely sure where.

djl:
> How about this: why don't we see if we can convince at least one other person that this would be a useful feature for their workflow? If we can find a second person who things "that sounds cool; I'd probably use it!" then let's add it. Maybe a post on the Discourse forum would help drum up such a person?

Sounds good to me!

Discourse is being awkward. It seems to hang when trying to post a new topic. I'll try again shortly.

jtpavlock:
Was there any follow-up to this?

djl:
Couldn't (and maybe still can't) post to Discourse :(

For now I've went back to baking this functionality into my own plugins.

jtpavlock:
I have to admit this is quite niche and the implementation, while simple, seems a bit confusing to the user. Let me try to explain. 

The idea that for or an alternate config whitelist, you should use `--config`, but for a blacklist, you should pass an environment variable doesn't sit right. That's two completely different ways to solve to a similar issue.

Ideally, tackling those two problems would be as similar as the problems themselves. Now, @sampsyo can weigh in more here, but if this were to be implemented, I'd like to see it in the config file somehow. Perhaps as a `!plugins:` line or any plugin listed as `!plugin` would be disabled.

djl:
> I have to admit this is quite niche and the implementation, while simple, seems a bit confusing to the user. Let me try to explain.
> 
> The idea that for or an alternate config whitelist, you should use `--config`, but for a blacklist, you should pass an environment variable doesn't sit right. That's two completely different ways to solve to a similar issue.
> 
> Ideally, tackling those two problems would be as similar as the problems themselves. Now, @sampsyo can weigh in more here, but if this were to be implemented, I'd like to see it in the config file somehow. Perhaps as a `!plugins:` line or any plugin listed as `!plugin` would be disabled, etc.

That's a neat idea and probably a good discussion for another issue.

Since that idea is better than mine (and my personal implementation works fine for now) I'm happy to close this.

PR #3397, opened 2019-10-10
Remove albums instead of tracks when removing duplicates

### LINKED ISSUE #2074: Remove art when removing album
When importing a higher quality version of an album already in my library, beet offered to remove the old version, which is very handy. But it left behind the old album's directory with the cover file in it. I would have expected beet to remove it completely.

logan-arens:
Fixes #2074.

Previously, remove_duplicates would remove each Item object (track) one by one, which would leave empty folders, or folders with just cover art.

Now I've replaced the duplicate_items function with duplicate_albums, so that it returns a list of Album objects rather than Item objects, and I changed remove_duplicates to remove albums instead of tracks.

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -552,7 +552,6 @@ class ImportTask(BaseImportTask):
         duplicate_albums = self.duplicate_albums(lib)
         log.debug(u'removing {0} old duplicated albums', len(duplicate_albums))
         for album in duplicate_albums:
-            print(album)
             album.remove(delete=True)
 
     def set_fields(self):
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -542,23 +542,18 @@ class ImportTask(BaseImportTask):
 
         autotag.apply_metadata(self.match.info, self.match.mapping)
 
-    def duplicate_items(self, lib):
-        duplicate_items = []
+    def duplicate_albums(self, lib):
+        duplicate_albums = []
         for album in self.find_duplicates(lib):
-            duplicate_items += album.items()
-        return duplicate_items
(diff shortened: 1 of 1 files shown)


logan-arens:
Sorry, for some reason the GitHub search showed me that duplicate_items was only used in that one place, but it was actually used elsewhere. I've added the code for it back in.

sampsyo:
Good catch! Thanks for looking into this.

There is a catch here, however—have you tested this with singleton imports? Those can also have duplicates, and I think when calling `remove_duplicates` on a `SingletonImportTask`, the invocation of `duplicate_albums` will likely crash.

logan-arens:
I haven't tested this with singletons, because I've never actually seen one in my beets library so I'm unsure how they work.

I used to think that importing a single track would result in a singleton, but at least in my case, that ends up creating the usual album directory structure instead of putting it into the Singletons folder like defined in my config. Is there some prerequisite for something to be classified as a singleton?

(Everything in my library is in MusicBrainz, or I add it to MusicBrainz if it is missing and then import, so I'm not sure if singletons only appear when you do "Use As Is")

sampsyo:
Yes! You can pass the `--singleton` flag to the `beet import` command. Or, the "as Tracks" interactive option in the importer temporarily switches to singleton mode.

jtpavlock:
@ghost are you still interested in continuing work on this?

jackwilsdon:
Looks like @ghost is a deleted user - I think this can be closed as the original creator has deleted their GitHub account.

jtpavlock:
Ha, good catch

PR #3398, opened 2019-10-10
"beet update" now confirms that the library path exists before updating.

### LINKED ISSUE #1934: Beet update when volume not mounted
Hi

Maybe I'm not using the correct config, but if you `beet update` when your music files volume is not mounted, your library gets all deleted without further prompting.

Is that the expected behaviour or is there a way to avoid it?
Thank you

ghost:
Fixes #1934.

Previously, if the library folder was renamed, unavailable (such as an offline network share), or unmounted (if stored on an external drive), then "beet update" would wipe out the user's entire database.

As this is not the best result for an unfortunate soul who doesn't realize that they left their USB drive unplugged, I've added a prompt that appears when the library is missing. Users can choose to continue the update, or to cancel it.

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1188,6 +1188,7 @@ def update_func(lib, opts, args):
     # Verify that the library folder exists to prevent accidental wipes.
     if not os.path.exists(lib.directory):
         ui.print_("Library path is unavailable or does not exist.")
+        ui.print_(lib.directory)
         if not ui.input_yn("Are you sure you want to continue (y/n)?", True):
             return
     update_items(lib, decargs(args), opts.album, ui.should_move(opts.move),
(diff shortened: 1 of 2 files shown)


sampsyo:
Sounds great! I have a few random suggestions:

- How about `isdir` instead of `exists`, just in case the path leads to a file instead of a directory?
- What do you think about printing out the library directory, to make it easier for the user to see why it might not exist?
- Would you mind adding a quick note to the changelog describing this new behavior?

logan-arens:
Is there any case where the library path would lead to a file instead of a directory?

sampsyo:
Sure—maybe an errant symlink, perhaps? I don't expect it to be _terribly_ common, but it seems like the more defensive check certainly couldn't hurt. :smiley:

logan-arens:
Simple enough :D

sampsyo:
Looks great; thank you!!

PR #3399, opened 2019-10-10
Fix various typos

jackwilsdon:
I found these using the wonderful [misspell](https://github.com/client9/misspell) tool.

### DIFF
--- a/beetsplug/replaygain.py
+++ b/beetsplug/replaygain.py
@@ -570,7 +570,7 @@ class FfmpegBackend(Backend):
         value = line.split(b":", 1)
         if len(value) < 2:
             raise ReplayGainError(
-                u"ffmpeg ouput: expected key value pair, found {0}"
+                u"ffmpeg output: expected key value pair, found {0}"
                 .format(line)
                 )
         value = value[1].lstrip()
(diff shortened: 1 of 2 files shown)


jackwilsdon:
Well I certainly wasn't expecting tests to fail on Travis for this - looks like there are some issues with packages we're installing? :thinking:

sampsyo:
Arg! Yeah, this is obviously not your fault—I don't know why this is happening now, but it looks like something we need to fix in the Travis config. Here's the error:

```
The following packages have unmet dependencies:
gstreamer1.0-plugins-bad : Depends: libgles2-mesa (>= 7.8.1) or
                                    libgles2
Unable to correct problems, you have held broken packages.
```

jackwilsdon:
Yeah I've been playing around with trying to fix it on another branch but can't seem to nail it down - it's weird that it is only failing on specific Python versions! I'll look into it later when I have time if it hasn't resolved itself.

jackwilsdon:
After reverting a4c413e64adc140d824512e183f76d609f4e29cb it seems like things are building again! :tada: must have been a temporary Travis outage.

I'll go ahead and merge this now :+1:.

sampsyo:
Awesome; thank you for figuring that out! The ways of Travis can sometimes be mysterious. And thanks for undoing my failed attempt at a fix…

PR #3400, opened 2019-10-11
Extended export plugin support

austinmm:
The original export plugin only exported as JSON so I added the ability to export in CSV or XML format. To do this I created a new option `-f` or `--format` which specifies the export format, the default is still JSON. For example...
`$ beet export -f csv -i "title,album" ACDC`
I also updated the plugin documentation to reflect the changes I've made to the plugin. Lastly, I also created a unit test for the export plugin; since one hadn't been created yet.

### DIFF
--- a/beetsplug/export.py
+++ b/beetsplug/export.py
@@ -67,9 +67,9 @@ class ExportPlugin(BeetsPlugin):
                 # XML module formatting options.
                 'formatting': {
                     # The output encoding.
-                    'encoding': 'utf-8',
+                    'encoding': 'unicode',
                     # Controls if XML declaration should be added to the file.
-                    'xml_declaration': True,
+                    # 'xml_declaration': True,
                     # Can be either "xml", "html" or "text" (default is "xml").
                     'method': 'xml'
                 }
@@ -199,20 +199,25 @@ class XMLFormat(ExportFormat):
 
     def export(self, data, **kwargs):
         # Creates the XML file structure.
-        library = ET.Element('library')
-        tracks_key = ET.SubElement(library, 'key')
-        tracks_key.text = "Tracks"
-        tracks_dict = ET.SubElement(library, 'dict')
+        library = ET.Element(u'library')
+        tracks_key = ET.SubElement(library, u'key')
+        tracks_key.text = u'tracks'
+        tracks_dict = ET.SubElement(library, u'dict')
         if data and isinstance(data[0], dict):
             for index, item in enumerate(data):
-                track_key = ET.SubElement(tracks_dict, 'key')
+                track_key = ET.SubElement(tracks_dict, u'key')
                 track_key.text = str(index)
-                track_dict = ET.SubElement(tracks_dict, 'dict')
-                track_details = ET.SubElement(track_dict, 'Track ID')
+                track_dict = ET.SubElement(tracks_dict, u'dict')
+                track_details = ET.SubElement(track_dict, u'Track ID')
                 track_details.text = str(index)
                 for key, value in item.items():
                     track_details = ET.SubElement(track_dict, key)
                     track_details.text = value
 
-        tree = ET.ElementTree(library)
-        tree.write(self.out_stream, **kwargs)
+        # tree = ET.ElementTree(element=library)
+        try:
+            data = ET.tostring(library, **kwargs)
+        except LookupError:
+            data = ET.tostring(library, encoding='utf-8', method='xml')
+
+        self.out_stream.write(data)
--- a/beetsplug/export.py
+++ b/beetsplug/export.py
@@ -42,6 +42,9 @@ class ExportPlugin(BeetsPlugin):
 
     def __init__(self):
         super(ExportPlugin, self).__init__()
+        # Used when testing export plugin
+        self.run_results = None
+        self.export_format = None
 
         self.config.add({
             'default_format': 'json',
@@ -68,7 +71,7 @@ class ExportPlugin(BeetsPlugin):
                 'formatting': {
                     'ensure_ascii': False,
                     'indent': 4,
-                    'separators': (''),
+                    'separators': ('>'),
                     'sort_keys': True
                 }
             }
@@ -105,13 +108,12 @@ class ExportPlugin(BeetsPlugin):
         return [cmd]
 
     def run(self, lib, opts, args):
-
         file_path = opts.output
         file_mode = 'a' if opts.append else 'w'
         file_format = opts.format
         format_options = self.config[file_format]['formatting'].get(dict)
 
-        export_format = ExportFormat.factory(
+        self.export_format = ExportFormat.factory(
             file_type=file_format, 
             **{
                 'file_path': file_path,
@@ -135,7 +137,9 @@ class ExportPlugin(BeetsPlugin):
                 continue
             data = key_filter(data)
             items += [data]
-        export_format.export(items, **format_options)
+
+        self.run_results = items
+        self.export_format.export(self.run_results, **format_options)
 
 
 class ExportFormat(object):
@@ -144,6 +148,8 @@ class ExportFormat(object):
         self.path = file_path
         self.mode = file_mode
         self.encoding = encoding
+        # Used for testing
+        self.results = None
 
     @classmethod
     def factory(cls, file_type, **kwargs):
@@ -167,11 +173,13 @@ class JsonFormat(ExportFormat):
         self.export = self.export_to_file if self.path else self.export_to_terminal
 
     def export_to_terminal(self, data, **kwargs):
-        json.dump(data, sys.stdout, cls=ExportEncoder, **kwargs)
+        r = json.dump(data, sys.stdout, cls=ExportEncoder, **kwargs)
+        self.results = str(r)
 
     def export_to_file(self, data, **kwargs):
         with codecs.open(self.path, self.mode, self.encoding) as f:
-            json.dump(data, f, cls=ExportEncoder, **kwargs)
+            r = json.dump(data, f, cls=ExportEncoder, **kwargs)
+            self.results = str(r)
 
 
 class CSVFormat(ExportFormat):
@@ -192,12 +200,15 @@ class CSVFormat(ExportFormat):
         writer = csv.DictWriter(sys.stdout, fieldnames=self.header)
         writer.writeheader()
         writer.writerows(data)
+        self.results = str(writer)
+
 
     def export_to_file(self, data, **kwargs):
         with codecs.open(self.path, self.mode, self.encoding) as f:
             writer = csv.DictWriter(f, fieldnames=self.header)
             writer.writeheader()
             writer.writerows(data)
+            self.results = str(writer)
 
 
 class XMLFormat(ExportFormat):
@@ -233,7 +244,9 @@ class XMLFormat(ExportFormat):
     
     def export_to_terminal(self, data, **kwargs):
         print(data)
+        self.results = str(data)
 
     def export_to_file(self, data, **kwargs):
         with codecs.open(self.path, self.mode, self.encoding) as f:
             f.write(data)
+            self.results = str(data)
--- a/beetsplug/export.py
+++ b/beetsplug/export.py
@@ -67,13 +67,11 @@ class ExportPlugin(BeetsPlugin):
                 # XML module formatting options.
                 'formatting': {
                     # The output encoding.
-                    'encoding': 'unicode',
+                    'encoding': 'utf-8',
                     # Controls if XML declaration should be added to the file.
                     'xml_declaration': True,
                     # Can be either "xml", "html" or "text" (default is "xml").
-                    'method': 'xml',
-                    # Controls formatting of elements that contain no content.
-                    'short_empty_elements': True
+                    'method': 'xml'
                 }
             }
             # TODO: Use something like the edit plugin
--- a/beetsplug/export.py
+++ b/beetsplug/export.py
@@ -42,9 +42,6 @@ class ExportPlugin(BeetsPlugin):
 
     def __init__(self):
         super(ExportPlugin, self).__init__()
-        # Used when testing export plugin
-        self.run_results = None
-        self.export_format = None
 
         self.config.add({
             'default_format': 'json',
@@ -60,19 +57,18 @@ class ExportPlugin(BeetsPlugin):
             'csv': {
                 # csv module formatting options
                 'formatting': {
-                    'ensure_ascii': False,
-                    'indent': 0,
-                    'separators': (','),
-                    'sort_keys': True
+                    'delimiter': ',', # column seperator
+                    'dialect': 'excel', # the name of the dialect to use
+                    'quotechar': '|'
                 }
             },
             'xml': {
                 # xml module formatting options
                 'formatting': {
-                    'ensure_ascii': False,
-                    'indent': 4,
-                    'separators': ('>'),
-                    'sort_keys': True
+                    'encoding': 'unicode', # the output encoding
+                    'xml_declaration':'True', # controls if an XML declaration should be added to the file
+                    'method': 'xml', # either "xml", "html" or "text" (default is "xml")
+                    'short_empty_elements': 'True' # controls the formatting of elements that contain no content.
                 }
             }
             # TODO: Use something like the edit plugin
@@ -103,17 +99,17 @@ class ExportPlugin(BeetsPlugin):
         )
         cmd.parser.add_option(
             u'-f', u'--format', default='json',
-            help=u'specify the format of the exported data. Your options are json (deafult), csv, and xml'
+            help=u"the output format: json (default), csv, or xml"
         )
         return [cmd]
 
     def run(self, lib, opts, args):
         file_path = opts.output
         file_mode = 'a' if opts.append else 'w'
-        file_format = opts.format
+        file_format = opts.format if opts.format else self.config['default_format'].get(str)
         format_options = self.config[file_format]['formatting'].get(dict)
 
-        self.export_format = ExportFormat.factory(
+        export_format = ExportFormat.factory(
             file_type=file_format, 
             **{
                 'file_path': file_path,
@@ -135,11 +131,11 @@ class ExportPlugin(BeetsPlugin):
             except (mediafile.UnreadableFileError, IOError) as ex:
                 self._log.error(u'cannot read file: {0}', ex)
                 continue
+
             data = key_filter(data)
             items += [data]
 
-        self.run_results = items
-        self.export_format.export(self.run_results, **format_options)
+        export_format.export(items, **format_options)
 
 
 class ExportFormat(object):
@@ -148,8 +144,8 @@ class ExportFormat(object):
         self.path = file_path
         self.mode = file_mode
         self.encoding = encoding
-        # Used for testing
-        self.results = None
+        # out_stream is assigned sys.stdout (terminal output) or the file stream for the path specified
+        self.out_stream = codecs.open(self.path, self.mode, self.encoding) if self.path else sys.stdout
 
     @classmethod
     def factory(cls, file_type, **kwargs):
@@ -170,16 +166,9 @@ class JsonFormat(ExportFormat):
     """Saves in a json file"""
     def __init__(self, file_path, file_mode=u'w', encoding=u'utf-8'):
         super(JsonFormat, self).__init__(file_path, file_mode, encoding)
-        self.export = self.export_to_file if self.path else self.export_to_terminal
 
-    def export_to_terminal(self, data, **kwargs):
-        r = json.dump(data, sys.stdout, cls=ExportEncoder, **kwargs)
-        self.results = str(r)
-
-    def export_to_file(self, data, **kwargs):
-        with codecs.open(self.path, self.mode, self.encoding) as f:
-            r = json.dump(data, f, cls=ExportEncoder, **kwargs)
-            self.results = str(r)
+    def export(self, data, **kwargs):
+        json.dump(data, self.out_stream, cls=ExportEncoder, **kwargs)
 
 
 class CSVFormat(ExportFormat):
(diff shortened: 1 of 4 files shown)


sampsyo:
Cool; thanks for getting this started!

Here's one question for context: while I can totally see the utility of a CSV export, I'm struggling to think of a use case for an XML export. Can you elaborate a little more on what you want to use CSV and XML data for? I think that would help calibrate our approach to these features.

beetsplug/export.py
```diff
@@ -40,6 +42,9 @@ class ExportPlugin(BeetsPlugin):
 
     def __init__(self):
         super(ExportPlugin, self).__init__()
+        # Used when testing export plugin
+        self.run_results = None
```
sampsyo: Hmm; I don't think I see where this is actually used?

beetsplug/export.py
```diff
@@ -40,6 +42,9 @@ class ExportPlugin(BeetsPlugin):
...
+        self.export_format = None
```
sampsyo: I also don't quite see why this is a class field instead of just being a local variable in the `run` method. Can we please revert to the old way?

beetsplug/export.py
```diff
@@ -52,6 +57,24 @@ def __init__(self):
...
+                }
+            },
+            'xml': {
+                # xml module formatting options
+                'formatting': {
+                    'ensure_ascii': False,
+                    'indent': 4,
+                    'separators': ('>'),
+                    'sort_keys': True
+                }
```
sampsyo: I don't think any of these config options are actually used in the CSV or XML emitters.

beetsplug/export.py
```diff
@@ -78,17 +101,21 @@ def commands(self):
             u'-o', u'--output',
             help=u'path for the output file. If not given, will print the data'
         )
+        cmd.parser.add_option(
+            u'-f', u'--format', default='json',
+            help=u'specify the format of the exported data. Your options are json (deafult), csv, and xml'
```
sampsyo: This could be shorter: perhaps, "the output format: json (default), csv, or xml".

beetsplug/export.py
```diff
@@ -78,17 +101,21 @@ def commands(self):
...
+        )
         return [cmd]
 
     def run(self, lib, opts, args):
-
         file_path = opts.output
-        file_format = self.config['default_format'].get(str)
         file_mode = 'a' if opts.append else 'w'
+        file_format = opts.format
```
sampsyo: Please preserve the effect of `default_format`.

beetsplug/export.py
```diff
@@ -108,44 +135,118 @@ def run(self, lib, opts, args):
             except (mediafile.UnreadableFileError, IOError) as ex:
                 self._log.error(u'cannot read file: {0}', ex)
                 continue
-
```
sampsyo: Any particular reason to delete this newline?

beetsplug/export.py
```diff
@@ -108,44 +135,118 @@ def run(self, lib, opts, args):
...
             data = key_filter(data)
             items += [data]
 
-        export_format.export(items, **format_options)
+        self.run_results = items
+        self.export_format.export(self.run_results, **format_options)
```
sampsyo: As mentioned above, I think it would be awesome to remove these side effects on mutable class fields…

beetsplug/export.py
```diff
@@ -108,44 +135,118 @@ def run(self, lib, opts, args):
...
-    def factory(cls, type, **kwargs):
-        if type == "json":
-            if kwargs['file_path']:
-                return JsonFileFormat(**kwargs)
-            else:
-                return JsonPrintFormat()
-        raise NotImplementedError()
+    def factory(cls, file_type, **kwargs):
+        if file_type == "json":
+            return JsonFormat(**kwargs)
```
sampsyo: I notice this now behaves differently from it did before—the `kwargs` are now forwarded unconditionally instead of only when the path is unspecified. Does this change the user-visible behavior?

beetsplug/export.py
```diff
@@ -108,44 +135,118 @@ def run(self, lib, opts, args):
...
         raise NotImplementedError()
 
 
-class JsonPrintFormat(ExportFormat):
-    """Outputs to the console"""
+class JsonFormat(ExportFormat):
+    """Saves in a json file"""
+    def __init__(self, file_path, file_mode=u'w', encoding=u'utf-8'):
+        super(JsonFormat, self).__init__(file_path, file_mode, encoding)
+        self.export = self.export_to_file if self.path else self.export_to_terminal
```
sampsyo: This technique where a method is defined conditionally is somewhat surprising. I now see how it works after some puzzling about it, but it could use an explanatory comment.

beetsplug/export.py
```diff
@@ -108,44 +135,118 @@ def run(self, lib, opts, args):
...
+        writer = csv.DictWriter(sys.stdout, fieldnames=self.header)
+        writer.writeheader()
+        writer.writerows(data)
+        self.results = str(writer)
 
-class JsonFileFormat(ExportFormat):
-    """Saves in a json file"""
 
+    def export_to_file(self, data, **kwargs):
+        with codecs.open(self.path, self.mode, self.encoding) as f:
```
sampsyo: I think it would be really useful to factor out the file-opening part of the logic so it doesn't need to be re-implemented for every export format. In general, the output format and output destination (file or stdout) seem orthogonal.

beetsplug/export.py
```diff
@@ -108,44 +135,118 @@ def run(self, lib, opts, args):
...
+
+class CSVFormat(ExportFormat):
+    """Saves in a csv file"""
+    def __init__(self, file_path, file_mode=u'w', encoding=u'utf-8'):
+        super(CSVFormat, self).__init__(file_path, file_mode, encoding)
+        self.header = []
 
     def export(self, data, **kwargs):
-        json.dump(data, sys.stdout, cls=ExportEncoder, **kwargs)
+        if data and type(data) is list and len(data) > 0:
```
sampsyo: I'm a little confused by the `list` type check here—when will the passed object _not_ be a list? And what happens when it's not—so `self.header` remains empty?

Also, this condition could be simplified to:

```py
if data and isinstance(data, list):
```

because `data` is only "true" if it is non-empty.

docs/plugins/export.rst
```diff
@@ -36,11 +39,18 @@ The ``export`` command has these command-line options:
 
 * ``--append``: Appends the data to the file instead of writing.
 
+* ``--format`` or ``-f``: Specifies the format the data will be exported as. If not informed, JSON will be used by default.
+  For example::
+
+      $ beet export -f csv beatles
+      $ beet export -f json beatles
+      $ beet export -f xml beatles
```
sampsyo: I think we can do without the full command-line examples here. Can we just list the options?

docs/plugins/export.rst
```diff
@@ -63,3 +73,15 @@ The default options look like this::
...
+                ensure_ascii: False
+                indent: 0
+                separators: [',']
+                sort_keys: true
+        xml:
+            formatting:
+                ensure_ascii: False
+                indent: 4
+                separators: ['>']
+                sort_keys: true
```
sampsyo: As above, I don't think these options are actually used for the new formats.

beetsplug/export.py
austinmm: I was going to use that class field as a way of checking the output during the testing of the export plugin but I ended up going another route so I will go ahead and remove it.

beetsplug/export.py
austinmm: I reverted it.

beetsplug/export.py
austinmm: Sounds good

beetsplug/export.py
austinmm: would you prefer I get rid of the `default=json` in...
```
cmd.parser.add_option(
            u'-f', u'--format', default='json',
            help=u"the output format: json (default), csv, or xml"
        )
```
and instead have something like...
```
file_format = opts.format if opts.format else self.config['default_format'].get(str)
```

beetsplug/export.py
austinmm: Sounds good, I'll add some comments explaining its purpose

beetsplug/export.py
austinmm: I was just trying to be extra cautious but I completely agree that it is a little overkill, this should work just fine instead; `if data and len(data) > 0:`

docs/plugins/export.rst
austinmm: Sounds good

beetsplug/export.py
austinmm: I was just trying to follow the same format, but you are right they are not being used. I have updated them to now be format specific to json, csv and xml and updated the csv and xml outputs to use them...
```
'json': {
                # json module formatting options
                'formatting': {
                    'ensure_ascii': False,
                    'indent': 4,
                    'separators': (',', ': '),
                    'sort_keys': True
                }
            },
            'csv': {
                # csv module formatting options
                'formatting': {
                    'delimiter': ',', # column seperator
                    'dialect': 'excel', # the name of the dialect to use
                    'quotechar':'|'
                }
            },
            'xml': {
                # xml module formatting options
                'formatting': {
                    'encoding': 'UTF-8', # the output encoding
                    'xml_declaration':'True', # controls if an XML declaration should be added to the file
                    'default_namespace': 'None', # sets the default XML namespace (for “xmlns”)
                    'method': 'xml', # either "xml", "html" or "text" (default is "xml")
                    'short_empty_elements': 'True' # controls the formatting of elements that contain no content.
                }
```

beetsplug/export.py
austinmm: Nope did that by mistake, I added it back

beetsplug/export.py
austinmm: will do

docs/plugins/export.rst
austinmm: I updated them to work correctly now with the added formats CSV and XML

austinmm:
I am still working on all the changes but almost done. Thank you for all the feedback as it was very helpful and constructive. Also to address your question about the XML format. The reason I added the ability to export in the XML format was that this is the preferred export format used by Apple in Apple Music and simply figured that if Apple does it then there must exist a good reason. For CSV exporting, I imagined that some people, including myself, might want to use Excel tools to better analyze their music library.

beetsplug/export.py
austinmm: This should not effect the user in any way. It simply simplifies the logic in my opinion.

beetsplug/export.py
austinmm: I'll explore that idea. Sounds like a better way

austinmm:
I have just pushed the changes that were discussed above. Please let me know if you recommend any other changes. :)

beetsplug/export.py
```diff
@@ -52,6 +54,23 @@ def __init__(self):
                     'sort_keys': True
                 }
             },
+            'csv': {
+                # csv module formatting options
+                'formatting': {
+                    'delimiter': ',', # column seperator
+                    'dialect': 'excel', # the name of the dialect to use
+                    'quotechar': '|'
```
sampsyo: I think you probably want `quotechar` to default to "? In fact, it doesn't seem like the most useful thing to want to override… maybe we can get away without this option.

beetsplug/export.py
```diff
@@ -52,6 +54,23 @@ def __init__(self):
                     'sort_keys': True
                 }
             },
+            'csv': {
+                # csv module formatting options
+                'formatting': {
+                    'delimiter': ',', # column seperator
+                    'dialect': 'excel', # the name of the dialect to use
```
sampsyo: This option seems useful enough! But maybe it's worth linking (in the documentation) to the place in the Python docs where dialects are described?

beetsplug/export.py
```diff
@@ -117,35 +140,74 @@ def run(self, lib, opts, args):
...
-class JsonFileFormat(ExportFormat):
-    """Saves in a json file"""
+class CSVFormat(ExportFormat):
+    """Saves in a csv file"""
+    def __init__(self, file_path, file_mode=u'w', encoding=u'utf-8'):
+        super(CSVFormat, self).__init__(file_path, file_mode, encoding)
+        self.header = []
 
+    def export(self, data, **kwargs):
+        if data and len(data) > 0:
```
sampsyo:
```suggestion
        if data:
```

docs/plugins/export.rst
```diff
@@ -63,3 +69,15 @@ The default options look like this::
...
```
sampsyo: I believe these examples are now out of sync with the implementation. It would also be good to describe the options in English above.

beetsplug/export.py
austinmm: Sounds good, I'll remove it

beetsplug/export.py
austinmm: Oh yes I forgot to do that, thank you for reminding me

beetsplug/export.py
austinmm: Will do

austinmm:
I have made the changes you suggested alongside some other just general code cleanup. I am not sure where I might be disobeying the PEP-8 conventions so I apologize if I am anywhere. I am still new to CIs and so would love any recommendations you might have so I can pass the two CI test.

sampsyo:
For the CI results, take a look at GitHub's "checks" interface on this page. It has "Details" links to the logs from Travis and CircleCI. For example, here's the log from the style checker:
https://travis-ci.org/beetbox/beets/jobs/597112966

You can also consider installing [flake8](http://flake8.pycqa.org/en/latest/) and running it locally.

Also, about the docs: while the text seems good, could you please move these from inline comments in the example YAML to the prose text above? There's already a section there with a bulleted list describing the options, before the page gets to the YAML example. That would be a nice place to add similar descriptions of the new options.

austinmm:
Both the test_export.py and export.py pass the flake8 test so I'm not sure why there is still an issue with the CLI. How do I run the test_export.py on my computer because if I just run `python test/test_export.py` it doesn't compile correctly and neither do any of the other python tests files?

sampsyo:
We have [a wiki page about testing](https://github.com/beetbox/beets/wiki/Testing) that might help! You might want to try using Tox or check out the section about test dependencies.

You can also see the results of the tests directly on Travis:
https://travis-ci.org/beetbox/beets/jobs/597334455

austinmm:
I was able to get all the tests to pass :). Please let me know if you want me to update anything else, but I believe it is production-ready.

beetsplug/export.py
```diff
@@ -78,17 +100,22 @@ def commands(self):
...
+        )
         return [cmd]
 
     def run(self, lib, opts, args):
-
         file_path = opts.output
-        file_format = self.config['default_format'].get(str)
         file_mode = 'a' if opts.append else 'w'
+        file_format = opts.format if opts.format else \
+            self.config['default_format'].get(str)
```
sampsyo: This should suffice: `opts.format or self.config['default_format'].get(str)`

beetsplug/export.py
```diff
@@ -117,35 +145,79 @@ def run(self, lib, opts, args):
...
 class ExportFormat(object):
     """The output format type"""
+    def __init__(self, file_path, file_mode=u'w', encoding=u'utf-8'):
+        self.path = file_path
+        self.mode = file_mode
+        self.encoding = encoding
+        """ self.out_stream =
+                sys.stdout if path doesn't exit
+                codecs.open(..) else
+        """
```
sampsyo: Perhaps this is out of date? Or it can be rephrased as a natural-language comment?

beetsplug/export.py
```diff
@@ -117,35 +145,79 @@ def run(self, lib, opts, args):
...
+                track_key = ET.SubElement(tracks_dict, u'key')
+                track_key.text = str(index)
+                track_dict = ET.SubElement(tracks_dict, u'dict')
+                track_details = ET.SubElement(track_dict, u'Track ID')
+                track_details.text = str(index)
+                for key, value in item.items():
+                    track_details = ET.SubElement(track_dict, key)
+                    track_details.text = value
+
+        # tree = ET.ElementTree(element=library)
```
sampsyo: Out of date?

docs/plugins/export.rst
```diff
@@ -36,19 +39,34 @@ The ``export`` command has these command-line options:
...
 
-- **ensure_ascii**: Escape non-ASCII characters with ``\uXXXX`` entities.
+    - **sort_keys**: Sorts the keys in JSON dictionaries.
 
-- **indent**: The number of spaces for indentation.
+- **CSV Formatting**
+    - **delimiter**: Used as the separating character between fields. The default value is a comma (,).
 
-- **separators**: A ``[item_separator, dict_separator]`` tuple.
+    - **dialect**: A dialect, in the context of reading and writing CSVs, is a construct that allows you to create, store, and re-use various formatting parameters for your data.
```
sampsyo: Instead of duplicating the text from the Python CSV documentation, can we instead link to the explanation there?

docs/plugins/export.rst
```diff
@@ -36,19 +39,34 @@ The ``export`` command has these command-line options:
...
 
-- **sort_keys**: Sorts the keys in JSON dictionaries.
+- **XML Formatting**
+    - **encoding**: Use encoding="unicode" to generate a Unicode string (otherwise, a bytestring is generated).
```
sampsyo: Hmm… how does this affect the usage of the plugin? Because we're writing to a file, it seems like the kind of Python string that gets generated is irrelevant. Perhaps there's an argument to be made that this should _not_ be configurable in that case?

docs/plugins/export.rst
```diff
@@ -36,19 +39,34 @@ The ``export`` command has these command-line options:
...
+
+    - **xml_declaration**: Controls if an XML declaration should be added to the file. Use False for never, True for always, None for only if not US-ASCII or UTF-8 or Unicode (default is None).
```
sampsyo: TBH I'm not sure if I know what an "XML declaration" is. Is this something that people will actually want to control?

docs/plugins/export.rst
```diff
@@ -36,19 +39,34 @@ The ``export`` command has these command-line options:
...
+
+    - **method**: Can be either "xml", "html" or "text" (default is "xml")
```
sampsyo: Maybe it's worth explaining what these options actually do? I'm not sure how setting `method` to `text` will affect the output, for example.

beetsplug/export.py
austinmm: Here is the new comment...
`# creates a file object to write/append or sets to stdout`

beetsplug/export.py
austinmm: sounds good

beetsplug/export.py
austinmm: yeah

docs/plugins/export.rst
austinmm: I will link to an external description but I would imagine it is still good to have some description in the document itself. I just shortened it a little. Let me know if you want me to get rid of the description altogether.

docs/plugins/export.rst
austinmm: I agree, it is not an config option now.

docs/plugins/export.rst
austinmm: Its been removed

austinmm:
I believe I have successfully implemented all your recommended changes. Let me know if I need to make any other changes. Also, here is the updated XML format...
```
<library>
    <tracks>
        <track>
            <title>Catch Me</title>
            <album>BRÅVES</album>
            ....
        </track>
        ....
    </tracks>
</library>
```

docs/plugins/export.rst
sampsyo: Hmm… correct me if I'm wrong, but isn't the HTML example here identical to the XML example? Is there a difference I'm missing?

sampsyo:
Cool! Thanks for continuing to make progress!

I have only two more things in mind:

- We'll need a changelog entry about the new feature.
- I'm still uncertain about the utility of XML export formats. It seems to me that "HTML" export is exactly the same as the "XML" format of XML, isn't it? And the text format doesn't seem all that useful to me…

docs/plugins/export.rst
austinmm: In this case, the HTML is the same but I don't think that is always the case. I agree the text is not very useful. Would you like me to remove this option?

sampsyo: I think so! Unless you can think of a compelling reason why someone would want to use it, let's take it out.

austinmm:
How do we create a changelog entry?

sampsyo:
There's a file in `docs/changelog.rst` that serves that purpose. :smiley:

austinmm:
Let me know if I didn't create the changelog entry correctly or if you notice anything else that needs fixing :)

docs/plugins/export.rst
```diff
@@ -36,24 +39,36 @@ The ``export`` command has these command-line options:
...
+    - **sort_keys**: Sorts the keys in JSON dictionaries.
 
 These options match the options from the `Python json module`_.
 
 .. _Python json module: https://docs.python.org/2/library/json.html#basic-usage
 
+- **CSV Formatting**
+    - **delimiter**: Used as the separating character between fields. The default value is a comma (,).
+
+    - **dialect**: A dialect is a construct that allows you to create, store, and re-use various formatting parameters for your data.
```
sampsyo:
```suggestion
    - **dialect**: The kind of CSV file to produce. The default is `excel`.
```

Keeping this focused on what matters to *users of this plugin*, not to Python programmers.

docs/plugins/export.rst
```diff
@@ -62,4 +77,8 @@ The default options look like this::
                 ensure_ascii: False
                 indent: 4
                 separators: [',' , ': ']
-                sort_keys: true
+                sort_keys: True
```
sampsyo: Lower case is OK here—it’s YAML, not Python.
```suggestion
                sort_keys: true
```

docs/plugins/export.rst
```diff
@@ -62,4 +77,8 @@ The default options look like this::
...
+        csv:
+            formatting:
+                delimiter: ','
+                dialect: 'excel'
```
sampsyo:
```suggestion
                dialect: excel
```

No quotes are usually necessary in YAML.

sampsyo:
Arg, turns out GitHub won’t let me commit my own suggestions on my phone. :( But I will wrap this up when I’m back at a normal computer.

austinmm:
Your changes look great! Thank you for all the help you provided me with this. This was my first time contributing to an open-source project so all your assistance was very much appreciated.

sampsyo:
Wow! In that case, thank you for your diligent efforts, congratulations, and welcome to the world of open source. :smiley: Nicely done!

PR #3405, opened 2019-10-15
Unimported files plugin

MrNuggelz:
Plugin to list all files in the library which are not imported.
Settings:

```yaml
unimported:
   ignore_extensions: [jpg,png] # default []
```


MrNuggelz:
fixed flake8 errors and added documentation

Issue #2835, opened 2018-03-10
hook: React to commands failures


jhermann:
I noticed that `hook.py` uses `subprocess.Popen(command_pieces).wait()` (BTW, why not `subprocess.call`), and does not check the return code (you ignore the `wait` result). It also might be you thought this code does what `check_call` does (namely, throw on non-OK exit). Right now, your exception handler is only triggered when the command cannot be found.

An improvement would be to log a warning on non-OK exit code, and allow failing commands to exit the beets process. Example:

```yaml
hook:
  hooks:
    - event: import
      command: chmod -R go+rX "{lib.directory}"
      on_error: abort
```
Choices for `on_error` would be  `abort`, `ignore` (the default), `log`, and `skip`.

`skip` would depend on the event, e.g. skip an album import if the command fails.

sampsyo:
Sounds good!

I’m not aware of a reason why that code doesn’t use the higher-level wrappers in `subprocess`. That would be nice to fix too.

jackwilsdon:
I wouldn't mind looking into this, however I'm not quite sure how the `skip` functionality would work. Do you have any specific events in mind for the `skip` operations @jhermann? The only tasks I can see which it would really work for are `import_task_created` and maybe a few of the other `import_task_*` events.

Similarly, how would `abort` work? Would it just throw an exception to force beets to exit? It seems like most events don't provide any cancellation functionality, so that looks like it might be the only way to do it.

Would `ignore` stop logging errors? If so then we should probably default to `log` which is what we do right now.

anarcat:
> Do you have any specific events in mind for the `skip` operations @jhermann? The only tasks I can see which it would really work for are `import_task_created` and maybe a few of the other `import_task_*` events.

old issue, but yeah, this would be relevant to avoid messing around with files that are handled by another manager, say a torrent client (#2617).

PR #3409, opened 2019-10-16
hook: switch to using check_call

jackwilsdon:
A small change but hopefully part of a larger piece of work around #2835.

### DIFF
--- a/beetsplug/hook.py
+++ b/beetsplug/hook.py
@@ -105,7 +105,10 @@ class HookPlugin(BeetsPlugin):
                             u' '.join(command_pieces), event)
 
             try:
-                subprocess.Popen(command_pieces).wait()
+                subprocess.check_call(command_pieces)
+            except subprocess.CalledProcessError as exc:
+                self._log.error(u'hook for {0} exited with status {1}',
+                                event, exc.returncode)
             except OSError as exc:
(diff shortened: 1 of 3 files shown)


sampsyo:
Seems right to me! Maybe we should also catch [CalledProcessError](https://docs.python.org/3.7/library/subprocess.html#subprocess.CalledProcessError) and print a friendly message when the hook fails?

jackwilsdon:
Aha, good catch! I clearly didn't look into the `check_call` documentation enough to see it can throw things other than an OSError :grimacing:

jackwilsdon:
I've gotten half way through implementing support for `on_error` - I've implemented `on_error: abort` by raising a `ui.UserError` which does seem to quit beets, although I'm not too sure on what state it actually leaves beets in :confused: I may as well go ahead and finish it off a bit more (thinking of just implementing `abort`, `log` and `ignore` without any of the `skip_*` handlers) and I'll open a PR for it so we can discuss whether or not it is worth implementing there.

sampsyo:
Perfect; thanks! I merged this as is.

That's cool; thanks for looking into `on_error`. And indeed, I'm actually also not entirely sure how dangerous just raising a `UserError` might be—it seems to depend a lot on the context, such as the particular event that is aborting things? Maybe we can get feedback from the original requester about what exactly they wanted to happen and why…

Issue #3298, opened 2019-06-06
title case kept lower when following a special character


doucheymcdoucherson:
Using the `%title` Title Case function in the path, it looks like characters that are preceded by a non-alpha are not changed. 

For example:

The Album Title `Childish Gambino - 2016 - “awaken, My Love!”` where `%title{$album%aunique{}}` is defined in the default path. The "a" in awaken is preceded by a quote character. 

Another example, the track name `08 - Bow Shock (interlude).mp3` where `$track - %title{$title}` is defined. 

Not sure if this is a known issue or not? Any workarounds if known?

Thanks!

Snippet from my config.yaml if it's helpful:

```
paths:
    default: $albumartist - $original_year - %title{$album%aunique{}} %upper{$albumtype_EP}%title{$albumtype_Single}%if{$multidisc,(Disc $disc)}/$track - %title{$title}
    singleton: Non-Album/$artist - $title
    comp: \#VA/$album%aunique{} - $original_year/$track - $artist - $title

plugins: lastgenre fetchart replaygain scrub discogs ftintitle beatport smartplaylist inline

item_fields:
    albumtype_EP: u'(' + albumtype + u')' if albumtype == 'ep' else ''
    albumtype_Single: u'(' + albumtype + u')' if albumtype == 'single' else ''
    multidisc: 1 if disctotal > 1 else 0
```

jackwilsdon:
It looks like we actually changed the original behaviour to do this, as previously it would convert strings such as "Don't" to "Don'T" (see #3033). I'm not sure what the right thing to do here is, as reading this it seems like we should maybe consider switching back to the old functionality but maintain a whitelist or blacklist of characters that shouldn/shouldn't appear before characters that are going to be capitalised.

arcresu:
Maybe we can learn from this project: https://github.com/ppannuto/python-titlecase (at least the test cases could be useful inspiration).

sampsyo:
Interesting find! It's clear that our title-case function could be improved, but I don't think we should switch back to the old version (just `.title()`), since it's also imperfect. I'm marking this as a feature request to make title-casing more robust and standard.

bennettca:
Created a pull request for this issue: https://github.com/beetbox/beets/pull/3411

henry-oberholtzer:
This feature has now been implemented by the plugin [Titlecase](https://github.com/beetbox/beets/commit/b9023521390aa372473d10ac28e3d6c7b93b3335).

PR #3411, opened 2019-10-21
Capitalize letters after punctuations

ColeBennett:
Addresses the issues noted in https://github.com/beetbox/beets/issues/3298

Revamps the current %title functionality, moving from Python's string.capwords() to using the [python-titlecase](https://github.com/ppannuto/python-titlecase) package. Supports the original functionality while also capitalizing the first letter of words that follow a punctuation (such as ', ", /, etc).

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -1482,10 +1482,91 @@ class DefaultTemplateFunctions(object):
         """Covert a string to upper case."""
         return s.upper()
 
+    # Regular expressions used by tmpl_title
+    SMALL = 'a|an|and|as|at|but|by|en|for|if|in|of|on|or|the|to|v\\.?|via|vs\\.?'  # noqa: E501
+    PUNCT = r"""!"“#$%&'‘()*+,\-–‒—―./:;?@[\\\]_`{|}~"""
+    SMALL_WORDS = re.compile(r'^(%s)$' % SMALL, re.I)
+    INLINE_PERIOD = re.compile(r'[a-z][.][a-z]', re.I)
+    UC_ELSEWHERE = re.compile(r'[%s]*?[a-zA-Z]+[A-Z]+?' % PUNCT)
+    UC_INITIALS = re.compile(r'^(?:[A-Z]{1}\.{1}|[A-Z]{1}\.{1}[A-Z]{1})+$')
+    CAPFIRST = re.compile(r"^[%s]*?[A-Za-z]" % PUNCT)
+    SMALL_FIRST = re.compile(r'^([%s]*)(%s)\b' % (PUNCT, SMALL), re.I)
+    SMALL_LAST = re.compile(r'\b(%s)[%s]?$' % (SMALL, PUNCT), re.I)
+    SUBPHRASE = re.compile(r'([:.;?!\-–‒—―][ ])(%s)' % SMALL)
+    APOS_SECOND = re.compile(r"^[dol]{1}['‘]{1}[a-z]+(?:['s]{2})?$", re.I)
+
     @staticmethod
     def tmpl_title(s):
         """Convert a string to title case."""
-        return string.capwords(s)
+        all_caps = s.upper() == s
+        words = re.split('[\t ]', s)
+        processed = []
+        for word in words:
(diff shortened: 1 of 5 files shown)


beets/library.py
```diff
@@ -1482,10 +1481,91 @@ def tmpl_upper(s):
         """Covert a string to upper case."""
         return s.upper()
 
+    # Regular expressions used by tmpl_title
+    SMALL = 'a|an|and|as|at|but|by|en|for|if|in|of|on|or|the|to|v\\.?|via|vs\\.?'  # noqa: E501
```
sampsyo: Instead of `#noqa`ing this line, it could just be broken up onto two lines:

```py
SMALL = 'some stuff here' \
  'more stuff here'
```

bennettca: Hi @sampsy. That's correct, it was taken from that repository and has been modified. I considered adding a dependency, however I found that that it is not as well maintained and has some unnessesary code. Do you think it would be worth adding in this custom version for beets?

sampsyo:
I really do think it is best to keep the code in a library, where it has a shot at being maintained independently. :smiley: Care to elaborate on what you mean by "not as well maintained"? Maybe we can help maintain it!

bennettca:
I've added the library. I agree with you now, it's best to have that managed in an external package. It looks stable in its current state and the unit tests I've added verify the functionality we need. Thanks for taking a look at this! Let me know if there's anything else needed to get this merged in.

nopoz:
Is it possible to get this pull reviewed and merged? Would love to get the title case working properly.

sampsyo:
Thanks for the ping! It looks like I dropped the thread.

There are some unrelated changes here that would be great to roll back so we can get a clean merge… I'll make some comments now.

beets/library.py
```diff
@@ -130,17 +130,17 @@ def format(self, value):
         return time.strftime(beets.config['time_format'].as_str(),
                              time.localtime(value or 0))
 
-    def parse(self, string):
+    def parse(self, s):
```
sampsyo: Let's change these argument names back (because the change is not relevant to the feature at hand).

test/test_library.py
```diff
@@ -585,13 +585,49 @@ def test_upper_case_variable(self):
         self._assert_dest(b'/base/THE TITLE')
 
     def test_title_case_variable(self):
-        self._setf(u'%title{$title}')
-        self._assert_dest(b'/base/The Title')
+        self._setf(u'%title{title}')
+        self._assert_dest(b'/base/Title')
```
sampsyo: I'm not sure why this test changed? We should add a new one if we just want to test non-variable title-casing.

nopoz:
@ColeBennett can you make these requested changes? I'm itching to get this merged into the main repo. :grin:

nopoz:
I tested this and while it fixes the issue of capitalization where there are preceding special characters as described in #3298, it makes undesirable changes to other title case behavior.  It adds an effect of making words like "and" lower case. From my interpretation of "title case", all words should start with a capital, so "and" should be "And" - this is the current behavior of the beets master branch. 

For example, with latest commit https://github.com/austinmm/beets/tree/capitalize_letters_after_punctuations:
```
01 - Me and Your Mama.mp3
```

With lastest beets master 939bc269b8a868e49b5e0301ab3e3ae4949dc41a: 
```
01 - Me And Your Mama.mp3
```

sampsyo:
FWIW, that behavior (not capitalizing “and”) is consistent with most English style guides for typesetting titles, such as the Chicago Manual of Style and the Associated Press.

nopoz:
> FWIW, that behavior (not capitalizing “and”) is consistent with most English style guides for typesetting titles, such as the Chicago Manual of Style and the Associated Press.

Could this be made a config parameter so the enduser can choose between the two?

sampsyo:
Hmm, probably not… I think the best we could do is switching between the new behavior from the proper title case library and the old behavior with Python's [str.title](https://docs.python.org/3/library/stdtypes.html#str.title), which has the flaws described in this issue.

Issue #2866, opened 2018-04-04
Storing metadata from multiple sources


dbogdanov:
Different metadata sources (be it MusicBrainz, Discogs, or something else) have different coverage and users might have to use both for tagging their music. As an example, very often, I cannot find metadata on MusicBrainz and I end up using Discogs instead, and this, as #604 mentions, pollutes the MBID tag space.

In general, it will be great to to make tagging approach more generic and agnostic to metadata sources because:
- Mixing metadata from different sources using the same ID3 tags makes the database inconsistent and this will be a disadvantage in the long term.
- More metadata sources is better (different sources might bring complementary information, why not to store it all? Some users will want this)

This can be achieved by adding an option to allow mining metadata from multiple sources simultaneously having a custom TXXX namespace for each one. To select which source metadata is used as a "master" source for the main ID3 tags, a user should be able to define his own order of preference (similar to a proposal in #2686). The TXXX namespaces for each source could be all customizable by a user too.

sampsyo:
Hi! This is an interesting direction, but I don't quite understand what the concrete suggestion is yet. Is the idea that you would like to, for example, fetch the "artist" field from *both* Discogs and MusicBrainz? The beets library database could store flexible attributes with names like `discogs_artist` and `mb_artist` in addition to the canonical (fixed) attribute, `artist`. Is that what you'd want?

If so, it's important to consider the UI implications. In the importer, you'd need to choose a match for *every* metadata source. If you have both MB and Discogs enabled, for instance, we'd first confirm the match from MB and then offer a second menu of matches from Discogs. Would that be desirable?

And just to check: do you know about beets's flexible attribute system already? If not, it might be worth reading up for background. [This blog post](http://beets.io/blog/flexattr.html) is a good starting point.

dbogdanov:
Yes, as an example, I would like to fetch as much metadata as possible from both Discogs and Musicbrainz: artist, titles, country, label, release number, IDs, genres, etc. The UI can offer matches from all sources that it found one by one, asking user whether to apply each source, as well as which source to use a a canonical. 

At the moment I am a bit outsider to beets, as I purely use it for fixing metadata in files and I don't use the library functionality. Therefore, what I am mostly interested in is writing those tags to the audio files themselves. I guess this should still be a good option to have (ideally, the audio files should be self-descriptive), as well as storing those tags in the library.

sampsyo:
[...]

PR #3413, opened 2019-10-24
Add `discogs_labelid` and `discogs_artistid` fields

msil:
This patch adds some extra discogs fields. I see this as a step forward towards not having the discogs plugin pollute/overwrite the `mb_albumid`/etc fields.


msil:
I have added some new tags to `mediafile` so we can store these values in the tags:
beetbox/mediafile#22

sampsyo:
Hi! This seems totally reasonable—storing the label and artist IDs seems useful. But for what it's worth, we don't actually need to add either of these to MediaFile to get started. They can just be stored in the database.

Any chance you could add a quick note to the changeling?

msil:
I've updated the changelog. Do you think it's worth it for me to start work on having the discogs plugin leave the `mb_*` fields alone and just touch the `discogs_*` fields going forward? I've had issues with some software (funkwhale for example) which are expecting proper MusicBrainz UUIDs in the `mb_*` tags.

msil:
This is part of the way to (hopefully) closing off #604. I'm also quite interested in looking at implementing something for the ideas discussed in #2866 unless anyone has any objections.

sampsyo:
Looks good! We can merge this now, if you think it's ready.

On those two other issues:

- Addressing #604 would be great! The next step, however, would be to change the core autotagger matching logic to allow empty MBID fields. It currently assumes it can use the ID fields to deduplicate redundant matches. Once that's done, it should be possible to remove the MBID "pollution" from the metadata plugins, including the Discogs one.
- Addressing #2866 would also be cool but, as a warning, seems to me like a really big project. It will have to touch the importer UI and the core importer pipeline, both of which are pretty extensive and intricate piles of hacks. Now you know! But don't let that stop you from trying to dive right in. :smiley:

msil:
I'm happy for this to get merged now, thanks.

I'll start looking at the other two issues over the next few days. Indeed #2866 is a bit daunting but I'm up for the challenge :)

Thanks for the advice!

sampsyo:
Awesome; thanks!!

dbogdanov:
This is missing `discogs_releasegroupid`. Add when available for consistency?

Issue #3176, opened 2019-03-05
Support for handling Primary Album Artist


dicolanl:
In beets when you have a CD that is "artist featuring artist" there is no way to name the folder as the primary artist.  This creates multiple folders and then they dont get imported into tools like lidarr.

Example
Artist - Album - Album Artist
As Is: 2 Chainz Featuring Drake - Big Amount - 2 Chainz Featuring Drake
Proposed: 2 Chainz Featuring Drake - Big Amount - 2 Chainz

If you look at the tags the albumartist MBid is for 2 Chainz.
The MBid for the album is as is
the MBid for release-group is as is.

Beets is handling "Album Artist" for Folder naming from the album artist property.  But the Album artist should be the pirmary artist if you want other tools like lidarr to manage this better.

It also creates a bunch of extra artist folders.  
Possible option would be to use the FtinTitle plugin and mirror to FtinAlbumArtist

![tag](https://user-images.githubusercontent.com/25964057/53819299-aece0800-3f37-11e9-9403-bc311513bfb1.png)
![artist](https://user-images.githubusercontent.com/25964057/53819305-b1c8f880-3f37-11e9-97d3-cdc2cb768208.png)
![release](https://user-images.githubusercontent.com/25964057/53819311-b392bc00-3f37-11e9-9733-7afdb5296719.png)

sampsyo:
Interesting! So would you want a new field called something like $primary_artist that only includes the name of the first artist on a track?

dicolanl:
I see two options.
1.  add a primary artist field, which could then be used by your folder naming config.
Pro:  simple.
Con: other tools like media monkey, lidarr, etc wont understand this field (https://help.mp3tag.de/main_tags.html)

2. allow a config that sets the albumartist field based the MBid artist.
pro: other tools would understand this.
con: harder to implement.  I think i saw antother issue indicating Artist MBid should support multiple artists like {artistid}/{artistid}.  The code i assume would pick the 1st.

sampsyo:
I think we would do the first option. You could then combine this with an orthogonal feature like #488 to get the behavior you want where this affects the "real" artist field.

However, I have to admit that I'm not *wild* about the idea of a $primary_artist field, because it implies that we'd also want a $secondary_artist and $tertiary_artist—where will the madness end?? Not to mention $albumartist versions, and then also $artist_credit and $albumartist_credit… finding some sort of a clearer way to access artists as "lists" would be much nicer if we can figure it out.

[...]

PR #3415, opened 2019-10-29
wip: Multiple artists plugin

GuilhermeHideki:
Simple plugin I made (which needs a lot of cleaning, tests and docs) to fetch multiple artists and join then by ';' so that I could split then and get the first artist when needed.

Usage:
- Add the `artists` plugin
- Add inline fields like:
```python
artist_1: artist.split(';')[0]
albumartist_1: albumartist.split(';')[0]
```
- change your paths to use the new fields
-  'beet artists $query' to update the fields

Opened this PR more as an example to how to implement #3176

### DIFF
--- a/beetsplug/artists.py
+++ b/beetsplug/artists.py
@@ -1,0 +1,191 @@
+# -*- coding: utf-8 -*-
+# This file is part of beets.
+# Copyright 2019, Guilherme Danno.
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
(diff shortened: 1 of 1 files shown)


jtpavlock:
@GuilhermeHideki do you intend to finish this off? Or should we close and keep as more of a reference as you mentioned.

GuilhermeHideki:
@jtpavlock closing this because the "right way" of implementing the feature is probably easier:

- edit `_flatten_artist_credit` to accept a `separator` parameter (`default None`)
- create a config `artist_separator` to join the artists (basically `artist_parts.append(separator or el)`
- change the places calling `_flatten_artist_credit` when necessary

I'm just not sure if there are some cons to this

PR #3418, opened 2019-10-31
Clarify getitem docstring. Log when falling back.

RollingStar:
I figure, let someone change the docstring again if the function gets a special case besides artist / album artist.

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -410,7 +410,8 @@ class FormattedItemMapping(dbcore.db.FormattedMapping):
             raise KeyError(key)
 
     def __getitem__(self, key):
-        """Get the value for a key. Certain unset values are remapped.
+        """Get the value for a key. `artist` and `albumartist`
+		are fallback values for each other when unmapped.
         """
         value = self._get(key)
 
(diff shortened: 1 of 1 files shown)


beets/library.py
```diff
@@ -410,16 +410,19 @@ def _get(self, key):
             raise KeyError(key)
 
     def __getitem__(self, key):
-        """Get the value for a key. Certain unset values are remapped.
+        """Get the value for a key. `artist` and `albumartist`
+		are fallback values for each other when unmapped.
```
RollingStar: Better wording: "when not set"?

sampsyo: Yeah, good point—"when not set" is better.

sampsyo:
Thanks! The docstring improvement is definitely better. About the logging, however—it seems like this could be pretty verbose! Just accessing a field like `bar = foo.artist` seems like it should probably not have side effects, but this will log every time you look up the field. Using `beet ls -f '$artist'` might print this out thousands of times, for example.

Maybe there's some more centralized place where we can put this logging that won't be quite so intrusive?

jtpavlock:
@RollingStar do you still have interest in continuing work on this?

sampsyo:
Yo; I rolled this back to just the docstring change so we can take that! Always happy to explore different logging in the future.

RollingStar:
Thanks. I'm still a bit of a mess with git.

PR #3422, opened 2019-11-03
Fix wrong section name (smartplaylist => playlist)

regagain:
See https://discourse.beets.io/t/playlist-vs-smartplaylist-plugins/975/2


sampsyo:
Thanks!!

PR #3425, opened 2019-11-10
Subsonic Playlist Plugin

MrNuggelz:
Plugin to import playlists from subsonic servers.

### DIFF
--- a/beetsplug/subsonicplaylist.py
+++ b/beetsplug/subsonicplaylist.py
@@ -19,7 +19,13 @@ from hashlib import md5
 import xml.etree.ElementTree as ET
 import random
 import string
+from urllib.parse import urlencode
+
 import requests
+from beets.dbcore.query import SubstringQuery
+
+from beets.dbcore import AndQuery, Query, MatchQuery
+
 from beets import plugins
 
 from beets.ui import Subcommand
@@ -46,19 +52,21 @@ class SubsonicPlaylistPlugin(BeetsPlugin):
 
     def update_tags(self, playlist_dict, lib):
         for query, playlist_tag in playlist_dict.items():
-            query = 'artist:"{}" album:"{}" title:"{}"'.format(*query)
+            query = AndQuery([SubstringQuery("artist", query[0]),
+                              SubstringQuery("album", query[1]),
+                              SubstringQuery("title", query[2])])
             items = lib.items(query)
-            if len(items) <= 0:
+            if not items:
                 self._log.warn(u"{} | track not found ({})", playlist_tag,
                                query)
                 continue
             for item in items:
-                item.update({'subsonic_playlist': playlist_tag})
+                item.subsonic_playlist = playlist_tag
                 with lib.transaction():
                     item.try_sync(write=True, move=False)
 
     def get_playlist(self, playlist_id):
-        xml = self.send('getPlaylist', '&id={}'.format(playlist_id)).text
+        xml = self.send('getPlaylist', {'id': playlist_id}).text
         playlist = ET.fromstring(xml)[0]
         if playlist.attrib.get('code', '200') != '200':
             alt_error = 'error getting playlist, but no error message found'
@@ -72,7 +80,7 @@ class SubsonicPlaylistPlugin(BeetsPlugin):
 
     def commands(self):
         def build_playlist(lib, opts, args):
-            self._parse_opts(opts)
+            self.config.set_args(opts)
             ids = self.config['playlist_ids'].as_str_seq()
             if len(self.config['playlist_names'].as_str_seq()) > 0:
                 playlists = ET.fromstring(self.send('getPlaylists').text)[
@@ -102,8 +110,6 @@ class SubsonicPlaylistPlugin(BeetsPlugin):
                         item.try_sync(write=True, move=False)
 
             self.update_tags(playlist_dict, lib)
-            ## notify plugins
-            plugins.send('database_change', lib=lib, model=self)
 
         subsonicplaylist_cmds = Subcommand(
             'subsonicplaylist', help=u'import a subsonic playlist'
@@ -117,24 +123,18 @@ class SubsonicPlaylistPlugin(BeetsPlugin):
         subsonicplaylist_cmds.func = build_playlist
         return [subsonicplaylist_cmds]
 
-    def _parse_opts(self, opts):
-
-        if opts.delete:
-            self.config['delete'].set(True)
-
-        self.opts = opts
-        return True
-
     def generate_token(self):
         salt = ''.join(random.choices(string.ascii_lowercase + string.digits))
         return md5(
             (self.config['password'].get() + salt).encode()).hexdigest(), salt
 
-    def send(self, endpoint, params=''):
+    def send(self, endpoint, params=None):
+        if params is None:
+            params = dict()
         url = '{}/rest/{}?u={}&t={}&s={}&v=1.12.0&c=beets'.format(
             self.config['base_url'].get(),
             endpoint,
             self.config['username'],
             *self.generate_token())
-        resp = requests.get(url + params)
+        resp = requests.get(url + urlencode(params))
         return resp
--- a/beetsplug/subsonicplaylist.py
+++ b/beetsplug/subsonicplaylist.py
@@ -22,7 +22,7 @@ import random
 import string
 import requests
 from beets.util import normpath, bytestring_path, mkdirall, syspath, \
-    path_as_posix, sanitize_path
+    path_as_posix
 
 from beets.ui import Subcommand
 
--- a/beetsplug/subsonicplaylist.py
+++ b/beetsplug/subsonicplaylist.py
@@ -24,9 +24,7 @@ from urllib.parse import urlencode
 import requests
 from beets.dbcore.query import SubstringQuery
 
-from beets.dbcore import AndQuery, Query, MatchQuery
-
-from beets import plugins
+from beets.dbcore import AndQuery
 
 from beets.ui import Subcommand
 
--- a/beetsplug/subsonicplaylist.py
+++ b/beetsplug/subsonicplaylist.py
@@ -20,6 +20,7 @@ import xml.etree.ElementTree as ET
 import random
 import string
 import requests
+from beets import plugins
 
 from beets.ui import Subcommand
 
@@ -101,6 +102,8 @@ class SubsonicPlaylistPlugin(BeetsPlugin):
                         item.try_sync(write=True, move=False)
 
             self.update_tags(playlist_dict, lib)
+            ## notify plugins
+            plugins.send('database_change', lib=lib, model=self)
 
         subsonicplaylist_cmds = Subcommand(
             'subsonicplaylist', help=u'import a subsonic playlist'
--- a/beetsplug/subsonicplaylist.py
+++ b/beetsplug/subsonicplaylist.py
@@ -1,0 +1,126 @@
+# -*- coding: utf-8 -*-
+# This file is part of beets.
+# Copyright 2019, Joris Jensen
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+
+from __future__ import absolute_import, division, print_function
+
+import os
+from hashlib import md5
+import xml.etree.ElementTree as ET
+import random
+import string
+import requests
+from beets.util import normpath, bytestring_path, mkdirall, syspath, \
+    path_as_posix, sanitize_path
+
+from beets.ui import Subcommand
+
+from beets.plugins import BeetsPlugin
+
+__author__ = 'https://github.com/MrNuggelz'
+
+
+class SubsonicPlaylistPlugin(BeetsPlugin):
+
+    def __init__(self):
+        super(SubsonicPlaylistPlugin, self).__init__()
+        self.config.add(
+            {
+                'relative_to': None,
+                'playlist_dir': '.',
+                'forward_slash': False,
+                'playlist_ids': [],
+                'playlist_names': [],
+                'username': '',
+                'password': ''
+            }
+        )
+        self.config['password'].redact = True
+
+    def create_playlist(self, xml, lib):
+        relative_to = self.config['relative_to'].get()
+        if relative_to:
+            relative_to = normpath(relative_to)
+
+        playlist = ET.fromstring(xml)[0]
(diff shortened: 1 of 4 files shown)


MrNuggelz:
Removed unused sanitize_path import

sampsyo:
Cool! This looks like a useful utility. But can I ask you something that might be a seriously naive question?

What does this gain from being a beets plugin? Other plugins in beets have _something_ to do with beets's music library database, its importer, or its music metadata abstraction. Unless I'm missing something, this works by querying Subsonic and writing the results directly to an m3u file—there's no interaction with beets queries, albums, or items. Could it work just as well as a standalone Python script?

MrNuggelz:
The beets library is used to get the file paths for all the tracks in the playlist

sampsyo:
Aha, I see! Thanks for explaining.

I think a slightly more orthogonal way to do this would be to have this plugin just store the relevant information in flexible attributes. Then, other stock plugins (i.e., `smartplaylist`) could help generate m3u files based on those. But this way, you can use the playlist information for other things aside from just generating m3us.

Does that sound useful to you?

MrNuggelz:
I think I understand. But i  don't know how to store a list of playlist names in one flexible attribute and using an arbitrary number of fields would mean you have to look up how the field is named. Do you have a idea how to solve this?

sampsyo:
Good question! A hacky but effective way would be to just store the names with a separator, such as `;`. So the `subsonic_playlist` field or similar might contain `Chill;Pop`, for instance.

MrNuggelz:
What would be the syntax to query such a field, just `subsonic_playlist:Chill`?

sampsyo:
Yep, exactly like that! Or we could consider setting up a special playlist query type that is sensitive to the separator, but that doesn’t seem strictly necessary.

MrNuggelz:
Just to be sure, would happen if there are playlists where the name is a substring of another playlists like:
`Rock;Surf Rock`
Wouldn't the query `subsonic_playlist:Rock` return items from both playlists?

sampsyo:
Yep, that's right—that's the downside of my "separator" proposal. It could be addressed by that custom query type, but that would only be necessary in situations like this.

Alternative, you can issue a regex query—something like `subsonic_playlist::[;^]Rock[;$]`, for example—to match only complete playlist names.

MrNuggelz:
I've implemented the changes, added an '--delete' flag and updated the documentation

docs/plugins/subsonicplaylist.rst
```diff
@@ -0,0 +1,33 @@
...
+flag. This resets all `subsonic_playlist` tag before importing playlists.
+Options to be defined in your config with their default value::
+
+    subsonicplaylist:
+        'base_url': "https://your.subsonic.server"
+        'delete': False,
+        'playlist_ids': [],
+        'playlist_names': [],
+        'username': '',
+        'password': ''
```
sampsyo, line 31:
```suggestion
    subsonicplaylist:
        base_url: "https://your.subsonic.server"
        delete: no
        playlist_ids: []
        playlist_names: []
        username: ''
        password: ''
```

Proper YAML syntax.

docs/plugins/subsonicplaylist.rst
```diff
@@ -0,0 +1,33 @@
...
+
+    subsonic_playlist: ";first playlist;second playlist"
+
+To get all items in a playlist use the query `;playlist name;`.
+
+Command Line Usage
+------------------
+
+To use the ``subsonicplaylist`` plugin, enable it in your configuration (see
+:ref:`using-plugins`). Then use it by invoking the ``subsonicplaylist`` command.
```
sampsyo, line 17: I think it would be great to include a minimal config, with required values, here. Something like this… "Next, configure the plugin to connect to your Subsonic server, like this:"

```yaml
subsonicplaylist:
    base_url: http://subsonic.example.com
    username: me
    password: seekrit
```

"Then you can use it by invoking the `subsonicplaylist` command."

docs/plugins/subsonicplaylist.rst
```diff
@@ -0,0 +1,33 @@
...
+By default only the tags of the items found for playlists will be updated.
+This means that, if one imported a playlist, then delete one song from it and
+imported the playlist again, the deleted song will still have the playlist set
+in its `subsonic_playlist` tag. To solve this problem one can use the `-d/--delete`
+flag. This resets all `subsonic_playlist` tag before importing playlists.
+Options to be defined in your config with their default value::
```
sampsyo, line 23: This should probably be a new paragraph, saying something like "here's an example configuration with all the available options."

docs/plugins/subsonicplaylist.rst
```diff
@@ -0,0 +1,33 @@
+Subsonic Playlist Plugin
+========================
+
+The ``subsonicplaylist`` plugin allows to import playlists from a subsonic server.
+This is done by retrieving the track infos from the subsonic server, searching
```
sampsyo, line 5:
```suggestion
This is done by retrieving the track info from the subsonic server, searching
```

docs/plugins/subsonicplaylist.rst
```diff
@@ -0,0 +1,33 @@
...
+them in the beets library and adding the playlist names to the
```
sampsyo, line 6:
```suggestion
for them in the beets library, and adding the playlist names to the
```

docs/plugins/subsonicplaylist.rst
```diff
@@ -0,0 +1,33 @@
...
+`subsonic_playlist` tag of the found items. The content of the tag has the format:
+
+    subsonic_playlist: ";first playlist;second playlist"
+
+To get all items in a playlist use the query `;playlist name;`.
```
sampsyo, line 11: Wouldn't this require that there be a trailing ; in the tag as well?

beetsplug/subsonicplaylist.py
```diff
@@ -0,0 +1,140 @@
...
+        self.opts = opts
+        return True
+
+    def generate_token(self):
+        salt = ''.join(random.choices(string.ascii_lowercase + string.digits))
+        return md5(
+            (self.config['password'].get() + salt).encode()).hexdigest(), salt
+
+    def send(self, endpoint, params=''):
+        url = '{}/rest/{}?u={}&t={}&s={}&v=1.12.0&c=beets'.format(
```
sampsyo, line 134: It might be best to use `urlencode` to avoid unpleasantness when there are special characters in the parameters:
https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode

beetsplug/subsonicplaylist.py
```diff
@@ -0,0 +1,140 @@
...
+            action='store_true',
+            help=u'delete tag from items not in any playlist anymore',
+        )
+        subsonicplaylist_cmds.func = build_playlist
+        return [subsonicplaylist_cmds]
+
+    def _parse_opts(self, opts):
+
+        if opts.delete:
+            self.config['delete'].set(True)
```
sampsyo, line 123: Confuse actually has a built-in way to do this with no explicit code! See the `set_args` method used elsewhere in beets:
https://github.com/beetbox/beets/blob/0c033bdaf10b78be04e6ebc8f998d9d3d5f70e3d/beetsplug/ftintitle.py#L101

beetsplug/subsonicplaylist.py
```diff
@@ -0,0 +1,140 @@
...
+                'playlist_names': [],
+                'username': '',
+                'password': ''
+            }
+        )
+        self.config['password'].redact = True
+
+    def update_tags(self, playlist_dict, lib):
+        for query, playlist_tag in playlist_dict.items():
+            query = 'artist:"{}" album:"{}" title:"{}"'.format(*query)
```
sampsyo, line 49: This is a little dangerous because the parameters might have quotes in them. A safer alternative would be to build a query programmatically—see other plugins that build things out of AndQuery and SubstringQuery, for example.

beetsplug/subsonicplaylist.py
```diff
@@ -0,0 +1,140 @@
...
+            items = lib.items(query)
+            if len(items) <= 0:
```
sampsyo, line 51:
```suggestion
            if not items:
```

beetsplug/subsonicplaylist.py
```diff
@@ -0,0 +1,140 @@
...
+                self._log.warn(u"{} | track not found ({})", playlist_tag,
+                               query)
+                continue
+            for item in items:
+                item.update({'subsonic_playlist': playlist_tag})
```
sampsyo, line 56: This can be written more simply as:
```suggestion
                item.sbsonic_playlist = playlist_tag
```

beetsplug/subsonicplaylist.py
```diff
@@ -0,0 +1,140 @@
...
+            # delete old tags
+            if self.config['delete']:
+                for item in lib.items('subsonic_playlist:";"'):
+                    item.update({'subsonic_playlist': ''})
+                    with lib.transaction():
+                        item.try_sync(write=True, move=False)
+
+            self.update_tags(playlist_dict, lib)
+            ## notify plugins
+            plugins.send('database_change', lib=lib, model=self)
```
sampsyo, line 106: Is this necessary? I think it already gets fired whenever we do `try_sync`.

docs/plugins/subsonicplaylist.rst
MrNuggelz: Your right. The example is wrong as every playlist named has a trailing semicolon

MrNuggelz:
I updated the code, but I also have one question. Is there a way to update multiple entries with one transaction? I noticed that for playlists with many songs it takes quite some time to execute all the updates.

sampsyo:
Nice! And yes! It just requires building up a list of the items that have changed and applying them in a loop, like this:
https://github.com/beetbox/beets/blob/f43a0c3270d18af19affea04fae9fb39ab02d784/beets/ui/commands.py#L1413-L1415

docs/plugins/subsonicplaylist.rst
```diff
@@ -0,0 +1,43 @@
...
+
+    subsonicplaylist:
+        base_url: "https://your.subsonic.server"
+        delete: no
+        playlist_ids: []
+        playlist_names: []
+        username: ''
+        password: ''
+
+Parameters `base_url`, `username` and `password` must be defined!
```
sampsyo, line 43:
```suggestion
The `base_url`, `username`, and `password` options are required.
```

beetsplug/subsonicplaylist.py
sampsyo: I was sorta recommending something like this:

```
params = {
    'u': self.config['base_url'].get(),
    't': endpoint,
    ...
}
```

As it stands, doing `+ urlencode(params)` might erroneously concatenate two query strings. (But `params` seems to be unused?)

beetsplug/subsonicplaylist.py
```diff
@@ -0,0 +1,138 @@
...
+    def commands(self):
+        def build_playlist(lib, opts, args):
+            self.config.set_args(opts)
+            ids = self.config['playlist_ids'].as_str_seq()
+            if len(self.config['playlist_names'].as_str_seq()) > 0:
+                playlists = ET.fromstring(self.send('getPlaylists').text)[
+                    0]
+                if playlists.attrib.get('code', '200') != '200':
+                    alt_error = 'error getting playlists,' \
+                                ' but no erro message found'
```
sampsyo, line 88: Misspelled?

beetsplug/subsonicplaylist.py
```diff
@@ -0,0 +1,138 @@
...
+            for playlist_id in ids:
+                name, tracks = self.get_playlist(playlist_id)
+                for track in tracks:
+                    if track not in playlist_dict:
+                        playlist_dict[track] = ';'
+                    playlist_dict[track] += name + ';'
+            # delete old tags
+            if self.config['delete']:
+                for item in lib.items('subsonic_playlist:";"'):
+                    item.update({'subsonic_playlist': ''})
```
sampsyo, line 106: As mentioned elsewhere, this can be an attribute assignment.

beetsplug/subsonicplaylist.py
```diff
@@ -0,0 +1,138 @@
...
+                    with lib.transaction():
+                        item.try_sync(write=True, move=False)
```
sampsyo, line 108: Since the new metadata is being saved anyway, is this necessary?

beetsplug/subsonicplaylist.py
```diff
@@ -0,0 +1,138 @@
...
+        name = playlist.attrib.get('name', 'undefined')
+        tracks = [(t.attrib['artist'], t.attrib['album'], t.attrib['title'])
+                  for t in playlist]
+        return name, tracks
+
+    def commands(self):
+        def build_playlist(lib, opts, args):
+            self.config.set_args(opts)
+            ids = self.config['playlist_ids'].as_str_seq()
+            if len(self.config['playlist_names'].as_str_seq()) > 0:
```
sampsyo, line 83: Recall that `if len(seq) > 0:` is the same as `if seq:`.

beetsplug/subsonicplaylist.py
```diff
@@ -0,0 +1,138 @@
...
+                    for playlist in playlists:
+                        if name == playlist.attrib['name']:
+                            ids.append(playlist.attrib['id'])
+            playlist_dict = dict()
+            for playlist_id in ids:
+                name, tracks = self.get_playlist(playlist_id)
+                for track in tracks:
+                    if track not in playlist_dict:
+                        playlist_dict[track] = ';'
+                    playlist_dict[track] += name + ';'
```
sampsyo, line 102: Perhaps the code would be a little easier to manage if this stuff—which builds up `playlist_dict`—were refactored into its own function?

beetsplug/subsonicplaylist.py
```diff
@@ -0,0 +1,138 @@
...
+from urllib.parse import urlencode
+
+import requests
+from beets.dbcore.query import SubstringQuery
+
+from beets.dbcore import AndQuery
+
+from beets.ui import Subcommand
+
+from beets.plugins import BeetsPlugin
```
sampsyo, line 31: Any particular reason for the double-spacing around these imports?

beetsplug/subsonicplaylist.py
MrNuggelz: this is for songs removed from a playlist as the aren't contained in the playlist_dict anymore. But I should be able to optimize it, so that it doesn't updates songs contained in the playlist_dict

beetsplug/subsonicplaylist.py
MrNuggelz: Guess it was from Pycharm autoformat

sampsyo:
This looks about right! Shall we call it done?

MrNuggelz:
Yes

sampsyo:
Awesome!! Very nice work!

MrNuggelz:
Sadly I just noticed that the substring query selects all songs containing the title. So if you have some song and also some remixes of that song all of them will be added to the playlist. Guess I have to open a new merge request for that

sampsyo:
Aha! Yes, you will want `MatchQuery`. (`StringFieldQuery` is an abstract base class and can't be used by itself.)

PR #3437, opened 2019-11-24
Increment playlist_version when a track is consumed.

aereaux:
This fixes the playlist not updating when in consume mode, at least in
ncmpcpp.

Someone who knows the mpd protocol better might want to make sure I'm doing this right, though,


sampsyo:
Sounds good! Could you please add a quick changelog entry?

aereaux:
Thanks for the quick response.  I added an entry.

sampsyo:
Excellent; thanks again. :sparkles:

PR #3440, opened 2019-11-24
Chris nguyen deliverible2task2

cmnguyen01:
fixed long method in ___init___.py


sampsyo:
Hi there—what's up with these pull requests? You keep opening them with no context and then immediately closing them. Is there something you're trying to do?

If you don't intend for these to be merged, please don't do this. It needlessly sends spam to lots of people.

PR #3448, opened 2019-12-06
Genius lyrics fix for API capitalization behaviors

xhocquet:


beetsplug/lyrics.py
```diff
@@ -395,13 +395,15 @@ def fetch(self, artist, title):
 
         song_info = None
         for hit in json["response"]["hits"]:
-            if hit["result"]["primary_artist"]["name"] == artist:
+            if hit["result"]["primary_artist"]["name"].strip(u'\u200b').lower() == artist.lower():
```
sampsyo: Can you please put the left-hand side into a variable, like this, so the line doesn't get to long?

```py
hit_artist = hit["result"]["primary_artist"]["name"].strip(u'\u200b').lower()
```

Also, it might be really useful to have a comment here that describes why the odd-looking `.strip(u'\u200b')` is necessary.

beetsplug/lyrics.py
```diff
@@ -395,13 +395,15 @@ def fetch(self, artist, title):
...
                 song_info = hit
                 break
 
         if song_info:
             song_api_path = song_info["result"]["api_path"]
             return self.lyrics_from_song_api_path(song_api_path)
+        else:
+            self._log.debug(u'Genius did not return a matching artist entry')
```
sampsyo: Suggest a shorter wording:
```suggestion
            self._log.debug(u'genius: No matching artist')
```

beetsplug/lyrics.py
xhocquet: Sounds good, I wasn't clear on when comments are used in the codebase and whether long lines were an issue for now, but I agree! Will address your comments 😄

xhocquet:
Hey @sampsyo , sorry for the delay over the past couple months but I wanted to wrap this up and get it merged! Tested locally again to make sure everything was smooth and added your suggestions.

sampsyo:
Yay; it looks perfect!! Thank you for sticking with this and figuring out a very tricky issue. :smiley:

PR #3449, opened 2019-12-06
Added schema variable and added more logging

jef:
### Description

The plugin does a great job updating, but if I wanted to use the 443 port (https) it fails. Example, going to https://google:80 fails because it doesn't know how to forward this. So in the users' case, they would have to assume that giving port 80 in the config will forward to 443 using https properly. This just makes it more obvious.

I also added some more logs to make it more obvious to the user what's going on if they use the `--verbose` option. And if they don't, then it will send a `log.info` for success cases.

In terms of logging, I think I did this right. Please let me know if I didn't. Also first time contributing something in Python -- sorry if it doesn't look right. Let me know if I need to update anything!

### DIFF
--- a/beetsplug/subsonicupdate.py
+++ b/beetsplug/subsonicupdate.py
@@ -37,51 +37,42 @@ from beets.plugins import BeetsPlugin
 __author__ = 'https://github.com/maffo999'
 
 
-def build_payload():
-    """ To avoid sending plaintext passwords, authentication will be
-    performed via username, a token, and a 6 random
-    letters/numbers sequence.
-    The token is the concatenation of your password and the 6 random
-    letters/numbers (the salt) which is hashed with MD5.
-    """
+def create_token():
+    """ Creates salt and token from given password.
 
-    user = config['subsonic']['user'].as_str()
+    :return: The generated salt and hashed token
+    """
     password = config['subsonic']['pass'].as_str()
 
     # Pick the random sequence and salt the password
     r = string.ascii_letters + string.digits
-    salt = "".join([random.choice(r) for n in range(6)])
-    t = password + salt
-    token = hashlib.md5()
-    token.update(t.encode('utf-8'))
+    salt = "".join([random.choice(r) for _ in range(6)])
+    salted_password = password + salt
+    token = hashlib.md5().update(salted_password.encode('utf-8')).hexdigest()
 
     # Put together the payload of the request to the server and the URL
-    return {
-        'u': user,
-        't': token.hexdigest(),
-        's': salt,
-        'v': '1.15.0',  # Subsonic 6.1 and newer.
-        'c': 'beets'
-    }
+    return salt, token
 
 
-def formal_url():
-    """ Formats URL to send request to Subsonic
-    DEPRECATED schema, host, port, contextpath; use ${url}
+def format_url():
+    """ Get the Subsonic URL to trigger a scan. Uses either the url
+    config option or the deprecated host, port, and context_path config
+    options together.
+
+    :return: Endpoint for updating Subsonic
     """
-
-    host = config['subsonic']['host'].as_str()
-    port = config['subsonic']['port'].get(int)
-
-    context_path = config['subsonic']['contextpath'].as_str()
-    if context_path == '/':
-        context_path = ''
 
     url = config['subsonic']['url'].as_str()
     if url and url.endsWith('/'):
         url = url[:-1]
 
+    # @deprecated("Use url config option instead")
     if not url:
+        host = config['subsonic']['host'].as_str()
+        port = config['subsonic']['port'].get(int)
+        context_path = config['subsonic']['contextpath'].as_str()
+        if context_path == '/':
+            context_path = ''
         url = "http://{}:{}{}".format(host, port, context_path)
 
     return url + '/rest/startScan'
@@ -105,8 +96,17 @@ class SubsonicUpdate(BeetsPlugin):
         self.register_listener('import', self.start_scan)
 
     def start_scan(self):
-        url = formal_url()
-        payload = build_payload()
+        user = config['subsonic']['user'].as_str()
+        url = format_url()
+        salt, token = create_token()
+
+        payload = {
+            'u': user,
+            't': token,
+            's': salt,
+            'v': '1.15.0',  # Subsonic 6.1 and newer.
+            'c': 'beets'
+        }
 
         response = requests.post(url, params=payload)
 
@@ -117,4 +117,4 @@ class SubsonicUpdate(BeetsPlugin):
         else:
             self._log.error(
                 u'Generic error, please try again later [Status Code: {}]'
-                    .format(response.status_code))
+                .format(response.status_code))
--- a/beetsplug/subsonicupdate.py
+++ b/beetsplug/subsonicupdate.py
@@ -25,14 +25,66 @@ a "subsonic" section like the following:
 """
 from __future__ import division, absolute_import, print_function
 
-from beets.plugins import BeetsPlugin
-from beets import config
-import requests
-import string
 import hashlib
 import random
+import string
+
+import requests
+
+from beets import config
+from beets.plugins import BeetsPlugin
 
 __author__ = 'https://github.com/maffo999'
+
+
+def build_payload():
+    """ To avoid sending plaintext passwords, authentication will be
+    performed via username, a token, and a 6 random
+    letters/numbers sequence.
+    The token is the concatenation of your password and the 6 random
+    letters/numbers (the salt) which is hashed with MD5.
+    """
+
+    user = config['subsonic']['user'].as_str()
+    password = config['subsonic']['pass'].as_str()
+
+    # Pick the random sequence and salt the password
+    r = string.ascii_letters + string.digits
+    salt = "".join([random.choice(r) for n in range(6)])
+    t = password + salt
+    token = hashlib.md5()
+    token.update(t.encode('utf-8'))
+
+    # Put together the payload of the request to the server and the URL
+    return {
+        'u': user,
+        't': token.hexdigest(),
+        's': salt,
+        'v': '1.15.0',  # Subsonic 6.1 and newer.
+        'c': 'beets'
+    }
+
+
+def formal_url():
+    """ Formats URL to send request to Subsonic
+    DEPRECATED schema, host, port, contextpath; use ${url}
+    """
+
+    host = config['subsonic']['host'].as_str()
+    port = config['subsonic']['port'].get(int)
+
+    context_path = config['subsonic']['contextpath'].as_str()
+    if context_path == '/':
+        context_path = ''
+
+    url = config['subsonic']['url'].as_str()
+    if url and url.endsWith('/'):
+        url = url[:-1]
+
+    if not url:
+        url = "http://{}:{}{}".format(host, port, context_path)
+
+    return url + '/rest/startScan'
 
 
 class SubsonicUpdate(BeetsPlugin):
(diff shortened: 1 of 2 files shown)


sampsyo:
Thanks for pointing this out; this could obviously be better!

I have a slightly different proposal to run by you: instead of the separate and varied config options `host`, `port`, and `contextpath`, and then layering on URL schemes on top, we should consider deprecating all of them and replacing them with a single `url` config parameter. That is, you'd just use `https://example.com:123` if that's what you need. In addition to being simpler, this would have the added benefit of letting you use SSL even on non-default ports.

jef:
> Thanks for pointing this out; this could obviously be better!
> 
> I have a slightly different proposal to run by you: instead of the separate and varied config options `host`, `port`, and `contextpath`, and then layering on URL schemes on top, we should consider deprecating all of them and replacing them with a single `url` config parameter. That is, you'd just use `https://example.com:123` if that's what you need. In addition to being simpler, this would have the added benefit of letting you use SSL even on non-default ports.

Good idea. In that case, I'll make some changes and have you take a look again!

jef:
Let me know what you think -- maybe needs other changes. Definitely need a documentation update.

beetsplug/subsonicupdate.py
```diff
@@ -78,10 +86,26 @@ def loaded(self):
             'v': '1.15.0',  # Subsonic 6.1 and newer.
             'c': 'beets'
         }
+
+        if port == 443:
+            schema = 'https://'
```
sampsyo: I actually think we can omit the automatic scheme switching based on the port altogether if we have the URL version… the idea being that, if you want to use https, you _must_ use the new-style configuration.

beetsplug/subsonicupdate.py
```diff
@@ -51,12 +53,18 @@ def __init__(self):
...
+        schema = 'http://'
+        # DEPRECATED host, port, user, passw, contextpath
+        # use ${url}
         host = config['subsonic']['host'].as_str()
         port = config['subsonic']['port'].get(int)
         user = config['subsonic']['user'].as_str()
         passw = config['subsonic']['pass'].as_str()
         contextpath = config['subsonic']['contextpath'].as_str()
 
+        url = config['subsonic']['url'].as_str()
```
sampsyo: I think this new setting will need a default in the `config['subsonic'].add(...)` call.

beetsplug/subsonicupdate.py
```diff
@@ -51,12 +53,18 @@ def __init__(self):
...
+        if url.endsWith('/'): url = url[:-1]
```
sampsyo: Please break this into two lines (PEP-8 style).

beetsplug/subsonicupdate.py
```diff
@@ -78,10 +86,26 @@ def loaded(self):
...
+        if port == 443:
+            schema = 'https://'
+
         if contextpath == '/':
             contextpath = ''
-        url = "http://{}:{}{}/rest/startScan".format(host, port, contextpath)
+
+        if not url:
+            url = "{}{}:{}{}/rest/startScan" \
+                .format(schema, host, port, contextpath)
```
sampsyo: Instead of doing this here, let's instead put something like this up near the initial configuration setup stuff:

``py
if not url:
    host = config['subsonic']['host'].as_str()
    # et cetera
    url = "http://{}:{}{}".format(host, port, contextpath)
```

Then, down here, we can just use `url` knowing that it is populated with the right thing.

File: beetsplug/subsonicupdate.py [hunk 5]
```diff
@@ -78,10 +86,26 @@ def loaded(self):
...
-        if response.status_code != 200:
-            self._log.error(u'Generic error, please try again later.')
+        if response.status_code == 403:
+            self._log.error(u'Forbidden, invalid permissions')
+        elif response.status_code == 404:
+            self._log.error(u'Not Found, resource cannot be found')
+        elif response.status_code == 200:
+            self._log.info(u'Updating Subsonic')
+        else:
+            self._log.error(u'Generic error, please try again later')
```
sampsyo: Maybe this should include the status code?

beetsplug/subsonicupdate.py
```diff
@@ -78,10 +86,26 @@ def loaded(self):
...
+            url += '/rest/startScan'
+
         response = requests.post(url, params=payload)
 
-        if response.status_code != 200:
-            self._log.error(u'Generic error, please try again later.')
+        if response.status_code == 403:
+            self._log.error(u'Forbidden, invalid permissions')
+        elif response.status_code == 404:
+            self._log.error(u'Not Found, resource cannot be found')
```
sampsyo: The message seems a little redundant. Maybe this would be better off with the generic error message below, just with the status code included?

beetsplug/subsonicupdate.py
```diff
@@ -78,10 +86,26 @@ def loaded(self):
...
+                .format(schema, host, port, contextpath)
+        else:
+            url += '/rest/startScan'
+
         response = requests.post(url, params=payload)
 
-        if response.status_code != 200:
-            self._log.error(u'Generic error, please try again later.')
+        if response.status_code == 403:
+            self._log.error(u'Forbidden, invalid permissions')
```
sampsyo: Does this happen when the username or password is wrong? Maybe it would be more useful to say something like "Server authentication failed."

beetsplug/subsonicupdate.py
```diff
@@ -78,10 +86,26 @@ def loaded(self):
...
+        elif response.status_code == 404:
+            self._log.error(u'Not Found, resource cannot be found')
+        elif response.status_code == 200:
+            self._log.info(u'Updating Subsonic')
```
sampsyo: Does this print whenever the update succeeds? That seems like it might be a little noisy… would it be OK to leave this off for now?

jef:
Oh wow... Sorry, I went quite a bit ahead before you commented.

beetsplug/subsonicupdate.py
jef: Yep, that's the one.

beetsplug/subsonicupdate.py
jef: Sure! I can put in debug.

jef:
Okay! I'm done playing with it now if you want to take another review. Thanks!

beetsplug/subsonicupdate.py
```diff
@@ -25,16 +25,68 @@
...
 
 __author__ = 'https://github.com/maffo999'
 
 
+def build_payload():
+    """ To avoid sending plaintext passwords, authentication will be
+    performed via username, a token, and a 6 random
+    letters/numbers sequence.
+    The token is the concatenation of your password and the 6 random
+    letters/numbers (the salt) which is hashed with MD5.
```
sampsyo: While this is useful context as well, maybe this docstring should first say a little more directly what the function _does_—that is, what the returned dict is supposed to be useful for, and mentioning the fact that the stuff it uses comes from the configuration.

beetsplug/subsonicupdate.py
```diff
@@ -25,16 +25,68 @@
...
+    return {
+        'u': user,
+        't': token.hexdigest(),
+        's': salt,
+        'v': '1.15.0',  # Subsonic 6.1 and newer.
+        'c': 'beets'
+    }
+
+
+def formal_url():
```
sampsyo: Is this supposed to be `format_url`? Not sure what's "formal" about it. :smiley:

beetsplug/subsonicupdate.py
```diff
@@ -25,16 +25,68 @@
...
+    """ Formats URL to send request to Subsonic
+    DEPRECATED schema, host, port, contextpath; use ${url}
+    """
```
sampsyo: Suggested docstring:

> Get the Subsonic URL to access to trigger a scan. Uses either the `url` config option or the deprecated `host`, `port`, and `context_path` config options together.

beetsplug/subsonicupdate.py
```diff
@@ -25,16 +25,68 @@
...
+
+    context_path = config['subsonic']['contextpath'].as_str()
+    if context_path == '/':
+        context_path = ''
+
+    url = config['subsonic']['url'].as_str()
+    if url and url.endsWith('/'):
+        url = url[:-1]
+
+    if not url:
```
sampsyo: I recommend moving all the config accesses to `host`, `port`, etc. below this (or an early `return`) so we don't bother looking up configuration information that we don't need.

docs/plugins/subsonicupdate.rst
```diff
@@ -12,6 +12,12 @@ Then, you'll probably want to configure the specifics of your Subsonic server.
 You can do that using a ``subsonic:`` section in your ``config.yaml``,
 which looks like this::
 
+    subsonic:
+        url: https://mydomain.com:443/subsonic
+        user: username
+        pass: password
+
+    # DEPRECATED
```
sampsyo: Let's just delete the deprecated configuration stuff altogether. Nobody needs to know it exists—they can just keep using it if it's already in their config file. :smiley:

docs/plugins/subsonicupdate.rst
```diff
@@ -12,6 +12,12 @@ Then, you'll probably want to configure the specifics of your Subsonic server.
 You can do that using a ``subsonic:`` section in your ``config.yaml``,
 which looks like this::
 
+    subsonic:
+        url: https://mydomain.com:443/subsonic
```
sampsyo: Let's use the officially sanctioned domain for examples:
```suggestion
        url: https://example.com:443/subsonic
```

docs/plugins/subsonicupdate.rst
```diff
@@ -30,6 +36,14 @@ Configuration
 
 The available options under the ``subsonic:`` section are:
 
+- **url**: The Subsonic server resource. Default: ``http://localhost:4040``
+
+Example: ``https://mydomain.com:443/subsonic``
+
+\* Note: context path is optional
+
+DEPRECATED:
```
sampsyo: Also OK to delete this stuff below.

docs/plugins/subsonicupdate.rst
```diff
@@ -30,6 +36,14 @@ Configuration
 
 The available options under the ``subsonic:`` section are:
 
+- **url**: The Subsonic server resource. Default: ``http://localhost:4040``
+
+Example: ``https://mydomain.com:443/subsonic``
+
+\* Note: context path is optional
```
sampsyo: I think this is OK to delete? We do already give an example that doesn't have an extra path.

beetsplug/subsonicupdate.py
jef: Lol...

docs/plugins/subsonicupdate.rst
jef: Didn't even know that was a thing. Thanks!

docs/plugins/subsonicupdate.rst
jef: I agree. Just want to be thorough here. But I imagine beets isn't for the weak hearted anyhow 😛

docs/plugins/subsonicupdate.rst
```diff
@@ -13,11 +13,11 @@ You can do that using a ``subsonic:`` section in your ``config.yaml``,
...
 
     subsonic:
-        host: X.X.X.X
-        port: 4040
+        url: https://example.com:443/subsonic
         user: username
         pass: password
-        contextpath: /subsonic
+
+\* NOTE: The pass config option can either be clear text or hex-encoded with a "enc:" prefix.
```
jef: http://www.subsonic.org/pages/api.jsp from there API docs

jef:
Not sure if I understand the CI problems... Is it trying to download packages that it can't? I don't think my PR is related.

docs/plugins/subsonicupdate.rst
```diff
@@ -30,8 +30,4 @@ Configuration
 
 The available options under the ``subsonic:`` section are:
 
-- **host**: The Subsonic server name/IP. Default: ``localhost``
-- **port**: The Subsonic server port. Default: ``4040``
-- **user**: The Subsonic user. Default: ``admin``
-- **pass**: The Subsonic user password. Default: ``admin``
```
sampsyo: These ones (`user` and `pass`) are still valid. :smiley:

docs/plugins/subsonicupdate.rst
sampsyo: Cool! Buy maybe this sentence should go below, in the bullet for the `pass` configuration option, rather than by itself here.

beetsplug/subsonicupdate.py
```diff
@@ -46,42 +89,32 @@ def __init__(self):
             'user': 'admin',
             'pass': 'admin',
             'contextpath': '/',
+            'url': 'http://localhost:4040'
```
sampsyo:
```suggestion
            'url': 'http://localhost:4040',
```

beetsplug/subsonicupdate.py
```diff
@@ -25,16 +25,59 @@
...
+    salt = "".join([random.choice(r) for _ in range(6)])
+    salted_password = password + salt
+    token = hashlib.md5().update(salted_password.encode('utf-8')).hexdigest()
+
+    # Put together the payload of the request to the server and the URL
+    return salt, token
+
+
+def format_url():
+    """ Get the Subsonic URL to trigger a scan. Uses either the url
```
sampsyo:
```suggestion
    """Get the Subsonic URL to trigger a scan. Uses either the url
```

beetsplug/subsonicupdate.py
```diff
@@ -25,16 +25,59 @@
...
+import requests
+
+from beets import config
+from beets.plugins import BeetsPlugin
 
 __author__ = 'https://github.com/maffo999'
 
 
+def create_token():
+    """ Creates salt and token from given password.
```
sampsyo:
```suggestion
    """Creates salt and token from given password.
```

Typically, the docstring style doesn't use a space after the opening """.

sampsyo:
Awesome; thank you again for seeing this through to its conclusion! :sparkles: :sparkles: Merged with a changelog entry.

jef:
Thanks for the support! I was fun doing Python for the first time.

Issue #3391, opened 2019-10-06
fetchart: Add itunes "high resolution" artwork


texneus:
### Use case
Use beets to tag, catalog, and fetch artwork.  The FetchArt plugin is enabled to gather album art work.  I have found [this website](https://bendodson.com/projects/itunes-artwork-finder/) that pulls itunes artwork giving the option of both standard (600x600) and hi-res coverart.  The resolution of the high resolution artwork varies, but from what I am seeing is almost always 1500x1500 minimum, and new-ish releases are much higher (for example The Cranberries "In the End" is a 4000x4000).  Besides the size, the artwork is very high quality.

Unfortunately, Beets/FetchArt does not seem to be able to find the high-resolution artwork.  To duplicate, install the FetchArt plugin, and configure FetchArt SOURCES to use only itunes, and set MINWIDTH to 1000.  Verify on the [iTunes Artwork Finder](https://bendodson.com/projects/itunes-artwork-finder/) that high-resolution itunes artwork exists for a target album (for example:  The Cranberries "In the End"), although if there is itunes artwork I have yet to find a case where a high-resolution artwork does not also exist.  FetchArt will fail to find any artwork.

### Solution
Enable FetchArt to fetch high resolution artwork from itunes.  The subject webpage offers his code (in JavaScript and PHP) on GitHub [here](https://github.com/bendodson/itunes-artwork-finder) for examination.

### Alternatives
The only alternative I am aware of is to manually fetch the artwork.

sampsyo:
Sounds great! This seems like a good idea.

wisp3rwind:
Assigned this to myself; however: I will not have time to look at this at least for another month, so if anyone else wants to go ahead, please do so

mathieupinet:
I just had a look at this- it seems the only change would be to fetch an image filename ending with `100000x100000-999` rather than `1200x1200` in fetchchart(as that's what the website cited in the PR retrieves) , in this line as well as L523:
https://github.com/beetbox/beets/blob/9f9af300e8de7be7b1bb6575e4ccb4464fd1ca34/beetsplug/fetchart.py#L513

I succeeded in getting the 4000x4000 _In the End_ cover art with this simple fix, but interestingly the image filesize was 20.7MB (as opposed to 608KB for the 1200x1200 version).

Therefore it probably shouldn't be the default for the iTunes source, as it has wider implications for disk and bandwidth usage than a user would expect?

sampsyo:
[...]

PR #3463, opened 2019-12-26
fetchart: New high_resolution config option (#3391)

mathieupinet:
Implementing fetching of high-res iTunes artwork, as discussed in #3391.

---
This change is [Reviewable](https://reviewable.io/reviews/beetbox/beets/3463)

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -470,14 +470,6 @@ class ITunesStore(RemoteArtSource):
     NAME = u"iTunes Store"
     API_URL = u'https://itunes.apple.com/search'
 
-    def __init__(self, *args, **kwargs):
-        super(ITunesStore, self).__init__(*args, **kwargs)
-        high_resolution = self._config['high_resolution'].get()
-        if high_resolution:
-            self.image_suffix = '100000x100000-999'
-        else:
-            self.image_suffix = '1200x1200bb'
-
     def get(self, album, plugin, paths):
         """Return art URL from iTunes Store given an album title.
         """
@@ -513,13 +505,18 @@ class ITunesStore(RemoteArtSource):
                             payload['term'])
             return
 
+        if self._config['high_resolution'].get():
+            image_suffix = '100000x100000-999'
+        else:
+            image_suffix = '1200x1200bb'
+
         for c in candidates:
             try:
                 if (c['artistName'] == album.albumartist
                         and c['collectionName'] == album.album):
                     art_url = c['artworkUrl100']
                     art_url = art_url.replace('100x100bb',
-                                              self.image_suffix)
+                                              image_suffix)
                     yield self._candidate(url=art_url,
                                           match=Candidate.MATCH_EXACT)
             except KeyError as e:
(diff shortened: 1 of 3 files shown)


beetsplug/fetchart.py
```diff
@@ -802,6 +813,7 @@ def __init__(self):
         self.cover_names = list(map(util.bytestring_path, cover_names))
         self.cautious = self.config['cautious'].get(bool)
         self.store_source = self.config['store_source'].get(bool)
+        self.high_resolution = self.config['high_resolution'].get(bool)
```
sampsyo: This (`self.high_resolution`) currently seems to be unused. We can either delete it or use it in directly in the code that needs the flag by using `plugin.high_resolution` instead of a `config[...].get(...)` call.

beetsplug/fetchart.py
```diff
@@ -470,6 +470,14 @@ class ITunesStore(RemoteArtSource):
     NAME = u"iTunes Store"
     API_URL = u'https://itunes.apple.com/search'
 
+    def __init__(self, *args, **kwargs):
+        super(ITunesStore, self).__init__(*args, **kwargs)
+        high_resolution = self._config['high_resolution'].get()
+        if high_resolution:
+            self.image_suffix = '100000x100000-999'
+        else:
+            self.image_suffix = '1200x1200bb'
```
sampsyo: Because `self.image_suffix` is only used in one location (i.e., the `get` method), maybe it makes more sense to do this work here. That is, we could avoid overriding `__init__` and the `super` baggage that comes with, along with the complexity of an additional configuration field, by just computing a local `image_suffix` variable inside the `get` method. This would just mean copying and pasting the `if` condition into `get` and replacing `self.image_suffix` with just `image_suffix`.

docs/changelog.rst
```diff
@@ -6,7 +6,11 @@ Changelog
 
 New features:
 
-* :doc:`plugins/discogs` now adds two extra fields: `discogs_labelid` and 
+* :doc:`plugins/fetchart`: Added new ``high_resolution`` config option to
```
```suggestion
* :doc:`plugins/fetchart`: Added a new ``high_resolution`` config option to
```

docs/changelog.rst
```diff
@@ -6,7 +6,11 @@ Changelog
...
+  allow downloading of higher resolution iTunes artwork (at the expense of
+  file size)
```
```suggestion
  file size).
```

docs/plugins/fetchart.rst
```diff
@@ -67,6 +67,9 @@ file. The available options are:
 - **store_source**: If enabled, fetchart stores the artwork's source in a
   flexible tag named ``art_source``. See below for the rationale behind this.
   Default: ``no``.
+- **high_resolution**: If enabled, fetchart retrieves artwork in the highest
+  resolution it can find (Warning: image files can sometimes reach >20MB)
```
sampsyo:
```suggestion
  resolution it can find (warning: image files can sometimes reach >20MB).
```

beetsplug/fetchart.py
mathieupinet: makes sense! Patched in 2593a5b.

beetsplug/fetchart.py
```diff
@@ -505,12 +505,18 @@ def get(self, album, plugin, paths):
                             payload['term'])
             return
 
+        if self._config['high_resolution'].get():
```
sampsyo:
```suggestion
        if self._config['high_resolution']:
```

The Confuse library is smart enough to turn this into a `bool` automatically without calling `get()`.

mathieupinet: Ah, good to know. Thanks for reviewing so quickly!

PR #3467, opened 2020-01-12
support for keyfinder-cli, which doesn't want the -f flag for paths

BrainDamage:
Added support for [keyfinder-cli](https://github.com/EvanPurkhiser/keyfinder-cli) to the keyfinder plugin, which doesn't want/rejects the -f flag before a path

### DIFF
--- a/beetsplug/keyfinder.py
+++ b/beetsplug/keyfinder.py
@@ -18,6 +18,7 @@
 
 from __future__ import division, absolute_import, print_function
 
+import os.path
 import subprocess
 
 from beets import ui
@@ -52,15 +53,19 @@ class KeyFinderPlugin(BeetsPlugin):
 
     def find_key(self, items, write=False):
         overwrite = self.config['overwrite'].get(bool)
-        bin = self.config['bin'].as_str()
+        command = [self.config['bin'].as_str()]
+        # The KeyFinder GUI program needs the -f flag before the path.
+        # keyfinder-cli is similar, but just wants the path with no flag.
+        if 'keyfinder-cli' not in os.path.basename(command[0]).lower():
+            command.append('-f')
 
         for item in items:
             if item['initial_key'] and not overwrite:
                 continue
 
             try:
-                output = util.command_output([bin, '-f',
-                                              util.syspath(item.path)]).stdout
+                output = util.command_output(command + [util.syspath(
+                                                        item.path)]).stdout
             except (subprocess.CalledProcessError, OSError) as exc:
                 self._log.error(u'execution failed: {0}', exc)
                 continue
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,6 +6,8 @@ Changelog
 
 New features:
 
+* :doc:`plugins/keyfinder`: Added support for `keyfinder-cli`_
+  Thanks to :user:`BrainDamage`.
 * :doc:`plugins/fetchart`: Added a new ``high_resolution`` config option to
   allow downloading of higher resolution iTunes artwork (at the expense of
   file size).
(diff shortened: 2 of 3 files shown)


sampsyo:
Cool; seems like a good idea overall! Here are a couple of suggestions:

- We should probably mention in the plugin's documentation that it also works with the new tool. And that the CLI compatibility relies on the command being named `keyfinder-cli`, which would probably not otherwise be obvious. (Leaving a detailed comment to this effect in the code would probably be a good idea too…)
- Instead of adding an empty string `''` to the command line, it would be cleaner to not add anything at all… for example, the command list could be built up like `cmd = [bin]; if not is_keyfinder_cli: cmd.append('-f'); cmd.append(path)`. Not all commands will take kindly to having an empty argument like that in the arguments. (Try doing `ls ''`, for example.)
- Would you mind adding a changelog entry?

BrainDamage:
I've added all the suggested changes. I am not familiar with rst's syntax, so I might've done some mistakes, apologies in advance.

beetsplug/keyfinder.py
```diff
@@ -52,15 +53,19 @@ def imported(self, session, task):
 
     def find_key(self, items, write=False):
         overwrite = self.config['overwrite'].get(bool)
-        bin = self.config['bin'].as_str()
+        command = [self.config['bin'].as_str()]
+        # if using KeyFinder GUI program, the path must be prefixed with -f flag
+        # keyfinder-cli instead wants the path bare
```
sampsyo: We tend to use full sentences in beets comments. Here's a suggested way to phrase it:

> The KeyFinder GUI program needs the `-f` flag before the path. keyfinder-cli is similar but just wants the path with no flag.

docs/changelog.rst
```diff
@@ -6,6 +6,8 @@ Changelog
 
 New features:
 
+* :doc:`plugins/keyfinder`: Added support for `keyfindercli`_
```
sampsyo: I know this may seem overly pedantic, but it seems like the project is named `keyfinder-cli` rather than `keyfindercli`?

docs/plugins/keyfinder.rst
```diff
@@ -24,9 +24,11 @@ configuration file. The available options are:
   a path to the binary. If you installed the KeyFinder GUI on a Mac, for
   example, you want something like
   ``/Applications/KeyFinder.app/Contents/MacOS/KeyFinder``.
+  If using `keyfindercli`_, the binary must be named keyfinder-cli.
```
sampsyo:
```suggestion
  If using `keyfinder-cli`_, the binary must be named ``keyfinder-cli``.
```

docs/plugins/keyfinder.rst
```diff
@@ -1,8 +1,8 @@
 Key Finder Plugin
 =================
 
-The `keyfinder` plugin uses the `KeyFinder`_ program to detect the
-musical key of track from its audio data and store it in the
+The `keyfinder` plugin uses either the `KeyFinder`_ or `keyfindercli`_ program
```
sampsyo:
```suggestion
The `keyfinder` plugin uses either `KeyFinder`_ or `keyfinder-cli`_
```

docs/plugins/keyfinder.rst
```diff
@@ -1,8 +1,8 @@
...
+to  detect the musical key of track from its audio data and store it in the
```
sampsyo:
```suggestion
to  detect the musical key of a track from its audio data and store it in the
```

docs/plugins/keyfinder.rst
```diff
@@ -24,9 +24,11 @@ configuration file. The available options are:
   a path to the binary. If you installed the KeyFinder GUI on a Mac, for
```
sampsyo: Perhaps the sentence above should read:

> The name of the program use for key analysis. You can use either 
\`KeyFinder\`_ or \`keyfinder-cli\`_.

BrainDamage:
done, I hope I didn't miss anything

PR #3470, opened 2020-01-15
WIP: plugin to stream from remote beetsplug.web server instance

droserasprout:
This plugin is a **beetsplug.play** with ability added to query **beetsplug.web** server instead of local beets library. Fetched URLs then passed to configured music player for streaming.

Configuration:
```
play:
    servers:
        local: http://127.0.0.1:8337
        some_alias: http://192.168.12.34:8337
```

Usage:
```
beet remote -s some_alias No Slow Rolling
```

Could this feature be useful for someone except me? If yes should it be a separate plugin or a beetsplug.play part?


sampsyo:
Wow; that's super cool! It's a very neat idea. I'd be interested to hear from others if they'd like to use it.

I'm sort of divided about where it should live:

- On the one hand, the way it has to interact with a _remote_ beets instance and works by _sending a query there_ rather than using the local library makes it quite different.
- On the other hand, it has to share the same functionality to actually play the music once it has obtained it.

Maybe adding it to the play plugin to begin with and then, perhaps over time, considering letting it live "on its own" later if other functionality also wants to query remote libraries?

Also just for fun, I thought I would point toward our fledging, semi-stalled effort to build a larger ecosystem around remote libraries via a new protocol: https://github.com/beetbox/aura

droserasprout:
Merged plugin with `beet play`. Can't manage to run test suite on macOS, so no tests yet.

jtpavlock:
@droserasprout thank you for your contribution and welcome to `beets`!

I don't run macOS, but maybe I can try to help figuring out setting up `tox` if you still are interested in working on this?

PR #3474, opened 2020-01-23
Support extra tags for MusicBrainz queries

x1ppy:
Here's a first go at the "hints" behavior discussed in https://discourse.beets.io/t/providing-hints-to-autotagger/1072/10.

This adds a single `extra_tags` config parameter that, when enabled, includes the following tags in the MusicBrainz query:  `media`, `year`, `country`, `label`, `catalognum`. The selection of tags to include is somewhat arbitrary, but based on a combination of the tags I needed in my particular use case and the tags already used by `likelies` in `current_metadata`.

### DIFF
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -614,17 +614,21 @@ def tracks_for_id(track_id):
 
 
 @plugins.notify_info_yielded(u'albuminfo_received')
-def album_candidates(items, artist, album, va_likely):
+def album_candidates(items, artist, album, va_likely, extra_tags):
     """Search for album matches. ``items`` is a list of Item objects
     that make up the album. ``artist`` and ``album`` are the respective
     names (strings), which may be derived from the item list or may be
     entered by the user. ``va_likely`` is a boolean indicating whether
-    the album is likely to be a "various artists" release.
+    the album is likely to be a "various artists" release. ``extra_tags``
+    is an optional dictionary of additional tags used to further
+    constrain the search.
     """
+
     # Base candidates if we have album and artist to match.
     if artist and album:
         try:
-            for candidate in mb.match_album(artist, album, len(items)):
+            for candidate in mb.match_album(artist, album, len(items),
+                                            extra_tags):
                 yield candidate
         except mb.MusicBrainzAPIError as exc:
             exc.log(log)
@@ -632,13 +636,15 @@ def album_candidates(items, artist, album, va_likely):
     # Also add VA matches from MusicBrainz where appropriate.
     if va_likely and album:
         try:
-            for candidate in mb.match_album(None, album, len(items)):
+            for candidate in mb.match_album(None, album, len(items),
+                                            extra_tags):
                 yield candidate
         except mb.MusicBrainzAPIError as exc:
             exc.log(log)
 
     # Candidates from plugins.
-    for candidate in plugins.candidates(items, artist, album, va_likely):
+    for candidate in plugins.candidates(items, artist, album, va_likely,
+                                        extra_tags):
         yield candidate
 
 
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -447,6 +447,12 @@ def tag_album(items, search_artist=None, search_album=None,
             search_artist, search_album = cur_artist, cur_album
         log.debug(u'Search terms: {0} - {1}', search_artist, search_album)
 
+        extra_tags = None
+        if config['musicbrainz']['extra_tags']:
+            tag_list = config['musicbrainz']['extra_tags'].get()
+            extra_tags = {k: v for (k, v) in likelies.items() if k in tag_list}
+            log.debug(u'Additional search terms: {0}', extra_tags)
+
         # Is this album likely to be a "various artist" release?
         va_likely = ((not consensus['artist']) or
                      (search_artist.lower() in VA_ARTISTS) or
@@ -457,7 +463,8 @@ def tag_album(items, search_artist=None, search_album=None,
         for matched_candidate in hooks.album_candidates(items,
                                                         search_artist,
                                                         search_album,
-                                                        va_likely):
+                                                        va_likely,
+                                                        extra_tags):
             _add_candidate(items, candidates, matched_candidate)
 
     log.debug(u'Evaluating {0} candidates.', len(candidates))
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -38,6 +38,14 @@ else:
 
 SKIPPED_TRACKS = ['[data track]']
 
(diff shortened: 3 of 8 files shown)


beets/autotag/match.py
```diff
@@ -447,6 +447,12 @@ def tag_album(items, search_artist=None, search_album=None,
             search_artist, search_album = cur_artist, cur_album
         log.debug(u'Search terms: {0} - {1}', search_artist, search_album)
 
+        extra_tags = None
+        if config['musicbrainz']['extra_tags']:
+            fields = {u'year', u'catalognum', u'country', u'media', u'label'}
```
sampsyo: We could even make the configuration option a list of fields that is by default empty. Then the user could specify exactly the set of fields to include, which could help with experimentation.

beets/autotag/mb.py
```diff
@@ -429,6 +429,28 @@ def match_album(artist, album, tracks=None):
...
+        if label:
+            criteria['label'] = six.text_type(label)
+
+        media = extra_tags.get('media', '').lower().strip()
+        if media:
+            criteria['format'] = six.text_type(media)
+
+        year = extra_tags.get('year', '')
+        if year:
+            criteria['date'] = six.text_type(year)
```
sampsyo: Maybe this could be implemented with a loop (except for the special-case handling of `catalognum`)? Something like this:

```py
FIELDS_TO_KEYS = {
    'media': 'format',
    'country': 'country',
}
for field, key in FIELDS_TO_KEYS.items():
    value = extra_tags.get(key, '').lower().strip()
    if value:
        criteria[field] = six.text_type(value)
```

beets/autotag/mb.py
```diff
@@ -429,6 +429,28 @@ def match_album(artist, album, tracks=None):
     if tracks is not None:
         criteria['tracks'] = six.text_type(tracks)
 
+    # Additional search cues from existing metadata.
+    if extra_tags is not None:
+        catalognum = extra_tags.get('catalognum', '').lower().strip()
+        if re.search(u'^[0-9a-z -]+$', catalognum):
```
sampsyo: This regex filter seems like a reasonably good idea, but just checking: is this based on a guess about what most catalog numbers look like, or is this based on some manner of MusicBrainz guidance?

beets/autotag/hooks.py
```diff
@@ -614,31 +614,34 @@ def tracks_for_id(track_id):
 
 
 @plugins.notify_info_yielded(u'albuminfo_received')
-def album_candidates(items, artist, album, va_likely):
+def album_candidates(items, artist, album, va_likely, extra_tags=None):
```
sampsyo: This is not directly relevant to this PR, but maybe someday we should remove `artist` and `album` as special parameters and just make them entries in a single, flat `search_tags` dict.

x1ppy: I started going down this road, but stopped once the change crept its way into [`plugins.py`](https://github.com/beetbox/beets/blob/fdbd95db21b8f2a686f8e0ce034e78c0f04f687e/beets/plugins.py#L175) since I'm not familiar with your protocol on breaking changes for plugins. Considering the scope, I imagine it's probably best left for a separate PR anyway.

beets/autotag/mb.py
x1ppy: Honestly, I don't remember why I added that regex check. I don't think there's any reason to safeguard against bad catalog numbers in particular when other fields aren't validated, so I just removed this.

Note that we do still need the special handling of `catno.replace(u' ', '')` since MediaBrainz appears to be unable to match catalog numbers containing spaces.

x1ppy:
I put this on hold far longer than intended, but finally got around to making the suggested changes. Please take another look at your leisure; thanks!

sampsyo:
Oops, I spoke too soon, however—it looks like the tests are failing:
https://travis-ci.org/github/beetbox/beets/jobs/664226712?utm_medium=notification&utm_source=github_status

I believe the problem is the mocking of the MusicBrainz module we do for importer tests:
https://github.com/beetbox/beets/blob/b34d1f71a93b3663f70f46a4d5ed08702b14c5ca/test/test_importer.py#L64

Any chance I could convince you to add the extra parameter to `match_album` there as well?

x1ppy:
I actually spent a few minutes yesterday trying to check the tests before I submitted, but wasn't able to get them running.

Here's what I tried:
~~~
git clone https://github.com/beetbox/beets.git
cd beets
mkvirtualenv beets-test
beets-test ❯ pip install -e .
beets-test ❯ pip install nose
~~~

Then I see a bunch of errors about missing test dependencies: `ERROR: Failure: ModuleNotFoundError (No module named 'pylast')`, `flask`, `requests`, etc.

I see these are all listed under [`tests_require`](https://github.com/beetbox/beets/blob/master/setup.py#L112), but that's where I got stuck. What's the recommended way to install the test dependencies? I could install these packages manually, but figured there's probably some command I'm missing.

sampsyo:
A good way to get started, which we should document better [on the wiki](https://github.com/beetbox/beets/wiki/Hacking), would be to use [tox](https://tox.readthedocs.io/en/latest/). If you just install tox and run the command in the beets directory, it should install everything it needs in a virtualenv automatically.

x1ppy:
Unfortunately, I'm not getting very far that way, either:
~~~
$> pip install tox
$> tox
GLOB sdist-make: /Users/x1ppy/sources/beets/setup.py
ERROR: invocation failed (exit code 1), logfile: /Users/x1ppy/sources/beets/.tox/log/GLOB-0.log
================================================================================================================= log start ==================================================================================================================
sphinx-build -b man -d _build/doctrees   . _build/man
make: sphinx-build: No such file or directory
make: *** [man] Error 1
  File "setup.py", line 54, in <module>
    build_manpages()
  File "setup.py", line 38, in build_manpages
    subprocess.check_call(['make', 'man'])
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/subprocess.py", line 363, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['make', 'man']' returned non-zero exit status 2.

================================================================================================================== log end ===================================================================================================================
ERROR: FAIL could not package project - v = InvocationError('/usr/local/opt/python/bin/python3.7 setup.py sdist --formats=zip --dist-dir /Users/x1ppy/sources/beets/.tox/dist', 1)
~~~

x1ppy:
Change made:
```
-    def match_album(self, albumartist, album, tracks):
+    def match_album(self, albumartist, album, tracks, extra_tags):
```

sampsyo:
Excellent! Thank you!! :fish: :rocket:

PR #3477, opened 2020-01-28
Remove --replaygain flag when checking bs1770gain availability

ybnd:
`bs1770gain` is marked as unavailable when it shouldn't be because `bs1770gain --replaygain` returns an error.

### DIFF
--- a/test/test_replaygain.py
+++ b/test/test_replaygain.py
@@ -40,7 +40,7 @@ if any(has_program(cmd, ['-v']) for cmd in ['mp3gain', 'aacgain']):
 else:
     GAIN_PROG_AVAILABLE = False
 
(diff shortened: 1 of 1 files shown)


sampsyo:
Huh; that's odd! Any guesses about why this was here in the first place? Was it necessary in previous versions, for example?

I'll merge this now in any case. :smiley:

ybnd:
I haven't used the previous versions so I don't know for sure, but it could be.

I've been having some more problems (seems like the format of the output XML changed, issue incoming on that) so it does seem that there were some major changes to `bs1770gain` at some point.

But I'm new to both `bs1770gain` and the replaygain plugin, so only guesses...

Issue #2224, opened 2016-10-11
replaygain: Analyze in parallel


pablocouto:
### Problem

The `replaygain` plugin doesn’t appear to make use of multicore capacities. On my machine, I see increased activity on only one core out of four.

In my opinion, it would be useful to be able to configure the plugin to analyze different albums or tracks at the same time.
### Setup
- OS: Arch Linux
- Python version: 2.7.12
- beets version: 1.3.18
- Turning off plugins made problem go away (yes/no): n/a

My configuration (output of `beet config`) is:

``` yaml
replaygain:
    backend: gstreamer
    auto: yes
    targetlevel: 89
    overwrite: no
embedart:
    remove_art_file: no
    compare_threshold: 0
    auto: yes
    ifempty: no
    maxwidth: 0

plugins: fromfilename embedart fetchart mbsync replaygain scrub duplicates missing random
directory: ~/Media/music

import:
    copy: no
    write: no
library: ~/Media/musiclibrary.blb
scrub:
    auto: yes
missing:
    count: no
    total: no
fetchart:
    auto: yes
    minwidth: 0
    sources:
    - filesystem
    - coverart
    - itunes
    - amazon
    - albumart
    google_engine: 001442825323518660753:hrh5ch1gjzm
    enforce_ratio: no
    cautious: no
    maxwidth: 0
    store_source: no
    google_key: REDACTED
    fanarttv_key: REDACTED
    cover_names:
    - cover
    - front
    - art
    - album
    - folder
duplicates:
    count: no
    full: no
    format: ''
    keys: []
    move: ''
    tag: ''
    path: no
    copy: ''
    tiebreak: {}
    album: no
    strict: no
    checksum: ''
    merge: no
    delete: no
```

sampsyo:
Good call. See also #673 and #663.

ybnd:
Do you think it makes sense to have this be specific to the replaygain plugin, without going further into https://github.com/beetbox/beets/pull/673 and https://github.com/beetbox/beets/pull/663?

sampsyo:
Sure, at least as a first step! We can always generalize it later.

nagisa:
I've been trying to call `beet replaygain --threads=32` but only one core is being used, still.

ybnd:
Which backend are you using? 
`gstreamer` and `audiotools` don't support parallel processing currently.

nagisa:
I tried gstreamer and then ffmpeg.

PR #3478, opened 2020-01-28
Implement parallel replaygain analysis

ybnd:
Implements [#2224](https://github.com/beetbox/beets/issues/2224)

* Add `--jobs` or `-j` to `replaygain` to set the pool size
* Single-threaded execution by default, if `--jobs` is unset

* If multithreaded, calls `Backend.compute_album_gain` or `Backend.compute_track_gain` asynchronously with metadata storing/writing in the callback


beetsplug/replaygain.py
```diff
@@ -1381,17 +1433,30 @@ def commands(self):
         def func(lib, opts, args):
             write = ui.should_write(opts.write)
             force = opts.force
+            jobs = opts.jobs
```
sampsyo: Instead of defaulting to off when jobs = 0, let's use our `cpu_count` utility:
https://github.com/beetbox/beets/blob/c1937e1b01d7130a27cdb3a5f589bb4f1bab9ce6/beets/util/__init__.py#L760

beetsplug/replaygain.py
```diff
@@ -1381,17 +1433,30 @@ def commands(self):
...
+            if hasattr(self, 'pool'):
+                self.pool.close()
+                self.pool.join()
+                self.pool.terminate()
+
         cmd = ui.Subcommand('replaygain', help=u'analyze for ReplayGain')
         cmd.parser.add_album_option()
+        cmd.parser.add_option(
+            "-j", "--jobs", dest="jobs", type=int, default=0,
+            help=u"worker pool size"
```
sampsyo: Perhaps a more helpful description:
```suggestion
            help=u"number of threads for parallel analysis"
```

beetsplug/replaygain.py
```diff
@@ -104,6 +104,8 @@ class Backend(object):
     """An abstract class representing engine for calculating RG values.
     """
 
+    do_parallel = False
```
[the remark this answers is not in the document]
ybnd: I don't know if it's dangerous dangerous, but I just haven't been able to get `gstreamer` to work in parallel with `ThreadLoop`. It looses track of threads non-determenistically; sometimes it writes one (or a couple) of the files, sometimes it throws 
image 
and sometimes it just fails silently. 

And for now, I haven't been able to test it with `audiotools` since it doesn't want to install on my machine for some reason.

I'm going to look into both of these issues, but I wanted to see if the whole thing would be useful first.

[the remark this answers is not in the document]
sampsyo: Sound good! I think it's fine to ship this feature before we figure out how to parallelize GStreamer.

beetsplug/replaygain.py
ybnd: Should we do something like default = math.ceil(cpu_count() / 4) to keep CPU usage reasonable by default for large queries?

beetsplug/replaygain.py
ybnd: Ok.

beetsplug/replaygain.py
```diff
@@ -1173,6 +1180,7 @@ def __init__(self):
         self.overwrite = self.config['overwrite'].get(bool)
         self.per_disc = self.config['per_disc'].get(bool)
         backend_name = self.config['backend'].as_str()
+
```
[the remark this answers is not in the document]
ybnd: Sorry about that, bad habit.

beetsplug/replaygain.py
```diff
@@ -1344,23 +1381,38 @@ def handle_track(self, item, write, force=False):
...
+                        item)
                 )
 
             store_track_gain(item, track_gains[0])
             if write:
                 item.try_write()
+            self._log.debug(u'done analyzing {0}', item)
+
+        try:
+            if hasattr(self, 'pool'):
```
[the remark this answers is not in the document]
ybnd: Yeah, that would be much better. Thanks!

beetsplug/replaygain.py
ybnd: And maybe have it default to no parallelism if configured with `threaded: no`?

sampsyo: Maybe the best policy would be to match the behavior of the `convert` plugin:
https://github.com/beetbox/beets/blob/master/beetsplug/convert.py#L119

The option is named `threads` there, and it does default to the CPU count. I think it's probably OK to default to that? If people want to run in the background without disruption, it's easy to turn down the count…

ybnd:
Ok, so I remembered why I made it default to non-parallel :)

While it *does* work with my on-disk library, it fails in `test_replaygain.py` on my machine, maybe because there the library is `:in_memory:`? For some reason it works in CI though.

The Travis build fails because I'm not raising errors to the main thread, working on that.

beetsplug/replaygain.py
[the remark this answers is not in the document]
ybnd: Implemented in https://github.com/beetbox/beets/pull/3478/commits/388d2d2c0de0b7bed66d661645b6adead79c3c09, with the exception of the optional `pool` argument, do the "pool or no pool" check in `_has_pool` instead. The plugin should still have `pool` as an attribute for parallel execution to work at import ~ https://github.com/beetbox/beets/pull/3478/commits/79c5535cf6ea80993ae1d2f00ea15911f2581c3c.

beetsplug/replaygain.py
```diff
@@ -1378,20 +1463,13 @@ def imported(self, session, task):
     def commands(self):
         """Return the "replaygain" ui subcommand.
         """
-        def func(lib, opts, args):
```
sampsyo: We generally keep these as nested functions in most plugins, just because they tend to be so simple.

beetsplug/replaygain.py
```diff
@@ -1402,5 +1480,24 @@ def func(lib, opts, args):
...
+        cmd.func = self.replaygain_func
         return [cmd]
+
+    def replaygain_func(self, lib, opts, args):
+        """Handle `replaygain` ui subcommand
+        """
+        write = ui.should_write(opts.write)
+        force = opts.force
+
+        if opts.threads != 0:
```
sampsyo: It seems like this condition might be supposed to be `!= 1`? Or better yet, the condition should come after checking the config file too?

beetsplug/replaygain.py
```diff
@@ -1312,21 +1333,35 @@ def handle_album(self, album, write, force=False):
...
+                                                album_gain.track_gains):
+                        store_track_gain(item, track_gain)
+                        store_album_gain(item, album_gain.album_gain)
+                        if write:
+                            item.try_write()
+                        self._log.debug(u'done analyzing {0}', item)
+                except ReplayGainError as e:
+                    self._log.info(u"ReplayGain error: {0}", e)
+                except FatalReplayGainError as e:
+                    raise ui.UserError(u"Fatal ReplayGain error: {0}".format(e))
```
sampsyo: I’m not entirely clear about why we need these exception handlers. Aren’t we just storing the data at this point, rather than doing any analysis? So is it really possible for this code to raise these exceptions?

beetsplug/replaygain.py
ybnd: * I wanted to include the option to completely bypass `ThreadPool` from the CLI with `--threads 0`, seems like a useful feature down the line in case we ever suspect the parallel processing is messing something up
* `ThreadPool(1)` still provides one worker thread that can work asynchronously which may be useful in some cases? I don’t know.

beetsplug/replaygain.py
ybnd: You're right, my bad.

sampsyo:
Looking pretty good! I assume we should probably figure out the exception logging thing before merging?

ybnd:
Added an exception handling thread.

`ReplayGainError` exceptions raised in the worker threads jost get logged, while `FatalReplayGainError` and other exceptions are reraised in the main thread.

jtpavlock:
Pinging @sampsyo in case he didn't see the last request for review.

Did you have anything else to add to this?

ybnd:
Fixed the conflicts on my end.

Some tests are failing though, fixing those now.

beetsplug/replaygain.py
```diff
@@ -1307,30 +1309,39 @@ def album_requires_gain(self, album):
...
                 for item in album.items()])
 
+    def _store(self, item):
+        try:
+            item.store()
+        except OperationalError:
+            # test_replaygain.py :memory: library can fail with
+            #    `sqlite3.OperationalError: no such table: items`
+            # but the second attempt succeeds
+            item.store()
```
ybnd, line 1319: As far as I can tell, the exception is never thrown when working with an 'actual' database.

sampsyo:
So sorry for the delay on this. Everything looks awesome, for the most part—I have no specific code-level comments on the actual parallelization.

However, I do think we need to do something about the `OperationalError` handler that @ybnd highlighted. `OperationalError` is a very generic exception and can arise in lots of cases where we really do want to know something went wrong with the database—for example, if we make an error in generating the SQL query. Is there any way we can move this handler/suppression to the test itself instead of putting it here in the plugin?

ybnd:
@sampsyo No problem, I know you've been super busy :)

Yeah, that would be way more sensible. I'll look into it!

beetsplug/replaygain.py
```diff
@@ -1310,13 +1309,14 @@ def album_requires_gain(self, album):
...
-            # but the second attempt succeeds
-            item.store()
+        """Store an item to the database.
+             When testing, item.store() sometimes fails non-destructively with
+             sqlite.OperationalError.
+             This method is here to be patched to a retry-once helper function
+             in test_replaygain.py, so that it can still fail appropriately
+             outside of these tests.
+        """
+        item.store()
```
ybnd, line 1319: Removed the `OperationalError` handler, added an explanation as to why this seemingly useless method is there

test/test_replaygain.py
```diff
@@ -55,10 +59,28 @@ def reset_replaygain(item):
...
+    try:
+        item.store()
+    except OperationalError:
+        # test_replaygain.py :memory: library can fail with
+        #    `sqlite3.OperationalError: no such table: items`
+        # but the second attempt succeeds
+        item.store()
+
+
+@patch.object(ReplayGainPlugin, '_store', _store_retry_once)
```
ybnd, line 82: As far as I can see there's no way to `try/except` on a higher level; the "false alarm" happens at *every* item, so it looks like we still need to ignore it for every item...

ybnd (commented):
I've pulled the `sqlite3.OperationalError` thing into `test_replaygain.py` as suggested.

Somewhat annoyingly this whole workaround is not "necessary" for the CI builds since they just skip all of the real tests...
I double-checked with the master branch and it's this PR that's causing this, for sure.

How do you think I should proceed?
* Leave it like this and submit an issue (which you could assign to me so I could keep looking into it)
* Look into it *before* going forward with this PR

sampsyo:
Awesome. This is really mysterious, but I like your temporary workaround to make the tests go through. Since you're comfortable with it, I'd like to forge ahead and merge this. Would you mind opening a new issue to get to the bottom of what's going on?

Many thanks once again!!

skapazzo:
This seems to give me problems: replaygain metadata doesn't get written to file on import.

If I import an album with `replaygain: auto: yes` and any backend that has `do_parallel = True` replaygain values get calculated and saved in the db but not written to the files.
If I set `do_parallel = False` for the backend I'm using the metadata gets written as expected; setting `threads = 0` doesn't help.

jackwilsdon:
@skapazzo could you please open a new issue for this?

skapazzo:
> @skapazzo could you please open a new issue for this?

Sure, I'll do it straight away.

Issue #3479, opened 2020-01-29
replaygain: bs1770gain v0.6.0 XML output can't be parsed


ybnd:
### replaygain: bs1770gain v0.6.0 XML output can't be parsed

Running `beet -vv replaygain` with the `bs1770gain` backend results in 
```
replaygain: analyzing Nyul -  When The Path Fades - Floatfall
replaygain: executing bs1770gain --ebu --xml -p /home/ybnd/code/beets/library/Nyul/2014 -  When The Path Fades/05 Floatfall.flac
replaygain: analysis finished: <!-- scanning 1 -->
<!-- analyzing ... -->
<bs1770gain norm="-23.00">
  <track total="1" number="1" file="05 Floatfall.flac">
    <integrated lufs="-10.99" lu="-12.01"/>
    <sample-peak spfs="-0.10" amplitude="0.99"/>
  </track>
  <integrated lufs="-10.99" lu="-12.01"/>
  <sample-peak spfs="-0.10" amplitude="0.99"/>
</bs1770gain>
<!-- done. -->

replaygain: ReplayGain error: The bs1770gain tool produced malformed XML. Using version >=0.4.10 may solve this problem.
```
for all tracks.

1. The actual XML output `text` that gets parsed at https://github.com/beetbox/beets/blob/ca3142b1d92d177a2d789775a4ff2ee5569b9ba6/beetsplug/replaygain.py#L318-L324 gets spliced with some progress percentages 
    ```
    b'<!-- scanning 1 -->
    <!-- analyzing ... -->
    <bs1770gain norm="-23.00">
      <track total="1" number="1" file="05 Floatfall.flac">
    0%\x08\x080%\x08\x080%\x08\x080%\x08\x08 ...
    ... 100%\x08\x08\x08\x08    \x08\x08\x08\x08    <integrated lufs="-10.99" lu="-12.01"/>
        <sample-peak spfs="-0.10" amplitude="0.99"/>
      </track>
      <integrated lufs="-10.99" lu="-12.01"/>
      <sample-peak spfs="-0.10" amplitude="0.99"/>
    </bs1770gain>
    <!-- done. -->'
    ```
2. The parser expects `factor` instead of `amplitude` in the `sample-peak` element, which *used to be* what the output looked like based on https://github.com/beetbox/beets/issues/2983): https://github.com/beetbox/beets/blob/ca3142b1d92d177a2d789775a4ff2ee5569b9ba6/beetsplug/replaygain.py#L296-L299
    

[...]

PR #3480, opened 2020-01-30
Update Bs1770gainBackend to handle bs1770gain v0.6.0 XML output

ybnd:
Fixes https://github.com/beetbox/beets/issues/3479.

### DIFF
--- a/beetsplug/replaygain.py
+++ b/beetsplug/replaygain.py
@@ -25,6 +25,7 @@ import enum
 import re
 import xml.parsers.expat
 from six.moves import zip
+from packaging import version
 
 from beets import ui
 from beets.plugins import BeetsPlugin
@@ -148,19 +149,18 @@ class Bs1770gainBackend(Backend):
 
         cmd = 'bs1770gain'
         try:
-            call([cmd, "--help"])
+            version_out = call([cmd, '--version'])
             self.command = cmd
-            try:
-                self.version = re.search(
-                    '([0-9]+.[0-9]+.[0-9]+), ',
-                    call([cmd, '--version']).stdout.decode('utf-8')
-                ).group(1)
-            except AttributeError:
-                self.version = '0.0.0'
+            self.version = re.search(
+                '([0-9]+.[0-9]+.[0-9]+), ',
+                version_out.stdout.decode('utf-8')
+            ).group(1)
         except OSError:
             raise FatalReplayGainError(
                 u'Is bs1770gain installed?'
             )
+        except AttributeError:
+            self.version = '0.0.0'
         if not self.command:
             raise FatalReplayGainError(
                 u'no replaygain command found: install bs1770gain'
@@ -263,7 +263,7 @@ class Bs1770gainBackend(Backend):
         cmd = [self.command]
         cmd += ["--" + method]
         cmd += ['--xml', '-p']
-        if self.version >= '0.6.0':
+        if version.parse(self.version) >= version.parse('0.6.0'):
             cmd += ['--unit=ebu']  # set units to LU
 
         # Workaround for Windows: the underlying tool fails on paths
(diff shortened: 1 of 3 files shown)


beetsplug/replaygain.py
```diff
@@ -312,10 +327,25 @@ def end_element_handler(name):
...
+                state['gain'] = state['peak'] = None
+
         parser.StartElementHandler = start_element_handler
         parser.EndElementHandler = end_element_handler
 
         try:
+            if type(text) == bytes:
+                text = text.decode('utf-8')
+            while '\x08' in text:
+                text = re.sub('[^\x08]\x08', '', text)
```
sampsyo: This replacement is somewhat surprising! Maybe a comment would be useful here to describe what’s going wrong?

beetsplug/replaygain.py
```diff
@@ -312,10 +327,25 @@ def end_element_handler(name):
...
+                        and album_state['peak'] is not None:
+                    album_gain["album"] = Gain(
+                        album_state['gain'], album_state['peak'])
+                state['gain'] = state['peak'] = None
+
         parser.StartElementHandler = start_element_handler
         parser.EndElementHandler = end_element_handler
 
         try:
+            if type(text) == bytes:
```
sampsyo:
```suggestion
            if isinstance(text, bytes):
```

beetsplug/replaygain.py
```diff
@@ -312,10 +327,25 @@ def end_element_handler(name):
...
+                text = text.decode('utf-8')
```
sampsyo: Is it necessary to do this conversion, or shall we perhaps just do the substitution directly on the bytes (with a bytes pattern)? That would avoid hard-coding an encoding, for example...

beetsplug/replaygain.py
```diff
@@ -312,10 +327,25 @@ def end_element_handler(name):
...
+            while '\x08' in text:
```
sampsyo: It might be safer to loop until the string stops changing. Otherwise, this runs the risk of this character being present but the pattern not matching (e.g., if it occurs at the beginning of the string).

beetsplug/replaygain.py
```diff
@@ -252,6 +260,8 @@ def compute_chunk_gain(self, items, is_album, target_level):
         cmd = [self.command]
         cmd += ["--" + method]
         cmd += ['--xml', '-p']
+        if self.version >= '0.6.0':
```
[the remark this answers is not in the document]
ybnd: Good point, it's not...
If we want to be really sure we could also go
``` python
$ from packaging import version
$ version.parse('0.10.0c') >= version.parse('0.6.0b')
True
```

beetsplug/replaygain.py
```diff
@@ -149,6 +150,13 @@ def __init__(self, config, log):
         try:
             call([cmd, "--help"])
             self.command = cmd
+            try:
+                self.version = re.search(
+                    '([0-9]+.[0-9]+.[0-9]+), ',
+                    call([cmd, '--version']).stdout.decode('utf-8')
+                ).group(1)
+            except AttributeError:
```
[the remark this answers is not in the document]
ybnd: Yes, it's the stdout attribute.
Ok, I suppose there's no real chance that some weird version `bs1770gain` *doesn't* handle `--version`.

beetsplug/replaygain.py
ybnd: To me it seems more readable to cover that case in the pattern by also replacing \x08 at the start of the string
```regex
b'[^\x08]\x08|^\x08'
```
then I can just explain it in a comment since I should write a bit about what's going on here anyway (:

beetsplug/replaygain.py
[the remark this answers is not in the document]
sampsyo: As nifty as this is, I would love to avoid using `packaging` because it is technically a new dependency—it is [not in the standard library](https://docs.python.org/3/library/packaging.html).

beetsplug/replaygain.py
```diff
@@ -147,12 +153,18 @@ def __init__(self, config, log):
...
             self.command = cmd
+            self.version = re.search(
+                '([0-9]+.[0-9]+.[0-9]+), ',
+                version_out.stdout.decode('utf-8')
+            ).group(1)
         except OSError:
             raise FatalReplayGainError(
                 u'Is bs1770gain installed?'
             )
+        except AttributeError:
```
sampsyo: Thanks for the reorganization, but I'm still not 100% sure when this exception can be raised. Isn't it the case that the `command_output` object will always have a `stdout` member if the call was successful?

beetsplug/replaygain.py
```diff
@@ -312,10 +334,36 @@ def end_element_handler(name):
...
         parser.EndElementHandler = end_element_handler
 
         try:
+            # Sometimes, the XML out put of `bs1770gain` gets spliced with
+            #   some progress percentages: b'9%\x08\x0810%\x08\x08'
+            #   that are supposed to be canceled out by appending
+            #   a b'\x08' backspace characters for every character,
+            #
+            # For some reason, these backspace characters don't get
+            #   resolved, resulting in mangled XML.
```
sampsyo: Aha; that's pretty weird! And there's no chance there's a "silent" flag we can pass to the tool to suppress the progress output?

ybnd: There is, I had overlooked it apparently!

beetsplug/replaygain.py
ybnd: There is an edge case in `test_replaygain` where the backend is tested while bypassing calls to `bs1770gain`.
It was patching `None` to stdout there, I made it patch `bs1770gain 0.0.0, ` instead.

Issue #1273, opened 2015-01-26
Access MusicBrainz series (releases, release groups, recordings)


Freso:
See https://musicbrainz.org/doc/Series

It would be great if beets could pull series information so you could sort e.g. podcasts like https://musicbrainz.org/series/f158ad31-bc44-4b22-94a4-025a8d6375cf into their own folder, using the ordering number in the filename, also releases/release groups that are part of a series could be sorted into some folder structure using the series or something. Either way, it would great to at least have the information handy. ;)

One caveat:
- Entities can be part of multiple series, so some way to select/prioritise between them will be needed.

sampsyo:
Interesting. Shall we make a list of useful information to pull from the series?

* ordering number
* series name?
* series MBID?

johtso:
This would be a great addition! Here's an example of some pain that would be alleviated by support for series :) http://musicbrainz.org/edit/31856400

Nippey:
Hi I just tried to find out what is necessary to achieve this functionality. But I had no success getting series information from the MB API.
Maybe you do already have this information?

I have put some questions here:
http://stackoverflow.com/questions/37991148/how-to-get-information-about-the-series-a-release-belongs-to-via-musicbrainz

If you can help me, I will help you :D

jackwilsdon:
You might be better asking in the MusicBrainz IRC channel (#musicbrainz on freenode.net) :smile:.

Nippey:
Good point! You'll hear from me if I had success

johtso:
Any thoughts how this might work?

I guess one approach is have some configuration where the user can list MBIDs of series they want to have grouped. Each MBID can also optionally have a flag indicating that all the recordings should be in one directory (often each release just contains a single recording, e.g. a podcast and you wouldn't want them to be in separate directories).

In the rare case a release belongs to more than one series that is listed in the config, throw an error?

PR #3482, opened 2020-02-08
Fetch series info from MusicBrainz

GuilhermeHideki:
A very simple way to fetch (only release and release group) series from MB. The command `beet series` needs:
 - the series id
 - beets query

related to https://github.com/beetbox/beets/issues/1273

## Checklist

- [ ] fix code issues
- [ ] add tests
- [ ] add docs

### DIFF
--- a/beetsplug/series.py
+++ b/beetsplug/series.py
@@ -218,7 +218,7 @@ class MbSeriesPlugin(BeetsPlugin):
                 self.albums(lib, query, series_id, move, pretend, write)
             else:
                 series = set()
-                for album in lib.albums('-mb_seriesid::^$'):
+                for album in lib.albums('mb_seriesid::.'):
                     series.add(album.mb_seriesid)
 
                 self._log.info(f'updating {len(series)} series')
--- a/beetsplug/series.py
+++ b/beetsplug/series.py
@@ -1,0 +1,276 @@
+# -*- coding: utf-8 -*-
+# This file is part of beets.
+# Copyright 2019, Guilherme Danno.
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+
+"""Update library's tags using MusicBrainz.
+"""
+from __future__ import division, absolute_import, print_function
+
+from beets.plugins import BeetsPlugin
+from beets import ui, util
+from beets.autotag.mb import _parse_id, musicbrainzngs
+from collections import defaultdict, namedtuple
+
+import re
+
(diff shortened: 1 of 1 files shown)


GuilhermeHideki:
Will check the flake8 warnings

sampsyo:
Nice; this looks cool so far! :+1:

the-solipsist:
It would be great if like "comp" and "singleton", there could be "series" special path.  Further, `$series_name` could be a variable that can be used in the config (containing the "release group series" name) to help with renaming file paths.

sampsyo:
Hi, @the-solipsist! You actually don't need a special path key or any changes to beets to make this work! You can make it happen today, like this:

```
paths:
    'series_name::.': Series/$series_name/$artist/...
```

That's a regex query used as a path key that enables the path for any item where $series_name is nonempty. You can set $series_name (a flexible attribute of your own invention) using `beet modify`.

Freso:
How does this handle when there are multiple series?

GuilhermeHideki:
@Freso The user can only set one series; The new value overwrites the last value

beetsplug/series.py
```diff
@@ -0,0 +1,283 @@
...
+            series = musicbrainzngs.get_series_by_id(series_id, SERIES_RELS)
+            return self._build_series(series)
+        except Exception:
+            return
+
+    def search(self, query):
+        # TODO: Pagination
+        result = musicbrainzngs.search_series(
+            query,
+            limit=self.item_limit)
```
GuilhermeHideki, line 114: @sampsyo do you have an opinion about how we should do pagination? I left this off because sometimes I couldn't find the item I wanted in the first 10 items. I don't know if people use terminals without scrolling

sampsyo: Good question—I have no strong opinion. It seems like a reasonable approach would be to have a limit by default and provide a command-line option for getting a longer list of results (or removing the limit altogether).

beetsplug/series.py
```diff
@@ -0,0 +1,283 @@
...
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+
+"""Update library's tags using MusicBrainz.
```
sampsyo, line 16: Seems like this docstring may be out of date. :smiley:

Kernald:
This is still relevant.

trapd00r:
What's the status of this? I came across this because I'd like to sort 'album series' properly in the filesystem:
```

▕ drwxr xr x ▏<  hour│   4K│Versus Video Games 2 [Daniel Tidwell] (2011)
▕ drwxr xr x ▏<  hour│   4K│Versus Video Games 3 [Daniel Tidwell] (2015)
▕ drwxr xr x ▏<  hour│   4K│Versus Video Games 4 [Daniel Tidwell] (2018)
▕ drwxr xr x ▏<  hour│   4K│Versus Video Games [Daniel Tidwell] (2010)
```

sampsyo:
Hi! I think it just needs a "champion," i.e., someone to push it forward and make sure it gets updated and merged?

Issue #3490, opened 2020-02-16
A new plugin was born: beets-bpmanalyser


adamjakab:
### News
Hi! I am an obsessive music geek. I realised this only since I installed Beets. I greatly enjoy this application and it helped me a great deal to sort out my music library. I played around the application and most of the plugins - actually the last two days I didn't do much else ;)

Anyways, I thought that with the tempo(bpm) information on the songs stored on acousticbrainz I would have all songs with tempo information once I imported my library. Only later I realised that this was not the case and I still have a good 30% of my music without tempo information. Since I am also a runner and I choose the music I take with me based on the type of training I am going out to do, this info for me is crucial (...yes,  I know, I am obsessive ... ;)). 

So I set out to make two plugins:
- a tempo(bpm) analyser for media files (beets-bpmanalyser)
- a file song exporter made for runners (beets-goingrunning)

While the second one may be interesting only for a limited number of people, the first one `beets-bpmanalyser` might be interesting for many people also for other reasons.

The good news is that I already created this plugin (`beets-bpmanalyser`) and it is available on [PyPi](https://pypi.org/project/beets-bpmanalyser/) and [Github](https://github.com/adamjakab/BeetsPluginBpmAnalyser).

### Question
I would like this plugin to be listed on the Plugins page under the [Other Plugins](https://beets.readthedocs.io/en/stable/plugins/index.html#other-plugins) section. I also thought about adding it as an internal plugin, but it has two quite heavy dependencies: [numpy](https://pypi.org/project/numpy/) and [aubio](https://pypi.org/project/aubio/) and maybe keeping it optional is a better choice. So, can you please add it? Thanks

PS: I realise this is not the best place for this - it is more like a forum post - but I don't know where else to post it.

sampsyo:
Awesome! I'm glad everything's going well for you so far!

This looks pretty rad, and we'd be happy to add a link to that "Other Plugins" section. Please open a pull request and we'll merge it! You can find the source under `docs`.

Just for fun, I'm going to link to the included bpm plugin, which is obviously very different but worth mentioning in this context: https://beets.readthedocs.io/en/stable/plugins/bpm.html

Also, while discussion issues like this are fine, we do also have a forum, if you're interested: https://discourse.beets.io

adamjakab:
Super! Thanks for the forum link.

sampsyo:
Fixed in #3491.

PR #3491, opened 2020-02-17
Added beets-bpmanalyser to the list of 'other plugins'

adamjakab:
I could not find a logical explanation to the order of the plugins so I assumed that they are in a chronological order and therefore I put mine to the bottom. (Maybe alphabetical would be better. What do you think?)


adamjakab:
PS: This fixes #3490 - discussion about listing the new plugin.

sampsyo:
Awesome; thanks!! Indeed—there's not really any significant order to that listing. :smiley:

PR #3492, opened 2020-02-17
parentwork: Only call store when the metadata has changed.

aereaux:
Otherwise, this rewrites all your files every time.


sampsyo:
Seems great!! Could you please add a quick changelog entry describing the change and we'll merge this?

(The AppVeyor failure above appears to be spurious.)

aereaux:
Sure, done.

sampsyo:
Perfect; thanks!

PR #3493, opened 2020-02-18
Allow the quality to be set for embedded/fetched cover art

danbee:
ImageMagick seems to default to 100% quality when resizing artwork which results in much larger files than necessary. This PR adds the ability to specify the quality level to pass to ImageMagick, which means file sizes can be much more reasonable with minimal visual quality loss.

I so far haven't been able to get the tests running here so I'm not sure what, if anything, I need to change there.

I've been using this branch for the last few months though and so far it's working well.

### DIFF
--- a/beets/util/artresizer.py
+++ b/beets/util/artresizer.py
@@ -59,7 +59,7 @@ def temp_file_for(path):
         return util.bytestring_path(f.name)
 
 
-def pil_resize(maxwidth, path_in, path_out=None):
+def pil_resize(maxwidth, quality, path_in, path_out=None):
     """Resize using Python Imaging Library (PIL).  Return the output path
     of resized image.
     """
@@ -72,7 +72,7 @@ def pil_resize(maxwidth, path_in, path_out=None):
         im = Image.open(util.syspath(path_in))
         size = maxwidth, maxwidth
         im.thumbnail(size, Image.ANTIALIAS)
-        im.save(util.py3_path(path_out))
+        im.save(util.py3_path(path_out), quality=quality)
         return path_out
     except IOError:
         log.error(u"PIL cannot create thumbnail for '{0}'",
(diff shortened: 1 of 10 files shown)


beets/util/artresizer.py
```diff
@@ -190,14 +191,14 @@ def __init__(self):
                 self.im_convert_cmd = ['magick']
                 self.im_identify_cmd = ['magick', 'identify']
 
-    def resize(self, maxwidth, path_in, path_out=None):
+    def resize(self, maxwidth, quality, path_in, path_out=None):
```
wisp3rwind: I think it is not a good idea to change the function signature here. Rather, `quality` should be appended as a keyword argument with a reasonable default value, as you did in `art.py`. I'm quite certain that otherwise, we will break third-party plugins without a good reason.

danbee:
@wisp3rwind Thanks for your comments. I did a bit more research and I was wrong about the default behaviour of ImageMagick:

> "For the JPEG and MPEG image formats, quality is 1 (lowest image quality and highest compression) to 100 (best quality but least effective compression). The default is to use the estimated quality of your input image if it can be determined, otherwise 92."

Therefore, to correctly get back the original behaviour we need to leave the quality setting off entirely if it's not set. I've made some changes which should achieve this.

danbee:
@sampsyo I've added documentation for the quality setting. I added some detail about differing default behaviour for different resizing backends. Let me know if you think that's not required.

I was also able to figure out that `images.webserv.nl` can take a quality parameter (`q`) so I've added code to pass that through as well. This means that all the image resizing backends now support the quality setting.

I think I updated all the relevant docstrings but please let me know if there's any you think I missed!

sampsyo:
Looks awesome; thank you for pursuing this!

One last small request: can we clarify in the documentation a bit about what the valid quality settings are and what they mean? Namely, I presume that it's the JPEG "quality number" value, which can range from 1 to 100 (worst to best). Is that right?

danbee:
@sampsyo I've added more information about the valid settings for quality. I also added the feature to the changelog. Let me know if that's correct.

sampsyo:
Looks great; thanks!! Merged with a few wording tweaks.

PR #3494, opened 2020-02-20
Added beets-goingrunning to the list of 'other plugins'

adamjakab:
Whilst I am at it, shall I order the plugins in the 'other plugins' section alphabetically?


sampsyo:
Awesome; thanks!! I'll just merge this now, but I don't object to another PR to do the alphabetization if you are so inclined. :smiley:

PR #3498, opened 2020-02-23
Raising error on missing configuration file inclusion

adamjakab:
I think it is better to know than ignoring it silently. I don't know how you feel about this but I am requesting this PR because I lost a good deal of time convinced that my configuration was read only to discover that I mistyped the path. We could also opt for a 'softer' warning...

### DIFF
--- a/beets/__init__.py
+++ b/beets/__init__.py
@@ -35,6 +35,8 @@ class IncludeLazyConfig(confuse.LazyConfig):
                 filename = view.as_filename()
                 if os.path.isfile(filename):
                     self.set_file(filename)
+                else:
+                    raise FileNotFoundError("Warning! Configuration file({0}) does not exist!".format(filename))
         except confuse.NotFoundError:
             pass
 
--- a/beets/__init__.py
+++ b/beets/__init__.py
@@ -36,7 +36,8 @@ class IncludeLazyConfig(confuse.LazyConfig):
                 if os.path.isfile(filename):
                     self.set_file(filename)
                 else:
-                    raise FileNotFoundError("Warning! Configuration file({0}) does not exist!".format(filename))
+                    raise FileNotFoundError(
+                        "Warning! Configuration file({0}) does not exist!".format(filename))
         except confuse.NotFoundError:
             pass
 
--- a/beets/__init__.py
+++ b/beets/__init__.py
@@ -15,8 +15,6 @@
 
 from __future__ import division, absolute_import, print_function
 
-import os
-
 import confuse
 
 __version__ = u'1.5.0'
@@ -32,8 +30,7 @@ class IncludeLazyConfig(confuse.LazyConfig):
 
         try:
             for view in self['include']:
-                filename = view.as_filename()
-                self.set_file(filename)
+                self.set_file(view.as_filename())
         except confuse.NotFoundError:
             pass
         except confuse.ConfigReadError as err:
--- a/beets/__init__.py
+++ b/beets/__init__.py
@@ -16,6 +16,7 @@
 from __future__ import division, absolute_import, print_function
 
 import confuse
+from sys import stderr
 
 __version__ = u'1.5.0'
 __author__ = u'Adrian Sampson <adrian@radbox.org>'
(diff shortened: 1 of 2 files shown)


sampsyo:
It seems fine to add a warning along these lines! I think I'd lean toward just a warning, however, rather than exiting with a traceback. Also, note that we already handle `NotFoundError`—maybe we should just print the message there instead of adding a new handler?

adamjakab:
Ok for the warning (no traceback) part. However, I think the decision to make here is if a missing configuration file constitutes to a bad-enough situation for which we want to interrupt the execution or not. In my opinion it does because the application would behave differently from how I would expect it to do with all the configuration files loaded.

sampsyo:
I guess I feel pretty agnostic about that precise distinction? But one argument for leaving it as a warning is to be less disruptive to people who were relying on the old behavior, who will now just get a warning when they run beets instead of being prevented from running it altogether… you could have an `import` in there that just points to a placeholder location that is only sometimes filled with an actual config file, for example.

adamjakab:
Ok. I see your point. Let's go with a simple warning then.

adamjakab:
This, with the following configuration:

```yml
include: 
    - inexistent.yml
```
produces:
```bash
$ ./beet
Warning! Missing configuration file! [Errno 2] No such file or directory: '/Users/jackisback/.config/beets/DEVEL/inexistent.yml'
Usage: 
  beet COMMAND [ARGS...]
  beet help COMMAND
...
````
and beets keeps running happily ;)

However, it does not catch badly formatted configuration files. With:
```yml
include: 
    - badformat.yml
```
+ the 'badformat.yml' file:
```yml
this
    will
        not
            pass
        as
    ok
yml
```
I expected the same warning to show because the `confuse.load_yaml()` function throws a ConfigReadError. It does not. This is what I get:
```bash
$ ./beet
configuration error: musicbrainz.host not found
```
and beets exits. But maybe this needs a separate PR.

sampsyo:
Great; thanks for continuing to sort this out!

A few random things:

- I wish we could use `logging` instead of `print`, but I think that would necessitate an impossible circular import. :cry:
- Barring that, let's at least print to stderr instead of stdout.
- I recommend this somewhat terser error message: "configuration \`import\` failed: "

I am also a tad confused by the issue with the malformed YAML… I would have expected that to throw an error too. Maybe we should dig into why that's not happening!

adamjakab:
Sure, no worries.

As for the malformed yaml issue I will open a separate issue (also referencing this PR) and we can dig into it. It seems odd.

beets/__init__.py
```diff
@@ -32,11 +31,13 @@ def read(self, user=True, defaults=True):
...
         try:
             for view in self['include']:
-                filename = view.as_filename()
-                if os.path.isfile(filename):
-                    self.set_file(filename)
+                self.set_file(view.as_filename())
         except confuse.NotFoundError:
             pass
+        except confuse.ConfigReadError as err:
+            stderr.write("Configuration 'import' failed: {}"
```
sampsyo:
```suggestion
            stderr.write("configuration `import` failed: {}"
```

beets/__init__.py
```diff
@@ -32,11 +31,13 @@ def read(self, user=True, defaults=True):
...
+                         .format(err.reason))
+            pass
```
sampsyo: This line is no longer necessary.

adamjakab: oooops

adamjakab:
> Could you please add a quick changelog entry?

Hi @sampsyo. This one: [docs/changelog.rst](https://github.com/beetbox/beets/blob/master/docs/changelog.rst)? Can you please give me an advice on what editor to use to edit rst files? I have not used this format before and I don't want to mess it up. tnx

beets/__init__.py
adamjakab: ok, coming with next commit

sampsyo:
Looks like you got it; thanks!!

I don't have specific text editor recommendations for editing ReST files (I just use vim :smiley:), but you might consider previewing changes to the docs. You can generate the HTML files by typing `make html` in the `docs` directory, which likely will require you to install [Sphinx](https://www.sphinx-doc.org/en/master/).

Issue #941, opened 2014-09-11
UI: confirm Chroma is operational


gwern:
The documentation for the [chroma fingerprinting plugin](http://beets.readthedocs.org/en/v1.3.7/plugins/chroma.html) specifically warns the user that it's tricky to set up and may not work correctly:

> First, it can be trickier to set up than beets itself (you need to set up the native fingerprinting library, whereas all of the beets core is written in pure Python).

The complexity of the dependencies section which follows suggests this as well, and [issue 885](https://github.com/sampsyo/beets/issues/885) warns me that the solution to it may also break chroma (or so I infer from the solution being to remove one version of the library when somone writes "We've seen some hangs when using both versions simultaneously, as can happen when the chroma plugin is also enabled.")

But while importing, there is no indication of whether the fingerprinting is working or if one is spending many hours doing tagging while crippled! A simple startup message like 'checking fingerprinting is working... fingerprinting seems to be working' would suffice.

(I've discovered that my chroma fingerprinting _does_ seem to be working when I turn on debugging output like `beet -v import` but this is (1) not something users will think to do, and (2) makes the import interface much harder to use.)

smichel17:
New beets user here.

I have a couple albums that have really bad metadata (eg, missing track names, just "Audiotrack 01", potentially multiple albums in the same directory, etc). I decided to use Acoustid to try to find metadata for these songs.

I enabled chroma and used `beet fingerprint` to (successfully) tag my library. Then I re-ran `beets import` to update the metadata now that everything is tagged... and I have no idea if it is working or not. It doesn't seem like I'm getting recommendations that are any different than before, but I have no idea if that's because I should be using a different set of option for import, or whether the fingerprints are not being used to find metadata.

I hope you'll reconsider how important this feature is; I consider it to be critical, as the chroma plugin is currently useless to me.

sampsyo:
@smichel17, for your particular predicament, have you tried turning on verbose mode to confirm things are happening?

And just to be clear, we (the usual developers) do think it's important! We just can't do everything at once. If you're particularly interested in a feature, please consider pitching in.

smichel17:
>@smichel17, for your particular predicament, have you tried turning on verbose mode to confirm things are happening?

I hadn't; I'll try that and open an issue elsewhere if I can't figure it out from there, so I don't clog up this issue.

>And just to be clear, we (the usual developers) do think it's important! We just can't do everything at once. If you're particularly interested in a feature, please consider pitching in.

[...]

PR #3504, opened 2020-03-01
Document running chroma plugin in verbose mode

smichel17:
For #941


sampsyo:
Thanks!! Merged with slightly different wording.

PR #3514, opened 2020-03-14
Fix error when art_filename contain subdiretories 

DurvalMenezes:
This is a fix for the error that will happen when you have, in config.yam, something like this: 
    art_filename: art/front

(note the "scan" subdirectory -- I do it this way because I segregate all art files in a subdir). 

Without this fix, the `beet import` command fails with:  

    Error: No such file or directory while copying /tmp/tmp5aczqcob.jpg to MY_DIR/ARTIST/ALBUM/art/front.jpg

as generally (at least the way I'm using beets) the subdir 'art' would not exist . 

With these modifications `beet import` will then create the missing subdir as needed, ditto `beet move` when moving. 

Please note that I'm no Python nor beets expert, so other/better fixes for this may exist. Please feel free to change or discard this PR if it is somehow innappropriate.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -167,8 +167,7 @@ Fixes:
 * Added a warning when configuration files defined in the `include` directive
   of the configuration file fail to be imported.
   :bug:`3498`
-* Automatically creates destination directories as needed when copying/moving files,   
-  so including directories as part of eg art_filename no longer lead to errors.
+* Enhanced code legibility/consistency in `beets/__init__.py`
   Thanks to :user:`DurvalMenezes`.
   :bug:`3514`
 * Added the normalize method to the dbcore.types.INTEGER class which now
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -169,7 +169,8 @@ Fixes:
   :bug:`3498`
 * Automatically creates destination directories as needed when copying/moving files,   
   so including directories as part of eg art_filename no longer lead to errors.
-  Thanks to :user:`DurvalMenezes`
+  Thanks to :user:`DurvalMenezes`.
+  :bug:`3514`
 * Added the normalize method to the dbcore.types.INTEGER class which now
   properly returns integer values, which should avoid problems where fields
   like ``bpm`` would sometimes store non-integer values.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -167,7 +167,7 @@ Fixes:
 * Added a warning when configuration files defined in the `include` directive
   of the configuration file fail to be imported.
   :bug:`3498`
-* Enhanced code legibility/consistency in `beets/__init__.py`
+* Enhanced code legibility/consistency in `util/__init__.py`
   Thanks to :user:`DurvalMenezes`.
   :bug:`3514`
 * Added the normalize method to the dbcore.types.INTEGER class which now
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -167,6 +167,9 @@ Fixes:
 * Added a warning when configuration files defined in the `include` directive
   of the configuration file fail to be imported.
   :bug:`3498`
+* Automatically creates destination directories as needed when copying/moving files,   
+  so including directories as part of eg art_filename no longer lead to errors.
+  Thanks to :user:`DurvalMenezes`
 
 For plugin developers:
 
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -457,14 +457,14 @@ def copy(path, dest, replace=False):
     """
     if samefile(path, dest):
         return
-    path = syspath(path)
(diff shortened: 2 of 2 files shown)


sampsyo:
Cool; this seems like it might be useful to have! Can I ask you to make a few small changes?

- Avoid the use of `exist_ok`, which does not exist in Python 2 (which we need to drop soon anyway but haven't yet).
- Probably remove the `mode` argument? I don't quite see why this is necessary, and making the directory world-writable by default is likely to be confusing.
- Consider using our `util.mkdirall` utility instead of `os.makedirs`. This has its own error handling, so it probably needs to come out of the exception handler.
- Add a quick changelog entry describing the fix.

DurvalMenezes:
Thanks for pointing me towards `util.mkdirall`, I wasn't aware of it. Will test it here and report back later; From a cursory examination of the code, it seems it will do what needs to be done better than  `os.makedirs`.

DurvalMenezes:
> Add a quick changelog entry describing the fix.
@sampsyo By this, you mean editing then pushing a commit for  beets/docs/changelog.rst , correct? Please confirm.

sampsyo:
Yes; exactly! Thanks!!

A couple of additional things:
- The calls to `mkdirall` should go outside of the `try` block, I think, because they already handle `OSError`s internally.
- I believe `mkdirall` needs the version of the path that has *not* gone through `syspath`. This is a really gnarly restriction, but it's important for getting paths right on Windows.

DurvalMenezes:
Hello @sampsyo ,

> * The calls to `mkdirall` should go outside of the `try` block, I think, because they already handle `OSError`s internally.

Of course, thanks for calling this to my attention. 
 
> * I believe `mkdirall` needs the version of the path that has _not_ gone through `syspath`. This is a really gnarly restriction, but it's important for getting paths right on Windows.

Agreed. I hadn't noticed that `mkdirall` calls syspath internally, which would result in the path being `syspath`'ed twice (not good). 

I will implement both changes in my next commit.

docs/changelog.rst
```diff
@@ -167,6 +167,9 @@ Fixes:
 * Added a warning when configuration files defined in the `include` directive
   of the configuration file fail to be imported.
   :bug:`3498`
+* Automatically creates destination directories as needed when copying/moving files,   
+  so including directories as part of eg art_filename no longer lead to errors.
+  Thanks to :user:`DurvalMenezes`
```
```suggestion
  Thanks to :user:`DurvalMenezes`.
  :bug:`3514`
```

DurvalMenezes:
> Looks good overall! Although I think the methods would read _slightly_ more cleanly if we didn't use `original_dest` and instead just used `syspath(dest)` where that's necessary? Could you give that a shot and see if that makes things a little more legible?

@sampsyo, thanks for the suggestion. My initial intention was being to change the code as least as possible, therefore `original_dest`, but you are right, it was ugly. My latest edit tries to remedy that, and besides doing away with `original_dest`, I do a similar thing with `path`, and also some more not-directly-related edits which I think makes the code more legible and consistent (more details in the commit comment). 

Please let me know what you think.

DurvalMenezes:
> Looks good overall! Although I think the methods would read _slightly_ more cleanly if we didn't use `original_dest` and instead just used `syspath(dest)` where that's necessary? Could you give that a shot and see if that makes things a little more legible?

@sampsyo, thanks for the suggestion. My initial intention was to try and change the code as least as possible (therefore `original_dest`) but you are right, it was ugly. My latest edit tries to remedy that, and besides doing away with `original_dest`, I do a parallel thing with `path`, and also some more not-directly-related edits which I think makes the code more legible and consistent (more details in the commit comment). 

Please let me know what you think.

sampsyo:
That looks great; thank you for rearranging this!

I have to apologize, however, that I have totally missed something important in this whole thread: that these changes are taking place in `util` as opposed to in `library`. I'm not sure how I didn't notice that before! But I think we need to take a somewhat different direction: changing the `move` and `copy` utilities will cause pervasive behavior changes across all of beets, and I don't think that's exactly what we want. It would probably be better to make a _targeted_ change to the album art behavior itself—it's likely that, in some other part of beets, we intentionally want to get an error when trying to copy/move to a directory that doesn't exist.

Could I please ask you to try out making the change in `Album.move_art` instead? That might even be a little simpler than changing the utility functions:
https://github.com/beetbox/beets/blob/b34d1f71a93b3663f70f46a4d5ed08702b14c5ca/beets/library.py#L1055

Sorry again for leading us on this wild goose chase while I didn't read the diffs carefully enough!

DurvalMenezes:
No problem, @sampsyo. What about I trim down this PR to just the legibility changes I mentioned earlier, and then open a new PR for the modification to `library.move_art`?

sampsyo:
That would be awesome! Thanks again.

PR #3517, opened 2020-03-15
Fix #3516 - Erroneous @classmethod on NoneQuery.match

### LINKED ISSUE #3516: Erroneous @classmethod on NoneQuery.match
### Problem

When running a query against the library that contains a NoneQuery such as:

> dbcore.query.NoneQuery(u'gender', fast=False)

beet exits with the following traceback:

```sh
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/bin/beet", line 8, in <module>
    sys.exit(main())
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/lib/python3.7/site-packages/beets/ui/__init__.py", line 1266, in main
    _raw_main(args)
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/lib/python3.7/site-packages/beets/ui/__init__.py", line 1253, in _raw_main
  File "/Users/jackisback/Documents/Projects/Python/BeetsPluginXtractor/beetsplug/xtractor/command.py", line 133, in func
    self.xtract()
  File "/Users/jackisback/Documents/Projects/Python/BeetsPluginXtractor/beetsplug/xtractor/command.py", line 157, in xtract
    if len(library_items) == 0:
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/lib/python3.7/site-packages/beets/dbcore/db.py", line 741, in __len__
    for obj in self:
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/lib/python3.7/site-packages/beets/dbcore/db.py", line 689, in _get_objects
    if not self.query or self.query.match(obj):
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/lib/python3.7/site-packages/beets/dbcore/query.py", line 462, in match
    return all([q.match(item) for q in self.subqueries])
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/lib/python3.7/site-packages/beets/dbcore/query.py", line 462, in <listcomp>
    return all([q.match(item) for q in self.subqueries])
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/lib/python3.7/site-packages/beets/dbcore/query.py", line 472, in match
    return any([q.match(item) for q in self.subqueries])
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/lib/python3.7/site-packages/beets/dbcore/query.py", line 472, in <listcomp>
    return any([q.match(item) for q in self.subqueries])
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/lib/python3.7/site-packages/beets/dbcore/query.py", line 168, in match
    return item[cls.field] is None
AttributeError: type object 'NoneQuery' has no attribute 'field'
```

The above example is using a flex attribute but the same error can be reproduced using a fixed field such as: `dbcore.query.NoneQuery(u'title')`

This behaviour has already been reported in #1634, identifying the problem within the `@classmethod` decorated match method using `self` in the NoneQuery class. The bug was fixed by correcting `self` to `cls` leaving the method itself decorated with `@classmethod`.

### Setup
beets version 1.4.9
Python version 3.7.6
plugins: info, xtractor

### Possible solution
The NotQuery.match method should probably not be decorated with `@classmethod`.

adamjakab:
It seems like tests are not hitting this method...

# Checklist
- [x] Fix
- [x] Test
- [x] Changelog

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -152,12 +152,16 @@ class NoneQuery(FieldQuery):
 
     def __init__(self, field, fast=True):
         super(NoneQuery, self).__init__(field, None, fast)
+        self.pattern = None
 
     def col_clause(self):
         return self.field + " IS NULL", ()
 
     def match(self, item):
-        return item.get(self.field, default=None) is None
+        return self.value_match(self.pattern, item.get(self.field, default=None))
+
+    def value_match(self, pattern, value):
+        return value is pattern
 
     def __repr__(self):
         return "{0.__class__.__name__}({0.field!r}, {0.fast})".format(self)
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -162,6 +162,10 @@ class NoneQuery(FieldQuery):
         except KeyError:
             return True
 
+    @classmethod
+    def value_match(cls, pattern, value):
+        return pattern == value
+
     def __repr__(self):
         return "{0.__class__.__name__}({0.field!r}, {0.fast})".format(self)
 
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -158,7 +158,8 @@ class NoneQuery(FieldQuery):
         return self.field + " IS NULL", ()
 
     def match(self, item):
-        return self.value_match(self.pattern, item.get(self.field, default=None))
+        return self.value_match(self.pattern,
+                                item.get(self.field, default=None))
 
     def value_match(self, pattern, value):
         return value is pattern
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -152,17 +152,12 @@ class NoneQuery(FieldQuery):
 
     def __init__(self, field, fast=True):
         super(NoneQuery, self).__init__(field, None, fast)
-        self.pattern = None
 
     def col_clause(self):
         return self.field + " IS NULL", ()
 
     def match(self, item):
-        return self.value_match(self.pattern,
-                                item.get(self.field, default=None))
-
-    def value_match(self, pattern, value):
-        return value is pattern
+        return item.get(self.field, default=None) is None
 
     def __repr__(self):
         return "{0.__class__.__name__}({0.field!r}, {0.fast})".format(self)
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -156,10 +156,9 @@ class NoneQuery(FieldQuery):
     def col_clause(self):
         return self.field + " IS NULL", ()
 
-    @classmethod
-    def match(cls, item):
+    def match(self, item):
         try:
-            return item[cls.field] is None
+            return item.get(self.field) is None
         except KeyError:
             return True
 
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -157,10 +157,7 @@ class NoneQuery(FieldQuery):
         return self.field + " IS NULL", ()
 
     def match(self, item):
-        try:
-            return item.get(self.field) is None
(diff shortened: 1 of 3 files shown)


adamjakab:
The NoneQuery class is now 100% covered. I added a missing value_match abstract method which later I removed because I was unable to reach/test it. I am not sure how it works but I think in the case of the NoneQuery this method can not be reached. Or am I wrong?

sampsyo:
Great; thanks!!

The `value_match` method is indeed unreachable; I think we should leave it off here. It is only used by the base implementation of `match` for cases where the behavior is best defined in terms of the pattern and value. Although the docstring says "Subclasses must provide a `value_match` class method," it should probably instead say that they should *either* provide `value_match` or just override `match` itself.

Looks like there's also a test failure related to the Python 3 behavior of `repr` on strings:
https://travis-ci.org/github/beetbox/beets/jobs/662725636#L912

Would you mind also adding a quick changelog entry describing the fix?

Thanks again for addressing this!

test/test_query.py
```diff
@@ -772,6 +772,38 @@ def test_match_after_set_none(self):
...
+            self.assertEquals("NoneQuery(u'{}', False)".format(fld),
+                              str(NoneQuery(fld, fast=False)))
+        else:
+            self.assertEquals("NoneQuery('{}', True)".format(fld),
+                              str(NoneQuery(fld)))
+            self.assertEquals("NoneQuery('{}', True)".format(fld),
+                              str(NoneQuery(fld, fast=True)))
+            self.assertEquals("NoneQuery('{}', False)".format(fld),
+                              str(NoneQuery(fld, fast=False)))
+
```
adamjakab: I am too n00b to Python to understand what is the story with unicode  there. This is not really part of the issue but I though with this we have all methods of the class covered. Can you pls have a look?

beets/dbcore/query.py
```diff
@@ -152,16 +152,17 @@ class NoneQuery(FieldQuery):
...
-        try:
-            return item[cls.field] is None
-        except KeyError:
-            return True
+    def match(self, item):
+        return self.value_match(self.pattern,
+                                item.get(self.field, default=None))
+
+    def value_match(self, pattern, value):
+        return value is pattern
```
adamjakab: I think this is ok. Also becuause the call from the match method also uses the `self.patern` initiated in the __init__.
Again, I am to new to Python to tell. Something like this in Php, not implementing an abstract class would fail at compile time... but then again this is not php. So your call: should I remove it?

docs/changelog.rst
```diff
@@ -171,6 +171,8 @@ Fixes:
   properly returns integer values, which should avoid problems where fields
   like ``bpm`` would sometimes store non-integer values.
   :bug:`762` :bug:`3507` :bug:`3508`
+* Removed `@classmethod`` decorator from dbcore.query.NoneQuery.match method
+  failing with AttributeError when called. It is now an instance method.`
```
```suggestion
* Removed ``@classmethod`` decorator from dbcore.query.NoneQuery.match method
  failing with AttributeError when called. It is now an instance method.
  :bug:`3516` :bug:`3517`
```

test/test_query.py
sampsyo: TBH I think we can just remove the tests for `repr`… the problem is that Python 2 string literals used syntax like `u'foo'` while on Python 3 they are just `'foo'`. Fortunately, `repr` formatting is so standard that IMO testing it is sort of overkill.

adamjakab:
@sampsyo - she is ready to go ;)

sampsyo:
Looks excellent; thanks!! :sparkles:

Issue #3344, opened 2019-08-07
lyrics: add tekstowo.pl as a source


octos:
### Use case

It would be nice if beets could fetch lyrics from https://www.tekstowo.pl/ – especially for more obscure Polish releases.

### Solution

https://github.com/hryniuk/tekstowo is pretty simple and works perfectly. We could integrate it into the Lyrics plugin.

sampsyo:
Sounds like a reasonable request! And it looks like the screen-scraping required is pretty light.

wisp3rwind:
Closed by https://github.com/beetbox/beets/pull/3519

PR #3519, opened 2020-03-19
Implement tekstowo lyrics provider

xhocquet:


xhocquet:
Howdy @sampsyo! I found a feature request for a lyrics provider from Poland which seemed appropriate since I worked around lyrics before! No tests yet and I'm not super familiar with testing in Python, so I'll need to work through that a bit. Besides that I've been testing using [this](https://www.tekstowo.pl/piosenka,boy_in_space,u_n_eye.html) popular track!

xhocquet:
Oh, and the request: https://github.com/beetbox/beets/issues/3344

beetsplug/lyrics.py
```diff
@@ -693,6 +738,13 @@ def __init__(self):
             )
             sources.remove('genius')
 
+        if 'tekstowo' in sources and not HAS_BEAUTIFUL_SOUP:
+            self._log.debug(
+                u'The Tekstowo.pl backend requires BeautifulSoup, which is not '
+                u'installed, so the source is disabled.'
+            )
+            sources.remove('tekstowo')
```
sampsyo: Since we now have three different sources that all need the same check and message, maybe it's time to factor this out into a function so we don't need to copy n' paste the error message?

xhocquet:
@sampsyo Updated! Please take a look 🙏

docs/plugins/lyrics.rst
```diff
@@ -166,6 +166,16 @@ library. Install it by typing::
 
 The backend is enabled by default.
 
+Activate Tekstowo.pl Lyrics
```
sampsyo: Maybe just merge this with the previous section, because it gives the same advice as for the Genius backend?

docs/plugins/lyrics.rst
```diff
@@ -166,6 +166,16 @@ library. Install it by typing::
...
+----------------------
```
sampsyo: The line has to be as long as the title above it:
https://travis-ci.org/github/beetbox/beets/jobs/664687146#L1104

beetsplug/lyrics.py
```diff
@@ -627,12 +677,14 @@ def fetch(self, artist, title):
 
 
 class LyricsPlugin(plugins.BeetsPlugin):
-    SOURCES = ['google', 'lyricwiki', 'musixmatch', 'genius']
+    SOURCES = ['google', 'lyricwiki', 'musixmatch', 'genius', 'tekstowo']
+    SOURCES_USING_BEAUTIFUL_SOUP = ['google', 'genius', 'tekstowo']
```
sampsyo: Let's just call this `BS_SOURCES` for brevity.

beetsplug/lyrics.py
```diff
@@ -705,6 +748,17 @@ def __init__(self):
         self.backends = [self.SOURCE_BACKENDS[source](self.config, self._log)
                          for source in sources]
 
+    def sanitize_beautiful_soup_sources(self, sources):
```
sampsyo: As above, I think `sanitize_bs_sources` will do fine.

xhocquet:
@sampsyo Checks are green!

beetsplug/lyrics.py
```diff
@@ -705,6 +748,17 @@ def __init__(self):
         self.backends = [self.SOURCE_BACKENDS[source](self.config, self._log)
                          for source in sources]
 
+    def sanitize_bs_sources(self, sources):
+        for source in self.BS_SOURCES:
+            if source in sources:
+                self._log.warning(u'To use the %s lyrics source, you must '
```
sampsyo: This should be a `debug` rather than a `warning` call. This is important so that there aren't spurious warnings when the user first enables the plugin—they should get the default set of non-BS sources working without any messages.

It was a warning for the Google backend because it needs to be enabled explicitly. It seems… OK, I suppose?… to leave it as a debug message for simplicity…

docs/plugins/lyrics.rst
```diff
@@ -59,11 +60,10 @@ configuration file. The available options are:
   sources known to be scrapeable.
 - **sources**: List of sources to search for lyrics. An asterisk ``*`` expands
   to all available sources.
-  Default: ``google lyricwiki musixmatch genius``, i.e., all the
+  Default: ``google lyricwiki musixmatch genius tekstowo``, i.e., all the
   available sources. The ``google`` source will be automatically
   deactivated if no ``google_API_key`` is setup.
-  Both it and the ``genius`` source will only be enabled if BeautifulSoup is
-  installed.
+  The following sources will only be enabled if BeatifulSoup is installed: ``[google, genius, tekstowo]``
```
sampsyo: Here's a slightly more human way of writing it:

```rest
The ``google``, ``genius``, and ``tekstowo`` sources will only be enabled if BeautifulSoup is installed.
```

docs/plugins/lyrics.rst
```diff
@@ -129,11 +129,6 @@ few suggestions.
 Activate Google Custom Search
 ------------------------------
 
-Using the Google backend requires `BeautifulSoup`_, which you can install
-using `pip`_ by typing::
-
-    pip install beautifulsoup4
-
```
sampsyo: I think I would actually like to leave this copy of the text here? Because that way the section for the Google backend can stand alone. (Otherwise, it might be easy to miss that there's another section.)

docs/plugins/lyrics.rst
```diff
@@ -156,15 +151,15 @@ After that, the lyrics plugin will fall back on other declared data sources.
 .. _pip: https://pip.pypa.io
 .. _BeautifulSoup: https://www.crummy.com/software/BeautifulSoup/bs4/doc/
 
-Activate Genius Lyrics
-----------------------
+Installing BeautifulSoup for Google, Genius, or Tekstowo.pl Backends
```
sampsyo: With the above in mind:
```suggestion
Active Genius and Tekstowo.pl Lyrics
```

xhocquet:
I plan on wrapping this up! Life got in the way 🙂

xhocquet:
Still plan on revisiting this

sampsyo (approved):
Awesome; this looks great!! Thanks for coming back to it.

PR #3529, opened 2020-03-26
Fix #3520 - Comparing type instances

### LINKED ISSUE #3520: Error in Type comparison of Flexible Fields declared by plugins
### Problem

When activating two plugins which [declare the type for a flexible field](https://beets.readthedocs.io/en/stable/dev/plugins.html#flexible-field-types), beets exits with the following exception:

```bash
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/bin/beet", line 8, in <module>
    sys.exit(main())
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/lib/python3.7/site-packages/beets/ui/__init__.py", line 1266, in main
    _raw_main(args)
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/lib/python3.7/site-packages/beets/ui/__init__.py", line 1249, in _raw_main
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/lib/python3.7/site-packages/beets/ui/__init__.py", line 1148, in _setup
    library.Item._types.update(plugins.types(library.Item))
  File "/Users/jackisback/opt/miniconda3/envs/beetsdevel/lib/python3.7/site-packages/beets/plugins.py", line 344, in types
    u'another type.'.format(plugin.name, field)
beets.plugins.PluginConflictException: Plugin acousticbrainz defines flexible field average_loudness which has already been defined with another type.
```

The above error can be reproduced by activating any two plugins (one example is`acousticbrainz`) which declare flexible fields like the code snippet below:

```python
item_types = {
        'average_loudness': types.Float(6)
}
```

The expected behaviour is that two plugins which declare the same field with the same type (for example: `types.Float()`) should be able to co-exist.

The actual behaviour is that beets exits with:

> beets.plugins.PluginConflictException: Plugin acousticbrainz defines flexible field average_loudness which has already been defined with another type.

The type comparison here seems to be wrong:

https://github.com/beetbox/beets/blob/b34d1f71a93b3663f70f46a4d5ed08702b14c5ca/beets/plugins.py#L342

because it compares the actual instances instead of their types.

adamjakab:


adamjakab:
@sampsyo - failing on python 2.7 - I don't know what is the issue. Can you advise?

beets/dbcore/types.py
```diff
@@ -117,6 +117,14 @@ def to_sql(self, model_value):
         """
         return model_value
 
+    def __eq__(self, other):
+        """Allow to compare type instances by their full namespaced name
+        """
+        def get_ns(o):
+            return "{m}.{n}".format(m=o.__class__.__module__,
+                                    n=o.__class__.__name__)
+        return get_ns(self) == get_ns(other)
```
sampsyo: This implementation of equity will make any two objects of the same type compare equal. That's not quite what we want—while we would like any object constructed with `Integer()` to be equal to any other, we would not want `PaddedInt(4)` to be equal to `PaddedInt(2)`. So unfortunately, I think there is no way around providing per-class equality functions.

For parameter-free classes like `Integer`, the implementation could do this to avoid stringifying things:

```py
def __eq__(self, other):
    if not isinstance(other, type(self)):
        return NotImplemented
    return True
```

adamjakab: I see your point. In fact, I also thought about the classes which are initialised with a length parameter after I created this PR. 

So, what you are saying is that the comparison needs to be created for each class and in the case of classes like Float or PaddedInt (with length parameter) also the length will need to be object of the comparison.

Q: in your example, why are you returning an error instead of simply returning `False`?

sampsyo: `NotImplemented` is a special value that says that the comparison is not valid—for example, you're not supposed to do `'a' == 5` in Python, and it should be an error rather than simply saying "no." This blog post has some useful background:
https://eev.ee/blog/2012/03/24/python-faq-equality/

In hindsight, we should probably only do `NotImplemented` for things that are not `Type`s. We should use `False` for two things that are both `Type` instances but not of the same subclass.

adamjakab: Ok. In the light of this I think that we will need to test each `Type` separately.

PR #3534, opened 2020-03-30
fetchart: Add Last.fm artwork source

### LINKED ISSUE #3530: fetchart: Add last.fm source
## Use case

When downloading album art via `fetchart`, it would be useful to be able to query [last.fm](https://last.fm). This functionality is present in [airsonic](https://github.com/airsonic/airsonic), and I tried to find how it works and came across [this commit](https://github.com/airsonic/airsonic/commit/72851f072128134f4ef8c103146aa7596735183d) which seems to implement this functionality.

## Solution

The `fetchart` plugin should have a new source called `lastfm` or `last.fm` that queries last.fm.

## Alternatives

For some releases, the sources currently available for album art don't suffice. Thus, the only suitable alternative is to grab the art manually.

jackwilsdon:
Fixes #3530.

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -21,6 +21,7 @@ from contextlib import closing
 import os
 import re
 from tempfile import NamedTemporaryFile
+from collections import OrderedDict
 
 import requests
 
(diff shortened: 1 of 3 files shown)


beetsplug/fetchart.py
```diff
@@ -742,11 +743,73 @@ def get(self, album, plugin, paths):
...
+            })
+        except requests.RequestException:
+            self._log.debug(u'lastfm: error receiving response')
+            return
+
+        try:
+            data = response.json()
+
+            if 'error' in data:
+                if data['error'] == 6:
```
jackwilsdon: The [API docs](https://www.last.fm/api/show/album.getInfo) for this endpoint suggest that error 6 is "Invalid parameters", however this is the error I get when seaching for an unknown MBID.

jackwilsdon:
Thanks!

PR #3542, opened 2020-04-06
Add beets-originquery plugin

x1ppy:
I'm not familiar with the process for adding new plugins on the plugins page, but here's a PR in case you'd like to include it.

[`beets-originquery`](https://github.com/x1ppy/beets-originquery) integrates with the new beets [`extra_tags`](https://github.com/beetbox/beets/blob/master/docs/reference/config.rst#id70) config setting to improve autotagger results by augmenting the MusicBrainz query with locally-sourced data. This allows you to more easily enable the `extra_tags` functionality since you don't have to update the tags first; tags can simply be specified in an _origin file_. The origin file format is extremely flexible and configurable, allowing for compatibility with the wide variety of supplemental metadata files people often include with their music.

### DIFF
--- a/docs/plugins/index.rst
+++ b/docs/plugins/index.rst
@@ -314,6 +314,9 @@ Here are a few of the plugins written by the beets community:
 
 * `beets-bpmanalyser`_ analyses songs and calculates their tempo (BPM).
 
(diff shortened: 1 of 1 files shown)


sampsyo:
Wow; this is extremely cool! Thanks for the addition. :smiley:

PR #3554, opened 2020-04-16
Fix #3535 - Genius Null Check

### LINKED ISSUE #3535: genius: Tolerate missing lyrics div in Genius scraper
```
  File "/home/greg/.local/lib/python3.7/site-packages/beetsplug/lyrics.py", line 410, in fetch
    return self.lyrics_from_song_api_path(song_api_path)
  File "/home/greg/.local/lib/python3.7/site-packages/beetsplug/lyrics.py", line 376, in lyrics_from_song_api_path
    lyrics = html.find("div", class_="lyrics").get_text()
AttributeError: 'NoneType' object has no attribute 'get_text'
```

Missing error handling when processing lyrics downloads, this shouldn't fail that hard.

thejli21:
Added the null check with a unit test and sample Genius page.

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -373,7 +373,13 @@ class Genius(Backend):
 
         # At least Genius is nice and has a tag called 'lyrics'!
         # Updated css where the lyrics are based in HTML.
-        lyrics = html.find("div", class_="lyrics").get_text()
+        try:
+            lyrics = html.find("div", class_="lyrics").get_text()
+        except AttributeError as exc:
+            # html is a NoneType cannot retrieve lyrics
+            self._log.debug(u'Genius lyrics for {0} not found: {1}',
+                            page_url, exc)
+            return None
 
         return lyrics
 
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -373,13 +373,14 @@ class Genius(Backend):
 
         # At least Genius is nice and has a tag called 'lyrics'!
         # Updated css where the lyrics are based in HTML.
-        try:
-            lyrics = html.find("div", class_="lyrics").get_text()
-        except AttributeError as exc:
-            # html is a NoneType cannot retrieve lyrics
-            self._log.debug(u'Genius lyrics for {0} not found: {1}',
-                            page_url, exc)
+        lyrics_div = html.find("div", class_="lyrics")
+
+        # nullcheck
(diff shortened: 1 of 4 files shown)


beetsplug/lyrics.py
```diff
@@ -373,7 +373,13 @@ def lyrics_from_song_api_path(self, song_api_path):
 
         # At least Genius is nice and has a tag called 'lyrics'!
         # Updated css where the lyrics are based in HTML.
-        lyrics = html.find("div", class_="lyrics").get_text()
+        try:
+            lyrics = html.find("div", class_="lyrics").get_text()
+        except AttributeError as exc:
```
sampsyo, line 378: Cool! I think this might be a little clearer (and not require the code-like reasoning that currently appears in a comment) if we instead check specifically for None instead of generically for a missing attribute:

```py
lyrics_div = html.find(...)
if lyrics_div is None:
    self._log...
    return None
lyrics = lyrics_div.get_text()
```

What do you think about that minor restructuring?

test/test_lyrics.py
```diff
@@ -395,6 +423,44 @@ def test_is_page_candidate_special_chars(self):
...
+    def test_no_lyrics_div(self):
+        """Ensure that `lyrics_from_song_api_path` doesn't crash when the html
+        for a Genius page contain <div class="lyrics"></div>
+        """
+        # https://github.com/beetbox/beets/issues/3535
+        # expected return value None
+        try:
+            self.assertEqual(genius.lyrics_from_song_api_path('/nolyric'),
+                             None)
+        except AttributeError:
```
sampsyo, line 459: Instead of catching the exception and assessing, I think we can just let the exception through. That is, you can remove the `try`/`except` handler here completely—the test will error out if the bug exists.

beetsplug/lyrics.py
thejli21: Seems like it should have the same functionality, I'll rewrite it and test it again to make sure! Checking for None in general seems like a better failsafe, I think I was just using try-except because that's what the other error checking seemed to be doing.

test/test_lyrics.py
thejli21: Sure thing!

thejli21:
@sampsyo made the recommended changes, trimmed the example html, and made a small note in the changelog.

sampsyo:
Awesome; looks great! Thanks for taking care of this!

PR #3557, opened 2020-04-18
Get performer names and roles

dosoe:
Gets the performer names and instruments/roles, puts them into a long list in the form ```performer1 (instrument1), performer2 (instrument2)``` . It has a few shortcomings:

-   it uses the names of the performers, which can be tricky (think of artists that use another alphabet)
-   if a performer has several roles, he appears several times
-   I don't look at the precise roles: if someone sings Carmen in an opera, she will appear as ```performer (soprano vocals)```
-   it creates one tag ```performer```, not one tag for each instrument/role

However, as discussed in https://github.com/beetbox/beets/pull/2563#issuecomment-554990758 this could be a start to tackle this issue.

### DIFF
--- a/beets/autotag/__init__.py
+++ b/beets/autotag/__init__.py
@@ -186,6 +186,7 @@ def apply_metadata(album_info, mapping):
                 'composer',
                 'composer_sort',
                 'arranger',
+                'performer',
                 'work',
                 'mb_workid',
                 'work_disambig',
--- a/beets/autotag/__init__.py
+++ b/beets/autotag/__init__.py
@@ -60,6 +60,8 @@ def apply_item_metadata(item, track_info):
         item.composer_sort = track_info.composer_sort
     if track_info.arranger is not None:
         item.arranger = track_info.arranger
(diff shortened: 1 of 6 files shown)


dosoe:
Right now it looks at attributes, which can either be the instruments/voices or one of the three: ```additional```, ```guest```, ```solo```. This means that I sometimes end up with performers that look like: ```Nikolaus Harnoncourt (conductor), Concentus Musicus Wien (performing orchestra), Ludwig Baumann (bass vocals, solo), Kurt Equiluz (solo, tenor vocals), Philip Langridge (solo, tenor vocals), Marjana Lipovšek (alto vocals, solo), Thomas Moser (solo, tenor vocals), Felicity Palmer (solo, soprano vocals), Arnold Schoenberg Chor (choir vocals)```. MB being what it is, the implementation of these additional attributes is wildly inconsistent and (at least for classical music) rarely used. However, while I would tend to remove these attributes, some may want to keep them, so what should be done?

PR #3558, opened 2020-04-18
Add artist_sort as the last entry in the artists to search

hashhar:
This will handle cases where the artist name includes some
characters which can cause a search failure for some
websites - mainly unicode.

Fixes https://github.com/beetbox/beets/issues/3340.

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -206,7 +206,10 @@ def search_pairs(item):
         # Remove any featuring artists from the artists name
         r"(.*?) {0}".format(plugins.feat_tokens())]
     artists = generate_alternatives(artist, patterns)
-    artists.append(artist_sort)
+    # Use the artist_sort as fallback only if it differs from artist to avoid
+    # repeated remote requests with the same search terms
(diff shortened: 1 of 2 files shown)


hashhar:
@sampsyo I've got this working. I was now looking to add tests and could figure out the new cases that will go into `LyricsPluginTest.test_search_artist` test cases.

But I can see that there are also some integration tests which make actual network requests to verify things but it looks like they are testing the web scraping part (beautiful soup) instead of the actual plugin logic.

What do you think I should do for this change? Is a new test case which verifies that the `search_pairs` includes the `artist_sort` is enough? Or should I do something else?

sampsyo:
Nice! Looks great!!

For this, I think just testing `search_pairs` should be enough. Plugging that in to a mocked web request seems like it doesn't add too much.

hashhar:
@sampsyo This looks good now. Let me know if there are any docs or changelogs that need editing.

sampsyo:
Looks perfect!! Merged with a quick changelog entry. :sparkles:

PR #3559, opened 2020-04-19
fetchart: improve description of -q flag

somasis:
"shows only quiet art" just seems a bit opaque to me.

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -979,7 +979,7 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
         cmd.parser.add_option(
             u'-q', u'--quiet', dest='quiet',
             action='store_true', default=False,
-            help=u'shows only quiet art'
+            help=u'quiet mode: do not output albums that already have artwork'
         )
 
         def func(lib, opts, args):
--- a/docs/plugins/fetchart.rst
+++ b/docs/plugins/fetchart.rst
@@ -126,8 +126,9 @@ art::
 
     $ beet fetchart [-q] [query]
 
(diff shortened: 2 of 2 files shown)


beetsplug/fetchart.py
```diff
@@ -979,7 +979,7 @@ def commands(self):
         cmd.parser.add_option(
             u'-q', u'--quiet', dest='quiet',
             action='store_true', default=False,
-            help=u'shows only quiet art'
+            help=u'only output about art that is fetched, or cannot be found'
```
wisp3rwind, line 982: Maybe `u'quiet mode: do not output albums that already have artwork'` is even clearer?

docs/plugins/fetchart.rst
```diff
@@ -127,7 +127,8 @@ art::
     $ beet fetchart [-q] [query]
 
 By default the command will display all results, the ``-q`` or ``--quiet``
-switch will only display results for album arts that are still missing.
+switch will only display results for album arts that is fetched, or that is
+still missing.
```
wisp3rwind, line 131: I think it's also not very clear what the `results` are in this context, maybe referring to the 'query` clears this up:
```
By default the command will display all albums matching the ``query``. When the ``-q`` or ``--quiet``
switch is given, only albums that were missing artwork, and whether or not it could be found, will be output.
```

sampsyo:
Wow; you're right… "shows only quiet art" sounds more like an existential cry of despair than a command-line flag help string. :scream:

somasis:
How's that look?

sampsyo:
Looks great to me! Thank you for your help!! :sparkles:

wisp3rwind:
Looks good to me :tada:

PR #3560, opened 2020-04-19
Album Art Resize to filesize

davidswarbrick:
The subject of this PR was previously a Twitter plugin (details of which are included below for reference), however I now think that plugin was too specific for my personal use-case to be included in the wider Beets project, especially due to possible longer-term maintenance & support that could be required by the volunteer Beets community! The one useful & generalised development of the PR was to adjust the ArtResizer API to be able to target a filesize rather than just a quality, thus the contents and title of this PR have been adjusted to reflect this. I hope that it now seems like a useful addition to the project, and one that can be squashed & merged in the future! Thanks for reading as always.

> A plugin to tweet the output of an album query, highlights include:
> - [x] Twitter API keys included in configuration file
> - ~~`pretend` option to test query is correct~~ (replaced with `cautious` configuration option)
> - [x] Template for tweet using beets syntax (default is `$albumartist - $album ($year)`)
> - [x] Album Art finding (using the `.artpath` field provided by fetchart)
> - [x] Album Art resizing using ArtResizer
> - [x] Option to tweet without album art
> - ~~Handling of errors from Twitter API~~ (currently no exceptions documented & therefore none handled)
> - [x] Testing
> - [x] Documentation
> - [x] Extension to ArtResizer API to choose final filesize
> 
> The plugin depends on the `twitter` python library. A functioning example is shown [here](https://twitter.com/swarbsturntable). This plugin was created as a foray into Twitter API integration, and in order to be linked into MPD/BPD or shell scripts to provide a nice digital listening diary. I thought it may be worth submitting a pull request in case there's sufficient interest for fully tested & documented addition to the beets project. I'm a huge fan of the project so would love to contribute, but also understand this is a particularly niche use-case. Thanks for reading!


sampsyo:
Wow; this is super cool! Awesome work on this so far. The Twitter feed with album art as attached images looks very snazzy.

For the image resizing bit, feel free to extend the ArtResizer API to permit a "resize to X bytes" operation. One option would be to use the proposed shortcut for the IM backend but resort to the [necessary](https://stackoverflow.com/q/10607468/39182) guess-and-check loop for PIL.

davidswarbrick:
Thanks for the input @sampsyo! I've implemented an addition to the ArtResizer API as you suggested, adding an extra function argument to resize to X bytes (`max_filesize`) which uses either the IM workaround or a guess-and-check PIL implementation. Hopefully that may find other uses, for example allowing users to save space on cover art storage.

In order to test it's working I've written tests in `test/test_tweet.py` to call the PIL and ImageMagick functions directly, however this has required me to include `Pillow` as a new dependency for testing which seems to imply we may have avoided testing the different image conversion methods previously? I'm similarly wary adding the `twitter` package to the testing dependencies as there are currently no tests requiring the package beyond import (as I don't think the package implements any exceptions to be handled, although there may be scope for quality checking responses), thus there may be a case for replacement with MagicMocks I suspect. The existing linting errors have been explained by @jtpavlock, as they should be solved by #3694.

jtpavlock:
Hey @davidswarbrick, I just want to clarify my explanation on the linting I gave on [our discussion](https://discourse.beets.io/t/fixing-flake8-errors-for-plugin-contribution/1454/2).

In particular, 

> I would avoid applying black to existing files as it would potentially alter other code outside the scope of your contribution.

Currently, it looks like you have unintentionally made a decent amount of formatting changes to other parts of code that don't deal with your plugin. I see two potential solutions to this:

1. Rollback the unrelated formatting changes so this pull request only deals with code relevant to your plugin.
2. Wait until #3694 is merged, and then rebase off the new changes it would introduce.

As for the docstring errors, those won't go away with #3694 😄

davidswarbrick:
Aha thanks @jtpavlock ! You're right about the reasonably large impact formatting changes have had, I was hoping they weren't too extensive but on review it's more than I realised. I'll address the docstring errors first then see where my efforts will be best spent on the transition to black.

davidswarbrick:
I've fixed the formatting changes & addressed the windows errors, should be ready for review now.

davidswarbrick:
After considering the state of the Tweet plugin and how niche the use-case is, I've decided to remove it from the subject of this PR, and transfer its status to that of a "third party plugin". However, I believe the additions to the album art resizer remain useful, in that the original PR introduced the functionality to resize an album art image to a particular filesize, which could be useful for optimising libraries with lots of large artwork.

beets/util/artresizer.py
```diff
@@ -83,14 +84,50 @@ def pil_resize(maxwidth, path_in, path_out=None, quality=0):
...
     except IOError:
         log.error(u"PIL cannot create thumbnail for '{0}'",
                   util.displayable_path(path_in))
         return path_in
+    except AssertionError:
+        log.error(
+            u"Percentage filesize decrease calculated incorrectly for '{0}'",
+            util.displayable_path(path_in),
+        )
+        return path_in
```
wisp3rwind, line 127: I think this should be dropped: Either the `assert` truly has the purpose of an assertion, i.e. checks for programmer error, and should be impossible to happen, or it is an error that depends on runtime state. Then, just directly raise some error or `return path_in`, without the intermediate assertion.

beets/util/artresizer.py
```diff
@@ -83,14 +84,50 @@ def pil_resize(maxwidth, path_in, path_out=None, quality=0):
...
+                lower_qual = lower_qual * frac_diff
+                # Restrict quality dropping below 10
+                if lower_qual < 10:
+                    lower_qual = 10
+                # Use optimize flag to improve filesize decrease
+                im.save(
+                    util.py3_path(path_out), quality=lower_qual, optimize=True
+                )
+            log.error(u"PIL Failed to resize file to below {0}B", max_filesize)
+            raise IOError
```
wisp3rwind, line 114: Maybe return the resized image here anyway? That's presumably still closer to the target size than the original.

test/test_art_resize.py
```diff
@@ -0,0 +1,87 @@
...
+        new_file_size = os.stat(im).st_size
+        # check size has decreased
+        self.assertLess(new_file_size, self.IMG_225x225_SIZE)
+        # check size has decreased enough
+        self.assertLess(new_file_size, self.LOWER_MAX_FILESIZE)
+
+    def test_pil_file_resize(self):
+        """Test PIL resize function is lowering file size."""
+        if get_pil_version():
+            self._test_img_resize(pil_resize)
```
wisp3rwind, line 73: Maybe add
```
else:
    self.skipTest("PIL not available")
```
or even shorter decorate the function with
```
@unittest.skipUnless(get_pil_version(), "PIL not available")
```
and the same for ImageMagick below. Otherwise, the tests appear to succeed in spite of not even being run.

wisp3rwind (changes requested):
> After considering the state of the Tweet plugin and how niche the use-case is, I've decided to remove it from the subject of this PR, and transfer its status to that of a "third party plugin". However, I believe the additions to the album art resizer remain useful, in that the original PR introduced the functionality to resize an album art image to a particular filesize, which could be useful for optimising libraries with lots of large artwork.

Absolutely, that does sound useful! I left a few comments about slight improvements to the logic and compatibility.

> In order to test it's working I've written tests in test/test_tweet.py to call the PIL and ImageMagick functions directly, however this has required me to include Pillow as a new dependency for testing which seems to imply we may have avoided testing the different image conversion methods previously?

I don't think there is a need to add Pillow to the requirements. The way this is handled for other dependencies (or platform-specific features) is to explicitly skip tests which are missing dependencies.

test/test_art_resize.py
```diff
@@ -0,0 +1,87 @@
...
+        self.setup_beets()
+
+    def tearDown(self):
+        """Called after each test, unloading all plugins."""
+        self.teardown_beets()
+
+    def _test_img_resize(self, resize_func):
+        """Wrapper function to test resizing based on file size."""
+        # Check "lower maximum filesize" is truly lower than original filesize
+        self.assertLess(self.LOWER_MAX_FILESIZE, self.IMG_225x225_SIZE)
```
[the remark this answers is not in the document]
davidswarbrick: Yes this is definitely a neater solution, I've implemented it, thank you!

docs/changelog.rst
```diff
@@ -16,9 +16,9 @@ New features:
 * A new :ref:`extra_tags` configuration option allows more tagged metadata
   to be included in MusicBrainz queries.
 * A new :doc:`/plugins/fish` adds `Fish shell`_ tab autocompletion to beets
-* :doc:`plugins/fetchart` and :doc:`plugins/embedart`: Added a new ``quality``
-  option that controls the quality of the image output when the image is
-  resized.
+* :doc:`plugins/fetchart` and :doc:`plugins/embedart`: Added new ``quality`` and
+  ``max_filesize`` options that control the quality & filesize of the output when an
+  image is resized.
```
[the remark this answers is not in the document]
davidswarbrick: Good point, I've exposed max_filesize via the fetchart plugin in the latest commit, although it's not yet documented and tested, will do those both soon.

beets/util/artresizer.py
```diff
@@ -83,14 +84,50 @@ def pil_resize(maxwidth, path_in, path_out=None, quality=0):
...
+                filesize = os.stat(path_out).st_size
+                log.debug(u"PIL Pass {0} : Output size: {1}B", i, filesize)
+                if filesize <= max_filesize:
+                    return path_out
+                # Calculate percentage decrease required. This is approximate,
+                # as quality may only loosely correlate with filesize.
+                frac_diff = 1 - ((filesize - max_filesize) / max_filesize)
+                assert frac_diff < 1
+                lower_qual = lower_qual * frac_diff
+                # Restrict quality dropping below 10
```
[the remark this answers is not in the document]
davidswarbrick: Regarding the second question, there isn't anything on the PIL docs but I've cast it to an integer just in case. I did test this algorithm on some files a while ago but I'll try and get some example numbers to share, my aim was to get below the target file size as quickly as possible, but yes I agree we could change it later easily too. Another question that's semi-related but worth asking - do you think there'd be any preference between `maxfilesize` and `max_filesize` (just aesthetically given `maxwidth`)?

beets/util/artresizer.py
davidswarbrick: Agreed, I think this was left in from writing the algorithm when I was making some basic mathematical errors, but it serves no purpose anymore.

test/test_art_resize.py
davidswarbrick: This is really helpful, thank you!

beets/util/artresizer.py
davidswarbrick: Yeah, I've done that now. Should I adjust the log to a warning?

wisp3rwind: That sounds like a good idea!

beets/util/artresizer.py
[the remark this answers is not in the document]
wisp3rwind: I don't really have a clear grasp on beets' preferences here, so sorry, I don't know. I guess there is no stringent style requirement in that regard. Maybe keep `max_filesize`, since the identifier is quite long and the underscore helps readability.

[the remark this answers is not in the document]
sampsyo: No particular institutional preference, IMO. But personally I think the more legible snake_case version is probably better. :smiley:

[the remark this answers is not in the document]
davidswarbrick: I've replaced this logic with -10 quality each time - this avoids issues with files being over double the max_filesize, and attempting to compensate for the unpredictable quality/filesize relationship for JPG which will be image-dependent. Testing with a 16.4MB jpg: quality 95 = 4.1MB, 80 = 1.3MB, 70 = 877.6kB, 10 = 130.3kB

docs/changelog.rst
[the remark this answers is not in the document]
davidswarbrick: Apologies for the delay, tests & documentation have been added.

davidswarbrick:
Hey @wisp3rwind / others, sorry for the delay on this but it should be ready for a final review or merging now.

beetsplug/fetchart.py
```diff
@@ -122,22 +128,49 @@ def _validate(self, plugin):
...
+                                         max_filesize=plugin.max_filesize)
+        elif self.check == self.CANDIDATE_DOWNSCALE:
             self.path = ArtResizer.shared.resize(plugin.maxwidth, self.path,
                                                  quality=plugin.quality)
+        elif self.check == self.CANDIDATE_DOWNSIZE:
+            # dimensions are correct, so maxwidth is set to maximum dimension
+            self.path = \
+                ArtResizer.shared.resize(max(self.size), self.path,
+                                         quality=plugin.quality,
+                                         max_filesize=plugin.max_filesize)
```
wisp3rwind, line 173: Wouldn't the following have the same effect? I don't see the need to set the `maxwidth` or `max_filesize` arguments in this way, the `resize` method should behave (essentially) the same
```suggestion
        if self.check in [self.CANDIDATE_DOWNSCALE,
                self.CANDIDATE_DOWNSIZE,
                self.CANDIDATE_DOWNSCALE_AND_DOWNSIZE]:
            self.path = ArtResizer.shared.resize(plugin.maxwidth, self.path,
                                         quality=plugin.quality,
                                         max_filesize=plugin.max_filesize)
```
Wouldn't that also mean that just one flag `CANDIDATE_DOWNSIZE` is enough since the action is the same for all three cases?

wisp3rwind (changes requested):
I left a few inline comments where I think tests might be improved and the plugin might be simplified.

When merging I would want to squash this PR to remove the remainder of the Twitter plugin. Is this fine with you, or what you rather reorganize these commits yourself?

beetsplug/fetchart.py
davidswarbrick: Your suggestion is certainly neater, however there remains a use-case where a user may want to target a maximum pixel width but doesn't care about the filesize, thus would prefer the original functionality of `CANDIDATE_DOWNSCALE`. In this case, `max_filesize` may have not been set, or the file is already lower than the maximum filesize.

Thanks for pointing out that `CANDIDATE_DOWNSIZE` is exactly the same as `CANDIDATE_DOWNSCALE_AND_DOWNSIZE` - I'll condense these down into one option.

davidswarbrick: Yeah sorry for deleting my earlier comment, I was trying to think through all of the possible combinations & any downsides. I think a good compromise, due to the default behaviour as you've pointed out, is to combine the two `DOWNSCALE` options. The `DOWNSIZE`-only option still has a use for those who don't want to rescale their images, just lower filesize (a use-case relevant for the Twitter example).

The only good reason I can think of to keep the combined option would be if setting `max_filesize` incurred a performance penalty when using the ImageMagick backend with an image _already below_ that filesize, however that's extremely unlikely. Using the PIL backend it would have no effect. Initially I wrote the code like this because I wasn't sure we could trust the initial values (I misinterpreted the checks for `if plugin.maxwidth` as them being set to `None` - I realise now they are initialised to zero in the default config). Thanks so much for your scrutiny here! Apologies again for my confusion.

test/test_art_resize.py
```diff
@@ -0,0 +1,101 @@
...
+        self.assertExists(im_unchanged)
+        self.assertExists(im_a)
+        self.assertExists(im_b)
+
+        # check size has decreased enough
+        # (unknown behaviour for quality-only setting)
+        self.assertLess(os.stat(syspath(im_a)).st_size,
+                        self.IMG_225x225_SIZE)
+        self.assertLess(os.stat(syspath(im_b)).st_size,
+                        self.IMG_225x225_SIZE)
```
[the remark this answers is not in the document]
davidswarbrick: Thanks a lot for this, that's a really good point and something I'd vaguely thought about but not addressed at all, I'll alter the tests now to make them more stringent.

davidswarbrick:
Thanks so much for the review @wisp3rwind . Please feel free to squash this PR, it has a particularly messy commit history, and the legacy twitter plugin has been moved to a separate personal branch.

wisp3rwind:
> Thanks so much for the review @wisp3rwind . Please feel free to squash this PR, it has a particularly messy commit history, and the legacy twitter plugin has been moved to a separate personal branch.

Sorry for dragging this along so long @davidswarbrick, I'll try to give this a final review soon!

wisp3rwind:
> > Thanks so much for the review @wisp3rwind . Please feel free to squash this PR, it has a particularly messy commit history, and the legacy twitter plugin has been moved to a separate personal branch.
> 
> Sorry for dragging this along so long @davidswarbrick, I'll try to give this a final review soon!

Sorry for the empty promise here! I went ahead and squashed and merged the PR. Thanks for the contribution :tada:

PR #3563, opened 2020-04-22
Fix #3533 - crash when calculating RG/R128 mix

### LINKED ISSUE #3533: replaygain: Don't crash when albums have a mix of ReplayGain and EBU R128 formats
```
beetsplug.replaygain.ReplayGainError: Mix of ReplayGain and EBU R128 detected for some tracks in album  -
```

The message isn't accurate: pretty much all of my library is Opus files with both tags — this is not about the actual tags in the files, this is about formats, the code checks `should_use_r128` — this actually seems to mean "you've mixed Opus and not-Opus in the same album". Which.. shouldn't be a fatal error??

autrimpo:
I've moved the check into the `try` block as per the discussion in #3533.
I've also rewritten the message, hopefully it is clearer now.

### DIFF
--- a/beetsplug/replaygain.py
+++ b/beetsplug/replaygain.py
@@ -1326,10 +1326,10 @@ class ReplayGainPlugin(BeetsPlugin):
 
         if (any([self.should_use_r128(item) for item in album.items()]) and not
                 all(([self.should_use_r128(item) for item in album.items()]))):
-            raise ReplayGainError(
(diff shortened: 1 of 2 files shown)


autrimpo:
I've moved the check back and replaced the exception with log + return :+1:

sampsyo:
Awesome; thanks!! I switched the log level to error and merged. :tada: :sparkles:

PR #3567, opened 2020-04-25
web plugin: support path queries

### LINKED ISSUE #3566: web plugin: Can't do path queries
### Problem

The beets web plugin provides a [JSON API](https://beets.readthedocs.io/en/stable/plugins/web.html?highlight=web#json-api). This JSON API has a `/item/query` endpoint. Using this endpoint I want to perform a [path query](https://beets.readthedocs.io/en/stable/reference/query.html#path-queries). For example:

```
path:/home/nmeum/music/Singles
```

Unfortunately, such path queries do not work with the JSON API due to the `QueryConverter` which is used for the `/query` endpoint:

 https://github.com/beetbox/beets/blob/d43d54e21cde97f57f19486925ab56b419254cc8/beetsplug/web/__init__.py#L175-L183

This class converts the query from above to:

```
path:home,nmeum,music,Singles
```

which obviously does not work as intended. It doesn't make a difference whether the path query is URL encoded or not. Maybe this could be fixed by invoking the `QueryConverter` before URL decoding is performed? Additionally, it seems that the `QueryConverter` behaviour is not actually documented in the JSON API doc.

This issue cannot be workaround by using the `/item/path/...` endpoint, because this endpoint returns a different JSON document.

See also: https://github.com/pallets/flask/issues/900

### Setup

* OS: Alpine Linux
* Python version: 3.8.2
* beets version: 1.4.9
* Turning off plugins made problem go away (yes/no): N/A

nmeum:
The QueryConverter was added without much further explanation in commit
004e9a8144cd120f6f245dd9865db469c243d34d. Unfortunately, the query
converter breaks path queries as flask cannot distinguish between an
URL encoded and a plain '/' character during routing [\[0\]][0]. Additionally,
I don't get why the QueryConverter is needed in the first place as ','
characters can just be used in the URL directly. For example:
`/item/query/foo,bar`. As conversions of slashes into comma characters
wasn't documented anyhow removing this code shouldn't cause much harm.

Fixes #3566

[0]: https://github.com/pallets/flask/issues/900

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -177,10 +177,11 @@ class QueryConverter(PathConverter):
     """
 
     def to_python(self, value):
-        return value.split('/')
+        queries = value.split('/')
+        return [query.replace('\\', os.sep) for query in queries]
 
     def to_url(self, value):
-        return ','.join(value)
+        return ','.join([v.replace(os.sep, '\\') for v in value])
 
 
 class EverythingConverter(PathConverter):
--- a/docs/plugins/web.rst
+++ b/docs/plugins/web.rst
@@ -213,7 +213,11 @@ If the server runs UNIX, you'll need to include an extra leading slash:
 Returns a list of tracks matching the query. The *querystring* must be a
 valid query as described in :doc:`/reference/query`. Path elements are
 joined as query keywords. For example, ``/item/query/foo/bar`` will be
-converted to the query ``foo,bar``. ::
+converted to the query ``foo,bar``. As this conflicts with using a slash
+character as a path seperator in path queries, a backlash character
+should be used in these queries instead. This character is converted to
+the path seperator actually used by the operating system before the
+query is performed. ::
 
     {
       "results": [
--- a/docs/plugins/web.rst
+++ b/docs/plugins/web.rst
@@ -210,7 +210,10 @@ If the server runs UNIX, you'll need to include an extra leading slash:
 ``GET /item/query/querystring``
 +++++++++++++++++++++++++++++++
 
-Returns a list of tracks matching the query. The *querystring* must be a valid query as described in :doc:`/reference/query`. ::
+Returns a list of tracks matching the query. The *querystring* must be a
+valid query as described in :doc:`/reference/query`. Path elements are
+joined as query keywords. For example, ``/item/query/foo/bar`` will be
+converted to the query ``foo,bar``. ::
 
     {
       "results": [


sampsyo:
Seems like a good idea overall, but let's do some due diligence to make sure this won't break anything obvious (because it's a breaking change that could cause problems for clients that relied on being able to query with slashes).

Unfortunately, a little searching has already turned up one such example: mopidy-beets is using this endpoint for all its music searches.
https://github.com/mopidy/mopidy-beets/blob/1af80f230da411eb9e203ffec41c1b5adcc2ac1c/mopidy_beets/client.py#L184

Maybe we need to think creatively here about how to support this kind of query without breaking stuff like that? A new path query endpoint, for example, or a separate style that allows arbitrary query strings?

nmeum:
> Maybe we need to think creatively here about how to support this kind of query without breaking stuff like that?

One *creative* quick-fix solution I was considering is using windows path separators, i.e. backslash, in the query URL. On Unix systems the web plugin can convert these back to Unix path separators, i.e. slash, before performing the actual query. Obviously a hacky solution but maybe a solution which introduces less breakage?

sampsyo:
Yeah, that's a reasonable idea! One other option would be to optionally support an HTTP query string version of the `query` endpoint. That would let `item/query?q=path:/foo/bar` work as an alternative way to specify the query string directly. It's also somewhat aesthetically pleasing because it reflects that queries can truly be arbitrary strings…

nmeum:
> Yeah, that's a reasonable idea!

Implemented this idea, I am not too familiar with flask and werkzeug so not sure if the implementation is entirely correct as is. If you like the implementation as is I would also add some documentation for both the `/ → ,` and the `\ → os.sep` behaviour. So let me know what you think.

beetsplug/web/__init__.py
```diff
@@ -177,10 +177,12 @@ class QueryConverter(PathConverter):
     """
 
     def to_python(self, value):
-        return value.split('/')
+        queries = value.split('/')
+        return list(map(lambda query: query.replace('\\', os.sep), queries))
```
sampsyo, line 181: An easier way to write `list(map(lambda...))` is with a list comprehension:

```py
[query.replace(...) for query in queries]
```

beetsplug/web/__init__.py
```diff
@@ -177,10 +177,12 @@ class QueryConverter(PathConverter):
...
 
     def to_url(self, value):
-        return ','.join(value)
+        value = map(lambda v: v.replace(os.sep, '\\'), value)
```
sampsyo, line 184: As above, a list comprehension could help here.

nmeum:
Adjusted the code for list comprehensions, didn't get around to testing this version yet though. Also added some documentation for the query end point of the web plugin.

nmeum:
> Adjusted the code for list comprehensions, didn't get around to testing this version yet though.

Finally gotten around to testing this again. Found a bug while at it, seems to work now.

sampsyo:
Looks great; thanks!! Merged with a few tweaks to the wording.

PR #3568, opened 2020-04-25
First try adding new albuminfo and trackinfo class

dosoe:
New attempt at dealing with https://github.com/beetbox/beets/issues/1547 learning from errors in https://github.com/beetbox/beets/pull/2650

### DIFF
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -148,6 +148,16 @@ class AlbumInfo(AttrDict):
             for track in self.tracks:
                 track.decode(codec)
 
+    def dup_albuminfo(self):
+        dupe = AlbumInfo()
+        dupe.update(self)
+        if 'tracks' in self:
+            tracks = []
+            for track in self.tracks:
+                tracks.append(track.dup_trackinfo())
+            dupe.tracks = tracks
+        return dupe
+
 
 class TrackInfo(AttrDict):
     """Describes a canonical track present on a release. Appears as part
@@ -208,6 +218,10 @@ class TrackInfo(AttrDict):
             value = getattr(self, fld)
             if isinstance(value, bytes):
                 setattr(self, fld, value.decode(codec, 'ignore'))
+
+    def dup_trackinfo(self):
+        dupe = TrackInfo()
+        dupe.update(self)
 
 
 # Candidate distance scoring.
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -165,6 +165,10 @@ class AlbumInfo(Map):
             if isinstance(value, bytes):
                 setattr(self, fld, value.decode(codec, 'ignore'))
 
+        if 'tracks' in self:
+            for track in self.tracks:
+                track.decode(codec)
+
 
 class TrackInfo(Map):
     """Describes a canonical track present on a release. Appears as part
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -60,10 +60,6 @@ class Map(dict):
     def __setattr__(self, key, value):
         self.__setitem__(key, value)
 
-    def __setitem__(self, key, value):
-        super(Map, self).__setitem__(key, value)
-        self.__dict__.update({key: value})
-
     def __delattr__(self, item):
         self.__delitem__(item)
 
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -100,14 +100,9 @@ class AlbumInfo(Map):
     ``mediums`` along with the fields up through ``tracks`` are required.
     The others are optional and may be None.
     """
-    def __init__(self, album, album_id, artist, artist_id, tracks, **kwargs):
-        self.album = album
-        self.album_id = album_id
-        self.artist = artist
-        self.artist_id = artist_id
-        self.tracks = tracks
+    def __init__(self, **kwargs):
         for arg in kwargs:
-            self.__setattr__(arg, kwargs[arg]) 
+            self.__setattr__(arg, kwargs[arg])
 
     # Work around a bug in python-musicbrainz-ngs that causes some
     # strings to be bytes rather than Unicode.
@@ -116,19 +111,11 @@ class AlbumInfo(Map):
         """Ensure that all string attributes on this object, and the
         constituent `TrackInfo` objects, are decoded to Unicode.
         """
-        for fld in ['album', 'artist', 'albumtype', 'label', 'artist_sort',
-                    'catalognum', 'script', 'language', 'country', 'style',
-                    'genre', 'albumstatus', 'albumdisambig',
-                    'releasegroupdisambig', 'artist_credit',
-                    'media', 'discogs_albumid', 'discogs_labelid',
-                    'discogs_artistid']:
+        for fld in self:
             value = getattr(self, fld)
-            if isinstance(value, bytes):
-                setattr(self, fld, value.decode(codec, 'ignore'))
-
-        if self.tracks:
-            for track in self.tracks:
-                track.decode(codec)
+            if type(value) == str:
+                if isinstance(value, bytes):
+                    setattr(self, fld, value.decode(codec, 'ignore'))
 
 
 class TrackInfo(Map):
@@ -142,22 +129,20 @@ class TrackInfo(Map):
     may be None. The indices ``index``, ``medium``, and ``medium_index``
     are all 1-based.
     """
-    def __init__(self, title, track_id, **kwargs):
-        self.title = title
-        self.track_id = track_id
+    def __init__(self, **kwargs):
         for arg in kwargs:
-            self.__setattr__(arg, kwargs[arg]) 
+            self.__setattr__(arg, kwargs[arg])
 
     # As above, work around a bug in python-musicbrainz-ngs.
     def decode(self, codec='utf-8'):
         """Ensure that all string attributes on this object are decoded
         to Unicode.
         """
-        for fld in ['title', 'artist', 'medium', 'artist_sort', 'disctitle',
-                    'artist_credit', 'media']:
+        for fld in self:
             value = getattr(self, fld)
-            if isinstance(value, bytes):
-                setattr(self, fld, value.decode(codec, 'ignore'))
+            if type(value) == str:
+                if isinstance(value, bytes):
+                    setattr(self, fld, value.decode(codec, 'ignore'))
 
 
 # Candidate distance scoring.
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -45,11 +45,6 @@ class Map(dict):
     m = Map({'first_name': 'Eduardo'}, last_name='Pool', age=24,
     sports=['Soccer'])
     """
-    def __init__(self, **kwargs):
-        super(Map, self).__init__(**kwargs)
-        if kwargs:
-            for k, v in kwargs.iteritems():
-                self[k] = v
 
     def __getattr__(self, attr):
         if attr in self:
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -45,13 +45,8 @@ class Map(dict):
     m = Map({'first_name': 'Eduardo'}, last_name='Pool', age=24,
     sports=['Soccer'])
     """
-    def __init__(self, *args, **kwargs):
-        super(Map, self).__init__(*args, **kwargs)
-        for arg in args:
-            if isinstance(arg, dict):
-                for k, v in arg.iteritems():
-                    self[k] = v
-
+    def __init__(self, **kwargs):
+        super(Map, self).__init__(**kwargs)
         if kwargs:
             for k, v in kwargs.iteritems():
                 self[k] = v
@@ -76,15 +71,8 @@ class Map(dict):
         super(Map, self).__delitem__(key)
         del self.__dict__[key]
 
-    def __getstate__(self):
-        return self.__dict__
-
-    def __setstate__(self, state):
-        for key in state:
-            self.__setattr__(key, state[key])
-
     def __hash__(self):
-        return hash(tuple(sorted(self.items())))
+        return self.id
 
 
 class AlbumInfo(Map):
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -39,49 +39,53 @@ except AttributeError:
 
 
 # Classes used to represent candidate options.
-class Map(dict): 
-     """ 
-     Example: 
-     m = Map({'first_name': 'Eduardo'}, last_name='Pool', age=24, sports=['Soccer']) 
-     """ 
-     def __init__(self, *args, **kwargs): 
-         super(Map, self).__init__(*args, **kwargs) 
-         for arg in args: 
-             if isinstance(arg, dict): 
-                 for k, v in arg.iteritems(): 
-                     self[k] = v 
-  
-         if kwargs: 
-             for k, v in kwargs.iteritems(): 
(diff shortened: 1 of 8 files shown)


dosoe:
This new implementation of AlbumInfo and TrackInfo allows flexible tags. A quick test with some of my data seems to work. It has one significant difference: because I overcharged ```__getattr__``` the function ```hasattr(track, tag)``` will always return ```True```. To check if a tag is set, one must check if ```track.tag==None```. This concerns especially plugins, as they sometimes add new tags, but it might be important elsewhere.

dosoe:
This behaviour of ```__getattr__``` is what makes the tests fail, as it is needed in python2 for ```copy.deepcopy(x)``` . One solution would be to adapt ```__getattr__``` by telling it to raise an ```AttributeError``` each time we want to get a tag that either does not exist or whose value is ```None```. @sampsyo , could that be a concern or can I implement it? Are there tags that should exist with the value ```None``` that should be kept as such?

dosoe:
It raises the question: The goal of this is to have some core attributes that are always populated and then some that are flexibly attributed. Which attributes are core attributes? From the definition of ```AlbumInfo```, it seems that only ```album, album_id, artist, artist_id, tracks``` are core (as the others are optional) but in the tests it calls for the other values. 
Do we want all tags to return ```None``` when called (and not populated) or do we want to raise an ```AttributeError```? The latter is compatible with ```copy.deepcopy``` in python2, the former isn't.

sampsyo:
Great! Thanks for getting this going!

Here's what I think we should do:

- *Everything* should be flexible, i.e., no built-in attributes.
- As a purely transitional matter, maybe we should provide a list of attributes that should produce `None` when they are missing instead of raising an `AttributeError`.
- When you access any other attribute, like `info.foo` and there's no value for `foo`, it should raise an exception rather than silently returning `None`.
- We should probably provide key-style access, like `info['foo']` and `info.get('foo')`. In the future, we can make code that consumes this stuff work more like that.

Does that seem reasonable? Perhaps I'm being too aggressive here about eliminating all built-in attributes, but I think it is probably the right thing to do.

dosoe:
I think in principle, yes, but I'm running into trouble adapting ```beets/autotag/__init__.py``` as some tags (like ```title```) are hardcoded. I'm looking into this.

dosoe:
Ok, that's gonna be a wild chase to find all the ```if item.attr:``` and replace them with ```if attr in item``` and check before calling each attribute if the attribute exists else I will bet ```AttributeError``` all over the place. Maybe you could help me to find all instances, @sampsyo ? Or maybe there is a better way?

dosoe:
>     * We should probably provide key-style access, like `info['foo']` and `info.get('foo')`. In the future, we can make code that consumes this stuff work more like that.

This already works.

>     * When you access any other attribute, like `info.foo` and there's no value for `foo`, it should raise an exception rather than silently returning `None`.

That's already the case. 
So far I'm working with everything being flexible, but then there's a lot to replace in the code, you might be able to help as you know the code better than I do.

dosoe:
That's a working prototype. Now I kept all the default values (all the tags that were previously set to ```None``` previously still are) as chasing all their occurrences and replacing every test has proven too difficult for me. However, that could be done one tag at a time now. Also, ```apply_metadata``` and ```apply_item_metadata``` from ```__init__.py``` have not be changed, that's also something that could be implemented later, however, I don't really know what it does so I don't know how to properly do it.

dosoe:
I do need some of the methods, because it needs to be hashable and I need a ```__setstate__``` and a ```__getstate__``` for deepcopy operations. As for the rest, I will try to clean up a bit the constructor and check which method is essential. What do you mean about the ```__dict__``` part of the functionality?

sampsyo:
I see… can you elaborate a little bit on which parts of the code require those two things (hashability and deep copying)?

For hashability, I think we might _not_ want a smart approach that hashes based on the data… we might want every object to be "unequal to" every other unique object. I think that's the default for `object` but not for `dict`.

dosoe:
I remember deepcopy being a problem at my previous attempt, especially for python2. It seems like it's linked to pickle and needs a ```__getstate__``` and a ```__setstate__``` method, that's also in the commentaries of this StackOverflow answer: https://stackoverflow.com/questions/2352181/how-to-use-a-dot-to-access-members-of-dictionary/32107024#32107024 . I also found this error in one of the test logs of my previous attemps to implement flexible classes (https://github.com/beetbox/beets/pull/2650): 
```

  File "/home/travis/build/beetbox/beets/test/test_autotag.py", line 436, in test_comp_track_artists_do_not_match

    self.assertNotEqual(self._dist(items, info), 0)

  File "/home/travis/build/beetbox/beets/test/test_autotag.py", line 340, in _dist

    return match.distance(items, info, self._mapping(items, info))

  File "/home/travis/build/beetbox/beets/beets/autotag/match.py", line 251, in distance

    dist.tracks[track] = track_distance(item, track, album_info.va)

TypeError: unhashable type: 'TrackInfo'
```

sampsyo:
OK—would you mind double-checking to see whether deep copying is still necessary anywhere? It might work to just delete those various methods and see if any tests fail…

That traceback is an example showing that the type does indeed need to be hashable, but this doesn't necessarily imply that the implementation should use the actual contents instead of the object identity. In fact, I think it would be incorrect to make two objects be considered equal in that context if their contents are equal. I think a default-ish implementation like `id(self)` would be appropriate?

dosoe:
```grep deepcopy``` shows it only appears in some tests (namely ```test_ui``` and ```test_autotag```) so we should be able to survive without. I will arrange the hash.

**Comment by dosoe** [comment]
It seems to work fine with all the methods I removed. There are only 4 methods remaining: 
```__init__```, ```__getattr__```, ```__setattr__``` and ```__hash__``` which are all needed somewhere.

**Comment by dosoe** [comment]
To me, the next logical step would be to scale down the number of tags that are preset to ```None``` . Another question: the tags that are decoded in ```AlbumInfo.decode()``` and ```TrackInfo.decode()``` seem to be chosen pretty randomly, is there a specific reason for these tags? When they don't get decoded they raise an error in ```hooks._string_dist_basic()``` , why? Can we arrange it to not be a list of tags but to detect which tags need to be decoded?

File: test/test_autotag.py [hunk 1]
```diff
@@ -749,13 +752,75 @@ def test_albumtype_applied(self):
         self.assertEqual(self.items[1].albumtype, 'album')
 
     def test_album_artist_overrides_empty_track_artist(self):
-        my_info = copy.deepcopy(self.info)
```
sampsyo, line 752: One way to make a copy of a TrackInfo or AlbumInfo object would be to use `dict`'s [`update` method](https://docs.python.org/3/library/stdtypes.html#dict.update). Something like this:

```
trackinfo1 = TrackInfo(title='foo')
trackinfo2 = TrackInfo()
trackinfo2.update(trackinfo1)
```

That has the effect of copying all the keys & values from one dict into another.

beets/autotag/hooks.py
```diff
@@ -120,6 +117,8 @@ def __init__(self, album, album_id, artist, artist_id, tracks, asin=None,
         self.discogs_albumid = discogs_albumid
         self.discogs_labelid = discogs_labelid
         self.discogs_artistid = discogs_artistid
+        for arg in kwargs:
+            self.__setattr__(arg, kwargs[arg])
```
sampsyo, line 121: A shorter way to do this:

```
self.update(kwargs)
```

beets/autotag/hooks.py
```diff
@@ -39,8 +39,27 @@
 
 
 # Classes used to represent candidate options.
+class Map(dict):
+    """
+    Example:
+    m = Map({'first_name': 'Eduardo'}, last_name='Pool', age=24,
+    sports=['Soccer'])
```
sampsyo, line 46: This class could use a more direct docstring. Also, maybe something like `AttrDict` is a better name than `Map`? A dict is already a map, after all…

test/test_autotag.py
dosoe: or even simpler: 
```
trackinfo1 = TrackInfo(title='foo')
trackinfo2 = TrackInfo(trackinfo1)
```
However, for updating ```AlbumInfo``` I need to recreate the tracks, don't I?

dosoe:
> But to validate that approach, maybe we should check whether, in a given run of the tool, any values are _supposed_ to be `bytes` and should _not_ be converted to `str`.

To me it would be shorter to make a list of the tags that shouldn't be decoded, rather than those that should, as most tags are just strings by default, except tracks, numbers and maybe dates.

sampsyo:
Sure! If only we could obtain a list of tags that should not be converted. :smiley: An experimental approach might be the best way to find out…

test/test_autotag.py
sampsyo: Indeed; good point if that's still supported in `__init__`. Maybe a good way to do this would be to create a little function for duplicating `AlbumInfo` objects—something like this?

```py
def dup_albuminfo(ai):
    return AlbumInfo(ai, tracks=[TrackInfo(ti) for ti in ai.tracks])
```

dosoe:
I did a few tries, it seems like ```decode``` mainly converts from ```str``` to ```unicode````.

sampsyo:
Yep, that's all it does—it only works on bytestrings (i.e., `bytes` on Python 3) to text strings (`str` on Python 3). The only question is whether there are fields that _want_ to remain as bytes.

dosoe:
The only ones I could think of are acoustic fingerprints and similar pseudo-strings

dosoe:
Whatever it is, it's probably simpler to just decode all strings into unicode and add a list of exceptions rather than doing it the other way. Is there a reason to convert everything to unicode?

sampsyo:
Yeah: the main reason to use the "positive" rather than "negative" approach is that, with this PR, the list of fields is meant to be extensible. So when someone comes along and adds a new field, perhaps in a plugin, we don't want to have to touch the rest of the code. We would need to choose a "default behavior," and the most sensible default is *not* to touch any data—to pass it along as the metadata source provided it.

dosoe:
But why do you actually bother to convert anything to unicode? Are regulat bytestrings not good enough? Is it for special characters (cyrillic, japanese, chinese etc.)?

sampsyo:
Yes—eventually, all strings that represent text in beets must be Unicode strings. That's the only way to reliably represent the full range of characters people use in their metadata.

dosoe:
Then it would make sense to make the default behaviour to convert every string to unicode unless stated otherwise.

sampsyo:
But that runs into the problem above: what if a plugin wants to add a field that is _supposed_ to contain bytes? Especially if it's not a built-in beets plugin, it would have no way to instruct beets core to skip the conversion.

dosoe:
Why not? A field gets converted only if it exists.

sampsyo:
Say I write a plugin that provides a fingerprint tag, $myfp. It holds a byte string, intentionally. The beets core has never heard of this before. But the loop you're proposing will look like this:

```py
for field in self.data:
     if isinstance(self[field], bytes) field not in do_not_convert_these_fields:
         self[field] = self[field].decode('utf8')
```

Because `myfp` is a brand-new field I just invented, it can't possibly be in the hard-coded `do_not_convert_these_fields` list. So it will get converted to text, and there's nothing I (the notional plugin developer) can do about it.

dosoe:
Then is there anything that holds this PR back from being merged? I'm thinking of adding a method like ```append(key,value): if self.key, self.key+=', '+value.decode(...)``` so that we don't need to make lists for all the tags that have multiple values, but that's not really a priority, I would like to first get this PR ready to be merged.

test/test_autotag.py
```diff
@@ -503,9 +505,9 @@ def test_reorder_when_track_numbers_incorrect(self):
         items.append(self.item(u'three', 2))
         items.append(self.item(u'two', 3))
         trackinfo = []
-        trackinfo.append(TrackInfo(u'one', None))
-        trackinfo.append(TrackInfo(u'two', None))
-        trackinfo.append(TrackInfo(u'three', None))
+        trackinfo.append(TrackInfo(title=u'one', track_id=None))
+        trackinfo.append(TrackInfo(title=u'two', track_id=None))
+        trackinfo.append(TrackInfo(title=u'three', track_id=None))
```
sampsyo, line 510: Seems like `track_id=None` may no longer be necessary?

test/test_autotag.py
```diff
@@ -595,7 +597,8 @@ def item(i, length):
         items.append(item(12, 186.45916150485752))
 
         def info(index, title, length):
-            return TrackInfo(title, None, length=length, index=index)
+            return TrackInfo(title=title, track_id=None, length=length,
```
sampsyo, line 600: As above.

test/test_autotag.py
```diff
@@ -749,13 +752,15 @@ def test_albumtype_applied(self):
         self.assertEqual(self.items[1].albumtype, 'album')
 
     def test_album_artist_overrides_empty_track_artist(self):
-        my_info = copy.deepcopy(self.info)
+        # make a deepcopy of self.info
```
sampsyo, line 755: This comment is probably not necessary every time?

test/test_ui.py
```diff
@@ -1136,7 +1137,12 @@ def test_summarize_items(self):
         summary = commands.summarize_items([self.item], False)
         self.assertEqual(summary, u"1 items, F, 4kbps, 10:54, 987.0 B")
 
-        i2 = deepcopy(self.item)
+        # make a copy of self.item
+        i2 = library.Item()
+        i2.bitrate = 4321
+        i2.length = 10 * 60 + 54
+        i2.format = "F"
```
sampsyo, line 1144: Any particular reason why this is not `i2 = self.item.copy()`?

dosoe: I wonder why this hasn't been done long ago, in a docstring of one of the parent classes it explicitly says that ```deepcopy()``` doesn't work on these objects.

beets/autotag/hooks.py
```diff
@@ -138,53 +144,41 @@ def decode(self, codec='utf-8'):
             if isinstance(value, bytes):
                 setattr(self, fld, value.decode(codec, 'ignore'))
 
-        if self.tracks:
+        if 'tracks' in self:
```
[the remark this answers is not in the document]
dosoe: I agree, to me it doesn't make sense to have an album without tracks.

dosoe:
Should I provide the changelog as well? Should the documentation be altered?

sampsyo:
Awesome! A changelog entry would be great. I can't think of anywhere else in the docs where we mention this stuff, so probably nothing else needs to change.

Thanks for all your work on this!

beets/autotag/hooks.py
```diff
@@ -39,8 +39,35 @@
...
+        value = info.tag
+    or  value = info[tag]
+    or  value = info.get(tag)
+    or  value = getattr(info, tag)
+    all raise AttributeError when info doesn't have tag
+    to set a tag value:
+        info.tag = value
+    or  info[tag] = value
+    or  setattr(info, tag, value)
+    """
```
sampsyo, line 55: Here's a simpler docstring.
```suggestion
    """A dictionary that supports attribute ("dot") access, so `d.field`
    is equivalent to `d['field']`.
    """
```

dosoe:
That should be it then.

sampsyo:
Awesome!! Thank you for your careful work on this. This is a long-standing request that will help enable lots of interesting additions in the future. Three cheers!

dosoe:
I'm coming back to this PR: I tried to add new fields, which should be easy now, but I found out we missed an important part in ```beet/autotag/__init__.py apply_metadata```. New PR incoming to sort this out.

Issue #3553, opened 2020-04-16
Use flexible attributes as aunique keys


gregforcey:
### Use case

I would like to be able to use flexible attributes as aunique keys for disambiguation. One use case of this would be when having multiple versions of the same album in different resolutions: 16/44 and 24/96, for example. In this case beets should treat the same album in different resolutions as a different album and thus not add the disambiguators at the end of the album name if this occurs.
### Solution
Something like the following is how it should be implemented:

```
aunique:
    keys: original_year albumartist album album_bitdepth album_samplerate
    disambiguators: albumtype label catalognum albumdisambig releasegroupdisambig
    bracket: '{}'

album_fields:  
      album_bitdepth: |
       total = 0
       for item in items:
           total += item.bitdepth
       return round(total / len(items))
    album_samplerate: |
       total = 0
       for item in items:
           total += item.samplerate
       return round(total / len(items) / 1000)
```

### Alternatives
I [posted a question](https://discourse.beets.io/t/using-custom-fields-as-aunique-keys/1213) in the forums and it was suggested I file a feature request. I don't believe there are any alternatives

### Related issue
https://github.com/beetbox/beets/issues/2678

sampsyo:
Thanks!!

For posterity: I linked in the aforementioned forum thread where the change would have to take place. When constructing a MatchQuery, the aunique template function would need to possibly set the `fast` parameter to `False` for non-fixed attributes.

PR #3576, opened 2020-05-02
library: Use flexible attributes as aunique keys (#3553)

drizzt:


sampsyo:
Nice work; thanks!! Any chance you could add a quick changelog entry describing the fix?

Issue #1130, opened 2014-12-03
fetchart: Extract embedded art


pescuma:
Is it possible to fetchart plugin to, if a file has an embedded cover, just copy it to a cover file instead of downloading?

PR #3577, opened 2020-05-03
fetchart: clean up tmp, add 'embedded' source

pants108:
**de83ffe fetchart: clean up invalid tmp files** 

The remote album art sources write into `NamedTemporaryFile`s in order to check image type and dimensions.  However, when an image fails validation, it doesn't seem to get removed from /tmp.  I was fetching art for a large library with aggressive size filters, and found a bunch of temp album art crud piling up there, so I wanted fetchart to clean up after itself a bit.

This commit introduces a `cleanup_tmp(candidate)` method on sources, which is only called on candidates that fail validation.  The `LocalArtSource` has a no-op implementation, but `RemoteArtSource` simply calls `util.remove(candidate.path)`.

There is probably some more tmp cleanup work that could be done in the future.  If you're importing with `copy` or `link`, for example, any fetched art that passes validation seems to remain in /tmp.  I also noticed that the `ArtResizer` class also seems to leave tmp files lying about.  But this solves at least one problem so I wanted to get some feedback.

**a8c8ae7 fetchart: add embedded source**

I wanted the inverse of the EmbedArt plugin; i.e. extract album art from tags, and save it as a local file.  This plugin introduces the "embedded" source type, which calls `art.extract()` on each member of `album.items()` until a validating image is found.

A new `EmbeddedArtSource`class is introduced, whose `fetch_image` simply calls `art.extract(item=candidate.item)`.  This meant I needed to keep track of each `item`, so I added that optional argument to the `Candidate` constructor.

I also found it easiest to let `art.extract()` create its own tempfile if the `outpath` argument is set to None.  This is mainly because that function internally adds a file extension to the provided `outpath`, and I don't know if there is a 100% safe way of generating a tmpfile prefix and guaranteeing that prefix + ".jpg" doesn't already exist (for example).  Maybe I'm just being too paranoid...

As part of that work, I added `util.temp_path_for` and moved that method out of artresizer, and renamed it slightly (was formerly `temp_file_for`).

Importantly, this also required me to change the signature of `art.extract()` to move the now-optional `outpath` argument to the end.  I think I updated all invocations of this function in the beets repo, but I'm not sure what the policy is on changing method signatures like that... as 3rd party plugins that call the function with positional arguments are probably out there.  I could easily update the PR to keep `art.extract()` and `artresizer.temp_file_for` unchanged if need be!

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -34,6 +34,7 @@ from beets.util import hidden
 import six
 from unidecode import unidecode
 from enum import Enum
+from tempfile import NamedTemporaryFile
 
 
 MAX_FILENAME_LENGTH = 200
@@ -543,6 +544,16 @@ def hardlink(path, dest, replace=False):
         else:
             raise FilesystemError(exc, 'link', (path, dest),
                                   traceback.format_exc())
+
+
+def tmp_path_for(path=None):
+    """Returns a path to a named temporary file with the same file extension as
+    ``path``.  If ``path`` is not provided, a named temporary file with no
+    extension is returned.
+    """
+    ext = os.path.splitext(path or '')[1]
+    with NamedTemporaryFile(suffix=py3_path(ext), delete=False) as f:
+        return bytestring_path(f.name)
 
 
 def unique_path(path):
--- a/beets/art.py
+++ b/beets/art.py
@@ -206,7 +206,7 @@ def extract(log, item, outpath=None):
 
     # Append suffix to outpath / create tmp file with suffix
     if outpath is not None:
-        outpath = bytestring_path(outpath + '.' + ext)
+        outpath = bytestring_path(outpath) + bytestring_path('.' + ext)
     else:
         outpath = tmp_path_for('_.' + ext)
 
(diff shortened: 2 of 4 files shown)


pants108:
I'll take a look at the test failures.

beets/util/__init__.py
```diff
@@ -545,6 +546,16 @@ def hardlink(path, dest, replace=False):
                                   traceback.format_exc())
 
 
+def tmp_path_for(path=None):
```
sampsyo, line 549: Could this just be `path=''` to avoid the need for `path or ''` below?

pants108:
Oh nice, hadn't seen #1130.  Taking a look now, but yeah can definitely split it out into its own PR.  

Should I just force push a single clean commit into this PR or what?

wisp3rwind:
> Oh nice, hadn't seen #1130. Taking a look now, but yeah can definitely split it out into its own PR.
> 
> Should I just force push a single clean commit (the first one) into this PR or what?

Yes, force-pushing to PRs is fine around here :) If it seems easier to you to create two new PRs, that's of course also an option.

pants108:
Going to close this and open 2x.  Thanks!

PR #3579, opened 2020-05-03
fetchart: clean up invalid tmp files

pants108:
The remote album art sources write into `NamedTemporaryFile`s in order to check image type and dimensions.  However, when an image fails validation, it doesn't seem to get removed from /tmp.  I was fetching art for a large library with aggressive size filters, and found a bunch of temp album art crud piling up there, so I wanted fetchart to clean up after itself a bit.

This commit introduces a `cleanup_tmp(candidate)` method on sources, which is only called on candidates that fail validation.  The `LocalArtSource` has a no-op implementation, but `RemoteArtSource` simply calls `util.remove(candidate.path)`.

There is probably some more tmp cleanup work that could be done in the future.  If you're importing with `copy` or `link`, for example, any fetched art that passes validation seems to remain in /tmp.  I also noticed that the `ArtResizer` class also seems to leave tmp files lying about.  But this solves at least one problem so I wanted to get some feedback.

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -210,12 +210,19 @@ class ArtSource(RequestMixin):
     def fetch_image(self, candidate, plugin):
         raise NotImplementedError()
 
+    def cleanup_tmp(self, candidate):
+        raise NotImplementedError()
+
 
 class LocalArtSource(ArtSource):
     IS_LOCAL = True
     LOC_STR = u'local'
 
     def fetch_image(self, candidate, plugin):
+        pass
+
+    def cleanup_tmp(self, candidate):
+        # local art source does not create tmp files
         pass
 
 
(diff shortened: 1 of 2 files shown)


beetsplug/fetchart.py
```diff
@@ -210,6 +210,9 @@ def _candidate(self, **kwargs):
     def fetch_image(self, candidate, plugin):
         raise NotImplementedError()
 
+    def cleanup_tmp(self, candidate):
+        raise NotImplementedError()
```
sampsyo, line 214: How about calling this method `cleanup` and making it `pass` in the base class? Then this will do nothing by default, and other sources can override it only if they need to do something on cleanup.

beetsplug/fetchart.py
```diff
@@ -1017,6 +1031,8 @@ def art_for_album(self, album, paths, local_only=False):
                             u'using {0.LOC_STR} image {1}'.format(
                                 source, util.displayable_path(out.path)))
                         break
+                    # else: remove tmp images created by this invalid candidate
```
sampsyo, line 1034:
```suggestion
                    # Remove temporary files for invalid candidates.
```

beetsplug/fetchart.py
pants108: That's fair.  Done.

beetsplug/fetchart.py
pants108: Done.

pants108:
I put this in changelog under "Fixes" but feel free to suggest a better place for it!

sampsyo:
Excellent; thanks!!

PR #3580, opened 2020-05-03
fetchart: add embedded source for #1130

pants108:
I wanted the inverse of the EmbedArt plugin; i.e. extract album art from tags, and save it as a local file.  This plugin introduces the "embedded" source type, which calls `art.extract()` on each member of `album.items()` until a validating image is found.

A new `EmbeddedArtSource`class is introduced, whose `fetch_image` simply calls `art.extract(item=candidate.item)`.  This meant I needed to keep track of each `item`, so I added that optional argument to the `Candidate` constructor.

I also found it easiest to let `art.extract()` create its own tempfile if the `outpath` argument is set to None.  This is mainly because that function internally adds a file extension to the provided `outpath`, and I don't know if there is a 100% safe way of generating a tmpfile prefix and guaranteeing that prefix + ".jpg" doesn't already exist (for example).  Maybe I'm just being too paranoid...

As part of that work, I added `util.temp_path_for` and moved that method out of artresizer, and renamed it slightly (was formerly `temp_file_for`).

Importantly, this also required me to change the signature of `art.extract()` to move the now-optional `outpath` argument to the end.  I think I updated all invocations of this function in the beets repo, but I'm not sure what the policy is on changing method signatures like that... as 3rd party plugins that call the function with positional arguments are probably out there.  I could easily update the PR to keep `art.extract()` and `artresizer.temp_file_for` unchanged if need be!

### DIFF
--- a/beets/art.py
+++ b/beets/art.py
@@ -24,7 +24,7 @@ import platform
 from tempfile import NamedTemporaryFile
 import os
 
-from beets.util import displayable_path, syspath, bytestring_path, tmp_path_for
+from beets.util import displayable_path, syspath, bytestring_path, tmp_file_for
 from beets.util.artresizer import ArtResizer
 import mediafile
 
@@ -121,7 +121,7 @@ def check_art_similarity(log, item, imagepath, compare_threshold):
     """A boolean indicating if an image is similar to embedded item art.
     """
     with NamedTemporaryFile(delete=True) as f:
-        art = extract(log, item=item, outpath=f.name)
+        art = extract(log, f.name, item)
 
         if art:
             is_windows = platform.system() == "Windows"
@@ -189,8 +189,9 @@ def check_art_similarity(log, item, imagepath, compare_threshold):
     return True
 
 
-def extract(log, item, outpath=None):
-    """Extract art from item into outpath, or into a tmp file if unset.
+def extract(log, outpath, item):
+    """Extract art from item into outpath with an appropriate file extension,
+    or into a tmp file if outpath is unset.
     """
     art = get_art(log, item)
     if not art:
@@ -208,7 +209,7 @@ def extract(log, item, outpath=None):
     if outpath:
         outpath = bytestring_path(outpath) + bytestring_path('.' + ext)
     else:
-        outpath = tmp_path_for('_.' + ext)
+        outpath = tmp_file_for('_.' + ext)
 
     log.info(u'Extracting album art from: {0} to: {1}',
              item, displayable_path(outpath))
@@ -219,7 +220,7 @@ def extract(log, item, outpath=None):
 
 def extract_first(log, outpath, items):
     for item in items:
-        real_path = extract(log, item=item, outpath=outpath)
+        real_path = extract(log, outpath, item)
         if real_path:
             return real_path
 
--- a/beets/art.py
+++ b/beets/art.py
@@ -198,14 +198,15 @@ def extract(log, outpath, item):
         log.info(u'No album art present in {0}, skipping.', item)
         return
 
-    # Discover image file extension
+    # Discover the image's file extension.
     ext = mediafile.image_extension(art)
     if not ext:
         log.warning(u'Unknown image type in {0}.',
                     displayable_path(item.path))
         return
 
-    # Append extension to outpath / create tmp file with extension
+    # Append the extension to destination path, creating a temporary file
+    # if no destination is specified.
     if outpath:
         outpath = bytestring_path(outpath) + bytestring_path('.' + ext)
     else:
--- a/beets/art.py
+++ b/beets/art.py
@@ -24,7 +24,7 @@ import platform
 from tempfile import NamedTemporaryFile
 import os
 
-from beets.util import displayable_path, syspath, bytestring_path
+from beets.util import displayable_path, syspath, bytestring_path, tmp_path_for
 from beets.util.artresizer import ArtResizer
 import mediafile
 
@@ -121,7 +121,7 @@ def check_art_similarity(log, item, imagepath, compare_threshold):
     """A boolean indicating if an image is similar to embedded item art.
     """
     with NamedTemporaryFile(delete=True) as f:
-        art = extract(log, f.name, item)
+        art = extract(log, item=item, outpath=f.name)
 
         if art:
             is_windows = platform.system() == "Windows"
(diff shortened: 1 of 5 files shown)


beets/util/__init__.py
```diff
@@ -545,6 +546,16 @@ def hardlink(path, dest, replace=False):
                                   traceback.format_exc())
 
 
+def tmp_path_for(path=''):
```
sampsyo, line 549: TBH I kind of liked the old name a little more… `tmp_file_for` more clearly suggests that this actually creates a file. Otherwise it might sound more like it computes a path but doesn't touch the filesystem.

beets/art.py
```diff
@@ -189,20 +189,26 @@ def check_art_similarity(log, item, imagepath, compare_threshold):
     return True
 
 
-def extract(log, outpath, item):
+def extract(log, item, outpath=None):
+    """Extract art from item into outpath, or into a tmp file if unset.
```
sampsyo, line 193: Perhaps explain here that `outpath` will get an extension applied automatically.

beetsplug/fetchart.py
```diff
@@ -219,6 +221,17 @@ def fetch_image(self, candidate, plugin):
         pass
 
 
+class EmbeddedArtSource(ArtSource):
```
sampsyo, line 224: A docstring here might be nice.

beetsplug/fetchart.py
```diff
@@ -454,11 +467,11 @@ def get(self, album, plugin, paths):
 
         matches = []
         # can there be more than one releasegroupid per response?
-        for mbid, art in data.get(u'albums', dict()).items():
+        for mbid, art_ in data.get(u'albums', dict()).items():
```
sampsyo, line 470: I think it's fine for `art` to shadow the module name here?

beetsplug/fetchart.py
```diff
@@ -743,6 +756,14 @@ def get(self, album, plugin, paths):
                                       match=Candidate.MATCH_FALLBACK)
 
 
+class Embedded(EmbeddedArtSource):
```
sampsyo, line 759: Any particular reason why `EmbeddedArtSource` and `Embedded` are separate classes? Could they just be one thing? I can't imagine having multiple subclasses of `EmbeddedArtSource` as it stands…

wisp3rwind:
> Importantly, this also required me to change the signature of `art.extract()` to move the now-optional `outpath` argument to the end. I think I updated all invocations of this function in the beets repo, but I'm not sure what the policy is on changing method signatures like that... as 3rd party plugins that call the function with positional arguments are probably out there. I could easily update the PR to keep `art.extract()` and `artresizer.temp_file_for` unchanged if need be!

IMO, we really should preserve this function signature. I couldn't name any plugin using this interface in particular, but as you observe, they probably exist. I don't think there's any real policy in this regard (correct me, @sampsyo), but I think the case here doesn't really warrant breaking compatibility. To me, explicitly specifying `None` as in
``` python
art.extract(log, None, item)
```
looks perfectly fine (even though it is not the interface one might choose if starting from scratch).

beets/util/__init__.py
pants108: Yeah that makes sense.  I was going off the idea that `tmp_file_for` implied some kind of file obj would come back, but I get your mental model is too.

beets/art.py
pants108: Done.

beetsplug/fetchart.py
pants108: I added docstrings to the methods defined in this class.  The other ArtSource classes don't have class docstrings so I didn't add one to Embedded either.  Let me know if you'd like more.

beetsplug/fetchart.py
pants108: Reverted, and added ``# noqa F402`` to the line.

beetsplug/fetchart.py
pants108: Mostly was just cargo culting what I saw in `FileSystem(LocalArtSource)`, as LocalArtSource also had just one class that extended it.  But I agree no good reason for this, so I dropped the `EmbeddedArtSource` class and moved all the functionality into `Embedded(ArtSource)`.

pants108:
@wisp3rwind I reverted the changes to `art.extract()` in the second commit in this PR.

I also have added a couple tests in `tests/test_art.py`.  I'm not sure that I wrote them according to your style exactly, but happy to revise if I can get some better pointers (stubbing out albums with items inside).

I'll take a look at the documentation next.  One main question: do we want `embedded` to be added to the default list of sources?  If so, where?  I have been using `filesystem embedded coverart itunes amazon albumart` in my config, but this would obviously change behavior for any user of the plugin who has not customized their sources.

beets/art.py
```diff
@@ -190,19 +190,26 @@ def check_art_similarity(log, item, imagepath, compare_threshold):
...
+    or into a tmp file if outpath is unset.
+    """
     art = get_art(log, item)
-    outpath = bytestring_path(outpath)
     if not art:
         log.info(u'No album art present in {0}, skipping.', item)
         return
 
-    # Add an extension to the filename.
+    # Discover image file extension
```
sampsyo, line 201:
```suggestion
    # Discover the image's file extension.
```

beets/art.py
```diff
@@ -190,19 +190,26 @@ def check_art_similarity(log, item, imagepath, compare_threshold):
...
     ext = mediafile.image_extension(art)
     if not ext:
         log.warning(u'Unknown image type in {0}.',
                     displayable_path(item.path))
         return
-    outpath += bytestring_path('.' + ext)
+
+    # Append extension to outpath / create tmp file with extension
```
sampsyo, line 208:
```suggestion
    # Append the extension to destination path, creating a temporary file
    # if no destination is specified.
```

We generally like to use complete sentences, with periods, in comments.

beets/util/__init__.py
```diff
@@ -545,6 +546,16 @@ def hardlink(path, dest, replace=False):
                                   traceback.format_exc())
 
 
+def tmp_file_for(path=''):
+    """Returns a path to a named temporary file with the same file extension as
```
sampsyo, line 550:
```suggestion
    """Return a path to a named temporary file with the same file extension as
```

[PEP 257](https://www.python.org/dev/peps/pep-0257/) recommends imperative verbs for all docstrings.

wisp3rwind:
> Nice! The tests look good to me.
> 
> I was wondering too about the source order thing. To be honest, I'm not really sure either way—about ether question (whether we should add it to the default source list, and where to put it if so). I'm slightly more confident about the latter: if we do add it, it probably makes sense for it to go before the network sources (and after the file source). But I'm trying to guess whether people will want that behavior by default—any thoughts on whether people will find this pleasant or unpleasant to have on by default, from you or anyone else who might be following this thread?
> 
> Maybe a quick post on the [Discourse](https://discourse.beets.io) to poll opinions would be a good idea?

No strong opinion from my side either --- I guess everyone who cares about their album art has a custom source list in place anyway, and specified some size threshold, such that changing the default can only deteriorate artwork quality to a limited extent. Querying a few more people on Discource sounds like a good idea, though (I really should start following the forum myself...).

PR #3586, opened 2020-05-10
Fetching parentwork if changed, even if not force

dosoe:
Add a new tag, ```mb_workid_current``` that is checked when performing ```beet parentwork``` to check if some works changed and update the parentworks, updating changed works without having to check all parentworks.

### DIFF
--- a/beetsplug/parentwork.py
+++ b/beetsplug/parentwork.py
@@ -174,16 +174,16 @@ add one at https://musicbrainz.org/recording/{}', item, item.mb_trackid)
 
         hasparent = hasattr(item, 'parentwork')
         workfalse = True
-        if hasattr(item, 'mb_workid_current'):
+        if hasattr(item, 'parentwork_workid_current'):
             workfalse = item.mb_workid_current != item.mb_workid
-        if force or (not hasparent) or workfalse:
+        if force or not hasparent or workfalse:
             try:
                 work_info, work_date = find_parentwork_info(item.mb_workid)
             except musicbrainzngs.musicbrainz.WebServiceError as e:
                 self._log.debug("error fetching work: {}", e)
                 return
             parent_info = self.get_info(item, work_info)
-            parent_info['mb_workid_current'] = item.mb_workid
+            parent_info['parentwork_workid_current'] = item.mb_workid
             if 'parent_composer' in parent_info:
                 self._log.debug("Work fetched: {} - {}",
                                 parent_info['parentwork'],
@@ -207,4 +207,4 @@ add one at https://musicbrainz.org/recording/{}', item, item.mb_trackid)
             item, fields=['parentwork', 'parentwork_disambig',
                           'mb_parentworkid', 'parent_composer',
                           'parent_composer_sort', 'work_date',
-                          'mb_workid_current'])
+                          'parentwork_workid_current'])
--- a/beetsplug/parentwork.py
+++ b/beetsplug/parentwork.py
@@ -173,13 +173,17 @@ add one at https://musicbrainz.org/recording/{}', item, item.mb_trackid)
             return
 
         hasparent = hasattr(item, 'parentwork')
-        if force or not hasparent:
+        workfalse = True
+        if hasattr(item, 'mb_workid_current'):
+            workfalse = item.mb_workid_current != item.mb_workid
+        if force or (not hasparent) or workfalse:
             try:
                 work_info, work_date = find_parentwork_info(item.mb_workid)
             except musicbrainzngs.musicbrainz.WebServiceError as e:
                 self._log.debug("error fetching work: {}", e)
                 return
             parent_info = self.get_info(item, work_info)
+            parent_info['mb_workid_current'] = item.mb_workid
             if 'parent_composer' in parent_info:
                 self._log.debug("Work fetched: {} - {}",
                                 parent_info['parentwork'],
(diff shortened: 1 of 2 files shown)


beetsplug/parentwork.py
```diff
@@ -173,13 +173,17 @@ def find_work(self, item, force):
             return
 
         hasparent = hasattr(item, 'parentwork')
-        if force or not hasparent:
+        workfalse = True
```
sampsyo, line 176: What does this variable represent? The `workfalse` name doesn't evoke a specific condition.

beetsplug/parentwork.py
```diff
@@ -173,13 +173,17 @@ def find_work(self, item, force):
...
+        if hasattr(item, 'mb_workid_current'):
+            workfalse = item.mb_workid_current != item.mb_workid
+        if force or (not hasparent) or workfalse:
```
sampsyo, line 179: I don't think the parentheses are necessary.

beetsplug/parentwork.py
dosoe: ```workfalse``` says if ```mb_workid``` and ```mb_workid_current``` have the same value, i. e. if the parentwork corresponds to the work or if the work has been changed since the parentwork was retrieved.

sampsyo: Sounds good—but can you maybe pick a name that reflects that use? Maybe something like work_changed?

docs/plugins/parentwork.rst
```diff
@@ -27,6 +27,9 @@ This plugin adds five tags:
 - **parent_composer_sort**: The sort name of the parent work composer.
 - **work_date**: The composition date of the work, or the first parent work
   that has a composition date. Format: yyyy-mm-dd.
+- **parentwork_workid_current**: The musicbrainz id of the work as it was when
```
sampsyo, line 30: Capitalize MusicBrainz.

dosoe:
For the changelog: Should I add a new changelog entry, or can I just add the PR to the changelog entry for adding the Parentwork plugin? After all, it's also in 1.50

dosoe:
I'm very confused by the failed test, it fails in files I never touched: ```replaygain.py```, ```test_playlist.py``` and ```test_player.py``` .

docs/plugins/parentwork.rst
```diff
@@ -27,6 +33,9 @@ This plugin adds five tags:
 - **parent_composer_sort**: The sort name of the parent work composer.
 - **work_date**: The composition date of the work, or the first parent work
   that has a composition date. Format: yyyy-mm-dd.
+- **parentwork_workid_current**: The Musicbrainz id of the work as it was when
```
sampsyo, line 36: It's "MusicBrainz," rather than "Musicbrainz." 🤪

docs/plugins/parentwork.rst
```diff
@@ -27,6 +33,9 @@ This plugin adds five tags:
...
+  the parentwork was retrieved. This tag exists only for internal bookkeeping,
+  to keep track of recordings whose work have changed.
```
sampsyo, line 38:
```suggestion
  to keep track of recordings whose works have changed. 
```

docs/plugins/parentwork.rst
```diff
@@ -18,7 +18,13 @@ example, all the movements of a symphony. This plugin aims to solve this
 problem by also fetching the parent work, which would be the whole symphony in
 this example.
 
-This plugin adds five tags:
+It uses the ``mb_workid`` tag to retrieve the *parent work* and stores the
+``mb_workid`` as ``parentwork_workid_current``. When one executes 
+``beet parentwork`` it checks if ``mb_workid`` has changed by comparing it
+with ``parentwork_workid_current``, if it has changed or if ``force`` is
+enabled it fetches the *parent work* again.
```
sampsyo, line 25: Here's a perhaps more direct way to tell the user why this works this way:

> The plugin can detect changes in `mb_workid` so it knows when to re-fetch other metadata, such as `parentwork`. To do this, when it first gets the parent work, it stores it *both* in `mb_parentworkid` *and* in a special bookkeeping field `parentwork_workid_current`. Then, you can later change `mb_parentworkid` (with the `modify` command, for example) and re-run the plugin. It will notice that the two tags no longer match and will re-fetch all the parent work metadata. You can also use the `force` option to re-fetch metadata for _all_ music.

dosoe: Actually, it's not how it works. ```parentwork_workid_current``` is a copy of ```mb_workid```, not a copy of ```mb_parentworkid```. I'm adapting, you tell me if it is fine.

sampsyo: Great; yes, that sounds good!

sampsyo:
Nice work! Thanks for sticking with this. :smiley:

aereaux:
It looks to me like you forgot to change one of the occurrences of `mb_workid_current`.  I get this error when running `beet parentwork`:
```Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/beets/dbcore/db.py", line 465, in __getattr__
    return self[key]
  File "/usr/lib/python3.8/site-packages/beets/dbcore/db.py", line 363, in __getitem__
    raise KeyError(key)
KeyError: 'mb_workid_current'

  File "/usr/bin/beet", line 11, in <module>
  File "/usr/lib/python3.8/site-packages/beets/ui/__init__.py", line 1267, in main
    _raw_main(args)
  File "/usr/lib/python3.8/site-packages/beets/ui/__init__.py", line 1254, in _raw_main
  File "/usr/lib/python3.8/site-packages/beetsplug/parentwork.py", line 92, in func
    changed = self.find_work(item, force_parent)
  File "/usr/lib/python3.8/site-packages/beetsplug/parentwork.py", line 178, in find_work
    work_changed = item.mb_workid_current != item.mb_workid
  File "/usr/lib/python3.8/site-packages/beets/dbcore/db.py", line 467, in __getattr__
    raise AttributeError(u'no such field {0!r}'.format(key))
AttributeError: no such field 'mb_workid_current'```

dosoe:
You're right. PR incoming.

PR #3587, opened 2020-05-10
Correct beet/autotag/__init__.py to adapt to flexible tags

dosoe:
Hi! 
Coming back to https://github.com/beetbox/beets/pull/3568, I tried to add a new tag (as should be much easier now) but I noticed that any changes wouldn't get applied. I eventually found out that the problem was in ```beet/autotag/__init__.py``` in ```apply_metadata```, which gets the metadata it has to apply from a list. As the tags are flexible now, this doesn't work. In this PR, I change it to: 
Do your stuff with special tags (```track_id``` becomes ```mb_trackid``` and similar adaptations)
Take all the remaining tags and apply the changes to those. 

I wonder by the way how the ```scrub``` plugin reacts to tags being flexible, I didn't look into it.

### DIFF
--- a/beets/autotag/__init__.py
+++ b/beets/autotag/__init__.py
@@ -35,6 +35,41 @@ from .match import Recommendation  # noqa
 # Global logger.
 log = logging.getLogger('beets')
 
+# metadata that is already hardcoded
+MISC_FIELDS = {
+    'album': (
+        'va',
+        'releasegroup_id',
+        'artist_id',
+        'album_id',
+        'mediums',
+        'tracks',
+        'year',
+        'month',
+        'day',
+        'artist',
+        'artist_credit',
+        'artist_sort',
+        'data_url'
+    ),
+    'track': (
+        'track_alt',
+        'artist_id',
+        'release_track_id',
+        'medium',
+        'index',
+        'medium_index',
+        'title',
+        'artist_credit',
+        'artist_sort',
+        'artist',
+        'track_id',
+        'medium_total',
+        'data_url',
+        'length'
+    )
+}
+
 
 # Additional utilities for the main interface.
 
@@ -52,23 +87,11 @@ def apply_item_metadata(item, track_info):
     if track_info.data_source:
         item.data_source = track_info.data_source
 
-    misc_fields = ['artist_id',
-                   'release_track_id',
-                   'title',
-                   'artist_credit',
-                   'artist_sort',
-                   'artist',
-                   'track_id',
-                   'data_url',
-                   'length'
-                   ]
-
-    for field in track_info.keys():
-        if field in misc_fields:
+    for field, value in track_info.items():
+        # only overwrite fields that are not already hardcoded
+        if field in MISC_FIELDS['track']:
             continue
-        clobber = field in config['overwrite_null']['track'].as_str_seq()
-        value = getattr(track_info, field)
-        if value is None and not clobber:
+        if value is None:
             continue
         item[field] = value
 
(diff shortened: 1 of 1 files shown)


sampsyo:
Hmm, what do you mean about the `scrub` plugin? It doesn't do anything with database metadata—it's just about on-disk tag manipulation, so it should not be affected.

beets/autotag/__init__.py
```diff
@@ -52,20 +52,25 @@ def apply_item_metadata(item, track_info):
...
-        item.composer_sort = track_info.composer_sort
-    if track_info.arranger is not None:
-        item.arranger = track_info.arranger
-    if track_info.work is not None:
-        item.work = track_info.work
-    if track_info.mb_workid is not None:
-        item.mb_workid = track_info.mb_workid
-    if track_info.work_disambig is not None:
-        item.work_disambig = track_info.work_disambig
+    misc_fields = ['artist_id',
```
sampsyo, line 55: Constants like this should be hoisted out to the module level and given descriptive constant-like (i.e., all caps) names. (I know that we had an old `misc_fields` value that disobeyed this dictum, but there's no time like the present to fix that.)

beets/autotag/__init__.py
```diff
@@ -52,20 +52,25 @@ def apply_item_metadata(item, track_info):
...
+                   'title',
+                   'artist_credit',
+                   'artist_sort',
+                   'artist',
+                   'track_id',
+                   'data_url',
+                   'length'
+                   ]
+
+    for field in track_info.keys():
```
sampsyo, line 66: Try using `for field, value in track_info.items()`.

beets/autotag/__init__.py
```diff
@@ -52,20 +52,25 @@ def apply_item_metadata(item, track_info):
...
+        if field in misc_fields:
```
sampsyo, line 67: A comment here describing why this is necessary would be useful.

beets/autotag/__init__.py
```diff
@@ -52,20 +52,25 @@ def apply_item_metadata(item, track_info):
...
+            continue
+        clobber = field in config['overwrite_null']['track'].as_str_seq()
```
sampsyo, line 69: Unless I'm mistaken, this looks like a change in behavior, right? The old version of `apply_item_metadata` did not have this check. For the time being, let's try to preserve exactly the same behavior as the old, fixed-field-set version.

dosoe:
I did an import of some data and made a ```diff``` of the output of ```beet info -l``` before and after the change and the only difference is the import time (as should be, I made two separate libraries for which I had to import it separately).

dosoe:
[out_old.txt](https://github.com/beetbox/beets/files/4610638/out_old.txt)
[out_updated.txt](https://github.com/beetbox/beets/files/4610639/out_updated.txt)
Here the two outputs

beets/autotag/__init__.py
```diff
@@ -35,6 +35,41 @@
 # Global logger.
 log = logging.getLogger('beets')
 
+# metadata that is already hardcoded
+MISC_FIELDS = {
```
sampsyo, line 39: This now does the *opposite* of what misc_fields used to do but it’s still using the old name. Let’s call this SPECIAL_FIELDS and use this comment:

> Most fields are applied directly using the same name as the field in the AlbumInfo or TrackInfo object. We handle a few fields specially, however, and avoid copying them directly to the object.

beets/autotag/__init__.py
```diff
@@ -52,20 +87,13 @@ def apply_item_metadata(item, track_info):
...
-    if track_info.arranger is not None:
-        item.arranger = track_info.arranger
-    if track_info.work is not None:
-        item.work = track_info.work
-    if track_info.mb_workid is not None:
-        item.mb_workid = track_info.mb_workid
-    if track_info.work_disambig is not None:
-        item.work_disambig = track_info.work_disambig
+    for field, value in track_info.items():
+        # only overwrite fields that are not already hardcoded
```
sampsyo, line 91: I don’t think we need to repeat this comment each time we use the filter mapping.

dosoe:
Oh sorry, these were ```beet info -l``` calls, using the data from the library.

dosoe:
Again, I don't understand this test failing, it's about pep-8 in files I never touched.

sampsyo:
That happens from time to time when flake8 gets updated with new warnings that weren't there before. I'll fix those on master!

sampsyo:
Overall, this is looking good! If you merge with master, those linter issues should be fixed. Thanks for being persistent with this, and for showing off those quick tests!!

Generally speaking, you may have noticed that comments in beets are complete sentences (with a capital letter and a period) that try to express full thoughts—we don't aim for terseness. Would you mind trying to do that to the new comments in this PR? That will be a huge help on my end.

dosoe:
As for the changelog, I think it still falls into the implementation of flexible tags.

sampsyo:
Looks great; thanks!

PR #3588, opened 2020-05-11
Fix: subsonicplaylist wrong song selection

MrNuggelz:
Fixes a bug introduced in #3425, where "MatchQuery" is used leading to wrong songs being added to a playlist.


sampsyo:
Excellent!

PR #3589, opened 2020-05-14
Fetch artists, performers, engineers etc flexibly

dosoe:
An attempt to solve https://github.com/beetbox/beets/issues/1547 by fetching all artists involved in a recording. The artists have tag of the form ```mb type - role (credit)```. For example, on this recording https://musicbrainz.org/recording/38fce51c-b8c9-4369-92d3-c354a8e09fd4 , tags would look like:
```mb vocals - bass vocals (Filippo)``` and ```mb vocals sort - bass vocals (Filippo)``` for the sort name, or ```mb vocals - choir vocals```
If several artists have the same type, role and credit, then the tag will contain a list of artists separated by ```; ```. To avoid cluttering over time, I also change the ```mbsync``` plugin to remove all tags that start with ```mb ``` (with a space) that are not recovered when fetching the data again. This fetches all performers, engineers, arrangers (replacing the ```arranger``` tag). Some type names (like ```orchestra```) have been modified, just for the convenience of having ```orchestra``` instead of ```performing orchestra``` or ```recording engineer``` instead of ```recording```. 

Finding all tags that start with ```mb ``` is easy on singletons, but for albums I don't know how to do it in the way it is implemented in ```mbsync```. @sampsyo could you help me for that please? @bearcatsandor does this correspond to your idea of fetching all metadata? There is still more data, like recording place and date, but that's a start, and once this works, it should be easy to adapt it for more tags.

### CODE REVIEW

**Comment by sampsyo** [comment]
Thanks for getting this started! This is pretty tricky… I think an important thing to focus on here is that this is a pretty specialized use case, and I'm not sure most users will want to have this enabled by default. We can consider it, under the hypothesis that extra tags may not bother anybody as long as they are not written to the file, but we may also want to consider creating some kind of new plugin hook that lets the MusicBrainz data extraction be extended. This way, users could customize the way MB data gets "flattened" into tags.

I'm also uncertain about the idea of using a naming convention to do synchronization—i.e., to "reserve" all tags starting with `mb `. It's certainly worth considering, especially with a more unique prefix that ties fields more obviously to a specific plugin, for example. But maybe it would be better to just use a built-in list of fields?

Finally, I definitely think that beets field names should not contain spaces. Otherwise, it's pretty hard to write things like template strings that use the fields.

**Comment by dosoe** [comment]
A built-in list of fields is not a good idea in my opinion. That would require creating a field for every 200+ instrument of MB, the overwhelming majority of them empty, not even talking about all the credits (in opera, typically), and now that we have flexible tags why not use them? In https://github.com/beetbox/beets/issues/1547#issuecomment-627950802 I interpret it as reserving certain fields for a plugin and singling them out by a prefix seems the easiest way of doing so. 
I wouldn't know how to create hooks to flatten the data. The data on performers seems cut into 3 parts: the type (performer, vocals, orchestra, instrument, sound engineer, recording engineer, balance engineer, producer and many others), the role (name of the instrument, the voice, some additional stuff like ```solo```, ```additional``` or ```invited```) and the credit (usually the name of the role for operas, sometimes some precision on the instrument or other similar stuff). It then has a name, a sort name, an id (sometimes also an id for the instrument) a date range and a few more fields. Of course, not all the fields are populated all the time. It should be possible to make something similar to ```path``` with these bricks, which would be effectively some tag name configuration, but I have very little experience on how to do this. 
I first thought about making a separate plugin and pretty much copy-pasting album_info and track_info (from ```mb.py```) with modifications but that's not great, it duplicates the code needlessly. Maybe put an optional boolean argument in them to fetch them or not?

**Comment by dosoe** [comment]
I can't speak for everyone, but as far as I'm concerned, the use case would just be to use beets to check on the info on a recording, not to use them to name files.

**Comment by dosoe** [comment]
Hi! I changed the code a little. A few quality of life improvements: ```beet mbsync``` will now also show changes on singletons. Another one: for releases with more than 500 tracks, when fetching an album with ```musicbrainzngs``` the simplified version of the album is used, that doesn't contain any works or performers. To get those, these releases need to be fetched track by track, which is much longer. I figure that the people that are interested in this kind of additional data would want to get this information and would be ready to wait a little more for their updates. 
I tied the additional data to an option in mbsync (option ```-I --more_info```, the name is temporary until I find a better one) and all tags populated this way will start with ```mbsync_``` so all tags like this are reserved for mbsync. I also replaced all spaces with underscores.

**Comment by dosoe** [comment]
This also uses two additional optional arguments in ```track_info``` and all functions that are passing over stuff to it: ```more_info``` says if we want to fetch the performers and ```by_rec``` says if the album has more than 500 tracks and the tracks need to be fetched individually. This seems to make the tests fail.

**Comment by dosoe** [comment]
> some kind of new plugin hook

What do you mean? The way it is now is that there is an additional argument in beets core that allows to choose whether one wants to get this additional data that is activated by a plugin, but that means changes in beets core. Do you have something else in mind?

**Comment by sampsyo** [comment]
What I meant there is that we could add a brand-new [event][] that allows plugins to process MusicBrainz responses and extract additional data. This would let the task of MusicBrainz-to-beets data extraction evolve independently of the beets core, which would remain somewhat minimal. If other people in the future come up with other swaths of additional data they'd like to fetch from MusicBrainz, we won't need to add new flags like `even_more_info` and `still_more_info_yet` for each new set of data.

[event]: https://beets.readthedocs.io/en/stable/dev/plugins.html#listen-for-events

**Comment by dosoe** [comment]
> What I meant there is that we could add a brand-new [event](https://beets.readthedocs.io/en/stable/dev/plugins.html#listen-for-events) that allows plugins to process MusicBrainz responses and extract additional data. 

Ah ok! Now I understand. I had a little thought at this, but I'm running into a problem: I can get the response of MB by setting up an event and process the response, but I somehow need the items linked to this response to apply the new tags to it. So I need both the MB response and the item, but those two are created in completely different parts of the code.

**Comment by dosoe** [comment]
For example, I tried to put up an event that gives the MB response: ```'trackdata_recieved'```: 
```
res = musicbrainzngs.get_recording_by_id(trackid, TRACK_INCLUDES)
beets.plugins.send(u'trackdata_recieved', info=res['recording'])
```
Then I fetch the response in ```mbsync```: ```self.register_listener('trackdata_recieved', self.track_performers)``` . This triggers a function ```track_performers```: 
```
def track_performers(self, recording):
        info = autotag.hooks.TrackInfo(...)
        artists = {}
...
        for key in artists:
            info[key] = u'; '.join(artists[key])

        info.decode()
```
Now I want to apply this info to the item it belongs to. How do we get it? I could create an event that triggers when an item is called, but there would be no guaranty that this would be the same item than the one we just got the MB data for. How do these triggers work btw? Does the whole program stop to process it when it triggers? Can ```track_performers``` return data? So far it doesn't seem so, but I'm not sure.

**Comment by sampsyo** [comment]
> Does the whole program stop to process it when it triggers? Can `track_performers` return data? So far it doesn't seem so, but I'm not sure.

It just turns into a normal function call, so yes, the rest of beets (in the current thread) "stops" while the event is handled. And yes, they can return values! :tada: Here's an example from elsewhere in beets:
https://github.com/beetbox/beets/blob/31855a93949a7b2e2636f709c2678d0efef1dcba/beets/ui/commands.py#L865-L866

So a sensible thing to do here might be to just return a `dict` with all the information extracted. Then the `mb` module can take that stuff and assign it onto the `*Info` object, which will later be applied to the `Items.

**Comment by dosoe** [comment]
Thanks! With the end of the lockdown and the start of the final phase of my PhD, I won't be able to work on this too much in the next weeks, but I'm keeping an eye on it.

**Comment by Holzhaus** [comment]
When writing performer information to files, does this use the same tag mapping that is used by Picard? We shouldn't roll our own tags, because then support by media players will be nonexistent. Picard's tags do have some support, at least they are shown correctly by mpd/ncmpcpp.

**Comment by dosoe** [comment]
It doesn't write anything to files so far. It's all only in the beets library.

**Comment by dosoe** [comment]
Hi! I'm coming to it again. Can something like ```performer_info = self.register_listener('trackdata_recieved', track_performers)``` (```track_performers``` and the ```'trackdata_recieved'``` being suitably defined) work? So far it returns ```None```, from looking into ```beets.plugin.register_listener``` it doesn't seem like it can return anything. Is there a way to get result equivalent to it (i. e. get the output of ```track_performers``` in a variable somewhere)?

**Comment by sampsyo** [comment]
Yes, that should work! When the "core beets" code calls `plugins.send(...)`, the return value consists of all the return values of all the listeners to the given event. So you should be able to collect the return values as in the example from https://github.com/beetbox/beets/pull/3589#issuecomment-649028921.

**Comment by dosoe** [comment]
Ok. I tried to do it, but ```self.register_listener('trackdata_recieved', track_performers)``` returns ```None``` eventhough ```track_performers``` has a return value.

**Comment by dosoe** [comment]
More details: 
```
def track_performers(info):
    artists = {}
...
    return artists
class MBSyncPlugin(BeetsPlugin):
    def __init__(self):
        super(MBSyncPlugin, self).__init__()
...
    def singletons(self, lib, query, move, pretend, write, more_info):
        """Retrieve and apply info from the autotagger for items matched by
        query.
        """
        for item in lib.items(query + [u'singleton:true']):
            item_formatted = format(item)
...
            # Get the MusicBrainz recording info.
            if more_info:
                performer_info = self.register_listener('trackdata_recieved', track_performers)
```
and the event: 
```
def track_for_id(releaseid):
    """Fetches a track by its MusicBrainz ID. Returns a TrackInfo object
    or None if no track is found. May raise a MusicBrainzAPIError.
    """
    trackid = _parse_id(releaseid)
    if not trackid:
        log.debug(u'Invalid MBID ({0}).', releaseid)
        return
    try:
        res = musicbrainzngs.get_recording_by_id(trackid, TRACK_INCLUDES)
        beets.plugins.send(u'trackdata_recieved', info=res['recording'])
```
It turns out that ```track_performers``` recieves ```info``` well, but ```performer_info``` ends up being ```None```

**Comment by sampsyo** [comment]
It's not `self.register_listener` that produces a return value—it's the "other end" of the event, `beets.plugins.send`. So whatever you return from the event handler function (`track_performers`) will be available to the sender. (What were you hoping to get returned from the registration call?)

**Comment by dosoe** [comment]
Well, what I was hoping is that I could obtain the return value of ```track_performers``` with the input given by the event.

**Comment by sampsyo** [comment]
But the call to `register_listener` happens at the wrong time. You call that method to register the plugin to receive events from beets, usually when you're setting up the plugin. But `track_performers` is not called then, so there's nothing to return from it. `track_performers` only gets called when the event actually happens.

**Comment by dosoe** [comment]
So is there a way to get the stuff the listener throws out to use it to tag tracks? I hope what I want to do is clear, do you know how I could achieve it?

**Comment by Holzhaus** [comment]
A shame that his went stale. I didn't follow development progress, is there another attempt to integrate musicbrainz performer support into beets?

**Comment by dosoe** [comment]
Hi! This is stale because I had other stuff to do, I'm working on it again.

**Comment by dosoe** [comment]
Hi @sampsyo , Now that I looked after the testing, I would like to look back on this one. So, if I understand well what has been discussed before, the way events work in beets is: 
-we create an ```event``` with ```beets.plugins.send('event',info='data')```
-we listen to this ```event``` in a plugin with ```self.register_listener('event',function)```
If ```event``` happens, then the listener will notice and will apply ```function``` using ```data``` as argument. Any return value of ```function``` will be the return value of ```beets.plugins.send('event',info='data')```. Is that clear? Is that correct? I fit is, then it seems I need to completely rework where I make my ```send``` and ```register_listener``` calls. 
By the way, this PR has been closed, could it be possible to reopen it?

**Comment by sampsyo** [comment]
Yep, I think you got it!

**Comment by dosoe** [comment]
Hi! I reworked the way I do my calls and I have a working prototype for ```track_info```. Is that fine for you, can I go over to doing the same to ```album_info```? 
It works fine on my laptop, I don't know why the tests aren't running.

**Comment by dosoe** [comment]
Is there a way to activate the listener only if an option is activated?

**Comment by sampsyo** [comment]
Sure, you can choose to do `register_listener` conditionally on a config option.

**Comment by sampsyo** [comment]
(I unfortunately will not have time to review the code for a few days. But other folks are welcome to weigh in!)

**Comment by dosoe** [comment]
This seems to be working just fine. Should I adapt the tests? Should I go over to the documentation? There's still the issue of how to mark flexible fields created by a plugin: https://discourse.beets.io/t/tag-naming-conventions/1626 and I'm not sure how I can use an option in ```__init__```. Else everything seems to be working fine. 

There is another issue, but that's for another PR: for releases with more than 500 tracks, MB doesn't give the performer or work info, no matter how we ask. The solution would be to get this info track by track, which is of course very long and only for those that really want it.

**Comment by dosoe** [comment]
Ultimately, one might also want a finer control: maybe only get performers, not recirding engineers/copyright holders etc., but again that's for another PR.

File: docs/changelog.rst [hunk 1]
```diff
@@ -173,6 +173,8 @@ New features:
 * :doc:`/plugins/replaygain` now does its analysis in parallel when using
   the ``command``, ``ffmpeg`` or ``bs1770gain`` backends.
   :bug:`3478`
+* The ``mbsync`` plugin now also fetches performers, if the ``performer_info``
+  option is enabled. Fixes :bug`1547`, thanks to :user:`dosoe`.
```
Holzhaus, line 177:
```suggestion
  option is enabled. Fixes :bug:`1547`, thanks to :user:`dosoe`.
```

dosoe: thanks for your vigilance!

dosoe:
Another thing to do would be to write these tags (or some of them) to the files using the picard tag mapping.

docs/changelog.rst
dosoe: Sorry I didn't merge your suggestion, I'm still struggling with git.

dosoe:
Any further comments or can this be merged? There is a lot to do with flexible, plugin-specific tags (like options to wipe all tags from a specific plugin or decide on a way to single them out). I go with this prefix method, I asked the best way to do it on the forum but didn't get any answer. I would like to wrap it up for now and merge, and maybe others will get new ideas.

sampsyo:
I almost hate to ask this, but is there any chance we could review a PR for the plugin hook separately (and first) from the plugin changes that use it? It would be great to think about those carefully by themselves, and that might be what I have the mental capacity to pay attention to. 😇

dosoe:
No problem. I will do it.

dosoe:
Now that https://github.com/beetbox/beets/pull/3831 has been merged, this can move on. I tried to start a discussion on tag naming here https://discourse.beets.io/t/tag-naming-conventions/1626 , now that tags are flexible and can potentially be conflicting, but that hasn't gained any traction. 
I mentioned previously using the tag mapping of picard, but as far as I know the tag mapping part of beets has moved to another repo, is that true? 
I can either try to do this as well in this PR or I can do it in a subsequent PR, whatever is more convenient.

aereaux:
Thanks for working on this.  For myself at least, this would be useful even if the tags are not written out to files, as I would like to use the tags in the file paths.

sampsyo:
@dosoe, just to keep this one in play, any chance you could summarize where this PR now stands? It looks like most of the new code is mbsync—is that the main thing that's now changed?

dosoe:
Well @sampsyo fetching and storing the performers works, the documentation is made, I don't know if there is anything to add to the tests. The PR meaningfully changes only the mbsync plugin (there is one quick change to the core, but nothing big). 
Two questions remain for me: 

- I'm adding new tags flexibly, these are characterized by the ```mbsync_``` prefix, I've asked on the discourse if this is the right way to characterize flexible tags created by plugins (we need something, else different plugins could conflict pretty easily)

- I'm not writing the new tags to files, which I could do for some of them (picard has a tag for conductor and a general tag for performer, so it's possible) but I would like to postpone it and get this merged, also I seem to remember that the bits relevant to writing tags and tag mapping are not in this repo anymore (I might be wrong though). However, if you would like me to do it I can give it a try.

docs/changelog.rst
```diff
@@ -224,7 +224,8 @@ Other new things:
   Thanks to :user:`arogl`.  :bug:`3906`
 * Get ISRC identifiers from musicbrainz
   Thanks to :user:`aereaux`.
-* :doc:`/plugins/metasync`: The ``metasync`` plugin now also fetches the ``Date Added`` field from iTunes databases and stores it in the``itunes_dateadded`` field.Thanks to :user:`sandersantema`.
+* :doc:`/plugins/metasync`: The ``metasync`` plugin now also fetches the ``Date Added`` field from iTunes databases and stores it in the``itunes_dateadded`` field.Thanks to :user:`sandersantema`.* The ``mbsync`` plugin now also fetches performers, if the ``performer_info``
```
sampsyo, line 227: Looks like a missing line break?

docs/changelog.rst
```diff
@@ -224,7 +224,8 @@ Other new things:
...
+  option is enabled. Fixes :bug:`1547`, thanks to :user:`dosoe`.
```
sampsyo, line 228: I'm not sure this does actually fix #1547… that one seems like more of a blanket issue for fetching all kinds of other metadata, and it's not really about mbsync?

docs/changelog.rst
sampsyo: Plugin-specific changelog entries are usually formatted like this:

```rest
:doc:`/plugins/mbsync`: Stuff about a change to `mbsync`.
```

docs/plugins/mbsync.rst
```diff
@@ -36,3 +36,8 @@ The command has a few command-line options:
 * To customize the output of unrecognized items, use the ``-f``
   (``--format``) option. The default output is ``format_item`` or
   ``format_album`` for items and albums, respectively.
+* To also get performer data from MusicBrainz, use the ``-P``
+  (``--performer_info``) option. This will add all track and album performers
+  as additional tags (recognisable by the ``mbsync_`` prefix). To autmmatically
```
sampsyo, line 41: Please check spelling in here.

beetsplug/mbsync.py
```diff
@@ -134,6 +229,12 @@ def albums(self, lib, query, move, pretend, write):
             # work for albums that have missing or extra tracks.
             mapping = {}
             for item in items:
+                # Clean up obsolete flexible fields
```
sampsyo, line 232: I admit I don't quite understand why this is necessary… any chance it could have a slightly more detailed explanation? Why do these tags need to be removed?

beetsplug/mbsync.py
```diff
@@ -134,6 +229,12 @@ def albums(self, lib, query, move, pretend, write):
...
+                if performer_info:
+                    for tag in item:
+                        if tag[:6] == 'mbsync' and (tag not in track_info or
```
sampsyo, line 235: It's a bit simpler to do `tag.startswith('mbsync_')`.

beetsplug/mbsync.py
```diff
@@ -86,13 +176,18 @@ def singletons(self, lib, query, move, pretend, write):
...
-
+            # Clean up obsolete flexible fields
+            if performer_info:
+                for tag in item:
+                    if tag[:6] == 'mbsync' and tag not in track_info:
+                        del item[tag]
             # Apply.
             with lib.transaction():
                 autotag.apply_item_metadata(item, track_info)
+                ui.show_model_changes(item)
```
sampsyo, line 187: This seems unrelated to the logic change at hand—any chance this was left over from debugging?

beetsplug/mbsync.py
```diff
@@ -27,10 +26,95 @@
 MBID_REGEX = r"(\d|\w){8}-(\d|\w){4}-(\d|\w){4}-(\d|\w){4}-(\d|\w){12}"
 
 
+def track_performers(info):
```
sampsyo, line 29: It would be really great to have detailed docstrings for this and `album_performers`. These are pretty complicated functions, so it's not exactly obvious what role they play.

beets/autotag/mb.py
```diff
@@ -66,6 +66,7 @@ def get_message(self):
             self._reasonstr(), self.verb, repr(self.query)
         )
 
+
```
sampsyo, line 69: Seems like some whitespace changes snuck in by accident. (One more below.)

sampsyo (commented):
OK! I guess the main thing I'm still looking for here is a succinct description of exactly what users should expect from the new feature. The docs now say:

> This will add all track and album performers as additional tags (recognisable by the ``mbsync_`` prefix).

Maybe it would be useful to say exactly which tags are added? I also feel like I may be forgetting something I suggested myself, but is there a reason why there is an `mbsync_` prefix on these fields? (I guess specifically: why are there different tag names from the original metadata fetcher?)

dosoe:
> Maybe it would be useful to say exactly which tags are added?

That's the thing, the tags are created flexibly, you don't want to create a tag for every instrument on MusicBrainz, not even speaking about all the roles in all the operas. I look at all the performers and their roles and create adapted tags on the fly. That's why I use this prefix: the tags are created on the fly. If other plugins do the same (and why shouldn't they), they could conflict, so we need something to prevent conflicts and the way I propose is to create a prefix for tags created for plugins (or at least flexible tags). There could be another way though, that's what I'm asking for in https://discourse.beets.io/t/tag-naming-conventions/1626 .

dosoe:
> I guess the main thing I'm still looking for here is a succinct description of exactly what users should expect from the new feature.

You mean here, or in the docs and changelog?

docs/changelog.rst
dosoe: This issue isn't about mbsync, but this PR fixes it as it fetches most of the extra fields from MusicBrainz (with the exception of performance dates and places).

beetsplug/mbsync.py
dosoe: This is because the tags are attributed 'on the fly'. If a performer is tagged as 'Percussions' for example, he will be tagged as 'mbsync_Percussions'. If then someone on MB corrects it to 'Cymbals' mbsync will add a tag 'mbsync_Cymbals' but will not remove the now obsolete 'mbsync_Percussions' tag, unless I remove all tags that start with 'mbsync_' and are not in the new tags. That's why I need this cleanup step.

beetsplug/mbsync.py
dosoe: It is not left from debugging, it is just an unrelated issue: I noticed that mbsync shows the changes for albums but not for singletons, this corrects it.

beetsplug/mbsync.py
dosoe: Done, is that enough?

dosoe:
For what you should expect, it looks like this: 
```
The Beatles - The Beatles - Long Tall Sally
  mbsync_engineer: Norman Smith -> Norman Smith; Norman Smith
  mbsync_engineer_sort: Smith, Norman -> Smith, Norman; Smith, Norman
  mbsync_instrument_-_electric_guitar: George Harrison; John Lennon -> 
  mbsync_instrument_sort_-_electric_guitar: Harrison, George; Lennon, John -> 
  mbsync_instrument_-_piano: George Martin -> 
  mbsync_instrument_sort_-_piano: Martin, George -> 
  mbsync_producer: George Martin -> George Martin; George Martin
  mbsync_producer_sort: Martin, George -> Martin, George; Martin, George
  mbsync_instrument_sort_-_electric_guitar_(rhythm_guitar): Lennon, John
  mbsync_instrument_-_electric_guitar_(lead_guitar): George Harrison
  mbsync_instrument_-_upright_piano: George Martin
  mbsync_engineer_sort_-_assistant: Langham, Richard; Emerick, Geoff
  mbsync_engineer_-_assistant: Richard Langham; Geoff Emerick
  mbsync_instrument_-_electric_guitar_(rhythm_guitar): John Lennon
  mbsync_instrument_sort_-_upright_piano: Martin, George
  mbsync_instrument_sort_-_electric_guitar_(lead_guitar): Harrison, George
```

aereaux:
I'm trying out this pull request, but I can't seem to get it to work.  `beet mbsync -pP Harmonielehre` does not give any output.  I would expect it to output some changed tags for the people involved.  I have this release: https://beta.musicbrainz.org/release/d19c62ec-8821-443a-b535-adf8d5dc0023

dosoe:
I checked it, I was mixing up the parameter name of the plugins.send calls, I corrected it, now it should work (it works for me).

beets/autotag/mb.py
```diff
@@ -270,7 +270,7 @@ def track_info(recording, index=None, medium=None, medium_index=None,
         info.arranger = u', '.join(arranger)
 
     # Supplementary fields provided by plugins
-    extra_trackdatas = plugins.send('mb_track_extract', data=recording)
+    extra_trackdatas = plugins.send('mb_track_extract', info=recording)
```
Holzhaus, line 273: Minor nit: data is already plural (singular is "datum"). Maybe use "datasets"?

dosoe: done!

sampsyo:
> That's the thing, the tags are created flexibly, you don't want to create a tag for every instrument on MusicBrainz, not even speaking about all the roles in all the operas.

Oh yes, I am 100% on board with using flexible attributes and not "real" on-disk tags!! What I was wondering about is why they have an `mbsync_` prefix. And in fact, I'm just not entirely sure why the `mbsync` behavior is any different from the original data fetch… isn't the idea with `mbsync` to "freshen" data, but to match the information so it looks exactly the same as if it had just been imported? With this PR, does `mbsync` do something different than the "original" fetch?

> You mean here, or in the docs and changelog?

Both, really! I'm (somewhat embarrassingly) having trouble understanding myself, so that means I'm interested in understanding right here in the thread, but it would also be nice to have a complete description of how the feature works in the docs for other users.

dosoe:
> Both, really!

So, basically when beets fetches data from MusicBrainz it iakes only a few selected tags (title, tracknumber, mb_trackid, composer, artist and a others). It doesn't take any performers. This PR takes the performer data from MusicBrainz, takes all the performers, fetches their roles/instruments, creates appropriate tags and fills them with all the performers. 

> What I was wondering about is why they have an `mbsync_` prefix.

If we go the road of flexible tags, every plugin could create/modify its own tags. This opens the possibility that different plugins uses the same tag for different things, basically creating use conflicts. So far, this problem was averted by the fact that there was a finite number of tags, they were all in a file and in the documentation so you could check a tag to see if it was already used by something else. With flexible tags that are created on the fly, there is no such list anymore, so there is no way of checking if a given tag is already in use by some other plugin. We therefore need a way to control tag names and ordering them by plugin with the use of a prefix is the solution I'm coming up with. Granted, in this specific case I don't think there is a risk of mixing up tags, but I think for the long-term health of beets and to avoid future conflicts it is necessary to create something of this type for flexible tags.

sampsyo:
OK, but I guess the fundamental thing I still don't get is why these tags are associated with `mbsync` in the first place. Wouldn't they want to be fetched during an initial import too? (Fundamentally, I wouldn't expect `mbsync` to fetch any tags that are not also fetched by the initial import. Does that make sense?)

dosoe:
You still need the ```mbsync``` plugin enabled as well as the corresponding option. 
I think they would be fetched during an initial import (if I understand events well enough) but keys are fetched at an initial import as well by ```keyfinder``` given the right options and nobody would argue against the plugin being in charge of fetching the key. 
I can change the prefix, but given the behaviour of flexible tags I would still argue for giving it some sort of prefix.

dosoe:
eventhough I find prefixes like this to be pretty ugly

sampsyo:
I'm sorry for still being so confused, but is there any chance you can break down in more detail why mbsync needs to be involved—or why the fields themselves are associated with mbsync? Is it the case that you get `mbsync_x` fields fetched even when the `mbsync` plugin is disabled? (Again, with apologies, I remain pretty confused here!)

aereaux:
Thanks for the fix, but now I get this error:
```
  File "/home/user/.local/bin//beet", line 23, in <module>
    beets.ui.main()
  File "/home/user/.local/lib/beets/beets/ui/__init__.py", line 1291, in main
    _raw_main(args)
  File "/home/user/.local/lib/beets/beets/ui/__init__.py", line 1278, in _raw_main
  File "/home/user/.local/lib/beets/beetsplug/mbsync.py", line 164, in func
    self.albums(lib, query, move, pretend, write, performer_info)
  File "/home/user/.local/lib/beets/beetsplug/mbsync.py", line 222, in albums
    album_info = hooks.album_for_mbid(a.mb_albumid)
  File "/home/user/.local/lib/beets/beets/autotag/hooks.py", line 558, in album_for_mbid
    album = mb.album_for_id(release_id)
  File "/home/user/.local/lib/beets/beets/autotag/mb.py", line 571, in album_for_id
    return album_info(res['release'])
  File "/home/user/.local/lib/beets/beets/autotag/mb.py", line 361, in album_info
    ti = track_info(
  File "/home/user/.local/lib/beets/beets/autotag/mb.py", line 273, in track_info
    extra_trackdatasets = plugins.send('mb_track_extract', info=recording)
  File "/home/user/.local/lib/beets/beets/plugins.py", line 498, in send
    result = handler(**arguments)
  File "/home/user/.local/lib/beets/beets/plugins.py", line 153, in wrapper
    return func(*args, **kwargs)
TypeError: import_listener() missing 1 required positional argument: 'data'
```

dosoe:
@aereaux That is strange, I don't have this problem. Are you sure you use the latest version?

dosoe:
However, one problem I do have, is that I need the ```performer_info``` option active in the config file for ```beets mbsync -P``` to work, which defeats the purpose of having this option. Is there a way to solve this? Can listeners be activated by an option in the command, or are they only by options in the config file?

aereaux:
Ah, looks like it's my fault.  I have a plugin that was causing the issue.  No idea why it's suddenly failing, though

beets/autotag/mb.py
```diff
@@ -270,8 +270,8 @@ def track_info(recording, index=None, medium=None, medium_index=None,
         info.arranger = u', '.join(arranger)
 
     # Supplementary fields provided by plugins
-    extra_trackdatas = plugins.send('mb_track_extract', data=recording)
-    for extra_trackdata in extra_trackdatas:
+    extra_trackdatasets = plugins.send('mb_track_extract', info=recording)
```
aereaux, line 273: Isn't the parameter called data in the docs?

https://github.com/beetbox/beets/blame/master/docs/dev/plugins.rst#L255

poo922:
Man I don't know my baby mama she always gets this b******* from GitHub and
all this answer this and answer this cheating and f****** around you know
what f*** y'all I've tried to unsubscribe y'all kiss my ass cuz y'all ain't
s*** y'all are chicken s*** have to cover everything hey honey got no balls
so y'all f*** off get the f*** off my email

beets/autotag/mb.py
dosoe: yes, sorry I screwed up

dosoe: it should be fine now

dosoe:
> Is it the case that you get `mbsync_x` fields fetched even when the `mbsync` plugin is disabled?

I did a test, they are not fetched if mbsync is disabled, even if the 'performer_info' option is active. 
Why am I associationg it with mbsync? mbsync goes over all items to fetch info from MusicBrainz and the performer info I want to fetch is from MusicBrainz, so it seems an obvious match. Also, to update the performer information, I need to go over all items and fetch their MB data, which mbsync already does. I don't think there is such high demand for this that it should be in beets core.

dosoe:
> Ah, looks like it's my fault. I have a plugin that was causing the issue. No idea why it's suddenly failing, though

Does it work now, with your plugin?

aereaux:
> > Ah, looks like it's my fault. I have a plugin that was causing the issue. No idea why it's suddenly failing, though
> 
> Does it work now, with your plugin?

Yep, changing the parameter name back fixed it.

Also, at least as a user, I think of mbsync as only updating the same information that is retrieved on initial import.  It seems like it would make more sense either to have this in core so that it gets retrieved on initial import and on mbsync (probably behind a feature flag) or to put it in a separate plugin like how parentwork works, although parentwork does do a lot of additional info fetching.

dosoe:
Actually, it is retrieved at the initial import if mbsync and the 'performer_info' option are active. That's probably the feature flag you're looking for.

aereaux:
> Actually, it is retrieved at the initial import if mbsync and the 'performer_info' option are active. That's probably the feature flag you're looking for.

Ah OK, I wasn't sure because it seemed like it required using both that parameter and the commandline flag when using mbsync.  But it seems like it works now, I might have done something wrong when I tried it before.

bearcatsandor:
I think I missed something. If this has been merged into master, what config options do I need to enable to try it out?

dosoe:
@bearcatsandor Yes, you need to enable the mbsync plugin and in the mbsync plusing you need to enable the fetch_performers option 
```
mbsync: 
    fetch_performers:yes
```
There is a command-line option that should override this option, but it doesn't work, I will probably remove it and performer fetching will be only possible by enabling the option in the config, unless you believe there could be a need/demand for it (something like ```beet mbsync -P``` to fetch the performers regardless of the option in the config file).

bearcatsandor:
So, unless I'm being  a doofus, I've placed that mbsync option into my config (the plugin was already enabled), and it doesn't appear to do anything.  Looking at this page, I'm not sure how to tell if it's been merged into beet:master.

sampsyo:
Hey @bearcatsandor—no, I think you're understanding correctly! This PR has not been merged: you can tell because the top of the page still says "open" in green. #3995 is an example of a merged (purple!) PR, meaning the changes are now part of the branch they were proposed for. To try out these changes, you'll need to check out @dosoe's branch.

sampsyo:
> I did a test, they are not fetched if mbsync is disabled, even if the 'performer_info' option is active.
> Why am I associationg it with mbsync? mbsync goes over all items to fetch info from MusicBrainz and the performer info I want to fetch is from MusicBrainz, so it seems an obvious match. Also, to update the performer information, I need to go over all items and fetch their MB data, which mbsync already does. I don't think there is such high demand for this that it should be in beets core.

Thanks for clarifying, @dosoe! This does help clarify things a bit for me.

Nonetheless, it does solidify my position that this probably _should not_ be part of mbsync. The original intention behind the mbsync plugin was to provide simple, fast alternative to re-importing your music—to just refresh exactly the same metadata that the original import process also provides. Like the documentation says:

> This is useful for updating tags as they are fixed in the MusicBrainz database, or when you change your mind about some config options that change how tags are written to files. 

That's the use case it's aiming for: when you know something has changed on MB (perhaps because you edited MB yourself!) and you'd like for your database to reflect that. Notably, it doesn't fetch any _more_ metadata than the original import process does—it just fetches the same stuff only "fresher."

From that perspective, it seems like fetching performer information would, in an ideal world, want to obey the same rules. It could get fetched on initial import and refreshed on `mbsync`. But there would be no need for it to be `mbsync`-specific, i.e., people who just want performer data shouldn't need to enable `mbsync` at all.

So can I make the recommendation that this be put into a separate, special-purpose plugin, perhaps called `performers` or similar? In v1.0, it could just provide an explicit `beet performers` command that fetches and stores all this data. The tags could be called `performer_*` to make the purpose clear. In a future revision, we could explore hooking this the importer and into `mbsync`, but that can come after an initial "explicit" invocation. Would this make sense?

dosoe:
It does make sense. However, to be honest, I'm not sure how to get the performer info without getting all the other MusicBrainz tags as well, as they get fetched during the call of ```hooks.track_for_mbid(item.mb_trackid)``` which fetches the data from MB. I could then just apply the tags related to performers, but that seems kind of a waste to have two plugins that do the same thing (because of the listener, each time ```mbsync``` is called, the performers would be updated). Maybe that's a problem with the way I implemented the listener, let me explain why:

If you create a new flexible tag, let's call it ```instrument_instr``` where ```instr``` is the name of the instrument. This tag gets added to ```track_info``` using the listener. Now someone changes the instrument. Then the listener will add the new tag, but not remove the old, now obsolete tag. This has to be done by hand, which is why I added a little loop that cleans all the tags with the ```mbsync_``` prefix in this PR. However, now I need to put this little loop not only in the ```mbsync``` plugin, but in each and every plugin that uses the listener and adds tags using it. Now if I have an ```instrument``` plugin that adds this instrument tag, I need to make a loop in ```mbsync``` that cleans up the instruments tag, else in case of a change it would add the new instrument and not remove the old one. Which means that the ```mb_track_extract``` event should allow to add tags, but also to remove some. I hope I'm clear, I don't know if it's a huge issue. Maybe a run of the ```instrument``` plugin would clean it up anyway, but that means that to update cleanly my data, I would need to call ```mbsync```, ```instruments``` and eventually ```performers``` which all make exactly the same calls to MB, which seems like a waste of calls and can take some time for my collection of 50k tracks.

sampsyo:
Yes, very clear; thanks for the explanation!! What makes this effort unique relative to other "apply some extra fields" efforts is the need to create a data-driven set of fields, rather than a fixed list, which implies the need to delete as well as add them.

What do you think about adding a new event just for doing this? We could emit the event in the `apply_item_metadata` function or thereabouts, to let plugins do whatever they want just before setting new fields on Items. The new plugin could just delete all the `performer_*` tags at that point and then rely on the normal process to set the new ones. Would that work (and be as simple as it sounds)?

(If that does work, it would be good to do one thorough check to see if some other, existing event is close enough to the right "time" that it world already service this purpose...)

aereaux:
One other possibility that I can think of, although I don't know how complicated it would be, would be to have some official way to associate specific tags with specific plugins.  This might help resolve any potential conflicts between plugins and also allow for beets to automatically delete the old versions of tags.

bearcatsandor:
If I can help keep this alive, let me know.  I've been eager for this for years.

dosoe:
Hi! Yes, that's greatly appreciated. I just can't find the time to do it, sorry. 
So that's where I think this is: Right now, I have two events ```mb_track_extract``` and ```mb_album_extract``` , I have two functions that get the performers ```track_performers``` and ```album_performers``` and I need to go through all tracks and albums and download musicbrainz data to feed it to these functions and then something to apply it to the files. Right now, I'm piggybacking on the ```mbsync``` plugin but from what I understand, it should change, maybe create a new plugin specifically for this. 
There is a catch however: The performer tags are created flexibly, which means that when they change, they are not replaced like-for-like, it can happen that new ones are created and the old ones just stay. 
Flexible tags also open the possibility for conflicts, if two plugins use the same tag for different purposes. Previously, all tags were explicitly listed, so the risk was low, you could just look up if the tag was already used for something else. Now, tags can be created on the fly, which means that there is no way to know in advance which tags will be used by which plugin. The solution I opted here was to create a prefix for all tags created by ```mbsync```. 
To clean up old potentially obsolete tags, I remove all tags with this prefix before applying the new ones, however, by using an event, all other plugins using triggering this event and applying tags would not make this cleanup (so far there are none, but if we create a new pluging for performers there will be: ```mbsync``` and this new plugin). 
I hope I'm clear and helpful, else just tell me.

PR #3591, opened 2020-05-15
Update genres.txt

thenktor:
Added more heavy metal genres: https://en.wikipedia.org/wiki/Heavy_metal_genres


sampsyo:
Seems cool! Could you add a quick changelog entry describing this and #3592? Even better if it were one PR, but I won't insist. :smiley:

thenktor:
Hopefully I got this right with this Github online editing feature...

PR #3592, opened 2020-05-15
Update genres-tree.yaml

thenktor:
Added more heavy metal genres: https://en.wikipedia.org/wiki/Heavy_metal_genres

### DIFF
--- a/beetsplug/lastgenre/genres-tree.yaml
+++ b/beetsplug/lastgenre/genres-tree.yaml
@@ -648,35 +648,51 @@
     - glam rock
     - hard rock
     - heavy metal:
-        - alternative metal
+        - alternative metal:
+            - funk metal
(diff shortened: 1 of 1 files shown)


beetsplug/lastgenre/genres-tree.yaml
```diff
@@ -719,6 +735,7 @@
             - street punk
             - thrashcore
         - horror punk
+        - oi!
```
sampsyo, line 738: Just checking: is this really a genre?

thenktor: See here: https://en.wikipedia.org/wiki/Oi!

sampsyo: Awesome and fascinating!

PR #3594, opened 2020-05-16
Scrape new Genius song page html

stlutz:
As noted in #3535, Genius now doesn't always produce html pages that the existing code can scrape. While the fix in #3554 stops the lyrics plugin from completely crashing, it now simply ignores these pages, even though they do contain the desired lyrics. I added a few lines to the algorithm to deal with this new layout.

While I was at it, I also removed the indirection over the /song api, so we only need to query Genius twice per song instead of thrice.

Another change was to include the artist in the search query sent to Genius. This produces much better search results for songs with very common names but less known arists.

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -359,14 +359,9 @@ class Genius(Backend):
             'User-Agent': USER_AGENT,
         }
 
-    def lyrics_from_song_api_path(self, song_api_path):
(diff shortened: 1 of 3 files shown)


sampsyo:
Woohoo; looks awesome! Would you mind adding a quick changelog entry describing how this works now?

sampsyo:
Awesome; thanks!!

PR #3598, opened 2020-05-22
fix: use `endwith()` instead of `endWith()`

jef:
### Description

Fixes method call error created by this: https://github.com/beetbox/beets/pull/3449

Moved the helper static functions to the `SubsonicUpdate` class and made them private static.

Added tests for various of URL input and made sure that it can create a token and format URL correctly.

### DIFF
--- a/beetsplug/subsonicupdate.py
+++ b/beetsplug/subsonicupdate.py
@@ -35,47 +35,6 @@ from beets.plugins import BeetsPlugin
 __author__ = 'https://github.com/maffo999'
 
 
-def create_token():
-    """Create salt and token from given password.
-
-    :return: The generated salt and hashed token
-    """
-    password = config['subsonic']['pass'].as_str()
-
-    # Pick the random sequence and salt the password
-    r = string.ascii_letters + string.digits
-    salt = "".join([random.choice(r) for _ in range(6)])
-    salted_password = password + salt
-    token = hashlib.md5().update(salted_password.encode('utf-8')).hexdigest()
-
-    # Put together the payload of the request to the server and the URL
-    return salt, token
-
-
-def format_url():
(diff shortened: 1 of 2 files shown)


jef:
Working on creating some tests, so we don't run into this problem again.

sampsyo:
So far, so good. :smiley:

beetsplug/subsonicupdate.py
```diff
@@ -90,10 +49,51 @@ def __init__(self):
...
+
+        :return: The generated salt and hashed token
+        """
+        password = config['subsonic']['pass'].as_str()
+
+        # Pick the random sequence and salt the password
+        r = string.ascii_letters + string.digits
+        salt = "".join([random.choice(r) for _ in range(6)])
+        salted_password = password + salt
+        token = hashlib.md5(salted_password.encode('utf-8')).hexdigest()
```
jef, line 64: Also changed this, but hard to tell in GitHub diff...

Used to be `hashlib.md5().update(salted_password.encode('utf-8')).hexdigest()`. Seems like a proper use of `hashlib`.

beetsplug/subsonicupdate.py
```diff
@@ -90,10 +49,51 @@ def __init__(self):
...
+    def __format_url():
+        """Get the Subsonic URL to trigger a scan. Uses either the url
+        config option or the deprecated host, port, and context_path config
+        options together.
+
+        :return: Endpoint for updating Subsonic
+        """
+
+        url = config['subsonic']['url'].as_str()
+        if url and url.endswith('/'):
```
jef, line 79: This was the main problem.

sampsyo:
Nice; looks prefect to me! Thanks!!

jef:
Thanks as always!

PR #3600, opened 2020-05-23
add drop2beets to community plugins

martinkirch:
Hello,
This is one more bullet in the list ; I hope the presentation says it all.

Thanks for beets !


sampsyo:
Wonderful; thanks!

PR #3602, opened 2020-05-24
feat: added github actions

jef:
### Description

You can see the actions success and time over [here](https://github.com/jef/beets/actions/runs/115199962).

Not sure if this was wanted, but I wanted to see how GitHub Actions compared with AppVeyor and Travis CI for this project.

Only side-effect right now is that it's only for Linux builds. It's a near exact replica of Travis CI.

I believe we should probably update tox to be more relevant to the build process and have the pipeline utilize what's in the `tox.ini` file.

Could be way off base here, but would love some feedback. We can move away from this too.


tox.ini
```diff
@@ -4,10 +4,9 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py27-test, py37-test, py27-flake8, docs
-
-# The exhaustive list of environments is:
-# envlist = py{27,34,35}-{test,cov}, py{27,34,35}-flake8, docs
+envlist =
+    py{27,35,36,37,38}-{test,cov,flake8}
+    docs
```
sampsyo, line 9: The reason we had this pared down to a smaller list is because it controls what happens by default when you just type `tox` at the command line. It's pretty annoying to have this launch all 5*3=15 versions in the common case, so we narrowed this down to a smaller list that is good for most development. The shorter list does not prevent CI from using whatever envs it wants.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,77 @@
...
+          python -m pip install --upgrade pip
+          python -m pip install tox sphinx
+          python -m pip install codecov || true
+      - name: Test with tox
+        run: |
+          sudo ln -s /dev/null /dev/raw1394
+          tox -e py27-test
+          tox -e py27-cov
+          codecov || true
+  build3x:
```
sampsyo, line 31: Just for clarity, maybe it would be slightly easier to read if these were, like, "build-py3x" and "build-py27"?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,77 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ 5, 6, 7, 8 ]
```
sampsyo, line 35: I understand why this variable is just one character (so we can interpolate it both for `python-version` below when doing `setup-python`, which wants "3.5," and as the tox environment, which wants "py35"), but it's a tiny bit unfortunate because it makes the CI output a little harder to read. Probably fine though!

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Install base dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install tox sphinx
+      - name: Test with tox
+        run: |
+          sudo ln -s /dev/null /dev/raw1394
+          if [[ ${{ matrix.python-version }} == '6' ]]; then
+              tox -e py3${{ matrix.python-version }}-test
+              tox -e py3${{ matrix.python-version }}-flake8
```
sampsyo, line 60: Instead of doing flake8 here as part of the build, what do you think about doing that as a separate action? I did that for another project there:
https://github.com/sampsyo/wideq/blob/64553d52e38d094a9d362a6527286376a9d30246/.github/workflows/style.yml#L6-L10

And it's _very_ cool because it puts the linting errors inline, visible right in the PR change view.

jef: Sure, I think that's great!

.github/workflows/ci.yaml
jef: It is unfortunate. Maybe we can work around this a different way.

jef:
Thanks for all the feedback!

> I also think a big advantage of switching to Actions would be that it has Windows support

I probably played around for about 30 min with the Windows side of things. I don't know exactly the whole process around beets building and the Windows wasn't doing that great with it. Well at least for me.

With the suggestions you've given, I don't really know what we want to accomplish. Like using flake8 (linting), tests, and building.

It seems that we want to do all of it on all the versions from your comments. So I think a refactor of the tox in general is good!

If you have any other suggestions, I don't mind trying to update the tox! Or feel free too.

I'll go ahead and put this back into draft for now and then we can finalize maybe later this week.

.github/workflows/ci.yaml
jef: Sure -- no skin off my back.

sampsyo:
Cool; I think it's fine to leave Windows off at first—we can get this into working order and merge it, then come back to the Windows side of things.

I don't 100% follow what you're suggesting needs to happen to the Tox setup—can you elaborate on what needs to happen there?

jef:
I suppose I'm lacking in information on what needs to happen with tox.

Like this for example:

> Instead of doing flake8 here as part of the build, what do you think about doing that as a separate action

What else we can pull out to make part of the build? I know sphinx also has some actions around that. As well as codecov.

I imagine that we want to run coverage on all the versions as well as coverage, but I don't think we are currently doing that. Is that assumption correct?

tox.ini
jef: Ah, that makes sense. So all these environment's on this list are automatically ran when you run `tox` I assume?

sampsyo:
Aha, I see! I think maybe part of the confusing here is that tox serves two purposes: both for CI and for running on your own computer, during development. So I would like, for example, to keep the flake8 environments around in the tox config, even if we don't use them in CI. And there's no harm in keeping the docs environments around either, even if we replace them with special-purpose Sphinx GitHub actions for CI.

I couldn't quite parse this:

> I imagine that we want to run coverage on all the versions as well as coverage

I actually don't know that we need coverage for all versions—it may make sense to just report coverage for a single run. It might just be confusing if there are multiple, slightly different coverage percentages per commit.

jef:
> It might just be confusing if there are multiple, slightly different coverage percentages per commit.

I suppose that makes sense. Does it make sense to run on latest then?

beets/autotag/hooks.py
```diff
@@ -332,7 +332,7 @@ def __init__(self):
         self._penalties = {}
 
     @LazyClassProperty
-    def _weights(cls):  # noqa
+    def _weights(self, cls):  # noqa
```
sampsyo, line 335: I don't think this is correct—the decorator only provides the class (like `@classmethod` or similar).

beetsplug/export.py
```diff
@@ -21,7 +21,7 @@
 import codecs
 import json
 import csv
-import xml.etree.ElementTree as ET
+import xml.etree.ElementTree as ElementTree
```
sampsyo, line 24: I actually kinda liked the `ET` name for `ElementTree`, which is very long, even if a pedantic name style checker thinks it looks like a constant. Maybe we should just call it `et` or something, which at least looks like a module name.

If we stick with this, though, the more standard way to write it would be:

```py
from xml.etree import ElementTree
```

beetsplug/smartplaylist.py
```diff
@@ -105,17 +105,17 @@ def build_queries(self):
 
             playlist_data = (playlist['name'],)
             try:
-                for key, Model in (('query', Item), ('album_query', Album)):
+                for key, model in (('query', Item), ('album_query', Album)):
```
sampsyo, line 108: The reason this was capitalized is that it's a class. I get that a naming style checker wouldn't like this because it's also a local variable. If we want to satisfy such a checker, I think we should call it `model_cls` to make it clear what's going on here.

beetsplug/metasync/amarok.py
```diff
@@ -49,7 +49,7 @@ class Amarok(MetaSource):
         'amarok_lastplayed':  DateType(),
     }
 
-    queryXML = u'<query version="1.0"> \
+    query_xml = u'<query version="1.0"> \
```
sampsyo, line 52: No objection here; this is obviously better!

tox.ini
sampsyo: Yep!

sampsyo:
> Does it make sense to run on latest then?

Yeah, sounds about right to me!

beets/autotag/hooks.py
jef: Not completely sure, flake8 made the suggestion.

beetsplug/export.py
jef: Flake8 also made this suggestion. Not sure if there is an explicit ignore comment we can make, but fine doing so if there is.

beetsplug/smartplaylist.py
jef: `clazz` seems to be a standard around certain parts. What do you think about that? `model_clazz`

jef:
image

Strange that github actions flake8 isn't picking it up. Maybe the `path` key isn't used how I read it to be.

.github/workflows/ci.yaml
jef: I looked through other actions as well. They didn't seem to include flake8 plugins like tox does. Let me know if you think we should work around this or leave as is. Maybe I missed something.

jef:
Alright, I think it's ready just for linux right now. I suppose we can reiterate for Windows soon enough!

.github/workflows/ci.yaml
sampsyo: Aha, got it. That's too bad! In the longer term, maybe we should fork one of these (such as [this one](https://github.com/TrueBrain/actions-flake8)) and add support for plugins.

sampsyo:
Argh, AppVeyor is now erroring out on Python 2 with this message:

> ERROR: Package 'setuptools' requires a different Python: 2.7.17 not in '>=3.5'

We can't drop py2 support soon enough, I guess.

beetsplug/smartplaylist.py
sampsyo: The reason `clazz` is a thing is for when you want to use `class` alone as a name, but Python reserves the word `class` by itself, so you can't. In any case, however, the convention in Python (and elsewhere in beets) is `cls` (see, for example, the [docs](https://docs.python.org/3/library/functions.html#classmethod) for the `@classmethod` decorator). So let's stick with `model_cls`.

tox.ini
```diff
@@ -4,54 +4,64 @@
...
 [tox]
-envlist = py27-test, py37-test, py27-flake8, docs
+envlist =
+    py27-{test}
+    py38-{test,cov,flake8}
+    docs
 
-# The exhaustive list of environments is:
-# envlist = py{27,34,35}-{test,cov}, py{27,34,35}-flake8, docs
+# exhaustive list
```
sampsyo, line 12: I kinda liked the complete sentence:
```suggestion
# The exhaustive list of environments is:
```

tox.ini
```diff
@@ -4,54 +4,64 @@
...
+passenv =
+    LANG # avoids output errors
+    NOSE_SHOW_SKIPPED # Undocumented feature of nose-show-skipped.
+deps =
+    {test,cov}: {[_test]deps}
+    flake8: {[_flake8]deps}
+    py27: pathlib
+commands =
+    py27-test: python -m nose {posargs}
+    py38-cov: python -bb -m nose --with-coverage {posargs}
```
sampsyo, line 30: Should this be `py3{4,5,6,7,8}-cov`? (No need to prevent other versions from supporting coverage, even if we don't use them.)

tox.ini
```diff
@@ -4,54 +4,64 @@
...
+    py38-flake8: flake8 {posargs} {[_flake8]files}
```
sampsyo, line 31: As above.

tox.ini
jef: I'll update these tonight!

tox.ini
jef: I thought the idea was that we would have the minimal defaults up here. Maybe I misunderstood. Do you mind making a suggestion and I'll commit it?

tox.ini
```diff
@@ -4,54 +4,64 @@
...
+    py3{4,5,6,7,8}-test: python -bb -m nose {posargs}
```
jef, line 32:
```suggestion
commands =
    py27-cov: python -m nose --with-coverage {posargs}
    py27-flake8: flake8 {posargs} {[_flake8]files}
    py27-test: python -m nose {posargs}
    py3{4,5,6,7,8}-cov: python -bb -m nose --with-coverage {posargs}
    py3{4,5,6,7,8}-flake8: flake8 {posargs} {[_flake8]files}
    py3{4,5,6,7,8}-test: python -bb -m nose {posargs}
```

.github/workflows/ci.yaml
jef: I agree, that'd be fun.

jef:
> Argh, AppVeyor is now erroring out on Python 2 with this message:
> 
> > ERROR: Package 'setuptools' requires a different Python: 2.7.17 not in '>=3.5'
> 
> We can't drop py2 support soon enough, I guess.

Not sure why. I didn't remove anything from AppVeyor or change tox's deps (too much).

sampsyo:
Hmm… there's definitely a chance that it's *not* related to any of the tox changes here and was some sort of dependency change, but it's really hard to tell. And TBH I'm at a loss for why it's only appearing on AppVeyor.

This seems to be the underlying problem:
https://github.com/pypa/setuptools/issues/1964#issuecomment-574761497

But I don't quite understand why the wrong version of setuptools is being installed in the first place.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,93 @@
...
+          tox -e py3${{ matrix.python-version }}-cov
+          tox -e py3${{ matrix.python-version }}-flake8
+      - name: Test with tox
+        if: matrix.python-version != '8'
+        run: tox -e py3${{ matrix.python-version }}-test
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python to build docs with sphinx
```
sampsyo, line 84:
```suggestion
      - name: Set up Python to build docs with Sphinx
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,93 @@
...
+    runs-on: ${{ matrix.platform }}
+    strategy:
+      matrix:
+        platform: [ ubuntu-latest ]
+    env:
+      PY_COLOR: 1
+      NOSE_SHOW_SKIPPED: 1
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up python 2.7
```
sampsyo, line 14:
```suggestion
      - name: Set up Python 2.7
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,93 @@
...
+    strategy:
+      matrix:
+        platform: [ ubuntu-latest ]
+        python-version: [ 5, 6, 7, 8 ]
+    env:
+      PY_COLOR: 1
+      NOSE_SHOW_SKIPPED: 1
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up python 3.${{ matrix.python-version }}
```
sampsyo, line 49:
```suggestion
      - name: Set up Python 3.${{ matrix.python-version }}
```

jef:
I'd prefer the latter, but I also don't know how to solve the Encoding problems I'm having there.

sampsyo:
Hmm, since we're _not_ seeing the AppVeyor problem on a more recent commit on master:
https://ci.appveyor.com/project/beetbox/beets/builds/33195829/job/nt7npa84vu48ei4k

I think we can conclude that there *is* something wrong with the tox.ini changes. I'm going to try reproducing the problem on a branch.

sampsyo:
My best guess at this point is that it's the `sphinx` dependency that was added to `[_test]deps`. (That doesn't seem necessary to me?)

But overall, I think a good idea for this PR might be to leave tox.ini alone—I know there's some stuff to clean up there, but we can leave that to a future pull request. :smiley: I also think the reason you saw all the flake8 naming errors was because tox.ini removed a version pin on `pep8-naming`. Again, that seems worth addressing, but one thing at a time—maybe let's just do the GitHub actions stuff for now?

Also, about the `gh-actions:env` section in `tox.ini`: that is only for Windows, right? (And would require [a plugin](https://pypi.org/project/tox-gh-actions/) we're not currently using?) If so, let's leave that off for now too.

Thanks for all your help with this!

jef:
> I think we can conclude that there _is_ something wrong with the tox.ini changes.

:sob: 

> My best guess at this point is that it's the `sphinx` dependency that was added to `[_test]deps`. (That doesn't seem necessary to me?)

Builds weren't succeeding without it in the GitHub Actions environment, but that could be that I changed the tox.ini quite a bit.

I kind of forget where I left off before all the tox changes, so yeah, lets pull back the tox changes and go from there.

> Also, about the `gh-actions:env` section in `tox.ini`: that is only for Windows, right? (And would require [a plugin](https://pypi.org/project/tox-gh-actions/) we're not currently using?) If so, let's leave that off for now too.

You're right. That was leftover from some testing I was doing. Must've missed that one. Thanks!

> Thanks for all your help with this!

You bet, fun to tinker around.

sampsyo:
OK, cool! Let's see what happens with an unmodified `tox.ini`. If the job is really failing without the Sphinx dependency, I'd like to figure out why—my best guess is that `sdist`-building the package requires building the manpages with Sphinx, but maybe that's best done *outside* of tox's virtualenv.

jef:
The `--min-version` on the flake8 options may have failing tests. Not sure if this is expected output:

```
FI18 __future__ import "annotations" missing
```

I tried in the past to put `annotations` import everywhere at one point, but that just made more errors. I also tried to ignore that flake8 rule, but started to warn on other imports. Not sure what the expected output is here. Sorry :disappointed:

jef:
Works with appveyor with default tox! That's nice. Let's see if I put back windows into the works...

jef:
I wonder if we should integrate codecov action since we have the plugin to upload the coverage. That way we get reports inline. Kind of what you talked about with the flake8. or maybe use coveralls or something. some sort of tool -- could be cool.

sampsyo:
Awesome!! Thanks for working on this!

To run down what I see as the outstanding or recently resolved issues:

- Encoding problem on Windows: Super weird. It looks like the failure is coming from tox itself, not from beets at all! (And at first glance, it looks like yet another consequence of the questionable decision in Python 3 to try to pretend that all OS interfaces are Unicode…) But it looks like you've disabled Windows on Actions for now, right? We can try to address this in a separate PR.
- Warning from flake8-future-import about `from __future__ import annotations`: it looks like this went away? Although I'm not entirely sure why… did something change? (The root cause is the configuration in `setup.cfg`. We can add one more `FI` error to the `ignore=` line.)
- The requirement for `self` as an argument to the `_weights` method: this also seems to be resolved, somehow?
- Removed `--min-version` to flake8: Seems fine! I actually don't know why we needed this flag in the first place… that option is also set globally here: https://github.com/beetbox/beets/blob/master/setup.cfg#L6

Also, about coverage tools: FWIW, here's [our codecov page](https://codecov.io/gh/beetbox/beets). Inline reports could be cool! I long ago disabled the automatic bot comment feature, where CodeCov would post a comment on any pull request saying how the coverage changed—it was mostly not very helpful, especially for PRs that aren't really relevant to tests such as documentation, and generated extra email. But maybe the special status report thing would be less noisy?

Anyway, this is looking great!! Unless any of the above things still have open questions, I do agree it's likely time to merge. :tada: :smiley: :rocket:

jef:
> It looks like the failure is coming from tox itself, not from beets at all!

Wow... really? I couldn't tell. I'll have to go back and look. That's frustrating.

> Warning from flake8-future-import about from __future__ import annotations: it looks like this went away? Although I'm not entirely sure why… did something change?

I removed the `--min-version` and that seemed to do it.

> The requirement for self as an argument to the _weights method: this also seems to be resolved, somehow?

This still seems [like a problem](https://github.com/jef/beets/runs/737343460?check_suite_focus=true#step:7:226) to me :(

I think after resolving that. We should merge! Or actually, before merging. Do you think I should toss Travis CI config?

sampsyo:
> Wow... really? I couldn't tell. I'll have to go back and look. That's frustrating.

Yeah, the traceback ends like this:

```
  File "c:\hostedtoolcache\windows\python\2.7.18\x64\lib\site-packages\tox\package\builder\__init__.py", line 7, in build_package
    return make_sdist(config, session)
  File "c:\hostedtoolcache\windows\python\2.7.18\x64\lib\site-packages\tox\package\builder\legacy.py", line 39, in make_sdist
    returnout=True,
  File "c:\hostedtoolcache\windows\python\2.7.18\x64\lib\site-packages\tox\action.py", line 118, in popen
    lines = out_path.read_text("UTF-8").split("\n")
  File "c:\hostedtoolcache\windows\python\2.7.18\x64\lib\site-packages\py\_path\common.py", line 171, in read_text
    return f.read()
  File "c:\hostedtoolcache\windows\python\2.7.18\x64\lib\codecs.py", line 314, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf8' codec can't decode byte 0x92 in position 30004: invalid start byte
```

So it's tox, calling [py.path](https://pypi.org/project/py/), which does a read-file-and-decode thing. I wish it were clearer which file it was reading!

> This still seems [like a problem](https://github.com/jef/beets/runs/737343460?check_suite_focus=true#step:7:226) to me :(

Aha, yes, I missed that! The real mystery to me is why the `# noqa` comment, which is supposed to disable *all* flake8 errors on that line, isn't doing its thing:
https://github.com/beetbox/beets/blob/a907dac16c685a013686cc10a6e092a7ca21ddcd/beets/autotag/hooks.py#L335

And about that, I'm pretty stumped for the time being. Maybe the syntax for "noqa" markers got updated in some recent version of flake8? Pretty mysterious…

jef:
> Aha, yes, I missed that! The real mystery to me is why the `# noqa` comment, which is supposed to disable _all_ flake8 errors on that line, isn't doing its thing:

Makes sense. Seems like there is a [new way to do this](https://flake8.pycqa.org/en/latest/user/violations.html#in-line-ignoring-errors)?

Once this gets fixed. I think we should be good to go.

sampsyo:
Cool, seems like a “targeted” noqa with a specific error code is worth a shot?

jef:
Alright. I remembered that I removed the limitation of `pep8-naming` to get latest. Not sure why it was pinned. Went to go look at the versions and seemed fine to me to update.

It worked without the extension on `# noqa`, but I included it anyway to be explicit.

I think we're ready!

sampsyo:
Awesome; thanks!! I'll review in more detail in the morning. But FWIW, I think the reason it was pinned was to avoid changes in its policy—namely, the ones that led to the `ElementTree` -> `ET` name becoming an error. But that doesn't seem too important!

jef:
Added `codecov` for the coverage upload. I think that was the only thing missing. I also believe notifications weren't working for IRC. I haven't it post in there yet, so I currently don't have it posting.

Other than that, it should be a direct replacement!

sampsyo:
Awesome! I think we can ignore the IRC thing—I have also stopped using IRC (in favor of Discourse and Gitter), and I don't think those notifications added very much even when we did use it more heavily.

sampsyo:
Merged! I also replaced the README badge. :tada: Thank you for all your work getting this going!!!

Here are a few items to explore in the future:

- Windows! (And disable AppVeyor.)
- Do we really need `hashFiles('**/requirements.txt')` in our cache step? We don't have a `requirements.txt`, and we're not using one to install tox and Sphinx.
- Explore separating out the flake8 check so we get inline annotations.
- Similarly, explore a [custom action for Sphinx](https://github.com/ammaraskar/sphinx-action) that can put errors inline.

jef:
> Do we really need hashFiles('**/requirements.txt') in our cache step? We don't have a requirements.txt, and we're not using one to install tox and Sphinx.

Good point, that was just a copy and paste and I forgot about bothering to update 😆 

Love all the custom actions. Those would be fun to write up. I'll make some tickets for them.

jtpavlock:
@jef Thanks for your work on this! Any chance you'd be interested in joining the beets team?

jef:
You bet, it was good to do. @jtpavlock That sounds great! I'd love to.

jtpavlock:
Awsome! I sent you an invite to the team chat and @sampsyo should get you an invite to the org

PR #3605, opened 2020-05-29
Fix `list -af` example in documentation

ybnd:
Didn't work since $tracktotal is not an Album field

image


sampsyo:
This is better; thanks!

PR #3607, opened 2020-05-31
importer: use tarfile.open to handle compressed archives

### LINKED ISSUE #3606: import: support extraction of bzip2 and gzip tarballs
### Use case

I'm trying to use beets to import albums from a set of bzip2/gzip tarballs, i.e. .tar.bz2 and .tar.gz files. Would be nice if we could just import them instead of having to extract them first.

### Solution

I see that the `tarfile` module can process tar archives with [gzip or bzip2 compression](https://docs.python.org/2/library/tarfile.html), so hopefully this is not too hard to do. Unfortunately I don't know Python so I cannot provide a patch.

### Alternatives

As a workaround I am extracting the tarballs first. Not the end of the world but it would be nice to be able to just point beets to the files, just like we do with rar and zip files.

Thanks for maintaining this great piece of software!

cota:
Call tarfile.open instead of tarfile.TarFile from the importer so that
we can import compressed tar archives.

Note that tarfile.TarFile does not handle compressed archives:
$ python3
Python 3.8.2 (default, Apr 27 2020, 15:53:34)
[GCC 9.3.0] on linux
>>> import tarfile
>>> tf = tarfile.TarFile("Lagrimas.tar.bz2")
[...]
tarfile.ReadError: invalid header
>>>

But tarfile.open does deal with them:
$ python3
Python 3.8.2 (default, Apr 27 2020, 15:53:34)
[GCC 9.3.0] on linux
>>> import tarfile
>>> tf = tarfile.open("Lagrimas.tar.bz2")
>>>

Tested:
$ ls Lagrimas/*.mp3 | wc -l
11
$ tar cjf Lagrimas.tar.bz2 Lagrimas/

- Before:
$ beet import Lagrimas.tar.bz2
extraction failed: invalid header
No files imported from /tmp/Lagrimas.tar.bz2

- After:
$ beet import Lagrimas.tar.bz2
[works]

Fixes #3606.

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -1034,8 +1034,8 @@ class ArchiveImportTask(SentinelImportTask):
             cls._handlers = []
             from zipfile import is_zipfile, ZipFile
             cls._handlers.append((is_zipfile, ZipFile))
-            from tarfile import is_tarfile, TarFile
-            cls._handlers.append((is_tarfile, TarFile))
+            import tarfile
+            cls._handlers.append((tarfile.is_tarfile, tarfile.open))
             try:
                 from rarfile import is_rarfile, RarFile
             except ImportError:
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -124,6 +124,8 @@ New features:
   separators to support path queries.
   Thanks to :user:`nmeum`.
   :bug:`3567`
+* ``beet import`` now handles tar archives with bzip2 or gzip compression.
+  :bug:`3606`
 
 Fixes:
 


beets/importer.py
```diff
@@ -1034,8 +1034,8 @@ def handlers(cls):
             cls._handlers = []
             from zipfile import is_zipfile, ZipFile
             cls._handlers.append((is_zipfile, ZipFile))
-            from tarfile import is_tarfile, TarFile
-            cls._handlers.append((is_tarfile, TarFile))
+            from tarfile import is_tarfile, open
+            cls._handlers.append((is_tarfile, open))
```
sampsyo, line 1038: To avoid a potentially confusing collision with the built-in `open` function in this scope:

```suggestion
            import tarfile
            cls._handlers.append((tarfile.is_tarfile, tarfile.open))
```

cota: Done, thanks.

sampsyo:
Perfect; thanks!!

Issue #3609, opened 2020-06-02
Playing albums or playlists via web


darakian:
### Use case

I'm trying to use beets as a music server with the webUI being the primary interface

### Solution
Extend the web interface to search on albums/artists/etc and to either continue playing on the current selection or to build playlists. For my use case playing albums is sufficient.

### Alternatives
I'm not aware of any, but maybe they exist.

sampsyo:
Hi there! This sounds cool but the web thingy is really meant as a bare-bones demo rather than a full-featured music player. I think the right medium-term solution here is to stop shipping our own player and instead rely on [AURA](https://github.com/beetbox/aura) to let innovation in the UI happen separately.

See also #718 and #738 for a vision for more full-fledged players.

darakian:
Oh cool. I'll check into those. For what it's worth I really like the minimalism of the built in web player. Either way I'll keep an eye on things. Thanks for the work :)

PR #3610, opened 2020-06-03
Enhance web ui search

darakian:
This is a pretty basic PR which adds three buttons for new search modes.
The basic idea is to prepend `artist:` or `album:` to the query string for the user.
An `all` button is added to avoid confusion around a general search.
While a manual search is possible this makes a qualified search more natural.

### DIFF
--- a/beetsplug/web/static/beets.css
+++ b/beetsplug/web/static/beets.css
@@ -53,6 +53,11 @@ body {
     text-align: center;
     margin: 0.25em 0;
 }
+#selectorButtons {
+  display: block;
+  text-align: center;
+  margin: 0.25em 0;
+}
 #query {
     width: 95%;
     font-size: 1em;
@@ -60,7 +65,7 @@ body {
 #entities ul {
     width: 17em;
 
-    position: fixed;
+    position: relative;
     top: 36px;
     left: 0;
     bottom: 0;
--- a/beetsplug/web/static/beets.js
+++ b/beetsplug/web/static/beets.js
@@ -226,10 +226,21 @@ var AppView = Backbone.View.extend({
     el: $('body'),
     events: {
         'submit #queryForm': 'querySubmit',
+        'click #albumSearch': 'queryAlbum',
+        'click #artistSearch': 'queryArtist',
+        'click #allSearch': 'querySubmit',
     },
     querySubmit: function(ev) {
         ev.preventDefault();
         router.navigate('item/query/' + encodeURIComponent($('#query').val()), true);
+    },
+    queryAlbum: function(ev) {
+        ev.preventDefault();
+        router.navigate('item/query/album:' + encodeURIComponent($('#query').val()), true);
+    },
+    queryArtist: function(ev) {
+        ev.preventDefault();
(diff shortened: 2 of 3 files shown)


darakian:
Related issue
https://github.com/beetbox/beets/issues/3609

darakian:
The visuals via Firefox on macos
image

sampsyo:
Hmm—I see what you're going for here, but I also kind of like the transparency of the current model where what you're typing is just a plain old beets query. It also seems a little odd that `foo bar` would run the query `artist:foo bar`, i.e., match `foo` in the artist field but `bar` in any field.

I do think there are probably interesting ways to think about querying by artist and album, but I'm not sure query prefixes are the right one. We probably want to do something like the equivalent of `beet ls -a` when searching by album, for example.

darakian:
> Hmm—I see what you're going for here, but I also kind of like the transparency of the current model where what you're typing is just a plain old beets query. It also seems a little odd that `foo bar` would run the query `artist:foo bar`, i.e., match `foo` in the artist field but `bar` in any field.

That's fair. That could be fixed to query on artist'foo bar', but I get the sense that you're not a fan on the buttons either way. I agree that it's a bit ugly for what it's worth :)

> We probably want to do something like the equivalent of beet ls -a when searching by album, for example.

That would be ideal I think, but I'm not sure how to map that to the UI. The list of results from the query seemed to only work with tracks when I was testing things. I could extend the results listing to handle album entries as well, but I see a few issues there
1. In a query do we display the contents of the album as well as an entry for an album directly? If not how does the user restrict the search?
2. when the user clicks on an album what happens? Does the set of results update with the contents of the album? Does the whole album play? Does a new UI element pop up to display the contents?

sampsyo:
Yeah, I indeed think that's the hard part—the UI for albums might want to be quite different from the UI for individual tracks! I don't have any great answers here, but I continue to be excited to unleash this kind of creative thinking by supporting alternative web UIs…

darakian:
Let me ask a different question then; where do you see the limits of the current interface? I completely understand that by making this PR I am asking you to take on the maintenance burden of supporting new features and I want to respect that you might not want to maintain my crap. 
That said, I'm also greedy here and I want you to maintain my crap :p

So, are you open to UI changes in the default web UI? If so what bounds on complexity (hard or soft) are you open to?

sampsyo:
That's a great question, and I don't have a 100% solid answer. On one hand, I think it would be cool to add features that don't take away from the bare-bones simplicity. On the other hand, I want to channel this energy in making better UIs in the bright [AURA][]-powered future where we can have lots of different UIs—simple ones, complex ones, weird ones—all developed independently and interoperable, rather than having the "one true UI" that is blessed by the beets distribution.

So that's not really a useful answer, I suppose, but maybe it gives some insight into roughly what's on my mind? But of course I'm just one person in the vast beets hivemind… it can always be nice to hear from others about what would be good to see, or to try getting folks to weigh in on the [Discourse][].

[aura]: https://github.com/beetbox/aura
[discourse]: https://discourse.beets.io

darakian:
After reading through the discourse a bit I think that's probably the place to have the conversation. Thanks and I'll close this PR.

PR #3637, opened 2020-06-27
fetchart: Use Cover Art Archive thumbnails

trolley:
The Cover Art Archive API offers pre-resized thumbnails of cover
art. If the `maxwidth` option of `fetchart` matches one of the
supported Cover Art Archive thumbnail sizes fetch it directly
instead of fetching the full size image then resizing it.

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -317,12 +317,26 @@ class CoverArtArchive(RemoteArtSource):
         """Return the Cover Art Archive and Cover Art Archive release group URLs
         using album MusicBrainz release ID and release group ID.
         """
+
+        # Cover Art Archive API offers pre-resized thumbnails at several sizes.
+        # If the maxwidth config matches one of the already available sizes
+        # fetch it directly intead of fetching the full sized image and
+        # resizing it.
+        release_url = self.URL
+        group_url = self.GROUP_URL
+        valid_thumbnail_sizes = [250, 500, 1200]
+        if plugin.maxwidth in valid_thumbnail_sizes:
+            size_suffix = "-" + str(plugin.maxwidth)
+            release_url += size_suffix
+            group_url += size_suffix
+
         if 'release' in self.match_by and album.mb_albumid:
-            yield self._candidate(url=self.URL.format(mbid=album.mb_albumid),
-                                  match=Candidate.MATCH_EXACT)
+            yield self._candidate(
+                url=release_url.format(mbid=album.mb_albumid),
+                                       match=Candidate.MATCH_EXACT)
         if 'releasegroup' in self.match_by and album.mb_releasegroupid:
             yield self._candidate(
-                url=self.GROUP_URL.format(mbid=album.mb_releasegroupid),
+                url=group_url.format(mbid=album.mb_releasegroupid),
                 match=Candidate.MATCH_FALLBACK)
 
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -214,6 +214,9 @@ Fixes:
   results or fetched lyrics
   :bug:`2805`
 * Adapt to breaking changes in Python's ``ast`` module in 3.8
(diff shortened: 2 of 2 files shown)


trolley:
Cover Art Archive API reference: https://musicbrainz.org/doc/Cover_Art_Archive/API#.2Frelease.2F.7Bmbid.7D.2F.28.7Bid.7D.7Cfront.7Cback.29-.28250.7C500.7C1200.29

Tested manually by setting the `fetchart` `maxwidth` option to:
* each of the supported values from Cover Art Archive (250, 500, 1200) and verify the fetched image is identical to the Cover Art Archive image
* a value that did not match one of those values and verified the final cover art file was resized to the proper size
* no value and verified that the full size art was retrieved

beetsplug/fetchart.py
```diff
@@ -314,6 +314,16 @@ class CoverArtArchive(RemoteArtSource):
         GROUP_URL = 'http://coverartarchive.org/release-group/{mbid}/front'
 
     def get(self, album, plugin, paths):
+        """Cover Art Archive API offers pre-resized thumbnails at several sizes.
+        If the maxwidth config matches one of the already available sizes
+        fetch it directly intead of fetching the full sized image and resizing
+        it."""
+        VALID_THUMBNAIL_SIZES = [250, 500, 1200]
```
sampsyo, line 321: `CAPITALIZED_NAMES` are generally only used for module-level constants. It's fine to make this lower-case, or to move it to a module-level constant with a more specific name.

beetsplug/fetchart.py
```diff
@@ -314,6 +314,16 @@ class CoverArtArchive(RemoteArtSource):
         GROUP_URL = 'http://coverartarchive.org/release-group/{mbid}/front'
 
     def get(self, album, plugin, paths):
+        """Cover Art Archive API offers pre-resized thumbnails at several sizes.
+        If the maxwidth config matches one of the already available sizes
+        fetch it directly intead of fetching the full sized image and resizing
+        it."""
```
sampsyo, line 320: In Python, multi-line string literals are used as [docstrings](https://en.wikipedia.org/wiki/Docstring), not comments. So by putting this here, at the top of the method, you have replaced the documentation for the method—which I don't think you intended. This can just be a plain old comment (i.e., using `#`). And both the comment and the code should come after the method's docstring.

beetsplug/fetchart.py
trolley: Thanks, I'm new with Python and beets so I'm learning as I go. I've already fixed this but didn't see your comment about the change log, so I'll do that as well now.

beetsplug/fetchart.py
```diff
@@ -317,6 +317,17 @@ def get(self, album, plugin, paths):
...
+
+        # Cover Art Archive API offers pre-resized thumbnails at several sizes.
+        # If the maxwidth config matches one of the already available sizes
+        # fetch it directly intead of fetching the full sized image and
+        # resizing it.
+        valid_thumbnail_sizes = [250, 500, 1200]
+        if plugin.maxwidth in valid_thumbnail_sizes:
+            size_suffix = "-" + str(plugin.maxwidth)
+            self.URL += size_suffix
+            self.GROUP_URL += size_suffix
```
sampsyo, line 329: Aha; one last thing I noticed: it will behave somewhat strangely to modify those class fields. Instead, let's change this locally:
```suggestion
        release_url = self.URL
        group_url = self.GROUP_URL
        if plugin.maxwidth in valid_thumbnail_sizes:
            size_suffix = "-" + str(plugin.maxwidth)
            release_url += size_suffix
            group_url += size_suffix
```

…and also use those local variables in the two calls below.

trolley: Done.

sampsyo:
Looks excellent; thanks again!

trolley:
@sampsyo - Unfortunately I discovered a problem with this after it was merged, which I logged here: https://github.com/beetbox/beets/issues/3638

I'll work on getting it fixed ASAP. Let me know if you'd rather I revert this change until then.

PR #3635, opened 2020-06-27
Fix genius lyrics backend artist matching with hyphens 

jtpavlock:
Fixes #3634

- [x] tests
- [x] changelog

PR #3641, opened 2020-06-28
Clean up code and tests for Genius Lyrics plugin backend

jtpavlock:
So this was bothering me while working on https://github.com/beetbox/beets/pull/3635, but I decided to keep things separated and clean things up on a new pr.

- [x] tests for Genius.fetch

do clean up commits like this require a changelog edit?


test/test_lyrics.py
```diff
@@ -457,28 +432,34 @@ def setUp(self):
...
+        mock = MockFetchUrl()
+        self.assertEqual(genius.scrape_lyrics_from_song_page(mock(url)), None)
+
+    def test_good_lyrics(self):
+        """Ensure we are able to scrape a page with lyrics"""
+        url = 'https://genius.com/Wu-tang-clan-cream-lyrics'
+        mock = MockFetchUrl()
+        self.assertIsNotNone(genius.scrape_lyrics_from_song_page(mock(url)))
+
+    # TODO: find an example of a lyrics page with multiple divs and test it
```
jtpavlock, line 459: I plan on leaving this one out as a `byte-sized` or perhaps `first-timers-only` issue ;)

test/test_lyrics.py
```diff
@@ -457,28 +432,34 @@ def setUp(self):
...
-    """Checks that Genius backend works as intended.
-    """
-    import requests
+class GeniusScrapeLyricsFromSongPageTest(GeniusBaseTest):
+    """tests Genius.scrape_lyrics_from_song_page()"""
 
     def setUp(self):
         """Set up configuration"""
-        LyricsGeniusBaseTest.setUp(self)
+        GeniusBaseTest.setUp(self)
```
jtpavlock, line 440: Lots of classes prefacing with 'Lyrics' even though we are in the lyrics.py namespace. I removed it, but is that common convention for this repo?

jtpavlock:
Looks like https://github.com/beetbox/beets/blob/533cc88df21dc0f4e3041a0dfd2eb2a82af9c4ed/.gitignore#L10 is causing some tests to fail. Is this intentional?

beetsplug/lyrics.py
```diff
@@ -359,16 +359,10 @@ def __init__(self, config, log):
...
-        self._log.debug(u'fetching lyrics from: {0}', page_url)
-        try:
-            page = requests.get(page_url)
-        except requests.RequestException as exc:
-            self._log.debug(u'Genius page request for {0} failed: {1}',
-                            page_url, exc)
-            return None
-        html = BeautifulSoup(page.text, "html.parser")
+    def scrape_lyrics_from_song_page(self, html):
+        """Scrape lyrics from a given genius.com song url"""
```
sampsyo, line 363: Seems like a good idea, but this docstring is not quite accurate because there is no URL involved here—just the HTML source.

test/test_lyrics.py
sampsyo: No, it's not a convention or anything. My best guess is that one of the initial tests did that, which sort of made sense in isolation, and then subsequent ones just copied it.

beetsplug/lyrics.py
jtpavlock: Yep, my mistake. I wrote that docstring before I outsourced the html fetching from this method.

test/test_lyrics.py
```diff
@@ -432,33 +431,90 @@ def setUp(self):
...
         """Ensure we are able to scrape a page with lyrics"""
         url = 'https://genius.com/Wu-tang-clan-cream-lyrics'
         mock = MockFetchUrl()
-        self.assertIsNotNone(genius.scrape_lyrics_from_song_page(mock(url)))
+        self.assertIsNotNone(genius._scrape_lyrics_from_html(mock(url)))
 
     # TODO: find an example of a lyrics page with multiple divs and test it
 
 
+class GeniusFetchTest(GeniusBaseTest):
```
jtpavlock, line 461: This is my first time really diving into unittests, so please let me know if there is a better way to do this

sampsyo:
Looks good! But it also appears that `test_good_lyrics` is failing (both on CI and on my local machine), apparently because of a missing `Wutangclancreamlyrics.txt`, presumably due to that `.gitignore` issue.

jtpavlock:
Whoops forgot to clean up that whole situation.. I'll see if I can figure that out

jtpavlock:
So do you not want me to check in any lyrics files? I suppose I could change the lyrics if needed cause they aren't actually important, just the structure is.

sampsyo:
Indeed, exactly, if that's not too annoying—that's what we've done with the "lyrics" files that are already checked in (they have lots of HTML structure but the lyrics themselves have been replaced with a description of beets from the README :smiley:).

jtpavlock:
Had some windows encoding issues, but seems like it's all settled now.

I chose not to check in all the google source files that were downloaded by the script, and instead just checked in ones that are currently used (just the genius ones). 

The script/some of the google sources also seemed to have some of their own issues which is why I opted to just avoid it. I plan on revisiting this lyrics plugin as a whole and rethinking some of the google sources when v1.5 gets released and I can transition the plugin to python3 as well.

Issue #2103, opened 2016-07-02
add a CONTRIBUTING.md file describing how to contribute to the project


jrobeson:
See https://help.github.com/articles/setting-guidelines-for-repository-contributors/ for details

sampsyo:
I'm for it.

jrobeson:
So, these tend to be markdown documents, but would we want the content in the documentation too? Also, how does it jive with the Hacking wiki page?

sampsyo:
Good question. Here's one possible route (not necessarily the best one):

* Move the content from the Hacking wiki page to the main Sphinx docs.
* Make the CONTRIBUTING file point to the docs for most of the technical stuff.
* CONTRIBUTING also contains the CoC link and overview—i.e., the nontechnical aspects.

jackwilsdon:
That sounds good to me; keep the main body of the documentation as Sphinx docs but add some information on the CoC to it.

What does GitHub think of using `.rst` for the `CONTRIBUTING` file? [Their documentation](https://help.github.com/articles/setting-guidelines-for-repository-contributors/) is a bit vague:

> Contributing guidelines filenames are not case sensitive, and **can have an extension such as .md or .txt**.

It looks like [MongoDB uses `CONTRIBUTING.rst`](https://github.com/mongodb/mongo/blob/master/CONTRIBUTING.rst), so it should work.

We're probably best putting it in the root directory instead of `.github`, as it's more of a general guideline than something GitHub-specific such as the issue template.

lengtche:
I was just talking to @sampsyo about this on IRC.

I did my first PR and had to go through the whole forking, cloning,  editable pip install, running tests locally, merging upstream to local before pushing to minimize conflicts, etc. While I'm not a new programmer by any means, it was my first time contributing to a project on GitHub. While it wasn't difficult, it did require a bit of tinkering before I got all pieces together.

I was thinking of submitting a PR once I get something fairly substantial in the Sphinx documentation.

FlaviaBastos:
Hey @lengtche 
I am new to this project and also noticed the lack of info for contributions. I spent some time documenting the first few (very) basic steps and created a 'development setup' rst doc to go under the [For Developers](http://beets.readthedocs.io/en/v1.4.3/dev/index.html) page. 
I was wondering if you made any progress with this or/and if you would be willing to check what I have so far. Maybe we can combine efforts and move this forward.

Please note my notes are very basic and assume the reader is new to the project and git/GitHub.
[...]

Issue #2386, opened 2017-01-13
Blog/Wiki idea: meet the users/meet the devs


nathdwek:
First of all let me apologize in advance if this seems unclear, inappropriate as an issue or too evangelistic. This is an idea I've had for some time but I'm only pulling the trigger late at night today, so expect some enthusiasm induced shenanigans.

The main idea would be to provide some content for the beets blog, to flesh out the docs with more practical infos, and for more advanced users to share tips and tricks, and to meet different usecases. The number of plugins and features of beets has been rising a lot lately, and I myself have been surprised by some functionalities I never suspected. A companion idea would be to attract some contributions as well with the "meet a dev" format, but that is less clear in my mind.

The concept would go along the lines of [this series of interviews](https://usesthis.com/). Basically, a comfortable user of beets tells us about how he came to use beets, what's his flow, what plugins he uses,... An interviewer wouldn't be needed, and the answers could or not have a more personal tone.

My first draft of a standardized set of questions would be

* Introduce yourself. How did you come in contacts with beets?
* How do you have beets installed? (OS + way of installing beets, especially interesting for windows users given the state of python (I think? Not really a windows user.), but also interesting for Mac and Unix users (package manager vs pip vs source))
* Where do you get your music from? What does an import look like for you?
* What packages do you use? Showcase some of them if you desire, elaborate on the config of some of them if you feel it is necessary.
* How do you listen to your music? If relevant, how does beet interact with your portable devices?
* If relevant, how do you store your music (Raid, on the cloud, sync between boxes, ...)
* Can you share your config with us? Are there last minute quirks you want to share?
* Exit comments?

This would also allow for newer users to have a look at a set of more rounded out config files that are actually in use currently, to summarize what mostly everyone is doing (I'll take a guess at fetchart/embedart, chroma and convert), and to also be put in contact with more exotic stuff.

Now for the dev part... I guess if we go through with this, the first couple of segments would be with actual beets contributors and regulars (even though I really, *really* hope we'd get to hear from other users as well!), and that those people would still make up a significant proportion them even long term.
These beet developers could answer some extra questions centred around, you guessed it, contributing to beets. This is less clear for me and there might actually be a lot less content to go over than I think, but some questions that come to mind:

* How did you start contributing to beets? Where do you recommend to start for a fresh contributor?
* Editor + beet-specific config + linters?
* How do you run the test suite?
[...]

PR #3643, opened 2020-06-28
add contributing.md & update readme

jtpavlock:
This would effectively replace the [hacking wiki page](https://github.com/beetbox/beets/wiki/Hacking)

Fixes https://github.com/beetbox/beets/issues/2103


contributing.md
```diff
@@ -0,0 +1,163 @@
...
+  changes.
+* GUI design. For the time being, beets is a command-line-only affair. But
+  that's mostly because I don't have any great ideas for what a good GUI
+  should look like. If you have those great ideas, please get in touch.
+* Benchmarks. We'd like to have a consistent way of measuring speed
+  improvements in beets' tagger and other functionality as well as a way of
+  comparing beets' performance to other tools. You can help by compiling a
+  library of freely-licensed music files (preferably with incorrect metadata)
+  for testing and measurement.
+* Think you have a nice config or unique use-case for beets? We'd love to hear about it! Submit a post to our [our forums](https://discourse.beets.io/) for a chance to get featured in our blog or use-case page.
```
jtpavlock, line 25: This is something that obviously isn't already implemented, but I think could be a cool idea to have a wiki page or similar that highlights unique or interesting things people do with beets

contributing.md
```diff
@@ -0,0 +1,163 @@
...
+* There are several ways to improve the tests in general (see [[Testing]]) and some places to think about performance optimization (see [[Optimization]]).
+* Not all of our code is up to our coding conventions. In particular, the [API documentation][apidocs] are currently quite sparse. You can help by adding to the docstrings in the code and to the documentation pages themselves.
+
+## Your first contribution
+If this is your first time contributing to an open source project, welcome! If you are confused at all about how to contribute or what to contribute, take a look at [this great tutorial](http://makeapullrequest.com/), or stop by our [gitter chat room](https://gitter.im/beetbox/beets#) if you have any questions. 
+
+We maintain a list of issues we feel are perfect for first-timers tagged "[good first issue](https://github.com/beetbox/beets/labels/good%20first%20issue)". These issues are perfect for anyone new to open source. In fact, we encourage only users who are new to open source to tackle these. Ideally, the goal for these issues is to get users comfortable with contributing to an open source project for the first time. Please, do not hesitate to ask any questions.
```
jtpavlock, line 79: So I mentioned `first-timers-only` on discourse, but I believe this is a good enough equivalent. Basically, imo, the philosophy should be issues where the only real thinking should be how to create a fork and pull request. Ideally, any issue issue with this tag would be restricted (or heavily encouraged) to only be tackled by users who are first-timers to open source projects. Additionally, any issues with this tag would outline in detail what steps the user should take to correct the issue to the extent of what files or lines to change.

README.rst
```diff
@@ -105,6 +103,17 @@ news and updates.
 .. _its Web site: https://beets.io/
 .. _@b33ts: https://twitter.com/b33ts/
 
+Contact
+-------
+* Encountered a bug you'd like to report or have an idea for a new feature? Check our our `issue tracker`_! If your issue or feature hasn't already been reported, please `open a new ticket`_ and we'll be in touch with you shortly.
+* Need help/support, would like to start a discussion, or would just like to introduce yourself to the team? Check out our `forums`_!
```
jtpavlock, line 109: Are you wanting support questions on github or discourse?

contributing.md
```diff
@@ -0,0 +1,163 @@
...
+
+
+### Code contribution ideas
+* We maintain a set of [issues marked as "bite-sized"](https://github.com/beetbox/beets/labels/bitesize). These are issues that would serve as a good introduction to the codebase. Claim one and start exploring!
+* Our [test coverage](https://codecov.io/github/beetbox/beets) is somewhat low. You can help out by finding low-coverage modules on Coveralls and adding new tests.
+* There are several ways to improve the tests in general (see [Testing](https://github.com/beetbox/beets/wiki/Testing)) and some places to think about performance optimization (see [Optimization](https://github.com/beetbox/beets/wiki/Optimization)).
+* Not all of our code is up to our coding conventions. In particular, the [API documentation][apidocs] are currently quite sparse. You can help by adding to the docstrings in the code and to the documentation pages themselves.
```
jtpavlock, line 74: Does beets enforce a particular type of docstring standard?

contributing.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Your first contribution
+If this is your first time contributing to an open source project, welcome! If you are confused at all about how to contribute or what to contribute, take a look at [this great tutorial](http://makeapullrequest.com/), or stop by our [gitter chat room](https://gitter.im/beetbox/beets#) if you have any questions. 
+
+We maintain a list of issues we feel are perfect for first-timers tagged "[good first issue](https://github.com/beetbox/beets/labels/good%20first%20issue)". These issues are perfect for anyone new to open source. In fact, we encourage only users who are new to open source to tackle these. Ideally, the goal for these issues is to get users comfortable with contributing to an open source project for the first time. Please, do not hesitate to ask any questions.
+
+## How to Submit Your Work
```
jtpavlock, line 81: I'm thinking a pull request template would be useful to define requirements for tests, documentation, and changelog entries. Something as simple as a todo checklist could help.

README.rst
```diff
@@ -105,6 +103,17 @@ news and updates.
...
+* If you are a developer or a prospective developer looking for some guidance, we have a `gitter chat`_. Feel free to stop by anytime!
```
jtpavlock, line 110: I'm hoping gitter can become more of a common-place dev-focused chat room in favor of any previously suggested irc (or whatever else) channels.

contributing.md
```diff
@@ -0,0 +1,163 @@
...
+* `mediafile`
+* `jellyfish`
+
+Please be aware that this list might be out of date, but [setup.py](https://github.com/beetbox/beets/blob/60bba370c0ecc58b4057d78867273e5936d22dee/setup.py#L89-L110) contains a list that is always up to date. With all the required modules in place, you can just type `./beet` in the source directory to run beets from there.
+
+
+### Code contribution ideas
+* We maintain a set of [issues marked as "bite-sized"](https://github.com/beetbox/beets/labels/bitesize). These are issues that would serve as a good introduction to the codebase. Claim one and start exploring!
+* Our [test coverage](https://codecov.io/github/beetbox/beets) is somewhat low. You can help out by finding low-coverage modules on Coveralls and adding new tests.
+* There are several ways to improve the tests in general (see [Testing](https://github.com/beetbox/beets/wiki/Testing)) and some places to think about performance optimization (see [Optimization](https://github.com/beetbox/beets/wiki/Optimization)).
```
jtpavlock, line 73: I noticed this mentions use of Nose, which is now unmaintained. Is this current info, or should beets perhaps consider other avenues? I could have sworn I've seen talk on this before, but I can't seem to find it now.

contributing.md
```diff
@@ -0,0 +1,163 @@
...
+* `confuse`
+* `mediafile`
+* `jellyfish`
+
+Please be aware that this list might be out of date, but [setup.py](https://github.com/beetbox/beets/blob/60bba370c0ecc58b4057d78867273e5936d22dee/setup.py#L89-L110) contains a list that is always up to date. With all the required modules in place, you can just type `./beet` in the source directory to run beets from there.
+
+
+### Code contribution ideas
+* We maintain a set of [issues marked as "bite-sized"](https://github.com/beetbox/beets/labels/bitesize). These are issues that would serve as a good introduction to the codebase. Claim one and start exploring!
+* Our [test coverage](https://codecov.io/github/beetbox/beets) is somewhat low. You can help out by finding low-coverage modules on Coveralls and adding new tests.
```
jtpavlock, line 72: I'm not familiar with this workflow and am actually confused how one goes about pinpointing what areas of code could use some better test coverage. I've only use pytest/pytest-cov before XD. I really note this because there's a decent chance I'm not alone in that

contributing.md
```diff
@@ -0,0 +1,163 @@
...
+## How to Submit Your Work
+
+Do you have a great bug fix, new feature, or documentation expansion you'd like to contribute? Follow these steps to create a GitHub pull request and your code will ship in no time.
+
+1. Fork the beets repository and clone it (see above) to create a workspace.
+2. Make your changes.
+3. Add tests. If you've fixed a bug, write a test to ensure that you've actually fixed it. If there's a new feature or plugin, please contribute tests that show that your code does what it says.
+4. Add documentation. If you've added a new command flag, for example, find the appropriate page under `docs/` where it needs to be listed.
+5. Add a changelog entry to `docs/changelog.rst` near the top of the document.
+6. Run the tests and style checker. The easiest way to run the full battery of tests is to use [tox](https://tox.readthedocs.org/en/latest/). This will run the same set of tests that we run on Travis.
```
jtpavlock, line 90: I noticed errors running tox and had to figure out how to install the specific python versions it requires for it's tests. Maybe the best place to explain that nuance is in the Testing wiki and then re-linking that page here?

contributing.md
```diff
@@ -0,0 +1,163 @@
...
+7. Push to your fork and open a pull request! We'll be in touch soon.
+8. If you add commits to a pull request, please *add a comment* after you push them. (GitHub doesn't notify us when commits are added, so we need your comment so we get an email alert.)
+
+Remember, code contributions have four parts: the code, the tests, the documentation, and the changelog entry. Thank you for contributing!
+
+
+# The Code
+
+The documentation has an [API documentation section][apidocs] that serves as an introduction to beets' design. There's also an [[Architecture]] page on this wiki.
```
jtpavlock, line 99: Both of these pages seem to be dead or non-existing

contributing.md
```diff
@@ -0,0 +1,163 @@
...
+[apidocs]: http://beets.readthedocs.org/en/latest/dev/api.html
+
+## Coding Conventions
+
+There are a few coding conventions we use in beets:
+
+* Whenever you access the library database, do so through the provided Library
+  methods or via a Transaction object. Never call `lib.conn.*` directly.
+  Transaction objects help control concurrent access to the database and assist
+  in debugging conflicting accesses.
```
jtpavlock, line 110: I think a code example of what to do vs what not to do could help here. This seems the least obvious of all the conventions.

README.rst
```diff
@@ -105,6 +103,17 @@ news and updates.
 .. _its Web site: https://beets.io/
 .. _@b33ts: https://twitter.com/b33ts/
 
+Contact
+-------
+* Encountered a bug you'd like to report or have an idea for a new feature? Check our our `issue tracker`_! If your issue or feature hasn't already been reported, please `open a new ticket`_ and we'll be in touch with you shortly.
```
jtpavlock, line 108: I'm not sure if you do this already internally, but it might be worth defining a way to sort feature requests. Something as simple as give a 👍 on features you'd most like to see implemented (at the cost of implementing other features). This way we can sort by 👍 on the issue tracker and go from there. Also hopefully cuts down on annoying +1s...

contributing.md
sampsyo: Yeah, seems cool! There was even a proposal to do something quite similar over in #2386. Nobody actually picked up the mantle but I still think it's a very cool idea.

contributing.md
sampsyo: Cool, that sounds good. I have no strong feelings about the name, certainly, but I like this idea of scoping these entry-level issues so they're appropriate for people who want to learn how to make a PR in the first place, for example. Maybe we want a separate tag for smallish issues that are more appropriate for experienced hackers, but maybe that's not really a significant category.

README.rst
sampsyo: Here's how I'd break it down: if you have a concrete problem to report (e.g., a traceback), make an issue. If you are not sure whether it's your fault or beets's, start a thread on Discourse and we might redirect you to an issue if it looks like a bug. So yeah, I think that means what you've said here: "help/support" goes to the forum.

contributing.md
jtpavlock: Yeah, for example, [flexget ](https://flexget.com/) maintains a [cookbook](https://flexget.com/Cookbook) for useful "recipes" (configurations) for both advanced stuff and more common simple use cases.

The "meet the devs/users" thing that post suggests is another cool idea, and perhaps the next step after implementing something like this.

contributing.md
sampsyo: We generally follow [PEP 257](https://www.python.org/dev/peps/pep-0257/). In some places, we also sometimes use ReST [autodoc](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) syntax for Sphinx to, for example, refer to a class name.

contributing.md
sampsyo: Yeah, sounds like a really good idea. Generally speaking, I definitely like to require docs and a changelog with every PR. (I can't count the number of times I've asked people to add a changelog entry—it would be great to have the PR template do that for me!)

Tests I'm a little less religious about—always nice to have, but some things in beets are _really_ nasty to test. Maybe we can make that more of a suggestion than a requirement.

contributing.md
sampsyo: I don't have a link to a previous thread, but the way things stand are that I actually do tend to use nose even though it's unmaintained. I should probably switch to pytest along with the rest of the world. But I do kind of like the status quo where our tests are framework-agnostic, in the sense that nose is *just* a test runner and pytest (or `python -m unittest` or [ptr](https://github.com/facebookincubator/ptr) or whatever) works just as well.

contributing.md
jtpavlock: That's actually what I think of when I think of `byte-sized`. Not necessarily any specific difficulty, but small in size. I think the more labels you create, the harder it can be to keep track of them while remaining consistent in using them, and more time gets spent on considering the difficulty and size of issues. Whether or not that's worthwhile, I'll leave up to you.

contributing.md
sampsyo: Well, one thing here is that the reference to Coveralls is very out of date. :smiley: We're using [Codecov](https://codecov.io) now—which is not too terribly different from using pytest-cov except that it reports the results in a GUI.

README.rst
jtpavlock: Did you see my [note on discourse](https://discourse.beets.io/t/forming-a-beets-core-team/639/15?u=jyrik) about gitter dropping mobile support? May make it less accessible than other options if you're willing to consider other avenues.

contributing.md
sampsyo: Yes, good point. We also have a [wiki page on testing](https://github.com/beetbox/beets/wiki/Testing) that should perhaps have some of its information collapsed here.

I don't know of a succinct description to link to, but here's a shot at capturing what we need to say:

> Tox tries to test on multiple Python versions by default. You'll see errors for the Python versions you don't have installed. It can be expedient to run the tests only on the version(s) of Python you already have. Use `tox -e py38-test` to only run the tests on Python 3.8, for example.

contributing.md
sampsyo: Yes, these have moved to the `dev` section of the docs:
https://beets.readthedocs.io/en/stable/dev/index.html

contributing.md
jtpavlock: In the effort of being actionable, and since I'm bad at coming up with names, I propose we add a label in discourse called 'recipe' then promote people to submit any cool or unique configurations or "recipes" to discourse that could then be added under a wiki page called "cookbook".

contributing.md
sampsyo: Indeed. Again with a rough synopsis, here's about what we need to say: to fetch `Item` objects from the database, use `lib.items(...)` and supply a query as an argument. Resist the urge to write raw SQL for your query. If you _must_ use lower-level queries into the database, do this:

```py
with lib.transaction() as tx:
    rows = tx.query('SELECT …')
```

(Insert the above explanation of why we have `Transaction`.) Here's an example in the web plugin:
https://github.com/beetbox/beets/blob/533cc88df21dc0f4e3041a0dfd2eb2a82af9c4ed/beetsplug/web/__init__.py#L152

README.rst
sampsyo: Yes, great idea! Explicitly recommending that folks search the issue tracker for similar issues before opening a new one (and voting!) seems very helpful.

contributing.md
jtpavlock: Yeah I can make a note in the pr template that tests are suggested but not required. Do you want issues to be created if people omit tests to help document what we're missing? Or just leave it out entirely?

contributing.md
sampsyo: As for a cookbook specifically, we do have a section of the docs that was intended to be something somewhat similar:
https://beets.readthedocs.io/en/stable/guides/advanced.html

Although posting in the Discourse is a lower-effort alternative. I like the Discourse idea, and we can "graduate" things to the docs if they need even more visibility. :smiley:

README.rst
```diff
@@ -87,14 +87,12 @@ Check out the `Getting Started`_ guide for more information.
 .. _Getting Started: https://beets.readthedocs.org/page/guides/main.html
 .. _software repositories: https://repology.org/project/beets/versions
 
-Contribute
+Contributing
```
sampsyo, line 90: I know this is a little silly, but the rest of the headings in the README are all imperative-mood verbs, so I kind of liked the aesthetic of letting this one stay the same way too. :smiley:

contributing.md
```diff
@@ -0,0 +1,163 @@
...
+dependencies automatically, but if you do it manually, you can use pip to
+install these Python modules:  
+* `mutagen`
+* `munkres`
+* `unidecode`
+* `pyYAML`
+* `musicbrainzngs`
+* `confuse`
+* `mediafile`
+* `jellyfish`
```
sampsyo, line 65: Maybe let's just get rid of this section—I think `pip install -e` should probably be the recommendation for basically everybody.

contributing.md
sampsyo: I think it's OK to leave it out.

contributing.md
jtpavlock: Yeah that's true, I'll see if I can make some changes to the Testing wiki to clarify some of this stuff.

README.rst
jtpavlock: I'll leave the gitter stuff in here as I assume we'll continue to use it until an alternative is decided upon.

contributing.md
jtpavlock: I'll try to clarify more in the wiki page and just link to it here

jtpavlock:
Alright, I think that's everything. Outstanding things that make sense following these changes are:

- [ ] add a discourse tag for posts that are config/use-case submissions to be featured in the docs (and also maybe advertise this new feature via twitter?)
- [ ] Add a pull request template (I can do this one)
- [ ] Add clarification/changes to the Testing wiki page (I can also do this one)
- [ ] Add `first-timers-only label` in place of current `good-first-issue`. If you don't care either way, I personally think `first-timers-only` seems more inviting and accessible to newcomers with the appearance that they have a spot already reserved for them.
- [ ] With the above, I'd like to see any issues marked `first-timers-only` to have more detailed instructions on how to tackle them for any potential contributors to be able to easily hit the ground running. 

If you're willing to permit me some access, I can also take care of the label changes. Up to you 😄

sampsyo:
Sounds great. I'll also to an end-to-end pass on the new contributing document.

I added you to the GitHub org and blessed you as a Discourse admin—let me now if that's sufficient permissions. :smiley:

contributing.md
jtpavlock: Made a new category on discourse called [ "Show and Tell"](https://discourse.beets.io/t/about-the-show-and-tell-category/1377) and advertised it [here](https://discourse.beets.io/t/new-categories/1378).

jtpavlock:
Since it seems the gitter is going to be abandoned in favor of other options, do you want to leave that in until we find an alternative, or just take it out advise going to the discourse for now?

README.rst
jtpavlock: Seems the gitter is likely to be abandoned now. Should we even advertise it? Or just direct people to the discourse?

sampsyo: I don't have a strong feeling either way—there's no particular harm in leaving the think for now, even if the Gitter is not long for this world.

jtpavlock:
What do you think about adding this to the docs? Maybe a `Contributing` section that could then break down into `for developrs` and also include things like code convention, testing info (the current testing wiki), etc.

jtpavlock:
[por que no los dos?](https://github.com/pytest-dev/pytest/blob/master/doc/en/contributing.rst)

jtpavlock:
Just removed some now outdated info and did another look over. Is there anything else I can do to help get this in?

sampsyo:
Did a quick editing/writing/style pass and I think this is good to go! Thank you for all the hard work on this, @jtpavlock!! :tada: :sparkles: :rocket:

jtpavlock:
Merged, thanks for the review!

PR #3646, opened 2020-06-28
update bug-report flow

jtpavlock:
So, I think the bug template could use some attention. My proposal is to create a wiki page that would go through the proper steps of reporting a bug and self-diagnosis, then hopefully have the template as a nice plug and play of the proper info. I made a sample wiki page at my fork [here](https://github.com/jtpavlock/beets/wiki/How-to-create-a-good-bug-report)

I also added a dropdown for the config file that can also be set to be open automatically. I figured this could make some of those issues with painfully long configs easier to read. The template isn't as explicit as before on purpose, to hopefully encourage users to take a look at the wiki page :)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -26,7 +26,7 @@ $ beet -vv (... paste here ...)
 ### Actual Result
 (if more explanation needed beyond command output)
 
-## Setup
+### Setup
 
 * OS: 
 * Python version: 
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -12,7 +12,7 @@ Hi there, thank you for reporting a bug! Please read our `how to create a good b
 The better the bug report, the quicker we can get to fixing it :rocket:
 
 ## Description
-
+(overview of the issue)
 
 ### Steps to reproduce:
 Running this command in verbose (`-vv`) mode:
@@ -21,7 +21,10 @@ $ beet -vv (... paste here ...)
 ```
 
 ### Expected Result
+(What did you expect to happen?)
 
+### Actual Result
+(if more explanation needed beyond command output)
 
 ## Setup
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -6,37 +6,33 @@ about: Report a problem with beets
 
 <!--
 Describe your problem, feature request, or discussion topic here.
-
-Please fill out this and the "Setup" section below and remember to include
-enough detail so that other people can reproduce the problem.
 -->
 
-### Problem
+Hi there, thank you for reporting a bug! Please read our `how to create a good bug report` wiki page to expedite the process of diagnosing the issue.
```
jtpavlock, line 11: this obviously turns into a link if the page existed

sampsyo:
Seems great, and the `<details>` thing seems like a good idea for the configuration.

There are two things about the old, slightly-more-specific template that I thought were effective:

- The rigid format that suggested that people paste a command and an output at the top. This doesn't _always_ apply, but it's good for about 90% of cases. With the more common but more vague "steps to reproduce"/"actual result" sections, I notice that people are sometimes encouraged to write in free-form prose about what they tried to do. That's nice to have too, but only when it's accompanied by the _actual command_ and and _actual, complete_ paste of the output. Including that `beet -vv` there helps people immediately see what they need to do to get that invaluable debug output.
- The "Turning off plugins made problem go away (yes/no):" prompt is also helpful because so many problems can be traced to plugins—but we really depend on people experimentally figuring out which one. So while it would be great to describe the process of disabling plugins one at a time in the accompanying wiki page, I like having that line there to be an obvious, visceral reminder that this should be your first recourse when trying to track down the root cause of a problem.

jtpavlock:
That makes sense, I added a bit more handholding in the template as you suggest borrowing from the old template. 

Keep in mind you're the one dealing with 99.9% of the issues on this repo, so I'm just trying to help you out. If you feel any of these changes are worse for you, you won't hurt my feelings.

jtpavlock:
So I goofed and removed my fork and consequently the "how to create a good bug report" wiki page. Not sure if any of these changes are worth merging (perhaps just the config dropdown?), but I can rebase this and change whatever.

PR #3644, opened 2020-06-28
fetchart: Use Cover Art Archive thumbnails

trolley:
The Cover Art Archive API offers pre-resized thumbnails of cover
art. If the `maxwidth` option of `fetchart` matches one of the
supported Cover Art Archive thumbnail sizes, and a thumbnail of
that size exists in the Cover Art Archive, fetch it directly
instead of fetching the full size image then resizing it.

PR #3647, opened 2020-06-28
fetchart: Use Cover Art Archive thumbnails

trolley:
The Cover Art Archive API offers pre-resized thumbnails of cover
art. If the `maxwidth` option of `fetchart` matches one of the
supported Cover Art Archive thumbnail sizes, and a thumbnail of
that size exists in the Cover Art Archive, fetch it directly
instead of fetching the full size image then resizing it.

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -305,6 +305,7 @@ class RemoteArtSource(ArtSource):
 class CoverArtArchive(RemoteArtSource):
     NAME = u"Cover Art Archive"
     VALID_MATCHING_CRITERIA = ['release', 'releasegroup']
+    VALID_THUMBNAIL_SIZES = [250, 500, 1200]
 
     if util.SNI_SUPPORTED:
         URL = 'https://coverartarchive.org/release/{mbid}/front'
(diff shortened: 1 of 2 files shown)


trolley:
This is an alternative to PR #3644. This is much simpler, now that I understand how `yield` works. It looks good with my test of an album that does not have my requested thumbnail size available:

```
fetchart: trying source coverart for album 88 Fingers Louie - Behind Bars
fetchart: downloading image: https://coverartarchive.org/release/80a2fbb8-06bc-4f6c-ad7d-f3e277ab7376/front-1200
fetchart: not a supported image: image/x-None
fetchart: downloading image: https://coverartarchive.org/release/80a2fbb8-06bc-4f6c-ad7d-f3e277ab7376/front
fetchart: downloaded art to: /tmp/tmp6s5i253z.jpg
fetchart: image size: (1080, 1066)
fetchart: using remote image /tmp/tmp6s5i253z.jpg
```

If this approach is preferred I'll close the other PR.

beetsplug/fetchart.py
```diff
@@ -317,13 +318,31 @@ def get(self, album, plugin, paths):
...
         using album MusicBrainz release ID and release group ID.
         """
+        release_url = self.URL.format(mbid=album.mb_albumid)
+        release_group_url = self.GROUP_URL.format(mbid=album.mb_releasegroupid)
+
+        # Cover Art Archive API offers pre-resized thumbnails at several sizes.
+        # If the maxwidth config matches one of the already available sizes
+        # fetch it directly intead of fetching the full sized image and
+        # resizing it.
+        size_suffix = ""
```
sampsyo, line 328: Perhaps this ought to be `None` to make it clear that we won't append this if it's not filled in.

trolley: Done.

PR #3655, opened 2020-07-03
Get the date of the parentwork, which may be different from the date of the child work.

### LINKED ISSUE #3650: work_date of parentwork with parentwork plugin
I currently have my beets library paths looking like this (I mostly have classical music, so I use the parentwork plugin): parent_composer_sort/work_date - parentwork/..., but if different subworks have different dates, this can store them in different directories.  Is there a way to work around this, perhaps in the path config?  Or would I need to write a plugin or edit the parentwork plugin?

aereaux:
##  Description

Fixes #3650.

## To Do

### DIFF
--- a/beetsplug/parentwork.py
+++ b/beetsplug/parentwork.py
@@ -131,6 +131,8 @@ class ParentWorkPlugin(BeetsPlugin):
                 if artist['type'] == 'composer':
                     parent_composer.append(artist['artist']['name'])
                     parent_composer_sort.append(artist['artist']['sort-name'])
+                    if 'end' in artist.keys():
+                        parentwork_info["parentwork_date"] = artist['end']
 
             parentwork_info['parent_composer'] = u', '.join(parent_composer)
             parentwork_info['parent_composer_sort'] = u', '.join(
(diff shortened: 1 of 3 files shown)


dosoe:
Looks fine to me! I'm wondering if we should keep the ```work_date``` tag, maybe this new one makes it redundant.

dosoe:
There are two ways to get the date: either you just go up the work tree and pick the first one you find (```work_date```) or you go up to the parentwork and take its date (```parentwork_date```). Generally they are the same, but on MB all works don't have a date so picking the first date you find makes it more probable that you will actually end up with a date. On the other hand, especially for collections, different parts of a work can have different composition dates so with ```work_date``` you end up with different dates for the same parentwork whereas with ```parentwork_date``` you most probably end up with no date at all in these cases, but at least a given parentwork always has the same ```parentwork_date```.

dosoe:
One maybe: is there some kind of formatting for dates in beets that the dates should be converted to?

aereaux:
I'm not sure if there is a specific date format style, but I copied this more or less from the way `work_date` is generated, so I don't think there is.

dosoe:
Then it should be fine and ready to merge.

sampsyo:
Excellent!

PR #3658, opened 2020-07-06
Add weekly integration test github action

### LINKED ISSUE #3645: feat(ci): Add github action that runs weekly(?) for integrated api tests
### Use case

Having integration tests that hit actual APIs and make real requests every once in a while can be a good idea to try and catch when they change things up on us. 

### Solution
Add something like the following to any tests making an actual request:
```
@unittest.skipUnless(os.environ.get(
        'INTEGRATION', '0') == '1', 'real integration tests not enabled')
```

Then, have a github action that runs all tests weekly with this flag enabled.

For a better idea of what the individual tests may look like, [check this out.](https://realpython.com/testing-third-party-apis-with-mocks/#conditionally-testing-scenarios)

I'm not super confident in what the appropriate terminology is here, but I think you get the idea.

Also, if this is implemented. Documentation in the [Testing wiki page](https://github.com/beetbox/beets/wiki/Testing) should include guidance on writing/skipping integration tests

jtpavlock:
## Description

Fixes #3645

Adds a weekly integration (hitting real apis) test as a github action. Also adjust `test_lyrics.py` to use the new integration tests.

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -1,0 +1,32 @@
+name: integration tests
+on:
+  schedule:
+    - cron: '0 0 * * SUN'  # run every Sunday at midnight
+jobs:
+  test integration:
+    runs-on: ubuntu-latest
+    strategy:
+      python-version: 3.8
+    env:
+      NOSE_SHOW_SKIPPED: 1
+      PY_COLOR: 1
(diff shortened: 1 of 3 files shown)


jtpavlock:
Yeah, that's seems like a good idea. I'm going to revisit tox in another PR and I'll that to the list of things to consider.

Issue #2969, opened 2018-06-29
Lyrics are wrong


Flurrywinde:
### Problem

I used all default values in my config file, and beets imported lyrics successfully. However, for Manowar's "Fast Taker", they were wrong, so I added the following to my config file (/home/kanon/.config/beets/config.yaml):
```
lyrics:
    google_API_key: [redacted]
    sources: google
```
The same wrong lyrics appeared, despite google's regular search turning up different, better lyrics. (The first result in the "beets custom search engine" is different from both of these.)

Is there a way to tell which source beet is getting the lyrics from?

Also, is there a way to select which google source is used?

Is there a list of sources? For example, I see "lyrics.com" and "lyricwiki" on beets.readthedocs.io.

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv  lyrics -fp manowar fast taker
```

Led to this problem:

```
user configuration: /home/kanon/.config/beets/config.yaml
data directory: /home/kanon/.config/beets
Sending event: pluginload
no config file found at /home/kanon/.beetsconfig
library database: /home/kanon/Music/musiclibrary.db
library directory: /home/kanon/Music
Sending event: library_opened
google lyrics backend error: accessNotConfigured
got lyrics from backend: fetch_lyricswiki
fetched lyrics: Manowar - Fast Taker
Sending event: write
Sending event: after_write
Sending event: database_change
[lyrics here]
```
I notice it says: "google lyrics backend error: accessNotConfigured". Here is how I set up my google api key:
![opera snapshot_2018-06-29_002558_console developers google com](https://user-images.githubusercontent.com/27073281/42078983-4360ec86-7b33-11e8-8927-cb87c5cff443.png)

Is that correct? Why isn't it working?

### Setup

* OS: Ubuntu 16.04
* Python version: Python 3.6.5 :: Anaconda, Inc.
* beets version: beets version 1.3.8
* Turning off plugins made problem go away (yes/no): no

My configuration (output of `beet config`) is:

```yaml
directory: ~/Music

import:
    copy: no
lyrics:
    google_API_key: AIzaSyDELPsVTGyXRywPrXx[redacted]
    sources: google
    auto: yes
    fallback:
    google_engine_ID: 009217259823014548361:lndtuqkycfu
library: ~/Music/musiclibrary.db

plugins: fromfilename fetchart lyrics
fetchart:
    cautious: no
    maxwidth: 0
    auto: yes
    remote_priority: no
    google_search: no
    cover_names:
[...]

PR #3668, opened 2020-07-12
Fix a regression in the heuristic for detecting valid lyrics from the google source

wisp3rwind:
See https://github.com/beetbox/beets/issues/2969


sampsyo:
Nice find. Thanks for fixing this.

PR #3670, opened 2020-07-13
Add possibility to select individual items to the remove CLI command

wisp3rwind:
I just wanted to remove a few individual tracks from my library, for which a query was not simple to construct. As it turns out, the `remove` command wasn't very helpful in this case. Hence, this adds interactive selection of items to the `remove` command, analogously to item selection in the `modify` command. The `yes/no` answers to the prompt work as before, and the `select` answer is added. Thus, there's no change of previous behaviour.

I also changed the output when `-a` is given to group albums more clearly.

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -389,17 +389,19 @@ def input_yn(prompt, require=False):
     return sel == u'y'
 
 
-def input_select_objects(prompt, objs, rep):
+def input_select_objects(prompt, objs, rep, prompt_all=None):
     """Prompt to user to choose all, none, or some of the given objects.
     Return the list of selected objects.
 
     `prompt` is the prompt string to use for each question (it should be
-    phrased as an imperative verb). `rep` is a function to call on each
-    object to print it out when confirming objects individually.
+    phrased as an imperative verb). If `prompt_all` is given, it is used
+    instead of `prompt` for the first (yes(/no/select) question.
+    `rep` is a function to call on each object to print it out when confirming
+    objects individually.
     """
     choice = input_options(
         (u'y', u'n', u's'), False,
-        u'%s? (Yes/no/select)' % prompt)
+        u'%s? (Yes/no/select)' % (prompt_all or prompt))
     print()  # Blank line.
 
     if choice == u'y':  # Yes.
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1232,31 +1232,53 @@ def remove_items(lib, query, album, delete, force):
     """
     # Get the matching items.
     items, albums = _do_query(lib, query, album)
+    objs = albums if album else items
 
     # Confirm file removal if not forcing removal.
     if not force:
         # Prepare confirmation with user.
-        print_()
+        album_str = u" in {} album{}".format(
+                 len(albums), u's' if len(albums) > 1 else u''
+            ) if album else ""
+
         if delete:
             fmt = u'$path - $title'
-            prompt = u'Really DELETE %i file%s (y/n)?' % \
-                     (len(items), 's' if len(items) > 1 else '')
+            prompt = u'Really DELETE'
+            prompt_all = u'Really DELETE {} file{}{}'.format(
+                 len(items), u's' if len(items) > 1 else u'', album_str
+            )
         else:
             fmt = u''
-            prompt = u'Really remove %i item%s from the library (y/n)?' % \
-                     (len(items), 's' if len(items) > 1 else '')
+            prompt = u'Really remove from the library?'
+            prompt_all = u'Really remove {} item{}{} from the library?'.format(
+                 len(items), u's' if len(items) > 1 else u'', album_str
+            )
+
+        # Helpers for printing affected items
+        def fmt_track(t):
+            ui.print_(format(t, fmt))
+
+        def fmt_album(a):
+            ui.print_()
+            for i in a.items():
+                fmt_track(i)
+
+        fmt_obj = fmt_album if album else fmt_track
 
         # Show all the items.
-        for item in items:
-            ui.print_(format(item, fmt))
+        for o in objs:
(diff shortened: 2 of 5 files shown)


beets/ui/__init__.py
```diff
@@ -399,7 +399,7 @@ def input_select_objects(prompt, objs, rep):
     """
     choice = input_options(
         (u'y', u'n', u's'), False,
-        u'%s? (Yes/no/select)' % prompt)
+        u'%s? (Yes/no/select)' % (prompt_all or prompt))
```
sampsyo, line 402: I'm a little confused about `prompt` vs. `prompt_all`. Two questions, I guess:

- Is it equivalent to just do `prompt_all or prompt` when *calling* this function, rather than inside the function? That seems a little simpler.
- Since the latter always overrides the former if provided, and both seem to be provided in `remove_items` below, what is the purpose of plain old `prompt`?

beets/ui/commands.py
```diff
@@ -1228,31 +1228,57 @@ def remove_items(lib, query, album, delete, force):
...
+        # Helpers for printing affected items
+        def fmt_track(t):
+            ui.print_(format(t, fmt))
+
+        def fmt_album(a):
+            ui.print_()
+            ui.print_(format(a, album_fmt))
+            ui.print_(u"which contains the tracks")
+            for i in a.items():
+                fmt_track(i)
```
sampsyo, line 1264: This display of albums and their contents is a little nonstandard for beets. What do you think about leaving the old behavior (just list the items)? This could still work by selecting albums when doing the interactive selection, but you would also still get the full list of files that would be deleted when "previewing" your selection.

beets/ui/__init__.py
wisp3rwind: It's not equivalent to `prompt_all or prompt` when calling, since `prompt` is used again later in `input_select_objects`: `prompt_all`, if given, is the message shown with the initial "Yes/no/select" query, whereas prompt is the message shown for each individual item if "select" is chosen. Maybe `prompt_all` is not a very descriptive name, but I couldn't quickly come up with anything better.

The modify CLI command also uses `input_select_objects`, but doesn't give `prompt_all`. The new argument allowed me to retain the initial prompt that the remove command used to print ("Really DELETE N files"), which however isn't suitable to be repeated for each item upon interactive selection.

beets/ui/commands.py
wisp3rwind: Ok, I agree that this is not really the current beets way. While I would really prefer some more structured output for albums, I shouldn't custom-build that into the remove command. Rather, the ui module should gain some unified facility to display albums. I don't have the time to revive the UI refactor, though, so I'll just restore the old behaviour :)

sampsyo: Yeah, that sounds cool—I like the idea of having an "orthogonal" option to display items grouped by album.

wisp3rwind:
i just force-pushed a change to the output to address @sampsyo's comment. Has anyone seen similar AppVeyor failures before?

beets/ui/__init__.py
sampsyo: Got it; I missed that other use of `prompt`! Thanks for clarifying. This totally makes sense.

docs/reference/cli.rst
```diff
@@ -230,10 +230,21 @@ remove
...
+touch the files on disk. To actually delete the files, use the ``-d`` flag.
+When the ``-a`` flag is given, the command operates on albums instead of
+individual tracks.
+
+When you run the ``remove`` command, it prints a list of all
+affected items in the library and asks for your permission before removing
+them. You can then choose to abort (type `n`), confirm (`y`), or interactively
+choose some of the items (`s`). In the latter case, the command will prompt you
+for every matching item or album and invite you to type `y` to apply the
+changes, `n` to discard them or `q` to exit and apply the changes selected up
```
sampsyo, line 243: "apply the changes" seems a little vague… maybe "remove the item"?

wisp3rwind: Thanks for spotting this, I copy&pasted this paragraph from the `modify` command and didn't properly adapt it.

wisp3rwind:
> And I can't help but notice the comment in there about determinism of the removal order. Is there any chance that's the culprit? But I note that, in that failure, it seems like both `path1` and `path2` are deleted?

That was also my first thought, but as you noticed, it doesn't seem to explain the failure. The issue about the determinism is unfortunate, but really fixing it would require extending our io stream mocks such that input could depend on previous CLI output. That would indeed be nice to have, but is more work than I'm willing to do right now.

> To get a little more clarity, maybe it would be helpful to try printing out those paths and asserting that they _do_ exist before removal is run…

I'll push a commit with some more debug prints, let's see...

wisp3rwind:
It appears that the AppVeyor failures were due to missing `syspath` conversions. The failures only occured on Python, 2.7 and 3.5 though, not 3.6 and 3.7, which I don't understand.

Moreover, there seem to be many missing `syspath` calls all over the tests, which is why there're many warnings like
```
C:\projects\beets\test\_common.py:186: DeprecationWarning: The Windows bytes API has been deprecated, use Unicode filenames instead
    if os.path.isdir(self.temp_dir):
```
in the AppVeyor logs. I'm going to open a new PR which adds these.

wisp3rwind:
Rebased on master. I think all of the comments above were addressed, so this should be good to go from my side. Would you mind having a brief look again, @sampsyo? Thanks!

> And I can't help but notice the comment in there about determinism of the removal order. Is there any chance that's the culprit? But I note that, in that failure, it seems like both `path1` and `path2` are deleted?

While that wasn't the issue for the previous test failure, I still changed the test to not depend on any ordering. That slightly decreases the power of the test, but should prevent it from exploding unexpectedly in the future.

sampsyo (approved):
Looks excellent! Thanks for the update.

PR #3671, opened 2020-07-13
first try at mocking get_work_by_id

### LINKED ISSUE #3332: parentwork: In tests, mock all MusicBrainz responses
I didn't notice this when we originally merged the parentwork plugin in #3279, but its tests rely on real communication with the MusicBrainz web service, i.e., they fail if there is no network connectivity. [This Travis job](https://travis-ci.org/beetbox/beets/jobs/558936634) is an example of a spurious failure caused by network interruptions.

We need to isolate these tests by mocking the MB queries so that no network traffic is every actually sent.

@dosoe, can you please look into this?

dosoe:
## Description

Fixes  #3332.

### DIFF
--- a/test/test_parentwork.py
+++ b/test/test_parentwork.py
@@ -28,7 +28,7 @@ from beetsplug import parentwork
 
 work = {'work': {'id': '1',
                  'work-relation-list': [{'type': 'parts',
-                                         'direction': 'backwards',
+                                         'direction': 'backward',
                                          'work': {'id': '2'}}],
                  'artist-relation-list': [{'type': 'composer',
                                            'artist': {'name':
@@ -37,7 +37,7 @@ work = {'work': {'id': '1',
                                                       'composer, random'}}]}}
 dp_work = {'work': {'id': '2',
                     'work-relation-list': [{'type': 'parts',
-                                            'direction': 'backwards',
+                                            'direction': 'backward',
                                             'work': {'id': '3'}}],
                     'artist-relation-list': [{'type': 'composer',
                                               'artist': {'name':
--- a/test/test_parentwork.py
+++ b/test/test_parentwork.py
@@ -145,12 +145,11 @@ class ParentWorkTest(unittest.TestCase, TestHelper):
         """Set up configuration"""
         self.setup_beets()
         self.load_plugins('parentwork')
+        musicbrainzngs.get_work_by_id = Mock(side_effect=mock_workid_response)
 
     def tearDown(self):
         self.unload_plugins()
         self.teardown_beets()
-
-    musicbrainzngs.get_work_by_id = Mock(side_effect=mock_workid_response)
 
     def test_normal_case(self):
         item = Item(path='/file', mb_workid='1', parentwork_workid_current='1')
--- a/test/test_parentwork.py
+++ b/test/test_parentwork.py
@@ -19,7 +19,6 @@ from __future__ import division, absolute_import, print_function
 
 import os
 import unittest
-from unittest.mock import patch
 from test.helper import TestHelper
 #from mock import Mock
 #import musicbrainzngs
@@ -138,7 +137,7 @@ class ParentWorkTest(unittest.TestCase, TestHelper):
         """Set up configuration"""
         self.setup_beets()
         self.load_plugins('parentwork')
-        self.patcher1 = patch('musicbrainzngs.get_work_by_id')
+        self.patcher1 = unittest.mock.patch('musicbrainzngs.get_work_by_id')
         self.mock_workid_response = self.patcher1.start()
         #musicbrainzngs.get_work_by_id = Mock(side_effect=mock_workid_response)
 
--- a/test/test_parentwork.py
+++ b/test/test_parentwork.py
@@ -147,7 +147,7 @@ class ParentWorkTest(unittest.TestCase, TestHelper):
         """Set up configuration"""
         self.setup_beets()
         self.load_plugins('parentwork')
-        musicbrainzngs.get_work_by_id = Mock(side_effect=mock_workid_response)
+        #musicbrainzngs.get_work_by_id = Mock(side_effect=mock_workid_response)
         self.patcher = patch('musicbrainzngs.get_work_by_id',
                              side_effect=mock_workid_response)
 
--- a/test/test_parentwork.py
+++ b/test/test_parentwork.py
@@ -27,6 +27,7 @@ from beets.library import Item
 from beetsplug import parentwork
 
 work = {'work': {'id': '1',
+                 'title': 'work',
                  'work-relation-list': [{'type': 'parts',
                                          'direction': 'backward',
                                          'work': {'id': '2'}}],
@@ -36,6 +37,7 @@ work = {'work': {'id': '1',
                                                       'sort-name':
                                                       'composer, random'}}]}}
 dp_work = {'work': {'id': '2',
+                    'title': 'directparentwork',
                     'work-relation-list': [{'type': 'parts',
                                             'direction': 'backward',
                                             'work': {'id': '3'}}],
@@ -46,6 +48,7 @@ dp_work = {'work': {'id': '2',
                                                          'composer, random'
                                                          }}]}}
 p_work = {'work': {'id': '3',
+                   'title': 'parentwork',
                    'artist-relation-list': [{'type': 'composer',
                                              'artist': {'name':
                                                         'random composer',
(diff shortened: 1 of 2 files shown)


dosoe:
I still can't get the tests to run locally, so unless it works quickly there will be a lot of commits

dosoe:
That one seems to work, is that what you had in mind?

jtpavlock:
Just want to add, it would be ideal to have these _in addition to_ the current integration tests (the ones that hit musicbrainz directly). That way we can test our local code quickly, but also be able to double-check every once in a while that musicbrainz didn't break our code either.

I plan on adding this info to the wiki at some point, but this code is for tests that hit external APIs, like musicbrainz:
```
@unittest.skipUnless(
        os.environ.get('INTEGRATION_TEST', '0') == '1',
        'integration testing not enabled')
```

dosoe:
I mean, the whole point of it was to not calling MB for the tests... 
So should I insert this snippet somewhere? Where?

dosoe:
Or should I just keep the tests as they were before, just add a second part with what I just did?

jtpavlock:
Any tests that have that snippet get run [weekly by github](https://github.com/beetbox/beets/blob/master/.github/workflows/integration_test.yaml). This functionality was added recently with https://github.com/beetbox/beets/issues/3645, so I understand the confusion as it was before #3332 was first created. 

Just keep the current tests, but also add yours (without the snippet).

dosoe:
Ok, done!

dosoe:
Apparently, you can't choose to mock a function in some tests and not in others. Now I have some kind of ugly hybrid mocking, that mocks the function in some cases but in others it just uses the real thing. I don't know if it is fine.

dosoe:
Hybrid mocking doesn't work. I tried to make two separate classes, one with mocking and one without but to no avail. Do you have an idea what's wrong?

dosoe:
Looking at the errors, it seems like the mocking doesn't work, I get network errors. That's strange because it happens only on the appveyor tests, not on the ci tests.

dosoe:
Finally, it works!

test/test_parentwork.py
```diff
@@ -85,14 +132,62 @@ def test_no_force(self):
...
+    def setUp(self):
+        """Set up configuration"""
+        self.setup_beets()
+        self.load_plugins('parentwork')
+
+    def tearDown(self):
+        self.unload_plugins()
+        self.teardown_beets()
+
+    musicbrainzngs.get_work_by_id = Mock(side_effect=mock_workid_response)
```
sampsyo, line 153: Doing this mocking "in class scope" will, unfortunately, *permanently* mock the function as soon as the class is loaded. Instead, there are a couple of options to "temporarily" mocking something:

- Use `patch` in `setUp` and `tearDown` as described in the `mock` docs: https://docs.python.org/3/library/unittest.mock.html#patch-methods-start-and-stop
- Use the `@patch` decorator: https://stackoverflow.com/a/23618258/39182
- Manually save the "old" `get_work_by_id` in `setUp` and restore it in `tearDown`.

test/test_parentwork.py
```diff
@@ -20,10 +20,50 @@
...
+def mock_workid_response(mbid, includes):
+    if mbid == '1':
+        return work
+    elif mbid == '2':
+        return dp_work
+    elif mbid == '3':
+        return p_work
+
 
 class ParentWorkTest(unittest.TestCase, TestHelper):
```
sampsyo, line 68: Maybe call this one `ParentWorkIntegrationTest`?

test/test_parentwork.py
dosoe: Ok I'll try it tomorrow (it's midnight here)

test/test_parentwork.py
dosoe: Done

test/test_parentwork.py
dosoe: if the class-level is the problem, could it be mocked at method-level?

dosoe: About your first example: if I use ```patch``` in ```setUp``` and ```tearDown``` , where do I control in which method I use the mock and in which I use the real one? I can't find it explained in the ```mock``` docs. 
I thought about this "hybrid" mocking, where I mock the functions for the arguments I use for the mock tests and I just keep the real function for all other arguments (like here https://github.com/beetbox/beets/pull/3671/commits/dba2e2320da48a653aaebcc7cb61b982fe867f99), could that be an option?

sampsyo: If you set something up in `setUp` and tear it down in `tearDown`, it will be "in effect" for all test invocations in that class. If you need a test that does not have that fixture, you can put it in a different class (such as the two classes you have currently defined).

I think that's likely to be simpler/easier to predict than deciding whether to do "mock" or "real" requests based on the argument…

dosoe: Does it correspond more to what you have in mind now @sampsyo ?

sampsyo:
Yes, this is getting there! This now sets up the mock in `setUp`; it now needs to tear it down in `tearDown` so the mock does not affect other tests. One way to do this is just to keep the old value around as an instance variable and assign it back in `tearDown`. Another is to use the `patch` functionality of the mock library:
https://docs.python.org/3/library/unittest.mock.html#patch-methods-start-and-stop

dosoe:
I'm trying to follow the example given in your link: 
```
class ParentWorkTest(unittest.TestCase, TestHelper):
    def setUp(self):
        """Set up configuration"""
        self.setup_beets()
        self.load_plugins('parentwork')
        self.patcher1 = unittest.mock.patch('musicbrainzngs.get_work_by_id')
        self.mock_workid_response = self.patcher1.start()

    def tearDown(self):
        self.unload_plugins()
        self.teardown_beets()
        self.patcher1.stop()

    def mock_workid_response(self, mbid, includes):
        if mbid == '1':
            return work
        elif mbid == '2':
            return dp_work
        elif mbid == '3':
            return p_work
```
Who is what in here? I have my function ```mock_workid_response``` that is meant to replace ```musicbrainzngs.get_work_by_id```. 
First question: ```mock_workid_response``` used to be a function, I changed it to a method to follow your example as closely as possible, is it necessary (and does it change anything)?
Second question: In the example, a class is mocked, here I want to mock a function (or a method), what needs to change? It's not clear to me.

sampsyo:
Thanks for the reminder, stalebot! @dosoe, to answer your most recent questions:

- No, a function or a method should not matter here. It can be a function because `self` is unused in this case.
- I don't think anything different needs to happen to mock a function rather than a class method.

Do you recall where the next steps are here?

dosoe:
Yes! I will look into it (it will take some time to get into it again) once
I have finished my thesis defence (next monday).

Le mar. 8 déc. 2020 à 13:42, Adrian Sampson <notifications@github.com> a
écrit :

sampsyo:
Wow! Good luck with the defense! :tada:

dosoe:
Hi @sampsyo ! I now have a working version, minus the style problems. It basically reverts to an older commit where I create a function externally and then use it to mock ```musicbrainzngs.get_work_by_id``` . Is that fine in your book?

test/test_parentwork.py
```diff
@@ -85,14 +134,72 @@ def test_no_force(self):
...
 
 
+class ParentWorkTest(unittest.TestCase, TestHelper):
+    def setUp(self):
+        """Set up configuration"""
+        self.setup_beets()
+        self.load_plugins('parentwork')
+        #self.patcher1 = unittest.mock.patch('musicbrainzngs.get_work_by_id')
+        #self.mock_workid_response = self.patcher1.start()
+        musicbrainzngs.get_work_by_id = Mock(side_effect=mock_workid_response)
```
sampsyo, line 152: This is close! The only trouble is that this will be _permanent_. That is, it's changing what `musicbrainzngs.get_work_by_id` points to when the test gets set up, and it never changes it back when the test is finished. So any tests that run after this will see the mock instead of the original function.

I think you want this:

```py
self.patcher = patch('musicbrainzngs.get_work_by_id', side_effect=mock_workid_response)
```

And then, in `tearDown`:

```py
self.patcher.stop()
```

to clean up.

sampsyo (commented):
Thanks for coming back to this! Here's one thought within.

test/test_parentwork.py
dosoe: Done!

dosoe:
This one should work. It is ready for merging?

sampsyo:
Looks about right to me! Thank you for seeing this through!!!

PR #3669, opened 2020-07-13
Unignore flake8 E123, E126, E305, E741, W504

arogl:
## Description

Un-ignore flake8 errors that require more than zero or minimal change to the codebase.

See the [the discourse discussion ](https://discourse.beets.io/t/updating-the-errors-we-ignore-with-flake8/1393)

Uploading changes based on PR #3666 for discussion

I have **ONLY** updated the main branch not the tests as yet

Could use some help to put change back on my feature branch.

PR #3666, opened 2020-07-12
Unignore flake8 E221, E226, E242, E704

jtpavlock:
## Description

Un-ignore flake8 errors that required zero or minimal change to the codebase.

See [the discourse discussion ](https://discourse.beets.io/t/updating-the-errors-we-ignore-with-flake8/1393)

PR #3673, opened 2020-07-14
Unignore flake8 E123, E126, E305, E741, W504

arogl:
Un-ignore flake8 errors that require more than zero or minimal change to the codebase.

See the the [discourse discussion](https://discourse.beets.io/t/updating-the-errors-we-ignore-with-flake8/1393)

Uploading changes based on PR #3666 and #3669 for discussion

## To Do

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -952,8 +952,8 @@ class Database(object):
     def load_extension(self, path):
         """Load an SQLite extension into all open connections."""
         if not self.supports_extensions:
-            raise ValueError(
-                    'this sqlite3 installation does not support extensions')
+            raise ValueError('this sqlite3 installation does not '
+                             'support extensions')
 
         self._extensions.append(path)
 
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -30,6 +30,7 @@ from beets.util import py3_path
 from beets.dbcore import types
 from .query import MatchQuery, NullSort, TrueQuery
 import six
+
 if six.PY2:
     from collections import Mapping
 else:
@@ -186,6 +187,7 @@ class LazyConvertDict(object):
 
 
 # Abstract base for model classes.
+
 
 class Model(object):
     """An abstract object representing an object in the database. Model
@@ -525,7 +527,7 @@ class Model(object):
                 tx.mutate(
                     'DELETE FROM {0} '
                     'WHERE entity_id=? AND key=?'.format(self._flex_table),
-                    (self.id, key)
+                    (self.id, key),
                 )
 
         self.clear_dirty()
@@ -547,12 +549,11 @@ class Model(object):
         self._check_db()
         with self._db.transaction() as tx:
             tx.mutate(
-                'DELETE FROM {0} WHERE id=?'.format(self._table),
-                (self.id,)
+                'DELETE FROM {0} WHERE id=?'.format(self._table), (self.id,)
             )
             tx.mutate(
                 'DELETE FROM {0} WHERE entity_id=?'.format(self._flex_table),
-                (self.id,)
+                (self.id,),
             )
 
     def add(self, db=None):
@@ -598,8 +599,9 @@ class Model(object):
         # Perform substitution.
         if isinstance(template, six.string_types):
             template = functemplate.template(template)
-        return template.substitute(self.formatted(for_path),
-                                   self._template_funcs())
+        return template.substitute(
+            self.formatted(for_path), self._template_funcs()
+        )
 
     # Parsing.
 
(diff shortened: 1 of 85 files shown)


arogl:
Sorry for the spamming

beets/dbcore/db.py
```diff
@@ -1039,12 +1039,11 @@ def _fetch(self, model_cls, query=None, sort=None):
...
-                model_cls._flex_table,
-                model_cls._table,
-                where or '1',
-            )
-        )
+            """.format(model_cls._flex_table,
+                       model_cls._table,
+                       where or '1',
+                       )
+                    )
```
sampsyo, line 1046: Weird—I think this conformant indentation is now _less_ clear, somehow? Maybe it would look less strange if we just put the closing parentheses right after the `'1'` instead of on their own lines.

test/test_logging.py
```diff
@@ -39,14 +39,14 @@ def test_logging_management(self):
         self.assertNotEqual(l1, l6)
 
     def test_str_format_logging(self):
-        l = blog.getLogger("baz123")
+        lteststrlogging = blog.getLogger("baz123")
```
sampsyo, line 42: Maybe `log` would be a clearer name for this variable?

test/test_ui.py
```diff
@@ -1238,63 +1238,69 @@ def tearDown(self):
         self.teardown_beets()
 
     def test_base(self):
-        l = self.run_with_output(u'ls')
-        self.assertEqual(l, u'the artist - the album - the title\n')
+        ltestbase = self.run_with_output(u'ls')
```
sampsyo, line 1241: Perhaps a better name for this variable would be `out` or something to indicate that it's the output log from the command.

test/test_ui_commands.py
```diff
@@ -96,10 +96,10 @@ def setUp(self):
     def tearDown(self):
         self.io.restore()
 
-    def remove_keys(self, l, text):
+    def remove_keys(self, lrm, text):
```
sampsyo, line 99: Perhaps a better name for this argument would be `lst` (the `l` is short for "list").

jtpavlock:
Regarding the indentation changes, I tend to agree that they seem a little strange. For comparison, I ran `db.py` through the [black](https://github.com/psf/black) python code formatter, and it chose some different indentations.

Here are some alternate relevant diffs that were mentioned

```diff
             """.format(
-                model_cls._flex_table,
-                model_cls._table,
-                where or '1',
-            )
+            model_cls._flex_table, model_cls._table, where or "1",
         )
```

and 

```diff
-            num = int(command_output([
-                '/usr/sbin/sysctl',
-                '-n',
-                'hw.ncpu',
-                ]).stdout)
+            num = int(command_output(["/usr/sbin/sysctl", "-n", "hw.ncpu",]).stdout)
```

Perhaps, it could be useful to switch to using black as a code formatter entirely? Then we wouldn't have to worry about these linting errors or correcting user's, as all the code would be unified.

arogl:
Have some time to come back to this. 
I reformatted the entire db.py file using [black](https://github.com/psf/black) for comparison with these options:
--line-length 79
--skip-string-normalization

I can run the entire code base through black if we wish to compare and have a conversation about a formatter and the default settings.
If we are going to python3 only support we could reformat the entire code base in another branch.

I'll tackle the variable names in the tests later today. I was reluctant to change those as they were tests. This depends how conformant we wish to be.

jtpavlock:
I was just using black as a comparison, not saying we should apply it completely or anything. Whether or not to use black probably deserves it's own discourse discussion. I imagine if we do switch though, these flake8 errors would go away as a side effect.

arogl:
> 
> 
> I was just using black as a comparison, not saying we should apply it completely or anything. Whether or not to use black probably deserves it's own discourse discussion. I imagine if we do switch though, these flake8 errors would go away as a side effect.

I had done all of the indenting manually.

The black format would fix it once if we accept the changes.

I still have to redo the variables

Now I need to work out how to fix the extra commit I did with the fetch of master :cold_sweat:

jtpavlock:
Yeah the variables are formatting agnostic, so that's good. Perhaps that could go in its own pull request while we decide what to do about the more sweeping formatting changes?

jtpavlock:
Also, [here's a great article on adjusting some commits](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History). I can't believe I used git for years before I found out about `git rebase -i`

jtpavlock:
Sorry for spamming PR, but I [started a discussion on discourse about black](https://discourse.beets.io/t/switching-to-using-a-universal-code-formatter/1431).

PR #1970, opened 2016-04-29
Add pydocstyle checking

jackwilsdon:
Sadly we have to add a new tox environment and test to the travis matrix for this, as we can't easily integrate this with flake8.

As we currently stand (using the default configuration), there are a large number of errors output by [`pydocstyle`](https://github.com/PyCQA/pydocstyle), although I'm sure we can fine tune it to suit our needs.

PR #3677, opened 2020-07-19
Add docstring checks to flake8

jtpavlock:
##  Description 

Adds [flake8-docstrings ](https://pypi.org/project/flake8-docstrings/) to lint for docstrings with the [google convention](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings) as discussed in #1970 

Currently, it ignores all errors on a per-file basis. This is to allow us to incrementally fix the errors and enable it for each file. I also described a method to upgrade each file in stages in `setup.cfg`

Stage 1: Fix current docstrings for each file. Then you can change `*.py:D` to `*.py:D1` for the specific file to now only ignore errors about mixing docstrings.
Stage 2: Add any missing docstrings. Then completely remove the mention of the file from `setup.cfg`

This also prevents new files from ignoring these errors.


sampsyo:
Great. Out of curiosity, have you taken a look yet at how overwhelming the errors are for your average beets source file? I'd be curious how _far_ out of compliance we are. :smiley:

jtpavlock:
No errors:
https://pastebin.com/7yqtn2GL

`wc -1` = 4914

Ignore missing docstring:
https://pastebin.com/HJUD9aza

`wc -l` = 2049

A lot of it seems to be missing docstrings and minor mis-formatting of the current docstring. I wonder if errors like `D415 First line should end with a period, question mark, or exclamation point` can be fixed via something like sed...

sampsyo:
Thanks!!

Yeah, we did discuss the "first line" thing (D415) in https://github.com/beetbox/beets/pull/1970#issuecomment-215824260. As @jackwilsdon points out, it looks like [PEP 257](https://www.python.org/dev/peps/pep-0257/) (while it's a bit ambiguous and doesn't say it outright) does seem to recommend something that I find somewhat silly… namely, that this is an OK docstring:

```py
"""This is a function.

Here are some more details.
"""
```

But this is not:

```py
"""This is a somewhat longer docstring whose first sentence happens to
exceed 72 characters.

And here are even more details.
"""
```

That is, the first sentence is not even allowed to wrap! It seems really hard to cram those synopses into the first line of characters, especially when some indentation is involved. I would be interested in seeing whether the Python community more broadly _actually_ thinks this rule is a good idea and, if so, why…

jtpavlock:
So the [PEP](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) actually mentions this

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line.

And I noticed both `numpy` and `google` recommend the same, so it does seem there's some sort of continuity there.

I don't see any restriction about it being 72 characters, though. There is always the option of increasing our line-length too. For example, [black](https://github.com/psf/black) uses 88 characters. After all, [PEP says it best](https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds).

jtpavlock:
Fixed conflicts and added the convention to CONTRIBUTING.rst. Anything else you think needs to be added?

jtpavlock:
Hmm, are there perhaps any sticking points to hash out some more before this can be merged?

sampsyo:
Thanks for continuing to shepherd this! Honestly, I'm still a little confused about D415 ("First line should end with a period, question mark, or exclamation point"). The style guides I've looked at all talk about making the first _sentence_ end with a period (and then two blank lines), which seems like a great idea! But I have not seen anyone justifying why it's a good idea for this first sentence to be so short that it is "unwrapped." I feel a little uncomfortable with just enforcing that recommendation if there's not a clear reason for it… especially because so many of our docstrings would apparently require human intervention to rephrase them so that the first sentence is (much) shorter. Does that hesitation make sense?

jtpavlock:
I think that comes from [google's style guide](https://google.github.io/styleguide/pyguide.html#381-docstrings):

> A docstring should be organized as a **summary line (one physical line)** terminated by a period, question mark, or exclamation point, followed by a blank line, followed by the rest of the docstring starting at the same cursor position as the first quote of the first line.

Which likely takes inspiration from the PEP I quoted above:

> Multi-line docstrings consist of a summary **line** just like a **one-line docstring**, followed by a blank line, followed by a more elaborate description. The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line.

I think the verbiage here is that it describes the summary line as a single _line_ and not necessarily plural _lines_. I'm curious where do you get the "sentence" description from?

If this is a sticking point, and I agree it would take a lot of manual adjustment (which I'm in no rush to do), then I am more than happy to ignore that error to get this through.

sampsyo:
OK, I think I'm starting to be convinced! I'm just unfamiliar with the convention and it immediately struck me as hard to follow. I think it is probably OK to enable this check for new files, although it will take some creativity to come up with a plan for enforcing it on existing files (and doing the necessary manual English rewriting work).

jtpavlock:
> it will take some creativity to come up with a plan for enforcing it on existing files (and doing the necessary manual English rewriting work).

Agreed. Hopefully, over time a good bit can be naturally adjusted as we go back to old methods or classes. We may have to manually enforce some of this for new functions as to not dig ourselves deeper into the hole until that specific file is compliant and taken off the ignore list.

> I think it is probably OK to enable this check for new files

So, is that a green light to merge? :smile:

sampsyo:
Sure, sounds good to me. 💚

PR #3678, opened 2020-07-21
Add windows to github action

### LINKED ISSUE #3613: feat(ci): integrate Windows
### Use case

From https://github.com/beetbox/beets/pull/3602#issuecomment-639520863

We'd like to get rid of AppVeyor and in favor of building Windows within GitHub Actions.

### Solution

Get `windows-latest` part of the matrix that GitHub Actions runs. Currently seeing encoding errors from `tox` when running.

jtpavlock:
## Description

Drops appveyor in favor of using github actions for windows CI

Fixes #3613 

[Currently blocked until python2.7 support is dropped](https://github.com/beetbox/beets/issues/3613#issuecomment-661413290)

Once python2.7 support is dropped, this should be able to be merged as-is


sampsyo:
I still want to do this and would love to keep it open.

jef:
> Once python2.7 support is dropped, this should be able to be merged as-is

Do you mean with beets or Python? Hasn't Python already dropped support? And with beets, what are the requirements for this to happen?

sampsyo:
Alas, beets's side. Basically, we can do this whenever we want. We just need to bite the bullet, make sure people know it's coming, and coordinate with various packagers. We're long overdue and we should absolutely do it.

jef:
> coordinate with various packagers

What does this look like?

I suppose I'm asking because I'd like to help in anyway I can!

jef:
@jtpavlock, let's remove `2.7` in the workflow for prep. As well as the appveyor config. What do you think?

wisp3rwind:
Do we want to support Python 2 for the next release? That might be nice for people who are still stuck on 2.7 (do they exist?). Then maybe we shouldn't cut CI just yet.

sampsyo:
Yes. Basically I think we should ship this last version with Python 2 support. As part of the announcement, we can include lots of warnings about how this is the last one you can expect before moving to 3.

jef:
> for the next release

When is this happening? It could be also interesting to have a cadence to releases as well when we announce the Python 2 deprecation. Perhaps we can make some workflows around that!

sampsyo:
Yes, I would love to! The basic answer is that our current release process is "@sampsyo is the bottleneck, and someday he finds time," which is not great. I would _love_ to have an actual release cadence and to empower everyone to do it.

A few discussions have started along these lines, but making a concrete decision would be awesome:
https://discourse.beets.io/t/should-we-adopt-the-git-flow-branching-model/1409/5?u=adrian

And there is ample evidence of me being like "I know I need to do this, and I promise to do it soon" ad nauseam, which is proof that we need a new system. To my great shame, for instance:
https://discourse.beets.io/t/what-is-the-timeframe-for-the-v1-5-0-release/1145/3?u=adrian

wisp3rwind:
Not stale, just waiting for a release.

sampsyo:
Please keep this alive.

PR #3679, opened 2020-07-21
Adjust wording of CONTRIBUTING.md

jtpavlock:
Wording seems a little off. I don't think "By following" should be repeated twice in a row


sampsyo:
Indeed. Sounds good!

PR #3681, opened 2020-07-21
introduce snapcraft.yaml

sergiusens:
## Description

This change introduces a snapcraft.yaml that would be handy for publishing to the Snap Store for Linux users. I am currently the publisher of the snap but if you want, I wouldn't mind transferring the ownership to you and from a _snap_ maintenance point of view I could remain as a collaborator of this snap if desired or warranted.

## To Do

### DIFF
--- a/docs/guides/main.rst
+++ b/docs/guides/main.rst
@@ -22,9 +22,7 @@ Beets works on `Python 2.7`_ and Python 3.4 or later.
   official package (`Debian details`_, `Ubuntu details`_), so try typing:
   ``apt-get install beets``. But the version in the repositories might lag
   behind, so make sure you read the right version of these docs. If you want
-  the latest version, you can get everything you need to install with pip
-  as described below by running:
-  ``apt-get install python-dev python-pip``
+  the latest version, you can use the `snap <snap_>`_.
 
 * On **Arch Linux**, `beets is in [community] <Arch community_>`_, so just run ``pacman -S
   beets``. (There's also a bleeding-edge `dev package <AUR_>`_ in the AUR, which will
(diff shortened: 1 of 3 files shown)


docs/guides/main.rst
```diff
@@ -22,9 +22,7 @@ Beets works on `Python 2.7`_ and Python 3.4 or later.
   official package (`Debian details`_, `Ubuntu details`_), so try typing:
   ``apt-get install beets``. But the version in the repositories might lag
   behind, so make sure you read the right version of these docs. If you want
-  the latest version, you can get everything you need to install with pip
-  as described below by running:
-  ``apt-get install python-dev python-pip``
+  the latest version, you can use the `snap <snap_>`_.
```
jtpavlock, line 25: pip still works... maybe worth noting both options

snap/snapcraft.yaml
```diff
@@ -0,0 +1,35 @@
...
+  beets:
+    source: .
+    override-pull: |
+      snapcraftctl pull
+      version="$(git describe --always --tags| sed -e 's/^v//;s/-/+git/;y/-/./')"
+      [ -n "$(echo $version | grep "+git")" ] && grade=devel || grade=stable
+      snapcraftctl set-version "$version"
+      snapcraftctl set-grade "$grade"
+    plugin: python
+    python-packages:
```
jtpavlock, line 30: are these supposed to be the dependencies of the project? This list seems different than what's in `setup.py`

jtpavlock:
Would you mind also explaining the benefit of having a snap vice the current installation methods for linux (pip or distro repo)?

docs/guides/main.rst
sergiusens: Can I reply with more of a meta question? pip install works for all platforms, so should it not be more of a generic installation method?

snap/snapcraft.yaml
sergiusens: These are extras_require in setup.py, for which there is no declarative way in Snapcraft to specify yet, but I can migrate to that once it is supported. I only considered some plugins and not all, perhaps I should just consider supporting all plugins.

sampsyo: Interesting! It's hard to say what to do if there's no room for optional dependencies… another option would be to include _no_ plugin dependencies and just leave it at that for now?

Issue #2544, opened 2017-05-03
Beets not picking up Discogs to tag


macdalor:
### Problem

I've just discovered this BRILLIANT platform and I have to say I'm quite impressed! Maybe I am doing something wrong and would appreciate your advice.

I am trying to get my mainly 80's & 90's electronic music collection re-tagged by Discogs as it is the best/most complete platform for my needs. The important point for me is to get the correct original year for each track.

I have enabled the disco's plugin within beets and have also assigned its source_weight to 0.0 as advised in the following post https://github.com/beetbox/beets/issues/2391 but unless I am missing something it's not picking it up although the tracks are in discogs.

Running this command in verbose (`-vv`) & singleton mode:

```sh
$ beet -vv import -s /path/to/my/music/files
```

Led to this problem:

```
Sending event: before_choose_candidate
Correcting track tags from:
     - Boomshanka - Do You Have the Power! (That Side mix)
To:
    Dig - Reinvent Yourself
URL:
    https://musicbrainz.org/recording/14e9d518-2ecd-4fe3-9bba-fec61e6ac4fd
(Similarity: 51.0%) (title, length)
Apply, More candidates, Skip, Use as-is, Enter search, enter Id, aBort? e
Artist: boomshanka
Track: Do You Have The Power?
Item search terms: boomshanka - Do You Have The Power?
Sending event: trackinfo_received
Sending event: trackinfo_received
Sending event: trackinfo_received
Sending event: trackinfo_received
Sending event: trackinfo_received
Found 5 candidates.
Sending event: before_choose_candidate
Finding tags for track " - Boomshanka - Do You Have the Power! (That Side mix)".
Candidates:
1. Boomshanka - Do What U Wanna Do (13.1%) (title, length)
2. Boomshanka - First Thing Tomorrow (10.1%) (title, length)
3. Boomshanka - Peter North (9.5%) (title, length)
4. Boomshanka - Full Circle (4.7%) (title, length)
5. Boomshanka - BFD (4.1%) (title, length)
# selection (default 1), Skip, Use as-is, Enter search, enter Id, aBort? 1
Correcting track tags from:
     - Boomshanka - Do You Have the Power! (That Side mix)
To:
    Boomshanka - Do What U Wanna Do
URL:
    https://musicbrainz.org/recording/e510dbe0-9882-4eaa-8aca-d8f8c0847e86
(Similarity: 13.1%) (title, length)
[...]

PR #3682, opened 2020-07-22
Allow Discogs Plugin to respond during Singleton Import Mode

benrbray:
## Description

The purpose of this change is to allow the Discogs plugin to respond as expected during singleton import mode.  Before, the Discogs plugin did not run at all in singleton mode, leading to some confusion, as in #2544 .  Changes to the Discogs plugin include:

* Implemented `DiscogsPlugin.item_candidates`, allowing the plugin to provide track candidates during singleton import.
* Implemented `DiscogsPlugin.track_for_id`, allowing the plugin to respond to Id queries of the form `release_id/position`.  For example, `5240807/B2` refers to position B2 of release 5240807.

This change required a few minor changes to how plugin output is handled during import.  This change does not break existing plugin behavior, but allows plugins to return more information than they did before.

* `Plugin.tracks_for_id` and `Plugin.item_candidates` can now optionally return / yield a `TrackAlbumTuple`, containing both `TrackInfo` and `AlbumInfo` objects.  In the case of Discogs, this is useful because Discogs release IDs correspond to albums rather than individual songs.  As far as I can tell, it makes sense to associate the extra AlbumInfo with a track, even during singleton import mode.  (thereby making tracks modified in this manner non-singletons)
* Every `TrackMatch` candidate can optionally specify some `album_info`, with a default value of `None`.

## Use case

I would like to use beets to assign album information to a random assortment of songs which currently only have a title and artist.  The discogs database has information for most of these songs, while MusicBrainz does not.  The `--group-albums` options works OK for this use case, but presents albums as candidates rather than tracks themselves.  I prefer to use singleton mode to search for tracks directly, but still have the option to attach album info based on the track I've chosen.

## To Do

- [x] **Documentation.** I briefly searched, but didn't identify any appropriate places in the docs to mention these changes.  Feedback welcome.
- [x] **Changelog.** (Add an entry to `docs/changelog.rst` near the top of the document.)
- [x] **Tests.**  I added a few tests of the new Discogs plugin, but could not identify an appropriate way to test the other changes (other than the fact that existing plugin tests still pass).  Feedback welcome.


benrbray:
There appears to be an issue with the Python 2.7 tests.  Something to do with string encodings?  Input appreciated, as I am not familiar with Python 2.  Although, since [Python 2.7] has [reached end-of-life](https://pythonclock.org/), are Python 2.7 tests needed?

jtpavlock:
Hey! Thanks for tackling this. As far as the errors, what may help is converting the strings to unicode strings e.g. `u'string'` 

Also, it may make your life easier if you use `tox` to run the tests/linting locally. Check out our[ CONTRIBUTING.rst](https://github.com/beetbox/beets/blob/master/CONTRIBUTING.rst#running-the-tests) for more info.

sampsyo:
Hi! Thanks so much for getting this started! The discogs thing will be very useful.

Is there any chance I could convince you to break this up into two separate PRs? The extension to the discogs plugin to work for singletons should be fairly modular and straightforward. The change to allow album-based matching of tracks (i.e., the new `TrackAlbumTuple`) is much more complex and will need to be treated with care because it touches multiple sensitive parts of the beets core. We will need to think about the right way to do this. If we can consider these changes separately, that would be extremely useful to both.

benrbray:
Sure, I wouldn't mind splitting it into two PRs.  Is the best way to do that just to copy the changes into a new PR entirely, or is there a way to split this PR into multiple?

At the moment my additions to the discogs plugin rely on the ability of plugins to return an `TrackAlbumTuple`, since it seemed like the least invasive way to add that functionality.  Perhaps the same thing could be accomplished by instead changing how plugins interact with the import session, but I wasn't able to figure out a way to make the Discogs fix without touching files in the main beets repo.  Suggestions are welcome!

Do you think it's possible for the Discogs plugin to attach album info during a singleton import session?  Or do you think it will have to wait until after deciding which changes should be made to the plugin system?

sampsyo:
Great. If possible, I think the best place to start would be by _just_ implementing `item_candidates` in Discogs, and—again, if possible—just providing TrackInfo for individual tracks. Even if that's missing some information, that will provide a good foundation for understanding where we need to go from there.

There's no magic GitHub way to split a single PR into multiple, unfortunately, but maybe a good place to start here would be by just removing the core changes for now in this branch (and possibly opening a new PR with them based on this one).

Issue #1409, opened 2015-04-09
Use pathlib everywhere in beets


brunal:
Pathlib is awesome. It abstracts path the platform for path manipulation. It would allow us to have more robust code and get rid to all the calls to `beets.util.*_path()` functions. It could be extended to provide desired `repr()`.

I feel like working on it. It is *huge* though.

- Am I missing a delicate point that prevents us from using it
- Should it be a 1.3.12 milestone?

sampsyo:
I'm all for it. Our current path handling stuff is too fiddly.

I have one important concern: encodings. The core problem is that Unix paths need to be bytes and Windows paths need to be Unicode strings (more or less). We need to store paths in a database, so eventually they need to be represented internally as one or the other. Beets currently chooses byte strings, and on Windows the "true" paths are encoded with UTF-8 just for storage and uniform manipulation. I'm not quite clear on how pathlib would interact with this constraint.

I also don't know how pathlib handles long names on Windows (the issue [described here](https://bugs.python.org/issue18199)).

Maybe the right approach here would be to start with a small prototype, see how it goes on different platforms, and then attempt the full migration.

brunal:
Currently,
- all paths are byte strings
- when there is a path in the input command we encode it with `beets.util._fsencoding()`
- when we display a path we decode it with that same encoding
- `displayable_path()` does that operation
- `syspath()` does some ugly dark magic
- `bytestring_path()` ensures always gives a byte string path, no matter the input
Is that right? Am I forgetting something?

Python 3.2 introduced `os.fs{en,de}code()`: https://docs.python.org/3/library/os.html#os.fsencode
The source is pretty simple (`Lib/os.py:800`) and shares some traits with our functions. However `mbcs` encoding only triggers another error mode: `strict` instead of `surrogateescape`.

brunal:
- `surrogateescape` is not natively available in python 2
- pathlib expects unicode paths

I think the solution is to use python-future (https://github.com/PythonCharmers/python-future) which provides a `future.utils.surrogateescape` module.

sampsyo:
Yes, surrogate escaping is probably now the Pythonic way of doing this. A few thoughts/questions:

* How hard will it be to be to make pathlib use surrogate escaping on Python 2 via python-future?
* How will this work with people's existing databases? Shall we incrementally "upgrade" current bytestring paths to surrogate-escaped Unicode?
[...]

PR #3690, opened 2020-07-26
Always use syspath conversions

wisp3rwind:
## Description

As mentioned in https://github.com/beetbox/beets/pull/3670, our tests often don't use `syspath` conversions. Since the tests usually involve nice short file/directory names, there have not been any issues. However, on Python 3.5, which deprecated the bytes API on Windows, we get a lot of `DeprecationWarning`s like

```
C:\projects\beets\test\_common.py:186: DeprecationWarning: The Windows bytes API has been deprecated, use Unicode filenames instead
    if os.path.isdir(self.temp_dir):
```
see for example https://ci.appveyor.com/project/beetbox/beets/builds/34307016/job/7ju959dh8uyxwq0f

Deprecation never happened, since these interface became a new meaning with [PEP 529](https://www.python.org/dev/peps/pep-0529/). However, PEP 529 does not imply that we don't need `syspath` anymore, since `syspath` also the adds the magic prefix that allows long file names. Related: https://bugs.python.org/issue27731

This attempts to add the missing conversions (I searched the codebase for `os\.` and `shutil\.` to find these). In some cases, it appears that the actual code (i.e. not the tests) also lacked those conversions.

No need to review this yet, I mainly pushed this to run it on all CI platforms.

## To Do

- [ ] Observe behaviour on CI, fix issues.

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -379,7 +379,8 @@ class ImportSession:
         """Mark paths and directories as merged for future reimport tasks.
         """
         self._merged_items.update(paths)
-        dirs = {os.path.dirname(path) if os.path.isfile(path) else path
+        dirs = {os.path.dirname(path)
+                if os.path.isfile(syspath(path)) else path
                 for path in paths}
         self._merged_dirs.update(dirs)
 
@@ -876,7 +877,7 @@ class ImportTask(BaseImportTask):
         the file still exists, no pruning is performed, so it's safe to
         call when the file in question may not have been removed.
         """
-        if self.toppath and not os.path.exists(filename):
+        if self.toppath and not os.path.exists(syspath(filename)):
             util.prune_dirs(os.path.dirname(filename),
                             self.toppath,
                             clutter=config['clutter'].as_str_seq())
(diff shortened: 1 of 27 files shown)


jtpavlock:
Is this something that would be a side effect of implementing https://github.com/beetbox/beets/issues/1409 or is this unrelated.

wisp3rwind:
> Is this something that would be a side effect of implementing #1409 or is this unrelated.

As far as I understand it, #1409 would get rid of all/many of these explicit conversions. I'm not familiar with pathlib, though, so I can't really judge the extent of #1409. This PR is only about ensuring that our current path handling is always applied.

sampsyo:
Wow; thank you for tracking all this down!! This is a heroic effort.

To weigh in briefly on the pathlib thing: it is also not clear to me how much pathlib solves the encoding issue. In particular, the library's documentation does not entirely explain how much it endeavors to preserve the bytes-oriented nature of paths on Unix, which is critical to reliably representing any path. It's possible it just Does The Right Thing, but it needs a deeper exploration to understand if so & how.

wisp3rwind:
Not stale, I should really just have another look at it and mark the PR as ready.

wisp3rwind:
Split the changes into more commits for easier review and rebased on master. The commit which should be looked at in particular is `add missing syspath conversions (3/3, beets)`. It changes beets' core and as such would have the largest fallout in case of incorrect path handling. Maybe I'll actually try to come up with tests to show that the old version of the respective code is indeed wrong.

sampsyo:
Awesome! I looked over all the changes in that commit and didn't see any red flags there. Thank you for pursuing this!

wisp3rwind:
Still not stale...

wisp3rwind:
All commits here have now been moved to separate PRs:
- #4830
- #4848
- #4849
- #4850

PR #3694, opened 2020-07-27
Add black as an auto formatter

jtpavlock:
## Description

[See discussion](https://discourse.beets.io/t/switching-to-using-a-universal-code-formatter/1431/5)

* Adds black a universal and automatic code formatter
* black is now enforced via an optional pre-commit hook (see below)
* flake8 is now included in that hook
* `tox -e lint` now runs `pre-commit run -a` (see below)

To add this to your workflow, I recommend creating a [git hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) with [pre-commit](https://pre-commit.com/). Note that this is optional and you can just run the checks as normal with `tox -e lint` if you don't want the hooks in your git workflow. Another option, if you don't want to run the hooks for some reason on any particular commit, you can use `git commit -n [--no-verify]`

To install the hook run `pre-commit install`. This will now enforce black and flake8 prior to completing the commit. It will only check files that were commited, so the performance hit isn't an issue.

Example workflow for a compliant commit:
```bash
$ git commit
black....................................................................Passed
flake8...................................................................Passed
[black 0ab1ef45] good commit
 1 file changed, 1 insertion(+), 1 deletion(-)
```

Example for a commit that is not compliant with the lint checks:
```bash
$ git commit
black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted setup.py
All done! ✨ 🍰 ✨
1 file reformatted.

flake8...................................................................Passed
```
Note that since it was a formatting issue, black will automatically apply the changes. However, you will need to re-stage and attempt the commit again. These checks will occur prior to writing the commit message if you don't use `git commit -m`.

To run the linting commands normally on all files (does not require the `pre-commit install` command) run `pre-commit run -a`. This is now exactly what `tox -e lint` will do.

Edit: this would also significantly affect https://github.com/beetbox/beets/pull/3673 as a lot of those changes would no longer be required.
 

## To Do

- [ ] Documentation - add info to CONTRIBUTING.rst

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,8 @@
+default_stages: [commit]
+
+repos:
+  - repo: https://github.com/psf/black
+    rev: stable
+    hooks:
+      - id: black
+        language_version: python3
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,3 +6,13 @@ repos:
     hooks:
       - id: black
         language_version: python3
+
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: master
+    hooks:
+      - id: flake8
+        additional_dependencies: [
+        flake8-blind-except,
+        flake8-coding,
+        flake8-future-import,
+        ]
--- a/beet
+++ b/beet
@@ -19,5 +19,5 @@ from __future__ import division, absolute_import, print_function
 
 import beets.ui
 
-if __name__ == '__main__':
+if __name__ == "__main__":
     beets.ui.main()
--- a/beets/__init__.py
+++ b/beets/__init__.py
@@ -18,25 +18,27 @@ from __future__ import division, absolute_import, print_function
 import confuse
 from sys import stderr
 
-__version__ = u'1.5.0'
-__author__ = u'Adrian Sampson <adrian@radbox.org>'
+__version__ = u"1.5.0"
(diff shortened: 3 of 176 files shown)


jtpavlock:
Fixed conflicts and rebased off master.

Since there seems to be some crickets on this PR, I'll make a few comments.

This is an incredibly large PR, and likely impossible to review every change. Merging this basically surrenders our formatting to an external program. While every change may not always be perfect, I believe that overall it will be a net positive as it will increase readability in most cases and also completely removes any formatting frustrations for both reviewers and contributors.

As a side note, the current testing errors are due to https://github.com/beetbox/beets/issues/3708. We can be sure black doesn't change any actual code meaning as it creates an AST before and after formatting.

arogl:
I was looking at #3673 earlier and was going to close, as we are looking at using black as the formatter.

I was then going to re-look at the variable names as suggested by @sampsyo

arogl:
> As a side note, the current testing errors are due to #3708. We can be sure black doesn't change any actual code meaning as it creates an AST before and after formatting.

Are we not looking to drop anything lees than 3.6 when we go to the pathlib changes you are proposing via #3704

jtpavlock:
> Are we not looking to drop anything lees than 3.6 when we go to the pathlib changes you are proposing via #3704

I believe the current plan for v1.5.0 is to be the last release that supports python 2.7, 3.4 and 3.5. But, until then, we still have to deal with issues for those python versions.

Issue #3625, opened 2020-06-17
New Release


lovesegfault:
Beets master is approaching 1000 commits since the last release, and so I'm here to ask for a new release to be cut :)

I can help with testing if an alpha/preliminary release is cut

sampsyo:
Hello! This post has more context on why releases are slow to come:
https://discourse.beets.io/t/forming-a-beets-core-team/639

The synopsis is that the project organization is currently bottlenecked on me to do releases, and I have somewhat extreme time constraints that make that hard. I'm still looking for help with this sort of thing if anyone wants to step up! :smiley:

nektro:
> and I have somewhat extreme time constraints that make that hard

the last message here was Jun 17. have you had no time in the last 4 months to do allocate to this?

lovesegfault:
@nektro Please do not harangue maintainers.

Adrian maintains this wonderful piece of software for free, despite the fact that I'm sure his day-job keeps him quite busy. If things take time, it is what it is.

Trying to guilt him, or any OSS maintainer, into doing work on _your_ pace is wrong. Please reconsider your approach to collaborating.

PR #3700, opened 2020-07-30
Delete after convert

### LINKED ISSUE #2947: convert: In `auto` mode, respect `import.move` by deleting old files

SamuelCook:
## Description

Fixes #2947.

- [x] Documentation.
- [x] Changelog.

This PR builds upon the abandoned https://github.com/beetbox/beets/pull/3394 and implements the comment in that PR, namely, the ability to enable this functionality explicitly via a `delete_originals` option.


docs/plugins/convert.rst
```diff
@@ -111,6 +111,8 @@ file. The available options are:
   This option overrides ``link``. Only works when converting to a directory
   on the same filesystem as the library.
   Default: ``false``.
+- **delete_originals**: Transcoded files will be copied or moved to their destination, depending on the import configuration. By default, the original files are not modified by the plugin. This option deletes the original files after the transcoding step has completed.
```
jtpavlock, line 114: #2947 seems to suggest that currently if you have `import: move` enabled in your config, then the convert plugin will not "move" the converted files (and delete the originals). This PR seems to just add a flag on whether or not to delete the original files. 

Basically, does this PR change the behavior of this plugin to match this:

>  Transcoded files will be copied or moved to their destination, depending on the import configuration.

SamuelCook: Good question.

The original PR did what you say - if you have `import: move` enabled, the convert plugin deletes the original files. However, the comments in that PR pointed out that this may not be desirable - an example given was that when transcoding from a lossless format to lossy format, the user may wish to keep the original files for archiving. Removing the source files might be surprising to the user, and is irreversible.

I found that argument compelling. Imho, it makes sense that the `import: move` command should refer to what happens to the _transcoded_ files - they would either be copied or moved into the library in accordance with this setting.

In this PR, if the user wishes to move the transcoded files into the library and to remove the source files after conversion (as I do), it would be necessary to specify both `import: move: yes` and `convert: delete_originals: yes`. This extra parameter seems to me to be a suitable compromise between arriving at the desired functionality while not causing surprising data loss for users that don't want this behaviour.

jtpavlock: Thanks for the clarification! I agree, your approach makes the most sense.

beetsplug/convert.py
```diff
@@ -532,11 +534,15 @@ def convert_on_import(self, lib, item):
...
             # Change the newly-imported database entry to point to the
             # converted file.
+            source_path = item.path
             item.path = dest
             item.write()
             item.read()  # Load new audio information data.
             item.store()
 
+            if self.config['delete_originals']:
+                util.remove(syspath(source_path), False)
```
jtpavlock, line 544: this is a tiny nitpick, but the `syspath` call here isn't necessary since it's called internally with `util.remove`

SamuelCook: That was part of the original PR, and to be honest I didn't really look too closely at it. I'll fix this. 👍🏻

jtpavlock:
@SamuelCook just let us know when you're ready and I'll take one last look and merge

SamuelCook:
All done. I added a unit test and it took me a few attempts to make the linter happy, but I’m finished messing with it now. Thanks for reviewing this, I appreciate it 👍🏻

jtpavlock:
Merged 🎉 Thanks for taking the time to see this through proper.

SamuelCook:
Apologies if this is not the right place to ask, but do you have any rough idea of when you're looking to cut a release? Do you have a release schedule?

jtpavlock:
It's on the horizon, see https://github.com/beetbox/beets/issues/3625

PR #3703, opened 2020-07-31
Link check

### LINKED ISSUE #3699: Minor documentation correction:  correct id3.org url
https://github.com/beetbox/beets/blob/master/docs/faq.rst#L303
refers to:
http://www.id3.org/id3v2.4.0-structure
as a reference url for a copy of the ID3v2.4 standard documentation, but this returns a "Not found" error. I've found 2 possibilities for the replacement:
https://id3.org/id3v2.4.0-structure
(with adverts) or
https://github.com/id3/ID3v2.4/raw/master/id3v2.40-structure.txt
(without adverts)

jtpavlock:
## Description

Fixes broken links throughout the docs and adds an external link checker to the weekly integration test

Fixes #3699

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,6 +28,13 @@ extlinks = {
     'stdlib': ('https://docs.python.org/3/library/%s.html', ''),
 }
 
+linkcheck_ignore = [
+    r'https://github.com/beetbox/beets/issues/',
+    r'https://github.com/\w+$',  # ignore user pages
+    r'.*localhost.*',
+    r'https://www.musixmatch.com/',  # blocks requests
+]
+
 # Options for HTML output
 htmlhelp_basename = 'beetsdoc'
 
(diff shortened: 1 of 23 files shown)


jef (approved):

docs/conf.py
```diff
@@ -28,6 +28,13 @@
     'stdlib': ('https://docs.python.org/3/library/%s.html', ''),
 }
 
+linkcheck_ignore = [
+    r'https://github.com/beetbox/beets/issues/',
+    r'https://github.com/\w+$',  # ignore user pages
```
sampsyo, line 33: Seems like a good idea. In which case, maybe we should undo the one remaining change to the `:user:` links in the changelog?

sampsyo (approved):

docs/conf.py
jtpavlock: Well, it does lead to a broken user page. The motivation for ignoring user and issue links is that they come up so often and end up timing out the github connection altogether as we get rate limited.

sampsyo: Legit, that makes sense.

Issue #3708, opened 2020-08-01
rarfile4.0 dependency only works with python3.6+


jtpavlock:
Newest version of rarfile (4.0) only works with python3.6+ https://pypi.org/project/rarfile/

Our current python 2.7 & 3.5 tests fail because of this

The fix should be as easy as targeting < 4.0 in setup.py

As an aside, I've been thinking about setting up better dependency management + something like dependabot to hopefully prevent stuff like this

PR #3711, opened 2020-08-01
Fix rarfile 4.0 only supports python3.6+

jtpavlock:
## Description

Fixes https://github.com/beetbox/beets/issues/3708


jtpavlock:
Okay I improved the dependency stuff to only target <4 if you're on python <3.6. Should be good to go now

Issue #3614, opened 2020-06-05
feat(ci): inline annotations


jef:
### Use case

From https://github.com/beetbox/beets/pull/3602#issuecomment-639520863

We'd like to see inline annotations if builds fail (e.g., sphinx, flake8). Here are some examples:

- [flake8-action](https://github.com/marketplace/actions/flake8-action): this is good, but doesn't allow for custom files
- [sphinx-action](https://github.com/ammaraskar/sphinx-action): looks good as-is

### Solution

We'll have to build some actions to make this possible or contribute to current solutions (if they're active).

PR #3714, opened 2020-08-02
Add flake8 problem matcher

jackwilsdon:
## Description

This PR implements the flake8 half of #3614 using a [problem matcher](https://github.com/actions/toolkit/blob/1cc56db0ff126f4d65aeb83798852e02a2c180c3/docs/problem-matchers.md), giving us nice inline errors like this on PRs:

inline error

I've had to disable `PY_COLORS` here, as the ANSI escape sequences were tripping up the regex used in the matcher.

## To Do

### DIFF
--- a/.github/flake8-problem-matcher.json
+++ b/.github/flake8-problem-matcher.json
@@ -4,12 +4,11 @@
             "owner": "flake8",
             "pattern": [
                 {
-                    "regexp": "^(.*?):(\\d+):(\\d+): ([^ ]+) (.*)$",
+                    "regexp": "^(.*?):(\\d+):(\\d+): (.*)$",
                     "file": 1,
                     "line": 2,
                     "column": 3,
-                    "code": 4,
-                    "message": 5
+                    "message": 4
                 }
             ]
         }
--- a/.github/flake8-problem-matcher.json
+++ b/.github/flake8-problem-matcher.json
@@ -1,0 +1,17 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "flake8",
+            "pattern": [
+                {
+                    "regexp": "^(.*?):(\\d+):(\\d+): ([^ ]+) (.*)$",
+                    "file": 1,
+                    "line": 2,
+                    "column": 3,
+                    "code": 4,
+                    "message": 5
+                }
+            ]
+        }
+    ]
+}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -65,9 +65,6 @@ jobs:
   lint:
     runs-on: ubuntu-latest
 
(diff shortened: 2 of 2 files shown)


jackwilsdon:
Hmm, I've noticed that even though we're setting the `code` group, GitHub isn't actually showing the error code anywhere. Is it worth us compounding this back into the message and dropping the code completely? We'd get something like this instead:

inline error

jtpavlock:
I just want to double check this won't be affected by the changes in https://github.com/beetbox/beets/pull/3694? In particular, the output from `pre-commit run -a` would display flake8 errors as such:

```
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

beetsplug/chroma.py:1:1: FI10 __future__ import "division" missing
beetsplug/chroma.py:1:1: FI11 __future__ import "absolute_import" missing
beetsplug/chroma.py:1:1: FI13 __future__ import "print_function" missing
```

jackwilsdon:
Provided we output `::add-matcher::.github/flake8-problem-matcher.json` in one of our steps, it should continue to work with flake8 output no matter where it is:

image

jtpavlock:
> Hmm, I've noticed that even though we're setting the `code` group, GitHub isn't actually showing the error code anywhere. Is it worth us compounding this back into the message and dropping the code group completely? We'd end up with the code in the message like this:

I like it with the error code, but probably not worth it if it's too much trouble

jackwilsdon:
Changed it to add the error code back in - was a nice and easy fix :+1:

jtpavlock:
will anything change if you do `[^\s]` rather than `[^ ]` ? (sorry for not asking this before you changed it 😄 )

jackwilsdon:
I don't think so — the match is working correctly but GitHub just doesn't display the code anywhere in their UI.

I'll give it a shot in a bit though to see if anything interesting happens.

jtpavlock:
> the match is working correctly but GitHub just doesn't display the code anywhere in their UI.

Ah okay, I see what you're saying now. Well this looks cool! I'm excited to see it in action. I think you should feel free to merge when you're ready 👍

jackwilsdon:
Awesome, thanks!

sampsyo:
Wow, this is awesome! I didn't know this was possible. This will be such a time saver!!

jef:
Wow... This was a lot easier than expected. Nice work guys.

jackwilsdon:
There may be some beneifts to using custom actions but this seemed like the path of least resistance for the time being.

jef:
That makes sense to me. Much easier just to put in this way rather than wrap it up another action.

We could've used local actions, but IMO that would be even worse (readability) I suppose.

Nice work!

PR #3716, opened 2020-08-03
Add sphinx problem matcher

### LINKED ISSUE #3614: feat(ci): inline annotations

jackwilsdon:
## Description

Fixes #3614. This PR implements the remaining sphinx half of #3614 using another problem matcher, giving us more nice inline errors like this on PRs:

inline error

Oddly enough, it seems like the errors reported by sphinx are for the first line of the paragraph containing the error, instead of the actual line of the error:

weird error positioning

I don't know if there's much we can do about this, as it looks like it's a problem with sphinx itself.

## To Do

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -59,6 +59,9 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install tox sphinx
 
+      - name: Add problem matcher
+        run: echo "::add-matcher::.github/sphinx-problem-matcher.json"
+
       - name: Build and check docs using tox
         run: tox -e docs
 
--- a/.github/sphinx-problem-matcher.json
+++ b/.github/sphinx-problem-matcher.json
@@ -1,0 +1,18 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "sphinx",
+            "pattern": [
+                {
+                    "regexp": "^Warning, treated as error:$"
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -59,6 +59,9 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install tox sphinx
 
+      - name: Add problem matcher
```
jtpavlock, line 62:
```suggestion
      - name: Add sphinx problem matcher
```

jackwilsdon: I thought it was clear enough that the problem matcher was for the current job (which is just `test-docs` here), but I'm happy enough to make this change. I guess we'd want it changing for flake8 too?

jtpavlock: That's a good point, I don't really care either way tbh

PR #3731, opened 2020-08-23
fix(plugin): `subsonicupdate` rest call

jef:
## Description

~Subsonic still isn't authenticating correctly.~

~I'm having some problems understanding how config works and how to overwrite it in the tests. I'm hoping I could get some help!~

Fixes REST method from `POST` to `GET`. Also includes some more tests and exception handling if the server isn't available or if Subsonic returns an error.

## To Do


beetsplug/subsonicupdate.py
```diff
@@ -40,60 +39,56 @@ def __init__(self):
         super(SubsonicUpdate, self).__init__()
 
         # Set default configuration values
-        config['subsonic'].add({
+        self.config.add({
```
sampsyo, line 42: The reason this was previously `config['subsonic']` instead of `self.config` is so that the configuration could show up in a shared `subsonic:` section, hypothetically to be used by other plugins that also want Subsonic configuration. Rather the same way that the `mbcollection` plugin uses a section called `musicbrainz`:
https://github.com/beetbox/beets/blob/eb6bbaeee8c2cb9ac143cbdc41be0f4c62d1fb94/beetsplug/mbcollection.py#L61

beetsplug/subsonicupdate.py
```diff
@@ -40,60 +39,56 @@ def __init__(self):
...
             'url': 'http://localhost:4040',
         })
 
-        config['subsonic']['pass'].redact = True
+        self.config['password'].redact = True
         self.register_listener('import', self.start_scan)
 
     @staticmethod
-    def __create_token():
+    def __create_token(self):
```
sampsyo, line 52: [Static methods](https://docs.python.org/3/library/functions.html#staticmethod) do not get a `self` parameter.

test/test_subsonic.py
```diff
@@ -54,19 +36,24 @@ def test_start_scan(self):
 
     @responses.activate
     def test_url_with_extra_forward_slash_url(self):
-        config["subsonic"]["url"] = "http://localhost:4040/contextPath"
+        self.config = {'url': 'http://localhost:4040/contextPath'}
```
sampsyo, line 39: I think the old way above is the right way to set config options for a test—that is, overwrite individual values. Doing `self.config = ...` will replace the entire configuration object (which is not a dict ordinarily, but a Confuse view), whereas what you want is to change the _contents_ of the configuration.

beetsplug/subsonicupdate.py
jef: Ahh. Okay, interesting. I was confused because it seemed that other configs were being initialized differently. This makes sense. I also was confused because my tests were not overwriting what was defined as default in this file.

beetsplug/subsonicupdate.py
jef: Ah! Makes sense. Thanks for link.

jef:
Thanks @sampsyo for the help! I reverted back the config changes. I appreciate that explanation.

You know whats fun is that I was using the wrong REST method this whole time... I started calling my own Airsonic (Subsonic) instance and was getting a CSFR error.

Can't believe it. Thought perhaps the username/password was getting malformed or perhaps something I didn't understand with Python (that's probably still the case :rofl:)

sampsyo (approved):
Awesome! Nice work figuring all this out. Looks good to go!

PR #3732, opened 2020-08-23
Update file metadata after generating fingerprints through the `submit` command.

PotcFdk:
This should be pretty self-explanatory.
I noticed that the file metadata was only updated when I used the `fingerprint` command, but not if I used the `submit` command and it generated the fingerprints while doing so. In the latter case, the database was updated, but the files were only updated after I ran `write` afterwards.
I must admit that I'm not fully sure if this is the intended behavior or not, but I thought creating this Pull Request shouldn't hurt, seeing as I fully expected its behavior to stay the same (i.e. to update the file after updating the database), regardless of whether it was triggered directly by `fingerprint` or indirectly via `submit`.

This PR makes the third argument match [this one](https://github.com/beetbox/beets/blob/06e7453a6066813edc68f888aab0e3e5806b6db7/beetsplug/chroma.py#L236).

### DIFF
--- a/beetsplug/chroma.py
+++ b/beetsplug/chroma.py
@@ -279,7 +279,7 @@ def submit_items(log, userkey, items, chunksize=64):
         del data[:]
 
     for item in items:
-        fp = fingerprint_item(log, item)
+        fp = fingerprint_item(log, item, write=ui.should_write())
 
         # Construct a submission dictionary for this item.
         item_data = {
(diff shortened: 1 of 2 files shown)


sampsyo:
Hi! Thanks for doing this! I can't think of a reason to prefer the current behavior, so this seems good. Would you mind adding a quick changelog entry?

PotcFdk:
@sampsyo Amended the commit to include a changelog entry. I hope this looks fine?

sampsyo:
Looks great; thank you!!

PR #3746, opened 2020-09-07
Added flac-specific samplerate-bitdepth reporting for duplicate imports

Sciencentistguy:
## Description
When importing duplicate flac files, it is often less important what the bitrate is and more important the sample rate and bit depth. My change adds this for flac files for convenience

## To Do

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -465,9 +465,13 @@ def summarize_items(items, singleton):
 
     if items:
         average_bitrate = sum([item.bitrate for item in items]) / len(items)
+        if items[0].format == "FLAC":
+            sample_bits = u'{}kHz/{} bit'.format(items[0].samplerate, items[0].bitdepth)
         total_duration = sum([item.length for item in items])
         total_filesize = sum([item.filesize for item in items])
         summary_parts.append(u'{0}kbps'.format(int(average_bitrate / 1000)))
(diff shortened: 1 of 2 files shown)


beets/ui/commands.py
```diff
@@ -465,9 +465,14 @@ def summarize_items(items, singleton):
...
     if items:
         average_bitrate = sum([item.bitrate for item in items]) / len(items)
+        if items[0].format == "FLAC":
+            sample_bits = u'{}kHz/{} bit'.format(
+                round(int(items[0].samplerate) / 1000, 1), items[0].bitdepth)
         total_duration = sum([item.length for item in items])
         total_filesize = sum([item.filesize for item in items])
         summary_parts.append(u'{0}kbps'.format(int(average_bitrate / 1000)))
+        if items[0].format == "FLAC":
+            summary_parts.append(sample_bits)
```
sampsyo, line 475: Any particular reason that this needs to be done in two steps, instead of just in one `if` block here?

sampsyo (commented):
Looks interesting! Could you please add a changelog entry describing the change?

beets/ui/commands.py
Sciencentistguy: You're right, I've merged them.

sampsyo:
Looks great; thanks!

PR #3748, opened 2020-09-11
fetchart: Improve Cover Art Archive source.

djl:
Instead of blindly selecting the first image, we now treat all "front" images as candidates.

This is useful where some digital releases have both an animated cover and a still image and the animated image is the first image returned from the API.

Example release: https://musicbrainz.org/release/dc20e054-0177-4652-b6e7-fe0c193ea8dd

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -318,6 +318,33 @@ class CoverArtArchive(RemoteArtSource):
         """Return the Cover Art Archive and Cover Art Archive release group URLs
         using album MusicBrainz release ID and release group ID.
         """
+
+        def get_image_url(url, size_suffix=None):
+            try:
+                response = self.request(url)
+            except requests.RequestException:
+                self._log.debug(u'{0}: error receiving response'
+                                .format(self.NAME))
+                return
+
+            try:
+                data = response.json()
+            except ValueError:
+                self._log.debug(u'{0}: error loading response: {1}'
+                                .format(self.NAME, response.text))
+                return
+
+            if 'images' in data.keys():
+                for item in data['images']:
+                    if not item['front']:
+                        continue
+
+                    if size_suffix:
+                        return item['thumbnails'][size_suffix]
+
+                    return item['image']
+
+
         release_url = self.URL.format(mbid=album.mb_albumid)
         release_group_url = self.GROUP_URL.format(mbid=album.mb_releasegroupid)
 
@@ -329,36 +356,13 @@ class CoverArtArchive(RemoteArtSource):
         if plugin.maxwidth in self.VALID_THUMBNAIL_SIZES:
             size_suffix = "-" + str(plugin.maxwidth)
 
-        url = None
         if 'release' in self.match_by and album.mb_albumid:
-            url = release_url
-        elif 'releasegroup' in self.match_by and album.mb_releasegroupid:
-            url = release_group_url
+            url = get_image_url(release_url, size_suffix)
+            yield self._candidate(url=url, match=Candidate.MATCH_EXACT)
 
-        try:
-            response = self.request(url)
-        except requests.RequestException:
-            self._log.debug(u'{0}: error receiving response'.format(self.NAME))
-            return
-
-        try:
-            data = response.json()
-        except ValueError:
-            self._log.debug(u'google: error loading response: {}'
-                            .format(response.text))
-            return
-
-        if 'images' in data.keys():
-            for item in data['images']:
-                if not item['front']:
-                    continue
-
-                if size_suffix:
-                    thumbnail_url = item['images'][size_suffix]
-                    yield self._candidate(url=thumbnail_url,
-                                          match=Candidate.MATCH_EXACT)
-                yield self._candidate(url=item['image'],
-                                      match=Candidate.MATCH_EXACT)
+        if 'releasegroup' in self.match_by and album.mb_releasegroupid:
+            url = get_image_url(release_group_url)
+            yield self._candidate(url=url, match=Candidate.MATCH_FALLBACK)
 
 
 class Amazon(RemoteArtSource):
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -319,7 +319,7 @@ class CoverArtArchive(RemoteArtSource):
         using album MusicBrainz release ID and release group ID.
         """
 
-        def get_image_url(url, size_suffix=None):
+        def get_image_urls(url, size_suffix=None):
             try:
                 response = self.request(url)
             except requests.RequestException:
@@ -340,9 +340,9 @@ class CoverArtArchive(RemoteArtSource):
                         continue
 
                     if size_suffix:
-                        return item['thumbnails'][size_suffix]
-
-                    return item['image']
+                        yield item['thumbnails'][size_suffix]
+                    else:
+                        yield item['image']
 
         release_url = self.URL.format(mbid=album.mb_albumid)
         release_group_url = self.GROUP_URL.format(mbid=album.mb_releasegroupid)
@@ -356,12 +356,12 @@ class CoverArtArchive(RemoteArtSource):
             size_suffix = "-" + str(plugin.maxwidth)
 
         if 'release' in self.match_by and album.mb_albumid:
-            url = get_image_url(release_url, size_suffix)
-            yield self._candidate(url=url, match=Candidate.MATCH_EXACT)
+            for url in get_image_urls(release_url, size_suffix):
+                yield self._candidate(url=url, match=Candidate.MATCH_EXACT)
 
         if 'releasegroup' in self.match_by and album.mb_releasegroupid:
-            url = get_image_url(release_group_url)
-            yield self._candidate(url=url, match=Candidate.MATCH_FALLBACK)
+            for url in get_image_urls(release_group_url):
+                yield self._candidate(url=url, match=Candidate.MATCH_FALLBACK)
 
 
 class Amazon(RemoteArtSource):
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -336,7 +336,7 @@ class CoverArtArchive(RemoteArtSource):
 
             if 'images' in data.keys():
                 for item in data['images']:
-                    if not item['front']:
+                    if 'Front' not in item['types']:
                         continue
 
                     if size_suffix:
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -334,8 +334,8 @@ class CoverArtArchive(RemoteArtSource):
                                 .format(self.NAME, response.text))
                 return
 
-            if 'images' in data.keys():
-                for item in data['images']:
+            for item in data.get('images', []):
+                try:
                     if 'Front' not in item['types']:
                         continue
 
@@ -343,6 +343,8 @@ class CoverArtArchive(RemoteArtSource):
                         yield item['thumbnails'][size_suffix]
                     else:
                         yield item['image']
+                except KeyError:
+                    pass
 
         release_url = self.URL.format(mbid=album.mb_albumid)
         release_group_url = self.GROUP_URL.format(mbid=album.mb_releasegroupid)
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -344,7 +344,6 @@ class CoverArtArchive(RemoteArtSource):
 
                     return item['image']
 
-
         release_url = self.URL.format(mbid=album.mb_albumid)
         release_group_url = self.GROUP_URL.format(mbid=album.mb_releasegroupid)
 
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -308,11 +308,11 @@ class CoverArtArchive(RemoteArtSource):
     VALID_THUMBNAIL_SIZES = [250, 500, 1200]
 
     if util.SNI_SUPPORTED:
-        URL = 'https://coverartarchive.org/release/{mbid}/front'
-        GROUP_URL = 'https://coverartarchive.org/release-group/{mbid}/front'
+        URL = 'https://coverartarchive.org/release/{mbid}'
+        GROUP_URL = 'https://coverartarchive.org/release-group/{mbid}'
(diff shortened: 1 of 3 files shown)


beetsplug/fetchart.py
```diff
@@ -329,20 +329,36 @@ def get(self, album, plugin, paths):
...
             size_suffix = "-" + str(plugin.maxwidth)
 
+        url = None
         if 'release' in self.match_by and album.mb_albumid:
-            if size_suffix:
-                release_thumbnail_url = release_url + size_suffix
-                yield self._candidate(url=release_thumbnail_url,
+            url = release_url
+        elif 'releasegroup' in self.match_by and album.mb_releasegroupid:
+            url = release_group_url
```
wisp3rwind, line 336: Previously, the art source would yield candidates from both the release and the release group, with match types `MATCH_EXACT` and `MATCH_FALLBACK`, respectively. Now, if for example, the album has both `release` and `releasegroup` set, and there's no art for the release, but for the group there is, the art source will not return it. Maybe a loop like `for match_by in ("release", "releasegroup")` could restore the behaviour?

wisp3rwind (changes requested):
Looks like an important improvement, thanks! I didn't look at the changes and test failures in detail, but I think there's one important (unintended) change of the plugin's behaviour that needs to be fixed, see the inline comment.

beetsplug/fetchart.py
```diff
@@ -329,20 +329,36 @@ def get(self, album, plugin, paths):
         if plugin.maxwidth in self.VALID_THUMBNAIL_SIZES:
             size_suffix = "-" + str(plugin.maxwidth)
 
+        url = None
```
sampsyo, line 332: This assignment is not necessary.

beetsplug/fetchart.py
```diff
@@ -329,20 +329,36 @@ def get(self, album, plugin, paths):
...
+        try:
+            response = self.request(url)
+        except requests.RequestException:
+            self._log.debug(u'{0}: error receiving response'.format(self.NAME))
+            return
+
+        try:
+            data = response.json()
+        except ValueError:
+            self._log.debug(u'google: error loading response: {}'
```
sampsyo, line 347: This log message doesn't seem to describe what's happening here (and references the wrong backend).

sampsyo (commented):
Thanks for getting this started! Here are some coding suggestions.

It's too bad that this will increase the number of API requests by 1 in the case where the first image suffices, but I suppose there's no way around that.

djl:
It seems I hit the "pull request" button too quickly :man_shrugging: 

I'll fix up the stuff mentioned in the comments ASAP.

beetsplug/fetchart.py
djl: Leaving this out causes a "reference before assignment" error in the tests.

I've refactored this locally so it's not required anymore.

beetsplug/fetchart.py
djl: That, as you can probably tell, was copied from the JSON parsing part from the `GoogleImages` class. If this isn't sufficient, I'm open to suggestions on how to improve it.

beetsplug/fetchart.py
djl: Good catch. Fixed in the latest commit.

djl:
Tests have been updated and should now be passing. They're passing when I run them locally but the CI is dragging its feet.

beetsplug/fetchart.py
```diff
@@ -308,16 +308,42 @@ class CoverArtArchive(RemoteArtSource):
...
+
+            if 'images' in data.keys():
+                for item in data['images']:
+                    if not item['front']:
+                        continue
+
+                    if size_suffix:
+                        return item['thumbnails'][size_suffix]
+
+                    return item['image']
```
wisp3rwind, line 345: > Instead of blindly selecting the first image, we now treat all "front" images as candidates.

Since you return instead of `yield`ing here, I think the original intent of the PR is not achieved, is it? This would of course require a `yield from` (or rather a loop) below.

beetsplug/fetchart.py
```diff
@@ -330,19 +356,12 @@ def get(self, album, plugin, paths):
...
         if 'release' in self.match_by and album.mb_albumid:
-            if size_suffix:
-                release_thumbnail_url = release_url + size_suffix
-                yield self._candidate(url=release_thumbnail_url,
-                                      match=Candidate.MATCH_EXACT)
-            yield self._candidate(url=release_url,
-                                  match=Candidate.MATCH_EXACT)
+            url = get_image_url(release_url, size_suffix)
+            yield self._candidate(url=url, match=Candidate.MATCH_EXACT)
+
```
wisp3rwind, line 361: According to the comment above, this should probably be
```Python
for url in get_image_urls(release_url, size_suffix):
    yield self._candidate(url=url, match=Candidate.MATCH_EXACT)
```

test/test_art.py
```diff
@@ -210,19 +208,6 @@ def test_main_interface_falls_back_to_aao(self):
...
-        album = _common.Bag(mb_albumid=self.MBID, asin=self.ASIN)
-        candidate = self.plugin.art_for_album(album, None)
-        self.assertIsNotNone(candidate)
-        self.assertEqual(len(responses.calls), 1)
-        if util.SNI_SUPPORTED:
-            url = "https://" + self.CAA_URL
-        else:
-            url = "http://" + self.CAA_URL
-        self.assertEqual(responses.calls[0].request.url, url)
-
```
wisp3rwind, line 225: I think this test should stay: The intent is slightly different from the one added in `CoverArtArchiveTest` since here, precedence of the art sources is tested.

wisp3rwind (changes requested):
The code is looking good now! I left some comments about the actual plugin logic. However, I'm not sure I really follow the goal of the PR: From your initial post, I thought the idea was to skip animated images if they are present. However, the JSON in the tests suggests that the `"front"` field doesn't distinguish such images. Thus, in its current form, the PR doesn't change anything, does it? From the [API docs](https://musicbrainz.org/doc/Cover_Art_Archive/API), I can see no easy way tell animated and static pictures apart, except for checking the file extension (if that's reliably correct on CAA). Also, to really yield all front covers (not just the "main front" cover), you should probably check the "types" field in the JSON response.

Did I understand your intent correctly? Could you clarify in some more detail how the plugin should handle multiple images, i.e. which should be preferred?

djl:
> However, I'm not sure I really follow the goal of the PR: From your initial post, I thought the idea was to skip animated images if they are present. However, the JSON in the tests suggests that the "front" field doesn't distinguish such images. Thus, in its current form, the PR doesn't change anything, does it?

My brain lapse with the yield/return definitely added to the confusion.

The initial problem isn't specific to gifs. The plugin currently only looks at the first front image returned from the CAA API. If that image is a gif or anything not in [`FetchArt.CONTENT_TYPES`](https://github.com/beetbox/beets/blob/master/beetsplug/fetchart.py#L40) then no art will be fetched. This change just makes all the front images [candidates](https://github.com/beetbox/beets/blob/master/beetsplug/fetchart.py#L47).

This is the same behavior as the Amazon and Google Images sources.

beetsplug/fetchart.py
djl: You are 100% correct. No idea how that slipped through. Fixed.

test/test_art.py
djl: It probably should stay but I couldn't find a simple way to mock the tests that didn't involve just copy/pasting the two large JSON responses that need to be mocked.

I'm open to ideas.

wisp3rwind:
> The initial problem isn't specific to gifs. The plugin currently only looks at the first front image returned from the CAA API. If that image is a gif or anything not in [`FetchArt.CONTENT_TYPES`](https://github.com/beetbox/beets/blob/master/beetsplug/fetchart.py#L40) then no art will be fetched. This change just makes all the front images [candidates](https://github.com/beetbox/beets/blob/master/beetsplug/fetchart.py#L47).

Right, I hadn't thought about the `CONTENT_TYPES` restriction, that alleviates my statement
> From the [API docs](https://musicbrainz.org/doc/Cover_Art_Archive/API), I can see no easy way tell animated and static pictures apart, except for checking the file extension (if that's reliably correct on CAA).

However, note that in the test JSON, the jpg has `"front": false`, thus even with the `return` -> `yield` change, it will not be provided as a candidate.

> [...], the JSON in the tests suggests that the `"front"` field doesn't distinguish such images. Thus, in its current form, the PR doesn't change anything, does it?  Also, to really yield all front covers (not just the "main front" cover), you should probably check the "types" field in the JSON response.

Both images have `"types": ["Front"]`, I guess this should be checked when looping through the JSON response? (if `"Front" in item["types"]` in `get_image_urls`) The current test doesn't notice this issue since the content type validation only happens in `RemoteArtSource.fetch_image`, not in `CoverArtArchive.get`.

test/test_art.py
```diff
@@ -416,6 +401,112 @@ def test_google_art_returns_no_result_with_malformed_response(self):
...
+    @responses.activate
+    def run(self, *args, **kwargs):
+        super(CoverArtArchiveTest, self).run(*args, **kwargs)
+
+    def mock_response(self, url, json):
+        responses.add(responses.GET, url, body=json,
+                      content_type='application/json')
+
+    def test_caa_finds_image(self):
+        album = _common.Bag(mb_albumid=self.MBID, mb_releasegroupid=self.MBID)
```
wisp3rwind, line 499: Maybe use two different IDs here? E.g. `MBID_RELEASE` and `MBID_GROUP` in analogy to `RESPONSE_RELEASE` and `RESPONSE_GROUP`.

test/test_art.py
wisp3rwind: I can't think of anything pretty offhand, either. But maybe just referencing the class members of `CoverArtArchiveTest` would be fine? Or a mixin class (`CoverArtArchiveTestData` for lack of a better name) could be added that contains just the URLs,  MBIDs and RESPONSEs.

djl:
> Both images have `"types": ["Front"]`, I guess this should be checked when looping through the JSON response? (if `"Front" in item["types"]` in `get_image_urls`) The current test doesn't notice this issue since the content type validation only happens in `RemoteArtSource.fetch_image`, not in `CoverArtArchive.get`.

Yep. Absolutely correct again. It seems the `front` field in the response is just there to signify that it the first image? Surely the order of the images in the response would suffice :confused. In any case, fixed the issue.

test/test_art.py
djl: Done.

test/test_art.py
djl: The latest commit adds a mixin class and brings back this test.

beetsplug/fetchart.py
```diff
@@ -308,16 +308,42 @@ class CoverArtArchive(RemoteArtSource):
...
+            if 'images' in data.keys():
+                for item in data['images']:
+                    if 'Front' not in item['types']:
+                        continue
+
+                    if size_suffix:
+                        yield item['thumbnails'][size_suffix]
+                    else:
+                        yield item['image']
+
```
wisp3rwind, line 346: One thing I just noticed: I might be worthwhile to not make too many assumptions about the JSON response, such that beets won't outright crash in case that ever changes, or does not necessarily contain all the fields. A minimal version could be:
```Python
for item in data.get('images', []):
    try:
        if 'Front' not in item['types']:
            continue

        if size_suffix:
            yield item['thumbnails'][size_suffix]
        else:
            yield item['image']
    except KeyError:
        pass
```

test/test_art.py
```diff
@@ -416,6 +506,30 @@ def test_google_art_returns_no_result_with_malformed_response(self):
...
+        album = _common.Bag(mb_albumid=self.MBID_RELASE,
+                            mb_releasegroupid=self.MBID_GROUP)
+        self.mock_caa_response(self.RELEASE_URL, self.RESPONSE_RELEASE)
+        self.mock_caa_response(self.GROUP_URL, self.RESPONSE_GROUP)
+        candidate = next(self.source.get(album, self.settings, []))
+        self.assertIsNotNone(candidate)
+        self.assertEqual(candidate.url,
+                         'http://coverartarchive.org/release/rid/12345.gif')
+        self.assertEqual(len(responses.calls), 1)
+        self.assertEqual(responses.calls[0].request.url, self.RELEASE_URL)
```
wisp3rwind, line 530: I guess this test would be more meaningful if it checked that `list(self.source.get(album, self.settings, []))` contains both images, which is what this PR is trying to achieve in the first place. the GIF is what the old implementation would have returned, too.

wisp3rwind (approved):
> Yep. Absolutely correct again. It seems the `front` field in the response is just there to signify that it the first image? Surely the order of the images in the response would suffice confused. In any case, fixed the issue.

Yeah, it seems like that is a pretty arbitrary label. I also didn't see any precise documentation on it (aside from marking the "main front" cover).

I left two more comments about making the code a little more robust. Otherwise, I like how this turned out, the tests also look pretty clean now :tada: I guess this could also be merged right now if you disagree with the requests or they are too much work.  It's not exactly uncommon for beets to exit with a stracktrace on all kinds of errors...

beetsplug/fetchart.py
djl: Fair. Fixed.

At some point this should probably be updated to log the failure rather than just `pass` but this is good enough for now.

test/test_art.py
djl: This should actually have been checking for three images. Two for the release and one for the release group. Fixed.

djl:
Assuming this good to go, what's the preference for merging? Regular merge or squashing to keep the history clean?

wisp3rwind:
> Assuming this good to go, what's the preference for merging? Regular merge or squashing to keep the history clean?

Looks good to me (assuming AppVeyor succeeds). I guess that's up to you, there's no policy on it, and I guess everyone has their own way of managing PRs. So merging as is would be fine. For my own PRs, I try to split them into independent patches and rebase+force push a lot to keep the history clean.

djl:
Merged+squashed via GitHub.

Thanks for reviews @wisp3rwind!

wisp3rwind:
Thank @djl for seeing this through :+1:

PR #3749, opened 2020-09-13
Explain disabling plugins

### LINKED ISSUE #3747: Two FAQ proposals - disable plugins and no file or directory
[Two FAQ proposals](https://beets.readthedocs.io/en/latest/faq.html):

# How do I...

Temporarily disable plugins? (A: Not sure about this one. Setting plugins to `''` (2 single quotes) seems to work.)

# Why does beets…

Complain about "no such file or directory" (A: Run beet update with a query if you have removed files without using beets)

If these questions make sense for the FAQ I will write them up.

RollingStar:
## Description

Partial close #3747

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -233,7 +233,7 @@ If you've already moved your music *outside* of beets, you have a few options:
 …disable plugins?
 -----------------
 
-Add a `#` to your :doc:`config.yaml </reference/config>` on the line where your :doc:`/plugins` are mentioned. For example::
+Add a `#` to your :doc:`config.yaml </reference/config>` on the line where your :doc:`plugins </plugins/index>` are mentioned. For example::
 
     # no plugins will be read if a # is at the start of the next line
     # plugins: the, convert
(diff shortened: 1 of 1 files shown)


sampsyo:
Thanks! Looks like there is a ReST error in one of the links. I also think it would be nice to clarify that this is not a special beets things—it's the YAML syntax for comments.

RollingStar:
Fixed now?

docs/faq.rst
```diff
@@ -227,6 +227,20 @@ If you've already moved your music *outside* of beets, you have a few options:
 - Resort to manually modifying the SQLite database (not recommended).
 
 
+
+.. _disable-plugins:
+
+…disable plugins?
+-----------------
+
+Add a `#` to your :doc:`config.yaml </reference/config>` on the line where your :doc:`plugins </plugins/index>` are mentioned. For example::
```
sampsyo, line 236: How about something like this to make it painfully obvious what's going on?

```rest
You can comment out the `plugins:` line in your `config.yaml` using a ``#`` character::
```

sampsyo (commented):
Cool! Sorry to drag this out, but here are a couple of ideas for clarity/concision.

PR #3755, opened 2020-09-20
Web plugin: support limited editing of items and albums

kdelwat:
## Description

This MR adds limited edit functionality to the REST API exposed by the `web` plugin.

It allows `DELETE`ing items and albums, by ID or query, and `PATCH`ing items. 

## To Do

- [x] Documentation
- [x] Changelog

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -143,14 +143,29 @@ def resource_query(name):
     """
     def make_responder(query_func):
         def responder(queries):
-            return app.response_class(
-                json_generator(
-                    query_func(queries),
-                    root='results', expand=is_expand()
-                ),
-                mimetype='application/json'
-            )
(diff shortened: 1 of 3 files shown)


sampsyo:
Wow; this is really neat! Out of curiosity, do you have a use case in mind for the extended API?

I also wanted to point you toward [AURA](https://github.com/beetbox/aura), which we hope will eventually replace the current web API, in case you're interested.

sampsyo (approved):

kdelwat:
Thanks! My use case is managing a beets library on a VPS which I stream music from - I especially want the ability to delete music from my library without having to log in to the server and use the CLI tool, as I add a lot of albums to "try out" but don't want to keep them forever.

I'll take a look at AURA, looks cool!

sampsyo:
Neat! I'd be interested to see if you end up writing any client-side code that uses these new capabilities.

PR #3761, opened 2020-10-01
export: Add --format=jsonlines option

ssssam:
## Description

This adds support for the JSON Lines format as documented at
https://jsonlines.org/.

In this mode the data is output incrementally, whereas the other
modes load every item into memory and don't produce output until
the end.

## To Do

### DIFF
--- a/beetsplug/export.py
+++ b/beetsplug/export.py
@@ -54,6 +54,14 @@ class ExportPlugin(BeetsPlugin):
                     'sort_keys': True
                 }
             },
+            'jsonlines': {
+                # JSON Lines formatting options.
+                'formatting': {
+                    'ensure_ascii': False,
+                    'separators': (',', ': '),
+                    'sort_keys': True
+                }
+            },
             'csv': {
                 # CSV module formatting options.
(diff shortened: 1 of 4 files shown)


ssssam:
@austinmm -- thanks for your great work on adding the `--format` option. I've extended that to support jsonlines because it's the most convenient format for piping into CLI tools. I'd appreciate your comments on the patch :)

sampsyo (approved):
Wahoo; this looks great to me!

ssssam:
Thanks for the review!
Thinking about it, we could simply always output a newline in the JSON emitter. It's a bit strange that it currently doesn't end the JSON output with `\n`

ssssam:
Spotted another issue while using this - pretty printing jsonlines is actually impossible since "Each Line is a Valid JSON Value". I'll remove some of the config options for the format.

sampsyo:
Seems great. To confirm, please feel free to merge whenever you're happy with it. :smiley:

ssssam:
Thanks! It's working fine for me now, so, let's go :)

Issue #3108, opened 2019-01-04
export: Don't try to compute unused fields, and probably don't include `filesize` by default in library exports


kuzi-moto:
### Problem

The main issue is that using the export plugin on large libraries (10k+ files) with a simple query is very slow. It took roughly an hour to export just the path and added fields from the library.

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv export -l -i "path" ultradyne theory
user configuration: /home/kuzi/.config/beets/config.yaml
data directory: /home/kuzi/.config/beets
plugin paths:
Sending event: pluginload
inline: adding item field multidisc
inline: adding item field disc_and_track
lastgenre: Loading canonicalization tree /usr/lib/python3.7/site-packages/beetsplug/lastgenre/genres-tree.yaml
library database: /home/kuzi/.config/beets/library.db
library directory: /home/kuzi/media/music
Sending event: library_opened
could not get filesize: [Errno 2] No such file or directory: b'/home/kuzi/media/music/Ultradyne/(2002) Unknown Works/01 - Theory Proliferation (Ultradyne Remix).mp3'
[
    {
        "path": "/home/kuzi/media/music/Ultradyne/(2002) Unknown Works/01 - Theory Proliferation (Ultradyne Remix).mp3"
    }
Sending event: cli_exit
]
```
Note: I purposely renamed that songs album to prove a point (otherwise nothing is mentioned). Why scan for the filesize if it's not needed? I'm not even querying the file metadata, just the library. Surely this will incur huge performance penalties in larger libraries.

Additionally there could be more verbose output for the plugin.

Here's a link to the music files that trigger the bug (if relevant): 
N/a

### Setup

* OS: arch linux
* Python version: 3.7.1
* beets version: 1.4.8
* Turning off plugins made problem go away (yes/no): no

My configuration (output of `beet config`) is:
Will post if needed

sampsyo:
Got it; thanks for investigating! You may notice that items have a $filesize field that gets the actual size of the underlying file:

```
$ beet ls -f '$filesize'
```

I'm not sure this is the wisest idea in the world—we should investigate why this field exists in the first place. But in ordinary circumstances, it doesn't matter—code that doesn't use this field doesn't need to compute it. The export plugin is an exception, of course. There are two things we should probably do:

- Don't access all the fields when we only want one. That is, don't get the value of $filesize when you only want $path.
- Probably don't include $filesize at all in the default settings when exporting from the library.

PR #1295, opened 2015-02-01
Info plugin can filter properties in output

geigerzaehler:
Resolves #1287

PR #3762, opened 2020-10-01
Optimise FormattedMapping when querying a specific set of fields

ssssam:
## Description

This changes greatly improves the speed of `beet export` and `beet info`
when the `--include-keys` option is used.

Listing all fields for an item requires querying the database to find
any flex attributes. This is slow when done for every item being
exported. We already have a way for the user to specify a fixed set
of keys, but we previously queried everything and filtered it afterwards.
The new approach is more efficient.

Code that iterates through all fields now have to handle invalid field
names. The export and info plugins output invalid fields as None.

Timings before:

    > /usr/bin/time beet export -i title,path,artist -l  Bob Dylan
    13.26user 20.22system 0:34.01elapsed 98%CPU (0avgtext+0avgdata 52544maxresident)k

    > /usr/bin/time beet export -l  Bob Dylan
    12.93user 20.15system 0:33.58elapsed 98%CPU (0avgtext+0avgdata 53632maxresident)k

Timings after:

    > /usr/bin/time beet export -l  Bob Dylan
    13.33user 20.17system 0:34.02elapsed 98%CPU (0avgtext+0avgdata 53500maxresident)k

    > /usr/bin/time beet export -i title,path,artist -l  Bob Dylan
    0.49user 0.07system 0:00.56elapsed 98%CPU (0avgtext+0avgdata 50496maxresident)k

Notice the dramatic speedup in the last example!

Fixes https://github.com/beetbox/beets/issues/3108

## To Do

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -52,15 +52,24 @@ class FormattedMapping(Mapping):
     The accessor `mapping[key]` returns the formatted version of
     `model[key]` as a unicode string.
 
+    The `included_keys` parameter allows filtering the fields that are
+    returned. By default all fields are returned. Limiting to specific keys can
+    avoid expensive per-item database queries.
+
     If `for_path` is true, all path separators in the formatted values
     are replaced.
     """
 
-    def __init__(self, model, for_path=False, compute_keys=True):
+    ALL_KEYS = '*'
+
+    def __init__(self, model, included_keys=ALL_KEYS, for_path=False):
         self.for_path = for_path
         self.model = model
-        if compute_keys:
-            self.model_keys = model.keys(True)
+        if included_keys == self.ALL_KEYS:
+            # Performance note: this triggers a database query.
+            self.model_keys = self.model.keys(True)
+        else:
+            self.model_keys = included_keys
 
     def __getitem__(self, key):
         if key in self.model_keys:
@@ -605,11 +614,11 @@ class Model(object):
 
     _formatter = FormattedMapping
 
-    def formatted(self, for_path=False):
+    def formatted(self, included_keys=_formatter.ALL_KEYS, for_path=False):
         """Get a mapping containing all values on this object formatted
         as human-readable unicode strings.
         """
-        return self._formatter(self, for_path)
+        return self._formatter(self, included_keys, for_path)
 
     def evaluate_template(self, template, for_path=False):
         """Evaluate a template (a string or a `Template` object) using
@@ -619,7 +628,7 @@ class Model(object):
         # Perform substitution.
         if isinstance(template, six.string_types):
             template = functemplate.template(template)
-        return template.substitute(self.formatted(for_path),
+        return template.substitute(self.formatted(for_path=for_path),
                                    self._template_funcs())
 
     # Parsing.
--- a/beets/library.py
+++ b/beets/library.py
@@ -374,12 +374,19 @@ class FormattedItemMapping(dbcore.db.FormattedMapping):
     Album-level fields take precedence if `for_path` is true.
     """
 
-    def __init__(self, item, for_path=False):
+    ALL_KEYS = '*'
+
+    def __init__(self, item, included_keys=ALL_KEYS, for_path=False):
         # We treat album and item keys specially here,
         # so exclude transitive album keys from the model's keys.
-        super(FormattedItemMapping, self).__init__(item, for_path,
-                                                   compute_keys=False)
-        self.model_keys = item.keys(computed=True, with_album=False)
+        super(FormattedItemMapping, self).__init__(item, included_keys=[],
+                                                   for_path=for_path)
+        self.included_keys = included_keys
+        if included_keys == self.ALL_KEYS:
+            # Performance note: this triggers a database query.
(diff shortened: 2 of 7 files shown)


beets/util/__init__.py
```diff
@@ -1108,3 +1108,31 @@ def wrapper(self):
...
+    glob_keys = set(k for k in included_keys if '*' in k)
+    keys = set(k for k in included_keys if k not in glob_keys)
+
+    if glob_keys:
+        all_keys = all_keys_cb()
+
+        matchers = []
+        for key in glob_keys:
+            pattern = re.escape(key).replace(r'\*', '.*') + '$'
+            matchers.append(re.compile(pattern))
```
sampsyo, line 1129: You may have considered this already, but could the [fnmatch.fnmatch](https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatch) function be useful for this to avoid constructing regexes explicitly? Or else [fnmatch.translate](https://docs.python.org/3/library/fnmatch.html#fnmatch.translate) if we still want to use proper regexes?

beets/util/__init__.py
```diff
@@ -1108,3 +1108,31 @@ def wrapper(self):
...
+
+        def filter_(keys):
+            for key in keys:
+                if any([m.match(key) for m in matchers]):
+                    yield key
+
+        keys.update(filter_(all_keys))
```
sampsyo, line 1136: Could perhaps use a comprehension?

`keys.update(key for key in keys if any(...))`

sampsyo (commented):
Hello! I am deeply embarrassed that it took me so long to actually look at this PR. That's not good maintainer behavior! With apologies, I did finally have a moment to get an idea of how this works.

Overall, this optimization definitely seems important and we should do it! However, an aspect of the design gives me a little bit of pause: namely, putting the `*`-expanding glob stuff _into dbcore_, which is our lowest-level database abstraction layer. Having an explicit list of fields to filter there seems like a perfectly good fit, but tying that layer to a specific style of glob expansion UI seems a little odd.

Is there any way we could get a similar effect to this optimization while keeping the glob-related work "owned" by the plugins that use them in their command-line interface (like `export`)? Maybe callbacks are the key to making that work, which would have its own weirdness, but it would make for a bit nicer separation of concern.

beets/util/__init__.py
ssssam: fnmatch implements a slightly more complex syntax (`?` wildcards, `[abc]` character sequences). Is that what we want?

beets/util/__init__.py
ssssam: No problem. Although it's longer ;)

ssssam:
Thanks for the review!

I think it would be possible to move the globbing syntax to a different level. The tricky thing is this is very performance sensitive when called from `beet export`. At present it works like this:

  * Call model.all_keys()
  * Query database to return all keys

I propose changing it to this:

  * Call model.all_keys()
  * If included_keys is set and doesn't contain '*', return included_keys
  * Otherwise, query database to find all keys, and filter against any globs in included_keys

If we move the functionality to a higher level. We'd need to have two hooks. One to determine if there are any expansions to be done, so we can default to fast codepath. And one to call to do the expansion.

  * Call model.all_keys
  * If included_keys is set and there's no expansion callbacks, or the is_expansion_needed() callback return False -> return included_keys
  * Otherwise, query database to find all keys, and filter result against do_expansion() callback

I understand the desire to keep a clean layering of concerns, but there's a tradeoff with complexity. Could we remove the globbing feature altogether? I'm not sure if it's used outside `beet export`?

beets/util/__init__.py
sampsyo: Yeah, good point. While they seem cool, I can't imagine anyone actually wanting to use them. So maybe it's not worth it.

sampsyo:
Thanks for the detailed background; that was a helpful way to put it.

> Could we remove the globbing feature altogether? I'm not sure if it's used outside `beet export`?

Brilliant. I think this would be a great idea. I think we probably added that feature too hastily in the first place and it wasn't very well considered. Let's give this a try with just removing it altogether—if anyone complains, we can consider a more careful implementation instead.

ssssam:
Hey! I found time to revisit this patch. Changes in the new version:

  * Remove globbing feature added along with `--include-keys` option in https://github.com/beetbox/beets/pull/1295
  * Resolve conflicts with https://github.com/beetbox/beets/pull/2988
  * Squash all changes in one commit.

sampsyo (approved):
Excellent work. I did a full review and don't have any comments. Thanks for taking care of this, @ssssam!

ssssam:
Great! Anything missing to merge it, or shall I push the big green button? :slightly_smiling_face:

sampsyo:
Ah yes, I should have said—go for it!! 🟢👈

ssssam:
Great! That's was the last thing on my to-do list, and now I can retire   :smile:

sampsyo:
Enjoy some well-earned rest in your golden years. 😆

kuzi-moto:
Just want to say thanks for the work on this!

PR #3765, opened 2020-10-05
Add --plugins flag

jackwilsdon:
## Description

This PR adds a `--plugins` flag to beets, allowing enabling/disabling of plugins without adding them to the user's configuration.

## To Do

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -35,6 +35,12 @@ Here's a link to the music files that trigger the bug (if relevant):
 * beets version: 
 * Turning off plugins made problem go away (yes/no): 
 
+<!--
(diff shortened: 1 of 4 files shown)


jackwilsdon:
I've added a changelog and some documentation for this now :+1:

sampsyo:
Nice! This looks great to me. The only thought I have here is whether a UI with multiple arguments, like `--plugin a --plugin b` might be more predictable/friendly than one with commas, like `--plugins a,b`. But I can definitely see it going either way. Other than that, I think this is ready to go.

jackwilsdon:
@sampsyo how would that work to disable all plugins? Would we need a new `--no-plugins` flag?

sampsyo:
Great point; that totally wouldn't work (and was the original suggestion that started this, after all). Let's keep it the way you have it!

jackwilsdon:
Great, cheers :tada: I'll go ahead and merge this now

PR #3766, opened 2020-10-05
Remove LyricWiki source

jackwilsdon:
## Description

[LyricWiki was shut down on 2020/09/21 and no longer serves lyrics.](https://en.wikipedia.org/wiki/LyricWiki#Edit_lock_and_closure)

The integration tests on our CI have been failing because of this.

## To Do


sampsyo:
Sad to hear about the demise! Thank you for tracking down the problem, though, and removing the source. Looks good!

PR #3769, opened 2020-10-08
Swap links

nathan78906:
## Description

Links were swapped and pointing to wrong locations. This re-swaps them so they point to the correct locations.

## To Do

- [N/A] Documentation. (If you've add a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)
- [N/A] Changelog. (Add an entry to `docs/changelog.rst` near the top of the document.)
- [N/A] Tests. (Encouraged but not strictly required.)


wisp3rwind:
Looks good, thanks @nathan78906!

PR #3777, opened 2020-10-21
Fix linkcheck on GitHub actions

jackwilsdon:
It looks like the existing GitHub user page regex wasn't quite right - I've updated this to match anything at the top-level of GitHub (it was missing usernames containing hyphens before).

I've added genius to the link ignore list too, as they're still blocking GitHub actions (see #3713).

Here's hoping this is the final nail in the integration test failure coffin 💪


sampsyo:
Awesome; thanks!! These changes look perfect.

PR #3778, opened 2020-10-23
Update faq.rst

KyleRConway:
A couple of suggested modifications primarily to clarify that this unordered list is of **options** to install the latest version―not a series of commands to complete a single install. Also suggested a layout change to explain the series of commands in option b.

Thanks for a great program!


sampsyo:
Cool, thanks! Here's a minor suggestion: instead of numbering the option bullets, what if we just label the list with something like, "choose one of these methods:"?

KyleRConway:
@sampsyo -- Thanks for the feedback. Made some modifications for review. :)

sampsyo:
Thanks! Merged with a couple of tweaks.

PR #3781, opened 2020-10-23
Add genre support using musicbrainz tags.

### LINKED ISSUE #3053: Get genres from MusicBrainz
### Problem

Next release of MusicBrainz will have genre support. An example can be seen at https://beta.musicbrainz.org/release-group/3bd76d40-7f0e-36b7-9348-91a33afee20e (JSON web service output: https://beta.musicbrainz.org/ws/2/release-group/3bd76d40-7f0e-36b7-9348-91a33afee20e?fmt=json&inc=genres )

beets should fetch these and allow MusicBrainz to be used as "genre authority".

aereaux:
## Description

Fixes #3053.

This used musicbrainz tags to populate the genre album tag.  This is still a WIP, and it requires this PR (https://github.com/alastair/python-musicbrainzngs/pull/266).  Let me know if anyone has any ideas on how this could work better.  Currently I just take the most-voted-for genre tag, breaking ties arbitrarily.

## To Do

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -74,6 +74,8 @@ RELEASE_INCLUDES = ['artists', 'media', 'recordings', 'release-groups',
 TRACK_INCLUDES = ['artists', 'aliases']
 if 'work-level-rels' in musicbrainzngs.VALID_INCLUDES['recording']:
     TRACK_INCLUDES += ['work-level-rels', 'artist-rels']
+if 'genres' in musicbrainzngs.VALID_INCLUDES['recording']:
+    RELEASE_INCLUDES += ['genres']
 
 
 def track_url(trackid):
@@ -414,6 +416,9 @@ def album_info(release):
     if release['medium-list']:
         first_medium = release['medium-list'][0]
         info.media = first_medium.get('format')
+
+    if release.get('genre-list'):
+        info.genre = ';'.join(g['name'] for g in release['genre-list'])
 
     info.decode()
     return info
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -417,8 +417,9 @@ def album_info(release):
         first_medium = release['medium-list'][0]
         info.media = first_medium.get('format')
 
-    if release.get('genre-list'):
-        info.genre = ';'.join(g['name'] for g in release['genre-list'])
+    genres = release.get('genre-list')
+    if genres:
+        info.genre = ';'.join(g['name'] for g in genres)
 
     info.decode()
     return info
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -159,6 +159,11 @@ New features:
 * :doc:`/plugins/web`: add DELETE and PATCH methods for modifying items
 * :doc:`/plugins/lyrics`: Removed LyricWiki source (shut down on 21/09/2020).
 * Added a ``--plugins`` (or ``-p``) flag to specify a list of plugins at startup.
+* Use musicbrainz genre tag api to get genre information.  This currently
+  depends on functionality that is currently unreleased in musicbrainzngs.
+  See https://github.com/alastair/python-musicbrainzngs/pull/247 and
+  https://github.com/alastair/python-musicbrainzngs/pull/266 .
+  Thanks to :user:`aereaux`.
 
 Fixes:
 


sampsyo:
Great! This seems like the right direction. Just choosing the most popular genre seems right to me, although an alternative would be to take _all_ the genres and join them (as the `lastgenre` plugin can optionally do, when configured appropriately).

aereaux:
Yeah, part of posting this PR early is to see what would be the best interface for this.  I mostly want it so that I can sort things in my library by genre (and in particular listing classical things by composer and work, and everything else by album).  It seems like the best solution would be a multi-valued tag, but it sounds like that's not supported yet.

One difficulty with my mentioned use-case is releases such as this: https://musicbrainz.org/release/3f4ef072-37fe-4759-8a2d-7e6cb61cfafd .  I think lastgenre uses a tag hierarchy to deal with this sort of thing.

aereaux:
OK, so I've changed this to add all tagged genres as a semicolon separated list.  Might need some testing on more popular collections, not sure how well-tagged some albums are.  I'm solving my use-case by using path queries on the genre tag, which works quite well.  Anyways, seems ready for review to me.

sampsyo:
Super cool!! Just to clarify, do we need to wait for the new version of python-musicbrainz-ngs? I think the answer is "no" (this will just do nothing when genres are missing), but I wanted to double-check.

aereaux:
Unfortunately, with v0.7.1 of mb-ngs I believe it should error because the changes to add genres to the list of supported includes is only in master so far (https://github.com/alastair/python-musicbrainzngs/pull/263).  Not sure what the best solution is.

sampsyo:
Got it. Maybe we should add an explicit check that asks whether `'genres' in musicbrainzngs.TAG_INCLUDES` or similar, and only then request genres?

aereaux:
OK, added.  Thanks for reviewing quickly.

sampsyo:
Cool; thanks!! Is there any chance you could confirm, just to be sure we don't do something dumb, that this version of the code does not break on the currently-released version of the library?

It also might be nice to mention in the changelog that getting this new support depends on a future release of the relevant library.

aereaux:
Yep, just reinstalled the non-git version and verified it.  (Although I would have expected it to get rid of my current genre information, but I guess it keeps it.)

beets/autotag/mb.py
```diff
@@ -415,6 +417,9 @@ def album_info(release):
         first_medium = release['medium-list'][0]
         info.media = first_medium.get('format')
 
+    if release.get('genre-list'):
+        info.genre = ';'.join(g['name'] for g in release['genre-list'])
```
sampsyo, line 421:
```suggestion
    genres = release.get('genre-list')
    if genres:
        info.genre = ';'.join(g['name'] for g in genres)
```

sampsyo (commented):
Looking good; thanks for following up. Looking at the code, I just found one simplification/removal of redundancy to suggest. If that looks good, then I think this is ready to merge.

trolley:
@aereaux Reading the comments on this PR am I correct in understanding that this will always add every genre that anyone has tagged a release with? Picard has really nice genre support in that it will let me optionally use only the genre's I've added to MB releases myself, it'll fall back to album artist's genre if the individual release or release group doesn't have a genre, and it lets you set a maximum number of genres to use. Will you consider adding support for options like that?

Also, looking at the code I don't see that you can disable genre adding in Beets, which I assume means this could overwrite genre tags at are already in the files. I personally wouldn't want that, so it would be nice to have this as an opt-in option until it's configurable.

sampsyo:
Aha, you know what, that is a good point—at the very least, it should be optional. Could you please add a config option, @aereaux? The single-user genre and fallback support sounds nice too but maybe we should make a feature request ticket for that to track it.

aereaux:
@trolley Correct, this will add all genres that have been tagged.  My use-case has been satisfied, so I am less motivated to implement additional functionality, but if you have a specific specification for how you want it to work, I could probably work on implementing it.

Currently tag overwriting shouldn't be an issue unless someone is testing it out with my musicbrainzngs PR, but I can create a PR for a config option once I get a chance.  (I don't think I've interacted with the config system before, so I need to figure that out.)

Issue #3779, opened 2020-10-23
beet config with empty configuration could show a message


Sudo-kun:
I'm new to the project so maybe I just forgot to do something.

 I was just looking around and found that when calling config while nothing was configured showed an empty array.

Maybe there could be a text that says "No existing configuration".

sampsyo:
Sure, sounds good. But I'd phrase it as "empty configuration" instead of "no configuration" because the symptom would be the same if there is just an empty file there.

jackwilsdon:
Fixed by #3782.

PR #3782, opened 2020-10-24
Feature/empty config

Sudo-kun:
"beet config with empty configuration could show a message #3779"

I added an if-clause if the config output was empty and print a different output if it is.

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1691,7 +1691,7 @@ def config_func(lib, opts, args):
     # Dump configuration.
     else:
         config_out = config.dump(full=opts.defaults, redact=opts.redact)
-        if config_out != '{}\n':
+        if config_out.strip() != '{}':
             print_(util.text_string(config_out))
         else:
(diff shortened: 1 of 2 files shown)


beets/ui/commands.py
```diff
@@ -1691,7 +1691,10 @@ def config_func(lib, opts, args):
     # Dump configuration.
     else:
         config_out = config.dump(full=opts.defaults, redact=opts.redact)
-        print_(util.text_string(config_out))
+        if config_out != '{}\n':
```
sampsyo, line 1694: Maybe `config_out.strip() != '{}'` to avoid white-space sensitivity?

docs/changelog.rst
```diff
@@ -6,6 +6,7 @@ Changelog
 
 New features:
 
+* When config is printed with no available configuration a new message is printed.
```
sampsyo, line 9: Linking to the issue number here might be cool.

sampsyo:
Thanks!!

Issue #3569, opened 2020-04-25
Importer: "Keep both" should be "keep all"


RollingStar:
### Problem

When importing duplicate albums, the importer will prompt to "keep both" if "ask" is set in the config.

https://github.com/beetbox/beets/blob/bf9b43a769e4fdc32ff6d2a958af6cfa124ec7de/beets/ui/commands.py#L806

This is true even when you have 2+ albums already in your library and are trying to add a duplicate. For simplicity and clarity, it should be "keep all".

If we make the change, a few docs and comments will have to change too. Nothing major.

https://github.com/beetbox/beets/search?q=%22keep+both%22&unscoped_q=%22keep+both%22

sampsyo:
Good call. Do you think it would be too crazy for it to dynamically switch between "keep both" (for 2) and "keep all" (for >=3)? Probably not necessary, just thought I'd wonder aloud to see if it seemed worthwhile. :smiley:

RollingStar:
Good question. IMO, if more verbose comparison dialog is ever implemented (such as https://github.com/beetbox/beets/issues/2852), "keep all" will be more intuitive even with just 2 albums.

PR #3783, opened 2020-10-24
Changed the phrasing asked in Issue 3569

Sudo-kun:
"Issue #3569 changed the the text from Keep both to keep all"

Changed "Keep both" to "Keep all"


sampsyo:
Looks good! As in #3782, however, it would be awesome to refer to the original issue number in the changeling.

sampsyo:
Awesome; thanks!

PR #3792, opened 2020-11-09
escape ? in fish completion

benneti:
## Description

Change the fish plugin to escape "?" in the completions file, without escping ? is interpreted as a regex and creates lots of errors that interfere with the actual useful completions.
This PR introduces changes to escape "?" in the command names in fish.py to ensure working completions.

## To Do

EDIT: I am not sure whether this should go into the changelog as the fish plugin is not even included in the latest release.

### DIFF
--- a/beetsplug/fish.py
+++ b/beetsplug/fish.py
@@ -132,6 +132,7 @@ class FishPlugin(BeetsPlugin):
         with open(completion_file_path, 'w') as fish_file:
             fish_file.write(totstring)
 
+
 def _escape(name):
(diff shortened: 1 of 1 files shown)


sampsyo:
Seems great! One small suggestion: what about making a small utility function called, like, `_escape` that does this replacement? So both places could just pass the relevant string through that function instead of duplicating the `if`.

benneti:
thanks, done

sampsyo:
Yay; thanks!!

Issue #2499, opened 2017-03-31
Import: Remove old removes an album if importing an album already in a library


RollingStar:
### Problem

Another silly one. Not a high priority by any means. I can try reproducing it with -vv enabled but it should be self explanatory. This is either a beets bug or a documentation clarity bug. Would this have worked if I pointed it at my global beets library folder rather than a subfolder for one album?

https://beets.readthedocs.io/en/v1.4.3/reference/cli.html#reimporting

>Just point the beet import command at a directory of files that are already catalogged in your library. Beets will automatically detect this situation and avoid duplicating any items.

In my case, "removing old" actually deleted the only copy of the file.

Expected behavior:

* if the new and old albums are the exact same files in the exact same spot on disk, just import the new (according to the current user path config)
* don't prompt in this case, or make a different prompt rather than the "old/new" setup, because importing the new is implicitly changing the old

Running this command in (non)verbose mode:

```sh
beet import "\Alan Menken - 1989 - The Little Mermaid (1997 CD) [96kbps MP3]"

\Alan Menken - 1989 - The Little Mermaid (1997 CD) [96kbps MP3] (1 items)
Tagging:
    Alan Menken - The Little Mermaid
URL:
    https://musicbrainz.org/release/eea4bcaa-f35e-3a7e-8d50-d6f7f1b4f17f
(Similarity: 51.3%) (missing tracks) (CD, 1997, US, Walt Disney Records, 1989 film soundtrack)
Missing tracks (19/20 - 95.0%):
 ! Fathoms Below (#1) (1:41)
 ! Main Titles (#2) (1:26)
 ! Fanfare (#3) (0:30)
 ! Daughters of Triton (#4) (0:40)
 ! Part of Your World (#5) (3:15)
 ! Part of Your World (reprise) (#7) (2:18)
 ! Poor Unfortunate Souls (#8) (4:51)
 ! Les Poissons (#9) (1:36)
 ! Kiss the Girl (#10) (2:43)
 ! Fireworks (#11) (0:38)
 ! Jig (#12) (1:34)
 ! The Storm (#13) (3:20)
 ! Destruction of the Grotto (#14) (1:55)
 ! Flotsam and Jetsam (#15) (1:25)
 ! Tour of the Kingdom (#16) (1:27)
 ! Bedtime (#17) (1:23)
 ! Wedding Announcement (#18) (2:19)
 ! Eric to the Rescue (#19) (3:43)
 ! Happy Ending (#20) (3:11)
[...]

PR #3796, opened 2020-11-18
Move files to trash instead of deleting via new config option

### LINKED ISSUE #149: Move music to trash
This issue was automatically migrated from Google Code.
_Original author: adrian.sampson (April 10, 2010 20:49:19)_
_Original issue: https://github.com/google-code-export/beets/issues/35_

justinmayer:
## Description

While it doesn't happen often, some folks have had experiences in which Beets unexpectedly deletes a song. All software has bugs, and data loss sometimes happens. But as someone who once lost an entire music collection due to data loss (not Beets-related), I think we should do everything we can to prevent unexpected file deletion, regardless of whether that happens via bugs or user mistakes. 

This pull request enhances the `util.remove()` function to move the specified file to the trash by default **if** there is a `trash` executable on `$PATH`. Otherwise, the file is deleted. This approach provides the option for a safer user experience without adding yet another Python dependency to the project.

The executable to be used for these safer file removal operations (default: `trash`) is configurable via a new `remove_command` configuration option.

Fixes #149 #2499

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -30,10 +30,18 @@ import traceback
 import subprocess
 import platform
 import shlex
-from beets.util import hidden
 import six
 from unidecode import unidecode
 from enum import Enum
+
+# This whole block can be removed once Python <3.3 support is dropped
+try:
+    from shutil import which
+except ImportError:
+    from distutils.spawn import find_executable as which
+
+from beets import config
+from beets.util import hidden
 
 
 MAX_FILENAME_LENGTH = 200
@@ -442,17 +450,34 @@ def samefile(p1, p2):
     return shutil._samefile(syspath(p1), syspath(p2))
 
 
-def remove(path, soft=True):
-    """Remove the file. If `soft`, then no error will be raised if the
-    file does not exist.
+def remove(path, soft=True, use_trash=True):
+    """Remove the file specified at `path`, using `trash` executable by default
+    (if found) or via the command specified in the `remove_command` setting if
+    found on $PATH. Otherwise, delete the file immediately. If `soft`, then no
+    error will be raised if the file does not exist.
     """
     path = syspath(path)
     if soft and not os.path.exists(path):
         return
-    try:
-        os.remove(path)
-    except (OSError, IOError) as exc:
-        raise FilesystemError(exc, 'delete', (path,), traceback.format_exc())
+    # If `use_trash` is invoked and the `remove_command` configuration setting
+    # (which defaults to `trash`) is found on $PATH, use that command to remove
+    # the specified file. Otherwise, delete it immediately.
+    # Change which() to shutil.which() once Python <3.3 support is dropped
+    remove_command = config["remove_command"].as_str()
+    if use_trash and which(remove_command) is not None:
+        try:
+            command_output([remove_command, path])
+        except (ValueError, OSError, subprocess.CalledProcessError) as exc:
+            raise FilesystemError(
+                exc, 'delete', (path,), traceback.format_exc()
+            )
+    else:
+        try:
+            os.remove(path)
+        except (OSError, IOError) as exc:
+            raise FilesystemError(
+                exc, 'delete', (path,), traceback.format_exc()
+            )
 
 
 def copy(path, dest, replace=False):
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -69,6 +69,7 @@ terminal_encoding:
 original_date: no
 artist_credit: no
 id3v23: no
+remove_command: "trash"
 va_name: "Various Artists"
 
 ui:
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -41,6 +41,11 @@ Major new features:
 
 Other new things:
 
+* File removal operations are performed by the command specified in the new
+  :ref:`remove_command` configuration option (default: ``trash``), allowing
+  users to optionally install and use safer file removal tools.
+  Thanks to :user:`justinmayer`.
+  :bug:`3796`
 * :doc:`/plugins/mpdstats`: Add a new `strip_path` option to help build the
   right local path from MPD information.
 * :doc:`/plugins/convert`: Conversion can now parallelize conversion jobs on
--- a/docs/reference/config.rst
+++ b/docs/reference/config.rst
@@ -347,6 +347,16 @@ By default, beets writes MP3 tags using the ID3v2.4 standard, the latest
 version of ID3. Enable this option to instead use the older ID3v2.3 standard,
 which is preferred by certain older software such as Windows Media Player.
 
+.. _remove_command:
+
+remove_command
+~~~~~~~~~~~~~~
+
+If Beets finds the specified executable on ``$PATH`` (default: ``"trash"``),
+file removal operations will be performed via that command. If this setting
+is set to ``""`` or to a value that does not appear to be an executable file,
+to-be-removed files will be deleted immediately via ``os.remove()``.
+
 .. _va_name:
 
 va_name


RollingStar:
FYI:

> Note: The Linux implementation is not very good and not maintained. Help welcome. If no one steps up to help maintain it, I will eventually remove Linux support.

justinmayer:
I saw that notice before and selected the project ([Trash](https://github.com/sindresorhus/trash)) as an example nonetheless. Reasons include:
* There are many projects that provide a `trash` command, but few of them are cross-platform.
* [Trash](https://github.com/sindresorhus/trash) is merely one such project, an example. Providing an exhaustive list is out-of-scope.
* Providing solid support across Linux distributions is non-trivial, so this note is unsurprising.
* Linux users are statistically the best-equipped to find an alternative that suits them, if necessary.
* Perhaps some Linux folks will step forward and help out with Linux support in [Trash](https://github.com/sindresorhus/trash).

sampsyo:
Yo! This is a good idea. Can I make these suggestions, however?

- Can we make the docs and the code changes separate PRs? They will both require separate consideration, although we probably want both of them.
- For "trashing" things, I think it would be great to make this a config option rather than the default. This would also let people configure the tool they want to use, if it does not happen to be the first `trash` program on `$PATH`.

justinmayer:
Hey Adrian!

> Can we make the docs and the code changes separate PRs? They will both require separate consideration, although we probably want both of them.

Certainly. They were related enough that it seemed sensible to submit them together, but I'm happy to separate them.

> For "trashing" things, I think it would be great to make this a config option rather than the default. This would also let people configure the tool they want to use, if it does not happen to be the first `trash` program on `$PATH`.

Sure, I can see the benefit in that. I'll have another pass at it.

ssssam:
On some (most?) Linux systems, the `gio trash` command provided by [gio](https://developer.gnome.org/gio/stable/gio.html) is a good way to send files to the trash. Perhaps we could also try that if available. 

If not, we could mention in the docs that creating a script like this can enable the feature on Linux:

    cat > ~/.local.bin/trash <<"EOF"
    #!/bin/sh
    gio trash $@
    EOF
    chmod +x ~/.local/bin/trash

sampsyo:
Nice; great point. All the more reason we should probably make the command configurable—so one could just configure it directly to do `gio trash`.

justinmayer:
Hey @sampsyo. Based on your feedback, I addressed the points you raised:

> Can we make the docs and the code changes separate PRs? They will both require separate consideration, although we probably want both of them.

I removed the documentation changes relating to backups, which I will submit in a separate pull request.

> For "trashing" things, I think it would be great to make this a config option rather than the default. This would also let people configure the tool they want to use, if it does not happen to be the first trash program on $PATH.

The executable is now configurable via the `remove_command` setting.

Anything left to do here before merging? ✨

beets/util/__init__.py
```diff
@@ -442,17 +450,29 @@ def samefile(p1, p2):
     return shutil._samefile(syspath(p1), syspath(p2))
 
 
-def remove(path, soft=True):
-    """Remove the file. If `soft`, then no error will be raised if the
-    file does not exist.
+def remove(path, soft=True, remove_command=config["remove_command"].as_str()):
```
wisp3rwind, line 453: Is it guaranteed that the config has actually been read by the time the util module is imported?

beets/util/__init__.py
```diff
@@ -442,17 +450,29 @@ def samefile(p1, p2):
...
+    """Remove file via command specified in the `remove_command` setting if
+    specified executable is found on $PATH. Otherwise, delete the file.
```
wisp3rwind, line 455: Maybe it would be safer to raise an Exception when the executable cannot be found? It appears to me that currently, a typo in the `remove_command` option would be completely silent. The same would happen by accidentally uninstalling the trash program.

wisp3rwind (commented):
I added a few remarks, I think the most important one would be the handling of invalid config values. I haven't been part of the discussion on trashing and the backup recommendation so far, so if this has been discussed before feel free to dismiss my comments.

It might be useful to grep the codebase for uses of `util.remove`; I could imagine it being used on temporary files that should not go to trash.

justinmayer:
Hey @wisp3rwind. Thanks for the comments. I did grep through the code, and I didn't see any areas in which `util.remove()` not deleting immediately would pose an issue.

beets/util/__init__.py
justinmayer: The Beets development docs seem to imply that this behavior can be relied upon.

beets/util/__init__.py
justinmayer: The idea is to provide — without any action on the part of end users — a sane, safe default, which is to use the `trash` executable if present. If not present, the current behavior gracefully falls back to the pre-existing `os.remove()` behavior. If an exception were raised when the executable isn't found, then the first-run experience for anyone without a `trash` executable on `$PATH` is going to see a surprising exception that they may not understand. I don't think that's a great user experience.

It's very easy to tell whether files are being moved to trash or not, so I don't think silently falling back to the pre-existing `os.remove()` behavior is problematic.

beets/util/__init__.py
```diff
@@ -442,17 +450,29 @@ def samefile(p1, p2):
...
     if soft and not os.path.exists(path):
         return
-    try:
-        os.remove(path)
-    except (OSError, IOError) as exc:
-        raise FilesystemError(exc, 'delete', (path,), traceback.format_exc())
+    # Change this to shutil.which() once Python <3.3 support is dropped
+    if remove_command and which(remove_command) is not None:
+        try:
+            remove = command_output([remove_command, path])  # NOQA:F841
```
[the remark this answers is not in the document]
justinmayer: Fine with me. Changed.

beets/util/__init__.py
wisp3rwind: Ok :+1: Do you have a link here? I'd actually be interested in reading up on this, it doesn't look obvious at all: I would have thought that this might depend on import order and thus be rather fragile.

beets/util/__init__.py
wisp3rwind: Right - I hadn't realised that the default would become `remove_command=trash` rather than `remove_command=""`. If using `trash` is going to be the default, you're right of course.

wisp3rwind:
> I did grep through the code, and I didn't see any areas in which `util.remove()` not deleting immediately would pose an issue.

I've just been doing the same, but I do think that there are a few cases in `test/` and `beetsplug/` that should be exempt. I'll open a PR relative to this one with the changes (I probably won't make it today anymore, though).

beets/util/__init__.py
justinmayer: From https://github.com/beetbox/beets/blob/master/docs/dev/plugins.rst#read-configuration-options:

> If you want to access configuration values _outside_ of your plugin's section, import the config object from the beets module. That is, just put `from beets import config` at the top of your plugin and access values from there.

While we're operating outside the context of a plugin here, it seems like this is intended to work even outside a plugin's realm.

wisp3rwind: > While we're operating outside the context of a plugin here, it seems like this is intended to work even outside a plugin's realm.

I'm not convinced ;) The thing is, for plugins I can see why it works: Plugins are imported manually in `beets.ui._load_plugins`. At that point, the main beets modules have surely been imported and set up. However, inside of beets core, that might not be the case. In fact, in `beets.ui._configure`, an overlay configuration given through the `--config` option might be applied. `beets.ui` already imports `beets.util` before that, so values from the overlay configuration will never make it to the `remove` kwarg default. For the "normal" non-overlay config I'm not sure, I haven't looked into confuse in a while, so I can't see right now when that would be loaded.

justinmayer: I agree that changing the interface to `use_trash=True` is an improvement, so I just pushed an updated commit that switches to that argument. The `remove()` function will use an alternate command if a `remove_command` value is found in the configuration, but now otherwise falls back to `trash`, thereby improving durability. I believe these changes constitute a reasonable compromise solution to the concerns raised here.

beets/util/__init__.py
```diff
@@ -442,17 +450,36 @@ def samefile(p1, p2):
...
-    except (OSError, IOError) as exc:
-        raise FilesystemError(exc, 'delete', (path,), traceback.format_exc())
+    # Use `remove_command` setting, if available. Otherwise assume `trash`.
+    if config["remove_command"]:
+        remove_command = config["remove_command"].as_str()
+    else:
+        remove_command = "trash"
+    # Use trash if `remove_command` is found on $PATH, else delete immediately.
+    # Change which() to shutil.which() once Python <3.3 support is dropped
+    if use_trash and which(remove_command) is not None:
```
wisp3rwind, line 469:
```suggestion
    # Use trash if `remove_command` (which defaults to `trash`) is found on $PATH,
    # else delete immediately.
    # Change which() to shutil.which() once Python <3.3 support is dropped
    remove_command = config["remove_command"].as_str()
    if use_trash and which(remove_command) is not None:
```

I think with confuse, you can immediately access the value: The `config_default.yaml` file will always be loaded, so this will return the default `trash` if the user didn't specify anything else. Accessing the configuration like this is a common pattern in beets.

wisp3rwind (changes requested):
I left one more comment about a slight simplification that can be made; I'll have a look at callers which need to opt-out of trash handling later today.

Otherwise, this is looking good :+1:

beets/util/__init__.py
justinmayer: Agreed. I removed the extraneous fallback in the latest commit. Aside from any areas where you think opting out from removal to trash is needed, I believe we are all good to go here.

sampsyo:
Hi, folks! Thanks for making so much progress on this; it looks fantastic so far.

Just to check, however: it looks like the current implementation will attempt to use `trash` by default, if the user doesn't do anything else? I think we should probably not do that—unless I'm reading it wrong, this would lead to errors on machines without `trash` installed in the current setup. And even if we turn this into a silent fallback, it will change the behavior for current users. Perhaps I'm misreading things, but maybe the default for `remove_command` should just be empty to preserve the current behavior?

It also occurs to me that the current structure of the `remove` function now does ~completely separate things depending on whether `trash=True`. Maybe it would be a little more straightforward to instead just have a separate function called `trash`, and to leave `remove` to mean "actually delete"? The removal "mode" would then be dictated by the client, which is what we already do for moving/copying/etc.:
https://github.com/beetbox/beets/blob/84f6d374c4cb918bb77bccb11593cd4765ebfa54/beets/library.py#L793

justinmayer:
Hey Adrian! Thanks for the kind words.

> It looks like the current implementation will attempt to use `trash` by default, if the user doesn't do anything else?

Yes — the idea is that this is a safer default for end users, eliminating potential foot-guns caused by unexpected and/or aberrant behavior, which can often result in catastrophic data loss. (See related linked issue in the description.)

> Unless I'm reading it wrong, this would lead to errors on machines without `trash` installed in the current setup.

I think you might be reading it wrong? There shouldn't be any errors in that situation. If `trash` (or another command defined in `remove_command`) is not found on `$PATH`, then `os.remove()` is used to immediately delete the file (i.e., the current default behavior).

> […] it will change the behavior for current users. Perhaps I'm misreading things, but maybe the default for `remove_command` should just be empty to preserve the current behavior?

I believe that (1) the safest default is the best default and (2) this safety outweighs preserving the less-safe current behavior. Plus, it only changes the behavior for current users who have done out of their way to install a `trash` tool, and those are likely the kind of users who will value the extra safety this new default removal behavior provides.

wisp3rwind:
Personally, I agree with @nathdwek in https://github.com/beetbox/beets/issues/149
> As a mainly linux guy, I didn't and still don't really feel the need for recycle-binning.

and TBH, I don't care what the default is, which is why I didn't press the point here. For my own workflow, I'm rarely in a situation where beets is deleting files anyway. This feature request (and the alleged footguns when using beets) seems to be brought up persistently (by a small number of people, however), so if trashing by default helps calming the waves, I'll just add `remove_command=""` to my configuration.


> Just to check, however: it looks like the current implementation will attempt to use trash by default, if the user doesn't do anything else? I think we should probably not do that—unless I'm reading it wrong, this would lead to errors on machines without trash installed in the current setup. And even if we turn this into a silent fallback, it will change the behavior for current users. Perhaps I'm misreading things, but maybe the default for remove_command should just be empty to preserve the current behavior?

I think @justinmayer is right in that this is currently already a silent fallback. I was suprised about that, too, see https://github.com/beetbox/beets/pull/3796#discussion_r591695157, but as outlined above, I won't be investing energy into discussing this point. Ideally of course, that would be polled on the forum.

> It also occurs to me that the current structure of the remove function now does ~completely separate things depending on whether trash=True. Maybe it would be a little more straightforward to instead just have a separate function called trash, and to leave remove to mean "actually delete"? The removal "mode" would then be dictated by the client, which is what we already do for moving/copying/etc.:

I'm not sure that is useful: In the example you made, there is still a `def move_file(self, dest, operation=MoveOperation.MOVE)` top-level function, that switches between the various modes. That appears very similar to `def remove(..., use_trash=True)` with a boolean instead of the enum.

In any case, this needs to go with https://github.com/beetbox/beets/pull/3876, lest we litter the trash with unexpected temporary files. I'd appreciate if someone (...) could have a brief look whether the additional commit in that PR looks sensible.

sampsyo:
Awesome; thanks. Glad that I totally misread the fallback thing! Sorry about that.

As for the other two issues:

* Thanks for clarifying about the intent to trash by default. I am still leaning toward *not* doing this and deleting by default. We already have several safeguards in place, such as not actually deleting unless you add `-d` to `beet rm` and then requiring a `Y` to confirm the action. In a sort of classically Unixy UI sense, I think it would be somewhat surprising that an action that's supposed to delete files actually calls a binary named `trash` instead. I really like having this as an option, but I think it makes the most sense to have it off by default.
* About whether to have the "trash decision" internal to `remove` or external in the client: thanks to @wisp3rwind for putting together #3876! As that PR illustrates, it seems like the places in the code where we actually want to remove files may actually be _more_ numerous than the places where we want to conditionally remove-or-trash files (which I imagine are only in `library.py`'s `remove` methods, which are already meant to centralize logic for removing actual "content" files, i.e., music and album art). So I'm worried that having `use_trash=True` be the default will lead to confusing bugs in the future when we add code that wants to do "normal" removals. I don't really object to having the check in `util` instead of `library`, but it still seems marginally more natural to me to do the latter? I dunno, no need to bikeshed and I don't feel terribly strongly about it, but maybe the above reasons will make a little bit of sense.

RollingStar:
Unless we're sure that beets' trash-related code is 100% bug-free, it strikes me as arrogant to not make recycle bin / trash the default. https://github.com/beetbox/beets/issues/2499

Add in an unbounded number of plugins that may also want to delete files, and it's a big source of errors for no benefit.

If your OS is that good, just automate the trash removal. I think mine does that already so there's no overhead to me as an end user, just more safety.

I skimmed the above conversation, so maybe I missed it: what is gained by not having trash functionality? If it just tries to call an outside program `trash` and silently falls back to the old behavior, what's the harm? Why do we need to avoid this?

sampsyo:
My argument for having this off by default boils down to [the principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment): is that beets is a Unix CLI tool, and the "normal" thing for a Unix CLI tool to do is to delete files when you type the `rm` command. Invoking an external `trash` binary is undeniably a useful thing to do for some users, but it is also surprising—no other Unix CLI tool I am familiar with works that way. If you have a weird binary sitting around in your $PATH called `trash` that does something else entirely, I think it would be pretty surprising for a seemingly-minor beets feature to suddenly start invoking it. And if the `trash` program you have fails for some reason, that's not really something we can recover from—so it's an added layer of complexity.

So I'm very enthusiastic about the functionality being available, but I still think it should be off by default.

wisp3rwind:
Having thought about it again, I probably care more than I realised, and I think I'm seeing a clear path now on how this should be approached:
- Change the interface to `def remove(..., use_trash=False)`, and in turn adapt https://github.com/beetbox/beets/pull/3876
- Make `remove_command=""` the default for now
- Merge this and https://github.com/beetbox/beets/pull/3876
- Open a new issue about changing the default for `remove_command`, which we review only after trash functionality has been in a release or two, so that we can have some more confidence in the correctness of the trash-related code. From what I can tell, that hasn't seen any relevant real-world usage right now.

The basic idea behind this would be that we can always change the defaults to a "safer" value, but should really avoid doing the reverse (e.g. because we found some regression). The latter would break expectations of anyone who comes to rely on trashing over direct deletion. I'm also under the impression that we're not going to agree on the default here, and I'd like to avoid blocking to merge this PR. There's of course the risk that when reviewing that new issue we will not be able to resolve the disagreement.

---

> Thanks for clarifying about the intent to trash by default. I am still leaning toward not doing this and deleting by default. We already have several safeguards in place, such as not actually deleting unless you add -d to beet rm and then requiring a Y to confirm the action.

I didn't even think about those safeguards, sounds like a good reason :+1:

> In a sort of classically Unixy UI sense, I think it would be somewhat surprising that an action that's supposed to delete files actually calls a binary named trash instead. I really like having this as an option, but I think it makes the most sense to have it off by default.

I'm not sure how large the actual risk is that there's indeed another program named `trash` that does not have the intended effect (how about Windows, though?). Still, one more reason to roll this out slowly.

> About whether to have the "trash decision" internal to remove or external in the client: thanks to @wisp3rwind for putting together #3876! As that PR illustrates, it seems like the places in the code where we actually want to remove files may actually be more numerous than the places where we want to conditionally remove-or-trash files (which I imagine are only in library.py's remove methods, which are already meant to centralize logic for removing actual "content" files, i.e., music and album art). So I'm worried that having use_trash=True be the default will lead to confusing bugs in the future when we add code that wants to do "normal" removals.

Yes, I now agree that trash should be opt-in, for the reasons scattered throughout this comment.

> I don't really object to having the check in util instead of library, but it still seems marginally more natural to me to do the latter? I dunno, no need to bikeshed and I don't feel terribly strongly about it, but maybe the above reasons will make a little bit of sense.

I believe that having it in `util` is useful, since it makes the trashing code available to plugins. There is one user already: The convert plugin, which might delete music on import when `delete_originals=True` (although one might argue that precisely because of this setting, trash should be bypassed...).

---

> Unless we're sure that beets' trash-related code is 100% bug-free, it strikes me as arrogant to not make recycle bin / trash the default. #2499

That seems like a contradiction. I'm quite confident about the current "trash-related" PR. I'm by no means 100% sure that there's not a problem with an edge case we didn't think about (where, say, beets crashes and leaves you with some half-imported items in the database). There is a value to introducing it somewhat conservatively.

> Add in an unbounded number of plugins that may also want to delete files, and it's a big source of errors for no benefit.

There is a very significant benefit: We should not ever move temporary files that we created ourselves to the trash. Such files with semi-random names popping up in the recycle bin are absolutely a cause for confusion, and exactly that will happen when implementing `def remove(..., use_trash=True)`.
I'm not convinced that preventing external plugins from doing dangerous things is a relevant concern. We can't prevent them from `os.remove`ing your data directly, anyway. Any (or at the least, most) plugins trying to delete your music directly is wrong already, since it should have used the `Item.remove` method. It could be worthwhile to add a section to https://github.com/beetbox/beets/blob/master/docs/dev/plugins.rst to explain usage of the `remove` function.

> I skimmed the above conversation, so maybe I missed it: what is gained by not having trash functionality? If it just tries to call an outside program trash and silently falls back to the old behavior, what's the harm? Why do we need to avoid this?

In addition to @sampsyo's answer, I think there simply is no solution here that will suit everyone's workflow out of the box. Why should "absolutely everything needs to go to the trash" be the only viable way forward?

---

FWIW, if we don't go by the steps outlined above, I think this PR needs one more change: I did account for direct calls to `util.remove` in the tests in https://github.com/beetbox/beets/pull/3876, but that misses anything going through, for example, the importer. For our tests, we should ensure that running the them doesn't move anything to the actual trash on the machine.

RollingStar:
Windows doesn't have that "rm = instant delete" paradigm to my knowledge, even in the CLI. Admittedly, Windows is half-supported right now (who on the dev team is enthusiastically using beets on Windows?). IIRC, there has been some discussion of just moving to the Windows Subsystem for Linux.

For temp files, Windows has the temporary files folder. This is all much lower level than I'm familiar with, but for temp files wouldn't the program be doing this anyway? https://en.wikipedia.org/wiki/Temporary_file#Creation

> There is a value to introducing it somewhat conservatively.

Fair enough. I guess what I mean is, what does this look like 1-3 years from now? For me, trash-by-default is the most user-friendly and indeed "least astonishing" behavior. rm == delete on unix, but for someone not used to the unix command line (ex. Windows user), it could just as easily be **r**e**m**ove. Even for me, knowing the rm command on Linux, I wouldn't have guessed that it deletes files irreversibly. That's not something I'm used to non-system programs doing (with limited Linux command line experience). Like, nothing I can do in Firefox Ubuntu will end with me deleting files without stopping in the trash first. It's splitting hairs though, if the config supports trashing, I can't complain much.

In my 4 years of using beets, I've had bugs cause beets to remove files that shouldn't have been removed, but I've never thought "I need these music files off my PC **now**, no backsies".

sampsyo:
Thanks, @wisp3rwind; that plan sounds great!

> I believe that having it in `util` is useful, since it makes the trashing code available to plugins. There is one user already: The convert plugin, which might delete music on import when `delete_originals=True` (although one might argue that precisely because of this setting, trash should be bypassed...).

That's a good reason! I too am not exactly sure what to do about that `convert` setting, but it is very easy to imagine other scenarios in plugins that want the same behavior.

> FWIW, if we don't go by the steps outlined above, I think this PR needs one more change: I did account for direct calls to `util.remove` in the tests in #3876, but that misses anything going through, for example, the importer. For our tests, we should ensure that running the them doesn't move anything to the actual trash on the machine.

Also a very good point (and a reason to keep `util.remove` as actually-deleting by default).

> In my 4 years of using beets, I've had bugs cause beets to remove files that shouldn't have been removed, but I've never thought "I need these music files off my PC **now**, no backsies".

Unfortunately, @RollingStar, I think there's a chance that these frustrating accidental deletions would not have been solved by a "trash" precaution. It's just as easy for a bug in beets to lose data by overwriting a file via a move, for instance, that would have slipped by this specific change to the `util.remove` function. So as inconvenient as it may be, I think the real solution here is to continue fixing those bugs…

wisp3rwind:
> Unfortunately, @RollingStar, I think there's a chance that these frustrating accidental deletions would not have been solved by a "trash" precaution. It's just as easy for a bug in beets to lose data by overwriting a file via a move, for instance, that would have slipped by this specific change to the `util.remove` function. So as inconvenient as it may be, I think the real solution here is to continue fixing those bugs…

Now you could argue that file moves should do backups first :) I don't have an overview over the situations we do moves in, so that might actually be worth discussing in a future issue/PR. In any case, let's get the uncontroversial part of the current PR merged. @justinmayer, how does the plan above sound to you? Do you have the time and energy to modify the defaults accordingly?

justinmayer:
> How does the plan above sound to you? Do you have the time and energy to modify the defaults accordingly?

@wisp3rwind: Assuming we're talking about the first two bullets points in your [previous comments](https://github.com/beetbox/beets/pull/3796#issuecomment-798557526), then yes — I should be able to make time to update this PR with those changes. I'm traveling at the moment but will do my best to get that done by the end of June.

wisp3rwind:
> > How does the plan above sound to you? Do you have the time and energy to modify the defaults accordingly?
> 
> @wisp3rwind: Assuming we're talking about the first two bullets points in your [previous comments](https://github.com/beetbox/beets/pull/3796#issuecomment-798557526), then yes — I should be able to make time to update this PR with those changes. I'm traveling at the moment but will do my best to get that done by the end of June.

Yep, exactly, that'd be great! Don't hurry, enjoy the travelling!

justinmayer:
@sampsyo: Can you re-open this? I'll work on it this week.

sampsyo:
Certainly!

PR #3807, opened 2020-12-10
chore: makes discussions more visible

jef:
## Description

I created a `config.yml` thats helps make [Discussions](https://github.com/beetbox/beets/discussions) more visible.

Example (from another repository):

image

Open links point to Discussions.

I have also removed feature request as it seems more fitting to have Discussions about a feature and then use the [Issues](https://github.com/beetbox/beets/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc) for valid tickets/things being worked on.


sampsyo:
Cool; thanks for getting this going!

FWIW, I do think it's cool to have tickets for feature requests as well as bugs in the tracker… but I get your intent to put vague "ideas" on the discussion board. Maybe we should keep the template around but just clarify that it's for cases where you know pretty much what you want and are ready to write a coherent proposal, whereas discussions are for when you just want to chat about an idea?

jef:
What do we think, should we add this in?

sampsyo:
Oh yeah, sounds good to me! But maybe we should leave the Discourse link for now just to see how things go?

jef:
> Oh yeah, sounds good to me! But maybe we should leave the Discourse link for now just to see how things go?

I agree!

sampsyo:
Nice; looks pretty good to me! And I think the CI failures may be resolved my updating with master? In any case, I think it's about good to go.

jef:
> And I think the CI failures may be resolved my updating with master

Not sure why, I can rerun if needed.

sampsyo:
Odd… just to be sure, maybe we should try merging #3907 first and seeing if that helps clear things up on this PR?

jef:
I also did a merge. I didn't realize that this was roughly _300 commits_ behind. Probably had something to do with it 😅

sampsyo:
Ah, that'd do it! LGTM, except that it looks like maybe the feature request template file might have gotten accidentally deleted?

jef:
Ah, good question. I removed it on purpose. Since the `config.yaml` will point folks to the Discussion board, they'll be able to discuss a feature in length. If we decide to pull it in, we can create put it on the board in a way we like.

This doesn't prohibit the user to define a feature request either way, but just leads them in the direction.

If you don't like this approach, I have no problem reverting.

sampsyo:
Got it! I don't object, certainly. Here's a question, though: when a feature request eventually makes its way over from vague discussion to a concrete proposal, will we want to follow a template then? (One option might be to add an encouragement to do discussion ahead of time to the template itself.)

jef:
Aha, makes sense to me. The feature issue could definitely lose it's structure overtime, so that's good idea! I'll update.

sampsyo:
Nice. This all looks good to go from here!

PR #3810, opened 2020-12-15
Use caching in plugins.find_plugins()

ssssam:
This function is called from library.Item._getters(), making it a hot
path when exporting data. We cache plugin instances but previously we
reran `import` for each module on each call. We now return the cached
values providing a speed boost for `beet export`.

This has a small effect on `beet ls` speed: for me it went 11.00s ->
10.40.

It had a significant effect on `beet export` speed when combined with
https://github.com/beetbox/beets/pull/3762/.

Before:

    $ time /home/sam/.local/bin/beet 'export' '--library' '--format' 'jsonlines' '--include-keys' 'artist,title,path,mb_artistid,mb_trackid' 'artist+ title+' > /dev/null
    Executed in   25.13 secs

After:

    $ time /home/sam/.local/bin/beet 'export' '--library' '--format' 'jsonlines' '--include-keys' 'artist,title,path,mb_artistid,mb_trackid' 'artist+ title+' > /dev/null
    Executed in   10.49 secs

## To Do

### DIFF
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -301,6 +301,11 @@ def find_plugins():
     currently loaded beets plugins. Loads the default plugin set
     first.
     """
+    if _instances:
+        # After the first call, use cached instances for performance reasons.
+        # See https://github.com/beetbox/beets/pull/3810
+        return list(_instances.values())
+
     load_plugins()
     plugins = []
     for cls in _classes:


sampsyo:
Neat! I think this is probably the right thing to do. But just to talk over the risks here… the potential circumstance where this could go wrong is some hypothetical situation where `find_plugins` gets called twice, once before and once after loading a given plugin module. The second one would fail to find the new plugin. The only circumstances I can think of where that would be realistic are:

1. In testing. (In fact, I'm a little surprised this doesn't expose any weirdness in our test suite—I guess we're just careful enough about cleaning up `_instances` already?)
2. When using beets as a library (which is not common).

Anyway, the "right" refactoring we should do here is to avoid module-level state and have some sort of class called `PluginManager` or something that keeps track of all the registered plugin classes and instances. That would be shared by a given beets execution, and programs embedding beets as a library would need to instantiate one too. This would encapsulate the global state and provide mechanisms for resetting stuff.

Absent that, however, I think this kind of caching behavior is unlikely to do any real harm.

Just one suggestion: would you mind adding a quick comment there saying why that check exists? (That is, it's "just" a performance optimization.) Agreed, no changelog entry necessary.

ssssam:
OK, done! Thanks for the quick feedback.

sampsyo:
Great; thanks!

PR #3814, opened 2020-12-19
Add ISRC identifiers from musicbrainz.

aereaux:
## Description
This adds ISRC information as a tag.  There are a couple of TODO I could use input on.

## To Do

- [ ] How to manage the issue where there are multiple related to a track.  Right now I just ';'-join them, but we could just take the first or something.
- [ ] How to add to files?  There is apparently a tag for it in common formats (https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#id12), but I'm not sure if anything needs to be added to mediafile.

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -71,14 +71,14 @@ log = logging.getLogger('beets')
 RELEASE_INCLUDES = ['artists', 'media', 'recordings', 'release-groups',
                     'labels', 'artist-credits', 'aliases',
                     'recording-level-rels', 'work-rels',
-                    'work-level-rels', 'artist-rels']
+                    'work-level-rels', 'artist-rels', 'isrcs']
 BROWSE_INCLUDES = ['artist-credits', 'work-rels',
                    'artist-rels', 'recording-rels', 'release-rels']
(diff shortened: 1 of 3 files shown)


sampsyo:
Neat! This seems like a good start. I don't have a strong feeling about what to do with multiple ISRCs—joining with `;` seems fine to me, but maybe others have more specific opinions.

It's possible to merge this independently of the necessary changes to MediaFile to get the tags written to metadata. But we will indeed need to add a tag mapping there to make that work.

aereaux:
Is there anything else that needs to be done before this is merged in beets?  I added a changelog entry, and it seems to me that using `;` is the best solution unless/until we get another way to have multi-valued tags.

sampsyo:
Awesome; thanks for adding that! This looks good to go.

JuniorJPDJ:
Multiple ISRCs? Isn't ISRC representing recording 1:1? Like one track can have only one ISRC?

aereaux:
@JuniorJPDJ: According to [here](https://usisrc.org/faqs/assignment.html) it is supposed to be.  But MB certainly allows many-many, such as https://beta.musicbrainz.org/recording/df569676-9c08-4bec-8137-daa122c04ca9 which has 5.  One reason might be multiple remasters that are counted in MB as one recording.

Issue #3809, opened 2020-12-15
ReplayGainPlugin throws false alarm sqlite3.OperationalError in test suite


ybnd:
### Problem
The tests in `test_replaygain.py` work with an in-memory SQLite database, which doesn't seem to play well with the parallelization introduced in https://github.com/beetbox/beets/pull/3478.

* [`item.store()`](https://github.com/beetbox/beets/blob/8645f56512fa8918fd05c4f53bd1b65ace87ab41/beetsplug/replaygain.py#L1311) raises

    ```
    Exception in thread Thread-90:
    Traceback (most recent call last):
      File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
        self.run()
      File "/home/ybnd/code/beets/beets/beetsplug/replaygain.py", line 1198, in run
        six.reraise(exc[0], exc[1], exc[2])
      File "/home/ybnd/.local/share/venve/beets/lib/python3.8/site-packages/six.py", line 703, in reraise
        raise value
      File "/home/ybnd/code/beets/beets/beetsplug/replaygain.py", line 1510, in wfunc
        return func(*args, **kwargs)
      File "/home/ybnd/code/beets/beets/beetsplug/replaygain.py", line 1460, in _store_track
        store_track_gain(item, track_gains[0])
      File "/home/ybnd/code/beets/beets/beetsplug/replaygain.py", line 1324, in store_track_gain
        self._store(item)
      File "/home/ybnd/code/beets/beets/beetsplug/replaygain.py", line 1319, in _store
        item.store()
      File "/home/ybnd/code/beets/beets/beets/library.py", line 347, in store
        super(LibModel, self).store(fields)
      File "/home/ybnd/code/beets/beets/beets/dbcore/db.py", line 516, in store
        tx.mutate(query, subvars)
      File "/home/ybnd/code/beets/beets/beets/dbcore/db.py", line 838, in mutate
        cursor = self.db._connection().execute(statement, subvals)
    sqlite3.OperationalError: no such table: items
    ```

  on the first call, **but just succeeds on the second call**

* Occurs on every call to `item.store()`
* Doesn't depend on the ReplayGain backend you use

### Reproducing

1. Make sure `test_replaygain.py` actually runs some of the tests (most are skipped in the CI build)
2. Comment out [the `_store_retry_once` patch](https://github.com/beetbox/beets/blob/8645f56512fa8918fd05c4f53bd1b65ace87ab41/test/test_replaygain.py#L82)
[...]

PR #3817, opened 2020-12-21
replaygain: Use on-disk database

wisp3rwind:
Removes the `_store_retry_once` hack in favor of using an on-disk database, we have support for that in the test harness already.

Accessing the database from a new thread opens a new connection, with the `:memory:` path, that actually creates a per-thread database. Presumably, even with the `OperationalError`, the first store actually created the missing the table which let the second store succeed.

Closes: #3809


sampsyo:
Perfect. Thanks so much—I should have thought of this as the solution originally.

PR #3818, opened 2020-12-21
Disable replaygain parallelism during import

ybnd:
## Description

Add an additional config parameter to enable parallelism on import. Defaults to `False` because this functionality is unpredictable and should be opt-in for now.

### DIFF
--- a/beetsplug/replaygain.py
+++ b/beetsplug/replaygain.py
@@ -1234,6 +1234,7 @@ class ReplayGainPlugin(BeetsPlugin):
             'auto': True,
             'backend': u'command',
             'threads': cpu_count(),
+            'parallel_on_import': False,
             'per_disc': False,
             'peak': 'true',
             'targetlevel': 89,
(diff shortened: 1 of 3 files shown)


beetsplug/replaygain.py
```diff
@@ -1553,8 +1554,10 @@ def close_pool(self):
     def import_begin(self, session):
         """Handle `import_begin` event -> open pool
         """
-        if self.config['auto']:
-            self.open_pool(self.config['threads'].get(int))
+        threads = self.config['threads'].get(int)
+
+        if self.config['parallel_on_import'] and threads:
```
sampsyo, line 1559: Should this also include `self.config['auto']` to avoid running the pool if we're not supposed to do RG at all?

sampsyo (commented):
Excellent; thanks for the quick fix. Just one idea within.

beetsplug/replaygain.py
ybnd: Yes it should, thanks!

sampsyo:
Perfect; thanks!

PR #3820, opened 2020-12-24
fix youtube video when using https

capital-G:
## Description

Should fix youtube video now shown on github.io page when using https due to using http within https

https://beets.io/

vs

http://beets.io/

(...)

## To Do


sampsyo:
Thank you!!

PR #3822, opened 2020-12-30
Fetch data for releases with more than 500 tracks

### LINKED ISSUE #3308: Some releases don't get their works fetched
### Problem

I implemented the ```work```, ```mb_workid``` and ```work_disambig``` tags not so long ago (https://github.com/beetbox/beets/pull/3272) . My problem is: for some recordings, the works just don't get fetched. It concerns especially the very big releases (20+ CDs like for example https://musicbrainz.org/release/9c5c043e-bc69-4edb-81a4-1aaf9c81e6dc or https://musicbrainz.org/release/9bcd75dd-995e-482b-8ba7-1ef074d253de ). I tried to backtrace the error (by putting random prints in the ```beets/autotag/mb.py``` and then  running ```beet mbsync``` on the problematic releases). 
What I can see is: 

while ```RELEASE_INCLUDES``` (in ```beets/autotag/mb.py```) does contain ```'work-rels', ``` and ```'work-level-rels'```, for the releases I'm looking at, ```TRACK_INCLUDES``` doesn't: ```musicbrainzngs.VALID_INCLUDES['recording']``` contains ```'work-rels'``` but not ```'work-level-rels'```, which is odd. 
At line 494 in ```beets/autotag/mb.py```, ```musicbrainzngs.get_release_by_id(albumid,RELEASE_INCLUDES)``` doesn't contain any works, even if the recordings do have works and  ```RELEASE_INCLUDES``` contains ```'work-rels', ``` and ```'work-level-rels'```. I first tried to check ```musicbrainzngs.get_recording_by_id(recording['id'], TRACK_INCLUDES)``` for all the recordings and it turns out it doesn't contain any works, because of the first error. If now I add ```'work-rels'``` to ```TRACK_INCLUDES``` and then look at ```musicbrainzngs.get_recording_by_id(recording['id'], TRACK_INCLUDES)``` then it contains the works just fine. 

So I'm wondering: why do we get the work relationships with ```musicbrainzngs.get_recording_by_id(recording['id'], TRACK_INCLUDES)``` but not with 
```musicbrainzngs.get_release_by_id(albumid, RELEASE_INCLUDES)``` even if both ask for  ```'work-rels``` and ```work-level-rels``` for ```musicbrainzngs.get_release_by_id``` ?

A quick and dirty fix would be to ask for ```musicbrainzngs.get_recording_by_id(recording['id'], TRACK_INCLUDES)``` for each track. The problem is, there is a significant performance loss if we have one musicbrainz query for each track instead of each release, but I didn't look at it in too much detail. It seems to me that ```musicbrainzngs``` doesn't send all info we ask for for very big releases, could that be because it is too big and they have a cap on the maximum size they can send? 

Of course, I checked: the concerned recordings do have works on MB. 

### Setup

* OS: Ubuntu
* Python version: 2.7.15
* beets version: 1.50

```yaml
asciify_paths: yes
bell: yes
color: yes
alternatives:
    by-work:
        directory: ../beets_by-work

        paths:
            albumartist_sort::'Wise Guys': ZZZ_Non-classical/$albumartist_sort/$album %ifdef{albumdisambig,($albumdisambig),}/$disc-$track $title
            ...
        formats: link
library: /media/soergeld/EEC7-74A2/Musique/beets/Musicdata_sync.db
embedart:
    remove_art_file: yes
    compare_threshold: 0
    auto: yes
    ifempty: no
    maxwidth: 0

plugins: duplicates info mbsync badfiles fromfilename mbsubmit mbcollection absubmit fetchart embedart chroma parentwork alternatives

musicbrainz:
    user: dosoe
    pass: REDACTED
absubmit:
    auto: no
    extractor: /home/soergeld/Personnel/acousticbrainz-client/streaming_extractor_music

paths:
    default: $albumartist_sort/$album/$disc-$track $title
    singleton: ZZZ_noalbum/$artist_sort/$title
parentwork:
    force: no
    auto: no
missing:
    count: yes
chroma:
    auto: no
duplicates:
    tiebreak:
        items: [bitrate]
    keys:
    - mb_releasegroupid
    - track
    - disc
    - title
    count: no
    full: no
    format: ''
    move: ''
    tag: ''
    path: no
    copy: ''
    album: no
    strict: no
    checksum: ''
    merge: no
    delete: no

ui:
    color: yes
    colors:
        text_success: green
        text_warning: yellow
        text_error: red
        text_highlight: red
        text_highlight_minor: lightgray
        action_default: turquoise
        action: blue
directory: /media/soergeld/EEC7-74A2/Musique/beets

import:
    copy: no
    write: no
    incremental: no
    detail: yes
    languages: en
    duplicate_action: keep
    per_disc_numbering: yes
    bell: yes
acoustid:
    apikey: REDACTED
mbsubmit:
    format: $title - $artist ($length)
    threshold: strong
mbcollection:
    auto: no
    remove: no
    collection: ''
fetchart:
    auto: yes
    minwidth: 0
    sources:
    - filesystem
    - coverart
    - itunes
    - amazon
    - albumart
    enforce_ratio: no
    cautious: no
    maxwidth: 0
    store_source: no
    google_key: REDACTED
    fanarttv_key: REDACTED
    cover_names:
    - cover
    - front
    - art
    - album
    - folder


```

dosoe:
## Description

Fixes #3308. 
```musicbrainzngs.get_release_by_id``` doesn't provide all data beets asks for when a release has more than 500 tracks. I use ```musicbrainzngs.browse_recordings``` as a complement to ```musicbrainzngs.get_release_by_id``` for releases with more than 500 tracks to get the missing data. It doesn't work perfectly yet, mainly because ```musicbrainzngs.browse_recordings``` does not allow to fetch quite the same data than  ```musicbrainzngs.get_release_by_id``` (notably, ```work_level_rels``` and ```aliases``` are missing for reasons I don't quite understand) but it's already much better and on the day ```musicbrainzngs.browse_recordings``` does fetch this data, the update should be trivial. 

## To Do

- [ ] Tests. (What should I test? Create a mock release with more than 500 tracks?)

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -74,6 +74,7 @@ RELEASE_INCLUDES = ['artists', 'media', 'recordings', 'release-groups',
 BROWSE_INCLUDES = ['artist-credits', 'work-rels',
                    'artist-rels', 'recording-rels', 'release-rels']
 BROWSE_CHUNKSIZE = 100
+v = 500
 TRACK_INCLUDES = ['artists', 'aliases']
 if 'work-level-rels' in musicbrainzngs.VALID_INCLUDES['recording']:
     TRACK_INCLUDES += ['work-level-rels', 'artist-rels']
@@ -293,7 +294,7 @@ def album_info(release):
     # The MusicBrainz API omits 'artist-relation-list' and 'work-relation-list'
     # when the release has more than 500 tracks. So we use browse_recordings
     # on chunks of tracks to recover the same information in this case.
-    if ntracks > 500:
+    if ntracks > BROWSE_MAXTRACKS:
         recording_list = []
         for i in range(0, ntracks, BROWSE_CHUNKSIZE):
             recording_list.extend(musicbrainzngs.browse_recordings(
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -296,18 +296,19 @@ def album_info(release):
     # on chunks of tracks to recover the same information in this case.
     if ntracks > BROWSE_MAXTRACKS:
         log.debug(u'Album {} has too many tracks', release['id'])
-        log.info(u'Fetching recordings in batches of ' + str(BROWSE_CHUNKSIZE))
         recording_list = []
         for i in range(0, ntracks, BROWSE_CHUNKSIZE):
+            log.debug(u'Retrieving tracks starting at {}', i)
             recording_list.extend(musicbrainzngs.browse_recordings(
                     release=release['id'], limit=BROWSE_CHUNKSIZE,
                     includes=BROWSE_INCLUDES,
-                    offset=BROWSE_CHUNKSIZE * i)['recording-list'])
+                    offset=i)['recording-list'])
+        track_map = {}
+        for recording in recording_list:
+            track_map[recording['id']] = recording
         for medium in release['medium-list']:
             for recording in medium['track-list']:
-                recording_info = list(filter(lambda track: track['id'] ==
-                                             recording['recording']['id'],
-                                             recording_list))[0]
+                recording_info = track_map[recording['recording']['id']]
                 recording['recording'] = recording_info
 
     # Basic info.
@@ -542,7 +543,6 @@ def album_for_id(releaseid):
     try:
         res = musicbrainzngs.get_release_by_id(albumid,
                                                RELEASE_INCLUDES)
-        log.info(u'Album ' + str(releaseid) + u' fetched from MusicBrainz')
     except musicbrainzngs.ResponseError:
         log.debug(u'Album ID match failed.')
         return None
@@ -562,7 +562,6 @@ def track_for_id(releaseid):
         return
     try:
         res = musicbrainzngs.get_recording_by_id(trackid, TRACK_INCLUDES)
-        log.info(u'Track ' + str(releaseid) + u' fetched from MusicBrainz')
     except musicbrainzngs.ResponseError:
         log.debug(u'Track ID match failed.')
         return None
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -73,7 +73,6 @@ RELEASE_INCLUDES = ['artists', 'media', 'recordings', 'release-groups',
                     'work-level-rels', 'artist-rels']
 BROWSE_INCLUDES = ['artist-credits', 'work-rels',
                    'artist-rels', 'recording-rels', 'release-rels']
-
 TRACK_INCLUDES = ['artists', 'aliases']
 if 'work-level-rels' in musicbrainzngs.VALID_INCLUDES['recording']:
     TRACK_INCLUDES += ['work-level-rels', 'artist-rels']
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -288,9 +288,7 @@ def album_info(release):
     artist_name, artist_sort_name, artist_credit_name = \
         _flatten_artist_credit(release['artist-credit'])
 
-    ntracks = 0
-    for medium in release['medium-list']:
-        ntracks += len(medium['track-list'])
+    ntracks = sum(len(m['track-list']) for m in release['medium-list'])
 
     # The MusicBrainz API omits 'artist-relation-list' and 'work-relation-list'
     # when the release has more than 500 tracks. So we use browse_recordings
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -295,7 +295,7 @@ def album_info(release):
     # when the release has more than 500 tracks. So we use browse_recordings
     # on chunks of tracks to recover the same information in this case.
     if ntracks > BROWSE_MAXTRACKS:
-        log.info(u'Album ' + str(release['id']) + u' has too many tracks')
+        log.debug(u'Album {} has too many tracks', release['id'])
         log.info(u'Fetching recordings in batches of ' + str(BROWSE_CHUNKSIZE))
         recording_list = []
         for i in range(0, ntracks, BROWSE_CHUNKSIZE):
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -295,6 +295,8 @@ def album_info(release):
     # when the release has more than 500 tracks. So we use browse_recordings
     # on chunks of tracks to recover the same information in this case.
     if ntracks > BROWSE_MAXTRACKS:
+        log.info(u'Album '+str(trackid)+u' has too many tracks')
+        log.info(u'Fetching recordings in batches of '+str(BROWSE_CHUNKSIZE))
         recording_list = []
         for i in range(0, ntracks, BROWSE_CHUNKSIZE):
             recording_list.extend(musicbrainzngs.browse_recordings(
@@ -540,6 +542,7 @@ def album_for_id(releaseid):
     try:
         res = musicbrainzngs.get_release_by_id(albumid,
                                                RELEASE_INCLUDES)
+        log.info(u'Album '+str(trackid)+u' fetched from MusicBrainz')
     except musicbrainzngs.ResponseError:
         log.debug(u'Album ID match failed.')
         return None
@@ -559,6 +562,7 @@ def track_for_id(releaseid):
         return
     try:
         res = musicbrainzngs.get_recording_by_id(trackid, TRACK_INCLUDES)
+        log.info(u'Track '+str(trackid)+u' fetched from MusicBrainz')
     except musicbrainzngs.ResponseError:
         log.debug(u'Track ID match failed.')
         return None
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -305,9 +305,7 @@ def album_info(release):
                     release=release['id'], limit=BROWSE_CHUNKSIZE,
                     includes=BROWSE_INCLUDES,
                     offset=i)['recording-list'])
-        track_map = {}
-        for recording in recording_list:
-            track_map[recording['id']] = recording
+        track_map = {r['id']: r for r in recording_list}
         for medium in release['medium-list']:
             for recording in medium['track-list']:
                 recording_info = track_map[recording['recording']['id']]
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by dosoe** [comment]
As it concerns few releases and 100 tracks apiece is a quick pace, I don't think this feature it is necessary to hide this feature behind a config option.

File: beets/autotag/mb.py [hunk 1]
```diff
@@ -285,6 +287,24 @@ def album_info(release):
     artist_name, artist_sort_name, artist_credit_name = \
         _flatten_artist_credit(release['artist-credit'])
 
+    ntracks = 0
+    for medium in release['medium-list']:
+        ntracks += len(medium['track-list'])
```
sampsyo, line 292:
```py
ntracks = sum(len(m['track-list']) for m in release['medium-list'])

File: beets/autotag/mb.py [hunk 2]
```diff
@@ -285,6 +287,24 @@ def album_info(release):
...
+
+    # for albums with more than 500 tracks
+    if ntracks > 500:
```
sampsyo, line 295: Is there perhaps any more reliable/obvious way to detect when MusicBrainz has omitted recording information of a too-big release? For example, maybe the API will just have all the `recording` fields missing? This would be nice because it would avoid hard-coding a number that might change in the API.

beets/autotag/mb.py
```diff
@@ -285,6 +287,24 @@ def album_info(release):
...
+        recording_list = []
+        for i in range((ntracks // 100) + 1):
```
sampsyo, line 297: Let's make the chunk size (100) into a global constant.

beets/autotag/mb.py
```diff
@@ -285,6 +287,24 @@ def album_info(release):
     artist_name, artist_sort_name, artist_credit_name = \
         _flatten_artist_credit(release['artist-credit'])
 
+    ntracks = 0
+    for medium in release['medium-list']:
+        ntracks += len(medium['track-list'])
+
+    # for albums with more than 500 tracks
```
sampsyo, line 294: A comment here to this effect might be nice:

> The MusicBrainz API omits [include details] when [condition]. So we use `browse_recordings` on chunks of tracks to recover the same information in this case.

beets/autotag/mb.py
sampsyo, line 297: Could also use `range(0, ntracks, RECORDING_CHUNK_SIZE)` to move in "chunks" of 100?

sampsyo (commented):
Cool! Seems neat! Here are some suggestions for how to make this a bit more maintainable.

beets/autotag/mb.py
dosoe: I checked but the only difference seems to be that the ```recording``` dicts contain more or less info, I tried comparing a very big release with a release that consists of one of the discs and the only meaningful differences were that the recordings of the bigger release contain no ```artist-relation-list``` and ```work-relation-list``` info, but that info cannot be used to discriminate, as it could also happen on small releases with incomplete info. 
I tried looking at the logs but there's nothing there either.

beets/autotag/mb.py
dosoe: Done!

beets/autotag/mb.py
dosoe: Done

beets/autotag/mb.py
dosoe: Done

beets/autotag/mb.py
dosoe: Done

beets/autotag/mb.py
sampsyo: Hmm, interesting. We may be stuck with the hard-coded threshold… could be worse, I guess. But at the very least, it would be nice to:

1. Check in with the MusicBrainz folks on IRC or similar to see if they know of any obvious way to tell when this was engaged?
2. Put the number into a module-level constant.

dosoe: I couldn't see anything, I asked on the forum and on IRC and didn't get an answer, so I guess we're stuck with 500.

sampsyo:
Thanks for the updates! I also noticed https://github.com/alastair/python-musicbrainzngs/issues/269. Does this mean that the data when browsing is still incomplete? If so, just checking: is this alternate mode in fact providing more data than the default way for >500 tracks?

One other thing: it would be awesome to add a little bit of logging to indicate what's going on in the verbose log. (Because this will require multiple requests to the MB servers, it would be great to tell the user what's taking so long.)

dosoe:
Yes, the alternate mode provides more data than the current one, it provides work-rels and artist-rels that the default doesn't provide for big releases. I'm getting started on the logging.

dosoe:
> Thanks for the updates! I also noticed [alastair/python-musicbrainzngs#269](https://github.com/alastair/python-musicbrainzngs/issues/269). Does this mean that the data when browsing is still incomplete? If so, just checking: is this alternate mode in fact providing more data than the default way for >500 tracks?

Yes, the data is still incomplete, namely work-level-rels are missing, however they should exist in the musicbrainz API (https://tickets.metabrainz.org/browse/MBS-8521) so the problems seems to be in the implementation of the API to python (i e musicbrainzngs), therefore I made this bug report.

beets/autotag/mb.py
```diff
@@ -285,6 +289,27 @@ def album_info(release):
     artist_name, artist_sort_name, artist_credit_name = \
         _flatten_artist_credit(release['artist-credit'])
 
+    ntracks = sum(len(m['track-list']) for m in release['medium-list'])
+
+    # The MusicBrainz API omits 'artist-relation-list' and 'work-relation-list'
+    # when the release has more than 500 tracks. So we use browse_recordings
+    # on chunks of tracks to recover the same information in this case.
+    if ntracks > BROWSE_MAXTRACKS:
+        log.info(u'Album ' + str(release['id']) + u' has too many tracks')
```
sampsyo, line 298:
```suggestion
        log.debug(u'Album {} has too many tracks', release['id'])
```

sampsyo: These should all be `debug`-level logging statements (so they are silent by default).

beets/autotag/mb.py
```diff
@@ -285,6 +289,27 @@ def album_info(release):
...
+        log.info(u'Fetching recordings in batches of ' + str(BROWSE_CHUNKSIZE))
```
sampsyo, line 299: Let's put this inside the `for` loop below. It can say `debug('fetching track batch starting at {}', i)`

beets/autotag/mb.py
```diff
@@ -285,6 +289,27 @@ def album_info(release):
...
+        recording_list = []
+        for i in range(0, ntracks, BROWSE_CHUNKSIZE):
+            recording_list.extend(musicbrainzngs.browse_recordings(
+                    release=release['id'], limit=BROWSE_CHUNKSIZE,
+                    includes=BROWSE_INCLUDES,
+                    offset=BROWSE_CHUNKSIZE * i)['recording-list'])
```
sampsyo, line 305: Just checking… shouldn't this just be `i`, not `BROWSE_CHUNKSIZE * i`? (Have you tried this out to make sure it's requesting the right tracks?)

beets/autotag/mb.py
```diff
@@ -285,6 +289,27 @@ def album_info(release):
...
+        for medium in release['medium-list']:
+            for recording in medium['track-list']:
+                recording_info = list(filter(lambda track: track['id'] ==
+                                             recording['recording']['id'],
+                                             recording_list))[0]
```
sampsyo, line 310: A list comprehension might be a bit easier to read… `[track for track in recording_list if track['id'] == ...]`.

Or even better would be to first create a dict, `track_map`, mapping from track IDs to tracks, and then just have something like `track_map[recording['recording']['id']]`. That would make sure this remains O(n) rather than re-scanning the list each time, which could actually matter when we have 500+ tracks.

beets/autotag/mb.py
```diff
@@ -517,6 +542,7 @@ def album_for_id(releaseid):
     try:
         res = musicbrainzngs.get_release_by_id(albumid,
                                                RELEASE_INCLUDES)
+        log.info(u'Album ' + str(releaseid) + u' fetched from MusicBrainz')
```
sampsyo, line 545: I don't think we need this log message. (We already have similar logs elsewhere.)

beets/autotag/mb.py
```diff
@@ -536,6 +562,7 @@ def track_for_id(releaseid):
         return
     try:
         res = musicbrainzngs.get_recording_by_id(trackid, TRACK_INCLUDES)
+        log.info(u'Track ' + str(releaseid) + u' fetched from MusicBrainz')
```
sampsyo, line 565: Ditto.

sampsyo (commented):
Cool! Here are some comments/questions.

beets/autotag/mb.py
dosoe: Will do. This is really a failsafe measure, normally we could just take the recordings in the order they are given.

beets/autotag/mb.py
dosoe: That's correct!

beets/autotag/mb.py
```diff
@@ -285,6 +291,28 @@ def album_info(release):
...
+        recording_list = []
+        for i in range(0, ntracks, BROWSE_CHUNKSIZE):
+            log.debug(u'Retrieving tracks starting at {}', i)
+            recording_list.extend(musicbrainzngs.browse_recordings(
+                    release=release['id'], limit=BROWSE_CHUNKSIZE,
+                    includes=BROWSE_INCLUDES,
+                    offset=i)['recording-list'])
+        track_map = {}
+        for recording in recording_list:
+            track_map[recording['id']] = recording
```
sampsyo, line 310: FWIW, this could also be `{r['id']: r for r in recording_list}`.

sampsyo (commented):
Cool! Let me know when you think this is done (i.e., you've tried it out on a couple of albums and everything seems to be working)?

dosoe:
I have done it on a very big album, there is no difference between with this PR and without apart from the additional information (track titles don't change for example) so I think this works.

beets/autotag/mb.py
dosoe: I'm not very familiar with this way to code, but let's go.

dosoe:
Now I get fails in completely unchanged places. Do you have an idea of what's going on?

sampsyo:
Looks like an [overactive linter package](https://github.com/elijahandrews/flake8-blind-except) update! We'll be disabling it soon.

jackwilsdon:
@dosoe this should be resolved as of dc0350d6ffb2e49428ff74a50988fd8a03084968, merging master back into your branch should resolve any issues you're having with the `B902 blind except Exception: statement` lint failure.

dosoe:
It is done, it works.

sampsyo (approved):
Looks great; thank you!

PR #3824, opened 2021-01-05
discogs plugin: fix index_track option

auguste42:
This PR fixes the behavior of the ``index_track`` option of the discogs plugin.

Currently, it only works when index tracks are of type "heading" in the discogs listing (like in https://www.discogs.com/Handel-Sutherland-Kirkby-Kwella-Nelson-Watkinson-Bowman-Rolfe-Johnson-Elliott-Partridge-Thomas-The-A/release/2026070), but not when they are of type "index" (like in https://www.discogs.com/fr/Glenn-Gould-Columbia-Symphony-Orchestra-Vladimir-Golschmann-Bach-The-Concertos-For-KeyboardClavier-S/release/10142332).
This is because index tracks are removed in function ``coalesce_tracks``. The fix is simply to keep them when there is no subtrack merging.

This also removes the semicolon when there is no index track. Currently, when the option is on, every track title will start with " : " when there is no index track.

### DIFF
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -499,7 +499,7 @@ class DiscogsPlugin(BeetsPlugin):
                 else:
                     # Promote the subtracks to real tracks, discarding the
                     # index track, assuming the subtracks are physical tracks.
-                    index_track = tracklist.pop()
+                    index_track = tracklist[-1]
                     # Fix artists when they are specified on the index track.
                     if index_track.get('artists'):
                         for subtrack in subtracks:
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -557,7 +557,8 @@ class DiscogsPlugin(BeetsPlugin):
         title = track['title']
         if self.config['index_tracks']:
             prefix = ', '.join(divisions)
-            title = ': '.join([prefix, title])
+            if prefix != '':
+                title = ': '.join([prefix, title])
         track_id = None
         medium, medium_index, _ = self.get_track_index(track['position'])
         artist, artist_id = MetadataSourcePlugin.get_artist(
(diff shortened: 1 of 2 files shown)


beetsplug/discogs.py
```diff
@@ -557,7 +557,8 @@ def get_track_info(self, track, index, divisions):
         title = track['title']
         if self.config['index_tracks']:
             prefix = ', '.join(divisions)
-            title = ': '.join([prefix, title])
+            if prefix != '':
```
sampsyo, line 560: Could be `if prefix`.

beetsplug/discogs.py
```diff
@@ -557,7 +557,8 @@ def get_track_info(self, track, index, divisions):
...
+                title = ': '.join([prefix, title])
```
sampsyo, line 561: Not your change, obviously, but it occurs to me that a clearer way to write this would be `'{}: {}'.format(prefix, title)`.

sampsyo (commented):
Seems good to me! I'm not a heavy `discogs` plugin user myself, so I'll have to trust you that removing the `pop()` doesn't cause other indexing problems. :smiley: But it seems likely enough to work!

Here are just a couple of tiny comments on the code.

auguste42:
Thanks for the comments! Anyway I just realized that this does not work as expected when there are indexed and non indexed tracks on the same release (e.g. https://www.discogs.com/fr/Bruno-Walter-Maestro-Generoso-/release/9142360). I think the title merging needs to be done inside ``coalesce_tracks``, when we still have the information on the hierarchy of indexes and tracks, I will try to implement that.

auguste42:
Here is the new version, which should be better since it will not modify the indexing at all when ``index_tracks`` is off.

sampsyo (approved):
Awesome; looks good to me! Thanks!

PR #3825, opened 2021-01-06
Add Python version check for mpdstats

### LINKED ISSUE #3798: mpdstats plugin needs updating to work with python-mpd2 2.0.0
(Sorry, this is just from source code inspection and a search of dependent packages in Debian, rather than a crash that I've genuinely reproduced - I don't currently use beets myself, although I keep meaning to try it.)

https://github.com/Mic92/python-mpd2 version 2.0.0 dropped support for the `use_unicode` parameter to the `MPDClient` constructor. This parameter doesn't do anything in Python 3, which always behaves like `use_unicode=True`. It was important in Python 2.

In Debian's python3-mpd package I've added back compatibility with the use_unicode parameter for now (see Mic92/python-mpd2#142) but I don't know whether that will be accepted upstream.

If beets only supports Python 3, dropping the parameter should be a sufficient fix: replace `MPDClient(use_unicode=True)` with `MPDClient()`.

If it still supports Python 2, something similar to https://github.com/multani/sonata/pull/116 should work.

agsimmons:
## Description

Fixes #3798.

Adds a check for Python 2/3 as to pass the correct init arguments to MPDClientWrapper. This removes the warning:
```
test/test_mpdstats.py::MPDStatsTest::test_update_rating
  /home/andrew/Documents/beets/beetsplug/mpdstats.py:55: DeprecationWarning: use_unicode parameter to ``MPDClient`` constructor is deprecated
    self.client = mpd.MPDClient(use_unicode=True)
```

~~Do the dependency constraints for python-mpd2 in setup.py need to be adjusted at all? Could there be a scenario where a user is using some version of Python 3 along with python-mpd2<2.0.0 that could cause issues?~~

## To Do

- [x] ~~Update dependency version constraints? (See question in description)~~

### DIFF
--- a/beetsplug/mpdstats.py
+++ b/beetsplug/mpdstats.py
@@ -18,6 +18,7 @@ from __future__ import division, absolute_import, print_function
 import mpd
 import socket
 import select
+import sys
 import time
 import os
 
@@ -52,7 +53,13 @@ class MPDClientWrapper(object):
         self.music_directory = (
             mpd_config['music_directory'].as_str())
 
-        self.client = mpd.MPDClient(use_unicode=True)
+        if sys.version_info < (3, 0):
+            # On Python 2, use_unicode will enable the utf-8 mode for
+            # python-mpd2
+            self.client = mpd.MPDClient(use_unicode=True)
+        else:
+            # On Python 3, python-mpd2 always uses Unicode
+            self.client = mpd.MPDClient()
 
     def connect(self):
         """Connect to the MPD.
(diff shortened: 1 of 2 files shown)


sampsyo (approved):
This looks great to me; thank you!

> Do the dependency constraints for python-mpd2 in setup.py need to be adjusted at all? Could there be a scenario where a user is using some version of Python 3 along with python-mpd2<2.0.0 that could cause issues?

I'm not quite following what might go wrong here… is there any chance you could elaborate? (Or maybe we should do this in a separate issue?

agsimmons:
> I'm not quite following what might go wrong here… is there any chance you could elaborate?

Sorry, I've read more into the code of python-mpd2 and realized that the issue I was talking about is not actually a thing.

This PR should be good to go, thanks!

sampsyo:
Great; thanks!

PR #3830, opened 2021-01-08
convert: uses new par_map to work in parallel

vvvrrooomm:
## Description

Builds on  #3153 
parallelizes convert using par_map on python3

## To Do

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -26,7 +26,6 @@ import shlex
 import six
 from string import Template
 import platform
-from itertools import repeat
 
 from beets import ui, util, plugins, config
 from beets.plugins import BeetsPlugin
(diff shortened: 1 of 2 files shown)


beetsplug/convert.py
```diff
@@ -183,8 +185,8 @@ def commands(self):
 
     def auto_convert(self, config, task):
         if self.config['auto']:
-            for item in task.imported_items():
-                self.convert_on_import(config.lib, item)
+            items = zip(repeat(config.lib), list(task.imported_items()))
+            par_map(self.convert_on_import, items)
```
wisp3rwind, line 189: Maybe make this `par_map(lambda item: self.convert_on_import(config.lib, item), task.imported_items())` with the old function signature of `convert_on_import`. In my opinion, that would be a lot clearer and should do precisely the same thing, right? I don't think there's any need to make a list out of `task.imported_items()`.

wisp3rwind (changes requested):

beetsplug/convert.py
vvvrrooomm: done

wisp3rwind (changes requested):
Looking good! Would you address the style issues and add a changelog entry? Then we can go ahead and merge.

wisp3rwind (approved):

wisp3rwind:
I went for squashing the PR since the final changeset is so much easier to understand. I hope that's alright with you! Thanks again :tada:

PR #3831, opened 2021-01-09
Add two events ```extracting_trackdata``` and ```extracting_albumdata```

dosoe:
## Description

Only the events as discussed in https://github.com/beetbox/beets/pull/3589

## To Do

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -23,6 +23,7 @@ import traceback
 from six.moves.urllib.parse import urljoin
 
 from beets import logging
+from beets import plugins
 import beets.autotag.hooks
 import beets
 from beets import util
@@ -259,6 +260,12 @@ def track_info(recording, index=None, medium=None, medium_index=None,
     if arranger:
         info.arranger = u', '.join(arranger)
 
+    # supplementary tags provided by plugins
+    extra_trackdatas = plugins.send('extracting_trackdata', info=recording)
+    for extra_trackdata in extra_trackdatas:
+        for key in extra_trackdata:
+            info[key] = extra_trackdata[key]
+
     info.decode()
     return info
 
@@ -446,6 +453,12 @@ def album_info(release):
     genres = release.get('genre-list')
     if config['musicbrainz']['genres'] and genres:
         info.genre = ';'.join(g['name'] for g in genres)
+
+    # supplementary tags provided by plugins
+    extra_albumdatas = plugins.send('extracting_albumdata', info=release)
+    for extra_albumdata in extra_albumdatas:
+        for key in extra_albumdata:
+            info[key] = extra_albumdata[key]
 
     info.decode()
     return info
--- a/docs/dev/plugins.rst
+++ b/docs/dev/plugins.rst
@@ -238,6 +238,11 @@ The events currently available are:
   during a ``beet import`` interactive session. Plugins can use this event for
   :ref:`appending choices to the prompt <append_prompt_choices>` by returning a
   list of ``PromptChoices``. Parameters: ``task`` and ``session``.
+  
+* `extracting_trackdata` and `extracting_albumdata`: called after the metadata
+  is obtained from MusicBrainz. The parameter is a ``dict`` containing the data
+  retrieved from MusicBrainz for a track or an album.
+  Allows to add additional tags.
 
 The included ``mpdupdate`` plugin provides an example use case for event listeners.
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -94,7 +94,7 @@ New features:
   Windows.
   Thanks to :user:`MartyLake`.
   :bug:`3331` :bug:`3334`
-* The 'data_source' field is now also applied as an album-level flexible
+* The ``data_source`` field is now also applied as an album-level flexible
   attribute during imports, allowing for more refined album level searches.
   :bug:`3350` :bug:`1693`
 * :doc:`/plugins/deezer`: Added Deezer plugin as an import metadata provider:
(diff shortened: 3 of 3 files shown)


dosoe:
Would this be fine @sampsyo ?

sampsyo:
Hi, @dosoe—this is slightly off topic, but I just wanted to point out that I'm not the only one who can review pull requests. I know I asked you to do this PR, and it is indeed in my queue, but just in case you're under the impression that I'm the only one to ping, I wanted to encourage you to look a little more broadly… everyone has limited attention to bring to PR reviews, which can be pretty time intensive, but it can be helpful to "shop" it around to get multiple perspectives if you're interested.

beets/autotag/mb.py
```diff
@@ -259,6 +260,12 @@ def track_info(recording, index=None, medium=None, medium_index=None,
     if arranger:
         info.arranger = u', '.join(arranger)
 
+    # supplementary tags provided by plugins
+    extra_trackdatas = plugins.send('extracting_trackdata', info=recording)
+    for extra_trackdata in extra_trackdatas:
+        for key in extra_trackdata:
+            info[key] = extra_trackdata[key]
```
sampsyo, line 267: Would `info.update(extra_trackdata)` work?

docs/dev/plugins.rst
```diff
@@ -238,6 +238,11 @@ The events currently available are:
   during a ``beet import`` interactive session. Plugins can use this event for
   :ref:`appending choices to the prompt <append_prompt_choices>` by returning a
   list of ``PromptChoices``. Parameters: ``task`` and ``session``.
+  
+* `extracting_trackdata` and `extracting_albumdata`: called after the metadata
+  is obtained from MusicBrainz. The parameter is a ``dict`` containing the data
+  retrieved from MusicBrainz for a track or an album.
+  Allows to add additional tags.
```
sampsyo, line 245: The prevailing style on this page seems to be to give every event its own bullet, even if they're similar.

sampsyo: We should also give the name of the parameter.

beets/autotag/mb.py
```diff
@@ -447,6 +454,12 @@ def album_info(release):
     if config['musicbrainz']['genres'] and genres:
         info.genre = ';'.join(g['name'] for g in genres)
 
+    # supplementary tags provided by plugins
+    extra_albumdatas = plugins.send('extracting_albumdata', info=release)
```
sampsyo, line 458: I think `info` might be a somewhat misleading name for the parameter, because it's not a `TrackInfo` or `AlbumInfo` object (as it is for other events). Maybe just `data`?

sampsyo, line 458: For the names of the events, I think it would be good to clarify that this is for MusicBrainz specifically. So perhaps `mb_album_extract` or similar?

beets/autotag/mb.py
```diff
@@ -447,6 +454,12 @@ def album_info(release):
     if config['musicbrainz']['genres'] and genres:
         info.genre = ';'.join(g['name'] for g in genres)
 
+    # supplementary tags provided by plugins
```
sampsyo, line 457: Probably no need to copy & paste the same comment.

beets/autotag/mb.py
```diff
@@ -259,6 +260,12 @@ def track_info(recording, index=None, medium=None, medium_index=None,
     if arranger:
         info.arranger = u', '.join(arranger)
 
+    # supplementary tags provided by plugins
```
sampsyo, line 263:
```suggestion
    # Supplementary fields provided by plugins.
```

sampsyo (changes requested):
Looking good so far; here are some comments!

dosoe:
> Hi, @dosoe—this is slightly off topic, but I just wanted to point out that I'm not the only one who can review pull requests.

Ok sorry. So I could just ping anyone in the core team? Is that a thing now? Or anyone in the beetbox team?

sampsyo:
Sure! And absolutely no worries; just thought I'd expand a bit. Feel free to post on Discourse asking for reviews, for example.

docs/dev/plugins.rst
```diff
@@ -238,6 +238,15 @@ The events currently available are:
...
   :ref:`appending choices to the prompt <append_prompt_choices>` by returning a
   list of ``PromptChoices``. Parameters: ``task`` and ``session``.
+  
+* `mb_track_extract`: called after the metadata is obtained from
+  MusicBrainz. The parameter is a ``dict`` containing the data
+  retrieved from MusicBrainz for a track.
+  Allows to add additional tags.
+  Parameter: ``data``
+
+* `mb_album_extract`: Like `mb_track_extract`, but for album tags.
```
wisp3rwind, line 248: One more nitpick here: I guess it would be helpful if it was spelled out how exactly plugins can add these additional tags. Similar to how the documentation for `before_choose_candidate` above says what type plugins need to return, the return type could be named here (a dict with the `field: value` mapping I suppose?).

wisp3rwind (commented):
I had a brief look, and this looks reasonable to me, and all of @sampsyo's suggestions are addressed. I added one more suggestion on how to improve the docs. Could you have a look at that? It would also be great if you could rebase this on master to ease merging.

Note that the autotagger is probably the part of beets that I know the least (i.e. not at all), so I can't assess whether this is really the right interface for what #3589 tries to achieve.

aereaux:
These changes would be useful for a small plugin I have (which currently relies on a hacky change in the importer, and which would be partially replaced by the rest of the changes from #3589).  I could port my plugin to use these changes and then either write up some more explicit documentation or at least publish it so that there is an example of how to use these hooks to add tags.

docs/dev/plugins.rst
dosoe: Done, does it correspond to what you have in mind?

dosoe: I added it, I hope it's clear and understandable. I also added precisions about priorities (i. e. who prevails when the field is the same).

PR #3832, opened 2021-01-10
add plugin subsonicupdate auth parameter

vincentDcmps:
## Description

this modification will permit that subsonicupdate works with subsonic fork like https://github.com/spl0k/supysonic

### DIFF
--- a/beetsplug/subsonicupdate.py
+++ b/beetsplug/subsonicupdate.py
@@ -20,7 +20,7 @@ a "subsonic" section like the following:
         url: https://mydomain.com:443/subsonic
         user: username
         pass: password
-        auth: enc or plain
+        auth: token or password
 """
 from __future__ import division, absolute_import, print_function
 
--- a/beetsplug/subsonicupdate.py
+++ b/beetsplug/subsonicupdate.py
@@ -88,7 +88,7 @@ class SubsonicUpdate(BeetsPlugin):
                 context_path = ''
             url = "http://{}:{}{}".format(host, port, context_path)
 
-        return url + '/rest/startScan'
+        return url + '/rest/startScan.view'
 
     def start_scan(self):
         user = config['subsonic']['user'].as_str()
--- a/beetsplug/subsonicupdate.py
+++ b/beetsplug/subsonicupdate.py
@@ -20,7 +20,6 @@ a "subsonic" section like the following:
         url: https://mydomain.com:443/subsonic
         user: username
         pass: password
-        auth: token or password
 """
 from __future__ import division, absolute_import, print_function
 
@@ -35,6 +34,7 @@ from beets import config
 from beets.plugins import BeetsPlugin
 
 __author__ = 'https://github.com/maffo999'
+authtokenversion = '1.12'
 
 
 class SubsonicUpdate(BeetsPlugin):
@@ -45,9 +45,16 @@ class SubsonicUpdate(BeetsPlugin):
             'user': 'admin',
             'pass': 'admin',
             'url': 'http://localhost:4040',
-            'auth': 'token',
         })
-
+        self.version = self.get_version()
+        if self.version > authtokenversion:
+            self._log.info(
+                    u'use token authent method')
+            self.auth = "token"
+        else:
+            self.auth = "password"
+            self._log.info(
+                    u'use password authent method')
         config['subsonic']['pass'].redact = True
         self.register_listener('import', self.start_scan)
 
@@ -69,10 +76,10 @@ class SubsonicUpdate(BeetsPlugin):
         return salt, token
 
     @staticmethod
-    def __format_url():
-        """Get the Subsonic URL to trigger a scan. Uses either the url
-        config option or the deprecated host, port, and context_path config
-        options together.
+    def __format_url(endpoint):
+        """Get the Subsonic URL to trigger a endpoint put in paramater.
+        Uses either the url config option or the deprecated host, port,
+        and context_path config options together.
 
         :return: Endpoint for updating Subsonic
         """
@@ -90,18 +97,37 @@ class SubsonicUpdate(BeetsPlugin):
                 context_path = ''
             url = "http://{}:{}{}".format(host, port, context_path)
 
-        return url + '/rest/startScan.view'
+        return url + '/rest/{}'.format(endpoint)
+
+    def get_version(self):
+        url = self.__format_url("ping.view")
+        payload = {
+            'c': 'beets',
+            'f': 'json'
+        }
+        try:
+            response = requests.get(url, params=payload)
+            if response.status_code == 200:
+                json = response.json()
+                version = json['subsonic-response']['version']
+                self._log.info(
+                    u'subsonic version:{0} '.format(version))
+                return version
+            else:
+                self._log.error(u'Error: {0}', json)
+        except Exception as error:
+            self._log.error(u'Error: {0}'.format(error))
 
     def start_scan(self):
         user = config['subsonic']['user'].as_str()
-        url = self.__format_url()
-        if config['subsonic']['auth'] == 'token':
+        url = self.__format_url("startScan.view")
+        if self.auth == 'token':
             salt, token = self.__create_token()
             payload = {
                 'u': user,
                 't': token,
                 's': salt,
-                'v': '1.15.0',  # Subsonic 6.1 and newer.
+                'v': self.version,  # Subsonic 6.1 and newer.
                 'c': 'beets',
                 'f': 'json'
             }
@@ -111,7 +137,7 @@ class SubsonicUpdate(BeetsPlugin):
             payload = {
                 'u': user,
                 'p': 'enc:{}'.format(encpass),
-                'v': '1.15.0',
+                'v': self.version,
                 'c': 'beets',
                 'f': 'json'
             }
--- a/beetsplug/subsonicupdate.py
+++ b/beetsplug/subsonicupdate.py
@@ -20,7 +20,7 @@ a "subsonic" section like the following:
         url: https://mydomain.com:443/subsonic
         user: username
         pass: password
-        auth: enc or token
+        auth: enc or plain
 """
 from __future__ import division, absolute_import, print_function
 
@@ -95,7 +95,7 @@ class SubsonicUpdate(BeetsPlugin):
     def start_scan(self):
         user = config['subsonic']['user'].as_str()
         url = self.__format_url()
-        if config['subsonic']['user'] == 'token':
+        if config['subsonic']['auth'] == 'token':
             salt, token = self.__create_token()
             payload = {
                 'u': user,
@@ -119,6 +119,7 @@ class SubsonicUpdate(BeetsPlugin):
         try:
             response = requests.get(url, params=payload)
             json = response.json()
+
             if response.status_code == 200 and \
                     json['subsonic-response']['status'] == "ok":
                 count = json['subsonic-response']['scanStatus']['count']
--- a/beetsplug/subsonicupdate.py
+++ b/beetsplug/subsonicupdate.py
@@ -46,15 +46,28 @@ class SubsonicUpdate(BeetsPlugin):
             'pass': 'admin',
             'url': 'http://localhost:4040',
         })
-        self.version = self.__get_version()
-        if self.version > AUTH_TOKEN_VERSION:
-            self.auth = "token"
-        else:
-            self.auth = "password"
-        self._log.info(
-            u"using '{}' authentication method".format(self.auth))
         config['subsonic']['pass'].redact = True
+        self._version = None
+        self._auth = None
         self.register_listener('import', self.start_scan)
+
+    @property
+    def version(self):
+        if (self._version is None):
+            self._version = self.__get_version()
+        return self._version
+
+    @property
+    def auth(self):
+        if (self._auth is None):
+            if(self.version is not None):
+                if self.version > AUTH_TOKEN_VERSION:
+                    self._auth = "token"
+                else:
+                    self._auth = "password"
+            self._log.info(
+                u"using '{}' authentication method".format(self._auth))
+        return self._auth
 
     @staticmethod
     def __create_token():
@@ -113,12 +126,15 @@ class SubsonicUpdate(BeetsPlugin):
                 return tuple(int(s) for s in version.split('.'))
             else:
                 self._log.error(u'Error: {0}', json)
+                return None
         except Exception as error:
             self._log.error(u'Error: {0}'.format(error))
+            return None
 
     def start_scan(self):
         user = config['subsonic']['user'].as_str()
         url = self.__format_url("startScan.view")
+
         if self.auth == 'token':
             salt, token = self.__create_token()
             payload = {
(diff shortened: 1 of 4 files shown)


beetsplug/subsonicupdate.py
```diff
@@ -88,26 +90,35 @@ def __format_url():
...
-
-        payload = {
-            'u': user,
-            't': token,
-            's': salt,
-            'v': '1.15.0',  # Subsonic 6.1 and newer.
-            'c': 'beets',
-            'f': 'json'
-        }
+        if config['subsonic']['user'] == 'token':
```
wisp3rwind, line 98: This should be `config['subsonic']['auth']`, right?

beetsplug/subsonicupdate.py
```diff
@@ -88,26 +90,35 @@ def __format_url():
...
+                'p': 'enc:{}'.format(encpass),
+                'v': '1.15.0',
+                'c': 'beets',
+                'f': 'json'
+            }
 
         try:
             response = requests.get(url, params=payload)
             json = response.json()
-
```
wisp3rwind, line 110: Is this whitespace change intentional? Personally, I'd say this block of code is sufficiently long to benefit from the additional linebreak.

docs/plugins/subsonicupdate.rst
```diff
@@ -30,5 +30,6 @@ The available options under the ``subsonic:`` section are:
 
 - **url**: The Subsonic server resource. Default: ``http://localhost:4040``
 - **user**: The Subsonic user. Default: ``admin``
+- **auth**: authentification method token or encoded password
```
wisp3rwind, line 33: I think this needs to be expanded: For me, this sounds as if the value should be the token or password, rather than just an indication of the authentication method. The valid choices (`token` and `enc`) should be indicated, and the description of the `pass` key below should reflect which values it can take in each case.

Since the `enc:` prefix seems to be allowed for tokens, too, maybe it would be better to rename this option to `plain`?

beetsplug/subsonicupdate.py
```diff
@@ -88,26 +90,35 @@ def __format_url():
                 context_path = ''
             url = "http://{}:{}{}".format(host, port, context_path)
 
-        return url + '/rest/startScan'
+        return url + '/rest/startScan.view'
```
wisp3rwind, line 93: Is this valid for all versions of subsonic?

wisp3rwind (changes requested):
I left a few comments on the code. I'm not familiar with subsonic itself, though, so I can't really comment on whether these changes are sensible. @jef you have worked on this plugin most recently, could you have a look?

beetsplug/subsonicupdate.py
vincentDcmps: correction done thanks

beetsplug/subsonicupdate.py
vincentDcmps: linebreak rollback

beetsplug/subsonicupdate.py
vincentDcmps: I have test with  http://demo.subsonic.org/ seem works

docs/plugins/subsonicupdate.rst
vincentDcmps: change enc by plain

docs/plugins/subsonicupdate.rst
```diff
@@ -30,5 +30,6 @@ The available options under the ``subsonic:`` section are:
 
 - **url**: The Subsonic server resource. Default: ``http://localhost:4040``
 - **user**: The Subsonic user. Default: ``admin``
+- **auth**: authentification method token or plain password
```
jef, line 33:
```suggestion
- **auth**: Authentication method. Default: ``token``. Use ``enc`` for APIs <= 1.12.0.
```

jef (changes requested):
I have a few comments here.

First of all, thanks for the contribution! This is great! Secondly, I think we could do something interesting here that could build other deprecated API rather than using `auth = token`.

I think if we call `/rest/ping` and extract the `version` (see http://demo.subsonic.org/rest/ping), we can call the proper API. That way the user doesn't get confused on what to use with `auth`.

This is not necessary by any means, but I think it could be interesting and wanted to bring it up.

I think we should go two routes:

a. review the comments and update some of the verbiage
b. implement the API version check

Feel free to choose any routes and if you have any concerns, let me know ☺️

Again, thank you very much for this!

beetsplug/subsonicupdate.py
```diff
@@ -20,6 +20,7 @@
         url: https://mydomain.com:443/subsonic
         user: username
         pass: password
+        auth: enc or plain
```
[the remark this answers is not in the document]
vincentDcmps: OK so I think in waiting to do the ping method best way is to put simply token or password

beetsplug/subsonicupdate.py
```diff
@@ -38,12 +40,12 @@
 class SubsonicUpdate(BeetsPlugin):
     def __init__(self):
         super(SubsonicUpdate, self).__init__()
-
         # Set default configuration values
         config['subsonic'].add({
             'user': 'admin',
             'pass': 'admin',
             'url': 'http://localhost:4040',
+            'auth': 'token',
```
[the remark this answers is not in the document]
vincentDcmps: to not break current config default must be to token, currently plugins use only token authent

beetsplug/subsonicupdate.py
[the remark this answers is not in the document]
jef: Awesome! I look forward to seeing it. Thanks so much for your time :)

Let me know if you need anything in the meantime.

vincentDcmps:
@jef I have implemented the version check like sugest

beetsplug/subsonicupdate.py
```diff
@@ -90,18 +97,37 @@ def __format_url():
                 context_path = ''
             url = "http://{}:{}{}".format(host, port, context_path)
 
-        return url + '/rest/startScan.view'
+        return url + '/rest/{}'.format(endpoint)
+
+    def get_version(self):
```
jef, line 102:
```suggestion
    def __get_version(self):
```

I believe we want to make this private.

beetsplug/subsonicupdate.py
```diff
@@ -90,18 +97,37 @@ def __format_url():
...
+        url = self.__format_url("ping.view")
```
jef, line 103:
```suggestion
        url = self.__format_url("ping")
```

Docs have `.view`, but I don't believe we need `.view`. This is also for the `starScan.view`.

jef (commented):
Awesome! Things look great to me. Just one last thing and I think we're good to go!

jef:
@vincentDcmps thanks for all the hard work on this, this looks great! This will also help us in the future with other API calls if we plan to have any!

beetsplug/subsonicupdate.py
vincentDcmps: like I say before I have put that way because .view work with classic subsonic but not with supysonic
 I will contact maintainer to know if we can add endpoint without .view. so for me it's better to let .view for now no?

beetsplug/subsonicupdate.py
jef: Ahh, it's ok! Luckily it works in both. I didn't realize this was also another deprecated API thing.

Thank you!

beetsplug/subsonicupdate.py
jef: Sorry, I didn't see your other comment.

No that's fine. Since `.view` is compatible with both versions, I think this is fine.

Thank you very much ☺️

jef (approved):
Looks good to me! Thank for for all your hard work and working with me on this!

jef:
@wisp3rwind, ready to go. Can you give it a once over?

test/test_subsonicupdate.py
```diff
@@ -39,9 +39,21 @@ def setUp(self):
...
+            'http://localhost:4040/rest/ping.view',
+            status=200,
+            body=self.PING_BODY
+        )
         self.subsonicupdate = subsonicupdate.SubsonicUpdate()
-
+    PING_BODY = '''
+{
+    "subsonic-response": {
+        "status": "failled",
```
jef, line 52:
```suggestion
        "status": "failed",
```

jef (approved):

beetsplug/subsonicupdate.py
```diff
@@ -29,23 +29,32 @@
 
 import requests
 
+from binascii import hexlify
 from beets import config
 from beets.plugins import BeetsPlugin
 
 __author__ = 'https://github.com/maffo999'
+authtokenversion = '1.12'
```
sampsyo, line 37: FWIW, it's generally nice for module-level constants to get `CAPITAL_NAMES`, like `AUTH_TOKEN_VERSION`.

sampsyo (commented):
Looking great! One tiny style suggestion within.

It also seems a little odd to compare version strings lexically, as opposed to doing `tuple(version.split('.')) > (1, 12)` to do numerical comparisons, but I can't actually concoct a reasonable example where this does the wrong thing, so maybe it's fine.

beetsplug/subsonicupdate.py
```diff
@@ -67,10 +76,10 @@ def __create_token():
         return salt, token
 
     @staticmethod
-    def __format_url():
-        """Get the Subsonic URL to trigger a scan. Uses either the url
-        config option or the deprecated host, port, and context_path config
-        options together.
+    def __format_url(endpoint):
+        """Get the Subsonic URL to trigger a endpoint put in paramater.
```
wisp3rwind, line 80: Maybe this wording is clearer?
```suggestion
        """Get the Subsonic URL to trigger the given endpoint.
```

wisp3rwind (approved):
Looking great! Thanks @vincentDcmps and @jef for the great work, the version check is a lot nicer than yet another config option!

I added some more nitpicks in the inline comments in addition to what @sampsyo said. None of them are really blocking the merge here. Please indicate if you'd rather be done with this PR, then I'll push those changes myself.

beetsplug/subsonicupdate.py
```diff
@@ -29,23 +29,32 @@
...
-
+        self.version = self.__get_version()
+        if self.version > authtokenversion:
+            self._log.info(
+                    u'use token authent method')
+            self.auth = "token"
+        else:
+            self.auth = "password"
+            self._log.info(
+                    u'use password authent method')
```
wisp3rwind, line 57: Minor rewording and style suggestion:
```suggestion
        if self.version > authtokenversion:
            self.auth = "token"
        else:
            self.auth = "password"
        self._log.info(u"using '{}' authentication method"
                .format(self.auth))
```

Is `> 1.12` the right version check, or should it rather be `>= 1.13`?

jef: Should be `<= 1.12`, so `> 1.12` is fine!

wisp3rwind: I'm asking because the API docs state that token authentication is new in `1.13.0`. There doesn't seem to be a version `1.12.1`, so this should indeed work. I don't know subsonic's versioning and release scheme, so no idea whether there's a chance that a `1.12.1` API version will ever be released (probably not, since `1.16.1` is current).

vincentDcmps:
Hi everyone I just push your last suggestion. 
just the string comparison let as before because I find ```tuple(int(s) for s in version.split('.')) > (1, 12)``` less Human readable but it's a point of view ;)

wisp3rwind:
> Hi everyone I just push your last suggestion.
> just the string comparison let as before because I find `tuple(int(s) for s in version.split('.')) > (1, 12)` less Human readable but it's a point of view ;)

Actually, looking at this again, lexical comparison of strings is incorrect:
```
>>> "1.13" > "1.12"
True
>>> "1.12" > "1.12"
False
>>> "1.11" > "1.12"
False
>>> "1.6" > "1.12"
True
```
note the last example.

sampsyo:
Good find! Not sure why I couldn't think of that when I was trying to cook up examples before. :smiley:

vincentDcmps:
OK I admit defeat ;) change done

wisp3rwind:
> OK I admit defeat ;) change done

Thanks! The code is really shaping up now :tada:

I did spot another issue with the version check, though:
- If an error occurs in `__get_version`, it returns nothing (i.e. it implicitly returns `None`). That's fine I would say, since `None` transports the notion of "can't determine the version" sufficiently well. It might be good for readability to make the `return None` explicit. However, the next thing that happens is the comparison
```
>>> None > (1, 12)
TypeError: '>' not supported between instances of 'NoneType' and 'tuple'
```
that is, beets crashes with an uncaught exception. It would be much better to exit without a traceback (`__get_version` already prints a diagnostic message, after all), or maybe even ignore the error, since apart from the subsonic update, an import session could still succeed. This is not an edge case that we could ignore, since this can fail for reasons such as the server being down.

- The version check is done too eagerly, namely during plugin initialization, which always happens on beets startup. However, the plugin does nothing unless `beet import` is being run. So the version check might be unnecessary, or worse could crash beets even if running some other command. The latter case might be remediated by fixing the first bullet point above. Still, we should not be contacting the server needlessly. So the version should only be queried when authentication needs to be done for the first time, which is in `start_scan`.
It might be easiest to just move the corresponding code from `__init__` to that function. Another option would be to turn `version` and `auth` into [properties](https://docs.python.org/3.9/library/functions.html#property) for which the getter functions contact the server when they're called for the first time. Right now, I don't see much of a difference between both methods. The solution using properties might be useful should the plugin ever gain additional functionality.

Let us know if you need some guidance for these changes!

wisp3rwind (changes requested):
Loading the plugin is safe now :+1:

I do think that the point that I raised about the version check being too eager should be addressed, too. One way to do so would be to move the version check to `start_scan`, and in that function simply exit early if the version check fails. That should result in the same overall behaviour as your current code (i.e. just log an error if unable to determine the version, but not abort the import session).

The issue is that if you run e.g. `beet ls <some query>` the plugin will still be loaded and attempt to connect to the subsonic server to determine the version. I really think that `beet ls` should not cause any network traffic, though. In addition, I guess this can delay such commands quite a bit if the server is slow to respond.

jef (approved):
Great use of `@property` in my opinion! I think this looks pretty good to me. What do you think @wisp3rwind?

Thank for all the hard work here @vincentDcmps, I appreciate you sticking around and seeing this through.

wisp3rwind (approved):

wisp3rwind:
Looks good to me :tada: If you want, you could add your credits to the [changelog](https://github.com/beetbox/beets/blob/master/docs/changelog.rst). Otherwise, I'm going to merge this soon.

vincentDcmps:
changelog updated, thanks for your support.
if you need beginners help you are welcome ;)

wisp3rwind:
> changelog updated, thanks for your support.
> if you need beginners help you are welcome ;)

By all means, please stay around and go on improving beets :)

jef:
> if you need beginners help you are welcome ;)

Better than me, I barely use Python 😆 Thanks for all the help!

PR #3836, opened 2021-01-18
fix bug 3804

### LINKED ISSUE #3804: missing plugin format for albums
Hi,

the missing plugin doesn't use the format provided when listing missing albums.

Here's an example:

```sh
$ beet -vv missing -f '$mb_releasegroupid | $albumtype | $albumartist | $album' -a Accept
user configuration: /config/config.yaml
data directory: /config
plugin paths: 
Sending event: pluginload
library database: /config/musiclibrary.blb
library directory: /music
Sending event: library_opened
Accept - Restless the Best
Accept - Breaker
Accept - The Best of Accept
Accept - I'm a Rebel / Breaker
Accept - Too Mean to Die
Accept - No Substitutes
Accept - Best of Accept
Accept - Restless & Wild / Balls to the Wall
Accept - The Collection
Accept - Metal Masters
Accept - Hungry Years
Sending event: cli_exit
```

### Setup

* OS: Centos 7 + docker + linuxserver/beets
* Python version: 3.8.2
* beets version: 1.4.9
* Turning off plugins made problem go away (yes/no): didn't try

```yaml
plugins: convert discogs embedart fetchart lastgenre missing replaygain scrub web
directory: /music
art_filename: albumart
threaded: yes
original_date: no
per_disc_numbering: no
format_album: $mb_releasegroupid § $albumtype § $albumartist § $album
format_item: $mb_releasegroupid § $albumtype § $albumartist § $album § $title

convert:
  auto: yes
  dest: /music
  ffmpeg: /usr/bin/ffmpeg
  format: mp3
  #max_bitrate: 640
  threads: 1

embedart:
  auto: no

import:
  write: yes
  copy: yes
  move: no
  resume: ask
  incremental: yes
  quiet_fallback: skip
  timid: no
  log: /config/beet.log

fetchart:
  auto: yes

lastgenre:
  auto: yes
  source: album

missing:
  format: $mb_releasegroupid § $albumtype § $albumartist § $album § $title
  count: no
  total: no

paths:
  default: $albumartist/$album%aunique{} [$year]/$track - $title
  singleton: Singleton/$artist - $title
  comp: Compilations/$album%aunique{} [$year]/$track - $title
  albumtype_soundtrack: Soundtracks/$album [$year]/$track - $title 

replaygain:
  auto: no

scrub:
  auto: yes

web:
  host: 0.0.0.0
  port: 8337
```

thanks..

syco:
Signed-off-by: Syco <alberto.rinaudo+zhds@gmail.com>

## Description

Fixes #3804

### DIFF
--- a/beetsplug/missing.py
+++ b/beetsplug/missing.py
@@ -254,7 +254,9 @@ class MissingPlugin(BeetsPlugin):
 
             for release_group in missing:
                 try:
-                    releases_object = musicbrainzngs.browse_releases(release_group=release_group['id'])
+                    releases_object = musicbrainzngs.browse_releases(
+                            release_group=release_group['id']
+                        )
                     albums_list = releases_object['release-list']
                 except MusicBrainzError as err:
                     self._log.info(
--- a/beetsplug/missing.py
+++ b/beetsplug/missing.py
@@ -229,8 +229,8 @@ class MissingPlugin(BeetsPlugin):
                 continue
 
             try:
-                resp = musicbrainzngs.browse_releases(artist=artist[1])
-                release_groups = resp['release-list']
+                resp = musicbrainzngs.browse_release_groups(artist=artist[1])
+                release_groups = resp['release-group-list']
             except MusicBrainzError as err:
                 self._log.info(
                     u"Couldn't fetch info for artist '{}' ({}) - '{}'",
@@ -252,10 +252,21 @@ class MissingPlugin(BeetsPlugin):
             if total:
                 continue
 
-            for album in missing:
-                for album_info in hooks.albums_for_id(album['id']):
-                    item = _album_item(album_info, album_info.album_id)
-                    print_(format(item, fmt))
+            for release_group in missing:
+                try:
+                    releases_object = musicbrainzngs.browse_releases(release_group=release_group['id'])
+                    albums_list = releases_object['release-list']
+                except MusicBrainzError as err:
+                    self._log.info(
+                        u"Couldn't fetch info for release group '{}' - '{}'",
+                        release_group['id'], err
+                    )
+                    continue
+
+                for album in albums_list:
+                    for album_info in hooks.albums_for_id(album['id']):
+                        item = _album_item(album_info, album_info.album_id)
+                        print_(format(item, fmt))
 
         if total:
             print(total_missing)
--- a/beetsplug/missing.py
+++ b/beetsplug/missing.py
@@ -84,6 +84,53 @@ def _item(track_info, album_info, album_id):
     })
 
 
+def _album_item(album_info, album_id):
+    """Build and return `item` from `album info`
+    objects. `item` is missing what fields cannot be obtained from
+    MusicBrainz alone (encoder, rg_track_gain, rg_track_peak,
+    rg_album_gain, rg_album_peak, original_year, original_month,
+    original_day, length, bitrate, format, samplerate, bitdepth,
+    channels, mtime.)
+    """
+    a = album_info
+
+    return Item(**{
+        'album_id':           album_id,
+        'album':              a.album,
+        'albumartist':        a.artist,
+        'albumartist_credit': a.artist_credit,
+        'albumartist_sort':   a.artist_sort,
+        'albumdisambig':      a.albumdisambig,
+        'albumstatus':        a.albumstatus,
+        'albumtype':          a.albumtype,
+        'artist':             a.artist,
+        'artist_credit':      a.artist_credit,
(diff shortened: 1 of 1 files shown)


beetsplug/missing.py
```diff
@@ -181,8 +229,8 @@ def _missing_albums(self, lib, query):
                 continue
 
             try:
-                resp = musicbrainzngs.browse_release_groups(artist=artist[1])
-                release_groups = resp['release-group-list']
+                resp = musicbrainzngs.browse_releases(artist=artist[1])
+                release_groups = resp['release-list']
```
sampsyo, line 233: Is this related to the formatting change?

beetsplug/missing.py
```diff
@@ -204,10 +252,10 @@ def _missing_albums(self, lib, query):
             if total:
                 continue
 
-            missing_titles = {rg['title'] for rg in missing}
-
-            for release_title in missing_titles:
-                print_(u"{} - {}".format(artist[0], release_title))
+            for album in missing:
+                for album_info in hooks.albums_for_id(album['id']):
+                    item = _album_item(album_info, album_info.album_id)
```
sampsyo, line 257: Sorry for being slow, but I don't quite understand what the deal is with the new `_album_item` filter. Can you explain a little more how this works?

sampsyo (commented):
Thanks!! A couple of questions within. Also, could you please add a changelog entry? (There are notes on how to do so in the PR template.)

beetsplug/missing.py
wisp3rwind, line 233: I'm wondering whether the changes from this PR should be hidden behind a command line flag (`--show-releases`, `--detailed-releases`?). Going from release groups to individual releases will substantially increase the amount of output (I'd imagine you'd easily see 5-10 times as many output lines for many artists), making it harder to parse visually. I'm not a user of this plugin myself, but I think I'd mostly be interested in knowing the release groups that I'm missing rather than each and every Vinyl, CD and digital release.

What do you think?

Style-wise, I think `release_groups` (and `rg`) below should be renamed to `releases` and `r` to avoid confusion.

beetsplug/missing.py
```diff
@@ -204,10 +252,10 @@ def _missing_albums(self, lib, query):
             if total:
                 continue
```
wisp3rwind, line 253: I'm pretty sure this filter is broken now, since it's comparing release group and release IDs. In my opinion, the right thing would to indeed filter by release group. But again, someone who actually uses the plugin should comment.

wisp3rwind (changes requested):

beetsplug/missing.py
syco: It is, but it's in the wrong place, I'll explain after the next commit.

beetsplug/missing.py
syco: To use the `format` function I needed an object previously returned by `_item`, but `_item` expects a `track_info` object.
Because this is applied to albums I don't have a `track_info` object to pass to `_item`, hence I created a new filter called `_album_item` that accept an `album_info`.

beetsplug/missing.py
syco: I've restored this lines, now it's comparing release group with release group as it should be

beetsplug/missing.py
```diff
@@ -252,10 +252,23 @@ def _missing_albums(self, lib, query):
...
+                    self._log.info(
+                        u"Couldn't fetch info for release group '{}' - '{}'",
+                        release_group['id'], err
+                    )
+                    continue
+
+                for album in albums_list:
+                    for album_info in hooks.albums_for_id(album['id']):
+                        item = _album_item(album_info, album_info.album_id)
+                        print_(format(item, fmt))
```
syco, line 271: So, to `print_` `format`, I need an item created by the filter at the top of the page, but the original object in the original code was the output of `musicbrainzngs.browse_release_groups`, these objects didn't contain any of the required attributes.
So for every missing release group I get the list of missing albums per release group.
This way I now have all the variables and can use the `format` function using all the placeholder already defined in the documentation.

beetsplug/missing.py
syco: Sorry for the confusion with the first commit, I should have tested more before opening the PR.
This change does increase the output, but the original `rg` object contains only release group id, artist, and album name.
If we want allow a formatting string that uses any other placeholder it's necessary to get the full album object.

syco:
This generates too many requests to musicbrainz and is far too slow to be usable.
Sorry I wasted your time, I'll be back in a few days when I have a better idea on how to proceed.
Thanks anyway.

Issue #2078, opened 2016-06-23
test_constrained_exception hangs the test suite in py3


jrobeson:
See https://github.com/beetbox/beets/blob/master/test/test_pipeline.py#L160

I have to `Ctrl+C` to stop it. 

When I `Ctrl+C` I get this:

``` sh
$ python3-tox -e py35-test -- test/test_pipeline.py -v
GLOB sdist-make: /home/johnny/src/beets/setup.py
py35-test inst-nodeps: /home/johnny/src/beets/.tox/dist/beets-1.3.19.zip
py35-test installed: beautifulsoup4==4.4.1,beets==1.3.19,click==6.6,cookies==2.2.1,coverage==4.1,enum34==1.1.6,Flask==0.11.1,itsdangerous==0.24,jellyfish==0.5.5,Jinja2==2.8,MarkupSafe==0.23,mock==2.0.0,munkres==1.0.7,musicbrainzngs==0.6,mutagen==1.32,nose==1.3.7,nose-show-skipped==0.1,pathlib==1.0.1,pbr==1.10.0,pylast==1.6.0,python-mpd2==0.5.5,pyxdg==0.25,PyYAML==3.11,rarfile==2.8,requests==2.10.0,responses==0.5.1,six==1.10.0,Unidecode==0.4.19,Werkzeug==0.11.10
py35-test runtests: PYTHONHASHSEED='418410290'
py35-test runtests: commands[0] | nosetests test/test_pipeline.py -v
test_pull (test.test_pipeline.BubbleTest) ... ok
test_run_parallel (test.test_pipeline.BubbleTest) ... ok
test_run_sequential (test.test_pipeline.BubbleTest) ... ok
test_constrained (test.test_pipeline.ConstrainedThreadedPipelineTest) ... ok

test_constrained_exception (test.test_pipeline.ConstrainedThreadedPipelineTest) ... ^CERROR: KEYBOARDINTERRUPT

----------------------------------------------------------------------
Ran 5 tests in 4.996s

OK
^CException ignored in: <module 'threading' from '/usr/lib64/python3.5/threading.py'>
Traceback (most recent call last):
  File "/usr/lib64/python3.5/threading.py", line 1288, in _shutdown
ERROR: keyboardinterrupt
    t.join()
  File "/usr/lib64/python3.5/threading.py", line 1054, in join
    self._wait_for_tstate_lock()
  File "/usr/lib64/python3.5/threading.py", line 1070, in _wait_for_tstate_lock
[...]

PR #3842, opened 2021-01-26
pipeline: Implement our own queue which needs no hacks for invalidation

wisp3rwind:
Commit message:
```
The old code relied on implementation details of queue.Queue from
Python's stdlib, which has caused issues before (namely, breaking beets
on Python 3.2 because of a minor change in the implementation).

This commit imports the implementation of the stdlib Queue. The code is
heavily stripped down since we don't use many of its methods. Instead,
invalidation and the acquire/release methods are now methods of the
queue itself, and it is fine to rely on Queue internals.
```

## Description
I've had this lying around for quite some time already, and finally cleaned it up a bit. I think this is how we should go about

> [...] and I'm also impressed at how miserable the problem is. In retrospect, the pipeline invalidation thing was definitely a hack -- we need a way to empty out a queue and instruct threads to shut down, but surely we can find a less invasive way to accomplish that. I'll add this to my to-do list.

_Originally posted by @sampsyo in https://github.com/beetbox/beets/issues/2078#issuecomment-229141698_

I tried to come up with some code that subclassed the stdlib `queue.Queue` and handles invalidation and the acquire/release pattern using the documented interface of `queue.Queue` only. However, the code turned into quite a mess, since:

- there's no way to wake up all threads that are blocked on the queue except for pushing dummy items
- I had to either use the queue's mutex (implementation detail) or add our own lock to the subclass
 
I didn't get that quite right yet, and the code had already grown to be pretty unpleasant to reason about. Thus, I'm convinced that the way to go is really just ditching the stdlib and implementing the queue ourselves, hence this PR.

I've mentioned the idea to do this before at https://github.com/beetbox/beets/pull/3075#issuecomment-436034945 where a little discussion can be found.

## To Do

- [x] Documentation. (not user-facing, has some code comments)
- [x] Changelog. (not user-facing)
- [ ] Tests. (The pipeline already has a number of tests.)

### DIFF
--- a/beets/util/queue.py
+++ b/beets/util/queue.py
@@ -1,0 +1,176 @@
+# -*- coding: utf-8 -*-
+# This file is part of beets.
+# Copyright 2021, @wisp3rwind
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+
+# The queue implementation is based on queue.Queue from Python 3.7, which is
+# licensed under the PSF (), and
+#
+# Copyright © 2001-2018 Python Software Foundation; All Rights Reserved
+#
+# All functionality that is not required has been stripped. Facilities for
+# counting threads that feed into a queue and for invalidating the queue have
+# been added.
+
+"""Custom queue for communication between pipeline stages."""
+
+from __future__ import division, absolute_import, print_function
+
+from collections import deque
+import threading
+
+
+class CountedInvalidatingQueue():
+    """A simple thread-safe queue derived from Python's `queue.Queue`.
+
+    Implements additional methods for
+
+    - tracking when all producers are done putting items into the queue,
+    - invalidating the queue, i.e. immediatly unblocking all threads that are
+      waiting to `put()` or `get()`.
+    """
+
+    def __init__(self, maxsize=0, on_finished=None):
+        """Create a queue object with a given maximum size.
+
+        If `maxsize` is <= 0, the queue size is infinite.
+
+        After all threads have released the queue it will be invalidated with
+        the value of `on_finished`.
+        """
+        self._maxsize = maxsize
+        self._on_finished = on_finished
+
+        self._queue = deque()
+        self._invalidated = False
+        self._finished = False
+        self._nthreads = 0
+
+        # _mutex must be held whenever the queue is mutating.  All methods
+        # that acquire _mutex must release it before returning.  _mutex
+        # is shared between the three conditions, so acquiring and
+        # releasing the conditions also acquires and releases _mutex.
+        self._mutex = threading.Lock()
+
+        # Notify not_empty whenever an item is added to the queue; a
+        # thread waiting to get is notified then.
+        self.not_empty = threading.Condition(self._mutex)
+
+        # Notify not_full whenever an item is removed from the queue;
+        # a thread waiting to put is notified then.
+        self.not_full = threading.Condition(self._mutex)
+
+    def acquire(self):
+        """Indicate that a thread will start putting into this queue.
+
+        Must not be called after the queue is already finished.
+        """
+        with self._mutex:
+            assert not self._finished
+            self._nthreads += 1
+
+    def release(self):
+        """Indicate that a thread that was putting into this queue has exited.
+
+        If this is the last thread using the queue, the queue is finished.
+        """
+        with self._mutex:
+            assert self._nthreads > 0
+            self._nthreads -= 1
+            if self._nthreads == 0:
+                # All threads are done adding to this queue. Poison it
+                # when it becomes empty.
+                self._finished = True
+                self.not_empty.notify_all()
+
+    def _invalidate(self, value):
+        """Internal method implementing invalidation, see `invalidate()`.
+
+        The caller must hold `self._mutex`.
+        """
+        self._invalid_value = value
+        if not self._invalidated:
+            self._invalidated = True
+            self.not_full.notifyAll()
+            self.not_empty.notifyAll()
+
+    def invalidate(self, value):
+        """Invalidate the queue with the given `value`.
+
+        When invalidated, the Queue will accept more items beyond its maxsize
+        and only ever return the `value`. Thus, it will never block, and wake
+        up all threads that are blocked in `put()` or `get()`.
+        While invalidated, the queue will not 'lose' any items such that it is
+        possible to `resume()` operation at a later point.
+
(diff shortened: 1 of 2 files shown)


sampsyo (approved):
Yay!!! :tada: This looks amazing—thanks for doing the annoying work of reimplementing `Queue`, which does look like the right thing to do. I've done a fairly close review of the methods on the new queue class and everything looks right to me, including the original standard queue methods. Awesome!

sampsyo:
> By the way — are you fine with deriving from that code and with the way I added the copyright notice? I'm not a lawyer, but I think this should be fine under the [PSF](https://docs.python.org/3/license.html).

Right, that's a good question—I saw that and (while also not anything even close to a lawyer) it seems fine to me? I'm glad we preserved their copyright in some form, at least.

> I'm going to push one more commit where I flattened the conditionals in `get()` and `put()`. Would you mind having a brief look whether I really didn't change them in the process?

Sure thing; will review now…

beets/util/queue.py
```diff
@@ -164,13 +170,15 @@ def get(self):
...
                     return self._invalid_value
-                elif not len(self._queue):
-                    if self._finished:
-                        self._invalidate(self._on_finished)
-                        return self._on_finished
-                    else:
-                        self.not_empty.wait()
-                else:
+
+                if len(self._queue):
```
sampsyo, line 174: Not sure about this, but could this be `if self._queue` (the same way lists work)?

beets/util/queue.py
```diff
@@ -164,13 +170,15 @@ def get(self):
...
                     item = self._queue.popleft()
                     self.not_full.notify()
                     return item
+
+                # When this code is reached, the queue is currently empty.
+                if self._finished:
+                    self._invalidate(self._on_finished)
+                    return self._on_finished
+
+                self.not_empty.wait()
```
sampsyo, line 184: This reorganization looks good to me.

beets/util/queue.py
```diff
@@ -146,10 +153,9 @@ def put(self, item):
         return immediately.
         """
         with self.not_full:
-            if not self._invalidated and self._maxsize > 0:
-                while (not self._invalidated and
-                       len(self._queue) >= self._maxsize):
-                    self.not_full.wait()
+            while (not self._invalidated and self._is_full()):
```
sampsyo, line 156: Parentheses are unnecessary here (since it fits on one line).

beets/util/queue.py
```diff
@@ -138,6 +138,13 @@ def resume(self):
             else:
                 assert len(self._queue) == 0
 
+    def _is_full(self):
+        """Predicate factored out to increase readability of `put()`.
+
+        For internal use only, `self._mutex` must be held.
+        """
+        return self._maxsize == 0 or len(self._queue) >= self._maxsize
```
sampsyo, line 146: If this (as the name implies) is meant to return True when the queue is full and can't currently accept any new elements, it seems intuitively that the `self._maxsize == 0 or` bit doesn't belong. That is, when `_maxsize == 0`, the queue is supposed to be infinite—i.e., *never* full, so this method should always return False. So I think it should instead be `self._maxsize > 0 and`?

beets/util/queue.py
sampsyo, line 156: This collapse of the `if` and `while` looks right to me with the above change to `_is_full`. But I think it's incorrect without that change: the intended behavior is to skip waiting altogether if `_maxsize == 0`, and I don't think we'd do that with the current version.

sampsyo (commented):
Just one correctness concern in the new `_is_full` method! I hope I'm thinking about this correctly.

beets/util/queue.py
wisp3rwind: Yep, that is certainly more pythonic!

beets/util/queue.py
wisp3rwind: Right, thanks for spotting this!

beets/util/queue.py
wisp3rwind: True, I got that wrong, `self._maxsize > 0 and [...]` is certainly correct.

beets/util/queue.py
wisp3rwind: Yes, `_is_full` was supposed to do what its name suggests.

wisp3rwind:
Thanks for the review!

The licensing probably needs some more thought, so I converted the PR to a draft.

The relevant sections in the PSF are

> 2. Subject to the terms and conditions of this License Agreement, PSF hereby
>    grants Licensee a nonexclusive, royalty-free, world-wide license to reproduce,
>    analyze, test, perform and/or display publicly, prepare derivative works,
>    distribute, and otherwise use Python 3.7.9 alone or in any derivative
>    version, provided, however, that PSF's License Agreement and PSF's notice of
>    copyright, i.e., "Copyright © 2001-2020 Python Software Foundation; All Rights
>    Reserved" are retained in Python 3.7.9 alone or in any derivative version
>    prepared by Licensee.

I did add the notice of copyright, so that requirement is fulfilled.

The difficult bit might be "retaining the license agreement". Previous licenses in Python's [license "stack"](https://wiki.python.org/moin/PythonSoftwareFoundationLicenseFaq#If_I_bundle_Python_with_my_application.2C_what_do_I_need_to_include_in_my_software_and.2For_printed_documentation.3F) also include this statement. 
This raises two related issues:
- Do we need to provide the full license stack with `queue.py`, too? I think so, because unlike the MIT, the PSF license does not limit this requirement to _substantial portions_ of Python, but really any derivative work. It might be worth checking when the queue module was added to the stdlib. Maybe not all licenses in the history of Python are relevant.
- Currently, beets is all-MIT. We might need to update at least the LICENSE file to state that the queue file is subject to a different license. Due to the previous point we cannot just redistribute it under MIT terms, since that doesn't contain the requirement of shipping the PSF license agreement.

> 3. In the event Licensee prepares a derivative work that is based on or
>    incorporates Python 3.7.9 or any part thereof, and wants to make the
>    derivative work available to others as provided herein, then Licensee hereby
>    agrees to include in any such work a brief summary of the changes made to Python
>    3.7.9.

`queue.py` has a brief paragraph stating that this is stripped down from  Python's stdlib queue, so this is fulfilled, too.

Combining MIT and PSF code by itself should not be a problem, since both allow you to do essentially all you want with the code, assuming proper attribution and copyright statements.
So the question is whether complicating our licensing is something we want to do. I don't see any fundamental problems with shipping some files with non-MIT licenses right now.

Relevant stackexchange answer: https://opensource.stackexchange.com/a/4336

sampsyo:
Ah, super confusing! First of all, yes, it seems like a good point that we need to mention in our LICENSE that we have a portion under this other license. It would be annoying to have to include the entire Python license *inside the queue.py file*, but that may be inevitable?

wisp3rwind:
Still going to wrap this up at some point, not stale.

arogl:
How does the move to Python 3 affect this request?

wisp3rwind:
> How does the move to Python 3 affect this request?

Not at all, pipeline invalidation is still a big hack. The tricky bit here is really to properly handle license restrictions; I'm rather confident about the code itself. I've also put this on the back burner since I've only little time for beets at the moment, and I've got other open PRs I'd like to address first.

PR #3845, opened 2021-01-31
Added Dockerfile

catap:
## Description

This commit introduced a `Dockerfile` that allows to build beet as simple docker container with almost all plugins :)

I feel that the next step should be adding github actions to automatically publish releases ;)

You can test this via `docker catap/beet`

Fixes #X.  

## To Do

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,3 +89,6 @@ ENV/
 /.pydevproject
 /.settings
 .vscode
+
+# GIT stuff, needs for docker
+/.git
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,45 @@
+FROM python:3.8
+
+# Copy current sources
+COPY . /src
+
+# install bets
+RUN pip install /src
+
+# install all possibly dependencies for plugins
+RUN pip install pigar
+RUN pigar -p /src/plugins-requirements.txt -P /src
+RUN pip install -r /src/plugins-requirements.txt
+
+# Install some 3rd party plugins
+RUN pip install beets-artistcountry
+RUN pip install beets-copyartifacts3
+RUN pip install git+git://github.com/geigerzaehler/beets-check.git@master
+RUN pip install beets-alternatives
+RUN pip install beets-follow
+RUN pip install beets-noimport
+RUN pip install pip install git+git://github.com/igordertigor/beets-usertag.git@master
+RUN pip install pip install git+git://github.com/8h2a/beets-barcode.git@master
+RUN pip install beets-ydl
+# some well known dependencies
+RUN pip install pylast
+RUN pip install flask
+RUN pip install flask
+RUN pip install beautifulsoup4
+
+# remove sources
+RUN rm -rf /src
+
+# run everything under defined user
+ARG APP_USER=beet
+RUN groupadd -r ${APP_USER} && useradd --no-log-init -r -g ${APP_USER} -d /${APP_USER} ${APP_USER}
+RUN mkdir -p /${APP_USER}
+RUN chown ${APP_USER}:${APP_USER} /${APP_USER}
+
+USER ${APP_USER}:${APP_USER}
+
+# move /beet to volume and ask beet to use /beet as work directory
+VOLUME /beet
+ENV BEETSDIR=/beet
+
+ENTRYPOINT ["beet"]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+.gitignore
--- a/docs/guides/main.rst
+++ b/docs/guides/main.rst
@@ -128,6 +128,31 @@ trouble or you have more detail to contribute here, please direct it to
 .. _get-pip.py: https://bootstrap.pypa.io/get-pip.py
 
 
+Installing on Docker
+^^^^^^^^^^^^^^^^^^^^
+
+You can install beets as docker containers with almost all plugins and all its
+dependencies by simple `docker build -t beets .`.
+
+This container keeps everything inside `/beet` and you can mount your music as
+second volume.
+
+For example if you user's home directory is `/Users/catap` and you keep your
+music at `~/Music` and beet's config and DB as `~/.beet` you can add a simple
+function to your shell rc file like `.zshrc` or `.bashrc`::
+
+    function beet() {
+      docker run --rm -ti -v /Users/catap/.beet:/beet -v /Users/catap/Music:/Music 'beet' $@
+    }
+
+and put config `~/.beet/config.yaml`::
+
+    directory: /Music
+    library: /beet/beets.db
+
+
+After that you can use beets as `docker run --rm -v $(pwd)/beets:/beets beets [CMD]`
+
 Configuring
 -----------
 


catap:
This allows to install `beet` on Synology NAS as simple as:
```
catap@diskstation:~$ cat /usr/local/bin/beet 
#!/bin/sh

sudo docker run --rm --user=$(id -u) --group-add=$(id -g) -ti -v $(echo $HOME)/.beet:/beet -v /volume1/music:/music 'catap/beet' $@
catap@diskstation:~$ 
```

.gitignore
```diff
@@ -89,3 +89,6 @@ ENV/
 /.pydevproject
 /.settings
 .vscode
+
+# GIT stuff, needs for docker
+/.git
```
sampsyo, line 94: Seems odd—wouldn't this be more logical to include in `.dockerignore`, not `.gitignore`?

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+# Copy current sources
+COPY . /src
+
+# install bets
+RUN pip install /src
+
+# install all possibly dependencies for plugins
+RUN pip install pigar
+RUN pigar -p /src/plugins-requirements.txt -P /src
+RUN pip install -r /src/plugins-requirements.txt
```
sampsyo, line 12: I don't entirely understand the need for `pigar` in this context. Would it make sense to just `pip install` beets with several extras to get all those optional dependencies?

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+# Install some 3rd party plugins
+RUN pip install beets-artistcountry
+RUN pip install beets-copyartifacts3
+RUN pip install git+git://github.com/geigerzaehler/beets-check.git@master
+RUN pip install beets-alternatives
+RUN pip install beets-follow
+RUN pip install beets-noimport
+RUN pip install pip install git+git://github.com/igordertigor/beets-usertag.git@master
+RUN pip install pip install git+git://github.com/8h2a/beets-barcode.git@master
+RUN pip install beets-ydl
```
sampsyo, line 23: For a Dockerfile we ship with this repository, I would like to _not_ include any third-party plugins. It seems a little strange to "endorse" some but not all… the way we endorse plugins is by including them with the main beets alternatives.

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+# some well known dependencies
+RUN pip install pylast
+RUN pip install flask
+RUN pip install flask
+RUN pip install beautifulsoup4
```
sampsyo, line 28: These should be covered by the extra dependencies above.

sampsyo (commented):
Thanks for getting this started; it looks like fun! I know a lot of people have rolled their own beets containers, so it will be nice to have a definitive one.

Here are a few suggestions. I would also love to get the attention of a true Docker expert who can audit this for adherence to commonly received "best practices" for writing Dockerfiles.

.gitignore
catap: Nope, because `.dockerignore` should be inside git (and it is a symlink to `.gitignoe`) ;)

Why I added it? Because this is an easy way to add `.git` inside `.dockerignore`

Dockerfile
catap: `pigar` scans everything inside `/src` and put all imports. Why? Because to enable some plugins user also needs to `pip install XXX` that is explained inside docs.

I'm a bit lazy to make it by hand and I've used `pigar` for it.

Dockerfile
catap: I've added all plugins that I can isntall via `pip` and that hadn't beet marked as archived.

Why? Because I'm using `follow` plugging and I did this container mainly for myself :)

Dockerfile
sampsyo: That's the thing—all those dependencies are already listed as "extras" in our setup.py. So we can just use those there:
https://github.com/beetbox/beets/blob/04ea754d00e2873ae9aa2d9e07c5cefd790eaee2/setup.py#L140

catap: I've missed it :(

catap: Yes, this makes things easy. Thanks!

catap: @sampsyo is it possibly to install all plugins by some magic command?

sampsyo: I don't know for sure, but I think the answer is no: https://github.com/beetbox/beets/blob/04ea754d00e2873ae9aa2d9e07c5cefd790eaee2/setup.py#L140

catap: @sampsyo I'm confused. Why? Because I will introduce one more list of plugins.

:(

sampsyo: Yeah, that's not great. But TBH I prefer it a bit to attempting to automatically re-discover dependencies that we have already pinned down based on import names.

catap: @sampsyo defentply. Maybe introducing a new extra "all plugins" is the best way?

sampsyo: No objection there. It seems like there needs to be an "all" list either in the Dockerfile or in setup.py, and I don't immediately see a reason to prioritize one or the other.

catap: @sampsyo any objections why it can't install all dependencies? :)

Honestly docs a bit confused because I read that I need run `pip install requests` and this sort of things, and I feel that plugins aren't so deep integrated.

Now I see that plugins is very deep integrated, and for me it isn't clear why have you used extras here.

sampsyo: Well, that is what pip extras are meant to express: things you might need if you use certain features, but you're not guaranteed to need. Making them optional dependencies means that you don't need to install as many libraries to get started.

.gitignore
jackwilsdon: I think @sampsyo is right here still - `.git` belongs in the `.dockerignore` really ([it seems pretty common to do this](https://github.com/search?l=Ignore+List&q=.git+filename%3A.dockerignore&type=Code)).

Dockerfile
jackwilsdon: The simplest option for users to install additional plugins is probably to make their own Dockerfile and extend ours (it'd be good to document this somewhere), as it keeps the official beets image lightweight and impartial to plugins.

Dockerfile
```diff
@@ -0,0 +1,45 @@
+FROM python:3.8
```
jackwilsdon, line 1: Are we able to extend from the `python:3.8-alpine` image here? It's a bit more lightweight ([35MB](https://microbadger.com/images/python:3.8-alpine) vs [333MB](https://microbadger.com/images/python:3.8)). It's probably worth checking that the majority of our plugins work under alpine though (i.e. their native requirements are available).

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+
+# Copy current sources
+COPY . /src
+
+# install bets
+RUN pip install /src
```
jackwilsdon, line 7: We should merge all of these `pip` commands into a single `RUN` so that we don't end up with a heap of layers.

.gitignore
catap: @jackwilsdon it requires to create this `.dockerfile`; right now I just make a symlink from `.gitingore` to `.dockeringore` because otherwise it requires to add everything into two places

jackwilsdon: Ah I didn't realise that `.dockerignore` was symlinked to `.gitignore` - I'm a bit hesitant about linking them like this as it doesn't really seem to be a common practice.

catap: @jackwilsdon but they very similar by idea :)

Dockerfile
GuilhermeHideki: If I'm not wrong, another option could be use an `ARG` variable, like an env variable to pass this dependencies at image build time, so you can add something like `RUN pip install $BEETS_PLUGINS`

.gitignore
GuilhermeHideki: the dockerignore and gitignore should be separated because it makes sense (to me) to have CI/CD config or the .github folder ignored by docker but not by git, for example.

Also, for "production-ready" Dockerfiles, I usually use an allowlist, so they look a bit different than my .gitignore, like
```ignore
*
!source_code
!dependencies
```

catap: @GuilhermeHideki and your way will include to doker image all `.pyc` files if you make `pip install -e .` ;)

GuilhermeHideki: @catap This should help
```ignore
# Exclude everything
*
# setup.py dependency
!README.rst
# Make sure that source folters are in dockerfile
!beets
!beetsplug
# Ignore cache files
**/__pycache__
**/*.pyc
# Include code / source files
!**/*.py
!**/*.yaml
!**/*.sh
```

catap: @GuilhermeHideki the idea was to not duplicated ignored/not ignored files.

The goal was to have everything that is ignored in one place.

Issue #3847, opened 2021-02-02
Create and update playlists


catap:
### Use case

Let assume that someone has some music library as series of files, and would like to manage also playlist for it.

I feel that it is expected that beets should have a way to create and update playlists per artist, albums, year, etc.

Right now playlist plugin can update paths to actual songs, but not created it.

### Solution

Update playlist plugin to support config like:
```
playlist:
  auto: yes
  relative_to: https://nas.domain.com/music/
  playlist_dir: ~/Music/playlists
  relative_url_encoded: yes
  dynamic_playlists: yes
```

This config creates a dynamic playlists in `~/Music/playlists` for each album, artist, year, genre, etc.

All paths to inside playlists are urlencoded and have form `https://nas.domain.com/music/[urldencoded path]` that allows to use this playlist almost everywhere with NAS that supports webdav/http access to your files.

### Alternatives

A lot of manual job. A lot.

sampsyo:
Hmm… just checking, have you tried the `smartplaylist` plugin? Is this close to what you're imagining?

catap:
@sampsyo looks almost that I need :)

It has missed part with urlencoded, otherwise it looks very close to that I need.

Anyway, I've missed how to write queries. Can you give me an example how can I define a query to create a playlists like for each album of each artist in library?

Thanks!

catap:
nevermind, I figured this out.

PR #3848, opened 2021-02-02
Stream-friendly smart playlists

catap:
## Description

This commit introduced a way to generate a stream-frienldy playlists.

This is that I've missed to make me very happy at #3847

## To Do

- [x] ~figured out that to do with complex unicode strings.~

## ~Unicode-related issue~

~Right now current code encodes at least at python `3.8.7` a string `Ayọ` as `Ayo%CC%A3` when it should be `Ay%E1%BB%8D`.~

~Until it is fixed I ask to don't merge it. @sampsyo if you have any idea how to fix it I'll be appreciated.~

### DIFF
--- a/beetsplug/smartplaylist.py
+++ b/beetsplug/smartplaylist.py
@@ -28,6 +28,12 @@ from beets.dbcore.query import MultipleSort, ParsingError
 import os
 import six
 
+try:
+    from urllib.request import pathname2url
+except ImportError:
+    # python2 is a bit different
+    from urllib import pathname2url
+
 
 class SmartPlaylistPlugin(BeetsPlugin):
 
@@ -39,8 +45,11 @@ class SmartPlaylistPlugin(BeetsPlugin):
             'auto': True,
             'playlists': [],
             'forward_slash': False,
+            'prefix': u'',
+            'urlencode': False,
         })
 
+        self.config['prefix'].redact = True  # May contain username/password.
         self._matched_playlists = None
         self._unmatched_playlists = None
 
@@ -201,6 +210,7 @@ class SmartPlaylistPlugin(BeetsPlugin):
                 if item_path not in m3us[m3u_name]:
                     m3us[m3u_name].append(item_path)
 
+        prefix = bytestring_path(self.config['prefix'].as_str())
         # Write all of the accumulated track lists to files.
         for m3u in m3us:
             m3u_path = normpath(os.path.join(playlist_dir,
@@ -210,6 +220,8 @@ class SmartPlaylistPlugin(BeetsPlugin):
                 for path in m3us[m3u]:
                     if self.config['forward_slash'].get():
                         path = path_as_posix(path)
-                    f.write(path + b'\n')
+                    if self.config['urlencode']:
+                        path = bytestring_path(pathname2url(path))
+                    f.write(prefix + path + b'\n')
 
         self._log.info(u"{0} playlists updated", len(self._matched_playlists))
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2166,6 +2166,9 @@ And many little fixes and improvements:
 * When there's a parse error in a query (for example, when you type a
   malformed date in a :ref:`date query <datequery>`), beets now stops with an
   error instead of silently ignoring the query component.
+* :doc:`/plugins/smartplaylist`: Stream-friendly smart playlists.
+  The ``splupdate`` command can now also add a URL-encodable prefix to every
+  path in the playlist file.
 
 For developers:
 
--- a/docs/plugins/smartplaylist.rst
+++ b/docs/plugins/smartplaylist.rst
@@ -101,3 +101,7 @@ other configuration options are:
   If you intend to use this plugin to generate playlists for MPD on
   Windows, set this to yes.
   Default: Use system separator.
+- **prefix**: Prepend this string to every path in the playlist file. For
+  example, you could use the URL for a server where the music is stored.
+  Default: empty string.
+- **urlencoded**: URL-encode all paths. Default: ``no``.


catap:
Status update here. I guess I discovered a wired bug in macOS:

It reads from FS as o plus modificator, as result I do have:
```
>>> from urllib.request import pathname2url
>>> pathname2url(u'Ayọ')
'Ayo%CC%A3'
>>> pathname2url(u'Ayọ')
'Ay%E1%BB%8D'
>>> 
```

catap:
Seems like using `asciify_paths` is a good way to do not care about unicode here.

beetsplug/smartplaylist.py
```diff
@@ -39,8 +44,12 @@ def __init__(self):
             'auto': True,
             'playlists': [],
             'forward_slash': False,
+            'started_with': u'',
+            'urlencoded': False,
```
sampsyo, line 48: Can we just call this `urlencode`?

beetsplug/smartplaylist.py
```diff
@@ -39,8 +44,12 @@ def __init__(self):
             'auto': True,
             'playlists': [],
             'forward_slash': False,
+            'started_with': u'',
```
sampsyo, line 47: Can we call this `prefix`?

beetsplug/smartplaylist.py
```diff
@@ -201,6 +210,7 @@ def update_playlists(self, lib):
                 if item_path not in m3us[m3u_name]:
                     m3us[m3u_name].append(item_path)
 
+        started_with = bytestring_path(self.config['started_with'].get())
```
sampsyo, line 213: I don't think `bytestring_path` is the right thing here because this isn't necessarily a path. I think we should encode it with whatever the text encoding is for the playlist file?

beetsplug/smartplaylist.py
```diff
@@ -210,6 +220,8 @@ def update_playlists(self, lib):
                 for path in m3us[m3u]:
                     if self.config['forward_slash'].get():
                         path = path_as_posix(path)
-                    f.write(path + b'\n')
+                    if self.config['urlencoded'].get():
```
sampsyo, line 223:
```suggestion
                    if self.config['urlencoded']:
```

beetsplug/smartplaylist.py
```diff
@@ -210,6 +220,8 @@ def update_playlists(self, lib):
...
+                        path = bytestring_path(pathname2url(path))
```
sampsyo, line 224: Again, I'm not sure `bytestring_path` is the right thing here…

docs/plugins/smartplaylist.rst
```diff
@@ -101,3 +121,6 @@ other configuration options are:
   If you intend to use this plugin to generate playlists for MPD on
   Windows, set this to yes.
   Default: Use system separator.
+- **started_with**: Started each path to track with specified string, for
+  example path to server where you host your music. Default empty string.
```
sampsyo, line 125: > Prepend this string to every path in the playlist file. For example, you could use the URL for a server where the music is stored. Default: empty string.

docs/plugins/smartplaylist.rst
```diff
@@ -101,3 +121,6 @@ other configuration options are:
...
+- **urlencoded**: URL encode all paths. Default ``no``.
```
sampsyo, line 126:
```suggestion
- **urlencoded**: URL-encode all paths. Default: ``no``.
```

docs/plugins/smartplaylist.rst
```diff
@@ -30,6 +30,26 @@ use spaces in the query, be sure to quote them (e.g., ``artist:"The Beatles"``).
 If you have existing files with the same names, you should back them up---they
 will be overwritten when the plugin runs.
 
+You also can generate a playlists to play music on your NAS or any server. Let
+assume that you host all your music at ``https://nas.domain.com/music``, that is
+mounted to your machine when you run beet as ``/mnt/music`` and you should login
+to it as user ``user`` with password ``secret``::
```
sampsyo, line 36: Could we put this in a separate section, because it's separate from the basic instructions? Also, I'm not sure we need an example config with all the other settings, such as `playlists:`. Can we keep the example compact (if an example is necessary at all)?

sampsyo (commented):
Seems interesting; thanks for getting this started. Could you please add a changelog entry?

beetsplug/smartplaylist.py
catap: Hm, shall I change `forward_slash` also?

beetsplug/smartplaylist.py
catap: @sampsyo without it CI fails ;)

Let me show it as separated commit

docs/plugins/smartplaylist.rst
catap: Let me remove this example because it is tricky one.

catap:
@sampsyo ping here

beetsplug/smartplaylist.py
```diff
@@ -210,6 +219,8 @@ def update_playlists(self, lib):
                 for path in m3us[m3u]:
                     if self.config['forward_slash'].get():
                         path = path_as_posix(path)
-                    f.write(path + b'\n')
+                    if self.config['urlencode'].get():
```
sampsyo, line 222:
```suggestion
                    if self.config['urlencode']:
```

beetsplug/smartplaylist.py
```diff
@@ -39,8 +44,12 @@ def __init__(self):
             'auto': True,
             'playlists': [],
             'forward_slash': False,
+            'prefix': u'',
+            'urlencode': False,
         })
 
+        # may contains user and password
+        self.config['prefix'].redact = True
```
sampsyo, line 52:
```suggestion
        self.config['prefix'].redact = True  # May contain username/password.

```

Preferring a blank line afterward because this is unrelated to the next two lines.

sampsyo (changes requested):
Looking pretty good! Could you please add a changelog entry?

catap:
@sampsyo ;)

beetsplug/smartplaylist.py
```diff
@@ -201,6 +209,7 @@ def update_playlists(self, lib):
                 if item_path not in m3us[m3u_name]:
                     m3us[m3u_name].append(item_path)
 
+        started_with = bytestring_path(self.config['prefix'].get())
```
sampsyo, line 212: `get` should be `as_str` here to enforce a type on the configuration. And can we call this variable `prefix`?

sampsyo (changes requested):
Oops; noticed one last small thing (or two)! Sorry for letting this slip by last time. This should be good to go afterward.

beetsplug/smartplaylist.py
```diff
@@ -28,6 +28,11 @@
 import os
 import six
 
+try:
+    from urllib.request import pathname2url
+except ImportError:
+    from urllib import pathname2url
+
```
justinmayer, line 35: I suggest adding a comment above to indicate which Python version(s) this compatibility block is intended to support. That will make it easier to find and remove this stanza when Beets drops support for said Python versions(s) in the future.

docs/changelog.rst
```diff
@@ -2066,6 +2066,9 @@ And many little fixes and improvements:
 * When there's a parse error in a query (for example, when you type a
   malformed date in a :ref:`date query <datequery>`), beets now stops with an
   error instead of silently ignoring the query component.
+* :doc:`/plugins/smartplaylist`: Stream-friendly smart playlists.
+  The ``splupdate`` command can now also add some prefix to every path
+  in the playlist file which can be URL-encode.
```
justinmayer, line 2071: Suggested re-phrasing:
```
The ``splupdate`` command can now also add a URL-encodable prefix to every path
in the playlist file.
```

justinmayer (commented):
Nice work, Kirill. I made a few minor suggestions. 💫

catap:
I defently forgot about it. Will try to finish in few days.

catap:
@sampsyo / @justinmayer a few days was a bit longer... well.. It is fixed! :)

sampsyo (approved):
Looks great! Thanks for seeing it though!!

PR #3851, opened 2021-02-08
switch encoding and finishing log to debug

vincentDcmps:
## Description

hello, when I have import large album file I have a lot of log entry from convert plugging for notify beginning and finishing encoding
that cause that sometime you don't see that beet is waiting an input, I suggest to switch this two entry in debug log

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -200,10 +200,7 @@ class ConvertPlugin(BeetsPlugin):
         assert isinstance(source, bytes)
         assert isinstance(dest, bytes)
 
-        quiet = self.config['quiet'].get(bool)
-
-        if not quiet and not pretend:
-            self._log.info(u'Encoding {0}', util.displayable_path(source))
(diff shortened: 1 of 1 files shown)


jackwilsdon:
Is there a reason you aren't running it in quiet mode by setting `quiet: true` in your configuration?

vincentDcmps:
because with the quiet mode you don't have input selection isn't it?

wisp3rwind:
@vincentDcmps I think @jackwilsdon is referring to the [`quiet` flag of the convert plugin](https://beets.readthedocs.io/en/latest/plugins/convert.html#configuration), not the global quiet setting.

There might still be some room for improvement, if someone would work out how to best address the following situation: You might want to suppress convert progress output during imports, but not when running `beet convert ...`. There's currently no way to configure beets to do so.

vincentDcmps:
ok my bad I haven't read documentation correctly quiet mod work for my use case

PR #3859, opened 2021-02-22
Add strip_path to mpdstats

akelge:
## Description

Add `strip_path` option to mpdstats plugin. Combined with `music_directory` option it can help mangling the path received from MPD daemon to match beets library one.

## To Do

### DIFF
--- a/beetsplug/mpdstats.py
+++ b/beetsplug/mpdstats.py
@@ -127,8 +127,9 @@ class MPDClientWrapper(object):
         entry = self.get('currentsong')
         if 'file' in entry:
             if not is_url(entry['file']):
-                file = entry['file'].replace(self.strip_path, '', count=1)
-                result = os.path.join(self.music_directory, file)
+                if entry.startswith(self.strip_path):
+                    entry = entry[len(self.strip_path):]
+                result = os.path.join(self.music_directory, entry)
             else:
                 result = entry['file']
         return result, entry.get('id')
--- a/beetsplug/mpdstats.py
+++ b/beetsplug/mpdstats.py
@@ -127,9 +127,10 @@ class MPDClientWrapper(object):
         entry = self.get('currentsong')
         if 'file' in entry:
             if not is_url(entry['file']):
-                if entry.startswith(self.strip_path):
-                    entry = entry[len(self.strip_path):]
-                result = os.path.join(self.music_directory, entry)
+                file = entry['file']
+                if file.startswith(self.strip_path):
+                    file = file[len(self.strip_path):]
+                result = os.path.join(self.music_directory, file)
             else:
                 result = entry['file']
         return result, entry.get('id')
--- a/beetsplug/mpdstats.py
+++ b/beetsplug/mpdstats.py
@@ -53,6 +53,13 @@ class MPDClientWrapper(object):
         self.music_directory = mpd_config['music_directory'].as_str()
         self.strip_path = mpd_config['strip_path'].as_str()
 
+        # Ensure strip_path end with '/'
+        if not self.strip_path.endswith('/'):
+            self.strip_path += '/'
+
+        self._log.debug('music_directory: {0}', self.music_directory)
+        self._log.debug('strip_path: {0}', self.strip_path)
+
         if sys.version_info < (3, 0):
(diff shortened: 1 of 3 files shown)


beetsplug/mpdstats.py
```diff
@@ -118,12 +121,16 @@ def currentsong(self):
...
+        In some cases, we need to remove the local path from MPD server,
+        we replace 'strip_path' with ''.
+        `strip_path` defaults to ''.
         """
         result = None
         entry = self.get('currentsong')
         if 'file' in entry:
             if not is_url(entry['file']):
-                result = os.path.join(self.music_directory, entry['file'])
+                file = entry['file'].replace(self.strip_path, '')
```
sampsyo, line 132: Is it perhaps worrisome that this will replace *all* occurrences of `strip_path` from the string—not just the leftmost one?

beetsplug/mpdstats.py
```diff
@@ -53,6 +53,9 @@ def __init__(self, log):
         self.music_directory = (
             mpd_config['music_directory'].as_str())
 
+        self.strip_path = (
+            mpd_config['strip_path'].as_str())
```
sampsyo, line 57: This can probably fit one one line (and therefore not need the somewhat odd-looking parentheses), right?

docs/plugins/mpdstats.rst
```diff
@@ -53,6 +53,9 @@ configuration file. The available options are:
 - **music_directory**: If your MPD library is at a different location from the
   beets library (e.g., because one is mounted on a NFS share), specify the path
   here.
+- **strip_path**: If your MPD library contains local path, specify the part to remove
+  here. Combining this with **music_directory** you can mangle MPD path to match the 
+  **beets library** one.
```
sampsyo, line 58: I would probably not bold **beets library** because it is not the name of an option.

sampsyo (commented):
Looks pretty straightforward; thanks!! There are a couple of comments within. Could you please also add a changelog entry?

beetsplug/mpdstats.py
akelge: Done, moved to one line

docs/plugins/mpdstats.rst
akelge: Fixed, makes sense

beetsplug/mpdstats.py
akelge: Added count=1 to replace to stop at first occurence

jackwilsdon: I don't think this will guarantee the first occurrence still? I think something like the following is probably best:

```Python
file = entry['file']

if file.startswith(self.strip_path):
    file = file[len(self.strip_path):]
```

It seems like Python 3.9+ has `removeprefix` but I think we still support <3.9.

sampsyo: Yes indeed, thanks @jackwilsdon! A little utility function that behaves like `removeprefix` may be in order.

akelge: I see your point, indeed a safer approach

akelge: I changed the code accordingly, please check latest commit

beetsplug/mpdstats.py
```diff
@@ -50,8 +50,9 @@ class MPDClientWrapper(object):
     def __init__(self, log):
         self._log = log
 
-        self.music_directory = (
-            mpd_config['music_directory'].as_str())
+        self.music_directory = (mpd_config['music_directory'].as_str())
+
+        self.strip_path = (mpd_config['strip_path'].as_str())
```
sampsyo, line 55:
```suggestion
        self.music_directory = mpd_config['music_directory'].as_str()
        self.strip_path = mpd_config['strip_path'].as_str()
```
No need for parentheses now (or the blank line).

sampsyo (commented):
Looks about right to me! Just one style suggestion within and we should be good to go. :sparkles: :rocket:

beetsplug/mpdstats.py
akelge: Fixed in last commit :)

sampsyo (approved):
Looks good to me; thanks for following through!!

PR #3868, opened 2021-03-06
Web artpath fix

### LINKED ISSUE #3866: web: GET /album/<n> hides artpath even when INCLUDE_PATHS is set
### Problem

`beet web` provides GET operations to fetch track items and albums. By default this removes paths
from the returned data, but the config setting INCLUDE_PATHS should allow the paths to be returned.

This works correctly for GET /item/... but not GET /album/... In the album case, the artpath is unconditionally deleted from the results.

### Setup

Add to config file:

web:
    include_paths: true

Use `beet web` to make a webserver available and do a GET /album/N, where N is the album id of an album in the database which has a cover art set. The JSON result should include the 'artpath' value but does not.

* OS: Linux (Debian Testing)
* Python version: 3.9.1-1
* beets version: 1.4.9-7
* Turning off plugins made problem go away (yes/no): bug in web plugin

Note this is a small issue, although I have hit it. I have a fix (and a regression test) which I will submit as a small PR once my first PR has been finished (so I can learn from the mistakes I made in that!).

GrahamCobb:
## Description

Fixes #3866.  

Change web/__init__.py to only remove `artpath` if `include_paths` is false.

## To Do

- [X] Documentation. No substantive change but made /art clearer.
- [X] Changelog. Added to bugs list.
- [X] Tests. Tests added.

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -59,7 +59,10 @@ def _rep(obj, expand=False):
         return out
 
     elif isinstance(obj, beets.library.Album):
-        del out['artpath']
+        if app.config.get('INCLUDE_PATHS', False):
+            out['artpath'] = util.displayable_path(out['artpath'])
+        else:
+            del out['artpath']
         if expand:
             out['items'] = [_rep(item) for item in obj.items()]
         return out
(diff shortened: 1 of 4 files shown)


sampsyo (approved):
Wahoo; seems perfect! Many thanks for the quick work!! :tada:

PR #3869, opened 2021-03-08
Web regex fix

### LINKED ISSUE #3867: web: web page search box doesn't work for regex searches
### Problem

This is not a problem in the web API itself, but in the web pages which provide the simple web user interface.

Bringing up the web interface and entering a query such as ``somefield::.`` never returns any results.
The problem is that the web page ends up double URI encoding the search before passing it to GET /item/query.

I have a fix (in `static/beets.js`) which I can submit once the current PR is done.

However, I have no idea how to create a test for this as it would mean starting the webserver, submitting an HTTP request and checking the resulting (complex) HTML. Does anyone have any example of doing that in the beets pytest environment? I know very little python and nothing about pytest but I may be able to steal a similar test if one exists!

EDIT: Actually, it is the last step - parsing and checking the resulting HTML which is hard (the rest is what the tests already do - but they are dealing with JSON responses, not HTML responses). Does anyone have any tools or examples of checking HTML responses? Or do I just do some simple string searches and hope nothing changes too much to change the page in the future?

GrahamCobb:
## Description

Fixes #3867. 

After discussion in #3867, this fix prevents replacement of `\` in web queries if and only if they include `::`, indicating that they are regex queries.

Several new tests have been created to test various queries of items and albums.

No documentation changes as this is expected behaviour.

## To Do

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -244,7 +244,9 @@ class QueryConverter(PathConverter):
 
     def to_python(self, value):
         queries = value.split('/')
-        return [query.replace('\\', os.sep) for query in queries]
+        """Do not do path substitution on regex value tests"""
+        return [query if '::' in query else query.replace('\\', os.sep)
+                for query in queries]
 
     def to_url(self, value):
         return ','.join([v.replace(os.sep, '\\') for v in value])
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -186,6 +186,8 @@ New features:
 
 Fixes:
 
+* :bug:`/plugins/web`: Allow use of backslash in regex web queries.
+  :bug:`3867`
 * :bug:`/plugins/web`: Fixed a small bug which caused album artpath to be
   redacted even when ``include_paths`` option is set.
   :bug:`3866`
--- a/test/test_web.py
+++ b/test/test_web.py
@@ -26,12 +26,22 @@ class WebPluginTest(_common.LibTestCase):
         # Add library elements. Note that self.lib.add overrides any "id=<n>"
         # and assigns the next free id number.
         # The following adds will create items #1, #2 and #3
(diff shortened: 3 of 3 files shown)


sampsyo:
Overall looking good! It seems like the test is failing on Windows currently, presumably due to differences in `os.sep`.

GrahamCobb:
Is there any way to get the log of the failing test on Windows? I don't own a single Windows system any more!

If not, I guess I could just try changing all the paths I can find in the tests to use os-independent tools and then submit for another go. But it would be nice to know I an fixing the right things :-)

sampsyo:
Sure thing! It's in our GitHub checks, from AppVeyor. You can find them under the little red "X" next to your most recent commit. Here's the most recent direct link to the AppVeyor job: https://ci.appveyor.com/project/beetbox/beets/builds/38118248

GrahamCobb:
My fix for Windows still doesn't work. Is there any way to run the CI Windows test without updating a PR? (I.e. on my changes before committing them?). Without a Windows system.

I am assuming not. So, I am going to create another separate PR which I will use just for forcing the tests to run and getting the output to examine. This should eventually be rejected, but I don't want to pollute the commit history for the mainline with lots of debug stuff by doing it in this PR.

sampsyo:
One option is to "mock" the `os` module to force a particular platform. You can see some of this going on in the `test_query` tests:
https://github.com/beetbox/beets/blob/3e8261393d57e33edd7e7956b71245f3d249d7c7/test/test_query.py#L589-L591

GrahamCobb:
Thanks. I have spent a few hours playing with that but can't get it to work for me. I will go back to the brute force method of a useless PR.

GrahamCobb:
I believe this is now complete - please consider merging it. I will open a new PR in a new branch for the readonly feature.

sampsyo:
Wow! Nice work tracking down a way around the test issues, and also great work cleaning up some miscellaneous issues in the tests. These look great; thank you!!

PR #3871, opened 2021-03-09
Web regex debug

GrahamCobb:
## Description

This isn't a real PR - it is a way to get the Windows tests to be run for me. This PR should be refused, but not yet, please!

Sorry for the noise.

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -244,7 +244,9 @@ class QueryConverter(PathConverter):
 
     def to_python(self, value):
         queries = value.split('/')
-        return [query.replace('\\', os.sep) for query in queries]
+        """Do not do path substitution on regex value tests"""
+        return [query if '::' in query else query.replace('\\', os.sep)
+                for query in queries]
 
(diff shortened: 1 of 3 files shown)


wisp3rwind:
> Sorry for the noise.

No need to worry! I'm not quite sure since I couldn't find this spelled out explicitly in Github's docs, but it might be that you can avoid notifying people by marking the PR as a draft right when opening it.

GrahamCobb:
Thanks for your help. I will try the draft thing if I need to do this again. I think I have now fixed the problem I was working on (and have pushed an update for #3869) but if you don't mind I will leave this PR open for another couple of days in case another problem with CI failures I can't reproduce appears.

wisp3rwind:
> if you don't mind I will leave this PR open for another couple of days in case another problem with CI failures I can't reproduce appears.

Sure, using PRs in this way is absolutely fine. I simply marked as a Draft since I think that's a useful way to highlight that "This PR should be refused". If it helps suppress some notifications, all the better, but it would be fine even if it didn't.

PR #3873, opened 2021-03-10
Refactor code quality issues

akshgpt7:
## Description

Hi, I ran [DeepSource](https://deepsource.io/) analysis on my fork of the repo, and found some [interesting code quality and performance issues](https://deepsource.io/gh/akshgpt7/beets/) that can improve the general code quality and performance here.
This PR fixes a few of the issues detected.

#### Summary of changes
- Use literals instead of `dict()` calls to improve performance.
- Use safer `sys.exit()` instead of bare `exit()`.
- Merge `isinstance` calls.
- Simplify `if` statement.
- Refactor unneeded list comprehension.

## To Do

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -729,10 +729,10 @@ class Results(object):
     def _get_indexed_flex_attrs(self):
         """ Index flexible attributes by the entity id they belong to
         """
-        flex_values = dict()
+        flex_values = {}
         for row in self.flex_rows:
             if row['entity_id'] not in flex_values:
-                flex_values[row['entity_id']] = dict()
+                flex_values[row['entity_id']] = {}
 
             flex_values[row['entity_id']][row['key']] = row['value']
 
--- a/beetsplug/bpd/__init__.py
+++ b/beetsplug/bpd/__init__.py
@@ -21,6 +21,7 @@ use of the wide range of MPD clients.
 from __future__ import division, absolute_import, print_function
 
 import re
+import sys
 from string import Template
 import traceback
 import random
@@ -334,7 +335,7 @@ class BaseServer(object):
 
     def cmd_kill(self, conn):
         """Exits the server process."""
-        exit(0)
+        sys.exit(0)
 
     def cmd_close(self, conn):
         """Closes the connection."""
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -187,7 +187,7 @@ class ImportSession(object):
         self.logger = self._setup_logging(loghandler)
         self.paths = paths
         self.query = query
-        self._is_resuming = dict()
+        self._is_resuming = {}
         self._merged_items = set()
         self._merged_dirs = set()
 
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -791,7 +791,7 @@ def _store_dict(option, opt_str, value, parser):
     if option_values is None:
         # This is the first supplied ``key=value`` pair of option.
         # Initialize empty dictionary and get a reference to it.
-        setattr(parser.values, dest, dict())
+        setattr(parser.values, dest, {})
         option_values = getattr(parser.values, dest)
 
     try:
--- a/beetsplug/export.py
+++ b/beetsplug/export.py
@@ -33,7 +33,7 @@ from beetsplug.info import make_key_filter, library_data, tag_data
 class ExportEncoder(json.JSONEncoder):
     """Deals with dates because JSON doesn't have a standard"""
     def default(self, o):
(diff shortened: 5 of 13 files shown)


wisp3rwind:
Hi!

> Hi, I ran [DeepSource](https://deepsource.io/) analysis on my fork of the repo, and found some [interesting code quality and performance issues](https://deepsource.io/gh/akshgpt7/beets/) that can improve the general code quality and performance here.

That does look like some useful static analysis. From a quick look, I rather doubt that most of this is actually affecting performance in any meaningful way, but more consistent and pythonic style alone makes for a good point of merging this.

You do fail to mention that you actually work for DeepSource, I feel like that should have been an obvious disclaimer (by the way: This has been pointed out before https://github.com/encode/httpx/pull/1501#issuecomment-794271355). In particular, you only fixed a few of the most simple issues that the tool found. I appears to me that many issues would require more carefull review to verify that the proposed fix is sensible. Given that ~800 issues are hightlighted, that is a substantial amount of work.
I think there a few points here:
- The style fixes in this PR could be merged as-is.
- We need to decide whether it is worthwhile to add DeepSource to our CI setup. If not, `.deepsoure.toml` should not be committed. If we do, we need to decide to place some trust in the service: I believe someone (...) would need to set up an account with DeepSource. It appears that DeepSource requests to "Act on your behalf" when logging in with Github, irrespective of whether or not one enables features that require this permission.
- Can DeepSource show only new issues introduced with PRs? It might not be feasible (or even correct) to address each and every issue, so without a means to filter for new problems, I see little use of the tool. In any case, the value will be little if no-one addresses the current "problems" - I won't do that in the near-term; is there someone else who would?

Note that this is  my personal opinion - not necessarily shared by everyone involved with beets. I'm not saying that DeepSource is not a useful tool, but it appears a lot like only a minimum effort was made in order to disguise what is essentially an ad. The decision on whether or not to use DeepSource should be independent of the style fixes from this PR.

test/test_replaygain.py
```diff
@@ -32,10 +32,9 @@
 except (ImportError, ValueError):
     GST_AVAILABLE = False
 
-if any(has_program(cmd, ['-v']) for cmd in ['mp3gain', 'aacgain']):
-    GAIN_PROG_AVAILABLE = True
-else:
-    GAIN_PROG_AVAILABLE = False
+GAIN_PROG_AVAILABLE = bool(
+    any(has_program(cmd, ['-v']) for cmd in ['mp3gain', 'aacgain'])
+)
```
sampsyo, line 37: Not wild about this one—I know it's equivalent, but for constant declarations, I think it's nicer to see the check made explicitly.

beets/dbcore/db.py
```diff
@@ -729,10 +729,10 @@ def __iter__(self):
     def _get_indexed_flex_attrs(self):
         """ Index flexible attributes by the entity id they belong to
         """
-        flex_values = dict()
+        flex_values = {}
```
sampsyo, line 732: Sure; `dict()` -> `{}` seems perfectly legitimate as a style improvement.

beetsplug/bpd/__init__.py
```diff
@@ -334,7 +335,7 @@ def cmd_idle(self, conn, *subsystems):
 
     def cmd_kill(self, conn):
         """Exits the server process."""
-        exit(0)
+        sys.exit(0)
```
sampsyo, line 338: Not sure why this is "safer"? I don't exactly object, but it doesn't seem too important either.

sampsyo (commented):
Thanks for writing up those high-level comments, @wisp3rwind, which reflect my thoughts exactly. For what it's worth, here are a few comments on the specific changes.

beetsplug/bpd/__init__.py
wisp3rwind: There's a "Description" link next to the issue on Deepsource with some reasoning. I didn't research this any further, though.

test/test_replaygain.py
akshgpt7: Fair enough, I've reverted the change.

sampsyo:
So we can resolve this for now, would you mind removing the DeepSource config file, at least for now? I think I'm in favor of the remaining changes at this point, so if others concur, we could just merge those fixes.

akshgpt7:
@sampsyo Sure, I've removed the file.

sampsyo:
Awesome; thanks! If @wisp3rwind and/or others are OK with these changes, I think we can merge them?

wisp3rwind:
Yes, that looks good to me, too. Thanks @akshgpt7 for adjusting the PR!

Since no one spoke up, I guess it doesn't look we'll investigate setting up DeepSource for now

PR #3877, opened 2021-03-11
Web readonly

### LINKED ISSUE #3870: web: add readonly config setting
### Use case

I want to use the web plugin to allow queries of my music on my LAN, but I want to maintain full control over updates, with no updates or deletions allowed using the web API.

Of course, I am not exposing the beets web API to the internet, but I don't even want to risk accidental changes by a web user if the web screens were to be updated to allow that, for example.

### Solution

While it is possible to use various external security capabilities, such as hiding behind a full-featured web server, etc., I would like a simple config setting to allow me to say "my beets web is only for read-only access".

I propose adding a simple config setting for web: `readonly` with a boolean true/false value. It would need to default to false since it was not added at the same time as DELETE and PATCH were added, but would still be useful to allow read-only usage.

I expect to have a simple implementation in a couple of days.

### Alternatives
- A full access control mechanism with accounts might be nice but is probably overkill. A simple `readonly` setting to disallow DELETE and PATCH and return to only allowing queries would be enough initially, and should have been provided when the additional commands were added.
- Presumably a separate user could be set up with read-only access to the database, but that is a lot of work just to prevent something which should not have been allowed by default in the first place.

GrahamCobb:
## Description

Fixes #3870

This PR adds a new web configuration option ``readonly``. If ``readonly`` is true (the default)
only _GET_ operations are allowed. ``readonly`` must be false to allow _DELETE_ or _PATCH_ operations.

**NOTE**: this is a change, which may break existing users of the _DELETE_ or _PATCH_ operations, but after discussion in #3870 it was decided that the default has to be read only. Re-enabling them is just a matter of adding ``readonly: no`` to the web configuration.

## To Do

- [X] Documentation. Documentation added for new rules and the ``readonly`` config option.
- [X] Changelog. Added
- [X] Tests. Many tests for _DELETE_ and _PATCH_ operations (with and without ``readonly``) have been added.

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -442,6 +442,7 @@ class WebPlugin(BeetsPlugin):
             'cors_supports_credentials': False,
             'reverse_proxy': False,
             'include_paths': False,
+            'readonly': True,
         })
 
     def commands(self):
@@ -461,6 +462,7 @@ class WebPlugin(BeetsPlugin):
             app.config['JSONIFY_PRETTYPRINT_REGULAR'] = False
 
             app.config['INCLUDE_PATHS'] = self.config['include_paths']
+            app.config['READONLY'] = self.config['readonly']
 
             # Enable CORS if required.
             if self.config['cors']:
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -116,6 +116,7 @@ def resource(name, patchable=False):
             entities = [entity for entity in entities if entity]
 
             if get_method() == "DELETE":
+
                 if app.config.get('READONLY', True):
                     return flask.abort(405)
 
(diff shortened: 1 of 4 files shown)


sampsyo:
Yay; thanks for getting this going!

I feel a little silly, but there's one major thing I'm not quite grokking about this code: what's the connection between `app.config['READONLY']` (i.e., the Flask app configuration) and `self.config['readonly']` (i.e., the beets configuration loaded from the `web:` section of `config.yaml`)? It looks like the code uses the former, but the documentation refers to the later. I feel like there may be some mechanism in here for setting one based on the other, but I don't quite see it—is there something I'm missing, perhaps?

GrahamCobb:
Oops! Great spot. It is me who needs to feel silly.

I spent so much time working on testing various different code paths in the test harness, I never did a real test with setting `readonly` false in a config file! And I forgot to set the app config from the file. 

Just goes to show that although tests are good, they are no substitute for careful reviewers.

Will update soon. Thanks.

GrahamCobb:
Now fixed. Thanks for noticing the problem!

sampsyo (approved):
Awesome; this looks perfect! Thanks once again for the extensive testing here.

Issue #3882, opened 2021-03-13
Query with asymmetric matching for accented characters


GrahamCobb:
### Use case

OK - let me admit it before I start... I am a dumb native English speaker and I don't like letters with accents and don't even know how to enter most of them on my keyboard. Yep, I know: the rest of the world despises me.

However, for people like me (and, to be fair, some other cases where accented letter use varies even between speakers of the same or very closely related languages, or has differed at various historical times), it would be very nice to have the __optional__ ability to do simplified comparisons when searching in beets.

Today, in beets, if I use `beet ls -a dvořák` I get 5 matches in my database. However, if I use `beet ls -a dvorak` I get no matches. Although I support MusicBrainz's requirement to use correct spelling, as a non-speaker of the language I would have to search online to look up the correct accents to use. It would be useful if beets allowed (perhaps optionally, based on a config setting) less strict searching. The case is even worse with Noel, which appears in various places in my database in both the forms Noël and Noel - similar problems exist for other partly _non-naturalised_ words which are sometimes spelt in English with accents and sometimes not.

### Solution

A popular, powerful and standardised by the Unicode consortium approach, is called [asymmetric matching](https://www.unicode.org/reports/tr10/tr10-41.html#Asymmetric_Search). In this case, speakers choose a language (e.g. English) and can write search terms using the characters of that language: the system then matches accented letters not present in that language with the unaccented forms in the query. However, if the user actually enters an accent in the search term, then only the same accented letter matches. The link above includes some examples of using the term "resume" to match the many words spelt "resume" but with accents on various different letters.

This solution feels very natural and is easy for both speakers and non-speakers of the language to use. Unfortunately, I have not found an existing Python implementation (but I haven't looked particularly hard).

### Alternatives

Regex searches can provide some capabilities but they require particular fields to be specified and require the use of the arcane regex format. For example, the Dvorak search would probably need to be entered as `artist::Dvo.a.' and requires me to know which letters are accented. It would also miss cases where the composer is mentioned in the title or albumtitle but not as an artist.

jackwilsdon:
The `fuzzy` plugin already kind of implements this, but instead using `difflib` with a threshold.

GrahamCobb:
Thanks for the suggestion but, unfortunately, `fuzzy` doesn't seem to work usefully for these sorts of cases.

[...]

PR #3883, opened 2021-03-15
Experimental "bare-ASCII" matching query

GrahamCobb:
Create prototype plugin for "bare-ASCII" matching query. Plugin is called bareasc and defaults to using # as the prefix.

## Description

For investigation and further discussion in feature request #3882.

## To Do

### DIFF
--- a/beetsplug/bareasc.py
+++ b/beetsplug/bareasc.py
@@ -78,7 +78,9 @@ class BareascPlugin(BeetsPlugin):
         # Copied from commands.py - list_items
         if album:
             for album in lib.albums(query):
-                print_(unidecode(format(album, fmt)))
+                bare = unidecode(six.ensure_text(format(album, fmt)))
+                print_(six.ensure_str(bare))
         else:
             for item in lib.items(query):
-                print_(six.ensure_str(unidecode(format(item, fmt))))
+                bare = unidecode(six.ensure_text(format(item, fmt)))
+                print_(six.ensure_str(bare))
--- a/beetsplug/bareasc.py
+++ b/beetsplug/bareasc.py
@@ -74,13 +74,12 @@ class BareascPlugin(BeetsPlugin):
         """Emulate normal 'list' command but with unidecode output."""
         query = decargs(args)
         album = opts.album
-        fmt = u''
         # Copied from commands.py - list_items
         if album:
             for album in lib.albums(query):
-                bare = unidecode(six.ensure_text(format(album, fmt)))
+                bare = unidecode(six.ensure_text(str(album)))
                 print_(six.ensure_text(bare))
         else:
             for item in lib.items(query):
-                bare = unidecode(six.ensure_text(format(item, fmt)))
+                bare = unidecode(six.ensure_text(str(item)))
                 print_(six.ensure_text(bare))
--- a/beetsplug/bareasc.py
+++ b/beetsplug/bareasc.py
@@ -26,6 +26,7 @@ from beets.ui import print_, decargs
 from beets.plugins import BeetsPlugin
 from beets.dbcore.query import StringFieldQuery
 from unidecode import unidecode
+import six
 
 
 class BareascQuery(StringFieldQuery):
@@ -80,4 +81,4 @@ class BareascPlugin(BeetsPlugin):
                 print_(unidecode(format(album, fmt)))
         else:
             for item in lib.items(query):
-                print_(unidecode(format(item, fmt)))
+                print_(six.ensure_str(unidecode(format(item, fmt))))
--- a/beetsplug/bareasc.py
+++ b/beetsplug/bareasc.py
@@ -30,7 +30,6 @@ from unidecode import unidecode
 class BareascQuery(StringFieldQuery):
     @classmethod
     def string_match(cls, pattern, val):
-        print('In BareascQuery' + ' ' + pattern + ' ' + val)
         # smartcase
         if pattern.islower():
             val = val.lower()
--- a/beetsplug/bareasc.py
+++ b/beetsplug/bareasc.py
@@ -79,8 +79,8 @@ class BareascPlugin(BeetsPlugin):
         if album:
             for album in lib.albums(query):
                 bare = unidecode(six.ensure_text(format(album, fmt)))
-                print_(six.ensure_str(bare))
+                print_(six.ensure_text(bare))
         else:
             for item in lib.items(query):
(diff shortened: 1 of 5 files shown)


sampsyo:
Awesome! I think this is really great to have, and it's fantastic that the implementation is so simple. Wahoo!

It sounds from https://github.com/beetbox/beets/issues/3882#issuecomment-799318251 that the speed is actually not _that_ bad, which is encouraging. In any case, I think we should forge ahead with this approach and worry about performance improvements later, if at all. :smiley:

Our CI checker is obviously being too aggressive about requiring docstrings in test files. Those should definitely be exempted from the requirement to have a docstring on every function/method…

GrahamCobb:
I think this is now working and is actually very useful to me. It would be nice to get feedback/testing from others in the feature discussion #3882 but it is up to you whether you want to merge without that.

sampsyo:
Yeah! What do you think about making a [Discourse](https://discourse.beets.io) post inviting testers to try it out?

GrahamCobb:
Good idea. Do you want me to point them to the PR or do you want to merge it and then we just tell them it is available for testing and comments?

Or do you want to make the post? Either works for me.

sampsyo:
I think pointing them to the PR makes sense for now? And yes, if you don't mind, it would be great to let you take all the credit! :smiley:

beetsplug/bareasc.py
```diff
@@ -0,0 +1,83 @@
...
+            u'Example: %prog -f \'$album: $title\' artist:beatles'
+        cmd.parser.add_all_common_options()
+        cmd.func = self.unidecode_list
+        return [cmd]
+
+    def unidecode_list(self, lib, opts, args):
+        """Emulate normal 'list' command but with unidecode output."""
+        query = decargs(args)
+        album = opts.album
+        fmt = u''
```
sampsyo, line 76: Is this intentional? Maybe it's supposed to get `fmt` from the options?

docs/plugins/bareasc.rst
```diff
@@ -0,0 +1,68 @@
...
+the output is passed through the bare-ASCII transformation before being printed.
+This allows you to easily check what the library data looks like in bare ASCII,
+which can be useful if you are trying to work out why a query is not matching.
+
+Using the same example track as above::
+
+    $ beet bareasc 'Dvořàk'
+    Istvan Kertesz - REQUIEM - Dvorak: Requiem, op.89 - Confutatis maledictis
+
+Note: the ``bareasc`` command does NOT automatically use bare-ASCII queries.
```
sampsyo, line 31:
```suggestion
Note: the ``bareasc`` command does *not* automatically use bare-ASCII queries.
```
Tiny thing, but because we're in rich text here, I generally tend to prefer italics over capitalization for emphasis.

docs/plugins/bareasc.rst
```diff
@@ -0,0 +1,68 @@
...
+Notes
+-----
+
+If the query string is all in lower case, the comparison ignores case as well as
+accents.
+
+The default ``bareasc`` prefix (``#``) is used as a comment character in some shells
+so may need to be protected (for example in quotes) when typed into the command line.
+
+The bare ASCII transformation is quite simple. It may not give the expected output
```
sampsyo, line 43:
```suggestion
The bare ASCII transliteration is quite simple. It may not give the expected output
```
This fancy word somewhat more precisely describes what unidecode does. :smiley:

docs/plugins/bareasc.rst
```diff
@@ -0,0 +1,68 @@
...
+To configure the plugin, make a ``bareasc:`` section in your configuration
+file. The only available option is:
+
+- **prefix**: The character used to designate bare-ASCII queries.
+  Default: ``#``, which may need to be escaped in some shells.
+
+Credits
+-------
+
+The hard work in this plugin is done in Sean Burke's Unidecode library.
```
sampsyo, line 66: Perhaps include a link to the PyPI page or GitHub repository for unidecode?

sampsyo (commented):
Looks awesome! Here are a few minor thoughts after a review.

docs/plugins/bareasc.rst
GrahamCobb: Good idea. Will add in next commit.

beetsplug/bareasc.py
GrahamCobb: Hmm. Actually I was confused, and never went back to ask for help.

I took the code from commands.py: list_func and list_items.  `list_func` has three args: `(lib, opts, arg)` -- I understand those, I think. It then calls `list_items` with `(lib, decargs(args), opts.album)` -- again I think I understand that. However, `list_items` is declared with a fourth optional argument `fmt=u''` which it passes to `format`. It seems that in this use case (handling the `beet list` command itself), fmt is not being specified so I wanted to make sure the call from the `bareasc` command gave identical results by explicitly calling `format` with second argument `''`.

I am not familiar enough with the python `format` function to know if setting the second arg to '' is the same as dropping it completely. If it is, I can just remove `fmt`. Or is there something else I should be doing?

jackwilsdon: If we're just calling `format` to invoke `__str__`, we should probably just do `str(item)` (not sure if this will cause any encoding issues though?).

GrahamCobb: Yes, that makes much more sense to me. And it seems to work fine (at least in Python3). I will push that change, along with a test case for it, and we can check if it works for Python2 as well.

GrahamCobb: Tests pass. Thanks for the help Adrian and Jack. I assume you have no problem with me marking this as resolved.

sampsyo: Awesome! Thanks to both of you. Of course; seems resolved to me. :smiley:

sampsyo (approved):
This is looking great! I think it's ready to merge any time. @GrahamCobb, I have added you to the beatbox organization (many thanks for your several recent contributions!!), so feel free to hit the green button yourself if you like. 🚀 ✨

GrahamCobb:
Wow, thanks @sampsyo! I am honoured to be able to contribute to such a professionally run and active community :trophy: You will have noticed that I am no Python programmer (C or assembler are my comfort zones :mage_man: !) so I will certainly continue to need review and comments. I am currently working on some suggestions for the web plugin which I will bring up for discussion shortly.

sampsyo:
Yay! Sounds fantastic.

Issue #1051, opened 2014-10-31
Problem with non-ascii characters when using the discogs plugin and oauth


fhchl:
Just found this bug: when trying to import with the awesome discogs plugin I get the following error:

```
beet -v import /Volumes/Speicher/Music/iTunes/iTunes\ Media/Music/Nikolai\ Lugansky/Rachmaninov_\ Préludes,\ Moments\ Musicaux
user configuration: /Users/Franz/.config/beets/config.yaml
data directory: /Users/Franz/.config/beets
Sending event: pluginload
library database: /Users/Franz/fun/data/musiclibrary.blb
library directory: /Users/Franz/fun/music
Sending event: library_opened
Sending event: import_task_start
Looking up: /Volumes/Speicher/Music/iTunes/iTunes Media/Music/Nikolai Lugansky/Rachmaninov_ Préludes, Moments Musicaux
Tagging Nikolai Lugansky - Rachmaninov: Préludes, Moments Musicaux
No album IDs found.
Search terms: Nikolai Lugansky - Rachmaninov: Préludes, Moments Musicaux
Album might be VA: False
Traceback (most recent call last):
  File "/Users/Franz/anaconda/bin/beet", line 9, in <module>
    load_entry_point('beets==1.3.9', 'console_scripts', 'beet')()
  File "/Users/Franz/anaconda/lib/python2.7/site-packages/beets-1.3.9-py2.7.egg/beets/ui/__init__.py", line 935, in main
    _raw_main(args)
  File "/Users/Franz/anaconda/lib/python2.7/site-packages/beets-1.3.9-py2.7.egg/beets/ui/__init__.py", line 925, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "/Users/Franz/anaconda/lib/python2.7/site-packages/beets-1.3.9-py2.7.egg/beets/ui/commands.py", line 872, in import_func
    import_files(lib, paths, query)
  File "/Users/Franz/anaconda/lib/python2.7/site-packages/beets-1.3.9-py2.7.egg/beets/ui/commands.py", line 844, in import_files
    session.run()
  File "/Users/Franz/anaconda/lib/python2.7/site-packages/beets-1.3.9-py2.7.egg/beets/importer.py", line 295, in run
    pl.run_parallel(QUEUE_SIZE)
  File "/Users/Franz/anaconda/lib/python2.7/site-packages/beets-1.3.9-py2.7.egg/beets/util/pipeline.py", line 299, in run
[...]

Issue #2387, opened 2017-01-13
discogs: sequence item 0: expected str instance, bytes found


bb010g:
### Problem

On Python 3, an import with discogs enabled fails.

Running this command in verbose (`-vv`) mode:

```sh
nixovo% beet -vv import opus
```

Led to this problem:

```
user configuration: /home/bb010g/.config/beets/config.yaml
data directory: /home/bb010g/.config/beets
plugin paths:
Sending event: pluginload
lyrics: Disabling google source: no API key configured.
library database: /home/bb010g/MusicLibrary.blb
library directory: /home/bb010g/Music
Sending event: library_opened
Sending event: import_begin
Sending event: import_task_created
Sending event: import_task_start
chroma: chroma: fingerprinted b'/data/arch/bb010g/Downloads/Cave Story/opus/XX - [Every track].opus'
chroma: no match found
Looking up: /data/arch/bb010g/Downloads/Cave Story/opus
Tagging Pixel - Cave Story
No album ID found.
Search terms: Pixel - Cave Story
Album might be VA: False
Sending event: albuminfo_received
Candidate: 開発室Pixel - Cave Story
Success. Distance: 0.32
Sending event: albuminfo_received
Candidate: 開発室Pixel - Cave Story
Success. Distance: 0.11
Sending event: albuminfo_received
Candidate: SUEMITSU & THE NODAME ORCHESTRA - Orchestra de Nodame Story
Success. Distance: 0.68
Sending event: albuminfo_received
Candidate: Various Artists - The Casablanca Records Story
Success. Distance: 0.76
Sending event: albuminfo_received
Candidate: The Idle Race - Back to the Story
Success. Distance: 0.65
chroma: acoustid album candidates: 0
Traceback (most recent call last):
  File "/usr/bin/beet", line 11, in <module>
    load_entry_point('beets==1.4.3', 'console_scripts', 'beet')()
  File "/usr/lib/python3.6/site-packages/beets/ui/__init__.py", line 1209, in main
    _raw_main(args)
  File "/usr/lib/python3.6/site-packages/beets/ui/__init__.py", line 1196, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "/usr/lib/python3.6/site-packages/beets/ui/commands.py", line 930, in import_func
    import_files(lib, paths, query)
  File "/usr/lib/python3.6/site-packages/beets/ui/commands.py", line 907, in import_files
    session.run()
[...]

PR #3885, opened 2021-03-17
Remove ascii encode step

nicksellen:
## Description

Fixes https://github.com/beetbox/beets/issues/1051 (and maybe a reoccurrence of https://github.com/beetbox/beets/issues/2387)

As described in [this comment](https://github.com/beetbox/beets/issues/1051#issuecomment-800584218) I encountered a `TypeError: sequence item 0: expected str instance, bytes found` error, and it seems the step to encode to ascii is no longer required. Removing it resolved this issue for me.

## To Do

- [ ] ~~Documentation. (If you've add a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)~~
- [ ] ~~Tests. (Encouraged but not strictly required.)~~

### DIFF
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -239,13 +239,10 @@ class DiscogsPlugin(BeetsPlugin):
         # cause a query to return no results, even if they match the artist or
         # album title. Use `re.UNICODE` flag to avoid stripping non-english
         # word characters.
-        # FIXME: Encode as ASCII to work around a bug:
-        # https://github.com/beetbox/beets/issues/1051
-        # When the library is fixed, we should encode as UTF-8.
-        query = re.sub(r'(?u)\W+', ' ', query).encode('ascii', "replace")
+        query = re.sub(r'(?u)\W+', ' ', query)
         # Strip medium information from query, Things like "CD1" and "disk 1"
         # can also negate an otherwise positive result.
-        query = re.sub(br'(?i)\b(CD|disc)\s*\d+', b'', query)
+        query = re.sub(r'(?i)\b(CD|disc)\s*\d+', '', query)
 
         self.request_start()
         try:


docs/changelog.rst
```diff
@@ -338,6 +338,8 @@ Fixes:
 * :doc:`/plugins/web`: DELETE and PATCH methods are disallowed by default.
   Set ``readonly: no`` web config option to enable them.
   :bug:`3870`
+* :doc:`/plugins/discogs`: Remove step to encode to ascii
```
sampsyo, line 341: Maybe something slightly more descriptive of the fix: "Avoid a crash when searching for music with non-ASCII metadata."?

sampsyo (approved):
Awesome; given the changes to the discogs-client library discussed in https://github.com/beetbox/beets/issues/1051#issuecomment-225280906, I think this is the right thing to do. I have one suggestion about the changelog, but otherwise I think this is good to go.

Many thanks once again for digging in and finding the problem!!

docs/changelog.rst
nicksellen: I don't think the crash was because the content of the query, but that it was being passed around as `bytes` datatype... 

... without having looked deeper into it I suspect the old discogs client still accepted that, but the new one (that was added a few weeks ago in https://github.com/beetbox/beets/commit/00252ab28ffc7d8e25cd5e420b6c086bfc178023) didn't?

Happy to change the text though to something else... just not sure what...

sampsyo: Got it! The new library would indeed explain it. Maybe we don't need a changelog at all, then, because this can essentially be folded into the change to use the new library.

nicksellen: Ok, sure! I'll remove it.

nicksellen: ... ok I just removed that last commit (force pushed it back one commit...).

sampsyo:
Perfect! Many thanks once again.

Issue #3880, opened 2021-03-13
mbsync: Double output


djl:
Recently the `mbsync` plugin has started to produce double output.
 ```
$ beet mbsync -p black sabbath vol 4
Black Sabbath - Vol 4 - Wheels of Confusion
  script:  -> Latn
  language:  -> eng
  script:  -> Latn
  language:  -> eng

...snip...

Black Sabbath - Vol 4 - Under the Sun
  script:  -> Latn
  language:  -> eng
  script:  -> Latn
  language:  -> eng
```

A quick `git bisect` reveals this was introduced in 72515448ad505ef0b73c0d0aa4aa424419b55755.

It looks like this could just be a cosmetic issue but I haven't had time to dive into the new code from #2988 to confirm yet.

wisp3rwind:
I'd guess that in

https://github.com/beetbox/beets/blob/c4347960eab78be2ad54f4e35b32928c47e287d1/beets/library.py#L620-L627

the keys should be round-tripped through a set, i.e. `return list(set(keys))`. Or maybe there's a way to deduplicate lists in-place?

sampsyo:
Ah, yes, that does look like the issue. Another simple way would be to just have `keys` be a set in the first place… i.e., something like `keys = set(super(...).keys(...)`. Perhaps the consumer doesn't care whether the return value is a set or a list, in which case it can just stay that way, or else we can do `return list(keys)` at the end.

wisp3rwind:
> Ah, yes, that does look like the issue. Another simple way would be to just have `keys` be a set in the first place… i.e., something like `keys = set(super(...).keys(...)`. Perhaps the consumer doesn't care whether the return value is a set or a list, in which case it can just stay that way, or else we can do `return list(keys)` at the end.

Yeah, I proposed wrapping in a `list` in order not to change the return type. This method very much resembles `dict.keys()`, so I'd say that consumers shouldn't rely on any properties of the result except for being an iterator. Whether that is really the case in our codebase, I don't know. We might just try returning a set, and see whether the tests succeed.

wisp3rwind:
So, locally on Python 3.9 at least, returning a `set` seems to work, i.e. the tests suceed. @FichteFoll does that look correct to you? I haven't followed the album field fallback PR, so I might be missing something.

FichteFoll:
As set definitely is the proper data structure for this task. I just didn't consider the keys to be iterated over when implementing fallback and thus the potential for duplicates.

[...]

PR #3886, opened 2021-03-18
slight performance optimization for ui.show_model_changes

wisp3rwind:
See https://github.com/beetbox/beets/issues/3880#issuecomment-801844630

```
I didn't actually benchmark this, but creating new FormattedItemMapping
views in each iteration of the loop appeared like a rather inefficient
way of doing things. In reality, this is probably completely irrelevant,
since one would usually not show huge numbers of changed Models at a
time.
```

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -666,10 +666,10 @@ def term_width():
 FLOAT_EPSILON = 0.01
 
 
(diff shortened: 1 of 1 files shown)


sampsyo (approved):
Great! Seems like a clear win to me (at the cost of a slight increase in complexity of the `_field_diff` function signature).

PR #3887, opened 2021-03-18
when iterating over item keys, do not return duplicates

### LINKED ISSUE #3880: mbsync: Double output

wisp3rwind:
... for keys that exist both as Item and Album fields

## Description

Fixes #3880

## To Do

- [ ] <del>Changelog. (Add an entry to `docs/changelog.rst` near the top of the document.)</del> Probably no need for this, the regression wasn't in a release.

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -623,7 +623,9 @@ class Item(LibModel):
         """
         keys = super(Item, self).keys(computed=computed)
         if with_album and self._cached_album:
-            keys += self._cached_album.keys(computed=computed)
+            keys = set(keys)
+            keys.update(self._cached_album.keys(computed=computed))
+            keys = list(keys)
         return keys
 
     def get(self, key, default=None, with_album=True):


sampsyo (approved):
Also very hard to find fault with! :smiley: Looks like exactly what we need.

PR #3888, opened 2021-03-18
replaygain: remove debugging leftovers

wisp3rwind:
Clean-up after #3817


sampsyo (approved):
Aha, yes, good catch. :sparkles:

PR #3889, opened 2021-03-18
Avoid deadlocks in Item.__repr__

wisp3rwind:
Shouldn't be a user-facing issue, but this took me a while to figure out while debugging.

I'm not sure about what exactly was going on, but it involved pytest trying to `repr()` some function arguments when logging an exception, and that then resulted in a deadlock.

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -617,6 +617,16 @@ class Item(LibModel):
                 return self._cached_album[key]
             raise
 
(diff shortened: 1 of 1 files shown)


sampsyo (approved):
Ah! Yes, that is a good point—without even knowing what was going on, I do heartily endorse the idea of making this not include the album stuff.

Issue #3872, opened 2021-03-09
ReplayGain plugin modifies source files in copy mode


qak:
### Problem

Importing music into the Beets library in copy mode with the ReplayGain plugin enabled causes the source files to be modified. The expected behaviour would be for Beets to only modify the files copied to the library and to keep the original files intact. For the purpose of this bug report I have reproduced the issue with the following config file:

```yaml
directory: ~/library
ui:
  color: no
plugins: replaygain
replaygain:
  backend: ffmpeg
```

Running command

```sh
$ beet -vv import "source/Aphex Twin - Flim (Laughing Man edit).flac"
```

results in the following output:

```
user configuration: /home/x/.config/beets/config.yaml
data directory: /home/x/.config/beets
plugin paths: 
Sending event: pluginload
library database: /home/x/.config/beets/library.db
library directory: /home/x/library
Sending event: library_opened
Sending event: import_begin
Sending event: import_task_created
Sending event: import_task_start
Looking up: /home/x/source/Aphex Twin - Flim (Laughing Man edit).flac
Tagging Aphex Twin - Flim (Laughing Man Edit)
No album ID found.
Search terms: Aphex Twin - Flim (Laughing Man Edit)
Album might be VA: False
Searching for MusicBrainz releases with: {'release': 'flim (laughing man edit)', 'artist': 'aphex twin', 'tracks': '1'}
Requesting MusicBrainz release 64006df9-ab6c-4b7d-a424-0dfd51ee8c74
primary MB release type: single
secondary MB release type(s): remix
Sending event: albuminfo_received
Candidate: Aphex Twin - Flim (Laughing Man edit) (64006df9-ab6c-4b7d-a424-0dfd51ee8c74)
Computing track assignment...
...done.
Success. Distance: 0.00
Requesting MusicBrainz release b26e7acb-1f99-4a28-a86c-e0ba67aa3b4e
primary MB release type: single
Sending event: albuminfo_received
Candidate: Aphex Twin - Analogue Bubblebath (Dumb Dan edit) (b26e7acb-1f99-4a28-a86c-e0ba67aa3b4e)
Computing track assignment...
...done.
Success. Distance: 0.47
Requesting MusicBrainz release 0ecef291-34ed-409f-9023-06c94644c5ce
primary MB release type: single
Sending event: albuminfo_received
Candidate: Aphex Twin - MARCHROMT30a Edit 2b 96 (0ecef291-34ed-409f-9023-06c94644c5ce)
[...]

PR #3890, opened 2021-03-18
replaygain: fix write/force flags on import

wisp3rwind:
Fixes https://github.com/beetbox/beets/issues/3872


sampsyo (approved):
Looks good to me—and nice that the code gets shorter too! (And more broadly, thanks for the flurry of very good work, @wisp3rwind.)

wisp3rwind:
In fact, this just restores this code to what it used to be :)

PR #3894, opened 2021-03-22
setup.py: add 3.9 classifier, add note about ffmpeg dep for replaygain

wisp3rwind:
Luckily, we _do_ support Python 3.9 :)


sampsyo:
Yay; excellent!

PR #3895, opened 2021-03-22
ci: install ffmpeg for replaygain tests; show skipped tests

wisp3rwind:
We lost the ability to show skipped tests when transitioning to gh actions. As it turns out, all of the replaygain tests were being skipped, so as a start, try to install at least one backend.

Let's see whether installing ffmpeg actually works, I don't really know what I'm doing regarding Github actions.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,6 +24,10 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install tox sphinx
 
+      - name: Install optional dependencies
+        run: |
+          sudo apt-get install ffmpeg  # For replaygain
+
       - name: Test with tox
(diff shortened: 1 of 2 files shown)


wisp3rwind:
@jef, @jtpavlock would that be the correct way to install extra dependencies with Github actions?

wisp3rwind:
Or should we rather be using https://github.com/FedericoCarboni/setup-ffmpeg ?

jef:
This seems totally fine to me. For something this small, especially if it can use pip to install, I'd rather inline it than use someone's action for something so minor.

sampsyo:
Great! Indeed; this looks like The Way to me.

wisp3rwind:
Nice, thanks for the swift feedback!

PR #3896, opened 2021-03-22
ci: also run on latest Python alpha, but allow failure

wisp3rwind:
Hopefully, this helps to spot breaking changes early. Unfortunately, the UI appears to be somewhat lacking: https://github.com/actions/toolkit/issues/399


wisp3rwind:
Rebased & merged.

PR #3899, opened 2021-03-23
Warn about dropping Python 3.5 support in the changelog

wisp3rwind:
While we're at it, we might announce the intent to drop 3.5 together with 2.7.


sampsyo:
Right; thanks!

PR #3900, opened 2021-03-24
badfiles: Optionally run checkers during import

### LINKED ISSUE #1633: badfiles: Check on import
How about running the badfiles-plugin on import, exactly like https://github.com/geigerzaehler/beets-check#usage-with-import (sans the checksum)?
Including the option to prevent import of a corrupted file.

Furthermore with at least mp3val an (automatic) fix could be possible?

Lanny:
## Description

Fixes #1633

Allows optionally running badfiles checkers during import. Issue this addresses had some suggestions for a general validation hook, although conversation was fairly old. Given the current set of hooks, this seemed like a relatively low-touch option that works well.

## To Do


sampsyo:
Thanks for getting this started; this looks really useful!

The one part of this change that we need to consider extremely carefully is the interaction with the core importer pipeline. This is a very fragile nest of logic and seemingly-benign changes can cause unexpected effects. For example, it is easy to create inconsistent behavior between album and item imports or to subtly break "incremental" mode when everything works fine when that option is disabled.

The current strategy is to move calls to `_get_choices` *before* `_summary_judgment` so that it is possible to *avoid calling the latter entirely*. An extra flag lets plugins modify the task state so that summary judgments can be skipped. That seems reasonable enough at first glance, but I also wonder whether it's possible to achieve a similar effect without rearranging the flow of logic here.

For example, maybe it's possible to make the hook behave very similar to `_summary_judgment`—that is, instead of mutating the task, it returns an action? That could make the core importer workflow treat the two routes to "early decisions" work the same. Would this make any sense, at a high level?

Lanny:
No problem, thanks for taking to the time look at it!

> The one part of this change that we need to consider extremely carefully is the interaction with the core importer pipeline

Fair point, and on closer examination the `skip_summary_judgement` flag doesn't work how I was hoping, it looks like further down the stack `choose_candidate` will skip prompting the user for input if there's a strong recommendation. This seems redundant but I'm guessing it exists to support item imports which don't have an equivalent `_summary_judgement` call before `choose_candidates`? Not sure trying to thread a "don't auto-apply strong recs" flag into there is something I'm confident doing given a lot of code seems to call through `choose_candidate`

>For example, maybe it's possible to make the hook behave very similar to _summary_judgment—that is, instead of mutating the task, it returns an action? That could make the core importer workflow treat the two routes to "early decisions" work the same. Would this make any sense, at a high level?

"the hook" here being `import_task_start` or `before_choose_candidate`? I don't think `import_task_start` would be suitable since it runs concurrently with the UI and wouldn't be able to present warnings or take input. `before_choose_candidate` already has a meaningful return value, and this re-ordering when it is called is still necessary since today `_summary_judgement` can prevent it from being called at all.

I do like the idea of returning an action from _some_ hook though. What would you think about a new hook, maybe `import_task_before_choice` to mirror `import_task_choice`? It could be called before `_summary_judgement` and have basically the same semantics as `_summary_judgement` like you suggested: returning None means continue as usual, otherwise an action means early return that action. It can run in the UI thread and present info or dialog as necessary (really just a "do you want to bail on this import" prompt in this case). This seems less likely to cause regression too, since it's one new hook that no plugin uses today.

sampsyo:
Yes! That design sounds very close to what I was hoping for, but more clearly articulated. I say let's give that a shot! Thanks for your careful consideration.

Lanny:
Awesome, pushed some more commits. I'm pretty happy with this approach, only thing that feels a bit awkward is the new `send_seq` pattern of passing the result of the last event handler to the next. I think it probably ends up being the most flexible but unfortunate that it pushes the work of coordination between multiple handlers onto the plugins. It was the best I could think of and ultimately I'm good with it if you are, but happy to do something else if you have a recommendation.

sampsyo:
Aha, I see! The `send_seq` thing is definitely a good idea overall for this specific case, and it's definitely nice that it foists the necessary complexity onto the plugins (in exchange for leaving the core importer pipeline alone :smiley:).

An even simpler option that might suffice, FWIW, would be to use a plain `send` and then make it an error to provide more than one action. That is, it would be disallowed to have two different plugins trying to make these determinations—all plugins must return `None` except for one. That's limiting, of course, but it's a sort of "YAGNI" approach: we can reevaluate and cook up the right solution if this ever becomes a problem.

Lanny:
That works for me! Not knowing what other use cases there might be for this hook makes it hard to determine what would be a good solution, so I like the idea of going with something simple and dealing with inter-plugin coordination in light of an actual need for it.

sampsyo (commented):
Looks awesome! I did a detailed review and found only one small suggestion to make. Otherwise, seems good to go!!

sampsyo (approved):
Excellent! Thanks for the very fast work!

PR #3901, opened 2021-04-01
Fix for Musixmatch multi-part lyrics

Ubiquitine:
## Description

Sometimes Musixmatch lyrics come in 2 parts.
Also, sometime non-existent lyrics return some content

### Lyrics: Improve Musixmatch backend
* Fix for lyrics that divided into 2 or more <div> elements.
* Ignore "fake" lyrics when lyrics not found but some content still being retrieved.

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -292,15 +292,11 @@ class MusiXmatch(Backend):
                               % url)
             return
         html_parts = html.split('<p class="mxm-lyrics__content')
-        # Sometimes lyrics come in 2 parts
-        if len(html_parts) > 1:
-            html_part1 = html_parts[-2]
-            lyrics_part1 = extract_text_between(html_part1, '>', '</p>')
-        else:
-            lyrics_part1 = ''
-        html_part2 = html_parts[-1]
-        lyrics_part2 = extract_text_between(html_part2, '>', '</p>')
-        lyrics = lyrics_part1 + '\n' + lyrics_part2
+        # Sometimes lyrics come in 2 or more parts
+        lyrics_parts = []
+        for html_part in html_parts:
+            lyrics_parts.append(extract_text_between(html_part, '>', '</p>'))
+        lyrics = '\n'.join(lyrics_parts)
         lyrics = lyrics.strip(',"').replace('\\n', '\n')
         # another odd case: sometimes only that string remains, for
         # missing songs. this seems to happen after being blocked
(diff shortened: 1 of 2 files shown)


beetsplug/lyrics.py
```diff
@@ -291,14 +291,25 @@ def fetch(self, artist, title):
...
+        html_parts = html.split('<p class="mxm-lyrics__content')
+        # Sometimes lyrics come in 2 parts
+        if len(html_parts) > 1:
+            html_part1 = html_parts[-2]
+            lyrics_part1 = extract_text_between(html_part1, '>', '</p>')
+        else:
+            lyrics_part1 = ''
+        html_part2 = html_parts[-1]
+        lyrics_part2 = extract_text_between(html_part2, '>', '</p>')
+        lyrics = lyrics_part1 + '\n' + lyrics_part2
```
sampsyo, line 303: Would it be simpler to do this with a loop over all the parts? Something roughly like this:

```py
for html_part in html_parts:
    lyrics_parts.append(extract_text_between(html_part1, '>', '</p>'))
lyrics = '\n'.join(lyrics_parts)
```

sampsyo (commented):
Looks perfectly reasonable; thanks!! Here's a suggestion for how to simplify the code. Also, could you please add a quick changelog entry?

beetsplug/lyrics.py
Ubiquitine: Thanks for the correction.
Indeed, this approach is more elegant. I've updated my branch to use the loop, as well as better non-existing lyrics detection.
I've tested it on couple songs I have, and it seems to work.

sampsyo:
Awesome; looks great! If you don't mind adding a quick changelog entry, I think this is good to go.

Ubiquitine:
I added changelog alongside the description. Or is there other place for this?

wisp3rwind:
> I added changelog alongside the description. Or is there other place for this?

I'm not seeing any changelog entry in this PR; did you actually push it? In any case, the changelog is at https://github.com/beetbox/beets/blob/master/docs/changelog.rst

Ubiquitine:
> > I added changelog alongside the description. Or is there other place for this?
> 
> I'm not seeing any changelog entry in this PR; did you actually push it? In any case, the changelog is at https://github.com/beetbox/beets/blob/master/docs/changelog.rst
Updated changelog.rst

sampsyo:
Looks great to me; thanks!!

PR #3907, opened 2021-04-16
ci: scope ci workflow

jef:
## Description

Changes `ci` workflow to only run when:

- A push to `master` happens
- A pull request gets opened and it's base is `master`
	- This will also kick of sequential `ci` runs if the branch is updated (synchronized: GitHub's terminology)

This makes for deduping the checks ran.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,5 +1,11 @@
 name: ci
-on: [push, pull_request]
+on:
+  push:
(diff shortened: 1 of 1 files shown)


sampsyo:
Ah, that's very useful! So this will avoid the doubled runs for "push" and "pull request" on every PR? This seems great!! Thanks!

jef:
Exactly!

PR #3923, opened 2021-05-03
Add HTTPS support/configuration option for Musicbrainz

jellekaufmann:
## Description

Enable HTTPS for musicbrainz.org API access by default. I noticed that beets was using HTTP with no way for the user to change this.

This patch does change default behavior since the use_https configuration option is set to yes by default. Setting this to no by default might be more appropriate.

## To Do

- [ ] Tests. I can't really think of a good way to write a test for this. I have verified that setting use_https works correctly by viewing a packet capture.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -46,6 +46,8 @@ Major new features:
 
 Other new things:
 
+* Enable HTTPS support to Musicbrainz by default and add configuration option
+  `use_https`.
 * :doc:`/plugins/mpdstats`: Add a new `strip_path` option to help build the
   right local path from MPD information.
 * :doc:`/plugins/convert`: Conversion can now parallelize conversion jobs on
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -47,7 +47,7 @@ Major new features:
 Other new things:
 
 * Enable HTTPS support to Musicbrainz by default and add configuration option
-  `use_https`.
+  `https` for custom servers.
 * :doc:`/plugins/mpdstats`: Add a new `strip_path` option to help build the
   right local path from MPD information.
 * :doc:`/plugins/convert`: Conversion can now parallelize conversion jobs on
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -46,7 +46,7 @@ Major new features:
 
 Other new things:
 
-* Enable HTTPS support to Musicbrainz by default and add configuration option
+* Enable HTTPS for MusicBrainz by default and add configuration option
   `https` for custom servers.
 * :doc:`/plugins/mpdstats`: Add a new `strip_path` option to help build the
   right local path from MPD information.
--- a/docs/reference/config.rst
+++ b/docs/reference/config.rst
@@ -701,8 +701,8 @@ under a ``musicbrainz:`` header, like so::
 
 The ``host`` key, of course, controls the Web server hostname (and port,
 optionally) that will be contacted by beets (default: musicbrainz.org).
-The ``https`` key makes the client use https instead of http, this setting applies
-only to custom servers. The official musicbrainz server always uses HTTPS (default: no).
+The ``https`` key makes the client use HTTPS instead of HTTP. This setting applies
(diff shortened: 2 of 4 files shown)


sampsyo:
Hi! Thanks for contributing this!

Just looking back at the history of this issue, we last addressed this in #2248 and https://github.com/beetbox/beets/commit/fbc0f322f6d7f781f5fa0a702dd39ca81ecf09d5, which added HTTPS for several web services, including MusicBrainz. We now attempt to automatically use HTTPS when possible for lookups, dictated by this condition:
https://github.com/beetbox/beets/blob/7254611a38221d4b56c179c2ef4c2a1536ff71d5/beets/autotag/mb.py#L34-L37

However, it looks like what you're saying is that this only affects the URLs we generate internally, and not the python-musicbrainz-ngs library. Which is really a shame.

Anyway, it seems like we should do something like this. A few thoughts:

- I'm worried about breaking backward compatibility for people with custom MB servers configured. Maybe I shouldn't be; that's perhaps a very niche use case. Moreover, I don't have a good idea about how to preserve the non-HTTPS default behavior for those users in an elegant way while making HTTPS the default for "normal" users, which is clearly the greater good.
- Should we be concerned about SNI support at this point? Maybe the default should depend on its presence?
- With apologies for the bikeshedding, would the simpler name `https` work, as opposed to `use_https`?

jellekaufmann:
Hi!

>     * I'm worried about breaking backward compatibility for people with custom MB servers configured. Maybe I shouldn't be; that's perhaps a very niche use case. Moreover, I don't have a good idea about how to preserve the non-HTTPS default behavior for those users in an elegant way while making HTTPS the default for "normal" users, which is clearly the greater good.
python-musicbrainz-ngs doesn't use HTTPS because Beets manually sets the musicbrainz hostname by calling musicbrainzngs.set_hostname . Only calling this function when a custom server is set is probably a simpler solution. I went this route to also make it possible to use HTTPS with a custom server, but that might be overkill.
>     * Should we be concerned about SNI support at this point? Maybe the default should depend on its presence?
I just checked and musicbrainz.org doesn't require SNI. So I guess we can enable HTTPS by default when no custom server is configured. But default to no HTTPS when a custom server is set?
>     * With apologies for the bikeshedding, would the simpler name `https` work, as opposed to `use_https`?
You're right >:), I simply copied the use_https argument name from python-musicbrainz-ngs without really thinking about it.

sampsyo:
OK, great, thanks for the thoughts! I'm convinced this is the way to go. :)

jellekaufmann:
Ok, great. I just implemented this. Let me know what you think.

docs/reference/config.rst
```diff
@@ -691,15 +691,18 @@ MusicBrainz Options
...
 
     musicbrainz:
         host: localhost:5000
+        https: no
         ratelimit: 100
 
 The ``host`` key, of course, controls the Web server hostname (and port,
 optionally) that will be contacted by beets (default: musicbrainz.org).
+The ``https`` key makes the client use https instead of http, this setting applies
+only to custom servers. The official musicbrainz server always uses HTTPS (default: no).
```
sampsyo, line 705:
```suggestion
The ``https`` key makes the client use HTTPS instead of HTTP. This setting applies
only to custom servers. The official MusicBrainz server always uses HTTPS. (Default: no.)
```

docs/changelog.rst
```diff
@@ -46,6 +46,8 @@ Major new features:
 
 Other new things:
 
+* Enable HTTPS support to Musicbrainz by default and add configuration option
```
sampsyo, line 49:
```suggestion
* Enable HTTPS for MusicBrainz by default and add configuration option
```

sampsyo (commented):
Wonderful! I just have a couple of docs comments, but other than this, I think it's good to go. Thanks for looking into it!!

jellekaufmann:
No problem. Thanks for maintaining this amazing piece of software!

sampsyo:
Looks great; thanks!

Issue #2800, opened 2018-01-29
include option to have set_field add fields to track vs album.


phunyguy:
### Problem

This is more of a feature request.  I would like set_fields option to be able to add things at the track level, rather than just the album level.  Without it, I have to run a second modify command after import to query for tracks where the field is NOT set to yes or no, and set it to yes if so.

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv import "some_dir"
```

Led to this problem:

```
(There isn't a problem in the output).
```

Here's a link to the music files that trigger the bug (if relevant): N/A

### Setup

* OS: Gentoo Linux
* Python version: 3.5.4
* beets version: 1.4.6
* Turning off plugins made problem go away (yes/no): n/a

My configuration (output of `beet config`) is:

```yaml
path_sep_replace: _
ignore:
- .AppleDouble
- .*
- '*~'
- System Volume Information
ftintitle:
    auto: yes
    drop: no
    format: (ft. {0})
fetchart:
    sources: coverart itunes google amazon wikipedia albumart filesystem
    google_key: REDACTED
    cautious: yes
    cover_names:
    - cover
    - front
    - art
    - album
    - folder
    maxwidth: 0
    fanarttv_key: REDACTED
    google_engine: 001442825323518660753:hrh5ch1gjzm
    auto: yes
    store_source: no
    minwidth: 0
    enforce_ratio: no

replace:
    '''': ''
    '[\\/]': _
    ^\.: _
    '[\x00-\x1f]': _
    '[<>"\?\*\|]': ''
    \.$: _
    \s+$: ''
    ^\s+: ''
    ^-: _
    ':': ','
embedart:
    auto: yes
    maxwidth: 1024
    remove_art_file: yes
    ifempty: no
    compare_threshold: 0

clutter:
- Thumbs.db
- .DS_Store
- '*.log'
- '*.txt'
- '*.cue'
- '*.pls'
- '*.m3u'
- '*.sfv'
- '*.nfo'

ui:
    terminal_width: 120
    color: yes
    colors:
        text_success: green
        text_warning: yellow
        text_error: red
        text_highlight: purple
        text_highlight_minor: lightgray
        action_default: cyan
        action: blue
art_filename: folder
ignore_hidden: no
library: ~/.config/beets/library.db
directory: ~/Music/vault
album_fields:
[...]

PR #3927, opened 2021-05-06
persist set_fields to media files

bertbesser:
## Description

Addresses #3925.

`set_fields` or `--set` write all fields through to the imported tracks, instead of setting them on the album only. This might be the most expected behavior, judging by [this](https://github.com/beetbox/beets/issues/2800) and [this](https://github.com/beetbox/beets/issues/2797) discussion.

Also, this PR wraps setting fields on album and tracks in lib transactions.

## To Do


beets/importer.py
```diff
@@ -572,18 +572,22 @@ def remove_duplicates(self, lib):
...
+            for field, view in config['import']['set_fields'].items():
+                value = view.get()
+                log.debug(u'Set field {1}={2} for {0}',
+                          displayable_path(self.paths),
+                          field,
+                          value)
+                self.album[field] = value
+                for item in self.imported_items():
+                    item[field] = value
+                    item.store()
```
bertbesser, line 589:
```suggestion
            for item in self.imported_items():
                item.store()
```

bertbesser, line 589: All `store()`-calls should be moved to a dedicated `for` loop outside `for field, view in ...`.

beets/importer.py
```diff
@@ -572,18 +572,22 @@ def remove_duplicates(self, lib):
...
-        self.album.store()
+        with lib.transaction():
+            for field, view in config['import']['set_fields'].items():
+                value = view.get()
+                log.debug(u'Set field {1}={2} for {0}',
+                          displayable_path(self.paths),
+                          field,
+                          value)
+                self.album[field] = value
+                for item in self.imported_items():
```
bertbesser, line 587: This loop occurs for each iteration of `for field, view in ...`. Therefore, `self.imported_items()` should probably be 'materialized' once before the outer loop. (iirc, `self.imported_items()` returns a list of 'unmaterialized' DB results.) I'm not sure, however, how this materialization should be implemented.

beets/importer.py
bertbesser: done

beets/importer.py
bertbesser: done

beets/importer.py
```diff
@@ -572,18 +572,24 @@ def remove_duplicates(self, lib):
...
-        self.album.store()
+        with lib.transaction():
+            items = self.imported_items()
+            for field, view in config['import']['set_fields'].items():
+                value = view.get()
+                log.debug(u'Set field {1}={2} for {0}',
+                          displayable_path(self.paths),
+                          field,
+                          value)
+                self.album[field] = value
```
bertbesser, line 587: The exact same loop as before, wrapped in a transaction plus amended with changing all items of the album.

beets/importer.py
```diff
@@ -946,18 +952,19 @@ def choose_match(self, session):
...
-        self.item.store()
+        with lib.transaction():
+            for field, view in config['import']['set_fields'].items():
+                value = view.get()
+                log.debug(u'Set field {1}={2} for {0}',
+                          displayable_path(self.paths),
+                          field,
+                          value)
+                self.item[field] = value
+            self.item.store()
```
bertbesser, line 967: The exact same loop as before, wrapped in a transaction.

sampsyo:
Nice! This seems pretty great to me. What do you think about checking in with folks on the Discourse or Discussions just to do a sanity check about whether this would break anything for anyone? Was anyone relying on using `--set` to _not_ assign values to individual tracks when in album mode? I can't imagine that there was, but maybe just posting about it and looking for feedback (moving on in the absence of comments in ~2-3 days) would be a useful sanity check.

bertbesser:
> Was anyone relying on using --set to not assign values to individual tracks when in album mode?

Yep, this was my fear, too. I also regard this very unlikely. Also, people seem to expect that `--set` will set each track's tags, like [here](https://discourse.beets.io/t/plugin-auto-set-comment-on-import/1762) or [here](https://discourse.beets.io/t/set-track-field-on-import/1235/5) or [here](https://discourse.beets.io/t/set-field-to-url/418/7).

But I'm also pro-sanity-check, so here we go: [discourse](https://discourse.beets.io/t/poll-whats-your-expected-behavior-set-fields-and-set/1764) and [github](https://github.com/beetbox/beets/discussions/3932).

bertbesser:
Hi,

from the feedback we've got I think our approach is right (one used additional config to circumvent the current behavior, the other thinks it's a bug). So imo we can integrate the change.

Cheers.

beets/importer.py
wisp3rwind: It think there's no need for the transaction here, the only interaction with the database here is in `item.store()` which already uses a transaction internally.

beets/importer.py
```diff
@@ -572,18 +572,24 @@ def remove_duplicates(self, lib):
...
+                log.debug(u'Set field {1}={2} for {0}',
+                          displayable_path(self.paths),
+                          field,
+                          value)
+                self.album[field] = value
+                for item in items:
+                    item[field] = value
+            for item in items:
+                item.store()
+            self.album.store()
```
wisp3rwind, line 592: It should be sufficient for the transaction to extend only over these three lines, setting the fields above shouldn't interact with the database.

test/test_importer.py
```diff
@@ -768,6 +774,10 @@ def test_set_fields(self):
             album.load()
             self.assertEqual(album.genre, genre)
             self.assertEqual(album.collection, collection)
+            for item in album.items():
+                self.assertEqual(item.genre, genre)
+                self.assertEqual(item.collection, collection)
+                self.assertEqual(item.comments, comments)
```
wisp3rwind, line 780: Have you checked that this test fails without the above changes to `set_fields`? I suspect that due to https://github.com/beetbox/beets/pull/2988, this part of the test was already successful before; it doesn't verify that the item tags are acutally set, though.

Maybe change to

```suggestion
            for item in album.items():
                self.assertEqual(item.get("genre", with_album=False), genre)
                self.assertEqual(item.get("collection", with_album=False), collection)
                self.assertEqual(item.get("comments", with_album=False), comments)
```
```
```

wisp3rwind (commented):
I've had a look at these changes: They generally look sound to me (just the implementation, I didn't follow the discussion on whether this is the right approach). I left a few comments: None of them need to block merging, but I think it'd be worthwhile to at least verify correctness of the test.

beets/importer.py
bertbesser: thx, fixed

beets/importer.py
bertbesser: thx. fixed.

test/test_importer.py
bertbesser: Yes, I checked. It failed without the change. In particular, it would not set the comments. Nevertheless I have adopted your suggestion for a stricter `get`.

sampsyo (approved):
This is looking great overall. I also think the discussion is trending toward convergence, with no one objecting to the new behavior. So whenever @bertbesser has enough confidence to move forward, I say we can hit that big green button.

wisp3rwind:
With you most recent changes (thanks for those!), a few style issues appear to have crept in @bertbesser, see the notes in the changes view: https://github.com/beetbox/beets/pull/3927/files

Would you mind addressing these, too?

bertbesser:
@wisp3rwind Sure. It was me, after all, to run the linter :-) Fixed (by hand).

Btw: What is the tox-way to reformat the code automatically?

bertbesser:
@sampsyo
> ... I say we can hit that big green button.
Let's do it :-)

wisp3rwind:
> @wisp3rwind Sure. It was me, after all, to run the linter :-) Fixed (by hand).

Thanks!

> Btw: What is the tox-way to reformat the code automatically?

There isn't as far as I know. You can run the linter with `tox -e py39-lint`, but there's no automatic formatter. There is some talk of moving to black, but that hasn't materialized yet: https://github.com/beetbox/beets/pull/3694

bertbesser:
> > @wisp3rwind Sure. It was me, after all, to run the linter :-) Fixed (by hand).
> 
> Thanks!
> 
> > Btw: What is the tox-way to reformat the code automatically?
> 
> There isn't as far as I know. You can run the linter with `tox -e py39-lint`, but there's no automatic formatter. There is some talk of moving to black, but that hasn't materialized yet: #3694

Yes, I used `tox -e py39-lint` and was searching for `tox -e py39-lint-fix` or similar. Thx for the pointer to the discussion on black.

Cheers.

PR #3934, opened 2021-05-08
Remove reference to rarfile version in link

arogl:
## Description

Remove reference to rarfile version in pypi link in `cli.rst`

## To Do


sampsyo:
Good call; thanks!

PR #3942, opened 2021-05-13
Add itunes_dateadded to metasync plugin

sandersantema:
## Description

This adds `itunes_dateadded` to the metasync plugin.

## To Do

- [x] Documentation: I believe this isn't required.
- [ ] Changelog: is "Add itunes_dateadded to metasync plugin" sufficient?
- [x] Tests: might need to be checked given that I've never really wrote tests before.

### DIFF
--- a/beetsplug/metasync/itunes.py
+++ b/beetsplug/metasync/itunes.py
@@ -69,6 +69,7 @@ class Itunes(MetaSource):
         'itunes_skipcount':   types.INTEGER,
         'itunes_lastplayed':  DateType(),
         'itunes_lastskipped': DateType(),
+        'itunes_dateadded':   DateType(),
     }
 
     def __init__(self, config, log):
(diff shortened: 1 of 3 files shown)


sandersantema:
I just remembered that the walrus operator was only introduced in python 3.8 and that beets support 3.6? Anyways I'll change it to not use the walrus operator.

wisp3rwind:
>     * [x]  Documentation: I believe this isn't required.

Well, in principle it would be nice to document the supported fields and their mapping to flexattrs in the `metasync` plugin documentation. However, the previously supported tags aren't documented either, so I guess it's not required to add this now for the new field.

>     * [ ]  Changelog: is "Add itunes_dateadded to metasync plugin" sufficient?

Maybe be a bit more explicit about what the plugin does exactly:
```
The ``metasync`` plugin now also fetches the ``Date Added`` field from iTunes databases and stores it in the``itunes_dateadded`` field.
```
Also, feel free to add your credits, cf. other changelog entries :)

>     * [x]  Tests: might need to be checked given that I've never really wrote tests before.

Otherwise, looks good to me :tada:

wisp3rwind:
As a side note, if there's an interest in adding further tags, it might be useful to not duplicate the code even further, but loop over the `item_types` dict (which would need to include iTunes'  name for the tag).

sandersantema:
That might be a good idea although I'd think that would require a configuration option because all those tags probably give quite a lot of clutter for people who don't want them. Furthermore a lot of these itunes tags are duplicated in beets itself while tags such as `Play count` can add to the play counts in beets and `Date added` doesn't exist in beets as far as I know i.e. as an immutable date at which the song was added for the first time to the library.

sandersantema:
Should be all done now!

wisp3rwind:
Great, thanks!

wisp3rwind:
> That might be a good idea although I'd think that would require a configuration option because all those tags probably give quite a lot of clutter for people who don't want them. Furthermore a lot of these itunes tags are duplicated in beets itself while tags such as `Play count` can add to the play counts in beets and `Date added` doesn't exist in beets as far as I know i.e. as an immutable date at which the song was added for the first time to the library. So a lot of those tags aren't really useful.

Yep, clearly this would need to be part of the configuration.

PR #3977, opened 2021-06-13
Update setup.py

arogl:
Remove the unsupported Python 3.5 from setup.py


wisp3rwind:
Now I'm not quite sure what our intention was: Support Python 3.5 until the next release (...), or drop it already now?

sampsyo:
TBH I also sort of lost track of whether we wanted to support 3.5 for the final time with this next release. Certainly not thereafter, though.

arogl:
Overtaken by #4018

PR #3982, opened 2021-06-16
Fix ifdef behavior with non-string fields (bug #3852)

### LINKED ISSUE #3852: %ifdef does not work on flexattr fields with int or float
### Problem
If I try `beet ls -a -f '$albumartist - (%if{$original_year,$original_year,$year}) - $album%ifdef{albumdisambig, (%title{$albumdisambig}),}%ifdef{albumrating, ($albumrating),}' year:2021` then beets will return the album rating for the albums that has the field defined. For albums without the rating, `$albumrating` is returned.

Example:
```
Artist1 - (2021) - Albumname1 (2.0)
Artist2 - (2021) - Albumname2 ($albumrating)
```

Further information on the forum:
https://discourse.beets.io/t/album-rating-as-a-flexible-attribute-in-5-increments/1687/3?u=pandatroubles

Comment from adrian:
_reading over the implementation of %ifdef, I can see now how it would be confused about types with a default value. That is, int or float fields will produce a '0' string for unset values through the formatter, which gets interpreted as non-missing. We clearly need better logic for this._

### Setup

* OS: Arch Linux
* beets version 1.5.0
* Python version 3.9.1
* plugins: edit, extrafiles, fetchart, importadded, importfeeds, info, inline, lastimport, mbcollection, mbsync, mpdstats, mpdupdate, originquery, scrub, smartplaylist, types, zero

adsultana:
## Description

Fixes #3852.

## To Do

- [x] Documentation: N/A, as it is a simple bug fix
- [x] Changelog: added
- [x] Tests: added. I wasn't sure whether the new test should belong with the template tests, or the types plugin tests. Since the bug only occurs when the non-string types from the types plugin are used, I thought the latter was the slightly more obvious choice.

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -1753,7 +1753,7 @@ class DefaultTemplateFunctions(object):
         :param falseval: The string if the condition is false
         :return: The string, based on condition
         """
-        if self.item.formatted().get(field):
+        if field in self.item:
             return trueval if trueval else self.item.formatted().get(field)
         else:
             return falseval
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -377,6 +377,9 @@ Fixes:
 * :doc`/reference/cli`: Remove reference to rarfile version in link
 * Fix :bug:`2873`. Duplicates can now generate checksums. Thanks user:`wisp3rwind`
   for the pointer to how to solve. Thanks to :user:`arogl`.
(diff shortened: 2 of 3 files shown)


sampsyo:
Nice; thank you!! I'm not sure why this was written up the old way, but this new way is clearly better. Nice work! :sparkles:

PR #3992, opened 2021-06-29
update irc link: freenode -> libera

wisp3rwind:
The website still pointed to freenode. Do we have a log bot around? (@Freso ?) For now, I've removed the link since `botbot.me` appears to be defunct.


sampsyo:
Thank you!! Too bad about BotBot.me; it was a nicely designed service while it lasted.

PR #3995, opened 2021-07-03
Crash-resilient Tekstowo lyrics source

wisp3rwind:
Follow up to https://github.com/beetbox/beets/pull/3994, see the comments there, and the message for the second commit in this PR:

```
The previous code had the potential to crash if (when?) Tekstowo changes
their website structure sufficiently.

The new code is rather ugly due to the explicit checks after each and
every function call. Unfortunately, the alternative would be to catch a
bunch of very generic Exceptions (AttributeError, ...), since there's no
such thing as a `BeautifulSoupNotFoundError`.
```

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -461,20 +461,27 @@ class Tekstowo(Backend):
         if not soup:
             return None
 
-        song_rows = soup.find("div", class_="content"). \
-            find("div", class_="card"). \
-            find_all("div", class_="box-przeboje")
+        content_div = soup.find("div", class_="content")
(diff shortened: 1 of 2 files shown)


sampsyo:
This looks perfect. It's one of those cases where "monadic" handling of Nones would actually be useful… but in any case, this seems like the right thing to do! Thanks!!!

wisp3rwind:
> It's one of those cases where "monadic" handling of Nones would actually be useful… 

That's a concept I've only been confronted with pretty recently, but I didn't recognize the pattern here, so thanks for pointing it out!

PR #4004, opened 2021-07-13
Fix subsonicupdate plugin

### LINKED ISSUE #4002: Subsonicupdate plugin broken in multiple paces
### Problem

Running the latest git version of beets causes the subsonicupdate plugin to fail with a generic "Error: 1".
The current code is broken in 2 places:

1. `def __get_version(self):`
Code is trying to retrieve the version in order to decide whether to authenticate with token or password: a much better choice than my original plan of assuming everyone would use an updated version and/or prefer token over password due to additional security.
In the current code, the REST API is called by passing only parameters 'c' and 'f': this API fails because it's not complete and the plugin stops processing.

2. `def start_scan(self):`
Assuming the authentication would go through, this call would fail anyway since the current code is splitting the API version in 3 values and passing them separately, e.g. `&v=1&v=61&v=1`, rather than `&v=1.61.1`

I have always been running an ancient version of beets with my original code for the plugin and I noticed this just today as I set up a new machine and installed beets form git repo through AUR.
It also seems I am the only user of this plugin given no other feedback has been filed so far, so I can understand if this fix doesn't get high priority.

### Setup

* OS: Arch Linux
* Python version: 3.9.6
* beets version: 1.5.0
* Turning off plugins made problem go away (yes/no):  disabling subsonicupdate plugin will prevent the error from happening

### Additional Info
Subsonic API reference: http://www.subsonic.org/pages/api.jsp

maffo999:
## Description

Fixes #4002.

Tested locally with Subsonic server 6.2 and library scans again on import

## To Do

- [?] Documentation. (If you've add a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)
- [?] Tests. (Encouraged but not strictly required.)

Unsure if documentation needs to change or not, the plugin now requires a new option (auth) under subsonic section in config.yaml


sampsyo:
Seems like it's headed in the right direction! Could you please remove the commented-out code? And since the configuration seems to have changed, it does seem like it would be great if the documentation reflected what users now need to do to use the plugin.

This may be a silly question, but is there a strong reason for removing the version query (and requiring the user to decide instead)? Is it just that it is annoying to get the version-check API call right?

maffo999:
> Could you please remove the commented-out code? And since the configuration seems to have changed, it does seem like it would be great if the documentation reflected what users now need to do to use the plugin.
will do and resubmit

> This may be a silly question, but is there a strong reason for removing the version query (and requiring the user to decide instead)? Is it just that it is annoying to get the version-check API call right?
In order to get the version from the server you need to make an API call (which requires authentication). Without any knowledge of what version of API is used by the server, we should default with plaintext password authentication which is not secure at all: if the Subsonic instance is hosted publicly, then login credentials could be easily found by a bad actor.

sampsyo:
> In order to get the version from the server you need to make an API call

That’s a very good reason; thanks for clarifying! 😊

maffo999:
not sure what's going on but I clearly cannot use git and this is getting messy, any chance I can provide you the fix in a different way? otherwise I guess I have to close this PR and open a new one

sampsyo:
TBH I don't mind messy histories much (we can always squash if they get out of hand!), but opening a new pull request is also free and 100% OK.

(I don't know about you, but sometimes if I get in over my head with git, I'll just give up and do `git diff > something.patch`, get back to a "clean" state, and then reapply `something.patch`. It's not pretty but it does get the changes you need out of a git quagmire.)

maffo999:
it's not really about the commit mess, but it seems Visual Studio added a whole bunch of unnecessary stuff in a ".vs" folder and I don't want to pollute your repo with it, e.g. commit 43b5f3f.
I'll close this one, clean up the repo on my side, and submit it again later today.

PR #4008, opened 2021-07-14
Fix for #4002: subsonicupdate plugin broken

### LINKED ISSUE #4002: Subsonicupdate plugin broken in multiple paces

maffo999:
## Description

Fixes #4002.
Tested locally on my Subsonic setup.

## To Do

### DIFF
--- a/beetsplug/subsonicupdate.py
+++ b/beetsplug/subsonicupdate.py
@@ -43,6 +43,7 @@ from beets.plugins import BeetsPlugin
 
 __author__ = 'https://github.com/maffo999'
 
+
 class SubsonicUpdate(BeetsPlugin):
     def __init__(self):
         super(SubsonicUpdate, self).__init__()
@@ -101,8 +102,8 @@ class SubsonicUpdate(BeetsPlugin):
         user = config['subsonic']['user'].as_str()
         auth = config['subsonic']['auth'].as_str()
         url = self.__format_url("startScan")
-        self._log.info(u'URL is {0}', url)
-        self._log.info(u'auth type is {0}', config['subsonic']['auth'])
+        self._log.debug(u'URL is {0}', url)
+        self._log.debug(u'auth type is {0}', config['subsonic']['auth'])
 
         if auth == "token":
             salt, token = self.__create_token()
--- a/beetsplug/subsonicupdate.py
+++ b/beetsplug/subsonicupdate.py
@@ -20,6 +20,14 @@ a "subsonic" section like the following:
         url: https://mydomain.com:443/subsonic
         user: username
         pass: password
+        auth: token
+For older Subsonic versions, token authentication
+is not supported, use password instead:
+    subsonic:
+        url: https://mydomain.com:443/subsonic
+        user: username
+        pass: password
+        auth: pass
 """
 from __future__ import division, absolute_import, print_function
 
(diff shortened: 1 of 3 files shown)


beetsplug/subsonicupdate.py
```diff
@@ -110,48 +97,30 @@ def __format_url(endpoint):
...
-            self._log.error(u'Error: {0}'.format(error))
-            return None
-
     def start_scan(self):
         user = config['subsonic']['user'].as_str()
-        url = self.__format_url("startScan.view")
+        auth = config['subsonic']['auth'].as_str()
+        url = self.__format_url("startScan")
+        self._log.info(u'URL is {0}', url)
+        self._log.info(u'auth type is {0}', config['subsonic']['auth'])
```
sampsyo, line 105: These should probably be `debug` logs, as opposed to `info`? (That will make them invisible by default, and I suspect most users don't need to see them.)

docs/changelog.rst
```diff
@@ -387,6 +387,7 @@ Fixes:
   all front images instead of blindly selecting the first one.
 * :doc:`/plugins/lyrics`: Removed the LyricWiki source (the site shut down on
   21/09/2020).
+* Fix :bug:`4002`. Subsonicupdate plugin is now functional again.
```
sampsyo, line 390: …and also has a new `auth` configuration parameter?

beetsplug/subsonicupdate.py
```diff
@@ -34,8 +42,6 @@
 from beets.plugins import BeetsPlugin
 
 __author__ = 'https://github.com/maffo999'
-AUTH_TOKEN_VERSION = (1, 12)
-
```
sampsyo, line 38: I think the style checker wants this blank line preserved.

sampsyo (changes requested):
Looks awesome; thanks!! Here are a couple of very minor suggestions.

maffo999:
All should be ok now but let me know if I missed anything else.

sampsyo (approved):
Perfect!! Thanks so much for sorting this out! 🚀 ✨

PR #4010, opened 2021-07-17
Add support for 256 colors

trapd00r:
## Description

Adds support for 256 colors in the ui. Color index is used as a name, like so:
```
ui:
  color: yes
  colors:
    text_success:         '40'
    text_warning:         '137'
    text_error:           '160'
    text_highlight:       '106'
    text_highlight_minor: '108'
    action_default:       '197'
    action:               '210'
```
256colors

## To Do

It would be nice to be able to specify (different) attributes on (different) COLOR_NAMES, like:
```yaml

text_success: 'bold 40'
text_warning: 'bold italic reverse 32'
```

but I'm unsure how to implement that right now. Feels like a waste to create more *_COLORS dictionaries, one for each attribute, like how DARK_COLORS and LIGHT_COLORS work right now for ANSI colors. Furthermore, attributes can be nested:  

```bash
printf "\033[38;5;196;1;3;4;7;48;5;220mfoobar\e[m" # perfectly legal
```

One _could_ allow the user to specify the relevant portion of the escape sequence themselves, like so:
```yaml
text_success: '40;1'      # bold 40
text_warning: '32;1;3;7' # bold italic reverse 32

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -534,9 +534,9 @@ def _colorize(color, text):
         # 38 == foreground color, but attributes as well as background
         # color can be specified in the same string if desired, i.e:
         # 38;5;196;1;3;4;5;7;48;5;220m
-        COLOR_ESCAPE = "\x1b[38;5"
+        COLOR_ESCAPE = "\x1b[38;5;"
         RESET_COLOR  = "\x1b[m"
-        escape = COLOR_ESCAPE + "%sm" (MORE_COLORS[color])
+        escape = COLOR_ESCAPE + MORE_COLORS[color] + "m"
     else:
         raise ValueError(u'no such color %s', color)
     return escape + text + RESET_COLOR
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -505,6 +505,13 @@ LIGHT_COLORS = {
     "cyan": 6,
     "white": 7
 }
+MORE_COLORS = { }
+# Generate color index mappings for 256 colors since there's no good way
+# to create meaningful names for them.
+for index in range(0, 256):
+    index = str(index)
+    MORE_COLORS[index] = index
+
 RESET_COLOR = COLOR_ESCAPE + "39;49;00m"
 
 # These abstract COLOR_NAMES are lazily mapped on to the actual color in COLORS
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: beets/ui/__init__.py [hunk 1]
```diff
@@ -505,6 +505,13 @@ def human_seconds_short(interval):
     "cyan": 6,
     "white": 7
 }
+MORE_COLORS = { }
+# Generate color index mappings for 256 colors since there's no good way
+# to create meaningful names for them.
+for index in range(0, 256):
+    index = str(index)
+    MORE_COLORS[index] = index
```
sampsyo, line 513: To avoid the need to create this mapping, could we instead just check whether the color given is a number? For example:

```py
try:
    index = int(color)
except ValueError:
    return
escape = ...  # something involving `index`
```

beets/ui/__init__.py
```diff
@@ -523,6 +530,13 @@ def _colorize(color, text):
         escape = COLOR_ESCAPE + "%im" % (DARK_COLORS[color] + 30)
     elif color in LIGHT_COLORS:
         escape = COLOR_ESCAPE + "%i;01m" % (LIGHT_COLORS[color] + 30)
+    elif color in MORE_COLORS:
+        # 38 == foreground color, but attributes as well as background
+        # color can be specified in the same string if desired, i.e:
+        # 38;5;196;1;3;4;5;7;48;5;220m
+        COLOR_ESCAPE = "\x1b[38;5;"
+        RESET_COLOR  = "\x1b[m"
+        escape = COLOR_ESCAPE + MORE_COLORS[color] + "m"
```
sampsyo, line 539: As with the other options in this file, maybe it's easier to use string formatting instead of concatenation:

```py
escape = "\x1b[38;5;{1}m".format(color)
```

beets/ui/__init__.py
```diff
@@ -523,6 +530,13 @@ def _colorize(color, text):
         escape = COLOR_ESCAPE + "%im" % (DARK_COLORS[color] + 30)
     elif color in LIGHT_COLORS:
         escape = COLOR_ESCAPE + "%i;01m" % (LIGHT_COLORS[color] + 30)
+    elif color in MORE_COLORS:
+        # 38 == foreground color, but attributes as well as background
+        # color can be specified in the same string if desired, i.e:
+        # 38;5;196;1;3;4;5;7;48;5;220m
+        COLOR_ESCAPE = "\x1b[38;5;"
+        RESET_COLOR  = "\x1b[m"
```
sampsyo, line 538: Shadowing a global variable with a local variable can make the flow of data pretty confusing… maybe we just need a new local variable that picks between the two strings?

sampsyo (changes requested):
Neat feature; thanks for contributing this! I have included a few low-level code comments. Could you please also add to the documentation to describe how to use the new feature?

beets/ui/__init__.py
trapd00r: Sure thing, however, that would make it harder to construct 'raw' strings for attributes, if we were to go down that route. Still don't know how to implement that though.

sampsyo: Indeed… it seems like that would need some creative parsing logic.

PR #4018, opened 2021-08-19
Require Python 3.6+

sampsyo:
This does the first-cut, initial work to remove support for Python 2, 3.4, and 3.5. As previewed with the v1.5.0 release, we now require a Python >=3.6. Things I have done:

* Remove the relevant Python versions from setup.py and such.
* Disable them in CI.
* Remove explicit `six.PY2` checks that led to different behavior on Python 2.
* Merge #3678, which drops AppVeyor for Windows CI in favor of GitHub Actions's Windows support. (We had trouble with this environment and Python 2 before.)

I think we should merge this sooner rather than later. There's of course so much we _could_ do during this expunging of old Python version, but I think we should leave at least these things for subsequent PRs:

* Fully removing six.
* Looking carefully for other dead code that may not be as obvious.
* Removing `u'…'` string literals and `__future__` imports.
* Someday, adding type annotations where we want them!


beets/library.py
```diff
@@ -37,13 +38,9 @@
...
 
 # To use the SQLite "blob" type, it doesn't suffice to provide a byte
-# string; SQLite treats that as encoded text. Wrapping it in a `buffer` or a
-# `memoryview`, depending on the Python version, tells it that we
-# actually mean non-text data.
-if six.PY2:
-    BLOB_TYPE = buffer  # noqa: F821
-else:
-    BLOB_TYPE = memoryview
+# string; SQLite treats that as encoded text. Wrapping it in a tells it
```
wisp3rwind, line 41:
```suggestion
# string; SQLite treats that as encoded text. Wrapping it in a `memoryview` tells it
```

beets/dbcore/types.py
```diff
@@ -104,7 +101,7 @@ def from_sql(self, sql_value):
         `sql_value` is either a `buffer`/`memoryview` or a `unicode` object`
```
wisp3rwind, line 101:
```suggestion
        `sql_value` is either a `memoryview` or a `unicode` object`
```

beets/plugins.py
```diff
@@ -129,14 +129,9 @@ def _set_log_level_and_params(self, base_log_level, func):
...
-            argspec = inspect.getargspec(func)
-            func_args = argspec.args
-            has_varkw = argspec.keywords is not None
-        else:
-            argspec = inspect.getfullargspec(func)
-            func_args = argspec.args
-            has_varkw = argspec.varkw is not None
+        argspec = inspect.getfullargspec(func)
+        func_args = argspec.args
+        has_varkw = argspec.varkw is not None
```
wisp3rwind, line 134: I think this auxiliary variables should be removed, and the right-hand side of these assignments substituted where they're used (which happens only once in each case).

beets/util/__init__.py
```diff
@@ -753,10 +753,7 @@ def as_string(value):
     """Convert a value to a Unicode object for matching with a query.
     None becomes the empty string. Bytestrings are silently decoded.
     """
-    if six.PY2:
-        buffer_types = buffer, memoryview  # noqa: F821
-    else:
-        buffer_types = memoryview
+    buffer_types = memoryview
```
wisp3rwind, line 756: Directly substitute `memoryview` in line 760 below.

beetsplug/convert.py
```diff
@@ -204,24 +203,25 @@ def encode(self, command, source, dest, pretend=False):
         if not quiet and not pretend:
             self._log.info(u'Encoding {0}', util.displayable_path(source))
 
-        if not six.PY2:
-            command = command.decode(arg_encoding(), 'surrogateescape')
```
wisp3rwind, line 208: This needs to stay (now unconditionally), right?

**EDIT:** I see you've added a decoding step in a second commit below, but now you're using `'utf8'` instead of `arg_encoding()`.

wisp3rwind (changes requested):
I've had a relatively quick look at these changes and added a few small style suggestions (which I'd say could also be fixed up later on, and don't need to be addressed before merging this).

The only thing that seems problematic is the slight change to argument decoding/encoding in the convert plugin, see the respective comment.

wisp3rwind:
> I think we should merge this sooner rather than later. There's of course so much we _could_ do during this expunging of old Python version, but I think we should leave at least these things for subsequent PRs:

:+1:

beetsplug/convert.py
sampsyo: Yep, this was a mistake! I think I made my life unintentionally difficult here by deleting this code when I should not have—I misread it as `if six.PY2`, i.e., the opposite sense, and then caused myself all manner of problems.

sampsyo:
Thanks; these were great suggestions!! All applied.

sampsyo:
I've ignored some more failing tests on Windows (we are up to 10 of these skips). I suppose we should add fixing these to that post-merge to-do list above. With that, GitHub actions is "all green." Unless there are objections, I think we should merge this and then disable AppVeyor. (I have already removed the AppVeyor configuration; all that remains is to deactivate the webhook so the check no longer runs.)

wisp3rwind (approved):

sampsyo:
Well, excellent! With that approval, I'll turn off AppVeyor and merge now. :tada: :sparkles: :rocket:

jef:
🚀 

Nice work y'all!

PR #4029, opened 2021-08-25
Add option to fetchart to store cover art as non-progressive.

khnsky:
## Description

Some DAPs like walkmans and ones using rockbox don't support progressive JPEGs.

This PR adds an option to fetchart plugin to store cover art without interlace, (and also functionality to util/artresizer to do so).

## To Do

Rockbox seems to not support PNGs either so it would be more future complete to allow to store all cover art as non-progressive JPEGs.

I tried to follow logic and conventions of the code base but as I'm not familiar with it there might be some issues, for exampled with util.syspath prefix=False which in the doctstring said to only use if you know what you are doing (I'm not) but I followed what other functions in artresizer did.

I'm not sure how correct the nomenclature of: progressive/non-progressive, interlace/deinterlace, baseline is.

The web backend seems per documentation to already serve images without interlace unless otherwise specified.


ArsenArsen:
This looks way better, thanks! And indeed, there was no activity.

khnsky:
@sampsyo I'm not sure if You want me to change the if-else thing and if so should I do ``progressive=not deinterlace`` or the kwargs trick. I don't know if I need to do something or just wait.

@ArsenArsen Thank You. I hope that I didn't step on Your toes or something, and that this PR works for Your use case, as it is a bit different in that it removes interlace from all formats, not only JPEG.

Finally for my use case I would probably also like to have cover art always in JPEG. If I were to write that feature, or change the deinterlace configuration option to take boolean or "jpeg" , so that only JPEG files would be converted to baseline, should I add to this PR or wait for this one to be merged and create another one.

This is my first PR so I don't really know how this works, thanks.

beets/util/artresizer.py
```diff
@@ -83,7 +86,12 @@ def pil_resize(maxwidth, path_in, path_out=None, quality=0, max_filesize=0):
             # Use PIL's default quality.
             quality = -1
 
-        im.save(util.py3_path(path_out), quality=quality)
```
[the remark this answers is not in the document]
ArsenArsen: When in doubt, check the sources: The JPEG plugin defaults this parameter to `False` when not set. https://github.com/python-pillow/Pillow/blob/a312b0e9d03e75d0d9c5184daf40c82fc6cef3f7/src/PIL/JpegImagePlugin.py#L737-L740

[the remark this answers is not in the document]
khnsky: removed the unnecessary if-else and replaced with specifying ``progressive=not deinterlace``

beets/util/artresizer.py
```diff
@@ -105,9 +113,12 @@ def pil_resize(maxwidth, path_in, path_out=None, quality=0, max_filesize=0):
                 if lower_qual < 10:
                     lower_qual = 10
                 # Use optimize flag to improve filesize decrease
-                im.save(
-                    util.py3_path(path_out), quality=lower_qual, optimize=True
```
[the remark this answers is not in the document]
khnsky: as above

sampsyo:
Nice! Looking pretty good, all told.

I have one last conceptual concern: since @ArsenArsen pointed out in https://github.com/beetbox/beets/pull/4029/files#r698801609 that `progressive=False` is the default in PIL, it occurs to me now that we might be changing beets's default behavior. That is, we were previously passing `progressive=False` (implicitly), but now, with the new default behavior, we'd be passing `progressive=True`. So rather than just failing to deinterlace progressive JPEGs, the default configuration will now be asking PIL to _create_ progressive JPEGs (where it previously didn't). Am I reading that right? If so, perhaps we should either change the default or just have `deinterlace` do something only when `im.save()` wouldn't otherwise be called (which should implicitly deinterlace files anyway)?

I think a separate config option to convert all art to JPEGs seems like a good idea but should probably live in a separate PR, FWIW.

ArsenArsen:
The default in beets should be to do no processing realistically, and just preserve the original values from the source image

ArsenArsen:
hey! I'd like to bump this PR, so that it can hopefully land in 1.5.1.

Thank you for your efforts.

sampsyo:
Thanks for the reminder! Everything is looking pretty good here, but can I just get some confirmation on the issue above? That is, as you say, the default *should* just be to preserve the original format of the image. But currently, the default configuration is `deinterlace: false`, and AFAICT, this passes `progressive=True` to `im.save` in a couple of places. It seems like this will create progressive JPEGs even when the originals were not progressive—is that right?

If I'm reading it right, then when doing `im.save`, it may be *impossible* to preserve the original format of the image. That is, we're converting the image anyway, so we have to pick the progressiveness attribute for the new file. Trying to detect what the old file was like and then generating a new file to match seems needlessly complex, especially because I suspect most people don't want progressive JPEGs in this day and age. So maybe a reasonable policy would be to leave these `im.save` calls unchanged, so the default `progressive=False` behavior stays and these converted files are _always_ non-progressive, regardless of the `deinterlace` option? That way, `deinterlace` really only controls the use of *extra* work that wouldn't otherwise be done to deinterlace the images.

ArsenArsen:
Right, that's what I mean, if there's no processing the file should be fully unaltered. Sorry for not being clear.

Depending on the most likely consumer the stored cover should default to baseline when there's no specified option but other processing has to happen. Progressive JPEG is beneficial on web for serving files (if you were ever on a slow connection you could notice a progressive JPEG get progressively more detailed).

Since the most likely consumer of beets-managed directories is local media players (MPD, IEMs, cmus, to name a couple), it's probably best to store as baseline due to the higher compatibility, rather than the previous value of the unconverted JPEG.

sampsyo:
Cool, I think we are on the same page about the goal—I suppose the thing I'm uncertain about is what the implementation in this PR currently does. Via the reasoning above, I guess I'm suspicious that images will now (in the default configuration) be made progressive, even when they were not originally?

khnsky:
I think You are right, it would make them progressive right now.

So if I understand correctly the best option would be:
- the resize functions are not affected by and don't take the `deinterlace` argument, instead default to non-progressive images. PIL and the web backend default to no interlace and from my limited testing so does ImageMagick, so the `progressive=False`/`-interlace none` only serves explicitness and could be omitted.
- the `deinterelace` option is to be used when no other processing is done, and when it is set to `yes` the deinterlace functions are used.

sampsyo:
Yep! I think that first option sounds great. It's simple and unlikely to cause problems.

khnsky:
Didn't mean this as either or :).

Updated so that the `deinterlace` argument is removed from the Artresizer resize functions and interlace is never used for the converted images, but if resizing/scaling is not performed the deinterlace function is called if `deinterlace: yes` is set (if it would otherwise determine the candidate to be exact but `detinerlace: yes` is specified it removes the interlace).

I left the explicit `progressive=False`/`-interlace none` but maybe those should be removed as they don't serve anything and are just noise.

beets/util/artresizer.py
```diff
@@ -77,7 +77,10 @@ def pil_resize(maxwidth, path_in, path_out=None, quality=0, max_filesize=0):
             # Use PIL's default quality.
             quality = -1
 
-        im.save(util.py3_path(path_out), quality=quality)
+        # progressive=False only affects JPEG writter and is the default
+        # leaving for explicitness
```
sampsyo, line 81: Maybe this comment could be a little more obvious for posterity—that is, something like:

> `progressive=False` only affects JPEGs and is the default, but we include it here for explicitness.

beets/util/artresizer.py
```diff
@@ -127,9 +129,11 @@ def im_resize(maxwidth, path_in, path_out=None, quality=0, max_filesize=0):
     # "-resize WIDTHx>" shrinks images with the width larger
     # than the given width while maintaining the aspect ratio
     # with regards to the height.
+    # no interlace seems to be default, specify for explicitness
```
sampsyo, line 132: Again, a full sentence here (with capital letter and period) might be a nice state to leave this in.

sampsyo (commented):
Looks great to me!! I just had a couple of comments on comments—they could be made a little more readable outside of this context. Other than that, I think this is good to go!

beets/util/artresizer.py
khnsky: fixed

beets/util/artresizer.py
khnsky: fixed

sampsyo (approved):
Yay; thank you for your excellent work on this!!! :tada: :sparkles:

ArsenArsen:
Thank you :D

PR #4036, opened 2021-08-30
permissions: set album art permissions

Duncaen:
## Description

This change makes the permissions plugin set the album art file permissions to the configurable file permissions whenever the `set_art` event is triggered.

## To Do


wisp3rwind:
This is looking good to me, what do you think @sampsyo ?

sampsyo (approved):
Thanks for the ping; yes, this all looks great! Nice work. Looks like it needs an update and then should be good to go.

Duncaen:
Thanks, rebased and fixed the change log conflict.

PR #4044, opened 2021-09-08
Get genres from both musicbrainz releases and release-groups.

aereaux:
## Description

This adds the votes from each.

This also changes the behavior to reset genre tags if the genre option is
enabled and no tags are received from musicbrainz.

This also sorts genres by the number of votes.

This is typically more useful because release-groups are usually better tagged, and now the %first function can be used to get the most popular genre tag.  Let me know if there are any other improvements I could make here.

## To Do

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -20,6 +20,7 @@ from __future__ import division, absolute_import, print_function
 import musicbrainzngs
 import re
 import traceback
+from collections import Counter
 from six.moves.urllib.parse import urljoin
 
 from beets import logging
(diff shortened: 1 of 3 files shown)


sampsyo:
Neat; this seems like a very good idea! I'm in favor.

It looks like there are a couple of style nitpicks. And could you also please add a note to the changelog? I'm also not sure whether there's somewhere in the documentation we should mention this; if we currently describe the MB genre stuff anywhere, maybe it needs a mention there.

aereaux:
Thanks for the quick response!  I added a changelog, docs tweak, and addressed the lint error

sampsyo:
Nice; this looks great! Thank you!!

PR #4045, opened 2021-09-08
Store secondary release types

### LINKED ISSUE #2200: Parse secondary album types from MusicBrainz

edgars-supe:
## Description

Closes #2200.

I've implemented some rudimentary processing for primary and secondary types, ~as well as a configurable field for use in paths~. I haven't written any docs or tests yet, because I'm not sure if it should be done this way (it's my first foray into beets's source code and Python).

### Solution
I've added `albumtypes` to the `Album` schema and `AlbumInfo`. Instead of simply logging the primary and secondary release types, I make a comma-separated string of them and store it in the new field. ~Then, I added a custom field (?) to `Album`, `atypes`, that outputs the album types according to the config.~ (removed, see [comment](https://github.com/beetbox/beets/pull/4045#pullrequestreview-749148643))

### Notes and questions
1) It seems to work _alright_ with queries. `beet ls albumtypes:remix` returns "On Remixes", as does `albumtypes:ep,remix`, but `remix,ep` doesn't.
2) Would it be possible and should I expose the album types (`albumtypes`) as a list (or set) instead of a plain `foo,bar` string?

## To Do


beets/autotag/hooks.py
```diff
@@ -88,6 +88,7 @@ def __init__(self, tracks, album=None, album_id=None, artist=None,
         self.tracks = tracks
         self.asin = asin
         self.albumtype = albumtype
+        self.albumtypes = albumtypes
```
sampsyo, line 91: There's actually no need to add `albumtypes` as an explicit parameter here; it can go into `**kwargs`. We have a bunch of fields listed here purely for legacy reasons; the `AlbumInfo` class is now "flexible" instead of baking in a fixed set of fields.

beets/autotag/mb.py
```diff
@@ -416,18 +416,16 @@ def album_info(release):
...
         rel_primarytype = release['release-group']['primary-type']
         if rel_primarytype:
-            log.debug('primary MB release type: ' + rel_primarytype.lower())
+            info.albumtypes = rel_primarytype.lower()
     if 'secondary-type-list' in release['release-group']:
         if release['release-group']['secondary-type-list']:
-            log.debug('secondary MB release type(s): ' + ', '.join(
-                [secondarytype.lower() for secondarytype in
-                    release['release-group']['secondary-type-list']]))
+            info.albumtypes += ',' + ','.join(
```
sampsyo, line 426: Using `+= ','` seems to run the risk of starting the string with a comma, if the album is missing a `primary-type`?

Maybe a simpler alternative would be to build up a list, `albumtypes`, that starts out empty and is then expanded under both `if` conditions, and then `join` it all together at the end.

sampsyo (commented):
Cool; thank you for getting this ball rolling! This seems like a promising direction.

I like the idea of adding a new `albumtypes` field. We don't currently have a lot of machinery for supporting list-valued fields, so a comma-separated string is likely the right way to go for now. We can consider more ambitious ways to store or query the data in the future… a simple hack could involve a comma-aware query type.

For now, I would prefer _not_ to add the more sophisticated `atype` field for formatting. We can explore designs like this in subsequent PRs, but I would love to merge the less-tricky part, just about fetching and storing the underlying data, before worrying about that. In particular, while fetching/storing the data is universal and belongs in "core" beets, the range of configuration and display styles are much more personal and might be best handled by a plugin…

beets/autotag/hooks.py
edgars-supe: Gotcha! I've removed what I had added here, seems to work still!

beets/autotag/mb.py
edgars-supe: Fair point! I had thought of that, but forgot about it just as quickly.

edgars-supe:
> For now, I would prefer not to add the more sophisticated `atype` field for formatting.

Gotcha, I've removed `atypes` from `Album`.

> the range of configuration and display styles are much more personal and might be best handled by a plugin…

Agreed. I've extracted `atypes` to a plugin (separate project). I'm not entirely sure what would be the best way to proceed with it, though. Do I just maintain it in my own repo? Or can I add it to the built-in plugins?

beets/library.py
```diff
@@ -494,6 +494,7 @@ class Item(LibModel):
         'mb_releasetrackid':    types.STRING,
         'trackdisambig':        types.STRING,
         'albumtype':            types.STRING,
+        'albumtypes':           types.STRING,
```
edgars-supe, line 497: I had to add this, because otherwise `beet import -A` would fail, because it sets values on an Item, iterating over `Album.item_keys`.

sampsyo: Seems right to me!

beets/autotag/mb.py
```diff
@@ -416,18 +416,17 @@ def album_info(release):
...
-            log.debug('primary MB release type: ' + rel_primarytype.lower())
+            albumtypes.append(rel_primarytype.lower())
     if 'secondary-type-list' in release['release-group']:
         if release['release-group']['secondary-type-list']:
-            log.debug('secondary MB release type(s): ' + ', '.join(
-                [secondarytype.lower() for secondarytype in
-                    release['release-group']['secondary-type-list']]))
+            for sec_type in release['release-group']['secondary-type-list']:
+                albumtypes.append(sec_type.lower())
+    info.albumtypes = ','.join(albumtypes)
```
sampsyo, line 429: It occurs to me somewhat belatedly, after reviewing #4044, that we've been using an existing convention for storing list-like data in string fields: namely, separating values with `; ` (semicolon-space) as opposed to just `,` (comma). I think this makes for somewhat more legible strings for humans and also is slightly less likely to get confused (because `,` seems more likely to occur within an individual element). Can we do that here?

sampsyo (commented):
Awesome; looking great! I've added one small suggestion. Could you also please add a changelog entry to `changelog.rst`?

In answer to your question about the plugin, it's completely OK to either maintain a plugin separately or open a PR to propose inclusion in beets. For this one, it's simple enough that it would be pretty easy to include with beets if you're interested! It does involve a teensy bit of extra work, such as writing a documentation page that matches the existing plugin docs, but that shouldn't be too hard.

beets/autotag/mb.py
edgars-supe: No problem; done!

edgars-supe:
I've added a line to `changelog.rst`. 

Okay, gotcha! I'll add it as a plugin to beets in a separate PR then. I think many people might find such a plugin quite useful out of the box; better than bothering with installing external plugins.

sampsyo:
Sounds great! Working on that in a separate PR will also be a good way to get other users' feedback about how the configuration should work, etc.

PR #4048, opened 2021-09-09
Add plugin for formatting albumtypes

edgars-supe:
## Description

Added a plugin to add a configurable album template field `$atypes` to format `albumtypes` of an Album (see #4045).

Should I rename the plugin to `atypes` since that's the field it adds, and to reduce possible confusion with the actual field `albumtypes`?

## To Do

### DIFF
--- a/beetsplug/albumtypes.py
+++ b/beetsplug/albumtypes.py
@@ -30,13 +30,14 @@ class AlbumTypesPlugin(BeetsPlugin):
         """Init AlbumTypesPlugin."""
         super(AlbumTypesPlugin, self).__init__()
         self.album_template_fields['atypes'] = self._atypes
-
-    def _atypes(self, item: Album):
         self.config.add({
             'types': [],
             'ignore_va': [],
             'brackets': '[]'
         })
+
+    def _atypes(self, item: Album):
+        """Returns a formatted string based on album's types."""
         types = self.config['types'].as_pairs()
         ignore_va = self.config['ignore_va'].as_str_seq()
         bracket = self.config['bracket'].as_str()
@@ -54,7 +55,7 @@ class AlbumTypesPlugin(BeetsPlugin):
         is_va = item.mb_albumartistid == VARIOUS_ARTISTS_ID
         for type in types:
             if type[0] in albumtypes and type[1]:
-                if not is_va or (not type[0] in ignore_va and is_va):
-                    res += bracket_l + type[1] + bracket_r
+                if not is_va or (type[0] not in ignore_va and is_va):
+                    res += f'{bracket_l}{type[1]}{bracket_r}'
 
         return res
--- a/beetsplug/albumtypes.py
+++ b/beetsplug/albumtypes.py
@@ -1,0 +1,60 @@
+# -*- coding: utf-8 -*-
+
+# This file is part of beets.
+# Copyright 2021, Edgars Supe.
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+
+"""Adds an album template field for formatted album types."""
+
+from __future__ import division, absolute_import, print_function
+
+from beets.autotag.mb import VARIOUS_ARTISTS_ID
+from beets.library import Album
+from beets.plugins import BeetsPlugin
+
+
+class AlbumTypesPlugin(BeetsPlugin):
+    """Adds an album template field for formatted album types."""
+
+    def __init__(self):
+        """Init AlbumTypesPlugin."""
+        super(AlbumTypesPlugin, self).__init__()
+        self.album_template_fields['atypes'] = self._atypes
+
+    def _atypes(self, item: Album):
+        self.config.add({
+            'types': [],
+            'ignore_va': [],
+            'brackets': '[]'
+        })
+        types = self.config['types'].as_pairs()
+        ignore_va = self.config['ignore_va'].as_str_seq()
+        bracket = self.config['bracket'].as_str()
+
+        # Assign a left and right bracket or leave blank if argument is empty.
+        if len(bracket) == 2:
+            bracket_l = bracket[0]
+            bracket_r = bracket[1]
+        else:
+            bracket_l = u''
+            bracket_r = u''
+
+        res = ''
+        albumtypes = item.albumtypes.split('; ')
+        is_va = item.mb_albumartistid == VARIOUS_ARTISTS_ID
+        for type in types:
+            if type[0] in albumtypes and type[1]:
+                if not is_va or (not type[0] in ignore_va and is_va):
+                    res += bracket_l + type[1] + bracket_r
+
+        return res
--- a/beetsplug/albumtypes.py
+++ b/beetsplug/albumtypes.py
@@ -31,9 +31,16 @@ class AlbumTypesPlugin(BeetsPlugin):
         super(AlbumTypesPlugin, self).__init__()
         self.album_template_fields['atypes'] = self._atypes
         self.config.add({
-            'types': [],
-            'ignore_va': [],
-            'brackets': '[]'
+            'types': [
+                ('ep', 'EP'),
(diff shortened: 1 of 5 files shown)


beetsplug/albumtypes.py
```diff
@@ -0,0 +1,60 @@
...
+
+class AlbumTypesPlugin(BeetsPlugin):
+    """Adds an album template field for formatted album types."""
+
+    def __init__(self):
+        """Init AlbumTypesPlugin."""
+        super(AlbumTypesPlugin, self).__init__()
+        self.album_template_fields['atypes'] = self._atypes
+
+    def _atypes(self, item: Album):
```
sampsyo, line 34: Perhaps add a docstring, like "generate a formatted string for an album's types"?

beetsplug/albumtypes.py
```diff
@@ -0,0 +1,60 @@
...
+        self.config.add({
+            'types': [],
+            'ignore_va': [],
+            'brackets': '[]'
+        })
```
sampsyo, line 39: This stuff should go in `__init__`; otherwise, it will try to set the defaults every time we use the `$atypes` field.

beetsplug/albumtypes.py
```diff
@@ -0,0 +1,60 @@
...
+        else:
+            bracket_l = u''
+            bracket_r = u''
+
+        res = ''
+        albumtypes = item.albumtypes.split('; ')
+        is_va = item.mb_albumartistid == VARIOUS_ARTISTS_ID
+        for type in types:
+            if type[0] in albumtypes and type[1]:
+                if not is_va or (not type[0] in ignore_va and is_va):
```
sampsyo, line 57: `not type[0] in ignore_va` is probably more clearly written `type[0] not in ignore_va`.

beetsplug/albumtypes.py
```diff
@@ -0,0 +1,60 @@
...
+                    res += bracket_l + type[1] + bracket_r
```
sampsyo, line 58: This might be a little easier to read as a [PEP 498](https://www.python.org/dev/peps/pep-0498/) format/template string? Like: `f'{bracket_l}{type[1]}{bracket_r}'`

docs/plugins/albumtypes.rst
```diff
@@ -0,0 +1,57 @@
+AlbumTypes Plugin
+=================
+
+The ``albumtypes`` plugin adds the ability to format and output album types,
+such as "Album", "EP", "Single", etc. List of available type can be found
+`here`_.
```
sampsyo, line 6: Maybe for this last link, I'd word it as: "For the list of available album types, see the MusicBrainz documentation." (with the last two words linked)

docs/plugins/albumtypes.rst
```diff
@@ -0,0 +1,57 @@
...
+
+To use the ``albumtypes`` plugin, enable it in your configuration
+(see :ref:`using-plugins`). Then, add ``$atypes`` to your path formats as
+desired.
```
sampsyo, line 10: "The plugin defines a new field `$atypes`, which you can use in your path formats or elsewhere."

docs/plugins/albumtypes.rst
```diff
@@ -0,0 +1,57 @@
...
+Configuration
+-------------
+
+To configure the plugin, make a ``albumtypes:`` section in your configuration
+file. The available options are:
+
+- **types**: An ordered list of album type to format mappings. The order of the
+  mappings determines their order in the output. If a mapping is missing or
+  blank, it will not be in the output.
+  Default: ``[]``.
```
sampsyo, line 23: Maybe just say "Default: None." (`[]` implies it's a list, not a mapping.)

But also… should the default be nothing? Does that mean that `$atypes` will always be empty by default? Perhaps this should come with a default mapping, like the one shown below, so the plugin is useful "out of the box"?

docs/plugins/albumtypes.rst
```diff
@@ -0,0 +1,57 @@
...
+- **ignore_va**: A list of types that should not be output for Various Artists
+  albums. Useful for not adding redundant information - various artist albums
+  are often compilations.
+  Default: ``[]``.
```
sampsyo, line 27: Again, perhaps the default here should be what's shown below—i.e., "compilation"? It seems like a sensible default for most people.

docs/plugins/albumtypes.rst
```diff
@@ -0,0 +1,57 @@
...
+            - remix: 'Remix'
+        ignore_va: compilation
+        bracket: '()'
+
+    paths:
+        default: $albumartist/($year)$atypes $album/...
+        albumtype:soundtrack Various Artists/$album ($year)$atypes)/...
+        comp: Various Artists/$album ($year)$atypes/...
+
+Example outputs::
```
sampsyo, line 51: "This configuration generates paths that look like this, for example::"

docs/plugins/albumtypes.rst
```diff
@@ -0,0 +1,57 @@
...
+
+    Aphex Twin/(1993)(EP)(Remix) On Remixes
+    Pink Flow/(1995)(Live) p·u·l·s·e
+    Various Artists/20th Century Lullabies (1999)
+    Various Artists/Ocean's Eleven (2001)(OST)
+
```
sampsyo, line 57: Probably no need for this extra line.

sampsyo (changes requested):
Wow; nice work! This looks awesome so far. I have a few suggestions for the code and some documentation improvements for your consideration.

beetsplug/albumtypes.py
edgars-supe: Good point! 

Would it not be better to extract the config values (next lines after this) in `__init__()` as well, so it's not done every time `_atypes`is called? I see both approaches in other plugins, so I suppose it's not a big concern, but converting config to those structs isn't exactly free.

docs/plugins/albumtypes.rst
edgars-supe: That's a good idea, I think. I've set the default values to the example config and adjusted the docs accordingly.

edgars-supe:
I've done the improvements you mentioned, added sane defaults, etc.

The comment about formatting output got me wondering, though: would it not be better to provide `prefix`, `separator` and `suffix` config values? That way with, say, `[`, `, `, `]` you could have `[EP, Remix]`, if that's what you prefer. Or `(`, `)(`, `)` to have the current output.

Also, still wondering if I should rename the plugin to `atypes`, because that's the template field it adds and to not confuse it with the actual field on `Album`. Thoughts?

beetsplug/albumtypes.py
sampsyo: That is less of a big deal, IMO, and it's fine to do it either way. The reason is that `config.add(...)` actually modifies `config`: it inserts a new "layer" with the default configuration values. So doing this every time actually adds multiple copies of that data. In contrast, `get` and `as_*` calls are read-only, so it doesn't matter as much where they appear.

docs/plugins/albumtypes.rst
```diff
@@ -0,0 +1,57 @@
...
+- **types**: An ordered list of album type to format mappings. The order of the
+  mappings determines their order in the output. If a mapping is missing or
+  blank, it will not be in the output.
+- **ignore_va**: A list of types that should not be output for Various Artists
+  albums. Useful for not adding redundant information - various artist albums
+  are often compilations.
+- **bracket**: Defines the brackets to enclose each album type in the output.
+
+The default configuration looks like this: ::
```
sampsyo, line 28:
```suggestion
The default configuration looks like this::
```

(I believe this is equivalent.)

sampsyo (approved):
Awesome! This looks great; I think it's basically good to go.

About the additional configuration values: I don't have a strong feeling either way! I definitely can envision people wanting a formatting like `[EP, Remix]`. But it's also totally fine to merge a minimal version of the plugin that fits your needs and to add more features later, "on demand," as people actually want them. (To philosophize for a moment, that strategy can be higher-overhead but lead to outcomes that more closely match what users want.) May I leave that decision up to you?

I usually have opinions about naming, but I am also finding myself without a strong feeling about the plugin name… I definitely see the argument for naming it the same as the field it creates, but the current name helps describe what it does. So again, I'm perfectly happy either way!

edgars-supe:
Either way is fine with me, so I'll just leave things as they are currently. I don't personally have a need to do `[EP, Remix]`. :D 
I'll perhaps work on that at some point, but if anyone expresses a strong need for such a feature, I can do it, no problem. 

I adjusted the docs based on your comment, so if there's nothing more, it's ready to be merged. :)

sampsyo:
Excellent! Nice work on this!! 🎉 🚀

PR #4060, opened 2021-09-14
Introduce atomic move and write of file

### LINKED ISSUE #3849: Option to copy files atomically
### Problem

When you importing a huge flac file to network disk via not so fast network (wifi for example) it requires a big time to complete.

If network disk has an agent that indexes files as soon as it was uploaded — for example default behaviour of Synology — it may start to index file *before* it is fully uploaded.

As result index will be broken.

The easy way to fix it is make inserting atomic by split it to two steps:
1) upload file to library with name like `$realName.$someTmpPostfix`
2) rename from `$realName.$someTmpPostfix` to `$realName`

catap:
The idea of this changes is simple: let move file to some temporary name inside distance folder, and after the file is already copy it renames to expected name.

When someone tries to save anything it also moves file to trigger OS level notification for change FS.

This commit also enforce that `beets.util.move` shouldn't be used to move directories as it described in comment.

Thus, this is fixed #3849

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -19,6 +19,7 @@ import sys
 import errno
 import locale
 import re
+import tempfile
 import shutil
 import fnmatch
 import functools
@@ -478,6 +479,11 @@ def move(path, dest, replace=False):
     instead, in which case metadata will *not* be preserved. Paths are
     translated to system paths.
     """
+    if os.path.isdir(path):
+        raise FilesystemError(u'source is directory', 'move', (path, dest))
+    if os.path.isdir(dest):
+        raise FilesystemError(u'destination is directory', 'move',
+                              (path, dest))
     if samefile(path, dest):
         return
     path = syspath(path)
@@ -487,15 +493,23 @@ def move(path, dest, replace=False):
 
     # First, try renaming the file.
     try:
-        os.rename(path, dest)
+        os.replace(path, dest)
     except OSError:
-        # Otherwise, copy and delete the original.
+        tmp = tempfile.mktemp(suffix='.beets',
+                              prefix=py3_path(b'.' + os.path.basename(dest)),
+                              dir=py3_path(os.path.dirname(dest)))
+        tmp = syspath(tmp)
         try:
-            shutil.copyfile(path, dest)
+            shutil.copyfile(path, tmp)
+            os.replace(tmp, dest)
+            tmp = None
             os.remove(path)
         except OSError as exc:
             raise FilesystemError(exc, 'move', (path, dest),
                                   traceback.format_exc())
+        finally:
+            if tmp is not None:
+                os.remove(tmp)
 
 
 def link(path, dest, replace=False):


catap:
@sampsyo any chances to review it?

sampsyo:
Hi, and thanks for your contribution! It's in my queue, but sometimes it can take me a while to have the bandwidth for a full review. Fortunately, the beets community is broad and doesn't just contain me, so other folks might also be able to take a look! You could consider putting out a broader call for reviewers in Discussions, etc., if you think that's warranted.

beets/library.py
```diff
@@ -760,6 +760,8 @@ def write(self, path=None, tags=None, id3v23=None):
         mediafile.update(item_tags)
         try:
             mediafile.save()
+            # Move file to trigger OS event if some processes subscribed to it
+            util.move(path, path, force=True)
```
wisp3rwind, line 764: Why is this necessary/should we do this? I guess this is forcing an update on file timestamps (is that actually guaranteed to happen, or is it just an implementation detail of the file system that this was tested on?). My first thought here would be that we should leave timestamp updated to the file system/mount option; and I would expect an mtime update to happen here, anyway.

beets/util/__init__.py
```diff
@@ -474,32 +475,45 @@ def copy(path, dest, replace=False):
                               traceback.format_exc())
 
 
-def move(path, dest, replace=False):
+def move(path, dest, replace=False, force=False):
```
wisp3rwind, line 478: See above: Do we need the force option?

beets/util/__init__.py
```diff
@@ -474,32 +475,45 @@ def copy(path, dest, replace=False):
...
         raise FilesystemError(u'file exists', 'rename', (path, dest))
 
     # First, try renaming the file.
     try:
-        os.rename(path, dest)
+        os.replace(path, dest)
     except OSError:
         # Otherwise, copy and delete the original.
+        if os.path.isdir(dest):
+            dest = os.path.join(dest, os.path.basename(path))
```
wisp3rwind, line 501: `dest` must not be a directory according to the doc string, so I'd add another assertion at function entry instead of this construction.

beets/util/__init__.py
```diff
@@ -474,32 +475,45 @@ def copy(path, dest, replace=False):
...
+        tmp = tempfile.mktemp(suffix='.beets',
+                              prefix=as_string(b'.' + os.path.basename(dest)),
+                              dir=as_string(os.path.dirname(dest)))
```
wisp3rwind, line 504: `as_string` is not supposed to be used for paths (see the docstring), in fact I'm pretty sure that it raises an exception for non-utf8 file names (maybe add a test for `beets.util.move()` with non-utf8 filenames?). If this conversion cannot be avoided, use `py3_path` (-> docstring). Maybe this would directly work with bytestrings if using a bytestring suffix `b'.beets'`?

beets/util/__init__.py
```diff
@@ -474,32 +475,45 @@ def copy(path, dest, replace=False):
...
+        tmp = syspath(tmp)
         try:
-            shutil.copyfile(path, dest)
+            shutil.copyfile(path, tmp)
+            os.replace(tmp, dest)
+            tmp = None
             os.remove(path)
```
wisp3rwind, line 510: I haven't thought this through, but this seems like a relatively long `try` clause: Does this always do the best thing to recover at each of the possible points of failure?

wisp3rwind (changes requested):
Not a full review, but some initial thoughts/requests/comments. Let's be certain that this is correct before merging; the fallout might otherwise be huge.

beets/util/__init__.py
catap: @wisp3rwind the root cause of this`as_string` is `tempfile.mktem`. It try to concate `prefix` and `suffix` with string which is obtained via `tempfile._get_candidate_names`.

Anyway, I love an idea about `py3_path`

beets/util/__init__.py
catap: I not sure that you suggest here :(

beets/util/__init__.py
catap: @wisp3rwind thanks!

beets/util/__init__.py
wisp3rwind: There's no direct suggestion, this is mostly a note to myself (and others!) to think about it in more depth. Maybe there's no issue, but at a first glance, this seemed like a critical section of the code to me.

beets/library.py
catap: Let me explain the case when `move` is required.

For example you're using wifi connection to work with your library which is stored as huge `flac` files, and you would like to insert a new and cool cover arts.

Music file can be big. Very big. dozen of megabytes. Cover arts with nice resolution also might be a few megs.

So, writing it via wifi isn't something fast. It requires time. Some time.

From another hand you may use smart NAS which indexes your music collection when it changed.

I do have such setup with Synology as NAS, and it starts to index a file when it appears. If file appears not fast enough => it corrupt it's index.

This move triggered forced re-index of file and `move` on the same FS is atomic, and it fires all listener who would like to know when file is updated *after* it is moved to distination.

beets/util/__init__.py
catap: @wisp3rwind split it into a few try blocks increases code dramatically. I prefer to keep it simple.

Do you agree?

beets/library.py
wisp3rwind: Ok, that's the issue that you reported elsewhere. I don't like the solution to be honest, it seems like a crude workaround that most people won't need, and it doesn't even prevent the intermediate broken index. A possible solution that I just thought of would be to use the [hook plugin](https://beets.readthedocs.io/en/stable/plugins/hook.html) to `touch` files in the `after_write` hook.

catap: @wisp3rwind touch isn't enough. File should be moved away ;)

wisp3rwind: Alternatively, it might be possible to implement proper atomic behaviour in `mediafile`.

catap: @wisp3rwind but it will be the same hack, isn't it?

wisp3rwind: > @wisp3rwind touch isn't enough. File should be moved away ;)

Well, you could do that in the hook, too. And there's still the issue that I don't think there's any guarantee that `os.replace(path, path)` does anything at all. The python 3.9 docs don't state this. In fact, if this is using `rename(3)` under the hood, POSIX even specifies
>If the old argument and the new  ar‐
       gument resolve to either the same existing directory entry or different directory entries for the same
       existing file, rename() shall return successfully and perform no other action.

wisp3rwind: > @wisp3rwind but it will be the same hack, isn't it?

No, the idea would be that `mediafile` created a new file (with a temporary suffix `.beets`) which would be renamed when it is fully written.

catap: @wisp3rwind I agree with your point to move it inside `mediafile`

catap: or inside `mutagen`

catap: and it has the issue: https://github.com/quodlibet/mutagen/issues/241 :)

catap:
@wisp3rwind / @sampsyo ping?

catap:
@sampsyo ping!

wisp3rwind:
It's not been forgotten! I'll try to have another look soon.

wisp3rwind (changes requested):
Looks good to me :tada: Sorry for the silence here, I've recently had much less time (and motivation) to spend on beets than before.

See the inline comment for a small typo in the log messages. Would you mind adding a changelog entry?

catap:
@wisp3rwind I've incorporated your changes and rebased to last master.

wisp3rwind:
> @wisp3rwind I've incorporated your changes and rebased to last master.

Great! A small style issue (line to long) has crept in in this last change, though.

catap:
@wisp3rwind everything is green now!

wisp3rwind:
Great! I've added a changelog entry for this in 5578d0713b3a9b60450ce3fb11c8e4a008001f7e

PR #4067, opened 2021-09-18
Add additional asciify documentation

### LINKED ISSUE #929: More 'asciify' documentation
[The documentation](http://beets.readthedocs.org/en/v1.3.7/reference/config.html#asciify-paths) for the `asciify_paths` option says

> Convert all non-ASCII characters in paths to ASCII equivalents. For example, if your path template for singletons is singletons/$title and the title of a track is “Café”, then the track will be saved as singletons/Cafe.mp3. 

This is clear enough for a Latin script (one would expect 'é' to be converted to 'e'), but it's unclear what this command would do for anything written in entirely different scripts or writing systems like Japanese kanji. (It says 'all' - does that mean they would be deleted since they have no ASCII equivalents?) I have no idea what it might do, and am too terrified to let this option anywhere near my files to figure it out empirically, so more documentation would be helpful for deciding whether to use this option.

emiham:
## Description

Fixes #929.

This was proposed but never dealt with, so after going through some confusion around this myself I figured it's about time.

## To Do

### DIFF
--- a/docs/reference/config.rst
+++ b/docs/reference/config.rst
@@ -165,6 +165,10 @@ take place before applying the :ref:`replace` configuration and are roughly
 equivalent to wrapping all your path templates in the ``%asciify{}``
 :ref:`template function <template-functions>`.
 
+This uses the `unidecode module`_ which is language agnostic, so some 
+characters may be transliterated from a different language than expected. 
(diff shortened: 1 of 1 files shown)


sampsyo:
Awesome; this will be helpful, I think! Thanks for adding it.

PR #4069, opened 2021-09-21
Added notes about MacPorts

catap:


catap:
@sampsyo I feel that you defently would like to adjust wording :)

docs/guides/main.rst
```diff
@@ -12,9 +12,11 @@ Installing
...
 Beets works on Python 3.6 or later.
 
-* **macOS** 11 (Big Sur) includes Python 3.8 out of the box.
-  You can opt for a more recent Python installing it via `Homebrew`_:
-  ``brew install python3``
+* On **macOS**, there's a `MacPorts`_ port. Just run ``port install beets`` or
+  ``port install beets-full`` to install it with all 3rd party plugins. macOS**
+  11 (Big Sur) includes Python 3.8 out of the box and you may install it via
+  ``pip``. You can opt for a more recent Python installing it via `Homebrew`_:
+  ``brew install python3`` for older versions.
```
sampsyo, line 19: Seems good, but let's keep the Homebrew thing first… that is, list the text beginning "11 (BigSur) includes…" first. Then, the next sentence can say, like, "There's also a MacPorts port. Just run…"

Also, I'm not entirely sure what "all 3rd party plugins" means… there are a *lot* of third-party plugins out there! Do you mean all dependencies for the included (first-party) plugins, perhaps?

sampsyo (commented):
Thanks!! Indeed; here are a couple of comments.

docs/guides/main.rst
catap: @sampsyo I mean that almost all plugins from [Other Plugins](https://beets.readthedocs.io/en/stable/plugins/index.html#other-plugins) are included as dedicated port and `beets-full` installs all of them. With all dependencies. Technically speaking I've [added 25 ports](https://ports.macports.org/search/?q=beets&name=on) which include 22 3rd party plugins,`beets`, `beets-devel` (the master branch)  and `beets-full` (install everything).

sampsyo: Incredible! Well, maybe I'd just say "lots of third-party plugins" or something to sidestep the potential confusion around claiming "all". 😃

sampsyo:
Tweaked it a bit more for concision, but looks good! Thanks again!

PR #4073, opened 2021-09-25
avoid modification of tags in case of autotag is no

### LINKED ISSUE #4072: comp vs. albumartist circle dependency with 'autotag: no' | 'autotag: no' but modification of tags
# Expected result
no modification of tags with 'autotag: no'

# Observed result

* import behaviour 1
	if comp is TRUE and albumartist is set
	beet import modifies 'comp: True -> False'
	
* import behaviour 2
	if comp is TRUE and albumartist is not set 
	beet import modifies 'albumartist: -> Various Artists'

Hint:
It seems that a comp with albumartist is not a comp,
but if no albumartist is set, a albumartist is set automatically which makes it indirectly with next import to a not comp.

Inconsistent behaviour.

# Version
v1.5.0

# Workaround
As workaround I commented out in file beets/importer.py function align_album_level_fields the part in if action.ASIS-plugins= version

# Configuration (output of `beet config`) is:

```yaml
directory: ~/MusicNew
library: ~/.cache/beets/musiclibrary.db

import:
    move: no
    copy: yes
    write: no
    autotag: no
    duplicate_action: ask

clutter:

plugins: inline fetchart duplicates

replace:
    \.: ''
item_fields:
    disc_and_track: u'%02i.%02i' % (disc, track) if disctotal > 1 else u'%02i' % (track)

paths:
    default: $albumartist/$album%aunique{} ($year)/$disc_and_track $title
    comp: $genre/$album%aunique{} ($year)/$track $title
    singleton: Non-Album/$artist/$title
    albumtype:soundtrack: Soundtracks/$album%aunique{} ($year)/$track $title
    albumtype:Audiobook: Audiobook/$album%aunique{} ($year)/$track $title
fetchart:
    auto: yes
    cover_names: cover front back art
    sources: filesystem
    minwidth: 0
    maxwidth: 0
    quality: 0
    max_filesize: 0
    enforce_ratio: no
    cautious: no
    google_key: REDACTED
    google_engine: xxx
    fanarttv_key: REDACTED
    lastfm_key: REDACTED
    store_source: no
    high_resolution: no
duplicates:
    album: no
    checksum: ''
    copy: ''
    count: no
    delete: no
    format: ''
    full: no
    keys: []
    merge: no
    move: ''
    path: no
    tiebreak: {}
    strict: no
    tag: ''
pathfields: {}
album_fields: {}
```

friedrich-daniel:
updated behaviour so that function align_album_level_fields is only executed in case that configuration parameter import.autotag is set to true
 
## Description

Fixes #4072

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -767,7 +767,8 @@ class ImportTask(BaseImportTask):
     def add(self, lib):
         """Add the items as an album to the library and remove replaced items.
         """
-        self.align_album_level_fields()
+        if config['import']['autotag'] is True:
+            self.align_album_level_fields()
         with lib.transaction():
             self.record_replaced(lib)
             self.remove_replaced(lib)
--- a/test/test_importer.py
+++ b/test/test_importer.py
@@ -259,7 +259,8 @@ class NonAutotaggedImportTest(_common.TestCase, ImportHelper):
         self.importer.run()
         albums = self.lib.albums()
         self.assertEqual(len(albums), 1)
-        self.assertEqual(albums[0].albumartist, u'Tag Artist')
+        if config['import']['autotag'] is True:
+            self.assertEqual(albums[0].albumartist, u'Tag Artist')
 
     def test_import_copy_arrives(self):
         self.importer.run()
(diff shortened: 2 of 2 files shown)


sampsyo:
Seems like a good first step! I think it's important to think through exactly how this would change the behavior of as-is imports. Is there any chance you could try to write up a brief description of exactly what will be different—and try to think carefully about what the potential downsides are? (That is, what hypothetical users might be _sad_ that this change occurred?)

sampsyo:
(I also notice that the tests are failing.)

Issue #4027, opened 2021-08-25
lyrics: Crash when BeautifulSoup is not installed


janpeeters70:
### Problem
The Lyrics plugin throws an error/crashes when ran without BeautifulSoup installed.

### Terminal output
_I currently can't run a verbose command because Beets is dragging itself through finding lyrics for many many albums._

The following error is displayed in Terminal when running:
```
$ beets lyrics artist song
/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/beets/util/confit.py:21: UserWarning: beets.util.confit is deprecated; use confuse instead
  warnings.warn("beets.util.confit is deprecated; use confuse instead")
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.7/bin/beet", line 11, in <module>
    load_entry_point('beets==1.5.0', 'console_scripts', 'beet')()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/beets/ui/__init__.py", line 1291, in main
    _raw_main(args)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/beets/ui/__init__.py", line 1274, in _raw_main
    subcommands, plugins, lib = _setup(options, lib)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/beets/ui/__init__.py", line 1153, in _setup
    plugins = _load_plugins(options, config)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/beets/ui/__init__.py", line 1139, in _load_plugins
    plugins.send("pluginload")
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/beets/plugins.py", line 497, in send
    for handler in event_handlers()[event]:
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/beets/plugins.py", line 480, in event_handlers
    for plugin in find_plugins():
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/beets/plugins.py", line 315, in find_plugins
    _instances[cls] = cls()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/beetsplug/lyrics.py", line 736, in __init__
    sources = self.sanitize_bs_sources(sources)
[...]

PR #4084, opened 2021-10-01
FAQ: Add another option for installing in editable mode

ctrueden:
It's another installation option that may be useful.

See also:
https://github.com/beetbox/beets/issues/4027#issuecomment-932232208


wisp3rwind:
LGTM, thanks!

PR #4086, opened 2021-10-02
feat(unimported): support excluding subdirectories

### LINKED ISSUE #4025: Lyrics plugin: Ignore folders or specify folder to run on
I use a hard drive as my Beets library 'folder'. 
Because of its size I also store some other non-imported music folders on that drive.
I ran into the situation that running 'beets unimported' showed me all the files in those unimported folders. 
It's logical that the plugin scans those too but a more specific scan would be great.
I could circumvent this by placing all Beets folders in another folder instead of the root of the drive but that would make for a deeper hierarchy which I wouldn't like.

### Proposed solution

Add extra options for the command line
`beets unimported /specific_folder`
or in config.yaml
```
lyrics:
    ignore_folders: non-imported-files
```

jaydesl:
## Support excluding specific subdirectories in _unimported_ plugin

Also refactored a bit because the comprehensions were a bit tough on readability.
Fixes #4025 

## To Do


docs/changelog.rst
```diff
@@ -33,6 +33,7 @@ Major new features:
 Other new things:
 
 * Permissions plugin now sets cover art permissions to the file permissions.
+* Unimported plugin now supports excluding specific subdirectories in library.
```
sampsyo, line 36: I know the previous bullet doesn't do it, but the way we usually refer to changes in plugins is like this:

```rest
* :doc:`/plugins/unimported`: Add a new configuration option to exclude subdirectories in the library.
```

docs/plugins/unimported.rst
```diff
@@ -9,9 +9,10 @@ Command Line Usage
...
-exclude file extensions using the configuration file::
+exclude file extensions or entire subdirectories using the configuration file::
 
     unimported:
         ignore_extensions: jpg png
+        ignore_subdirectories: NonMusic data temp
 
-The default configuration list all unimported files, ignoring no extensions.
\ No newline at end of file
+The default configuration list all unimported files, ignoring no extensions.
```
sampsyo, line 18:
```suggestion
The default configuration lists all unimported files, ignoring no extensions
or subdirectories.
```

beetsplug/unimported.py
```diff
@@ -40,11 +40,19 @@ def commands(self):
         def print_unimported(lib, opts, args):
             ignore_exts = [('.' + x).encode() for x
                            in self.config['ignore_extensions'].as_str_seq()]
-            in_folder = {
-                os.path.join(r, file) for r, d, f in os.walk(lib.directory)
-                for file in f if not any(
-                    [file.endswith(extension) for extension in
-                     ignore_exts])}
+            ignore_dirs = [lib.directory + ('/' + x.strip('/')).encode() for x
```
sampsyo, line 43: Using `os.path.join` would be a bit safer (and easier to read) than the manual string manipulation. Similarly, `os.path.normpath` can do the stripping of `/` as well as other normalization that might be desirable in this situation.

beetsplug/unimported.py
```diff
@@ -40,11 +40,19 @@ def commands(self):
...
+            in_folder = set()
+            for r, d, f in os.walk(lib.directory):
+                if r in ignore_dirs:
+                    continue
+                files = {
+                    os.path.join(r, file)
+                    for file in f if not 
+                    any([file.endswith(ext) for ext in ignore_exts])
+                }
+                in_folder = in_folder | files
```
sampsyo, line 55: Maybe I'm missing something obvious, but was there a specific reason to switch away from the set comprehension and to use a `for` loop instead? It seems like adding `r not in ignore_dirs and...` would have the desired effect, right?

sampsyo (commented):
Awesome; looks good so far! Here are a few suggestions.

docs/changelog.rst
jaydesl: Hi @sampsyo thanks for the quick review. How's the new code?

beetsplug/unimported.py
jaydesl: agreed - changed in the PR

beetsplug/unimported.py
jaydesl: Honestly, just for readability. I'm happy to keep comprehensions for the sake of speed, but I feel like it makes the code less readable. What do you think?

sampsyo: Ah, oddly enough, I actually found the comprehension version easier to read (because it's closer to "declaratively" writing down what the desired set contains). It could certainly use some additional whitespace/line-breaks, but can we give it a shot and see how bad it looks?

sampsyo:
Oops; it looks like there are a couple of tiny lint problems—should be easy to fix! Let's merge after that turns "green."

jaydesl:
@sampsyo Thanks for your patience! Should be fine now :)

janpeeters70:
Thanks a lot for implementing the 'ignore_subdirectories' request.

I'm trying to figure out how I can add a list of subdirectory's with a space and '(' in their names. 
I've tried putting ' around them, but that didn't work. Thanks.

janpeeters70:
Thanks for the pointer @sampsyo. I feel a bit stupid, I researched and tried many options still with no success. 
What I tried. I paste various `ignore_subdirectories` after each other. I know this doesn't work just to share what I tried.

```
unimported:
    ignore_extensions: jpg, jpeg
    ignore_subdirectories: [Albums (niet geimporteerd), Compilations, DJ Music]
    ignore_subdirectories: 'Albums (niet geimporteerd)', Compilations, 'DJ Music'
    ignore_subdirectories: "Albums (niet geimporteerd)", Compilations, "DJ Music"
    ignore_subdirectories:
          - Albums (niet geimporteerd)
          - Compilations
          - DJ Music
    ignore_subdirectories:
          - 'Albums (niet geimporteerd)'
          - Compilations
          - 'DJ Music'
    ignore_subdirectories:
          - "Albums (niet geimporteerd)"
          - Compilations
          - "DJ Music"
 ```

What am I missing?

janpeeters70:
That's an idea, maybe I did something wrong with the editable install. I'll check that and report back.

PR #4090, opened 2021-10-04
docs: fix links that were failing sphinx' linkcheck

wisp3rwind:
As the integration tests show, some of our links are rotting...

...also, sphinx is trying to access AURA at 127.0.0.1.


sampsyo (approved):
Good catches; thanks!!

PR #4092, opened 2021-10-05
Update CI to the latest 3.10 release

arogl:
## Description

Updates CI to latest 3.10 release


wisp3rwind:
Maybe let's wait for the first Python 3.11 alpha to hit github actions (should be soon: https://www.python.org/dev/peps/pep-0664/) so that we can keep the current structure of the configuration?

sampsyo:
According to the CI results, it looks like the name of the version may be "3.10" rather than "3.10.0"?

arogl:
My first attempt was just for 3.10, but the CI system interpreted that as 3.1

wisp3rwind:
> Not sure why coverage failed on 3.9, windows, when it has worked on other requests.

I had a brief look at the log: The test in question spawns another process running the bpd daemon, which does indeed access the file that could not be deleted. Presumably, in

https://github.com/beetbox/beets/blob/afbb12d68b5279d8a2d9207abd0dd2c89f5043d1/test/test_player.py#L338

the timeout can occasionally be too short if something blocks the process. Then, the daemon has not yet quit on test teardown. We may want to increase that timeout a little (1-2 seconds? This is only a safeguard in case the daemon actually locks up, so making it quite long shouldn't be a problem) to make the test more reliable.

arogl:
This should be merged and a new PR for 3.11-a05

wisp3rwind:
> This should be merged and a new PR for 3.11-a05

I think it would be easier to update this PR to include the 3.11 alpha. Splitting the PR would mean switching the logic for older/current/upcoming Python versions back and forth. Do you have the time to do the update?

.github/workflows/ci.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, windows-latest]
-        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10', '3.11-a05']
+        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10', '3.11.0-alpha.5']
```
wisp3rwind, line 15: Is it possible to specify `'3.11' here to always get the latest available development version?

wisp3rwind (changes requested):
Thanks @arogl for taking care of this. I have one comment about hardcoding the specific versions, which if possible would be nice to avoid for easier maintenance.

.github/workflows/ci.yaml
arogl: Updated to use 3.11

wisp3rwind:
Any ideas why the last commit didn't trigger the CI?

I've found some documentation on specifying python versions at https://github.com/actions/setup-python#usage, apparently the "canonical" way to require the latest pre-release is a range specification like `python-version: '3.11.0-alpha - 3.11.0'`.

sampsyo:
Weird about Actions not being triggered… I'm also stumped?

Using the range specifier for the Python version seems like a cool idea (since it will avoid becoming out of date as releases happen).

wisp3rwind:
> Any ideas why the last commit didn't trigger the CI?
> 
> I've found some documentation on specifying python versions at https://github.com/actions/setup-python#usage, apparently the "canonical" way to require the latest pre-release is a range specification like `python-version: '3.11.0-alpha - 3.11.0'`.

@arogl, would you have the time to update the PR accordingly?

arogl:
> > Any ideas why the last commit didn't trigger the CI?
> > I've found some documentation on specifying python versions at [actions/setup-python#usage](https://github.com/actions/setup-python#usage), apparently the "canonical" way to require the latest pre-release is a range specification like `python-version: '3.11.0-alpha - 3.11.0'`.
> 
> @arogl, would you have the time to update the PR accordingly?

@wisp3rwind I have updated to use the ```-dev``` option and dropped Python 3.6

arogl:
> Thanks!
> 
> We do still support Python 3.6, don't we? I think we should keep testing it then.

3.6 has been dropped from support since Sept 2021

https://peps.python.org/pep-0494/

wisp3rwind:
Fair enough. Then we should drop Python 3.6 entirely (see the new issue I just opened).

PR #4093, opened 2021-10-05
[WIP] list field support

Duncaen:
I'm interested in getting support for list fields and this is my initial draft for which I would like to get feedback, to see if this is the right approach since list fields are a long standing open issue and there were other attempts like the genre field.

This adds additional tables per list field to the database, which are linked to the entity like flex attributes.

Adding the artists lists using `beets modify artists="foo;bar"` already works and mediafile will write the tag for each separate value.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -255,6 +255,16 @@ class Model:
     do not relate to any specific field.
     """
 
+    _list_tables = {}
+    """A mapping of "fixed" list fields on this type. The keys are field
+    names and the values are table names.
+    """
+
+    _list_types = {}
+    """A mapping of "fixed" list fields on this type. The keys are field
+    names and the values are `Type` objects.
+    """
+
     _always_dirty = False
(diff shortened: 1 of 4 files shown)


beets/dbcore/db.py
```diff
@@ -1057,6 +1130,20 @@ def _make_attribute_table(self, flex_table):
                     ON {0} (entity_id);
                 """.format(flex_table))
 
+    def _make_list_tables(self, list_tables, list_types):
+        """Create tables for list types.
+        """
+        with self.transaction() as tx:
+            for key, name in list_tables.items():
+                tx.script(f"""
```
sandersantema, line 1138: Isn't this supposed to be unsafe? i.e. one should use `.format` instead of f-strings? It would in any case be more consistent as this is what you do in the other sql strings. But this is just a guess as I'm only reading trough this PR because it seems like a really nice function to have and I don't have a lot of knowledge of the beets codebase.

PR #4095, opened 2021-10-06
Formatted modify and import --set-field.

### LINKED ISSUE #488: Formatted modify
There are a number of cool things we could do if the `modify` command supported templates in the value part. For example, you could put the Echo Nest liveness into the comments field like so:

```
$ beet mod comments='liveness: $liveness'
```

It might also be nice to expose these rewritings in an import hook also.

Duncaen:
## Description

Fixes #488 

## To Do


sampsyo:
Awesome!! Thanks for putting this together! This looks really good at first glance; I will give it a thorough review soon.

beets/ui/commands.py
```diff
@@ -1424,7 +1421,9 @@ def modify_items(lib, mods, dels, query, write, move, album, confirm):
            .format(len(objs), 'album' if album else 'item'))
     changed = []
     for obj in objs:
-        if print_and_modify(obj, mods, dels) and obj not in changed:
+        obj_mods = {key: model_cls._parse(key, format(obj, value))
+                    for key, value in mods.items()}
```
sampsyo, line 1425: It occurs to me that we are losing a couple of optimizations by moving this logic inside the loop:

1. The template is compiled *n* times. This is not required; we could hoist the template compilation out of the loop and reuse it in place of the `value` string.
2. `_parse` is called *n* times instead of once. This is kind of fundamental; the formatted value might be different and need to be parsed differently on every iteration. We could consider detecting whether the value actually uses any templating and avoid this if the value happens to be a constant—the way `beet modify` used to behave.

I propose we attempt to address 1 now, as part of this PR, and create a separate issue for addressing item 2. The former should (hopefully) be easy. The latter will require a bit more complexity and IMO is not really necessary to ship this.

sampsyo (commented):
Awesome! This overall looks really solid; thanks for your patience while I did a complete review.

I just have one comment in here about performance. What do you think about deferring the actual solution to a separate issue? It seems like it would be cool to ship this as-is and mostly worry about performance later. But there may be one piece of low-hanging fruit (pre-compiling the string template) that might be nice to address now.

beets/ui/commands.py
kergoth: Looks like @Duncaen added a couple new commits to this PR, is this review still accurate?

sampsyo: No; looks like this is now rearranged a bit! I think this is due for another look—it may be ready to go as-is?

sampsyo:
This is still alive.

PR #4102, opened 2021-10-16
Add popularity tag

### LINKED ISSUE #4094: Spotify: Get track popularity from Spotify
Track popularity is probably one of the most important attributes. Spotify directly provides `popularity` information in its [Tracks API](https://developer.spotify.com/documentation/web-api/reference/#category-tracks). I am surprised that we don't import this attribute, given that we already are using the [Tracks API](https://github.com/beetbox/beets/blob/master/beetsplug/spotify.py#L43). Wondering if this is just an oversight or something that we should implement. I saw an older [thread](https://github.com/beetbox/beets/issues/3578) where @sampsyo agreed to include Spotify-specific attributes. 

### Proposed solution

Retrieve track popularity from Spotify and include it along with other tags. 

### Objective

Get track popularity information to enable discovery. 

#### Goals

Create better playlists and recommendations.

happiestbee:
## Get popularity tag for tracks from track api in Spotify plugin 

Fixes #4094 .  

Makes use of the popularity information for tracks

## To Do

### DIFF
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -161,9 +161,8 @@ class TrackInfo(AttrDict):
                  data_source=None, data_url=None, media=None, lyricist=None,
                  composer=None, composer_sort=None, arranger=None,
                  track_alt=None, work=None, mb_workid=None,
-                 spotify_popularity=None, work_disambig=None,
-                 bpm=None, initial_key=None, genre=None,
-                 **kwargs):
+                 work_disambig=None, bpm=None, initial_key=None,
+                 genre=None, **kwargs):
         self.title = title
         self.track_id = track_id
         self.release_track_id = release_track_id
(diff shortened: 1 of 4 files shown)


sampsyo:
Seems good overall to me! Can you also please add a changelog entry?

happiestbee:
> Seems good overall to me! Can you also please add a changelog entry?

Ok, that should be done

arsaboo:
While we are at it...can we also save the track_id (maybe part of `items` table)? We can then use the track_id to get other [audio features](https://developer.spotify.com/console/get-audio-features-several-tracks/) (in a subsequent PR).

arsaboo:
So, I tested this a little more and it looks like the `track` API will not work. I started getting 429 errors (API rate limit exceeded) and I am guessing it is because every track is triggering an API call and that can easily exceed the [API rate limit](https://developer.spotify.com/documentation/web-api/guides/rate-limits/) for a large album. Instead of `track` API call, we should use `tracks` API call that allows multiple tracks (maximum of 50) to be called.

PR #4107, opened 2021-10-17
Clarify config docs for `per_disc_numbering`

ali-ramadhan:
## Description

Originally I found the `per_disc_numbering` docs lacking since the path format suggested adds disc numbers to all tracks, even single-disc tracks.

I then found out about the inline plugin which even had a nice `$disc_and_track` example which I thought should be referenced in the section about `per_disc_numbering`.

## To Do

Does this require a changelog entry? If so, happy to add one.


sampsyo:
It would be great to add in that extra link here!

PR #4124, opened 2021-10-26
Add a basic "--album" flag to "beet info"

### LINKED ISSUE #2134: info: Add per-album option
Branched from https://github.com/beetbox/beets/issues/2132
### Problem

There is currently no option to find all the metadata from an album.
This would be especially handy for templating as path variables prefer album metadata.

`beet info -a Album` seems like a good idea, returning just the album metadata, not the info for the tracks.

tianon:
This currently implies `--library` because I'm not sure what `album info` of the tags of individual files would mean. 😅

(Honestly it's felt really off that `beet info` shows details of the tags by default when every other command shows information from the beet database by default, but looking at the code I learned it could be used directly on unimported files, so I guess that makes some sense. :innocent:)

I've tested this successfully in my own `pluginpath` directory, although admittedly haven't done much more than `beet info --album` (haven't tested the combination with other flags, but from the code I don't see any reasons for them not to work).

I'm happy to adjust, rebase, amend, etc, although I'll probably need some specific guidance if you want code changes because I don't write a lot of Python. :sweat_smile:

Fixes #2134.

## To Do

### DIFF
--- a/beetsplug/export.py
+++ b/beetsplug/export.py
@@ -149,8 +149,6 @@ class ExportPlugin(BeetsPlugin):
                 if isinstance(value, bytes):
                     data[key] = util.displayable_path(value)
 
-            items += [data]
-
             if file_format_is_line_based:
                 export_format.export(data, **format_options)
             else:
--- a/beetsplug/export.py
+++ b/beetsplug/export.py
@@ -88,6 +88,10 @@ class ExportPlugin(BeetsPlugin):
             help='show library fields instead of tags',
         )
         cmd.parser.add_option(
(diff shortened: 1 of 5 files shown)


sampsyo:
This all looks awesome; thank you!!

tianon:
oops: :sweat_smile:

https://github.com/beetbox/beets/blob/be82fd0f4a94d6ea820cf05d0105f999119daf03/beetsplug/export.py#L82

(missed a `TODO` comment that should've been deleted :innocent:)

sampsyo:
Ah—can you please open a PR to delete it?

PR #4129, opened 2021-10-30
Use "colordiff" to highlight "beet move" path differences

tianon:
## Description

This uses the common `colordiff` function such that path pairs shown by `beet move` have their differences highlighted, which makes them _way_ easier to scan through visually (more so if they were lined up better or if the added vs subtracted colors were different, but this makes a huge difference by itself).

## To Do

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -757,15 +757,21 @@ def show_path_changes(path_changes):
     if max_width > col_width:
         # Print every change over two lines
         for source, dest in zip(sources, destinations):
-            log.info('{0} \n  -> {1}', source, dest)
+            color_source, color_dest = colordiff(source, dest)
+            print_('{0} \n  -> {1}'.format(color_source, color_dest))
     else:
         # Print every change on a single line, and add a header
         title_pad = max_width - len('Source ') + len(' -> ')
 
-        log.info('Source {0} Destination', ' ' * title_pad)
+        print_('Source {0} Destination'.format(' ' * title_pad))
         for source, dest in zip(sources, destinations):
             pad = max_width - len(source)
-            log.info('{0} {1} -> {2}', source, ' ' * pad, dest)
+            color_source, color_dest = colordiff(source, dest)
+            print_('{0} {1} -> {2}'.format(
+                color_source,
+                ' ' * pad,
+                color_dest,
+            ))
 
 
 # Helper functions for option parsing.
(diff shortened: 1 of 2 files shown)


tianon:
Here's a screenshot of how it looks:

image

tianon:
I suppose I should add that I've been running with this patch for a few weeks, and it's really helpful for me to dial in my path rewrites since I can do a mass `beet move --pretend` and then scroll through and verify the changes very very quickly (whereas before it was tedious to even look at).

wisp3rwind:
The result looks nice! However, I'm not sure this is the right implementation: This passes colorized text through the logging infrastructure. Is that a thing we want to do/that works reliably? Or should this be changed to present its UI through `beets.ui.print_` instead? Maybe someone more familiar with the UI code can chime in?

sampsyo:
Hey! Indeed, this seems pretty snazzy. I think @wisp3rwind is right here that we should probably be `print_`ing this stuff instead of `log.info`ing. it. Maybe we should do that with all of the output from the `move` command, in fact? Why don't we just switch that over, and then this should be good to go.

tianon:
Ok, I think that finally does it.  There's one more `log.info` as part of `beet move` (that I can see):

https://github.com/beetbox/beets/blob/361ff05b5112377b3c05bef56b8b33f42b038ee2/beets/ui/commands.py#L1556-L1560

However, I _think_ this one makes sense to leave as `log.info` so that users who don't want any output at all and just want files moved can do so?  (Since this path diffing command is only used with `--pretend` and confirmation mode within this function, I think.)

Hopefully I haven't misunderstood. :sweat_smile: (Happy to make more changes if so!)

I think we could probably improve this further by making each path relative to the library directory so they're all typically shorter (at least the right hand side would then almost always be a lot shorter), but that's probably more controversial? :innocent:

sampsyo (approved):
Yay; this all looks great to me! I think it's ready to merge (but would love to have a second pair of eyes on it if possible).

I think leaving that remaining `log.info` seems right to me. To be honest, the line here is pretty blurry—it's not always precisely clear what is "logging" and what is "console output". But this division here seems perfectly reasonable in this case.

Making the paths relative to the library directory is interesting too! And yeah, perhaps less of a slam-dunk than the colorization. I'd be happy to discuss that too.

wisp3rwind:
> Yay; this all looks great to me! I think it's ready to merge (but would love to have a second pair of eyes on it if possible).

Looks good to me!

> I think leaving that remaining `log.info` seems right to me. To be honest, the line here is pretty blurry—it's not always precisely clear what is "logging" and what is "console output". But this division here seems perfectly reasonable in this case.

Yep, at a first glance that also seems rather blurry to me.

PR #4131, opened 2021-10-30
Remove completed TODO item

tianon:
## Description

Removes a `TODO` comment that was completed in #4124 (but I didn't notice the comment in the code until after it was merged and I was looking at the file again later :facepalm:).

## To Do


wisp3rwind:
Thanks!

PR #4133, opened 2021-11-02
Add option to convert cover art file format.

ArsenArsen:
## Description

Introduces an option to force an image format, as some IEMs (notably, Rockbox) might lack support for some image formats.

## To Do

I want to do tests, but I am unsure as to how to implement a test that changes the resulting cover file name.

PS: there might be a bug in cleanup of the temporary files here, however, my `/tmp` is too polluted to test, so consider this a call for help. I also don't have Windows anywhere to test on.

### DIFF
--- a/docs/plugins/fetchart.rst
+++ b/docs/plugins/fetchart.rst
@@ -90,6 +90,10 @@ file. The available options are:
   instructed to store cover art as non-progressive JPEG. You might need this if
   you use DAPs that don't support progressive images.
   Default: ``no``.
+- **cover_format**: If enabled, forced the cover image into the specified
+  format. Most often, this will be either ``JPEG`` or ``PNG`` [#imgformats]_.
+  Also respects ``deinterlace``.
+  Default: None (leave unchanged).
 
 Note: ``maxwidth`` and ``enforce_ratio`` options require either `ImageMagick`_
 or `Pillow`_.
@@ -105,6 +109,12 @@ or `Pillow`_.
 .. _beets custom search engine: https://cse.google.com.au:443/cse/publicurl?cx=001442825323518660753:hrh5ch1gjzm
 .. _Pillow: https://github.com/python-pillow/Pillow
 .. _ImageMagick: https://www.imagemagick.org/
+.. [#imgformats] Other image formats are available, though the full list
+   depends on your system and what backend you are using. If you're using the
+   ImageMagick backend, you can use ``magick identify -list format`` to get a
+   full list of all supported formats, and you can use the Python function
+   PIL.features.pilinfo() to print a list of all supported formats in Pillow
+   (``python3 -c 'import PIL.features as f; f.pilinfo()'``).
 
 Here's an example that makes plugin select only images that contain ``front`` or
 ``back`` keywords in their filenames and prioritizes the iTunes source over
--- a/beets/util/artresizer.py
+++ b/beets/util/artresizer.py
@@ -18,6 +18,7 @@ public resizing proxy if neither is available.
 
 import subprocess
 import os
+import os.path
 import re
 from tempfile import NamedTemporaryFile
 from urllib.parse import urlencode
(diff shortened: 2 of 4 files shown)


docs/plugins/fetchart.rst
```diff
@@ -90,6 +90,10 @@ file. The available options are:
   instructed to store cover art as non-progressive JPEG. You might need this if
   you use DAPs that don't support progressive images.
   Default: ``no``.
+- **cover_format**: If enabled, forced the cover image into the specified
+  format. Valid values line up with ImageMagick/Pillow decoder names (for
+  instance, ``PNG`` or ``JPEG``). Also respects ``deinterlace``.
```
sampsyo, line 95: Maybe we should just give a list of all the formats we expect to work? We could just start with `PNG` and `JPEG`, for example, and not bother people with the somewhat subtle advice to look in the ImageMagick/Pillow documentation. (And if folks want other formats, maybe we can test those and add them to the docs.)

sampsyo (commented):
Seems great so far! I too think it would be awesome to get some help trying out the whole `/tmp` cleanup stuff here, if anyone has any extra cycles…

docs/plugins/fetchart.rst
ArsenArsen: This might be a bit more complex. These two are the ones I tested and my confidence isn't high that the two libraries use matching names for others. I'll look for the lists in a bit.

ArsenArsen:
I figured out a solution for testing: `firejail --noprofile --private-tmp`. There is, indeed, an issue with cleanup, I'll fix it.

docs/plugins/fetchart.rst
ArsenArsen: The list can be obtained via `magick identify -list format` and via `PIL.features.pilinfo()` for Pillow.

ArsenArsen:
No more junk left in `/tmp`!

docs/plugins/fetchart.rst
sampsyo: Yeah, I think that's the practical thing—just listing a few common formats that 99.99% of people will want to use seems like the most helpful thing. And yes, supplementing that with a docs link will help satisfy the curiosity of the remaining 0.01%.

ArsenArsen: Got it, I'll change that and squash the commits.

sampsyo:
Seems awesome! Thanks for your work on massaging the documentation, which looks very neat now. I'll merge this immediately! :sparkles:

ArsenArsen:
Thank you!

ArsenArsen:
oh, that's true, I haven't used those so I'd have forgotten to also add that special case, good catch

wisp3rwind:
> EDIT: do you think it'd be a good idea to modify this a bit, and instead of returning REFORMAT/DEINTERLACE, merge those into one and make them a flag, so that it happens for all kinds of candidates?

In fact, with the various recent additions to fetchart/artresizer, I think that the design of these function needs to be rethought. The current code is quite hard to reason about. Maybe, instead of these mutually exclusive flags, we should have something like a bitfield (in python, probably a list of required actions?), which is passed directly to a single artresizer function, which handles them all? For example, this could take the form of an `options=["downsize", "reformat", "deinterlace"]` argument for `ArtResizer.shared.resize`.

ArsenArsen:
a set is more likely the appropriate function, and I disagree

I'd do one of two things, one of which is more drastic:

- reformat all inputs unconditionally (which will simplify the code a lot and realistically not add an unreasonable cost in the modern age), or
- run a single all-purpose reencode if something should be changed

sampsyo:
You're right that a refactoring is in order here, @wisp3rwind. I like your idea of returning a set of options (perhaps a bitfield, or any other way of "unioning" enum values) instead of these mutually exclusive action flags, which have outlived their usefulness.

PR #4146, opened 2021-11-16
Send the pluginload event after types and queries are available

kergoth:
## Description

Making these types and queries available is part of fully loading the plugins,
so the event should not be sent until this work is done. This allows plugins
to make use of those types and queries in a pluginload listener.

## To Do

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -1129,7 +1129,6 @@ def _load_plugins(options, config):
         plugin_list = config['plugins'].as_str_seq()
 
     plugins.load_plugins(plugin_list)
-    plugins.send("pluginload")
     return plugins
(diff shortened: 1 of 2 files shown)


kergoth:
I attempted to add a test for this, but the way plugins are mocked ensures that the function altered here is not used at all at this time. It'd require reworking that. Thoughts?

kergoth:
I considered just moving the queries/types update into the plugin load function, but the _setup function is specifically described as updating globals, so it seemed more appropriate to move the send there instead.

sampsyo (approved):
Nice; this seems great!! It's good to merge, IMO. I tend to agree that this would be super awkward to test, so it's probably not worth doing.

Issue #4151, opened 2021-11-19
Failde unit test on macOS via MacPorts


catap:
### Problem

```sh
$ sudo port -v test beets
...
==== FAILURES ====
____ DestinationTest.test_destination_escapes_leading_dot ____

self = <test.test_library.DestinationTest testMethod=test_destination_escapes_leading_dot>

    def test_destination_escapes_leading_dot(self):
        self.i.album = '.something'
        dest = self.i.destination()
        self.assertTrue(b'something' in dest)
>       self.assertFalse(b'/.' in dest)
E       AssertionError: True is not false

test/test_library.py:230: AssertionError

```

### Setup

* OS: macOS 11.6
* Python version:  3.9.9
* beets version: 1.5.0 and f3552f09cfa330b7e875526bf6162cb5f3ba174e
* Turning off plugins made problem go away (yes/no): no

### Reproducer

This is quite easy to reproduce by applying this patch:

```python
diff --git a/test/test_library.py b/test/test_library.py
index e64f7556..f94130ac 100644
--- a/test/test_library.py
+++ b/test/test_library.py
@@ -148,7 +148,7 @@ class GetSetTest(_common.TestCase):
 class DestinationTest(_common.TestCase):
     def setUp(self):
         super().setUp()
-        self.lib = beets.library.Library(':memory:')
+        self.lib = beets.library.Library(':memory:', '~/.Music')
         self.i = item(self.lib)
 
     def tearDown(self):
```

PR #4152, opened 2021-11-19
Prevent fails of tests when path contains dots

catap:
Unit test may fails when path to temprorary library contains `.`; to garantue that bug wasn't here, it forces to use one more `.` inside path.

Fixes: https://github.com/beetbox/beets/issues/4151

### DIFF
--- a/test/test_library.py
+++ b/test/test_library.py
@@ -148,7 +148,10 @@ class GetSetTest(_common.TestCase):
 class DestinationTest(_common.TestCase):
     def setUp(self):
         super().setUp()
-        self.lib = beets.library.Library(':memory:')
+        # default directory is ~/Music and the only reason why it was switched
+        # to ~/.Music is to confirm that tests works well when path to
+        # temporary directory contains .
+        self.lib = beets.library.Library(':memory:', '~/.Music')
         self.i = item(self.lib)
 
     def tearDown(self):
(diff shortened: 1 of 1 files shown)


test/test_library.py
```diff
@@ -224,7 +224,7 @@ def test_destination_escapes_leading_dot(self):
         self.i.album = '.something'
         dest = self.i.destination()
         self.assertTrue(b'something' in dest)
-        self.assertFalse(b'/.' in dest)
+        self.assertFalse(b'/.something' in dest)
```
sampsyo, line 227: This seems reasonable!

test/test_library.py
```diff
@@ -148,7 +148,7 @@ def test_album_fallback(self):
 class DestinationTest(_common.TestCase):
     def setUp(self):
         super().setUp()
-        self.lib = beets.library.Library(':memory:')
+        self.lib = beets.library.Library(':memory:', '~/.Music')
```
sampsyo, line 151: I'm not quite sure I understand this—what does this directory argument do for the tests? Are we in any danger of writing files to this directory? Is there a reason it's located in the home directory (`~`) instead of in the root (`/`)?

sampsyo (commented):
Cool; thanks! I have a couple of questions within.

test/test_library.py
catap: @sampsyo default directory is `~/Music` and the only reason why I had switched to `~/.Music` is to trigger a bug outside of MacPorts by adding second `.` in the path which is triggered a false fail in unit test.

sampsyo: OK, got it, thanks. Do you think you could add a comment that clarifies why this is necessary? It would be great to leave clues for future maintainers—otherwise, this could look pretty mysterious if you don't have the context.

catap: @sampsyo here it is!

PR #4153, opened 2021-11-20
Add the item fields bitrate_mode, encoder_info and encoder_settings

jcassette:
## Description

See beetbox/mediafile#58
See #3094

## To Do

- [x] Documentation
- [x] Changelog
- [x] Tests (nothing to do)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -10,6 +10,7 @@ New features:
 
 * :doc:`/plugins/kodiupdate`: Now supports multiple kodi instances
   :bug:`4101`
+* Add the item fields ``bitrate_mode``, ``encoder_info`` and ``encoder_settings``.
 
 Bug fixes:
 
@@ -28,6 +29,7 @@ For packagers:
 
 * We fixed a version for the dependency on the `Confuse`_ library.
   :bug:`4167`
+* The minimum required version of :pypi:`mediafile` is now 0.9.0.
 
 
 1.6.0 (November 27, 2021)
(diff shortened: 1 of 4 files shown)


docs/changelog.rst
```diff
@@ -51,6 +51,7 @@ Other new things:
   yes`` in your configuration to enable.
 * :doc:`/plugins/fetchart`: A new option to change cover art format. Useful for
   DAPs that do not support some image formats.
+* :doc:`/dev/library`: Add property bitrate_mode.
```
sampsyo, line 54: We generally only use these prefixes in the changelog for plugins. Also, you might want to surround the name of the new field in backticks to make it render nicely.

sampsyo (commented):
Thanks!! Maybe we should hold off on merging this until the MediaFile support is released, and then we should require that version of the MediaFile library?

docs/changelog.rst
jcassette: Okay, I have removed the prefix and added the backticks

jcassette:
Sure, I have updated the changelog and setup.py regarding the new required version of mediafile :-)

jcassette:
I have also added the properties encoder_info and encoder_settings since they are available.
This would fix #3094 by allowing to get information about the VBR quality.

jcassette:
Hi @sampsyo I have updated this :-)

sampsyo:
Looks great; thanks!!

jcassette:
Thank you 👍

PR #4155, opened 2021-11-22
Add an 'album_removed' event.

djl:
## Description

This works similarly to the existing 'item_removed' event but is called with an `Album` object.

My primary goal for writing this is to scratch my own itch. I'd like to update the `mbcollection` plugin to support automatically removing albums when an album is removed from the library. There was no easy way to do this (e.g. you have to use the `item_removed` event and keep track of all the items in an album) so here we are 😄 

It's been a while since I've had a look at the code so please feel free to let me know if this is the correct way to approach this.

## To Do

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -878,7 +878,6 @@ class Item(LibModel):
             album = self.get_album()
             if album and not album.items():
                 album.remove(delete, False)
-                plugins.send('album_removed', album=album)
 
         # Send a 'item_removed' signal to plugins
         plugins.send('item_removed', item=self)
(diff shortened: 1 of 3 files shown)


sampsyo:
Seems good overall! But perhaps this should go right in the `Album.remove` method, instead of in the `Item.remove` case for album-having items?
https://github.com/beetbox/beets/blob/ca1974ff863ddd07a2a185537c73fd6cc95f0d76/beets/library.py#L1136

djl:
Ah yes. Good point.

Should the event be sent before or after the items are removed? Which makes more sense? Does it even matter?

sampsyo:
Hmm, good question. I guess one reason to prefer "before" would be that the current `item_removed` event comes before the files are actually deleted. And putting the new `album_removed` event after the `item.remove()` calls would mean that it would come after the files are deleted, whereas putting it before would preserve this ordering.

djl:
Moved it to the Album class, just below the `super().remove()` call.

sampsyo:
Looks great; thanks!!

PR #4175, opened 2021-11-30
Add lib64 to .gitignore

smichel17:
lib64/ was already included, but sometimes lib64 is a symlink to lib/


sampsyo:
Hmm, maybe we should just remove the slash on both lib and lib64 to cover both cases?

sampsyo:
Alright!

PR #4176, opened 2021-11-30
Rewrite plugin: Prioritize explicit albumartist rules

smichel17:
## Description

The `rewrite` plugin applies patterns for 'artist' to 'albumartist' too.

However, this means you didn't know that the patterns are checked in
order, you might write something like this:

    rewrite:
        artist .*jimi hendrix.*: The Jimi Hendrix Experience
        albumartist .*jimi hendrix.*: Jimi Hendrix

..and be sad when your alubumartist becomes The Jimi Hendrix Experience.

This change always prioritizes explicit 'albumartist' rules over those
inherited from 'artist', regardless of which order they're defined.

## Caveats

I'm not sure if this is a good idea, since it reduces power/flexibility. Before, if you *wanted* to have an artist rule selectively override an albumartist rule, you could put it in front. Now you can't. My main motivation was getting my feet wet in the beets code base, so I don't mind if this is not merged :)

Maybe a better solution is just to document the ordering— I wasn't aware that the keys in yaml dictionaries were ordered (maybe they're not supposed to be and this is an implementation quirk?). Either way, this is not ready for merge, so I'll leave it as a draft for now.

## To Do

### DIFF
--- a/beetsplug/rewrite.py
+++ b/beetsplug/rewrite.py
@@ -60,10 +60,11 @@ class RewritePlugin(BeetsPlugin):
             self._log.debug('adding template field {0}', key)
             pattern = re.compile(pattern.lower())
             rules[fieldname].append((pattern, value))
-            if fieldname == 'artist':
-                # Special case for the artist field: apply the same
-                # rewrite for "albumartist" as well.
-                rules['albumartist'].append((pattern, value))
+
+        # Special case for the "artist" field: apply the same rewrite for
+        # "albumartist" as well (only if no other "albumartist" rule matches)
+        if 'artist' in rules:
+            rules['albumartist'] += rules['artist']
 
         # Replace each template field with the new rewriter function.
         for fieldname, fieldrules in rules.items():


sampsyo:
Interesting—yeah, I'm attracted to documenting the ordering so folks can figure this out themselves. One-off priority rules seem a little brittle here. Care to take a stab at the docs?

smichel17:
I'm happy to document it. But I'm also a little concerned with that approach, since in the [yaml spec](https://yaml.org/spec/1.2.2/#3221-mapping-key-order), mappings don't have any natural order; it's up to whatever program serializes them to decide on an order. Ours *happens* to do it in the defined order, but I'm not super comfortable codifying that, either.

I noticed the other day that Confuse has the `as_pairs()` option. What if we also allowed `rewrite` rules in that format?

```yaml
rewrite:
    - artist .*jimi hendrix.*: The Jimi Hendrix Experience
    - albumartist .*jimi hendrix.*: Jimi Hendrix
```

If someone uses pairs, the rules are evaluated in that order. If they use a simple mapping, then it's some arbitrary order.

_Maybe_ in the future, we could add some kind of specificity heuristic for the regex, like [reach router](https://reach.tech/router/ranking) does with paths.

sampsyo:
It is probably relevant that Confuse deviates from the YAML spec in this respect. Dictionaries are ordered. We use this, for example, for the `paths:` section of the configuration, which is similarly prioritized.

smichel17:
Okay! I'll close this and send a documentation PR instead.

..although, I do like the idea of accepting both a dictionary and a list of pairs, and documenting that they both currently function the same way, but only the list of pairs is guaranteed to preserve order in the future.

Issue #4162, opened 2021-11-27
Lyrics: Confusion with own genius api key


kjboa:
After I had problems with the given genius api key, I tried to put in my own in order to see if the problem of Lyrics not being pulled was about beets or genius.com. The Relevant Section of my config file looks like this:

lyrics:      
   sources: genius    
   genius_api_key: Client_ID Client_Secret

However if I know try to request lyrics with e.g. beet -vv lyrics playboi carti magnolia it fails with:

`user configuration: /root/.config/beets/config.yaml
data directory: /root/.config/beets
plugin paths:
Sending event: pluginload
library database: /mnt/2TB/media/Beets/Library/library.db
library directory: /mnt/2TB/media/Beets/Music
Sending event: library_opened
Traceback (most recent call last):
  File "/usr/local/bin/beet", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/dist-packages/beets/ui/__init__.py", line 1285, in main
    _raw_main(args)
  File "/usr/local/lib/python3.9/dist-packages/beets/ui/__init__.py", line 1272, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "/usr/local/lib/python3.9/dist-packages/beetsplug/lyrics.py", line 820, in func
    self.fetch_item_lyrics(
  File "/usr/local/lib/python3.9/dist-packages/beetsplug/lyrics.py", line 919, in fetch_item_lyrics
    lyrics = [self.get_lyrics(artist, title) for title in titles]
  File "/usr/local/lib/python3.9/dist-packages/beetsplug/lyrics.py", line 919, in <listcomp>
    lyrics = [self.get_lyrics(artist, title) for title in titles]
  File "/usr/local/lib/python3.9/dist-packages/beetsplug/lyrics.py", line 952, in get_lyrics
    lyrics = backend.fetch(artist, title)
  File "/usr/local/lib/python3.9/dist-packages/beetsplug/lyrics.py", line 356, in fetch
    for hit in json["response"]["hits"]:
KeyError: 'response'`

Did i maybe put the API key in the wrong format or set a wrong redirect url in the genius api panel? Or might this be a beets problem?

Issue #4137, opened 2021-11-09
[Lyrics Plugin] Genius failed to find artist


kjboa:
Hi, uppon requesting lyrics with beet -vv lyrics the genius source always fails and regardless of artist and song always gives this message.
In the Case of Halsey:
`lyrics: Genius failed to find a matching artist for 'Halsey'`
Maybe this is a problem with how the plugin handles Charactersets or smth, since on the regular Genius Site Lyrics for this song exist.

sampsyo:
That's odd! To try debugging this, you might try seeing what a search like `https://api.genius.com/search?q=Halsey` produces directly from the Genius API. You'll need to provide an API key (you could try using beets's own, which is in its source code) and figure out the appropriate curl or httpie incantation to authenticate the request.

kjboa:
Sorry for the late answer. I could find the API key in lyrics.py but sadly I wasn't able to figure out how to make the proper request. I however would really like to have genius lyrics in the file tags, and solve this problem. Is there any other way I could debug this?

wisp3rwind:
> Sorry for the late answer. I could find the API key in lyrics.py but sadly I wasn't able to figure out how to make the proper request. I however would really like to have genius lyrics in the file tags, and solve this problem. Is there any other way I could debug this?

There are a few related comments at the bottom of https://github.com/beetbox/beets/issues/3903 (fwiw, the query still doesn't seem to work for me...)

kjboa:
I had tried some other curl commands  but none worked for me using the given API key. One time though beet lyrics worked for a very short time, but pulled the same lyrics for all songs in the album...

kjboa:
After few more tries with various installs this problem still persists. Is there maybe an option to use an own genius api key? Does redacted in this context mean that the genius api key option can't be used any longer or that it will be retired soon? 1.5.1 also seems to include some changes with genius is there a way to upgrade to the unfinished version using pip?

sampsyo:
No, "redacted" is just our way of preventing people from accidentally leaking secrets when they post their config online.

To run the latest version from source, there are instructions in the FAQ.

kjboa:
Thanks! Now I have 1.5.1, and want to put in a own api key. Does it make any difference at all what I pick for redirection url in the genius api panel?

kjboa:
Still haven't figured it out

PR #4177, opened 2021-11-30
[Lyrics] Genius uses query param not body

piu130:
## Description

Currently, the Genius lyrics search does not work.
It throws the following error for all files I tried:
```
lyrics: Genius failed to find a matching artist for <artist>
```
I tested it locally and I now see:
```
lyrics: got lyrics from backend: Genius
lyrics: fetched lyrics: <song>
```

## To Do

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -377,7 +377,7 @@ class Genius(Backend):
         data = {'q': title + " " + artist.lower()}
         try:
             response = requests.get(
-                search_url, data=data, headers=self.headers)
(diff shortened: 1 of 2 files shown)


sampsyo:
Aha, thank you! This looks right to me! Maybe this fixes #4162 and #4137 as reported by @kjboa?

piu130:
@sampsyo Thank you for the fast review. You are right. It should fix #4137 and #4162 since the second issue is a following up of the first.

kjboa:
Just tested it and it really fixes the genius plugin. Thanks a lot 👍

PR #4178, opened 2021-11-30
Add template functionality in rewrite plugin

smichel17:
## Description

Amongst other things, the goal is to be able to prevent overriding the `albumartist` just because you want to override the `album`. E.g.:

```yaml
rewrite:
    albumartist: $albumartist
    artist .*jimi hendrix.*: Jimi Hendrix
```

It also moves the `rewrite` plugin closer to the `inline` plugin in terms of functionality. So the example from the docs,

> ```yaml
> rewrite:
>     artist .*jimi hendrix.*: Jimi Hendrix
> ```

can be replaced with

```yaml
rewrite:
    artist .*: %ifdef{$artist_override,$artist_override,$artist}
```
Followed by `beet modify artist_override="Jimi Hendrix" artist::'.*jimi hendrix.*'`. This allows much more fine-grained canonicalizations, since you can select individual tracks or albums to modify by adding custom attributes. Or you can just write more fine-grained rewrite rules, like only modifying the artist on *some* albums.

## Status

Currently templates work as long as there's no (infinite) recursion, and crash if there is any recursion, despite my attempts to avoid this. I'm out of energy to bang my head against this for now. `Template.substitute()` must be independently fetching the list of template fields somewhere, since the one I pass in is definitely overridden with the previous, non-recursive, template.

@sampsyo If you're interested in me getting this good-enough to merge, I'd appreciate any guidance on what I'd need to touch in order to prevent the infinite recursion (following things around the call stack has been making my head spin a little…).

## To Do

- [ ] Tests. (~~Encouraged but not strictly required.~~ Definitely required for this one :P)

### DIFF
--- a/beetsplug/rewrite.py
+++ b/beetsplug/rewrite.py
@@ -49,10 +49,12 @@ def rewriter(field, rules, og_templ_funcs):
         return value
     return fieldfunc
 
+
 def og_getters(prev_templs, new_getters):
     def blankfield(item):
         return ""
-    return { k: prev_templs.get(k, blankfield) for k in new_getters.keys() }
+    return {k: prev_templs.get(k, blankfield) for k in new_getters.keys()}
+
 
 class RewritePlugin(BeetsPlugin):
     def __init__(self):
--- a/beetsplug/rewrite.py
+++ b/beetsplug/rewrite.py
@@ -22,23 +22,37 @@ from collections import defaultdict
 from beets.plugins import BeetsPlugin
 from beets import ui
 from beets import library
+from beets import plugins
+from beets.library import DefaultTemplateFunctions
+from beets.util.functemplate import template
 
 
-def rewriter(field, rules):
+def rewriter(field, rules, og_templ_funcs):
     """Create a template field function that rewrites the given field
-    with the given rewriting rules. ``rules`` must be a list of
-    (pattern, replacement) pairs.
+    with the given rewriting rules.
+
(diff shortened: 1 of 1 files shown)


sampsyo:
Detecting and preventing recursion seems quite hard in general! There is a chance that the only way to win is not to play, i.e., just document that you shouldn’t do that and let the thing crash in an ugly way if you do. I don’t love that solution but a straightforward way to detect recursion seems like it may not exist.

smichel17:
I agree that preventing recursion in general is hard-to-impossible, and would likely require a major rewrite of the template system. I don't think we should do that.

At the same time, it would be pretty unfortunate if we can't prevent recursion just within this one plugin itself. It would prevent most of the more interesting uses for this, like both of the examples above (the second example was actually miswritten yesterday; I just edited it to fix it).

The approach I was *trying* to take here is to take a snapshot of the template functions before this plugin adds the rewrite templates, and only use those when applying the template. This wouldn't prevent infinite recursion if there is another plugin which modifies the templates (and calls back to this plugin), but it would catch the most important cases, I think… if it worked! For some reason this approach didn't stop the recursion— I'm not sure why.

smichel17:
This has fallen very far down my priority list, so I'm unlikely to return to it any time soon. I'd still like to do it some day, but there's no sense in leaving this draft PR open until then. Anyone who wants to try to fix it, feel free.

PR #4179, opened 2021-12-01
Add default for unimported.ignore_subdirectories

tummychow:
## Description

Adds a default value for the new option introduced in #4086 . Without this, the plugin requires you to set it:

```
$ beet -vv unimported
user configuration: /home/sjung/.config/beets/config.yaml
data directory: /home/sjung/.config/beets
plugin paths:
inline: adding item field disc_track_padded
Sending event: pluginload
library database: /home/sjung/music/beets.db
library directory: /home/sjung/music
Sending event: library_opened
configuration error: unimported.ignore_subdirectories not found
```

### DIFF
--- a/beetsplug/unimported.py
+++ b/beetsplug/unimported.py
@@ -32,7 +32,8 @@ class Unimported(BeetsPlugin):
         super().__init__()
         self.config.add(
             {
-                'ignore_extensions': []
+                'ignore_extensions': [],
+                'ignore_subdirectories': []
             }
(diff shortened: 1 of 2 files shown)


sampsyo:
Thank you!! Would you mind adding a quick changelog entry?

tummychow:
Oh, of course - changelog added.

sampsyo:
Many thanks once again for sorting this out so neatly.

PR #4182, opened 2021-12-03
beetsplug/web: fix translation of query path

sumpfralle:
The routing map translator `QueryConverter` was misconfigured:
* decoding (parsing a path): splitting with "/" as tokenizer
* encoding (translating back to a path): joining items with "," as separator

Instead the encoding should have used the same delimiter (the slash) for the backward conversion.

How to reproduce:
* query: `/album/query/albumartist::%5Efoo%24/original_year%2B/year%2B/album%2B`
* resulting content in parsed argument `queries` in the `album_query` function:
    * previous (wrong): `['albumartist::^foo$,original_year+,year+,album+']`
    * new (correct): `['albumartist::^foo$', 'original_year+', 'year+', 'album+']`

I am not sure, why the issue was not noticed before. Maybe `werkzeug`s handling of routing maps has changed.

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -261,7 +261,7 @@ class QueryConverter(PathConverter):
                 for query in queries]
 
     def to_url(self, value):
-        return ','.join([v.replace(os.sep, '\\') for v in value])
+        return '/'.join([v.replace(os.sep, '\\') for v in value])
 
 
 class EverythingConverter(PathConverter):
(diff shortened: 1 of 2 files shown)


sampsyo:
I think this looks right—I imagine it was just broken before! Thanks for tracking this down.

This seems related to https://github.com/beetbox/beets/pull/3567, so perhaps @nmeum would be able to take a look to make sure we're not doing anything wrong?

And in the mean time, @sumpfralle, any chance you could add a quick changelog entry describing the fix?

sumpfralle:
I added a changelog entry.
Thank you for your time!

sampsyo:
Looks good; thank you! Please feel free to weigh in if we are doing something wrong, @nmeum.

PR #4184, opened 2021-12-03
clarify `-a` option for `beet modify`

dschrempf:
See https://github.com/beetbox/beets/discussions/4172.

I think the confusion arises because the documentation refers to the
query. That is, when `-a` is given, albums are queried, not tracks. This
is especially clear when using `beet list`, because then it truly lists
"albums instead of items". However, for other commands, the distinction
between what is queried and what is acted on should be made more clear.

This PR fixes the section for `modify`, but there are more questions:

- `remove` command: The documentation states that it acts on albums
instead of individual tracks. I guess we should also amend that? I think
the complete album including the tracks is deleted, or is that not true?

- `move` command: I think the same is true for this command. If `-a` is
  given, the queried albums including all tracks are moved.

- `update` command: The `-a` flag is not explained here.

### DIFF
--- a/docs/reference/cli.rst
+++ b/docs/reference/cli.rst
@@ -263,11 +263,11 @@ artist="Tom Tom Club"`` will change the artist for the track "Genius of Love."
 To remove fields (which is only possible for flexible attributes), follow a
 field name with an exclamation point: ``field!``.
 
-The ``-a`` switch operates on albums instead of individual tracks. Without
-this flag, the command will only change *track-level* data, even if all the
-tracks belong to the same album. If you want to change an *album-level* field,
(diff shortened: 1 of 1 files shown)


sampsyo:
Nice; this wording looks like a solid improvement to me!

I guess a subtlety that's worth acknowledging in the rest of these commands you've mentioned is that the text is trying to imply that albums _include_ tracks. That is, when we say that "`remove -a` removes albums instead of tracks," what we mean is "`remove -a` removes _entire_ albums instead of _merely_ individual tracks." We can definitely clarify this further, but I thought I'd attempt to explain the original reasoning behind this "albums instead of tracks" wording that we've used several times.

dschrempf:
I see. Thanks for the clarification!

PR #4190, opened 2021-12-08
Limit plugin

patrick-nicholson:
## Description

The `limit` plugin adds the command `lslimit` and the query prefix `<` for limiting query results to the head or tail. (Tail is supported by command only.) 

The impetus for this change was for `smartplaylists` with size limits. Such a feature has been requested before and there looks to have long been talk about this behavior through pagination (see #750).

### DIFF
--- a/beetsplug/limit.py
+++ b/beetsplug/limit.py
@@ -11,6 +11,15 @@
 # The above copyright notice and this permission notice shall be
 # included in all copies or substantial portions of the Software.
 
+"""Adds head/tail functionality to list/ls.
+
+1. Implemented as `lslimit` command with `--head` and `--tail` options. This is
+   the idiomatic way to use this plugin.
+2. Implemented as query prefix `<` for head functionality only. This is the
+   composable way to use the plugin (plays nicely with anything that uses the
+   query language).
+"""
+
 from beets.dbcore import FieldQuery
 from beets.plugins import BeetsPlugin
 from beets.ui import Subcommand, decargs, print_
--- a/beetsplug/limit.py
+++ b/beetsplug/limit.py
@@ -28,7 +28,7 @@ from itertools import islice
 
 
 def lslimit(lib, opts, args):
-    """Query command with head/tail"""
+    """Query command with head/tail."""
 
     if (opts.head is not None) and (opts.tail is not None):
         raise ValueError("Only use one of --head and --tail")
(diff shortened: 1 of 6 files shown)


docs/plugins/limit.rst
```diff
@@ -0,0 +1,52 @@
...
+So why does the query prefix exist? Because it composes with any other 
+query-based API or plugin (see :doc:`/reference/query`). For example, 
+you can use the query prefix in ``smartplaylists`` (see :doc:`/plugins/
+smartplaylists`) to limit the number of tracks in a smart playlist for 
+applications like most played and recently added.
+
+Configuration
+=============
+
+Enable the ``last`` plugin in your configuration (see
```
sampsyo, line 37:
```suggestion
Enable the ``limit`` plugin in your configuration (see 
```

beetsplug/limit.py
```diff
@@ -0,0 +1,95 @@
...
+    query = decargs(args)
+    if opts.album:
+        objs = lib.albums(query)
+    else:
+        objs = lib.items(query)
+    
+    if head:
+        objs = islice(objs, head)
+    elif tail:
+        objs = deque(objs, tail)
```
sampsyo, line 39: Very clever use of `deque`. 💖

beetsplug/limit.py
```diff
@@ -0,0 +1,95 @@
...
+lslimit_cmd.parser.add_all_common_options()
+lslimit_cmd.func = lslimit
+
+
+class LsLimitPlugin(BeetsPlugin):
+    def commands(self):
+        return [lslimit_cmd]
+
+
+class HeadPlugin(BeetsPlugin):
```
sampsyo, line 73: It's probably not necessary to group this into two separate `BeetsPlugin` objects. The typical way to do this (in most plugins) is to just put everything into one class—i.e., this latter plugin could have a `commands` method as well.

sampsyo (commented):
Wow; this is super neat! We have always wanted to support limits in the internal database engine (cf #750), but this plugin will do quite nicely as a stopgap. (And perhaps as a prototyping area for how the feature might work when given first-class support eventually.)

I have a few comments within, and the style checker has a few comments of its own, but I think this is close to ready.

sampsyo (approved):
All looking great!! Thanks again for your attentiveness to detail here. I think this is good to merge whenever you deem it ready.

patrick-nicholson:
Thanks @sampsyo!

patrick-nicholson:
Dangit I clicked without reading

PR #4191, opened 2021-12-08
Updates docstrings and some comments in library.py

katecoded:
## Description

- Updates existing docstrings to more closely follow google's docstring format (for consistency). 
- Adds a docstring to the Items class.
- Fixes a few typos in comments/docstrings.
- Updates some block comments to follow the PEP 8 style guide.

All of these changes are in library.py.

## To Do

- [ ] Documentation (not applicable)
- [ ] Tests (not applicable - no actual code was changed)
- [ ] Changelog entry

Does this kind of change require a changelog entry? I am happy to add one if needed. Otherwise, please let me know what you think!

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -53,7 +53,7 @@ class PathQuery(dbcore.FieldQuery):
     """
 
     def __init__(self, field, pattern, fast=True, case_sensitive=None):
-        """Creates a path query.
+        """Create a path query.
 
         `pattern` must be a path, either to a file or a directory.
 
@@ -80,7 +80,7 @@ class PathQuery(dbcore.FieldQuery):
 
     @classmethod
     def is_path_query(cls, query_part):
-        """Tries to guess whether a unicode query part is a path query.
+        """Try to guess whether a unicode query part is a path query.
 
         Condition: separator precedes colon and the file exists.
         """
@@ -152,7 +152,7 @@ class PathType(types.Type):
     model_type = bytes
 
     def __init__(self, nullable=False):
-        """Creates a path type object.
+        """Create a path type object.
 
         `nullable` controls whether the type may be missing, i.e., None.
         """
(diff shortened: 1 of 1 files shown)


beets/library.py
```diff
@@ -283,21 +287,24 @@ def key(x):
...
 class FileOperationError(Exception):
     """Indicates an error when interacting with a file on disk.
+
     Possibilities include an unsupported media type, a permissions
     error, and an unhandled Mutagen exception.
     """
 
     def __init__(self, path, reason):
-        """Create an exception describing an operation on the file at
+        """Creates an exception describing an operation on the file at
```
sampsyo, line 296: I notice that Google guide explicitly says to use descriptive verbs, which directly contradicts PEP 257's recommendation to use an imperative mood. Weird! That is going to be a _very_ hard habit for me to break, personally… and it even makes me question whether we really do want to embrace the Google style wholeheartedly.

sampsyo (commented):
This all looks really great; thank you for cleaning up so many docstrings! There is one piece that left me a little worried, though: personally, I have a somewhat averse reaction to Google's recommendation for verb tense. 😬 I'd be interested in others' opinions, but maybe we could consider sticking with Google's style *except* for this anti-PEP-257 aspect that I would find hard to adopt myself?

beets/library.py
katecoded: I would be fine with going back and changing the tense if that would be preferred! Just let me know if that is something I should do - I can go with whatever style is wanted.

sampsyo: Sure! Why don't we go back to imperative verbs while we discuss a broader policy elsewhere?

katecoded:
I have now altered the docstrings to use imperative-style language.

sampsyo (approved):
Awesome! Thank you again for this cleanup & consistency enforcement!

PR #4197, opened 2021-12-12
kodiupdate: Support multiple instances

### LINKED ISSUE #4101: kodiupdate: support updating multiple Kodi instances
Pretty straitforward ask, it would be nice if when running beets using the Kodi update client that you could enter multiple clients

syntax would be something to the tune of ...

kodi:
    host: x.x.x.x
    port: 8080
    user: user
    pwd: password
kodi:
    host: x.x.x.x
    port: 8080
    user: user
    pwd: password
kodi:
    host: x.x.x.x
    port: 8080
    user: user
    pwd: password
can this please be implemented

ybnd:
## Description

Fixes #4101

* Read `kodi` configuration as an array and handle all configured instances.
* Backwards compatible with previous config (non-array single instance)
* When failing to connect to/update an instance, log and move on to the next one instead of bailing at the first failure

## To Do

### DIFF
--- a/beetsplug/kodiupdate.py
+++ b/beetsplug/kodiupdate.py
@@ -54,11 +54,12 @@ class KodiUpdate(BeetsPlugin):
         super().__init__()
 
         # Adding defaults.
-        config['kodi'].add({
+        config['kodi'].add([{
             'host': 'localhost',
             'port': 8080,
             'user': 'kodi',
-            'pwd': 'kodi'})
+            'pwd': 'kodi'
+        }])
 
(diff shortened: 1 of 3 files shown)


sampsyo (commented):
Looks good overall! I note that the old configuration code here was not very type-safe, i.e., it uses Confuse's plain old `.get()` instead of `as_str()` and friends to enforce specific types. The new strategy here would make this somewhat more complicated to add in the future, but at least it's no worse than the status quo.

Nice work!

PR #4198, opened 2021-12-18
fix spotify pagination

### LINKED ISSUE #1480: On OS X, `beet config -e` doesn't work for initial setup by default
By default (when `$EDITOR` is not set), the `beet config -e` command will try to use OS X's `open` command to edit the config file on that platform. Unfortunately, that command doesn't work when the file does not exist—when setting up beets for the first time, for example.

We may need to run `touch` first to create the file, and then use `open -t` to get the default text editor.

Background in a mailing-list thread:
https://groups.google.com/d/topic/beets-users/FDa_3fO3sBM/discussion

### LINKED ISSUE #4180: spotify: Use pagination to fetch beyond the first 50 tracks of a release
Not sure if this is a problem related to the plugin, or if the Spotify API doesn't serve more than 50 tracks per release. The other tracks are reported "Unmatched" during import.

dschrempf:
Basically, keep fetching tracks until there are no more available for
the specified album.

Fixes #4180 .

I tested this by locally importing some albums, but please let me know if you want more extended tests. I am not sure how to write them, though.

EDIT: Fix issue number.

### DIFF
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -194,9 +194,16 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
                 )
             )
 
+        tracks_data = album_data['tracks']
+        tracks_items = tracks_data['items']
+        while tracks_data['next']:
+            tracks_data = self._handle_response(requests.get,
+                                                tracks_data['next'])
+            tracks_items.extend(tracks_data['items'])
+
         tracks = []
         medium_totals = collections.defaultdict(int)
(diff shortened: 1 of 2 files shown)


sampsyo (commented):
Looks perfect to me! :tada: Could you add a quick changelog entry to this effect?

It also looks like the style checker is asking for one line to be wrapped onto two.

dschrempf:
I satisfied the linter and added a sentence to the change log!

sampsyo:
Yay! Perfect! Thank you for sorting this out.

PR #4199, opened 2021-12-18
Allow to configure which fields are used to find duplicates

jcassette:
## Description

This adds an option `duplicate_keys` which allows to configure the behavior of the importer when it searches for duplicates.
For example, one could add the `format` field so that the importer will always keep albums that have different formats.

Can you review this please? I still have to implement it for singletons.
Also what do you think about allowing to use a Python function in `duplicate_action` for full customization?

## To Do

- [x] Documentation
- [x] Changelog
- [x] Tests


beets/importer.py
```diff
@@ -533,6 +533,17 @@ def chosen_ident(self):
...
             return (self.match.info.artist, self.match.info.album)
 
+    def chosen_info(self):
+        """Returns a dictionnary of metadata about the current choice.
+        May only be called when the choice flag is ASIS or RETAG
+        (in which case the data comes from the files' current metadata)
+        or APPLY (in which case the data comes from the choice).
+        """
+        if self.choice_flag in (action.ASIS, action.RETAG):
+            return self.cur_info
```
sampsyo, line 543: I wonder if this might actually be easier to understand/maintain if it called `current_metadata` right here instead of using the stored field (which would avoid the need to compute `current_metadata` in the `lookup_candidates` stage).

docs/changelog.rst
```diff
@@ -10,6 +10,7 @@ New features:
 
 * :doc:`/plugins/kodiupdate`: Now supports multiple kodi instances
   :bug:`4101`
+* Allow to configure which fields are used to find duplicates
```
sampsyo, line 13: It's nice to link to the docs for the new option.

beets/importer.py
```diff
@@ -675,10 +688,16 @@ def find_duplicates(self, lib):
 
         duplicates = []
         task_paths = {i.path for i in self.items if i}
-        duplicate_query = dbcore.AndQuery((
-            dbcore.MatchQuery('albumartist', artist),
-            dbcore.MatchQuery('album', album),
-        ))
+        keys = config['import']['duplicate_keys'].as_str().split()
+        info = self.chosen_info().copy()
```
sampsyo, line 692: We could even make `chosen_info` itself call `copy()` or `dict()` to make its contract neater: i.e., it returns a temporary/disposable dict.

beets/importer.py
```diff
@@ -675,10 +688,16 @@ def find_duplicates(self, lib):
...
+        info['albumartist'] = artist
```
sampsyo, line 693: It's not entirely clear to me here why `info` needs to be updated with the results from `chosen_ident`… maybe this should be done inside `chosen_info`? In fact, maybe we want to remove `chosen_ident` altogether if we're replacing it with `chosen_info`?

beets/importer.py
```diff
@@ -675,10 +688,16 @@ def find_duplicates(self, lib):
...
+        album = library.Album(None, **info)
```
sampsyo, line 694: It's also not entirely clear to me why we need to construct this `Album` object. Is there something that doesn't work about looking directly in `info`?

beets/importer.py
```diff
@@ -675,10 +688,16 @@ def find_duplicates(self, lib):
...
+        subqueries = []
+        for key in keys:
+            value = album.get(key)
+            fast = key in library.Album.item_keys
```
sampsyo, line 698: This should probably use the *album-level* keys, not the item-level keys (`item_keys`), because we're doing an album-level query.

It's possible it would be a good idea to introduce some kind of utility method on `Album` and `Item` just for constructing `MatchQuery` objects that correctly set the `fast` flag…

docs/reference/config.rst
```diff
@@ -670,6 +670,16 @@ with the ``-a`` flag to the :ref:`import-cmd` command.)
 
 Default: ``yes``.
 
+.. _duplicate_keys:
+
+duplicate_keys
+~~~~~~~~~~~~~~
+
+The fields used to find duplicates in import task.
+If several items have the same value for each key, they will be considered duplicates.
```
sampsyo, line 679: It occurs to me here that we search for duplicates for both albums and items, and maybe we want the list of fields to be configurable separately.

sampsyo (commented):
Nice; thank you for getting this started! Overall, this seems like a good direction—we can generalize `chosen_ident` to include all the metadata we could ever want, and then construct a `MatchQuery` using a chosen set of field names. :+1: Nice design!

This is a tricky part of the code overall to tweak, so I have several thoughts within about how to carry things out.

docs/changelog.rst
jcassette: Ok I thought this was for plugins only

beets/importer.py
jcassette: It is to allow using flexible attributes, I added a comment

beets/importer.py
```diff
@@ -533,6 +533,17 @@ def chosen_ident(self):
...
+    def chosen_info(self):
+        """Returns a dictionnary of metadata about the current choice.
+        May only be called when the choice flag is ASIS or RETAG
+        (in which case the data comes from the files' current metadata)
+        or APPLY (in which case the data comes from the choice).
+        """
+        if self.choice_flag in (action.ASIS, action.RETAG):
+            return self.cur_info
+        elif self.choice_flag is action.APPLY:
+            return self.match.info
```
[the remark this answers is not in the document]
jcassette: I forgot that sorry

docs/reference/config.rst
jcassette: Actually it should use album keys only

jcassette:
Hi @sampsyo and thank you for your review
I have updated this with your comments

jcassette:
It seems I can't trigger the CI tests somehow...

sampsyo:
Awesome! I believe the problem is that the CI doesn't trigger when the branch has conflicts. I've resolved the conflict (in the changelog file) and the tests are running now.

allesmi:
This pull request would enable my use case where I want to import several releases of the same artist which are untitled. Other fields like the catalog number or year would identify the releases however.

@jcassette  After reading your pull request I tried to re-implement a very basic working example for the fields (artist, album, year).

I found out that I can import the same album again and again without getting the duplicate warning if one of the fields is None, like the year for example.

jcassette:
Hi @allesmi and thanks for trying this
Can you provide a test case which reproduces this issue?
I tried to set "flex" fields to None in `test_keep_when_extra_key_is_different` and the album was not imported.

jcassette:
Sorry about the noise - I got a bit confused with git
I have fixed the tests and implemented the feature for singletons
@sampsyo can you review this please ?
I think this PR is complete now

allesmi:
> Hi @allesmi and thanks for trying this Can you provide a test case which reproduces this issue? I tried to set "flex" fields to None in `test_keep_when_extra_key_is_different` and the album was not imported.

Unfortunately I do not have time to show this as code. Let me instead describe what I have manually tried with your most recent commit:
I have imported the album Ketsa - May Starlight Find You ([musicbrainz release](https://musicbrainz.org/release/25c9d201-f2a7-44b3-a1b3-74e75dec89c9), [download from Free Music Archive](https://freemusicarchive.org/music/Ketsa/May_Starlight_Find_You)). Note that musicbrainz does not have a year for this release.

My config file consists of these options for `duplicate_keys` and is other than that empty:

```
import:
  duplicate_keys:
    album: albumartist album year
```

The first import into an empty database completes successfully. Importing the album again does not warn me about the duplicate, does not ask me what to do with old and new files and instead imports as if it is not a duplicate. After several imports my library ends up like this:

```
$ tree ~/Music/Ketsa/ -d
/home/me/Music/Ketsa/
├── May Starlight Find You
├── May Starlight Find You [2]
└── May Starlight Find You [3]
```

It seems as if the `year` field is an issue. If I remove it from `duplicate_keys` and import the album again I get asked what I want to do with three old albums and one new album.

beets/importer.py
```diff
@@ -521,17 +521,18 @@ def skip(self):
...
-        (artist, title). May only be called when the choice flag is ASIS
-        or RETAG (in which case the data comes from the files' current
-        metadata) or APPLY (data comes from the choice).
+    def chosen_info(self):
+        """Return a dictionary of metadata about the current choice.
+        May only be called when the choice flag is ASIS or RETAG
+        (in which case the data comes from the files' current metadata)
+        or APPLY (in which case the data comes from the choice).
         """
+        assert(self.choice_flag in (action.ASIS, action.RETAG, action.APPLY))
```
sampsyo, line 530: I don't think the parentheses are necessary:
```suggestion
        assert self.choice_flag in (action.ASIS, action.RETAG, action.APPLY)
```

sampsyo: In fact, another (perhaps) way to catch this is to just put `assert False` at the end, after the `elif`.

beets/library.py
```diff
@@ -607,6 +607,20 @@ def from_path(cls, path):
         i.mtime = i.current_mtime()  # Initial mtime.
         return i
 
+    @classmethod
+    def construct_match_queries(cls, **info):
+        subqueries = []
+        for (key, value) in info.items():
```
sampsyo, line 613: I don't think the parentheses are necessary.
```suggestion
        for key, value in info.items():
```

sampsyo (commented):
Overall, this is looking great! I like the simplification, and packaging up the query construction has made things significantly more readable.

I'm sorry for being slow, but I'm still having some trouble understanding why `tmp_album` and `tmp_item` are necessary. You mentioned above (and in the comments) that this is to make things work with flexible attributes. But I don't quite see why this is necessary for flexible attributes. Couldn't something like `duplicates` and `construct_match_queries` work using only information from the `Album` and `Item` _classes_ rather from temporary instances? (Again, sorry for missing the insight here!)

sampsyo:
One last thing I forgot to mention: it would be great to avoid the duplication in the model classes. At least some of `duplicates` and `construct_match_queries` seem like they could be moved to `dbcore` to be available on all `Model` subclasses.

jcassette:
Thanks for the review @sampsyo ! I have made the changes you requested.

> I'm sorry for being slow, but I'm still having some trouble understanding why `tmp_album` and `tmp_item` are necessary. You mentioned above (and in the comments) that this is to make things work with flexible attributes. But I don't quite see why this is necessary for flexible attributes. Couldn't something like `duplicates` and `construct_match_queries` work using only information from the `Album` and `Item` _classes_ rather from temporary instances? (Again, sorry for missing the insight here!)

No problem, I probably didn't explain well.
If you checkout the first commit and run the test case of the second commit with you will see an error related to the flexible attribute.
Also `chosen_info` only returns values contained in the tags of the audio files, but for inline fields I think I understood that the values are computed in `Album` and `Item` classes, that's why I needed to create the temporary instances.
I haven't dug into it so much so I hope this helps, or let me know if you want to discuss this further.

sampsyo:
Computed (inline) fields are an interesting point! I guess creating a temporary object may be the easiest/lowest-effort way to obtain those, if people want to use them as deduplication keys.

I suppose there's an underlying confusion I have about flexible attributes, though: since we're dealing with newly-imported albums/items, how could they have flexible attributes on them in the first place? I thought we would _only_ be able to distinguish newly-imported albums based on the data either from disk or from the metadata source… maybe one of the metadata sources uses a flexible attribute? But if so, I thought that would be in the `chosen_info` dict already? I guess I'm missing something about the underlying problem here…

jcassette:
Sorry about the confusion, I had been using the term "flexible attributes" to actually mean "inline fieds"...
I may have confused because they seem to work similarly for database queries.

> since we're dealing with newly-imported albums/items, how could they have flexible attributes on them in the first place?

I was not trying to make this to work, only inline fields.
But maybe one album/item could already have flexible attributes if it is being reimported?
In this case, with the current code these attributes cannot be used in `duplicate_keys`.

> I thought we would _only_ be able to distinguish newly-imported albums based on the data either from disk or from the metadata source…

Yes, that's what is implemented. The data of `chosen_info` is provided by `autotag.current_metadata` (from disk) or `self.match.info` (from metadata source), and inline fields derived from that data are provided by the `Album`/`Item` classes.

sampsyo:
Got it; thanks for clarifying! Computed fields (e.g., from the `inline` plugin) seem like an interesting use case for this. I'm convinced, I think, that creating temporary objects is the easiest route to doing this.

I'll look in more detail at the low-level utilities here shortly…

sampsyo:
It's still very much on my to-do list to rearrange some of the low-level query manipulation stuff here… 😕

sampsyo:
Thanks for your (extreme) patience while I finally got around to giving this PR the attention it deserved. It was 99% there, but I wanted to simplify the low-level query utilities that found their way into the `library` and `dbcore` modules. I moved us toward some even simpler, broadly applicable utilities for constructing queries, which I can imagine using many other places throughout the codebase. I think this should also somewhat simplify the logic in the `importer` module itself.

Please let me know if you have any thoughts! Meanwhile, I'll merge this so it can go out in the upcoming release.

PR #4200, opened 2021-12-18
fix deezer pagination

dschrempf:
See #4180, and #4198.


sampsyo (approved):
Looks perfect! Nice work figuring out the API for this. 👏

PR #4192, opened 2021-12-08
Fixed `can't concat str to byte` inside move

catap:
## Description

`os.path.basename` returns the same type that was used.

So, I have no guarantee that I needed here: `b'.''` or `'.'`.

Let use something universal ;)

Closes #4168

PR #4209, opened 2021-12-27
Improved fix for Unicode mistakes in atomic move

### LINKED ISSUE #4168: Unicode error in new atomic move
### Problem

I’ve stumbled upon a bug when importing any file, seemingly related to the use of the `os.replace` function and the `import: move` option. using the copy option instead fixes the bug. According to adrian, it would be related to [Introduce atomic move and write of file](https://github.com/beetbox/beets/pull/4060)

```sh
  File "C:\Users\hydro\anaconda3\lib\site-packages\beets\util\__init__.py", line 496, in move
    os.replace(path, dest)
OSError: [WinError 17] The system cannot move the file to a different disk drive: '\\\\?\\C:\\Users\\hydro\\AppData\\Local\\Temp\\tmp02lvwxx1.flac' -> '\\\\?\\M:\\Musique lossless\\Ancient Methods\\2021 - The Jericho Remixes\\01 Swordplay (Hypnoskull vocal remix).flac'

  File "C:\Users\hydro\anaconda3\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\hydro\anaconda3\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\hydro\anaconda3\Scripts\beet.exe\__main__.py", line 7, in <module>
  File "C:\Users\hydro\anaconda3\lib\site-packages\beets\ui\__init__.py", line 1285, in main
    _raw_main(args)
  File "C:\Users\hydro\anaconda3\lib\site-packages\beets\ui\__init__.py", line 1272, in _raw_main
  File "C:\Users\hydro\anaconda3\lib\site-packages\beets\ui\commands.py", line 973, in import_func
  File "C:\Users\hydro\anaconda3\lib\site-packages\beets\ui\commands.py", line 943, in import_files
    session.run()
  File "C:\Users\hydro\anaconda3\lib\site-packages\beets\importer.py", line 340, in run
    pl.run_parallel(QUEUE_SIZE)
  File "C:\Users\hydro\anaconda3\lib\site-packages\beets\util\pipeline.py", line 446, in run_parallel
  File "C:\Users\hydro\anaconda3\lib\site-packages\beets\util\pipeline.py", line 358, in run
    self.coro.send(msg)
  File "C:\Users\hydro\anaconda3\lib\site-packages\beets\util\pipeline.py", line 170, in coro
  File "C:\Users\hydro\anaconda3\lib\site-packages\beets\importer.py", line 1566, in manipulate_files
    task.manipulate_files(
  File "C:\Users\hydro\anaconda3\lib\site-packages\beets\importer.py", line 757, in manipulate_files
    item.move(operation)
  File "C:\Users\hydro\anaconda3\lib\site-packages\beets\library.py", line 922, in move
    self.move_file(dest, operation)
  File "C:\Users\hydro\anaconda3\lib\site-packages\beets\library.py", line 821, in move_file
```

### Setup

* OS: Windows 10
* Python version: 3.9.7
* beets version: 1.6.0
* Turning off plugins made problem go away (yes/no): irrelevant

The relevant bit of my configuration (output of `beet config`) is:

```yaml
# Import options
import:
  write: yes
  # copy: yes
  move: yes
  resume: no
  autotag: yes
```

sampsyo:
The original problem was that this implementation of atomic moves, in #4060, used bytestrings incorrectly to invoke `mktemp`. This caused a regression that crashed beets when it tried to move files across filesystems (which, of course, actually involves a copy).

However, `mktemp` is deprecated, so this PR just avoids it altogether. Fortunately, the non-deprecated APIs in `tempfile` support all-bytes arguments.

Fixes #4168. Also closes #4192, which it supersedes.

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -495,13 +495,20 @@ def move(path, dest, replace=False):
     try:
         os.replace(path, dest)
     except OSError:
-        tmp = tempfile.mktemp(suffix='.beets',
-                              prefix=py3_path(b'.' + os.path.basename(dest)),
-                              dir=py3_path(os.path.dirname(dest)))
-        tmp = syspath(tmp)
+        # Copy the file to a temporary destination.
+        tmp = tempfile.NamedTemporaryFile(suffix=b'.beets',
+                                          prefix=b'.' + os.path.basename(dest),
+                                          dir=os.path.dirname(dest),
+                                          delete=False)
         try:
-            shutil.copyfile(path, tmp)
-            os.replace(tmp, dest)
+            with open(path, 'rb') as f:
+                shutil.copyfileobj(f, tmp)
+        finally:
+            tmp.close()
+
+        # Move the copied file into place.
+        try:
+            os.replace(tmp.name, dest)
             tmp = None
             os.remove(path)
         except OSError as exc:
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -496,8 +496,9 @@ def move(path, dest, replace=False):
         os.replace(path, dest)
     except OSError:
         # Copy the file to a temporary destination.
+        base = os.path.basename(bytestring_path(dest))
         tmp = tempfile.NamedTemporaryFile(suffix=b'.beets',
-                                          prefix=b'.' + os.path.basename(dest),
+                                          prefix=b'.' + base,
                                           dir=os.path.dirname(dest),
                                           delete=False)
         try:
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -486,32 +486,33 @@ def move(path, dest, replace=False):
                               (path, dest))
     if samefile(path, dest):
         return
-    path = syspath(path)
-    dest = syspath(dest)
-    if os.path.exists(dest) and not replace:
+    if os.path.exists(syspath(dest)) and not replace:
         raise FilesystemError('file exists', 'rename', (path, dest))
 
     # First, try renaming the file.
     try:
-        os.replace(path, dest)
+        os.replace(syspath(path), syspath(dest))
     except OSError:
         # Copy the file to a temporary destination.
-        base = os.path.basename(bytestring_path(dest))
-        tmp = tempfile.NamedTemporaryFile(suffix=b'.beets',
-                                          prefix=b'.' + base,
-                                          dir=os.path.dirname(dest),
-                                          delete=False)
+        basename = os.path.basename(bytestring_path(dest))
+        dirname = os.path.dirname(bytestring_path(dest))
+        tmp = tempfile.NamedTemporaryFile(
+            suffix=syspath(b'.beets', prefix=False),
+            prefix=syspath(b'.' + basename, prefix=False),
+            dir=syspath(dirname),
+            delete=False,
+        )
         try:
-            with open(path, 'rb') as f:
+            with open(syspath(path), 'rb') as f:
                 shutil.copyfileobj(f, tmp)
         finally:
             tmp.close()
 
         # Move the copied file into place.
         try:
-            os.replace(tmp.name, dest)
+            os.replace(tmp.name, syspath(dest))
             tmp = None
-            os.remove(path)
+            os.remove(syspath(path))
         except OSError as exc:
             raise FilesystemError(exc, 'move', (path, dest),
                                   traceback.format_exc())
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -479,9 +479,9 @@ def move(path, dest, replace=False):
     instead, in which case metadata will *not* be preserved. Paths are
     translated to system paths.
     """
-    if os.path.isdir(path):
+    if os.path.isdir(syspath(path)):
         raise FilesystemError(u'source is directory', 'move', (path, dest))
-    if os.path.isdir(dest):
+    if os.path.isdir(syspath(dest)):
         raise FilesystemError(u'destination is directory', 'move',
                               (path, dest))
     if samefile(path, dest):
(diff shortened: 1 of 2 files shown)


sampsyo:
Paging @catap, @LeoSebal, and @dertuxmalwieder to please try out this improved fix, which supersedes #4192.

dertuxmalwieder:
I am currently using the patch from #4192 to move files from my local hard disk to a NAS and it has not crashed for me. But I will try your patch and see if I notice any difference. :-)

catap:
@sampsyo the strange things that I’m used original code and it works fine :)

anyway, I run it on macOS and probably it is OS specific things

dertuxmalwieder:
This patch reintroduces the old bug:

      File "C:\Python311\Lib\site-packages\beets\util\__init__.py", line 500, in move
        prefix=b'.' + os.path.basename(dest),
               ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
    TypeError: can't concat str to bytes

sampsyo:
Thanks for trying it out, everybody! @dertuxmalwieder, do you have the complete traceback I could look at?

dertuxmalwieder:
I already undid the patch, I am currently running beets over my whole collection, that will take a while...
I'll come back to it these days. :-)

(If nobody else does...)

sampsyo:
Awesome; thanks! It's partially a stab in the dark, but I have pushed an additional tweak that should almost certainly avoid that kind of problem in the odd cases where it might arise.

dertuxmalwieder:
It seems to be hard to reproduce the crashes as there does not seem to be a pattern.

wisp3rwind:
> Awesome; thanks! It's partially a stab in the dark, but I have pushed an additional tweak that should almost certainly avoid that kind of problem in the odd cases where it might arise.

I've only had a very brief look at this (and I haven't read the issues and PRs leading up to this one); but I'm not sure this fix is fully consistent yet: You're now using `basename(bytestring_path(syspath(dest)))` (`syspath` might cause the problem if on Windows, since it yields `str`ings), whereas it might be more natural to do `syspath(basename(dest))`, i.e. convert using `syspath` only at the very edge to foreign API. Moreover, all arguments (`dir`, `suffix`, `prefix`) to `NamedTemporaryFile` should have the same type, which I think is not guaranteed right now. They probably all need to be passed through syspath (cf. [mkstemp](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp), which `NamedTemporaryFile` apparently delegates to).

sampsyo:
Ah, you're very right about that, @wisp3rwind. Thanks for the sanity check. Let me try simplifying things with `syspath` in mind.

sampsyo:
OK, let's give the above commit a try. The result may look a little ugly, but I think it's the right thing. The strategy is:

* I eliminated the somewhat messy up-front `syspath` conversion. Instead, we use `syspath` directly before handing off strings to stdlib APIs, as usual.
* To call `NamedTemporaryFile`, we first make sure we have bytes everywhere for the three arguments using `bytestring_path`.
* Then, for Windows, we call `syspath` on *all three* arguments. This way, the arguments will all be `str` objects on Windows and all be `bytes` objects on Unix, as god intended. We need to use `prefix=False` for the two fragments (incomplete paths) that are passed as arguments. But using `prefix=True` (the default) for the `dir` argument means that we can still handle long paths on Windows.

wisp3rwind:
This seems to be a stubborn problem (IIRC, I'm at least partially responsible for reviewing the original PR...), thanks for dealing with it!

> The result may look a little ugly, but I think it's the right thing.

I see what you mean, but I also can't think of a way to improve things.

> * To call `NamedTemporaryFile`, we first make sure we have bytes everywhere for the three arguments using `bytestring_path`.

I would claim that this is superfluous, if anything is calling `move()` with non-`bytestring` paths, that should be treated as a bug, right? In any case, the additional `bytestring_path`s shouldn't hurt.

> * Then, for Windows, we call `syspath` on _all three_ arguments. This way, the arguments will all be `str` objects on Windows and all be `bytes` objects on Unix, as god intended. We need to use `prefix=False` for the two fragments (incomplete paths) that are passed as arguments. But using `prefix=True` (the default) for the `dir` argument means that we can still handle long paths on Windows.

Sounds correct to me :+1:

wisp3rwind:
While we're at it, the two `os.path.isdir()` checks at the function's entrance should also be garnished with `syspath()`, shouldn't they?

sampsyo:
> I would claim that this is superfluous, if anything is calling `move()` with non-`bytestring` paths, that should be treated as a bug, right? In any case, the additional `bytestring_path`s shouldn't hurt.

Yeah, you're absolutely right that this should be considered a bug. I was just trying to "defensively program" here: to tolerate those bugs instead of crashing, just because I'm not terribly confident that every client does the right thing (given the other tracebacks I've seen). It would be good, in some future iteration, to back away from this and actually enforce this requirement (with `assert isinstance(path, bytes)` or similar) to see what breaks…

> While we're at it, the two `os.path.isdir()` checks at the function's entrance should also be garnished with `syspath()`, shouldn't they?

Yep, good call. I'll add these.

sampsyo:
Alright, with that little bit of evolution out of the way, I would be ever so grateful if the interested parties could give this a shot! I'm reasonably confident in it, however, so I'd like to merge this soonish if there are no obvious red flags.

nopeinomicon:
> Alright, with that little bit of evolution out of the way, I would be ever so grateful if the interested parties could give this a shot! I'm reasonably confident in it, however, so I'd like to merge this soonish if there are no obvious red flags.

It seems to be working on my end! I was having this issue before and your fix works. Thank you much!

Gorasa:
Will there be a bugfix version with this fix any time soon?

PR #4226, opened 2022-01-08
Makes the import converter respect the quiet and pretend flags

mousecloak:
## Description

Makes the import converter respect the quiet and pretend flags.

When the `delete_originals` is set, beets prints the following, regardless of the presence of the quiet parameter:

```
convert: Removing original file /path/to/file.ext
```

This PR ensures that the log is only printed when quiet is not present.

In addition, I noticed that some tests were annotated with the following:
```python
@unittest.skipIf(sys.platform, 'win32')  # FIXME: fails on windows
```

I believe the first parameter to this annotation is a condition and the second a description, which means the tests were always skipped regardless of platform. I corrected these annotations to be

```python
@unittest.skipIf(sys.platform == 'win32', 'win32')  # FIXME: fails on windows
```

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -28,6 +28,10 @@ Bug fixes:
   ``r128_album_gain`` fields was changed from integer to float to fix loss of
   precision due to truncation.
   :bug:`4169`
+* :doc:`/plugins/convert`: Files are no longer converted when running import in
+  ``--pretend`` mode.
+* :doc:`/plugins/convert`: Deleting the original files during conversion no
+  longer logs output when the ``quiet`` flag is enabled.
 
 For packagers:
 
--- a/test/test_convert.py
+++ b/test/test_convert.py
@@ -107,7 +107,8 @@ class ImportConvertTest(unittest.TestCase, TestHelper):
         item = self.lib.items().get()
         self.assertFileTag(item.path, 'convert')
 
-    @unittest.skipIf(sys.platform, 'win32')  # FIXME: fails on windows
+    # FIXME: fails on windows
+    @unittest.skipIf(sys.platform == 'win32', 'win32')
     def test_import_original_on_convert_error(self):
         # `false` exits with non-zero code
         self.config['convert']['command'] = 'false'
--- a/test/test_convert.py
+++ b/test/test_convert.py
@@ -122,9 +122,28 @@ class ImportConvertTest(unittest.TestCase, TestHelper):
         self.importer.run()
         for path in self.importer.paths:
             for root, dirnames, filenames in os.walk(path):
-                self.assertTrue(len(fnmatch.filter(filenames, '*.mp3')) == 0,
-                                'Non-empty import directory {}'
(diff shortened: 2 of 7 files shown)


test/test_convert.py
```diff
@@ -107,7 +107,7 @@ def test_import_converted(self):
         item = self.lib.items().get()
         self.assertFileTag(item.path, 'convert')
 
-    @unittest.skipIf(sys.platform, 'win32')  # FIXME: fails on windows
+    @unittest.skipIf(sys.platform == 'win32', 'win32')  # FIXME: fails on windows
```
sampsyo, line 110: Good catch!

sampsyo (commented):
Awesome; this all seems great! Thanks for catching the erroneous test skippage. 😄 Would you mind adding a quick changelog entry about the new behavior of `pretend` and `quiet`?

mousecloak:
Done! Thanks for reviewing this.

docs/changelog.rst
```diff
@@ -28,6 +28,8 @@ Bug fixes:
   ``r128_album_gain`` fields was changed from integer to float to fix loss of
   precision due to truncation.
   :bug:`4169`
+* :doc:`/plugins/convert`: The ``pretend`` and ``quiet`` flags are now respected
+  when ``delete_originals`` is enabled.
```
sampsyo, line 32: Hmm, maybe I'm missing something, but I would have expected this to say "when `auto` is enabled" or "when automatically converting on import"? I think we're just changing what happens when doing the convert-on-import thing—regardless of `delete_originals`, if I'm not mistaken.

mousecloak: Yes you're right, sorry. I'm conflating two different things here.

I originally started this PR because I saw `delete_originals` still print logs despite `quiet` being enabled. While fixing that I noticed that `pretend` was also not respected - as you point out, by the converter in general, regardless of `delete_originals` - so I figured I'd fix that too.

I'll make these two separate changelog entries to avoid the confusion.

sampsyo:
Looks great!! Many thanks once again.

wisp3rwind:
Do the changes here really make sense, though? At a first sight, it seems to me that:
- `pretend` as used here is only the CLI option to `beet convert`. That can never be set in the importer in `convert_on_import`, right? Moreover, when running `import --pretend`, the convert plugin should not even be invoked, since only the `log_files` pipeline stage is run, and none of the plugin stages (see `beets.importer.ImportSession.run()`). In which exact situations did you observe unexpected behavior that led to this change?
- I would say that the log message should not be removed entirely, but only changed to `debug` level if quiet is set. We do want all the details in `--verbose` mode.

sampsyo:
Thanks for bringing this back to reality, @wisp3rwind. I don't know how I missed the thing about `--pretend` mode, and you're right about changing the verbosity. I'm going to put together a quick PR to rectify these things.

Meanwhile, @mousecloak, maybe you can clarify what you were getting at with the import/pretend interaction?

Issue #2617, opened 2017-06-28
torrent clients support (transmission-daemon)


anarcat:
### Problem

Some of the files managed by beets may be seeded on torrent files to share with your friends.

If beets modifies those files, this will  break the checksums from the torrent files, and some torrent clients will obviously be confused by this. Some will just warn the user, others won't do anything. In general, one can assume clients will start re-downloading the affected chunks, overwriting the changes performed by beets.

### Setup

N/A

### Proposed solution

I envision a plugin that would either deny writes on torrent-managed files or remove files from the torrent client when beets modifies them. This behavior could obviously be customizable from the configuration file.

The same question as with #2616 exists here: I assume I would need to hook into `before_item_moved` and `write` to do this. But i'm not clear how or if i can deny moving items at all?

With transmission, my current client of choice, there's a commandline tool ([transmission-remote](https://manpages.debian.org/stretch/transmission-cli/transmission-remote.1.en.html)) that can be used to manipulate torrent files (add/remove/list etc). This is backed by an [RPC](https://trac.transmissionbt.com/wiki/rpc) interface that's basically JSON over HTTP which should be fairly easy to use.

sampsyo:
Interesting idea! Currently, I think most people just disable `write`, `move`, and `copy` to leave all the files in place. Would that suffice for your case?

anarcat:
On 2017-06-29 20:36:52, Adrian Sampson wrote:
> Interesting idea! Currently, I think most people just disable `write`, `move`, and `copy` to leave all the files in place. Would that suffice for your case?

that is exactly what i do now, but i would like to fix tags for files
that are *not* covered by torrent files at least then make a choice for
those that are: if the change is minor, leave it alone, but if the tags
are too messed up, remove it from the torrent.

sampsyo:
Interesting. Maybe one strategy, then, would be to leave the `write` option off and explicitly use `beet write` after the fact to update non-torrenting files?

If you're interested in putting together a plugin, yes, the `write` event does seem like the right one to use. You should be able to (conditionally) undo any updates—taking a look at the `zero` plugin might be a good place to start.

anarcat:
does `write` cover the `copy`, `move`, etc updates as well? i was wondering if i could interrupt move operations with `before_item_moved` as well...

sampsyo:
No, `write` is just about updating tags. You might indeed need to hook into other events if you want to prevent moving files too.

anarcat:
[...]

PR #4230, opened 2022-01-13
Draft: Allow plugins to inhibit file operations via hooks

anarcat:
This reflects the `try_write` approach will handles write failures
elegantly, by logging the error and continuing. We do the same with
`move`, `art_set` and `move_art`.

We also handle exceptions on `before_item_moved` and `art_set` plugins
hooks, the latter of which is moved *before* the file operations to
remain consistent with other hook configurations.

That might be a mistake and API-breaking change, another approach
would be to have a new `before_art_set` hook instead.

We also introduce a new hook (`move_art`) for those operations as
well.

The point of this patch is to make it possible for plugins to send a
signal (through the already FileOperationError exception) to callers
that it should skip a specific item or artwork.

This is essential to allow beets to better integrate with other
utilities like bittorrent clients which may rewrite those files. The
rationale here is that some music collections will have *parts* of
them managed by such clients in which case we should be careful not to
overwrite or move those files.

Operations like copy or hardlink are not handled by this, for that
reason. We may also want to do proper error handling for those as
well, that said, but that seems out of scope for this specific
issue (#2617).

## To Do

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -747,14 +747,14 @@ class ImportTask(BaseImportTask):
                 if (operation != MoveOperation.MOVE
                         and self.replaced_items[item]
                         and session.lib.directory in util.ancestry(old_path)):
-                    item.move()
-                    # We moved the item, so remove the
-                    # now-nonexistent file from old_paths.
-                    self.old_paths.remove(old_path)
+                    if item.try_move():
+                        # We moved the item, so remove the
+                        # now-nonexistent file from old_paths.
+                        self.old_paths.remove(old_path)
                 else:
                     # A normal import. Just copy files and keep track of
                     # old paths.
-                    item.move(operation)
+                    item.try_move(operation)
 
             if write and (self.apply or self.choice_flag == action.RETAG):
                 item.try_write()
--- a/beets/library.py
+++ b/beets/library.py
@@ -826,8 +826,12 @@ class Item(LibModel):
         if not util.samefile(self.path, dest):
             dest = util.unique_path(dest)
         if operation == MoveOperation.MOVE:
-            plugins.send("before_item_moved", item=self, source=self.path,
-                         destination=dest)
+            try:
+                plugins.send("before_item_moved", item=self, source=self.path,
+                             destination=dest)
+            except FileOperationError as exc:
+                log.error("{0}", exc)
+                return
             util.move(self.path, dest)
             plugins.send("item_moved", item=self, source=self.path,
                          destination=dest)
@@ -948,6 +952,19 @@ class Item(LibModel):
         # Prune vacated directory.
         if operation == MoveOperation.MOVE:
             util.prune_dirs(os.path.dirname(old_path), self._db.directory)
+
+    def try_move(self, *args, **kwargs):
+        """Call `move()` but catch and log `FileOperationError`
+        exceptions.
+
+        Return `False` an exception was caught and `True` otherwise.
+        """
+        try:
+            self.move(*args, **kwargs)
+            return True
+        except FileOperationError as exc:
+            log.error("{0}", exc)
+            return False
 
     # Templating.
 
@@ -1195,6 +1212,11 @@ class Album(LibModel):
         if new_art == old_art:
             return
 
+        try:
+            plugins.send('move_art', album=self)
+        except FileOperationError as exc:
+            log.error("{0}", exc)
+            return
         new_art = util.unique_path(new_art)
         log.debug('moving album art {0} to {1}',
                   util.displayable_path(old_art),
@@ -1239,8 +1261,8 @@ class Album(LibModel):
         # Move items.
         items = list(self.items())
         for item in items:
-            item.move(operation, basedir=basedir, with_album=False,
-                      store=store)
+            item.try_move(operation, basedir=basedir, with_album=False,
+                          store=store)
 
         # Move art.
         self.move_art(operation)
@@ -1327,6 +1349,11 @@ class Album(LibModel):
             self.artpath = path
             return
 
+        try:
+            plugins.send('art_set', album=self)
+        except FileOperationError as exc:
+            log.error("{0}", exc)
+            return
         # Normal operation.
         if oldart == artdest:
             util.remove(oldart)
@@ -1337,7 +1364,6 @@ class Album(LibModel):
             util.move(path, artdest)
         self.artpath = artdest
 
-        plugins.send('art_set', album=self)
 
     def store(self, fields=None):
         """Update the database with the album information.


wisp3rwind:
I haven't thought about this in depth, but here are some initial comments. Generally, I think one point we need to pay attention to is never to update the database with new paths when the operation is actually skipped (seems to be the case for now).

> This reflects the `try_write` approach will handles write failures elegantly, by logging the error and continuing. We do the same with `move`, `art_set` and `move_art`.

As explained in more detail below, I think we should really distinguish actual errors, and a plugin saying "Let's skip this".

> We also handle exceptions on `before_item_moved` and `art_set` plugins hooks, the latter of which is moved _before_ the file operations to remain consistent with other hook configurations.
> 
> That might be a mistake and API-breaking change, another approach would be to have a new `before_art_set` hook instead.

Yes, we should not move that hook. It'd be much more consistent to have the `before_art_set` hook, and also think about it's precise design, cf. below.

> We also introduce a new hook (`move_art`) for those operations as well.
> 
> The point of this patch is to make it possible for plugins to send a signal (through the already FileOperationError exception) to callers that it should skip a specific item or artwork.

I think this API design is extremely confusing, since no real `FileOperationError` is happening, I'd much prefer to use something here that clearly means "skip this move operation". Plugin event handlers _can_ return values, so one alternative would be to have a special `MoveOperation.SKIP`, and use it along the lines of
```python
results = plugins.send("before_item_moved", item=self, source=self.path,
                 destination=dest)
if any(r == MoveOperation.SKIP for r in results):
    # Really not an error, a plugin _chose_ not to move the file.
    log.warning("Skipping move of item {self}")
    return
```
Incidentally, the `before_item_moved `event is placed in an unfortunate location, it would be much more useful to have an event _before_ the `MoveOperation` switch and pass the particular `MoveOperation` as an argument to this event. Maybe, we could add such an event (`before_item_moved_v2`?), and deprecate `before_item_moved`?

Relatedly, from #2617
> we also need to somehow pass information from the plugin to the caller. right now the plugins.send callers mostly ignore the return value, at least in the above hooks (there are rare exceptions elsewhere in the code). i wonder if we should return a truth value or raise an exception to signal a problem.
>
> maybe the safest would be to raise an exception from the plugin. older versions of beets would just fail to do the operation and crash (a possible acceptable failure mode, as in "fail closed"), newer versions could handle the exception and behave correctly (e.g. skip or abort, depending on the exception)?

I don't think that justifies designing a confusing new API. Plugins that require this functionality should simply declare a dependency on a sufficiently recent version of beets.


> Operations like copy or hardlink are not handled by this, for that reason. We may also want to do proper error handling for those as well, that said, but that seems out of scope for this specific issue (#2617).

By having `before_item_moved_v2` as detailed above, plugins could do that without further effort.

wisp3rwind:
I went ahead and rephrased the PR's title to my understanding of what we really want to achieve here.

anarcat:
On 2022-01-15 01:35:54, Benedikt wrote:
> I haven't thought about this in depth, but here are some initial comments. Generally, I think one point we need to pay attention to is never to update the database with new paths when the operation is actually skipped (seems to be the case for now).

That's a good point, it was a rough draft and it certainly misses some
key problems like that.

>> This reflects the `try_write` approach will handles write failures elegantly, by logging the error and continuing. We do the same with `move`, `art_set` and `move_art`.
>
> As explained in more detail below, I think we should really distinguish actual errors, and a plugin saying "Let's skip this".

Agreed.

>> We also handle exceptions on `before_item_moved` and `art_set` plugins hooks, the latter of which is moved _before_ the file operations to remain consistent with other hook configurations.
>> 
>> That might be a mistake and API-breaking change, another approach would be to have a new `before_art_set` hook instead.
>
> Yes, we should not move that hook. It'd be much more consistent to have the `before_art_set` hook, and also think about it's precise design, cf. below.

Agreed.

>> We also introduce a new hook (`move_art`) for those operations as well.
>> 
>> The point of this patch is to make it possible for plugins to send a signal (through the already FileOperationError exception) to callers that it should skip a specific item or artwork.
>
> I think this API design is extremely confusing, since no real `FileOperationError` is happening, I'd much prefer to use something here that clearly means "skip this move operation". Plugin event handlers _can_ return values, so one alternative would be to have a special `MoveOperation.SKIP`, and use it along the lines of
> ```python
> results = plugins.send("before_item_moved", item=self, source=self.path,
>                  destination=dest)
> if any(r == MoveOperation.SKIP for r in results):
>     # Really not an error, a plugin _chose_ not to move the file.
>     log.warning("Skipping move of item {self}")
>     return
> ```

I had the feeling that returning an error was more confusing, because
right now *data* is returned by hooks that do return data, not
errors. Raising an exception makes it easier to handle it, and to handle
it in a backwards-compatible way.

If `FileOperationError` is not the right exception (and I tend to agree,
i just picked that because it was already there), we could make a new
one. If it's a subclass of HumanReadableException, then it would even be
backwards-compatible because currently such errors are handled
gracefully (with an exit) in beets.

I have this in my checkout, which we could reuse:

```diff
modified   beets/plugins.py
@@ -25,6 +25,7 @@
 
 import beets
 from beets import logging
+from beets.util import HumanReadableException
 import mediafile
 
 
@@ -63,6 +64,15 @@ def filter(self, record):
         return True
 
 
+class PluginSkipException(HumanReadableException):
+    """An exception raised from a plugin to signal the caller to skip
+    whatever operation they are currently doing.
+
+    For example, this could signal a `move` operation to skip this
+    particular item because it is under control of another system that
+    would overwrite its changes, say a bittorrent client.
+    """
+
 # Managing the plugins themselves.
 
 class BeetsPlugin:
```

> Incidentally, the `before_item_moved `event is placed in an unfortunate location, it would be much more useful to have an event _before_ the `MoveOperation` switch and pass the particular `MoveOperation` as an argument to this event. Maybe, we could add such an event (`before_item_moved_v2`?), and deprecate `before_item_moved`?

Yeah, so the way plugins are fired now is really confusing. `write` is
fired before, but `item_moved` is after. Maybe it would make sense to
rewire *all* of them in a v2 event set?

> Relatedly, from #2617
>> we also need to somehow pass information from the plugin to the caller. right now the plugins.send callers mostly ignore the return value, at least in the above hooks (there are rare exceptions elsewhere in the code). i wonder if we should return a truth value or raise an exception to signal a problem.
>>
>> maybe the safest would be to raise an exception from the plugin. older versions of beets would just fail to do the operation and crash (a possible acceptable failure mode, as in "fail closed"), newer versions could handle the exception and behave correctly (e.g. skip or abort, depending on the exception)?
>
> I don't think that justifies designing a confusing new API. Plugins that require this functionality should simply declare a dependency on a sufficiently recent version of beets.

I didn't know you could do that, maybe that would work. But how about
using a new exception instead?

>> Operations like copy or hardlink are not handled by this, for that reason. We may also want to do proper error handling for those as well, that said, but that seems out of scope for this specific issue (#2617).
>
> By having `before_item_moved_v2` as detailed above, plugins could do that without further effort.

but that would still require error handling in beets itself.

wisp3rwind:
> I had the feeling that returning an error was more confusing, because right now *data* is returned by hooks that do return data, not errors. Raising an exception makes it easier to handle it, and to handle it in a backwards-compatible way.

IMO, handling the return value is just as easy (unless you'd want to handle the exception at a higher level, which I'd argue is a bad idea).

> Yeah, so the way plugins are fired now is really confusing. `write` is fired before, but `item_moved` is after. Maybe it would make sense to rewire *all* of them in a v2 event set?

Well, I guess they've all been added individually whenever a plugin author needed another hook, so it's not very surprising that they're not completely consistent. I'd only change (as in, add "v2" versions) events for which you have a use right now, otherwise there's probably a risk of making changes that nobody will use, or where the new call site is even slightly wrong for whoever ends up actually needing it.

> > Plugins that require this functionality should simply declare a dependency on a sufficiently recent version of beets.
> I didn't know you could do that, maybe that would work.

Not via beets' plugin mechanism, but beets plugins are just python packages, which can have version constraints on their dependencies.

anarcat:
On 2022-01-17 14:11:46, Benedikt wrote:
>> I had the feeling that returning an error was more confusing, because right now *data* is returned by hooks that do return data, not errors. Raising an exception makes it easier to handle it, and to handle it in a backwards-compatible way.
>
> IMO, handling the return value is just as easy (unless you'd want to handle the exception at a higher level, which I'd argue is a bad idea).

Well that's the thing: if we do use a derivative of
HumanReadableException, it *will* be handled at a higher level by older
versions of beets.

>> Yeah, so the way plugins are fired now is really confusing. `write` is fired before, but `item_moved` is after. Maybe it would make sense to rewire *all* of them in a v2 event set?
>
> Well, I guess they've all been added individually whenever a plugin author needed another hook, so it's not very surprising that they're not completely consistent. I'd only change (as in, add "v2" versions) events for which you have a use right now, otherwise there's probably a risk of making changes that nobody will use, or where the new call site is even slightly wrong for whoever ends up actually needing it.

Makes sense.

wisp3rwind:
It's unfortunate that this has gone stale; but as explained above at length, I'm not sold on using an `Exception` to communicate plugin decisions here, it seems like the wrong API for the problem. Do others have an opinion or new insight in this matter?

>> IMO, handling the return value is just as easy (unless you'd want to handle the exception at a higher level, **which I'd argue is a bad idea**).
> 
> Well that's the thing: if we do use a derivative of HumanReadableException, it *will* be handled at a higher level by older
versions of beets.

(emphasis added to the quote)


I still don't see why we would _want_ to handle this at a higher level. Any plugin that uses the new functionality will need to declare that it requires `beets>=1.6.X`. With that in place, if we are still concerned that incompatible versions of beets and plugins might be installed (which I'm not sure of that we should be), the solution is probably to verify that version requirement on plugin load (and abort then, instead of a late crash midway during an import).

PR #4244, opened 2022-01-20
Fix exception handling in tests

wisp3rwind:
Apparently we haven't hit these exceptions in a while; see the commit messages for details.

tl;dr: `raise e0, None, e2` was incorrectly translated to `None.with_traceback(e2)`, but clearly, `None` has no such attribute ( `BaseException` has). `e0.with_traceback(e2)` would be correct, but we need here can be expressed in a simpler way on Python 3.


sampsyo (approved):
Very nice. That broken code looks very funny in retrospect. 😄 I really like the "traceback is in the `Exception` object" behavior in Python 3; it might be my favorite under-appreciated Python 2->3 improvement.

PR #4251, opened 2022-01-25
Add query prefixes :~ and :=

rcrowell:
## Description

Adds 2 new query prefixes to specify exact matches for strings; `=` for case-sensitive matches and `~` for case-insensitive.  For example, the queries `~braid` and `=Braid` both match "Braid" but not "Braids".

I have often wished for a way to do this without having to write a regex on the command line.  Perhaps there is a better way than what I am suggesting in this PR, but I wanted to get the ball rolling on this (imo) handy feature.

## To Do

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -191,7 +191,7 @@ class StringQuery(StringFieldQuery):
 
     @classmethod
     def string_match(cls, pattern, value):
-        return pattern == value
+        return pattern.lower() == value.lower()
 
 
 class SubstringQuery(StringFieldQuery):
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -177,6 +177,23 @@ class StringFieldQuery(FieldQuery):
         raise NotImplementedError()
 
 
+class StringQuery(StringFieldQuery):
+    """A query that matches a whole string in a specific item field."""
+
+    def col_clause(self):
+        search = (self.pattern
+                  .replace('\\', '\\\\')
+                  .replace('%', '\\%')
+                  .replace('_', '\\_'))
+        clause = self.field + " like ? escape '\\'"
+        subvals = [search]
+        return clause, subvals
+
+    @classmethod
+    def string_match(cls, pattern, value):
+        return pattern == value
+
+
 class SubstringQuery(StringFieldQuery):
     """A query that matches a substring in a specific item field."""
 
--- a/beets/library.py
+++ b/beets/library.py
@@ -1385,7 +1385,11 @@ def parse_query_parts(parts, model_cls):
     special path query detection.
     """
     # Get query types and their prefix characters.
-    prefixes = {':': dbcore.query.RegexpQuery}
+    prefixes = {
+        ':': dbcore.query.RegexpQuery,
+        '~': dbcore.query.StringQuery,
+        '=': dbcore.query.MatchQuery,
+    }
     prefixes.update(plugins.queries())
 
     # Special-case path-like queries, which are non-field queries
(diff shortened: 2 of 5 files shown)


beets/dbcore/query.py
```diff
@@ -177,6 +177,23 @@ def string_match(cls, pattern, value):
...
 
+class StringQuery(StringFieldQuery):
+    """A query that matches a whole string in a specific item field."""
+
+    def col_clause(self):
+        search = (self.pattern
+                  .replace('\\', '\\\\')
+                  .replace('%', '\\%')
+                  .replace('_', '\\_'))
+        clause = self.field + " like ? escape '\\'"
```
sampsyo, line 188: Could you elaborate a little bit on why this uses SQLite's `LIKE` operator instead of just plain `=`? Maybe I'm missing something, but it seems like that should work without any escaping…

sampsyo (commented):
Truly awesome!! I have thought for a long time that we needed something simple like this, and requests for it come up fairly frequently. (I can't seem to find an issue for such a feature request, but it does come up all the time informally.) The docs are especially clear; I really liked the use of examples.

I have one small technical question within; otherwise, I think this should be good to go.

beets/dbcore/query.py
rcrowell: I am using `LIKE` to perform a case-insensitive string match, though you are right that there are several ways to achieve that which I can think of:

1. `$field LIKE $value`
    - Requires value to be escaped when building the query
    - Match can be satisfied by a `COLLATE NOCASE` index
2. `$field = $value COLLATE NOCASE`
    - Can be applied on a per-clause basis, so `WHERE artist = 'braid' COLLATE NOCASE AND album = 'No Coast'` works as expected
    - Match can be satisfied by a `COLLATE NOCASE` index
3. `UPPER($field) = UPPER($value)`.
    - Simple to understand
    - Not sure how to get sqlite to use an index for this match

Seems like sqlite does not understand how to change the case of non-ascii characters out of the box, so none of these methods will do the right thing for those strings... not sure if this is a deal-breaker?  If so, I am sure we can come up with some hack that will work reasonably well.

sampsyo: Ohhhh, sorry for misunderstanding! I got it backward and thought this was the case-sensitive version; of course, that's just plain old `MatchQuery` in this PR. In that case, you're absolutely right and I don't think there's a strong reason to prefer either of the first two (good point about the index for option 3).

But in that case, is it a bug that `string_match` uses `pattern == value`? It should perhaps use `pattern.lower() == value.lower()` or similar for a similar effect in slow queries.

sampsyo: I think we will just have to live with the consequences for non-ASCII characters. This is already the case for `SubstringQuery` for similar reasons. It's not great, but the complexity of working around it is also not terribly attractive.

rcrowell: Yep, good call on `string_match` being screwed up.  Fixed that (and the test I added which was asserting the incorrect behavior...) in a [second commit](https://github.com/rcrowell/beets/commit/2cab2d670aa011006f4322a59176ba3dbb6bb22b) just now.

Looks like I was the one who got it backwards!

sampsyo:
Looks perfect. Thanks so much for addressing this with style!

kergoth:
@sampsyo Does this now conflict with the `fuzzy` plugin?

sampsyo:
Ack, yes, you're absolutely right, @kergoth—at least with the default prefix. (Fortunately, the plugin uses a configurable prefix.) It might be nice to at least document this, although I don't think it's the end of the world in itself.

kergoth:
> Ack, yes, you're absolutely right, @kergoth—at least with the default prefix. (Fortunately, the plugin uses a configurable prefix.) It might be nice to at least document this, although I don't think it's the end of the world in itself.

Agreed, thanks!

sampsyo:
Hmm, yeah, `=~` is a pretty good idea… it makes it more obviously related to the `=` prefix for exact-match queries. Let's do this!

PR #4258, opened 2022-01-29
convert: Revert some recent changes to convert-on-import path

sampsyo:
This rectifies a couple of things in #4226, as pointed in https://github.com/beetbox/beets/pull/4226#issuecomment-1011499620:

- Undo the `pretend` sensitivity in the import path, because it's not clear how this setting could ever be true.
- Preserve the log message in debug mode, even when quiet.

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -22,6 +22,7 @@ import subprocess
 import tempfile
 import shlex
 from string import Template
+import logging
 
 from beets import ui, util, plugins, config
 from beets.plugins import BeetsPlugin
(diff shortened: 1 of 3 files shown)


wisp3rwind:
Thanks for looking at this again!

Reading the `convert` source again, my initial assessment of the `pretend` "config option" wasn't entirely precise, but backing out this change is still the right thing:

> pretend as used here is only the CLI option to beet convert. That can never be set in the importer in convert_on_import, right?

It can in fact be set, if someone has
```
convert:
    pretend: True
```
in their config, since the plugin uses the `pretend` config key to store the commandline flag when running `beet convert`. When in import mode, that doesn't happen, so the value from the config file (or the default `False`) is still visible. However, setting that to `True` doesn't really make a lot of sense since it would effectively do the same as setting `convert.auto = False`, and for that reason is also undocumented.

tl;dr: `convert.pretend` can be True on import, but that's not how it's meant to be used.

sampsyo:
Thanks for merging! Yes, that's exactly the issue—it's not clear what the semantics of `pretend` ought to be, when the plugin is running "on import." So I don't quite see the use case for empowering `pretend` as an alternative to disabling `auto`.

PR #4259, opened 2022-01-30
zsh completion: handle destination for "convert"

mklein-de:
## Description

Use directory name completion for -d/--dest option.


sampsyo:
Great! I admit I don't fully understand the zsh completion setup, but I'm willing to trust you that this does the right thing. Thank you! :smiley:

PR #4261, opened 2022-01-30
convert plugin: add destination directory pruning

mklein-de:
## Description

This adds `-P`/`--prune` to delete any extra files in the destination directory after converting.

## To Do

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -146,6 +146,7 @@ class ConvertPlugin(BeetsPlugin):
             'copy_album_art': False,
             'album_art_maxwidth': 0,
             'delete_originals': False,
+            'prune': False,
         })
         self.early_import_stages = [self.auto_convert]
 
@@ -174,6 +175,8 @@ class ConvertPlugin(BeetsPlugin):
                               dest='hardlink',
                               help='hardlink files that do not \
                               need transcoding. Overrides --link.')
+        cmd.parser.add_option('-P', '--prune', action='store_true', dest='prune',
+                              help=u'prune destination directory after converting')
         cmd.parser.add_album_option()
         cmd.func = self.convert_func
         return [cmd]
@@ -268,6 +271,10 @@ class ConvertPlugin(BeetsPlugin):
                 if should_transcode(item, fmt):
                     dest = replace_ext(dest, ext)
                 converted = dest
+
+            # Keep track of converted files, so we can prune the destination
+            # later.
+            self.converted_paths.add(util.normpath(converted))
 
             # Ensure that only one thread tries to create directories at a
             # time. (The existence check is not atomic with the directory
@@ -376,6 +383,10 @@ class ConvertPlugin(BeetsPlugin):
         dest = os.path.join(*util.components(dest)[:-1])
 
         dest = album.art_destination(album.artpath, item_dir=dest)
+
+        # Keep track of converted files, so we can prune the destination later.
+        self.converted_paths.add(util.normpath(dest))
+
         if album.artpath == dest:
             return
 
@@ -480,6 +491,7 @@ class ConvertPlugin(BeetsPlugin):
                 self.copy_album_art(album, dest, path_formats, pretend,
                                     link, hardlink)
 
+        self.converted_paths = set()
         convert = [self.convert_item(dest,
                                      opts.keep_new,
                                      path_formats,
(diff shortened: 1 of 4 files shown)


beetsplug/convert.py
```diff
@@ -491,6 +503,26 @@ def convert_func(self, lib, opts, args):
         pipe = util.pipeline.Pipeline([iter(items), convert])
         pipe.run_parallel()
 
+        if opts.prune or self.config['prune'].get():
```
sampsyo, line 506: Configuration options are automatically converted to `bool`:

```suggestion
        if opts.prune or self.config['prune']:
```

beetsplug/convert.py
```diff
@@ -480,6 +491,7 @@ def convert_func(self, lib, opts, args):
                 self.copy_album_art(album, dest, path_formats, pretend,
                                     link, hardlink)
 
+        self.converted_paths = set()
```
sampsyo, line 494: The use of a mutable field on the plugin object creates some potential hazards for bugs in the future. Namely, it can be subtle to see how `self.converted_paths` builds up across the various function invocations. As a result, for example, it seems like it's tricky to get tests exactly right: to make sure that each "job" independently clears the set of converted paths, and that we're capturing exactly the set that we want.

There are also some potential problems with racy accesses when running conversion jobs in parallel.

Is there any chance we could instead have the set be returned from the various functions involved, such as `convert_item`? This would make it easier to see where the data flows and to keep track of the set for a given conversion "transaction."

docs/plugins/convert.rst
```diff
@@ -54,6 +54,11 @@ instead, passing ``-H`` (``--hardlink``) creates hard links.
 Note that album art embedding is disabled for files that are linked.
 Refer to the ``link`` and ``hardlink`` options below.
 
+by default, any extra files in the destination directory are left untouched. To
```
sampsyo, line 57:
```suggestion
By default, any extra files in the destination directory are left untouched. To
```

docs/plugins/convert.rst
```diff
@@ -54,6 +54,11 @@ instead, passing ``-H`` (``--hardlink``) creates hard links.
...
+remove any files that have no matching library file after converting, use the
```
sampsyo, line 58: I'm a little confused by this wording: what exactly is a "matching library file"? Maybe some examples would help.

sampsyo (commented):
Thanks for getting this started! It seems potentially useful.

I have two main comments: one technical one about how to build up the set of converted paths, and one more general one: I don't _quite_ understand the use case, or how we should explain this to users. It seems a like a few more sentences of context in the docs could help… maybe you could start by explaining what the problem is in your current setup that this addresses for you? Is it _pre-existing_ clutter in the place where you're putting converted files, for example? Or is it detritus from multiple conversion runs? Or other non-music files?

wisp3rwind:
> I have two main comments: one technical one about how to build up the set of converted paths, and one more general one: I don't _quite_ understand the use case, or how we should explain this to users. It seems a like a few more sentences of context in the docs could help… maybe you could start by explaining what the problem is in your current setup that this addresses for you? Is it _pre-existing_ clutter in the place where you're putting converted files, for example? Or is it detritus from multiple conversion runs? Or other non-music files?

I agree, without knowing the intended use case, I'd object to merging this: This functionality seems particularly dangerous with respect to accidental data loss.

mklein-de:
Thank you for your insights!

> Thanks for getting this started! It seems potentially useful.
> 
> It seems a like a few more sentences of context in the docs could help… maybe you could start by explaining what the problem is in your current setup that this addresses for you?

Sure; I use the convert together with the playlist plugin to sync music files to different portable media, typically via one playlist per media. This works fine as long as tracks are only added to a playlist, but fails in the case where tracks are removed from a playlist or library files we moved or renamed. So it's always clutter from earlier conversion runs that I want to get rid of.

wisp3rwind:
That sounds like it might better be solved by https://github.com/geigerzaehler/beets-alternatives, which explicitly tracks the files it manages (which works for tag changes/moves, as long as the file is not removed entirely from the library).

sampsyo:
Got it. Aside from beets-alternatives (which might indeed be a good fit), am I correct in understanding that the final result should resemble doing `rm -rf $dest ; mkdir $dest ; beet convert -d $dest ...`? In other words, the idea is to get rid of everything in the destination that is not a converted file.

(Of course, I imagine that this would in practice be faster because the `convert` command gets to skip over the already-converted destination files.)

mklein-de:
Yes, beets-alternatives looks indeed like a perfect fit, and since there's not much point in duplicating existing functionality, I think this PR can be closed.

sampsyo:
Awesome! Yeah, it's a pretty great plugin. Thanks for looking into this anyway!!

PR #4262, opened 2022-01-31
Improve lyrics tests

wisp3rwind:
## Description

Inspired by the recurring test failures of the kind seen in https://github.com/beetbox/beets/runs/4993930755?check_suite_focus=true

Generate more useful output on failed assertion in the lyrics integration tests, in particular provides more details for the failures in https://github.com/beetbox/beets/runs/4993930755?check_suite_focus=true

In addition, `subTest` is used to better separate tests of different backends, and avoids aborting the entire test function if one backend fails.

## To Do

Nothing, this is only improves the tests.

### DIFF
--- a/test/test_lyrics.py
+++ b/test/test_lyrics.py
@@ -231,13 +231,24 @@ class MockFetchUrl:
         return content
 
 
-def is_lyrics_content_ok(title, text):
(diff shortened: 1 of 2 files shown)


sampsyo (approved):
Nice; this looks great. Getting better information back will help a lot! ✨🚀

PR #4263, opened 2022-02-01
confit: Improve deprecation warning

wisp3rwind:
Show the actual origin of the import statement, cf. #4024.

### DIFF
--- a/beets/util/confit.py
+++ b/beets/util/confit.py
@@ -16,7 +16,13 @@
 import confuse
 
 import warnings
-warnings.warn("beets.util.confit is deprecated; use confuse instead")
+warnings.warn(
+    "beets.util.confit is deprecated; use confuse instead",
+    # Show the location of the `import confit` statement as the warning's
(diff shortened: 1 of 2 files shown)


sampsyo:
Wow; cool! I didn't know about the `stacklevel` argument to `warn` here, which is pretty nifty. The Python docs are oddly vague about it (merely "the *stacklevel* argument can be used by wrapper functions written in Python"), but it seems like exactly the right thing.

wisp3rwind:
> Wow; cool! I didn't know about the `stacklevel` argument to `warn` here, which is pretty nifty. The Python docs are oddly vague about it (merely "the _stacklevel_ argument can be used by wrapper functions written in Python"), but it seems like exactly the right thing.

Yep, I didn't know about it before, either. I'm quite happy that the "fix" turned out so concise; it would have been more work to manually modify warning printing from the plugin loader. I wrote a test for this precisely because I wasn't sure to understand the documentation correctly.

PR #4264, opened 2022-02-01
test_replaygain: fix exception handling again

wisp3rwind:
This is the same as d26b0bc1 (from #4244), which I reintroduced again in a new test
due to a botched rebase...


sampsyo:
Oops! Thanks!!

PR #4280, opened 2022-02-15
Fix error while opening config on Windows (with beet config -e command)

### LINKED ISSUE #2847: Opening config produces Errno 22

JohnKara:
## Description

Fixes #2847 

## To Do

### DIFF
--- a/beetsplug/fish.py
+++ b/beetsplug/fish.py
@@ -85,8 +85,8 @@ class FishPlugin(BeetsPlugin):
             '-o',
             '--output',
             default='~/.config/fish/completions/beet.fish',
-            help='save the script to a specific file, by default it will be'
-                 'saved to ~/.config/fish/completions')
+            help='where to save the script. default: '
+                 '~/.config/fish/completions')
         return [cmd]
 
     def run(self, lib, opts, args):
--- a/docs/plugins/fish.rst
+++ b/docs/plugins/fish.rst
@@ -51,5 +51,7 @@ for example, very large numbers of genres/artists may result in higher memory
 utilization, completion latency, et cetera. This option is not meant to replace
 database queries altogether.
 
-By default the completion file will be generated at ``~/.config/fish/completions/``.
-If you want to save it somewhere else you can use the ``-o`` or ``--output``.
+By default, the completion file will be generated at
+``~/.config/fish/completions/``.
+If you want to save it somewhere else, you can use the ``-o`` or ``--output``
+option.


beets/util/__init__.py
```diff
@@ -949,17 +949,21 @@ def interactive_open(targets, command):
...
+        except ValueError:  # Malformed shell tokens.
+            args = [command]
 
-    args.insert(0, args[0])  # for argv[0]
+    # Unnecessary for subprocess.call()
+    # args.insert(0, args[0])  # for argv[0]
 
     args += targets
-
-    return os.execlp(*args)
```
sampsyo, line 962: Well, there is a reason we're using `execlp`… it "replaces" the beets process with the new editor process, which is a bit nicer on people's systems. There isn't a useless Python process just taking up memory, waiting for the editor to close.

beets/util/__init__.py
```diff
@@ -949,17 +949,21 @@ def interactive_open(targets, command):
     """
     assert command
 
-    # Split the command string into its arguments.
-    try:
-        args = shlex.split(command)
-    except ValueError:  # Malformed shell tokens.
+    if os.name == 'nt':
         args = [command]
```
sampsyo, line 953: Does this work for cases when the command has arguments? Like `vim -b`, for example?

sampsyo (commented):
Thanks for getting this started! Do you think it's impossible to support using `execlp` on Windows? If so, then we should indeed switch to `subprocess`, but probably on Windows only. But I think `execlp` *is* supposedly functional on Windows? Can we perhaps get a little more specific about what's going wrong here—perhaps it's just the erroneous escaping—and work around it in a more direct way?

JohnKara:
> Thanks for getting this started! Do you think it's impossible to support using `execlp` on Windows? If so, then we should indeed switch to `subprocess`, but probably on Windows only. But I think `execlp` _is_ supposedly functional on Windows? Can we perhaps get a little more specific about what's going wrong here—perhaps it's just the erroneous escaping—and work around it in a more direct way?

When I try to edit the config file from command prompt using 'beet config -e', it gives me error 2 file not found. In this situation the os.execlp cannot find the 'start' command for some reason. When I add EDITOR variable to 'C:\Windows\System32\notepad.exe' it is converted from shlex.split(command) to 'C:WindowsSystem32notepad.exe' and again gives me error 2 file not found. If I remove this command it works with the EDITOR variable. 

So I thought maybe it's better not to use the shlex.split with Windows and I also read that subprocess.call is better than os.execlp. 

If there is a problem with subprocess.call on Linux, I can leave it as it is for Linux and make these changes only for Windows systems.

JohnKara:
I did something wrong with the commits but the one we want is this: https://github.com/beetbox/beets/pull/4280/commits/c816525d652bdce430bc6eeb764785b7fed85eb6

JohnKara:
I think this is what we want
```python
assert command

    if os.name == 'nt':
        if command == 'start':
            args = ['cmd.exe']
            targets = ['/C start ' + targets[0]]
        else:
            args = [command]
    else:
        # Split the command string into its arguments.
        try:
            # Doesn't work well with Windows
            # (removes all the '\' from path resulting in an invalid path)
            args = shlex.split(command)
        except ValueError:  # Malformed shell tokens.
            args = [command]

    args.insert(0, args[0])  # for argv[0]

    args += targets
    return os.execlp(*args)
```
I commited and pushed the changes to my forked repository.

sampsyo:
Hmm; it looks like this PR has other changes in it… maybe you could try force-pushing a new branch with the just changes you've settled on, or closing this and opening a new PR if all else fails? (That would make it much easier to review. Thanks!!)

PR #4281, opened 2022-02-15
fish plugin: Add --output option

jaimeMF:
I find it useful because it would allow us to generate the completion script when creating an .rpm package.

## To Do

### DIFF
--- a/beetsplug/fish.py
+++ b/beetsplug/fish.py
@@ -85,7 +85,8 @@ class FishPlugin(BeetsPlugin):
             '-o',
             '--output',
             default=None,
-            help='save the script to a specific file, by default it will be saved to ~/.config/fish/completions')
+            help='save the script to a specific file, by default it will be'
+                 'saved to ~/.config/fish/completions')
         return [cmd]
 
     def run(self, lib, opts, args):
--- a/beetsplug/fish.py
+++ b/beetsplug/fish.py
@@ -105,11 +105,7 @@ class FishPlugin(BeetsPlugin):
             completion_file_path = os.path.join(completion_dir, 'beet.fish')
 
         if completion_dir != '':
-            try:
-                os.makedirs(completion_dir)
-            except OSError:
-                if not os.path.isdir(completion_dir):
-                    raise
+            os.makedirs(completion_dir, exist_ok=True)
 
         nobasicfields = opts.noFields  # Do not complete for album/track fields
         extravalues = opts.extravalues  # e.g., Also complete artists names
--- a/beetsplug/fish.py
+++ b/beetsplug/fish.py
@@ -84,7 +84,7 @@ class FishPlugin(BeetsPlugin):
         cmd.parser.add_option(
             '-o',
             '--output',
-            default=None,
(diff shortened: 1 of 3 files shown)


beetsplug/fish.py
```diff
@@ -89,14 +95,22 @@ def run(self, lib, opts, args):
...
+            home_dir = os.path.expanduser("~")
+            completion_dir = os.path.join(home_dir, '.config/fish/completions')
+            completion_file_path = os.path.join(completion_dir, 'beet.fish')
+
+        if completion_dir != '':
+            try:
+                os.makedirs(completion_dir)
+            except OSError:
+                if not os.path.isdir(completion_dir):
+                    raise
```
sampsyo, line 112: I know this is not new, but I think we should probably just do `makedirs(completion_dir, exist_ok=True)`, which avoids the need for the exception handler.

beetsplug/fish.py
```diff
@@ -89,14 +95,22 @@ def run(self, lib, opts, args):
...
-            if not os.path.isdir(completion_dir):
-                raise
-        completion_file_path = os.path.join(completion_dir, 'beet.fish')
+
+        completion_file_path = opts.output
+        if completion_file_path is not None:
+            completion_dir = os.path.dirname(completion_file_path)
+        else:
+            home_dir = os.path.expanduser("~")
+            completion_dir = os.path.join(home_dir, '.config/fish/completions')
```
sampsyo, line 104: Maybe it would be easier to just make `~/.config/fish/completions/beet.fish` the actual `default` parameter to `add_option` above, and then stick to `completion_dir = os.path.dirname(...)` regardless of whether the parameter is set? As an added bonus, this would let the parameter include `~`, because we'd need to include `expanduser` on the entire path.

docs/plugins/fish.rst
```diff
@@ -50,3 +50,6 @@ with care when specified fields contain a large number of values. Libraries with
 for example, very large numbers of genres/artists may result in higher memory
 utilization, completion latency, et cetera. This option is not meant to replace
 database queries altogether.
+
+By default the completion file will be generated at ``~/.config/fish/completions/``,
+if you want to save it somewhere else you can use the ``-o`` or ``--output``.
```
sampsyo, line 55:
```suggestion
By default the completion file will be generated at ``~/.config/fish/completions/``.
If you want to save it somewhere else you can use the ``-o`` or ``--output``.
```
(Avoid a comma splice.)

sampsyo (changes requested):
Awesome; thanks for taking care of this!! Looking pretty good so far. I have a couple of suggestions for keeping things simple, most of which pre-exist this PR but would be great to address with this additional change.

beetsplug/fish.py
jaimeMF: Done.

jaimeMF:
Thank you for the review, i have fixed the issues. If you prefer that I squash all or some commit, just tell me.

sampsyo (approved):
Looks great; thank you!!

PR #4286, opened 2022-02-18
Fix error while opening config on Windows (with beet config -e command)

### LINKED ISSUE #2847: Opening config produces Errno 22

JohnKara:
## Description

Fixes #2847

## To Do

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -949,11 +949,20 @@ def interactive_open(targets, command):
     """
     assert command
 
-    # Split the command string into its arguments.
-    try:
-        args = shlex.split(command)
-    except ValueError:  # Malformed shell tokens.
-        args = [command]
+    if os.name == 'nt':
+        if command == 'start':
+            args = ['cmd.exe']
+            targets = ['/C start ' + targets[0]]
+        else:
+            args = [command]
+    else:
+        # Split the command string into its arguments.
+        try:
+            # Doesn't work well with Windows
+            # (removes all the '\' from path resulting in an invalid path)
+            args = shlex.split(command)
+        except ValueError:  # Malformed shell tokens.
+            args = [command]
 
     args.insert(0, args[0])  # for argv[0]
 
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -921,7 +921,7 @@ def open_anything():
     if sys_name == 'Darwin':
         base_cmd = 'open'
     elif sys_name == 'Windows':
-        base_cmd = 'start'
+        base_cmd = 'cmd.exe'
     else:  # Assume Unix
         base_cmd = 'xdg-open'
     return base_cmd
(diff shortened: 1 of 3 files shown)


sampsyo:
Awesome; thanks for continuing to push this forward!

Instead of detecting when the command is `start` and changing it here, what do you think about addressing the root cause—that is, changing the command where we select it in the first place?
https://github.com/beetbox/beets/blob/d0a5f812fded18287ef7e8db55112ff015cc95fe/beets/util/__init__.py#L924

JohnKara:
You are right, sorry I hadn't noticed it before.

beets/util/__init__.py
```diff
@@ -949,12 +949,21 @@ def interactive_open(targets, command):
...
 
-    # Split the command string into its arguments.
-    try:
-        args = shlex.split(command)
-    except ValueError:  # Malformed shell tokens.
+    if os.name == 'nt':
         args = [command]
 
+        if command == 'cmd.exe':
+            targets = ['/C start ' + targets[0]]
```
wisp3rwind, line 956: You're silently dropping any `targets` beyond the first here; that seems like a bad idea. A brief `grep` suggests that we're not using `targets` lists with more than one items, but the signature of `itneractive_open` currently suggests that this is a supported thing. Since, according to https://github.com/beetbox/beets/issues/2847#issuecomment-375395370, opening several files at once is probably not going to work on Windows, we might want to change that. However, I would prefer to actually change the function signature to take a single filename instead of a list.

beets/util/__init__.py
```diff
@@ -949,12 +949,21 @@ def interactive_open(targets, command):
...
+    else:
+        # Split the command string into its arguments.
+        try:
+            # Doesn't work well with Windows
+            # (removes all the '\' from path resulting in an invalid path)
+            args = shlex.split(command)
+        except ValueError:  # Malformed shell tokens.
+            args = [command]
+
     args.insert(0, args[0])  # for argv[0]
```
wisp3rwind, line 967: According to https://github.com/beetbox/beets/issues/2847#issuecomment-375387406, that should only be done on *nix, is this still the case?

wisp3rwind:
> Awesome; thanks for continuing to push this forward!
> 
> Instead of detecting when the command is `start` and changing it here, what do you think about addressing the root cause—that is, changing the command where we select it in the first place?
> 
> https://github.com/beetbox/beets/blob/d0a5f812fded18287ef7e8db55112ff015cc95fe/beets/util/__init__.py#L924

Couldn't this be simplified by returning a list (i.e. the result of splitting the command) from `open_anything` and  `editor_command` instead of strings? According to https://github.com/beetbox/beets/pull/2851#discussion_r178920618, it should be possible to treat Windows/*nix (almost) identically  in `interactive_open` then. That is, `open_anything` could simply return `["cmd.exe", "start", "/C"]` on Windows, and `interactive_open` would simply append `targets[0]` to that list.

sampsyo:
Yes, precisely—returning a list containing the necessary command incarnation on Windows would be perfect!

PR #4299, opened 2022-02-26
fix: dont deduplicate matches if musicbrainz ID is not set for album matches

Neurrone:
First discussed in https://github.com/beetbox/beets/discussions/4269#discussioncomment-2251701

## Description

Fixes album candidates being treated as duplicates when the album id (assumed to be musicbrainz id) is not set.

(...)

## To Do

### DIFF
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -357,8 +357,8 @@ def _add_candidate(items, results, info):
         log.debug('No tracks.')
         return
 
-    # Don't duplicate.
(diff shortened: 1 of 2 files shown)


sampsyo:
Looks great to me overall! Would you mind adding a quick changelog entry, and then we can merge this?

sampsyo:
Looks great; thanks!! Slightly tweaked the wording, but this is good to go. 🙏

Issue #4087, opened 2021-10-02
Prevent Discogs from overwriting genre tags


arsaboo:
Discogs is a useful source for meta-data, however, the genre information in Discogs is pretty useless. Currently, when we use Discogs as a tag source, everything is overwritten by Discogs information. I posted it on [Discourse here](https://discourse.beets.io/t/prevent-discogs-plugin-from-overwriting-existing-genre-tags/1884) and Adrian suggested that currently, the Discogs plugin is not configurable. 

In general, I think it would be a good idea to make plugins configurable so that they don't overwrite other details that the user may not want. 

### Proposed solution

Make Discogs plugin configurable so that users can choose fields to be ignored. The current behavior (where it imports everything) can be the default so that this is not a breaking change. 

### Objective

Avoid getting unwanted information from Discogs. 

#### Goals

Provide more fine-grained user control.

wisp3rwind:
Maybe the solution here should be a more general mechanism to configure auto-tagging rather than a discogs-specific approach? Something like a `keep-fields` importer option? The `*sync` plugins might want to respect such a setting, too.

wisp3rwind:
In fact, we could have something like the following to also address (a part of) #4071

```yaml
import:
    autotag: yes|no
    autotag-fields:
        genre: 'append'  # see #4071
        year: 'as-is'   # import as-is, don't apply autotagger data
        comment: 'set: foo'   # set field to a constant value
        title: 'auto'   # the default, autotag this field
        albumartist_sort: 'zero'   # clear the field
```

and deprecate the `set_fields` option.

sampsyo:
That's a pretty good idea! We could start with just the "filter" functionality (selecting fields not to apply) and then later expand it to take on the role of stuff like `set_fields`.

arsaboo:
I like the idea of configuring inputs (with sensible defaults) that other auto-taggers could follow.

snejus:
I've got something similar implemented in the `beetcamp` plugin, where most of the fields can be excluded. See the [readme](https://github.com/snejus/beetcamp) or the [implementation](https://github.com/snejus/beetcamp/blob/master/beetsplug/bandcamp/_metaguru.py#L587)
```yaml
bandcamp:
  ...
  exclude_extra_fields: []
  ...
```

snejus:
Running import without exclusions:
![image](https://user-images.githubusercontent.com/16212750/143325854-5d18a832-6dfa-426e-9760-82c248048068.png)

Running import with
[...]

PR #4300, opened 2022-02-26
Avoid writing a user-configurable set of file tags

JOJ0:
## Description

- Adds a configuration option(import.autotag_exclude_fields) that let's a user decide which file tags should never be written by beets.
- Fixes parts of #4087

## To Do

- [ ] Documentation
- [ ] Changelog
- [ ] Tests

### DIFF
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -30,6 +30,7 @@ import:
     duplicate_action: ask
     bell: no
     set_fields: {}
+    autotag_exclude_fields: []
 
 clutter: ["Thumbs.DB", ".DS_Store"]
 ignore: [".*", "*~", "System Volume Information", "lost+found"]
--- a/beets/library.py
+++ b/beets/library.py
@@ -753,9 +753,13 @@ class Item(LibModel):
             id3v23 = beets.config['id3v23'].get(bool)
 
         # Get the data to write to the file.
+        exclude = beets.config['import']['autotag_exclude_fields'].as_str_seq()
         item_tags = dict(self)
         item_tags = {k: v for k, v in item_tags.items()
-                     if k in self._media_fields}  # Only write media fields.
+                     if k in self._media_fields  # Only write media fields
+                     and not k in exclude}  # and respect excluded fields.
+
+        log.debug('Excluding tags from write: {0}', ' '.join(exclude))
         if tags is not None:
             item_tags.update(tags)
         plugins.send('write', item=self, path=path, tags=item_tags)


JOJ0:
This an idea to address parts of the linked issue as suggested in this comment: https://github.com/beetbox/beets/issues/4087#issuecomment-932956587

It forbids the actual write to file at the lowest possible level, the library Item.write method and states it in the log. I wanted to get in touch and discuss if this "once and for all" approach is the way to go here and could be a basis for further features discussed in the issue, or if the exclusions should happen on "upper layers" in the code.

There is a few issues with this approach currently:
- If there are autotag_exclude_fields configured that _are_ stated: in the "what will change view" of the imported then it's confusing to the user because he sees that something will change that won't be changed in the end.

For example I configured tags `title` and `genre`to not be written, certainly still the importer tells me that title would be changed:

```
Enter search, enter Id, aBort, eDit, edit Candidates, plaY? i
Enter release ID: 1315756
Tagging:
    Sade - Your Love Is King
URL:
    https://www.discogs.com/release/1315756-Sade-Your-Love-Is-King
(Similarity: 59.8%) (id, tracks, catalognum) (Discogs, Vinyl, US, Epic, 12-4137)
 * Your Love Is King FLAC - A - 135 (3:36)   -> Your Love Is King (3:58) (id, length)
 * Smooth Operator/Snake Bite FLAC - Am - 60 -> Smooth Operator / Snake Bite (id)
 * Love Affair With Life FLAC - Bm - 88      -> Love Affair With Life (id)
[A]pply, More candidates, Skip, Use as-is, as Tracks, Group albums,
Enter search, enter Id, aBort, eDit, edit Candidates, plaY?
```

- Certainly the naming of the config option is just a first idea and could be changed to something as proposed here: https://github.com/beetbox/beets/issues/4087#issuecomment-932901710 to be prepared for future improvements.

sampsyo:
Interesting; thanks for getting this started! As you mentioned in the commend above, I can see two ways to address #4087 and others:

1. Avoid *writing the field values from the database to disk*, as this PR does. This would not just affect the importer but *all* occasions that beets has to write values into files' metadata (such as `beet write`). (If this were the route, I would suggest that we would *not* want this to include "autotag" in the option name, nor would we want it to be nested under the `import:` section.)
2. Avoid *applying metadata from a match source to the database*. This would happen in the `apply_metadata` and `apply_item_metadata` functions in the `beets.autotag` module.

I think the spirit of the request in #4087 leans toward option 2 above. This also seems more generally useful to me: it allows for field exclusion even when someone doesn't write tags to files at all. And when tags _are_ being written to files, it also avoids the somewhat confusing situation where the database and the files become "out of sync." So I think I lean that direction, although folks interested in the aforementioned behavior should also weigh in.

JOJ0:
Hi @sampsyo and @wisp3rwind, thanks a lot for your quick replies and valuable opinions. Yes, I probably didn't think this through to the end but wanted to get the ball rolling and thought I give it a shot, it might be a possible direction.

> I also think that option 2 should be the way to go: beets design is fundamentally based on treating the database as the ground truth.

After reading your answer @wisp3rwind and even when I wrote this code I recalled having read somewhere in some doc, an issue, a statement of one of you guys, ....?,  that the basic idea of beets is to "keep the database and file tags" in sync. This got me thinking but I tried out the approach anyway.

`beet info` is well aware of the fact that there might be a difference between database and file tag, so I thought that actually it _is_ a consistent state for beets:
```
(beets) jojo@gin ~/git/beets (skip_write) $ beet info sade smooth | grep title
           disctitle:
               title: Smooth Operator/Snake Bite FLAC - Am - 60
(beets) jojo@gin ~/git/beets (skip_write) $ beet info -l sade smooth | grep title
           disctitle:
               title: Smooth Operator / Snake Bite
(beets) jojo@gin ~/git/beets (skip_write) $
```

`beet edit` and `modify` on the other hand are not aware of an "unsynced" situation, `beet edit` with this PR's code now just lets one edit the database value.

JOJ0:
One last question on the direction go though: To me personally having the option of not overwriting a field, in my case the title field is (weirdly enough) important. So as a first step I suggest from these options: https://github.com/beetbox/beets/issues/4087#issuecomment-932901710

to implement the 'as-is' and the 'auto' values. 

What do you think, would this be enough for a first PR and staying consisten with beets behaviour?

PR #4302, opened 2022-02-28
Add auto_keep option to convert plugin - attempt to fix #1840

### LINKED ISSUE #1840: convert: In auto mode, option to keep originals instead of converted files
Hey All,

I'm looking to use beets to manage my lossless library, but I have devices in the house that do not support high bit-rate formats (Sonos devices). So I'm looking to use Beets to import music into it's database and at the same time create a copy of the Lossless files, convert them into lossy/.mp3 and store in a different location.

Couple questions:
1. Can this all be done with one Beets instance / database or should I use one for the lossy and one for the lossless?
2. How would this be configured? I'm using the below configuration and it only keeps the MP3 version and ignores the FLACs.

```
plugins: fetchart embedart convert scrub replaygain lastgenre chroma web
directory: /music
art_filename: albumart
threaded: yes
original_date: no
per_disc_numbering: no

convert:
    auto: yes
    ffmpeg: /usr/bin/ffmpeg
    max_bitrate: 320
    threads: 5
    copy_album_art: yes
    dest: /music/lossy

paths:
    default: $albumartist/$album%aunique{}/$track - $title
    albumtype_soundtrack: Soundtracks/$album/$track $title

import:
    write: yes
    copy: yes
    move: no
    resume: ask
    incremental: yes
    quiet_fallback: skip
    timid: no
    log: /config/beet.log

lastgenre:
    auto: yes
    source: album

embedart:
    auto: yes

fetchart:
    auto: yes

replaygain:
    auto: no

scrub:
    auto: yes

replace:
    '^\.': _
    '[\x00-\x1f]': _
    '[<>:"\?\*\|]': _
    '[\xE8-\xEB]': e
    '[\xEC-\xEF]': i
    '[\xE2-\xE6]': a
    '[\xF2-\xF6]': o
    '[\xF8]': o
    '\.$': _
    '\s+$': ''

web:
    host: 0.0.0.0
    port: 8337
```

DucNg:
## Description

Fixes https://github.com/beetbox/beets/issues/1840

My idea was to add a new option so people using the old option won't have to change anything. This way it's a non breaking change. Even if being the default for the auto option would make more sens.

I don't think option naming is ideal so I'm very open to suggestions on this.

Using beet-alternatives to do exactly that is also a possibility but seemed a little overkill for my use case. Figured could be nice to have it build in.

## To Do

Should I had tests for this? Changes seems minimal.


sampsyo:
Thanks for getting this started; this seems like a good approach! (That is, an entirely separate option that's like `auto` but does the other thing with the generated files.)

I actually don't mind the name. `auto_keep` seems perfectly serviceable to me.

Here is an ill-posed question, if you don't mind: I'm curious about why there is so much code in the new `auto_convert_keep` method. It needs to cover all the configuration bases, of course. But I wonder if this would be a good opportunity to refactor some of that stuff out so that it can share setup/configuration logic with the other two modes (`auto` and explicitly with the `beet convert` command). But how to do this is not immediately obvious and would require some creativity…

DucNg:
You're right, all the new code is basically copy paste from the convert function. Now might be a good time to refactor. I'll see what I can do.

-------- Original Message --------

DucSkello:
@sampsyo I tried to refactor common code as best as I could.

beetsplug/convert.py
```diff
@@ -187,31 +187,13 @@ def auto_convert(self, config, task):
...
-                raise ui.UserError('no convert destination set')
-            dest = util.bytestring_path(dest)
-
-            path_formats = ui.get_path_formats(self.config['paths'] or None)
-
-            hardlink = self.config['hardlink'].get(bool)
-            link = self.config['link'].get(bool)
-
-            threads = self.config['threads'].get(int)
+            empty_opts = self.commands()[0].parser.get_default_values()
```
DucSkello, line 190: Is there a better way to do this? It doesn't feel quite right.

DucNg, line 190: Is there a better way to do this? It doesn't feel quite right.

DucNg:
@sampsyo I tried to refactor common code as best as I could.

docs/plugins/convert.rst
```diff
@@ -66,6 +66,10 @@ file. The available options are:
   default configuration) non-MP3 files over the maximum bitrate before adding
   them to your library.
   Default: ``no``.
+- **auto_keep**: As opposite to **auto**, import non transcoded versions of
+  your files but still convert them to **dest**. It uses the default format
+  you have defined in your config file.
+  Default: ``no``.
```
sampsyo, line 72: Maybe this should be slightly more explicit that this does this _automatically on import_? In that sense, it is similar to (rather than solely different from) `auto`.

beetsplug/convert.py
```diff
@@ -546,3 +526,45 @@ def _cleanup(self, task, session):
...
+            hardlink = opts.hardlink
+            link = False
+        elif opts.link is not None:
+            hardlink = False
+            link = opts.link
+        else:
+            hardlink = self.config['hardlink'].get(bool)
+            link = self.config['link'].get(bool)
+
+        return (dest, threads, path_formats, fmt, pretend, hardlink, link)
```
sampsyo, line 557: No parentheses are necessary here.

beetsplug/convert.py
```diff
@@ -546,3 +526,45 @@ def _cleanup(self, task, session):
                 if os.path.isfile(path):
                     util.remove(path)
                 _temp_files.remove(path)
+
+    def _get_opts_and_config(self, opts):
```
sampsyo, line 530: A quick docstring on this new method would be super helpful. (No need to even list all the options exhaustively, IMO—just giving the gist of what this does would be useful.)

beetsplug/convert.py
```diff
@@ -546,3 +526,45 @@ def _cleanup(self, task, session):
...
+            hardlink = False
+            link = opts.link
+        else:
+            hardlink = self.config['hardlink'].get(bool)
+            link = self.config['link'].get(bool)
+
+        return (dest, threads, path_formats, fmt, pretend, hardlink, link)
+
+    def _parallel_convert(self, dest, keep_new, path_formats, fmt,
+                          pretend, link, hardlink, threads, items):
```
sampsyo, line 560: A docstring here could be really useful too.

beetsplug/convert.py
sampsyo: Very creative! While it doesn't bother me too much, it does seem a little odd that we'd be using the CLI arguments here even though this method doesn't have anything to do with the command itself.

It's not exactly very clean, but one option would be to let `_get_opts_and_config`'s argument be `None`. I think all of the conditionals in there would end up ignoring the `opts` value for this mode anyway, right? At least it would make it clear that we are using only the configuration values, nothing from the command line.

sampsyo (commented):
Awesome; looking good!! Here are a few granular suggestions.

beetsplug/convert.py
DucNg: Actually that was my first attempt but it doesn't work.

For example for dest it will do:
`dest = None.dest ....`
Resulting in: `'NoneType' object has no attribute 'dest'`

A elegant workaround would have been to to this: `dest = opts?.dest` for every options. But unfortunately safe navigation operator is not available in python.

The only option left then would be to manually check if opts is None for every option witch is not very elegant in my opinion. What do you think? I think `parser.get_default_values()` works well in this context but looks very hacky at the same time.

sampsyo: Ah, yes, the `opts.foo if opts else config[‘foo’].get()` dance was pretty much what I was recommending here. But you make a good point that this would bring a high complexity tax for something that doesn’t seem worth it.

I think a better solution would probably be to [use Confuse’s support for argparse options](https://confuse.readthedocs.io/en/latest/usage.html#command-line-options). But maybe the best route would be to note that this is the next step and merge this PR as-is, hoping to tackle the Confuse-based alternative in a subsequent PR (since it will involve some refactoring of its own)?

DucNg: Fine with me

mannp:
Thanks for the additional option... I wondered if the new option respects if auto convert is disabled?

I have the following config, which I would not expect to convert new imported items, but it is?
```
convert:
    auto: no
    auto_keep: True,
```

sampsyo:
It's a good point; it might be nice to document the interaction between the two options. The intent is that they're mutually exclusive; you enable one or the other. Care to expand on that in the docs at all, @DucNg?

mannp:
I'd understood 'auto' was to convert or not to convert on import... while I convert by running the convert command... if it's working the same as auto, it would still keep the originals in place, but only do it when the convert command is issued.

Will auto_keep: 'no' also work, to keep it the same as auto? 

Just observations, but thanks again for making the changes :)

sampsyo:
Yes! The default for `auto_keep` is `no`, so you don't to change anything if you don't want to use it. But keeping it off will let you manually convert stuff by typing `beet convert`.

DucNg:
> It's a good point; it might be nice to document the interaction between the two options. The intent is that they're mutually exclusive; you enable one or the other. Care to expand on that in the docs at all, @DucNg?

Yes that might be a good idea. Even if enabling both is theoretically possible.

PR #4304, opened 2022-03-02
feat: discogs: allow style to be appended to genre

jdheyburn:
## Description

- Adds a configuration that, when enabled, will append the style to genre
- Rationale is to have more verbose genres in genre tag of players that only support genre
- [Discussed here](https://github.com/beetbox/beets/discussions/4301)

## To Do

### DIFF
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -57,6 +57,7 @@ class DiscogsPlugin(BeetsPlugin):
             'user_token': '',
             'separator': ', ',
             'index_tracks': False,
+            'append_style_genre': False,
         })
         self.config['apikey'].redact = True
         self.config['apisecret'].redact = True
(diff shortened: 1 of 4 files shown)


docs/plugins/discogs.rst
```diff
@@ -76,6 +76,12 @@ whereas with ``index_tracks`` disabled you'd get::
 
 This option is useful when importing classical music.
 
+Other configurations available under ``discogs:`` are:
+
+- **append_style_genre**: Appends the style (if found) to the genre tag, useful if you would like more granular genre styles added to music file tags
```
sampsyo, line 81: Maybe including an example here would be useful? Your "Electronic; Techno" example is pretty illustrative, for instance. You could say that, by default, you just get "Electronic" but get the longer value when the option is enabled.

beetsplug/discogs.py
```diff
@@ -318,7 +319,13 @@ def get_album_info(self, result):
         country = result.data.get('country')
         data_url = result.data.get('uri')
         style = self.format(result.data.get('styles'))
-        genre = self.format(result.data.get('genres'))
```
sampsyo, line 321: Maybe this could be simplified a bit by keeping the genre itself in a variable, like:

```py
base_genre = self.format(result.data.get('genres'))
```

…and using that in the assignments below.

sampsyo (commented):
Nice; looks great so far!! I just have a couple of extremely minor suggestions.

jdheyburn:
@sampsyo Thanks for the review, both very valid points (what happens when I late-night code :joy:). I've pushed the latest changes now.

sampsyo (approved):
Looks great!! I tweaked the docs a tiny bit, but this is good to go. 🚀 ✨

jdheyburn:
Amazing, thanks for the support!

PR #4312, opened 2022-03-12
add interative usage in ihate.rst

linhanzhang:
## Description
Add a section ``interactive usage`` in the page of plugin-ihate

### DIFF
--- a/docs/plugins/ihate.rst
+++ b/docs/plugins/ihate.rst
@@ -35,3 +35,11 @@ Here's an example::
             - album:christmas
 
 The plugin trusts your decision in "as-is" imports.
+
+Interactive Usage
+-------------
+
+- When the music library you import contains a album or item in``skip`` list, the console will prompt a "skipped: item information" message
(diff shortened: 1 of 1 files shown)


wisp3rwind:
Hey, thanks for working on improving our documentation! I do feel like this addition might be somewhat confusing, too: There is no actual interactivity in the plugin, which operates fully automatically. Its behaviour cannot be influenced at runtime. So, the information that you're really clarifying here (correct me if I'm wrong) is that in both cases (skip or warn lists), a message is logged to the terminal. I feel like that could equally well be expressed by saying something along the lines of

> In both cases, a message will be logged to the console for each imported album/singleton track that matches the given queries.

just before the `Here's an example'  line. What do you think?

linhanzhang:
> In both cases, a message will be logged to the console for each imported album/singleton track that matches the given queries.

Hey, I agree with you. I have remove the interactive usage part and add the description that you suggested

PR #4275, opened 2022-02-12
Fix embedart with compare_threshold on ImageMagick 7

wisp3rwind:
Fixes #4272. In the process, I found that I didn't like some details of the code for checking art similarity and `ArtResizer` in general, so this includes a fair bit of refactoring. For review, best read the commits individually and look at the commit messages.

PR #4313, opened 2022-03-12
Refactor: ArtResizer backends are classes

wisp3rwind:
A follow-up to https://github.com/beetbox/beets/pull/4275. Previously, `ArtResizer` had several backend -> method mappings, which seems to be a poor substitute for proper classes given that the `ArtResizer` has outgrown its name and supports many methods these days. This also allows for a cleaner separation of backend-specific concerns such as handling the modern/legacy ImageMagick invocations.

Some patterns in this code still feel a little clunky (such as the repeated checks for local/proxy backend), but in my opinion, this is at least a small improvement. Do you agree? In particular, I'd be interested if the last commit in this series, which moves some code from the thumbnails plugin to `ArtResizer`, seems sensible. Or is this not really universal enough to warrant moving to beets' core?

No functional changes, so no documentation/changelog/test updates should be required. As usual, this is best reviewed commit-by-commit: All commits should pass the tests, and only move limited amounts of code around such that the diff is still readable.

### DIFF
--- a/beets/util/artresizer.py
+++ b/beets/util/artresizer.py
@@ -221,6 +221,23 @@ class IMBackend(LocalBackend):
         except subprocess.CalledProcessError:
             return None
 
+    def convert_format(self, source, target, deinterlaced):
+        cmd = self.convert_cmd + [
+            syspath(source),
+            *(["-interlace", "none"] if deinterlaced else []),
+            syspath(target),
+        ]
+
+        try:
+            subprocess.check_call(
+                cmd,
+                stderr=subprocess.DEVNULL,
+                stdout=subprocess.DEVNULL
+            )
+            return target
+        except subprocess.CalledProcessError:
+            return source
+
 
 class PILBackend(LocalBackend):
     NAME="PIL"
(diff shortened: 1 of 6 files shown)


beets/util/artresizer.py
```diff
@@ -61,327 +57,409 @@ def temp_file_for(path):
...
-        else:
-            return path_out
-    except OSError:
-        log.error("PIL cannot create thumbnail for '{0}'",
-                  displayable_path(path_in))
-        return path_in
+class IMBackend(LocalBackend):
+    NAME = "ImageMagick"
+    _version = None
+    _legacy = None
```
sampsyo, line 80: Maybe worth a quick comment to indicate what this field means—i.e., whether we're using the `magick` binary?

beets/util/artresizer.py
```diff
@@ -503,13 +600,17 @@ def reformat(self, path_in, new_format, deinterlaced=True):
...
         # file path was removed
         result_path = path_in
         try:
-            result_path = func(self, path_in, path_new, deinterlaced)
+            result_path = self.local_method.convert_format(
+                path_in, path_new, deinterlaced
+            )
+        except Exception:
+            # FIXME: Should probably issue a warning?
+            pass
```
sampsyo, line 613: I could be missing something, but I think this method didn't previously suppress exceptions—the `finally` clause does cleanup even in the case of exceptions, but the exception was still propagated.

sampsyo (commented):
Wow; this looks awesome!! The `artresizer` module is certainly the victim of slowly increasing complexity: it started fairly simple and its design did not scale well. Adding backends made it progressively weirder, of course, but adding so many image-manipulating capabilities was the real problem. I really like this class-based approach, and everything looks in order from here—I have a couple of extremely minor comments, but I think this should be good to go.

beets/util/artresizer.py
wisp3rwind: You're right, the `finally` must have confused me — I've restored the old (and more sensible) behaviour now.

sampsyo (approved):
Looks great to me! I say ship it!! :tada:

PR #4314, opened 2022-03-13
README.md: HTTP => HTTPS

Schweinepriester:
## Description

Hi beets, long time no see 👋

Just a tiny change today, checked the link and made it HTTPS, skipping the redirect HTTP => HTTPS this way 0:-)

As [the link](https://www.w3.org/TR/html-markup/audio.html) is being redirected and most likely not showing what was intended before it potentially could even be replaced, for example with either of
- https://en.wikipedia.org/wiki/HTML5_audio
- https://html.spec.whatwg.org/multipage/media.html#the-audio-element
- https://caniuse.com/audio

I even went back and checked where the link originally pointed at: https://web.archive.org/web/20161125120800/http://www.w3.org/TR/html-markup/audio.html

Just say the word if I should change something :)

## To Do

- N/A Documentation. (If you've add a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)
- N/A Changelog. (Add an entry to `docs/changelog.rst` near the top of the document.)
- N/A Tests. (Encouraged but not strictly required.)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -59,7 +59,7 @@ shockingly simple if you know a little Python.
 .. _writing your own plugin:
     https://beets.readthedocs.org/page/dev/plugins.html
 .. _HTML5 Audio:
-    https://www.w3.org/TR/html-markup/audio.html
+    https://html.spec.whatwg.org/multipage/media.html#the-audio-element
 .. _albums that are missing tracks:
     https://beets.readthedocs.org/page/plugins/missing.html
 .. _duplicate tracks and albums:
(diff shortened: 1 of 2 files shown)


sampsyo:
Hello! Great to see you too! 😃

Thanks for noticing that this link had gone away. Of those options, I might like [the WHATWG link](https://html.spec.whatwg.org/multipage/media.html#the-audio-element) the best—it seems to be more-or-less a canonical way to refer to the element.

Schweinepriester:
Done :)

sampsyo:
Looks good to me; thanks!

PR #4316, opened 2022-03-13
Add test for function `summary` in `ihate.py`

doris962:
## Description

We add a test for function `summary` in `ihate.py`.

## To Do

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -68,6 +68,8 @@ For packagers:
 * We fixed a version for the dependency on the `Confuse`_ library.
   :bug:`4167`
 * The minimum required version of :pypi:`mediafile` is now 0.9.0.
+* We added a test function, named `test_summary` in the `test_ihate.py` file, that
+  increases the coverage of `test_iahate.py`.
 
 Other new things:
 
--- a/test/test_ihate.py
+++ b/test/test_ihate.py
@@ -5,6 +5,7 @@ import unittest
 from beets import importer
 from beets.library import Item
 from beetsplug.ihate import IHatePlugin
+from beetsplug.ihate import summary
 
 
 class IHatePluginTest(unittest.TestCase):
@@ -42,6 +43,17 @@ class IHatePluginTest(unittest.TestCase):
         match_pattern = ["album:testalbum genre:testgenre",
                          "artist:testartist album:notthis"]
         self.assertTrue(IHatePlugin.do_i_hate_this(task, match_pattern))
+
+    def test_summary(self):
+
+        # Task is not album
+        test_item = Item(
+            artist='TestArtist',
+            title='TestTitle')
+        task = importer.SingletonImportTask(None, test_item)
+        self.assertEqual(
+            summary(task),
+            'TestArtist - TestTitle')
 
 
 def suite():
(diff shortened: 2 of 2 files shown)


docs/changelog.rst
```diff
@@ -68,6 +68,8 @@ For packagers:
 * We fixed a version for the dependency on the `Confuse`_ library.
   :bug:`4167`
 * The minimum required version of :pypi:`mediafile` is now 0.9.0.
+* We added a test function, named `test_summary` in the `test_ihate.py` file, that
+  increases the coverage of `test_iahate.py`.
```
wisp3rwind, line 72: I think we usually don't track purely internal changes like this in the changelog (unless they address bugs, which is not the case here as far as I can tell). It's certainly irrelevant for packagers (assuming the test is well-written and works reliably on their systems).

test/test_ihate.py
```diff
@@ -43,6 +44,17 @@ def test_hate(self):
...
+
+        # Task is not album
+        test_item = Item(
+            artist='TestArtist',
+            title='TestTitle')
+        task = importer.SingletonImportTask(None, test_item)
+        self.assertEqual(
+            summary(task),
+            'TestArtist - TestTitle')
+
```
wisp3rwind, line 57: I'm not convinced that this is a useful test: It doesn't test any of the plugin's business logic, and the tested function is really very straightforward. The test only asserts that we're not inadvertently modifying the UI. However, this is not machine-readable output, and doesn't come with any stability guarantees in the first place. So, it'd say the only real result here is to make it more difficult for anybody who intentionally wants to modify the plugin's UI.

What was you motivation to add it?

wisp3rwind, line 57: I'm not convinced that this is a useful test: It doesn't test any of the plugin's business logic, and the tested function is really very straightforward. The test only asserts that we're not inadvertently modifying the UI. However, this is not machine-readable output, and doesn't come with any stability guarantees in the first place. So, it'd say the only real result here is to make it more difficult for anybody who intentionally wants to modify the plugin's UI.

What was you motivation to add it?

wisp3rwind (commented):
Hi, thanks for working on our test suite! However, I do have some concerns on whether this PR is really improving the situation, see the inline comments.

wisp3rwind:
> I just wanted to document the changes and thought this might be the most suitable place to add.

Personally, I'd argue that the git log (and the code itself, which is rather self-explanatory here) is sufficient documentation in this case.

> And for the test case, my motivation is to cover more lines of code in ihate.py, and the way I test is too straightforward which I have to admit. So could you please give me some hints for improvement?

Well, just "cover[ing] more lines" is probably not the most important reason to add tests. That's a simple metric to judge the extent of the test suite, but shouldn't be an end in itself. So the question is really, can we add tests that would really help preventing regressions and enable safe refactoring of our code? (There's certainly a lot of gaps in beets' testsuite where this can still be done.)

That aside, the largest untested chunk in the plugin is apparently the interaction with the importer, i.e. the `import_task_choice_event` hook. I guess it would be possible to come up with a useful test that checks the logic and logging from this function. For reference, we have test utilities for running dummy import sessions: Search for `create_importer` in the `test/` folder, in particular `test_importer.py` and `test_convert.py` (the actual function for creating these import sessions are defined in `test/helper.py`).

Issue #2686, opened 2017-09-10
Always prefer a different metadata source over MusicBrainz


0x1f415:
### Problem

I really dislike how musicbrainz normalizes capitalization across its entire database, ignoring the original stylization of the source material. I buy a lot of music from bandcamp, and would prefer to use the bandcamp plugin for metadata whenever possible. however, it seems there is not existing way to prefer an external metadata source _above_ musicbrainz, only _equal to_ musicbrainz, and even when both sources match 100% beets still prefers musicbrainz over bandcamp:

![image](https://user-images.githubusercontent.com/6064300/30252064-df1ffd18-9639-11e7-9acf-906e96a02844.png)

this could be alleviated if I could set the `source_weight` of bandcamp to a negative number, or if I could assign musicbrainz a `source_weight` of >0.0 (if this is already possible, it should be documented).

### Setup

* OS: windows 10, running beets in msys2
* Python version: 3.6.1
* beets version: 1.4.5
* Turning off plugins made problem go away (yes/no): haven't tried (disabling the beets plugin would be counterproductive)

My configuration (output of `beet config`) is:

```yaml
directory: /f/music/
library: /f/music/musiclibrary.lib

plugins: info discogs bandcamp edit badfiles missing #fetchart #copyartifacts
pluginpath:
    - ~/.config/beets/
    - ~/src/beets-bandcamp/beetsplug/
    - ~/src/beets-copyartifacts/beetsplug/

bandcamp:
    source_weight: 0

musicbrainz:
    source_weight: 0.5

copyartifacts:
    print_ignored: yes
    extensions: .*

ignore:
    - '*.mp4'
    - '*.avi'
paths:
    default: %ifdef{GROUPING,%ifdef{ERA,$GROUPING/$ERA/[$year-$month-$day] $albumartist - $album/$track $title,$GROUPING/$albumartist/[$year-$month-$day] $album/$track $title},$albumartist/[$year-$month-$day] $album/$track $title}
    singleton: %ifdef{GROUPING,%ifdef{ERA,$GROUPING/$ERA/[$year-$month-$day] $albumartist - $album/$track $title,$GROUPING/$albumartist/[$year-$month-$day] $album/$track $title},$albumartist/[$year-$month-$day] $album/$track $title}
[...]

PR #4319, opened 2022-03-15
Ability to optionally disable the musicbrainz metadata source

### LINKED ISSUE #400: Optionally disable the MusicBrainz data source
I'd like to perform an series of imports using beets but instead of using the musicbrainz autotagger I'd like to write my own matcher as a plugin.
There is a config option `import.autotag`, but this disables the tagger completely.

@sampsyo has suggested another config option to handle this, but it might be a little messy to implement, so I'm asking for some feedback.
My current idea would be to put a config option check in `beets/autotag/hooks.py`, in each of the methods that can be extended in the plugins, e.g. 
https://github.com/sampsyo/beets/blob/master/beets/autotag/hooks.py#L534
The alternative would be to move the check higher up to somewhere like https://github.com/sampsyo/beets/blob/master/beets/autotag/match.py#L409, but this might require some larger changes to the plugin system.

The other problem I can see is that there is a significant reliance on musicbrainz track ids, e.g. https://github.com/sampsyo/beets/blob/master/beets/autotag/match.py#L450. Any chance of working around these without a large rewrite?

snejus:
## Description

Fixes #400.  
Related to #2686

As I mentioned in https://github.com/beetbox/beets/issues/400#issuecomment-1067666254, this adds a patch that allows to disable the MusicBrainz metadata source during the import process.

I had a little go at making the `MusicBrainz Options` section in the docs a little bit more consistent, and copied over the entire default config section for clarity.

## To Do

### DIFF
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -598,7 +598,7 @@ def tracks_for_id(track_id):
             yield t
 
 
-def handle_exc(call_func, *args):
+def invoke_mb(call_func, *args):
     if not config["musicbrainz"]["enabled"]:
         return ()
 
@@ -622,11 +622,11 @@ def album_candidates(items, artist, album, va_likely, extra_tags):
     common_args = [album, len(items), extra_tags]
     # Base candidates if we have album and artist to match.
     if artist and album:
-        yield from handle_exc(mb.match_album, artist, *common_args)
+        yield from invoke_mb(mb.match_album, artist, *common_args)
 
     # Also add VA matches from MusicBrainz where appropriate.
     if va_likely and album:
-        yield from handle_exc(mb.match_album, None, *common_args)
+        yield from invoke_mb(mb.match_album, None, *common_args)
 
     # Candidates from plugins.
     yield from plugins.candidates(items, artist, album, va_likely, extra_tags)
@@ -641,7 +641,7 @@ def item_candidates(item, artist, title):
 
     # MusicBrainz candidates.
     if artist and title:
-        yield from handle_exc(mb.match_track, artist, title)
+        yield from invoke_mb(mb.match_track, artist, title)
 
     # Plugin candidates.
     yield from plugins.item_candidates(item, artist, title)
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -598,6 +598,16 @@ def tracks_for_id(track_id):
             yield t
 
 
+def handle_exc(call_func, *args):
+    if not config["musicbrainz"]["enabled"]:
+        return ()
+
+    try:
+        return call_func(*args)
+    except mb.MusicBrainzAPIError as exc:
+        exc.log(log)
+
+
 @plugins.notify_info_yielded('albuminfo_received')
 def album_candidates(items, artist, album, va_likely, extra_tags):
     """Search for album matches. ``items`` is a list of Item objects
@@ -609,25 +619,17 @@ def album_candidates(items, artist, album, va_likely, extra_tags):
     constrain the search.
     """
 
+    common_args = [album, len(items), extra_tags]
     # Base candidates if we have album and artist to match.
     if artist and album:
-        try:
-            yield from mb.match_album(artist, album, len(items),
-                                      extra_tags)
-        except mb.MusicBrainzAPIError as exc:
-            exc.log(log)
+        yield from handle_exc(mb.match_album, artist, *common_args)
 
     # Also add VA matches from MusicBrainz where appropriate.
     if va_likely and album:
-        try:
-            yield from mb.match_album(None, album, len(items),
-                                      extra_tags)
-        except mb.MusicBrainzAPIError as exc:
-            exc.log(log)
+        yield from handle_exc(mb.match_album, None, *common_args)
 
     # Candidates from plugins.
-    yield from plugins.candidates(items, artist, album, va_likely,
-                                  extra_tags)
+    yield from plugins.candidates(items, artist, album, va_likely, extra_tags)
 
 
 @plugins.notify_info_yielded('trackinfo_received')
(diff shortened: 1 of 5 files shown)


docs/reference/config.rst
```diff
@@ -716,6 +716,39 @@ Default: ``{}`` (empty).
 MusicBrainz Options
 -------------------
 
+Default configuration::
```
sampsyo, line 719: Just for my own sanity, could we keep this PR's docs changes focused on the new option? Some refactoring, reordering, and adding examples would be nice too, but it will be easier to think about in a separate PR. Thank you!

sampsyo (commented):
Awesome; thanks for getting this started! I actually think that it might be a little cleaner to handle this disabling in the `beets.autotag.hooks` module, which is the "dispatcher" that aggregates results from multiple sources. For example, this is the line that asks MB for matches:
https://github.com/beetbox/beets/blob/3f6da68af2444b77807692023abfd607f0146750/beets/autotag/hooks.py#L615

Maybe it would make a little more sense to skip those calls there, rather than having them all return None?

docs/reference/config.rst
snejus: Of course, I'll move them out to another branch!

beets/autotag/mb.py
```diff
@@ -482,6 +482,9 @@ def match_album(artist, album, tracks=None, extra_tags=None):
     The query consists of an artist name, an album name, and,
     optionally, a number of tracks on the album and any other extra tags.
     """
+    if not config["musicbrainz"]["enabled"].get(bool):
```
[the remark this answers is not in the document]
snejus: That's helpful, thanks!

snejus:
I do agree, will add the logic to `hooks.py` in a second.

snejus:
It did end up being cleaner once the error handling logic got abstracted away.

beets/autotag/hooks.py
```diff
@@ -609,25 +619,17 @@ def album_candidates(items, artist, album, va_likely, extra_tags):
...
 
+    common_args = [album, len(items), extra_tags]
     # Base candidates if we have album and artist to match.
     if artist and album:
-        try:
-            yield from mb.match_album(artist, album, len(items),
-                                      extra_tags)
-        except mb.MusicBrainzAPIError as exc:
-            exc.log(log)
+        yield from handle_exc(mb.match_album, artist, *common_args)
```
wisp3rwind, line 625: In my opinion, this pattern would be much more readable as

```python
    if artist and album and config["musicbrainz"]["enabled"]:
        yield from handle_exc(mb.match_album, artist, *common_args)
```

with the check for MB being enabled removed from `handle_exc`. This would separate exception handling and checking whether to query musicbrainz somewhat more cleanly.

If not going this way, renaming `handle_exc` -> `invoke_mb` or similar would also help.

snejus: Good point, @wisp3rwind. I've just renamed the function to `invoke_mb`

beets/autotag/hooks.py
```diff
@@ -598,6 +598,16 @@ def tracks_for_id(track_id):
             yield t
 
 
+def invoke_mb(call_func, *args):
+    if not config["musicbrainz"]["enabled"]:
+        return ()
```
sampsyo, line 603: I do sort of like @wisp3rwind's suggestion to consider moving this out of the `invoke_mb` utility—it might make it a little clearer what's going on!

beets/autotag/hooks.py
```diff
@@ -598,6 +598,16 @@ def tracks_for_id(track_id):
...
+
+    try:
+        return call_func(*args)
+    except mb.MusicBrainzAPIError as exc:
+        exc.log(log)
```
sampsyo, line 608: I think this might cause a crash when an exception occurs. Namely, it seems like this function needs to return an iterable, right? So I think this exception handler should probably `return ()` or similar…

beets/autotag/hooks.py
```diff
@@ -609,25 +619,17 @@ def album_candidates(items, artist, album, va_likely, extra_tags):
     constrain the search.
     """
 
+    common_args = [album, len(items), extra_tags]
```
sampsyo, line 622: I know this might be bikeshedding, but I actually think the repetition (passing the same set of arguments to the two calls to `mb.match_album`) is a little more intelligible.

beets/autotag/hooks.py
```diff
@@ -609,25 +619,17 @@ def album_candidates(items, artist, album, va_likely, extra_tags):
...
     # Base candidates if we have album and artist to match.
     if artist and album:
```
sampsyo, line 624: This stanza and the next could share a single `if config["musicbrainz"]["enabled"]:`.

docs/reference/config.rst
```diff
@@ -716,6 +716,17 @@ Default: ``{}`` (empty).
...
+.. _musicbrainz.enabled:
+
+enabled
+~~~~~~~
+
+This option allows you to disable using MusicBrainz as a metadata source. This applies
+if you use plugins that fetch data from alternative sources and should make the import
+process quicker.
+
+Default: ``yes``.
```
sampsyo, line 728: I think this documentation stuff might be out of order—it makes the "top-level" description below appear to be part of the "enabled" subsection. I think this should appear among the other subsections below (which start with "searchlimit").

sampsyo (commented):
Nice; looking good overall! I have a few low-level suggestions.

beets/autotag/hooks.py
wisp3rwind: I think I'm also fine with the current version (renamed to `invoke_mb`); I was mostly concerned about `handle_exc` being actively misleading since it does more than handling exceptions. So, from my side, I'd say it's up to you @snejus which variant you prefer.

beets/autotag/hooks.py
wisp3rwind: Nice catch! Yes, I'd suspect the same.

snejus: Well caught! Fixed in e1ffadb2

docs/reference/config.rst
snejus: Agree - have moved it below.

sampsyo (approved):
Looks great to me; thank you for your attention to detail here! 😃

PR #4327, opened 2022-03-27
Database parental directories creation if not exist

alicezou:
## Description

Add prompt for users to select whether to create parental directories when they do not exist.

## To Do

- [x] Tests. (Encouraged but not strictly required.) [still working on it]

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -1206,22 +1206,15 @@ def _configure(options):
               util.displayable_path(config.config_dir()))
     return config
 
-# Check whether parental directories exist.
 
-
-def database_dir_creation(path):
-    # Ask the user for a choice.
-    return input_yn("The database directory {} does not \
-                       exists. Create it (Y/n)?"
-                    .format(util.displayable_path(path)))
-
-
-def _check_db_directory_exists(path):
+def _ensure_db_directory_exists(path):
     if path == b':memory:':  # in memory db
         return
     newpath = os.path.dirname(path)
     if not os.path.isdir(newpath):
-        if database_dir_creation(newpath):
+        if input_yn("The database directory {} does not \
+                       exist. Create it (Y/n)?"
+                    .format(util.displayable_path(newpath))):
             os.makedirs(newpath)
 
 
@@ -1229,7 +1222,7 @@ def _open_library(config):
     """Create a new library instance from the configuration.
     """
     dbpath = util.bytestring_path(config['library'].as_filename())
-    _check_db_directory_exists(dbpath)
+    _ensure_db_directory_exists(dbpath)
     try:
         lib = library.Library(
             dbpath,
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -1209,12 +1209,18 @@ def _configure(options):
 # Check whether parental directories exist.
 
 
+def database_dir_creation(path):
+    # Ask the user for a choice.
+    return input_yn("The database directory {} does not \
+                       exists. Create it (Y/n)?"
+                    .format(util.displayable_path(path)))
+
+
 def _check_db_directory_exists(path):
     if path == b':memory:':  # in memory db
         return
     newpath = os.path.dirname(path)
     if not os.path.isdir(newpath):
-        from beets.ui.commands import database_dir_creation
         if database_dir_creation(newpath):
             os.makedirs(newpath)
 
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -1206,11 +1206,24 @@ def _configure(options):
               util.displayable_path(config.config_dir()))
     return config
 
+# Check whether parental directories exist.
+
+
+def _check_db_directory_exists(path):
+    if path == b':memory:':  # in memory db
+        return
+    newpath = os.path.dirname(path)
+    if not os.path.isdir(newpath):
+        from beets.ui.commands import database_dir_creation
+        if database_dir_creation(newpath):
+            os.makedirs(newpath)
+
 
 def _open_library(config):
     """Create a new library instance from the configuration.
     """
     dbpath = util.bytestring_path(config['library'].as_filename())
+    _check_db_directory_exists(dbpath)
     try:
         lib = library.Library(
             dbpath,
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1404,7 +1404,7 @@ default_commands.append(version_cmd)
 
 def database_dir_creation(path):
     # Ask the user for a choice.
-    return ui.input_yn("{} does not exist, create it (Y/n)?"
+    return ui.input_yn("The database directory {} does not exists, create it (Y/n)?"
                        .format(displayable_path(path)))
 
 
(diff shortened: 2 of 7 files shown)


beets/ui/commands.py
```diff
@@ -1398,6 +1398,15 @@ def show_version(lib, opts, args):
 version_cmd.func = show_version
 default_commands.append(version_cmd)
 
+# database_location: return true if user
+# wants to create the parent directories.
+
+
+def database_dir_creation(path):
+    # Ask the user for a choice.
+    return ui.input_yn("{} does not exist, create it (Y/n)?"
```
wisp3rwind, line 1407: It might be helpful to more explicity describe the issue here; the path alone might not be enough to understand the problem: `The database directory {} does not exists, ...`?

test/test_dbcore.py
```diff
@@ -760,8 +762,31 @@ def test_no_results(self):
...
 
 
+class ParentalDirCreation(unittest.TestCase):
+    @mock.patch('builtins.input', side_effect=['y', ])
+    def test_create_yes(self, _):
+        non_exist_path = "ParentalDirCreationTest/nonexist/" + str(random())
+        try:
+            dbcore.Database(non_exist_path)
+        except OSError as e:
+            raise e
```
wisp3rwind, line 772: That would behave identically with the `try ... catch` removed, right?

test/test_dbcore.py
```diff
@@ -760,8 +762,31 @@ def test_no_results(self):
             ModelFixture1, dbcore.query.FalseQuery()).get())
 
 
+class ParentalDirCreation(unittest.TestCase):
+    @mock.patch('builtins.input', side_effect=['y', ])
+    def test_create_yes(self, _):
+        non_exist_path = "ParentalDirCreationTest/nonexist/" + str(random())
```
wisp3rwind, line 768: I think this should be using a path within a temporary directory, to be certain that we don't pollute the environment where the test is run.

test/test_dbcore.py
```diff
@@ -760,8 +762,31 @@ def test_no_results(self):
...
+        try:
+            dbcore.Database(non_exist_path)
+        except OSError as e:
+            raise e
+        shutil.rmtree("ParentalDirCreationTest")
+
+    @mock.patch('builtins.input', side_effect=['n', ])
```
wisp3rwind, line 775: It might be easier to use the `control_stdin` context manager from `test/helper.py` (see its docstring for details).

wisp3rwind (changes requested):
Hi, thanks for  getting this started! I've left a few inline comments on how I think the design of this PR might be improved.

alicezou:
Hi Benedikt,
Thank you for your review! I have updated my commits.

beets/ui/commands.py
```diff
@@ -1398,6 +1398,16 @@ def show_version(lib, opts, args):
...
 default_commands.append(version_cmd)
 
+# database_location: return true if user
+# wants to create the parent directories.
+
+
+def database_dir_creation(path):
+    # Ask the user for a choice.
+    return ui.input_yn("The database directory {} does not \
+                       exists, create it (Y/n)?"
```
sampsyo, line 1408:
```suggestion
                       exists. Create it (Y/n)?"
```

So it's two full sentences.

docs/changelog.rst
```diff
@@ -7,6 +7,7 @@ Changelog
 Changelog goes here!
 
 New features:
+* Create the parental directories for database if they do not exist.
```
sampsyo, line 11:
```suggestion
* Create the parental directories for database if they do not exist.
  :bug:`3808` :bug:`4327`
```

Including references to where this was discussed.

docs/guides/main.rst
```diff
@@ -142,7 +142,8 @@ place to start::
 
 Change that first path to a directory where you'd like to keep your music. Then,
 for ``library``, choose a good place to keep a database file that keeps an index
-of your music. (The config's format is `YAML`_. You'll want to configure your
+of your music. Beets will prompt you if the parental directories for database do
+not exist. (The config's format is `YAML`_. You'll want to configure your
```
sampsyo, line 146: TBH I think it's probably OK to leave this sentence off? This is an edge case, and the prompt clearly explains to users what's going on, so I don't think we need to warn people that this might happen.

sampsyo (commented):
Looking great!! Here are a few granular suggestions.

I actually think it would be fine (and a bit simpler) to keep the prompt code directly in the `beets.ui` module (i.e., `ui/__init__.py`) rather than having a short function in the `beets.ui.commands` module. The latter module is centered around providing implementations for the `beet <foo>` subcommands, and this functionality is actually a global thing that is not associated with any particular subcommand—so it would be a bit simpler to just keep everything in the new `_check_db_directory_exists` function.

I also just wanted to state for the record that this issue will *probably* only affect people who have changed the `library` configuration option… beets automatically creates `~/.config/beets` by default and puts `library.db` in there. But custom database locations would not enjoy this automation (until now).

alicezou:
Hi Adrian,
Thank you for the feedback! I have updated the related files.

alicezou:
Hi, 
It's been a week since my last update. Are there any changes that I can make to improve this pull request? I am looking forward to hearing from you. :)
@wisp3rwind @sampsyo

beets/ui/commands.py
```diff
@@ -1398,6 +1398,9 @@ def show_version(lib, opts, args):
 version_cmd.func = show_version
 default_commands.append(version_cmd)
 
+# database_location: return true if user
+# wants to create the parent directories.
+
```
sampsyo, line 1403: This comment looks like it's out of date (it can just be deleted).

beets/ui/__init__.py
```diff
@@ -1206,11 +1206,30 @@ def _configure(options):
...
     return config
 
+# Check whether parental directories exist.
+
+
+def database_dir_creation(path):
+    # Ask the user for a choice.
+    return input_yn("The database directory {} does not \
+                       exists. Create it (Y/n)?"
+                    .format(util.displayable_path(path)))
```
sampsyo, line 1216: I suggest just "inlining" this code, since it's pretty simple and not called anywhere else. (That is, just do this prompting right inside the function below instead of having a separate function.)

beets/ui/__init__.py
```diff
@@ -1206,11 +1206,30 @@ def _configure(options):
...
+
+
+def _check_db_directory_exists(path):
```
sampsyo, line 1219: Maybe should be called `_ensure_db_directory_exists` since it does more than just check?

beets/ui/__init__.py
```diff
@@ -1206,11 +1206,30 @@ def _configure(options):
               util.displayable_path(config.config_dir()))
     return config
 
+# Check whether parental directories exist.
+
+
+def database_dir_creation(path):
+    # Ask the user for a choice.
+    return input_yn("The database directory {} does not \
+                       exists. Create it (Y/n)?"
```
sampsyo, line 1215: exists -> exist

sampsyo (changes requested):
Looking good!! I have just a few very small comments, but then I think this should be good to go.

alicezou:
Hi Adrian, 
Thank you for your comments! I have updated the relevant parts.

sampsyo (approved):
Looks perfect. Thanks again!!

Issue #4325, opened 2022-03-23
Disabling plugins CLI option


JOJ0:
The current CLI option:
```
-p PLUGINS, --plugins=PLUGINS
                        a comma-separated list of plugins to load
```
is tedious to use when many plugins are in use. Often a user might just want to disable a single plugin or a tiny list of plugins to try something/do something in a specific beets run. 

### Proposed solution

Add a CLI option named eg. `--disable-plugins/-P` that allows mentioning a comma separated list of plugin-names to stay disabled for that beets run.

sampsyo:
Sounds OK to me! It is probably worth noting that `--plugins` already can disable _all_ plugins, so I suppose this would just be about disabling one plugin?

It could also be worth looking at the #3765 thread for inspiration about how to tackle this.

arbanhossain:
Hey, if I am understanding this right, I sort of have an implementation of this in my fork. It uses `-x/--exclude`. Should I do a PR for you to review?

sampsyo:
A PR would be great!

arsaboo:
Can we close this issue since #4332 is merged?

sampsyo:
Yes, thanks for the reminder!!

PR #4332, opened 2022-04-15
Add --exclude flag

arbanhossain:
## Description

Add a `--exclude` (or `-x`) flag to allow disabling of one/multiple plugin(s). Related issue: #4325 

## To Do

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -1127,6 +1127,10 @@ def _load_plugins(options, config):
                        if len(options.plugins) > 0 else [])
     else:
         plugin_list = config['plugins'].as_str_seq()
+    
+    # Exclude any plugins that were specified on the command line
+    if options.exclude is not None:
+        plugin_list = [p for p in plugin_list if p not in options.exclude.split(',')]
 
     plugins.load_plugins(plugin_list)
     return plugins
@@ -1261,6 +1265,8 @@ def _raw_main(args, lib=None):
                       help='path to configuration file')
     parser.add_option('-p', '--plugins', dest='plugins',
                       help='a comma-separated list of plugins to load')
+    parser.add_option('-x', '--exclude', dest='exclude',
+                      help='a comma-separated list of plugins to disable')
     parser.add_option('-h', '--help', dest='help', action='store_true',
                       help='show this help message and exit')
     parser.add_option('--version', dest='version', action='store_true',
--- a/docs/reference/cli.rst
+++ b/docs/reference/cli.rst
@@ -445,6 +445,10 @@ import ...``.
   specified, the plugin list in your configuration is ignored. The long form
   of this argument also allows specifying no plugins, effectively disabling
   all plugins: ``--plugins=``.
+* ``-x EXCLUDE``: specify a comma-separated list of plugins to disable in a
+  specific beets run. If specified, it will exclude plugins from your configuration
+  and/or plugins specified using the ``-p`` flag. To disable ALL plugins, use
+  ``--plugins=`` instead.
 
 Beets also uses the ``BEETSDIR`` environment variable to look for
 configuration and data.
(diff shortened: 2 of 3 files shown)


sampsyo:
Neat; thank you for getting this started!!

I would prefer to spell the option a little more explicitly: `--no-plugin` or `--exclude-plugin` or `--disable-plugin` and to use the short flag `-P` to convey the "opposite" of `-p`. There are a lot of things we could be excluding (files? albums?), so being a little more explicit seems important.

arbanhossain:
Okay, I've pushed the changes.

docs/reference/cli.rst
```diff
@@ -445,6 +445,10 @@ import ...``.
   specified, the plugin list in your configuration is ignored. The long form
   of this argument also allows specifying no plugins, effectively disabling
   all plugins: ``--plugins=``.
+* ``-P EXCLUDE``: specify a comma-separated list of plugins to disable in a
```
sampsyo, line 448: Use `-P plugins`, as with the previous bullet point.

docs/reference/cli.rst
```diff
@@ -445,6 +445,10 @@ import ...``.
...
+  specific beets run. If specified, it will exclude plugins from your configuration
+  and/or plugins specified using the ``-p`` flag. To disable ALL plugins, use
```
sampsyo, line 450: Not sure we need the second sentence? It seems like this is pretty logical.

sampsyo, line 450: `ALL` -> `*all*`

beets/ui/__init__.py
```diff
@@ -1261,6 +1266,8 @@ def _raw_main(args, lib=None):
                       help='path to configuration file')
     parser.add_option('-p', '--plugins', dest='plugins',
                       help='a comma-separated list of plugins to load')
+    parser.add_option('-P', '--disable-plugin', dest='exclude',
```
sampsyo, line 1269: Argh, sorry for the bad advice!! It should be `--disable-plugins`, i.e., symmetric with the "positive" flag name.

beets/ui/__init__.py
```diff
@@ -1128,6 +1128,11 @@ def _load_plugins(options, config):
     else:
         plugin_list = config['plugins'].as_str_seq()
 
+    # Exclude any plugins that were specified on the command line
+    if options.exclude is not None:
+        plugin_list = [p for p in plugin_list
+                      if p not in options.exclude.split(',')]
```
sampsyo, line 1134: Just making sure: does this work if `-P` is not specified, even without something like `default=''`? I would have expected this to crash because `None` does not have a method `split`…

sampsyo (commented):
Great!! Here are a few comments. Looks like flake8 has one small nitpick as well.

beets/ui/__init__.py
arbanhossain: I'm not sure I follow you here. Do you mean if it works when I input `--disable-plugins=` or when I just run a command without `-P` at all like `beet list`?

docs/reference/cli.rst
arbanhossain: I agree there is some redundancy here, but I think it is worth mentioning that it overwrites `-p`?

sampsyo: Ah, sure, that seems worth preserving.

beets/ui/__init__.py
sampsyo: The latter—surely it must work because the tests seem to pass! I'm just a little surprised that the case: i.e., I would have expected leaving `-P` off altogether would crash.

arbanhossain: It works. I believe it runs only if the option is `not None`. I will post some logs for verification if that's okay?

sampsyo: Oh jeez, I totally didn't see the `if options.exclude is not None:` conditional. That would do it!! Sorry for being slow.

sampsyo:
Looking good to me, except for the extremely nitpicky linter output! Sorry about that; it wants you to remove one space.

arbanhossain:
Oh, that's honestly my bad. I totally forgot to check that. I am changing that

sampsyo:
This looks great! Thank you for your persistence!! :sparkles: :smiley:

PR #4333, opened 2022-04-15
Genius lyrics header bug fixed and updated test case for lyrics plugin

### LINKED ISSUE #4305: lyrics: Genius backend includes heading in lyrics
### Problem

Lyrics fetched from the Genius backend include text from the heading above the lyrics
For example, [this page](https://genius.com/Ttng-chinchilla-lyrics) results in
```
Chinchilla LyricsSo far we've lost focus
Let's just concentrate on words that could mean everything
...
```
AFAIK this heading is always `<track> Lyrics` so there's no reason for us to keep it in.

olgarrahan:
Fixes #4305  

Lyrics fetched in Genius backend no longer include extraneous header information

Tests: changed test_good_lyrics file in test_lyrics.py to test html from updated Genius html


wisp3rwind:
Looks good at a first glance, let's wait for the CI to run!

I suspect that the fallback code after this change (`if not lyrics_div: ...`) might be outdated, too, but I guess we can just merge this as-is and deal with that later, if necessary.

beetsplug/lyrics.py
```diff
@@ -401,7 +401,12 @@ def _scrape_lyrics_from_html(self, html):
         # all of the lyrics can be found already correctly formatted
         # Sometimes, though, it packages the lyrics into separate divs, most
         # likely for easier ad placement
-        lyrics_div = soup.find("div", class_="lyrics")
+
+        lyrics_div = soup.find("div", {"data-lyrics-container": True})
+
+        for br in lyrics_div.find_all("br"):
```
sampsyo, line 407: Looks like some tests are failing here because `lyrics_div` may be `None`. Maybe this just needs to go inside the `if` block below?

sampsyo (commented):
Thanks for getting this started! I added one inline note pointing to the likely source of the test failure.

olgarrahan:
Got it, just made the fix!

beetsplug/lyrics.py
```diff
@@ -401,7 +401,13 @@ def _scrape_lyrics_from_html(self, html):
         # all of the lyrics can be found already correctly formatted
         # Sometimes, though, it packages the lyrics into separate divs, most
         # likely for easier ad placement
-        lyrics_div = soup.find("div", class_="lyrics")
+
+        lyrics_div = soup.find("div", {"data-lyrics-container": True})
+
+        if lyrics_div:
+            for br in lyrics_div.find_all("br"):
+                br.replace_with("\n")
```
sampsyo, line 409: I'm not sure this is important enough to change, but we could reduce some duplication by factoring this out into a function (named `replace_br` or something?), which we'd also use below in the per-verse case.

sampsyo (approved):
Overall this looks good; I think it's good to go! I have one thought about a possible refactoring to reduce duplicated code, but it may not even really be worth it…

beetsplug/lyrics.py
olgarrahan: Just pushed a commit with a replace_br function to reduce code duplication

sampsyo:
All looks good! I think all we need is a changelog entry and we should be good to go? (Sorry I didn't notice that was still outstanding earlier!)

olgarrahan:
Yes, just updated the changeling!

sampsyo:
Looks good from here. Thank you for your work on this!!

PR #4335, opened 2022-04-18
Front-end improvements for the Web plug-in

martinkirch:
Hello,

As a long-time user of the web plug-in, I've made a few changes to its front-end that might be worth merging.

This PR:
 * adds the `controls` attribute to the `audio` element - it's the simplest way to have play/pause buttons, play times, volume and navigation bar
 * adds a `Newest music` link so I don't have to search `added:-1w..` manually
 * fixes an unicode bug
 * makes the results list sortable (thanks to [SortableJS](http://sortablejs.github.io/Sortable/))
 * makes individual items removable
 * lets the use choose between replacing or adding search results to current items in the left bar

Overall, this turns the page into a rudimentary but decent playlist.


sampsyo:
Super cool-looking stuff; thanks for getting in touch! I would be interested to hear from people who are more active users of the web interface. In general, it's a bit easier for maintainers to review one new feature at a time rather than an "omnibus" of changes over time. (Of course, I totally understand why you organized it this way! You've been hard at work over some time and it would take work to disentangle the pieces.) So depending on interest/demand from users, maybe we can focus our work on the most desired pieces of your implementation?

Perhaps posting on our Discussions section (or Discourse forum) would be a good way to get some feedback from other web interface users!

martinkirch:
Indeed I accumulated a few things here. But I could easily roll back each point, depending on the community's interest. I've started #4336

sampsyo:
Let's keep this alive! I still think it would be cool to break this into a few smaller components, and to gauge community interest in each piece.

sampsyo:
Reviving once again.

PR #4337, opened 2022-04-21
chore(ci): add informational Codecov status checks

thomasrockhu-codecov:
## Description
Hi, Tom from Codecov here. I noticed that you were using Codecov but weren't actually getting any notifications on pull requests. I figured it would be useful to get some idea if code being changed is being tested, but also not blocking CI/merging. Let me know if this makes sense or if we can do something that would be helpful.

## To Do

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,6 +2,7 @@
 comment: off
 
 # Sets non-blocking status checks
+# https://docs.codecov.com/docs/commit-status#informational
 coverage:
   status:
     project:
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,9 +1,13 @@
 # Don't post a comment on pull requests.
 comment: off
 
-# I think this disables commit statuses?
+# Sets non-blocking status checks
 coverage:
-    status:
-        project: no
-        patch: no
-        changes: no
+  status:
+    project:
+      default:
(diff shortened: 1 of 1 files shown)


wisp3rwind:
We disabled those on purpose a while back, since the reports seemed overly intrusive (https://github.com/beetbox/beets/commit/0395e7fd8095a1c03d2b9d1c3c0576ea4b95ce6e). Is there any documentation (or rather, examples) on what they look like nowadays?

wisp3rwind:
Ok, thanks for clarification. Let's wait for a few more opinions on this one, though.

sampsyo:
OK; thanks for looking into this! Let's at least give it a shot and see how it looks. I've installed the Codecov GitHub app; presumably that will make this PR start showing the status on next update?

sampsyo:
Hmm… oddly, I'm not seeing anything here! The normal GitHub Actions are running as usual, but I don't see anything from Codecov…

wisp3rwind:
Screenshot from 2022-04-24 13-01-42

Seems to work here.

sampsyo:
Got it! I think I just didn't wait long enough. Looks good to me; let's give it a shot! Thanks again for your help.

PR #4349, opened 2022-05-08
Use config option for Spotify

arsaboo:
## Description

Partially Fixes #4347.  

Use config option for Spotify. Tested with the following configs and both work:
```
spotify:
    source_weight: 0
    client_id: REDACTED
    client_secret: REDACTED
```

```
spotify:
    client_id: REDACTED
    client_secret: REDACTED
```

## To Do

### DIFF
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -61,8 +61,8 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
                 'track_field': 'title',
                 'region_filter': None,
                 'regex': [],
-                'client_id': '4e414367a1d14c75a5c5129a627fcab8',
-                'client_secret': 'f82bdc09b2254f1a8286815d02fd46dc',
+                'client_id': config['spotify']['client_id'].as_str(),
+                'client_secret': config['spotify']['client_secret'].as_str(),
                 'tokenfile': 'spotify_token.json',
             }
         )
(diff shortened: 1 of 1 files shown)


arsaboo:
Tests are failing and, unfortunately, I have no idea how to fix them :(

sampsyo:
Hmm… I'm a tiny bit confused here. Because this is already in a `config.add` call, it should in fact already be possible to override these defaults in your configuration, just as you've outlined above. Have you given that a try without any changes to the code? (If so, what happens?)

arsaboo:
Yes, it works even without any change to the code. with the above config. I thought you wanted to remove the hard-coded id/secret so that we don't overload them.

sampsyo:
Ah, I see! Do we know that there's a problem yet with sharing API keys? For some services, at least, baking in an API key for a specific project is exactly what the API provider wants us to do…

The way to make a config option *required*, FWIW, is just to not provide a default. So we would just delete those lines from the `config.add` call altogether.

arsaboo:
I don't think it should be a problem. I am also associated with Home Assistant and they also require a developer account and they have not included the key in their code

sampsyo:
I guess what I was trying to clarify is: have you (or anyone else) been experiencing any problems with the status quo?

arsaboo:
No, I am not. We can leave it as is. Maybe I misunderstood your comment. I think I will close this PR as it is not adding much. 

Now that we have the track_id, it will be great if we can now have the other audio features using the `tracks` API.

sampsyo:
Maybe we should chat a bit more in #4347 a bit more about exactly what the design should look like? (For example, one good way to start could be with the existing old PR and incrementally improving that…)

Issue #3578, opened 2020-05-03
Get audio features from spotify


flovanc:
hello, i'm a Home DJ who use beets for tag music with mostly beatport plugin and thanks for this great feature .

Since i have discover http://www.playlistmachinery.com/ i use these tools for adding tag like Genre, Energy, Danceability and Valence in the comment section of id3 for help me to make smooth playlist for my DJ software "traktor"

for do this i ..

- in 1 add music in "to tag"playlist inside spotify 

- in 2 open http://organizeyourmusic.playlistmachinery.com/#

- in 3 manually copy those tag to the comment section of id3 tag

needless to say it takes a very long time  , so I'm looking for a way to automate everything.

is anyone interested in this and would like to help me implement this with the help of the spotify API ?

https://developer.spotify.com/documentation/web-api/reference/tracks/get-audio-features/

FLO

sampsyo:
Hi! Have you seen the AcousticBrainz plugin that ships with beets? It provides another (open-source) way to get these acoustic metrics.

flovanc:
yes it only works with musicbrainz. But in my case I use 100% the beatport plugin and forbid access to musicbrainz, because they are new releases of electronic music from beatport.

sampsyo:
Aha, that's too bad. But it looks like the Spotify API there requires a Spotify ID—do your files have that?

Laski:
@flovanc if your files don't have an Spotify ID you can probably import or adapt the code from https://github.com/beetbox/beets/blob/master/beetsplug/spotify.py in order to get the IDs first.

P.S: Just discovered all the data that is available in the Spotify API, thanks! Now I'm looking forward to having an automated way to import that data into my library too. Let me know how can I help you.

flovanc:
No i don't have spotify ID inside my files .
i make a try to import with spotify plugin but i have this error:

> beet -vv import /home/payason/downloads/music\ in
user configuration: /home/payason/.config/beets/config.yaml
data directory: /home/payason/.config/beets
plugin paths:
Sending event: pluginload
library database: /home/payason/.config/beets/musiclibrary.db
library directory: /home/payason/Music
Sending event: library_opened
Sending event: import_begin
Sending event: import_task_created
Sending event: import_task_start
Looking up: /home/payason/downloads/music in/DEADWALKMAN - Valleys Of Saturn EP
Tagging DEADWALKMAN - Valleys Of Saturn EP
No album ID found.
[...]

PR #4351, opened 2022-05-12
Obtain Spotify track (popularity and audio features) attributes

arsaboo:
This PR introduces a new command `beet spotifysync` to obtain all the audio features available for a track. This includes the `popularity` information and `audio_features`. The following `audio_features` are included: acousticness, danceability, energy, instrumentalness, key, liveness, loudness, mode, speechiness, tempo, time_signature, valence. See details [here](https://developer.spotify.com/documentation/web-api/reference/#/operations/get-audio-features). 

This features only processes tracks that already have the spotify_track_id (which was introduced in #4348). So, the workflow will be to run `beet import` which will populate the `spotify_track_id`. This information can then be used to run `beet spotifysync`. 

Fixes #4347 #3578 #4094

Config required:

```
spotifysync:
    client_id: SPOTIFY_CLIENT_ID
    client_secret: SPOTIFY_CLIENT_SECRET
    force: no
```

My hope is that we get the initial PR so that folks can start using it and improve upon it. 

## To Do

### DIFF
--- a/beetsplug/spotifysync.py
+++ b/beetsplug/spotifysync.py
@@ -160,7 +160,7 @@ class SpotifySyncPlugin(BeetsPlugin):
                 audio_features = self.track_audio_features(item.spotify_track_id)
                 for key in audio_features.keys():
                     self._log.info('key: {}',SPOTIFY_AUDIO_FEATURES[key][0])
-                    self._log.info('audio feature: {}',audio_features[key])
+                    self._log.info('audio feature: {} for {}',audio_features[key], item)
                     item[SPOTIFY_AUDIO_FEATURES[key][0]] = audio_features[key]
                 # item['spotify_track_acousticness'] = audio_features["acousticness"]
                 # item['spotify_track_danceability'] = audio_features["danceability"]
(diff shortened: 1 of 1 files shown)


sampsyo:
Nice; thanks for getting this started! What do you think about rolling this new command directly into the existing Spotify plugin? That could avoid a bunch of the duplicated code and allow users to share the token between the two bits of functionality, for example.

arsaboo:
I am not 100% versed with the existing Spotify plugin. It may be possible to combine the features, but I am a little hesitant. The existing plugin is written for the auto-tagger and serves the basic purpose of getting the tag information from Spotify. 

Also, not everyone may require all these features (although that can be addressed with a config option). Unless you are completely opposed to this idea, it may be a good idea to keep it separate so that additional features can be added here.

arsaboo:
Also note that it will make the importer significantly slow due to the timeouts and the additional processing required.

sampsyo:
Ah, to be clear, I just meant adding this command to the Spotify plugin. That would mean you'd still have to run the command explicitly to fetch the new data; it would not run as part of the import. In the code, this is just a matter of defining the `commands` method of the existing plugin appropriately.

I don't have a strong opinion about whether it's one or two plugins _per se_, but I do strongly prefer not to duplicate code... and keeping it in one plugin seems most likely to avoid that duplication.

arsaboo:
Is there any plugin template that I can follow?

arsaboo:
closing in favor of #4352

PR #4352, opened 2022-05-13
Extend Spotify plugin to obtain (popularity and audio features) track attributes

arsaboo:
This PR extends the `spotify` plugin to add another command `beet spotifysync` to obtain all the audio features available for a track. This includes the `popularity` information and `audio_features`. The following `audio_features` are included: acousticness, danceability, energy, instrumentalness, key, liveness, loudness, mode, speechiness, tempo, time_signature, valence. See details [here](https://developer.spotify.com/documentation/web-api/reference/#/operations/get-audio-features). 

This command only processes tracks that already have the `spotify_track_id` (which was introduced in #4348). So, the workflow will be to run `beet import` to obtain the `spotify_track_id`. This information can then be used to run `beet spotifysync`. 

Fixes #4347 #3578 #4094

optional config:

```
spotify:
    client_id: SPOTIFY_CLIENT_ID
    client_secret: SPOTIFY_CLIENT_SECRET
```

## To Do

### DIFF
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -251,6 +251,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
             artist=artist,
             artist_id=artist_id,
             spotify_artist_id=artist_id,
+            spotify_track_popularity=track_data['popularity'],
             length=track_data['duration_ms'] / 1000,
             index=track_data['track_number'],
             medium=track_data['disc_number'],
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -251,6 +251,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
             artist=artist,
             artist_id=artist_id,
             spotify_artist_id=artist_id,
+            spotify_track_popularity=track_data['tracks']['popularity'],
             length=track_data['duration_ms'] / 1000,
             index=track_data['track_number'],
             medium=track_data['disc_number'],
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -413,8 +413,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
 
         def func(lib, opts, args):
             items = lib.items(ui.decargs(args))
-            self._fetch_info(items, ui.should_write(),
-                             opts.force_refetch or self.config['force'])
+            self._fetch_info(items, ui.should_write(), opts.force_refetch)
 
         sync_cmd.func = func
         return [spotify_cmd, sync_cmd]
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -597,8 +597,6 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
                     self.track_audio_features(item.spotify_track_id)
                 for feature in audio_features.keys():
                     if feature in spotify_audio_features.keys():
-                        self._log.info('feature: {}',
-                                       audio_features[feature])
                         item[spotify_audio_features[feature][0]] = \
                             audio_features[feature]
                 item.store()
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -586,9 +586,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
                 # If we're not forcing re-downloading for all tracks, check
                 # whether the popularity data is already present
                 if not force:
-                    spotify_track_popularity = \
-                        item.get('spotify_track_popularity', '')
-                    if spotify_track_popularity:
+                    if 'spotify_track_popularity' in item:
                         self._log.debug('Popularity already present for: {}',
                                         item)
                         continue
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -612,8 +612,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
             requests.get, self.track_url + track_id
         )
         self._log.debug('track_data: {}', track_data['popularity'])
-        track_popularity = track_data['popularity']
-        return track_popularity
+        return track_data['popularity']
 
     def track_audio_features(self, track_id=None):
         """Fetch track audio features by its Spotify ID."""
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -356,6 +356,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
         return response_data
 
     def commands(self):
+        # autotagger import command
         def queries(lib, opts, args):
             success = self._parse_opts(opts)
             if success:
@@ -382,7 +383,23 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
             ),
         )
         spotify_cmd.func = queries
-        return [spotify_cmd]
+
+        # spotifysync command
+        sync_cmd = ui.Subcommand('spotifysync',
+                                 help="fetch track attributes from Spotify")
+        sync_cmd.parser.add_option(
+            '-f', '--force', dest='force_refetch',
+            action='store_true', default=False,
+            help='re-download data when already present'
+        )
+
+        def func(lib, opts, args):
+            items = lib.items(ui.decargs(args))
+            self._fetch_info(items, ui.should_write(),
+                             opts.force_refetch or self.config['force'])
+
+        sync_cmd.func = func
+        return [spotify_cmd, sync_cmd]
 
     def _parse_opts(self, opts):
         if opts.mode:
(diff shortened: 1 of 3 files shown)


arsaboo:
@sampsyo I was thinking that instead of `force`, we should have separate overwrite flags. `audio_features` are static and may not have to be updated, whereas `popularity` is dynamic and users may want to update them. We don't have to tie them together. In addition, we may want to add other attributes later. Not everyone may want all the attributes, so it may be better to decouple them. Here's the config that I have in mind (with defaults indicated):
```yaml
spotify:
    client_id: SPOTIFY_CLIENT_ID
    client_secret: SPOTIFY_CLIENT_SECRET
    popularity: yes
    popularity_overwrite: no
    audio_features: no
    audio_features_overwrite: no
```

beetsplug/spotify.py
```diff
@@ -64,6 +66,7 @@ def __init__(self):
                 'client_id': '4e414367a1d14c75a5c5129a627fcab8',
                 'client_secret': 'f82bdc09b2254f1a8286815d02fd46dc',
                 'tokenfile': 'spotify_token.json',
+                'force': False,
```
sampsyo, line 69: Seems OK to make this a CLI-only flag for now, and omit the configuration option. It will be a little less confusing that way because the `force` option *only* applies to the new command, not the remaining functionality of the plugin.

beetsplug/spotify.py
```diff
@@ -536,3 +556,71 @@ def _output_match_results(self, results):
...
+            'instrumentalness': ['spotify_track_instrumentalness'],
+            'key': ['spotify_track_key'],
+            'liveness': ['spotify_track_liveness'],
+            'loudness': ['spotify_track_loudness'],
+            'mode': ['spotify_track_mode'],
+            'speechiness': ['spotify_track_speechiness'],
+            'tempo': ['spotify_track_tempo'],
+            'time_signature': ['spotify_track_time_sig'],
+            'valence': ['spotify_track_valence'],
+        }
```
sampsyo, line 575: Maybe this should be a module-level constant?

Also, I'm not 100% clear what he values here are lists instead of single strings. Is the idea that there may be multiple mappings per attribute, someday?

beetsplug/spotify.py
```diff
@@ -536,3 +556,71 @@ def _output_match_results(self, results):
...
+        import time
```
sampsyo, line 576: This can be a top-level `import` at the very top of the file.

beetsplug/spotify.py
```diff
@@ -536,3 +556,71 @@ def _output_match_results(self, results):
...
+
+        no_items = len(items)
+        self._log.info('Total {} tracks', no_items)
+
+        for index, item in enumerate(items, start=1):
+            time.sleep(.5)
+            self._log.info('Processing {}/{} tracks - {} ',
+                           index, no_items, item)
```
sampsyo, line 584: I think it's OK to use `len(items)` here; it's unlikely to cause measurable performance problems.

beetsplug/spotify.py
```diff
@@ -536,3 +556,71 @@ def _output_match_results(self, results):
...
+            'mode': ['spotify_track_mode'],
+            'speechiness': ['spotify_track_speechiness'],
+            'tempo': ['spotify_track_tempo'],
+            'time_signature': ['spotify_track_time_sig'],
+            'valence': ['spotify_track_valence'],
+        }
+        import time
+
+        no_items = len(items)
+        self._log.info('Total {} tracks', no_items)
```
sampsyo, line 579: Maybe this should be a `debug` message instead of `info`?

beetsplug/spotify.py
```diff
@@ -536,3 +556,71 @@ def _output_match_results(self, results):
...
+        track_popularity = track_data['popularity']
+        return track_popularity
+
+    def track_audio_features(self, track_id=None):
+        """Fetch track audio features by its Spotify ID."""
+        track_data = self._handle_response(
+            requests.get, self.audio_features_url + track_id
+        )
+        audio_features = track_data
+        return audio_features
```
sampsyo, line 626:
```suggestion
        return track_data
```

beetsplug/spotify.py
```diff
@@ -536,3 +556,71 @@ def _output_match_results(self, results):
...
+                pass
+
+    def track_popularity(self, track_id=None):
+        """Fetch a track popularity by its Spotify ID."""
+        track_data = self._handle_response(
+            requests.get, self.track_url + track_id
+        )
+        self._log.debug('track_data: {}', track_data['popularity'])
+        track_popularity = track_data['popularity']
+        return track_popularity
```
sampsyo, line 618:
```suggestion
        return track_data['popularity']
```

beetsplug/spotify.py
```diff
@@ -536,3 +556,71 @@ def _output_match_results(self, results):
...
+                for feature in audio_features.keys():
+                    if feature in spotify_audio_features.keys():
+                        item[spotify_audio_features[feature][0]] = \
+                            audio_features[feature]
+                item.store()
+                if write:
+                    item.try_write()
+            except AttributeError:
+                self._log.debug('No track_id present for: {}', item)
+                pass
```
sampsyo, line 609: I think his `try`/`except` handler should probably go _just_ around the `item.spotify_track_id` access above, to make it clear to readers that this is the only thing we expect to fail.

beetsplug/spotify.py
```diff
@@ -536,3 +556,71 @@ def _output_match_results(self, results):
...
+            time.sleep(.5)
+            self._log.info('Processing {}/{} tracks - {} ',
+                           index, no_items, item)
+            try:
+                # If we're not forcing re-downloading for all tracks, check
+                # whether the popularity data is already present
+                if not force:
+                    spotify_track_popularity = \
+                        item.get('spotify_track_popularity', '')
+                    if spotify_track_popularity:
```
sampsyo, line 591:
```suggestion
                    if 'spotify_track_popularity' in item:
```

sampsyo (changes requested):
Awesome; looking good so far! Here are a few suggestions/questions.

Do you also think you might be able to add to the docs for the plugin to describe the new command? A changelog entry would also be super helpful.

beetsplug/spotify.py
```diff
@@ -536,3 +556,71 @@ def _output_match_results(self, results):
...
+            'time_signature': ['spotify_track_time_sig'],
+            'valence': ['spotify_track_valence'],
+        }
+        import time
+
+        no_items = len(items)
+        self._log.info('Total {} tracks', no_items)
+
+        for index, item in enumerate(items, start=1):
+            time.sleep(.5)
```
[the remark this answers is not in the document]
arsaboo: To avoid [API rate limit](https://developer.spotify.com/documentation/web-api/guides/rate-limits/).

beetsplug/spotify.py
sampsyo: I guess one other option would be that we could just make _all_ of these `spotify_X` where `X` is the original Spotify name. That could be a little more predictable and would not require an explicit mapping.

arsaboo: Changed

arsaboo:
Thanks for the review @sampsyo. Addressed all the comments and updated the documents/changelog.

JOJ0:
Just here to say woohoo! Very nice feature! Thanks to the submitter 💣💥

arsaboo:
Thanks @JOJ0 

I have a few additional features in mind once this PR gets merged.

beetsplug/spotify.py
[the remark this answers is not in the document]
wisp3rwind: I think it would be great to have a code comment that explains this. Would it even be possible to check for a 429 response code in `SpotifyPlugin._handle_response`, and retry after the provided `Retry-After` value (and maybe default to some exponential backoff, the documentation doesn't seem to guarantee that this header is always pressent)?

beetsplug/spotify.py
wisp3rwind: This doesn't seem to be addressed in your revised PR, was that intentional? For which reason?

arsaboo: @wisp3rwind I am not sure exactly what is expected here. I could move [this block](https://github.com/beetbox/beets/blob/7e648db2583774e80834642b3a75ded7b3fa25ab/beetsplug/spotify.py#L588-L594)  outside the `try` block. Would that help?

beetsplug/spotify.py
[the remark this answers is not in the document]
arsaboo: @wisp3rwind I agree it would be great, but I am not sure about the implementation at this point. I have tried not to touch the `_handle_response` function for precisely that reason. Once this PR is merged, hopefully, we can add that feature to make the plugin more robust. With the `sleep` and if someone is using their own keys, we should not hit 429.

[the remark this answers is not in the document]
arsaboo: Meanwhile, I have added a comment to reflect that.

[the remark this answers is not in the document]
wisp3rwind: Fair enough, let's open a new issue to remember the issue of rate limiting.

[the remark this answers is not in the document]
arsaboo: Created an issue https://github.com/beetbox/beets/issues/4369 to keep a track of it.

beetsplug/spotify.py
wisp3rwind: I think what @sampsyo meant is this:

```python
            try:
                spotify_track_id = item.spotify_track_id
            except AttributeError:
                self._log.debug('No track_id present for: {}', item)
                continue
            
            popularity = self.track_popularity(spotify_track_id)
            item['spotify_track_popularity'] = popularity
            audio_features = \
                self.track_audio_features(spotify_track_id)
            for feature in audio_features.keys():
                if feature in self.spotify_audio_features.keys():
                    item[self.spotify_audio_features[feature]] = \
                        audio_features[feature]
            item.store()
            if write:
                item.try_write()
```

wisp3rwind:
Looks good to me! @sampsyo all your previous comments have been addressed, any other concerns? Or can we go ahead and merge this?

sampsyo (approved):
Nice!! This looks really great. Thank you for seeing this through!!

PR #4359, opened 2022-05-31
Make implicit path queries explicit and simplify their handling

### LINKED ISSUE #1865: Implicit path queries in a boolean OR yield a match-all query
This is best illustrated by example:
- `beet list /path/to/library/A` — works
- `beet list /path/to/library/B` — works
- `beet list /path/to/library/A , /path/to/library/B` — **lists the whole library**
- `beet list path:/path/to/library/A , path:/path/to/library/B` — works

snejus:
## Description

Fixes #1865.  

* [As per @intelfx suggestion](https://github.com/beetbox/beets/issues/1865#issuecomment-180027742),
  implicit path queries are replaced with explicit ones by prepending `path::` to them.

* @sampsyo thanks for adding a test for this issue in 19a5d8c - it's been unskipped.

## To Do

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -1394,7 +1394,7 @@ def parse_query_parts(parts, model_cls):
 
     # Special-case path-like queries, which are non-field queries
     # containing path separators (/).
-    parts = [f"path::{s}" if PathQuery.is_path_query(s) else s for s in parts]
+    parts = [f"path={s}" if PathQuery.is_path_query(s) else s for s in parts]
 
     case_insensitive = beets.config['sort_case_insensitive'].get(bool)
 
--- a/beets/library.py
+++ b/beets/library.py
@@ -1394,7 +1394,7 @@ def parse_query_parts(parts, model_cls):
 
     # Special-case path-like queries, which are non-field queries
     # containing path separators (/).
-    parts = [f"path={s}" if PathQuery.is_path_query(s) else s for s in parts]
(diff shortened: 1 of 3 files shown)


beets/library.py
```diff
@@ -1394,28 +1394,14 @@ def parse_query_parts(parts, model_cls):
...
     # Special-case path-like queries, which are non-field queries
     # containing path separators (/).
-    path_parts = []
-    non_path_parts = []
-    for s in parts:
-        if PathQuery.is_path_query(s):
-            path_parts.append(s)
-        else:
-            non_path_parts.append(s)
+    parts = [f"path:{s}" if PathQuery.is_path_query(s) else s for s in parts]
```
sampsyo, line 1397: Because this looks somewhat odd, just leaving a note to myself about why we're doing this string-manipulation thing instead of constructing the `PathQuery` directly. The problem is outlined here: https://github.com/beetbox/beets/issues/1865#issuecomment-180017864

sampsyo (approved):
Awesome!! Thank you for sorting this out. Everything looks good to me. I'd love a second pair of eyes to confirm nothing's out of order, and then we can merge it!

wisp3rwind:
> Awesome!! Thank you for sorting this out. Everything looks good to me. I'd love a second pair of eyes to confirm nothing's out of order, and then we can merge it!

LGTM, too!

Issue #3515, opened 2020-03-15
Track-level query returning list of albums containing those tracks


Galicarnax:
There are use cases where it is nice to be able to query beets DB on track level, but show results on album level. Currently, If I use `beet ls -a ...`, then I do see albums only, but the queries involving track fields are ignored. It would be great to see the list of albums in which at least one track matched a track-level query.

sampsyo:
Sounds good! One boring part of designing this feature would be the CLI… how should commands indicate that they want this query-by-track, print-by-album behavior? (Maybe it should be a form of query, actually, instead of a command-line flag, so it could work with *any* command instead of just `ls`?)

GuilhermeHideki:
Sometimes I use something like `beet ls -f '$album_id' trackLevelQuery | sort -u |  xargs <some flags> sh -c "beet ls -af 'some format' id:<xargs replacer>"`, but would like something more "user-friendly"

snejus:
Reopening, since #5240 is getting reverted #5326

PR #4362, opened 2022-06-05
Querying improvements / removing 'slow' queries

### LINKED ISSUE #4360: beet list -a path:some/path queries are extremely slow
### Problem

```sh
0s 943ms 734us
```

```sh
13s 951ms 836us
```

```
15s 398ms 109us
```

```
Tracks: 5697
Artists: 2634
Albums: 1304
Album artists: 901
```

### Setup

* Python version: 3.10.4

snejus:
## Description

Fixes #4360 #3515.  

This is a WIP pull request to act as a basis for further discussions regarding improvements to queries. See #4360 for more details. There may be more issues that have to do with this functionality that I am unaware of, so feel free to link any relevant discussions.

What's been done at this point:

- [x] Faster regex queries, `beet list album::album`
  - Added db function that does matching directly
- [x] Faster album path queries, `beet list -a path::some/path` 
  - Track-level metadata is queried directly through a join
- [x] Faster flexible attributes queries, both albums and tracks, `beet list play_count:10` 
  - The corresponding attribute table is also immediately accessible through a join
- [x] (Fast) Ability to list albums with track-level (and vice-versa) **db** field queries, `beet list -a title:something` 
  - `items` and `albums` tables are joined so that data is available for filtering
- [x] (Fast) Ability to list tracks with album-level **flexible** field queries, `beet list artpath:cover`
  - Since track data contains `album_id`, it's fairly straightforward to join `album_attributes` into track queries
- [x] Pass current tests
  - Except 3 `lslimit` plugin tests that have to do with the '<' query prefix. I had a look at its implementation and found that it's fairly fragile and complicated to adjust. Therefore, for now I added a flag `-l --limit` which accepts a value for the limit. See 734883fa. If it looks sensible I reckon it could be moved to another PR. 
  - `bareasc` and `playlist` plugins got slightly adjusted to enable fast db-level queries as well

## To Do
- [x] (Fast) Ability to list albums with track-level **flexible** field queries, `beet list -a title:something` 
  - It's kind of done but it caused a couple of tests to fail, so I left it out for now

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -15,20 +15,28 @@
 """The central Model and Database constructs for DBCore.
 """
 
+import json
 import time
 import os
 import re
+from functools import reduce
+from itertools import groupby
+from operator import add
 from collections import defaultdict
 import threading
 import sqlite3
 import contextlib
 
+from unidecode import unidecode
 import beets
 from beets.util import functemplate
 from beets.util import py3_path
 from beets.dbcore import types
 from .query import MatchQuery, NullSort, TrueQuery
 from collections.abc import Mapping
+
+# converter to load json strings produced by the 'group_to_json' aggregate
+sqlite3.register_converter("json_str", json.loads)
 
 
 class DBAccessError(Exception):
@@ -649,8 +657,7 @@ class Results:
     constructs LibModel objects that reflect database rows.
     """
 
-    def __init__(self, model_class, rows, db, flex_rows,
-                 query=None, sort=None):
+    def __init__(self, model_class, rows, db, query=None, sort=None):
         """Create a result set that will construct objects of type
         `model_class`.
 
@@ -670,7 +677,6 @@ class Results:
         self.db = db
         self.query = query
         self.sort = sort
-        self.flex_rows = flex_rows
 
         # We keep a queue of rows we haven't yet consumed for
         # materialization. We preserve the original total number of
@@ -693,9 +699,6 @@ class Results:
         first.
         """
 
-        # Index flexible attributes by the item ID, so we have easier access
-        flex_attrs = self._get_indexed_flex_attrs()
-
         index = 0  # Position in the materialized objects.
         while index < len(self._objects) or self._rows:
             # Are there previously-materialized objects to produce?
@@ -708,14 +711,11 @@ class Results:
             else:
                 while self._rows:
                     row = self._rows.pop(0)
-                    obj = self._make_model(row, flex_attrs.get(row['id'], {}))
-                    # If there is a slow-query predicate, ensurer that the
-                    # object passes it.
-                    if not self.query or self.query.match(obj):
-                        self._objects.append(obj)
-                        index += 1
-                        yield obj
-                        break
+                    obj = self._make_model(row)
+                    self._objects.append(obj)
+                    index += 1
+                    yield obj
+                    break
 
     def __iter__(self):
         """Construct and generate Model objects for all matching
@@ -730,28 +730,13 @@ class Results:
             # Objects are pre-sorted (i.e., by the database).
             return self._get_objects()
 
-    def _get_indexed_flex_attrs(self):
-        """ Index flexible attributes by the entity id they belong to
-        """
-        flex_values = {}
-        for row in self.flex_rows:
-            if row['entity_id'] not in flex_values:
-                flex_values[row['entity_id']] = {}
-
-            flex_values[row['entity_id']][row['key']] = row['value']
-
-        return flex_values
-
-    def _make_model(self, row, flex_values={}):
-        """ Create a Model object for the given row
-        """
-        cols = dict(row)
-        values = {k: v for (k, v) in cols.items()
-                  if not k[:4] == 'flex'}
+    def _make_model(self, row):
+        """ Create a Model object for the given row."""
+        values = dict(row)
+        flex_values = values.pop("flex_attrs", {})
 
         # Construct the Python object
-        obj = self.model_class._awaken(self.db, values, flex_values)
-        return obj
+        return self.model_class._awaken(self.db, values, flex_values)
 
     def __len__(self):
         """Get the number of matching objects.
@@ -954,9 +939,14 @@ class Database:
         # bytestring paths here on Python 3, so we need to
         # provide a `str` using `py3_path`.
         conn = sqlite3.connect(
-            py3_path(self.path), timeout=self.timeout
+            py3_path(self.path),
+            timeout=self.timeout,
+            # enable mapping types defined in column names as "col [type]"
+            # to sqlite converters
+            detect_types=sqlite3.PARSE_COLNAMES
         )
 
+        self.add_functions(conn)
         if self.supports_extensions:
             conn.enable_load_extension(True)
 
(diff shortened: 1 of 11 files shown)


beets/dbcore/db.py
```diff
@@ -1059,47 +1085,94 @@ def _make_attribute_table(self, flex_table):
...
-        )
-        )
+        table = model_cls._table
+
+        fields = {}
+        for fast, _fields in groupby(self._get_fields(query), lambda x: x[1]):
+            fields[fast] = {f[0] for f in _fields}
+
+        flex_fields = fields.get(False, set())
+        other_model_fields = fields.get(True, set()) - set(model_cls._fields)
```
wisp3rwind, line 1120: `groupby` expects its input to be sorted: https://docs.python.org/3/library/itertools.html#itertools.groupby. From what I can tell, this is not the case here.

One fix would be to drop the `groupby` entirely:
```python
fields = defaultdict(set)
for field, fast in self._get_fields(query):
    fields[fast].add(field)

flex_fields, model_fields = fields[False], fields[True]
other_model_fields = model_fields - set(model_cls._fields)
```

beets/dbcore/db.py
```diff
@@ -1059,47 +1085,94 @@ def _make_attribute_table(self, flex_table):
...
+
+        _from = table
+        select_columns = [f"{table}.*"]
+        if other_model_fields:
+            _from = "items JOIN albums ON items.album_id = albums.id"
```
wisp3rwind, line 1125: This breaks the `dbcore` abstraction and introduces specifics of the higher-level `beets.library` module. Maybe we can come up with a way to avoid this and store information on related tables in the `Model` subclasses, similar to how the table name is specified in `model_cls._table`?

(There's another reference to explicit table names in `_ids_predicate` below).

wisp3rwind (commented):
This looks like a pretty clever way to compound our queries!

I've only partially read the code now (and I may not be the best reviewer in the first place for any of the database code). Here's two initial comments:

beets/dbcore/db.py
snejus: Good point!

beets/dbcore/db.py
snejus: I agree, this would indeed be cleaner.

snejus:
@wisp3rwind Your comments have been addressed.

Furthermore, now albums can be filtered using track-level queries.

snejus:
Included a few simple tests as well. Except for the changelog entry, the functionality is complete at this point, so I'm marking the PR as ready for review.

beets/dbcore/db.py
```diff
@@ -1059,43 +1076,155 @@ def _make_attribute_table(self, flex_table):
...
 
         with self.transaction() as tx:
             rows = tx.query(sql, subvals)
-            flex_rows = tx.query(flex_sql, subvals)
 
         return Results(
-            model_cls, rows, self, flex_rows,
-            None if where else query,  # Slow query component.
+            model_cls, rows, self,
             sort if sort.is_slow() else None,  # Slow sort component.
```
snejus, line 1228: This should probably also go actually.

wisp3rwind:
Hi @snejus, looking great! I have essentially no time to review this (or anything for beets) at the moment, so it will take me some time to get around to it. It's on top of my beets todo list!

beets/dbcore/db.py
snejus: Attempted making sorting 'quick' but found that it is not at all straightforward and that it's already well optimized.

wisp3rwind:
For now, I've only looked at the `lslimit` changes. It's great to get rid of the brittle implementation in the `HeadQuery` and use SQL `LIMIT` instead, we should absolutely do this!

How exactly are the tests failing?

---

If we can't fix them easily by changes local to the `limit` plugin, I think we should still resolve this issue in one way or the other now:  Otherwise, we risk to "temporarily" disable the `<` query prefix, and never add it again. Two approaches could be

1) What you did here, but make it permanent instead of considering it to be a stopgap solution: That is, remove the `<` query prefix entirely, and add a `--limit` argument to relevant commands.
The downside is that we'd usually want to mark the prefix as deprecated for a while and issue a warning, before finally removing it. Maybe, we could at least turn the `HeadQuery` into what is essentially a `TrueQuery`, but displays a warning on first invocation? Doing this would still require appropriate communication in the changelog, etc., since we'd nevertheless have to remove the prefix on short notice. Would that be acceptable?

2) We come up with a way to keep the `<` prefix working, which will involve supporting it in query parsing in `dbcore.queryparse.parse_sorted_query`. This function would need to return something like `(query, sort, limit) ` instead of the current `(query, sort)`.
One question is whether to enable this by default, or still use the `limit` "plugin" to enable this prefix, in order to keep exact backwards compatibility.

@sampsyo (or anyone!) do you have any strong opinion here? This is only about commit 1 and 3 (734883fa9af0320b1f00bc97a286a3d3da8df803 and b9666ee9977d148ba6c467388414b21b9380d3d2) of this PR.

---

In any case, would it be possible for you (once we have decided what to do exactly) to split this into a separate preparatory PR, and rebase this PR on top of it? I think that would simplify review a bit.

snejus:
Hi @wisp3rwind, thanks for your response! See below for the failed tests:
```
FAILED test/test_limit.py::LimitPluginTest::test_prefix - AssertionError: 10 != 5
FAILED test/test_limit.py::LimitPluginTest::test_prefix_when_correctly_ordered - AssertionError: 10 != 5
FAILED test/test_limit.py::LimitPluginTest::test_prefix_when_incorrectly_ordred - AssertionError: 10 != 0
```
Let me know if you want to know the specifics of why it is the case. 

Regarding your suggestions,

1. This would probably not be ideal, though luckily this plugin has not been released yet - it got added a month later, after `1.6.0` got released, therefore a deprecation notice may not be needed. However, we also know that a fair bit of users install the `master` branch, so this change would be relevant to them.

2. I'd be very happy to go ahead with this solution. The implementation, as you suggested, is fairly straightforward. From the user's perspective, ability to limit the results using the same `list` command is great. Personally, my main use case is 
    * I find that a query yields too many results
    * Press up
    * Append `-l10`
    * Press enter

Some users may have already gotten used to be using the prefix, so supporting it going forward makes sense. 

Regarding the `lslimit` command, we could maybe print a notice like `Limiting is now supported in the list command, please use it instead` __before we release the next beets version__. We could remove this when it's released, since `limit` functionality doesn't exist in `1.6.0` anyways.

Let me know your thoughts. 

To confirm, I'll be happy to open a separate pull request for this.

arsaboo:
@snejus This looks useful for something that I using. If I am using something like `query = MatchQuery("plex_ratingkey", track.ratingKey, fast=False)` to match on the flexible attribute `plex_ratingkey`, will it work as is? Right now, the query is slow, which I am hoping your PR seeks to address.

snejus:
@arsaboo it should indeed work. 

You may as well try checking out this PR and attempt to, say, list tracks using your filter: `beet ls plex_ratingkey:xyz`. I'd be happy to hear whether there's any change in performance :)

wisp3rwind:
Alright, sorry for the long silence, but I'll have a little more time in the next weeks :)

---

> Let me know if you want to know the specifics of why it is the case.

Ah, ok, the entire `FieldQuery.match -> `*Query.value_match` code path is dead now, isn't it, since it all pushed down to SQL?

---

I had completely missed that `lslimit` has not been released; so let's just go ahead, remove it and add the functionality to beet's core :)

A deprecation notice on loading the plugin would indeed be nice. We already have (at least) two plugins that do so; see `gmusic.py` and `freedesktop.py`.

There's in fact a rationale for keeping the query prefix, see the docs https://github.com/beetbox/beets/pull/4190/files What's more, `LIMIT`ing properly would also solve the issue that's mentioned there about query ordering.

About the implementation, a few initial thoughts:
- `--limit/--head`: Easy enough, just factor out the limit keyword argument and implementation from this PR.
- `--tail`: Seems more complicated: This makes little sense unless a sort other than the default `NullSort` is used. Then, we would need to reverse the sort in the SQL, `LIMIT` the results, and then reverse the result set again. Reversing the SQL seems somewhat tricky: maybe it could be done by adding a `reverse` kwargs to `dbcore.query.*Sort.order_clause`.  Or, is there an easier option to formulate this directly in SQL?
@patrick-nicholson how important is the `--tail` command to your usecase? Maybe, it would be an option not to implement it for now to limit the complexity of the PR.
- the query prefix: I guess we would need a notion of "query modifiers" or "query options", such as a limit, that would be returned by `dbcore.queryparse.parse_sorted_query`.

wisp3rwind:
A semi-related issue is https://github.com/beetbox/beets/issues/750

patrick-nicholson:
@wisp3rwind I don't think `--tail` is all that necessary (at least on *nix) if sorting is correct, `--head` results can be reversed.

snejus:
Thanks for a reminder, will try reviving it in the coming days.

beets/dbcore/db.py
```diff
@@ -1059,43 +1076,155 @@ def _make_attribute_table(self, flex_table):
...
+
+        _from = self._relation_join(model)
+        if _from != model._table:
+            joins.append(join_tmpl.format(model._relation._flex_table,
+                                          f"{model._relation._table}.id"))
+
+        ids = set()
+        with self.transaction() as tx:
+            for join in joins:
+                sql = f"SELECT {id_field} FROM {_from} {join} WHERE {where}"
```
RollingStar, line 1154: Is code like this vulnerable to SQL injection and do we care?

RollingStar (commented):
For what it's worth, the code looks good. I work with databases but not on a level to thoroughly understand what's going on here.

RollingStar:
We should keep this active

snejus:
I've been intending to come back to this but have been too busy recently. This PR crosses my mind every weekend and I know I will be coming back to this soon :)

arsaboo:
We should keep this active....

snejus:
Superseded by #4746

PR #4363, opened 2022-06-05
Provide tutorial for setting up beets in Windows

abmeyers1:
## Description

Fixes #X.  

(...)

## To Do

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -84,6 +84,9 @@ You can install beets by typing ``pip install beets``.
 Beets has also been packaged in the `software repositories`_ of several
 distributions. Check out the `Getting Started`_ guide for more information.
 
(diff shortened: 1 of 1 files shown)


arsaboo:
Wondering if [this](https://beets.readthedocs.io/en/stable/guides/main.html#installing-on-windows) is a better place to link the same.

wisp3rwind:
> Wondering if [this](https://beets.readthedocs.io/en/stable/guides/main.html#installing-on-windows) is a better place to link the same.

In fact, since this is not really about installation but rather about first steps when installed, I think it would best fit into a new subsection along the lines of `Third-party tutorials` at the end of the `Getting started` page.

PR #4370, opened 2022-06-12
Add Spotify 429 (too many requests) API error handling

### LINKED ISSUE #4369: Add the ability to handle error 429 (too many requests) in Spotify plugin
Spotify has a rate-limiting feature. If there are too many requests to the Spotify App it will prevent you from sending further requests. Fortunately, the response includes the `retry-after` value that we should be using to back off. 

I am just creating this issue so that we can keep a track of it based on @wisp3rwind comment [here](https://github.com/beetbox/beets/pull/4352#discussion_r895174817). 

Here are some reference links to implement this feature:
https://github.com/skuethe/MMM-Spotify/issues/172
https://github.com/felix-hilden/tekore/issues/163

arsaboo:
## Description
Added ability to handle API error (too many requests). 

Fixes #4369 

## To Do

### DIFF
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -170,7 +170,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
                                                DEFAULT_WAITING_TIME)
                 self._log.debug('Too many API requests. Retrying after {} \
                     seconds.', seconds)
-                time.sleep(int(seconds))
+                time.sleep(int(seconds) + 1)
                 return self._handle_response(request_type, url, params=params)
             else:
                 raise ui.UserError(
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -167,8 +167,8 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
             elif response.status_code == 429:
                 seconds = response.headers['Retry-After']
                 time.sleep(int(seconds))
-                self._log.info('Too many API requests. Retrying after {} seconds.',
-                               seconds)
+                self._log.info('Too many API requests. Retrying after {} \
+                    seconds.', seconds)
                 return self._handle_response(request_type, url, params=params)
             else:
                 raise ui.UserError(
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -17,20 +17,19 @@
 Spotify playlist construction.
 """
 
-import re
-import json
 import base64
+import collections
+import json
+import re
 import time
 import webbrowser
-import collections
 
-import unidecode
-import requests
 import confuse
-
+import requests
+import unidecode
 from beets import ui
 from beets.autotag.hooks import AlbumInfo, TrackInfo
-from beets.plugins import MetadataSourcePlugin, BeetsPlugin
+from beets.plugins import BeetsPlugin, MetadataSourcePlugin
 
 DEFAULT_WAITING_TIME = 5
 
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -164,6 +164,11 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
                 )
                 self._authenticate()
                 return self._handle_response(request_type, url, params=params)
+            elif response.status_code == 429:
+                seconds = response.headers['Retry-After']
+                time.sleep(int(seconds))
+                self._log.info('Too many API requests. Retrying after {} seconds.', seconds)
+                return self._handle_response(request_type, url, params=params)
             else:
                 raise ui.UserError(
                     '{} API error:\n{}\nURL:\n{}\nparams:\n{}'.format(
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -32,6 +32,7 @@ from beets import ui
 from beets.autotag.hooks import AlbumInfo, TrackInfo
 from beets.plugins import MetadataSourcePlugin, BeetsPlugin
 
+DEFAULT_WAITING_TIME = 5
 
 class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
     data_source = 'Spotify'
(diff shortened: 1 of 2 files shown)


beetsplug/spotify.py
```diff
@@ -164,6 +164,12 @@ def _handle_response(self, request_type, url, params=None):
                 )
                 self._authenticate()
                 return self._handle_response(request_type, url, params=params)
+            elif response.status_code == 429:
+                seconds = response.headers['Retry-After']
```
wisp3rwind, line 168: Spotify does not seem to guarantee that this header is present; I think we should gracefully handle that case


```suggestion
                seconds = response.headers.get('Retry-After', DEFAULT_WAITING_TIME)
```

where I don't really know a good value for `DEFAULT_WAITING_TIME`. One of the links from your issue uses 5 seconds.

beetsplug/spotify.py
```diff
@@ -164,6 +164,12 @@ def _handle_response(self, request_type, url, params=None):
...
+                time.sleep(int(seconds))
+                self._log.info('Too many API requests. Retrying after {} \
```
wisp3rwind, line 170: Maybe swap these two (and reword the message accordingly), such that it's clear what's happening even if the delay becomes quite long for some reason.

sampsyo (commented):
This is definitely a step in the right direction! Handling the error and retrying seems much better than crashing when we exceed the rate-limit.

Can we remove the *other* `time.sleep` call here with this in place?
https://github.com/beetbox/beets/blob/c25e7ad51176f6ed5f6f7cec78ca4e86a4e146c2/beetsplug/spotify.py#L582

As I mentioned in https://github.com/beetbox/beets/issues/4369#issuecomment-1153299567, however, I think it's possible to do even better (perhaps in a subsequent PR?). The rough idea would be to explicitly account for [Spotify's 30-second rolling window](https://developer.spotify.com/documentation/web-api/guides/rate-limits/) on the client side by counting up the number of requests we've made recently and delaying only ones that exceed the limit. But like I said above, I this reactive-rather-than-proactive approach in this PR is a great 80% solution.

beetsplug/spotify.py
arsaboo: Set the `DEFAULT_WAITING_TIME` to 5 for now (based on the other issue).

beetsplug/spotify.py
arsaboo: Done!

arsaboo:
> Can we remove the _other_ `time.sleep` call here with this in place?

I think we can, but it may generate frequent 429 errors. We can reduce the `sleep` to 0.1 from 0.5 for now. I can test things out after this PR is merged and then maybe we can remove it entirely. Let me know what you think.

beetsplug/spotify.py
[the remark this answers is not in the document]
arsaboo: My thought was to keep the user informed about the delay. It will also be nice to get those reports if users see those messages so that we can adjust the timeout. In any case, I have changed it to `debug`.

sampsyo:
Looking better and better! But I guess I would like to ask a question about this:

> I think we can, but it may generate frequent 429 errors.

Have you observed that this happens? It might be a really useful experiment to run: what does the plugin behave like (how many errors does it see) if there is no extra `sleep`?

I ask for two reasons: first, intuitively, it seems like the 429 error's "please back off by N seconds" response could, in fact, work to avoid repeating the error too many times. That is, we'd inevitably get _one_ such error after doing many tracks, but then things would settle down after the error tells us how long to sleep for. Second, reading that aforelinked Spotify API documentation page made it sound like following the 429 errors was the _recommended_ way to do rate limiting. It doesn't even say what the rate limits are, which would be critical to avoiding 429 errors! So maybe this the technique we're supposed to use?

arsaboo:
I did receive 429 with the other PR (https://github.com/beetbox/beets/pull/4102#issuecomment-957495781). While Spotify does not explicitly specify the limits, I [read](https://jmeyers44.github.io/blog/2015/04/26/builder-beware-the-limitations-of-popular-apis/) that it should be able to handle 2000 calls per hour (so slightly less than 1 call per second). 

However, as you suggested, using the `retry-after` should be able to handle those cases. Thus, I have removed the sleep for now. We can add it back later if required.

sampsyo:
Great!! Just out of curiosity, have you gotten a chance to try it out on a large-ish number of files to see how well/poorly it performs?

arsaboo:
So far, I have not hit 429 (but that may be because I don't have my entire library on my test setup). Once we have this merged and hopefully released, I can test it out extensively. I have other features in mind for Spotify and Plex integration, so I will be sure to fix any issues that may come up :)

beetsplug/spotify.py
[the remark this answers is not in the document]
sampsyo: It looks like this changed back to `info`?

docs/changelog.rst
```diff
@@ -32,7 +32,9 @@ New features:
 
 Bug fixes:
 
-* Fix implicit paths OR queries (e.g. ``beet list /path/ , /other-path/``) 
+* Added Spotify 429 (too many requests) API error handling
+  :bug:`4370`
```
sampsyo, line 36: Maybe we should also say the bottom line for what the user can expect here: namely, we now respect the Spotify API's rate limiting, which avoids crashing when the API reports code 429 (too many requests).

sampsyo (commented):
Looking great!! Just one last suggestion within.

docs/changelog.rst
arsaboo: I changed it back to `info` so that the users are not guessing. See my argument [here](https://github.com/beetbox/beets/pull/4370#issuecomment-1157634268). If you prefer `debug`, I can make the change.

arsaboo: Changed logging it to `debug`.

sampsyo: Thanks!!! I do think this is the right thing. The philosophy is that now, with this new strategy, 429 responses are part of the _normal operation_ of the command—it's how we're enforcing the rate limit logic. So there's nothing actionable for the user to interpret when these errors occur.

sampsyo (approved):
Wahoo! Thank you for working on this!!

PR #4380, opened 2022-06-20
Define item_types for Spotify attributes to enable queries

arsaboo:
## Description

Right now the Spotify attributes do not have any type defined, which makes it impossible to use queries. This PR adds item_types for all the Spotify attributes 

## To Do


arsaboo:
Feel free to close the PR, if we prefer the `types` plugin instead.

sampsyo:
LGTM; thanks! I think it's nice for plugins to provide types for the fields they fill in. :smiley:

PR #4386, opened 2022-06-26
Handle audio features are not available Spotify API error 

### LINKED ISSUE #4385: Spotifysync crashes when tracks do not have audio-features
Some tracks in Spotify do not have audio-features. This causes `spotifysync` to crash:

```
spotify: Processing 48/471 tracks - Arijit Singh, Arko - OM - Rashtra Kavach - Seher
error: Spotify API error:
{
  "error" : {
    "status" : 404,
    "message" : "analysis not found"
  }
}
URL:
https://api.spotify.com/v1/audio-features/5pjSpt2mstf5JTf46FbT48
params:
None
```

`spotifysync` should gracefully handle such errors. Maybe we need a try-except block [here](https://github.com/beetbox/beets/blob/master/beetsplug/spotify.py#L622). Thoughts?

arsaboo:
## Description

Fixes #4385 

## To Do

### DIFF
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -193,7 +193,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
                     seconds.', seconds)
                 time.sleep(int(seconds) + 1)
                 return self._handle_response(request_type, url, params=params)
-            elif 'analysis not found' in response.text:
+            elif response.status_code == 400:
                 raise SpotifyAPIError("API Error {0.status_code} for {1}"
                                       .format(response, url))
             else:
@@ -650,9 +650,8 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
     def track_audio_features(self, track_id=None):
         """Fetch track audio features by its Spotify ID."""
         try:
-            track_data = self._handle_response(
+            return self._handle_response(
                 requests.get, self.audio_features_url + track_id)
         except SpotifyAPIError as e:
             self._log.debug('Spotify API error: {}', e)
-            track_data = None
-        return track_data
+            return None
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -622,7 +622,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
             audio_features = \
                 self.track_audio_features(spotify_track_id)
             if audio_features is None:
-                self._log.debug('No audio features found for: {}', item)
+                self._log.info('No audio features found for: {}', item)
                 continue
             for feature in audio_features.keys():
                 if feature in self.spotify_audio_features.keys():
@@ -645,8 +645,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
         try:
             track_data = self._handle_response(
                 requests.get, self.audio_features_url + track_id)
-            track_data.raise_for_status()
-        except requests.exceptions.RequestException as e:
+        except AttributeError:
             self._log.debug('Audio feature update failed: {0}', str(e))
             track_data = None
         return track_data
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -193,7 +193,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
                     seconds.', seconds)
                 time.sleep(int(seconds) + 1)
                 return self._handle_response(request_type, url, params=params)
-            elif response.status_code == 400:
+            elif response.status_code == 404:
                 raise SpotifyAPIError("API Error {0.status_code} for {1}"
                                       .format(response, url))
             else:
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -189,9 +189,6 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
                     seconds.', seconds)
                 time.sleep(int(seconds) + 1)
                 return self._handle_response(request_type, url, params=params)
-            elif 'analysis not found' in response.text:
-                self._log.debug('No audio analysis found')
-                return None
             else:
                 raise ui.UserError(
                     '{} API error:\n{}\nURL:\n{}\nparams:\n{}'.format(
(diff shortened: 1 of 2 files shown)


beetsplug/spotify.py
```diff
@@ -189,6 +189,9 @@ def _handle_response(self, request_type, url, params=None):
                     seconds.', seconds)
                 time.sleep(int(seconds) + 1)
                 return self._handle_response(request_type, url, params=params)
+            elif 'analysis not found' in response.text:
```
sampsyo, line 192: Would it be possible to check the status code instead of looking at the text? For example, `response.status_code == 404`.

beetsplug/spotify.py
```diff
@@ -621,6 +624,9 @@ def _fetch_info(self, items, write, force):
             item['spotify_track_popularity'] = popularity
             audio_features = \
                 self.track_audio_features(spotify_track_id)
+            if audio_features is None:
+                self._log.debug('No audio features found for : {}', item)
```
sampsyo, line 628: It seems like this one could be an `info`-level log—the user probably wants to know if the data is actually missing for a given song.

sampsyo, line 628:
```suggestion
                self._log.debug('No audio features found for: {}', item)
```

docs/changelog.rst
```diff
@@ -32,6 +32,8 @@ New features:
 
 Bug fixes:
 
+* Fix Spotify API error when tracks do not have audio-features.
+  :bug:`4385`
```
sampsyo, line 36: Probably no need for an extra bugfix entry in the changelog, since we already have an entry for the new `spotifysync` command itself.

sampsyo (changes requested):
Awesome; looking good! Thanks for the quick work!! Here are a few low-level suggestions.

beetsplug/spotify.py
arsaboo: I'm wondering if there may be other 404, which we have not encountered yet. As I test this more extensively over time, we can change this to 404 if required. Right now, my focus was to capture all the errors that we can encounter. Hope that makes sense. Let me know if you still prefer to change it to 404. I'll fix the other issues soon.

sampsyo: Ah, I think I see now, and I was a bit too hasty. Namely, this is in `_handle_response`, which is used for all sorts of API calls, not just for querying the audio analysis. If it were the latter, then I think it would be safe to assume "404" meant "no audio analysis found."

With that in mind, it's a teensy bit troublesome that there is endpoint-specific logic in this general API wrapper function. It now means that, _technically_, the `_handle_response` function _could_ return `None` for any of its calls in this plugin. I wonder if we shouldn't instead just do `response.raise_for_status()` and handle the exception in the audio-features-specific code?

docs/changelog.rst
arsaboo: Done

beetsplug/spotify.py
arsaboo: Done

beetsplug/spotify.py
arsaboo: Moved it to `track_audio_features` function. Can you check there...thanks!

beetsplug/spotify.py
```diff
@@ -639,7 +642,11 @@ def track_popularity(self, track_id=None):
 
     def track_audio_features(self, track_id=None):
         """Fetch track audio features by its Spotify ID."""
-        track_data = self._handle_response(
-            requests.get, self.audio_features_url + track_id
-        )
+        try:
+            track_data = self._handle_response(
+                requests.get, self.audio_features_url + track_id)
+            track_data.raise_for_status()
```
sampsyo, line 648: I don't think this will work, unfortunately… `_handle_response` returns `response.json()`, so the value will be a `dict` instead of a `Response` object.

sampsyo (commented):
Hello! Thanks for continuing to work on this! I think the current setup will be an `AttributeError` because `_handle_response` does not return a `Response` object. Is there a way to try out the new code by deliberately running the plugin on a song with missing data?

arsaboo:
> Is there a way to try out the new code by deliberately running the plugin on a song with missing data?

I have tried installing the fix using `python3 -m pip install git+https://github.com/arsaboo/beets.git@spotify_error` and it has worked so far. It is possible that Spotify added the missing attributes (they may have been missing while the song was being processed). In any case, I will keep an eye on this or any other error that may appear.

arsaboo:
@sampsyo all good so far....have been testing this with over 5000 tracks. 

We can merge this for now if you have no other concerns.

arsaboo:
@sampsyo Updated the error handling so that we can address other errors in the future. I followed the [beatport](https://github.com/beetbox/beets/blob/master/beetsplug/beatport.py) approach. The current approach works perfectly. Here's the log (in debug mode)
```
spotify: Processing 547/5849 tracks - Ankit Tiwari, Manoj Muntashir - Ek Villain Returns - Galliyan Returns (From "Ek Villain Returns")
spotify: track_data: 0
spotify: Spotify API error: API Error 404 for https://api.spotify.com/v1/audio-features/2eTkxblLeqdTSsb2ivafDX
spotify: No audio features found for: Ankit Tiwari, Manoj Muntashir - Ek Villain Returns - Galliyan Returns (From "Ek Villain Returns")
spotify: Processing 548/5849 tracks - Ankit Tiwari, Nikhita Gandhi - Badhaai Do - Bandi Tot
```

If the user is not running in debug mode, the user will just see a message and the next track will be processed. 
```
spotify: No audio features found for: Ankit Tiwari, Manoj Muntashir - Ek Villain Returns - Galliyan Returns (From "Ek Villain Returns")
```

beetsplug/spotify.py
```diff
@@ -639,7 +649,10 @@ def track_popularity(self, track_id=None):
 
     def track_audio_features(self, track_id=None):
         """Fetch track audio features by its Spotify ID."""
-        track_data = self._handle_response(
-            requests.get, self.audio_features_url + track_id
-        )
+        try:
+            track_data = self._handle_response(
+                requests.get, self.audio_features_url + track_id)
```
sampsyo, line 654: This could be simplified to `return self._handle_response(…)`.

beetsplug/spotify.py
```diff
@@ -639,7 +649,10 @@ def track_popularity(self, track_id=None):
...
+        except SpotifyAPIError as e:
+            self._log.debug('Spotify API error: {}', e)
+            track_data = None
```
sampsyo, line 657: And this could be `return None`.

beetsplug/spotify.py
sampsyo: Looking like this is going in the right direction! But since this is raising a generic `SpotifyAPIError`, it seems like this condition should now just be `response.status_code == 404` instead of looking for this particular substring in the response? That way, different uses of `_handle_response` can still handle 404s in different ways.

sampsyo (commented):
Looking great; thanks for continuing to push this forward! I just have a few comments within.

arsaboo:
> Looking great; thanks for continuing to push this forward! I just have a few comments within.

Thanks....fixed all the issues.

sampsyo (commented):
Looking good! Just want to check that the status code here is right…

sampsyo (approved):
LGTM; thanks for your work on this!

PR #4387, opened 2022-06-26
feat(import): Add support for reading skipped paths from logfile

### LINKED ISSUE #4379: Allow importing from logfile
When importing large collections, it makes sense to do the automated import first, then do the manual part afterwards:

    $ beet import -l skipped.log -q /path/to/library

It would be great if it was possible to continue the import using the logfile instead of a directory:

    $ beet import skipped.log

Beets should be able to parse the logfile and call import on each directory that was skipped.

Holzhaus:
## Description

Fixes #4379.

## To Do

- [X] Test. (Encouraged but not strictly required.)

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -933,31 +933,6 @@ def import_files(lib, paths, query):
     """Import the files in the given list of paths or matching the
     query.
     """
-    # Check the user-specified directories.
-    paths_to_import = []
-    for path in paths:
-        normalized_path = syspath(normpath(path))
-        if not os.path.exists(normalized_path):
-            raise ui.UserError('no such file or directory: {}'.format(
-                displayable_path(path)))
-
-        # Read additional paths from logfile.
-        if os.path.isfile(normalized_path):
-            try:
-                paths_to_import.extend(list(
-                    _paths_from_logfile(normalized_path)))
-            except ValueError:
-                # We could raise an error here, but it's possible that the user
-                # tried to import a media file instead of a logfile. In that
-                # case, logging a warning would be confusing, so we skip this
-                # here.
-                pass
-            else:
-                # Don't re-add the log file to the list of paths to import.
-                continue
-
-        paths_to_import.append(paths)
-
     # Check parameter consistency.
     if config['import']['quiet'] and config['import']['timid']:
         raise ui.UserError("can't be both quiet and timid")
@@ -978,11 +953,11 @@ def import_files(lib, paths, query):
             config['import']['quiet']:
         config['import']['resume'] = False
 
-    session = TerminalImportSession(lib, loghandler, paths_to_import, query)
+    session = TerminalImportSession(lib, loghandler, paths, query)
     session.run()
 
     # Emit event.
-    plugins.send('import', lib=lib, paths=paths_to_import)
+    plugins.send('import', lib=lib, paths=paths)
 
 
 def import_func(lib, opts, args):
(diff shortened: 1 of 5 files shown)


beets/ui/commands.py
```diff
@@ -79,6 +79,26 @@ def _do_query(lib, query, album, also_items=True):
...
 
+def _paths_from_logfile(path):
+    """Parse the logfile and yield skipped paths to pass to the `import`
+    command.
+    """
+    with open(path, mode="r", encoding="utf-8") as fp:
+        for line in fp:
+            verb, sep, paths = line.rstrip("\n").partition(" ")
+            if not sep:
+                raise ValueError("malformed line in logfile")
```
sampsyo, line 90: I'm not sure `ValueError` is the right thing here. I believe we want to surface this kind of error to the user (assuming we change the interface to be explicit about the file type), so maybe `ui.UserError` is more appropriate—or a new subclass thereof?

sampsyo (commented):
Super cool; thank you for looking into this!

If it doesn't seem overly complex to implement, I think I'd prefer to use an explicit command-line flag (like `beet import --read-log foo.log`) as opposed to implicitly allowing logs as arguments alongside files/directories. The main benefit would be that it will be more "debuggable" from the perspective of the user: if the log file isn't want they expect or is malformed, they can get a useful error message instead of letting the text file be treated as a media file. I hope this makes sense!

beets/ui/commands.py
Holzhaus: I thought that at this point the error is "internal". `os.path.commonpath` might raise a `ValueError` too. The error is caught in the calling code and could be used to raise a `ui.UserError`. I didn't do that yet because without an additional flag we can't  distinguish logfiles from audio files yet.

Holzhaus: error handling is here: https://github.com/beetbox/beets/pull/4387/commits/99ced5cc7def450181cc62fb7190be5ef4697eba#diff-e324b20657a7d6b43b8b7aeb5754b96774f5062294b5ba7f1e3062845e9e7044R982-R991

beets/ui/commands.py
```diff
@@ -960,7 +974,25 @@ def import_func(lib, opts, args):
...
+        paths_from_logfiles = []
+        for logfile in opts.from_logfiles or []:
+            try:
+                paths_from_logfiles.extend(
+                    list(_paths_from_logfile(syspath(normpath(logfile))))
+                )
+            except ValueError as err:
+                raise ui.UserError('malformed logfile {}'.format(
+                    util.displayable_path(logfile),
+                )) from err
```
sampsyo, line 989: For this one, at least, perhaps it would be nice to include some evidence about why the log file is malformed? Because `UserError` just prints the associated string by default (and doesn't display the "chained" exception), just including `str(err)` in the message could make a difference in usability.

beets/ui/commands.py
```diff
@@ -960,7 +974,25 @@ def import_func(lib, opts, args):
     else:
         query = None
         paths = args
-        if not paths:
+
+        # The paths from the logfiles go into a separate list to allow handling
+        # errors differently from user-specified paths.
```
sampsyo, line 979: This seems good! Just for the sake of organization and keeping the function short, what do you think about factoring this stanza out into its own helper function?

sampsyo (commented):
Awesome; thanks for taking this "flag" approach! I think this is basically there—I have one or two very low-level comments for your consideration.

beets/ui/commands.py
Holzhaus: https://github.com/beetbox/beets/pull/4387/commits/cfc34826a20203240e1bc8f2ba0cee328dfe18df

beets/ui/commands.py
Holzhaus: https://github.com/beetbox/beets/pull/4387/commits/cfc34826a20203240e1bc8f2ba0cee328dfe18df

sampsyo (approved):
Looking great. Thanks for the fast and clean work.

Issue #4392, opened 2022-06-29
importadded plugin don't seem to work when importing from a zip file


zuzzurro:
Importing a disc from a zip file while having the importadded set results in file times not corresponding to the content of the zip archive itself, the files times are current. Repeating the same import on a folder extracted from the zip file using the unzip tool works fine.

I'm using 1.6.0 on Fedora 36 (self compiled) as the distro still has 1.4.9.

I can provide more information, but the issue is quite easy to reproduce as I explained above.

sampsyo:
Interesting! I think the first step to understanding this would be to know whether the library we use to extract these files, namely [the `zipfile.extractall` function](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.extractall), can preserve mtimes. If it does, then we would need to find where we're discarding those. If it doesn't, then we may be out of luck. Any chance you'd be able to investigate?

zuzzurro:
It looks like if people want to extract the correct metadata it needs to be done manually afterwards. Maybe similar to this?

[https://www.burgundywall.com/post/preserving-file-perms-with-python-zipfile-module](url)

arogl:
I found the following, so it seems a limitation of pythons implementation of zip extraction:

https://stackoverflow.com/questions/9813243/extract-files-from-zip-file-and-retain-mod-date

zuzzurro:
Would it be possible to adopt one of the proposed solutions? Since it's for internal use I don't see much possibly going wrong by changing the time by hand after the extraction.

sampsyo:
Nice find, @arogl! It seems technically possible, but somewhat annoying to implement because we can no longer just use that `extractall` function… but I'll mark this as a feature request in case anyone is interested in giving it a shot.

zuzzurro:
I haven't tested if the functionality works for RAR files, but since I think the corresponding functionality is implemented by directly calling an external utility (unrar) maybe that's the case.

arogl:
@sampsyo 

Could this work?

In [importer.py#L1080](https://github.com/beetbox/beets/blob/master/beets/importer.py#L1080) add

``` py

# From here:
# https://stackoverflow.com/questions/9813243/extract-files-from-zip-file-and-retain-mod-date
# fixing #4392

def RestoreTimestampsOfArchiveContents(archivename, extract_dir):
    for f in archivename.infolist():
        # path to this extracted f-item
        fullpath = os.path.join(extract_dir, f.filename)
[...]

PR #4396, opened 2022-07-02
Preserve mtimes from archives #4392

arogl:
## Description

Fixes #X.  

(...)

## To Do


arogl:
Initial change

Still to do
 [ ] test which archive type 
 [ ] how to check whether preserve_mtimes is set

arogl:
> Is this still relevant? If so, what is blocking it? Is there anything you can do to help move it forward?
> 
> This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

Still relevant.

arogl:
This should be an easy pull, as it mimics normal unzip behaviour of extracting files with times from the zip file.

Python unzip does not modify the files date when extracting, so it gets the current time.

beets/importer.py
```diff
@@ -1090,6 +1090,19 @@ def extract(self):
         archive = handler_class(util.py3_path(self.toppath), mode='r')
         try:
             archive.extractall(extract_to)
+
+            # From here:
+            # https://stackoverflow.com/questions/9813243/extract-files-from-zip-file-and-retain-mod-date
+            # fixing #4392
+
+            for f in archive.infolist():
+                # path to this extracted f-item
```
sampsyo, line 1099: This comment doesn't seem important.

beets/importer.py
```diff
@@ -1090,6 +1090,19 @@ def extract(self):
...
+                fullpath = os.path.join(extract_to, f.filename)
+                # still need to adjust the dt o/w item will have the current dt
+                date_time = time.mktime(f.date_time + (0, 0, -1))
+                # update date_time
+                os.utime(fullpath, (date_time, date_time))
```
sampsyo, line 1104: Maybe we should just put the relevant assignment closer to here?
```suggestion
                 fullpath = os.path.join(extract_to, f.filename)
                os.utime(fullpath, (date_time, date_time))
```

beets/importer.py
```diff
@@ -1090,6 +1090,19 @@ def extract(self):
...
+
+            # From here:
+            # https://stackoverflow.com/questions/9813243/extract-files-from-zip-file-and-retain-mod-date
+            # fixing #4392
+
+            for f in archive.infolist():
+                # path to this extracted f-item
+                fullpath = os.path.join(extract_to, f.filename)
+                # still need to adjust the dt o/w item will have the current dt
+                date_time = time.mktime(f.date_time + (0, 0, -1))
```
sampsyo, line 1102: I don't yet understand the need for `+ (0, 0, -1)` here… in fact, I'm not entirely sure what type `f.date_time` is. Maybe this needs a closer look?

beets/importer.py
```diff
@@ -1090,6 +1090,19 @@ def extract(self):
         archive = handler_class(util.py3_path(self.toppath), mode='r')
         try:
             archive.extractall(extract_to)
+
+            # From here:
+            # https://stackoverflow.com/questions/9813243/extract-files-from-zip-file-and-retain-mod-date
+            # fixing #4392
+
```
sampsyo, line 1097:
```suggestion
            # Adjust the files' mtimes to match the information from the archive. Inspired by:
            # https://stackoverflow.com/q/9813243
```

sampsyo (commented):
If it works, it works!! Here are a few suggestions, especially about the comments.

beets/importer.py
arogl: Comment removed

beets/importer.py
arogl: I have updated the code

beets/importer.py
arogl: I added the question and answer from the stackoverflow link

beets/importer.py
arogl: Changed as requested

sampsyo:
Looking good! I think all this needs is to polish the comments so they're self-contained and oriented toward the code-reader of the future rather than summarizing our discussion here, and it will be good to go.

sampsyo:
Very nice work; thanks for your effort on this!!

PR #4397, opened 2022-07-02
docs: fix simple typo, seperated -> separated

timgates42:
There is a small typo in beetsplug/web/static/jquery.js.

Should read `separated` rather than `seperated`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


sampsyo:
Thank you for the contribution, but I don't think we want to change the copy of jQuery we have vendored here.

PR #4399, opened 2022-07-04
convert: New feature "Write m3u playlist to destination folder"

JOJ0:
## Description

- Enables the functionality proposed in https://github.com/beetbox/beets/discussions/4373
- The convert plugin is suitable to be used as a media file exporter already.
- Thus it makes sense to add a feature to save a playlist file into the export folder containing all exported media files in the order they were exported/converted.
- Since with the convert plugin all sorts of "beets queries" could be exported/converted, the feature uses a more general approach than described in the initial proposal (above linked discussion). To achieve exactly what's described in the discussion, additionally the beets `playlist` feature can be used.
- Relative paths should be used to make the playlist file portable, e.g used on another computer.
- The classic m3u format does not do well with special characters in media file names. Since the beets "path" variables could be used to write all sorts of characters, the m3u8 format which requires unicode for all media file entries, is required.

## To Do

- [x] Tests.
  - [x] convert plugin tests.
  - [x] M3UFile class tests (macOS/Unix/General).
  - [x] M3UFile class tests (Windows).
- [x] Currently there is an issue with VLC media player not reading the generated files correctly. It seems to still have a problem with special characters (e.g square brackets: []). -> Fixed, it was a bug in VLC 3.0.16, it works with 3.0.17.3, see https://forum.videolan.org/viewtopic.php?f=2&t=158920&p=529284#p529284
- [x] It seems that there is another bug (?) in the current VLC version 3.0.17.3 that files containing the # character are not opening, also see same post: https://forum.videolan.org/viewtopic.php?f=2&t=158920&p=529284#p529284
        I'm not considering this a problem with my implementation here. Let's see what happens on that forum post, but for now I check this task as done.
- [x] Test correct reading of the files in other software. So far working fine (all tested on macOS 10.15.3 Catalina):
    - [x] iTunes (latest on Catalina)
    - [x] Traktor Scratch Pro 2 (latest, 2.11.3.17)
    - [x] VLC (3.0.17.3)
- [x] ~Ensure cross-platform compatibility~Ensure compatibility of playlists generated _on_ an OS _for_ an OS. For example, lists generated on Windows should work on Windows.

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -494,7 +494,7 @@ class ConvertPlugin(BeetsPlugin):
         if playlist:
             self._log.info("Creating playlist file: {0}", playlist)
             if not pretend:
-                with open(os.path.join(dest, playlist), "w") as playlist_file:
+                with open(playlist, "w") as playlist_file:
                     playlist_file.write("#EXTM3U" + "\n")
 
         self._parallel_convert(dest, opts.keep_new, path_formats, fmt, pretend,
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -149,6 +149,7 @@ class ConvertPlugin(BeetsPlugin):
             'copy_album_art': False,
             'album_art_maxwidth': 0,
             'delete_originals': False,
+            'playlist': '',
         })
         self.early_import_stages = [self.auto_convert, self.auto_convert_keep]
 
@@ -177,6 +178,8 @@ class ConvertPlugin(BeetsPlugin):
                               dest='hardlink',
                               help='hardlink files that do not \
                               need transcoding. Overrides --link.')
+        cmd.parser.add_option('-m', '--playlist', action='store',
+                              help='set the name of the playlist file to be created')
         cmd.parser.add_album_option()
         cmd.func = self.convert_func
         return [cmd]
@@ -257,7 +260,7 @@ class ConvertPlugin(BeetsPlugin):
                            util.displayable_path(source))
 
     def convert_item(self, dest_dir, keep_new, path_formats, fmt,
-                     pretend=False, link=False, hardlink=False):
+                     pretend=False, link=False, hardlink=False, playlist=''):
         """A pipeline thread that converts `Item` objects from a
         library.
         """
@@ -281,6 +284,13 @@ class ConvertPlugin(BeetsPlugin):
                 if should_transcode(item, fmt):
                     dest = replace_ext(dest, ext)
                 converted = dest
+
+            # Quick, dirty, playlist
+            # converted_filename = Path(str(converted)).name
+            converted_filename = os.path.basename(converted)
+            self._log.info("Appending to playlist file {0}".format(playlist))
+            with open(playlist, "ab") as playlist_file:
+                playlist_file.write(converted_filename + b"\n")
 
             # Ensure that only one thread tries to create directories at a
             # time. (The existence check is not atomic with the directory
@@ -436,7 +446,7 @@ class ConvertPlugin(BeetsPlugin):
 
     def convert_func(self, lib, opts, args):
         (dest, threads, path_formats, fmt,
-         pretend, hardlink, link) = self._get_opts_and_config(opts)
+         pretend, hardlink, link, playlist) = self._get_opts_and_config(opts)
 
         if opts.album:
             albums = lib.albums(ui.decargs(args))
@@ -461,8 +471,16 @@ class ConvertPlugin(BeetsPlugin):
                 self.copy_album_art(album, dest, path_formats, pretend,
                                     link, hardlink)
 
+        if playlist:
+            print("################")
+            print(f"Playlist: {playlist}")
+            print("################")
+            if not pretend:
+                with open(playlist, "w") as playlist_file:
+                    playlist_file.write("#EXTM3U" + "\n")
+
         self._parallel_convert(dest, opts.keep_new, path_formats, fmt,
-                               pretend, link, hardlink, threads, items)
+                               pretend, link, hardlink, threads, items, playlist)
 
     def convert_on_import(self, lib, item):
         """Transcode a file automatically after it is imported into the
@@ -544,6 +562,8 @@ class ConvertPlugin(BeetsPlugin):
 
         fmt = opts.format or self.config['format'].as_str().lower()
 
+        playlist = opts.playlist or self.config['playlist'].get()
+
         if opts.pretend is not None:
             pretend = opts.pretend
         else:
@@ -559,10 +579,11 @@ class ConvertPlugin(BeetsPlugin):
             hardlink = self.config['hardlink'].get(bool)
             link = self.config['link'].get(bool)
 
-        return dest, threads, path_formats, fmt, pretend, hardlink, link
+
+        return dest, threads, path_formats, fmt, pretend, hardlink, link, playlist
 
     def _parallel_convert(self, dest, keep_new, path_formats, fmt,
-                          pretend, link, hardlink, threads, items):
+                          pretend, link, hardlink, threads, items, playlist):
         """Run the convert_item function for every items on as many thread as
         defined in threads
         """
@@ -572,7 +593,8 @@ class ConvertPlugin(BeetsPlugin):
                                      fmt,
                                      pretend,
                                      link,
-                                     hardlink)
+                                     hardlink,
+                                     playlist)
                    for _ in range(threads)]
         pipe = util.pipeline.Pipeline([iter(items), convert])
         pipe.run_parallel()
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -299,11 +299,16 @@ class ConvertPlugin(BeetsPlugin):
                     "Appending to playlist file {0}",
                     util.displayable_path(playlist)
                 )
+                # The classic m3u format doesn't support special characters in
+                # media file paths, thus we use the m3u8 format which requires
+                # media file paths to be unicode. Additionally we use relative
+                # paths to ensure readability of the playlist on remote
+                # computers.
+                dest_relative = util.displayable_path(dest).replace(
+                    util.displayable_path(dest_dir) + os.sep, ""
+                )
                 with open(playlist, "a") as playlist_file:
-                    # The classic m3u format doesn't support special characters
-                    # in media file paths, thus we use the m3u8 format which
-                    # requires media file paths to be unicode.
-                    playlist_file.write(util.displayable_path(dest) + "\n")
+                    playlist_file.write(dest_relative + "\n")
 
             # Ensure that only one thread tries to create directories at a
             # time. (The existence check is not atomic with the directory
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -31,7 +31,7 @@ from beets import art
 from beets.util.artresizer import ArtResizer
 from beets.library import parse_query_string
 from beets.library import Item
-from beets.util import M3UFile
+from beets.util.m3u import M3UFile
 
 _fs_lock = threading.Lock()
 _temp_files = []  # Keep track of temporary transcoded files for deletion.
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -304,8 +304,10 @@ class ConvertPlugin(BeetsPlugin):
                 # media file paths to be unicode. Additionally we use relative
                 # paths to ensure readability of the playlist on remote
                 # computers.
-                dest_relative = util.displayable_path(dest).replace(
-                    util.displayable_path(dest_dir) + os.sep, ""
+                dest_relative = item.destination(
+                    basedir=dest_dir,
+                    path_formats=path_formats,
+                    fragment=True
                 )
                 with open(playlist, "a") as playlist_file:
                     playlist_file.write(dest_relative + "\n")
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -476,16 +476,13 @@ class ConvertPlugin(BeetsPlugin):
                                link, hardlink, threads, items)
 
         if playlist:
-            # When playlist arg is passed create an m3u8 file in dest folder.
-            #
-            # The classic m3u format doesn't support special characters in
-            # media file paths, thus we use the m3u8 format which requires
-            # media file paths to be unicode. Additionally we use relative
-            # paths to ensure readability of the playlist on remote
-            # computers.
+            # Playlist paths are understood as relative to the dest directory.
             pl_normpath = util.normpath(playlist)
             pl_dir = os.path.dirname(pl_normpath)
             self._log.info("Creating playlist file {0}", pl_normpath)
+            # Generates a list of paths to media files, ensures the paths are
+            # relative to the playlist's location and translates the unicode
+            # strings we get from item.destination to bytes.
             items_paths = [
                 os.path.relpath(util.bytestring_path(item.destination(
                     basedir=dest, path_formats=path_formats, fragment=False
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -267,7 +267,7 @@ class ConvertPlugin(BeetsPlugin):
                            util.displayable_path(source))
 
     def convert_item(self, dest_dir, keep_new, path_formats, fmt,
-                     pretend=False, link=False, hardlink=False, playlist=None):
+                     pretend=False, link=False, hardlink=False):
         """A pipeline thread that converts `Item` objects from a
         library.
         """
@@ -291,26 +291,6 @@ class ConvertPlugin(BeetsPlugin):
                 if should_transcode(item, fmt):
                     dest = replace_ext(dest, ext)
                 converted = dest
-
-            # When the playlist argument is passed, add the current filename to
-            # an m3u8 playlist file located in the destination folder.
-            if playlist:
-                self._log.debug(
-                    "Appending to playlist file {0}",
-                    util.displayable_path(playlist)
-                )
-                # The classic m3u format doesn't support special characters in
-                # media file paths, thus we use the m3u8 format which requires
-                # media file paths to be unicode. Additionally we use relative
-                # paths to ensure readability of the playlist on remote
-                # computers.
-                dest_relative = item.destination(
-                    basedir=dest_dir,
-                    path_formats=path_formats,
-                    fragment=True
-                )
-                with open(playlist, "a") as playlist_file:
-                    playlist_file.write(dest_relative + "\n")
 
             # Ensure that only one thread tries to create directories at a
             # time. (The existence check is not atomic with the directory
@@ -492,13 +472,26 @@ class ConvertPlugin(BeetsPlugin):
                                     link, hardlink)
 
         if playlist:
+            # When playlist arg is passed create an m3u8 file in dest folder.
+            #
+            # The classic m3u format doesn't support special characters in
+            # media file paths, thus we use the m3u8 format which requires
+            # media file paths to be unicode. Additionally we use relative
+            # paths to ensure readability of the playlist on remote
+            # computers.
             self._log.info("Creating playlist file: {0}", playlist)
+            items_paths = [
+                item.destination(
+                    basedir=dest, path_formats=path_formats, fragment=True
+                ) for item in items
+            ]
+            items_paths = ["#EXTM3U"] + items_paths
             if not pretend:
                 with open(playlist, "w") as playlist_file:
-                    playlist_file.write("#EXTM3U" + "\n")
+                    playlist_file.writelines('\n'.join(items_paths))
 
         self._parallel_convert(dest, opts.keep_new, path_formats, fmt, pretend,
-                               link, hardlink, threads, items, playlist)
+                               link, hardlink, threads, items)
 
     def convert_on_import(self, lib, item):
         """Transcode a file automatically after it is imported into the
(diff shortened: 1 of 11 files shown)


JOJ0:
## Usage examples

### Simple example

Using a regular beets query to save all playlist of all files of an artist sorted chronologically:

```
beet convert -y "artist:Amon Tobin" year+ --pretend --dest ~/Downloads/tmp --playlist 0-amon_tobin_chronologically.m3u
```

### More advanced example

A smart playlist should be exported to the export folder alongside its media files.

Required config:

```
plugins:
    - convert
    - playlist

playlist:
    playlist_dir: ~/Music/Techno/0-playlists
convert:
    dest: /Users/jojo/Music/Techno/0-export
    no_convert: format:AAC, format:WMA, format:ALAC, format:FLAC
    threads: 1
    paths:
        default: "$artist - $album - $title [$catalognum] ($year)"
        singleton: "$artist - $album - $title [$catalognum] ($year)"
        comp: "$artist - $album - $title [$catalognum] ($year)"
```

```
beet convert -y playlist:/Users/jojo/Music/Techno/0-playlists/110..130_TechnoJungleDub_Lo.m3u --dest ~/Downloads/tmp --playlist 0-play.m3u8
```

Certainly any playlist file could be exported, no matter if it was created with a beets plugin or manually.

It's intended that the file is saved alongside the media files, in this case the path provided by the `--dest` option. If `--dest` is not provided with the command it's taken from the config file (as the convert plugin handles it already).

beetsplug/convert.py
```diff
@@ -281,6 +291,24 @@ def convert_item(self, dest_dir, keep_new, path_formats, fmt,
...
+                self._log.debug(
+                    "Appending to playlist file {0}",
+                    util.displayable_path(playlist)
+                )
+                # The classic m3u format doesn't support special characters in
+                # media file paths, thus we use the m3u8 format which requires
+                # media file paths to be unicode. Additionally we use relative
+                # paths to ensure readability of the playlist on remote
+                # computers.
+                dest_relative = util.displayable_path(dest).replace(
```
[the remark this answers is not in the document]
JOJ0: done

JOJ0:
yes there will most probably be an issue with threading I guess. Actually I wanted to ask you for advice on that topic :-) Thanks for the early review I'll think about your ideas and will hopefully get back to this next week and probably have more questions.

sampsyo:
Sounds good! I wonder if there's a way to "batch up" all the new paths and then write them out all at once…

JOJ0:
This is stalling.

There is summer/vacations/other projects need finishing but most of all because I don't get the unittest to work.

Anyway, I have 5 minutes and ask for help. I'm currently trying to write a unittest that checks whether the playlist.m3u8 file was written and is existing. I'm feeling bad already for asking Adrian all the time, he probably has a lot on the plate with this project already. @arsaboo I'm not sure if you're officially a maintainer of the project but you seem highly motivated lately and since you have quite some experience with your Spotify and Plex stuff going on you might be able to help: I'm trying to check if a file was created, but the file doesn't seem to be there: https://github.com/beetbox/beets/pull/4399/commits/147e01885bae522da613f63f5d54c6c1cee88213
This is what I get:
```
(beets) ➜  beets git:(convert_playlist) python -m unittest test.test_convert.ConvertCliTest.test_playlist
Sending event: item_copied
no user configuration found at /var/folders/3t/l575q0f96fx91pp61zb158z00000gn/T/tmpettdj4j2/config.yaml
data directory: /var/folders/3t/l575q0f96fx91pp61zb158z00000gn/T/tmpettdj4j2
plugin paths:
Sending event: pluginload
the artist - älbum - tïtle 0
Convert? (Y/n) convert: Creating playlist file: /var/folders/3t/l575q0f96fx91pp61zb158z00000gn/T/tmpettdj4j2/convert_dest/playlist.m3u8
E
ERROR: test_playlist (test.test_convert.ConvertCliTest)
  File "/Users/jojo/git/beets/test/test_convert.py", line 254, in test_playlist
    self.run_convert('--playlist', 'playlist.m3u8')
  File "/Users/jojo/git/beets/test/test_convert.py", line 156, in run_convert
    return self.run_convert_path(self.item.path, *args)
  File "/Users/jojo/git/beets/test/test_convert.py", line 152, in run_convert_path
    return self.run_command('convert', *args)
  File "/Users/jojo/git/beets/test/helper.py", line 451, in run_command
    beets.ui._raw_main(_convert_args(list(args)), lib)
  File "/Users/jojo/git/beets/beets/ui/__init__.py", line 1291, in _raw_main
  File "/Users/jojo/git/beets/beetsplug/convert.py", line 491, in convert_func
    m3ufile.write()
  File "/Users/jojo/git/beets/beets/util/__init__.py", line 184, in write
    with open(self.path, "w") as playlist_file:
FileNotFoundError: [Errno 2] No such file or directory: b'/var/folders/3t/l575q0f96fx91pp61zb158z00000gn/T/tmpettdj4j2/convert_dest/playlist.m3u8'

Ran 1 test in 0.068s

FAILED (errors=1)
(beets) ➜  beets git:(convert_playlist)
(beets) ➜  beets git:(convert_playlist) ls -l /var/folders/3t/l575q0f96fx91pp61zb158z00000gn/T/tmpettdj4j2/
total 72
drwxr-xr-x  3 jojo  staff     96 Aug 10 07:24 libdir
-rw-r--r--  1 jojo  staff  36864 Aug 10 07:24 library.db
(beets) ➜  beets git:(convert_playlist)
```


I think I misunderstand something pretty basic here. Maybe you have an idea to further debug this. Thanks so much in advance!

sampsyo:
Hmm… my best guess here is that the code currently tries to write the playlist file (i.e., does `m3ufile.write()`) _before_ the actual conversion happens (`self._parallel_convert()`). That latter call has the responsibility to create the destination folder—in the test harness, that `convert_dest` directory. So maybe this would just work if the playlist file were created _after_ that?

JOJ0:
Hi, thanks so much @sampsyo, it's working now, some progress is to note and some new questions arise:

- I'm not sure how I would ensure the compatibility with playlists created _on_ Windows _for_ Windows. Some pointers to helper functions would  be great.
- Would it be ok for this PR to go through if the M3UFile class only provides a good basis for further implementation of features required by other m3u playlist oriented plugins?
  - What would be required to achieve that? Currently I'm testing, if I haven't forgotten something, all the features the M3UFile class is capable of, even if not all is required by the convert+playlist feature (e.g load())

Reason is: I'm a little afraid that trying to port all of the existing m3u file things in playlist, smartplaylist and play plugins is a lot of errorprone and delicate work. I'm not sure if I'm willing to do that and as said want to provide a good and rock solid basis for further advancing the M3UFile class. What do you think?

sampsyo:
Well, awesome! I'm glad this seems to be working.

It's a good question about Windows. I'm not a Windows expert either… I think the best route here would probably be to just assume that everything's fine on that platform (as long as we're correctly using `syspath` as we do elsewhere in beets) and then let the bug reports come in. :smiley: Having concrete evidence of how things go wrong can really help point in the direction of the right fix, in my experience.

And sure! Starting with a special-purpose M3U-writing utility, and expanding its functionality on demand, seems like a perfectly good plan to me.

arsaboo:
@JOJ0 sorry....was out for a conference and could not respond. I am not a maintainer....just an active user of beets and trying to add features that I care about. In any case, looks like the issue was resolved. I hope to see this feature in beets soon. I would like to use parts of it to covert playlists.

JOJ0:
It's currently untested when this feature is used together with
the `--keep-new` option. The `dest` option seems to be mandatory, so supposedly there shouldn't be a problem with writing the playlist file to this path. I'm a little afraid to test this right now, anyone?

JOJ0:
Interesting, now I'm getting exactly the error I was kind of hoping for. Something is different on Windows but atm I don't know how to proceed: Testfile was created on macOS Catalina, which I guess uses unicode. Windows can't read this path, even though I use syspath() wrapped around each and every entry in a playlist:
https://github.com/beetbox/beets/runs/8009869557?check_suite_focus=true#step:7:115

JOJ0:
Hi @sampsyo , I think this PR finally is ready for the final review. I tried to make sure that unicode files reading and writing is working on Windows by adding some Windows specific unittests. I'm still not 100% sure if I understood syspath() entirely. I disabled the prefix when adding file  paths to the playlist. I'm not sure if syspath() would  be required at all when adding to an m3u playlist file but I did it anyway (https://github.com/beetbox/beets/pull/4399/files#diff-f20a2ab41ac3a3c83575cac2b871653999b0e4d28c27336a1f3897a004f5d6daR49)
Hopefuilly if someone experiences issues with that approach they will report it and I'll be able to fix it. Please ping me if an issue like that would come in. Thanks!

Please specifically review the Windows tests (and the corresponding fixture files) and tell me whether you think they make sense like that:
https://github.com/beetbox/beets/pull/4399/files#diff-277edfc80947a6330096d584f8c3a0e418445dbc69976834a2ea474ac856c681R65-R89

https://github.com/beetbox/beets/pull/4399/files#diff-277edfc80947a6330096d584f8c3a0e418445dbc69976834a2ea474ac856c681R108-R118

JOJ0:
Okay, okay, I did some homework and finally read the msdn article that's linked within the syspath() function (https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247.aspx). I think I get it now. It enables Windows to handle paths longer than the maximum of 260 characters. Using syspath() always adds the prefix of \\?\ to paths, because it can't hurt, even when paths are shorter. Still I'm unsure if it's correct to apply it when loading (https://github.com/beetbox/beets/pull/4399/files#diff-f20a2ab41ac3a3c83575cac2b871653999b0e4d28c27336a1f3897a004f5d6daR49) an m3u playlist's entry and when adding one (https://github.com/beetbox/beets/pull/4399/files#diff-f20a2ab41ac3a3c83575cac2b871653999b0e4d28c27336a1f3897a004f5d6daR49).

JOJ0:
I don't want to leave all the clutter of the last 3 messages standing as the last message. All this is probably esoteric at the moment and as always there might be room for improvement. A final review and merge could be a sensible next step. Thanks a lot in advance.

JOJ0:
Hi @sampsyo and @wisp3rwind, how can I help getting this feature merged? It might just be a matter of time and resources on your end. Please tell me if there's anything I can do to ease up the final review for you!

JOJ0:
There where some changes in master in test_convert.py and changelog.rst, rebased them in, fixed conflicts, did some test runs (`tox -e lint`, `python -m unittest discover -p 'test_convert*'` and `python -m unittest discover -p 'test_m3u*'`, which ran fine.

gh actions checks seem to be fine again as well.

JOJ0:
Ready for a final review.

arsaboo:
@JOJ0 I am trying to add a feature to import playlists (e.g., from Spotify). While I want to import the playlist in Plex eventually, I thought it may be a good idea to use m3u as an intermediate step. So, we can import any playlist (once we have the parsers ready) and create an m3u playlist. I know this PR does not search for tracks (based on album, title, etc.), but I am thinking that we can add that feature to make it more generic. What are your thoughts on that?

JOJ0:
@arsaboo I like the idea of importing playlists from Spotify, though I can't imagine how this would work if a track initially wasn't imported into beets by using the Spotify source plugin but any other source plugin or MusicBrainz. Would you then try to match by searching for artist and title? That could work in some cases, but probably not safe enough.

As for the M3UFile class in this PR: It's purpose should be to become a general way of reading and writing such files from anywhere in beets. Currently it has limited features:

A list of files needs to be compiled beforehand and then passed to the class: https://github.com/beetbox/beets/pull/4399/files#diff-f20a2ab41ac3a3c83575cac2b871653999b0e4d28c27336a1f3897a004f5d6daR53-R65
Then it can be written to disk using the `write` method.

For loading an m3u file from disk there is the `load` method.

That's it currently, but could be a foundation for future refactoring of all the existing plugins that do stuff with m3u files. If you want to be the first "user" of the class, I'd be happy about it. I didn't implement any more "magic" of querying into it on purpose because when I looked into what all the other plugins do with playlists and how they do it it felt like a neverending story to integrate them all to use this class. So I tried to keep it as minimal as possible.

arsaboo:
I don't want you to change anything with the current PR....let us get this merged first (it has been pending for a while anyways :p)

My plan was to first search using the Spotify id. If not found, match using album, title, and artist (some sort of fuzzy match). Note that Spotify is only of the sources, there are other sources that this approach could be applied to - JioSaavn, Gaana (or any other source that can provide structured lists). 

We can use any pre-defined source to input the playlist, identify the matching tracks, and log the tracks that were not found. Once we have a list of matched tracks, we can create an m3u file or directly import it into something like Plex (which is where I want them to be).

JOJ0:
Hi @Serene-Arc sorry for bothering with a rather huge PR but I think you are the right person to have a look at this. I'm wondering if my tests make sense. Could you have a look at:

- whether the tests that check the M3UFile class cover the functionality as good as possible or what could be added (or even what is redundant)= https://github.com/beetbox/beets/pull/4399/files#diff-277edfc80947a6330096d584f8c3a0e418445dbc69976834a2ea474ac856c681 
- and if the tests for the actual feature could be better, I only have 2 tests implemented here and no idea if there could be done more: https://github.com/beetbox/beets/pull/4399/files#diff-bbbb8de7fcf63a7338c349f85b860817966c3400530c0ffeac2518e5cf2c1fab

Thanks for any hints on how to make sure this is tested well. I'd like to get it merge-ready finally :-)

JOJ0:
@wisp3rwind may I ask for your advice on a specific thing here: I recently came across about a PR where you cleaned up a little and moved the wrapping through syspath to the very last thing happening _before_ a filesystem operation is done by beets: https://github.com/beetbox/beets/pull/4675/files#diff-1ab0ce830f9dd337e8c57f809d3488b4dfb78c12a61a2a526100038e05a7ca54R29

I think back when I wrote this feature / opened this PR I did not have a good understanding of what and why syspath() wrapping is required. I did this: 

https://github.com/beetbox/beets/pull/4399/files#diff-ac0584bc04d691dc3bb362557c68b83e7e71f491b7e4e0b42013e3a73cf30dd3R488-R496

In words: I'm generating a list of paths to media files that then will be saved into an m3u playlist file. I wrap every one of these paths into `syspath()`

I now think that this actually is not the purpose of syspath() since I do not actually do a filesystem operation from within beets. I should not use `syspath()` here. Is that correct?

beetsplug/convert.py
```diff
@@ -461,8 +472,28 @@ def convert_func(self, lib, opts, args):
...
+                           util.normpath(playlist))
+            items_paths = [
+                util.syspath(item.destination(
+                    basedir=dest, path_formats=path_formats, fragment=True
+                )) for item in items
+            ]
+            if not pretend:
+                m3ufile = M3UFile(playlist)
+                m3ufile.set_contents(items_paths)
+                m3ufile.write()
```
JOJ0, line 496: @wisp3rwind may I ask for your advice on a specific thing here: I recently came across about a PR where you cleaned up a little and moved the wrapping through syspath to the very last thing happening _before_ a filesystem operation is done by beets: https://github.com/beetbox/beets/pull/4675/files#diff-1ab0ce830f9dd337e8c57f809d3488b4dfb78c12a61a2a526100038e05a7ca54R29

I think back when I wrote this feature / opened this PR I did not have a good understanding of what and why syspath() wrapping is required. Have a look at line 488 above.

In words: I'm generating a list of paths to media files that then will be saved into an m3u playlist file. I wrap every one of these paths into `syspath()`

I now think that this actually is not the purpose of syspath() since I do not actually do a filesystem operation from within beets. I should not use `syspath()` here. Is that correct?

Serene-Arc:
@JOJ0 I'll start going through it!

beetsplug/convert.py
Serene-Arc: As an aside, I'm preparing to go through the codebase and convert everything to `pathlib` which will negate the need for syspath calls.

JOJ0: Oh yeah pathlib! That is a cool package! I use it all the time in my projects! Great idea!

JOJ0:
Any chance you found time looking through some of the tests @Serene-Arc? :-)

beetsplug/convert.py
wisp3rwind: Sounds right: Currently, beets' code represents every path internally as a `bytes` object. To keep things manageable, paths should be converted to this representation immediately upon ingestion from any source (CLI, input files, library APIs, ...), and converted to other representations as late as possible (usually just upon leaving beets' direct control, i.e. when writing them to a file or the DB, etc).

To be honest, I'm also not entirely sure without studying the code again what exactly is the right approach when receiving paths from, say, the commandline. Anyway, yes, `syspath` seems like the wrong function here, at least without further research on whether it does the right thing on each platform. The question to answer here would be how m3u files should normally be encoded in each case, and ensure that beets adheres to that.

As a matter of fact, does your code actually run on Linux?  At a first glance, I would expect that `M3UFile.write()` would crash when trying to join the paths (`bytes` as returned by `syspath`) with newlines.

Sorry for the late reply, I missed that you mentioned me here!

JOJ0: Hi @wisp3rwind thanks a lot for the reply, late or not, it's helpful! :-)

One thing before I answer more: Is it possible that you are confusing the `syspath` function with `normpath` or `bytestring_path` (which is called by `normpath` for example.

Reason being: `syspath` doesn't do anything on Linux, it just returns as-is. As I understood `syspath`, it actually is ment for sanitizing Windows paths:

https://github.com/beetbox/beets/blob/2e18f84b380911c1d3ec2e5b76513b3924db2188/beets/util/__init__.py#L395-L426

https://github.com/beetbox/beets/blob/2e18f84b380911c1d3ec2e5b76513b3924db2188/beets/util/__init__.py#L138-L144

JOJ0: > The question to answer here would be how m3u files should normally be encoded in each case, and ensure that beets adheres to that.
 
Exactly! I did my research back then when I started to code this around last summer. It turns out that there is no real standard of what m3u is. It is rather loose, therefore tools do what the dev's thought it should do. The best we can do is do a compromise to work for the most circumstances. The main points from my research are:

- The best summary / documentation seems to be this wikipdia page: https://en.wikipedia.org/wiki/M3U
  - Intersting points:
    - m3u contains **platform specific paths**
    - m3u**8** ensures paths are unicode! -> This is what I'm aiming for, this is how this feature is documented so far. So this is my actual review question: Am I aiming at this goal?
  - Even proprietory DJ Tools (which are one of my target tools as a DJ obviously) might not do a 100% the right thing, thus this feature needs some flexibility. (One tiny example: Native Instruments Traktor understands m3u8 formatted playlists (utf-8 paths) but it does not understand the file ending of m3u8)

JOJ0: >

JOJ0: > > As a matter of fact, does your code actually run on Linux? At a first glance, I would expect that `M3UFile.write()` would crash when trying to join the paths (`bytes` as returned by `syspath`) with newlines.
> 
> Code is tested on macOS Catalina and Manjaro Linux (my former dev machine and my current main dev machine ;-))
> 
> One thing before I answer more: Is it possible that you are confusing the `syspath` function with `normpath` or even `bytestring_path` (which is called by `normpath`)
> 
> Reason being: `syspath` doesn't do anything on Linux/Posix, it just returns as-is. As I understood `syspath`, it actually is ment for sanitizing Windows paths:


Ah, sorry, now I understand: Since internally we are having `bytes` paths,. `syspath`, even when not doing anything on Non-Windows systems, will return exactly that: The internatl representation of paths, we still have `bytes`

So to make my question more concrete now: My goal is to now make sure that all paths that goe in the m3u list should be converted to UTF-8. Which function from utils could I use?

`displayable_path()`? Sounds odd but actually it translates `bytes` to `unicode`.

JOJ0: https://github.com/beetbox/beets/blob/2e18f84b380911c1d3ec2e5b76513b3924db2188/beets/util/__init__.py#L376-L392

JOJ0:
Your review was instrumental @wisp3rwind, I went through all the things in this PR again and realised that I didn't follow the "rule" in beets to handle stuff internally as bytes at all.

Without talking too much, I think the changes in these commits speak for themselves even without having much context:

The writing part:

https://github.com/beetbox/beets/pull/4399/commits/8b1e226397c873c8acb80f3725b855aec5647b5e

The reading part:

https://github.com/beetbox/beets/pull/4399/commits/c859367d3fc962cab489cecf9cfedfa26cd0a974

Both of you @Serene-Arc and @wisp3rwind, if you want to take a quick look that's appreciated. 

Some background: Back when I started to work on this I had the goal of writing m3u8, i.e. unicode encoded paths in files. I took the route of let Python handle the files in text mode and encode unicode that way. Now I changed everything to do the "beets way" and keep everything as bytes internally, writing and reading in binary mode and making sure encoding is handled via the `bytestring_path` function. I think / I hope I'm on the right track now :-)

JOJ0:
I think this is finally ready to go. There is error handling, unittests for Posix and Windows, syspath and bytestring_path wrappers are used and I use it often personally for exporting from my collection. There have been reviews by experienced beeters and I tried to incorporate there suggestions. The one thing that is lacking is testing thoroughly on Windows. We'll see if it holds and I'll be here for fixing.

The m3u class that's included in this PR is ready to be used in other parts of beets. It's minimal functionality is intended but I think by adding just a few features a next step could be to use it in the smartplaylist plugin, which is the one place in beets where I stole some error handling practices from ;-)

JOJ0:
Whoops, reminder to myself: A markdown hyperlink is not a rst hyperlink. Quickfix this in master ;-)

PR #4405, opened 2022-07-08
Spotify: Add album name for singleton imports

arsaboo:
As discussed in #4398, this PR adds the Spotify album information when importing tracks. I understand there are unresolved issues about which album to import for other taggers (e.g., MB). Thus, this PR only works for Spotify taggers. The behavior is unchanged for other sources. 

## To Do

### DIFF
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -297,7 +297,6 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
             album = track_data['album']['name']
         except KeyError:
             album = None
-            pass
         return TrackInfo(
             title=track_data['name'],
             track_id=track_data['id'],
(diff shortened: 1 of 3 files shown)


beetsplug/spotify.py
```diff
@@ -291,11 +291,19 @@ def _get_track(self, track_data):
         :rtype: beets.autotag.hooks.TrackInfo
         """
         artist, artist_id = self.get_artist(track_data['artists'])
+
+        # Get album information for spotify tracks
+        try:
+            album = track_data['album']['name']
+        except KeyError:
+            album = None
+            pass
```
sampsyo, line 300: No need for this line.

sampsyo (commented):
Looks great to me!! I think it's good to go, modulo an extraneous `pass` statement, assuming it works on your music how you want it to?

arsaboo:
> Looks great to me!! I think it's good to go, modulo an extraneous `pass` statement, assuming it works on your music how you want it to?

Indeed it does. At least now when we encounter folders with random tracks we can import them as singletons and then process them further.

sampsyo:
Excellent; many thanks for your work on this!

Issue #4404, opened 2022-07-08
LibModel.items() confusion


govynnus:
The docs describe [`LibModel.items()`](https://beets.readthedocs.io/en/stable/dev/library.html#beets.library.LibModel.items) to
> Iterate over (key, value) pairs that this object contains. Computed fields are not included.

This is the case for `Item.items()`, but [`Album.items()`](https://github.com/beetbox/beets/blob/472c3ab7975e5933f8111a2182f177a0fe5642d5/beets/library.py#L1145) gives all the `Item`s associated with the `Album`.

`LibModel.items()` is actually defined in [`Model.items()`](https://github.com/beetbox/beets/blob/472c3ab7975e5933f8111a2182f177a0fe5642d5/beets/dbcore/db.py#L170), and overwritten in the `Album` class.

So I'm wondering what the intended behavior is :)

I know I have used `Album.items()` as it currently is in the past (and it seems to me a sensibly named function). Just now I tried to use it in the sense defined in the docs and ran into this problem (although I want computed fields so not using it any more). In the docs sense I guess the `items()` mimics `dict.items()`, but does create some confusion with `Item`s :D

sampsyo:
You've hit the nail on the head. It's an unfortunate naming collision—one that exists mainly for historical reasons. Namely, the `Item` and `Album` objects predate `dbcore`; we refactored the latter out of the core beets library to bring a little more abstraction and modularity. So the `Album.items()` method is very old and special-purpose, while the new(er) `Model.items()` is generic.

So yes, it's not a great situation, but it's not clear what we should do instead. Perhaps rename `Album.items()` to something else someday?

wisp3rwind:
> So yes, it's not a great situation, but it's not clear what we should do instead. Perhaps rename `Album.items()` to something else someday?

While unfortunate; this seems like a change that would break many third-party plugins. In my opinion, we should avoid renaming `.items()` (unless there's a stronger reason than being somewhat confusing). Maybe extending the `Album.items()` docstring to point this out would help to alleviate the issue?

govynnus:
Yep I think just documenting the conflict would be good. Doesn't seem like it's been too much of an issue in the past.

wisp3rwind:
> Yep I think just documenting the conflict would be good. Doesn't seem like it's been too much of an issue in the past.

Would you be able to make a quick PR for this?

govynnus:
Sure, I'll get round to it later or tomorrow morning (UTC)

PR #4412, opened 2022-07-12
Document Album.items() / LibModel.items() conflict

govynnus:
Closes: https://github.com/beetbox/beets/issues/4404

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -1145,6 +1145,9 @@ class Album(LibModel):
     def items(self):
         """Return an iterable over the items associated with this
         album.
+
+        This method predates the :meth:`LibModel.items` method inherited
+        from :meth:`beets.dbcore.Model.items`.
         """
         return self._db.items(dbcore.MatchQuery('album_id', self.id))
 
--- a/beets/library.py
+++ b/beets/library.py
@@ -1146,8 +1146,10 @@ class Album(LibModel):
         """Return an iterable over the items associated with this
         album.
 
-        This method predates the :meth:`LibModel.items` method inherited
-        from :meth:`beets.dbcore.Model.items`.
+        This method conflicts with :meth:`LibModel.items`, which is
+        inherited from :meth:`beets.dbcore.Model.items`.
+        Since :meth:`Album.items` predates these methods, and is
+        likely to be used by plugins, we keep this interface as-is.
         """
         return self._db.items(dbcore.MatchQuery('album_id', self.id))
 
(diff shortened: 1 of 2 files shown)


docs/dev/library.rst
```diff
@@ -99,6 +99,10 @@ synchronized (via load or store) with the database.
 
     .. automethod:: items
 
+    .. note::
+       The :py:meth:`Album.items` method is not inherited from
+       :py:meth:`LibModel.items` for historical reasons.
+
```
wisp3rwind, line 105: Wouldn't this make more sense below with the `Album.items` docs?

beets/library.py
```diff
@@ -1145,6 +1145,9 @@ def _getters(cls):
     def items(self):
         """Return an iterable over the items associated with this
         album.
+
+        This method predates the :meth:`LibModel.items` method inherited
+        from :meth:`beets.dbcore.Model.items`.
```
wisp3rwind, line 1150: Maybe make it even clearer what this is about (at the risk of becoming overly verbose)? Along the lines of

```suggestion
        This method conflicts with `.items()` in Python's mapping protocol, 
        which is implemented by :meth:`LibModel.items` method (inherited
        from :meth:`beets.dbcore.Model.items`).
        Since :meth:`Album.items` predates these methods, and is likely to be used
        by plugins, we keep this interface as-is.
```

wisp3rwind (commented):
Thanks! I have a few suggestions how this might be reworded for clarity. What do you think, would that indeed improve clarity?

docs/dev/library.rst
govynnus: I think some kind of note here makes sense because it would be reasonable to assume methods documented in LibModel can also be used by its children, but that's not the case here.

Although the main explanation is left with the Album.items docs as you've expanded upon.

beets/library.py
govynnus: Yes that's a good idea. I think the bit about the mapping protocol might be overly verbose; maybe just have something like this?

```
        This method conflicts with :meth:`LibModel.items`, which is inherited
        from :meth:`beets.dbcore.Model.items`.
        Since :meth:`Album.items` predates these methods, and is likely to be
        used by plugins, we keep this interface as-is.
```

docs/dev/library.rst
wisp3rwind: Fair enough.

beets/library.py
wisp3rwind: I don't have a clear preference here: It's up to you :)

sampsyo (approved):
Looks perfect to me! Thanks for the thoughtful wording, @govynnus.

PR #4416, opened 2022-07-14
Change the option 'Remove old' -> 'Replace old' in the importer CLI

Sciencentistguy:
## Description

The CLI currently provides the following options to handle adding a duplicate album:

```
[S]kip new, Keep all, Remove old, Merge all?
```

This PR replaces the word `Remove` with `Replace`:

```
[S]kip new, Keep all, Replace old, Merge all?
```

In my opinion, this more accurately reflects what the operation does.

## To Do

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -1473,7 +1473,7 @@ def resolve_duplicates(session, task):
                 # Keep both. Do nothing; leave the choice intact.
                 pass
             elif duplicate_action == 'r':
-                # Remove old.
+                # Replace old.
                 task.should_remove_duplicates = True
             elif duplicate_action == 'm':
                 # Merge duplicates together
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -857,7 +857,7 @@ class TerminalImportSession(importer.ImportSession):
             ))
 
             sel = ui.input_options(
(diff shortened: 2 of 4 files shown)


sampsyo:
Hi! Thanks for the interesting proposal! It's true that the end effect of this option is that the old music will be replaced with the new music. I certainly don't have a strong objection, but the reason it's currently "Remove" is to suggest exactly what will happen to the old music: it's like running the `beet remove` command.

Anyway, it's a subtle difference, so no strong feeling either way—but maybe it would be nice to get other perspectives about what this rephrasing would mean to them.

Sciencentistguy:
My primary disliking of the current phrasing is that "remove existing" does not immediately say to me "and import this on top of it" (i.e. replace)

arsaboo:
The other problem (and I only realized it recently) is that "remove old" removes all the old files and not only the files being replaced. So, for example, if there are 5 tracks already in the library and you add one new track that already exists and you choose "remove old", it will delete all the 5 tracks in the library and only retain the one new track. With the current behavior (of deleting the old tracks) maybe `Remove` is not a bad label. 

Having said that, I prefer the `replace` behavior, when only the duplicate tracks are replaced. Once we have that behavior, I think we should change the label.

PR #4429, opened 2022-07-30
Fix old alias

aereaux:
It looks like the convert option for wma used to be called windows media.  We could just remove this alias, but might be good to keep for backwards compatibility.

## Description

Fixes error when trying to use wma format in convert.

## To Do


sampsyo:
Thanks for noticing this! Looks good.

PR #4431, opened 2022-07-31
[Small PR] Add spotify_lastupdatedat field

arsaboo:
## Description
Added `spotify_lastupdatedat` field that can be used to economize on the API calls. For example, we can use this field to update only those tracks that have not been updated in the last 6 months. 

## To Do


sampsyo:
OK, sounds good! I think this may need a quick changelog entry?

Also, it might make sense to consider a simpler name… maybe even just `spotify_updated` would suffice to make it clear what's going on?

sampsyo (approved):
Excellent; thanks!!

PR #4433, opened 2022-08-03
Fix get item file in web plugin

vicholp:
## Description

Hi! In the last version of Flask, the `attach_filename` parameter was removed of the `send_file` function, used by the web plugin to stream the file. So thing change broke the web plugin. [Flask changelog](https://flask.palletsprojects.com/en/2.2.0/changes/#version-2-2-0)

Also, I added a test for that feature.

## To Do

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -324,7 +324,7 @@ def item_file(item_id):
     response = flask.send_file(
         item_path,
         as_attachment=True,
-        attachment_filename=safe_filename
+        download_name=safe_filename
     )
     response.headers['Content-Length'] = os.path.getsize(item_path)
     return response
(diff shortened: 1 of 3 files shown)


arogl:
Do we also need to check or lock the flask level [setup.py](https://github.com/beetbox/beets/blob/master/setup.py#L104)?

vicholp:
> Do we also need to check or lock the flask level [setup.py](https://github.com/beetbox/beets/blob/master/setup.py#L104)?

I think that is way to go to avoid breaks like this, and `flask~=2.2.0` in web plugin and testing dependencies should works. But I'm not sure if that can have unintended effects.

sampsyo (approved):
Awesome; thanks for noticing and fixing this!! This all looks good, so let's merge once the CI results come back.

Adding a version constraint in `setup.py` would also probably be a good idea. But let's tackle that in a separate PR and see if affects anything unexpected?

PR #4438, opened 2022-08-12
Add path template "sunique" to disambiguate between singleton tracks

jaimeMF:
Solution for https://github.com/beetbox/beets/discussions/4436. It is mostly copied from the `aunique` template. I don't know if there could be better defaults for the `disambiguators`.

## To Do

### DIFF
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -55,6 +55,11 @@ aunique:
     disambiguators: albumtype year label catalognum albumdisambig releasegroupdisambig
     bracket: '[]'
 
+sunique:
+    keys: artist title
+    disambiguators: year trackdisambig
+    bracket: '[]'
+
 overwrite_null:
   album: []
   track: []
--- a/beets/library.py
+++ b/beets/library.py
@@ -1683,75 +1683,17 @@ class DefaultTemplateFunctions:
         if album_id is None:
             return ''
 
-        memokey = ('aunique', keys, disam, album_id)
+        memokey = self._tmpl_unique_memokey('aunique', keys, disam, album_id)
         memoval = self.lib._memotable.get(memokey)
         if memoval is not None:
             return memoval
 
-        keys = keys or beets.config['aunique']['keys'].as_str()
-        disam = disam or beets.config['aunique']['disambiguators'].as_str()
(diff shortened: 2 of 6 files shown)


sampsyo:
Wow! It's pretty cool that it was possible to make this work. I would have suspected it would be more complicated.

It seems like there are two major things we'd need to do to make this merge-ready:

* Abstract out the shared functionality between `aunique` and `sunique`. We'd want to avoid any code duplication by writing a single method that can work like both, and then turn the two template functions into different calls to that function.
* Make sure `_memotable` is handled correctly elsewhere. It's a complicated bit of functionality that requires invalidation whenever the database changes. We'll want to make sure that database changes that could affect `sunique` correctly trigger this invalidation.

jaimeMF:
> It seems like there are two major things we'd need to do to make this merge-ready:
> 
>     * Abstract out the shared functionality between `aunique` and `sunique`. We'd want to avoid any code duplication by writing a single method that can work like both, and then turn the two template functions into different calls to that function.

I will refactor it to a common method.

>     * Make sure `_memotable` is handled correctly elsewhere. It's a complicated bit of functionality that requires invalidation whenever the database changes. We'll want to make sure that database changes that could affect `sunique` correctly trigger this invalidation.

I will try to have a look at how is handled and if any change is needed.

sampsyo:
Wow; awesome work!! This is looking very snazzy already.

Would you mind trying your hand at the documentation? We'll want to add a section about `sunique` that ideally makes reference to the `aunique` docs (again avoiding the need for exhaustive duplication of the same concepts). We will also need a brief changelog entry.

jaimeMF:
> Would you mind trying your hand at the documentation? We'll want to add a section about `sunique` that ideally makes reference to the `aunique` docs (again avoiding the need for exhaustive duplication of the same concepts). We will also need a brief changelog entry.

I will do it in the following days.

jaimeMF:
I have added a short descriptions to the docs and an entry to the changelog.

sampsyo:
Truly awesome!! Thank you for the fast and high-quality work addressing this. I'm super impressed this was achievable without too much fuss! 👏

PR #4441, opened 2022-08-17
Change the prefix for exact match queries

sampsyo:
#4251 added exact match queries, which are great, but it was subsequently pointed out in https://github.com/beetbox/beets/pull/4251#issuecomment-1069455483  that the `~` query prefix was already in use.

So this changes the prefix from `~` to `=~`. A little longer, but hopefully it makes the relationship to the similarly-new `=` prefix obvious.


wisp3rwind:
LGTM!

PR #4444, opened 2022-08-19
Add Beetstream in the plugin list

BinaryBrain:
## Description

Adding my plugin in the documentation as it was suggested [in the discussions](https://github.com/beetbox/beets/discussions/4375).

Also, would you like if I add this plugin code in the `beetsplug` folder of this repo? What are the prerequisites for it?

### DIFF
--- a/docs/plugins/index.rst
+++ b/docs/plugins/index.rst
@@ -287,6 +287,8 @@ Here are a few of the plugins written by the beets community:
 
 * `beetcamp`_ enables **bandcamp.com** autotagger with a fairly extensive amount of metadata.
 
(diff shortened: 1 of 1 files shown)


sampsyo:
Great; thanks for adding this!

There aren't hard-and-fast rules about when it makes sense to contribute a plugin to the central beets repository. The main thing is that we, as a project, commit to a certain level of code quality and to ongoing maintenance—so it needs a certain amount of broad appeal and some work to make sure it meets our usual coding standards and such. But other than that, no particular objection here!

BinaryBrain:
Great! I think what's really missing in term of quality is unit tests. Otherwise I feel quite comfortable with it.

PR #4463, opened 2022-08-30
Fix typo

Serene-Arc:
## Description

This fixes an issue where fetchart, when the config file was set to move files but the command-line option `-c` was passed, would remove the cover art files anyway, even though the rest of the music files were left alone. This meant it was changing the on-disk files during a copy.

The solution I coded was simply copying the cover art to a named temporary file when drawing a candidate from the filesystem. I rewrote most of the failing tests but I haven't added any new ones.

## To Do

- [-] Tests. (Encouraged but not strictly required.)


wisp3rwind:
I think I see the problem, but at this point I don't quite understand why the temporary file is necessary — could you clarify why that is needed, and it's not enough to simply ensure that `src_removed` is always set correctly? (Which might require that the art sources return a flag indicating whether to remove the file.)

Serene-Arc:
@wisp3rwind I tried that to begin with but if you look at the code, you can see that any modifications to the file overwrite the original. So for my configuration, I have it set to convert everything to png. But if the source file is a jpeg, then it will be overwritten with the png version. To get the source file to remain untouched, I thought it would be simplest to just shift all of the processing to a temporary file.

PR #4132, opened 2021-11-01
Add VGMdb support

Eldarock:
Add vgmdb support for matching and fetchart.
HIGHLY ALPHA, just wanted to do a PR because i want your feedbacks and don't know where exactly to contact you.
Let me know what is wrong (it's like a "night coding session" work), i'm already using it on my massive VGM library.

All comments are welcomed, just started python 2 days ago.

Credits to early vgmdb plugin : 
https://github.com/whtsky/VGMdb-BeetsPlug
https://github.com/SomeTiramisu/VGMdb-BeetsPlug

PR #4469, opened 2022-09-02
Add VGMdb Support, Again

jojoxd:
## Description

This is a continuation of #4132, but with the requested changes and some bugfixes.

Note that this is a work in progress.

## To Do


jojoxd:
Note that Track Credits (artist, lyricist, composer, arranger on `TrackInfo`) can't be loaded from VGMdb yet (See https://github.com/vgmdb/tracker/issues/26)

Eldarock solved this by setting the global album attributions to the track data, which is _super_ incorrect, IMHO, as this will create a list in the metadata. (That happens [here](https://github.com/beetbox/beets/pull/4469/files#diff-0fa26dda0f9b870dc5c05676285449e07b97aec3b6eea95156641d3dc5724ba2R221))

VGMdb adds barcodes to their data, which can be used to query MusicBrainz (Although most likely, most albums won't be in their dataset)

I'm guessing we can't chain MusicBrainz to fetch the correct track data for artist, lyricist etc.?

sampsyo:
Let's keep this alive! @jojoxd, would it makes sense to merge a limited version of this first, even if it doesn't fetch everything we potentially _could_ fetch in the first version?

PR #4488, opened 2022-09-16
WIP tests: Render ui output to HTML

wisp3rwind:
This is a bare-bones implementation of an idea I've had in mind for a while: Record unit test output and render it to HTML to ease development and review regarding our UI.

Does that sound like a useful tool (@davidswarbrick)? Any other thoughts?

Commit message:
> experiment to help with development of the commandline user interface. The idea is to record representative output from tests, instead of needing to manually run some beet commands to get feedback. This works by
>
> - recording pytest output in a new tox run (tox -e py310-render_ui)
> - parsing stdout in the tox commands_post, and turning it into a basic html document with ansi2html
>
> To be useful, a bunch of new tests should be written that showcase most of the ui and enable color. (It might also be useful to only run this script for those specific tests, instead of the entire test suite.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,7 @@ htmlcov/
 coverage.xml
 *,cover
 .hypothesis/
+ui_render/
 
 # Flask stuff:
 instance/
--- a/extra/render-test-outputs.py
+++ b/extra/render-test-outputs.py
@@ -1,0 +1,83 @@
+#! /usr/bin/env python3
+
+from ansi2html import Ansi2HTMLConverter
+from jinja2 import Template
+import json
+from pathlib import Path
+import re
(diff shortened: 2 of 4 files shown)


sampsyo:
Wow; this is incredibly cool!! I really like this idea—it could be useful not only for tests but also for producing nice screenshots and the like.

sampsyo:
I think it would be great to keep this in play!

wisp3rwind:
I'm probably not going to continue working on this anytime soon; but I came across two related projects that might serve as inspiration:

- https://github.com/borgbackup/borg/pull/6197
- https://github.com/k9withabone/autocast?tab=readme-ov-file

PR #4492, opened 2022-09-18
Fix spotify_updated bug so that the date is not "null"

### LINKED ISSUE #4491: spotify: spotify_updated is always a "null" date
I just checked the `spotify_updated` values and all the values are `1969-12-31 19:00:00`, which prevents the use of any date-related queries. 
 
I have seen this elsewhere, but cannot figure out how to fix the issue. Here's how the date is currently being saved

https://github.com/beetbox/beets/blob/44a7cc74bfe673bcf960c5dc18716b38e33d5ad0/beetsplug/spotify.py#L651

Any idea how to fix it?

arsaboo:
## Description

Fixes #4491


arsaboo:
Tested this and it works as expected.

wisp3rwind:
Thanks!

Issue #4484, opened 2022-09-13
[FEATURE] Adding typings to the codebase


Serene-Arc:
Typings have been supported since python 3.5 and could be applied to the codebase. This would help readability a lot, help with IDE syntax, etc

### Objective

Adding typings to the entirety of the codebase would be the goal although that's obviously a very long way in the future. 

I'm happy to start and try and add these a module at a time.

sampsyo:
Certainly no objection! One thing I'm a little mystified by in this department is how to prioritize. We probably don't want to do it all at once, so we'd need to pick a couple modules or classes to start with. Any thoughts about how to choose those?

Serene-Arc:
@sampsyo There's really no single best approach imo. Perhaps for git reasons, changing the least used parts would be easiest first and less likely to impact other people's PRs and so on. Otherwise maybe doing similar ones in one fell swoop? Like doing all of the plugins (or most of them) at once.

Another approach would be starting with the core-most modules and then propogating outwards.

wisp3rwind:
I would speculate that it might be easiest to start adding types from the bottom up, in order to be able to know the types of called functions when trying to infer types for higher-level code.

It would also be great to add mypy to our CI to avoid adding incorrect annotations.

sampsyo:
Very true. Perhaps starting with `dbcore` would be a useful way in? It's kind of at the bottom of everything, touches all other modules (at least indirectly), and is meant to be abstract enough that the types won't be too crazy.

Serene-Arc:
That's probably a good idea, I can start with that. There is a pitfall that I and everyone else who attempts this has to be aware of though. The typings system as a whole was introduced in 3.5, it was expanded in 3.9 with additional types and annotations. These expanded types can't be used and will cause errors. So if beets is going to maintain compatibility with anything prior to 3.9, the collections typings especially will have to use the older versions.

Serene-Arc:
So I've started typing `dbcore` and I'll have that up in a PR soon for people to view. I still need to go and crossreference things and try and reconcile the typings, but I thought I should say: there are some places where the typings show that something is strange and possibly wrong. Do you want me to part or comment on these? Things like a list being passed when a dictionary is expected.

sampsyo:
Sure! Yeah, it's definitely possible that something's wrong, or that the typing is just kind of subtle and will require some thought to annotate correctly.

[...]

PR #4495, opened 2022-09-22
Add typing to dbcore types modules

Serene-Arc:
## Description

This is part of  #4484, adding typings to the `dbcore` submodule. It's very much still a work in progress. I've made a couple of `# TYPING ERROR` comments where there are obvious things that don't match up with what the typings would seem to suggest. MyPy gives a *lot* more errors than my IDE though so there's clearly some work to be done.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -22,6 +22,10 @@ from collections import defaultdict
 import threading
 import sqlite3
 import contextlib
+from sqlite3 import Connection
+from types import TracebackType
+from typing import Iterable, Type, List, Tuple, NoReturn, Optional, Union, \
+    Dict, Any, Generator
 
 from unidecode import unidecode
 
@@ -29,8 +33,12 @@ import beets
 from beets.util import functemplate
 from beets.util import py3_path
 from beets.dbcore import types
-from .query import MatchQuery, NullSort, TrueQuery, AndQuery
+from .query import MatchQuery, NullSort, TrueQuery, AndQuery, Query, \
+    FieldQuery, Sort
 from collections.abc import Mapping
+
+from ..library import LibModel
+from ..util.functemplate import Template
 
 
 class DBAccessError(Exception):
@@ -58,7 +66,12 @@ class FormattedMapping(Mapping):
 
     ALL_KEYS = '*'
 
-    def __init__(self, model, included_keys=ALL_KEYS, for_path=False):
+    def __init__(
+            self,
+            model: 'Model',
+            included_keys: str = ALL_KEYS,
+            for_path: bool = False,
+    ):
         self.for_path = for_path
         self.model = model
         if included_keys == self.ALL_KEYS:
@@ -73,10 +86,10 @@ class FormattedMapping(Mapping):
         else:
             raise KeyError(key)
 
-    def __iter__(self):
+    def __iter__(self) -> Iterable:
         return iter(self.model_keys)
 
-    def __len__(self):
+    def __len__(self) -> int:
         return len(self.model_keys)
 
     def get(self, key, default=None):
@@ -107,7 +120,7 @@ class LazyConvertDict:
     """Lazily convert types for attributes fetched from the database
     """
 
-    def __init__(self, model_cls):
+    def __init__(self, model_cls: 'Model'):
         """Initialize the object empty
         """
         self.data = {}
@@ -148,12 +161,12 @@ class LazyConvertDict:
         if key in self.data:
             del self.data[key]
 
-    def keys(self):
+    def keys(self) -> List:
         """Get a list of available field names for this object.
         """
         return list(self._converted.keys()) + list(self.data.keys())
 
-    def copy(self):
+    def copy(self) -> 'LazyConvertDict':
         """Create a copy of the object.
         """
         new = self.__class__(self.model_cls)
@@ -169,7 +182,7 @@ class LazyConvertDict:
         for key, value in values.items():
             self[key] = value
 
-    def items(self):
+    def items(self) -> Generator[Tuple]:
         """Iterate over (key, value) pairs that this object contains.
         Computed fields are not included.
         """
@@ -185,12 +198,12 @@ class LazyConvertDict:
         else:
             return default
 
-    def __contains__(self, key):
+    def __contains__(self, key) -> bool:
         """Determine whether `key` is an attribute on this object.
         """
         return key in self.keys()
 
-    def __iter__(self):
+    def __iter__(self) -> Iterable:
         """Iterate over the available field names (excluding computed
         fields).
         """
@@ -269,14 +282,14 @@ class Model:
     """
 
     @classmethod
-    def _getters(cls):
+    def _getters(cls: Type['Model']):
         """Return a mapping from field names to getter functions.
         """
         # We could cache this if it becomes a performance problem to
         # gather the getter mapping every time.
         raise NotImplementedError()
 
-    def _template_funcs(self):
+    def _template_funcs(self) -> NoReturn:
         """Return a mapping from function names to text-transformer
         functions.
         """
@@ -285,7 +298,7 @@ class Model:
 
     # Basic operation.
 
-    def __init__(self, db=None, **values):
+    def __init__(self, db: Optional['Database'] = None, **values):
         """Create a new object with an optional Database association and
         initial field values.
         """
@@ -299,7 +312,12 @@ class Model:
         self.clear_dirty()
 
     @classmethod
-    def _awaken(cls, db=None, fixed_values={}, flex_values={}):
+    def _awaken(
+            cls: Type['Model'],
+            db: 'Database' = None,
+            fixed_values: Mapping = {},
+            flex_values: Mapping = {},
+    ) -> 'Model':
         """Create an object with values drawn from the database.
 
         This is a performance optimization: the checks involved with
@@ -312,7 +330,7 @@ class Model:
 
         return obj
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return '{}({})'.format(
             type(self).__name__,
             ', '.join(f'{k}={v!r}' for k, v in dict(self).items()),
@@ -326,7 +344,7 @@ class Model:
         if self._db:
             self._revision = self._db.revision
 
-    def _check_db(self, need_id=True):
+    def _check_db(self, need_id: bool = True):
         """Ensure that this object is associated with a database row: it
         has a reference to a database (`_db`) and an id. A ValueError
         exception is raised otherwise.
@@ -338,7 +356,7 @@ class Model:
         if need_id and not self.id:
             raise ValueError('{} has no id'.format(type(self).__name__))
 
-    def copy(self):
+    def copy(self) -> 'Model':
         """Create a copy of the model object.
 
         The field values and other state is duplicated, but the new copy
@@ -356,7 +374,7 @@ class Model:
     # Essential field accessors.
 
     @classmethod
-    def _type(cls, key):
+    def _type(cls, key) -> types.Type:
         """Get the type of a field, a `Type` instance.
 
         If the field has no explicit type, it is given the base `Type`,
@@ -364,7 +382,7 @@ class Model:
         """
         return cls._fields.get(key) or cls._types.get(key) or types.DEFAULT
 
-    def _get(self, key, default=None, raise_=False):
+    def _get(self, key, default: bool = None, raise_: bool = False):
         """Get the value for a field, or `default`. Alternatively,
         raise a KeyError if the field is not available.
         """
@@ -431,7 +449,7 @@ class Model:
         else:
             raise KeyError(f'no such field {key}')
 
-    def keys(self, computed=False):
+    def keys(self, computed: bool = False):
         """Get a list of available field names for this object. The
         `computed` parameter controls whether computed (plugin-provided)
         fields are included in the key list.
@@ -457,19 +475,19 @@ class Model:
         for key, value in values.items():
             self[key] = value
 
-    def items(self):
+    def items(self) -> Generator:
         """Iterate over (key, value) pairs that this object contains.
         Computed fields are not included.
         """
         for key in self:
             yield key, self[key]
 
-    def __contains__(self, key):
+    def __contains__(self, key) -> bool:
         """Determine whether `key` is an attribute on this object.
         """
         return key in self.keys(computed=True)
 
-    def __iter__(self):
+    def __iter__(self) -> Iterable:
         """Iterate over the available field names (excluding computed
         fields).
         """
@@ -500,7 +518,7 @@ class Model:
 
     # Database interaction (CRUD methods).
 
-    def store(self, fields=None):
+    def store(self, fields: bool = None):
         """Save the object's metadata into the library database.
         :param fields: the fields to be stored. If not specified, all fields
         will be.
@@ -581,7 +599,7 @@ class Model:
                 (self.id,)
             )
 
-    def add(self, db=None):
+    def add(self, db: Optional['Database'] = None):
         """Add the object to the library database. This object must be
         associated with a database; you can provide one via the `db`
         parameter or use the currently associated database.
@@ -610,13 +628,21 @@ class Model:
 
     _formatter = FormattedMapping
 
-    def formatted(self, included_keys=_formatter.ALL_KEYS, for_path=False):
+    def formatted(
+            self,
+            included_keys: str = _formatter.ALL_KEYS,
+            for_path: bool = False,
+    ):
         """Get a mapping containing all values on this object formatted
         as human-readable unicode strings.
         """
         return self._formatter(self, included_keys, for_path)
 
-    def evaluate_template(self, template, for_path=False):
+    def evaluate_template(
+            self,
+            template: Union[str, Template],
+            for_path: bool = False,
+    ) -> str:
         """Evaluate a template (a string or a `Template` object) using
         the object's fields. If `for_path` is true, then no new path
         separators will be added to the template.
@@ -630,7 +656,7 @@ class Model:
     # Parsing.
 
     @classmethod
-    def _parse(cls, key, string):
+    def _parse(cls, key, string: str) -> types.Type:
         """Parse a string as a value for the given key.
         """
         if not isinstance(string, str):
@@ -638,7 +664,7 @@ class Model:
 
         return cls._type(key).parse(string)
 
-    def set_parse(self, key, string):
+    def set_parse(self, key, string: str):
         """Set the object's key to a value represented by a string.
         """
         self[key] = self._parse(key, string)
@@ -646,12 +672,21 @@ class Model:
     # Convenient queries.
 
     @classmethod
-    def field_query(cls, field, pattern, query_cls=MatchQuery):
+    def field_query(
+            cls,
+            field,
+            pattern,
+            query_cls: Type[FieldQuery] = MatchQuery,
+    ) -> FieldQuery:
         """Get a `FieldQuery` for this model."""
         return query_cls(field, pattern, field in cls._fields)
 
     @classmethod
-    def all_fields_query(cls, pats, query_cls=MatchQuery):
+    def all_fields_query(
+            cls: Type['Model'],
+            pats: Mapping,
+            query_cls: Type[FieldQuery] = MatchQuery,
+    ):
         """Get a query that matches many fields with different patterns.
 
         `pats` should be a mapping from field names to patterns. The
@@ -670,8 +705,15 @@ class Results:
     constructs LibModel objects that reflect database rows.
     """
 
-    def __init__(self, model_class, rows, db, flex_rows,
-                 query=None, sort=None):
+    def __init__(
+            self,
+            model_class: Type[LibModel],
+            rows: List[Mapping],
+            db: 'Database',
+            flex_rows,
+            query: Optional[FieldQuery] = None,
+            sort=None,
+    ):
         """Create a result set that will construct objects of type
         `model_class`.
 
@@ -703,7 +745,7 @@ class Results:
         # consumed.
         self._objects = []
 
-    def _get_objects(self):
+    def _get_objects(self) -> Generator[Model]:
         """Construct and generate Model objects for they query. The
         objects are returned in the order emitted from the database; no
         slow sort is applied.
@@ -738,7 +780,7 @@ class Results:
                         yield obj
                         break
 
-    def __iter__(self):
+    def __iter__(self) -> Iterable[Model]:
         """Construct and generate Model objects for all matching
         objects, in sorted order.
         """
@@ -751,7 +793,7 @@ class Results:
             # Objects are pre-sorted (i.e., by the database).
             return self._get_objects()
 
-    def _get_indexed_flex_attrs(self):
+    def _get_indexed_flex_attrs(self) -> Mapping:
         """ Index flexible attributes by the entity id they belong to
         """
         flex_values = {}
@@ -763,7 +805,7 @@ class Results:
 
         return flex_values
 
-    def _make_model(self, row, flex_values={}):
+    def _make_model(self, row, flex_values: Dict = {}) -> Model:
         """ Create a Model object for the given row
         """
         cols = dict(row)
@@ -774,7 +816,7 @@ class Results:
         obj = self.model_class._awaken(self.db, values, flex_values)
         return obj
 
-    def __len__(self):
+    def __len__(self) -> int:
         """Get the number of matching objects.
         """
         if not self._rows:
@@ -792,12 +834,12 @@ class Results:
             # A fast query. Just count the rows.
             return self._row_count
 
-    def __nonzero__(self):
+    def __nonzero__(self) -> bool:
         """Does this result contain any objects?
         """
         return self.__bool__()
 
-    def __bool__(self):
+    def __bool__(self) -> bool:
         """Does this result contain any objects?
         """
         return bool(len(self))
@@ -819,7 +861,7 @@ class Results:
         except StopIteration:
             raise IndexError(f'result index {n} out of range')
 
-    def get(self):
+    def get(self) -> Optional[Model]:
         """Return the first matching object, or None if no objects
         match.
         """
@@ -840,10 +882,10 @@ class Transaction:
     current transaction.
     """
 
-    def __init__(self, db):
+    def __init__(self, db: 'Database'):
         self.db = db
 
-    def __enter__(self):
+    def __enter__(self) -> 'Transaction':
         """Begin a transaction. This transaction may be created while
         another is active in a different thread.
         """
@@ -856,7 +898,12 @@ class Transaction:
             self.db._db_lock.acquire()
         return self
 
-    def __exit__(self, exc_type, exc_value, traceback):
+    def __exit__(
+            self,
+            exc_type: Type[Exception],
+            exc_value: Exception,
+            traceback: TracebackType,
+    ):
         """Complete a transaction. This must be the most recently
         entered but not yet exited transaction. If it is the last active
         transaction, the database updates are committed.
@@ -872,14 +919,14 @@ class Transaction:
             self._mutated = False
             self.db._db_lock.release()
 
-    def query(self, statement, subvals=()):
+    def query(self, statement: str, subvals: Iterable = ()) -> List:
         """Execute an SQL statement with substitution values and return
         a list of rows from the database.
         """
         cursor = self.db._connection().execute(statement, subvals)
         return cursor.fetchall()
 
-    def mutate(self, statement, subvals=()):
+    def mutate(self, statement: str, subvals: Iterable = ()) -> Any:
         """Execute an SQL statement with substitution values and return
         the row ID of the last affected row.
         """
@@ -898,7 +945,7 @@ class Transaction:
             self._mutated = True
             return cursor.lastrowid
 
-    def script(self, statements):
+    def script(self, statements: str):
         """Execute a string containing multiple SQL statements."""
         # We don't know whether this mutates, but quite likely it does.
         self._mutated = True
@@ -922,7 +969,7 @@ class Database:
     data is written in a transaction.
     """
 
-    def __init__(self, path, timeout=5.0):
+    def __init__(self, path, timeout: float = 5.0):
         if sqlite3.threadsafety == 0:
             raise RuntimeError(
                 "sqlite3 must be compiled with multi-threading support"
@@ -956,7 +1003,7 @@ class Database:
 
     # Primitive access control: connections and transactions.
 
-    def _connection(self):
+    def _connection(self) -> Connection:
         """Get a SQLite connection object to the underlying database.
         One connection object is created per thread.
         """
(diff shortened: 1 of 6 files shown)


Serene-Arc:
So given that I'd like to try converting beets into using `pathlib`, I thought I'd try typing the path functions that beets uses. Honestly, I don't think it's possible though. Due to Python's ambiguous treatment of bytes and strings, there is a lot of code that doesn't distinguish, or only distinguishes some of the time, despite what it apparently says. 

For pretty much every function, the best that I can do is say it's either a string or a bytestring and leave it at that because anything more is just not possible. If this were statically typed then I honestly think it wouldn't even compile but because of Python's duck-typing system, I would imagine that it just works because it converts when it needs to and doesn't when it doesn't.

I'm happy to try opening another PR with the `pathlib` conversion and typing that as I go. That'll solve both problems at once and given that there's a large test base, it *should* be fine.

wisp3rwind:
> So given that I'd like to try converting beets into using `pathlib`, I thought I'd try typing the path functions that beets uses. Honestly, I don't think it's possible though. Due to Python's ambiguous treatment of bytes and strings, there is a lot of code that doesn't distinguish, or only distinguishes some of the time, despite what it apparently says.

In fact, beets has a convention of using ` bytestring`  for all internal paths, and only convert to anything else (such as `str`) at the boundary to external APIs. As such, I think any deviation from this should in fact be considered a bug. But, I'm also pretty sure that there are in fact such bugs still around. See for example https://github.com/beetbox/beets/pull/3690 which unfortunately I never got around to finishing up.

> I'm happy to try opening another PR with the `pathlib` conversion and typing that as I go. That'll solve both problems at once and given that there's a large test base, it _should_ be fine.

Skipping paths for now sounds fine! I'm pretty sure that our test suite has blind spots in this regard, however. It'll hopefully prevent most regressions, however.

beets/dbcore/db.py
```diff
@@ -71,10 +84,10 @@ def __getitem__(self, key):
         else:
             raise KeyError(key)
 
-    def __iter__(self):
+    def __iter__(self) -> Iterable:
```
wisp3rwind, line 87: Couldn't this be made more specific as `Iterable[str]`?

beets/dbcore/db.py
```diff
@@ -146,12 +159,12 @@ def __delitem__(self, key):
         if key in self.data:
             del self.data[key]
 
-    def keys(self):
+    def keys(self) -> List:
```
wisp3rwind, line 162: `List[str]`?

beets/dbcore/db.py
```diff
@@ -183,12 +196,12 @@ def get(self, key, default=None):
...
             return default
 
-    def __contains__(self, key):
+    def __contains__(self, key) -> bool:
         """Determine whether `key` is an attribute on this object.
         """
         return key in self.keys()
 
-    def __iter__(self):
+    def __iter__(self) -> Iterable:
```
wisp3rwind, line 204: ` Iterable[str]`?

beets/dbcore/db.py
```diff
@@ -167,7 +180,7 @@ def update(self, values):
         for key, value in values.items():
             self[key] = value
 
-    def items(self):
+    def items(self) -> Generator[Tuple, None, None]:
```
wisp3rwind, line 183: According to https://docs.python.org/3/library/typing.html#typing.Generator, this could also be written more concisely as `-> Iterable[Tuple[str, Any]]`.

beets/dbcore/db.py
```diff
@@ -267,14 +280,14 @@ class Model:
...
-    def _getters(cls):
+    def _getters(cls: Type['Model']):
         """Return a mapping from field names to getter functions.
         """
         # We could cache this if it becomes a performance problem to
         # gather the getter mapping every time.
         raise NotImplementedError()
 
-    def _template_funcs(self):
+    def _template_funcs(self) -> NoReturn:
```
wisp3rwind, line 290: Is it really correct to type this as `NoReturn`, when it is expected to be overridden with a function `-> Mapping[str, Callable[[str], str]]`   ? (not sure that I got the callable right, should check actual users of `_template_funcs`)

beets/dbcore/db.py
```diff
@@ -455,19 +473,19 @@ def update(self, values):
         for key, value in values.items():
             self[key] = value
 
-    def items(self):
+    def items(self) -> Generator:
```
wisp3rwind, line 476: `-> Iterator[Tuple[str, Any]]`

beets/dbcore/db.py
```diff
@@ -455,19 +473,19 @@ def update(self, values):
...
             yield key, self[key]
 
-    def __contains__(self, key):
+    def __contains__(self, key) -> bool:
         """Determine whether `key` is an attribute on this object.
         """
         return key in self.keys(computed=True)
 
-    def __iter__(self):
+    def __iter__(self) -> Iterable:
```
wisp3rwind, line 488: ` Iterable[str]`?

beets/dbcore/db.py
```diff
@@ -628,28 +654,37 @@ def evaluate_template(self, template, for_path=False):
     # Parsing.
 
     @classmethod
-    def _parse(cls, key, string):
+    def _parse(cls, key, string: str) -> types.Type:
```
wisp3rwind, line 657: I think `parse` actually return the field type ( `int`, `bool`, etc), so this should probably be `-> Any`.

beets/dbcore/db.py
```diff
@@ -701,7 +743,7 @@ def __init__(self, model_class, rows, db, flex_rows,
         # consumed.
         self._objects = []
 
-    def _get_objects(self):
+    def _get_objects(self) -> Generator[Model, None, None]:
```
wisp3rwind, line 746: `Iterable[Model]` in order to match `__iter__` below.

beets/util/__init__.py
```diff
@@ -355,7 +370,7 @@ def _fsencoding():
     return encoding
 
 
-def bytestring_path(path):
+def bytestring_path(path: Bytestring) -> bytes:
```
wisp3rwind, line 373: Should probably take `Union[Bytestring, str]`)? Or maybe even `Union[bytes, str]`? I don't think we expect any other `Bytestring` subtype here.

beets/util/__init__.py
```diff
@@ -376,10 +391,10 @@ def bytestring_path(path):
         return path.encode('utf-8')
 
 
-PATH_SEP = bytestring_path(os.sep)
+PATH_SEP: bytes = bytestring_path(os.sep)
 
 
-def displayable_path(path, separator='; '):
+def displayable_path(path, separator: str = '; ') -> str:
```
wisp3rwind, line 397: Should be `path: bytes`,  I think. If it's not, that would be a bug in the caller.

beets/util/__init__.py
```diff
@@ -398,7 +413,7 @@ def displayable_path(path, separator='; '):
         return path.decode('utf-8', 'ignore')
 
 
-def syspath(path, prefix=True):
+def syspath(path: Bytestring, prefix: bool = True) -> Bytestring:
```
wisp3rwind, line 416:
```suggestion
def syspath(path: bytes, prefix: bool = True) -> Union[bytes, str]:
```

Output type is platform-dependent, unfortunately. Maybe, there's a way to encode this is in the types?

beets/util/__init__.py
```diff
@@ -432,14 +448,14 @@ def syspath(path, prefix=True):
...
-def samefile(p1, p2):
+def samefile(p1: Bytestring, p2: Bytestring) -> bool:
     """Safer equality for paths."""
     if p1 == p2:
         return True
     return shutil._samefile(syspath(p1), syspath(p2))
 
 
-def remove(path, soft=True):
+def remove(path: AnyStr, soft: bool = True):
```
wisp3rwind, line 458: `path: bytes` (same for the functions below)

beets/util/__init__.py
```diff
@@ -636,7 +657,10 @@ def unique_path(path):
 ]
 
 
-def sanitize_path(path, replacements=None):
+def sanitize_path(
+        path: Bytestring,
+        replacements: Optional[Sequence[Sequence[Pattern | str]]] = None,
+) -> str | bytes:
```
wisp3rwind, line 663: Shouldn't this be `str -> str` ?

beets/util/__init__.py
```diff
@@ -762,12 +798,12 @@ def py3_path(path):
...
 
 
-def str2bool(value):
+def str2bool(value: str) -> bool:
     """Returns a boolean reflecting a human-entered string."""
     return value.lower() in ('yes', '1', 'true', 't', 'y')
 
 
-def as_string(value):
+def as_string(value: Optional[memoryview | bytes]) -> str:
```
wisp3rwind, line 806: Need to check the call sites, but if I remember correctly this can take `Any`, not just `bytes`.

beets/util/__init__.py
```diff
@@ -1056,7 +1092,7 @@ def asciify_path(path, sep_replace):
     return os.sep.join(path_components)
 
 
-def par_map(transform, items):
+def par_map(transform: Callable, items: Sequence):
```
wisp3rwind, line 1095:
```suggestion
def par_map(transform: Callable, items: Iterable):
```

wisp3rwind (commented):
Had a quick look through this (without actually running mypy, though); here are a few comments on where I think the types can be made more precise or are not quite right.

beets/dbcore/db.py
Serene-Arc: IMO it's better to type a generator as a generator, since it can have different uses to a more general iterable. IT's rare but people do call specific generator functions such as `next()`.

beets/util/__init__.py
Serene-Arc: No, there isn't. That isn't how the typing works. The way to do that would be different functions that are typed differently. 

I defined `Bytestring` as equivalent to the union of strings and bytes. Any reason why you changed it? I think the return might be a string no matter what since the conversion happens but that has knock on effects with the rest of the code.

Also `path` can clearly be a string, not bytes, because there's an `isinstance` check in that function that sees whether it's in string or byte form.

beets/util/__init__.py
Serene-Arc: If that's the case then there are flow-in effects. The `path` variable goes through `_legalize_stage()` and `legalize_path()` without being changed, so if `sanitize_path` only accepts strings, then so must those functions to keep it consistent.

I've made these changes in 9453a513

Serene-Arc:
This all makes me think `pathlib` should be implemented ASAP lol

beets/util/__init__.py
Serene-Arc: Quite possibly, these are just the usages that occur in this file. Once it's expanded to the entirety of the project, many typings will likely have to be broadened.

beets/dbcore/db.py
Serene-Arc: I honestly don't know the conventions here. It might be a matter of preference and style whether to mark functions that should be overloaded. Perhaps it will help for IDEs/linters? It's a minor thing, but the typing does exist and this is its explicit purpose in the Python docs.

Serene-Arc:
@wisp3rwind If I were to go about converting to pathlib in another PR, would that be fine vis-a-vis this PR? It would make all of the commits for the util file obsolete but there wouldn't be any confusion and it would be much simpler to type.

beets/dbcore/db.py
wisp3rwind: Yes, I don't know either what the "recommended" approach is. However, in the present case of an abstract method it seems more natural to me to declare the interface that should be implemented, rather than that of the stub method in the super class. I couldn't really find many examples for this; the mypy docs do something similar at https://mypy.readthedocs.io/en/stable/error_code_list.html?highlight=notimplementederror#check-that-call-to-an-abstract-method-via-super-is-valid-safe-super

beets/dbcore/db.py
wisp3rwind: Sure, in fact I'm not at all certain that I got these suggestions for `Iterator`, `Iterable`,  and `Generator` right. But the example that you mention seems to be about the distinction between the former two (`Iterator`  has `__next__` , `Iterable` only has `__iter__`). So, `from my understanding, Iterator`  would in fact be correct here, since not full generality of generators is allowed. `Generator[Tuple[..], None, None]`, seems to be synonymous, and I don't feel strongly about using either of the two expresssions.

beets/util/__init__.py
wisp3rwind: Oh, what happened here is that I confused the custom `Bytestring` with [`typing.ByteString`](https://docs.python.org/3/library/typing.html#typing.ByteString). Seems useful to have an alias here, but we may want to come up with a dfferent name.

> I think the return might be a string no matter what since the conversion happens but that has knock on effects with the rest of the code.

No, the conversion does not always happen, only on Windows.

> Also path can clearly be a string, not bytes, because there's an isinstance check in that function that sees whether it's in string or byte form.

You're right, that is indeed the case. From my understanding, correct usage of that function shouldn't pass `str`,  but it probably makes sense to keep that check as a safeguard for now.

wisp3rwind:
> @wisp3rwind If I were to go about converting to pathlib in another PR, would that be fine vis-a-vis this PR? It would make all of the commits for the util file obsolete but there wouldn't be any confusion and it would be much simpler to type.

I'm not quite sure that I follow your question here: pathlib and the current PR seem to be mostly independent projects, and I don't think these initial typings need to be completely perfect before we merge this (seems to be unrealistic, since we can't really run mypy on the full codebase yet). Converting to pathlib seems like a huge project however, and I suspect that having (partial) type annotations might actually help with the conversion (so, I'd keep the `util` part of this PR). Also, as discussed at https://github.com/beetbox/beets/issues/1409 and https://github.com/beetbox/beets/pull/3690#issuecomment-664074016, its not completely obvious that pathlib can get us rid of all manual conversions (it might, but nobody has really thought about this yet).

Serene-Arc:
I intended to try my hand at the pathlib conversion soon but I mentioned a while back it might be good to try this typing project beforehand to make sure that there were fewer issues. That might still be the case but it seems to have less of an impact.

I've used pathlib extensively and in general I don't think conversions are required. The `Path` object accepts bytes or strings and then can be passed to all of the other standard library functions. Windows and Linux can be treated more or less the same in my other projects once the strings have been sanitised. IIRC the only problems that we had were with malformed Unicode bytes.

beets/util/__init__.py
Serene-Arc: Oops wrong one

beets/util/__init__.py
Serene-Arc: Fixed in 5abe5d54

beets/util/__init__.py
Serene-Arc: Fixed in 4d1dec3c

beets/dbcore/db.py
Serene-Arc: Fixed in e57d66e2

beets/dbcore/db.py
Serene-Arc: Fixed in e57d66e2

beets/dbcore/db.py
Serene-Arc: Fixed in e57d66e2

beets/util/__init__.py
Serene-Arc: This is my defined type so it already is the union. I'll change the name to something less confusing in a coming commit.

Serene-Arc:
I've addressed all of the issues that you've raised but I think there's a lot of work to be done yet. Do you have a direction you'd like me to take? Way I see it, there are two main paths:

1. Start using mypy to firm down the typings in the modules that I've already started
2. Expand the typings project to other modules, keeping the rough ones and using mypy at the end on the project as a whole to resolve conflicting typings

wisp3rwind:
Hey, sorry for the late reply; the time that I can invest into beets its fairly limited these days!

> I intended to try my hand at the pathlib conversion soon but I mentioned a while back it might be good to try this typing project beforehand to make sure that there were fewer issues. That might still be the case but it seems to have less of an impact.
> 
> I've used pathlib extensively and in general I don't think conversions are required. The `Path` object accepts bytes or strings and then can be passed to all of the other standard library functions. Windows and Linux can be treated more or less the same in my other projects once the strings have been sanitised. IIRC the only problems that we had were with malformed Unicode bytes.

One point of friction might be the connection of pathlib to our database model (which always stored bytestrings). But you're probably better qualified to talk about this than me; my own usage of pathlib has been fairly limited and mostly in ASCII-only scenarios. I think the only real constraint should be backwards-compatibility (or, conscious decisions to break it with appropriate warnings to users).


> I've addressed all of the issues that you've raised but I think there's a lot of work to be done yet. Do you have a direction you'd like me to take? Way I see it, there are two main paths:
> 
> 1. Start using mypy to firm down the typings in the modules that I've already started
> 
> 2. Expand the typings project to other modules, keeping the rough ones and using mypy at the end on the project as a whole to resolve conflicting typings

I guess I'd like to see mypy rather sooner than later (I also find its errors useful when adding missing types); but in principle, whatever motivates you!

From the reviewers perspective, I think it would be very helpful to split things into manageable PRs, even that implies that they might not be perfect initially (which is harmless, after all).
In particular, could you move the new commits for the autotag module to a separate PR (relative to this one)? Then, I'd give the dbcore types another quick glance and merge it. Similarly, mypy could be added to the CI in a PR of its own.

Serene-Arc:
Sure I can do that. Mypy is more strict so there will likely have to be changes. Also, there are additional changes to the `dbcore` typing that are becoming clear as I move through modules that reference them, so hopefully that won't cause issues with the branch splitting.

Serene-Arc:
This commit here should be pretty good for merging.

wisp3rwind:
> Sure I can do that. Mypy is more strict so there will likely have to be changes. Also, there are additional changes to the `dbcore` typing that are becoming clear as I move through modules that reference them, so hopefully that won't cause issues with the branch splitting.

Thanks!

I guess we don't need to enforce mypy compliance (yet), but it would be nice to have it available via tox (and maybe even in CI, but without actually failing the entire build, similar to how we test python alphas?). I might actually have a look at that myself.

wisp3rwind:
> > Sure I can do that. Mypy is more strict so there will likely have to be changes. Also, there are additional changes to the `dbcore` typing that are becoming clear as I move through modules that reference them, so hopefully that won't cause issues with the branch splitting.
> 
> Thanks!
> 
> I guess we don't need to enforce mypy compliance (yet), but it would be nice to have it available via tox (and maybe even in CI, but without actually failing the entire build, similar to how we test python alphas?). I might actually have a look at that myself.

About issues with branch splitting: I'm not sure I get what you mean, but wouldn't you avoid any problem by rebasing the autotag one on top of this one?

wisp3rwind:
When this is ready for a final review, can you mark the PR as such?

@sampsyo (and anyone else!): Any concerns on going ahead with typings in beets?

---

One thought that I had when thinking about paths in beets: It might be a good idea to have newtypes for bytes and str here to encode the semantics more clearly, i.e.

```python
InternalPath = NewType("InternalPath", bytes)
SysPath = NewType("SysPath", Union[bytes, str])  # <- or, maybe, define this per-platform?

def syspath(path: InternalPath) -> SysPath:
    ...

def displayable_path(path: InternalPath) -> str:
    ...

# etc
```

(Doesn't have to happen in this PR already, but I'd be happy about any opinions!)

Serene-Arc:
@wisp3rwind I suppose it wouldn't be an issue for the branch splitting if I base it off this one but generally I like to hold to the principle that each PR is a self-contained change to the development or master branch. Basing PRs off of other PRs that have yet to be merged sounds like a recipe for chaos, but if it's allowed then sure, I can do that.

I think that it's ready for final review now, unless I need to make it mypy passable right now?

The internal path type alias can be done, but like I said above, there's no way to type in a platform dependent manner. The way to do that is to have separate functions, there's no way in Python to define the same function parameter to be bytes on windows and strings everywhere else.

beets/dbcore/query.py
```diff
@@ -163,14 +169,18 @@ class StringFieldQuery(FieldQuery):
     """
 
     @classmethod
-    def value_match(cls, pattern, value):
+    def value_match(cls: Type['StringFieldQuery'], pattern: str, value: Any):
```
wisp3rwind, line 172: Let's just drop the type for `cls` here?  I suppose type checkers don't need it, and if we really need to type `cls`, I think we should wait until we drop support for older Python which doesn't have the `Self` type yet. (Same for `cls` annotations below.)

wisp3rwind (changes requested):
Hi @Serene-Arc , I've finally looked through the entire changeset, and left a bunch of inline comments. I think I'd be good to address at least the ones for `util/__init__.py` before merging this.

For `dbcore` my impression is that, after all, thi is not the easiest component of beets to type: There are a lot of abstract base classes, things that seem to require what Rust calls "associated types", and I guess we will quite a few iterations to really nail down the typing there (and potentially, first fix the code to remove some nasty hacks). Thus, I would merge this without going through all the comments that I left.

wisp3rwind:
> @wisp3rwind I suppose it wouldn't be an issue for the branch splitting if I base it off this one but generally I like to hold to the principle that each PR is a self-contained change to the development or master branch. Basing PRs off of other PRs that have yet to be merged sounds like a recipe for chaos, but if it's allowed then sure, I can do that.

Fair enough. However we do it exactly, what I would really like to achieve is to keep these PRs relatively small. This one is already rather large, and reviewing it entirely literally takes hours, and I already looked at some parts quite superficially.

> I think that it's ready for final review now, unless I need to make it mypy passable right now?

No, let's not bother too much with making mypy happy for now.

> The internal path type alias can be done, but like I said above, there's no way to type in a platform dependent manner. The way to do that is to have separate functions, there's no way in Python to define the same function parameter to be bytes on windows and strings everywhere else.

Yeah, I'm not sure about how to do this either, and it's certainly something for a separate PR. What I had in mind is something along the lines of
```python
import sys
if sys.platform == "win32":
    SysPath = NewType("SysPath", str)
else:
    SysPath = NewType("SysPath", bytes)
```
but the question is of course whether type checkers can deal with that. According to https://mypy.readthedocs.io/en/stable/common_issues.html#version-and-platform-checks, it might actually work.

Serene-Arc:
The `dbcore` modules were recommended to me as the 'base' of the system so the typings would be easier to propogate out but in retrospect, not the easiest place to start.

beets/util/__init__.py
```diff
@@ -136,7 +142,7 @@ class MoveOperation(Enum):
     REFLINK_AUTO = 5
 
 
-def normpath(path):
```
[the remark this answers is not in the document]
Serene-Arc: Not that I know of, changed it.

beets/util/__init__.py
```diff
@@ -145,7 +151,7 @@ def normpath(path):
     return bytestring_path(path)
 
 
-def ancestry(path):
```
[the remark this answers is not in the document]
Serene-Arc: `AnyStr` means that it is bytes or a string but in a way that is consistent through the function e.g. if there are multiple instances of variables of type `AnyStr` they are all either bytes or all strings.

For a single instance it is essentially a union when viewed as a single function but mypy might propagate the typing restrictions down the stack, I don't know enough about it to be sure.

beets/util/__init__.py
```diff
@@ -226,14 +237,14 @@ def sorted_walk(path, ignore=(), ignore_hidden=False, logger=None):
...
-def path_as_posix(path):
+def path_as_posix(path: bytes) -> bytes:
     """Return the string representation of the path with forward (/)
     slashes.
     """
     return path.replace(b'\\', b'/')
 
 
-def mkdirall(path):
+def mkdirall(path: str):
```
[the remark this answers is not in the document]
Serene-Arc: This function calls `ancestry()` which has a `str` in the function description example. If that is wrong, then it should be changed to bytes as it's called directly with the path parameter used here.

beets/util/__init__.py
```diff
@@ -686,7 +716,7 @@ def _legalize_stage(path, replacements, length, extension, fragment):
 
     # Encode for the filesystem.
     if not fragment:
-        path = bytestring_path(path)
+        path = bytestring_path(path)  # type: ignore
```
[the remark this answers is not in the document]
Serene-Arc: Hopefully once I get the pathlib PR going this won't be an issue. This was mostly for mypy since it doesn't like when you change the type of named variables, it always throws an error.

beets/dbcore/types.py
```diff
@@ -35,7 +36,7 @@ class Type:
     """The `Query` subclass to be used when querying the field.
     """
 
-    model_type = str
+    model_type: Callable[[Any], str] = str
```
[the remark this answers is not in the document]
Serene-Arc: Right now this is saying that it takes in any parameters and outputs a str, and is defaulting to the builtin function `str()`. Not the clearest syntax.

I don't know if there's a way to do that with a TypeVar since you would have to dynamically bind it to the builtin cast function, seems to me like that would require another if-else block to do easily. 

Plus, for this function, we know that it always returns a string since it's bound to the string cast function, so a TypeVar would actually be incorrect as it's always going to be a string.

beets/dbcore/queryparse.py
```diff
@@ -158,7 +172,13 @@ def construct_query_part(model_cls, prefixes, query_part):
         return out_query
 
 
-def query_from_strings(query_cls, model_cls, prefixes, query_parts):
+# TYPING ERROR
```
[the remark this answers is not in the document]
Serene-Arc: I think that it doesn't fit the `__init__` parameters of every child of the `Query` class. It's submitting a list of queries to the `__init__` of the specified class but a lot of classes require more than one parameter. My IDE is highlighting it as an issue.

I should do more detailed comments.

beets/dbcore/queryparse.py
```diff
@@ -34,8 +38,14 @@
 )
 
 
-def parse_query_part(part, query_classes={}, prefixes={},
-                     default_class=query.SubstringQuery):
+# TYPING ERROR
+# not sure what the query type should be
```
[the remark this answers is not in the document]
Serene-Arc: Apparently no I can't because I have no idea what I was referring to. Seems to have been resolved by other changes or is somewhere in the usage. I'll remove the comments.

beets/dbcore/types.py
```diff
@@ -125,7 +131,7 @@ class Integer(Type):
     query = query.NumericQuery
     model_type = int
 
-    def normalize(self, value):
+    def normalize(self, value: str) -> Union[int, str]:
```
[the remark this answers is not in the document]
Serene-Arc: We could define a type alias per class?

wisp3rwind:
> The `dbcore` modules were recommended to me as the 'base' of the system so the typings would be easier to propogate out but in retrospect, not the easiest place to start.

Sorry for that! Thanks a lot for sticking around anyway!

beets/dbcore/queryparse.py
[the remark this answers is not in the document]
wisp3rwind: Hmm, I'm not quite following here. I'll mark this as resolved for now, and will have a look when this is merged.

beets/dbcore/types.py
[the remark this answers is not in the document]
wisp3rwind: Oh, ok, there seem to be two approaches how to model the types here:

- what you did, understand `model_type` as the function which parses `Any` -> `<the target type>`, which works since classes `int`, `float`, etc are also callable objects which return an instance of that class
- what I had in mind: `model_type` is the class that values of the given type should have, which also happens to be a callable

However, I'm pretty sure that the output type is not actually `str`, but at this point you cannot be more precise than `Any`. For example, consider when `model_type = int`.

The best way to express all of this might be using generic, cf. https://stackoverflow.com/questions/74885777/how-to-implement-associated-types-in-python-mypy-or-what-to-do-when-wanting-su

Let's not do that now I'd say, but go for the `Callable` approach.

beets/dbcore/types.py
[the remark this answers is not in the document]
wisp3rwind: Seems like the solution is to make `Type` generic over `model_type` and `null_type` cf. https://stackoverflow.com/questions/74885777/how-to-implement-associated-types-in-python-mypy-or-what-to-do-when-wanting-su

How much time would you still like to invest into this project? If you like, go ahead, otherwise, I'd also offer to merge this quickly, and help with improving it myself afterwards.

beets/util/__init__.py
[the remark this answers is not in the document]
wisp3rwind: Exactly; what I meant to suggest is that this should be `def ancestry(path: AnyStr) -> List[AnyStr]:`

beets/util/__init__.py
```diff
@@ -376,10 +391,10 @@ def bytestring_path(path):
         return path.encode('utf-8')
 
 
-PATH_SEP = bytestring_path(os.sep)
+PATH_SEP: bytes = bytestring_path(os.sep)
 
 
-def displayable_path(path, separator='; '):
+def displayable_path(path: bytes, separator: str = '; ') -> str:
```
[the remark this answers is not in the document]
wisp3rwind: I think you accidentally added the comment to `bytestring_path`?

beets/util/__init__.py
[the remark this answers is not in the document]
wisp3rwind: mypy complains for good reason, changing the type in this way really seems like a recipe for disaster. I'm not sure that pathlib magically makes this disappear, I suspect that the callers of this function might still need a different type depending on whether dealing with path fragments or an entire path.

wisp3rwind (commented):
I left a few more small comments, what do you think?

beets/dbcore/types.py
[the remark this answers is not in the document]
Serene-Arc: So leave as is?

beets/dbcore/types.py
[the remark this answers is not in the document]
Serene-Arc: I'm happy to invest more time into this but maybe in another PR? You seem to have a good idea of what you'd like the system to be for these classes' typing, and it would also mean I can work on the pathlib PR more easily.

Serene-Arc:
I've left comments on them but more broadly, if you'd like to merge what there is now, I'd be fine with that. As I expand the typings across modules, there's going to be at least some changes and merging nowish would make that easier, plus allow a bit more collaboration on the different parts.

Sorry that this turned into a rather mammoth and wide-reaching PR, there's a lot to type.

JOJ0:
Without having an idea about the details and what's left but it seems like that you both agreed that this could be merged now and the "typing project" will be tackled further in subsequent PR's. Should we just merge it and if you think it's required leave a note in form of an issue of what's left to be "typed"?

Serene-Arc:
@JOJ0 yes I think that this can be merged. There isn't likely to be any actual issues since it's all typings and mypy hasn't been integrated into the checks yet.

Serene-Arc:
I need to resolve the conflicts but my brain isn't working right now and I keep messing up, so I'll get to it soon.

JOJ0:
LOL yeah I know that issue. Resolving conflicts sometimes requires more cpu-power than it deserves!

wisp3rwind:
Thanks for wrapping this up :tada:

Serene-Arc:
Finally got the tests working and the conflicts resolved!

PR #4496, opened 2022-09-24
Add a `import.ignored_alias_types` option to ignore alias types.

djl:
Sometimes a user may want to use an artist's locale-specific alias but *not* want to use their legal name, for example.

## Description

A good example of this would be the artist [Qeight](https://musicbrainz.org/artist/9888f61e-3ec8-4dcd-ae11-3115da898f0a). On MusicBrainz, their `en` locale is set to their legal name. While technically correct for the data in MusicBrainz, this probably isn't what a user expects when using beets.

(No tests yet but I can add them if this PR is going to be merged)

## To Do

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -117,11 +117,20 @@ def _preferred_alias(aliases):
     # Only consider aliases that have locales set.
     aliases = [a for a in aliases if 'locale' in a]
 
+    # Get any ignored alias types and lower case them to prevent case issues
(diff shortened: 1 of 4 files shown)


sampsyo:
Wow; this is a great idea and looks like an ideal implementation! I've resolved some lingering CI problems recently and updated the branch, so let's see if things go green.

sampsyo:
Indeed it has!  💚 🎉

PR #4504, opened 2022-09-30
Fix beets#4472 - set 'max_bitrate' default param to none

### LINKED ISSUE #4472: Convert plugin's max_bitrate does not match documentation
### Problem

Convert plugin's parameter ```max_bitrate``` has default value of ```500``` (see line 139 in ```beetsplug/convert.py```), but in docs it's ```none```.
This results in unexpected copying instead of transcoding.
I'd expect ```none``` to be a reasonable default.

### Setup

Irrelevant

Tomasito665:
## Description

Fixes #4472.

## To Do

- [ ] Documentation. N/A because the documentation was already correct.
- [X] Changelog.
- [X] Tests.

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -26,7 +26,7 @@ import logging
 
 from beets import ui, util, plugins, config
 from beets.plugins import BeetsPlugin
-from confuse import ConfigTypeError
+from confuse import ConfigTypeError, Optional
 from beets import art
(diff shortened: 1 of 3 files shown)


sampsyo:
Nice; this looks perfect! Thank you for the tidy fix and the test!! Really nice work here.

I'll merge this once the CI goes green.

PR #4505, opened 2022-10-01
Adapt tests to latest Unidecode version

sampsyo:
Unidecode 1.3.5 (a yanked PyPI version) [changed the behavior](https://github.com/avian2/unidecode/blob/414199246e4871669896756279e3460ff0364839/ChangeLog#L5) for some specific characters:

> Remove trailing space in replacements for vulgar fractions.

As luck would have it, our tests used the 1/2 character specifically to test the behavior when these characters decoded to contain slashes. We now pin a sufficiently recent version of Unidecode and adapt the tests to match the new behavior.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -85,7 +85,7 @@ setup(
     },
 
     install_requires=[
-        'unidecode',
+        'unidecode>=1.3.6',
         'musicbrainzngs>=0.4',
         'pyyaml',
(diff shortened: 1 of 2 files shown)


sampsyo:
This is just one of two recent intermittent test failures that has recently cropped up… so, still red, but getting there.

PR #4506, opened 2022-10-01
replaygain: Fix error handling for parallel runs

sampsyo:
This is a second fix to errors that have been cropping up in CI lately. Put this one into the bucket labeled "I don't know why this ever worked," but it's causing problems only on the Windows CI builds.

The parallelism strategy in #3478, in retrospect, used a pretty funky way to deal with exceptions in the asynchronous work—since `apply_async` has an `error_callback` parameter that's meant for exactly this. The problem is that the wrapped function would correctly log the exception *and then return `None`*, confusing any downstream code. Instead of just adding `None`-awareness to the callback, let's just avoid running the callback altogether in the case of an error.

### DIFF
--- a/beetsplug/replaygain.py
+++ b/beetsplug/replaygain.py
@@ -49,13 +49,14 @@ class FatalGstreamerPluginReplayGainError(FatalReplayGainError):
     loading the required plugins."""
 
 
-def call(args, **kwargs):
+def call(args, log, **kwargs):
     """Execute the command and return its output or raise a
     ReplayGainError on failure.
     """
     try:
         return command_output(args, **kwargs)
     except subprocess.CalledProcessError as e:
+        log.debug(e.output.decode('utf8', 'ignore'))
         raise ReplayGainError(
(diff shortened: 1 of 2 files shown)


sampsyo:
This fixes the obvious crash in the error handling. However, there is still a problem with invoking `ffmpeg` on Unicode filenames on Windows. I'll work on that in another PR—but if that effort fails (it might!), then we can always go back to giving up entirely in situations like this. (Unicode stuff on Windows has always been a mess.)

wisp3rwind:
> This fixes the obvious crash in the error handling. However, there is still a problem with invoking `ffmpeg` on Unicode filenames on Windows. I'll work on that in another PR—but if that effort fails (it might!), then we can always go back to giving up entirely in situations like this. (Unicode stuff on Windows has always been a mess.)

:+1:

A few days ago, I've also realized this when testing #4374, which was also plagued by mysterious CI failures that I couldn't trace back to any of the changes there. However, your fix is much cleaner than the quick workaround I came up with!

PR #4507, opened 2022-10-01
Use Python's standard filesystem encoding for command-line arguments

sampsyo:
This is very much an experiment, but it's an attempt to address persistent issues with invoking command-line tools with non-ASCII arguments. For example, the `replaygain` plugin wants to invoke `ffmpeg` on filenames that might contain special characters. Doing this in a cross-platform way has always been extremely fraught—it's hard to get it to work right simultaneously on Unix (where filenames are plain bytes that need not obey *any* specific encoding) and on Windows (where filenames are weird, but they are closer to being Unicode strings).

The experiment here is to centralize on the Python standard library's implicit recommendations for this: `fsencode` and `fsdecode`. These use surrogate escaping (at least in most configurations??), so they probably shouldn't crash. The `subprocess` module also apparently assumes this is the right encoding for CLI argument stuff, at least on Windows:
https://github.com/python/cpython/blob/54bbb5e3363ef3634f1fd7521ba3f3c42c81a331/Lib/subprocess.py#L561

…so hopefully using this encoding everywhere should make things better on Windows and no worse on Unix.

The hope is to address some CI failures that come from the `replaygain` invoking `ffmpeg` on Windows.

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -1092,21 +1092,3 @@ def lazy_property(func):
         return value
 
     return wrapper
-
-
-def decode_commandline_path(path):
-    """Prepare a path for substitution into commandline template.
-
-    On Python 3, we need to construct the subprocess commands to invoke as a
-    Unicode string. On Unix, this is a little unfortunate---the OS is
-    expecting bytes---so we use surrogate escaping and decode with the
-    argument encoding, which is the same encoding that will then be
-    *reversed* to recover the same bytes before invoking the OS. On
-    Windows, we want to preserve the Unicode filename "as is."
-    """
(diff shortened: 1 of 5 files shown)


sampsyo:
Incredibly, this actually seems to have addressed the `ffmpeg`-related errors on Windows in CI! I'm going to give this a bit of a "cooling-off" period and then merge it so we can get master back to green.

sampsyo:
OK, this does seem to work—at least for our tests—and I haven't thought of a good reason it should be wrong over the last couple of days. So I'm going to merge this one, with considerable trepidation, and we'll see if anything breaks in the wild!

PR #4509, opened 2022-10-03
ftintitle overhaul

dvcky:
### Reworking ftintitle for better coverage of weirder cases
There are many weird cases in which our current method (regular expressions) really struggle to identify. This fork intends to improve upon those issues and give the user more accessibility with the program.
## To Do

- [x] **Basic Implementation** - get it working _alongside_ current code
- [ ] **Cover more music platforms** - create a way for platforms other than MusicBrainz to work with ftintitle
- [ ] **Other specific cases?** - account for other issues if found and agreed are common enough to need implementation

### DIFF
--- a/docs/plugins/index.rst
+++ b/docs/plugins/index.rst
@@ -324,7 +324,7 @@ Here are a few of the plugins written by the beets community:
 * `beets-originquery`_ augments MusicBrainz queries with locally-sourced data
   to improve autotagger results.
 
-* `beets-popularity`_ fetches popularity values from Spotify.
+* `beets-popularity`_ fetches popularity values from Deezer.
 
 * `beets-setlister`_ generate playlists from the setlists of a given artist.
 


sampsyo:
Thanks for getting this started! I admit that I don't yet fully understand how this works exactly yet, but one high-level question seems relevant: in what situations are you worried about this interfering with existing workflows (if it were enabled by default)? It seems like we should aim for a policy that is "one-size-fits-all," i.e., that doesn't behave too weirdly for normal releases but also works most of the time for VA releases.

dvcky:
Hi @sampsyo! As far as I've been able to test, I don't think it should have issues with existing workflows, but because I don't necessarily know all possible edge cases, I thought it would be good to keep disabled until it's been tested on several collections of music and approved to generally work from several users. If we want to make it enabled by default though, we certainly can!

As for how it functions, here is a pretty good run-down:

1. Check if `albumartist` is "Various Artists". If it is, assign `albumartist` to `artist` of the specific song without the feature.

This is probably the most difficult part to explain. Like I mentioned in the pull request, **ftintitle determines where the feature artist is by looking at the `albumartist` as being a "substring" of `artist`**

|albumartist|artist|output (artist, feat_part)|
|-|-|-|
|Foo|Foo feat. Bar|Foo, feat. Bar|
|Various Artists|Foo feat. Bar|Foo feat. Bar, _|

The first row shown in the table above is an example of an album that works as expected. Because "Foo" is a substring of "Foo feat. Bar", ftintitle knows there is a feature and will split it. The second row is a common example of what would happen without my code. Because "Various Artists" is not a substring of "Foo feat. Bar", **ftintitle assumes the artist is entirely separate, and won't do anything with it.**

**TL;DR: my code preemptively changes `albumartist` for each individual song in that album, so that it acts like the first row in the table.**

2. Pass the artist value through `update_metadata()`. While it won't necessarily be used, there is no other way to access it in update_metadata() if needed, so this is a quick and easy way to pass the value.

That's it! Now that I have explained it, you might be realizing that how this plugin works at it's core has some issues. For example, **if it wasn't a compilation album (meaning albumartist is "Various Artists"), but still contains songs with different main artists, this won't work.** While rare, it does exist. (Second link example provided by @Verrus [here](https://github.com/beetbox/beets/pull/1011#issuecomment-58928114)) In that case, the track with the artist "Jay-Z feat. Beyonce" would be skipped, even with my code.

Because of this, I suggest that we look into refactoring the entire plugin (hence why I have the task in the To-Do list). Until then, this smaller-scale addition exists with the intent of working with the current code and changing as little as possible to promote stability and compatibility.

dvcky:
I am realizing now the larger scale of this project. Right now, we have to use regular expressions to split the artist from the feature, but I think we will find that this becomes quite difficult with more complicated scenarios. (example shown [here](https://beta.musicbrainz.org/recording/180298db-82f9-4abe-b7f1-8f961b70eb77)) If possible, it might be better for us to store the main artist in artist, but also have a new array in Item that stores features (with each featured artist being an entry in the array) separately. This way we know where the artist starts and ends without having to account for all complicated and specific scenarios. I realize that this might be a semi-daunting task though, and might conflict with lots of code, so I'd love to hear an opinion on it.

It might also be worth storing the feature "relation" in Item. What I mean by that is that some features aren't just "features". [Here](https://beta.musicbrainz.org/recording/f926de9c-7f43-4165-a20a-96e4f137eff8) is an example where the feature is considered a duet.

sampsyo:
Sounds interesting! For an initial "minimum viable product" here, I think we should focus on what we can do without changing the database schema. That likely means yielding good results _for very restricted scenarios_—we don't have to handle _every_ possible case; we can just handle one or two cases for now (but do it right when we do it). Would that make sense as a place to start?

dvcky:
I'm 100% on board there, and I think I have some ideas for how to implement it without changing the database schema. It would be a bit redundant, but instead of grabbing the data from Item, we could re-fetch the track data using `item.mb_trackid` and store it inside ftintitle itself. That way we have the benefits of getting the data I wanted to implement my code, without the downside of having to mess around in Item.

**This is my current example script to showcase how I want it to function: [ftintitle-pseudo.zip](https://github.com/beetbox/beets/files/9765708/ftintitle-pseudo.zip)** _(this runs on it's own, independent from beets. requires Python and musicbrainzngs to work)_

If people could look it over and tell me what they like and dislike about how it behaves, that would be great! Otherwise, I'll start working on it!

dvcky:
I got a bit distracted, but I finally got some ground on my code for the overhauled ftintitle! 

### Changes:

- Removed my previous **enable_for_va_compilations** with intent of replacing with this better method
- Changed default for format variable to a more common title standard (added parenthesis)
- _added_ **use_new**: set by default to `False`. When `True`, this will enable my new method for ftintitle. Reasoning for why it is set to `False` by default at the bottom*
- _added_ **collab_cases**: an array of feature types that should indicate a collaboration work rather than a feature. In other words, which feature types should be ignored by ftintitle.

### ft_in_title_new
The idea behind this is that there are a lot of cases that make the old implementation of ftintitle really hard to work with or improve upon. So, instead of using regular expressions to try and figure out where artists start and end in a string, why not use the data MusicBrainz gives us? When we do this, we can hope to get a much improved, more all-encompassing performance from ftintitle. `ft_in_title_new` aims to do just that! Here's how it works:

1. Fetch MusicBrainz data and distribute the important stuff into arrays stored in the method.
2. Build a title and artist from that data in a sensible way.
3. Apply newly created title and artist to track and metadata. Profit!

*This is at a point in development where it is very usable and seems to cover pretty much every case I can throw at it. Which is great!..for people using MusicBrainz. While that will be a large percentage of the users, I know that some may prefer Discogs or some other music database, which may have slightly different data. In those cases, the original method (while covering a lot less cases and being a lot less thorough) might be better for the sake of consistency with their preferred database. I would like to implement these other platforms, **but I am hoping that my original idea of storing the artist and feature type arrays in Item itself will be accepted so that we can simplify things to an immense degree.** If not, that is understandable, and I can always implement every platform individually in the plugin itself, even if that is redundant and less efficient.

sampsyo:
Interesting! That is a very different approach—one that is specific to MusicBrainz (and wouldn't work if you already had metadata on your files, or were using another metadata source). Maybe it's worth considering this as a new, separate, coexisting plugin?

dvcky:
The more I think about that, the better it sounds! Great idea! That should be a fairly quick thing, I'll get to working on that later today. Does `ftintitle_mb` sound like a proper title for this as a separate plugin?

sampsyo:
Sounds reasonable enough to me. But I suppose I have one very high-level question: is there any sense in trying to make this work _on import_ rather than in a post-processing pass? That is, maybe the plugin should change the way we translate MusicBrainz data into beets metadata _on the fly_ rather than fetching data from MusicBrainz again.

That would, of course, require brand-new plugin hooks we don't currently have, but it may be worth investigating anyway.

dvcky:
I think that would be great, and as I mentioned earlier a solution like that would eliminate redundancy of fetching data from MusicBrainz twice. I'll be perfectly honest, I don't entirely understand the import process all that well, so any suggestions on where to start with that would be greatly appreciated!

sampsyo:
Sure! The main place I _think_ is relevant is in the `beets.autotag.mb` module, which is responsible for extracting all the data we get from the MusicBrainz API into a flat set of beets fields:
https://github.com/beetbox/beets/blob/master/beets/autotag/mb.py

There is also a chance that `apply_metadata` may be relevant:
https://github.com/beetbox/beets/blob/bcfc86df04f247e5a812e4940adfd7aeedb6b3b4/beets/autotag/__init__.py#L98

That's the function that takes the flattened data out of an `AlbumInfo` object and applies it directly to database fields.

PR #4513, opened 2022-10-04
Allow `mb_track_extract` to have the final say for all fields

### LINKED ISSUE #4512: `mb_track_extract`'s changes overwritten by `album_info`
### Background
I've been working on a beets plugin to adjust tags and paths of imported music to my preferences. I wasn't particularly happy with how "ftintitle" works, given that beets itself renders the structured data down into a string, and the string is then post-processed by that plugin. I had a go at implementing an alternative, using the `mb_track_extract` and `mb_album_extract` hooks, which render _from the source data_ the desired output for multiple artist credits.

### Problem
While the hooks work mostly as expected...

(at the end of processing a track, calls a hook and merges the return value into the TrackInfo object)
https://github.com/beetbox/beets/blob/cfb32d9bc5450982a9a9a513cedf5ea553ecc8d2/beets/autotag/mb.py#L279-L281

(at the end of processing an album, calls a hook and merges the return value into the AlbumInfo object)
https://github.com/beetbox/beets/blob/cfb32d9bc5450982a9a9a513cedf5ea553ecc8d2/beets/autotag/mb.py#L478-L480

...There is an additional case where the album_info function overwrites the track_info data with "reprocessed" information based on the album's `tracks` results:
https://github.com/beetbox/beets/blob/cfb32d9bc5450982a9a9a513cedf5ea553ecc8d2/beets/autotag/mb.py#L354-L390

The result is that, if an `mb_track_extract` hook updates a track's title or artist info, that info is dicarded when importing in album mode because the `album_info` call explicitly overwrites it.

I intended to send a PR with a fix (the solution here could take multiple forms, but the simplest is to call the hook again before `track_infos.append`), but I was a little surprised that I couldn't find any existing tests around the hook behavior of the musicbrainz import hooks. I don't have a good enough understanding of the codebase to write some from scratch, so I'm hoping that somebody can guide me:

How can I mock out the appropriate data and processes so that I can write tests around correct behavior of hooks being called when importing albums and/or tracks? Which file is most appropriate, and what mechanisms must I mock or satisfy to accomplish this task?

myndzi:
## Description
In the source, the `mb.album_info` function calls track_info for each recording in a release... but then overrides certain values such as `title`, `artist`, and `length`. This unexpectedly causes hooks for `mb_track_extract` that return values for these fields to not have those values honored.

This PR defers calling the hook when `mb.track_info` is called from `mb.album_info` until after these album-context modifications have been made, thus allowing plugins to have the final say on the values for any fields.

Fixes #4512.  

## To Do

- [n/a] Documentation. (If you've add a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -395,7 +395,8 @@ def album_info(release):
                 ti.length = int(track['length']) / (1000.0)
 
             # Supplementary fields provided by plugins
-            extra_trackdatas = plugins.send('mb_track_extract', data=track)
+            extra_trackdatas = plugins.send(
+                'mb_track_extract', data=track['recording'])
             for extra_trackdata in extra_trackdatas:
                 ti.update(extra_trackdata)
 
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -203,7 +203,7 @@ def _flatten_artist_credit(credit):
 
 
 def track_info(recording, index=None, medium=None, medium_index=None,
-               medium_total=None):
+               medium_total=None, extract_hook=True):
     """Translates a MusicBrainz recording result dictionary into a beets
     ``TrackInfo`` object. Three parameters are optional and are used
     only for tracks that appear on releases (non-singletons): ``index``,
@@ -275,10 +275,16 @@ def track_info(recording, index=None, medium=None, medium_index=None,
     if arranger:
         info.arranger = ', '.join(arranger)
 
-    # Supplementary fields provided by plugins
-    extra_trackdatas = plugins.send('mb_track_extract', data=recording)
-    for extra_trackdata in extra_trackdatas:
-        info.update(extra_trackdata)
+    # When calling track_info from album_info, we must defer this hook
+    # since album_info performs a few extra assignments. Therefore,
+    # the argument extract_hook=False can be used to avoid processing
+    # the hook. For track-mode imports, the default (extract_hook=True)
+    # should be used.
+    if extract_hook:
+        # Supplementary fields provided by plugins
+        extra_trackdatas = plugins.send('mb_track_extract', data=recording)
+        for extra_trackdata in extra_trackdatas:
+            info.update(extra_trackdata)
 
     info.decode()
     return info
@@ -370,6 +376,7 @@ def album_info(release):
                 int(medium['position']),
                 int(track['position']),
                 track_count,
+                extract_hook=False,
             )
             ti.release_track_id = track['id']
             ti.disctitle = disctitle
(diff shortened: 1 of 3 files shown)


myndzi:
(Integration testing with my plugin revealed an oversight -- I passed `data=track` to the hook in `album_info` which didn't match the `track['recording']` that gets passed to `track_info` and from there to `mb_track_extract`. I updated the tests to be self-referential to their own data to catch this sort of thing, and fixed it)

sampsyo:
_Super_ sorry for letting this PR languish, since it is clearly almost ready. One thing I would love to understand more deeply, but don't quite yet, is why we need a `extract_hook` flag to the `track_info` function. I understand from the comment is that there is a problem with "interference" between `track_info` and `album_info`, but I don't quite grok what that interference consists of… I would be interested in thinking through this a bit so we can simplify the interface.

myndzi:
@sampsyo Apologies, I was checking occasionally on the issue but not the PR. I didn't notice you'd asked a question :( 

> Super sorry for letting this PR languish, since it is clearly almost ready. One thing I would love to understand more deeply, but don't quite yet, is why we need a extract_hook flag to the track_info function. I understand from the comment is that there is a problem with "interference" between track_info and album_info, but I don't quite grok what that interference consists of… I would be interested in thinking through this a bit so we can simplify the interface.

I honestly don't remember anything about it, I'd given up on this PR. I believe this comment in the test offers some additional context:
https://github.com/beetbox/beets/pull/4513/files#diff-a7c72f70798fb60fdc13080e14bdf135969e77be5c2b4a639124e8ee8ea1393aR502-R505

and here:
https://github.com/beetbox/beets/pull/4513/files#diff-a7c72f70798fb60fdc13080e14bdf135969e77be5c2b4a639124e8ee8ea1393aR531-R532

By my vague recollection, when processing a full recording, the function that handles that calls the function that processes a track. But then, after that function returns, it sets values on the tracks. So, if it goes album -> track -> hook (modify track) -> back to album, album can overwrite the result of the track-processing hook.

myndzi:
https://github.com/beetbox/beets/blob/57b5642b7cb6cacc54c8b0d619724acd19cb09df/beets/autotag/mb.py#L373-L384

This is probably what I was referring to. The code assigns the result of "track_info(...)" to "ti", then sets properties on "ti". The diff is confusing to read because Python, but the changes are actually to separate functions. The code here:
https://github.com/beetbox/beets/pull/4513/files#diff-ff51bca3f15a32bf8652a69e127b582652187bae2fd2fd7a1ccb58d0feeb97b2R283

is in the `track_info` function, and calls the hook to allow it to modify the track results.

The code here:
https://github.com/beetbox/beets/pull/4513/files#diff-ff51bca3f15a32bf8652a69e127b582652187bae2fd2fd7a1ccb58d0feeb97b2R379

is in the **album** info function, and tells the track info function to _not_ run the hooks, since the result of the hook might be overwritten. Then, after performing its normal work, it separately calls the hook for each track (instead of allowing track_data to call the hook):

https://github.com/beetbox/beets/pull/4513/files#diff-ff51bca3f15a32bf8652a69e127b582652187bae2fd2fd7a1ccb58d0feeb97b2R397-R401

arsaboo:
It looks like this PR was close to getting merged...may be we keep it open for now.

snejus:
@myndzi in case you haven't given up there 💪🏼 we can re-open, but I'm closing it for now

myndzi:
Who knows what the codebase is like at this point. I could probably touch up anything needed, but nobody's ever actually told me what is needed...

PR #4521, opened 2022-10-14
Fix recent bug that would not transcode when format was wrong

aereaux:
## Description

Bug introduced in c6d623241b578d1c409affaa2cef27c4f27cb02c, should_transcode would return False even when the format was different.

## To Do

### DIFF
--- a/docs/plugins/convert.rst
+++ b/docs/plugins/convert.rst
@@ -88,10 +88,11 @@ file. The available options are:
 - **embed**: Embed album art in converted items. Default: ``yes``.
 - **id3v23**: Can be used to override the global ``id3v23`` option. Default:
   ``inherit``.
-- **max_bitrate**: All lossy files with a higher bitrate will be
-  transcoded and those with a lower bitrate will simply be copied. Note that
-  this does not guarantee that all converted files will have a lower
-  bitrate---that depends on the encoder and its configuration.
+- **max_bitrate**: By default, the plugin does not transcode files that are
+  already in the destination format. This option instead also transcodes files
+  with high bitrates, even if they are already in the same format as the
+  output.  Note that this does not guarantee that all converted files will have
+  a lower bitrate---that depends on the encoder and its configuration.
   Default: none.
 - **no_convert**: Does not transcode items matching provided query string
   (see :doc:`/reference/query`). (i.e. ``format:AAC, format:WMA`` or
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -102,10 +102,9 @@ def should_transcode(item, fmt):
             not (item.format.lower() in LOSSLESS_FORMATS):
         return False
     maxbr = config['convert']['max_bitrate'].get(Optional(int))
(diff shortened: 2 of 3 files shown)


aereaux:
Actually, looks like this is a little more complicated, not entirely sure what the logic should be here.  The problem I'm getting is that my setup used to be to convert from flac to opus, and that no longer works.

sampsyo:
Argh; that's a thorny problem… maybe throwing in some debugging `print` statements would help work out why the transcoding isn't happening in your specific case?

aereaux:
Yeah, it's not working because my config has the default max_bitrate (None).  This means that with the current logic it never transcodes because I have a flac and I want opus, so it never gets to the last line in the function that checks if the formats are different.  What is the max_bitrate option supposed to do?

sampsyo:
Ah, yeah, that does seem wrong. `max_bitrate` is supposed to do this:

1. If it's None, then do nothing. (The logic only depends on formats, etc.)
2. Otherwise, then transcode files *even when the format would not change* if their bitrate is above the given threshold.

It looks like 2 may be working, but 1 (the default behavior!) is not… does that seem right to you?

aereaux:
Yeah, that's the logic I'd expect, and you're right about where the problem I'm experiencing is.  This PR should fix up the code to match, I just need to fix up the tests.  I should be able to get to that later today or tomorrow.

Thanks!

sampsyo:
Sounds great!

aereaux:
OK, I've fixed up the tests and clarified the documentation.  Hopefully this matches what it should be doing now!

docs/plugins/convert.rst
```diff
@@ -88,10 +88,10 @@ file. The available options are:
 - **embed**: Embed album art in converted items. Default: ``yes``.
 - **id3v23**: Can be used to override the global ``id3v23`` option. Default:
   ``inherit``.
-- **max_bitrate**: All lossy files with a higher bitrate will be
-  transcoded and those with a lower bitrate will simply be copied. Note that
-  this does not guarantee that all converted files will have a lower
-  bitrate---that depends on the encoder and its configuration.
+- **max_bitrate**: If the convert specifies the same format that the file is
+  already in and the existing bitrate is higher than this option, convert
+  anyways. Note that this does not guarantee that all converted files will have
```
sampsyo, line 93: Perhaps a slightly more straightforward phrasing:

> By default, the plugin does not transcode files that are already in the destination format. This option instead also transcodes files with high bitrates, even if they are already in the same format as the output.

sampsyo (commented):
Looks like the right logic to me!! The code style CI is complaining about a couple of too-long lines; then, we should be good to go.

docs/plugins/convert.rst
aereaux: :+1: Much better phrasing than mine.

aereaux:
Great, I abbreviated the test names a bit to fit.

sampsyo:
Perfect! Thank you!!

PR #4522, opened 2022-10-17
Refactor `beets/dbcore` files with exception module

iamibi:
## Description

This pull request is for restructuring the `beets/dbcore` exceptions into its own independent module. This allows a centrally holding pattern for an exception module. Created a file called `beets/dbcore/dbcore_exceptions.py` to cater all the exceptions related to `beets/dbcore`. Additionally, refactored the relevant locations where the exceptions were being used and corrected the comments for the classes which said otherwise.

## To Do


sampsyo:
I don't have any particular objections here! Any chance I could ask how this came up for you? I know this is a common thing to do, but it would be useful/entertaining to know how this reorganization affected you and your use case in particular.

I'd also prefer a simpler name `beets.dbcore.exceptions` to `beets.dbcore.dbcore_exceptions`, if that seems sensible.

iamibi:
> I don't have any particular objections here! Any chance I could ask how this came up for you? I know this is a common thing to do, but it would be useful/entertaining to know how this reorganization affected you and your use case in particular.
> 
> I'd also prefer a simpler name `beets.dbcore.exceptions` to `beets.dbcore.dbcore_exceptions`, if that seems sensible.

Hey @sampsyo, the usual way to organize a project that I have observed is that we try to move all the exceptions in a single place, settings in a single place (maybe they are read from a config file or env values?) and utility methods. This is because these modules are mostly required in different places by the application and I feel that a logical separation of the common modules make sense. I plan to refactor other modules which fit this criteria if you feel that is the right path to go.

As for the naming convention, sure, that works. The long refactor plan that I have essentially takes into account that we bring all these exceptions into a single place. For now, I am good with `beets.db_core.exceptions` as you suggested.

iamibi:
@sampsyo Would you mind adding the hacktoberfest label for this one? Totally optional, but would be appreciated.

sampsyo:
Indeed, I do understand the _general_ principle (and have used `exceptions` modules in the past), but what I was asking here was what your specific need for the reorganization was. I guess I don't have a strong objection to moving things around, but I'm also not terribly enthusiastic about change for the sake of change if the current setup is not causing any problems…

wisp3rwind:
Closing this; I don't think we're convinced that this change is useful in isolation (except for collecting Hacktoberfest points.)

PR #4525, opened 2022-10-20
Corrects popularity service from spotify to deezer to match plugin

phillprice:
## Description

Corrects popularity service from spotify to deezer to match plugin


sampsyo:
Excellent; thank you!!

arsaboo:
The `beets-popularity` [plugin](https://github.com/abba23/beets-popularity) is archived. May be we should just remove it.

PR #4526, opened 2022-10-21
updated CONTRIBUTING.rst with string literals guidelines

mvkwong:
This adds string literals guidelines from [the wiki](https://github.com/beetbox/beets/wiki/String-literals) to `CONTRIBUTING.rst` as suggested in [#3283](https://github.com/beetbox/beets/issues/3283).


sampsyo:
Aha, that's an interesting idea… the thing is, that wiki page has quite recently become out of date, now that beets is a Python 3-only project. Therefore, for example, there is never a need for `u'...'` literals, and there is not really any such thing as "native strings" anymore. What do you think about revising this for the Python 3 era?

PR #4531, opened 2022-10-25
Added Code of Conduct Documentation

judylee05:
## Description
Hello! I'm a first time contributor and thought it'd be great if the repo could have a Code of Conduct to help foster a positive and inclusive environment and give guidelines to help curb negative behaviors. Thus, I created the CODE_OF_CONDUCT.rst document based on the Contributor Covenant v2.1.

## To Do

### DIFF
--- a/CODE_OF_CONDUCT.rst
+++ b/CODE_OF_CONDUCT.rst
@@ -1,96 +1,139 @@
+####################################
 Contributor Covenant Code of Conduct
+####################################
+
 Our Pledge
+==========
+
 We as members, contributors, and leaders pledge to make participation in our
 community a harassment-free experience for everyone, regardless of age, body
 size, visible or invisible disability, ethnicity, sex characteristics, gender
 identity and expression, level of experience, education, socio-economic status,
 nationality, personal appearance, race, caste, color, religion, or sexual
 identity and orientation.
+
 We pledge to act and interact in ways that contribute to an open, welcoming,
 diverse, inclusive, and healthy community.
+
 Our Standards
+=============
+
 Examples of behavior that contributes to a positive environment for our
 community include:
 
-* Demonstrating empathy and kindness toward other people
-* Being respectful of differing opinions, viewpoints, and experiences
-* Giving and gracefully accepting constructive feedback
-* Accepting responsibility and apologizing to those affected by our mistakes,
-and learning from the experience
-* Focusing on what is best not just for us as individuals, but for the overall
-community
+- Demonstrating empathy and kindness toward other people
+- Being respectful of differing opinions, viewpoints, and experiences
+- Giving and gracefully accepting constructive feedback
+- Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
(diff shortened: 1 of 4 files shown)


sampsyo:
Let's continue to consider this! We will want to have a broader conversation about what we want our CoC to look like.

miertla:
This was something I also noticed would be a great contribution (see  #4782). I'm not sure what conflicts need to be resolved so this PR can be merged, but there are also different versions of the Contributor Covenant. I recommended adapting the Contributor Covenant Code of Conduct version 1.4: [https://www.contributor-covenant.org/version/1/4/code-of-conduct/](url). Version 2.1 is the most updated one, but they all vary slightly and I felt this one would work well for this project.

catap:
Personally I vote to use something like https://github.com/domgetter/NCoC

miertla:
That is a good option. It is unlike the traditional CoC which is the point, but I like it. However, I worry that some people may be turned away from contributing due to the blunt language, but I like the idea behind it a lot. It seems fitting considering this community seems to already unofficially be following something similar to it. There should be no need to babysit adults.

Serene-Arc:
Personally, I don't like the linked NCoC. The focus of the document is on the people who want the CoC because of the protection that it brings, and it is entirely negative. It is saying that the burden is on the victim to attempt to both bring up the issue and try to have it resolved by the people in charge of the project, either in private, or just suck it up and deal with it. And I'm sorry, no.

This is not how the real world works. 'Adults' are not inherently trustworthy and it is not babysitting to have a CoC. The people that make people *want* a CoC are usually the ones that are already established. This seems very much like it is someone's attempt to justify their lack of intervention in a community that they control. The people that the CoC is made for *will* take advantage of any ambiguity in the rules, and requiring the people that they are, in this case, abusing to be the active party in resolving the conflict is a non-started.

I think we should have a CoC but it should certainly not be the NCoC linked. It should be an actual CoC as in the one currently in this PR. I am a member of several minorities and a woman in tech. Having explicit rules saying what isn't allowed *does* make me feel better about working on a project, and it *does* make me feel more comfortable. If these rules feel overly burdensome, then you should take a step back and think on why you feel that way about what really is the lowest bar.

Serene-Arc:
Now, after that screed, if we want to adopt this CoC then there needs to be some formatting changes to make use of the RST format, things like headings and stuff in the file. It's been copy-pasted by the look of it so that will need to be added or retrieved from the source. They do offer a markdown version that we can use, or convert it to RST.

sampsyo:
Thanks, @Serene-Arc. I fully agree about rejecting NCoC, for exactly the reasons you stated.

I have been part of several projects that have used [Contributor Covenant](https://www.contributor-covenant.org/) and think it is a perfectly good place to start (and better than not having a CoC!). Logistically, I'd be fine with grabbing the Markdown from the official site and just using that.

Serene-Arc:
I had to do a bit of manual conversion. Took a while but it's all working now. I'll go ahead and merge it when the checks go green.

PR #4534, opened 2022-10-29
Sort and comment "replace"

RollingStar:
## Description

We don't currently comment anything in the default config. For more advanced users, occasional comments may be helpful. I think this is most true in the "replace" section, which otherwise requires an eagle eye and a memorization of Python regex escape rules + how regex works.

Comments should be sparing, analogous to code comments vs. code docs in other projects. In our actual docs I may toss a link to https://regexr.com/.

## To Do

### DIFF
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -40,12 +40,12 @@ ignore: [".*", "*~", "System Volume Information", "lost+found"]
 ignore_hidden: yes
 
 replace:
-    # Bad chars to _
+    # Replace bad characters with _
     # prohibited in many filesystem paths
     '[<>:\?\*\|]': _
     # double quotation mark "
-    "\"": _
-    # \ or /
+    '\"': _
+    # path separators: \ or /
     '[\\/]': _
     # starting and closing periods
     '^\.': _
@@ -53,9 +53,8 @@ replace:
     # control characters
     '[\x00-\x1f]': _
     # dash at the start of a filename (causes command line ambiguity)
-    # https://unix.stackexchange.com/questions/16357/usage-of-dash-in-place-of-a-filename
     '^-': _
-    # Bad chars to nothing
+    # Replace bad characters with nothing
     # starting and closing whitespace
     '\s+$': ''
     '^\s+': ''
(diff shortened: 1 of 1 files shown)


beets/config_default.yaml
```diff
@@ -40,14 +40,25 @@ ignore: [".*", "*~", "System Volume Information", "lost+found"]
...
     '\.$': _
+    # control characters
+    '[\x00-\x1f]': _
+    # dash at the start of a filename (causes command line ambiguity)
+    # https://unix.stackexchange.com/questions/16357/usage-of-dash-in-place-of-a-filename
+    '^-': _
+    # Bad chars to nothing
+    # starting and closing whitespace
     '\s+$': ''
     '^\s+': ''
```
sampsyo, line 61: FWIW, I think trailing spaces are actually illegal on Windows! More here:
https://learn.microsoft.com/en-us/troubleshoot/windows-client/shell-experience/file-folder-name-whitespace-characters

And even more about rules for Windows filenames here:
https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file

beets/config_default.yaml
```diff
@@ -40,14 +40,25 @@ ignore: [".*", "*~", "System Volume Information", "lost+found"]
 ignore_hidden: yes
 
 replace:
+    # Bad chars to _
+    # prohibited in many filesystem paths
+    '[<>:\?\*\|]': _
+    # double quotation mark "
+    "\"": _
+    # \ or /
```
sampsyo, line 48: Maybe mention that these are path separators?

sampsyo (commented):
Seems like a good idea! Maybe phrasing these as full sentences would make them look nice? But not a big deal, obviously.

beets/config_default.yaml
RollingStar: "Bad chars to nothing" i.e. "Bad chars [converted] to nothing", not that they're not bad to any OS. Should it say something else? I've decided that most/all URLs should be in the docs, not the config.

sampsyo: That seems perfectly good! Yeah, I get the blanket usage of "bad" here. Merging shortly!

PR #4546, opened 2022-11-10
Force Tekstowo to check result for match, added Tekstowo test suite

### LINKED ISSUE #4406: lyrics: tekstowo backend does not check if the found song actually matches
### Problem

```sh
$ beet -vv lyrics -p 'kelly bailey'
```

Led to this problem:

```
lyrics: failed to fetch: https://www.musixmatch.com/lyrics/Kelly-Bailey/Black-Mesa-Inbound (404)
lyrics: Genius failed to find a matching artist for 'Kelly Bailey'
lyrics: got lyrics from backend: Tekstowo
lyrics: fetched lyrics: Kelly Bailey - Half-Life 2 - Black Mesa Inbound
Sending event: write
Sending event: after_write
Black eyed Susan
Sun shines in your veins
If the clouds are moving
Never hear her complain
Yeah, black eyed Susan
Just waiting on a drop of rain
…
```

…there are no lyrics. It's an instrumental track.

The backend seems to blindly trust that the first row in search results would be the actual expected song:

https://github.com/beetbox/beets/blob/7c670711aeb2e2b3c9e1c02845c4ae96deaea82b/beetsplug/lyrics.py#L486-L494

But the site absolutely can return something with vaguely similar names!!!

image

### Setup

* OS: FreeBSD -CURRENT
* Python version: 3.9.4
* beets version: current git

luharder:
## Description

Fixes #4406. 

This fix forces the Tekstowo lyrics backend to check if the song retrieved from the search results actually matches the song being searched. The matching check is performed by running string_dist (from beets.autotag.hooks) on both the song name and artist name (e.g. `string_dist(song_name_from_query, song_name_from_search_result)`). If either edit distance is greater than 0.1, the song is considered not a match, and the backend returns that lyrics were not found. If both edit distances are less than (or equal to) 0.1, the song is considered a match, and the backend returns the lyrics for this song. 

NOTE: For efficiency purposes, my fix currently only checks the first song in the search results for a match. Considering only the first search result is essentially what the code did before my fix (except for the fact that it did not check if the search result matched). This is certainly something that can be discussed. 

My fix also adds a stronger test suite in test_lyrics.py for the Tekstowo source. Before my fix, there did not seem to be many (if any) Tekstowo tests. I added both unit tests and integration tests for Tekstowo that mimic some of the tests for other lyrics sources. I additionally added tests that verify that the bug in #4406 is fixed (the "non-matching result" case).

## To Do


beetsplug/lyrics.py
```diff
@@ -493,14 +494,31 @@ def parse_search_results(self, html):
...
+        if not info_elements:
+            return None
+
+        html_title = info_elements[-1].get_text()
+        html_artist = info_elements[-2].get_text()
+
+        title_dist = string_dist(html_title, title)
+        artist_dist = string_dist(html_artist, artist)
+
+        if title_dist > 0.1 or artist_dist > 0.1:
```
sampsyo, line 519: What do you think about making this configurable? Even if we make it completely undocumented (probably a good idea; not much of a reason to customize this), separating out the magic number will make it easy to adapt this in the future if we need to.

This would require something like:


```py
thresh = self.config['dist_thresh'].get(float)
if title_dist > thresh or artist_dist > thresh:
```

…as well as adding the appropriate key to the `self.config.add(...)` call in `LyricsPlugin` to set the default.

sampsyo (commented):
Wow; *very* impressive! This solution looks just perfect. Nice work putting this together!

I only have one minor suggestion, about moving the 0.1 distance threshold to an undocumented config option. Otherwise this looks very shiny and ready to go! :clap:

beetsplug/lyrics.py
luharder: Yes, that would make sense! I just committed a change that adds this configurable distance threshold.

sampsyo:
Hmm, it looks like one of the versions is failing in CI because of a comparison between `float` and a `MagicMock` object… but I don't on first glance see where the `MagicMock` is coming from. Any chance that's obvious to you?
https://github.com/beetbox/beets/actions/runs/3526080150/jobs/5917792758#step:7:255

luharder:
Whoops! I forgot to check if my tests were affected by those changes - turns out they were! The `MagicMock` was showing up because the tests create a `Tekstowo` object with a `MagicMock` object in place of a config object. So when the `extract_lyrics` function was looking in the config for the dist_thresh, it was actually getting a `MagicMock` object.

I fixed this by setting the `Tekstowo` config object to the default config for each unit test class that requires a config. There are probably cleaner ways to do this, but I did not want to mess with the existing design patterns too much.

sampsyo (approved):
Very nice!! Thanks for tracking that down. I agree this fix is a little bit of a "band-aid," but it seems like exactly the right thing given the testing strategy. Thanks again for tackling this!

PR #4549, opened 2022-11-15
Added remixer field

### LINKED ISSUE #4428: Fetch `remixer` field from MusicBrainz
### Problem
Musicbrainz seems to have a field called `remixer` for remixes, this info isn't written to my files.

Example: https://musicbrainz.org/release/7c7f7ddf-c021-4ee8-993d-d1c330b4a36a

Using it for the artist field, similarly to the `artist_credit` option would also be nice.

According to the [Picard docs](https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html) there would even be a proper place to store it in ID3.

### Setup

* OS: Arch Linux
* Python version:  3.10
* beets version: 1.6.0

Bootjewolf:
## Description

Fixes #4428

Adds the remixer field from the musicbrainz database to the beets library. I haven't added any documentation since I didn't know if it is necessary for the changes I made.

## To Do

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -202,6 +202,19 @@ def _flatten_artist_credit(credit):
     )
 
 
+def _get_remixer_names(relations):
+    """ Given a list representing the artist relationships extract the names of
+    the remixers and concatenate them.
+    """
+    remixers = []
+
+    for relation in relations:
+        if relation['type'] == 'remixer':
+            remixers.append(relation['artist']['name'])
+
+    return ', '.join(remixers)
+
+
 def track_info(recording, index=None, medium=None, medium_index=None,
                medium_total=None):
     """Translates a MusicBrainz recording result dictionary into a beets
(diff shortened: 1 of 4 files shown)


beets/autotag/mb.py
```diff
@@ -202,6 +202,19 @@ def _flatten_artist_credit(credit):
     )
 
 
+def _get_remixer_names(relations):
+    """ Given a list representing the artist relationships extract the names of
```
sampsyo, line 206:
```suggestion
    """Given a list representing the artist relationships extract the names of
```

beets/autotag/mb.py
```diff
@@ -202,6 +202,19 @@ def _flatten_artist_credit(credit):
     )
 
 
+def _get_remixer_names(relations):
```
sampsyo, line 205: Maybe we should very slightly generalize this for future-proofing… we could make it take the relation type as a parameter, and make it default to `'remixer'` for now? I can imagine wanting to collect other relationships in a similar way.

sampsyo (commented):
Wahoo; looks awesome!! I just have two very tiny comments.

Bootjewolf:
I updated the PR to include you feedback. I also noticed I introduced a failing test by changing the order of the parameters of the _make_track function so I added the remixer parameter as the last parameter.

sampsyo:
Awesome!! Seems like everything's in order. Thanks again!!

Bootjewolf:
I really liked working on beets since I also use it myself. This will probably not be my last PR for beets.

PR #4556, opened 2022-11-18
The from_scratch flag is called --from-scratch on the command line

sweist:
Fixed the documentation


sampsyo:
Nice catch! Many thanks!!

PR #4563, opened 2022-11-28
Improve playlist plugin docs

JOJ0:
## Description

Improve documentation around the playlist plugin. Clarify what it does and what it does not.

### DIFF
--- a/docs/plugins/playlist.rst
+++ b/docs/plugins/playlist.rst
@@ -24,6 +24,16 @@ name::
 
     $ beet ls playlist:anotherplaylist
 
+A playlist query will use the paths found in the playlist file to match items
+in the beets library. Since `playlist:` submits a regular beets query like any
+other query syntax (eg. `artist:`, `year:`, ...), the sorting order of either
+the configuartion setting `sort_item` or any sorting patterns submitted with
+the query command will be used::
+
+    $ beet ls playlist:/path/to/someplaylist.m3u artist+ year+
+
+Currently retaining the original sort order of the playlist is not supported.
+
 The plugin can also update playlists in the playlist directory automatically
 every time an item is moved or deleted. This can be controlled by the ``auto``
 configuration option.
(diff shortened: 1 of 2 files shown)


docs/plugins/playlist.rst
```diff
@@ -24,6 +24,16 @@ name::
 
     $ beet ls playlist:anotherplaylist
 
+A playlist query will use the paths found in the playlist file to match items
+in the beets library. Since `playlist:` submits a regular beets query like any
```
sampsyo, line 28: Maybe consider linking to the [query page](https://github.com/beetbox/beets/blob/master/docs/reference/query.rst)? Then we probably don't need the "e.g." parenthetical to explain more.

docs/plugins/playlist.rst
```diff
@@ -24,6 +24,16 @@ name::
...
+other query syntax (eg. `artist:`, `year:`, ...), the sorting order of either
+the configuartion setting `sort_item` or any sorting patterns submitted with
+the query command will be used::
```
sampsyo, line 31: Maybe this should be phrased as: "If you want the list in any particular order, you can use the standard beets query syntax for sorting:"

docs/plugins/playlist.rst
```diff
@@ -24,6 +24,16 @@ name::
...
+
+    $ beet ls playlist:/path/to/someplaylist.m3u artist+ year+
+
+Currently retaining the original sort order of the playlist is not supported.
```
sampsyo, line 35: To avoid passive voice: "Playlist queries do not reflect the original order in the `m3u` file."

sampsyo (commented):
Thank you!! Here are some ideas for wording for maximizing clarity.

JOJ0:
Thanks a lot for your suggestions. Tried to incorporate them. That's the new version rendered:
image

sampsyo:
Looks excellent! Thanks for iterating!!

PR #4566, opened 2022-12-01
unimported: ignore_subdirectories works recursively

vemek:
The `ignore_subdirectories` option for the unimported plugin does not ignore directories recursively; that is, if there are directories under the ignored directory, they are not ignored. I believe this is a bug, and this commit corrects that behaviour.

I store my music in a folder checked into git (using [git-annex](https://git-annex.branchable.com/)). This means that there are many, many subdirectories under the .git folder and I am unable to list them individually in the config file ahead of time. I expected this config option to behave for the entire tree under the ignored directory, but it does not. I considered whether other people might find this change in behaviour surprising, but I can't think of a good use-case for the existing behaviour, and I don't think it goes against [the intention of the original feature request](https://github.com/beetbox/beets/issues/4025).

## To Do

### DIFF
--- a/beetsplug/unimported.py
+++ b/beetsplug/unimported.py
@@ -53,7 +53,7 @@ class Unimported(BeetsPlugin):
                 for file in f
                 if not any(
                     [file.endswith(ext) for ext in ignore_exts]
-                    + [r in ignore_dirs]
+                    + [r.startswith(ignore_dir) for ignore_dir in ignore_dirs]
                 )
             }
             in_library = {x.path for x in lib.items()}
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,6 +8,7 @@ Changelog goes here!
 
 New features:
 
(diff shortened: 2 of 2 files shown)


beetsplug/unimported.py
```diff
@@ -53,7 +53,7 @@ def print_unimported(lib, opts, args):
                 for file in f
                 if not any(
                     [file.endswith(ext) for ext in ignore_exts]
-                    + [r in ignore_dirs]
+                    + [r.startswith(ignore_dir) for ignore_dir in ignore_dirs]
```
jackwilsdon, line 56: It's probably a rare scenario, but it'd be good to handle the case where an ignored directory matches the prefix of a directory that the user doesn't want to ignore (e.g. with this change, ignoring `abc` would also ignore the directory `abcdef`).

We'd probably need some special handling to check if the name starts with `ignore_dir` followed by a path separator?

vemek:
You're right - I'll do this properly. I'll extract the check to a method
and add a test for these cases 👍

PR #4579, opened 2022-12-10
Avoid calling chmod in some scenarios.

night199uk:
## Description

The permissions plugin itself walks up the library from the directory we created to the root by default. Therefore it's possible we come across parent directories that weren't created by beets during import and aren't therefore owned by us. Calling os.chmod on a directory not owned by us causes an exception in python; even if the permissions on the parent directory were already correct.

### DIFF
--- a/beetsplug/permissions.py
+++ b/beetsplug/permissions.py
@@ -6,6 +6,7 @@ like the following in your config.yaml to configure:
             dir: 755
 """
 import os
+import stat
 from beets import config, util
 from beets.plugins import BeetsPlugin
 from beets.util import ancestry
@@ -25,7 +26,7 @@ def check_permissions(path, permission):
     """Check whether the file's permissions equal the given vector.
     Return a boolean.
     """
-    return oct(os.stat(path).st_mode & 0o777) == oct(permission)
+    return oct(stat.S_IMODE(os.stat(path).st_mode)) == oct(permission)
 
 
 def assert_permissions(path, permission, log):
@@ -103,7 +104,8 @@ class Permissions(BeetsPlugin):
                 'setting file permissions on {}',
                 util.displayable_path(path),
             )
-            os.chmod(util.syspath(path), file_perm)
+            if not check_permissions(util.syspath(path), file_perm):
+                os.chmod(util.syspath(path), file_perm)
 
             # Checks if the destination path has the permissions configured.
             assert_permissions(path, file_perm, self._log)
@@ -115,7 +117,8 @@ class Permissions(BeetsPlugin):
                 'setting directory permissions on {}',
                 util.displayable_path(path),
             )
-            os.chmod(util.syspath(path), dir_perm)
+            if not check_permissions(util.syspath(path), dir_perm):
+                os.chmod(util.syspath(path), dir_perm)
 
             # Checks if the destination path has the permissions configured.
             assert_permissions(path, dir_perm, self._log)


beetsplug/permissions.py
```diff
@@ -103,7 +103,9 @@ def set_permissions(self, files=[], dirs=[]):
                 'setting file permissions on {}',
                 util.displayable_path(path),
             )
-            os.chmod(util.syspath(path), file_perm)
+            perm = stat.S_IMODE(os.stat(util.syspath(path)).st_mode)
+            if perm != file_perm:
```
wisp3rwind, line 107: We already have a helper function for this check above, I think it would be nice to re-use that here:

```suggestion
            if not check_permissions(syspath(path), file_perm)):
```

Probably, it would be cleaner to change that to use ` S_IMODE`?

wisp3rwind (changes requested):
> The permissions plugin itself walks up the library from the directory we created to the root by default. Therefore it's possible we come across parent directories that weren't created by beets during import and aren't therefore owned by us. Calling os.chmod on a directory not owned by us causes an exception in python; even if the permissions on the parent directory were already correct.

Thanks, these guards look like a safe change! I think some code duplication should be avoided, see the inline comment.

Just to understand more precisely what's really going on here, could you describe your issue in a bit more detail? Briefly skimming the code, I would expect beets not to ` chmod`  any parent directories of the library directory due to the ` startswith`  check in ` dirs_in_library`. Could it be that this check in fact does not work as intended?

night199uk:
> Thanks, these guards look like a safe change! I think some code duplication should be avoided, see the inline comment.

Good catch, I'm going to push a new version that re-uses check_permissions instead of duplicating, thanks!

> Just to understand more precisely what's really going on here, could you describe your issue in a bit more detail? Briefly skimming the code, I would expect beets not to ` chmod` any parent directories of the library directory due to the ` startswith` check in ` dirs_in_library`. Could it be that this check in fact does not work as intended?

So here's the setup I use:
My media is centralized and shared by a number of users, therefore the library permissions are owned and manage by groups; not users. viz a simplified example. user:group; permissions; path.
media:media  drwxrwsr-x   MyMiedia
media:media  drwxrwsr-x   MyMedia/Albums

When I use beet import to import a new album under MyMedia/Albums as some user, I want to set the permissions only on that imported folder to 2775:
someuser:media  drwxrwsr-x   MyMedia/Albums/SomeAlbum

However because of the walk up the tree we attempt to set permissions ALSO on MyMedia/Albums, and MyMedia. This fails because chmod MyMedia (not owned by us will fail with a Python exception running as someuser. This is a realy nasty way to fail the import. It took me a while to figure this out because the cmdline chmod in Linux already does the same check as in this PR; and never fails since the permissions are already 2775. :facepalm: 

I could also have implemented a flag, like "dont_walk_up_the_library" or "only_modify_imported" or some such which would more accurately reflect what I want to do. But, since I trust that the permissions on MyMedia & MyMedia/Albums are always already 2775; this simple check fixes my problem in the minimal way would introducing too much noise or new features/flags.

beetsplug/permissions.py
night199uk: Feedback actions, please check the latest push. Thanks!

wisp3rwind:
Looking good! Would you mind addressing the issue that the style checker is highlighting (imports should be split into separate lines)?

Thanks for explanation; sounds like a very specific situation; but I don't see a drawback to these safeguards, so let's do this!

pobthebuilder:
Oops, sorry. I kind of forgot about this PR, until I had to build a new machine without this patch. I just made the requested changes and pushed with the style fix. Hopefully this is ready to merge!

night199uk:
Oops; I forgot about this PR until I had to build a new machine without the patch. Just actioned the remaining feedback. Hopefully this is ready to merge!

wisp3rwind:
Thanks!

PR #4583, opened 2022-12-13
Add a flowchart of beet's import process to the documentation resources

Serene-Arc:
In an attempt to learn more about the codebase of beets, I've been making a flowchart of the process of importing something into beets. There isn't much in the documentation about the technical details, the nitty gritty stuff, so I thought that, if this would be appropriate for the documentation, that the chart might be helpful for other people wanting to familiarise themselves with the process that beets goes through.

I'm still working on it but I'm making a draft PR if anyone would like to help, either by pointing out mistakes I've made, increasing the detail of parts, or parts that needs to be added to the flowchart. So, thoughts anyone?


RollingStar:
Probably doesn't matter here, but I believe you can generate charts like this from the code itself.

Serene-Arc:
You can generate call charts but my goal with this isn't the exact same thing. I'm trying to explain the process, which isn't the same as the calls to various functions, though it is close in some places.

JOJ0:
I think a chart like this is really helpful. I try to understand bits and pieces of the flow of beets import time and time again when working on beets but it's often that it's hard to keep the whole big picture in mind. 

Do you have an idea where in the docs this could live?

Also it should have a timestamp and tell that it's displaying how things work(ed) at a specific point in time. Certainly a lot if parts are pretty stable and wont change any time soon but still, noting the time of generation makes sense I guess if it should be part of the developers docs.

Serene-Arc:
I would place it in the developers notes in the docs, in a page perhaps elaborating on it. My goal is to have a big, sprawling flowchart where people can zoom in if needed. Like when I need to know about the matching process, there's a sub-flowchart that I can look to specifically.

Adding a specific commit/timestamp is a great idea, that should definitely be done. I'm not sure how much it would really change between versions since it's not supposed to be a call graph, but a step-by-step process that only sometimes corresponds with a function call. I wouldn't expect that to change significantly.

Over time it could be expanded with additional parts but I was going through the import process so that's what represented at the moment.

PR #4588, opened 2022-12-16
Add typing for autotag modules

Serene-Arc:
As discussed, I'm breaking out the changes for the autotag module into its own PR for checking.


beets/autotag/mb.py
```diff
@@ -571,7 +590,7 @@ def match_track(artist, title):
         yield track_info(recording)
 
 
-def _parse_id(s):
+def _parse_id(s: str) -> Optional[Union[str, bytes]]:
```
wisp3rwind, line 593: I don't think this can return `bytes`, can it?

wisp3rwind (changes requested):
I left a few comments (also note the failing style checks); otherwise this looks good to go to me. I'm not 100% sure about the types for the `AlbumInfo`/`TrackInfo` constructors, either, but none of them look obviously wrong. mypy should help us to fix any potential issue later on.

Any particular reason why this is still marked as draft?

Serene-Arc:
@wisp3rwind no real reason for it to be a draft PR other than I expected a fair bit of comments and revision.

beets/autotag/mb.py
Serene-Arc: No it cannot, not sure why I did that.

beets/autotag/hooks.py
```diff
@@ -444,7 +500,7 @@ def update(self, dist):
 
     # Adding components.
 
-    def _eq(self, value1, value2):
+    def _eq(self, value1: Union['Distance', Patterntype], value2) -> bool:
```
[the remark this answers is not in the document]
Serene-Arc: I did that because I kind of assumed that the function would be used similarly to the `__eq__` function. It does seem to be Any based on the uses and other typings.

Serene-Arc:
The failing tests have revealed a problem: there is a circular import between `mb.py` and `hooks.py`. I think it exists regardless of the typing but the typing has caused it to fail here. There are a two options:

- Add imports to the five functions in `hooks.py` that call on the other module
- Move functions around so that there aren't circular imports

Due to the beets version being imported in `mb.py`, the import-only-when-type-checking solution won't work, since the `beets` module needs to be imported regardless.

wisp3rwind:
> Move functions around so that there aren't circular imports

Sounds like the proper solution to me: The classes at the beginning of `hooks.py` and the five glue function at the end are conceptually pretty much unrelated, and it would make more sense to have them in separate files, anyway. I guess we could either just split the file into two (`match_types.py` and `interface.py` or somesuch). At a first glance, it seems that me might also just move the top of `hooks.py` to `match.py`.

Serene-Arc:
Splitting the file seems to be the cleanest option, I'll do that.

Serene-Arc:
Okay a brief attempt revealed that this won't be as simple as I thought it would be, it'll take some effort and rearranging.

Serene-Arc:
So, fun fact: I refactored it all and then realised that it didn't need to be done. There's a PEP, PEP 563, that shunts the parsing of annotations to last place when importing. It's available as a future in Python 3.7 so it should be good to use here and it solves the problem nicely.

Serene-Arc:
I think this PR should be ready for merging if you're good with it @wisp3rwind

beets/autotag/hooks.py
```diff
@@ -68,17 +76,45 @@ class AlbumInfo(AttrDict):
...
+            artist_credit: Optional[str] = None,
+            original_year: Optional[str] = None,
+            original_month: Optional[str] = None,
+            original_day: Optional[str] = None,
+            data_source: Optional[str] = None,
+            data_url: Optional[str] = None,
+            discogs_albumid: Optional[str] = None,
+            discogs_labelid: Optional[str] = None,
+            discogs_artistid: Optional[str] = None,
+            **kwargs,
```
sampsyo, line 116: This is a heroic set of annotations! We should probably someday drop all these positional parameters and cram everything into `kwargs`. But this will do for now!

beets/autotag/hooks.py
```diff
@@ -332,7 +391,7 @@ def __init__(self):
         self._penalties = {}
 
     @LazyClassProperty
-    def _weights(cls):  # noqa: N805
+    def _weights(cls) -> Dict:  # noqa: N805
```
sampsyo, line 394: `Dict[str, float]`, I believe?

sampsyo (approved):
This looks great!! I gave this a careful hand review, but I didn't run MyPy. Things look right from here, and relying on [PEP 563](https://peps.python.org/pep-0563/) seems like the thing to do.

I understand that PEP 563 is causing some trouble: it was originally slated to be enabled by default for Python 3.10, but [it has been pushed back](https://www.reddit.com/r/Python/comments/muyz5h/pep_563_getting_rolled_back_from_python_310/) and I'm not entirely clear about when/if it is supposed to be re-enabled. Anyway, maybe this is just something we need to keep an eye on.

Serene-Arc:
Fixed the typing you suggested @sampsyo. Should be ready to merge!

wisp3rwind:
Thanks for getting this started! I'm just looking at what mypy says about this, there seem to be some issues. I'll try to do a quick follow-up PR.

PR #4591, opened 2022-12-17
Add mypy to tox and CI

wisp3rwind:
Add mypy to tox and CI configurations, and add a few type stubs to the dependencies. This allows failures, similar to the nightly Python tests. Unsurprisingly, mypy does fail currently.

cf. https://github.com/beetbox/beets/pull/4495#issuecomment-1356458837


sampsyo:
Great; this looks good to me. Just having it in place (with failures allowed) will be a solid first step toward actually making it pass…

PR #4600, opened 2022-12-22
fromfilename: Fix regex ordering and add debug logging

### LINKED ISSUE #4561: fromfilename: Fails to detect `<track> <title>` filename patterns
### Problem

```sh
$ beet import path/to/unknown_music_tracks
path/to/unknown_music_tracks (5 items)
No matching release found for 5 tracks.
For help, see: https://beets.readthedocs.org/en/latest/faq.html#nomatch
[S]kip, Use as-is, as Tracks, Group albums, Enter search, enter Id, aBort?
$ u
```

to Use as-is, will led to this problem:

the track numbers are not recognized. For instance, if I run `beet ls`:

```
$ beet ls
 -  - 01 深き森にて
 -  - 02 緋龍の翼
 -  - 03 風の国ルイヴェスタ
 -  - 04 終りなき狂詩曲
 -  - 05 無慈悲な夜に
```

the track numbers are still part of the files.

If I disable fromfilename plugin, the whole output is empty (no track number nor track name).

Use any unknown track, such as a dummy sound file you would create, or rename the folder/files so artist/track cannot be identified (this is however a real use case; I'm simply downloading lesser known tracks from indie bands, etc.)

### Setup

* OS: Ubuntu 20.04
* Python version: 3.8.10
* beets version: 1.6.0

```yaml
directory: ~/path/to/Music
import:
    copy: no
    write: no
plugins: fromfilename
```

Note: write: yes didn't help. In fact, it didn't even seem to change the track names, but that's another issue. I assume `beet ls` alone should show expected information.

JOJ0:
## Description

- Fixes #4561 
- Adds debug log messages that inform when the plugin is replacing bad (missing) artist, title or track number fields.

## To Do

- [x] ~Documentation.~
- [x] Changelog.
- [x] ~Tests.~


JOJ0:
I swapped the other of the regex as suggestin #4561 but would like to discuss whether iut shouldn't be moved to the very end of the regex list.

This is what we would have now:
https://github.com/beetbox/beets/pull/4600/files#diff-5d4a5c24257905f266f635b0e6605e82e4cc8e8f6d877f77abc622c21beea7faR32-R36

wisp3rwind:
Thanks for getting this started! As you already noticed in #4561, this pattern should really go to the very end of the list, since it's the least specific.

wisp3rwind:
LGTM; any reason why this is still marked as a draft?

JOJ0:
> LGTM; any reason why this is still marked as a draft?

I'm not sure if the debug log messages with a simple "textual" prefix of `fromfilename: ` are the correct way of doing it. Usually plugins being a child of BeetsPlug, have a nice logging magic built-in: https://github.com/beetbox/beets/blob/master/beets/plugins.py#L88-L91

Implementing this would be quite some hassle because the method the logging takes places in my PR is not part of the plugin. If you think it's sufficient like this and not worth the reafactoring hassle I'm very fine with it.

Also: I'll add a quick changelog entry.

wisp3rwind:
> > LGTM; any reason why this is still marked as a draft?
> 
> I'm not sure if the debug log messages with a simple "textual" prefix of `fromfilename: ` are the correct way of doing it. Usually plugins being a child of BeetsPlug, have a nice logging magic built-in: https://github.com/beetbox/beets/blob/master/beets/plugins.py#L88-L91
> 
> Implementing this would be quite some hassle because the method the logging takes places in my PR is not part of the plugin. If you think it's sufficient like this and not worth the reafactoring hassle I'm very fine with it.

You're right, that would be much cleaner to do, thanks for thinking ahead! I think it's not a lot of work actually: Make `filename_task` a method of `FromFilenamePlugin` (i.e. indent it and add the `self` parameter), then pass the `self._log` logger to `apply_matches`. There are many plugins that do similar things, for example fetchart.

Would you mind doing that change?

> Also: I'll add a quick changelog entry.

:+1:

JOJ0:
Sure thing. Will do tomorrow or soonish! Appreciate your hints.

JOJ0:
Hi, changes as suggested. Tested a little and works well. Also changed log level to INFO. I figured that it should be enough to pass one `-v` to see what messages. Is there any case for that plugin where INFO level could be inappropriate?

If the latter is fine, ready for a final review.

wisp3rwind:
> Hi, changes as suggested. Tested a little and works well. Also changed log level to INFO. I figured that it should be enough to pass one `-v` to see the messages. Is there any case for this plugin where INFO level could be inappropriate?

Looks good, our default log level for plugin stages appears to be `WARNING`, so `INFO` wont show up by default.

> There seems to be an issue with the CI. Could be something with Sphinx build. I'm not sure if it's my fault.
> 
> The only thing I did which was Sphinx related was the changelog. I reverted that commit to see if the error persists. Turns out it does.

Weird, I don't immediately have a glue what might be going on here, the CI logs do seem to point towards an issue with Sphinx.

> Maybe something is broken with the current Sphinx version or even something with the Python install process broken on github's end? Not sure but please look into the CI logs. Thanks a lot! I'll continue work on this PR earliest on the 3rd of Jan.

jackwilsdon:
I've fixed the Sphinx issue in 2106f471affd1dab35b4b26187b9c74d034528c5 - if you update your branch from master then you should be good :+1:

JOJ0:
Thanks for the fix @jackwilsdon. Ready to review again @wisp3rwind.

sampsyo:
Nicely done, all! And great work tracking down that Sphinx version problem, @jackwilsdon!!!

PR #4633, opened 2023-01-11
Refactor metadata source ID extraction utilities

JOJ0:
## Description

Tidy up and streamline ID extraction utilities used by metadata source plugins. Enable the usage of those utilities from everywhere.

- Introduce a new module `beets.util.id_extractors`
- Move the Discogs ID extractor function to the new module.
- Move the regex variables for the following services to the new modul:
  - Spotify
  - Deezer
  - Beatport
  - Leave a note about Bandcamp ID's
- Refactor the MetadataSourcePlugin._get_id() to use the new utility location.
- Fix the Beatport plugin to use the shared functionality of MetadataSourceplugin which it already inherits.
- Fix the Discogs ID extractor function to support a legacy format that is still in use.

## To Do

- [x] ~Documentation.~
- [x] Changelog. Didn't add one for now. Does it make sense for such an "internal change"? We don't change anything from a user perspective.
- [x] Tests.

### DIFF
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -705,22 +705,27 @@ class MetadataSourcePlugin(metaclass=abc.ABCMeta):
 
         return artist_string, artist_id
 
-    def _get_id(self, url_type, id_):
+    @staticmethod
+    def _get_id(url_type, id_, id_regex):
         """Parse an ID from its URL if necessary.
 
         :param url_type: Type of URL. Either 'album' or 'track'.
         :type url_type: str
         :param id_: Album/track ID or URL.
         :type id_: str
+        :param id_regex: A dictionary containing a regular expression
+            extracting an ID from an URL (if it's not an ID already) in
+            'pattern' and the number of the match group in 'match_group'.
+        :type id_regex: dict
         :return: Album/track ID.
         :rtype: str
         """
-        self._log.debug(
-            "Searching {} for {} '{}'", self.data_source, url_type, id_
+        log.debug(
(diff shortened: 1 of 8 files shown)


JOJ0:
One feature for future use of this refactoring PR is https://github.com/beetbox/beets/pull/4474, which currently is set back to draft state until I get to rebasing to only include the relevant parts.

JOJ0:
@sampsyo, this one is for you <3, hope you like it. Suggestions on "naming things" and documentation/changelog are very welcome.

JOJ0:
Hi @sampsyo, I know this one is probably a low priority one since it actually doesn't change any behaviour, nor does it fix anything (except a little oversight in the currently defunct beatport plugin). Certainly I ultimately did this to get a better view of what https://github.com/beetbox/beets/pull/4474 will actually change.
I think merging this one is worthwile in any case since it just does some tidying up and most of all I think "it's a rather low hanging fruit" :-)

- Tests are fixed.
- One remaining question: Do you think a changelog would be good? Something like "internal change/reorganization of metadata ID extraction" - something like that - or not too important and just leave it out?

JOJ0:
I'm merging this now since:

- I've added additional tests that check whether Spotify, Deezer and Bandcamp URLs and ID's are translated/detected correctly.
- It currently is a good time since no PR's are "in flioght" that seem ot touch does parts in the code.
- @sampsyo had agreed a long time ago already that it's a good idea to split the ID extraction tools out to the utils module.
- I'm taking full responsibility if something breaks but I think we are covered with tests well and this is ready to merge.

PR #4636, opened 2023-01-16
Deprecate `absubmit` and `acousticbrainz` plugins

### LINKED ISSUE #4627: acousticbrainz and absubmit plugins should be deprecated
Acousticbrainz has [shut down](https://blog.metabrainz.org/2022/02/16/acousticbrainz-making-a-hard-decision-to-end-the-project/) and will go completely offline in early 2023 according to that blog post. The `acousticbrainz` and `absubmit` plugins should be deprecated.

Attempting to use it now just results in a lot of "acousticbrainz did not provide info about..." messages.

trolley:
## Description

Fixes #4627.

AcousticBrainz is shutting down as of early 2023. Deprecate the `absubmit` and `acousticbrainz` plugins, and modify them to only function if an AcousticBrainz server has been configured.

## To Do

### DIFF
--- a/beetsplug/absubmit.py
+++ b/beetsplug/absubmit.py
@@ -56,10 +56,13 @@ class AcousticBrainzSubmitPlugin(plugins.BeetsPlugin):
     def __init__(self):
         super().__init__()
 
+        self._log.warning("This plugin is deprecated.")
+
         self.config.add({
             'extractor': '',
             'force': False,
-            'pretend': False
+            'pretend': False,
+            'base_url': ''
         })
 
         self.extractor = self.config['extractor'].as_str()
@@ -79,7 +82,7 @@ class AcousticBrainzSubmitPlugin(plugins.BeetsPlugin):
             except OSError:
                 raise ui.UserError(
                     'No extractor command found: please install the extractor'
-                    ' binary from https://acousticbrainz.org/download'
+                    ' binary from https://essentia.upf.edu/'
                 )
             except ABSubmitError:
                 # Extractor found, will exit with an error if not called with
@@ -96,7 +99,15 @@ class AcousticBrainzSubmitPlugin(plugins.BeetsPlugin):
             self.extractor_sha.update(extractor.read())
         self.extractor_sha = self.extractor_sha.hexdigest()
 
-    base_url = 'https://acousticbrainz.org/api/v1/{mbid}/low-level'
+        self.url = ''
+        base_url = self.config['base_url'].as_str()
+        if base_url:
+            if not base_url.startswith('http'):
+                raise ui.UserError('AcousticBrainz server base URL must start '
+                                   'with an HTTP scheme')
+            elif base_url[-1] != '/':
+                base_url = base_url + '/'
+            self.url = base_url + '{mbid}/low-level'
 
     def commands(self):
         cmd = ui.Subcommand(
@@ -118,10 +129,17 @@ only files which would be processed'
         return [cmd]
 
     def command(self, lib, opts, args):
-        # Get items from arguments
-        items = lib.items(ui.decargs(args))
-        self.opts = opts
-        util.par_map(self.analyze_submit, items)
+        if not self.url:
+            raise ui.UserError(
+                'This plugin is deprecated since AcousticBrainz no longer '
+                'accepts new submissions. See the base_url configuration '
+                'option.'
+            )
+        else:
+            # Get items from arguments
+            items = lib.items(ui.decargs(args))
+            self.opts = opts
+            util.par_map(self.analyze_submit, items)
 
     def analyze_submit(self, item):
         analysis = self._get_analysis(item)
@@ -179,7 +197,7 @@ only files which would be processed'
     def _submit_data(self, item, data):
         mbid = item['mb_trackid']
         headers = {'Content-Type': 'application/json'}
-        response = requests.post(self.base_url.format(mbid=mbid),
+        response = requests.post(self.url.format(mbid=mbid),
                                  json=data, headers=headers)
         # Test that request was successful and raise an error on failure.
         if response.status_code != 200:
--- a/beetsplug/acousticbrainz.py
+++ b/beetsplug/acousticbrainz.py
@@ -22,7 +22,6 @@ import requests
 from beets import plugins, ui
 from beets.dbcore import types
 
-ACOUSTIC_BASE = "https://acousticbrainz.org/"
 LEVELS = ["/low-level", "/high-level"]
 ABSCHEME = {
     'highlevel': {
@@ -138,11 +137,22 @@ class AcousticPlugin(plugins.BeetsPlugin):
     def __init__(self):
         super().__init__()
 
+        self._log.warning("This plugin is deprecated.")
+
         self.config.add({
             'auto': True,
             'force': False,
-            'tags': []
+            'tags': [],
+            'base_url': ''
         })
+
+        self.base_url = self.config['base_url'].as_str()
+        if self.base_url:
+            if not self.base_url.startswith('http'):
+                raise ui.UserError('AcousticBrainz server base URL must start '
+                                   'with an HTTP scheme')
+            elif self.base_url[-1] != '/':
+                self.base_url = self.base_url + '/'
 
         if self.config['auto']:
             self.register_listener('import_task_files',
@@ -171,8 +181,13 @@ class AcousticPlugin(plugins.BeetsPlugin):
         self._fetch_info(task.imported_items(), False, True)
 
     def _get_data(self, mbid):
+        if not self.base_url:
+            raise ui.UserError(
+                'This plugin is deprecated since AcousticBrainz has shut '
+                'down. See the base_url configuration option.'
+            )
         data = {}
-        for url in _generate_urls(mbid):
+        for url in _generate_urls(self.base_url, mbid):
             self._log.debug('fetching URL: {}', url)
 
             try:
@@ -328,8 +343,8 @@ class AcousticPlugin(plugins.BeetsPlugin):
                                 'because key {} was not found', subdata, v, k)
 
 
-def _generate_urls(mbid):
+def _generate_urls(base_url, mbid):
     """Generates AcousticBrainz end point urls for given `mbid`.
     """
     for level in LEVELS:
-        yield ACOUSTIC_BASE + mbid + level
+        yield base_url + mbid + level
--- a/docs/plugins/absubmit.rst
+++ b/docs/plugins/absubmit.rst
@@ -1,8 +1,17 @@
 AcousticBrainz Submit Plugin
 ============================
 
-The ``absubmit`` plugin lets you submit acoustic analysis results to the
-`AcousticBrainz`_ server.
+The ``absubmit`` plugin lets you submit acoustic analysis results to an
+`AcousticBrainz`_ server. This plugin is now deprecated since the
+AcousicBrainz project has been shut down.
+
+As an alternative the `beets-xtractor`_ plugin can be used.
+
+Warning
+-------
+
+The AcousticBrainz project has shut down. To use this plugin you must set the
+``base_url`` configuration option to a server offering the AcousticBrainz API.
 
 Installation
 ------------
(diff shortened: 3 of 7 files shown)


JOJ0:
@trolley I don't see a problem with finally merging this one. Would you be so kind and resolve conflicts and mark remaining checkmarks.

Could you post how this deprecation now would present itself to a beets user?

beetsplug/acousticbrainz.py
```diff
@@ -15,321 +15,28 @@
...
+        return [deprecated]
 
-def _generate_urls(mbid):
-    """Generates AcousticBrainz end point urls for given `mbid`.
-    """
-    for level in LEVELS:
-        yield ACOUSTIC_BASE + mbid + level
+    def deprecation_message(self, lib, opts, args):
+        ui.print_("This plugin is deprecated. AcousticBrainz "
+                  "is shutting down.")
```
JOJ0, line 42: Maybe we could be a little more elaborate here. Something similar to what you've put in the changelog sounds a little nicer:

This plugin is deprecated and no longer has any function since the AcousicBrainz project has been shut down.

trolley:
Given a config file with `absubmit` and/or `acousticbrainz` plugins in it, like:
```
plugins: absubmit acousticbrainz

absubmit:
  auto: yes
```
These messages are printed to the user when beets is invoked:
```
(beets-dev)  π ~ ❯ beet ls
acousticbrainz: This plugin is deprecated and no longer functions.
absubmit: This plugin is deprecated and no longer functions.
```

If either plugin is manually invoked the same messages appear, along with the longer message:
```
(beets-dev)  π ~ ❯ beet absubmit
acousticbrainz: This plugin is deprecated and no longer functions.
absubmit: This plugin is deprecated and no longer functions.
This plugin is deprecated and no longer has any function since AcousticBrainz no longer accepts new submissions.
(beets-dev)  π ~ ❯ beet acousticbrainz
acousticbrainz: This plugin is deprecated and no longer functions.
absubmit: This plugin is deprecated and no longer functions.
This plugin is deprecated and no longer has any function since the AcousicBrainz project has been shut down.
```

Serene-Arc:
`asubmit` is likely pointless moving forwards but can't there be a local version of acousticbrains? Such as in a docker instance? It might be a little premature to say that the entire plugin is deprecated to the point of uselessness.

trolley:
If the maintainers decide they'd prefer that I will adjust the PR.

Personally, I don't see much point considering the maintainers of AcousticBrainz themselves say:

> Unfortunately, a number of things happened with the data that we collected which made us decide that the quality of the data isn’t as useful as we had hoped.

JOJ0:
> `asubmit` is likely pointless moving forwards but can't there be a local version of acousticbrains? Such as in a docker instance? It might be a little premature to say that the entire plugin is deprecated to the point of uselessness.

There is a local version of this plugin already that does everything absubmit die and more. It is a third party plugin but I'm co-maintaining it: https://github.com/adamjakab/BeetsPluginXtractor

JOJ0:
@Serene-Arc @trolley and I'd like to include @sampsyo into this discussion: Now after having thought about this a little more I'd like to go one step further and even add to the deprecation message that it should be considered to move from acousticbrainz plugin to BeetsPluginXtractor. 

If adding it to the deprecation message is not decided to be the best way, we could also add it to the acousticbrainz plugins docs page: https://beets.readthedocs.io/en/stable/plugins/acousticbrainz.html

What do you think about this idea? I would hope to get some movement to the third party plugin.

trolley:
Output with latest change:
```
(beets-dev)  π ~ ❯ beet absubmit
acousticbrainz: This plugin is deprecated and no longer functions.
absubmit: This plugin is deprecated and no longer functions.
The AcousticBrainz project has been shut down and thus no longer accepts new submissions.
(beets-dev)  π ~ ❯ beet acousticbrainz
acousticbrainz: This plugin is deprecated and no longer functions.
absubmit: This plugin is deprecated and no longer functions.
The AcousicBrainz project has been shut down.
```

JOJ0:
@sampsyo any last wording advices for this deprecation message? Sorry for bothering, I just feel a little afraid while having to do a final decision for kicking out a part of beets that's  been in there for years and did good work. Even if it's of no use any more. Help! Thanks so much in advance! :-)

sampsyo:
The deprecation messages look great; thank you so much to everybody for iterating on them! Short and simple and clear.

I'll defer to others about whether it's the right thing to delete the code (vs. just printing the deprecation message). As @Serene-Arc implies, if there's any chance people have their own local thing running, then maybe that's a reason to keep the code. But as @JOJ0 mentions also, `BeetsPluginXtractor` is clearly the spiritual successor.

JOJ0:
@Serene-Arc are you fine with kicking out the code or should we keep it? Your idea of "dockerizing" AB and keeping this plug as-is (in general) is a different approach to the xtractor plug. On the other hand, git never forgets so it's not lost forever. Your call, I'm fine either way! :-) Thanks!

Serene-Arc:
I think that, if the code isn't completely useless, which it isn't because there are docker services, then there's no real cost to keeping it in the codebase. It might disrupt a few workflows to remove it and if there's no real reason to remove it, then I think it should be kept.

The deprecation message will make sure that people know to switch if they're not explicitly running their own service, and moving forwards we can discuss whether to decline any further development on the plugins.

zaptrem:
Fuzzy but sort-of correct data like this is extremely useful to ML researchers working on music generation.

JOJ0:
> I think that, if the code isn't completely useless, which it isn't because there are docker services, then there's no real cost to keeping it in the codebase. It might disrupt a few workflows to remove it and if there's no real reason to remove it, then I think it should be kept.

Alright then let's keep it :-) I would have believed in git, saving everything for us until the end of time but certainly if this plugin stays here, it's not hidedn in old commits and the chance someone picks up the idea and creates a dockerized versoin of AcousticBrainz might be more likely.

@trolley sorry for the hassle but would you be so kind and bring back the deleted stuff. I hope it's not too much of a hassle now :-)

> The deprecation message will make sure that people know to switch if they're not explicitly running their own service, and moving forwards we can discuss whether to decline any further development on the plugins.

Should deprecation messages be altered now or are they fine like that? Not sure....

JOJ0:
> Fuzzy but sort-of correct data like this is extremely useful to ML researchers working on music generation.

That's out of question! Yes! And just to clarify: The deprecation of this defunct plugin is not done because we don't think the outcome isn't useful because someone (MB) said that!.

I still think though, that the alternative that currently _is_ stable and should be supported is https://github.com/adamjakab/BeetsPluginXtractor. You might have missed that in the long threads above but it uses _exactly_ the algorithms / the software-stack AcousticBrainz was using, only that it allows to even use _newer_ versions of that software (https://github.com/mtg/essentia) that might be of even more use because they might have fixed some bugs AB was facing in the meantime.v I think that's awesome and my personal experience is now even better since I started to use xtractor instead of acousticbrainz plugin. The main problem I had with acousticbrainz was that every now and then some files could never be analysed and it was kind of a blackbox what the problem was. Was it my file, was it a server error over at acousticbrainz.org, I never new. Now everything under control because it runs locally with newer versions and debug logging if I want that.

trolley:
I restored the acousticbrainz plugin but added a new config option `hostname` that's required.

I tested it with a couple scenarios:

- without the `hostname` config set:

```
(beets-dev)  π ~ ❯ beet acousticbrainz
error: AcousticBrainz server hostname not specified
```

- with the `hostname` config set, but to a hostname that doesn't work:

```
(beets-dev)  π ~ ❯ beet acousticbrainz
acousticbrainz: getting data for: Thrice - Horizons / East - The Color of the Sky
acousticbrainz: request error: HTTPConnectionPool(host='fakeserver', port=80): Max retries exceeded with url: /a24cd93d-7a48-4971-90de-38f628a7a8ff/low-level (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f1cb1b38eb0>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution'))
```

I'll try to set up my own docker instance of AcousticBrainz to try it with that when I can.

beetsplug/acousticbrainz.py
```diff
@@ -141,9 +140,17 @@ def __init__(self):
...
             'auto': True,
             'force': False,
-            'tags': []
+            'tags': [],
+            'hostname': ''
         })
 
+        self.hostname = self.config['hostname'].as_str()
+        if self.hostname:
+            if self.hostname[-1] != '/':
```
trolley, line 149: I'm considering a couple things about this:
- I think this option should be renamed to `base_url` since it needs to include the scheme and possibly port.
- Should more validation happen here, like verifying the option _does_ include a scheme? Only `http` or `https` would be supported.

trolley:
I had no luck building their docker image. Could someone else give it a try?

docs/plugins/absubmit.rst
```diff
@@ -1,66 +1,8 @@
...
-  the items which would be processed.
-  Default: ``no``.
-
-.. _streaming_extractor_music: https://acousticbrainz.org/download
-.. _FAQ: https://acousticbrainz.org/faq
-.. _pip: https://pip.pypa.io
-.. _requests: https://requests.readthedocs.io/en/master/
-.. _github: https://github.com/MTG/essentia
-.. _AcousticBrainz: https://acousticbrainz.org
+.. _AcousticBrainz: https://acousticbrainz.org/
```
JOJ0, line 8: Since we decided the plugin should stay and allow future tinkering with dockerised self-hosting, I suggest we bring back all the docs above this line (including the deprecation message).

Also I would now propse that since we have BeetsPluginXtractor in the community-plugins list already (https://beets.readthedocs.io/en/stable/plugins/index.html#other-plugins) I'd like to include a sentence next to the deprecation message that says:

As an alternative to the `absubmit` and `acouticbrainz` plugins ´[beets-xtractor](https://github.com/adamjakab/BeetsPluginXtractor)` can be used

docs/plugins/index.rst
```diff
@@ -155,9 +155,6 @@ Autotagger Extensions
 Metadata
 --------
 
-:doc:`absubmit <absubmit>`
-   Analyse audio with the `streaming_extractor_music`_ program and submit the metadata to the AcousticBrainz server
-
```
JOJ0, line 160: Also this should be brought back I suppose. Since a dockerized version would require to have exactly the same system of using `absubmit` to analyze and push to server/container and `acousticbrainz` then to retrieve it back from there.

JOJ0 (changes requested):
The goals of this PR from my point of view would be:

- Print deprecation messages in both plugins. DONE
- Allow future tinkering / containerization. Code should be kept in-tact.
- Proposal: Mention beets-xtractor plugin as a currently working replacement in the docs of both plugins.

Non-goals IMHO:
- Provide a functioning container version.
- Alter the plugin further to improve those future ideas.

We certainly can leave the hostname feature you already implemented @trolley , it is a good idea and a nice start for future tinkerers! :-)

@serene-arc, what do you think about setting these goals to finalise this PR?

trolley:
Alright, I took another crack at it. Now both plugins remain functional if you provide a configuration option to set the server. Both still print a deprecation warning if included in a config file, and a more detailed deprecation message pointing to the new config option of explicitly invoked.

beetsplug/absubmit.py
```diff
@@ -79,7 +82,7 @@ def __init__(self):
             except OSError:
                 raise ui.UserError(
                     'No extractor command found: please install the extractor'
-                    ' binary from https://acousticbrainz.org/download'
+                    ' binary from https://essentia.upf.edu/'
```
trolley, line 85: The download links on the AcousticBrainz site are dead now, so I updated this to the official Essentia project homepage where it's still available.

trolley: Actually it's only the 64-bit Linux version that's dead, but it's probably still a good idea to reference the official project site instead of the mirrored downloads on the AcousticBrainz site which may not be around for long.

JOJ0 (approved):
Thanks for the changes in wording and link changes. That is much appreciated and very thoughtful! Great! And many thanks for the patience since it took some time to really know where and how this deprecation should be going. I think it was all worth it and I am finally....merging :bomb: :

duckfromdiscord:
I have a quick question about this PR, feel free to hide my message if it seems unimportant.

I wanted to thank you for including the custom server option. My question is, are there any custom servers out there we can point the plugin to?

JOJ0:
Not that I'm aware of. FWIW, have you seen the suggested alternative for getting similar information into your library noted in the second paragraph here: https://beets.readthedocs.io/en/latest/plugins/acousticbrainz.html

duckfromdiscord:
`beets-xtractor`? I've seen it also, I was going to ask there but I noticed they didn't have a base URL option. I wanted to see if there was an online dataset somewhere that would include analyses of other music. Thank you for the help.

JOJ0:
xtractor does the analysis on your machine ☝️😁

PR #4643, opened 2023-01-20
Add `sphinx` and `sphinx_rtd_theme` as deps for a new `docs` extra

Steffo99:
## Description

Currently, the dependencies to build the documentation are not specified anywhere.

This PR introduces a new `docs` extra that specifies what is required to build the documentation locally.

## To Do


docs/changelog.rst
```diff
@@ -8,6 +8,7 @@ Changelog goes here!
 
 New features:
 
+* Add ``sphinx`` and ``sphinx_rtd_theme`` as dependencies for a new ``docs`` extra
```
sampsyo, line 11: This could go under "For packagers" (if relevant) or just "Other changes" below, since it's not really a new feature for users.

sampsyo (commented):
Thanks for looking into this!! Even though it's super simple, I have a couple of questions:

* Do we actually depend on `sphinx_rtd_theme` to build the docs locally? At least here, I don't seem to use anything other than the default theme when building things…
* Is the goal with the `docs` extra just for manual installation, i.e., to let people manually type in `pip install .[docs]` or similar? Or is there some automated scenario I'm not yet thinking of that would use this extra?

docs/changelog.rst
Steffo99: Thanks for the notice! I didn't see it, since "Other changes" is not an header :sweat_smile:

Steffo99:
> Do we actually depend on sphinx_rtd_theme to build the docs locally? At least here, I don't seem to use anything other than the default theme when building things…
> —@sampsyo

Beets currently doesn't depend on it, but it has the theme [applied by default when the documentation is compiled by readthedocs.io](https://beets.readthedocs.io/en/stable/).

I think it would be a good idea to use the same theme locally, so that whoever makes an edit to the docs can see immediately how it will be displayed in the "official" version.

> Is the goal with the docs extra just for manual installation, i.e., to let people manually type in pip install .[docs] or similar? Or is there some automated scenario I'm not yet thinking of that would use this extra?
> —@sampsyo

Manual installation is the use case I had in mind, but, if the project ever wanted to move off readthedocs.io, the new Continuous Deployment infrastructure could use that extra to install the dependencies to build the docs.

sampsyo (approved):
Cool; thanks for clarifying! Having this around as a manual option seems perfectly reasonable to me.

JOJ0:
Hi @Steffo99, I would like to take over this PR. If you could fix conflicts in changelog.rst I think we are ready to merge since @sampsyo had approved it already :-)

PR #4644, opened 2023-01-20
Configure `sphinx_rtd_theme`

Steffo99:
## Description

This PR makes the documentation build locally with `sphinx_rtd_theme` instead of the default theme, matching [how the documentation is displayed on Read the Docs](https://beets.readthedocs.io/en/stable/).

Additionally, this configures the theme to use the Beets logo color as the background of the nav header, customizing the documentation page for the project.

A screenshot of the locally built docs after the changes. The page now uses `sphinx_rtd_theme`. The background of the header containing the project's name and the searc bar is dark red.

## To Do

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,6 +8,8 @@ Changelog goes here!
 
 New features:
 
+* Recolor the Sphinx documentation header to use the Beets' logo color
+* Configure Sphinx so that it uses ``sphinx_rtd_theme`` locally
 * We now import the remixer field from Musicbrainz into the library.
   :bug:`4428`
 * :doc:`/plugins/mbsubmit`: Added a new `mbsubmit` command to print track information to be submitted to MusicBrainz after initial import.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -34,6 +34,7 @@ linkcheck_ignore = [
 ]
 
 # Options for HTML output
+html_theme = 'sphinx_rtd_theme'
 htmlhelp_basename = 'beetsdoc'
 
 # Options for LaTeX output
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -35,6 +35,9 @@ linkcheck_ignore = [
 
 # Options for HTML output
 html_theme = 'sphinx_rtd_theme'
+html_theme_options = {
+    "style_nav_header_background": "#a23632",
+}
 htmlhelp_basename = 'beetsdoc'
 
 # Options for LaTeX output


sampsyo:
I appreciate your looking into this! To be perfectly honest, I have never really loved the RTD theme, and it seems to get more out-of-date-looking with time. I have actually been considering trying to go in the opposite direction: to use the default [Alabaster](https://alabaster.readthedocs.io/en/latest/) theme everywhere, which IMO looks great and more up-to-date in general, including on RTD. Would that be a crazy thing to do?

Steffo99:
> ...to use the default [Alabaster](https://alabaster.readthedocs.io/en/latest/) theme everywhere...
> —@sampsyo 

That's your decision to take!

I personally find Alabaster more difficult to read for some reason, perhaps because of the more muted colors, the serif font, and the less visually distinct sidebar...  
I'm not sure about other people, though. 

I think it would be a good idea to gather some more feedback before switching!

> Would that be a crazy thing to do?
> —@sampsyo

It definitely wouldn't be a crazy thing to do: I know a few packages which made the same decision, such as [pandaSDMX](https://pandasdmx.readthedocs.io/en/v1.0/) and [docker-py](https://docker-py.readthedocs.io/en/stable/index.html)...

sampsyo:
Yes, gathering some feedback would be great!!

JOJ0:
A poll around the theme would be nice. I don't know where we could easily do a pll around that? GitHub discussion? The beets discourse forum?

My personal opinions:

Having the same theme as in the final rendered online-docs when building locally helps me a lot.#

My eyes are very used the readthedocs theme because it is everywhere. 

I do like the alabaster theme and I think in for example that case I am also very used to it: https://click.palletsprojects.com/en/8.1.x/

I'm not quite sure yet if for the beets docs I would prefer alabaster.

Is it easily possible to use "any" theme with readthedocs?

I recently ame across a very nice theme I now forgot the name..... It had possibilities like "Tips" and "Warning" boxes and stuff. I think that provides intersting new possibilities....FIXME  need to find that again ;-)

Steffo99:
@JOJ0 Click does not use Alabaster, and uses [a variation of the Flask theme](https://github.com/pallets/click/blob/main/docs/conf.py#L33) instead.

JOJ0:
Furo theme is what I was trying to say: https://sphinx-themes.org/sample-sites/furo/

I like the "important boxes" but otherwise I'm not so sure anymore.

Anyway what holds us back to just change to set rtd for local rendering for now? @sampsyo would you be ok with that?

I'm not so sure about the color change because it would provide a different look between off/online afain but my opinion shouldnt hold a merge back.

@Steffo99 a poll for future design could happen as a separate thing anyway, would you like to make one? https://github.blog/changelog/2022-04-12-discussions-polls/

Steffo99:
> Furo theme is what I was trying to say: https://sphinx-themes.org/sample-sites/furo/
>
> I like the "important boxes" but otherwise I'm not so sure anymore.
> 
> —@JOJ0 

It isn't too bad!

(All themes support `.. important::`, `.. note::`, `.. warning::` and `.. danger::` boxes, as they are implemented directly in Sphinx, they just aren't as apparent.)

> I'm not so sure about the color change because it would provide a different look between off/online again but my opinion shouldnt hold a merge back.
>
> —@JOJ0 

The color change is applied online too, as it is officially supported by Read the Docs!

See the documentation of [my WIP Python config library](https://cfig.readthedocs.io/en/latest/) for an example :D

> a poll for future design could happen as a separate thing anyway, would you like to make one? https://github.blog/changelog/2022-04-12-discussions-polls/
>
> —@JOJ0 

I don't contribute enough to Beets to feel comfortable making one, and I don't know what themes to offer as options: I think that it would be more appropriate if a project member made one instead!

JOJ0:
@sampsyo I would like to merge this as-is (fixing changelog only)

I'm fine with (finally) making a local docs build look like our readthedocs pages.

And I'm also fine with the tiny color change - give it a little twist for once :-)

If you strongly advise against it, I could remove the color change and just merge the "streamlining local/rtd build". No problem

Mr "get rid of low hanging fruit PR's" ;-)

JOJ0:
Whoops @sampsyo I guess I didn't get the memo, we seem to use Alabaster theme now for the "latest" docs: https://beets.readthedocs.io/en/latest/

Was that an intentional change (which is perfectly fine for me, actually I like that theme too :-), or was it due to defaults-change on readthedocs end (I had some fiddling with new config yaml thing recently in other projects since RTD is deprecating some config-related things)

sampsyo:
Huh, I'm surprised to see that change too! I don't remember making any changes myself that would switch the default theme. I'm OK with it too (I prefer Alabaster over the default RTD theme, purely aesthetically)… so maybe let's just roll with that and make this the "official" theme?

JOJ0:
Yeah why not! Let's just stick with it 😆

JOJ0:
I'm sorry I was too lazy for a PR but I a quickfixed the sidebar in Alabaster. It did not stay "in place" like we had it before. More adjustments might come in proper PR.

I'm closing this one. Thanks @Steffo99 anyway, and I'd like to invite you to keep suggestions around around the docs formatting coming! I'm sure there is room for improvement with Alabaster.

PR #4645, opened 2023-01-20
Improve formatting in the `plugins/plexupdate` documentation page

Steffo99:
## Description

This PR tries to improve formatting on the `plugins/plexupdate` documentation page.

## To Do

_None of the checkboxes seems applicable to this pull request._

### DIFF
--- a/docs/plugins/plexupdate.rst
+++ b/docs/plugins/plexupdate.rst
@@ -37,34 +37,15 @@ Configuration
 
 The available options under the ``plex:`` section are:
 
-``host``
-   The Plex server name.
-
-   Default: ``"localhost"``.
-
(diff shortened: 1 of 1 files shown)


sampsyo:
Hello again! Adding language declarations to the code blocks seems perfectly reasonable if you're interested, but the list of options is a bit more of a bugaboo. The other plugin pages tend to use a pretty consistent formatting, so I'd be wary of just changing one without changing the rest. And there is a trade-off here: adding separate paragraphs for each bullet will make the page longer and less dense… maybe it's worth being careful about whether this improves readability/scannability?

sampsyo:
Looks great; thanks!!

PR #4647, opened 2023-01-20
Use definition lists with the plugins' names in the documentation

Steffo99:
## Description

This PR changes the list of plugins in the documentation:

* it converts the *unordered lists* of plugins into *definition lists*, more appropriate for this kind of listing
* it changes the terms of the lists from the documentation page name to the relative plugin's id, allowing easier search when the exact plugin name is known

## To Do

_None of the checkboxes seems applicable to this pull request._

### DIFF
--- a/docs/plugins/index.rst
+++ b/docs/plugins/index.rst
@@ -132,13 +132,21 @@ following to your configuration::
 Autotagger Extensions
 ---------------------
 
-* :doc:`chroma`: Use acoustic fingerprinting to identify audio files with
-  missing or incorrect metadata.
-* :doc:`discogs`: Search for releases in the `Discogs`_ database.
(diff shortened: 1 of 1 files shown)


sampsyo:
Neato! Any chance you could share a screenshot of what the definition list looks like?

Steffo99:
<details>
<summary>Screenshot</summary>

Screenshot of the plugin list page.

</details>

As you mentioned in #4645, definition lists *do* make the less dense, but I think in this case it is a positive, as the plugins are many, and this change in my opinion makes the list easier to read.

sampsyo:
Nice!!! I really like the way this looks. I honestly didn't know ReST had support for definition lists; this is great!

PR #4649, opened 2023-01-22
Add typings to the art resizer module

Serene-Arc:
Continuing my effort to type the modules of beets, here's another module typed. Few comments:

- Pillow's Image module is imported at the top so I could use the typings. There are other duplicate imports in the code; the tests run so it doesn't cause problems there but I don't know why it was done in the first place so worth noting
	- If it's fine, then there are duplicate imports that should be removed
- @wisp3rwind said that beets considers every path beyond the core to be a string, or whatever needed by external APIs. I have here typed them to be `AnyStr` which includes both bytes and strings for maximum breadth, but if they are truly strings, then this can be tightened.

### DIFF
--- a/beets/util/artresizer.py
+++ b/beets/util/artresizer.py
@@ -23,6 +23,8 @@ import os.path
 import platform
 import re
 from tempfile import NamedTemporaryFile
+from typing import AnyStr, Tuple, Optional, Mapping, Union
+from PIL import Image
 from urllib.parse import urlencode
 from beets import logging
 from beets import util
@@ -33,7 +35,7 @@ PROXY_URL = 'https://images.weserv.nl/'
 log = logging.getLogger('beets')
 
 
-def resize_url(url, maxwidth, quality=0):
+def resize_url(url: str, maxwidth: int, quality: int = 0) -> str:
     """Return a proxied image URL that resizes the original image to
     maxwidth (preserving aspect ratio).
     """
@@ -48,7 +50,7 @@ def resize_url(url, maxwidth, quality=0):
     return '{}?{}'.format(PROXY_URL, urlencode(params))
 
 
-def temp_file_for(path):
+def temp_file_for(path: AnyStr):
     """Return an unused filename with the same extension as the
     specified path.
     """
@@ -131,8 +133,14 @@ class IMBackend(LocalBackend):
             self.identify_cmd = ['magick', 'identify']
             self.compare_cmd = ['magick', 'compare']
 
-    def resize(self, maxwidth, path_in, path_out=None, quality=0,
-               max_filesize=0):
+    def resize(
+        self,
+        maxwidth: int,
+        path_in: AnyStr,
+        path_out: Optional[AnyStr] = None,
+        quality: int = 0,
+        max_filesize: int = 0,
+    ) -> AnyStr:
         """Resize using ImageMagick.
 
         Use the ``magick`` program or ``convert`` on older versions. Return
@@ -172,7 +180,7 @@ class IMBackend(LocalBackend):
 
         return path_out
 
-    def get_size(self, path_in):
+    def get_size(self, path_in: str) -> Optional[Tuple[int, ...]]:
         cmd = self.identify_cmd + [
             '-format', '%w %h', syspath(path_in, prefix=False)
         ]
@@ -193,7 +201,7 @@ class IMBackend(LocalBackend):
             log.warning('Could not understand IM output: {0!r}', out)
             return None
 
-    def deinterlace(self, path_in, path_out=None):
+    def deinterlace(self, path_in: AnyStr, path_out: Optional[AnyStr] = None) -> AnyStr:
         path_out = path_out or temp_file_for(path_in)
 
         cmd = self.convert_cmd + [
@@ -209,7 +217,7 @@ class IMBackend(LocalBackend):
             # FIXME: Should probably issue a warning?
             return path_in
 
-    def get_format(self, filepath):
+    def get_format(self, filepath: AnyStr) -> Optional[bytes]:
         cmd = self.identify_cmd + [
             '-format', '%[magick]',
             syspath(filepath)
@@ -221,7 +229,12 @@ class IMBackend(LocalBackend):
             # FIXME: Should probably issue a warning?
             return None
 
-    def convert_format(self, source, target, deinterlaced):
+    def convert_format(
+            self,
+            source: AnyStr,
+            target: AnyStr,
+            deinterlaced: bool,
+    ) -> AnyStr:
         cmd = self.convert_cmd + [
             syspath(source),
             *(["-interlace", "none"] if deinterlaced else []),
@@ -240,10 +253,15 @@ class IMBackend(LocalBackend):
             return source
 
     @property
-    def can_compare(self):
+    def can_compare(self) -> bool:
         return self.version() > (6, 8, 7)
 
-    def compare(self, im1, im2, compare_threshold):
+    def compare(
+            self,
+            im1: Image,
+            im2: Image,
+            compare_threshold: float,
+    ) -> Optional[bool]:
         is_windows = platform.system() == "Windows"
 
         # Converting images to grayscale tends to minimize the weight
@@ -309,10 +327,10 @@ class IMBackend(LocalBackend):
         return phash_diff <= compare_threshold
 
     @property
-    def can_write_metadata(self):
+    def can_write_metadata(self) -> bool:
         return True
 
-    def write_metadata(self, file, metadata):
+    def write_metadata(self, file: AnyStr, metadata: Mapping):
         assignments = list(chain.from_iterable(
             ('-set', k, v) for k, v in metadata.items()
         ))
@@ -338,8 +356,14 @@ class PILBackend(LocalBackend):
         """
         self.version()
 
-    def resize(self, maxwidth, path_in, path_out=None, quality=0,
-               max_filesize=0):
+    def resize(
+            self,
+            maxwidth: int,
+            path_in: AnyStr,
+            path_out: Optional[AnyStr] = None,
+            quality: int = 0,
+            max_filesize: int = 0,
+    ) -> AnyStr:
         """Resize using Python Imaging Library (PIL).  Return the output path
         of resized image.
         """
@@ -396,7 +420,7 @@ class PILBackend(LocalBackend):
                       displayable_path(path_in))
             return path_in
 
-    def get_size(self, path_in):
+    def get_size(self, path_in: AnyStr) -> Optional[Tuple[int, int]]:
         from PIL import Image
 
         try:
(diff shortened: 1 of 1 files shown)


wisp3rwind:
> - Pillow's Image module is imported at the top so I could use the typings. There are other duplicate imports in the code; the tests run so it doesn't cause problems there but I don't know why it was done in the first place so worth noting
>       
>   - If it's fine, then there are duplicate imports that should be removed

The purpose of the delayed imports is to only require these libraries if they actually end up being used. See the `__init__` methods on the two backend classes. Thus, this should stay as is. For typing, we could maybe hide this behind a `https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING` check?

> - @wisp3rwind said that beets considers every path beyond the core to be a string, or whatever needed by external APIs. I have here typed them to be `AnyStr` which includes both bytes and strings for maximum breadth, but if they are truly strings, then this can be tightened.

Can be both, since on Linux, the bytes are directly passed on (via `syspath` returning early in that case).

beets/util/artresizer.py
```diff
@@ -240,10 +259,15 @@ def convert_format(self, source, target, deinterlaced):
...
 
     @property
-    def can_compare(self):
+    def can_compare(self) -> bool:
         return self.version() > (6, 8, 7)
 
-    def compare(self, im1, im2, compare_threshold):
+    def compare(
+            self,
+            im1: Image,
```
ybnd, line 267: Looks like leaving these type hints like this causes tox tests to fail; I would have expected the hints to be ignored, but apparently they're not.

Typing these as `'Image'` seems to fix this.

beets/util/artresizer.py
```diff
@@ -66,7 +70,7 @@ class LocalBackendNotAvailableError(Exception):
 
 class LocalBackend:
     @classmethod
-    def available(cls):
+    def available(cls) -> bool:
         try:
             cls.version()
```
ybnd, line 75: We should have an abstract method `LocalBackend.version()` for this line to pass the type check.

beets/util/artresizer.py
```diff
@@ -84,7 +88,7 @@ class IMBackend(LocalBackend):
     _legacy = None
 
     @classmethod
-    def version(cls):
+    def version(cls) -> Optional[Union[object, Tuple[int, int, int]]]:
```
ybnd, line 91: This may be out of scope for a typing PR, but I think this annotation is a bit much; if anything it highlights that the version logic is a bit hard to work with.

I'd suggest
* Some other way to handle "no version available" instead of the raw object instance
  * `None` would be way more clear
  * Could allow for that by changing the check below to `not hasattr(...)` instead of `... is None`
* Replace the "raw tuple" with a `NamedTuple`

All in all this would simplify those annotations by a lot:

```py
class Version(NamedTuple):
  major: int
  minor: int
  patch: int


class IMBacked(LocalBackend):
  @classmethod
  def version(cls) -> Optional[Version]:
    ...
```

beets/util/artresizer.py
```diff
@@ -240,10 +259,15 @@ def convert_format(self, source, target, deinterlaced):
             return source
 
     @property
-    def can_compare(self):
+    def can_compare(self) -> bool:
         return self.version() > (6, 8, 7)
```
ybnd, line 263: Type checker flags this because it doesn't take into account that the version may be non-existent.
The original signature is a bit too complicated, see my previous comment.

beets/util/artresizer.py
```diff
@@ -309,10 +333,10 @@ def compare(self, im1, im2, compare_threshold):
         return phash_diff <= compare_threshold
 
     @property
-    def can_write_metadata(self):
+    def can_write_metadata(self) -> bool:
         return True
 
-    def write_metadata(self, file, metadata):
+    def write_metadata(self, file: AnyStr, metadata: Mapping):
         assignments = list(chain.from_iterable(
```
ybnd, line 340: Looks like something about this makes it "untypeable", I think we basically have no choice but to explicitly type it as `Any`...

```
beets/util/artresizer.py:344: error: Cannot determine type of "assignments"  [has-type]
```

Doesn't seem liek too big of a shortcut to me given that `util.command_output` is untyped anyway.

ybnd (changes requested):
Did some local tests/type checks on this branch, highlighted some problems.

Other than the comments there are still a few minor problems that need to be addressed too:
* Some `Optional` values where `None` is not taken into account
* Some `str` vs. `bytes` mismatches

beets/util/artresizer.py
Serene-Arc: There's steps that need to be taken to fix circular imports. I do that as the last thing because it's a gigantic pain. Don't worry about failing tests; at this stage the typings are what matter.

beets/util/artresizer.py
Serene-Arc: I think it is out of scope. I don't think I should be changing the logic of the methods I'm typing in this PR. If you feel differently then we can put changes in motion but I feel like there should be a different PR for those. There's a great deal of places where things could be changed or made simpler, but there's only so many changes before things become unmanageable.

beets/util/artresizer.py
Serene-Arc: If you're using mypy to get these hints, then you should know that I have been working under the understanding that mypy is not really relevant at this stage. It is much more strict and even without this PR, throws a huge number of errors. See [here for details](https://github.com/beetbox/beets/pull/4495#issuecomment-1328068170). 

That MyPy gives this error doesn't mean it's true. It doesn't mean that it is actually `Any`. Those are the typings that are presented by the codebase and those ones I'm sure of.

beets/util/artresizer.py
Serene-Arc: Mypy is not usable at the current stage as an indicator of typing correctness. See next comment.

beets/util/artresizer.py
Serene-Arc: Is that in scope for this typing PR or a comment on something that needs to be done at another time?

beets/util/artresizer.py
wisp3rwind: Well, I wouldn't say mypy is completely irrelevant, it can definitely provide useful input.

What is certainly the case that having zero mypy issues left is not a requirement for a PR like this to be merged, since that might simply not be achievable without substantial refactoring to resolve issues that mypy reveals.

An example would be the `version()` problem mentioned in another comment: The refactoring proposed by @ybnd looks like a great idea, but I also agree that we might want to hold it off for another PR in order to avoid blowing up this one too much (although, in this case, I'd be fine either way: the PR is not that huge yet).

---

As to the particular issue, I'm somewhat surprised that mypy cannot resolve this to at least a `List[Any]`, which would be enough to verify that the list unpacking in the argument list is valid. Maybe explicitly annotating it as `List[Any]` would be useful? Or, maybe specifying the `Mapping` in the argument more concretely. I suspect this should be `Mapping[str, Any]`, maybe even `Mapping[str, str]` (but verifying the latter would require some research into the backend methods).

beets/util/artresizer.py
wisp3rwind: Up to you, but since this just means adding a

```python
def version(cls) -> ...:
    raise NotImplementedError()
```

I think it's not such a big change that would make this PR intractable. (Actually making `LocalBackend` an `abc.ABC` and using the `@abstractmethod` decorator would be some more work, but I guess with the above stopgap solution we already get quite far, and the `@abstractmethod` way would require it anyway.)

beets/util/artresizer.py
```diff
@@ -172,7 +182,7 @@ def resize(self, maxwidth, path_in, path_out=None, quality=0,
 
         return path_out
 
-    def get_size(self, path_in):
+    def get_size(self, path_in: str) -> Optional[Tuple[int, ...]]:
```
wisp3rwind, line 185:
```suggestion
    def get_size(self, path_in: bytes) -> Optional[Tuple[int, int]]:
```

Note that `path_in` is passed through `syspath`, i.e. it is a path in beets-internal bytestring-representation.

The output type should match the PIL backend. Probably, we should eventually make the code here validate the IM result completely, including the being exactly length 2 (just catching the `IndexError` is definitely not enough (if such an error can even happen?)).

beets/util/artresizer.py
```diff
@@ -209,7 +223,7 @@ def deinterlace(self, path_in, path_out=None):
             # FIXME: Should probably issue a warning?
             return path_in
 
-    def get_format(self, filepath):
+    def get_format(self, filepath: AnyStr) -> Optional[bytes]:
```
wisp3rwind, line 226: Probably out-of scope here; but this code seems to be broken: All backends should really have the same return type, so we should probably be decoding the stdout to a `str` here.

beets/util/artresizer.py
```diff
@@ -430,7 +464,12 @@ def get_format(self, filepath):
             log.exception("failed to detect image format for {}", filepath)
             return None
 
-    def convert_format(self, source, target, deinterlaced):
+    def convert_format(
+            self,
+            source: AnyStr,
+            target: AnyStr,
+            deinterlaced: bool,
+    ) -> str:
```
wisp3rwind, line 472:
```suggestion
    ) -> AnyStr:
```

given that this returns either of `source` or `target`.

On second thought: Probably all of these should be `bytes` (again, these are paths in beets internal representation, as being evidenced by the `syspath` call).

beets/util/artresizer.py
```diff
@@ -557,13 +610,13 @@ def proxy_url(self, maxwidth, url, quality=0):
...
     @property
-    def local(self):
+    def local(self) -> bool:
         """A boolean indicating whether the resizing method is performed
         locally (i.e., PIL or ImageMagick).
         """
         return self.local_method is not None
 
-    def get_size(self, path_in):
+    def get_size(self, path_in: AnyStr) -> Union[Tuple[int, int], AnyStr]:
```
wisp3rwind, line 619: Can you correct this method to return `None` in the `else` branch? What's currently going on is just wrong (probably due to my own fault... the git blame is somewhat confusing, not sure what happened. Maybe a botched merge? copy&paste bug?)

wisp3rwind (changes requested):
Just had a look through this (notably, I didn't actually run any type checker on it!), and added a few comments.

As @ybnd  already pointed out, there are some inconsistencies with the design of the backend classes. Adding an abstract `Backend` base class seems like a good way to get a hold on them. However, I'd say let's keep that for a follow-up.

beets/util/artresizer.py
Serene-Arc: I actually intend to start my grand project to convert to pathlib today on the core modules, so this should be fixed soon regardless.

beets/util/artresizer.py
Serene-Arc: I don't think there's a way to conditionally return a single type. I can add it to be `Optional` though so it recognises that a null value is possible.

wisp3rwind: Sorry, what I meant is that this should return `Optional[Tuple[int, int]]`, and in the else case `return None` instead of `return path_in`.

wisp3rwind:
Hi @Serene-Arc, I just came across this and was a little surprised that this never made it over the finish line: Are you still interested in working on this? If not, would you mind if I picked this up?

Serene-Arc:
@wisp3rwind by all means, you can take over! I'm not entirely sure what stopped me from completing this. I've been a bit slammed with my honours studies of late.

wisp3rwind:
Closing in favor of #5734.

PR #4663, opened 2023-01-31
Fix album store method to cascade flex field deletions to items

### LINKED ISSUE #4662: Removing a flexible field from an album doesn't remove corresponding field from its items
### Problem

When importing an album with the `--set` flag, beets imports the album with "flex1" set on both the album and each of its tracks:

```sh
$ beet import <path> --set flex1=something
```

Later, when deleting the flex field with `-a`, the value is updated on the album, but is still set on the items:

```sh
$ beet modify -a "Some album" flex1!
```

After modifying, search for matching tracks, and they still return

```sh
$ beet ls flex1::something
```
```
Some artist - Some album - Track name 1
Some artist - Some album - Track name 2
...
```

Search for matching albums, returns nothing

```sh
$ beet ls -a flex1::something
```
```sh
# no results
```

### Setup

* OS: Arch
* Python version: 3.10.9
* beets version: 1.6.1

```
(tested with default settings)
```

### Extra info

Also of note, this issue naturally doesn't come up if the flex field was added to just the album after import using `beet mod -a flex1=something` which is how I configured a lot of my library, so this issue only happens to albums imported with `import --set`

katedickey:
## Description

Fixes #4662.

## To Do

- [ ] N/A ~Documentation. (If you've add a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)~

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -1355,9 +1355,12 @@ class Album(LibModel):
         """
         # Get modified track fields.
         track_updates = {}
-        for key in self.item_keys:
-            if key in self._dirty:
+        track_deletes = set()
+        for key in self._dirty:
+            if key in self.item_keys:
                 track_updates[key] = self[key]
+            elif key not in self:
+                track_deletes.add(key);
 
         with self._db.transaction():
             super().store(fields)
@@ -1365,6 +1368,12 @@ class Album(LibModel):
                 for item in self.items():
                     for key, value in track_updates.items():
                         item[key] = value
+                    item.store()
+            if track_deletes:
+                for item in self.items():
+                    for key in track_deletes:
+                        if key in item:
+                            del item[key]
                     item.store()
 
     def try_sync(self, write, move):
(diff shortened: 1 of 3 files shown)


sampsyo (approved):
Wow; this is a tricky issue! Great work tracking this down to the root cause.

Reading over the new policy in `Album.store`, this can look surprising at first, but it seems like the right thing. Namely, to spell it out in English, the idea is that the album looks for its fields that are *both dirty and missing* and then deletes those things from the item. It would be incorrect to look for fields that are *only missing* on the album, because it's possible for items to be harmlessly "out of sync" with their albums in that way. But when the field is _also_ dirty, then that indicates and update that we want to propagate to the items—in the same way we do as when the field's value has just changed.

Let's let the CI run just in case, but I think this is the right change and should be mergeable once the CI is green.

sampsyo (commented):
Ah, flake8 has a couple of minor style suggestions. Any chance you could give those a quick look, @katedickey?

sampsyo:
That was fast!! Looks great from here.

PR #4677, opened 2023-02-17
Fix cover art archive fetching

trolley:
## Description

PR #3748 changed the way cover art is fetched from the cover art archive, but the manual addition of a `-` to the width suffix that was needed when the image URI was being constructed manually was not removed. Because of this the plugin would try to look up the property under `thumbnails` that didn't exist (for example `-1200` instead of `1200`), which would fail.

## To Do


sampsyo:
Nice find; thanks!!

PR #4678, opened 2023-02-19
Docs: Add a link to the walkthrough from the main tutorial guide

sampsyo:
Supersedes #4382, with a somewhat simpler set of context around the link. Also fixes a link to Twitter (we're on Mastodon now).


wisp3rwind:
Looks good to me!

PR #4679, opened 2023-02-19
typing: fixes according to mypy for autotag.*

wisp3rwind:
Follow-up to #4588.

### DIFF
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -19,7 +19,7 @@ from collections import namedtuple
 from functools import total_ordering
 import re
 from typing import Dict, List, Tuple, Iterator, Union, NewType, Any, Optional,\
-    Iterable, Callable, TypeVar
+    Iterable, Callable, TypeVar, cast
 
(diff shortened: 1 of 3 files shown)


sampsyo (approved):
All looks great! Really nice work sorting out all the places to put in `cast` calls.

wisp3rwind:
> All looks great! Really nice work sorting out all the places to put in `cast` calls.

Well, at least they made mypy happy... I'm not entirely sure that this is the canonical way to solve the problem: At least for `confuse`, the proper solution would be to also type that eventually (the various getters should have well-defined return types, so that should be quite doable). The `Item`  and `Album` field accesses seem to be more tricky: In this case, the type really can't be determined statically.

sampsyo:
Indeed; it would be great to add proper type annotations to the Confuse library. They are a somewhat funky and dynamic but hopefully not too bad.

And yes, the core model classes are a much more… interesting problem. I imagine there is something reasonable to be done with them, but it will take some creativity.

PR #4690, opened 2023-03-04
Improve "Changelog goes here note" in changelog.rst

JOJ0:
## Description

Added a note to the very top of the changelog.rst that editors should add there "new things" to the bottom to the list. Some people add it to the top, some to the bottom. Basically it doesn't really matter in the end where a feature/fix is placed in the final release notes, but since we constantly have to resolve merge conflicts within this file I think it would reduce confusion at least a little if we know that "news" are at the bottom _always_.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,7 +4,7 @@ Changelog
 1.6.1 (in development)
 ----------------------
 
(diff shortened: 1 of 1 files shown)


sampsyo:
Sure; seems like a good idea!

PR #4698, opened 2023-03-09
discogs_client version check

ghbrown:
## Description

Addresses #4691, providing a warning if the user attempts to use a non-compliant version of `discogs_client` (any version of the abandoned package `discogs-client` or `python3-discogs-client` < 2.3.10). These non-compliant versions give no error when running `beet import`, but instead manifest as a large number of `No matching records found` due to rate limiting (which is expected to be implemented by `discogs_client`).

## To Do

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -111,7 +111,7 @@ setup(
             'requests_oauthlib',
             'reflink',
             'rarfile',
-            'python3-discogs-client',
+            'python3-discogs-client>=2.3.15',
             'py7zr',
         ],
         'lint': [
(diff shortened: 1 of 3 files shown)


JOJ0:
Hi @ghbrown, thank you very much for the PR, this is great! I am only wondering if we should, while we are at it, require an even higher version. I jhust recalled that there where reports that came out of the beets community about a bytes vs str error around a year ago: https://github.com/joalla/discogs_client/issues/80

It took us a while to fix it but we did in this PR: https://github.com/joalla/discogs_client/pull/94

It was released in 2.3.15: https://github.com/joalla/discogs_client/releases/tag/v2.3.15

JOJ0:
Looked through release notes and I don't see a reason why 2.3.15 minimum could be problematic.

Also please increase the version requirement of python3-discogs-client in `setup.py` please!

setup.py
```diff
@@ -111,7 +111,7 @@ def build_manpages():
             'requests_oauthlib',
             'reflink',
             'rarfile',
-            'python3-discogs-client',
+            'python3-discogs-client>=2.3.15',
```
ghbrown, line 114: Made the requested changes and also added a minimum version for the `extras_require 'test'`. Not sure why there wasn't one before, but if there's a reason I'm happy to revert.

sampsyo:
That seems like a very good idea. Thank you!!

PR #4699, opened 2023-03-10
Fix typo in beetsplug/deezer.py

arogl:
Found another typo


JOJ0:
It seems I do not love the deezer plugin.... :-/

arogl:
All OK, I missed it as well, even when trying to use deezer IDs

PR #4705, opened 2023-03-12
Add instructions to install directly from Github.

arsaboo:
Since the release is getting delayed, I think it may be worthwhile to add installation instructions directly from GitHub.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -80,7 +80,7 @@ shockingly simple if you know a little Python.
 Install
 -------
 
-You can install beets by typing ``pip install beets`` or directly from Github (see details [here](https://beets.readthedocs.io/en/stable/faq.html#run-the-latest-source-version-of-beets)).
+You can install beets by typing ``pip install beets`` or directly from Github (see details [here](https://beets.readthedocs.io/en/latest/faq.html#run-the-latest-source-version-of-beets)).
 Beets has also been packaged in the `software repositories`_ of several
 distributions. Check out the `Getting Started`_ guide for more information.
 
--- a/README.rst
+++ b/README.rst
@@ -80,7 +80,7 @@ shockingly simple if you know a little Python.
 Install
 -------
 
-You can install beets by typing ``pip install beets`` or directly from github using ``python -m pip install git+https://github.com/beetbox/beets.git``. command.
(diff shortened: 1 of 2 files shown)


JOJ0:
Hi @arsaboo that is actually a very good idea for improving the install documentation!

Please do me a favor and have a look into the developers documentation on read the docs. I think something similar is in there already. We could additionally link to that. Thanks in advance!

arsaboo:
See now.

JOJ0:
Hi, this is looking good so far. I'm to lazy to check it out and generate docs myself. Can you do me a favor and build docs locally and provide a screenshot snippet of your locally generated docs? That would help a lot! Thanks so much!

Not sure if building docs you find in the docs ;-), but it should be as simple as `cd docs; make html`, then open build/index.html or something like that.

arsaboo:
Not sure I follow you here. I don't recall doing anything like that in the past for doc changes. I did the `make html` part, but not sure about the next step.

JOJ0:
> Not sure I follow you here. I don't recall doing anything like that in the past for doc changes. I did the `make html` part, but not sure about the next step.

Looking at the docs in there rendered form would help me to understand if everything is in the right placce and looking good :-)

Just open the file you just rendered, usually you find an index.html in the docs build directory, I think its docs/build/index.html. Open that with your browser and send me a screenshot snippet of the parts you added. Thanks!

docs/faq.rst
```diff
@@ -137,6 +137,8 @@ it's helpful to run on the "bleeding edge". To run the latest source:
    ``pip uninstall beets``.
 2. Install from source. Choose one of these methods:
 
+   -  Directly from github using
```
JOJ0, line 140: Spell as "GitHub" when in regular sentence and not in a link. Thanks!

arsaboo:
Here's how it appears in preview. 

image

README.rst
```diff
@@ -80,7 +80,7 @@ shockingly simple if you know a little Python.
 Install
 -------
 
-You can install beets by typing ``pip install beets``.
+You can install beets by typing ``pip install beets`` or directly from Github (see details [here](https://beets.readthedocs.io/en/stable/faq.html#run-the-latest-source-version-of-beets)).
```
JOJ0, line 83: Point to latest docs instead of stable docs here, otherwise your change will not be shown.

We currently have stable as the default for beets.readthedocs.io, that is IMO another problem / thing to be discussed currently.

JOJ0 (changes requested):

Issue #4711, opened 2023-03-14
Add option to specify which formats to convert


arsaboo:
The `convert` plugin currently uses `no_convert` to specify which formats to not convert. Given the multiple audio formats, it may be difficult to specify all the formats. Instead, it will be nice to be able to specify the formats to convert. 

### Proposed solution

We should add another option `convert_format` to specify the formats to be converted. This will not be a breaking change as the current configuration will work as is. We will just introduce another option to specify the formats to be converted. 

For example, if I want to only convert mp4 files, the current option requires me to list all the formats that I DON'T want to convert, e.g., `no_convert: path::\.(flac|ogg|m4a|mp3)$` (and I may still miss a few formats). Instead, it will be a lot easier to just specify `convert_format::\.(mp4)$`
`

jackwilsdon:
It seems like this should already be achievable with [query term negation](https://beets.readthedocs.io/en/stable/reference/query.html#query-term-negation):

```
no_convert: ^path::\.(flac|ogg|m4a|mp3)$
```

I'll admit it's a bit hard to understand as it's a double negative, but maybe we could just document this instead of introducing a new option?

arsaboo:
Oh...love it. Created a PR to add that example.

PR #4712, opened 2023-03-15
Add additional example to convert plugin 

arsaboo:
## Description

Fixes https://github.com/beetbox/beets/issues/4711

Add another example to `convert` plugin to allow specifying a format to convert.


sampsyo:
Looks valuable; thank you!! :sparkles:

PR #4714, opened 2023-03-15
resolve transl-tracklisting relations for pseudo releases

### LINKED ISSUE #654: Follow MusicBrainz pseudo-releases
The MB database uses "pseudo-releases" to provide translations of track names when there was, in fact, never a translated release. Apparently, [the correct way to handle these](http://tickets.musicbrainz.org/browse/PICARD-145) is to use the original release's IDs but the pseudo-release's human-readable metadata. The autotagger would need to follow the [translation/transliteration relationship](http://musicbrainz.org/relationship/fc399d47-23a7-4c28-bfcf-0607a562b644) to the original release.

TypicalFence:
## Description

Fixes #654 

As in the issue suggested I implemented the "easiest/most obvious approach".

It fetches release relations and "follows" the ones for the type [transl-tracklisting](https://musicbrainz.org/relationship/fc399d47-23a7-4c28-bfcf-0607a562b644) for pseudo releases. The additional data gets merged with the pseudo-release. I am not 100% sure if I included all applicable fields for the merge. I left the pseudo-releases id's as is.

This change makes it near impossible to differentiate the pseudo releases from the actual releases in the ui, so I added that to `disambig_string`.  Are there any unwanted side effects from that change?

A small screenshot for reference:
Screenshot 2023-03-15 at 22 42 24

This implementation has almost the same approach as the [plugin](https://github.com/TypicalFence/beets-resolve-pesudo-releases) I wrote recently. Said plugin hasn't caused any issues for me yet. This PR handles things much cleaner tho (it doesn't have to fetch pseudo-releases twice for instance).

I feel like my code ended up being a bit awkward, I haven't written any serious python in a long while and pep8 feels a bit tedious. 

## To Do

- [ ] Add a Setting for people who really don't care about this?

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -73,7 +73,7 @@ log = logging.getLogger('beets')
 RELEASE_INCLUDES = ['artists', 'media', 'recordings', 'release-groups',
                     'labels', 'artist-credits', 'aliases',
                     'recording-level-rels', 'work-rels',
-                    'work-level-rels', 'artist-rels', 'isrcs', 'url-rels']
+                    'work-level-rels', 'artist-rels', 'isrcs', 'url-rels', 'release-rels']
 BROWSE_INCLUDES = ['artist-credits', 'work-rels',
                    'artist-rels', 'recording-rels', 'release-rels']
 if "work-level-rels" in musicbrainzngs.VALID_BROWSE_INCLUDES['recording']:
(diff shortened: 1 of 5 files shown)


beets/autotag/mb.py
```diff
@@ -607,6 +607,61 @@ def _parse_id(s: str) -> Optional[str]:
...
+
+def _find_actual_release_from_pseudo_release(pseudo_rel: Dict)\
+      -> Optional[Dict]:
+    relations = pseudo_rel['release']["release-relation-list"]
+
+    # currently we only support trans(liter)ation's
+    actual_id = next(filter(_is_trans, relations), {'target': None})['target']
+
+    if actual_id is None:
+        return None
```
sampsyo, line 623: I think this could be a teensy bit simpler so it's easier to read/maintain. Something like this?

```py
translations = [r for r in relations if _is_trans(r)]
if not translations:
    return None
actual_id = translations[0]['target']
```

beets/autotag/mb.py
```diff
@@ -607,6 +607,61 @@ def _parse_id(s: str) -> Optional[str]:
...
+        "month": actual.month,
+        "day": actual.day,
+        "original_year": actual.original_year,
+        "original_month": actual.original_month,
+        "original_day": actual.original_day,
+        "label": actual.label,
+        "asin": actual.asin,
+        "style": actual.style,
+        "genre": actual.genre,
+    })
```
sampsyo, line 661: I don't know if this helps much, be we could reduce some of the typing here by just using a list of fields. That is:

```py
from_actual = {k: actual[k] for k in ["media", "mediums", "country", ...]}
merged.update(from_actual)
```

beets/autotag/mb.py
```diff
@@ -620,13 +675,27 @@ def album_for_id(releaseid: str) -> Optional[beets.autotag.hooks.AlbumInfo]:
...
         log.debug('Album ID match failed.')
         return None
     except musicbrainzngs.MusicBrainzError as exc:
         raise MusicBrainzAPIError(exc, 'get release by ID', albumid,
                                   traceback.format_exc())
-    return album_info(res['release'])
+
+    release = album_info(res['release'])
+
+    if actual_res is not None:
```
sampsyo, line 694: Maybe this needs a comment to explain that this is for pseudo-releases.

beets/ui/commands.py
```diff
@@ -211,6 +211,9 @@ def disambig_string(info):
             disambig.append(info.catalognum)
         if info.albumdisambig:
             disambig.append(info.albumdisambig)
+        # pseudo releases can't be differentiated from real release otherwise
```
sampsyo, line 214: Or, to be slightly more direct:

> Let the user differentiate between pseudo and actual releases.

sampsyo (commented):
Super cool; thank you for getting this started!!! Here are just a few code-level suggestions.

beets/autotag/mb.py
```diff
@@ -607,6 +607,61 @@ def _parse_id(s: str) -> Optional[str]:
     return None
 
 
+# this was defined within the function below but pep8 made me move it here
```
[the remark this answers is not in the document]
TypicalFence: yea defining a function is probably much better

beets/autotag/mb.py
TypicalFence: this for syntax is pretty new to me tbh, I've added it to make adding fields less painful

beets/autotag/mb.py
TypicalFence: added comments for there

beets/ui/commands.py
TypicalFence: updated that as well

sampsyo:
Awesome! It looks like there is one minor conflict with a recent change to `RELEASE_INCLUDES`—any chance you could take a shot at resolving that? Then perhaps we'll have a clean merge.

TypicalFence:
@sampsyo I'll try to get around rebasing soon

TypicalFence:
hey @sampsyo sorry that it took months but I've rebase the branch

sampsyo (changes requested):
Awesome; this is looking great!! Could you please just add a really brief changelog entry? There is also a super minor lint thing (a long line that will be easy to wrap).

sampsyo (approved):
Many thanks again for tracking this down!

PR #4717, opened 2023-03-17
OLD Support providing album information on singleton imports

### LINKED ISSUE #4716: discogs: Save album information on singleton imports
When the discogs plugin is used for singleton imports consider additionally saving album information to the beets library.

### Proposed solution

On singleton imports the user specifically selects a Discogs release (Discogs release URL) where the track information should be taken from. This could be saved to the library field `album` of the track item (`album` row in the `items` table in the database) and to the corresponding metadata tag of the file.

Let's also compare this feature idea to a singleton import against the MusicBrainz database: Here we have a distinction between an album and a recording. A recording / a track is its own entity and does have a 1:n connection to an album, thus by selecting a recording on a singleton import, we don't know which album we are referring to! A feature like that would not make sense with MusicBrainz. With Discogs on the other hand, we have a 1:1 connection from a track to an album.

#### Goals

Save to the track item's `album` field (`items` table, `album` column). Provide an easy way for subsequent re-tagging of such singletons to albums, by having the required album info already.

#### Non-goals

Adding `album` and `album_id` to the `album` table in the database.

JOJ0:
## Description

Implements a way to allow supplying album information on singleton imports and makes use of it in the Discogs metadata source plugin.

Resolves #4716

## To Do

- [ ] Documentation.
- [ ] Changelog.
- [ ] Tests.

### DIFF
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -727,7 +727,17 @@ def item_candidates(item: Item, artist: str, title: str) -> Iterable[Tuple]:
 
     # MusicBrainz candidates.
     if config["musicbrainz"]["enabled"] and artist and title:
-        yield from invoke_mb(mb.match_track, artist, title)
+        try:
+            for candidate in mb.match_track(artist, title):
+                yield TrackAlbumTuple(candidate, None)
+        except mb.MusicBrainzAPIError as exc:
+            exc.log(log)
 
     # Plugin candidates.
-    yield from plugins.item_candidates(item, artist, title)
+    for candidate in plugins.item_candidates(item, artist, title):
+        # Allow (track_info, album_info) tuples
+        plugins.send(u'trackinfo_received', info=candidate)
+        if isinstance(candidate, TrackAlbumTuple):
+            yield candidate
+        else:
+            yield TrackAlbumTuple(candidate, None)
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -613,7 +613,10 @@ class Distance:
 AlbumMatch = namedtuple('AlbumMatch', ['distance', 'info', 'mapping',
                                        'extra_items', 'extra_tracks'])
 
-TrackMatch = namedtuple('TrackMatch', ['distance', 'info'])
+TrackMatch = namedtuple('TrackMatch', ['distance', 'info', 'album_info'],
+                        defaults=[None, None, None])
+TrackAlbumTuple = namedtuple('TrackAlbumTuple', ['track_info', 'album_info'],
+                             defaults=[None, None])
 
 
 # Aggregation of sources.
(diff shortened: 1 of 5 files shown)


JOJ0:
@ghbrown you wann give this an early review/test? Especially along these lines I'm not sure if all of it makes sense to a user in a import session where having MB as well as Discogs results: https://github.com/beetbox/beets/pull/4717/files#diff-e324b20657a7d6b43b8b7aeb5754b96774f5062294b5ba7f1e3062845e9e7044R450-R468

ghbrown:
@JOJ0 I'm a bit swamped this week, and I also have been working with MusicBrainz turned off, so not sure I'm the best candidate here.

However, depending on how long this takes, I might be able to take a look this weekend.

JOJ0:
Weekend sounds great, also with MB turned off a review will be superhelpful. No need to hurry here, I would like this to become a wonderful feature that just helps. I'm sure you will like what you see. I also improved the disambiguation string so it shows which track actually would is selected e.g. in my case often I tag from vinyl records. You'd see something like this:

```
Correcting track tags from:
    Calibre - My Chances
To:
    Calibre - My Chances
    Album: Peso / My Chances
URL:
    https://www.discogs.com/release/1163433-Calibre-Peso-My-Chances
(Similarity: 50.0%) (id) (Discogs, Index 2, Track B )
Apply, More candidates, Skip, Use as-is, Enter search, enter Id, aBort,
plaY, eDit, edit Candidates? m
Finding tags for track "Calibre - My Chances".
Candidates:
1. Calibre - My Chances (100.0%) (Discogs, Index 2, Track b )
2. Calibre - My Chances (50.0%) (id) (Discogs, Index 2, Track B )
# selection (default 1), Skip, Use as-is, Enter search, enter Id, aBort,
plaY, eDit, edit Candidates? 
```

sampsyo:
Awesome; thanks for getting this going!

Here is one super basic question: could this be simpler if we just added an (optional) field to `TrackInfo` that points to an `AlbumInfo` object? That is, instead of creating and passing around a new `TrackAlbumTuple` type, we could stick with plain `TrackInfo` objects, but add a `TrackInfo.album_info` object that is allowed to be `None`. Would this make sense, and possibly simplify some of the "plumbing" that shuffles around these matches?

JOJ0:
> 

Well yeah, that sounds so much simpler! I'll give that idea a shot. Thanks a lot!

JOJ0:
> Well yeah, that sounds so much simpler! I'll give that idea a shot. Thanks a lot!

> Here is one super basic question: could this be simpler if we just added an (optional) field to `TrackInfo` that points to an `AlbumInfo` object? That is, instead of creating and passing around a new `TrackAlbumTuple` type, we could stick with plain `TrackInfo` objects, but add a `TrackInfo.album_info` object that is allowed to be `None`. Would this make sense, and possibly simplify some of the "plumbing" that shuffles around these matches?

Well yeah, that sounds so much simpler! I'll give that idea a shot. Thanks a lot!

PR #4723, opened 2023-03-20
Add JioSaavn to the community plugins list

arsaboo:
## Description

Added JioSaavn.com as a tag source.

I used Deezer and Beatport as the base for this plugin. It works well for me. I am happy to create a PR if we want to have this as part of beets.


JOJ0:
Since there is so much to do within beets, I think having it in the community list is the best way for now and probably makes things even easier for you :-) I'll merge this. Thanks! Always great to see a new plugin being added to the list!

PR #4728, opened 2023-03-25
play plugin: $playlist marker for precise control where the playlist …

cvx35isl:
…file is placed in the command

## Description

see included doc; placing the playlist filename at the end of command just isn't working for all players

I have this in use with `mpv`

## To Do

- [ ] Documentation. (done, let me know if more is needed)
- [ ] Tests. (let me know if needed)

### DIFF
--- a/beetsplug/play.py
+++ b/beetsplug/play.py
@@ -30,6 +30,11 @@ import shlex
 # If this is missing, they're placed at the end.
 ARGS_MARKER = '$args'
 
+# Indicate where the playlist file (with absolute path) should be inserted into
+# the command string. If this is missing, its placed at the end, but before
+# arguments.
+PLS_MARKER = '$playlist'
+
 
 def play(command_str, selection, paths, open_args, log, item_type='track',
          keep_open=False):
@@ -128,6 +133,15 @@ class PlayPlugin(BeetsPlugin):
 
         open_args = self._playlist_or_paths(paths)
         command_str = self._command_str(opts.args)
+
+        if PLS_MARKER in command_str:
+            if not config['play']['raw']:
+                ui.print_(ui.colorize('text_warning', 'DEBUG: ' + command_str))
+                command_str = command_str.replace(PLS_MARKER, ''.join(open_args))
+                ui.print_(ui.colorize('text_warning', 'DEBUG: ' + command_str))
+                open_args = []
+            else:
+                command_str = command_str.replace(PLS_MARKER, " ")
 
         # Check if the selection exceeds configured threshold. If True,
         # cancel, otherwise proceed with play command.
--- a/docs/plugins/play.rst
+++ b/docs/plugins/play.rst
@@ -48,8 +48,14 @@ To configure the plugin, make a ``play:`` section in your
 configuration file. The available options are:
 
 - **command**: The command used to open the playlist.
-  Default: ``open`` on OS X, ``xdg-open`` on other Unixes and ``start`` on
-  Windows. Insert ``$args`` to use the ``--args`` feature.
+
+  - Default: ``open`` on OS X, ``xdg-open`` on other Unixes and ``start`` on
+    Windows.
+  - Insert ``$args`` to use the ``--args`` feature.
+  - Insert ``$playlist`` to specify precisely where the playlist path should
+    appear in the command; only works if a playlist is generated, that is when
+    **raw** = ``no`` is configured.
+
 - **relative_to**: If set, emit paths relative to this directory.
   Default: None.
 - **use_folders**: When using the ``-a`` option, the m3u will contain the
@@ -97,6 +103,13 @@ example::
 
 indicates that you need to insert extra arguments before specifying the
 playlist.
+
+The above example, however, does not work with current ``mpv`` because the
+``--playlist`` argument wants a different syntax. To satisfy this, the optional
+``$playlist`` can be used to meet that::
+
+    play:
+        command: mpv $args --playlist=$playlist
 
 The ``--yes`` (or ``-y``) flag to the ``play`` command will skip the warning
 message if you choose to play more items than the **warning_threshold** 


beetsplug/play.py
```diff
@@ -129,6 +134,15 @@ def _play_command(self, lib, opts, args):
         open_args = self._playlist_or_paths(paths)
         command_str = self._command_str(opts.args)
 
+        if PLS_MARKER in command_str:
+            if not config['play']['raw']:
+                ui.print_(ui.colorize('text_warning', 'DEBUG: ' + command_str))
```
sampsyo, line 139: Seems like these should probably be proper debug logs… like `log.debug(...)`?

sampsyo (commented):
Looks good overall! But it would be nice to use proper debug logging so the messages are suppressed by default. Also, could you please add a quick changelog entry?

beetsplug/play.py
cvx35isl: done. totally overlooked this.

M1113R (approved):

sampsyo:
Cool; thanks for moving this along! It looks like there are a couple of suggestions from the style checker tool.

Also, it occurs to me after thinking about the code for a moment that there could be problems with shell escaping. Do you think it would make sense to pass each of the `open_args` values through [shlex.quote](https://docs.python.org/3/library/shlex.html#shlex.quote)? That would prevent issues when an argument has a space in it, for example.

cvx35isl:
thinking some more I'll probably have to unquote again right before `subprocess.call()`, since if its not a shell it won't do that. It would try to execute and play files that have quoting characters in their names ... and fail.

So quoting may really only be able to improve the splitting in case `config.play.raw=true` is used.

(tell me if I get this totally wrong - I dont do much with python, really)

sampsyo:
I think the main thing is that, when we eventually called `shlex.split` on the newly-interpolated command, we will want the filename to be preserved as a single shell token. It's true that `open_args` will typically not contain any weird characters because it's a generated `/tmp/...` name when `raw=False`, but I guess it seems like a good practice to make sure that no filename can ever trip up the command-line construction?

cvx35isl:
hello, sorry for the long delay, i was otherwise busy


this branch / PR is now rebased upon the current master and should just merge

I did more digging in how and where to make use of `shlex.quote()` and don't see it. Cause the `paths` list, especially when `raw=True` is configured, is composed by calling `Item.path` which returns a 'byte' object. `shlex.quote()` does not accept byte input, so I would need to stringify first...  but then I see `util.interactive_open()` is aware of this and somehow handles these bytes as Unicode or passes them on to whatever handles Unicode in the OS. I don't understand that part really, but all goes into `os.execlp()` where each path is handed over as one argN. No shell in sight.

If `raw=False` is configured the bytes get separated by newline and written straight into a .m3u file where its then up to the actual player program to properly read it.  In most cases this won't be a shell either unless the user specifically does so.

To my understanding there is no chance of unintentionally running into a escaping problem.

sampsyo:
Consider the case where the config is:

```yaml
play:
    command: mpv $playlist
```

So the `_command_str` function returns the string `"mpv $playlist"`. Eventually, we will call `shlex.split` on this string (after interpolating it) and then hand it off to `os.execlp`. So, if the playlist path is `/foo/bar`, then we will effectively call:

```py
os.execlp(['mpv', 'mpv', '/foo/bar'])
```

However, if the playlist path is `/foo/bar baz/qux`, then we will construct the string `"mpv /foo/bar baz/qux"` and *then* split it, so we will effectively call:

```py
os.execlp(['mpv', 'mpv', '/foo/bar', 'baz/qux'])
```

…and the command will see the wrong number of arguments. Does this make sense?

cvx35isl:
yes, that could happen if a space or something happens to be in that tmp dir path -> I added a `shlex.quote()` where that path gets created, before it gets into `command_str`.

I don't see how to properly quote the `beet play <query> --args ...` list if given, since it would need `shlex.split()` first and then quoting these parts would pursue the problem ... that must be done right by the user.

regarding the failed tests: think that wasn't me, some package server was unreachable

sampsyo:
Thanks! Since it seems like somewhere where things could easily go wrong, maybe it would be instructive to have a few tests to show that this is working correctly in any configuration setup we can think of?

cvx35isl:
how to implement these, in general?

i think the main imponderables are  `play.command:` and the library itself, esp. the paths of where it is on the system and how the user structured it using the `paths:` section. 

to test `play.command:`  a mock-player could be written that checks how it gets called and/or does wrong things when being called. Or do you think of calling a real music player(s) and verify their outputs / return codes?

As for the library: I dont know. Does some sort of mock-library exist in the tests already? Something with weird characters that are known to cause trouble?

cvx35isl:
I will bring it up to master over the next days.

Regarding the security: i still have no idea how to effectively improve it and dont see a problem either. It works flawless for me since published.

cvx35isl:
i started merging current master and see about the tests + security

(in fact I forgot about this PR as i am using my branch for years now)

cvx35isl:
simple test is added, like how I use it

biggest of not only risk I see here is the user messing up his command -> he'll notice quickly

docs/plugins/play.rst
```diff
@@ -107,6 +107,15 @@ string, use ``$args`` to indicate where to insert them. For example:
 indicates that you need to insert extra arguments before specifying the
 playlist.
 
+The above example, however, does not work with current ``mpv`` because the
+``--playlist`` argument wants a different syntax. To satisfy this, the optional
+``$playlist`` can be used to meet that:
```
JOJ0:
```suggestion
Some players require a different syntax. For example, with ``mpv`` the optional
``$playlist`` variable can be used to match the syntax of the ``--playlist`` option:
```

Let's try not to focus on mpv _only_ in the docs, but it is definitely worth mentioning that especially with this player it is required.

Updated the suggestion again. What do you think @cvx35isl, is this how we can  put it?

test/plugins/test_play.py
```diff
@@ -105,6 +105,19 @@ def test_raw(self, open_mock):
...
+            "echo --some params --playlist=$playlist --some-more params"
+        )
+
+        self.run_command("play", "nice")
+
+        open_mock.assert_called_once
+
+        commandstr = open_mock.call_args_list[0][0][1]
+        assert commandstr.startswith("echo --some params --playlist=")
+        assert commandstr.endswith("--some-more params")
```
JOJ0:
```suggestion
        assert commandstr.endswith(" --some-more params")
```

tiny improvement idea but still you don't really test that the `actual playlist path is in the middle. But probably that would be overkill and introduce more difficulties. I'm fine with just testing start and end.

JOJ0 (changes requested):
Hi, this is looking good already, tiny suggestions. Thanks!

test/plugins/test_play.py
cvx35isl: @JOJ0 yes, thanks. I agree with these, no problem. Just had no time to do it yet ...

JOJ0: ok please apply the suggestions then. and do another rebase with master (I don't seem to be able to do it via the UI currently, it says because of conflicts) Tell me if that doesn't work and I'll fix and push it :-) Thanks!

JOJ0: Hi again @cvx35isl I applied the changes above. I also tried to help and rebase this branch on current master but running out of time. There is quite som conflict resolving to do and I better leave this to you since you know by heart what 's important to use. Maybe start putting your changes on top of the current master is quicker than trying to resolve many conflicts, mostly only because of docs and code formatting issues? Not sure, your call, just an idea and HTH :-)

cvx35isl:
Hello @JOJ0 ,  thanks!

got it rebalanced and all the linting passed - hope its good now.

JOJ0:
> Hello @JOJ0 , thanks!
> 
> got it rebalanced and all the linting passed - hope its good now.

Hi, thanks, but I think something went clearly wrong here. 65 files changed. I see loads of recent changes in here. Are you sure you rebased with current master?

cvx35isl:
> > Hello @JOJ0 , thanks!
> Hi, thanks, but I think something went clearly wrong here. 65 files changed. I see loads of recent changes in here. Are you sure you rebased with current master?

yes, seen that. I rebalanced this time, but not all the time in this branch (its longgoing): once or twice I merged upstream/master (your master) in. Thats why it knows these commits.

I assume (!) all changes but mine will be noops when you merge, as your branch also knows these commits allready. But sure I am not.

If you cant try / do that safely, I need to find some other way. Re-create the PR, perhaps. Or my branch? Don't know yet. Never had this.

JOJ0:
> > > Hello @JOJ0 , thanks!
> > > Hi, thanks, but I think something went clearly wrong here. 65 files changed. I see loads of recent changes in here. Are you sure you rebased with current master?
> 
> yes, seen that. I rebalanced this time, but not all the time in this branch (its longgoing): once or twice I merged upstream/master (your master) in. Thats why it knows these commits.
> 
> I assume (!) all changes but mine will be noops when you merge, as your branch also knows these commits allready. But sure I am not.
> 
> If you cant try / do that safely, I need to find some other way. Re-create the PR, perhaps. Or my branch? Don't know yet. Never had this.

What sometimes worked for me if all is messed up:
- Backup your branch locally `git switch -c backup_branch` 
- Go back to the broken branch `git switch -`
- Reset it to current master `git fetch -av; git reset --hard origin/master`
- Cherry-pick only  the commits you need, one by one and carfully resolve conflicts: `git log backup_branch` `git cherry-pick 1234asdf`

It's almost a kind of recreating it. If you only have a hand full of commits it will work.

Hope that helps!

cvx35isl:
thanks! 

I ended up doing this slightly different:  used `git checkout backup_branch -- <file>` on the 4 files I changed instead of cherry-picking. Then committed this again.

Makes the diff to current master much shorter.

Here the github UI still lists 54 commits to merge: all mine and the two by you. Then the _"Files changed"_ tab has the 4 files I changed only. Looks right to me.

JOJ0:
whoops I ment to merge and squash, not close. On mobile here.. 🤣

JOJ0 (approved):
Thanks for the rebase again, it is still weird, most commits are here two times or more. The diff looks perfect though. I will merge and squash this so we simply get rid of those commits and just one remains - it's not so much actual code changes in the end.

I just merged in the latest master in via the button - no conflicts this time :-)

Thanks for you patience with this!

PR #4731, opened 2023-03-27
Add `-l` flag to `list` command to limit query results

snejus:
## Description

I'm slowly coming back #4362: splitting the `--limit` flag into a separate PR is something that I promised I'd do!

## To Do

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -1078,7 +1078,7 @@ class Database:
 
     # Querying.
 
-    def _fetch(self, model_cls, query=None, sort=None):
+    def _fetch(self, model_cls, query=None, sort=None, limit=None):
         """Fetch the objects of type `model_cls` matching the given
         query. The query may be given as a string, string sequence, a
         Query object, or None (to fetch everything). `sort` is an
@@ -1089,10 +1089,11 @@ class Database:
         where, subvals = query.clause()
         order_by = sort.order_clause()
 
-        sql = ("SELECT * FROM {} WHERE {} {}").format(
+        sql = ("SELECT * FROM {} WHERE {} {} {}").format(
             model_cls._table,
             where or '1',
             f"ORDER BY {order_by}" if order_by else '',
+            f"LIMIT {limit}" if limit else '',
         )
 
         # Fetch flexible attributes for items matching the main query.
--- a/beets/library.py
+++ b/beets/library.py
@@ -1512,7 +1512,7 @@ class Library(dbcore.Database):
 
     # Querying.
 
-    def _fetch(self, model_cls, query, sort=None):
+    def _fetch(self, model_cls, query, sort=None, limit=None):
         """Parse a query and fetch.
 
         If an order specification is present in the query string
@@ -1533,9 +1533,7 @@ class Library(dbcore.Database):
         if parsed_sort and not isinstance(parsed_sort, dbcore.query.NullSort):
             sort = parsed_sort
 
-        return super()._fetch(
-            model_cls, query, sort
-        )
+        return super()._fetch(model_cls, query, sort, limit)
 
     @staticmethod
     def get_default_album_sort():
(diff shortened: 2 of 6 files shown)


beets/dbcore/db.py
```diff
@@ -1078,7 +1078,7 @@ def _make_attribute_table(self, flex_table):
 
     # Querying.
 
-    def _fetch(self, model_cls, query=None, sort=None):
+    def _fetch(self, model_cls, query=None, sort=None, limit=None):
         """Fetch the objects of type `model_cls` matching the given
         query. The query may be given as a string, string sequence, a
         Query object, or None (to fetch everything). `sort` is an
```
sampsyo, line 1084: Seems like the docstring may need a mention of the new `limit` parameter.

sampsyo (commented):
Thank you for getting this started as an independent PR!!

I am a little concerned that there may be some problems with this approach—namely with providing the limit *solely* as a `LIMIT` clause on the underlying SQL query. The issue has to do with how beets queries work when they deal with flexible attributes and other complex (slow) queries. You'll see that the `Results` object does some _post-processing_ on the query result, including sorting and filtering the objects if the query can't be done in "pure SQL."

This means that doing, for example, `beet ls --limit=5 my_great_field=foo` will actually look for the first 5 items in the entire library and *then* find the ones matching the query `my_great_field=foo`. The same is true for "slow" sorting; we'll be sorting stuff after applying the limit.

I think this probably shouldn't be *too* hard to address—it just requires detecting slow queries and then teaching `Results` how to apply limiting at that level. We will also want tests for this kind of query to show that limiting still works.

snejus:
> Thank you for getting this started as an independent PR!!
> 
> I am a little concerned that there may be some problems with this approach—namely with providing the limit _solely_ as a `LIMIT` clause on the underlying SQL query. The issue has to do with how beets queries work when they deal with flexible attributes and other complex (slow) queries. You'll see that the `Results` object does some _post-processing_ on the query result, including sorting and filtering the objects if the query can't be done in "pure SQL."
> 
> This means that doing, for example, `beet ls --limit=5 my_great_field=foo` will actually look for the first 5 items in the entire library and _then_ find the ones matching the query `my_great_field=foo`. The same is true for "slow" sorting; we'll be sorting stuff after applying the limit.
> 
> I think this probably shouldn't be _too_ hard to address—it just requires detecting slow queries and then teaching `Results` how to apply limiting at that level. We will also want tests for this kind of query to show that limiting still works.

Ah snap, I've completely forgotten about the slow queries 🤦🏽‍♂️ Thanks a lot for this, I will update the implementation accordingly!

Just to double-check: are we looking to remove the 'slow' queries at some point in the future, where filtering and sorting is performed directly with appropriate SQL queries?

sampsyo:
> Just to double-check: are we looking to remove the 'slow' queries at some point in the future, where filtering and sorting is performed directly with appropriate SQL queries?

That is a very good question that unfortunately doesn't have a succinct answer. In one sense, yes: if we could somehow convert everything we wanted to do into proper SQLite queries, then we could get rid of the whole second path for "slow" queries. On the other hand, however, no: "slow" queries offer important functionality that can be surprisingly hard to duplicate with "fast" queries. In short, the reason "slow" queries exist is so that we can do arbitrary filtering/sorting that is difficult (or perhaps impossible) to do with an actual SQLite clause. These "difficult" cases fall into two categories:

* First, anything having to do with flexible attributes. It is possible that we could manage these with a sufficiently smart `JOIN` setup and perhaps a schema change, but we haven't figured it out yet.
* Queries that are implemented in Python. For example, regular expression queries use Python's `re` module—it would be really hard to replicate functionality this in pure SQL. A possible alternative would be pre-registering SQLite callbacks, but it's not clear this complexity would be worth it.

So while it's a good long-term goal, changing everything to live in "fast" query land might be a pretty big project!

snejus:
This sounds great @sampsyo. I do completely agree with the pain points you listed. 

> First, anything having to do with flexible attributes. It is possible that we could manage these with a sufficiently smart JOIN setup and perhaps a schema change, but we haven't figured it out yet.

This is something I've been extensively looking at and is the main push behind #4362. I think that PR got slightly noisy regarding some of the nice-to-haves, so I have now extracted the key bits into #4746. Looking forward to your feedback!

> Queries that are implemented in Python. For example, regular expression queries use Python's re module—it would be really hard to replicate functionality this in pure SQL. A possible alternative would be pre-registering SQLite callbacks, but it's not clear this complexity would be worth it.

Does #4741 cover the case you mentioned above? I've been hunting for Python-only queries and adding SQL-compatibility there. Relatedly, see `SingletonQuery` in #4744

PR #4733, opened 2023-03-27
Update beetcamp URL in id_extractors

snejus:
## Description

A tiny beetcamp URL fixup

## To Do


JOJ0:
Oh yeah @snejus that was my bad! Only since a couple of days I'm using your version! Awesome work! Thanks for that and yes we definitely should state your fork in the code, if any! I see in readthedocs we already are linking to yours. Nice!

Issue #4298, opened 2022-02-25
Compatibility with BPM format of DJ programs i.e. BPM's as floats


sandersantema:
### Proposed solution

Offer the ability to store BPM's as floats.

### Objective

I've seen some discussion about this in this previous [issues](https://github.com/beetbox/beets/issues/762).

In certain use-cases BPM's as integers are fine, but if you're using beets in tandem with for instance a DJ program such as Traktor floats are really needed or in general if you'd like to have accurate BPM's. That is because a lot of songs have non integer BPM values which I'd like to store in the tags of my files. 

I'd guess that BPM's as floats are however not suitable for most other functions of beets e.g. if you'd want a playlist with songs having a BPM of `130` you'd probably want to include songs with a BPM of `129.9` and `130.1` as well. 

What I think is important to figure out is whether it is already possible to circumvent this issue using existing functions of beets, for instance the playlist problem which I proposed is solvable with a trivial regex. This is instead of increasing the scope of this and for instance implementing a function where you can give a range of BPM's or a function to round them off for the purpose of for instance creating a playlist. 

Another option would be a setting to allow the user to choose whether they'd want to store BPM's as integers or floats but this would again increase the scope.

#### Goals

- Allow for the storage of BPM's as floats.
- Find out whether it is necessary to round off these floats for other things than storage or if the user already has the capability to do this themselves.

#### Non-goals

- A setting to choose whether you'd want to store your BPM's as floats or integers. This would be really nice but seems of a lot less importance than the more pressing issue of not being able to store BPM's as floats in general and thereby essentially throwing away data because of the round-off.

#### Anti-goals

What could go wrong? This I don't know and is the reason why I opened this issue instead of trying to create a PR. It seems that this change has impact all over the codebase/ parts of beets, if anyone could point out to me where exactly this is and what code would need to change that seems like a great starting point for a PR.

sampsyo:
I think switching to `float` is a good idea; it's actually really hard to see anything that could go wrong with a coordinated switch. The comments in https://github.com/beetbox/beets/issues/762#issuecomment-43098216 still stand, i.e., we'd want to simultaneously change the database type and the MediaFile type.

With suitable formatting options, it should be possible to let people print the value out rounded to an integer, even if the internal storage uses a `float`.

JOJ0:
[...]

PR #4734, opened 2023-03-27
Change bpm to be a float (instead of int) field

baelany:
## Description

Fixes issue #4298 .  
To address the feature outlined in issue #4298, we modified `beets/library.py` and `beetsplug/bpm.py` to have bpm support float type instead of integer type. We locally tested our changes by making modifications to `test/test_beatport.py` and `test/test_query.py`.

## To Do

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -495,7 +495,7 @@ class Item(LibModel):
         'disctotal': types.PaddedInt(2),
         'lyrics': types.STRING,
         'comments': types.STRING,
-        'bpm': types.INTEGER,
+        'bpm': types.FLOAT,
         'comp': types.BOOLEAN,
         'mb_trackid': types.STRING,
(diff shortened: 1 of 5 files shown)


sampsyo:
Thank you for getting this going!!

This is a somewhat sensitive change, so we will want to proceed with caution. In particular, there are two potential pitfalls on my mind:

* If someone has an existing library with a bunch of integer `bpm` values, will using that database with the new version of beets cause any problems?
* We will want to change [MediaFile](https://github.com/beetbox/mediafile) to use `float` also. Will it cause any problems for the two to temporarily disagree on the type of the field? Or do we need to make a coordinated change—updating `MediaFile`, releasing a new version, and changing beets' dependencies to require the updated version of `MediaFile` before release?

I would love any help thinking through these issues in detail!

PR #4741, opened 2023-04-04
Perform regex and `bareasc` lookups using SQL

snejus:
## Description

* Define a custom function which performs regex lookups natively in SQL. This improves
performance of lookups like `beet list path::hello`.
* Define a function which runs `unidecode` for the `bareasc` lookups

## To Do

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -975,6 +975,7 @@ class Database:
         conn = sqlite3.connect(
             py3_path(self.path), timeout=self.timeout
         )
+        self.add_functions(conn)
 
         if self.supports_extensions:
             conn.enable_load_extension(True)
@@ -986,6 +987,14 @@ class Database:
         # Access SELECT results like dictionaries.
         conn.row_factory = sqlite3.Row
         return conn
+
+    def add_functions(self, conn):
+        def regexp(value, pattern):
+            if isinstance(value, bytes):
+                value = value.decode()
+            return re.search(pattern, str(value)) is not None
+
+        conn.create_function("regexp", 2, regexp)
 
     def _close(self):
         """Close the all connections to the underlying SQLite database
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -23,6 +23,8 @@ import threading
 import sqlite3
 import contextlib
 
+from unidecode import unidecode
+
 import beets
 from beets.util import functemplate
 from beets.util import py3_path
@@ -995,6 +997,7 @@ class Database:
             return re.search(pattern, str(value)) is not None
 
         conn.create_function("regexp", 2, regexp)
+        conn.create_function("unidecode", 1, unidecode)
 
     def _close(self):
         """Close the all connections to the underlying SQLite database
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -231,6 +231,9 @@ class RegexpQuery(StringFieldQuery):
                                                  "a regular expression",
                                                  format(exc))
 
+    def col_clause(self):
+        return f" regexp({self.field}, ?)", [self.pattern.pattern]
+
     @staticmethod
     def _normalize(s):
         """Normalize a Unicode string's representation (used on both
--- a/beetsplug/bareasc.py
+++ b/beetsplug/bareasc.py
@@ -42,6 +42,14 @@ class BareascQuery(StringFieldQuery):
         val = unidecode(val)
         return pattern in val
 
+    def col_clause(self):
+        """Compare ascii version of the pattern."""
+        clause = f"unidecode({self.field})"
+        if self.pattern.islower():
+            clause = f"lower({clause})"
+
+        return rf"{clause} LIKE ? ESCAPE '\'", [f"%{unidecode(self.pattern)}%"]
+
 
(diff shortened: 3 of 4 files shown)


sampsyo:
Super cool; thanks for getting this started!

I think it would be really cool to set up a generic way to register functions like this—instead of hard-coding one case for regexes, is there a way we can make this reasonably extensible? This would work well with beets's current facility that lets plugins register new query types via prefixes, for instance.

Second, I'd be really interested to know how this affects performance! Have you had a chance to measure that? Even just a few runs of a sample query with `time` could be really insightful.

snejus:
> Second, I'd be really interested to know how this affects performance! Have you had a chance to measure that? Even just a few runs of a sample query with time could be really insightful.

Of course! Have just tested filtering `items` with `path` and `title` fields using three regex filters
* `::a` picks up >70% of items
* `::aa` picks up 3% of paths and 0.3% of titles
* `::aaa` picks up just a few of each

Used the default config on my ~7K / 322GB items library. Measurements were done with `hyperfine`.

image

(the durations are shown in _seconds_)

As you can see there's some correlation between query specificity and the performance of the SQL regex query vs python regex filtering. 

The SQL regex query seems to make the inversely proportional relationship between **query specificity** and **time** more linear since (I guess) the performance is not impacted by the need to loop across and filter all data in Python. 

For example, `title::aaa` returns only 4 items, but the python filter still needs 0.775s while SQL filter needs 0.232s. 

On the other hand, timings are more or less the same when the entire database is being returned.

snejus:
> I think it would be really cool to set up a generic way to register functions like this—instead of hard-coding one case for regexes, is there a way we can make this reasonably extensible? This would work well with beets's current facility that lets plugins register new query types via prefixes, for instance.

Regarding your idea above, see 1fc3df81. Are you thinking about something in these lines?

sampsyo:
Wow; really nice work doing a thorough microbenchmark for this new approach! There is a clear performance win here for low-selectivity queries, it seems—so it's definitely worth taking this seriously.

And that's an interesting design for doing "generic" registration of these SQLite callbacks. However, I must confess that I am made pretty nervous (having been bitten many times in the past) by mutable global state for registration systems like the `FUNCTIONS_TO_REGISTER` array. This kind of pattern exists in a few places within beets already, and it causes two main categories of problems: (1) subtle test failures, because unit tests need to be careful to clean up after themselves, and (2) problems with using beets as a library, because importing the library now has side-effects. On the other hand, I 100% see why you took this approach: because query classes are global (not associated with any specific `dbcore.Database` object), it's not clear where we should be keeping track of this registration info except for in global state.

With that in mind, and with *sincere* apologies for flip-flopping, perhaps we should move forward in this PR *without* the generic mechanism, and just make it work for `regex`. We could address that in a separate PR and give it the careful thought it deserves. (Maybe we can come up with a strategy to lazily register query callbacks on-demand when they are first used, which would let us avoid an up-front registration phase.) Would that make sense for this PR?

snejus:
I see - I have had similar thoughts when I analyzed `beets` code in detail and found the way the global state was used. Given that the use case for registering such functions is limited (at least currently!) I am more than happy to hypothesize a possible design at some point in the future. Would it be possible for me to get the _push-to-some-branch-which-is-not-the-main-branch_ permission for `beetbox/beets` so that I can save this for later reference?

> With that in mind, and with sincere apologies for flip-flopping, perhaps we should move forward in this PR without the generic mechanism, and just make it work for regex. We could address that in a separate PR and give it the careful thought it deserves. (Maybe we can come up with a strategy to lazily register query callbacks on-demand when they are first used, which would let us avoid an up-front registration phase.) Would that make sense for this PR?

It makes complete sense - you can see that this is the way they got registered in #4362 in any case.

snejus:
@sampsyo There was only one other `Query` that also requires registering a SQL function - `BareascQuery` so I added it to this PR too.

snejus:
See the performance of `bareasc` update below

image

You can see that it adds a slight overhead of around 30ms to the empty query (I guess that's because the SQL now runs `unidecode` for every row before filtering). On the other hand, the query duration is nicely nicely consistent regardless of how many rows are being returned.

sampsyo (approved):
Awesome; this looks perfect!! And yes, I've also sent you an invitation to the @beetbox org. So I'll let you do the honors of pressing the green button to merge. :smiley:

snejus:
Though, I'm suspicious about the, actually, _**weirdly** consistently quick queries_ in the last test given the timings that we saw in the previous [here](https://github.com/beetbox/beets/pull/4741#issuecomment-1500747400). On the other hand, all queries are faster today - to be honest my system is far less busy today in comparison to two days ago.

sampsyo:
That's the way with benchmarking, I guess—it can be surprisingly hard to know everything has been controlled. :shrug:

snejus:
> That's the way with benchmarking, I guess—it can be surprisingly hard to know everything has been controlled. :shrug:

Thanks a lot!! I'm absolutely thrilled about this! :)

snejus:
See [this commit](https://github.com/beetbox/beets/commit/2a6fd33d56895fa52df464e2849db8084c148590) on the [generic-db-functions](https://github.com/beetbox/beets/tree/generic-db-functions) branch

PR #4744, opened 2023-04-08
Define SingletonQuery to perform filtering through SQL

snejus:
## Description

Added `SingletonQuery` which queries 
* `album_id is NULL` for `singleton:true`
* `not (album_id is NULL)` for `singleton:false`

As a result
* A nice speed-up of the `singleton:true` query 
* `singleton:1` and `singleton:0` now work fine!

This is ultimately building towards replacing python-only queries with SQL equivalents.

### Timings
Tested with default config on my ~7K items library that has 350 singletons. Numbers represent seconds, and each command was run 10 times.

image

## To Do

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -45,6 +45,15 @@ log = logging.getLogger('beets')
 # Library-specific query types.
 
 class SingletonQuery(dbcore.FieldQuery):
+    """This query is responsible for the 'singleton' lookup.
+
+    It is based on the FieldQuery and constructs a SQL clause
+    'album_id is NULL' which yields the same result as the previous filter
+    in Python but is more performant since it's done in SQL.
+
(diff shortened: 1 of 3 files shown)


sampsyo:
Wow; incredible!! I'm absolutely thrilled that this fit in so nicely, requiring almost no code to make it work, and that the performance win is measurable. Really good thinking on this!

Any chance you could add a quick docstring to the new `SingletonQuery` class? It seems like it would be nice to mention that it takes the place of a generic `FieldQuery` for the `singleton` field, which also works the same way, but is just faster.

snejus:
@sampsyo thanks, of course! Have done so :)

sampsyo (approved):
Looks perfect! Many thanks again!!

Issue #1866, opened 2016-02-04
Let mbsubmit plugin open seeded "add release" page


Profpatsch:
When packaging I found mbsubmit, but I can’t figure out how to access it’s functionality from the command line.

sampsyo:
There's actually no CLI interface to the plugin. Basically, all it does is:

    $ beet ls -f '$track. $title - $artist ($length)'

except as an option at the importer prompts.

Profpatsch:
I see.
The code states that it cannot submit to Musicbrainz. Picard actually has a plugin that can do that, via a generated HTML form and a post request (that’s kind of hacky, alright, but it works :) ).

Profpatsch:
So the workflow is: 

1. enable the plugin
2. use “as is”
3. copy the output strings into musicbrainz?
4. update tags after some time?

sampsyo:
Whoa; crazy. Maybe we should steal that feature. :smiley:

Yes, that's approximately my understanding. It might be worth getting feedback from @awesomer and @diego-plan9, who are more familiar with the plugin (@diego-plan9 wrote it :smiley:).

Profpatsch:
Hm, the question is how the workflow should look;

My idea:

1. The release doesn’t exist, so I hit `use as-is`
2. After the import I filter out all albums without a musicbrainz ID
3. I use `beet mbsubmit <album filter>` to submit the data; it opens a window in the browser with known tags already filled in (care has to be taken that only whole albums can be submitted, also ones with multiple mediums)

- If `use as-is` imports everything in the folder (which it should imho) mbsubmit could also submit the album cover somehow
- Problem I had: `beet mbsync` skips releases without mbid instead of trying to fetch the correct release

For reference, here is the Picard plugin that does something similar: https://github.com/musicbrainz/picard/blob/master/contrib/plugins/addrelease.py

awesomer:
The workflow for the current mbsubmit plugin is :

1) try to import the release
2) it doesn't exist, so hit "[P]" for print tracklist
3) add the release to musicbrainz, including using the tracklist durations if you can't get them via a musicbrainz importer (https://github.com/murdos/musicbrainz-userscripts)
4) hit "[I]" and input the MBID of the release you just added

The whole point of mbsubmit is to prevent you from having to import the files "As Is" in order to "print" their tracklist.

sampsyo:
Oops; sorry for misunderstanding your comment, @Profpatsch. Yes, that sounds like a good plan.

[...]

PR #4749, opened 2023-04-10
plugins: add markunmatched

doronbehar:
## Description

Add a plugin useful for bulk importers.

## To Do

### DIFF
--- a/beetsplug/markunmatched.py
+++ b/beetsplug/markunmatched.py
@@ -1,0 +1,53 @@
+from pathlib import Path
+import subprocess
+import shlex
+
+from beets.autotag.match import Recommendation
+from confuse import NotFoundError as ConfigParmNotFound
+from beets.plugins import BeetsPlugin
+from beets.ui.commands import PromptChoice
+from beets.importer import action
+
+
+class MarkUnmatched(BeetsPlugin):
+    def __init__(self):
+        super(MarkUnmatched, self).__init__()
+        self.register_listener('before_choose_candidate',
+                               self.before_choose_candidate_event)
+
+    def before_choose_candidate_event(self, session, task):
+        if task.rec == Recommendation.strong:
+            self._log.info("recommendation is strong, not notifying")
+            return
+        notify_command = ""
+        try:
+            notify_command = shlex.split(str(self.config['notify-command']))
+        except ConfigParmNotFound:
+            self._log.info("no notify-command was configured")
+        if notify_command:
+            self._log.info("notifying no-match via command configured: %s",
+                           self.config['notify-command'])
+            print_cmd = [
+                "printf", "%s",
+                "failed to find match for paths:\n",
+                "\n".join([path.decode('utf-8') for path in task.paths])
+            ]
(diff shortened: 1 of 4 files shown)


beetsplug/markunmatched.py
```diff
@@ -0,0 +1,53 @@
...
+                print(stdout.decode('utf-8'), end='')
+        return [PromptChoice(
+            'f',
+            'create a \'beets-unmatched\' File and skip',
+            self.touch_unmatched
+        )]
+
+    def touch_unmatched(self, session, task):
+        for path in task.paths:
+            Path(path.decode('utf-8') + "/beets-unmatched").touch()
```
ybnd, line 52: I imagine you wanted this file to be easily visible from the terminal/file browser, but my first thought was that it should probably be hidden (i.e. `.beets-unmatched`).

Could be useful to make the filename configurable too.

ybnd (commented):
I think this has a lot of potential as a more generic plugin -- some random thoughts:
* Instead of only focusing on "weak" matches, the plugin could map recommendation levels to different commands. 
* These commands don't have to be limited to notifications
* The prompt option to create a "marker file" could also be a configurable command!
  ```sh
  xargs -I $ touch $/beets-unmatched
  ```

doronbehar:
Hey and thanks for the review @ybnd!

Your points have made me think for a while about the purpose of this plugin, and I think I've reached good conclusions. In any case, I'd like to hear about use cases for the ideas you've suggested, in particular:

> - These commands don't have to be limited to notifications
> - The prompt option to create a "marker file" could also be a configurable command!

Although I am interested in hearing ideas, I tend to not want to add to many moving parts to this plugin. In fact, I was thinking even further, that perhaps what's implemented here could become an addition to the `mbsubmit` plugin, instead of a whole new plugin. Because all it tries to do is help making bulk imports easier. I'd be happy to get feedback for the following idea :):

Add another `PromptChoice` for `mbsubmit.py`, that will print the tracks _into a **F**ile_ in the same format as the regular printing. The file name indeed should be configurable, as pointed out by @ybnd. This brings us sort of 1 step closer to resolve #1866.

Then, a user will be able to run something like:

```
picard $(dirname **/beets-unmatched)
```

Which will open `picard` with all of those unmatched directories, making it easy to bulk submitting those items. And also, the `beets-unmatched` file will have useful contents, which perhaps other usages will be found by users.

Eventually though, copying the [`addrelease.py` logic from picard](https://github.com/metabrainz/picard-plugins/blob/2.0/plugins/addrelease/addrelease.py?rgh-link-date=2020-03-15T09%3A49%3A41Z) I mentioned in [#1866](https://github.com/beetbox/beets/issues/1866#issuecomment-599188611), into beets will be the ideal improvement to `mbsubmit` plugin.

What do you think @sampsyo? Perhaps I should open a new PR with at least that idea added to `mbsubmit.py`?

sampsyo:
I'm sorry that I don't have the bandwidth to carefully review this, but on principle, I certainly have no objection to adding that "file export" option to the `mbsubmit` plugin!

doronbehar:
Closing this, as progress seems better at https://github.com/beetbox/beets/pull/4807 .

PR #4771, opened 2023-04-27
Add spotify as a fetchart source

arsaboo:
Added `spotify` as a fetchart source. 

Let me know if this approach looks fine. 

## To Do

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -15,25 +15,26 @@
 """Fetches album art.
 """
 
-from contextlib import closing
 import os
 import re
-from tempfile import NamedTemporaryFile
 from collections import OrderedDict
-from bs4 import BeautifulSoup
+from contextlib import closing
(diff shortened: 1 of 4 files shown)


wisp3rwind:
Looks reasonable at a first glance :tada: (Without checking anything Spotify-specific.) One small thing: We list `beautifulsoup` as an optional dependency, so we need to import in conditionally. See for example the `lyrics` plugin and the dependency specification in `setup.py`.

wisp3rwind:
Looking good! Would you mind rebasing on (or merging) master in order to get https://github.com/beetbox/beets/pull/4772 for a proper CI run?

arsaboo:
Thanks @wisp3rwind for that fix...all green now ✅

wisp3rwind:
Great! I didn't check whether the Spotify "API" usage is correct, but since this is well-contained and otherwise the code looks good, I'm just going ahead and merge.

PR #4774, opened 2023-04-30
fetchart: Move some source-specific code from base plugin into source classes

wisp3rwind:
Small follow-up to #4771: We have a bunch of source-specific code in the plugin constructor: It seems cleaner to me to move that into the source classes and properly separate concerns.


sampsyo (approved):
This is awesome. Distributing the `available` checks is especially valuable; this was beginning to get pretty unwieldy. :tada:

PR #4775, opened 2023-04-30
CI: update python versions

wisp3rwind:
Given that Python 3.11 is released, we've not been testing the most recent versions anymore. Thus, (try to) use semver x-ranges in order to avoid needing manual updates of this in the future.

~~Don't merge yet, I'm not sure that the range specification for `-dev` versions is actually valid as I used it here.~~
Unfortunately, [semver intentionally doesn't give a way to specify `3.x-dev`](https://github.com/npm/node-semver#prerelease-tags) to always get the latest development version. Thus, the 3.12 specification will still need to be updated manually.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -975,7 +975,11 @@ class Database:
         # bytestring paths here on Python 3, so we need to
         # provide a `str` using `py3_path`.
         conn = sqlite3.connect(
-            py3_path(self.path), timeout=self.timeout
+            py3_path(self.path),
+            timeout=self.timeout,
+            # We have our own same-thread checks in _connection(), but need to
+            # call conn.close() in _close()
+            check_same_thread=False,
         )
         self.add_functions(conn)
 
@@ -1005,7 +1009,9 @@ class Database:
         unusable; new connections can still be opened on demand.
         """
         with self._shared_map_lock:
-            self._connections.clear()
+            while self._connections:
+                _thread_id, conn = self._connections.popitem()
+                conn.close()
 
     @contextlib.contextmanager
     def _tx_stack(self):
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -923,6 +923,11 @@ class Database:
     """
 
     def __init__(self, path, timeout=5.0):
+        if sqlite3.threadsafety == 0:
+            raise RuntimeError(
+                "sqlite3 must be compiled with multi-threading support"
+            )
+
         self.path = path
         self.timeout = timeout
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, windows-latest]
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11-dev']
+        python-version: ['3.7', '3.8', '3.9', '3.x', '3.12-dev']
 
     env:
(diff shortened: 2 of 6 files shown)


wisp3rwind:
Turns out that our 3.11 CI has been failing on Windows already, but nobody noticed because `continue-on-error` job steps show up as successful...

sampsyo:
Aha; that's cool! I didn't realize `3.x` was an option; this makes lots of sense to me.

It's nice but also worrisome that this turned up so many perplexing-looking filesystem-related errors on Windows… it looks like something having to do with shutting down importer threads properly? That will certainly be fun to debug. Maybe we should suppress this failure (as it was being unintentionally suppressed before) in order to merge this PR and work on that separately?

wisp3rwind:
> It's nice but also worrisome that this turned up so many perplexing-looking filesystem-related errors on Windows… it looks like something having to do with shutting down importer threads properly? That will certainly be fun to debug. Maybe we should suppress this failure (as it was being unintentionally suppressed before) in order to merge this PR and work on that separately?

I'll make an attempt to fix it quickly here; if that doesn't work this sounds like a good strategy!

The failures happen mostly in test shutdown (where the database file seems to unexpectedly remain open), but there's a different failure in the replaygain tests (due to an unexpectly missing `threads` config value), as well as a failure in `PlaylistTestRelativeToLib.test_name_query_with_absolute_paths_in_playlist` which is (superficially at least) different.
So, at least part of the issue is related to multi-threaded execution.

One thing that I noticed is that we never explicitly close sqlite3 connections (But leave that up to the GC I presume?). Doing so seems to fix the dbcore migration tests. Unfortunately, this is much harder to fix in general: I attempted to close the connections in `dbcore.db.Database._close`, but this errors out saying that sqlite3 objects can only be used from the threads that created them. Which would mean we need to ensure that any thread that shuts down closes its connection. I expect that to be quite tricky.

Maybe, someone (...) should sift through the sqlite3 source code changes between 3.10 and 3.11 and see whether any detailed explanation and rational can be found?

wisp3rwind:
We're not the first to discover this: https://github.com/python/cpython/issues/97641

beets/dbcore/db.py
```diff
@@ -975,7 +980,11 @@ def _create_connection(self):
         # bytestring paths here on Python 3, so we need to
         # provide a `str` using `py3_path`.
         conn = sqlite3.connect(
-            py3_path(self.path), timeout=self.timeout
+            py3_path(self.path),
+            timeout=self.timeout,
+            # We have our own smae-thread checks in _connection(), but need to
```
sampsyo, line 985:
```suggestion
            # We have our own same-thread checks in _connection(), but need to
```

sampsyo (approved):
Wow; this looks like a tricky issue, according to https://github.com/python/cpython/issues/97641#issuecomment-1291654345. Nice work tracking this down. It's a relief that it suffices to close everything in our existing `self._connections` dict!

wisp3rwind:
> Wow; this looks like a tricky issue, according to [python/cpython#97641 (comment)](https://github.com/python/cpython/issues/97641#issuecomment-1291654345). Nice work tracking this down.

Well, luckily, I didn't need to track down the Cpython side of this. I could have saved quite some time by immediately searching for this problem in their issue tracker 😆 

> It's a relief that it suffices to close everything in our existing `self._connections` dict!

Luckily, the `check_same_thread` flag exists, so that doing this is in fact viable. Initially, I didn't know about it; and at that point things would indeed have become very tricky

PR #4778, opened 2023-05-02
Fetchart: use cover_art_url flex attr as a cover art source

### LINKED ISSUE #4707: fetchart: Use MusicBrainz' Bandcamp/Discogs/... association to fetch art
### Discussed in https://github.com/beetbox/beets/discussions/4676

<div type='discussions-op-text'>

<sup>Originally posted by **evur** February 12, 2023</sup>
There are instances where covers are only available from one of these and not from the currently implemented sources.

The Discogs ID can be retrieved from the MusicBrainz data. This allows for finding a cover precisely should Discogs have it and MusicBrainz (via CoverArtArchive) not.</div>

arsaboo:
## Description

Fixes #4707 

It turned out to be a lot simpler than I thought. As discussed in #4707, this PR creates a generic function that will use the image URL from `cover_art_url` field and update the album art. The idea was discussed in the linked issue to allow any plugin to save the cover art link to a flexible field `cover_art_url` and then `fetchart` should be able to use that field. Users can also manually update this field to specify a covert art image. 

Tested this with the [JioSaavn](https://github.com/arsaboo/beets-jiosaavn) plugin and it works perfectly. 

## To Do

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -837,7 +837,7 @@ class CoverArtUrl(RemoteArtSource):
 
     def get(self, album, plugin, paths):
         try:
-            url = album.item().get().cover_art_url
+            url = album.items().get().cover_art_url
             print(url)
         except ValueError:
             self._log.debug('Cover art URL not found for {0}', album)
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -15,24 +15,21 @@
 """Fetches album art.
 """
 
-from contextlib import closing
 import os
 import re
-from tempfile import NamedTemporaryFile
+import tempfile
 from collections import OrderedDict
+from contextlib import closing
+from mimetypes import guess_extension
+from tempfile import NamedTemporaryFile
 
-import requests
-
-from beets import plugins
-from beets import importer
-from beets import ui
-from beets import util
-from beets import config
-from mediafile import image_mime_type
-from beets.util.artresizer import ArtResizer
-from beets.util import sorted_walk
-from beets.util import syspath, bytestring_path, py3_path
 import confuse
+import requests
+from mediafile import image_mime_type
+
+from beets import config, importer, plugins, ui, util
+from beets.util import bytestring_path, py3_path, sorted_walk, syspath
+from beets.util.artresizer import ArtResizer
 
 CONTENT_TYPES = {
     'image/jpeg': [b'jpg', b'jpeg'],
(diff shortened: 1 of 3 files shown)


beetsplug/fetchart.py
```diff
@@ -975,6 +974,26 @@ def get(self, album, plugin, paths):
             self._log.debug('Spotify: error loading response: {}'
                             .format(response.text))
             return
+
+
+class CoverArtUrl(RemoteArtSource):
```
sampsyo, line 979: Maybe a docstring would be nice here to explain the (somewhat nuanced) role this source plays?

beetsplug/fetchart.py
```diff
@@ -975,6 +974,26 @@ def get(self, album, plugin, paths):
...
+    NAME = "Cover Art URL"
+
+    def get(self, album, plugin, paths):
+        image_url = None
+        try:
+            image_url = album.items().get().cover_art_url
+            self._log.debug(f'Cover art URL {image_url} found for {album}')
+        except (AttributeError, TypeError):
```
sampsyo, line 987: I'm not entirely sure when a `TypeError` would occur here, but I'll take your word for it!

docs/plugins/fetchart.rst
```diff
@@ -266,6 +266,11 @@ Spotify backend is enabled by default and will update album art if a valid Spoti
 .. _pip: https://pip.pypa.io
 .. _BeautifulSoup: https://www.crummy.com/software/BeautifulSoup/bs4/doc/
 
+Cover Art URL
+'''''''''''''
+
+The `fetchart` plugin can also use a flexible attribute field ``cover_art_url`` where you can manually specify the image URL to be used as cover art. Any custom plugin can use this field to provide the cover art and ``fetchart`` will use it as a source. Just add ``cover_art_url`` to the list of sources in your configuration to enable this source.
```
sampsyo, line 272: Perhaps the last sentence is not necessary, since it seems to be enabled by default?

beetsplug/fetchart.py
```diff
@@ -975,6 +974,26 @@ def get(self, album, plugin, paths):
...
             return
+
+
+class CoverArtUrl(RemoteArtSource):
+    NAME = "Cover Art URL"
+
+    def get(self, album, plugin, paths):
+        image_url = None
+        try:
+            image_url = album.items().get().cover_art_url
```
sampsyo, line 985: I am somewhat surprised to see that this only looks at the item-level attribute, rather than an attribute on the album also. I presume this is because this is how metadata-source plugins want to provide this data, which makes sense. But perhaps we also want to check `album.cover_art_url` in case that changes in the future?

sampsyo (commented):
Nice; this is looking pretty straightforward!! Very good work here. I just found a few suggestions I wanted to make within!

beetsplug/fetchart.py
arsaboo: Added

docs/plugins/fetchart.rst
arsaboo: Updated

beetsplug/fetchart.py
arsaboo: Changed the logic to first look for `album.cover_art_url`

sampsyo (approved):
Very nice work on this! Wahoo!!

docs/changelog.rst
```diff
@@ -11,6 +11,8 @@ for Python 3.6).
 
 New features:
 
+* Added option use `cover_art_arl` as an album art source in the `fetchart` plugin.
```
tianon, line 14: Minor typo (missing "to", "arl" instead of "url") 👀

```suggestion
* Added option to use `cover_art_url` as an album art source in the `fetchart` plugin.
```

PR #4779, opened 2023-05-02
Fix reimport bug #4776

### LINKED ISSUE #4776: Reimporting from another source should not remove the specific information from other sources
https://github.com/beetbox/beets/pull/4735 did a great job cleaning up the reimport process. However, now that I have been using this for some time, I realized that it introduced a bug. Specifically, we are not even preserving ids from any previous imports. This seems wrong. We need ids from every possible source (for specific purposes). For example, even if we use MusicBrainz, we may need `spotify_album_id` to obtain popularity or music attributes. 

The reimport process should only overwrite the common fields (and not touch any fields from other sources). Thus, reimporting with MB should not remove `spotify_album_id`

cc: @JOJ0

arsaboo:
Fixes #4776 

@JOJ0 the fix you suggested https://github.com/beetbox/beets/issues/4776#issuecomment-1531354757 works. I went ahead and created a PR to get it fixed. Thanks for the quick resolution! 

Don't think it requires a changelog or any documentation. 

## To Do

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -826,6 +826,7 @@ class ImportTask(BaseImportTask):
             existing_fields = dict(existing_fields)
             overwritten_fields = [k for k in existing_fields
                                   if k in overwrite_keys
+                                  and new_obj.get(k)
                                   and existing_fields.get(k) != new_obj.get(k)]
             if overwritten_fields:
                 log.debug(


JOJ0:
Well, you are awesome, thanks for the great teamwork! Merging 💪

PR #4781, opened 2023-05-05
fetchart: revert a cleanup from #4774 that could break plugins

wisp3rwind:
cf. @arogl's comment https://github.com/beetbox/beets/commit/254bb297c894c0323174012ff27500867416d02b#commitcomment-111922347

> Now that this has been merged, external plugins that add to the fetchart plugin now fail with:
>
> ```AttributeError: module 'beetsplug.fetchart' has no attribute 'SOURCES_ALL'``

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -977,6 +977,12 @@ class Spotify(RemoteArtSource):
             return
 # Try each source in turn.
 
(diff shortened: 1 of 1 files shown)


sampsyo (approved):
Looks great! I also actually wasn't aware that external plugins were monkey-patching fetchart in this way. Maybe someday, we should consider a supported interface for these "sub-plugins"…

PR #4786, opened 2023-05-12
Add Youtube to the community plugins list

arsaboo:
Added Youtube as a tag source.

It works well for me. I am happy to create a PR if we want to have this as part of beets.


sampsyo:
Looks like a very cool plugin!! I don't have a strong feeling about whether to include most plug-ins or leave them as external. But happy to help explore including it if you're interested!

PR #4798, opened 2023-05-22
Prevent MusicBrainz fields from being overwritten by non-MB data

### LINKED ISSUE #4797: Musicbrainz fields should not be overwritten by non-MB data
### Problem

There are several Musicbrainz (MB) fields `mb_*` that are currently being overwritten by other plugins during import. This does not make sense. Now, I understand that many of those choices were made given that MB was the only tag source and beets was designed around MB. However, now we have several other tag sources. Thus, non-MB sources should not touch `mb_*` fields. 

Here's the logic that we should follow:

1. During the initial import (even by non-MB sources), it is fine to populate `mb_*` field as we cannot have them empty. 
2. During re-imports, we should not overwrite these fields unless they are from MB. 

Based on my understanding, we need to add additional checks in the `apply_item_metadata`:
https://github.com/beetbox/beets/blob/a19bc0570cd462aa8c978634fa82b1049330fcd1/beets/autotag/__init__.py#L76

and the `apply_metadata` functions
https://github.com/beetbox/beets/blob/a19bc0570cd462aa8c978634fa82b1049330fcd1/beets/autotag/__init__.py#L100
before overwriting `mb_*` fields. 

Would love your thoughts on this and possibly a quick fix.

arsaboo:
## Description

Fixes #4797 

Quick PR to fix the aforementioned bugs. 

## To Do

### DIFF
--- a/beets/autotag/__init__.py
+++ b/beets/autotag/__init__.py
@@ -89,7 +89,7 @@ def apply_item_metadata(item: Item, track_info: TrackInfo):
     fields = [('mb_trackid', track_info.track_id),
               ('mb_releasetrackid', track_info.release_track_id)]
     for field, value in fields:
-        if re.match(MBID_REGEX, value) or item.get(field) is None:
+        if re.match(MBID_REGEX, str(value)) or item.get(field) is None:
             item[field] = value
 
     for field, value in track_info.items():
@@ -180,7 +180,7 @@ def apply_metadata(album_info: AlbumInfo, mapping: Mapping[Item, TrackInfo]):
                   ('mb_albumartistid', album_info.artist_id),
                   ('mb_releasegroupid', track_info.artist_id)]
         for field, value in fields:
-            if re.match(MBID_REGEX, value) or item.get(field) is None:
+            if re.match(MBID_REGEX, str(value)) or item.get(field) is None:
                 item[field] = value
         # Compilation flag.
         item.comp = album_info.va
--- a/beets/autotag/__init__.py
+++ b/beets/autotag/__init__.py
@@ -87,7 +87,7 @@ def apply_item_metadata(item: Item, track_info: TrackInfo):
     item.title = track_info.title
 
     fields = [('mb_trackid', 'track_info.track_id'),
-                ('mb_releasetrackid', 'track_info.release_track_id')]
+              ('mb_releasetrackid', 'track_info.release_track_id')]
 
     for field, attr_name in fields:
         attr_value = getattr(item, field)
--- a/beets/autotag/__init__.py
+++ b/beets/autotag/__init__.py
@@ -86,16 +86,11 @@ def apply_item_metadata(item: Item, track_info: TrackInfo):
     item.artist_credit = track_info.artist_credit
     item.title = track_info.title
 
-    fields = [('mb_trackid', 'track_info.track_id'),
-              ('mb_releasetrackid', 'track_info.release_track_id')]
-
-    for field, attr_name in fields:
-        attr_value = getattr(item, field)
-        if not re.match(MBID_REGEX, attr_value):
-            setattr(item, field, eval(attr_name))
-    if track_info.artist_id and not re.match(MBID_REGEX,
-                                             item.mb_artistid):
-        item.mb_artistid = track_info.artist_id
+    fields = [('mb_trackid', track_info.track_id),
+              ('mb_releasetrackid', track_info.release_track_id)]
+    for field, value in fields:
+        if re.match(MBID_REGEX, value) or item.get(field) is None:
+            item[field] = value
 
     for field, value in track_info.items():
         # We only overwrite fields that are not already hardcoded.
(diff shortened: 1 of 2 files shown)


beets/autotag/__init__.py
```diff
@@ -80,10 +85,12 @@ def apply_item_metadata(item: Item, track_info: TrackInfo):
...
-    item.mb_trackid = track_info.track_id
-    item.mb_releasetrackid = track_info.release_track_id
-    if track_info.artist_id:
-        item.mb_artistid = track_info.artist_id
+    fields = [('mb_trackid', track_info.track_id),
+              ('mb_releasetrackid', track_info.release_track_id),
+              ('mb_artistid', track_info.artist_id)]
+    for field, value in fields:
+        if re.match(MBID_REGEX, str(value)) or item.get(field) is None:
+            item[field] = value
```
sampsyo, line 93: Would it suffice to just check whether these fields are None? That is, something like this:

```py
for field in ['mb_trackid', 'mb_releasetrackid', 'mb_artistid']:
    value = track_info.get(field)
    if value:
         item[field] = value
```

sampsyo (commented):
It's probably a good idea to do something like this! I have an inline question about possibly just ignoring all _empty_ values of these fields.

beets/autotag/__init__.py
arsaboo: Actually, that may not be enough as it will prevent overwriting the `mb_*` fields (should the user choose to update tags using another MB release. The current approach allows updating the `mb_*` fields if it is empty or if the new value is valid. Hope that makes sense.

arsaboo:
Also, I am not sure about how to fix the failing tests...can you please help me with those...thanks!

sampsyo:
Sure! It's always hard to track down a test failure, but it can be instructive about how things are supposed to work. Looking at the CI output, it seems like an indicative test that is failing is `test_update_library`, which is an mbsync test found here:
https://github.com/beetbox/beets/blob/5aae612f45716cfdea2884e3ae873c58e59ef64e/test/test_mbsync.py#L40

To set up for debugging tests, it is really helpful to be able to run them on your own machine. You'll need `pytest` and the test dependencies for beets. Something like this will work (in the beets directory) (you may need to slightly adapt the pip commands):

```
$ python3 -m pip install --user pytest
$ python3 -m pip install --user -e .[test]
```

You can then run this specific test like this:

```
$ pytest -k test_update_library test/test_mbsync.py
```

You should be able to see the failure there then.

It looks to me like updated metadata from MB is not being applied even though it should be.

beets/autotag/__init__.py
sampsyo: I guess the question I was getting at was: do we really need to check for validity? It makes the logic here quite a bit more complex, and it would be really nice to avoid the need for the special case—hopefully by just not supplying invalid data in the first place.

arsaboo:
The problem is our tests require `reflink` and I am not able to get it installed on my computer. I looked everywhere and tried everything, but looks like `reflink` is not being [maintained](https://gitlab.com/rubdos/pyreflink/-/issues/15). It has to be something simple, but I am failing at it :(

arsaboo:
The tests are failing somwhere here
https://github.com/beetbox/beets/blob/5aae612f45716cfdea2884e3ae873c58e59ef64e/test/test_edit.py#L359-L360

What I am not sure about is why are we checking for `match ' in i.mb_trackid`? This should always be `false`, which is what the failed test suggests.

sampsyo:
> The problem is our tests require `reflink` and I am not able to get it installed on my computer. I looked everywhere and tried everything, but looks like `reflink` is not being [maintained](https://gitlab.com/rubdos/pyreflink/-/issues/15). It has to be something simple, but I am failing at it :(

Ah, that is too bad! What error do you get when installing reflink?

It seems like we should be able to avoid requiring reflink to run the tests… I'll give something a try briefly.

> What I am not sure about is why are we checking for `match ' in i.mb_trackid`? This should always be `false`, which is what the failed test suggests.

I don't immediately understand this `edit` test either. The `mbsync` test above may be an easier way to start…

sampsyo:
Perhaps #4808 can avoid the requirement to install reflink!

arsaboo:
So, the `mbsync` test is fixed. Also, local tests work now after #4808. 

Out of the remaining errors, three are related to some weird comparisons like `match ' in i.mb_trackid` and one is related to `self.assertEqual(item.mb_trackid, 'new trackid')`. Both of them don't make sense to me. There are also some other tests `self.assertIn('albumid', self.lib.albums()[0].mb_albumid)`, which don't make sense. Specifically, in this case, `albumid` will never be in `mb_albumid`. My guess is those tests assume any string in the `mb_*` fields.

sampsyo:
I think the problem is the validity checks, as discussed here:
https://github.com/beetbox/beets/pull/4798/files#r1204562676

The issue is that all the tests are roughly of this format: "When I try to put the string 'my great albumid' into the `mb_albumid` field, does the `mb_albumid` field then contain the string 'my great albumid'?" And that fact does not hold because the string doesn't pass the regex-based valid-ID check. So I'd love to return to the central question: do we need to be checking for valid IDs? Or can we just rethink the underlying issue, which is that the plugins are providing invalid IDs in the first place?

arsaboo:
It is not so much as the plugins are providing invalid IDs. My understanding is that because MB is so deeply integrated, the `mb_*` fields are automatically populated during initial import even if we use a non-MB tag source. The test that checks if mb_albumid field contains the string 'my great albumid' should never be allowed (IMHO). This was fine when we only had MB as the tag source as the assumption was we can only have MB-related data. However, that is no longer the case. Ideally, we should change this behavior and only allow valid MB data to be written to MB fields (but that is a long-term exercise and a more significant undertaking - well beyond what I can handle). 

In the short term, an acceptable solution could be that we don't allow non-MB data in `mb_*` after the initial import (which this PR implements).

sampsyo:
> My understanding is that because MB is so deeply integrated, the `mb_*` fields are automatically populated during initial import even if we use a non-MB tag source.

Ah, that is actually not the case! These fields come directly from the data source. So to the extent that any of this data arrives when importing a discogs release, for example, that data is produced by the discogs plugin (and not by anything MB-related). The underlying problem may be #604, i.e., the plugins themselves are putting bad data in those fields.

arsaboo:
Ah...indeed `album_id` is used in almost all plugins (including the plugins that I wrote). I always assumed `album_id` is a required field - is it not the case? Also, just to make sure if plugins don't update `album_id`, MBID will not be updated? Is that the solution to this problem? If that is the case, I can test it out with some of my community plugins.

arsaboo:
So, we should remove this line?
https://github.com/beetbox/beets/blob/c948c0f62e0b9084234ba84ac76797e1973671ba/beetsplug/spotify.py#L268

sampsyo:
Yes, it's currently required for the reason sketched here: https://github.com/beetbox/beets/issues/4803#issuecomment-1569221397

Namely, it's used for collapsing duplicate results. The main thing we need to fix to make this work is to find a different way of removing duplicates with identical IDs in a source-agnostic way.

arsaboo:
Given that all ID fields have the format `source_id`, can we not just get the list of all identifiers and use them to find duplicates? It will also help with issues like [this one](https://github.com/beetbox/beets/discussions/4762).  

In any case, that seems like a @sampsyo kind of problem 😜 to be fixed (and something that we should be doing relatively soon). I will close this PR (and continue to use this fix in my custom function) in the meantime.

sampsyo:
Yes, something like that! I think it poses some minor logistical problems because all the plugins use different unique fields for their IDs, so it's not clear how to construct the "source/ID tuple" to use as a globally unique identifier.

> In any case, that seems like a @sampsyo kind of problem 😜

🤪 Maybe so! But at least it is a separate problem… hopefully we can come up with a simple solution under the banner of #604.

arsaboo:
Well, I am happy (and optimistic) that this may be fixed soon....closing it now.

PR #4801, opened 2023-05-25
Prevent reimporting album if an album is permanently removed from Spotify

### LINKED ISSUE #4800: [Spotify] Do not try to reimport album if it is permanently removed from Spotify
### Problem

Spotify often removes content from its library. There are two types of content removal:

In the first type, only the tracks are removed, but the data is still available. See [this](https://open.spotify.com/album/1t1AmX5GwiKa0UYESMJG2E) for example. Here's how it appears:
image

In the second type, everything is removed. See [this](https://open.spotify.com/album/61galBMmvNvVtuFjeRndZ4) for example. Here's how it appears (notice there is no information available here):

image

(Re)Import is perfectly fine for the first type of content as the data is still available. However, (re)importing (with `search-id`) in case of the second type of removal removes all the metadata. 

image

We should add a check to see if valid data is being returned from Spotify before applying the metadata.

arsaboo:
## Description

Fixes #4800

(...)

## To Do

### DIFF
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -219,6 +219,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
         album_data = self._handle_response(
             requests.get, self.album_url + spotify_id
         )
+        self._log.debug('Album data: {}', album_data)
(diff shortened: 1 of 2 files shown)


sampsyo:
Nice; looks good to me!!

PR #4804, opened 2023-05-26
Fetchart: Use the right field to obtain Spotify album id

### LINKED ISSUE #4803: Fetchart uses incorrect field for spotify_album_id
Fetchart currently uses `album.mb_albumid` to get `spotify_album_id`. This assumes `mb_albumid` has the `spotify_album_id`, which may be the case after the initial import with Spotify. However, `mb_albumid` can be overwritten with a valid `mb_albumid`, which will result in an error. 

We should be using `spotify_album_id` instead, which will always have the right value.

arsaboo:
## Description

Use `spotify_album_id` instead of `mb_albumid` to always obtain the right values. 

Fixes #4803

## To Do

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -956,7 +956,11 @@ class Spotify(RemoteArtSource):
         return HAS_BEAUTIFUL_SOUP
 
     def get(self, album, plugin, paths):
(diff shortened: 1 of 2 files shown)


sampsyo:
Looks good; thanks!!

PR #4708, opened 2023-03-13
Save external ID's via MusicBrainz

JOJ0:
## Description

MusicBrainz provides links to related websites via the "External links" section , usually found at the bottom right of a release page:

image

Fortunately this information is available via the MusicBrainz API. This PR adds the fetching of these URL's, extracts the "metadata source ID's" from those and saves them in the corresponding fields in the Beets library. 

Currently the following sources are supported: Discogs, Bandcamp, Spotify, Deezer, Beatport. Typically **a link to a Discogs release is very likely** to be provided and seems to be considered a "quasi standard" for MusicBrainz editors. It seems worth for Beets to access this information and use it with  future features. The rest of the supported external sources are not that likely to be found.

- Fixes #4220
- Is required for future feature ideas:
  - #4707

Note: This is a replacement for #4474, which was split into two separate pull requests. The first one which was a prerequesite, was merged already: #4633 

## To Do

- [x] Documentation. 
- [x] Changelog.
- [x] ~Tests~.

PR #4805, opened 2023-05-27
Allow importing Tidal album id from MusicBrainz

arsaboo:
Following #4708, this PR adds the ability to also import Tidal album id, which is also sometimes available on the release pages. 

## To Do


sampsyo:
Looks good. Someday, we might want to reconsider the hard-coding of specific ID sources here:
https://github.com/beetbox/beets/blob/95898a3055b42d5615896baca0a8c2a74d35b32b/beets/autotag/mb.py#L520-L536

This used to be reasonable when there were only one or two such IDs, but now there is a lot of duplicated code. It would be great if this could be turned into a loop over a list of source names.

PR #4807, opened 2023-05-30
mbsubmit: Add picard `PromptChoice`

doronbehar:
Make it possible to open picard from the import menu when there are weak recommendations. Add another menu option to print the tracks into a beets-unmatched file for later parsing.

## Description

A better alternative to #4749 - closes #4749 . Also provides an intermediate solution to https://github.com/beetbox/beets/issues/1866 .

cc @ybnd .

## To Do

- [x] Tests - `test_mbsubmit.py` was modified to accommodate the change.

### DIFF
--- a/beetsplug/mbsubmit.py
+++ b/beetsplug/mbsubmit.py
@@ -21,11 +21,13 @@ implemented by MusicBrainz yet.
 [1] https://wiki.musicbrainz.org/History:How_To_Parse_Track_Listings
 """
 
+import subprocess
 
 from beets import ui
 from beets.autotag import Recommendation
 from beets.plugins import BeetsPlugin
 from beets.ui.commands import PromptChoice
+from beets.util import displayable_path
 from beetsplug.info import print_data
 
 
@@ -37,6 +39,7 @@ class MBSubmitPlugin(BeetsPlugin):
             {
                 "format": "$track. $title - $artist ($length)",
                 "threshold": "medium",
+                "picard_path": "picard",
             }
         )
 
@@ -56,7 +59,21 @@ class MBSubmitPlugin(BeetsPlugin):
 
     def before_choose_candidate_event(self, session, task):
         if task.rec <= self.threshold:
-            return [PromptChoice("p", "Print tracks", self.print_tracks)]
+            return [
+                PromptChoice("p", "Print tracks", self.print_tracks),
+                PromptChoice("o", "Open files with Picard", self.picard),
+            ]
+
+    def picard(self, session, task):
+        paths = []
+        for p in task.paths:
+            paths.append(displayable_path(p))
+        try:
+            picard_path = self.config["picard_path"].as_str()
+            subprocess.Popen([picard_path] + paths)
+            self._log.info("launched picard from\n{}", picard_path)
+        except OSError as exc:
+            self._log.error(f"Could not open picard, got error:\n{exc}")
 
     def print_tracks(self, session, task):
         for i in sorted(task.items, key=lambda i: i.track):
--- a/docs/plugins/mbsubmit.rst
+++ b/docs/plugins/mbsubmit.rst
@@ -1,23 +1,40 @@
 MusicBrainz Submit Plugin
 =========================
 
-The ``mbsubmit`` plugin provides an extra prompt choice during an import
-session and a ``mbsubmit`` command that prints the tracks of the current
-album in a format that is parseable by MusicBrainz's `track parser`_.
+The ``mbsubmit`` plugin provides extra prompt choices when an import session
+fails to find a good enough match for a release. Additionally, it provides an
+``mbsubmit`` command that prints the tracks of the current album in a format
+that is parseable by MusicBrainz's `track parser`_. The prompt choices are:
+
+- Print the tracks to stdout in a format suitable for MusicBrainz's `track
+  parser`_.
+
+- Open the program `Picard`_ with the unmatched folder as an input, allowing
+  you to start submitting the unmatched release to MusicBrainz with many input
+  fields already filled in, thanks to Picard reading the preexisting tags of
+  the files.
+
+For the last option, `Picard`_ is assumed to be installed and available on the
+machine including a ``picard`` executable. Picard developers list `download
+options`_. `other GNU/Linux distributions`_ may distribute Picard via their
+package manager as well.
 
 .. _track parser: https://wiki.musicbrainz.org/History:How_To_Parse_Track_Listings
+.. _Picard: https://picard.musicbrainz.org/
+.. _download options: https://picard.musicbrainz.org/downloads/
+.. _other GNU/Linux distributions: https://repology.org/project/picard-tagger/versions
 
 Usage
 -----
 
 Enable the ``mbsubmit`` plugin in your configuration (see :ref:`using-plugins`)
-and select the ``Print tracks`` choice which is by default displayed when no
-strong recommendations are found for the album::
+and select one of the options mentioned above. Here the option ``Print tracks``
+choice is demonstrated::
 
     No matching release found for 3 tracks.
     For help, see: https://beets.readthedocs.org/en/latest/faq.html#nomatch
     [U]se as-is, as Tracks, Group albums, Skip, Enter search, enter Id, aBort,
-    Print tracks? p
+    Print tracks, Open files with Picard? p
     01. An Obscure Track - An Obscure Artist (3:37)
     02. Another Obscure Track - An Obscure Artist (2:05)
     03. The Third Track - Another Obscure Artist (3:02)
@@ -53,6 +70,11 @@ file. The following options are available:
   Default: ``medium`` (causing the choice to be displayed for all albums that
   have a recommendation of medium strength or lower). Valid values: ``none``,
   ``low``, ``medium``, ``strong``.
+- **picard_path**: The path to the ``picard`` executable. Could be an absolute
+  path, and if not, ``$PATH`` is consulted. The default value is simply
+  ``picard``. Windows users will have to find and specify the absolute path to
+  their ``picard.exe``. That would probably be:
+  ``C:\Program Files\MusicBrainz Picard\picard.exe``.
 
 Please note that some values of the ``threshold`` configuration option might
 require other ``beets`` command line switches to be enabled in order to work as
--- a/test/plugins/test_mbsubmit.py
+++ b/test/plugins/test_mbsubmit.py
@@ -45,7 +45,7 @@ class MBSubmitPluginTest(
 
         # Manually build the string for comparing the output.
         tracklist = (
-            "Print tracks? "
+            "Open files with Picard? "
             "01. Tag Title 1 - Tag Artist (0:01)\n"
             "02. Tag Title 2 - Tag Artist (0:01)"
         )
@@ -61,7 +61,9 @@ class MBSubmitPluginTest(
                 self.importer.run()
 
         # Manually build the string for comparing the output.
-        tracklist = "Print tracks? " "02. Tag Title 2 - Tag Artist (0:01)"
+        tracklist = (
+            "Open files with Picard? " "02. Tag Title 2 - Tag Artist (0:01)"
+        )
         self.assertIn(tracklist, output.getvalue())
 
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,6 +17,7 @@ Major new features:
 
 New features:
 
+* :doc:`plugins/mbsubmit`: add new prompt choices helping further to submit unmatched tracks to MusicBrainz faster.
 * :doc:`plugins/spotify`: We now fetch track's ISRC, EAN, and UPC identifiers from Spotify when using the ``spotifysync`` command.
   :bug:`4992`
 * :doc:`plugins/discogs`: supply a value for the `cover_art_url` attribute, for use by `fetchart`.


JOJ0:
Hi, interesting idea. Why not make the tag-editor configurable? Me personally I prefer puddle-tag or mp3tag (on macOS) over Picard when tags are so bad that beets alone is not sufficient.

doronbehar:
> Hi, interesting idea. Why not make the tag-editor configurable? Me personally I prefer puddle-tag or mp3tag (on macOS) over Picard when tags are so bad that beets alone is not sufficient.

The point of opening Picard is not for tagging the files of the release, but to read the already tagged files and open a 'add release' page in musicbrainz' website, that will include most of the information needed for a proper release entry in the database. Then, once it's in the database, you continue the import process, using the release id you got from musicbrainz, and beet tags the files for you.

JOJ0:
Oh I see! That is indeed a cool feature!!

jackwilsdon:
Is it possible for us to link directly to the release creation page on MusicBrainz (with the fields pre-filled) instead of doing it via Picard?

doronbehar:
> Is it possible for us to link directly to the release creation page on MusicBrainz (with the fields pre-filled) instead of doing it via Picard?

That is the idea suggested in https://github.com/beetbox/beets/issues/1866 . It's a great idea, but it's much harder to implement. See [my comment there linking picard's relevant source code](https://github.com/beetbox/beets/issues/1866#issuecomment-599188611).

arsaboo:
Short of being able to directly add information to MB, using Picard seems like a nice middle ground. Looking forward to this 👍

arsaboo:
@doronbehar can you resolve the conflicts so that we can move forward with this?
@jackwilsdon @JOJ0 Assuming we are fine with using Picard as the middle ground, is there anything else that needs to be done?

doronbehar:
Fixed merge conflicts via a rebase.

doronbehar:
Hmm not sure why CI says my order of imports is incorrect...

arsaboo:
> Hmm not sure why CI says my order of imports is incorrect...

Use `isort` to sort imports.

doronbehar:
> > Hmm not sure why CI says my order of imports is incorrect...
> 
> Use `isort` to sort imports.

Done!

doronbehar:
Fixed `black` formatting as well.

arsaboo:
There are still a couple minor lint errors.

beetsplug/mbsubmit.py
```diff
@@ -56,7 +58,33 @@ def __init__(self):
...
+                PromptChoice(
+                    "o", "Open Picard with the relevent files", self.picard
+                ),
+            ]
+
+    def markunmatched(self, session, task):
+        # TODO: Implement
+        fmt = self.config["format"].as_str()
+        for item in task.items:
+            unmatchedPath = (
```
arsaboo, line 77: change the variable name to `unmatchedpath`

doronbehar:
Thanks for help. Hopefully it'll be green now.

doronbehar:
Added docs and a changelog line.

test/plugins/test_mbsubmit.py
```diff
@@ -61,7 +61,7 @@ def test_print_tracks_output_as_tracks(self):
                 self.importer.run()
 
         # Manually build the string for comparing the output.
-        tracklist = "Print tracks? " "02. Tag Title 2 - Tag Artist (0:01)"
+        tracklist = "Print tracks," "02. Tag Title 2 - Tag Artist (0:01)"
```
arsaboo, line 64: probably revert this change

doronbehar:
I don't understand how that test passed... The error now, from CI is:

```
AssertionError: 'Print tracks,02. Tag Title 2 - Tag Artist (0:01)' not found in '\n/tmp/tmpcyx2ztde/testsrcdir/the_album (2 items)\n\nFinding tags for album "Tag Artist - Tag Album".\n  Candidates:\n  1. (51.5%) Applied Artist M - Applied Album M\n             ≠ album, artist, tracks\n             match_source, None, None, None, None, None, None\n  2. (49.8%) Applied Artist MM - Applied Album MM\n             ≠ album, artist, tracks\n             match_source, None, None, None, None, None, None\n➜ # selection (default 1), Skip, Use as-is, as Tracks, Group albums,\nEnter search, enter Id, aBort, Print tracks,\nprint tracks into a "beets-unmatched" File,\nOpen Picard with the relevent files? \n/tmp/tmpcyx2ztde/testsrcdir/the_album/track_1.mp3\n\n  Match (61.5%):\n  Applied Artist - Applied Title 1\n  ≠ title, artist\n  None, Index 0, Track None, \n  ≠ Artist: Tag Artist -> Applied Artist\n  ≠ Title: Tag Title 1 -> Applied Title 1\n➜ Apply, More candidates, Skip, Use as-is, Enter search, enter Id, aBort,\nPrint tracks, print tracks into a "beets-unmatched" File,\nOpen Picard with the relevent files? \n/tmp/tmpcyx2ztde/testsrcdir/the_album/track_2.mp3\n\n  Match (61.5%):\n  Applied Artist - Applied Title 2\n  ≠ title, artist\n  None, Index 0, Track None, \n  ≠ Artist: Tag Artist -> Applied Artist\n  ≠ Title: Tag Title 2 -> Applied Title 2\n➜ Apply, More candidates, Skip, Use as-is, Enter search, enter Id, aBort,\nPrint tracks, print tracks into a "beets-unmatched" File,\nOpen Picard with the relevent files? 02. Tag Title 2 - Tag Artist (0:01)\n\n  Match (61.5%):\n  Applied Artist - Applied Title 2\n  ≠ title, artist\n  None, Index 0, Track None, \n  ≠ Artist: Tag Artist -> Applied Artist\n  ≠ Title: Tag Title 2 -> Applied Title 2\n➜ Apply, More candidates, Skip, Use as-is, Enter search, enter Id, aBort,\nPrint tracks, print tracks into a "beets-unmatched" File,\nOpen Picard with the relevent files? '
```

I changed it now to something which might work.

docs/plugins/mbsubmit.rst
```diff
@@ -1,23 +1,36 @@
...
+``mbsubmit`` command that prints the tracks of the current album in a format
+that is parseable by MusicBrainz's `track parser`_. The prompt choices are:
+
+- Print the tracks to stdout in a format suitable for MusicBrainz's `track
+  parser`_.
+
+- Print the tracks into a ``beets-unmatched`` file in the same format as in the
+  above.
+
+- Open the program `Picard`_ with the unmatched folder as an input, allowing
```
arsaboo, line 15: I am assuming this requires Picard to be installed. Please clarify that and link to the install docs.

doronbehar: > I am assuming this requires Picard to be installed. Please clarify that and link to the install docs.

Thanks, good idea. Did something.

doronbehar:
Now I'm reading the docs, and the `beet mbsubmit QUERY` command I never used has got me thinking: Perhaps it should accept an option such as `--picard` now? That would open `picard` with the files the QUERY give? Tell me what you think :)

jackwilsdon:
Can we maybe only enable this prompt choice if an option is set? Maybe only show it if `mbsubmit.picard_path` is present (we should also use this to actually find picard instead of expecting it to be on the user's PATH).

doronbehar:
Finally CI is green :)

> Can we maybe only enable this prompt choice if an option is set?

I wouldn't mind it too much personally, but why hiding this feature? Not cluttering the choices prompt?

> Maybe only show it if `mbsubmit.picard_path` is present (we should also use this to actually find picard instead of expecting it to be on the user's PATH).

I agree it'd be nice to configure it, especially for Windows users. I won't have time to work on this detail in the next 3 days though.

JOJ0:
This PR is shaping. Nice! I like @jackwilsdon suggestion! No worries with time @doronbehar, beets is a patient project (;-)))) 

When you find time to come back to here, I have som nitpicking on the usage of pathlib and path handling in general. We have some guidlines here: https://beets.readthedocs.io/en/latest/contributing.html#handling-paths

Especially the part about using `syspath` might be useful.

JOJ0:
> Finally CI is green :)
> 
> > Can we maybe only enable this prompt choice if an option is set?
> 
> I wouldn't mind it too much personally, but why hiding this feature? Not cluttering the choices prompt?

IMO, yes, because of the cluttering the choices prompt:

```
Enter search, enter Id, aBort, eDit, edit Candidates, plaY,
Print tracks, print tracks into a "beets-unmatched" File,
Open Picard with the relevent files?
```

Don't get me wrong, I like this feature and think it adds to the value of beets and submitting to musicbrainz should be well supported by beets! Great idea in general!

But I don't have an immediate idea on how this prompt could be kept shorter and be understood as well, other than hiding it and making it configurable only. Maybe if it is configurable, and an advanced beets user enables it, they might most probably have read the docs already. Which means: The texts here could be even shorter because the user is already aware of what those new choices mean! Does that make sense?

JOJ0:
> Can we maybe only enable this prompt choice if an option is set? Maybe only show it if `mbsubmit.picard_path` is present (we should also use this to actually find picard instead of expecting it to be on the user's PATH).

And yes, there needs to be some kind of error handling, currently on my machine without Picard installed:

```
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'picard'
```

Jack's suggestion that picard's install-path is the way to go. Otherwise or if it's not found, fail gracefully, and don't quit the importer.

doronbehar:
OK so I fixed the error handling issue, and I added a `picard-path` configuration key. Haven't documented it yet because I need your approval on it. Also, regarding: 

> IMO yes, because of cluttering the choices prompt:

I don't find anything bad about it personally, it doesn't feel to cluttered to me... I am open to suggestions though, especially if they'd be concrete.

JOJ0:
I think you dont need to handle unicode stuff yourself, use the provided functions in beets.utils. They are bamed: syspath, bytestring_path, displayable_path, normpath

I'm sure one of them fits your needs. It would make sure we follow beets coding guidelines.

JOJ0:
Another nitpick: What happens if a user have their album tracks not each in one directory (because of their custom path formats). The file name beets-unmatched would require some disambiguation in that case.

doronbehar:
> What happens if a user have their album not each in one directory (using --group-albums on the import command). The file name beets-unmatched would require some disambiguation in that case.

I'm not sure what you mean regarding disambiguation, but the current behavior is rather simple: Every track information is written to a `beets-unmatched` file in the same folder as the track file. I haven't tested, but if `--group-albums` is used, I assume beets enters an import process with every folder, and the `beets-unmatched` would be written to each folder, if needed. I think I had a small mistake regarding that logic that the last commit that was pushed introduced. Now I removed it in the last force push.

Besides that, I haven't found time to switch to using unicode / str / bytes conversions from `beets.util`, will find time for it next weekend.

JOJ0:
> > What happens if a user have their album not each in one directory (using --group-albums on the import command). The file name beets-unmatched would require some disambiguation in that case.
> 
> 
> 
> I'm not sure what you mean regarding disambiguation, but the current behavior is rather simple: Every track information is written to a `beets-unmatched` file in the same folder as the track file. I haven't tested, but if `--group-albums` is used, I assume beets enters an import process with every folder, and the `beets-unmatched` would be written to each folder, if needed. I think I had a small mistake regarding that logic that the last commit that was pushed introduced. Now I removed it in the last force push.


I was thinking of a situation where let's say all the tracks of 5 albums are in one source folder. The first unmatched prompt would write a beets-unmatched file. In case another unmatched prompt occurs, that beets-unmatched file would get overwritten. I didnt study the code too much to be honest.

With ambiguation I ment, making sure the beets-unmatched gets a unique filename. You could for example gen a random number and append to the name, like beets-unmatched-12345.
> 
> 
> Besides that, I haven't found time to switch to using unicode / str / bytes conversions from `beets.util`, will find time for it next weekend.

👍

beetsplug/mbsubmit.py
```diff
@@ -56,7 +60,44 @@ def __init__(self):
...
+            unmatched_path = (
+                Path(item.path.decode("utf-8")).parent / "beets-unmatched"
+            )
+            with open(unmatched_path, "a") as p:
+                p.write(format(item, fmt) + "\n")
+
+    def picard(self, session, task):
+        picard_path = which(self.config["picard-path"].as_str())
+        if picard_path is None:
+            if picard_path.startswith("/"):
```
JOJ0, line 87: This second if-condition will never return True if picard_path was None already. Maybe u wanted to use `if picard_path is not None`

beetsplug/mbsubmit.py
```diff
@@ -56,7 +60,44 @@ def __init__(self):
...
+        fmt = self.config["format"].as_str()
+        for item in task.items:
+            unmatched_path = (
+                Path(item.path.decode("utf-8")).parent / "beets-unmatched"
+            )
+            with open(unmatched_path, "a") as p:
+                p.write(format(item, fmt) + "\n")
+
+    def picard(self, session, task):
+        picard_path = which(self.config["picard-path"].as_str())
```
JOJ0, line 85: You could use the .as_pathname method from confuse here, which gives you a basic sanity check. 

An entirely different approach further down below could be to use a try except wrapper around the subprocess call. Then catch a file not found error gracefully.

JOJ0:
> OK so I fixed the error handling issue, and I added a `picard-path` configuration key. Haven't documented it yet because I need your approval on it. Also, regarding:
> 
> > IMO yes, because of cluttering the choices prompt:
> 
> I don't find anything bad about it personally, it doesn't feel to cluttered to me... I am open to suggestions though, especially if they'd be concrete.

Maybe you are right, it doesn't get much shorter and concise than what you have already.

I'm reconsidereing here a little and came to my personal conclusion that what's most important here is to encourage users to submit to MusicBrainz (and I think it's awesome that beets makes it as easy and obivour as possible!).

So at first I thought that hiding the Picard part of the prompt could be useful to unclutter the prompt but now after giving int some though I think that way: Showing the prompt might even encourage users to install Picard and help submitting stuff to MusicBrainz in the future. So why not let them try it, they get a friendly error message, and might even install Picard because of that :-)

There is a tiny typo btw. Should be "relevant" spelled with an a

```
...
Open Picard with the relevant files?
```

JOJ0:
> Let us take the middle path. We can make `picard` configurable and only show the Picard prompts if the user has Picard configured correctly; otherwise fail gracefully. 

Define "configured correctly" please :-)


> 
> 
> While we want users to submit the MB, most don't, so that it may be excessive for most people. Also, we don't want folks to submit junk there, i.e., only knowledgeable users should submit valid entries there. 

We should not forget that we are talking about the mbsubmit plugin. If a user enables it, they _do_ want to submit to MB! ☝️😆

arsaboo:
> Define "configured correctly" please :-)

Basically, we should not worry about user error (as @sampsyo mentioned in another PR). We assume that the user has provided the correct path. We can try to run and if it fails, error out gracefully. 

> We should not forget that we are talking about the mbsubmit plugin. If a user enables it, they _do_ want to submit to MB! ☝️😆

Oops....my bad. In that case, perfectly fine with showing all the prompt choices 👍

JOJ0:
> > Define "configured correctly" please :-)
> 
> 
> 
> Basically, we should not worry about user error (as @sampsyo mentioned in another PR). We assume that the user has provided the correct path. We can try to run and if it fails, error out gracefully. 
> 
Yep, totally agree. A valid path set in the conf should be sufficient to try running.

beetsplug/mbsubmit.py
doronbehar: > This second if-condition will never return True if picard_path was None already. Maybe u wanted to use `if picard_path is not None`

Thanks for pointing that out, I changed the logic there to what I meant.

doronbehar:
Hey all, these are busy times for me nowadays.. I'm not sure whether the discussions from ~2.5 weeks have reached a concrete decision. I'll point out my view after reading your discussion:

The idea below is great, and it indeed helps regarding the "cluttered prompt choices" complaint, so I pushed it:

> I do have one suggestion to make this a little shorter:
>
>     ...
>     open files with Picard

As for whether to allow this prompt choice or not, depending on whether `picard` is installed or not, and depending on the config: My view was also voiced here by @arsaboo: a user enabling the `mbsubmit` should expect that the plugin will help them submit releases to musicbrainz - per the name of the plugin. Whether they have or don't have picard installed, or whether they are even aware of it, should be their responsibility. Beets' responsibility IMO should be to not exit completely and gracefully not launch `picard` if it's not available.

As for:

> I was thinking of a situation where let's say all the tracks of 5 albums are in one source folder. The first unmatched prompt would write a beets-unmatched file. In case another unmatched prompt occurs, that beets-unmatched file would get overwritten. I didnt study the code too much to be honest.

You are right @JOJ0 - I haven't yet fully finished defining the `beets-unmatched` file behavior... As I'm thinking about it I wonder perhaps this prompt choice could be removed - then all of the prompt choices will be less cluttered. What do you think?

beetsplug/mbsubmit.py
doronbehar: > You could use the .as_filename method from confuse here, which gives you a basic sanity check.

Good idea :) Thanks.

JOJ0:
Agree with everything, work away :-)

JOJ0 (commented):
Thanks for the updates. No worries about response times. We are getting there! :-) One question within.

beetsplug/mbsubmit.py
```diff
@@ -56,7 +60,29 @@ def __init__(self):
...
+                PromptChoice("o", "Open files with Picard", self.picard),
+            ]
+
+    def picard(self, session, task):
+        picard_path_orig = self.config["picard-path"].as_filename()
+        picard_path = which(picard_path_orig)
+        if picard_path is None:
+            if picard_path_orig.startswith("/"):
+                self._log.error(
+                    "picard-path defined in config at\n"
```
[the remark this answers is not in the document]
doronbehar: > Sorry I still don't quite understand this part. Can you explain what the indented functionality is?

It's OK :) I updated this section with 2 more comments.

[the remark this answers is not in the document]
JOJ0: Hmm still I don't understand why we need that much logic to try running a path that's configured by the user anyway.


Simplification suggestion:

remove the which(),

just try running and fsil gracefully.

I think with this conversation we decided on that approach: fixme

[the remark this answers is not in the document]
JOJ0: Maybe a simplified version of like the play plugin does it. A try/except block using subprocess.run and then fail gracefully if that doesn't work out?

https://github.com/beetbox/beets/blob/951b0f576329cbf5d3646036b7e0602156024d05/beetsplug/play.py#L50-L58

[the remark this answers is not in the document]
doronbehar: > Maybe a simplified version of like the play plugin does it.

Thanks for the link :) I used it as a reference, though haven't tested in an environment with no `picard` available or so..

> using `subprocess.call`

Is there an insistence on `subprocess.call` and not `subprocess.Popen`? [Python describes `call` as an older-level API](https://docs.python.org/3/library/subprocess.html#older-high-level-api)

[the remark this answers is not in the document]
JOJ0: yeah, goot catch. then use .Popen. sounds good.

I've always used subprocess.run in the past. not sure if that makes sense here.

JOJ0:
> It'd be nice if someone else will give this PR a test drive on a machine without `picard` installed...

could be easily done by just providing a non-existent path in the config:

confuse's .path() wont fail, it just checks if "it looks like" a proper path.

when you try running with subprocess.Popen, your error should be catched and handled by the except clause.

So just to clarify, what I think would be the most straight forward way and will work on all platforms and can be tested easily by yourself:

- check if the user has a path configured by using the config handlers own sanity check (you do that already)
- now that we know we have a proper path configured we can instantly run the command within a try block.
- if it doesn'T work out we print a nicely formatted error message (in the except block)

doronbehar:
> > It'd be nice if someone else will give this PR a test drive on a machine without `picard` installed...
> 
> could be easily done by yourself, by just providing a non-existent path in the config:

You are right :) I wasn't thinking about it. Here's what I added to my config for testing:

```yaml
mbsubmit:
    picard-path: picard-missing
```

> 
> confuse's ~.path()~ .as_filename() wont fail, it just checks if "it looks like" a proper path/filename.
> 
> when you try running with subprocess.Popen, your error should be catched and handled by the except clause.
> 
> So just to clarify and prevent further misunderstandings, what I think would be the most straight forward way and will work on all platforms and can be tested easily by yourself:
> 
>   * check if the user has a path configured by using the config handler's (confuse) own sanity check (you do that already)
>   * now that we know we have a proper path configured we can instantly run the command within a try block.
>   * if it doesn'T work out we print a nicely formatted error message (in the except block)
> 
> 
> please doublecheck my assumption about ~.path()~ .as_filename() in the confuse docs: [confuse.readthedocs.io/en/latest/examples.html#path](https://confuse.readthedocs.io/en/latest/examples.html#path)

And I got the following experience:

```
➜ Apply, More candidates, Skip, Use as-is, as Tracks, Group albums,
Enter search, enter Id, aBort, Print tracks, Open files with Picard,
eDit, edit Candidates? O
error: Could not open picard due to error:
[Errno 2] No such file or directory: '/home/doron/.config/beets/picard-missing'
```

And the `beet` exits completely. I'm not satisfied with this behavior. Firstly, I don't like how `confuse`' behavior is to always get a path relative to the config. Even if I comment out the `mbsubmit:` config section, using `.as_path()` or `.as_filename()` makes it look for `picard` in the directory I'm at, and not in `$PATH`. So again, what is the reason not to use `.as_str()`?

Here is the behavior I get with an implementation similar to the previous - using `shutil.which()`, and `.as_str()`:

```
➜ Apply, More candidates, Skip, Use as-is, as Tracks, Group albums,
Enter search, enter Id, aBort, eDit, edit Candidates, Print tracks,
Open files with Picard? O
mbsubmit: couldn't find picard executable 'picard-missing'
```

Also note, that the import process doesn't stop completely when failing to run `picard-missing`, because I don't `raise ui.UserError`, as opposed to the previous behavior.

> edit: But I can certainly test it out on my machine were currently is no Picard installed, once you are finished with implementing.

Sure :) now that I learned to test it, I'll notify you when I'm satisfied with the behavior. I pushed (without `--force` this time) a certain behavior more similar to the previous behavior, but this time I tested it with the following options:

1. `mbsubmit.picard-path` is defined as `picard-missing`, causes an error message to come out, without exiting the process alltogether.
2. Using an absolute and a relative (to `$PATH`) path in `mbsubmit.picard-path` is both supported, thanks to `shutil.which`.

JOJ0:
Hi, you are right using as_filename() might not be the best choice here. My impression was that as_filename() checks for an absolute path and errors if that's not the case. Thea magic it does with relative paths is not what we want here. Sorry I should have read docs more closely.

My idea actually was that it should be required for a user to always configure an absolute path since I've not seen automatic retrieval of paths for external programs to be automatically be found in other places in beets. I might be wrong. I still think it's sufficient and most safe if don't put too much magic here since we can't test on all platforms. which() works on Windows too as far as I read in the docs, but the executable will probably could picard.exe or something. So the default of `picard` work anyway. But yeah if you want we do it with which, no problem. Default will work on Linux and Macs probably.

Because uf exiting: You don't _have_ to raise another exception, you could just print the error using `self._log.error()` - it will be seen by the user and beets will stay alive.

Your current code looks much better now. My main problem with it was probably the cascaded if conditions and especially the check for `....startswith("/") - it was just hard to read what's going on here. That's why I suggeste a simple try/except.

What do you think of something like that:

```
    def picard(self, session, task):
        picard_path = self.config["picard_path"].as_str()
        paths = []
        for p in task.paths:
            paths.append(p.decode("utf-8"))
        try:
            subprocess.Popen([picard_path] + paths)
        except OSError as exc:
            self._log.error(f"Could not open picard.\n{exc}")
```

```
mbsubmit: Could not open picard.
[Errno 2] No such file or directory: 'tmp/picard_exec'
```

JOJ0:
For decoding paths there is a utility in `beets.util` you could use:

```
from beets.util import displayable_path
...
...
            paths.append(displayable_path(p))
```

doronbehar:
Thanks for all the suggestions! I liked even more the behavior with the `try-except` you demonstrated. I hope the usage with the `displayable_path` won't cause issues in obscure encoding environments. I pushed both of your suggestions and squashed everything to the 1st commit.

> I still think it's sufficient and most safe if we don't put too much magic here since we can't test on all platforms. which() works on Windows too as far as I read in the docs, but the executable will probably be could picard.exe or something. So the default of `picard` won't work anyway. But yeah if you want we do it with which, no problem. Default will work on Linux and Macs probably.

I agree. Additionally, you should remember that even if a user installs picard and get a `picard.exe` executable somewhere in their `C:\Program Files` or whatever, it is likely this executable's directory doesn't get added to the system's `$PATH`. I also presume that when installing Picard via an installer, the user gets to choose the install location. Hence it is natural to ask all windows users to specify that path in their config.

JOJ0:
Awesome, code looking perfect now. Some nitpicking on the docs: Maybe state that picard path needs to be configured.

doronbehar:
Great :) I just installed Picard on a Windows platform and amended a sentence to the `picard_path` config option. Mainly meant for windows.

JOJ0:
That help now reads very complete and I think it tells verything one needs to know on how to use these new features. Thanks a lot.

I'm happily merging this now. At this point: Thanks a lot or baring with us with those loads of iterations and refinments. Often it is not so easy for all participants to see the most straight-forward way for a feature/plugin. After all I personally think it was worth it that we didn't rush a merge and thanks again for your patients and the good work! :-)

jackwilsdon:
@JOJ0 

> So at first I thought that hiding the Picard part of the prompt (if Picard not installed) could be useful to unclutter the prompt but now after giving it some thought, I think that way: Showing the prompt might even encourage users to install Picard and help submitting stuff to MusicBrainz in the future. So why not let them try it, they get a friendly error message, and might even install Picard because of that :-)

I (and a lot of other users) use beets on headless machines, which don't really have the option for using `picard`. I'm still in favor of hiding the prompt when Picard isn't installed.

JOJ0:
Hi @jackwilsdon, I just saw your comment about headless machines. We didn't take this intersting thought into account yet, I might have missed this perspective in our previous conversations. Please could you open a feature request issue for that? I think it won't be hard to implement. Then let's move all the discussion on how it should work over there. Sorry again I missed this part and thanks in advance!

doronbehar:
I'm also a bit surprised that headless machines were a strong reason in favor of hiding the Picard prompt choice... we talked about it long ago and I got the impression that we settled down upon not hiding it. I am also a bit surprised though that users of headless machines even use the `mbsubmit` plugin - how did the plugin help you submit tracks up until now? If the plugin is not enabled, the picard prompt choice is disabled.

The logic @jackwilsdon suggested looks not bad, and I haven't thought about it myself back when we discussed this detail. However, I still think that having this `PromptChoice` enabled anyway encourages users to use Picard to quickly submit releases to MusicBrainz. I also hardly believe that there are so many users of headless machines, and that they will consider it a significant inconvenience.

jackwilsdon:
> how did the plugin help you submit tracks up until now?

The `p` prompt prints out the track listing in a format MusicBrainz can understand, which I copy from my terminal and paste into the MusicBrainz website which is open on my local machine.

> I also hardly believe that there are so many users of headless machines, and that they will consider it a significant inconvenience.

From how many people use the [LinuxServer.io Docker image](https://hub.docker.com/r/linuxserver/beets) (over 10M pulls, although this seems quite inflated), I'd say there's probably a decent number of people using beets without a desktop environment. It's not that it's really an inconvenience, it just seems less than ideal to be showing an option that we know will throw an error.

I'm not overly bothered by it at the end of the day (not enough to open a feature request), it just seemed like it'd be nice if we didn't show an extra prompt choice if we didn't need to (as there's already a good number of choices).

PR #4809, opened 2023-06-01
Store the release group's title.

djl:
beets does not currently store the "canonical" release group title, only each release's title and disambiguation. Oddly, beets does store the release group disambiguation.

Without this there is no way to determine a release group's title without separately querying MusicBrainz with `mb_releasegroupid`.

## To Do


sampsyo:
Seems cool overall! But maybe we should just call the field `release_group_title` or similar… since it's not an MBID, it's not really all that MusicBrainz-specific (and you can imagine other data sources hypothetically supplying equivalent data). That would also avoid the need to do any special handling in `apply_metadata`; we could use the same field name all the way through.

djl:
You're totally right, I'm so used to thinking in MB terms. Changed the code and changelog to use `release_group_title` instead.

beets/autotag/__init__.py
```diff
@@ -40,6 +40,7 @@
     'album': (
         'va',
         'releasegroup_id',
+        'release_group_title',
```
sampsyo, line 43: I think we don't need this—if we leave this off, this data will get applied automatically.

beets/autotag/__init__.py
```diff
@@ -176,6 +177,7 @@ def apply_metadata(album_info: AlbumInfo, mapping: Mapping[Item, TrackInfo]):
             item.mb_artistid = album_info.artist_id
         item.mb_albumartistid = album_info.artist_id
         item.mb_releasegroupid = album_info.releasegroup_id
+        item.release_group_title = album_info.release_group_title
```
sampsyo, line 180: Regarding the above: I think we can just delete this line, and the `for field, value in album_info.items():` loop below should have the same effect.

docs/changelog.rst
```diff
@@ -88,6 +88,8 @@ New features:
 * :doc:`/plugins/convert`: Add support for generating m3u8 playlists together
   with converted media files.
   :bug:`4373`
+* Add release_group_title field
```
sampsyo, line 91:
```suggestion
* Fetch the ``release_group_title`` field from MusicBrainz.
```

sampsyo (commented):
Looks great! I think we can get away with deleting a couple of lines in the metadata application bit now that we use the same field name consistently.

djl:
Addressed all the above comments.

sampsyo (approved):
Excellent! Thanks for seeing this through!

PR #4816, opened 2023-06-06
Make sure release_group_title is applied to album objects.

djl:
Missed this while implementing #4809.


sampsyo:
Makes sense; thanks!

PR #4819, opened 2023-06-14
Fix fanarttv sorting

lonebyte:
## Description

fanart.tv uses a string to output the number of likes (see https://fanart.tv/api-docs/api-v3/). In order to sort numerically we need to convert the string into an int.

This was tested locally with `http://webservice.fanart.tv/v3/music/albums/1cbc1a5a-5512-3f8e-997b-d9281b83722b?api_key=...`

## To Do


sampsyo:
Seems like a great idea; thank you for catching this!

PR #4823, opened 2023-06-16
Streamline album attributes modification behaviour and allow override via CLI

### LINKED ISSUE #4822: Album flexible attributes modifications are not inherited to tracks
As discussed in https://github.com/beetbox/beets/discussions/4742, changing flexible attributes via a `beet modify -a` command does not inherit the changes to the tracks of the album. 

Changing a fixed attribute on an album _does_ inherit the changes to the album-tracks though.

### Problem

Running

```sh
$ beet mod -a flex_attr=new_value search term
```

Does not change `flex_attr` on the album`s tracks,

while running

```sh
$ beet mod -a genre=new_value search term  # genre is a fixed attribute available to albums an album-tracks
```

_does_ change `genre` on the album's tracks.

### Proposed solution

- Streamline the behaviour of `beet modify -a` - Fixed attributes changes as well as flexible attributes changes should be inherited to album-tracks.
- Add a CLI option that prevents inheritance, allowing changing fixed/flex attributes on an album **only**

JOJ0:
## Description

- Fixes #4822
- Keeps the default behaviour of inheriting album (fixed) attributes modifications to album tracks.
- Fixes album modifications of flexible attributes to be inherited to album-tracks as well.
- Adds a new CLI option `--noinherit/-I` that prevents inheriting to album-tracks, which allows changing fixed/flexible attributes on an album only.

## To Do

- [x] Documentation.
- [x] Changelog.
- [x] Tests.

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -1369,7 +1369,7 @@ class Album(LibModel):
 
         plugins.send('art_set', album=self)
 
-    def store(self, fields=None):
+    def store(self, fields=None, inherit=True):
         """Update the database with the album information.
 
         The album's tracks are also updated.
@@ -1381,11 +1381,11 @@ class Album(LibModel):
         track_updates = {}
         track_deletes = set()
         for key in self._dirty:
-            if key in self.item_keys:
+            if key in self.item_keys and inherit:
                 track_updates[key] = self[key]
-            elif key not in self:
+            elif key not in self and inherit:
                 track_deletes.add(key)
-            else:  # Must be a flex attr
+            elif inherit:  # Must be a flex attr
                 track_updates[key] = self[key]
 
         with self._db.transaction():
@@ -1402,7 +1402,7 @@ class Album(LibModel):
                             del item[key]
                     item.store()
 
-    def try_sync(self, write, move):
+    def try_sync(self, write, move, inherit=True):
         """Synchronize the album and its items with the database.
         Optionally, also write any new tags into the files and update
         their paths.
@@ -1411,7 +1411,7 @@ class Album(LibModel):
         `move` controls whether files (both audio and album art) are
         moved.
         """
-        self.store()
+        self.store(inherit=inherit)
         for item in self.items():
             item.try_sync(write, move)
 
--- a/beets/library.py
+++ b/beets/library.py
@@ -1385,7 +1385,7 @@ class Album(LibModel):
                 track_updates[key] = self[key]
             elif key not in self and inherit:
                 track_deletes.add(key)
-            elif inherit:  # Must be a flex attr
+            elif key != 'id' and inherit:  # Could be a flex attr or id
                 track_updates[key] = self[key]
 
         with self._db.transaction():
--- a/beets/library.py
+++ b/beets/library.py
@@ -1385,6 +1385,8 @@ class Album(LibModel):
                 track_updates[key] = self[key]
             elif key not in self:
                 track_deletes.add(key)
+            else:  # Must be a flex attr
+                track_updates[key] = self[key]
 
         with self._db.transaction():
             super().store(fields)
--- a/beets/library.py
+++ b/beets/library.py
@@ -1383,12 +1383,13 @@ class Album(LibModel):
         track_updates = {}
         track_deletes = set()
         for key in self._dirty:
-            if key in self.item_keys and inherit:  # Fixed attr
-                track_updates[key] = self[key]
-            elif key not in self and inherit:  # Fixed or flex attr
-                track_deletes.add(key)
-            elif key != 'id' and inherit:  # Could be a flex attr or id (fixed)
-                track_updates[key] = self[key]
+            if inherit:
+                if key in self.item_keys:  # is a fixed attribute
+                    track_updates[key] = self[key]
+                elif key not in self:  # is a fixed or a flexible attribute
+                    track_deletes.add(key)
+                elif key != 'id':  # is a flexible attribute
+                    track_updates[key] = self[key]
 
         with self._db.transaction():
             super().store(fields)
--- a/beets/library.py
+++ b/beets/library.py
@@ -1372,20 +1372,22 @@ class Album(LibModel):
     def store(self, fields=None, inherit=True):
         """Update the database with the album information.
 
-        The album's tracks are also updated.
-
         `fields` represents the fields to be stored. If not specified,
         all fields will be.
+
+        The album's tracks are also updated when the `inherit` flag is enabled.
+        This applies to fixed attributes as well as flexible ones. The `id`
+        attribute of the album will never be inherited.
         """
         # Get modified track fields.
         track_updates = {}
         track_deletes = set()
         for key in self._dirty:
-            if key in self.item_keys and inherit:
+            if key in self.item_keys and inherit:  # Fixed attr
                 track_updates[key] = self[key]
-            elif key not in self and inherit:
+            elif key not in self and inherit:  # Fixed or flex attr
                 track_deletes.add(key)
-            elif key != 'id' and inherit:  # Could be a flex attr or id
+            elif key != 'id' and inherit:  # Could be a flex attr or id (fixed)
                 track_updates[key] = self[key]
 
         with self._db.transaction():
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1489,7 +1489,7 @@ default_commands.append(version_cmd)
 
 # modify: Declaratively change metadata.
 
-def modify_items(lib, mods, dels, query, write, move, album, confirm):
+def modify_items(lib, mods, dels, query, write, move, album, confirm, inherit):
     """Modifies matching items according to user-specified assignments and
     deletions.
 
@@ -1542,7 +1542,7 @@ def modify_items(lib, mods, dels, query, write, move, album, confirm):
     # Apply changes to database and files
     with lib.transaction():
         for obj in changed:
-            obj.try_sync(write, move)
+            obj.try_sync(write, move, inherit)
 
 
 def print_and_modify(obj, mods, dels):
@@ -1585,7 +1585,8 @@ def modify_func(lib, opts, args):
     if not mods and not dels:
         raise ui.UserError('no modifications specified')
     modify_items(lib, mods, dels, query, ui.should_write(opts.write),
-                 ui.should_move(opts.move), opts.album, not opts.yes)
+                 ui.should_move(opts.move), opts.album, not opts.yes,
+                 opts.inherit)
 
 
 modify_cmd = ui.Subcommand(
@@ -1612,6 +1613,10 @@ modify_cmd.parser.add_format_option(target='item')
 modify_cmd.parser.add_option(
     '-y', '--yes', action='store_true',
     help='skip confirmation'
+)
+modify_cmd.parser.add_option(
+    '-I', '--noinherit', action='store_false', dest='inherit', default=True,
+    help="Don't inherit album-changes to tracks"
 )
 modify_cmd.func = modify_func
 default_commands.append(modify_cmd)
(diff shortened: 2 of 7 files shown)


JOJ0:
Hi @sampsyo we talked about this a couple of months ago and agreed that it would make sense to have the same behaviour of inheritance of album-mods no matter if fixed or flex attr https://github.com/beetbox/beets/discussions/4742#discussioncomment-5565092

Now before I add documentation and changelog to this PR I would like to discuss a usability thing: I just added a capital option of `-I` to `beets modify`. Would it make sense to add a non-capital option `-i` that states `Inherit album-changes to tracks (default)" in its help text? One one hand it would make it clear to the user immediately what the default behaviour is, on the other `-i` would be kind of never to be used, since it's the default anyway and it's not configurable. Am I making sense? I just want to make sure we don't overengineer this feature and make it as logical as possible. Certainly docs of `modify` will mention what the default behaviour is, and how it can be changed.

sampsyo:
Thanks for getting this started!

At a very high level, I think it may make more sense to make the "inheritance" happen when making the change on the model, not when calling `store`. The `store` command is supposed to mean "take these temporary changes and commit to them, storing them to the on-disk database." In general, it is not supposed to change the values of fields; it is just supposed to *store* the changes that have already been made. Does that make any sense?

This could happen either in `__setitem__`, which we already override for `Item` although not `Album` yet:
https://github.com/beetbox/beets/blob/0c3f428a601cb40c5fd463791df6229d51b0635e/beets/library.py#L633

Or in the `modify` command, if we think this is a `modify`-specific behavior we want.

wisp3rwind:
> At a very high level, I think it may make more sense to make the "inheritance" happen when making the change on the model, not when calling `store`. The `store` command is supposed to mean "take these temporary changes and commit to them, storing them to the on-disk database." In general, it is not supposed to change the values of fields; it is just supposed to _store_ the changes that have already been made. Does that make any sense?

My initial reaction was to agree with you, but after a brief glance at the code, I'm not so sure anymore: `Album.store` already has the `track_updates`/`track_deletes` mechanism. Even if one might design this differently now, it would also be unfortunate to introduce a second code path that propagates tags to items. Also, if I'm not mistaken, `Album`s don't keep their tracks around, but `Album.items()` fetches them freshly from the DB. I doubt we would want to do that in each `__setitem__` call, but once per `store` seems reasonable (and, is already being done for `track_updates`).

sampsyo:
Hmm; you're right about that—I hadn't thought that through completely (that `store` is where we currently "propagate" changes as it stands). Which means that you can think of synchronizing the album and item metadata as part of the "sync" that `store` represents. Thanks for catching this!

JOJ0:
Well, thanks to both of you for clarifying. I have to admit that @sampsyo's first suggestion confused me a little, but no worries, I'm glad we seem to be back on track :-)

Good, before I move on to wrapping my head around all the tests that are failing now, any other suggestions if there is a better way than passing a kwarg through a few levels to make this "configurable"?

Also any opinions on whether or not it makes sense to add an additional `-i` option. I would suggest to not do that. A single option `-I` (capital since we often see that in beets for negation) I'd consider sufficient.

arsaboo:
@JOJ0 I thought `-I` is used for non-incremental or something. This is my go-to import command `beet import -m -I -t ~/shared/music/`

sampsyo:
> any other suggestions if there is a better way than passing a kwarg through a few levels to make this "configurable"?

The main way would be to make this a proper beets configuration option (i.e., make it appear in `config_default.yaml`). These can be overridden by command-line flags.

JOJ0:
> @JOJ0 I thought `-I` is used for non-incremental or something. This is my go-to import command `beet import -m -I -t ~/shared/music/`

Hmm, yes that's true but actually I thought that its not problem to have the same letter for a different subcommand. ‘beet modify -I’ is not ‘beet import -I‘

But yeah if you guys can think of a different letter I'm up for suggestions.

‘beet mod -n‘ better? (but -n often is somethin like --no-op in a lot of cli tools, didnt like it because of that)

arsaboo:
If it does not conflict, I am fine with `beet modify -I`

JOJ0:
> > any other suggestions if there is a better way than passing a kwarg through a few levels to make this "configurable"?
> 
> The main way would be to make this a proper beets configuration option (i.e., make it appear in `config_default.yaml`). These can be overridden by command-line flags.

Sorry for being unclear with my question. I used the word configurable but put it in double quotes because I was trying to say something like: ....any better way...to implement this?

Well, you did kind of answer to that as well. A regular beets config option that a cli option overrides is probably a typicial form of implementing this. Now my initial idea was to have a config option but after thinking this through a couple of times I came to the conclusion that actually I  consider it really necessary. If I want to modify something, then a consistent default should be used - which we finally achieve with these changes. Finally album mods propagate to items for flex attrs as well (not touching the current behaviour for fixed attrs). If this is not what the user wants, they can use this new cli option to modify.

That werer my thoughts, also trying to keep this change minimal. We can certainly discuss it. Maybe having an option does not hurt....

Any suggestions where an option like this would "live" in config_default.yaml?

Ideas:

```
modify:
    inherit: yes
```

```
modify_inherit: yes
```

```
inherit_modifications: yes
```

```
album_to_item_mods: yes
```

And now, _if_ we would add such an option it would further be required to have a second cli option that can alter the behaviour.

Suggestion:

If user config option set to `no`, this can be used to alter:

`beet modify -i`

If user config option set to `yes` (or using the default), this can be used to alter:

`beet modify -I`

Thoughts welcome! Thanks!

arsaboo:
I prefer:
```yaml
modify:
    inherit: yes
```
I am also thinking that the default behavior should be to inherit attributes.

JOJ0:
Next up is this test:
https://github.com/beetbox/beets/blob/b19b961035da23a7cbe2a049d2e3f19e7ce220d5/test/test_library.py#L1025-L1030

Failing like so:
```
FAIL: test_albuminfo_change_artist_does_not_change_items (test.test_library.AlbumInfoTest.test_albuminfo_change_artist_does_not_change_items)
  File "/home/jojo/git/beets/test/test_library.py", line 1030, in test_albuminfo_change_artist_does_not_change_items
    self.assertNotEqual(i.artist, 'myNewArtist')
AssertionError: 'myNewArtist' == 'myNewArtist'

Ran 12 tests in 0.216s

FAILED (failures=1)
```

It seems like inheriting change of artist (of the album) shouldn't be inherited to items which this check proves.

I could easily fix this test by passing `inherit=False` to the `store()` when running the test:
https://github.com/beetbox/beets/pull/4823/commits/9d0a4ddb12a5b80b2f6b2a2d28f0164d3f4bf102

but is this what we want in general: Should editing of artist always not be inherited to items?

On the other hand editing of an albumartist (of an album) is checked whether succesfully propagated to items with this test:
https://github.com/beetbox/beets/blob/b19b961035da23a7cbe2a049d2e3f19e7ce220d5/test/test_library.py#L1017-L1023

sampsyo:
> I'm not sure if this is what we want. High level question: I would say the 'id' field should always be prevented to propagate from album mods to items? If yes, there might be other fields that should always not be inherited. I could put a CONSTANT somewhere that collects these fields. Would that be the right way?

Good question. Yes, absolutely; we would never ever want the `id` field to propagate to items; that does indeed seem like a recipe for confusion. But it's a good question to ask whether there are _other_ things that are similar… my gut reaction says "no, `id` stands alone," but I could very well be proven wrong. Maybe we should just watch out for other related effects where this kind of alignment is clearly undesirable?

> I could easily fix this test by passing `inherit=False` to `store()` when running the test: [9d0a4dd](https://github.com/beetbox/beets/commit/9d0a4ddb12a5b80b2f6b2a2d28f0164d3f4bf102)
> 
> but is this what we want in general? Should editing of artist perhaps never propagate to items?

It seems like this test was expressly meant to "lock in" the old behavior, which we are now changing. I think what seems reasonable here is that we should split this into two tests: one with `inherit=True` and one with `inherit=False`, and we check that the resulting behavior differs in the expected way between the two.

JOJ0:
Ready for a final review @sampsyo, maybe have a look at the docs I wrote, I also reworded existing stuff for `mod -a` a little. Is that easily understandable and correct like that: https://github.com/beetbox/beets/pull/4823/commits/d30ac6032cbdc317a479ec760acef3e0da411c52

I left the check for ìd`simply inline with the if-statement. An extra GLOBAL_NEVER_INHERIT_THIS_FIELD or something like that felt overkill: https://github.com/beetbox/beets/pull/4823/commits/7313fe1d5f3806e2612ee45b3578f9ed7068da64

As a final step I tried to clarify and extend docstring and comments within this function. Nitpicking appreciated since it is a very very core-thing everybody who comes accross it should have the chance to understand quickly. What do you say? Does that make sense? https://github.com/beetbox/beets/pull/4823/commits/a6937f08f293f04a32a127204855ef643afd1e1d

beets/library.py
```diff
@@ -1369,22 +1369,26 @@ def set_art(self, path, copy=True):
...
         track_deletes = set()
         for key in self._dirty:
-            if key in self.item_keys:
+            if key in self.item_keys and inherit:  # Fixed attr
                 track_updates[key] = self[key]
-            elif key not in self:
+            elif key not in self and inherit:  # Fixed or flex attr
                 track_deletes.add(key)
+            elif key != 'id' and inherit:  # Could be a flex attr or id (fixed)
+                track_updates[key] = self[key]
```
sampsyo, line 1391: I think you alluded to this in the main comment thread, but this could be a little easier to read if we put the entire stanza under a `if inherit:` condition. That would make it clear to the reader that _nothing_ happens here when `inherit=False`.

beets/ui/commands.py
```diff
@@ -1613,6 +1614,10 @@ def modify_func(lib, opts, args):
     '-y', '--yes', action='store_true',
     help='skip confirmation'
 )
+modify_cmd.parser.add_option(
+    '-I', '--noinherit', action='store_false', dest='inherit', default=True,
+    help="Don't inherit album-changes to tracks"
```
sampsyo, line 1619:
```suggestion
    help="when modifying albums, don't also change item data"
```

Just trying something even more explicit! To clarify that this does nothing when in item mode.

docs/changelog.rst
```diff
@@ -207,6 +207,11 @@ Bug fixes:
   :bug:`4726`
 * :doc:`/plugins/fetchart`: Correctly select the cover art from fanart.tv with
   the highest number of likes
+* Modifying flexible attributes of albums now cascade to the individual album
+  tracks, similar to how fixed album attributes have been cascading to tracks
+  already. A new option ``--inherit/-I`` to :ref:`modify <modify-cmd>`
```
sampsyo, line 212:
```suggestion
  already. A new option ``--noinherit/-I`` to :ref:`modify <modify-cmd>`
```

docs/reference/cli.rst
```diff
@@ -274,13 +274,17 @@ name into the artist field for all your tracks,
...
+The ``-a`` option changes to querying album fields instead of track fields and
+also enables to operate on albums in addition to the individual tracks.
 Without this flag, the command will only change *track-level* data, even if all
 the tracks belong to the same album. If you want to change an *album-level*
 field, such as ``year`` or ``albumartist``, you'll want to use the ``-a`` flag
 to avoid a confusing situation where the data for individual tracks conflicts
 with the data for the whole album.
 
+Modifications issued using ``-a`` by default cascade to individual tracks. To
+prevent this behaviour, ``-I/--noinherit`` can be used.
```
sampsyo, line 286:
```suggestion
prevent this behavior, use ``-I``/``--noinherit``.
```

Mostly just for the avoidance of passive voice here. :smiley:

sampsyo (commented):
Nice!!! This is looking great. I have just a couple of small suggestions, but I think we are basically in business!

beets/library.py
JOJ0: Good point! I changed it to this: https://github.com/beetbox/beets/pull/4823/files#diff-d6c57a8deb6bf406de08bd231833fafb4515c984bf0d854a50a0e6cb3dcd3d1fR1386-R1392

JOJ0:
That was a tough one but I think it was worth it. Thanks @sampsyo for all the guidance!

PR #4826, opened 2023-06-23
typings: corrections for dbcore/queryparse

wisp3rwind:
Some improvements to dbcore typings, see the commit messages for details and rationale.

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -16,10 +16,23 @@
 """
 
 from __future__ import annotations
+
+from abc import ABC, abstractmethod
 import re
 from operator import mul
-from typing import Union, Tuple, List, Optional, Pattern, Any, Type, Iterator,\
-    Collection, MutableMapping, Sequence
+from typing import (
+    Any,
+    Collection,
+    Iterator,
+    List,
+    MutableMapping,
+    Optional,
+    Pattern,
+    Sequence,
+    Tuple,
+    Type,
+    Union,
+)
 
 from beets import util
 from datetime import datetime, timedelta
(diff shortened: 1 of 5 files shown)


beets/dbcore/query.py
```diff
@@ -93,7 +110,21 @@ def __eq__(self, other) -> bool:
         return type(self) == type(other)
 
     def __hash__(self) -> int:
-        return 0
+        """Minimalistic default implementation of a hash.
+
+        Given the implementation if __eq__ above, this is
+        certainly correct.
+        """
+        return hash(type(self))
```
sampsyo, line 118: It may not be much, but it's better than `0`! :smiley:

beets/dbcore/query.py
```diff
@@ -93,7 +110,21 @@ def __eq__(self, other) -> bool:
...
+
+
+class NamedQuery(Query):
+    """Non-field query, i.e. the query prefix is not a field but identifies the
+    query class.
+    """
+    @abstractmethod
+    def __init__(self, pattern):
+        ...
```
sampsyo, line 127: This is interesting: it's a good way to capture what is in common between queries that have a pattern but no field (i.e., are "half" of a `FieldQuery`). This makes sense, and it's a good way to deal with the typing requirements.

beetsplug/playlist.py
```diff
@@ -65,7 +66,7 @@ def __init__(self, pattern):
             f.close()
             break
 
-    def col_clause(self):
+    def clause(self) -> Tuple[Optional[str], Sequence[Any]]:
```
sampsyo, line 69: Wow, very nice catch. That's the cost of having "transparent" fallback for queries with no SQL implementation; you don't notice when the SQL implementation is just silently not being used.

sampsyo (approved):
Looks excellent! Nice work all around (with a couple of comments inline).

PR #4827, opened 2023-06-23
dbcore/query: remove spurious dependency on library

wisp3rwind:
Slight cleanup of dbcore typings:

> It seems like `Query`s and `Sort`s should really be generic over `Model`.
>
> Also, make use of the postponed evaluation of annotations that we get with `from future import __annotations__`.


sampsyo (approved):
Awesome; nice work on this! And good catches in the docstrings as well.

PR #4828, opened 2023-06-23
typing: corrections for dbcore/types.py

wisp3rwind:
Yet another follow-up to dbcore typing. I've had this in the pipeline for a while, but only now got around to finishing it up and figure out the last bits around handling the `model_type` `Type`-valued field.

Please have a close look at this PR; does the approach look sensible? I restructured the class hierarchy a bit to avoid some conflicting signature. This could potentially break plugins that depend on details of the db and type handling (as it turns out, the edit plugin does).

### DIFF
--- a/beets/dbcore/types.py
+++ b/beets/dbcore/types.py
@@ -16,8 +16,9 @@
 """
 
 from abc import ABC
+import sys
 import typing
-from typing import Any, cast, Generic, List, TypeVar, Union
+from typing import Any, cast, Generic, List, TYPE_CHECKING, TypeVar, Union
 from .query import BooleanQuery, FieldQuery, NumericQuery, SubstringQuery
 from beets.util import str2bool
 
@@ -25,13 +26,18 @@ from beets.util import str2bool
 # Abstract base.
 
 
-class ModelType(Protocol):
-    """Protocol that specifies the required constructor for model types, i.e.
-    a function that takes any argument and attempts to parse it to the given
-    type.
-    """
-    def __init__(self, value: Any = None):
-        ...
+# FIXME: unconditionally define the Protocol once we drop Python 3.7
+if TYPE_CHECKING and sys.version_info >= (3, 8):
+    class ModelType(typing.Protocol):
+        """Protocol that specifies the required constructor for model types,
+        i.e. a function that takes any argument and attempts to parse it to the
+        given type.
+        """
+        def __init__(self, value: Any = None):
+            ...
+else:
+    # No structural subtyping in Python < 3.8...
+    ModelType = Any
 
 
 # Generic type variables, used for the value type T and null type N (if
--- a/beets/dbcore/types.py
+++ b/beets/dbcore/types.py
@@ -15,28 +15,47 @@
 """Representation of type information for DBCore model fields.
 """
 
-from typing import Union, Any, Callable
-from . import query
+from abc import ABC
+import typing
+from typing import Any, cast, Generic, List, TypeVar, Union
+from .query import BooleanQuery, FieldQuery, NumericQuery, SubstringQuery
 from beets.util import str2bool
 
 
 # Abstract base.
 
-class Type:
+
+class ModelType(Protocol):
+    """Protocol that specifies the required constructor for model types, i.e.
+    a function that takes any argument and attempts to parse it to the given
+    type.
+    """
+    def __init__(self, value: Any = None):
+        ...
+
+
+# Generic type variables, used for the value type T and null type N (if
+# nullable, else T and N are set to the same type for the concrete subclasses
+# of Type).
+N = TypeVar("N")
+T = TypeVar("T", bound=ModelType)
+
+
+class Type(ABC, Generic[T, N]):
     """An object encapsulating the type of a model field. Includes
     information about how to store, query, format, and parse a given
     field.
     """
 
-    sql = 'TEXT'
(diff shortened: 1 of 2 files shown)


beets/dbcore/types.py
```diff
@@ -45,26 +64,26 @@ class Type:
     """
 
     @property
-    def null(self) -> model_type:
+    def null(self) -> N:
         """The value to be exposed when the underlying value is None.
         """
-        return self.model_type()
+        return cast(N, self.model_type())
```
sampsyo, line 76: It's overall very interesting that we ended up with null values that don't match the "main" type… like, I know it's not the case, but you would imagine that the null value would always agree with the underlying Python type that the `Type` represents. Maybe this is a good nudge to revisit this in the future and someday make this the case, which could reduce some unintentional weirdness.

On the other hand, maybe it is exactly what we want, in the sense that the actual Python type of the underlying values should be considered to be `Union[N, T]`. That lets us deal with `T` in the common case but occasionally provide `N`, knowing that `Union[None, T]` is the same as `Optional[T]`, and `Union[T, T]` is the same as `T`.

beets/dbcore/types.py
```diff
@@ -168,52 +209,66 @@ def format(self, value: int) -> str:
         return '{}{}'.format((value or 0) // self.unit, self.suffix)
 
 
-class Id(Integer):
+class Id(NullInteger):
```
sampsyo, line 218: I understand why this is slightly worrisome, but it is clearly the right thing to do and it's hard to imagine how this could actually break anything…

beets/dbcore/types.py
```diff
@@ -168,52 +209,66 @@ def format(self, value: int) -> str:
...
         self.digits = digits
 
-    def format(self, value: float) -> str:
+    def format(self, value: Union[float, N]) -> str:
         return '{0:.{1}f}'.format(value or 0, self.digits)
 
 
-class NullFloat(Float):
+class Float(BaseFloat[float]):
+    pass
```
sampsyo, line 247: Nice; this is a particularly good idea to make a new base class here. You could imagine leaving `null` unimplemented in the base, which would more closely match `BaseInteger`, but it doesn't seem to matter too much.

beets/dbcore/types.py
```diff
@@ -168,52 +209,66 @@ def format(self, value: int) -> str:
...
-class String(Type):
+class BaseString(Type[T, N]):
     """A Unicode string type.
     """
     sql = 'TEXT'
-    query = query.SubstringQuery
+    query = SubstringQuery
 
-    def normalize(self, value: str) -> str:
+    def normalize(self, value: N) -> Union[T, N]:
```
sampsyo, line 264: Perhaps `value` should have type `Any`, as it does elsewhere?

beets/dbcore/types.py
```diff
@@ -168,52 +209,66 @@ def format(self, value: int) -> str:
...
 
 
-class DelimitedString(String):
+class String(BaseString[str, Any]):
+    """A Unicode string type.
+    """
+    model_type = str
+
+
+class DelimitedString(BaseString[list, Any]):
```
sampsyo, line 277: We may be at the point of diminishing returns for static typing here, but perhaps we should restrict `T` here to be `list[str]`? It's possible that the null type here should also always be the empty string, meaning that this would subclass `BaseString[list[str], list[str]]`, perhaps.

sampsyo (commented):
Wow, this is very tricky indeed! It was a really good idea to parameterize `Type` on both a "main" type and a null type; this seems to help clarify types work in general (i.e., on "input," you're supposed to provide the main type; on "output," you need to expect that you might get the null type as well). Really nice work nailing this down!

I have few assorted thoughts scattered throughout.

beets/dbcore/types.py
wisp3rwind: Good catch, thanks!

beets/dbcore/types.py
wisp3rwind: Done! Matching `Float` and `Integer` makes a lot of sense here.

beets/dbcore/types.py
wisp3rwind: :+1:

beets/dbcore/types.py
wisp3rwind: I think currently, we can't avoid this: We do have types where e.g. `T = int` and `N = None`, i.e.

> in the sense that the actual Python type of the underlying values should be considered to be `Union[N, T]`

Maybe there's a world where we would have a more generic way for values to be either of `model_type` or `None`, instead of having the various explicit NULL-able and on-NULL-able types? Actually, I don't really know my way around beets auto-tagging super well and can't tell at the moment why we can't just allow all types to be `None` (as in not set).

wisp3rwind: The `cast` here is a particularly ugly hack: In the abstract `Type` class, `N` and `T` may in fact be incompatible. This default implementation only works for the case `N = T`. More cleanly, this should be done in a subclass. Alternatively, we could add a `null_type` similar to `model_type` and use a `Protocol` to restrict it's constructor signature.

For now, I think I'll leave it as-is to keep the code simple.

wisp3rwind:
Thanks for the swift and thorough review! I've addressed the comments and rebased, let's see what CI says.

sampsyo (approved):
Awesome; this looks great!!

PR #4829, opened 2023-06-24
test_ui: Fix spurious warnings in completion test

wisp3rwind:
For some reason, the completion test is always failing for me locally. While this PR doesn't actually fix that; at least it removes a spurious warning.


sampsyo (approved):
Looks like clearly the right thing to me (i.e., paths are bytes, and we use `syspath` consistently before calling `isfile`).

PR #4830, opened 2023-06-24
Always use syspath conversions (#3690 split up, part 1)

wisp3rwind:
I've had #3690 in the pipeline for a long time; and it has grown too big to look at in one piece. This is just the first commit rebased on master. It is still quite big, but only touches test code, so should be safe (assuming CI is green).

### DIFF
--- a/test/_common.py
+++ b/test/_common.py
@@ -29,6 +29,7 @@ import beets.library  # noqa: E402
 from beets import importer, logging  # noqa: E402
 from beets.ui import commands  # noqa: E402
 from beets import util  # noqa: E402
+from beets.util import bytestring_path, syspath  # noqa: E402
 import beets  # noqa: E402
 
 # Make sure the development versions of the plugins are used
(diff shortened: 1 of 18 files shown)


sampsyo (approved):
Awesome; thanks for splitting this off. There are obviously a lot of lines in the diff, but I scanned over everything and it checks out at a glance.

The one weird thing in here is the identification of some trouble with the `thumbnails` plugin. That plugin is a surprisingly complicated bugaboo relative to how simple (and perhaps how widely used) it is; it could certainly use an overhaul but may not have enough users to have anyone interested in doing it.

wisp3rwind:
> The one weird thing in here is the identification of some trouble with the `thumbnails` plugin. That plugin is a surprisingly complicated bugaboo relative to how simple (and perhaps how widely used) it is; it could certainly use an overhaul but may not have enough users to have anyone interested in doing it.

I haven't looked at the thumbnails plugin recently; so can't really tell anything about it's state. However, once I wrap up the rest of #3690, the path handling should hopefully be addressed.

PR #4831, opened 2023-06-27
Add beets-more to community plugins

TheLastZombie:
I wrote a plugin that scans MusicBrainz and Discogs for versions of releases that have more tracks than the ones in the user's database, such as deluxe, expanded or anniversary editions. For example:

```
beet more "Electric Light Orchestra"
```

```
more: Electric Light Orchestra - Discovery
more:     Fetching data from MusicBrainz...
more:     Version with 12 tracks found (have 9).
more:     See https://musicbrainz.org/release/23bf913d-3709-4f84-a3df-f19d3928cb2b​.
more: Electric Light Orchestra - Out of the Blue
more:     Fetching data from MusicBrainz...
more:     Version with 20 tracks found (have 17).
more:     See https://musicbrainz.org/release/9a86feb3-66db-3946-a477-67312641cc93​.
```

This PR adds it to the community plugins list.

### DIFF
--- a/docs/plugins/index.rst
+++ b/docs/plugins/index.rst
@@ -543,4 +543,4 @@ Here are a few of the plugins written by the beets community:
 .. _beets-originquery: https://github.com/x1ppy/beets-originquery
 .. _drop2beets: https://github.com/martinkirch/drop2beets
 .. _beets-audible: https://github.com/Neurrone/beets-audible
-.. _beets-more: https://forgejo.sny.sh/sun/beetsplug
+.. _beets-more: https://forgejo.sny.sh/sun/beetsplug/src/branch/main/more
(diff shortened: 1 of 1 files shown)


docs/plugins/index.rst
```diff
@@ -540,3 +543,4 @@ Here are a few of the plugins written by the beets community:
 .. _beets-originquery: https://github.com/x1ppy/beets-originquery
 .. _drop2beets: https://github.com/martinkirch/drop2beets
 .. _beets-audible: https://github.com/Neurrone/beets-audible
+.. _beets-more: https://forgejo.sny.sh/sun/beetsplug
```
arsaboo, line 546: Looks like you have all the plugins in the same folder. Can you please separate it out into a separate plugin so that users can install only the plugin they want? It is just a matter of reorganizing the files.

TheLastZombie: Sure. I also updated the URL to point to the specific plugin directory.

PR #4842, opened 2023-07-09
Import additional fields from Deezer and add function to update rank

### LINKED ISSUE #4841: Import rank from Deezer
Deezer api also provides `rank` information. Rank is the indicator of track popularity (ranging from 1 to 1 million) - the higher the rank is the more popular the track is. 

We should import this information as well.

arsaboo:
## Description

Fixes #4841 
Imports few other fields provided by Deezer. 

In addition, I added a function `deezerupdate` that updates the rank information without going through the import process. The rank information is updated daily and users may want to refresh the rank information periodically. 

## To Do

### DIFF
--- a/beetsplug/deezer.py
+++ b/beetsplug/deezer.py
@@ -76,7 +76,7 @@ class DeezerPlugin(MetadataSourcePlugin, BeetsPlugin):
             return None
 
         album_data = requests.get(self.album_url + deezer_id).json()
-        artist, artist_id = self.get_artist(album_data['contributors'])
+        artist, artist_id = self.get_artist(album_data.get('contributors'))
 
         release_date = album_data['release_date']
         date_parts = [int(part) for part in release_date.split('-')]
--- a/beetsplug/deezer.py
+++ b/beetsplug/deezer.py
@@ -192,6 +192,10 @@ class DeezerPlugin(MetadataSourcePlugin, BeetsPlugin):
             if deezer_id is None:
                 return None
             track_data = requests.get(self.track_url + deezer_id).json()
+            if 'error' in track_data:
+                self._log.debug(f"Error fetching track {track_id}: "
+                                f"{track_data['error']['message']}")
+                return None
         track = self._get_track(track_data)
 
         # Get album's tracks to set `track.index` (position on the entire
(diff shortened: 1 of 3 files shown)


beetsplug/deezer.py
```diff
@@ -54,7 +76,15 @@ def album_for_id(self, album_id):
             return None
 
         album_data = requests.get(self.album_url + deezer_id).json()
-        artist, artist_id = self.get_artist(album_data['contributors'])
+        if 'error' in album_data:
+            self._log.debug(f"Error fetching album {album_id}: "
+                            f"{album_data['error']['message']}")
+            return None
```
sampsyo, line 82: Cool; this seems like an important addition! Maybe it deserves its own changelog bullet point?

beetsplug/deezer.py
```diff
@@ -129,14 +160,18 @@ def _get_track(self, track_data):
...
             artist=artist,
             artist_id=artist_id,
             length=track_data['duration'],
             index=track_data.get('track_position'),
             medium=track_data.get('disk_number'),
+            deezer_track_rank=track_data.get('rank'),
             medium_index=track_data.get('track_position'),
             data_source=self.data_source,
             data_url=track_data['link'],
+            deezer_updated=time.time(),
```
sampsyo, line 174: I don't 100% see yet the need for this field… we only seem to be writing to it, never reading from it. Is it useful for some future functionality?

docs/changelog.rst
```diff
@@ -11,6 +11,8 @@ for Python 3.6).
 
 New features:
 
+* Deezer: Import rank and other attributes from Deezer during import and add a function to update the rank of existing items.
```
sampsyo, line 14:
```suggestion
* :doc:`/plugins/deezer`: Import rank and other attributes from Deezer during import and add a function to update the rank of existing items.
```

sampsyo (commented):
Looks pretty good overall! I just had a couple of minor questions based on a review.

beetsplug/deezer.py
arsaboo: Yes, it is useful to only update based on time elapsed. For example, update the tracks that have not been updated in the last 2 months. In the absence of this information, we will have to update all the tracks.

beetsplug/deezer.py
arsaboo: Not sure if this is required, but added it anyways :)

beetsplug/deezer.py
sampsyo: Got it; that makes sense!

sampsyo (approved):
All looks good; thanks!!

PR #4847, opened 2023-07-15
typings: Fixes and improvements for dbcore.query

wisp3rwind:
The next iteration of introducing typings into dbcore. This is splitting out only `dbcore.query`; with some additional changes, I have clean mypy on `dbcore`, but that'd be a bit too much for a single PR.

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -23,14 +23,16 @@ from operator import mul
 from typing import (
     Any,
     Collection,
+    Generic,
     Iterator,
     List,
-    MutableMapping,
+    MutableSequence,
     Optional,
     Pattern,
     Sequence,
     Tuple,
     Type,
+    TypeVar,
     Union,
 )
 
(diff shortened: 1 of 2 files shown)


beets/dbcore/query.py
```diff
@@ -257,25 +264,26 @@ def string_match(cls, pattern: str, value: str) -> bool:
...
-            self.pattern = re.compile(self.pattern)
+            pattern_re = re.compile(pattern)
         except re.error as exc:
             # Invalid regular expression.
             raise InvalidQueryArgumentValueError(pattern,
                                                  "a regular expression",
                                                  format(exc))
 
-    def col_clause(self):
+        super().__init__(field, pattern_re, fast)
```
sampsyo, line 284: [non-actionable] The diff confused me for a sec here, so for future reviewers: here's where the `super` call moved to.

beets/dbcore/query.py
```diff
@@ -290,47 +298,57 @@ def string_match(cls, pattern: Pattern, value: str) -> bool:
...
+                bytes_pattern = pattern
+            self.buf_pattern = memoryview(bytes_pattern)
         elif isinstance(self.pattern, memoryview):
             self.buf_pattern = self.pattern
-            self.pattern = bytes(self.pattern)
+            bytes_pattern = bytes(self.pattern)
+        else:
+            raise ValueError("pattern must be bytes, str, or memoryview")
+
+        super().__init__(field, bytes_pattern)
```
sampsyo, line 343: Another solid improvement to the old version, where the mutation was admittedly hard to follow.

sampsyo (approved):
Excellent! This looks right everywhere, according to my "brain type checker."

beets/dbcore/query.py
wisp3rwind: That's really just pleasing mypy, which does not like shadowing of variables at all.

PR #4848, opened 2023-07-16
Always use syspath conversions (#3690 split up, part 2)

wisp3rwind:
Follow-up to #4830:

This is part 2, which mostly concerns plugins (except for a small fix to the `artresizer` module such that it's internally consistent).

### DIFF
--- a/beets/util/artresizer.py
+++ b/beets/util/artresizer.py
@@ -460,11 +460,11 @@ class PILBackend(LocalBackend):
 
         # FIXME: Detect and handle other file types (currently, the only user
         # is the thumbnails plugin, which generates PNG images).
-        im = Image.open(file)
+        im = Image.open(syspath(file))
         meta = PngImagePlugin.PngInfo()
         for k, v in metadata.items():
             meta.add_text(k, v, 0)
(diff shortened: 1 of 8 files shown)


sampsyo (approved):
Very very cool—these all seem right to me.

I'll use this space to again lament that Python 3 went in the direction it did for paths; I think the paths-are-`str` approach was worth trying but did not work out in the end, and now it's too late to change.

wisp3rwind:
> I'll use this space to again lament that Python 3 went in the direction it did for paths; I think the paths-are-`str` approach was worth trying but did not work out in the end, and now it's too late to change.

One more place where I think Rust is getting it right — by having an OS-dependent type instead of trying to abstract away too much.

On the other hand, I wonder whether there's really any need to support non-UTF-8 paths these days.

sampsyo:
Yeah, the Rust approach makes a lot of sense to me. But you're right—if we were building beets from scratch today, it would be tempting to simply reject all files with non-UTF-8 names. Crashing seems perhaps too harsh, but (for example) simply logging a message and skipping any nonconforming file could work!

PR #4849, opened 2023-07-16
Always use syspath conversions (#3690 split up, part 3)

wisp3rwind:
Follow-up to https://github.com/beetbox/beets/pull/4830 and #4848:

This is part 3, with some changes to beets' core. Thus, this is probably the most sensitive PR in the series.


sampsyo (approved):
Excellent! I can't see anything wrong here—it is comforting to know that `syspath` is (supposed to be) idempotent, so it's unlikely to break things too bad by calling it unnecessarily. (And anyway, all of these seem to me to qualify as "necessary.")

PR #4850, opened 2023-07-16
tests: make use of our custom filesystem assertions for conciseness

wisp3rwind:
Follow-up to #4830, #4848 and #4849:

this replaces assertions of the form
```py
self.assertTrue(os.path.exists(syspath(path)))
```
by
```py
self.assertExists(path)
```
which includes the syspath conversion and is much easier to read.

Occurences where located using
```sh
git grep -E 'assert(True|False).*(isdir|isfile|exist)'
```

### DIFF
--- a/test/_common.py
+++ b/test/_common.py
@@ -148,6 +148,18 @@ class Assertions:
         self.assertFalse(os.path.exists(syspath(path)),
                          f'file exists: {path!r}')
 
+    def assertIsFile(self, path):  # noqa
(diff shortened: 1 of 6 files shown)


wisp3rwind:
I'll just go ahead on this one and merge; given that it is a fairly straightforward refactoring of tests, there seems little need for a thorough review.

sampsyo:
Yay; very good call!

PR #4851, opened 2023-07-16
replaygain: Assorted refactoring

wisp3rwind:
I'm trying to revive my work on improved parallelization (in particular for replaygain, cf. #3893). A while ago, I've been exploring an alternative approach to the the elaborate async approach of #3893 using beets's pipeline threads + an additional pool of worker threads for CPU intense computation. In this way, no intrusive refactoring of beets' pipeline is required and the resulting code feels easier to understand.

This PR is preparatory work for the new approach, which is however quite independent from the actual rewrite of parallelization; and could be merged already.


beetsplug/replaygain.py
```diff
@@ -1373,12 +1375,13 @@ def _interrupt(self, signal, frame):
             pass
 
     def close_pool(self):
-        """Close the `ThreadPool` instance in `self.pool` (if there is one)
+        """Regularly lose the `ThreadPool` instance in `self.pool`.
```
sampsyo, line 1378:
```suggestion
        """Regularly close the `ThreadPool` instance in `self.pool`.
```

beetsplug/replaygain.py
```diff
@@ -308,42 +307,47 @@ def compute_album_gain(self, task):
...
-            # album peak is maximum track peak
-            album_peak = max(album_peak, track_gain.peak)
+        # Album peak is maximum track peak
+        album_peak = max(tg.peak for tg in track_gains)
 
-            # prepare album_gain calculation
-            # total number of blocks is sum of track blocks
-            n_blocks += track_n_blocks
+        # Total number of BS.1770 gating blocks
+        n_blocks = sum(nb for _tg, nb in track_results)
```
sampsyo, line 328: It's very satisfying that we can now use `max` and `sum` here, as ~God~ Guido intended.

beetsplug/replaygain.py
```diff
@@ -730,13 +730,13 @@ def _import_gst(self):
         self.GLib = GLib
         self.Gst = Gst
 
-    def compute(self, files, target_level, album):
-        self._error = None
-        self._files = list(files)
-
-        if len(self._files) == 0:
+    def compute(self, items, target_level, album):
+        if len(items) == 0:
```
sampsyo, line 734: Could perhaps be `if not items:`?

beetsplug/replaygain.py
```diff
@@ -875,7 +880,7 @@ def _set_file(self):
         # Set a new file on the filesrc element, can only be done in the
         # READY state
         self._src.set_state(self.Gst.State.READY)
-        self._src.set_property("location", py3_path(syspath(self._file.path)))
+        self._src.set_property("location", py3_path(syspath(self._file)))
```
sampsyo, line 883: It's also nice that `_file` now actually contains a file rather than an Item object.

sampsyo (approved):
All looks good from here! I just encountered a couple of extremely tiny suggestions.

beetsplug/replaygain.py
```diff
@@ -308,42 +307,47 @@ def compute_album_gain(self, task):
...
-            # prepare album_gain calculation
-            # total number of blocks is sum of track blocks
-            n_blocks += track_n_blocks
+        # Total number of BS.1770 gating blocks
+        n_blocks = sum(nb for _tg, nb in track_results)
 
+        def sum_of_track_powers(track_gain, track_n_blocks):
             # convert `LU to target_level` -> LUFS
-            track_loudness = target_level_lufs - track_gain.gain
+            loudness = target_level_lufs - track_gain
```
ybnd, line 332: This line seems broken now; after track gains are computed the album gain errors out with
```
TypeError: unsupported operand type(s) for -: 'int' and 'Gain'
```

Looks like it should have been called with `sum_of_track_powers(tg.gain, nb)` below.

ybnd (changes requested):
I don't have much to add to @sampsyo's comments in terms of the refactor itself.

Ran `beet replaygain` with the `command`, `ffmpeg` and `gstreamer` backends (I could not get `audiotools` installed for some reason)

Confirmed that all of the track/album gain calculations and write operations remain identical compared to the `master` branch.

Found a tiny typo with album gain in `ffmpeg`, see below.

beetsplug/replaygain.py
wisp3rwind: Ouch, thanks for testing this thoroughly (which I obviously didn't)!

What happens here is that that our audio test files contain only silence, which the final result that the tests don't actually hit this calculation (more detail: the number of gating blocks with loudness of above the BS1770 gating threshold `nblocks` is zero, even though at ~1s the track is in principle longer that a single gating block of 400ms).

@sampsyo it looks like you added the test files (`full.mp3`, etc). Is there a script around that you used to generate them? Maybe we could replace them with a simple sine wave or so, instead of just silence.

sampsyo: Good question—I just generated those by taking a short silent WAV and using various tools (ffmpeg? maybe even iTunes, back in the day?) to manually convert it to different formats. It's a good idea to do something similar with an off-the-shelf one-second sine wave or similar…

wisp3rwind:
Thanks for the feedback! This should be good to go now: I've added a version of the ffmpeg tests with non-silent files (turns out there's SoX which makes it really easy to generate audio files with some synthesized sounds) and addressed all of your comments.

ybnd (approved):

PR #4854, opened 2023-07-18
Support providing album information on singleton imports via Discogs

### LINKED ISSUE #4716: discogs: Save album information on singleton imports

JOJ0:
## Description

The Discogs metadata plugin now saves album information when importing singletons.

- Fixes #4716
- Is a refactored / simplified version of #4717
- Presents album information in the singleton importer UI as:

```
# selection (default 1), Skip, Use as-is, Enter search, enter Id, aBort,
eDit, edit Candidates? 4
Tagging track: Portishead - Numb
               Album: Numb
URL:
    https://www.discogs.com/release/1385011-Portishead-Numb
(Similarity: 100.0%) (Discogs, Index 1, Track A1)
Apply, More candidates, Skip, Use as-is, Enter search, enter Id, aBort,
eDit, edit Candidates? 
```

- Additionally this PR extends the "disambiguation string" for _all_ singleton imports with track index and alternative name (track_alt field):

```
Candidates:
1. Portishead - Numb (100.0%) (Spotify, Index 7)
2. Portishead - Numb (100.0%) (Spotify, Index 1)
3. Portishead - Numb (100.0%) (Spotify, Index 27)
4. Portishead - Numb (100.0%) (Discogs, Index 1, Track A1)
5. Portishead - Numb (100.0%) (Discogs, Index 1, Track A1)
6. Portishead - Numb (100.0%) (Discogs, Index 1, Track 1)
7. Portishead - Numb (100.0%) (Discogs, Index 1, Track 1)
8. Portishead - Numb (100.0%) (Discogs, Index 1, Track 1)
```
- Furthermore if an album name is found, it's appended to the end of the disambiguation string.  This is currently supported by the Discogs and the Spotify plugin:

```
Candidates:
1. Portishead - Numb (100.0%) (Discogs, Index 1, Track A1, [Numb])
2. Portishead - Numb (100.0%) (Discogs, Index 1, Track A1, [Numb])
3. Portishead - Numb (100.0%) (Discogs, Index 1, Track 1, [Numb])
4. Portishead - Numb (100.0%) (Discogs, Index 1, Track 1, [Numb])
5. Portishead - Numb (100.0%) (Discogs, Index 1, Track 1, [Numb])
6. Portishead - Numb (100.0%) (Spotify, Index 7, [Dummy])
7. Portishead - Numb (100.0%) (Spotify, Index 1, [Numb])
8. Portishead - Numb (100.0%) (Spotify, Index 27, [Crazy Mix 5])
9. Portishead - Numbed In Moscow (64.3%) (title) (Spotify, Index 2, [Numb])
```
- This feature is enabled by default but if such a long disambiguation string is not desired by the user, displaying of `[Album]` can be turned off using a new configuration option:
```
import:
    singleton_album_disambig: no
```

## To Do

- [x] Documentation. 
- [x] Changelog.
- [x] ~Tests.~

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -220,6 +220,9 @@ def disambig_string(info):
             disambig.append("Index {}".format(str(info.index)))
         if info.track_alt:
             disambig.append("Track {}".format(info.track_alt))
+        if (config['import']['singleton_album_disambig'].get()
+                and info.get('album')):
+            disambig.append("[{}]".format(info.album))
 
     if disambig:
         return ', '.join(disambig)
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -450,18 +450,25 @@ def show_item_change(item, match):
     """
     cur_artist, new_artist = item.artist, match.info.artist
     cur_title, new_title = item.title, match.info.title
+    cur_album = item.album if item.album else ""
+    new_album = match.info.album if match.info.album else ""
 
-    if cur_artist != new_artist or cur_title != new_title:
+    if (cur_artist != new_artist or cur_title != new_title
+            or cur_album != new_album):
         cur_artist, new_artist = ui.colordiff(cur_artist, new_artist)
         cur_title, new_title = ui.colordiff(cur_title, new_title)
+        cur_album, new_album = ui.colordiff(cur_album, new_album)
 
         print_("Correcting track tags from:")
         print_(f"    {cur_artist} - {cur_title}")
+        print_(f"    Album: {cur_album}") if cur_album else None
         print_("To:")
         print_(f"    {new_artist} - {new_title}")
+        print_(f"    Album: {new_album}") if new_album else None
 
     else:
         print_(f"Tagging track: {cur_artist} - {cur_title}")
+        print_(f"               Album: {new_album}") if cur_album else None
 
     # Data URL.
     if match.info.data_url:
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -461,14 +461,17 @@ def show_item_change(item, match):
 
         print_("Correcting track tags from:")
         print_(f"    {cur_artist} - {cur_title}")
-        print_(f"    Album: {cur_album}") if cur_album else None
+        if cur_album:
+            print_(f"    Album: {cur_album}")
         print_("To:")
         print_(f"    {new_artist} - {new_title}")
-        print_(f"    Album: {new_album}") if new_album else None
+        if new_album:
+            print_(f"    Album: {new_album}")
 
     else:
         print_(f"Tagging track: {cur_artist} - {cur_title}")
-        print_(f"               Album: {new_album}") if cur_album else None
+        if cur_album:
+            print_(f"               Album: {new_album}")
 
     # Data URL.
     if match.info.data_url:
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -215,6 +215,12 @@ def disambig_string(info):
         if info.albumstatus == 'Pseudo-Release':
             disambig.append(info.albumstatus)
 
+    if isinstance(info, hooks.TrackInfo):
+        if info.index:
+            disambig.append("Index {}".format(str(info.index)))
+        if info.track_alt:
+            disambig.append("Track {}".format(info.track_alt))
+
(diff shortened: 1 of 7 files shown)


beets/ui/commands.py
```diff
@@ -444,18 +450,25 @@ def show_item_change(item, match):
...
-    if cur_artist != new_artist or cur_title != new_title:
+    if (cur_artist != new_artist or cur_title != new_title
+            or cur_album != new_album):
         cur_artist, new_artist = ui.colordiff(cur_artist, new_artist)
         cur_title, new_title = ui.colordiff(cur_title, new_title)
+        cur_album, new_album = ui.colordiff(cur_album, new_album)
 
         print_("Correcting track tags from:")
         print_(f"    {cur_artist} - {cur_title}")
+        print_(f"    Album: {cur_album}") if cur_album else None
```
wisp3rwind, line 464:
```suggestion
        if cur_album:
            print_(f"    Album: {cur_album}")
```

I don't think I've ever seen this idiom in Python code and it took me a second to parse properly — let's avoid this.

wisp3rwind:
> Without having investigated too much, it seemed to me that the TrackInfo objects generated during this early stage in the import process are passed all the way through to the database abstraction layer. where it finally wants to store this unsupported field. Removing it on this level again felt unnecessary complicated so I decided to go down an even simpler route of just applying, what I actually need, to the TrackInfo object: The `album` name. This is actually the way the Spotify plugin does it too.

Yes, looks like it's trying to assign the `AlbumInfo` to an item field, which makes sense since you just set it as any other `TrackInfo` field.

> What do you think about this solution?

Looks alright to me, we can still come up with a more elaborate solution should we ever need more album fields on `TrackInfo`.

beets/ui/commands.py
JOJ0: Thanks! Well, trying to stay brief/one-line is not _always_ the best idea. I'll also change the other occurence of it.

JOJ0:
> > What do you think about this solution?
> 
> Looks alright to me, we can still come up with a more elaborate solution should we ever need more album fields on `TrackInfo`.

Perfect!

Thanks so much for the review!

JOJ0:
Alright, I think I'm happy with this feature now. In the end I decided to additionally add the album name to the disambiguation string, which makes the selection of the "correct" release even more straight forward. Also I realized that it makes sense to make this globally configurable since the Spotify plugin will benefit from it as well and some users might find such a long disambiguation string distracting. See my additions to the inital PR description.

I tried to explain as brief and clear as possible but nitpicking on my docs wording is always very welcome :-)

https://github.com/beetbox/beets/pull/4854/commits/a3749779534d579b0f032db83ff530724f6b52d1#diff-6d7bb6bc84b63dd7d8b09e64b4d5b117575f689ecb5a492c25b9aeeff1b5b38fR765-R774

wisp3rwind (approved):
LGTM; however, I've only looked at the code changes in isolation, since I don't know the discogs plugin (in fact, any of the metadata sources) in detail.

JOJ0:
Thanks @wisp3rwind, that's perfectly fine and helpful. Thanks for chrecking through! I'm happy to finally have merged this feature. I had been successfully using its first version for quite some time now, which shares parts of the Discogs plugin code we have here and I'm also very happy that now it's so much simpler than it was in the first place. (thanks again @sampsyo for the hints to simplify ;-)))

PR #4857, opened 2023-07-21
Fix link format in README

Decad:
Fix issue with link in readme.


wisp3rwind:
Looks like there was some confusion between Mardown and rst; thanks for the fix!

PR #4858, opened 2023-07-22
typing: Wrap up dbcore

wisp3rwind:
Finally, mypy is happy about dbcore :tada:

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -16,6 +16,7 @@
 """
 
 from __future__ import annotations
+from abc import ABC
 import time
 import os
 import re
@@ -25,22 +26,36 @@ import sqlite3
 import contextlib
 from sqlite3 import Connection
 from types import TracebackType
-from typing import Iterable, Type, List, Tuple, Optional, Union, \
-    Dict, Any, Generator, Iterator, Callable
+from typing import (
+    Any,
+    Callable,
+    cast,
+    DefaultDict,
+    Dict,
+    Generator,
+    Generic,
+    Iterable,
+    Iterator,
+    List,
+    Mapping,
+    Optional,
+    Sequence,
+    Set,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+)
 
 from unidecode import unidecode
 
 import beets
 from beets.util import functemplate
 from beets.util import py3_path
-from beets.dbcore import types
+from . import types
 from .query import MatchQuery, NullSort, TrueQuery, AndQuery, Query, \
-    FieldQuery, Sort
-from collections.abc import Mapping
+    FieldQuery, Sort, FieldSort
 
-from typing import TYPE_CHECKING
-if TYPE_CHECKING:
-    from beets.library import LibModel
 from ..util.functemplate import Template
 
 
@@ -53,7 +68,7 @@ class DBAccessError(Exception):
     """
 
 
-class FormattedMapping(Mapping):
+class FormattedMapping(Mapping[str, str]):
     """A `dict`-like formatted view of a model.
 
     The accessor `mapping[key]` returns the formatted version of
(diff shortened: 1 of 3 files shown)


beets/dbcore/db.py
```diff
@@ -201,21 +234,30 @@ def get(self, key, default=None):
...
+        # NOTE: It would be nice to use the following:
+        # yield from self._converted
+        # yield from self._data
+        # but that won't work since some code relies on modifying `self`
+        # during iteration.
         return iter(self.keys())
 
+    def __len__(self) -> int:
+        # FIXME: This is incorrect due to duplication of keys
+        return len(self._converted) + len(self._data)
```
sampsyo, line 255: Would it work to do `len(self.keys())`, even if that's obviously inefficient? I suppose this probably doesn't matter much!

beets/dbcore/db.py
```diff
@@ -359,6 +401,8 @@ def _check_db(self, need_id: bool = True):
         if need_id and not self.id:
             raise ValueError('{} has no id'.format(type(self).__name__))
 
+        return self._db
```
sampsyo, line 404: This is a clever way to get rid of some dependence on statefulness!

sampsyo (approved):
🎉🎉🎉

This all looks great!! Amazing work seeing this through.

beets/dbcore/db.py
```diff
@@ -83,31 +98,39 @@ def __init__(
...
+    def _get_formatted(self, model: Model, key: str) -> str:
         value = model._type(key).format(model.get(key))
         if isinstance(value, bytes):
             value = value.decode('utf-8', 'ignore')
 
         if self.for_path:
-            sep_repl = beets.config['path_sep_replace'].as_str()
-            sep_drive = beets.config['drive_sep_replace'].as_str()
+            sep_repl = cast(str, beets.config['path_sep_replace'].as_str())
+            sep_drive = cast(str, beets.config['drive_sep_replace'].as_str())
```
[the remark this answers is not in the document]
wisp3rwind: Yep, just needs a volunteer ;)

beets/dbcore/db.py
```diff
@@ -119,67 +142,77 @@ def _get_formatted(self, model, key):
...
 
+# NOTE: This seems like it should be a `Mapping`, i.e.
+# ```
+# class LazyConvertDict(Mapping[str, Any])
+# ```
+# but there are some conflicts with the `Mapping` protocol such that we
+# can't do this without changing behaviour: In particular, iterators returned
+# by some methods build intermediate lists, such that modification of the
+# `LazyConvertDict` becomes safe during iteration. Some code does in fact rely
+# on this.
```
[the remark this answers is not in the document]
wisp3rwind: Just changing `LazyConvertDict` should be easy enough. In fact, I did that, but that resulted in test failures due to mutation during iteration. At that point, I just reverted and didn't check what the offending code actually did. Maybe it's not so bad.

beets/dbcore/db.py
wisp3rwind: Actually no, because one of the peculiarities of `LazyConvertDict` is that `keys` gives a list, which as far as I understand, should have the same duplicates.

I have some code locally that changes the implementation (basically, `pop` entries from `_data` when adding to `_converted`) to avoid this.

But maybe I'm just misunderstanding something here: It is somewhat surprising that no issues were caused by such duplicates (but maybe, we just don't have much code that iterates `LazyConvertDict`?).

sampsyo: Ah, got it! So this essentially reflects a regrettable quirk, but it reflects that quirk accurately.

Yes, I agree with you that this hasn't caused problems historically. I agree that there probably are not too many places where we iterate over the mapping, and where we do, the situation could easily be idempotent (e.g., when inserting into another map, which will deduplicate things). 🤷

PR #4860, opened 2023-07-24
Update faq.rst for linuxserver.io

RollingStar:
## Description

Second fix for #4838.  

(...)

## To Do


RollingStar:
There are other ways this could happen - not just with Linuxserver. The sin isn't just moving, it's moving and writing tags at the same time. No undo.

sampsyo:
I suppose that's right. But for this purpose, it seems like the main thing we want to do is catch the eye of LinuxServer container users… I'm not sure if other people (who have presumably configured things themselves?) would be able to identify their problem given a short description.

snejus:
Closing this for the same reason as #4859

PR #4861, opened 2023-07-25
Log smart playlist create msg as info log

JOJ0:
Log the creation of a smart playlist as info instead of debug log level. 

## Description

In an explicit run (not while importing) it doesn't hurt to get some progress to the UI since creating/updating smart playlists can take a singinficant amount of time.

## To Do

- [x] ~Documentation.~
- [x] Changelog.
- [x] ~Tests.~


sampsyo:
Good call! And you're right; this won't disrupt automatic (`beet import`) runs.

PR #4862, opened 2023-07-25
Trigger subsonicupdate on database and smart playlist changes

JOJ0:
## Description

- Fixes triggering subsonicupdate on a database_change event instead of only at an import event.
- Sends a new event from the smartplaylist plugin whenever lists are updated/created and has subsonicupdate plugin listen to it.
- Both events register new listeners that launch the actual Subsonic library update _at the very end of a beets run_ (cli_exit) instead of being triggered by each change (similar to how the mpdupdate plugin does it).
- It doesn't hurt if Subsonic is informed on subsequent changes of smart playlists (while the user is heavily experimenting with them) since once a subsonic-rescan is triggered, it keeps running and doesn't restart from scratch.

## To Do

- [x] Documentation.
- [x] Changelog.
- [x] ~Tests.~

### DIFF
--- a/beetsplug/smartplaylist.py
+++ b/beetsplug/smartplaylist.py
@@ -17,6 +17,7 @@
 
 
 from beets.plugins import BeetsPlugin
+from beets.plugins import send as send_event
 from beets import ui
 from beets.util import (mkdirall, normpath, sanitize_path, syspath,
                         bytestring_path, path_as_posix, displayable_path)
@@ -232,6 +233,8 @@ class SmartPlaylistPlugin(BeetsPlugin):
                         if self.config['urlencode']:
                             path = bytestring_path(pathname2url(path))
                         f.write(prefix + path + b'\n')
+            # Send an event when playlists were updated.
+            send_event("smartplaylist_update")
 
         if pretend:
             self._log.info("Displayed results for {0} playlists",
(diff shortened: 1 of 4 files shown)


sampsyo (approved):
Interesting; this seems pretty good! One odd thing to note is that I don't think we have ever emitted an even _from a plugin_ before. (Grepping for `send_event` in the beetsplug directory yields only false-positives in the `bpd` plugin.) I think it's fine for this to happen, but I just thought it was worth remarking on the fact that this is the first instance of plugin-to-plugin communication.

JOJ0:
Oh wow! That is indeed interesting. I wouldn't have thought of that this was never used before. As usual, thanks a lot for the insight!

One thing I was wondering while putting this PR together was: "Should I include the same feature for mpd/bpd plugins?" I didn't in the end because I never felt that for my use-cases mpd requires to be informed about playlist changes. But I'm not a heavy mpd user, and have a rather unusual setup.

tl;dr; I would be interested what you think about it. Would it make sense to receive this event in the `mpdupdate` plugin?

sampsyo:
I guess I think it's OK to just add this one feature! No particular need to preemptively cover other plugins… if a user of bpd requests this behavior there, we can consider it then.

PR #4866, opened 2023-07-31
Optional more verbose duplicates prompt

JOJ0:
## Description

A new config option allows to enable a more detailed view when a duplicate is detected during import of an album or singleton.

- New `import` config option:

```
import:
    duplicate_verbose_prompt: no
```
- Disabled by default.

## To Do

- [x] Documentation.
- [x] Changelog. 
- [x] ~Tests.~

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -873,8 +873,11 @@ class TerminalImportSession(importer.ImportSession):
                     not task.is_album,
                 ))
                 if config['import']['duplicate_verbose_prompt']:
-                    for dup in duplicate.items():
-                        print(f"  {dup}")
+                    if task.is_album:
+                        for dup in duplicate.items():
+                            print(f"  {dup}")
+                    else:
+                        print(f"  {duplicate}")
 
             print_("New: " + summarize_items(
                 task.imported_items(),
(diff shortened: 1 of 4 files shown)


beets/ui/commands.py
```diff
@@ -872,11 +872,20 @@ def resolve_duplicate(self, task, found_duplicates):
                     list(duplicate.items()) if task.is_album else [duplicate],
                     not task.is_album,
                 ))
+                if config['import']['duplicate_verbose_prompt']:
+                    if not task.is_album:
+                        print(f"  {duplicate}")
+                    else:
+                        for dup in duplicate.items():
+                            print(f"  {dup}")
```
sampsyo, line 880: This might be slightly easier to read if it were reversed! (That is, start with `if task.is_album:`.)

JOJ0: Hahaha, I had it the other way round and thought: That looks badly readable and switched it around. But actually I agree with you, starting with a negated condition when there is an else branch anyway, is often confusing. I'll change it back! :-)

arsaboo:
Thanks @JOJ0 

It would be nice to get [this related PR ](https://github.com/beetbox/beets/pull/4471)to handle upgrades merged soon.

JOJ0:
Alright, I'm quite happy with this feature now and it's ready for a final review. Some nitpicking on my docs-wording and format might be good! Thanks in advance!

sampsyo (approved):
Nice! No objections here—looks pretty solid to me.

PR #4868, opened 2023-08-05
Fix compatibility with pillow 10

mweinelt:
The `ANTIALIAS` attribute has long been deprecated, and was finally removed in pillow 10.0.

The recommendation is to migrate to `Resampling.LANCZOS` instead.

https://pillow.readthedocs.io/en/stable/deprecations.html#constants

## Description

Fixes #X.  

(...)

## To Do

### DIFF
--- a/beets/util/artresizer.py
+++ b/beets/util/artresizer.py
@@ -353,7 +353,7 @@ class PILBackend(LocalBackend):
         try:
             im = Image.open(syspath(path_in))
             size = maxwidth, maxwidth
-            im.thumbnail(size, Image.ANTIALIAS)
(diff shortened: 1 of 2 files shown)


wisp3rwind:
Thanks for catching this!

Would you mind adding a changelog entry so that people who run into this issue can find the solution by searching through the log?

(For reference: The `LANCZOS` constant [was added in Pillow 1.1.3](https://pillow.readthedocs.io/en/stable/handbook/concepts.html#concept-filters) which was released in 2010, so I think we should be fine without explicitly stating any version requirement in `setup.py`).

mweinelt:
Added to the changelog. PTAL.

wisp3rwind:
Thanks again!

PR #4878, opened 2023-08-16
Tutorial documentation

bodnguye:
## Description

Recreated a Tutorial through series of steps to complete a task for a project for newcomers to be able find quickly.

## To Do


sampsyo:
Hi there! Thanks for your interest in contributing. This tutorial is pretty similar to our [Getting Started guide](https://beets.readthedocs.io/en/stable/guides/main.html). We don't want to duplicate the text in two places—that would create a tricky maintenance burden where we'd need to modify two files whenever we need to make a change. Can you explain a little more about what you wanted to address here so we can think about how to improve our existing documentation?

JOJ0:
Basically I always welcome new docs efforts, but as @sampsyo mentioned, I also wondered why this tutorial seems to be very much the same as the original getting started guide with the difference being that in some parts it is much shorter chapters. So yeah, let me second that question: What is your goal with this doc? How can we make sure we don't double information and soon get outdated docs because of the maintenance-burden Adrian already mentioned.

One idea would be that you try to update the getting started guide with the first goal being: kicking out everything that _is not_ true anymore / is outdated.

PR #4879, opened 2023-08-16
Getting Started Guide

jontlo:
## Description

Adds documentation to update pip to the current version in "Installing" section.
Edits documentation for how to navigate to Environment Variables in Windows 11 in "Installing on Windows" section.
Adds documentation for how to run beets package if scripts are not in PATH in "Installing on Windows" section.

## To Do

- [ x ] Documentation. (If you've add a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)
- [ x ] Changelog. (Add an entry to `docs/changelog.rst` near the top of the document.)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -237,6 +237,8 @@ Other changes:
   :bug:`4585`
 * :doc:`/faq`: :ref:`multidisc`: Elaborated the multi-disc FAQ :bug:`4806`
 * :doc:`/faq`: :ref:`src`: Removed some long lines.
+* :doc:`/guides`: :ref:`main.rst`: Added documentation for updating pip,
+  adding to PATH on Windows, and running the beets package.
 
 1.6.0 (November 27, 2021)
 -------------------------
--- a/docs/guides/main.rst
+++ b/docs/guides/main.rst
@@ -58,7 +58,7 @@ Beets works on Python 3.7 or later.
 .. _MacPorts: https://www.macports.org
 
 If you have `pip`_, just say ``pip install beets`` (or ``pip install --user
-beets`` if you run into permissions problems).
+beets`` if you run into permissions problems). If you have pip but wish to update (uninstall and install the current version), run ``python -m pip install –upgrade pip``. Ensure the directory with pip’s scripts are added to ``PATH``.
 
 To install without pip, download beets from `its PyPI page`_ and run ``python
 setup.py install`` in the directory therein.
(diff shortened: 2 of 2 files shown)


docs/changelog.rst
```diff
@@ -237,6 +237,8 @@ Other changes:
   :bug:`4585`
 * :doc:`/faq`: :ref:`multidisc`: Elaborated the multi-disc FAQ :bug:`4806`
 * :doc:`/faq`: :ref:`src`: Removed some long lines.
+* :doc:`/guides`: :ref:`main.rst`: Added documentation for updating pip,
```
JOJ0, line 240: Should be main instead of main.rst but on the other hand: We don't really need a changelog entry for these small docs updates. Just remove the line and we are good :-)

docs/guides/main.rst
```diff
@@ -60,6 +60,10 @@ Beets works on Python 3.7 or later.
 If you have `pip`_, just say ``pip install beets`` (or ``pip install --user
 beets`` if you run into permissions problems).
 
+If you have pip but wish to update (uninstall and install the current version),
+run ``python -m pip install –upgrade pip``. Ensure the directory with pip’s scripts
+are added to ``PATH``.
+
```
JOJ0, line 66: Not sure if we should add this. Handling pip upgrade is kind of out of scope for the beets project. I'd like to keep this already very long guide as short as possible. An alternative idea would be to link to the official pip docs that explain upgrading.

docs/guides/main.rst
```diff
@@ -99,19 +103,18 @@ get it right:
...
-   then hit the "Environmental Variables..." button, and then look for the PATH 
-   variable in the table. Add the following to the end of the variable's value: 
-   ``;C:\Python37;C:\Python37\Scripts``. You may need to adjust these paths to 
-   point to your Python installation.
+   include Python and its scripts. To do so, press Win + R to open the Run dialog
+   box. Next, type ``sysdm.cpl`` to open "System Properties". After, under the 
+   “Advanced” tab, click the “Environment Variables” button, and look for the
+   ``PATH`` variable in the table. Add the following to the end of the variable's
+   value: ``;C:\Python37;C:\Python37\Scripts``. You may need to adjust these paths
+   to point to your Python installation.
```
JOJ0, line 111: I like this change. Not much of an Windows expert but I think using sysdm.cpl tool directly is the more sustainable method! Nice!

JOJ0 (changes requested):
Sorry for the superlong delay with reviewing. Hope you're still up for these tiny updates and I'll merge soon :-)

docs/guides/main.rst
sampsyo: We in fact have an FAQ about `$PATH`, which could be worth linking to:
https://beets.readthedocs.io/en/stable/faq.html#say-command-not-found

JOJ0:
@jontlo what do you think of our suggestions? Questions? Don't like them? Thanks for any response :-)

JOJ0:
@jontlo? only tiny fixes away of a merge!

JOJ0:
@jontlo should we close this? Or do you want to do those tiny changes and we can finally merge it! Would certainly appreciate the latter! :-)

JOJ0:
I'll close this now. Some information about upgrading pip which is actually off-topic. The thing about the PATH variable on Windows might be of interest to some maybe, but actually also off-topic. It's Windows system administration know how and all not really beets-specific. Will be outdated info anyway at some point....

PR #4883, opened 2023-08-26
Add embyupdate `userid` config option

### LINKED ISSUE #4402: embyupdate: Allow direct configuration of user ID for private users
In the embyupdate.py plugin, if the admin user listed in the beets config is hidden from login screens, the embyupdate plugin will fail and not find the username. 

I am using an API key, which is server wide so the username does not really matter. Therefore the get_user function should just return the user provided in the beets config

We should also allow for the emby user ID (string/hash) to be supplied directly from the beets config to avoid this situation. 

can this change be made please?

shoriminimoe:
## Description

Fixes #4402.  

Adds a `userid` config option for the `embyupdate` plugin. When the
`userid` is provided, the Emby user lookup is bypassed. This avoids a
failure to get the user ID for private users.

## To Do

### DIFF
--- a/beetsplug/embyupdate.py
+++ b/beetsplug/embyupdate.py
@@ -167,6 +167,7 @@ class EmbyUpdate(BeetsPlugin):
         port = config['emby']['port'].get()
         username = config['emby']['username'].get()
         password = config['emby']['password'].get()
+        userid = config['emby']['userid'].get()
         token = config['emby']['apikey'].get()
 
         # Check if at least a apikey or password is given.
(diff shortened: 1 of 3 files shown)


shoriminimoe:
One of the original requests from #4402 was to change what is returned by `get_user`. I chose not to change that because that functions returns the JSON object for the user from the Emby api and not just a username and ID. Instead, the user lookup is skipped entirely if a `userid` is supplied in the config.

sampsyo:
Nice; this looks perfect! Many thanks!!

PR #4890, opened 2023-09-01
Correcting pip specific version install from "=" to "==" in Contributing.rst

samtaylorr:
## Description

Changed line 291 from  ``python -m pip install tox=3.8.3`` to  ``python -m pip install tox==3.8.3`` as shown in the 7th example: https://pip.pypa.io/en/stable/cli/pip_install/#examples, this is the correct way to install specific versions.


sampsyo:
Good catch; thank you!

PR #4899, opened 2023-09-09
Support excluding fields with update command

fracai:
## Description

Adds `-e`, `–exclude-field` for selecting fields to _not_ update.

## To Do

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1196,10 +1196,12 @@ default_commands.append(list_cmd)
 
 # update: Update library contents according to on-disk tags.
 
-def update_items(lib, query, album, move, pretend, fields):
+def update_items(lib, query, album, move, pretend, fields, exclude_fields):
     """For all the items matched by the query, update the library to
     reflect the item's embedded tags.
     :param fields: The fields to be stored. If not specified, all fields will
+    be.
+    :param exclude_fields: The fields to not be stored. If not specified, all fields will
     be.
     """
     with lib.transaction():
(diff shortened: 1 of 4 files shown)


beets/ui/commands.py
```diff
@@ -1210,6 +1213,12 @@ def update_items(lib, query, album, move, pretend, fields):
             fields.append('path')
         items, _ = _do_query(lib, query, album)
 
+        item_fields = fields or library.Item._media_fields
+        album_fields = fields or library.Album._fields.keys()
```
sampsyo, line 1217: There's nothing specific to do here, but I just wanted to highlight that this looks somewhat odd when put next to each other (`Item` uses `_media_fields`, `Album` uses all `_fields`). There's a good reason for this (the former needs to pick only the relevant fields that might come from files); it just looks somewhat odd.

sampsyo (approved):
Nice! This looks very straightforward; thanks for the clean implementation and the clear documentation. I think this is ready to go whenever the CI is green.

sampsyo:
Hmm; it looks like a few `update`-related tests are failing in CI, which may mean that my judgment in code review was wrong and the behavior may have changed for the default case (not excluding any fields). Maybe this deserves a slightly deeper investigation?

fracai:
> Hmm; it looks like a few `update`-related tests are failing in CI, which may mean that my judgment in code review was wrong and the behavior may have changed for the default case (not excluding any fields). Maybe this deserves a slightly deeper investigation?

I think I actually saw these when I tested locally, but ignored them because they related to "missing files". I'll take a look at what the behavior change could be.

beets/ui/commands.py
fracai: Setting `item_fields` to `_fields.keys()` does fix the the failing tests, but I'm investigating other solutions in case there's a tighter fix.

fracai: yup, found a smaller set of fields to update. It's just `_media_fields` and conditionally `path`

sampsyo (approved):
Nice; looks good to me! Thank you!

PR #4909, opened 2023-09-16
Make `mb_albumartistids` available at the album level.

djl:
As this is a field related to albums, it probably should be accessible as `album.mb_albumartistids`.


JOJ0:
@jmbannon please a quick review. From my wild guess this could be safely merged, agree?

jmbannon:
I think it needs to be added to `item_keys` below as well. Wouldn't hurt to have a test to be sure it works

djl:
Added the field to `item_keys` and added a test to ensure `album.mb_albumartistids` matches the album's first item's `mb_albumartistids`. There's probably a more sensible or robust way to test it, though.

jmbannon:
In the assert, could you add the hard coded list containing the expected strings. That should tell us for sure it works

djl:
So at the point where I added the test, it seems not all the metadata has been applied. `item.mb_albumartistids` is an empty list.

It's been like two years since I last looked at beets core. Could someone point in the right direction?

JOJ0:
keep open

snejus:
Have only noticed this now. I've actually had this fixed in my fork for a while now, and haven't realized this wasn't available upstream! It should go in. @djl are you by any chance still around?

djl:
I had completely forgotten about this! @snejus if you have a more up to date implementation then please open a PR. I won't have time to look into this for a while.

snejus:
Thanks for your contribution @djl! I'll rebase your commit and take it from here

jmbannon:
Lgtm cc @JOJ0

JOJ0 (approved):

PR #4919, opened 2023-09-22
Try adding a ReadTheDocs config

### LINKED ISSUE #4912: readthedocs config deprecation
https://blog.readthedocs.com/migrate-configuration-v2/

We should take care of it really soon, due date Sep, 25th!

sampsyo:
Fixes #4912. I'm not sure if there's a way to test this before merging into the main branch, but I'm opening this PR just to show what's happening and then immediately merging…

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,16 @@
+version: 2
+
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
+sphinx:
+  configuration: doc/conf.py
+
+python:
+  install:
+    - method: pip
+      path: .
+      extra_requirements:
+        - docs


.readthedocs.yaml
```diff
@@ -0,0 +1,16 @@
+version: 2
+
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
+sphinx:
+  configuration: doc/conf.py
```
JOJ0, line 9: Isnt this the wrong path? In beets we have docs/ 

sorry was just an as-is copy from other project!

sampsyo: Yep, absolutely! No problem; it broke the build and I fixed it in the subsequent commit: ad73a61ec51c614c9d66e02202a532fc0caabb6d

Issue #2895, opened 2018-05-02
Log a message when changing filenames to avoid conflicts


alch-emi:
### Summary ###
When moving an item with the `move` command, if a file that's not registered with beets exists in the destination directory, beets gives the file a unique filename, but neglects to tell the user that the move didn't go as planned (even if in verbose mode), and continues to claim that the file was moved if the command is run again, even if the file isn't moved.  If the command is run with the `-p` flag in order to cause a dry run, beets claims that it is moving the file to the original file name, not the unique-ified name.

### Setup ###
* OS: Linux 4.16.5-1-ARCH
* Python version: Python 3.6.5
* beets version: 1.4.6 [no plugins; default options]

### Reproduction ###
1. With an empty library, import an item. 
2. Change the library directory in the config to point to a new, empty directory.
3. Create a dummy file in the new directory mirroring the actual music file.  For example, if the real file is in `library/DummyArtist/DummyAlbum/DummySong.mp3`, and your dummy library is `newlibrary`, create the file in `newlibrary/DummyArtist/DummyAlbum/DummySong.mp3`.  The content is irrelevant.  It can be identical, empty, or corrupt.
4. Issue `beet -vv move -p`.  Notice that beets declares that it is moving the song to the same location as the dummy file, and doesn't mention any problem with this
5. Issue `beet -vv move`.  Beets moves the item to `newlibrary/DummyArtist/DummyAlbum/DummySong.1.mp3`, and doesn't mention it, even with verbose mode enabled.
5. Issue `beet -vv move -p`  and `beet -vv move`  again.  In both cases, beets claims that it is moving the file, despite not moving the file at all.

### Expected / Suggested Behavior ###
I understand why beets is doing what it is doing, but I believe that leaving the user in the dark about it can only cause confusion.  I would suggest that beets do one of the following instead:

* On a move conflict, provide a dialog to the user, allowing them to choose, to skip the file, overwrite it, or use a unique name.  This would essentially force the user to make a decision and enable them to deal with it easily, from withing the beets interface.  It might get a little annoying if the user wants the behavior in the third bullet, or without options to assume an answer (e.g. overwrite all)
* On a move conflict, print a warning to console specifying where the error occurred.  This is the closest to what beets currently does, and seems like a simple solution that gives the most control to the user without being too silent or interrupting an otherwise one-click action.  If the user wants to deal with the problem themselves, however, they need to manually remove the conflicting item, which might be annoying if there are hundreds of conflicts
[...]

PR #4924, opened 2023-09-24
Add log message on verbose, when beets renames destination file during move.

fleur:
## Description

First step in fix of #2895.  Add log message on verbose, when beets renames destination file during move.  The issue says it may be trickier than it looks.  It did not seem tricky, so only doing the first step suggested, to see if I missed something.

## To Do

- [x] Documentation is not needed since functional behavior is not changing.
- [x] Changelog
- [x] Tests are not needed because there are already tests in test/test_files.py that exercise the code path that pass.

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -856,7 +856,13 @@ class Item(LibModel):
         `operation` should be an instance of `util.MoveOperation`.
         """
         if not util.samefile(self.path, dest):
+            old_dest = dest
             dest = util.unique_path(dest)
+            log.debug(
+                'File already exists at dest, changing final path slightly, '
+                "to be unique, from '{}', to '{}'.",
+                old_dest, dest
+            )
         if operation == MoveOperation.MOVE:
             plugins.send("before_item_moved", item=self, source=self.path,
                          destination=dest)
--- a/beets/library.py
+++ b/beets/library.py
@@ -859,9 +859,8 @@ class Item(LibModel):
             old_dest = dest
             dest = util.unique_path(dest)
             log.debug(
-                'File already exists at dest, changing final path slightly, '
-                "to be unique, from '{}', to '{}'.",
-                old_dest, dest
+                'File exists at {}. Avoiding conflict with new name: {}',
+                util.displayable_path(old_dest), util.displayable_path(dest)
             )
         if operation == MoveOperation.MOVE:
             plugins.send("before_item_moved", item=self, source=self.path,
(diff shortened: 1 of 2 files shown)


beets/library.py
```diff
@@ -856,7 +856,13 @@ def move_file(self, dest, operation=MoveOperation.MOVE):
         `operation` should be an instance of `util.MoveOperation`.
         """
         if not util.samefile(self.path, dest):
+            old_dest = dest
             dest = util.unique_path(dest)
+            log.debug(
+                'File already exists at dest, changing final path slightly, '
+                "to be unique, from '{}', to '{}'.",
```
sampsyo, line 863: Just a suggestion to slightly simplify the wording here… how about this?

> File exists at `{}`. Avoiding conflict with new name: `{}`

beets/library.py
```diff
@@ -856,7 +856,13 @@ def move_file(self, dest, operation=MoveOperation.MOVE):
         `operation` should be an instance of `util.MoveOperation`.
         """
         if not util.samefile(self.path, dest):
+            old_dest = dest
             dest = util.unique_path(dest)
+            log.debug(
```
sampsyo, line 861: I think you may need a `if dest != old_dest` here. This may also need some careful testing to make sure it doesn't fire spuriously when these point to the same path but are just normalized differently… hopefully that doesn't happen, but a few test runs should reveal the answer!

sampsyo (commented):
This seems like a nice idea! Here are some suggestions.

beets/library.py
```diff
@@ -856,7 +856,13 @@ def move_file(self, dest, operation=MoveOperation.MOVE):
...
+                'File already exists at dest, changing final path slightly, '
+                "to be unique, from '{}', to '{}'.",
+                old_dest, dest
```
sampsyo, line 864: These will need to use `displayable_path` wrappers.

beets/library.py
fleur: If you have any specific suggestions of test runs, and have the time to, please let me know.  All tests pass, so I'm assuming you mean something outside of those.

sampsyo: About being "normalized differently," I'm just thinking about all the various ways in which different paths can refer to the same file. For example, `os.path.normpath` or `os.path.abspath` are different ways of transforming paths so they still refer to the same file. It's just worth double-checking to make sure that we're not doing that kind of thing when transforming the path (or that the original path is already normalized in the same way, which it probably is).

I suggested `if dest != old_dest` so that (normalization aside) this message only gets printed when there is an actual change.

For testing, all I mean is running this on a few files—perhaps in the same cases that presumably motivated you to add this log message—and make sure that the message shows up when it's supposed to (and doesn't appear when it isn't).

JOJ0:
Is there something to do here still @fleur? Was the `old != new` suggestion sorted out? Or new ideas? I think this is a candidate for merging soon :-) HTH

PR #4940, opened 2023-10-07
FetchArt : fix an error with CAA where the pre-sized thumbs of release groups would be ignored

AdrienCos:
## Description

This patch fixes two bugs with the Cover Art Archive provider of the `FetchArt` plugin:

-  the pre-sized thumbnails of the release group of an album would be ignored when the `maxwidth` option matches one of the available sizes (250, 500, and 1200 pixels).
    - This allows the release group to behave exactly like the release itself (which does take the pre-sized thumbnails into account).
- no cover would be found and downloaded when the `maxsize` option is defined, is in the pre-sized thumbnails list, but CAA does not provide thumbnails for this release/release group

This patch can significantly speed up the fetching process of some albums, as the default size of some cover arts can be quite large (e.g. [this cover ](https://coverartarchive.org/release-group/1cd77af1-5fe8-45ac-9953-710315c85604) is 56 MB and 7300 pixels wide, while the 1200px version is 450KB).

This patch also add a section in the `FetchArt` plugin documentation, detailing this behavior.

## To Do

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -399,10 +399,15 @@ class CoverArtArchive(RemoteArtSource):
                     if 'Front' not in item['types']:
                         continue
 
-                    if preferred_width:
-                        yield item['thumbnails'][preferred_width]
-                    else:
-                        yield item['image']
+                    # If there is a pre-sized thumbnail of the desired size
+                    # we select it. Otherwise, we return the raw image.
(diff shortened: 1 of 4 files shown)


sampsyo (approved):
Awesome!! This looks really handy. Thanks for the clean fix, and for extending the documentation to match. When/if the CI goes green, this looks ready to merge.

PR #4951, opened 2023-10-16
Separate plugin tests from the core tests

Serene-Arc:
As discussed on Zulip, this refactors the plugin tests into a specific folder. This is a minor but important first step in increasing the test base and tidying everything up, and allows developers to exclude the plugin tests with ease.

There are no changes to actual tests, beyond anything relating to paths, and all tests succeed as they did before.


JOJ0:
Would it be a bad idea tpshorten the path a little? ie test/plugin/... instead of test/plugin_tests/...

(plugin singular!)

I'm thinking of firing one particular test is rather long path already

python -m unittest test.test_blabla.BlaBla.test_something

Serene-Arc:
We should really be encouraging people to use pytest, not the unittest module. Then it's easier to ignore. 

I'm not wedded to the directory name. In general I think more descriptive names are better, but if you feel strongly about it, I can change it.

JOJ0:
> We should really be encouraging people to use pytest, not the unittest module. Then it's easier to ignore.
> 
> I'm not wedded to the directory name. In general I think more descriptive names are better, but if you feel strongly about it, I can change it.

For now I'm still up to changing it, yes. I think it's descriptive that a directory plugin/ within a directory named test/ is referring to unittests  :-)

Question on the side: Please open an issue about updating the docs regarding pytest. I'll try to take care of it, but since I'm a noob around pytest still probably it would be good if you provide some basic info about it. Just some quick basiscs on how to run tests in quick words, I'll try to do the calligraphy then :-))

First question: How to run a single test with e.g. my example command above with pytest.

Serene-Arc:
You literally just write the command `pytest` in the root of the project and it will collect and run all the tests. I will write specific instructions in a different PR, though. Pytest is better for a variety of reasons: there are a bunch of plugins we can make use of, and going forward there I intend to simplify and expand the tests wherever possible, which will probably make use of pytest features.

Serene-Arc:
@JOJ0  I just pushed the rename to just `plugin`. If you give your approval, we can merge.

JOJ0:
> You literally just write the command `pytest` in the root of the project and it will collect and run all the tests. I will write specific instructions in a different PR, though. Pytest is better for a variety of reasons: there are a bunch of plugins we can make use of, and going forward there I intend to simplify and expand the tests wherever possible, which will probably make use of pytest features.

I was referring to: How to run single tests :-)

JOJ0 (approved):
You forgot to change the imports to also use that new module name :-) Please go ahead and merge once that's done. Thanks!

Serene-Arc:
Always something you miss lol

sampsyo:
Just wanted to express gratitude for doing this reorganization! Seems like a great idea.

Serene-Arc:
Seemed like a good first step for improving our testing

PR #4961, opened 2023-10-20
Some CI fixes

sampsyo:
Attempt to fix some recent CI failures that are cropping up in the Windows build. (Although I can't yet see why these problems are Windows-specific.)


sampsyo:
This seems to have done it!

Issue #4955, opened 2023-10-18
path queries not returning all results


vredesbyyrd:
```
beets version 1.6.1
Python version 3.11.5
```
Note: I originally posted this as a discussion, as I was not confident the problem is a bug, now I am leaning more towards bug.

Hello,

Some background on the problem. My music collection is separated into 3 categories: `Artists`, `Composers` & `Scores`. The Directory structure reflects the categories. Years ago I accomplished this by using 3 separate beets configs, one for each category. I always thought this was convoluted so I dug into the doc looking for a more sane approach - Using [flexible attributes](https://beets.readthedocs.io/en/stable/guides/advanced.html#choose-a-path-style-manually-for-some-music) appears to be what I was looking for. 

The relevant part of my config now looks like: 

```yaml
library: /mnt/TARS/Media/library.db
directory: /mnt/TARS/Media/Music
paths:
    default: Artists/%the{$albumartist}/$year - $album%aunique{}/%if{$multidisc,Disc $disc/}$track - $title
    category:artists: Artists/%the{$albumartist}/$year - $album%aunique{}/%if{$multidisc,Disc $disc/}$track - $title
    category:composers: Composers/%asciify{$albumartist}/$year - %asciify{$album%aunique{}}/%if{$multidisc,Disc $disc/}%asciify{$track - $title}
    category:scores: Scores/%the{$albumartist}/$year - $album%aunique{}/%if{$multidisc,Disc $disc/}$track - $title
``` 
I first tried to tag all the files in each category by doing: 

```yaml
beet modify category="artists" path:/mnt/TARS/Media/Music/Artists
beet modify category="composers" path:/mnt/TARS/Media/Music/Composers
beet modify category="scores" path:/mnt/TARS/Media/Music/Scores
```
But this was not finding any files. To be sure:

`beet ls path:/mnt/TARS/Media/Music/Artists`

would not return anything. 

By a happy accident typo I noticed prepending `path::` to the filepath instead of `path:` would return all the files. I know the double `::` is for regex, but not sure why this finds the files when the single `:` does not. 

Perhaps wrecklessly, I ran the `beet modify category="foobar" path::/foo/bar` for each category and everything looks correct, e.g, `beet ls category:artists | wc -l` return the expected count of files, `24878`

So that's the background, this is the issue I cannot figure out - Running: 

```yaml
beet ls path:/mnt/TARS/Media/Music/Artists | wc -l
```
[...]

PR #4962, opened 2023-10-21
library: debug log parsed queries (and convert to f-strings)

wisp3rwind:
To help with debugging queries that behave unexpectedly. Not sure whether this is really going to be helpful, but while looking at #4955 I realized I would like to know exactly what the queries parse to, because to be honest, I never remember the details our query syntax.

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -166,8 +166,8 @@ class FieldQuery(Query, Generic[P]):
 
     def __repr__(self) -> str:
         return (
-            "{0.__class__.__name__}({0.field!r}, {0.pattern!r}, "
-            "{0.fast})".format(self)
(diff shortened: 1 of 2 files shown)


sampsyo (approved):
Very good idea! This also helps with the other "direction" I sometimes run into: I can remember the text query syntax, but I can't remember how to construct that query programmatically. With `beet -vv ls whatever`, now I can get a reminder.

Looks right to me, modulo one linter hint about a long line.

sampsyo:
(Oh, and I also meant to say thanks for the `__repr__` modernization that came along with this!)

wisp3rwind:
Fixed the overly long line and rebased for #4654 and #4961. There are some new lint failures now, but they're unrelated and already present in #4654.

PR #4964, opened 2023-10-21
Add retry logic for Spotify 

arsaboo:
## Description

Part 2 of the PR - #4943 

Currently, the plugin will endlessly keep retrying. I added the max-retries so that it stops after three retries. 

## To Do

### DIFF
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -157,7 +157,8 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
         with open(self.tokenfile, 'w') as f:
             json.dump({'access_token': self.access_token}, f)
 
-    def _handle_response(self, request_type, url, params=None):
+    def _handle_response(self, request_type, url, params=None, retry_count=0,
+                         max_retries=3):
         """Send a request, reauthenticating if necessary.
 
         :param request_type: Type of :class:`Request` constructor,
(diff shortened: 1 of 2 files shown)


beetsplug/spotify.py
```diff
@@ -195,12 +196,17 @@ def _handle_response(self, request_type, url, params=None):
                 raise SpotifyAPIError(f'API Error: {e.response.status_code}\n'
                                       f'URL: {url}\nparams: {params}')
             elif e.response.status_code == 429:
-                seconds = response.headers.get('Retry-After',
-                                               DEFAULT_WAITING_TIME)
-                self._log.debug(f'Too many API requests. Retrying after '
-                                f'{seconds} seconds.')
-                time.sleep(int(seconds) + 1)
-                return self._handle_response(request_type, url, params=params)
+                if retry_count < max_retries:
```
sampsyo, line 199: This could be _slightly_ easier to read if the logic were inverted, i.e.:

```py
if retry_count >= max_retries:
    raise SpotifyAPIError('Maximum retries reached.')
seconds = ...
```

…then at least the "normal case" doesn't even need more reindenting.

docs/changelog.rst
```diff
@@ -142,6 +142,7 @@ New features:
 
 Bug fixes:
 
+* Fix retry-logic for Spotify plugin.
```
sampsyo, line 145: It's nice to prefix plugin-specific fixes with `:doc:`/plugins/spotify`. Also maybe this could be slightly more specific: "Add a limit of 3 retries, instead of retrying endlessly when the API is not available."

sampsyo (commented):
Great! Here are a couple of suggestions.

sampsyo:
Excellent!!

PR #4970, opened 2023-10-22
fix formatting_check CI config

wisp3rwind:
cf. https://github.com/beetbox/beets/pull/4654, the `formatting_check.yml` is invalid and never runs, immediately aborting on parsing this configuration file.

### DIFF
--- a/.github/workflows/formatting_check.yml
+++ b/.github/workflows/formatting_check.yml
@@ -6,7 +6,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Install dependencies
-      - uses: actions/checkout@v3
-      - uses: paolorechia/pox@v1.0.1
+        uses: actions/checkout@v3
+
+      - name: Run formatting check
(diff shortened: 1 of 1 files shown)


wisp3rwind:
@Serene-Arc do we actually need the `pox` GH action? In our main CI config, as far as I can tell, we just directly execute `tox`. `pox` also seems to be essentially unmaintained (although that might not be a problem in principle; maybe this is really just done and doesn't require much maintenance). However, it does trigger some deprecation warnings, cf. https://github.com/beetbox/beets/actions/runs/6602808494

Serene-Arc:
@wisp3rwind probably not? I didn't realise the action was broken, sorry for that. I tried to keep it as close to the existing one as possible. It's likely that no, we don't need it.

wisp3rwind:
> @wisp3rwind probably not? I didn't realise the action was broken, sorry for that. I tried to keep it as close to the existing one as possible. It's likely that no, we don't need it.

No problem; I was just a little confused about the CI failures I got for #4962 :)

PR #4972, opened 2023-10-22
discogs: Populate `cover_art_url` attribute

### LINKED ISSUE #429: fetchart: Fetch cover art from discogs
I have the experience that discogs usually provides the best artwork at least for the music I need to tag and Amazon either does not have an image or it is..frankly speaking...crap.

henges:
## Description

Fixes #429.

#4778 added the `cover_art_url` flex attribute, allowing any plugin to supply a URL for cover art, which can then be downloaded by `fetchart`. This PR updates the `discogs` plugin to populate this attribute using the first value returned in the `images` field on the `Release` object from the Discogs API. 

This PR also fixes an issue in `fetchart` where an `AttributeError` was thrown by calling `album.cover_art_url` when this attribute isn't set on the album - even though it may be set on the first track of that album. 

I haven't included tests due to the low impact of the changes and the lack of preexisting test coverage for analogous functionality in these plugins, but am happy to do so if desired. It looks like this would require mocking out the Discogs client, so if this is needed, it'd be helpful to get some pointers to prior art around mocking in this project, if any.

Thanks in advance for your review! I love this project. : )

## To Do

- [-] Tests. (Encouraged but not strictly required.)


beetsplug/discogs.py
```diff
@@ -425,6 +425,8 @@ def get_album_info(self, result):
             catalogno = result.data["labels"][0].get("catno")
             labelid = result.data["labels"][0].get("id")
 
+        cover_art_url = self.select_cover_art(result)
```
arsaboo, line 428: Maybe we only update the value if the function does not return `None`.

Serene-Arc:
Hi. Thanks for the PR. You'll need to conform to the new formatting standards we have in place by running `tox-e format`.

beetsplug/discogs.py
henges: Sorry @arsaboo, I'm not sure what you mean. Do you mean we should not pass `cover_art_url` to the `AlbumInfo` constructor at all if its value is `None`?

sampsyo (approved):
Looks great! Many thanks!!

PR #4975, opened 2023-10-23
Use gender-neutral language

leetfin:
## Description

Remove a few missed gendered pronouns from code comments.


JOJ0 (approved):
Thanks a lot! Even though it's just code comments, still language is important! I stumbled upon exactly some of those weird wordings at some point already, and admittedly was to lazy to fix them, so many thanks again! Much appreciated!

PR #4987, opened 2023-11-03
 Sort and categorize config_default.yaml

RollingStar:
## Description

Sorted the default config into categories, more or less. Added a FIXME and URL for an undocumented feature but otherwise it just re-arranges the existing config. No changes to substance.

## To Do

### DIFF
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -160,9 +160,6 @@ sort_case_insensitive: yes
 
 # --------------- Autotagger ---------------
 
-# FIXME undocumented feature
-# https://github.com/beetbox/beets/blob/c15ccb16bf3bceadfdf50268818dd64f04de25f8/beets/autotag/__init__.py#L219
-
 overwrite_null:
   album: []
   track: []
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -145,7 +145,8 @@ New features:
   plugin which allows to replace fields based on a given library query.
 * :doc:`/plugins/lyrics`: Add LRCLIB as a new lyrics provider and a new
   `synced` option to prefer synced lyrics over plain lyrics.
-* Sorted the default configuration file :ref: into categories.
+* Sorted the default configuration file into categories.
+  :bug:`4987`
 
 Bug fixes:
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -145,6 +145,7 @@ New features:
   plugin which allows to replace fields based on a given library query.
 * :doc:`/plugins/lyrics`: Add LRCLIB as a new lyrics provider and a new
   `synced` option to prefer synced lyrics over plain lyrics.
+* Sorted the default configuration file :ref: into categories.
 
 Bug fixes:
 


JOJ0:
Yeah why not sort this. It might help contributors to find a good place for their potential new default. I think it's an awesome idea personally since I often look into that file and still am overwhelmed with all the things that are in there :-)

Some things definitely  _are_ "opinionated" like maybe for some persons: "hardlinking" is the most natural thing to do with beets, haha :-), but I'm not all opposed to group them like that. For the "mainstream" beets users, probably your grouping is just right and you won't hurt much feelings with it ;-)

beets/config_default.yaml
```diff
@@ -135,13 +158,14 @@ sort_album: albumartist+ album+
...
-    comp: Compilations/$album%aunique{}/$track $title
+# --------------- Autotagger ---------------
 
-statefile: state.pickle
+# FIXME undocumented feature
+# https://github.com/beetbox/beets/blob/c15ccb16bf3bceadfdf50268818dd64f04de25f8/beets/autotag/__init__.py#L219
 
+overwrite_null:
+  album: []
+  track: []
```
JOJ0, line 168: That is a very interesting catch! Do you want to quickly open a "missing docs" issue for it. Someone "who knows" might want do document it.

sampsyo: FWIW it's not _too_ uncommon for us to have undocumented config options. The way I think about this is that an undocumented option is better than not having the option at all (i.e., to have a magic number in the code). Sometimes these are undocumented intentionally, because we mostly hope people will not use them except in special circumstances, and sometimes we just haven't gotten around to it yet. In yet a third category, it might just be _so_ specific that it seems like not worth the space in the docs page!

I'm not 100% certain which category to put `overwrite_null` in. Probably just missing docs that we should add? But maybe it's just hyper-specific with an audience that rounds down to zero. FWIW, it was added in #3133.

RollingStar: The use case of this is I'm in my config or config_default more often than you might think. But it feels like I'm changing "basic" parts more often. Plugins, timid/quiet, replace/inline/paths are probably my most common changes. The esoteric and undocumented parts most people will never touch. This PR is about taming the beast of config_default and make it a more useful starting point for a user config.

This PR is secretly part 1 of 3 or more:
1. sort config
2. sort config page of docs
3. split config page of docs into basic/advanced
4. split config yaml into basic/advanced yamls (not sure how hard it would be to have cascading configs; I think JOJ0 does this now).
5. move inline function definitions to their own user_inline.**py** file to take advantage of code markup and symbol definitions in IDEs

In an ideal world I think every config option would at least point to the PR it was added, or the part of the code it exists in. That's often what I'm looking at when trying to figure out if a config feature is something I need to change. But even linking that stuff may be too much work for very esoteric options.

RollingStar:
> Some things definitely _are_ "opinionated" like maybe for some persons: "hardlinking" is the most natural thing to do with beets, haha :-), but I'm not all opposed to group them like that. For the "mainstream" beets users, probably your grouping is just right and you won't hurt much feelings with it ;-)

Hardlinks makes a lot of sense. But they apparently don't work on Windows (can we confirm that?) so I think it's instantly not a novice-level feature.

https://beets.readthedocs.io/en/latest/reference/config.html#link

Lots of stuff on Google looks like you can hardlink or something like it on Windows. https://www.reddit.com/r/sonarr/comments/w8vt85/can_someone_plesae_explain_how_to_hardlink_the/

beets/config_default.yaml
RollingStar: > I really hope that most people don't need to copy & paste the entirety of config_default.yaml

Okay, I'm not married to that idea. The reason I started with copying config_default for my personal config is I wanted to read through the config page and make sure I wasn't "missing" any options that would be helpful. Keeping the default value doesn't hurt anything (unless the value changes). [Also, I think part of it was I'm using the LSIO container and they change the config to their own default, and I wanted to change it back and take back control of the config.]

I wonder if there's a way to give people a blank config with the categories in this PR already filled in (possibly in the docs page). The user is free to use that or not, but it might help them keep some semblance of organization. Configs can get out of hand.

A sorted config_default also helps with maintenance on our end.

JOJ0: I suggest as a first step we could finalize this PR by doing these things:

- Remove the comment about the undocumented feature but add a proper entry to the documentation. Or if that's not possible at this time or seems tedious, just leave an issue about a missing config entry.
- Merge the sorted config_default.yaml

sampsyo: Sounds good to me!

docs/changelog.rst
RollingStar, line 1: I can't get docs to render on my PC so I can't test this.

docs/changelog.rst
```diff
@@ -145,6 +145,7 @@ New features:
   plugin which allows to replace fields based on a given library query.
 * :doc:`/plugins/lyrics`: Add LRCLIB as a new lyrics provider and a new
   `synced` option to prefer synced lyrics over plain lyrics.
+* Sorted the default configuration file :ref: into categories.
```
JOJ0, line 148: If I don't have an issue that is fixed/implemented I always put the id of the PR implementing it into the changelog. I don't know if that's good practice but I do it. I wanted to fixed your :ref:link link to point to config_default.yaml. I don't know how to do it, do you have an idea? i couldn't find anything in the docs in a quick search where we link to actual files in the code repo. I'm sure we are doing this somewhere / I did this before. 

Anyway, if anything renders wrongly we'll see it in the final docs built from master and could quickfix it there (no time to test locally atm myself)

```suggestion
* Sorted the default configuration file :ref: into categories.
  :bug:`4987`
```


For your problem with rendering docs locally open another issue or discussion or something describing your issues, I'll try to help sort it out :-)

beets/config_default.yaml
JOJ0: > The use case of this is I'm in my config or config_default more often than you might think. But it feels like I'm changing "basic" parts more often. Plugins, timid/quiet, replace/inline/paths are probably my most common changes. The esoteric and undocumented parts most people will never touch. This PR is about taming the beast of config_default and make it a more useful starting point for a user config.

I feel you. Often I find myself changing things back and forrh between beets runs. What I find most helpful is controlling things via cli options. If we don't have them we should add then. Recently @


> 
> 
> This PR is secretly part 1 of 3 or more:
> 
> 1. sort config
> 
> 2. sort config page of docs
> 
> 3. split config page of docs into basic/advanced
> 
> 4. split config yaml into basic/advanced yamls (not sure how hard it would be to have cascading configs; I think JOJ0 does this now).
> 
> 5. move inline function definitions to their own user_inline.**py** file to take advantage of code markup and symbol definitions in IDEs
> 
> 
> 
> In an ideal world I think every config option would at least point to the PR it was added, or the part of the code it exists in. That's often what I'm looking at when trying to figure out if a config feature is something I need to change. But even linking that stuff may be too much work for very esoteric options.

RollingStar:
Is this ready for merge now that I merged @JOJ0's suggestion?

PR #4989, opened 2023-11-05
Add LRCLIB as a provider for the lyrics plugin

mumumumu:
## Description

This PR adds support for [LRCLIB ](https://lrclib.net) as a lyrics provider.

I recently set up Navidrome on my home server and the web UI for that [only supports embedded, synced lyrics](https://github.com/navidrome/navidrome/issues/491#issuecomment-946782864). From what I can tell, the current lyric providers only return plain lyrics. I did a little research and found LRCLIB, which offers an [easy to use API](https://lrclib.net/docs) that includes both synced and plain lyrics. Note that all four parameters (track, artist, album, duration) to the `/get` endpoint are required.

Inspired by https://github.com/tranxuanthang/lrcget/issues/2#issuecomment-1471512530.

## To Do

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -290,8 +290,29 @@ class Backend:
             self._log.debug("failed to fetch: {0} ({1})", url, r.status_code)
             return None
 
-    def fetch(self, artist, title):
+    def fetch(self, artist, title, album=None, length=None):
         raise NotImplementedError()
+
+
+class LRCLib(Backend):
+    base_url = "https://lrclib.net/api/get"
+
+    def fetch(self, artist, title, album=None, length=None):
+        params = {
+            "artist_name": artist,
+            "track_name": title,
+            "album_name": album,
+            "duration": length,
+        }
+
+        try:
+            response = requests.get(self.base_url, params=params)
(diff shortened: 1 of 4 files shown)


sampsyo (commented):
Wow; this looks really great! I've reviewed the code carefully and everything looks good from here.

The one high-level question I have is: Do we want to have this source enabled by default, given that it returns synced lyrics by default? To be specific, "synced" lyrics from LRCLIB look like this:

```
[00:09.17] Come back from San Francisco
[00:12.97] It can't be all that pretty
[00:16.75] When all of New York City misses you
...
```

…which may be surprising to most users of the plugin, who (from other sources) are expecting plain "unsynced" lyrics.

One option could be to add a new top-level configuration option to the plugin, `synced`. This would prefer synced lyrics for any source that provides them (currently just this new source), and otherwise would only try to get plain lyrics. Would this make sense?

mumumumu:
@sampsyo thanks for the quick review, and that makes total sense. Added a `synced` option and updated the plugin docs.

sampsyo (approved):
Wonderful; this looks great! Thanks again for the awesome work on this.

PR #4994, opened 2023-11-10
Remove reference to detox

emiham:
`detox` has been superceded by a native option in `tox`, so it is no longer maintained. 

See: https://github.com/tox-dev/detox


wisp3rwind:
Thanks :tada:

PR #4996, opened 2023-11-12
Support writing ISRC to tracks in Spotify plugin

### LINKED ISSUE #4992: Support writing ISRC to tracks in Spotify plugin
Support ISRC track field in Spotify plugin.

The difficulty is that we currently if I'm not mistaken, only get a "Simplified Track Object" track object from Spotify while fetching album info: https://developer.spotify.com/documentation/web-api/reference/get-an-album -> look for this in the docs:

Screenshot 2023-11-10 at 09 48 10

I wanted to instantly open a PR to just add external_ids.isrc to the plugin but it's just not there in the response we currently get from the albums endpoint. We would need an additional request to this endpoint to get it: https://developer.spotify.com/documentation/web-api/reference/get-track

Because we need this data:

Screenshot 2023-11-10 at 09 51 16

There might be a better way, maybe asking the albums endpoint to include that field. Not sure if that's possible.

arsaboo:
## Description

Fixes #4992

## To Do

- [x] Documentation.
- [x] Changelog.
- [x] ~Tests.~

### DIFF
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -658,7 +658,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
                 self._log.debug("No track_id present for: {}", item)
                 continue
 
-            popularity, iscr, ean, upc = self.track_info(spotify_track_id)
+            popularity, isrc, ean, upc = self.track_info(spotify_track_id)
             item["spotify_track_popularity"] = popularity
             item["isrc"] = isrc
             item["ean"] = ean
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -678,7 +678,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
                 item.try_write()
 
     def track_info(self, track_id=None):
-        """Fetch a track popularity and other external_ids by its Spotify ID."""
+        """Fetch track popularity and other external_ids using its Spotify ID."""
         track_data = self._handle_response(
             requests.get, self.track_url + track_id
         )
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -305,7 +305,6 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
             mediums=max(medium_totals.keys()),
             data_source=self.data_source,
             data_url=album_data["external_urls"]["spotify"],
-            isrc=album_data["external_ids"].get("isrc"),
         )
 
     def _get_track(self, track_data):
@@ -338,7 +337,6 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
             medium_index=track_data["track_number"],
             data_source=self.data_source,
             data_url=track_data["external_urls"]["spotify"],
-            isrc=track_data["external_ids"].get("isrc"),
         )
 
     def track_for_id(self, track_id=None, track_data=None):
@@ -660,8 +658,9 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
                 self._log.debug("No track_id present for: {}", item)
                 continue
 
-            popularity = self.track_popularity(spotify_track_id)
-            item["spotify_track_popularity"] = popularity
+            info = self.track_info(spotify_track_id)
+            item["spotify_track_popularity"] = info[0]
+            item["isrc"] = info[1]
             audio_features = self.track_audio_features(spotify_track_id)
             if audio_features is None:
                 self._log.info("No audio features found for: {}", item)
@@ -676,13 +675,16 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
             if write:
                 item.try_write()
 
-    def track_popularity(self, track_id=None):
-        """Fetch a track popularity by its Spotify ID."""
+    def track_info(self, track_id=None):
+        """Fetch a track popularity and ISRC by its Spotify ID."""
         track_data = self._handle_response(
             requests.get, self.track_url + track_id
         )
-        self._log.debug("track_data: {}", track_data.get("popularity"))
-        return track_data.get("popularity")
+        self._log.debug("track_popularity: {} and track_isrc: {}",
+                        track_data.get("popularity"),
+                        track_data.get("external_ids").get("isrc"))
+        return [track_data.get("popularity"),
+                track_data.get("external_ids").get("isrc")]
 
     def track_audio_features(self, track_id=None):
         """Fetch track audio features by its Spotify ID."""
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -661,6 +661,8 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
             info = self.track_info(spotify_track_id)
             item["spotify_track_popularity"] = info[0]
             item["isrc"] = info[1]
+            item["ean"] = info[2]
+            item["upc"] = info[3]
             audio_features = self.track_audio_features(spotify_track_id)
             if audio_features is None:
                 self._log.info("No audio features found for: {}", item)
@@ -676,7 +678,7 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
                 item.try_write()
 
     def track_info(self, track_id=None):
-        """Fetch a track popularity and ISRC by its Spotify ID."""
+        """Fetch a track popularity and other external_ids by its Spotify ID."""
         track_data = self._handle_response(
             requests.get, self.track_url + track_id
         )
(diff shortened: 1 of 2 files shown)


arsaboo:
@JOJ0 can you test this out with an album that has ISRC? It seems Spotify returns ISRC at both the album and track levels. Right now, I am importing both. I have never used ISRC, so I'm not sure if that is the right approach. I will look for your guidance here.

JOJ0:
@arsaboo thanks for getting this started. In the meantime I quickfixed this for myself so I could move on with my personal project. Please just cherry-pick this commit into your PR branch: https://github.com/JOJ0/beets/commit/30e0fb12da7106eaecf65fd2923bae2a1a883e77

It implements proper fetching of the full track details and then writing to the TrackInfo object. It's tested, I'm using it.

Your current implementation won't work since, as I tried to explain [here](https://github.com/beetbox/beets/issues/4992#issue-1987191172), we currently only fetch a, so called "SimplifiedTrackObject", which definitely does not contain the track's ISRC. To get it we need to fetch a full track object!

JOJ0:
Regarding **album** ISRC I don't have a finished commit at hand but I just read the docs again and can report that I would understand that the album-level ISRC is already included via the endpoint we fetch, which is again this one: https://developer.spotify.com/documentation/web-api/reference/get-an-album and the example response contains it already :-) So your current implementation will do! 

If it helps, I can test from your branch, once you implemented both track-level and album-level ISRC's :-)

arsaboo:
Good catch. 

Would it make sense to create a separate function to update ISRC for those who want to use it? Otherwise, we are significantly increasing the number of API calls.

JOJ0:
> Good catch.
> 
> Would it make sense to create a separate function to update ISRC for those who want to use it? Otherwise, we are significantly increasing the number of API calls.

Yeah I'm also thinking about "too many requests errors" (where is our rate-limiting code actually? do we have one? :-))

On the other hand, beets fetches track ISRC per default when using MusicBrainz and Deezer. So making it configurable with Spotify plugin only would be odd or at least would be "non-streamlined" behaviour within beets. IMHO the way to go is trying to support similar things as good as possible throughout sources and try to have similar defaults, to get a streamlined user experience. (repeat: as good as possible)

This is how Deezer plugin does it in the `album_for_id` method, I tried to implement it similarly: 

https://github.com/beetbox/beets/blob/708a04d4a88eb955f261a21f0b1146c9e2ae0a55/beetsplug/deezer.py#L111-L117

JOJ0:
I was curious, we have ISRC for tracks fetching in MusicBrainz since 2020: b6fda634c42be59ce9805dab544ff5575a6f1216

BTW the MusicBrainz API is smarter about "I now want more info that is not available by default" - you can just tell the API to "include more info" on the inital requests. I wish Spotify's API could do that with album's - it could simply make it configurable of what info to include for the items object.

### Off-topic but interesting

From another personal project I played around a lot with the playlists endpoint. Here we have a 'fields" option that allows to "include additional" information as well as "restrict to what we really need", I see it as a kind of "filter" possibility. Have a look at the fields option in the docs here: https://developer.spotify.com/documentation/web-api/reference/get-playlist

arsaboo:
We do have the `_fetch_info` function that can be updated to fetch the ISRC:
https://github.com/beetbox/beets/blob/6cda741143781cc6d6b7309d31f10ecb325103ef/beetsplug/spotify.py#L640

JOJ0:
The only place where this function is called is here: https://github.com/beetbox/beets/blob/6cda741143781cc6d6b7309d31f10ecb325103ef/beetsplug/spotify.py#L480-L482

Will it be called when doing a regular import? I'm having a hard time reading that code at the  moment......

https://github.com/beetbox/beets/blob/6cda741143781cc6d6b7309d31f10ecb325103ef/beetsplug/spotify.py#L438-L485

arsaboo:
It is not called during regular import. The user will have to call `beet spotifysync`, which will refresh the optional information. While we are at it, I also included EAN and UPC.

JOJ0:
> It is not called during regular import. The user will have to call `beet spotifysync`, which will refresh the optional information. While we are at it, I also included EAN and UPC.

Well, then this is not the solution I was aiming for. I think this should be similar to other plugins and ISRC should definitely be fetched on initial import. Why make it different? I do not want a second step when importing my tracks via Spotify. I do not have to issue a second step when I use MusicBrainz or Deezer.

arsaboo:
The problem with that approach is that it will definitely cause API errors. Additional API calls for all the tracks for all the albums (even those that are not used) will cause excessive API usage (especially when importing a large number of albums). Even with the current implementation, I sometimes get API errors (partly fixed with #4943). I don't think we should add multiple calls for everyone (not sure how popular/useful the ISRC field is). 

Just run `beet import ... && beet spotifysync` 😉

arsaboo:
While the rate limits for Spotify are not published, folks [suggest](https://community.spotify.com/t5/Spotify-for-Developers/Web-API-ratelimit/m-p/5503153/highlight/true#M7931) that Spotify allows for approximately 180 requests per minute without returning the error 429. For comparison, Deezer [allows](https://stackoverflow.com/questions/28264392/does-the-deezer-api-request-quota-applies-by-url-or-oauth-token) around 50 per 5 seconds, i.e., 600 calls per minute.

JOJ0 (changes requested):
Hi and thanks a lot for the detailed information about the why's and how's! I now understand that at this point just adding an additional fetch to the tracks endpoint will cause rate-limiting trouble. I also now see that you already put some effort into rate-limiting in #4943 and #4946 I'm wondering why this still is not enough, it's actually pretty much what the docs suggest. Should we open another issue, or de we have one, to further discuss rate limiting?

arsaboo:
The problem is that Spotify often (or rather mostly) does not return the retry-after interval. The only option is to give up after a few tries, which is what I did in the previous PR. We can open another issue for a better retry logic, if required. 

I'll try to fix the other issues tomorrow (thanks for your detailed review).

beetsplug/spotify.py
```diff
@@ -674,13 +677,22 @@ def _fetch_info(self, items, write, force):
             if write:
                 item.try_write()
 
-    def track_popularity(self, track_id=None):
-        """Fetch a track popularity by its Spotify ID."""
+    def track_info(self, track_id=None):
+        """Fetch track popularity and other external_ids using its Spotify ID."""
```
JOJ0, line 681: The reason I corrected this sentence is that it was, and now still is, wrong in terms of content. It means this:

We fetch an ID named track popularity and we also fetch *other* external ID's. -> That's not true.

To make this sound correct you at least have to remove the word "other" in this sentence (if you don't like my suggested longer version)

Sorry for being picky here. Even it is just a developer's comment, I'd still consider this "documentation" and thus I'm worried about correctness a lot :-)

JOJ0:
> The problem is that Spotify often (or rather mostly) does not return the retry-after interval. The only option is to give up after a few tries, which is what I did in the previous PR. We can open another issue for a better retry logic, if required.

Yes please, if you find a minute, please open one. You already started in the very right direction with your retry logic. Going a little further there I'd like to tell you what we've learned when we implemented rate-limiting magic in python3-discogs-client a couple of years ago.

> I'll try to fix the other issues tomorrow (thanks for your detailed review).

Welcome! Thanks for quickly implementing this! Awesome new feature! Since you

JOJ0:
..., I sometimes get API errors (partly fixed with #4943). I don't think we should add multiple calls for everyone (not sure how popular/useful the ISRC field is).

Regarding ratelimiting: As suggested, let's better open a new issue and let's collect ideas together :-)

Regarding "how useful is ISRC / what can it be used for": As far as I understand it and use it, it is a unique ID throughout tagging sources and I think it currently is the best way to "link" Deezer, MusicBrainz and Spotify tagged items together

My personal usecase is, that I'm generating offline-m3u files as pools for my DJ use. I have most of the music I collect on Spotify in my offline-collection on my DJ computer as well. I'm generating m3u playlists using the ISRC. I have fallback searches for artist/title but actually ISRC is the most accurate way to generate those playlists.
 
> Once we have the track_ids, we can limit the calls for only those ids. This will significantly economize the number of calls.

Makes sense!
 
> Just run `beet import ... && beet spotifysync` 😉 or we can even use the `hook` plugin, which will have the same effect

Oh, great idea using the hook plugin! Thanks!

JOJ0:
> While the rate limits for Spotify are not published, folks [suggest](https://community.spotify.com/t5/Spotify-for-Developers/Web-API-ratelimit/m-p/5503153/highlight/true#M7931) that Spotify [allows](https://stackoverflow.com/questions/28264392/does-the-deezer-api-request-quota-applies-by-url-or-oauth-token) for approximately 180 requests per minute without returning the error 429. For comparison, Deezer allows around 50 per 5 seconds, i.e., 600 calls per minute.

Discogs has even less, 60 per minute: https://www.discogs.com/developers#page:home,header:home-rate-limiting

-> Let's discuss in new issue

PR #4998, opened 2023-11-18
Fix some docs links

cubicvoid:
## Description

This is just a small doc fix: update the Twitter link in README to point to the Mastodon account (the pinned forwarding address on X isn't accessible to non-logged-in users anymore), and fix a typo that broke the link to the code of conduct in CONTRIBUTING.

## To Do

- [x] ~~Changelog. (Add an entry to `docs/changelog.rst` to the bottom of one of the lists near the top of the document.)~~
- [x] ~~Tests. (Very much encouraged but not strictly required.)~~

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -24,7 +24,7 @@ We love to get contributions from our community—you! There are many ways
 to contribute, whether you’re a programmer or not.
 
 The first thing to do, regardless of how you'd like to contribute to the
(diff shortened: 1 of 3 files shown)


Serene-Arc:
Hi! Thanks for the PR. A changelog entry isn't required for this, and we don't have a workflow that requires it. That was under discussion but it was changed. 

It looks like the linting for the docs have failed. If you could fix that so it passes, I'll merge it.

PR #5003, opened 2023-11-23
Allow multiple plugins to provide getters for one field

### LINKED ISSUE #5002: beetsplug: only one plugin can set `template_fields` per field
### Problem

I ran into this when using the new `advancedrewrite` plugin on my production system together with the old `rewrite` plugin.
When using these two (or any two other plugins modifying `BeetsPlugin.template_fields`) together and specifying rules that both target the same field (e.g., `artist`), only the rules of one plugin concerning this field will be applied. Even worse, which plugin will be used depends on the plugin loading order (the one loaded last will be used), which seems to be random-ish: in the importer, the `rewrite` plugin is loaded last, when just invoking `beet`, the `advancedrewrite` plugin is loaded last. The rules for that specific field from the previously loaded plugin(s) are silently dropped.

### Investigation

I've pinned this bug down to the following line: [`beets.plugins.item_field_getters`](https://github.com/beetbox/beets/blob/36454a3883c9e5c656805152d2220e9496a7455c/beets/plugins.py#L454). The reason this breaks is that the `update` function on the `dict` overwrites previous getters, instead of merging them somehow.
Of course, the problem lies in the fact that rewrites can generally affect each other and do thus need a deterministic way of applying them.

### Suggested solution (implemented in #5003)

I suggest addressing this issue the following way:

* Rework the getter functions applied to `template_fields` or `album_template_fields` so that they either return a replacement value **or `None`** (possibly breaking change for 3rd-party plugins).
* Merge getter functions into a dict with lists of functions in `plugins.item_field_getters`.
* When evaluating the getters in `beets.dbcore.Model._get`, the **list is iterated and the value of the first value that isn't `None` is used**.
* To make sure the value that's used is actually deterministic, the getters can be sorted by the class names of the plugin providing them (alphabetically). In my opinion, that's not ideal, but it's at least consistent.

### Setup

* OS: Ubuntu Server 22.04
* Python version: 3.10
* beets version: 1.6.1-master (2023-11-23)
* Turning off plugins made problem go away: not applicable, bug specifically affects plugin handling

Maxr1998:
Fixes a bug where it wasn't possible for multiple plugins to define replacement functions for the same field. See #5002 for more details.

Closes #5002.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -430,8 +430,12 @@ class Model(ABC):
         """
         getters = self._getters()
         if key in getters:  # Computed.
-            return getters[key](self)
-        elif key in self._fields:  # Fixed.
+            # Use the result of the first getter that returns a value.
+            for getter in getters[key]:
+                result = getter(self)
+                if result is not None:
+                    return result
+        if key in self._fields:  # Fixed.
             if key in self._values_fixed:
                 return self._values_fixed[key]
             else:
--- a/beets/library.py
+++ b/beets/library.py
@@ -662,8 +662,8 @@ class Item(LibModel):
     @classmethod
     def _getters(cls):
         getters = plugins.item_field_getters()
-        getters["singleton"] = lambda i: i.album_id is None
-        getters["filesize"] = Item.try_filesize  # In bytes.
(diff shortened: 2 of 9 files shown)


Maxr1998:
Superseded by #5050.

PR #5006, opened 2023-11-24
Add more detail to option about terminal width in the docs

Serene-Arc:
I was going through the code, preparing to do a PR that makes it find the terminal width dynamically, but to my surprise that is already the case. So this is a short thing that just adds that information to the docs.


sampsyo:
This is great; thank you!

PR #5022, opened 2023-12-04
expose incremental_skip_later as cli option

### LINKED ISSUE #4958: Expose import.incremental_skip_later config option as CLI option
To reduce the time needed for manual prompting during the import of a large number of tracks and to import as much as possible without prompting, I run the import once with the `-i --quiet` options and the `import.incremental_skip_later` config option enabled and then again without the `--quiet` option and `import.incremental_skip_later` disabled within the configuration, letting beets prompt for the previously skipped tracks only.
However, currently that requires me to change the configuration file during the workflow which I'd like to avoid.

### Proposed solution

In addition to the configuration file the  `import.incremental_skip_later` option should be configurable via the CLI as e.g. `--incremental-skip-later` option of the `import` command.

Usage example:
```sh
beet import -siq --incremental-skip-later # no questions asked, tracks skipped in doubt
beet import -si                           # prompting for previously skipped tracks only
```

#### Goals

Improving the import workflow / making the import command more flexible.

#### Non-goals

Changing the import behaviour internally.

#### Anti-goals

Any additional complexity or more prompts. :smile:

mgoltzsche:
## Description

Fixes #4958

Exposes `import.incremental_skip_later` configuration option as CLI option.

## To Do

- [x] ~~Tests~~. (Very much encouraged but not strictly required.)

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1507,6 +1507,20 @@ import_cmd.parser.add_option(
     help="do not skip already-imported directories",
 )
 import_cmd.parser.add_option(
+    "-R",
+    "--incremental-skip-later",
+    action="store_true",
+    dest="incremental_skip_later",
+    help="do not record skipped files during incremental import",
+)
+import_cmd.parser.add_option(
+    "-r",
+    "--noincremental-skip-later",
+    action="store_false",
+    dest="incremental_skip_later",
+    help="record skipped files during incremental import",
+)
+import_cmd.parser.add_option(
     "--from-scratch",
     dest="from_scratch",
     action="store_true",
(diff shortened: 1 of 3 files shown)


sampsyo:
Looks good! I would like to invite everybody to please bikeshed the CLI's long option names… `--incremental-skip-later` and `--noincremental-skip-later`, while faithful transcriptions of the config option names, are a bit unwieldy (and the second one looks especially weird because of the missing `-` in an option that otherwise has `-`s, although I know all the other `--noX` options follow this pattern). Is there any sense in picking something more legible, even if it means the correspondence to the config option is further obscured?

mgoltzsche:
I just rebased the PR to resolve the conflict.

Serene-Arc:
Renaming a config option would be backwards incompatible with existing configurations from previous releases. We should refrain from doing that, unless the configuration option has not been officially released yet.

sampsyo:
Awesome; many thanks. Given the various other downsides, I think we should stick with the proposed name. Maybe someday we can get the bandwidth to discuss more holistic renamings of these related settings (in a backwards-compatible way that would allow the old names in old config files).

PR #5027, opened 2023-12-06
Fix failing test: test_nonexistant_db

Phil305:
## Description

Fixes #5021

The `test_nonexistant_db` test in `test_ui` would attempt to read from `stdin` under certain conditions (particularly when run directly, i.e. `pytest -k nonex test/test_ui.py`).

This diff fixes the issue.

**note:** I didn't **really** figure out what the root cause of the initial issue was (since the tests could sometimes pass when run with the entire test-suite), but if this works on other folks machines, then hey, good enough I suppose...

### DIFF
--- a/test/test_ui.py
+++ b/test/test_ui.py
@@ -937,7 +937,8 @@ class ConfigTest(unittest.TestCase, TestHelper, _common.Assertions):
             config.write("library: /xxx/yyy/not/a/real/path")
 
         with self.assertRaises(ui.UserError):
-            self.run_command("test", lib=None)
+            with control_stdin("n"):
+                self.run_command("test", lib=None)
 
     def test_user_config_file(self):
         with self.write_config_file() as file:


test/test_ui.py
```diff
@@ -934,10 +934,12 @@ def test_default_paths_preserved(self):
 
     def test_nonexistant_db(self):
         with self.write_config_file() as config:
-            config.write("library: /xxx/yyy/not/a/real/path")
+            with control_stdin("y"):
+                config.write("library: /xxx/yyy/not/a/real/path")
```
wisp3rwind, line 938: Is this actually doing anything? From quick glance at the code, there should be no way this could prompt for user input.

wisp3rwind (commented):
I think I'd be good to understand what is really going on here; this seems a lot like there's some global state not being properly restored after some preceding test. Or maybe some intricacies of stdin/out handling does not work as intended.

In general, it is quite unclear to me which error path the test is actually supposed to hit, because there are two possibilities:

- the library directory doesn't even exist (this is what leads to the prompt you see, and given the invalid path I'd say this is expected)
- the library directory exists, but opening the DB fails (Maybe the file isn't there, or it is corrupt. This raises `UserError`.)

Arguably, there should be tests for both cases.

test/test_ui.py
sampsyo: Right; I think we probably don't need the `control_stdin` wrapper here.

sampsyo (commented):
Thanks! It's likely this does address the proximate problem, but I think it might be nice to do a little extra homework here to identify the hidden test ordering dependency. In particular, we could do some experiments that would run tests in different orders/subsets to identify which one is causing this test to pass when it doesn't alone; that could identify some _other_ test that is (worrisomely) not resetting its changes to global state.

Phil305:
> Thanks! It's likely this does address the proximate problem, but I think it might be nice to do a little extra homework here to identify the hidden test ordering dependency. In particular, we could do some experiments that would run tests in different orders/subsets to identify which one is causing this test to pass when it doesn't alone; that could identify some _other_ test that is (worrisomely) not resetting its changes to global state.

Sounds good to me;

I think there might be some other hidden problems that reordering will bring out.

When I shuffled the test order around for a trial, the `test_tags_not_restored` test broke as well as the one I initially opened this issue for:
```
ERROR: test_nonexistant_db (test_ui.ConfigTest.test_nonexistant_db)
  File "/Users/pjulien/repos/beets/test/test_ui.py", line 940, in test_nonexistant_db
    self.run_command("test", lib=None)
  File "/Users/pjulien/repos/beets/test/helper.py", line 457, in run_command
    beets.ui._raw_main(_convert_args(list(args)), lib)
  File "/Users/pjulien/repos/beets/beets/ui/__init__.py", line 1848, in _raw_main
                                ^^^^^^^^^^^^^^^^^^^^
  File "/Users/pjulien/repos/beets/beets/ui/__init__.py", line 1697, in _setup
    lib = _open_library(config)
          ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pjulien/repos/beets/beets/ui/__init__.py", line 1757, in _open_library
    _ensure_db_directory_exists(dbpath)
  File "/Users/pjulien/repos/beets/beets/ui/__init__.py", line 1751, in _ensure_db_directory_exists
    os.makedirs(newpath)
  File "<frozen os>", line 215, in makedirs
  File "<frozen os>", line 215, in makedirs
  File "<frozen os>", line 215, in makedirs
  [Previous line repeated 1 more time]
  File "<frozen os>", line 225, in makedirs
OSError: [Errno 30] Read-only file system: b'/xxx'

FAIL: test_tags_not_restored (test_importer.ScrubbedImportTest.test_tags_not_restored)
  File "/Users/pjulien/repos/beets/test/test_importer.py", line 307, in test_tags_not_restored
    self.assertEqual(imported_file.artist, None)
AssertionError: 'Tag Artist' != None
```

Unfortunately, I didn't save the test order for this run, but I'll be playing around and learning more about the beets test suite ordering to see what's going on...

Also, I'm using `python3 test/testall.py` instead of `tox` to run all the tests; would that make any difference, or everything should be the same?

Serene-Arc:
I do plan to do a test overhaul 'soon'. I'll keep this in mind.

Phil305:
> I do plan to do a test overhaul 'soon'. I'll keep this in mind.

What do you have in mind for the overhaul (how big is the scope), and why is it needed?

I only have ~5 hours a week to contribute, but I'm interested in helping; seems like the perfect opportunity for me to learn more about the codebase

Phil305:
(whoops, accidentally closed, reopening until I finish fixing the bug)

test/test_ui.py
Phil305: I accidentally globbered by old commit and history with a botched `git push -f`, so I can't see what I had before, but this is what it **should** have been:

```python
with self.write_config_file() as config:
    config.write("library: /xxx/yyy/not/a/real/path")

with self.assertRaises(ui.UserError):
    with control_stdin("n"):
        self.run_command("test", lib=None)
```

Without it, I don't see anything globally set that's mocking stdin.
With `control_stdin` here, the call the call to `input` ([here](https://github.com/beetbox/beets/blob/master/beets/ui/__init__.py#L202)) is handled w/o need for user input, thereby making the test pass.

I'm not sure how this test was passing before **without** this?

I re-added it and it's passing now.
Just to make sure, I rebased on commit `c1a232ec` of `master`, and it's working.

Let me push this change again, this time correctly 🙂

Serene-Arc:
> What do you have in mind for the overhaul (how big is the scope), and why is it needed?

A lot of things. We have spotty test coverage with a lot of things not being tested at all. We also have some complicated tests that could be improved with pytest to make setup and teardown easier. I haven't done a full inventory and I'll be doing it piecemeal, file by file. I just need time/energy/motivation.

sampsyo:
> Hey guys, I'm wondering how this test was ever passing if it's supposed to fail on stdin?

Indeed; I do think that's what's suspicious here… it seems like some kind of input-wrapping state is being held over from some _other_ test and therefore suppressing the failure here. So I could be wrong about this, but I think we actually have two problems:

* This test is broken and it _should_ fail.
* Some other test is leaving beets in a state that suppresses this test's failure.

Does that seem right?

PR #5034, opened 2023-12-07
Bring back NO_COLOR support.

### LINKED ISSUE #5028: Bring back NO_COLOR support?
Back in 2019 I added support for the [NO_COLOR environment variable](https://no-color.org/). This was removed in be290e54441987fdb80437f7982f4d5ad01bac59 as part of a refactor. Unless I'm mistaken it seems there's no way to disable colors from the command line anymore. It must be done in the config file. That's really inconvenient.

What was the reason for removing `NO_COLOR` support? Could it be brought back?

djl:
Fixes #5028.

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -591,7 +591,7 @@ def colorize(color_name, text):
     """Colorize text if colored output is enabled. (Like _colorize but
     conditional.)
     """
-    if config["ui"]["color"]:
+    if config["ui"]["color"] and "NO_COLOR" not in os.environ:
         global COLORS
         if not COLORS:
             # Read all color configurations and set global variable COLORS.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -13,7 +13,7 @@ Major new features:
 
 * The beets importer UI received a major overhaul. Several new configuration
   options are available for customizing layout and colors: :ref:`ui_options`.
-  :bug:`3721`
+  :bug:`3721` :bug:`5028`
 
 New features:
 


JOJ0:
Thanks for the fix @djl, please add a quick changelog entry.

djl:
I didn't add a changelog entry for this as the original entry from when it was added is still there and there hasn't been a release with the removal yet. Does it make sense to add a changelog entry for this?

JOJ0:
Fun fact: I had similar thoughts: when do we need changelog, when not. I thought we had this fearure 1.6.0 already though. didnt check.

help @Serene-Arc @RollingStar

JOJ0:
@Serene-Arc agreed that actually we wouldn't need the entry, but I think I've made up my mind on what I think would be bestin this situation: Dear @djl please add just the bug/issue ID to the existing changelog entry for the UI overhaul. That way we make sure we link to all the relevant pieces of that particular entry in the changelog (at the very top "major ..." list):

https://beets.readthedocs.io/en/latest/changelog.html#in-development

@RollingStar we recently talked about stuff like that elsewhere. The changelog is currently our best bet to record PR talk alongside actual code changes. If we don't add this "extra PR/fix" to the changelog, this information would get lost. It wouldn't show up if you were to investigate using the changelog on "hat happened exactly when the UI was redisgned"

Certainly with that approach there would be some "after UI overhaul fixes" be missing at exactly that changelog entry.

Does that idea make sense to you all?

djl:
Done!

JOJ0:
I think this is safe to merge. Fellow maintainers can still comment on my general "how to changelog question"

PR #5046, opened 2023-12-13
smartplaylist: add extm3u/extinf/m3u8 support

mgoltzsche:
## Description

Adds the boolean configuration option `smartplaylist.extm3u` to allow generating extm3u playlists that contain additional information (artist, title, length) per track. See [Extended M3U (Wikipedia)](https://en.wikipedia.org/wiki/M3U#Extended_M3U) and [playlist specification for HTTP Live Streaming (RFC8216)](https://datatracker.ietf.org/doc/html/rfc8216#section-4.3.2.1).
This is to be able to display meaningful metadata and search tracks within a playlist within a player without having to load the linked audio files of that playlist.

## To Do

### DIFF
--- a/beetsplug/smartplaylist.py
+++ b/beetsplug/smartplaylist.py
@@ -49,6 +49,7 @@ class SmartPlaylistPlugin(BeetsPlugin):
                 "prefix": "",
                 "urlencode": False,
                 "pretend_paths": False,
+                "extm3u": False,
             }
         )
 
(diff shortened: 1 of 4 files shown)


Serene-Arc:
Thanks! This is a very useful change, and it all looks good to me.

mgoltzsche:
I realize it is disadvantageous to chose a boolean flag to specify the format because we may as well want to specify JSON as playlist output format at some point.
Thus, do you mind if I create another PR to change the flag to `--format` (or so) and make it of `type=string`?

Serene-Arc:
Sure! Go right ahead.

mgoltzsche:
Okay, I created #5053 now to fix this.

PR #5047, opened 2023-12-14
smartplaylist: expose config as CLI options

mgoltzsche:
## Description

Add CLI options to `splupdate` command:
* `--playlist-dir`, `-d`
* `--relative-to`
* `--prefix`
* `--urlencode`
* `--forward-slash`
* `--pretend-paths`

## To Do

- [x] Documentation. (If you've added a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)
- [x] Changelog. (Add an entry to `docs/changelog.rst` to the bottom of one of the lists near the top of the document.)
- [x] ~~Tests~~. (Very much encouraged but not strictly required.)

Issue #5037, opened 2023-12-08
Expose playlists within web API


mgoltzsche:
I am using mopidy with the mopidy-beets plugin to play music that it streams from a beets web server that runs on another machine. However, while mopidy-beets lets me search for tracks and browse albums, unfortunately it doesn't let me access playlists (probably because the beets web API doesn't expose them).
On the other hand, using beets' smartplaylist plugin, I have defined rules to generate playlists within my beets config under `smartplaylist.playlists` already. Now ideally I'd like to browse those playlists within mopidy.
One way to do that could be to enhance the mopidy-beets plugin to be able to configure the same rules as for the smartplaylist beets plugin. However, that would require users to duplicate the playlist configuration and maintain it in two different places.
Instead, to maintain the playlist configuration in a single/central place, the beets web API could expose the playlists defined by the smartplaylist plugin's configuration. Alternatively, there could be a separate object introduced within the beets config to define playlists and both the web and smartplaylist plugin would obtain their configuration also from that configuration object. Either way, clients such as mopidy-beets could be enhanced to provide more useful and individual/personalized ways of browsing a beets library using the playlists defined within that library.

### Proposed solution

Let the beets web API expose the playlists defined by the existing `smartplaylist.playlists` beets configuration.
Two new GET endpoints should be added that each return a JSON response:
* `/playlist/` to list the available playlists (only their names really).
* `/playlist/{name}` to list the tracks contained within a particular playlist (specified by path param) using the same response format as the `/item/` endpoint. The endpoint would run all configured queries for the referenced playlist and return the merged results in the same way the smartplaylist plugin does.

Example response body of the `/playlist/` endpoint:
```json
{
  "playlists": [
    {
      "name": "jazz.m3u"
    },
  ]
}
```

Example response body of the `/playlist/{name}` endpoint:
```json
{
  "items": [
    {
      "id": 1,
      "title": "Bag's Groove",
      "artist": "Miles Davis",
      ...
    },
    ...
  ]
}
```

### Objective

Allow remote clients to easily access the playlists defined within the beets library.

#### Goals

Expose the playlists that can be defined and generated using the smartplaylist plugin within the beets web API.

#### Non-goals

[...]

PR #5048, opened 2023-12-15
smartplaylist: add --uri-format option

mgoltzsche:
Beets web API already allows remote players to access audio files but it doesn't provide a way to expose the playlists defined using the smartplaylist plugin.
Now the smartplaylist plugin provides an option to generate ID-based item URIs/URLs instead of paths.
Once playlists are generated this way, they can be served using a regular HTTP server such as nginx.

To provide sufficient flexibility for various ways of integrating beets remotely (e.g. beets API, beets API with context path, AURA API, mopidy resource URI, etc), the new option has been defined as a template with an `$id` placeholder (assuming each remote integration requires a different path schema but they all rely on using the beets item `id` as identifier/path segment).

To prevent local path-related plugin configuration from leaking into a HTTP URL-based playlist generation (invoked with CLI option in addition to the local playlists generated into another directory), setting the new option makes the plugin ignore the other path-related options `prefix`, `relative_to`, `forward_slash` and `urlencode`.

Usage examples:
* `beet splupdate --uri-format 'http://beets:8337/item/$id/file'` (for beets web API)
* `beet splupdate --uri-format 'http://beets:8337/aura/tracks/$id/audio'` (for AURA API)

(While it was already possible to generate playlists containing HTTP URLs previously using the `prefix` option, it did not allow to generate ID-based URLs pointing to the beets web API but required to expose the audio files using a web server directly and refer to them using their file system `$path`.)

Depends on:
* #5047

Relates to:
* #5037
* #5046
* #5053

## To Do

### DIFF
--- a/beetsplug/smartplaylist.py
+++ b/beetsplug/smartplaylist.py
@@ -45,6 +45,7 @@ class SmartPlaylistPlugin(BeetsPlugin):
                 "playlist_dir": ".",
                 "auto": True,
                 "playlists": [],
+                "uri_format": None,
                 "forward_slash": False,
                 "prefix": "",
                 "urlencode": False,
@@ -108,6 +109,12 @@ class SmartPlaylistPlugin(BeetsPlugin):
             "--urlencode",
             action="store_true",
             help="URL-encode all paths.",
+        )
+        spl_update.parser.add_option(
+            "--uri-format",
+            dest="uri_format",
+            type="string",
+            help="playlist item URI template, e.g. http://beets:8337/item/$id/file.",
         )
         spl_update.parser.add_option(
             "--output",
@@ -247,6 +254,8 @@ class SmartPlaylistPlugin(BeetsPlugin):
 
         playlist_dir = self.config["playlist_dir"].as_filename()
         playlist_dir = bytestring_path(playlist_dir)
+        tpl = self.config["uri_format"].get()
+        prefix = bytestring_path(self.config["prefix"].as_str())
         relative_to = self.config["relative_to"].get()
         if relative_to:
             relative_to = normpath(relative_to)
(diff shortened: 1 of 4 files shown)


sampsyo:
Looks good overall, and it doesn't seem likely to disrupt other users of this plugin!

Here is one small design-choice: currently, the template uses the literal string `$id` and a plain old `str.replace` to fill it in. Would it make sense to future-proof this a little by using proper Python string templates? The placeholder would be `{id}`, and we'd fill it in like `tpl.format(id=...)`. Maybe that's over-engineering and we never want that level of complexity, but I just thought I'd ask. :smiley:

mgoltzsche:
@sampsyo I am fine with making the change you suggested. However, can you elaborate in which way your suggested change is making it more future-proof please.
My reasoning to pick the `$id` placeholder was to provide a UX that is onsistent with the path templates within the configuration while at the same time I didn't want to use `item.evaluate_template` for this purpose yet since it supports a bunch of other placeholders that are not required yet (as far as I can see) but eventually somebody may require them one day and even have the need to add the song title as URL-encoded path segment or so which could be possible using the template functionality that is also used for the file paths. If it would be decided one day that such a feature is required, `tpl.replace("$id", ...)` could simply be replaced with `item.evaluate_template()`.

mgoltzsche:
Thinking about consistency, I realize it would be more consistent to call the option `--format` instead of `--uri-template`, aligned with the `beet ls` command's option with the same name.
Correspondingly, the option to enable m3u8 should be renamed to `--output` (or so)...
WDYT?

mgoltzsche:
I adjusted both PRs correspondingly (`--output` and `--format` options). Though, corresponding to the `ls` command's `--format` option, it is still using `$id` as placeholder. Do you want me to change that?

mgoltzsche:
I renamed the option again: it is now called `--uri-format` to be more intuitively usable and leave the door open for templating additional fields in the future consistently.
Now the PR is really ready for re-review :sweat_smile:.
Do you still want me to use `{id}` as placeholder, taking the UX consistency into account?

sampsyo:
> I am fine with making the change you suggested. However, can you please elaborate in which way your suggested change is making it more future-proof.

Basically, I just meant that—if we ever want to have more complex formatting going on here than just substituting a single number with no conversion—we might want a more flexible way to specify the format. But you make a good point that, even if we did that, we might still want beets's `$foo`-style templating instead of Python's `{foo}`-style templating!

sampsyo (approved):
This all looks great to me!! I like the new name `uri_format`; it is a bit more consistent. I'll merge this now!

PR #5051, opened 2023-12-15
advancedrewrite: Add note about quoting issues to docs

Maxr1998:
Ran into this earlier and decided to add a note to the docs about this.


Maxr1998:
Yep, quotes inside the query are retained, those around the query are removed, haha. But generally it's useful to also support and queries like for `beet ls`.

PR #5052, opened 2023-12-15
advancedrewrite: Apply same rewrite to more corresponding album fields

Maxr1998:
Felt a bit inconsistent as is, so the same rewrites are now applied to the simple and advanced rules, and more rules were added.


sampsyo (approved):
Very nice; thanks!!

PR #5053, opened 2023-12-15
smartplaylist: change option --extm3u to --output

mgoltzsche:
The boolean flags `--extm3u` and `--no-extm3u` are replaced with a string option `--output=m3u|m3u8`.
This reduces the amount of options and allows to evolve the CLI to support more playlist output formats in the future (e.g. JSON) without polluting the CLI at that point.

This is a follow-up of #5046.
Relates to #5048.

## To Do


mgoltzsche:
I renamed the option from `--format` to `--output` now to allow #5048 to use `--format`, aligned with the `beet ls` command.

PR #5058, opened 2023-12-20
Add initial version of the Listenbrainz plugin

### LINKED ISSUE #1719: ListenBrainz plugin
We should add a plugin to integrate with [ListenBrainz](https://listenbrainz.org). ListenBrainz is a new MetaBrainz project similar to Last.fm: it tracks your listening habits.

An initial version could look very much like the existing [lastimport](http://docs.beets.io/en/v1.3.15/plugins/lastimport.html) plugin: just download listen counts and such for music in your database. The data would be stored in flexible attributes.

arsaboo:
## Description

Fixes #1719

I added a basic version of the LB plugin that imports the play history. It uses the same process used in the Lastimport plugin to update playcounts. 

There are few additional functions that I added here to import playlists created by Listenbrainz. These playlists can then be imported in other programs such as Plex. 

## To Do

### DIFF
--- a/beetsplug/lastimport.py
+++ b/beetsplug/lastimport.py
@@ -204,9 +204,13 @@ def process_tracks(lib, tracks, log):
 
     for num in range(0, total):
         song = None
-        trackid = tracks[num]["mbid"].strip() or None
-        artist = tracks[num]["artist"].get("name", "").strip() or None
-        title = tracks[num]["name"].strip() or None
+        trackid = tracks[num]["mbid"].strip() if tracks[num]["mbid"] else None
+        artist = (
+            tracks[num]["artist"].get("name", "").strip()
+            if tracks[num]["artist"].get("name", "")
+            else None
+        )
+        title = tracks[num]["name"].strip() if tracks[num]["name"] else None
         album = ""
         if "album" in tracks[num]:
             album = tracks[num]["album"].get("name", "").strip()
--- a/beetsplug/lastimport.py
+++ b/beetsplug/lastimport.py
@@ -204,9 +204,9 @@ def process_tracks(lib, tracks, log):
 
     for num in range(0, total):
         song = None
-        trackid = tracks[num]["mbid"].strip()
-        artist = tracks[num]["artist"].get("name", "").strip()
-        title = tracks[num]["name"].strip()
+        trackid = tracks[num]["mbid"].strip() or None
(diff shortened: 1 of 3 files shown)


sampsyo:
Awesome! This already looks fantastic! No objection to merging right away if you think this is at a usable initial state.

I also noticed this contains a few changes to the last.fm plugin. Was that intentional? I didn't review them carefully, although they seem credible. If they fix bugs in the plugin, maybe that deserves a changelog entry of its own.

arsaboo:
@sampsyo Yes, I have tested it, and it is ready to be merged. 

I made minor changes to the `process_tracks` function in `lastimport` to improve error handling and reuse the same function. I am not entirely happy with the `process_tracks` function as it does a poor job of matching (e.g., right now, it only matches 3 of the 25 tracks even though they are all in my library). It may be due to the fact that I am using Spotify as the tag source. I do want to create a better matching function, but that can be done in a separate PR.

sampsyo:
Sounds good! Would you mind summarizing the changes to `lastimport` in their own changelog entry? Then this is clearly good to go!

Kernald:
Correct me if I'm wrong, but this seems to be fetching the whole ListenBrainz history on each run - could we leverage `min_ts` somehow to avoid that maybe?

arsaboo:
Yes, all the elements could (and should) be configured. We still need to carefully think about how to use them (which is why I called this PR `initial version` 😉). I honestly don't use the `play_count` information, but I am sure someone will find it useful and update the code to use those variables (heck, even I may do it at some point if no one gets to it). 

The current code has been reviewed and approved by @sampsyo, so I would not change it. Once this PR is merged, we can create a feature request for additional configurations and discuss the best approach to implement the same.

Serene-Arc:
I'll merge this then, since it's been approved. I should note that I have attempted my own tool for listenbrainz [here](https://github.com/Serene-Arc/listenbrainz-playlist-uploader), though it is implemented in Rust instead of Python. I still have to reach out to the listenbrainz team to try and get the song matching working as well as it can. Right now I can get it to about 90%+ but still not 100%.

arsaboo:
@Serene-Arc I will try to extend this plugin to include playlist uploads. The song-matching issue could be on LB's end...see a related [issue](https://community.metabrainz.org/t/how-is-listenbrainz-failing-to-identity-the-correct-recording/639331).

PR #5044, opened 2023-12-13
advancedrewrite: Support simple syntax and improve advanced syntax

Maxr1998:
## Description

Rework the `advancedrewrite` plugin as discussed in #5002.

## Checklist

- [X] Update documentation
- [x] Write tests
- [ ] Add changelog

PR #5063, opened 2024-01-01
advancedrewrite: Fix simple rules being overwritten by advanced rules

Maxr1998:
Well, this is awkward. The second for-loop for the advanced rules ended up overwriting the template fields for the simple rules, nullifying the main reason for merging the simple rule support into this plugin.

This bug was introduced in #5044, and previous test cases unfortunately didn't cover it. A combined test case has thus been introduced as a regression test for this bug.
Please let me know if you'd like to have a changelog entry for this issue.


Serene-Arc:
lgtm! Thanks for expanding the test coverage too.

PR #5064, opened 2024-01-01
advancedrewrite: Fix multi-valued fields specified as a single string

Maxr1998:
Unfortunately, I mistakenly assumed that `list()` was equivalent to `[]`. However, this transforms a rewrite `Value` into `['V', 'a', 'l', 'u', 'e']` instead of `['Value']` as it was intended.


Serene-Arc:
Seems good to me!

PR #5111, opened 2024-02-14
Update lyrics.rst

swishyswoosh:
looks like this backend was replaced with musixmatch

## Description

Fixes #X.  

(...)

## To Do

### DIFF
--- a/docs/plugins/lyrics.rst
+++ b/docs/plugins/lyrics.rst
@@ -59,7 +59,7 @@ configuration file. The available options are:
   sources known to be scrapeable.
 - **sources**: List of sources to search for lyrics. An asterisk ``*`` expands
   to all available sources.
-  Default: ``google genius tekstowo lrclib``, i.e., all the available sources. The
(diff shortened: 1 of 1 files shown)


jackwilsdon:
The defaults for 1.6.0 (the latest release) are `google`, `musixmatch`, `genius` and `tekstowo`: https://github.com/beetbox/beets/blob/19371805e70c34e06ddd1ee3bd2e23a23827f845/beetsplug/lyrics.py#L688

This was correctly documented at the time: https://github.com/beetbox/beets/blob/v1.6.0/docs/plugins/lyrics.rst#configuration

Since 1.6.0, the new `lrclib` source has been added (in #4989) and `musixmatch` has been removed from the default list (in #4616). These changes aren't yet in a release but are present on master, so the documentation is correct.

You can also see the difference on the hosted documentation:
* 1.6.0: https://beets.readthedocs.io/en/v1.6.0/plugins/lyrics.html#configuration
* master: https://beets.readthedocs.io/en/latest/plugins/lyrics.html#configuration

PR #5121, opened 2024-02-24
smartplaylist: allow exporting item fields, rename m3u8 output to extm3u

mgoltzsche:
## Description

Allow generating extm3u playlists so that they contain additional item fields such as the `id`.
    
The feature is required by the [webm3u plugin](https://github.com/mgoltzsche/beets-webm3u) (M3U server) to transform playlists using a request based item URI template which may require additional fields such as the `id`, e.g. `beets:library:track;$id`.

## To Do

### DIFF
--- a/test/plugins/test_smartplaylist.py
+++ b/test/plugins/test_smartplaylist.py
@@ -241,6 +241,58 @@ class SmartPlaylistTest(_common.TestCase):
             + b"http://beets:8337/files/tagada.mp3\n",
         )
 
+    def test_playlist_update_output_m3u8_fields(self):
+        spl = SmartPlaylistPlugin()
+
+        i = MagicMock()
+        type(i).artist = PropertyMock(return_value="Fake Artist")
+        type(i).title = PropertyMock(return_value="fake Title")
+        type(i).length = PropertyMock(return_value=300.123)
+        type(i).path = PropertyMock(return_value=b"/tagada.mp3")
+        a = {"id": 456, "genre": "Fake Genre"}
+        i.__getitem__.side_effect = a.__getitem__
+        i.evaluate_template.side_effect = lambda pl, _: pl.replace(
+            b"$title",
+            b"ta:ga:da",
+        ).decode()
+
+        lib = Mock()
+        lib.replacements = CHAR_REPLACE
+        lib.items.return_value = [i]
+        lib.albums.return_value = []
+
+        q = Mock()
+        a_q = Mock()
+        pl = b"$title-my<playlist>.m3u", (q, None), (a_q, None)
+        spl._matched_playlists = [pl]
+
+        dir = bytestring_path(mkdtemp())
+        config["smartplaylist"]["output"] = "m3u8"
+        config["smartplaylist"]["relative_to"] = False
+        config["smartplaylist"]["playlist_dir"] = py3_path(dir)
+        config["smartplaylist"]["fields"] = ["id", "genre"]
+        try:
+            spl.update_playlists(lib)
+        except Exception:
+            rmtree(syspath(dir))
+            raise
+
+        lib.items.assert_called_once_with(q, None)
+        lib.albums.assert_called_once_with(a_q, None)
+
+        m3u_filepath = path.join(dir, b"ta_ga_da-my_playlist_.m3u")
+        self.assertExists(m3u_filepath)
+        with open(syspath(m3u_filepath), "rb") as f:
+            content = f.read()
+        rmtree(syspath(dir))
+
+        self.assertEqual(
+            content,
+            b"#EXTM3U\n"
+            + b'#EXTINF:300 id="456" genre="Fake Genre",Fake Artist - fake Title\n'
+            + b"/tagada.mp3\n",
+        )
+
     def test_playlist_update_uri_format(self):
         spl = SmartPlaylistPlugin()
 
--- a/test/plugins/test_smartplaylist.py
+++ b/test/plugins/test_smartplaylist.py
@@ -191,7 +191,7 @@ class SmartPlaylistTest(_common.TestCase):
 
         self.assertEqual(content, b"/tagada.mp3\n")
 
-    def test_playlist_update_output_m3u8(self):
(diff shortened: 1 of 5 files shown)


Serene-Arc (approved):

Serene-Arc:
Thank you for this PR! The code seems fine but might I ask that you add some more to the docs? Perhaps the keys to the attributes most likely to be used and an explanation for when they might be required, such as the web service you specified.

JOJ0:
Hi @mgoltzsche, great addition the smartplaylist and very interesting plugin you wrote. Please also add a link to your plugin's repo to this section of the docs when you are coming back to working on this PR: https://beets.readthedocs.io/en/latest/plugins/index.html#interoperability

mgoltzsche:
I rebased the PR now to resolve the merge conflict, added an entry to the interoperability section of the plugins/index.rst, added more info to the smartplaylist.rst.
Also, I changed the option name from `attributes` to `fields`.
Unfortunately, the gh cache of the PR seems to be inconsistent since it doesn't show my force-pushed commit :disappointed:....

mgoltzsche:
Okay, as described [here](https://stackoverflow.com/questions/45626986/updating-pull-request-after-review-changes-commit-doesnt-show-up-in-the-pull-re) re-selecting `beetbox:master` as base branch within the GitHub GUI updates the PR and triggers CI but now the diff also includes all other commits on the master branch since this PR was created :disappointed:. However, my changes are still within a single commit which you could re-review.

JOJ0:
Hi, I don't find time to really look into the code you wrote but I think @Serene-Arc approved that already. Just one nitpick about terminology. I'm not sure if m3u8 is the right one, I think that just tells that characters could be unicode and not ascii. At least according to wikipedia. Maybe that is not accurate anyway? I don't know, what was your source?

I think it should rather be "Extended M3U", and maybe the output option should call that `extm3u` as well. It looks like this was incorrect before your changes here! Not sure if we should fix that here now if we don't want to break existing users configs though! :-/

https://en.wikipedia.org/wiki/M3U#Extended_M3U

https://en.wikipedia.org/wiki/M3U#M3U8

And I'm still not sure if I interpret that wikipedia article entirely correct...

mgoltzsche:
Indeed, I added that m3u8 format feature in a previous PR and indeed, it should be called extm3u instead and I was going to address that within a separate PR.
Eventually I'll add a commit to this PR to change that tomorrow...

mgoltzsche:
Since I rebased this PR, added documentation as @Serene-Arc requested and changed the option name from `attributes` to `fields`, technically she has to re-review it :sweat_smile:.

JOJ0:
> Indeed, I added that m3u8 format feature in a previous PR and indeed, it should be called extm3u instead and I was going to address that within a separate PR.
> 
> Eventually I'll add a commit to this PR to change that tomorrow...

Ah ol thanks for letting me know, so m3u8 format option us maybe only existing since a couple of months? In that case yeah, the impact might not be too massive on the userbase.

I agree that changing is a good idea.

mgoltzsche:
> so m3u8 format option is maybe only existing since a couple of months?

Yes, the `--output=m3u|m3u8` option was introduced in #5053.

mgoltzsche:
Okay, now I renamed the `output` option value `m3u8` to `extm3u`.
Also, I added a link to the [Beetstream plugin](https://github.com/BinaryBrain/Beetstream) to the interoperability section of the plugin documentation, given that its playlist endpoints [now also serve the smartplaylist generated playlists](https://github.com/BinaryBrain/Beetstream/pull/13).

mgoltzsche:
I updated my previous comment after I made a few more changes to the documentation.
Please re-review!

mgoltzsche:
@JOJ0 are you okay with having the webm3u plugin link within the "Other plugins" section of the documentation as opposed the interoperability section? Otherwise, do you think the webrouter plugin should also be listed in the interoperability documentation section?

JOJ0:
Ah yes please put everything into the Other Plugins section. Seems like we file all of externally hosted plugins there. My suggestion here was wrong. I'm sorry: https://github.com/beetbox/beets/pull/5121#issuecomment-1982637778

mgoltzsche:
Rebased to resolve merge conflict within changelog.

JOJ0:
Sorry for the massive delay in reviewing, I wanted to do this earlier. Code is fine, there is only one thing that I suggest to improve: Please add one or two example lines to the docs. How will a playlist entry look like with a specific config.

test/plugins/test_smartplaylist.py
```diff
@@ -241,6 +241,58 @@ def test_playlist_update_output_m3u8(self):
...
+        m3u_filepath = path.join(dir, b"ta_ga_da-my_playlist_.m3u")
+        self.assertExists(m3u_filepath)
+        with open(syspath(m3u_filepath), "rb") as f:
+            content = f.read()
+        rmtree(syspath(dir))
+
+        self.assertEqual(
+            content,
+            b"#EXTM3U\n"
+            + b'#EXTINF:300 id="456" genre="Fake Genre",Fake Artist - fake Title\n'
```
JOJ0, line 292: Something like this: Add a config example and the resulting line on the playlist file to the docs.

JOJ0 (changes requested):
Sorry for the massive delay in reviewing, I wanted to do this earlier. Code is fine, there is only one thing that I suggest to improve: Please add one or two example lines to the docs. How will a playlist entry look like with a specific config?

mgoltzsche:
Okay, I added an example configuration and extm3u playlist to the documentation now. Please re-review!

mgoltzsche:
After some problems with the docs linter now the PR is ready for review again.
Also, I made the documentation mention the autogenre plugin.

JOJ0 (approved):

JOJ0:
BTW I like the tiny PlaylistItem class, something like that could also help with my never finished project of providing a general m3u handling solution in beets. Currently it's rather unspectacular, it has save/load functionality only: 

oops oof-topic alert :-)

Issue #2300, opened 2016-12-02
WAV Support


haakobja:
Please allow WAV-files to be imported

sampsyo:
Hi! As with #1160, do you have any references for how metadata tags might be saved in .wav files? Is there a standard way, or even a *de facto* common practice?

You might also be interested in reading #250, the thread about AIFF support (which we now have).

haakobja:
It seems that the WAV-format may contain metadata-blocks of different types: http://wavmetadata.blogspot.com
https://en.wikipedia.org/wiki/Resource_Interchange_File_Format

I do not know of if there is a de facto common practice. But it would be nice to be able to import wav-files. Currently I haven't been able to do so. It may of course be that I haven't found the correct way to import wav files into my beets library.

sampsyo:
You're right; it's currently not possible.

The problem, as you've no doubt surmised, is that we don't know exactly how to write tags to WAV files. Mutagen doesn't support it, so this is unlikely to happen immediately.

I know this isn't much of a solution, but have you considered switching to a more widely used lossless format, such as FLAC or ALAC?

haakobja:
Yes, I could convert the wav to flac. In that case, it would be nice to use the convert plugin during import. I do not think it is run during import now, or  have i configured it wrong?

sampsyo:
Hmm, interesting—no, convert-on-import won't quite work here, because beets needs to be able to read the metadata even for files that it's just converting. So, at least for the moment, you'd need to use some other tool to do the conversion. Interesting idea, though!

brandonwkipp:
What if instead of trying to insert metadata tags into wav files, which sort of isn't recommended since it messes with the container structure, one still had the ability to rename wav files based on user configuration?

sampsyo:
Yes, that's another option—supporting some file types that can't store metadata at all. It's worth considering, but we haven't done that for any other formats yet, so it would require a few architectural (and documentation) changes.

RollingStar:
This may create even more problems, but what about converting the WAV to FLAC, leaving the WAVs where they are (or wherever the user wants them), and importing the FLAC files to the normal sorted library location for beets?

sampsyo:
Right—in fact, that could be accomplished already with the `convert` plugin if beets was allowed to scan WAV files.

sllawap:
I would be quite happy with an "import wav, convert to FLAC, write metadata" approach. Would this be a lower friction route than full support for non-metadata storing formats?

[...]

PR #5122, opened 2024-02-24
convert: Correctly identify WAVE format as lossless

Bobo1239:
## Description

I've noticed that `beet convert` with the `never_convert_lossy_files: yes` just copies `.wav` files instead of converting them. This fixes that. (see commit message for details)

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -40,7 +40,7 @@ ALIASES = {
     "vorbis": "ogg",
 }
 
-LOSSLESS_FORMATS = ["ape", "flac", "alac", "wav", "aiff"]
+LOSSLESS_FORMATS = ["ape", "flac", "alac", "wave", "aiff"]
 
 
 def replace_ext(path, ext):
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -153,6 +153,9 @@ New features:
 * :doc:`/plugins/smartplaylist`: Add new option `smartplaylist.uri_format`.
 * Sorted the default configuration file into categories.
   :bug:`4987`
+* :doc:`/plugins/convert`: Don't treat WAVE (`.wav`) files as lossy anymore
+  when using the `never_convert_lossy_files` option. They will get transcoded
+  like the other lossless formats.
 
 Bug fixes:
 


arsaboo:
Isn't copying the wav file the desired option in this case? I am not sure what the bug is here.

Bobo1239:
The reasoning behind `never_convert_lossy_files` is that we don't want to re-encode lossy file formats to the target format due to [quality loss of repeated conversion](https://en.wikipedia.org/wiki/Generation_loss). WAVE is a lossless format so should be re-encoded just like FLACs do. (especially since they're uncompressed)

wisp3rwind:
In any case, `wav` was in the list of lossless formats already, but that was apparently the incorrect key. If I'm reading the code correctly at a glance, the `Item.format` field is initialized from the human-readable representation of `mediafile`'s format field, which would be `WAVE`, rather than its internal `type` field, which would indeed take the value `wav`.

JOJ0:
Related unmerged PR https://github.com/beetbox/beets/pull/4656

Serene-Arc:
I prefer that change tbh, since we don't have the test coverage to determine if this is a breaking change on some systems.

Bobo1239:
@JOJ0 Thanks for finding that. Not sure how I missed it before submitting this one...

Did some `git blame`ing and imo `wav` should never have been returned by `item.format`. (unless they manually modified mediafile to fix this bug...) Relevant commits:
- https://github.com/beetbox/mediafile/commit/010a359cd62f7041440a315da873a9b3e27a86e3 (2010) is the commit that added the the `format` property and already uses the `TYPES` dict for the returned value. No support for WAVE.
- https://github.com/beetbox/beets/commit/9d55179d2d88e586d1051b6e404659e579e640f7 (2014) added the `never_convert_lossy_files` option which uses `item.format`.
- https://github.com/beetbox/mediafile/commit/832f3d895c665edfc29f8cd7fca41ce628b608af (2021) Adds `wav: WAVE` to the the `TYPES` dict for the first time. Previously WAVE files weren't even importable according to https://github.com/beetbox/beets/issues/2300. So I think the last commit just wasn't able to test that WAVE files worked as intended.

JOJ0:
No worries, that other PR is abandoned anyway, I'll close it now!

My efforts of moving on with the topic went silent when I couldnt push to that existing PR's branch.

Anyway, thanks for picking this issue up. About time we fix it!

Serene-Arc:
Definitely! I need to take a look at our formatting CI task too, it's malfunctioning.

Serene-Arc:
Formatting CI fixed with #5131

Serene-Arc:
Thanks for doing the digging for this @Bobo1239, it seems good to merge.

PR #5142, opened 2024-03-11
Support non-POSIX text editor file paths for Windows.

sgunn80808:
## Description

Fixes issue with `shlex.split` parsing Windows editor program file paths as POSIX.

This resulted in the following error for a program located in `C:\Program Files` when running `beet edit`

```
error: could not run editor command 'C:Program': [WinError 2] The system cannot find the file specified
```

Setting `posix = not is_windows` changes the parsing mode for Windows platform and allows Windows file paths to be used to specify the text editor.

## To Do

- [x] ~Documentation~
- [x] Changelog
- [ ] Tests

### DIFF
--- a/beetsplug/edit.py
+++ b/beetsplug/edit.py
@@ -17,6 +17,7 @@
 
 import codecs
 import os
+import platform
 import shlex
 import subprocess
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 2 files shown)


sgunn80808:
I marked `Tests` as non-mandatory because I couldn't find a good place to add a test for this change. Please comment here if you have any suggestions for where a test should go and I can add one. Thanks!

wisp3rwind:
I'm wondering whether on Windows, we should be doing any splitting at all: If I remember correctly, things will be concatenated into a single string to pass to Windows APIs anyway. Wouldn't it also work to do

```py
cmd = util.editor_command()
cmd = [cmd] if is_windows else shlex.split()
```

? The documentation of `shlex` seems to suggest that it generally shouldn't be used on non-Unix-like platforms.

snejus:
Closing due to lack of author response.

PR #5147, opened 2024-03-16
autotagger: match file name if the file has no title metadata

mklein-de:
This improves autotagging of items without metadata

### DIFF
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -19,6 +19,7 @@ releases and tracks.
 from __future__ import annotations
 
 import datetime
+import os
 import re
 from enum import IntEnum
 from typing import (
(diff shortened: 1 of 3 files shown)


arsaboo:
We already have a [fromfilename](https://beets.readthedocs.io/en/v2.0.0/plugins/fromfilename.html) plugin. Does this add any additional features?

mklein-de:
> We already have a [fromfilename](https://beets.readthedocs.io/en/v2.0.0/plugins/fromfilename.html) plugin. Does this add any additional features?

ah, didn't see the plugin... This doesn't add any feature, except that it work in the default configuration without any plugins enables

snejus:
I have mixed feelings about this change. Currently, when I try to import a file that has no metadata, I can instantly see what's going on: there are 0 results, and I see in the logs that artist/album fields are empty.

If we default to the filename, the autotagger may return some results, but the logs won't make it clear that the tags in the file were empty. Those results will not be relevant for any other filename except when it has the title.

Serene-Arc:
I think if we have the plugin, we shouldn't add this to the base code. Besides, that plugin is actually packaged with beets. You don't have to install it so much as enable it. Thanks for the PR @mklein-de but I think this isn't needed. Even if we were to merge this, we would probably have to have a configuration option governing it anyway, so it wouldn't actually change the user experience. If you think we should include the `fromfilename` plugin in the default configuration for beets, that's another PR.

Issue #5171, opened 2024-03-31
beets leaks file descriptors


Serranya:
### Problem

I'm running `beet import` with the chroma plugin enabled. After some time beets cannot open any more files. I think what happens is that beet runs out of the maximum allowed file descriptors.

When I run `lsof | grep python` I can see that there are over 600 open sockets.

The problem goes away when I disable the chroma plugin.

The problem seems to occur because the fingerprinting via library fails on some files. While the fingerprinting eventually succeeds as chroma tries the next backend (fpcalc) the socket from the first attemt seems to stay open.

I patched beets locally to pass `force_fpcalc` to https://github.com/beetbox/pyacoustid. This also causes the problem to dissapear.

### Setup

* OS: Alpine Linux 3.19
* Python version: Python 3.11.8
* beets version: 1.6.0
* Turning off plugins made problem go away (yes/no): yes

My configuration (output of `beet config`) is:

```yaml
directory: /serra-storage/Music
library: /serra-storage/.Musiclibrary.db

plugins: chroma

import:
    move: no
    copy: no
per_disc_numbering: yes
chroma:
    auto: yes
acoustid:
    apikey: REDACTED
```

PR #5172, opened 2024-03-31
Allow the user to control how chroma does the fingerprinting

Serranya:
## Description

This is a workaround to #5171. It does not adress the underlying problem. But
I think the patch might be worth including anyway.

## To Do


wisp3rwind:
While we do have configurable backends for other plugins (fetchart, lyrics, etc), I don't think we should be adding a new configuration option if its only purpose is to workaround a bug. Let's first try to figure out what exactly the problem in the linked issues is. If it can be fixed, we should do that. If it can't be fixed easily (or we can't determine the problem), maybe we should simply always use the CLI tool.

PR #5184, opened 2024-04-09
Enhancement: Refactor Test Code for Efficiency and Quality

freddiewanah:
## Description
As a first-time contributor with a background in researching Python unit tests, I conducted a thorough static analysis of the project's test code. This examination revealed various "test smells" that could potentially degrade both the efficiency and quality of our tests. To address these issues, I've refactored the relevant code segments. The modifications I propose not only aim to eliminate these inefficiencies but are also expected to reduce execution time within GitHub Actions. This reduction could contribute to lower operational costs for the project.

Test smells refactored:

- Suboptimal Assert: Instead of using statements such as assertIsNone, assertIsInstance, always simply use assertTrue or assertFalse. This will decrease the code overall readability and increase the execution time as extra logic needed.
- Similar Test Methods: Test methods within the same test class that have similar code. Refactor such cases with unittest.Subtest can not only reduce code redundancy, but also saves execution time by caching similar code base.
- General Fixture: Not all parameters in the setup function were used. Given that Setup function will be triggered everytime when the test method under the same class is executed, we should comment out the unused fixtures for the efficiency concern. For some the test class, we should also consider to break them into small classes.
- Redundant Print: Having print called within the test method. This was considered not a good practice as printing costs extra time and will normally hidden in the long output from the CICD. Instead, put them into the assert messages for debug purposes.
- Sleepy Test: occurs when developers need to pause the execution of statements in a test case for a certain duration. Sometimes the sleep command can be inevitable, but instead of time.sleep, consider to use asyncio.sleep. Also, consider to comment out the sleep is that is not inevitable.

I didn't change any of the testing logic or testing values. The refactoring only focuses on reducing the test smells.

Please feel free to let me know if you are interested in more refactoring like this or if there're any changes that you don't wish.

### DIFF
--- a/test/plugins/test_bareasc.py
+++ b/test/plugins/test_bareasc.py
@@ -3,7 +3,6 @@
 
 """Tests for the 'bareasc' plugin."""
 
-
 import unittest
 
 from beets import logging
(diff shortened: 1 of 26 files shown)


Serene-Arc:
Hi, thanks for the PR. I'm fine merging these changes but the tests need to be working first. If you could review the failing tests in the workflow and fix them, that would be great and we can move forwards after that.

freddiewanah:
Hi @Serene-Arc thank you for the review, I've fixed the failings in the test and passed all the tests on my forked repository ([here](https://github.com/freddiewanah/beets/actions/runs/8650535580)). Should be good now :)

test/plugins/test_player.py
```diff
@@ -549,7 +549,7 @@ def test_cmd_noidle(self):
             # Manually send a command without reading a response.
             request = client.serialise_command("idle")
             client.sock.sendall(request)
-            time.sleep(0.01)
+            asyncio.sleep(0.01)
```
Dr-Blank, line 552: need to await the asynchronous sleep, for that need to make the test method async and for that need to make the class inherit from `unittest.IsolatedAsyncioTestCase` and after that every method will have to be made async. Highly suggest to make this a different pr to make test cases asynchronous.

test/plugins/test_bareasc.py
```diff
@@ -28,96 +27,44 @@ def setUp(self):
...
+                ["without umlaut or e", "with umlaut"],
+            ),  # Bare-ASCII search, with umlaut.
+        ]
+
+        for query, expected_titles in test_cases:
+            with self.subTest(query=query, expected_titles=expected_titles):
+                items = self.lib.items(query)
+                titles = [item.title for item in items]
+                self.assertEqual(len(items), len(expected_titles))
+                self.assertEqual(titles, expected_titles)
```
Dr-Blank, line 68: redundant checks
```suggestion
                self.assertListEqual(
                    [item.title for item in items], expected_titles
                )

```

test/plugins/test_player.py
Serene-Arc: Ah, I missed that, might be why the test is failing. If this is a problem, I suggest waiting until I attempt to refactor tests to use pytest.

freddiewanah: In that case, I'll revert it back to time.sleep. And maybe leave a TODO mark here?

freddiewanah: Yeah, for `time.sleep` I believe it's the only one.

Serene-Arc: Leave it as time.sleep then. And yeah we use it to run the tests but most of the test framework is still unittest. So it'd be good to use the features of the better framework.

Serene-Arc:
If you could change the `time` statement back @freddiewanah, I think it'll be fine to merge.

PR #5186, opened 2024-04-11
Fix CI run of #5152 

Dr-Blank:
## Description
minor fix

Fixes CI run of #5152 

did not want to make it a part of that PR since it is unrelated and blame gets assigned correctly.

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -938,7 +938,7 @@ def command_output(
     if proc.returncode:
         raise subprocess.CalledProcessError(
             returncode=proc.returncode,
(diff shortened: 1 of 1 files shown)


Dr-Blank:
Is the changelog really necessary for this pr?

Serene-Arc:
It's not necessary. The failure is just a reminder that it hasn't been done, and we're still discussing whether it should be changed to a message. Don't worry about it.

Serene-Arc:
Given that this is contained in #5152, I'm going to close this one. Doing both is unnecessary and will create a merge conflict if we do both.

PR #5192, opened 2024-04-14
Fix extension issue in convert plugin when exporting a playlist

pierreay:
## Description

Fix extension substitution inside path of the exported playlist.

Before this, the exported playlist contained relative paths pointing to the
converted files BUT the extension were not substituted comparing to before and
the after the conversion. Therefore, running the playlist will fail for files
which have been converted and where extension have changed.

Example:
1. Convert `/path/to/library/artist.flac` to `/path/to/converted/artist.mp3` using the `-m playlist.m3u` command-line flag.
2. Open the generated playlist, and find the incorrect path `/path/to/converted/artist.flac` inside.

## To Do

- [X] Documentation
- [X] Changelog
- [X] Tests

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -48,6 +48,8 @@ def replace_ext(path, ext):
 
     The new extension must not contain a leading dot.
     """
+    assert isinstance(path, bytes)
+    assert isinstance(ext, bytes)
     ext_dot = b"." + ext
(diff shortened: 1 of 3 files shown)


pierreay:
I screwed up the commits. Sorry for the noise. I will perform two cleaned pull requests.

PR #5203, opened 2024-04-20
 Fix extension issue in convert plugin when exporting a playlist

pierreay:
## Description

Fix extension substitution inside path of the exported playlist.

Before this, the exported playlist contained relative paths pointing to the
converted files BUT the extension were not substituted comparing to before and
the after the conversion. Therefore, running the playlist will fail for files
which have been converted and where extension have changed.

Example:
1. Convert `/path/to/library/artist.flac` to `/path/to/converted/artist.mp3` using the `-m playlist.m3u` command-line flag.
2. Open the generated playlist, and find the incorrect path `/path/to/converted/artist.flac` inside.

## To Do

- [X] Documentation
- [X] Changelog
- [X] Tests

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -48,6 +48,8 @@ def replace_ext(path, ext):
 
     The new extension must not contain a leading dot.
     """
+    assert isinstance(path, bytes)
+    assert isinstance(ext, bytes)
     ext_dot = b"." + ext
     return os.path.splitext(path)[0] + ext_dot
 
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -599,6 +599,8 @@ class ConvertPlugin(BeetsPlugin):
             items,
         )
 
+        # If the user supplied a playlist name, create a playlist containing
+        # all converted titles using this name.
         if playlist:
             # Playlist paths are understood as relative to the dest directory.
             pl_normpath = util.normpath(playlist)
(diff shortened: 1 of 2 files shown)


beetsplug/convert.py
```diff
@@ -48,6 +48,8 @@ def replace_ext(path, ext):
 
     The new extension must not contain a leading dot.
     """
+    assert isinstance(path, bytes)
+    assert isinstance(ext, bytes)
```
Serene-Arc, line 52: Asserting things like this isn't a good way to handle it. It's the equivalent of throwing up your hands and saying 'screw it'. Is there a way to test for the problem and do something? The best solution would be to make sure that something that isn't bytes isn't passed in. Do you have reason to believe that something else will be passed in?

Serene-Arc (changes requested):
Hi! Thanks for the PR. I've left a comment or two. Would you also be able to add a test that fails under the old code and succeeds with the new?

pierreay:
> Hi! Thanks for the PR. I've left a comment or two. Would you also be able to add a test that fails under the old code and succeeds with the new?

I'm not familiar with unit tests in Python, but since you ask, I will give a try!

beetsplug/convert.py
pierreay: I think it's just a matter of another (bad ?) developer habit. Since I sometimes struggles to find what kind of data type is the input of a function, and since passing another data type will lead to errors, I'm used to perform sanity check using assertion to catch developers errors (not user errors). 

That being say, I did not had strong reasons to do so here nor to believe that something else would be passed in, except a developer error. If it is not appreciated by the coding habits / rules of Beets, for sure I will remove this.

pierreay:
I added the appropriate test case! ;)

pierreay:
Hello, kind reminder, is there anything that I can do on my side to see this PR being merged? :)

PR #5204, opened 2024-04-20
Add feature of refreshing converted files if original files are modified

pierreay:
## Description

By default, the `convert` plugin will skip every pending conversion if the converted file is already existing on the filesystem. However, when making changes (_e.g._, tags) in the original files, the changes are not reflected to the converted library. Adding the `--refresh` command-line option and the `refresh` option to the `convert` plugin allows to re-convert the original files if they are modified more recently than the time when converted files were converted.

## To Do

- [X] Documentation.
- [X] Changelog.
- [X] Tests.

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -428,9 +428,10 @@ class ConvertPlugin(BeetsPlugin):
         # Delete existing destination files when original files have been
         # modified since the last conversion. NOTE: Only when not using the
         # --keep-new option because I'm not sure what to do in this case.
-        if ((refresh and not keep_new) and
-            (os.path.exists(util.syspath(dest))) and
-            (os.path.getmtime(util.syspath(item.path)) > os.path.getmtime(util.syspath(dest)))):
+        if ((refresh and not keep_new)
+            and (os.path.exists(util.syspath(dest)))
+            and (os.path.getmtime(util.syspath(item.path)) 
+                 > os.path.getmtime(util.syspath(dest)))):
             if pretend:
                 self._log.info(
                     "rm {0} (original file modified)",
@@ -438,7 +439,8 @@ class ConvertPlugin(BeetsPlugin):
                 )
             else:
                 self._log.info(
-                    "Removing {0} (original file modified)", util.displayable_path(dest)
+                    "Removing {0} (original file modified)",
+                    util.displayable_path(dest)
                 )
                 util.remove(util.syspath(dest))
 
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -428,10 +428,14 @@ class ConvertPlugin(BeetsPlugin):
         # Delete existing destination files when original files have been
         # modified since the last conversion. NOTE: Only when not using the
         # --keep-new option because I'm not sure what to do in this case.
-        if ((refresh and not keep_new)
+        if (
+            (refresh and not keep_new)
             and (os.path.exists(util.syspath(dest)))
-            and (os.path.getmtime(util.syspath(item.path))
-                 > os.path.getmtime(util.syspath(dest)))):
+            and (
+                os.path.getmtime(util.syspath(item.path))
+                > os.path.getmtime(util.syspath(dest))
+            )
+        ):
             if pretend:
                 self._log.info(
                     "rm {0} (original file modified)",
@@ -440,7 +444,7 @@ class ConvertPlugin(BeetsPlugin):
             else:
                 self._log.info(
                     "Removing {0} (original file modified)",
-                    util.displayable_path(dest)
+                    util.displayable_path(dest),
                 )
                 util.remove(util.syspath(dest))
 
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -440,14 +440,16 @@ class ConvertPlugin(BeetsPlugin):
 
         if os.path.exists(dest):
             # Delete existing destination files when original files have been modified since the last convert run.
-            if refresh and os.path.getmtime(original) > os.path.getmtime(dest):
+            if refresh and os.path.getmtime(original) > os.path.getmtime(
+                dest
+            ):
                 self._log.info(
                     "Removing {0} (original file modified)",
                     util.displayable_path(dest),
                 )
                 if not pretend:
                     util.remove(dest)
-            else
+            else:
                 self._log.info(
                     "Skipping {0} (target file exists)",
                     util.displayable_path(item.path),
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -439,7 +439,8 @@ class ConvertPlugin(BeetsPlugin):
             converted = dest
 
         if os.path.exists(dest):
-            # Delete existing destination files when original files have been modified since the last convert run.
+            # Delete existing destination files when original files have
+            # been modified since the last convert run.
             if refresh and os.path.getmtime(original) > os.path.getmtime(
                 dest
             ):
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -153,6 +153,7 @@ class ConvertPlugin(BeetsPlugin):
             "--refresh",
             action="store_true",
             dest="refresh",
+            default=self.config["refresh"].get(),
             help="reconvert if original file is newer than converted file",
         )
         cmd.parser.add_option(
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -263,6 +263,10 @@ class ConvertPlugin(BeetsPlugin):
         return self.config["force"].get(bool)
 
     @cached_property
+    def refresh(self) -> bool:
+        return self.config["refresh"].get(bool)
+
+    @cached_property
     def hardlink(self) -> bool:
         return self.config["hardlink"].get(bool)
 
@@ -315,7 +319,7 @@ class ConvertPlugin(BeetsPlugin):
             # Filter items based on should_transcode function
             items = [item for item in items if self.should_transcode(item)]
 
-            self._parallel_convert(items, refresh=False, keep_new=False)
+            self._parallel_convert(items, keep_new=False)
 
     # Utilities converted from functions to methods on logging overhaul
 
@@ -409,9 +413,14 @@ class ConvertPlugin(BeetsPlugin):
         )
 
     @pipeline.mutator_stage
-    def convert_item(self, refresh: bool, keep_new: bool, item: Item) -> None:
+    def convert_item(self, keep_new: bool, item: Item) -> None:
         """Convert an Item from the library."""
-        pretend, link, hardlink = self.pretend, self.link, self.hardlink
+        pretend, link, hardlink, refresh = (
+            self.pretend,
+            self.link,
+            self.hardlink,
+            self.refresh,
+        )
         command, ext = self.command
 
         dest = self.get_item_destination(item)
@@ -701,9 +710,7 @@ class ConvertPlugin(BeetsPlugin):
                 items_paths.append(os.path.relpath(item_path, pl_dir))
 
         self._parallel_convert(
-            items,
-            refresh=self.config["refresh"].get(bool),
-            keep_new=self.config["keep_new"].get(bool),
+            items, keep_new=self.config["keep_new"].get(bool)
         )
 
         if playlist:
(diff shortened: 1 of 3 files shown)


beetsplug/convert.py
```diff
@@ -372,6 +384,29 @@ def convert_item(
...
+                    self._log.info(
+                        "rm {0} (original file modified)",
+                        util.displayable_path(dest),
+                    )
+                else:
+                    self._log.info(
+                        "Removing {0} (original file modified)",
+                        util.displayable_path(dest),
+                    )
+                    util.remove(util.syspath(dest))
```
snejus, line 408: You can simplify this

```suggestion
                self._log.info(
                    "Removing {0} (original file modified)",
                    util.displayable_path(dest),
                )
                if not pretend:
                    util.remove(dest)
```

beetsplug/convert.py
```diff
@@ -372,6 +384,29 @@ def convert_item(
...
 
+            # Delete existing destination files when original files have been
+            # modified since the last conversion. NOTE: Only when not using the
+            # --keep-new option because I'm not sure what to do in this case.
+            if (
+                (refresh and not keep_new)
+                and (os.path.exists(util.syspath(dest)))
+                and (
+                    os.path.getmtime(util.syspath(item.path))
+                    > os.path.getmtime(util.syspath(dest))
```
snejus, line 395: No need to call `util.syspath` here - both `item.path` and `item.destination` outputs are already in the expected form.

beetsplug/convert.py
```diff
@@ -372,6 +384,29 @@ def convert_item(
                 with _fs_lock:
                     util.mkdirall(dest)
 
+            # Delete existing destination files when original files have been
+            # modified since the last conversion. NOTE: Only when not using the
+            # --keep-new option because I'm not sure what to do in this case.
+            if (
```
snejus, line 390: I think this may result in some unintended behaviour: see `should_transcode` check _below_: it seems that already existing encoded files will be removed regardless of this check (whether this music file should be transcoded)?

beetsplug/convert.py
```diff
@@ -372,6 +384,29 @@ def convert_item(
...
+                (refresh and not keep_new)
```
snejus, line 391: I think the number of duplicate checks is getting a bit too confusing: now `keep_new` is checked 3 times in 3 separate conditional branches, while existence of `dest` is checked twice. 

I think we want to combine/nest these.

snejus (changes requested):

beetsplug/convert.py
pierreay: Indeed, my new commit `b96def17f` is removing all redundant `util.syspath` calls. ;)

beetsplug/convert.py
pierreay: If I understand correctly, you mean that we should not remove a destination file because of the `--refresh` option if the original file should *not* be transcoded because it will be symlinked or hardlinked (thus, removing the need of performing the `convert` operation another time because of an original file being modified). Is that right? It yes, I think you're right and this can be improved, while not being harmful, isn't it?

beetsplug/convert.py
pierreay: Indeed, this was a bit messy. My last commit `a956a10bc` try to address this issue.

beetsplug/convert.py
pierreay: Well, even if `should_transcode` is false, the file can be copied too (instead of linked). So that makes sense to remove `dest` such that the copy is done again when `refresh` is true from my understanding. What am I missing here?

pierreay:
Well, I just rebased on master anticipating a future merge, but I'm not sure the moment was right. ^^"

beetsplug/convert.py
snejus: I think this looks fine now!

snejus:
Now the only missing bit is a note in the changelog as this is a new feature :)

pierreay:
> Now the only missing bit is a note in the changelog as this is a new feature :)

@snejus Done! In fact the changelog entry was already there, but was lost in others entries of nearly one year ago since this is a pretty old PR ahah. :)

pierreay:
Rebased on master for `v2.3.0`. :) Anything else that I can do?

Serene-Arc:
Sorry I have been very busy and had to step back for a bit, but I'll try and review this today.

beetsplug/convert.py
```diff
@@ -382,12 +393,24 @@ def convert_item(
...
+                # Delete existing destination files when original files have
+                # been modified since the last convert run.
+                if refresh and os.path.getmtime(original) > os.path.getmtime(
+                    dest
+                ):
+                    self._log.info(
+                        "Removing {0} (original file modified)",
+                        util.displayable_path(dest),
+                    )
+                    if not pretend:
```
snejus: Thanks for tidying this up. It's much more clear now, where the logic is clear.

The only question I have left: what happens when `pretend=True`?

As far as I can see, the file will not be removed and the logic will continue on line 415. Before, the logic would never reach line 415 if the file existed. 

Is this intentional?

snejus (commented):
Apologies for the delay!

beetsplug/convert.py
```diff
@@ -382,12 +393,24 @@ def convert_item(
...
+                        util.remove(dest)
```
snejus: Also, I think we want to log this _only_ when `pretend` is disabled, otherwise this may scare me as a user 😅 

```suggestion
                    if not pretend:
                        self._log.info(
                            "Removing {0} (original file modified)",
                            util.displayable_path(dest),
                        )
                        util.remove(dest)

```

JOJ0:
Hi @pierreay, I'd like to ressurect this PR. There is a few suggestions from @snejus you might want to address, rebase against master and some polish and we should be good to go. I'll try to help move this forward. Thanks!

pierreay:
Hi @JOJ0 @snejus! Sorry for the delay, I've been really busy these days, unfortunately. I'll try to finish this PR properly by the next days. I did not forget, and I'm still using my own fork of Beets including my few PRs since the last nearly 2 years. :) See you soon!

beetsplug/convert.py
pierreay: Hi! Thanks for asking. Yep, this is totally intentional. Before, the logic would never reach 415 because we always wanted to exit if the `dest` file already exist -- this is exactly the reason that pushed me to create this PR.

Here, is `refresh` is `true` and the `original` file is newer that `dest` file, we will test for `pretend`. If `pretend` is `false`, we remove `dest`, and we jump to 415 with `dest` that do not exists, as previously to this PR. And if `pretend` is `true` (which is your question), we jump to 415 with the `dest` that is still existing. However ! Since `pretend` is true, all future lines will only log to inform the user what would be done -- without doing anything. Hence, to me, this is the intended behavior.

However, looking more closely at the code made my wondering about another potential (and last, I hope) issue. See my future comments below. :)

beetsplug/convert.py
pierreay: Ahah, I totally agree that would be scary to a user. However, I also think that this is in the interest of the user to see with `pretend` being `true` what would be actually done -- otherwise, `pretend` lost its interest. So, I suggest that:
- When `pretend=false`, we log something as `Removing {0} (original file modified)`
- When `pretend=true`, we log something as `Would remove {0} (original file modified)` or `Pretend to remove {0} (original file modified)` (any message that make it clear that this is only to inform the user of what would be done without the `--pretend` flag)

pierreay:
Hello @JOJ0 @snejus ! Sorry for the long delay since my last message, I'll strive to answer quickly as of now. Before rebasing on `master` and fixing logging message (the last questions of snejus), I would like to raise a final question about the usage of `refresh` and `keep_new` simultaneously.

# Question

- In the case of `keep_new` is `false` (the default), the use of `refresh` is clear: if we modified the library and the destination exist, we want to remove the destination and perform the conversion again (this is often the my use case, with my library being lossless and the destination being a lossy version of my library).
- In the case of `keep_new` is `true`, this is where things are unclear to me, because I'm not a user for this feature, hence, I do not know if a `refresh` feature would make sense here. Currently, in the way I implemented things (`convert.py`, lines 396 to 414), `original` timestamp is compared (with a `>`) with `dest`, which means that we are comparing the same files (only when `keep_new` is `true`). Therefore, the comparison will always lead to `false`, hence, the `refresh` branch (line 399) will never be taken and `convert.py` will just be behave as before this PR. 

While this behavior is harmless, the logic could be unclear for users and future developers. This confusion comes from the fact I am using the `original` variable, which depends on `keep_new`, and the `dest` variable, which do not depends on `keep_new` (I made a little diagram to help me with those variables, see attachment). Therefore, I suppose that a good logic would be to use both `item.path` AND `dest` (which do not depends on `keep_new`), or both `original` AND `converted` variables (which depends on `keep_new`). 

So finally, we have two potential scenarios here:
1. This do not make sense to have a `refresh` option when using `keep_new`, so `refresh` should do nothing when using `keep_new`. This is what is currently implemented (because we compare timestamps of `original > dest`), but this could be implemented in a clearer way. In addition, this would also not makes sense to remove `dest` (the file in the converted destination, theoretically, a backup) if `item.path` (the file in the library) is newer, because it will always be true if a previous conversion with `keep_new` has been issued, since converted files ended up in the library. Therefore, it do not makes sense to compare timestamps of `item.path > dest`.
2. This makes sense to remove `item.path` (the file in the library) if `dest` (the file in the converted destination) has been modified and is newer, when using both `refresh` and `keep_new`. Is this a use case for users using `keep_new`? I do not know, and I hope you may help me to decide here. I that case, we would need to compare timestamps of `original > converted`.

I hope this is clear, despite it is a long message. Sorry for not have been more concise. ^^'

# TL;DR

When using `keep_new`, should `refresh` do something, and is it the second scenario I presented, or simply `refresh` should do nothing?

# Diagram with variables from `convert.py`, filesystem side (library vs. convert destination), and scenarios (with or without `keep_new`)

image

snejus:
Hey @pierreay, thanks for the detailed write-up and the diagram!

I think option 1 is the right call here. `keep_new` is already a niche feature, and adding `refresh` behavior on top of it would be a niche-of-niche code path that nobody has asked for yet. Since you yourself don't use `keep_new` and aren't sure whether it makes sense - that's a strong signal to not build it.

The current behavior (silently no-op) is harmless, so I'd say just make it explicit with an early guard and maybe a debug log message, e.g.:

```py
if keep_new:    
    self._log.debug(                                                                                                                                                              
        "Skipping refresh: not supported with keep_new"
    )
```

If a real user comes along later and needs scenario 2, we can build it then with a concrete use case to guide the implementation.

Let's keep the scope tight and get this merged - the PR has been open long enough!

snejus:
@pierreay just pinging to check whether you're still happy to finish this up?

pierreay:
> @pierreay just pinging to check whether you're still happy to finish this up?

Hi @snejus, yes I am! April and May were really busy but I will complete this PR very soon in the next days. ;)

(Should be done by the 22nd of june)

pierreay:
- [x] Implement better logging
- [x] Implement better logic to handle `keep_new` and `refresh`
- [x] Rebase (lot of changes, take some time...)
- [x] Tests are passing
- [x] Final live test with a MWE of a database

<details><summary>Details</summary>
<p>
Should be finished by the 23
</p>
</details>

pierreay:
Hello @snejus @JOJ0 ! I have been able to fix and implement all the things we discussed above, and rebasing on `master`. :) I think this time this PR should be ready to merge! (at least, hope so!)

beetsplug/convert.py
```diff
@@ -400,7 +409,7 @@ def get_item_destination(self, item: Item) -> bytes:
         )
 
     @pipeline.mutator_stage
-    def convert_item(self, keep_new: bool, item: Item) -> None:
+    def convert_item(self, refresh: bool, keep_new: bool, item: Item) -> None:
```
snejus: Define a new `@cached_property` called `refresh` and read the value from the config there. (See the rest of variables)

snejus (changes requested):
Looks good, just make `refresh` a cached property and access it directly from the methods. I passed `keep_new` around due to another reason, so it has to be passed in.

pierreay:
> Looks good, just make `refresh` a cached property and access it directly from the methods. I passed `keep_new` around due to another reason, so it has to be passed in.

Nice catch, done! :)

snejus (approved):
Finally, this is ready and about to get merged! <3

PR #5210, opened 2024-04-26
Enforce the same interface across all `...Query` implementations

snejus:
### Make `PlaylistQuery` a `FieldQuery`
While working on the DB optimization and looking at updates upstream I discovered one query which does not follow the `FieldQuery` interface —`PlaylistQuery`, so I looked into it in more detail and ended up integrating it as a `FieldQuery`.

One special thing about it is that it uses **IN** SQL operator, so I added implementation for this sort of query outside the playlist context, see `InQuery`.

Otherwise, it seems like `PlaylistQuery` is a field query with a special way of resolving values it wants to query. In the future, we may want to consider moving this kind of custom _initialization_ logic away from `init` methods to factory/@classmethod: this should make it more clear that the purpose of such logic is to resolve the data that is required to define a particular `FieldQuery` class fully.

### Remove `NamedQuery` since it is unused

This simplifies query parsing logic in `queryparse.py`. We know that this logic can only receive `FieldQuery` classes thus I adjusted types and removed the logic that handles other cases.

Effectively, this means that the query parsing logic does not need to care whether the query is named by the corresponding DB field. Instead, queries like `SingletonQuery` and `PlaylistQuery` are initialized with the same data as others and take things from there themselves: in this case they translate `singleton` and `playlist` queries to the underlying DB filters.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -12,8 +12,7 @@
 # The above copyright notice and this permission notice shall be
 # included in all copies or substantial portions of the Software.
 
-"""The central Model and Database constructs for DBCore.
-"""
+"""The central Model and Database constructs for DBCore."""
 
 from __future__ import annotations
 
@@ -309,7 +308,7 @@ class Model(ABC):
     are subclasses of `Sort`.
     """
 
-    _queries: Dict[str, Type[Query]] = {}
+    _queries: Dict[str, Type[FieldQuery]] = {}
     """Named queries that use a field-like `name:value` syntax but which
     do not relate to any specific field.
     """
@@ -599,8 +598,7 @@ class Model(ABC):
             # Deleted flexible attributes.
             for key in self._dirty:
                 tx.mutate(
-                    "DELETE FROM {} "
-                    "WHERE entity_id=? AND key=?".format(self._flex_table),
+                    f"DELETE FROM {self._flex_table} WHERE entity_id=? AND key=?",
                     (self.id, key),
                 )
 
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -116,7 +116,7 @@ class Query(ABC):
 
 
 P = TypeVar("P")
-SQLiteType = Union[str, float, int, memoryview]
+SQLiteType = Union[str, bytes, float, int, memoryview]
 AnySQLiteType = TypeVar("AnySQLiteType", bound=SQLiteType)
 
 
@@ -416,7 +416,7 @@ class NumericQuery(FieldQuery):
                 return "1", ()
 
 
-class InQuery(FieldQuery[Sequence[AnySQLiteType]]):
+class InQuery(Generic[AnySQLiteType], FieldQuery[Sequence[AnySQLiteType]]):
     """Query which matches values in the given set."""
 
     field: str
@@ -424,10 +424,10 @@ class InQuery(FieldQuery[Sequence[AnySQLiteType]]):
     fast: bool = True
 
     @property
-    def subvals(self) -> Sequence[AnySQLiteType]:
+    def subvals(self) -> Sequence[SQLiteType]:
         return self.pattern
 
-    def col_clause(self) -> Tuple[str, Sequence[AnySQLiteType]]:
+    def col_clause(self) -> Tuple[str, Sequence[SQLiteType]]:
         placeholders = ", ".join(["?"] * len(self.subvals))
         return f"{self.field} IN ({placeholders})", self.subvals
 
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -115,15 +115,6 @@ class Query(ABC):
         return hash(type(self))
 
 
-class NamedQuery(Query):
-    """Non-field query, i.e. the query prefix is not a field but identifies the
-    query class.
-    """
-
-    @abstractmethod
-    def __init__(self, pattern): ...
-
-
 P = TypeVar("P")
 SQLiteType = Union[str, float, int, memoryview]
 AnySQLiteType = TypeVar("AnySQLiteType", bound=SQLiteType)
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -12,8 +12,7 @@
 # The above copyright notice and this permission notice shall be
 # included in all copies or substantial portions of the Software.
 
-"""The Query type hierarchy for DBCore.
-"""
+"""The Query type hierarchy for DBCore."""
 
 from __future__ import annotations
 
@@ -155,9 +154,7 @@ class FieldQuery(Query, Generic[P]):
 
     @classmethod
     def value_match(cls, pattern: P, value: Any):
-        """Determine whether the value matches the pattern. Both
-        arguments are strings.
-        """
+        """Determine whether the value matches the pattern."""
         raise NotImplementedError()
 
     def match(self, obj: Model) -> bool:
@@ -426,6 +423,28 @@ class NumericQuery(FieldQuery):
                 return f"{self.field} <= ?", (self.rangemax,)
             else:
                 return "1", ()
+
+
+class InQuery(FieldQuery[Sequence[AnySQLiteType]]):
+    """Query which matches values in the given set."""
+
+    field: str
+    pattern: Sequence[AnySQLiteType]
+    fast: bool = True
+
+    @property
+    def subvals(self) -> Sequence[AnySQLiteType]:
+        return self.pattern
+
+    def col_clause(self) -> Tuple[str, Sequence[AnySQLiteType]]:
+        placeholders = ", ".join(["?"] * len(self.subvals))
+        return f"{self.field} IN ({placeholders})", self.subvals
+
+    @classmethod
+    def value_match(
+        cls, pattern: Sequence[AnySQLiteType], value: AnySQLiteType
+    ) -> bool:
+        return value in pattern
 
 
 class CollectionQuery(Query):
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -427,7 +427,7 @@ def displayable_path(
         return path.decode("utf-8", "ignore")
 
 
-def syspath(path: bytes, prefix: bool = True) -> Bytes_or_String:
+def syspath(path: Bytes_or_String, prefix: bool = True) -> Bytes_or_String:
     """Convert a path for use by the operating system. In particular,
     paths on Windows must receive a magic prefix and must be converted
     to Unicode before they are sent to the OS. To disable the magic
--- a/beetsplug/playlist.py
+++ b/beetsplug/playlist.py
@@ -23,8 +23,12 @@ from beets.library import BLOB_TYPE
 from beets.util import path_as_posix
 
 
-class PlaylistQuery(InQuery):
+class PlaylistQuery(InQuery[bytes]):
     """Matches files listed by a playlist file."""
+
+    @property
+    def subvals(self) -> Sequence[BLOB_TYPE]:
+        return [BLOB_TYPE(p) for p in self.pattern]
 
     def __init__(self, _, pattern: str, __):
         config = beets.config["playlist"]
@@ -72,10 +76,6 @@ class PlaylistQuery(InQuery):
             f.close()
             break
         super().__init__("path", paths)
-
-    @property
-    def subvals(self) -> Sequence[BLOB_TYPE]:
-        return [BLOB_TYPE(p) for p in self.pattern]
 
 
 class PlaylistPlugin(beets.plugins.BeetsPlugin):
--- a/beetsplug/playlist.py
+++ b/beetsplug/playlist.py
@@ -15,17 +15,18 @@
 import fnmatch
 import os
 import tempfile
-from typing import Any, Optional, Sequence, Tuple
+from typing import Sequence
 
 import beets
+from beets.dbcore.query import InQuery
+from beets.library import BLOB_TYPE
 from beets.util import path_as_posix
 
 
-class PlaylistQuery(beets.dbcore.NamedQuery):
+class PlaylistQuery(InQuery):
     """Matches files listed by a playlist file."""
 
-    def __init__(self, pattern):
-        self.pattern = pattern
+    def __init__(self, _, pattern: str, __):
         config = beets.config["playlist"]
 
         # Get the full path to the playlist
@@ -39,7 +40,7 @@ class PlaylistQuery(beets.dbcore.NamedQuery):
             ),
         )
 
-        self.paths = []
+        paths = []
         for playlist_path in playlist_paths:
             if not fnmatch.fnmatch(playlist_path, "*.[mM]3[uU]"):
                 # This is not am M3U playlist, skip this candidate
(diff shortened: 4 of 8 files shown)


snejus:
> Did you check whether this refactoring still passes type checking (i.e. mypy)?

It's happy!

image

mypy runs live in my editor as I code where it's allowed to control my implementation :)

You've done an amazing job designing the generic types for `FieldQuery` class - you can see here how the pattern supported the new `InQuery` nicely out of the box.

beets/dbcore/db.py
```diff
@@ -308,7 +307,7 @@ class Model(ABC):
     are subclasses of `Sort`.
     """
 
-    _queries: Dict[str, Type[Query]] = {}
+    _queries: Dict[str, Type[FieldQuery]] = {}
```
wisp3rwind, line 310: I think this should remain unchanged, at least if we keep the `Query` base class around.

beets/dbcore/query.py
```diff
@@ -428,6 +416,28 @@ def col_clause(self) -> Tuple[str, Sequence[SQLiteType]]:
                 return "1", ()
 
 
+class InQuery(FieldQuery[Sequence[AnySQLiteType]]):
```
wisp3rwind, line 419:
```suggestion
class InQuery(FieldQuery[Sequence[Generic[AnySQLiteType]]]):
```

Making `InQuery` generic over the pattern type would allow subclasses to narrow this down (e.g. to `BLOB_TYPE a.k.a. memoryview` for `PlayListQuery`). (I didn't test this change, but I think this should work.)

beets/dbcore/query.py
```diff
@@ -428,6 +416,28 @@ def col_clause(self) -> Tuple[str, Sequence[SQLiteType]]:
...
+    """Query which matches values in the given set."""
+
+    field: str
+    pattern: Sequence[AnySQLiteType]
+    fast: bool = True
+
+    @property
+    def subvals(self) -> Sequence[AnySQLiteType]:
```
wisp3rwind, line 427: Why do we need `subvals`? Couldn't subclass constructors simply convert the `pattern` to the appropriate type and pass that to `InQuery.__init__`?

wisp3rwind (changes requested):
Oh, ok, so the point is that we don't really have any non-field query at this point. One could in principle imagine this: Maybe a query that involves several fields (although we also have the computed fields machinery, which might express that as well), or a plugin maintaining an extra table and defining a query on top of that. Thus, we don't need to accommodate for such queries at this point.

Looks good to me; I left a few notes where things might be improved or I'm not quite sure I understand why you coded things the way you did.

beets/dbcore/db.py
snejus: My thinking here was that `Query` was required in order to accommodate this one single `PlaylistQuery` that was not a `FieldQuery`. Seeing that all values in `Album._queries` and `Item._queries` now subclass a more specific `FieldQuery`, I aligned it accordingly.

Do you reckon it may be worth enforcing `FieldQuery` here, so that maintainers and plugin developers building on this have slightly more specific requirements/guidelines over how their queries should be implemented?

beets/dbcore/query.py
snejus: Completely! I started thinking about this but I couldn't see a quick and straightforward way of implementing this. Let me have a look what can be done here

beets/dbcore/query.py
snejus: Ah I didn't end up adding a docstring here - but yes you got the purpose of this right.

I saw two reasons for this:

1. Conversion to `memoryview` is only required because we use SQLite backend, and since this class has wider responsibilities than simply querying the database, it didn't seem right to couple its implementation to what's required by SQLite. This mostly comes from the Dependency inversion principle.

    Relatedly, storing items as a different type may break this method

    ```python
    @classmethod
    def value_match(
        cls, pattern: Sequence[AnySQLiteType], value: AnySQLiteType
    ) -> bool:
        return value in pattern
    ```

2. Regarding the implementation, in my head `__init__` is responsible for initialising the object, i.e. resolving the attributes that it needs, in this case the list of filenames to be stored as `pattern`. Given what I mentioned above, I see type conversion to `memoryview` as a separate task that only needs to happen when we're querying the database. Thus I moved it away from `__init__` to a separate method/attribute that has only this responsibility. I should have named it appropriately though!

beets/dbcore/query.py
wisp3rwind: I'm not totally sure, but I think what I suggested above might already be enough, since `AnySQLiteType` is a `TypeVar` with an appropriate bound. Then, it would simply become `class PlayListQuery(InQuery[BLOB_TYPE])`.

beets/dbcore/db.py
wisp3rwind: My thinking was that `db` only really uses the interface defined by `Query`, nothing added by `FieldQuery` on top.

So either, `Query` and `FieldQuery` remain distinct, in which case my comment from above applies.

If we decide that there's really nothing more basic than a `FieldQuery`, we should just remove one of them and simplify the class hierarchy. On first thought, I wouldn't do this, since conceptually there might well be different types of queries, and as far as I can tell, this doesn't really increase the complexity of our code very much.

Does this make sense? I might very well be missing something, it's been a while since I've looked at the `dbcore` code in any depth.

beets/dbcore/query.py
wisp3rwind: Ok, sounds good! I don't think it's absolutely necessary to have a docstring here (that'd be way more strict than what we have in many other places).

beets/dbcore/db.py
snejus: > My thinking was that db only really uses the interface defined by Query, nothing added by FieldQuery on top.

I haven't adjusted what the db expects here, for example see `Database._fetch` definition which has `Query`

```python
    def _fetch(
        self,
        model_cls: Type[AnyModel],
        query: Optional[Query] = None,
        sort: Optional[Sort] = None,
    ) -> Results[AnyModel]:
```

The change above adjusts `Model._queries`, which defines the type for `Item._queries` and `Album._queries`:

```python
class Item(LibModel):
    ...

    _queries = {"singleton": SingletonQuery}
```

This is also the dictionary that keeps named queries like `PlaylistQuery` once the plugins are parsed.

I have just attempted to change this to `_queries: Dict[str, Query]` and I found this made `mypy` upset in `queryparse.py`:
image

This is most likely why I adjusted it in the first place

beets/dbcore/query.py
snejus: Have tried it out but mypy has different plans 😅 
image

wisp3rwind: Oh, I think it needs to be
```py
class InQuery(Generic[AnySQLiteType], FieldQuery[Sequence[AnySQLiteType]]):
```
that seems to type-check for me locally, and in hindsight makes much more sense: `InQuery` is generic over `AnySQLiteType`, and then we immediately use that type variable again for naming the parent class.

beets/dbcore/db.py
wisp3rwind: I see the problem now: `db` does indeed not require this (anyway, it does not even access the `_queries` attribute), but `queryparse` uses the `db` default queries to initialize the `query_classes`.
So, I guess your change is fine. It still feels a little rough at the edges, because `Model._queries` used to be the non-field-queries, which are now all `FieldQuery` in disguise. I guess we can just keep these things as is, and figure out a better way if either non-field-queries need to return at some point, or if we decide to completely remove this distinction.

beets/dbcore/query.py
snejus: Thanks! This works fine now, see the latest commits

snejus:
@wisp3rwind you can see that I replaced `memoryview` with `bytes` in `PlaylistQuery` to query BLOB column `Item.path`.

Is there a reason we're using `BLOB_TYPE = memoryview` instead of `bytes`? I tested querying using `bytes` and I see to get the expected results

wisp3rwind:
> Is there a reason we're using `BLOB_TYPE = memoryview` instead of `bytes`? I tested querying using `bytes` and I see to get the expected results

https://github.com/beetbox/beets/blob/92fb830559530a4913b6966059c053fdd1e46bcf/beets/library.py#L41-L44

snejus:
> > Is there a reason we're using `BLOB_TYPE = memoryview` instead of `bytes`? I tested querying using `bytes` and I see to get the expected results
> 
> https://github.com/beetbox/beets/blob/92fb830559530a4913b6966059c053fdd1e46bcf/beets/library.py#L41-L44

Thanks!

This may be slightly outdated or am I missing something?

```sh
🐶 beet list label:ᴄʟᴇʀɢʏ -f '$path'
/run/media/sarunas/music/Music/🔥/03_daxj_stealthmode.flac
```

```python
[ins] In [17]: db = sqlite3.connect("/home/sarunas/.music/beets/library.db")

[ins] In [18]: res = list(db.execute("select path from items where path = ?", ("/run/media/sarunas/music/Music/🔥/03_daxj_stealthmode.flac".encode(),)))

[ins] In [19]: res
Out[19]: [(b'/run/media/sarunas/music/Music/\xf0\x9f\x94\xa5/03_daxj_stealthmode.flac',)]

[ins] In [20]: res[0][0].decode()
Out[20]: '/run/media/sarunas/music/Music/🔥/03_daxj_stealthmode.flac'
```

wisp3rwind:
> > > Is there a reason we're using `BLOB_TYPE = memoryview` instead of `bytes`? I tested querying using `bytes` and I see to get the expected results
> > 
> > 
> > https://github.com/beetbox/beets/blob/92fb830559530a4913b6966059c053fdd1e46bcf/beets/library.py#L41-L44
> 
> Thanks!
>
> [...]
> 
> This may be slightly outdated (this was introduced ~8 years ago) or am I missing something?

I really don't know. Is there any benefit to changing this? If at all possible, I'd prefer not to do so, at least not without a very good understanding of the original reasoning. Both to keep this PR focused, and because I'm afraid that this has the potential to break things in unexpected ways on different systems (with different OS, filesystem, terminal encodings).

Note that `memoryview` was specifically introduced for Python 3, so it doesn't look like a leftover of Python 2 times.

beets/dbcore/query.py
```diff
@@ -428,6 +416,24 @@ def col_clause(self) -> Tuple[str, Sequence[SQLiteType]]:
...
 
 
+class InQuery(Generic[AnySQLiteType], FieldQuery[Sequence[AnySQLiteType]]):
+    """Query which matches values in the given set."""
+
+    field: str
+    pattern: Sequence[AnySQLiteType]
+    fast: bool = True
+
+    def col_clause(self) -> Tuple[str, Sequence[SQLiteType]]:
```
wisp3rwind, line 426:
```suggestion
    def col_clause(self) -> Tuple[str, Sequence[AnySQLiteType]]:
```
This returns `self.pattern` which is `AnySQLiteType`, which is more specific than `SQLiteType`.

beets/util/__init__.py
```diff
@@ -427,7 +427,7 @@ def displayable_path(
         return path.decode("utf-8", "ignore")
 
 
-def syspath(path: bytes, prefix: bool = True) -> Bytes_or_String:
+def syspath(path: Bytes_or_String, prefix: bool = True) -> Bytes_or_String:
```
wisp3rwind, line 430: Did this sneak in unintendedly? I'm pretty sure that the original was correct, and if that fails to type-check, the error is probably with the caller.

beetsplug/playlist.py
```diff
@@ -15,17 +15,16 @@
...
 import tempfile
-from typing import Any, Optional, Sequence, Tuple
 
 import beets
+from beets.dbcore.query import InQuery
 from beets.util import path_as_posix
 
 
-class PlaylistQuery(beets.dbcore.NamedQuery):
+class PlaylistQuery(InQuery[bytes]):
```
wisp3rwind, line 24: See the `bytes` vs `memoryview` discussion.

wisp3rwind (changes requested):
Another, hopefully last, round of nitpicks :)

beets/util/__init__.py
snejus: That's intentional: since I now added types to `PlaylistQuery` implementation, `mypy` flagged an issue with this code

image

I had a look at `syspath` implementation and I found this situation

image

Given the above and that the function actually expects to receive `str` type (it converts `bytes` to `str`: `if not isinstance(path, str):` and below), I fixed the typing.

beetsplug/playlist.py
snejus: Added back the type mapping to `memoryview` now. Though `bytes` type stays - I found I made a mistake initially assuming it was `str` here.

beets/dbcore/query.py
snejus: I wasn't sure how to type the return type here, especially now that we're mapping the initial `bytes` sequence in `pattern` to a sequence of `memoryview`.

If I use `AnySQLiteType`, then it breaks since `memoryview` != `bytes`. Ideally, I'd want to define a type variable to make sure the return type from `subvals` is the same as here

```python
    if not isinstance(path, str):
    @property
    def subvals(self) -> Sequence[T]:
        return self.pattern

    def col_clause(self) -> Tuple[str, Sequence[T]]:
        placeholders = ", ".join(["?"] * len(self.subvals))
        return f"{self.field} IN ({placeholders})", self.subvals
```

But then trying to use a type variable here gave me this mess 😁 
image

snejus: Note that the interface defines this as `SQLiteType`

```python
class FieldQuery(Query, Generic[P]):
    """An abstract query that searches in a specific field for a
    pattern. Subclasses must provide a `value_match` class method, which
    determines whether a certain pattern string matches a certain value
    string. Subclasses may also provide `col_clause` to implement the
    same matching functionality in SQLite.
    """
    ...
    def col_clause(self) -> Tuple[Optional[str], Sequence[SQLiteType]]:
        return None, ()
```

beets/util/__init__.py
wisp3rwind: From my understanding, `syspath` doesn't really expect `str`, but this more of a defensive programming safeguard against incorrect usage. Almost always, it should be given `bytes`, which is the canonical internal representation for paths.

However, things seem to be more complicated than I originally thought: I'd argue that the playlist plugin should convert to bytes itself, but additionally, there is a different kind of usage such as in `normpath`. It might be that `syspath` is really conflating two purposes.

So, let's stick to your `Bytes_or_String` change for now, but I'm taking a mental note to revisit this at some point.

beets/dbcore/query.py
wisp3rwind: Ok, looks like I didn't check the `FieldQuery` interface properly myself. As you noted, it is conceivable that `pattern` and `col_clause` have distinct types. Let's keep it as is, if there's really the potential to do so, we can always tighten up our typing later on.

beetsplug/playlist.py
wisp3rwind: Oh, ok, so this is the same distinction as we discussed regarding the `[Any]SQLiteType` above: The input pattern type is `bytes`, but `col_clause` will turn that into `memoryview` for the SQLite API. It's also related to whether we do this type conversion in the constructor or only in `col_clause` which we discussed elsewhere.

wisp3rwind (approved):
Ok, looks good to me now. Thanks :+1:

beets/util/__init__.py
snejus: I found that the majority of typing issues across the codebase were found in this module, so I couldn't do much else but to address them, see #5215

PR #5214, opened 2024-04-30
Handle NULL values in the path column

snejus:
As I was devving, I did something wrong and had `beet mv` command fail on me.

Later, having spent an hour investigating why beets kept throwing me 'User-defined function raised exception' I discovered that it was failing because that previous `beet mv` command ended up writing value `NULL` in one of the items' `path` column. This was not handled well by the `BYTELOWER` implementation.

Since we do not have a **NOT NULL** constraint for the `path` column in the db, it's best to insure ourselves against this kind of stuff anyways.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -29,6 +29,7 @@ from sqlite3 import Connection
 from types import TracebackType
 from typing import (
     Any,
+    AnyStr,
     Callable,
     DefaultDict,
     Dict,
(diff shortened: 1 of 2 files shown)


wisp3rwind:
lgtm :tada:

wisp3rwind (approved):

Serene-Arc:
...wth why did the changelog bot suddenly work for this PR and not any other?

wisp3rwind:
I guess because you pushed this branch to `beetbox/` rather than `snejus/`. So, actually, there seems to be some permission error, the docs of the github action mention something along those lines.

snejus:
@Serene-Arc Could I by any chance get access to the repo settings? I've had this issue before and fixed it, but now I don't remember which specific settings were to blame, so I'd be happy just to have a look at what we've currently got.

Serene-Arc:
It is a permission error, 100%, it's in the PR for the change I think. I don't have the access required to change it, but that makes sense as to why it works for this.

So @snejus no, because I don't have access either. You'll have to as @sampsyo for those permissions (and I wouldn't mind getting them either if it's not a security risk).

snejus:
Cool, will drop him a message :)

PR #5217, opened 2024-05-03
Draft: importer: Add ability to verify copy and move operations

oliv3r:
To avoid corrupting files and worse, remove the original files (move), a quick check is added to ensure files are still in order.

On failure the crc is printed to aid in discovering if the problem was with the source (unlikely) or destination file. Using `filecmp` would only tell us of a miss-match, but making it harder to figure out why/where things fail.

Since `beets import` can be destructive (move), we want to be delicate here.

## Description

Fixes #X.  

(...)

## To Do


wisp3rwind:
> But also being thrown about with `NameError: name 'import_cmd_parser' is not defined` both when running `tox -e int` but also just trying to run the application ...

It's a typo: `import_cmd.parser` vs. `import_cmd_parser`.

oliv3r:
I taken it it's fixed on master so I van just rebase? :)

wisp3rwind:
> I taken it it's fixed on master so I van just rebase? :)

No, you made a typo when adding the new verify command, which is why the tests fail.

snejus:

PR #5220, opened 2024-05-04
Improve GitHub build output and configure coverage

snejus:
### `pytest`

- Ensure that test summary includes the list of failed tests (which is missing right now).

### `isort` and `black`

- Show which files are failing and include the diff of what sort of fix needs applying.
  Currently, neither the files nor the diff is available.

### Coverage

Add `pytest-cov` which automatically measures coverage while tests are executing and update `tox` and GitHub CI configuration accordingly.

#### Configuration

1. Enable analysing logical branches
2. Save coverage details as HTML for local development use
3. Configure the HTML output to include coverage context: on the right-hand side of each covered code line there is an expandable section which lists every test that runs/covers this line. See below 
image


setup.py
```diff
@@ -108,6 +108,7 @@ def build_manpages():
             "mock",
             "pylast",
             "pytest",
+            "pytest-cov",
```
wisp3rwind, line 111: We can now remove the `coverage` dependency above, right?

tox.ini
```diff
@@ -25,8 +25,7 @@ deps =
     mypy: {[_mypy]deps}
 passenv = INTEGRATION_TEST
 commands =
-    test: python -bb -m pytest -rs {posargs}
```
wisp3rwind, line 28: The `-rs` is now handled by the options in `setup.cfg`, right?

wisp3rwind (commented):
Thanks! I left two small questions.

Have you tested whether the diff reporting works as expected (i.e. just force-push some intentionally broken code to this PR)?

tox.ini
snejus: Indeed. `-rs` only selected to show the SKIPPED tests but ignored the rest. `-ra` shows everything except passed
image

snejus:
Failed isort check
image

snejus:
> Thanks! I left two small questions.
> 
> Have you tested whether the diff reporting works as expected (i.e. just force-push some intentionally broken code to this PR)?

Good point, forgot to do it!

snejus:
I've noticed that the coverage [is not getting uploaded](https://github.com/beetbox/beets/actions/runs/8946776110/job/24577880217). I wonder whether this has to do with the fact that this PR is coming from the fork, and potentially `CODECOV_TOKEN` is not available in the build?

I generated coverage for the last couple of commits on master and uploaded it upstream, so you can have a glance [here](https://app.codecov.io/github/beetbox/beets/tree/master?trend=6%20months&displayType=list)

snejus:
That's the output from failed `black` check 
image

snejus:
And finally, failed tests are now appearing at the end of the output 
image

wisp3rwind:
> I've noticed that the coverage [is not getting uploaded](https://github.com/beetbox/beets/actions/runs/8946776110/job/24577880217). I wonder whether this has to do with the fact that this PR is coming from the fork, and potentially `CODECOV_TOKEN` is not available in the build?

Not sure, I don't have access to the repo settings, so can't check whether the token is in fact set. In any case, according to https://docs.codecov.com/docs/adding-the-codecov-token#github-actions, we would need to explicitly use the token in the CI config.

Maybe let's merge this PR, check whether things work, and if required follow up with a small fixup PR?

In any case: Looks good to me, and thanks again! (In fact, I remember adding the failed tests output back when we used nosetest, but it seems that this got lost in the transition to pytest.)

wisp3rwind:
This breaks local tox runs for me, since pytest seems to not recognize the `--cov...` arguments. @snejus does this work for you?

arogl:
> This breaks local tox runs for me, since pytest seems to not recognize the `--cov...` arguments. @snejus does this work for you?

I had to locally install `pip install pytest-cov`. 

@wisp3rwind rerun setup for test @snejus added it to the config.

snejus:
You can run `pip install -e .[test]` in your virtual environment

PR #5223, opened 2024-05-04
Fix most types in beets.util.__init__

snejus:
## Description

This PR is Part 1 of the work that fixes typing issues in `beets.util.__init__` module. 

It addresses simple-to-fix / most of the issues in this module.

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -24,12 +24,13 @@ import shlex
 import shutil
 import subprocess
 import sys
-import tempfile
 import traceback
 from collections import Counter, namedtuple
+from contextlib import suppress
 from enum import Enum
 from logging import Logger
 from multiprocessing.pool import ThreadPool
+from tempfile import NamedTemporaryFile
 from typing import (
     Any,
     AnyStr,
@@ -54,7 +55,7 @@ from beets.util import hidden
 MAX_FILENAME_LENGTH = 200
 WINDOWS_MAGIC_PREFIX = "\\\\?\\"
 T = TypeVar("T")
-Bytes_or_String: TypeAlias = Union[str, bytes]
+BytesOrStr: TypeAlias = Union[str, bytes]
 
 
 class HumanReadableException(Exception):
@@ -154,7 +155,7 @@ class MoveOperation(Enum):
     REFLINK_AUTO = 5
 
 
-def normpath(path: bytes) -> bytes:
+def normpath(path: AnyStr) -> bytes:
     """Provide the canonical form of the path suitable for storing in
     the database.
     """
@@ -163,7 +164,7 @@ def normpath(path: bytes) -> bytes:
     return bytestring_path(path)
 
 
-def ancestry(path: bytes) -> List[str]:
+def ancestry(path: AnyStr) -> List[AnyStr]:
     """Return a list consisting of path's parent directory, its
     grandparent, and so on. For instance:
 
@@ -172,7 +173,7 @@ def ancestry(path: bytes) -> List[str]:
 
     The argument should *not* be the result of a call to `syspath`.
     """
-    out = []
+    out: List[AnyStr] = []
     last_path = None
     while path:
         path = os.path.dirname(path)
@@ -188,28 +189,28 @@ def ancestry(path: bytes) -> List[str]:
 
 
 def sorted_walk(
-    path: AnyStr,
-    ignore: Sequence = (),
+    path: BytesOrStr,
+    ignore: Sequence[BytesOrStr] = (),
     ignore_hidden: bool = False,
     logger: Optional[Logger] = None,
-) -> Generator[Tuple, None, None]:
+) -> Generator[Tuple[bytes, Sequence[bytes], Sequence[bytes]], None, None]:
     """Like `os.walk`, but yields things in case-insensitive sorted,
     breadth-first order.  Directory and file names matching any glob
     pattern in `ignore` are skipped. If `logger` is provided, then
     warning messages are logged there when a directory cannot be listed.
     """
     # Make sure the paths aren't Unicode strings.
-    path = bytestring_path(path)
-    ignore = [bytestring_path(i) for i in ignore]
+    bytes_path = bytestring_path(path)
+    bytes_ignore = [bytestring_path(i) for i in ignore]
 
     # Get all the directories and files at this level.
     try:
-        contents = os.listdir(syspath(path))
+        contents = os.listdir(syspath(bytes_path))
     except OSError as exc:
         if logger:
             logger.warning(
                 "could not list directory {}: {}".format(
-                    displayable_path(path), exc.strerror
+                    displayable_path(bytes_path), exc.strerror
                 )
             )
         return
@@ -220,11 +221,11 @@ def sorted_walk(
 
         # Skip ignored filenames.
         skip = False
-        for pat in ignore:
+        for pat in bytes_ignore:
             if fnmatch.fnmatch(base, pat):
                 if logger:
-                    logger.debug(
-                        "ignoring {} due to ignore rule {}".format(base, pat)
+                    logger.error(
+                        "ignoring '{}' due to ignore rule '{}'", base, pat
                     )
                 skip = True
                 break
@@ -232,7 +233,7 @@ def sorted_walk(
             continue
 
         # Add to output as either a file or a directory.
-        cur = os.path.join(path, base)
+        cur = os.path.join(bytes_path, base)
         if (ignore_hidden and not hidden.is_hidden(cur)) or not ignore_hidden:
             if os.path.isdir(syspath(cur)):
                 dirs.append(base)
@@ -242,11 +243,11 @@ def sorted_walk(
     # Sort lists (case-insensitive) and yield the current level.
     dirs.sort(key=bytes.lower)
     files.sort(key=bytes.lower)
-    yield (path, dirs, files)
+    yield (bytes_path, dirs, files)
 
     # Recurse into directories.
     for base in dirs:
-        cur = os.path.join(path, base)
+        cur = os.path.join(bytes_path, base)
         # yield from sorted_walk(...)
         yield from sorted_walk(cur, ignore, ignore_hidden, logger)
 
@@ -289,7 +290,7 @@ def fnmatch_all(names: Sequence[bytes], patterns: Sequence[bytes]) -> bool:
 
 def prune_dirs(
     path: str,
-    root: Optional[Bytes_or_String] = None,
+    root: Optional[AnyStr] = None,
     clutter: Sequence[str] = (".DS_Store", "Thumbs.db"),
 ):
     """If path is an empty directory, then remove it. Recursively remove
@@ -299,33 +300,33 @@ def prune_dirs(
     emptiness. If root is not provided, then only path may be removed
     (i.e., no recursive removal).
     """
-    path = normpath(path)
-    if root is not None:
-        root = normpath(root)
+    bytes_path = normpath(path)
+    root_bytes = normpath(root) if root else None
+    ancestors = ancestry(bytes_path)
 
-    ancestors = ancestry(path)
     if root is None:
         # Only remove the top directory.
         ancestors = []
-    elif root in ancestors:
-        # Only remove directories below the root.
-        ancestors = ancestors[ancestors.index(root) + 1 :]
+    elif root_bytes in ancestors:
+        # Only remove directories below the root_bytes.
+        ancestors = ancestors[ancestors.index(root_bytes) + 1 :]
     else:
         # Remove nothing.
         return
 
+    bytes_clutter = [bytestring_path(c) for c in clutter]
+
     # Traverse upward from path.
-    ancestors.append(path)
+    ancestors.append(bytes_path)
     ancestors.reverse()
     for directory in ancestors:
         directory = syspath(directory)
         if not os.path.exists(directory):
             # Directory gone already.
             continue
-        clutter: List[bytes] = [bytestring_path(c) for c in clutter]
         match_paths = [bytestring_path(d) for d in os.listdir(directory)]
         try:
-            if fnmatch_all(match_paths, clutter):
+            if fnmatch_all(match_paths, bytes_clutter):
                 # Directory contains only clutter (or nothing).
                 shutil.rmtree(directory)
             else:
@@ -380,7 +381,7 @@ def _fsencoding() -> str:
     return encoding
 
 
-def bytestring_path(path: Bytes_or_String) -> bytes:
+def bytestring_path(path: BytesOrStr) -> bytes:
     """Given a path, which is either a bytes or a unicode, returns a str
     path (ensuring that we never deal with Unicode pathnames). Path should be
     bytes but has safeguards for strings to be converted.
@@ -427,7 +428,7 @@ def displayable_path(
         return path.decode("utf-8", "ignore")
 
 
-def syspath(path: Bytes_or_String, prefix: bool = True) -> Bytes_or_String:
+def syspath(path: AnyStr, prefix: bool = True) -> AnyStr:
     """Convert a path for use by the operating system. In particular,
     paths on Windows must receive a magic prefix and must be converted
     to Unicode before they are sent to the OS. To disable the magic
@@ -438,51 +439,57 @@ def syspath(path: Bytes_or_String, prefix: bool = True) -> Bytes_or_String:
     if os.path.__name__ != "ntpath":
         return path
 
-    if not isinstance(path, str):
+    if isinstance(path, bytes):
         # Beets currently represents Windows paths internally with UTF-8
         # arbitrarily. But earlier versions used MBCS because it is
         # reported as the FS encoding by Windows. Try both.
         try:
-            path = path.decode("utf-8")
+            str_path = path.decode("utf-8")
         except UnicodeError:
             # The encoding should always be MBCS, Windows' broken
             # Unicode representation.
             assert isinstance(path, bytes)
             encoding = sys.getfilesystemencoding() or sys.getdefaultencoding()
-            path = path.decode(encoding, "replace")
+            str_path = path.decode(encoding, "replace")
+    else:
+        str_path = path
 
     # Add the magic prefix if it isn't already there.
     # https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247.aspx
-    if prefix and not path.startswith(WINDOWS_MAGIC_PREFIX):
-        if path.startswith("\\\\"):
+    if prefix and not str_path.startswith(WINDOWS_MAGIC_PREFIX):
+        if str_path.startswith("\\\\"):
             # UNC path. Final path should look like \\?\UNC\...
-            path = "UNC" + path[1:]
-        path = WINDOWS_MAGIC_PREFIX + path
+            str_path = "UNC" + str_path[1:]
+        str_path = WINDOWS_MAGIC_PREFIX + str_path
 
-    return path
+    return str_path if isinstance(path, str) else str_path.encode()
 
 
-def samefile(p1: bytes, p2: bytes) -> bool:
+def samefile(p1: AnyStr, p2: AnyStr) -> bool:
     """Safer equality for paths."""
     if p1 == p2:
         return True
-    return shutil._samefile(syspath(p1), syspath(p2))
+    with suppress(OSError):
+        return os.path.samefile(syspath(p1), syspath(p2))
+
+    return False
 
 
-def remove(path: Optional[bytes], soft: bool = True):
+def remove(path: AnyStr, soft: bool = True):
     """Remove the file. If `soft`, then no error will be raised if the
     file does not exist.
     """
-    path = syspath(path)
     if not path or (soft and not os.path.exists(path)):
         return
+
+    path = syspath(path)
     try:
         os.remove(path)
     except OSError as exc:
         raise FilesystemError(exc, "delete", (path,), traceback.format_exc())
 
 
-def copy(path: bytes, dest: bytes, replace: bool = False):
+def copy(path: AnyStr, dest: AnyStr, replace: bool = False):
     """Copy a plain file. Permissions are not copied. If `dest` already
     exists, raises a FilesystemError unless `replace` is True. Has no
     effect if `path` is the same as `dest`. Paths are translated to
@@ -524,30 +531,31 @@ def move(path: bytes, dest: bytes, replace: bool = False):
         # Copy the file to a temporary destination.
         basename = os.path.basename(bytestring_path(dest))
         dirname = os.path.dirname(bytestring_path(dest))
-        tmp = tempfile.NamedTemporaryFile(
+        tmp = NamedTemporaryFile(
             suffix=syspath(b".beets", prefix=False),
             prefix=syspath(b"." + basename, prefix=False),
             dir=syspath(dirname),
             delete=False,
         )
         try:
-            with open(syspath(path), "rb") as f:
+            with open(syspath(path)) as f:
                 shutil.copyfileobj(f, tmp)
         finally:
             tmp.close()
 
         # Move the copied file into place.
+        tmp_filename = tmp.name
         try:
-            os.replace(tmp.name, syspath(dest))
-            tmp = None
+            os.replace(tmp_filename, syspath(dest))
+            tmp_filename = ""
             os.remove(syspath(path))
         except OSError as exc:
             raise FilesystemError(
                 exc, "move", (path, dest), traceback.format_exc()
             )
         finally:
-            if tmp is not None:
-                os.remove(tmp)
+            if tmp_filename:
+                os.remove(tmp_filename)
 
 
 def link(path: bytes, dest: bytes, replace: bool = False):
@@ -669,7 +677,7 @@ def unique_path(path: bytes) -> bytes:
 # Unix. They are forbidden here because they cause problems on Samba
 # shares, which are sufficiently common as to cause frequent problems.
 # https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247.aspx
-CHAR_REPLACE: List[Tuple[Pattern, str]] = [
+CHAR_REPLACE: List[Tuple[Pattern[str], str]] = [
     (re.compile(r"[\\/]"), "_"),  # / and \ -- forbidden everywhere.
     (re.compile(r"^\."), "_"),  # Leading dot (hidden files on Unix).
     (re.compile(r"[\x00-\x1f]"), ""),  # Control characters.
@@ -681,7 +689,7 @@ CHAR_REPLACE: List[Tuple[Pattern, str]] = [
 
 def sanitize_path(
     path: str,
-    replacements: Optional[Sequence[Sequence[Union[Pattern, str]]]] = None,
+    replacements: Optional[Sequence[Tuple[Pattern[str], str]]] = None,
 ) -> str:
     """Takes a path (as a Unicode string) and makes sure that it is
     legal. Returns a new path. Only works with fragments; won't work
@@ -722,11 +730,11 @@ def truncate_path(path: AnyStr, length: int = MAX_FILENAME_LENGTH) -> AnyStr:
 
 def _legalize_stage(
     path: str,
-    replacements: Optional[Sequence[Sequence[Union[Pattern, str]]]],
+    replacements: Optional[Sequence[Tuple[Pattern[str], str]]],
     length: int,
     extension: str,
     fragment: bool,
-) -> Tuple[Bytes_or_String, bool]:
+) -> Tuple[BytesOrStr, bool]:
     """Perform a single round of path legalization steps
     (sanitation/replacement, encoding from Unicode to bytes,
     extension-appending, and truncation). Return the path (Unicode if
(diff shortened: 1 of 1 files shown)


beets/util/__init__.py
```diff
@@ -438,51 +439,57 @@ def syspath(path: Bytes_or_String, prefix: bool = True) -> Bytes_or_String:
...
-def samefile(p1: bytes, p2: bytes) -> bool:
+def samefile(p1: AnyStr, p2: AnyStr) -> bool:
     """Safer equality for paths."""
     if p1 == p2:
         return True
-    return shutil._samefile(syspath(p1), syspath(p2))
+    with suppress(OSError):
+        return os.path.samefile(syspath(p1), syspath(p2))
+
+    return False
```
wisp3rwind, line 475:
```suggestion
   return shutil._samefile(syspath(p1), syspath(p2))  # 2)) # type: ignore[attr-defined]
```

Did you check why we're using the internal `shutil` function here in the first place? Unless we properly understand this, let's not change the code, but instead ignore the typing error.

beets/util/__init__.py
```diff
@@ -438,51 +439,57 @@ def syspath(path: Bytes_or_String, prefix: bool = True) -> Bytes_or_String:
...
 
 
-def remove(path: Optional[bytes], soft: bool = True):
+def remove(path: AnyStr, soft: bool = True):
```
wisp3rwind, line 478:
```suggestion
def remove(path: bytes, soft: bool = True):
```

beets/util/__init__.py
```diff
@@ -438,51 +439,57 @@ def syspath(path: Bytes_or_String, prefix: bool = True) -> Bytes_or_String:
...
     """Remove the file. If `soft`, then no error will be raised if the
     file does not exist.
     """
-    path = syspath(path)
```
wisp3rwind, line 476: This was important, `os.path.exists` needs to be called with the `syspath` transformation.

beets/util/__init__.py
```diff
@@ -438,51 +439,57 @@ def syspath(path: Bytes_or_String, prefix: bool = True) -> Bytes_or_String:
...
+
+    path = syspath(path)
     try:
         os.remove(path)
     except OSError as exc:
         raise FilesystemError(exc, "delete", (path,), traceback.format_exc())
 
 
-def copy(path: bytes, dest: bytes, replace: bool = False):
+def copy(path: AnyStr, dest: AnyStr, replace: bool = False):
```
wisp3rwind, line 492:
```suggestion
def copy(path: bytes, dest: bytes, replace: bool = False):
```

beets/util/__init__.py
```diff
@@ -524,30 +531,31 @@ def move(path: bytes, dest: bytes, replace: bool = False):
...
+        tmp = NamedTemporaryFile(
             suffix=syspath(b".beets", prefix=False),
             prefix=syspath(b"." + basename, prefix=False),
             dir=syspath(dirname),
             delete=False,
         )
         try:
-            with open(syspath(path), "rb") as f:
+            with open(syspath(path)) as f:
                 shutil.copyfileobj(f, tmp)
```
wisp3rwind, line 542:
```suggestion
            with open(syspath(path), "rb") as f:
                # mypy bug:
                # - https://github.com/python/mypy/issues/15031
                # - https://github.com/python/mypy/issues/14943
                # Fix not yet released:
                # - https://github.com/python/mypy/pull/14975
                shutil.copyfileobj(f, tmp)  # type: ignore[misc]
```
Why remove the "rb"? I had another issue with mypy at this place, see the suggestion.

beets/util/__init__.py
```diff
@@ -524,30 +531,31 @@ def move(path: bytes, dest: bytes, replace: bool = False):
...
         finally:
             tmp.close()
 
         # Move the copied file into place.
+        tmp_filename = tmp.name
         try:
-            os.replace(tmp.name, syspath(dest))
-            tmp = None
+            os.replace(tmp_filename, syspath(dest))
+            tmp_filename = ""
```
wisp3rwind, line 550: nice workaround! I had dealt with this differently, introducing an extra variable, but yours is more concise.

beets/util/__init__.py
```diff
@@ -1094,7 +1102,7 @@ def asciify_path(path: str, sep_replace: str) -> str:
     return os.sep.join(path_components)
 
 
-def par_map(transform: Callable, items: Iterable):
+def par_map(transform: Callable[..., Any], items: Iterable[Any]) -> None:
```
wisp3rwind, line 1105:
```suggestion
def par_map(transform: Callable[[T], None], items: Iterable[T]) -> None:
```
We can make this even more specific.

beetsplug/web/__init__.py
```diff
@@ -327,7 +327,7 @@ def item_file(item_id):
 
     try:
         # Imitate http.server behaviour
-        base_filename.encode("latin-1", "strict")
+        base_filename.decode().encode("latin-1", "strict")
```
wisp3rwind, line 330: Could you explain this change? I think `base_filename` should be `str` here, but mypy might not be able to infer this. In any case, I don't understand what this code does at all, it might need a proper cleanup.

wisp3rwind (changes requested):
I just added a bunch of comments: Many of them are concerned with the tension between what types of arguments functions would accept vs. what we actually want to call them with (i.e. often it should only by `bytes` that are accepted, but the code does not obviously fail for `str`).

In other cases, it's a bit unclear why a specific change was made; it would be great if you could elaborate.

beets/util/__init__.py
```diff
@@ -220,19 +221,19 @@ def sorted_walk(
 
         # Skip ignored filenames.
         skip = False
-        for pat in ignore:
+        for pat in bytes_ignore:
             if fnmatch.fnmatch(base, pat):
                 if logger:
-                    logger.debug(
-                        "ignoring {} due to ignore rule {}".format(base, pat)
+                    logger.error(
```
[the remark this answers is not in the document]
snejus: Completely. I used `error` to test it, but forgot to change it back!

beets/util/__init__.py
```diff
@@ -188,28 +189,28 @@ def ancestry(path: bytes) -> List[str]:
 
 
 def sorted_walk(
-    path: AnyStr,
-    ignore: Sequence = (),
+    path: BytesOrStr,
```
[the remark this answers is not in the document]
snejus: I was wondering why should paths be represented as `bytes` internally? This requirement may be less relevant now that python 2 is dropped and not supported anymore, see https://beets.io/blog/paths.html

> This way, Python 3’s Unicode [str](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str) can represent arbitrary bytes in filenames. (The first commit to beets happened a bit before Python 3.0 was released, so perhaps the project can be forgiven for not adopting this approach in the first place.)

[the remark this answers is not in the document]
snejus: Ideally, I'd imagine these functions should accept the same types that the original ones do. In this case it's `os.walk`, which accepts any path-like `str` or `bytes` and returns _the same type_ that it received.

On the other hand, this PR was about fixing types rather than adjusting the implementation, so I focused on underlining what the current implement does and handles to make life easier to the person who will need to refactor this module later.

beets/util/__init__.py
```diff
@@ -289,7 +290,7 @@ def fnmatch_all(names: Sequence[bytes], patterns: Sequence[bytes]) -> bool:
 
 def prune_dirs(
     path: str,
-    root: Optional[Bytes_or_String] = None,
+    root: Optional[AnyStr] = None,
```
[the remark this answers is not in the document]
snejus: See my comments above: in summary, I think we'd rather use types that reflect the current implementation. In addition to the two reasons I've given above, I see another risk: if we were to restrict the accepted type to `bytes` here, in a fully-typed codebase we'd expect `mypy` check to fail anywhere where a `str` is being fed into this function.

Right now, much the codebase is not typed, so the check passes silently. Whoever ends up adding types to the part of the codebase that sends `str` into this function will get slightly bamboozled: they will need to read this code and understand what it does and they will have to either 

1. Adjust the implementation of that code
2. Relax the type here

Both of these actions will be out of scope of what they're doing. Remember, I had to do this in `playlist.py` when I ended up relaxing the type in `syspath`?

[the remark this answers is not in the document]
snejus: I'd be happy with making these functions to handle `bytes` only if

1. We explicitly handle `bytes` only - remove the code that casts `str` to `bytes` from the beginning of the implementation
2. Rename the functions to make the _only-bytes-accepted-here_ requirement clear, since the equivalents in `os` (`normpath`, `samefile`, ...) handle `str` and `bytes` types. Therefore, to not confuse people used to the builtins, we'd want to name these functions accordingly I suppose.

beets/util/__init__.py
snejus: I did - `samefile` was not supported on Windows until Python 3.2, see [here](https://docs.python.org/3/library/os.path.html#os.path.samefile)


> os.path.samefile(path1, path2)
Return True if both pathname arguments refer to the same file or directory. This is determined by the device number and i-node number and raises an exception if an [os.stat()](https://docs.python.org/3/library/os.html#os.stat) call on either pathname fails.
Changed in version 3.2: Added Windows support.
Changed in version 3.4: Windows now uses the same implementation as all other platforms.
Changed in version 3.6: Accepts a [path-like object](https://docs.python.org/3/glossary.html#term-path-like-object).

beets/util/__init__.py
snejus: Ah good point, thanks!

beets/util/__init__.py
snejus: Whoa, well caught! I changed precisely because `mypy` check was failing. I also didn't realise that `NamedTemporaryFile` opens the file in bytes mode by default. Just checked my change would actually make this fail.

I wonder why this isn't tested?

beetsplug/web/__init__.py
snejus: This has to do with the change I made in `syspath` 
image

You can see that it now **always** returns the same type that was fed into it. 

```python
    return str_path if isinstance(path, str) else str_path.encode()
```

It previously returned undeterministic `BytesOrStr`: for non-windows systems it would return the same type that was fed in, but on Windows it would always return `str`. This caused a fair bit of typing issues down the line.

Now the code in `web/__init__.py` is a bit of a mess

```python
    # On Windows under Python 2, Flask wants a Unicode path. On Python 3, it
    # *always* wants a Unicode path.
    if os.name == "nt":
        item_path = util.syspath(item.path)
    else:
        item_path = util.py3_path(item.path)

    base_filename = os.path.basename(item_path)
    # FIXME: Arguably, this should just use `displayable_path`: The latter
    # tries `_fsencoding()` first, but then falls back to `utf-8`, too.
    if isinstance(base_filename, bytes):
        try:
            unicode_base_filename = base_filename.decode("utf-8")
        except UnicodeError:
            unicode_base_filename = util.displayable_path(base_filename)
    else:
        unicode_base_filename = base_filename

    try:
        # Imitate http.server behaviour
        base_filename.decode().encode("latin-1", "strict")
```

You can see it calls `syspath` on Windows, which now returns `bytes` instead of `str`. It seemingly expects `base_filename` to have type `bytes` and assigns `unicode_base_filename`, but then ends up using `base_filename` in the last line here, assuming that it's a `str` but it is still `bytes`.

snejus: Given this comment

```python
    # On Windows under Python 2, Flask wants a Unicode path. On Python 3, it
    # *always* wants a Unicode path.
    if os.name == "nt":
        item_path = util.syspath(item.path)
    else:
        item_path = util.py3_path(item.path)
```

I think it should be safe to simply call `py3_path` in all cases now.

wisp3rwind: I agree that `syspath` is a tricky regarding typing — but this is in fact a change which I think has the potential to cause a big fallout if we get it wrong, and at this point I'm not fully convinced that it is correct.

Up to now, `syspath` is intended to return `str` on Windows, because all Windows filesystems and APIs use some Unicode encoding and the previous default `mbcs` encoding (which I think would have been used when passing `bytes` to OS APIs) would in fact have been incorrect, given that beets internally uses UTF-8 as stated in the code comment.

Due to https://peps.python.org/pep-0529/ your change might in fact be correct now (note that there's a legacy mode that uses MBCS, even though I'd say users who explicitly enable that are on their own).

I'd prefer not to change the behaviour for now (at least not in this PR, let's keep the code changes to a minimum when the primary goal is introducing typings). I'm not confident that our testsuite exercises non-ASCII filenames sufficiently to catch all encoding edge cases.

For now, another solution might be to define types conditionally (mypy understands the `sys.platform` switch IIRC):

```python
if sys.platform == "win32":
    SysPath: TypeAlias = str
else:
    SysPath: TypeAlias = bytes
def syspath(path: bytes, prefix: bool = True) -> SysPath:
```
(we should probably also run mypy on Windows then).


---

Regarding the `web` plugin code: If I'm not completely mistaken, `base_filename` is guaranteed to be `str` when `decode` is invoked (irrespective of the `syspath` change, in fact), but only `bytes` can be `decode()`ed.

Actually, it looks like the whole section
```python
    base_filename = os.path.basename(item_path)
    # FIXME: Arguably, this should just use `displayable_path`: The latter
    # tries `_fsencoding()` first, but then falls back to `utf-8`, too.
    if isinstance(base_filename, bytes):
        try:
            unicode_base_filename = base_filename.decode("utf-8")
        except UnicodeError:
            unicode_base_filename = util.displayable_path(base_filename)
    else:
        unicode_base_filename = base_filename
```
is a no-op nowadays (always hitting the `else` arm. (It's not hit in tests at least: https://app.codecov.io/gh/beetbox/beets/blob/master/beetsplug%2Fweb%2F__init__.py but note that there, also the `UnicodeError` never happens, i.e. this is a place where we don't seem to actually test non-ASCII (actually, non-latin1) filenames.)

beets/util/__init__.py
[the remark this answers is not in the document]
wisp3rwind: > I was wondering why should paths be represented as bytes internally? This requirement may be less relevant now that python 2 is dropped and not supported anymore, see https://beets.io/blog/paths.html

Yes, that could be changed in principle, but that's in fact not completely straightforward: We still need conversions, at a minimum when dealing with the database (which currently stores bytes, not text. Would SQLite even support storing text with surrogate escapes?).
At that point, we might want to swtich to pathlib, however (in which case it's still not completely obvious what we should store to the database).

beets/util/__init__.py
wisp3rwind: Ok, and it appears that `shutil._samefile` also handles `OSError` in a similar way.

Any reason this is not just
```python
try:
    return os.path.samefile(...)
except OSError:
    return False
```

It's not more verbose in this case and the much more common idiom.

beets/util/__init__.py
wisp3rwind: That's actually unsurprising: This is only in the error handler for a cross-filesystem move. To test this, we would need to set up an extra filesystem (maybe on a loop device? mabye a bind-mount would suffice?). It would in fact be great to have that, since I suppose it's a common use case (e.g. have the music on some external device, but new files to import in your Downloads folder).

beetsplug/web/__init__.py
wisp3rwind: Actually, the `decode` is where the tests fail.

wisp3rwind:
I left a few more comments. The one thing that remains and where we seem to disagree is this:

---

> Ideally, I'd imagine these functions should accept the same types that the original ones do. In this case it's os.walk, which accepts any path-like str or bytes and returns the same type that it received.
>
> On the other hand, this PR was about fixing types rather than adjusting the implementation, so I focused on underlining what the current implement does and handles to make life easier to the person who will need to refactor this module later.

---

> I'd be happy with making these functions to handle bytes only if
>
>    We explicitly handle bytes only - remove the code that casts str to bytes from the beginning of the implementation
>    Rename the functions to make the only-bytes-accepted-here requirement clear, since the equivalents in os (normpath, samefile, ...) handle str and bytes types. Therefore, to not confuse people used to the builtins, we'd want to name these functions accordingly I suppose.

---

I guess there are two philosophies here:

1. Add typings according to anything the code accepts currently. This significantly reduces how much the type checker can help us to verify that our current path handling is correct, but might be the quicker path to a clean mypy run.

2. Add typings according to what is vaguely defined to be the intended usage (e.g. always call `syspath` with `bytes`). There may be some code around that would fail to type check due to this. I'd argue that this is actually a good thing, and probably indicates that the calling code needs to be fixed. In this interpretation, more lenient implementation (i.e. no `assert isinstance(path, bytes)` is a form of defensive programming.

I don't think we should start adding such assert statements immediately: This is too likely to cause issues. If we want to go this way, I think the steps should be

- Type the entire beets code base and fix any callers
- Add a warning where the assert should ultimately be
- Do a release, and mention in the release notes that plugin authors need to anticipate this change
- swap warnings for asserts

I'm not sure what's the best way to proceed here. I would very much like to see 2. happen eventually.
A broader roadmap could be:

- Type the entire beets code base, erring on the more lenient side.
- Prepare an updated version of https://beets.io/blog/paths.html, this should probably live in this repository rather than a blog. This should have clear guidance on what to do when
    - passing paths to `os.path.` and similar APIs
    - reading/writing paths or path fragments from/to files (e.g. playlist files)
    - passing paths to subprocesses
    - storing paths in the database
    - reading paths or path fragments from the configuration
    - ...
- Tighten up our typings accordingly and add warnings where the ultimately we would want assertions.
- Do a release, and mention in the release notes that plugin authors need to anticipate this change
- swap warnings for asserts
- At this point, I think we'd be in a good position to revisit a conversion to `pathlib`. We should start by updating our design document from the second point.

wisp3rwind:
See also https://github.com/beetbox/beets/issues/1409

beets/util/__init__.py
snejus: Perplexity summarised what I wanted to respond with very well
image

I initially learned about this from `flake8-simplify` [SIM-105](https://github.com/MartinThoma/flake8-simplify?tab=readme-ov-file#sim105)

```python
# Bad
try:
    foo()
except ValueError:
    pass

# Good
from contextlib import suppress

with suppress(ValueError):
    foo()
```

wisp3rwind: I'd disagree in this case, because the `except` arm is not `pass`, but rather an actual statement. In any case, I don't have a very strong opinion here, go with the option that you prefer.

beets/util/__init__.py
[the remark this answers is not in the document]
snejus: SQLite stores unicode fine these days (we may want to include its icu extension to cover users that have some ancient versions of SQLite).

For example, I locally store bandcamp releases in this db:

```sql
CREATE TABLE release (
    source TEXT,
    label TEXT,
    name TEXT,
    album TEXT,
    artist TEXT,
    title TEXT,
    comment TEXT,
    url TEXT,
    email_id TEXT,
    added TEXT,
    done INTEGER DEFAULT 0,
    done_date TEXT,
    bought INTEGER DEFAULT 0,
    next INTEGER DEFAULT 0,
    invalid INTEGER DEFAULT 0,
    PRIMARY KEY(email_id, added, source, name) ON CONFLICT IGNORE
  );
``` 

See some of the data that I've got there 
image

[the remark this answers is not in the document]
snejus: On the other hand I'd strongly suggest to keep the database as it is, at least for now, since this means opening a can of worms of migrating existing user databases.

What I meant by 'internally', I guess, is the code handling the paths before it hits the database (I assume that regardless of whether we supply a `str` or `bytes`, db code appropriately converts it to `bytes` for storage).

On the other hand paths being stored as `bytes` in db doesn't mean that we need to handle them as such across the codebase. I agree that `pathlib` is indeed the right way forward.

beetsplug/web/__init__.py
snejus: > For now, another solution might be to define types conditionally (mypy understands the `sys.platform` switch IIRC):
> 
> ```python
> if sys.platform == "win32":
>     SysPath: TypeAlias = str
> else:
>     SysPath: TypeAlias = bytes
> def syspath(path: bytes, prefix: bool = True) -> SysPath:
> ```

This is a nice solution - though the catch here is that for non-win32 systems the function actually returns `AnyStr`, and I'm not sure how to type it to avoid this issue

image

Regarding the failing tests, oops, I was meant to use `os.fsdecode` there

snejus: > Actually, it looks like the whole section
> 
> ```python
>     base_filename = os.path.basename(item_path)
>     # FIXME: Arguably, this should just use `displayable_path`: The latter
>     # tries `_fsencoding()` first, but then falls back to `utf-8`, too.
>     if isinstance(base_filename, bytes):
>         try:
>             unicode_base_filename = base_filename.decode("utf-8")
>         except UnicodeError:
>             unicode_base_filename = util.displayable_path(base_filename)
>     else:
>         unicode_base_filename = base_filename
> ```
> 
> is a no-op nowadays (always hitting the `else` arm. (It's not hit in tests at least: https://app.codecov.io/gh/beetbox/beets/blob/master/beetsplug%2Fweb%2F__init__.py

Note that the coverage is produced by running the tests in the `ubuntu` machine, so it doesn't show the logical branches that Windows hits 🙁

snejus: Regarding `syspath`, I reckon we could use the good ol' `cast`?

```python
    return cast(AnyStr, str_path)
```

snejus:
Weird stuff: as you've seen my change made that single test fail on Windows, see [this build](https://github.com/beetbox/beets/actions/runs/8966693983/job/24622737166) [22e2a3c](https://github.com/beetbox/beets/pull/5223/commits/22e2a3cd2c0842205cfb6b4be1a2f10adb704cda) for example

```
FAILED test/plugins/test_web.py::WebPluginTest::test_get_item_file - Attribut...
===== 1 failed, 1639 passed, 123 skipped, 9 xfailed in 199.92s (0:03:19) ======
```

Now in [the next commit](https://github.com/beetbox/beets/actions/runs/8968451206/job/24627879526) https://github.com/beetbox/beets/pull/5223/commits/abd8447218c6e340a9f672b140e4c0fcd71318c2 I reverted my change and I got bombarded with this
```
FAILED test/plugins/test_convert.py::ImportConvertTest::test_delete_originals
FAILED test/plugins/test_convert.py::ImportConvertTest::test_import_converted
FAILED test/plugins/test_convert.py::ConvertCliTest::test_convert - TypeError...
FAILED test/plugins/test_convert.py::ConvertCliTest::test_convert_keep_new - ...
FAILED test/plugins/test_convert.py::ConvertCliTest::test_convert_with_auto_confirmation
FAILED test/plugins/test_convert.py::ConvertCliTest::test_embed_album_art - T...
FAILED test/plugins/test_convert.py::ConvertCliTest::test_format_option - Typ...
FAILED test/plugins/test_convert.py::ConvertCliTest::test_no_transcode_when_maxbr_set_high_and_different_formats
FAILED test/plugins/test_convert.py::ConvertCliTest::test_playlist - TypeErro...
FAILED test/plugins/test_convert.py::ConvertCliTest::test_transcode_when_maxbr_set_low_and_different_formats
FAILED test/plugins/test_convert.py::ConvertCliTest::test_transcode_when_maxbr_set_low_and_same_formats
FAILED test/plugins/test_convert.py::ConvertCliTest::test_transcode_when_maxbr_set_to_none_and_different_formats
FAILED test/plugins/test_convert.py::NeverConvertLossyFilesTest::test_transcode_from_lossless
FAILED test/plugins/test_convert.py::NeverConvertLossyFilesTest::test_transcode_from_lossy
===== 14 failed, 1628 passed, 121 skipped, 9 xfailed in 194.90s (0:03:14) =====
```

https://github.com/beetbox/beets/assets/16212750/56e3cb3a-daca-41dc-b8bb-f5300b953173

beets/util/__init__.py
snejus: This makes sense!

beetsplug/web/__init__.py
snejus: `syspath` implementation is now back to where it was before

snejus:
@wisp3rwind as you suggested I used `bytes` wherever possible and resolved the typing issues without having to use `AnyStr` everywhere.

I'd be happy to move this forward if you have time to have another look!

beets/util/__init__.py
```diff
@@ -310,34 +313,34 @@ def prune_dirs(
...
-        ancestors = ancestors[ancestors.index(root) + 1 :]
+    elif root_bytes in ancestors:
+        # Only remove directories below the root_bytes.
+        ancestors = ancestors[ancestors.index(root_bytes) + 1 :]
     else:
         # Remove nothing.
         return
 
+    bytes_clutter = [bytestring_path(c) for c in clutter]
+
```
wisp3rwind, line 330: Nice catch, hoisting this out of the loop!

beets/util/__init__.py
```diff
@@ -838,7 +854,7 @@ def plurality(objs: Sequence[T]) -> T:
     return c.most_common(1)[0]
 
 
-def convert_command_args(args: List[bytes]) -> List[str]:
+def convert_command_args(args: List[AnyStr]) -> List[str]:
```
wisp3rwind, line 857:
```suggestion
def convert_command_args(args: List[bytes | str]) -> List[str]:
```
Not entirely sure, but I'd guess that we actually use this with lists containing both `bytes` and `str`. I think `AnyStr` forces this to be a list with a fixed item type.

beets/util/__init__.py
```diff
@@ -857,10 +873,7 @@ class CommandOutput(NamedTuple):
     stderr: bytes
 
 
-def command_output(
-    cmd: List[Bytes_or_String],
-    shell: bool = False,
-) -> CommandOutput:
+def command_output(cmd: List[AnyStr], shell: bool = False) -> CommandOutput:
```
wisp3rwind, line 876: Same as above regarding `bytes | str` vs. `AnyStr`.

beets/util/__init__.py
```diff
@@ -894,7 +907,7 @@ def command_output(
     if proc.returncode:
         raise subprocess.CalledProcessError(
             returncode=proc.returncode,
-            cmd=" ".join(map(str, cmd)),
+            cmd=" ".join(converted_cmd),
```
wisp3rwind, line 910: I'm wondering where this `map(...)` came from historically... good thing to clean this up!

wisp3rwind (commented):
I've had a quick look, nothing much that sticks out! There's one thing regarding subprocess command formatting that I'm unsure about.

Otherwise, it seems that there have been some changes around `syspath` and the `bytes` vs `str` vs `Path` question. I don't think I'm qualified to comment on the correctness of related changes here, I'm definitely out of the loop on what the correct usage of these types would currently be.

beets/util/__init__.py
snejus: To be honest `AnyStr` is only useful when it's used as one of the input parameter types and the output type - it communicates that the type stays the same. For example, `os.path.normpath`:

```python
def normpath(path: _PathLike[AnyStr]) -> AnyStr
def normpath(path: AnyStr) -> AnyStr
```

In this case there's no need for it, so I'm making the adjustment you suggested!

beets/util/__init__.py
snejus: Potentially lack of clarity regarding the output type of `convert_command_args`?

```python
    converted_cmd = convert_command_args(cmd)
```

snejus:
> Otherwise, it seems that there have been some changes around `syspath` and the `bytes` vs `str` vs `Path` question. I don't think I'm qualified to comment on the correctness of related changes here, I'm definitely out of the loop on what the correct usage of these types would currently be.

@wisp3rwind note that `syspath` implementation has not been adjusted:

image

wisp3rwind:
> > Otherwise, it seems that there have been some changes around `syspath` and the `bytes` vs `str` vs `Path` question. I don't think I'm qualified to comment on the correctness of related changes here, I'm definitely out of the loop on what the correct usage of these types would currently be.
> 
> @wisp3rwind note that `syspath` implementation has not been adjusted:

Not here; what I meant is that it has changed recently (in 01c24bb3a72760cd82c3f3417b56c5796ac55077), at a time when I've already been following beets development less closely.

snejus:
> > > Otherwise, it seems that there have been some changes around `syspath` and the `bytes` vs `str` vs `Path` question. I don't think I'm qualified to comment on the correctness of related changes here, I'm definitely out of the loop on what the correct usage of these types would currently be.
> > 
> > 
> > @wisp3rwind note that `syspath` implementation has not been adjusted:
> 
> Not here; what I meant is that it has changed recently (in [01c24bb](https://github.com/beetbox/beets/commit/01c24bb3a72760cd82c3f3417b56c5796ac55077)), at a time when I've already been following beets development less closely.

Ah I see!

That wasn't a very significant change to be honest: I added support for the `Path` type to `syspath` and `bytestring_path`, thus the change involved making sure that the input is cast appropriately.

snejus:
Is there anything else that needs addressing here @wisp3rwind?

wisp3rwind:
> Is there anything else that needs addressing here @wisp3rwind?

Looks fine!

snejus:
Cool, will merge it in once you've approved it 🙂

wisp3rwind (approved):

PR #5227, opened 2024-05-06
Replace `py3_path` and `cpu_count` implementations with builtins

snejus:
- **Replace py3_path by standard os.fsdecode**
- **Use native os.cpu_count**
- **Replace bytestring_path with os.fsencode**

## Description

These functions had been written before builtins achieving the same have been made available (in
Python 3.2 and 3.4). 

Now that they are available though, this PR updates the codebase to use builtin implementations rather than the custom ones.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -51,7 +51,7 @@ from typing import (
 from unidecode import unidecode
 
 import beets
-from beets.util import functemplate, py3_path
+from beets.util import functemplate
 
 from ..util.functemplate import Template
 from . import types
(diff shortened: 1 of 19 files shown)


snejus:
Tip: consider using `git diff --color-words="[[:alnum:]_]+|[^[:alnum:]_.[:space:]]+" master..` to go through the changes in this PR

snejus:
The changelog update workflow is failing for some reason

image

snejus:
Removed the commit which replaced `bytestring_path`. Leaving this to be tackled together with `syspath` later.

Serene-Arc (commented):
So I notice that you've replaced every instance of the `py3_path` function with `os.fsdecode`. What does this do when the parameter is already a string? Because if it was it appears, and it's just returned, there are possibly places where it could be removed entirely. Not a big deal, since pathlib with change it likely anyway, but still worth noting.

Serene-Arc:
Otherwise the tests seem to pass, so hopefully this means that the database is getting and giving the correct paths too, so I don't see a reason this shouldn't be merged. I'll need to rebase my pathlib PR on it but that should be easy.

Serene-Arc (approved):

snejus:
> What does this do when the parameter is already a string? 

It indeed just passes it through. 🙂

PR #5232, opened 2024-05-07
Configure mypy

snejus:
Move `mypy` configuration to `setup.cfg` and define the paths to lint in the
configuration, removing the need to explicitly specify them in `tox.ini`.


wisp3rwind (approved):
lgtm :tada:

PR #5233, opened 2024-05-07
Test aura plugin

snejus:
In #4746 I was making a small adjustment in `beetsplug/aura.py` and found that the module
wasn't tested. So this PR adds some high-level tests to act a safeguard for any future
adjustments.


snejus:
Reopening it from the `beets` repository

PR #5234, opened 2024-05-07
Remove flake8 docstrings and its exclusion rules

snejus:
Have come across this interesting case since I submitted a PR with an edit in a file that falls out of these exclusion rules: #5233 . :grin:

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -649,21 +649,6 @@ pycodestyle = ">=2.9.0,<2.10.0"
 pyflakes = ">=2.5.0,<2.6.0"
 
 [[package]]
-name = "flake8-docstrings"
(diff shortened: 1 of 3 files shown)


wisp3rwind:
So that point is that we're not enforcing this at all, so there's no point in even having the check; is that correct?

snejus:
> So that point is that we're not enforcing this at all, so there's no point in even having the check; is that correct?

Indeed.

Serene-Arc:
Adding to that, it's removing a rule that enforces a practice from a defunct practice. Since that's the case, and we're not even really using it, I think it can be removed.

Merge when you want @snejus

Serene-Arc (approved):

Issue #4709, opened 2023-03-14
Avoid using double-quoted string literals in SQL queries


TheRealDev0:
### Problem

With a default beets installation (via pkg or ports) on FreeBSD (13.1), any action requiring interaction with the SQLite database will cause beets to crash.

This is due to Double-quoted String Literals being disabled by default in the 3.41.0,1 version of SQLite - enabling DQS in the make options for the SQLite port corrects this issue and allows beets to complete database functions normally.

The maintainers of the SQLite port on FreeBSD have since re-enabled DQS by default but are planning to disable DQS by default no later than 20240101.

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv ls
```

Led to this problem:

```
Traceback (most recent call last):
  File "/usr/local/bin/beet", line 33, in <module>
    sys.exit(load_entry_point('beets==1.6.0', 'console_scripts', 'beet')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/beets/ui/__init__.py", line 1285, in main
    _raw_main(args)
  File "/usr/local/lib/python3.11/site-packages/beets/ui/__init__.py", line 1272, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "/usr/local/lib/python3.11/site-packages/beets/ui/commands.py", line 1089, in list_func
    list_items(lib, decargs(args), opts.album)
  File "/usr/local/lib/python3.11/site-packages/beets/ui/commands.py", line 1084, in list_items
    for item in lib.items(query):
                ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/beets/library.py", line 1529, in items
    return self._fetch(Item, query, sort or self.get_default_item_sort())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/beets/library.py", line 1503, in _fetch
    return super()._fetch(
           ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/beets/dbcore/db.py", line 1093, in _fetch
    rows = tx.query(sql, subvals)
           ^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #5235, opened 2024-05-07
db: disable DQS on Python >= 3.12

wisp3rwind:
cf. https://github.com/beetbox/beets/issues/4709, let's see how badly this breaks CI

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -20,6 +20,7 @@ import contextlib
 import os
 import re
 import sqlite3
+import sys
 import threading
 import time
 from abc import ABC
@@ -1068,6 +1069,12 @@ class Database:
             # call conn.close() in _close()
             check_same_thread=False,
         )
+
+        if sys.version_info >= (3, 12) and sqlite3.sqlite_version_info >= (3, 29, 0):
+            # If possible, disable double-quoted strings
+            conn.setconfig(sqlite3.SQLITE_DBCONFIG_DQS_DDL, 0)
+            conn.setconfig(sqlite3.SQLITE_DBCONFIG_DQS_DML, 0)
+
         self.add_functions(conn)
 
         if self.supports_extensions:


wisp3rwind:
ah, the Ubuntu LTS on CI has an ancient SQLite...

sungo:
to preface, I know precisely jack about python. I'm cutting and pasting to try and get my setup back online :) but after rolling a python 3.12 pyenv on freebsd, I tried this patch. it doesn't work for me unless those 0s are changed to 1s. so
`conn.setconfig(sqlite3.SQLITE_DBCONFIG_DQS_DDL, 1)` etc. 

with that change, I seem to be up and running again on freebsd 13.1

wisp3rwind:
Oh, but if DQS can still be enabled at runtime, we could merge that change quickly as a workaround until we really remove all DQS usage (at least if the latter turns out to be a lot of work).

wisp3rwind:
Oh, but if DQS can still be enabled at runtime, we could merge that change quickly as a workaround until we really remove all DQS usage (at least if the latter turns out to be a lot of work).

wisp3rwind:
Rebased on master to see what happens with #5447, but CI cannot properly test this these days: https://github.com/beetbox/beets/issues/5450

beets/dbcore/db.py
```diff
@@ -1070,6 +1071,16 @@ def _create_connection(self) -> Connection:
             # call conn.close() in _close()
             check_same_thread=False,
         )
+
+        if sys.version_info >= (3, 12) and sqlite3.sqlite_version_info >= (
+            3,
+            29,
```
snejus, line 1130: Why is the version `3.29.0` instead of `3.41.0` as you specified in the [issue description](https://github.com/beetbox/beets/issues/4709)?

> This is due to Double-quoted String Literals being disabled by default in the 3.41.0,1 version of SQLite

wisp3rwind: 3.29 added the the config options used here, 3.41 changed the defaults; see https://www.sqlite.org/changes.html

So 3.29 is the earliest version for which we can ensure to disable DQS. Starting from 3.41 it's disabled anyway with default build options, but it seems prudent to me to enforce it here to ensure that we won't regress.

snejus: I see, thanks. What would you think about removing this conditional logic and instead *always* enforcing double quotes?

wisp3rwind: Now, I'm confused as to what you propose: double quoted string literals are the problematic feature, we want to _prohibit_ them so that we're sure to immediately see an error if a new instance of them is added accidentally. (Rather than only get a bug report later on when someone runs beets with a newer SQLite, which happened in #4709.) The idea with this PR was:

- SQLite < 3.29: DQS enabled by default, but nothing we can do about it, since SQLite doesn't have those config options
- SQLite >= 3.29, < 3.41: DQS enabled by default, but we can disable them
- SQLite >= 3.41, < 3.41: DQS disabled by default, no harm in disabling them; in fact, we might want to since people might build SQLite with different defaults (not sure whether anybody does)

We also need Python 3.12 to do anything, because that added `setconfig`.

In other words, if we attempt run this unconditionally, I'd expect either the Python library or SQLite itself to error out.

The relevance of this PR is probably decreasing, since newer version of Python and SQLite will become more prevalent. I'd argue to add it anyway, it's pretty simple, and it makes obvious and certain that DQS are disabled.

snejus: Apologies, I was meant to say the opposite: always enforcing **single** quotes 🤦🏼

wisp3rwind: Ok :+1: But, as I tried to explain, I don't think there's a way to do that unconditionally, at least not via the mechanism used here. Of course, the idea would be that DQS are generally disallowed in beets. We can't enforce that in every environment where beets might be run. This PR tries to make it more certain that we actually verify this at least in some cases (e.g. some CI runs): The problem with https://github.com/beetbox/beets/issues/4709 was that none of our CI runs actually used a version of SQLite where DQS are disabled by default, so we never saw the issue. Maybe the best way to think of this PR is indeed as an additional CI check.

snejus:
I should have approved this a while ago, apologies @wisp3rwind.

PR #5236, opened 2024-05-08
Fix for #4709

### LINKED ISSUE #4709: Avoid using double-quoted string literals in SQL queries

arogl:
## Description

Fixes #4709.

Changelog to complete if testing works out

## To Do

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,7 @@ htmlcov/
 coverage.xml
 *,cover
 .hypothesis/
+.reports/
 
 # Flask stuff:
 instance/
--- a/beets/library.py
+++ b/beets/library.py
@@ -310,7 +310,7 @@ class SmartArtistSort(dbcore.query.Sort):
         collate = "COLLATE NOCASE" if self.case_insensitive else ""
         return (
             "(CASE {0}_sort WHEN NULL THEN {0} "
-            'WHEN "" THEN {0} '
+            "WHEN '' THEN {0} "
             "ELSE {0}_sort END) {1} {2}"
         ).format(field, collate, order)
 
--- a/beets/library.py
+++ b/beets/library.py
@@ -308,11 +308,8 @@ class SmartArtistSort(dbcore.query.Sort):
         order = "ASC" if self.ascending else "DESC"
         field = "albumartist" if self.album else "artist"
         collate = "COLLATE NOCASE" if self.case_insensitive else ""
-        return (
-            "(CASE {0}_sort WHEN NULL THEN {0} "
-            "WHEN '' THEN {0} "
-            "ELSE {0}_sort END) {1} {2}"
-        ).format(field, collate, order)
+
+        return f"COALESCE(NULLIF({field}_sort, ''), {field}) {collate} {order}"
 
     def sort(self, objs):
         if self.album:
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -166,6 +166,7 @@ New features:
 
 Bug fixes:
 
+* :bug:`4709`
 * :doc:`/plugins/lastimport`: Improve error handling in the `process_tracks` function and enable it to be used with other plugins.
 * :doc:`/plugins/spotify`: Improve handling of ConnectionError.
 * :doc:`/plugins/deezer`: Improve Deezer plugin error handling and set requests timeout to 10 seconds.
--- a/test/test_library.py
+++ b/test/test_library.py
@@ -59,7 +59,7 @@ class StoreTest(_common.LibTestCase):
         self.i.store()
         new_year = (
             self.lib._connection()
-            .execute("select year from items where " 'title="the title"')
+            .execute("select year from items where title='the title'")
             .fetchone()["year"]
         )
         self.assertEqual(new_year, 1987)
@@ -70,7 +70,7 @@ class StoreTest(_common.LibTestCase):
         self.i.store()
         new_genre = (
             self.lib._connection()
-            .execute("select genre from items where " 'title="the title"')
+            .execute("select genre from items where title='the title'")
             .fetchone()["genre"]
         )
         self.assertEqual(new_genre, original_genre)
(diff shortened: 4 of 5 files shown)


arogl:
Doing further testng with #5235 enabled on windows to compare failures

arogl:
I also noted the runners for python 3.x DO NOT run any tests apart from mypy 😢 they have been excluded.

They also no longer test the next beta version.

arogl:
Thsi should be good of further tetsing before merging.
I'll try out on ubuntu 22.04 (WSL), and teh latest version of as much as possible

wisp3rwind:
> I also noted the runners for python 3.x DO NOT run any tests apart from mypy 😢 they have been excluded.
> 
> They also no longer test the next beta version.

yes, I realized the same yesterday, we should revisit our CI first. I also remember that we used to have a run for the next Python alpha; we should bring that back: In particular, we already know that Python 3.13 is going to break beets due to the `imghdr` removal.

arogl:
Tested on fully updated Arch and Ubuntu-24.04 with #5235 set with DQS=0|1

beets/library.py
```diff
@@ -310,7 +310,7 @@ def order_clause(self):
         collate = "COLLATE NOCASE" if self.case_insensitive else ""
         return (
             "(CASE {0}_sort WHEN NULL THEN {0} "
-            'WHEN "" THEN {0} '
+            "WHEN '' THEN {0} "
```
snejus, line 313: May be worth to consider replacing this entire thing with something like

```python
        order = "ASC" if self.ascending else "DESC"
        collate = "COLLATE NOCASE" if self.case_insensitive else ""

        return f"COALESCE(NULLIF({self.field}_sort, ''), {self.field}) {collate} {order}"
```

test/test_library.py
```diff
@@ -59,7 +59,7 @@ def test_store_changes_database_value(self):
         self.i.store()
         new_year = (
             self.lib._connection()
-            .execute("select year from items where " 'title="the title"')
+            .execute("select year from items where title='the title'")
```
snejus, line 62: While you're here it may be good to replace this with
```suggestion
            .execute("select year from items where title = ?", (i.title,))
```

Same applies to the rest of the updates. **the title** is taken out of context and it isn't clear whether it's looking for the the item that has just been stored or some other one

snejus: See slightly below, this pattern is being used here for example

```python
    def test_albuminfo_for_two_items_doesnt_duplicate_row(self):
        i2 = item(self.lib)
        self.lib.get_album(self.i)
        self.lib.get_album(i2)

        c = self.lib._connection().cursor()
        c.execute("select * from albums where album=?", (self.i.album,))
        # Cursor should only return one row.
```

Just checked this test in question, and realised I made a typo. It should instead be

```python
            .execute("select year from items where title = ?", (self.i.title,))
```

The test works fine on my end.

arogl: Updated to user passed values rather than hard coded values

.gitignore
```diff
@@ -56,6 +56,7 @@ htmlcov/
 coverage.xml
 *,cover
 .hypothesis/
+.reports/
```
snejus, line 59: Thanks for adding this, nice!

docs/changelog.rst
```diff
@@ -166,6 +166,7 @@ New features:
 
 Bug fixes:
 
+* :bug:`4709`
```
snejus, line 169: Could you add a small description of the fix?

snejus (changes requested):
Sorry for not coming back to this earlier, completely forgot about it 😞 Just one last thing missing - a description of the fix in the changelog

arogl:
Reworked in another PR

PR #5239, opened 2024-05-09
Test aura

snejus:
In #4746 I was making a small adjustment in beetsplug/aura.py and found that the module wasn't tested. So this PR adds some high-level tests to act a safeguard for any future adjustments.

### DIFF
--- a/beetsplug/aura.py
+++ b/beetsplug/aura.py
@@ -20,7 +20,7 @@ import re
 from dataclasses import dataclass
 from mimetypes import guess_type
 from os.path import getsize, isfile
-from typing import Mapping
+from typing import ClassVar, Mapping, Type
 
 from flask import (
     Blueprint,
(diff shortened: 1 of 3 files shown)


snejus:
@Serene-Arc could you have a glance at this one? It should be an easy one!

Serene-Arc (approved):
I'm not the most familiar with aura as a part of beets but I looked through all the changes and nothing seems untoward. Nice deduplication of some of it, and it's great that we can use the dataclass attribute now! There's probably a couple other places that could use it.

You do have a change to `setup.py` which means that this is going to cause a conflict with #5266. I'll approve but won't merge; you can decide in what order to merge these two so that it doesn't cause problems.

snejus:
Going to merge this one first. #5266 removes `setup.py` altogether and already has `pytest-flask`, so I'll rebase the PR to resolve the conflict :)

PR #5240, opened 2024-05-09
Make queries fast, filter all flexible attributes

### LINKED ISSUE #3515: Track-level query returning list of albums containing those tracks

### LINKED ISSUE #4360: beet list -a path:some/path queries are extremely slow

snejus:
## Description

Another and (hopefully) final attempt to improve querying speed.

Fixes #4360 
Fixes #3515
and possibly more issues to do with slow queries.

This PR supersedes #4746.

## What's been done
The `album` and `item` tables are joined, and corresponding data from `item_attributes` and `album_attributes` is merged and made available for filtering. This enables to achieve the following:

- [x] Faster album path queries, `beet list -a path::some/path` 
- [x] Faster flexible attributes queries, both albums and tracks, `beet list play_count:10` 
- [x] (New) Ability to filter albums with track-level (and vice-versa) **db** field queries, `beet list -a title:something` 
- [x] (New) Ability to filter tracks with album-level **flexible** field queries, `beet list artpath:cover`
- [x] (New) Ability to filter albums with track-level **flexible** field queries, `beet list -a art_source:something`

## Benchmarks
image

You can see that now querying speed is more or less constant regardless of the query, and the speed is mostly influenced by how many results need to be printed out
image

Compare this with what we had previously
image

## To Do

## Later
https://github.com/beetbox/beets/issues/5318
https://github.com/beetbox/beets/issues/5319

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -93,11 +93,8 @@ class Query(ABC):
         Return (clause, subvals) where clause is a valid sqlite
         WHERE clause implementing the query and subvals is a list of
         items to be substituted for ?s in the clause.
-
-        The default implementation returns None, falling back to a slow query
-        using `match()`.
         """
-        return None, ()
+        raise NotImplementedError
 
     @abstractmethod
     def match(self, obj: Model):
@@ -146,17 +143,16 @@ class FieldQuery(Query, Generic[P]):
 
     @property
     def col_name(self) -> str:
+        if not self.fast:
+            return f'json_extract("flex_attrs [json_str]", "$.{self.field}")'
+
         return f"{self.table}.{self.field}" if self.table else self.field
 
     def col_clause(self) -> Tuple[str, Sequence[SQLiteType]]:
-        return self.col_name, ()
+        raise NotImplementedError
 
-    def clause(self) -> Tuple[Optional[str], Sequence[SQLiteType]]:
-        if self.fast:
-            return self.col_clause()
-        else:
-            # Matching a flexattr. This is a slow query.
-            return None, ()
+    def clause(self) -> Tuple[str, Sequence[SQLiteType]]:
+        return self.col_clause()
 
     @classmethod
     def value_match(cls, pattern: P, value: Any):
@@ -305,7 +301,17 @@ class RegexpQuery(StringFieldQuery[Pattern[str]]):
         return pattern.search(cls._normalize(value)) is not None
 
 
-class BooleanQuery(MatchQuery[int]):
+class NumericColumnQuery(MatchQuery[AnySQLiteType]):
+    """A base class for queries that work with NUMERIC SQLite affinity."""
+
+    @property
+    def col_name(self) -> str:
+        """Cast a flexible attribute column (string) to NUMERIC affinity."""
+        col_name = super().col_name
+        return col_name if self.fast else f"CAST({col_name} AS NUMERIC)"
+
+
+class BooleanQuery(NumericColumnQuery[bool]):
     """Matches a boolean field. Pattern should either be a boolean or a
     string reflecting a boolean.
     """
@@ -357,7 +363,7 @@ class BytesQuery(FieldQuery[bytes]):
         return pattern == value
 
 
-class NumericQuery(FieldQuery[str]):
+class NumericQuery(NumericColumnQuery[Union[int, float]]):
     """Matches numeric fields. A syntax using Ruby-style range ellipses
     (``..``) lets users specify one- or two-sided ranges. For example,
     ``year:2001..`` finds music released since the turn of the century.
@@ -483,7 +489,7 @@ class CollectionQuery(Query):
     def clause_with_joiner(
         self,
         joiner: str,
-    ) -> Tuple[Optional[str], Sequence[SQLiteType]]:
+    ) -> Tuple[str, Sequence[SQLiteType]]:
         """Return a clause created by joining together the clauses of
         all subqueries with the string joiner (padded by spaces).
         """
@@ -491,9 +497,6 @@ class CollectionQuery(Query):
         subvals = []
         for subq in self.subqueries:
             subq_clause, subq_subvals = subq.clause()
-            if not subq_clause:
-                # Fall back to slow query.
-                return None, ()
             clause_parts.append("(" + subq_clause + ")")
             subvals += subq_subvals
         clause = (" " + joiner + " ").join(clause_parts)
@@ -533,7 +536,7 @@ class AnyFieldQuery(CollectionQuery):
     def field_names(self) -> Set[str]:
         return set(self.fields)
 
-    def clause(self) -> Tuple[Optional[str], Sequence[SQLiteType]]:
+    def clause(self) -> Tuple[str, Sequence[SQLiteType]]:
         return self.clause_with_joiner("or")
 
     def match(self, obj: Model) -> bool:
@@ -572,7 +575,7 @@ class MutableCollectionQuery(CollectionQuery):
 class AndQuery(MutableCollectionQuery):
     """A conjunction of a list of other queries."""
 
-    def clause(self) -> Tuple[Optional[str], Sequence[SQLiteType]]:
+    def clause(self) -> Tuple[str, Sequence[SQLiteType]]:
         return self.clause_with_joiner("and")
 
     def match(self, obj: Model) -> bool:
@@ -582,7 +585,7 @@ class AndQuery(MutableCollectionQuery):
 class OrQuery(MutableCollectionQuery):
     """A conjunction of a list of other queries."""
 
-    def clause(self) -> Tuple[Optional[str], Sequence[SQLiteType]]:
+    def clause(self) -> Tuple[str, Sequence[SQLiteType]]:
         return self.clause_with_joiner("or")
 
     def match(self, obj: Model) -> bool:
@@ -602,14 +605,9 @@ class NotQuery(Query):
         """Return a set with field names that this query operates on."""
         return self.subquery.field_names
 
-    def clause(self) -> Tuple[Optional[str], Sequence[SQLiteType]]:
+    def clause(self) -> Tuple[str, Sequence[SQLiteType]]:
         clause, subvals = self.subquery.clause()
-        if clause:
-            return f"not ({clause})", subvals
-        else:
-            # If there is no clause, there is nothing to negate. All the logic
-            # is handled by match() for slow queries.
-            return clause, subvals
+        return f"not ({clause})", subvals
 
     def match(self, obj: Model) -> bool:
         return not self.subquery.match(obj)
@@ -816,7 +814,7 @@ class DateInterval:
         return f"[{self.start}, {self.end})"
 
 
-class DateQuery(FieldQuery[str]):
+class DateQuery(NumericColumnQuery[int]):
     """Matches date fields stored as seconds since Unix epoch time.
 
     Dates can be specified as ``year-month-day`` strings where only year
@@ -910,7 +908,7 @@ class Sort:
         return sorted(items)
 
     def is_slow(self) -> bool:
-        """Indicate whether this query is *slow*, meaning that it cannot
+        """Indicate whether this sort is *slow*, meaning that it cannot
         be executed in SQL and must be executed in Python.
         """
         return False
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -144,7 +144,7 @@ class FieldQuery(Query, Generic[P]):
     @property
     def col_name(self) -> str:
         if not self.fast:
-            return f'json_extract("flex_attrs [json_str]", "$.{self.field}")'
+            return f'json_extract(all_flex_attrs, "$.{self.field}")'
 
         return f"{self.table}.{self.field}" if self.table else self.field
 
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -135,7 +135,7 @@ class FieldQuery(Query, Generic[P]):
     """
 
     def __init__(self, field: str, pattern: P, fast: bool = True):
-        self.field = field
+        self.table, _, self.field = field.rpartition(".")
         self.pattern = pattern
         self.fast = fast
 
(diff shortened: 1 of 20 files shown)


snejus:
I'm using the `test-aura` branch as the base since it depends on it getting merged. Though for now, I will change the base to `master` in order to run the tests.

beets/importer.py
```diff
@@ -1019,7 +1019,7 @@ def find_duplicates(self, lib):
         # temporary `Item` object to generate any computed fields.
         tmp_item = library.Item(lib, **info)
         keys = config["import"]["duplicate_keys"]["item"].as_str_seq()
-        dup_query = library.Album.all_fields_query(
+        dup_query = library.Item.match_all_query(
```
snejus, line 1022: It seemed to me this was supposed to be `Item` instead of `Album`?

Serene-Arc:
I'll be able to review in a week or two, just end of semester push at the moment.

Serene-Arc:
I've left some comments on specific lines and things, but some general comments. I noticed that you changed `field` to `col_name` in the definition of `FieldQuery` which is reflected in `bareasc`. Do you think that this will have an impact on the compatibility with plugins more broadly? Going through the plugins with grep, I couldn't find any other instances but it is something to keep in mind. You haven't altered any of the interfaces in our docs so it isn't *too* much of an issue, but it will definitely have to be mentioned in the changelog.

I think it's quite clever how you merged all of the flexible attributes and optimised the SQL queries! This is a great PR, will really improve how beets works. 

Btw, with those later PRs to do, I'd open bugs/enhancements for them so we don't forget!

Also, those benchmarks are great. We should add them as a poetry command I think, in another PR would probably be best. What have you used to make those?

snejus:
> I've left some comments on specific lines and things

Hm I'm not being able to see them! Did you submit the review?

Serene-Arc (changes requested):

Serene-Arc:
...I did not! I have now.

snejus:
> I noticed that you changed field to col_name in the definition of FieldQuery which is reflected in `bareasc`. Do you think that this will have an impact on the compatibility with plugins more broadly? Going through the plugins with grep, I couldn't find any other instances but it is something to keep in mind. 

Well noted @Serene-Arc. I think I adjusted every internal plugin that used this functionality. Otherwise, I haven't seen any external plugins based on this interface.

> You haven't altered any of the interfaces in our docs so it isn't too much of an issue, but it will definitely have to be mentioned in the changelog.

You're right, this completely needs a mention in the changelog!

snejus:
> Also, those benchmarks are great. We should add them as a poetry command I think, in another PR would probably be best. What have you used to make those?

This one's a bit complicated. 😅 In short, 
* used `hyperfine` to measure the performance between the two versions and save the results as JSON
* `jq` to handle the JSON data
* `rich-tables` to render it
* `zsh` as the glue between these parts

## In detail

### 1. Prepare each executable for testing
* Install `beets` from `master` branch as `beetmaster`: (in the repository) `pipx install . --suffix master` 
* Install `beets` from my branch as `beetfast`: `pipx install -e . --suffix fast`.

Now one can run `beetmaster` and `beetfast` to test each version accordingly.

### 2. Prepare some queries for testing

I added the following contents to a file `./queries.txt`. One query per line

```
artpath::a
artpath::a -a
path::a
path::a -a
art_source::a
art_source::a -a
title:a
title:a -a
play_count:5..
play_count:5.. -a
play_count:50..
play_count:50.. -a
art_source::a play_count::5 city:Be
art_source::a play_count::5 city:Be -a
art_source::. play_count:..5 city::.*
art_source::. play_count:..5 city::.* -a
```

### 3. Use `hyperfine` to measure the performance between the two

```zsh
compare_performance() {
  # args: <queries-file> <results-json-file>
  # info: read queries from _queries-file, compare performance and save results to _results-json-file
  # read queries
  local queries_file=$1
  local results_json_target=$2

  local -a queries
  queries=(${(f@)"$(grep -v "^#" $queries_file)"})

  # compare performance of beetmaster and beetfast using hyperfine.
  # For each of the given queries, hyperfine runs each command 5 times
  # and exports the results to a json file.
  hyperfine --parameter-list query ${(j:,:)queries} --runs 5 --ignore-failure \
    --export-json $results_json_target \
    'beetmaster -l /home/sarunas/.music/beets/library.db list {query}' \
    'beetfast -l /home/sarunas/.music/beets/library.db list {query}'
}
```

### 4. Count how many items each command returns and insert it into the results

```zsh
add_results_count() {
  # args: <performance-results-json> <performance-with-results-count-json-target>
  # info: count the number of results each command returns for each query, add
  ## this to the performance results json file and save it in the given target file
  local performance_json=$1
  local target_json=$2
  local -a runs
  runs=(${(f@)"$(jq '.results[].command' < $performance_json -r)"})

  local run
  rm results_count.txt
  # save each command (which includes the executable and query) and the number of results it returned
  for run in $runs; do
    print "$run $($run | wc -l)" >> results_count.txt
  done

  # parse the results to JSON objects
  jq '
    capture("(?<command>.*) (?<results>[^ ]+)$")
    | {(.command): (.results | tonumber)}
  ' -R <results_count.txt |
    jq add -s > results_count.json

  # add the results count to the performance results json
  jq '.results |= map(.results_count = $index[.command])' \
    --argjson index "$(<results_count.json)" \
    <$performance_json >$target_json
}
```

### 5. Get relevant data from the results JSON and display them with `rich-tables`
```zsh
show_hyperfine_results() {
  # args: <results-json-file>
  # info: print hyperfine results nicely
  local -a json
  zparseopts -D -E j=json
  jq '
    def mv(before; after): .[after] = .[before] | del(.[before]);
    def rm(regex): gsub(regex; "");
    def round(num): . * pow(10; num) | round / pow(10; num);
    def group_map(field): 
        (field | split(",") | map(rm("^ *| *$"))) as $key
        | sort_by(.[$key[]])
        | group_by(.[$key[]]) 
        | sort_by(length)
        | (map({([.[0][$key[]]| if type == "array" then join(", ") end|tostring] | join(" | ")): map(del(.[$key[]]))}) | add);

    def group_map_dict(field): 
        (field | split(",")) as $key
        | sort_by(.[$key[]])
        | group_by(.[$key[]]) 
        | sort_by(length)
        | map((.[0] as $item | $key | map({(.): $item[.]}) | add) + {"items": map(del(.[$key[]]) | select(len(.) > 0))});

    .results
    | map(
      del(.times, .stddev, .median, .system, .user, .exit_codes)
      | (.mean, .min, .max) |= round(2)
      | mv("mean"; "duration")
      | mv("results_count"; "results")
      | .command |= (rm("BEETSDIR= ") | rm(" .*"))
      | .parameters |= .query
      | if (.parameters | test(" -a")) then .type = "album" else .type = "item" end
      | {type, command, parameters, results, duration}
    )
    | group_map("type")
    # | map_values(
      # group_map_dict("parameters")
    # )
    # | sort_by(.items[0].mean - .items[1].mean)
  ' $1 | {
    if (( $#json )); then
      jq
    else
      table
    fi
  }
}
```

### Steps 3-5
```zsh
benchmark_beets() {
  # args:
  # info: benchmark beetmaster and beetfast
  local queries_file=queries.txt
  local performance_json=master-vs-beetfast-counts.json
  local results_json=benchmark_with_counts.json

  compare_performance $queries_file $performance_json
  add_results_count $performance_json $results_json
  show_hyperfine_results $results_json
}
```

beets/importer.py
snejus: I don't think it's the same, as it uses `Album._fields` and `Item._fields` which are different, see

```python
    @classmethod
    def field_query(
        cls,
        field,
        pattern,
        query_cls: Type[FieldQuery] = MatchQuery,
    ) -> FieldQuery:
        """Get a `FieldQuery` for this model."""
        return query_cls(field, pattern, field in cls._fields)

    @classmethod
    def all_fields_query(
        cls: Type["Model"],
        pats: Mapping,
        query_cls: Type[FieldQuery] = MatchQuery,
    ):
        """Get a query that matches many fields with different patterns.

        `pats` should be a mapping from field names to patterns. The
        resulting query is a conjunction ("and") of per-field queries
        for all of these field/pattern pairs.
        """
        subqueries = [cls.field_query(k, v, query_cls) for k, v in pats.items()]
        return AndQuery(subqueries)
```

beets/dbcore/query.py
```diff
@@ -81,17 +82,19 @@ def __init__(self, what, expected, detail=None):
 class Query(ABC):
     """An abstract class representing a query into the database."""
 
+    @property
+    def field_names(self) -> Set[str]:
+        """Return a set with (field, fast) tuples."""
+        return set()
```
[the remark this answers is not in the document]
snejus: It's the docstring that's misleading here! It used to return tuples but has been changed :)

Serene-Arc:
That is a lot of steps for a benchmark! It's useful though, so I might make an issue for it as an enhancement and we can explore additional methods and any simplifications that can be done to the process. It's a good thing to include though, especially when this gets so complex. 

We could even include a GitHub action that tests it and gives a message if a PR increases average time by 50% or 25% or whatever we specify, just to give maintainers a hint to scrutinise it a little more closely.

docs/changelog.rst
```diff
@@ -6,6 +6,16 @@ Unreleased
...
+New features:
+
+* Ability to query albums with track-level (and vice-versa) **db** or
+  **flexible** field queries, for example `beet list -a title:something`, `beet
+  list artpath:cover`.
+* Queries have been made faster, and their speed is constant regardless of
+  their complexity or the type of queried fields. Notably, album queries for
+  the `path` field and those that involve flexible attributes have seen the
+  most significant speedup.
+
```
Serene-Arc, line 18: I think that there should be a mention of the interface that changed, along with the fact that it was not public, just to give any plugin developers a hint in case something does go wrong. It's not *really* our responsibility to manage non-public APIs in the program but if it was used in some plugins, it's reasonable to assume it could have been in others.

Serene-Arc (changes requested):

docs/changelog.rst
snejus: See the last commit - I attempted to keep the same `FieldQuery.field` interface like we had before. What do you think?

Serene-Arc: Ah, then there's no issue at all!

Serene-Arc (approved):

arsaboo:
@snejus I just updated after this PR, and it seems that everything is painfully slow. Even a simple `beet update` takes a significantly long time. Are we supposed to do anything else?

Serene-Arc:
@arsaboo what kind of requests are taking longer?

arsaboo:
Beet update seems like the worst hit; even beet import seems slow. Let me know what additional information you would like.

snejus:
hmm, interesting. @Serene-Arc , do no release any new versions just yet, I will investigate this over the coming days

snejus:
@arsaboo, can you provide the output of `beet stats` here?

arsaboo:
I added time to the command....even `stats` takes time. 
```bash
time beet stats
Tracks: 55012
Total time: 27.3 weeks
Approximate total size: 1.2 TiB
Artists: 16399
Albums: 11414
Album artists: 2361

real    0m27.519s
user    0m24.341s
sys     0m3.152s
```

Here's another output from `beet import` even after I specified the id (it took over a minute after accepting the prompt). This used to be almost instantaneous:

```bash
$ time beet import -m -I -t ~/shared/music/ --set genre="Filmi" --search-id 0vGMpTlGXYZ
deezer: Deezer API error: no data

/home/arsaboo/shared/music/Jubin Nautiyal/REDACTED (2024) (1 items)

  Match (88.2%):
  Jubin Nautiyal - REDACTED
  ≠ album, tracks
  Spotify, None, 2024, None, Tips Industries Ltd, None, None
  https://open.spotify.com/album/0vGMpTlGXYZ
  * Artist: Jubin Nautiyal
...
Enter search, enter Id, aBort, Print tracks, Open files with Picard,
eDit, edit Candidates?

real    1m15.144s
user    1m1.506s
sys     0m10.224s
```

``bash
~$ sqlite3 --version
3.37.2 2022-01-06 13:25:41 872ba256cbf61d9290b571c0e6d82a20c224ca3ad82971edc46b29818d5dalt1
$ python3 --version
Python 3.10.12
```

**Comment by snejus** [comment]
I see. You've got old SQLite version which doesn't have built-in JSON functionality - could you try upgrading your SQLite to a version higher than `3.38.0`?

**Comment by snejus** [comment]
@arsaboo can you also confirm you're using Ubuntu?

**Comment by arsaboo** [comment]
yes, I am on Ubuntu

**Comment by arsaboo** [comment]
Shouldn't sqlite be automatically updated?

**Comment by snejus** [comment]
I am reverting this change, look for an incoming pr

**Comment by snejus** [comment]
See https://github.com/beetbox/beets/pull/5326

**Comment by snejus** [comment]
> Shouldn't sqlite be automatically updated?

Unfortunately Ubuntu does not keep your packages up to date :-(

**Comment by Serene-Arc** [comment]
How unfortunate! This PR was very well done. It does highlight that we should have testing in place though for benchmarking purposes, at least where these types of PRs are concerned. And perhaps a wider range of machines on CI for benchmarking, such as specific ubuntu distros and so on, rather than just the latest.

**Comment by snejus** [comment]
@arsaboo have you updated to Ubuntu 24?

**Comment by snejus** [comment]
> How unfortunate! This PR was very well done. It does highlight that we should have testing in place though for benchmarking purposes, at least where these types of PRs are concerned. And perhaps a wider range of machines on CI for benchmarking, such as specific ubuntu distros and so on, rather than just the latest.

Agree with you! I was aware that it may be painfully slow on older versions of SQLite, it just that I wasn't aware that at that point Ubuntu shipped with such an old version of SQLite.

**Comment by arsaboo** [comment]
Yes, updated now (after the initial feedback). Does it work better on 24?

**Comment by snejus** [comment]
The SQLite version on 22 did not yet include the JSON extension, so it was painfully slow. 

Try checking out the `only-fast-filtering` branch: how fast are queries? Is the speed comparable to their speed on `master`?

**Comment by arsaboo** [comment]
@Serene-Arc, can you rebase the `only-fast-filtering` branch so I can try it again?

**Comment by snejus** [comment]
> @snejus can you rebase the `only-fast-filtering` branch so I can try it again?

Done!

**Comment by arsaboo** [comment]
@snejus I ran `beet update` with the current master branch and compared to the `only-fast-filtering` branch and here are the results:

only-fast-filtering

RUN 1:
```bash
real    65m39.027s
user    26m2.579s
sys     5m32.047s
```
RUN 2:
```bash
real    54m50.369s
user    25m52.297s
sys     5m31.133s
```bash

Current MASTER
RUN 1
```bash
real    21m21.877s
user    5m7.913s
sys     1m12.769s
```
RUN 2
```bash
real    41m39.848s
user    5m11.727s
sys     1m16.404s
```

While the fast-filtering is slower than the master, it is significantly faster than my previous experience with this.

Issue #5167, opened 2024-03-29
fetchart plugin does not import cover art when import is run with `move` and `quiet` enabled


mgoltzsche:
When I import a directory that includes a `cover.jpg` file into my beets library with both `import.move` and `import.quiet` set to `true`, the cover art is not imported.
When I disable `move` or `quiet`, it imports the cover art.

(However, when importing the cover art, the fetchart plugin does not use the `cover.jpg` that exists within the album source directory but fetches it from coverartarchive.org. Correspondingly, the `cover.jpg` file within the import source directory is not deleted after the import although `move` is enabled.)

### Problem

Imagine I have the import options `group_albums`, `move` and `quiet` enabled and the following file system structure to import from:
* `/data/ytimport/albums/some artist - some album`
  * `01 - First Track.opus`
  * `02 - Second Track.opus`
  * `cover.jpg`

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv import /data/ytimport/albums
```

Does not import any cover art.
The log shows that the `fetchart` plugin tries to lookup the cover art within the file system and that it doesn't try the other cover art sources:
```
$ beet -vv import /data/ytimport/albums/
user configuration: /data/beets/config.yaml
data directory: /data/beets
plugin paths: 
lastgenre: Loading canonicalization tree /usr/local/lib/python3.12/site-packages/beetsplug/lastgenre/genres-tree.yaml
fetchart: google: Disabling art source due to missing key
fetchart: lastfm: Disabling art source due to missing key
fetchart: To use Spotify as an album art source, you must install the beautifulsoup4 module. See the documentation for further details.
Sending event: pluginload
library database: /data/musiclibrary.db
library directory: /data/music
Sending event: library_opened
Sending event: import_begin
state file could not be read: [Errno 2] No such file or directory: '/data/beets/state.pickle'
Sending event: import_task_created
Sending event: import_task_created
Sending event: import_task_start
chroma: chroma: fingerprinted b'/data/ytimport/albums/Yussef Kamaal - Black Focus (Full Album Upload) [4D8YPDdsxYU]/01 - Black Focus.opus'
chroma: matched recordings ['b39a012c-ae1a-475a-bdaf-78ba6c04efc3'] on releases ['e6a5e707-7782-4258-8b22-9364629810bf', 'f5f47021-c189-42dd-89db-d64c161a8c6d', '1c06b236-d48a-4ff8-b047-7629b732f628']
[...]

PR #5244, opened 2024-05-13
fetchart: defer file removal config option evaluation

mgoltzsche:
## Description

Defer the evaluation of the source file removal options (`import.delete` and `import.move`) to the point where the fetchart plugin is actually called instead of only evaluating those configuration options on plugin initialization.
This is to allow other plugins (such as the [ytimport](https://github.com/mgoltzsche/beets-ytimport/blob/v1.8.1/beetsplug/ytimport/__init__.py#L194) plugin) to invoke the import directly (within the same python process; implicitly invoking the fetchart plugin) with temporarily overwritten configuration options.

Addresses https://github.com/beetbox/beets/issues/5167#issuecomment-2106465172

## To Do

### DIFF
--- a/test/plugins/test_art.py
+++ b/test/plugins/test_art.py
@@ -825,9 +825,13 @@ class ArtImporterTest(UseThePlugin):
         self.assertExists(self.art_file)
 
     def test_delete_original_file(self):
-        self.plugin.src_removed = True
-        self._fetch_art(True)
-        self.assertNotExists(self.art_file)
+        prev_move = config["import"]["move"].get()
+        try:
+            config["import"]["move"] = True
+            self._fetch_art(True)
+            self.assertNotExists(self.art_file)
+        finally:
+            config["import"]["move"] = prev_move
 
     def test_do_not_delete_original_if_already_in_place(self):
         artdest = os.path.join(os.path.dirname(self.i.path), b"cover.jpg")
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -1253,10 +1253,6 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
         self.cautious = self.config["cautious"].get(bool)
         self.store_source = self.config["store_source"].get(bool)
 
-        self.src_removed = config["import"]["delete"].get(bool) or config[
-            "import"
-        ]["move"].get(bool)
-
         self.cover_format = self.config["cover_format"].get(
             confuse.Optional(str)
         )
@@ -1340,10 +1336,11 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
         """Place the discovered art in the filesystem."""
         if task in self.art_candidates:
             candidate = self.art_candidates.pop(task)
+            removal_enabled = _is_source_file_removal_enabled()
 
-            self._set_art(task.album, candidate, not self.src_removed)
+            self._set_art(task.album, candidate, not removal_enabled)
 
-            if self.src_removed:
+            if removal_enabled:
                 task.prune(candidate.path)
 
     # Manual album art fetching.
@@ -1443,3 +1440,9 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
                 else:
                     message = ui.colorize("text_error", "no art found")
                 self._log.info("{0}: {1}", album, message)
+
+
+def _is_source_file_removal_enabled():
+    delete_enabled = config["import"]["delete"].get(bool)
+    move_enabled = config["import"]["move"].get(bool)
+    return delete_enabled or move_enabled
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -1294,6 +1294,10 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
             for s, c in sources
         ]
 
+    @staticmethod
+    def _is_source_file_removal_enabled():
+        return config["import"]["delete"] or config["import"]["move"]
+
     # Asynchronous; after music is added to the library.
     def fetch_art(self, session, task):
         """Find art for the album being imported."""
@@ -1336,7 +1340,7 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
         """Place the discovered art in the filesystem."""
         if task in self.art_candidates:
             candidate = self.art_candidates.pop(task)
-            removal_enabled = _is_source_file_removal_enabled()
+            removal_enabled = FetchArtPlugin._is_source_file_removal_enabled()
 
             self._set_art(task.album, candidate, not removal_enabled)
 
@@ -1440,9 +1444,3 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
                 else:
                     message = ui.colorize("text_error", "no art found")
                 self._log.info("{0}: {1}", album, message)
-
-
-def _is_source_file_removal_enabled():
-    delete_enabled = config["import"]["delete"].get(bool)
-    move_enabled = config["import"]["move"].get(bool)
-    return delete_enabled or move_enabled
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -208,6 +208,8 @@ New features:
 * Add support for `barcode` field.
   :bug:`3172`
 * :doc:`/plugins/smartplaylist`: Add new config option `smartplaylist.fields`.
+* :doc:`/plugins/fetchart`: Defer source removal config option evaluation to
+  the point where they are used really, supporting temporary config changes.
 
 Bug fixes:
 


Serene-Arc (approved):

Serene-Arc (changes requested):
Whoops, wrong button there. Accidentally approved it.

There are failing tests. The code seems fine at first glance, but it's affected the tests.

mgoltzsche:
I adjusted the test correspondingly now. Please re-review.

test/plugins/test_art.py
```diff
@@ -825,9 +825,13 @@ def test_leave_original_file_in_place(self):
         self.assertExists(self.art_file)
 
     def test_delete_original_file(self):
-        self.plugin.src_removed = True
-        self._fetch_art(True)
-        self.assertNotExists(self.art_file)
+        prev_move = config["import"]["move"].get()
```
snejus, line 828: Instead of editing the config try patching it (or the relevant method in `fetchart.py`):

```python
from unittest.mock import patch
...

        with patch("beetsplug.fetchart.FetchArtPlugin.should_delete", True):
            ...
```

snejus (changes requested):

test/plugins/test_art.py
mgoltzsche: hm, I am not sure I follow you here:
1. I don't see a `should_delete` property or method within the `FetchArtPlugin` that could be mocked this way.
2. The point of the test is to ensure that another plugin that calls the fetchart plugin CAN temporarily overwrite the config effectively in the way the test does it. [This](https://github.com/mgoltzsche/beets-ytimport/blob/v1.8.1/beetsplug/ytimport/__init__.py#L194) is why I created the PR in the first place. Mocking that piece of logic away within the test makes the test pointless.

snejus: Thanks for linking up the reason why it's needed - this makes more sense now.

I'm concerned about the idea of a plugin overriding global settings, especially a setting like `move`. I'm not sure whether this would be the expected behaviour for users that have setups with `move: false`.

Meanwhile, I think I see why this may be required for your plugin which firstly __downloads__  and only then imports music.

If your plugin functionality completely depends on `move` being set to True, have you considered validating it instead?

```python
if not config["import"]["move"]:
    raise confuse.ConfigValueError("'move' must be set to True in order to use ytimport")
```

Ultimately, from `beets` implementation perspective, allowing global config adjustments in flight is an anti-pattern and in my opinion should be avoided.

mgoltzsche: > I'm not sure whether this would be the expected behaviour for users that have setups with move: false.

The user is not calling `beet import` but `beet ytimport` which has its own configuration (that it also applies to the import operation). Thus, the ytimport plugin overwriting some import parameters is not really unexpected imho.
Also, ytimport downloads into a temporary directory. As a user I don't want to switch my import configuration every time I import from a different source.

> Ultimately, from beets implementation perspective, allowing global config adjustments in flight is an anti-pattern and in my opinion should be avoided.

I agree it is not great. For instance if somebody wanted to run multiple imports in parallel they couldn't do that within a single Python process with different configuration each like this.
However, as of now there is no need to run multiple in parallel and the alternative is to start a separate Python process.

snejus: > The user is not calling `beet import` but `beet ytimport` which has its own configuration (that it also applies to the import operation). Thus, the ytimport plugin overwriting some import parameters is not really unexpected to the user imho. 

For me as a user this is unexpected in the following ways:

1. First and foremost - the fact that this command does not respect my global `move` configuration and **always** moves music into my library is not documented (didn't see it in the readme nor the command help).
2. I have no control of it (the command doesn't have `--move` nor `--no-move` flag)

Based on other commands that may move files, what I'm used to is
1. By default my global configuration being respected - if my configuration does not allow any filename adjustments, no adjustments are being made
2. Being able to override global configuration using the flag `--move`, `--write` etc.

My previous `beets` configuration had `move: no` and `copy: no` preventing any path adjustments during import. After the import, I did some additional processing on the files, and only then ran a command to move the files to my library directory. 


> As a user I don't want to edit/switch my import configuration every time I perform a different import operation.

I'm struggling to think of a use case where I'd have to switch the configuration, i.e. want to have `move: no` for the `import` command and `move: yes` for `ytimport`. I presume users may want the same configuration to apply for both commands consistently?

> ytimport downloads into and imports from a temporary directory that it manages itself

This makes sense. It sounds like once `ytimport` downloads a file to the temp directory, the situation is more or less equivalent to a user importing a local file manually. Did you consider delegating the import to the `beet import` command?

mgoltzsche: > First and foremost - the fact that this command does not respect my global move configuration and always moves music into my library is not documented (didn't see it in the readme nor the command help).

True but like I said: ytimport is importing from a temporary directory that it manages itself. In a way the fact that ihe ytimport plugin uses the beets' import plugin under the hood is more of an implementation detail.

> I have no control of it (the command doesn't have --move nor --no-move flag)

You kind of have that control: the ytimport plugin supports a `--no-import` option which makes it download the files but not import them yet. Calling `beet ytimport` imports all files within the temp download directory that weren't imported previously and that weren't skipped explicitly. Leaving the files in the temp dir after they were imported makes no sense to me. The additional postprocessing and renaming of files after the import that you mentioned sounds more of a hacky workaround to me than a well designed workflow.

> I'm struggling to think of a use case where I'd have to switch the configuration, i.e. want to have move: no for the import command and move: yes for ytimport. I presume users may want the same configuration to apply for both commands consistently?

Here's an example use-case: I am importing from another library. Because I don't want that to be a destructive operation, I just copy the files but leave the source files untouched. Because of the destructiveness of the move operation, I'd actually like it to be set to false in my [beets configuration](https://github.com/mgoltzsche/beets-container/blob/v0.18.1/config.yaml#L91) and just have the ytimport plugin enable it temporarily for its managed temp dir import.

> Did you consider delegating the import to the beet import command?

Yes, in fact it was like that initially before [I changed it](https://github.com/mgoltzsche/beets-ytimport/commit/3dfa195ab78e3c8afc557e6b781e546d304a411a#diff-d41b7d7f5b3c086c955a2710191b21ba0db2e3aa785a71cf31d9ec1f59f20100) to using the Python API as opposed to the CLI. This was to avoid spawning a new Python process for the import. Apart from the unnecessary overhead, spawning a new Python process comes with the disadvantage that global beets CLI options are not applied to the beets child process oob and would each have to be passed through to the child process explicitly, covering all beets options exhaustively, ideally. E.g. when beets' config file option is not passed through, it can result in the child process using a different beets configuration than its parent process unexpectedly. Even with the option being passed through the configuration could change during execution of the `ytimport` command since the child process reloads it. To give another example: the log verbosity (`-vvv`) of the parent process is not applied to the child process oob. In this matter it is beneficial to be able to call a Python API from within the same Python process as opposed to having to call a CLI, spawning a new process.
After all why is beets' `import_file` function public when it shouldn't be called by 3rd party Python code? Now that it is public, why would you still want to force users to call the CLI and be fine with the CLI overwriting the `--move` option but not allow doing that via the Python API (especially given the fact that this already works with all the other config options but the `move` option)?

snejus: > True but like I said: ytimport is importing from a temporary directory that it manages itself.

In addition to this it also by default manages my library directory even when beets is configured not to do so.

> The additional postprocessing and renaming of files after the import that you mentioned sounds more of a hacky workaround than a well designed/clean workflow to me.

It may have been hacky, but that's the beauty of beets with the flexibility it provides.

> Yes, in fact it was like that initially before [I changed it](https://github.com/mgoltzsche/beets-ytimport/commit/3dfa195ab78e3c8afc557e6b781e546d304a411a#diff-d41b7d7f5b3c086c955a2710191b21ba0db2e3aa785a71cf31d9ec1f59f20100) to using the Python API as opposed to the CLI. This was to avoid spawning a new Python process for the import. Apart from the unnecessary overhead, spawning a new Python process comes with the disadvantage that global beets CLI options are not applied to the beets child process oob and would each have to be passed through to the child process explicitly, covering all beets options exhaustively, ideally. E.g. when beets' config file option is not passed through, it can result in the child process using a different beets configuration than its parent process unexpectedly. Even with the option being passed through the configuration could change during execution of the `ytimport` command since the child process reloads it. To give another example: the log verbosity (`-vvv`) of the parent process is not applied to the child process oob. In this matter it is beneficial to be able to call a Python API from within the same Python process as opposed to having to call a CLI, spawning a new process.
> 
> After all why is beets' `import_files` method public when it shouldn't be called by 3rd party Python code? Now that it is public, why would you still want to force users to call the CLI and be fine with the CLI overwriting the `--move` option but not allow doing the same via the Python API (especially given the fact that this already works with all the other config options but the `move` option)?

Sorry, I should have clarified that I meant using the Python API, but instead of the `import_files` function, using the import `Subcommand` (I think it's found in `beets.commands` as `import_cmd` variable) and extending it with flags custom to `ytimport`. This would ensure consistent import behaviour.

I imagine you'd also want to adjust `import_cmd.func` which runs `beets.commands::import_func` to something like

```python
def ytimport_func(lib, opts, args):
    download(...)
    # add the files to args, I guess
    ...
    import_func(...)
```

mgoltzsche: How is calling the [`import_func`](https://github.com/beetbox/beets/blob/0a985fdfb4add8eeb5d2bb73233a78057e690745/beets/ui/commands.py#L1337-L1400) better than calling the [`import_files` function](https://github.com/beetbox/beets/blob/0a985fdfb4add8eeb5d2bb73233a78057e690745/beets/ui/commands.py#L1305) that the former function delegates to? I think the approach you're suggesting is disadvantageous: It would require the ytimport plugin to write all options it needs to overwrite into the opts dict just to have the `import_func` copy it into the config as opposed to letting the ytimport plugin do that directly. The `import_func` function also doesn't treat these config changes as temporary but permanent ones, still requiring the ytimport plugin to restore the original configuration afterwards which is less obvious to maintain and more error-prone when it does not make the config changes directly but via the `opts` but then still has to access the config directly to restore its original values. (The latter is actually optional since usually no other code/plugin is run within the same Python process after the `ytimport` CLI command terminates but it leaves the door open for another 3rd party plugin to call the ytimport plugin via its Python API directly and run additional code afterwards within the same Python process based on the original user-specified configuration.)

Also, the problem this PR is solving still exists in the approach you're describing (the global config is modified after the fetchart plugin was initialized) and as I understand it also exists within Beets internally when you consider that e.g. the import plugin [overwrites the `move` option](https://github.com/beetbox/beets/blob/0a985fdfb4add8eeb5d2bb73233a78057e690745/beets/ui/commands.py#L1343) explicitly after the fetchart plugin was already initialized with the original/previous value of that option which results in the plugin ignoring the overwritten value.

snejus: > It would require the ytimport plugin to write all options it needs to overwrite into the opts dict just to have the import_func copy it into the config as opposed to letting the ytimport plugin do that directly. 

Given that these options are only relevant for the import, and importing is delegated to beets, why should it be partly reimplemented in ytimport? I am assuming that `ytimport` does not overwrite any import-specific options but it has flags `--import` and `--no-import` to control whether the `import_func` is called.

> The import_func function also doesn't treat these config changes as temporary but permanent ones, still requiring the ytimport plugin to restore the original configuration afterwards which is less obvious to maintain and more error-prone

The global config object is updated this way since the flags override what is specified in the configuration file. This way plugins that run during the import respect whatever is configured by the flags.

> CLI command terminates but it leaves the door open for another 3rd party plugin to call the ytimport plugin via its Python API directly and run additional code afterwards within the same Python process based on the original user-specified configuration.

Let's consider the example I used above

```python
def ytimport_func(lib, opts, args):
    download(...)
    # add the files to args, I guess
    ...
    import_func(...)
```

Assuming that the global configuration is adjusted by `import_func` only, the case you described can only happen if the 3rd party plugin extends `ytimport` command with some functionality that runs _after_ the import is complete. Following my suggestion, we would have a command like

```
beet post-ytimport --help

...

Options:
  ...
  -m, --move ...
  ... the rest of beet import options
  ... beet ytimport options
  ... beet post-ytimport options
```

If it happens to depend on `import` configuration, I believe it is a good idea to **keep** the configuration that `import_func` overrides, which would happen if the user runs `beet post-ytimport --move`, for example.

Realistically, though, I think, it is more likely that a 3rd party plugin is going to be more interested in the functionality specific to your plugin: the `download` function in my example above (everything except the import).


> Also, the problem this PR is solving still exists in the approach you're describing (the global config is modified after the fetchart plugin was initialized) and as I understand it also exists within Beets internally when you look at e.g. the import plugin overwriting the move option explicitly after the fetchart plugin was already initialized with the original/previous value of that option, resulting in the fetchart plugin not honouring the new value.

Have just checked this - you're completely right. Fetchart has essentially been using the default configuration all this time.

I'm going to approve this in a second, just going to quickly scan whether this issue has been reported before and add them to the PR.

Thanks for your patience!

snejus (approved):

PR #5266, opened 2024-05-29
Replace setup.py packaging by Poetry

snejus:
Migrate `beets` package configuration to Poetry which nowadays seems to be the gold standard.

I have been using Poetry since 2019 and I have mostly been happy a happy user: it makes local dev setup easy and has the tools I need to maintain python packages day to day, including reliable dependency resolution, versioning and publishing to Pypi. 

It's a user-friendly tool, so it should make it more straightforward for contributors to setup and navigate the codebase, and ultimately, hopefully facilitate more frequent releases!

Since poetry manages local virtual environment, we do not have much need for tox any more. Therefore, it was replaced by a task runner `poethepoet`. Type `poe` in the project directory to see the available commands.


Serene-Arc:
I'm fine with this change to poetry in theory, once all the tests and linting errors have been fixed, but a couple of things: 

- What are the purposes of the changes to `release.py`? A lot of the things that I removed seem to have been added back in, some of which are definitely not supposed to be there?
- What are the purposes of changing the tox file, specifically calling binaries instead of `python -m black` for instance?
- There are some issues with changing version in the release workflow; does this affect any of that?

snejus:
> I'm fine with this change to poetry in theory, once all the tests and linting errors have been fixed, but a couple of things:
> 
> * What are the purposes of the changes to `release.py`? A lot of the things that I removed seem to have been added back in, some of which are definitely not supposed to be there?

These commits have been made before I noticed that `release.py` has been adjusted on `master` - it needs rebasing where it will get removed :)

> * What are the purposes of changing the tox file, specifically calling binaries instead of `python -m black` for instance?

Some of the changes have to do with `tox` upgrade to `4.0.0`. Regarding calling the binaries directly - that's the accepted standard, given that every command will be run within the `tox` environment, where it will be resolved automatically. It is slightly shorter as well.

> * There are some issues with changing version in the release workflow; does this affect any of that?

When I made changes here, the release workflow did not exist yet, so I made adjustments to `release.py` instead. When I rebase, I will make sure to update the release workflow instead.

Serene-Arc:
> These commits have been made before I noticed that release.py has been adjusted on master - it needs rebasing where it will get removed :)

Fair enough!

> Some of the changes have to do with tox upgrade to 4.0.0. Regarding calling the binaries directly - that's the accepted standard, given that every command will be run within the tox environment, where it will be resolved automatically. It is slightly shorter as well.

Great, was just worried with path stuff, but if it's only for the tox environment internally, makes sense.

> When I made changes here, the release workflow did not exist yet, so I made adjustments to release.py instead. When I rebase, I will make sure to update the release workflow instead.

Cool.

With regards to the `release.py` changes, do you want to keep them here or move them to #5274? It might make more sense to get the release workflow working for the current project before doing a big tool change like poetry.

snejus:
> do you want to keep them here or move them to https://github.com/beetbox/beets/pull/5274? It might make more sense to get the release workflow working for the current project before doing a big tool change like poetry.

I'm completely with you here @Serene-Arc - I'm moving the relevant commits over.

Serene-Arc:
@snejus I made some changes to the actions. Everything else looks good so if you could review the changes, I think it's ready to merge!

snejus:
Just noticed that neither of the workflows managed to run due to some issues: 
image

.github/workflows/ci.yaml
```diff
@@ -5,48 +5,50 @@ on:
...
-      - uses: actions/setup-python@v5
+      - name: Install Python
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-          cache: poetry
-
-      - name: Install PyGobject dependencies
-        if: matrix.platform == 'ubuntu-latest'
+          cache: 'poetry'
```
snejus, line 27: This unfortunately does not work, since `poetry` needs to be installed before `actions/setup-python@v5` is pulled in. That's why there was `pipx-install-action` which sets up both `poetry` and `poe`.

.github/workflows/ci.yaml
```diff
@@ -5,48 +5,50 @@ on:
...
+      
+      - name: Install PyGobject dependencies on Ubuntu
+        if: startsWith(matrix.platform, 'ubuntu-')
```
snejus, line 30: I think we would want to keep `if: matrix.platform == 'ubuntu-latest'` here since it's more explicit, and we only have a single option

.github/workflows/ci.yaml
```diff
@@ -5,48 +5,50 @@ on:
     branches:
       - master
 env:
-  PYTHON_VERSION: 3.8
+  MIN_PYTHON_VERSION: 3.8
   PY_COLORS: 1
+  IS_MAIN_PYTHON: ${{ matrix.python-version == env.MIN_PYTHON_VERSION && matrix.platform == 'ubuntu-latest' }}
```
snejus, line 10: I like how you removed the duplication by defining it in the environment!

Though, I think it may fail here since `matrix` context does not exist in the global environment. We will need to define this in the job environment instead

.github/workflows/ci.yaml
```diff
@@ -5,48 +5,50 @@ on:
...
 
-      - if: matrix.python-version == env.PYTHON_VERSION && matrix.platform == 'ubuntu-latest'
-        name: Upload code coverage
+      - name: Run tests
+        run: |
+          if [ "${{ env.IS_MAIN_PYTHON }}" = "true" ]; then
+            poe test
+          else
+            poe test --no-cov
+          fi
```
snejus: A couple of reasons why it's preferable to have this logic as two separate steps

```yaml
  - if: ! env.IS_MAIN_PYTHON
    name: Test without coverage
    run: poe test --no-cov

  - if: env.IS_MAIN_PYTHON
    name: Test with coverage
    uses: liskin/gh-problem-matcher-wrap@v3
    with:
      linters: pytest
      run: poe test
```

1. More explicit: name of the step communicates everything and there's no need to read the implementation. This makes maintenance easier and is generally easier for devs to deal with
1. This uses GitHub Actions functionality, and thus we do not need add any custom shell code which would again add to the required maintenance
1. GitHub nicely displays which steps have been skipped 
image
image
image

snejus (commented):
I love that you caught the issue with changed docs files and deduplicate some things in `ci.yaml`!

I see a great attempt at DRY-ing up the jobs linting workflow, unfortunately GitHub does not support such design (I really wish it did!).

I've noticed that you made adjustments to how `poetry` and `poe` are set up and it broke the workflows - this is completely my bad for not documenting this properly and making it clear why `poetry` needs to be installed before `setup-python`, and what's the intention behind `poe`. 

Regarding the tests setup in `ci.yaml`, I wanted to hear a bit more reasoning behind the adjustments (for example replacing  explicit `Test with coverage` and `Test without coverage` steps by a single one which has the condition written in Shell).

Serene-Arc:
Yeah I didn't do full testing on the actions yet, since they're very finnicky and rely a lot on what repo they're running on. Wanted to see if you're happy with the changes before I spend that extra time x.x

Serene-Arc:
Thanks for fixing up my code! 

The one thing I can see is that you've regressed the commit changing the paths of the format commands, format and the format-check. I still think that they should be made explicit as to what directories to change. I think it'll make the project less likely to somehow get non-standard code added by mistake and it will mean that virtual environments don't get arbitrarily formatted. Perhaps adding the docs and `extra/release.py` to the previous folders would work?

Other than that, I think it's good to be merged.

snejus:
> Thanks for fixing up my code!
> 
> The one thing I can see is that you've regressed the commit changing the paths of the format commands, format and the format-check. I still think that they should be made explicit as to what directories to change. I think it'll make the project less likely to somehow get non-standard code added by mistake and it will mean that virtual environments don't get arbitrarily formatted. Perhaps adding the docs and `extra/release.py` to the previous folders would work?
> 
> Other than that, I think it's good to be merged.

Unfortunately neither `black` nor `isort` has configuration for __default__ paths to check that could be overridden by command-line arguments, like `mypy` has. This means that if we hardcode `beets`, `beetsplug` etc. these tools will _always_ run across the entire codebase, and thus we won't be able to check the changed files _only_.

You're making a good point regarding virtual environments and other unwanted files - luckily both `black` and `isort` by default ignores everything defined in `.gitignore`. Since our `.gitignore` has `.venv` and `venv`, those files should not get formatted 👍🏼 if we find that locally something unexpected is getting reformatted, we can just add those paths to `.gitignore` too since they should most likely be ignored by `git` as well.

Serene-Arc:
I'm not sure that checking against the changed files only is the best way to do this. Like when we did the release, it is technically possible for unformatted code to make its way into the codebase, however that happens. Currently, any code will always be found by the next PR because it'll come up with a change to the file or a format error in the CI. If we make it so only changed files are checked, this code can be missed for who-knows-how-long.

Maybe it's paranoid but I prefer the formatting tools checking the entirety of the codebase every time. It makes sure that the formatting rules are actually being enforced at every stage possible.

snejus:
Thanks for clarifying @Serene-Arc, I now see what you mean! Given that the code base is already neat and formatted, I agree that checking everything every time makes more sense! I'm adjusting it now.

I think the __only-check-what-has-changed__ approach makes more sense when the code base __is not yet formatted__, and is getting tidied up iteratively (if you adjust a file, you must format it properly), which is what we did in most cases in my previous experience. 😅

snejus:
As you predicted @Serene-Arc, there was indeed an unformatted file `docs/conf.py` that would have been missed! It's fixed now, thanks for your persistence regarding this change! <3

Serene-Arc:
Wonderful! That's all my concerns fixed then, merge whenever you'd like.

Serene-Arc (approved):

PR #5272, opened 2024-05-30
Add missed `HAVE_REFLINK` in test_files.py

catap:
## Description

A trivial changes to skip test which requires reflink.


Serene-Arc:
Changelog isn't required, and the formatting issue isn't your fault, so this seems good to me! Thanks for the contribution.

PR #5277, opened 2024-06-01
Add aliases support

sezaru:
## Description

This PR add support for aliases to releases, release-groups and recordings.

This PR is a must have (IMO at least) for people that listen to Japanese, Chinese and other songs that has other symbols for letters. With this, not only the artist name will use the alias if available, but now the album and track name will also use aliases.

NOTE: This PR requires a fork of the `python-musicbrainzngs` package. That package seems to be dead (more than 2 years without any commit) and is lacking in a lot of updates to Musicbrainz API.

Here is my PR adding support for the new fields https://github.com/alastair/python-musicbrainzngs/pull/289 but I have no hope that will be merged since the package seems to be dead.

Personally I think beets can create a fork of that package and just add the bare minimum to it to make new features work (for example this one).

I didn't add any tests and changelog changes yet since this PR requires forking `python-musicbrainzngs`, so first I want to discuss if this can be merged and then after getting the confirmation I will add the missing parts.

### DIFF
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -563,7 +563,17 @@ def album_info(release: Dict) -> beets.autotag.hooks.AlbumInfo:
     info.albumstatus = release.get("status")
 
     if release["release-group"].get("title"):
-        info.release_group_title = release["release-group"].get("title")
+        if "alias-list" in release["release-group"]:
+            alias = _preferred_alias(release["release-group"]["alias-list"])
+        else:
+            alias = None
+
+        if alias:
+            title = alias["alias"]
+        else:
+            title = release["release-group"].get("title")
+
+        info.release_group_title = title
 
     # Get the disambiguation strings at the release and release group level.
     if release["release-group"].get("disambiguation"):
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -527,8 +527,19 @@ def album_info(release: Dict) -> beets.autotag.hooks.AlbumInfo:
             track_infos.append(ti)
 
     album_artist_ids = _artist_ids(release["artist-credit"])
+
+    if "alias-list" in release:
+        alias = _preferred_alias(release["alias-list"])
+    else:
+        alias = None
+
+    if alias:
+        title = alias["alias"]
+    else:
+        title = release["title"]
+
     info = beets.autotag.hooks.AlbumInfo(
-        album=release["title"],
+        album=title,
         album_id=release["id"],
         artist=artist_name,
         artist_id=album_artist_ids[0],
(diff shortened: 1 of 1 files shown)


Serene-Arc:
I would really like this functionality. However, there is #4936, which completely replaces the `musicbrainzngs` package and instead embeds that functionality in beets itself. That seems like the best option, given that the other project is more or less dead. 

I would ask you to consider looking at that PR and how you can use it, since I will begin reviewing it now.

Morikko:
> I would really like this functionality. However, there is https://github.com/beetbox/beets/pull/4936, which completely replaces the musicbrainzngs package

The replacement is stale for quite a while now. And because of the size of the change and impact, it requires a lot of developers and beets maintainers availability. Pragmatically, It can not be expected to happen soon.  

> Personally I think beets can create a fork of that package and just add the bare minimum to it to make new features work (for example this one).

I tend to agree with that statement, this and maybe other propositions may continue to improve beets in the meantime without being blocked by uncertain deep refactoring.

As long as new propositions are tested, it should mitigate the risk of regression that the new musicbrain binding change could be subjected to.

Could it be reconsidered ?

beets/autotag/mb.py
```diff
@@ -288,8 +288,18 @@ def track_info(
     ``medium_index``, the track's index on its medium; ``medium_total``,
     the number of tracks on the medium. Each number is a 1-based index.
     """
+    if "alias-list" in recording:
```
snejus, line 291: This is duplicated in 4 places. Could you define a function to deduplicate it?

Morikko:
I don't know if @sezaru plans to resume his work else I can continue it. But before, there is this requirement that needs to be sorted:

> Here is my PR adding support for the new fields https://github.com/alastair/python-musicbrainzngs/pull/289 but I have no hope that will be merged since the package seems to be dead.

> Personally I think beets can create a fork of that package and just add the bare minimum to it to make new features work

Would the beets maintainers agree to fork it to accept MR to extend beets features ?

snejus:
@Morikko, this may be slightly complicated. We won't be able to publish a new version of the package and will have to depend on the GitHub URL of the fork, which is not ideal.

MusicBrainz have a simple JSON API, I wonder whether we could remove this dependency and talk to MB directly. Let me see how much effort would this require.

snejus:
Here we go: https://github.com/beetbox/beets/pull/6052

Morikko:
@snejus I have made a new PR (https://github.com/beetbox/beets/pull/6231) based on your changes (https://github.com/beetbox/beets/pull/6052). Great changes by the way :clap:

snejus:
@Morikko great, thank you! I will have a look at this soon. Closing this one a it's been superseded by https://github.com/beetbox/beets/pull/6231

PR #5282, opened 2024-06-03
Resolve some 'mypy' errors

bal-e:
## Description

I noticed that some development work was waiting on cleaner `mypy` runs, so I thought I would help out a bit.  This PR just contains a number of miscellaneous fixes across the codebase that I hit.  I'm happy to add more, or remove any problematic changes (though I don't think I've broken anything, only type annotations have changed).

## To Do

- [x] ~Documentation~
- [ ] Changelog
- [x] ~Tests~

### DIFF
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -238,7 +238,7 @@ def distance(
     if album_info.media:
         # Preferred media options.
         patterns = config["match"]["preferred"]["media"].as_str_seq()
-        patterns = cast(Sequence, patterns)
+        patterns = cast(Sequence[str], patterns)
         options = [re.compile(r"(\d+x)?(%s)" % pat, re.I) for pat in patterns]
         if options:
             dist.add_priority("media", album_info.media, options)
@@ -276,7 +276,7 @@ def distance(
 
     # Preferred countries.
     patterns = config["match"]["preferred"]["countries"].as_str_seq()
-    patterns = cast(Sequence, patterns)
+    patterns = cast(Sequence[str], patterns)
     options = [re.compile(pat, re.I) for pat in patterns]
     if album_info.country and options:
         dist.add_priority("country", album_info.country, options)
@@ -440,7 +440,9 @@ def _add_candidate(
         return
 
     # Discard matches without required tags.
-    for req_tag in cast(Sequence, config["match"]["required"].as_str_seq()):
+    for req_tag in cast(
+        Sequence[str], config["match"]["required"].as_str_seq()
+    ):
         if getattr(info, req_tag) is None:
             log.debug("Ignored. Missing required tag: {0}", req_tag)
             return
@@ -469,7 +471,7 @@ def tag_album(
     items,
     search_artist: Optional[str] = None,
     search_album: Optional[str] = None,
-    search_ids: List = [],
+    search_ids: List[str] = [],
 ) -> Tuple[str, str, Proposal]:
     """Return a tuple of the current artist name, the current album
     name, and a `Proposal` containing `AlbumMatch` candidates.
@@ -561,7 +563,7 @@ def tag_item(
     item,
     search_artist: Optional[str] = None,
     search_title: Optional[str] = None,
-    search_ids: List = [],
+    search_ids: Optional[List[str]] = None,
 ) -> Proposal:
     """Find metadata for a single track. Return a `Proposal` consisting
     of `TrackMatch` objects.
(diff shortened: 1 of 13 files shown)


bal-e:
Updated to account for merge conflicts.

bal-e:
Many remaining mypy errors are from `util/__init__.py`, but I'm not touching it right now since there are other PRs addressing it.  The remaining type errors mostly stem from APIs that allow for arbitrary-type data storage, such as the database and `AlbumInfo` / `TrackInfo`.  For the latter two, some refactoring could eliminate type errors (organizing expected fields into new classes with strongly-typed attributes and turning `AttrDict` into a container object managing those classes).  I'm not sure how to eliminate mypy errors related to `dbcore` and the query subsystems.

beets/library.py
```diff
@@ -49,7 +50,7 @@
 # Library-specific query types.
 
 
-class SingletonQuery(dbcore.FieldQuery):
+class SingletonQuery(dbcore.FieldQuery[NoReturn]):
```
snejus, line 53: Shouldn't this rather be `str`?

It's a weird one: I've written this class defining `__new__` method with the same parameters like `__init__` has in the base class. It's essentially a 'factory' class that returns an instance of `NotQuery`.

`mypy` won't be able to map the defined to the parameter in `__new__` definition, but since we expect it to have a string there, it may be worth defining it so?

snejus (commented):
Love this, thank you. Just one small comment

beets/autotag/match.py
```diff
@@ -561,7 +563,7 @@ def tag_item(
     item,
     search_artist: Optional[str] = None,
     search_title: Optional[str] = None,
-    search_ids: List = [],
+    search_ids: Optional[List[str]] = None,
```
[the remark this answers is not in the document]
bal-e: Yeah, I think the `[t for t in [item.mb_trackid] if t]` looks ridiculous, but I couldn't figure out a better way to express it that concisely.

beets/library.py
bal-e: Oh, I didn't realize that `SingletonQuery` can be used to search for `album_id is None` as well as `album_id is not None` (and even then, `()` would have been a better fit than `NoReturn`, I suppose).  Since we're just calling `util.str2bool()` on the provided pattern, maybe we should just take a `bool` there.  What do you think?

bal-e: It looks like `SingletonQuery` gets pulled in to `beets.dbcore.db.Model._queries` (of type `Dict[str, Type[FieldQuery[Any]]]`), with the documentation "Named queries that use a field-like `name:value` syntax".  In that case, I understand why `SingletonQuery` would take `str` instead of `bool` -- but shouldn't every query in `_queries` also be of type `FieldQuery[str]`?  Is there any reason for the pattern to be a generic parameter at all?

beets/dbcore/types.py
```diff
@@ -232,7 +232,7 @@ class BaseFloat(Type[float, N]):
     """
 
     sql = "REAL"
-    query = NumericQuery
+    query: typing.Type[FieldQuery[Any]] = NumericQuery
```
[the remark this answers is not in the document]
snejus: no, it's all good! I just like the wordplay here 😄

beets/library.py
snejus: > Since we're just calling `util.str2bool()` on the provided pattern, maybe we should just take a `bool` there. What do you think?

For this exact reason, I thought `str` would be more suitable, given that it's the string input that is being converted to boolean?

snejus: > but shouldn't every query in _queries also be of type FieldQuery[str]? Is there any reason for the pattern to be a generic parameter at all?

The `FieldQuery` being constructed from the input (string) query is only one of the uses cases - in other cases it's actually initialised with the correct type, so it needs to be generic, as far as I remember.

bal-e: Alright, I see.  I'm going to keep it as `str` then.  However, I am concerned with the use of `util.str2bool()`: at some point, it should be changed to only accept unambiguously positive or negative answers (i.e. it should be able to fail), to avoid confusion for users, but that's out of scope for this PR.

bal-e:
I don't think this PR needs a changelog entry, but please confirm if so.

snejus:
Nah I don't think the changelog entry is needed!

snejus (approved):
Looks great, thank you!

PR #5283, opened 2024-06-05
Upgrade to Python 3.8 / drop Python 3.7

### LINKED ISSUE #5222: We should drop Python 3.7
In #5215, our minimum supported Python version came up: We currently state 3.7, but that's been EOL since June 2023. I think we should drop that and allow 3.8+ features. Any objections?

snejus:
## Description

Fixes #5222.  

Drop Python 3.7. `pyupgrade` is responsible for most of the changes in the code. I undid some of the bits it attempted to update that aren't strictly necessary:

1. Converting `List/Dict/Tuple` -> `list/dict/tuple` in modules that have `from __future__ import annotations` import. This should be done in a separate PR, and for all modules
2. Converting some `.format(` calls to f-strings. It didn't do it consistently, and it should also be done in a separate PR, I believe.

Python upgrade unblocks several other PRs, for example #5266 and #5248.

## To Do

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, windows-latest]
-        python-version: ['3.7', '3.8', '3.9', '3.x']
+        python-version: ["3.8", "3.9", "3.x"]
 
     env:
       PY_COLORS: 1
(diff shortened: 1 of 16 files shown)


Serene-Arc (approved):
Changes seem good to me. Nice that 3.8 has its own lazy property functionality in functools.

PR #5288, opened 2024-06-07
Fix `reflink: "auto"` option

lnikkila:
## Description

The docs say:

> The `auto` option uses reflinks when possible and falls back to plain
> copying when necessary.

I've been using this option for a while, and recently discovered that despite the option, copying fails between two BTRFS filesystems with:

    Error: OS/filesystem does not support reflinks. during link of paths /mnt/fs1/file, /mnt/fs2/file

I tracked this down to how the configuration is handled in the importer.

I love beets dearly, thanks for your great work :)

## To Do


Serene-Arc:
Thank you for your contribution. It looks great to me.

PR #5300, opened 2024-06-12
Feature/mbsubmit create release

jonaswinkler:
## Description

Implement support to create new releases on musicbrainz from unmatched albums directly.

1. When the new option is chosen, the MB release editor is opened in the browser, with known metadata already filled in. The plugin must launch an internal web server to do that.
2. When the release is saved / created on musicbrainz, musicbrainz redirects to the local web server, submitting the new release ID. beets proceeds to import the album with that ID, similar to what happens when an ID is provided manually.

See documentation and comments for details.

## To Do

- [ ] All todo marks still left in the code.
- [ ] Add some more key comments to the code.


beetsplug/mbsubmit.py
```diff
@@ -79,6 +407,36 @@ def print_tracks(self, session, task):
...
+        elif self.server_open_method == "show_link":
+            print_(f"Open the following URL in your browser: {url}")
+        else:
+            return
+
+        print_("Waiting for MusicBrainz release ID...")
+
+        while (mbid := self.server_task_results.get(task_key, None)) is None:
+            time.sleep(0.1)
+            # TODO need a better way to do this
```
jonaswinkler, line 435: Not sure what to do here!

I want to wait until the release_mbid is available (by saving the release on musicbrainz). This works. What also works is creating a threading.Event(), triggering that in the RequestHandler, and waiting here.

BUT: In both cases the user can't exit with CTRL+c, which is bad...

Don't see a way to catch KeyboardInterrupt here either, since that's handled way further up... Can I somehow check if the current thread is aborted and should exit?

jonaswinkler: Solved.

jonaswinkler:
Hey @doronbehar, If you're around, have a look at this. Maybe I can remove the "Open with Picard" option now that this here is implemented?

beetsplug/mbsubmit.py
```diff
@@ -18,31 +18,334 @@
...
+
+    result_release_mbid: Optional[str] = None
+    """
+    Contains the release ID returned by MusicBrainz after the release was created.
+    """
+
+    browser_opened: bool = False
+    """
+    True when the user has opened the link for this task in the browser.
+    """
```
Serene-Arc, line 105: Docstring comments go above variables. It's only classes and functions that have them below the definition. Confusing, I know

docs/plugins/mbsubmit.rst
```diff
@@ -9,11 +9,24 @@ that is parseable by MusicBrainz's `track parser`_. The prompt choices are:
...
   you to start submitting the unmatched release to MusicBrainz with many input
   fields already filled in, thanks to Picard reading the preexisting tags of
   the files.
 
+To create new releases on MusicBrainz with this plugin you need to install the
+`PyJWT`_ library with:
+
+.. code-block:: console
+
+   $ pip install beets[mbsubmit]
```
Serene-Arc, line 26: With the switch to poetry, this is now outdated. Should be changed to the poetry equivalent.

beetsplug/mbsubmit.py
```diff
@@ -18,31 +18,334 @@
...
+def join_phrase(i, total):
+    if i < total - 2:
+        return ", "
+    elif i < total - 1:
+        return " & "
+    else:
+        return ""
+
+
+def build_formdata(items: list, redirect_uri: Optional[str]):
```
Serene-Arc, line 223: Your use of typings is inconsistent. Would you consider typing all of these functions? Both the parameters and return types?

Serene-Arc (changes requested):
Thanks for the PR! I've made a couple comments but I'll also note that a fair amount of the changes in this PR are fairly boilerplate HTTP webserver stuff. Have you considered using something like [simple-http-server](https://pypi.org/project/simple-http-server/) or another package to deal with this? It would reduce the number of lines in the PR and make it easier to change in the future.

jonaswinkler:
> Thanks for the PR! I've made a couple comments but I'll also note that a fair amount of the changes in this PR are fairly boilerplate HTTP webserver stuff. Have you considered using something like [simple-http-server](https://pypi.org/project/simple-http-server/) or another package to deal with this? It would reduce the number of lines in the PR and make it easier to change in the future.

I thought about not pulling int too many new dependencies. I can also use flask instead, since that's already used by a couple other plugins...?

Serene-Arc:
I don't think many dependencies is an issue that is too pressing. If it makes the code easier to read and maintain, that wins out over another dependency in my book. If flask is used by other plugins though, by all means use it!

docs/plugins/mbsubmit.rst
jonaswinkler: Is it?

https://github.com/beetbox/beets/blob/master/docs/plugins/fetchart.rst?plain=1#L12

I can add quotation marks for consistency.

beetsplug/mbsubmit.py
jonaswinkler: Do they? See https://peps.python.org/pep-0258/#attribute-docstrings... also, PyCharm correctly picks up the docstrings this way.

jonaswinkler:
Thank you for the review! It's now much cleaner and shorter with flask.

doronbehar:
> Hey @doronbehar, If you're around, have a look at this. Maybe I can remove the "Open with Picard" option now that this here is implemented?

Hey! Personally I won't mind it getting removed, as I'd definitely use this functionality once it will land! It also makes sense to me from a general point of view to remove it, but that needs an approval from the maintainers of beets.

Thanks for working on this feature!

beetsplug/mbsubmit.py
```diff
@@ -18,31 +18,134 @@
...
+
+    browser_opened: bool = False
+    """
+    True when the user has opened the link for this task in the browser.
+    """
+
+    result_release_mbid: Optional[str] = None
+    """
+    Contains the release ID returned by MusicBrainz after the release was created.
+    """
```
Serene-Arc, line 96: Comments go above variables, please.

beetsplug/mbsubmit.py
```diff
@@ -57,14 +160,250 @@ def __init__(self):
...
+    def _stop_server(self):
+        if self._server:
+            self._server.close()
+            self._server = None
+            self._server_port = None
+
+    def _wait_for_condition(self, condition: Callable):
+        t = threading.current_thread()
+        while not condition():
+            time.sleep(0.5)
```
Serene-Arc, line 378: Is this necessary? This is a blocking call on the other thread, so is there any penalty to removing it?

Serene-Arc (changes requested):

snejus:
Closing due to lack of activity

PR #5302, opened 2024-06-14
Make the release send a toot on Fosstodon

Serene-Arc:
This adds a step to the GitHub release so that it sends a toot on Fosstodon, beets' Mastodon account, to publicise that and get some engagement!

Not tested, but this is lifted straight from the action repo so it *should* be okay. @snejus for a sanity check if you wouldn't mind

### DIFF
--- a/.github/workflows/make_release.yaml
+++ b/.github/workflows/make_release.yaml
@@ -73,12 +73,19 @@ jobs:
           name: python-package-distributions
           path: dist/
       - name: Create a GitHub release
+        id: make_release
         uses: ncipollo/release-action@v1
         with:
           tag: ${{ steps.tag_version.outputs.new_tag }}
           name: Release ${{ steps.tag_version.outputs.new_tag }}
           body: ${{ env.CHANGELOG }}
           artifacts: dist/*
+      - name: Send release toot to Fosstodon
+        uses: cbrgm/mastodon-github-action@v2
+        with:
+          access-token: ${{ secrets.MASTODON_ACCESS_TOKEN }}
+          url: ${{ secrets.MASTODON_URL }}
+          message: "Version v${{ inputs.version }} of beets has been released! Check out all of the new changes at ${{ steps.create_release.outputs.html_url }}"
 
   publish_to_pypi:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/make_release.yaml
```diff
@@ -73,12 +73,19 @@ jobs:
...
           tag: ${{ steps.tag_version.outputs.new_tag }}
           name: Release ${{ steps.tag_version.outputs.new_tag }}
           body: ${{ env.CHANGELOG }}
           artifacts: dist/*
+      - name: Send release toot to Fosstodon
+        uses: cbrgm/mastodon-github-action@v2
+        with:
+          access-token: ${{ secrets.MASTODON_ACCESS_TOKEN }}
+          url: ${{ secrets.MASTODON_URL }}
+          message: "Version v${{ inputs.version }} of beets has been released! Check out all of the new changes at ${{ steps.create_release.outputs.html_url }}"
```
snejus, line 88: Just a tiny thing but I think you can use this output here

```suggestion
          message: "Version ${{ steps.tag_version.outputs.new_tag }} of beets has been released! Check out all of the new changes at ${{ steps.create_release.outputs.html_url }}"
```

Serene-Arc:
Great! committed that change, do you think it's ready to merge?

bal-e:
Just curious: who owns the `beets` Mastodon account?  Does a single person have access to the keys or is it just available to the `beetbox` organization as a whole?

Serene-Arc:
We don't have any social media organisation software so Adrian just shared the email and password with me, and I entered the token into the repository secrets.

bal-e:
Right, good to know!  After the xz fiasco, it just feels more important to have this kind of information available for any open-source software with a team of contributors.  But I don't think that is something to worry about for beets.

Serene-Arc:
Yeah it's not ideal but xz is hopefully an exception. It's not a good idea for projects to be infiltrated but honestly that kind of thing is extraordinary, which is why it got the attention it did.

To use myself as an example, I've probably got the most power over this project in administrative terms other than Adrian, since he gave me privileges to deal with the release stuff. I control this repo as an admin, as well as the PyPi project and the mastodon account (which has been inactive but still). *I* know I'm trustworthy but others don't, so to some extent it's a leap of faith.

The biggest threat to a project like beets from this isn't an APT trying to get a vulnerability in but people trying to take the project in a different direction and doing the digital equivalent of a coup, worst case just maybe deleting the repo? I don't know. If there are any suggestions for how to share responsibility better, I'm all ears. At some point though, you just need to have some trust in maintainers and admins.

PR #5303, opened 2024-06-14
Remove the `beet` executable and update shebangs

### LINKED ISSUE #4604: Update /usr/bin shebangs to Python3?
### Problem

YouTuber found error because beets calls for Python3 but actually requires python36+.

https://youtu.be/_-WeZhlU0_I?t=492

Probably this code needs to change:

https://github.com/beetbox/beets/search?q=%2Fusr%2Fbin

**Open question**: Specify Python36 in the shebangs? I bet this would cause compatibility issues with shells that only have generic python3 installed with no subversion number (3.**6**).

### Setup

* OS: Linux

bal-e:
## Description

This is a very simple PR, just removing the (apparently unused) `beet` executable from the repository and substituting `bin/env python` for `bin/python3` where possible.

Fixes #4604.

## To Do

- [x] ~Documentation.~
- [x] Changelog.
- [x] ~Tests.~


bal-e:
I'm using this as a quick testing ground for #5305, hence the strange rebase.  I'll re-adjust it once the checks complete.

bal-e:
Restored, ready for review / merge.

Serene-Arc:
Hmm, does this impact calling beets on the command with `beet`?

bal-e:
In a venv, I ran `poetry install` and checked `$(which beet)`.  It's definitely a different script than the one I removed from the repository (I'm sure it's automatically generated):

```python
#!<path/to/my/venv>/bin/python
import sys
from beets.ui import main

if __name__ == '__main__':
    sys.exit(main())
```

So as long as Poetry is being used to build `beets`, which it should be, everything should be fine.  Even the `/usr/bin/beet` which is part of the Arch Linux `beets` package looks like a different file.

Serene-Arc:
Wonderful, should be fine to remove then.

wisp3rwind:
For context: The script that is actually being installed is generated by poetry via the `sripts` configuration in `pyproject.toml` (previously, it was generated via `entry_points` in `setup.py`.

Serene-Arc:
Great, wasn't entirely sure and I thought it best to double check! Don't want to suddenly be unable to use beet as a command

bal-e:
Yeah, absolutely!  Honestly, I should've investigated how the actual `beet` executable was generated a bit better before making the PR.  Thanks for making sure!

PR #5304, opened 2024-06-14
Fix changelog reminder workflow

snejus:
Quote the breaking glob as noted by @bal-e.


wisp3rwind:
For reference, this is superseded by/a duplicate of #5305

PR #5305, opened 2024-06-14
Fix changelog reminder workflow

snejus:
Quote the breaking glob as noted by @bal-e.

### DIFF
--- a/.github/workflows/changelog_reminder.yaml
+++ b/.github/workflows/changelog_reminder.yaml
@@ -16,7 +16,8 @@ jobs:
         id: changed-python-files
         uses: tj-actions/changed-files@v44
         with:
-          files: **.py
+          files: |
+            **.py
 
       - name: Check for the changelog update
         id: changelog-update
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -147,7 +147,7 @@ class BeetsPlugin:
         return wrapper
 
     def queries(self):
-        """Should return a dict mapping prefixes to Query subclasses."""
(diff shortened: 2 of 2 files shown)


snejus:
I was expecting it to remind me to update the changelog on this PR too 🕵🏼 🤨

bal-e:
I'm going to try rebasing #5303 on top of this to trigger the changelog action.

snejus:
Not sure, I think there's only one way to test it 🤷🏼‍♂️ at least now we know it shouldn't break!

Serene-Arc:
How bizarre. This is why testing GH actions are such a pita. They run in very strange ways. If you look at the checks, it hasn't been attempted. Maybe there's a restriction on running jobs that are being updated?

Serene-Arc:
It might still be running the version on master, rather than the one on the PR?

bal-e:
Yeah, maybe we should just try merging this PR and seeing if it helps.  I can't think of a better option, which sucks.

Serene-Arc:
GitHub actions and CI are amazing and useful but not easy to test and honestly, half of the time the only way TO test it is to push to production and just see what breaks. When I was writing the release workflow, I pushed to MY repo, and did a bunch of breaking changes, and then just reset it to upstream. And it *still* didn't find everything.

snejus:
I think it's happier now: https://github.com/beetbox/beets/actions/runs/9525145560/job/26258903712

Serene-Arc:
Wonderful

PR #5306, opened 2024-06-14
Fix black security vulnerability

snejus:
See https://github.com/beetbox/beets/security/dependabot/7

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -32,6 +32,7 @@ jobs:
         with:
           files: |
             **.py
+            poetry.lock
 
   format:
     if: needs.changed-files.outputs.any_python_changed == 'true'
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,7 +15,7 @@ jobs:
     name: Get changed files
     outputs:
       any_docs_changed: ${{ steps.changed-doc-files.outputs.any_changed }}
-      any_python_changed: ${{ steps.changed-python-files.outputs.any_changed }}
+      any_python_changed: ${{ steps.raw-changed-python-files.outputs.any_changed }}
       changed_doc_files: ${{ steps.changed-doc-files.outputs.all_changed_files }}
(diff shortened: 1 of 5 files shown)


wisp3rwind (approved):
Fine with me (even though I'm not sure this really has any security impact). Was there a reason that we capped `black` at a specific minor version (24.3)? 

Since the format check was skipped in CI, it seems possible that changed `black` defaults in version `25` might break our style checks which we can't see here.

snejus:
> Fine with me (even though I'm not sure this really has any security impact). Was there a reason that we capped `black` at a specific minor version (24.3)?
> 
> Since the format check was skipped in CI, it seems possible that changed `black` defaults in version `25` might break our style checks which we can't see here.

Good point, let's make the workflow to get triggered by a change in `poetry.lock` too!

snejus:
Now let me just figure out how to make `flake8` ignore `poetry.lock` here 😅

snejus:
I made the linting workflow to check the entire repository whenever `poetry.lock` is updated

```
Run poe check-format
Poe => black --check --diff --color .
All done! ✨ 🍰 ✨
191 files would be left unchanged.
Poe => isort --check --diff --color .
Skipped 1 files
```

snejus:
Ignoring flake8-docstrings in `aura.py` before https://github.com/beetbox/beets/pull/5234 is merged

snejus:
> Since the format check was skipped in CI, it seems possible that changed black defaults in version 25 might break our style checks which we can't see here.

@wisp3rwind `black` version is constrained between `>=24.3` and `<25` so it can't be updated to `25`, see `pyproject.toml`: 

```toml
black = ">=24.3,<25"
```

snejus:
Merging this in as I think I addressed your comments @wisp3rwind

wisp3rwind:
> I made the linting workflow to check the entire repository whenever `poetry.lock` is updated

Nice!

 
> @wisp3rwind `black` version is constrained between `>=24.3` and `<25` so it can't be updated to `25`, see `pyproject.toml`:

Obviously, I wasn't sufficiently awake when I commented here yesterday 😅

PR #5308, opened 2024-06-15
Update release workflow with poetry

snejus:
This was missed out in #5266: build the Python package distribution using Poetry in the release workflow.

In addition to the above, I slightly simplified the release pipeline which should hopefully make it easier to debug if things go wrong in the future.

### DIFF
--- a/.github/workflows/make_release.yaml
+++ b/.github/workflows/make_release.yaml
@@ -49,7 +49,7 @@ jobs:
         run: poetry build
 
       - name: Store the distribution packages
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: python-package-distributions
           path: dist/
@@ -78,7 +78,7 @@ jobs:
 
       - name: Tag the commit
         id: tag_version
-        uses: mathieudutour/github-tag-action@v6.1
+        uses: mathieudutour/github-tag-action@v6
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           custom_tag: ${{ env.NEW_VERSION }}
@@ -117,7 +117,7 @@ jobs:
       id-token: write
     steps:
       - name: Download all the dists
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         with:
           name: python-package-distributions
           path: dist/
--- a/.github/workflows/make_release.yaml
+++ b/.github/workflows/make_release.yaml
@@ -67,10 +67,28 @@ jobs:
           name: python-package-distributions
           path: dist/
 
+  publish-to-pypi:
+    name: Publish distribution 📦 to PyPI
+    runs-on: ubuntu-latest
+    needs: build
+    environment:
+      name: pypi
+      url: https://pypi.org/p/beets
+    permissions:
+      id-token: write
+    steps:
+      - name: Download all the dists
+        uses: actions/download-artifact@v4
+        with:
+          name: python-package-distributions
+          path: dist/
+      - name: Publish distribution 📦 to PyPI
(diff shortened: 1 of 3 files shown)


.github/workflows/make_release.yaml
```diff
@@ -7,99 +7,108 @@ on:
...
+
       - name: Store the distribution packages
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: python-package-distributions
           path: dist/
 
-  make_github_release:
+  publish-and-release:
```
Serene-Arc, line 70: My intent with this is that the pypi release and the github release would be two independent actions. If, say, there's a problem with PyPi, the github release will still be created and vice versa. They both are only made if the build succeeds but otherwise are separate.

Do you think that combining them is the better choice?

.github/workflows/make_release.yaml
```diff
@@ -7,99 +7,108 @@ on:
...
-        uses: actions/setup-python@v4
-        with:
-          python-version: "3.9"
-      - name: Obtain the changelog
-        run: echo CHANGELOG="$(python ./extra/release.py changelog)" >> $GITHUB_ENV
-      - name: Bump version and push tag
+      - name: Tag the commit
         id: tag_version
-        uses: mathieudutour/github-tag-action@v6.1
+        uses: mathieudutour/github-tag-action@v6.2
```
Serene-Arc, line 84: This might have been an error of mine. Can be simply use v6? I think that automatically updates to whatever the latest version is in most GH actions.

.github/workflows/make_release.yaml
```diff
@@ -7,99 +7,108 @@ on:
...
+    name: Create GitHub release and publish distribution 📦 to PyPI
     runs-on: ubuntu-latest
     needs: build
+    env:
+      CHANGELOG: ${{ needs.build.outputs.changelog }}
+    environment:
+      name: pypi
+      url: https://pypi.org/p/beets
+    permissions:
+      id-token: write
```
Serene-Arc, line 80: Are there any issues, security or otherwise, with making this permission wider? Just want to be sure, because before it was only applied to the PyPi release action, now it's for a lot of different steps.

Serene-Arc (changes requested):

.github/workflows/make_release.yaml
snejus: This is a very very very good point, thank you! In my world security always wins over anything else, thus I will undo the last commit

.github/workflows/make_release.yaml
snejus: > If, say, there's a problem with PyPi, the github release will still be created and vice versa.

In my head the GitHub release action depends on the package being uploaded to PyPI successfully: It feels a bit misleading to announce a new release to the users while the package has failed to be updated on PyPI. What do you reckon?

Given your comment below re security, I have split these two action back how they were defined previously. I am thinking of making `make-github-release` depend on `publish-to-pypi`

.github/workflows/make_release.yaml
snejus: Good point. I updated it. Any ideas on how to test this?

.github/workflows/make_release.yaml
Serene-Arc: That would work fine. Honestly it's just that pypi can be updated manually and until a couple weeks ago, I had no concrete proof that the PyPi would work as intended, whereas I knew the GH release would. So make it dependent if you think that's better.

snejus: We'll investigate it if it fails :)

Serene-Arc (approved):

snejus:
test

Issue #5218, opened 2024-05-03
Crash KeyError: 'title'


eric-sofisoftwarellc:
`Import fails with KeyError: 'title'`

The problem seems to be the file name doesn't have a space. If a track is named "01LadyOfAutumn.mp3", it crashes. If I change all the tracks to have a space after the number, eg "01 LadyOfAutumn.mp3", it succeeds.

Turning off plugins works, it does not crash.

I have latest version from git (pipx install https://github.com/beetbox/beets/tarball/master).

```
(beets) ➜  ~ beet import test
Traceback (most recent call last):
  File "/Users/user/.local/pipx/venvs/beets/bin/beet", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/user/.local/pipx/venvs/beets/lib/python3.12/site-packages/beets/ui/__init__.py", line 1865, in main
    _raw_main(args)
  File "/Users/user/.local/pipx/venvs/beets/lib/python3.12/site-packages/beets/ui/__init__.py", line 1852, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "/Users/user/.local/pipx/venvs/beets/lib/python3.12/site-packages/beets/ui/commands.py", line 1395, in import_func
    import_files(lib, paths, query)
  File "/Users/user/.local/pipx/venvs/beets/lib/python3.12/site-packages/beets/ui/commands.py", line 1326, in import_files
    session.run()
  File "/Users/user/.local/pipx/venvs/beets/lib/python3.12/site-packages/beets/importer.py", line 360, in run
    pl.run_parallel(QUEUE_SIZE)
  File "/Users/user/.local/pipx/venvs/beets/lib/python3.12/site-packages/beets/util/pipeline.py", line 447, in run_parallel
    raise exc_info[1].with_traceback(exc_info[2])
  File "/Users/user/.local/pipx/venvs/beets/lib/python3.12/site-packages/beets/util/pipeline.py", line 312, in run
    out = self.coro.send(msg)
          ^^^^^^^^^^^^^^^^^^^
  File "/Users/user/.local/pipx/venvs/beets/lib/python3.12/site-packages/beets/util/pipeline.py", line 195, in coro
    func(*(args + (task,)))
  File "/Users/user/.local/pipx/venvs/beets/lib/python3.12/site-packages/beets/importer.py", line 1490, in lookup_candidates
    plugins.send("import_task_start", session=session, task=task)
[...]

PR #5311, opened 2024-06-15
fromfilename plugin: proposed fix for #5218, some improvements

Vrihub:
### b5216a0 is a proposed fix for #5218

We don't assume anymore that if the pattern lacks the "artist" group it also has the "title" group (this was causing the "title" KeyError, when the pattern containg only the `?P<track>` group is used).
Instead, we check for existence of "title" before assigning "title_field", otherwise we let the last pattern (which will always match) to assign it. The only slight drawback is that              files like "01.mp3" will also get "01" as title (besides the "1" track number), instead of having no title, but I guess that is not going to cause any major problem in the rest of the import procedure (e.g. track matching after a search). On the other hand, I guess allowing a file to have no title would require a substantial rewriting of the plugin, since the empty string is in BAD_TITLE_PATTERNS.

### 0966042 adds a log message about the pattern being tried
This is useful while debugging changes to the regexps in PATTERNS.
I used loglevel "debug" for this message, should we use "debug" also for other messages (currently using level "info")?

### e6b7735 refactors the regexps in PATTERNS

I reviewed the regexps (especially after the changes I made some time ago in 84cf336) and tried to make them cleaner.
- Allow " - " and "-" as separator between track/artist/title fields: that should cover two common filename conventions: using spaces (e.g. `01 - Artist Name - Title of the song`) or not using spaces (e.g. `01-artist_name-title_of_the_song`). Some regexp groups are non-greedy, to avoid capturing unwanted leading/trailing spaces in artist/titles; we could think about some other enhancements, e.g. automatically converting underscores to spaces in artist/title names, but I guess the main use of the plugin is not to produce a perfectly formatted title, but a sufficient meaningful one to be used as reference in the following stages of the update procedure.
- Allow `?P<tag>` groups by making them optional in the first two patterns (instead of having two versions of the pattern, with or without tag). BTW I'm not sure what these groups are about: we optionally check for them in the plugin, but we never use them to assign any metadata.
- Allow optional "." after the track number
- In the third pattern: also allow "_" as separator (e.g. for file names like `01_some_tack.mp3`)


snejus:
Could you by any chance add some tests that show the range of filenames that can be parsed? This would be very useful going forward.

Serene-Arc:
Second @snejus here! Thanks for the PR too. Regex is so complicated and obtuse to many people, so it's always good to have as complete a testing regimen in place as possible, both to find bugs with the provided patterns and to ensure that the same strings are matched going forward with any changes.

Vrihub:
Can I have any feedback about this PR?

I'm wondering especially if anything has changed after adopting poetry regarding how tests should be implemented, thanks!

snejus:
Apologies - I didn't see this since I did not receive a review request (best to do this whenever your PR is ready to be re-reviewed).

There have been a fair bit of changes merged in since the last commit here, so you want to merge / rebase on `master` first, and then address formatting/linting issues.

regagain:
> Is this still relevant?

Yes, I ran into this issue today and this change fixed it. Beets 2.1.0 / Linux.

JOJ0:
@Vrihub I'd like to resurrect this PR. Do you feel like rebasing, fix the tests, etc??

I'll try to help as good as I can to get this merge ready. Thanks in advance!

Vrihub:
> @Vrihub I'd like to resurrect this PR. Do you feel like rebasing, fix the tests, etc??

Sure, first I have to switch my working copy to poetry. I'll give a shot at it later today or tomorrow.

Vrihub:
OK, sorry for the delay (bad week), I've updated the PR to the latest changes in master, poetry etc. 

I've been using this fix to the plugin during all this time and I haven't found problems when importing albums. The new tests seem to work fine too. So unless you have further suggestions, I think we should be ready to merge.

Vrihub:
Note: with [0ac7fb4](https://github.com/beetbox/beets/pull/5311/commits/0ac7fb42f50c923ca149fcffaed6c93f5e60e0ea) I undo the fix implemented in https://github.com/beetbox/beets/pull/5907 because the case of a `Null` title is already handled (excluded) in the new version of the plugin contained in this PR.

semohr:
The actual changes to the plugin look good to me!

Would you be open to rewriting the tests using pytest tho? It should be fairly straightforward and could significantly reduce boilerplate. I think most of the tests could be handled with a single pytest.mark.parametrize. I can also give it a try if you like.

In my opinion, using fixtures would make the tests more concise and easier to maintain over time.

Vrihub:
> Would you be open to rewriting the tests using pytest tho?

Ah sure, I'll take care of this.

test/plugins/test_fromfilename.py
```diff
@@ -0,0 +1,142 @@
...
+        taskconfig = {"is_album": True, "items": [self.item1, self.item2]}
+        self.task = Mock(**taskconfig)
+
+    def tearDown(self):
+        del self.session, self.task, self.item1, self.item2
+
+    def test_sep_sds(self):
+        """Test filenames that use " - " as separator."""
+
+        self.item1.path = "/music/files/01 - Artist Name - Song One.m4a"
```
snejus: I think it's enough to define just a single path for testing - the second one doesn't add much value, only makes the tests more verbose.

test/plugins/test_fromfilename.py
```diff
@@ -0,0 +1,142 @@
...
+
+"""Tests for the fromfilename plugin."""
+
+import unittest
+from unittest.mock import Mock
+
+from beetsplug import fromfilename
+
+
+class FromfilenamePluginTest(unittest.TestCase):
```
snejus: This entire test class can be reduced to just a couple of lines if you use `@pytest.mark.parametrize`. Please see `test/test_query.py`, `test/autotag/test_distance` and `test/plugins/test_lyrics.py` for some examples how we use it.

beetsplug/fromfilename.py
```diff
@@ -25,12 +25,12 @@
...
-    r"^(?P<artist>.+)[\-_](?P<title>.+)$",
-    r"^(?P<track>\d+)[\s.\-_]+(?P<artist>.+)[\-_](?P<title>.+)$",
-    r"^(?P<track>\d+)[\s.\-_]+(?P<title>.+)$",
-    r"^(?P<track>\d+)\s+(?P<title>.+)$",
+    (
+        r"^(?P<track>\d+)\.?\s*-\s*(?P<artist>.+?)\s*-\s*(?P<title>.+?)"
+        r"(\s*-\s*(?P<tag>.*))?$"
+    ),
+    r"^(?P<artist>.+?)\s*-\s*(?P<title>.+?)(\s*-\s*(?P<tag>.*))?$",
+    r"^(?P<track>\d+)\.?[\s\-_]+(?P<title>.+)$",
```
snejus: No need for a backlash prefix before the dash `-` if you move it to the end of the range (the start also works).
```suggestion
    r"^(?P<track>\d+)\.?[\s_-]+(?P<title>.+)$",
```

Vrihub: > No need for a backlash prefix before the dash `-` if you move it to the end of the range (the start also works).

Good point.

test/plugins/test_fromfilename.py
Vrihub: The second path was used (in some tests) to allow for the case when an optional "." follows the track number. I'll replace this with a proper case with pytest parametrize.

JOJ0:
Hi @snejus and @Vrihub, I fixed the changelog conflict and updated from master. Now ran into a tiny formatting issue which I solved in 152cafbf69de036253898a11349b8d59ecb6618b . @Vrihub maybe you want to rebase again to make this prettier. If not, all good. (I couldn't rebase/force-push to this branch but realized that too late). 

By the way I'm using this feature in my main beets branch now and it works well so far :-) 

Also had a look at the pytest refactor and like it. In short: This LGTM, @snejus might want to give it the final review! Thanks in advance to both of you!

PR #5327, opened 2024-06-20
Filter albums by tracks fields and vice versa

### LINKED ISSUE #4360: beet list -a path:some/path queries are extremely slow

snejus:
Fixes #4360

This PR enables querying albums by track fields and tracks by album fields, and speeds up querying albums by `path` field.

It originally was part of #5240, however we found that the changes related to the flexible attributes caused degradation in performance. So this PR contains the first part of #5240 which joined `items` and `albums` tables in queries.

### DIFF
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -39,7 +39,7 @@ from unidecode import unidecode
 from beets import config, logging, plugins
 from beets.autotag import mb
 from beets.library import Item
-from beets.util import as_string
+from beets.util import as_string, cached_classproperty
 
 log = logging.getLogger("beets")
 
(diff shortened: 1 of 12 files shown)


Serene-Arc (changes requested):
In the original PR, the interface for `field` was reverted back, can we do that for this one as well? Just in case that any third-part plugins use it.

Serene-Arc (approved):
Ah, my mistake! Looks fine to merge then

PR #5331, opened 2024-06-22
Use typing_extensions only when needed

qarkai:
## Description

Use `typing_extensions` only for Python <= 3.10.

`Self` was added in Python 3.11.
`TypeAlias` was added in Python 3.10.

## To Do

- [x] Documentation.
- [x] Changelog.
- [x] Tests.

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -45,7 +45,11 @@ from typing import (
     Union,
 )
 
-from typing_extensions import TypeAlias
+try:
+    # Python 3.10+
+    from typing import TypeAlias
+except ImportError:
+    from typing_extensions import TypeAlias
 from unidecode import unidecode
 
 from beets.util import hidden
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -45,11 +45,11 @@ from typing import (
     Union,
 )
 
-try:
-    # Python 3.10+
+if sys.version_info >= (3, 10):
     from typing import TypeAlias
-except ImportError:
+else:
     from typing_extensions import TypeAlias
+
 from unidecode import unidecode
 
 from beets.util import hidden
--- a/beetsplug/aura.py
+++ b/beetsplug/aura.py
@@ -17,6 +17,7 @@
 
 import os
 import re
+import sys
 from dataclasses import dataclass
 from mimetypes import guess_type
 from typing import ClassVar, Mapping, Type
@@ -30,10 +31,9 @@ from flask import (
     send_file,
 )
 
-try:
-    # Python 3.11+
+if sys.version_info >= (3, 11):
     from typing import Self
-except ImportError:
+else:
     from typing_extensions import Self
 
 from beets import config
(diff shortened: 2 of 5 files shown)


bal-e:
Since type annotations are only used by mypy, maybe we should make these statements conditional on [`typing.TYPE_CHECKING`](https://mypy.readthedocs.io/en/latest/runtime_troubles.html#typing-type-checking) as well?  In any case, I think an explicit `if` statement (e.g. on the running Python version) would be preferable over catching an `ImportError`.

(Disclaimer: I'm not a maintainer, I'm just providing some feedback.)

qarkai:
Ruff outputs [runtime-import-in-type-checking-block (TCH004)](https://docs.astral.sh/ruff/rules/runtime-import-in-type-checking-block/) warning if `from typing_extensions import` is in type-checking block. Not sure why.

bal-e:
I suspect Ruff special-cases imports in type-checking blocks, only allowing things from modules it knows about.  It definitely looks like they support _some_ imports in type-checking blocks: it's mentioned [here](https://docs.astral.sh/ruff/settings/#lint_flake8-type-checking_quote-annotations) for example.

Serene-Arc:
Hi! Thanks for the PR. What is the benefit of doing this rather than what we already have? Python 3.8 is currently our minimum supported version.

qarkai:
> Hi! Thanks for the PR. What is the benefit of doing this rather than what we already have? Python 3.8 is currently our minimum supported version.

Use built-in module in modern Python.
Reduce dependencies, make life easier for downstream maintainers.
Simplify upgrading to newer Python versions in the future, i.e. when minimum supported version is Python 3.11 typing_extensions could be removed completely.

Serene-Arc:
Those benefits would only happen when we have Python 3.10 as the minimally supported version version though, wouldn't it? There'd be no difference until we reach that point, since it will still be on the dependency list regardless.

qarkai:
Nope, if distro has Python >= 3.11 maintainer would be able to drop `typing_extensions` dependency for the next beets release with this patch even if the minimal supported version is still 3.8.
It's in `typing_extensions = { version = "*", python = "<=3.10" }` line.

Serene-Arc (approved):
Very well then, thanks for explaining.

semohr:
To give you a heads up, we're considering reversing this decision. In hindsight, it significantly degraded the developer experience. For example:
```python
if sys.version_info >= (3, 11):
    from typing import NotRequired
else:
    from typing_extensions import NotRequired
```
vs simply:
```python
from typing_extensions import NotRequired
```

Right now, we're effectively limited to using typing features that are natively available in Python <= 3.11.  This becomes a problem when we want to adopt newer features that are safely backported via `typing_extensions`, For instance default values for `TypeVar` or typeguards using `TypeIs`, both of which are introduced in Python 3.13.

I suspect this will be a recurring issue, and locking a dependency to a specific version just creates unnecessary friction. `typing_extensions` maintains a complete list of backported features, so switching to the stdlib when dropping support for older Python versions would be straightforward.

PR #5357, opened 2024-07-10
Perform a regex substitution in the substitute plugin

nichobi:
## Description

This utilises regex substitution in the substitute plugin. The previous approach only used regex to match the pattern, then replaced it with a static string. This change allows more complex substitutions, where the output depends on the input.

### Example use case
Say we want to keep only the first artist of a multi-artist credit, as in the following list:
```
Neil Young & Crazy Horse -> Neil Young
Michael Hurley, The Holy Modal Rounders, Jeffrey Frederick & The Clamtones -> Michael Hurley
James Yorkston and the Athletes -> James Yorkston
````
This would previously have required three separate rules, one for each resulting artist. By using a regex substitution, we can get the desired behaviour in a single rule: 
```yaml
substitute:
  ^(.*?)(,| &| and).*: \1
```
(Capture the text until the first `,` ` &` or ` and`, then use that capture group as the output)

### Notes
I've kept the previous behaviour of only applying the first matching rule, but I'm not 100% sure it's the ideal approach.
I can imagine both cases where you want to apply several rules in sequence and cases where you want to stop after the first match.

## To Do

### DIFF
--- a/beetsplug/substitute.py
+++ b/beetsplug/substitute.py
@@ -35,7 +35,7 @@ class Substitute(BeetsPlugin):
         """Do the actual replacing."""
         if text:
             for pattern, replacement in self.substitute_rules:
-                new_text, subs_made = re.subn(pattern, replacement, text)
+                new_text, subs_made = pattern.subn(replacement, text)
                 if subs_made > 0:
                     return new_text
             return text
@@ -48,7 +48,7 @@ class Substitute(BeetsPlugin):
         Get the configuration, register template function and create list of
         substitute rules.
         """
-        super(Substitute, self).__init__()
+        super().__init__()
         self.substitute_rules = []
         self.template_funcs["substitute"] = self.tmpl_substitute
 
@@ -70,7 +70,7 @@ class Substitute(BeetsPlugin):
                     """
                 )
             )
-            pairs = [(key, view.as_str()) for key, view in self.config.items()]
+            pairs = self.config.flatten().items()
         else:
             pairs = self.config.as_pairs()
 
--- a/beetsplug/substitute.py
+++ b/beetsplug/substitute.py
@@ -18,6 +18,7 @@ Uses user-specified substitution rules to canonicalize names for path formats.
 """
 
 import re
+import textwrap
 
 from beets.plugins import BeetsPlugin
 
@@ -51,7 +52,30 @@ class Substitute(BeetsPlugin):
         self.substitute_rules = []
         self.template_funcs["substitute"] = self.tmpl_substitute
 
-        for key, view in self.config.items():
-            value = view.as_str()
+        match self.config.get():
+            case list():
+                pairs = self.config.as_pairs()
+            case dict():
+                pairs = [
+                    (key, view.as_str()) for key, view in self.config.items()
+                ]
+                self._log.warning(
+                    "Unordered configuration is deprecated, as it leads to"
+                    + " unpredictable behaviour on overlapping rules.\n"
+                    + textwrap.dedent(
+                        """
+                        Old syntax:
+                          substitute:
+                            a: b
+                            c: d
+
+                        New syntax:
+                          substitute:
+                          - a: b
+                          - c: d
+                        """
+                    )
+                )
+        for key, value in pairs:
             pattern = re.compile(key, flags=re.IGNORECASE)
             self.substitute_rules.append((pattern, value))
--- a/beetsplug/substitute.py
+++ b/beetsplug/substitute.py
@@ -34,7 +34,7 @@ class Substitute(BeetsPlugin):
         """Do the actual replacing."""
         if text:
             for pattern, replacement in self.substitute_rules:
-                if pattern.match(text.lower()):
+                if pattern.match(text):
                     return replacement
             return text
         else:
@@ -52,5 +52,5 @@ class Substitute(BeetsPlugin):
 
         for key, view in self.config.items():
             value = view.as_str()
-            pattern = re.compile(key.lower())
+            pattern = re.compile(key, flags=re.IGNORECASE)
             self.substitute_rules.append((pattern, value))
--- a/beetsplug/substitute.py
+++ b/beetsplug/substitute.py
@@ -34,9 +34,9 @@ class Substitute(BeetsPlugin):
         """Do the actual replacing."""
         if text:
             for pattern, replacement in self.substitute_rules:
-                new_string, number_of_subs_made = re.subn(pattern, replacement, text)
-                if number_of_subs_made > 0:
-                    return new_string
+                new_text, subs_made = re.subn(pattern, replacement, text)
+                if subs_made > 0:
+                    return new_text
             return text
         else:
             return ""
--- a/beetsplug/substitute.py
+++ b/beetsplug/substitute.py
@@ -18,7 +18,6 @@ Uses user-specified substitution rules to canonicalize names for path formats.
 """
 
 import re
-import textwrap
 
 from beets.plugins import BeetsPlugin
 
@@ -49,31 +48,8 @@ class Substitute(BeetsPlugin):
         substitute rules.
         """
         super().__init__()
-        self.substitute_rules = []
         self.template_funcs["substitute"] = self.tmpl_substitute
-
-        if isinstance(self.config.get(), dict):
-            self._log.warning(
-                "Unordered configuration is deprecated, as it leads to"
-                + " unpredictable behaviour on overlapping rules.\n"
-                + textwrap.dedent(
-                    """
-                    Old syntax:
-                      substitute:
-                        a: b
-                        c: d
-
-                    New syntax:
-                      substitute:
-                      - a: b
-                      - c: d
-                    """
-                )
-            )
-            pairs = self.config.flatten().items()
-        else:
-            pairs = self.config.as_pairs()
-
-        for key, value in pairs:
-            pattern = re.compile(key, flags=re.IGNORECASE)
-            self.substitute_rules.append((pattern, value))
+        self.substitute_rules = [
+            (re.compile(key, flags=re.IGNORECASE), value)
+            for key, value in self.config.flatten().items()
+        ]
--- a/beetsplug/substitute.py
+++ b/beetsplug/substitute.py
@@ -48,34 +48,32 @@ class Substitute(BeetsPlugin):
         Get the configuration, register template function and create list of
         substitute rules.
         """
-        super().__init__()
+        super(Substitute, self).__init__()
         self.substitute_rules = []
         self.template_funcs["substitute"] = self.tmpl_substitute
 
-        match self.config.get():
-            case list():
-                pairs = self.config.as_pairs()
-            case dict():
-                pairs = [
-                    (key, view.as_str()) for key, view in self.config.items()
-                ]
-                self._log.warning(
-                    "Unordered configuration is deprecated, as it leads to"
-                    + " unpredictable behaviour on overlapping rules.\n"
-                    + textwrap.dedent(
-                        """
-                        Old syntax:
-                          substitute:
-                            a: b
-                            c: d
+        if isinstance(self.config.get(), dict):
+            self._log.warning(
+                "Unordered configuration is deprecated, as it leads to"
+                + " unpredictable behaviour on overlapping rules.\n"
+                + textwrap.dedent(
+                    """
+                    Old syntax:
+                      substitute:
+                        a: b
+                        c: d
 
-                        New syntax:
-                          substitute:
-                          - a: b
-                          - c: d
-                        """
-                    )
+                    New syntax:
+                      substitute:
+                      - a: b
+                      - c: d
+                    """
                 )
+            )
+            pairs = [(key, view.as_str()) for key, view in self.config.items()]
+        else:
+            pairs = self.config.as_pairs()
+
         for key, value in pairs:
             pattern = re.compile(key, flags=re.IGNORECASE)
             self.substitute_rules.append((pattern, value))
--- a/beetsplug/substitute.py
+++ b/beetsplug/substitute.py
@@ -34,8 +34,9 @@ class Substitute(BeetsPlugin):
         """Do the actual replacing."""
         if text:
             for pattern, replacement in self.substitute_rules:
-                if pattern.match(text):
-                    return replacement
+                new_string, number_of_subs_made = re.subn(pattern, replacement, text)
+                if number_of_subs_made > 0:
+                    return new_string
             return text
         else:
             return ""
--- a/beetsplug/substitute.py
+++ b/beetsplug/substitute.py
@@ -34,9 +34,7 @@ class Substitute(BeetsPlugin):
         """Do the actual replacing."""
         if text:
             for pattern, replacement in self.substitute_rules:
-                new_text, subs_made = pattern.subn(replacement, text)
-                if subs_made > 0:
-                    return new_text
+                text = pattern.sub(replacement, text)
             return text
         else:
             return ""
--- a/docs/plugins/substitute.rst
+++ b/docs/plugins/substitute.rst
@@ -13,6 +13,7 @@ Enable the ``substitute`` plugin (see :ref:`using-plugins`), then make a ``subst
 Each rule consists of a case-insensitive regular expression pattern, and a
 replacement string. For example, you might use:
 
+.. code-block:: yaml
     substitute:
       - .*jimi hendrix.*: Jimi Hendrix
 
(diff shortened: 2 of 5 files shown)


Serene-Arc:
Hi! Thanks for the PR. Are you comfortable adding tests for this enhancement? Anything with regex should be heavily tested to make sure it behaves exactly as expected.

nichobi:
> Hi! Thanks for the PR. Are you comfortable adding tests for this enhancement? Anything with regex should be heavily tested to make sure it behaves exactly as expected.

Thanks for the feedback, and sorry it's taken me so long to respond. I've added some testing now, although I've intentionally left one failing.

I had assumed that the order of properties would be preserved when loading the configuration, but this does not seem to be the case. This means it's quite unpredictable what rule will be triggered if the input matches several rules. This would also be the case in the current version of the plugin.

Perhaps the configuration should be refactored into a list of rules, rather than an object, so order is preserved?

Maybe you have thoughts, as the creator of the plugin @fdaniele85?

fdaniele85:
Hi! I agree that it should be a list of rules and order should be preserved. I never had an overlap of two rules, therefore I did not think about this case.

nichobi:
I've updated the configuration to use a list of rules, while deprecating the previous syntax.
The old syntax will still work, but with a warning.

Old syntax:
```yaml
substitute:
  a: b
  c: d
```

New syntax:
```yaml
substitute:
- a: b
- c: d
```

docs/plugins/substitute.rst
```diff
@@ -9,15 +9,30 @@ Experience to be sorted into the same folder as solo Hendrix albums.
...
+      - .*jimi hendrix.*: Jimi Hendrix
+
+The replacement can be an expression utilising the matched regex, allowing us
+to create more general rules. Say for example, we want to sort all albums by
+multiple artists into the directory of the first artist. We can thus capture
+everything before the first ``,``, `` &`` or `` and``, and use this capture
+group in the output, discarding the rest of the string.
+
+    substitute:
+      - ^(.*?)(,| &| and).*: \1
```
snejus, line 26: Could you make this and above code snippets to be colorized in `yaml` syntax?
```suggestion
.. code-block:: yaml
    substitute:
      - ^(.*?)(,| &| and).*: \1
```

test/plugins/test_substitute.py
```diff
@@ -0,0 +1,116 @@
...
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+
+"""Test the substitute plugin regex functionality."""
+
+from beets.test.helper import PluginTestCase, capture_log
+from beetsplug.substitute import Substitute
+
+PLUGIN_NAME = "substitute"
```
snejus, line 20: Is this required?

test/plugins/test_substitute.py
```diff
@@ -0,0 +1,116 @@
...
+            for input, expected in cases:
+                assert Substitute().tmpl_substitute(input) == expected
+
+    def test_case_insensitivity(self):
+        with self.configure_plugin([{"a": "b"}]):
+            assert Substitute().tmpl_substitute("A") == "b"
+
+    def test_unmatched_input_preserved(self):
+        with self.configure_plugin([{"a": "b"}]):
+            assert Substitute().tmpl_substitute("c") == "c"
```
snejus, line 45: I think you can remove some duplication found in each of the test cases by defining a `run_substitute` method which accepts the configuration under test and the test cases. I 

```python
class SubstitutePluginTest(PluginTestCase):
    plugin = "substitute"
    preload_plugin = False

    def run_substitute(self, config, cases):
        with self.configure_plugin(config):
            for input, expected in cases:
                assert Substitute().tmpl_substitute(input) == expected

    def test_simple_substitute(self):
        self.run_substitute(
            [
                {"a": "b"},
                {"b": "c"},
                {"c": "d"},
            ],
            [("a", "b"), ("b", "c"), ("c", "d")],
        )

    def test_case_insensitivity(self):
        self.run_substitute([{"a": "b"}], [("A", "b")])

    def test_unmatched_input_preserved(self):
        self.run_substitute([{"a": "b"}], [("c", "c")])

    ...
```

beetsplug/substitute.py
```diff
@@ -46,11 +48,32 @@ def __init__(self):
         Get the configuration, register template function and create list of
         substitute rules.
         """
-        super().__init__()
+        super(Substitute, self).__init__()
```
snejus, line 51: Was there a reason for this change? This sort of syntax is outdated I think

beetsplug/substitute.py
```diff
@@ -46,11 +48,32 @@ def __init__(self):
...
+                    """
+                    Old syntax:
+                      substitute:
+                        a: b
+                        c: d
+
+                    New syntax:
+                      substitute:
+                      - a: b
+                      - c: d
```
snejus, line 69: ✍🏼 Nit:

```suggestion
                        - a: b
                        - c: d
```

beetsplug/substitute.py
```diff
@@ -34,8 +35,9 @@ def tmpl_substitute(self, text):
         """Do the actual replacing."""
         if text:
             for pattern, replacement in self.substitute_rules:
-                if pattern.match(text.lower()):
-                    return replacement
+                new_text, subs_made = re.subn(pattern, replacement, text)
+                if subs_made > 0:
+                    return new_text
```
snejus, line 40: Does this mean that only the very first matching pattern is applied?

beetsplug/substitute.py
```diff
@@ -46,11 +48,32 @@ def __init__(self):
...
         self.substitute_rules = []
         self.template_funcs["substitute"] = self.tmpl_substitute
 
-        for key, view in self.config.items():
-            value = view.as_str()
-            pattern = re.compile(key.lower())
+        if isinstance(self.config.get(), dict):
+            self._log.warning(
+                "Unordered configuration is deprecated, as it leads to"
+                + " unpredictable behaviour on overlapping rules.\n"
```
snejus, line 58: I'm slightly confused about this issue. Using the old configuration, does the issue have to do with a configuration like this?

```yaml
substitute:
  a: b
  a: c
```

If so, I think it should be pretty much predictable: the last duplicate key wins, since it's equivalent to `{"a": "b", "a": "c"}` in Python.

beetsplug/substitute.py
```diff
@@ -34,8 +35,9 @@ def tmpl_substitute(self, text):
         """Do the actual replacing."""
         if text:
             for pattern, replacement in self.substitute_rules:
-                if pattern.match(text.lower()):
-                    return replacement
+                new_text, subs_made = re.subn(pattern, replacement, text)
```
snejus, line 38: You may want to use the compiled pattern here
```suggestion
                new_text, subs_made = pattern.subn(replacement, text)
```

snejus (commented):
Added several comments

beetsplug/substitute.py
nichobi: Yes, and this was the case previously as well. I'm not 100% certain it is the optimal behaviour, but it does have its advantages. One example is it makes it easy to exclude certain matches, by adding a no-op rule. For example, to prevent splitting a certain artist with `and` in its title:
```yaml
substitute:
- It's Jo and Danny: It's Jo and Danny
- ^(.*?)(,| &| and).*: \1
```

beetsplug/substitute.py
nichobi: This ties in to the above question, that we only apply the first matching rule. Because our keys are regex patterns, multiple *different* keys can match the same input. With the below configuration, what do you expect `a` to be substituted with?
```yaml
substitute:
  a: b
  (a|x): c
```

I initially believed it would result in `b`, since that's the first defined rule, but in testing I found it to be `c`.

I now believe it is because the keys end up in alphabetical order (or at least some other order), which makes it hard to predict or control which rule should be applied. By turning it into a list, we preserve the order of rules as defined by the user. Behaviour then becomes simple to define: the first matching rule is applied.

beetsplug/substitute.py
nichobi: I believe this is somewhat of a divisive matter :)

I too used the indented syntax you are suggesting for many years, but after repeatedly encountering both varieties being used (often even within the same document). I eventually got tired of the inconsistency and tried to find out which was "correct" (both are of course valid yaml).

It seems the [yaml specification](https://yaml.org/spec/1.2.2/#example-mapping-scalars-to-sequences) favours the less indented variety and uses it in all its examples. If beets favours the indented variety, I am of course happy to be consistent.

test/plugins/test_substitute.py
nichobi: Apparently not, thank you! I saw it in another test (advanced_rewrite) and assumed it was required.

beetsplug/substitute.py
```diff
@@ -46,11 +48,32 @@ def __init__(self):
...
+                        c: d
+
+                    New syntax:
+                      substitute:
+                      - a: b
+                      - c: d
+                    """
+                )
+            )
+            pairs = [(key, view.as_str()) for key, view in self.config.items()]
```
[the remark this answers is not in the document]
nichobi: I still needed the call to items, but that still makes it a lot nicer, thanks!
```py
pairs = self.config.flatten().items()
```

test/plugins/test_substitute.py
nichobi: This simplified things a lot, many thanks!

beetsplug/substitute.py
snejus: Regarding the codebase, I thought the plugin documentation may be suitable place to check:

```fish
$ grep '^ *- [a-z]+: ' -iEB2 docs/**/*.rst
docs/plugins/advancedrewrite.rst-
docs/plugins/advancedrewrite.rst-    advancedrewrite:
docs/plugins/advancedrewrite.rst:      - match: "mb_artistid:dec0f331-cb08-4c8e-9c9f-aeb1f0f6d88c year:..2022"
--
docs/plugins/advancedrewrite.rst-
docs/plugins/advancedrewrite.rst-    advancedrewrite:
docs/plugins/advancedrewrite.rst:      - match: "artist:배유빈 feat. 김미현"
--
docs/plugins/advancedrewrite.rst-    advancedrewrite:
docs/plugins/advancedrewrite.rst-      # BAD, DON'T DO THIS!
docs/plugins/advancedrewrite.rst:      - match: album:THE ALBUM
--
docs/plugins/advancedrewrite.rst-    advancedrewrite:
docs/plugins/advancedrewrite.rst-      # Note the quotes around the query string!
docs/plugins/advancedrewrite.rst:      - match: album:"THE ALBUM"
--
docs/plugins/albumtypes.rst-    albumtypes:
docs/plugins/albumtypes.rst-        types:
docs/plugins/albumtypes.rst:            - ep: 'EP'
docs/plugins/albumtypes.rst:            - single: 'Single'
docs/plugins/albumtypes.rst:            - soundtrack: 'OST'
docs/plugins/albumtypes.rst:            - live: 'Live'
docs/plugins/albumtypes.rst:            - compilation: 'Anthology'
docs/plugins/albumtypes.rst:            - remix: 'Remix'
--
docs/plugins/fetchart.rst-        sources:
docs/plugins/fetchart.rst-            - filesystem
docs/plugins/fetchart.rst:            - coverart: release
docs/plugins/fetchart.rst-            - itunes
docs/plugins/fetchart.rst:            - coverart: releasegroup
```

It seems like the indentation is preferred.

I personally just follow whatever is the preference of the most popular formatter for the language: in this case both `yamlfix` and `yamlfmt` add indentation by default.

beetsplug/substitute.py
snejus: Maybe I got it wrong, but it seems to me that this design will only ever apply the first substitution it matches?

Say, I want to replace all **ft**s with a comma, and only then split the artists:

```yaml
substitute:
  - \ +f(ea)?t(\.|uring)?: ", "
  - ^([^,]+),.*: \1
```

If I have `Artist ft. Someone`, I would expect to end up with `Artist`, but it seems like I would end up with `Artist, Someone` since only the first pattern would get applied.

beetsplug/substitute.py
snejus: How did you test this? And most importantly, which Python version are you using?

`config["substitute"]` returns an `OrderedDict`

```python
>>> from beets import config

>>> list(config["substitute"].items())
Out[11]: [('a', <Subview: substitute.a>), ('(a|x)', <Subview: substitute.(a|x)>)]

>>> config["substitute"].flatten()
Out[12]: OrderedDict([('a', 'b'), ('(a|x)', 'c')])
```

So I'd expect the plugin to apply the replacements in the sequence they are defined. And it seems like it's working as expected:

```python
>>> from beetsplug.substitute import Substitute

>>> s = Substitute()

>>> s.substitute_rules
Out[7]: [(re.compile(r'a', re.UNICODE), 'b'), (re.compile(r'(a|x)', re.UNICODE), 'c')]

>>> s.tmpl_substitute("a")
Out[8]: 'b'
```

nichobi: I encountered it in one of the test cases I had written, on my own Python 3.12 setup and in the [Github Actions tests](https://github.com/beetbox/beets/actions/runs/11093855139/job/30820644945#step:8:121) in this PR. 

I can however replicate your findings on my machine, and it seems the order is in fact preserved when loaded from the yaml configuration. Perhaps this is then only an error in the testing setup.

My attempts to fix the test case however have not been successful. Even declaring the configuration as an OrderedDict still results in a jumbled up configuration. 

```py
def test_break_on_first_match(self):
    self.run_substitute(
        OrderedDict([
            ("a", "x"),
            ("[ab]", "y"),
        ]),
        [
            ("a", "x"),
            ("b", "y"),
        ],
    )
```

This test still fails the first case, as `a` becomes substituted with `y`.

snejus: Ahh, I see. Have tested it and found that the issue is with how the configuration is applied in the tests:

```python
    @contextmanager
    def configure_plugin(self, config: list[Any] | dict[str, Any]):
        if isinstance(config, list):
            beets.config[self.plugin] = config
        else:
            for key, value in config.items():
                beets.config[self.plugin][key] = value
        self.load_plugins(self.plugin)

        yield

        self.unload_plugins()
```

Apologies - I've written this context manager myself and back then I did not know the inner working of `beets.config` yet.

Try replacing this function in `beets.test.helper` with

```python
    @contextmanager
    def configure_plugin(self, config: Any):
        beets.config[self.plugin].set(config)
        self.load_plugins(self.plugin)

        yield

        self.unload_plugins()
```

beetsplug/substitute.py
nichobi: Your understanding is correct. I'm however not sure applying every matching rule in sequence would actually give that much more power. You could just as well write a single rule that transforms `Artist ft. Someone` to `Artist` directly.
```yaml
substitute:
  - (.+) +f(ea)?t(\.|uring)?.*: \1
```

I also do think it's worth being able to create exceptions by interrupting the flow, as with my `It's Jo and Danny` example above (where the group itself has 'and' in its name). I'm not sure how you'd handle that if every rule will be applied in sequence.

I personally think there's value to the simplicity of each case being handled by a single rule.

beetsplug/substitute.py
nichobi: > Apologies - I've written this context manager myself and back then I did not know the inner working of `beets.config` yet.

Haha, no worries! Thank you very much for figuring out where it goes wrong :)

This does simplify this PR quite a lot then, if I can skip the change of configuration format. Should I make your above change as part of this PR, or should it be a separate change?

beetsplug/substitute.py
nichobi: I was going to make the change, but I guess it's not needed now, as the list format won't be needed :)

beetsplug/substitute.py
nichobi: Okay, I think that would also work for my purposes. It may get a bit messy to define rules a lot of exceptions (I currently only have around 5), but it's doable.

Your `(?i)`s do make me wonder if we should also leave it up to the rules to define case insensitivity. Currently it is enabled for all rules. This would however be a breaking change as the plugin hitherto has enabled it by default, affecting even the example configuration in the docs:
```yaml
substitute:
  .*jimi hendrix.*: Jimi Hendrix
```

snejus: I don't see an issue with ignoring the case - to be honest my rules above would have looked slightly more organised if the patterns were case insensitive 😅

snejus (approved):
This looks great!!

snejus:
Apologies - I forgot to merge this in 😅

PR #5359, opened 2024-07-11
Rewrite the `fish` plugin from scratch

bal-e:
This plugin needed significant reworking.  It now uses `StringIO` to construct the completion script and `pathlib` to manage paths uniformly. Functions now have type annotations and the flow of information through the entire script is much more readable.

More importantly, the actual contents of the script produced have been rewritten entirely.  A more sophisticated parsing function is now used to determine when to show global option completions.  Based on Fish`s source code, a much more comprehensive string escaping function is now used to put data in the completion script.  The completion for metadata values (as provided by `--extravalues`) actually works now: it needed to provide completions for the word the user was in the middle of typing (e.g. `artist:ab` -> `artist:abba`), rather than the word following.

- [x] Documentation
- [ ] Changelog
- [ ] Tests

### DIFF
--- a/beetsplug/fish.py
+++ b/beetsplug/fish.py
@@ -1,6 +1,7 @@
 # This file is part of beets.
 # Copyright 2015, winters jean-marie.
 # Copyright 2020, Justin Mayer <https://justinmayer.com>
+# Copyright 2024, Arav K. <gq28uu827qlnqpgi@bal-e.org>
 #
 # Permission is hereby granted, free of charge, to any person obtaining
 # a copy of this software and associated documentation files (the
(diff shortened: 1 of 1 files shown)


bal-e:
Fish's `complete` builtin offers a `-C` flag which lets you test what completions would be offered for a certain command-line; this could be used to perform automated testing.  I suppose this is doable, by making any such tests conditional on a Fish executable being found.  But I don't know whether that should be a separate PR or not.  This one is already a complete rewrite of the plugin.

bal-e:
cc @Serene-Arc, this follows on from the small changes I made to `fish.py` in #5337.  The more I looked at the code, the more I wanted to rewrite it.

Serene-Arc (changes requested):
I'm not familiar with fish but the code looks fine to me. One thing that seems absolutely paramount is tests though. If I can't understand everything that is being done, I'd advocate for tests normally but here it's paramount.

When I look at this, a programmatically generated shell script that is meant to be autoloaded, that screams 'attack vector' to me. At the very least, we need tests that take the generated script, write it to a file, and then the test compares that written script to one we have a string in the document. That will make it *crystal* clear to developers (and users, if they are inclined to look) what exactly all this code generates, and what changes with each option. Plus, if someone changes the logic, they'll need to change the tests, which will show exactly what will change without any obfuscation.

bal-e:
I agree, a comprehensive testing solution is important.  I was also concerned about the security of these sorts of shell scripts, but I hadn't thought about just comparing the results to a fixed script string in the tests -- I'll definitely implement that.

bal-e:
@snejus, I know you were working on the testing infrastructure in #5362.  I've ended up getting started on the `pytest` migration by defining some shared fixtures here, based on the code from your PR -- can you take a look (at `beets/test/fixtures.py` in particular) to make sure it all checks out?  It can coexist with the existing testing infrastructure so we can migrate over time.

snejus (approved):

snejus:
Sorry - my github script has gona haywire and I accidentally approved this PR 😅

bal-e:
no worries @snejus!

beets/test/fixtures.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # Reset the global configuration state.
+    beets.config.clear()
+    beets.config._materialized = False
+
+
+@pytest.fixture(scope="function")
+def library(
+    tmp_path: Path,
+    config: Configuration,
```
snejus, line 73: `config` fixture is not used within this fixture.

If it was placed here to ensure it runs, consider [making it an autouse fixture](https://docs.pytest.org/en/stable/how-to/fixtures.html#autouse-fixtures-fixtures-you-don-t-have-to-request).


```python
@pytest.fixture(autouse=True)
def config(...)
```

alternatively, if you only need the config setup in some specific tests

```python
# note leading underscore which is pytest convention for defining a fixture that doesn't return anything
@pytest.fixture
def _config(...):
    ...


@pytest.mark.usefixtures("_config")
def test_something(...):
    ...
```

or use it across the entire module
```python
# test/test_something.py
pytestmark = pytest.mark.usefixtures("_config")

...

def test_something(...):
    ...
```

beets/test/fixtures.py
```diff
@@ -0,0 +1,92 @@
...
+    monkeypatch: pytest.MonkeyPatch,
+) -> Iterator[Library]:
+    """
+    A fixture for the Beets library.
+    """
+
+    # Beets needs a location to store library contents.
+    lib_dir = tmp_path / "lib_dir"
+    lib_dir.mkdir(exist_ok=False)
```
snejus, line 82: Instead of `tmp_path` try using [`tmp_path_factory`](https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmp-path-factory-fixture):

```suggestion
    lib_dir = tmp_path_factory.mktemp("lib_dir")
```

beets/test/fixtures.py
```diff
@@ -0,0 +1,92 @@
...
+    monkeypatch.setenv("BEETSDIR", str(lib_dir))
+
+    # Create the Beets library object.
+    lib = Library(":memory:", str(lib_dir))
```
snejus, line 86: Maybe it'd be worth defining `lib_dir` as a fixture to remove the need to cast to `str` twice?

```python
@pytest.fixture(scope="session")
def lib_dir(tmp_path_factory: pytest.TempPathFactory) -> str:
    return str(tmp_path_factory.mktemp("lib_dir"))
```

snejus (changes requested):
Thanks for starting the pytest setup!!! I left a couple of comments there. Will review the stuff in `beetsplug/fish.py` in a bit.

beets/test/fixtures.py
bal-e: I didn't want to do this because these interfaces should be switched over to `pathlib` relatively soon: I want to explicitly leave the `str` casts in so that we can notice and replace them later.

beets/test/fixtures.py
```diff
@@ -0,0 +1,92 @@
+# This file is part of beets.
```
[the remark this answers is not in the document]
bal-e: I chose to put them in the actual `beets` library for the same reason that the `beets.test` module exists at all -- other Python projects directly using `beets` can make use of these fixtures in their own tests.  If we don't foresee that being necessary, I can move this to `conftest.py`, but then we should also plan on removing the entire `beets.test` module in the future.

test/plugins/test_fish.py
```diff
@@ -0,0 +1,67 @@
...
+import optparse
+from pathlib import Path
+
+import pytest
+
+from beets.library import Library
+from beetsplug.fish import FishPlugin
+
+# Tell 'pytest' to pull in Beets' test fixtures.
+pytest_plugins = "beets.test.fixtures"
```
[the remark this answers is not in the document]
bal-e: I was worried introducing `conftest.py` already (when it's just being used for a single module) would be a bit overkill, but yeah, if this is the start of our `pytest` migration, then I should probably add it anyway.

beets/test/fixtures.py
```diff
@@ -0,0 +1,92 @@
...
+
+    return Path(__file__).parents[2] / "test" / "rsrc"
+
+
+@pytest.fixture(scope="function")
+def config(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> Iterator[Configuration]:
+    """
```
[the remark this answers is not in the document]
bal-e: Yep, I usually follow that convention for function descriptions, but I wasn't sure whether it made sense in the case of fixtures (since from a user perspective, they're really some kind of automatically generated _value_, not a function).  I'm happy to switch to something like "Prepare configuration for the library" instead.

(P.S. I personally disagree with the stated syntax convention for one-liners, I like keeping the quotes on their own lines.  But if this is a popular convention then I'm happy to adjust these function descriptions accordingly.)

beets/test/fixtures.py
```diff
@@ -0,0 +1,92 @@
...
+from beets.library import Library
+
+
+@pytest.fixture(scope="session")
+def resource_dir() -> Path:
+    """
+    A fixture returning the resource directory for tests.
+    """
+
+    return Path(__file__).parents[2] / "test" / "rsrc"
```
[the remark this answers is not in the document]
bal-e: Oh, that's awesome, thank you!

beets/test/fixtures.py
bal-e: This is interesting.  I based this fixture on the way the existing testing infrastructure works.  When setting up a `BeetsTestCase`, the line `self.config = beets.config` occurs; this means that `self.config` is basically an alias for the same global variable.  I think the motivation behind this was that the Beets API could one day be re-entrant, relying on a `config` object passed into the `Library` rather than just the global `beets.config`.  This is why the `config` fixture returns a reference to `beets.config`: it is the direct replacement for `self.config` in `BeetsTestCase`.

While the `library` fixture does not _directly_ rely on the `config` fixture, it does rely on the `beets.config` global variable to be reset at the start of every test.  On the other hand, I don't think the `config` fixture should be auto-use; at least for now, I want users to explicitly mention it, so we can gain some understanding of exactly how it is used across the tests.  For the same reason, I don't think it should be renamed to `_config`: the fact that we don't directly reference it in `library()` is more of an implementation detail, and it won't be true for other users of the fixture.

beets/test/fixtures.py
bal-e: Yep, thanks!

beets/test/fixtures.py
snejus: > For the same reason, I don't think it should be renamed to `_config`: the fact that we don't directly reference it in `library()` is more of an implementation detail, and it won't be true for other users of the fixture.

I agree - at the end of the day it __does__ return the configuration object.

The unused `config` argument makes some of the linters screech (for example `flake8-unused-arguments`). You can add `pytest.mark.usefixtures("config")` above this fixture definition. I like `usefixtures` because it communicates that the `library` fixture __only__ requires the `config` fixture set up - it does not care what `config` returns.

beets/test/fixtures.py
snejus: I thought that we're using pytest's `tmp_path_factory` here instead?

snejus: > I didn't want to do this because these interfaces should be switched over to `pathlib` relatively soon: I want to explicitly leave the `str` casts in so that we can notice and replace them later.

I agree that a separate fixture may be an overkill here - consider simply defining a variable `lib_dir = tmp_path_factory.mktemp("lib_dir")` to avoid the double cast and make sure that we only need to a single replacement once we switch to `pathlib`.

beets/test/fixtures.py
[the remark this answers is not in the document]
snejus: Tbh ideally I'd see the docstrings removed from here - you've done a good job naming the fixtures and the implementation is simple - I don't see any reason for the docstring. 

In such cases I find docstrings counter-productive - they just occupy space and add noise 🤷🏼

beets/test/fixtures.py
bal-e: Alright, will do.  Thanks!

beets/test/fixtures.py
bal-e: Unfortunately, `monkeypatch.setenv` [requires](https://docs.pytest.org/en/stable/_modules/_pytest/monkeypatch.html#MonkeyPatch.setenv) a `str`.  That will be true even once we switch over to `pathlib`, so at least one `str` cast will always exist.  What we currently have -- where a second cast is performed for `Library` -- is optimal, since it's the only line that will change.

beets/test/fixtures.py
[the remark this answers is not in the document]
bal-e: Actually, I want to add a bit more (non-obvious) information in these docstrings.  I can't expect every user to read and understand the implementation, even though it is simple.  I'll add some shortly -- if you still think it's not necessary, we can remove it all.

bal-e:
@snejus bump on review?

snejus:
Ah sorry, completely missed out on this! Will have a look now!

beets/test/fixtures.py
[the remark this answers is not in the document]
snejus: This makes sense. I did not realise that `beets.tests` package was being used by other projects!

beets/test/fixtures.py
```diff
@@ -0,0 +1,107 @@
...
+    they should be placed within an appropriately named subdirectory.
+
+    :return: the path to `test/rsrc` in the repository.
+    """
+
+    return pytestconfig.rootpath / "test" / "rsrc"
+
+
+@pytest.fixture
+def config(
```
snejus, line 48: I'm a bit concerned about multiple sources of truth for the same beets config setup in `setup_beets` and this fixture. If these two configurations get out of sync, we are more likely to see unexpected test failures attempting to port tests from `unittest` to `pytest`. This can be a nightmare to debug due to the lack of clear indication of the root cause.

What would you think about defining the logic in your `config` fixture as a context manager in `beets.test.helper`? This way we could reuse it in both `BeetsTestCase` and `pytest` fixtures, and remove the duplicate implementation from `TestHelper.setup_beets`.

For example

```python
# beets/test/helper.py

@contextmanager
def setup_config(directory: Path):
    config = beets.config
    config.clear()
    config.read()

    config["plugins"] = []
    config["verbose"] = 1
    config["ui"]["color"] = False
    config["threaded"] = False
    config["directory"] = str(tmp_path)

    yield config

    self.config.clear()
    beets.config.read(user=False, defaults=True)
```

It encapsulates all setup and cleanup logic and it can be used both in the `setUp` code and `pytest` fixtures:

```python 
# beets/test/helper.py
class TestHelper:
    ...
    def setup_with_context_manager(self, cm):
        val = cm.__enter__()
        self.addCleanup(cm.__exit__, None, None, None)
        return val

    def setUp(self):
        ...
        self.config = self.setup_with_context_manager(setup_config(self.tmp_path))
        ...


# conftest.py
from beets.test.helper import setup_config

@pytest.fixture
def config(tmp_path):
    with setup_config(tmp_path) as config:
        yield config
```

This way we keep a single source of truth for the beets config setup, maintaining consistency regardless of the test runner. This pattern is also applicable to the rest of the setup (lib, importer etc.). What do you think?

A nice side effect of this approach is that we can use `conftest.py` to define our fixtures since the stuff useful to people is defined in `beets.test.helper` and not in the test case itself. This makes the test cases more readable and easier to maintain.

beets/test/fixtures.py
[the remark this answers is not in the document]
snejus: I see, it's good stuff!

snejus (commented):
Leaving a comment regarding beets config setup getting out of sync between unittest and pytest.

Will have a look at the actual fish plugin changes later today!

beets/test/fixtures.py
bal-e: I like this plan, thanks!  Once we're done with `unittest` we can revert back to the existing layout, though.  I'll leave a TODO in the code for it.

beetsplug/fish.py
```diff
@@ -13,329 +14,393 @@
...
+    """,
+]
+
+
+def fish_config_dir() -> Path:
+    """
+    The directory for the user's Fish configuration.
+    """
+
+    config_home: Path
```
snejus, line 134: Consider using `platformdirs`

beetsplug/fish.py
```diff
@@ -13,329 +14,393 @@
...
+    def run(
+        self,
+        lib: library.Library,
+        opts: optparse.Values,
+        args: List[str],
+    ):
+        # Get the user-provided options.
+        include_fields = not getattr(opts, "noFields")
+        extra_comp_fields = cast(List[str], getattr(opts, "extravalues") or [])
+        output = Path(getattr(opts, "output"))
```
snejus, line 298: `noFields` and `output` options define defaults, so they can be accessed as a property I think. Consider configuring `extravalues` with a default too 🙂

beetsplug/fish.py
```diff
@@ -13,329 +14,393 @@
...
+        assert len(args) == 0
```
snejus, line 299:
```suggestion
        assert not args
```

snejus: I think it may be worth adding some message for the users here

beetsplug/fish.py
```diff
@@ -13,329 +14,393 @@
...
-                    if option.help
-                    else ""
-                )
-                cmd_arglist = (
-                    (" -a " + wrap(" ".join(option.choices)))
-                    if option.choices
-                    else ""
+        if extra_comp_fields:
+            # The set of values for every user-specified extra field.
+            extra_values: Dict[str, set[str]] = dict.fromkeys(
```
snejus, line 381: `from __future__ import annotations` may be missing at the top of the file

beetsplug/fish.py
```diff
@@ -13,329 +14,393 @@
...
-    # Need " or ' around strings but watch out if they're in the string
-    sptoken = '"'
-    if ('"') in word and ("'") in word:
-        word.replace('"', sptoken)
-        return '"' + word + '"'
-
-    tok = '"' if "'" in word else "'"
-    return tok + word + tok
+        # Write the buffered text to the file.
+        output.write_text(script.getvalue())
```
snejus, line 406: As a user I would probably expect the completions to be printed to stdout, since that's what most other tools do. For example, some users may be using a different path to store their completions.

snejus (commented):
That's a very clean rewrite! Added a couple of small comments, and a suggestion to print the completions to stdout.

beetsplug/fish.py
bal-e: Oh, shoot.  I completely forgot about the `platformdirs` PR I was supposed to open.  Do you mind if I leave this as-is for now and address it in that PR?

beetsplug/fish.py
bal-e: Yeah, but `mypy` doesn't like it, unfortunately.  I think this also makes sense given that the argument parser dynamically adds new fields to the object.  I'll add a default to `extravalues`, thanks for the tip!

beetsplug/fish.py
bal-e: Oh, I didn't think this could be triggered!  I figured the argument parser would throw up an error earlier.  I tested it myself, and it does occur.  I'll add a user-friendly error message.

beetsplug/fish.py
snejus: Which version of `mypy` are you using? It seems to be happy on my end

image

beetsplug/fish.py
bal-e: That import would activate [PEP 563](https://peps.python.org/pep-0563/).  It's not necessary in this module, and even the Python 3.8 check doesn't complain in CI.  However, if we're instituting it as a convention across the codebase, we need some shared issue to point everybody to about it.

beetsplug/fish.py
snejus: Sounds good. You may be able to drop this altogether if you decide to print the completions to stdout instead of saving them to a file 😉

beetsplug/fish.py
bal-e: It's possible to configure a different path using `-o`.  As a Fish user, I find the current behaviour quite convenient, since I'm not fully familiar with where these completions need to go.  I can add support for printing to `stdout` using `-o -`, but I still think the current behaviour is a good default.  Also note backward compatibility -- users may rely on the current behaviour as part of e.g. periodic scripts.

beetsplug/fish.py
bal-e: Huh, maybe I confused it for my IDE's warnings.  I'll change it back (and fix the `Set` annotation).  Thanks!

beetsplug/fish.py
snejus: image

`mypy` complains about it since it cannot type-check it (this notation is only supported from Python 3.9 onwards). However, the CI is configured to ignore `mypy` errors.

This also fails **FA100** check from `flake8-future-annotations` linter, which nicely describes what it's about:

> # future-rewritable-type-annotation (FA100)
> 
> Derived from the **flake8-future-annotations** linter.
> 
> ## What it does
> Checks for missing `from __future__ import annotations` imports upon
> detecting type annotations that can be written more succinctly under
> PEP 563.
> 
> ## Why is this bad?
> PEP 585 enabled the use of a number of convenient type annotations, such as
> `list[str]` instead of `List[str]`. However, these annotations are only
> available on Python 3.9 and higher, _unless_ the `from __future__ import annotations`
> import is present.
> 
> Similarly, PEP 604 enabled the use of the `|` operator for unions, such as
> `str | None` instead of `Optional[str]`. However, these annotations are only
> available on Python 3.10 and higher, _unless_ the `from __future__ import annotations`
> import is present.
> 
> By adding the `__future__` import, the pyupgrade rules can automatically
> migrate existing code to use the new syntax, even for older Python versions.
> This rule thus pairs well with pyupgrade and with Ruff's pyupgrade rules.
> 
> This rule respects the [`target-version`] setting. For example, if your
> project targets Python 3.10 and above, adding `from __future__ import annotations`
> does not impact your ability to leverage PEP 604-style unions (e.g., to
> convert `Optional[str]` to `str | None`). As such, this rule will only
> flag such usages if your project targets Python 3.9 or below.
> 
> ## Example
> 
> ```python
> from typing import List, Dict, Optional
> 
> 
> def func(obj: Dict[str, Optional[int]]) -> None: ...
> ```
> 
> Use instead:
> 
> ```python
> from __future__ import annotations
> 
> from typing import List, Dict, Optional
> 
> 
> def func(obj: Dict[str, Optional[int]]) -> None: ...
> ```
> 
> After running the additional pyupgrade rules:
> 
> ```python
> from __future__ import annotations
> 
> 
> def func(obj: dict[str, int | None]) -> None: ...
> ```
> 
> ## Options
> - `target-version`

We have this import in every module that uses `list` / `dict` / `set` annotations.

bal-e: Thanks for the detailed explanation!  I've added the import.  However, I still have to use `typing.List` in one place: the explicit `cast(list[str], ...)` on line 299.  Since this expression is evaluated at runtime, Python 3.8 will fail with the standard "'type' object is not subscriptable" error.  As such, I haven't updated any of the types yet.

bal-e:
It looks like the Windows builds are broken because of the `reflink` issue.  @snejus, feel free to review anyway.

snejus:
> It looks like the Windows builds are broken because of the `reflink` issue. @snejus, feel free to review anyway.

You can now rebase as the fix has been merged upstream  🙂

beetsplug/fish.py
snejus: Is that `cast` call required? 

Seems like `mypy` is happy with a simple annotation, and the plugin runs fine
image

bal-e: Hm, I guess it's fine.  Will fix.

test/rsrc/fish/simple.fish
```diff
@@ -0,0 +1,71 @@
+
+# An 'argparse' description for global options for 'beet'.
+function __fish_beet_global_optspec
```
snejus, line 3: I was meant to add a comment to ask to move this to a `.fish` file but forgot. Happy to see that it was done nevertheless!

snejus:
Everything looks good - there's only one bit that's left, see https://github.com/beetbox/beets/pull/5359#pullrequestreview-2246989763. What did we decide to do about the test setup?

bal-e:
I'll implement it now, @snejus!

bal-e:
There seems to be some weird issue with test coverage.  @snejus, any idea what's going on?

snejus:
> There seems to be some weird issue with test coverage. @snejus, any idea what's going on?

I have a feeling that's because the workflow is being run from your fork, where the environmental variable doesn't exist.

Try replacing `pull_request` with `pull_request_target` in `.github/workflows/ci.yaml`?

image

snejus:
@bal-e sorry I didn't realise this was ready! If you rebase the PR off master the coverage should run fine now.

bal-e:
@snejus I tried `pull_request_target`, but it looks like that disabled CI checking for the PR entirely.  Is there any good way to disable specifically the upload-coverage action for PRs?

snejus:
@bal-e see #5479 where I did the same change which later got reverted since I discovered it indeed disabled PR checks 😅 

> Is there any good way to disable specifically the upload-coverage action for PRs?

Haven't yet looked into this... For now, I guess, you can ignore this failure

snejus:
For the future PRs, note that since you're a member of the organization you can push your work directly to a branch in this repo (`beetbox/beets`) - this issue does not come up. That's what I've been doing myself.

beets/test/helper.py
```diff
@@ -194,46 +207,44 @@ def setup_beets(self):
...
+        self.contexts = []
 
-        self.config["plugins"] = []
-        self.config["verbose"] = 1
-        self.config["ui"]["color"] = False
-        self.config["threaded"] = False
+        # Set up the basic configuration with a HOME directory.
+        self.create_temp_dir()
+        temp_dir = Path(os.fsdecode(self.temp_dir))
+        self.libdir = temp_dir / "libdir"
```
snejus, line 216: `self.libdir` starts as a `Path` and later gets cast to `bytes`. As a `Path`, it is only used within this method, so you may want to assign a local variable for it, and set `self.libdir` with `bytes` later:

```python
    libdir = temp_dir / "libdir"
    ...
    self.libdir = util.bytestring_path(libdir)
```

Note the usage of `bytestring_path`: remember the weird magic/bytes/Windows logic? 😅

beets/test/helper.py
```diff
@@ -194,46 +207,44 @@ def setup_beets(self):
...
+        self.libdir = bytes(self.libdir)
 
         # Initialize, but don't install, a DummyIO.
         self.io = _common.DummyIO()
 
     def teardown_beets(self):
-        self.env_patcher.stop()
         self.io.restore()
-        self.lib._close()
+        while len(self.contexts) != 0:
```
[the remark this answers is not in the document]
snejus: I think this should remove the need for the `contexts` list.

snejus (commented):
Nice, looks good. A couple of small comments but we're close to completing this!

snejus:

Issue #5361, opened 2024-07-12
Migrate tests from `unittest` to `pytest`


snejus:
Following our discussion under https://github.com/beetbox/beets/issues/5229, I am adding this issue to track the progress.

### Proposed solution

Make our tests use `pytest` instead of `unittest`.

### Objective

This updates our testing codebase to be based on the currently accepted best practice - `pytest`.

bal-e:
I'd like to take on this, though most of it will have to wait until #5362, #5363, and #5229 are resolved.  Once #5362 is complete, I'm happy to start refactoring some of the smaller test modules in small PR batches.

PR #5362, opened 2024-07-12
Centralize test setup

snejus:
This PR deduplicates shared tests setup that was used across test files. It attempts to centralise the setup into a single source of truth implementation in `beets.test.helper`.

This works towards the eventual migration to `pytest` (#5361) making it easier to replace the tests setup with `pytest` fixtures.

It builds upon the temporary files cleanup in #5345.

### Test setup

* Mostly duplicate test setup implementations in `beets.test._common.TestCase` and `beets.test.helper.TestHelper` were merged into a single implementation as `beets.test.helper.BeetsTestCase`.
* Replaced `unittest.TestCase` and `beets.test.helper.TestHelper` combination with `beets.test.helper.ImportTestCase`
* Refactored `test/test_plugins.py` removing duplicate import files setup.
* Centralised setting up the database on disk.
* Centralised plugin test setup into `beets.test.helpers.PluginMixin`.

### Removals

* Around 1/3 of the removed lines correspond to the removal of `def suite()` functions that were previously used for tests collection.
* Removed redundant `setUp` and `tearDown` methods from many test files given that the functions that they performed had been centralised in `beets.test.helper`.
* Removed redundant library initialisations

### Importing

* Centralised import directory creation (only gets created on the first access).
* Deduplicated `_setup_import_session` implementations in `TerminalImportSessionSetup` and `ImportHelper`.
* Merged `ImportHelper._create_import_dir` and `TestHelper.create_importer` implementations into `ImportHelper.setup_importer`.
* Merged various takes on import files prep into `ImportHelper.prepare_albums_for_import` and `ImportHelper.prepare_album_for_import`.
* Seeing that many tests used it, defined `AsIsImporterMixin` which provides a method `run_asis_importer` to setup _asis_ (non-autotag) importer and run it

### DIFF
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -541,6 +541,8 @@ class ImportHelper:
     autotagging library and several assertions for the library.
     """
 
+    importer: importer.ImportSession
+
     def setup_beets(self, disk=False):
         super().setup_beets(disk)
         self.lib.path_formats = [
@@ -598,35 +600,33 @@ class ImportHelper:
             self.media_files.append(medium)
         self.import_media = self.media_files
 
+    def _get_import_session(self, import_dir: str) -> None:
+        self.importer = ImportSessionFixture(
+            self.lib,
+            loghandler=None,
+            query=None,
+            paths=[import_dir],
+        )
+
     def _setup_import_session(
         self,
         import_dir=None,
-        delete=False,
-        threaded=False,
-        copy=True,
         singletons=False,
         move=False,
         autotag=True,
-        link=False,
-        hardlink=False,
     ):
-        config["import"]["copy"] = copy
-        config["import"]["delete"] = delete
+        config["import"]["copy"] = True
+        config["import"]["delete"] = False
         config["import"]["timid"] = True
         config["threaded"] = False
         config["import"]["singletons"] = singletons
         config["import"]["move"] = move
         config["import"]["autotag"] = autotag
         config["import"]["resume"] = False
-        config["import"]["link"] = link
-        config["import"]["hardlink"] = hardlink
+        config["import"]["link"] = False
+        config["import"]["hardlink"] = False
 
-        self.importer = ImportSessionFixture(
-            self.lib,
-            loghandler=None,
-            query=None,
-            paths=[import_dir or self.import_dir],
-        )
+        self._get_import_session(import_dir or self.import_dir)
 
     def assert_file_in_lib(self, *segments):
         """Join the ``segments`` and assert that this path exists in the
@@ -759,37 +759,17 @@ class TerminalImportSessionFixture(TerminalImportSession):
             raise Exception("Unknown choice %s" % choice)
 
 
-class TerminalImportSessionSetup:
-    """Overwrites ImportHelper._setup_import_session to provide a terminal importer"""
+class TerminalImportMixin(ImportHelper):
+    """Provides_a terminal importer for the import session."""
 
-    def _setup_import_session(
-        self,
-        import_dir=None,
-        delete=False,
-        threaded=False,
-        copy=True,
-        singletons=False,
-        move=False,
-        autotag=True,
-    ):
-        config["import"]["copy"] = copy
-        config["import"]["delete"] = delete
-        config["import"]["timid"] = True
-        config["threaded"] = False
-        config["import"]["singletons"] = singletons
-        config["import"]["move"] = move
-        config["import"]["autotag"] = autotag
-        config["import"]["resume"] = False
-
-        if not hasattr(self, "io"):
-            self.io = _common.DummyIO()
+    def _get_import_session(self, import_dir: str) -> None:
         self.io.install()
         self.importer = TerminalImportSessionFixture(
             self.lib,
             loghandler=None,
             query=None,
             io=self.io,
-            paths=[import_dir or self.import_dir],
+            paths=[import_dir],
         )
 
 
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -39,6 +39,7 @@ import sys
 import unittest
 from contextlib import contextmanager
 from enum import Enum
+from functools import cached_property
 from io import StringIO
 from tempfile import mkdtemp, mkstemp
 from typing import ClassVar
@@ -504,6 +505,12 @@ class ImportHelper:
 
     importer: importer.ImportSession
 
+    @cached_property
+    def import_dir(self):
+        import_dir = os.path.join(self.temp_dir, b"import")
+        os.makedirs(syspath(import_dir), exist_ok=True)
+        return import_dir
+
     def setup_beets(self):
         super().setup_beets()
         self.lib.path_formats = [
@@ -526,10 +533,6 @@ class ImportHelper:
 
         :param count:  Number of files to create
         """
-        self.import_dir = os.path.join(self.temp_dir, b"testsrcdir")
-        if os.path.isdir(syspath(self.import_dir)):
-            shutil.rmtree(syspath(self.import_dir))
-
         album_path = os.path.join(self.import_dir, b"the_album")
         os.makedirs(syspath(album_path))
 
@@ -595,14 +598,10 @@ class ImportHelper:
         Copies the specified number of files to a subdirectory of
         `self.temp_dir` and creates a `ImportSessionFixture` for this path.
         """
-        import_dir = os.path.join(self.temp_dir, b"import")
-        if not os.path.isdir(syspath(import_dir)):
-            os.mkdir(syspath(import_dir))
-
         album_no = 0
         while album_count:
             album = util.bytestring_path(f"album {album_no}")
-            album_dir = os.path.join(import_dir, album)
+            album_dir = os.path.join(self.import_dir, album)
             if os.path.exists(syspath(album_dir)):
                 album_no += 1
                 continue
@@ -639,7 +638,7 @@ class ImportHelper:
         config["import"]["resume"] = False
 
         return ImportSessionFixture(
-            self.lib, loghandler=None, query=None, paths=[import_dir]
+            self.lib, loghandler=None, query=None, paths=[self.import_dir]
         )
 
     def assert_file_in_lib(self, *segments):
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -151,9 +151,11 @@ class TestHelper(_common.Assertions):
     fixtures.
     """
 
+    db_on_disk: ClassVar[bool] = False
+
     # TODO automate teardown through hook registration
 
-    def setup_beets(self, disk=False):
+    def setup_beets(self):
         """Setup pristine global configuration and library for testing.
 
         Sets ``beets.config`` so we can safely use any functionality
@@ -199,7 +201,7 @@ class TestHelper(_common.Assertions):
         os.mkdir(syspath(self.libdir))
         self.config["directory"] = os.fsdecode(self.libdir)
 
-        if disk:
+        if self.db_on_disk:
             dbpath = util.bytestring_path(self.config["library"].as_filename())
         else:
             dbpath = ":memory:"
@@ -538,8 +540,8 @@ class ImportHelper:
 
     importer: importer.ImportSession
 
-    def setup_beets(self, disk=False):
-        super().setup_beets(disk)
+    def setup_beets(self):
+        super().setup_beets()
         self.lib.path_formats = [
             ("default", os.path.join("$artist", "$album", "$title")),
             ("singleton:true", os.path.join("singletons", "$title")),
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -520,19 +520,14 @@ class BeetsTestCase(unittest.TestCase, TestHelper):
         self.teardown_beets()
 
 
-class LibTestCase(BeetsTestCase):
+class ItemInDBTestCase(BeetsTestCase):
     """A test case that includes an in-memory library object (`lib`) and
     an item added to the library (`i`).
     """
 
     def setUp(self):
         super().setUp()
-        self.lib = beets.library.Library(":memory:")
         self.i = _common.item(self.lib)
-
-    def tearDown(self):
-        self.lib._connection().close()
-        super().tearDown()
 
 
 class ImportHelper:
@@ -985,3 +980,4 @@ class CleanupModulesMixin:
         """Remove files created by the plugin."""
         for module in cls.modules:
             clean_module_tempdir(module)
+        self.lib = beets.library.Library(":memory:")
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -43,7 +43,7 @@ from functools import cached_property
 from io import StringIO
 from pathlib import Path
 from tempfile import mkdtemp, mkstemp
-from typing import ClassVar
+from typing import Any, ClassVar
 from unittest.mock import patch
 
 import responses
@@ -497,6 +497,19 @@ class PluginMixin:
         Album._types = getattr(Album, "_original_types", {})
         Item._queries = getattr(Item, "_original_queries", {})
         Album._queries = getattr(Album, "_original_queries", {})
+
+    @contextmanager
+    def configure_plugin(self, config: list[Any] | dict[str, Any]):
+        if isinstance(config, list):
+            beets.config[self.plugin] = config
+        else:
+            for key, value in config.items():
+                beets.config[self.plugin][key] = value
+        self.load_plugins(self.plugin)
+
+        yield
+
+        self.unload_plugins()
 
 
 class PluginTestCase(PluginMixin, BeetsTestCase):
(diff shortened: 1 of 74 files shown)


test/test_plugins.py
```diff
@@ -266,8 +222,12 @@ def import_task_created_event(self, session, task):
         self.assertEqual(
             logs,
             [
-                "Singleton: {}".format(displayable_path(self.file_paths[0])),
-                "Singleton: {}".format(displayable_path(self.file_paths[1])),
+                "Singleton: {}".format(
+                    displayable_path(self.import_media[0].path)
+                ),
```
bal-e, line 227: This definitely looks better than using an f-string, but I can't wait for that to change with `pathlib` integration.

test/plugins/test_filefilter.py
```diff
@@ -91,31 +94,28 @@ def __create_import_dir(self, count):
...
             metadata["track"] = i + 5
-            metadata["title"] = "Tag Title Misc %d" % (i + 1)
+            metadata["title"] = "Tag Track Misc %d" % (i + 1)
             track_file = bytestring_path("track_%d.mp3" % (i + 1))
             dest_path = os.path.join(self.misc_path, track_file)
             self.__copy_file(dest_path, metadata)
             self.misc_paths.append(dest_path)
 
-    def __run(self, expected_lines, singletons=False):
+    def _run(self, expected_lines):
```
bal-e, line 103: `__run` has been renamed to `_run`, but then shouldn't `__create_import_dir` also be renamed to `_create_import_dir`?

test/plugins/test_replaygain.py
```diff
@@ -51,6 +52,36 @@ def reset_replaygain(item):
...
+        # Implemented by Mixins, see above. This may decide to skip the test.
+        self.test_backend()
+
+        super().setUp()
+        self.config["replaygain"]["backend"] = self.backend
+
+        try:
+            self.load_plugins("replaygain")
+        except Exception:
+            self.tearDown()
```
bal-e, line 69: Why not use `PluginTestCase` here?  And why do you have to manually call `tearDown()` if an exception occurs? (won't `pytest` do that for you automatically?)

bal-e (commented):
This is an incredibly comprehensive PR!  I've left a few notes, but everything looks good.  Perhaps you could use `pathlib` in the functions you've rewritten, so that they don't need to be migrated later?  Or we can leave that for the later `pathlib`-specific PR(s).

Serene-Arc:
I'll try and get my review in over the next day or two! I want to really sit down and go through it all.

test/plugins/test_filefilter.py
snejus: Realised I missed out on removing the import files setup from here. Ended up refactoring this module significantly - the aforementioned method is now gone and the tests should be more clear to follow.

test/plugins/test_embedart.py
```diff
@@ -47,21 +47,15 @@ class EmbedartCliTest(TestHelper, FetchImageHelper):
     abbey_differentpath = os.path.join(_common.RSRC, b"abbey-different.jpg")
 
     def setUp(self):
```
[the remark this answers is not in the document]
snejus: This one's an interesting one - it doesn't seem to run `setup_beets` and thus doesn't seem to require the entire temporary directory / lib setup.

I think it only needs a few methods that `TestHelper` provides. Though now that you brought this to attention, I can see that this test class is actually being ignored:

```python
class EmbedartCliTest(TestHelper, FetchImageHelper):
```

It doesn't inherit from `unittest.TestCase`, so it doesn't get picked up. Will sort this out and check whether we have more cases like this one.

test/test_plugins.py
snejus: > I can't wait for that to change with `pathlib` integration.

Oh yes, completely. This will remove amounts of boilerplate from `beets.util`.

test/plugins/test_replaygain.py
snejus: You will see there are a more test modules which run `load_plugins` and `unload_plugins` manually and they all have something in common: they set up plugin configuration before loading it.

Some of them test various configurations, like `test/plugin/test_filtefilter.py`, where the plugin is loaded midway through the test.

`PluginMixin / PluginTestCase` loads the plugin at the end of the setup, before the test methods runs, therefore plugin configuration defined in the test method simply has no effect. This mixin is only useful for those plugin tests that simply need the plugin to be loaded (majority of them). Otherwise, as you see here, the plugin is loaded manually.

I think it may be a good idea to define a method `PluginMixin.reload_plugin` which does something like

```python
class PluginMixin:
    ...
    def reload_plugin(self, config: dict):
        self.unload_plugins()
        beets.config[self.plugin].update(config)
        self.load_plugins(self.plugin)
```

snejus:
> Perhaps you could use `pathlib` in the functions you've rewritten, so that they don't need to be migrated later? Or we can leave that for the later `pathlib`-specific PR(s).

That's a very good point. I was careful to not try and refactor anything to do with paths (out of scope here, and the PR size would have quickly gotten out of hand) but I think it's a good idea to use `pathlib` in rewrites/new stuff.

For now I made the new methods in `ImportHelper` to use `pathlib` and you can see it's already made it cleaner!

test/plugins/test_embedart.py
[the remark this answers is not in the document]
bal-e: It might not need the additional setup, but it might be worth it to inherit from `BeetsTestCase` just for consistency with other tests.  Otherwise, there should be a comment explaining why `BeetsTestCase` should _not_ be inherited from.

test/plugins/test_replaygain.py
bal-e: Right, thank you for explaining!  But would a `test_plugin_load()` method (set config; load plugin; unload plugin) be more appropriate, if tests call `reload_plugin()` to try loading a plugin temporarily?

bal-e:
> That's a very good point. I was careful to not try and refactor anything to do with paths (out of scope here, and the PR size would have quickly gotten out of hand) but I think it's a good idea to use `pathlib` in rewrites/new stuff.

I agree.  When we fully migrate to `pytest`, we can use its tempdir fixtures which already use `pathlib`.  That should also end up migrating the majority of the tests to `pathlib`.

test/plugins/test_embedart.py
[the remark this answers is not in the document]
snejus: See #5363 where I enabled this test and a couple of others that are skipped.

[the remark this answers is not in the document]
snejus: @bal-e, maybe it would be a good idea to move the the functionality that this test needs away from `BeetsTestCase` to a `LibMixin` (or similar)? There are a couple of other tests (namely `test_query.py`) that are based from `BeetsTestCase` but they actually do not require the entire temporary dir setup which slows them down slightly.

test/plugins/test_replaygain.py
snejus: @bal-e I'm not sure whether I understood what you meant correctly. Maybe something like

```python
from contextlib import contextmanager

...

class PluginMixin:
    @contextmanager
    def configure_plugin(self, config):
        beets.config[self.plugin].update(config)
        self.load_plugins(self.plugin)

        yield

        self.unload_plugins()

class PluginTest(PluginMixin, ...):
    def test_plugin_config(self):
        with self.configure_plugin({"option": "value"}):
            self.assertEqual(beets.config[self.plugin]["option"].get(), "value")
```

This would clean up the configuration on test exit.

test/plugins/test_embedart.py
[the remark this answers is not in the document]
snejus: Ignore my message above - it actually requires the entire setup! I have now made it inherit from `BeetsTestCase`.

test/test_importer.py
```diff
@@ -1755,96 +1679,60 @@ def test_reimported_album_not_preserves_flexattr(self):
...
-            self.importer.run()
-
-        logs = [line for line in logs if not line.startswith("Sending event:")]
+            importer.run()
 
         self.assertEqual(len(self.lib.items()), 0)
         self.assertEqual(len(self.lib.albums()), 0)
 
-        return logs
+        return {line for line in logs if not line.startswith("Sending event:")}
```
bal-e, line 1705: Why have you made the log list a set?  Doesn't that obscure the order of the logs and whether they were produced multiple times?

bal-e:
I like the new uses of `pathlib` and f-strings, the code feels a lot cleaner.  It also looks like the net line change is becoming even more negative, which is awesome.  Good job @snejus!

test/plugins/test_mbsubmit.py
```diff
@@ -52,8 +48,8 @@ def test_print_tracks_output(self):
         # Manually build the string for comparing the output.
         tracklist = (
             "Open files with Picard? "
-            "01. Tag Title 1 - Tag Artist (0:01)\n"
-            "02. Tag Title 2 - Tag Artist (0:01)"
+            "01. Tag Track 1 - Tag Artist (0:01)\n"
+            "02. Tag Track 2 - Tag Artist (0:01)"
```
Serene-Arc, line 52: Double checking here, what's the reason for this change? We're changing what the test compares to, which means it should have been failing before if there's no change to the program code. If it wasn't, then that's an internal rule in our tests that should be marked.

test/plugins/test_replaygain.py
Serene-Arc: I'll point out that this context manager approach is very, very similar to the pytest approach with fixtures. If we do that, it'll be less of a change when converting the remainder of the files.

beets/test/helper.py
```diff
@@ -490,99 +456,164 @@ def touch(self, path, dir=None, content=""):
...
-    def _create_import_dir(self, count=3):
-        """Creates a directory with media files to import.
-        Sets ``self.import_dir`` to the path of the directory. Also sets
-        ``self.import_media`` to a list :class:`MediaFile` for all the files in
-        the directory.
+    def prepare_track_for_import(
+        self,
+        track_id: int,
+        album_path: Path,
+        album_id: int | None = None,
```
Serene-Arc, line 546: I had a big comment written out but then I looked it up and PEP 604 was backported through the futures. I'll have to update my typing habits. Great that you're using the more modern way, we'll have to make the change through the codebase (or at least the parts with typing).

test/test_importer.py
Serene-Arc: I don't think order should matter in this specific part, but the duplicate entry issue is something we should test for.

Serene-Arc (changes requested):

Serene-Arc:
I think it would be good if I opened a pathlib PR (as I keep saying I will) after this PR is done. Like this PR, it will have a lot of small line changes that are just replacing function calls and simplifying, but this PR being merged will make that shorter and easier to parse.

test/plugins/test_replaygain.py
snejus: @Serene-Arc completely agree. I have done this in 28ae414.

test/test_importer.py
snejus: Since I made the import dir to use the import directory as it is, it didn't seem possible to predict whether the album or the single gets picked up first (sometimes it's former and sometimes it's the latter).

snejus: @Serene-Arc I think we could assert the number of imported items/albums to verify that duplicate imports are not happening. See `test_filefilter` where I replaced the logs assertions by library assertions.

test/plugins/test_mbsubmit.py
snejus: See 5643358. We had two similar ways how the import files were being prepared, one of them used **Title** in the item title, the other one used **Track**. I found that **Track** is used in more tests, so I removed the other one and updated the tests accordingly.

snejus:
> I think it would be good if I opened a pathlib PR (as I keep saying I will) after this PR is done. Like this PR, it will have a lot of small line changes that are just replacing function calls and simplifying, but this PR being merged will make that shorter and easier to parse.

Sounds like a good plan!

snejus:
I changed this PR to be based on #5363, which is in turn based on #5345:

#5345 <- #5363 <- #5362.

#5345 is ready and can be merged once approved.
#5363 needs a bit more work regarding the windows runner. I will be away from today until 28th so I will come back to it then. Once #5363 is merged, this one follows!

test/test_importer.py
Serene-Arc: That would be a good solution I think, it would also skip the unlikely situation where the logs are wrong, which would be a separate issue but it's better to make sure that there's actually no duplicate imports.

test/plugins/test_mbsubmit.py
Serene-Arc: Cool, thanks for explaining that.

Serene-Arc:
Ah, great. I'll try and get a review in for #5345 over the next day or two so we can get that merged and get the ball rolling on these PRs.

snejus:
> I changed this PR to be based on #5363, which is in turn based on #5345:
> 
> #5345 <- #5363 <- #5362.
> 
> #5345 is ready and can be merged once approved. #5363 needs a bit more work regarding the windows runner. I will be away from today until 28th so I will come back to it then. Once #5363 is merged, this one follows!

@Serene-Arc actually ignore this - I'm not sure what made me to base this PR off #5363 as it actually does not depend on any changes from there. I now rebased this PR off master, so it's ready to be reviewed and tests should be happy.

test/plugins/test_hook.py
```diff
@@ -13,162 +13,98 @@
...
+            self.assertIn(
+                "hook: hook for test_event exited with status 1", logs
+            )
 
     def test_hook_non_existent_command(self):
-        self._add_hook("test_event", "non-existent-command")
+        with self._configure_logs("non-existent-command") as logs:
+            self.assertTrue(
+                any(
+                    message.startswith("hook: hook for test_event failed: ")
```
bal-e, line 61: Can we implement this as `self.assertIn` instead?  What does the error message contained here look like, and how does it vary?  If this style is necessary, a comment explaining the variation in the message is needed.

bal-e (approved):
This is looking really good!  I've outlined a few nitpicks, but overall this is awesome.  @snejus, I love the amount of code clean-up you've managed in this ginormous PR, good job!

test/test_importer.py
snejus: > Since I made the import dir to use the import directory as it is, it didn't seem possible to predict whether the album or the single gets picked up first (sometimes it's former and sometimes it's the latter).

I have reviewed this and found I am telling lies here: the import sequence is predictable - the importer picks up whichever is first in the alphabet, in this case album.

I have now made this a list.

test/plugins/test_hook.py
snejus: I think this kind of improvement may be out of scope of the structural change I did here

test/plugins/test_hook.py
```diff
@@ -13,162 +13,98 @@
...
-
-
-def suite():
-    return unittest.TestLoader().loadTestsFromName(__name__)
-
-
-if __name__ == "__main__":
-    unittest.main(defaultTest="suite")
+        self.paths = [p.encode() for p in self.paths]
+        self._test_command(lambda *_: "{path}", send_path_kwarg=True)
```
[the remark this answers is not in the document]
snejus: Nope, Python won't accept two arguments that have the same name: 
image

In this case we want to scrap all arguments, so I think `*_ / *args` may be the most common way to do so: https://www.perplexity.ai/search/python-ignore-positional-argum-f9pNbJEiR76a6eVlJKrgPA

test/plugins/test_hook.py
```diff
@@ -13,162 +13,98 @@
...
-    @unittest.skipIf(sys.platform == "win32", "win32")
-    def test_hook_no_arguments(self):
-        temporary_paths = [
-            get_temporary_path() for i in range(self.TEST_HOOK_COUNT)
+    def setUp(self):
+        super().setUp()
+        temp_dir = os.fsdecode(self.temp_dir)
+        self.paths = [os.path.join(temp_dir, e) for e in self.events]
+
+    def _test_command(
```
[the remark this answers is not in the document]
snejus: Added a docstring with some details

snejus:
I'm back! @bal-e addressed your comments!

test/plugins/test_hook.py
bal-e: Alright, could you leave a TODO comment so that somebody can pick it up in the future?

test/plugins/test_hook.py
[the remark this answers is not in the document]
bal-e: Alright, cool.  Leave it as it is.

test/plugins/test_hook.py
snejus: Had a quick look into this and got it done!

bal-e (approved):
Amazing work @snejus!

Serene-Arc:
I have no objections! Merge whenever you'd like @snejus

PR #5363, opened 2024-07-16
Enable skipped tests

snejus:
Under #5362 me and @bal-e found out that `EmbedartCliTest` does not run since it does not
inherit from `unittest.TestCase`.

This PR fixes the above and another test that also had the same issue:
`LRCLibIntegrationTest`.

Additionally, I went ahead and removed `unittest.skip` from tests that are skipped because
of missing dependencies, and added installation instructions for those dependencies in the
build.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,7 +33,12 @@ jobs:
         if: matrix.platform == 'ubuntu-latest'
         run: |
           sudo apt update
-          sudo apt install ffmpeg gobject-introspection libgirepository1.0-dev
+          sudo apt install \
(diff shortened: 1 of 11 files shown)


snejus:
Now I'm waiting to see failures from the `windows` runner and will try to install the dependencies on this runner.

Serene-Arc:
Given that this isn't passing the tests and #5362 is much larger, I think we should merge the other first.

semohr:
@snejus Is that still relevant? I think lots of things changed since opening this :)

snejus:
Yes,but needs rebasing. Will re-open once it's done.

Issue #5314, opened 2024-06-16
Undocumented config feature not in the docs: overwrite_null


RollingStar:
See here for info on overwrite_null:

https://github.com/beetbox/beets/pull/4987

This is a good first issue / bitesize! Even something as simple as a link to that issue, from the config docs page, would be helpful for future coders.

bal-e:
Should the documentation basically just be "These config options allow overwriting blank metadata fields for albums or tracks?"

RollingStar:
Honestly, it's beyond me. Whatever the code and past comments say.

PR #5373, opened 2024-07-31
Document `overwrite_null`

bal-e:
Solves #5314.

I'm not entirely sure what this documentation should look like, especially the examples (I've never personally had to use this feature anyway).  Still, this is better than leaving it undocumented.

### DIFF
--- a/docs/reference/config.rst
+++ b/docs/reference/config.rst
@@ -281,11 +281,11 @@ That is, if this option is turned on, then ``year`` will always equal
 overwrite_null
 ~~~~~~~~~~~~~~
 
-In some situations, data for a particular field is not available, and it ends up being
-set to an empty value (an empty string).  To allow such fields to be overwritten once
-the appropriate data does become available (e.g. by a re-import or a manual update),
-the name of the metadata field can be added to the list `album` (for albums) or `track`
-(for tracks).  For example::
+This confusingly-named option indicates which fields have meaningful `null` values.  If
+an album or track field is in the corresponding list, then an existing value for this
+field in an item in the database can be overwritten with `null`.  By default, however,
+`null` is interpreted as information about the field being unavailable, so it would not
+overwrite existing values.  For example::
 
     overwrite_null:
         album: ["albumid"]
--- a/docs/reference/config.rst
+++ b/docs/reference/config.rst
@@ -276,6 +276,21 @@ Either ``yes`` or ``no``, indicating whether matched albums should have their
 That is, if this option is turned on, then ``year`` will always equal
 ``original_year`` and so on. Default: ``no``.
 
+.. _overwrite_null:
+
+overwrite_null
+~~~~~~~~~~~~~~
+
+In some situations, data for a particular field is not available, and it ends up being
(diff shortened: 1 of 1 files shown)


docs/reference/config.rst
```diff
@@ -276,6 +276,21 @@ Either ``yes`` or ``no``, indicating whether matched albums should have their
 That is, if this option is turned on, then ``year`` will always equal
 ``original_year`` and so on. Default: ``no``.
 
+.. _overwrite_null:
+
+overwrite_null
```
snejus, line 281: See the logic where this configuration is being utilized:

```python
SPECIAL_FIELDS = {
    "album": (
        "artist", "artist_credit", "artist_id", "artists", "artists_credit",
        "artists_ids", "artist_sort", "artists_sort", "data_url",

        "album_id", "day", "mediums", "month",
        "releasegroup_id", "tracks", "va", "year",
    ),
    "track": (
        "artist", "artist_credit", "artist_id", "artists", "artists_credit",
        "artists_ids", "artist_sort", "artists_sort", "data_url",

        "index", "length", "medium", "medium_index", "medium_total",
        "release_track_id", "title", "track_alt", "track_id",
    ),
}

...

        # Don't overwrite fields with empty values unless the
        # field is explicitly allowed to be overwritten
        for field, value in album_info.items():
            if field in SPECIAL_FIELDS["album"]:
                continue
            clobber = field in config["overwrite_null"]["album"].as_str_seq()
            if value is None and not clobber:
                continue
            item[field] = value

        for field, value in track_info.items():
            if field in SPECIAL_FIELDS["track"]:
                continue
            clobber = field in config["overwrite_null"]["track"].as_str_seq()
            value = getattr(track_info, field)
            if value is None and not clobber:
                continue
            item[field] = value
```

To be honest, this configuration should rather be called `apply_null_value` to
slightly clarify what it does.

In short, this has to do with applying empty metadata, i.e. fields that have value `None`.

Let's say you're reimporting an album which currently has `field="something"`.
What happens to this field when you apply metadata that has `field=None`?

**By default**, `None` is treated like an **absence of value**: no update takes place thus the current field value is preserved

If the field is configured in **`overwrite_null`** config, `None` is seen as a **value**: thus the field value is cleared/reset to `None`. Note that many fields aren't allowed to be configured here - see the **SPECIAL_FIELDS** dictionary.

snejus (changes requested):

docs/reference/config.rst
bal-e: Oof, thanks @snejus, I completely misinterpreted this!  I thought it affected whether a metadata field that is already `None` can be overwritten, rather than whether a metadata field can be overwritten with `None`.  Thanks for the catch, I'll rewrite the documentation.

I agree that the name of the option should be changed, but I don't know what the procedure for that is, particularly because of backward compatibility requirements.  I can leave a TODO here, open an issue, and leave it for later.

snejus (approved):
Looks great!

bal-e:
Thanks @snejus!  I can't merge it myself for some reason, please do.

snejus:
You should be able to do it now!

PR #5376, opened 2024-08-01
Fix the no_convert option of the convert plugin stopping conversion when there is only a partial match.

sanpoChew:
## Description

I was running the `convert` plugin with the following config:

```
no_convert: samplerate:..48000 bitdepth:..16
```

but anything that was 24/48 was also not being converted, this was due to the code returning `False` for `should_transcode` if any part of the query matches, rather than considering the whole query. This meant that `bitdepth:...16` was being ignored.

I have changed this so the `no_convert` value is considered as one query.

## To Do

- [X] ~Documentation.~
- [X] Changelog.
- [X] Tests.

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -84,17 +84,20 @@ def get_format(fmt=None):
 
     return (command.encode("utf-8"), extension.encode("utf-8"))
 
+def in_no_convert(item: Item) -> bool:
+    no_convert_query = config["convert"]["no_convert"].as_str()
 
+    if no_convert_query:
+        query, _ = parse_query_string(no_convert_query, Item)
+        return query.match(item)
+    else:
+        return False
+    
 def should_transcode(item, fmt):
     """Determine whether the item should be transcoded as part of
     conversion (i.e., its bitrate is high or it has the wrong format).
     """
-    no_convert_query = config["convert"]["no_convert"].as_str()
-    if no_convert_query:
-        query, _ = parse_query_string(no_convert_query, Item)
-        if query.match(item):
-            return False
-    if (
+    if in_no_convert(item) or (
         config["convert"]["never_convert_lossy_files"]
         and item.format.lower() not in LOSSLESS_FORMATS
     ):
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -89,12 +89,11 @@ def should_transcode(item, fmt):
     """Determine whether the item should be transcoded as part of
     conversion (i.e., its bitrate is high or it has the wrong format).
     """
-    no_convert_queries = config["convert"]["no_convert"].as_str_seq()
-    if no_convert_queries:
-        for query_string in no_convert_queries:
-            query, _ = parse_query_string(query_string, Item)
-            if query.match(item):
-                return False
+    no_convert_query = config["convert"]["no_convert"].as_str()
+    if no_convert_query:
+        query, _ = parse_query_string(no_convert_query, Item)
+        if query.match(item):
+            return False
     if (
         config["convert"]["never_convert_lossy_files"]
         and item.format.lower() not in LOSSLESS_FORMATS
--- a/test/plugins/test_convert.py
+++ b/test/plugins/test_convert.py
@@ -18,10 +18,13 @@ import os.path
 import re
 import sys
 import unittest
+import pytest
 
 from mediafile import MediaFile
 
 from beets import util
+from beetsplug import convert
+from beets.library import Item
 from beets.test import _common
 from beets.test.helper import (
     AsIsImporterMixin,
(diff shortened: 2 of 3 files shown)


bal-e:
I'm looking at the documentation for `no_convert`:

> Does not transcode items matching the query string provided (see [Queries](https://beets.readthedocs.io/en/stable/reference/query.html)). For example, to not convert AAC or WMA formats, you can use `format:AAC, format:WMA` or `path::\.(m4a|wma)$`. If you only want to transcode WMA format, you can use a negative query, e.g., `^path::\.(wma)$`, to not convert any other format except WMA.

I understand that the comma should work as a logical OR operator.  @sanpoChew, can you confirm that `format:AAC, format:WMA` works as the documentation suggests, with your patches applied?

Also, this may break users' configurations, please add some information in the changelog so that users know how to adjust.

Serene-Arc (approved):
Thanks for the PR, and for catching this bug! This seems like a good fix.

Serene-Arc (changes requested):
Actually, just one request. Can you add a test that confirms that the comma OR functionality still works as expected with this change? Just as a test-coverage thing.

After that, if @bal-e is fine, this can be merged.

sanpoChew:
> I'm looking at the documentation for `no_convert`:
> 
> > Does not transcode items matching the query string provided (see [Queries](https://beets.readthedocs.io/en/stable/reference/query.html)). For example, to not convert AAC or WMA formats, you can use `format:AAC, format:WMA` or `path::\.(m4a|wma)$`. If you only want to transcode WMA format, you can use a negative query, e.g., `^path::\.(wma)$`, to not convert any other format except WMA.
> 
> I understand that the comma should work as a logical OR operator. @sanpoChew, can you confirm that `format:AAC, format:WMA` works as the documentation suggests, with your patches applied?
> 
> Also, this may break users' configurations, please add some information in the changelog so that users know how to adjust.

That is correct yes, if users want to maintain the existing behaviour they would need to add a comma. I have updated the changelog entry so that it is more helpful 👍

sanpoChew:
> Actually, just one request. Can you add a test that confirms that the comma OR functionality still works as expected with this change? Just as a test-coverage thing.
> 
> After that, if @bal-e is fine, this can be merged.

That makes, sense have added that test 👍

bal-e (approved):
Sorry for the delay, @sanpoChew.  LGTM!

test/plugins/test_convert.py
```diff
@@ -347,3 +347,59 @@ def test_transcode_from_lossy_prevented(self):
...
+                "mp3": self.tagged_copy_cmd("mp3"),
+            },
+        }
+
+    def test_no_transcode_when_no_convert_set(self):
+        [item] = self.add_item_fixtures(ext="ogg")
+        with control_stdin("y"):
+            self.run_convert_path(item.path)
+        converted = os.path.join(self.convert_dest, b"converted.mp3")
+        self.assertFalse(os.path.exists(converted))
```
snejus, line 375: Once you rebase off/merge `master`, you will find that `unittest`-specific assertions have been replaced with `assert != ...` statements as we're migrating to `pytest`. SO it would be good to update your tests accordingly

snejus (changes requested):

test/plugins/test_convert.py
sanpoChew: I have updated these assertions

sanpoChew:
There seems to be some failures unrelated to my code in the lint checks

snejus:
> There seems to be some failures unrelated to my code in the lint checks

Seems like this was caused by GitHub failing DNS resolution internally. Have re-ran the workflows and it worked fine now.

test/plugins/test_convert.py
```diff
@@ -335,3 +335,59 @@ def test_transcode_from_lossy_prevented(self):
...
+
+    def test_no_transcode_when_or_query_no_convert_set_partial_match(self):
+        self.config["convert"]["no_convert"] = "format:OGG , bitrate:..256"
+        [item] = self.add_item_fixtures(ext="ogg")
+        item.bitrate = 320
+        item.store()
+        with control_stdin("y"):
+            self.run_convert_path(item.path)
+        converted = os.path.join(self.convert_dest, b"converted.mp3")
+        self.assertNotExists(converted)
```
snejus, line 393: There is a lot of duplication and boilerplate in this test class. This makes it hard to tell what specifically is being tested and the difference between each test case.

Given that you're only interested in the `no_convert` configuration logic, it's best to isolate it to a separate function (see my suggestion above) and create a focused unit test, for example:

```suggestion
class TestNoConvert:
    """Test the effect of the `no_convert` option."""

    @pytest.mark.parametrize(
        "config_value, should_skip",
        [
            ("", False),
            ("bitrate:320", False),
            ("bitrate:320 format:ogg", False),
            ("bitrate:320 , format:ogg", True),
        ],
    )
    def test_no_convert_skip(self, config_value, should_skip):
        item = Item(format="ogg", bitrate=256)

        convert.config["convert"]["no_convert"] = config_value

        assert convert.in_no_convert(item) == should_skip
```

beetsplug/convert.py
```diff
@@ -89,12 +89,11 @@ def should_transcode(item, fmt):
     """Determine whether the item should be transcoded as part of
     conversion (i.e., its bitrate is high or it has the wrong format).
     """
-    no_convert_queries = config["convert"]["no_convert"].as_str_seq()
-    if no_convert_queries:
-        for query_string in no_convert_queries:
-            query, _ = parse_query_string(query_string, Item)
-            if query.match(item):
-                return False
+    no_convert_query = config["convert"]["no_convert"].as_str()
```
snejus, line 92: I'd suggest creating a separate function for this check, something like:

```python
def in_no_convert(item: Item) -> bool:
    # check if item is excluded by the configured query
    ...

def should_transcode(item, fmt):
    """Determine whether the item should be transcoded as part of
    conversion (i.e., its bitrate is high or it has the wrong format).
    """
    if in_no_convert(item) or (
        config["convert"]["never_co...
```

This simplifies testing a lot, see the comment in the test file

snejus (changes requested):
Suggested something which should simplify testing

test/plugins/test_convert.py
sanpoChew: Thanks a lot for this that is much nicer, I have created the `in_no_convert` function and updated the tests 👍

snejus (dismissed):

snejus (approved):
Amazing stuff, looks super clean now! Thanks for catching this.

sanpoChew:
fantastic! thanks a lot for the guidance all 👍

PR #5379, opened 2024-08-04
Remove redundant unicode decoding

snejus:
These days musicbrainz returns unicode strings as expected.

### DIFF
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -168,42 +168,6 @@ class AlbumInfo(AttrDict):
         self.discogs_artistid = discogs_artistid
         self.update(kwargs)
 
-    # Work around a bug in python-musicbrainz-ngs that causes some
-    # strings to be bytes rather than Unicode.
-    # https://github.com/alastair/python-musicbrainz-ngs/issues/85
-    def decode(self, codec: str = "utf-8"):
-        """Ensure that all string attributes on this object, and the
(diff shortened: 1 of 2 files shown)


bal-e (approved):
This is great, we get to remove redundant code in the core library.  I'm happy with the changes, and clearly the tests pass.

My only question is, since `beets` is technically a library others can import, does this count as an API breaking change?  Or do we count this as implementation code that we can change freely?

bal-e:
Also, I just wanted to leave a link to <https://github.com/alastair/python-musicbrainzngs/issues/85> here for GitHub to notice.

snejus:
> My only question is, since `beets` is technically a library others can import, does this count as an API breaking change? Or do we count this as implementation code that we can change freely?

The way `decode` method was defined here indeed looks like something that is exposed, especially given that `AlbumInfo` and `TrackInfo` objects are imported by external autotaggers (for example [beetcamp](https://github.com/snejus/beetcamp) imports them in multiple modules).

On the other hand, if we consider where does this decoding functionality come into play why was it here, we find that it's only relevant to the musicbrainz autotagger, and I guess we can assume that there's no plugin that provides another autotagger based on musicbrainz 😅 

Regarding code design and best practices, this kind of logic should have ideally been defined somewhere else, i.e. in `beets/autotag/mb.py` since that's the only module that needs it and we'd know it's safe to remove it.

I did a quick github code search regardless `gh search code info.decode --language python` and luckily did not find anything relevant!

PR #5383, opened 2024-08-10
Fix/sqlite func exceptions

ThinkChaos:
## Description

A bit niche but I tried setting my bareasc prefix to an empty string,
and was getting an obtuse error. This should help make clearer what is
happening when queries fail.

The exception is not properly raised up the stack in the first place
because it happens across 2 FFI boundaries: the DB query
(Python -> SQLite), and the custom DB function (SQLite -> Python).
Thus Python cannot forwarded it back to itself through SQLite, and it's
treated as an "unraisable" exception.

We could override `sys.unraisablehook` to not print anything for the
original exception, and store it in a global for the outer Python
interpreter to fetch and raise properly, but that's pretty hacky,
limited to a single DB instance and query at once, and risks swallowing
other "unraisable" exceptions.
Instead we just tell the user to look above for what Python prints.

Sample output:
```
Exception ignored in: <function unidecode_expect_ascii at 0x7f7fa20bb060>
  File "site-packages/unidecode/__init__.py", line 60, in unidecode_expect_ascii
    bytestring = string.encode('ASCII')
                 ^^^^^^^^^^^^^
AttributeError: 'bytes' object has no attribute 'encode'
  File "site-packages/beets/dbcore/db.py", line 988, in query
sqlite3.OperationalError: user-defined function raised exception

  File "site-packages/beets/__main__.py", line 9, in <module>
    sys.exit(main())
             ^^^^^^
  File "site-packages/beets/ui/__init__.py", line 1865, in main
    _raw_main(args)
  File "site-packages/beets/ui/__init__.py", line 1852, in _raw_main
  File "site-packages/beets/ui/commands.py", line 1599, in list_func
  File "site-packages/beets/ui/commands.py", line 1594, in list_items
    for item in lib.items(query):
                ^^^^^^^^^^^^^^^^
  File "site-packages/beets/library.py", line 1695, in items
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "site-packages/beets/library.py", line 1673, in _fetch
    return super()._fetch(model_cls, query, sort)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "site-packages/beets/dbcore/db.py", line 1301, in _fetch
    rows = tx.query(sql, subvals)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "site-packages/beets/dbcore/db.py", line 991, in query
    raise DBCustomFunctionError()
beets.dbcore.db.DBCustomFunctionError: beets defined custom SQLite function failed; see the other errors above for details
```

## To Do

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,8 +2,7 @@ name: Test
 on:
   pull_request:
   push:
-    branches:
-      - master
+
 env:
   PY_COLORS: 1
 
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -20,11 +20,12 @@ import contextlib
 import os
 import re
 import sqlite3
+import sys
 import threading
 import time
 from abc import ABC
 from collections import defaultdict
-from sqlite3 import Connection
+from sqlite3 import Connection, sqlite_version_info
 from types import TracebackType
 from typing import (
     Any,
@@ -1143,9 +1144,14 @@ class Database:
 
             return bytestring
 
-        conn.create_function("regexp", 2, regexp)
-        conn.create_function("unidecode", 1, unidecode)
-        conn.create_function("bytelower", 1, bytelower)
+        deterministic = {}
+        if sys.version_info >= (3, 8) and sqlite_version_info >= (3, 8, 3):
+            # Let sqlite make extra optimizations
+            deterministic["deterministic"] = True
+
+        conn.create_function("regexp", 2, regexp, **deterministic)
+        conn.create_function("unidecode", 1, unidecode, **deterministic)
+        conn.create_function("bytelower", 1, bytelower, **deterministic)
 
     def _close(self):
         """Close the all connections to the underlying SQLite database
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -73,6 +73,16 @@ class DBAccessError(Exception):
     """
 
 
+class DBCustomFunctionError(Exception):
+    """A sqlite function registered by beets failed."""
+
+    def __init__(self):
+        super().__init__(
+            "beets defined SQLite function failed; "
+            "see the other errors above for details"
+        )
+
+
 class FormattedMapping(Mapping[str, str]):
     """A `dict`-like formatted view of a model.
 
@@ -970,6 +980,12 @@ class Transaction:
             self._mutated = False
             self.db._db_lock.release()
 
+        if (
+            isinstance(exc_value, sqlite3.OperationalError)
+            and exc_value.args[0] == "user-defined function raised exception"
+        ):
+            raise DBCustomFunctionError()
+
     def query(self, statement: str, subvals: Sequence = ()) -> List:
         """Execute an SQL statement with substitution values and return
         a list of rows from the database.
@@ -1027,6 +1043,10 @@ class Database:
             raise RuntimeError(
                 "sqlite3 must be compiled with multi-threading support"
             )
+
+        # Print tracebacks for exceptions in user defined functions
+        # See also `self.add_functions` and `DBCustomFunctionError`.
+        sqlite3.enable_callback_tracebacks(True)
 
         self.path = path
         self.timeout = timeout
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -2,8 +2,7 @@ name: Test
 on:
   pull_request:
   push:
-    branches:
-      - master
```
snejus, line 6: Were these two changes in github workflows intended?

beets/dbcore/db.py
```diff
@@ -1028,6 +1045,10 @@ def __init__(self, path, timeout: float = 5.0):
                 "sqlite3 must be compiled with multi-threading support"
             )
 
+        # Print tracebacks for exceptions in user defined functions
+        # See also `self.add_functions` and `DBCustomFunctionError`.
+        sqlite3.enable_callback_tracebacks(True)
```
snejus, line 1050: Do you reckon we should only do this if the logging level is `DEBUG`?

beets/dbcore/db.py
```diff
@@ -1123,9 +1144,14 @@ def bytelower(bytestring: Optional[AnyStr]) -> Optional[AnyStr]:
 
             return bytestring
 
-        conn.create_function("regexp", 2, regexp)
-        conn.create_function("unidecode", 1, unidecode)
-        conn.create_function("bytelower", 1, bytelower)
+        deterministic = {}
+        if sys.version_info >= (3, 8) and sqlite_version_info >= (3, 8, 3):
+            # Let sqlite make extra optimizations
+            deterministic["deterministic"] = True
```
snejus, line 1150: Consider using `partial` here, it may make it just slightly cleaner

```python
create_function = conn.create_function
...
    create_function = partial(create_function, deterministic=True)

create_function("regexp", 2, regexp)
...
```

beets/dbcore/db.py
ThinkChaos: I'm not sure, I think that would hurt discoverability/UX.

I didn't find anything about performance online, or a reason for why it's disabled by default.
Most plausible is the reasoning in https://github.com/simonw/datasette/issues/891: not pollute stderr by default.
So I think it'd be helpful to keep it on unless we have a reason not to.

Also, based on that above issue, I added feature detection to support PyPy.

.github/workflows/ci.yaml
ThinkChaos: Yes, I want to run all the tests *before* opening the PR :)
I removed that from this PR.

beets/dbcore/db.py
ThinkChaos: Done with https://github.com/beetbox/beets/pull/5383/commits/e4c42cdf8249822e8b9c127bf5c1005f4ccbe2af

ThinkChaos:
Sorry it took me a while to get back after the review. Forgot about it and just got to cleaning up enough of my notifications to see it again.

I retested it to make sure it still works, and found the motivation to add an automated test.

snejus:
Nice, thanky you! What would you think about fixing the original issue as well?

test/test_dbcore.py
```diff
@@ -776,3 +777,24 @@ def test_no_results(self):
             self.db._fetch(ModelFixture1, dbcore.query.FalseQuery()).get()
             is None
         )
+
+
+class ExceptionTest(unittest.TestCase):
```
snejus: Since we're migrating away from `unittest`, we've been writing all new tests using pytest. I reckon this one should be trivial to adapt?

ThinkChaos: Done in https://github.com/beetbox/beets/pull/5383/commits/8283b4f78509c0d2464a40d327bad6121de5a5f6

ThinkChaos:
> What would you think about fixing the original issue as well?

I can't reproduce the error anymore with empty bareasc prefix... :sweat_smile:

snejus:
Just rebase from master and we should be good to go!

ThinkChaos:
I rebased it.

In case you haven't seen it before, a Musicbrainz test failed on first try due to network flakiness: https://github.com/beetbox/beets/actions/runs/16862383319/job/47764495282?pr=5383
Retrying worked.

snejus (approved):

ThinkChaos:
Thank you!

PR #5385, opened 2024-08-11
Fixup: use fast queries for fields shared between albums and items tables

snejus:
I surprisingly found that many queries for fields that exist in the database are 'slow'. Querying items with `month:0` used a slow query while `title:0` used a fast one.

I found the mistake in the logic: the field existence in the db check was done *after* prepending the table name to the field. This meant that the logic wrongly checked e.g. `items.field` instead of `field`.

This commit fixes the above and renames the variables for more clarity.

### DIFF
--- a/beets/dbcore/queryparse.py
+++ b/beets/dbcore/queryparse.py
@@ -152,14 +152,15 @@ def construct_query_part(
     # Field queries get constructed according to the name of the field
     # they are querying.
     else:
-        key = key.lower()
-        if key in model_cls.shared_db_fields:
+        field = table = key.lower()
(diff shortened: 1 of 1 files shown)


Serene-Arc (approved):
Nice catch! These changes look good to me.

Serene-Arc:
Merge when ready, unless you want another review @snejus

Serene-Arc:
btw again, great catch. It took me like 5 different readings to get what changed in that code lol

snejus:
> btw again, great catch. It took me like 5 different readings to get what changed in that code lol

Sounds like the code should be improved regarding its readability 😆

PR #5388, opened 2024-08-16
fix(import): don't throw away album flexible fields

### LINKED ISSUE #4785: Dirty flex attr data_source not actually written to album table
Sorry for being brief, I'll elaborate later. I didn't add a verbose output for now since I think it is almost impossible to follow. If someone does want to read one, I'm happy to add one in a second post.

While working on fixing this issue which I though relates to an old beets version that I was using in the past was just missing a feature (writing `data_source` to album table): https://github.com/beetbox/beets/discussions/4784 I realize I _do_ have found a bug in a current version of beets that exists (I'm not 100% sure yet) **only** on my macOS 10.15.7 installation of beets (and not on my Linux machine). Maybe rather something like an sqlite bug? Just vagues ideas yet....

Using the master branch enriched with a couple of debug statements I realise that

- On an initial import the `data_source` field is marked dirty correctly
- It seems the database writes it
- I get back a transaction id from the database
- but in the end the field is **not** written to the database
- When I reimport that same album, the write is done properly

```
$ beet import /Users/jojo/Music/Techno/079-spin-23-04-remotebeets23/Aphrodite\ -\ Bass\ Lights\ \(Addict\ Mix\)\ _\ Solaris

/Users/jojo/Music/Techno/079-spin-23-04-remotebeets23/Aphrodite - Bass Lights (Addict Mix) _ Solaris (1 items)
Tagging:
    Aphrodite - Bass Lights (Addict Mix) / Solaris
URL:
    https://www.discogs.com/release/3161097-Aphrodite-Bass-Lights-Addict-Mix-Solaris
(Similarity: 73.0%) (id) (Discogs, Vinyl, 1994, UK, Aphrodite Recordings, APH 14)
 * Bass Lights (Addict Mix)
Missing tracks (1/2 - 50.0%):
 ! Solaris (# 2)
[A]pply, More candidates, Skip, Use as-is, as Tracks, Group albums,
Enter search, enter Id, aBort, eDit, edit Candidates, plaY?

In model.store

In model.store
Key is dirty: track_alt
Value is: A
Transaction executing:
278188

flex attr data_source found in loop, value is:
Discogs
Key is dirty: data_source
Value is: Discogs
Transaction executing:
278189

Right after tx.mutate statement. The flex attr was data_source. The first time setting an attr named like this must have been the album object.

data_source was found on first item of album!

data_source set on album:
Discogs

This is replaced album:
None


In model.store
flex attr data_source found in loop, value is:
Discogs

In model.store
flex attr data_source found in loop, value is:
Discogs

In model.store

In model.store
Key is dirty: art_source
Value is: google
Transaction executing:
8072


In model.store
flex attr data_source found in loop, value is:
Discogs

In model.store

In model.store
flex attr data_source found in loop, value is:
Discogs
```

Album 1957 was properly added

Screenshot 2023-05-12 at 14 27 11

art_source flex attr was written correctly but data_source is missing:

Screenshot 2023-05-12 at 14 28 59

It seems I have this bug on my mac for a long time already (several iterations of beets versions, always installed from git repo) but never realized that my albums are missing data_source (and it's only set on the tracks).

ThinkChaos:
## Description

I believe this fixes #4785, but I might be misunderstanding the description and it might already be fixed.

Tagging @JOJ0 since they created both the bug linked above, and 5bf4e3d92f5b4010a19cd3e8222fbf28b7b844e0 which mentions it.

## To Do

- [ ] ~Documentation. (If you've added a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)~

### DIFF
--- a/beets/autotag/__init__.py
+++ b/beets/autotag/__init__.py
@@ -14,10 +14,10 @@
 
 """Facilities for automatically determining files' correct metadata.
 """
-from typing import Mapping
+from typing import Mapping, Sequence, Union
 
 from beets import config, logging
-from beets.library import Item
+from beets.library import Album, Item
 
 # Parts of external interface.
 from .hooks import (  # noqa
@@ -80,6 +80,29 @@ SPECIAL_FIELDS = {
 # Additional utilities for the main interface.
 
 
+def _apply_metadata(
+    info: Union[AlbumInfo, TrackInfo],
+    db_obj: Union[Album, Item],
+    nullable_fields: Sequence[str] = [],
+):
+    """Set the db_obj's metadata to match the info."""
+    special_fields = SPECIAL_FIELDS[
+        "album" if isinstance(info, AlbumInfo) else "track"
+    ]
+
+    for field, value in info.items():
+        # We only overwrite fields that are not already hardcoded.
+        if field in special_fields:
+            continue
+
+        # Don't overwrite fields with empty values unless the
+        # field is explicitly allowed to be overwritten.
+        if value is None and field not in nullable_fields:
+            continue
+
+        db_obj[field] = value
+
+
 def apply_item_metadata(item: Item, track_info: TrackInfo):
     """Set an item's metadata from its matched TrackInfo object."""
     item.artist = track_info.artist
@@ -96,16 +119,15 @@ def apply_item_metadata(item: Item, track_info: TrackInfo):
     if track_info.artists_ids:
         item.mb_artistids = track_info.artists_ids
 
-    for field, value in track_info.items():
-        # We only overwrite fields that are not already hardcoded.
-        if field in SPECIAL_FIELDS["track"]:
-            continue
-        if value is None:
-            continue
-        item[field] = value
+    _apply_metadata(track_info, item)
 
     # At the moment, the other metadata is left intact (including album
     # and track number). Perhaps these should be emptied?
+
+
+def apply_album_metadata(album_info: AlbumInfo, album: Album):
+    """Set the album's metadata to match the AlbumInfo object."""
+    _apply_metadata(album_info, album)
 
 
 def apply_metadata(album_info: AlbumInfo, mapping: Mapping[Item, TrackInfo]):
@@ -218,21 +240,14 @@ def apply_metadata(album_info: AlbumInfo, mapping: Mapping[Item, TrackInfo]):
         # Track alt.
         item.track_alt = track_info.track_alt
 
-        # Don't overwrite fields with empty values unless the
-        # field is explicitly allowed to be overwritten
-        for field, value in album_info.items():
-            if field in SPECIAL_FIELDS["album"]:
-                continue
-            clobber = field in config["overwrite_null"]["album"].as_str_seq()
-            if value is None and not clobber:
-                continue
-            item[field] = value
+        _apply_metadata(
+            album_info,
+            item,
+            nullable_fields=config["overwrite_null"]["album"].as_str_seq(),
+        )
 
-        for field, value in track_info.items():
-            if field in SPECIAL_FIELDS["track"]:
-                continue
-            clobber = field in config["overwrite_null"]["track"].as_str_seq()
-            value = getattr(track_info, field)
-            if value is None and not clobber:
-                continue
-            item[field] = value
+        _apply_metadata(
+            track_info,
+            item,
+            nullable_fields=config["overwrite_null"]["track"].as_str_seq(),
+        )
--- a/beets/autotag/mb.py
+++ b/beets/autotag/mb.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import re
 import traceback
 from collections import Counter
+from itertools import product
 from typing import Any, Dict, Iterator, List, Optional, Sequence, Tuple, cast
 from urllib.parse import urljoin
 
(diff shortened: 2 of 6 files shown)


beets/autotag/mb.py
ThinkChaos, line 1: The changes to this file don't change functionality, it's just a refactor (see the commit split).

test/test_importer.py
```diff
@@ -1672,6 +1672,12 @@ def test_reimported_item_preserves_art(self):
         if new_artpath != old_artpath:
             self.assertNotExists(old_artpath)
 
+    def test_reimported_album_has_new_flexattr(self):
+        self._setup_session()
+        self.assertIsNone(self._album().get("bandcamp_album_id"))
+        self.importer.run()
+        self.assertEqual(self._album().bandcamp_album_id, "bc_url")
```
ThinkChaos, line 1679: I confirmed this fails without my change :)

beets/autotag/mb.py
```diff
@@ -618,65 +618,49 @@ def album_info(release: Dict) -> beets.autotag.hooks.AlbumInfo:
...
-        if config["musicbrainz"]["external_ids"]["deezer"].get():
-            fetch_deezer = True
-        if config["musicbrainz"]["external_ids"]["beatport"].get():
-            fetch_beatport = True
-        if config["musicbrainz"]["external_ids"]["tidal"].get():
-            fetch_tidal = True
+    external_ids = config["musicbrainz"]["external_ids"].get()
+    wanted_ids = {site for site, wanted in external_ids.items() if wanted}
+    if "url-relation-list" in release and wanted_ids:
+        urls = {}
```
snejus, line 624: You can simplify this entire block as

```python
from itertools import product

...

    external_ids = config["musicbrainz"]["external_ids"].get()
    wanted_sources = {site for site, wanted in external_ids.items() if wanted}
    if wanted_sources and (urls := release.get("url-relation-list", [])):
        urls = {}

        for source, url in product(wanted_sources, urls):
            if f"{source}.com" in (target := url["target"]):
                urls[source] = target
                log.debug(
                    "Found link to {} release via MusicBrainz",
                    source.capitalize(),
                )
```

See the diff below

image

beets/importer.py
```diff
@@ -815,9 +815,16 @@ def add(self, lib):
         with lib.transaction():
             self.record_replaced(lib)
             self.remove_replaced(lib)
+
             self.album = lib.add_album(self.imported_items())
-            if "data_source" in self.imported_items()[0]:
-                self.album.data_source = self.imported_items()[0].data_source
+            if self.choice_flag == action.APPLY:
+                # Copy album flexible fields to the DB
+                # TODO: change the flow so we create the `Album` object earlier,
```
snejus, line 822: Nice, I like this TODO

docs/changelog.rst
```diff
@@ -18,6 +18,10 @@ Bug fixes:
 * Improved naming of temporary files by separating the random part with the file extension.
 * Fixed the ``auto`` value for the :ref:`reflink` config option.
 * Fixed lyrics plugin only getting part of the lyrics from ``Genius.com`` :bug:`4815`
+* Album flexible fields are now correctly saved. For instance MusicBrainz external links
+  such as `bandcamp_album_id` will be available on albums in addition to tracks.
+  Fields will need to be re-added for the data to be saved, meaning it will require a re-import.
```
snejus, line 23: How much effort would it take to save them properly as part of the initial import?

test/test_importer.py
snejus: Thanks for adding the test!

I'd suggest rebasing this PR off `master` - you will find that all unittest assertions like `assertIsNone` and `assertEqual` have been replaced by `assert` statements.

beets/autotag/__init__.py
```diff
@@ -80,6 +80,29 @@
...
+    special_fields = SPECIAL_FIELDS[
+        "album" if isinstance(info, AlbumInfo) else "track"
+    ]
+
+    for field, value in info.items():
+        # We only overwrite fields that are not already hardcoded.
+        if field in special_fields:
+            continue
+
+        # Don't overwrite fields with empty values unless the
```
snejus, line 98: 👍🏼

snejus (changes requested):
Very clean PR, thank you! Added a suggestion, and requesting changes since the PR needs rebasing :)

docs/changelog.rst
ThinkChaos: Future initial imports will work, this is applies to pre-existing albums. I reworded the changelog entry, let me know if it's clearer.

I also realized that `mbsync` isn't enough, so removed that bit.

beets/autotag/mb.py
ThinkChaos: I made a small tweak before applying it because 2 vars were now called `urls`.

It seems no test actually exercises this path :slightly_frowning_face:  
I did do a real import + checked the result though, so it should be working.

.github/workflows/ci.yaml
```diff
@@ -2,8 +2,7 @@ name: Test
 on:
   pull_request:
   push:
-    branches:
-      - master
+
```
snejus, line 5: Forgot to ask about this in my previous review: why was this removed?

snejus (commented):
Just a small comment regarding the change in the workflows

docs/changelog.rst
snejus: Thanks for rewording - it is indeed more clear now 😃

.github/workflows/ci.yaml
ThinkChaos: To run the CI on all pushes. I can revert it if you'd like, but it's useful if you're working in a branch to have the CI check everything even before making a MR.

snejus: It's probably best to keep this limited to master, otherwise pushes in all forks will start triggering these builds. The CI build submits coverage and requires credentials setup, so it will start failing in forks, presumably.

Personally, I often push updates in my own beets fork and I'm kind of glad that I'm not receiving notification about failed linting builds and such.

snejus: Given your use case, a potential solution could be to open a draft PR while you're working on it?

ThinkChaos: I dropped the commit from the PR.

I really don't mind keeping a commit just for my fork and removing it before making a PR, but a couple things don't match my experience:
- there's an extra step before the CI runs in a fork, you have to manually go into the actions tab and click "yes I understand my workflows"
- Codecov works out of the (beet)box: https://app.codecov.io/gh/beetbox/beets/tree/ThinkChaos%2Fbeets%3Afix%2Fsave-album-flexible-fields/beets
- I don't get any notifications for failed CI, so there's likely a user preference somewhere

snejus (approved):
Thanks for the adjustments, you're good to go!

.github/workflows/ci.yaml
ThinkChaos: Good point about the non PR branches, they don't show up in the UI AFAICT, and they also don't fail

ThinkChaos:
Thanks for the review!

JOJ0:
@ThinkChaos thank you very much for this PR! I love the refactoring you did to the *_id fetching code! So much cleaner and shorter now! Also many thanks to @snejus for taking over with a review, much appreciated! ❤️

Issue #5185, opened 2024-04-10
Crash in autobpm plugin / Docs incomplete


JOJ0:
Sorry for being brief. A quick report on a crash I'm experiencing with the rather new autobpm plugin. Something with librosa. I'm on Python 3.11.
### Problem

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv autbpm label:syncopathic
```

Led to this problem:

```
Traceback (most recent call last):
  File "/home/jojo/.pyenv/versions/beets/bin/beet", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/jojo/git/beets/beets/ui/__init__.py", line 1865, in main
    _raw_main(args)
  File "/home/jojo/git/beets/beets/ui/__init__.py", line 1852, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "/home/jojo/git/beets/beetsplug/autobpm.py", line 46, in command
    self.calculate_bpm(lib.items(ui.decargs(args)), write=ui.should_write())
  File "/home/jojo/git/beets/beetsplug/autobpm.py", line 81, in calculate_bpm
    tempo, _ = beat.beat_track(y=y, sr=sr)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jojo/.pyenv/versions/3.11.3/envs/beets/lib/python3.11/site-packages/librosa/beat.py", line 185, in beat_track
    beats = __beat_tracker(onset_envelope, bpm, float(sr) / hop_length, tightness, trim)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jojo/.pyenv/versions/3.11.3/envs/beets/lib/python3.11/site-packages/librosa/beat.py", line 433, in __beat_tracker
    beats = __trim_beats(localscore, beats, trim)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jojo/.pyenv/versions/3.11.3/envs/beets/lib/python3.11/site-packages/librosa/beat.py", line 507, in __trim_beats
    smooth_boe = scipy.signal.convolve(localscore[beats], scipy.signal.hann(5), "same")
                                                          ^^^^^^^^^^^^^^^^^
AttributeError: module 'scipy.signal' has no attribute 'hann'
```

[...]

PR #5389, opened 2024-08-16
Fix autobpm and improve its docs

### LINKED ISSUE #5289: Error in autobpm module (at least with Python 3.7)
### Problem

Trying to import an album with the new autobpm plugin, first I had to install `librosa` and `resampy` to get it to run, those dependencies should probably be documented.

Then it failed with:

```
  File "/home/user/bin/beet", line 8, in <module>
    sys.exit(main())
  File "/home/user/.local/pipx/venvs/beets2/lib/python3.7/site-packages/beets/ui/__init__.py", line 1865, in main
    _raw_main(args)
  File "/home/user/.local/pipx/venvs/beets2/lib/python3.7/site-packages/beets/ui/__init__.py", line 1852, in _raw_main
  File "/home/user/.local/pipx/venvs/beets2/lib/python3.7/site-packages/beets/ui/commands.py", line 1395, in import_func
  File "/home/user/.local/pipx/venvs/beets2/lib/python3.7/site-packages/beets/ui/commands.py", line 1326, in import_files
    session.run()
  File "/home/user/.local/pipx/venvs/beets2/lib/python3.7/site-packages/beets/importer.py", line 360, in run
    pl.run_parallel(QUEUE_SIZE)
  File "/home/user/.local/pipx/venvs/beets2/lib/python3.7/site-packages/beets/util/pipeline.py", line 447, in run_parallel
  File "/home/user/.local/pipx/venvs/beets2/lib/python3.7/site-packages/beets/util/pipeline.py", line 312, in run
    out = self.coro.send(msg)
  File "/home/user/.local/pipx/venvs/beets2/lib/python3.7/site-packages/beets/util/pipeline.py", line 195, in coro
    func(*(args + (task,)))
  File "/home/user/.local/pipx/venvs/beets2/lib/python3.7/site-packages/beets/importer.py", line 1661, in plugin_stage
    func(session, task)
  File "/home/user/.local/pipx/venvs/beets2/lib/python3.7/site-packages/beets/plugins.py", line 143, in wrapper
    return func(*args, **kwargs)
  File "/home/user/.local/pipx/venvs/beets2/lib/python3.7/site-packages/beetsplug/autobpm.py", line 49, in imported
    self.calculate_bpm(task.imported_items())
  File "/home/user/.local/pipx/venvs/beets2/lib/python3.7/site-packages/beetsplug/autobpm.py", line 82, in calculate_bpm
    bpm = round(tempo)
TypeError: type numpy.ndarray doesn't define __round__ method
```

Adding an `import numpy` and changing the indicated line to `bpm = numpy.round(tempo)` fixed it here. Might just be a Python 3.7 problem, but the 2.0.0 release notes still claim Python 3.7 support.

### Setup

* OS:  Debian 10.13
* Python version: 3.7.3
* beets version: 2.0.0
* Turning off plugins made problem go away (yes/no): Fixing the plugin made the problem go away

snejus:
## Description

Fixes #5289 and #5185

Tried using the `autobpm` plugin and found a couple of issues:

1. #5185 `librosa` dependency was missing in `pyproject.toml`
2. Simply including the plugin in the configuration made `beet` take over 4 seconds to start up.
3. BPM detection failed due to another missing dependency, `resampy`
4. #5289 Librosa `beat_track` function returned unexpected type which made the plugin and the entire import process fail.

Addressed each of the above, slightly refactored the plugin and added tests.

## To Do


beets/library.py
```diff
@@ -658,6 +659,11 @@ def relation_join(cls) -> str:
             f"ON {cls._table}.album_id = {cls._relation._table}.id"
         )
 
+    @property
+    def filepath(self) -> Path | None:
+        """The path to the item's file as pathlib.Path."""
+        return Path(os.fsdecode(self.path)) if self.path else self.path
+
```
bal-e, line 666: Yesss pathlib!!!

docs/changelog.rst
```diff
@@ -12,12 +12,19 @@ New features:
   `beet list -a title:something` or `beet list artpath:cover`. Consequently
   album queries involving `path` field have been sped up, like `beet list -a
   path:/path/`.
+* :doc:`plugins/autobpm`: Addea new configuration option
```
bal-e, line 15: Typo

beetsplug/autobpm.py
```diff
@@ -11,81 +11,71 @@
...
+                self._log.error("Failed to load {}: {}", path, exc)
                 continue
-            except ValueError as exc:
-                self._log.error(
-                    "ValueError: failed to load {0} {1}",
-                    util.displayable_path(item.path),
-                    exc,
-                )
+
+            kwargs = self.config["beat_track_kwargs"].flatten()
```
bal-e, line 69: Just checking, there's no security considerations here, right?  cc @Serene-Arc

beetsplug/autobpm.py
```diff
@@ -11,81 +11,71 @@
...
+    def commands(self) -> list[Subcommand]:
+        cmd = Subcommand(
             "autobpm", help="detect and add bpm from audio using Librosa"
         )
         cmd.func = self.command
         return [cmd]
 
-    def command(self, lib, opts, args):
-        self.calculate_bpm(lib.items(ui.decargs(args)), write=ui.should_write())
+    def command(self, lib: Library, _, args: list[str]) -> None:
```
bal-e, line 49: I agree that the unused arguments should be marked, but I think their names should be preserved, i.e. using `_opts` instead of just `_`.  It's just a bit clearer, especially since the whole codebase isn't well-typed yet.  Also, a type annotation should still be used for the argument.

bal-e (changes requested):
I haven't used this feature myself, but this looks really good, @snejus!  I'm particularly happy about the use of `pytest` and `pathlib` :)

beets/library.py
snejus: Ultimately we will want to adjust `PathType` here I think (right, @Serene-Arc?)

```python
    def from_sql(self, sql_value) -> Path | None:
        value = self.normalize(sql_value)
        return Path(os.fsdecode(value)) if value else value
```

but it's too early yet since the entire codebase is expecting this to return `bytes` 😅 .

Thus this `filepath` property kind of allows us to migrate to `filepath.Path` step-by-step 😃

beetsplug/autobpm.py
snejus: I agree regarding general clarity, but do you reckon this is needed here given that this argument is unused? I find that `_` in such case clearly communicates that. I find the presence of the type a bit confusing if the argument is not used.

Note that my opinion here is mostly based on refactoring interfaces like the one that `BeetsPlugin` provides: in this case using `_` notation allows the interface (and the type) to change without the need to update the code here.

Essentially, I'm applying the [ISP principle](https://en.wikipedia.org/wiki/Interface_segregation_principle) to the function arguments here

> Within [object-oriented design](https://en.wikipedia.org/wiki/Object-oriented_design), [interfaces](https://en.wikipedia.org/wiki/Interface_(object-oriented_programming)) provide layers of abstraction that simplify code and create a barrier preventing coupling to [dependencies](https://en.wikipedia.org/wiki/Dependency_inversion_principle). A system may become so coupled at multiple levels that it is no longer possible to make a change in one place without necessitating many additional changes.[[1]](https://en.wikipedia.org/wiki/Interface_segregation_principle#cite_note-ASD-1) Using an interface or an [abstract class](https://en.wikipedia.org/wiki/Abstract_class) can prevent this side effect.

bal-e: That makes sense, I'm just not used to it.  Any preference, @Serene-Arc?

snejus:
@JOJ0 I noticed your comment under #4923 regarding difficulties calculating BPM for offbeaty tracks: I've been having the same issue - I found that adjusting `start_bpm` parameter in the `beat_track` call helps to control the 'taggable' BPM range. You can do so using the new configuration option:

```yaml
autobpm:
  beat_track_kwargs:
    start_bpm: 160
```

Just giving a shout making sure you're aware!

snejus:
@Serene-Arc @bal-e do we think this could be merged?

bal-e (approved):

tandy-1000:
thanks for doing this!

JOJ0:
Also from my end many thanks @snejus and thanks for the tip with configuring bpm min value. will try that!

PR #5393, opened 2024-08-25
Improved NamedTuple declaration

amogus07:
## Description

Utilize a [new way of declaring NamedTuples](https://docs.python.org/3/library/typing.html#typing.NamedTuple), which allows for typechecking as well. Maybe the latter is now redundant in other places, but I'm not that familiar with the codebase yet, so I just changed the declarations (and hopefully used the correct types). While I was at it, I also ran `poetry update`, but I'll revert poetry.lock in case I wasn't supposed to do that. This is my first commit here, so I hope I didn't do anything wrong...

## To Do

- [x] ~Documentation.~
- [x] ~Changelog.~
- [ ] Tests.

### DIFF
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -17,6 +17,7 @@ releases and tracks.
 """
 
 
+from __future__ import annotations
 import datetime
 import re
 from typing import (
@@ -78,7 +79,7 @@ class Recommendation(OrderedEnum):
 
 
 class Proposal(NamedTuple):
-    candidates: Sequence[Union[AlbumMatch, TrackMatch]]
+    candidates: Sequence[AlbumMatch | TrackMatch]
     recommendation: Recommendation
 
 
@@ -354,7 +355,7 @@ def match_by_id(items: Iterable[Item]):
 
 
 def _recommendation(
-    results: Sequence[Union[AlbumMatch, TrackMatch]],
+    results: Sequence[AlbumMatch | TrackMatch],
 ) -> Recommendation:
     """Given a sorted list of AlbumMatch or TrackMatch objects, return a
     recommendation based on the results' distances.
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -78,7 +78,7 @@ class Recommendation(OrderedEnum):
 
 
 class Proposal(NamedTuple):
-    candidates: List[Any]
+    candidates: Sequence[AlbumMatch | TrackMatch]
     recommendation: Recommendation
 
 
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -19,12 +19,12 @@ releases and tracks.
 
 import datetime
 import re
-from collections import namedtuple
 from typing import (
     Any,
     Dict,
     Iterable,
     List,
+    NamedTuple,
     Optional,
     Sequence,
     Tuple,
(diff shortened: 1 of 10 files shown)


poetry.lock
```diff
@@ -74,13 +74,13 @@ test = ["tox"]
 
 [[package]]
 name = "babel"
-version = "2.15.0"
+version = "2.16.0"
```
snejus, line 77: Presumable this update was not intentional? It would have to be done in a separate PR on its own

snejus (changes requested):
I like this update!

Just checked and it seems like there's one replacements that's been left out?

image

amogus07:
uhhhh… can't reproduce this locally?!

amogus07:
@snejus What should I do now? Can you reproduce it?

snejus:
`dict` is only supported from Python 3.9 onwards, so try using `typing.Dict` instead. Also, I think you may want to use `str` type for the key instead of `Any`, `Dict[str, Any]`.

beets/dbcore/db.py
```diff
@@ -322,7 +322,7 @@ class Model(ABC):
     """
 
     @cached_classproperty
-    def _relation(cls) -> type[Model]:
+    def _relation(cls) -> Model:
```
snejus, line 325: Was this change intentional? This needs to be `type[Model]`

beets/plugins.py
```diff
@@ -291,7 +292,7 @@ def load_plugins(names=()):
             )
 
 
-_instances = {}
+_instances: Dict[Any, Any] = {}
```
snejus, line 295: I think it should be possible to specify accurate types for both the key and value here

beets/ui/commands.py
```diff
@@ -665,7 +664,7 @@ def show_match_tracks(self):
         """
         # Tracks.
         # match is an AlbumMatch named tuple, mapping is a dict
-        # Sort the pairs by the track_info index (at index 1 of the namedtuple)
+        # Sort the pairs by the track_info index (at index 1 of the NamedTuple)
```
snejus, line 667: If you're adjusting this, could you clarify what **at index 1 of the NamedTuple** means? Can this be replaced by the name of the namedtuple attribute?

beets/util/__init__.py
```diff
@@ -833,7 +838,7 @@ def plurality(objs: Sequence[T]) -> T:
     return c.most_common(1)[0]
 
 
-def convert_command_args(args: List[bytes]) -> List[str]:
+def convert_command_args(args: Sequence[Bytes_or_String]) -> List[str]:
```
snejus, line 841: The type fixes in this module are unfortunately out of scope of the `namedtuple` -> `NamedTuple` replacement. Please undo them and submit a separate PR as they need to be reviewed carefully.

snejus (changes requested):
Ideally we want to keep this PR focused on the namedtuple replacement, so I asked to submit the rest of the type changes as a separate PR.

beets/autotag/match.py
```diff
@@ -78,7 +78,7 @@ class Recommendation(OrderedEnum):
 
 
 class Proposal(NamedTuple):
-    candidates: Sequence[AlbumMatch | TrackMatch]
+    candidates: Sequence[Union[AlbumMatch, TrackMatch]]
```
snejus, line 81: Forgot to mention - there's no need for this adjustment as it already utilizes the new pipe syntax.

amogus07: it doesn‘t work below 3.10 though

snejus: Try including `from __future__ import annotations` import at the top of the module

beets/ui/commands.py
amogus07: I think the person who wrote this would have to explain what it means…

amogus07: @JOJ0 can you help out?

snejus (approved):
Thanks for making the requested adjustments, it all looks good now!

PR #5407, opened 2024-09-04
Update dependencies and do not install reflink on Windows for tests

snejus:
See my comment under #5406 for context

> The build on win32 is failing to install reflink because it's [only supported until Python 3.7](https://gitlab.com/rubdos/pyreflink/-/blob/master/setup.py?ref_type=heads).
>
> I will address this in a separate PR and rebase this one accordingly once the fix is merged.
>
> Note: this issue popped up now because I added a new requests-mock dependency which invalidated cached dependencies.

### DIFF
--- a/test/test_files.py
+++ b/test/test_files.py
@@ -86,12 +86,10 @@ class MoveTest(BeetsTestCase):
         self.i.move(operation=MoveOperation.COPY)
         self.assertExists(self.path)
 
-    @NEEDS_REFLINK
     def test_reflink_arrives(self):
         self.i.move(operation=MoveOperation.REFLINK_AUTO)
         self.assertExists(self.dest)
 
-    @NEEDS_REFLINK
     def test_reflink_does_not_depart(self):
         self.i.move(operation=MoveOperation.REFLINK_AUTO)
         self.assertExists(self.path)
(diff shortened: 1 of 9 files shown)


test/test_files.py
```diff
@@ -585,7 +582,6 @@ def test_unsuccessful_copy(self):
         with pytest.raises(util.FilesystemError):
             util.copy(self.path, self.otherpath)
 
-    @unittest.skipUnless(_common.HAVE_REFLINK, "need reflink")
```
bal-e, line 588: This should be `@NEEDS_REFLINK` too, right?

bal-e (changes requested):

test/test_files.py
snejus: I found that this test fails before reflink import is being attempted, see below:

```python
if samefile(path, dest):
    return

    raise FilesystemError("target exists", "rename", (path, dest))

if fallback:
    with suppress(Exception):
        return import_module("reflink").reflink(path, dest)
...
```

I have now included the `match` parameter in `pytest.raises` to clarify this!

bal-e: Ah, that makes sense.

bal-e (approved):

bal-e:
LGTM, I think?

Issue #5467, opened 2024-10-18
Detect Microsoft Store installation and display warning


jackwilsdon:
### Proposed solution

We should detect whether Python is installed via the Microsoft Store and issue a warning if so. The warning should say something along the following lines:

> Warning: your Python installation is from the Microsoft Store which restricts access to AppData. Please follow https://beets.readthedocs.io/en/stable/guides/main.html#installing-on-windows.

### Objective

Microsoft Store installations of Python seem to be sandboxed which causes confusion (see #5466).

#### Goals

Avoid confusion when using Microsoft Store installation.

snejus:
To start with, we could potentially document this somewhere in the **Windows** / **Installation** docs?

arogl:
with this 

```
$ python
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.exec_prefix
```

and detecting a path within this directory `C:\Users\username\AppData\Local\Microsoft\WindowsApps\` we should be able to stop beets execution and advise user about options. 

Will need some thoughts about config options and editing via `beet config -e`

PR #5470, opened 2024-10-19
Add a quick check for MST Store Python install

arogl:
## Description

Quick fix for #5467.

Checks if the path for python is under the windows store folder then error and point the user to the beets [documentation](https://beets.readthedocs.io/en/stable/guides/main.html).

Happy for feedback to improve, but thought it best to exit as early as possible.

## To Do

- [x] Documentation.
- [x] Changelog.
- [x] Tests.


beets/ui/__init__.py
```diff
@@ -1858,6 +1858,16 @@ def main(args=None):
...
     exception handlers that print friendly error messages.
     """
+    if "AppData\\Local\\Microsoft\\WindowsApps" in sys.exec_prefix:
+        log.error(
+            "error: beets is unable to use the Microsoft Store version of "
+            "Python. Please install Python from python.org."
+        )
+        log.error(
+            "error: More details can be found here "
+            "https://beets.readthedocs.io/en/stable/guides/main.html"
```
snejus, line 1868: Consider using a single `log.error` call:

```suggestion
            "error: beets is unable to use the Microsoft Store version of "
            "Python. Please install Python from python.org. See "
            "https://beets.readthedocs.io/en/stable/guides/main.html."
```

snejus (changes requested):

beets/ui/__init__.py
arogl: Updated

snejus:
Apologies for not addressing this earlier - all looks good, just needs a rebase on latest master.

snejus (approved):

arogl:
> Apologies for not addressing this earlier - all looks good, just needs a rebase on latest master.

I believe I have managed a rebase 😃

PR #5492, opened 2024-10-30
Fix docs build

snejus:
## Description

Seems like readthedocs build expects docs dependencies to be available
as an extra, see one of the failing builds: https://app.readthedocs.org/projects/beets/builds/26079213/.

You can see this has been failing due to missing 'pydata_sphinx_theme' which is one of our dependencies:

```
    raise ThemeError(__('no theme named %r found (missing theme.toml?)') % name)
    sphinx.errors.ThemeError: no theme named 'pydata_sphinx_theme' found (missing theme.toml?)
    Theme error:
    no theme named 'pydata_sphinx_theme' found (missing theme.toml?)
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,6 @@
 name: Test
 on:
-  pull_request_target:
+  pull_request:
   push:
     branches:
       - master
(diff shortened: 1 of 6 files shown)


snejus:
The two failing checks are from the workflow that used `pull_request_target` trigger. As you can see they are failing even when the code has been fixed, because they are checking the code on `master`.

Thus the last commit reverts the commit that introduced `pull_request_target` to fix coverage uploads from forks. We will find another way to fix that issue.

Serene-Arc (approved):
lgtm, merge when you're ready

PR #5503, opened 2024-11-13
Update python to current supported versions. Also add experimental support for next python release

Fale:
## Description

Bump python version in the CI to current versions.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,8 +14,17 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ubuntu-latest, windows-latest]
-        python-version: ["3.9"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        experimental: [false]
+        include:
+          - python-version: "3.13"
+            platform: ubuntu-latest
+            experimental: true
+          - python-version: "3.14-dev"
+            platform: ubuntu-latest
+            experimental: true
     runs-on: ${{ matrix.platform }}
+    continue-on-error: ${{ matrix.experimental }}
     env:
       IS_MAIN_PYTHON: ${{ matrix.python-version == '3.9' && matrix.platform == 'ubuntu-latest' }}
     steps:
@@ -33,7 +42,7 @@ jobs:
         if: matrix.platform == 'ubuntu-latest'
         run: |
           sudo apt update
-          sudo apt install ffmpeg gobject-introspection libgirepository1.0-dev
+          sudo apt install ffmpeg gobject-introspection libgirepository1.0-dev llvm libopenblas-dev
           poetry install --extras=replaygain --extras=reflink
 
       - name: Install Python dependencies
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,10 +14,10 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ubuntu-latest, windows-latest]
-        python-version: ["3.8", "3.9"]
+        python-version: ["3.9"]
     runs-on: ${{ matrix.platform }}
     env:
-      IS_MAIN_PYTHON: ${{ matrix.python-version == '3.8' && matrix.platform == 'ubuntu-latest' }}
+      IS_MAIN_PYTHON: ${{ matrix.python-version == '3.9' && matrix.platform == 'ubuntu-latest' }}
     steps:
       - uses: actions/checkout@v4
       - name: Install Python tools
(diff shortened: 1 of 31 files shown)


.github/workflows/ci.yaml
```diff
@@ -14,10 +14,16 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ubuntu-latest, windows-latest]
-        python-version: ["3.8", "3.9"]
+        python-version: ["3.11", "3.12"]
```
snejus, line 17: I don't think that we want to stop supporting Python 3.9 and Python 3.10 yet.

snejus (changes requested):

Fale:
I've added back 3.9 and 3.10. Also I've pushed 3.12 to be ok to fail, since now it is failing. The best scenario would be to have up to 3.13 working properly, considering that 3.13 is already in Fedora 41 (released last month) and very soon other distros will move to it

snejus:
I see that you're dropping support for Python 3.8 - note that this needs a bit more effort:

1. Looking through the release notes, checking the codebase against any removals/deprecations and updating the codebase accordingly.
2. Typing updates: we should use `pyupgrade` to align types used in the codebase with what Python 3.9 supports.
3. Update of Python constraints in the package.
4. Relevant updates in the documentation.
5. And finally the changelog note.

That's a couple of things just off the top of my head - there's likely a couple of other things. 

Are you happy to go ahead with the above or would you like to limit the scope of this PR to simply add support for the most recent version without removing Python 3.8?

Fale:
I've added back 3.8.

The reason why I removed it in the first place was that it is EOL, but I don't mind it being there. The part I'm interested into is the support to the newest versions. I don't have anything against the old versions :).

snejus:
Deps installation on Python 3.12+ seems to be failing due to `numpy 1.24.4` which does not support Python 3.12. It added support in `1.26.0`, however in the same release it also removed support for Python 3.8.

Poetry resolved this `numpy` version because we are still supporting Python 3.8. Just tried setting minimum python version to 3.9 in the package and resolved the dependencies - `numpy` indeed gets upgraded, which should resolve this issue.

I will shortly submit a PR that removes Python 3.8 support and fixes this issue for you.

Fale:
Thanks a lot :)

As soon as the other PR is merged, I'll rebase this one :)

snejus:
> Thanks a lot :)
> 
> As soon as the other PR is merged, I'll rebase this one :)

I think you may be able to adjust the base branch for this PR to point to `drop-py38` branch. Once that PR it's merged, the base will automatically reset to `master`.

Fale:
Thanks! I've rebased on the `drop-py38` branch. It now works properly on py3.12 but not yet with py3.13. I'm not sure if we want to move py3.13 to be ok to fail or you think it makes more sense to fix the compatibility to py3.13 directly

snejus:
> Thanks! I've rebased on the `drop-py38` branch. It now works properly on py3.12 but not yet with py3.13. I'm not sure if we want to move py3.13 to be ok to fail or you think it makes more sense to fix the compatibility to py3.13 directly

I guess we'd need to understand why is that issue occurring: if you want to try to fix it you will have to wrestle with Poetry a bit and find out why is that dependency installation failing.

As far as I remember, this may have to do with `scipy` - I think googling up that error may give some clues about what's going on.

Fale:
It seems like there is an issue with `numba`. We are importing `0.60.0` (the latest released one), while `0.61.0` will [have](https://github.com/numba/numba/blob/main/docs/source/release/0.61.0-notes.rst#python-3-13-support) Python 3.13 support. The first rc should be released in the coming days.

snejus:
@Fale sounds good. Feel free to review https://github.com/beetbox/beets/pull/5508 if you have time - if all good I will merge it in.

snejus:
Alternatively, you can also split this into two PRs:

1. Add support for 3.10 and 3.11, base it on the current `master`
2. Add support for 3.12, base it on `drop-py38`.

Fale:
I've moved the 3.13 to be ok to fail so that this can be merged. We can work on 3.13 support when `numba 0.61.0` gets released :).

Thanks a lot :)

snejus:
I bet that numba will get fixed before #5508 is reviewed and ready to merge 😁 I think that's the main bottleneck here.

snejus:
Oh I stupidly removed the `drop-py38` immediately after the merge and GitHub did not have time to change the base to `masster` and thus closed this PR since its base disappeared.

Apologies!!

PR #5505, opened 2024-11-15
Use up-to-date namespace package setup for plugins

snejus:
### Description

Refactor `beetsplug` to use native namespace packages by removing `__init__.py`. Update documentation and `setup.cfg` to support namespace packages.

### Changes

- Deleted `beetsplug/__init__.py`.
- Updated `docs/dev/plugins.rst` to reflect namespace package structure.
- Updated `mypy` configuration in `setup.cfg`

### Motivation

Adopt PEP 420 native namespace packages to simplify plugin management and eliminate the need for `__init__.py`.

See https://realpython.com/python-namespace-package.

This setup is backwards-compatible, so plugins using the old pkgutil-based setup will continue working fine.

The advantage with this setup is that external plugins will now be able to import modules from 'beetsplug' package for typing purposes. Previously, mypy could not resolve these modules due to presence of `__init__.py`.


docs/dev/plugins.rst
```diff
@@ -3,46 +3,55 @@
...
 
+.. code-block:: yaml
+
+    # config.yaml
+    plugins:
+      - myawesomeplugin
+
+and you're good to go!
+
+.. _Stack Overflow question about namespace packages: https://stackoverflow.com/a/27586272/9582674
```
JOJ0, line 54: Should we better link to the realpython article here?

docs/dev/plugins.rst
```diff
@@ -3,46 +3,55 @@
...
+            myawesomeplugin.py
+
+.. attention::
 
-    from pkgutil import extend_path
-    __path__ = extend_path(__path__, __name__)
+    You do not anymore need to add a ``__init__.py`` file to the ``beetsplug``
+    directory. Python treats your plugin as a namespace package automatically,
+    thus we do not depend on ``pkgutil``-based setup in the ``__init__.py``
+    file anymore.
```
JOJ0, line 26: If the general info about (current) namespace packages is already explained in the realpython article, maybe here you can point out this chapter of the python docs just to make sure we document everything potentially interesting for plugin developers (and also those who adopt old/orphaned plugins;-)): https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#legacy-namespace-packages

setup.cfg
```diff
@@ -37,3 +37,5 @@ allow_any_generics = false
 # FIXME: Would be better to actually type the libraries (if under our control),
 # or write our own stubs. For now, silence errors
 ignore_missing_imports = true
+namespace_packages = true
+explicit_package_bases = true
```
JOJ0, line 41: LGTM. https://mypy.readthedocs.io/en/stable/config_file.html

JOJ0 (changes requested):
Built the docs and read them. I like the "attention" box. Some suggestions within.

What I miss in general in our docs is for plugin developers if there is anything specific they need to do in there pyproject.toml or setup.py files.

I'm not sure but is it enough to do this in a setup.py?

```
packages=find_packages(),
```

Here is some information about it: https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#finding-simple-packages

Should we add the most important parts that need to be in the package config. Maybe only focusing on pyproject.toml and setup.py verisions (I suppose they are the most common) and maybe also link to the above? What do you think? Too much information? Less is more?

docs/dev/plugins.rst
snejus: Could you share the link? 😅

JOJ0: the one you posted in the pr description 😁 

https://realpython.com/python-namespace-package

snejus: 🤦🏼 🤦🏼 🤦🏼

snejus: Added the link to the realpython article as well!

docs/dev/plugins.rst
snejus: Removed most of this section and added links to poetry and setuptools

JOJ0:
> I prefer explicit (latter) over implicit. For example in beetcamp the following works fine:
> 
> ```toml
> packages = [{ include = "beetsplug" }]
> ```
> 
> I'm not sure we should include these instructions in the docs since we've got a variety of build tools and each has its own instructions. We haven't got anything special going on here so as long as they somehow include the `beetsplug` directory in their package things will work fine. Addings links to `setuptools` and `poetry` docs is a good idea though!

I agree. Less is more and better maintainable. With these links added I think there is more than enough detailed information for a plugin developer. I rendered the article and it reads very straight forward now. Thanks! Merge it!

PR #5508, opened 2024-11-17
Remove support for Python 3.8

snejus:
## Description

- Drop support for EOL Python 3.8 making Python 3.9 the minimum supported version

- Take advantage of Python 3.9+ type hint syntax by:
  - Using `list[T]` instead of `List[T]` etc. from typing module
  - Using `Type | None` syntax for unions instead of `Union[Type, None]`
  - Moving collection type hints from `typing` to `collections.abc`
  - Using `TYPE_CHECKING` guard for runtime import optimization

Note: in #5503 we found that we cannot support Python 3.12 unless we upgrade our minimum support Python to 3.9.

## To Do

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,7 +7,7 @@ on:
       - master
 
 env:
-  PYTHON_VERSION: 3.8
+  PYTHON_VERSION: 3.9
 
 jobs:
   changed-files:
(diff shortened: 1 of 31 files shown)


.github/workflows/lint.yml
```diff
@@ -131,7 +131,7 @@ jobs:
         run: echo "::add-matcher::.github/sphinx-problem-matcher.json"
 
       - name: Build docs
-        run: |
+        run: |-
```
Serene-Arc, line 134: What's this do?

Serene-Arc (approved):
Looks good to me!

.github/workflows/lint.yml
snejus: This was accidental actually. See 
image

PR #5516, opened 2024-11-25
Skip autobpm tests if librosa isn't available

stefanor:
## Description

Debian doesn't have `librosa`. Allow the tests to continue.

### DIFF
--- a/test/plugins/test_autobpm.py
+++ b/test/plugins/test_autobpm.py
@@ -1,6 +1,13 @@
+import importlib.util
+import os
+
 import pytest
 
 from beets.test.helper import ImportHelper, PluginMixin
+
+github_ci = os.environ.get("GITHUB_ACTIONS") == "true"
+if not github_ci and not importlib.util.find_spec("librosa"):
+    pytest.skip("librosa isn't available", allow_module_level=True)
 
 
 class TestAutoBPMPlugin(PluginMixin, ImportHelper):


snejus:
Does this fix the issue for you?

I'd be surprised if it does - `autobpm` is not expected to exist as a command. Rather, the tests run `beet autobpm`. If the dependencies are missing on your end, you can install them using the `autobpm` extra, see [here](https://beets.readthedocs.io/en/latest/plugins/autobpm.html).

test/plugins/test_autobpm.py
```diff
@@ -1,7 +1,12 @@
+import shutil
+
 import pytest
 
 from beets.test.helper import ImportHelper, PluginMixin
 
+if shutil.which("autobpm") is None:
+    pytest.skip("autobpm isn't available", allow_module_level=True)
```
snejus, line 8: I didn't know about `allow_module_level` before, that's cool!

stefanor:
Ah, I misunderstood `self.run_command()`. Of course it worked by always skipping the tests.

test/plugins/test_autobpm.py
```diff
@@ -1,7 +1,12 @@
+import importlib.util
+
 import pytest
 
 from beets.test.helper import ImportHelper, PluginMixin
 
+if not importlib.util.find_spec("librosa"):
```
snejus, line 7: Nice. I initially thought 'why not simple `import librosa` in a try catch block' but why import it if it's not required in this module.

Note that we _always_ want to run these tests in CI: after this change, if librosa installation was accidentally removed from the CI, these tests would be skipped silently,

See [test_lyrics.py](https://github.com/beetbox/beets/blob/37a2ceccd1a0ca7f6e256d6bf23f92ea0ef6a38f/test/plugins/test_lyrics.py#L261) for an example condition to check whether the tests are run in CI or not.

snejus (changes requested):

test/plugins/test_autobpm.py
stefanor: > Nice. I initially thought 'why not simple import librosa in a try catch block' but why import it if it's not required in this module.

I had that, and the CI told me not to do it that way :P

snejus: Ah, the good ol' F401

snejus (approved):
Nicely done, thanks!

Issue #1250, opened 2015-01-21
chroma tells me that audio could not be decoded


xsteadfastx:
i cant get the chroma plugin working on all my ubuntu unicorn boxes. i have the feeling that i didnt had the problem with the previous beets version. i checked all installed packages named in the doc.

* python-gst0.10-dev
* gstreamer0.10-plugins-good
* gstreamer0.10-plugins-bad
* gstreamer0.10-plugins-ugly

and i even installed `libchromaprint-tools` to have the `fpcalc` command. i still get the error:

```
fingerprinting of '/home/myuser/Musik/Kill Holiday/Somewhere Between the Wrong Is Right/01 Somewhere Between the Wrong Is Right.flac' failed: audio could not be decoded
```

if i use the fpcalc command on that file... it gives me the fingerprint. so i think that it can be decoded somehow. im pretty stuck with finding the error. anything i could try?

my config:
```
directory: ~/Musik/beets                                                          
paths:                                                                            
        singleton: Non-Album/$artist/$title                                       
plugins: chroma                                                        
```

brunal:
This error comes from `pyacoustid`, in `_fingerprint_audio_read()`: it catches a `DecodeError` from `audioread`, which happens because no backend could read the music. Please try to determine which backends are available: try the following
`import gst`
`import mad`
And do you have `ffmpeg` available on the commad-line?

brunal:
If using GST, you need 3 elements: uridecodebin, audioconvert and appsink. You may be missing one of those 3. Are there other plugin packages available on ubuntu? Is something like gstreamer0.10-plugins-base available?

xsteadfastx:
i "investigated" a little and i think the problem is that i try to run beets in a virtualenv. `import gst` doesnt work in the venv. now i installed everything system-wide and it works. mmhhh not the best way but it works now.

brunal:
Right! That's a recurring problem in the python ecosystem (cf. matplotlib, opencv...). You should be able to get it working in a virtual environment by creating/updating it with the `--system-site-packages` option.

xsteadfastx:
so... im sorry for generating just hot air with this bugreport  and thanks for all the informations and help :)

brunal:
No problem, the doc should probably mention that point (I'll update it). Enjoy beets!



—
Tagada tsouin tsouin

bolisaf:
I've the same problem but I'm not as python-savvy as xsteadfastx
I've installed beets using pip.
[...]

PR #5533, opened 2024-12-02
Add FFmpeg requirement info and links to installation guides

hardyoyo:
## Description

- Add documentation re: the requirement for FFmpeg, and links to installation guides for FFmpeg.
- Refs #1250

### DIFF
--- a/docs/guides/main.rst
+++ b/docs/guides/main.rst
@@ -9,7 +9,7 @@ collection better.
 Installing
 ----------
 
-You will need Python.
+You will need FFmpeg and Python.
 Beets works on Python 3.8 or later.
 
 * **macOS** 11 (Big Sur) includes Python 3.8 out of the box.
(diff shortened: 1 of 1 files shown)


snejus:
By default beets does not depend on ffmpeg. Plugins like `replaygain` and `convert` do, so maybe docs of these plugins would be more relevant for these installation instructions?

hardyoyo:
You're right, and the Chroma plugin *does* [mention needing FFmpeg](https://beets.readthedocs.io/en/v1.3.17/plugins/chroma.html) (among other things). I believe this PR would be better if it instead directed users to investigate the dependencies of any plugins they wish to use, instead of blindly instructing folks to install FFmpeg. I'll revise it.

hardyoyo:
actually, this seems like an unnecessary hint... I'm just going to close this PR

snejus:
> instead directed users to investigate the dependencies of any plugins they wish to use,

Some more clarity in our docs sounds like a good idea!

PR #5544, opened 2024-12-10
Update python in CI

Fale:
## Description
Bump python version in the CI to current versions.

This is the same as #5503 that for some reason got closed when the merge of the `drop-py38` branch happened

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ubuntu-latest, windows-latest]
-        python-version: ["3.9"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
     runs-on: ${{ matrix.platform }}
     env:
       IS_MAIN_PYTHON: ${{ matrix.python-version == '3.9' && matrix.platform == 'ubuntu-latest' }}


snejus:
I'm thinking maybe we should comment out 3.13 and 3.14 for now with a note that this depends on `numba`s support for 3.13.

snejus:
Unfortunately new version of `numba` dropped Python 3.9 altogether, which means we can only support versions up to and including 3.12.

Fale:
Why would dropping 3.9 be a problem?
Seems like:
* Ubuntu 22.04 and higher have Python 3.10 (and higher)
* Fedora: all supported version have Python 3.12 and higher
* Debian: from bookworm (current stable) onward they have Python 3.10 or higher

So, it seems dropping 3.9 should not be a problem today.

Would you agree?

snejus:
Historically, beets has never dropped support for a non-EOL Python version. If we do so, those people that use Python 3.9 will be forced to upgrade their setups early in order to upgrade beets, which is not ideal. People have various legitimate reasons for using the earliest support Python: for example, maintaining their own open-source packages.

I checked monthly beets download for each Python version:

image

Seems like there's still a considerable number of people who use it with Python 3.9 :)

Fale:
the assumption, though, is that all those users are on the latest version of beets/ Is that the case?

Also, how are those statistics fetched? I was not aware that beets has a telemetry that was able to provide such info, was I under the wrong assumption os are those from somewhere else?

snejus:
That's just download statistics from pypi, see [pypistats](https://pypi.org/project/pypistats/). 

I can see on [pepy.stats](https://pepy.tech/projects/beets?timeRange=threeMonths&category=version&includeCIDownloads=true&granularity=daily&viewType=chart&versions=2.2.0%2C2.1.0%2C2.0.0) that nearly all users download the most recent version indeed.

Fale:
I see, thanks!
Sooner or later beets will have to make the step. I think it will be important to understand when it will make sense for beets to do it.

snejus:
Most likely this is going to happen with EOL of Python 3.9. We should support 3.11 and 3.12 anyways though!

snejus:
@Fale Would you be happy to remove 3.13 and 3.14 for now so that we can merge this in? Otherwise I could take it from here myself

Fale:
Should be ok now :)

snejus (approved):

snejus:
Thank you!

PR #5545, opened 2024-12-10
Properly type db on the model, sort out generic types issues

snejus:
Thanks to @wisp3rwind's suggestion this PR adds types to the relationship between `Model`, `Database` and `Library`.

Then I worked through the rest of the issues found in the edited files. Most of this involved providing type parameters for generic types (or defining defaults, rather 😉).

There `queryparse` module had a somewhat significant issue where the sorting construction logic only expected to receive `FieldSort` subclasses, while `SmartArtistSort` was not one. Thus `SmartArtistSort` has now been forced to behave and is a `FieldSort` subclass. It's also been moved to `query.py` module which is where the rest of sorts are defined.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -38,6 +38,7 @@ from .query import (
     AndQuery,
     FieldQuery,
     FieldQueryType,
+    FieldSort,
     MatchQuery,
     NullSort,
     Query,
@@ -294,7 +295,7 @@ class Model(ABC, Generic[D]):
     """Optional Types for non-fixed (i.e., flexible and computed) fields.
     """
 
-    _sorts: dict[str, type[Sort]] = {}
+    _sorts: dict[str, type[FieldSort]] = {}
     """Optional named sort criteria. The keys are strings and the values
     are subclasses of `Sort`.
     """
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -48,6 +48,8 @@ from .query import (
 if TYPE_CHECKING:
     from types import TracebackType
 
+    from .query import SQLiteType
+
     D = TypeVar("D", bound="Database", default=Any)
 else:
     D = TypeVar("D", bound="Database")
@@ -579,7 +581,7 @@ class Model(ABC, Generic[D]):
 
         # Build assignments for query.
         assignments = []
-        subvars = []
+        subvars: list[SQLiteType] = []
         for key in fields:
             if key != "id" and key in self._dirty:
                 self._dirty.remove(key)
@@ -959,14 +961,14 @@ class Transaction:
             self._mutated = False
             self.db._db_lock.release()
 
-    def query(self, statement: str, subvals: Sequence = ()) -> list:
+    def query(self, statement: str, subvals: Sequence[SQLiteType] = ()) -> list:
         """Execute an SQL statement with substitution values and return
         a list of rows from the database.
         """
         cursor = self.db._connection().execute(statement, subvals)
         return cursor.fetchall()
 
-    def mutate(self, statement: str, subvals: Sequence = ()) -> Any:
+    def mutate(self, statement: str, subvals: Sequence[SQLiteType] = ()) -> Any:
         """Execute an SQL statement with substitution values and return
         the row ID of the last affected row.
         """
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -37,6 +37,7 @@ from . import types
 from .query import (
     AndQuery,
     FieldQuery,
+    FieldQueryType,
     MatchQuery,
     NullSort,
     Query,
@@ -293,7 +294,7 @@ class Model(ABC, Generic[D]):
     are subclasses of `Sort`.
     """
 
-    _queries: dict[str, type[FieldQuery]] = {}
+    _queries: dict[str, FieldQueryType] = {}
     """Named queries that use a field-like `name:value` syntax but which
     do not relate to any specific field.
     """
@@ -718,7 +719,7 @@ class Model(ABC, Generic[D]):
         cls,
         field,
         pattern,
-        query_cls: type[FieldQuery] = MatchQuery,
+        query_cls: FieldQueryType = MatchQuery,
     ) -> FieldQuery:
         """Get a `FieldQuery` for this model."""
         return query_cls(field, pattern, field in cls._fields)
@@ -727,7 +728,7 @@ class Model(ABC, Generic[D]):
     def all_fields_query(
         cls: type[Model],
         pats: Mapping,
-        query_cls: type[FieldQuery] = MatchQuery,
+        query_cls: FieldQueryType = MatchQuery,
     ):
         """Get a query that matches many fields with different patterns.
 
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -55,6 +55,9 @@ else:
     D = TypeVar("D", bound="Database")
 
 
+FlexAttrs = dict[str, str]
+
+
 class DBAccessError(Exception):
     """The SQLite database became inaccessible.
 
@@ -729,7 +732,7 @@ class Model(ABC, Generic[D]):
     @classmethod
     def all_fields_query(
         cls: type[Model],
-        pats: Mapping,
+        pats: Mapping[str, str],
         query_cls: FieldQueryType = MatchQuery,
     ):
         """Get a query that matches many fields with different patterns.
@@ -841,9 +844,9 @@ class Results(Generic[AnyModel]):
             # Objects are pre-sorted (i.e., by the database).
             return self._get_objects()
 
-    def _get_indexed_flex_attrs(self) -> Mapping:
+    def _get_indexed_flex_attrs(self) -> dict[int, FlexAttrs]:
         """Index flexible attributes by the entity id they belong to"""
-        flex_values: dict[int, dict[str, Any]] = {}
+        flex_values: dict[int, FlexAttrs] = {}
         for row in self.flex_rows:
             if row["entity_id"] not in flex_values:
                 flex_values[row["entity_id"]] = {}
@@ -852,7 +855,9 @@ class Results(Generic[AnyModel]):
 
         return flex_values
 
-    def _make_model(self, row, flex_values: dict = {}) -> AnyModel:
+    def _make_model(
+        self, row: sqlite3.Row, flex_values: FlexAttrs = {}
+    ) -> AnyModel:
         """Create a Model object for the given row"""
         cols = dict(row)
         values = {k: v for (k, v) in cols.items() if not k[:4] == "flex"}
@@ -961,7 +966,9 @@ class Transaction:
             self._mutated = False
             self.db._db_lock.release()
 
-    def query(self, statement: str, subvals: Sequence[SQLiteType] = ()) -> list:
+    def query(
+        self, statement: str, subvals: Sequence[SQLiteType] = ()
+    ) -> list[sqlite3.Row]:
         """Execute an SQL statement with substitution values and return
         a list of rows from the database.
         """
@@ -1129,7 +1136,7 @@ class Database:
                 conn.close()
 
     @contextlib.contextmanager
-    def _tx_stack(self) -> Generator[list]:
+    def _tx_stack(self) -> Generator[list[Transaction]]:
         """A context manager providing access to the current thread's
         transaction stack. The context manager synchronizes access to
         the stack map. Transactions should never migrate across threads.
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -47,6 +47,10 @@ from .query import (
 if TYPE_CHECKING:
     from types import TracebackType
 
+    D = TypeVar("D", bound="Database", default=Any)
+else:
+    D = TypeVar("D", bound="Database")
+
 
 class DBAccessError(Exception):
     """The SQLite database became inaccessible.
@@ -236,7 +240,7 @@ class LazyConvertDict:
 # Abstract base for model classes.
 
 
-class Model(ABC):
+class Model(ABC, Generic[D]):
     """An abstract object representing an object in the database. Model
     objects act like dictionaries (i.e., they allow subscript access like
     ``obj['field']``). The same field set is available via attribute
@@ -306,7 +310,7 @@ class Model(ABC):
     """
 
     @cached_classproperty
-    def _relation(cls) -> type[Model]:
+    def _relation(cls):
         """The model that this model is closely related to."""
         return cls
 
@@ -347,7 +351,7 @@ class Model(ABC):
 
     # Basic operation.
 
-    def __init__(self, db: Database | None = None, **values):
+    def __init__(self, db: D | None = None, **values):
         """Create a new object with an optional Database association and
         initial field values.
         """
@@ -363,7 +367,7 @@ class Model(ABC):
     @classmethod
     def _awaken(
         cls: type[AnyModel],
-        db: Database | None = None,
+        db: D | None = None,
         fixed_values: dict[str, Any] = {},
         flex_values: dict[str, Any] = {},
     ) -> AnyModel:
@@ -393,7 +397,7 @@ class Model(ABC):
         if self._db:
             self._revision = self._db.revision
 
-    def _check_db(self, need_id: bool = True) -> Database:
+    def _check_db(self, need_id: bool = True) -> D:
         """Ensure that this object is associated with a database row: it
         has a reference to a database (`_db`) and an id. A ValueError
         exception is raised otherwise.
@@ -637,7 +641,7 @@ class Model(ABC):
                 f"DELETE FROM {self._flex_table} WHERE entity_id=?", (self.id,)
             )
 
-    def add(self, db: Database | None = None):
+    def add(self, db: D | None = None):
         """Add the object to the library database. This object must be
         associated with a database; you can provide one via the `db`
         parameter or use the currently associated database.
@@ -749,8 +753,8 @@ class Results(Generic[AnyModel]):
     def __init__(
         self,
         model_class: type[AnyModel],
-        rows: list[Mapping],
-        db: Database,
+        rows: list[sqlite3.Row],
+        db: D,
         flex_rows,
         query: Query | None = None,
         sort=None,
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -19,7 +19,7 @@ from __future__ import annotations
 import re
 import unicodedata
 from abc import ABC, abstractmethod
-from collections.abc import Collection, Iterator, MutableSequence, Sequence
+from collections.abc import Iterator, MutableSequence, Sequence
 from datetime import datetime, timedelta
 from functools import reduce
 from operator import mul, or_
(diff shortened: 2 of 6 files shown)


beets/dbcore/db.py
```diff
@@ -306,7 +317,7 @@ class Model(ABC):
     """
 
     @cached_classproperty
-    def _relation(cls) -> type[Model]:
+    def _relation(cls):
```
wisp3rwind, line 320: I suspect this should really be ` -> Self`? Maybe we could do

```py
if sys.version_info >= (3, 11):
    from typing import Self
else:
    from typing_extensions import Self
```

to make this work? (We already do this in another place: See the `aura` plugin.)

beets/dbcore/query.py
```diff
@@ -30,6 +30,11 @@
 
 if TYPE_CHECKING:
     from beets.dbcore import Model
+    from beets.dbcore.db import AnyModel
+
+    P = TypeVar("P", default=Any)
+else:
+    P = TypeVar("P")
```
wisp3rwind, line 37: Same question regarding the `default` here as above: Do we really need it; and if so, shouldn't it be a more specific type?

beets/dbcore/query.py
```diff
@@ -107,9 +112,9 @@ def __hash__(self) -> int:
         return hash(type(self))
 
 
-P = TypeVar("P")
-SQLiteType = Union[str, bytes, float, int, memoryview]
+SQLiteType = Union[str, bytes, float, int, memoryview, None]
 AnySQLiteType = TypeVar("AnySQLiteType", bound=SQLiteType)
+FieldQueryType = type["FieldQuery"]
```
wisp3rwind, line 117: Personally, I'd be in favor of sticking with the explicit `type[FieldQuery]`, it seems more obvious to me what it is supposed to be (note how `Type` means different things in `AnySQLiteType` and `FieldQueryType`).

In any case, not a strong opinion here: Feel free to stick to your preference!

beets/dbcore/query.py
```diff
@@ -1061,3 +1066,23 @@ def __eq__(self, other) -> bool:
...
+        field = self.field
+
+        return f"COALESCE(NULLIF({field}_sort, ''), {field}) {collate} {order}"
+
+    def sort(self, objs: list[AnyModel]) -> list[AnyModel]:
+        def key(o):
+            val = o[f"{self.field}_sort"] or o[self.field]
+            return val.lower() if self.case_insensitive else val
+
+        return sorted(objs, key=key, reverse=not self.ascending)
```
wisp3rwind, line 1088: Nice catch, making this general and independent of the specifics of beets' library items! Maybe, it should also be renamed to something more generic (`SmartSort`?) with a corresponding docstring (which could still mention the `artist`/`artist_sort` example)?

beets/dbcore/db.py
```diff
@@ -47,6 +49,15 @@
 if TYPE_CHECKING:
     from types import TracebackType
 
+    from .query import SQLiteType
+
+    D = TypeVar("D", bound="Database", default=Any)
+else:
+    D = TypeVar("D", bound="Database")
```
wisp3rwind, line 56: Do we need to set `default` here? (I wouldn't have expected it: `Model` is abstract, so we can't ever create `Model` objects, where `D` would not be specified.)

Wouldn't it also be easier to do this unconditionally? (If `not TYPE_CHECKING`, `D` isn't really used anyway.)

```suggestion
D = TypeVar("D", bound="Database")
```

or, if we do need the default, make it more specific?

```suggestion
D = TypeVar("D", bound="Database", default="Database")
```

beets/dbcore/types.py
```diff
@@ -37,8 +43,12 @@ def __init__(self, value: Any = None): ...
...
 # nullable, else T and N are set to the same type for the concrete subclasses
 # of Type).
-N = TypeVar("N")
-T = TypeVar("T", bound=ModelType)
+if TYPE_CHECKING:
+    N = TypeVar("N", default=Any)
+    T = TypeVar("T", bound=ModelType, default=Any)
+else:
+    N = TypeVar("N")
+    T = TypeVar("T", bound=ModelType)
```
wisp3rwind, line 51: See above regarding `default` (here, in particular, I'd argue that there should be no default, such that the type checker forces one to specify these types).

wisp3rwind (commented):
Woohoo, this really looks like the pieces are falling into place for wrapping up typings of beets' core :tada:

I left a few small nitpicks, overall this is looking great!

beets/dbcore/db.py
wisp3rwind: Or rather, `type[Self]`.

snejus: I was hoping it would be possible to express this with `Self`, but there are two issues:

1. This returns a model class, so it's more like `type[Self]` 😅 
2. It returns a sibling class compared with the class that the property is under. If I understood their docs correctly

> ... you can import the special type typing.Self that is automatically transformed into a method-level type parameter with the _current class as the **upper bound**_,

this is probably not suitable for our use case?

Ultimately this is just a generic property definition - both of the models define specific returns types where this property is used:

* `Album`:
```py
    @cached_classproperty
    def _relation(cls) -> type[Item]:
        return Item
```
* `Item`:
```py
    @cached_classproperty
    def _relation(cls) -> type[Album]:
        return Album
```

I was going to `raise NotImplementedError` in the generic `Model._relation` implementation, but I found that this kind of broke the `ModelFixture` model in `test_dbcore`, so I just left it to return `cls`.

wisp3rwind: Oh, I misunderstood what's going on here: Even `type[Self]` is definitely wrong, for the reason you explained. Now I'm confused as to why `type[Model]` as it was before doesn't work: From the [mypy docs](https://mypy.readthedocs.io/en/stable/class_basics.html#overriding-statically-typed-methods)

> You can also vary return types covariantly in overriding. For example, you could override the return type Iterable[int] with a subtype such as list[int]. Similarly, you can vary argument types contravariantly – subclasses can have more general argument types.

But feel free to not address this in this PR, and just get in what we have quickly :)

beets/dbcore/query.py
snejus: This saves us from having to specify `Any` when we need to use the generic `FieldQuery` in type definitions, like

```py
    _queries: dict[str, type[FieldQuery]] = {}
```

Without the default `mypy` will complain asking for some type to be specified.

> shouldn't it be a more specific type?

I experimented with this, for example I tried using `object` instead:

```
beets/dbcore/types.py:64: error: Incompatible types in assignment (expression has type "type[SubstringQuery]", variable has type "type[FieldQuery[object]]")  [assignment]
beets/dbcore/types.py:163: error: Incompatible types in assignment (expression has type "type[NumericQuery]", base class "Type" defined the type as "type[FieldQuery[object]]")  [assignment]
beets/dbcore/types.py:244: error: Incompatible types in assignment (expression has type "type[NumericQuery]", variable has type "type[FieldQuery[object]]")  [assignment]
beets/dbcore/types.py:274: error: Incompatible types in assignment (expression has type "type[SubstringQuery]", base class "Type" defined the type as "type[FieldQuery[object]]")  [assignment]
beets/dbcore/types.py:315: error: Incompatible types in assignment (expression has type "type[BooleanQuery]", base class "Type" defined the type as "type[FieldQuery[object]]")  [assignment]
```

This broke it, essentially. As far as I understood anything more specific than `Any` set as default ties to that type, so it kind of loses its generic properties.

beets/dbcore/query.py
snejus: One of the main motivations I had for this was the presence of `types.Type`: in some places it was confusing to see `types.Type` next to `type`, so this helped to remove some of the ambiguity.

beets/dbcore/query.py
snejus: Ah yeah, good point! 

Ah actually, if you look at `sort` implementation, you can see that we've got something hardcoded there that ties it to the `artist` field, as it firstly tries to sort by `{field}_sort` field.

beets/dbcore/db.py
snejus: We use plain `Model` in type definitions without any type on it, so I'd expect mypy to complain. Have just removed the default and ran the analysis:

```
beets/dbcore/query.py:94: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/query.py:159: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/query.py:199: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/query.py:394: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/query.py:533: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/query.py:572: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/query.py:582: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/query.py:608: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/query.py:627: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/query.py:637: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/query.py:828: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/query.py:1001: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/db.py:89: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/db.py:125: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/db.py:156: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/db.py:346: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/db.py:419: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/db.py:735: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/db.py:752: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/db.py:1012: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/queryparse.py:115: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/queryparse.py:183: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/queryparse.py:200: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/queryparse.py:231: error: Missing type parameters for generic type "Model"  [type-arg]
beets/dbcore/queryparse.py:250: error: Missing type parameters for generic type "Model"  [type-arg]
```

snejus: And that's what we see if we try to set the default to `Database`:

(see my message above for my theory regarding this 😅)

```
beets/dbcore/db.py:367: error: Argument 1 to "LazyConvertDict" has incompatible type "Model[D]"; expected "Model[Database]"  [arg-type]
beets/dbcore/db.py:368: error: Argument 1 to "LazyConvertDict" has incompatible type "Model[D]"; expected "Model[Database]"  [arg-type]
beets/dbcore/db.py:432: error: Incompatible return value type (got "Model[D]", expected "Model[Database]")  [return-value]
beets/dbcore/db.py:432: note: Perhaps you need a type annotation for "new"? Suggestion: "Model[Database]"
beets/dbcore/db.py:449: error: Invalid self argument "type[Model[D]]" to class attribute function "_getters" with type "Callable[[type[Model[Database]]], Any]"  [misc]
beets/dbcore/db.py:517: error: Invalid self argument "type[Model[D]]" to class attribute function "_getters" with type "Callable[[type[Model[Database]]], Any]"  [misc]
beets/dbcore/db.py:686: error: Argument 1 has incompatible type "Model[D]"; expected "Model[Database]"  [arg-type]
Found 6 errors in 1 file (checked 1 source file)
```

beets/dbcore/types.py
snejus: As I mentioned this has to do with those type annotations where we use the generic type, in this case `Type`. For example, let's take `db.py` which has these annotations:

```py
    _fields: dict[str, types.Type] = {}
    _types: dict[str, types.Type] = {}
    def _type(cls, key) -> types.Type:
    def _make_table(self, table: str, fields: Mapping[str, types.Type]):
```

By default `mypy` is unhappy with `Missing type parameters for generic type "Type"`

A couple of possible ways to deal with this:

1. Explicit `Any`
```py
    _fields: dict[str, types.Type[Any]] = {}
    _types: dict[str, types.Type[Any]] = {}
    def _type(cls, key) -> types.Type[Any]:
    def _make_table(self, table: str, fields: Mapping[str, types.Type[Any]]):
```

2. Explicit _dry_ `Any`
```py
AnyType = types.Type[Any]

    _fields: dict[str, AnyType] = {}
    _types: dict[str, AnyType] = {}
    def _type(cls, key) -> AnyType:
    def _make_table(self, table: str, fields: Mapping[str, AnyType]):
```

3. `Any` defined as default on the typevar

```py
T = TypeVar("T", bound=ModelType, default=Any)
```

4. Or just disable this check with `allow_any_generics = true`


The main issue with **1** is that it adds complexity to the type definitions and creates noise. **2** is slightly less bad, however we'd need to create such variable for each generic type we want to use, it needs importing and adds indirecttion.

I found that specifying the default on the typevar addresses most of the issues above - we can use the actual type directly, it doesn't add any noise and we don't need to define additional variables. We do need to define it in a funny way using `TYPE_CHECKING` however 😅 

Of course we could always just disable this check, but this will affect all generic types, and mypy won't anymore enforce specific types for `list`, `dict` etc.

beets/dbcore/query.py
wisp3rwind: Ok! Probably, whatever we do, there will some ambiguity around the term "type" in `dbcore`.

beets/dbcore/query.py
wisp3rwind: > Ah actually, if you look at `sort` implementation, you can see that we've got something hardcoded there that ties it to the `artist` field, as it firstly tries to sort by `{field}_sort` field.

Ok, but conceptually, this is a more generic mechanism now which could be used for other fields.

I think this boils down to deciding whether we treat this as generic dbcore code (in that case, name and docs should be adapted) or as `Library`-specific (in that case, this code should remain in `beets/library.py`). Currently, it's somewhere half-way, eroding the separation of concerns we have between `library` and `dbcore`.

beets/dbcore/db.py
wisp3rwind: Ok, fair enough: I've also been playing around with this locally for a bit. My impression is that in some cases (e.g. `dbcore.db.FormattedMapping`), we really want `Model[Any]` since the resulting type really behaves the same for any `Model` subclass. In other cases (e.g. some `Database` methods), we could be more precise about the specific subclass (because there, we know that `D = Self`). Yet in other cases, we could get stricter typing by making more classes generic (e.e. `dbcore.db.Results`), but that would make things more complicated, and it also seems that I run into https://github.com/python/typing/issues/548.

I'd conclude to keep the `default=Any`, but try to remember that there are places where we could revisit and tighten typing at some point.

beets/dbcore/types.py
wisp3rwind: Yes, let's stick to 3, cf. my comment in another thread.

beets/dbcore/db.py
wisp3rwind: More precisely, I'd say this should be `type[Model[D]]`, but right now, it seems that for me locally, mypy doesn't really check this at all: I don't even manage to make it complain about that line.

snejus: > Now I'm confused as to why type[Model] as it was before doesn't work: From the [mypy docs](https://mypy.readthedocs.io/en/stable/class_basics.html#overriding-statically-typed-methods)

Now this has to do with the custom descriptor `cached_classproperty` - I have spent far too many hours battling with mypy trying to make it aware that it works the same way like `classmethod` and that `cls` is the **class** and not the object instance. I even learned how to write mypy plugins but I still failed 😅

Try adding the output type and see what happens

```py
    @cached_classproperty
    def _relation(cls) -> type[Model]:
        """The model that this model is closely related to."""
        return cls  # mypy: Incompatible return value type (got "Model[D]", expected "type[Model[Any]]")
```

wisp3rwind: That's what I did — but nothing bad happened... but I'm not going to follow you into that rabbit hole and start investigating this now ;)

wisp3rwind (approved):
Looks good to me! Feel free to merge this after addressing the question of where the `SmartArtistSort` should go.

beets/dbcore/db.py
snejus: Best of luck! It would be amazing if you can figure this out.

snejus: Just thought I'd note that I use it in [`beetcamp`](https://github.com/snejus/beetcamp/blob/main/beetsplug/bandcamp/catalognum.py) slightly differently and it works as expected there:

```py
class cached_classproperty(Generic[T]):
    def __init__(self, getter: Callable[..., T]) -> None:
        self.getter = getter
        self.cache: dict[type[object], T] = {}

    def __get__(self, instance: object, owner: type[object]) -> T:
        if owner not in self.cache:
            self.cache[owner] = self.getter(owner)

        return self.cache[owner]


def cached_patternprop(
    pattern: str, flags: int = 0
) -> cached_classproperty[Pattern[str]]:
    """Pattern is compiled and cached the first time it is accessed."""
    return cached_classproperty(lambda _: re.compile(pattern, flags))


@dataclass
class Catalognum:
    ...
    start_end = cached_patternprop(rf"((^{MATCH})|({MATCH}$))", re.M | re.VERBOSE)
    # enclosed by parens or square brackets, but not ending with MIX
    delimited = cached_patternprop(rf"(?:[\[(])({MATCH})(?:[])]|$)", re.VERBOSE)
```

`Catalognum.start_end` and `Catalognum.delimited` are always correctly resolved as `re.Pattern`.

beets/dbcore/query.py
wisp3rwind: I think we actually agree here 😅

I don't have much of an opinion on where it belongs, I'd just argue for a rename and doc rewording in the case of moving it to `dbcore`.

beets/dbcore/db.py
wisp3rwind: I think the reason is that the typings for `cached_classproperty` exist in beetcamp, but not in beets ;)

On the other hand, maybe we should simply remove the decorator from `_relation` and turn it into a regular `classmethod`: It's probably too simple to benefit from caching, and in any case, it seems to only be called from methods which in turn are cached using `cached_classproperty`. When I do that, `-> type[Model[D]]` seems to work for mypy.

snejus: To be honest `cached_` is just a buzzword there, because ultimately it is a _class_ property, and a class is evaluated just once, so 'caching' is kind of inherent here.

The reason why I didn't use a classmethod here comes down to whether the returned data is constant or dynamic. Since it's constant, I thought using a classmethod would be slightly misleading.

Ideally we'd have them set as attributes on the class:

```py
class Album(LibModel):
    _relation = Item
    

class Item(LibModel):
    _relation = Album
```

But you can see this is a chicken and egg problem 😅 since `Item` class is undefined when `Album` class wants to use it. I created this classproperty to work around this limitation, so that they behave as if they were defined as attributes on the class.

beets/dbcore/query.py
wisp3rwind: TBH, I don't have a really good idea, either...

Let's not block the PR over this, feel free to merge it as-is!

beets/dbcore/db.py
wisp3rwind: Yeah, I played around with that code myself, only to run into the same issue 😅

Maybe you could open a quick PR to add typings to `cached_classproperty`? (i.e. just copy them over from beetcamp.)

PR #5540, opened 2024-12-07
Stop perpetually writing `mb_artistid`, `mb_albumartistid` and `albumtypes` fields

snejus:
This PR fixes an issue where the `beet write` command repeatedly shows differences for certain fields (`mb_artistid`, `mb_albumartistid`, `albumtype`) even after writing the tags. 
This happens because these fields are actually stored as lists in the media files (`mb_artistids`, `mb_albumartistids`, `albumtypes`), but beets maintains both single and list versions in its database.

This PR addresses this issue in a non-invasive way: the fix ensures consistency between single fields and their list counterparts by:
1. When setting a single field value, making it the first element of the corresponding list
2. When setting a list, using its first element as the single field value

This resolves long-standing issues #5265, #5371, and #4715 where users experienced persistent "differences" in these fields despite writing tags multiple times.

Changes:
- Added `ensure_consistent_list_fields()` function to synchronize field pairs
- Applied this function during metadata application
- Added tests for all field combinations

Fixes #5265, #5371, #4715

**Note**: your music needs to be reimported with `beet import -LI` or synchronised with `beet mbsync` in order to fix these fields!

PR #5548, opened 2024-12-15
Add check for empty album and handle them gracefully

arsaboo:
## Description

Fixes https://github.com/beetbox/beets/pull/5540#issuecomment-2543934811


arsaboo:
@snejus I tested this and can confirm that it works.

docs/changelog.rst
```diff
@@ -89,6 +89,7 @@ New features:
 
 Bug fixes:
 
+* Add empty album check in ``mbsync``.
```
snejus, line 92:
```suggestion
* :doc:`plugins/mbsync`: Skip empty albums.
```

snejus (commented):
I'm now thinking: how come do you have empty albums in your library? 😅

arsaboo:
Haha...I wish I knew 😜

Serene-Arc:
Perhaps we should add a function to `beet update` to purge these empty albums? We use the `update` command to do a lot of cleanup and other functions, so this might be good to fold in. Is there a reason we'd want to keep empty albums in a library database?

snejus:
> Is there a reason we'd want to keep empty albums in a library database?

Definitely not. The funny thing here is that `beets` never ends up _**adding** empty albums in the first place_. The only times I've come across empty albums in my library have been when this happened due to some funky adjustments I made in the source. Could this potentially be the reason why you have them as well @arsaboo?

arsaboo:
Indeed, it's quite possible those empty albums were created during one of my many tests. We can close this PR for now. If it's comes back again, we can remove the empty albums (which may be a better solution). Unfortunately, I am unable to reproduce it right now to fix it.

PR #5552, opened 2024-12-18
pyproject.toml: don't install docs et al into the wheel

ArsenArsen:
## Description

wheels are directly unpacked into site-packages, so this means likely conflict with other packages

## To Do

if a changelog entry is needed, I am happy to add one.  tests/docs seem not applicable (as this is purely a packaging issue)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,11 +26,11 @@ packages = [
     { include = "beetsplug" },
 ]
 include = [ # extra files to include in the sdist
-    "docs",
-    "extra",
-    "man/**/*",
-    "test/*.py",
-    "test/rsrc/**/*",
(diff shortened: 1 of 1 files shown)


snejus:
image

I should have not trusted [poetry docs](https://python-poetry.org/docs/pyproject/#include-and-exclude) which say that `format` parameter defaults to `sdist` 🤡

ArsenArsen:
> image
> 
> I should have not trusted [poetry docs](https://python-poetry.org/docs/pyproject/#include-and-exclude) which say that `format` parameter defaults to `sdist` 🤡 

heh, they seem to have changed it between versions.. which means one can't rely on it right now 

thanks for merging!

PR #5566, opened 2024-12-27
Future proof BucketPluginTest.test_year_single_year_last_folder

stefanor:
2025 won't be in the future, forever.

Fixes: https://bugs.debian.org/1091495


snejus (approved):
Well spotted, thanks!

PR #5568, opened 2024-12-29
Album art browser in web plugin UI

valrus:
## Description

This PR adds a "browse albums" page to the web plugin's UI, with a couple tweaks (via query params, to preserve backwards compatibility) to the API in order to enable it. This is a feature that's been [requested](https://discourse.beets.io/t/anyone-know-a-way-to-browse-by-cover-art/1073) and [kind of hacked together](https://github.com/beetbox/beets/discussions/4063) in the past. This implementation has the following characteristics:

1. Album covers displayed in a grid, sorta similar to [CoverGrid](https://www.suruatoel.xyz/codes/mcg) for mpd.
2. The grid is not interactable (at this time) — it's only for browsing.
3. Albums will be loaded in _random_ order. This is more fun than seeing the alphabetically first albums on every load.
4. Album covers load as they scroll into view, to avoid making thousands of requests to the server at once for big libraries. (But I think it's still possible to send a lot of requests at once by dragging the scrollbar.)
5. It's possible to switch between the item view and album browser and the playing track will continue to play.

It includes assorted other improvements to the web plugin as well. A list summarizing the changes:

1. Entering a search query now shows "Loading..." in the results list while the request is in progress.
2. The search form is now "sticky" and stays at the top of the results list when you scroll down.
3. The player now shows the name of the playing song, if one is playing.
4. Most CSS has been converted from floats to CSS flexbox and grid as appropriate. (This may break compatibility with some very old browsers, so I'm fine reverting it if it's important to keep that compatibility.)
5. [underscore.js](https://underscorejs.org/) and [backbone.js](https://backbonejs.org/) have been updated to the most recent versions. (Ditto potential compatibility issues; I haven't checked but I will if it's important.)
6. Non-breaking web API change: if a request to the `album_art` route includes the `b64` query string (regardless of its value), the art data will be returned as base64 data rather than binary. This was mainly because this conversion was easier (for me) to do in Python than in JavaScript.
7. Non-breaking web API change: if a request to the `all_albums` route includes the `random` query string (regardless of value), the albums will be returned in random order.

Here is what the browser looks like. I figured others will likely have libraries like mine where the album art situation is patchy so I tried to make the alt text look nice:
beets_browser

## To Do

I haven't done any of these yet because I'm not sure this will even get merged, much less what from its current state will end up making it in. It doesn't really change any of the documented functionality of the web plugin. And there are no tests as the JS is completely untested and the API changes are not straightforward or particularly interesting to test (randomness? base-64 encoding?)

That said, I'm happy to do any of these that reviewers would like me to and I'll certainly add a changelog entry if this gets to a mergeable state.

- [ ] Documentation.
- [ ] Changelog.
- [ ] Tests.

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -20,6 +20,7 @@ import os
 
 import flask
 from flask import g, jsonify
+import random
 from unidecode import unidecode
 from werkzeug.routing import BaseConverter, PathConverter
 
@@ -374,10 +375,14 @@ def get_album(id):
 
 
 @app.route("/album/")
+@app.route("/album")
 @app.route("/album/query/")
 @resource_list("albums")
 def all_albums():
-    return g.lib.albums()
+    albums = list(g.lib.albums())
+    if 'random' in request.args:
+        random.shuffle(albums)
+    return albums
 
 
 @app.route("/album/query/<query:queries>", methods=["GET", "DELETE"])
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -77,11 +77,8 @@ def json_generator(items, root, expand=False):
     :returns:     generator that yields strings
     """
     yield '{"%s":[' % root
-    first = True
-    for item in items:
-        if first:
-            first = False
-        else:
+    for index, item in enumerate(items):
+        if index != 0:
             yield ","
         yield json.dumps(_rep(item, expand=expand))
     yield "]}"
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -17,10 +17,12 @@
 import base64
 import json
 import os
+from io import BytesIO
 
 import flask
-from flask import g, jsonify
+import mimetypes
 import random
+from flask import g, jsonify, request
 from unidecode import unidecode
 from werkzeug.routing import BaseConverter, PathConverter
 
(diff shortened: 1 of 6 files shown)


semohr:
Just noticed this PR, thanks for putting in the work! I’ll try to carve out some time to review it soon. At +3,980 / −1,708 lines, it’s definitely a hefty one. 

---

As a small FYI, you might also be interested in the album-grid implementation we’ve been working on in beets-flask:
https://github.com/pSpitzner/beets-flask

Here’s a quick preview:
image
 
We’d be more than happy to welcome more contributions if you’re interested in helping shape the frontend. :upside_down_face:

valrus:
Oh that's super cool. I was actually just brushing off this PR to see if I could maybe package it separately so it wouldn't have to merge into beets itself. Because it is a chonky one for some "just for fun" functionality and I'm not sure reviewing it is the best use of maintainers' time.

Let me close this out so you don't feel the need to review it while I explore other options.

semohr:
Alright, thank you!

We’re more than happy to welcome additional contributors in bees-flask ;) If you feel our goals align and want to pull in the same direction with us, please don’t hesitate to reach out!

valrus:
This is now a 3rd party plugin! https://pypi.org/project/beetsplug-covergrid/

PR #5570, opened 2024-12-29
Fix sorting on missing non-string fields

### LINKED ISSUE #5512: mpdstats: Error when querying `play_count` with a sort order
### Error when sorting certain mpdstats queries

```sh
$ beet -vv list play_count+
```

Led to this problem:

```
user configuration: ~/.config/beets/config.yaml
data directory: ~/.config/beets
plugin paths: 
Sending event: pluginload
library database: ~/Music/musiclibrary.db
library directory: ~/Music
Sending event: library_opened
Parsed query: AndQuery([TrueQuery()])
Parsed sort: SlowFieldSort('play_count', ascending=True)
  File "/usr/bin/beet", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/lib/python3.12/site-packages/beets/ui/__init__.py", line 1865, in main
    _raw_main(args)
  File "/usr/lib/python3.12/site-packages/beets/ui/__init__.py", line 1852, in _raw_main
  File "/usr/lib/python3.12/site-packages/beets/ui/commands.py", line 1599, in list_func
  File "/usr/lib/python3.12/site-packages/beets/ui/commands.py", line 1594, in list_items
    for item in lib.items(query):
                ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/beets/dbcore/db.py", line 823, in __iter__
    objects = self.sort.sort(list(self._get_objects()))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/beets/dbcore/query.py", line 984, in sort
    return sorted(objs, key=key, reverse=not self.ascending)
TypeError: '<' not supported between instances of 'int' and 'str'
```
### Setup

* OS: Artix Linux
* Python version: 3.12.7
* beets version: 1.6.1
* Turning off plugins made problem go away (yes/no): yes

```yaml
directory: ~/Music

library: ~/Music/musiclibrary.db

plugins: mpdstats

import:
    copy: no
mpd:
    music_directory: ~/Music
    strip_path: ''
    rating: yes
    rating_mix: 0.75
    host: localhost
    port: 6600
    password: REDACTED
```

It also happens with `list rating+` and `list skip_count+`. It happens with both ascending (`play_count+`) and descending (`play_count-`) sorting.

valrus:
## Description

Fixes #5512. When sorting on a field, if the field is missing from some items and it has a type, use the type's `null` value. Otherwise, continue to fall back to an empty string, as I don't think there's much to be done in that case.

The new test `test_int_field_present_in_some_items` fails without the fix in `query.py`.

- [x] Documentation.
- [x] Changelog.
- [x] Tests.


test/test_sort.py
```diff
@@ -514,6 +519,24 @@ def test_field_present_in_some_items(self):
         # items without field last
         assert [i.id for i in results_desc] == [ids[2], ids[1], ids[0], ids[3]]
 
+    def test_int_field_present_in_some_items(self):
+        """Test ordering by a field not present on all items."""
+        # append int-valued 'foo' to two items (1,2)
+        Item._types = {"foo": types.Integer()}
```
snejus, line 525: Instead of overwriting global `Item` try patching it:

```
    @patch("beets.library.Item._types", {"foo": types.Integer()})
    def test_int_field_present_in_some_items(self):
```

Then you can remove the `tearDown` that got added in line 475.

test/test_sort.py
```diff
@@ -514,6 +519,24 @@ def test_field_present_in_some_items(self):
...
+        items[1].store()
+        items[2].store()
+
+        results_asc = list(self.lib.items("foo+ id+"))
+        # items without field first
+        assert [i.id for i in results_asc] == [ids[0], ids[3], ids[1], ids[2]]
+        results_desc = list(self.lib.items("foo- id+"))
+        # items without field last
+        assert [i.id for i in results_desc] == [ids[2], ids[1], ids[0], ids[3]]
+
```
snejus, line 539: Consider naming variables in a slightly more clear way, something like this

```suggestion
    @patch("beets.library.Item._types", {"foo": types.Integer()})
    def test_int_field_present_in_some_items(self):
        """Test ordering by a field not present on all items."""
        # append int-valued 'foo' to two items (1,2)
        items = self.lib.items("id+")
        lower_foo_item, higher_foo_item, *items_without_foo = items
        lower_foo_item.foo = 1
        higher_foo_item.foo = 2
        lower_foo_item.store()
        higher_foo_item.store()

        items_without_foo_ids = [i.id for i in items_without_foo]
        results_asc = list(self.lib.items("foo+ id+"))
        # items without field first
        assert [i.id for i in results_asc] == [
            *items_without_foo_ids,
            lower_foo_item.id,
            higher_foo_item.id,
        ]
        results_desc = list(self.lib.items("foo- id+"))
        # items without field last
        assert [i.id for i in results_desc] == [
            higher_foo_item.id,
            lower_foo_item.id,
            *items_without_foo_ids,
        ]

```

snejus (changes requested):
Thanks for this, I tested it and it works fine! Added a couple of comments

test/test_sort.py
valrus: Nice, thank you. I copied this pattern from some tests in `test_query.py` so I updated those to patch as well, for consistency's sake.

test/test_sort.py
valrus: Done. Like the other change this one was c/p from elsewhere so I modified the source too.

test/test_sort.py
snejus: Amazing stuff!

snejus:
Looks great. I see that `mypy` complains about missing `mock` stubs. Could you by any chance add `types-mock` to testing dependencies?

valrus:
> Could you by any chance add types-mock to testing dependencies?

I'm guessing you meant _typing_ dependencies and added it there instead, but let me know if I misunderstood. There are a bunch of typecheck errors now but none of them are in the changed files. Is something wrong or is this expected?

snejus:
> > Could you by any chance add types-mock to testing dependencies?
> 
> I'm guessing you meant _typing_ dependencies and added it there instead, but let me know if I misunderstood. 

You're completely correct! Thanks

> There are a bunch of typecheck errors now but none of them are in the changed files. Is something wrong or is this expected?

That's expected - `mypy` should only check the changed files but it ends up reporting issues in related files for some reason 🤷🏼‍♂️ but that's to be sorted separately

snejus (approved):
Good stuff, thanks for this!

PR #5571, opened 2024-12-30
Fix track matching regression

snejus:
## Problem
A regression was introduced when adjusting the track matching logic to use `lapjv` instead of `munkres`. The `lapjv` algorithm returns `-1` for unmatched items, which wasn't being handled correctly in the matching logic. This caused incorrect track assignments when importing new music.

## Solution
- Modified the mapping creation to filter out unmatched items (where index is `-1`)
- Updated test case to properly catch this scenario

### DIFF
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -127,15 +127,21 @@ def assign_items(
     objects. These "extra" objects occur when there is an unequal number
     of objects of the two types.
     """
+    log.debug("Computing track assignment...")
     # Construct the cost matrix.
     costs = [[float(track_distance(i, t)) for t in tracks] for i in items]
-    # Find a minimum-cost bipartite matching.
-    log.debug("Computing track assignment...")
-    cost, _, assigned_idxs = lap.lapjv(np.array(costs), extend_cost=True)
+    # Assign items to tracks
+    _, _, assigned_item_idxs = lap.lapjv(np.array(costs), extend_cost=True)
     log.debug("...done.")
 
-    # Produce the output matching.
-    mapping = {items[i]: tracks[t] for (t, i) in enumerate(assigned_idxs)}
+    # Each item in `assigned_item_idxs` list corresponds to a track in the
+    # `tracks` list. Each value is either an index into the assigned item in
+    # `items` list, or -1 if that track has no match.
+    mapping = {
+        items[iidx]: t
+        for iidx, t in zip(assigned_item_idxs, tracks)
+        if iidx != -1
+    }
     extra_items = list(set(items) - mapping.keys())
     extra_items.sort(key=lambda i: (i.disc, i.track, i.title))
     extra_tracks = list(set(tracks) - set(mapping.values()))
--- a/test/test_autotag.py
+++ b/test/test_autotag.py
@@ -23,7 +23,7 @@ from beets import autotag, config
 from beets.autotag import AlbumInfo, TrackInfo, correct_list_fields, match
 from beets.autotag.hooks import Distance, string_dist
 from beets.library import Item
-from beets.test.helper import BeetsTestCase
+from beets.test.helper import BeetsTestCase, ConfigMixin
 from beets.util import plurality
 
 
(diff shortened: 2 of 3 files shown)


beets/autotag/match.py
```diff
@@ -131,11 +131,15 @@ def assign_items(
...
-    cost, _, assigned_idxs = lap.lapjv(np.array(costs), extend_cost=True)
+    cost, _, assigned_item_idxs = lap.lapjv(np.array(costs), extend_cost=True)
     log.debug("...done.")
 
     # Produce the output matching.
-    mapping = {items[i]: tracks[t] for (t, i) in enumerate(assigned_idxs)}
+    mapping = {
+        items[iidx]: tracks[tidx]
+        for tidx, iidx in enumerate(assigned_item_idxs)
+        if iidx != -1
```
wisp3rwind, line 141: Maybe, to be extra-safe against this type of issue (which seems to be pretty subtle, at a first glance I was very confused why the testcase didn't catch it in the first place), the comparison could be turned into
```suggestion
        if iidx >= 0
```


This might also be worth a code comment about the meaning of `-1`? At a first glance, the `lapvj` documentation doesn't mention it.

test/test_autotag.py
```diff
@@ -560,8 +560,8 @@ def test_order_works_with_missing_tracks(self):
             items, trackinfo
         )
         assert extra_items == []
-        assert extra_tracks == [trackinfo[1]]
-        assert mapping == {items[0]: trackinfo[0], items[1]: trackinfo[2]}
+        assert extra_tracks == [trackinfo[2]]
+        assert mapping == {items[0]: trackinfo[0], items[1]: trackinfo[1]}
```
wisp3rwind, line 564: I think I understand why this change was made (the original code didn't fail the test due to indexing the arrays from the back with the spurious -1 index), and this made the test fail.

However, the original test case seems to cover something slightly different (track assignment with missing tracks with gaps, i.e. just `zip(track_infos, items)` (stopping when the shortest argument ends) wouldn't give the correct assignment.

Maybe, it would make sense to enhance the test case such that it tests missing tracks at various positions:

```py
    def test_order_works_with_missing_tracks(self):
    	all_items = [
            self.item("one", 1),
            self.item("two", 2),
            self.item("three", 3),
        ]
        all_trackinfos = [
            TrackInfo(title="one"),
            TrackInfo(title="two"),
            TrackInfo(title="three"),
        ]
    	for missing_idx in range(3):
            items = all_items[:]
            items.pop(missing_idx)
            trackinfos = all_trackinfos[:]
            trackinfos.pop(missing_idx)
            mapping, extra_items, extra_tracks = match.assign_items(
                items, all_trackinfos
            )
            assert extra_items == []
            assert extra_tracks == [trackinfo[missing_idx]]
            assert mapping == dict(zip(items, trackinfos))
```

I didn't test this, but this should contain both the original and your modified test case.

wisp3rwind (changes requested):
Thanks for the quick follow-up fix!

This looks fine in principle, I left a few suggestions that might make this more robust.

beets/autotag/match.py
snejus: I added some comments in the logic and tried making it a bit cleaner. Do you prefer `iidx >= 0` over `iidx != -1`? I find the latter more explicit and self-documenting regarding the expected index for unmatched tracks.

test/test_autotag.py
snejus: I refactored the tests and added the test cases that you mentioned above

beets/autotag/match.py
wisp3rwind: Sounds good! Let's stick to the current version: I guess it's exceedingly unlikely that lapjv could ever start returning other negative values. (And either way, our tests should hopefully catch that case now.)

snejus:
@wisp3rwind note that I 'borrowed' the `ConfigMixin` implementation (see the last commit) from #5452 where it was introduced for the same purpose - to remove tests dependency on local configuration.

arsaboo:
@snejus just tested this branch, and the tracks match correctly.

Issue #2661, opened 2017-08-20
Problems with missing plugin, reporting albums in database


gboysko:
### Problem

When I try to find missing albums, it reports albums already in my database.

```sh
$ beet ls --album tycho
Tycho - Awake
Tycho - Epoch
$ beet missing -a tycho
Tycho - Division Remix
Tycho - Descent
Tycho - Spectre Remix
Tycho - Division
Tycho - Elsewhere
Tycho - Dive
Tycho - Awake Remix
Tycho - Hours
Tycho - Montana
Tycho - Awake
Tycho - Sunrise Projector
Tycho - The Daydream / The Disconnect
Tycho - Past Is Prologue
Tycho - Spectre
Tycho - Adrift / From Home
Tycho - Black Sunrise
Tycho - EDC Sunrise 2016
Tycho - Epoch
Tycho - See Remix
Tycho - Coastal Brake
Tycho - See
Tycho - Awake Remixes
```
Note that `Tycho - Epoch` and `Tycho - Awake` are both listed in the `missing -a` output.

### Setup

* OS: Ubuntu 16.04
* Python version: 2.7
* beets version: 1.4.5
* Turning off plugins made problem go away (yes/no): No. By removing all plugins (other than `missing`), I still see the incorrect `missing` output.

My configuration (output of `beet config`) is:

```yaml
directory: /mnt/980425CD0425AF66/MP3s.new
import:
#    move: yes
    write: yes
original_date: yes
asciify_paths: yes
id3v23: yes
# plugins: edit embedart fetchart lastgenre scrub missing info replaygain gmusic
plugins: missing
paths:
    default: $albumartist/$album%aunique{}/$track $title
replaygain:
    backend: gstreamer
embedart:
    ifempty: yes
    remove_art_file: yes
gmusic:
    email: 'REDACTED'
    password: 'REDACTED'
```

sampsyo:
Looking a little closer at how the plugin works, it seems to match based on the MusicBrainz release group ID (`mb_releasegroupid`), not the title. Can you check, perhaps using `beet ls -f '$mb_releasegroupid'`, whether it's reporting albums with duplicate IDs?

gboysko:
```sh
$ beet ls -a -f '$mb_releasegroupid' tycho
b4eab0c5-3d27-411f-b2b7-a03a30ced66b
51842820-7c80-4e23-be8d-6793a4bb7935
```
[...]

PR #5587, opened 2025-01-11
Change missing plugin to allow for filtering albums by release type

bgrassy:
## Description

Addresses #2661.  

Currently, the missing plugin when ran in album mode only allows for getting all release groups attached to a single artist. Users may want to restrict this search to only show releases of a specific type (such as albums or compilations). This CR adds a new `--release-type` flag to the missing plugin. If users want to filter to a specific type (or set of types), they simply need to provide this flag for every release type that they want included. 

As part of this change, the default behavior has been shifted to only select `album` type releases as is suggested in the issue- to avoid breaking default behavior I could easily switch this back. I am also wondering if it might make sense to address the following idea (https://github.com/beetbox/beets/discussions/5101) in a follow-up.

The tests I wrote are written using `pytest`, as it is suggested in the documentation to stay away from `unittest`. As a result, I required a slightly different mixin inheritance schema than is usually seen. If there are any other recommended ways to accomplish this I'd be happy to switch!

## To Do

### DIFF
--- a/beetsplug/missing.py
+++ b/beetsplug/missing.py
@@ -18,14 +18,14 @@
 from collections import defaultdict
 
 import musicbrainzngs
-from musicbrainzngs.musicbrainz import MusicBrainzError
+from musicbrainzngs.musicbrainz import VALID_RELEASE_TYPES, MusicBrainzError
 
 from beets import config
 from beets.autotag import hooks
 from beets.dbcore import types
 from beets.library import Item
 from beets.plugins import BeetsPlugin
-from beets.ui import Subcommand, decargs, print_
+from beets.ui import Subcommand, decargs
 
 
 def _missing_count(album):
@@ -98,6 +98,7 @@ class MissingPlugin(BeetsPlugin):
                 "count": False,
                 "total": False,
                 "album": False,
+                "release_type": ["album"],
             }
         )
 
@@ -125,6 +126,15 @@ class MissingPlugin(BeetsPlugin):
             action="store_true",
             help="show missing albums for artist instead of tracks",
         )
+        self._command.parser.add_option(
+            "--release-type",
+            dest="release_type",
+            action="append",
+            help=(
+                "select release types for missing albums for artist "
+                f"from ({', '.join(VALID_RELEASE_TYPES)})"
+            ),
+        )
         self._command.parser.add_format_option()
 
     def commands(self):
@@ -149,7 +159,7 @@ class MissingPlugin(BeetsPlugin):
         fmt = config["format_album" if count else "format_item"].get()
 
         if total:
-            print(sum([_missing_count(a) for a in albums]))
+            self._log.info(str(sum([_missing_count(a) for a in albums])))
             return
 
         # Default format string for count mode.
@@ -159,17 +169,18 @@ class MissingPlugin(BeetsPlugin):
         for album in albums:
             if count:
                 if _missing_count(album):
-                    print_(format(album, fmt))
+                    self._log.info(format(album, fmt))
 
             else:
                 for item in self._missing(album):
-                    print_(format(item, fmt))
+                    self._log.info(format(item, fmt))
 
     def _missing_albums(self, lib, query):
         """Print a listing of albums missing from each artist in the library
         matching query.
         """
         total = self.config["total"].get()
+        release_type = self.config["release_type"].get() or ["album"]
 
         albums = lib.albums(query)
         # build dict mapping artist to list of their albums in library
@@ -193,7 +204,10 @@ class MissingPlugin(BeetsPlugin):
                 continue
 
             try:
-                resp = musicbrainzngs.browse_release_groups(artist=artist[1])
+                resp = musicbrainzngs.browse_release_groups(
+                    artist=artist[1],
+                    release_type=release_type,
+                )
                 release_groups = resp["release-group-list"]
             except MusicBrainzError as err:
                 self._log.info(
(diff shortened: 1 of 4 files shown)


beetsplug/_utils/musicbrainz.py
```diff
@@ -235,8 +236,15 @@ def browse_release_groups(
         """Browse release groups related to the given entities.
 
         At least one of artist, collection, or release must be provided.
+        Optionally filter by release_type (e.g., ["album", "ep"]).
         """
-        return self._get_resource("release-group", **kwargs)["release-groups"]
+        api_params: dict[str, Any] = dict(kwargs)
```
snejus: Maybe it would a bit cleaner to format `type` into the expected shape in `missing.py`?

bgrassy: No worries, thanks for the review! Agreed, made the tweak. Just rebased on top of some other changes that have been merged in the interim

beetsplug/missing.py
```diff
@@ -108,6 +127,7 @@ def __init__(self):
                 "count": False,
                 "total": False,
                 "album": False,
+                "release_type": ["album"],
```
snejus:
```suggestion
                "release_types": ["album"],
```

beetsplug/missing.py
```diff
@@ -191,13 +223,17 @@ def _missing_albums(self, lib: Library, query: list[str]) -> None:
             # reporting the same set of missing albums. Instead, we should
             # group by `mb_albumartistid` field only.
             artist = (album["albumartist"], album["mb_albumartistid"])
-            album_ids_by_artist[artist].add(album)
+            album_ids_by_artist[artist].add(album["mb_releasegroupid"])
 
         total_missing = 0
+        release_type = self.config["release_type"].get() or ["album"]
```
snejus: 1. Call `as_str_seq()` here to make sure it handles `release_types: a b c`
2. No need to default to `["album"]` as that's already the default. I think we should leave users the option to specify an empty list `release_types: []`.

```suggestion
        release_type = self.config["release_type"].as_str_seq()
```

beetsplug/missing.py
```diff
@@ -191,13 +223,17 @@ def _missing_albums(self, lib: Library, query: list[str]) -> None:
...
         calculating_total = self.config["total"].get()
         for (artist, artist_id), album_ids in album_ids_by_artist.items():
             try:
-                resp = self.mb_api.browse_release_groups(artist=artist_id)
+                resp = self.mb_api.browse_release_groups(
+                    artist=artist_id,
+                    type="|".join(release_type),
```
snejus: Given my comment above, we may have an empty list here. What happens if we submit `type=""` to MB API? If the results are the same as if no types are supplied, then we can leave it as it is.

Otherwise, maybe we should check that the list is not empty and only then supply this argument?

docs/changelog.rst
```diff
@@ -15,10 +15,17 @@ New features
...
   ``media``, and ``year``) in Discogs search queries.
+- :doc:`plugins/missing`: When running in missing album mode, allows users to
+  specify MusicBrainz release types to show using the ``--release-type`` flag.
+  The default behavior is also changed to just show releases of type ``album``.
+  :bug:`2661`
 
 Bug fixes
 ~~~~~~~~~
 
+- :doc:`plugins/missing`: Fix ``--album`` mode incorrectly reporting albums
```
snejus, line 31: Love it!

test/plugins/test_missing.py
```diff
@@ -37,26 +38,184 @@ class TestMissingAlbums(IOMixin, PluginMixin):
...
+                album="Album",
+                albumartist="Artist",
+                mb_albumartistid=artist_mbid,
+                mb_albumid="album",
+                mb_releasegroupid="release_group_in_lib",
+            )
+        )
         requests_mock.get(
-            f"/ws/2/release-group?artist={self.album_in_lib.mb_albumartistid}",
+            f"/ws/2/release-group?artist={artist_mbid}",
```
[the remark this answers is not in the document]
bgrassy: Updated this (and the below)

docs/plugins/missing.rst
```diff
@@ -27,12 +27,18 @@ library:
     -c, --count           count missing tracks per album
     -t, --total           count totals across the entire library
     -a, --album           show missing albums for artist instead of tracks for album
+    --release-type        show only missing albums of specified release type.
+                          You can provide this argument multiple times to
+                          specify multiple release types to filter to. If not
+                          provided, defaults to just the "album" release type.
```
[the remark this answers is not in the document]
bgrassy: Done!

beetsplug/missing.py
bgrassy: Just checked this- `type=""` returns the same results as not passing type, so I left it as is. One question though- this does change default behavior from running no type filter to filtering just to `--release-type album`. Should I instead set the default to empty in the config?

beetsplug/missing.py
bgrassy: I changed the default on line 229 (still have the question linked to below surrounding if we actually want this default to be `["album"]`)

beetsplug/missing.py
snejus: Is it still possible to override this with `missing: release_types: []` using the config?

snejus: By the way, rename the flag to `--release-types` as well!

bgrassy: Renamed (and changed the flag to support comma separated multiple inputs). One of the tests checks that `release_types: []` in the config passes an empty `type` string to the musicbrainz request, so that should match that behavior.

beetsplug/missing.py
snejus: Makes sense!

snejus (approved):

PR #5589, opened 2025-01-12
Fix matcher typo

valrus:
## Description

Fixes a confusing typo when setting the MusicBrainz matcher in a few tests. It looks like the matcher defaults to `IDENT` so change it to that, since that would have been the value used in the tests with the typo.

- [x] Documentation: N/A
- [x] Changelog. N/A, I think
- [x] Tests. N/A, I think


snejus (approved):
Looks good. Just rebase off updated `master` and I'll merge this.

PR #5606, opened 2025-01-27
Multi genres / only "genres" field wip

snejus:

PR #5609, opened 2025-01-31
Fix musicbrainz genres fetching

JOJ0:
- genres are now called tags
- tags needs to be in "mb fetch includes"
- release-group has them
- release has them
- and recording as well but we don't use them
- not sure what this outdated check was doing, it looked at "recordings" in the valid list even though genre fetching is only included in the `album_info` code and not `track_info` 🤷 
- see musicbrainz.VALID_INCLUDES for reference

## Description

Fixes #X.  

(...)

## To Do


JOJ0:
While looking at https://github.com/beetbox/beets/pull/5606/ I realized that genres fetching from musicbrainz is actually broken since.....I don't know :-/

beets/autotag/mb.py
```diff
@@ -104,8 +104,11 @@ def get_message(self):
 TRACK_INCLUDES = ["artists", "aliases", "isrcs"]
 if "work-level-rels" in musicbrainzngs.VALID_INCLUDES["recording"]:
     TRACK_INCLUDES += ["work-level-rels", "artist-rels"]
-if "genres" in musicbrainzngs.VALID_INCLUDES["recording"]:
-    RELEASE_INCLUDES += ["genres"]
+if "tags" in [
+    musicbrainzngs.VALID_INCLUDES["release"],
```
snejus, line 108: I have a feeling this line doesn't behave as intended. Did you consider adding tests?

JOJ0: Nope actually I didn't, just trying to quickly fix what was broken and on to more important things :-)

Actually I'm rather wondering why these kinds of checks are of any use here. What is there purpose? Why is it useful to check if something is valid if it is allowed to get silently broken anyway? Nobody will notice when this check is false the outcome just is: "there are no genres for this album". No error no nothing. Useless in my opinion. 

This is by the way the valid_includes constants dict:

https://github.com/alastair/python-musicbrainzngs/blob/1638c6271e0beb9560243c2123d354461ec9f842/musicbrainzngs/musicbrainz.py#L36-L84

snejus: The reason for my comment was that this check seems like:

```python
if "tags" in [
    ["artists", "labels", "recordings", ...],
    ["artists", "releases", "discids", ...],
]:
    RELEASE_INCLUDES += ["tags"]
```

i.e. this won't possibly evaluate to `True` because the list contains two __lists__ of tags. I think you intended to unpack them?

snejus: > Why is it useful to check if something is valid if it is allowed to get silently broken anyway?

I think the checks are here to prevent a more serious issue: if musicbrainzngs is asked to include a field it doesn't support it will raise an exception. We can ensure that we only ask for supported fields in a slightly more straightforward way without using a conditional:

```python
RELEASE_INCLUDES = {
    "artists",
    "media",
    "recordings",
    "release-groups",
    "labels",
    "artist-credits",
    "aliases",
    "recording-level-rels",
    "work-rels",
    "work-level-rels",
    "artist-rels",
    "isrcs",
    "url-rels",
    "release-rels",
    "tags",
} & set(musicbrainzngs.VALID_INCLUDES["release"])
```

JOJ0: > The reason for my comment was that this check seems like:
> 
> ```python
> if "tags" in [
>     ["artists", "labels", "recordings", ...],
>     ["artists", "releases", "discids", ...],
> ]:
>     RELEASE_INCLUDES += ["tags"]
> ```
> 
> i.e. this won't possibly evaluate to `True` because the list contains two **lists** of tags. I think you intended to unpack them?

Yes I understand, I was trying to look into a list in a list, kind of, that can't work. Ok.

But let's leave this aside for a moment and bear with me: I'm thinking of removing this check entirely, since it would give us the option to see if something changes and breaks - a slient bug like this could not have happened, someboday would have reported it earlier!

```
$ beet import -L dom roland
Error: "Bad includes: genres is not a valid include" in get release by ID with query '9506bb13-5ef1-42c9-bd53-f6c9387638c8'
Error: "Bad includes: genres is not a valid include" in get release by ID with query '9506bb13-5ef1-42c9-bd53-f6c9387638c8'
```

musicbrainzngs fails gracefully and telling us pretty detailed what the problem is. I would very much prefer a behaviour like that, than trying to reimplement what we had before - which tries to be smart about "not failing" but actually only leads to "silently ignoring" what better should fail and warn.


The second thought I always have when I see these top-level musicbrainz-requests-include-things, and I had that feeling and tried to implement something already back when I implemented something else here 1 year ago: Wouldn't it be much more useful if we somehow check if the user actually *wants* genres (config check) and only _then_ we ask musicbrainz to fetch those for us (add "tags" to the module level constant)

I'm not sure where to put such a check

snejus: I see your point regarding how quickly would we find out that this is broken.

On the other hand, beets importing functionality is now completely broken. Try looking at this from an average user's perspective: suddenly beets importing functionality is broken, and I cannot do anything about it since this didn't happen because of my bad configuration: the issue is in the source code.

A minority of users will consider creating an issue on GitHub, hoping that this will get addressed soon. Given the number of maintainers available for this nowadays, how long would it take to get this issue fixed? By the time it gets fixed some users will be already be gone, since they had to search for another tool that does the job and found it.

I agree we would find out about this issue soonish, however I think I'd prioritize reliability and user satisfaction instead 😅

JOJ0: Have a look at my two latest commits. Sanity checks for both release and track level fetches.

JOJ0:
What can we do about this @snejus ?

Screenshot 2025-02-04 at 08 14 02

snejus:
You want to use `{ ... }` to define the fields as a set. Note that musicbrainzngs only accept a `list` of fields there, so you'll need to cast it:

```python
RELEASE_INCLUDES = list(
    {
        "artists",
        "media",
        "recordings",
        "release-groups",
        "labels",
        "artist-credits",
        "aliases",
        "recording-level-rels",
        "work-rels",
        "work-level-rels",
        "artist-rels",
        "isrcs",
        "url-rels",
        "tags",
    }
    & set(musicbrainzngs.VALID_INCLUDES["release"])
)
```

snejus:
Nice, thank you. One final question: how does this play with the `musicbrainz.genres` configuration option?

JOJ0:
> Nice, thank you. One final question: how does this play with the `musicbrainz.genres` configuration option?

The option is checked for its state here:

https://github.com/beetbox/beets/pull/5609/files#diff-ff51bca3f15a32bf8652a69e127b582652187bae2fd2fd7a1ccb58d0feeb97b2L608-L612

But you might be referring to the fact that genres/tags are fetched in any case. Yes that is true.

I recall foggily that once I added `url-rels` to the includes-list  I initially submitted a config check there but Adrian advised that it might not be a good idea to check for config settings at the module level. Found the comment: https://github.com/beetbox/beets/pull/4474#discussion_r986066535

Or were you referring to something else @snejus ?

snejus:
Oh yeah this sounds good - I was just interested whether we actually performs that configuration check!

snejus (approved):
Nice, thanks for this

JOJ0:
Thanks for reviewing! Another step in the beets genres bucket list ;-)

JOJ0:
whoops I wanted to add a changelog. 🤣😅 I'll do it in master 😉

PR #5618, opened 2025-02-08
Fix lyrics tests

snejus:
Adjust lyrics integration tests to make sure that expected lyrics from `musica` match what's actually returned.


snejus:
It's a tiny test change so I'm merging it in, to unblock #5611

PR #5636, opened 2025-02-16
`mbsync` / `missing` plugins: consider all metadata backends

snejus:
This PR adds support for alternate metadata backends in the `mbsync` and `missing` plugins, allowing them to work with any metadata backend the items are tagged with instead of being limited to just MusicBrainz.

Key changes:
- Removed redundant `albums_for_id` and `items_for_id` methods.
- Updated `album_for_id` and `track_for_id` implementations to return a single album/track as their names suggest.
- Updated both plugins to use the above functions instead of MusicBrainz-specific ones.
- Updated `missing` plugin docs to make it clear that missing albums for artists can only be obtained for the `musicbrainz` metadata backend
  - This is possible with other backends but implementation was outside of this PR's scope.
  - Slightly simplified the existing implementation

The changes maintain backwards compatibility while enabling users to leverage metadata from sources like Deezer, Discogs or Bandcamp when syncing their libraries.

**Note**: this change came about because I'm working on making `musicbrainz` a separate plugin: this will be submitted in the next PR.

### DIFF
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -602,8 +602,7 @@ def album_for_mbid(release_id: str) -> AlbumInfo | None:
     if the ID is not found.
     """
     try:
-        album = mb.album_for_id(release_id)
-        if album:
+        if album := mb.album_for_id(release_id):
             plugins.send("albuminfo_received", info=album)
         return album
     except mb.MusicBrainzAPIError as exc:
@@ -616,8 +615,7 @@ def track_for_mbid(recording_id: str) -> TrackInfo | None:
     if the ID is not found.
     """
     try:
-        track = mb.track_for_id(recording_id)
-        if track:
+        if track := mb.track_for_id(recording_id):
             plugins.send("trackinfo_received", info=track)
         return track
     except mb.MusicBrainzAPIError as exc:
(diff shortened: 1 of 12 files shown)


wisp3rwind:
While I won't have time to review this in depth anytime soon, here's one thing that's unclear to me at a superficial glance:

> - Removed redundant albums_for_id and items_for_id methods.
> - Updated album_for_id and track_for_id implementations to return a single album/track as their names suggest.

Looking just at the changed code, it appears that `albums_for_id` could previously indeed return several candidates given a single `search_id`, whereas now only the first one is used (i.e. musicbrainz if available, else the first plugin match).

Maybe this makes sense (possibly, a given `search_id` is only ever expected to return _a single result_ for _a single service_?), but I feel like it would warrant a code comment explaining why we potentially drop additional results.

snejus:
> Maybe this makes sense (possibly, a given search_id is only ever expected to return a single result for a single service?), but I feel like it would warrant a code comment explaining why we potentially drop additional results.

That's exactly my thinking - I have worked with all available metadata sources, and I'm yet to discover a case where a single ID yields more than one album / singleton. I will add a comment under these methods.

JOJ0:
Played around a little with mbsync Discogs and bandcamp. The latter keeps showing this. Not sure what this field is:

```
$ beet mbsync alves
Alves - Episode 05 - 01 - Dark Era Of Amnesia (Resurrection) (2016)  [] deep house, experimental, industrial, techno
  track_alt:
Alves - Episode 05 - 02 - Black Mamba (2016)  [] deep house, experimental, industrial, techno
  track_alt:
Alves - Episode 05 - 03 - Viper (2016)  [] deep house, experimental, industrial, techno
  track_alt:
Alves - Episode 05 - 04 - Problem Magnet (2016)  [] deep house, experimental, industrial, techno
  track_alt:
Alves - Episode 05 - 05 - ADHD (2016)  [] deep house, experimental, industrial, techno
  track_alt:
Alves - Episode 05 - 06 - The Chase (2016)  [] deep house, experimental, industrial, techno
  track_alt:
  ```

JOJ0:
> Comments suppressed due to low confidence (1)
> **beetsplug/mbsync.py:86**
> 
> * The log message uses the same placeholder twice. It likely should display different values (for example, one for the recording ID and one for additional context such as the track title).
> 
> ```
>             self._log.info("Recording ID not found: {0} for track {0}", item.mb_trackid)
> ```

I can't find it in the github diff. Did you even touch it or is copilot checking unchanged things too? Anyway, you could fix it.

Other than that if you are happy so far I think this could be merged. What do you think @snejus ? (and yes the track_alt thing, not sure about that)

JOJ0 (approved):
Also a ton of simplifications in this one! Thanks for that and certainly a very handy new feature!!! The `track_alt` thing is minor, let's merge it and see what other things come up with metadata backends. We can't test them all anyway.

snejus:
> > Comments suppressed due to low confidence (1)
> > **beetsplug/mbsync.py:86**
> > 
> > * The log message uses the same placeholder twice. It likely should display different values (for example, one for the recording ID and one for additional context such as the track title).
> > 
> > ```
> >             self._log.info("Recording ID not found: {0} for track {0}", item.mb_trackid)
> > ```
> 
> I can't find it in the github diff. Did you even touch it or is copilot checking unchanged things too? Anyway, you could fix it.
> 
> Other than that if you are happy so far I think this could be merged. What do you think @snejus ? (and yes the track_alt thing, not sure about that)

Fixed the logging issue.

Regarding the `track_alt` field, my guess is that this is due to autotaggers returning an empty string value `""` for this field, which probably doesn't get handled appropriately given lack of arguments (I guess) `bpsync` provides to `ui.show_item_changes`. 

I spent an hour trying to understand this now and I'm more confused than I was previously, so I don't have a clue to be honest 🤣

PR #5641, opened 2025-02-21
Allow to pickle db models by removing the current connection.

semohr:
## Description

This might be a quick one, depending on how you feel about it... It allows you to pickle DB model objects. I don't think this is used directly in Beets, but it might be useful in general. For instance, we encountered an issue where we wanted to quickly pickle an Item or Album. This sometimes worked and other times failed, which seemed quite inconsistent.

Some DB model methods and properties have the side effect of attaching an SQLite connection to self (._db), which prevents serialization. The fix is quite straightforward, so I thought we might want to integrate this into beets directly.

## To Do

- [x] Changelog
- [x] Tests

### DIFF
--- a/test/test_library.py
+++ b/test/test_library.py
@@ -1394,16 +1394,3 @@ class LibraryFieldTypesTest(unittest.TestCase):
         beets.config["format_raw_length"] = True
         assert 61.23 == t.format(61.23)
         assert 3601.23 == t.format(3601.23)
-
-
-class PickleTest(ItemInDBTestCase):
-    def test_pickle_dump(self):
-        """Tries to pickle an item. This tests the __getstate__ method
-        of the Model ABC"""
-        import pickle
-
-        item = self.i
-        # Fetch album
-        item._cached_album
-
-        pickle.dumps(item)
--- a/test/test_library.py
+++ b/test/test_library.py
@@ -1394,3 +1394,16 @@ class LibraryFieldTypesTest(unittest.TestCase):
         beets.config["format_raw_length"] = True
         assert 61.23 == t.format(61.23)
         assert 3601.23 == t.format(3601.23)
+
+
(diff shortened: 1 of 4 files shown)


wisp3rwind:
Not sure how I feel about this, given that beets doesn't require it directly. Of how much help would it be for a project that uses beets as a library to have this upstream?

Note that a related issue led to the implementation of `Model.copy`: https://github.com/beetbox/beets/pull/2668

test/test_library.py
```diff
@@ -1394,3 +1394,16 @@ def test_durationtype(self):
...
+
+class PickleTest(ItemInDBTestCase):
+    def test_pickle_dump(self):
+        """Tries to pickle an item. This tests the __getstate__ method
+        of the Model ABC"""
+        import pickle
+
+        item = self.i
+        # Fetch album
+        item._cached_album
```
wisp3rwind, line 1407: What's the point of this exactly? Is it an indirect way to ensure that the `item` is in fact associated to a database?

This should probably just be
```suggestion
        item.store()
```
which would raise a `ValueError` if there's no database. You could also consider adding an explicit `assert item._db is not None`.

wisp3rwind (changes requested):
Fair enough, I'd say this can be merged. Whether or not pickling (rather than actually consistently persisting import state to the db, maybe with a flexible attribute saying that it's incomplete) is the best solution to your original problem might be a different issue. Having pickle support seems generally in line with the vision that the `dbcore` module should be as generic and reusable as possible, rather than being specific to beets data model.

I think the test should go to `test/dbcore.py`, however. Would you mind moving it there?

semohr:
We are actually storing the pickled value in a database :grin: It is quite hacky but any other solution would come with a pile of other issues which tbh wasn't worth the time for us right now.

wisp3rwind (approved):
LGTM :+1:

PR #5651, opened 2025-02-27
fix: plugins/listenbrainz: Fix UnboundLocalError in cases where 'mbid' is not defined

allendema:
## Description

Fix ocurrence of `UnboundLocalError` in plugins/listenbrainz > `get_tracks_from_listens()` when `mbid` is not available.
Removed a print statment.
Fix link to config.yaml.
Fix link to Listenbrainz "get the token" documentation.

## To Do


snejus:
Please add a PR description to link this to the issue and it should be good to go.

PR #5654, opened 2025-03-03
Fix import_task_before_choice event for singleton imports

marksdestiny:
## Description

This PR fixes a bug which caused the `import_task_before_choice` event not to be sent for singleton imports.

I suppose this was just an oversight, as this behavior is not mentioned in the [documentation](https://beets.readthedocs.io/en/stable/dev/plugins.html#listen-for-events).

## To Do

### DIFF
--- a/beets/ui/commands.py
+++ b/beets/ui/commands.py
@@ -1117,6 +1117,22 @@ class TerminalImportSession(importer.ImportSession):
         """
         print_()
         print_(displayable_path(task.item.path))
+
+        # Let plugins display info or prompt the user before we go through the
+        # process of selecting candidate.
+        results = plugins.send(
+            "import_task_before_choice", session=self, task=task
+        )
+        actions = [action for action in results if action]
+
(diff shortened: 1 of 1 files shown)


beets/ui/commands.py
```diff
@@ -1117,6 +1117,22 @@ def choose_item(self, task):
...
+        )
+        actions = [action for action in results if action]
+
+        if len(actions) == 1:
+            return actions[0]
+        elif len(actions) > 1:
+            raise plugins.PluginConflictError(
+                "Only one handler for `import_task_before_choice` may return "
+                "an action."
+            )
```
snejus, line 1134: Seems like the logic is exactly the same as inside `choose_match`. Would you mind deduping it?

snejus (changes requested):
Apologies for a late review - I think you will want to rebase this work

snejus:
Closing due to lack of author response.

PR #5657, opened 2025-03-05
feat: mpdstats: adds config option for remaining time threshold to determine if track was played.

5061726b6572:
## Description
Add new configuration option for mpdstats plugin, `played_ratio_threshold`, to allow configuring the percentage the song must be played for it to be counted as played instead of skipped.

## To Do

- [x] Documentation.
- [x] Changelog. 
- [x] Tests.

### DIFF
--- a/beetsplug/mpdstats.py
+++ b/beetsplug/mpdstats.py
@@ -27,6 +27,7 @@ from beets.util import displayable_path
 # much time should we wait between retries?
 RETRIES = 10
 RETRY_INTERVAL = 5
+DUPLICATE_PLAY_THRESHOLD = 10.0
 
 
 mpd_config = config["mpd"]
@@ -143,7 +144,9 @@ class MPDStats:
 
         self.do_rating = mpd_config["rating"].get(bool)
         self.rating_mix = mpd_config["rating_mix"].get(float)
-        self.time_threshold = 10.0  # TODO: maybe add config option?
+        self.played_ratio_threshold = mpd_config["played_ratio_threshold"].get(
+            float
+        )
 
         self.now_playing = None
         self.mpd = MPDClientWrapper(log)
@@ -216,10 +219,8 @@ class MPDStats:
 
         Returns whether the change was manual (skipped previous song or not)
         """
-        diff = abs(song["remaining"] - (time.time() - song["started"]))
-
-        skipped = diff >= self.time_threshold
-
+        elapsed = song["elapsed_at_start"] + (time.time() - song["started"])
+        skipped = elapsed / song["duration"] < self.played_ratio_threshold
         if skipped:
             self.handle_skipped(song)
         else:
@@ -256,13 +257,10 @@ class MPDStats:
 
     def on_play(self, status):
         path, songid = self.mpd.currentsong()
-
         if not path:
             return
 
         played, duration = map(int, status["time"].split(":", 1))
-        remaining = duration - played
-
         if self.now_playing:
             if self.now_playing["path"] != path:
                 self.handle_song_change(self.now_playing)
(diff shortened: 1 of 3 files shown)


beetsplug/mpdstats.py
```diff
@@ -142,7 +142,9 @@ def __init__(self, lib, log):
 
         self.do_rating = mpd_config["rating"].get(bool)
         self.rating_mix = mpd_config["rating_mix"].get(float)
-        self.time_threshold = 10.0  # TODO: maybe add config option?
+        self.time_threshold = mpd_config[
+            "maximum_remaining_seconds_for_track_to_be_considered_played"
```
snejus: This must be the longest name for an option I've seen in this codebase! 😅 Instead of using an absolute value, I'd suggest using a ratio here and naming it **played_ratio_threshold**. 

This approach will make it suitable for longer tracks. For example, I have many sets in my library that are longer than one hour. I'd like to set this configuration to `0.5` to ensure that anything I listen to halfway through is counted as 'played', regardless of the duration.

beetsplug/mpdstats.py
```diff
@@ -272,7 +274,7 @@ def on_play(self, status):
                 # after natural song start.
                 diff = abs(time.time() - self.now_playing["started"])
 
-                if diff <= self.time_threshold:
+                if diff <= 10.0:
```
snejus: This should use the configured value, I think?

docs/plugins/mpdstats.rst
```diff
@@ -58,6 +58,8 @@ configuration file. The available options are:
   Default: ``yes``.
 - **rating_mix**: Tune the way rating is calculated (see below).
   Default: 0.75.
+- **maximum_remaining_seconds_for_track_to_be_considered_played**: Maximum number of seconds remaining in the track for it to be considered played. If remaining time is greater than this value skip count will be incremented instead of play count.
```
snejus: Try to fit this description within the line length limit of 80.

snejus (changes requested):
Finally! This is a well needed feature, thanks. I commented with a suggestion to use a ratio instead of an absolute value.

beetsplug/mpdstats.py
5061726b6572: Based on the comment above (lines 271-274) it seems this is for handling duplicate events from mpd so I left the value unchanged. I do not fully understand this or the subsequent conditional block, and since it seems to be working figure it's best to leave it alone.

beetsplug/mpdstats.py
5061726b6572: My first version is based on a ratio, but switched to this since it was less of a behavioral change. Not a big difference in code so I'll try to update the PR with the ratio version soon. Thanks for suggesting using a ratio and a config name; hated the sentence for a name but couldn't think of anything shorter that still had sufficient meaning.

beetsplug/mpdstats.py
snejus: I see. Maybe we could create a constant, such as

```py
DUPLICATE_PLAY_THRESHOLD = 10.0
```

to clarify this a bit more

docs/plugins/mpdstats.rst
5061726b6572: Thought it would be auto-formatted. Should be better now.

docs/changelog.rst
```diff
@@ -30,6 +30,9 @@ New features:
   :bug:`5829`
 * :doc:`plugins/mbcollection`: When getting the user collections, only consider
   collections of releases, and ignore collections of other entity types.
+* :doc:`plugins/mpdstats`: Add new configuration option,
+  ``played_ratio_threshold``, to allow configuring the percentage the song must
```
snejus, line 34: Make sure that the option is called the same way in the code and the docs - currently it is `ratio_played_threshold`

snejus (changes requested):

docs/changelog.rst
5061726b6572: My mind was telling me one the names didn't match but I was too tired the other day to not read what I was expecting. Thanks for catching it.

Issue #5648, opened 2025-02-26
UnicodeEncodeError: 'utf-8' codec can't encode character '\udce9' in position 14: surrogates not allowed


kalon33:
### Problem

Running this command in verbose (`-vv`) mode:

```sh
$ $ ./bin/beet -vv import -gC /media/nicolas/musique/
user configuration: /home/nicolas/.config/beets/config.yaml
data directory: /home/nicolas/.config/beets
plugin paths: 
lyrics: Disabling google source: no API key configured.
fetchart: google: Disabling art source due to missing key
fetchart: lastfm: Disabling art source due to missing key
Sending event: pluginload
library database: /media/nicolas/musique/musiclibrary.db
library directory: /media/nicolas/musique
Sending event: library_opened
Sending event: import_begin
state file could not be read: [Errno 2] No such file or directory: '/home/nicolas/.config/beets/state.pickle'
ignoring '.directory' due to ignore rule '.*'
ignoring '.folder.png' due to ignore rule '.*'
Traceback (most recent call last):
  File "/home/nicolas/beets/lib/python3.12/site-packages/beets/util/__init__.py", line 413, in bytestring_path
    return str_path.encode(_fsencoding())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeEncodeError: 'utf-8' codec can't encode character '\udce9' in position 14: surrogates not allowed

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/nicolas/beets/./bin/beet", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/nicolas/beets/lib/python3.12/site-packages/beets/ui/__init__.py", line 1870, in main
    _raw_main(args)
  File "/home/nicolas/beets/lib/python3.12/site-packages/beets/ui/__init__.py", line 1849, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "/home/nicolas/beets/lib/python3.12/site-packages/beets/ui/commands.py", line 1398, in import_func
    import_files(lib, paths, query)
  File "/home/nicolas/beets/lib/python3.12/site-packages/beets/ui/commands.py", line 1329, in import_files
    session.run()
  File "/home/nicolas/beets/lib/python3.12/site-packages/beets/importer.py", line 360, in run
    pl.run_parallel(QUEUE_SIZE)
[...]

PR #5662, opened 2025-03-10
Use os.fsencode to encode bytes

snejus:
# Remove Python 2 compatibility code for string encoding/decoding

This PR simplifies the codebase by removing Python 2 compatibility code related to string encoding and decoding operations. Key changes:

- Remove custom `_convert_args`, `arg_encoding()`, `_fsencoding()` and `convert_command_args()` functions
- Replace with standard library's `os.fsencode()` and `os.fsdecode()` for file system path handling
- Simplify bytestring/string conversions throughout the codebase
- Update test code to use modern string handling

These changes reduce code complexity and maintenance burden since Python 2 support is no longer needed.

### DIFF
--- a/test/plugins/test_convert.py
+++ b/test/plugins/test_convert.py
@@ -143,18 +143,13 @@ class ConvertCommand:
     in tests.
     """
 
-    def run_convert_path(self, path, *args):
+    def run_convert_path(self, item, *args):
         """Run the `convert` command on a given path."""
-        # The path is currently a filesystem bytestring. Convert it to
-        # an argument bytestring.
-        path = os.fsencode(os.fsdecode(path))
-
-        args = args + (b"path:" + path,)
-        return self.run_command("convert", *args)
+        return self.run_command("convert", *args, f"path:{item.filepath}")
 
     def run_convert(self, *args):
         """Run the `convert` command on `self.item`."""
-        return self.run_convert_path(self.item.path, *args)
+        return self.run_convert_path(self.item, *args)
 
 
 @_common.slow_test()
@@ -320,7 +315,7 @@ class NeverConvertLossyFilesTest(ConvertTestCase, ConvertCommand):
     def test_transcode_from_lossless(self):
         [item] = self.add_item_fixtures(ext="flac")
         with control_stdin("y"):
-            self.run_convert_path(item.path)
+            self.run_convert_path(item)
         converted = os.path.join(self.convert_dest, b"converted.mp3")
         self.assertFileTag(converted, "mp3")
 
@@ -328,14 +323,14 @@ class NeverConvertLossyFilesTest(ConvertTestCase, ConvertCommand):
         self.config["convert"]["never_convert_lossy_files"] = False
         [item] = self.add_item_fixtures(ext="ogg")
         with control_stdin("y"):
-            self.run_convert_path(item.path)
+            self.run_convert_path(item)
         converted = os.path.join(self.convert_dest, b"converted.mp3")
         self.assertFileTag(converted, "mp3")
 
     def test_transcode_from_lossy_prevented(self):
         [item] = self.add_item_fixtures(ext="ogg")
         with control_stdin("y"):
-            self.run_convert_path(item.path)
+            self.run_convert_path(item)
         converted = os.path.join(self.convert_dest, b"converted.ogg")
         self.assertNoFileTag(converted, "mp3")
 
--- a/test/plugins/test_convert.py
+++ b/test/plugins/test_convert.py
@@ -147,7 +147,7 @@ class ConvertCommand:
         """Run the `convert` command on a given path."""
         # The path is currently a filesystem bytestring. Convert it to
         # an argument bytestring.
-        path = path.decode(util._fsencoding()).encode(util.arg_encoding())
+        path = os.fsencode(os.fsdecode(path))
 
         args = args + (b"path:" + path,)
         return self.run_command("convert", *args)
--- a/test/plugins/test_thumbnails.py
+++ b/test/plugins/test_thumbnails.py
@@ -100,12 +100,10 @@ class ThumbnailsTest(BeetsTestCase):
 
     @patch("beetsplug.thumbnails.ThumbnailsPlugin._check_local_ok", Mock())
     @patch("beetsplug.thumbnails.ArtResizer")
-    @patch("beetsplug.thumbnails.util")
+    @patch("beets.util.syspath", Mock(side_effect=lambda x: x))
     @patch("beetsplug.thumbnails.os")
     @patch("beetsplug.thumbnails.shutil")
-    def test_make_cover_thumbnail(
-        self, mock_shutils, mock_os, mock_util, mock_artresizer
-    ):
+    def test_make_cover_thumbnail(self, mock_shutils, mock_os, mock_artresizer):
         thumbnail_dir = os.path.normpath(b"/thumbnail/dir")
         md5_file = os.path.join(thumbnail_dir, b"md5")
         path_to_art = os.path.normpath(b"/path/to/art")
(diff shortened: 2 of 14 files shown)


beets/autotag/hooks.py
```diff
@@ -283,8 +283,8 @@ def _string_dist_basic(str1: str, str2: str) -> float:
     """
     assert isinstance(str1, str)
     assert isinstance(str2, str)
-    str1 = as_string(unidecode(str1))
-    str2 = as_string(unidecode(str2))
+    str1 = unidecode(str1)
+    str2 = unidecode(str2)
```
wisp3rwind, line 287: Nice catch that this was a no-op these days!

beets/dbcore/query.py
```diff
@@ -216,7 +217,7 @@ def value_match(cls, pattern: P, value: Any):
         """Determine whether the value matches the pattern. The value
         may have any type.
         """
-        return cls.string_match(pattern, util.as_string(value))
+        return cls.string_match(pattern, os.fsdecode(str(value or "")))
```
wisp3rwind, line 220: I'm not sure about this, it definitely deviates from the previous behaviour:

- `str(b'abc') == "b'abc'"`
- Does it really make sense to involve `fsdecode` here? The `value` has no immediate relation to the filesystem, but is meant to match database fields (path queries are handled separately in `library.PathQuery`). Before, the `ignore` error handler was used: That way, there are definitely no decoding errors. Now, I'm not sure what could happen for decoding errors: Surrogates maybe? What would that mean for `string_match`.

wisp3rwind (commented):
I'm definitely out of my depth here; to review this properly, I'm lacking knowledge on the whole en/decoding business.

That said, I left a comment on one change that seemed unclear.

Superficially, this seems plausible as a fix to #5648, since that issue is probably caused by partial conversion of en/decoding usage.

FlorentLM:
Damn I was about to do this exact thing, glad I checked the issues first haha

Beets is as-is unusable for me with the current `util._fsencoding` function (surrogated characters in place of some swedish chars) - the parsing errors and I can't import the rest of the music.

Using `os.fsencode` works.

beets/dbcore/query.py
snejus: I just removed the last commit, keeping `as_string` in place :)

test/plugins/test_convert.py
```diff
@@ -147,9 +147,9 @@ def run_convert_path(self, path, *args):
         """Run the `convert` command on a given path."""
         # The path is currently a filesystem bytestring. Convert it to
         # an argument bytestring.
-        path = os.fsencode(os.fsdecode(path))
+        path = os.fsdecode(path)
 
-        args = args + (b"path:" + path,)
+        args = args + ("path:" + path,)
```
wisp3rwind, line 152: This should simplify to

```py
def run_convert_path(self, path, *args):
    """Run the `convert` command on a given path."""
    return self.run_command("convert", *args, b"path:" + path)
```

unless of course, we don't actually call it with `path: bytes` consistently. In that case, the `fsdecode` should stay.

test/plugins/test_thumbnails.py
```diff
@@ -100,12 +100,10 @@ def exists(path):
 
     @patch("beetsplug.thumbnails.ThumbnailsPlugin._check_local_ok", Mock())
     @patch("beetsplug.thumbnails.ArtResizer")
-    @patch("beetsplug.thumbnails.util")
+    @patch("beets.util.syspath", Mock(side_effect=lambda x: x))
```
wisp3rwind, line 103: Oh, wow, that's a test with a lot of mocks in place. I'm wondering whether that can't be done in a more robust way, relying less on the implementation. Anyway, that's for a different PR.

test/test_ui.py
```diff
@@ -865,6 +865,9 @@ def setUp(self):
 
         # Custom BEETSDIR
         self.beetsdir = os.path.join(self.temp_dir, b"beetsdir")
+        self.cli_config_path = os.path.join(
+            os.fsdecode(self.temp_dir), "config.yaml"
+        )
```
wisp3rwind, line 870:
```suggestion
        self.cli_config_path = os.path.join(self.temp_dir, b"config.yaml")
```

Let's stick to the convention of storing paths internally as `bytes` for now, until a change to a `pathlib` future is actually underway.

wisp3rwind (changes requested):
This much more focused version of the PR looks good to me now. Just a few nitpicks.

The reason that the first two commits should be fine is [PEP529](https://peps.python.org/pep-0529/) (had forgotten about that before), which implies that we shouldn't ever encounter `mbcs` again.

While I can't find this stated in the documentation, `subprocess` also applies `fsdecode` to argument lists, see `list2cmdline` in the subprocess source. Thus, the third commit should also be ok.

Regarding https://github.com/beetbox/beets/issues/5648, I think it would be good to verify again that it is actually fixed by this: It's not completely obvious to me what happens there, and why this should be a fix. (It might be due to the `surrogatepass` error handler being used by `fs{en,de}code`. It's still unclear to me what's the real origin of #5648.)

test/plugins/test_convert.py
snejus: Nice cleanup! Unfortunately `run_command` does not support `bytes` arguments, so this needs to be a string:

```python
    def run_convert_path(self, path, *args):
        """Run the `convert` command on a given path."""
        return self.run_command("convert", *args, f"path:{os.fsdecode(path)}")
```

test/plugins/test_thumbnails.py
snejus: Yeah, that's pretty much useless and overly fragile implementation tests that will break for any implementation adjustment 🤷🏼‍♂️

test/test_ui.py
snejus: I don't think this is a good idea, since `run_command` does not support `bytes`.

This is precisely why I removed `_convert_args` from test helpers which previously decoded `bytes`-based args to `str` for some tests:

image

There's no need for back-and-forth encoding and decoding when we just need to use strings.

snejus:
> There have been a bunch of changes in beets relatively recently: I'm not sure that it's consistently using any single type for paths these days (There used to be the very clear convention of handling everything as `bytes` internally. That probably wasn't 100% correctly implemented at any time, but at least there was a clear design decision.)

I agree that handling everything as bytes made sense when supporting Python 2, but as demonstrated in this PR, this approach introduced unnecessary complexity. For instance, tests attempting to run beets CLI commands with `bytes` arguments require decoding somewhere along the way. This seems counterintuitive, given that the `beets` CLI typically only receives arguments of `str` type.


> Right now, there seems to be more flexibility regarding `bytes`/`str`/`Path` than before, but that somehow evolved without a real strategy. (Except for the over-arching goal of eventually moving to pathlib completely: That's a worthwhile goal in my opinion, but also a large project.) 

It’s clear that trying to do both a full migration to pathlib and maintain strict consistency across the whole codebase (using only bytes or only Path) all at once would be really challenging. So, I’m happy with the gradual progress as long as we keep that bigger picture in mind.

I've just grepped the codebase for Path usage, and I'm somewhat surprised to see that it's present in just a few modules - I thought it was more widely adopted. The codebase is still overwhelmingly using paths based on `bytes`.

test/plugins/test_convert.py
wisp3rwind: Ah, sorry, must have mixed it up. Thanks for fixing up my suggestion :)

test/test_ui.py
wisp3rwind: Makes sense. I was thinking too much in terms of implementation rather than testing code: In `/beets/...` I'd argue that my point on sticking to `bytes` for now stands, but for testing code as it is here, that might not be the best choice.

wisp3rwind (approved):

Issue #5664, opened 2025-03-11
Tarfile import fails with extraction failed: 'TarFile' object has no attribute 'infolist'


bcotton:
Seems `TarFile.infolist` was removed in Python 3: https://github.com/python/cpython/issues/47289#issuecomment-1093424447 when `ZipFileCompat` was removed from `TarFile`

introduced in https://github.com/beetbox/beets/pull/4396

### Problem

Running this command in verbose (`-vv`) mode:

```sh
beet -vv import -m Elton\ John\ -\ Your\ Songs\ \(1985\)\ \(flac\).tar-xpost/Elton_John_-_Your_Songs_\(1985\)_\(flac\).tar
```

Led to this problem:

```
user configuration: /home/bcotton/.config/beets/config.yaml
data directory: /home/bcotton/.config/beets
plugin paths:
library database: /media/music/musiclibrary.db
library directory: /media/music/curated
Sending event: library_opened
Sending event: import_begin
Extracting archive: /media/music/downloads/lidarr/Elton John - Your Songs (1985) (flac).tar-xpost/Elton_John_-_Your_Songs_(1985)_(flac).tar
extraction failed: 'TarFile' object has no attribute 'infolist'
No files imported from /media/music/downloads/lidarr/Elton John - Your Songs (1985) (flac).tar-xpost/Elton_John_-_Your_Songs_(1985)_(flac).tar
```

Here's a link to the music files that trigger the bug (if relevant):

### Setup

* OS: Linux
* Python version: Python 3.12.8
* beets version:  master as of 2025-03-11
* Turning off plugins made problem go away (yes/no):  no

My configuration (output of `beet config`) is:

```yaml
directory: /media/music/curated
id3extract:
    mappings:
        WOAS: spotify_track_id

import:
    copy: yes
    move: no
# --------------- Main ---------------

library: /media/music/musiclibrary.db
musicbrainz:
    enabled: yes
    external_ids:
        bandcamp: yes
        beatport: yes
        deezer: yes
        discogs: yes
        spotify: yes
        tidal: yes
        youtube: yes
    extra_tags: []
    genres: yes
    host: musicbrainz.org
    https: no
    ratelimit: 1
    ratelimit_interval: 1
    searchlimit: 5

# --------------- Plugins ---------------

plugins:
- spotify
- info
- fetchart
- lyrics
- lastgenre
- id3extract
spotify:
    mode: open
    region_filter: US
    show_failures: 'on'
    source_weight: 0
[...]

PR #5666, opened 2025-03-11
fix: TarFile import

bcotton:
## Description

Failing test for #5664. Proposed solution to failing test

## To Do

- [x] Fix the problem

### DIFF
--- a/beets/importer.py
+++ b/beets/importer.py
@@ -1225,7 +1225,33 @@ class ArchiveImportTask(SentinelImportTask):
             cls._handlers.append((is_zipfile, ZipFile))
             import tarfile
 
-            cls._handlers.append((tarfile.is_tarfile, tarfile.open))
+            class TarInfo(tarfile.TarInfo):
+                def __init__(self, member):
+                    super().__init__(member.name)
+                    self.name = member.name
+                    self.size = member.size
+                    self.mode = member.mode
+                    self.mtime = member.mtime
+                    self.type = member.type
+                    self.linkname = member.linkname
+                    self.uid = member.uid
+                    self.gid = member.gid
+                    self.uname = member.uname
+                    self.gname = member.gname
+                    self.date_time = time.gmtime(member.mtime)[:6]
+
+                @property
+                def filename(self):
+                    return self.name
+
+            class TarWrapper(tarfile.TarFile):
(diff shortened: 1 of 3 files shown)


beets/importer.py
```diff
@@ -1225,7 +1225,33 @@ def handlers(cls):
             cls._handlers.append((is_zipfile, ZipFile))
             import tarfile
 
-            cls._handlers.append((tarfile.is_tarfile, tarfile.open))
+            class TarInfo(tarfile.TarInfo):
```
bcotton, line 1228: I am not a python dev. Not sure this is correct, but is working on my install.

beets/importer.py
```diff
@@ -1225,7 +1225,33 @@ def handlers(cls):
...
+                    self.size = member.size
+                    self.mode = member.mode
+                    self.mtime = member.mtime
+                    self.type = member.type
+                    self.linkname = member.linkname
+                    self.uid = member.uid
+                    self.gid = member.gid
+                    self.uname = member.uname
+                    self.gname = member.gname
+                    self.date_time = time.gmtime(member.mtime)[:6]
```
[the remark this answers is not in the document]
bcotton: Seems to be passing without it. 🤷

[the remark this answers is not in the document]
wisp3rwind: That's because the import is actually there; maybe copilot is missing that due to the nested class definition?

beets/importer.py
wisp3rwind: The idea right correct; however I think the definition of `TarInfo` and `TarWrapper` should be moved out of `set_choice` directly to the top-level of this file.

beets/importer.py
wisp3rwind, line 1241: I think this could be stripped down to just the bare minimum that we need here: Thus, ` TarInfo` can be a new class (not wrap the original `TarInfo`), and only set `date_time` and `filename`.

wisp3rwind (changes requested):
Thanks for tackling this! The overall approach of wrapping the standard library seems right to me, but I think there's some refactoring that should be done to clean up the code a little: See the inline comments.

snejus:
@bcotton are you planning to come back to this?

snejus:

PR #5684, opened 2025-03-23
FIX: Dereference symlinks before hardlinking

### LINKED ISSUE #5676: convert:  Relative symlinks broken upon convert with -H
When converting with the hardlink option, existing relative symlinks in the original beets library will be converted to broken symlinks in the destination.

### Problem

With the configuration:

```yaml
directory: ~/BugDemo
library: ~/BugDemo/beets.db
plugins: convert
convert:
```

Run the following commands:

```bash
# Setting up the initial library
cd ~
mkdir -p BugDemo/MyArtist/MyAlbum
ln -s ../../../MyTrack.ogg BugDemo/MyArtist/MyAlbum/ # This should be a real, lossy music file, and a relative symlink

tree BugDemo/
# BugDemo/
# └── MyArtist
#     └── MyAlbum
#         └── MyTrack.ogg -> /path/to/MyTrack.ogg

beet import -CW BugDemo/ # metadata is irrelevant

# Running the transcode
beet -vv convert -Hyd ./path/to/Destination/

# user configuration: /home/myuser/.config/beets/config.yaml
# data directory: /home/myuser/.config/beets
# plugin paths: 
# Sending event: pluginload
# library database: /home/myuser/BugDemo/beets.db
# library directory: /home/myuser/BugDemo
# Sending event: library_opened
# Parsed query: AndQuery([TrueQuery()])
# Parsed sort: NullSort()
# MyArtist - MyAlbum - MyTrack
# Parsed query: AndQuery([NoneQuery('album_id', True)])
# Parsed sort: NullSort()
# Parsed query: AndQuery([BooleanQuery('comp', 1, fast=True)])
# Parsed sort: NullSort()
# convert: Hardlinking /home/myuser/BugDemo/MyArtist/MyAlbum/01 MyTrack.ogg
# Sending event: write
# Sending event: after_write
# Sending event: after_convert
# Sending event: cli_exit

# The Result
ls -l ./path/to/Destination/MyArtist/MyAlbum/
# lrwxrwxrwx 2 myuser users 86 Mar 18 16:26 '01 MyTrack.ogg' -> '../../../MyTrack.ogg'
```

Notice that this symlink now points to nothing, as the original `MyTrack.ogg` would be at `../../../../../MyTrack.ogg`.  In other words, the relative path pointed to by the symlink was not updated to its new location.

The behavior that I *would* expect is either:
- Before creating the new symlink, convert the referent to an absolute path, so that the new symlink points to the same file as the original
- Or, even better, dereference the symlink and hardlink the referent to the new location

### Setup

* OS: NixOS 24.11
* Python version: 3.12.8
* beets version: 2.1.0 (from Nix package manager)
* Turning off (other) plugins made problem go away:  no

Alch-Emi:
When creating a hardlink, either during import or `beet convert`, if the origin of the hardlink was a symlink, that symlink used to be directly copied.  This could create broken symlinks if the origin symlink was relative, and in either case, probably wasn't the user's desired behavior.

This change de-references all symlinks before creating a hardlink, such that the end result is a normal file with the same inode as the original file.  See #5676 for more discussion about the original issue.

Fixes #5676

- [X] ~~Documentation~~ (N/A)
- [X] Tests. (Very much encouraged but not strictly required.)

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -605,7 +605,9 @@ def hardlink(path: bytes, dest: bytes, replace: bool = False):
     if os.path.exists(syspath(dest)) and not replace:
         raise FilesystemError("file exists", "rename", (path, dest))
     try:
-        os.link(syspath(path), syspath(dest))
+        # This step dereferences any symlinks and converts to an absolute path
+        resolved_origin = Path(syspath(path)).resolve()
+        os.link(resolved_origin, syspath(dest))
     except NotImplementedError:
         raise FilesystemError(
             "OS does not support hard links." "link",
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -29,6 +29,10 @@ Bug fixes:
 * :doc:`plugins/fetchart`: Fix fetchart bug where a tempfile could not be deleted due to never being
   properly closed.
   :bug:`5521`
+* When hardlinking from a symlink (e.g. importing a symlink with hardlinking
+  enabled), dereference the symlink then hardlink, rather than creating a new
+  (potentially broken) symlink
+  :bug:`5676`
 * :doc:`plugins/lyrics`: LRCLib will fallback to plain lyrics if synced lyrics
   are not found and `synced` flag is set to `yes`.
 * Synchronise files included in the source distribution with what we used to
--- a/test/test_files.py
+++ b/test/test_files.py
@@ -35,7 +35,8 @@ class MoveTest(BeetsTestCase):
         super().setUp()
 
         # make a temporary file
-        self.path = join(self.temp_dir, b"temp.mp3")
+        self.temp_music_file_name = b"temp.mp3"
+        self.path = join(self.temp_dir, self.temp_music_file_name)
         shutil.copy(
             syspath(join(_common.RSRC, b"full.mp3")),
             syspath(self.path),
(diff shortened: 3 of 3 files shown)


snejus:
@Alch-Emi conflicts need to be resolved here, I think

alch-emi:
Done!

alch-emi:
Merge conflicts re-resolved!  Please let me know if there are other changes or considerations that you believe should be addressed before this can be merged.

beets/util/__init__.py
```diff
@@ -581,7 +581,9 @@ def hardlink(path: bytes, dest: bytes, replace: bool = False):
     if os.path.exists(syspath(dest)) and not replace:
         raise FilesystemError("file exists", "rename", (path, dest))
     try:
-        os.link(syspath(path), syspath(dest))
+        # This step dereferences any symlinks and converts to an absolute path
+        resolved_origin = Path(syspath(path)).resolve()
```
semohr: I do not think the syspath call is needed if you use Pathlib `Path` here. Also instead of using os.link you should use `Path.hardlink_to`.

If you are up for it I would love to get rid of the `os.*` calls in favor for the pathlib versions.

Ref:
https://docs.python.org/3/library/pathlib.html#corresponding-tools
https://beets.readthedocs.io/en/latest/dev/paths.html

alch-emi: Makes sense!  I implemented this in b405d2f, though there's still a call to `os.fsdecode` that I couldn't figure out how to get rid of.

Unfortunately, it seems like [`pathlib.Path.hardlink_to()`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.hardlink_to) was only added in Python 3.10, which caused the type check and tests for Python 3.9 to fail (see [this test log](https://github.com/beetbox/beets/actions/runs/19179901670/job/54833887027?pr=5684)).

I've force-pushed an update that leaves the `os.link` call in and includes the `Path.hardlink_to` call in a comment with a note to swap it out once Python 3.9 support is dropped.  Though if there's a different way you'd like to handle this, I'm happy to implement that instead.

semohr: > I've force-pushed an update that leaves the os.link call in and includes the Path.hardlink_to call in a comment with a note to swap it out once Python 3.9 support is dropped. Though if there's a different way you'd like to handle this, I'm happy to implement that instead.

Thank you! 

You can probably disregard the 3.9 workaround as we are [dropping the support for 3.9 any day](https://github.com/beetbox/beets/pull/6144). I don't think waiting until this is merged should be an issue for this PR ;)

alch-emi: Sounds good!  Force pushed back to the original commit!

alch-emi:
Changes to master from the 3.9 drop have been merged and tests are passing!

PR #5685, opened 2025-03-23
BUG:  Wrong path edited when running config -e

### LINKED ISSUE #5652: config -e edits the wrong file
I use multiple instances of beet to organize different kinds of music. The configurations are saved in the music directory below music_x/.beets . For simplicity, I linked the default music directory .beets to .config/beets:

```sh
$ ls -al .config/beets
lrwxrwxrwx 1 user user 24 Feb 23 21:30 .config/beets -> /path/to/music_x/.beets/
```

### Problem

```sh
$ beet -c /path/to/music_y/.beets/config.yaml config -e
```

Led to this problem: The configuration file of the default beet installation is opened. While executing simply

```sh
$ beet -c /path/to/music_y/.beets/config.yaml config
```

shows the correct configuration output.

### Setup

* OS: Ubuntu 24.04 LTS
* Python version: python 3.12.3
* beets version: 2.2.0 - installed via pip

alch-emi:
As per #5652, `beet --config <path> config -e` edited the default config path, even though that's not the config that would be used by beets.

It seems like this was the result of a deliberate short-circuit in [`_raw_main()`](https://github.com/Alch-Emi/beets/blob/c2de6feada5c84d8bd0235969d551b6a4fce1277/beets/ui/__init__.py#L1832).  The short-circuit prevents malformed configs from causing a crash before opening the editor, but also prevents the setup function from loading the custom config at all.

The solution used here is to just expose the CLI options to `edit_config()`, so that it can use the custom config path if its set.  I also suspect that the branch in [`config_func()`](https://github.com/Alch-Emi/beets/blob/c2de6feada5c84d8bd0235969d551b6a4fce1277/beets/ui/commands.py#L2354) which is getting short circuited is actually unreachable, but I left it in with a note just in case.

Fixes #5652

## To Do

- [x] ~~Documentation~~ (N/A)
- [X] Changelog
- [X] Tests

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -958,7 +958,7 @@ class LyricsPlugin(RequestHandler, plugins.BeetsPlugin):
 
     @cached_property
     def backends(self) -> list[Backend]:
-        user_sources = self.config["sources"].get()
+        user_sources = self.config["sources"].as_str_seq()
 
         chosen = sanitize_choices(user_sources, self.BACKEND_BY_NAME)
         if "google" in chosen and not self.config["google_API_key"].get():
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -31,6 +31,8 @@ Bug fixes:
   calls and rate limit exhaustion.
 - Running `beet --config <mypath> config -e` now edits `<mypath>` rather than
   the default config path. :bug:`5652`
+- :doc:`plugins/lyrics`: Accepts strings for lyrics sources (previously only
+  accepted a list of strings). :bug:`5962`
 
 For plugin developers:
 


Alch-Emi:
Oops, I thought I checked that.  (force-push to fix formatting)

test/test_config_command.py
```diff
@@ -128,3 +128,15 @@ def test_edit_invalid_config_file(self):
...
+        alt_config_path = os.path.join(
+            self.temp_dir.decode(), "alt_config.yaml"
+        )
+        with open(self.config_path, "w") as file:
+            file.write("option: alt value\n")
+
+        os.environ["EDITOR"] = "myeditor"
+        with patch("os.execlp") as execlp:
+            self.run_command("--config", alt_config_path, "config", "-e")
+        execlp.assert_called_once_with("myeditor", "myeditor", alt_config_path)
```
wisp3rwind, line 142: We already have a secondary config file around here, cf. the `test_show_user_config_with_cli` test: Can you change this test to use `self.cli_config_path` instead of generating `alt_config_path`?

wisp3rwind (changes requested):
Thanks, this looks like the correct fix :+1:

I think the test can be simplified a bit by using existing infrastructure please see the inline comment.

test/test_config_command.py
alch-emi: Done!  Sorry that took so long i 100% forgot about this PR

beets/ui/commands.py
```diff
@@ -2322,7 +2322,10 @@ def config_func(lib, opts, args):
 
     # Open in editor.
     elif opts.edit:
-        config_edit()
+        # Note:  This branch *should* be unreachable
+        # since the normal flow should be short-circuited
+        # by the special case in ui._raw_main
+        config_edit(opts)
```
henry-oberholtzer: This does seem like it'll be unreachable, but I can't think of a situation where it'll reach this. Although there may be an edgecase I haven't thought of. (two subcommands maybe? -- I'll need to check if the parser already handles that.) I think it's fine to be here though.

henry-oberholtzer (approved):

alch-emi:
Just pushed a merge to fix the conflicts that cropped up following the refactoring of `commands.py`.  However, I'd like to (re)request a review from @wisp3rwind or someone with merge access so that I don't have to keep doing conflict resolution.

henry-oberholtzer:
Hi! All looks good, I've merged this in!

PR #5692, opened 2025-03-26
Feature: Add progress bars

peterjdolan:
## Description

Many long-running commands produce little or no feedback in the terminal to indicate that they're progressing, and none of them provide estimates of how long the operation will run. This change introduces the `enlighten` python package, which displays progress bars akin to TQDM below the existing terminal output. This PR does not modify the import workflow, rather only modifies several built-in plugins. Modifying the import workflow to display a progress bar will be done in a follow-up PR.

See [tracking issue](https://github.com/beetbox/beets/issues/5711)

## To Do

## Example screenshot

Screenshot 2025-04-10 103457


peterjdolan:
Import operation partially implemented; still needs some work. Example screenshot above.

peterjdolan:
@snejus I'm not sure why the latest push is failing the automatic lint and formatting; it passes 'poetry run poe lint && poetry run poe format' on my machine just fine. Any suggestions? I've already sync'd my fork to beets master and rebased.

wisp3rwind:
This seems to be related to some combination of (although I'm not seeing the full picture yet)
- you regenerated the lockfile, which now chooses a newer version of ruff
- https://github.com/astral-sh/ruff/blob/6bdffc3cbff2194afdff0e46fccb03cf6147d367/CHANGELOG.md?plain=1#L987
- https://github.com/astral-sh/ruff/pull/14511
- maybe, you're still running another ruff version locally

Someone should probably open a new PR to fix all TC006 warnings, include `TC` in our ruff rule selection instead of TCH, and probably fix ruff to a major version (a `0.X.*`). Since major ruff updates might contain breaking rule/style changes, I'd guess that we'd only want to bump it manually.

semohr:
Is there a way to disable the progress bar?

We’re [running](https://github.com/pSpitzner/beets-flask) in a containerized environment without a cli UI, where all `stdout` and `stderr` are redirected into log files. In similar cases, progress libraries (e.g. `tqdm`) tend to generate a lot of noisy log output, as each terminal update is written as a new line. I’m not sure how `enlighten` handles this and was not able to derive this from their docs.

Just wanted to mention this here to spare us some trouble down the line. Do you see any other problems regarding this use case?

jackwilsdon:
It looks like this needs to replicate the check `enlighten.get_manager` does - disabling the manager if the output is not a TTY.

You might be best actually just using `enlighten.get_manager` and passing enabled to that, as it will only enable the manager if the provided enable value was true and the stream is a TTY.

peterjdolan:
> It looks like this needs to replicate the check `enlighten.get_manager` does - disabling the manager if the output is not a TTY.
> 
> You might be best actually just using `enlighten.get_manager` and passing enabled to that, as it will only enable the manager if the provided enable value was true and the stream is a TTY.

Great points, thank you. I'll add that condition as well.

peterjdolan:
This is now ready for review. I'm not certain how to request a review, though. Anybody have pointers?

beetsplug/badfiles.py
```diff
@@ -200,8 +202,19 @@ def command(self, lib, opts, args):
...
+            ) as (_, n_unchanged, n_errors):
+                success, error_lines = self.check_item(item)
+                if success:
+                    n_unchanged.update()
+                else:
+                    n_errors.update()
+                    for line in error_lines:
+                        ui.print_(line)
 
         par_map(check_and_print, items)
```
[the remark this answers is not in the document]
peterjdolan: Resolved otherwise.

beetsplug/replaygain.py
```diff
@@ -1511,21 +1511,26 @@ def command_func(
...
                     f"Analyzing {len(items)} tracks ~"
                     f" {self.backend_name} backend..."
                 )
-                for item in items:
+                for item in ui.iprogress_bar(
+                    items,
+                    desc="Analyzing tracks",
+                    unit="tracks",
+                ):
                     self.handle_track(item, write, force)
```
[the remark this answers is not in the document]
peterjdolan: Resolved manually.

docs/changelog.rst
```diff
@@ -47,6 +51,48 @@ Other changes:
...
+
+2.4.0 (September 13, 2025)
+--------------------------
+
+New features:
+
+- Added progress bars for most commands, showing how many albums or songs have
+  been changed (white), how many were unchanged (blue), and how many errors
+  there were (red). The progress bar includes how much time has passed and an
+  estimate of how much time is left. :bug:`5711`
```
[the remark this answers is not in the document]
peterjdolan: Added progress bar docs, and updated changelog.

beetsplug/lastgenre/__init__.py
```diff
@@ -532,11 +532,15 @@ def lastgenre_func(lib, opts, args):
...
+                for album in ui.iprogress_bar(
+                    lib.albums(args),
+                    desc="Fetching genres",
+                    unit="albums",
+                ):
                     album_genre, src = self._get_genre(album)
                     prefix = "Pretend: " if pretend else ""
                     self._log.info(
-                        '{}genre for album "{.album}" ({}): {}',
+                        '{}genre for album "{.album}" ({}): {.genre}',
```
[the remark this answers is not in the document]
peterjdolan: Resolved manually.

beetsplug/badfiles.py
```diff
@@ -55,12 +64,12 @@ def __init__(self):
...
         )
 
-    def run_command(self, cmd):
+    def run_command(self, cmd: list[str]) -> CheckResult:
         self._log.debug(
             "running command: {}", displayable_path(list2cmdline(cmd))
         )
         try:
-            output = check_output(cmd, stderr=STDOUT)
+            output: bytes = check_output(cmd, stderr=STDOUT)
```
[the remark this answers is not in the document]
peterjdolan: Thank you, but I'm going to leave this for future work, as I'm not changing the logical flow in this PR.

beets/ui/__init__.py
```diff
@@ -1325,6 +1344,80 @@ def add_all_common_options(self):
...
+U = TypeVar("U")
+
+
+class SizedIterable(Protocol[T], Sized, Iterable[T]):
+    pass
+
+
+def iprogress_bar(
+    sequence: Union[Iterable[U], SizedIterable[U]], **kwargs
+) -> Generator[U, None, None]:
```
snejus:
```suggestion
) -> Iterator[U]:
```

beets/ui/__init__.py
```diff
@@ -1325,6 +1344,80 @@ def add_all_common_options(self):
         self.add_format_option()
 
 
+T = TypeVar("T", covariant=True)
```
snejus: Name it using the standard name for a covariant type variable.

```suggestion
T_co = TypeVar("T_co", covariant=True)
```

beets/ui/__init__.py
```diff
@@ -1325,6 +1344,80 @@ def add_all_common_options(self):
...
+
+def iprogress_bar(
+    sequence: Union[Iterable[U], SizedIterable[U]], **kwargs
+) -> Generator[U, None, None]:
+    """Construct and manage an `enlighten.Counter` progress bar while iterating.
+
+    Example usage:
+    ```
+    for album in ui.iprogress_bar(
+        lib.albums(), desc="Updating albums", unit="albums"):
```
snejus: Needs to be formatted the same way as it's used in the actual code

beets/ui/__init__.py
```diff
@@ -1325,6 +1344,80 @@ def add_all_common_options(self):
...
+        kwargs: Additional keyword arguments to pass to the `enlighten.Counter`
+            constructor.
+
+    Yields:
+        The items from the sequence.
+    """
+    if sequence is None:
+        log.error("sequence must not be None")
+        return
+
```
snejus: No need for this defensive check - the type of this variable clearly indicates that `None` is not accepted, so any attempt to supply `None` will be caught by `mypy` already.

beetsplug/badfiles.py
```diff
@@ -194,16 +208,27 @@ def on_import_task_before_choice(self, task, session):
...
-            for error_line in self.check_item(item):
-                ui.print_(error_line)
-
-        par_map(check_and_print, items)
+            success, error_lines = self.check_item(item)
+            if not success:
+                for line in error_lines:
+                    ui.print_(line)
+
+        with concurrent.futures.ThreadPoolExecutor() as executor:
```
snejus, line 207: Why are we not simply looping across the items, just like everywhere else?

beetsplug/badfiles.py
```diff
@@ -102,8 +111,9 @@ def get_checker(self, ext):
             return self.check_mp3val
         if ext == "flac":
             return self.check_flac
+        return None
 
-    def check_item(self, item):
+    def check_item(self, item: library.Item) -> tuple[bool, list[str]]:
```
snejus: Why did you add types to this plugin?

beetsplug/duplicates.py
```diff
@@ -284,9 +285,13 @@ def _group_by(self, objs, keys, strict):
...
         counts = collections.defaultdict(list)
-        for obj in objs:
+        for obj in ui.iprogress_bar(
+            objs,
+            desc="Finding duplicates",
+            unit="items",
+        ):
             values = [getattr(obj, k, None) for k in keys]
-            values = [v for v in values if v not in (None, "")]
+            values = list(filter(lambda v: v not in (None, ""), values))
```
snejus: No need for this change - the list comprehension is simpler to read and is less verbose.

beetsplug/missing.py
```diff
@@ -157,13 +156,13 @@ def _missing_tracks(self, lib, query):
...
-            if count:
+            for album in ui.iprogress_bar(
+                albums, desc="Analyzing albums", unit="albums"
+            ):
                 if _missing_count(album):
                     print_(format(album, fmt))
-
-            else:
+        else:
+            for album in albums:
```
snejus: I don't think I follow how does this conditional logic work. Could you explain?

beetsplug/parentwork.py
```diff
@@ -88,12 +88,18 @@ def func(lib, opts, args):
...
+                desc="Identifying parent works",
+                unit="songs",
+            ):
                 changed = self.find_work(item, force_parent, verbose=True)
                 if changed:
                     item.store()
-                    if write:
-                        item.try_write()
+
+                if write:
```
snejus: Why did you unindent this block? I doubt this is what we want to have.

beetsplug/permissions.py
```diff
@@ -31,10 +31,13 @@ def check_permissions(path, permission):
...
 
 
-def assert_permissions(path, permission, log):
+def assert_permissions(path, permission, log) -> bool:
     """Check whether the file's permissions are as expected, otherwise,
     log a warning message. Return a boolean indicating the match, like
     `check_permissions`.
+
+    Returns:
+        True if the permissions are as expected, False otherwise.
```
snejus: This is redundant given the `Return a boolean indicating the match` bit in the docstring.

```suggestion
```

It seems like the docstring has been lying until now though, thanks for fixing it!

beetsplug/replaygain.py
```diff
@@ -1511,21 +1511,29 @@ def command_func(
...
+                    unit="track",
+                ):
                     self.handle_track(item, write, force)
-
-            self.close_pool()
         except (SystemExit, KeyboardInterrupt):
             # Silence interrupt exceptions
             pass
+        finally:
+            self.close_pool()
```
snejus, line 1542: This is lovely, thanks

beetsplug/unimported.py
```diff
@@ -41,20 +41,38 @@ def print_unimported(lib, opts, args):
...
+                )
+
+            def file_filter(file):
+                # ignore files with ignored extensions or that are art files
+                return (
+                    not any(file.endswith(ext) for ext in ignore_exts)
+                    and file not in art_files
+                )
+
+            with ui.changes_and_errors_pbars(
```
snejus: I don't think I'm seeing a definition for this function?

beetsplug/unimported.py
```diff
@@ -41,20 +41,38 @@ def print_unimported(lib, opts, args):
                 os.path.join(lib.directory, x.encode())
                 for x in self.config["ignore_subdirectories"].as_str_seq()
             ]
-            in_folder = set()
```
snejus, line 44: It doesn't seem like this refactor is required for your change - best to stay within the scope and submit a separate PR for this.

docs/dev/plugins/other/progressbars.rst
```diff
@@ -0,0 +1,57 @@
...
+.. code-block:: python
+
+    items = ...
+    num_items = ...
+
+    for item in ui.iprogress_bar(items, desc="Analyzing", unit="track", total=num_items):
+        self.handle_track(item)
+
+A simple method for displaying progress while applying a function to all results
+in parallel is to use the ``concurrent.futures.ThreadPoolExecutor#map`` method,
```
snejus: What do we gain from parallelising this? Is the progress bar display logic so slow that it needs to be parallelised?

pyproject.toml
```diff
@@ -81,6 +81,7 @@ pydata-sphinx-theme = { version = "*", optional = true }
 sphinx = { version = "*", optional = true }
 sphinx-design = { version = "^0.6.1", optional = true }
 sphinx-copybutton = { version = "^0.5.2", optional = true }
+enlighten = "^1.14.1"
```
snejus: Note that our `pyproject.toml` file is nicely organised (I know, it's a bit unbvelievable). Since we've got a fair bit of dependencies specified there, we need to be careful about placing this in the correct place - please include this higher up, together with the rest of main (non-optional) dependencies - and follow the alphabetical order.

snejus: Additionally, replace this with the following constraint to make sure this gets automatically updated

```suggestion
enlighten = ">=1.14.1"
```

Poetry is just useless in this regard, enforcing their **^** constraint which makes no sense (allowing only _patch_ upgrades of the dependency).

poetry.lock
```diff
@@ -1,11 +1,13 @@
-# This file is automatically @generated by Poetry 1.8.5 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.2.1 and should not be changed by hand.
```
snejus: We should probably be a bit more clear about this in the docs but for now we support 'poetry<2', see the [following issue](https://github.com/beetbox/beets/issues/5783).

1. Undo all changes in `poetry.lock`
2. Update the dependency constraint following my comment
3. Run `uv tool 'poetry<2' lock --no-update`

beets/ui/__init__.py
```diff
@@ -1325,6 +1344,80 @@ def add_all_common_options(self):
...
 
+T = TypeVar("T", covariant=True)
+U = TypeVar("U")
+
+
+class SizedIterable(Protocol[T], Sized, Iterable[T]):
+    pass
+
+
+def iprogress_bar(
```
snejus, line 1261: The docs refer to this as if it was a plugin, however I see no logic which disables this by default, and enables it when `progressbar` (or equivalent) plugin is enabled?

beetsplug/embedart.py
```diff
@@ -121,7 +121,11 @@ def embed_func(lib, opts, args):
                 if not opts.yes and not _confirm(items, not opts.file):
                     return
 
-                for item in items:
+                for item in ui.iprogress_bar(
+                    items,
+                    desc="Embedding artwork",
```
snejus, line 130: There's a fair bit of duplication in every each of these function calls.

Have you considered the following?

```py
# beets/library/library.py
class Library(...):
    ...
    def items_with_progress(self, desc: str, **kwargs) -> Iterator[Item]:
        for item in ui.iprogress_bar(
            self.items(**kwargs),
            desc="Embedding artwork",
            unit="items",
        ):
            yield item
```

and then

```py
# beetsplug/embedart.py
                ...
                items = lib.items_with_progress("Embedding artwork", args)
                ...
```

snejus (changes requested):
Added a fair bit of comments

snejus:
@peterjdolan these are great changes. Are you planning to push this to completion?

peterjdolan:
@snejus thank you. Realistically I probably can't complete this any time soon - life has thrown several things in my way, and I can't spend time on it like I used to. If anybody else is interested in picking it up and carrying it across the line I would be very glad to see it.

peterjdolan:
I'm picking this back up now. I'll respond to the PR comments shortly.

beets/ui/__init__.py
peterjdolan: Done

beets/ui/__init__.py
peterjdolan: Done

beets/ui/__init__.py
peterjdolan: Done

beets/ui/__init__.py
peterjdolan: Done

beetsplug/embedart.py
peterjdolan: I like it. Done.

beetsplug/badfiles.py
peterjdolan: The base implementation uses par_map to parallelize the operation instead of simply looping in the main thread. I'm trying to maintain all existing logic, so I'm maintaining the thread-parallel structure

However, par_map's implementation is incompatible with the enlighten progress bar - it delegates work to threads and it's not possible to update the progress bar from different threads. So, instead we use a thread pool executor to achieve the same type of parallelism while keeping the progress bar iterator operating in the main thread.

If you'd rather, I could simplify this by removing the thread-parallelism, but in my testing it improves speed (over a NAS connection) by about an order of magnitude.

beetsplug/badfiles.py
peterjdolan: Ah, I thought I'd do a little code health improvements while I'm here, but yes it's muddying the commit's changes. I'll revert them.

beets/ui/__init__.py
peterjdolan: Oh, no I don't mean this to be a plugin. I mean it to be a part of the core library. In the documentation, I call it out as functionality that plugin _authors_ should be aware of and consider using, as it presents a consistent way for plugin authors to handle progress bars and progress reporting across the app.

beetsplug/duplicates.py
peterjdolan: Reverted

beetsplug/missing.py
peterjdolan: Ah, my mistake. I had misread the original logic and thought there was a bug, but I understand it now. Reverted.

beetsplug/parentwork.py
peterjdolan: My mistake, thank you.

beetsplug/permissions.py
peterjdolan: De-redundantized :)

beetsplug/replaygain.py
peterjdolan: You bet!

beetsplug/unimported.py
peterjdolan: Reverted. Thank you, and will do.

beetsplug/unimported.py
peterjdolan: My mistake, it was left over from a previous revision. I've resolved this error, and we should improve test coverage for this plugin.

pyproject.toml
peterjdolan: Done

peterjdolan: Done

poetry.lock
peterjdolan: Done, thank you.

peterjdolan (commented):
I've addressed your comments to the best of my ability, and left unresolved the comments that I had questions on. Thanks for your review!

peterjdolan:
@snejus it seems that 'poe format-docs' reformats virtually all of the docs; should I do so, or would you like another PR that does so that I can submit first?

snejus:
> @snejus it seems that 'poe format-docs' reformats virtually all of the docs; should I do so, or would you like another PR that does so that I can submit first?

Try rebasing on `master` and running `poetry install` - I've now updated it and configured it correctly.

masroof-maindak:
Many thanks to everyone involved.

Could we please get an expected date for when this will be merged?

~If it's a while in the future, does anyone know an easy way to install this fork using `uv`?~ Edit: the `--from` flag.

PR #5696, opened 2025-03-28
library: add subtitle (id3 TIT3) field

### LINKED ISSUE #6563: Subtitle support
Support for track subtitles (remix, remaster, etc)

### Proposed solution

Add support for subtitle field

### Objective

https://github.com/beetbox/beets/discussions/6409

daniele-athome:
## Description

Adds support for a subtitle field, corresponding to the `TIT3` id3 tag. Requires beetbox/mediafile#82

I don't think tests are required for this - I can't really see a comprehensive test suite for all available fields - but I might be wrong.

I'll add the changelog entry after your review.

## To Do

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -31,6 +31,7 @@ New features
   ``albums_for_ids`` and ``tracks_for_ids``. It also enables search by query as
   well as identifier-based retrieval, with support for ISRC codes (tracks) and
   barcode/EANs (albums).
+- Add support for adding or modifying a subtitle (ID3 tag ``TIT3``) field
 
   This is an initial, relatively minimal implementation, but already fully
   usable for common metadata workflows. We welcome feedback, improvement ideas,
--- a/beets/library/fields.py
+++ b/beets/library/fields.py
@@ -54,6 +54,7 @@ TYPE_BY_FIELD: dict[str, types.Type] = {
     "format": types.STRING,
     "genres": types.MULTI_VALUE_DSV,
     "grouping": types.STRING,
+    "subtitle": types.STRING,
     "id": types.PRIMARY_ID,
     "initial_key": types.MusicalKey(),
     "isrc": types.STRING,
--- a/beets/library/models.py
+++ b/beets/library/models.py
@@ -642,6 +642,7 @@ class Item(LibModel):
         "encoder_settings",
         "format",
         "grouping",
+        "subtitle",
         "initial_key",
         "isrc",
         "length",
--- a/docs/reference/pathformat.rst
+++ b/docs/reference/pathformat.rst
@@ -241,6 +241,7 @@ Ordinary metadata:
 - lyricists: The track lyricists as a multi-valued field.
 - remixers: The track remixers as a multi-valued field.
 - grouping
+- subtitle
 - year, month, day: The release date of the specific release.
 - original_year, original_month, original_day: The release date of the original
   version of the album.


michaelkrieger:
Much needed as seems actively used by Picard now and in Navidrome.

michaelkrieger:
The [musicbrainz plugin](https://github.com/beetbox/beets/blob/master/beetsplug/musicbrainz.py) should also be updated to pull the field and populate it as well.

Did this addition stall? Seems increasingly commonly used with Picard and now [Navidrome’s metadata enhancements](https://github.com/navidrome/navidrome/issues/1036).

daniele-athome:
> Did this addition stall?

I don't know... there are frequent commits on the main branch but some PRs are ignored. There must surely be a reason but the core devs would have to share that. Or they simply forgot about this :-)

beets/library.py
```diff
@@ -546,6 +546,7 @@ class Item(LibModel):
         "work_disambig": types.STRING,
         "arranger": types.STRING,
         "grouping": types.STRING,
+        "subtitle": types.STRING,
```
snejus: How does this field get populated? Which metadata sources support it? We may want to add some logic for this.

Or maybe you handle this using an external plugin?

snejus (changes requested):
Added a comment. I reckon this change should be included in the changelog too

beets/library.py
daniele-athome: For the moment, no plugin will populate the field. However, adding the field will allow it to be managed and read/written from/to files.

daniele-athome:
I added a changelog entry.

snejus (approved):

snejus:
Aw snap just realised this depends on an unmerged PR in mediafile repo :/

semohr:
I just merged https://github.com/beetbox/mediafile/pull/82 :upside_down_face:. You up to rebase and revive this?

daniele-athome:
Rebased!! :rocket:

semohr:
Awesome! Let me see if I'm able to create a new mediafile version and how to do that :fearful:

semohr:
I don't think I'm able to do a version bump for mediafile. I raised the issue in our internal chat tho. We will have to wait some more, sorry :see_no_evil:

daniele-athome:
Hi there! Any luck in releasing a new version of mediafile? We are like a meter from the finish line! :smile: 

I'm ready to rebase this whenever needed.

semohr:
We should be able to do that in the near future :tada: We sill need to create a (maintainable) release workflow but other than that we should be able to create a release.

snejus:
@daniele-athome new `mediafile` version `0.14.0` has been released, try updating with 

```sh
$ uv tool run 'poetry<2' update mediafile
```

daniele-athome:
On it!

daniele-athome:
It took some time - sorry, life got in the way - but I'm finally there. Rebased on latest master and mediafile upgraded.

poetry.lock
```diff
@@ -1870,21 +1870,21 @@ files = [
 
 [[package]]
 name = "mediafile"
-version = "0.13.0"
-description = "Handles low-level interfacing for files' tags. Wraps Mutagen to"
+version = "0.14.0"
```
snejus: It's best to also update the requirement in `pyproject.toml` to make sure it's clear we need at least `0.14.0` of `mediafile`.

docs/changelog.rst
```diff
@@ -45,6 +45,7 @@ New features:
   of brackets are supported and a new ``bracket_keywords`` configuration option
   allows customizing the keywords. Setting ``bracket_keywords`` to an empty list
   matches any bracket content regardless of keywords.
+- Add support for importing and writing a subtitle (ID3 tag `TIT3`) field
```
snejus: I actually wanted to ask about the following:

- Add support for _importing_ and writing

It doesn't seem like this PR adds support for importing this field? Currently, this field is ignored - it can only be populated manually using `beet modify ... subtitle=...`.

What is your use case for it?

snejus (changes requested):

docs/changelog.rst
daniele-athome: It doesn't work when importing metadata from an external source (e.g. MusicBrainz), but it worked for importing the field from source files (e.g. an existing TIT3 field in an mp3 file).

My specific use case was to edit the field after import (I wanted to remove it).

snejus: Have you considered removing this field from your music files before importing them?

For mp3 files you could use `id3v2 --remove-frame TIT3 <files>` and for flacs you may use `metaflac --remove-tag SUBTITLE`

daniele-athome: I did consider it, but since I use beets for anything related to my music collection, I thought it would be nice to implement it directly in beets and contribute it.

I know this is not 100% complete (because there is no way to actually import the field from sources, e.g. MusicBrainz), but it can pave the way for the next person to implement that (or I might even implement it myself if I find the time).

snejus: I share that concern. Adding a field to the core database just to remove it feels wrong architecturally.

The **`scrub` plugin** is designed exactly for this—stripping unwanted metadata before writing. But it has a limitation: it removes *all* unknown tags, with no way to selectively target specific ones.

Your actual need is granular control: remove TIT3 without touching other tags. Neither `scrub` nor `zero` supports that today.[4][5]

Rather than adding subtitle to the schema, **consider extending `scrub` with a blacklist option** to remove specific frames by name. Something like:[3]

```yaml
scrub:
  auto: yes
  remove_frames: [TIT3, COMM, USLT]
```

This solves the problem cleanly for anyone managing legacy metadata, not just subtitle fields. It's also better than shelling out to `id3v2` or `metaflac` externally.

What do you think?

daniele-athome: > Rather than adding subtitle to the schema, consider extending scrub with a blacklist option to remove specific frames by name.

It's a nice idea and I might do that as well, another time.

> I share that concern. Adding a field to the core database just to remove it feels wrong architecturally.

After your explanation about `scrub`, I totally agree. Does that mean that you are considering not adding support for the field in beets afterall? Despite its intended use by me, the new field can be used for importing (given someone - or myself - implement it in the MusicBrainz plugin for example).

daniele-athome: (I was about to rebase again but at this point I think I'll wait for an answer before doing it)

daniele-athome:
Hi there! Any news on this? I can rebase (again) if necessary.

snejus:
Thanks for your patience and for all the work you've put into keeping this rebased!

> After your explanation about scrub, I totally agree. Does that mean that you are considering not adding support for the field in beets afterall? Despite its intended use by me, the new field can be used for importing (given someone - or myself - implement it in the MusicBrainz plugin for example).

To answer your question directly: yes, I'm inclined not to merge the subtitle field into core beets at this point. The primary use case described here - removing an unwanted TIT3 tag from existing files - doesn't justify adding a first-class field to the database schema. A field in the core library implies a full lifecycle: importing from metadata sources, writing back, querying, and path formatting. Right now none of that is wired up for subtitle, and merging it in an incomplete state sets a precedent that's hard to walk back.

The right home for this is the `scrub` plugin, extended with a `remove_frames` (or similar) option that lets users target specific raw tags by name. That cleanly solves the "I want to strip TIT3 without blowing away everything else" problem, works for any tag - not just subtitle - and doesn't require touching the core schema at all. It'd be a genuinely useful, standalone contribution.

If you'd like to pursue that instead, I'd be happy to review a PR against `beetsplug/scrub.py`. And if subtitle support in MusicBrainz import ever gets fully fleshed out, we can revisit adding it to core then with a complete implementation.

Sorry for the slow feedback loop on this one - I appreciate your persistence. ❤️

daniele-athome:
>  A field in the core library implies a full lifecycle: importing from metadata sources, writing back, querying, and path formatting. Right now none of that is wired up for subtitle, and merging it in an incomplete state sets a precedent that's hard to walk back.

Although I don't agree completely (I believe that, especially in the open-source volunteering world, small iterations count), I understand your point for maintainability and future-proofing.

> The right home for this is the scrub plugin, extended with a remove_frames (or similar) option that lets users target specific raw tags by name. That cleanly solves the "I want to strip TIT3 without blowing away everything else" problem, works for any tag - not just subtitle - and doesn't require touching the core schema at all. It'd be a genuinely useful, standalone contribution.

I might contribute that if I find the time :+1:

snejus:
@daniele-athome, I think we've got a use case for this field now! See https://github.com/beetbox/beets/issues/6561

snejus:
Would you mind rebasing this on `master`? Undo your changes in `poetry.lock` - `mediafile` has already been updated to later versions.

daniele-athome:
Sure, just rebased!

snejus:
There's just one issue in the changelog :)

snejus (approved):
Thanks!

daniele-athome:
> There's just one issue in the changelog :)

Yeah realized it a few minutes later eheh  
Just fixed, I think.

PR #1277, opened 2015-01-27
Create thumbnails from covers for album folders

brunal:
Relates to #1276. Thumbnailing works fine.
I still have an issue with some folders not displaying their thumbnail. This does not seem encoding-related.
This is only the code, so I still have to write tests, docs and changelog.

PR #5708, opened 2025-04-04
plugins/thumbnails: fix FFI with GIO on s390x

Hyask:
## Description

Using the correct function signature for g_file_new_for_path fixes the tests on s390x.
I do not have the full story on why this failed consistently only on s390x, but I guess the big endian might have something to play with this.

Here is how the tests were failing:
```
169s ___________________________ ThumbnailsTest.test_uri ____________________________
169s
169s self = <test.plugins.test_thumbnails.ThumbnailsTest testMethod=test_uri>
169s
169s     def test_uri(self):
169s         gio = GioURI()
169s         if not gio.available:
169s             self.skipTest("GIO library not found")
169s
169s >       assert gio.uri("/foo") == "file:///"  # silent fail
169s E       AssertionError: assert '' == 'file:///'
169s E
169s E         - file:///
169s
169s test/plugins/test_thumbnails.py:268: AssertionError
```
You can see a full log here [1] and a history of consistent failure here [2]. Both links are bound to expire at some point, sorry future archeologist 🤷.

[1]: https://autopkgtest.ubuntu.com/results/autopkgtest-plucky/plucky/s390x/b/beets/20250403_162414_5d1da@/log.gz#S5
[2]: https://autopkgtest.ubuntu.com/packages/beets/plucky/s390x

## To Do

- Discuss the behavior in case a string is passed. I've currently updated the test to reflect the exception being raised, but I did not do a thorough check of the code to know if that exception would need to be caught elsewhere in the code, or if that API might be used externally.

### DIFF
--- a/beetsplug/thumbnails.py
+++ b/beetsplug/thumbnails.py
@@ -246,7 +246,7 @@ class GioURI(URIGetter):
         if self.available:
             self.libgio.g_type_init()  # for glib < 2.36
 
-            self.libgio.g_file_get_uri.argtypes = [ctypes.c_char_p]
+            self.libgio.g_file_new_for_path.argtypes = [ctypes.c_char_p]
             self.libgio.g_file_new_for_path.restype = ctypes.c_void_p
 
             self.libgio.g_file_get_uri.argtypes = [ctypes.c_void_p]
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -24,6 +24,8 @@ New features:
 
 Bug fixes:
 
+* Fix API call to GIO on big endian architectures (like s390x) in thumbnails plugin.
+  :bug:`5708`
 * :doc:`plugins/listenbrainz`: Fix rST formatting for URLs of Listenbrainz API Key documentation and config.yaml.
 * :doc:`plugins/listenbrainz`: Fix ``UnboundLocalError`` in cases where 'mbid' is not defined.
 * :doc:`plugins/fetchart`: Fix fetchart bug where a tempfile could not be deleted due to never being
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -24,7 +24,8 @@ New features:
 
 Bug fixes:
 
-* Fix API call to GIO on big endian architectures (like s390x) in thumbnails plugin.
+* :doc:`plugins/thumbnails`: Fix API call to GIO on big endian architectures
+  (like s390x) in thumbnails plugin.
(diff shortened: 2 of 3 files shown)


wisp3rwind (approved):
Wow, that copy-and-paste error (I presume) seems to have been around forever (10 years, to be precise: https://github.com/beetbox/beets/commit/530b42156b0e4eb1814093c2735edff73881e920); thanks for the fix :tada:

Given https://docs.gtk.org/gio/type_func.File.new_for_path.html, this appears to be correct.

> Discuss the behavior in case a string is passed. I've currently updated the test to reflect the exception being raised, but I did not do a thorough check of the code to know if that exception would need to be caught elsewhere in the code, or if that API might be used externally.

While undocumented, unchecked and untyped (and I'm unfamiliar with this part of the code), I'm fairly confident that this is only expected to take bytestrings. That's (historically) the internal representation of paths that beets chooses (with the caveat that I'm quite sure that was never 100% correctly implemented). Additionally, with the previous code, ctypes would pass the argument as `wchar_t *`, which presumably ended up being interpreted as `b"\\\0"` on little-endian systems (and the empty string on big-endian systems, which might explain the test failures). In any case, the result would definitely be broken thumbnails (or rather failure to even write them).

So, the only sense in which this could regress seems to be by different failures modes, in particular during import: Logging a warning/silent failure due to the broken code (before) and a crash with `ArgumentError` (after).

Given that the thumbnail plugin seems to be essentially unmaintained (the only contributions since its inception seem to be small bug fixes by drive-by contributors, and repo-wide refactorings), I'd argue to just merge this and deal with potential fall-out later. @snejus @JOJ0 any objections?

Eventually, it would be nice if the thumbnails plugin would also be subject to typing (slowly, beets is getting type annotations throughout its codebase) to highlight incorrect usage of these functions.

I'm also wondering whether `gio` is even required here these days: From a quick glance at the [PR which added it](https://github.com/beetbox/beets/pull/1277), there used to be some discrepancy between Python and glib, but that's really a long time ago. Maybe, the much simpler pathlib implementation is sufficient nowadays?

wisp3rwind:
Would you mind adding a quick changelog entry?

Hyask:
Thanks for the review, and the hindsight on that little-endian/big-endian thing. That was more or less my conclusion too, but I really don't know enough about these things to be confident in a clear explanation.

I've added an entry to the changelog, hope it's good enough :-)

PR #5716, opened 2025-04-09
Add fetchart typing v2

wisp3rwind:
Adds typings to fetchart; a revised and modernized version of https://github.com/beetbox/beets/pull/5213.

The former pull request became stale because I did more refactoring there, in order to express properties of the `Candidate` (whether or not it is validated) via the type system (motivated by the type state pattern which is more common in Rust.
The result was a little contentious, since it became a little unclear what the purpose of the `Candidate` class should even be at that point.

I think this was a case of combining too many things into a single PR. Thus, this version does not perform any refactoring of that extent. It does sprinkle in a few `assert`s to make thinks clear for the type checker.

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -262,6 +262,11 @@ class Candidate:
         """Resize the candidate artwork according to the plugin's
         configuration and the specified check.
         """
+        # This must only be called when _validate returned something other than
+        # ImageAction.Bad or ImageAction.EXACT; then path and size are known.
+        assert self.path is not None
+        assert self.size is not None
+
         if check == ImageAction.DOWNSCALE:
             self.path = ArtResizer.shared.resize(
                 plugin.maxwidth,
@@ -270,9 +275,6 @@ class Candidate:
                 max_filesize=plugin.max_filesize,
             )
         elif check == ImageAction.DOWNSIZE:
-            # This must only be called when _validate returned something other
-            # than ImageAction.Bad or ImageAction.EXACT; then the size is known.
-            assert self.size is not None
             # dimensions are correct, so maxwidth is set to maximum dimension
             self.path = ArtResizer.shared.resize(
                 max(self.size),
@@ -424,7 +426,11 @@ class RemoteArtSource(ArtSource):
         """Downloads an image from a URL and checks whether it seems to
         actually be an image.
         """
+        # This must only be called for candidates that were returned by
+        # self.get, which are expected to have a url and no path (because they
+        # haven't been downloaded yet).
         assert candidate.path is None
+        assert candidate.url is not None
 
         if plugin.maxwidth:
             candidate.url = ArtResizer.shared.proxy_url(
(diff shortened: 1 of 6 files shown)


wisp3rwind:
Did a quick test with ImageMagick in CI, since otherwise, most fetchart test were skipped. Now all fetchart tests succeed, but one embedart test fails in a way that pytest handles really badly when rewriting the assertion. Locally, this test succeeds for me.

This might be due to https://github.com/ImageMagick/ImageMagick/discussions/5191

beetsplug/fetchart.py
```diff
@@ -14,10 +14,15 @@
...
 
+from __future__ import annotations
+
 import os
 import re
+from abc import ABC, abstractmethod
 from collections import OrderedDict
 from contextlib import closing
+from enum import Enum
+from typing import TYPE_CHECKING, AnyStr, ClassVar, Literal, Tuple, Type
```
semohr, line 25: You might want to consider omitting AnyStr and definig the constained type yourself. `AnyStr` is getting deprecated in 3.13.

beetsplug/fetchart.py
```diff
@@ -188,23 +216,29 @@ def _validate(self, plugin, skip_check_for=None):
...
-
-    def resize(self, plugin):
+            return ImageAction.DEINTERLACE
+        if downsize and (ImageAction.DOWNSIZE not in skip_check_for):
+            return ImageAction.DOWNSIZE
+        return ImageAction.EXACT
+
+    def validate(
+        self,
+        plugin,
```
semohr, line 235: `plugin: FetchArtPlugin`

beetsplug/fetchart.py
```diff
@@ -568,13 +659,18 @@ def add_default_config(config):
         config["google_key"].redact = True
 
     @classmethod
-    def available(cls, log, config):
+    def available(cls, log, config: confuse.ConfigView) -> bool:
```
semohr, line 662: `log:logging.Logger`

wisp3rwind:
This should be good to go from my side now. The embedart CI thing is a bit of a hack, but from my perspective still an improvement compared to before (where we didn't run many art tests in CI at all).

semohr (changes requested):
I think the changelog needs an update to move the changes to the unreleased section. Other than that, I just had another look and the changes seem good to me :+1:

semohr (approved):

snejus:
It seems like this was more than just typing! Just found this broke `beetcamp`:

```
  File "/home/sarunas/.local/bin/beet", line 5, in <module>
    main()
  File "/home/sarunas/repo/beets/beets/ui/__init__.py", line 1878, in main
    _raw_main(args)
  File "/home/sarunas/repo/beets/beets/ui/__init__.py", line 1853, in _raw_main
  File "/home/sarunas/repo/beets/beets/ui/__init__.py", line 1693, in _setup
    plugins.send("pluginload")
  File "/home/sarunas/repo/beets/beets/plugins.py", line 634, in send
    result = handler(**arguments)
  File "/home/sarunas/repo/beets/beets/plugins.py", line 206, in wrapper
    return func(*args, **kwargs)
  File "/home/sarunas/repo/beetcamp/beetsplug/bandcamp/__init__.py", line 178, in loaded
    fetchart.ART_SOURCES[self.data_source] = BandcampAlbumArt
TypeError: 'set' object does not support item assignment
```

PR #5718, opened 2025-04-10
Parallelize several built-in plugin operations

peterjdolan:
## Description

Fixes #X.  

Several built-in plugins' operations can be parallelized, as they operate on albums or tracks individually, thus avoiding any race or synchronization issues. Thus, we can safely parallelize them with the Python multiprocessing libraries. This PR introduces parallelism to the ftintitle command.

On testing, the mbsync command does not gain any significant performance, as it is limited not by processing or database operations, but by the rate-limited requests made to the Musicbrainz backend.

FtInTitle, on the other hand, gains a substantial improvement, going from 2 titles / second to 200-2000 titles / second on a computer with 32 cores.

## To Do

### DIFF
--- a/beets/library.py
+++ b/beets/library.py
@@ -904,6 +904,8 @@ class Item(LibModel):
         `move` controls whether the path should be updated. In the
         latter case, files are *only* moved when they are inside their
         library's directory (if any).
+        `with_album` controls whether the album should also be updated when the
+        item is moved.
 
         Similar to calling :meth:`write`, :meth:`move`, and :meth:`store`
         (conditionally).
@@ -1713,7 +1715,7 @@ class Library(dbcore.Database):
 
     # Convenience accessors.
 
-    def get_item(self, id):
+    def get_item(self, id: int) -> Item | None:
         """Fetch a :class:`Item` by its ID.
 
         Return `None` if no match is found.
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -177,6 +177,10 @@ def should_move(move_opt=None):
     )
 
 
+def threaded():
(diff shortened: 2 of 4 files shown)


wisp3rwind:
> FtInTitle, on the other hand, gains a substantial improvement, going from 2 titles / second to 200-2000 titles / second on a computer with 32 cores.

That doesn't make a whole lot of sense, the speedup should be closer to a factor 32 then. It seems like more is going here than just parallelization, possibly the behavior actually changed in some unintended way?

In any case, this PR seems to contain a bunch of unrelated changes and it's quite hard to get a reasonable overview of what really changed. It would be helpful if it could be focused on just introducing the parallelism without any extra changes to the plugin logic or logging (if that is actually relevant for the speed-up, it would be useful to point that out).

peterjdolan:
Yes, you're absolutely right. I looked at it again and I think the main performance improvement comes from only calling `item.store()` when the item was changed. I'll rewrite this to include only that modification.

wisp3rwind (approved):
Looks good to me, thanks!

I went ahead, rebased this on master & squashed.

wisp3rwind:
For reference, the fact that this allows for such a large speed-up makes me wonder:

- What is the impact of `Model.store()` in general? Given that there has been a bunch of `dbcore` optimizations, it would be surprising if this remains inefficient. (But maybe optimizations mainly concerned queries?)
- Should there be an early exit in `Model.store()` if there are no dirty fields?
- Is this not really an issue with `model.store()`, but rather with `LibItem.store()` which notifies plugins of `database_change` combined with a plugin that performs some slow operation upon receiving that event? @peterjdolan is there anything particular about your beets configuration that would make this likely?

PR #5729, opened 2025-04-15
pipeline: remove old tests, integrate with our test suite

wisp3rwind:
These have probably not been run by anyone in ages, better to move the code to our test suite where it is regularly exercised. In fact, the latter covers most of the cases already. The only missing tests seem to be those were exceptions are raised in the first or last stage. Thus, this adds such tests.


semohr:
Funnily we used these tests recently when we re-implemented the pipeline in an async execution pattern. Was useful for timing and benchmark the current model against our new and a bit more complex implementation.

Nonetheless, I think it should not really used anymore.

wisp3rwind:
> Funnily we used these tests recently when we re-implemented the pipeline in an async execution pattern. Was useful for timing and benchmark the current model against our new and a bit more complex implementation.
> 
> Nonetheless, I think it should not really used anymore.

A bit off-topic, but where could I find your re-implementation? I've been working on better parallelism in beets a while ago (see for example https://github.com/beetbox/beets/pull/3893, which I never finished up, and according to the comment in https://github.com/beetbox/beets/pull/4851 I came to the conclusion that it's not the best approach), so I'd be interested to see the approach that you chose.

semohr:
I put the relevant pipeline code into a gist [here](https://gist.github.com/semohr/8df181d5485cd3b26e0f38fb07a291ef). Feel free to leave some comments if you have questions.

snejus (approved):
Looks good to me

snejus:
Have set this to merge automatically once the branch is rebased.

Issue #2572, opened 2017-05-21
Feature request: Query to find albums with/without embedded cover art


kupietools:
Would like to be able to run a query for albums with/without embedded cover art. As it is, I have to import my mp3s back into iTunes, create a smart playlist, and then tag them under 'grouping' as having or not having cover art, and then search in Beets on that. Would like to be able to do it without needing iTunes as an intermediary.

sampsyo:
This has come up before, and it wouldn't be hard to add. We should totally do it. Thanks for filing the issue!

kuzi-moto:
+1 from me. I asked about this on Discourse and meant to open a request myself. I would really like to more easily find my albums missing embedded art!

BardsSong:
Hi! I'm looking to implement this feature, but I'm struggling a little with understanding how to create a subclass of Query, is there another similar plugin that I can look at to see an example?

jackwilsdon:
@BardsSong take a look at these:
 * [Extend the Query Syntax](https://github.com/beetbox/beets/blob/9f9af300e8de7be7b1bb6575e4ccb4464fd1ca34/docs/dev/plugins.rst#extend-the-query-syntax)
 * [`playlist.py`](https://github.com/beetbox/beets/blob/9392256993b0cf6f4e2a545d9c504decb71542d9/beetsplug/playlist.py#L24)

It seems like we only have one plugin which uses item queries and none that use album queries at the moment, so I hope the links above are enough to get you started :+1:

kergoth:
Adding a simple query for this is straightforward, but `art.get_art()` is really slow, since it has to pull it out of the file with mediafile, so doing such a query takes ages with a large library. Better than nothing, sure. Alternatively, we could probably have such a plugin add a new field and then keep the field in sync with the file via appropriate hooks, to avoid calling `get_art()` repeatedly during the query.

Side note, beets will check every Query.match against everything up front. That is, if you do a `beet ls billy boy on poison has_art:1` it still runs HasArtQuery.match() against every Item, not just the ones that already match the rest of the query. It makes sense, as I'm sure it'd be more complex for the implementation to try to do the query objects that have an sql clause first and then match against that, but worth taking note of, particularly in the case of slow field queries like this one.

cz3k:
Would love to be able to query for missing album art as well. 

My current workaround is creating a smart playlist with the [Feishin](https://github.com/jeffvli/feishin) client for Navidrome. It provides the `Has CoverArt` query.

PR #5730, opened 2025-04-15
Add art:true/false album query to detect embedded cover art

CarolineKevnick:
Adds support for a new `art:true` / `art:false` album-level query that allows users to filter albums based on whether they contain embedded cover art in their audio files.

- Adds a new `art` boolean field to the Album model
- Detects embedded art using `MediaFile.art` during import
- Persists the field in the library database
- Registers a custom `ArtQuery` so users can run CLI queries
- Example usage: `beet ls -a art:true`

This feature improves the usability of beets for users who embed artwork directly into music files and want to better manage or detect it.

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -1052,3 +1052,13 @@ class SmartArtistSort(FieldSort):
             return val.lower() if self.case_insensitive else val
 
         return sorted(objs, key=key, reverse=not self.ascending)
+
+
+from beets.dbcore.query import FieldQuery, BooleanQuery
+from beets import util
+
+class ArtQuery(FieldQuery):
+    """Query that matches albums with or without embedded art."""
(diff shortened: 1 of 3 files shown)


wisp3rwind:
Hi, thanks for contributing to beets!

However, I see a few issues here:

- Please read & follow our development guide at https://github.com/beetbox/beets/blob/master/CONTRIBUTING.rst, this PR messes up our code style in many (also unrelated) files, making it impossible to review. Eventually, this should also have tests, but it might make sense to first discuss here whether the code seems like a sensible approach to the problem.
- I suspect this is motivated by https://github.com/beetbox/beets/issues/2572? It'd be great to mention that: In this particular case, your pull requests give me the impression that this might be a class project (correct me if I'm wrong!) rather than improving beets for your actual usecase. On its own, I don't think that would be sufficient motivation for adding a feature that potentially nobody uses.

snejus:

PR #5734, opened 2025-04-16
Art resizer typings v2

wisp3rwind:
A rebased & amended version of https://github.com/beetbox/beets/pull/4649.

I left a FIXME note regarding a cleanup that I feel should be done at some point, but that is out-of-scope here when just adding initial typings.

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -40,6 +40,7 @@ from typing import (
     Any,
     AnyStr,
     Callable,
+    Generic,
     Iterable,
     NamedTuple,
     TypeVar,
@@ -1039,6 +1040,30 @@ class cached_classproperty:
             self.cache[owner] = self.getter(owner)
 
         return self.cache[owner]
+
+
+T = TypeVar("T")
+
+
+class LazySharedInstance(Generic[T]):
+    """A descriptor that provides access to a lazily-created shared instance of
+    the containing class, while calling the class constructor to construct a
+    new object works as usual.
+    """
+
+    _instance: T | None = None
+
+    def __get__(self, instance: T | None, owner: type[T]) -> T:
+        if instance is not None:
+            raise RuntimeError(
+                "shared instances must be obtained from the class property, "
+                "not an instance"
+            )
+
+        if self._instance is None:
+            self._instance = owner()
+
+        return self._instance
 
 
 def get_module_tempdir(module: str) -> Path:
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -1046,6 +1046,32 @@ class LazySharedInstance(Generic[T]):
     """A descriptor that provides access to a lazily-created shared instance of
     the containing class, while calling the class constructor to construct a
     new object works as usual.
+
+    ```
+    ID: int = 0
+
+    class Foo:
+        def __init__():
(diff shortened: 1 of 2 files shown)


beets/util/__init__.py
```diff
@@ -1094,6 +1095,30 @@ def __get__(self, instance, owner):
         return self.cache[owner]
 
 
+T = TypeVar("T")
+
+
+class LazySharedInstance(Generic[T]):
+    """A descriptor that provides access to a lazily-created shared instance of
+    the containing class, while calling the class constructor to construct a
+    new object works as usual.
```
semohr, line 1104: Can we add a usage example here?

Is there any benefit in using this pattern instead of the more traditional singleton pattern?

beets/util/artresizer.py
```diff
@@ -51,30 +61,100 @@ class LocalBackendNotAvailableError(Exception):
...
-    def available(cls):
+    def available(cls) -> bool:
         try:
             cls.version()
             return True
         except LocalBackendNotAvailableError:
             return False
 
+    @abstractmethod
+    def resize(
```
semohr, line 90: Would it be possible to get some docstrings for the abstract methods?  I'm not too familiar with the artresizer, a sentence on each each of them would be highly appreciated.

beets/util/artresizer.py
```diff
@@ -194,12 +284,22 @@ def get_size(self, path_in):
             )
             return None
         try:
-            return tuple(map(int, out.split(b" ")))
+            size = tuple(map(int, out.split(b" ")))
```
semohr, line 287: The size isn't returned anymore or am I missing something?

beets/util/__init__.py
```diff
@@ -1094,6 +1095,30 @@ def __get__(self, instance, owner):
         return self.cache[owner]
 
 
+T = TypeVar("T")
```
semohr, line 1098: `T`  is already defined at the top, should be removed here.

beets/util/artresizer.py
wisp3rwind: Done.

beets/util/__init__.py
wisp3rwind: If I remember correctly, this was the result of refactoring the existing code (with the `ArtResizer.shared` accessor) in a way that I managed to write typings for and seemed a bit simpler than the previous implementation.

Whether we should have the `ArtResizer.shared` and `ArtResizer()` to get shared and owned instances, or just a singleton is a separate discussion that I'd prefer not to open up in this PR, because it would require changing all uses of the art resizer. Currently, this PR leaves the interface unchanged.

I'll add an example.

semohr: All right thanks!

I was just wondering why this changed. It is the only none typing part in this PR  in comparison to the previous implementation.

If it works tho, I'm happy with it :+1:

wisp3rwind: There was something similar before (the `Shareable` metaclass), but I moved it from `artresizer.py` to `__init__.py` and rewrote it as "just" a descriptor instead of a full-blown metaclass.

wisp3rwind:
Should be ready from my side now.

semohr (approved):

PR #5735, opened 2025-04-17
Rewrite the import pipeline as an asynchronous state machine.

peterjdolan:
## Description

See discussion https://github.com/beetbox/beets/discussions/5736

Implements an asynchronous state machine which can process tasks in
multiple states concurrently, and transition the tasks between the
states based on the task's properties. This is a better representation
of the Beets import process than the Pipeline, as it makes explicit the
possibility to skip stages based on the user's configuration, and to
return to previous states in a cycle to support iterative
decision-making by the user. The Pipeline supported only a linear
progression of stages, and implemented cyclic behavior by constructing
and running sub-pipelines within pipeline stages. This was potentially
fragile, difficult to test, and difficult to understand.

The state machine replacement for the pipeline is implemented using the
asyncio Python libraries, specifically to support concurrent processing
of multiple items in multiple stages of the import process while not
incurring the overhead and complexity of multi-threading and
synchronization. Many long-running or I/O-intensive operations are
scheduled in a ThreadPoolExecutor by the methods processing import tasks
for various states, but the operations performed on those threads are
always contained to the scope of a single state's processing method and
avoid any need for synchronization.

Fixes #X.  

(...)

## To Do


semohr:
Heya, I have just seen this and have implemented something similar quite recently, see [here](https://gist.github.com/semohr/8df181d5485cd3b26e0f38fb07a291ef).

 We use my implementation going forward from the next release of our [beets web interface](https://github.com/pSpitzner/beets-flask) instead of the normal threaded and blocking execution. Wanted to make a PR myself once we are done with our release.
 
 Might be of value/interest for you.

peterjdolan:
Closing as infeasible, see discussion.

> Heya, I have just seen this and have implemented something similar quite recently, see [here](https://gist.github.com/semohr/8df181d5485cd3b26e0f38fb07a291ef).
> 
> We use my implementation going forward from the next release of our [beets web interface](https://github.com/pSpitzner/beets-flask) instead of the normal threaded and blocking execution. Wanted to make a PR myself once we are done with our release.
> 
> Might be of value/interest for you.

Nice! Thanks for the pointer. That kicked me into rethinking my approach, and I ultimately decided it's not worth the complexity tradeoff for the core pipeline, at least as I've approached it. I think yours is much simpler. I'll be eager to see what performance tradeoffs you find - have you done any baseline measurements?

semohr:
Ah, definitely didn’t mean to suggest you should stop pursuing it altogether — just wanted to share in case it helped with ideas or comparison!

I’ve done a little bit of benchmarking, and so far the parallel execution seems promising, especially for request-bound tasks, which tend to be the most time-consuming in beets.

Once we’ve got things stabilized post-release, I’m hoping to dive a bit deeper into that. Would be great to trade notes if you end up revisiting it!

peterjdolan:
Thanks! And no, I didn't think you were suggesting I stop - I found your pointers and your code's approach really helpful in re-thinking how I was approaching it. Changing my thinking was entirely on my part.

I'd love to trade notes! I'm probably going to keep poking at this because I can't leave things like this alone :)

PR #5739, opened 2025-04-17
plugins: ensure to log the offending plugin on instantiation failure

wisp3rwind:
Came up in https://github.com/beetbox/beets/pull/5701#discussion_r2049220759 where an error can't easily be traced back to a plugin.

(Draft because I want to avoid creating merge conflicts for https://github.com/beetbox/beets/pull/5701 before that is merged<del>, and because I might a few additional improvements here</del>.)

### DIFF
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -356,32 +356,30 @@ def load_plugins(names: Sequence[str] = ()) -> None:
     """
     for name in names:
         modname = f"{PLUGIN_NAMESPACE}.{name}"
-        try:
-            try:
-                namespace = __import__(modname, None, None)
-            except ImportError as exc:
-                # Again, this is hacky:
-                if exc.args[0].endswith(" " + name):
-                    log.warning("** plugin {0} not found", name)
-                else:
-                    raise
-            else:
-                for obj in getattr(namespace, name).__dict__.values():
-                    if (
-                        isinstance(obj, type)
-                        and issubclass(obj, BeetsPlugin)
-                        and obj != BeetsPlugin
-                        and obj != MetadataSourcePlugin
-                        and obj not in _classes
-                    ):
-                        _classes.add(obj)
 
-        except Exception:
-            log.warning(
-                "** error loading plugin {}:\n{}",
-                name,
-                traceback.format_exc(),
-            )
+        try:
+            namespace = __import__(modname, None, None)
+        except ImportError as exc:
+            # Again, this is hacky:
+            if exc.args[0].endswith(" " + name):
+                log.warning("** plugin {0} not found", name)
+            else:
+                log.warning(
+                    "** error loading plugin {}:\n{}",
+                    name,
+                    traceback.format_exc(),
+                )
+            continue
+
+        for obj in getattr(namespace, name).__dict__.values():
+            if (
+                isinstance(obj, type)
+                and issubclass(obj, BeetsPlugin)
+                and obj != BeetsPlugin
+                and obj != MetadataSourcePlugin
+                and obj not in _classes
+            ):
+                _classes.add(obj)
 
 
 _instances: dict[type[BeetsPlugin], BeetsPlugin] = {}
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -50,7 +50,7 @@ else:
 
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable
+    from collections.abc import Iterable, Iterator
 
     from confuse import ConfigView
 
@@ -346,7 +346,7 @@ class BeetsPlugin:
         return helper
 
 
-_classes: set[type[BeetsPlugin]] = set()
+_instances: dict[type[BeetsPlugin], BeetsPlugin] = {}
 
 
 def load_plugins(names: Sequence[str] = ()) -> None:
@@ -355,6 +355,8 @@ def load_plugins(names: Sequence[str] = ()) -> None:
     package in sys.path; the module indicated should contain the
     BeetsPlugin subclasses desired.
     """
+    classes = set()
+
     for name in names:
         try:
             mod = import_module(f".{name}", package=PLUGIN_NAMESPACE)
@@ -369,43 +371,39 @@ def load_plugins(names: Sequence[str] = ()) -> None:
             )
             continue
 
-        for _name, obj in inspect.getmembers(mod, inspect.isclass):
+        for _name, cls in inspect.getmembers(mod, inspect.isclass):
             if (
-                issubclass(obj, BeetsPlugin)
-                and obj != BeetsPlugin
-                and obj != MetadataSourcePlugin
+                issubclass(cls, BeetsPlugin)
+                and cls != BeetsPlugin
+                and cls != MetadataSourcePlugin
+                and cls not in _instances
             ):
-                _classes.add(obj)
+                classes.add(cls)
+
+    for cls in classes:
+        _register_plugin(cls)
 
 
-_instances: dict[type[BeetsPlugin], BeetsPlugin] = {}
+def _register_plugin(cls: type[BeetsPlugin]) -> None:
+    if cls in _instances:
+        return
+
+    try:
+        _instances[cls] = cls()
+    except Exception:
+        log.error(
+            "failed to initialize plugin class {}.{}",
+            cls.__module__,
+            cls.__qualname__,
+        )
+        raise
 
 
-def find_plugins() -> list[BeetsPlugin]:
-    """Returns a list of BeetsPlugin subclass instances from all
-    currently loaded beets plugins. Loads the default plugin set
-    first.
+def find_plugins() -> Iterator[BeetsPlugin]:
+    """Returns an iterator of BeetsPlugin subclass instances from all
+    currently loaded beets plugins.
     """
-    if _instances:
-        # After the first call, use cached instances for performance reasons.
-        # See https://github.com/beetbox/beets/pull/3810
-        return list(_instances.values())
-
-    plugins = []
-    for cls in _classes:
-        # Only instantiate each plugin class once.
-        if cls not in _instances:
-            try:
-                _instances[cls] = cls()
-            except Exception:
-                log.error(
-                    "failed to initialize plugin class {}.{}",
-                    cls.__module__,
-                    cls.__qualname__,
-                )
-                raise
-        plugins.append(_instances[cls])
-    return plugins
+    yield from _instances.values()
 
 
 # Communication with plugins.
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -23,6 +23,7 @@ import sys
 import traceback
 from collections import defaultdict
 from functools import wraps
+from importlib import import_module
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -355,10 +356,8 @@ def load_plugins(names: Sequence[str] = ()) -> None:
     BeetsPlugin subclasses desired.
     """
     for name in names:
-        modname = f"{PLUGIN_NAMESPACE}.{name}"
-
         try:
-            namespace = __import__(modname, None, None)
+            mod = import_module(f".{name}", package=PLUGIN_NAMESPACE)
         except ModuleNotFoundError:
             log.warning("** plugin {} not found", name)
             continue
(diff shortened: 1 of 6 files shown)


beets/plugins.py
```diff
@@ -270,31 +270,28 @@ def load_plugins(names=()):
...
-                    if (
-                        isinstance(obj, type)
-                        and issubclass(obj, BeetsPlugin)
-                        and obj != BeetsPlugin
-                        and obj not in _classes
-                    ):
-                        _classes.add(obj)
-
-        except Exception:
+            namespace = __import__(modname, None, None)
```
snejus, line 275: Perhaps we could use the more standard `importlib.import_module`?

beets/plugins.py
```diff
@@ -270,31 +270,28 @@ def load_plugins(names=()):
...
+            continue
+        except ImportError:
             log.warning(
                 "** error loading plugin {}:\n{}",
                 name,
                 traceback.format_exc(),
             )
+            continue
+
+        for obj in getattr(namespace, name).__dict__.values():
```
snejus, line 287: Consider using `inspect.getmembers`

beets/plugins.py
```diff
@@ -270,31 +270,28 @@ def load_plugins(names=()):
...
+            if (
+                isinstance(obj, type)
```
snejus, line 289: ... and `inspect.isclass` here

beets/plugins.py
```diff
@@ -315,7 +312,15 @@ def find_plugins():
...
-            _instances[cls] = cls()
+            try:
+                _instances[cls] = cls()
+            except Exception:
+                log.error(
+                    "failed to initialize plugin class {}.{}",
+                    cls.__module__,
+                    cls.__qualname__,
+                )
+                raise
```
snejus, line 323: Shouldn't we just allow it to die?

beets/plugins.py
wisp3rwind: Good call!

beets/plugins.py
wisp3rwind: :+1:

beets/plugins.py
wisp3rwind: Maybe I mis-read this and thought you were commmenting on another line: This re-raise should lead to beets exiting.
However, it won't exit on import failures (a few lines above).

snejus: I was really unclear here in any case 🤣 

I was meant to say: shall we handle plugin failure gracefully? Ideally, a single plugin failing to initialise should not stop beets execution. I'm thinking of something like

```python
def find_plugins() -> Iterable[BeetsPlugin]:
    """Returns a list of BeetsPlugin subclass instances from all
    currently loaded beets plugins. Loads the default plugin set
    first.
    """
    if not _instances:
        load_plugins()
        for _class in list(_classes):
            try:
                _instances[_class] = _class()
            except Exception:
                log.error(...)
                _classes.remove(_class)

    return _instances.values()
```

wisp3rwind: Interesting, so we actually have completely opposite opinions here :laughing:

In any case; I'd say that things are not really consistent now:
- if a plugin is not found (e.g. because it was pip-installed for Python 3.X, but then a Python update to 3.(X+1) happened) or can't be imported (e.g. because there are syntax errors), we just warn about it
- if a plugin can be imported, but raises on `__init__`, we abort/crash

My reasoning is that some plugin actions are actually tricky to recover from if unexpectedly not run during import. For example `copyartifacts` (although that is broken anyway), `fetchart` with local art files that are in the same folder at the to-be-imported album, and probably more. Others are easier to recover from (say fetching lyrics, where we only have remote sources IIRC).

So, this seems to warrant some discussion, see #5740. Let's keep this PR focused on code cleanup/debugging improvements, and not change behavior.

wisp3rwind:
This should be ready to go now; it's easiest to review commit-by-commit.

beets/plugins.py
```diff
@@ -350,57 +351,55 @@ def load_plugins(names: Sequence[str] = ()) -> None:
...
-                        isinstance(obj, type)
-                        and issubclass(obj, BeetsPlugin)
-                        and obj != BeetsPlugin
-                        and obj != MetadataSourcePlugin
-                        and obj not in _classes
-                    ):
-                        _classes.add(obj)
-
-        except Exception:
+            mod = import_module(f".{name}", package=PLUGIN_NAMESPACE)
```
snejus: This may be slightly misleading in some cases, for example:

```py
# beetsplug/discogs.py
...
import beets.ibrary
```

```sh
$ beet import -LI sel
...
** plugin discogs not found
```

beets/plugins.py
```diff
@@ -350,57 +351,55 @@ def load_plugins(names: Sequence[str] = ()) -> None:
...
+        except ModuleNotFoundError:
```
snejus: Maybe we should catch a generic `Exception` here? Otherwise every other exception will make beets die, for example:

```py
# beetsplug/discogs.py
...
abc
```

```sh
$ beet import -LI sel
...
  File "/home/sarunas/.local/bin/beet", line 5, in <module>
    main()
  File "/home/sarunas/repo/beets/beets/ui/__init__.py", line 1878, in main
    _raw_main(args)
  File "/home/sarunas/repo/beets/beets/ui/__init__.py", line 1853, in _raw_main
  File "/home/sarunas/repo/beets/beets/ui/__init__.py", line 1680, in _setup
    plugins = _load_plugins(options, config)
  File "/home/sarunas/repo/beets/beets/ui/__init__.py", line 1669, in _load_plugins
    plugins.load_plugins(plugin_list)
  File "/home/sarunas/repo/beets/beets/plugins.py", line 358, in load_plugins
    mod = import_module(f".{name}", package=PLUGIN_NAMESPACE)
  File "/home/sarunas/.local/share/pyenv/versions/3.9.20/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/home/sarunas/repo/beets/beetsplug/discogs.py", line 32, in <module>
    abc
NameError: name 'abc' is not defined
```

beets/plugins.py
```diff
@@ -350,57 +351,55 @@ def load_plugins(names: Sequence[str] = ()) -> None:
...
                 name,
                 traceback.format_exc(),
             )
+            continue
 
+        for _name, cls in inspect.getmembers(mod, inspect.isclass):
+            if (
+                issubclass(cls, BeetsPlugin)
+                and cls != BeetsPlugin
+                and cls != MetadataSourcePlugin
```
snejus:
```suggestion
                and cls not in {BeetsPlugin, MetadataSourcePlugin}
```

beets/plugins.py
```diff
@@ -350,57 +351,55 @@ def load_plugins(names: Sequence[str] = ()) -> None:
...
+                and cls not in _instances
```
snejus: Do we need to perform this check? It's unlikely we will come across dupes, and even if we do, `classes` is defined as a set, so no dupes are allowed.

beets/plugins.py
```diff
@@ -350,57 +351,55 @@ def load_plugins(names: Sequence[str] = ()) -> None:
...
 
-_instances: dict[type[BeetsPlugin], BeetsPlugin] = {}
+    for cls in classes:
+        _register_plugin(cls)
+
+
+def _register_plugin(cls: type[BeetsPlugin]) -> None:
+    if cls in _instances:
+        return
+
```
snejus, line 395: I doubt this check is required
```suggestion
```

beets/plugins.py
```diff
@@ -350,57 +351,55 @@ def load_plugins(names: Sequence[str] = ()) -> None:
...
+            continue
 
+        for _name, cls in inspect.getmembers(mod, inspect.isclass):
+            if (
+                issubclass(cls, BeetsPlugin)
+                and cls != BeetsPlugin
+                and cls != MetadataSourcePlugin
+                and cls not in _instances
+            ):
+                classes.add(cls)
```
snejus: Do we need to store classes in this intermediate set? Can't we just `_register_plugin` here?

beets/plugins.py
```diff
@@ -350,57 +351,55 @@ def load_plugins(names: Sequence[str] = ()) -> None:
...
+            cls.__qualname__,
+        )
+        raise
 
 
-def find_plugins() -> list[BeetsPlugin]:
-    """Returns a list of BeetsPlugin subclass instances from all
-    currently loaded beets plugins. Loads the default plugin set
-    first.
+def find_plugins() -> Iterator[BeetsPlugin]:
```
snejus, line 407: Why are we returning an iterator here instead of the list?

wisp3rwind: There is no list, we'd need to explicitly create one from `_instances.values()`. However, none of the callers of `find_plugins` need a list, since all they ever do is iterate over the plugins. So, it seemed more natural to just `yield from` the iterator that we already have.

beets/plugins.py
wisp3rwind: That's a left over from the previous code. Yes, it should be possible to remove the intermediate set.

beets/plugins.py
wisp3rwind: It's definitely also redundant with the check that `_register_plugin` does.

beets/plugins.py
wisp3rwind: :+1:

beets/plugins.py
wisp3rwind: I think this also a left over from old code: We did use to do this repeatedly in the previous version of `find_plugins` (at least before the caching optimization was added).

If anything, it might make sense to log a warning here if a plugin is instantiated repeatedly. (Say if `load_plugins` is called repeatedly with the same plugins, which I don't think should happen.)

beets/plugins.py
wisp3rwind: I think that was actually an unintended change in https://github.com/beetbox/beets/pull/5739/commits/d977b6cd9690b1229ebb71f7f445c62d8e070b4b.

Recall that we disagree on what should happen here: https://github.com/beetbox/beets/issues/5740. I'll restore the warning for now.

beets/plugins.py
wisp3rwind: Oh, that one looks a bit tricky. Ideally it would be possible to split locating and actually importing the module. I'll see what I can do.

wisp3rwind: Ok, can be done by checking `ModuleNotFoundError.name`, I added a test for all error cases.

wisp3rwind:
Should be good to go, but I'll rebase on https://github.com/beetbox/beets/pull/5841 once its merged.

Note that this isn't quite what the title says anymore: Logging wasn't as incomplete as I thought in the context of #5701. Thus, this PR only slightly improves logging, notably by fixing the "not found" error case (detection of which was quite hacky before, and in fact broken), and by logging the fully-qualified path to offending plugin classes.

Additionally, the PR refactors plugin loading, reducing nesting of the code, and actually instantiating plugins in `load_plugins` instead of `find_plugins` (which makes more sense conceptually, and the laziness in the former approach was irrelevant or even detrimental in practice).

beets/plugins.py
```diff
@@ -355,56 +356,59 @@ def load_plugins(names: Sequence[str] = ()) -> None:
...
+                _register_plugin(cls)
+
+
+def _register_plugin(cls: type[BeetsPlugin]) -> None:
+    if cls in _instances:
+        log.debug(
+            "repeated initialization of plugin class {}.{}",
+            cls.__module__,
+            cls.__qualname__,
+        )
```
snejus, line 394:
```suggestion
```

Let's remove this check and ensure uniqueness with

```py
for name in set(names):
...
```

on line 357.

snejus (approved):
Looks good to me. I've re-run the build and added a small comment

semohr:
Do you feel like this was addressed? We had quite some related PRs recently.

We log the plugin name if it raises an error during import (see [here](https://github.com/beetbox/beets/blob/b06f3f6aa6e29ce3bd3f84f18ee3ff0cf5cef470/beets/plugins.py#L392)). We also test if all plugins can be loaded without an error (see [here](https://github.com/beetbox/beets/blob/b06f3f6aa6e29ce3bd3f84f18ee3ff0cf5cef470/test/test_plugins.py#L484)).

snejus:
> Do you feel like this was addressed? We had quite some related PRs recently.
> 
> We log the plugin name if it raises an error during import (see [here](https://github.com/beetbox/beets/blob/b06f3f6aa6e29ce3bd3f84f18ee3ff0cf5cef470/beets/plugins.py#L392)). We also test if all plugins can be loaded without an error (see [here](https://github.com/beetbox/beets/blob/b06f3f6aa6e29ce3bd3f84f18ee3ff0cf5cef470/test/test_plugins.py#L484)).

It seems like most of this has indeed been addressed now. Closing this now, @wisp3rwind if you feel like something's missing, feel free to re-open of course!

Issue #5690, opened 2025-03-25
`beet mod -a` splits a single provided `artists` tag into a multi-tagged `artists` with one `artists` per character


afontenot:
### Problem

I was trying to fix the `artists` tag that was tagged incorrectly on one of my albums. This album has a single artist, so I only needed to specify one. I'm encountering a strange issue, though. When specifying the tag when in album mode, the track metadata is still set (as expected), but the provided artist name gets split. This doesn't happen when setting the tag directly on the track.

Here's an illustration

```sh
$ beet mod -a -w 'brat' artists='Charli XCX'
Modifying 2 albums.
Charli XCX - BRAT
  artists: Charli xcx -> Charli XCX
Charli XCX - Brat and it’s completely different but also still brat
  artists: Charli xcx -> Charli XCX
Really modify, move and write tags? (Yes/no/select) 

$ beet info 'i might say something' | grep ' artists:'
            artists: C; h; a; r; l; i;  ; X; C; X
            artists: C; h; a; r; l; i;  ; X; C; X
            artists: C; h; a; r; l; i;  ; X; C; X
$ beet mod -w 'I might say something stupid' artists='Charli XCX'
Modifying 3 items.
Charli XCX - BRAT - I might say something stupid
  artists: C\␀h\␀a\␀r\␀l\␀i\␀ \␀X\␀C\␀X -> Charli XCX
Charli XCX - Brat and it’s completely different but also still brat - I might say something stupid
  artists: C\␀h\␀a\␀r\␀l\␀i\␀ \␀X\␀C\␀X -> Charli XCX
Charli XCX - Brat and it’s completely different but also still brat - I might say something stupid
  artists: C\␀h\␀a\␀r\␀l\␀i\␀ \␀X\␀C\␀X -> Charli XCX
Really modify, move and write tags? (Yes/no/select) 

$ beet info 'i might say something' | grep ' artists:'
            artists: Charli XCX
            artists: Charli XCX
            artists: Charli XCX
```

### Setup

* OS: Ubuntu 22.04
* Python version: 3.10.12
* beets version: 2.2.0
* Turning off plugins made problem go away (yes/no): no

My configuration (output of `beet config`) is:

```yaml
directory: /mnt/data/audio
# --------------- Main ---------------

library: ~/.config/beets/music.blb
pluginpath: ~/.local/share/beets/plugins

# --------------- Plugins ---------------

plugins: convert copyartifacts fetchart inline permissions replaygain zero info mbsync
[...]

PR #5743, opened 2025-04-18
Adding functionality for using "artists" as album level query

amelimik:
## Description
I have added the ability to query based on the "artists" field at an album level, along with the existing functionality for "albumartists" field. I have also written 2 tests ensuring that this saves correctly and is correctly query-able.
I have done this by adding "artists" with a string type to the Album class fields. 

Addresses Issue #5690, by adding this functionality that user expected, but was previously not supported. 

- [X ] ~Documentation.~ I believe documentation changes are not necessarily required for such a small change, beyond updating the Changelog. Correct me if I'm wrong, but I didn't see a place in the documentation where this small change would feel appropriate to mention. 
- [X ] Changelog. 
- [X ] Tests. Added 2 tests in test_query.py in RelatedQueriesTest class


wisp3rwind:
I'm not sure that this is a good fix to the problem, rather than just papering over it. I'm fairly sure that not having `artists` (and also `artist`: note that with this PR, the situation becomes inconsistent) defined as an album field is intentional.

Please also see my comment at #5690.

amelimik:
Thank you for the feedback @wisp3rwind! I went more in depth in my response on #5690 but will also mention here that if anyone is to continue working on this issue I'd be happy to try my best to help! :) My "quick fix" here is definitely not addressing the root of the issue.

snejus:
Thanks for your time looking at this @amelimik. As @wisp3rwind said, this issue is more complex than it first appears - it touches on the broader interaction between `Item`- and `Album`-level fields, particularly around multi-valued fields like `artists`. While I appreciate the effort here, merging a quick fix risks introducing further inconsistencies without a proper understanding of the underlying design intent. I'm going to close this PR for now, but the discussion in #5690 is a great starting point if you or anyone else wants to take a deeper look at a more comprehensive solution in the future. Thanks again for contributing!

PR #5745, opened 2025-04-20
fixup #5701

wisp3rwind:
which added some mypy config to pyproject.toml, leading to mypy ignoring setup.cfg (this shows up in CI output for #5701, but is not very visible since we currently ignore mypy errors)

Related: https://github.com/beetbox/beets/pull/5728

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -281,12 +281,3 @@ ignore-variadic-names = true
 [tool.ruff.lint.pep8-naming]
 classmethod-decorators = ["cached_classproperty"]
 extend-ignore-names = ["assert*", "cached_classproperty"]
-
-# Temporary, until we decide on a mypy
-# config for all files.
-[[tool.mypy.overrides]]
-module = "beets.plugins"
-disallow_untyped_decorators = true
-disallow_any_generics = true
-check_untyped_defs = true
-allow_redefinition = true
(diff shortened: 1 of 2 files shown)


wisp3rwind:
I'm going ahead and merge this, since this breaks mypy for me locally and should be pretty safe.

Note that we don't have any CI feedback on the change, since the skip rules don't take into account `pyproject.toml` and `setup.cfg` changes for mypy (and linting).

semohr:
Why do you think we don't have feedback in the CI for this? I think mypy should automatically load the configs from the pyproject toml file. Otherwise, if this does not work we can define a config file with the `--config-file` mypy option.

https://github.com/beetbox/beets/blob/43301c4caad638346a05a50e300d72ecdbb7b450/.github/workflows/lint.yml#L89C1-L111C128

semohr:
Thanks for the clarification!

I think only running mypy on changed python files makes sense. One could create an exception in the ci to run the full mypy checks on changes to the pyproject.toml file, but tbh I think this is rarely needed.

I thought we only use a `pyproject.toml` file. I have just seen #5728, which might make this a bit easier if merged! Mypy has an order for config files, see [here](https://mypy.readthedocs.io/en/stable/config_file.html) and `setup.cfg` has the lowest priority.

PR #5761, opened 2025-05-05
Centralize release id extraction

snejus:
## Description

Refactor: Centralize release ID extraction

This change introduces a new utility function `extract_release_id` in `beets.util.id_extractors` to handle the parsing of release IDs (or URLs containing IDs) for various metadata sources (Spotify, Deezer, Beatport, Discogs, MusicBrainz, Bandcamp, Tidal).

Key changes:
- Added `extract_release_id` function and `PATTERN_BY_SOURCE` regex dictionary.
- Converted `MetadataSourcePlugin._get_id` static method to an instance method which uses the `data_source` property to pick the correct id extractor.
- Removed the `id_regex` property and updated `_get_id` calls in all modules.
- Replaced old tests related to ID parsing in individual plugin test files with `test/util/test_id_extractors.py` that tests the `extract_release_id` function.

This refactoring simplifies the codebase, reduces redundancy, and makes it easier to manage and extend ID extraction logic for different sources.

### DIFF
--- a/test/util/test_id_extractors.py
+++ b/test/util/test_id_extractors.py
@@ -1,0 +1,34 @@
+import pytest
+
+from beets.util.id_extractors import extract_release_id
+
+
+@pytest.mark.parametrize(
+    "source, id_string, expected",
+    [
+        ("spotify", "39WqpoPgZxygo6YQjehLJJ", "39WqpoPgZxygo6YQjehLJJ"),
+        ("spotify", "blah blah", None),
+        ("spotify", "https://open.spotify.com/album/39WqpoPgZxygo6YQjehLJJ", "39WqpoPgZxygo6YQjehLJJ"),  # noqa: E501
+        ("deezer", "176356382", "176356382"),
+        ("deezer", "blah blah", None),
+        ("deezer", "https://www.deezer.com/album/176356382", "176356382"),
+        ("beatport", "3089651", "3089651"),
+        ("beatport", "blah blah", None),
+        ("beatport", "https://www.beatport.com/release/album-name/3089651", "3089651"),  # noqa: E501
+        ("discogs", "http://www.discogs.com/G%C3%BCnther-Lause-Meru-Ep/release/4354798", "4354798"),  # noqa: E501
+        ("discogs", "http://www.discogs.com/release/4354798-G%C3%BCnther-Lause-Meru-Ep", "4354798"),  # noqa: E501
+        ("discogs", "http://www.discogs.com/G%C3%BCnther-4354798Lause-Meru-Ep/release/4354798", "4354798"),  # noqa: E501
+        ("discogs", "http://www.discogs.com/release/4354798-G%C3%BCnther-4354798Lause-Meru-Ep/", "4354798"),  # noqa: E501
+        ("discogs", "[r4354798]", "4354798"),
+        ("discogs", "r4354798", "4354798"),
+        ("discogs", "4354798", "4354798"),
+        ("discogs", "yet-another-metadata-provider.org/foo/12345", None),
+        ("discogs", "005b84a0-ecd6-39f1-b2f6-6eb48756b268", None),
+        ("musicbrainz", "28e32c71-1450-463e-92bf-e0a46446fc11", "28e32c71-1450-463e-92bf-e0a46446fc11"),  # noqa: E501
+        ("musicbrainz", "blah blah", None),
+        ("musicbrainz", "https://musicbrainz.org/entity/28e32c71-1450-463e-92bf-e0a46446fc11", "28e32c71-1450-463e-92bf-e0a46446fc11"),  # noqa: E501
+        ("bandcamp", "https://nameofartist.bandcamp.com/album/nameofalbum", "https://nameofartist.bandcamp.com/album/nameofalbum"),  # noqa: E501
+    ],
+)  # fmt: skip
+def test_extract_release_id(source, id_string, expected):
+    assert extract_release_id(source, id_string) == expected
--- a/test/util/test_id_extractors.py
+++ b/test/util/test_id_extractors.py
@@ -1,3 +1,5 @@
+from typing import NamedTuple
+
 import pytest
 
 from beets.util.id_extractors import extract_release_id
@@ -32,3 +34,28 @@ from beets.util.id_extractors import extract_release_id
 )  # fmt: skip
 def test_extract_release_id(source, id_string, expected):
     assert extract_release_id(source, id_string) == expected
+
+
+class SourceWithURL(NamedTuple):
+    source: str
+    url: str
+
+
+source_with_urls = [
+    SourceWithURL("spotify", "https://open.spotify.com/album/39WqpoPgZxygo6YQjehLJJ"),
+    SourceWithURL("deezer", "https://www.deezer.com/album/176356382"),
+    SourceWithURL("beatport", "https://www.beatport.com/release/album-name/3089651"),
+    SourceWithURL("discogs", "http://www.discogs.com/G%C3%BCnther-Lause-Meru-Ep/release/4354798"),
+    SourceWithURL("musicbrainz", "https://musicbrainz.org/entity/28e32c71-1450-463e-92bf-e0a46446fc11"),
+]  # fmt: skip
+
+
+@pytest.mark.parametrize("source", [s.source for s in source_with_urls])
+@pytest.mark.parametrize("source_with_url", source_with_urls)
+def test_match_source_url(source, source_with_url):
+    if source == source_with_url.source:
+        assert extract_release_id(source, source_with_url.url)
+    else:
+        assert not extract_release_id(source, source_with_url.url), (
+            f"Source {source} pattern should not match {source_with_url.source} URL"
+        )
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -37,6 +37,7 @@ import mediafile
 
 import beets
 from beets import logging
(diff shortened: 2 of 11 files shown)


semohr:
Thanks for already implementing this! 

Are we set on the functional approach here? I would prefer if this would be part of the plugins themself (e.g. defined as an required property or method in the abc). Is in my opinion closer to the actual logic and would required anyone who creates a new source plugin to properly register their ids.

I know not all plugins are `SourcePlugins` yet, but we might want to implement the shared parent class first before continuing with this. In my opinion such shared behavior (across all source plugins) is exactly a use case where abstract base classes shine bright.

wisp3rwind:
> Are we set on the functional approach here? I would prefer if this would be part of the plugins themself (e.g. defined as an required property or method in the abc). Is in my opinion closer to the actual logic and would required anyone who creates a new source plugin to properly register their ids.

:+1: I only had a _very_ superficial look at the PR, but I'd agree that this looks like plugins should register their ID patterns. Right now, `beetsplug` and `beets` are unexpectedly intertwined.

semohr:
I just created a small proposal for the base class, how do we feel about this? 

https://gist.github.com/semohr/c1b12b026b51dbb346c4cd869f31ba54

snejus:
In an ideal world, I would also imagine that metadata source plugins have something like what @semohr suggested:

> I would prefer if this would be part of the plugins themself (e.g. defined as an required property or method in the abc)

And they indeed originally had been part of the plugins, however they have been moved out #4633 in order to support saving external IDs by the `musicbrainz` plugin (see the relevant changes in `beetsplug/musicbrainz.py` in this PR).

Arguably, we may be able to remove this ID parsing from `musicbrainz` and instead just save the URLs? The code that will eventually make use of these IDs can parse them from URLs. Currently, we have no functionality that makes use of them.

On the other hand, what do we do about `bandcamp` and `tidal` id regexes? These plugins are not present in the codebase, so how would we parse them if we'd expect the plugins to define them?

wisp3rwind:
> In an ideal world, I would also imagine that metadata source plugins have something like what @semohr suggested:
> 
> > I would prefer if this would be part of the plugins themself (e.g. defined as an required property or method in the abc)
> 
> And they indeed originally had been part of the plugins, however they have been moved out #4633 in order to support saving external IDs by the `musicbrainz` plugin (see the relevant changes in `beetsplug/musicbrainz.py` in this PR).

That looks like a tricky issue indeed, not sure what's the best way to proceed.

> Arguably, we may be able to remove this ID parsing from `musicbrainz` and instead just save the URLs? The code that will eventually make use of these IDs can parse them from URLs. Currently, we have no functionality that makes use of them.
> 
> On the other hand, what do we do about `bandcamp` and `tidal` id regexes? These plugins are not present in the codebase, so how would we parse them if we'd expect the plugins to define them?

I was thinking of something like you say, and like @semohr implements in his proposal (the `to_release_id` function). However, that doesn't quite work for the "external id" use in the musicbrainz source, because the plugins might not be loaded. So, maybe it does make sense to have some id extractors in beets core, that the corresponding built-in plugins as well as the musicbrainz plugin can make use of (but community metadata source plugins can still bring their own ID extraction).

I'm not going to review this PR further; the whole metadata source part of beets is really not my area of expertise. You guys go ahead with what you think is appropriate!

semohr:
I don't think this is too complicated. We could change the call signature of the `self.to_release_id` to something like this:

```python
def to_release_id(self, url: str) -> Tuple[str, str] | None:
    """extract id and source from an url"""
    return [id,source]
```

This would allow simple metadata plugins to just define a regex (as currently proposed). In the mb plugin we could use a more complex logic or even call all other metadata plugins to resolve additional data source urls. Similarly how we call e.g. 
`item_candidates`.

I think the logic of id parsing should not be parts of beets itself if mb is moved to a beets plugin

wisp3rwind:
> In the mb plugin we could use a more complex logic or even call all other metadata plugins to resolve additional data source urls. Similarly how we call e.g. `item_candidates`.

The thing is that calling other metadata plugins would only work if the corresponding plugins are enabled, and will therefore be loaded in `plugins.py`. Thus, for example, the musicbrainz plugin would only be able to  set the discogs ID if the discogs plugin is enabled. I doubt that this would be the desired outcome.

snejus:
I see. See my question from above:

> On the other hand, what do we do about bandcamp and tidal id regexes? These plugins are not present in the codebase, so how would we parse them if we'd expect the plugins to define them?

semohr:
These additional regexes are only used in the musicbrainz plugin, right? If this is the case, it wouldn't be too big of an issue to just have them as part of the mb plugin. 

I'm not super set on this approach, but I generally think the extraction should be closer to the plugins.

semohr:
I had a try at this myself and implemented the extraction on plugin level in an earlier commit of #5764. 

To be honest my approach gets convoluted and incomprehensive quickly, even if working in theory. I would be happy to go with your inital approach for id extraction as it is more understandable and maintainable.

Could we move the extractors utils into the beetsplug folder tho? That way they are close to where the are used. I think that should be a decent compromise.

snejus:
> Could we move the extractors utils into the beetsplug folder tho? That way they are close to where the are used. I think that should be a decent compromise.

The extraction function is called from the generic `MetadataSourcePlugin`, i.e., it's imported in beets core `beets.plugins`. I think the consensus is that beets core does not import from `beetsplug` package since it's for plugins only.

snejus:
> Apologies for the back and forth—this really feels more like a design decision we should get right to ensure it's future-proof than a simple refactor.

No need to apologize! All good 😆 

> The _get_id method is not required by the MetadataSourcePlugin abstract base class

`_get_id` is an internal plugin util rather than a public method which must exist in order to fulfil plugin's designed purpose (like `candidates` etc.). Each metadata source plugin requires it (`musicbrainz` and `discogs` use `extract_release_id` function directly since they don't inherit `MetadataSourcePlugin`). One of the reasons why I want to make these two inherit `MetadataSourcePlugin` is so that they use the unified `_get_id` implementation.

> the Chroma plugin does not use ID extraction at all

Indeed - another reason why `_get_id` shouldn't be abstract. 

I don't think we need to overthink this too much - this method simply follows DRY design and defines shared functionality within the base class.

semohr:
I do not think thought `_get_id` should be abstract. I think it is probably not necessary at all if we define the patterns together in one place 😅

We should really think about moving the extractor utils into the beetsplug folder tho. It is kinda confusing that the regex patterns are defined in beets, because they are not directly used in beets but only by beetsplug i.e. inside specific plugin implementations.

Nonetheless Im still pretty happy with these enhancments as they are a mayor improvment over the previous behaviour. We can still move/remove it in favor of direct imports later.

test/util/test_id_extractors.py
```diff
@@ -0,0 +1,34 @@
...
+        ("discogs", "4354798", "4354798"),
+        ("discogs", "yet-another-metadata-provider.org/foo/12345", None),
+        ("discogs", "005b84a0-ecd6-39f1-b2f6-6eb48756b268", None),
+        ("musicbrainz", "28e32c71-1450-463e-92bf-e0a46446fc11", "28e32c71-1450-463e-92bf-e0a46446fc11"),  # noqa: E501
+        ("musicbrainz", "blah blah", None),
+        ("musicbrainz", "https://musicbrainz.org/entity/28e32c71-1450-463e-92bf-e0a46446fc11", "28e32c71-1450-463e-92bf-e0a46446fc11"),  # noqa: E501
+        ("bandcamp", "https://nameofartist.bandcamp.com/album/nameofalbum", "https://nameofartist.bandcamp.com/album/nameofalbum"),  # noqa: E501
+    ],
+)  # fmt: skip
+def test_extract_release_id(source, id_string, expected):
```
semohr, line 33: I think we should also test the matching without sources. The regexes should only ever match urls from the specific source.

In `[plugins.py](https://github.com/beetbox/beets/blob/2ee5de4767d223b4b9a1dd2d0dfa55b2c028735a/beets/plugins.py#L522)` file we try to find a match over all plugins, there should not be false positives here, as they trigger expensive external lookups.

snejus: Good point! I'll extend the test cases accordingly and rebase this on master to make this independent of #5748 so that we can merge it.

snejus: I was looking at this and realised that `bandcamp` is a bit tricky to test because artists may have custom domains; see [this issue](https://github.com/snejus/beetcamp/issues/12) for context.

Thus, the pattern I use in beetcamp [is very generic](https://github.com/snejus/beetcamp/blob/64c7afc9d87682fb2b7c9f2deb76525e44afb248/beetsplug/bandcamp/__init__.py#L64):

```py
    BANDCAMP_URL_PAT = cached_patternprop(r"http[^ ]+/(album|track)/")
```

which means it will match deezer and spotify URLs. Meanwhile, this pattern is only used in `track_by_id` and `album_by_id` methods to skip 'proper' IDs from other backends as they are never stored as URLs.

semohr: Well that is inconvenient... For now we could just ignore bandcamp in the test the other cases?

I would just iterate overall all patterns and test that any defined input is only matched by exactly pattern.

snejus: That's what I did - see the latest commit

snejus:
Just need your approval @semohr if this looks good

semohr (approved):

semohr:
Looks good to me :+1:  Once we are done with this, It would be awesome to continue with #5787, since it partially depends on this PR.

PR #5764, opened 2025-05-06
Refactor of metadata plugin and opt in all metadata plugins to new baseclass.

semohr:
## Description

At the moment the `MetaDataSourcePlugin` has multiple responsibilities:

- fetch data via `_search` api
- defines contract for interaction within the beets autotag lookup

I propose splitting these responsibilities, as it would enable us to use the `MetaDataSourcePlugin` baseclass with plugins that use external packages to fetch data.
  
This follows from discussion in #5761 and https://github.com/beetbox/beets/pull/5748#discussion_r2075070638.

Feedback is highly appreciated, as this is mainly architectural decision and I would prefer if the new behavior is a shared consensus. 

## To Do

- [x] Opt in plugins into the new `MetaDataSourcePlugin` 
     - [x] Spotify
    - [x] Musicbrainz
    - [x] Deezer
    - [x] Beatport
    - [x] Chroma
    - [x] Disccogs
- [x] Remove old MetaDataSourcePlugin and related functions
- [x] Documentation on the ontology of plugins
- [ ] Changelog


snejus:
Aplogies, didn't realise this existed and I updated my branch with the most recent commit from master. You probably want to update

semohr:
From the logic standpoint this should be ready for review. I will add a section in the documentation and update the change-log once we are happy with the proposed changes.

semohr:
Well I don't think we can reopen this as the source branch is deleted now. I guess I create another PR for this :)

PR #5765, opened 2025-05-07
Update dependencies and address security issues

snejus:
This PR addresses Dependabot vulnerability alerts and secret scanner reports by updating dependencies and redacting some of the sensitive configuration field.

*   **Dependency Updates:** Updated various dependencies, including those flagged by Dependabot, as reflected in `poetry.lock`. This resolves known vulnerabilities.
*   **Secret Redaction:** Addressed secret scanner findings by redacting sensitive configuration values (e.g., `tokenfile`, `apikey`, `password`, `username`, `userid`) in `beatport`, `discogs`, `embyupdate`, `fetchart`, `lastimport`, `spotify`, and `subsonicupdate` plugins.
*   **Python Version:** Standardized on Python 3.9 across GitHub Actions workflows.
*   **Code Style:** Minor adjustments for `ruff` compatibility, including string concatenations and f-string quote consistency.

### DIFF
--- a/test/plugins/lyrics_pages.py
+++ b/test/plugins/lyrics_pages.py
@@ -605,5 +605,6 @@ lyrics_pages = [
         Children at your feet
         Wonder how you manage to make ends meet
         """,
+        marks=[pytest.mark.xfail(reason="Tekstowo seems to be broken again")],
     ),
 ]


beetsplug/beatport.py
```diff
@@ -296,6 +296,7 @@ def __init__(self):
         )
         self.config["apikey"].redact = True
         self.config["apisecret"].redact = True
+        self.config["tokenfile"].redact = True
```
semohr, line 299: The tokenfile is the path to tokens on disk, right? Why does it need to be redacted. I guess it is okay, but I don't get the reasoning behind it.

If I run `beet config`, I want to know where the tokens are place or am I wrong?

semohr (approved):
Everything looks good to me. I just have one open question regarding the redacted token files, but aside from that, it’s ready to merge.

wisp3rwind (commented):
Looks good to me beyond the comments by @semohr.

Regarding the bump to libgirepository2, the comment I left in https://github.com/beetbox/beets/pull/5728/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R116-R121 might be interesting (in general, it appears that we don't document that dependency at all?).

FWIW, I've been working on initial typing support for confuse a few weeks ago. With those merged, many of the annotations added to remove the explicit `cast`s become unnecessary (and there would be some actual type checking). I've not opened a PR for them yet, but two preparatory cleanups, see https://github.com/beetbox/confuse/pulls. Would be happy about a review there.

beets/dbcore/types.py
```diff
@@ -79,7 +79,7 @@ def null(self) -> N:
         # Note that this default implementation only makes sense for T = N.
         # It would be better to implement `null()` only in subclasses, or
         # have a field null_type similar to `model_type` and use that here.
-        return cast(N, self.model_type())
+        return cast("N", self.model_type())
```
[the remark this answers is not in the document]
snejus: See `ruff rule TS006`:

(actually, now that I'm reading it again, you are completely right - we've got the annotations import so it's not relevant to us. I should instead just disable this rule.)

> # runtime-cast-value (TC006)
> 
> Derived from the **flake8-type-checking** linter.
> 
> Fix is always available.
> 
> ## What it does
> Checks for unquoted type expressions in `typing.cast()` calls.
> 
> ## Why is this bad?
> This rule helps enforce a consistent style across your codebase.
> 
> It's often necessary to quote the first argument passed to `cast()`,
> as type expressions can involve forward references, or references
> to symbols which are only imported in `typing.TYPE_CHECKING` blocks.
> This can lead to a visual inconsistency across different `cast()` calls,
> where some type expressions are quoted but others are not. By enabling
> this rule, you ensure that all type expressions passed to `cast()` are
> quoted, enforcing stylistic consistency across all of your `cast()` calls.
> 
> In some cases where `cast()` is used in a hot loop, this rule may also
> help avoid overhead from repeatedly evaluating complex type expressions at
> runtime.
> 
> ## Example
> ```python
> from typing import cast
> 
> x = cast(dict[str, int], foo)
> ```
> 
> Use instead:
> ```python
> from typing import cast
> 
> x = cast("dict[str, int]", foo)
> ```
> 
> ## Fix safety
> This fix is safe as long as the type expression doesn't span multiple
> lines and includes comments on any of the lines apart from the last one.

beetsplug/beatport.py
snejus: If I put myself in an attacker's shoes, I could gain some sort of advantage by knowing where sensitive files reside on the target system? 

This may be an overkill though 😁

semohr: Honestly, it's not much of an advantage. In most cases, you'd check the default paths for sensitive files anyway—since the vast majority of people never change them. So revealing the file's location, which is likely just the default, doesn’t seem like a major risk to me.

If your system’s already compromised, you've probably got bigger problems than someone finding your Beatport or Spotify token files :rofl:

snejus: Have unapplied my change 😅

PR #5768, opened 2025-05-07
Add two new plugin events for recommendation and assign_items

lukeIam:
## Description

I wanted to be able to influence

1. recommendation behavior
2. Item to TrackInfo matching

from a plugin.
So I added those to events to the beets code.

Please let me know if you are generally interested in merging this to beets.
Then I will update the docs and changelog...

More background:
I'm using beets for sorting audiobooks using audible data. But if there is no clean match beets mixes the order of my tracks.
With that plugin event I'm able to control the matching and make sure that the order is keeps the same.

I plan to use the recommendation event to downgrade a strong recommendation if the next candidate is closer than a threshold.

## To Do

### DIFF
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -390,6 +390,11 @@ def _recommendation(
         rec = Recommendation.low
     else:
         # No conclusion. Return immediately. Can't be downgraded any further.
+        # but allow plugins to modify the recommendation
+        new_rec_list = plugins.send("recommendation", results=results, org_rec=Recommendation.none)
+        # take the first recommendation from a plugin
+        if len(new_rec_list) > 0 and isinstance(new_rec_list[0], Recommendation):
+            return new_rec_list[0]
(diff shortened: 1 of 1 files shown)


lukeIam:
example plugin: https://gist.github.com/lukeIam/d371d5f8ac53e645b6d46ece3649d1d0

beets/autotag/match.py
```diff
@@ -411,6 +435,11 @@ def _recommendation(
             )
             rec = min(rec, max_rec)
 
+    # allow plugins to modify the recommendation
```
snejus, line 438: Seems like both of these blocks are nearly identical. Consider defining a function for this logic.

snejus (changes requested):
Thanks for this. Please add a note to the changelog and add these events to the documentation.

snejus:
@lukeIam are you planning to come back to this by any chance?

PR #5775, opened 2025-05-11
Move queries and types to respective modules

snejus:
This PR moves query and type definitions away from `library.py` to `dbcore` to improve modularity and organization.

**Key Changes:**

*   **Query and Type Relocation:**
    *   `PathQuery` and `SingletonQuery` moved from `beets.library` to `beets.dbcore.query`.
    *   `DateType`, `PathType` (and its variants `NullPathType`), `MusicalKey`, and `DurationType` moved from `beets.library` to `beets.dbcore.types`.
    *   The `BLOB_TYPE` definition was moved from `beets.library` to `beets.dbcore.query` and then referenced in `beets.dbcore.types`.
    *   The `human_seconds_short` utility function was moved from `beets.ui` to `beets.util` due to circular dependency.
*   **Test Modernization:**
    *   The `PathQueryTest` class in `test/test_query.py` has been rewritten to use `pytest.mark.parametrize` for more concise and readable test cases.
*   **Import Updates:** All internal references to these moved classes and functions have been updated across the codebase.

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -40,6 +40,8 @@ else:
 # To use the SQLite "blob" type, it doesn't suffice to provide a byte
 # string; SQLite treats that as encoded text. Wrapping it in a
 # `memoryview` tells it that we actually mean non-text data.
+# needs to be defined in here due to circular import.
+# TODO: remove it from this module and define it in dbcore/types.py instead
 BLOB_TYPE = memoryview
 
 
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -283,26 +283,20 @@ class PathQuery(FieldQuery[bytes]):
     and case-sensitive otherwise.
     """
 
-    def __init__(self, field, pattern, fast=True, case_sensitive=None):
+    def __init__(self, field, pattern, fast=True):
         """Create a path query.
 
         `pattern` must be a path, either to a file or a directory.
-
-        `case_sensitive` can be a bool or `None`, indicating that the
-        behavior should depend on the filesystem.
         """
         super().__init__(field, pattern, fast)
 
         path = util.normpath(pattern)
 
-        # By default, the case sensitivity depends on the filesystem
-        # that the query path is located on.
-        if case_sensitive is None:
-            case_sensitive = util.case_sensitive(path)
-        self.case_sensitive = case_sensitive
+        # Case sensitivity depends on the filesystem that the query path is located on.
+        self.case_sensitive = util.case_sensitive(path)
 
         # Use a normalized-case pattern for case-insensitive matches.
-        if not case_sensitive:
+        if not self.case_sensitive:
             # We need to lowercase the entire path, not just the pattern.
             # In particular, on Windows, the drive letter is otherwise not
             # lowercased.
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -16,6 +16,7 @@
 
 from __future__ import annotations
 
+import os
 import re
 import unicodedata
 from abc import ABC, abstractmethod
@@ -35,6 +36,11 @@ if TYPE_CHECKING:
     P = TypeVar("P", default=Any)
 else:
     P = TypeVar("P")
+
+# To use the SQLite "blob" type, it doesn't suffice to provide a byte
+# string; SQLite treats that as encoded text. Wrapping it in a
+# `memoryview` tells it that we actually mean non-text data.
+BLOB_TYPE = memoryview
 
 
 class ParsingError(ValueError):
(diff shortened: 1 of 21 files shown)


beets/util/__init__.py
```diff
@@ -987,6 +987,14 @@ def raw_seconds_short(string: str) -> float:
     return float(minutes * 60 + seconds)
 
 
+def human_seconds_short(interval):
```
semohr: Comment (no need to fix now): 
Interesting that this function exists like it is, I think using a [timedelta](https://docs.python.org/3/library/datetime.html#datetime.timedelta) is preferable. Also proper formatting for hours would be nice.

beets/dbcore/query.py
```diff
@@ -328,22 +322,20 @@ def is_path_query(cls, query_part):
...
             query_part = "(BYTELOWER({0}) = BYTELOWER(?)) || \
                          (substr(BYTELOWER({0}), 1, ?) = BYTELOWER(?))"
 
+        dir_blob = BLOB_TYPE(os.path.join(self.pattern, b""))
         return query_part.format(self.field), (
-            file_blob,
+            BLOB_TYPE(self.pattern),
             len(dir_blob),
             dir_blob,
         )
```
wisp3rwind, line 341: If I'm not mistaken, the SQL predicate and the `match()` function have slightly different behavior now, since `match` changed.

For example, think of matching `PathQuery("/foo/bar")` against a model with

- `path = "/foo/bar" (matches for old and new code)
- `path = "/foo/bar/"  (matches for old and new code)
- `path = "/foo/barbaz"  (matches only with new code)

I didn't check when and with which rationale the `file_path` and `dir_path` checks were added. Probably, the exact behavior doesn't matter too much; the new one might even be less surprising. If that's what we want to keep, the SQL needs to be adjusted to perform just the equivalent of `.startswith`.

wisp3rwind (commented):
Not a thorough review, just a few initial comments.

In general, this PR feels too me like it starts to erode the separation between high-level `library` code and low-level `dbcore` a little (Consider for example the `MusicalKey` type). I think @sampsyo mentioned somewhere that the original intent was to keep the latter as a fully re-usable and generic module, without any specifics for music libraries. Usage of `dbcore` for anything but `Item` and `Album` never happened, though.

I haven't really formed a clear opinion yet on where the ideal abstraction boundary lies here.

beets/dbcore/query.py
```diff
@@ -304,13 +301,10 @@ def __init__(self, field, pattern, fast=True):
...
-        # Match the path as a single file.
-        self.file_path = path
-        # As a directory (prefix).
-        self.dir_path = os.path.join(path, b"")
+        super().__init__(field, path, fast)
 
-    @classmethod
-    def is_path_query(cls, query_part):
+    @staticmethod
+    def is_path_query(query_part: str) -> bool:
```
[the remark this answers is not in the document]
wisp3rwind: nevermind, I didn't read that line properly

beets/dbcore/query.py
snejus: I think this keeps the previous logic in tact:

Previously
```py
    def __init__(...):
        super().__init__(field, pattern, fast)
        path = util.normpath(pattern)
        ...
        # Match the path as a single file.
        self.file_path = path
        # As a directory (prefix).
        self.dir_path = os.path.join(path, b"")

    def col_clause(self):
        file_blob = BLOB_TYPE(self.file_path)
        dir_blob = BLOB_TYPE(self.dir_path)
        ...
        return query_part.format(self.field), (
            file_blob,
            len(dir_blob),
            dir_blob,
        )
```

Now
```py
    def __init__(...):
        path = util.normpath(pattern)
        ...
        super().__init__(field, path, fast)

    def col_clause(self):
        ...
        dir_blob = BLOB_TYPE(os.path.join(self.pattern, b""))
        return query_part.format(self.field), (
            BLOB_TYPE(self.pattern),
            len(dir_blob),
            dir_blob,
        )
```

Instead of assigning `self.file_path` and `self.dir_path` this now initialises the object with the normalized `path` where `dir_path` is calculated directly for the query construction:

```py
        dir_blob = BLOB_TYPE(os.path.join(self.pattern, b""))
```

beets/dbcore/types.py
```diff
@@ -388,15 +372,30 @@ def normalize(self, value):
         else:
             return value
 
-    def from_sql(self, sql_value):
-        return self.normalize(sql_value)
```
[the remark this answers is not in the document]
snejus: Nice catch, I don't remember this being intentional 👀 

Brought it back

beets/dbcore/query.py
wisp3rwind: Yes, I agree, `col_clause`  seems to be unchanged.

However, given my understanding, `col_clause` and `match` should result in the same test: Once in SQL, and once in Python. I think that's not the case anymore, since the SQL still encodes the expression that `match` (which is changed now) used to have.

snejus: Ah, I see what you mean... I completely misunderstood you, even your message was clear, now that I read it again 🤦🏼‍♂️ 

I simplified `match` check seeing that it essentially did

```py
    return obj.path == self.pattern or obj.path.startswith(self.pattern + "b/")
```

which seemed to me functionally equivalent to

```py
    return obj.path.startswith(self.pattern)
```

thus I don't think I adjusted the behaviour?

I should simplify the SQL query accordingly too, though.

snejus: Thanks to a failing test in `test_web.py` I figured that I'm completely dumb.

These are not at all equivalent 😁 

> I simplified `match` check seeing that it essentially did
> 
> ```python
>     return obj.path == self.pattern or obj.path.startswith(self.pattern + "b/")
> ```
> 
> which seemed to me functionally equivalent to
> 
> ```python
>     return obj.path.startswith(self.pattern)
> ```

beets/dbcore/types.py
```diff
@@ -312,14 +312,145 @@ class Boolean(Type):
...
 
     def format(self, value: bool) -> str:
         return str(bool(value))
 
     def parse(self, string: str) -> bool:
-        return str2bool(string)
+        return util.str2bool(string)
+
+
+class DateType(Float):
```
semohr, line 326: What is our current approach for Time-zones? SQLite does not have a proper datetime format and every date is normally written as numeric (float). Do we always write UTC timestamps?

If yes, time.localtime would return the wrong times here.

I know it is just a move and no real changes here just out of interest.

beets/ui/commands.py
```diff
@@ -795,7 +795,7 @@ def summarize_items(items, singleton):
                 round(int(items[0].samplerate) / 1000, 1), items[0].bitdepth
             )
             summary_parts.append(sample_bits)
-        summary_parts.append(ui.human_seconds_short(total_duration))
+        summary_parts.append(util.human_seconds_short(total_duration))
         summary_parts.append(ui.human_bytes(total_filesize))
```
semohr: Seems a bit inconsistent now, that the humanize functions aren't placed in the same file anymore. Could we move all humanize functions into a shared file?
I think `util/units.py` should be appropriate.

semohr (changes requested):

beets/util/__init__.py
snejus: agreed

beets/dbcore/types.py
snejus: We don't handle timezones and do not do any conversion between the timezones, I think

beets/ui/commands.py
snejus: Have moved them under `beets.util.units` as suggested

semohr:
Thank you for the PR! I really appreciate it, this will make our lives much easier in the long run!

PR #5782, opened 2025-05-14
Pin poetry version below <2

### LINKED ISSUE #5770: 2.3.0 PyPi tarball mangling mtime on files
### Problem

When trying to run the tests:
```
______________________________ RmTempTest.test_rm ______________________________

self = <test.test_importer.RmTempTest testMethod=test_rm>

    def test_rm(self):
>       zip_path = create_archive(self)

test/test_importer.py:259: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
test/test_importer.py:241: in create_archive
    archive.write(syspath(os.path.join(_common.RSRC, b"full.mp3")), "full.mp3")
/usr/lib/python3.13/zipfile/__init__.py:1889: in write
    zinfo = ZipInfo.from_file(filename, arcname,
/usr/lib/python3.13/zipfile/__init__.py:596: in from_file
    zinfo = cls(arcname, date_time)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <[AttributeError("'zipfile.ZipInfo' object has no attribute 'compress_type'") raised in repr()] ZipInfo object at 0x7f3b2c2717d0>
filename = 'full.mp3', date_time = (1970, 1, 1, 0, 0, 0)

    def __init__(self, filename="NoName", date_time=(1980,1,1,0,0,0)):
        self.orig_filename = filename   # Original file name in archive
    
        # Terminate the file name at the first null byte and
        # ensure paths always use forward slashes as the directory separator.
        filename = _sanitize_filename(filename)
    
        self.filename = filename        # Normalized file name
        self.date_time = date_time      # year, month, day, hour, min, sec
    
        if date_time[0] < 1980:
>           raise ValueError('ZIP does not support timestamps before 1980')
E           ValueError: ZIP does not support timestamps before 1980

/usr/lib/python3.13/zipfile/__init__.py:412: ValueError
```

Looking at the file:
```
$ stat test/rsrc/full.mp3                                                                                                                                                                                          
  File: test/rsrc/full.mp3
  Size: 12820     	Blocks: 33         IO Block: 13312  regular file
Device: 0,30	Inode: 9091031     Links: 1
Access: (0644/-rw-r--r--)  Uid: ( 1000/    joel)   Gid: ( 1000/    joel)
Access: 2025-05-08 08:58:23.524990591 -0400
Modify: 1969-12-31 19:00:00.000000000 -0500
Change: 2025-05-08 08:58:23.524990591 -0400
 Birth: 2025-05-08 08:58:23.524990591 -0400
```

### Setup

* OS: Void Linux
* Python version: 3.13.3
* beets version: 2.3.0

snejus:
Force `poetry` version below 2 to avoid it mangling file modification times in `sdist`.

Fixes #5770


wisp3rwind (approved):
Looks good to me 👍 I think we should eventually figure out how to deal with poetry 2, but this seems like a reasonable workaround to quickly get a fixed tarball out.

semohr (approved):

semohr:
I agree that we should keep track of this. Could you create an issue for upgrading to poetry >= 2, so we don’t lose sight of it? Ideally, include a brief note on why the upgrade is not feasible at this point.

snejus:
Good point, see https://github.com/beetbox/beets/issues/5783

Issue #5695, opened 2025-03-27
typing_extensions requirement inconsistency


minusf:
It is claimed in the documentation and in `pyproject.toml` that `typing_extensions` is needed only for `python>=3.10`. However if plugins are enabled, beets crashes with `ModuleNotFoundError: No module named 'typing_extensions'`.

The plugins functionality is trying to import `typing_extensions` unconditionally.

I think it would be more user friendly and practical to have the same requirements as the official plugins. (`requests` is a similar plugin hidden dependency that needs to be added manually to a newly installed beets)

Now every time I update the `venv` or `uv tool upgrade`, I have to manually readd `typing_extensions` as it's being removed.

Would it be possible to either only import it conditionally, or declaring it in pyproject.toml so it's not being removed with every update?

Here is the string search for `typing_extensions` in the repo:
https://github.com/search?q=repo%3Abeetbox%2Fbeets%20typing_extensions&type=code

### Problem

Running this command in verbose (`-vv`) mode:

```sh
$ beet
```

Led to this problem:

```
...
** error loading plugin edit:
Traceback (most recent call last):
  File "/Users/minusf/src/tmp/beetspip/venv/lib/python3.12/site-packages/beets/plugins.py", line 268, in load_plugins
    namespace = __import__(modname, None, None)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/minusf/src/tmp/beetspip/venv/lib/python3.12/site-packages/beetsplug/edit.py", line 27, in <module>
    from beets.importer import action
  File "/Users/minusf/src/tmp/beetspip/venv/lib/python3.12/site-packages/beets/importer.py", line 37, in <module>
    from beets.util import (
  File "/Users/minusf/src/tmp/beetspip/venv/lib/python3.12/site-packages/beets/util/pipeline.py", line 41, in <module>
    from typing_extensions import TypeVar, TypeVarTuple, Unpack
ModuleNotFoundError: No module named 'typing_extensions'
```

### Setup

* OS: macos
* Python version: 3.12.9
* beets version: 2.2.0 (https://github.com/beetbox/beets.git)
* Turning off plugins mode problem go away (yes/no): yes

My configuration (output of `beet config`) is:

```yaml
(paste here)
```

wisp3rwind:
That's a regression due to https://github.com/beetbox/beets/pull/5611

semohr:
[...]

PR #5787, opened 2025-05-17
Refactor of metadata plugin and opt in all metadata plugins to new baseclass

semohr:
## Description

At the moment the `MetaDataSourcePlugin` has multiple responsibilities:

- fetch data via `_search` api
- defines contract for interaction within the beets autotag lookup

I propose splitting these responsibilities, as it would enable us to use the `MetaDataSourcePlugin` baseclass with plugins that use external packages to fetch data.
  
This follows from discussion in #5761 and https://github.com/beetbox/beets/pull/5748#discussion_r2075070638.

Feedback is highly appreciated, as this is mainly architectural decision and I would prefer if the new behavior is a shared consensus. 

## To Do

- [x] Opt in plugins into the new `MetaDataSourcePlugin` 
     - [x] Spotify
    - [x] Musicbrainz
    - [x] Deezer
    - [x] Beatport
    - [x] Chroma
    - [x] Disccogs
- [x] Remove old MetaDataSourcePlugin and related functions
- [x] Documentation on the ontology of plugins
- [ ] Changelog

This PR was initially #5764 and was accidentally closed as the target branch was deleted. Wasn't able to recover the original PR.


snejus:
You may want to consider updating your (dependent) branches with `git rebase` + force pushes rather than `git merge`. GitHub seems to support it a bit better, for example, see these events under https://github.com/beetbox/beets/pull/5761: it automatically updated the base branch from `make-musicbrainz-a-plugin` to `master` at the point `make-musicbrainz-a-plugin` got merged to `master`. 

image

semohr:
> You may want to consider updating your (dependent) branches with `git rebase` + force pushes rather than `git merge`. GitHub seems to support it a bit better, for example, see these events under #5761: it automatically updated the base branch from `make-musicbrainz-a-plugin` to `master` at the point `make-musicbrainz-a-plugin` got merged to `master`.
> 
> image

Was already on it, before I have seen this :+1: Should be way cleaner now.

beets/metadata_plugins.py
```diff
@@ -0,0 +1,365 @@
...
+    # ---------------------------------- search ---------------------------------- #
+
+    @abc.abstractmethod
+    def candidates(
+        self,
+        items: Sequence[Item],
+        artist: str,
+        album: str,
+        va_likely: bool,
+        extra_tags: dict[str, Any] | None = None,
```
snejus, line 188: Note I removed `extra_tags` from all signatures in this PR: https://github.com/beetbox/beets/pull/5789/commits/d47558b2465160801129c3cf13d1e3a12044cdce as it's unused

semohr: Interesting, I already thought they are mostly unused. Want me to remove it here too or do we want to prioritize #5789 first?

semohr: As we merged the PR #5789 into master, I now removed all extra_tags in this PR too.

beets/metadata_plugins.py
```diff
@@ -0,0 +1,352 @@
...
+    """Get TrackInfo object for the given ID string.
+
+    A single ID can yield just a single track, so we return the first match.
+    """
+    for plugin in find_metadata_source_plugins():
+        if info := plugin.track_for_id(_id):
+            send("trackinfo_received", info=info)
+            return info
+
+    return None
```
snejus, line 77:
```suggestion
        yield from plugins.tracks_for_ids([_id])
```

I'd say we can move the logging message under `tracks_for_ids`:
1. Reuse `None` filtering logic in there
2. Gives an opportunity to include the plugin name in the log message (which I've personally been missing).
3. Should be able to drop `notify_info_yielded` from around `def candidates/item_candidates` above

beets/metadata_plugins.py
```diff
@@ -0,0 +1,352 @@
...
+    artist: NotRequired[str]
+    album: NotRequired[str]
+
+
+R = TypeVar("R", bound=IDResponse)
+
+
+class SearchApiMetadataSourcePlugin(
+    Generic[R], MetadataSourcePlugin, metaclass=abc.ABCMeta
+):
```
snejus, line 245: Nice, I like the way you solved the issue of `_search_api` not being required for discogs and musicbrainz.

beets/metadata_plugins.py
```diff
@@ -0,0 +1,352 @@
...
+        album: str,
+        va_likely: bool,
+    ) -> Iterable[AlbumInfo]:
+        query_filters: SearchFilter = {"album": album}
+        if not va_likely:
+            query_filters["artist"] = artist
+
+        results = self._search_api("album", query_filters)
+        if not results:
+            return
```
snejus, line 285: Is this required?

beets/metadata_plugins.py
```diff
@@ -0,0 +1,352 @@
...
+
+        yield from filter(
+            None, self.tracks_for_ids([result["id"] for result in results])
+        )
+
+
+def artists_to_artist_str(
+    artists,
```
snejus, line 304: Missing type

beetsplug/deezer.py
```diff
@@ -61,31 +69,20 @@ def func(lib, opts, args):
...
-        if "error" in data:
-            self._log.debug("Deezer API error: {}", data["error"]["message"])
-            return None
-        return data
+    # --------------------------------- id lookup -------------------------------- #
 
     def album_for_id(self, album_id: str) -> AlbumInfo | None:
         """Fetch an album by its Deezer ID or URL."""
-        if not (deezer_id := self._get_id(album_id)):
+        if not (deezer_album_id := extract_release_id("deezer", album_id)):
```
snejus, line 76: Why has this been decentralized?

beetsplug/deezer.py
```diff
@@ -45,7 +53,7 @@ class DeezerPlugin(MetadataSourcePlugin, BeetsPlugin):
     track_url = "https://api.deezer.com/track/"
 
     def __init__(self):
-        super().__init__()
+        super().__init__("Deezer")
```
snejus, line 56: As I mentioned above, we can resolve this centrally from the plugin/class name.

snejus (changes requested):
Good stuff! I added a couple of comments in the interfaces, however I'm unable to see how this affects the plugins because they have also been refactored.

Could you move optional changes to a separate PR to minimize any side effects this change may cause? I tried testing it and I found spotify stuck in some authentication exception loop, for example:

```
...done.
Success. Distance: 0.58
Candidate: DRS - Mid Mic Crisis (6817507)
Computing track assignment...
...done.
Success. Distance: 0.58
spotify: Searching Spotify for 'album:Skeptical Answers artist:Vulkanski'
spotify: Spotify access token has expired. Reauthenticating.
spotify: Spotify access token has expired. Reauthenticating.
spotify: Spotify access token has expired. Reauthenticating.
spotify: Spotify access token has expired. Reauthenticating.
spotify: Spotify access token has expired. Reauthenticating.
spotify: Spotify access token has expired. Reauthenticating.
spotify: Spotify access token has expired. Reauthenticating.
spotify: Spotify access token has expired. Reauthenticating.
spotify: Spotify access token has expired. Reauthenticating.
spotify: Spotify access token has expired. Reauthenticating.
spotify: Spotify access token has expired. Reauthenticating.
```

beets/metadata_plugins.py
```diff
@@ -0,0 +1,352 @@
...
+    This base class implements a contract for plugins that provide metadata
+    from a specific source. The plugin must implement the methods to search for albums
+    and tracks, and to retrieve album and track information by ID.
+    """
+
+    data_source: str
+
+    def __init__(self, data_source: str, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self.data_source = data_source or self.__class__.__name__
```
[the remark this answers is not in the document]
semohr: We had a similar logic beforehand. The issue is, that if `data_source` is just a property you can't properly enforce plugins to define them. Having the `data_source` as a parameter in the constructor solves that and requires child classes to define it.

We can still make it a getter if you feel like it should not be dynamic, i.e.:

```python
class Plug():
    def __init__(self, src):
        self._data_source = src 

   @property
   def data_source(self):
        return self._data_source
```

beets/metadata_plugins.py
```diff
@@ -0,0 +1,352 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from confuse import ConfigView
+
+    from .autotag import Distance
+    from .autotag.hooks import AlbumInfo, Item, TrackInfo
+
+
+def find_metadata_source_plugins() -> list[MetadataSourcePlugin]:
```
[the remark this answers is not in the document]
semohr: I wasn't really sure where to place all of the code, tbh. We could move both the default plugin and metadataplugin source into a shared folder, would make importing a bit easier.

Maybe something like this could work:
```shell
...
plugins
- __init__.py
- metadata_plugin.py
- beets_plugin.py
```

beets/metadata_plugins.py
semohr: We could do this, I see the following issue: in theory the plugin implementation can overwrite these functions and ensuring that the send is always called will be more difficult (calling by_id vs by_ids). Also it might be introduce duplication of code.

We can always include the plugin name in the send, something like this could also work:

```python
        if info := plugin.album_for_id(album_id=_id):
            send("albuminfo_received", info=info, data_source=plugin.data_source)
            return info
```

beets/metadata_plugins.py
```diff
@@ -0,0 +1,352 @@
...
+    def albums_for_ids(self, ids: Sequence[str]) -> Iterable[AlbumInfo | None]:
+        """Batch lookup of album metadata for a list of album IDs.
+
+        Given a list of album identifiers, yields corresponding AlbumInfo
+        objects. Missing albums result in None values in the output iterator.
+        Plugins may implement this for optimized batched lookups instead of
+        single calls to album_for_id.
+        """
+
+        return iter(self.album_for_id(id) for id in ids)
```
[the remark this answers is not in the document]
semohr: I agree with point 1, but I disagree with point 2.

If you're retrieving resources by IDs and pass an array of length 3, I would expect to get back an array of the same length and in the same order. That makes sense to me from a UX perspective. I'd also prefer to keep None values, as they preserve the original indexing and make it easier to reason about the results.

beetsplug/deezer.py
snejus: But... It's imported in 5 modules and used twice in each, where the source name argument is hardcoded in every call?

```
  metadatasource-cleanup│?17│ 205 ⮞  grep -RE -e 'import extract_release_id' -e 'extract_release_id\(' beets*                                            beets 2.3.1 │ 🐍 3.9 🐶
beets/util/id_extractors.py:def extract_release_id(source: str, id_: str) -> str | None:
beetsplug/beatport.py:from beets.util.id_extractors import extract_release_id
beetsplug/beatport.py:        if not (release_id := extract_release_id("beatport", album_id)):
beetsplug/deezer.py:from beets.util.id_extractors import extract_release_id
beetsplug/deezer.py:        if not (deezer_album_id := extract_release_id("deezer", album_id)):
beetsplug/deezer.py:        if not (deezer_id := extract_release_id("deezer", track_id)):
beetsplug/discogs.py:from beets.util.id_extractors import extract_release_id
beetsplug/discogs.py:        discogs_id = extract_release_id("discogs", album_id)
beetsplug/discogs.py:        discogs_albumid = extract_release_id("discogs", result.data.get("uri"))
beetsplug/musicbrainz.py:from beets.util.id_extractors import extract_release_id
beetsplug/musicbrainz.py:                    info, f"{source}_album_id", extract_release_id(source, url)
beetsplug/musicbrainz.py:        if not (albumid := extract_release_id("musicbrainz", album_id)):
beetsplug/musicbrainz.py:        if not (trackid := extract_release_id("musicbrainz", track_id)):
beetsplug/spotify.py:from beets.util.id_extractors import extract_release_id
beetsplug/spotify.py:        if not (spotify_album_id := extract_release_id("spotify", album_id)):
b
```

beets/metadata_plugins.py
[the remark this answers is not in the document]
semohr: Yeah that should work in most cases. We might have issues with special characters down the line and inheritance will be tricky. 

Am a bit more in favor of using the constructor as this makes the process very explicit which I like. If you have strong feeling tho we can change it. I don't mind either approach too much.

beetsplug/deezer.py
semohr: Fair enough! I suppose adding a method to MetadataSourcePlugin wouldn’t hurt too much. I’m just a bit cautious about adding too many features to the base class—that’s one of the reasons it became cluttered and hard to maintain in the first place.

beets/metadata_plugins.py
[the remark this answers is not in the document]
snejus: I would strongly prefer my suggestion both as a maintainer and a plugin developer. 

For example, I've been testing `DiscogsPlugin` loads, and this line is in my muscle memory by now:

```py
d = DiscogsPlugin()
```

thus having to provide `"discogs"` as a param just seems unnecessary.

When I developed `beetcamp`, I set `data_source = "bandcamp"`, and only later I discovered that the rest of metadata source plugins use capital letters. I wish that I never had control over this (and be bothered about the inconsistency) and I could just focus on the core logic.

beets/metadata_plugins.py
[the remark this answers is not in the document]
snejus: Fair, I can indeed see a use case where this can be useful.

beets/metadata_plugins.py
```diff
@@ -0,0 +1,352 @@
...
+        plugin
+        for plugin in find_plugins()
+        if isinstance(plugin, MetadataSourcePlugin)
+    ]
+
+
+@notify_info_yielded("albuminfo_received")
+def candidates(*args, **kwargs) -> Iterable[AlbumInfo]:
+    """Return matching album candidates by using all metadata source
+    plugins."""
```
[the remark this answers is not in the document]
snejus: I have always been fond of PEP 257! And yes, I do think google / numpy formats are too verbose and mostly just useless 😁 I have a custom CopilotChat prompt for documentation:

> You are an expert programmer focused on writing maintainable documentation. Your goal is to explain the *purpose* and *behavior* of the code, not its specific implementation details, making the documentation resilient to future refactoring. Focus on *why* the code exists and *what* it achieves at a high level. Strictly avoid duplicating information already present in the code signature, such as parameter names, types, or return types. Assume type hints or clear naming conventions convey this information. Produce concise, well-formatted text that is easy to read and understand. Use PEP 247 conventions for Python docstrings.

beets/plugins.py
```diff
@@ -367,7 +297,7 @@ def load_plugins(names: Sequence[str] = ()) -> None:
                         isinstance(obj, type)
                         and issubclass(obj, BeetsPlugin)
                         and obj != BeetsPlugin
-                        and obj != MetadataSourcePlugin
```
snejus, line 370: I'm seeing this

```
** error loading plugin importreplace:
  File "/home/sarunas/repo/beets/beets/plugins.py", line 298, in load_plugins
    and issubclass(obj, BeetsPlugin)
  File "/home/sarunas/.local/share/pyenv/versions/3.9.20/lib/python3.9/abc.py", line 123, in __subclasscheck__
    return _abc_subclasscheck(cls, subclass)
TypeError: issubclass() arg 1 must be a class
```

beets/metadata_plugins.py
```diff
@@ -0,0 +1,348 @@
...
+        Plugins may implement this for optimized batched lookups instead of
+        single calls to album_for_id.
+        """
+
+        return (self.album_for_id(id) for id in ids)
+
+    def tracks_for_ids(self, ids: Sequence[str]) -> Iterable[TrackInfo | None]:
+        """Batch lookup of track metadata for a list of track IDs.
+
+        Given a list of track identifiers, yields corresponding TrackInfo objects.
```
snejus, line 196: ✍🏼 Nit

These two docstrings have inconsistent formatting:


>         Given a list of track identifiers, yields corresponding TrackInfo objects.
>         Missing tracks result in None values in the output iterator. Plugins may


>        Given a list of album identifiers, yields corresponding AlbumInfo
>        objects. Missing albums result in None values in the output iterator.

beets/metadata_plugins.py
snejus: Makes sense

beets/metadata_plugins.py
snejus: `deezer` and `spotify` seem to return `None`s. Both have essentially duplicate implementations. For example, `spotify`:

```py
    def _search_api(
        self,
        query_type: Literal["album", "track"],
        filters: SearchFilter | None = None,
        keywords="",
    ) -> Sequence[IDResponse] | None:
        """Query the Spotify Search API for the specified ``keywords``,
        applying the provided ``filters``.

        :param query_type: Item type to search across. Valid types are:
            'album', 'artist', 'playlist', and 'track'.
        :param filters: (Optional) Field filters to apply.
        :return: JSON data for the class:`Response <Response>` object or None
            if no search results are returned.
        """
        query = self._construct_search_query(keywords=keywords, filters=filters)
        if not query:
            return None
        self._log.debug(f"Searching {self.data_source} for '{query}'")
        try:
            response = self._handle_response(
                "get",
                self.search_url,
                params={"q": query, "type": query_type},
            )
        except SpotifyAPIError as e:
            self._log.debug("Spotify API error: {}", e)
            return []
        response_data: Sequence[IDResponse] = response.get(
            query_type + "s", {}
        ).get("items", [])
        self._log.debug(
            "Found {} result(s) from {} for '{}'",
            len(response_data),
            self.data_source,
            query,
        )
        return response_data
```

This is a no-op since `filters` are guaranteed to have at least `album`:

```py
        query = self._construct_search_query(keywords=keywords, filters=filters)
        if not query:
            return None
```

beetsplug/deezer.py
snejus: Resolving this as we're discussing it in another thread.

beets/metadata_plugins.py
```diff
@@ -0,0 +1,348 @@
...
+        if not results:
+            return []
+
+        return filter(
+            None,
+            self.tracks_for_ids([result["id"] for result in results if result]),
+        )
+
+
+def artists_to_artist_str(
```
snejus, line 299: I doubt we want this to be split away from `MetadataSourcePlugin`.

It's only ever used by `MetadataSourcePlugin`, and only relevant in this context, why has it been made independent?

It now also needs an additional import.

setup.cfg
```diff
@@ -49,3 +49,8 @@ disallow_untyped_decorators = true
 disallow_any_generics = true
 check_untyped_defs = true
 allow_redefinition = true
```
snejus, line 51: Can we remove this globally, please? Unfortunately, I missed the PR where this made it in, otherwise I would have raised it

semohr: This is only for the plugins file, it is not global. If I remember correctly you were happy with it.

snejus: See my comment [here](https://github.com/beetbox/beets/pull/5738#issuecomment-2909819896)

beets/metadata_plugins.py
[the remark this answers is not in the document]
semohr: Just rebased on master and added the proposed implementation using the class name.

beets/metadata_plugins.py
semohr: I separated it because there's no reason for it to be part of the base class. While the import becomes slightly more verbose, the function doesn't need access to the plugin instance. Attaching standalone functions to the abstract base class, when they don’t interact with it, leads to poor design.

As I mentioned earlier, one of the reasons the previous design became unmaintainable was precisely because unrelated functions were lumped into the base class just because they were "related." Keeping concerns separate makes the codebase easier to manage and reason about.

beets/plugins.py
semohr: That is strange, as far as I can see the error happens before the changes. I.e. at `issubclass()` which I didn't touch.

beets/metadata_plugins.py
semohr: So I guess the method signature for the plugins needs some change than. You say filters are always required? If yes this should be reflected in the function arguments.

E.g. at the moment we have `def _construct_search_query(filters=None, keywords="")` which has an optional filter which should be required than. Same for `_search_api` abstract.

beets/metadata_plugins.py
```diff
@@ -0,0 +1,364 @@
...
+            data_source=self.data_source, info=info, config=self.config
+        )
+
+    @cached_classproperty
+    def data_source(cls) -> str:
+        """The data source name for this plugin.
+
+        This is inferred from the plugin name.
+        """
+        return cls.__name__.replace("Plugin", "")
```
snejus: image
I saw the exact same issue when I implemented `beetsplug.lyrics.Backend` class. I attempted to add types to `cached_classproperty` but failed and ended up using a metaclass:

```py
class BackendClass(type):
    @property
    def name(cls) -> str:
        """Return lowercase name of the backend class."""
        return cls.__name__.lower()


class Backend(RequestHandler, metaclass=BackendClass):
    def __init__(self, config, log):
        self._log = log
        self.config = config

    def fetch(
        self, artist: str, title: str, album: str, length: int
    ) -> tuple[str, str] | None:
        raise NotImplementedError
```

beets/metadata_plugins.py
snejus: Yes, you're right. If you look at the source code `_search_api` is **always** called with some filters. And those filters always include at least **album** key.

beets/plugins.py
snejus: That's the weird bit. I don't get this error on `master` but I get it here 😁

beets/metadata_plugins.py
semohr: I thought so, should be fixed here: https://github.com/beetbox/beets/pull/5815 and here: https://github.com/beetbox/beets/pull/5814

beets/metadata_plugins.py
snejus: > the function doesn't need access to the plugin instance

It never had access to the plugin __instance__ since it was a `staticmethod`.



> As I mentioned earlier, one of the reasons the previous design became unmaintainable was precisely because unrelated functions were lumped into the base class just because they were "related." Keeping concerns separate makes the codebase easier to manage and reason about.

I don't think I understand this point. How come is this function unrelated to the metadata source functionality, while it's imported and used by every data source, except musicbrainz? And it's not used anywhere else

```sh
$ oneline_python beetsplug/*.py | grep artists_to_artist_str
beetsplug/beatport.py: from beets.metadata_plugins import MetadataSourcePlugin, artists_to_artist_str
beetsplug/beatport.py:         return artists_to_artist_str(artists=artists, id_key=0, name_key=1)
beetsplug/deezer.py: from beets.metadata_plugins import (IDResponse, SearchApiMetadataSourcePlugin, SearchFilter, artists_to_artist_str)
beetsplug/deezer.py:             artist, artist_id = artists_to_artist_str(contributors)
beetsplug/deezer.py:         return AlbumInfo(album=album_data["title"], album_id=deezer_album_id, deezer_album_id=deezer_album_id, artist=artist, artist_credit=artists_to_artist_str([album_data["artist"]])[0], artist_id=artist_id, tracks=tracks, albumtype=album_data["record_type"], va=(len(album_data["contributors"]) == 1
beetsplug/deezer.py:         artist, artist_id = artists_to_artist_str(track_data.get("contributors", [track_data["artist"]]))
beetsplug/discogs.py: from beets.metadata_plugins import MetadataSourcePlugin, artists_to_artist_str
beetsplug/discogs.py:         artist, artist_id = artists_to_artist_str([a.data for a in result.artists], join_key="join")
beetsplug/discogs.py:         artist, artist_id = artists_to_artist_str(track.get("artists", []), join_key="join")
beetsplug/spotify.py: from beets.metadata_plugins import (IDResponse, SearchApiMetadataSourcePlugin, SearchFilter, artists_to_artist_str)
beetsplug/spotify.py:         artist, artist_id = artists_to_artist_str(album_data["artists"])
beetsplug/spotify.py:         artist, artist_id = artists_to_artist_str(track_data["artists"])
```

meanwhile, on `master`:

```sh
$ oneline_python beetsplug/*.py | grep '\.get_artist('
beetsplug/beatport.py:         return MetadataSourcePlugin.get_artist(artists=artists, id_key=0, name_key=1)
beetsplug/deezer.py:             artist, artist_id = self.get_artist(contributors)
beetsplug/deezer.py:         return AlbumInfo(album=album_data["title"], album_id=deezer_id, deezer_album_id=deezer_id, artist=artist, artist_credit=self.get_artist([album_data["artist"]])[0], artist_id=artist_id, tracks=tracks, albumtype=album_data["record_type"], va=(len(album_data["contributors"]) == 1
beetsplug/deezer.py:         artist, artist_id = self.get_artist(track_data.get("contributors", [track_data["artist"]]))
beetsplug/discogs.py:         artist, artist_id = MetadataSourcePlugin.get_artist([a.data for a in result.artists], join_key="join")
beetsplug/discogs.py:         artist, artist_id = MetadataSourcePlugin.get_artist(track.get("artists", []), join_key="join")
beetsplug/spotify.py:         artist, artist_id = self.get_artist(album_data["artists"])
beetsplug/spotify.py:         artist, artist_id = self.get_artist(track_data["artists"])
```

It's simply cleaner to keep it as a `staticmethod`. I expect to see `self.get_artist` in each case above.

semohr: The only practical difference with it being a static method is that it requires an additional import. Also, we don’t need access to the class type within the `get_artist` function.

My focus is on designing a clean and minimal base class without unnecessary components, primarily as a contract for all Source plugins.

I understand that from the consumer (plugin) perspective, it might be slightly easier to call an instance method, but I don’t think that should be the highest priority here.

This is similar to how the search API used to be part of the base class but was later removed since it’s not a core method defining the contract, rather, it’s more of a helper.

> I don't think I understand this point. How come is this function unrelated to the metadata source functionality, while it's imported and used by every data source, except musicbrainz? And it's not used anywhere else

I'm not saying it is unrelated for the metadata plugin implementations, I'm just saying it is unrelated from the beets core perspective as it is not needed for auto-tagging functionalities i.e. for the contract.

I hope this makes my thoughts here a bit more understandable.

beets/metadata_plugins.py
semohr: Seems a bit complicated or too complex with the ABC metaclass imo. I just made it an instance method instead :shrug:

beets/metadata_plugins.py
snejus: > The only practical difference with it being a static method is that it requires an additional import.

For me (a maintainer) that is a considerable difference since a duplicate import in multiple modules is less maintainable than the alternative.

> Also, we don’t need access to the class type within the get_artist function.

That's why it's a `staticmethod`, not a `classmethod`.

> My focus is on designing a clean and minimal base class without unnecessary components, primarily as a contract for all Source plugins.

I think I see what you mean - it seems like your focus is the classic interface design (in Java or so), i.e., __definition__ of abstract methods (contract) that children classes should implement. This is great - we're on the same page here. From my point of view, that's Part 1. 

Part 2 is centralisation/unification of shared functionality:

For `MetadataSourcePlugin` (all plugins):
* Name initialisation `data_source`
* Id extraction `_get_id`
* `get_artist` static method definition

For `SearchApiMetadataSourcePlugin`:
* `_search_api` _implementation_ (did you notice `deezer` and `spotify` implementations are equivalent/duplicate?
* maybe something else (haven't looked into this in detail)

snejus: When it comes to duplication, I think a single duplicate is acceptable, however anything more than this impacts long-term maintainability and indicates design issues.

If it helps, I can share my thoughts and actions encountering such duplicate import like `artists_to_artist_str` in the wild: 
* I can see multiple `artists_to_artist_str` imports
* I notice that it is _always_ imported with `MetadataSourcePlugin`, never independently.
  * For me this is a small indication that it maybe should be part of `MetadataSourcePlugin`
* I check the source code and see that it is indeed only used by `MetadataSourcePlugin` methods. In other words, `MetadataSourcePlugin` on it, and nothing else does, thus it should be defined under this class.
* It be renamed to `get_artist`, and the input parameter type should be updated to show the expected input shape instead
* Reasoning: this makes sure that functions are defined within the relevant context. For example, if we were to relocate `MetadataSourcePlugin` in the future, we will only need to update
  ```diff
  -from beets.metadata_plugins import MetadataSourcePlugin
  +from beets.new_location import MetadataSourcePlugin
  ```
  in all modules where it's imported. We don't have to bother about `get_artist` / `artists_to_artist_str` since it's under this class and accessed as `self.get_artist`.
  On the other hand, if `artists_to_artist_str` is kept as a separate function, it will have to be treated independently, and will require extra effort to check where it's used etc., only to find that it needs to be moved together with `MetadataSourcePlugin`. This is the maintenance cost we'd have to pay which is my primary issue with it.

semohr: > For me (a maintainer) that is a considerable difference since a duplicate import in multiple modules is less maintainable than the alternative.

I get your point, and I agree that reducing redundant imports can improve readability. However, for me, maintainability is more about separation of concerns and enabling future maintainers to reason about the code easily. Encapsulation of logic, especially when it's plugin-specific, is a key part of that.

> I think I see what you mean - it seems like your focus is the classic interface design (in Java or so), i.e., definition of abstract methods (contract) that children classes should implement. This is great - we're on the same page here. From my point of view, that's Part 1.

Yes, I completely agree—defining a clean and minimal base class as a contract is essential. That`s Part 1. But Part 2, in my view, is ensuring shared logic is centralized in the right layer to support long-term maintainability.

Currently, we`re mixing plugin logic into the main package more than necessary. The base class ends up taking responsibility for functionality that really belongs in the plugin namespace. Ideally, the core (beets) should be agnostic to how plugins implement their logic, caring only about the interfaces they expose.

That`s why I lean toward isolating shared functionality (like get_artist) into separate, plugin-side helpers. This makes them easier to move or restructure later, keeps the base class focused on its role as an interface, and avoids unnecessary entanglement.


> Reasoning: this makes sure that functions are defined within the relevant context. For example, if we were to relocate MetadataSourcePlugin in the future, we will only need to update...

I understand the benefit here. But there`s also a big tradeoff: embedding small helper functions into the class makes it harder to extract and reuse them later. If we want to decouple parts of the plugin logic, or apply multi-inheritance or composition, it becomes more complicated when the functionality is deeply nested inside a base class.

From a long-term design perspective, keeping logic modular and flat—especially when it's not strictly tied to the instance or class state—makes it more maintainable. It allows for easier testing, reusability, and ultimately supports the goal of moving all plugin logic out of the core namespace.

TLDR: I believe the most maintainable approach is one where core maintainers never have to touch plugin code and cleanly separating shared responsibilities helps us get there.

beets/plugins.py
snejus: Oh, I just realised that the error is thrown **not** inside the `issubclass` check on line 298 - it instead comes from `abc` module!

beets/metadata_plugins.py
snejus: I think we should not fall into the trap of [designing for the future](https://www.perplexity.ai/search/the-trap-of-attempting-to-desi-mNBju3FrTreoYK8.QwHq8g#1). We do not know whether this method will ever need to be reused outside this context. We should look at what we currently have: as things stand, [I would refactor](https://github.com/beetbox/beets/pull/5787#discussion_r2122333927) multiple imports (or any alternative design) back into the static `MetadataSourcePlugin.get_artist` method, exactly as this was defined previously. 

To get on the same page, let's just look at this pragmatically: assume that you came across this code in the wild, do you disagree with any specific step in my refactoring approach above?

beetsplug/deezer.py
semohr: > However, We do need to consider what happens to sources that don't have their ID pattern defined in id_extractors.py. I suggest we return the original _id:

Seems like a good idea, added in the latest commit.

beets/plugins.py
semohr: Still weird, the only change I introduced here was the check for ABC classes and skipping them. 

I added this line with the hope it fixes the issue:
`and inspect.isclass(obj)`

setup.cfg
semohr: Removed for the metadata_plugins section, which is the relevant one for this PR imo.

```toml
[[mypy-beets.metadata_plugins]]
disallow_untyped_decorators = true
disallow_any_generics = true
check_untyped_defs = true
```

beets/plugins.py
snejus: See my comments under https://github.com/beetbox/beets/pull/5739, maybe it's relevant. 

> I added this line with the hope it fixes the issue:
and inspect.isclass(obj)

If so, I think you can remove the `isinstance(obj, type)` check

setup.cfg
snejus: As I mentioned above, I want to remove `allow_redefinition = true` completely. This may be out of scope of this PR, so for now we can simply go ahead without __making any changes__ in `setup.cfg`.

beets/metadata_plugins.py
semohr: > To get on the same page, let's just look at this pragmatically: assume that you came across this code in the wild, do you disagree with any specific step in my refactoring approach above?

To be honest, I would disagree with the changes and would not approve of them or at least be reluctant.  Especially looking at it from beets core module it is not really visible why this method lives here at all. Further it introduces a bidirectionallity.
If you remember, I had similar concerns with your `extract release id` PR some time ago. I still think it would be better to move the code from the core module into e.g. the beetsplug namespace.

To not block this PR further tho, also as this more of a general disagreement. I added the get_artist statismethod to the ABC, but I do not stand behind it.

setup.cfg
semohr: Yeah I have seen this. We can work around the redefinition, it is just a bit annoying and might require additional variables in some cases, but I see the benefits for iterators. 

> without making any changes

To get you right, you want me to remove the metadata_plugin section again? I think going for this incremental opt in approach is quite nice for adaption.

beets/metadata_plugins.py
```diff
@@ -0,0 +1,380 @@
...
+            if join_key and artist.get(join_key, None):
+                name += f" {artist[join_key]} "
+            else:
+                name += ", "
+        artist_string += name
+
+    return artist_string, artist_id
+
+
+MetadataSourcePlugin.get_artist_str.__doc__ = artists_to_artist_str.__doc__
```
snejus: I struggle to understand why do we introduce this complexity instead of simply moving the function implementation under `MetadataSourcePlugin.get_artist`?

Additionally, now we have duplicate type definitions in the method and in the function.

There is an issue with `artists` type `Iterable[dict]` being too generic (need to specify the shape of the expected input dictionary), so we better fix it in a single place. :)

beets/metadata_plugins.py
snejus: Maybe we could instead just add `type: ignore[type-defined]`?

Technically, __it is__ a class attribute, and this error comes up because `cached_classproperty` is not properly typed, so we can safely ignore it, I think.

beets/metadata_plugins.py
```diff
@@ -0,0 +1,380 @@
...
+    def extract_release_id(self, url: str) -> str | None:
+        """Extract an ID from a URL for this metadata source plugin.
+
+        Uses the plugin's data source name to determine the ID format and
+        extracts the ID from a given URL.
+        """
+        return extract_release_id(self.data_source, url)
+
+    @staticmethod
+    def get_artist_str(
```
snejus:
```suggestion
    def get_artist(
```

Let's use the same name as previously - embedding type into method names is an anti-pattern, and we generally don't use this across the codebase.

beets/metadata_plugins.py
semohr: > I struggle to understand why do we introduce this complexity instead of simply moving the function implementation under MetadataSourcePlugin.get_artist?

Moved the implementation to the classmethod. 

> There is an issue with artists type Iterable[dict] being too generic (need to specify the shape of the expected input dictionary), so we better fix it in a single place. :)

Mypy did complain about a missing parameter for the generic, thus I added it. We can keep it as dict[Any, Any] but I would prefer to keep it as is.

beets/metadata_plugins.py
```diff
@@ -0,0 +1,365 @@
...
+        self,
+        item: Item,
+        info: TrackInfo,
+    ) -> Distance:
+        """Calculate the distance for a track based on its item and track info."""
+        return _get_distance(
+            data_source=self.data_source, info=info, config=self.config
+        )
+
+    @cached_classproperty  # type: ignore[type-defined]
```
snejus: The annotation needs to be on line 231.

@semohr do you run `mypy` locally?

snejus:
Can we keep this PR focused on what the title suggests?

> Refactor of metadata plugin and opt in all metadata plugins to new baseclass

Could you move commits with plugins and documentation improvements to separate branches, so that we can review them separately? I'm keen to merge the changes that have been reviewed and would like to have a final look without any noise.

beets/metadata_plugins.py
semohr: Oh yea,  you are right. Thought I had run it, also the type-defined should be a attr-defined.

semohr:
I will have another look here and rebase it after we merge the #5814 PR.

semohr:
Moved the docs changes into #5861. Feel free to have another look. Just rebased onto master and should good to go from my side.

snejus:
Thanks for moving the docs changes out! The PR still contains plugin refactors, namely `beatport`, `deezer` and `spotify`. I see the changes you've made in `discogs` and I expect an equivalent amount of changes to be made in the rest of the plugins, not more - to keep them focused on the new functionality.

semohr:
While beatport has seen more significant updates, changes to the other plugins are minimal and primarily related to the MetadataSourcePlugin interface.

For the spotify plugin, I introduced stricter types for query responses. These changes do not affect the core logic and are closely related to the MetadataSourcePlugin improvements, so I believe they belong in this PR.

We can drop the beatport changes if you'd prefer, though I probably won’t move them to a separate PR as I’m not using Beatport myself, just noticed some type issues that needed fixing.

beetsplug/discogs.py
```diff
@@ -40,7 +40,7 @@
 from beets import config
 from beets.autotag.distance import string_dist
 from beets.autotag.hooks import AlbumInfo, TrackInfo
-from beets.plugins import BeetsPlugin, MetadataSourcePlugin, get_distance
+from beets.metadata_plugins import MetadataSourcePlugin
 from beets.util.id_extractors import extract_release_id
```
snejus: This is not needed. The plugin should use the method available on `MetadataSourcePlugin`.

beetsplug/musicbrainz.py
```diff
@@ -362,9 +362,7 @@ def _merge_pseudo_and_actual_album(
     return merged
 
 
-class MusicBrainzPlugin(BeetsPlugin):
-    data_source = "Musicbrainz"
-
+class MusicBrainzPlugin(MetadataSourcePlugin):
```
snejus, line 364: Now that it inherits `MetadataSourcePlugin`, it should use the provided method to validate the release id. There are two lines that call `extract_release_id("musicbrainz", ...)` that need adjusting.

beetsplug/spotify.py
```diff
@@ -257,7 +290,7 @@ def album_for_id(self, album_id: str) -> AlbumInfo | None:
         :return: AlbumInfo object for album
         :rtype: beets.autotag.hooks.AlbumInfo or None
         """
-        if not (spotify_id := self._get_id(album_id)):
+        if not (spotify_id := self.extract_release_id(album_id)):
             return None
```
snejus, line 294: Why has the method been renamed? I prefer a different name for it to help to distinguish between the original function and the method. Each accepts a different set of parameters, so it avoids confusion too.

The underscore prefix in the original method name was intentional - we're not expecting this method to be used by anyone else except the plugin instance.

beetsplug/spotify.py
```diff
@@ -570,12 +604,12 @@ def _match_library_tracks(self, library: Library, keywords: str):
...
 
             # Apply market filter if requested
-            region_filter = self.config["region_filter"].get()
+            region_filter: str = self.config["region_filter"].get()
             if region_filter:
                 response_data_tracks = [
                     track_data
                     for track_data in response_data_tracks
-                    if region_filter in track_data["available_markets"]
+                    if region_filter in track_data.get("available_markets", [])
```
snejus: Is there a need to be defensive here? As far as I can see the key is always present in the data.

beetsplug/spotify.py
```diff
@@ -43,13 +48,41 @@
 DEFAULT_WAITING_TIME = 5
 
 
+class SpotifySearchResponseAlbums(IDResponse):
```
snejus: No need to prefix it with the plugin name as it only exists within this context.

```suggestion
class AlbumSearchResponse(IDResponse):
```

beetsplug/spotify.py
```diff
@@ -43,13 +48,41 @@
...
+
+    see https://developer.spotify.com/documentation/web-api/reference/search
+    """
+
+    album_type: str
+    available_markets: Sequence[str]
+    name: str
+
+
+class SpotifySearchResponseTracks(IDResponse):
```
snejus:
```suggestion
class TrackSearchResponse(IDResponse):
```

beetsplug/spotify.py
```diff
@@ -43,13 +48,41 @@
...
 class SpotifyAPIError(Exception):
     pass
 
 
-class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
-    data_source = "Spotify"
-
+class SpotifyPlugin(
+    SearchApiMetadataSourcePlugin[
+        Union[SpotifySearchResponseAlbums, SpotifySearchResponseTracks]
```
snejus:
```suggestion
        SpotifySearchResponseAlbums | SpotifySearchResponseTracks
```

beets/metadata_plugins.py
```diff
@@ -0,0 +1,367 @@
...
+
+
+class IDResponse(TypedDict):
+    """Response from the API containing an ID."""
+
+    id: str
+
+
+class SearchFilter(TypedDict):
+    artist: NotRequired[str]
```
snejus, line 322: The album is always present as far as I remember

```suggestion
    artist: str
```

beets/plugins.py
```diff
@@ -369,10 +297,10 @@ def load_plugins(names: Sequence[str] = ()) -> None:
             else:
                 for obj in getattr(namespace, name).__dict__.values():
                     if (
-                        isinstance(obj, type)
```
snejus, line 372: I'm still seeing

```
** error loading plugin importreplace:
  File "/home/sarunas/repo/beets/beets/plugins.py", line 303, in load_plugins
    and issubclass(obj, BeetsPlugin)
  File "/home/sarunas/.local/share/pyenv/versions/3.9.20/lib/python3.9/abc.py", line 123, in __subclasscheck__
    return _abc_subclasscheck(cls, subclass)
TypeError: issubclass() arg 1 must be a clas
```

This is failing because `BeetsPlugin` is now an abstract class. I've had a look at this and this fails when `obj` is a typing alias, for example `JSONDict = dict[str, Any]`.

You can solve this by adding

```py
from types import GenericAlias

...

and not isinstance(obj, GenericAlias)
```

Before the `BeetsPlugin` subclass check.

beetsplug/chroma.py
```diff
@@ -19,12 +19,14 @@
...
 from functools import cached_property, partial
+from typing import Iterable
 
 import acoustid
 import confuse
 
-from beets import config, plugins, ui, util
-from beets.autotag.distance import Distance
+from beets import config, ui, util
+from beets.autotag.hooks import Distance
```
snejus: It should be imported from the `distance` module as before

beetsplug/deezer.py
```diff
@@ -61,9 +64,11 @@ def func(lib: Library, opts, args):
 
         return [deezer_update_cmd]
 
+    # --------------------------------- id lookup -------------------------------- #
```
snejus: Let's remove these from all source files.

beetsplug/deezer.py
```diff
@@ -242,13 +256,17 @@ def _search_api(
...
         keywords="",
-    ) -> Sequence[Response]:
+    ) -> Sequence[IDResponse]:
         """Query the Deezer Search API for the specified ``keywords``, applying
         the provided ``filters``.
 
-        :param query_type: The Deezer Search API method to use.
+        :param query_type: Valid types are: 'album', 'artist', 'history',
+            'playlist', 'podcast', 'radio', 'track', 'user', and 'track'.
+            But only `track` and `album` are used.
```
snejus: This is redundant given the type above

beetsplug/deezer.py
```diff
@@ -242,13 +256,17 @@ def _search_api(
...
+        :param filters: (Optional) Field filters to apply.
+        :type filters: dict
```
snejus: Filters aren't optional. Best to just remove this and trust types as documentation.
```suggestion
```

beetsplug/musicbrainz.py
```diff
@@ -20,19 +20,19 @@
...
 import beets
 import beets.autotag.hooks
 from beets import config, plugins, util
-from beets.plugins import BeetsPlugin
+from beets.metadata_plugins import MetadataSourcePlugin
 from beets.util.id_extractors import extract_release_id
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator, Sequence
+    from collections.abc import Sequence
```
snejus: Duplicate import

beetsplug/spotify.py
```diff
@@ -390,7 +423,7 @@ def track_for_id(self, track_id: str) -> None | TrackInfo:
         return track
 
     def _construct_search_query(
```
snejus, line 425: Maybe this method should be moved to the base class, since it's duplicated in spotify and deezer?

beetsplug/spotify.py
```diff
@@ -595,7 +629,11 @@ def _match_library_tracks(self, library: Library, keywords: str):
                     len(response_data_tracks),
                 )
                 chosen_result = max(
-                    response_data_tracks, key=lambda x: x["popularity"]
+                    response_data_tracks,
+                    key=lambda x: x[
+                        # We are sure this is a track response!
+                        "popularity"  # type: ignore[typeddict-item]
```
snejus, line 635: `popularity` is [available on Albums too](https://developer.spotify.com/documentation/web-api/reference/get-an-album)!

beetsplug/spotify.py
```diff
@@ -694,13 +732,13 @@ def track_info(self, track_id: str):
         self._log.debug(
             "track_popularity: {} and track_isrc: {}",
             track_data.get("popularity"),
-            track_data.get("external_ids").get("isrc"),
+            track_data.get("external_ids", {}).get("isrc"),
```
snejus: Maybe we could just assign a variable once?

```py
external_ids = track_data["external_ids"]
```

These fields also need to be specified on the track typeddict.

pyproject.toml
```diff
@@ -52,7 +52,7 @@ musicbrainzngs = ">=0.4"
 numpy = ">=1.24.4"
 platformdirs = ">=3.5.0"
 pyyaml = "*"
-typing_extensions = { version = "*", python = "<=3.10" }
```
snejus: Why?

docs/changelog.rst
```diff
@@ -61,9 +61,20 @@ For plugin developers:
...
   Custom art sources might need to be adapted.
-
+* We split the responsibilities of plugins into two base classes
+    #. :class:`beets.plugins.BeetsPlugin` 
+        is the base class for all plugins, any plugin needs to inherit from this class.
+    #. :class:`beets.metadata_plugin.MetadataSourcePlugin`
+        allows plugins to act like metadata sources. E.g. used by the MusicBrainz plugin. All plugins
+        in the beets repo are opted into this class where applicable. If you are maintaining a plugin
+        that acts like a metadata source, i.e. you expose any of `track_for_id,
+        album_for_id, candidates, item_candidates, album_distance, track_distance` methods,
```
```suggestion
        that acts like a metadata source, i.e. you expose any of ``track_for_id``,
        ``album_for_id``, ``candidates``, ``item_candidates``, ``album_distance``, ``track_distance`` methods,
```

docs/changelog.rst
```diff
@@ -61,9 +61,20 @@ For plugin developers:
...
+        please update your plugin to inherit from the new baseclass, as otherwise it will
+        not be registered as a metadata source and wont be usable going forward.
```
snejus: Oh snap, I didn't realize this was the case! Just tried importing and found that beetcamp is indeed gone.

We need to come up with a way to deprecate the old functionality but keep it functional, so that plugin developers have time to migrate. This will warrant a major version bump as well.

beetsplug/spotify.py
semohr: Nope it is not :) Have a look at the search endpoint, which returns SimplifiedAlbumsObject without popularity.

semohr: https://developer.spotify.com/documentation/web-api/reference/search

pyproject.toml
semohr: Line 14 in metadataplugin.py:
```
from typing_extensions import NotRequired
```
Needs python 3.11 or typing extensions.

beetsplug/spotify.py
semohr: Generices are a bit weird regarding that. Even tho we use the future import this requires a Union. Otherwise I get the following mypy error:

```
beetsplug/spotify.py:82: error: Type expected within [...]
```

beetsplug/spotify.py
semohr: Indeed we could move it to the SearchApiMetadataSourcePlugin. But I wouldn't move it into the MetadataSourcePlugin.

beetsplug/spotify.py
semohr: From the spotify api documentation it is not clear if the external_ids are always set for tracks. They might be optional here. But I agree a variable might help here.

beets/plugins.py
semohr: Is that python version dependent? I have never run into this issue in testing.

docs/changelog.rst
semohr: I guess we can add a deprecation warning and do the full change another time. Will be a bit difficult tho as we would need to inspect each plugin if it has some metadatasource related functions.

added a proposal see [here](https://github.com/beetbox/beets/pull/5787/commits/259c9cc7a3f51fadca851ee1800e9daa2b2c48b1)

beetsplug/spotify.py
semohr: `get_id` was too generic in my opinion and wasn't a too good descriptor of the method. We can rename to something more descriptive. Any suggestions?

I don't mind the duplicate name too much tho.

beets/metadata_plugins.py
semohr: I think it is not required. In the item_candidates method we use:
```python
results = self._search_api("track", {"artist": artist}, keywords=title)
```
vs candidate method:
```python
query_filters: SearchFilter = {"album": album}
if not va_likely:
    query_filters["artist"] = artist

results = self._search_api("album", query_filters)
```

beetsplug/spotify.py
semohr: I've given this some more thought, and I don't think we should move the function into the base class. The similarity between the deezer and spotify implementations is mostly due to their nearly identical query syntax. For another (as yet unknown) SearchApiMetadataSourcePlugin, the requirements might differ significantly. Having a shared base implementation could become confusing or misleading down the line.

Even tho we have some duplication now I think it is more clear.

beetsplug/spotify.py
snejus: `_extract_id` maybe?

beets/plugins.py
snejus: Try `pip install beetcamp` and enable `bandcamp` in your plugins list

beetsplug/spotify.py
snejus: > Indeed we could move it to the SearchApiMetadataSourcePlugin

That's indeed what I am suggesting.

> For another (as yet unknown) SearchApiMetadataSourcePlugin, the requirements might differ significantly. 

That's something to consider in the future, when we have another plugin using. Today we're working with what we have - remember that this class was created just for these two plugins anyways.

pyproject.toml
snejus: That's exactly what this requirement specified, I think?

docs/changelog.rst
snejus: Checking for presence of `data_source` attribute should be enough. The real problem is that many of the previously present methods are gone from `BeetsPlugin` now.

pyproject.toml
semohr: Oh you are right I just assumed that was the reason why I changed it as I did remember there was an issue but was not too sure anymore what the actual issue was. Must have been a part of the code I removed. 

Can we keep it tho? The whole python version tango for type hints is very annoying and issues with it are not catched by the CI pipeline properly as we have seen before (e.g. #5695). It also enables us to use default values in TypeVar (which we already do in some places even tho it is a feature introduced in python 3.13, ref dbcore/db.py:52) and more python 3.13 typing features.

docs/changelog.rst
semohr: Indeed, I guess the solution is not optimal. Maybe we should backport the methods into the `BeetsPlugin` and add a deprecation warning to them. That would allow legacy metadata plugins to still work (hopefully). We need to properly remove the methods at some point tho.

I just came up with this deprecation instead, which seems to work and all methods are now still available in the base class.

```python
    """Returns a list of MetadataSourcePlugin subclass instances

    Resolved from all currently loaded beets plugins.
    """

    all_plugins = find_plugins()
    metadata_plugins = []
    for plugin in all_plugins:
        if isinstance(plugin, MetadataSourcePlugin):
            metadata_plugins.append(plugin)
        elif hasattr(plugin, "data_source"):
            # TODO: Remove this elif in the future, only used for legacy support
            warnings.warn(
                f"{plugin.__class__.__name__} is used as a legacy metadata source. "
                "It should extend MetadataSourcePlugin instead of BeetsPlugin. "
                "Support for this will be removed in a future release!",
                DeprecationWarning,
                stacklevel=2,
            )
            metadata_plugins.append(plugin)

    return metadata_plugins
    
# Dynamically copy methods to BeetsPlugin for legacy support
# TODO: Remove this in a future release
for name, method in inspect.getmembers(
    MetadataSourcePlugin, predicate=inspect.isfunction
):
    if not hasattr(BeetsPlugin, name):
        setattr(BeetsPlugin, name, method)

```

beets/plugins.py
semohr: Seems indeed to be python version specific. Am not getting this issue on python 3.12 but on python 3.9 I do have the issue too.

docs/changelog.rst
semohr:
```python
from beets import plugins, metadata_plugins
plugins.load_plugins(["bandcamp"])

metadata_plugins.find_metadata_source_plugins() # <- Triggers warning
<ipython-input-4-beae3b674f7d>:1: DeprecationWarning: BandcampPlugin is used as a legacy metadata source. It should extend MetadataSourcePlugin instead of BeetsPlugin. Support for this will be removed in a future release!
metadata_plugins.find_metadata_source_plugins()
```

pyproject.toml
snejus: Honestly, I think it's indeed a better idea to just have

```toml
typing_extensions = "*"
```

and remove it once we upgrade to Python 3.11.

On the other hand, this means that for consistency we need to remove all these conditional blocks from the codebase

```
beetsplug/aura.py-if sys.version_info >= (3, 11):
beetsplug/aura.py-    from typing import Self
beetsplug/aura.py-else:
beetsplug/aura.py:    from typing_extensions import Self
```

so I think it's out of scope of this PR.

docs/changelog.rst
snejus: This is very nice, I love it!!!

You can specifically say

```py
# TODO: Remove this in the future major release, v3.0.0
```

beets/plugins.py
snejus: Nice, thanks for investigating!

pyproject.toml
semohr: Created an issue (#5869) to keep track of that.

beetsplug/spotify.py
semohr: Overall I agree we could/should do that. Seems a bit out of scope here tho. Let's tackle this with #5860 after this PR is merged.

beetsplug/spotify.py
semohr: Renamed to `_extract_id`. I also added a method for `_get_id` to maintain legacy support.

beets/metadata_plugins.py
```diff
@@ -251,14 +251,31 @@ def data_source(cls) -> str:
...
+        """
+        warnings.warn(
+            "Support for this method is deprecated and will be removed in the future. "
+            "Use _extract_id instead. "
+            "Support for this will be removed in the v3.0.0 release!",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self._extract_id(url)
+
```
snejus: This method lived for just a couple of weeks and since it's been named to show that it's only used internally, we can do whatever we want with it - thus we can demolish it without any deprecation.

semohr:
> This may be my last comment 😅 Check the build for typing error, I think there's one popping up:
> 
> ```
> Argument 1 to "append" of "list" has incompatible type "BeetsPlugin"; expected "MetadataSourcePlugin"
> ```

Seemly like this was an issue introduced with the legacy support. Sadly the fix needs a type ignore but we should be able to remove it with v3.0.0.

snejus:
Just rebase it and it will get merged.

PR #5795, opened 2025-05-21
Streamlined auto api referencing for documentation

semohr:
## Description

The current developer documentation feels somewhat cluttered due to inline auto-generated API references for certain classes. To improve readability and maintainability, this PR introduces a more streamlined approach that aligns better with best practices observed in other PyData ecosystem documentation.

Specifically, this PR:
- Adds a dedicated `api/` folder to the documentation structure.
- Moves all auto-generated references (classes, methods, etc.) to this folder.
- Enables clean, concise linking to API elements from the narrative documentation—without interrupting human-written content with large autogenerated blocks.

This separation makes the documentation easier to navigate and maintain, while still providing full API reference coverage where needed.

- [x] Documentation
- [x] Changelog


JOJ0 (approved):
I rendered this on my machine and compared it to the original docs. This reads much better now! It really was cluttered having straight API autodoc stuff and human written stuff mixed. I would say we merge this. We can't check through each and everything anyway. I suppose you have checked with the rst links checker thingy?

One thing, but that is an actual content change and not only format change, include it now or later: MetaDataSourcePlugin is missing in the api docs. I think it's worth including. Or maybe better after your work on it (splitting PR) is done?

semohr:
The #5787 PR already has docs added on the MetaDataSourcePlugin. Let's omit it here for now. The dev docs need a bit more love anyways.

PR #5814, opened 2025-06-02
Minor improvements to deezer plugin typing.

semohr:
## Description

Added some more typehints to deezer plugin.

I know, it is properly not used much and we don't even have test for the deezer plugin but I want to make this a bit more maintainable, mainly to prepare for https://github.com/beetbox/beets/pull/5787 and make migration a bit easier.

### DIFF
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -28,6 +28,7 @@ from typing import (
     Any,
     Callable,
     Generic,
+    Literal,
     Sequence,
     TypedDict,
     TypeVar,
@@ -737,7 +738,7 @@ class MetadataSourcePlugin(Generic[R], BeetsPlugin, metaclass=abc.ABCMeta):
     @abc.abstractmethod
     def _search_api(
         self,
-        query_type: str,
+        query_type: Literal["album", "track"],
         filters: dict[str, str] | None,
         keywords: str = "",
     ) -> Sequence[R]:
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -739,7 +739,7 @@ class MetadataSourcePlugin(Generic[R], BeetsPlugin, metaclass=abc.ABCMeta):
     def _search_api(
         self,
         query_type: Literal["album", "track"],
-        filters: dict[str, str] | None,
+        filters: dict[str, str],
         keywords: str = "",
     ) -> Sequence[R]:
         raise NotImplementedError
--- a/beetsplug/deezer.py
+++ b/beetsplug/deezer.py
@@ -41,9 +41,11 @@ class DeezerPlugin(MetadataSourcePlugin[Response], BeetsPlugin):
         "deezer_track_id": types.INTEGER,
         "deezer_updated": types.DATE,
     }
-
-    def __init__(self):
-        super().__init__()
+    # Base URLs for the Deezer API
+    # Documentation: https://developers.deezer.com/api/
+    search_url = "https://api.deezer.com/search/"
+    album_url = "https://api.deezer.com/album/"
+    track_url = "https://api.deezer.com/track/"
 
     def commands(self):
         """Add beet UI commands to interact with Deezer."""
@@ -206,13 +208,6 @@ class DeezerPlugin(MetadataSourcePlugin[Response], BeetsPlugin):
             data_url=track_data["link"],
             deezer_updated=time.time(),
         )
-
-    # ------------------------------- Data fetching ------------------------------ #
-    # Base URLs for the Deezer API
-    # Documentation: https://developers.deezer.com/api/
-    search_url = "https://api.deezer.com/search/"
-    album_url = "https://api.deezer.com/album/"
-    track_url = "https://api.deezer.com/track/"
 
     @staticmethod
     def _construct_search_query(
--- a/beetsplug/deezer.py
+++ b/beetsplug/deezer.py
@@ -232,7 +232,16 @@ class DeezerPlugin(MetadataSourcePlugin[Response], BeetsPlugin):
 
     def _search_api(
         self,
-        query_type: Literal["album", "track"],
+        query_type: Literal[
+            "album",
+            "track",
+            "artist",
+            "history",
+            "playlist",
+            "podcast",
+            "radio",
+            "user",
+        ],
         filters: dict[str, str],
         keywords="",
     ) -> Sequence[Response]:
@@ -240,8 +249,6 @@ class DeezerPlugin(MetadataSourcePlugin[Response], BeetsPlugin):
         the provided ``filters``.
 
         :param query_type: The Deezer Search API method to use.
-            Valid types are: 'album', 'artist', 'history', 'playlist',
-            'podcast', 'radio', 'track', 'user', and 'track'.
         :param keywords: (Optional) Query keywords to use.
         :return: JSON data for the class:`Response <Response>` object or None
             if no search results are returned.
--- a/beetsplug/deezer.py
+++ b/beetsplug/deezer.py
@@ -26,10 +26,10 @@ import unidecode
 from beets import ui
 from beets.autotag import AlbumInfo, TrackInfo
 from beets.dbcore import types
-from beets.library import DateType, Item, Library
 from beets.plugins import BeetsPlugin, MetadataSourcePlugin, Response
 
 if TYPE_CHECKING:
+    from beets.library import Item, Library
     from beetsplug._typing import JSONDict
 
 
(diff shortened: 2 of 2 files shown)


beetsplug/deezer.py
```diff
@@ -252,25 +240,23 @@ def _construct_search_query(filters=None, keywords=""):
             query = query.decode("utf8")
         return unidecode.unidecode(query)
 
-    def _search_api(self, query_type, filters=None, keywords=""):
+    def _search_api(
+        self,
+        query_type: Literal["album", "track"],
+        filters: dict[str, str],
```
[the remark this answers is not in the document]
semohr: This is intended, see discussion [here](https://github.com/beetbox/beets/pull/5787#discussion_r2120819953).

beetsplug/deezer.py
```diff
@@ -18,18 +18,22 @@
...
-from beets.plugins import BeetsPlugin, MetadataSourcePlugin
+from beets.library import DateType, Item, Library
+from beets.plugins import BeetsPlugin, MetadataSourcePlugin, Response
 
+if TYPE_CHECKING:
+    from beetsplug._typing import JSONDict
 
-class DeezerPlugin(MetadataSourcePlugin, BeetsPlugin):
+
+class DeezerPlugin(MetadataSourcePlugin[Response], BeetsPlugin):
```
snejus, line 36: Shouldn't we define a deezer-specific `Response`?

beetsplug/deezer.py
```diff
@@ -230,18 +181,50 @@ def track_for_id(self, track_id=None, track_data=None):
...
+            data_url=track_data["link"],
+            deezer_updated=time.time(),
+        )
+
+    # ------------------------------- Data fetching ------------------------------ #
+    # Base URLs for the Deezer API
+    # Documentation: https://developers.deezer.com/api/
+    search_url = "https://api.deezer.com/search/"
+    album_url = "https://api.deezer.com/album/"
+    track_url = "https://api.deezer.com/track/"
```
snejus: Why have these been moved? They are static class properties so should be defined at the top, I think.

beets/plugins.py
```diff
@@ -737,7 +738,7 @@ def track_url(self) -> str:
     @abc.abstractmethod
     def _search_api(
         self,
-        query_type: str,
+        query_type: Literal["album", "track"],
```
snejus, line 741: I'd suggest defining

```py
QueryType = Literal["album", "track"]
```

And import it in `beetsplug/deezer.py` instead of defining it in multiple places

beetsplug/deezer.py
```diff
@@ -252,25 +235,23 @@ def _construct_search_query(filters=None, keywords=""):
             query = query.decode("utf8")
         return unidecode.unidecode(query)
 
-    def _search_api(self, query_type, filters=None, keywords=""):
+    def _search_api(
+        self,
+        query_type: Literal["album", "track"],
+        filters: dict[str, str],
```
snejus, line 245: See `mypy` error on this line

beets/plugins.py
semohr: Yeah something alike would be preferable, I might revisit it once we merge the metadata plugin PR.

beetsplug/deezer.py
semohr: Optimally yes! Adding one should be relative straight forward, this Response here only defines an `id` attribute. 

I did not want to create an deezer account to look at the api documentation tbh. But maybe someone else wants to tackle this.

beetsplug/deezer.py
semohr: I always try to group related code, but I think this was not intentional in this PR. Was a cherrypick :grin: 

Gonna revert it later today.

beetsplug/deezer.py
semohr: This is temporary until the metadata plugin PR is merged. We approach it similarly in #5815

beets/plugins.py
semohr: Also I just noticed that the deezer api can take other query types, thus defining it globally might intertwine the types more the necessary. E.g. for deezer the complete type-hint would be:
```python
Literal[
    "album",
    "track",
    "artist",
    "history",
    "playlist",
    "podcast",
    "radio",
    "user",
],
```

snejus: I see. I don't think the interface/abstract definition should be coupled to any specific child/plugin implementation.

We know that this method is only ever called with values `album` and `track`, so I think we should just add these two in the type definition.

semohr: For the main definition that is the case i.e. in the MetadatasourcePlugin the method only supports album and track. 

It should be fine if the specific plugins support more than we need from the core perspective. Just added it here to indicate that (in theory) the deezer api supports these endpoint and removed the comment regarding the endpoints.

snejus:
Not sure if you've seen my review comment

> Would it be possible to isolate method moves into a separate commit? It's a bit complicated to see the relevant changes when the entire method implementations are red and green 😅

Would you be happy to do so?

beets/plugins.py
snejus: Makes sense. As long as this satisfies the Liskov substitution principle, I'm happy with it!

semohr: It does, I think mypy would complain if it didnt  :+1:

semohr:
> Not sure if you've seen my review comment
> 
> > Would it be possible to isolate method moves into a separate commit? It's a bit complicated to see the relevant changes when the entire method implementations are red and green 😅
> 
> Would you be happy to do so?

Sure, the moves should now be contained in one commit. I also added that one to the .git-blame-ignore-revs

beetsplug/deezer.py
```diff
@@ -18,62 +18,49 @@
...
         """Add beet UI commands to interact with Deezer."""
         deezer_update_cmd = ui.Subcommand(
             "deezerupdate", help=f"Update {self.data_source} rank"
         )
 
-        def func(lib, opts, args):
+        def func(lib: Library, opts, args):
             items = lib.items(ui.decargs(args))
-            self.deezerupdate(items, ui.should_write())
+            self.deezerupdate(list(items), ui.should_write())
```
snejus, line 58: Why are items casted to list? I can see that the logic in `deexerupdate` loops across them just once, so any `Iterable` input should be safe. Actually, it would be best to update the type annotation to clarify this!

beetsplug/deezer.py
```diff
@@ -230,18 +183,43 @@ def track_for_id(self, track_id=None, track_data=None):
...
 
-        :param filters: (Optional) Field filters to apply.
-        :type filters: dict
+        :param filters: Field filters to apply.
         :param keywords: (Optional) Query keywords to use.
-        :type keywords: str
         :return: Query string to be provided to the Search API.
-        :rtype: str
         """
         query_components = [
```
snejus, line 224: Probably out of scope of your PR but this implementation could be simplified to:

```py
        parts = [keywords, *(f'{k}:"{v}"' for k, v in filters.items())]
        return unidecode.unidecode(" ".join(filter(None, parts)))
```

snejus (commented):
Aplogies, didn't realise that I forgot to submit this review 🤦🏼 . Just a pair of small questions

beetsplug/deezer.py
semohr: Because the `deezerupdate` uses the `len()` function which is incompatible with `Iterable`. It is only used in a log statement and we could fix that, but I did not want to change the implementation in this PR.

beetsplug/deezer.py
semohr: Maybe we also want to remove the unidecode here or move it behind a feature flag. We did something similar with the spotify plugin recently.

snejus: You're right, we do indeed want to remove it. See
```
Sending event: trackinfo_received
Item search terms: Betwixt & Between - たんこぶへそ太郎の放浪
spotify: Searching Spotify for 'たんこぶへそ太郎の放浪 artist:Betwixt & Between'
spotify: Found 0 result(s) from Spotify for 'たんこぶへそ太郎の放浪 artist:Betwixt & Between'
Sending event: trackinfo_received
Sending event: trackinfo_received
deezer: Searching Deezer for 'tankobuhesoTai Lang noFang Lang  artist:"Betwixt & Between"'
deezer: Found 25 result(s) from Deezer for 'tankobuhesoTai Lang noFang Lang  artist:"Betwixt & Between"'
Sending event: trackinfo_received
Sending event: trackinfo_received
Sending event: trackinfo_received
Sending event: trackinfo_received
```

semohr: see #5860

snejus (approved):

PR #5815, opened 2025-06-02
Minor improvements to spotify plugin typing.

semohr:
## Description

Added some more typehints to the spotify plugin. Also added a method to get the tokenfile and changed to logic for the handle_response to use `requests.request`.

This is done mainly to prepare for https://github.com/beetbox/beets/pull/5787, see also https://github.com/beetbox/beets/pull/5814

### DIFF
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -25,6 +25,7 @@ import json
 import re
 import time
 import webbrowser
+from typing import TYPE_CHECKING, Any, Literal, Sequence
 
 import confuse
 import requests
@@ -33,8 +34,11 @@ import unidecode
 from beets import ui
 from beets.autotag.hooks import AlbumInfo, TrackInfo
 from beets.dbcore import types
-from beets.library import DateType
-from beets.plugins import BeetsPlugin, MetadataSourcePlugin
+from beets.library import DateType, Library
+from beets.plugins import BeetsPlugin, MetadataSourcePlugin, Response
+
+if TYPE_CHECKING:
+    from beetsplug._typing import JSONDict
 
 DEFAULT_WAITING_TIME = 5
 
@@ -107,33 +111,33 @@ class SpotifyPlugin(MetadataSourcePlugin, BeetsPlugin):
         self.config["client_id"].redact = True
         self.config["client_secret"].redact = True
 
-        self.tokenfile = self.config["tokenfile"].get(
-            confuse.Filename(in_app_dir=True)
-        )  # Path to the JSON file for storing the OAuth access token.
         self.setup()
 
     def setup(self):
         """Retrieve previously saved OAuth token or generate a new one."""
+
         try:
-            with open(self.tokenfile) as f:
+            with open(self._tokenfile()) as f:
                 token_data = json.load(f)
         except OSError:
             self._authenticate()
         else:
             self.access_token = token_data["access_token"]
 
+    def _tokenfile(self) -> str:
+        """Get the path to the JSON file for storing the OAuth token."""
+        return self.config["tokenfile"].get(confuse.Filename(in_app_dir=True))
+
     def _authenticate(self):
         """Request an access token via the Client Credentials Flow:
         https://developer.spotify.com/documentation/general/guides/authorization-guide/#client-credentials-flow
         """
+        c_id: str = self.config["client_id"].as_str()
+        c_secret: str = self.config["client_secret"].as_str()
+
         headers = {
             "Authorization": "Basic {}".format(
-                base64.b64encode(
-                    ":".join(
-                        self.config[k].as_str()
-                        for k in ("client_id", "client_secret")
-                    ).encode()
-                ).decode()
+                base64.b64encode(":".join([c_id, c_secret]).encode()).decode()
             )
         }
         response = requests.post(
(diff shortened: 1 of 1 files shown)


beetsplug/spotify.py
```diff
@@ -107,33 +111,33 @@ def __init__(self):
...
+
         headers = {
             "Authorization": "Basic {}".format(
-                base64.b64encode(
-                    ":".join(
-                        self.config[k].as_str()
-                        for k in ("client_id", "client_secret")
-                    ).encode()
-                ).decode()
+                base64.b64encode(":".join([c_id, c_secret]).encode()).decode()
```
wisp3rwind: This doesn't look like the prototypical usecase for `join`, maybe it would be even more concise and clear as

```suggestion
                base64.b64encode(f"{c_id}:{c_secret}").encode()).decode()
```

?

beetsplug/spotify.py
```diff
@@ -154,27 +158,32 @@ def _authenticate(self):
...
 
     def _handle_response(
-        self, request_type, url, params=None, retry_count=0, max_retries=3
-    ):
+        self,
+        method: Literal["get", "post", "put", "delete"],
+        url: str,
+        params: Any = None,
+        retry_count=0,
+        max_retries=3,
```
wisp3rwind: While you're at it, maybe also add the remaining argument types?

```suggestion
        retry_count: int = 0,
        max_retries: int = 3,
```

beetsplug/spotify.py
```diff
@@ -404,34 +409,31 @@ def _construct_search_query(filters=None, keywords=""):
...
-        :type filters: dict
         :param keywords: (Optional) Query keywords to use.
-        :type keywords: str
-        :return: JSON data for the class:`Response <Response>` object or None
-            if no search results are returned.
-        :rtype: dict or None
         """
         query = self._construct_search_query(keywords=keywords, filters=filters)
-        if not query:
-            return None
```
wisp3rwind, line 424: Maybe this should be replaced by a `if not keywords` check before the call to `_construct_search_query` instead of removing the check entirely?

beetsplug/spotify.py
```diff
@@ -683,11 +682,9 @@ def _fetch_info(self, items, write, force):
             if write:
                 item.try_write()
 
-    def track_info(self, track_id=None):
+    def track_info(self, track_id: str):
         """Fetch a track's popularity and external IDs using its Spotify ID."""
-        track_data = self._handle_response(
-            requests.get, self.track_url + track_id
-        )
+        track_data = self._handle_response("get", self.track_url + track_id)
```
wisp3rwind, line 687: Seems like there should be an early return here if 'track_data is None`?

wisp3rwind (changes requested):
A few small nitpicks, otherwise this looks good to me :+1: I particularly like the `request_type` refactor, there's really no need to pass around functions here.

I'm not up to date on the related in-flight PRs on refactoring metadata source plugins in general, so I only looked at this in isolation. I suppose mypy's complaints about the `_search_api` signature would be resolved via those PRs?

beetsplug/spotify.py
semohr: I don't think we need a check here: We always return track_data from the `_handle_response` function. If it fails the function will raise. Was different beforehand but makes things easier to reason about imo.

beetsplug/spotify.py
semohr: It is fine if no keywords are given afaiks.  The `_construct_search_query` method  only requires filters, keywords are optional.

semohr:
> I'm not up to date on the related in-flight PRs on refactoring metadata source plugins in general, so I only looked at this in isolation. I suppose mypy's complaints about the _search_api signature would be resolved via those PRs?

The typing issue should be fixed once the abstract base class used by the spotify plugin changes. I might need to revisit this tho. I felt like the `requests.request` addition does not fit the metadataplugin PR thus I opened this one.

beetsplug/spotify.py
wisp3rwind: Oh, what happened here is that I thought that the mypy warnings below are due to the possibility of `track_data` being `None`: As you say, that's not actually true. What mypy really seems to be complaining about is nested accesses. Not sure whether those should have more validation, but that's probably out-of-scope in this PR. (Is there some higher-level function that would catch the exception if `track_data` wouldn't match the expected schema?)

beetsplug/spotify.py
wisp3rwind: My concern was that (at least superficially), it seems like the code previously prevented a completely empty query being sent to the server (which I'd guess might have unintended side-effects, if only by yielding completely random data).
Maybe there's no way such a query would ever be constructed by our code and that's fine?

In any case, I don't know the code around metadata handling sufficiently well to say much about this; I'll leave this up to your judgement :)

semohr: I think this None check was some historical code that can never be reached in the current version. A query is always returned by `_construct_search_query` as we always give a filter. At least in the beets first part code. Maybe it was different at some point but at the moment it shouldn't be an issue as far as I can see.

In future if we run into issues regarding this, I would opt to just raise an error in the `_construct_search_query` function instead. Will keep my eyes open for issues regarding this :+1:

beetsplug/spotify.py
semohr: I don't think we have any validation in place. We just hope for the best and pray I guess 😅

PR #5841, opened 2025-06-24
Fix pipx install action

snejus:
## Fix pipx-install-action Windows compatibility issue

Replace `BrandonLWhite/pipx-install-action@v1.0.3` with `threeal/pipx-install-action@v1.0.0` in CI workflow due to Windows compatibility issues.

**Changes:**
- Switch to alternative pipx-install-action that works on Windows
- Add explicit package specification for poethepoet and poetry
- Resolves GitHub issue: https://github.com/BrandonLWhite/pipx-install-action/issues/62

**Edit**: received a reply from `pipx-install-action` maintainer with a fix: simply remove the breaking cache from GitHub Actions: this worked fine.

Thus I reverted the commit which introduced another action on Windows. The only change that this PR makes is `pipx-install-action` version upgrade.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,7 +21,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install Python tools
-        uses: BrandonLWhite/pipx-install-action@v1.0.3
+        # BrandonLWhite/pipx-install-action@v1.0.3 fails on Windows, thus we're using an alternative action here
+        uses: threeal/pipx-install-action@v1.0.0
+        with:
+          packages: poethepoet>=0.26 poetry<2
       - name: Setup Python with poetry caching
         # poetry cache requires poetry to already be installed, weirdly
         uses: actions/setup-python@v5
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,7 +21,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install Python tools
-        uses: BrandonLWhite/pipx-install-action@v1.0.1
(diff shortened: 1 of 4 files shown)


semohr (approved):

wisp3rwind:
Looks good to me as a quick fix to CI :+1: Copilot's suggestions around consistency probably make sense?

However, I am wondering about the reliability of the new action compared to the previous one: The code looks fine superficially. It's also a one-man-show, where the author seems to have limited time to maintain it, and I'm not sure it's in widespread use.

Maybe this would in fact be a reason to push for `uv`: It does come with a reusable action by the authors (see https://docs.astral.sh/uv/guides/integration/github/) which supports caching, and it can manage python and tools. (I think we could also use `uv` only in this role for now, and continue building with poetry.)

snejus:
> Looks good to me as a quick fix to CI 👍 Copilot's suggestions around consistency probably make sense?
> 
> However, I am wondering about the reliability of the new action compared to the previous one: The code looks fine superficially. It's also a one-man-show, where the author seems to have limited time to maintain it, and I'm not sure it's in widespread use.
> 
> Maybe this would in fact be a reason to push for `uv`: It does come with a reusable action by the authors (see https://docs.astral.sh/uv/guides/integration/github/) which supports caching, and it can manage python and tools. (I think we could also use `uv` only in this role for now, and continue building with poetry.)


I started looking into `uv` and meanwhile received a comment under the issue with a fix, so  received a reply from pipx-install-action maintainer with a fix: simply remove the breaking cache from GitHub Actions which worked fine.

Thus I reverted the commit which introduced another action on Windows. The only change that this PR makes is pipx-install-action version upgrade.

PR #5846, opened 2025-06-28
Preserve line breaks for example cases in substitution plugin docs

kchr:
## Description

Adds line block markup to example substitutions in the plugin documentation, so that each case is shown on a separate line:

> The replacement can be an expression utilising the matched regex, allowing us to create more general rules. Say for example, we want to sort all albums by multiple artists into the directory of the first artist. We can thus capture everything before the first ,, `` &`` or `` and``, and use this capture group in the output, discarding the rest of the string.
> 
> ```yaml
> substitute:
>  ^(.*?)(,| &| and).*: \1
> ```
>
> This would handle all the below cases in a single rule:
>
>> Bob Dylan and The Band -> Bob Dylan
>> Neil Young & Crazy Horse -> Neil Young
>> James Yorkston, Nina Persson & The Second Hand Orchestra -> James Yorkston

## To Do

- [X] Documentation. (If you've added a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)
- [X] ~Tests. (Very much encouraged but not strictly required.)~


semohr:
Thanks for the fix!

I've just enabled auto-merge. If you update the branch, it should merge into `master` automatically. Let me know if you run into any issues.

PR #5833, opened 2025-06-22
Convert flexible field values to SQL before storing them

Maxr1998:
## Description

This is required to support multivalued fields as flexible fields.
See #5698 for more details.

## To Do

- [ ] Changelog. (Add an entry to `docs/changelog.rst` to the bottom of one of the lists near the top of the document.)
- [ ] Tests. (Very much encouraged but not strictly required.)

(will add those later)

PR #5847, opened 2025-06-29
Import MusicBrainz composer/lyricist/arranger ids

### LINKED ISSUE #5698: Support storing multivalued composer/lyricist/arranger MBIDs in the database
## Motivation

Currently, beets stores the names of writers and arrangers in the database and the audio file tags. I'm currently using these to build automated playlists (with the `smartplaylist` plugin) based on these writers. However, this has several issues:
- Names of these writers are not necessarily unique, creating collisions and thus including unrelated tracks in my playlists
- Matching these single-valued tags makes it even more complex, since some names might have substring overlaps, but you can't do exact matches (because there can be multiple writers in a single field), and handling the separator char (comma) with RegEx is unnecessarily complex.

## Proposal

Thus, I suggest adding new fields to the database and possibly to files:
- `mb_composerids`
- `mb_lyricistids`
- `mb_arrangerids`

All of these should be multivalued fields for future proofing. Adding legacy fields that simply join the values with a separator would be possible, but isn't strictly necessary in my opinion.

### As a plugin

This feature could be implemented as a plugin, in theory:

<details>
<summary>Plugin implementation</summary>

```python

MB_COMPOSERIDS = "mb_composerids"
MB_LYRICISTIDS = "mb_lyricistids"
MB_ARRANGERIDS = "mb_arrangerids"


class WriterMbidsPlugin(BeetsPlugin):
    item_types = {
        MB_COMPOSERIDS: types.MULTI_VALUE_DSV,
        MB_LYRICISTIDS: types.MULTI_VALUE_DSV,
        MB_ARRANGERIDS: types.MULTI_VALUE_DSV,
    }

    def __init__(self):
        super().__init__()
        self.register_listener("mb_track_extract", self.track_extract)

    @staticmethod
    def track_extract(data):
        recording = data
        composer_mbids = []
        lyricist_mbids = []
        arranger_mbids = []

        for work_relation in recording.get("work-relation-list", ()):
            if work_relation["type"] != "performance":
                continue

            work = work_relation["work"]
            for artist_relation in work.get("artist-relation-list", ()):
                if "type" in artist_relation:
                    artist_type = artist_relation["type"]
                    if artist_type == "composer":
                        composer_mbids.append(artist_relation["artist"]["id"])
                    elif artist_type == "lyricist":
                        lyricist_mbids.append(artist_relation["artist"]["id"])

        for artist_relation in recording.get("artist-relation-list", ()):
            if "type" in artist_relation:
                artist_type = artist_relation["type"]
                if artist_type == "arranger":
                    arranger_mbids.append(artist_relation["artist"]["id"])

        info = {}
        if composer_mbids:
            info[MB_COMPOSERIDS] = composer_mbids
        if lyricist_mbids:
            info[MB_LYRICISTIDS] = lyricist_mbids
        if arranger_mbids:
            info[MB_ARRANGERIDS] = arranger_mbids
        return info
```
</details>

However, there's one issue currently preventing this: plugins cannot use multivalued fields because beets doesn't handle those for flexible fields:

<details>
<summary>Stacktrace</summary>

```
…
  File "/usr/lib/python3.13/site-packages/beets/dbcore/db.py", line 679, in add
    self.store()
    ~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/beets/library.py", line 396, in store
    super().store(fields)
    ~~~~~~~~~~~~~^^^^^^^^
  File "/usr/lib/python3.13/site-packages/beets/dbcore/db.py", line 616, in store
    tx.mutate(
    ~~~~~~~~~^
        "INSERT INTO {} "
        ^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        (self.id, key, value),
        ^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/lib/python3.13/site-packages/beets/dbcore/db.py", line 987, in mutate
sqlite3.ProgrammingError: Error binding parameter 3: type 'list' is not supported
```
</details>

From what I understand, this is because beets doesn't use the sql type conversions for flexible fields, only for built-in fields:

<details>
<summary>Current code</summary>

```python
# dbcore.py
…

   def store(self, fields: Optional[Iterable[str]] = None):
        …
        for key in fields:
            if key != "id" and key in self._dirty:
                self._dirty.remove(key)
                assignments.append(key + "=?")
                value = self._type(key).to_sql(self[key]) # !!! TYPE IS CONVERTED !!!
                subvars.append(value)
        with db.transaction() as tx:
            # Main table update.
            if assignments:
                query = "UPDATE {} SET {} WHERE id=?".format(
                    self._table, ",".join(assignments)
                )
                subvars.append(self.id)
                tx.mutate(query, subvars)

            # Modified/added flexible attributes.
            for key, value in self._values_flex.items():
                if key in self._dirty:
                    self._dirty.remove(key)
                    tx.mutate(
                        "INSERT INTO {} "
                        "(entity_id, key, value) " # !!! NO TYPE CONVERSION !!!
                        "VALUES (?, ?, ?);".format(self._flex_table),
                        (self.id, key, value),
                    )
        …
```
</details>

To fix this, the value could be converted to SQL for flexible fields as well:

```diff
diff --git a/beets/dbcore/db.py b/beets/dbcore/db.py
old mode 100644
new mode 100644
index 4645d4b..53c462d
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -611,6 +611,7 @@ class Model(ABC):
             for key, value in self._values_flex.items():
                 if key in self._dirty:
                     self._dirty.remove(key)
+                    value = self._type(key).to_sql(value)
                     tx.mutate(
                         "INSERT INTO {} "
                         "(entity_id, key, value) "
```

### As a core feature

Alternatively, implementation as a core feature could also work. This would require a few changes to the `Item` model, but could otherwise be handled mostly in the `mbsync` plugin. The "bug" described above wouldn't impact this implementation, but should probably still be addressed.

## Implementation

I'm interested in implementing this myself, please tell me which option you prefer, and I will open a pull request.

Maxr1998:
## Description

Updates the MusicBrainz plugin to also import MBIDs for composers/lyricists/arrangers, and adds them as multi-valued fields.

Closes #5698.

## To Do

- [X] ~~Documentation.~~ (not required for this change)
- [x] Changelog
- [x] Tests

### DIFF
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -455,19 +455,17 @@ class TrackInfo(Info):
         **kwargs,
     ) -> None:
         self.arrangers = arrangers
-        # Stay consistent with arrangers if ids are not supplied by provider
+        # Remove existing IDs if they are not provided by the data source
         self.arrangers_ids = arrangers_ids or []
         self.bpm = bpm
         self.composers = composers
         self.composer_sort = composer_sort
-        # Stay consistent with composers if ids are not supplied by provider
         self.composers_ids = composers_ids or []
         self.disctitle = disctitle
         self.index = index
         self.initial_key = initial_key
         self.length = length
         self.lyricists = lyricists
-        # Stay consistent with lyricists if ids are not supplied by provider
         self.lyricists_ids = lyricists_ids or []
         self.mb_workid = mb_workid
         self.medium = medium
@@ -475,7 +473,6 @@ class TrackInfo(Info):
         self.medium_total = medium_total
         self.release_track_id = release_track_id
         self.remixers = remixers
-        # Stay consistent with remixers if ids are not supplied by provider
         self.remixers_ids = remixers_ids or []
         self.title = title
         self.track_alt = track_alt


semohr:
Heya, thank you for the PR!

I'm not too sure if we want to mirror more ids from musicbrainz in the autotag functionalities if they are not actively used in the lookup process. Especially since mb was moved into a plugin and other metadata sources do not supply the same information.

Maybe instead of specifically defining all different id properties, we can find a more generic way to allow adding arbitrary IDs  (Optimally without the need to manually edit the autotag module)? Seems to me like we might have aquie some duplication for multivalue id fields.

I would be highly interested in @snejus thoughts on this. Let's discuss this a bit before we continue with the merge. 

Best, Sebastian

Maxr1998:
> I'm not too sure if we want to mirror more ids from musicbrainz in the autotag functionalities if they are not actively used in the lookup process. Especially since mb was moved into a plugin and other metadata sources do not supply the same information.
> 
> Maybe instead of specifically defining all different id properties, we can find a more generic way to allow adding arbitrary IDs  (Optimally without the need to manually edit the autotag module)? Seems to me like we might have aquie some duplication for multivalue id fields.

Yes, this is a good point. In my original issue in #5698 I wrote a separate plugin that contained all these properties inside, we can likely do the same for the new `musicbrainz` plugin. However, the  we definitely need #5833 first because otherwise those fields can't store multiple values as a list.

Maxr1998:
Rebased my changes now. Instead of adding the new fields to beets itself (library, hooks, autotag), putting them only in the plugin should be possible now after my previous PR has been merged.

@snejus what do you think? Additionally, I'm still a bit confused as to why there's both a `artists_ids` and `mb_artistids` field. Should this be mirrored for composers/lyricists/arrangers? What should be managed by the plugin, what shouldn't?

semohr:
> I'm still a bit confused as to why there's both a artists_ids and mb_artistids field. Should this be mirrored for composers/lyricists/arrangers? What should be managed by the plugin, what shouldn't?

There’s also `mb_albumartistid`. The main database tables are in need of a cleanup and unification; the naming is inconsistent, as you’ve probably noticed. From a database normalization perspective, we’re carrying around a lot of redundant fields that really should be consolidated.

As for composers/lyricists/arrangers: duplicating these fields across artist_ids vs. mb_* variants probably isn’t the right long-term direction. Ideally the core schema would handle the distinction cleanly, with plugins consuming or extending it rather than re-implementing.

For this specific case, I think it’s fine to add here for now and revisit later. I don’t think we should try to tackle that larger issue in this PR.

Maxr1998:
Ok, that makes sense.

So, should the fields still be moved to the plugin, or can I keep them in the beets core for now?

semohr:
In my opinion moving the fields to the plugin itself would be preferred and will spare us some work down the line. How would you achieve that?

Maxr1998:
You can basically just declare them within the plugin and they'll be picked up by beets. I wonder though if that makes sense for all of these fields. A more generic `composers_ids` might still belong in the core so that other plugins can also provide that value. The redundant `mb_composerids` should be moved to the plugin, or eventually just removed (and maybe not even added in this PR).

Maxr1998:
@semohr did you have some time to further look into this or discuss the changes? I can resolve the current conflicts, but I see there's some more refactoring going on presently (#6052), which would cause conflicts again if this can't be merged before.

Maxr1998:
I updated the PR once more and removed the duplicate `mb_` tags for now, only adding the multi-valued, unbranded tags. (cc @semohr @snejus)

Maxr1998:
Rebased the changes once more and fixed the changelog conflict 🙂

Maxr1998:
Is there anything else I can do to get this reviewed and merged?

snejus:
Sorry - ignore my activity there as I misread which fields you are introducing.

Note that I just merged some changes where we know store each of these artist credits in multi-valued fields so you will need to rebase to resolve a fair bit of conflicts and keep your change to ID-based fields only.

I will review this as soon as possible after this. Sorry for my late response here!

Maxr1998:
> Sorry - ignore my activity there as I misread which fields you are introducing.
> 
> Note that I just merged some changes where we know store each of these artist credits in multi-valued fields so you will need to rebase to resolve a fair bit of conflicts and keep your change to ID-based fields only.
> 
> I will review this as soon as possible after this. Sorry for my late response here!

@snejus no worries, rebasing wasn't too bad, since most changes are now already handled by your earlier changes. Thus, the PR only got smaller. The `apply_item_metadata` changes in `autotag/__init__.py` seem not to be necessary anymore, so I removed those.

Maxr1998:
Added remixer ids as well now. Gonna fix the tests in a bit.

Maxr1998:
There we go. Everything is ready to review and merge now.

snejus:
Cool, will have a look at this later today!

beets/autotag/hooks.py
```diff
@@ -452,14 +455,17 @@ def __init__(
         **kwargs,
     ) -> None:
         self.arrangers = arrangers
+        self.arrangers_ids = arrangers_ids or []
```
snejus, line 459: Add a comment to explain why are we defaulting to an empty list. Presumably, that's to make sure that previous IDs are cleared if the backend does not populate them.

beetsplug/musicbrainz.py
```diff
@@ -371,7 +364,9 @@ def track_info(
             info.isrc = ";".join(recording["isrcs"])
 
         lyricists = []
-        composer = []
+        lyricists_ids = []
+        composers = []
```
snejus, line 368: Thanks!

docs/changelog.rst
```diff
@@ -34,7 +34,9 @@ New features
   ``arranger`` fields. Existing libraries are migrated automatically, and
   :doc:`plugins/musicbrainz` now preserves each MusicBrainz ``remixer``,
   ``lyricist``, ``composer``, and ``arranger`` relation as a separate value.
-  :bug:`5698`
+  MBIDs for each relation are also stored in new multi-valued fields
```
snejus: I like your attention to detail! I mistakenly included this bug in my previous PR.

Do add a separate entry for this change though.

beets/autotag/hooks.py
Maxr1998: I think I mainly did that because the old multi-value code also did, I'll have to check if it's actually necessary or if we can keep the default.

Maxr1998: Ok, yes, we can remove the fallback. Good catch!

beets/autotag/hooks.py
```diff
@@ -451,14 +454,17 @@ def __init__(
         **kwargs,
     ) -> None:
         self.arrangers = arrangers
+        self.arrangers_ids = arrangers_ids
```
snejus: Let's say I've got an album from `MusicBrainz` where `remixers` and `remixers_ids` have been populated by `musicbrainz`.

If I then reimport the same album with `beet import -LI` but I pick the candidate from `bandcamp`, `remixers` will get populated, however `bandcamp` does not populate `remixers_ids`, so the track will retain the IDs from MusicBrainz.

Now the question is whether we want to retain these IDs? Defaulting to an empty list would remove them.

docs/changelog.rst
```diff
@@ -34,7 +34,9 @@ New features
   ``arranger`` fields. Existing libraries are migrated automatically, and
   :doc:`plugins/musicbrainz` now preserves each MusicBrainz ``remixer``,
   ``lyricist``, ``composer``, and ``arranger`` relation as a separate value.
-  :bug:`5698`
+- Store MBIDs for remixers, lyricists, composers, and arrangers in the new
```
```suggestion
- :doc:`plugins/musicbrainz`: Store MBIDs for remixers, lyricists, composers, and arrangers in the new
```

beets/autotag/hooks.py
Maxr1998: Really good question. When I wrote the changes originally, MusicBrainz was still the only provider. I think for the case you describe, consistency is more important than completeness, so I'd replace the IDs with an empty list. Should I replace the default `None` in the constructor with an empty list then?

snejus: I think that's the right direction forward. Yes, please!

snejus: Note keep `None` in the argument list and have `remixers_ids = remixers_ids or []`

Maxr1998: Ok, I'll do that! Should I also change the other id fields back or just remixers? In both cases, I'll add an explanation comnent as well.

snejus: All four please!

Maxr1998: Made the requested changes. I also noticed that I hadn't added the remixers_ids model yet, I also did that now.

beets/autotag/hooks.py
```diff
@@ -451,20 +455,28 @@ def __init__(
         **kwargs,
     ) -> None:
         self.arrangers = arrangers
+        # Stay consistent with arrangers if ids are not supplied by provider
```
snejus:
```suggestion
        # Remove existing IDs if they are not provided by the data source
```

And remove the rest of the comments - one is enough.

Maxr1998: Fixed. Feel free to squash the changes when merging, by the way. But I can also rebase locally and only squash the review feedback PRs if you prefer that.

snejus: It's all good, don't worry!

Maxr1998: Alright!

Maxr1998:
Thanks for the review and all the feedback!

PR #5856, opened 2025-07-05
Only consider release collections in mbcollection plugin

nicomem:
## Description

Currently, `mbcollection` takes the first collection it finds (if the user does not explicitely indicates one in the configuration).

The problem is that the user may have collections of types that we are not interested in (e.g. recording collections, artist collections, etc.).

Since the `get_collections` response returns the collections types, we can only keep the `release` collections.

## To Do

- [x] ~Documentation.~ (If you've added a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)

### DIFF
--- a/beetsplug/mbcollection.py
+++ b/beetsplug/mbcollection.py
@@ -70,10 +70,14 @@ class MusicBrainzCollectionPlugin(BeetsPlugin):
         if not collections["collection-list"]:
             raise ui.UserError("no collections exist for user")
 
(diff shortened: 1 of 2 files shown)


snejus (approved):
Looks great, thank you. There's only a conflict in the changelog that needs resolving.

nicomem:
> Looks great, thank you. There's only a conflict in the changelog that needs resolving.

I rebased the branch and fixed the conflict

PR #5859, opened 2025-07-07
Remove decargs and other references to Python 2

snejus:
This PR modernizes the codebase by removing Python 2 compatibility code and simplifying several areas:

- Deleted `BytesQuery` class (replaced with `PathQuery`)
- Removed `decargs()` function that was a no-op in Python 3
- Simplified `print_()` function signature and implementation
- Removed coding-related workarounds in various modules


beets/ui/__init__.py
```diff
@@ -104,30 +104,16 @@ def _stream_encoding(stream, default="utf-8"):
...
     The `end` keyword argument behaves similarly to the built-in `print`
     (it defaults to a newline).
     """
-    if not strings:
-        strings = [""]
-    assert isinstance(strings[0], str)
-
-    txt = " ".join(strings)
-    txt += kwargs.get("end", "\n")
+    txt = " ".join(strings or ("",))
```
semohr: Nit:
```
txt = " ".join(strings or ("",)) + end
```

might be a bit more efficient due to fewer temporary objects

beets/ui/__init__.py
```diff
@@ -1311,7 +1297,7 @@ def _set_format(
         if fmt:
             value = fmt
         elif value:
-            (value,) = decargs([value])
+            (value,) = [value]
```
semohr: This seems very unnecessary :)

beetsplug/bpsync.py
```diff
@@ -65,7 +65,7 @@ def func(self, lib, opts, args):
         move = ui.should_move(opts.move)
         pretend = opts.pretend
         write = ui.should_write(opts.write)
-        query = ui.decargs(args)
+        query = args
```
semohr: Seems a bit unnecessary now.

semohr (changes requested):
Just had a short look and this looks mostly good! Thank you for the cleanup!

There are a number of places where we now have `query = args`, maybe we should rethink this. Will have another look once the tests are fixed (I guess this was done before #5775 was merged)

semohr (approved):

PR #5861, opened 2025-07-07
Metadatasource cleanup docs

semohr:
This PR includes documentation updates for the new `metadatasource` plugin architecture, as requested by @snejus. The docs changes were split out from the original implementation to keep things focused and reviewable.

**Details**

* Introduces comprehensive documentation for the new metadata plugin system.
* Performs a general cleanup of the plugin-related developer documentation for clarity and consistency.

**Note**
This PR should be reviewed and merged **together with** [#5787]


semohr:
@beetbox/maintainers Could we get a review here please?
I split it out from #5787, but it's still very closely related. We're currently missing documentation for the new features, so this helps address that. 😉

wisp3rwind:
Could you briefly point out which sections of the docs have actual changes to their content, and not just layout tweaks? Since the first commit moves around a lot of text, but also makes some changes to them, this is a little tricky to review. Thanks!

snejus:
@semohr note you can just run `poe docs` from the repo root - no need for `cd docs && make docs`

semohr:
> Could you briefly point out which sections of the docs have actual changes to their content, and not just layout tweaks? Since the first commit moves around a lot of text, but also makes some changes to them, this is a little tricky to review. Thanks!

You're right, the first commit does move a lot of content around, which makes it tricky to spot the actual content changes. As far as I can tell, the only completely new section is the `Extending the Autotagger` section in `other.rst`.

If you'd prefer, I can split the structural changes and the content edits into separate commits for easier review. That might take a bit of time though, and since this is just developer docs, I figured a single pass might be sufficient. Let me know what works best for you!

semohr:
Fot the current CI error see https://github.com/LilSpazJoekp/docstrfmt/issues/127

semohr:
@wisp3rwind The initial move is now contained in a single commit.

docs/dev/plugins/index.rst
```diff
@@ -0,0 +1,102 @@
+Plugin Development
+==================
+
+Beets plugins are Python modules or packages that extend the core functionality
+of beets. The plugin system is designed to be flexible, allowing developers to
+add virtually any type of features to beets.
+
+For instance you can create plugins that add new commands to the command-line
+interface, listen for events in the beets lifecycle or extend the autotagger
+with new metadata sources. .. _writing-plugins:
```
snejus:
```suggestion
with new metadata sources.
```

docs/dev/plugins/index.rst
```diff
@@ -0,0 +1,102 @@
...
+
+.. _writing-plugins:
```
snejus: This should probably follow the section heading. Relevant references should be adjusted too

```suggestion
.. _basic-plugin-setup:
```

docs/dev/plugins/index.rst
```diff
@@ -0,0 +1,102 @@
...
+
+Basic Plugin Setup
+------------------
+
+A beets plugin is just a Python module or package inside the ``beetsplug``
+namespace [namespace]_ package. To create the basic plugin layout, create a
```
snejus: Does it look as intended?

Image

docs/dev/plugins/index.rst
```diff
@@ -0,0 +1,102 @@
...
+directory called ``beetsplug`` and add either your plugin module:
+
+.. code-block:: shell
+
+    beetsplug/
+    └── myawesomeplugin.py
```
snejus, line 24: Love these unicode lines, thanks!

docs/dev/plugins/events.rst
```diff
@@ -0,0 +1,206 @@
...
+
+    class SomePlugin(BeetsPlugin):
+        def __init__(self):
+            super().__init__()
+            self.register_listener("pluginload", self.loaded)
+
+        def loaded(self):
+            self._log.info("Plugin loaded!")
+
+.. list-table:: Plugin Events
```
snejus: Well done, this reads much more clearly now.

Could we make this a definition list? For me this improves readability even further, we can easily add references to each event, and we don't need to hardcode column widths:

Image


See the relevant chat with CopilotChat which reformatted the table accordingly:

<details>
<summary>Conversation Log</summary>

## User (cb9c1a5b-6910-4fb2-b457-a9c72a392d90)---

Is there a better way to format a restructured text table?

## Copilot (2694d432-c0cb-42ec-9c1a-8c18b81a13ca)---

Yes. Options (pick what optimizes readability + diffs):

1. Definition list (often clearest)
EventName
  :Parameters: lib (Library), paths (list[str])
  :Description: Called after the import command finishes.
Pros: Easy to diff, no column wrapping issues. Cons: Not a literal table.

2. Field list per event (same idea)
.. rubric:: pluginload
:Parameters: (none)
:Description: Called after ...

3. csv-table directive
.. csv-table:: Plugin Events
   :header: Event, Parameters, Description
   :widths: 15,25,60

   pluginload, ,Called after all plugins load
   import,"lib; paths[]","Called after import command finishes"
Pros: Compact source. Cons: Awkward for long multi‑paragraph cells; commas need quoting.

4. Grid table
+--------+-----------+-------------+
| Event  | Parameters| Description |
+--------+-----------+-------------+
Pros: Renders everywhere. Cons: Pain to edit, wide diffs.

5. Keep list-table but shorten cells: move repeated type prefixes into footnotes or a legend, or replace :py:class: roles with backticks if not needed.

Recommendation: Use a definition list per event with :Parameters: and :Description: fields; it stays readable and avoids fragile table syntax while still structured enough for Sphinx to style.

## User (fadf55ff-ddaa-48e6-8501-9acf5346b9d8)---

Replace my table with the definition list then please

## Copilot (5e2804ad-e98e-4fc3-9654-33f037087380)---

```rst path=/home/sarunas/repo/beets/docs/dev/plugins/events.rst start_line=47 end_line=203
.. rubric:: Plugin Events

``pluginload``
  :Parameters: (none)
  :Description: Called after all plugins have been loaded after the ``beet`` command starts.

``import``
  :Parameters: :py:class:`lib <beets.library.Library>`, ``paths`` (list of path strings)
  :Description: Called after the ``import`` command finishes.

``album_imported``
  :Parameters: :py:class:`lib <beets.library.Library>`, :py:class:`album <beets.library.Album>`
  :Description: Called every time the importer finishes adding an album to the library.

``album_removed``
  :Parameters: :py:class:`lib <beets.library.Library>`, :py:class:`album <beets.library.Album>`
  :Description: Called every time an album is removed from the library (even when its files are not deleted from disk).

``item_copied``
  :Parameters: :py:class:`item <beets.library.Item>`, ``source`` (path), ``destination`` (path)
  :Description: Called whenever an item file is copied.

``item_imported``
  :Parameters: :py:class:`lib <beets.library.Library>`, :py:class:`item <beets.library.Item>`
  :Description: Called every time the importer adds a singleton to the library (not called for full-album imports).

``before_item_imported``
  :Parameters: :py:class:`item <beets.library.Item>`, ``source`` (path), ``destination`` (path)
  :Description: Called with an ``Item`` object immediately before it is imported.

``before_item_moved``
  :Parameters: :py:class:`item <beets.library.Item>`, ``source`` (path), ``destination`` (path)
  :Description: Called with an ``Item`` object immediately before its file is moved.

``item_moved``
  :Parameters: :py:class:`item <beets.library.Item>`, ``source`` (path), ``destination`` (path)
  :Description: Called with an ``Item`` object whenever its file is moved.

``item_linked``
  :Parameters: :py:class:`item <beets.library.Item>`, ``source`` (path), ``destination`` (path)
  :Description: Called with an ``Item`` object whenever a symlink is created for a file.

``item_hardlinked``
  :Parameters: :py:class:`item <beets.library.Item>`, ``source`` (path), ``destination`` (path)
  :Description: Called with an ``Item`` object whenever a hardlink is created for a file.

``item_reflinked``
  :Parameters: :py:class:`item <beets.library.Item>`, ``source`` (path), ``destination`` (path)
  :Description: Called with an ``Item`` object whenever a reflink is created for a file.

``item_removed``
  :Parameters: :py:class:`item <beets.library.Item>`
  :Description: Called with an ``Item`` object every time an item (singleton or part of an album) is removed from the library (even when its file is not deleted from disk).

``write``
  :Parameters: :py:class:`item <beets.library.Item>`, ``path`` (path), ``tags`` (dict)
  :Description: Called just before a file's metadata is written to disk. Handlers may modify ``tags`` or raise ``library.FileOperationError`` to abort.

``after_write``
  :Parameters: :py:class:`item <beets.library.Item>`
  :Description: Called after a file's metadata is written to disk.

``import_task_created``
  :Parameters: :py:class:`task <beets.importer.ImportTask>`, :py:class:`session <beets.importer.ImportSession>`
  :Description: Called immediately after an import task is initialized. May return a list (possibly empty) of replacement tasks.

``import_task_start``
  :Parameters: :py:class:`task <beets.importer.ImportTask>`, :py:class:`session <beets.importer.ImportSession>`
  :Description: Called before an import task begins processing.

``import_task_apply``
  :Parameters: :py:class:`task <beets.importer.ImportTask>`, :py:class:`session <beets.importer.ImportSession>`
  :Description: Called after metadata changes have been applied in an import task (on the UI thread; keep fast). Prefer a pipeline stage otherwise (see :ref:`plugin-stage`).

``import_task_before_choice``
  :Parameters: :py:class:`task <beets.importer.ImportTask>`, :py:class:`session <beets.importer.ImportSession>`
  :Description: Called after candidate search and before deciding how to import. May return an importer action (only one handler may return non-None).

``import_task_choice``
  :Parameters: :py:class:`task <beets.importer.ImportTask>`, :py:class:`session <beets.importer.ImportSession>`
  :Description: Called after a decision has been made about an import task. Use ``task.choice_flag`` to inspect or change the action.

``import_task_files``
  :Parameters: :py:class:`task <beets.importer.ImportTask>`, :py:class:`session <beets.importer.ImportSession>`
  :Description: Called after filesystem manipulation (copy/move/write) for an import task.

``library_opened``
  :Parameters: :py:class:`lib <beets.library.Library>`
  :Description: Called after beets starts and initializes the main Library object.

``database_change``
  :Parameters: :py:class:`lib <beets.library.Library>`, :py:class:`model <beets.library.Model>`
  :Description: A modification has been made to the library database (may not yet be committed).

``cli_exit``
  :Parameters: :py:class:`lib <beets.library.Library>`
  :Description: Called just before the ``beet`` command-line program exits.

``import_begin``
  :Parameters: :py:class:`session <beets.importer.ImportSession>`
  :Description: Called just before a ``beet import`` session starts.

``trackinfo_received``
  :Parameters: :py:class:`info <beets.autotag.TrackInfo>`
  :Description: Called after metadata for a track is fetched (e.g., from MusicBrainz). Handlers can modify the tags seen by later pipeline stages or adjustments (e.g., ``mbsync``).

``albuminfo_received``
  :Parameters: :py:class:`info <beets.autotag.AlbumInfo>`
  :Description: Like ``trackinfo_received`` but for album-level metadata.

``before_choose_candidate``
  :Parameters: :py:class:`task <beets.importer.ImportTask>`, :py:class:`session <beets.importer.ImportSession>`
  :Description: Called before prompting the user during interactive import. May return a list of ``PromptChoices`` to append to the prompt (see :ref:`append_prompt_choices`).

``mb_track_extract``
  :Parameters: :py:class:`data <dict>`
  :Description: Called after metadata is obtained from MusicBrainz for a track. Must return a (possibly empty) dict of additional ``field: value`` pairs to apply (overwriting existing fields).

``mb_album_extract``
  :Parameters: :py:class:`data <dict>`
  :Description: Like ``mb_track_extract`` but for album tags. Overwrites tags set at the track level with the same field.
```

## User (e0b8e3ce-a512-40d3-90b5-aa6a90a1bf18)---
</details>

docs/dev/plugins/events.rst
```diff
@@ -0,0 +1,206 @@
...
+        dictionary to customize the tags that are written to the media file.
+        Event handlers may also raise a ``library.FileOperationError`` exception
+        to abort the write operation. Beets will catch that exception, print an
+        error message, and continue.
+    - - `after_write`
+      - :py:class:`item <beets.library.Item>`
+      - called with an ``Item`` object after a file's metadata is written to
+        disk (i.e., just after the file on disk is closed).
+    - - `import_task_created`
+      - :py:class:`task <beets.importer.ImportTask>`, :py:class:`session
```
snejus: For some reason these are not resolved in rendered docs. I tried using explicit import references, e.g. `beets.importer.tasks.ImportTask` but it didn't help. 

I guess that's because we haven't added importer classes to the API reference?

Image

docs/dev/plugins/events.rst
```diff
@@ -0,0 +1,206 @@
...
+
+    - - Event
+      - Parameters
+      - Description
+    - - `pluginload`
+      -
+      - called after all the plugins have been loaded after the ``beet`` command
+        starts
+    - - `import`
+      - :py:class:`lib <beets.library.Library>`, ``paths`` is a list of paths
```
snejus: We may want to define substitutions for repeated references:

```rst
.. rubric:: Plugin Events

.. |Library| replace:: :class:`beets.library.Library`

...

``import``
  :Parameters: lib (|Library|), paths (list[path])
  :Description: Called after the ``import`` command finishes.
```

<details><summary>All replacements</summary>

```rst

.. rubric:: Plugin Events

.. |Library| replace:: :class:`~beets.library.Library`
.. |Album| replace:: :class:`~beets.library.Album`
.. |Item| replace:: :class:`~beets.library.Item`
.. |Model| replace:: :class:`~beets.library.Model`
.. |ImportTask| replace:: :class:`~beets.importer.ImportTask`
.. |ImportSession| replace:: :class:`~beets.importer.ImportSession`
.. |TrackInfo| replace:: :class:`~beets.autotag.TrackInfo`
.. |AlbumInfo| replace:: :class:`~beets.autotag.AlbumInfo`

``pluginload``
  :Parameters: (none)
  :Description: Called after all plugins have been loaded after the ``beet`` command starts.

``import``
  :Parameters: lib (|Library|), paths (list[path])
  :Description: Called after the ``import`` command finishes.

``album_imported``
  :Parameters: lib (|Library|), album (|Album|)
  :Description: Called every time the importer finishes adding an album to the library.

``album_removed``
  :Parameters: lib (|Library|), album (|Album|)
  :Description: Called every time an album is removed from the library (even when its files are not deleted from disk).

``item_copied``
  :Parameters: item (|Item|), source (path), destination (path)
  :Description: Called whenever an item file is copied.

``item_imported``
  :Parameters: lib (|Library|), item (|Item|)
  :Description: Called every time the importer adds a singleton to the library (not called for full-album imports).

``before_item_imported``
  :Parameters: item (|Item|), source (path), destination (path)
  :Description: Called with an Item immediately before it is imported.

``before_item_moved``
  :Parameters: item (|Item|), source (path), destination (path)
  :Description: Called immediately before an item's file is moved.

``item_moved``
  :Parameters: item (|Item|), source (path), destination (path)
  :Description: Called whenever an item's file is moved.

``item_linked``
  :Parameters: item (|Item|), source (path), destination (path)
  :Description: Called whenever a symlink is created for a file.

``item_hardlinked``
  :Parameters: item (|Item|), source (path), destination (path)
  :Description: Called whenever a hardlink is created for a file.

``item_reflinked``
  :Parameters: item (|Item|), source (path), destination (path)
  :Description: Called whenever a reflink is created for a file.

``item_removed``
  :Parameters: item (|Item|)
  :Description: Called every time an item (singleton or part of an album) is removed from the library (even when its file is not deleted from disk).

``write``
  :Parameters: item (|Item|), path (path), tags (dict)
  :Description: Called just before a file's metadata is written to disk. Handlers may modify ``tags`` or raise ``library.FileOperationError`` to abort.

``after_write``
  :Parameters: item (|Item|)
  :Description: Called after a file's metadata is written to disk.

``import_task_created``
  :Parameters: task (|ImportTask|), session (|ImportSession|)
  :Description: Called immediately after an import task is initialized. May return a list (possibly empty) of replacement tasks.

``import_task_start``
  :Parameters: task (|ImportTask|), session (|ImportSession|)
  :Description: Called before an import task begins processing.

``import_task_apply``
  :Parameters: task (|ImportTask|), session (|ImportSession|)
  :Description: Called after metadata changes have been applied in an import task (on the UI thread; keep fast). Prefer a pipeline stage otherwise (see :ref:`plugin-stage`).

``import_task_before_choice``
  :Parameters: task (|ImportTask|), session (|ImportSession|)
  :Description: Called after candidate search and before deciding how to import. May return an importer action (only one handler may return non-None).

``import_task_choice``
  :Parameters: task (|ImportTask|), session (|ImportSession|)
  :Description: Called after a decision has been made about an import task. Use ``task.choice_flag`` to inspect or change the action.

``import_task_files``
  :Parameters: task (|ImportTask|), session (|ImportSession|)
  :Description: Called after filesystem manipulation (copy/move/write) for an import task.

``library_opened``
  :Parameters: lib (|Library|)
  :Description: Called after beets starts and initializes the main Library object.

``database_change``
  :Parameters: lib (|Library|), model (|Model|)
  :Description: A modification has been made to the library database (may not yet be committed).

``cli_exit``
  :Parameters: lib (|Library|)
  :Description: Called just before the ``beet`` command-line program exits.

``import_begin``
  :Parameters: session (|ImportSession|)
  :Description: Called just before a ``beet import`` session starts.

``trackinfo_received``
  :Parameters: info (|TrackInfo|)
  :Description: Called after metadata for a track is fetched (e.g., from MusicBrainz). Handlers can modify the tags seen by later pipeline stages or adjustments (e.g., ``mbsync``).

``albuminfo_received``
  :Parameters: info (|AlbumInfo|)
  :Description: Like ``trackinfo_received`` but for album-level metadata.

``before_choose_candidate``
  :Parameters: task (|ImportTask|), session (|ImportSession|)
  :Description: Called before prompting the user during interactive import. May return a list of ``PromptChoices`` to append to the prompt (see :ref:`append_prompt_choices`).

``mb_track_extract``
  :Parameters: data (dict)
  :Description: Called after metadata is obtained from MusicBrainz for a track. Must return a (possibly empty) dict of additional ``field: value`` pairs to apply (overwriting existing fields).

``mb_album_extract``
  :Parameters: data (dict)
  :Description: Like ``mb_track_extract`` but for album tags. Overwrites tags set at the track level with the same field.
```
</details>

docs/dev/plugins/other.rst
```diff
@@ -0,0 +1,391 @@
...
+
+Plugins can also be used to extend the autotagger functions i.e. the metadata
+lookup from external sources. For this your plugin has to extend the
+:py:class:`MetadataSourcePlugin` base class and implement all abstract methods.
+
+On metadata lookup, the autotagger will try to find matching candidates from all
+enabled metadata source plugins. To do this, we will call the
+:py:meth:`MetadataSourcePlugin.candidates` (or
+:py:meth:`MetadataSourcePlugin.item_candidates`) with all available (local)
+metadata. The list of retrieved candidates will be ranked by their
```
snejus: 1. Clarify what each of `candidates` and `item_candidates` methods are for
2. We can remove _with all available (local) metadata_ since this may be slightly misleading in some contexts (for example, search).
3. Info about distance should probably be moved to a separate paragraph.

```suggestion
:py:meth:`MetadataSourcePlugin.candidates` for albums or
:py:meth:`MetadataSourcePlugin.item_candidates` for singletons.

The list of retrieved candidates will be ranked by their
```

docs/dev/plugins/other.rst
```diff
@@ -0,0 +1,391 @@
+Further Reading
+===============
+
+.. contents:: Table of Contents
+    :local:
+    :depth: 2
+
+Extending the Autotagger
```
snejus: Can we have autotagger docs in a separate file?

docs/dev/plugins/other.rst
```diff
@@ -0,0 +1,391 @@
...
+:py:meth:`MetadataSourcePlugin.candidates` (or
+:py:meth:`MetadataSourcePlugin.item_candidates`) with all available (local)
+metadata. The list of retrieved candidates will be ranked by their
+:py:meth:`MetadataSourcePlugin.album_distance` (or
+:py:meth:`MetadataSourcePlugin.track_distance`) and be presented to the user for
+selection (or automatically selected if the threshold is met).
+
+Please have a look at the ``beets.autotag`` and especially the
+``beets.metadata_plugin`` modules for more information. Additionally, for a
+comprehensive example, see the ``musicbrainz`` or ``chroma`` plugins, which are
```
snejus: I don't think any of these two is a great example of how `MetadataSourcePlugin` should be extended:

* `musicbrainz` has been developed in isolation and its implementation is really not great
* `chroma` is not a 'true' autotagger plugin.

I think we'd rather refer people to `spotify` / `deezer` or `discogs`.

docs/dev/plugins/other.rst
```diff
@@ -0,0 +1,391 @@
...
+Logging
+-------
+
+Each plugin object has a ``_log`` attribute, which is a ``Logger`` from the
+`standard Python logging module`_. The logger is set up to `PEP 3101`_,
+str.format-style string formatting. So you can write logging calls like this:
+
+::
+
+    self._log.debug('Processing {0.title} by {0.artist}', item)
```
snejus: Thanks for reformatting the code blocks! I think that's the last one's left
```suggestion
.. code-block:: python

    self._log.debug("Processing {0.title} by {0.artist}", item)
```

docs/dev/plugins/other.rst
```diff
@@ -0,0 +1,391 @@
...
+
+        @property
+        def album_types(self):
+            return {"rating": types.INTEGER}
+
+A plugin may define two attributes: `item_types` and `album_types`. Each of
+those attributes is a dictionary mapping a flexible field name to a type
+instance. You can find the built-in types in the `beets.dbcore.types` and
+`beets.library` modules or implement your own type by inheriting from the `Type`
+class.
```
snejus: Not sure if this was intentional but we should use double ticks (``) for inline code (applies to the rest of the changes).

```suggestion
A plugin may define two attributes: ``item_types`` and ``album_types``. Each of
those attributes is a dictionary mapping a flexible field name to a type
instance. You can find the built-in types in the ``beets.dbcore.types`` and
``beets.library`` modules or implement your own type by inheriting from the ``Type``
class.
```

docs/dev/plugins/other.rst
```diff
@@ -0,0 +1,391 @@
...
+
+Plugins provide stages as functions that take two arguments: ``config`` and
+``task``, which are ``ImportSession`` and ``ImportTask`` objects (both defined
+in ``beets.importer``). Add such a function to the plugin's ``import_stages``
+field to register it:
+
+.. code-block:: python
+
+    from beets.plugins import BeetsPlugin
+    from beets.importer import ImportSession, ImportTask
```
snejus: Nit

```suggestion
    from beets.importer import ImportSession, ImportTask
    from beets.plugins import BeetsPlugin
```

docs/dev/plugins/other.rst
```diff
@@ -0,0 +1,391 @@
...
+Extend MediaFile
+----------------
+
+MediaFile_ is the file tag abstraction layer that beets uses to make
+cross-format metadata manipulation simple. Plugins can add fields to MediaFile
+to extend the kinds of metadata that they can easily manage.
+
+The ``MediaFile`` class uses ``MediaField`` descriptors to provide access to
+file tags. If you have created a descriptor you can add it through your plugins
+:py:meth:`beets.plugins.BeetsPlugin.add_media_field()`` method.
```
snejus:
```suggestion
:py:meth:`beets.plugins.BeetsPlugin.add_media_field()` method.
```

docs/dev/plugins/other.rst
```diff
@@ -0,0 +1,391 @@
...
+:py:class:`MetadataSourcePlugin` base class and implement all abstract methods.
+
+On metadata lookup, the autotagger will try to find matching candidates from all
+enabled metadata source plugins. To do this, we will call the
+:py:meth:`MetadataSourcePlugin.candidates` (or
+:py:meth:`MetadataSourcePlugin.item_candidates`) with all available (local)
+metadata. The list of retrieved candidates will be ranked by their
+:py:meth:`MetadataSourcePlugin.album_distance` (or
+:py:meth:`MetadataSourcePlugin.track_distance`) and be presented to the user for
+selection (or automatically selected if the threshold is met).
```
snejus: A fair bit of detail has disappeared from this paragraph. If we don't want to expand about each of the methods in detail here, we should probably add the removed documentation to their docstrings and refer to them accordingly.

I think we should mention `album_for_id` and `track_for_id` methods instead of `album_distance` and `traack_distance` since the former are abstract while the latter aren't.

docs/dev/plugins/events.rst
semohr: Yeah at the moment we only export the plugin classes. But having the references already backed in allows us to reference them in the future.

docs/dev/plugins/index.rst
semohr: No it doesn't. A bit weird as the syntax is correct, might be an issue upstream in the template. Citations should work like that:
https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#citations

I tried to replace them with footnotes, but the docstrfmt tool has some [issues](https://github.com/LilSpazJoekp/docstrfmt/issues/133) with them. Last fix was quick, so let's cross our fingers for an update soon :)

docs/dev/plugins/other.rst
semohr: Added quite a bit of content.

docs/dev/plugins/other.rst
semohr: Added quite a bit of content.

docs/dev/plugins/other.rst
semohr: Added quite a bit of content.

docs/dev/plugins/other.rst
semohr: Nope wasnt intentional. Might be copy paste error when I moved reset the branch onto the docstrfmt.

docs/dev/plugins/events.rst
semohr: That is a good idea, tables are always a bit janky. Especially on mobile devices.

docs/dev/plugins/events.rst
semohr: Tried to get the substitution to work for some minutes but docstrfmt complained about it being empty. Gave up on it for time reasons, feel free to have a look too. Might be a bug upstream too.

docs/dev/plugins/other.rst
semohr: Done. Have another look also addresses the comments below.

docs/dev/plugins/index.rst
semohr: Funnily enough if the citations are numbers instead, it works as expected. I think this is a `pydata_theme` issue.

docs/dev/plugins/autotagger.rst
```diff
@@ -0,0 +1,103 @@
+Extending the Autotagger
```
snejus, line 1: Do you reckon it's worth mentioning the logic which resolves `data_source` automatically?

docs/dev/plugins/autotagger.rst
```diff
@@ -0,0 +1,103 @@
...
+        def album_for_id(self, album_id: str): ...
+
+How Metadata Lookup Works
+-------------------------
+
+When beets runs the autotagger, it queries **all enabled metadata source
+plugins** for potential matches:
+
+- For **albums**, it calls :py:meth:`~MetadataSourcePlugin.candidates`.
+- For **individual items**, it calls
```
snejus:
```suggestion
- For **singletons**, it calls
```

docs/dev/plugins/autotagger.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Implementing ID-based Lookups
+-----------------------------
+
+Your plugin must also define:
+
+- :py:meth:`~MetadataSourcePlugin.album_for_id` — fetch album metadata by ID.
+- :py:meth:`~MetadataSourcePlugin.track_for_id` — fetch track metadata by ID.
+
+These methods should return `None` if your source doesn`t support ID lookups.
```
snejus: I doubt this is needed. I reckon we can safely assume that any metadata source will be performing id lookups.

```suggestion
```

The only exception to this is the `chroma` plugin, but I'd argue it isn't a 'proper' metadata source plugin but rather an alternative track/album distance calculation for the `musicbrainz` autotagger.

Longer term I'd imagine it should be downgraded from a plugin to a configuration option under `musicbrainz` autotagger, probably.

docs/_templates/autosummary/class.rst
snejus, line 1: Seems like properties decorated with custom `cached_classproperty` don't appear in the docs. Would it be possible to include them too?

docs/api/plugins.rst
```diff
@@ -7,3 +7,11 @@ Plugins
     :toctree: generated/
 
     BeetsPlugin
+
```
snejus, line 10: For some reason `get_artist` method is documented under `BeetsPlugin` in rendered docs, it seems.

Image

docs/dev/plugins/autotagger.rst
```diff
@@ -0,0 +1,103 @@
...
+            va_likely: bool,
+        ): ...
+
+        def item_candidates(self, item: Item, artist: str, title: str): ...
+
+        def track_for_id(self, track_id: str): ...
+
+        def album_for_id(self, album_id: str): ...
+
+How Metadata Lookup Works
```
snejus:
```suggestion
Metadata lookup
```

(note only the first word is capitalised)

docs/dev/plugins/autotagger.rst
```diff
@@ -0,0 +1,103 @@
...
+The results are combined and scored. By default, candidate ranking is handled
+automatically by the beets core, but you can customize weighting by overriding:
+
+- :py:meth:`~MetadataSourcePlugin.album_distance`
+- :py:meth:`~MetadataSourcePlugin.track_distance`
+
+This is optional, if not overridden, both methods return a constant distance of
+`0.5`.
+
+Implementing ID-based Lookups
```
snejus:
```suggestion
ID-based lookups
```

docs/dev/plugins/autotagger.rst
```diff
@@ -0,0 +1,103 @@
...
+
+- :py:meth:`~MetadataSourcePlugin.album_for_id` — fetch album metadata by ID.
+- :py:meth:`~MetadataSourcePlugin.track_for_id` — fetch track metadata by ID.
+
+These methods should return `None` if your source doesn`t support ID lookups.
+IDs are expected to be strings. If your source uses specific formats, consider
+contributing an extractor regex to the core module:
+:py:mod:`beets.util.id_extractors`.
+
+Best Practices
```
snejus:
```suggestion
Best practices
```

docs/dev/plugins/autotagger.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Beets already ships with several metadata source plugins. Studying these
+implementations can help you follow conventions and avoid pitfalls. Good
+starting points include:
+
+- `spotify`
+- `deezer`
+- `discogs`
+
+Migration Guidance
```
snejus:
```suggestion
Migration guidance
```

docs/dev/plugins/autotagger.rst
```diff
@@ -0,0 +1,103 @@
...
+------------------
+
+Older metadata plugins that extend :py:class:`beets.plugins.BeetsPlugin` should
+be migrated to :py:class:`MetadataSourcePlugin`. Legacy support will be removed
+in **beets v3.0.0**.
+
+.. seealso::
+
+    - :py:mod:`beets.autotag`
+    - :py:mod:`beets.metadata_plugins`
```
snejus, line 105: We may want to refer to respective sections in the docs instead of the module in the source code.

docs/dev/plugins/other.rst
```diff
@@ -0,0 +1,368 @@
+Further Reading
```
snejus: What do you think about increasing `level` by 1 in the left sidebar? Personally, being able to see section names that comprise __Further Reading__ without having to jump into the docs is kinda nice.

Image

docs/dev/plugins/autotagger.rst
semohr: Is there a paragraph somewhere that explains singletons?

docs/dev/plugins/autotagger.rst
semohr: It might be more common than we expect, e.g. ai-sauce also does not support that. Removed it for now, should also be clear from the type-hints anyways.

docs/_templates/autosummary/class.rst
semohr: Should work in theory, might need a bit more investigation. Let's postpone that, the auto api docs need quite some work anyways. 

Much of the issues also come from the default template, which we do not edit too much at the moment.

docs/api/plugins.rst
semohr: Sadly that is the default. Will need some more investigation, these autodoc templates are not something one looks at often. 

Would keep it like this for now, we might want to switch to `sphinx-autoapi` instead of `sphinx.autodoc` too.

docs/dev/plugins/other.rst
semohr: I don't mind either way. I can't get it to work tho, I increased the maxdepth but that does not seem to change anything (both in the `dev/index` and `dev/plugin/index` file).

docs/dev/plugins/autotagger.rst
semohr: Not sure what are the relevant parts for you, added some references to the normal docs here.

docs/dev/plugins/autotagger.rst
semohr: I considered that too. Added a sentence.

docs/dev/plugins/other.rst
JOJ0: Where did you set that maxdepth option? Let me double check this. I want to build it on my machine.

semohr: In the mentioned files, e.g.

```
.. toctree::
    :maxdepth: 3
    :titlesonly:

    plugins/index
    library
    importer
    cli
```

docs/_templates/autosummary/class.rst
snejus: What if you try adding `:undoc-members:` in the class template?

semohr: It would technically work, but we’d lose the docstring in the generated documentation, which makes it pretty unhelpful in practice. The deeper issue here is with `@cached_classproperty`: the more we use it, the more it feels like an antipattern. Typing support is fragile, Sphinx integration is limited, and overall it creates quite a few problems.

I tried a few things to get the docstring propagated but nothing was successful.

docs/dev/plugins/other.rst
JOJ0: toctree settings don't change anything on my machine either. Have you tried this setting? https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/navigation.html#control-how-many-navigation-levels-are-shown-by-default

add it somewhere here. near the bottom are html_theme.. things: https://github.com/beetbox/beets/blob/master/docs/conf.py

JOJ0: or maybe it's this one: https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/navigation.html#control-the-number-of-navigation-levels

semohr: Good find! I thought there has to be something like that. Tbh tho, I would prob. not adjust the global setting just for the dev pages.

JOJ0: Hi @semohr, I pushed two commits. The main issue is that it's just not possible to reveal chapters that are only headlines in a document, so:

- I split those chapters to separate files
- And set those pydata theme settings mentioned above so that we always see them in the primary sidebar

Things I noticed:

- Do we actually want that additional "Further Reading" chapter? We could technically remove it at all, and move all the chapters one level up. I think it would still be integral without having to change content, right? The only reason to keep it in my opinion is that we don't overwhelm with subchapters at first and keep it slim to 4. So maybe that's a good thing and we want to keep it that way?
- The subchapter "Extend the query syntax" is one level lower than the others. I think this is a mistake in the original documentation already. I suggest we move it one level up. Or maybe I don't see something that makes it only work one level below content-wise? Please check!

Screenshot 2025-08-31 at 09 52 18

semohr: > Do we actually want that additional "Further Reading" chapter? We could technically remove it at all, and move all the chapters one level up. I think it would still be integral without having to change content, right? The only reason to keep it in my opinion is that we don't overwhelm with subchapters at first and keep it slim to 4. So maybe that's a good thing and we want to keep it that way?

I added it mainly for the reason that the main file was extremely overwhelming. Before the refactor the plugin dev guide was just one big file. I dont mind either way, as long as all content is not contained on a single page.

semohr: > The subchapter "Extend the query syntax" is one level lower than the others. I think this is a mistake in the original documentation already. I suggest we move it one level up. Or maybe I don't see something that makes it only work one level below content-wise? Please check!

Indeed seems wrong.

semohr: > I split those chapters to separate files

Make sure to also add the commit hash to the git blame file.

docs/api/plugins.rst
JOJ0: Ok so I invested a couple of hours this morning trying to remove inherited methods from BeetsPlugin and I give up now. I agree that moving on here is not worth the pain. Trying something else? Sure! Do it, but for now I suggest we move on with getting this PR merge-ready. This PR improves things a lot already and we can leave things for future improvements.

JOJ0: Just one thing if maybe in the future you want to keep working on it here or somewhere else in the future. I tried to exclude stuff from BeetsPlugin like this:

```
{% if objname == 'BeetsPlugin' %}
.. autoclass:: {{ objname }}
   :members:
   :private-members:
   :show-inheritance:
{% else %}
.. autoclass:: {{ objname }}
   :members:
   :private-members:
   :show-inheritance:
   :inherited-members:
{% endif %}

   {% block methods %}
   .. automethod:: __init__

<and so on...>
```

or

```
   .. autosummary::
   {% for item in methods %}
      {%- if not (objname == 'BeetsPlugin' and item == 'get_artist') %}
      ~{{ name }}.{{ item }}
      {%- endif %}
   {%- endfor %}

   <and so on...>

```

docs/dev/plugins/other.rst
JOJ0: I don't understand. git blame file?

semohr: I mean the git-blame-ignore file. As essentially the changes are only file moves, we don't want the changes to pollute the history. At least this is how I understand it :) 

I have already added one commit to the ignore file in this PR. We might want to add the commit which moves the docs from other.rst into sperate files aswell.

JOJ0: > I added it mainly for the reason that the main file was extremely overwhelming. Before the refactor the plugin dev guide was just one big file. I dont mind either way, as long as all content is not contained on a single page. 

Ah ok I understand. If that was your intention I totally agree to keep it like that. The 4 main chapters to get an idea and not overwhelm and to dig deeper there is more. Good thinking!

JOJ0: I've just added that commit to the blame ignore file. Also I fixed one poe format issue and included it in the original commit, thus force-pushed!

JOJ0:
@snejus @semohr what do you think? good enough for now and ready to merge (after resolving conflicts)? I'd say yes and will hit that approve button once I find it... ;-)

semohr:
@JOJ0 LGTM. I just rebased it on main. Feel free to merge

docs/api/plugins.rst
JOJ0: ...and then it bothered me... Why the hell would Sphinx render members of a child class in its parent class? In our case, why would BeetsPlugin show members of MetadataSourcePlugin.

After countless and useless prompting Claude Sonnet 4 and GPT5 a very direct "Why the crap would Sphinx do that?", I got the answer: The reason is our legacy/compatibility code where we copy all MetadataSourcePlugin stuff back to BeetsPlugin here: https://github.com/beetbox/beets/blob/6f61488af9ce107f01526aba04cf0ce8879f0b0f/beets/metadata_plugins.py#L421-L428

GPT5 (preview)'s full answer:
> Root cause: BeetsPlugin shows get_artist because metadata_plugins.py dynamically copies every function from MetadataSourcePlugin onto BeetsPlugin for legacy support (the for inspect.getmembers(...) loop near the end). That literally adds get_artist (and the other interface methods) to BeetsPlugin at runtime, so autodoc correctly sees them as real members. They’re not inherited—they’re monkey‑patched. We added a tracking list _legacy_metadata_methods and updated skip_member to hide those on the BeetsPlugin docs. After rebuilding, those copied methods should disappear from the BeetsPlugin page while remaining on MetadataSourcePlugin.

I tried the suggested fix with a "legacy collect magic" so Sphinx can succesfully ignore those methods in BeetsPlugin at least - but no success and again: I give up! It's super-annoying!

So @snejus and @semohr removing this legacy code is the only way I see to really make the docs right. So please help me decide here how to move on:

- Do we really need this legacy code? Do we have an idea of how many 3rd party source plugins we would break out there if we just remove it?
- Keep the badly rendered BeetsPlugin docs as-is? (pretty misleading, I don't like it....and yes I kind of reconsidered from Yesterday-Jojo's opiniont that this is good enough :-))
- Remove BeetsPlugin docs page entirely? (quickfix)
- Write a manual .rst file that only lists the BeetsPlugin methods we want - one by one? (more work but if one of you would want to take over again? since I'm running out of time here!)

Sorry for the longish post and thanks for reading and your thoughts! :-)

JOJ0:
> @semohr note you can just run `poe docs` from the repo root - no need for `cd docs && make docs`

I'll add `clean`to this `poe`command @snejus, when working with Sphinx builds it's pretty essential IMO to always include that, otherwise the outcome is unpredictable. It takes a second longer to build but that's better than getting docs that won't be the real docs anyway! ;-)

docs/api/plugins.rst
JOJ0: (and hey don't get me wrong @semohr you did amazing work here! I love the autosummary overview tables with methods+docstrings only! Even though I was fighting with all this for hours I learned a lot! i didn't even know that features like this and templates with jinja support even existed! So many thanks! Much appreciated!)

JOJ0: Anf for the record: Yes I did test that removing that mentioned legacy-keep-code at the end of metadata_plugins.py would fix the docs to then render properly:


Screenshot 2025-09-03 at 07 28 48

semohr: Maybe it is possible to add a deprecated marker to the docstring of the decorated functions `__doc__ += deprecated` .

I would keep it as is for now. It is an upgrade either way and we are only talking about dev docs anyways.

Thanks that you also had a look! Sphinx is often one of the most frustrating parts for me, seems very hacky in general and I absolutly hate jinja.

JOJ0: Please try and add such a deprecated marker, I'm out for today! I agree that something like that and moving on with our lives without Sphinx for a while is good!  🤣

snejus:
Please keep it in place since a 'clean' build is considerably longer than just 1s:

A clean build from zero:
```sh
$ poe docs
...
46s 266ms 162us | 60 pids
```

A rebuild __without cleanup__ after adjusting a line in `docs/plugins/lyrics.rst`:
```sh
$ poe docs
...
12s 427ms 549us | 24 pids
```

snejus:
> when working with Sphinx builds it's pretty essential IMO to always include that, otherwise the outcome is unpredictable

Could you expand by any chance? Not sure if I have ever experienced this: on my end, Sphinx always updates all HTML files that are relevant for the documentation. I don't think I've ever come across a page that was somehow out of sync.

semohr:
I have the same issue quite often, especially jinja template or tocs not updating correctly. I don't mind either way, since Im mostly building the docs with `make clean` and `make html` tho :shrug:

snejus:
We seem to have introduced an inconsistency here, see the **Reference** page:
image

Previously, the **For developers** page had the same structure (high level section navigation on the left, while the more detailed tree is within the page on the right
 image


However, now this page seems to follow the opposite pattern:

image

JOJ0:
> We seem to have introduced an inconsistency here, see the **Reference** page: image Previously, the **For developers** page had the same structure (high level section navigation on the left, while the more 


> However, now this page seems to follow the opposite pattern:
> 

Yep, this is just how the pydata theme works. On-page nav vs. "several page" nav - primary sidebar vs. secondary sidebar.

Do you feel this is a major show-stopper? I just wanted to click that merge button and I don't want to revert all the efforts @semohr put in restructuring dev-docs already. So either this gets a neverending story or we just live with this inconsistency for now. What do you think? 

In my opinion the other docs pages (not dev) have the same problem @semohr already addressed with dev docs - one huge page for everything which is hard to comprehend. so maybe a restructure is in order there as well (but possibly not now aka never-ending-story ;-)

docs/dev/plugins/events.rst
snejus: Managed to make it work once I defined global substitutions in `docs/conf.py`:

```py
# Global substitutions that can be used anywhere in the documentation.
rst_epilog = """
.. |Album| replace:: :class:`~beets.library.models.Album`
.. |AlbumInfo| replace:: :class:`beets.autotag.hooks.AlbumInfo`
.. |ImportSession| replace:: :class:`~beets.importer.session.ImportSession`
.. |ImportTask| replace:: :class:`~beets.importer.tasks.ImportTask`
.. |Item| replace:: :class:`~beets.library.models.Item`
.. |Library| replace:: :class:`~beets.library.library.Library`
.. |Model| replace:: :class:`~beets.dbcore.db.Model`
.. |TrackInfo| replace:: :class:`beets.autotag.hooks.TrackInfo`
"""
```

docs/_templates/autosummary/class.rst
snejus: Have you tried this? 

```py
    data_source: str = cached_classproperty(
        lambda cls: cls.__name__.replace("Plugin", "")
    )  # type: ignore[assignment]
    """The data source name for this plugin."""
```

No need for this for now though

docs/api/plugins.rst
snejus: Guys, I think we forgot that this exists... 😁 

```py
# Dynamically copy methods to BeetsPlugin for legacy support
# TODO: Remove this in the future major release, v3.0.0

for name, method in inspect.getmembers(
    MetadataSourcePlugin, predicate=inspect.isfunction
):
    if not hasattr(BeetsPlugin, name):
        setattr(BeetsPlugin, name, method)
```

snejus: Fixed it

docs/dev/plugins/events.rst
snejus: Adjusted it

snejus:
I've added 4 commits:
* Defined replacements for the commonly used classes (see https://github.com/beetbox/beets/pull/5861#discussion_r2296830209)
* Removed `MetadataSourcePlugin` methods from `BeetsPlugin`: moved the logic under `BeetsPlugin.__init_subclass__` where they are added to plugins that subclass `BeetsPlugin`. See https://github.com/beetbox/beets/pull/5861#discussion_r2302776225
* Fixed formatting issues (need to use double ``` `` ``` to format code inline). For some reason, our linter isn't catching this :/
* Made the **For developers** page look somewhat more consistent with the rest (as mentioned in my previous comment).
* Added missing `docs/api/index.rst` to make sure this section appears is in the main **Contents** section and children TOCs
  * Before: 
    image 
  * After: 
    image

That's it from my end I think.

snejus (approved):

snejus:
Ah one more thing:

```toml
[tool.poe.tasks.docs]
help = "Build documentation"
args = [{ name = "COMMANDS", positional = true, multiple = true, default = "html" }]
cmd = "make -C docs $COMMANDS"
```

Adjusted it back to run just `html` by default, but you can now alternatively run `poe docs clean html` @JOJ0

semohr:
@snejus I think we have to be careful with rebases if we use the git-blame-ignore file as a rebase changes the hashes. Changed it to the new hashes. 

Good enough to merge from my side :upside_down_face:

snejus:
Apologies! Completely forgot that `.git-blame-ignore-revs` was adjusted here 🤦🏼

docs/api/plugins.rst
JOJ0: > 

Didn't I just mention that here @snejus ;-) https://github.com/beetbox/beets/pull/5861#discussion_r2317821694

Thanks so much for fixing it! That really was misleading and is so much better now!

JOJ0:
> > when working with Sphinx builds it's pretty essential IMO to always include that, otherwise the outcome is unpredictable
> 
> Could you expand by any chance? Not sure if I have ever experienced this: on my end, Sphinx always updates all HTML files that are relevant for the documentation. I don't think I've ever come across a page that was somehow out of sync.

Especially when restructuring things - you WILL get problems. I have it always when working with Sphinx :-) I had many frustrating WTF moments and will never forget about including clean anymore ;-) 

I actually was assuming that `poe docs` includes `clean` - it's that essential and logic to me. But anyway! Now I know that it isn't....

One more aspect of why I still think that it's useful to always include it is to give contributors a super-failsafe way of working with it. In my opinion it's better to wait 5 times longer than having them fight with inconsistencies and they give up and learn to hate fixing beets' docs because it's annoying.

snejus:
> I actually was assuming that `poe docs` includes `clean` - it's that essential and logic to me. But anyway! Now I know that it isn't....

Note you can now run `poe docs clean html` and it will clean them before building!


> In my opinion it's better to wait 5 times longer than having them fight with inconsistencies and they give up and learn to hate fixing beets' docs because it's annoying.

Personally, I would give up before I even find any inconsistencies if I had to wait too long 😁😁

PR #5867, opened 2025-07-09
Unimported plugin - new option to add globbing

Mikaciu:
## Description

Add a new option, `ignore_as_globs` to consider the `ignore_subdirectories` as globs, the same way the [`ignore`](https://beets.readthedocs.io/en/stable/reference/config.html#ignore) section works.

Using this flag, the `unimported` plugin is walking through the library using the `util.sorted_walk` method, the same way the importer does.

## To Do
- I didn't find a way to access the beets config from a plugin (maybe due to the fact that plugin config is namespaced). An improvement would have been to use the `ignore` config value instead of duplicating it.
- [x] Changelog. (Add an entry to `docs/changelog.rst` to the bottom of one of the lists near the top of the document.) <-- will be done after review ;)


snejus (commented):
Hi, thanks for this, and apologies for a late review. Would you mind adding a test for this?

snejus:
Closing due to lack of author's response - please reopen if you're still around!

Mikaciu:
I'll have a look ;)

PR #5870, opened 2025-07-13
Fix a pair of regressions, types and enforce mypy in CI

snejus:
**Bug Fixes**:
   - Wrong `raw_seconds_short` module in `DurationQuery`.
   - Missing `byte_paths` variable in `import_func`.

Otherwise

- Addressed all static type checking errors in the codebase (except `spotify` plugin which has already been fixed in a separate PR)
- Removed `continue-on-error` from in `mypy` CI job, which means any errors in type checking will now fail the CI job.


beets/importer/stages.py
```diff
@@ -143,9 +144,7 @@ def lookup_candidates(session: ImportSession, task: ImportTask):
 
     # Restrict the initial lookup to IDs specified by the user via the -m
     # option. Currently all the IDs are passed onto the tasks directly.
-    task.search_ids = session.config["search_ids"].as_str_seq()
```
semohr, line 146: Good change :tada: 

Makes it easier for us (beets-flask) to use the task functions without the need for complex config patching.

snejus:
> I expected way more typing issues to be left as I'm normally only running mypy on a specific file. Awesome that we are done with that 🎉
> 
> For this branch I'm still seeing the following (non spotify) issues locally
> 
> ```shell
> test/plugins/test_player.py:40:5: error: Argument 1 to "module_from_spec" has incompatible type "Optional[ModuleSpec]"; expected "ModuleSpec"  [arg-type]
> test/plugins/test_player.py:49:1: error: Module has no attribute "_GstPlayer"  [attr-defined]
> test/plugins/test_player.py:71:1: error: Module has no attribute "GstPlayer"  [attr-defined]
> test/plugins/test_player.py:941:6: error: Argument 1 to "skip" has incompatible type "Callable[[BPDPlaylistsTest], Any]"; expected "str"  [arg-type]
> test/plugins/test_player.py:1131:6: error: Argument 1 to "skip" has incompatible type "Callable[[BPDConnectionTest], Any]"; expected "str"  [arg-type]
> ```

Good point, I forgot to check the `test` folder locally. Fixed now!

semohr:
LGTM. Maybe we want to merge the metadata PR first? As it includes the missing pieces for spotify :thinking:

PR #607, opened 2014-03-13
Refactor plugin

geigerzaehler:
In this PR we try to develop a new plugin API. The API has two parts: the `BeetsPlugin` class to be used by plugins and the functions beets uses to interact with plugins. See also the [wiki](https://github.com/sampsyo/beets/wiki/Refactoring#wiki-plugin-api-overhaul).

At the moment the only change is the addition of a `Registry` class to encapsulate all interactions with plugins.
##### TODO
- [ ]  Use `BeetsPlugin` methods to expose event handlers. If I understood @sampsyo in #603 correctly this means that a plugin simply needs to implement a `on_write` method to register a 'write' event handler.
- [ ]  Think of a clever way to expose the default registry without a calling `plugins.registry` or migrate all code to the new registry API.
- [x] ~~Make subclasses of `BeetsPlugin` singeltons by overwriting the `__new__` function.~~ [Obsolete](#issuecomment-38404746)
- [x] Clarify the usage of namespace packages. At the moment I’m leaning towards keeping then, since they actually [work quite well](https://gist.github.com/geigerzaehler/9508410) if we are careful.
- [ ] Devise a way to migrate to the new API or provide a compatibility layer.

PR #5872, opened 2025-07-15
Register types and queries from plugins on models dynamically

snejus:
## Fix dynamic plugin type and query registration

This PR refactors the plugin system to properly handle dynamic type and query registration by converting static class attributes to cached class properties.

**Problem**: Plugin types and queries were stored as mutable class attributes that were manually updated during plugin loading/unloading. This caused issues where:
- Plugin types weren't properly registered in test environments
- Shared mutable state between test runs caused inconsistent behavior  
- Manual cleanup was error-prone and incomplete

See https://github.com/beetbox/beets/pull/5833 and specifically https://github.com/beetbox/beets/pull/5833#issuecomment-3016635209 for the context.

**Solution**: 
- Convert `_types` and `_queries` from static dictionaries to `@cached_classproperty` decorators
- Types and queries are now dynamically computed from loaded plugins when accessed
- Eliminates manual mutation of class attributes during plugin loading
- Properly clears cached properties when plugins are loaded/unloaded
- Ensures plugin types are available immediately after registration

**Key Changes**:
- `Model._types` and `LibModel._queries` now use `@cached_classproperty` 
- Removed manual `_types.update()` calls in plugin loading code
- Added proper cache clearing in test infrastructure
- Plugin types are now inherited through the class hierarchy correctly

This fixes the developer's issue where `item_types` weren't being registered properly in tests - the new dynamic property approach ensures plugin types are available as soon as plugins are loaded.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -289,9 +289,10 @@ class Model(ABC, Generic[D]):
     terms.
     """
 
-    _types: dict[str, types.Type] = {}
-    """Optional Types for non-fixed (i.e., flexible and computed) fields.
-    """
+    @cached_classproperty
+    def _types(cls) -> dict[str, types.Type]:
+        """Optional types for non-fixed (flexible and computed) fields."""
+        return {}
 
     _sorts: dict[str, type[FieldSort]] = {}
     """Optional named sort criteria. The keys are strings and the values
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -299,10 +299,12 @@ class Model(ABC, Generic[D]):
     are subclasses of `Sort`.
     """
 
-    _queries: dict[str, FieldQueryType] = {}
-    """Named queries that use a field-like `name:value` syntax but which
-    do not relate to any specific field.
-    """
+    @cached_classproperty
+    def _queries(cls) -> dict[str, FieldQueryType]:
+        """Named queries that use a field-like `name:value` syntax but which
+        do not relate to any specific field.
+        """
+        return {}
 
     _always_dirty = False
     """By default, fields only become "dirty" when their value actually
--- a/beets/library/models.py
+++ b/beets/library/models.py
@@ -42,6 +42,14 @@ class LibModel(dbcore.Model["Library"]):
     path: bytes
 
     @cached_classproperty
+    def _types(cls) -> dict[str, types.Type]:
+        """Return the types of the fields in this model."""
+        return {
+            **plugins.types(cls),  # type: ignore[arg-type]
+            "data_source": types.STRING,
+        }
+
+    @cached_classproperty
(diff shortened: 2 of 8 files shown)


snejus:
> Looks mostly good to me. A short question, shouldn't we clear or update the cache whenever the load_plugins function is called? I know for beets itself it probably shouldn't matter but might for some external apps.

Plugins are only ever loaded once, right? If some external app does something shady and wants to somehow force reload them, it will have to manually modify global variables in `beets/plugins.py`. If some app goes this far, then it's going to be their responsibility to clear the cache too 😅

We don't have a functional requirement to do this except in tests. I considered adding a classmethod `reset` to `cached_classproperty` to make it more convenient to clear the cache, but decided against it _purely_ because this is only required by tests.

semohr:
I thought we overwrite the global plugin instances on each `load_plugins` call but it seems like we just add to it. So having a cache clear inside does not really make sense, you are right.

I would highly appreciate a common `unload_plugins(names)` functions tho, as we (beets-flask) are starting to juggle with multiple beets environments in the same runtime and that would be very helpful :upside_down_face:. We could place it in test helpers as we could make use of this function is tests (e.g. #5878 could make use of it).

From an external perspective it  was quite difficult to follow side effects of plugin imports as they are not idempotent.

snejus:
> I would highly appreciate a common `unload_plugins(names)` functions tho, as we (beets-flask) are starting to juggle with multiple beets environments in the same runtime and that would be very helpful 🙃. We could place it in test helpers as we could make use of this function is tests (e.g. #5878 could make use of it).

Does `beets.test.helper.PluginMixin` not fit the bill here?


> From an external perspective it was quite difficult to follow side effects of plugin imports as they are not idempotent.

I know exactly what you mean - it took me a while to get a good grasp around them. Honestly, it's so complicated __simply__ because the logic is scattered across multiple modules: most of the responsibility falls on `beets.plugins`, but some is taken over by `beets.ui.__init__`. I have a branch where I simplified it all and moved everything under a single function in `beets.plugins`, so I may as well rebase it and submit a PR.

semohr:
> Does beets.test.helper.PluginMixin not fit the bill here?

Sadly not completely, an external package can't call the class methods easily. Further, the cleanup is missing the unimport (del sys.modules) call and you can't target specific plugins. 

Maybe we could streamline this a bit in the future. Might be helpful for understanding too. 

>  I have a branch where I simplified it all and moved everything under a single function in beets.plugins, so I may as well rebase it and submit a PR.

Sounds reasonable! Feel free to do, I will definitely have a look. I think we could also get rid of the `_classes` and `_instances` global objects and could go for a singleton pattern. Might be simpler but may need quite a bit of refactoring.

wisp3rwind:
Note that there's an earlier draft of a refactored plugin registry: https://github.com/beetbox/beets/pull/607

PR #5873, opened 2025-07-15
Update deps to address security vulnerabilities

snejus:
This PR updates dependencies to address security vulnerabilities

### Key changes
- **GitHub Actions:** Updated `tj-actions/changed-files` from v44/v45 to v46 across all workflows
- **Python package updates:**
image

### DIFF
--- a/.github/workflows/changelog_reminder.yaml
+++ b/.github/workflows/changelog_reminder.yaml
@@ -1,6 +1,6 @@
 name: Verify changelog updated
 
-on: 
+on:
   pull_request_target:
     types:
       - opened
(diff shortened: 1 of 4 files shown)


wisp3rwind (approved):
Looks sensible to me :+1: 

FWIW, the new `pathspec` dependency comes from https://github.com/python/mypy/pull/18696. Some discussion around that here: https://github.com/python/mypy/issues/19275

Pygments has become a new dep of `pytest`, which is why it's non-optional now.

I'm a bit puzzled at how this fits with closing https://github.com/beetbox/beets/pull/5749; but anyway, let's merge this!

snejus:
@wisp3rwind I see how closing #5749 is confusing: the main reason why I ended up updating the action here was that I wanted to tick off each of the issues reported [here](https://github.com/beetbox/beets/security/dependabot). I don't think I had initially realised that the issue with the action was present in this list.

Issue #5876, opened 2025-07-16
beet command fail after #5787


arogl:
beet commands "fail" after #5787 

it is a soft error, as the command continues

### Problem

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv ls -ap "Natalie Jane" | sort
```

Led to this problem:

```
user configuration: C:\Users\user\AppData\Roaming\beets\config.yaml
data directory: C:\Users\user\AppData\Roaming\beets
plugin paths: C:\Users\user\AppData\Roaming\beets\beetsplug
** error loading plugin chroma:
Traceback (most recent call last):
  File "E:\Git\beets\beets\plugins.py", line 291, in load_plugins
    namespace = __import__(modname, None, None)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Git\beets\beetsplug\chroma.py", line 29, in <module>
    from beets.metadata_plugins import MetadataSourcePlugin, TrackInfo
ImportError: cannot import name 'TrackInfo' from 'beets.metadata_plugins' (E:\Git\beets\beets\metadata_plugins.py)

inline: adding item field first_artist
lastgenre: Loading canonicalization tree E:\Git\beets\beetsplug\lastgenre\genres-tree.yaml
fetchart: google: Disabling art source due to missing key
fetchart: lastfm: Disabling art source due to missing key
Sending event: pluginload
library database: C:\Users\user\AppData\Roaming\beets\library.db
library directory: E:\Music
Sending event: library_opened
Parsed query: AndQuery([OrQuery([SubstringQuery('album', 'Natalie Jane', fast=True), SubstringQuery('albumartist', 'Natalie Jane', fast=True), SubstringQuery('genre', 'Natalie Jane', fast=True)])])
Parsed sort: NullSort()
Sending event: cli_exit
```

Here's a link to the music files that trigger the bug (if relevant):

### Setup

* OS: Microsoft Windows [Version 10.0.19045.6093]
* Python version: Python 3.12.10
* beets version: Head f67d40c7db7b79d49daef4a75c8fbc450cf4007c
* Turning off plugins made problem go away (yes/no): yes

My configuration (output of `beet config`) is:

```
Not relevant as the chroma plugin was updated by this commit
```
From VSCode, I can follow the import links ok.

<img width="1095" height="602" alt="Image" src="https://github.com/user-attachments/assets/c830f168-0058-4a95-8946-b6ece7ea0f4c" />

[...]

PR #5878, opened 2025-07-16
Added a load all plugins test.

semohr:
Adds a test that loads all available plugins in the beetsplug namespace. This came up during #5876.

### DIFF
--- a/test/test_plugins.py
+++ b/test/test_plugins.py
@@ -570,6 +570,11 @@ class TestImportAllPlugins(PluginMixin):
         yield
         self.unimport_plugins()
 
+    @pytest.mark.skipif(
+        os.environ.get("GITHUB_ACTIONS") != "true",
+        reason="Requires all dependencies to be installed, "
+        + "which we can't guarantee in the local environment.",
+    )
     @pytest.mark.parametrize("plugin_name", get_available_plugins())
     def test_import_plugin(self, caplog, plugin_name):  #
         """Test that a plugin is importable without an error using the
@@ -581,22 +586,9 @@ class TestImportAllPlugins(PluginMixin):
 
         # Check for warnings, is a bit hacky but we can make full use of the beets
         # load_plugins code that way
-        # We skip ModuleNotFoundError to allow local pytest runs to pass if plugin
-        # dependencies are not installed e.g. librosa for autobpm
-        records = []
-        pattern = r"ModuleNotFoundError: No module named '(.*?)'"
-        for record in caplog.records:
-            match = re.search(pattern, str(record))
-            if match:
-                module_name = match.group(1)
-                if not self._is_spec_available(module_name):
-                    # If the module is not found, we skip it
-                    continue
-            records.append(record)
-
-        assert len(records) == 0, (
+        assert len(caplog.records) == 0, (
             f"Plugin '{plugin_name}' has issues during import. ",
-            records,
+            caplog.records,
         )
 
     def _is_spec_available(self, spec_name):
--- a/test/test_plugins.py
+++ b/test/test_plugins.py
@@ -19,7 +19,6 @@ import itertools
 import logging
 import os
 import pkgutil
-import re
 import sys
 from unittest.mock import ANY, Mock, patch
 
@@ -582,7 +581,7 @@ class TestImportAllPlugins(PluginMixin):
 
         caplog.set_level(logging.WARNING)
         caplog.clear()
-        plugins.load_plugins(plugin_name)
+        plugins.load_plugins([plugin_name])
 
         # Check for warnings, is a bit hacky but we can make full use of the beets
         # load_plugins code that way
--- a/test/test_plugins.py
+++ b/test/test_plugins.py
@@ -36,7 +36,12 @@ from beets.importer import (
 )
 from beets.library import Item
 from beets.test import helper
-from beets.test.helper import AutotagStub, ImportHelper, TerminalImportMixin
+from beets.test.helper import (
+    AutotagStub,
+    ImportHelper,
+    PluginMixin,
+    TerminalImportMixin,
+)
 from beets.test.helper import PluginTestCase as BasePluginTestCase
 from beets.util import displayable_path, syspath
 
@@ -550,17 +555,14 @@ def get_available_plugins():
     return [name for name in plugin_names if name != "_typing"]
 
 
-class TestImportAllPlugins:
+class TestImportAllPlugins(PluginMixin):
     @pytest.mark.parametrize("plugin_name", get_available_plugins())
     def test_import_plugin(self, caplog, plugin_name):  #
         """Test that a plugin is importable without an error using the
         load_plugins function."""
 
-        # Try to load all plugins
-        from beets.plugins import load_plugins
-
         caplog.set_level(logging.WARNING)
-        load_plugins([plugin_name])
+        self.load_plugins(plugin_name)
 
         # Check for warnings, is a bit hacky but we can make full use of the beets
         # load_plugins code that way
@@ -575,12 +577,13 @@ class TestImportAllPlugins:
                     continue
             records.append(record)
 
+        self.unload_plugins()
+        self.unimport_plugins()
+
         assert len(records) == 0, (
             f"Plugin '{plugin_name}' has issues during import. ",
             records,
         )
-
-        self.unimport_plugins()
 
     def _is_spec_available(self, spec_name):
         """Check if a module is available by its name."""
--- a/test/test_plugins.py
+++ b/test/test_plugins.py
@@ -13,8 +13,13 @@
 # included in all copies or substantial portions of the Software.
 
 
+import importlib
+import importlib.util
 import itertools
+import logging
 import os
+import pkgutil
+import re
 from unittest.mock import ANY, Mock, patch
 
 import pytest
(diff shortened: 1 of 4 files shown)


semohr:
It might be worth revisiting the plugin loading infrastructure, particularly with respect to side effects. Ideally, loading and unloading a plugin should be safe and idempotent. However, in practice, it appears that loading a plugin currently introduces significant side effects.

A list with plugin loading side effects would be useful for the dev docs.

test/test_plugins.py
```diff
@@ -531,3 +537,62 @@ def foo(self, session, task):
...
+            records,
+        )
+
+        self.unimport_plugins()
+
+    def _is_spec_available(self, spec_name):
+        """Check if a module is available by its name."""
+        return importlib.util.find_spec(spec_name) is not None
+
+    def unimport_plugins(self):
```
jackwilsdon: The test failures might be caused by not unloading plugins properly, as it seems like the tests only start failing after running this one.

To unload plugins, you don't just need to remove them from `sys.modules`, you also need to make beets forget about them.

Your best bet is to make this test class extend `beets.test.helper.PluginTestCase` and use `self.load_plugin(plugin_name)` and `self.unload_plugins()`.

semohr: Ah thanks! I just reviewed #5872 which will hopefully fix this too. So in theory no change should be necessary here :crossed_fingers:

test/test_plugins.py
```diff
@@ -531,3 +542,63 @@ def foo(self, session, task):
...
+
+def get_available_plugins():
+    """Get all available plugins in the beetsplug namespace."""
+    namespace_pkg = importlib.import_module("beetsplug")
+    plugin_names: list[str] = []
+    for _, module_name, _ in pkgutil.iter_modules(namespace_pkg.__path__):
+        plugin_names.append(module_name)
+
+    # skip _typing
+    return [name for name in plugin_names if name != "_typing"]
```
snejus: This can be slightly simplified:

```suggestion
    return [m.name for m in pkgutil.iter_modules() if not m.name.startswith("_")]
```

You can also make the check a bit more generic

test/test_plugins.py
```diff
@@ -531,3 +542,63 @@ def foo(self, session, task):
...
+        self.unimport_plugins()
+
+        assert len(records) == 0, (
+            f"Plugin '{plugin_name}' has issues during import. ",
+            records,
+        )
+
+    def _is_spec_available(self, spec_name):
+        """Check if a module is available by its name."""
+        return importlib.util.find_spec(spec_name) is not None
```
snejus: What is this for? Didn't `pkgutil.iter_modules` only return legitimate names?

test/test_plugins.py
```diff
@@ -531,3 +542,63 @@ def foo(self, session, task):
...
+        from beets.plugins import load_plugins
+
+        caplog.set_level(logging.WARNING)
+        caplog.clear()
+        load_plugins(plugin_name)
+
+        # Check for warnings, is a bit hacky but we can make full use of the beets
+        # load_plugins code that way
+        records = []
+        pattern = r"ModuleNotFoundError: No module named '(.*?)'"
```
snejus: Why are we looking _specifically_ for this issue only? What about any other issues?

test/test_plugins.py
```diff
@@ -531,3 +542,63 @@ def foo(self, session, task):
...
+            match = re.search(pattern, str(record))
+            if match:
+                module_name = match.group(1)
+                if not self._is_spec_available(module_name):
+                    # If the module is not found, we skip it
+                    continue
+            records.append(record)
+
+        self.unload_plugins()
+        self.unimport_plugins()
```
snejus: Place this in an independent `autouse` fixture:

```py
    @pytest.fixture(autouse=True)
    def clean_plugins(self):
        yield
        
        self.unload_plugins()
 
        for mod in list(sys.modules):
            if mod.startswith("beetsplug."):
                del sys.modules[mod]
```

**Note**: if you call `list` on `sys.modules`, Python won't complain about the _dict changed size during iteration_ (or whatever is the error you may have seen, I guess).

test/test_plugins.py
```diff
@@ -531,3 +542,63 @@ def foo(self, session, task):
...
+    return [name for name in plugin_names if name != "_typing"]
+
+
+class TestImportAllPlugins(PluginMixin):
+    @pytest.mark.parametrize("plugin_name", get_available_plugins())
+    def test_import_plugin(self, caplog, plugin_name):  #
+        """Test that a plugin is importable without an error using the
+        load_plugins function."""
+
+        from beets.plugins import load_plugins
```
snejus: I don't think you need this import as you can just use `plugins.load_plugins`.

test/test_plugins.py
```diff
@@ -531,3 +542,63 @@ def foo(self, session, task):
...
+
+        caplog.set_level(logging.WARNING)
+        caplog.clear()
+        load_plugins(plugin_name)
```
snejus: This test is not guaranteed to run _before_ tests that may import plugins from `beetsplug` package. Maybe we should unload / unimport plugins _before_ this test rather than _after_?

test/test_plugins.py
semohr: In combination with the `_is_spec_available` function (see below) this is used to skip plugins only if they fail to load with a ModuleNotFoundError. This only happens if a dependency of a plugin is not installed. 

Locally this allows that the tests still finish if e.g. librosa is not installed.
Improved the comment a bit :upside_down_face:

test/test_plugins.py
semohr: I think at the moment we run all tests are in serial, dont we? Assuming all other tests are also idempotent, it should be fine ;)

But yeah would probably be the best to just clear the plugins before and after. Updated the code to call the cleanup before and after.

test/test_plugins.py
snejus: Just exclude them explicitly - I'm not convinced that dynamism is worth the complexity cost

test/test_plugins.py
semohr: Yes but we check for different things here. This function just check if a dependency of a plugin is installed. 

Not the prettiest thing but a workaround as we dont know if a plugin needs a dependency and which one is optional. See also [above](https://github.com/beetbox/beets/pull/5878/files/176544535bf510949e08aea82f031a175a12ea95#r2216986540).

test/test_plugins.py
semohr: I would argue keeping a list of plugins with optimal dependencies is a bit more cumbersome? We might be able to parse the pyproject toml for that, seems more complex and unmaintainable if you ask me.

snejus: Wait.. What am I saying... In CI, on `ubuntu-latest` and Python 3.9 _all_ plugins are expected to load (at least the ones that would get imported by tests). Maybe we can make this test run in this environment only? See `.github/workflows/ci.yaml`: you can check for `IS_MAIN_PYTHON == "true"` in the environment.

snejus: And of course we can skip it locally by checking presence of GITHUB_ACTIONS env var (see other tests that use it)

semohr: So you'd skip the tests entirely when running locally? We could do that, but honestly, the current solution, while a bit hacky, still feels more useful.

snejus: > So you'd skip the tests entirely when running locally?

Yes, in order to reduce complexity and make these tests deterministic. This way, one at least knows how to activate them locally, simply provide the relevant env variable: `IS_MAIN_PYTHON=true poe test`.

Generally, tests need to be as dumb as possible: otherwise, if they have conditional logic, we end up in a situation where _tests_ need to be **tested**.

I think, I see another issue with the current approach: seems like it will ignore legitimate issues due to missing plugin dependencies.

test/test_plugins.py
snejus: > I think at the moment we run all tests are in serial, dont we? Assuming all other tests are also idempotent, it should be fine ;)

I don't think I understand what you mean by idempotency here.

Relatedly, see `test_discogs.py` as an example:

```py
# test_discogs.py
...
from beetsplug.discogs import DiscogsPlugin
...
```

If there's a problem with the import, this exception will be raised before tests had a chance to start. Is this something we should consider/and or handle somehow?

semohr: > If there's a problem with the import, this exception will be raised before tests had a chance to start. Is this something we should consider/and or handle somehow?

Yes but only for plugins which are actually properly tested. I just started this PR as this is actually not the case. Otherwise we would have catched this: https://github.com/beetbox/beets/issues/5876

Idempotent just means that test do not have side effects i.e. we clean up after them properly. Which is not the case for plugin test as they e.g. populate the global _classes variable. I'm just saying that we would not have to make sure everything is unloaded **before** the test, if everything actually was idempotent. More of a side note, not meant too seriously ;)

semohr:
We run into some interesting issues now, which we wanted to catch with this test in the first place!

- [x] absubmit: distutils was removed with python 3.12
    - replaced with shutil.which
- bpd: Extras gstreamer not installed in tests for python != 3.9
- chroma: Extras not installed in tests at all (pyacoustid missing)
- sonosupdate: Extras not installed in tests at all (soco missing)

How to we want to approach the tests in the future? We could e.g. add all dependencies to the CI run.

semohr:
@snejus I added some fixes for the failing plugin import. And an exception for gstreamer plugins on windows. We still fine with the changes?

test/test_plugins.py
```diff
@@ -13,8 +13,13 @@
 # included in all copies or substantial portions of the Software.
 
 
+import importlib
+import importlib.util
```
snejus: Isn't it enough just to import `importlib`?
```suggestion
```

snejus (approved):

Issue #5869, opened 2025-07-11
Remove version dependency for typing_extensions


semohr:
Currently, we only install the `typing_extensions` package for Python versions `< 3.11`. This has caused some issues, for example #5695. Additionally, we rely on conditional version checks like:

```python
if sys.version_info < (3, 11):
    from typing_extensions import ...
else:
    from typing import ...
```

We should eliminate these version checks and consistently use `typing_extensions` regardless of the python version. This simplifies the codebase and avoids subtle bugs caused by discrepancies between `typing` and `typing_extensions`.

Right now, we're effectively limited to using typing features that are natively available in Python <= 3.11.  This becomes a problem when we want to adopt newer features that are safely backported via `typing_extensions`, For instance default values for `TypeVar` or typeguards using `TypeIs`, both of which are introduced in Python 3.13.

Additionally, `typing_extensions` maintains a complete list of backported features, so switching to the stdlib when dropping support for older Python versions would be straightforward.

wisp3rwind:
Part of why we have this is https://github.com/beetbox/beets/pull/5331: Apparently someone maintaining this downstream prefers not to ship `typing_extensions`.

I don't have a strong opinion here, but I'd lean towards simplifying this on our end (I don't think we test that we're actually consistently using such version checks, i.e. we're likely to regress in this regard) and have people patch it out if it's really an issue for them.

semohr:
We already started using 3.13 typing features. This has not caused issues yet but in theory would need a typing_extension import even for python>3.11. No idea why this wasn't an issue yet.

E.g. default values for typevars are introduced in 3.13 and backported with typing_extension. In our code [here](https://github.com/beetbox/beets/blob/24dd40eed213bde14341e8c994466b29bf71de0e/beets/dbcore/db.py#L51C1-L52C52), typing ref [here](https://docs.python.org/3/library/typing.html#typing.TypeVar.__default__).

PR #5879, opened 2025-07-16
Remove version dependency for typing_extensions

semohr:
Removes  the python version dependency for typing_extensions, for more information see #5869

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -26,8 +26,9 @@ from abc import ABC
 from collections import defaultdict
 from collections.abc import Generator, Iterable, Iterator, Mapping, Sequence
 from sqlite3 import Connection
-from typing import TYPE_CHECKING, Any, AnyStr, Callable, Generic, TypeVar
+from typing import TYPE_CHECKING, Any, AnyStr, Callable, Generic
 
+from typing_extensions import TypeVar  # default value support
 from unidecode import unidecode
 
 import beets
@@ -50,9 +51,6 @@ if TYPE_CHECKING:
     from .query import SQLiteType
 
     D = TypeVar("D", bound="Database", default=Any)
-else:
-    D = TypeVar("D", bound="Database")
-
 
 FlexAttrs = dict[str, str]
 
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -50,7 +50,7 @@ if TYPE_CHECKING:
 
     from .query import SQLiteType
 
-    D = TypeVar("D", bound="Database", default=Any)
+D = TypeVar("D", bound="Database", default=Any)
 
 FlexAttrs = dict[str, str]
 
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -10,19 +10,15 @@ from __future__ import annotations
 import abc
 import inspect
 import re
-import sys
 import warnings
 from typing import TYPE_CHECKING, Generic, Literal, Sequence, TypedDict, TypeVar
+
+from typing_extensions import NotRequired
 
 from beets.util import cached_classproperty
 from beets.util.id_extractors import extract_release_id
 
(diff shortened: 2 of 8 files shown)


beets/dbcore/db.py
```diff
@@ -26,8 +26,9 @@
 from collections import defaultdict
 from collections.abc import Generator, Iterable, Iterator, Mapping, Sequence
 from sqlite3 import Connection
-from typing import TYPE_CHECKING, Any, AnyStr, Callable, Generic, TypeVar
+from typing import TYPE_CHECKING, Any, AnyStr, Callable, Generic
 
+from typing_extensions import TypeVar  # default value support
```
snejus, line 31: Hmmm, is this change needed? As far as I remember, it was about `mypy` supporting this from a certain __`mypy`__ version

semohr: I think it is needed. We use `TypeVar(default=something)` some lines below which is only supported since version 3.13 in the std lib.

If the typevar line is needed is another question. I just assumed it is here for a good reason :)

snejus: How come has it been working fine until now?

semohr: I was wondering the same some time ago. I'm still not 100% sure tho. Mypy seems to be less strict with future features and also in TYPE_CHECKING blocks.

snejus: I'm not convinced that this is required - let's continue importing it from `typing`. We can adjust it once we see any issues.

semohr: https://docs.python.org/3/library/typing.html#typing.TypeVar see after has_default.

I would keep it as otherwise we need a if statement and this is way easier imo.

snejus: OK, I think I finally see what you mean - this allows to instantiate 

```py
D = TypeVar("D", bound="Database", default=Any)
```

outside the `TYPE_CHECKING` condition.

Does this bring any value though? I quite like that the way you defined all types in `plugins.py` here:

```py
if TYPE_CHECKING:
    from collections.abc import Iterable

    from confuse import ConfigView

    from beets.dbcore import Query
    from beets.dbcore.db import FieldQueryType
    from beets.dbcore.types import Type
    from beets.importer import ImportSession, ImportTask
    from beets.library import Album, Item, Library
    from beets.ui import Subcommand

    # TYPE_CHECKING guard is needed for any derived type
    # which uses an import from `beets.library` and `beets.imported`
    ImportStageFunc = Callable[[ImportSession, ImportTask], None]
    T = TypeVar("T", Album, Item, str)
    TFunc = Callable[[T], str]
    TFuncMap = dict[str, TFunc[T]]

    AnyModel = TypeVar("AnyModel", Album, Item)

    P = ParamSpec("P")
    Ret = TypeVar("Ret", bound=Any)
    Listener = Callable[..., None]
    IterF = Callable[P, Iterable[Ret]]
```

semohr: It simplifies (i.e. removes the duplicated definition), other than that not really much value added. Sadly the `D` TypeVar cant be in the TYPE_CHECKING block as it is used as an value below (`class Model(ABC, Generic[D]):`).

snejus: What about wrapping it with double quotes?

semohr: That seems to work for some reason. The more I use typehints the less I feel like understanding them, ah well.

snejus: Double quotes can go a long way with Python typehints, especially when it comes to forward references 😅 happy it works here, it means we can keep importing from `typing` - less lines to migrate later.

semohr: Yeah this actually did not work :/ Generics are quirky. 

I reverted this to the `typing_extensions` import. Still less lines than before imo.

wisp3rwind:
LGTM. Maybe the changelog entry should go to the "For packagers" section?

PR #5882, opened 2025-07-17
Avoid Unidecoding Query Strings in Deezer Metadata Plugin

### LINKED ISSUE #5860: Avoid Unidecoding Query Strings in Deezer Metadata Plugin
**Description:**
Currently, when searching for tracks using the Deezer metadata source, the query string is passed through `unidecode`. This can lead to incorrect or unexpected results, particularly when users search for non-Latin script content (e.g., Japanese track titles).

We’ve encountered similar issues before (see refs #5699 and #5705), and a fix was already implemented in the Spotify plugin to preserve the original query string.

**Suggested Solution:**
Adapt the approach used in the Spotify plugin to avoid unidecoding the query in the Deezer plugin. This should improve search accuracy and better respect user input for international content.

semohr:
Moved the `_construct_search_query` method that was duplicated in the spotify and deezer plugins into the (new) shared helper baseclass ( `SearchApiMetadataSourcePlugin`).

By default deezer now also does not ascii encode a query when it is send but the legacy behaviour can be restored using the `deezer.search_query_ascii` option.

closes #5860  

- [x] Changelog
- [x] Documentation

### DIFF
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -14,6 +14,7 @@ import warnings
 from typing import TYPE_CHECKING, Generic, Literal, Sequence, TypedDict, TypeVar
 
 from typing_extensions import NotRequired
+import unidecode
 
 from beets.util import cached_classproperty
 from beets.util.id_extractors import extract_release_id
@@ -334,6 +335,14 @@ class SearchApiMetadataSourcePlugin(
     of identifiers for the requested type (album or track).
     """
 
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self.config.add(
+            {
+                "search_query_ascii": False,
+            }
+        )
+
     @abc.abstractmethod
     def _search_api(
         self,
@@ -381,6 +390,34 @@ class SearchApiMetadataSourcePlugin(
             None,
             self.tracks_for_ids([result["id"] for result in results if result]),
         )
+
+    def _construct_search_query(
+        self, filters: SearchFilter, keywords: str = ""
+    ) -> str:
+        """Construct a query string with the specified filters and keywords to
+        be provided to the Spotify (or similar) Search API.
+
+        At the moment, this is used to construct a query string for:
+        - Spotify (https://developer.spotify.com/documentation/web-api/reference/search).
+        - Deezer (https://developers.deezer.com/api/search).
+
+        :param filters: Field filters to apply.
+        :param keywords: Query keywords to use.
+        :return: Query string to be provided to the Search API.
+        """
+
+        query_components = [
+            keywords,
+            " ".join(f'{k}:"{v}"' for k, v in filters.items()),
+        ]
+        query = " ".join([q for q in query_components if q])
+        if not isinstance(query, str):
+            query = query.decode("utf8")
+
+        if self.config["search_query_ascii"].get():
+            query = unidecode.unidecode(query)
+
+        return query
 
 
 # Dynamically copy methods to BeetsPlugin for legacy support
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -13,8 +13,8 @@ import re
 import warnings
 from typing import TYPE_CHECKING, Generic, Literal, Sequence, TypedDict, TypeVar
 
-from typing_extensions import NotRequired
 import unidecode
+from typing_extensions import NotRequired
 
 from beets.util import cached_classproperty
 from beets.util.id_extractors import extract_release_id
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -397,15 +397,16 @@ class SearchApiMetadataSourcePlugin(
         self, filters: SearchFilter, query_string: str
     ) -> str:
         """Construct a query string with the specified filters and keywords to
-        be provided to the Spotify (or similar) Search API.
+        be provided to the spotify (or similar) search API.
 
-        At the moment, this is used to construct a query string for:
-        - Spotify (https://developer.spotify.com/documentation/web-api/reference/search).
-        - Deezer (https://developers.deezer.com/api/search).
+        The returned format was initially designed for spotify's search API but
+        we found is also useful with other APIs that support similar query structures.
+        see `spotify <https://developer.spotify.com/documentation/web-api/reference/search>`_
+        and `deezer <https://developers.deezer.com/api/search>`_.
 
         :param filters: Field filters to apply.
         :param query_string: Query keywords to use.
-        :return: Query string to be provided to the Search API.
+        :return: Query string to be provided to the search API.
         """
 
         components = [query_string, *(f'{k}:"{v}"' for k, v in filters.items())]
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -348,13 +348,13 @@ class SearchApiMetadataSourcePlugin(
         self,
         query_type: Literal["album", "track"],
         filters: SearchFilter,
-        keywords: str = "",
+        query_string: str = "",
     ) -> Sequence[R]:
         """Perform a search on the API.
 
         :param query_type: The type of query to perform.
         :param filters: A dictionary of filters to apply to the search.
-        :param keywords: Additional keywords to include in the search.
+        :param query_string: Additional query to include in the search.
 
         Should return a list of identifiers for the requested type (album or track).
         """
@@ -382,7 +382,9 @@ class SearchApiMetadataSourcePlugin(
     def item_candidates(
         self, item: Item, artist: str, title: str
     ) -> Iterable[TrackInfo]:
-        results = self._search_api("track", {"artist": artist}, keywords=title)
+        results = self._search_api(
+            "track", {"artist": artist}, query_string=title
+        )
         if not results:
             return []
 
@@ -392,7 +394,7 @@ class SearchApiMetadataSourcePlugin(
         )
 
     def _construct_search_query(
-        self, filters: SearchFilter, keywords: str = ""
+        self, filters: SearchFilter, query_string: str
     ) -> str:
         """Construct a query string with the specified filters and keywords to
         be provided to the Spotify (or similar) Search API.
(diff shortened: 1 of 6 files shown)


beets/metadata_plugins.py
```diff
@@ -386,6 +396,34 @@ def item_candidates(
...
+        - Deezer (https://developers.deezer.com/api/search).
+
+        :param filters: (Optional) Field filters to apply.
+        :param keywords: (Optional) Query keywords to use.
+        :return: Query string to be provided to the Search API.
+        """
+
+        query_components = [
+            keywords,
+            " ".join(f"{k}:{v}" for k, v in filters.items()),
```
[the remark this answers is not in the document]
semohr: Good bot!

beets/metadata_plugins.py
```diff
@@ -386,6 +396,34 @@ def item_candidates(
...
+        query_components = [
+            keywords,
+            " ".join(f'{k}:"{v}"' for k, v in filters.items()),
+        ]
+        query = " ".join([q for q in query_components if q])
+        if not isinstance(query, str):
+            query = query.decode("utf8")
+
+        if self.config["search_query_ascii"].get():
+            query = unidecode.unidecode(query)
```
snejus, line 416: I'd say we should just bomb `unidecode`. I just checked that `deezer` returns crap if I `unidecode` a japanese album name, and it returns the correct album if I don't.

`requests` these days are more than capable of dealing with any sort of strings, so I think we can safely `unidecode` for good.

beets/metadata_plugins.py
```diff
@@ -386,6 +396,34 @@ def item_candidates(
             self.tracks_for_ids([result["id"] for result in results if result]),
         )
 
+    def _construct_search_query(
+        self, filters: SearchFilter, keywords: str = ""
```
snejus: `keywords` parameter is present in every `_construct_search_query` call in the codebase, so we can tighten this up, make it required and remove the default value.

snejus: And you can rename the parameter to `query_string` for clarification too

beets/metadata_plugins.py
```diff
@@ -386,6 +396,34 @@ def item_candidates(
...
+        :param filters: Field filters to apply.
+        :param keywords: Query keywords to use.
+        :return: Query string to be provided to the Search API.
+        """
+
+        query_components = [
+            keywords,
+            " ".join(f'{k}:"{v}"' for k, v in filters.items()),
+        ]
+        query = " ".join([q for q in query_components if q])
```
snejus: I think this can be shortened a bit

```suggestion
        components = [keywords, *(f'{k}:"{v}"' for k, v in filters.items())]
        return " ".join(filter(None, components))
```

beets/metadata_plugins.py
```diff
@@ -386,6 +396,34 @@ def item_candidates(
...
+        if not isinstance(query, str):
```
snejus: This doesn't seem to be possible

beets/metadata_plugins.py
```diff
@@ -386,6 +396,34 @@ def item_candidates(
...
+        - Deezer (https://developers.deezer.com/api/search).
+
+        :param filters: Field filters to apply.
+        :param keywords: Query keywords to use.
+        :return: Query string to be provided to the Search API.
+        """
+
+        query_components = [
+            keywords,
+            " ".join(f'{k}:"{v}"' for k, v in filters.items()),
```
snejus: What happens if there is a double quote character in the query?

beets/metadata_plugins.py
```diff
@@ -386,6 +396,34 @@ def item_candidates(
...
+        be provided to the Spotify (or similar) Search API.
+
+        At the moment, this is used to construct a query string for:
+        - Spotify (https://developer.spotify.com/documentation/web-api/reference/search).
+        - Deezer (https://developers.deezer.com/api/search).
+
+        :param filters: Field filters to apply.
+        :param keywords: Query keywords to use.
+        :return: Query string to be provided to the Search API.
+        """
```
snejus: The interface sets up this method, but it shouldn't know the specifics about what implements it. 

Since the parameters' names are clear, I don't think we need to repeat this information in the docstring.

```suggestion
        """Construct a query with the specified filters and query string."""
```

snejus (changes requested):

beets/metadata_plugins.py
semohr: I thought so too before I looked into why this was implemented in the first place. I don't think we should just remove it. The query never was the issue. The main issue is the terminal or badly pasted queries if I remember correctly.

Please have a look at #5705

beets/metadata_plugins.py
semohr: I agree, the issue is tho that we lose the references to the docs if we just remove it, as these methods aren't present in the plugins file anymore. There is no repeated information.

beets/metadata_plugins.py
semohr: No idea. Depends on spotify or deezer, I guess. I check that both support quoted queries and they do but have not checked double quoted.

Deezer even needs quoted strings otherwise the api just returns rubbish. Before this PR spotify did not quote string whereas deezer did. Was the only difference in the functions.

beets/metadata_plugins.py
semohr: Also this function kinda is bound to the implementation specific as it only makes sense in the context of these two plugins and for the deezer and spotify api. Again kinda an ambiguity with our deeper structure if you ask me:
```
plugin = beetsplug(beets)
# vs
plugin = beets(beetsplug(beets))
```

snejus: This docstring sort of violates the [Open/Closed SOLID principle](https://en.wikipedia.org/wiki/SOLID#Open–closed_principle): if we add another data source that inherits `SearchApiMetadataSourcePlugin` and uses this method, we will need to update this docstring, which is bad practice. Practically speaking, this docstring would not get updated and end up being misleading.

This method (including the docstring) should be implemented/documented in a generic manner and closed for modification in order to prevent this issue.

You can move the URLs under respective `_search_api` implementations instead.

semohr: Yeah, I can move the comment, but I still think it fits better alongside the query function. The root issue here is that this setup already violates the dependency inversion principle, so moving the comment doesn’t really solve anything, it just reinforces the design flaw.

> This docstring sort of violates the [Open/Closed SOLID principle](https://en.wikipedia.org/wiki/SOLID#Open%E2%80%93closed_principle): if we add another data source that inherits SearchApiMetadataSourcePlugin and uses this method, we will need to update this docstring, which is bad practice. Practically speaking, this docstring would not get updated and end up being misleading.

I get the concern about the docstring violating the Open/Closed Principle, and that’s valid, but we’re already operating in a structure that makes adhering to these principles tricky. Rather than selectively applying them, I think we should either address the underlying design or be explicit about the compromises we're making.

semohr: I mainly think it is important for someone looking at the function to know why the query string looks that way. And this is only clear in the context of the spotify api documentation.

snejus: I don't think I'm following you regarding these two points:

> The root issue here is that this setup already violates the dependency inversion principle 

> And this is only clear in the context of the spotify api documentation

semohr: 1. I've been trying to point out for quite some time that there's a problem with how plugins and core logic are not properly separated (see, for example, https://github.com/beetbox/beets/pull/5882#discussion_r2217014242 or the get_artists discussion). Our ongoing debates stem from this fundamental issue. Since you value design principles, this essentially comes down to a violation of the dependency inversion principle.

2. Consider encountering the `create_query` function somewhere without any context. A comment explaining why the query is constructed that way, such as linking to spotifys or deezers specifications, would be very helpful. Having to dive into the spotify or deezer plugin just to understand the rationale makes maintenance harder, even if adding such comments slightly bends some design principles.

beets/metadata_plugins.py
snejus: This makes sense, thanks!

beets/metadata_plugins.py
snejus: > 1. I've been trying to point out for quite some time that there's a problem with how plugins and core logic are not properly separated (see, for example,

I'm completely lost - I'm struggling to see what this means

>> plugin = beetsplug(beets)
>> \# vs
>> plugin = beets(beetsplug(beets))

nor how any of this relates to the question that this thread was started for: _should `_construct_search_query` method docstring in the base class provide links to Spotify and Deezer APIs?

> A comment explaining why the query is constructed that way

I completely agree with this!



> Having to dive into the spotify or deezer plugin just to understand the rationale makes maintenance harder

Agree, the docstring should contain all relevant information - this is exactly why I'm not expecting to see any external URLs there.

> such as linking to spotifys or deezers specifications, would be very helpful.

Following your idea above, we should not expect devs to navigate these URLs in order to understand what this method does.

semohr: >I'm completely lost - I'm struggling to see what this means nor how any of this relates to the question that this thread was started for: _should _construct_search_query method docstring in the base class provide links to Spotify and Deezer APIs?

What I'm trying to say is that we’re only running into this issue, and this whole discussion, because our core and plugin layer aren't properly separated. More generally, we have an inversion problem, which I think is directly related here.

> Following your idea above, we should not expect devs to navigate these URLs in order to understand what this method does. 

That would be ideal, but it’s just not realistic. We're interfacing with external APIs, and the construction of the query string is entirely dependent on the specifics of the Spotify and Deezer APIs. There's no way to fully explain or document this method without referencing those APIs in some form.

semohr: I updated the docstring slightly, are we good with this version? I think it’s more reasonable now since it provides some context:

```python
    def _construct_search_query(
        self, filters: SearchFilter, query_string: str
    ) -> str:
        """Construct a query string with the specified filters and keywords to
        be provided to the spotify (or similar) search API.

        The returned format was initially designed for spotify's search API but
        we found is also useful with other APIs that support similar query structures.
        see `spotify <https://developer.spotify.com/documentation/web-api/reference/search>`_
        and `deezer <https://developers.deezer.com/api/search>`_.

        :param filters: Field filters to apply.
        :param query_string: Query keywords to use.
        :return: Query string to be provided to the search API.
        """
```

snejus (approved):

PR #5887, opened 2025-07-19
Simplify plugin loading

snejus:
This PR centralises plugin loading logic inside `beets.plugins` module.

- Removed intermediatery `_classes` variable by initialising plugins immediately.
- Simplified listeners registration by defining listener variables in the base
`BeetsPlugin` class, and making the `register_listener` method a `@classmethod`.
- Simplified plugin test setup accordingly.


beets/plugins.py
```diff
@@ -275,69 +270,97 @@ def helper(func: TFunc[Item]) -> TFunc[Item]:
...
-_instances: dict[type[BeetsPlugin], BeetsPlugin] = {}
 
+_instances: list[BeetsPlugin] = []
 
-def find_plugins() -> list[BeetsPlugin]:
-    """Returns a list of BeetsPlugin subclass instances from all
-    currently loaded beets plugins. Loads the default plugin set
-    first.
+
+def load(*args, **kwargs) -> None:
```
semohr: Could we remove the args and kwargs in favor for the actual parameters here?

beets/plugins.py
```diff
@@ -275,69 +270,97 @@ def helper(func: TFunc[Item]) -> TFunc[Item]:
...
 
-    load_plugins()
-    plugins = []
-    for cls in _classes:
-        # Only instantiate each plugin class once.
-        if cls not in _instances:
-            _instances[cls] = cls()
-        plugins.append(_instances[cls])
-    return plugins
+def find_plugins() -> Iterable[BeetsPlugin]:
```
semohr, line 393: Find plugins does not call load anymore. This might introduce issues downstream as before the behavior was different.

semohr (commented):
Looks good overall. I’ve left a few minor comments that may need addressing.

On a broader note, it might be worth considering moving the listener logic out of the BeetsPlugin ABC in the long term. Extracting it into a helper mixin, for example, could help clean up the class and enable proper event typing. It would also benefit plugin maintainers, since this logic isn’t essential for them to understand and currently adds unnecessary noise to the base class.

beets/plugins.py
snejus: I went with `*args/**kwargs` here to avoid parameter drilling, which helps keeping the function decoupled from the plugin configuration details, and saves us from having to update its signature if the discovery logic changes.

I also find that this pattern makes it easier to trace where arguments are ultimately consumed. This is a very trivial case as we only have one intermediate function, so it does not make much difference - see `beets.dbcore.queryparse` which has larger parameter drilling issue - that's the ideal refactoring target for the pattern I described.

I'm aware that this is a fairly unpopular opinion, so I'm happy to adjust if you think otherwise!

beets/plugins.py
snejus: Under typical `beets` operation (not testing) `load` is called during setup. Thus, by the time `find_plugins` is called, the plugins appear to be guaranteed to have been loaded. It seemed to me that attempting to load plugins again within `find_plugins` was redundant, so I removed that step in the hope of making this loading mechanism more deterministic. 

Under this assumption, I believe the behaviour should remain the same - what do you think?

beets/plugins.py
semohr: My main concern is that `load` might be used by external packages, possibly in contexts where beets isn’t part of the development environment or CI tooling. Without explicit parameters or type hints, the function can be less approachable for users unfamiliar with the internals.

Also, features like code completion won’t work well with `*args/**kwargs` in tools like IntelliSense or other language servers, since they depend on explicit parameter definitions to provide useful suggestions. This can make integration more frustrating than necessary.

I'd lean toward defining explicit parameters here purely for improved developer experience.

beets/plugins.py
semohr: It’s only called during setup if you’re using the CI/UI entry point, right? I don’t think plugins are loaded by default on beets import.

I was mainly thinking about our `beets-flask` package here, which doesn’t use the UI entry point at all. We just call `find_plugins`, which used to load all plugins. That’s no longer the case, of course.

Not a big issue at all since the fix is straightforward, but it might be worth adding a small note to the changelog. Could be that someone else is doing similar shenanigans.

beets/plugins.py
snejus: See the latest commit: I removed these arguments altogether by making the configuration the only source of truth.

beets/plugins.py
snejus: Makes sense - I'm adding a note in the changelog!

semohr: Thanks!

beets/plugins.py
```diff
@@ -105,6 +104,14 @@ class BeetsPlugin(metaclass=abc.ABCMeta):
     the abstract methods defined here.
     """
 
+    _raw_listeners: ClassVar[dict[EventType, list[Listener]]] = defaultdict(
```
semohr, line 148: Since you are at it anyways, could we make the raw listeners to a set? :upside_down_face: 

E.g. instead of:
```
ClassVar[dict[EventType, list[Listener]]]
```
```
ClassVar[dict[EventType, set[Listener]]]
```

snejus: I considered this when I looked to remove this condition

```py
        if func not in self._raw_listeners[event]:
```

But decided against it since ordering may have been an intentional design choice here.

semohr: Good point! I always forget python does not use ordered sets. We could use some sort of OrderedDict instead but I wouldn't bother.

snejus: We've got a small util which does a similar thing:

```py
def unique_list(elements: Iterable[T]) -> list[T]:
    """Return a list with unique elements in the original order."""
    return list(dict.fromkeys(elements))
```

But of course it requires to be called.

snejus:
@semohr is there anything else that is missing?

beets/plugins.py
semohr: You could also do something like:

```python
class OrderedSet:
    def __init__(self):
        self.ordered_dict = OrderedDict()

    def add(self, element):
        if element not in self.ordered_dict:
            self.ordered_dict[element] = None

    def __iter__(self):
        return iter(self.ordered_dict.keys())

    def __len__(self):
        return len(self.ordered_dict)

```

semohr:
Looks good to me. 

You changed the tests quite a bit so I quickly checked the coverage: Coverage decreased in the plugins file, we might have missed something. Hard to say tho since the `beets/plugins.py` code also changed but might be worth a bit more investigation.
```
Name                                     Stmts   Miss Branch BrPart   Cover   Missing
-------------------------------------------------------------------------------------
# old
beets/plugins.py                           225     23     82     11  86.32%   94-95, 121->123, 123->125, 125->128, 248-254, 264-270, 290, 332->334, 397-399, 413, 448-449, 533, 539, 541->544
# new
beets/plugins.py                           219     24     66     12  84.56%   93-94, 128->130, 130->132, 132->135, 243-249, 259-265, 296, 317, 321->338, 424-426, 440, 475-476, 547, 553, 555->558
```
The missing branch increase is a bit fishy.

snejus:
> Looks good to me.
> 
> You changed the tests quite a bit so I quickly checked the coverage: Coverage decreased in the plugins file, we might have missed something. Hard to say tho since the `beets/plugins.py` code also changed but might be worth a bit more investigation.
> 
> ```
> Name                                     Stmts   Miss Branch BrPart   Cover   Missing
> -------------------------------------------------------------------------------------
> # old
> beets/plugins.py                           225     23     82     11  86.32%   94-95, 121->123, 123->125, 125->128, 248-254, 264-270, 290, 332->334, 397-399, 413, 448-449, 533, 539, 541->544
> # new
> beets/plugins.py                           219     24     66     12  84.56%   93-94, 128->130, 130->132, 132->135, 243-249, 259-265, 296, 317, 321->338, 424-426, 440, 475-476, 547, 553, 555->558
> ```
> 
> The missing branch increase is a bit fishy.

I think this is due to the move of the logic from `ui/__init__.py` to `plugins.py`. Previously in `ui/__init__.py`:
image

Now in `plugins.py`:

image

semohr:
> I think this is due to the move of the logic from ui/__init__.py to plugins.py. Previously in ui/__init__.py

Makes sense.

beets/plugins.py
```diff
@@ -270,69 +265,99 @@ def helper(func: TFunc[Item]) -> TFunc[Item]:
...
+    to notify other components.
     """
-    if _instances:
-        # After the first call, use cached instances for performance reasons.
-        # See https://github.com/beetbox/beets/pull/3810
-        return list(_instances.values())
+    if not _instances:
+        names = get_plugin_names()
+        log.info("Loading plugins: {}", ", ".join(sorted(names)))
+        _instances.extend(filter(None, map(_get_plugin, names)))
```
semohr, line 388: Yet another short question before I approve this:

We now omit the `and obj not in _classes` check. If a user now wants to load a plugin two times, what will happen? Is that an issue, will the plugin be loaded twice? 
I think there had to be a reason for this check in the previous version.

snejus: `load_plugins` will not attempt to reload the plugins if `_instances` is not empty and `get_plugin_names` is guaranteed to return a unique list.

If someone needs to reload plugins, they need to clear the list of `_instances`.

semohr: I mean `get_plugin_names` can return the same name multiple times if the plugin is present in the config multiple times. This will load the same plugin multiple times if I am not missing something obvious here.

Previously these duplicates were skipped with the _classes list.

snejus: You're right, have just made an adjustment to prevent it:

```py
    plugins = unique_list(beets.config["plugins"].as_str_seq())
```

semohr: Awesome! Guess that works. Maybe warning the user would also be appropriate?

snejus:
@semohr Note I made a couple of adjustments in the plugin loading test

beets/plugins.py
snejus: I think it's fine to _just_ handle it as there's no damage done if they specify duplicates. This also handles duplicates in the command line, e.g. `beet --plugins=discogs,deezer,discogs import ...`

semohr (approved):

PR #5888, opened 2025-07-20
Add mbpseudo plugin for pseudo-release proposals

asardaes:
## Description

Hi there, I wanted to implement this for my foreign music and I would be glad to get some feedback from the maintainers. I will address all the TODOs, but first I'd like to know if this fits.

The auto-tagger can already resolve official releases from pseudo-releases, but this only happens if the user provides search IDs manually. Similarly, `import.languages` can be used to search for artist aliases (and that does happen automatically), but that doesn't apply to album and track names.

This plugin proactively searches for pseudo-releases during import and adds them as candidates. Since it also depends on MusicBrainz, there are some special considerations for the default logic (which is now a plugin as well). However, at the very least it expects a list of desired [names of scripts](https://en.wikipedia.org/wiki/ISO_15924) in the configuration, for example:

```yaml
mbpseudo:
    scripts:
    - Latn
```

It will use that to search for pseudo-releases that match some of the desired scripts, but will only do so if the input tracks match against an official release that is not in one of the desired scripts.

## Standalone Usage

This would be the recommended approach, which involves disabling the `musicbrainz` plugin. The `mbpseudo` plugin will manually delegate the initial search to it. Since the data source of official releases will still match MusicBrainz, weights are still relevant:

```yaml
mbpseudo:
    source_weight: 0.0
    scripts:
    - Latn

musicbrainz:
    source_weight: 0.1
```

A setup like that would ensure that the pseudo-releases have slightly more preference when choosing the final proposal.

## Combined Usage

I initially thought it would be important to coexist with the `musicbrainz` plugin when it's enabled, and reuse as much of its data as possible to avoid redundant calls to the MusicBrainz API. I have the impression this is not really important in the end, and maybe things could be simplified if we decide that both plugins shouldn't coexist.

As it is right now, using both plugins at the same time would still work, but it'll only avoid redundancy if `musicbrainz` emits its candidates before `mbpseudo`, ~which is why I modified the plugin-loading logic slightly to guarantee ordering. I'm not sure if you think this could be an issue, but I think the `musicbrainz` plugin is also used by other plugins and I can imagine it's good to guarantee the order that is declared in the configuration?~

If the above is fulfilled, the `mbpseudo` plugin will use listeners to intercept data emitted by the `musicbrainz` plugin and check if any of them have pseudo-releases that might be desirable.

## To Do

- [x] Documentation.
- [x] Changelog.
- [x] Tests.

### DIFF
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -24,7 +24,7 @@ from typing import TYPE_CHECKING, Any, NamedTuple, TypeVar
 import lap
 import numpy as np
 
-from beets import config, logging, metadata_plugins
+from beets import config, logging, metadata_plugins, plugins
 from beets.autotag import AlbumInfo, AlbumMatch, TrackInfo, TrackMatch, hooks
 from beets.util import get_most_common_tags
 
@@ -274,12 +274,17 @@ def tag_album(
             log.debug("Searching for album ID: {}", search_id)
             if info := metadata_plugins.album_for_id(search_id):
                 _add_candidate(items, candidates, info)
+                if opt_candidate := candidates.get(info.album_id):
+                    plugins.send("album_matched", match=opt_candidate)
 
     # Use existing metadata or text search.
     else:
         # Try search based on current ID.
         if info := match_by_id(items):
             _add_candidate(items, candidates, info)
+            for candidate in candidates.values():
+                plugins.send("album_matched", match=candidate)
+
             rec = _recommendation(list(candidates.values()))
             log.debug("Album ID match recommendation is {}", rec)
             if candidates and not config["import"]["timid"]:
@@ -313,6 +318,8 @@ def tag_album(
             items, search_artist, search_album, va_likely
         ):
             _add_candidate(items, candidates, matched_candidate)
+            if opt_candidate := candidates.get(matched_candidate.album_id):
+                plugins.send("album_matched", match=opt_candidate)
 
     log.debug("Evaluating {} candidates.", len(candidates))
     # Sort and get the recommendation.
--- a/docs/plugins/mbpseudo.rst
+++ b/docs/plugins/mbpseudo.rst
@@ -23,12 +23,17 @@ Since this plugin first searches for official releases from MusicBrainz, all
 options from the `musicbrainz` plugin's :ref:`musicbrainz-config` are supported,
 but they must be specified under `mbpseudo` in the configuration file.
 Additionally, the configuration expects an array of scripts that are desired for
-the pseudo-releases. Therefore, the minimum configuration for this plugin looks
-like this:
+the pseudo-releases. For ``artist`` in particular, keep in mind that even
+pseudo-releases might specify it with the original script, so you should also
+configure import :ref:`languages` to give artist aliases more priority.
+Therefore, the minimum configuration for this plugin looks like this:
 
 .. code-block:: yaml
 
     plugins: mbpseudo # remove musicbrainz
+
+    import:
+        languages: en
 
     mbpseudo:
         scripts:
@@ -37,13 +42,16 @@ like this:
 Note that the `search_limit` configuration applies to the initial search for
 official releases, and that the `data_source` in the database will be
 "MusicBrainz". Nevertheless, `data_source_mismatch_penalty` must also be
-specified under `mbpseudo` (see also
+specified under `mbpseudo` if desired (see also
 :ref:`metadata-source-plugin-configuration`). An example with multiple data
 sources may look like this:
 
 .. code-block:: yaml
 
     plugins: mbpseudo deezer
+
+    import:
+        languages: en
 
     mbpseudo:
         data_source_mismatch_penalty: 0
@@ -52,3 +60,44 @@ sources may look like this:
 
     deezer:
         data_source_mismatch_penalty: 0.2
+
+By default, the data from the pseudo-release will be used to create a proposal
+that is independent from the official release and sets all properties in its
+metadata. It's possible to change the configuration so that some information
+from the pseudo-release is instead added as custom tags, keeping the metadata
+from the official release:
+
+.. code-block:: yaml
+
+    mbpseudo:
+        # other config not shown
+        custom_tags_only: yes
+
+The default custom tags with this configuration are specified as mappings where
+the keys define the tag names and the values define the pseudo-release property
+that will be used to set the tag's value:
+
+.. code-block:: yaml
+
+    mbpseudo:
+        album_custom_tags:
+            album_transl: album
+            album_artist_transl: artist
+        track_custom_tags:
+            title_transl: title
+            artist_transl: artist
+
+Note that the information for each set of custom tags corresponds to different
+metadata levels (album or track level), which is why ``artist`` appears twice
+even though it effectively references album artist and track artist
+respectively.
+
+If you want to modify any mapping under ``album_custom_tags`` or
+``track_custom_tags``, you must specify *everything* for that set of tags in
+your configuration file because any customization replaces the whole dictionary
+of mappings for that level.
+
+.. note::
+
+    These custom tags are also added to the music files, not only to the
+    database.
--- a/docs/plugins/mbpseudo.rst
+++ b/docs/plugins/mbpseudo.rst
@@ -1,0 +1,56 @@
+MusicBrainz Pseudo-Release Plugin
+=================================
+
+The `mbpseudo` plugin can be used *instead of* the `musicbrainz` plugin to
+search for MusicBrainz pseudo-releases_ during the import process, which are
+added to the normal candidates from the MusicBrainz search.
+
+.. _pseudo-releases: https://musicbrainz.org/doc/Style/Specific_types_of_releases/Pseudo-Releases
+
+This is useful for releases whose title and track titles are written with a
+script_ that can be translated or transliterated into a different one.
+
+.. _script: https://en.wikipedia.org/wiki/ISO_15924
+
+Pseudo-releases will only be included if the initial search in MusicBrainz
+returns releases whose script is *not* desired and whose relationships include
+pseudo-releases with desired scripts.
+
+Configuration
+-------------
+
+Since this plugin first searches for official releases from MusicBrainz, most
+options from the `musicbrainz` plugin's :ref:`musicbrainz-config` are supported,
+but they must be specified under `mbpseudo` in the configuration file.
+Additionally, the configuration expects an array of scripts that are desired for
+the pseudo-releases. Therefore, the minimum configuration for this plugin looks
+like this:
+
+.. code-block:: yaml
+
+    plugins: mbpseudo # remove musicbrainz
+
+    mbpseudo:
+        scripts:
+        - Latn
+
+Note that the `search_limit` configuration applies to the initial search for
+official releases, and that the `data_source` in the database will be
+"MusicBrainz". Because of this, the only configuration that must remain under
+`musicbrainz` is `data_source_mismatch_penalty` (see also
+:ref:`metadata-source-plugin-configuration`). An example with multiple data
+sources may look like this:
+
+.. code-block:: yaml
+
+    plugins: mbpseudo deezer
+
+    mbpseudo:
+        scripts:
+        - Latn
+
+    musicbrainz:
+        data_source_mismatch_penalty: 0
+
+    deezer:
+        data_source_mismatch_penalty: 0.5
--- a/docs/plugins/mbpseudo.rst
+++ b/docs/plugins/mbpseudo.rst
@@ -19,7 +19,7 @@ pseudo-releases with desired scripts.
 Configuration
 -------------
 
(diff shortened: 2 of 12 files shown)


semohr:
I was wondering if it woudn't be easier to just extend the default `Musicbrainz` plugin instead of using the event system. Have you tried that?

asardaes:
@semohr I think that would be fine, I just didn't want to start by modifying the core plugin. I'm not sure how to go about the source weights though, would the distance calculation need to be modified to consider script somehow? Or how could the user nudge the proposals toward the pseudo releases?

semohr:
> I think that would be fine, I just didn't want to start by modifying the core plugin. 

Instead, I was thinking along the lines of extending it, while keeping the logic isolated in its own plugin/file. Something like:
```python
from beetsplug.musicbrainz import MusicbrainzPlugin

class MBPseudoPlugin(MusicbrainzPlugin):
    pass
```
That way, the original plugin remains untouched, and this extension can encapsulate the new behavior.
It might also be good to raise a warning or error if both plugins are enabled at the same time, just to avoid confusing overlaps.

> I'm not sure how to go about the source weights though, would the distance calculation need to be modified to consider script somehow? Or how could the user nudge the proposals toward the pseudo releases?

I don’t think special handling is strictly necessary here, but I’m not 100% certain. You might want to take a look at the functions in `autotag.distance`.

asardaes:
Ah I see, I imagine that should be fine. I'm away from my pc for some time, but I'll take a look at that option.

asardaes:
@semohr I changed the implementation as suggested and I think it looks good. The config I mentioned in the original post still applies. I've been validating with [this release](https://musicbrainz.org/release/a5ce1d11-2e32-45a4-b37f-c1589d46b103) and I do get the pseudo-release as the main proposal.

If this is ok, I'll add some tests and docs.

semohr (changes requested):
This looks pretty amazing and way cleaner now! Nicely done.

Some more things before we merge this. I think you wanted to do this either way but the plugin needs some documentation. Also while not required tests would be highly appreciated.

Another minor thing: We are trying to not use `Optional[...]` instead please use ` ... | None` (this might need a `from __future__ import annotations` ). Also if you want to continue maintaining this plugin we would highly appreciate if you could add you handle to the `CODEOWNERS` file. This way you get updates if any PRs try to apply changes to the plugin (totally optional).

asardaes:
@semohr I initially didn't use Optional, but then the CI fails because it says that's not supported for python 3.9 I think.

Edit: [here's the failed CI](https://github.com/beetbox/beets/actions/runs/16849430046) that said:

> X | Y syntax for unions requires Python 3.10

I'll have to change a couple things after #6077, but I'll start looking at the documentation and the tests.

semohr:
> X | Y syntax for unions requires Python 3.10

I think adding the `from __future__ import annotations` import should fix this. See more information [here](https://docs.python.org/3/library/__future__.html).

asardaes:
@semohr after the latest changes, the source weights (and the new source penalties) don't really work because all the proposals end up coming from the same plugin, but I figured it's not really necessary because we can just emit the pseudo-releases first and they'll have a higher priority after sorting by distance. However, I had to add an event in order to intercept the final proposal after the distance calculation, I hope this is ok; I documented it as well.

docs/plugins/mbpseudo.rst
```diff
@@ -0,0 +1,56 @@
...
+    plugins: mbpseudo # remove musicbrainz
+
+    mbpseudo:
+        scripts:
+        - Latn
+
+Note that the `search_limit` configuration applies to the initial search for
+official releases, and that the `data_source` in the database will be
+"MusicBrainz". Because of this, the only configuration that must remain under
+`musicbrainz` is `data_source_mismatch_penalty` (see also
```
asardaes: Oh this also won't work because the code only returns configured penalties if the corresponding source's plugin is actually loaded, and `musicbrainz` won't be loaded 🤔

beets/autotag/match.py
```diff
@@ -280,6 +280,9 @@ def tag_album(
         # Try search based on current ID.
         if info := match_by_id(items):
             _add_candidate(items, candidates, info)
+            for candidate in candidates.values():
+                plugins.send("album_matched", match=candidate)
```
semohr, line 286: Can you shortly outline why the event is needed?

asardaes: For example when the input Items are in Japanese and MusicBrainz has a translation as a pseudo-release in English, the pseudo-release's `PseudoAlbumInfo` should use the official release in Japanese for the distance calculation so that the Items are matched correctly, but after that we need to switch to the pseudo-release's info for the final proposal, otherwise the DB will not reflect the pseudo-release. The only place where this could happen right now is between the addition of candidates and the recommendation computation, so I needed an event in between to intercept the match and do the adjustment.

asardaes:
@semohr I fixed the issue with the data source penalty.

docs/plugins/mbpseudo.rst
```diff
@@ -0,0 +1,54 @@
...
+specified under `mbpseudo` (see also
+:ref:`metadata-source-plugin-configuration`). An example with multiple data
+sources may look like this:
+
+.. code-block:: yaml
+
+    plugins: mbpseudo deezer
+
+    mbpseudo:
+        data_source_mismatch_penalty: 0
```
semohr, line 57: I would remove the `data_source_mismatch_penalty` here. It is not really that helpful in the first place as it is only applied on reimport/update and the default value should be fine for most people. Might be a bit more confusing than helpful here.

If you really want to include a remark you could either put it into a [dropdown](https://sphinx-design.readthedocs.io/en/pydata-theme/dropdowns.html) or an admonition.

semohr (changes requested):
Other than the one request this looks pretty good to me now! 

I will schedule this for 2.6.0 👍

docs/plugins/mbpseudo.rst
asardaes: In the final version implemented in #6077 the logic does apply during new imports, I've validated that locally precisely with `mbpseudo` and `deezer` active and with the values I documented here.

asardaes: Looks like this:

```
Finding tags for album "Lilas Ikuta - 百花繚乱".
  Candidates:
  1. (97.6%) Lilas Ikuta - In Bloom
             ≠ media
             MusicBrainz, Digital Media, 2025, XW, None, Lilas-020, None
  2. (97.6%) Lilas Ikuta - 百花繚乱
             ≠ media
             MusicBrainz, Digital Media, 2025, XW, None, Lilas-020, None
  3. (96.1%) Lilas Ikuta - 百花繚乱
             ≠ data source, tracks
             Deezer, None, 2025, None, 幾田りら, None, Non
```

semohr: I'm a bit confused now. The following will only ever add a distance if `before != None` which is never the case for new imports. https://github.com/beetbox/beets/blob/af022683fe4141cfc9f7e26b0d17cb17fa6d15e4/beets/autotag/distance.py#L348

asardaes: The first comparison is `before != after`, so e.g. `None != "MusicBrainz"`, which is True. Then `or len(metadata_plugins.find_metadata_source_plugins()) > 1` is True as well.

semohr: Sorry, missed that you are right :+1:

asardaes:
I had to rebase due to changelog updates for 2.5.1

miblodelcarpio:
Cheers for all your work on this. I've been wondering, would this plugin let us take the metadata from both the original and the translated pseudo-release, and apply both sets to the same file? For example, using the original metadata for the file's tags, and the pseudo-release metadata for its directory / filename.

My reading of [the docs](https://github.com/beetbox/beets/blob/7c22db9389d9d104154815be1b04acdc6e1b6a6c/docs/plugins/mbpseudo.rst) suggests that it wouldn't let us do this, but I figured I'd better check.

asardaes:
@miblodelcarpio I'm personally not sure, the plugin basically emits `AlbumInfo`s, and if I had to guess, I'd say the rest of the import flow uses the same properties from those objects to determine tags and file names, so I imagine it's currently not possible to do that with any plugin, but I'd have to go through the code to make sure.

I'm also not sure that would be ideal, for example [navidrome](https://github.com/navidrome/navidrome) completely ignores file names and only cares about tags.

Maybe it would be possible to add an option so that the pseudo-release's info is added as custom additional tags instead of replacing the official release, but that wouldn't affect the directory/file names. I'd also have to check if the plugin machinery currently allows such custom tags.

miblodelcarpio:
Ah cheers, @asardaes. Yeah, I appreciate it's a pretty weird and specific use-case! I actually already do this file renaming manually, to let me keep my files tagged canonically in their original language / script, while also giving me a way to meaningfully recognise the titles of tracks and albums, and artists whose native language I sadly don't understand.

Supposing the full thing isn't (yet) possible, I do like your idea to add the pseudo-release's info as custom tags, if that is possible. Pretty sure I could easily get those tags out using `mediainfo`.

asardaes:
@miblodelcarpio I looked at the code and, if I understand correctly, custom tags will only be added to the music file if they are registered in [mediafile](https://github.com/beetbox/mediafile), otherwise they only end up in the beets DB. I do see some [MusicBrainz-specific tags](https://github.com/beetbox/mediafile/blob/83157ca0f6643bab50e46f8d5378d062ae85fb39/mediafile.py#L2119), but they are for IDs, and I don't see anything that might fit a translation/transliteration in a generic way. I think we would have to track such a change separately, but I think it could be nice to have tags for translations that any plugin could modify, not sure what @semohr thinks about that.

semohr:
I'm not following the discussion closely, but I guess you are looking for the [`add_media_field`](https://github.com/beetbox/beets/blob/c26c342cc155945db83f178c948f870d10175010/beets/plugins.py#L339C9-L339C24) method. It should allow you to add specific tag fields that are also written to the file.

miblodelcarpio:
Lovely, thank you for looking into this, @asardaes, and @semohr for suggesting [`add_media_field`](https://github.com/beetbox/beets/blob/c26c342cc155945db83f178c948f870d10175010/beets/plugins.py#L339C9-L339C24). Do you want to gauge whether or not this custom tag idea is in scope for the PR? Naturally I'd welcome it, but equally I'd be happy to see the plugin be merged before going into the broader ideas of "applying metadata from multiple sources to the same file" and "tags for translations that any plugin could modify". There is also issue #2866 from 2018 which I've just discovered and reckon gets to the nub of it, with @amogus07 even [currently working](https://github.com/beetbox/beets/issues/2866#issuecomment-3396303682) on a plugin with the same requirement!

amogus07:
Something like in my custom base class could be implemented in MetadataSourcePlugin, so plugins can like mine easily define fields in their namespace which is inferred from `self.name`.

amogus07:
Honestly, I don't care about having multiple artists, titles, etc. I just think it's useful for id fields. Everything else can be inferred from the ids by reimporting. It would also fix multiple plugins trying to override the same album or item when running `beet mbsync`, since they would all use different id fields.

asardaes:
@miblodelcarpio I added the option to only add custom tags. Since beets already supports [custom path configurations](https://beets.readthedocs.io/en/stable/reference/config.html#path-format-configuration), it seems this would accomplish what you wanted:

```yaml
paths:
    default: %ifdef{album_artist_transl, $album_artist_transl, $albumartist}/%ifdef{album_transl, $album_transl, $album}%aunique{}/$track%ifdef{title_transl, $title_transl, $title}
```

miblodelcarpio:
@asardaes Oh wow, thank you so much! Really looking forward to giving this a spin. Coupled with the conditional logic happening in `item_fields` from the [`inline`](https://beets.readthedocs.io/en/stable/plugins/inline.html) plugin, this'll be a dream. I'm thinking:

```yaml
item_fields:
    artist: %ifdef{artist_transl, $artist_transl, $artist}
    album: %ifdef{album_transl, $album_transl, $album}
    title: %ifdef{title_transl, $title_transl, $title}
```

While `$albumartist_sort` already comes in the language I'm after, English, with `$artist` being used for `comp` releases.

Update (2026-03-25): It does exactly what I want, with the following config changes as per [the docs](https://beets.readthedocs.io/en/stable/plugins/mbpseudo.html):

```yaml
plugins:
    mbpseudo # Replacing musicbrainz

mbpseudo:
    data_source_mismatch_penalty: 0
    scripts:
    - Latn
    custom_tags_only: yes

items_fields:
    artist_en: artist_transl if artist_transl else artist
    album_en: album_transl if album_transl else album
    title_en: title_transl if title_transl else title

paths:
    # Appending _en to all instances of artist, album and title
```

asardaes:
@semohr I don't have further changes if you agree with the latest additions.

semohr:
In my head this is already approved! I'm waiting for some free time to have one closer look and merge this.

I haven't forgotten you, please be patient for a bit longer 🙏

semohr (approved):

PR #5892, opened 2025-07-24
Deprecate imports, decargs

snejus:
## Summary

See https://github.com/beetbox/beets/discussions/5891 for context.

This PR introduces a deprecation system for module structure changes and deprecated functionality in the beets codebase:

### Key Changes

**New deprecation infrastructure:**
- Add `deprecate_imports()` utility function to handle deprecated module imports with warnings
- Implement `__getattr__` hooks in modules to intercept and redirect deprecated imports
- Deprecate previously removed imports in `beets.library` and `beets.autotag`
- Deprecate `decargs` function.

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -28,6 +28,7 @@ import struct
 import sys
 import textwrap
 import traceback
+import warnings
 from difflib import SequenceMatcher
 from typing import TYPE_CHECKING, Any, Callable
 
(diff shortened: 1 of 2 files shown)


JOJ0:
Yes this is a great solution @snejus , but somehow I don't see the deprecation warnings with this branch checked out. Must be something stupid on my end. 

Eg. I issue `beet describe something` and the plugin _just works_. I don't get import errors anymore but also I don't get a deprecation warning.

wisp3rwind:
> Yes this is a great solution @snejus , but somehow I don't see the deprecation warnings with this branch checked out. Must be something stupid on my end.
> 
> Eg. I issue `beet describe something` and the plugin _just works_. I don't get import errors anymore but also I don't get a deprecation warning.

Could it be that the code in `deprecate_imports` raises an exception, but that is caught somewhere else unintentionally, e.g. during plugin load?

snejus:
As a user, you are not expected to see deprecation warnings - these are primarily intended for other developers. If you want to see them, you need to [pass `-Wd` to the python invocation](https://docs.python.org/3.9/library/warnings.html#updating-code-for-new-versions-of-dependencies).

These warnings are enabled by default by `unittest` and `pytest`, which is how we expect external developers to encounter them (assuming their test setup does not disable warnings 😁).

JOJ0:
> As a user, you are not expected to see deprecation warnings - these are primarily intended for other developers. If you want to see them, you need to [pass `-Wd` to the python invocation](https://docs.python.org/3.9/library/warnings.html#updating-code-for-new-versions-of-dependencies).
> 
> 
> 
> These warnings are enabled by default by `unittest` and `pytest`, which is how we expect external developers to encounter them (assuming their test setup does not disable warnings 😁).

Thanks for clarifying, that sounds even more useful and most of all enduserfriendly! Nice one!

PR #5893, opened 2025-07-25
discogs: don't crash if a release is deleted

9999years:
## Description

Detect an `HTTPError` when hitting the Discogs API to prevent a crash when a release has been deleted.

## To Do


snejus:
@9999years please fix the issues reported by CI

beetsplug/discogs.py
```diff
@@ -47,6 +48,9 @@
 
     from beets.library import Item
 
+# Global logger.
+log = logging.getLogger("beets")
```
snejus: It doesn't seem like it's being used?

beetsplug/discogs.py
```diff
@@ -56,7 +60,9 @@
     ConnectionError,
     socket.error,
     http.client.HTTPException,
+    HTTPError,
     ValueError,  # JSON decoding raises a ValueError.
+    JSONDecodeError,  # Except sometimes it raises a `JSONDecodeError`
```
snejus: Unneeded - `JSONDecodeError` inherits from `ValueError` which is already present in the list.

beetsplug/discogs.py
```diff
@@ -56,7 +60,9 @@
     ConnectionError,
     socket.error,
     http.client.HTTPException,
+    HTTPError,
```
snejus: Unneeded - `HTTPError` inherits from `DiscogsAPIError` which is already present in the list.

beetsplug/discogs.py
```diff
@@ -305,7 +311,15 @@ def get_album_info(self, result):
...
         # present if the result is from a `discogs_client.search()`.
         if not result.data.get("artists"):
-            result.refresh()
+            try:
+                result.refresh()
+            except CONNECTION_ERRORS:
+                self._log.debug(
+                    "Connection error in release lookup: {0}",
+                    result,
+                    exc_info=True,
```
snejus: I doubt that the exception trace will be useful to the user - I think the `result` is enough.
```suggestion
```

snejus (changes requested):

beetsplug/discogs.py
9999years: Ah, thank you, this was meant for #5897 but slipped in here.

9999years:
Addressed comments, PTAL :)

PR #5896, opened 2025-07-27
Use python3 print in plugins.rst

dmcdougall:
## Description

Use python3 print in plugins.rst docs

## To Do


wisp3rwind (approved):
Makes sense, thanks :+1:

PR #5262, opened 2024-05-25
Add timeout argument to all `requests` calls

droserasprout:
## Description

Fixes bug where some plugin commands hang indefinitely. 8 years have passed since opening https://github.com/psf/requests/issues/3070, and we still have to deal with this stuff 😔 10 seconds value have been chosen as it's used multiple times in the codebase.

## To Do

- [x] ~Documentation. (If you've added a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)~
- [x] Changelog. (Add an entry to `docs/changelog.rst` to the bottom of one of the lists near the top of the document.)
- [x] ~Tests. (Very much encouraged but not strictly required.)~

PR #5898, opened 2025-07-28
chroma: set a default timeout of 10 seconds

9999years:
## Description

TODO: Configurable timeouts

## To Do


snejus:
Looks good to me, just need a changelog note. Is there a reason this is a draft?

9999years:
@snejus Thanks, that's good to hear! I kinda wanted to write a test for it and/or make it configurable. It's OK as a patch for me to run but I'd like it to be configurable for others...

snejus:
I think very few people would every want to adjust this timeout, probably... I think your adjustment goes in line with this previous PR: #5262, where we added `timeout` argument to every `requests` invocation. Thus, I'm happy with what you have here plus a note in the changelog :)

Issue #3262, opened 2019-05-09
Use best width image as fallback in fetchart plugin


msdos:
### Use case

I have a `minwidth` of 1024 for `fetchart` plugin. `source` priority is filesystem.

My `cover.jpg` image in filesystem is 500x500. Because of that, it tries to get a better picture using the `fetchart` plugin.

Problem is: the fetched image is 300x300. In this situation, since the fetched image is `worse` than my filesystem image, I would prefer to use it instead of the downloaded one.

### Solution

I would like to have a better approach, but don't know exactly what to do:

- Create a new config, `download_all` that tries to get the image from all sources, and chooses the best one? This is a completely new development of the fetchart feature;
- Create a new config, `fallback_worse_fetched_ratio`, that uses the filesystem `cover.jpg` instead of the fetched image if the downloaded resource has a worse ratio.

### Alternatives

Today is a manual approach.

sampsyo:
Interesting idea! Seems like a reasonable thing to add.

wisp3rwind:
With respect to not respecting `minwidth`, that would be an actual bug. However, looking at the code, the only way this could happen appears to missing dependencies: `fetchart` requires either of Pillow or ImageMagick to be installed (this is rather hidden in [the docs](https://beets.readthedocs.io/en/v1.4.7/plugins/fetchart.html#configuration)). If that is indeed the issue at hand, when running beets in verbose mode, you should see a warning "Could not get size of image...". Could you confirm whether or not these are installed? If so, and if the issue persists, would you mind posting a verbose log of beets replacing such an image?

About the `download_all` feature request: The infrastructure for such an addition would actually be mostly in place because I considered implementing a related feature in #1917 (which I didn't continue to work on because after all, beets was usually finding sufficiently good artwork without manual intervention).

sampsyo:
Hmm, that’s true. Upon more careful reflection, I’m not sure what the desired behavior would be here—if not `minwidth`, how should the system decide whether a local image is high enough resolution? Reductively: should a 2x2 image be applied, if it’s the highest-resolution image available?

msdos:
> If so, and if the issue persists, would you mind posting a verbose log of beets replacing such an image?

@wisp3rwind I think I mixed concepts here since I was importing like 12 albums and refining my yaml configuration at the same time and I'm sorry about that.

[...]

PR #5902, opened 2025-07-29
Add fetchart option to set the algorithm used to fetch album art.

redlamb:
Add configuration option (`match`) to the fetchart plugin that allows user to set the algorithm used when fetching album art. Possible values are as follows:

- `first` (default): Returns the first image that meets the specs defined by the config.
- `max_resolution`: Fetch one image from each of the configured sources and return the one with the highest resolution.

## Description

Possible solution for #3262. To be nice to backend providers, `max_resolution` will only download _one_ image from each configured source before selecting the image with the highest resolution.

## To Do

- [x] Documentation.
- [x] Changelog.
- [ ] Tests.


snejus:
Apologies for a very late review, but would you mind rebasing this on a recent master?

snejus:

PR #5906, opened 2025-07-30
Add py.typed marker file to support PEP 561 typing

semohr:
Add `py.typed` marker file to support PEP 561 typing

This PR adds a `py.typed` marker file to the package directory to indicate that the package includes inline type hints and is PEP 561 compliant.

### Why this is needed

* The `py.typed` file signals to type checkers (e.g. `mypy`, `pyright`) that type information is available within the package.
* Without this file, external projects using this package will not be able to consume its type hints.
* This improves developer experience and enables better static analysis for downstream users.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -107,7 +107,7 @@ jobs:
         uses: liskin/gh-problem-matcher-wrap@v3
         with:
           linters: mypy
-          run: poe check-types --show-column-numbers --no-error-summary ${{ needs.changed-files.outputs.changed_python_files }}
+          run: poe check-types --show-column-numbers --no-error-summary .
 
   docs:
     if: needs.changed-files.outputs.any_docs_changed == 'true'
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -92,6 +92,10 @@ For plugin developers:
 
   Old imports are now deprecated and will be removed in version ``3.0.0``.
 * ``beets.ui.decargs`` is deprecated and will be removed in version ``3.0.0``.
+* Beets is now pep 561 compliant, which means that it provides type hints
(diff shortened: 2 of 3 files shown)


wisp3rwind:
Makes sense to me in general; a few questions for my understanding:

- This does not make any claim from our side that typed functions are part of an official API, right? Rather, this simply documents the fact that is safe to run a type checker against our code?
- If we do this, we should be sure that we actually always pass type checking. https://github.com/beetbox/beets/pull/5870 requires mypy to succeed in CI. However, https://github.com/beetbox/beets/blob/268075063fbf5f796ee1bc419e5268f593c85f4d/.github/workflows/lint.yml#L110 seems to only run mypy against changed files. That does not in fact guarantee that type checking all of beets continues to succeed: If mypy config is changed, or if an interface is broken that we use elsewhere in beets, CI could succeed and still the codebase might end up in a state that does not type check. Should we set up CI to unconditionally run mypy against the entire codebase?

snejus:
Nice! Should we add it under `beetsplug` too?

@wisp3rwind you're making a very good point here, I think we should indeed run `mypy` across the entire codebase.

semohr:
> This does not make any claim from our side that typed functions are part of an official API, right? Rather, this simply documents the fact that is safe to run a type checker against our code?

If we put the file into our src folder, we only claim that we care about typehints and have taken  some effort adding typehints. I think only mypy really cares about this file anyways.

> Should we set up CI to unconditionally run mypy against the entire codebase?

Agreed, I don't see why we shouldn't do this. Added it to this PR!

> Nice! Should we add it under beetsplug too?

I'm pretty sure there are some plugins which are still not typed yet. I would omit the py.typed marker for now. Also I don't think many ppl actually consume the beetsplug package/namespace as dependency.

snejus (approved):

PR #5909, opened 2025-07-31
Adds a `zero_disc_number_if_single_disc` to the zero plugin

michaelkrieger:
Adds a `zero_disc_number_if_single_disc` boolean to the zero plugin for writing to files.  Adds the logic that, if disctotal is set and there is only one disc in disctotal, that the disc is not set.

This keeps tags cleaner, only using disc on multi-disc albums. The disctotal is not touched, particularly as this is not usually displayed in most clients.

The field is removed only for writing the tags, but the disc number is maintained in the database to avoid breaking anything that may depend on a disc number or avoid possible loops or failed logic.

A column of disc 1 makes me feel there should be a disc 2, when most albums are a single disc only.

## Description

This addresses concerns such as https://beets-users.narkive.com/fxh0cgmC/beets-remove-disc-field-for-single-disc-albums and accomplishes things like https://github.com/beetbox/beets/discussions/5670 but in tags instead of just in the path.

## To Do


michaelkrieger:
Would love to hear some feedback on this or its implementation.

beetsplug/zero.py
```diff
@@ -123,8 +124,13 @@ def set_fields(self, item, tags):
         """
         fields_set = False
 
+        if 'disc' in tags and self.config["zero_disc_number_if_single_disc"].get(bool):
+            if getattr(item, 'disctotal', None) == 1:
```
snejus: I think you can access this attribute directly
```suggestion
            if item.disctotal == 1:
```

beetsplug/zero.py
```diff
@@ -123,8 +124,13 @@ def set_fields(self, item, tags):
...
+                self._log.debug("{0}: {1} -> None", 'disc', item['disc'])
```
snejus:
```suggestion
                self._log.debug("disc: {.disc} -> None", item)
```

michaelkrieger:
Incorporated all suggestions and fixed the formatting issues.  All CI now tests okay.
Also fixed merge conflicts created since the original change.
In my view, very ready for merging @snejus

michaelkrieger:
Anything else need to be changed @snejus ?

fredrik:
> Would love to hear some feedback on this or its implementation.

I love this improvement! I've been looking for a solution for a while and here it is.

> The disctotal is not touched, particularly as this is not usually displayed in most clients.

For my use case, I want to zero disctotal as well for single disc releases. Foobar 2000 shows it by default in the properties list. Can we please add in `tags["disctotal "] = None`?


> Adds a zero_disc_number_if_single_disc boolean ..

Calling it "zero_disc_number_if_single_disc" is fine and make sense from an English perspective, but I was initially confused thinking that "disc_number" was referring to a field named just that. I would maybe call it "zero_disc_if_single_disc".

> very ready for merging

Agreed. I've tried your branch with a minimal config on a few releases and it works well!

However, with the configuration I intend to use (see below) where neither `fields` nor `keep_fields` are set, the warning "no fields list to remove" is printed for every item. The plugin docs say "The plugin can work in one of two modes" and this setting, I think, introduces a third mode. What do you think? 

```
zero:
  zero_disc_number_if_single_disc: true
```


Either way, thanks for bringing this change to the table!

michaelkrieger:
> I love this improvement! I've been looking for a solution for a while and here it is.

Glad to have helped.

> > The disctotal is not touched, particularly as this is not usually displayed in most clients.
> For my use case, I want to zero disctotal as well for single disc releases. Foobar 2000 shows it by default in the properties list. Can we please add in `tags["disctotal "] = None`?

Lots of things will show in the properties list.  That said, these typically aren't shown in grid-view or similar, which the disc number typically is.  Once this is merged, you could of course add a _configurable_ option, but I would want it to work as it is for my own use case and preference.

> > Adds a zero_disc_number_if_single_disc boolean ..
> Calling it "zero_disc_number_if_single_disc" is fine and make sense from an English perspective, but I was initially confused thinking that "disc_number" was referring to a field named just that. I would maybe call it "zero_disc_if_single_disc".

I understand your point, and since the field is commonly only referred to as 'disc', I've made that adjustment in the plugin and docs. I do think it's confusing (ie: what is it doing to my disc?), but documentation will clear that up.

> > very ready for merging
> Agreed. I've tried your branch with a minimal config on a few releases and it works well!

Thank you for testing it.

> However, with the configuration I intend to use (see below) where neither `fields` nor `keep_fields` are set, the warning "no fields list to remove" is printed for every item. The plugin docs say "The plugin can work in one of two modes" and this setting, I think, introduces a third mode. What do you think?

I do think that it's odd, but I also think it's the correct behaviour.  There is no list of fields (as defined by the `fields` configuration parameter) for the zero plugin to remove.  This is a warning-level message to help you find configuration challenges, and I don't think it's incorrect.  If you have a configuration problem with fields and keep_fields, I do think it should still throw this warning.

This is why the warning was changed from "no fields, nothing to do" to "no fields list to remove", to remove the (incorrect) warning that it is doing nothing.

> Either way, thanks for bringing this change to the table!

Anytime!

PR #5910, opened 2025-07-31
spotify: Handle API failures

9999years:
Lots of these weren't handled, leading to crashes while importing.

## To Do


beetsplug/spotify.py
```diff
@@ -293,7 +293,13 @@ def album_for_id(self, album_id: str) -> AlbumInfo | None:
...
             return None
 
-        album_data = self._handle_response("get", self.album_url + spotify_id)
+        try:
+            album_data = self._handle_response(
+                "get", self.album_url + spotify_id
+            )
+        except APIError:
+            return None
+
```
snejus, line 296: This approach introduces a lot of duplication in the code. I think a cleaner approach would be to _never raise APIError_ in the first place.

See `_handle_response` method - I'd suggest removing all lines that raise `APIError` and returning `None` instead.

snejus (changes requested):

semohr:
Maybe it is worth taking a step back here:

How do we want to handle errors in metadata plugins in general? E.g. what should happen if one plugin fails because a remote service is currently unreachable? Do we want beets to crash here or just warn a user?

I think raising errors in the specific metadata plugin (such as spotify or deezer) is completely fine. We should handle errors globally for any metadata plugins tho and warn users.

Proposal:  Let's add error handling to the `metadata_plugins.py` file and warn the user that some plugin(s) did not complete a lookup or function call successfully. This would also move responsibilities away from the plugin developers and reduce unnecessary & duplicate error handling.

semohr:
@9999years Are you planning to keep looking into the metadataplugin layer for adding error handling? If you want to take the lead on it, go ahead. I’d be glad to chime in or help out later, since it’s also needed for the [aisauce](https://github.com/metasauce/beets-aisauce) plugin. From what I can tell it shouldn’t be too difficult to add.

9999years:
@semohr Feel free, I might get back to this in the future but I'm not sure when.

PR #5913, opened 2025-08-02
Tighten up docs CI checks

snejus:
This PR adds `sphinx-lint` check for docs which catches issues like line length.


pyproject.toml
```diff
@@ -289,3 +290,7 @@ ignore-variadic-names = true
 [tool.ruff.lint.pep8-naming]
 classmethod-decorators = ["cached_classproperty"]
 extend-ignore-names = ["assert*", "cached_classproperty"]
+
+[tool.esbonio.sphinx]
+buildCommand = ["sphinx-build", "-M", "dirhtml", "docs", "${defaultBuildDir}"]
+pythonCommand = ["${venv:env}"]
```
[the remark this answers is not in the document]
snejus: oops

.github/workflows/lint.yml
```diff
@@ -126,11 +131,13 @@ jobs:
...
         run: poetry install --extras=docs
 
-      - name: Add Sphinx problem matcher
-        run: echo "::add-matcher::.github/sphinx-problem-matcher.json"
+      - name: Add Sphinx problem matchers
+        run: |
+          echo "::add-matcher::.github/problem-matchers/sphinx-build.json"
+          echo "::add-matcher::.github/problem-matchers/sphinx-lint.json"
 
       - name: Build docs
```
semohr: Wouldn't it make more sense to run the linter before running the build command? Or am i missing something here?

snejus: Good point, moved them around now

snejus:
> Looks good generally! I guess you did not do all the changes manually. Could you add a poe command for the auto docs formatting? Also a reference/sentence in the contribution guide would be highly appreciated!

Funnily, I initially __only__ addressed the linting issues (lines being too long) selectively. On the other hand, I guess, we want to have consistent formatting across **all** docs. Thus now I have reformatted everything using `docstrfmt` and added relevant `poe` commands for formatting and linting.

semohr:
Would it be possible to group all automated changes into a single commit and list its hash in the .git-blame-ignore-revs file?

snejus:
That's exactly what I will do once you confirm that we're good to go!

semohr:
> That's exactly what I will do once you confirm that we're good to go!

I guess we should be good to go. All seems reasonable to me :+1:

semohr:
Maybe one more thing could you add the command to the precommit hook too? :pray:

snejus:
> Maybe one more thing could you add the command to the precommit hook too? 🙏

Cool, done it. Squashed docs adjustments to two commits and added them to blame ignore revs.

PR #5924, opened 2025-08-09
Refactored beets/random.py  and moved into beetsplug/random.py

semohr:
The `beets/random.py` module was only used by the random plugin, so I moved its functions into `beetsplug/random.py` to keep core modules cleaner.

Changes:
- Moved beets/random.py functions into beetsplug/random.py
- Added typehints for better readability and tooling support
- Added additional tests for improved coverage
- General tidy up and refactor, keeping the core functionality unchanged


test/plugins/test_random.py
```diff
@@ -77,3 +77,74 @@ def experiment(field, histogram=False):
...
+        """Test _equal_chance_permutation with single item."""
+        result = list(random._equal_chance_permutation([self.item1], "artist"))
+        assert result == [self.item1]
+
+    def test_equal_permutation_single_artist(self):
+        """Test _equal_chance_permutation with items from one artist."""
+        items = [self.create_item(artist=self.artist1) for _ in range(5)]
+        result = list(random._equal_chance_permutation(items, "artist"))
+        assert set(result) == set(items)
+        assert len(result) == len(items)
```
snejus: Any chance you could use `pytest.mark.parametrize` for these tests and refactor the rest to use pytest too?

beetsplug/random.py
```diff
@@ -64,3 +71,107 @@ def random_func(lib, opts, args):
...
     def commands(self):
         return [random_cmd]
+
+
+def _length(obj: Item | Album) -> float:
+    """Get the duration of an item or album."""
+    if isinstance(obj, Album):
+        return sum(i.length for i in obj.items())
+    else:
+        return obj.length
```
snejus: I would suggest defining an abstract property `total_duration` on `LibModel` and implementing it for each model.

This will remove the need for this function as you can just use `item.duration` instead.

beetsplug/random.py
```diff
@@ -64,3 +71,107 @@ def random_func(lib, opts, args):
...
+
+
+def _take(
+    iter: Iterable[T],
+    num: int,
+) -> list[T]:
+    """Return a list containing the first `num` values in `iter` (or
+    fewer, if the iterable ends early).
+    """
+    return list(islice(iter, num))
```
snejus: I doubt there's a need for this function, as it's only used in a single place. There's also no need to cast it to a list.

beetsplug/random.py
```diff
@@ -64,3 +71,107 @@ def random_func(lib, opts, args):
...
+
+
+def _take_time(
```
snejus, line 112: Alternatively, this function could be written in a purely functional way:


```py
from itertools import accumulate, compress

...

def _take_time(item_iter: Iterable[LibModel], secs: float) -> Iterator[LibModel]:
    items = list(item_iter)
    durations = (item.duration for item in items)
    mask = (d <= secs for d in accumulate(durations))
    return compress(items, mask)
```

Again, it's fine to return an iterator here instead of a list.

beetsplug/random.py
```diff
@@ -64,3 +71,107 @@ def random_func(lib, opts, args):
...
+    if isinstance(obj, Album):
+        return sum(i.length for i in obj.items())
+    else:
+        return obj.length
+
+
+def _equal_chance_permutation(
+    objs: Sequence[Item | Album],
+    field: str = "albumartist",
+    random_gen: random.Random | None = None,
```
snejus:
```suggestion
```

This function is never called with these arguments - use them directly within the function implementation instead

beetsplug/random.py
```diff
@@ -64,3 +71,107 @@ def random_func(lib, opts, args):
...
+    """Generate (lazily) a permutation of the objects where every group
+    with equal values for `field` have an equal chance of appearing in
+    any given position.
+    """
+    rand = random_gen or random
+
+    # Group the objects by artist so we can sample from them.
+    key = attrgetter(field)
+    objs = sorted(objs, key=key)
+    objs_by_artists = {}
```
snejus:
```suggestion
    objs_by_artist = {}
    for artist, v in groupby(sorted(objs, key=key), key):
        objs_by_artist[artist] = list(v)
        random.shuffle(objs_by_artist[artist])
```

1. It should be a single artist for correctness
2. A single `random.shuffle` call is less expensive than calling `random.randint` in a loop

beetsplug/random.py
```diff
@@ -64,3 +71,107 @@ def random_func(lib, opts, args):
...
+    return out
+
+
+def random_objs(
+    objs: Sequence[Item | Album],
+    number=1,
+    time: float | None = None,
+    equal_chance: bool = False,
+    random_gen: random.Random | None = None,
+):
```
snejus: This function definition doesn't seem to be correct given the way it's called:

```py
    objs = random_objs(
        objs, opts.album, opts.number, opts.time, opts.equal_chance
    )
```

I think `opts.album` is missing? `random_gen` argument is never used, so you can use `random.shuffle` directly. The rest of arguments should have the defaults removed to reduce confusion - values are _always_ supplied (in a single call).

snejus (changes requested):

beetsplug/random.py
semohr: Leaving this here but it applies to all comments:

This PR was never intended to fix coding related issues, just wanted to remove the random.py file from the beets core. All of the code has no logic changes, I just moved it and added type hints.

That said I can have a iteration here and fix some more of the obvious things, you also noted. Just didnt want to spend this much time here initially, especially as I'm not sure this plugin is even used by many ppl.

semohr: I think this `_take` function is the only one I modified, sine it was obviously highly inefficient and I was wondering what the initial code was even doing.

snejus: I'd be happy to have a small refactor as part of this PR if you're up for it! 😅

beetsplug/random.py
semohr: I opted for a `.length` property on album instead to maintain consistency with the item length attr.

snejus: beautiful

beets/library/models.py
```diff
@@ -616,6 +616,11 @@ def try_sync(self, write, move, inherit=True):
         for item in self.items():
             item.try_sync(write, move)
 
+    @property
```
snejus: This should be a cached property, otherwise it will end up hitting the db on every access

beetsplug/random.py
```diff
@@ -64,3 +62,95 @@ def random_func(lib, opts, args):
...
+            yield obj
+            total_time += length
+
+
+def random_objs(
+    objs: Sequence[T],
+    number: int = 1,
+    time_minutes: float | None = None,
+    equal_chance: bool = False,
+    random_gen: random.Random | None = None,
```
snejus: Defaults are still present even though they have no effect. Actually, why not just passing the entire `opts` object?

beetsplug/random.py
```diff
@@ -64,3 +62,95 @@ def random_func(lib, opts, args):
...
+    rand: random.Random = random_gen or random.Random()
+
+    # Group the objects by artist so we can sample from them.
+    key = attrgetter(field)
+
+    def get_attr(obj: T) -> Any:
+        try:
+            return key(obj)
+        except AttributeError:
+            return NOT_FOUND_SENTINEL
```
snejus: I don't think this can possibly happen?

beetsplug/random.py
```diff
@@ -64,3 +62,95 @@ def random_func(lib, opts, args):
...
+
+    groups: dict[Any, list[T]] = {
+        NOT_FOUND_SENTINEL: [],
+    }
+    for k, values in groupby(objs, key=get_attr):
```
snejus: `objs` need to be sorted before grouping but it seems to have disappeared

snejus:
```suggestion
    for albumartist, objs_iter in groupby(objs, key=get_attr):
```

beetsplug/random.py
```diff
@@ -64,3 +62,95 @@ def random_func(lib, opts, args):
...
+    # Group the objects by artist so we can sample from them.
+    key = attrgetter(field)
+
+    def get_attr(obj: T) -> Any:
+        try:
+            return key(obj)
+        except AttributeError:
+            return NOT_FOUND_SENTINEL
+
+    groups: dict[Any, list[T]] = {
```
snejus: This function only ever handles the `albumartist` field, so the variable name can be more clear and the type of the key a fixed `str`.

```suggestion
    objs_by_albumartist: dict[str, list[T]] = {
```

beetsplug/random.py
```diff
@@ -64,3 +62,95 @@ def random_func(lib, opts, args):
...
+    Args:
+    - objs: The sequence of objects to choose from.
+    - number: The number of objects to select.
+    - time_minutes: If specified, the total length of selected objects
+      should not exceed this many minutes.
+    - equal_chance: If True, each artist has the same chance of being
+      selected, regardless of how many tracks they have.
+    - random_gen: An optional random generator to use for shuffling.
+    """
+    rand: random.Random = random_gen or random.Random()
```
snejus: Why was `random` module replaced by `random.Random` instance?

beetsplug/random.py
```diff
@@ -1,37 +1,35 @@
-# This file is part of beets.
-# Copyright 2016, Philippe Mongeau.
-#
-# Permission is hereby granted, free of charge, to any person obtaining
-# a copy of this software and associated documentation files (the
-# "Software"), to deal in the Software without restriction, including
-# without limitation the rights to use, copy, modify, merge, publish,
```
snejus, line 7: This should probably be kept in place.

beetsplug/random.py
```diff
@@ -1,37 +1,35 @@
...
 from beets.plugins import BeetsPlugin
-from beets.random import random_objs
 from beets.ui import Subcommand, print_
 
+if TYPE_CHECKING:
+    import optparse
+
+    from beets.library import LibModel, Library
+
+    T = TypeVar("T", bound=LibModel)
```
snejus: Is there a reason for a type var? 

I think we should get away by replacing it with `LibModel` since we only depend on `length` and `albumartist` attributes which exist on both models.

For clarity and to help type-checking it would be ideal to define their types under `LibModel`:

```py
    _format_config_key: str
    albumartist: str
    length: float
    path: bytes
```

snejus (changes requested):
A couple of more comments. I found that unused default arguments are still present, so I unresolved my previous two comments where I commented about them.

beetsplug/random.py
semohr: Let's do parameter parsing as early as possible. Am not a big fan passing unknown/untyped values.

We use the default parameters in the tests all over the place. I would keep them just for usability even tho they are not used in the main function call. Doesn't really hurt imo.

beetsplug/random.py
semohr: This can occur if we reference a field that might not be defined on an object. In practice, this is very unlikely and currently not possible given how the function is invoked. However, we do have a test for it, as it is technically within the realm of possibility.

I’d prefer to keep the function as generic as possible, since this flexibility could be useful if we decide to extend the random plugin with additional features in the future.

beetsplug/random.py
semohr: You have seen that this was not the standard copyright notice?

If we want to keep it, should I add my own name?

beetsplug/random.py
semohr: Indeed, typevar here should be Item|Album here.

beetsplug/random.py
semohr: It is mostly the same thing. The random module is a hidden `random.Random` instance. Typehints do not work with the default random module for some reason. Also this is more readable in my opinion.

beetsplug/random.py
semohr: At the moment that is true but as I said in another comment. I’d prefer to keep the function as generic as possible, since this flexibility could be useful if we decide to extend the random plugin with additional features in the future.

Beeing able to define any field is a nice feature even if unused atm.

beetsplug/random.py
semohr: See other comments. Also we need to keep the random gen variable to ensure reproducible tests here.

beets/library/models.py
semohr: I thought the items function is cached, but I guess that PR was not merged yet.

beetsplug/random.py
semohr: See other comment on fields. Also seems like a nit not really worth our time tbh ;)

beetsplug/random.py
semohr: > For clarity and to help type-checking it would be ideal to define their types under LibModel

Yeah would be nice to identify more shared functionalities. I think this is out of scope here tho.

beetsplug/random.py
snejus: We should solve the problem at hand and not introduce unnecessary complexity based on a __guess__ about the future.

I expanded about the dangers of this anti-pattern at length under #5787.

beetsplug/random.py
snejus: The issue with type hints disappears once it's removed from function parameters and we use `random` module directly in those three places.

Regarding

> It is mostly the same thing. The random module is a hidden random.Random instance.

> Also we need to keep the random gen variable to ensure reproducible tests here.

Not sure if I'm missing something but seeding the `random` module globally once seems to be more straightforward and reliable than depending on a seeded `random.Random` instance being passed across the functions?

beetsplug/random.py
semohr: I see this as a slightly different situation, we’re adjusting existing features where the original code didn’t fully cover edge cases to begin with.  

That said, I can easily add support for random candidate selection based on another field. Maybe we could focus on that as a compromise? 

---

I'll be honest about my intentions here: I mainly wanted to clean up the main package by moving the unused random logic. I don't have particular investment in this plugin itself.

I'm growing concerned that we might be over-optimizing certain areas at the expense of more impactful work. In my experience, many of these patterns are matters of preference rather than clear right-or-wrong decisions.

While I respect that we may have different opinions on implementation details, I've been consciously trying to adopt a 'good enough' approach for changes that:
1. Don't break ux and dx 
2. Are functionally correct
3. Maintain reasonable code quality

This seems especially relevant for non-core components like random plugins, where perfect might be the enemy of good. Could we discuss how to better align on review priorities? I'd love to find a way to move faster on uncontroversial improvements while reserving deeper scrutiny for truly critical changes

beetsplug/random.py
semohr: I wouldn't remove the random parameter as this makes our tests easily reproducible and importantly isolated. While it is probably possible to adapt the tests and create a proper cleanup function for a global rng state, I dont really want to spend the effort on it since the current version is functional.

Feel free to jump in and drop a commit tho. I don't mind.

beetsplug/random.py
snejus: How come? 
image

beetsplug/random.py
snejus: I did edit the source code before I commented to double-check that it was valid, so I committed and pushed it now.

beetsplug/random.py
semohr: Ship of Theseus :smile:

beetsplug/random.py
snejus: Apologies if I wasn't clear - I don't see why are we using `TypeVar` here instead of the simpler `LibModel` alternative.

As I said above, this relates to the `length` property, which is currently not resolved properly (unknown for `Item`) and thus isn't type-checked:

```py
def _take_time(
    iter: Iterable[T],
    secs: float,
) -> Iterable[T]:
    """Return a list containing the first values in `iter`, which should
    be Item or Album objects, that add up to the given amount of time in
    seconds.
    """
    total_time = 0.0
    for obj in iter:
        length = obj.length
        reveal_type(obj.length)  # mypy: Revealed type is "Union[Any, builtins.float]"
        if total_time + length <= secs:
            yield obj
            total_time += length
```

Instead, defining `LibModel.length` type

```py
class LibModel(dbcore.Model["Library"]):
    """Shared concrete functionality for Items and Albums."""

    # Config key that specifies how an instance should be formatted.
    _format_config_key: str
    path: bytes
    length: float
```

and replacing the type var with `LibModel` fixes it:

```py
def _take_time(
    iter: Iterable[LibModel],
    secs: float,
) -> Iterable[LibModel]:
    """Return a list containing the first values in `iter`, which should
    be Item or Album objects, that add up to the given amount of time in
    seconds.
    """
    total_time = 0.0
    for obj in iter:
        length = obj.length
        reveal_type(obj.length)  # mypy: Revealed type is "builtins.float"
        if total_time + length <= secs:
            yield obj
            total_time += length
```

Additionally, this prevents `Item` or `Album` from returning any other type for this property.

semohr: Indeed, didn't get your original point! I did not notice the db fields getters are not typed for the libmodels. Your solution would work :+1: 

As an alternative, how do you feel about generating types for the attrgetters in general? Would also make using beets downstream alot easier. I have an idea on how that could work and some time later this week to implementing a prototype.

snejus: Yeah, this would be great!

snejus: Shall I commit the replacement of `T` by `LibModel` or are you happy doing it yourself?

semohr: Feel free! I don't think we can add a `.length` property to the LibModel without introducing mypy issues. Keeping it as `T = Union[Item, Album]` might be the only way for now.

snejus: > I don't think we can add a `.length` property to the LibModel

Note we're adding a __type__ definition to `LibModel` instead of a property. I mean, ultimately I am trying to make `mypy` happy rather than angry 😁

beetsplug/random.py
snejus: Please keep the original copyright - this PR only refactors the plugin and adds an additional field

beetsplug/random.py
snejus: > That said, I can easily add support for random candidate selection based on another field. Maybe we could focus on that as a compromise?

Of course!

beetsplug/random.py
```diff
@@ -57,10 +54,105 @@ def random_func(lib, opts, args):
     type="float",
     help="total length in minutes of objects to choose",
 )
+random_cmd.parser.add_option(
```
snejus, line 71: I see you've added support for a configurable equal chance field. However, this breaks the plugin: 

```sh
$ beet version
** error loading plugin random:
  File "/home/sarunas/repo/beets/beets/plugins.py", line 286, in load_plugins
    namespace = __import__(modname, None, None)
  File "/home/sarunas/repo/beets/beetsplug/random.py", line 65, in <module>
    random_cmd.parser.add_all_common_options()
  File "/home/sarunas/repo/beets/beets/ui/__init__.py", line 1386, in add_all_common_options
    self.add_format_option()
  File "/home/sarunas/repo/beets/beets/ui/__init__.py", line 1380, in add_format_option
    self.add_option(opt)
  File "/home/sarunas/.local/share/uv/python/cpython-3.10.18-linux-x86_64-gnu/lib/python3.10/optparse.py", line 1008, in add_option
    self._check_conflict(option)
  File "/home/sarunas/.local/share/uv/python/cpython-3.10.18-linux-x86_64-gnu/lib/python3.10/optparse.py", line 980, in _check_conflict
    raise OptionConflictError(
optparse.OptionConflictError: option -f/--format: conflicting option string(s): -f

beets version 2.3.1
Python version 3.10.18
plugins: autobpm, bandcamp, bareasc, bench, deezer, discogs, edit, embedart, fetchart, importreplace, info, inline, lastgenre, lastimport, lyrics, mbcollection, mbsync, missing, mpdstats, mpdupdate, musicbrainz, parentwork, replaygain, scrub, spotify, types, unimported
```

beetsplug/random.py
```diff
@@ -57,10 +54,105 @@ def random_func(lib, opts, args):
...
+    "-f",
+    "--field",
+    action="store",
+    type="string",
+    help="field to use for equal chance sampling (default: albumartist)",
```
snejus, line 76:
```suggestion
    type="string",
    default="albumartist",
    help="field to use for equal chance sampling (default: albumartist)",
```

beetsplug/random.py
```diff
@@ -57,10 +54,105 @@ def random_func(lib, opts, args):
     type="float",
     help="total length in minutes of objects to choose",
 )
+random_cmd.parser.add_option(
+    "-f",
+    "--field",
```
snejus, line 72: References to `artist` need to be replaced with a generic field:

```diff
diff --git a/beetsplug/random.py b/beetsplug/random.py
index ba687f477..70e48c565 100644
--- a/beetsplug/random.py
+++ b/beetsplug/random.py
@@ -47,3 +47,3 @@ def random_func
     action="store_true",
-    help="each artist has the same chance",
+    help="each equal chance field has the same chance",
 )
@@ -82,3 +82,3 @@ def _equal_chance_permutation
     """
-    # Group the objects by artist so we can sample from them.
+    # Group the objects by field so we can sample from them.
     key = attrgetter(field)
@@ -137,2 +137,3 @@ def random_objs
     - objs: The sequence of objects to choose from.
+    - equal_chance_field: ...
     - number: The number of objects to select.
@@ -140,4 +141,4 @@ def random_objs
       should not exceed this many minutes.
-    - equal_chance: If True, each artist has the same chance of being
-      selected, regardless of how many tracks they have.
+    - equal_chance: If True, each equal_chance_field has the same chance of
+      being selected, regardless of how many tracks they have.
     - random_gen: An optional random generator to use for shuffling.
@@ -145,3 +146,3 @@ def random_objs
     # Permute the objects either in a straightforward way or an
-    # artist-balanced way.
+    # equal_chance_field-balanced way.
     perm: Iterable[LibModel]
```

beetsplug/random.py
```diff
@@ -57,10 +54,105 @@ def random_func(lib, opts, args):
...
+            yield obj
+            total_time += length
+
+
+def random_objs(
+    objs: Iterable[LibModel],
+    number: int = 1,
+    time_minutes: float | None = None,
+    equal_chance: bool = False,
+    equal_chance_field: str = "albumartist",
```
snejus: Once `default` is specified, `equal_chance_field` is always present as a `str` so we can remove the ambiguity by making it a required argument

```suggestion
    objs: Iterable[LibModel],
    equal_chance_field: str,
    number: int = 1,
    time_minutes: float | None = None,
    equal_chance: bool = False,
```

beetsplug/random.py
```diff
@@ -57,10 +54,105 @@ def random_func(lib, opts, args):
...
 class Random(BeetsPlugin):
     def commands(self):
         return [random_cmd]
+
+
+NOT_FOUND_SENTINEL = object()
+
+
+def _equal_chance_permutation(
+    objs: Iterable[LibModel], field: str = "albumartist"
```
snejus: Same here

```suggestion
    objs: Iterable[LibModel], field: str
```

beetsplug/random.py
```diff
@@ -57,10 +54,105 @@ def random_func(lib, opts, args):
...
+      should not exceed this many minutes.
+    - equal_chance: If True, each artist has the same chance of being
+      selected, regardless of how many tracks they have.
+    - random_gen: An optional random generator to use for shuffling.
+    """
+    # Permute the objects either in a straightforward way or an
+    # artist-balanced way.
+    perm: Iterable[LibModel]
+    if equal_chance:
+        perm = _equal_chance_permutation(objs, field=equal_chance_field)
```
snejus:
```suggestion
        perm = _equal_chance_permutation(objs, equal_chance_field)
```

beetsplug/random.py
```diff
@@ -57,10 +54,105 @@ def random_func(lib, opts, args):
...
+        group = random.choice(list(groups.keys()))
+        yield groups[group].pop()
+        if not groups[group]:
+            del groups[group]
+
+
+def _take_time(
+    iter: Iterable[LibModel],
+    secs: float,
+) -> Iterable[LibModel]:
```
snejus, line 115: Return type should be as specific as possible

```suggestion
) -> list[LibModel]:
```

beetsplug/random.py
```diff
@@ -57,10 +54,105 @@ def random_func(lib, opts, args):
...
+    - number: The number of objects to select.
+    - time_minutes: If specified, the total length of selected objects
+      should not exceed this many minutes.
+    - equal_chance: If True, each artist has the same chance of being
+      selected, regardless of how many tracks they have.
+    - random_gen: An optional random generator to use for shuffling.
+    """
+    # Permute the objects either in a straightforward way or an
+    # artist-balanced way.
+    perm: Iterable[LibModel]
```
snejus: No need

```suggestion
```

snejus (changes requested):
Since you've added `field` argument, I added a couple of related comments. This should be mentioned in the changelog as well.

beetsplug/random.py
semohr: `_take_time` returns a generator. We can do `Generator[LibModel,None,None]` instead but I would prefer `Iterable` here.

beetsplug/random.py
semohr: Removed the `-f` shorthand.

beetsplug/random.py
snejus: Apologies 🤦🏼 I was supposed to comment under `_equal_chance_permutation` function

semohr: The `_equal_chance_permutation` also yields and is a generator as far as i can see :thinking: I might be missing something but I dont see it.

snejus: Now I see... I'm officially a muppet

beetsplug/random.py
```diff
@@ -57,10 +68,104 @@ def random_func(lib, opts, args):
...
+        try:
+            return key(obj)
+        except AttributeError:
+            return NOT_FOUND_SENTINEL
+
+    sorted(objs, key=get_attr)
+
+    groups: dict[str | object, list[LibModel]] = {
+        NOT_FOUND_SENTINEL: [],
+    }
```
snejus: `LibModel` is a dictionary - so this can be simplified with using `LibModel.get` instead of try/except + sentinel. 

```suggestion
    # Group the objects by field so we can sample from them.
    get_attr = methodcaller("get", field)

    sorted(objs, key=get_attr)

    groups = {}
```

beetsplug/random.py
```diff
@@ -57,10 +68,104 @@ def random_func(lib, opts, args):
...
+    # Group the objects by field so we can sample from them.
+    key = attrgetter(field)
+
+    def get_attr(obj: LibModel) -> Any:
+        try:
+            return key(obj)
+        except AttributeError:
+            return NOT_FOUND_SENTINEL
+
+    sorted(objs, key=get_attr)
```
snejus: This has no effect - you want to either move it to 
```py
    for k, values in groupby(sorted(objs, key=get_attr), key=get_attr):
```

or better: revert to what we had previously by casting entities to `list` and send `list[LibModel]` through:

```diff
diff --git a/beetsplug/random.py b/beetsplug/random.py
index 9714c8e53..44adc9fbe 100644
--- a/beetsplug/random.py
+++ b/beetsplug/random.py
@@ -33,7 +33,7 @@
 def random_func(lib: Library, opts: optparse.Values, args: list[str]):
     """Select some random items or albums and print the results."""
     # Fetch all the objects matching the query into a list.
-    objs = lib.albums(args) if opts.album else lib.items(args)
+    objs = list(lib.albums(args) if opts.album else lib.items(args))
 
     # Print a random subset.
     for obj in random_objs(
```

which allows

```suggestion
    objs.sort(key=get_attr)
```

beetsplug/random.py
```diff
@@ -57,10 +68,104 @@ def random_func(lib, opts, args):
...
+    groups: dict[str | object, list[LibModel]] = {
+        NOT_FOUND_SENTINEL: [],
+    }
+    for k, values in groupby(objs, key=get_attr):
+        groups[k] = list(values)
+        # shuffle in category
+        random.shuffle(groups[k])
+
+    # Remove items without the field value.
+    del groups[NOT_FOUND_SENTINEL]
```
snejus: 1. Need to handle list fields like `artists`, `albumtypes` - we can just cast the keys to `str`
2. Skip missing (`None`) values removing the need to delete them later

```suggestion
        if k is not None:
            vals = list(values)
            # shuffle in category
            random.shuffle(vals)
            groups[str(k)] = vals

```

snejus (changes requested):
I gave this a test and found two bugs:

1. List fields are not handled
```sh
$ beet random --field artists -e
  File "/home/sarunas/.local/bin/beet", line 8, in <module>
    main()
  File "/home/sarunas/repo/beets/beets/ui/__init__.py", line 1629, in main
    _raw_main(args)
  File "/home/sarunas/repo/beets/beets/ui/__init__.py", line 1608, in _raw_main
  File "/home/sarunas/repo/beets/beetsplug/random.py", line 39, in random_func
    for obj in random_objs(
  File "/home/sarunas/repo/beets/beetsplug/random.py", line 112, in _equal_chance_permutation
    groups[k] = list(values)
TypeError: unhashable type: 'list'```
```

2. items aren't sorted before grouping in equal chance permutation. See `beet list` output:

```sh
$ beet list path::aaa title- -f '$album: $artist - $title'
BLAH!!!! malakaa!! todotiernos de mierdaaa: DER SANDMANN - Triple Masala
BLAH!!!! malakaa!! todotiernos de mierdaaa: DER SANDMANN - Machete
BLAH!!!! malakaa!! todotiernos de mierdaaa: DER SANDMANN - Lost in Time
BLAH!!!! malakaa!! todotiernos de mierdaaa: DER SANDMANN - Freaks at Work
BLAH!!!! malakaa!! todotiernos de mierdaaa: DER SANDMANN - der killer wahl 185bpm
RAVE SLUTZ: Fallen Shrine & dj Christian NXC - DEEEJAAAY
BLAH!!!! malakaa!! todotiernos de mierdaaa: DER SANDMANN - Blah
Netikras albumas: Karpiz - AAA
```

vs

```sh
$ beet random path::aaa title- -f '$album: $artist - $title' --field album -e -n10
RAVE SLUTZ: Fallen Shrine & dj Christian NXC - DEEEJAAAY
Netikras albumas: Karpiz - AAA
BLAH!!!! malakaa!! todotiernos de mierdaaa: DER SANDMANN - Blah
```

I expected to see the same items but in random order.

I'm keen to finally wrap this up, so feel free to just paste this fix:

```diff
diff --git a/beetsplug/random.py b/beetsplug/random.py
index 9714c8e53..aa65dd5d5 100644
--- a/beetsplug/random.py
+++ b/beetsplug/random.py
@@ -19,4 +19,4 @@
 from itertools import groupby, islice
-from operator import attrgetter
-from typing import TYPE_CHECKING, Any
+from operator import methodcaller
+from typing import TYPE_CHECKING
 
@@ -86,5 +86,2 @@ def commands
 
-NOT_FOUND_SENTINEL = object()
-
-
 def _equal_chance_permutation(
@@ -97,22 +94,12 @@ def _equal_chance_permutation
     # Group the objects by field so we can sample from them.
-    key = attrgetter(field)
-
-    def get_attr(obj: LibModel) -> Any:
-        try:
-            return key(obj)
-        except AttributeError:
-            return NOT_FOUND_SENTINEL
-
-    sorted(objs, key=get_attr)
-
-    groups: dict[str | object, list[LibModel]] = {
-        NOT_FOUND_SENTINEL: [],
-    }
-    for k, values in groupby(objs, key=get_attr):
-        groups[k] = list(values)
-        # shuffle in category
-        random.shuffle(groups[k])
-
-    # Remove items without the field value.
-    del groups[NOT_FOUND_SENTINEL]
+    get_attr = methodcaller("get", field)
+
+    groups = {}
+    for k, values in groupby(sorted(objs, key=get_attr), key=get_attr):
+        if k is not None:
+            vals = list(values)
+            # shuffle in category
+            random.shuffle(vals)
+            groups[str(k)] = vals
+
     while groups:
```

snejus (approved):

PR #5926, opened 2025-08-10
"Add support for creating indices on arbitrary database tables

### LINKED ISSUE #5809: Missing Indexes in SQLite Database Causing Performance Bottlenecks
I've noticed that certain queries on the SQLite database used by beets—particularly album.items()—are quite slow on my machine.

After some investigation, I found that the database does not currently have any indexes defined. Specifically, the items table lacks an index on the album_id column, which seems to be the root cause of the slowdown when querying items by album.

I manually created an index on album_id, and the performance improvement was significant—query times dropped by approx a factor of 4 in my benchmarks.

## Suggestion

Adding indexes for frequently queried fields (like `album_id`) could provide a substantial performance boost for many users. However, I'm not very familiar with the dbcore layer and I'm unsure how or where to properly define these indexes within the beets codebase. I dont think we have a mechanism in place for applying any kind of database migrations.

## Temporary Workaround

For anyone else running into similar performance issues, or if you just want to test this, here's a small script you can use to manually add the index and run a benchmark on your library.

```python
import sqlite3
import time
from pathlib import Path

from beets.library import Library

# Configure your db path
db_path = Path("./library.db")
idx_name = "items_album_id_idx"


def run_sql_exc(path, query):
    conn = sqlite3.connect(path)
    cursor = conn.cursor()
    cursor.execute(query)
    conn.commit()
    conn.close()


def run_benchmark(lib, indexed=True):
    time_start = time.perf_counter()
    albums = lib.albums()
    items = []
    for album in albums:
        items.extend(album.items())
    duration = time.perf_counter() - time_start
    nA = len(albums)
    nI = len(items)

    print(
        f"Benchmark with{'out' if indexed else None} index took {duration:.2f} seconds"
    )
    print(f"\tnAlbums: {nA} (avgtime {duration * 1000 / nA} ms per matchquery)")
    print(f"\tnItems:  {nI} (avgtime {duration * 1000 / nI:.2f} ms per item)")


if __name__ == "__main__":
    run_sql_exc(db_path, 'DROP INDEX IF EXISTS "{}"'.format(idx_name))
    print("Benchmark without index...")
    run_benchmark(Library(db_path.name, str(db_path.resolve().parent)))
    print("Creating index on items.album_id...")
    run_sql_exc(
        db_path, 'CREATE INDEX IF NOT EXISTS "{}" ON items (album_id)'.format(idx_name)
    )
    print("Benchmark with index...")
    run_benchmark(Library(db_path.name, str(db_path.resolve().parent)))
    print("Done.")
```

For me adding a singular index on items yields the following improvements:
```shell
Benchmark without index...
Benchmark without index took 2.27 seconds
	nAlbums: 521 (avgtime 4.364256262957434 ms per matchquery)
	nItems:  3977 (avgtime 0.57 ms per item)
Creating index on items.album_id...
Benchmark with index...
Benchmark with index took 0.66 seconds
	nAlbums: 521 (avgtime 1.2675718119012231 ms per matchquery)
	nItems:  3977 (avgtime 0.17 ms per item)
Done.
```

semohr:
Creating indexes turned out to be relatively straightforward!
While we can’t remove them yet, that doesn’t seem necessary for now. Interestingly, much of the infrastructure for database additions was already in place.

- This PR introduces a new type, `Index`, which can be used to create an index on any defined table. :tada:
- The `items` table now automatically registers an index on `album_id`

Closes #5809


beets/dbcore/types.py
```diff
@@ -454,6 +462,21 @@ def parse(self, string):
                 return self.null
 
 
+class Index(NamedTuple):
```
snejus: I don't think this is needed (I usually avoid creating a data container until a have a need to add some sort of property or a method to it). 

I think a plain

```py
_indices: Sequence[tuple[str, [tuple[str, ...]]]

...

_indices = (("idx_item_album_id", ("album_id")))
```

Is clear enough.

That said, the following method would make it worth keeping:

```py
class Index(NamedTuple):

    ...

    def create(self, tx: Transaction, table: str) -> None:
        tx.script(f"CREATE INDEX {self.name} ON {table} ({', '.join(self.columns)})")
```

beets/dbcore/types.py
```diff
@@ -454,6 +462,21 @@ def parse(self, string):
...
+    """A database index.
+
+    A helper class to represent the index
+    information in the schema.
+    """
```
snejus:
```suggestion
    """A helper class to represent the index information in the schema."""
```

beets/dbcore/types.py
```diff
@@ -454,6 +462,21 @@ def parse(self, string):
...
+
+    name: str
+    columns: Sequence[str]
+    table: str | None = None
+    """As a index is normally bound to a table
+    we may omit it here.
+    """
```
snejus: No need for this as it's not used.
```suggestion
```

Though note that if you ever need to document an attribute, you may consider the following format

```py
    #: As a index is normally bound to a table we may omit it here.
    table: str | None = None
```

https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#doc-comments-and-docstrings

beets/dbcore/types.py
snejus: Move this to `dbcore/db.py` to remove any potential ambiguity regarding database field types.

docs/changelog.rst
```diff
@@ -57,6 +57,9 @@ Bug fixes:
   :bug:`5699`
 * :doc:`plugins/discogs`: Beets will no longer crash if a release has been
   deleted, and returns a 404.
+* Beets now creates an index for ``album_id`` on the ``items`` sqlite table.
+  This speeds up queries that filter items by their album significantly.
+  :bug:`5809`
```
```suggestion
* Beets now creates an index for ``album_id`` field in the ``items`` table.
  This significantly speeds up queries that filter items by their album.
  :bug:`5809`
```

beets/dbcore/db.py
```diff
@@ -1153,20 +1160,32 @@ def load_extension(self, path: str):
 
     # Schema setup and migration.
 
-    def _make_table(self, table: str, fields: Mapping[str, types.Type]):
+    def _make_table(
+        self,
+        table: str,
+        fields: Mapping[str, types.Type],
+        indices: Sequence[types.Index],
```
snejus: Let's just pass the model instance here to make it tidier

beets/dbcore/db.py
```diff
@@ -1153,20 +1160,32 @@ def load_extension(self, path: str):
...
+        # Skip table creation if the current schema matches the
+        # requested schema (and no indexes are missing).
         field_names = set(fields.keys())
-        if current_fields.issuperset(field_names):
-            # Table exists and has all the required columns.
+        index_names = {index.name for index in indices}
+        if current_fields.issuperset(
+            field_names
+        ) and current_indices.issuperset(index_names):
             return
```
snejus: I'm not a big fan of this early return and the ambiguous `else` condition below
```suggestion
```

I think the logic would be more clear if the condition is explicit:

```diff
-        else:
+        elif fields.keys() <= current_fields:
```

beets/dbcore/db.py
```diff
@@ -1189,6 +1208,17 @@ def _make_table(self, table: str, fields: Mapping[str, types.Type]):
...
+        # Index handling
+        with self.transaction() as tx:
+            for index in indices:
+                if index.name in current_indices:
+                    continue
+
+                columns_str = ", ".join(index.columns)
+                tx.script(
+                    f"CREATE INDEX {index.name} ON {table} ({columns_str})"
+                )
```
snejus: Based on my previous comment, I think it's also more clear to avoid GOTO equivalents like `continue`

```suggestion
        current_indices = {row[1] for row in index_rows}
        with self.transaction() as tx:
            for index in indices:
                if index.name not in current_indices:
                    index.create(tx, table)

```

semohr: I prefer to keep control flow simple and avoid deep nesting. That’s why I often lean toward early returns and continue statements, they help. 

Also I wanted the new code style to match the old code, which also used the continue. Maybe we want to change this around for all code in a more overarching PR?

beets/dbcore/types.py
semohr: I mainly use them as typehints tbh and in this case also for the increased performance (in the case of NamedTuple). 

Adding a create method is a good idea tho!

beets/dbcore/types.py
semohr: I don't think the one you quoted from the sphinx docs works for standard python intropection. I will continue using the [PEP258](https://peps.python.org/pep-0258/) standard for attribute docstrings :upside_down_face: 

FYI: Sphinx also supports the normal attribute comments, as stated in their docs.

beets/dbcore/db.py
snejus: > I prefer to keep control flow simple

We're on the same page! Thus my suggestion 😆 

I asked for an objective opinion about which alternative is simpler: https://www.perplexity.ai/search/which-one-is-simpler-to-compre-iOZtRnDoRUySe6HMw1wVVQ

> I often lean toward early returns and continue statements, they help.

I use the former for the exact same purpose as you mentioned:

> to keep control flow simple and avoid deep nesting

And I __avoid__ any goto-like statements like `continue` or `break` since my brain finds continuous logic easier to deal with.

snejus: > Also I wanted the new code style to match the old code, which also used the continue. Maybe we want to change this around for all code in a more overarching PR? 

You're extending the functionality of this function significantly, refactoring/optimising the existing conditional logic seems to be within the scope of your change. And especially so because this impacts the new logic you're adding.

snejus: Note I considered suggesting to replace multiple `SELECT * FROM table/index_info` with a single `SELECT * FROM sqlite_master` query to return data about **_all_** models and indices immediately but thought this would probably be out of scope of our adjustment.

semohr: > > I prefer to keep control flow simple
> 
> We're on the same page! Thus my suggestion 😆
> 
> I asked for an objective opinion about which alternative is simpler: https://www.perplexity.ai/search/which-one-is-simpler-to-compre-iOZtRnDoRUySe6HMw1wVVQ
> 
> > I often lean toward early returns and continue statements, they help.
> 
> I use the former for the exact same purpose as you mentioned:
> 
> > to keep control flow simple and avoid deep nesting
> 
> And I **avoid** any goto-like statements like `continue` or `break` since my brain finds continuous logic easier to deal with.

It’s really just a personal preference, probably influenced by my Fortran and C background. I’m fine either way. Early returns tho can be great for skipping unnecessary checks, especially in interpreted code, where (if I understand correctly) they prevent evaluating the rest of the conditions.

> > Also I wanted the new code style to match the old code, which also used the continue. Maybe we want to change this around for all code in a more overarching PR?
> 
> You're extending the functionality of this function significantly, refactoring/optimising the existing conditional logic seems to be within the scope of your change. And especially so because this impacts the new logic you're adding.

We do use continue in this way extensively, 42 times in beets and 81 times in beetsplug, so if we want to change that, I think it’s best tackled in an overarching style review. I did adapt the new code here, but I wouldn’t want to start reworking well-functioning code purely for style. That tends to make git blame less useful and can distract from the main goal of the PR.

I’d rather keep changes here focused on the feature itself and save broader style adjustments for a dedicated pass, that way we keep moving quickly without losing sight of code consistency.

> Note I considered suggesting to replace multiple `SELECT * FROM table/index_info` with a single `SELECT * FROM sqlite_master` query to return data about **_all_** models and indices immediately but thought this would probably be out of scope of our adjustment.

I also considered creating the Index class directly from a sqlite_master query, but that would require some string parsing, which I’d prefer to avoid. I opted to add a from_db method and do the checks with sets directly which is way cleaner now :+1:

beets/dbcore/db.py
```diff
@@ -1207,6 +1213,33 @@ def _make_attribute_table(self, flex_table: str):
...
+
+        if current_indices.issuperset(_indices):
+            return
+
+        # May also include missing indices.
+        changed_indices = _indices - current_indices
+
+        with self.transaction() as tx:
+            for index in changed_indices:
+                index.recreate(tx, table)
```
snejus: This can be slightly simplified

```suggestion
        with self.transaction() as tx:
            current = {
                Index.from_db(tx, r[1])
                for r in tx.query(f"PRAGMA index_list({table})")
            }
            for index in set(indices) - current:
                index.recreate(tx, table)
```

beets/dbcore/db.py
```diff
@@ -1276,3 +1309,42 @@ def _get(
...
+        updated.
+        """
+        tx.script(f"DROP INDEX IF EXISTS {self.name}")
+        self.create(tx, table)
+
+    def create(self, tx: Transaction, table: str) -> None:
+        """Create the index in the database."""
+        return tx.script(
+            f"CREATE INDEX {self.name} ON {table} ({', '.join(self.columns)})"
+        )
```
snejus: Is there a reason why these two instructions are in separate methods?

```suggestion
        tx.script(f"""
            DROP INDEX IF EXISTS {self.name};
            CREATE INDEX {self.name} ON {table} ({", ".join(self.columns)})
        """)
```

beets/library/models.py
```diff
@@ -716,6 +716,12 @@ class Item(LibModel):
         "mtime": types.DATE,
         "added": types.DATE,
     }
+    _indices = (
+        dbcore.Index(
+            name="idx_item_album_id",
+            columns=("album_id",),
+        ),
+    )
```
snejus: Would be nice to see it fit in a single line
```suggestion
    _indices = (dbcore.Index("idx_item_album_id", ("album_id",),)
```

beets/dbcore/db.py
```diff
@@ -1276,3 +1309,42 @@ def _get(
...
         return self._fetch(model_cls, MatchQuery("id", id)).get()
+
+
+class Index(NamedTuple):
+    """A helper class to represent the index
+    information in the database schema.
+    """
+
+    name: str
+    columns: Sequence[str]
```
snejus: Is there a need for this type to be flexible? 
```suggestion
    columns: tuple[str, ...]
```

Any reason for a flexible type? I'd rather enforce a tuple and remove the `tuple` call under `__hash__`.

The issue with Sequence is that it's happy receiving a string 😅

beets/dbcore/db.py
```diff
@@ -1276,3 +1309,42 @@ def _get(
...
+        """Create an Index object from the database if it exists.
+
+        The name has to exists in the database! Otherwise, an
+        Error will be raised.
+        """
+        rows = tx.query(f"PRAGMA index_info({name})")
+        columns = [row[2] for row in rows]
+        return cls(name, columns)
+
+    def __hash__(self) -> int:
```
snejus: A neat way to reduce the comparison complexity! We just may want to define it [above the other methods](https://stackoverflow.com/a/36967648)

beets/dbcore/db.py
```diff
@@ -1276,3 +1309,42 @@ def _get(
...
+        self.create(tx, table)
+
+    def create(self, tx: Transaction, table: str) -> None:
+        """Create the index in the database."""
+        return tx.script(
+            f"CREATE INDEX {self.name} ON {table} ({', '.join(self.columns)})"
+        )
+
+    @classmethod
+    def from_db(cls, tx: Transaction, name: str) -> Index:
```
snejus: Good idea!

beets/dbcore/db.py
semohr: Nope, had some more functions for testing and never removed the create function :+1:

beets/dbcore/db.py
semohr: Indeed, tuple should be fine :+1:

beets/dbcore/db.py
semohr: Is the index that important? I always try to order by importance top down.

snejus:
I just stumbled upon this function in the same module:

```py
    def _make_attribute_table(self, flex_table: str):
        """Create a table and associated index for flexible attributes
        for the given entity (if they don't exist).
        """
        with self.transaction() as tx:
            tx.script(
                """
                CREATE TABLE IF NOT EXISTS {0} (
                    id INTEGER PRIMARY KEY,
                    entity_id INTEGER,
                    key TEXT,
                    value TEXT,
                    UNIQUE(entity_id, key) ON CONFLICT REPLACE);
                CREATE INDEX IF NOT EXISTS {0}_by_entity
                    ON {0} (entity_id);
                """.format(flex_table)
            )
```

I didn't realise we've already been creating indices for attributes tables! 

As you mentioned, consistency is important, and it seems that now we use two methods to create indices. Looking at the simplicity of the above command, I'm now questioning whether we really need to depend on the entire complexity of checking what indices we already have in the database. Can't we get away with the following instead?

```diff
diff --git a/beets/dbcore/db.py b/beets/dbcore/db.py
index 2e0df3fec..a54dabd7f 100755
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -1159,10 +1159,11 @@ def load_extension
 
     # Schema setup and migration.
 
-    def _make_table(self, table: str, fields: Mapping[str, types.Type]):
+    def _make_table(self, model_cls: Model):
         """Set up the schema of the database. `fields` is a mapping
         from field names to `Type`s. Columns are added if necessary.
         """
+        table, fields = model_cls._table, model_cls._fields
         # Get current schema.
         with self.transaction() as tx:
             rows = tx.query("PRAGMA table_info(%s)" % table)
@@ -1192,6 +1193,14 @@ def _make_table
                     table, name, typ.sql
                 )
 
+        create_idx_cmds = [
+            f"CREATE INDEX IF NOT EXISTS {i.name} ON {table} ({', '.join(i.columns)});"
+            for i in model_cls._indices
+        ]
+
+        setup_sql += "\n".join(create_idx_cmds)
+
         with self.transaction() as tx:
             tx.script(setup_sql)
 
```

semohr:
> I didn't realise we've already been creating indices for attributes tables!

I have seen that  when creating the indices. For me the whole set of attribute tables seem a bit inconsistent in the first place as they feel like a bolt-on rather than a first-class part of the schema, more like a key/value store glued on top of the ORM than something that plays nicely with the rest of the core tables. Long term, I think they’d probably benefit from a bigger refactor to align with the rest of the core layer. That’s also why I originally ignored it :upside_down_face:

---

From what I understand, the core was originally meant to function like a lightweight ORM, with an eye toward adding more tables and features to beets in the future. If we only care about functionality, we could simplify things quite a bit in that layer. Personally, though, I’m fine with a bit of added complexity here since it fits with the original ORM idea.

The advantage of this approach is that it gives us a path to evolve the schema more cleanly, e.g. adding or migrating indices in a structured way, which isn’t really possible with your proposal. Once created they stay if we don't add additional logic. For me this comes down to whether we want to stick with the ORM-like abstraction in the core, or move toward a leaner, purely functional approach.

I don’t have a huge preference either way. I do plan to work on some schema migration in the future, which might benefit from the ORM-style complexity in general, but you’re right that it’s not urgent right now. How do you see this in the bigger picture?

snejus:
See this blog post for more context regarding flexible attributes: https://beets.io/blog/flexattr.html. Certain implementation details can potentially be optimised, but the core concept is going to stay, I think.

> How do you see this in the bigger picture?

I don't think I see how this relates to the bigger picture. I'm aware of the requirement:

 **we need to index `items.album_id` to optimise the lookup performance**.

And I want to see _the **simplest** solution that satisfies this requirement **but** doesn't create any tech debt and is easy to maintain_.

semohr:
> See this blog post for more context regarding flexible attributes: https://beets.io/blog/flexattr.html. Certain implementation details can potentially be optimised, but the core concept is going to stay, I think.

I don’t disagree with having an attribute table in general I think you got that wrong, they’re useful as a key–value store. My concern is just that they shouldn’t be treated differently from “normal” tables in the core. A table is a table, regardless of its content. Especially now that we’ve also introduced indices on the albums table, the differences are even smaller. Treating them more uniformly might allow us to simplify some logic quite a bit. 

> I don't think I see how this relates to the bigger picture. I'm aware of the requirement:

I’m just trying to explore improvements here. For example, we could reuse the index data object on the attribute tables (and similarly in the migration function), or apply the same, or a similar, approach for field migrations in the future. That said, this doesn’t really fit into this PR, so I’m fine taking it one step at a time. If you prefer not to change existing code, that’s fine, but I like to maintain a codebase that I feel confident and happy working with. Adding features is much easier when the groundwork is set properly.

> And I want to see the simplest solution that satisfies this requirement but doesn't create any tech debt and is easy to maintain

If the absolute main goal is just to optimize `items.album_id` lookup performance, then sure, we could hardcode the migration. However, this would reduce flexibility, make testing and reuse harder, and increase maintenance effort, essentially creating technical debt over time. There’s definitely a trade-off here.

snejus:
I'm largely fine with the changes but since this affects one of the key pieces of the code base I think we'd benefit from hearing what @wisp3rwind thinks about it as well.

test/test_dbcore.py
```diff
@@ -784,3 +787,48 @@ def test_no_results(self):
...
+        return Index(name="sample_index", columns=("field_one",))
+
+    def test_index_creation(self, db, sample_index):
+        """Test creating an index and checking its existence."""
+        with db.transaction() as tx:
+            sample_index.recreate(tx, "test")
+            indexes = (
+                db._connection().execute("PRAGMA index_list(test)").fetchall()
+            )
+        assert any(sample_index.name in index for index in indexes)
```
wisp3rwind: Does this test really add much value compared to `test_from_db`?

test/test_dbcore.py
```diff
@@ -784,3 +787,48 @@ def test_no_results(self):
...
+        """Test the hashing and set behavior of the Index class."""
+        index_set = {sample_index}
+        similar_index = Index(name="sample_index", columns=("field_one",))
+
+        assert similar_index in index_set  # Should recognize similar attributes
+
+        different_index = Index(name="other_index", columns=("other_field",))
+        index_set.add(different_index)
+
+        assert len(index_set) == 2  # Should recognize distinct index
```
wisp3rwind: To be really thorough with testing the hash, we might also test indices with differences in just one component like.

```py
Index(name="sample_index", columns=("field_one",))
Index(name="sample_index", columns=("field_two",))
Index(name="sample_index", columns=("field_one", "field_two"))
Index(name="other_index", columns=("field_one",))
```

beets/dbcore/db.py
semohr, line 1278: There might be an argument using the Index class here too 🤔

semohr:
@snejus @wisp3rwind 

I would like to revive this and merge it if we still think the changes are fine. I had another look here and still like the approach. I also run a micro benchmark against using tx.script directly (without the `Index` class) and there does not seem to be any difference.

```
------------------------------------------------------------------------------------------ benchmark: 2 tests -----------------------------------------------------------------------------------------
Name (time in us)                     Min                 Max               Mean            StdDev             Median               IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_index_raw_benchmark          29.5750 (1.0)      117.1070 (1.10)     31.6931 (1.00)     5.5332 (1.05)     30.4920 (1.0)      0.4530 (1.06)     913;2520       31.5526 (1.00)      22167           1
test_index_recreate_benchmark     29.6570 (1.00)     106.1600 (1.0)      31.5888 (1.0)      5.2643 (1.0)      30.5230 (1.00)     0.4260 (1.0)      397;1197       31.6568 (1.0)       11565           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

snejus:
I tested this locally and discovered a performance issue with the current implementation. It queries the database separately for each existing index on the table (N+1 problem):

```sql
$ BEETS_DEBUG=1 beet --plugins= version
sql
└── PRAGMA table_info(items)
sql
└── 
    CREATE TABLE IF NOT EXISTS item_attributes
      (id INTEGER
         PRIMARY KEY,
         entity_id INTEGER, key TEXT, value TEXT, UNIQUE(entity_id, key) ON CONFLICT REPLACE);
    
    
    CREATE INDEX IF NOT EXISTS item_attributes_by_entity ON item_attributes (entity_id);
sql
└── PRAGMA index_list(items)
sql
└── PRAGMA index_info(idx_item_album_id)
sql
└── PRAGMA index_info(items_album_id_idx)
sql
└── PRAGMA index_info(items_by_album)
sql
└── PRAGMA table_info(albums)
sql
└── 
    CREATE TABLE IF NOT EXISTS album_attributes
      (id INTEGER
         PRIMARY KEY,
         entity_id INTEGER, key TEXT, value TEXT, UNIQUE(entity_id, key) ON CONFLICT REPLACE);
    
    
    CREATE INDEX IF NOT EXISTS album_attributes_by_entity ON album_attributes (entity_id);
sql
└── PRAGMA index_list(albums)
sql
└── SELECT items.*
    FROM (items)
    WHERE id = 0
    GROUP BY items.id
sql
└── SELECT *
    FROM item_attributes
    WHERE entity_id IN
        (SELECT id
         FROM
           (SELECT items.*
            FROM (items)
            WHERE id = 0
            GROUP BY items.id))
beets version 2.5.1
Python version 3.10.18
no plugins loaded
```

Notice how it executes `PRAGMA index_list(items)` followed by a separate `PRAGMA index_info(...)` for each existing index. This overhead happens on every beets command invocation.

I apologize for missing this during review - the implementation details weren't immediately obvious. This is a good reminder why I advocate for keeping things simple when our needs are straightforward. Given that we're adding a single index, could we use the simpler approach that avoids these extra queries? SQLite's `CREATE INDEX IF NOT EXISTS` handles the existence check efficiently at the engine level:

```diff
diff --git a/beets/dbcore/db.py b/beets/dbcore/db.py
index 0a3477371..db3430008 100755
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -1088,9 +1088,9 @@ def __init__
 
         # Set up database schema.
         for model_cls in self._models:
-            self._make_table(model_cls._table, model_cls._fields)
+            self._make_table(model_cls)
             self._make_attribute_table(model_cls._flex_table)
-            self._migrate_indices(model_cls._table, model_cls._indices)
+            # self._migrate_indices(model_cls._table, model_cls._indices)
 
     # Primitive access control: connections and transactions.
 
@@ -1217,21 +1217,16 @@ def load_extension
             conn.load_extension(path)
 
     # Schema setup and migration.
-
-    def _make_table(self, table: str, fields: Mapping[str, types.Type]):
+    def _make_table(self, model_cls: type[Model]):
         """Set up the schema of the database. `fields` is a mapping
         from field names to `Type`s. Columns are added if necessary.
         """
+        table, fields = model_cls._table, model_cls._fields
         # Get current schema.
         with self.transaction() as tx:
             rows = tx.query(f"PRAGMA table_info({table})")
         current_fields = {row[1] for row in rows}
 
-        field_names = set(fields.keys())
-        if current_fields.issuperset(field_names):
-            # Table exists and has all the required columns.
-            return
-
         if not current_fields:
             # No table exists.
             columns = []
@@ -1249,8 +1244,16 @@ def _make_table
                     f"ALTER TABLE {table} ADD COLUMN {name} {typ.sql};\n"
                 )
 
-        with self.transaction() as tx:
-            tx.script(setup_sql)
+        create_idx_cmds = [
+            f"CREATE INDEX IF NOT EXISTS {i.name} ON {table} ({', '.join(i.columns)});"
+            for i in model_cls._indices
+        ]
+
+        setup_sql += "\n".join(create_idx_cmds)
+
+        if setup_sql:
+            with self.transaction() as tx:
+                tx.script(setup_sql)
 
     def _make_attribute_table(self, flex_table: str):
         """Create a table and associated index for flexible attributes
```

This eliminates the PRAGMA queries entirely:

```sql
$ BEETS_DEBUG=1 beet --plugins= version
sql
└── PRAGMA table_info(items)
sql
└── CREATE INDEX IF NOT EXISTS idx_item_album_id ON items (album_id);
sql
└── 
    CREATE TABLE IF NOT EXISTS item_attributes
      (id INTEGER
         PRIMARY KEY,
         entity_id INTEGER, key TEXT, value TEXT, UNIQUE(entity_id, key) ON CONFLICT REPLACE);
    
    
    CREATE INDEX IF NOT EXISTS item_attributes_by_entity ON item_attributes (entity_id);
sql
└── PRAGMA table_info(albums)
sql
└── 
    CREATE TABLE IF NOT EXISTS album_attributes
      (id INTEGER
         PRIMARY KEY,
         entity_id INTEGER, key TEXT, value TEXT, UNIQUE(entity_id, key) ON CONFLICT REPLACE);
    
    
    CREATE INDEX IF NOT EXISTS album_attributes_by_entity ON album_attributes (entity_id);
sql
└── SELECT items.*
    FROM (items)
    WHERE id = 0
    GROUP BY items.id
sql
└── SELECT *
    FROM item_attributes
    WHERE entity_id IN
        (SELECT id
         FROM
           (SELECT items.*
            FROM (items)
            WHERE id = 0
            GROUP BY items.id))
beets version 2.5.1
Python version 3.10.18
no plugins loaded
```

This approach also maintains consistency with how we already handle indices for the attribute tables. We can always revisit the abstraction layer if we need more sophisticated migration capabilities in the future, but for now this solves the immediate performance need without the overhead.

semohr:
I was well aware of this and it was _intended_! It is a **migration** system by design, we need to get the existing indices to check if we need to recreate them if they changed. This is the same approach we take for table migration so it is consistent with the current implementations.

The PRAGMA index_list and index_info have essentially no overhead as they just retrieve the schema metadata. On my machine the overhead here it is negligible (~2us per table).

Also the number of indices is limited, thus it is not really this much of a scaling issue as you made it out to be here...

snejus:
I appreciate the clarification on the intent. A few thoughts:

**On the migration system design:**

I understand this mirrors the table migration approach, and I can see the code does drop and recreate indices when they differ from what's defined. However, there's a key difference in how we'll actually use this: column definitions change over time (new fields get added, types evolve), but the indices we're defining have unique, stable names. Once `idx_item_album_id` is created, it won't change - we're not going to redefine it with different columns or rename it. 

So while the migration capability exists, we're unlikely to use it in practice. Indices aren't like columns where we need to detect and handle schema evolution - we define them once with unique names and they stay that way. This means we're paying the query cost (PRAGMA index_list + N × PRAGMA index_info) on every beets invocation for migration detection we won't actually need.

**On the performance impact:**

You're right that individual PRAGMA calls are fast. My concern isn't about absolute numbers but about principle: we're doing work (N+1 queries) that `CREATE INDEX IF NOT EXISTS` does internally in one step. It's not about catastrophic slowdowns - it's about avoiding unnecessary work when SQLite already solves this for us.

**On consistency:**

I actually see this as an opportunity to improve consistency. The flexible attribute tables already use `CREATE INDEX IF NOT EXISTS` directly. Rather than making the main tables more complex to match the column migration pattern, we could align with the simpler pattern we already use successfully.

**My ask:**

Given that:
- Indices have unique names and won't need migration once created
- `CREATE INDEX IF NOT EXISTS` is simpler and used elsewhere in the codebase  
- The immediate goal is adding one index to improve query performance

Could we start with the simpler approach and add migration infrastructure if/when we have concrete migration requirements? I'm genuinely trying to avoid premature complexity, not dismiss the value of your work.

semohr:
> Could we start with the simpler approach and add migration infrastructure if/when we have concrete migration requirements? I'm genuinely trying to avoid premature complexity, not dismiss the value of your work.

This is a longterm goal with beets for me. I want to unify naming in the database and properly normalize it. This will also require changes to the indices. There are also some alternative valid approaches to a migration systems but I will require this in some form!

My concern is mostly about workflow and reviewability. If I drop a massive overhaul PR with thousands of line changes and additions in the core database module, nobody will realistically be able to review it. I’m trying hard to avoid that situation by breaking things into small, comprehensible steps. To do that, _I need a way to land incremental additions_ even if the features are not fully used yet, otherwise the only alternative is a huge “all-or-nothing” patch that people won’t want to look at. We/I need a way to make these additions without it being dismissed as premature complexity...

snejus:
> If I drop a massive overhaul PR with thousands of line changes and additions in the core database module, nobody will realistically be able to review it.

I get the concern, but what specifically requires thousands of lines? When you're actually doing normalization work, you'll have concrete requirements in front of you - "rename column X and update the index" or "change index columns." That's when you'll know what migration logic you need. Right now you're building a general-purpose detection system without knowing what you're actually migrating.

> I need a way to land incremental additions even if the features are not fully used yet

I'm with you on incremental work. The thing is, there's a difference between landing foundational pieces and landing speculative infrastructure. The `Index` class and `_indices` attributes? Those are solid. But `_migrate_indices()` with PRAGMA queries on every startup - that's assuming indices will change after they're defined. I don't think that's a safe assumption.

> This will also require changes to the indices.

What does that actually look like? Are you renaming indices, changing their columns, creating new ones alongside old ones? I need to understand the concrete migration scenario here, not the abstract "we might need this someday."

I'm keen to get this merged - the index on `album_id` is needed. I just want to make sure we're not adding overhead we don't need yet.

---

**What I'd propose:**

1. Land the `Index` class and declarative `_indices` - that's the foundation
2. Keep it simple with `CREATE INDEX IF NOT EXISTS` for now
3. When you start the normalization work and actually need index migration, add the detection logic then

You get the foundation you need, we avoid ~100 lines of unused code, and when you build the migration stuff, it's informed by real requirements instead of guessing.

```diff
diff --git a/beets/dbcore/__init__.py b/beets/dbcore/__init__.py
index fa20eb00d..0b5e700cb 100644
--- a/beets/dbcore/__init__.py
+++ b/beets/dbcore/__init__.py
@@ -18,3 +18,3 @@
 
-from .db import Database, Model, Results
+from .db import Database, Index, Model, Results
 from .query import (
@@ -45,2 +45,3 @@
     "Type",
+    "Index",
     "parse_sorted_query",
diff --git a/beets/dbcore/db.py b/beets/dbcore/db.py
index cc172d0d8..0920065cc 100755
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -37,3 +37,3 @@
 from sqlite3 import Connection, sqlite_version_info
-from typing import TYPE_CHECKING, Any, AnyStr, Generic
+from typing import TYPE_CHECKING, Any, AnyStr, Generic, NamedTuple
 
@@ -308,2 +308,7 @@ class Model
 
+    _indices: Sequence[Index] = ()
+    """A sequence of `Index` objects that describe the indices to be
+    created for this table.
+    """
+
     @cached_classproperty
@@ -1068,2 +1073,3 @@ class Database
             self._make_attribute_table(model_cls._flex_table)
+            self._create_indices(model_cls._table, model_cls._indices)
 
@@ -1245,2 +1251,15 @@ def _make_attribute_table
 
+    def _create_indices(
+        self,
+        table: str,
+        indices: Sequence[Index],
+    ):
+        """Create indices for the given table if they don't exist."""
+        with self.transaction() as tx:
+            for index in indices:
+                tx.script(
+                    f"CREATE INDEX IF NOT EXISTS {index.name} "
+                    f"ON {table} ({', '.join(index.columns)});"
+                )
+
     # Querying.
@@ -1314 +1333,10 @@ def _get
         return self._fetch(model_cls, MatchQuery("id", id)).get()
+
+
+class Index(NamedTuple):
+    """A helper class to represent the index
+    information in the database schema.
+    """
+
+    name: str
+    columns: tuple[str, ...]
diff --git a/beets/library/models.py b/beets/library/models.py
index cbee2a411..f57b4201a 100644
--- a/beets/library/models.py
+++ b/beets/library/models.py
@@ -718,2 +718,3 @@ class Item
     }
+    _indices = (dbcore.Index("idx_item_album_id", ("album_id",)),)
 
diff --git a/test/test_dbcore.py b/test/test_dbcore.py
index 653adf298..88d94f914 100644
--- a/test/test_dbcore.py
+++ b/test/test_dbcore.py
@@ -25,3 +25,3 @@
 from beets import dbcore
-from beets.dbcore.db import DBCustomFunctionError
+from beets.dbcore.db import DBCustomFunctionError, Index
 from beets.library import LibModel
@@ -68,2 +68,3 @@ class ModelFixture1
     }
+    _indices = (Index("field_one_index", ("field_one",)),)
 
@@ -139,2 +140,3 @@ class AnotherModelFixture
     }
+    _indices = (Index("another_foo_index", ("foo",)),)
 
@@ -240,2 +242,10 @@ def test_extra_model_adds_table
 
+    def test_index_creation(self):
+        """Test that declared indices are created on database initialization."""
+        db = DatabaseFixture1(":memory:")
+        with db.transaction() as tx:
+            rows = tx.query("PRAGMA index_info(field_one_index)")
+            assert len(rows) > 0  # Index exists
+        db._connection().close()
+
```

snejus (approved):

PR #5927, opened 2025-08-11
fish: complete files in more places

9999years:
## Description

Normally, the Fish completion plugin won't complete filenames, which is useful for `beet import` and similar! This removes the `-f` (no filename completion) flag from various places in the output.

## To Do


snejus (approved):
Nice - looks good. Just resolve the conflict in the changelog and we're good to go!

9999years:
@snejus Fixed the merge conflict. In the future, it might be worth having a system where PRs add release notes in different files to a common directory, and then we have a script that gets run when we release which combines all those files into a single changelog, to avoid merge conflicts.

PR #5943, opened 2025-08-18
feat(FtInTitle): Support tracks by artists != album artist

treyturner:
## Description

I have different expectations around this [FtInTitle test case](https://github.com/beetbox/beets/blob/fcc9341360255f7a85e9a55341ab6a6c80083788/test/plugins/test_ftintitle.py#L147-L151), which represents a song on an album by a **guest artist featuring a third artist**:

```
{
    "artist": "Alice ft. Carol",
    "album_artist": "Bob",
    "feat_part": None,
},
```

If this were Alice's album, the plugin would process the track, moving `Carol` to the `feat_part` and changing the `artist` to `Alice`. But because it's Bob's album, nothing happens. I don't want `Alice ft. Carol` as an artist on Bob's album any more than I want it on Alice's though 😭

This may be a bit of a corner case but it does happen reasonably often in the wild:

```
Album: Flying Lotus - ASH (OST) (2025)
Track 26: Kuedo feat. Miguel Atwood-Ferguson - WHAT'S WRONG PEACH？
```

More commonly, this case applies to __compilation albums which use `Various Artists` as an albumartist__.

Processing doesn't occur on these tracks in the current implementation because of how [`find_feat_part()`](https://github.com/beetbox/beets/blob/fcc9341360255f7a85e9a55341ab6a6c80083788/beetsplug/ftintitle.py#L57-L82) kicks off:

```
# Look for the album artist in the artist field. If it's not present, give up.
albumartist_split = artist.split(albumartist, 1)
if len(albumartist_split) <= 1:
    return None
```

As best I can tell, the code is setup this way to enable parsing of more complex cases , like a song by `Hall & Oates` on an Oates album, which is pretty clever. But giving up in cases where albumartist isn't in the artist field seems premature when a reasonable parsing method exists that doesn't require it.

So this PR proposes modifications to code and tests that enables FtInTitle to process items whose artist doesn't contain the albumartist. It seems like what someone loading a `FtInTitle` plugin would want by default, but if that's a contentious take then I'm happy to put it behind a config flag.

Thanks for your consideration!

## To Do

- [ ] ~Documentation.~
  - This case isn't specifically addressed; to me it seems like it would make an intuitive default, but I can either address the case if desired or will  document the config property if one is requested.
- [x] Changelog
- [x] Tests
  - Modified + expanded


whyman:
Dropping a comment here to say this is great and would a nice addition.

 Its coded pretty defensively too and has nice test coverage, shame its been sat here for weeks.

test/plugins/test_ftintitle.py
```diff
@@ -40,19 +40,69 @@ def _ft_set_config(
         self.config["ftintitle"]["auto"] = auto
         self.config["ftintitle"]["keep_in_artist"] = keep_in_artist
 
+    def test_functional_no_featured_artist(self):
```
semohr: The added tests are all quite similar. You should consider using `@pytest.mark.parameterize`. This makes tests more maintainable imo.

semohr:
@whyman Even though `beets` may seem like a big project but the actual active maintainers can be counted on one hand. For us all this is a side project and time is sadly limited. 

That said, we’re always happy to welcome new contributors! If you’re interested, please reach out, we’re especially looking for people who can help with maintenance tasks like triaging issues and reviewing PRs.

Delays in reviews, especially for non-core code, are sadly the norm. :cry:

whyman:
@semohr firstly thanks for your efforts, beets is awesome. 

Apologies for negative vibes there, I've been on the recieving end of such commentary and it's not fun, especially when everyone is here as a hobby, so not cool for me to post that.

treyturner:
> Wouldn't it be reasonable to just populate the `artists` field with `[Alice, Carol]`. In my opinion multiple artists are reasonably well supported in beets today.

Only speaking for my own interests here, but it bothers me when media players that automatically build artist lists (like many automotive entertainment systems for example) will include unique pairings with guest artists as a top-level artist. These often contain only one track and make navigating the artist list cumbersome, especially when you're forced to pause every X entries to monitor road conditions 😅

treyturner:
Ready for another look whenever you find some time @semohr ! 🙏 🚀

semohr (approved):
Thank you very much for improving the test! They are way more readable now.

PR #5948, opened 2025-08-20
Add stripfeat plugin

ARTINKEL:
## Description

Adds new plugin ``stripfeat``, which allows the user to automatically replace the featuring artist token in the artist metadata tag with a delimiter of their choice.

This is particularly useful when using beets in conjunction with something like Jellyfin, which expects multiple artists to be formatted in a certain way, and will create separate artists for every feature combination otherwise.

**Example**

A song with artist metadata ``underscores feat. Jane Remover`` would become ``underscores;Jane Remover``.

## To Do

### DIFF
--- a/beetsplug/stripfeat.py
+++ b/beetsplug/stripfeat.py
@@ -1,0 +1,110 @@
+# This file is part of beets.
+# Copyright 2025, Austin Tinkel, <github.com/ARTINKEL/beets-plugin-stripfeat>
+#
+# Permission is hereby granted, free of charge, to any person obtaining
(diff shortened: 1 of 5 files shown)


whyman:
Your problem is fixed in the next release of jellyfin, where if you enable "prefer ARTISTS" it will use ALBUMARTISTS, which is populated as a multi-value tag and parsed cleanly.

Obviously that doesnt help you with the current release, but FYI.

snejus:
@ARTINKEL is this still relevant or you've upgraded to the new release?

snejus:

PR #5960, opened 2025-08-25
Add configurable search limit to Spotify and Deezer plugins, and make the option name consistent

snejus:
- Add configurable `search_limit` option to Deezer and Spotify plugins (default 5) and enforce it when returning search results.
- Rename `musicbrainz.searchlimit` to `musicbrainz.search_limit` (old key still read with deprecation warning; slated for removal in 3.0.0).
- Update docs and changelog to reflect the changes.

### DIFF
--- a/beetsplug/deezer.py
+++ b/beetsplug/deezer.py
@@ -248,7 +248,10 @@ class DeezerPlugin(SearchApiMetadataSourcePlugin[IDResponse]):
         try:
             response = requests.get(
                 f"{self.search_url}{query_type}",
-                params={"q": query},
+                params={
+                    "q": query,
+                    "limit": self.config["search_limit"].get(),
+                },
                 timeout=10,
             )
             response.raise_for_status()
@@ -266,7 +269,7 @@ class DeezerPlugin(SearchApiMetadataSourcePlugin[IDResponse]):
             self,
             query,
         )
-        return response_data[: self.config["search_limit"].get()]
+        return response_data
 
     def deezerupdate(self, items: Sequence[Item], write: bool):
         """Obtain rank information from Deezer."""
--- a/beetsplug/musicbrainz.py
+++ b/beetsplug/musicbrainz.py
@@ -809,7 +809,7 @@ class MusicBrainzPlugin(MetadataSourcePlugin):
         )
         try:
             method = getattr(musicbrainzngs, f"search_{query_type}s")
-            res = method(limit=self.config["search_limit"].get(int), **filters)
+            res = method(limit=self.config["search_limit"].get(), **filters)
         except musicbrainzngs.MusicBrainzError as exc:
             raise MusicBrainzAPIError(
                 exc, f"{query_type} search", filters, traceback.format_exc()
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -442,7 +442,11 @@ class SpotifyPlugin(
             response = self._handle_response(
                 "get",
                 self.search_url,
-                params={"q": query, "type": query_type},
+                params={
+                    "q": query,
+                    "type": query_type,
+                    "limit": self.config["search_limit"].get(),
+                },
             )
         except APIError as e:
             self._log.debug("Spotify API error: {}", e)
@@ -454,7 +458,7 @@ class SpotifyPlugin(
             self,
             query,
         )
-        return response_data[: self.config["search_limit"].get()]
+        return response_data
 
     def commands(self) -> list[ui.Subcommand]:
         # autotagger import command
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -249,7 +249,7 @@ class DiscogsPlugin(MetadataSourcePlugin):
 
         try:
             results = self.discogs_client.search(query, type="release")
-            results.per_page = self.config["search_limit"].as_number()
+            results.per_page = self.config["search_limit"].get()
             releases = results.page(1)
         except CONNECTION_ERRORS:
             self._log.debug(


beetsplug/musicbrainz.py
```diff
@@ -385,6 +387,13 @@ def __init__(self):
                 "extra_tags": [],
             },
         )
+        # TODO: Remove in 3.0.0
+        with suppress(NotFoundError):
+            self.config["search_limit"] = config["match"]["searchlimit"].get()
```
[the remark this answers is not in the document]
snejus: Good spot, thanks.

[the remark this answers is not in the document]
snejus: On the other hand, you autocompleted this yourself earlier 🙄

semohr:
Maybe we want to add this to the MetadataPlugin abc? Seems like logic for how many candidates are returned should be handled by every metadata plugin. Does it make a difference from an api calling perspective for the deezer and spotify endpoint i.e. do we use less api calls if we limit the returned candidate? 

Alternatively, we could also have one global config for this e.g. `candidates_per_metadata_plugin` and early quite the iteration the plugin core logic?

JOJ0:
> Maybe we want to add this to the MetadataPlugin abc? Seems like logic for how many candidates are returned should be handled by every metadata plugin. Does it make a difference from an api calling perspective for the deezer and spotify endpoint i.e. do we use less api calls if we limit the returned candidate?

It doesn't look like it in this implementation - full results are fetched and then simply cut off. So there shouldn't be an actual difference in performance.

> 
> Alternatively, we could also have one global config for this e.g. `candidates_per_metadata_plugin` and early quite the iteration the plugin core logic?

Do you mean that a global option sets all source plugins in use to one value always, eg sets a limit of 5 to all of them always? This wouldn't be a good idea in my opinion since I often had the feeling that some sources benefit from having more results available to the user than others. So individual config possiblity per source plugin should be kept I think.

I think we should merge this PR as-is since most of all it is a streamlining of options. Moving some to metadata source plugin after that. And you are the expert for that @semohr :-)

JOJ0:
> > Maybe we want to add this to the MetadataPlugin abc? Seems like logic for how many candidates are returned should be handled by every metadata plugin. Does it make a difference from an api calling perspective for the deezer and spotify endpoint i.e. do we use less api calls if we limit the returned candidate?
> 
> It doesn't look like it in this implementation - full results are fetched and then simply cut off. So there shouldn't be an actual difference in performance.
> 

Now i was curious and investigated how for example the Discogs plugin handles that seach_limit option. In that case it _does_ limit the api calls in the first place since that `per_page` option in the underlying library handles that. In case you're interested it generates a URL that only requests that much for a page and we only request one page (first page) in our plugin code:

https://github.com/joalla/discogs_client/blob/e3bd52998023a4f0624e9a28e788c445dda76370/discogs_client/models.py#L345-L351

https://github.com/beetbox/beets/blob/6c21482b7aa293808f0222472da310c311e03b47/beetsplug/discogs.py#L251-L255

So handling this is on a per source plugin basis I think and what this PR does is ok for now but if the libs we use support it, it could be improved. Too lazy right now to look into what deezer and spotify libs (if we use any) support.

Let's merge this and keep an issue open as a reminder? What do you all think?

semohr:
> Do you mean that a global option sets all source plugins in use to one value always, eg sets a limit of 5 to all of them always? This wouldn't be a good idea in my opinion since I often had the feeling that some sources benefit from having more results available to the user than others. So individual config possiblity per source plugin should be kept I think.

We would still have distinct options for each plugin but would handle the configuration option/logic more globally and generic. For example we use the same approach for the `source_weight` config option. 
https://github.com/beetbox/beets/blob/6c21482b7aa293808f0222472da310c311e03b47/beets/metadata_plugins.py#L151

> Let's merge this and keep an issue open as a reminder? What do you all think?

Agreed

beetsplug/spotify.py
```diff
@@ -480,7 +481,7 @@ def _search_api(
             self.data_source,
             query,
         )
-        return response_data
+        return response_data[: self.config["search_limit"].get()]
```
semohr: The spotify api has a limit option, which we might want to use instead:

https://developer.spotify.com/documentation/web-api/reference/search

beetsplug/deezer.py
```diff
@@ -283,7 +287,7 @@ def _search_api(
             self.data_source,
             query,
         )
-        return response_data
+        return response_data[: self.config["search_limit"].get()]
```
semohr: Deezer’s API uses pagination. If you set a higher search limit, some results may still be skipped and not appear. Currently, we do not yet handle pagination in our implementation but maybe it is time to think about this here?

e.g. https://developers.deezer.com/api/explorer?url=search%3Fq%3Deminem

beetsplug/spotify.py
JOJ0: opened separate issue

beetsplug/deezer.py
JOJ0: opened separate issue.

JOJ0:
> We would still have distinct options for each plugin but would handle the configuration option/logic more globally and generic. For example we use the same approach for the `source_weight` config option.
> 
> https://github.com/beetbox/beets/blob/6c21482b7aa293808f0222472da310c311e03b47/beets/metadata_plugins.py#L151
> 
> > Let's merge this and keep an issue open as a reminder? What do you all think?
> 
> Agreed

I opened that reminder issue just now. But would adding a simple line like this already be a win in that abstract class (and removing it from the plugins in this PR)? 
```
        self.config.add({"search_limit": 5})
```

Then we could as well add it already. Or should there be more to it in the first place?

JOJ0:
Sorry for hijacking @snejus , I fixed that conflict in spotify.py and force-pushed. Hope you don't mind 😬

beetsplug/spotify.py
snejus: This is done now

beetsplug/deezer.py
snejus: Also done

snejus:
* Moved `search_limit` configuration option initialisation under the common `MetadataSourcePlugin` interface
* Following @semohr suggestions `spotify` and `deezer` results are now limited using the request parameters.

snejus:
Relatedly, I started work on centralising the entire search functionality in these autotaggers: https://github.com/beetbox/beets/pull/5982

PR #5961, opened 2025-08-25
Add CODEOWNERS file

snejus:
Add CODEOWNERS file assigning all files (*) to [`@beetbox/maintainers`](https://github.com/orgs/beetbox/teams/maintainers) so that team is auto-requested for reviews.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# assign the entire repo to the maintainers team
+* @beetbox/maintainers
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -180,8 +180,7 @@ Your First Contribution
 If this is your first time contributing to an open source project, welcome! If
 you are confused at all about how to contribute or what to contribute, take a
 look at `this great tutorial <http://makeapullrequest.com/>`__, or stop by our
-`discussion board <https://github.com/beetbox/beets/discussions/>`__ if you have
-any questions.
+`discussion board`_ if you have any questions.
 
 We maintain a list of issues we reserved for those new to open source labeled
 `first timers only`_. Since the goal of these issues is to get users comfortable
@@ -215,6 +214,15 @@ will ship in no time.
 
 Remember, code contributions have four parts: the code, the tests, the
 documentation, and the changelog entry. Thank you for contributing!
+
+.. admonition:: Ownership
+
+    If you are the owner of a plugin, please consider reviewing pull requests
+    that affect your plugin. If you are not the owner of a plugin, please
+    consider becoming one! You can do so by adding an entry to
+    ``.github/CODEOWNERS``. This way, you will automatically receive a review
+    request for pull requests that adjust the code that you own. If you have any
+    questions, please ask on our `discussion board`_.
 
 The Code
 --------
@@ -393,6 +401,8 @@ In order to add such a test, mark your test with the ``integration_test`` marker
 This way, the test will be run only in the integration test suite.
 
 .. _codecov: https://codecov.io/github/beetbox/beets
+
+.. _discussion board: https://github.com/beetbox/beets/discussions
 
 .. _documentation: https://beets.readthedocs.io/en/stable/
 


semohr:
Seems like a good way to define responsibilities in the codebase! For anyone else who is a bit confused on what this even is, check out the [github docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners?apiVersion=2022-11-28).

jackwilsdon:
Are we best putting this in `.github` to keep it out of the way? It seems like it can be in `.github/`, root or `docs/`: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-file-location

snejus:
> Are we best putting this in `.github` to keep it out of the way? It seems like it can be in `.github/`, root or `docs/`: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-file-location

Just had a glance at GitHub docs: isn't it ironic that Copilot suggestions contradict GitHub docs 😁 

Will bring this back under the `docs` directory. I followed [this guide](https://www.aviator.co/blog/a-modern-guide-to-codeowners/#).

semohr:
Should we add a note to the docs reminding contributors that when they introduce a new plugin, they should also add themselves (or their team) to the CODEOWNERS file so they’re responsible for maintaining it?

jackwilsdon:
Sounds like a good idea. I still think this probably belongs in `.github` instead of `docs` as it's a GitHub-specific file and isn't actually rendered into the docs.

JOJ0:
> Sounds like a good idea. I still think this probably belongs in `.github` instead of `docs` as it's a GitHub-specific file and isn't actually rendered into the docs.

We need a final decision here. The guide snejus posted points out that docs/ applies to github and gitlab as well (even though beets might never move there it's a common location for it)

It's not rendered in the docs, that is true, but we _could_ actually add it and handle it as a documentation of sorts: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude

In any case a few lines the contribution/developer docs I too think is a very good idea. If plugin contributors mention themselves there and get automatically notified this would be a big win - it will take time until this file is populated (unless we ask people "can we add you and auto-bother you since you contributed this ages ago" ;-)

snejus:
> Should we add a note to the docs reminding contributors that when they introduce a new plugin, they should also add themselves (or their team) to the CODEOWNERS file so they’re responsible for maintaining it?

I see where you're coming from - it might be a bit optimistic to expect contributors to come back and maintain what they've added. In reality, once a plugin is merged, it's usually on us maintainers to take care of it long term. Adding people to CODEOWNERS is still helpful to show who originally worked on it, but I wouldn't rely on that as a real hand-off of responsibility.

On the other hand, I think we should consider assigning ourselves to the parts of the codebase
that each of us has expertise on. For example, just off the top of my head:

* `beetsplug/lastgenre` owned by @JOJ0 
* `beets/metadata_plugins.py` owned by @semohr 
* `beetsplug/lyrics.py` owned by me etc.

We could even analyse git blame to determine modules that have mostly been contributed by a single person.

semohr:
> I see where you're coming from - it might be a bit optimistic to expect contributors to come back and maintain what they've added. In reality, once a plugin is merged, it's usually on us maintainers to take care of it long term. Adding people to CODEOWNERS is still helpful to show who originally worked on it, but I wouldn't rely on that as a real hand-off of responsibility.

I’m not expecting new contributors to take on full responsibility for a plugin long-term. The idea is just to add a small note in the docs suggesting that contributors can add their name to the CODEOWNERS file if they’d like to stay in the loop. Totally optional, but might encourage some people to stick around as longer-term contributors. 

We should still get updates as maintainers regardless.

snejus:
I've added a note about ownership to `CONTRIBUTING.rst`.

snejus:
> The idea is just to add a small note in the docs suggesting that contributors can add their name to the CODEOWNERS file if they’d like to stay in the loop. Totally optional, but might encourage some people to stick around as longer-term contributors.

I see what you mean - it's a great idea. I've added a note to `CONTRIBUTING.rst`:
image

Issue #5903, opened 2025-07-29
Network errors abort import process


9999years:
### Problem

If `beets` encounters a (transient) network error, the entire import process is aborted. It would be nice if `beets` instead skipped the particular files that failed and continued with the import.

When running `beet import ~/Music/Music/ --quiet --quiet-fallback=skip --incremental`:

```
2025-07-29 00:35:14 musicbrainzngs.musicbrainz DEBUG: GET request for https://musicbrainz.org/ws/2/release/f11c115e-afe8-45ab-a94f-91cc92656fe4?inc=isrcs+release-groups+artist-credits+release-rels+artist-rels+tags+aliases+work-level-rels+recording-level-rels+artists+work-rels+url-rels+media+labels+recordings
2025-07-29 00:50:44 beets ERROR: Error: MusicBrainz not reachable in get release by ID with query 'f11c115e-afe8-45ab-a94f-91cc92656fe4'
2025-07-29 00:50:44 beets DEBUG: Traceback (most recent call last):
  File "/nix/store/f1jvr1vx9mwnj5sp7azx9mf0y52jcws5-python3.13-musicbrainzngs-0.7.1/lib/python3.13/site-packages/musicbrainzngs/musicbrainz.py", line 501, in _safe_read
    f = opener.open(req, timeout=timeout)
  ...
ConnectionResetError: [Errno 54] Connection reset by peer

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/nix/store/lfv9ns20hz2bg6d44js378vcxjfm9261-beets-2.3.1/lib/python3.13/site-packages/beetsplug/musicbrainz.py", line 845, in album_for_id
    res = musicbrainzngs.get_release_by_id(albumid, RELEASE_INCLUDES)
  ...
  File "/nix/store/f1jvr1vx9mwnj5sp7azx9mf0y52jcws5-python3.13-musicbrainzngs-0.7.1/lib/python3.13/site-packages/musicbrainzngs/musicbrainz.py", line 536, in _safe_read
    raise NetworkError(cause=exc)
musicbrainzngs.musicbrainz.NetworkError: caused by: [Errno 54] Connection reset by peer
```

<details><summary>Debug logs / stack trace</summary>

```
[...]

Issue #5553, opened 2024-12-18
MusicBrainz connection network error handling differences between beets version 1.6 and 2+


jellybuns0x:
I've been trying to get to the bottom of a network issue for the last few weeks while importing that has been preventing me from being able to successfully run through a few thousand albums. While I cannot solve the underlying issue to prevent this error from happening, it seems the way beets handles this error has changed since 1.6 [NetworkError from musicbrainzng](https://github.com/alastair/python-musicbrainzngs/blob/v0.7.1/musicbrainzngs/musicbrainz.py#L520C1-L525C33). 

In 1.6 the import process hangs for a while, the hung thread eventually fails gracefully, an exception is logged, a new set of threads are spawned and the import process continues. In 2.2 I've noticed when this same exception occurs, the beets process is killed completely. Both installs of beets I'm using to test this are on 1.7.1 of musibrainzng bindings. I've come across this problem in both the latest linuxserver.io docker image and a manual setup on a debian vm. Behavior is also the same regardless of threaded value set yes/no.

### Problem

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv import /downloads
```

Led to this problem:

```
Sending event: import_task_created
Traceback (most recent call last):
  File "/usr/lib/python3.11/urllib/request.py", line 1348, in do_open
    h.request(req.get_method(), req.selector, req.data, headers,
  File "/usr/lib/python3.11/http/client.py", line 1282, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/usr/lib/python3.11/http/client.py", line 1328, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.11/http/client.py", line 1277, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/usr/lib/python3.11/http/client.py", line 1037, in _send_output
    self.send(msg)
  File "/usr/lib/python3.11/http/client.py", line 975, in send
    self.connect()
  File "/usr/lib/python3.11/http/client.py", line 1454, in connect
    self.sock = self._context.wrap_socket(self.sock,
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/ssl.py", line 517, in wrap_socket
    return self.sslsocket_class._create(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #4789, opened 2023-05-12
Fatal Spotify plugin error - Bad gateway


sleroq:
When importing a lot of music with Spotify plugin, sometimes I get a "Bad gateway" error which is fatal. With quiet imports I can't resume the import and have to start again.

It would be great if this error could be made non-fatal.
Maybe that the song could be retried later?

### Problem

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv import -q -s folder
```

Led to this problem:

```
spotify: Searching Spotify for '[Blue Archive] Theme 99 (Onsen/Hotspring Event Lobby BGM) artist:Salieri AS Beta'
Sending event: import_task_created
error: Spotify API error:
{
  "error" : {
    "status" : 502,
    "message" : "Bad gateway."
  }
}
URL:
https://api.spotify.com/v1/search
params:
{'q': '[Blue Archive] Theme 99 (Onsen/Hotspring Event Lobby BGM) artist:Salieri AS Beta', 'type': 'track'}
```

Here's a link to the music files that trigger the bug (if relevant):

### Setup

* OS: NixOS 23.05pre481752.897876e4c48 (Stoat) x86_64
* Python version: Python 3.10.11
* beets version: beets version 1.6.1
* Turning off plugins made problem go away (yes/no): I mean.. yes

My configuration (output of `beet config`) is:

```yaml
lyrics:
    bing_lang_from: []
    sources: genius tekstowo
    auto: yes
    bing_client_secret: REDACTED
    bing_lang_to:
    google_API_key: REDACTED
    google_engine_ID: REDACTED
    genius_api_key: REDACTED
    fallback:
    force: no
    local: no
    dist_thresh: 0.1
directory: ~/Sync/storj/music
library: ~/Music/storj-music-library.db

import:
    move: yes

plugins: discogs spotify fetchart lyrics lastgenre web
discogs:
    user_token: REDACTED
    apikey: REDACTED
    apisecret: REDACTED
    tokenfile: discogs_token.json
    source_weight: 0.5
    separator: ', '
    index_tracks: no
    append_style_genre: no
spotify:
    source_weight: 0.5
    mode: list
    tiebreak: popularity
    show_failures: no
    artist_field: albumartist
    album_field: album
    track_field: title
    region_filter:
    regex: []
    client_id: 4e414367a1d14c75a5c5129a627fcab8
    client_secret: REDACTED
    tokenfile: spotify_token.json
lastgenre:
    whitelist: yes
[...]

PR #5965, opened 2025-08-26
Safely handle metadata plugin exceptions.

semohr:
## Description

When a metadata plugin raises an exception during the auto-tagger process, the entire operation crashes. This behavior is not desirable, since metadata lookups can legitimately fail for various reasons (e.g., temporary API downtime, network issues, or offline usage).

This PR introduces a safeguard by adding general exception handling around metadata plugin calls. Instead of causing the whole process to fail, exceptions from individual plugins are now caught and logged. This ensures that the auto-tagger continues to function with the remaining available metadata sources. I used a proxy pattern here as this
seems like an elegant solution to me.

This replaces the efforts from #5910

## Decisions needed

How do we want to name the configuration option which controls if an error should be propagated and also where/how do we want this to be documented?

## Todos:
- [x] Changelog.
- [ ] Documetnation.

### DIFF
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -8,18 +8,15 @@ implemented as plugins.
 from __future__ import annotations
 
 import abc
-import inspect
 import re
-from functools import cache, cached_property, wraps
+from functools import cache, cached_property
 from typing import (
     TYPE_CHECKING,
     Callable,
-    ClassVar,
     Generic,
     Literal,
     TypedDict,
     TypeVar,
-    overload,
 )
 
 import unidecode
@@ -396,73 +393,28 @@ class SafeProxy(base):
     """
 
     _plugin: MetadataSourcePlugin
-    _SAFE_METHODS: ClassVar[set[str]] = {
-        "candidates",
-        "item_candidates",
-        "album_for_id",
-        "track_for_id",
-    }
 
     def __init__(self, plugin: MetadataSourcePlugin):
         self._plugin = plugin
 
     def __getattribute__(self, name):
-        if (
-            name == "_plugin"
-            or name == "_handle_exception"
-            or name == "_SAFE_METHODS"
-            or name == "_safe_execute"
-        ):
+        if name in {
+            "_plugin",
+            "_handle_exception",
+            "candidates",
+            "item_candidates",
+            "album_for_id",
+            "track_for_id",
+        }:
             return super().__getattribute__(name)
-
-        attr = getattr(self._plugin, name)
-
-        if callable(attr) and name in SafeProxy._SAFE_METHODS:
-            return self._safe_execute(attr)
-        return attr
+        else:
+            return getattr(self._plugin, name)
 
     def __setattr__(self, name, value):
         if name == "_plugin":
             super().__setattr__(name, value)
         else:
             self._plugin.__setattr__(name, value)
-
-    @overload
-    def _safe_execute(
-        self,
-        func: Callable[P, Iterable[R]],
-    ) -> Callable[P, Iterable[R]]: ...
-    @overload
-    def _safe_execute(self, func: Callable[P, R]) -> Callable[P, R | None]: ...
-    def _safe_execute(
-        self, func: Callable[P, R]
-    ) -> Callable[P, R | Iterable[R] | None]:
-        """Wrap any function (generator or regular) and safely execute it.
-
-        Limitation: This does not work on properties!
-        """
-
-        @wraps(func)
-        def wrapper(
-            *args: P.args, **kwargs: P.kwargs
-        ) -> R | Iterable[R] | None:
-            try:
-                result = func(*args, **kwargs)
-            except Exception as e:
-                self._handle_exception(func, e)
-
-                return None
-
-            if inspect.isgenerator(result):
-                try:
-                    yield from result
-                except Exception as e:
-                    self._handle_exception(func, e)
-                return None
-            else:
-                return result
-
-        return wrapper
 
     def _handle_exception(self, func: Callable[P, R], e: Exception) -> None:
         """Helper function to log exceptions from metadata source plugins."""
@@ -476,17 +428,26 @@ class SafeProxy(base):
         )
         log.debug("Exception details:", exc_info=True)
 
-    # Implement abstract methods to satisfy the ABC
-    # this is only needed because of the typing hack above.
-
-    def album_for_id(self, album_id: str):
-        raise NotImplementedError
+    def album_for_id(self, *args, **kwargs):
+        try:
+            return self._plugin.album_for_id(*args, **kwargs)
+        except Exception as e:
+            return self._handle_exception(self._plugin.album_for_id, e)
 
     def track_for_id(self, track_id: str):
-        raise NotImplementedError
+        try:
+            return self._plugin.track_for_id(track_id)
+        except Exception as e:
+            return self._handle_exception(self._plugin.track_for_id, e)
 
     def candidates(self, *args, **kwargs):
-        raise NotImplementedError
+        try:
+            yield from self._plugin.candidates(*args, **kwargs)
+        except Exception as e:
+            return self._handle_exception(self._plugin.candidates, e)
 
     def item_candidates(self, *args, **kwargs):
-        raise NotImplementedError
+        try:
+            yield from self._plugin.item_candidates(*args, **kwargs)
+        except Exception as e:
+            return self._handle_exception(self._plugin.item_candidates, e)
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -9,33 +9,26 @@ from __future__ import annotations
 
 import abc
 import re
-from functools import cache, cached_property
-from typing import (
-    TYPE_CHECKING,
-    Callable,
-    Generic,
-    Literal,
-    TypedDict,
-    TypeVar,
-)
+from contextlib import contextmanager, nullcontext
+from functools import cache, cached_property, wraps
+from typing import TYPE_CHECKING, Generic, Literal, TypedDict, TypeVar
 
 import unidecode
 from confuse import NotFoundError
-from typing_extensions import NotRequired, ParamSpec
+from typing_extensions import NotRequired
 
 from beets import config, logging
 from beets.util import cached_classproperty
 from beets.util.id_extractors import extract_release_id
 
-from .plugins import BeetsPlugin, find_plugins, notify_info_yielded, send
+from .plugins import BeetsPlugin, find_plugins, notify_info_yielded
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable, Sequence
+    from collections.abc import Callable, Iterable, Iterator, Sequence
 
     from .autotag.hooks import AlbumInfo, Item, TrackInfo
 
-    P = ParamSpec("P")
-    R = TypeVar("R")
+    Ret = TypeVar("Ret")
 
 # Global logger.
 log = logging.getLogger("beets")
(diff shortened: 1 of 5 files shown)


semohr:
Related:
- #5553 will benefit from this
- #4789 will not crash anymore but be skipped 
- #5903 Network errors will log warning but the import will continue

jackwilsdon:
It does feel like we should pause the import until the user has acknowledged the issue (maybe providing options to skip importing the album or continue with existing metadata?), as otherwise you could end up with missing metadata that is awkward to fix later if for example you had network issues or the provider was down.

Not sure how this should interact with `--quiet` and `--quiet-fallback` though.

semohr:
A prompt makes sense in theory, but I'm really hesitant to add any user interaction because it breaks headless scripts, automation, and make monkey-patching way harder which is a big part of my beets use case. Instead, I'm leaning towards adding a simple config option that lets you tell the plugin to just fail hard if it runs into trouble, so your import would stop and you'd know right away.

On a practical level, I'm also not convinced that skipping these errors is a major issue in the first place. The plugin's job is to suggest `candidates`, and it's already robust in how it handles failure: if it can't find any candidates, it just doesn't add any, and the import continues with other plugins. Same goes for `get_by_id`, it's designed to return None, so the higher-level functions are already built to handle that. Imo the introduced behavior should be good enough for most users.

henry-oberholtzer (approved):
Somehow completely missed this PR when you first put it in - but I've been encountering this exact issue a lot myself and it makes wanting to use batch import really discouraging, especially with the Musicbrainz rate limiting. This is a great fix for it.

semohr:
Im not too happy with the duplicate `_safe_call` and `_safe_yield` functions. I would be happy to get another idea here on how to make this a bit more convenient and maintainable.

 We also need a config option to re-enable to old behavior before I'm comfortable enough to merge this.

semohr:
@henry-oberholtzer I rethought my approach here a bit and opted to use a proxy pattern instead. Seems a bit more clear and maintainable imo.

semohr:
@snejus Would you mind having a look here too? I would be happy to know your thoughts on this. Im pretty happy with the general implementation now and can finalize this if we are happy with the approach.

I like the proxy approach (while a bit complex) it is very unintrusive, only hooking into our existing logic at one place.

henry-oberholtzer:
I'm not the most familiar with proxies, but this looks good to me, same logic, neater application.

beets/metadata_plugins.py
```diff
@@ -26,12 +35,18 @@
...
+# Global logger.
+log = logging.getLogger("beets")
+
 
 @cache
 def find_metadata_source_plugins() -> list[MetadataSourcePlugin]:
     """Return a list of all loaded metadata source plugins."""
     # TODO: Make this an isinstance(MetadataSourcePlugin, ...) check in v3.0.0
-    return [p for p in find_plugins() if hasattr(p, "data_source")]  # type: ignore[misc]
+    return [SafeProxy(p) for p in find_plugins() if hasattr(p, "data_source")]  # type: ignore[misc,arg-type]
```
snejus: I guess this is where we do this selectively given the new configuration value?

beets/metadata_plugins.py
```diff
@@ -357,3 +372,83 @@ def _construct_search_query(
...
+        )
+        log.debug("Exception details:", exc_info=True)
+
+    def album_for_id(self, *args, **kwargs):
+        try:
+            return self.__plugin.album_for_id(*args, **kwargs)
+        except Exception as e:
+            return self.__handle_exception(self.__plugin.album_for_id, e)
+
+    def track_for_id(self, track_id: str):
```
snejus: This needs `*args, **kwargs` for consistency

beets/metadata_plugins.py
```diff
@@ -357,3 +372,83 @@ def _construct_search_query(
...
+    def __setattr__(self, name, value):
+        if name == "_SafeProxy__plugin":
+            super().__setattr__(name, value)
+        else:
+            self.__plugin.__setattr__(name, value)
+
+    def __handle_exception(self, func: Callable[P, R], e: Exception) -> None:
+        """Helper function to log exceptions from metadata source plugins."""
+        if config["raise_on_error"].get(bool):
+            raise e
```
snejus: I think we would preferably handle this configuration option in `find_metadata_source_plugins`. This way, when we have, `raise_on_error: no`, exceptions will be raised _truly_ directly. This way, we also have a fallback in case something goes bad with `SafeProxy` implementation, since we have an option to disable it.

snejus (commented):
That's a very clean way to handle exceptions, nicely done! See my note about at which point should the new configuration come into play here.

beets/metadata_plugins.py
semohr: Sounds reasonable. I was thinking we handle it here incase we want to attach the option to each plugin instead. Would allow for more granual control.

Although, I'm not sure we want to go this route.

snejus: I cannot really see a use case for granular control here.

FYI `beetcamp` already handles its exceptions, where it points users to a URL to submit a GitHub issue when something unexpected happens.

semohr: You are right, plugins should handle their exceptions gracefully! I think this logic here is just a last resort in-case something really unexpected happens.

I moved the config check into the `find_metadata_source_plugins` function.

snejus:
@semohr As far as I remember this was initially implemented using decorator pattern. Why did you switch to the proxy pattern? I'm most likely missing context, but, I think, the decorator pattern would allow this to be type-checked?

semohr:
Typechecking should work either way with the small `mypy` hack.

Here’s a brief recap of my reasoning:

One challenge with using a decorator is that it needs to be applied to each plugin individually. Simply wrapping a main function like `def album_for_id(...)` isn’t sufficient, because it would prevent other plugins from continuing the auto-tagging process if an error occurs. I initially tried something like:

```python
if info := _safe_call(plugin.album_for_id)(_id):
    ...
```

…but I found this approach somewhat intrusive and not very Pythonic. It is needed in every plugin wrapper call.

Another complication arises from using `yield form`, which requires additional decorator logic to handle the generator (I used `_safe_yield_from` here). This approach would end up duplicating a lot of the decorator code. In theory, we could combine the “yield” and “call” decorators, but that would require `overload` typing and fairly convoluted inspection logic to determine whether the return value supports `yield from` syntax.

Given this, I decided to try a proxy pattern and am a bit more happy with it.

snejus:
> One challenge with using a decorator is that it needs to be applied to each plugin individually

Can we not use the `metadata_plugins.candidates`, `metadata_plugins.item_candidates` etc for this?

semohr:
Im not too sure what you mean. The methods are overwritten by each plugins implementation as they are abstract. You cant apply the decorator directly, you always need to wrap the functions dynamically.

As mentioned in the comment above something like this should work:

```python
def item_candidates(*args,**kwargs):
    For plugin in find_plugins():
         yield from _safe_yield(plugin.item_candidates)(*args, **kwargs)
````

**Comment by snejus** [comment]
I'm thinking something in these lines indeed

```py
def _safe_call(
    plugin: MetadataSourcePlugin,
    method_name: str,
    *args,
    **kwargs,
):
    """Safely call a plugin method, catching and logging exceptions."""
    try:
        method = getattr(plugin, method_name)
        return method(*args, **kwargs)
    except Exception as e:
        log.error(
            "Error in '{}.{}': {}",
            plugin.data_source,
            method_name,
            e,
        )
        log.debug("Exception details:", exc_info=True)
        return None


@notify_info_yielded("albuminfo_received")
def candidates(*args, **kwargs) -> Iterable[AlbumInfo]:
    """Return matching album candidates from all metadata source plugins."""
    for plugin in find_metadata_source_plugins():
        if result := _safe_call(plugin, "candidates", *args, **kwargs):
            yield from result
```

Another question: does `raise_an_error` configuration provide any value? I'd be happy to **always** handle these exceptions.

semohr:
This is mostly how i had it before. This does not catch the errors during the yield and handling this properly requires a bit more work. Also typing is stripped here.

I think retaining the initial behavior is wanted, sometimes one doesn't want to continue if an error occurs as some crucial lookup (for some) might be missing. See also https://github.com/beetbox/beets/pull/5965#issuecomment-3274063030

snejus:
> This is mostly how i had it before. This does not catch the errors during the yield and handling this properly requires a bit more work. Also typing is stripped here.

A pure decorator-based approach may be the way to go forward, then:

```py
@contextmanager
def handle_plugin_error(method_name: str):
    """Safely call a plugin method, catching and logging exceptions."""
    try:
        yield
    except Exception as e:
        log.error("Error in '{}': {}", method_name, e)
        log.debug("Exception details:", exc_info=True)


def safe_yield(func: IterF[P, Ret]) -> IterF[P, Ret]:
    @wraps(func)
    def wrapper(*args: P.args, **kwargs: P.kwargs) -> Iterable[Ret]:
        with handle_plugin_error(func.__qualname__):
            yield from func(*args, **kwargs)

    return wrapper


@notify_info_yielded("albuminfo_received")
@safe_yield
def candidates(*args, **kwargs) -> Iterable[AlbumInfo]:
    """Return matching album candidates from all metadata source plugins."""
    for plugin in find_metadata_source_plugins():
        yield from plugin.candidates(*args, **kwargs)
```

To make sure this decorator is suitable for `album_for_id` and `track_for_id`, consider using `plugin.albums_for_ids` and `plugin.tracks_for_ids` inside those functions. This way, you can also wrap them up `notify_info_yielded` accordingly.



> I think retaining the initial behavior is wanted, sometimes one doesn't want to continue if an error occurs as some crucial lookup (for some) might be missing. See also [#5965 (comment)](https://github.com/beetbox/beets/pull/5965#issuecomment-3274063030)

Presumably, this depends on us _handling_ errors somewhere? Otherwise, right now, an unhandled plugin error kills beets execution.

semohr:
> Presumably, this depends on us handling errors somewhere? Otherwise, right now, an unhandled plugin error kills beets execution.

 This seems like the desired behavior in that case for me. E.g. I want the process to exit completely if musicbrainz is currently down.
 
 > To make sure this decorator is suitable for album_for_id and track_for_id, consider using plugin.albums_for_ids and plugin.tracks_for_ids inside those functions. This way, you can also wrap them up notify_info_yielded accordingly.
 
The issue with that approach is that it does not log the plugin name correctly (here the method_name would be `metada_plugins.candidates`). Imo it is quite important for the error message to include the actual name of the plugin which has issues. I have really tried quite a few approaches here and this will even get more complex ;)

I can try to restore my initial decorator based approach but it is way more complex and more intrusive. I think the proxy based approach is more maintainable.

pSpitzner:
Hi all,

chiming in because Im currently stumbling across this a lot in beets-flask - hitting API timeouts with musicbrainz every third run or so.

I looked through the current PR state and am wondering what the downside of the proxy pattern is?
As it currently stands we get: typing, user-configurable raising, automatically applies to all plugins, correctly logs plugin-name to global logger.

The only downside I could imagine is, that the ProxyClass has to be updated when new abstract methods are added to the MetaPlugins, but I suppose this is in some form also the case with the decorators.

snejus:
> I looked through the current PR state and am wondering what the downside of the proxy pattern is?

- Proxy hides behavior from plugin maintainers; the base class is empty so the error handling is invisible.
- Introspection/debugging regresses: `mbp = MusicBrainzPlugin(); mbp.candidates??` points at `SafeProxy`, not the plugin.
- Changes object identity (`SafeProxy` instead of the real plugin), which can break `isinstance`/`__class__` assumptions.
- Config is cached via `find_metadata_source_plugins`; toggling `raise_on_error` will not take effect until cache clear.
- This feels non-Pythonic vs explicit wrappers, and the typing is not transparent (the "fake subclass" hack is a maintainability smell).

semohr:
> Proxy hides behavior from plugin maintainers; the base class is empty so the error handling is invisible.

The proxy is only ever used in the metadata_plugins.py file. Plugin developers should not interact with it in any way as far as I can see.

> Config is cached via find_metadata_source_plugins; toggling raise_on_error will not take effect until cache clear.

Isn't this the case for all config values in beets? Once they are loaded with confuse, future calls to the same confuse object will always return the value in memory. I think having a live relodable config is out of scope here.

> This feels non-Pythonic vs explicit wrappers, and the typing is not transparent (the "fake subclass" hack is a maintainability smell).

Agreed it does not feel too pythonic. Feelings aside tho, it does work and is pretty nonintrussive which is a big plus from the maintainability side for me.

---

We might want to create a full decorator based implementation to compare in another branch. Im traveling currently but might come around it, I guees latest in a month or so.

snejus:
Decorator-based approach that should satisfy each of our requirements:

```py
@contextmanager
def handle_plugin_error(plugin: MetadataSourcePlugin, method_name: str):
    """Safely call a plugin method, catching and logging exceptions."""
    try:
        yield
    except Exception as e:
        log.error("Error in '{}.{}': {}", plugin.data_source, method_name, e)
        log.debug("Exception details:", exc_info=True)


def _yield_from_plugins(
    func: Callable[..., Iterable[Ret]],
) -> Callable[..., Iterator[Ret]]:
    method_name = func.__name__

    @wraps(func)
    def wrapper(*args, **kwargs) -> Iterator[Ret]:
        for plugin in find_metadata_source_plugins():
            method = getattr(plugin, method_name)
            with handle_plugin_error(plugin, method_name):
                yield from filter(None, method(*args, **kwargs))

    return wrapper


@notify_info_yielded("albuminfo_received")
@_yield_from_plugins
def candidates(*args, **kwargs) -> Iterator[AlbumInfo]:
    yield from ()


@notify_info_yielded("trackinfo_received")
@_yield_from_plugins
def item_candidates(*args, **kwargs) -> Iterator[TrackInfo]:
    yield from ()


@notify_info_yielded("albuminfo_received")
@_yield_from_plugins
def albums_for_ids(*args, **kwargs) -> Iterator[AlbumInfo]:
    yield from ()


@notify_info_yielded("trackinfo_received")
@_yield_from_plugins
def tracks_for_ids(*args, **kwargs) -> Iterator[TrackInfo]:
    yield from ()


def album_for_id(_id: str) -> AlbumInfo | None:
    return next(albums_for_ids([_id]), None)


def track_for_id(_id: str) -> TrackInfo | None:
    return next(tracks_for_ids([_id]), None)
```

semohr:
Looks reasonable :+1: Some details ofc missing but we should be able to adapt that.

snejus:
Going to merge this in as it is now since I'm keen to have this in the new release. We can revise the details a bit later if needed!

snejus (approved):

semohr:
Just seen the change.
I think coupling the names of the functions that tightly with the implementation is pretty convoluted. I get that it minimizes code but it is not really understandable in the long run and I don't think mypy can check this properly. Performance should also be worse compared to calling the functions directly.

We should find another way here imo.

PR #5979, opened 2025-08-30
lastgenre: Move file loading to helpers

JOJ0:
## Description

I split out this refactor from #5744 to make the blacklist PR just a little bit easier to read. ~No code changes in here just moving to helpers!~

- Separate _load methods for whitelist and canonicalization tree file loading.
- Add return type hints for those methods.
- Use `pathlib.Path()` instead of `normpath()`
- Additionally respect None value to fall back to built-in settings for `canonical` and `whitelist` files (was only `""` before) - this code exists for backwards compatibility (see https://github.com/beetbox/beets/commit/233f71a4574b5a10d1182a3a63ce5cd23c4268f4), and might be better sanity checked and vanish in the future... (a None value is set eg. `canoncial:`) -> #5994 

## To Do

- [x] Documentation. (not required)
- [x] Changelog.
- [x] Tests. (already available, nothing changes)

## Summary by Sourcery

Extract file loading logic for the whitelist and canonicalization tree into dedicated helper methods and add return type annotations

Enhancements:
- Move whitelist file parsing into a new _load_whitelist helper and return the set of genres
- Move canonicalization tree loading into a new _load_c14n_tree helper and return branches and canonicalize flag
- Import and use syspath with normpath for consistent file path handling
- Add return type hints for the new helper methods
- Simplify setup by delegating to the new helper methods and adjusting debug formatting

### DIFF
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -120,7 +120,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         if wl_filename in (True, ""):  # Indicates the default whitelist.
             wl_filename = WHITELIST
         if wl_filename:
-            text = Path(wl_filename).read_text(encoding="utf-8")
+            text = Path(wl_filename).expanduser().read_text(encoding="utf-8")
             for line in text.splitlines():
                 if (line := line.strip().lower()) and not line.startswith("#"):
                     whitelist.add(line)
@@ -140,7 +140,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         # Read the tree
         if c14n_filename:
             self._log.debug("Loading canonicalization tree {}", c14n_filename)
-            with Path(c14n_filename).open(encoding="utf-8") as f:
+            with Path(c14n_filename).expanduser().open(encoding="utf-8") as f:
                 genres_tree = yaml.safe_load(f)
             flatten_tree(genres_tree, [], c14n_branches)
         return c14n_branches, canonicalize
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -24,6 +24,7 @@ https://gist.github.com/1241307
 
 import os
 import traceback
+from pathlib import Path
 from typing import Union
 
 import pylast
@@ -140,9 +141,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         # Read the tree
         if c14n_filename:
             self._log.debug("Loading canonicalization tree {}", c14n_filename)
-            with open(
-                syspath(normpath(c14n_filename)), "r", encoding="utf-8"
-            ) as f:
+            with Path(c14n_filename).open(encoding="utf-8") as f:
                 genres_tree = yaml.safe_load(f)
             flatten_tree(genres_tree, [], c14n_branches)
         return c14n_branches, canonicalize
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -111,9 +111,11 @@ class LastGenrePlugin(plugins.BeetsPlugin):
             self.import_stages = [self.imported]
 
         self._genre_cache = {}
+        self.whitelist = self._load_whitelist()
+        self.c14n_branches, self.canonicalize = self._load_c14n_tree()
 
-        # Read the whitelist file if enabled.
-        self.whitelist = set()
+    def _load_whitelist(self) -> set[str]:
+        whitelist = set()
         wl_filename = self.config["whitelist"].get()
         if wl_filename in (True, ""):  # Indicates the default whitelist.
             wl_filename = WHITELIST
@@ -123,27 +125,27 @@ class LastGenrePlugin(plugins.BeetsPlugin):
                 for line in f:
                     line = line.decode("utf-8").strip().lower()
                     if line and not line.startswith("#"):
-                        self.whitelist.add(line)
+                        whitelist.add(line)
+        return whitelist
 
-        # Read the genres tree for canonicalization if enabled.
-        self.c14n_branches = []
+    def _load_c14n_tree(self) -> tuple[list[list[str]], bool]:
+        c14n_branches = []
         c14n_filename = self.config["canonical"].get()
-        self.canonicalize = c14n_filename is not False
-
+        canonicalize = c14n_filename is not False
         # Default tree
         if c14n_filename in (True, ""):
             c14n_filename = C14N_TREE
-        elif not self.canonicalize and self.config["prefer_specific"].get():
+        elif not canonicalize and self.config["prefer_specific"].get():
             # prefer_specific requires a tree, load default tree
             c14n_filename = C14N_TREE
-
         # Read the tree
         if c14n_filename:
             self._log.debug("Loading canonicalization tree {}", c14n_filename)
             c14n_filename = normpath(c14n_filename)
             with codecs.open(c14n_filename, "r", encoding="utf-8") as f:
                 genres_tree = yaml.safe_load(f)
-            flatten_tree(genres_tree, [], self.c14n_branches)
+            flatten_tree(genres_tree, [], c14n_branches)
+        return c14n_branches, canonicalize
 
     @property
     def sources(self) -> tuple[str, ...]:
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -117,7 +117,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
     def _load_whitelist(self) -> set[str]:
         """Load the whitelist from a text file.
 
-        Default whitelist is used if config is True or empty string.
+        Default whitelist is used if config is True, empty string or set to "nothing".
         """
         whitelist = set()
         wl_filename = self.config["whitelist"].get()
@@ -135,8 +135,8 @@ class LastGenrePlugin(plugins.BeetsPlugin):
     def _load_c14n_tree(self) -> tuple[list[list[str]], bool]:
         """Load the canonicalization tree from a YAML file.
 
-        Default tree is used if config is True or empty string, or if
-        prefer_specific is enabled.
+        Default tree is used if config is True, empty string, set to "nothing"
+        or if prefer_specific is enabled.
         """
         c14n_branches: list[list[str]] = []
         c14n_filename = self.config["canonical"].get()
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -22,7 +22,6 @@ The scraper script used is available here:
 https://gist.github.com/1241307
 """
 
-import codecs
 import os
 import traceback
 from typing import Union
@@ -141,7 +140,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         # Read the tree
         if c14n_filename:
             self._log.debug("Loading canonicalization tree {0}", c14n_filename)
-            with codecs.open(
+            with open(
                 syspath(normpath(c14n_filename)), "r", encoding="utf-8"
             ) as f:
                 genres_tree = yaml.safe_load(f)
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -140,9 +140,10 @@ class LastGenrePlugin(plugins.BeetsPlugin):
             c14n_filename = C14N_TREE
         # Read the tree
         if c14n_filename:
-            self._log.debug("Loading canonicalization tree {}", c14n_filename)
-            c14n_filename = normpath(c14n_filename)
-            with codecs.open(c14n_filename, "r", encoding="utf-8") as f:
+            self._log.debug("Loading canonicalization tree {0}", c14n_filename)
+            with codecs.open(
+                str(normpath(c14n_filename)), "r", encoding="utf-8"
+            ) as f:
                 genres_tree = yaml.safe_load(f)
             flatten_tree(genres_tree, [], c14n_branches)
         return c14n_branches, canonicalize
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -32,7 +32,7 @@ import yaml
 
 from beets import config, library, plugins, ui
 from beets.library import Album, Item
-from beets.util import normpath, plurality, syspath, unique_list
+from beets.util import plurality, unique_list
 
 LASTFM = pylast.LastFMNetwork(api_key=plugins.LASTFM_KEY)
 
@@ -120,12 +120,11 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         if wl_filename in (True, ""):  # Indicates the default whitelist.
             wl_filename = WHITELIST
         if wl_filename:
-            wl_filename = normpath(wl_filename)
-            with open(wl_filename, "rb") as f:
-                for raw_line in f:
-                    line = raw_line.decode("utf-8").strip().lower()
-                    if line and not line.startswith("#"):
-                        whitelist.add(line)
+            text = Path(wl_filename).read_text(encoding="utf-8")
+            for line in text.splitlines():
+                if (line := line.strip().lower()) and not line.startswith("#"):
+                    whitelist.add(line)
+
         return whitelist
 
     def _load_c14n_tree(self) -> tuple[list[list[str]], bool]:
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -132,10 +132,10 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         c14n_filename = self.config["canonical"].get()
         canonicalize = c14n_filename is not False
         # Default tree
-        if c14n_filename in (True, ""):
-            c14n_filename = C14N_TREE
-        elif not canonicalize and self.config["prefer_specific"].get():
+        if c14n_filename in (True, "") or (
             # prefer_specific requires a tree, load default tree
+            not canonicalize and self.config["prefer_specific"].get()
+        ):
             c14n_filename = C14N_TREE
         # Read the tree
         if c14n_filename:
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -139,7 +139,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
             c14n_filename = C14N_TREE
         # Read the tree
         if c14n_filename:
-            self._log.debug("Loading canonicalization tree {0}", c14n_filename)
+            self._log.debug("Loading canonicalization tree {}", c14n_filename)
             with open(
                 syspath(normpath(c14n_filename)), "r", encoding="utf-8"
             ) as f:
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -124,6 +124,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         if wl_filename in (True, "", None):  # Indicates the default whitelist.
             wl_filename = WHITELIST
         if wl_filename:
+            self._log.debug("Loading whitelist {}", wl_filename)
             text = Path(wl_filename).expanduser().read_text(encoding="utf-8")
             for line in text.splitlines():
                 if (line := line.strip().lower()) and not line.startswith("#"):
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -114,7 +114,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         self.whitelist = self._load_whitelist()
         self.c14n_branches, self.canonicalize = self._load_c14n_tree()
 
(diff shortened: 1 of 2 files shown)


beetsplug/lastgenre/__init__.py
```diff
@@ -123,27 +125,27 @@ def setup(self):
                 for line in f:
                     line = line.decode("utf-8").strip().lower()
                     if line and not line.startswith("#"):
-                        self.whitelist.add(line)
+                        whitelist.add(line)
+        return whitelist
```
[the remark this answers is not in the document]
JOJ0: Currently and since ever we fail like this if whitelist file not found:


```
** error loading plugin lastgenre
  File "/Users/jojo/git/beets/beets/plugins.py", line 367, in _get_plugin
    return obj()
           ^^^^^
  File "/Users/jojo/git/beets/beetsplug/lastgenre/__init__.py", line 106, in __init__
    self.setup()
  File "/Users/jojo/git/beets/beetsplug/lastgenre/__init__.py", line 114, in setup
    self.whitelist = self._load_whitelist()
                     ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jojo/git/beets/beetsplug/lastgenre/__init__.py", line 124, in _load_whitelist
    with open(wl_filename, "rb") as f:
         ^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: b'/Users/jojo/git/home_config/lastgenre_conf/BOLLOCKS.txt'
error: unknown command 'lastgenre'
```

that bad? Since it's rather fatal I think the current behavior is ok. What do you all think?

beetsplug/lastgenre/__init__.py
```diff
@@ -123,27 +125,27 @@ def setup(self):
...
-
         # Read the tree
         if c14n_filename:
             self._log.debug("Loading canonicalization tree {0}", c14n_filename)
             c14n_filename = normpath(c14n_filename)
             with codecs.open(c14n_filename, "r", encoding="utf-8") as f:
                 genres_tree = yaml.safe_load(f)
-            flatten_tree(genres_tree, [], self.c14n_branches)
+            flatten_tree(genres_tree, [], c14n_branches)
+        return c14n_branches, canonicalize
```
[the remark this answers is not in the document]
JOJ0: wrong filename fails like this which IMHO is ok:

```
** error loading plugin lastgenre
  File "/Users/jtiefenbacher/git/beets/beets/plugins.py", line 367, in _get_plugin
    return obj()
           ^^^^^
  File "/Users/jtiefenbacher/git/beets/beetsplug/lastgenre/__init__.py", line 106, in __init__
    self.setup()
  File "/Users/jtiefenbacher/git/beets/beetsplug/lastgenre/__init__.py", line 115, in setup
    self.c14n_branches, self.canonicalize = self._load_c14n_tree()
                                            ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jtiefenbacher/git/beets/beetsplug/lastgenre/__init__.py", line 145, in _load_c14n_tree
    with codecs.open(c14n_filename, "r", encoding="utf-8") as f:
  File "<frozen codecs>", line 918, in open
FileNotFoundError: [Errno 2] No such file or directory: b'/Users/jtiefenbacher/git/home_config/lastgenre_conf/BOLLOCKS.yaml'
error: unknown command 'lastgenre'
```

It's documented that canonical must be explicitly set to no/False to really disable it. Setting to "" will NOT disable it but load the default anyway. 

And all that is ignored if prefer_specific=yes/True, the default will be loaded if there is none configured. And that also is well documented. 

So far all good here....

If genres-tree.yaml is faulty, we fail ugly but fail and quit which is good:

```
...
...
  File "/Users/jojo/.pyenv/versions/3.12.7/envs/beets312/lib/python3.12/site-packages/yaml/parser.py", line 392, in parse_block_sequence_entry
    raise ParserError("while parsing a block collection", self.marks[-1],
yaml.parser.ParserError: while parsing a block collection
  in "b'/Users/jojo/git/home_config/lastgenre_conf/genres-tree.yaml'", line 1, column 1
expected <block end>, but found '<scalar>'
  in "b'/Users/jojo/git/home_config/lastgenre_conf/genres-tree.yaml'", line 546, column 2
 ```

File: beetsplug/lastgenre/__init__.py [hunk 3]
```diff
@@ -111,39 +111,41 @@ def setup(self):
...
         wl_filename = self.config["whitelist"].get()
         if wl_filename in (True, ""):  # Indicates the default whitelist.
             wl_filename = WHITELIST
         if wl_filename:
             wl_filename = normpath(wl_filename)
             with open(wl_filename, "rb") as f:
-                for line in f:
-                    line = line.decode("utf-8").strip().lower()
+                for raw_line in f:
+                    line = raw_line.decode("utf-8").strip().lower()
```
JOJ0: So I fixed the mypy errors I got:

```
beetsplug/lastgenre/__init__.py:126: error: Name "line" already defined on line 125  [no-redef]
beetsplug/lastgenre/__init__.py:127: error: Argument 1 to "startswith" of "bytes" has incompatible type "str"; expected "Buffer | tuple[Buffer, ...]"  [arg-type]
```

I need some advice here @snejus, @semohr, @wisp3rwind or @Serene-Arc : Is this the right way to do it? And isn't the loading as rb and then translating each line to str here anyway kind of redundant? Thanks in advance!

beetsplug/lastgenre/__init__.py
```diff
@@ -111,39 +111,41 @@ def setup(self):
...
                     if line and not line.startswith("#"):
-                        self.whitelist.add(line)
+                        whitelist.add(line)
+        return whitelist
 
-        # Read the genres tree for canonicalization if enabled.
-        self.c14n_branches = []
+    def _load_c14n_tree(self) -> tuple[list[list[str]], bool]:
+        c14n_branches: list[list[str]] = []
```
JOJ0, line 141: Also here, please double check if that is the correct way to do it. I fixed this one:

```
beetsplug/lastgenre/__init__.py:132: error: Need type annotation for "c14n_branches" (hint: "c14n_branches: list[<type>] = ...")  [var-annotated]
```

beetsplug/lastgenre/__init__.py
```diff
@@ -111,39 +111,42 @@ def setup(self):
...
             c14n_filename = C14N_TREE
-
         # Read the tree
         if c14n_filename:
             self._log.debug("Loading canonicalization tree {0}", c14n_filename)
-            c14n_filename = normpath(c14n_filename)
-            with codecs.open(c14n_filename, "r", encoding="utf-8") as f:
+            with codecs.open(
+                str(normpath(c14n_filename)), "r", encoding="utf-8"
+            ) as f:
```
JOJ0: Had a mypy error here that only happened in the CI, not locally:
```
beetsplug/lastgenre/__init__.py:144: error: Argument 1 to "open" has incompatible type "bytes"; expected "str"  [arg-type]
```
 Seems like codecs.open wants str., normpath returns bytes though. normpath is essential here since it does all kinds of sanity checks and transforms so I'd like to keep it. Wrapping with str() is the most straightforward way I suppose? 
 
 Another way that works would be to wrap with syspath(), would this be better / more "beetonic"? @wisp3rwind you are always a good address with such questions ;-) thanks in advance! :-)

beetsplug/lastgenre/__init__.py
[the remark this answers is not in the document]
JOJ0: so we agree then :-) resolved.

JOJ0:
oh my, now loading of genres-tree is broken in the ci. will fix that tomorrow and back to draft.... .-/

beetsplug/lastgenre/__init__.py
JOJ0: I think I answered this myself. Yes syspath(normpath()) is the way to go here. I had these errors in the CI:

```
FAILED test/plugins/test_lastgenre.py::LastGenrePluginTest::test_c14n_only - FileNotFoundError: [Errno 2] No such file or directory: "b'/home/runner/work/beets/beets/beetsplug/lastgenre/genres-tree.yaml'"
```

which are now fixed. Sorry the bother and nevermind! Resolved.

beetsplug/lastgenre/__init__.py
```diff
@@ -111,39 +111,42 @@ def setup(self):
...
         if wl_filename:
             wl_filename = normpath(wl_filename)
             with open(wl_filename, "rb") as f:
-                for line in f:
-                    line = line.decode("utf-8").strip().lower()
+                for raw_line in f:
+                    line = raw_line.decode("utf-8").strip().lower()
                     if line and not line.startswith("#"):
-                        self.whitelist.add(line)
+                        whitelist.add(line)
```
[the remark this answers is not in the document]
JOJ0: That diff is wrong and I also questioned this elsewhere. set to resolved.

beetsplug/lastgenre/__init__.py
```diff
@@ -111,39 +111,42 @@
...
         # Read the tree
         if c14n_filename:
-            self._log.debug("Loading canonicalization tree {}", c14n_filename)
-            c14n_filename = normpath(c14n_filename)
-            with codecs.open(c14n_filename, "r", encoding="utf-8") as f:
+            self._log.debug("Loading canonicalization tree {0}", c14n_filename)
+            with codecs.open(
+                syspath(normpath(c14n_filename)), "r", encoding="utf-8"
+            ) as f:
                 genres_tree = yaml.safe_load(f)
```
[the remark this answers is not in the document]
JOJ0: Did some research and most probably this is a relict of python2/3 compatibility. let's accept this. using open() is a good idea and ought to be a little faster as well.

beetsplug/lastgenre/__init__.py
JOJ0: ok I did my homework and researched about it. It's a good solution. Let's keep it. Resolved. Nevermind. @sourcery-ai review

beetsplug/lastgenre/__init__.py
JOJ0: also here decision that it's ok. resolved.

JOJ0:
@sourcery-ai review

beetsplug/lastgenre/__init__.py
```diff
@@ -111,39 +110,42 @@ def setup(self):
...
+        self.c14n_branches, self.canonicalize = self._load_c14n_tree()
 
-        # Read the whitelist file if enabled.
-        self.whitelist = set()
+    def _load_whitelist(self) -> set[str]:
+        whitelist = set()
         wl_filename = self.config["whitelist"].get()
         if wl_filename in (True, ""):  # Indicates the default whitelist.
             wl_filename = WHITELIST
         if wl_filename:
```
snejus, line 126: You can simplify file loading with

```py
from pathlib import Path
...

text = Path(wl_filename).read_text()
for line in text.splitlines():
    if (line := line.strip().lower()) and not line.startswith("#"):
        whitelist.add(line)
```

beetsplug/lastgenre/__init__.py
```diff
@@ -111,39 +110,42 @@ def setup(self):
...
-        # Read the genres tree for canonicalization if enabled.
-        self.c14n_branches = []
+    def _load_c14n_tree(self) -> tuple[list[list[str]], bool]:
+        c14n_branches: list[list[str]] = []
         c14n_filename = self.config["canonical"].get()
-        self.canonicalize = c14n_filename is not False
-
+        canonicalize = c14n_filename is not False
         # Default tree
         if c14n_filename in (True, ""):
```
snejus: Note that an empty configuration reads `None` instead of an empty string. I think you can join both conditions into one:

```py
        if c14n_filename in {True, None} or (
            # prefer_specific requires a tree, load default tree
            not canonicalize and self.config["prefer_specific"].get()
        ):
            c14n_filename = C14N_TREE
```

beetsplug/lastgenre/__init__.py
```diff
@@ -111,39 +110,42 @@ def setup(self):
...
             c14n_filename = C14N_TREE
-
         # Read the tree
         if c14n_filename:
             self._log.debug("Loading canonicalization tree {}", c14n_filename)
-            c14n_filename = normpath(c14n_filename)
-            with codecs.open(c14n_filename, "r", encoding="utf-8") as f:
+            with open(
+                syspath(normpath(c14n_filename)), "r", encoding="utf-8"
+            ) as f:
```
snejus: You can simplify this with `pathlib.Path`:
```suggestion
            with Path(c14n_filename).open() as f:
```

beetsplug/lastgenre/__init__.py
snejus: And replace

```diff
-        if wl_filename in (True, ""):
+        if wl_filename in (True, None):
```

beetsplug/lastgenre/__init__.py
JOJ0: But changing to {True None} breaks the intended behavior:

```
FAILED test/plugins/test_lastgenre.py::LastGenrePluginTest::test_empty_string_enables_canonical - AssertionError: assert ['delta blues'] == ['blues']
FAILED test/plugins/test_lastgenre.py::LastGenrePluginTest::test_empty_string_enables_whitelist - AssertionError: assert ['iota blues'] == []
======================================================================== 2 failed, 31 passed, 1 warning in 0.57s ========================================================================
```

beetsplug/lastgenre/__init__.py
JOJ0: This breaks paths including `~`:

```
text = Path(wl_filename).read_text()
```

I went with:

```
text = Path(wl_filename).expanduser().read_text()
```

beetsplug/lastgenre/__init__.py
JOJ0: Note that here the default is `False`, so when user doesn't set that configuration it won't be `None`

snejus: > Note that here the default is False, so when user doesn't set that configuration it won't be None

By _empty configuration_ I meant 

```yaml
lastgenre:
  canonical:
```



```py
>>> from beets import config
>>> type(config["lastgenre"]["canonical"].get())
NoneType
```
Ppresumably, this is the case that the logic is intended to handle? Or did I make the  wrong assumption?

JOJ0: wrong assumption. see the test output above:


test_empty_string_enables_canonical

JOJ0: See bc23fc69c, let's simply check for both "" and None

JOJ0: and it also doesn't hurt to show which whitelist file is loading exactly: 8d83c59ec

snejus: Can you tell me what value do I need to place in my configuration in order to read an empty string here?

```py
        wl_filename = self.config["whitelist"].get()
```

I'm fairly convinced that those tests you mentioned are set up incorrectly and are thus misleading.

JOJ0:
```
lastgenre:
    whitelist: ""
    canonical: ""
```

JOJ0:
```
    def test_empty_string_enables_canonical(self):
        """For backwards compatibility, setting the `canonical` option
        to the empty string enables it using the default tree.
        """
        self._setup_config(canonical="", count=99)
        assert self.plugin._resolve_genres(["delta blues"]) == ["blues"]

    def test_empty_string_enables_whitelist(self):
        """Again for backwards compatibility, setting the `whitelist`
        option to the empty string enables the default set of genres.
        """
        self._setup_config(whitelist="")
        assert self.plugin._resolve_genres(["iota blues"]) == []
  ```

The original commit that added those tests: https://github.com/beetbox/beets/commit/233f71a4574b5a10d1182a3a63ce5cd23c4268f4

Comment (JOJ0) [reply beetsplug/lastgenre/__init__.py]:
```
 jtb311  dev  ~/git/beets   lastgenre_forbidden  python
Python 3.11.10 (main, Oct  3 2024, 09:12:28) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
>>> from beets import config
>>> config["lastgenre"].get()
OrderedDict([('force', False), ('keep_existing', False), ('whitelist', ''), ('canonical', '~/git/home_config/lastgenre_conf/genres-tree.yaml'), ('prefer_specific', False), ('count', 4), ('separator', ', '), ('auto', False), ('source', 'track'), ('min_weight', 10), ('extended_debug', True), ('blacklist', '~/git/home_config/lastgenre_conf/blacklist-dev.txt'), ('aliases', '')])
>>> config["lastgenre"]["whitelist"].get()
''
>>> type(config["lastgenre"]["whitelist"].get())
<class 'str'>
>>>
```

Comment (snejus) [reply beetsplug/lastgenre/__init__.py]:
```
> >>> config["lastgenre"]["whitelist"].get()
> ''
```

This is not an empty string, it's two single quotes `"''"` 😅 

It seems like an empty string cannot be possibly read? The tests _do not read_ an actual config from the file system, they just set the configuration to the given value:

```py
    def _setup_config(
        self, whitelist=False, canonical=False, count=1, prefer_specific=False
    ):
        self.config["lastgenre"]["canonical"] = canonical
```

Therefore, this can be set to an empty string and it works. However, this does not reflect the reality: as I've shown above an empty configuration reads `None`. Thus, the tests should be updated accordingly:

```py
    def test_empty_configuration_enables_canonical(self):
        """For backwards compatibility, setting the `canonical` option
        to the empty string enables it using the default tree.
        """
        self._setup_config(canonical=None, count=99)
        assert self.plugin._resolve_genres(["delta blues"]) == ["blues"]
```

Meanwhile, the logic should only handle `None` value - please remove `""` from there to make sure this does not cause confusion to anyone else that comes across this in the future.

JOJ0: I don't understand, I'm sorry. Let's leave out the tests for now:

If I set this in my actual config (double quotes aka empty string?):

```
lastgenre:
    whitelist: ""
    canonical: ""
```

I read a typ str:

```
Python 3.11.10 (main, Oct  3 2024, 09:12:28) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
>>> from beets import config
>>> type(config["lastgenre"]["whitelist"].get())
<class 'str'>
>>> type(config["lastgenre"]["canonical"].get())
<class 'str'>
>>>
```

I'm using: confuse version 2.0.1

JOJ0: So what I'm trying to say is: A user might have exactly that config because they want to fall back to the defaults. If I remove "" from that check as you suggest, their config will suddenly behave differently - We don't want that, do we? At least I do not want to break users' experience! Really I don't. Maybe you misunderstood something here still? We do not want to check for missing config! We want to check for "" or '' :-)

JOJ0: And regarding

>  please remove "" from there to make sure this does not cause confusion to anyone else that comes across this in the future.

Have you seen the docstring I put on top of those functions? It clearly explains the behavior. This will prevent confusion to anyone else in the future :-)

```
    def _load_whitelist(self) -> set[str]:
        """Load the whitelist from a text file.

        Default whitelist is used if config is True, empty string or set to "nothing".
        """
 ```
 
 ```
     def _load_c14n_tree(self) -> tuple[list[list[str]], bool]:
        """Load the canonicalization tree from a YAML file.

        Default tree is used if config is True, empty string, set to "nothing"
        or if prefer_specific is enabled.
        """
```

I think we've made the configuration handling even more robust by adding the None check too. So thanks for the suggestion! Appreciated! But I really do not want users having to change anything in their config only because of this refactoring change!

snejus: OK, I see - sorry, I was confused since this behaviour is not mentioned in the docs. My intuition told me that this configuration
```yaml
  canonical:
```
is more likely to mean _use the default whitelist_. Whereas

```yaml
  canonical: ""
```

seems to be more of an explicit empty override (_don't use any_ / same as setting it to **no**).

This is the sort of pattern that I've seen in the codebase:
1. Autotagging: if a metadata source returns value `None` for a field (__no value__), it is ignored. Meanwhile, an empty string `""` is treated as a 'legitimate' value and it replaces the previous value of that field.
2. **Lyrics** plugin `fallback` configuration which is used as a default when lyrics are not found for the track. If users set `fallback: ""`, the plugin clears this field in case lyrics aren't found. On the other hand, when they set `fallback:`, it means 'there's no fallback', and the previous lyrics are not cleared.

In any case, I agree with you that we should not force users to update their configs, so I'm happy with this behaviour (for now 😉).

semohr (approved):

beetsplug/lastgenre/__init__.py
JOJ0: Yes, for now let's leave this "" fallback to defaults (which is here for whatever reason @sampsyo put it there as we see in https://github.com/beetbox/beets/commit/233f71a4574b5a10d1182a3a63ce5cd23c4268f4. I don't find it useful to investigate further at this point - for now let's not break existing configs (fun fact: mine often is like that. I switch between "" and an actual custom path often during testing, even though I could write `yes` to get the default :-O)

But the real fix I think - including possibly also in other places as you point out - is in my opinion to sanity check for "", None and then throw a `UserError` - in case of the whitelist setting something like this:

`Invalid whitelist setting - True, False or path to custom file allowed`


This would fail gracefully - not silent - and give the user a chance to update with clear instructions.

PR #5981, opened 2025-08-31
lastgenre: Use albumartists field to improve last.fm results

JOJ0:
Often last.fm does not find artist genres for delimiter-separated artist names (eg. "Artist One, Artist Two") or where multiple artists are combined with "concatenation words" like "and" , "+", "featuring" and so on.

This fix gathers each artist's last.fm genre separately by using Beets' mutli-valued `albumartists` field to improve the likeliness of finding genres in the artist genre fetching stage.

Refactoring was done along the existing genre fetching helper functions (`fetch_album_genre`, `fetch_track_genre`, ...):

- last.fm can be asked for genre for these combinations of metadata:
   - albumartist/album
   - artist/track
   - artist
- Instead of passing `Album` or `Item` objects directly to these helpers., generalize them and pass the (string) metadata directly.
- Passing "what's to fetch" in the callers instead of hiding it in the methods also helps readability in `_get_genre()`
- And reduces the requirement at hand for another additional method (or adaptation) to support "multi-albumartist genre fetching" 

- [x] ~Documentation~ 
- [x] Changelog
- [x] Tests

### DIFF
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -422,6 +422,19 @@ class LastGenrePlugin(plugins.BeetsPlugin):
             elif obj.albumartist != config["va_name"].as_str():
                 new_genres = self.fetch_artist_genre(obj.albumartist)
                 stage_label = "album artist"
+                if not new_genres:
+                    self._tunelog(
+                        'No album artist genre found for "{}", '
+                        "trying multi-valued field...",
+                        obj.albumartist,
+                    )
+                    for albumartist in obj.albumartists:
+                        self._tunelog(
+                            'Fetching artist genre for "{}"', albumartist
+                        )
+                        new_genres += self.fetch_artist_genre(albumartist)
+                    if new_genres:
+                        stage_label = "multi-valued album artist"
             else:
                 # For "Various Artists", pick the most popular track genre.
                 item_genres = []
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -438,6 +438,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
             else:
                 # For "Various Artists", pick the most popular track genre.
                 item_genres = []
+                assert isinstance(obj, Album)  # Type narrowing for mypy
                 for item in obj.items():
                     item_genre = None
                     if "track" in self.sources:
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -28,7 +28,7 @@ import os
 import traceback
 from functools import singledispatchmethod
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Callable
 
 import pylast
 import yaml
@@ -259,9 +259,11 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         valid_tags = [t for t in tags if self._is_valid(t)]
         return valid_tags[:count]
 
-    def fetch_genre(self, lastfm_obj):
-        """Return the genre for a pylast entity or None if no suitable genre
-        can be found. Ex. 'Electronic, House, Dance'
+    def fetch_genre(
+        self, lastfm_obj: pylast.Album | pylast.Artist | pylast.Track
+    ) -> list[str]:
+        """Return genres for a pylast entity. Returns an empty list if
+        no suitable genres are found.
         """
         min_weight = self.config["min_weight"].get(int)
         return self._tags_for(lastfm_obj, min_weight)
@@ -278,8 +280,10 @@ class LastGenrePlugin(plugins.BeetsPlugin):
 
     # Cached last.fm entity lookups.
 
-    def _last_lookup(self, entity, method, *args):
-        """Get a genre based on the named entity using the callable `method`
+    def _last_lookup(
+        self, entity: str, method: Callable[..., Any], *args: str
+    ) -> list[str]:
+        """Get genres based on the named entity using the callable `method`
         whose arguments are given in the sequence `args`. The genre lookup
         is cached based on the entity name and the arguments.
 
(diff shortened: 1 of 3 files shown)


beetsplug/lastgenre/__init__.py
```diff
@@ -421,20 +417,36 @@ def _try_resolve_stage(stage_label: str, keep_genres, new_genres):
...
+                        self._tunelog(
+                            'Fetching artist genre for "{}"', albumartist
+                        )
+                        new_genres += self.fetch_artist_genre(albumartist)
+                    if new_genres:
+                        stage_label = "multi-valued album artist"
             else:
                 # For "Various Artists", pick the most popular track genre.
                 item_genres = []
+                assert isinstance(obj, Album)  # Type narrowing for mypy
```
JOJ0, line 445: This was quite tricky to fix. The original mypy error was:

```
$ mypy beetsplug/lastgenre/__init__.py --show-error-codes
beetsplug/lastgenre/__init__.py:443: error: "tuple[str, Any]" has no attribute "artist"  [attr-defined]
beetsplug/lastgenre/__init__.py:443: error: "tuple[str, Any]" has no attribute "title"  [attr-defined]
beetsplug/lastgenre/__init__.py:446: error: "tuple[str, Any]" has no attribute "artist"  [attr-defined]
Found 3 errors in 1 file (checked 1 source file)
```

With the assert I'm convincing mypy to consider `obj` being an `Album` object, otherwise it would have detected it as an `[Iterable[tuple[str, Any]]]` which `LibModel.items()` returns.  

This seemed to be the most easy fix. I tried to fiddle around with changing types within `models.py`:

https://github.com/beetbox/beets/blob/f79c125d151bdc17c29dce3249a04d22116fbe1f/beets/library/models.py#L37

but I'm out of my depth here. Some advice for a better fix or if that's a good enough one already please @semohr ?

beetsplug/lastgenre/__init__.py
```diff
@@ -421,20 +417,36 @@ def _try_resolve_stage(stage_label: str, keep_genres, new_genres):
...
+                        "trying multi-valued field...",
+                        obj.albumartist,
+                    )
+                    for albumartist in obj.albumartists:
+                        self._tunelog(
+                            'Fetching artist genre for "{}"', albumartist
+                        )
+                        new_genres += self.fetch_artist_genre(albumartist)
+                    if new_genres:
+                        stage_label = "multi-valued album artist"
```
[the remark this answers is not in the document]
JOJ0: deduplication happpens in `_resolve_genres` using the `unique_list` helper from utils package

JOJ0:
I'd like to rerequest some eyes here @snejus @semohr @henry-oberholtzer 

also i don't think it interferes much with https://github.com/beetbox/beets/pull/6169

Nedra1998:
This solves the issues I was having, thank you for this.

A minor note is that since it only tries to use the albumartists for the `fetch_artist_genre`, there could be a potential album match with `fetch_album_genre` for the individual artists (e.g. `fetch_album_genre(obj.albumartists[0], obj.album)`). Which might have more applicable tags than just falling back to the artists.

But I'm not sure how common that is, and the artist tags are plenty for my use case.

beetsplug/lastgenre/__init__.py
```diff
@@ -300,24 +300,20 @@ def _last_lookup(self, entity, method, *args):
...
+        return self._last_lookup("artist", LASTFM.get_artist, artist)
 
-    def fetch_artist_genre(self, item):
-        """Returns raw track artist genres from Last.fm for this Item."""
-        return self._last_lookup("artist", LASTFM.get_artist, item.artist)
-
-    def fetch_track_genre(self, obj):
-        """Returns raw track genres from Last.fm for this Item."""
+    def fetch_track_genre(self, trackartist, tracktitle):
+        """Return genres from Last.fm for the track by artist."""
```
henry-oberholtzer: It's not a big deal, but I think adding type hints to these could be nice & consistent with the rest of the plugin.

beetsplug/lastgenre/__init__.py
henry-oberholtzer: This does seem like the simplest fix [according to mypy's own documentation](https://mypy.readthedocs.io/en/stable/type_narrowing.html#type-narrowing-expressions), so I can't think there would be much else we'd want to do? I'm also not the most familiar with mypy but this seems like the right track.

henry-oberholtzer (changes requested):
Looks great to me, for my part I think some type hints would be good to add to the fetch_album_ fields while we're at it.

JOJ0:
> Looks great to me, for my part I think some type hints would be good to add to the fetch_album_ fields while we're at it.

thanks! will do!

JOJ0:
Hi @henry-oberholtzer finally gotten around to add type hints to the `_fetch_...` methods. I realized that the plugin at this point was not far from full type hint coverage, so I finally added ALL of them even though some of them are not related to this PR.

Slightly more hassle for your re-review but I'm sure you'll manage. It's not tooo much actually and I finally would like to [get this task on my bucket list done](https://github.com/beetbox/beets/discussions/5915#discussion-8686159). Thanks in advance!

beetsplug/lastgenre/__init__.py
```diff
@@ -584,11 +608,12 @@ def _tags_for(self, obj, min_weight=None):
         # Work around an inconsistency in pylast where
         # Album.get_top_tags() does not return TopItem instances.
         # https://github.com/pylast/pylast/issues/86
+        obj_to_query: Any = obj
         if isinstance(obj, pylast.Album):
-            obj = super(pylast.Album, obj)
+            obj_to_query = super(pylast.Album, obj)
```
JOJ0: just fyi @henry-oberholtzer: here a new variable was the solution to not changing the type along the way which mypy does not like

beetsplug/lastgenre/__init__.py
```diff
@@ -278,8 +280,10 @@ def _is_valid(self, genre: str) -> bool:
 
     # Cached last.fm entity lookups.
 
-    def _last_lookup(self, entity, method, *args):
-        """Get a genre based on the named entity using the callable `method`
+    def _last_lookup(
+        self, entity: str, method: Callable[..., Any], *args: str
```
JOJ0, line 284: @henry-oberholtzer can you take a closer look here please. Not sure about the method type. thanks

Update: 
I researched this (was a suggestion by copilot I just accepted) and now understand that this is the right way to do it.

JOJ0:
> ... so I finally added ALL of them even though some of them are not related to this PR.

I reconsidered and split out the remaining types to a new PR #6239 . Makes both of them easier to read and review. Thanks in advance for a final look on this one as a first step.

henry-oberholtzer:
No problem! I'm pro more type hints everywhere so happy to review that one too. I should be able to take a look at it this weekend.

henry-oberholtzer (approved):
Looks great!!

PR #6004, opened 2025-09-13
Issue#5998

cp90-pixel:
## Summary
- fall back to stdlib JSONDecodeError if requests lacks it
- catch JSONDecodeError via shared alias in lyrics plugin

## Testing

- [x] - `pre-commit run --files beetsplug/lyrics.py`
- [x] - `pytest test/plugins/test_lyrics.py test/plugins/lyrics_pages.py

`

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -31,6 +31,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Iterable, Iterator, NamedTuple
 from urllib.parse import quote, quote_plus, urlencode, urlparse
 
+import json
 import langdetect
 import requests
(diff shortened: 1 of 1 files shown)


snejus (changes requested):
Thanks for addressing this! I think, we'd rather keep using `requests.JSONDecodeError` unconditionally in `lyrics.py` and instead update the `requests` dependency requirement to ensure that users do not install outdated versions.

Would you be happy to do so?

1. Update `requests` version requirement to require the version that `JSONDecodeError` was introduced as the minimum (I think it was `>=2.27.0`).
2. Run `uv tool run poetry<2 update requests`

PR #6013, opened 2025-09-15
Move `art.py` to `beetsplug._utils` package to avoid polluting core namespace

semohr:
This PR moves the `art.py` module, which is only used by the plugin architecture, to avoid polluting the main beets namespace. 

Added commit where file is moved to the `.git-blame-ignore-revs` file.

- [x] Changelog

## Summary by Sourcery

Move the plugin-only art module out of the core namespace and into the beetsplug._utils package

Enhancements:
- Relocate art.py to beetsplug._utils to prevent namespace pollution
- Update plugin imports in convert and embedart to reference the new art location

Documentation:
- Add changelog entry for the module relocation

Chores:
- Add the moved file commit to .git-blame-ignore-revs


snejus:
It's only used by plugins - I don't think it belongs under beets core. What about `beetsplug/_art.py`?

semohr:
Thought initially a shared folder for common beetsplug functions might be useful.We have the same issue with the `vfs.py` file in #6017. 


Am not sure what is more maintainable in the long run. No strong opinion either way as long as it is kinda consistent.

snejus:
`beets/vfs.py` is also only used by plugins, so it should be placed under `beetsplug` methinks.

semohr:
Glad that we agree here :+1: That's why I started the PRs...

- ATM the files are placed in `beetsplug/_utils/*.py`.
- We can move them if you feel like they should be move

As I said:
> Am not sure what is more maintainable in the long run. No strong opinion either way as long as it is kinda consistent.

snejus:
Sorry - I'm an idiot. For some reason, I assumed that you moved them under `beets/_utils` 🤦🏼

snejus (approved):

semohr:
No worries. I already thought you might have misread it. That's why I clarified :upside_down_face:

PR #6015, opened 2025-09-15
Adds a zero_disc_if_single_disc to the zero plugin

michaelkrieger:
Adds a `zero_disc_if_single_disc` boolean to the zero plugin for writing to files.  Adds the logic that, if disctotal is set and there is only one disc in disctotal, that the disc is not set.

This keeps tags cleaner, only using disc on multi-disc albums. The disctotal is not touched, particularly as this is not usually displayed in most clients.

The field is removed only for writing the tags, but the disc number is maintained in the database to avoid breaking anything that may depend on a disc number or avoid possible loops or failed logic.

A column of disc 1 makes me feel there should be a disc 2, when most albums are a single disc only.

NOTE: This was originally at https://github.com/beetbox/beets/pull/5909 but I managed to screw up the source/patch repo while trying to bring things up to date with master, so I just recreated the final version of it.  I've also changed the parameter to remove the 'number' which I'm told won't always translate well (and technically the disc could be a letter such as 'A').

## Description

This addresses concerns such as https://beets-users.narkive.com/fxh0cgmC/beets-remove-disc-field-for-single-disc-albums and accomplishes things like https://github.com/beetbox/beets/discussions/5670 but in tags instead of just in the path.

## To Do

## Summary by Sourcery

Introduce a zero_disc_if_single_disc configuration to the zero plugin, skipping the disc tag on single-disc albums when writing tags.

New Features:
- Add zero_disc_if_single_disc boolean option to the zero plugin to omit the disc tag when only one disc is present.

Documentation:
- Update plugin documentation and changelog to describe the zero_disc_if_single_disc option.

### DIFF
--- a/beetsplug/zero.py
+++ b/beetsplug/zero.py
@@ -41,6 +41,7 @@ class ZeroPlugin(BeetsPlugin):
                 "fields": [],
                 "keep_fields": [],
                 "update_database": False,
+                "zero_disc_if_single_disc": False,
             }
         )
 
@@ -123,8 +124,12 @@ class ZeroPlugin(BeetsPlugin):
         """
         fields_set = False
 
+        if "disc" in tags and self.config["zero_disc_if_single_disc"].get(bool) and item.disctotal == 1:
+            self._log.debug("disc: {.disc} -> None", item)
+            tags["disc"] = None
+
         if not self.fields_to_progs:
-            self._log.warning("no fields, nothing to do")
+            self._log.warning("no fields list to remove")
             return False
 
         for field, progs in self.fields_to_progs.items():
--- a/docs/plugins/zero.rst
+++ b/docs/plugins/zero.rst
@@ -31,6 +31,8 @@ to nullify and the conditions for nullifying them:
   ``keep_fields``---not both!
 - To conditionally filter a field, use ``field: [regexp, regexp]`` to specify
   regular expressions.
+- Set ``zero_disc_if_single_disc`` to ``True`` to zero the disc number field
+  only if the album contains a disctotal count and is a single disc.
 - By default this plugin only affects files' tags; the beets database is left
   unchanged. To update the tags in the database, set the ``update_database``
   option to true.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,6 +8,9 @@ Unreleased
 ----------
 
 New features:
+* :doc:`plugins/zero`: Add new configuration option,
+  ``zero_disc_if_single_disc``, to allow zeroing the disc number on
+  write for single-disc albums. Defaults to False.
 
 Bug fixes:
 


michaelkrieger:
@semohr Thank you for the suggestion.  It has been implemented changing the parameter name to omit_single_disc (vs previously zero_disc_if_single_disc).

Also, with this update, I have:
* Add return of 'fields_set' so that, if triggered by the command line `beets zero`, it will still effect the item.write.
* Added some tests for completeness.  This tries a permutation both as the only option and along with a field-removal.

docs/plugins/zero.rst
```diff
@@ -31,6 +31,8 @@ to nullify and the conditions for nullifying them:
   ``keep_fields``---not both!
 - To conditionally filter a field, use ``field: [regexp, regexp]`` to specify
   regular expressions.
+- Set ``omit_single_disc`` to ``True`` to zero the disc number field only if the
```
semohr: How about

```rst
Set ``omit_single_disc`` to ``True`` to omit writing the ``disc`` number for albums with only a single disc
(``disctotal == 1``). By default, beets will number the disc even if the album contains only one disc in total.
```

test/plugins/test_zero.py
```diff
@@ -249,6 +249,54 @@ def test_fields_removes_preserved_tags(self):
 
         assert "id" not in z.fields_to_progs
 
+    def test_omit_single_disc_with_tags_single(self):
```
semohr, line 252: The added tests are all pretty similar. You could consider using `@pytest.mark.paramterize`.

beetsplug/zero.py
```diff
@@ -123,9 +124,15 @@ def set_fields(self, item, tags):
...
+            if item.disctotal == 1:
+                fields_set = True
+                self._log.debug("disc: {.disc} -> None", item)
+                tags["disc"] = None
+
         if not self.fields_to_progs:
-            self._log.warning("no fields, nothing to do")
-            return False
+            self._log.warning("no fields list to remove")
+            return fields_set
```
semohr: I think we can remove the return here. Should make thinking about the code flow easier. If `fields_to_progs` is empty the iteration below is not done anyways. We should keep the log message tho.

semohr (changes requested):

beetsplug/zero.py
michaelkrieger: Change made.  Agreed.

docs/plugins/zero.rst
michaelkrieger: Change made.  Agreed.

test/plugins/test_zero.py
michaelkrieger: Agree that they could be combined. Tried to use parametrize and I must be missing something as couldn't get it to not throw errors (including putting it in its own class and other approaches).  Here is what I had in case it's a starting point.  Otherwise just left it as is for now and unless it's an easy fix, would rather just remove the tests. #regretsPuttingInTests.

```
    @pytest.mark.parametrize(
        "disc_value, disctotal_value, fields_value, result_value",
        [
            (1, 1, ["comments"], 0),
            (1, 4, ["comments"], 1),
            (1, 1, False, 0),
            (1, 4, False, 1),
        ],
    )
    def test_omit_single_disc(self, disc_value, disctotal_value, fields_value, result_value):
        item = self.add_item_fixture(
            disc=disc_value, disctotal=disctotal_value, comments="test comment"
        )
        item.write()
        with self.configure_plugin(
            {"omit_single_disc": True, "fields": fields_value}
        ):
            item.write()

        mf = MediaFile(syspath(item.path))
        assert mf.disc == result_value
```

semohr: Oh yeah sorry! We are currently migrating to pytest and it seem like the `PluginTestCase` base class still uses the UnitTest pattern.

Something like this should work:
```python
class TestOmitSingleDisc(PluginMixin, TestHelper):
    plugin = "zero"

    @pytest.fixture(autouse=True)
    def setup(self):
        self.setup_beets()
        yield
        self.teardown_beets()

    @pytest.mark.parametrize(
        "disc_value, disctotal_value, fields_value, result_value",
        [
            (1, 1, ["comments"], 0),
            (1, 4, ["comments"], 1),
            (1, 1, False, 0),
            (1, 4, False, 1),
        ],
    )
    def test_omit_single_disc(
        self, disc_value, disctotal_value, fields_value, result_value
    ):
        item = self.add_item_fixture(
            disc=disc_value, disctotal=disctotal_value, comments="test comment"
        )
        item.write()
        with self.configure_plugin(
            {"omit_single_disc": True, "fields": fields_value}
        ):
            item.write()

        mf = MediaFile(syspath(item.path))
        assert mf.disc == result_value
```

michaelkrieger:
Tried to implement and just got more problems.  Left it as is for now (could be cleaned up once the migration of tests gets done for these plugins) _unless you'd rather just remove the tests entirely._

Ideally just want to merge this important feature and move forward so it doesn't miss out on another release of Beets and aren't chasing other people's merges.

michaelkrieger:
Another release cycle, (2.5.0) another time this **feature-complete** PR has been passed by.  Can we please merge it?  It works.  All requested changes are made.  Simplifying the tests into fewer functions can be done later or the tests outright removed if you'd prefer.  @semohr

PR #6016, opened 2025-09-16
Remove beets.mediafile deprecation warning

semohr:
I propose we remove the deprecation warning for `beets.mediafile`. This warning was introduced in version 1.6.0.

While we could schedule the removal for version 3.0.0, it should be safe to remove it now since the feature has been deprecated for over three years.

ref: #4263


JOJ0 (approved):
Yes! it's time! :-) Thanks!

snejus (changes requested):
We need a major version increase in order to remove it.

semohr:
Added 3.0.0 as milestone. Might help us keep track of it.

PR #6008, opened 2025-09-14
lastgenre: Add --pretend option for previewing genre changes

arsaboo:
## Description

Introduce a `--pretend` option to the lastgenre plugin, allowing users to preview genre changes without making any modifications to their library. This feature enhances user control by showing potential changes before they are applied.

## To Do

- [x] Documentation. (If you've added a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)
- [X] Changelog. (Add an entry to `docs/changelog.rst` to the bottom of one of the lists near the top of the document.)
- [X] Tests. (Very much encouraged but not strictly required.)

cc: @JOJ0

## Summary by Sourcery

Introduce a --pretend option to the lastgenre plugin to allow users to preview genre assignments without altering their library files.

New Features:
- Add --pretend flag to preview genre changes without modifying files

Enhancements:
- Conditionally skip updating item and album genres and file writes when --pretend is enabled

Documentation:
- Document the --pretend option in the plugin documentation and update the changelog

PR #6021, opened 2025-09-17
lastgenre: Refactor genre applying and pretend mode

JOJ0:
## Description

Shorten final album/genre-apply code. This was longish code already and since the `--pretend` option is here it is even longer, thus I'd like to smarten/shorten that. Also behavior is changed so that `--pretend` is automatically enabling force mode (`-p` or `-f` is what the user usually expects to being the opposite of each other, but we can discuss that of course)

## To Do

- [x] ~Documentation.~ Not required. No behavioral changes
- [x] Changelog.
- [x] Test refactored to pytest

### DIFF
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -468,7 +468,9 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         """Fetch genre and log it."""
         self._log.info(str(obj))
         obj.genre, label = self._get_genre(obj)
-        self._log.info("Resolved ({}): {}", label, obj.genre)
+        self._log.debug("Resolved ({}): {}", label, obj.genre)
+
+        ui.show_model_changes(obj, fields=["genre"], print_obj=False)
 
     @singledispatchmethod
     def _process(self, obj: LibModel, write: bool) -> None:
--- a/test/plugins/test_lastgenre.py
+++ b/test/plugins/test_lastgenre.py
@@ -152,13 +152,10 @@ class LastGenrePluginTest(PluginTestCase):
             raise AssertionError("Unexpected store call")
 
         # Verify that try_write was never called (file operations skipped)
-        with (
-            patch("beetsplug.lastgenre.Item.store", unexpected_store),
-            self.assertLogs() as logs,
-        ):
-            self.run_command("lastgenre", "--pretend")
+        with patch("beetsplug.lastgenre.Item.store", unexpected_store):
+            output = self.run_with_output("lastgenre", "--pretend")
 
-        assert "Mock Genre" in str(logs.output)
+        assert "Mock Genre" in output
         album.load()
         assert album.genre == "Original Genre"
         assert album.items()[0].genre == "Original Genre"
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -1078,7 +1078,9 @@ def _field_diff(field, old, old_fmt, new, new_fmt):
     return f"{oldstr} -> {newstr}"
 
 
-def show_model_changes(new, old=None, fields=None, always=False):
+def show_model_changes(
+    new, old=None, fields=None, always=False, print_obj: bool = True
+):
     """Given a Model object, print a list of changes from its pristine
     version stored in the database. Return a boolean indicating whether
     any changes were found.
@@ -1117,7 +1119,7 @@ def show_model_changes(new, old=None, fields=None, always=False):
         )
 
     # Print changes.
-    if changes or always:
+    if print_obj and (changes or always):
         print_(format(old))
     if changes:
         print_("\n".join(changes))


JOJ0:
Setting this to "ready to review" since I'd like to get this merged before any new feature I have in draft already and to get a sourcery review already..

JOJ0:
Hi  @beetbox/maintainers, this does not include documentation yet. Can I get a review on the code especially the decorator implementation? I personally love this change since the final genre-apply-code is so much shorter now (even though it also has two new features!

Definitive questions:

- Would you consider the actual behavioral change that missing track genres falling back to album genres should be configurable?
- I'm not sure about the naming of that decorator `dry_run` - I could rename it to `pretend` ok, but still it does 3 things: Log, decide whether to apply genres or just log (in a slightly different way)


I'll handle docs and changelog after this is reviewed and final decisions made, of course! Thanks in advance!

JOJ0:
@sourcery-ai guide

@sourcery-ai summary

snejus:
I think it would be a good idea to add `--pretend` flag in a separate PR, and we already have a PR open: #6008. Would you be happy to adjust this PR to __not__ add the `--pretend` flag and keep it focused on the refactor?

arsaboo:
@snejus  if it's not too much of inconvenience, I would rather have @JOJ0  make changes to last genre. They are much closer to it and have other features planned. I'm happy to close the other PR.

JOJ0:
@snejus the refactor does not really make sense without the pretend. it's easier this way.

i would like to close @arsaboo 's PR



what i can do here is remove the track to album fallback feature. 

would that be ok with you @snejus ?

JOJ0:
> @snejus the refactor does not really make sense without the pretend. it's easier this way.
> 
> i would like to close @arsaboo 's PR
> 
> what i can do here is remove the track to album fallback feature.
> 
> would that be ok with you @snejus ?

Ah let me reconsider @snejus: I see you already reviewed the pretend PR from @arsaboo, so if you think this is ready to merge soon, please go ahead you both and finishe it!

I can  work from there an adapt this PR to only make pretend conditional smarter/shorter with my decorator refactor idea here!

I think I misunderstood earlier and you ment it that way @snejus right?

JOJ0:
> > @snejus the refactor does not really make sense without the pretend. it's easier this way.
> > i would like to close @arsaboo 's PR
> > what i can do here is remove the track to album fallback feature.
> > would that be ok with you @snejus ?
> 
> Ah let me reconsider @snejus: I see you already reviewed the pretend PR from @arsaboo, so if you think this is ready to merge soon, please go ahead you both and finishe it!
> 
> I can work from there an adapt this PR to only make pretend conditional smarter/shorter with my decorator refactor idea here!
> 
> I think I misunderstood earlier and you ment it that way @snejus right?

@snejus ? Was this your intention?

snejus:
@JOJ0 indeed - it's mostly finished so it's we can finish it there and you can base your work on top of it!

JOJ0:
@sourcery-ai dismiss

JOJ0:
> * Consider adding a configuration flag to enable or disable the automatic fallback to album genre, so users can opt out of inherited genres if desired.

The feature was removed from this PR.

> * Add unit tests specifically for the fallback-to-album-genre path to ensure tracks only inherit album genres under the correct conditions.

Removed this feature from this PR. Just refactoring of genre applying/pretend handling now.

> * You might rename the `label` parameter to something like `source` for clearer alignment with the existing genre-source terminology.

Terminology changed to "logging label / stage label" a while ago already. This is fine and consistent with other parts of the code.

JOJ0:
@semohr or @snejus I pulled out the test of the unittest class to a self-contained pytest and installed pytest-mock. I didn't like the cascaded patch.object statements. Test is similar in length now (altough self-contained) but I think a bit easier to read. What do you think?

test/plugins/test_lastgenre.py
```diff
@@ -209,6 +172,52 @@ def test_sort_by_depth(self):
         assert res == ["ambient", "electronic"]
 
 
+def test_pretend_option_skips_library_updates(mocker):
```
snejus: This test can be simplified and made more clear by:

1. Keeping the test within the `LastGenrePluginTest` to remove the manual need to run `setUp` and `tearDown` (which negates `pyetst` benefits).

1. Using our `self.run_command` abstraction to run the command. Note you will need to inherit from `PluginTestCase` and define `plugin` attribute for this:
   ```py
   from beets.test.helper import PluginTestCase


   class LastGenrePluginTest(PluginTestCase):
       plugin = "lastgenre"
   ```

1. Using idiomatic `self.assertLogs()` to catch logging output.
1. Patching `Item.store` with a function that 'explodes' with a clear message. Note that I think we're supposed to test `store` rather than `write`, I think?

```py
    @patch("beets.ui.should_write", Mock(return_value=True))
    @patch(
        "beetsplug.lastgenre.LastGenrePlugin._get_genre",
        Mock(return_value=("Mock Genre", "mock stage")),
    )
    def test_pretend_option_skips_library_updates(self):
        item = self.create_item(
            album="Pretend Album",
            albumartist="Pretend Artist",
            artist="Pretend Artist",
            title="Pretend Track",
            genre="Original Genre",
        )
        album = self.lib.add_album([item])

        def unexpected_store(*_, **__):
            raise AssertionError("Unexpected store call")

        # Verify that try_write was never called (file operations skipped)
        with (
            patch("beetsplug.lastgenre.Item.store", unexpected_store),
            self.assertLogs() as logs,
        ):
            self.run_command("lastgenre", "--pretend")

        assert "Mock Genre" in str(logs.output)
        album.load()
        assert album.genre == "Original Genre"
        assert album.items()[0].genre == "Original Genre"
```

beetsplug/lastgenre/__init__.py
```diff
@@ -527,64 +566,37 @@ def commands(self):
 
         def lastgenre_func(lib, opts, args):
             write = ui.should_write()
-            pretend = getattr(opts, "pretend", False)
             self.config.set_args(opts)
+            if opts.pretend:
```
snejus: What is it for?

beetsplug/lastgenre/__init__.py
```diff
@@ -527,64 +566,37 @@ def commands(self):
...
-                            if not pretend:
-                                item.genre = item_genre
-                                item.store()
-
-                        if write and not pretend:
-                            item.try_write()
+                            item_genre, label = self._get_genre(item)
+
+                            if not item_genre:
+                                self._log.info(
```
snejus: Shouldn't we have the same logic in two other places?

1. the singleton track on line 599
2. the equivalent operation for album tracks under `imported`, line 617

beetsplug/lastgenre/__init__.py
```diff
@@ -527,64 +566,37 @@ def commands(self):
...
-                        src,
-                        item_genre,
-                    )
-                    if not pretend:
-                        item.genre = item_genre
-                        item.store()
-                    if write and not pretend:
-                        item.try_write()
+                    singleton_genre, label = self._get_genre(item)
+                    self._apply_item_genre(item, label, singleton_genre)
```
snejus: Why are we not writing singleton like we write album items?

beetsplug/lastgenre/__init__.py
```diff
@@ -593,34 +605,21 @@ def imported(self, session, task):
...
-                    self._log.debug(
-                        'genre for track "{0.title}" ({1}): {0.genre}',
-                        item,
-                        src,
-                    )
-                    item.store()
-            # Store the album genre and inherit to tracks.
-            else:
-                album.store()
+                    item_genre, label = self._get_genre(item)
```
snejus: To ensure that album tracks and singletons are processed consistently regardless of whether it's run by the command or the import, I'd suggest defining

```py
    def _process_item(self, item: Item, write: bool = False):
        genre, label = self._get_genre(item)

        if genre:
            self._apply_item_genre(item, label, genre)
            if write and not self.config["pretend"]:
                item.try_write()
        else:
            self._log.info('No genre found for track "{.title}"', item)

```

beetsplug/lastgenre/__init__.py
```diff
@@ -527,64 +566,37 @@ def commands(self):
...
+
+                            if not item_genre:
+                                self._log.info(
+                                    'No genre found for track "{0.title}"',
+                                    item,
+                                )
+                            else:
+                                self._apply_item_genre(item, label, item_genre)
+                                if write:
+                                    item.try_write()
```
snejus: Seems like files will get written even if `--pretend` is given?

Maybe it should be 
```suggestion
                                if write and self.config["pretend"]:
                                    item.try_write()
```

beetsplug/lastgenre/__init__.py
```diff
@@ -593,34 +605,21 @@ def imported(self, session, task):
         """Event hook called when an import task finishes."""
         if task.is_album:
             album = task.album
-            album.genre, src = self._get_genre(album)
-            self._log.debug(
-                'genre for album "{0.album}" ({1}): {0.genre}', album, src
-            )
+            album_genre, label = self._get_genre(album)
+            self._apply_album_genre(album, label, album_genre)
```
snejus: Seems like album logic is the same in `lastgenre_func` and `imported`, so you can probably define

```py
    def _process_album(self, album: Album, write: bool = False):
        album_genre, label = self._get_genre(album)
        self._apply_album_genre(album, label, album_genre)

        # If we're using track-level sources, store the album genre only (this
        # happened in _apply_album_genre already), then also look up individual
        # track genres.
        if "track" in self.sources:
            for item in album.items():
                self._process_item(item, write=write)
```

snejus (changes requested):
Added a comment regarding making the test simpler and a couple of suggestions how to deduplicate item and album logic further

snejus:
@JOJ0 shall I commit what I've got here? I refactored things exactly as I suggested as I reviewed the PR.

JOJ0:
I was about to finally do it this week. Was again out of town this weekend. Yes please, it would help speed up things. Thank you!!! 🙏

snejus:
I tried to make the output a bit more user-friendly and used
1. `str(obj)` to log the object using the user configured `format_item` / `format_album` values
2. `ui.show_model_changes` to show the change in `genre` with colors

#### Non-verbose output
image

#### Verbose output
image

JOJ0 (commented):
Thanks @snejus! Let's merge it. Can't press approve since I'm the initial author but LGTM. Great idea with the colors and the singledispatch 🤩

snejus:
🫡

PR #6022, opened 2025-09-17
Refactor `test_plugin.py`: Streamline plugin tests and migrate to pytest

semohr:
## Overview

This PR addresses historical testing debt in `test_plugins.py` by standardizing plugin test infrastructure.
- `PluginTestCase` is now using pytest, old test mixing is still available via `PluginUnitTestCase` (unittest-based)
- Enhances `PluginMixin` with programmatic plugin registration/unregistration while still working through standard `load_plugins`, improving test isolation and avoiding fragile disk/module hacks.
- Simplifies and modernizes individual tests in `test_plugins.py` where possible, reducing boilerplate and making future plugin tests easier to extend.

There is still the `PromptChoicesTest` left which I have not touched, tbh I have no idea why this test is located in `test_plugins.py` at all...

## Coverage
```
# before
Name               Stmts   Miss Branch BrPart   Cover   Missing
---------------------------------------------------------------
beets/plugins.py     228     57     70      6  70.13%   119, 131-136, 184->186, 186->188, 188->191, 279-282, 299-305, 315-321, 352, 366-387, 473-475, 489, 523-529, 544-547, 575-578, 595-607
---------------------------------------------------------------
TOTAL                228     57     70      6  70.13%

# after
Name               Stmts   Miss Branch BrPart   Cover   Missing
---------------------------------------------------------------
beets/plugins.py     228     45     70      8  74.83%   131-136, 184->186, 186->188, 188->191, 279-282, 286->exit, 299-305, 315-321, 352, 372->387, 473-475, 489, 523-529, 544-547, 575-578, 595-607
---------------------------------------------------------------
TOTAL                228     45     70      8  74.83%
```

## Open questions / decisions

Since test_plugins.py has effectively been rewritten, should the legacy
Copyright 2016, Thomas Scholtes notice be removed?

Current authorship distribution per git blame:
```
git blame --line-porcelain ./test/test_plugins.py | grep '^author ' | sort | uniq -c | sort -nr
    341 author Sebastian Mohr
     76 author Šarūnas Nejus
     51 author Diego Moreda
     31 author Serene-Arc
     22 author Thomas Scholtes
     17 author Bruno Cauet
     15 author Malte Ried
      7 author Maxr1998
      6 author Andrew Rogl
      2 author pscn
      1 author Lars Kruse
      1 author Johnny Robeson
      1 author Jack Wilsdon
      1 author Carl Suster
```

Currently the `PluginTestCase` loads the plugin before each function and unloads it after each function, this is most likely not what we want as this comes with quite a bit of overhead. We might need some adjustments to run as class level fixture instead.

## Long term direction

This PR is an early step toward modernizing the plugin test suite and reducing reliance on legacy unittest-style patterns, with the goal of making pytest the primary testing model going forward.

### DIFF
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -641,7 +641,7 @@ def get_available_plugins():
     ]
 
 
-class PluginTestCasePytest(PluginMixin, TestHelper):
+class PluginTest(PluginMixin, TestHelper):
     @pytest.fixture(autouse=True)
     def _setup_teardown(self):
         self.setup_beets()
@@ -650,7 +650,7 @@ class PluginTestCasePytest(PluginMixin, TestHelper):
 
 
 class PluginTestCase(PluginMixin, BeetsTestCase):
-    """DEPRECATED: Use PluginTestCasePytest instead for new code!"""
+    """DEPRECATED: Use PluginTestCase instead for new code using pytest!"""
 
     pass
 
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -507,7 +507,7 @@ class PluginMixin(TestHelper):
     def teardown_beets(self):
         super().teardown_beets()
         self.unload_plugins()
-        self.unregister_plugin(self.plugin)
+        self.unregister_plugin(name=self.plugin)
 
     @staticmethod
     def register_plugin(
@@ -555,12 +555,6 @@ class PluginMixin(TestHelper):
             if hasattr(parent_pkg, name):
                 delattr(parent_pkg, name)
 
-    def unregister_all_plugins(self) -> None:
-        # Remove plugin modules from sys.modules
-        for mod in list(sys.modules):
-            if mod.startswith("beetsplug."):
-                del sys.modules[mod]
-
     def load_plugins(self, *plugins: str) -> None:
         """Load and initialize plugins by names.
 
@@ -600,17 +594,26 @@ class PluginMixin(TestHelper):
         raise ValueError(f"No plugin found with name {name}")
 
     @contextmanager
-    def plugins(self, *plugins: tuple[str, type[beets.plugins.BeetsPlugin]]):
+    def plugins(
+        self, *plugins: tuple[str, type[beets.plugins.BeetsPlugin]] | str
+    ):
         """Context manager to register and load multiple plugins."""
         self.unload_plugins()
-        for name, plugin_type in plugins:
-            self.register_plugin(plugin_type, name)
-        self.load_plugins(*(name for name, _ in plugins))
+
+        names = []
+        for plug in plugins:
+            if isinstance(plug, str):
+                names.append(plug)
+            else:
+                names.append(plug[0])
+                self.register_plugin(plug[1], plug[0])
+        self.load_plugins(*names)
 
         yield
 
         self.unload_plugins()
-        self.unregister_all_plugins()
+        for name in names:
+            self.unregister_plugin(name)
 
     @contextmanager
     def configure_plugin(self, config: Any):
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -25,11 +25,14 @@ information or mock the environment.
 
 from __future__ import annotations
 
+import importlib
 import os
 import os.path
+import pkgutil
 import shutil
 import subprocess
 import sys
+import types
 import unittest
 from contextlib import contextmanager
 from dataclasses import dataclass
@@ -50,6 +53,7 @@ from beets import importer, logging, util
 from beets.autotag.hooks import AlbumInfo, TrackInfo
 from beets.importer import ImportSession
 from beets.library import Item, Library
+from beets.plugins import PLUGIN_NAMESPACE
 from beets.test import _common
 from beets.ui.commands.import_.session import TerminalImportSession
 from beets.util import (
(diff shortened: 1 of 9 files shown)


semohr:
@beetbox/maintainers If we are in general fine with these changes, I can start with migrating the plugins from `PluginUnitTestCase` to `PluginTestCase` either in this or incrementally in the following PRs.

This was sitting for quite a while but I just got around rebasing it.

beets/test/helper.py
```diff
@@ -467,23 +472,130 @@ def setUp(self):
         self.i = _common.item(self.lib)
 
 
-class PluginMixin(ConfigMixin):
-    plugin: ClassVar[str]
+class PluginMixin(TestHelper):
+    """A mixin that handles plugin loading, unloading and registration.
```
henry-oberholtzer, line 432: I love this extended docstring!

henry-oberholtzer (commented):
This looks good to me! I'm not seeing anything that stands out as concerning to me, looks like a good step & a nice refactor. Great work with the extended coverage of plugins too!

I'm happy to take on reworking tests for a few plugin tests as well to refactor switch more to being pytest based.

As for authorship, I think at the very least you can definitely put your name at the top too - I don't know what the original form of the test suite was, but we've all been building off it, so I don't know if removing an author would be the way to go.

Some of this test patching stuff does feel a little beyond my expertise, so I'll just leave this as a comment in case anyone else wants to weigh in.

semohr:
> As for authorship, I think at the very least you can definitely put your name at the top too - I don't know what the original form of the test suite was, but we've all been building off it, so I don't know if removing an author would be the way to go.

I would like to get rid of it entierly tbh. Im not too fond of the additional notice in every file as it is bundled with every release. Effectively noone adds their name to these notices and they do drift significantly against the acctual ownership. We do have a general license and a git log for this imo.

> I'm happy to take on reworking tests for a few plugin tests as well to refactor switch more to being pytest based.

I would be happy to have you onboard here, I guess migration should be mostly straightforward.

henry-oberholtzer:
Git log is a point I hadn't considered - I think it'd be fine to remove it or replace with your name in that case.

semohr:
@snejus Any thought in general on this PR?

beets/test/helper.py
```diff
@@ -467,23 +472,130 @@ def setUp(self):
...
+                yield
+                self.teardown_beets()
+
+            def test_something(self):
+                ...
+
+    This will load the plugin named `myplugin` before each test and unload it
+    afterwards. This requires a `myplugin.py` file in the `beetsplug` namespace.
+
+    If you need to register a plugin class directly, set the `plugin_type` class
```
snejus, line 457: We've got two types of plugin tests:

1. Specific plugins: each plugin under `beetsplug`, with tests under `test/plugins/*`
   `PluginMixin` was designed as a helper for these tests: it's got a simple task: load and unload the plugin, and provide a method to adjust the configuration.

2. `beets/plugins.py` or `BeetsPlugin`, with tests in `test/test_plugins.py`.
   These tests are much more complicated since they test the `BeetsPlugin` interface and depend on virtual plugin creation. 

Thus, I think, we should keep `PluginMixin` simple to focus on its original use case. Since the rest of the functionality (multiple and virtual plugins) is only applicable within `test/test_plugins.py`, we should extend it there:

```py
# test/test_plugins.py

class BeetsPluginMixin(PluginMixin):
    plugin_type: ...
    def register_plugin
    def plugins...
    ...
```

beets/test/helper.py
```diff
@@ -492,29 +604,99 @@ def load_plugins(self, *plugins: str) -> None:
...
+    namespace_pkg = importlib.import_module("beetsplug")
+
+    return [
+        m.name
+        for m in pkgutil.iter_modules(namespace_pkg.__path__)
+        if not m.name.startswith("_")
+    ]
+
+
+class PluginTestCase(PluginMixin, TestHelper):
```
snejus: Let's remove __Case__ from the name to decouple it from `unittest` completely

```suggestion
class PluginTest(PluginMixin, TestHelper):
```

beets/test/helper.py
```diff
@@ -492,29 +604,99 @@ def load_plugins(self, *plugins: str) -> None:
...
+    @pytest.fixture(autouse=True)
+    def _setup_teardown(self):
+        self.setup_beets()
+        yield
+        self.teardown_beets()
+
 
+class PluginUnitTestCase(PluginMixin, BeetsTestCase):
```
snejus: Given my comment above you can keep the previous name here

```suggestion
class PluginTestCase(PluginMixin, BeetsTestCase):
```

snejus (changes requested):

beets/test/helper.py
semohr: Yeah, that makes sense! Thinking of it as a virtual plugin is a nice approach. We could split out the register functionality into a `VirtualPluginMixin(PluginMixin)` 👍. I still think it makes sense to keep it in the helpers file, though, it seems generally useful and could be handy for external plugin developers as well.

(I can see us use that in beets-flask if it is shipped with beets, there is no easy way to import something from the tests files afaik)

snejus:
@semohr is this one still relevant?

semohr:
I don't think it is. Might be some things we can salvage from here but would need some effort.

PR #6032, opened 2025-09-19
Improved `beets/logging.py` typing

semohr:
This PR enhances `beets/logging.py` with improved typing and tests:

* `getLogger` now returns the precise logger type (`BeetsLogger` or `RootLogger`).
* Tests use `pytest` and `parametrize` for more concise and readable coverage.

### DIFF
--- a/beets/logging.py
+++ b/beets/logging.py
@@ -37,9 +37,7 @@ from logging import (
     RootLogger,
     StreamHandler,
 )
-from typing import TYPE_CHECKING, Any, Mapping, TypeVar, overload
-
-from typing_extensions import TypeAlias
+from typing import TYPE_CHECKING, Any, Mapping, TypeVar, Union, overload
 
 __all__ = [
     "DEBUG",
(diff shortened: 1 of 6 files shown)


beets/logging.py
```diff
@@ -83,40 +96,45 @@ class StrFormatLogger(Logger):
...
+        stacklevel: int = 1,
         **kwargs,
     ):
         """Log msg.format(*args, **kwargs)"""
-        m = self._LogMessage(msg, args, kwargs)
 
-        stacklevel = kwargs.pop("stacklevel", 1)
-        stacklevel = {"stacklevel": stacklevel}
+        if isinstance(msg, str):
+            msg = self._LogMessage(msg, args, kwargs)
```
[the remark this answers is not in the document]
semohr: This is how it should work tho! Only strings may have `%` style string formatting. For all other types we use the default log message.

beets/logging.py
```diff
@@ -83,40 +91,45 @@ class StrFormatLogger(Logger):
...
-        level,
-        msg,
-        args,
-        exc_info=None,
-        extra=None,
-        stack_info=False,
+        level: int,
+        msg: object,
+        args: logging._ArgsType = (),
+        exc_info: logging._ExcInfoType = None,
```
snejus: `_ArgsType` and `_ExcInfoType` are not intended to be used outside of `logging` module (original 😅) implementation.

My IDE does not jump into them nor resolve their definitions, so they are a mystery to me. Can you define them explicitly in this module, please?

beets/logging.py
```diff
@@ -83,40 +91,45 @@ class StrFormatLogger(Logger):
...
         self,
-        level,
-        msg,
-        args,
-        exc_info=None,
-        extra=None,
-        stack_info=False,
+        level: int,
+        msg: object,
+        args: logging._ArgsType = (),
```
snejus: `args` doesn't have a default in the base definition?

beets/logging.py
semohr: The types are resolved in [typeshed](https://github.com/python/typeshed/blob/main/stdlib/logging/__init__.pyi). My IDE also has an issue with that, but it works with mypy. 

Did not want to copy the types initially but they are now included in the latest commit.

beets/logging.py
semohr: Indeed, shouldn't matter too much but removed it :shrug:

PR #6040, opened 2025-09-23
Discogs featured artist fix

henry-oberholtzer:
## Description

Fixes #6038 - Appends featured artists in the extraartists field to the artist tag, similar to the MusicBrainz plugin. Works well with ftintitle for consistency as well.

## To Do

May need adjustment for common artist delimiters used - but appears to match the MusicBrainz standard at the moment.

- [x] Documentation. (If you've added a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)
- [x] Changelog. (Add an entry to `docs/changelog.rst` to the bottom of one of the lists near the top of the document.)
- [x] Tests. (Very much encouraged but not strictly required.)

Issue #3354, opened 2019-08-30
discogs: consider use of ANV (Artist Name Variation) to set artist_credit


kergoth:
See https://support.discogs.com/hc/en-us/articles/360005054753-Database-Guidelines-2-Artist#Artist_Name_Variation_ANV. There's an `anv` field in the `artists` and `extraartists`/`credits` lists as returned by the API.

sampsyo:
Interesting idea! Could you provide a few examples to make the case that this is the right data to use?

I notice that the style guide says:

> 2.5.3. ANVs should not be used if the artist uses the name to differentiate their work, in this case an Alias is more appropriate.

> 2.5.4. ANVs should not be used for any fundamental name changes, such as surname changes. In these cases, an Alias is more appropriate. If in doubt, do not create an ANV.

But I thought MusicBrainz credits would include things like differentiated artist names and surname changes.

kergoth:
Track 2 on the Amazon Groove album. See https://musicbrainz.org/release/a32deed2-088c-41e4-a23d-72d9af74dac4 
 and https://musicbrainz.org/ws/2/release/a32deed2-088c-41e4-a23d-72d9af74dac4?inc=aliases+artist-credits+labels+discids+recordings, track 2, which shows the current artist name as 'Friendly', but credited as 'Andrew Friendly' on the album.

In the [discogs](https://www.discogs.com/the-amazon-groove/release/2088224) info, the name is 'Friendly', the anv is 'Andrew Friendly', which aligns with musicbrainz as artist and artist_credit, respectively.

Given this, I think we should be setting `_credit` fields from `anv`.

kergoth:
There's a band named This Town Needs Guns that changed their name after the release of a certain album to TTNG. MusicBrainz says the artist is TTNG and the credit is 'This Town Needs Guns' on the older albums. Discogs still has the artist entry as 'This Town Needs Guns', but indicates the rename in the description and under name variations, and on newer albums the ANV is set to TTNG.

This shows up as a 'name variation' in the API for the artist, however, not as an alias, so this may just be a slightly larger delta but still considered a variation on the same name due to the acronym.

Just another data point. I'm looking for a more substantial performance under a different name for additional data, but this is another example of ANV lining up with artist_credit. I'm going to find another example or two before I submit the pull request.

henry-oberholtzer:
[...]

PR #6050, opened 2025-09-27
Discogs: artist name variation support + artist credit support.

henry-oberholtzer:
## Description

Implements #3354, updates #6040

The name variations provided by Discogs can now be used by the auto-tagger through 3 config options.
By default, the plugin will write the variation to the tag `artist_credit`, but through three config options can also write it
to the album artist tag, the track's artist tag, or any combination of the three.

This PR contains a small addition on #6040, letting the string used to join featured artists be customized.

The new configuration options available, and their defaults, are as follows:

```yaml
discogs:
  featured_string: "Feat."
  anv:
    artist_credit: True
    album_artist: False
    artist: False

```

## To Do

Configuration options could be renamed, but I think they're decently self documenting at the moment.


beetsplug/discogs.py
```diff
@@ -442,7 +469,7 @@
         else:
             return None
 
-    def get_tracks(self, tracklist, album_artist, album_artist_id):
+    def get_tracks(self, tracklist, album_artist_data):
```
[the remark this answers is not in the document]
henry-oberholtzer: this will come with time sourcery-ai, we're getting there, don't worry.

beetsplug/discogs.py
```diff
@@ -376,9 +404,7 @@
         # Additional cleanups
         # (various artists name, catalog number, media, disambiguation).
         if va:
-            artist = config["va_name"].as_str()
-        else:
-            artist = self.strip_disambiguation(artist)
+            album_artist, artist_credit = config["va_name"].as_str()
```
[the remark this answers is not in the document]
henry-oberholtzer: Oops! Fixed.

docs/plugins/discogs.rst
```diff
@@ -112,6 +112,14 @@ Other configurations available under ``discogs:`` are:
 - **strip_disambiguation**: Discogs uses strings like ``"(4)"`` to mark distinct
   artists and labels with the same name. If you'd like to use the discogs
   disambiguation in your tags, you can disable it. Default: ``True``
+- **featured_string**: Configure the string used for noting featured artists.
```
semohr, line 115: Could we make it clearer in the docs what an **Artist Name Variation (ANV)** is? Right now, I had to follow the linked PRs and read [this Discogs guideline](https://support.discogs.com/hc/en-us/articles/360005054753-Database-Guidelines-2-Artist#Artist_Name_Variation_ANV) before I fully understood the context.

I think adding a short explanation and a concrete example would help users a lot. For instance, something like:

> *An Artist Name Variation (ANV) is an alternate way an artist’s name appears on a release. For example, “The Beatles” might be listed as “Beatles” or “Beatles, The” on certain records.*

That way, people don’t need to dig into external links just to understand what’s meant here. From just the config option descriptions, I suspect quite a few users won’t take the extra step to research it.

henry-oberholtzer: Great call out! I've been too deep a discogs user for years to have caught that. I'll update that accordingly.

beetsplug/discogs.py
```diff
@@ -636,38 +669,56 @@ def strip_disambiguation(self, text: str) -> str:
...
-        extraartists = track.get("extraartists", [])
-        featured = [
-            artist["name"]
-            for artist in extraartists
-            if "Featuring" in artist["role"]
-        ]
-        if featured:
-            artist = f"{artist} feat. {', '.join(featured)}"
-        artist = self.strip_disambiguation(artist)
+        if extraartists := track.get("extraartists", []):
```
semohr, line 771: I have noticed you opted to use the featured string to join the artist. Have you thought about using the `artists` field instead?

henry-oberholtzer: Oh wow, I had missed those plural fields since they aren't currently filled by this plugin!

Since featured artist was already applied in PR #6040, I'll make a new PR to test & utilize the `artists` and `artists_credit` field for the album and tracks in the discogs plugin, inline with how the musicbrainz plug treats them.

semohr: Sounds good! I think our support for multi value field is also pretty new in the context of beets lifetime :laughing:

beetsplug/discogs.py
```diff
@@ -97,8 +97,12 @@ def __init__(self):
                 "user_token": "",
                 "separator": ", ",
                 "index_tracks": False,
+                "featured_string": "Feat.",
                 "append_style_genre": False,
                 "strip_disambiguation": True,
+                "album_artist_anv": False,
```
semohr: It is not a requirement but nesting these options might be a bit more clear. E.g. `config["anv"]["track_artist"]`.

beetsplug/discogs.py
henry-oberholtzer: Definitely good to have as far as making tag writing dynamic goes.

beetsplug/discogs.py
```diff
@@ -442,7 +475,7 @@ def format(self, classification):
         else:
             return None
 
-    def get_tracks(self, tracklist, album_artist, album_artist_id):
+    def get_tracks(self, tracklist, album_artist_data):
```
semohr: Can we get some typehints for the function definition or would you recon that to be rather difficult? Same for the `get_track_info` function.

While not a necessity it helps with maintenance :D

docs/plugins/discogs.rst
```diff
@@ -96,6 +96,11 @@ whereas with ``index_tracks`` disabled you'd get:
 
 This option is useful when importing classical music.
 
+### Handling Artist Name Variations (ANVs)
+
+An ANV is an alternate way that an artist may be credited on a release. If the
+band name changes or is misspelled on different releases. The artist name ac
```
semohr: I think there are some issues with the docs here 🙃

beetsplug/discogs.py
henry-oberholtzer: I'll get those applied - I didn't add them since the initial function didn't have them, but they really should have type hints.

beetsplug/discogs.py
henry-oberholtzer: That was my initial thought but I couldn't find any other plugins / examples of nested config so wasn't sure if that was something we could do. I'll switch it to that!

beetsplug/discogs.py
semohr: Would be highly appreciated if you could add some!

beetsplug/discogs.py
semohr: You can have a look at the `convert` plugin. I think it uses nested configs.

PR #6052, opened 2025-09-29
Make musicbrainz plugin talk to musicbrainz directly

snejus:
This PR refactors the MusicBrainz plugin implementation by replacing the `musicbrainzngs` library with direct HTTP API calls using `requests` and `requests-ratelimiter`.

**Key Changes:**

- **New utilities module**: Added `beetsplug/_utils/requests.py` with `TimeoutSession` class and HTTP error handling (`HTTPNotFoundError`, `CaptchaError`)
- **MusicBrainz API rewrite**: Replaced `musicbrainzngs` dependency with custom `MusicBrainzAPI` class using direct HTTP requests
- **Rate limiting**: Integrated `requests-ratelimiter` for API rate limiting instead of `musicbrainzngs.set_rate_limit()`
- **Data structure updates**: Updated field names to match MusicBrainz JSON API v2 format (e.g., `medium-list` → `media`, `track-list` → `tracks`)
- **Dependency management**: 
  - Made `musicbrainzngs` optional and added it to plugin-specific extras (`listenbrainz`, `mbcollection`, `missing`, `parentwork`). Updated plugin docs accordingly.
  - Made `requests` a required dependency to ensure backwards compatibility (ideally, we would make it an optional dependency under `musicbrainz` extra).
- **Error handling**: Simplified error handling by removing `MusicBrainzAPIError` wrapper class

**Benefits:**
- Direct control over HTTP requests
- Consistent rate limiting across all network requests
- Better alignment with modern MusicBrainz API responses

The changes maintain backward compatibility while modernizing the underlying implementation.

Fixes #5553
Fixes #5095

Issue #4788, opened 2023-05-12
Extending REIMPORT_FRESH_FIELDS from plugins


arsaboo:
@JOJ0 looks like there are still some issues with reimports. 

I am reimporting with [youtube](https://github.com/arsaboo/beets-youtube) and here are the relevant logs:

```
Reimported album 1. Preserving attribute ['added']. Path: /home/arsaboo/shared/music/Tu Jhoothi Main Makkaar (2023)
Reimported album 1. Preserving flexible attributes ['data_source']. Path: /home/arsaboo/shared/music/Tu Jhoothi Main Makkaar (2023)
Reimported item 1. Preserving attribute ['added']. Path: /home/arsaboo/shared/music/Tu Jhoothi Main Makkaar (2023)/Tu Jhoothi Main Makkaar (2023) - Jaadui.mp3
Reimported item 1. Preserving flexible attributes ['data_source', 'yt_album_id', 'yt_artist_id', 'yt_track_id', 'track_alt', 'cover_art_url', 'yt_updated', 'yt_views', 'spotify_album_id', 'spotify_artist_id', 'spotify_track_id', 'jiosaavn_album_id', 'jiosaavn_artist_id', 'jiosaavn_perma_url', 'jiosaavn_track_id', 'jiosaavn_updated', 'danceability', 'beats_count', 'average_loudness', 'danceable', 'gender', 'is_male', 'is_female', 'genre_rosamerica', 'voice_instrumental', 'is_voice', 'is_instrumental', 'mood_acoustic', 'mood_aggressive', 'mood_electronic', 'mood_happy', 'mood_sad', 'mood_party', 'mood_relaxed', 'mood_mirex', 'mood_mirex_cluster_1', 'mood_mirex_cluster_2', 'mood_mirex_cluster_3', 'mood_mirex_cluster_4', 'mood_mirex_cluster_5', 'spotify_track_popularity', 'spotify_danceability', 'spotify_energy', 'spotify_key', 'spotify_loudness', 'spotify_mode', 'spotify_speechiness', 'spotify_acousticness', 'spotify_instrumentalness', 'spotify_liveness', 'spotify_valence', 'spotify_tempo', 'spotify_time_signature', 'spotify_updated']. Path: /home/arsaboo/shared/music/Tu Jhoothi Main Makkaar (2023)/Tu Jhoothi Main Makkaar (2023) - Jaadui.mp3
Sending event: database_change
```

[...]

PR #4735, opened 2023-03-28
Fix data_source not updated on re-import and refactor `reimport_metadata`

JOJ0:
## Description

- Fixes #4726 
- Refactors preserving of flexible attributes in the importer's `reimport_metadata()` method.
- New logging message when flexible attributes are excluded from the preserving process and _actually_ set to a new value.
- Reformats all existing logging messages and streamline their looks.
- Adds a new unittest that checks if the data_source album field is _not_ reset to its original value on a reimport.

## To Do

- [X] ~Documentation.~
- [X] Changelog.
- [X] Tests.

PR #6064, opened 2025-10-02
Refresh flexible MusicBrainz metadata on reimport

### LINKED ISSUE #6036: import: Media type won't get updated when retagging album
### Problem

When retagging an already imported album with a MusicBrainz release id with `import -L -S <mbid>` and the id used for importing has a different media type, the new media type won't get written.

### Steps to reproduce:

1. Import an album from MusicBrainz that has releases in different formats (I tested with release [c5292a70-ed3c-3bf3-b251-8a1e1e2abf23](https://beta.musicbrainz.org/release/c5292a70-ed3c-3bf3-b251-8a1e1e2abf23))
2. Retag the same album using a release with different media, for example:

    ```sh
    beet import -L -S "4429cdd4-b0dd-3e69-bc0c-fc521c998550" album:Mezmerize
    ```

3. Check the album's media:

    ```sh
    beet ls -a -f '$media' album:Mezmerize
    ```

    Media shows up as cd even though it should be 12" vinyl.

### Setup

* OS: Windows 11
* Python version: 3.12.1
* beets version: 2.4.0

```yaml
directory: ~/mediabugtest/imported

library: ~/mediabugtest/library.db

import:
    write: yes
    move: no

plugins: musicbrainz
disabled_plugins: []
musicbrainz:
    search_limit: 5
    source_weight: 0.5
    host: musicbrainz.org
    https: no
    ratelimit: 1
    ratelimit_interval: 1
    genres: no
    external_ids:
        discogs: no
        bandcamp: no
        spotify: no
        deezer: no
        tidal: no
    extra_tags: []```

### PR DESCRIPTION
**Pull request by arsaboo**
Fixes #6036

### DIFF
--- a/beets/importer/tasks.py
+++ b/beets/importer/tasks.py
@@ -58,7 +58,6 @@ REIMPORT_FRESH_FIELDS_ITEM = [
     "deezer_album_id",
     "beatport_album_id",
     "tidal_album_id",
-    "releasegroup_id",
     "data_url",
 ]
 REIMPORT_FRESH_FIELDS_ALBUM = [*REIMPORT_FRESH_FIELDS_ITEM, "media"]

### CODE REVIEW

File: beetsplug/musicbrainz.py [hunk 1]
```diff
@@ -55,6 +59,26 @@
     "year": "date",
 }
 
+_MB_REIMPORT_FRESH_FIELDS_ALBUM = [
```
semohr: I had a quick look in the `album_info` function and was wondering if we need to put more of the fields into this list? 

Is there an easy way to know which field is a flex attribute?

arsaboo: I was thinking about the same. We want all the fields to be updated during reimport. I am wondering why this guard was introduced in the first place.

arsaboo: @semohr I asked AI this question and here's the response:

```
You don’t need to whitelist most MusicBrainz fields: the core MB fields (e.g., mb_albumid, mb_trackid, mb_releasegroupid, albumtypes, albumstatus, label, barcode, etc.) are standard model fields on beets.library.Item and beets.library.Album, so they are already overwritten on re-import. The REIMPORT_FRESH_FIELDS_* mechanism only governs flexible attributes (i.e., fields not in the schema).
```

In my plugins, everything is a flexible attribute. That is why I had to add everything there. We can merge this for now (if it looks good) and then see if we need to include other fields

semohr: > I was thinking about the same. We want all the fields to be updated during reimport. I am wondering why this guard was introduced in the first place.

Most likely historic dept :joy: An opt out approach would be better in my opinion.

snejus (changes requested):
We need a different approach - this is not a good idea

beetsplug/musicbrainz.py
```diff
@@ -55,6 +59,26 @@
...
+_MB_REIMPORT_FRESH_FIELDS_ITEM = [
+    "data_url",
+]
+
+
+def _extend_reimport_fresh_fields() -> None:
+    """Ensure MusicBrainz fields stored as flex attrs refresh on reimport."""
+    for field in _MB_REIMPORT_FRESH_FIELDS_ALBUM:
+        if field not in REIMPORT_FRESH_FIELDS_ALBUM:
+            REIMPORT_FRESH_FIELDS_ALBUM.append(field)
```
snejus: This adjusts definitions of these lists that are defined as **constants** which is a nightmare to debug.


If someone attempts to understand/debug this functionality in the future, they will see

```py
# beets/importer/tasks.py

REIMPORT_FRESH_FIELDS_ALBUM = [
    "data_source",
    "bandcamp_album_id",
    "spotify_album_id",
    "deezer_album_id",
    "beatport_album_id",
    "tidal_album_id",
]
```

And they will be left scratching their heads asking why `beets` adjusts some additional fields not present in this list.

beetsplug/musicbrainz.py
```diff
@@ -55,6 +59,26 @@
     "year": "date",
 }
 
+_MB_REIMPORT_FRESH_FIELDS_ALBUM = [
+    "media",
+    "releasegroup_id",
```
snejus: Seems like both `Album` and `Item` have `mb_releasegroupid` field present in their tables?

semohr:
Do we want to rewrite this now? Sadly the current logic is how it was intended for some reason originally.

It alsso seems a bit strange to me that we have to define these field in the first place. An opt out list would be more appropriet. 

We should also check why and when this was introduced.

snejus:
We can (and should) simply extend those list constants explicitly with new fields. They aren't specific to `musicbrainz` only - for example, `beetcamp` returns both `data_url` and `media`.

JOJ0:
> Do we want to rewrite this now? Sadly the current logic is how it was intended for some reason originally.
> 
> It alsso seems a bit strange to me that we have to define these field in the first place. An opt out list would be more appropriet. 

AFAIR this _is_ an opt out list: Everything should always be refreshed on a reimport, except not for those fields (because it would be just wrong for them)


> We should also check why and when this was introduced.

I introduced this a couple of years ago. Not much time right now to reread what was all the reasons back then but I at least looked up the PR for you guys :-) https://github.com/beetbox/beets/pull/4735

snejus:
I think I see this logic in play in my `importlog`:

```
Reimported album 9355. Not preserving flexible attributes ['data_source']. Path: /run/media/sarunas/music/Music/soma/uun
Reimported album 9355. Preserving attribute ['added']. Path: /run/media/sarunas/music/Music/soma/uun
Reimported album 9355. Preserving flexible attributes ['art_source']. Path: /run/media/sarunas/music/Music/soma/uun
Reimported item 43942. Preserving attribute ['added']. Path: /run/media/sarunas/music/Music/soma/uun/02_crueltyisnothingnew.flac
Reimported item 43942. Not preserving flexible attributes ['data_source']. Path: /run/media/sarunas/music/Music/soma/uun/02_crueltyisnothingnew.flac
Reimported item 43942. Preserving flexible attributes ['lastfm_play_count', 'bandcamp_artist_id', 'art_source', 'bandcamp_album_id', 'times_bought', 'rating', 'artpath', 'price', 'helicopta', 'play_count', 'art_url', 'bandcamp_track_id', 'currency', 'last_played']. Path: /run/media/sarunas/music/Music/soma/uun/02_crueltyisnothingnew.flac
Reimported item 43943. Preserving attribute ['added']. Path: /run/media/sarunas/music/Music/soma/uun/03_givethemrope.flac
Reimported item 43943. Not preserving flexible attributes ['data_source']. Path: /run/media/sarunas/music/Music/soma/uun/03_givethemrope.flac
Reimported item 43943. Preserving flexible attributes ['lastfm_play_count', 'skip_count', 'bandcamp_artist_id', 'art_source', 'bandcamp_album_id', 'times_bought', 'rating', 'artpath', 'price', 'play_count', 'art_url', 'bandcamp_track_id', 'currency', 'last_played']. Path: /run/media/sarunas/music/Music/soma/uun/03_givethemrope.flac
```

JOJ0:
> We can (and should) simply extend those list constants explicitly with new fields. They aren't specific to `musicbrainz` only - for example, `beetcamp` returns both `data_url` and `media`.
> 
> 

Explicitly! Yes please! Keep it readable. _If_ there would be something that is absolutely plugin specific, we could simply note it in a comment?

JOJ0:
> I think I see this logic in play in my `importlog`:

exactly! this is it!

snejus:
For now, I think, we can just extend those lists and merge this PR. Longer term, we need to find why flexible attribute update logic differs from db attribute update.

DB attributes are always updated with values other than `None`. `None` indicates _absence_ of a field, and thus this field is ignored in updates. This can be changed by `overwrite_null.album` and `overwrite_null.track` configuration: when a field is present in this list, `None` means _empty_ and the previous value is cleared on updates.

arsaboo:
Here's another discussion on this topic https://github.com/beetbox/beets/issues/4788

So, how do we want to resolve this PR? I am happy if one of you would like to take this forward.

beetsplug/musicbrainz.py
arsaboo: Haha...I am still scratching my head about this convoluted logic. Reimport should update everything.

beetsplug/musicbrainz.py
arsaboo: updated

arsaboo:
@snejus, I have updated the list for now. We can merge this for now and have a separate discussion about the changes required to fix the underlying issue.  It may also not be a bad idea for you to fix this in #6052
@semohr I can create a separate PR for #6076 once this one is done (so I know what is needed)

beetsplug/musicbrainz.py
snejus: What I meant by this comment was: these fields are in the DB for each relation, so they can't be flexible attributes, I think?

arsaboo:
@snejus check now. 

This may be beyond the scope of the current PR, but I assume that title, album, albumartist, etc. are still not updated on reimport. I don't see any reason why that should be the case. Like you said 

> Longer term, we need to find why flexible attribute update logic differs from db attribute update.

snejus:
@arsaboo it's impossible to have a flexible attribute named after any of the database fields. If you try to set `item.title = "new_title"`, you will update _the database field item.title_. See the following logic under `class Model`:

```py
    def _setitem(self, key, value):
        """Assign the value for a field, return whether new and old value
        differ.
        """
        # Choose where to place the value.
        if key in self._fields:
            source = self._values_fixed
        else:
            source = self._values_flex
```

`_values_fixed` property holds database fields like `album`, `title`, `albumartist`, `artist` etc., while `_values_flex` contains everything else.

beets/importer/tasks.py
```diff
@@ -58,8 +58,13 @@
     "deezer_album_id",
     "beatport_album_id",
     "tidal_album_id",
+    "media",
+    "releasegroup_id",
+    "data_url",
+]
+REIMPORT_FRESH_FIELDS_ITEM = [
```
snejus: Consider defining them in the opposite order:

```py
REIMPORT_FRESH_FIELDS_ITEM = [
    ...
]
REIMPORT_FRESH_FIELDS_ALBUM = [*REIMPORT_FRESH_FIELDS_ITEM, "media"]
```

beets/importer/tasks.py
```diff
@@ -58,8 +58,13 @@
     "deezer_album_id",
     "beatport_album_id",
     "tidal_album_id",
+    "media",
+    "releasegroup_id",
```
snejus: What prompted you to add this field, actually?

I don't have a single item or an album with such flexible attribute:

```fish
$ sqlite3 ~/.music/beets/library.db "select * from album_attributes where key == 'releasegroup_id'"
$ sqlite3 ~/.music/beets/library.db "select * from item_attributes where key == 'releasegroup_id'"
```

And I can see it is saved as `item.mb_releasegroupid` field in the database:

```py
beets/autotag/__init__.py-313-        item.mb_albumartistid = album_info.artist_id
beets/autotag/__init__.py-314-        item.mb_albumartistids = album_info.artists_ids
beets/autotag/__init__.py:315:        item.mb_releasegroupid = album_info.releasegroup_id
beets/autotag/__init__.py-316-
beets/autotag/__init__.py-317-        # Compilation flag.
```

arsaboo: I thought it was internally referred to as `releasegroup_id`. In the MusicBrainz plugin and in autotag/hooks.py, the field is called releasegroup_id; in the library models, it's stored as `mb_releasegroupid`; in the reimport fresh fields, it's using releasegroup_id (the internal field name used during the matching process)

So the field name `releasegroup_id` in REIMPORT_FRESH_FIELDS_ITEM maps to `mb_releasegroupid` in the database during the import process (as shown in line 315 of autotag/__init__.py: `item.mb_releasegroupid = album_info.releasegroup_id`).

BTW, these are some of the issues I hope you can resolve in your other PR #6052

semohr:
> Longer term, we need to find why flexible attribute update logic differs from db attribute update.

Before merging, could you please create an issue summarizing the situation, what’s different, and how or why we might want to change it? That way we can keep track of it and organize our thoughts for future work. 😉

arsaboo:
> > Longer term, we need to find why flexible attribute update logic differs from db attribute update.
> 
> Before merging, could you please create an issue summarizing the situation, what’s different, and how or why we might want to change it? That way we can keep track of it and organize our thoughts for future work. 😉

HAHA....I wish I had that clarity 🤣

I honestly did not even completely follow @snejus [comment](https://github.com/beetbox/beets/pull/6064#issuecomment-3378967115). @JOJ0 and @snejus are the closest to this, and I will let them summarize and resolve this in a subsequent PR.

semohr:
I kinda lost the sauce too. That's why im asking :D

snejus:
I often fail to communicate properly... I hope that this helps, guys, see: https://www.perplexity.ai/search/can-you-explain-what-is-snejus-eO3.NbE8TM.d.1FOARnPqA

arsaboo:
> I often fail to communicate properly... I hope that this helps, guys, see: https://www.perplexity.ai/search/can-you-explain-what-is-snejus-eO3.NbE8TM.d.1FOARnPqA

It was not your explanation, but my lack of understanding. This helps. To summarize (and for posterity), 

**Database Fields vs. Flexible Attributes:** Beets treats two types of fields differently:

Database Fields (_values_fixed): 
- These are predefined schema fields like title, album, albumartist, artist, etc.
- Stored directly in the SQLite database tables as columns
- Always updated during reimport when new values are found (unless the value is None)
- Updates are controlled by the overwrite_null configuration for handling None values

Flexible Attributes (_values_flex)
- Custom fields that users can add (like mood, play_count, etc.)
- Stored in separate key-value tables (album_attributes and item_attributes)
- Follow different update logic during reimport operations
- Subject to the REIMPORT_FRESH_FIELDS lists that determine which flexible attributes get refreshed

The REIMPORT_FRESH_FIELDS lists act as an allowlist - only flexible attributes in these lists get updated during reimport, while all others are preserved.

Let me know if anything else is required in this PR.

snejus:
There's no need to add `releasegroup_id` - see the last section in https://www.perplexity.ai/search/can-you-explain-what-is-snejus-eO3.NbE8TM.d.1FOARnPqA#2.
image

arsaboo:
> Thanks for your patience @arsaboo. We ended up having a very involved discussion relative to the size of this tiny change.
> 
> This only proves that this is one of the most complicated and opaque parts of the codebase and that we need to prioritize improving this to make it clearer to everyone.

Thanks for patiently explaining the issue. I feel I have a (slightly) better understanding :)

PR #6065, opened 2025-10-02
Empty metadata support for autotagger plugins

### LINKED ISSUE #6060: beet import -s fails with "at least one query term is required"
I'm trying to import a bunch of songs to my library, using `beet import -s .`

### Problem

```sh
beet -vv import -s . 
```

Led to this problem:

```
user configuration: /home/soso/.config/beets/config.yaml
data directory: /home/soso/.config/beets
plugin paths: []
Loading plugins: musicbrainz
Sending event: pluginload
library database: /home/soso/.config/beets/library.db
library directory: /home/soso/Music
Sending event: library_opened
Sending event: import_begin
Looking up: /home/soso/Downloads/music/Agar Agar ~ Prettiest Virgin [Yp257b5APOg].opus
Item search terms:  -
musicbrainz: Searching for MusicBrainz recordings with: {}
  File "/usr/bin/beet", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/usr/lib/python3.13/site-packages/beets/ui/__init__.py", line 1713, in main
    _raw_main(args)
    ~~~~~~~~~^^^^^^
  File "/usr/lib/python3.13/site-packages/beets/ui/__init__.py", line 1692, in _raw_main
  File "/usr/lib/python3.13/site-packages/beets/ui/commands.py", line 1378, in import_func
    import_files(lib, byte_paths, query)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/beets/ui/commands.py", line 1322, in import_files
    session.run()
    ~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/beets/importer/session.py", line 234, in run
    pl.run_parallel(QUEUE_SIZE)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/beets/util/pipeline.py", line 468, in run_parallel
  File "/usr/lib/python3.13/site-packages/beets/util/pipeline.py", line 333, in run
    out = self.coro.send(msg)
  File "/usr/lib/python3.13/site-packages/beets/util/pipeline.py", line 216, in coro
    func(*(args + (task,)))
    ~~~~^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/beets/importer/stages.py", line 144, in lookup_candidates
    task.lookup_candidates(session.config["search_ids"].as_str_seq())
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/beets/importer/tasks.py", line 686, in lookup_candidates
    self.candidates, self.rec = autotag.tag_item(
                                ~~~~~~~~~~~~~~~~^
        self.item, search_ids=search_ids
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/lib/python3.13/site-packages/beets/autotag/match.py", line 373, in tag_item
    for track_info in metadata_plugins.item_candidates(
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        item, search_artist, search_title
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ):
    ^
  File "/usr/lib/python3.13/site-packages/beets/plugins.py", line 473, in wrapper
    for v in func(*args, **kwargs):
             ~~~~^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/beets/metadata_plugins.py", line 69, in item_candidates
    yield from plugin.item_candidates(*args, **kwargs)
  File "/usr/lib/python3.13/site-packages/beetsplug/musicbrainz.py", line 837, in item_candidates
    None, map(self.track_info, self._search_api("recording", criteria))
                               ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/beetsplug/musicbrainz.py", line 812, in _search_api
    res = method(limit=self.config["search_limit"].get(), **filters)
  File "/usr/lib/python3.13/site-packages/musicbrainzngs/musicbrainz.py", line 971, in search_recordings
    return _do_mb_search('recording', query, fields, limit, offset, strict)
  File "/usr/lib/python3.13/site-packages/musicbrainzngs/musicbrainz.py", line 773, in _do_mb_search
    raise ValueError('at least one query term is required')
ValueError: at least one query term is required
```

There needs to be at least on music file in the directory where this is called. An empty opus file generated by `ffmpeg` is enough: `ffmpeg -f lavfi -i anullsrc -ac 2 -ar 48000 -t 30 -c:a libopus file.opus`

### Setup

* OS:  Arch Linux
* Python version: 3.13.7 
* beets version: beets version 2.4.0
* Turning off plugins made problem go away (yes/no): yes (the only plugin I use is musicbrainz)

Config: 
```yaml
disabled_plugins: []
musicbrainz:
    search_limit: 5
    source_weight: 0.5
    host: musicbrainz.org
    https: no
    ratelimit: 1
    ratelimit_interval: 1
    genres: no
    external_ids:
        discogs: no
        bandcamp: no
        spotify: no
        deezer: no
        tidal: no
    extra_tags: []
```

semohr:
## Description

It is possible for an metadata lookup to be performed with an empty string for both `artist` and `title/album`. This PR add handling for this edgecase for the metadata lookup of `musibrainz`, `spotify`, `discogs` and `beatport`. 

Seems like the issue was not catched earlier, since the typehints were
not propagated correctly in the `metadata_plugin.item_candidates` function.

closes #6060
#5965 might  have helped here too

### DIFF
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -236,19 +236,6 @@ def _add_candidate(
     )
 
 
-def _parse_search_terms_with_fallbacks(
-    *pairs: tuple[str | None, str],
-) -> tuple[str, ...]:
-    """Given pairs of (search term, fallback), return a tuple of
-    search terms. If **all** search terms are non-empty, return them. Otherwise,
-    return the fallback terms.
-    """
-    if all(term for term, _ in pairs):
-        return tuple(term or default for term, default in pairs)
-    else:
-        return tuple(default for _, default in pairs)
-
-
 def tag_album(
     items,
     search_artist: str | None = None,
@@ -307,24 +294,23 @@ def tag_album(
                         Proposal(list(candidates.values()), rec),
                     )
 
-        # Manually provided search terms or fallbacks.
-        _search_artist, _search_album = _parse_search_terms_with_fallbacks(
-            (search_artist, cur_artist),
-            (search_album, cur_album),
-        )
-        log.debug("Search terms: {} - {}", _search_artist, _search_album)
+        # Search terms.
+        if not (search_artist and search_album):
+            # No explicit search terms -- use current metadata.
+            search_artist, search_album = cur_artist, cur_album
+        log.debug("Search terms: {} - {}", search_artist, search_album)
 
         # Is this album likely to be a "various artist" release?
         va_likely = (
             (not consensus["artist"])
-            or (_search_artist.lower() in VA_ARTISTS)
+            or (search_artist.lower() in VA_ARTISTS)
             or any(item.comp for item in items)
         )
         log.debug("Album might be VA: {}", va_likely)
 
         # Get the results from the data sources.
         for matched_candidate in metadata_plugins.candidates(
-            items, _search_artist, _search_album, va_likely
+            items, search_artist, search_album, va_likely
         ):
             _add_candidate(items, candidates, matched_candidate)
 
@@ -336,7 +322,7 @@ def tag_album(
 
 
 def tag_item(
-    item: Item,
+    item,
     search_artist: str | None = None,
     search_title: str | None = None,
     search_ids: list[str] | None = None,
@@ -378,18 +364,14 @@ def tag_item(
         else:
             return Proposal([], Recommendation.none)
 
-    # Manually provided search terms or fallbacks.
-    _search_artist, _search_title = _parse_search_terms_with_fallbacks(
-        (search_artist, item.artist),
-        (search_title, item.title),
-    )
-    log.debug("Item search terms: {} - {}", _search_artist, _search_title)
+    # Search terms.
+    search_artist = search_artist or item.artist
+    search_title = search_title or item.title
+    log.debug("Item search terms: {} - {}", search_artist, search_title)
 
     # Get and evaluate candidate metadata.
     for track_info in metadata_plugins.item_candidates(
-        item,
-        _search_artist,
-        _search_title,
+        item, search_artist, search_title
     ):
         dist = track_distance(item, track_info, incl_artist=True)
         candidates[track_info.track_id] = hooks.TrackMatch(dist, track_info)
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -236,6 +236,19 @@ def _add_candidate(
     )
 
 
+def _parse_search_terms_with_fallbacks(
+    *pairs: tuple[str | None, str],
+) -> tuple[str, ...]:
+    """Given pairs of (search term, fallback), return a tuple of
+    search terms. If **all** search terms are non-empty, return them. Otherwise,
+    return the fallback terms.
+    """
+    if all(term for term, _ in pairs):
+        return tuple(term or default for term, default in pairs)
+    else:
+        return tuple(default for _, default in pairs)
+
+
 def tag_album(
     items,
     search_artist: str | None = None,
@@ -294,23 +307,24 @@ def tag_album(
                         Proposal(list(candidates.values()), rec),
                     )
 
-        # Search terms.
-        if not (search_artist and search_album):
-            # No explicit search terms -- use current metadata.
-            search_artist, search_album = cur_artist, cur_album
-        log.debug("Search terms: {} - {}", search_artist, search_album)
+        # Manually provided search terms or fallbacks.
+        _search_artist, _search_album = _parse_search_terms_with_fallbacks(
+            (search_artist, cur_artist),
+            (search_album, cur_album),
+        )
+        log.debug("Search terms: {} - {}", _search_artist, _search_album)
 
         # Is this album likely to be a "various artist" release?
         va_likely = (
             (not consensus["artist"])
-            or (search_artist.lower() in VA_ARTISTS)
+            or (_search_artist.lower() in VA_ARTISTS)
             or any(item.comp for item in items)
         )
         log.debug("Album might be VA: {}", va_likely)
 
         # Get the results from the data sources.
         for matched_candidate in metadata_plugins.candidates(
-            items, search_artist, search_album, va_likely
+            items, _search_artist, _search_album, va_likely
         ):
             _add_candidate(items, candidates, matched_candidate)
 
@@ -322,7 +336,7 @@ def tag_album(
 
 
 def tag_item(
-    item,
+    item: Item,
     search_artist: str | None = None,
     search_title: str | None = None,
     search_ids: list[str] | None = None,
@@ -364,14 +378,18 @@ def tag_item(
         else:
             return Proposal([], Recommendation.none)
 
-    # Search terms.
-    search_artist = search_artist or item.artist
-    search_title = search_title or item.title
-    log.debug("Item search terms: {} - {}", search_artist, search_title)
+    # Manually provided search terms or fallbacks.
+    _search_artist, _search_title = _parse_search_terms_with_fallbacks(
+        (search_artist, item.artist),
+        (search_title, item.title),
+    )
+    log.debug("Item search terms: {} - {}", _search_artist, _search_title)
 
     # Get and evaluate candidate metadata.
     for track_info in metadata_plugins.item_candidates(
-        item, search_artist, search_title
+        item,
+        _search_artist,
+        _search_title,
     ):
         dist = track_distance(item, track_info, incl_artist=True)
         candidates[track_info.track_id] = hooks.TrackMatch(dist, track_info)
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -35,17 +35,35 @@ def find_metadata_source_plugins() -> list[MetadataSourcePlugin]:
 
 
 @notify_info_yielded("albuminfo_received")
-def candidates(*args, **kwargs) -> Iterable[AlbumInfo]:
+def candidates(
+    items: Sequence[Item],
+    artist: str,
+    album: str,
+    va_likely: bool,
+) -> Iterable[AlbumInfo]:
     """Return matching album candidates from all metadata source plugins."""
     for plugin in find_metadata_source_plugins():
-        yield from plugin.candidates(*args, **kwargs)
+        yield from plugin.candidates(
+            items=items,
+            artist=artist,
+            album=album,
+            va_likely=va_likely,
+        )
 
 
 @notify_info_yielded("trackinfo_received")
-def item_candidates(*args, **kwargs) -> Iterable[TrackInfo]:
-    """Return matching track candidates fromm all metadata source plugins."""
+def item_candidates(
+    item: Item,
+    artist: str,
+    title: str,
+) -> Iterable[TrackInfo]:
+    """Return matching track candidates from all metadata source plugins."""
     for plugin in find_metadata_source_plugins():
-        yield from plugin.item_candidates(*args, **kwargs)
+        yield from plugin.item_candidates(
+            item=item,
+            artist=artist,
+            title=title,
+        )
 
 
 def album_for_id(_id: str) -> AlbumInfo | None:
@@ -157,15 +175,22 @@ class MetadataSourcePlugin(BeetsPlugin, metaclass=abc.ABCMeta):
 
     @abc.abstractmethod
     def item_candidates(
-        self, item: Item, artist: str, title: str
+        self,
+        item: Item,
+        artist: str,
+        title: str,
     ) -> Iterable[TrackInfo]:
         """Return :py:class:`TrackInfo` candidates that match the given track.
 
         Used in the autotag functionality to search for tracks.
 
         :param item: Track item
-        :param artist: Track artist
-        :param title: Track title
+        :param artist: Track artist, either a search manually provided or
+            preprocessed from the item. If no metadata is available an empty string
+            is passed.
+        :param title: Track title, either a search manually provided or
+            preprocessed from the item. If no metadata is available an empty string
+            is passed.
         """
         raise NotImplementedError
 
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -35,35 +35,17 @@ def find_metadata_source_plugins() -> list[MetadataSourcePlugin]:
 
 
 @notify_info_yielded("albuminfo_received")
-def candidates(
-    items: Sequence[Item],
-    artist: str,
-    album: str,
-    va_likely: bool,
-) -> Iterable[AlbumInfo]:
+def candidates(*args, **kwargs) -> Iterable[AlbumInfo]:
     """Return matching album candidates from all metadata source plugins."""
     for plugin in find_metadata_source_plugins():
-        yield from plugin.candidates(
-            items=items,
-            artist=artist,
-            album=album,
-            va_likely=va_likely,
-        )
+        yield from plugin.candidates(*args, **kwargs)
 
 
 @notify_info_yielded("trackinfo_received")
-def item_candidates(
-    item: Item,
-    artist: str,
-    title: str,
-) -> Iterable[TrackInfo]:
-    """Return matching track candidates from all metadata source plugins."""
+def item_candidates(*args, **kwargs) -> Iterable[TrackInfo]:
+    """Return matching track candidates fromm all metadata source plugins."""
     for plugin in find_metadata_source_plugins():
-        yield from plugin.item_candidates(
-            item=item,
-            artist=artist,
-            title=title,
-        )
+        yield from plugin.item_candidates(*args, **kwargs)
 
 
 def album_for_id(_id: str) -> AlbumInfo | None:
@@ -175,22 +157,15 @@ class MetadataSourcePlugin(BeetsPlugin, metaclass=abc.ABCMeta):
 
     @abc.abstractmethod
     def item_candidates(
-        self,
-        item: Item,
-        artist: str,
-        title: str,
+        self, item: Item, artist: str, title: str
     ) -> Iterable[TrackInfo]:
         """Return :py:class:`TrackInfo` candidates that match the given track.
 
         Used in the autotag functionality to search for tracks.
 
         :param item: Track item
-        :param artist: Track artist, either a search manually provided or
-            preprocessed from the item. If no metadata is available an empty string
-            is passed.
-        :param title: Track title, either a search manually provided or
-            preprocessed from the item. If no metadata is available an empty string
-            is passed.
+        :param artist: Track artist
+        :param title: Track title
         """
         raise NotImplementedError
 
--- a/beetsplug/beatport.py
+++ b/beetsplug/beatport.py
@@ -457,6 +457,16 @@ class BeatportPlugin(MetadataSourcePlugin):
         # Strip medium information from query, Things like "CD1" and "disk 1"
         # can also negate an otherwise positive result.
         query = re.sub(r"\b(CD|disc)\s*\d+", "", query, flags=re.I)
+
+        # query may be empty strings
+        # We want to skip the lookup in this case.
+        if not query.strip():
+            self._log.debug(
+                "Empty search query after preprocessing, skipping {.data_source}.",
+                self,
+            )
+            return
+
         for beatport_release in self.client.search(query, "release"):
             if beatport_release is None:
                 continue
@@ -522,8 +532,18 @@ class BeatportPlugin(MetadataSourcePlugin):
         """
         return self.get_artist(artists=artists, id_key=0, name_key=1)
 
-    def _get_tracks(self, query):
+    def _get_tracks(self, query: str):
         """Returns a list of TrackInfo objects for a Beatport query."""
+
+        # query may be empty strings
+        # We want to skip the lookup in this case.
+        if not query.strip():
+            self._log.debug(
+                "Empty search query after preprocessing, skipping {.data_source}.",
+                self,
+            )
+            return []
+
         bp_tracks = self.client.search(query, release_type="track")
         tracks = [self._get_track_info(x) for x in bp_tracks]
         return tracks
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -231,7 +231,10 @@ class DiscogsPlugin(MetadataSourcePlugin):
         return track_info
 
     def item_candidates(
-        self, item: Item, artist: str, title: str
+        self,
+        item: Item,
+        artist: str,
+        title: str,
     ) -> Iterable[TrackInfo]:
         albums = self.candidates([item], artist, title, False)
 
@@ -290,6 +293,15 @@ class DiscogsPlugin(MetadataSourcePlugin):
         # Strip medium information from query, Things like "CD1" and "disk 1"
         # can also negate an otherwise positive result.
         query = re.sub(r"(?i)\b(CD|disc|vinyl)\s*\d+", "", query)
+
+        # query may be empty strings
+        # We want to skip the lookup in this case.
+        if not query.strip():
+            self._log.debug(
+                "Empty search query after preprocessing, skipping {.data_source}.",
+                self,
+            )
+            return []
 
         try:
             results = self.discogs_client.search(query, type="release")
--- a/beetsplug/musicbrainz.py
+++ b/beetsplug/musicbrainz.py
@@ -807,6 +807,11 @@ class MusicBrainzPlugin(MetadataSourcePlugin):
         self._log.debug(
             "Searching for MusicBrainz {}s with: {!r}", query_type, filters
         )
+
+        if not filters:
+            self._log.debug("No valid filters provided, skipping search.")
+            return []
+
         try:
             method = getattr(musicbrainzngs, f"search_{query_type}s")
             res = method(limit=self.config["search_limit"].get(), **filters)
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -434,6 +434,13 @@ class SpotifyPlugin(
             filters=filters, query_string=query_string
         )
 
+        if not query.strip():
+            self._log.debug(
+                "Empty search query after applying filters, skipping {.data_source}.",
+                self,
+            )
+            return []
+
         self._log.debug("Searching {.data_source} for '{}'", self, query)
         try:
             response = self._handle_response(
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -24,9 +24,10 @@ For packagers:
 Other changes:
 
 - Standardized ``search_*`` parameter handling in autotag matchers. Manual album
-  and singleton searches now behave consistently: when the prompt does not
-  specify a search query, the system defaults to using the corresponding
-  metadata value.
+  and singleton searches now behave consistently: when a user does not specify a
+  search query in the prompt, the system defaults to using the corresponding
+  value from the metadata. This was already the case for albums but not for
+  singletons.
 
 2.5.1 (October 14, 2025)
 ------------------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -23,12 +23,6 @@ For packagers:
 
 Other changes:
 
-- Standardized ``search_*`` parameter handling in autotag matchers. Manual album
-  and singleton searches now behave consistently: when a user does not specify a
-  search query in the prompt, the system defaults to using the corresponding
-  value from the metadata. This was already the case for albums but not for
-  singletons.
-
 2.5.1 (October 14, 2025)
 ------------------------
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -15,13 +15,18 @@ New features:
 
 Bug fixes:
 
+- :doc:`plugins/discogs`, :doc:`plugins/beatport`, :doc:`plugins/spotify`,
+  :doc:`plugins/musicbrainz`: Fix an issue where no metadata in a file would
+  crash the import process :bug:`6060`
+
 For packagers:
 
 Other changes:
 
-- Standardized ``search_*`` parameter handling in autotag matchers. Manual album and
-  singleton searches now behave consistently: when the prompt does not specify a
-  search query, the system defaults to using the corresponding metadata value.
+- Standardized ``search_*`` parameter handling in autotag matchers. Manual album
+  and singleton searches now behave consistently: when the prompt does not
+  specify a search query, the system defaults to using the corresponding
+  metadata value.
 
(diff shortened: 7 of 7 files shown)


beetsplug/musicbrainz.py
```diff
@@ -801,6 +801,9 @@ def _search_api(
         using the provided criteria. Handles API errors by converting them into
         MusicBrainzAPIError exceptions with contextual information.
         """
+        if not filters:
+            return []
+
         filters = {
             k: _v for k, v in filters.items() if (_v := v.lower().strip())
         }
```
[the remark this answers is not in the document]
semohr: filters is always a `dict[str,str]` here!

test/plugins/test_musicbrainz.py
```diff
@@ -1030,16 +1030,33 @@ def test_get_album_criteria(
...
         )
 
-        candidates = list(mb.item_candidates(Item(), "hello", "there"))
+        candidates = list(mb.item_candidates(Item(), artist, title))
 
-        assert len(candidates) == 1
-        assert candidates[0].track_id == self.RECORDING["id"]
+        assert len(candidates) == expected_count
+        if expected_count == 1:
+            assert candidates[0].track_id == self.RECORDING["id"]
```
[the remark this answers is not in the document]
semohr: This would than require another duplication of this test. I think here this is the clean approach.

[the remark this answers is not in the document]
semohr: I told you before https://github.com/beetbox/beets/pull/6065#discussion_r2398979289.

beets/autotag/match.py
```diff
@@ -366,9 +366,15 @@ def tag_item(
 
     # Search terms.
     search_artist = search_artist or item.artist
-    search_title = search_title or item.title
+    search_title = search_title or item.title or item.filepath.stem
```
snejus: We shouldn't have this here - if users want to use file names, they can enable the `fromfilename` plugin.

snejus:
> Seems like the issue was not catched earlier, since the typehints were
not propagated correctly in the metadata_plugin.item_candidates function.

I may be missing something, but it seems to me they have been correct? Both candidates functions always received strings - including empty ones.

beets/metadata_plugins.py
```diff
@@ -383,12 +394,21 @@ def candidates(
...
+        artist: str | None,
+        title: str | None,
     ) -> Iterable[TrackInfo]:
-        results = self._search_api(
-            "track", {"artist": artist}, query_string=title
-        )
-        if not results:
+        filters: SearchFilter = {}
+        if artist is not None:
+            filters["artist"] = artist
```
snejus: I don't see how the introduction of `None` brings more clarity - this now says '_any_ string is fine, including an empty one'. 

On the other hand, if one is aware of `tag_item` implementation, they know that `None` just stands for an empty string.

Thus, I think, we should continue using strings and simply handle `if not artist` etc. if needed.

beets/metadata_plugins.py
```diff
@@ -383,12 +394,21 @@ def candidates(
...
+
+        if title is None:
```
snejus: This is a big change in behaviour - every metadata source handles absence of `title` fine when `artist` is given.

snejus: In fact, both linked issues indicate that the issue is limited to `musicbrainz` *only* and only when both `artist` and `title` are empty.

beetsplug/beatport.py
```diff
@@ -409,9 +409,17 @@ def candidates(
             return
 
     def item_candidates(
-        self, item: Item, artist: str, title: str
+        self, item: Item, artist: str | None, title: str | None
     ) -> Iterable[TrackInfo]:
-        query = f"{artist} {title}"
+        query = ""
```
snejus: We can keep the same `str` types and use the following instead

```py
query = " ".join(map(str.strip, [artist, title]))
```

beetsplug/musicbrainz.py
```diff
@@ -829,9 +832,17 @@ def candidates(
...
     def item_candidates(
-        self, item: Item, artist: str, title: str
+        self,
+        item: Item,
+        artist: str | None,
+        title: str | None,
     ) -> Iterable[beets.autotag.hooks.TrackInfo]:
-        criteria = {"artist": artist, "recording": title, "alias": title}
+        criteria: dict[str, str] = {}
+        if artist is not None:
```
snejus: This logic is redundant - `_search_api` filters out empty values _consistently_ for both singletons and albums. No change is needed here

beetsplug/musicbrainz.py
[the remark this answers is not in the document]
snejus: Copilot is confused now that `None`s have entered the game - let's keep them out.

snejus, line 806: Once you've reverted to handling empty strings, the only adjustment that this module needs is adding an early return here:

```py
if not filters:
    return []
```

test/plugins/test_musicbrainz.py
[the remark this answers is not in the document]
snejus: This is a good point by Copilot - you can remove the conditional logic by using `expected_ids` parameter instead of `expected_count`.

snejus (changes requested):
What is the practical benefit of replacing empty strings by `None`? I outlined a couple of issues with this approach.

I'm a bit confused about the 'why' here, especially so because this only applies to tracks/singletons and not albums

beets/autotag/match.py
semohr: I think it doesn't really hurt here and makes this one edgecase a bit better. Some search results are better than none imo. 
But don't get me wrong I agree that people are better of using the `fromfilename` plugin.

beets/metadata_plugins.py
semohr: You are right, using empty strings is the easier way. Will change it in the next commit.

To give you some context how we got here: I just fixed the mypy issues that followed from typing the args and kwargs in the `item_candidates` function.  

The main issue with the types stems from shadowing not working as expected for [`search_artists/titles`](https://github.com/beetbox/beets/pull/6065/files#diff-ced3cc4e79a836f174228f2b7cc20a54a7c5e01192fd6c3edfa6d5c3222f84feR326). Every other function down the line expected the search strings to be `str | None`.  The fix is straight forward, just didn't see this initially and thought `item.title` can return a `None` too :facepalm:

beets/metadata_plugins.py
semohr: I will probably change it in the next commit, but I don't think it is a big change. At least from the typehint perspective this is how it is supposed to be imo. We use `NotRequired` for both the `title` and `artist` filter.

semohr: > In fact, both linked issues indicate that the issue is limited to musicbrainz only and only when both artist and title are empty.

Maybe this is also a good reason for creating a proper general tests suit for all `metadata` plugins. I will see if I can think about how to approach this properly.

beetsplug/musicbrainz.py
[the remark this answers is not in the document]
semohr: Agreed. I think taking them out and just putting a remark in the docstrings should be good enough. A test also should help for the long term 👍

beetsplug/musicbrainz.py
semohr: Sadly not. I tested it: If you for example input `{"artist":None, "title":"something"}` the `_search_api` will crash. But the logic should not matter if we properly support empty strings.

test/plugins/test_musicbrainz.py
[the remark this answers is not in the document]
semohr: I think this might still need an if for the candidate indexing or a filter which is basically the same logic. I will see if it works tho.

semohr:
> > Seems like the issue was not catched earlier, since the typehints were
> > not propagated correctly in the metadata_plugin.item_candidates function.
> 
> I may be missing something, but it seems to me they have been correct? Both candidates functions always received strings - including empty ones.


see https://github.com/beetbox/beets/pull/6065#discussion_r2401359869

beets/autotag/match.py
snejus: It changes the behaviour for external plugins that expect this string to be either the search string or the `title` of the item.

beetsplug/musicbrainz.py
snejus: Apologies - forgot to clarify that my comment applies only if we handle strings _only_ 😅

beets/autotag/match.py
semohr: Now I'm wondering why we even do this simple parsing here in the first place.

Item is included in the function call anyways. Having the manual search strings for artist and title makes sense, but the `search_title or item.title` seems a bit like duplication that is not clear for the consumers.

What do you think?

test/plugins/test_musicbrainz.py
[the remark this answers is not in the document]
snejus: I meant something like this

```py
    @pytest.mark.parametrize(
        "artist, title, expected_ids",
        [
            ("Artist", "Title", [RECORDING["id"]]),
            (None, "Title", [RECORDING["id"]]),
            ("Artist", None, [RECORDING["id"]]),
            (None, None, []),
        ],
    )
    def test_item_candidates(self, monkeypatch, mb, artist, title, expected_ids):
        monkeypatch.setattr(
            "musicbrainzngs.search_recordings",
            lambda *_, **__: {"recording-list": [self.RECORDING]},
        )
        ids = [c.track_id for c in mb.item_candidates(Item(), artist, title)]
        assert ids == expected_ids
```

beets/autotag/match.py
snejus: I have just realised that I inadvertently changed this logic in f1dc75f7431987e6f17eb8cbf0d610c3b2180183:

```diff
     # Search terms.
-    if not (search_artist and search_title):
-        search_artist, search_title = item.artist, item.title
+    search_artist = search_artist or item.artist
+    search_title = search_title or item.title
     log.debug("Item search terms: {0} - {1}", search_artist, search_title)
```

Just like album search above, this should default to `item` data __only__ when no search terms are given at all.

semohr: This would complicate it again tho, right? Essentially this would allow for `None` again :laughing: Seems like we are going in circles. Why dont we just use the raw `search_title` and `search_artist` and let plugins decide on their own how to handle this? The `item` is available in plugins too.

test/plugins/test_musicbrainz.py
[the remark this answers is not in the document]
semohr: Does this work? I think you can't access `self` in parametrize. I will have a look once I touch this again, Im sure there is a workaround :shrug:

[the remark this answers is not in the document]
snejus: > Does this work? I think you can't access `self` in parametrize. I will have a look once I touch this again, Im sure there is a workaround 🤷

You're right - you cannot access `self` within parametrize, but are you talking about my suggestion? It doesn't do so.

`parametrize` decorator (and the test function) is defined **on the class**, thus it can make use of class properties/attributes defined before it as they are just variables within the same scope.

```py
class TestMusicBrainzPlugin(PluginMixin):
    RECORDING = {"title": "foo", "id": "bar", "length": 42}
    ...
    @pytest.mark.parametrize(
        "artist, title, expected_ids",
        [
            ("Artist", "Title", [RECORDING["id"]]),
            (None, "Title", [RECORDING["id"]]),
            ("Artist", None, [RECORDING["id"]]),
            (None, None, []),
        ],
    )
    def test_item_candidates(self, monkeypatch, mb, artist, title, expected_ids):
        ...
```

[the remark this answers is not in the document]
snejus: See an example in `test_query.py`:

```py
class TestQuery:
    ALBUM = "album title"
    SINGLE = "singleton"

    @pytest.fixture(scope="class")
    def lib(self, helper):
        helper.add_album(
            title=self.ALBUM,
            comp=True,
            flexbool=True,
            bpm=120,
            flexint=2,
            rg_track_gain=0,
        )
        helper.add_item(
            title=self.SINGLE, comp=False, flexbool=False, rg_track_gain=None
        )

        with pytest.MonkeyPatch.context() as monkeypatch:
            monkeypatch.setattr(
                Item,
                "_types",
                {"flexbool": types.Boolean(), "flexint": types.Integer()},
            )
            yield helper.lib

    @pytest.mark.parametrize("query_class", [MatchQuery, StringFieldQuery])
    def test_equality(self, query_class):
        assert query_class("foo", "bar") == query_class("foo", "bar")

    @pytest.mark.parametrize(
        "make_q, expected_msg",
        [
            (lambda: NumericQuery("year", "199a"), "not an int"),
            (lambda: RegexpQuery("year", "199("), r"not a regular expression.*unterminated subpattern"),  # noqa: E501
        ]
    )  # fmt: skip
    def test_invalid_query(self, make_q, expected_msg):
        with pytest.raises(ParsingError, match=expected_msg):
            make_q()

    @pytest.mark.parametrize(
        "q, expected_titles",
        [
            # Boolean value
            _p("comp:true", {ALBUM}, id="parse-true"),
            _p("flexbool:true", {ALBUM}, id="flex-parse-true"),
            _p("flexbool:false", {SINGLE}, id="flex-parse-false"),
            _p("flexbool:1", {ALBUM}, id="flex-parse-1"),
            _p("flexbool:0", {SINGLE}, id="flex-parse-0"),
            # TODO: shouldn't this match 1 / true instead?
            _p("flexbool:something", {SINGLE}, id="flex-parse-true"),
            # Integer value
            _p("bpm:120", {ALBUM}, id="int-exact-value"),
            _p("bpm:110..125", {ALBUM}, id="int-range"),
            _p("flexint:2", {ALBUM}, id="int-flex"),
            _p("flexint:3", set(), id="int-no-match"),
            _p("bpm:12", set(), id="int-dont-match-substring"),
            # None value
            _p(NoneQuery("album_id"), {SINGLE}, id="none-match-singleton"),
            _p(NoneQuery("rg_track_gain"), {SINGLE}, id="none-value"),
        ],
    )
    def test_value_type(self, lib, q, expected_titles):
        assert {i.title for i in lib.items(q)} == expected_titles
```

beetsplug/beatport.py
```diff
@@ -457,6 +457,16 @@ def _get_releases(self, query: str) -> Iterator[AlbumInfo]:
...
         query = re.sub(r"\b(CD|disc)\s*\d+", "", query, flags=re.I)
+
+        # query may be empty strings
+        # We want to skip the lookup in this case.
+        if not query.strip():
+            self._log.debug(
+                "Empty search query after preprocessing, skipping {.data_source}.",
+                self,
+            )
+            return
```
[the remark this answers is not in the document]
semohr: Raises a type error if i return an empty array.

beets/autotag/match.py
```diff
@@ -236,6 +236,13 @@ def _add_candidate(
...
 
 
+def _parse_search_terms(
+    *pairs: tuple[str | None, str],
+) -> tuple[str, ...]:
+    """Determine the search terms to use, falling back to current values."""
+    return tuple(term or current for term, current in pairs)
+
+
 def tag_album(
```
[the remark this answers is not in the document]
semohr: You tripping.

beets/autotag/match.py
```diff
@@ -295,22 +302,23 @@ def tag_album(
                     )
 
         # Search terms.
-        if not (search_artist and search_album):
-            # No explicit search terms -- use current metadata.
-            search_artist, search_album = cur_artist, cur_album
-        log.debug("Search terms: {} - {}", search_artist, search_album)
+        _search_artist, _search_album = _parse_search_terms(
```
snejus: Do we need this indirection? I preferred the previous logic which explicitly said

```py
        if not (search_artist and search_album)
```

beets/autotag/match.py
```diff
@@ -365,13 +373,17 @@ def tag_item(
...
 
     # Search terms.
-    search_artist = search_artist or item.artist
-    search_title = search_title or item.title
-    log.debug("Item search terms: {} - {}", search_artist, search_title)
+    _search_artist, _search_title = _parse_search_terms(
+        (search_artist, item.artist),
+        (search_title, item.title),
+    )
+    log.debug("Item search terms: {} - {}", _search_artist, _search_title)
```
snejus: I'd rather revert this back to what we had before my change:

```py
        if not (search_artist and search_album):
            search_artist = item.artist
            search_title = item.title
```

beets/metadata_plugins.py
```diff
@@ -35,17 +35,35 @@ def find_metadata_source_plugins() -> list[MetadataSourcePlugin]:
 
 
 @notify_info_yielded("albuminfo_received")
-def candidates(*args, **kwargs) -> Iterable[AlbumInfo]:
+def candidates(
```
snejus: I'm not convinced it is a good idea to hardcode the arguments here. It passes all arguments through _unchanged_ and calls another function, yielding the results.

Simply put, it's out of scope for this function to know what specific arguments are being fed into it. Same applies to `item_candidates`.

beetsplug/discogs.py
```diff
@@ -291,6 +294,15 @@ def get_albums(self, query: str) -> Iterable[AlbumInfo]:
         # can also negate an otherwise positive result.
         query = re.sub(r"(?i)\b(CD|disc|vinyl)\s*\d+", "", query)
 
+        # query may be empty strings
+        # We want to skip the lookup in this case.
+        if not query.strip():
```
snejus, line 299: I don't like this logic being duplicated in every data source. Can we not check this in `plugins.item_candidates`?

beets/autotag/match.py
semohr: I introduced the extra function as I noticed the logic was different between the singleton and album implementation. Having the private function here couples the two cases, which I like. 
It is not enterily required but might be a good idea as we already had some drift here.

beets/metadata_plugins.py
semohr: I agree in theory but would value the added type safety more highly here.

beetsplug/discogs.py
semohr: I think we can't move this logic easily. Some metadataplugins do not require artist or title/album at all. We can't move the check one layer up as this would break plugins only depending on item.

beets/autotag/match.py
snejus: As I mentioned in the comment below, I introduced that inconsistency _accidentally_. This logic should be as follows:

```py
        if not (search_artist and search_album):
            search_artist = cur_artist
            search_title = cur_album
```

_Only_ when **both** are missing we should reset to `cur_artist` and `cur_album`. Hiding this logic in a method just makes this harder to see and understand.

beets/metadata_plugins.py
snejus: There's no need for type safety for a method that has no interest in parameters that are being fed into it. On `master` branch, autocompletion already correctly resolves the parameters it accepts:

Image

Let's keep this PR focused on the fix - I don't want to block it because of this.

semohr: Mypy strips types here which I would prefer to not do:
```
def candidates(*arg,**kwargs) -> Iterable[AlbumInfo]:
    reveal_type(arg)
    ...
beets/metadata_plugins.py:39: note: Revealed type is "builtins.tuple[Any, ...]"
```
This is definitely an issue and can introduce further problems if we are not careful. For instance passing wrong parameter to the `plugins.candidates` function would currently not be raised as mypy error. 

This is the issue why `metadata_plugins.*candidates` in the `tag_item` and `tag_album` function did accept the `str|None` type for album, title and artist in the first place.

beets/autotag/match.py
semohr: We can change it back I dont mind which implementation we use. Would be nice to just overwrite one value tho, e.g. use `search_album` while keep the artist from the metadata. I don't think this is possible right now, might need some adjustments to the command input itself.

beets/autotag/match.py
semohr: I would still like to keep the small helper function. The proper typing needs a bit of additional fluff which I would like to abstract or omit. I renamed the function and made it clear what happens in the docstring. For me this makes it easier to understand.

I prefer
```python

def _parse_search_terms_with_fallbacks(
    *pairs: tuple[str | None, str],
) -> tuple[str, ...]:
    """Given pairs of (search term, fallback), return a tuple of
    search terms. If **all** search terms are non-empty, return them. Otherwise,
    return the fallback terms.
    """
    if all(term for term, _ in pairs):
        return tuple(term or default for term, default in pairs)
    else:
        return tuple(default for _, default in pairs)

# ...

 _search_artist, _search_title = _parse_search_terms_with_fallbacks(
        (search_artist, item.artist),
        (search_title, item.title),
)
 ```
over
```python
    _search_artist: str
    _search_title: str
    if not (search_artist and search_title):
        _search_artist = item.artist
        _search_title = item.title
    else:
        _search_artist = search_artist
        _search_title = search_title
  ```

File: beets/metadata_plugins.py (hunk 12 above)
Comment (snejus) [reply beets/metadata_plugins.py]: As I mentioned, let's keep this PR focused on the fix. You can open a separate PR and we can specifically discuss `*args, **kwargs` there.

File: beets/autotag/match.py (hunk 10 above)
Comment (snejus) [reply beets/autotag/match.py]: It's three lines of code in two places:

```py
    if not search_artist and not search_title:
        search_artist = item.artist
        search_title = item.title
```

and this is definitely more clear than the helper function above.

semohr: Please have a full look here this wont work because we do not allow shadowing! We need an additional variable if we want proper typing.

beetsplug/discogs.py
snejus: Seems like this logic is present in all internal plugins. I also don't see a use case in `beetcamp` where **both** `artist` and `title` / `album` are missing - I think we can safely handle this one layer up and remove the duplication.

When does `fromfilename` plugin intercept these parameters and insert data from the filename?

semohr: Have a look at chroma.

beets/autotag/match.py
snejus: If that's the case, we need to adjust `MetadataSourcePlugin.candidates` and `item_candidates` to take into account the fact that either `artist` and/or `title`/`album` can be `None`:

```py
    @abc.abstractmethod
    def candidates(
        self,
        items: Sequence[Item],
        artist: str | None,
        album: str | None,
        va_likely: bool,
    ) -> Iterable[AlbumInfo]:
```

We should keep the previous functionality where we allow either of these two to be missing.

semohr: We are turning in circles here, lol. Using none was the initial implementation in this PR. We agreed that the added complexity for it is not worth it.

beetsplug/discogs.py
snejus: Completely forgot about this one 😞

beets/autotag/match.py
snejus: Didn't we agree that this function only ever receives strings? Regardless - we are keeping the functionality as it was - and thus the simple logic that we had.

semohr: Exactly, this is the reason why we are missing a type error: `*args, **kwargs` are not typed in the downstream functions. We can reverse this but as I said in another comment this is an issue! 
Seeing you are also confused by how this works, the types should be absolutely worth it ;)

semohr: Signature `match.py` (notice the default none for no values):
```
def tag_item(
    item: Item,
    search_artist: str | None = None,
    search_title: str | None = None,
    search_ids: list[str] | None = None,
) -> Proposal:
```
Signature `plugins.py` (no default value):
```
def item_candidates(
    item: Item,
    artist: str,
    title: str,
) -> Iterable[TrackInfo]:
```

During `tag_item` we need to drop the `None`. We decided to not allow shadowing, we can't just reuse `search_artist`, `search_title` here as they would be `str|None` instead of `str` -> type error.

semohr: Moved to #6117

beets/metadata_plugins.py
semohr: Moved to #6117

semohr:
@snejus How do we want to continue here? 

The main concern was the duplication as far as I can remember. It might be possible to move the duplication check into an api layer (or shared session) at some point but right now I dont see how we can dedupe this otherwise.

snejus:
Forgot about this PR! My main concern is that `chroma` plugin is very different from the rest of metadata source plugins - and whether we gain any value from _forcing_ it to implement `MetadataSourcePlugin`? It does not implement `track_for_id` and `album_for_id` which also indicates that this interface may not the best fit for it.

I'm leaning towards making it inherit `BeetsPlugin` instead:
1. Define `data_source` explicitly to make sure it's still treated as a data source
2. Centralise the empty search handling for `MetadataSourcePlugin` subclasses

semohr:
I think the proposed change creates more problems than it solves. Maybe we are stuck here without a bigger refactor.

Chroma does define `track_for_id` and `album_for_id` they just return empty results. And that actually makes sense: the plugin supports the interface contract, but simply doesn’t produce lookups because fingerprinting doesn’t map cleanly to ID-based metadata sources (which it could in theory). That’s a perfectly valid implementation and doesn’t justify removing it from `MetadataSourcePlugin`.

Switching it to inherit directly from `BeetsPlugin'` blurs the separation between the two interfaces. `MetadataSourcePlugin` has a clear purpose imo _ID-based lookup and search semantics_, while `BeetsPlugin` is intentionally generic. Moving `chroma` out of the metadata-source abstraction weakens that clarity and makes the inheritance structure harder to reason about.

It also goes against one of `beets` strengths: being open and flexible in how plugins can be implemented. We shouldn’t impose rules that don’t fit every plugin. Chroma isn’t the only case, there are external metadata plugins that also purely work on `Item` data without necessary requiring titles and artist (e.g. [aisauce](https://github.com/metasauce/beets-aisauce/blob/ca9591d3ca2b3845da6ef62ded96d1b25d5ab4ef/beetsplug/aisauce/aisauce.py#L272-L279) or [audible](https://github.com/Neurrone/beets-audible/blob/cfc9b1cfd250906a254dc827fe2e073e342cf06c/beetsplug/audible.py#L140)).

I agree that from a maintainer standpoint, the proposal may simplify internal maintenance slightly. But it adds ambiguity for plugin authors, increases the risk of mismatched expectations in future core changes, and could even break existing workflows for users.

---

TLDR: chroma fits the interface well enough, and `beets` has always thrived by not forcing plugin authors into rigid patterns. This change would move in the opposite direction.

beetsplug/discogs.py
```diff
@@ -231,7 +231,10 @@ def get_track_from_album(
         return track_info
 
     def item_candidates(
-        self, item: Item, artist: str, title: str
```
snejus, line 218: Nothing seems to have changed in this line. Re your own comment: https://github.com/beetbox/beets/pull/6184#discussion_r2554118022

beetsplug/beatport.py
```diff
@@ -522,8 +532,18 @@ def _get_artist(self, artists):
         """
         return self.get_artist(artists=artists, id_key=0, name_key=1)
 
-    def _get_tracks(self, query):
+    def _get_tracks(self, query: str):
         """Returns a list of TrackInfo objects for a Beatport query."""
+
+        # query may be empty strings
+        # We want to skip the lookup in this case.
+        if not query.strip():
```
snejus, line 540: Can we define a helper method in `MetadataSourcePlugin` to deduplicate this logic?

```py
    ...
    def is_query_empty(query: str) -> bool:
        if not query.strip():
            self._log.debug(
                "Empty search query after preprocessing, skipping {.data_source}.",
                self,
            )
            return True

        return False
```

and then have

```py
    if self.is_query_empty(query):
        return []
```

here.

snejus (changes requested):
Agree with your last comment - let's just dedupe this logic and we're good to go

beetsplug/beatport.py
semohr: In theory we could add `MetadataSourcePlugin.is_query_empty`. And I thought about this too. The issue I see here is that we again start to pollute our currently very clean `MetadataSourcePlugin` interface with functions that are not used in the core and only used by some subclasses. This kinda signals wrong abstraction boundaries to me, think interface segregation principle.

We can although add a helper function to the `beetsplug._utils` module and I would be more than fine with that.

snejus: I understand the concern about interface segregation, but looking at the alternative, I strongly disagree that `beetsplug._utils` is the cleaner abstraction.

If we move this validation there, we force every metadata source plugin to import from a module that is currently a "junk drawer" of unrelated concerns. `_utils` contains a wide variety of utils (image manipulation, requests, vfs).

Forcing plugins to depend on that chaotic module just to perform a simple string check is a much stronger violation of separation of concerns than adding a single protected helper method to the base class they already inherit from.

`MetadataSourcePlugin` is the cohesive home for shared behaviour of metadata sources. `is_query_empty` is specific to the domain of these plugins (handling query inputs). Placing it on the base class keeps the dependencies clean and the logic where it belongs, rather than scattering it into a global utility bucket.

semohr: > If we move this validation there, we force every metadata source plugin to import from a module that is currently a "junk drawer" of unrelated concerns. _utils contains a wide variety of utils (image manipulation, requests, vfs).

The core issue for me is that **not every metadata plugin** actually needs this helper. If all metadata sources depended on it, then abstracting it into core would make sense. But that’s not the case thus I’m hesitant to add it to the main interface. Interfaces should define the minimal, common contract, not accumulate convenience helpers that only some implementations need.

> MetadataSourcePlugin is the cohesive home for shared behaviour of metadata sources.

At the moment, `MetadataSourcePlugin` is intentionally designed as a boundary between external plugin code and the beets core. After the split from `BeetsPlugin`, we were quite deliberate about moving any shared functionality that isn’t required by the core layer into more specific abstractions like `APIMetadataSourcePlugin`.

---

I completely understand the motivation here, having shared helpers in a common namespace often does feel like the right direction. However, in this case, adding common functions directly to the interface risks muddying the separation between the plugin layer and the core layer. For me that separation has been a conscious design goal with the split into `MetadataSourcePlugin`.

Issue #6066, opened 2025-10-03
Musicbrainz broken in 2.4.0


natchapman:
### Problem
Albums which imported automatically under 2.3.1 are not found or are not strong candidates under 2.4.0

```sh
[user@host music]$ beet -vv import -t phone/
no user configuration found at /home/user/.config/beets/config.yaml
data directory: /home/user/.config/beets
plugin paths: []
Loading plugins: musicbrainz
Sending event: pluginload
library database: /home/user/.config/beets/library.db
library directory: /home/user/Music
Sending event: library_opened
Sending event: import_begin
Sending event: import_task_created
Sending event: import_task_start
Looking up: /home/user/downloads/music/phone
Tagging Lemon Demon - Spirit Phone
No album ID found.
Search terms: Lemon Demon - Spirit Phone
Album might be VA: False
musicbrainz: Searching for MusicBrainz releases with: {'release': 'spirit phone', 'alias': 'spirit phone', 'tracks': '27', 'artist': 'lemon demon'}
musicbrainz: Requesting MusicBrainz release 80f33b28-89b3-4ec7-bf30-bc7c943104c2
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_album_extract
Sending event: albuminfo_received
Candidate: Lemon Demon - Spirit Phone (80f33b28-89b3-4ec7-bf30-bc7c943104c2)
Computing track assignment...
...done.
Success. Distance: 0.12
musicbrainz: Requesting MusicBrainz release 2124ff41-7e9a-421f-98a9-99405dca7a24
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
[...]

PR #6077, opened 2025-10-06
Fix autotagger results penalisation due to the data source field

snejus:
# PR Summary

This PR refactors the metadata source plugin architecture to fix incorrect data source penalty calculations during import matching.

**Key Changes:**

- **Fixed distance calculation**: Removed `track_distance()` and `album_distance()` methods from `MetadataSourcePlugin`. Data source mismatch penalty is now calculated directly in `beets/autotag/distance.py`, same as most of the rest of field-based penalties.

- **Plugin registration**: Metadata source plugins now register themselves in `_instance_by_data_source` dict on instantiation, enabling efficient penalty lookups.

- **Renamed configuration options**:
  - `source_weight` → `data_source_mismatch_penalty` (plugin-level)
  - `match.distance_weights.source` → `match.distance_weights.data_source` (global matching)

- **Documentation**: added a `yaml` block with default configuration values to each metadata source documentation page. Included the renamed penalty field in each to underline that it's availeble for each data source.


test/autotag/test_distance.py
```diff
@@ -297,3 +293,48 @@ def test_solo_pattern(self):
...
+    @pytest.mark.parametrize(
+        "item,info,penalty,weight,expected_distance",
+        [
+            _p("first", "first", 0.5, 1.0, 0.0, id="match"),
+            _p("first", "second", 0.5, 1.0, 0.125, id="mismatch"),
+            _p("first", "unknown", 0.5, 1.0, 0.125, id="unknown-mismatch"),
+            _p("first", None, 0.5, 1.0, 0.125, id="no-info"),
+            _p(None, "first", 0.5, 1.0, 0.0, id="none-originally-first-matches"),  # noqa: E501
+            _p(None, "second", 0.5, 1.0, 0.0, id="none-originally-second-matches"),  # noqa: E501
+            _p("unknown", "unknown", 0.5, 1.0, 0.0, id="unknown-match"),
```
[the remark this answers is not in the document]
snejus: You either stop trying to overengineer literally everything you see or we'll have to part ways.

beets/metadata_plugins.py
```diff
@@ -145,14 +93,28 @@ class MetadataSourcePlugin(BeetsPlugin, metaclass=abc.ABCMeta):
     and tracks, and to retrieve album and track information by ID.
     """
 
+    @cached_classproperty
```
semohr, line 100: Any reason for moving this above the init?

beets/metadata_plugins.py
```diff
@@ -145,14 +93,28 @@ class MetadataSourcePlugin(BeetsPlugin, metaclass=abc.ABCMeta):
...
+    def data_source_mismatch_penalty(self) -> float:
+        return self.config["data_source_mismatch_penalty"].as_number()
+
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         self.config.add(
             {
                 "search_limit": 5,
-                "source_weight": 0.5,
+                "data_source_mismatch_penalty": 0.5,
```
semohr: I think we should still have backwards compatibility for `source_weight`. Maybe schedule deprecation for 3.0.0?

Also shouldn't the new default be 0.0 instead of 0.5?

beets/metadata_plugins.py
```diff
@@ -18,54 +19,27 @@
...
-                "It should extend MetadataSourcePlugin instead of BeetsPlugin. "
-                "Support for this will be removed in the v3.0.0 release!",
-                DeprecationWarning,
-                stacklevel=2,
-            )
-            metadata_plugins.append(plugin)
-
-    # typeignore: BeetsPlugin is not a MetadataSourcePlugin (legacy support)
-    return metadata_plugins  # type: ignore[return-value]
+_instance_by_data_source: dict[str, MetadataSourcePlugin] = {}
```
semohr: To be honest, the whole `_instance_by_data_source` machinery feels a bit over-engineered to me and introduces complexity which might bite us in the long term. If we really think caching is necessary for the lookup (which I’m not entirely sure about, but I’ll trust your judgment), couldn’t we just replace the whole thing with a `@functools.cache` decorator on a single function call?


That would be much cleaner, and we wouldn’t have to manually handle cache invalidation. In normal beets workflows, loaded plugins shouldn’t change anyway.

Edit: Just noticed you did the same for the `get_penalty` function.

beets/metadata_plugins.py
```diff
@@ -87,54 +61,28 @@ def track_for_id(_id: str) -> TrackInfo | None:
...
-    """
-    from beets.autotag.distance import Distance
-
-    dist = Distance()
-    for plugin in find_metadata_source_plugins():
-        dist.update(plugin.track_distance(item, info))
-    return dist
-
+@cache
+def get_penalty(data_source: str | None) -> float:
```
semohr, line 78: Very nice!

beets/plugins.py
```diff
@@ -160,12 +161,47 @@ class BeetsPlugin(metaclass=abc.ABCMeta):
...
+            "Support for this will be removed in the v3.0.0 release!",
+            DeprecationWarning,
+            stacklevel=3,
+        )
+
+        original_init = cls.__init__
+
+        def init(self, *args, **kwargs):
+            original_init(self, *args, **kwargs)
+            metadata_plugins._instance_by_data_source[self.data_source] = self
```
semohr: The enhanced comment is nice! In general I'm not convinced we need an additional registry for metadata plugins. We shouldn’t have too many plugins active at once, so looping over all of them shouldn’t add significant overhead. I’m not sure the added complexity is worth it. How do you see this?

docs/changelog.rst
```diff
@@ -62,12 +66,22 @@ Other changes:
...
+- :ref:`match-config`: ``match.distance_weights.source`` configuration has been
+  renamed to ``match.distance_weights.data_source`` for consistency with the
+  name of the field it refers to.
 
 For developers and plugin authors:
 
 - Typing improvements in ``beets/logging.py``: ``getLogger`` now returns
   ``BeetsLogger`` when called with a name, or ``RootLogger`` when called without
   a name.
+- **Breaking Change**: The ``track_distance()`` and ``album_distance()`` methods
```
semohr: Do we want to deprecate this functions instead? I see them as internal api, might not be necessary.

beets/config_default.yaml
```diff
@@ -166,13 +166,14 @@ match:
...
     distance_weights:
-        source: 2.0
+        data_source: 2.0
         artist: 3.0
         album: 3.0
         media: 1.0
         mediums: 1.0
         year: 1.0
         country: 0.5
+        data_source: 0.5
```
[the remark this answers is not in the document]
semohr: Copilot has a point here 🙃

beets/metadata_plugins.py
snejus: It's just the simplest version of the plugin registry pattern, see https://www.perplexity.ai/search/explain-the-registry-pattern-f-bH5iHV4jTgSrDgk7KKhvlg#0

> That would be much cleaner, and we wouldn’t have to manually handle cache invalidation
Where do we do this?

semohr: We already have a registry in `plugins.py`, which we’re essentially just wrapping here. I don’t think it’s strictly necessary to maintain a separate one, metadata plugins are just a subset of the normal plugins. Or do you have a plan for metadata plugins that I’m not aware of?

It seems simpler to rely on cached functions if performance is a concern and we want to avoid looping through all plugins. Is that why this was introduced? (Although I’m not entirely convinced this is a real performance issue, given that the number of plugins is limited.)

```python
@cache
def meta_plugin_by_name(name: str):
    pass

@cache
def all_metaplugins():
    pass
```

beets/metadata_plugins.py
snejus: Two reasons:

1. Community-accepted best practices re expected order of definitions https://www.perplexity.ai/search/python-best-practices-what-s-t-eZHB065RTc.AaJy6Y8Htkg#0
2. Following the order these are defined in practice: 
   1. `@cached_classproperty` is a _class_ attribute (just a dynamic one) so it's available on the class and doesn't depend on the instance being created
   2. `@property` is an _instance_ attribute
   3. `__init__` defines how an instance is created from this class (it can make use of class properties which makes sense, however, confusingly, it can also make use `@property`ies thus I define it below them).

beets/metadata_plugins.py
snejus: > I think we should still have backwards compatibility for source_weight. Maybe schedule deprecation for 3.0.0?

Good point, added it

snejus: > Also shouldn't the new default be 0.0 instead of 0.5?

Once I fixed the logic, I realised that the default should indeed be non-zero: we do want to penalize matches from a non-matching data source.

beets/metadata_plugins.py
snejus: When I investigated this bug, I was taken aback seeing how often this function gets called in reality:

```py
    print("Finding metadata source plugins...")
```

```
user configuration: /home/sarunas/.config/beets/config.yaml
data directory: /home/sarunas/.config/beets
plugin paths: []
Loading plugins: autobpm, bandcamp, bareasc, bench, deezer, discogs, edit, embedart, fetchart, importreplace, info, inline, lastimport, lyrics, mbcollection, mbsync, missing, mpdstats, mpdupdate, musicbrainz, random, replaygain, scrub, spotify, types, unimported
inline: adding item field has_lyrics
inline: adding item field label_or_albumartist
inline: adding item field singleton_track_artist
inline: adding item field track_artist
inline: adding item field album_name
inline: adding item field track_identification
inline: adding item field withdrawn
inline: adding album field label_or_albumartist
inline: adding album field multiple_artists
Sending event: pluginload
library database: /home/sarunas/.music/beets/library.db
library directory: /run/media/sarunas/music/Music
Sending event: library_opened
Sending event: import_begin
Parsed query: AndQuery([SubstringQuery('album', 'Signs of a struggle', fast=True)])
Parsed sort: NullSort()
yielding album 12085: Mattafix - Signs of a Struggle
Looking up: /run/media/sarunas/music/Music/emiinternational/mattafix_signsofastruggle/02_bigcitylife.mp3
Tagging Mattafix - Signs of a Struggle
Searching for discovered album ID: 4a6d6b9b-02f8-413c-a672-9bbe535b36d8
Finding metadata source plugins...
bandcamp: Not a bandcamp URL, skipping
deezer: Deezer API error: no data
discogs: Searching for release 4a6d6b9b-02f8-413c-a672-9bbe535b36d8
musicbrainz: Requesting MusicBrainz release 4a6d6b9b-02f8-413c-a672-9bbe535b36d8
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_album_extract
Candidate: Mattafix - Signs of a Struggle (4a6d6b9b-02f8-413c-a672-9bbe535b36d8)
Computing track assignment...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
...done.
Finding metadata source plugins...
Finding metadata source plugins...
Success. Distance: 0.00
Album ID match recommendation is 3
Search terms: Mattafix - Signs of a Struggle
Album might be VA: False
Finding metadata source plugins...
bandcamp: Searching releases of type 'a' for query 'Mattafix - Signs of a Struggle' using '{'query': 'Mattafix - Signs of a Struggle', 'artist': 'Mattafix', 'label': 'EMI International', 'search_type': 'a'}'
deezer: Searching Deezer for 'album:'Signs of a Struggle' artist:'Mattafix''
deezer: Found 1 result(s) from Deezer for 'album:'Signs of a Struggle' artist:'Mattafix''
Candidate: Mattafix - Signs Of A Struggle (678437231)
Computing track assignment...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
...done.
Finding metadata source plugins...
Finding metadata source plugins...
Success. Distance: 0.41
discogs: Getting master release 103390
Candidate: Mattafix - Signs Of A Struggle (1616194)
Computing track assignment...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
...done.
Finding metadata source plugins...
Finding metadata source plugins...
Success. Distance: 0.32
spotify: Searching Spotify for 'album:'Signs of a Struggle' artist:'Mattafix''
spotify: Found 1 result(s) from Spotify for 'album:'Signs of a Struggle' artist:'Mattafix''
Candidate: Mattafix - Signs Of A Struggle (57m6FbCoUXfTxw6REFc3RZ)
Computing track assignment...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
Finding metadata source plugins...
...done.
Finding metadata source plugins...
Finding metadata source plugins...
Success. Distance: 0.30
musicbrainz: Searching for MusicBrainz releases with: {'release': 'signs of a struggle', 'alias': 'signs of a struggle', 'tracks': '1', 'artist': 'mattafix'}
musicbrainz: Requesting MusicBrainz release 4a6d6b9b-02f8-413c-a672-9bbe535b36d8
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_track_extract
Sending event: mb_album_extract
Candidate: Mattafix - Signs of a Struggle (4a6d6b9b-02f8-413c-a672-9bbe535b36d8)
Duplicate.
Evaluating 4 candidates.
```

snejus: > ca73d9cbc...

I agree - made adjustments keeping the original `find_metadata_source_plugins` in place!

beets/plugins.py
snejus: Absolutely agree. This complexity has now been removed - the only change I made was to move the deprecation warning from `metadata_plugins.find_metadata_source_plugins` to here.

docs/changelog.rst
snejus: Should have verified this AI-generated bit with more scrutiny 😅

beets/util/__init__.py
```diff
@@ -836,9 +836,10 @@ def get_most_common_tags(
         "country",
         "media",
         "albumdisambig",
+        "data_source",
     ]
     for field in fields:
-        values = [item[field] for item in items if item]
+        values = [item.get(field) for item in items if item]
```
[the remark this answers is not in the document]
snejus: What a hallucination - going straight to the hall of fame.

The only change in behaviour is the fact that this happily returns `None` for keys that aren't present in `item` instead of throwing an exception.

snejus:
@semohr it's ready for a re-review.

beets/autotag/distance.py
```diff
@@ -526,6 +529,9 @@ def distance(
         dist.add("unmatched_tracks", 1.0)
 
     # Plugins.
-    dist.update(metadata_plugins.album_distance(items, album_info, mapping))
-
+    if (
+        likelies["data_source"]
```
asardaes: Sorry if I'm missing something, but isn't this always `None` since `data_source` isn't considered in `get_most_common_tags`?

Even if `get_most_common_tags` changes, is `data_source` set for `Item` in any scenario?

I thought the idea of the weights was to control preference between multiple sources, not really to penalize mismatches.

semohr: > Sorry if I'm missing something, but isn't this always None since data_source isn't considered in get_most_common_tags?

`data_source` was added to the list in `get_most_common_tags` in this PR.

> Even if get_most_common_tags changes, is data_source set for Item in any scenario?

Yes it is set :) Im too lazy to lookup where this is set exactly but quickly looked into my database and the `item_attributes` table includes quite some `data_source` fields. 

> I thought the idea of the weights was to control preference between multiple sources, not really to penalize mismatches.

I thought so too initially but this was never implemented this way. It was always implemented a penalty for the `data_source` string. I think it should be implemented different too but this is another PR in my opinion. See my idea on this [here](https://github.com/beetbox/beets/issues/6066#issuecomment-3368925412)

snejus: > I thought the idea of the weights was to control preference between multiple sources, not really to penalize mismatches.

_preferences between multiple sources_ are indeed implemnted by **penalizing mismatches** under the hood.

Note that two configuration variables are in play here. One for the **weight** and one for the **penalty**, in a simplified form, we have:

```py
data_source_dist = (
    config["match"]["distance_weights"]["data_source"]
    *
    config[data_source]["data_source_mismatch_penalty"]
)
```

Let's compare this to, say, the logic behind matching the `catalognum` field:

```py
def string_penalty(before, after):
    """Normalise difference between strings into the range [0.0, 1.0],
    0.0 meaning 'strings are the same' and 1.0 meaning 'strings are completely different'
    """

catalognum_dist = (
    config["match"]["distance_weights"]["catalognum"]
    *
    string_penalty(expected_catalognum, actual_catalognum)
)
```

Essentially, the penalty for every other field is **dynamically calculated** while the penalty for _data_source_ is **static** for a certain data source. This is exactly how we are able to express preferences for certain data sources.

asardaes: Ah I don't know how I missed the change to `get_most_common_tags`. Nevertheless, it makes sense to me that that attribute is set in the database, but during the auto-tagging process the `Item`s refer to files that usually don't have such tags, no? A new file that has never been in the DB and is about to be imported cannot have a data source in the `beets` sense.

semohr: In theory `data_source` can be set as custom attribute in the audiofile, right?

beets/metadata_plugins.py
```diff
@@ -145,12 +92,38 @@ class MetadataSourcePlugin(BeetsPlugin, metaclass=abc.ABCMeta):
...
+
+    @cached_property
+    def data_source_mismatch_penalty(self) -> float:
+        try:
+            source_weight = self.config["source_weight"].as_number()
+        except NotFoundError:
+            return self.config["data_source_mismatch_penalty"].as_number()
+
+        # TODO: Remove in v3.0.0
+        warnings.warn(
```
semohr: Can we do a normal log.warning message too? Warnings are not shown for most normal users.

docs/plugins/index.rst
```diff
@@ -47,21 +47,33 @@ some, you can use ``pip``'s "extras" feature to install the dependencies:
...
-  disable. Default: ``0.5``.
+.. _data_source_mismatch_penalty:
 
-For example, to equally consider matches from Discogs and MusicBrainz add the
-following to your configuration:
+- **data_source_mismatch_penalty**: Penalty applied to matches during import.
+  Any decimal number between 0 and 1. Default: ``0.5``.
 
-.. code-block:: yaml
+  Penalize this data source to prioritize others. For example, to prefer Discogs
```
semohr: Is this really how it works? From the docstring, it sounds like if I have two identical matches, one from `spotify` and one from `discogs`  - `spotify` would be preferred if its penalty is smaller. But that’s not actually the case here, right?

As I understand it, the penalty is only applied when the `data_source` doesn’t match, and it’s just a flat offset added to the total distance score.

Am I interpreting that correctly? I might have confused myself here.

beets/autotag/distance.py
asardaes: I'm personally not sure, and even then, I imagine it won't happen commonly in practice. Don't get me wrong, if you believe the functionality is not working, I get that, but this new behavior might end up being pointless, maybe it would be better to just remove it.

I will mention that what I've been experimenting with in #5888 made use of source weights and it actually worked as expected in my tests.

snejus: > In theory data_source can be set as custom attribute in the audiofile, right?

It can't. Custom attributes only exist in beets database. @asardaes is right here:

> A new file that has never been in the DB and is about to be imported cannot have a data source in the beets sense

For file that is not in beets database yet

```py
likelies = {
    ...
    "data_source": None
}
```

This is why we add a distance penalty for `data_source` on condition `if likelies["data_source"]`.


> but this new behavior might end up being pointless, maybe it would be better to just remove it.

This behaviour is neither new nor pointless - it only affects **reimports** and it's been broken until this adjustment (please read #6066 in detail).

docs/plugins/index.rst
asardaes: I think this is what I was thinking about for new imports. If you have multiple metadata sources configured and you're importing files that have never been in the DB, `data_source` won't be set for any `Item`, so its penalties won't apply, so it won't be possible to prioritize spotify over discovs or viceversa.

snejus: I updated the documentation, hopefully it is a bit more clear now.

> if I have two identical matches, one from spotify and one from discogs - spotify would be preferred if its penalty is smaller. But that’s not actually the case here, right?

This essentially describes the behaviour that we had before this change! 

> As I understand it, the penalty is only applied when the data_source doesn’t match

Indeed

> and it’s just a flat offset added to the total distance score.

The configured `data_source_mismatch_penalty` multiplied by the configured `match.distance_weights.data_source`.

asardaes: I have to disagree, I don't think this is only about reimports, you're removing the possibility of prioritizing sources for new imports; I believe setting all source weights/penalties to 0 by default should be the (non-temporary) solution, there's already confirmation that that works and doesn't remove existing functionality.

snejus: Setting this configuration to 0 would completely change the previous (and expected) behaviour where data sources are penalised by default (the issue was that this penalisation was a bit too eager, and it penalised even the candidates that had matching data sources). I don't seem to understand how _removing prioritization altogether_ fixes the issue you mentioned?

> you're removing the possibility of prioritizing sources for new import

It seems to me we can instead resolve this by **not ignoring** the absence of `data_source` field (treat `None` as a legitimate value). This way, every data source gets penalised, taking configured preferences into account. Is this the behaviour you'd expect?

asardaes: I think that, to truly keep previous behavior, maybe a bit of both is needed. I do think None would have to be taken into account, but additionally the weight of _only_ MusicBrainz should be 0 by default? I'm not sure about that, but I saw you mentioned in the other thread that the issue didn't appear before because MusicBrainz wasn't a plugin, so its penalization was effectively 0 because it wasn't considered, no?

snejus: Once we've made MusicBrainz a plugin like every other data source, people realised that they can't any more see 100% matches for their musicbrainz candidates.

Once I dug deeper, I found out about this

> so its penalization was effectively 0 because it wasn't considered

which uncovered a fundamental flaw in this data source matching implementation: it had always been unfairly biased, preferring candidates from musicbrainz. The previous implementation never recognised and matched candidates from any other data source correctly.

Personally, I'd noticed some weird bias towards musicbrainz years ago but could never pinpoint what exactly caused it. This [frustrated a fair bit of users](https://github.com/beetbox/beets/issues/2686) and led to me adding the ability to disable musicbrainz, and ultimately making it a separate plugin.

snejus: I hope this clarifies why we're not keeping the previous behaviour regarding

> but additionally the weight of only MusicBrainz should be 0 by default?

beets/metadata_plugins.py
snejus: This would mean that users see this warning even when it's not their fault: this setting may be applied by an external data source plugin.

I added logic to check whether this config is coming from the user or a plugin.

snejus:
@sourcery-ai review

beets/metadata_plugins.py
```diff
@@ -145,12 +91,29 @@ class MetadataSourcePlugin(BeetsPlugin, metaclass=abc.ABCMeta):
...
+        This is inferred from the plugin name.
+        """
+        return cls.__name__.replace("Plugin", "")  # type: ignore[attr-defined]
+
+    @cached_property
+    def data_source_mismatch_penalty(self) -> float:
+        try:
+            return self.config["source_weight"].as_number()
+        except NotFoundError:
+            return self.config["data_source_mismatch_penalty"].as_number()
```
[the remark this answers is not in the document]
snejus: Deprecation warning is raised using `BeetsPlugin.verify_config`.

docs/plugins/index.rst
arsaboo: I can 100% agree with the bias towards MusicBrainz.

snejus: @asardaes @semohr @arsaboo see [this comment](https://github.com/beetbox/beets/issues/6066#issuecomment-3384389499) where I asked users to vote for the behaviour they prefer regarding the very first import. Please cast your vote!

beets/metadata_plugins.py
semohr: I think we had this discussion before and I'm more in favour of sorting by importance. I don't care too much about it here but I personally wont conform to this if it is not strictly enforced by a linter 😆

beets/metadata_plugins.py
```diff
@@ -23,36 +24,13 @@
...
-                "Support for this will be removed in the v3.0.0 release!",
-                DeprecationWarning,
-                stacklevel=2,
-            )
-            metadata_plugins.append(plugin)
-
-    # typeignore: BeetsPlugin is not a MetadataSourcePlugin (legacy support)
-    return metadata_plugins  # type: ignore[return-value]
+    """Return a list of all loaded metadata source plugins."""
+    return [p for p in find_plugins() if hasattr(p, "data_source")]  # type: ignore[misc]
```
semohr, line 34: This might be a bit dangerous if we migrate to v3.0.0 and could yield false positives. I think the instance check would be cleaner in the future but I see why you changed it here. 

Do you mind readding a comment that we consider changing this to an `isinstance` check for v3.0.0?

snejus: Oh yeah, good point. This needs to be a TODO

docs/plugins/index.rst
```diff
@@ -47,21 +47,68 @@ some, you can use ``pip``'s "extras" feature to install the dependencies:
...
 
-For example, to equally consider matches from Discogs and MusicBrainz add the
-following to your configuration:
+- **data_source_mismatch_penalty**: Penalty applied when the data source of a
+  match candidate differs from the original source of your existing tracks. Any
+  decimal number between 0.0 and 1.0. Default: ``0.5``.
 
-.. code-block:: yaml
+  This setting controls how much to penalize matches from different metadata
+  sources during import. The penalty is applied when beets detects that a match
```
semohr, line 60: Can we make `beets detects that a match
  candidate comes from a different t data source` bold?

snejus: Is it necessary? I understand that the internal implementation is not easy to grasp, but regarding above, I think, just the name of this configuration option makes things crystal clear. 😅

semohr: It should be mostly clear👍I think this is the main takeaway from the section, might help some people who only skim the docs. Do as you see fit.

PR #6090, opened 2025-10-12
Add custom feat words for ftintitle

EmberLightVFX:
## Description

For non English tracks (Swedish in my case) feat. words might be something that ftintitle doesn't pick up.
Eg. for the song `Promoe med Afasi - Inflation` [https://musicbrainz.org/recording/8e236347-61d6-4e11-9980-52f4cc6b905f](https://musicbrainz.org/recording/8e236347-61d6-4e11-9980-52f4cc6b905f) the word `med` is `feat.` in Swedish.
With this PR you can add what ever word you wish to match as feat. so it should cover any kind of language.

The config.yaml could look like this:
ftintitle:
  custom_feat_words: ["med"]

## To Do

### DIFF
--- a/docs/plugins/ftintitle.rst
+++ b/docs/plugins/ftintitle.rst
@@ -28,8 +28,8 @@ file. The available options are:
 - **keep_in_artist**: Keep the featuring X part in the artist field. This can be
   useful if you still want to be able to search for features in the artist
   field. Default: ``no``.
-- **custom_words**: List of additional words that will be treated as a marker for
-  artist features. Default: ``[]``.
+- **custom_words**: List of additional words that will be treated as a marker
+  for artist features. Default: ``[]``.
 
 Running Manually
 ----------------


docs/changelog.rst
```diff
@@ -9,6 +9,8 @@ Unreleased
 
 New features:
 
+- Added argument for custom feat. words in ftintitle.
```
henry-oberholtzer: We can add the specific plugin documentation link to this.

```yaml

:doc:`plugins/fitintitle`: Added config option for custom words.
```

docs/plugins/ftintitle.rst
```diff
@@ -28,6 +28,8 @@ file. The available options are:
 - **keep_in_artist**: Keep the featuring X part in the artist field. This can be
   useful if you still want to be able to search for features in the artist
   field. Default: ``no``.
+- **custom_feat_word**: Add custom words to the feat list; any words you add will
+  also be treated as "feat" tokens. Default: ``[]``.
```
henry-oberholtzer: I think we can do a simpler word for the config option - maybe just `custom_words`. 'feat tokens' doesn't mean a lot to the user, so maybe more along the lines of

```yaml
**custom_words**: List of additional words that will be treated as a marker for artist features.
```

beetsplug/ftintitle.py
```diff
@@ -120,10 +137,13 @@ def func(lib, opts, args):
...
             drop_feat = self.config["drop"].get(bool)
             keep_in_artist_field = self.config["keep_in_artist"].get(bool)
+            custom_feat_words = self.config["custom_feat_words"].get(list)
             write = ui.should_write()
 
             for item in lib.items(args):
-                if self.ft_in_title(item, drop_feat, keep_in_artist_field):
+                if self.ft_in_title(
+                    item, drop_feat, keep_in_artist_field, custom_feat_words
+                ):
```
henry-oberholtzer, line 142: I'd say that we could use the class data members for these member functions, but it appears the original plugin isn't really written to make full use of that either.

If you'd like to restructure it to make more use that with member functions and data members, that'd be super, but I don't think it's necessary this time.

docs/plugins/ftintitle.rst
EmberLightVFX: Yeah that's waaay better

beetsplug/ftintitle.py
EmberLightVFX: As this is my first PR to beets and I'm not super used to the code base I think it's better if someone else would do it so I don't mess anything up.

docs/changelog.rst
```diff
@@ -9,6 +9,8 @@ Unreleased
 
 New features:
 
+- :doc:`plugins/fitintitle`: Added argument for custom feat. words in ftintitle.
```
henry-oberholtzer: looks like github actions ran into a hiccup here! `plugins/ftintitle` should fix it.

EmberLightVFX:
Ran all the poe lines and fixed the misspelling!

I did get an error when running `poe check-types`:
```
Poe => mypy --pretty
usage: mypy [-h] [-v] [-V] [more options; see below]
            [-m MODULE] [-p PACKAGE] [-c PROGRAM_TEXT] [files ...]
mypy: error: Missing target module, package, files, or command.
```

I'm running python 3.9.13 and latest poetry.
I did modified the task to run `cmd = "mypy ."` and got `Success: no issues found in 216 source files` so I guess it's all ok!

henry-oberholtzer:
Great! Merged in, thanks for the PR!

PR #6101, opened 2025-10-14
Legacy plugin copy not copying properties.

semohr:
The new `data_source_mismatch_penalty` property in the MetadataPlugin class was not copied in the backwards compatibility layer. This PR introduces a fixes this such that `cached_properties` are copied to legacy metadata plugins.

This also includes a test for the expected behavior.

See also [beetcamp issue](https://github.com/snejus/beetcamp/issues/85#issuecomment-3399273892).

Note: I wanted to use the predicate but it is difficult to get the name of a cache property in the predicate function. Using a loop was the easier solution imo.

### DIFF
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -192,24 +192,21 @@ class BeetsPlugin(metaclass=abc.ABCMeta):
             stacklevel=3,
         )
 
+        method: property | cached_property[Any] | Callable[..., Any]
         for name, method in inspect.getmembers(
             MetadataSourcePlugin,
-            predicate=lambda f: (
+            predicate=lambda f: (  # type: ignore[arg-type]
                 (
-                    isinstance(f, cached_property)
-                    and f.attrname is not None
-                    and not hasattr(BeetsPlugin, f.attrname)
-                )
-                or (
-                    isinstance(f, property)
-                    and f.fget is not None
-                    and f.fget.__name__ is not None
-                    and not hasattr(BeetsPlugin, f.fget.__name__)
+                    isinstance(f, (property, cached_property))
+                    and not hasattr(
+                        BeetsPlugin,
+                        getattr(f, "attrname", None) or f.fget.__name__,  # type: ignore[union-attr]
+                    )
                 )
                 or (
                     inspect.isfunction(f)
                     and f.__name__
-                    not in MetadataSourcePlugin.__abstractmethods__
+                    and not getattr(f, "__isabstractmethod__", False)
                     and not hasattr(BeetsPlugin, f.__name__)
                 )
             ),


beets/plugins.py
```diff
@@ -192,15 +192,21 @@ def __init_subclass__(cls) -> None:
...
+            if hasattr(cls, name) or name in abstracts:
+                continue
+
+            # Copy functions, methods, and properties
+            if (
+                inspect.isfunction(method)
+                or inspect.ismethod(method)
+                or isinstance(method, cached_property)
+            ):
+                setattr(cls, name, method)
```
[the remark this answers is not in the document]
semohr: Not needed yet. We may add them once we have a need for it.

test/test_plugins.py
```diff
@@ -523,3 +523,23 @@ def test_import_plugin(self, caplog, plugin_name):
         assert "PluginImportError" not in caplog.text, (
             f"Plugin '{plugin_name}' has issues during import."
         )
+
+
+class TestDeprecationCopy:
```
snejus, line 528: I'm a big fan of testing this kind of thing in action:

```diff
diff --git a/test/autotag/test_distance.py b/test/autotag/test_distance.py
index b327bbe44..22f2e7937 100644
--- a/test/autotag/test_distance.py
+++ b/test/autotag/test_distance.py
@@ -11,6 +11,7 @@
 )
 from beets.library import Item
 from beets.metadata_plugins import MetadataSourcePlugin, get_penalty
+from beets.plugins import BeetsPlugin
 from beets.test.helper import ConfigMixin
 
 _p = pytest.param
@@ -310,8 +311,8 @@
             def candidates(self, *args, **kwargs): ...
             def item_candidates(self, *args, **kwargs): ...
 
-        class OriginalPlugin(TestMetadataSourcePlugin):
-            pass
+        class OriginalPlugin(BeetsPlugin):
+            data_source = "Original"
 
         class OtherPlugin(TestMetadataSourcePlugin):
             @property
@@ -332,6 +333,7 @@
         [
             _p("Original", "Original", 0.5, 1.0, True, MATCH, id="match"),
             _p("Original", "Other", 0.5, 1.0, True, MISMATCH, id="mismatch"),
+            _p("Other", "Original", 0.5, 1.0, True, MISMATCH, id="mismatch"),
             _p("Original", "unknown", 0.5, 1.0, True, MISMATCH, id="mismatch-unknown"),  # noqa: E501
             _p("Original", None, 0.5, 1.0, True, MISMATCH, id="mismatch-no-info"),  # noqa: E501
             _p(None, "Other", 0.5, 1.0, True, MISMATCH, id="mismatch-no-original-multiple-sources"),  # noqa: E501
```

semohr: I don't have an issue with this as long as we add a docstring with reasoning. Added it in the latest commit

PR #6102, opened 2025-10-16
Ftintitle: Continue even if albumartist and artist is the same

EmberLightVFX:
## Description

This small PR allows ftintitle to process even if the artist/s in the artist and albumartist fields are the same.

This fixes the problem with a lot of singles like [Porter Robinsons song Shelter](https://musicbrainz.org/release/ccc261b9-e4cc-4965-81b8-7c92a5d28601) and even [Rihanas's album Umbrella](https://musicbrainz.org/release/60f8f1f5-485b-4637-8574-23f2bb98531f)

Without this fix the songs would end up with the feat. artist in the artists folder-name and not the feat. in the songs filename.
Without:
`Rihanna feat. JAY‐Z\(2007) Umbrella\01 - Umbrella (radio edit).flac`
`Porter Robinson feat. Madeon\(2016) Shelter\01 - Shelter.flac`
With:
`Rihanna\(2007) Umbrella\01 - Umbrella (radio edit) feat. JAY‐Z.flac`
`Porter Robinson\(2016) Shelter\01 - Shelter feat. Madeon.flac`

I left the current way ftintitle works as the default so stuff doesn't randomly change for users, but maybe it should is changed as the PR that changed the ftintitle's behavour is only ~2 month old https://github.com/beetbox/beets/pull/5943
Thoughts?

I'm also not super happy with the args name `skip_if_artist_and_album_artists_is_the_same` so any suggestion what it could be instead is more than welcome 😅 

## To Do

### DIFF
--- a/docs/plugins/ftintitle.rst
+++ b/docs/plugins/ftintitle.rst
@@ -28,8 +28,8 @@ file. The available options are:
 - **keep_in_artist**: Keep the featuring X part in the artist field. This can be
   useful if you still want to be able to search for features in the artist
   field. Default: ``no``.
-- **preserve_album_artist**: If the artist and the album
-  artist are the same, skip the ftintitle processing. Default: ``yes``.
+- **preserve_album_artist**: If the artist and the album artist are the same,
+  skip the ftintitle processing. Default: ``yes``.
 - **custom_words**: List of additional words that will be treated as a marker
   for artist features. Default: ``[]``.
 


beetsplug/ftintitle.py
```diff
@@ -133,12 +134,19 @@ def func(lib, opts, args):
             self.config.set_args(opts)
             drop_feat = self.config["drop"].get(bool)
             keep_in_artist_field = self.config["keep_in_artist"].get(bool)
+            skip_if_artist_and_album_artists_is_the_same = self.config[
```
henry-oberholtzer: I definitely agree with you - this is a little cumbersome! I think `preserve_album_artist` could be a good alternative, since we're essentially seeing if the album artist matches the track artist, and then skipping.

PR #6108, opened 2025-10-18
Fix/cache validation

HarshitR2004:
## Description

Fixed cache invalidation bug in `Library.add_album()` where the `_memotable` cache was not cleared after adding albums, unlike the `Library.add()` method. This could lead to stale template cache data when adding albums with multiple items.

## To Do

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -13,7 +13,9 @@ New features:
 
 Bug fixes:
 
-- Fixed cache invalidation bug in `Library.add_album()` where the `_memotable` cache was not cleared after adding albums, unlike the `Library.add()` method, potentially causing stale template cache data when adding albums with multiple items.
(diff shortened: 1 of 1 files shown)


semohr:
Thanks for submitting this PR and your interest in contributing to beets.

I took a look at the change and had some questions about the motivation. The original `_memotable` functionality (from #4438) was added for a specific reason related to singleton tracks. I'm not sure the same issue applies to albums, and the PR description doesn't clarify the specific problem this solves.

To help us understand, could you please provide:
- A concrete example of the issue or bug this change fixes when working with albums.
- The specific motivation or use case that led you to make this change.

As your contributions appears to be AI-generated to me, I will take a bit more caution here! Often such PRs  "fix" non-existent issues without a clear understanding of the codebase. These can unintentionally introduce bugs. If that's not the case here, I apologize for the assumption, and I look forward to your clarification so we can move forward.

PR #6110, opened 2025-10-19
Use `Generic` instead of `Any` for `cached_class_property`

amogus07:
## Description

In a plugin, self.data_source currently returns Any instead of str. This change allows `cached_classproperty` to retain the type returned by the decorated method.

## To Do

- [x] ~Documentation~
- [ ] Changelog
- [x] ~Tests~


snejus (approved):
Lovely, thanks for this. I never managed to make this work!

PR #6112, opened 2025-10-19
Fix transaction context manager signature

amogus07:
## Description

Pyrefly reported the contect manager signature of Transaction as invalid, so I corrected it.

## To Do

- [x] ~Documentation~
- [x] ~Tests~

### DIFF
--- a/.github/workflows/changelog_reminder.yaml
+++ b/.github/workflows/changelog_reminder.yaml
@@ -10,7 +10,7 @@ jobs:
   check_changes:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       - name: Get all updated Python files
         id: changed-python-files
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,12 +25,12 @@ jobs:
     env:
       IS_MAIN_PYTHON: ${{ matrix.python-version == '3.9' && matrix.platform == 'ubuntu-latest' }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Install Python tools
         uses: BrandonLWhite/pipx-install-action@v1.0.3
       - name: Setup Python with poetry caching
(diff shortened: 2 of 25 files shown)


snejus:
What was the issue it reported? I wonder why is it not caught by mypy?

snejus:
Sorry - I pressed the wrong button there 😅

amogus07:
> What was the issue it reported? I wonder why is it not caught by mypy?

https://pyrefly.org/en/docs/error-kinds/#bad-context-manager. It came up when I enabled Pyrefly in a plugin repo that uses `with lib.transaction()`. Idk about mypy, but basedpyright didn't catch it either.

semohr:
Seems strange to me as the class had both an enter and exit method. From skimming the [Context Manager Types docs](https://docs.python.org/3/library/stdtypes.html#context-manager-types) it does not seem to that return values are required. Not a bad thing we added a return value but you might want to raise that with the pyrefly people too :shrug:

amogus07:
> Seems strange to me as the class had both an enter and exit method. From skimming the [Context Manager Types docs](https://docs.python.org/3/library/stdtypes.html#context-manager-types) it does not seem to that return values are required. Not a bad thing we added a return value but you might want to raise that with the pyrefly people too 🤷

Tbh I'm not too familiar with context managers, I just tried to make Pyrefly shut up without changing any runtime logic. As I understand it, `__exit__` should return a boolean indicating wether an exception passed as a parameter should be ignored or not. ~It also says in the docs that no exceptions should be raised in `__exit__`, but I'm not sure if I understand that correctly.~ Maybe this needs a closer look...

semohr:
Now I was curious, the official typeshed types allow for None return values. https://github.com/python/typeshed/blob/11e7d904b9745bf33fe5b9b64bcd274ff788b189/stdlib/contextlib.pyi#L87

And no return should be inferred as `None` or am i getting something wrong?

amogus07:
> And no return should be inferred as `None` or am i getting something wrong?

Just checked again, can confirm it works without explicit `None`.

semohr:
Was just out of curiosity :wink: Thanks for the PR!

PR #6114, opened 2025-10-19
make `cached_classproperty` more robust

amogus07:
## Description

continuation of #6110 

- [x] ~Documentation~
- [ ] Changelog
- [ ] Tests

### DIFF
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -64,8 +64,7 @@ if TYPE_CHECKING:
 MAX_FILENAME_LENGTH = 200
 WINDOWS_MAGIC_PREFIX = "\\\\?\\"
 T = TypeVar("T")
-T_co = TypeVar("T_co", covariant=True)
-_R_co = TypeVar("_R_co", covariant=True)
+T2 = TypeVar("T2")
 StrPath = str | Path
 PathLike = StrPath | bytes
 Replacements = Sequence[tuple[Pattern[str], str]]
@@ -1064,7 +1063,7 @@ def par_map(transform: Callable[[T], Any], items: Sequence[T]) -> None:
     pool.join()
 
 
-class cached_classproperty(Generic[T_co, T]):
+class cached_classproperty(Generic[T, T2]):
     """Descriptor implementing cached class properties.
 
     Provides class-level dynamic property behavior where the getter function is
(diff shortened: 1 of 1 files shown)


snejus:
I want to understand what issue does this solve as this balloons its implementation complexity.

amogus07:
@snejus feel free to revert changes you think are unnecessary

> I want to understand what issue does this solve as this balloons its implementation complexity.

The changes mitigate possible race conditions by ensuring thread-safety and aim to prevent misuse by marking the cache as private and introducing `cached_property.clear_cache`. If you think it's too complex and there will never be a situation where the parameters are used, they could be removed along with the related logic.

snejus:
I think, we'd rather solve this when we have an issue. For now, we use it in very limited contexts (simply as a dynamic class property) and I doubt this logic is required (at least for now).

amogus07:
if it's ever needed, just revert the commit above

snejus:
> if it's ever needed, just revert the commit above

Could you clarify?

amogus07:
> > if it's ever needed, just revert the commit above
> 
> Could you clarify?

I removed the optional `owner` and `name` attributes from `clear_cache` to simplify it. So now, always clears the entire cache. So, if those parameters are ever needed, that commit could be reverted.

snejus:
Thank you for the effort here - I appreciate the thoughtfulness behind this work. However, after review, the added complexity doesn't align with how we actually use this decorator.

The original implementation is simple, works correctly for our use cases, and has had no issues in production. Looking at the 24 places where `cached_classproperty` is used across the codebase, the pattern is consistent: read configuration once, compute schema metadata once, return the cached value. While there may be concurrent access when beets runs in threaded mode, these properties are typically accessed during initialization rather than in hot paths, and we've had no reported issues with race conditions. There's no dynamic class creation, and no evidence of the problems this PR aims to solve.

We follow YAGNI principles - we add thread-safety, weak references, or other robustness features when we have concrete evidence they're needed. For now, the 15-line original is the right level of complexity for this codebase. When an actual issue arises, we can revisit.

That said, the `clear_cache()` simplification you made is solid. If you'd like, feel free to open a separate PR with just that change - it'd be easier to review and merge on its own.

Thanks for the work on this.

PR #6116, opened 2025-10-21
docs: Rewrite Handling Paths chapter (pathlib vs utils)

JOJ0:
## Description

Updates the docs chapter "Handling Paths" describing how to modernise old code and intentionally includes historical details. Examples should further guide contributors while refactoring.

## To Do

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -26,6 +26,10 @@ For packagers:
 
 Other changes:
 
+- The documentation chapter :doc:`dev/paths` has been moved to the "For
+  Developers" section and revised to reflect current best practices (pathlib
+  usage).
+
 2.5.1 (October 14, 2025)
 ------------------------
 


semohr:
How do we feel about moving this into the developer documentation? I don't think it should be included in the contribution guide.

CONTRIBUTING.rst
```diff
@@ -289,27 +289,55 @@ documentation formatting and check for any issues.
...
+  correctly expands ``~`` and resolves symlinks.
+- Cross-platform details like path separators, Unicode handling, and long-path
+  support are handled automatically by ``pathlib``.
+
+Examples
+--------
+
+Old style::
+
+    displayable_path(item.path)
```
semohr: Can we use the full `.. code-block:: python` blocks? Otherwise my editor wont pick up highlighting 😢

JOJ0:
> How do we feel about moving this into the developer documentation? I don't think it should be included in the contribution guide. 

Good idea! I keep forgetting n starting to look for it there anyway!! 😂

CONTRIBUTING.rst
JOJ0: done @semohr

JOJ0:
@semohr moved it to the dev docs but now realize that actually this is the plugin writing section. It kind of fits there but is not only related to plugins but a general thing. Have a better idea where to put it or is fine there? Thanks for any feedback on the wording too!

semohr:
Awesome, although why not just create a file in `docs/dev` directly? I don't think the path handling is plugin specific or do you feel like this needs to be under the plugins umbrella for a specific reason?

Otherwise I'm happy with the new sections. I don't think the wording is that important for the dev docs in general. It is a major improvement either way and we can always iterate it if something feels off.

docs/dev/plugins/other/paths.rst
```diff
@@ -0,0 +1,58 @@
...
+  Python (≥3.6) handles this automatically through its wide-character APIs.
+- ``normpath()`` normalized slashes and removed ``./`` or ``..`` parts but did
+  not expand ``~``. It was used mainly for paths from user input or config
+  files.
+- ``bytestring_path()`` converted paths to ``bytes`` for storage in the
+  database. Paths in the database are still stored as bytes today, though there
+  are plans to eventually store ``pathlib.Path`` objects directly.
+- ``displayable_path()`` converted byte paths to Unicode for display or logging.
+
+These utilities remain safe to use when maintaining older code, but new code and
```
semohr: Let's move the how to do it right now to the top. I feel like the flow of this section should be permuted. Maybe something like:
- How we want dev to handle paths right now
- Historic methods you might still find in the codebase
- Examples

snejus:
This makes me think that we should extend the `pathlib.Path` in our codebase and perform the `expanduser` and `resolve` resolution on init automatically. I think this would make it much easier to use it going forward. What do you think?

docs/dev/plugins/other/paths.rst
JOJ0: Yes good point, actually in my first attempts I tried to do it that way but it didn't read too well. I now tried again and think it's getting there, see my 2nd version @semohr , some polishing might still be required. Thanks for further feedabck in advance!

JOJ0: Screenshot 2025-10-25 at 23 25 33

JOJ0:
> This makes me think that we should extend the `pathlib.Path` in our codebase and perform the `expanduser` and `resolve` resolution on init automatically. I think this would make it much easier to use it going forward. What do you think?

You mean for the `.filepath` property? Good idea!

Related question, do we recommend to store as bytes in such cases as well, I just recommended it that way over there: https://github.com/beetbox/beets/pull/4748/files#diff-5ddf2bcdeaa73deda8e015085ca50330be268bd9352cfda7ed68bcd0879b8c51R56

So basically I'm asking: In case we store paths not for `Item`s or `Album`s but anything else, In flexible fields for example. We still want to store that as bytes, right?  (In that particular example item.path returns bytes already, so simply storing it as-is in that flexible field doesnt require wrapping through `bytestring_path`)

JOJ0:
> Awesome, although why not just create a file in `docs/dev` directly? I don't think the path handling is plugin specific or do you feel like this needs to be under the plugins umbrella for a specific reason?

Nope, as mentioned already, I don't think it fits under plugins :-)
 
> Otherwise I'm happy with the new sections. I don't think the wording is that important for the dev docs in general. It is a major improvement either way and we can always iterate it if something feels off.

Not a native-speaker here but IMHO at least it should be halfway good english - even for dev docs :-) I get goosebumps when reading all wrongly worded documentation - no matter if for developers or endusers. Haha! ;-)

You are right we can always change it if something is not 100% correct. 

I finally found a good place for the chapter in the main docs level. Right between "Library Database API" and "Music importer". the former talks about `Item` and `Album` objects, following the paths chapter which talks about the `.path` and `.file_path` properties seems to fit perfectly.

Also I reworded and clarified this paragraph:

Screenshot 2025-10-28 at 07 24 23

Added it to the changelog and merging it now. Thanks for the feedback! @semohr ah but please give me an approval so I can press the button. Thanks!

CONTRIBUTING.rst
```diff
@@ -289,27 +289,55 @@ documentation formatting and check for any issues.
...
+    normpath("~/Music/../Artist")
+    syspath(path)
+
+New style::
+
+    item.filepath
+    Path("~/Music/../Artist").expanduser().resolve()
```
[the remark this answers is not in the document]
JOJ0: We clearly state what normpath does and what expanduser.resolve() does, so... no intention for a 1:1 replacement and I think it's made clear enough to the reader.

PR #6117, opened 2025-10-21
Unified search string construction between albums and items.

semohr:
## Description

Matching albums and singletons has slightly different behavior between singletons and albums. 

Albums: When a user does not specify any search query in the manual prompt, the system defaults to using the corresponding value from the metadata. 
Singletons: Overwrite metadata by any given noneempty search query.

Additionally I noticed that we have an unintentional type error in the `metadata_plugins.candidates` call. Since `*arg` and `**kwargs` are not typed they accepted `artist` and `title/album` of type `str|None` while they should only accept `str`. 

## Changes

This PR aligns both methods and uses a common function to parse the search terms. This intentionally couples these two occasions to hopefully prevent drift in the future and makes clear that we use the same logic in these two places.

Added explicit typehints to `candidates` and `item_candidates` function.

### DIFF
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -236,6 +236,25 @@ def _add_candidate(
     )
 
 
+def _parse_search_terms_with_fallbacks(
+    *pairs: tuple[str | None, str],
+) -> tuple[str, ...]:
+    """Given pairs of (search term, fallback), return a tuple of
+    search terms. If **all** search terms are empty, returns the fallback. Otherwise,
+    return the search terms even if some are empty.
+
+    Examples:
+    (("", "F1"), ("B", "F2")) -> ("", "B")
+    (("", "F1"), ("", "F2")) -> ("F1", "F2")
+    (("A", "F1"), (None, "F2")) -> ("A", "")
+    ((None, "F1"), (None, "F2")) -> ("F1", "F2")
+    """
+    if any(term for term, _ in pairs):
+        return tuple(term or "" for term, _ in pairs)
+    else:
+        return tuple(fallback for _, fallback in pairs)
+
+
 def tag_album(
     items,
     search_artist: str | None = None,
@@ -299,23 +318,24 @@ def tag_album(
                         Proposal(list(candidates.values()), rec),
                     )
 
-        # Search terms.
-        if not (search_artist and search_album):
-            # No explicit search terms -- use current metadata.
-            search_artist, search_album = cur_artist, cur_album
-        log.debug("Search terms: {} - {}", search_artist, search_album)
+        # Manually provided search terms or fallbacks.
+        _search_artist, _search_album = _parse_search_terms_with_fallbacks(
+            (search_artist, cur_artist),
+            (search_album, cur_album),
+        )
+        log.debug("Search terms: {} - {}", _search_artist, _search_album)
 
         # Is this album likely to be a "various artist" release?
         va_likely = (
             (not consensus["artist"])
-            or (search_artist.lower() in VA_ARTISTS)
+            or (_search_artist.lower() in VA_ARTISTS)
             or any(item.comp for item in items)
         )
         log.debug("Album might be VA: {}", va_likely)
 
         # Get the results from the data sources.
         for matched_candidate in metadata_plugins.candidates(
-            items, search_artist, search_album, va_likely
+            items, _search_artist, _search_album, va_likely
         ):
             _add_candidate(items, candidates, matched_candidate)
             if opt_candidate := candidates.get(matched_candidate.album_id):
@@ -329,7 +349,7 @@ def tag_album(
 
 
 def tag_item(
-    item,
+    item: Item,
     search_artist: str | None = None,
     search_title: str | None = None,
     search_ids: list[str] | None = None,
@@ -371,14 +391,18 @@ def tag_item(
         else:
             return Proposal([], Recommendation.none)
 
-    # Search terms.
-    search_artist = search_artist or item.artist
-    search_title = search_title or item.title
-    log.debug("Item search terms: {} - {}", search_artist, search_title)
+    # Manually provided search terms or fallbacks.
+    _search_artist, _search_title = _parse_search_terms_with_fallbacks(
+        (search_artist, item.artist),
+        (search_title, item.title),
+    )
+    log.debug("Item search terms: {} - {}", _search_artist, _search_title)
 
     # Get and evaluate candidate metadata.
     for track_info in metadata_plugins.item_candidates(
-        item, search_artist, search_title
+        item,
+        _search_artist,
+        _search_title,
     ):
         dist = track_distance(item, track_info, incl_artist=True)
         candidates[track_info.track_id] = hooks.TrackMatch(dist, track_info)
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -259,7 +259,7 @@ def tag_album(
     items,
     search_artist: str | None = None,
     search_album: str | None = None,
-    search_ids: list[str] = [],
+    search_ids: list[str] | None = None,
 ) -> tuple[str, str, Proposal]:
     """Return a tuple of the current artist name, the current album
     name, and a `Proposal` containing `AlbumMatch` candidates.
--- a/beets/autotag/match.py
+++ b/beets/autotag/match.py
@@ -259,7 +259,7 @@ def tag_album(
     items,
     search_artist: str = "",
     search_album: str = "",
-    search_ids: Sequence[str] | None = None,
+    search_ids: Sequence[str] = (),
 ) -> tuple[str, str, Proposal]:
     """Return a tuple of the current artist name, the current album
     name, and a `Proposal` containing `AlbumMatch` candidates.
@@ -288,16 +288,15 @@ def tag_album(
     candidates: dict[Any, AlbumMatch] = {}
 
     # Search by explicit ID.
-    if search_ids:
-        for search_id in search_ids:
-            log.debug("Searching for album ID: {}", search_id)
-            if info := metadata_plugins.album_for_id(search_id):
-                _add_candidate(items, candidates, info)
-                if opt_candidate := candidates.get(info.album_id):
-                    plugins.send("album_matched", match=opt_candidate)
+    for search_id in search_ids:
+        log.debug("Searching for album ID: {}", search_id)
+        if info := metadata_plugins.album_for_id(search_id):
+            _add_candidate(items, candidates, info)
+            if opt_candidate := candidates.get(info.album_id):
+                plugins.send("album_matched", match=opt_candidate)
 
     # Use existing metadata or text search.
-    else:
+    if not search_ids:
         # Try search based on current ID.
         if info := match_by_id(items):
             _add_candidate(items, candidates, info)
@@ -352,7 +351,7 @@ def tag_item(
     item: Item,
     search_artist: str = "",
     search_title: str = "",
-    search_ids: list[str] | None = None,
+    search_ids: Sequence[str] = (),
 ) -> Proposal:
     """Find metadata for a single track. Return a `Proposal` consisting
     of `TrackMatch` objects.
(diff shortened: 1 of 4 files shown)


beets/autotag/match.py
```diff
@@ -236,11 +236,30 @@ def _add_candidate(
     )
 
 
+def _parse_search_terms_with_fallbacks(
```
snejus, line 239: Seeing that you still want to use this helper method after reading my comments, I'm now concerned about my ability to understand the code.

I honestly have an issue trying to understand what this function does. I failed to do this by reading the code, so I had to manually run search in order to see what it does.

snejus: I've got a simple mp3 track:

```sh
$ tag track.mp3
IDv2 tag info for track.mp3
TALB=album
TIT2=title
```

### Previously

| import    | artist search | album/title search | resulting search terms |
| --------- | ------------- | ------------------ | ---------------------- |
| album     | `None`        | `None`             | ` - album`             |
| album     | `"Artist"`    | `""`               | ` - album`             |
| album     | `""`          | `"Album"`          | ` - album`             |
| album     | `"Artist"`    | `"Album"`          | `Artist - Album`       |
| singleton | `None`        | `None`             | ` - title`             |
| singleton | `"Artist"`    | `""`               | `Artist - title`       |
| singleton | `""`          | `"Title"`          | ` - Title`             |
| singleton | `"Artist"`    | `"Title"`          | `Artist - Title`       |

I'm actually surprised to see that album search requires **both** terms. Apparently, I misunderstood even the simple `if not (search_artist and search_album)` condition 😁. 

### Now

| import    | artist search | album/title search | resulting search terms |
| --------- | ------------- | ------------------ | ---------------------- |
| album     | `None`        | `None`             | ` - album`             |
| album     | `"Artist"`    | `""`               | `Artist - `            |
| album     | `""`          | `"Album"`          | ` - Album`             |
| album     | `"Artist"`    | `"Album"`          | `Artist - Album`       |
| singleton | `None`        | `None`             | ` - title`             |
| singleton | `"Artist"`    | `""`               | `Artist - `            |
| singleton | `""`          | `"Title"`          | ` - Title`             |
| singleton | `"Artist"`    | `"Title"`          | `Artist - Title`       |

I take back what I said regarding 'we're keeping the previous functionality'. What you've got here aligns with my expectation: we only need one of the search terms to be present for the search to override item data.

beets/autotag/match.py
```diff
@@ -236,11 +236,30 @@ def _add_candidate(
...
+    if any(term for term, _ in pairs):
+        return tuple(term or "" for term, _ in pairs)
+    else:
+        return tuple(fallback for _, fallback in pairs)
+
+
 def tag_album(
     items,
     search_artist: str | None = None,
     search_album: str | None = None,
```
snejus: I suggest we handle _only_ strings here:

```py
    search_artist: str,
    search_album: str,
```

And update the `tag_item` and `tag_album` calls in `beets.importer.tasks` to explicitly provide empty strings.

beets/autotag/match.py
```diff
@@ -294,23 +313,24 @@ def tag_album(
...
-        # Search terms.
-        if not (search_artist and search_album):
-            # No explicit search terms -- use current metadata.
-            search_artist, search_album = cur_artist, cur_album
-        log.debug("Search terms: {} - {}", search_artist, search_album)
+        # Manually provided search terms or fallbacks.
+        _search_artist, _search_album = _parse_search_terms_with_fallbacks(
+            (search_artist, cur_artist),
+            (search_album, cur_album),
+        )
```
snejus, line 324: Based on my comment above, we don't need to handle `None`s any more, so this can be simplified to

```py
if not search_artist and not search_album:
    search_artist, search_album = cur_artist, cur_album
```

beets/autotag/match.py
```diff
@@ -236,11 +236,30 @@ def _add_candidate(
...
+    else:
+        return tuple(fallback for _, fallback in pairs)
+
+
 def tag_album(
     items,
     search_artist: str | None = None,
     search_album: str | None = None,
-    search_ids: list[str] = [],
+    search_ids: list[str] | None = None,
```
snejus: This is a nice use case for a `Sequence` with a default of empty `tuple`:


```suggestion
    search_ids: Sequence[str] = (),
```

This way, you can simplify the following:

```diff
-    if search_ids:
-        for search_id in search_ids:
+    for search_id in search_ids:
```

beets/metadata_plugins.py
```diff
@@ -35,17 +35,35 @@ def find_metadata_source_plugins() -> list[MetadataSourcePlugin]:
...
+) -> Iterable[AlbumInfo]:
     """Return matching album candidates from all metadata source plugins."""
     for plugin in find_metadata_source_plugins():
-        yield from plugin.candidates(*args, **kwargs)
+        yield from plugin.candidates(
+            items=items,
+            artist=artist,
+            album=album,
+            va_likely=va_likely,
+        )
```
snejus, line 45: I'm happy with making arguments explicit, but let's delegate them unchanged (using the same `plugin.candidates` call):

```suggestion
def candidates(
    items: Sequence[Item], artist: str, album: str, va_likely: bool
) -> Iterable[AlbumInfo]:
    """Return matching album candidates from all metadata source plugins."""
    for plugin in find_metadata_source_plugins():
        yield from plugin.candidates(items, artist, album, va_likely)
```

beets/metadata_plugins.py
```diff
@@ -157,15 +175,22 @@ def candidates(
...
+        title: str,
     ) -> Iterable[TrackInfo]:
         """Return :py:class:`TrackInfo` candidates that match the given track.
 
         Used in the autotag functionality to search for tracks.
 
         :param item: Track item
-        :param artist: Track artist
-        :param title: Track title
+        :param artist: Track artist, either a search manually provided or
```
snejus: Let's undo this - the plugin is not supposed to know where this data comes from. This is going to get out of sync given how plugins may extend it (and it probably is given the `fromfilename` plugin).

beets/autotag/match.py
```diff
@@ -364,14 +384,18 @@ def tag_item(
         else:
             return Proposal([], Recommendation.none)
 
-    # Search terms.
-    search_artist = search_artist or item.artist
-    search_title = search_title or item.title
-    log.debug("Item search terms: {} - {}", search_artist, search_title)
+    # Manually provided search terms or fallbacks.
+    _search_artist, _search_title = _parse_search_terms_with_fallbacks(
+        (search_artist, item.artist),
```
snejus, line 395: Currently, `item.artist` and `item.title` types are not present. Let's add them:

```py
class LibModel(dbcore.Model["Library"]):
    ...
    path: bytes
    artist: str

...

class Item(LibModel):
    title: str
```

snejus (changes requested):

beets/autotag/match.py
semohr: Yeah I also had issue with understanding the logic especially since it was different in singletons and albums. Have you seen the test I added?

snejus: I did, see the rest of the comments

beets/autotag/match.py
semohr: This would than also allow for non list entries. I remember we opted to not use Sequence[str] in some other places because of this reason.
```python
# both valid:
search_ids: Sequence[str] = "fooo"
search_ids: Sequence[str] = ["fooo", "baar"]
```

beets/metadata_plugins.py
semohr: Why can't we use trailing commas? https://docs.astral.sh/ruff/rules/missing-trailing-comma/

snejus: I prefer to not use it if not needed, for verbosity reasons. Scrolling the code does take some time at the end of the day 🤷🏼

beets/metadata_plugins.py
semohr: The issue I see here, is that there are two places to get this information from. E.g. either a plugin gets it from `item.artist` or from `artist`. For me, until looking into the source, it was not clear what are the differences. Giving some more information here makes this a bit more clear in my opinion.  We can change the wording but we should definitely mention this somehow.

A bit in the line of how are `artist` and `title` more special in this case as every other field defined in `item`?

beets/metadata_plugins.py
semohr: I would prefer a cleaner git blame all day tbh :shrug: 

Changed it to the one line rep tho.

beets/metadata_plugins.py
snejus: The main question is: do these two fields *always* come *from the item*?

> Track artist, either a search manually provided or preprocessed from the item

semohr: > The main question is: do these two fields always come from the item?

No they are either a manual provided search or the items values. That's why I want to mentioned it the differences somehow. Just looking at the wording, there is no difference between `item.artist` and using the `artist` argument. While I do not expect every plugin developer to read the full source-code and understand the nuances here,  I would argue everyone who develops a metadata plugin reads the docstring for this function.

Think abit about a new developer writing such a plugin, why would I use `artist` if there is also `item.artists`which might have more information for me to query an external service with. A single comment mentioning that artist might be a manual query would give me a hint to not use `item.artists` here but rather rely on artist.

> the plugin is not supposed to know where this data comes from

I agree that with a perfect api it should not matter but we have a bit of ambiguity here. It might be better to establish a unique representation but as I have mentioned somewhere else, changing the api here is close to impossible as external plugins rely on it heavily.

beets/autotag/match.py
semohr: I did not add this here as it would required a bit of refactoring or type ignores in other files which does not seem too related here. 

Let's look at that in another PR.

beets/metadata_plugins.py
semohr: I replaced the comment in the parameters directly with a note under the parameter list:

```
Note that `artist` and `album` may contain additional user-supplied search terms
intended to refine the query. When relevant, prefer these values over
metadata extracted from item directly.
```

Hope this allows us to converge here :upside_down_face:

semohr:
@snejus How do we want to continue here? 

We could remove the `_parse_search_terms_with_fallbacks` function and revert to an inline approach, but I’m a bit concerned that an inline solution might drift out of sync again over time. I’d also prefer to keep the new tests, since they effectively serve as a contract for the search-parsing behavior.

That said, in the grand scheme I don't care too much as long as the tests remain in place in some form.

snejus:
I think I'm missing something... Is this condition so complex it needs testing? https://github.com/beetbox/beets/pull/6117#discussion_r2454370978

semohr:
I mean it kinda is, otherwise we would not have had this issue here (see [also](https://github.com/beetbox/beets/pull/6117/files#r2454351503)). It is not really about the complexity but ensuring the implementation is well defined and stays as expected, we would use the tests here to ensure the functional requirements.

snejus:
Remember the reason I did the investigation behind that comment: after your adjustment, I could not any more understand what this code does. The test does not help. 

On the positive side, I'm happy that you drew attention to it, because it does not work the way I'd expect it to, and thus we need to adjust it:

```diff
-        if not (search_artist and search_name):
+        if not search_artist and not search_album:
            search_artist, search_name = cur_artist, cur_album

...

-    search_artist = search_artist or item.artist
-    search_title = search_title or item.title
+    if not search_artist and not search_title:
+        search_artist, search_title = item.artist, item.title

```

I think we are on the same page regarding this simple conditional being unambigous and not needing testing?

---

I would be keen to keep it simple and stupid and finally get it merged.

semohr:
The simple and stupid approach has introduced this issue and has allowed the desync to be introduced in the first place. I would like this a bit more controlled, either have the implementations coupled with a private function and/or have this properly tested. As you are against a private function to couple the item and album logic (which tbh I don't understand, as we pretty much do the same for the `_sort_candidate` function), we should at-least enforce the behavior by testing it.

PR #6119, opened 2025-10-21
Refactor `ui/commands.py` monolith into modular structure

semohr:
## Description

This one’s a big one :fishing_pole_and_fish: Proceed with care and a bit of time ;)

The `ui/commands.py` file had grown into an unwieldy monolith (2000+ lines) over time, so this PR breaks it apart into a modular structure i.e. **one file per command**, plus some cleanup and reorganization along the way.

---

###  What changed

* **Commands modularized:**
  Every command (`help`, `list`, `move`, `update`, `remove`, etc.) now lives in its own file under `ui/commands/`.
* **Support code reorganized:**
  * Utility functions moved into a separate helper module.
  * `commands.py` converted into `commands/__init__.py` for better import handling.
  * The `import` command (and related helpers) moved into its own folder:
    * `importer/session.py` for import session logic
    * `importer/display.py` for display-related functions
* **Tests cleaned up:**
  * Each command’s tests now live in their own file.
  * All UI-related tests were moved into a dedicated folder for clarity.

---

### Review notes

Try to look at **one commit at a time** - the changes are organized logically across commits to make the refactoring easier to follow. The source code stayed mostly unchanged.

### Open questions

How should we approach deprecations? While we could re-export all functions and classes that aren’t prefixed with an underscore, most of the module’s functions feel effectively private, so we might want to consider limiting the public API and phasing out anything that shouldn’t be used externally.

How do we want to add this to the `.git-blame-ignore-revs` file? I feel like putting all commits into it might be a bit spammy. There has to be a better way here.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -75,5 +75,7 @@ d93ddf8dd43e4f9ed072a03829e287c78d2570a2
 28aee0fde463f1e18dfdba1994e2bdb80833722f
 # Refactor `ui/commands.py` into multiple modules
 59c93e70139f70e9fd1c6f3c1bceb005945bec33
-a59e41a88365e414db3282658d2aa456e0b3468a
+# Moved ui.commands._utils into ui.commands.utils
 25ae330044abf04045e3f378f72bbaed739fb30d
+# Refactor test_ui_command.py into multiple modules 
+a59e41a88365e414db3282658d2aa456e0b3468a


beets/ui/commands/import_/__init__.py
```diff
@@ -0,0 +1,281 @@
...
+                continue
+
+            byte_paths.append(path)
+
+        # If all paths were read from a logfile, and none of them exist, throw
+        # an error
+        if not paths:
+            raise ui.UserError("none of the paths are importable")
+
+    import_files(lib, byte_paths, query)
```
amogus07, line 131: query is passed to import_files as a `list[str]`, which passes it to the constructor of `TerminalImportSession`, which expects a `Query` object. How does that work?

semohr: I think the string is converted to a query at some point. Not sure where this happens. This is not a change introduced in this PR tho. Was just migrated to this file from here: https://github.com/beetbox/beets/blob/043581e0c9af6e2fbdd103d8707d0b8ea26b51f7/beets/ui/commands.py#L1328-L1378

Also notice this [commit](https://github.com/beetbox/beets/pull/6119/commits/df2be2c866cb76dde9c815f41b2c2cc82161d5e5), where this part of the code was moved.

amogus07: I looked into it more, I can't find any point in the code where it gets converted. It originates from `optparse.parse_args` (used in `Subcommand.parse_args`), which returns it as a `list[str]`. This value ends up in `_raw_main`and is direcly passed to subcommand.func. It never gets converted from what I can tell. So is TerminalInputSession able to accept use `list[str]` as it's query parameter?

amogus07: yay, found it! it gets converted in `Library._get`. That means the type of ImportSession.query should be `str | list[str] | tuple[str] | Query`, since all of these are currently valid.

beets/ui/commands/import_/session.py
```diff
@@ -0,0 +1,557 @@
...
+            raise plugins.PluginConflictError(
+                "Only one handler for `import_task_before_choice` may return "
+                "an action."
+            )
+
+        # Take immediate action if appropriate.
+        action = _summary_judgment(task.rec)
+        if action == importer.Action.APPLY:
+            match = task.candidates[0]
+            show_change(task.cur_artist, task.cur_album, match)
```
amogus07, line 58: match is of type `AlbumMatch | TrackMatch` while show_change only accepts `AlbumMatch`. Something isn't adding up here

beets/ui/commands/import_/__init__.py
semohr: We can address this after the restructure :+1: Fixing and finding these type issues is one of the reasons I have started this.

beets/ui/commands/import_/session.py
amogus07: The only solution I can think of is to use `show_item_change` if `Match` is a `TrackMatch`. But then again, according to the docstring, choose_match isn't supposed to return a TrackInfo object, even though I can't see anything that would prevent it

snejus: This entire method is a mess to type, note it may return either of Match subclasses OR `importer.Action`. I've been working on splitting `ImporterTask` into two - one for item, one for album. This is out of scope for this restructure though.

snejus:
Well done, this must have required a significant time investment to undertake!! <3

Had a quick glance and I see a circular import issue:

```fish
$ grepr 'from.*ui$' beets
beets/ui/commands/import_/__init__.py:from beets import config, logging, plugins, ui
beets/ui/commands/import_/display.py:from beets import autotag, config, logging, ui
beets/ui/commands/import_/session.py:from beets import autotag, config, importer, logging, plugins, ui
beets/ui/commands/_utils.py:from beets import ui
beets/ui/commands/completion.py:from beets import library, logging, plugins, ui
beets/ui/commands/fields.py:from beets import library, ui
beets/ui/commands/help.py:from beets import ui
beets/ui/commands/list.py:from beets import ui
beets/ui/commands/modify.py:from beets import library, ui
beets/ui/commands/remove.py:from beets import ui
beets/ui/commands/stats.py:from beets import logging, ui
beets/ui/commands/update.py:from beets import library, logging, ui
beets/ui/commands/version.py:from beets import plugins, ui
beets/ui/commands/write.py:from beets import library, logging, ui
```

beets/ui/commands/import_/session.py
amogus07: so what to for now? something like this?
```suggestion
            match = task.candidates[0]
            if isinstance(match, AlbumMatch):
                show_change(task.cur_artist, task.cur_album, match)
            else:
                show_item_change(task.cur_artist, task.cur_album, match)
```

snejus: No change to the logic within this PR

amogus07:
@snejus I've already fixed the import cycles, I'm working on merging this pr with #6125.

semohr:
> Well done, this must have required a significant time investment to undertake!! <3
> 
> Had a quick glance and I see a circular import issue:
> 
> ```fish
> $ grepr 'from.*ui$' beets
> beets/ui/commands/import_/__init__.py:from beets import config, logging, plugins, ui
> beets/ui/commands/import_/display.py:from beets import autotag, config, logging, ui
> beets/ui/commands/import_/session.py:from beets import autotag, config, importer, logging, plugins, ui
> beets/ui/commands/_utils.py:from beets import ui
> beets/ui/commands/completion.py:from beets import library, logging, plugins, ui
> beets/ui/commands/fields.py:from beets import library, ui
> beets/ui/commands/help.py:from beets import ui
> beets/ui/commands/list.py:from beets import ui
> beets/ui/commands/modify.py:from beets import library, ui
> beets/ui/commands/remove.py:from beets import ui
> beets/ui/commands/stats.py:from beets import logging, ui
> beets/ui/commands/update.py:from beets import library, logging, ui
> beets/ui/commands/version.py:from beets import plugins, ui
> beets/ui/commands/write.py:from beets import library, logging, ui
> ```

How do you recon we fix these? Intriguing as this doesn't appear to be issue for the test runs :thinking:

amogus07:
@semohr see #6129. I split up most of `ui/__init__.py` into `ui._common` `ui.core` and `ui.colors`. Importing from the respective modules within ui fixes the import cycles, while other modules can still import from `ui` directly.

semohr:
This was not necessarily what I meant. I want to keep the efforts here focused on the commands.py file. Also changing things around in the ui base layer seems a bit out of scope here.

The major surprise for me is that this does not introduce a runtime error and I would like to understand why.

amogus07:
> The major surprise for me is that this does not introduce a runtime error and I would like to understand why.

It's because the modules in commands don't actually import anything from ui, they just get a reference to it. By the time the functions that use attributes of the module are called, it happens to have fully initialized, which obscures the problem

amogus07:
ui.commands is only imported in _stetup and _raw_main. Those happen to be below everything the commands depend on

semohr:
Than this does not seem like an actual issue here. As this PR only restructures the module and the imports where handled identically beforehand the problem was existing [beforehand](https://github.com/beetbox/beets/blob/201677ae6297f176f77aae2789e30e496c03c1e5/beets/ui/commands.py#L30) too.

@snejus Let's omit this here for now and keep this focused on the restructure. We can fix this in a followup more easily anyways.

amogus07:
import cycles don't cause errors directly but can lead attribute access errors

amogus07:
@semohr so I can expect this to get merged and build on top right? Then after this gets merged, I would be able to submit my pr directly to master as a followup

semohr:
Yeah, I guess we will merge this in one form or another. Feel free to base your upcomming PR on this branch.

snejus:
They don't cause any runtime errors because you only import the `ui` module. Try importing

```py
from beets.ui import print_
```

in one of the command modules and you will see what I mean. Fixing these is most likely out of scope of this PR since it does require a fair bit of restructuring following the idea that @amogus07 described [here](https://github.com/beetbox/beets/pull/6119#issuecomment-3448293266).

beetsplug/edit.py
```diff
@@ -25,7 +25,8 @@
 from beets import plugins, ui, util
 from beets.dbcore import types
 from beets.importer import Action
-from beets.ui.commands import PromptChoice, _do_query
+from beets.ui.commands._utils import do_query
```
snejus: Seems like the module should be called `utils` since it's used outside of the `commands` submodule.

test/ui/commands/test_import.py
```diff
@@ -0,0 +1,263 @@
...
+        # ToDo: fix color_len to handle mid-text color escapes, and thus
+        # split colored texts over newlines (potentially with dashes?)
+        split_txt = ["\x1b[31mtest\x1b[39;49;00mt", "est", "test", "test"]
+        txt = ui.split_into_lines(colored_text, [5, 5, 5])
+        assert txt == split_txt
+
+    def test_album_data_change_wrap_newline(self):
+        # Patch ui.term_width to force wrapping
+        with patch(
+            "beets.ui.commands.import_.display.ui.term_width", return_value=30
```
snejus:
```suggestion
            "beets.ui.term_width", return_value=30
```

beets/ui/commands/_utils.py
```diff
@@ -0,0 +1,67 @@
...
+        try:
+            yield from paths_from_logfile(syspath(normpath(logfile)))
+        except ValueError as err:
+            raise ui.UserError(
+                f"malformed logfile {displayable_path(logfile)}: {err}"
+            ) from err
+        except OSError as err:
+            raise ui.UserError(
+                f"unreadable logfile {displayable_path(logfile)}: {err}"
+            ) from err
```
snejus: These both should be defined in `beets/ui/commands/import_/__init__.py`.

beets/ui/commands/import_/display.py
```diff
@@ -0,0 +1,573 @@
...
+
+from beets import autotag, config, logging, ui
+from beets.autotag import hooks
+from beets.util import displayable_path
+from beets.util.units import human_seconds_short
+
+VARIOUS_ARTISTS = "Various Artists"
+
+# Global logger.
+log = logging.getLogger("beets")
```
snejus: Doesn't seem to be used in this module.

beets/ui/commands/move.py
```diff
@@ -0,0 +1,154 @@
...
+        else:
+            ui.show_path_changes(
+                [(obj.path, obj.destination(basedir=dest)) for obj in objs]
+            )
+    else:
+        if confirm:
+            objs = ui.input_select_objects(
+                f"Really {act}",
+                objs,
+                lambda o: ui.show_path_changes(
```
snejus: `show_path_changes` is not used anywhere else in the codebase except this module, so you can relocate it here too.

beets/ui/commands/move.py
```diff
@@ -0,0 +1,154 @@
+"""The 'move' command: Move/copy files to the library or a new base directory."""
+
+import os
+
+from beets import logging, ui, util
```
snejus: This module is now inconsistent with others regarding imports from `beets.util`. The rest import specifically what they need.

beets/ui/commands/config.py
```diff
@@ -0,0 +1,89 @@
+"""The 'config' command: show and edit user configuration."""
+
+import os
+
+from beets import config, ui, util
```
snejus: Also needs specific imports from `util`.

test/ui/commands/test_update.py
```diff
@@ -0,0 +1,205 @@
...
+        self.i.move(operation=MoveOperation.COPY)
+        self.i2.move(operation=MoveOperation.COPY)
+        self.album = self.lib.add_album([self.i, self.i2])
+
+        # Album art.
+        artfile = os.path.join(self.temp_dir, b"testart.jpg")
+        _common.touch(artfile)
+        self.album.set_art(artfile)
+        self.album.store()
+        util.remove(artfile)
```
snejus: Just import `remove` from `utils` specifically (I'm aware it was like this originally but best to update regardless).

beets/ui/commands/completion.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def completion_script(commands):
+    """Yield the full completion shell script as strings.
+
+    ``commands`` is alist of ``ui.Subcommand`` instances to generate
+    completion data for.
+    """
+    base_script = os.path.join(
+        os.path.dirname(__file__), "../completion_base.sh"
```
snejus: You probably want to move `completion_base.sh` under `beets/ui/commands/` too.

beets/ui/commands/import_/__init__.py
```diff
@@ -0,0 +1,281 @@
...
+    dest="from_logfiles",
+    action="append",
+    metavar="PATH",
+    help="read skipped paths from an existing logfile",
+)
+import_cmd.parser.add_option(
+    "--set",
+    dest="set_fields",
+    action="callback",
+    callback=ui._store_dict,
```
snejus: `_store_dict` is also only used in this module, so ideally it should be defined here.

snejus (changes requested):
A couple of small comments. Otherwise, it's largely ready.

snejus:
> How do we want to add this to the `.git-blame-ignore-revs` file? I feel like putting all commits into it might be a bit spammy. There has to be a better way here.

I imagine we'd probably squash all commits for `test/` directory into one, and all commits for `beets/` directory into another one.

semohr:
> I imagine we'd probably squash all commits for test/ directory into one, and all commits for beets/ directory into another one.

Should be done now. The full PR is now included in 4 commits.

.git-blame-ignore-revs
```diff
@@ -73,3 +73,7 @@ d93ddf8dd43e4f9ed072a03829e287c78d2570a2
 33f1a5d0bef8ca08be79ee7a0d02a018d502680d
 # Moved art.py utility module from beets into beetsplug
 28aee0fde463f1e18dfdba1994e2bdb80833722f
+# Refactor `ui/commands.py` into multiple modules
+59c93e70139f70e9fd1c6f3c1bceb005945bec33
```
snejus, line 77: Are you happy to include commit summaries for each of these commits, same as the rest of the entries above?

semohr:
> Very well done, GitHub is having a real struggle to render this PR properly due to size 😆 Just add commit summaries for each of the commits in the blame ignore revs following the rest of the file

Performance wise, Github is really taking a dump for me too. Noticing issues since they have update their UI some months ago.

PR #6122, opened 2025-10-24
add lots of type hints

amogus07:
## Description

I added type hints ~and refactored the modules to fix import cycles~

## To Do

- [ ] Documentation
- [ ] Changelog
- [ ] Tests


semohr:
Hey, thanks a lot for the contribution, really appreciate the effort you’ve put into this!

This PR is quite large, which makes it a bit difficult to review thoroughly. Would you mind breaking it down into smaller, more focused commits or separate PRs? For instance, you could move file relocations or refactors into their own commits to make the changes easier to follow. You can take a look at the PR linked below for an example of how to structure these kinds of refactors.

It also looks like this PR touches on multiple areas of the codebase. To keep things maintainable and easier to review, it’s best if each PR focuses on a single concern or goal, rather than combining several changes at once.

As a side note, have you had a chance to look at #6119? It seems to address some of the same issues you’re working on here, it might be worth aligning with that effort to avoid overlap.

Thanks again for your work on this! We really appreciate your recent efforts!

beets/autotag/distance.py
```diff
@@ -384,11 +384,19 @@ def track_distance(
     cached because this function is called many times during the matching
     process and their access comes with a performance overhead.
     """
-    dist = Distance()
+    dist: Distance = Distance()
```
snejus, line 387: This is redundant

beets/autotag/distance.py
```diff
@@ -345,7 +345,7 @@ def add_string(self, key: str, str1: str | None, str2: str | None):
         dist = string_dist(str1, str2)
         self.add(key, dist)
 
-    def add_data_source(self, before: str | None, after: str | None) -> None:
+    def add_data_source(self, before: object, after: str | None) -> None:
```
snejus, line 348: Why?

beets/autotag/distance.py
```diff
@@ -384,11 +384,19 @@ def track_distance(
...
     process and their access comes with a performance overhead.
     """
-    dist = Distance()
+    dist: Distance = Distance()
 
     # Length.
+    info_length: float | None
     if info_length := track_info.length:
-        diff = abs(item.length - info_length) - get_track_length_grace()
+        diff: float = (
```
snejus, line 392: Again, redundant

beets/dbcore/db.py
```diff
@@ -310,7 +302,7 @@ def _types(cls) -> dict[str, types.Type]:
     """
 
     @cached_classproperty
-    def _queries(cls) -> dict[str, FieldQueryType]:
+    def _queries(cls) -> dict[str, type[Query]]:
```
snejus, line 322: All of these changes from `FieldQuery` -> `Query` are incorrect. Why?

beets/dbcore/db.py
```diff
@@ -744,7 +736,7 @@ def __getstate__(self):
 AnyModel = TypeVar("AnyModel", bound=Model)
 
 
-class Results(Generic[AnyModel]):
+class Results(Generic[AnyModel, D]):
```
snejus, line 766: Why?

beets/dbcore/queryparse.py
```diff
@@ -18,37 +18,44 @@
...
 if TYPE_CHECKING:
+    import sys
     from collections.abc import Collection, Sequence
 
-    from ..library import LibModel
-    from .query import FieldQueryType, Sort
+    from .query import Sort
+
+    if not sys.version_info < (3, 10):
+        from typing import TypeAlias  # pyright: ignore[reportUnreachable]
```
snejus, line 34: We use `mypy`, please remove this and setup your editor to use `mypy` in your environment for this project.

snejus: And see #5879 #5869

beets/dbcore/queryparse.py
```diff
@@ -18,37 +18,44 @@
...
 
 PARSE_QUERY_PART_REGEX = re.compile(
     # Non-capturing optional segment for the keyword.
-    r"(-|\^)?"  # Negation prefixes.
-    r"(?:"
-    r"(\S+?)"  # The field key.
-    r"(?<!\\):"  # Unescaped :
-    r")?"
-    r"(.*)",  # The term itself.
+    r"(-|\^)?"  # Negation prefixes. #  noqa: ISC003
```
snejus, line 43: Why? Please undo

beets/dbcore/queryparse.py
```diff
@@ -94,15 +101,17 @@ def parse_query_part(
     """
     # Apply the regular expression and extract the components.
     part = part.strip()
-    match = PARSE_QUERY_PART_REGEX.match(part)
+    match: re.Match[str] | None = PARSE_QUERY_PART_REGEX.match(part)
```
snejus, line 104: This is redundant

beets/dbcore/queryparse.py
```diff
@@ -137,26 +146,30 @@ def construct_query_part(
...
+        out_query = model_cls.field_query(
+            key.lower(), pattern, cast(type[FieldQuery], query_class)
+        )
 
-    if key is None:
+    else:
         # If there's no key (field name) specified, this is a "match anything"
         # query.
         out_query = model_cls.any_field_query(pattern, query_class)
-    else:
```
snejus: Why switching the order of the conditional?

Why did you replace `FieldQuery` in the function input and then introduce `cast` here?

beets/dbcore/queryparse.py
```diff
@@ -137,26 +146,30 @@ def construct_query_part(
...
     # `Query` classes.
-    query_classes: dict[str, FieldQueryType] = {}
+    query_classes: dict[str, type[Query]] = {}
     for k, t in itertools.chain(
         model_cls._fields.items(), model_cls._types.items()
     ):
         query_classes[k] = t.query
-    query_classes.update(model_cls._queries)  # Non-field queries.
+    query_classes.update(
+        model_cls._queries.items()  # Non-field queries.
```
snejus, line 155: Why?

beets/dbcore/queryparse.py
```diff
@@ -18,37 +18,44 @@
 
 import itertools
 import re
-from typing import TYPE_CHECKING
```
snejus: I don't see any improvements in this module except for redundant types and adjustments in function input types which then force you to use `cast` and `type: ignore`.

snejus (changes requested):
Thanks for your contribution!

I reviewed a couple of files but stopped at `queryparse.py` since I found myself confused about the motivation for these changes?

I'm struggling to find any improvements here, while I see a fair bit off issues. Please see the comments.

snejus:
Apologies for my language - I realise I sound dismissive and terse there - I am tired and about to go to sleep, so this is my fault, not your changes <3

I take that most of these type adjustments stem from you using `pyright` to type check this project. In an ideal world, `pyright` and `mypy` would show the same issues, but they don't in reality. This project uses `mypy` as our type checker, so changes should address `mypy`'s requirements rather than `pyright`'s. Some of these type hints may be fixing issues that `pyright` sees but `mypy` doesn't, because it manages to resolve the types.

Would you mind running `mypy` on your changes to see which type issues actually need addressing for our setup? That would help focus the PR on changes that are truly necessary. Please use `resolve_type` to see whether a type is missing for a variable.

Thanks again for your effort on this - I really do appreciate it!

amogus07:
@snejus Thank you for taking the time to review this! Yeah, I probably should've left it as a draft for now, I pretty much expected some changes would have to be reverted. I did realize that this project uses mypy, but either it doesn't play well with my editor (Zed) or just it's LSP isn't as good as Basedpyright, which is what I'm using. As for the "redundant" type annotations, that's just my personal preference. Sure, you can live without them, but using I just don't like the concept of inferred types at all. They can change without me noticing if I accidentally reuse a variable for an unintended purpose. My goal is to get beets closer to a state where it complies with mypy's strict mode. Although I might suggest migrating to a different type checker like basedpyright at some point. On the other hand, we could wait until pyrefly becomes stable, which is significantly faster then both mypy and basedpyright. I will go over your feedback in detail when I have time and use it as guidance for future contributions.

beets/autotag/distance.py
amogus07: I'd argue it's not. The annotation explicitly defines that, within the scope of the function, dist should always be of type Distance. Without it, dist could accidentally be changed to a value of any type, e. g. an int without a warning. The stricter the annotations are, the faster you can get to the root cause of e. g. a regression. Consider this oversimplified example:

```Python
x = 0
x = "foo" # this line is the source of the problem, but it isn't highlighted by the type checker
do_something(x) # this function only accepts int, so x is highlighted
```

```Python
x: int = 0
x = "foo" # now "foo" is highlighted correctly
do_something(x)
```
Obviously this example is pretty silly but rather than deciding for each variable whether it should be annotated or not, I'd just annotate everything I come across.

snejus: That's incorrect:

```py
    dist = Distance()
    dist = 5  # mypy: assignment: Incompatible types in assignment (expression has type "int", variable has type "Distance")
    dist = "foo"  # mypy: assignment: Incompatible types in assignment (expression has type "str", variable has type "Distance")
```

I repeat, you must use `mypy` and not `basedpyright` to type-check this codebase.

snejus:
I am closing this as most of the type annotations do not add value to the codebase. 

> On the other hand, we could wait until pyrefly becomes stable, which is significantly faster then both mypy and basedpyright. I will go over your feedback in detail when I have time and use it as guidance for future contributions.

Have you heard of `zuban`? It beats both `mypy` and `basedpyright` in speed, and uses much less memory than `ty`. It's `jedi` and `mypy` (and parts of `basedpyright`) morphed into one monster of a tool.

Issue #6121, opened 2025-10-23
Certain albums are not searched on musicbrainz with any plugin enabled?


0xE0-rng:
Hi! Thanks for the cool tool! I was a user a few years ago and just got into it again. Unfortunatly with some small issue :/ 
### Problem
I am sure it has to do with my setup, but I cannot find the issue. With a minimal config, some Albums are just not searched on musicbrainz. One example is 
https://musicbrainz.org/release/34547462-bb50-435d-a8ba-eae72430875b
or 
https://musicbrainz.org/release/f72d8759-1d81-4902-9cb1-403937271c51
This happens with any plugins enabled, however if I disable the plugins it works as expected. 

```sh
root@b6692f89a99d:/downloads# beet -vv import RÜFÜS\ -\ Inhale\ _\ Exhale\ Remixed\,\ Vol.\ 1/
user configuration: /config/config.yaml
data directory: /config
plugin paths: []
Loading plugins: convert
Sending event: pluginload
library database: /config/musiclibrary.blb
library directory: /data/newMusic
Sending event: library_opened
Sending event: import_begin
Sending event: import_task_created
Sending event: import_task_start
Looking up: /downloads/RÜFÜS - Inhale _ Exhale Remixed, Vol. 1
Tagging RÜFÜS - Inhale / Exhale Remixed, Vol. 1
No album ID found.
Search terms: RÜFÜS - Inhale / Exhale Remixed, Vol. 1
Album might be VA: False
Evaluating 0 candidates.

/downloads/RÜFÜS - Inhale _ Exhale Remixed, Vol. 1 (15 items)
Sending event: import_task_before_choice
Sending event: before_choose_candidate
No matching release found for 15 tracks.
For help, see: https://beets.readthedocs.org/en/latest/faq.html#nomatch
➜ [S]kip, Use as-is, as Tracks, Group albums, Enter search, enter Id, aBort? B
Sending event: import
Sending event: cli_exit
```

As soon as I disabled the one plugin from the config this happens:
```
root@b6692f89a99d:/downloads# beet -vv import RÜFÜS\ -\ Inhale\ _\ Exhale\ Remixed\,\ Vol.\ 1/
user configuration: /config/config.yaml
data directory: /config
plugin paths: []
Loading plugins: musicbrainz
Sending event: pluginload
library database: /config/musiclibrary.blb
library directory: /data/newMusic
Sending event: library_opened
Sending event: import_begin
Sending event: import_task_created
Sending event: import_task_start
[...]

PR #6127, opened 2025-10-25
Add deprecation warning for `musicbrainz.enabled` but use it to load the plugin, centralise deprecations handling

snejus:
Fixes: #6121

This PR introduces a centralized deprecation system and adjusts `musicbrainz` plugin loading to properly handle the deprecated `musicbrainz.enabled` configuration option.

#### MusicBrainz

- Added deprecation warnings for the `musicbrainz.enabled` configuration option:
  - When set to `true`, warns users to explicitly add `musicbrainz` to their `plugins` configuration and adds it if not already present
  - When set to `false`, warns users and adds the plugin to `disabled_plugins` (list
    received by the `--disable-plugins` flag)

#### Deprecations

- Created new `beets/util/deprecation.py` module with standardized deprecation helpers:
  - `deprecate_for_user()` - logs warnings visible to end users
  - `deprecate_for_maintainers()` - emits `DeprecationWarning` for developers
  - `deprecate_imports()` - handles deprecated module imports with automatic version calculation
  - `_format_message()` - generates consistent deprecation messages that auto-calculate next major version

- Migrated all deprecation handling to use the new centralized functions:
  - Replaced inline `warnings.warn()` calls throughout codebase
  - Updated `deprecate_imports()` signature to remove explicit `version` parameter
  - Converted user-facing deprecation warnings in plugins to use logger-based `deprecate_for_user()`

### DIFF
--- a/beets/ui/commands/__init__.py
+++ b/beets/ui/commands/__init__.py
@@ -39,7 +39,7 @@ def __getattr__(name: str):
         __name__,
         {
             "TerminalImportSession": "beets.ui.commands.import_.session",
-            "PromptChoice": "beets.ui.commands.import_.session",
+            "PromptChoice": "beets.util",
         },
         name,
     )
--- a/beets/ui/commands/import_/session.py
+++ b/beets/ui/commands/import_/session.py
@@ -1,10 +1,9 @@
 from collections import Counter
 from itertools import chain
-from typing import Any, NamedTuple
 
 from beets import autotag, config, importer, logging, plugins, ui
 from beets.autotag import Recommendation
-from beets.util import displayable_path
+from beets.util import PromptChoice, displayable_path
 from beets.util.units import human_bytes, human_seconds_short
 
 from .display import (
@@ -366,12 +365,6 @@ def _summary_judgment(rec):
     elif action == importer.Action.ASIS:
         ui.print_("Importing as-is.")
     return action
-
-
-class PromptChoice(NamedTuple):
-    short: str
-    long: str
-    callback: Any
 
 
 def choose_candidate(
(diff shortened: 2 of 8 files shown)


henry-oberholtzer:
I like the additional deprecation warnings, but adding a disabled plugin list doesn't feel right to me when a plugin should be considered disabled by merit of not being on the list of plugins to begin with.

[My thoughts on a solution to the Musicbrainz problem:](https://github.com/beetbox/beets/issues/6121#issuecomment-3447778548)

> Throw an error, and ask the user if they'd like to use musicbrainz.
> 
> I'm not sure how possible it is for us to modify the user's config file, but if possible we could just add it to the list of enabled plugins and then the warning shouldn't appear again, unless the user intentionally switches for just discogs, or spotify, or similar.

snejus:
> I like the additional deprecation warnings, but adding a disabled plugin list doesn't feel right to me when a plugin should be considered disabled by merit of not being on the list of plugins to begin with.

Agree with you. I have now updated the logic and added a test to show the behaviour (otherwise there's too much to wrap one's head around!)

beets/__init__.py
```diff
@@ -26,13 +26,7 @@
...
     """Handle deprecated imports."""
     return deprecate_imports(
-        old_module=__name__,
-        new_module_by_name={
-            "art": "beetsplug._utils",
-            "vfs": "beetsplug._utils",
-        },
-        name=name,
-        version="3.0.0",
+        __name__, {"art": "beetsplug._utils", "vfs": "beetsplug._utils"}, name
```
semohr: Could you try to keep the trailing commas here? I recon we might add more imports here in the future.

See [COM812](https://docs.astral.sh/ruff/rules/missing-trailing-comma/):
The presence of a trailing comma can reduce diff size when parameters or elements are added or removed from function calls, function definitions, literals, etc.

snejus:
@semohr would be keen to merge it if it looks good.

Issue #152, opened 2013-02-28
Fix first letter caps


ghost:
This issue was automatically migrated from Google Code.
_Original author: j.kazem...@gmail.com (January 22, 2012 23:19:43)_
_Original issue: https://github.com/google-code-export/beets/issues/307_

minchinweb:
There is the [titlecase](https://pypi.org/project/titlecase/) module that may be of use for this.

PR #6133, opened 2025-10-26
New Plugin: Titlecase

henry-oberholtzer:
This plugin aims to address the shortcomings of the %title function, as brought up in issues #152, #3298 and an initial look to improvement with #3411. It supplies a new string format command, `%titlecase` which doesn't interfere with any prior expected behavior of the `%title` format command.

It also adds the ability to apply titlecase logic to metadata fields that a user selects, which is useful if you, like me, are looking for stylistic consistency and the minor stylistic differences between Musizbrainz, Discogs, Deezer etc, with title case are slightly infuriating.

This will add an optional dependency of [titlecase](https://pypi.org/project/titlecase/), which allows the titlecase core logic to be externally maintained.

If there's not enough draw to have this as a core plugin, I can also spin this into an independent one, but it seemed like a recurring theme that the %title string format didn't really behave as expected, and I wanted my metadata to match too.

- [x] Tests. - Not 100% coverage, but didn't see a lot of other plugins with testing for import stages.

### DIFF
--- a/beetsplug/titlecase.py
+++ b/beetsplug/titlecase.py
@@ -199,7 +199,6 @@ class TitlecasePlugin(BeetsPlugin):
         # Check we should split this into two substrings.
         if self.seperators:
             if len(splits := self.seperators.findall(text)):
-                print(splits)
                 split_cased = "".join(
                     [self.titlecase(s[0], field) + s[1] for s in splits]
                 )


beetsplug/titlecase.py
```diff
@@ -0,0 +1,184 @@
...
+    def __preserve_words__(self, preserve: list[str]) -> None:
+        for word in preserve:
+            if " " in word:
+                self.preserve_phrases[word] = re.compile(
+                    re.escape(word), re.IGNORECASE
+                )
+            else:
+                self.preserve[word.upper()] = word
+
+    def __preserved__(self, word, **kwargs) -> str | None:
```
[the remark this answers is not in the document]
henry-oberholtzer: I guess we *technically* do support 3.9...

beetsplug/titlecase.py
```diff
@@ -0,0 +1,184 @@
...
+        self.__init_fields_to_process__(
+            self.config["fields"].as_str_seq(),
+        )
+
+    def __init_fields_to_process__(self, fields: list[str]) -> None:
+        """Creates the set for fields to process in tagging.
+        Only uses fields included.
+        Last, the EXCLUDED_INFO_FIELDS are removed to prevent unitentional modification.
+        """
+        initial_field_list = set(fields)
```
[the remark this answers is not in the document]
henry-oberholtzer: beet looking at my past commits have we? saw the ideas i abandoned and wanted to bring back 'em back, eh?

beetsplug/titlecase.py
```diff
@@ -0,0 +1,192 @@
...
+    preserve: dict[str, str] = {}
+    preserve_phrases: dict[str, Pattern[str]] = {}
+    force_lowercase: bool = True
+    fields_to_process: set[str]
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        # Register template function
+        self.template_funcs["titlecase"] = self.titlecase  # type: ignore
```
henry-oberholtzer: Could not get mypy to play nice with this type here - any suggestions appreciated.

semohr: I think the issue is that the type is 
```python
# plugins.py
template_funcs: TFuncMap[str] | None = None
# and should be
template_funcs: ClassVar[TFuncMap[str]] = {}
```

Notice the union with None. Should be fine here as this is more of an issue with the plugin implementation.

beetsplug/titlecase.py
```diff
@@ -0,0 +1,192 @@
...
+                "fields": [],
+                "force_lowercase": False,
+                "small_first_last": True,
+            }
+        )
+
+        """
+        auto - Automatically apply titlecase to new import metadata.
+        preserve - Provide a list of words/acronyms with specific case requirements.
+        fields - Fields to apply titlecase to, default is all.
```
semohr: I can't seem to find the `default is all` case. Where is this handled?

beetsplug/titlecase.py
```diff
@@ -0,0 +1,192 @@
...
+from beets.library import Item
+from beets.plugins import BeetsPlugin
+
+__author__ = "henryoberholtzer@gmail.com"
+__version__ = "1.0"
+
+# These fields are excluded to avoid modifying anything
+# that may be case sensistive, or important to database
+# function
+EXCLUDED_INFO_FIELDS: set[str] = {
```
semohr: This list seems a bit like a potential footgun to me:
- How can we be sure these are all available sensitive fields?
- How do we make sure this is maintained?
- What about plugins that add field?

Maybe we should turn this around? Could we have a list with default fields and let the user add more if needed?

beetsplug/titlecase.py
henry-oberholtzer: Ah! This is a holdover from my previous version where I initially just had it do all fields by default. Thanks for catching it. Now it only does fields as specified, which I think can play into a solution to your other comment!

beetsplug/titlecase.py
henry-oberholtzer: Good point - I was a little worried about the maintenance of it. Thinking on it, since the fields to tag have to be manually specified, we could remove this and let users apply titlecase to any field they please. We can then rely on re-importing / re-tagging to fix any issues if a user, for some reason, chooses to titlecase a musicbrainz id.

henry-oberholtzer:
Want to add an option intercept candidates for auto tagging before the distances have been calculated, so the changes the plugin makes are more obvious on import - have reverted to draft until then.

beetsplug/titlecase.py
```diff
@@ -0,0 +1,208 @@
...
+                self.register_listener(
+                    "trackinfo_received", self.received_info_handler
+                )
+                self.register_listener(
+                    "albuminfo_received", self.received_info_handler
+                )
+
+    def _get_config(self):
+        self.force_lowercase = self.config["force_lowercase"].get(bool)
+        self.replace = self.config["replace"].as_pairs()
```
semohr: Nit:

I think we should not attach the config values directly as class attributes. Using a property should be fine i.e.
```python
@property
def force_lowercase(self) -> bool:
    return self.config["force_lowercase"].get(bool)
```

This allows proper typing while keeping the class a bit cleaner.

docs/plugins/titlecase.rst
```diff
@@ -0,0 +1,175 @@
...
+    the first being the target, and the second being the replacement
+
+.. conf:: force_lowercase
+    :default: no
+
+    Force all strings to lowercase before applying titlecase, but can cause
+    problems with all caps acronyms titlecase would otherwise recognize.
+
+.. conf:: small_first_last
+
```
semohr, line 137: Default value documentation is missing here. Same for `the_artist` and `after_choice`.

docs/plugins/titlecase.rst
```diff
@@ -0,0 +1,175 @@
...
+~~~~~~~~~~~~~~~~
+
+``titlecase`` only ever modifies string fields, however, this doesn't prevent
+you from selecting a case sensitive field that another plugin or feature may
+rely on.
+
+In particular, including any of the following in your configuration could lead
+to unintended behavior:
+
+.. code-block:: bash
```
semohr, line 167: I like that we just put that into the docs now! This gives users all the freedom in using the plugin however they see fit.

docs/plugins/titlecase.rst
```diff
@@ -0,0 +1,175 @@
...
+    Whether to automatically apply titlecase to new imports.
+
+.. conf:: fields
+
+    A list of fields to apply the titlecase logic to. You must specify the fields
+    you want to have modified in order for titlecase to apply changes to metadata.
+
+    A good starting point is below, which will titlecase artists, album and track titles.
+
+.. code-block:: yaml
```
semohr, line 79: Do you think we want to make this the default value? Could improve initial usability for new users.

semohr (changes requested):
This looks great to me! I added some minor comment.
 
We might want to cleanup the git commit history a bit. E.g. there are quite a few merges from master which seem unnecessary to me. We could also do a squash, depends a bit which commit information we want to keep.  

Further the `poetry.lock` file changed a bit more than I expected for a singular addition of the `titlecase` dependency. We might want to not update the other dependencies here.

Changelog entry is missing.

docs/plugins/titlecase.rst
henry-oberholtzer: I like this idea, but I worry about causing the confusion with a default value that is completely overridden once a user makes a change. I think it'd be similar to the confusion we're seeing with musicbrainz being invisibly enabled by default.

Since this plugin isn't enabled by default and has to be sought out and installed, I think it's safer to assume users will be more likely to read the docs for it and configure it. Let me know if you see it otherwise though.

docs/plugins/titlecase.rst
henry-oberholtzer: Thanks for the catch! I'll get these in.

beetsplug/titlecase.py
henry-oberholtzer: Oh cool! I like this style. My main thought with storing the config values as class attributes was to prevent re-reading the config value for each titlecase operation.

I might want to add a `@cached_property` decorator to the properties in this case to prevent it from having to read the config file for each operation, unless that's a feature of the property decorator I didn't pick up in the docs.

henry-oberholtzer:
> We might want to cleanup the git commit history a bit. E.g. there are quite a few merges from master which seem unnecessary to me. We could also do a squash, depends a bit which commit information we want to keep.

I think we could squash and clean this up to whatever degree feels good, I'll defer to your opinion on it since you have more experience with this project.

> Further the `poetry.lock` file changed a bit more than I expected for a singular addition of the `titlecase` dependency. We might want to not update the other dependencies here.

I'll double check the pyproject.toml and lock file to make sure they're not too different from the master branch, I think we might be seeing the merge in of the recent updates? It should've run with --no-update....

docs/plugins/titlecase.rst
semohr: Sounds good to me :+1:

semohr:
> I'll double check the pyproject.toml and lock file to make sure they're not too different from the master branch, I think we might be seeing the merge in of the recent updates? It should've run with --no-update....

I just used the github diff view, it is a bit flakey recently. Should be gone after a rebase than :+1:

henry-oberholtzer:
Think we should be good to go now - probably easier just to squash merge this, especially since it's just one whole new feature. I'll make sure to squash irrelevant commits as I go next time.

beetsplug/titlecase.py
```diff
@@ -0,0 +1,237 @@
...
+        force_lowercase - Lowercases the string before titlecasing.
+        small_first_last - If small characters should be cased at the start of strings.
+        the_artist - If the plugin infers the field to be an artist field
+        (e.g. the field contains "artist")
+        It will capitalize a lowercase The, helpful for the artist names
+        that start with 'The', like 'The Who' or 'The Talking Heads' when
+        they are not at the start of a string. Superceded by preserved phrases.
+        """
+        # Register template function
+        self.template_funcs["titlecase"] = self.titlecase  # type: ignore
```
semohr: We can remove the this `type: ignore` now 🎉 We recently merged a PR with a typing fix.

henry-oberholtzer: Yay!

henry-oberholtzer:
Got that test added. It is isn't as much of an integration test with the whole import flow, but it makes sure that the logic of the function does what it is supposed to.

semohr (approved):

PR #6138, opened 2025-10-30
Spotify: gracefully handle deprecated audio-features API

arsaboo:
Spotify has deprecated many of its APIs that we are still using, wasting calls and time on these API calls; also results in frequent rate limits. 

This PR introduces a dedicated `AudioFeaturesUnavailableError` and tracks audio feature availability with an `audio_features_available` flag. If the audio-features endpoint returns an HTTP 403 error, raise a new error, log a warning once, and disable further audio-features requests for the session. 

The plugin now skips attempting audio-features lookups when disabled (avoiding repeated failed calls and rate-limit issues). 

Also, update the changelog to document the behavior.

## To Do

### DIFF
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -193,8 +193,7 @@ class SpotifyPlugin(
             response.raise_for_status()
         except requests.exceptions.HTTPError as e:
             raise ui.UserError(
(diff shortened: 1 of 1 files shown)


henry-oberholtzer (approved):
No concerns on this from me, nicely handled with the multi-threaded implementation.

PR #6140, opened 2025-10-31
New Plugin: Save skipped imports

EmberLightVFX:
## Description

This plugin will save the name of the skipped song/album to a text file during import for later review.

It will also allow you to try to find the Spotify link for the skipped songs if the Spotify plugin is installed and configured.
This information can later be used together with other MB importers like Harmony.

If any song has already been written to the file, it will not be written again.

## To Do

## Extra Info

I haven't written any tests for this plugin as I'm not sure how to write tests for this kind of thing. If anyone knows how to write some good tests for this plugin I'm more than happy to add them to this PR!

I personally wrote this plugin to later use with Harmony as I currently manually save the skipped song-info from my library of not found songs, search spotify for them and add them through Harmony. Currently I have over 100 skipped songs and I'm only on the letter D in my library so this plugin will save me A LOT of time getting the information to later add.

### DIFF
--- a/beetsplug/saveskippedsongs.py
+++ b/beetsplug/saveskippedsongs.py
@@ -1,0 +1,157 @@
+# This file is part of beets.
+# Copyright 2025, Jacob Danell.
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+
+"""
+Save all skipped songs to a text file for later review.
+This plugin uses the Spotify plugin (if available) to try to find
+the Spotify links for the skipped songs.
+"""
+
+import os
+from typing import TYPE_CHECKING, Optional
+
+from beets import plugins
+from beets.importer import Action
+from beets.plugins import BeetsPlugin
+
+if TYPE_CHECKING:
+    from beets.metadata_plugins import SearchFilter
+    from beets.plugins import ImportSession, ImportTask
+
+__author__ = "jacob@emberlight.se"
+__version__ = "1.0"
+
+
+def summary(task: "ImportTask"):
+    """Given an ImportTask, produce a short string identifying the
+    object.
+    """
+    if task.is_album:
+        return f"{task.cur_artist} - {task.cur_album}"
+    else:
+        item = task.item  # type: ignore[attr-defined]
+        return f"{item.artist} - {item.title}"
+
+
+class SaveSkippedSongsPlugin(BeetsPlugin):
+    def __init__(self):
+        """Initialize the plugin and read configuration."""
+        super().__init__()
+        self.config.add(
+            {
+                "spotify": True,
+                "path": "skipped_songs.txt",
+            }
+        )
+        self.register_listener("import_task_choice", self.log_skipped_song)
+
+    def log_skipped_song(self, task: "ImportTask", session: "ImportSession"):
+        if task.choice_flag == Action.SKIP:
+            # If spotify integration is enabled, try to match with Spotify
+            link = None
+            if self.config["spotify"].get(bool):
+                link = self._match_with_spotify(task, session)
+
+            result = f"{summary(task)}{' (' + link + ')' if link else ''}"
+            self._log.info(f"Skipped: {result}")
+            path = self.config["path"].get(str)
+            if path:
+                path = os.path.abspath(path)
+                try:
+                    # Read existing lines (if file exists) and avoid duplicates.
+                    try:
+                        with open(path, "r", encoding="utf-8") as f:
+                            existing = {line.rstrip("\n") for line in f}
+                    except FileNotFoundError:
+                        existing = set()
+
+                    if result not in existing:
+                        with open(path, "a", encoding="utf-8") as f:
+                            f.write(f"{result}\n")
+                    else:
+                        self._log.debug(f"Song already recorded in {path}")
+                except OSError as exc:
+                    # Don't crash import; just log the I/O problem.
+                    self._log.debug(
+                        f"Could not write skipped song to {path}: {exc}"
+                    )
+
+    def _match_with_spotify(
+        self, task: "ImportTask", session: "ImportSession"
+    ) -> Optional[str]:
+        """Try to match the skipped track/album with Spotify by directly
+        calling the Spotify API search.
+        """
+        try:
+            # Try to get the spotify plugin if it's already loaded
+            spotify_plugin = None
+            for plugin in plugins.find_plugins():
+                if plugin.name == "spotify":
+                    spotify_plugin = plugin
+                    break
+
+            # If not loaded, try to load it dynamically
+            if not spotify_plugin:
+                try:
+                    from beetsplug.spotify import SpotifyPlugin
+
+                    spotify_plugin = SpotifyPlugin()
+                    self._log.debug("Loaded Spotify plugin dynamically")
+                except ImportError as e:
+                    self._log.debug(f"Could not import Spotify plugin: {e}")
+                    return
+                except Exception as e:
+                    self._log.debug(f"Could not initialize Spotify plugin: {e}")
+                    return
+
+            # Build search parameters based on the task
+            query_filters: SearchFilter = {}
+            if task.is_album:
+                query_string = task.cur_album or ""
+                if task.cur_artist:
+                    query_filters["artist"] = task.cur_artist
+                search_type = "album"
+            else:
+                # For singleton imports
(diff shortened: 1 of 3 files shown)


beetsplug/saveskippedsongs.py
```diff
@@ -0,0 +1,157 @@
...
+            }
+        )
+        self.register_listener("import_task_choice", self.log_skipped_song)
+
+    def log_skipped_song(self, task: "ImportTask", session: "ImportSession"):
+        if task.choice_flag == Action.SKIP:
+            # If spotify integration is enabled, try to match with Spotify
+            link = None
+            if self.config["spotify"].get(bool):
+                link = self._match_with_spotify(task, session)
```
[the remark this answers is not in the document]
EmberLightVFX: Skipping this fix as get(bool) is being used everywhere else in the code base and not as_bool()

beetsplug/saveskippedsongs.py
```diff
@@ -0,0 +1,157 @@
...
+                if item.album:
+                    query_filters["album"] = item.album
+                search_type = "track"
+
+            self._log.info(
+                f"Searching Spotify for: {query_string} ({query_filters})"
+            )
+
+            # Call the Spotify API directly via the plugin's search method
+            results = spotify_plugin._search_api(  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
EmberLightVFX: _search_api is needed instead of re-implementing the whole spotify-connection codebase. If the spotify plugin would update it's easier to just update this plugin too than duplicating the whole code.

beetsplug/saveskippedsongs.py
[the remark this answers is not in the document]
henry-oberholtzer: yep -[ get() comes from confuse](https://confuse.readthedocs.io/en/latest/api.html#confuse.core.ConfigView.get)

beetsplug/saveskippedsongs.py
```diff
@@ -0,0 +1,162 @@
...
+        """Try to match the skipped track/album with Spotify by directly
+        calling the Spotify API search.
+        """
+        try:
+            # Try to get the spotify plugin if it's already loaded
+            spotify_plugin = None
+            for plugin in plugins.find_plugins():
+                if plugin.name == "spotify":
+                    spotify_plugin = plugin
+                    break
```
henry-oberholtzer, line 107: I think we can assume that if a user doesn't have spotify loaded as a plugin, they may not be interested in the spotify feature.

Could also avoid having to make as second API call as well, since by the time the user application or may skip a song, it will have probably already attempted to grab candidates from the import task. It should be available under `ImportTask.candidates` [here](https://github.com/beetbox/beets/blob/master/beets/importer/), and then it'd just be a member of the `AlbumInfo.info` or `TrackMatch.info` [object](https://github.com/beetbox/beets/blob/master/beets/autotag/hooks.py) - which should come with the distance already calculated nicely too. Could let the user just filter what database source URLs they wanted printed with it in a config option.

beetsplug/saveskippedsongs.py
```diff
@@ -0,0 +1,162 @@
...
+
+if TYPE_CHECKING:
+    from beets.importer import ImportSession, SingletonImportTask
+    from beets.metadata_plugins import SearchFilter
+
+__author__ = "jacob@emberlight.se"
+__version__ = "1.0"
+
+
+def summary(task: "SingletonImportTask"):
```
henry-oberholtzer, line 36: I think the more general ImportTask is a better suited typehint than the than the derived SingletonImportTask - I don't think there's any functionality in this plugin that appears to be Singleton only?

docs/plugins/saveskippedsongs.rst
```diff
@@ -0,0 +1,25 @@
...
+Configuration
+-------------
+
+To configure the plugin, make a ``saveskippedsongs:`` section in your
+configuration file. The available options are:
+
+- **spotify**: Search Spotify for the song/album and return the link. Default:
+  ``yes``.
+- **path**: Path to the file to write the skipped songs to. Default:
+  ``skipped_songs.txt``.
```
henry-oberholtzer, line 25: Might be good to specify that it stores in the user's home directory.

henry-oberholtzer (changes requested):
Hi! This looks great so far! It is probably is better to hook into the existing database querying mechanism's results (as I noted in the review) rather than having to re-invent the Spotify API wheel. It'll also make it really easy to increase how useful this is for users who may not user Spotify.

docs/plugins/saveskippedsongs.rst
EmberLightVFX: For me it will save the file in the dir that I run the beets command from. Should I maybe change the default path to `~/skipped_songs.txt` to make it a more specified path?

beetsplug/saveskippedsongs.py
EmberLightVFX: It would be if you skipp a singleton song?

beetsplug/saveskippedsongs.py
EmberLightVFX: I probably just don't really understand your explenation so if I'm wrong please correct me.

For the addition of the spotify links to be added you would need the spotify plugin to be configured but disabled in your config. If it's active beets will mostly just pick the spotify match as the best match and move on (This is some info I should add to the documentation now when thinking about it).

If the spotify plugin is disabled we would need to do the API call when the user presses skip to see if there is any Spotify matches (without picking it as the beets match)

JOJ0:
Hi @EmberLightVFX, thank you for the submission! Interesting idea for a plugin. On a very high level and only having skimmed through this PR's conversation and code quickly, I would like to mention that I see quite similar use-cases for this plugin and the built-in log-file feature of Beets. Before moving on with this plugin, I would like to invite you to compare what benefits this plugin offers over the built-in import log (except the Spotify feature)

So first of all, in case one of you (including @henry-oberholtzer here), hasn't used it yet. This is the crash course:

- Configure in config.yaml to write skipped imports to a predefined log file (https://beets.readthedocs.io/en/stable/reference/config.html#log)
- Or use `beet import --log= ....` for specific import runs
- Run `beet import --quiet`
- Complete interactive-required imports later with `beet import --from-logfile ....`

Quick and dirty here, I can certainly go into more details if you want. I'm also thinking that this feature might not be too obvious to find out how to use properly for new Beets users since it's rather poorly documented and writing a proper tutorial might also be worth putting effort into.

JOJ0:
@EmberLightVFX @henry-oberholtzer any thoughts? Did you find time to think about my comparision with the internal log-file feature? I'm setting this PR to draft for now. Should we close it already and maybe only the spotify-suggest will be a future feature? How should we proceed?

henry-oberholtzer:
Import from log file seems to do a lot of what this plug aims to do. It could certainly do with more highlighting in the documentation! This plugin seems to differ from `--from-logfile` in that it saves candidates, so maybe that'd be a different plugin direction.

JOJ0:
> Import from log file seems to do a lot of what this plug aims to do. It could certainly do with more highlighting in the documentation! This plugin seems to differ from `--from-logfile` in that it saves candidates, so maybe that'd be a different plugin direction.

I agree that --from-logfile is kind of badly documented. When I first replied to this thread I wanted to simplay send you docs but found they are lacking detail on how to really use it, so I started to describe it quickly myself.

Is this plugin in its current state saving candidates to the text file already? From what I read here it only saves "artist - title" (which in my opinion is not very useful for a reimport later on. _Is_ the purpose of this plugin even to move on with importing "difficult" to find tracks/albums or did I misunderstand it in the first place? Please rephrase again (sorry I might be slow here) @EmberLightVFX what you tried to do in the docs already (Harmony MusicBrainz importer for example)

https://github.com/beetbox/beets/pull/6140/files#diff-748c48fa17458aae81d7c28f64bf287aaf5864c919b16cbb8b62333daab78acbR42

https://github.com/beetbox/beets/pull/6140/files#diff-bbab3c937e587079482d4ff607fbd08ffb8886972381c44df243746a71d8e1bdR4-R5

I don't want to be at all opposed on creating a new and fancy plugin that helps people better import stuff into their Beets library but I'm not yet sure what we are really aiming for. 

This plugin does a lot of file reading/writing, must handle errors, and even calls spotify and records things. There is a lot that can go wrong. The spotify search is only artist/title, that will bring a lot of false positives. Well, this could be a little improved if it would search via `ISRC` and then only fall back to artist/title/album. But even then I'm questioning: When a file to import has been skipped in the importer, most probably because it was hard to find from all the sources (mb,discogs,spotify,deezer,....) why will it be helpful to search for it on spotify again? If we didn't find it already, we won't find it then either, right?

Ooor asking again: Is the purpose not to find out what is difficult to import / no matches found / save tedious and manualy importing work in a logfile to work through in batch? Is there a different use-case?

So to wrap this up:

- What do both of you think about taking the time to thoroughly document the from-logifle feature first, then we would see whether it can be improved (I'm sure there is much room for improvement) and were a plugin like we have here _add_ even more value to the existing feature (or do something else entirely)

- Also: Please provide some examples of how that textfile can look like, I'm having a hard time figuring out how it looks from the mentioned code snippets above, I have an idea but might not see all the details. That would help for a better high level review or maybe a concept refinement going forward.

PR #6142, opened 2025-11-02
Introduce Info.name property and add types to match details functions

snejus:
# Generalize some of common tagging functionality to `Info` and `Match` base classes

This PR centralises some common tagging functionality between singletons and albums allowing to simplify `ChangeRepresentation` and importing functionality. This is prep work for a larger PR which refactors and simplifies the entire tagging workflow.

## Changes

- **Core type changes**: Changed `mapping` parameter from `Mapping[Item, TrackInfo]` to `list[tuple[Item, TrackInfo]]` in `apply_metadata()`, `distance()`, and `assign_items()` functions
- **Match dataclasses**: Converted `AlbumMatch` and `TrackMatch` from `NamedTuple` to `@dataclass`, introducing base `Match` class with common functionality
- **New properties**: Added `name` cached property to `Info` class for unified name access
- **ChangeRepresentation refactor**: Converted to `@dataclass` with lazy property evaluation, replacing `cur_album`/`cur_title` with unified `cur_name` field
- **UI improvements**: Simplified display logic by using `match.info.name` instead of type-specific field checks
- **Parameter renaming**: Renamed `search_album`/`search_title` parameters to `search_name` for consistency across singleton and album workflows

The changes maintain backward compatibility in behavior while improving type safety and code clarity.

### DIFF
--- a/beets/ui/commands/import_/display.py
+++ b/beets/ui/commands/import_/display.py
@@ -17,6 +17,7 @@ if TYPE_CHECKING:
 
     import confuse
 
+    from beets.autotag.distance import Distance
     from beets.library.models import Item
     from beets.ui import ColorName
 
@@ -47,7 +48,7 @@ class ChangeRepresentation:
         return ui.colorize("changed", "\u2260")
 
     @cached_property
-    def _indentation_config(self) -> confuse.ConfigView:
+    def _indentation_config(self) -> confuse.Subview:
         return config["ui"]["import"]["indentation"]
 
     @cached_property
@@ -196,7 +197,7 @@ class ChangeRepresentation:
             return str(index)
 
     def make_track_numbers(
-        self, item, track_info: hooks.TrackInfo
+        self, item: Item, track_info: hooks.TrackInfo
     ) -> tuple[str, str, bool]:
         """Format colored track indices."""
         cur_track = self.format_index(item)
@@ -307,7 +308,7 @@ class ChangeRepresentation:
             }
             return (lhs, rhs)
 
-    def print_tracklist(self, lines):
+    def print_tracklist(self, lines: list[tuple[Side, Side]]) -> None:
         """Calculates column widths for tracks stored as line tuples:
         (left, right). Then prints each line of tracklist.
         """
@@ -315,7 +316,7 @@ class ChangeRepresentation:
             # If no lines provided, e.g. details not required, do nothing.
             return
 
-        def get_width(side):
+        def get_width(side: Side) -> int:
             """Return the width of left or right in uncolorized characters."""
             try:
                 return len(
@@ -458,7 +459,7 @@ def show_item_change(item: Item, match: hooks.TrackMatch) -> None:
     change.show_match_details()
 
 
-def disambig_string(info):
+def disambig_string(info: hooks.Info) -> str:
     """Generate a string for an AlbumInfo or TrackInfo object that
     provides context that helps disambiguate similar-looking albums and
     tracks.
@@ -524,7 +525,7 @@ def get_album_disambig_fields(info: hooks.AlbumInfo) -> Sequence[str]:
     return out
 
 
-def dist_colorize(string, dist):
+def dist_colorize(string: str, dist: Distance) -> str:
     """Formats a string as a colorized similarity string according to
     a distance.
     """
@@ -537,7 +538,7 @@ def dist_colorize(string, dist):
     return string
 
 
-def dist_string(dist):
+def dist_string(dist: Distance) -> str:
     """Formats a distance (a float) as a colorized similarity percentage
     string.
     """
@@ -545,7 +546,7 @@ def dist_string(dist):
     return dist_colorize(string, dist)
 
 
-def penalty_string(distance, limit=None):
+def penalty_string(distance: Distance, limit: int | None = None) -> str:
     """Returns a colorized string that indicates all the penalties
     applied to a distance object.
     """
@@ -561,3 +562,5 @@ def penalty_string(distance, limit=None):
         # Prefix penalty string with U+2260: Not Equal To
         penalty_string = f"\u2260 {', '.join(penalties)}"
         return ui.colorize("changed", penalty_string)
+
+    return ""


semohr:
Just had a short look and this will absolutly break beets-flask downstream. We use the mappings dict quite a bit... Can we move this change in particular to a different PR?

I think the other things are mainly no brainers.

I need to think abit about the implications a how to migrate this in beets-flask. Just a bit more time here would be highly appreciated :)

snejus:
> Just had a short look and this will absolutly break beets-flask downstream. We use the mappings dict quite a bit... Can we move this change in particular to a different PR?
> 
> I think the other things are mainly no brainers.
> 
> I need to think abit about the implications a how to migrate this in beets-flask. Just a bit more time here would be highly appreciated :)

No worries - I kept `Match.mapping` the same now - instead, I introduced `item_info_pairs` property.

semohr:
Renaming does not really help me here :laughing: 

Where is the usecase for `list[tuple[Item, TrackInfo]]` over `Mapping[Item, TrackInfo]`? I feel like the later is cleaner if there is not a specific edgecase I'm missing.

snejus:
> Renaming does not really help me here 😆

I'm not sure that I understand - `AlbumMatch.mapping` is now present like it was before. I just introduce a new property `item_info_pairs`:

```py
@dataclass
class AlbumMatch(Match):
    info: AlbumInfo
    mapping: dict[Item, TrackInfo]
    extra_items: list[Item]
    extra_tracks: list[TrackInfo]

    @property
    def item_info_pairs(self) -> list[tuple[Item, TrackInfo]]:
        return list(self.mapping.items())
```


> Where is the usecase for `list[tuple[Item, TrackInfo]]` over `Mapping[Item, TrackInfo]`? I feel like the later is cleaner if there is not a specific edgecase I'm missing.

It's just a wrong data type for how this property is used across the codebase - it's used for recording and looping across matching `(item, track_info)` pairs for the album. No functionality in the codebase makes use of the *mapping* nature of it - instead, values are accessed as `mapping.items()` which just gets the underlying list of `(item, track_info)` pairs.

I thus clarified it with the new property.

snejus:
Is there anything missing here @semohr?

beets/autotag/match.py
```diff
@@ -86,16 +86,17 @@ def assign_items(
...
-    mapping = {
-        items[iidx]: t
+    item_info_pairs = [
+        (items[iidx], t)
         for iidx, t in zip(assigned_item_idxs, tracks)
         if iidx != -1
-    }
-    extra_items = list(set(items) - mapping.keys())
+    ]
+    item_info_pairs.sort(key=lambda it: (it[0].disc, it[0].track, it[0].title))
```
semohr: Have you done some light performance evaluation? We introduce a sort here which was not necessary beforehand as we had used a hashmap previously for the mappings (now item_info_pairs). For long compilation albums that might be noticeable.

beets/autotag/hooks.py
```diff
@@ -214,16 +230,32 @@ def __init__(
 
 
 # Structures that compose all the information for a candidate match.
+@dataclass
```
semohr, line 233: Do we require the dataclass somewhere? Dataclasses do introduce some overhead, especially for attribute access. I would remove them if they are not entirely required and add them once we really do need them.

beets/autotag/match.py
```diff
@@ -86,16 +86,17 @@ def assign_items(
...
+    extra_items = list(set(items) - {i for i, _ in item_info_pairs})
     extra_items.sort(key=lambda i: (i.disc, i.track, i.title))
-    extra_tracks = list(set(tracks) - set(mapping.values()))
+    extra_tracks = list(set(tracks) - {t for _, t in item_info_pairs})
```
semohr: We could use a numpy ndarray for `item_info_pairs`. Would simplify this to `item_info_pairs[:,0]` and gernally might help in some other places. Not sure we want to go this route tho, as Im not sure how well numpy behaves with none primitive types. But might be worth a considerarion.

If you decide to do some performance eval I would be very interested in how numpy compares here.

beets/autotag/match.py
snejus: Not entirely sure now why did I introduce sort here - will remove it.

beets/autotag/hooks.py
snejus: It allows to abstract `TrackMatch` and `AlbumMatch`!

beets/autotag/match.py
snejus: Not sure if performance is very relevant here as the difference is going to be negligible given the rest of the pipeline, e.g., autotagger search. I'd rather keep this a tuple since it's fed into `AlbumMatch`.

beets/ui/commands/import_/display.py
```diff
@@ -198,12 +195,15 @@ def format_index(self, track_info):
         else:
             return str(index)
 
-    def make_track_numbers(self, item, track_info):
+    def make_track_numbers(
+        self, item, track_info: hooks.TrackInfo
```
semohr: Partial typehint (missing item here). I think there is a mypy rule which could catch this 🤔

beets/ui/commands/import_/display.py
```diff
@@ -426,21 +425,17 @@ def show_match_tracks(self):
...
 
-def show_change(cur_artist, cur_album, match):
+def show_change(
+    cur_artist: str, cur_album: str, match: hooks.AlbumMatch
+) -> None:
     """Print out a representation of the changes that will be made if an
     album's tags are changed according to `match`, which must be an AlbumMatch
     object.
     """
-    change = AlbumChange(
```
semohr, line 441: Is the removing of the kwargs intended here? Does not seem that way to me.

I dont really mind either way but modifies the git history unnecessarily.

beets/ui/commands/import_/display.py
```diff
@@ -452,13 +447,11 @@ def show_change(cur_artist, cur_album, match):
     change.show_match_tracks()
 
 
-def show_item_change(item, match):
+def show_item_change(item: Item, match: hooks.TrackMatch) -> None:
     """Print out the change that would occur by tagging `item` with the
     metadata from `match`, a TrackMatch object.
     """
-    change = TrackChange(
```
semohr, line 459: see https://github.com/beetbox/beets/pull/6142/changes#r2628854338

beets/autotag/hooks.py
semohr: Doesn't work for namedtuples or extending from dict directly I guess. We would have to define an init function any way in this case, ah well seems like it is the cleanest way than. Thanks!

beets/autotag/match.py
semohr: For most cases you are right it should not matter. Although since we now change the pattern I think we could profit from a bit of evaluation. Now is the time, otherwise we probably wont touch/evaluate this anymore. I recon since we ship numpy anyways as dependency it might be a minor and easy improvement and worth the short check.

It is not required for users to run with metadata-sources that trigger online (long running) lookups. In theory local sources exist and can be used. Beets is pretty impartial to its usage and for some this might be beneficial :shrug:

semohr:
Btw I really appreciate that you started to properly type some parts in the commands submodule. I was hopeing someone would start with this after the bigger refactor :P

beets/autotag/match.py
snejus: [Asked AI](https://www.perplexity.ai/search/https-github-com-beetbox-beets-ZLRNYfGHQ36OrOxu4dNDUg#0) for an opinion:
 
> ### My Assessment: **Not a Good Idea**
> 
> The ndarray suggestion is **not valid** for several reasons:
> 
> 1. **Type Incompatibility**: `item_info_pairs` contains tuples of `(Item, TrackInfo)` objects - these are complex Python objects, not primitive numeric types. NumPy arrays work best with homogeneous numeric data, and using them with arbitrary Python objects defeats most of their performance advantages.
> 
> 2. **No Real Performance Benefit**: The reviewer acknowledges this is speculative. Since:
>    - The operations are simple indexing/extraction on small-to-medium sized lists (typically dozens of items, not millions)
>    - The bottleneck is network I/O for metadata lookups, not in-memory list operations
>    - NumPy's overhead for object arrays can actually be **slower** than native Python lists for small datasets
> 
> 3. **Added Complexity & Dependencies**: While NumPy is already a dependency (likely for audio processing), forcing it into the matching logic adds unnecessary coupling and complexity for operations Python handles natively and efficiently.
> 
> 4. **Code Readability**: `[item for item, _ in item_info_pairs]` is Pythonic and clear. The NumPy syntax `item_info_pairs[:,0]` isn't meaningfully simpler and may confuse contributors unfamiliar with NumPy.
> 
> 5. **Structural Mismatch**: As the author notes, these tuples are passed directly to `AlbumMatch`. Converting to/from NumPy arrays would add conversion overhead without benefit.

beets/autotag/hooks.py
```diff
@@ -214,16 +230,32 @@ def __init__(
...
     extra_tracks: list[TrackInfo]
 
+    @property
+    def item_info_pairs(self) -> list[tuple[Item, TrackInfo]]:
+        return list(self.mapping.items())
 
-class TrackMatch(NamedTuple):
-    distance: Distance
+    @property
+    def items(self) -> list[Item]:
```
semohr, line 255: This should be `self.mapping.keys()` instead no need to loop over the `item_info_pairs`.

beets/autotag/match.py
semohr: Would still be interested in the performance regress since we replaced direct memory-views (`.keys` `.values`) with two python loops here. Tbh, the AI comment is not helpful here as it just repeats what we have said already. It did not take into consideration the changes to the previous version. 

I was trying to find a better solution here that omits the two new loops without requiring us to go back to a dictionary. Numpy array indexing  (`[:, 0]`) as I have suggested uses memory-views internally. This is not about a premature optimization but a potential performance **regression** in the `assign_items` function.

---

I think you got wrong where I'm going here: I dont really care if we use numpy or tuples or mapping here but there should **not** be a measurable performance degradation. Alternatively, keeping the original mapping here and returning `item_info_pairs` might also be possible. This would allow use to keep the new syntax in other places but keep the old version of the assign_items function.

beets/ui/commands/import_/display.py
snejus: Added missing type defs using `--strict`.

It also complained about missing types in confuse:
image

So I got completely distracted and spent a couple of hours adding types to `confuse` 😅

beets/ui/commands/import_/display.py
snejus: Yes. I removed unnecessary verbosity.

One example where this is relevant: I'm refactoring this file further in another branch where I'm replacing `show_item_change` and `show_change` function pair with a single unified function. Having

```py
    change = AlbumChange(cur_artist, cur_album, match)
    ...
    change = TrackChange(item.artist, item.title, match)
```

it's easier to tell the difference between the two calls.

beets/autotag/match.py
semohr: I just created a small benchmark and tried to be fair here:

```python
def tuple_v(
    items: Sequence[Item],
    tracks: Sequence[TrackInfo],
):
    item_info_pairs = [
        (items[iidx], t) for iidx, t in enumerate(tracks) if iidx != -1
    ]
    extra_items = list(set(items) - {i for i, _ in item_info_pairs})
    extra_items.sort(key=lambda i: (i.disc, i.track, i.title))
    extra_tracks = list(set(tracks) - {t for _, t in item_info_pairs})
    extra_tracks.sort(key=lambda t: (t.index, t.title))
    return item_info_pairs, extra_items, extra_tracks


def mapping_v(
    items: Sequence[Item],
    tracks: Sequence[TrackInfo],
):
    mapping = {items[iidx]: t for iidx, t in enumerate(tracks) if iidx != -1}
    extra_items = list(set(items) - mapping.keys())
    extra_items.sort(key=lambda i: (i.disc, i.track, i.title))
    extra_tracks = list(set(tracks) - set(mapping.values()))
    extra_tracks.sort(key=lambda t: (t.index, t.title))
    return mapping.items(), extra_items, extra_tracks


def test_bench_tuple(benchmark):

    items = [Item(mb_artistid="1", mb_artistids=["1", "2"])] * 100
    tracks = [TrackInfo(title="Track")] * 100
    benchmark(tuple_v, items, tracks)


def test_bench_mapping(benchmark):

    items = [Item(mb_artistid="1", mb_artistids=["1", "2"])] * 100
    tracks = [TrackInfo(title="Track")] * 100
    benchmark(mapping_v, items, tracks)
```

We get the following results (mostly independent of the number of tracks/items:

```
----------------------------------------------------------------------------------- benchmark: 2 tests ----------------------------------------------------------------------------------
Name (time in us)         Min                 Max              Mean            StdDev            Median               IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_bench_mapping     3.1290 (1.0)       58.2490 (1.0)      3.2620 (1.0)      0.4868 (1.0)      3.2200 (1.0)      0.0690 (1.0)      611;1876      306.5620 (1.0)       97192           1
test_bench_tuple       4.9070 (1.57)     101.3280 (1.74)     5.6402 (1.73)     2.5126 (5.16)     5.0750 (1.58)     0.0970 (1.41)    3742;5101      177.3000 (0.58)      58658           1
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

This shows the mapping version is approx **1.7x** faster with much lower variance.

---

Considering this we should keep the old version here.

beets/autotag/hooks.py
snejus: I kept the `mapping` only because you requested so - it's gone from the rest of the codebase 😅

semohr: It is still the main data container right now for the pairs, I'm open to changing this in general at some point but this needs a bit more thought in my opinion. Maybe properties for extra_* and pairs might be an elegant way forward with one data container for items and one for tracks and index mappings for the properties. How I think about this is kinda like a venn diagram.
```
// items ∩ tracks = pairs
// items' ∩ tracks = extra_items
// items ∩ tracks' = extra_tracks
// pairs + extra_items + extra_tracks = items + tracks
```

So all these properties are derived from some items and tracks with some kind of index mapping.

--- 

Might allow us to remove the whole mapping/sorting step in the `assign_items` function and move it into the match class.

beets/ui/commands/import_/display.py
semohr: True :+1:

beets/autotag/match.py
snejus: Tested it:

```py
>>> from beets.library import *
... from beets.autotag.hooks import TrackInfo
... from beets.autotag import match
... 
... def item(i, length):
...     return Item(
...         artist="ben harper",
...         album="burn to shine",
...         title=f"ben harper - Burn to Shine {i}",
...         track=i,
...         length=length,
...     )
... 
... items = []
... items.append(item(1, 241.37243007106997))
... items.append(item(2, 342.27781704375036))
... items.append(item(3, 245.95070222338137))
... items.append(item(4, 472.87662515485437))
... items.append(item(5, 279.1759535763187))
... items.append(item(6, 270.33333768012))
... items.append(item(7, 247.83435613222923))
... items.append(item(8, 216.54504531525072))
... items.append(item(9, 225.72775379800484))
... items.append(item(10, 317.7643606963552))
... items.append(item(11, 243.57001238834192))
... items.append(item(12, 186.45916150485752))
... 
... def info(index, title, length):
...     return TrackInfo(title=title, length=length, index=index)
... 
... trackinfo = []
... trackinfo.append(info(1, "Alone", 238.893))
... trackinfo.append(info(2, "The Woman in You", 341.44))
... trackinfo.append(info(3, "Less", 245.59999999999999))
... trackinfo.append(info(4, "Two Hands of a Prayer", 470.49299999999999))
... trackinfo.append(info(5, "Please Bleed", 277.86599999999999))
... trackinfo.append(info(6, "Suzie Blue", 269.30599999999998))
... trackinfo.append(info(7, "Steal My Kisses", 245.36000000000001))
... trackinfo.append(info(8, "Burn to Shine", 214.90600000000001))
... trackinfo.append(info(9, "Show Me a Little Shame", 224.0929999999999))
... trackinfo.append(info(10, "Forgiven", 317.19999999999999))
... trackinfo.append(info(11, "Beloved One", 243.733))
... trackinfo.append(info(12, "In the Lord's Arms", 186.13300000000001))
```

### `master`
```py
>>> timeit(lambda: match.assign_items(items, trackinfo), number=100)
2.0667866870062426

>>> timeit(lambda: match.assign_items(items, trackinfo), number=100)
1.9120936749968678

>>> timeit(lambda: match.assign_items(items, trackinfo), number=100)
1.9020310329506174
```

### this branch
```py
>>> timeit(lambda: match.assign_items(items, trackinfo), number=100)
2.1010622539324686

>>> timeit(lambda: match.assign_items(items, trackinfo), number=100)
1.9093402739381418

>>> timeit(lambda: match.assign_items(items, trackinfo), number=100)
1.8998970040120184
```

snejus: On the other hand you raised a good point regarding `mapping.keys` and `mapping.values` - they keep the code cleaner, so I reverted to constructing the mapping for these operations.

semohr: Im a bit confused by the results now tbh. In my micro bench it was more clear, in your more macroscopic bench it looks like the tuple version is mostly the same  :thinking: 

Give this I would most likely just go with the most consistent or ergonomic version.

snejus: Well, my test underlines that the difference is negligible since `assign_items` function spends most of the time calculating item assignment :)

semohr: True always depends on the scale. Im kinda opposed in adding python loops if the yare not required that's mainly why my alarm bells went off here.

beets/autotag/hooks.py
snejus: That's the general idea I'm following in my WIP refactor of the `match.py` module. Unfortunately, there's one complicating factor, see:

```py
     1	@dataclass
     2	class AlbumMatch(Match[AlbumInfo]):
     3	    mapping: dict[Item, TrackInfo]
     4	    extra_items: list[Item] = field(default_factory=list)
     5	    extra_tracks: list[TrackInfo] = field(default_factory=list)
     6	
     7	    @cached_classproperty
     8	    def required_fields(cls) -> Sequence[str]:
     9	        return config["match"]["required"].as_str_seq()
    10	
    11	    @cached_classproperty
    12	    def ignored_fields(cls) -> Sequence[str]:
    13	        return config["match"]["ignored"].as_str_seq()
    14	
    15	    def __post_init__(self) -> None:
    16	        plugins.send("album_matched", match=self)
    17	
    18	    @classmethod
    19	    def try_create(cls, info: AlbumInfo, source: Source) -> AlbumMatch | None:
    20	        """Validate and create an AlbumMatch from the given info and source."""
    21	        log.debug("Candidate: {!r}", info)
    22	
    23	        # Discard albums with zero tracks.
    24	        if not info.tracks:
    25	            log.debug("No tracks.")
    26	            return None
    27	
    28	        # Discard matches without required tags.
    29	        if missing_tags := {
    30	            f for f in cls.required_fields if info.get(f) is None
    31	        }:
    32	            log.debug("Ignored. Missing required tag: {}", missing_tags)
    33	            return None
    34	
    35	        # Find mapping between the items and the track info.
    36	        items = source.items
    37	        mapping, extra_items, extra_tracks = assign_items(items, info.tracks)
    38	
    39	        # Get the change distance.
    40	        _distance = distance(
    41	            source.data, info, mapping, len(items) - len(mapping)
    42	        )
    43	
    44	        # Skip matches with ignored penalties.
    45	        penalties = [key for key, _ in _distance]
    46	        for penalty in cls.ignored_fields:
    47	            if penalty in penalties:
    48	                log.debug("Ignored. Penalty: {}", penalty)
    49	                return None
    50	
    51	        log.debug("Success. Distance: {}", _distance)
    52	        return cls(_distance, info, mapping, extra_items, extra_tracks)
```

See validation on lines 46-49 - this depends on `distance`, which in turns depends on the assignment. Essentially, we need to perform item assignment before constructing `AlbumMatch` object :-(

semohr: I was more thinking of a data structure like this.

```python
class AlbumMatch:
    items: list[Item] # on disk
    tracks: list[TrackInfo] # from source
    _pairs_idx: list[tuple[int,int] # matching via assignment problem  (lsa)
    
    @property
    def pairs(self):
        return ...
 ```
 
Creating the assignments outside the class is fine in my book. I see this more as a data container with some helper functions to get the transformed data.

PR #6144, opened 2025-11-05
Drop support for Python 3.9

snejus:
Drop support for Python 3.9 and pyupgrade the codebase.

Dependency upgrades:

image

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -3,6 +3,10 @@ on:
   workflow_dispatch:
   schedule:
     - cron: "0 0 * * SUN" # run every Sunday at midnight
+
+env:
+  PYTHON_VERSION: "3.10"
+
 jobs:
   test_integration:
     runs-on: ubuntu-latest
@@ -12,7 +16,7 @@ jobs:
         uses: BrandonLWhite/pipx-install-action@v1.0.3
       - uses: actions/setup-python@v6
         with:
-          python-version: 3.9
+          python-version: ${{ env.PYTHON_VERSION }}
           cache: poetry
 
       - name: Install dependencies
--- a/beetsplug/bpd/__init__.py
+++ b/beetsplug/bpd/__init__.py
@@ -283,7 +283,7 @@ class BaseServer:
         if not self.ctrl_sock:
             self.ctrl_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
             self.ctrl_sock.connect((self.ctrl_host, self.ctrl_port))
-        self.ctrl_sock.sendall((f"{message}\n").encode("utf-8"))
+        self.ctrl_sock.sendall((f"{message}\n").encode())
 
     def _send_event(self, event):
         """Notify subscribed connections of an event."""
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -23,7 +23,7 @@ from collections import OrderedDict
 from contextlib import closing
 from enum import Enum
 from functools import cached_property
-from typing import TYPE_CHECKING, AnyStr, ClassVar, Literal, Tuple, Type
+from typing import TYPE_CHECKING, AnyStr, ClassVar, Literal
 
 import confuse
 import requests
@@ -86,7 +86,7 @@ class Candidate:
         path: None | bytes = None,
         url: None | str = None,
         match: None | MetadataMatch = None,
-        size: None | Tuple[int, int] = None,
+        size: None | tuple[int, int] = None,
     ):
         self._log = log
         self.path = path
@@ -682,7 +682,7 @@ class GoogleImages(RemoteArtSource):
         """
         if not (album.albumartist and album.album):
             return
-        search_string = f"{album.albumartist},{album.album}".encode("utf-8")
+        search_string = f"{album.albumartist},{album.album}".encode()
 
         try:
             response = self.request(
@@ -1293,7 +1293,7 @@ class CoverArtUrl(RemoteArtSource):
 
 
 # All art sources. The order they will be tried in is specified by the config.
-ART_SOURCES: set[Type[ArtSource]] = {
+ART_SOURCES: set[type[ArtSource]] = {
     FileSystem,
     CoverArtArchive,
     ITunesStore,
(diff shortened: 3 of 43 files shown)


beets/test/_common.py
```diff
@@ -19,12 +19,10 @@
 import unittest
 from contextlib import contextmanager
 
-import beets
-import beets.library
```
semohr, line 23: How is this change related to the 3.9 deprecation?

beetsplug/bpd/__init__.py
```diff
@@ -28,11 +28,10 @@
 from string import Template
 from typing import TYPE_CHECKING
 
-import beets
-import beets.ui
```
semohr, line 32: I'm not convinced these import issues should be fixed in this PR. This is not really related to the python 3.9 deprecation.

beetsplug/bpd/__init__.py
```diff
@@ -283,7 +282,7 @@ def _ctrl_send(self, message):
         if not self.ctrl_sock:
             self.ctrl_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
             self.ctrl_sock.connect((self.ctrl_host, self.ctrl_port))
-        self.ctrl_sock.sendall((f"{message}\n").encode("utf-8"))
```
semohr, line 286: This change is not related to the 3.9 deprecation. The default value of `utf-8` was always a thing.

beetsplug/fetchart.py
```diff
@@ -86,7 +86,7 @@ def __init__(
         path: None | bytes = None,
         url: None | str = None,
         match: None | MetadataMatch = None,
-        size: None | Tuple[int, int] = None,
+        size: None | tuple[int, int] = None,
```
semohr, line 89: You fixed a lot of minor issues with exports from typing, it might be beneficial to enable some more ruff rules to make this consistent and prevent future (miss)use.

https://docs.astral.sh/ruff/rules/deprecated-import/
https://docs.astral.sh/ruff/rules/non-pep585-annotation/

Especially, we should also define a target version if we haven't done so already.

beetsplug/fetchart.py
```diff
@@ -682,7 +682,7 @@ def get(
         """
         if not (album.albumartist and album.album):
             return
-        search_string = f"{album.albumartist},{album.album}".encode("utf-8")
```
semohr, line 685: Same as above, I don't think this should change in this PR, does not seem related to the deprecation.

semohr (changes requested):
Looks mostly reasonable to me 👍  

You upgraded quite alot of dependencies, we might want to mention this in the change-log too. I guess this was necessary for the 3.10 bump?

I mentioned it in a comment too, but it might be worth it to add some more ruff rules now. Especially some of the pyupgrade ones. There are many rules that could be useful for us, these ones are the primary issues addressed in the PR: 
https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/ 
https://docs.astral.sh/ruff/rules/non-pep585-annotation/
https://docs.astral.sh/ruff/rules/super-call-with-parameters/

.github/workflows/integration_test.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
         uses: BrandonLWhite/pipx-install-action@v1.0.3
       - uses: actions/setup-python@v6
         with:
-          python-version: 3.9
+          python-version: 3.10
```
semohr: Should be `"3.10"`

beets/test/_common.py
```diff
@@ -99,7 +97,7 @@ def item(lib=None, **kwargs):
         album_id=None,
         mtime=12345,
     )
-    i = beets.library.Item(**{**defaults, **kwargs})
+    i = Item(**{**defaults, **kwargs})
```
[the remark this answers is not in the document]
sakgoyal: should this be `**(defaults | kwargs)` instead?

sakgoyal:
You should also squash these commits into a single commit instead of the current 5 commits. it keeps the git history cleaner. 

```sh
git reset --soft HEAD~5 # or however many commits you want to squash
git add .
git commit -m "..."
git push --force
```

beetsplug/bpd/__init__.py
snejus: `pyupgrade` replaced it 🤷🏼

beetsplug/bpd/__init__.py
snejus: I think I sneakily enabled `ruff check --preview` for this one. 😅 I have undone both changes.

.github/workflows/integration_test.yaml
snejus: Have replaced it with `PYTHON_VERSION` env var to make it consistent with the reset of the workflows.

snejus:
> You should also squash these commits into a single commit instead of the current 5 commits. it keeps the git history cleaner.
> 
> ```shell
> git reset --soft HEAD~5 # or however many commits you want to squash
> git add .
> git commit -m "..."
> git push --force
> ```

Each commit is tidy and intentional:

```
* 9 minutes ago             Šarūnas Nejus 9682bfd00 Enable all pyupgrade lint rules (HEAD -> drop-python-39)
* 3 days ago                Šarūnas Nejus f324bb0ce Ignore pyupgrade blame
# this commit is added to .git-blame-ignore-revs to keep original authors in the blame
* 3 days ago                Šarūnas Nejus 0f6250b52 pyupgrade Python 3.10
* 3 days ago                Šarūnas Nejus b9fde9a47 Update python version references
# shows which dependencies were upgraded specifically due to Python version upgrade
* 3 days ago                Šarūnas Nejus a7830beba Update python requirement and dependencies
# gives meaning to the next commit
* 3 days ago                Šarūnas Nejus d64efbb6c Upgrade deps before upgrade
```

snejus:
> You upgraded quite alot of dependencies, we might want to mention this in the change-log too. I guess this was necessary for the 3.10 bump?

Yes, and this may enable support for Python 3.14, actually. Let me check. I'm not sure how dependency version upgrades are relevant to users or maintainers - we update them often without a mention in the changelog.

semohr:
Not necessarily for users but for people using beets as a library it might be of interest. I agree tho, no need to add them to the changelog.

Just noticed we have scipy twice as dependency, you know how this happend?

snejus:
> Just noticed we have scipy twice as dependency, you know how this happend?

See `pyproject.toml` and the PR where I added support for Python 3.13 - it needed a higher version on Python 3.13 to fetch the wheel from PyPI.

semohr (approved):

PR #6149, opened 2025-11-07
Accept lyrics source as a string

### LINKED ISSUE #5962: lyrics: sources can be a string and beets doesn't complain
### Problem

#5931 discovered that `sources` can be a string and beets doesn't complain at all. This seems like an easy mistake to make, so we should probably catch it.

Annoyingly, `config["sources"].get(list)` doesn't actually throw an error on strings! So we probably just need some manual error checking.

DogeTheBeast:
## Description

Fixes #5962 .  

The fix was shared in the issue. Now it uses ```as_str_seq``` similarly to other plugins.

## To Do

- [X] ~Documentation. (If you've added a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)~

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -29,7 +29,6 @@ Bug fixes:
   audio-features endpoint, the plugin logs a warning once and skips audio
   features for all remaining tracks in the session, avoiding unnecessary API
   calls and rate limit exhaustion.
-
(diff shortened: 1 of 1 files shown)


DogeTheBeast:
This is my first PR. Please feel free to give feedback or things I should change.

semohr:
Thank you for the PR!

We're thrilled to have you contributing, welcome to the community!

semohr (approved):

PR #6152, opened 2025-11-08
Add Media Field for Album 

gcoppola19:
Fixes 6092  

Add a media field for album, when a user runs beets fields, media will be listed for album. This way, media of album can be kept track of.

 Documentation. (If you've added a new command-line flag, for example, find the appropriate page under docs/ to describe it.)
 Changelog. (Add an entry to docs/changelog.rst to the bottom of one of the lists near the top of the document.)
[x ] Tests.
test_media_field.py


henry-oberholtzer:
Hi Grace! Thanks for the PR and your interest in contributing to Beets!

We currently have the `media` field provided by the Item class, which represents an individual song or track, which currently can be queried on an album level, and allows support for albums to have tracks from different media types (e.g. tracks from both CD and DVD ripped to digital file).

If there's a use case that isn't addressed by the current implementation, more context & background for the issues this addresses would be appreciated!

Thanks!

endcredits33:
Hi @henry-oberholtzer, I created the original issue. The context here is when you run certain queries at album level, $media is not provided as it is not an album level field.

For example, running `beet ls -af '$albumartist - $album ($media)' label:Nonplus` gives:

```
Actress - Harrier ATTK / Gershwin ($media)
Actress - Machine And Voice ($media)
ASC - Nothing Is Certain ($media)
ASC - Nothing Is Certain LP Sampler ($media)
ASC - Porcelain / Focus Inwards ($media)
Barry Manalogue - Analogue / Koyo Front ($media)
```

arsaboo:
@endcredits33 indeed, because `media` for an album does not always make sense. What would you expect if one of the tracks is vinyl and the other track is DVD-ripped?

henry-oberholtzer:
@endcredits33 @gcoppola19 Ah, I see! I missed that.

I could see the album level query returning a list of the distinct media types from its tracks? I assume we never have albums without at least one Item.

endcredits33:
> @endcredits33 indeed, because `media` for an album does not always make sense. What would you expect if one of the tracks is vinyl and the other track is DVD-ripped?

Is this a common occurence? I think generally most releases tend to have a single media type. At the moment we're not displaying media at all for albums just because some releases _may_ contain several types.

henry-oberholtzer:
I think we need to be able to account for those instances, even if they're rare, and it should be possible to through DB queries.

gcoppola19:
@henry-oberholtzer @endcredits33 Can I work on implementing this? Where for an album, it would list all media types of the songs in the album?

arsaboo:
I don't see the value of doing this. Have you tried [album_fields](https://beets.readthedocs.io/en/stable/plugins/inline.html#album-fields)?

endcredits33:
Couldn't we make the same argument about other fields, e.g label, genre, catalog number? These can vary across the same release (anecdotally more often than media) and yet they are included as album level fields.

gcoppola19:
I wanted to implement it so that if a user wants to see the media types of songs in an album, it lists all the types, and made test cases to test if there was one media type, none, or multiple. Let me know if it needs any chances in functionality, I hope it's helpful!

snejus:
@gcoppola19 looking at the diff I'm struggling to understand how does the new `media-types` calculated field work?

Would you mind rebasing this on recent master? That issue in discogs has already been fixed as well, as far as I know.

snejus:

PR #5447, opened 2024-09-28
Reworked #4709 after latest release

arogl:
## Description

Fixes #4709 SQL use of Double Quoted Strings.

Replacing #5236

- [X] Changelog. (Add an entry to `docs/changelog.rst` to the bottom of one of the lists near the top of the document.)

PR #6158, opened 2025-11-11
Plugins/web: fix endpoints `/…/values/…`

tobast:
Following #4709 and #5447, the web plugin used single-quotes (ie. string litteral) in the SQL query for table columns. 

Thus, for instance, the query `GET /item/values/albumartist` would return the litteral "albumartist" instead of a list of unique album artists.

This prevents the Mopidy beets integration from working, returning the single artist "albumartist".

### DIFF
--- a/beetsplug/web/__init__.py
+++ b/beetsplug/web/__init__.py
@@ -17,9 +17,10 @@
 import base64
 import json
 import os
+import typing as t
 
 import flask
-from flask import g, jsonify
+from flask import jsonify
 from unidecode import unidecode
 from werkzeug.routing import BaseConverter, PathConverter
 
@@ -27,6 +28,17 @@ import beets.library
 from beets import ui, util
 from beets.dbcore.query import PathQuery
 from beets.plugins import BeetsPlugin
+
+# Type checking hacks
+
+if t.TYPE_CHECKING:
+
+    class LibraryCtx(flask.ctx._AppCtxGlobals):
+        lib: beets.library.Library
+
+    g = LibraryCtx()
+else:
+    from flask import g
 
 # Utilities.
 


beetsplug/web/__init__.py
```diff
@@ -232,7 +232,7 @@ def _get_unique_table_field_values(model, field, sort_field):
         raise KeyError
     with g.lib.transaction() as tx:
         rows = tx.query(
-            f"SELECT DISTINCT '{field}' FROM '{model._table}' ORDER BY '{sort_field}'"
+            f"SELECT DISTINCT {field} FROM {model._table} ORDER BY {sort_field}"
```
semohr, line 247: We should be a bit more careful here. Could we use subvals for field and sort_field here?

tobast: That indeed looked unsafe, but the [contribution guide](https://github.com/beetbox/beets/blob/master/CONTRIBUTING.rst#general) has more or less this precise example using direct formatting without any injection protection. Should this be updated too?

tobast: I actually spoke a bit too quickly: Python's sqlite3 placeholders are only for values (eg. string literals), not column names. They are formatted as single-quoted strings, and would do exactly the same as before this PR.

There seems to be no idiomatic way to do this. I expect it to be actually safe to use the current code, as the lines just above check that both `field` and `sort_field` are indeed column names.

semohr: Indeed, should be fine 👍

tobast:
> Seems like this is your first one! Glad to have you onboard 🎉

Thanks!

> If you are up for it, we could also add some typehints for the `g` object. See [here](https://github.com/pSpitzner/beets-flask/blob/b1b04d26341ab8cbe7283253e81b1b56c7cb44a2/backend/beets_flask/server/routes/library/__init__.py#L27-L34)

Why not! Would you rather have it in a separated PR or here?

semohr:
Up to you :upside_down_face: Here is totally fine with me. The web plugin needs a bit more love imo.

tobast:
I'm not certain this is the cleanest way, but that's the cleanest I found. Does it look fine to you? I didn't start the work of typing functions, however.

semohr (approved):

semohr:
Looks clean to me ;) 

Thank you for the contribution!

PR #6163, opened 2025-11-15
Support multiple pseudo-releases and reimport in mbpseudo

asardaes:
## Description

Hi again, today I tried to reimport an album to see if I could get the data from its pseudo-release and I realized there were a couple of things that could be improved.

First, determination of the best ref in `PseudoAlbumInfo` was not triggered during reimport because the flow is slightly different. To fix that I added a plugin method `before_album_info_emitted` that is explicitly called for metadata plugins. ~figured I could use the `albuminfo_received` event if I modified it slightly to add the `Iterable[Item]` to the event. I feel like this is ok because those 2 parameters are closely related, but let me know what you think; it seems no other plugin was using the event so far.~ See also my comment below.

Second, multiple pseudo-releases ([example](https://musicbrainz.org/release-group/08eaf50a-5389-42b8-be35-51a1df593c61)) were not really supported since the plugin always took the first one from the list, so I adjusted the code to compute some priority depending on the import languages from the config file.

Example where `jp` was first in the language list:

```
Finding tags for album "YOASOBI - 夜に駆ける".
  Candidates:
  1. (97.6%) YOASOBI - Yoru ni Kakeru
             ≠ media
             MusicBrainz, Digital Media, 2019, XW, None, YOASOBI-001, None
  2. (97.6%) YOASOBI - Racing Into The Night
             ≠ media
             MusicBrainz, Digital Media, 2019, XW, None, YOASOBI-001, None
  3. (97.6%) YOASOBI - 夜に駆ける
             ≠ media
             MusicBrainz, Digital Media, 2019, XW, None, YOASOBI-001, None
```

### DIFF
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -35,15 +35,12 @@ def find_metadata_source_plugins() -> list[MetadataSourcePlugin]:
     return [p for p in find_plugins() if hasattr(p, "data_source")]  # type: ignore[misc]
 
 
+@notify_info_yielded("albuminfo_received")
 def candidates(items, *args, **kwargs) -> Iterable[AlbumInfo]:
     """Return matching album candidates from all metadata source plugins."""
     for plugin in find_metadata_source_plugins():
         for info in plugin.candidates(items, *args, **kwargs):
-            send(
-                "albuminfo_received",
-                info=plugin.before_album_info_emitted(items, info),
-            )
-            yield info
+            yield plugin.before_album_info_emitted(items, info)
 
 
 @notify_info_yielded("trackinfo_received")
@@ -63,10 +60,8 @@ def album_for_id(
     """
     for plugin in find_metadata_source_plugins():
         if info := plugin.album_for_id(album_id=_id):
-            send(
-                "albuminfo_received",
-                info=plugin.before_album_info_emitted(items, info),
-            )
+            info = plugin.before_album_info_emitted(items, info)
+            send("albuminfo_received", info=info)
             return info
 
     return None


beetsplug/mbpseudo.py
```diff
@@ -112,6 +116,25 @@ def _on_plugins_loaded(self):
...
+        info: AlbumInfo,
+        items: Iterable[Item],
+    ):
+        if isinstance(info, PseudoAlbumInfo):
+            for item in items:
+                # particularly relevant for reimport but could also happen during import
+                if "mb_albumid" in item:
+                    del item["mb_albumid"]
+                if "mb_trackid" in item:
+                    del item["mb_trackid"]
```
asardaes, line 128: During reimport this leads to a lot of penalties in the distance calculation, so I hope doing something like this is fine; the changes aren't "flushed" and I guess those attributes will be set again after the reimport is done.

asardaes:
Now that I know that plugins can be hosted in external repositories, I'm not sure if it's really possible to just modify the `albuminfo_received` event, I guess that would break things. I think there was no other place where I could get access to the `Item`s, so maybe I won't be able to support reimport with this plugin 🤔

snejus:
@asardaes I'm finding that the way this plugin is designed is causing issues in

1. https://github.com/beetbox/beets/pull/6052 where I'm struggling to rebase this work on `master` because this plugin is so tightly coupled to `musicbrainz` but is defined as a separate plugin
2. The way it dynamically adjusts `AlbumMatch` prevents proper refactoring of tagging functionality, see https://github.com/beetbox/beets/pull/6184.

Therefore, it needs to be restructured. From what I've seen, it seems that this functionality should be part of the existing `musicbrainz` plugin (and toggled/enabled using the configuration). We should probably consider creating a `musicbrainz` directory under `beetsplug` to organise functionality related to MB more effectively:

```diff
beetsplug/
-├── musicbrainz.py
+└── musicbrainz/
+    ├── mbpseudo.py
+    └── musicbrainz.py
```

I already started working on this and will get on it next once I'm finished with #6052, so for now please do not make significant changes in this plugin, thanks!

--- 

If you're interested I asked AI to review this plugin architecture:

<details>
<summary>Key architectural concerns in <code>mbpseudo</code></summary>

- **Mutually exclusive design:** `_on_plugins_loaded` raises if the standard `musicbrainz` plugin is enabled, meaning `mbpseudo` reimplements large chunks of MusicBrainz logic instead of composing with, or extending, the existing plugin. This tight coupling makes maintenance difficult (any upstream change must be duplicated) and prevents users from mixing behaviors that might otherwise be complementary.

- **Deep reach into MusicBrainz internals:** The plugin calls private helpers like `_merge_pseudo_and_actual_album` and `_preferred_alias`, taps `musicbrainzngs.get_release_by_id` directly, and overrides both `candidates()` and `album_info()`. Because it relies on internal, non-public functions and data structures, even minor upstream refactors can break `mbpseudo`. A more stable architecture would expose an explicit hook or shared service for "alternate releases" rather than patching the full candidate pipeline.

- **Complex stateful `AlbumInfo` wrapper:** `PseudoAlbumInfo` overrides `__getattr__`, toggles `self.__dict__['_pseudo_source']`, mutates mappings after `album_matched`, and implements custom deepcopy logic. This "dual view" approach works but complicates reasoning: downstream code sees one object whose semantics change depending on hidden flags, which increases the risk of subtle bugs (e.g., caching or serialization issues). A cleaner design would keep pseudo and official releases as separate `AlbumInfo` instances with explicit metadata about preference rather than switching identity midstream.

- **Configuration/model mismatch:** The plugin inserts custom `MediaFields` for every configured tag at runtime, even though they may already exist or conflict. The `custom_tags_only` switch branches behavior inside `album_info`, mutating album or track fields in place. This suggests the architecture should separate "metadata decoration" from "candidate selection" so each concern can evolve independently (e.g., use hook-based transformers to add tags, and a separate strategy object for pseudo-release selection).

Overall, the plugin achieves its goal but does so by tightly grafting onto MusicBrainz internals and using stateful wrappers. A more modular architecture, sharing infrastructural hooks with `musicbrainz`, exposing a dedicated pseudo-release provider interface, and avoiding mutable proxy objects, would make the feature easier to reason about, test, and maintain.

</details>

asardaes:
Hi @snejus, what would need to change for #6052? The only call made directly with `musicbrainzngs` is somewhat abstracted by:

```py
self._release_getter = musicbrainzngs.get_release_by_id
```

so I would have expected it's easy to swap that if necessary.

As for the other issue, I think it might be fine trying to just merge everything into the main `musicbrainz` plugin, but I don't know if that's enough to ease the issues with `AlbumMatch`. Let me know where the issue is, I can try to improve that too.

snejus:
We've now merged the relevant adjustments in #6052 - feel free to rebase and see what's changed.

> but I don't know if that's enough to ease the issues with `AlbumMatch`

Could you describe why there is a need to make adjustments to `AlbumMatch`? Couldn't we simply return `AlbumInfo` candidates for the original and each of the pseudo releases?

asardaes:
@snejus So the idea was the following. Imagine the file you're about to import is for a track titled `百花繚乱`. After searching in MB, we find the release and a pseudo-release translating the track as `In Bloom`. If we simply emit the album info for the pseudo-release, the distance between the file's title (`百花繚乱`) and the pseudo-release's title (`In Bloom`) will be very large, so it won't be a good candidate. What the plugin's logic does is: it initially uses the original title and exposes that information to the distance calculations, and after the distance has been computed, it switches to the translated title in the final album match so the translation can be used for the imported file's name and in the DB.

snejus:
Makes sense, thanks!

Does it account for the reverse scenario? Let's say my item has `title="In Bloom"` in the DB. Am I going to see both `title="In Bloom"` and `title=百花繚乱` as top candidates when I search?

asardaes:
@snejus I think that should work, at least in theory. The initial search in MB might yield the pseudo-release, and the core logic from the `musicbrainz` plugin would resolve the main release from that, so both should end up in the candidates.

Did I understand correctly that I should wait for #6184 before I take another look at these changes? Or should we agree on the expectations before that?

snejus:
Try rebasing on the current `master` and I will try to review this :)

asardaes:
I opened a separate PR that merges the functionality into the main plugin.

PR #6164, opened 2025-11-15
Add album template value in ftintitle plugin

asardaes:
## Description

I was hoping to use the functionality from `ftintitle` to set the path's album artist as the main artist, but that wasn't possible, so I added a template value `album_artist_no_feat`.

### DIFF
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -151,9 +151,9 @@ class BeetsPlugin(metaclass=abc.ABCMeta):
         list
     )
     listeners: ClassVar[dict[EventType, list[Listener]]] = defaultdict(list)
-    template_funcs: ClassVar[TFuncMap[str]] = {}
-    template_fields: ClassVar[TFuncMap[Item]] = {}
-    album_template_fields: ClassVar[TFuncMap[Album]] = {}
+    template_funcs: ClassVar[TFuncMap[str]] | TFuncMap[str] = {}  # type: ignore[valid-type]
+    template_fields: ClassVar[TFuncMap[Item]] | TFuncMap[Item] = {}  # type: ignore[valid-type]
+    album_template_fields: ClassVar[TFuncMap[Album]] | TFuncMap[Album] = {}  # type: ignore[valid-type]
 
     name: str
     config: ConfigView
@@ -219,14 +219,14 @@ class BeetsPlugin(metaclass=abc.ABCMeta):
         self.name = name or self.__module__.split(".")[-1]
         self.config = beets.config[self.name]
 
-        # Set class attributes if they are not already set
-        # for the type of plugin.
+        # If the class attributes are not set, initialize as instance attributes.
+        # TODO: Revise with v3.0.0, see also type: ignore[valid-type] above
         if not self.template_funcs:
-            self.template_funcs = {}  # type: ignore[misc]
+            self.template_funcs = {}
         if not self.template_fields:
-            self.template_fields = {}  # type: ignore[misc]
+            self.template_fields = {}
         if not self.album_template_fields:
-            self.album_template_fields = {}  # type: ignore[misc]
+            self.album_template_fields = {}
 
         self.early_import_stages = []
         self.import_stages = []


beetsplug/ftintitle.py
```diff
@@ -129,6 +134,12 @@ def __init__(self) -> None:
         if self.config["auto"]:
             self.import_stages = [self.imported]
 
+        # help mypy .-.
+        if album_template_fields := self.album_template_fields:
```
semohr: You could try something like:
```python
self.album_template_fields = {
    **(self.album_template_fields or {}),
    "album_artist_no_feat": _album_artist_no_feat,
}
```
`album_template_fields` is not used as class variable so it should be fine to overwrite on self.

We should improve the typing for `album_template_fields`, `template_fields`  and `template_funcs`. I dont think the None union is entirely required.

asardaes: I improved the types instead, it was pretty straightforward.

beets/plugins.py
```diff
@@ -368,8 +363,6 @@ def template_func(cls, name: str) -> Callable[[TFunc[str]], TFunc[str]]:
         """
 
         def helper(func: TFunc[str]) -> TFunc[str]:
-            if cls.template_funcs is None:
```
asardaes, line 371: Maybe this helper and the one below aren't even needed anymore.

semohr: I think we want to keep `album_template_fields`, `template_fields`, `template_funcs` as class variables, these were intended to be a shared registry for all plugins. This is more or less a `borg` pattern if you want to readup on it.

Not sure how and where this is used in practice but I think there was a reason for this at some point. We should be able to add `ClassVar[TFuncMap]` as typhint and initialize them before the constructor. Using the same pattern as `listeners`.

asardaes: I see, maybe the union with `None` was used because otherwise mypy throws an error, so I instead suppressed mypy on those lines, would you agree with that?

semohr: What is the error?

Can't we remove this entirely now?
```
        if not self.album_template_fields:
            self.album_template_fields = {}
            self.album_template_fields = {}  # type: ignore[misc]
```

semohr: Interestingly this part of the code removes the classvar definitions entirely. Each plugin will create its own `album_template_fields` it was never used as classvar.

Also we have the following, which does not make too much sense with classvars too.
```
def album_field_getters() -> TFuncMap[Album]:
    """As above, for album fields."""
    funcs: TFuncMap[Album] = {}
    for plugin in find_plugins():
        _check_conflicts_and_merge(plugin, plugin.album_template_fields, funcs)
    return funcs
```

Seems like we might be better of removing the classvar entierly. Sorry for the confusion

asardaes: From the comments I thought that the idea was that the class variables could contain methods for specific child classes, something like 

```python
class FtInTitlePlugin(plugins.BeetsPlugin):

    album_template_fields = {
        "album_artist_no_feat": _album_artist_no_feat,
    }

    def __init__(self) -> None:
    # do not modify album_template_fields at all
```

So when any instance of `FtInTitlePlugin` dispatches to the `__init__` method of its parent, `if not self.album_template_fields` would be false and the *instance* variables would not be created at all. I haven't validated that, but maybe it's not worth having that flexibility anyway?

BTW the error was `error: Cannot assign to class variable "template_funcs" via instance  [misc]`

semohr: This makes sense in theory. Do we use that somewhere in practice? I agree this might not be worth the effort to maintain. We have to be very careful here tho. We should not break any existing plugins.

asardaes: I don't see any explicit references to this from class variables, so if there's no reflection magic that searches for template fields/funcs, I think it can be removed:

image

semohr: Just researched this a bit:

It was possible at some point (and still is) to use the classmethods as decorator:

```
@MyPlugin.template_func("initial")
def _t_func():
    pass
```

There might be some external plugins (not in the beets repo) which use that pattern. It was mentioned in the documentation at some point. If we change this, this breaks these plugins (althought I could not find any that use this).

I'm not that confident we can remove the ClassVars entirely here without a major version bump. We could schedule it for deprecation in 3.0.0 tho and/or add a comment to have another look here with 3.0.0.

asardaes: Got it, then maybe it is worth it.

semohr (approved):
Thanks again for the PR! 

Taking the effort to enhance the `template_funcs` stuff is highly appreciated 👍 I think we just leave it like it is with the `type: ignore` for now. I added an additional comment for us to revise this with 3.0.0.

I enabled auto-merge, once you update the branch it should merge into master.

asardaes:
Sorry, it seems I shouldn't rebase since it disables your merge configuration

Issue #6181, opened 2025-11-23
Autotagger considers candidates with same album id but from different data sources as duplicates


snejus:
### Problem

While fixing #6178 I discovered another problem: apparently we do not take into account
the `data_source` field when we check whether an incoming candidate is a duplicate.

I have a test that reproduces this issue (included the data source in the logs for clarity):

```sh
===================================================================================== FAILURES =====================================================================================
_________________________________________________________________ TestGetAlbumMultipleDataSources.test_search_ids __________________________________________________________________

self = <test.autotag.test_match.TestGetAlbumMultipleDataSources object at 0x7fb73532e560>, shared_id = 'album-12345'

    def test_search_ids(self, shared_id):
        _, _, proposal = match.tag_album([Item()], search_ids=[shared_id])
    
        sources = [
            candidate.info.data_source for candidate in proposal.candidates
        ]
>       assert len(sources) == 2
E       AssertionError: assert 1 == 2
E        +  where 1 = len(['Deezer'])

test/autotag/test_match.py:148: AssertionError
------------------------------------------------------------------------------- Captured stderr call -------------------------------------------------------------------------------
Tagging  - 
Searching for album ID: album-12345
Sending event: albuminfo_received
Candidate: Albumartist - Album (album-12345) from Deezer
Computing track assignment...
...done.
Success. Distance: 1.00
Sending event: album_matched
Sending event: albuminfo_received
Candidate: Albumartist - Album (album-12345) from Discogs
Duplicate.
Sending event: album_matched
Evaluating 1 candidates.
```

I cannot reproduce this with my library data because (apparently) `discogs` plugin returns `AlbumInfo.album_id` field as an integer while `deezer` returns a string :grin:.

### Setup

* beets version: 2.5.1

PR #6179, opened 2025-11-23
Discogs: fixes for: 6177, 6068

henry-oberholtzer:
## Description

Fixes #6177, #6068.

I fixed the issue in #6177 by removing the derived class interface, and moving those fields back into function variables. They're a bit unwieldy still, but that's the algorithm it came with. There's a lot of room to continue to improve the clarity of the code in that section, but I think that'll require a deeper overhaul.

For #6068, I created the `ArtistInfo` and `AlbumArtistInfo` typed dictionaries, and was able to centralize the logic of building the artist info into `build_artistinfo` and `build_albumartistinfo`. Tests for these scenarios were created largely by expanding existing tests to incorporate the new fields.

I think I'll have to re-think the entire algorithm for 6030 to make it more flexible at parsing the issue for #6030, so I'll move that to a later PR in the interest of getting the flex attr fix in.

## To Do

- [x] Changelog. (Add an entry to `docs/changelog.rst` to the bottom of one of the lists near the top of the document.)
- [x] Tests.

PR #6184, opened 2025-11-23
Return candidates from all data sources on id search

### LINKED ISSUE #6178: Autotagger returns only the first candidate for an ID that is present in multiple sources
### Problem

I am reimporting an album from Discogs that has an ID `853460`. To underline the issue,
I use only **Discogs** and **Deezer** data sources and I set `search_limit: 1` for both:

```sh
$ beet -vv import -LI now music 65 --disable-plugins=musicbrainz,spotify,bandcamp
user configuration: /home/sarunas/.config/beets/config.yaml
data directory: /home/sarunas/.config/beets
plugin paths: []
Loading plugins: autobpm, bareasc, bench, deezer, discogs, edit, embedart, fetchart, importreplace, info, inline, lastgenre, lastimport, lyrics, mbcollection, mbsync, missing, mpdstats, mpdupdate, random, replaygain, scrub, types, unimported
inline: adding item field has_lyrics
inline: adding item field label_or_albumartist
inline: adding item field singleton_track_artist
inline: adding item field track_artist
inline: adding item field album_name
inline: adding item field track_identification
inline: adding item field withdrawn
inline: adding album field label_or_albumartist
inline: adding album field multiple_artists
lastgenre: Loading whitelist /home/sarunas/repo/beets/beetsplug/lastgenre/genres.txt
Sending event: pluginload
library database: /home/sarunas/.music/beets/library.db
library directory: /run/media/sarunas/music/Music
Sending event: library_opened
Sending event: import_begin
Parsed query: AndQuery([OrQuery([SubstringQuery('album', 'now', fast=True), SubstringQuery('albumartist', 'now', fast=True), SubstringQuery('genre', 'now', fast=True)]), OrQuery([SubstringQuery('album', 'music', fast=True), SubstringQuery('albumartist', 'music', fast=True), SubstringQuery('genre', 'music', fast=True)]), OrQuery([SubstringQuery('album', '65', fast=True), SubstringQuery('albumartist', '65', fast=True), SubstringQuery('genre', '65', fast=True)])])
Parsed sort: NullSort()
yielding album 12136: Various Artists - Now That's What I Call Music! 65
Looking up: /run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_01_scissorsisters_idontfeellikedancin.flac; /run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_02_shakirafeatwyclefjeanfeatwyclefjean_hipsdontlie.flac; /run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_05_amywinehouse_rehab.flac; /run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_09_justintimberlake_sexyback.flac; /run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_10_nellyfurtadofeattimbalandfeattimbaland_promiscuous.flac; /run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_11_beyonce_dejavu.flac; /run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_12_chamillionairefeatkrayziebonefeatkrayziebone_ridin.flac; /run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_14_feddelegrand_putyourhandsupfordetroit.flac; /run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_15_fergie_londonbridge.flac; /run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_16_thepussycatdolls_idontneedaman.flac; /run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_18_cassie_meu.flac; /run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_20_rihanna_unfaithful.flac
Tagging Various Artists - Now That's What I Call Music! 65
Searching for discovered album ID: 853460
Candidate: JJ Symon & The Monochromes - Black, White & Grey (853460)
Computing track assignment...
...done.
Success. Distance: 0.84
Sending event: album_matched
Album ID match recommendation is 1
Search terms: Various Artists - Now That's What I Call Music! 65
Album might be VA: True
deezer: Searching Deezer for 'album:'Now That's What I Call Music! 65''
deezer: Found 1 result(s) from Deezer for 'album:'Now That's What I Call Music! 65''
Candidate: Eric Dane Jaqua - Now That's What I Call EDJOMB! (796914581)
Computing track assignment...
...done.
Success. Distance: 0.73
Sending event: album_matched
Candidate: Various Artists - Now That's What I Call Rock (25179484)
Computing track assignment...
...done.
Success. Distance: 0.71
Sending event: album_matched
Evaluating 3 candidates.

/run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_01_scissorsisters_idontfeellikedancin.flac
/run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_02_shakirafeatwyclefjeanfeatwyclefjean_hipsdontlie.flac
/run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_05_amywinehouse_rehab.flac
/run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_09_justintimberlake_sexyback.flac
/run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_10_nellyfurtadofeattimbalandfeattimbaland_promiscuous.flac
/run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_11_beyonce_dejavu.flac
/run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_12_chamillionairefeatkrayziebonefeatkrayziebone_ridin.flac
/run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_14_feddelegrand_putyourhandsupfordetroit.flac
/run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_15_fergie_londonbridge.flac
/run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_16_thepussycatdolls_idontneedaman.flac
/run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_18_cassie_meu.flac
/run/media/sarunas/music/Music/virgin/nowthatswhaticallmusic65/1_20_rihanna_unfaithful.flac (12 items)
Sending event: import_task_before_choice
Sending event: before_choose_candidate

Finding tags for album "Various Artists - Now That's What I Call Music! 65".
  Candidates:
  1. (29.2%) Various Artists - Now That's What I Call Rock
             ≠ label, year, album, ...
             Discogs, Various Artists, Now That's What I Call Rock, 3xVinyl, 2022, None, None, LPRKNOW1, Sony Music, LP, Compilation, Stereo, None, Europe, rock
  2. (27.0%) Eric Dane Jaqua - Now That's What I Call EDJOMB!
             ≠ artist, label, year, ...
             Deezer, Eric Dane Jaqua, Now That's What I Call EDJOMB!, None, 2025, 7, 2, None, Quo Vadis, album, None, None, None
  3. (15.6%) JJ Symon & The Monochromes - Black, White & Grey
             ≠ artist, album, label, ...
             Deezer, JJ Symon & The Monochromes, Black, White & Grey, None, 2009, 11, 21, None, E.O.I. Productions, album, None, None, None
```

The most relevant logs are here:

```sh
Searching for discovered album ID: 853460
Candidate: JJ Symon & The Monochromes - Black, White & Grey (853460)
Computing track assignment...
...done.
Success. Distance: 0.84
Sending event: album_matched
Album ID match recommendation is 1
Search terms: Various Artists - Now That's What I Call Music! 65
...

Finding tags for album "Various Artists - Now That's What I Call Music! 65".
  Candidates:
  1. (29.2%) Various Artists - Now That's What I Call Rock
             ≠ label, year, album, ...
             Discogs, Various Artists, Now That's What I Call Rock, 3xVinyl, 2022, None, None, LPRKNOW1, Sony Music, LP, Compilation, Stereo, None, Europe, rock
  2. (27.0%) Eric Dane Jaqua - Now That's What I Call EDJOMB!
             ≠ artist, label, year, ...
             Deezer, Eric Dane Jaqua, Now That's What I Call EDJOMB!, None, 2025, 7, 2, None, Quo Vadis, album, None, None, None
  3. (15.6%) JJ Symon & The Monochromes - Black, White & Grey
             ≠ artist, album, label, ...
             Deezer, JJ Symon & The Monochromes, Black, White & Grey, None, 2009, 11, 21, None, E.O.I. Productions, album, None, None, None
```

`beets` found a match on **Deezer** with the same ID `853460` and it did not attempt to try
any further data sources, even though the album I am looking for comes from **Discogs**.

This is especially an issue considering that the candidate from **Deezer** is not a good
match to my original album.

### Setup

* beets version: 2.5.1

snejus:
Closes #6178 (multiple metadata source results per ID) and #6181 (duplicate/overwrite of candidates).

- (#6178) Replace `album_for_id` / `track_for_id` with `albums_for_ids` / `tracks_for_ids` in `metadata_plugins` that yield candidates from all metadata sources
- (#6181) Use `Info.identifier` (`(data_source, id)`) as candidate keys to avoid cross-source ID collisions.
- Add tests (`test/autotag/test_match.py`) for assignment logic and multi-source ID matching
- Simplify `match_by_id`
- Dedupe `album_matched` event emission by moving it to `AlbumMatch.__post_init__` (and convert `AlbumMatch` / `TrackMatch` to dataclasses)

I am refactoring a couple of other things in `beets.autotag.match` module because this thing is a hot mess.

### DIFF
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 import abc
 import re
-from contextlib import contextmanager, nullcontext
+from contextlib import contextmanager
 from functools import cache, cached_property, wraps
 from typing import (
     TYPE_CHECKING,
@@ -27,7 +27,7 @@ from beets import config, logging
 from beets.util import cached_classproperty
 from beets.util.id_extractors import extract_release_id
 
-from .plugins import BeetsPlugin, find_plugins, notify_info_yielded
+from .plugins import BeetsPlugin, find_plugins, notify_info_yielded, send
 
 Ret = TypeVar("Ret")
 QueryType = Literal["album", "track"]
@@ -49,14 +49,27 @@ def find_metadata_source_plugins() -> list[MetadataSourcePlugin]:
     return [p for p in find_plugins() if hasattr(p, "data_source")]  # type: ignore[misc]
 
 
+@cache
+def get_metadata_source(name: str) -> MetadataSourcePlugin | None:
+    """Get metadata source plugin by name."""
+    name = name.lower()
+    plugins = find_metadata_source_plugins()
+    return next((p for p in plugins if p.data_source.lower() == name), None)
+
+
 @contextmanager
-def handle_plugin_error(plugin: MetadataSourcePlugin, method_name: str):
+def maybe_handle_plugin_error(plugin: MetadataSourcePlugin, method_name: str):
     """Safely call a plugin method, catching and logging exceptions."""
-    try:
+    if config["raise_on_error"]:
         yield
-    except Exception as e:
-        log.error("Error in '{}.{}': {}", plugin.data_source, method_name, e)
-        log.debug("Exception details:", exc_info=True)
+    else:
+        try:
+            yield
+        except Exception as e:
+            log.error(
+                "Error in '{}.{}': {}", plugin.data_source, method_name, e
+            )
+            log.debug("Exception details:", exc_info=True)
 
 
 def _yield_from_plugins(
@@ -68,11 +81,7 @@ def _yield_from_plugins(
     def wrapper(*args, **kwargs) -> Iterator[Ret]:
         for plugin in find_metadata_source_plugins():
             method = getattr(plugin, method_name)
-            with (
-                nullcontext()
-                if config["raise_on_error"]
-                else handle_plugin_error(plugin, method_name)
-            ):
+            with maybe_handle_plugin_error(plugin, method_name):
                 yield from filter(None, method(*args, **kwargs))
 
     return wrapper
@@ -102,12 +111,26 @@ def tracks_for_ids(*args, **kwargs) -> Iterator[TrackInfo]:
     yield from ()
 
 
-def album_for_id(_id: str) -> AlbumInfo | None:
-    return next(albums_for_ids([_id]), None)
+def album_for_id(_id: str, data_source: str) -> AlbumInfo | None:
+    """Get AlbumInfo object for the given ID and data source."""
+    if plugin := get_metadata_source(data_source):
+        with maybe_handle_plugin_error(plugin, "album_for_id"):
+            if info := plugin.album_for_id(_id):
+                send("albuminfo_received", info=info)
+                return info
+
+    return None
 
 
-def track_for_id(_id: str) -> TrackInfo | None:
-    return next(tracks_for_ids([_id]), None)
+def track_for_id(_id: str, data_source: str) -> TrackInfo | None:
+    """Get TrackInfo object for the given ID and data source."""
+    if plugin := get_metadata_source(data_source):
+        with maybe_handle_plugin_error(plugin, "track_for_id"):
+            if info := plugin.track_for_id(_id):
+                send("trackinfo_received", info=info)
+                return info
+
+    return None
 
 
 @cache
--- a/beetsplug/mbsync.py
+++ b/beetsplug/mbsync.py
@@ -72,17 +72,19 @@ class MBSyncPlugin(BeetsPlugin):
         query.
         """
         for item in lib.items([*query, "singleton:true"]):
-            if not item.mb_trackid:
+            if not (track_id := item.mb_trackid):
                 self._log.info(
                     "Skipping singleton with no mb_trackid: {}", item
                 )
                 continue
 
             if not (
-                track_info := metadata_plugins.track_for_id(item.mb_trackid)
+                track_info := metadata_plugins.track_for_id(
+                    track_id, item.get("data_source", "MusicBrainz")
+                )
             ):
                 self._log.info(
-                    "Recording ID not found: {0.mb_trackid} for track {0}", item
+                    "Recording ID not found: {} for track {}", track_id, item
                 )
                 continue
 
@@ -97,15 +99,20 @@ class MBSyncPlugin(BeetsPlugin):
         """
         # Process matching albums.
         for album in lib.albums(query):
-            if not album.mb_albumid:
+            if not (album_id := album.mb_albumid):
                 self._log.info("Skipping album with no mb_albumid: {}", album)
                 continue
 
+            data_source = album.get("data_source") or album.items()[0].get(
+                "data_source", "MusicBrainz"
+            )
             if not (
-                album_info := metadata_plugins.album_for_id(album.mb_albumid)
+                album_info := metadata_plugins.album_for_id(
+                    album_id, data_source
+                )
             ):
                 self._log.info(
-                    "Release ID {0.mb_albumid} not found for album {0}", album
+                    "Release ID {} not found for album {}", album_id, album
                 )
                 continue
 
--- a/beetsplug/missing.py
+++ b/beetsplug/missing.py
@@ -227,10 +227,15 @@ class MissingPlugin(MusicBrainzAPIMixin, BeetsPlugin):
         if len(album.items()) == album.albumtotal:
             return
 
-        item_mbids = {x.mb_trackid for x in album.items()}
         # fetch missing items
         # TODO: Implement caching that without breaking other stuff
-        if album_info := metadata_plugins.album_for_id(album.mb_albumid):
+        data_source = album.get("data_source") or album.items()[0].get(
+            "data_source", "MusicBrainz"
+        )
+        if album_info := metadata_plugins.album_for_id(
+            album.mb_albumid, data_source
+        ):
+            item_mbids = {x.mb_trackid for x in album.items()}
             for track_info in album_info.tracks:
                 if track_info.track_id not in item_mbids:
                     self._log.debug(
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -23,10 +23,15 @@ Bug fixes
 - :ref:`import-cmd` Autotagging by explicit release or recording IDs now keeps
   candidates from all enabled metadata sources instead of dropping matches when
   different providers share the same ID. :bug:`6178` :bug:`6181`
+- :doc:`plugins/mbsync` and :doc:`plugins/missing` now use each item's stored
+  ``data_source`` for ID lookups, with a fallback to ``MusicBrainz``.
 
-..
-    For plugin developers
-    ~~~~~~~~~~~~~~~~~~~~~
+For plugin developers
+~~~~~~~~~~~~~~~~~~~~~
+
+- :py:func:`beets.metadata_plugins.album_for_id` and
+  :py:func:`beets.metadata_plugins.track_for_id` now require a ``data_source``
+  argument and query only that provider.
 
 Other changes
 ~~~~~~~~~~~~~
--- a/docs/plugins/mbsync.rst
+++ b/docs/plugins/mbsync.rst
@@ -17,6 +17,9 @@ Enable the ``mbsync`` plugin in your configuration (see :ref:`using-plugins`)
 and then run ``beet mbsync QUERY`` to fetch updated metadata for a part of your
 collection (or omit the query to run over your whole library).
 
+ID lookups use each item's stored ``data_source``. If a row has no
+``data_source``, ``mbsync`` falls back to ``MusicBrainz``.
+
 This plugin treats albums and singletons (non-album tracks) separately. It first
 processes all matching singletons and then proceeds on to full albums. The same
 query is used to search for both kinds of entities.
--- a/docs/plugins/missing.rst
+++ b/docs/plugins/missing.rst
@@ -10,6 +10,8 @@ Usage
 
 The ``beet missing`` command fetches album information from the origin data
 source and lists names of the **tracks** that are missing from your library.
+Track-level checks use the album's stored ``data_source`` and fall back to
+``MusicBrainz`` when no source is stored.
 
 It can also list the names of missing **albums** for each artist, although this
 is limited to albums from the MusicBrainz data source only.
--- a/test/plugins/test_mbsync.py
+++ b/test/plugins/test_mbsync.py
@@ -45,10 +45,15 @@ class MbsyncCliTest(PluginTestCase):
             album="old album",
             mb_albumid="album id",
             mb_trackid="track id",
+            data_source="data_source",
         )
         self.lib.add_album([album_item])
 
-        singleton = Item(title="old title", mb_trackid="singleton id")
+        singleton = Item(
+            title="old title",
+            mb_trackid="singleton id",
+            data_source="data_source",
+        )
         self.lib.add(singleton)
 
         self.run_command("mbsync")
--- a/test/test_metadata_plugins.py
+++ b/test/test_metadata_plugins.py
@@ -32,6 +32,7 @@ class TestMetadataPluginsException(PluginMixin):
     @pytest.fixture(autouse=True)
     def setup(self):
         metadata_plugins.find_metadata_source_plugins.cache_clear()
+        metadata_plugins.get_metadata_source.cache_clear()
         self.register_plugin(ErrorMetadataMockPlugin)
         yield
         self.unload_plugins()
@@ -45,25 +46,23 @@ class TestMetadataPluginsException(PluginMixin):
         return _call
 
     @pytest.mark.parametrize(
-        "method_name,error_method_name,args",
+        "method_name,args",
         [
-            ("candidates", "candidates", ()),
-            ("item_candidates", "item_candidates", ()),
-            ("albums_for_ids", "albums_for_ids", (["some_id"],)),
-            ("tracks_for_ids", "tracks_for_ids", (["some_id"],)),
-            # Currently, singular methods call plural ones internally and log
-            # errors from there
-            ("album_for_id", "albums_for_ids", ("some_id",)),
-            ("track_for_id", "tracks_for_ids", ("some_id",)),
+            ("candidates", ()),
+            ("item_candidates", ()),
+            ("albums_for_ids", (["some_id"],)),
+            ("tracks_for_ids", (["some_id"],)),
+            ("album_for_id", ("some_id", "ErrorMetadataMock")),
+            ("track_for_id", ("some_id", "ErrorMetadataMock")),
         ],
     )
-    def test_logging(self, caplog, call_method, error_method_name):
+    def test_logging(self, caplog, call_method, method_name):
         self.config["raise_on_error"] = False
 
         call_method()
 
         assert (
-            f"Error in 'ErrorMetadataMock.{error_method_name}': Mocked error"
+            f"Error in 'ErrorMetadataMock.{method_name}': Mocked error"
             in caplog.text
         )
 
@@ -72,8 +71,10 @@ class TestMetadataPluginsException(PluginMixin):
         [
             ("candidates", ()),
             ("item_candidates", ()),
-            ("album_for_id", ("some_id",)),
-            ("track_for_id", ("some_id",)),
+            ("albums_for_ids", (["some_id"],)),
+            ("tracks_for_ids", (["some_id"],)),
+            ("album_for_id", ("some_id", "ErrorMetadataMock")),
+            ("track_for_id", ("some_id", "ErrorMetadataMock")),
         ],
     )
     def test_raising(self, call_method):


beets/autotag/hooks.py
```diff
@@ -214,16 +235,23 @@ def __init__(
...
-class AlbumMatch(NamedTuple):
-    distance: Distance
+@dataclass
+class AlbumMatch(Match):
     info: AlbumInfo
     mapping: dict[Item, TrackInfo]
     extra_items: list[Item]
     extra_tracks: list[TrackInfo]
 
+    def __post_init__(self) -> None:
```
semohr, line 286: I would move the event trigger out of the class constructor. It is possible that an Match object is constructed independent of a beets pipeline. I do not want to couple this that strongly if not necessary.

E.g. we have some serialization logic in beets-flask and I don't want to trigger this whenever I load an Match entry for cold storage.

beets/autotag/match.py
```diff
@@ -179,33 +177,31 @@ def _recommendation(
...
 def _sort_candidates(candidates: Iterable[AnyMatch]) -> Sequence[AnyMatch]:
     """Sort candidates by distance."""
     return sorted(candidates, key=lambda match: match.distance)
 
 
 def _add_candidate(
-    items: Sequence[Item],
-    results: dict[Any, AlbumMatch],
-    info: AlbumInfo,
+    items: Sequence[Item], results: Candidates[AlbumMatch], info: AlbumInfo
```
semohr: Can we try to keep the trailing comma inplace if it exists already. This makes the git diff easier to read, I was wondering for a moment what changed here.

beets/autotag/hooks.py
```diff
@@ -54,6 +57,16 @@ def __hash__(self) -> int:  # type: ignore[override]
...
     """Container for metadata about a musical entity."""
 
+    Identifier = tuple[str | None, str | None]
+
+    @property
+    def id(self) -> str | None:
+        raise NotImplementedError
+
+    @property
+    def identifier(self) -> Identifier:
```
semohr, line 71: Using a tuple as internal Identifier makes sense! Would have been my approach too.

beets/autotag/hooks.py
```diff
@@ -214,16 +235,23 @@ def __init__(
...
 # Structures that compose all the information for a candidate match.
+@dataclass
+class Match:
+    distance: Distance
+    info: Info
 
 
-class AlbumMatch(NamedTuple):
-    distance: Distance
+@dataclass
```
semohr, line 279: Do we really need dataclasses here? They introduce a bit of overhead compared to named tuples. We are not doing any instance checks and if we remove the post init I dont see the added value.

beets/autotag/hooks.py
```diff
@@ -54,6 +57,16 @@ def __hash__(self) -> int:  # type: ignore[override]
 class Info(AttrDict[Any]):
     """Container for metadata about a musical entity."""
 
+    Identifier = tuple[str | None, str | None]
+
+    @property
```
semohr, line 65: I do not like that we raise an NotImplementedError here. We should make the `Info` class abstract or a protocol if want to define a contract for the inheritance.

beets/metadata_plugins.py
```diff
@@ -48,30 +48,18 @@ def item_candidates(*args, **kwargs) -> Iterable[TrackInfo]:
         yield from plugin.item_candidates(*args, **kwargs)
 
 
-def album_for_id(_id: str) -> AlbumInfo | None:
-    """Get AlbumInfo object for the given ID string.
-
-    A single ID can yield just a single album, so we return the first match.
-    """
+@notify_info_yielded("albuminfo_received")
+def albums_for_ids(_id: str) -> Iterable[AlbumInfo]:
```
semohr: Should be named `albums_for_id` if we keep the current call signature. Btw. this seems like a breaking change to me. Can we do this without a major version increment?

beets/autotag/hooks.py
snejus: I considered using `ABC` + `@abstractmethod` here, but opted against it for these reasons:

1. Limited scope: `Info` only has 2 concrete subclasses - it's not a public plugin interface where we need strict enforcement at instantiation time.

2. Template pattern, not an interface: The base class provides real shared functionality (`identifier` property, `__repr__`, common `__init__` parameters). The `id` and `name` properties are just internal adapters mapping to different field names in subclasses (e.g., `album_id` vs `track_id`).

3. **Testing overhead**: Making it an ABC would require either creating stub implementations or monkeypatching `__abstractmethods__` in tests. Since we're not exposing `Info` for external extension, the ceremony doesn't add value.

The `NotImplementedError` approach clearly documents "subclasses must override this" without the ABC machinery. If we later expose `Info` for some plugin extendability, I'd absolutely convert it to ABC at that point.

Happy to reconsider if you feel strongly about it though.

beets/autotag/hooks.py
snejus: I get the concern about this side effect but moving this to `__post_init__` was intentional:

1. The `send("album_matched")` call was duplicated across multiple sites in the autotagger - always right after creating `AlbumMatch`
2. This _guarantess_ that this event is sent on every `AlbumMatch` creation. This is especially relevant given that I'm currently refactoring this functionality extensively.

This is actually a textbook use of `__post_init__` - PEP 557 explicitly recommends it for side effects that must always happen during initialization. The alternative (keeping manual emissions everywhere) was provably bug-prone.

**Re: beets-flask serialization**: `AlbumMatch`/`TrackMatch` are internal to beets' autotagger, not public plugin API. For deserialization, you could bypass `__post_init__` with:

```python
match = object.__new__(AlbumMatch)
match.__dict__.update(serialized_data)
```

Or even better, serialize just the match data rather than the objects themselves. If there's broader need for match serialization, we could discuss adding a proper public API for it. 

I want to avoid blocking necessary refactoring of beets' internals based on downstream usage of internal classes. Does the deserialization workaround work for your use case?

semohr: I still think the event should not be sent on every initialization. To me, the event is more closely coupled with the tagging logic than with the match object itself, although I do agree that the current approach makes the code a bit cleaner.

> AlbumMatch/TrackMatch are internal to beets' autotagger, not public plugin API. 

How does one identify public api in this case? We do not use `__all__` and there is no underscore in the naming. As there is no internal use indicator, `AlbumMatch/TrackMatch` are public api according to pep-8. 

Historically beets has treated only the plugin API as public, yes, but the project is also used as a library, and I think that use case deserves consideration as well. Without explicit boundaries, users reasonably assume that importable classes are fair game.

> Or even better, serialize just the match data rather than the objects themselves. If there's broader need for match serialization, we could discuss adding a proper public API for it.

Just to clarify: I meant only deserialization. This was what the "load from cold storage" comment was referencing.

> I want to avoid blocking necessary refactoring of beets' internals based on downstream usage of internal classes. Does the deserialization workaround work for your use case?

We routinely do block changes, or at least adjust them, because of potential downstream usage. Wanting to avoid that concern here feels a bit inconsistent.

The workaround would work, but it shifts the burden entirely onto downstream users and breaks existing programs.

---

I’m not opposed to the change in principle, and I don’t think it needs to block progress, but by any reasonable definition, this is a breaking change. In my view that implies either a major-version bump or, alternatively, introducing a minimal public deserialization format so maintainers can refactor freely without silently breaking consumers.

beets/autotag/hooks.py
semohr: > Template pattern, not an interface: The base class provides real shared functionality (identifier property, __repr__, common __init__ parameters). The id and name properties are just internal adapters mapping to different field names in subclasses (e.g., album_id vs track_id).

Abstract classes or protocols can also provide real shared functionality.

> Testing overhead: Making it an ABC would require either creating stub implementations or monkeypatching __abstractmethods__ in tests. Since we're not exposing Info for external extension, the ceremony doesn't add value.

We actually don’t construct raw Info instances in tests. Tests only ever instantiate AlbumInfo or TrackInfo directly. So converting Info to an ABC wouldn’t add test-work in practice.

> The NotImplementedError approach clearly documents "subclasses must override this" without the ABC machinery. If we later expose Info for some plugin extendability, I'd absolutely convert it to ABC at that point

It looks like this usage has already begun showing up in plugins (mbpseudo.py is one example). Given that the system is already being extended, it might be safer to formalize the interface now rather than later.

beets/autotag/hooks.py
snejus: Fair points - you're right about the API ambiguity and that beets is used as a library.

How about adding a `from_dict()` classmethod that skips the event?

```python
@classmethod
def from_dict(cls, data: dict) -> AlbumMatch:
    """Reconstruct from serialized data without emitting events."""
    obj = object.__new__(cls)
    obj.__dict__.update(data)
    return obj
```

This gives library users a stable deserialization path, keeps the `__post_init__` enforcement for the autotagger, and avoids the major version bump debate. Would that work for beets-flask?

beets/autotag/hooks.py
snejus: I think there's some confusion here - `mbpseudo.py` doesn't subclass `Info`. It's a metadata source plugin that returns `AlbumInfo`/`TrackInfo` instances, but it doesn't create new `Info` subclasses.

On the other hand, `discogs` does - however, that approach is being reverted in #6179 since this subclass unintentionally introduced flexible attributes that ended up written into the database.

Given the above (that they actually _should not_ be subclassed), I'd prefer to keep it simple with `NotImplementedError` for now. We can formalize it as an ABC if/when there's an actual need for plugin extensibility.

beets/metadata_plugins.py
snejus: I named it following the method names available on `MetadataSourcePlugin` for consistency. I'll update the parameter to accept a list of IDs!

> Btw. this seems like a breaking change to me. Can we do this without a major version increment?

As far as I'm aware, these functions are only used internally.

snejus: This is one of the reasons why I tend towards using `*args, **kwargs`...

snejus: Dug the removed functions back out from the grave - didn't realise that we _do_ use them - and added `data_source` parameter to both.

beets/metadata_plugins.py
```diff
@@ -214,9 +237,9 @@ def albums_for_ids(self, ids: Iterable[str]) -> Iterable[AlbumInfo | None]:
         single calls to album_for_id.
         """
 
-        return (self.album_for_id(id) for id in ids)
+        return filter(None, (self.album_for_id(id) for id in ids))
```
semohr: Please do not filter `None` values in this function. Returning `None` or `Info` was an intentionally decision to preserves the positional mapping between the input ids and the yielded results. Reconstructing the original order after filtering introduces unnecessary overhead for callers.

Same applies for the `tracks_for_ids` function.

beets/metadata_plugins.py
```diff
@@ -49,14 +49,27 @@ def find_metadata_source_plugins() -> list[MetadataSourcePlugin]:
     return [p for p in find_plugins() if hasattr(p, "data_source")]  # type: ignore[misc]
 
 
+@cache
```
semohr, line 52: I don’t think we need to create another cached function here. Instead, we could have `find_metadata_source_plugins` return a `dict[str, MetadataSourcePlugin]`, keyed by `data_source`.

This way, `find_metadata_source_plugins().get("spotify")` can quickly look up a plugin by name without introducing a separate cache.

If we want to keep the old `find_metadata_source_plugins` for deprecation purposes, we could invert the logic:

```python
@deprecate
def find_metadata_source_plugins():
    return new_function.values()
```

snejus: This would introduce inconsistencies (see `find_plugins`) and a need for deprecation for a limited use case in `track_for_id` and `album_for_id` that are used by two plugins only.

semohr: internally* ;) Again beets is used programmatically. Im using specifically this module quite a lot in private scripts and proper deprecations are always appreciated. Especially a function signature change, which is hard to catch.

I would always vote simplicity over consistency, if a cache is needed for lookup why not create it ourself. This is quite more code added for something that I would expect could be a simple `meta_plugins.get(data_source)`.

snejus: This is precisely why I'm introducing a new function instead of adjusting an existing one, keeping the existing callers for `find_metadata_source_plugins` unchanged.

beets/metadata_plugins.py
snejus: Actually, this change is not even required any more. This PR was created _before_ #5965 which introduced `filter(None, ...` in `_yield_from_plugins`.

beets/metadata_plugins.py
```diff
@@ -102,12 +111,26 @@ def tracks_for_ids(*args, **kwargs) -> Iterator[TrackInfo]:
...
+                return info
+
+    return None
+
 
+def track_for_id(_id: str, data_source: str) -> TrackInfo | None:
+    """Get TrackInfo object for the given ID and data source."""
+    if plugin := get_metadata_source(data_source):
+        with maybe_handle_plugin_error(plugin, "track_for_id"):
+            if info := plugin.track_for_id(_id):
```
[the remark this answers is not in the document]
snejus: All callers and tests updated mate

beets/metadata_plugins.py
```diff
@@ -49,14 +49,27 @@ def find_metadata_source_plugins() -> list[MetadataSourcePlugin]:
     return [p for p in find_plugins() if hasattr(p, "data_source")]  # type: ignore[misc]
 
 
+@cache
+def get_metadata_source(name: str) -> MetadataSourcePlugin | None:
+    """Get metadata source plugin by name."""
+    name = name.lower()
+    plugins = find_metadata_source_plugins()
+    return next((p for p in plugins if p.data_source.lower() == name), None)
```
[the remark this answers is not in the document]
snejus: Will do when there's a need for this.

beets/autotag/match.py
```diff
@@ -346,25 +339,22 @@ def tag_item(
...
-                candidates[info.track_id] = hooks.TrackMatch(dist, info)
-                # If this is a good match, then don't keep searching.
-                rec = _recommendation(_sort_candidates(candidates.values()))
-                if (
-                    rec == Recommendation.strong
-                    and not config["import"]["timid"]
-                ):
-                    log.debug("Track ID match.")
-                    return Proposal(_sort_candidates(candidates.values()), rec)
+        log.debug("Searching for track IDs: {}", ", ".join(trackids))
```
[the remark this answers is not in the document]
snejus: Guaranteed to be strings

beets/autotag/match.py
```diff
@@ -269,38 +267,35 @@ def tag_album(
...
 
     # Search by explicit ID.
     if search_ids:
-        for search_id in search_ids:
-            log.debug("Searching for album ID: {}", search_id)
-            if info := metadata_plugins.album_for_id(search_id):
-                _add_candidate(items, candidates, info)
-                if opt_candidate := candidates.get(info.album_id):
-                    plugins.send("album_matched", match=opt_candidate)
+        log.debug("Searching for album IDs: {}", ", ".join(search_ids))
```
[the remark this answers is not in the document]
snejus: Guaranteed to be strings.

beets/autotag/match.py
```diff
@@ -346,25 +339,22 @@ def tag_item(
...
-                    rec == Recommendation.strong
-                    and not config["import"]["timid"]
-                ):
-                    log.debug("Track ID match.")
-                    return Proposal(_sort_candidates(candidates.values()), rec)
+        log.debug("Searching for track IDs: {}", ", ".join(trackids))
+        for info in metadata_plugins.tracks_for_ids(trackids):
+            dist = track_distance(item, info, incl_artist=True)
+            candidates[info.identifier] = hooks.TrackMatch(dist, info)
+
```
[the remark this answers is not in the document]
snejus: `track_id` is essentially guaranteed.

beetsplug/mbsync.py
```diff
@@ -97,15 +105,24 @@ def albums(self, lib, query, move, pretend, write):
...
                 self._log.info("Skipping album with no mb_albumid: {}", album)
                 continue
 
             if not (
-                album_info := metadata_plugins.album_for_id(album.mb_albumid)
+                data_source := album.get("data_source")
+                or album.items()[0].get("data_source")
+            ):
+                self._log.info("Skipping album without data source: {}", album)
+                continue
```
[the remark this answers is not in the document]
snejus: It's guaranteed to be non-empty.

beets/autotag/match.py
```diff
@@ -180,33 +178,33 @@ def _recommendation(
...
     # Discard albums with zero tracks.
     if not info.tracks:
         log.debug("No tracks.")
         return
 
     # Prevent duplicates.
-    if info.album_id and info.album_id in results:
+    if info.album_id and info.identifier in results:
         log.debug("Duplicate.")
         return
```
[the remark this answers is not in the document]
snejus: `album_id` is essentially guaranteed to exist

beetsplug/missing.py
```diff
@@ -227,10 +227,17 @@ def _missing(self, album: Album) -> Iterator[Item]:
...
             return
 
-        item_mbids = {x.mb_trackid for x in album.items()}
         # fetch missing items
         # TODO: Implement caching that without breaking other stuff
-        if album_info := metadata_plugins.album_for_id(album.mb_albumid):
+        if (
+            data_source := album.get("data_source")
+            or album.items()[0].get("data_source")
+        ) and (
```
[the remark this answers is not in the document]
snejus: This is impossible

beetsplug/mbsync.py
```diff
@@ -97,15 +99,20 @@ def albums(self, lib, query, move, pretend, write):
...
                 continue
 
+            data_source = album.get("data_source") or album.items()[0].get(
+                "data_source", "MusicBrainz"
+            )
             if not (
-                album_info := metadata_plugins.album_for_id(album.mb_albumid)
+                album_info := metadata_plugins.album_for_id(
+                    album_id, data_source
+                )
```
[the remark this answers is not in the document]
snejus: I know it's a confusingly named field but `mb_albumid` contains ids from any data source.

beetsplug/missing.py
```diff
@@ -227,10 +227,15 @@ def _missing(self, album: Album) -> Iterator[Item]:
...
-        item_mbids = {x.mb_trackid for x in album.items()}
         # fetch missing items
         # TODO: Implement caching that without breaking other stuff
-        if album_info := metadata_plugins.album_for_id(album.mb_albumid):
+        data_source = album.get("data_source") or album.items()[0].get(
+            "data_source", "MusicBrainz"
+        )
+        if album_info := metadata_plugins.album_for_id(
+            album.mb_albumid, data_source
+        ):
```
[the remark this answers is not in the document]
snejus: I know it's a confusingly named field but `mb_albumid` contains ids from any data source.

beets/metadata_plugins.py
semohr: You are twisting my words here... 

I suggested inverting the logic which allows to drop a cache decorator and also skip some iterations in the by name lookup. 

Not that it matters much anymore since you merged this already but my suggestion was something like this:

```python
    """Return a list of all loaded metadata source plugins."""
    return list(get_metadata_sources.values())

@cache
def get_metadata_sources() -> dict[str, MetadataSourcePlugin]:
    """Get a dictionary of all metadata source plugins."""
    # TODO: Make this an isinstance(MetadataSourcePlugin, ...) check in v3.0.0
    # This should also allow us to remove the type: ignore comments below.
    return {p.name: p for p in p in find_plugins() if hasattr(p, "data_source")}
```

With the merged changes, if I want to reload plugins programmatically, I have to invalidate one more cache which should not be necessary in my opinion.

PR #6192, opened 2025-11-27
Remove gmusic plugin

semohr:
Sometimes it is time to let go of old things:
This PR removes the old gmusic plugin and all related docs. 

---

The google play music service was shutdown in 2020 and already deprecated in beets 1.6.0.


henry-oberholtzer:
Wow that had been sitting around for a while! Nice catch @semohr

PR #6197, opened 2025-12-01
Improves SubsonicUpdate error messages when server is unavailable #5635

### LINKED ISSUE #5635: subsonicupdate fails cryptically when server is not available
### Problem

Running this command:

```sh
$ beet -vv mbsync
```

Led to this problem:

```
subsonicupdate: URL is https://[redacted].com/rest/startScan
subsonicupdate: auth type is password
subsonicupdate: Error: Expecting value: line 1 column 1 (char 0)
```

This happened because my Subsonic server was down, so the issue is just that beets should print a better error message when this happens. 

### Setup

* OS: 
* Python version: 3.13.1
* beets version: 2.2.0

```yaml
(paste here)
```
I'll retry in a few days with a minimal config, I'm assuming `mbsync` needs a change to happen to actually run `subsonicupdate`.

jasling920:
## Description

Fixes #5635: subsonicupdate fails cryptically when server is not available

The plugin previously produced unhelpful JSON-parsing errors when the server was down or returned unexpected data. This PR improves the error handling so that these situations now produce clear, user-friendly messages instead of cryptic failures such as: 

> Error: Expecting value: line 1 column 1 (char 0)

This update adds explicit handling for:
- HTTP connection errors and timeouts 
- Invalid or non-JSON responses 
- Missing subsonic-response fields in returned data 

Mocked Subsonic responses confirm the improved behavior: 

- Normal response --> unchanged 
- Timeout/connection error --> now produce a clear and readable error 
- Invalid JSON --> now reported as “Invalid JSON from Subsonic” and shows the server’s returned text for context
- Missing subsonic-response --> now produces a clear message identifying that the required field is missing 

These changes ensure that subsonicupdate failures are now clearly reported and no longer cryptic. 
- [X] Documentation. (Only changelog was updated. No additional docs required.)
- [X] Changelog. (Added an entry to `docs/changelog.rst`.)
- [X] Tests. While no formal unit tests were added, the fix was verified using local mock servers simulating various Subsonic server conditions: 
- Normal response: Verified scanning proceeds normally with a valid server response (test_subsonic.py). 
- Timeout: Verified that server timeout produces a clear, user-friendly error message (test_subsonic_timeout.py). 
- Invalid JSON: Verified that invalid JSON produces a clear error (test_subsonic_invalid_json.py). 
- Missing subsonic-response field: Verified that a clear error message is shown when the expected field is missing (test_subsonic_missing_response.py).

### DIFF
--- a/beetsplug/subsonicupdate.py
+++ b/beetsplug/subsonicupdate.py
@@ -138,20 +138,29 @@ class SubsonicUpdate(BeetsPlugin):
                 params=payload,
                 timeout=10,
             )
-            json = response.json()
-
+            try:
+                json = response.json()
+            except ValueError:
+                self._log.error("Invalid JSON from Subsonic: {}", response.text[:200])
+                return
+            resp = json.get("subsonic-response")
+            if not resp:
+                self._log.error("Missing 'subsonic-response' field: {}", json)
+                return
+            status = resp.get("status")
             if (
                 response.status_code == 200
-                and json["subsonic-response"]["status"] == "ok"
+                and status == "ok"
             ):
-                count = json["subsonic-response"]["scanStatus"]["count"]
+                count = resp.get("scanStatus", {}).get("count", 0)
                 self._log.info("Updating Subsonic; scanning {} tracks", count)
             elif (
                 response.status_code == 200
-                and json["subsonic-response"]["status"] == "failed"
+                and status == "failed"
             ):
+                msg = resp.get("error", {}).get("message", "Unknown error")
                 self._log.error(
-                    "Error: {[subsonic-response][error][message]}", json
+                    "Error: {}", msg
                 )
             else:
                 self._log.error("Error: {}", json)
--- a/beetsplug/subsonicupdate.py
+++ b/beetsplug/subsonicupdate.py
@@ -138,6 +138,7 @@ class SubsonicUpdate(BeetsPlugin):
                 params=payload,
                 timeout=10,
             )
+            response.raise_for_status()
             try:
                 json = response.json()
             except ValueError:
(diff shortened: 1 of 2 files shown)


beetsplug/subsonicupdate.py
```diff
@@ -138,20 +138,29 @@ def start_scan(self):
...
                 timeout=10,
             )
-            json = response.json()
-
+            try:
+                json = response.json()
+            except ValueError:
+                self._log.error("Invalid JSON from Subsonic: {}", response.text[:200])
+                return
+            resp = json.get("subsonic-response")
```
semohr: The can be simplified to:
```python
if not (resp := json.get("subsonic-response")):
    self._log.error("Missing 'subsonic-response' field: {}", json)
    return
```

beetsplug/subsonicupdate.py
```diff
@@ -138,20 +138,29 @@ def start_scan(self):
                 params=payload,
                 timeout=10,
             )
-            json = response.json()
```
semohr, line 137: Typically we should do `response.raise_for_status()` here and handle errors properly in the try expect block below. Should also allow us to remove the `if ( response.status_code == 200)` conditional.

beetsplug/subsonicupdate.py
```diff
@@ -138,20 +138,29 @@ def start_scan(self):
...
+            try:
+                json = response.json()
+            except ValueError:
+                self._log.error("Invalid JSON from Subsonic: {}", response.text[:200])
+                return
+            resp = json.get("subsonic-response")
+            if not resp:
+                self._log.error("Missing 'subsonic-response' field: {}", json)
+                return
+            status = resp.get("status")
```
semohr, line 152: Do we have to be defensive here? The same applies to `resp.get("scanStatus", {}).get("count", 0)` below.

semohr (changes requested):
Thank you for the PR and initiative!

Since this seems to be your first contribution you might want to checkout our [contribution guide](https://beets.readthedocs.io/en/stable/contributing.html#getting-the-source). We are happy to have you onboard!

I’ve left a few comments that should help improve this PR.

beetsplug/subsonicupdate.py
jaslingm: Thank you. I have simplified it.

beetsplug/subsonicupdate.py
```diff
@@ -138,21 +138,24 @@ def start_scan(self):
...
                 )
+                return
+            if not (resp := json.get("subsonic-response")):
+                self._log.error("Missing 'subsonic-response' field: {}", json)
+                return
+            status = resp.get("status")
+            if status == "ok":
+                count = resp.get("scanStatus", {}).get("count", 0)
+                self._log.info("Updating Subsonic; scanning {} tracks", count)
+            elif status == "failed":
```
semohr, line 156: If we raise for status, `status == "failed"` can never be reached. Atleast I think this is the case or does subsonic not use HTTPErrorCodes?

snejus:
@jasling920 are you planning to come back to this by any chance?

PR #6200, opened 2025-12-03
Fix edit plugin cancel flow restoring in-memory tags (#6104)

### LINKED ISSUE #6104: Bad interaction: fromfilename and edit plugins
### Problem

When using both the `fromfilename` and `edit` plugin, aborted/canceled edit operations fail to restore the temporary tag data provided by the fromfilename plugin.

#### Steps to reproduce

1. Import some files lacking metadata tags, but containing useful information in the file name
2. The fromfilename plugin will create tags for track number, artist, title which will be used by the autotagger, e.g. for a search on musicbrainz
3. Select an album release candidate from the musicbrainz search results. In the "match preview" list, on the lefthand side you can see the files identified by the tags generated by the fromfilename plugin (and on the right side the suggested matches from the candidate release).
4. Choose "edit Candidates" and quit your text editor without making any changes
5. Select another candidate (even the same selected in step 3.): now on the lefthand side your files will be identified by the filename, no more info about the track number and titles: those tags generated by the fromfilename plugin are lost.

#### Notes

- The same happens if you choose "eDit" in step 4. The resulting text file will be correctly populated with tags from the fromfilename plugin, but if you quit the editor and go to step 5. above, the same bug occurs.
- The same happens if, instead of quitting the editor without changes, you actually make some changes, save the file, and select "Cancel" in the next menu.
- Other actions from the import menu don't cause the bug, i.e. I can do multiple searches on musicbrainz by artist/album or release id, and the fromfilename tags are always preserved
- If I don't use the fromfilename plugin and try to import files that already contain some tags, the bug doesn't happen. Tags coming from the file metadata are correctly remembered across aborted edit sessions, only tags coming from the fromfilename plugin are lost.
- "Mixed tags case": if I try to import files that contain tags for artist and track number, but non for title: the fromfilename plugin will generate title tags (equal to the track number, that's expected behaviour), and in step 5. the artist and track tags (coming from the file metadata) will be remembered, while the title tag generated by the fromfilename plugin will be lost.

#### Hints?

I haven't studied the `edit` plugin yet, but looking at the history, the bug might be related to 9898451 and/or surrounding commits dealing with correctly restoring objects in multiple edit sessions? Maybe the restoring logic should be fixed also in our "Cancel" case?

Or maybe the bug might be in the `fromfilename` plugin: should it create the temporary tags in a more "robust" way? I don't know much about beets internals to assess this. I can only observe the tags correctly survive to multiple musicbrainz searches (see third note above), they are only lost after aborted `edit` sessions.

By the way: the edit plugin documentation https://beets.readthedocs.io/en/stable/plugins/edit.html states:

> Also, please be aware that the edit Candidates choice can only be used with the
> matches found during the initial search (and currently not supporting the candidates
> found via the Enter search or enter Id choices). 

it seems this is not true anymore, I can do multiple searches and edit the candidates resulting from each search correctly, so I guess we should update the docs. 

### Setup

* OS: Linux
* Python version: 3.13.5
* beets version: 2.5.1

gabe-push:
## Description

Fixes [#6104](https://github.com/beetbox/beets/issues/6104).

When using the `fromfilename` and `edit` plugins together during import, aborted edit sessions could silently discard the temporary tags injected by `fromfilename` (e.g., track number and title derived from the filename). This happened when using `eDit` or `edit Candidates` and then cancelling: the edit plugin reverted objects by re-reading from disk, which does not contain the `fromfilename`-generated metadata.

This PR changes the `edit` plugin so that cancel and “continue Editing” both roll back objects to the original in-memory snapshot captured before opening the editor, instead of reloading from the files. This preserves temporary tags provided by other plugins (like `fromfilename`) across aborted edit sessions, while still only writing to disk when the user chooses Apply.

`importer_edit` is also updated to rely on this rollback behavior when edits are cancelled, rather than re-reading from disk, so interactive imports resume with the same in-memory metadata they started with.

## To Do

- [x] Documentation
- [x] Changelog (Entry added under the Unreleased “Bug fixes” section.)
- [x] Tests (Existing `edit` plugin tests exercised and extended to cover the updated cancel/rollback behavior.)

### DIFF
--- a/beetsplug/edit.py
+++ b/beetsplug/edit.py
@@ -282,16 +282,18 @@ class EditPlugin(plugins.BeetsPlugin):
                 if choice == "a":  # Apply.
                     return True
                 elif choice == "c":  # Cancel.
+                    # Revert all temporary changes made in this edit session
+                    # so that objects return to their original in-memory
+                    # state (including tags provided by other plugins such as
+                    # `fromfilename`).
+                    self.apply_data(objs, new_data, old_data)
                     return False
                 elif choice == "e":  # Keep editing.
-                    # Reset the temporary changes to the objects. I we have a
-                    # copy from above, use that, else reload from the database.
-                    objs = [
-                        (old_obj or obj) for old_obj, obj in zip(objs_old, objs)
-                    ]
-                    for obj in objs:
-                        if not obj.id < 0:
-                            obj.load()
+                    # Revert changes on the objects, but keep the edited YAML
+                    # file so the user can continue editing from their last
+                    # version. On the next iteration, differences will again
+                    # be computed against the original state (`old_data`).
+                    self.apply_data(objs, new_data, old_data)
                     continue
 
         # Remove the temporary file before returning.
@@ -380,9 +382,11 @@ class EditPlugin(plugins.BeetsPlugin):
             # to the files if needed without re-applying metadata.
             return Action.RETAG
         else:
-            # Edit cancelled / no edits made. Revert changes.
-            for obj in task.items:
-                obj.read()
+            # Edit cancelled / no edits made. `edit_objects` has already
+            # restored each object to its original in-memory state, so there
+            # is nothing more to do here. Returning None lets the importer
+            # resume the candidate prompt.
+            return None
 
     def importer_edit_candidate(self, session, task):
         """Callback for invoking the functionality during an interactive
--- a/beetsplug/edit.py
+++ b/beetsplug/edit.py
@@ -275,24 +275,17 @@ class EditPlugin(plugins.BeetsPlugin):
                     ui.print_("No changes to apply.")
                     return False
 
-                # Confirm the changes.
+                # For cancel/keep-editing, restore objects to their original
+                # in-memory state so temp edits don't leak into the session
                 choice = ui.input_options(
                     ("continue Editing", "apply", "cancel")
                 )
                 if choice == "a":  # Apply.
                     return True
                 elif choice == "c":  # Cancel.
-                    # Revert all temporary changes made in this edit session
-                    # so that objects return to their original in-memory
-                    # state (including tags provided by other plugins such as
-                    # `fromfilename`).
                     self.apply_data(objs, new_data, old_data)
                     return False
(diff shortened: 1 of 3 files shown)


semohr (changes requested):
Thanks for the PR. Finally had some time to have a look here. 

Could you simplify the comments a bit they feel quite verbose. Also could you please add a test for the `["e","c"]` case?

Other than that this looks like a quite elegant solution to me.

JOJ0:
> Thanks for the PR. Finally had some time to have a look here.
> 
> Could you simplify the comments a bit they feel quite verbose. Also could you please add a test for the `["e","c"]` case?
> 
> Other than that this looks like a quite elegant solution to me.

Hi @gabe-push I also had a look here yesterday and thought the same. Even though the comments are well written and useful they could be a little shorter, or, as an alternative idea for documenting what's happening, you could try to to describe everything in one comment block per _section_. Keeping code readable / together and putting one larger block above it to describe more briefly what the flow of that section is. Or maybe sometimes the whole flow of a function can be briefly described in the docstring (that strategy won't work well here I suppose, just mentioning)

This plugin had an extensive amount of comments already, I just realized now. We won't be mad if you even try to shorten some of the existing one's if it fits your with your new descriptions better!

Hope that helps! 

And by the way: I love this feature / this fix! Very nice! Thanks!

gabe-push:
Hello @semohr and @JOJ0 , thank you both for the helpful feedback! I agree and apologize for the very verbose comments - definitely don't want to negatively effect code readability and maintainability in any way. I've just updated the comments and added a test for the `["e","c"]` case. Please let me know if there's anything missing! Thank you

semohr (approved):
Looks good from my side. Thanks again for the PR!

JOJ0:
Hi @gabe-push, thanks for the comments changes! Great! Please resolve the changelog conflict, I've set the PR to auto-merge once that's done. Many thanks again for the contribution! See you next time!

semohr:
Since the PR was sitting for a bit I resolved the issues myself before we start to forget about this :upside_down_face:

PR #6201, opened 2025-12-03
Duplicating Artist if Featured

### LINKED ISSUE #6166: Discogs: Featured artists appear twice.
On the current implementation of the discogs plugin, for supporting featured artists, the plugin does not account for cases where an artist may be credited as featured in the artist name, and also in the track credits.

When importing [this album:](https://www.discogs.com/release/1734236-Filteria-Daze-Of-Our-Lives), track 6 ends up with duplicate credits. Discogs current has the track credited as follows.
```
6 - Filteria Feat. Ukiro - Infinite Regression
    Written-By, Producer, Featuring – Ukiro
```
On import, it results in the following:
```
Filteria Feat. Ukiro Feat. Ukiro - Infinite Regression
```

The provided Discogs information may be redundant, but there doesn't seem to be a concise guideline on whether it's a database violation either, so I think we will have to figure out a way to handle it.

gcoppola19:
## Description

Fixes #6166   

(...)

## Checking if an Artist is already present in featured, and not adding it from Discogs plug in

I didn't add chaneglog or documentation, and can do so, but wanted to make sure this was the intended usage and implementation. 
- [ ] Documentation. 
- [ x] Tests. (Very much encouraged but not strictly required.)

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -95,12 +95,18 @@ def in_no_convert(item: Item) -> bool:
         return False
 
 
-def should_transcode(item, fmt):
+def should_transcode(item, fmt, force: bool = False):
     """Determine whether the item should be transcoded as part of
     conversion (i.e., its bitrate is high or it has the wrong format).
+
+    If ``force`` is True, safety checks like ``no_convert`` and
+    ``never_convert_lossy_files`` are ignored and the item is always
+    transcoded.
     """
+    if force:
+        return True
     if in_no_convert(item) or (
-        config["convert"]["never_convert_lossy_files"]
+        config["convert"]["never_convert_lossy_files"].get(bool)
         and item.format.lower() not in LOSSLESS_FORMATS
     ):
         return False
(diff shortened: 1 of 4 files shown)


beetsplug/discogs.py
```diff
@@ -780,10 +780,21 @@ def get_track_info(
...
-                artist += f" {self.config['featured_string']} {featured}"
-                artist_credit += (
-                    f" {self.config['featured_string']} {featured_credit}"
-                )
+                featured_string = self.config["featured_string"].as_str()
+                token = f"{featured_string} {featured}".lower()
+                token_credit = f"{featured_string} {featured_credit}".lower()
+
+                # Only append if this featured artist isn't already present
+                if token not in artist.lower():
```
jackwilsdon: Is there something we can do other than a substring match? This is going to avoid appending the featured artist if the main artist contains a subset of the featured artist's name (e.g. `ABCD feat. D`).

gcoppola19: I updated the logic to avoid relying on a substring match, and the new approach normalizes both the main artist and featured artist into lists and compares them, ensuring we only skip appending a featured artist when it is already present. I also added a test case for the example you provided to test it. Would this be a better implementation?

Serene-Arc:
Hi, couple things:

- Why was a test file deleted?
- For anything regarding string parsing, I'd want a ton of tests for edge cases.

snejus:

Issue #5783, opened 2025-05-14
Upgrade poetry to v2


snejus:
### Problem

We attempted to upgrade to poetry v2 however we found this caused issues: https://github.com/beetbox/beets/issues/5770. See the linked issue for more context about the problem.

For now, [we downgraded to '<2'](https://github.com/beetbox/beets/pull/5782) however we ultimately want to be able to upgrade. 

Unfortunately, `poetry` won't make this an easier task because they don't seem to see an issue here: https://github.com/python-poetry/poetry/issues/10083

amogus07:
How about migrating to [uv](https://github.com/astral-sh/uv)? It‘s seriously fast and seems like a solid replacement that eliminates the need for many other tools

wisp3rwind:
> How about migrating to [uv](https://github.com/astral-sh/uv)? It‘s seriously fast and seems like a solid replacement that eliminates the need for many other tools

Related: https://github.com/beetbox/beets/pull/5728

That said, it's conceivable that `uv` does the same as `poetry` here, for the same reason: Reproducible builds.

We might consider defaulting to `uv`, but I think working around this issue should not be the driver: Note also my comment above.

gbcox:
Fedora is upgrading to Poetry 2 for F43, while F42 remains on 1.8.5. My understanding was that Poetry wasn't upgraded in F42 due to breakage caused by poetry-core>=2 that affects pyproject.toml files, which the maintainers preferred not to introduce to F42.

I want to stress that my intention isn't to disparage the Poetry project; it might very well be the best fit for this scenario. However, I found it curious that another package, yubikey-manager, moved to Poetry a few years ago, and then to a 'Poetry 2' setup within the last few months. Now, they've switched to UV going forward.

Of course, beets isn't yubikey-manager. But if you're on the fence about a decision, it might behoove you to check with the yubikey-manager project. Their recent experience, navigating these rapid changes in the Python packaging ecosystem, could offer valuable insights.

sakgoyal:
I would also vote to migrate to `uv`. It removes the need for an additional dependency as you can specify dependencies in the standard python format instead of needing to use poetry dependencies.

amogus07:
@sakgoyal doesn't poetry support the standard format too? Also, [fyn](https://github.com/duriantaco/fyn) is another option, now that Astral (the company behind uv) was bought by OpenAI. It also has real-world benefits compared to uv like a task runner (which would replace poethepoet I suppose) as well as a `shell` command.

[...]

PR #6204, opened 2025-12-06
Update poetry command

Serene-Arc:
I was getting my development environment set up so I could start contributing code again (yay!) when I noticed that the docs are out of date. `poetry shell` is no longer part of the core poetry tool, but a plugin. Using `poetry env activate` is given as a replacement. This just updates the docs to reflect that. I also added a shortcut so people could understand how to install all the extras, since that's pretty necessary for developers.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -124,7 +124,7 @@ command. Instead, you can activate the virtual environment in your shell with:
 
     $ poetry env activate
 
-You should see ``(beets-py3.10)`` prefix in your shell prompt. Now you can run
+Now you can run
 commands directly, for example:
 
 ::


CONTRIBUTING.rst
```diff
@@ -122,7 +122,7 @@ command. Instead, you can activate the virtual environment in your shell with:
 
 ::
 
-    $ poetry shell
+    $ poetry env activate
 
 You should see ``(beets-py3.10)`` prefix in your shell prompt. Now you can run
```
JOJ0:
```suggestion
Now you can run
```

CONTRIBUTING.rst
JOJ0, line 132:
```suggestion
    $ pytest
```

JOJ0 (changes requested):
Welcome back again @Serene-Arc 👋 👍 

sorcery-ai has a point. we mention that the prompt changes which doesn't with `poetry env activate`. 

there is other commands for showing that info ar adjusting one's prompt manuyll `poetry env info` for example. We could add that or we just link to the poetry docs which I find better actually. (we might do that already, didn't check)

henry-oberholtzer:
I think before we'd change this, we would want to look into bumping our poetry version above v2. We're still specifying [poetry <2.0](https://github.com/beetbox/beets/blob/2bd77b9895fa9779818ff1c0430b9f9738d1616b/pyproject.toml#L188C1-L190C19) which means that the `poetry shell` command would [still be the recommended way to set up the environment.](https://python-poetry.org/docs/1.8/basic-usage/#activating-the-virtual-environment)

snejus:
> I think before we'd change this, we would want to look into bumping our poetry version above v2. We're still specifying [poetry <2.0](https://github.com/beetbox/beets/blob/2bd77b9895fa9779818ff1c0430b9f9738d1616b/pyproject.toml#L188C1-L190C19) which means that the `poetry shell` command would [still be the recommended way to set up the environment.](https://python-poetry.org/docs/1.8/basic-usage/#activating-the-virtual-environment)

Indeed, @Serene-Arc see https://github.com/beetbox/beets/issues/5783 for the context.

Serene-Arc:
@snejus @wisp3rwind perhaps it's time to try a switch to `uv`, or perhaps implementing a fix to the issue in #5783 at least? Is there any particular reason it's not moved forwards at all? I would prefer using it over poetry and pipenv and I'm willing to investigate the integration (and this isn't a real reason, but I do so enjoy supporting rust tools).

semohr:
> @snejus @wisp3rwind perhaps it's time to try a switch to `uv`, or perhaps implementing a fix to the issue in #5783 at least? Is there any particular reason it's not moved forwards at all? I would prefer using it over poetry and pipenv and I'm willing to investigate the integration (and this isn't a real reason, but I do so enjoy supporting rust tools).

Tbh, I’d also prefer using `uv`, but there are a few arguments against adopting it right now, mainly that it would further segregate our development env across the various subpackages. Still, if you want to evaluating a uv-based setup for beets, you can take a look at https://github.com/beetbox/mediafile/pull/87, which replaces `poetry` & `pipx` with `uv` in the `mediafile` package. It may serve as a useful reference point.

snejus:
The reason why we pinned poetry is here: https://github.com/beetbox/beets/issues/5770

snejus:
Closing as the change is not relevant for the current poetry version that we use.

Issue #5668, opened 2025-03-12
`beet modify` writes to file when no writeable tags changed


loafofpiecrust:
### Problem

Running this command:

```sh
$ beet modify -a onplayer=true queen
```

Led to this problem:

- Media files got updated mtime and the command took 25s, telling me it wrote to the files even though no writable media tags were changed. If I run with `-W` instead, the command takes <1s. 
- As far as I can tell, all custom tags are filtered out in `library.py` when writing to media files, so `beet modify` can avoid writing entirely in this case. I modify this tag for example often to change what goes on my iPod and always use `-W` but would rather not have to worry about having the argument.

### Setup

* OS: NixOS
* Python version: 3.12.8
* beets version: 2.1.0
* Turning off plugins made problem go away (yes/no): no

My configuration (output of `beet config`) is:

```yaml
directory: /mnt/music
library: ~/music/beets-library.db
original_date: yes
format_album: $albumartist/[$year] $album ($release_year, $country, $label, $media)
format_item: $artist - $album - $disc.$track. $title
sort_album: albumartist+ year+ album+
per_disc_numbering: yes
import:
  detail: yes
  autotag: yes
  copy: no
  timid: yes
  languages: en
  incremental: yes
  incremental_skip_later: no
  write: yes
match:
  preferred:
    countries: ["US", "GB|UK", "JP"]
    media: ["CD", "DigitalMedia|File", "Vinyl"]
    original_year: yes
musicbrainz:
  extra_tags: [year, catalognum, country, media]
paths:
  albumtype:soundtrack: "0. Soundtracks/[$realyear] $album/%if{$multidisc,Disc %right{$disc,1}/}$track. $title"
  albumtype:single: %the{$albumartist}/Singles/[$realyear]%if{$subartist,[$subartist]} $album/$track. $title
  comp: "0. %first{$genre}/[$realyear]%if{$subartist,[$subartist]} $album%ifdef{atypes, $atypes}/%if{$multidisc,%if{$grouping,$grouplabel,$disclabel}/,%if{$grouping,$grouping/}}$track. $title"
  default: %the{$albumartist}/[$realyear]%if{$subartist,[$subartist]} $album%ifdef{atypes, $atypes}/%if{$multidisc,%if{$grouping,$grouplabel,$disclabel}/,%if{$grouping,$grouping/}}$track. $title
item_fields:
  mono: 1 if channels == 1 else 0
  multidisc: 1 if disctotal > 1 else 0
[...]

PR #6208, opened 2025-12-08
Avoid unnecessary writes when modifying non-media fields

weiqianwang123:
Summary

Some beets commands such as:

beet modify -a onplayer=true queen

update only custom fields that are not mapped to media tags.
However, the current implementation still performs MediaFile.save(), causing:

unnecessary file writes

updated mtimes

20–30 second delays on large libraries

slow iPod sync / alternative device workflows

Root Cause

Item.write() filters _media_fields, but it always writes even if:

none of the media-tag fields changed

only flexible/custom database fields changed

Fix

Before writing, compute whether any media-tag-backed field changed.
If no media tag differences exist, skip the write entirely.

This preserves existing behavior for:

imports

replaygain

zero / scrub

metadata updates

Testing

A 0.5s sine-wave mp3 was generated via ffmpeg:

ffmpeg -f lavfi -i sine=frequency=440:duration=0.5 tiny.mp3
beet import tiny.mp3

Then:

# Should NOT write to file
beet modify -a onplayer=true tiny

# Should write to file
beet modify title="new title" tiny

Screenshots attached.

Performance improvement

Before: 25 seconds
After: <1 second
(repeated 200× batch modify)

DCO

All commits signed-off.

Related Issue: #5668

### DIFF
--- a/beets/library/models.py
+++ b/beets/library/models.py
@@ -995,6 +995,12 @@ class Item(LibModel):
             log.error("{}", exc)
             return False
 
+    def _media_tags_changed(self, original, modified):
+        for f in self._media_tag_fields:   # this is defined in the class already
+            if getattr(original, f) != getattr(modified, f):
+                return True
+        return False
+
     def try_sync(self, write, move, with_album=True):
         """Synchronize the item with the database and, possibly, update its
         tags on disk and its path (by moving the file).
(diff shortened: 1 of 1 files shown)


henry-oberholtzer:
Hi! Thanks for the PR. It looks like the current approach is causing side effects that break the current test suite, possibly due to an indentation error on this line. A test for a change like this would also be good, especially since it's dealing with core code, to make sure we're not making the performance issues worse.

snejus:
Thanks for working on this @weiqianwang123. I'm closing the PR for now since the current approach calls `Item.from_path` unconditionally, which undermines the goal when `write` is off. 

Please feel free to open a new PR that avoids constructing the Item in the no-write case.

PR #5644, opened 2025-02-23
Feat: Add replace plugin

Uncorrupt3318:
## Description

Adds replace plugin. The plugin allows the user to replace the audio file of a song, while keeping the tags and file name. 

Some music servers keep track of favourite songs via paths and tags. Now there won't be a need to 'refavourite'. Plus, this skips the import/merge steps.

My first contribution to any project and first python project. Be gentle haha

Also adds shell.nix for easier development on NixOS.

## To Do

- [x] Documentation.
- [x] Changelog.
- [x] Tests.

PR #6215, opened 2025-12-12
Fixes to Replace Plugin

### LINKED ISSUE #6260: Replace plugin does not work
### Problem

```sh
$ beet -vv replace id:42 /
user configuration: /Users/agross/.config/beets/config.yaml
data directory: /Users/agross/.config/beets
plugin paths:
inline: adding item field albumartist_no_va
inline: adding item field maybe_disc
inline: adding item field maybe_soundtrack
fetchart: lastfm: Disabling art source due to missing key
artresizer: method is ImageMagick
Sending event: pluginload
library database: /Users/agross/.config/beets/library.db
library directory: /Users/agross/Nextcloud/Music
Sending event: library_opened
loadext: loading extension /Users/agross/.config/beets/libicu
  File "/opt/homebrew/bin/beet", line 7, in <module>
    sys.exit(main())
             ~~~~^^
  File "/opt/homebrew/lib/python3.13/site-packages/beets/ui/__init__.py", line 1865, in main
    _raw_main(args)
    ~~~~~~~~~^^^^^^
  File "/opt/homebrew/lib/python3.13/site-packages/beets/ui/__init__.py", line 1852, in _raw_main
TypeError: ReplacePlugin.run() takes 3 positional arguments but 4 were given
```

### Setup

* OS: macOS
* Python version: 3.13
* beets version: beets version 2.5.1
* Turning off plugins made problem go away (yes/no): The problem is related to a plugin

My configuration (output of `beet config`) is: irrelevant

willburden:
## Description

This is my first contribution here, hope it makes sense!

When running the [Replace Plugin](https://beets.readthedocs.io/en/v2.5.1/plugins/replace.html) it fails due to the plugin's callback method having the wrong signature.

```
➜ ~ beet replace bowie changes ~/Downloads/changes.flac
  File "/home/will/.local/bin/beet", line 7, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/will/.local/share/pipx/venvs/beets/lib64/python3.13/site-packages/beets/ui/__init__.py", line 1713, in main
    _raw_main(args)
    ~~~~~~~~~^^^^^^
  File "/home/will/.local/share/pipx/venvs/beets/lib64/python3.13/site-packages/beets/ui/__init__.py", line 1692, in _raw_main
TypeError: ReplacePlugin.run() takes 3 positional arguments but 4 were given
```

On fixing this, I noticed that when replacing a file, the tags in the database are kept intact but are not written to the newly swapped-in file's metadata.

```
$ beet ls hunky dory changes
David Bowie - Hunky Dory - Changes

$ beet replace hunky dory changes ~/Downloads/changes_no_metadata.mp3 
Matching songs:
1. David Bowie - Hunky Dory - Changes
Which song would you like to replace? [1-1] (0 to cancel): 1

Replacing: /home/will/Downloads/changes_no_metadata.mp3 -> /home/will/Music/library/shared/David Bowie/Hunky Dory/01 Changes.mp3
Are you sure you want to replace this track? (y/N): y
Replacement successful.

$ beet ls hunky dory changes
David Bowie - Hunky Dory - Changes   (the database still has the tags)

$ beet write -p hunky dory changes   (but the file doesn't, so the user needs to run beet write)
 -  - 
  title:  -> Changes
  artist:  -> David Bowie
  artists:  -> David Bowie
  artist_sort:  -> David Bowie
  [...]
```

So I've updated it to call `Item.write()` immediately after replacing. To me this is a more intuitive behaviour but if it's preferred that the user should have to manually run `beet write`, I'm happy to undo this second change and just update the docs to reflect that.

I've written a test for the replacement behaviour.

## To Do

- [x] Decide if automatically writing the metadata to the new file is okay.
- [x] Documentation.
- [X] Tests.
- [x] Changelog.

### DIFF
--- a/beetsplug/replace.py
+++ b/beetsplug/replace.py
@@ -23,7 +23,9 @@ class ReplacePlugin(BeetsPlugin):
         cmd.func = self.run
         return [cmd]
 
-    def run(self, lib: Library, opts: optparse.Values, args: list[str]) -> None:
+    def run(
+        self, lib: Library, _opts: optparse.Values, args: list[str]
+    ) -> None:
         if len(args) < 2:
             raise ui.UserError("Usage: beet replace <query> <new_file_path>")
 
@@ -61,7 +63,7 @@ class ReplacePlugin(BeetsPlugin):
         except mediafile.FileTypeError as fte:
             raise ui.UserError(fte)
 
-    def select_song(self, items: list[Item]):
+    def select_song(self, items: list[Item]) -> Item | None:
         """Present a menu of matching songs and get user selection."""
         ui.print_("Matching songs:")
         for i, item in enumerate(items, 1):
@@ -86,7 +88,7 @@ class ReplacePlugin(BeetsPlugin):
             except ValueError:
                 ui.print_("Invalid input. Please type in a number.")
 
-    def confirm_replacement(self, new_file_path: Path, song: Item):
+    def confirm_replacement(self, new_file_path: Path, song: Item) -> bool:
         """Get user confirmation for the replacement."""
         original_file_path: Path = Path(song.path.decode())
 
--- a/beetsplug/replace.py
+++ b/beetsplug/replace.py
@@ -101,12 +101,8 @@ class ReplacePlugin(BeetsPlugin):
             f"\nReplacing: {util.displayable_path(new_file_path)} "
             f"-> {util.displayable_path(original_file_path)}"
         )
-        decision: str = (
-            input("Are you sure you want to replace this track? (y/N): ")
-            .strip()
-            .casefold()
-        )
-        return decision in {"yes", "y"}
+
+        return ui.input_yn("Are you sure you want to replace this track (y/n)?")
 
     def replace_file(self, new_file_path: Path, song: Item) -> None:
         """Replace the existing file with the new one."""
@@ -128,7 +124,7 @@ class ReplacePlugin(BeetsPlugin):
                 raise ui.UserError(f"Could not delete original file: {e}")
 
         # Update the path to point to the new file.
-        song.path = str(dest).encode()
+        song.path = util.bytestring_path(dest)
         song.store()
 
         # Write the metadata in the database to the song file's tags.
--- a/beetsplug/replace.py
+++ b/beetsplug/replace.py
@@ -71,24 +71,19 @@ class ReplacePlugin(BeetsPlugin):
         for i, item in enumerate(items, 1):
             ui.print_(f"{i}. {util.displayable_path(item)}")
 
-        while True:
-            try:
-                index = int(
-                    input(
-                        f"Which song would you like to replace? "
-                        f"[1-{len(items)}] (0 to cancel): "
-                    )
-                )
-                if index == 0:
-                    return None
-                if 1 <= index <= len(items):
-                    return items[index - 1]
-                ui.print_(
-                    f"Invalid choice. Please enter a number "
-                    f"between 1 and {len(items)}."
-                )
-            except ValueError:
-                ui.print_("Invalid input. Please type in a number.")
+        index = ui.input_options(
+            [],
+            require=True,
+            prompt=(
+                f"Which song would you like to replace? "
+                f"[1-{len(items)}] (0 to cancel):"
+            ),
+            numrange=(0, len(items)),
+        )
+
+        if index == 0:
+            return None
+        return items[index - 1]
 
     def confirm_replacement(self, new_file_path: Path, song: Item) -> bool:
         """Get user confirmation for the replacement."""
--- a/beetsplug/replace.py
+++ b/beetsplug/replace.py
@@ -7,6 +7,8 @@ from typing import TYPE_CHECKING
 import mediafile
 
 from beets import ui, util
+from beets.library import Item, Library
+from beets.library.exceptions import FileOperationError
 from beets.plugins import BeetsPlugin
 
 if TYPE_CHECKING:
@@ -113,7 +115,7 @@ class ReplacePlugin(BeetsPlugin):
 
         try:
             shutil.move(util.syspath(new_file_path), util.syspath(dest))
-        except Exception as e:
+        except OSError as e:
             raise ui.UserError(f"Error replacing file: {e}")
 
         if (
@@ -122,7 +124,7 @@ class ReplacePlugin(BeetsPlugin):
         ):
             try:
                 original_file_path.unlink()
-            except Exception as e:
+            except OSError as e:
                 raise ui.UserError(f"Could not delete original file: {e}")
 
         # Update the path to point to the new file.
@@ -131,7 +133,7 @@ class ReplacePlugin(BeetsPlugin):
         # Write the metadata in the database to the song file's tags.
         try:
             song.write()
-        except Exception as e:
+        except FileOperationError as e:
             raise ui.UserError(f"Error writing metadata to file: {e}")
 
         # Commit the new path to the database.
--- a/beetsplug/replace.py
+++ b/beetsplug/replace.py
@@ -123,7 +123,14 @@ class ReplacePlugin(BeetsPlugin):
             except Exception as e:
                 raise ui.UserError(f"Could not delete original file: {e}")
 
+        # Store the new path in the database.
         song.path = str(dest).encode()
         song.store()
+
+        # Write the metadata in the database to the song file's tags.
+        try:
+            song.write()
+        except Exception as e:
+            raise ui.UserError(f"Error writing metadata to file: {e}")
 
         ui.print_("Replacement successful.")
--- a/beetsplug/replace.py
+++ b/beetsplug/replace.py
@@ -10,6 +10,8 @@ from beets import ui, util
 from beets.plugins import BeetsPlugin
 
 if TYPE_CHECKING:
+    import optparse
+
     from beets.library import Item, Library
 
 
@@ -21,7 +23,7 @@ class ReplacePlugin(BeetsPlugin):
         cmd.func = self.run
         return [cmd]
 
-    def run(self, lib: Library, args: list[str]) -> None:
+    def run(self, lib: Library, opts: optparse.Values, args: list[str]) -> None:
         if len(args) < 2:
             raise ui.UserError("Usage: beet replace <query> <new_file_path>")
 
@@ -61,7 +63,7 @@ class ReplacePlugin(BeetsPlugin):
 
     def select_song(self, items: list[Item]):
         """Present a menu of matching songs and get user selection."""
-        ui.print_("\nMatching songs:")
+        ui.print_("Matching songs:")
         for i, item in enumerate(items, 1):
             ui.print_(f"{i}. {util.displayable_path(item)}")
 
--- a/beetsplug/replace.py
+++ b/beetsplug/replace.py
@@ -102,7 +102,9 @@ class ReplacePlugin(BeetsPlugin):
             f"-> {util.displayable_path(original_file_path)}"
         )
 
-        return ui.input_yn("Are you sure you want to replace this track (y/n)?")
+        return ui.input_yn(
+            "Are you sure you want to replace this track (y/n)?", require=True
+        )
 
     def replace_file(self, new_file_path: Path, song: Item) -> None:
         """Replace the existing file with the new one."""
--- a/beetsplug/replace.py
+++ b/beetsplug/replace.py
@@ -129,14 +129,12 @@ class ReplacePlugin(BeetsPlugin):
 
         # Update the path to point to the new file.
         song.path = str(dest).encode()
+        song.store()
 
         # Write the metadata in the database to the song file's tags.
         try:
             song.write()
         except FileOperationError as e:
             raise ui.UserError(f"Error writing metadata to file: {e}")
-
-        # Commit the new path to the database.
-        song.store()
 
         ui.print_("Replacement successful.")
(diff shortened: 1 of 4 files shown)


test/plugins/test_replace.py
```diff
@@ -5,46 +5,54 @@
...
 from beets import ui
+from beets.library import Item, Library
 from beets.test import _common
+from beets.test.helper import TestHelper
 from beetsplug.replace import ReplacePlugin
 
 replace = ReplacePlugin()
 
 
 class TestReplace:
```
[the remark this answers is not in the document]
willburden: I've added a pair of tests similar to the suggestion.

willburden:
- Refined the exception types being caught.
- Improved test coverage for the run function itself, which should catch a regression of the initial bug.
- Improved test coverage for replace.replace_file, covering IO error paths.
- Sourcery suggested saving the path only after successfully writing metadata, but I think it makes more sense to update the path either way. Making that change would just mean that failing to write metadata also causes the path in the database to be wrong.

I'll add the changelog entry after review since it depends on the decision made.

JOJ0:
regarding also including write in the plugin, please research if you find something about it in the original implementation or in the docs. i don't know if that was intentional back then. HTH

https://github.com/beetbox/beets/pull/5644

willburden:
I had a look through that PR, and couldn't see any confirmation either way. The documentation added in the PR says it "replaces the audio file of a track, while keeping the name and tags intact." As I read that, it could mean either keeping the tags of the new file or the tags in the database intact, so I think it's still ambiguous? But maybe it's clearer to someone more familiar with this project.

To resolve this, perhaps the plugin author @Uncorrupt3318 has a view? (For context, we're wondering if the replace plugin should automatically copy the tags stored in the database for a track onto the newly swapped-in file.)

Alternatively, I'm happy to remove that change and leave it so the user has to remember to call `beet write` afterward. That would at least allow merging in the most important fix which is just to #6260

Uncorrupt3318:
Hi! Yeah, that's a bit ambiguous, now that I read it again haha

The original idea was to copy the tags from the database to the new file. In other words: you get a music file, you add it to the library, you realise it's the wrong one, you get the good music file, you replace the file without having to go through the tagging process again.

Please keep the change, I don't know if I forgot to add the `write()` call or something, but it should just work without having to use `beet write`. It seems like the plugin didn't work to begin with, that's weird, sorry about that.

I would fix this myself, but I'm not in a great place in my life to do that (moving, new job, etc. etc.). You can tag me and ask, but feel free to do what you think is best with the plugin @willburden . I think I outlined the original idea well enough :)

willburden:
Thanks for the reply, that's a great help! That's exactly my use case, yeah, and it's a really convenient plugin so thanks for making it :)

In that case, I'll leave this PR as is for now, and I've updated the changelog to reflect the fixed/changed behaviour.

JOJ0:
Thanks to both of you! Great news. Let's go ahead with the current implementation then! I'll take a final look soon!

beetsplug/replace.py
JOJ0, line 102: Since we are at it, could you possibly change this to use the provided util for yes/no questions: https://github.com/beetbox/beets/blob/master/beets/ui/__init__.py#L390
Thanks!!

beetsplug/replace.py
```diff
@@ -113,10 +117,17 @@ def replace_file(self, new_file_path: Path, song: Item) -> None:
         ):
             try:
                 original_file_path.unlink()
-            except Exception as e:
+            except OSError as e:
                 raise ui.UserError(f"Could not delete original file: {e}")
 
+        # Update the path to point to the new file.
         song.path = str(dest).encode()
```
JOJ0: Use the `bytestring_path` utility to store the path as bytes.

```
from beets.util import bytestring_path
```

beetsplug/replace.py
```diff
@@ -113,10 +117,17 @@ def replace_file(self, new_file_path: Path, song: Item) -> None:
...
         song.store()
 
+        # Write the metadata in the database to the song file's tags.
+        try:
+            song.write()
```
JOJ0: Would it be possible to use the `try_sync` method on the item? Something like this: 

```
song.try_sync(write=True, move=False)
```

There is also a `try_write` method that does not update the mtime in the db.

Which one would suit the usecase best?

https://github.com/beetbox/beets/blob/master/beets/library/models.py#L989-L1021

JOJ0 (approved):
Some improvement suggestions within, also after our latest merges to master there is a conflict you need to resolve. Thanks!

Update: Sorry hit the wrong button. I wanted to request changes not approve right away! Sorry, please address the suggestions and soon we'll be good to go :-)

JOJ0 (changes requested):
Since an approval review can't be changed, sending anoter review as a "change request". Sorry for the confusion :-)

beetsplug/replace.py
willburden: I think this order of operations makes sense:

1. Update the `Item`'s `path` field to point to the new file.
2. Call `try_write` on the `Item`. This updates the new file's metadata to match the `Item`, and the `Item`'s `mtime` field gets updated (unless preserved by the ImportAdded plugin).
3. Call `store` on the `Item`. This updates the database to match the `Item`, which in this case will update the `path` and `mtime` fields in the database.

Steps 2 and 3 are equivalent to `try_sync(write=True, move=False)`, so that's what I've changed it to. It does pass unit tests and I've tried running it to update a file in my music and it updates all the metadata correctly.

If the write fails, it will log it and still perform the store, so it will have correctly replaced the file but failed only to write the database metadata to the new file. The user can then try again by manually doing `beet write`.
 
Let me know if any of that doesn't seem right.

beetsplug/replace.py
willburden: Done!

beetsplug/replace.py
willburden: Done!

willburden:
Thanks for the review! I'm not sure about this Sourcery issue, it seems unrelated. Other than that, should all be resolved.

docs/plugins/replace.rst
```diff
@@ -15,5 +15,8 @@ and then type:
 The plugin will show you a list of files for you to pick from, and then ask for
 confirmation.
 
+The file you pick will be replaced with the file at `path`, and the tags in the
+database will be written to that file's metadata.
```
JOJ0: Should we be super-explicit here and even mention that updating mtime in the DB item is then taken care of as well? Or ist that "too much information" because people would assume that anyway? Not sure.

In the end if someone is aware of a query like `beet ls mtime:....` they would assume this would still work as promised (DB in sync with file)

But maybe you change your wording to "file synced with the DB" similar to the code comment above.

test/plugins/test_replace.py
```diff
@@ -1,50 +1,138 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+    from beets.library import Library
+
 replace = ReplacePlugin()
 
 
+def always(x):
+    return lambda *args, **kwargs: x
```
JOJ0: I'm not sure about this. Is there maybe a more pythonic / pytest-builtin way for doing this?

test/plugins/test_replace.py
```diff
@@ -1,50 +1,138 @@
...
+
+
+def always_raise(x):
+    def err(*args, **kwargs):
+        raise x
```
JOJ0: also here. maybe use a builtin pytest way if possible.

JOJ0 (changes requested):
Hi, I just realized that I never finished, nor sent my pending review. Submitting it now. Apologiez for the delay! :-)

beetsplug/replace.py
JOJ0: This is good now! Many thanks!

test/plugins/test_replace.py
```diff
@@ -105,11 +193,77 @@ class Song:
...
+        item = Item()
+        item.path = bytes(tmp_path / "not_a_song.mp3")
+
+        with pytest.raises(ui.UserError):
+            replace.replace_file(tmp_path / "not_a_file.opus", item)
+
+    def test_replace_file_delete_fails(
+        self, library, mp3_file, opus_file, monkeypatch
+    ):
+        monkeypatch.setattr(Path, "unlink", always_raise(OSError))
```
JOJ0:
```suggestion
        fail_unlink = Mock(side_effect=OSError("cannot delete"))
        monkeypatch.setattr(Path, "unlink", fail_unlink)
```

would this  be slightly easier to read? what do you think?

emiham:
Sorry to hijack this and go in a bit of a different direction, but I think it would be better for this plugin to copy the tags from the original file to the new one, rather than to read them from the database. 
1. If you have any tags that were not written with beets those would be lost (could be edited tags that exist in beets or tags that don't even exist in beets)
2. You can always overwrite the file tags with the information from beets later, but once beets writes to the file the original tags are lost

Then we could either add a flag (and probably a config option to default to this behavior) to write tags from the database, or clarify in the documentation that you might want to run `beet write` after replacing tags.

beetsplug/replace.py
```diff
@@ -84,7 +89,7 @@ def select_song(self, items: list[Item]):
             except ValueError:
                 ui.print_("Invalid input. Please type in a number.")
```
nolsto: I had a separate PR ready for submission before finding this one. The sole concern in mine was removing all raw `input()` prompts with their analogues in `beets.ui`. I propose this block should be replaced with:
~~~python3
        index = ui.input_options(
            [],
            require=True,
            prompt=(
                f"Which song would you like to replace? "
                f"[1-{len(items)}] (0 to cancel):"
            ),
            numrange=(0, len(items)),
        )
        if index == 0:
            return None
        return items[index - 1]
~~~

beetsplug/replace.py
```diff
@@ -95,12 +100,10 @@ def confirm_replacement(self, new_file_path: Path, song: Item):
             f"\nReplacing: {util.displayable_path(new_file_path)} "
             f"-> {util.displayable_path(original_file_path)}"
         )
-        decision: str = (
-            input("Are you sure you want to replace this track? (y/N): ")
-            .strip()
-            .casefold()
+
+        return ui.input_yn(
+            "Are you sure you want to replace this track (y/n)?", require=True
```
nolsto, line 100: I don't think the `require` param should be used as True here. The yes action is destructive and should probably not be the default without being explicitly given. That is how it was before.

nolsto (changes requested):
Some advice concerning this plugin's input prompting.

beetsplug/replace.py
willburden: Yep makes sense, I've made the change!

beetsplug/replace.py
willburden: I'm not clear on this. My understanding is that `require=True` means the yes action does have to be explicitly given. If it's not there, simply pressing Enter on the confirmation prompt is considered confirmation.

willburden:
> Sorry to hijack this and go in a bit of a different direction, but I think it would be better for this plugin to copy the tags from the original file to the new one, rather than to read them from the database.

@emiham That does make sense to me, I agree that we should favour the least destructive action. I will change the behaviour to this when I have some time.

> Then we could either add a flag (and probably a config option to default to this behavior) to write tags from the database, or clarify in the documentation that you might want to run beet write after replacing tags.

I think for this PR we should clarify in the documentation, and then maybe flags and config options can be added in a later PR? This PR has already expanded a little (due to my own eagerness!) beyond the initial issue which was just fixing the replace plugin CLI being broken.

beetsplug/replace.py
nolsto: Oops. You're right. I had it exactly opposite in my head. Carry on!

nolsto: Github is telling me it "Failed to resolve thread" when I try. 🤷

JOJ0:
I agree that this PR fixes the initial issue now and we should open a new issue/PR for this new feature idea. Which I find is a good one btw! The current behavior though is good enough or rather what beets does per design - being the source of truth.

JOJ0:
Hmm some unanswered change requests/ideas in here and a rebase is to be done @willburden

docs/plugins/replace.rst
willburden: I've made it a bit more explicit that the metadata is synced with the database as you suggested. It now mentions that most tags are written from the database to the file but that `path` and `mtime` are written from the file to the database.

test/plugins/test_replace.py
willburden: Yep, this does look better, I agree that the `always` thing looks a bit clunky/unintuitive in Python.

I've removed `always` and `always_raise` and replaced all of their usages with `Mock`, using either the `side_effect` or `return_value` argument as needed.

willburden:
@JOJ0 I believe I've now addressed everything. Sorry for the long delays in this thread! (I've been moving to a new job/city.)

willburden:
Regarding @emiham's suggestion, I've left the behaviour as syncing the new file with the database rather than with the old file. Syncing with the old file can be a new feature request as discussed :)

test/plugins/test_replace.py
```diff
@@ -105,11 +183,77 @@ class Song:
...
+        replace.replace_file(opus_file, item)
+
+        # Check that the file has been replaced.
+        assert opus_file.exists()
+        assert not mp3_file.exists()
+
+        # Check that the database path and mtime have been updated.
+        item.load()
+        assert item.path == bytes(opus_file)
+        assert item.mtime > 0
```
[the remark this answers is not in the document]
snejus: @willburden note that `item.filepath` contains `Path` instance with the path if that's in any way helpful

snejus:
Hi @willburden, are you planning to come back to this?

snejus:
Seems like this issue had been fixed in another PR!

PR #6220, opened 2025-12-17
Titlecase Plugin Improvements

henry-oberholtzer:
- Add preserving strings that are all lowercase or all upper case
- Fix spelling of 'separator' in config, docs and code
- Move most of the logging for the plugin to debug to keep log cleaner.

Improvements I found a need for in my daily use with the plugin.

- [x] Changelog. (Skipping as the plugin has not been released yet)

### DIFF
--- a/test/plugins/test_titlecase.py
+++ b/test/plugins/test_titlecase.py
@@ -140,8 +140,8 @@ class TestTitlecasePlugin(PluginTestCase):
             (False, "RBMK1000", "Rbmk1000"),
             (True, "P A R I S!", "P A R I S!"),
             (True, "pillow dub...", "Pillow Dub..."),
-            (False, "P A R I S!", "P a R I S!")
-                ]
+            (False, "P A R I S!", "P a R I S!"),
+        ]
         for testcase in testcases:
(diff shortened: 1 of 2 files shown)


semohr (approved):
Good stuff!

test/plugins/test_titlecase.py
```diff
@@ -129,8 +129,34 @@ def test_seperators(self):
             ),
         ]
         for testcase in testcases:
-            seperators, given, expected = testcase
-            with self.configure_plugin({"seperators": seperators}):
+            separators, given, expected = testcase
+            with self.configure_plugin({"separators": separators}):
+                assert TitlecasePlugin().titlecase(given) == expected
+
+    def test_all_caps(self):
```
semohr, line 136: This seems like a prime example of where a `pytest.parametrize` would be be my preferred syntax. We should spent some time in the future to fully migrate our test setup to pytest :/

henry-oberholtzer: I agree! I didn't really want to re-invent the wheel for this plugin, but maybe a PR adding equivalent/replacement helper methods in Pytest could be good to look into.

PR #6224, opened 2025-12-17
Add multiartist support for Spotify

Nedra1998:
## Description

This PR improves/adds multi-artist support for the spotify plugin and the lastgenre plugin.

For spotify, it simply returns the artists in a list as the `artists` and `albumartists` fields respectively, similar to how the musicbrainz plugin handled this.

For lastgenre, it first does a lookup using the `artist` or `albumartist`. But if there are no results, then it will fallback to trying the individual artists if they are available, returning the first result that includes any genres. For example this album:
[Project Skylate & Sugar Shrill - Akiba Night](https://musicbrainz.org/release-group/32cceacf-eea3-4c91-a374-83926288fb5a). The `artist` and `albumartist` is set to "Project Skylate & Sugar Shrill". This results in finding [this match](https://www.last.fm/music/Project+Skylate+&+Sugar+Shrill/_/Akiba+Nights) from last.fm which has no tags, instead of [this one](https://www.last.fm/music/Project+Skylate/_/Akiba+Nights) which is under the first artist and is significantly more popular and includes some tags.

> [!NOTE]
> This is my first change to the codebase so I'm not the most familiar with it, just trying to fix some things that I ran into with my own music library. If there are better ways to address these please let me know and I'd love to learn more.

## To Do

- [X] ~Documentation. (If you've added a new command-line flag, for example, find the appropriate page under `docs/` to describe it.)~
- [X] Tests. (Very much encouraged but not strictly required.)

### DIFF
--- a/beetsplug/spotify.py
+++ b/beetsplug/spotify.py
@@ -304,7 +304,7 @@ class SpotifyPlugin(
 
     def _multi_artist_credit(
         self, artists: list[dict[str | int, str]]
-    ) -> tuple[list[str], list[str | None]]:
+    ) -> tuple[list[str], list[str]]:
         """Given a list of artist dictionaries, accumulate data into a pair
         of lists: the first being the artist names, and the second being the
         artist IDs.
@@ -312,11 +312,8 @@ class SpotifyPlugin(
         artist_names = []
         artist_ids = []
         for artist in artists:
-            # Still use the get_artist helper to handle the artical
-            # normalization for each individual artist.
-            name, id = self.get_artist([artist])
-            artist_names.append(name)
-            artist_ids.append(id)
+            artist_names.append(artist["name"])
+            artist_ids.append(artist["id"])
         return artist_names, artist_ids
 
     def album_for_id(self, album_id: str) -> AlbumInfo | None:
--- a/test/plugins/test_spotify.py
+++ b/test/plugins/test_spotify.py
@@ -251,20 +251,36 @@ class SpotifyPluginTest(PluginTestCase):
             assert query.isascii()
 
     @responses.activate
-    def test_multi_artist_album(self):
-        """Tests if plugin is able to map multiple artists in an album"""
+    def test_multiartist_album_and_track(self):
+        """Tests if plugin is able to map multiple artists in an album and
+        track info correctly"""
 
         # Mock the Spotify 'Get Album' call
         json_file = os.path.join(
-            _common.RSRC, b"spotify", b"multi_artist_request.json"
+            _common.RSRC, b"spotify", b"multiartist_album.json"
         )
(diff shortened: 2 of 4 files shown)


Nedra1998:
Ah, sorry I missed that other PR. I've removed the lastgenre changes from this PR to just focus on Spotify. And I'll test out the other PR and add a comment on it.

beetsplug/spotify.py
```diff
@@ -302,6 +302,23 @@ def _handle_response(
...
+        """Given a list of artist dictionaries, accumulate data into a pair
+        of lists: the first being the artist names, and the second being the
+        artist IDs.
+        """
+        artist_names = []
+        artist_ids = []
+        for artist in artists:
+            # Still use the get_artist helper to handle the artical
+            # normalization for each individual artist.
+            name, id = self.get_artist([artist])
```
henry-oberholtzer: I don't think we need this call to .get_artist anymore, the only part that ends up being used is the part that moves articles to front. Spotify doesn't place articles at the end of artist names so we don't need to worry about that here, as far as I am aware.

test/plugins/test_spotify.py
```diff
@@ -249,3 +249,27 @@ def test_japanese_track(self):
...
+            f"{spotify.SpotifyPlugin.album_url}0yhKyyjyKXWUieJ4w1IAEa",
+            body=response_body,
+            status=200,
+            content_type="application/json",
+        )
+
+        album_info = self.spotify.album_for_id("0yhKyyjyKXWUieJ4w1IAEa")
+        assert album_info is not None
+        assert album_info.artist == "Project Skylate, Sugar Shrill"
+        assert album_info.artists == ["Project Skylate", "Sugar Shrill"]
```
henry-oberholtzer, line 291: Let's test the artist_ids here and make sure those are processed as expected too.

Testing for the track information to make sure those multiartist fields are capture correctly too would be good. Modifying the track's artist fields to be different than the album artist's fields for clarity would be good as well.

beetsplug/spotify.py
Nedra1998: Ah, good to know that we don't need it for Spotify, I didn't think it was needed but wasn't entirely sure, I've removed that to just get the name and id directly.

test/plugins/test_spotify.py
Nedra1998: Good suggestion, I've updated the test to check the `artists_ids`, as well as checking the `album_info.tracks` that those fields are set appropriately. And I also added a check for the track info from `track_for_id`. And tweaked the API JSON so that the track artists are different than the album artist.

henry-oberholtzer (approved):
This looks great to me! Thanks for making those changes.

PR #6233, opened 2025-12-23
Fix lyrics Unicode corruption and escaped quotes in Genius plugin

grillofran:
## Problem
The lyrics plugin has two bugs that corrupt fetched lyrics:

1. **Unicode corruption**: Characters like `ò`, `è`, `à` are corrupted to `√≤`, `√®`, etc.
2. **Escaped quotes**: Quotes appear as `\"` instead of `"` in lyrics

### Issue 1: MacRoman encoding misdetection
- **Location**: `RequestHandler.fetch_text()` line 220
- **Cause**: Setting `r.encoding = None` forces requests to use `apparent_encoding`
- **Problem**: For Genius.com (and others), requests incorrectly detects MacRoman instead of UTF-8
- **Result**: UTF-8 bytes `c3 b2` (ò) decoded as MacRoman produces "√≤" (U+221A U+2264)

### Issue 2: Incomplete JSON unescape
- **Location**: `Genius.scrape()` line 576
- **Cause**: The `remove_backslash` regex doesn't handle all escape patterns in JSON
- **Problem**: Genius embeds lyrics in JSON with patterns like `\\"` and `\\\\"` 
- **Result**: After BeautifulSoup processing, escaped quotes remain in final text

### Fix 1: Trust server encoding, fallback to UTF-8
```python
# OLD: r.encoding = None
# NEW:
if not r.encoding:
    r.encoding = 'utf-8'
```
- Respects server's declared encoding (UTF-8 for Genius)
- Falls back to UTF-8 if no encoding specified (safer than apparent_encoding)
- Preserves original intent of handling misconfigured servers

### Fix 2: Iteratively clean escaped quotes
```python
while '\\"' in lyrics:
    lyrics = lyrics.replace('\\"', '"')
```
- Handles variable escape levels (`\"`, `\\\"`, `\\\\\"`)
- Minimal change - keeps original `remove_backslash` regex
- Applied after BeautifulSoup to avoid interfering with HTML parsing

## Testing

Tested with:
- Caparezza - "Argenti Vive" (Italian, many accented characters)
- WestsideGunn - "Heel Cena" (escaped quotes in lyrics)

Before:
```
mi si par√≤ davanti
\\"I got big moves\\"
```

After:
```
mi si parò davanti
"I got big moves"
```

## Impact
- Fixes lyrics for all languages with non-ASCII characters
- Fixes Genius lyrics with quotes
- No breaking changes - maintains backward compatibility
- Minimal code changes (14 lines total)

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -190,7 +190,11 @@ class LyricsRequestHandler(RequestHandler):
         return f"{url}?{urlencode(params)}"
 
     def get_text(
-        self, url: str, params: JSONDict | None = None, force_utf8: bool = False, **kwargs
+        self,
+        url: str,
+        params: JSONDict | None = None,
+        force_utf8: bool = False,
+        **kwargs,
     ) -> str:
         """Return text / HTML data from the given URL.
 
@@ -546,9 +550,9 @@ class Genius(SearchBackend):
     def headers(self) -> dict[str, str]:
         return {"Authorization": f"Bearer {self.config['genius_api_key']}"}
 
-    def get_text(self, *args, **kwargs) -> str:
+    def get_text(self, url: str, params: JSONDict | None = None, **kwargs) -> str:
         """Force UTF-8 encoding for Genius to avoid MacRoman misdetection."""
-        return super().get_text(*args, force_utf8=True, **kwargs)
+        return super().get_text(url, params, force_utf8=True, **kwargs)
 
     def search(self, artist: str, title: str) -> Iterable[SearchResult]:
         search_data: GeniusAPI.Search = self.get_json(
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -198,15 +198,16 @@ class LyricsRequestHandler(RequestHandler):
     ) -> str:
         """Return text / HTML data from the given URL.
 
-        By default, trust the server's encoding and requests' apparent_encoding
-        detection. When force_utf8=True, default to UTF-8 if server doesn't
-        specify encoding (avoids MacRoman misdetection on some sites like Genius).
+        Set encoding to None to let requests auto-detect (works for most sites).
+        For Genius, force UTF-8 to avoid MacRoman misdetection.
         """
         url = self.format_url(url, params)
         self.debug("Fetching HTML from {}", url)
         r = self.get(url, **kwargs)
         if force_utf8:
             r.encoding = r.encoding or "utf-8"
+        else:
+            r.encoding = None
         return r.text
 
     def get_json(self, url: str, params: JSONDict | None = None, **kwargs):
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -190,21 +190,19 @@ class LyricsRequestHandler(RequestHandler):
         return f"{url}?{urlencode(params)}"
 
     def get_text(
-        self, url: str, params: JSONDict | None = None, **kwargs
+        self, url: str, params: JSONDict | None = None, force_utf8: bool = False, **kwargs
     ) -> str:
         """Return text / HTML data from the given URL.
 
-        Set the encoding to None to let requests handle it because some sites
-        set it incorrectly.
+        By default, trust the server's encoding and requests' apparent_encoding
+        detection. When force_utf8=True, default to UTF-8 if server doesn't
+        specify encoding (avoids MacRoman misdetection on some sites like Genius).
         """
         url = self.format_url(url, params)
         self.debug("Fetching HTML from {}", url)
         r = self.get(url, **kwargs)
-        """Trust server's encoding,
-        but default to UTF-8 if not specified
-        """
-        if not r.encoding:
-            r.encoding = "utf-8"
+        if force_utf8:
+            r.encoding = r.encoding or "utf-8"
         return r.text
 
     def get_json(self, url: str, params: JSONDict | None = None, **kwargs):
@@ -548,6 +546,10 @@ class Genius(SearchBackend):
     def headers(self) -> dict[str, str]:
         return {"Authorization": f"Bearer {self.config['genius_api_key']}"}
 
+    def get_text(self, *args, **kwargs) -> str:
+        """Force UTF-8 encoding for Genius to avoid MacRoman misdetection."""
+        return super().get_text(*args, force_utf8=True, **kwargs)
+
     def search(self, artist: str, title: str) -> Iterable[SearchResult]:
         search_data: GeniusAPI.Search = self.get_json(
             self.SEARCH_URL,
@@ -562,8 +564,7 @@ class Genius(SearchBackend):
         if m := cls.LYRICS_IN_JSON_RE.search(html):
             html_text = cls.remove_backslash(m[0]).replace(r"\n", "\n")
             lyrics = cls.get_soup(html_text).get_text().strip()
-            # Clean up any remaining escaped quotes (may need multiple passes)
-            # Use regex to handle all escape levels in one pass
+            # Genius embeds lyrics in JSON; escape sequences remain after parsing
             lyrics = re.sub(r'\\+"', '"', lyrics)
             return lyrics
 
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -569,6 +569,7 @@ class Genius(SearchBackend):
 
         return None
 
+
 class Tekstowo(SearchBackend):
     """Fetch lyrics from Tekstowo.pl."""
 
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -200,7 +200,11 @@ class LyricsRequestHandler(RequestHandler):
         url = self.format_url(url, params)
         self.debug("Fetching HTML from {}", url)
         r = self.get(url, **kwargs)
-        r.encoding = None
+        """Trust server's encoding,
+        but default to UTF-8 if not specified
+        """
+        if not r.encoding:
+            r.encoding = 'utf-8'
         return r.text
 
     def get_json(self, url: str, params: JSONDict | None = None, **kwargs):
@@ -557,10 +561,13 @@ class Genius(SearchBackend):
     def scrape(cls, html: str) -> str | None:
         if m := cls.LYRICS_IN_JSON_RE.search(html):
             html_text = cls.remove_backslash(m[0]).replace(r"\n", "\n")
-            return cls.get_soup(html_text).get_text().strip()
+            lyrics = cls.get_soup(html_text).get_text().strip()
+            # Clean up any remaining escaped quotes (may need multiple passes)
+            while '\\"' in lyrics:
+                lyrics = lyrics.replace('\\"', '"')
+            return lyrics
 
         return None
-
 
 class Tekstowo(SearchBackend):
     """Fetch lyrics from Tekstowo.pl."""
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -563,8 +563,8 @@ class Genius(SearchBackend):
             html_text = cls.remove_backslash(m[0]).replace(r"\n", "\n")
             lyrics = cls.get_soup(html_text).get_text().strip()
             # Clean up any remaining escaped quotes (may need multiple passes)
-            while '\\"' in lyrics:
-                lyrics = lyrics.replace('\\"', '"')
+            # Use regex to handle all escape levels in one pass
+            lyrics = re.sub(r'\\+"', '"', lyrics)
             return lyrics
 
         return None
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -550,9 +550,15 @@ class Genius(SearchBackend):
     def headers(self) -> dict[str, str]:
         return {"Authorization": f"Bearer {self.config['genius_api_key']}"}
 
-    def get_text(self, url: str, params: JSONDict | None = None, **kwargs) -> str:
+    def get_text(
+        self,
+        url: str,
+        params: JSONDict | None = None,
+        force_utf8: bool = True,
+        **kwargs,
+    ) -> str:
         """Force UTF-8 encoding for Genius to avoid MacRoman misdetection."""
-        return super().get_text(url, params, force_utf8=True, **kwargs)
+        return super().get_text(url, params, force_utf8=force_utf8, **kwargs)
 
     def search(self, artist: str, title: str) -> Iterable[SearchResult]:
         search_data: GeniusAPI.Search = self.get_json(
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -204,7 +204,7 @@ class LyricsRequestHandler(RequestHandler):
         but default to UTF-8 if not specified
         """
         if not r.encoding:
-            r.encoding = 'utf-8'
+            r.encoding = "utf-8"
         return r.text
 
     def get_json(self, url: str, params: JSONDict | None = None, **kwargs):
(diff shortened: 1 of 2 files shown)


grillofran:
Is the 66% patch coverage requirement strict for bug fixes? The changes are small edge-case improvements. Happy to add tests if required!

beetsplug/lyrics.py
```diff
@@ -200,7 +200,11 @@ def get_text(
         url = self.format_url(url, params)
         self.debug("Fetching HTML from {}", url)
         r = self.get(url, **kwargs)
-        r.encoding = None
+        """Trust server's encoding,
```
semohr: We can simplify this:
```
# trust server encoding and default to utf-8
r.encoding = r.encoding or "utf-8"
```

beetsplug/lyrics.py
```diff
@@ -557,7 +561,11 @@ def search(self, artist: str, title: str) -> Iterable[SearchResult]:
     def scrape(cls, html: str) -> str | None:
         if m := cls.LYRICS_IN_JSON_RE.search(html):
             html_text = cls.remove_backslash(m[0]).replace(r"\n", "\n")
-            return cls.get_soup(html_text).get_text().strip()
+            lyrics = cls.get_soup(html_text).get_text().strip()
+            # Clean up any remaining escaped quotes (may need multiple passes)
```
semohr: Comments should be minimal and focus on the why rather than the how:
```
#  Strip HTML and normalize quotes that were left because of JSON...
return re.sub(r'\\+"', '"', lyrics)
```
It should explain why the escaped quotes exist without repeating how the regex works. You have quite some information on this in your PR, maybe we can condense it into a sentence?

semohr (changes requested):
Thank you for the PR. Since this seems to be your first contribution to beets, you might want to have a look at our [developer guides](https://beets.readthedocs.io/en/stable/dev/index.html). Welcome aboard!

> Is the 66% patch coverage requirement strict for bug fixes? The changes are small edge-case improvements. Happy to add tests if required!

While it is not a requirement, for new code tests are always highly appreciated. Here for example showing that JSON encoded lyrics can be parsed as expected would be a good addition.  Same for the inverse case showing that "normal" html lyrics still work as expected would be good (I guess the test already exists, havent look at our testsuit 😁 )

----

More of a general question: how come this was never raised as an issue? Seems like it should be a common issue to me. Have you tried to find related open issues yet?

beetsplug/lyrics.py
snejus: @grillofran setting encoding to `None` is required for some websites under `Google` provider, such as **AZLyrics**. Your fix should apply to `Genius` backend only.

You could potentially add a new parameter `reset_encoding: bool=False` to `LyricsRequestHandler.get_text`, and adjust `Google.get_text` accordingly?

snejus:
> Tested with:
> 
>     * Caparezza - "Argenti Vive" (Italian, many accented characters)
> 
>     * WestsideGunn - "Heel Cena" (escaped quotes in lyrics)

Are you happy to add both cases to `test/plugins/lyrics_pages.py`?

grillofran:
Thanks for the feedback! I've updated the PR to address all your points.

- Made the encoding fix Genius-only with a force_utf8 parameter (thanks @snejus)
- Simplified the encoding check to r.encoding = r.encoding or "utf-8" (thanks @semohr)
- Updated the comment to focus on why the escapes exist
- Added test cases for Caparezza (Italian accents) and Arctic Monkeys (quoted lyrics), as the other one was... too explicit

Funny enough.... I couldn't find any issues reporting this in the repo's issue tracker, which I thought was strange given how many people must use Genius. Also, this is my first contribution to beets so apologies if I missed anything. Still learning the ropes! Let me know if there's anything else that needs adjusting.

beetsplug/lyrics.py
```diff
@@ -190,17 +190,23 @@ def format_url(url: str, params: JSONDict | None) -> str:
...
+        By default, trust the server's encoding and requests' apparent_encoding
+        detection. When force_utf8=True, default to UTF-8 if server doesn't
+        specify encoding (avoids MacRoman misdetection on some sites like Genius).
         """
         url = self.format_url(url, params)
         self.debug("Fetching HTML from {}", url)
         r = self.get(url, **kwargs)
-        r.encoding = None
+        if force_utf8:
+            r.encoding = r.encoding or "utf-8"
```
snejus, line 223: This only handles `Genius` requirements: to clarify, for `Google`, we want:

```py
r.encoding = None
```

same it was before.

snejus (changes requested):

beetsplug/lyrics.py
semohr: Could we add a test for the specifc case where this is an issue please?

For me (without researching and testing it further) using the encoding of the response seems reasonable.

snejus: Luckily, I managed to dig this URL out from my `ipython` history. Add this to `test/plugins/lyrics_pages.py`

```py
    LyricsPage.make(
        "https://www.azlyrics.com/lyrics/derivakat/bountyhunter.html",
        """
        [Romanized:]

        Living legend made of myths and made of stories
        But I'm present, find the mark and I'll take the glory
        Hit the dead-end, cornered like an animal, it just
        Takes one second, I'm nothing if not practical
        But I don't really care if you're right or you're wrong
        Bùguǎn nǐ de wùhuì or you're just a disgrace
        Get in the zone when I put on my song
        Bìrán de xiànshí when I'm up in your face
        Zhēn tài kě qíng kùn zài lóng lǐ de nǐ
        But you'll find in the end that I'll leave you erased
        (If they call?) I'm the one
        (Got a job?) Get it done
        All alone, you and me, I'll be leaving no trace

        No place where I won't go to
        No limits I won't break through
        No godless deed I won't do
        If they got problems, then they know to

        Call the bounty hunter
        Yeah, they got my number
        I'm the problem solver
        Leave you six feet under
        You won't see another
        Winter, spring or summer
        Wǒ huì ràng nǐ xiāoshī
        No way to recover (Hey!)

        Run it up
        Night or day (Hey!)
        Hunt you down
        Run away

        No mercy from this mercenary
        Take away your sanctuary
        Shìlì jiù xiàng 20/20
        Way I see you'll end up buried
        Don't deviate
        Zhǎodào nǐ so I terminate
        Wúlùn, wǒ huì make you pay
        I'll double tap you like bang, bang

        (Dead lilac) A ghost you only heard in rumor
        (Dead lilac) But only if you heard me sooner
        (Dead lilac) Losing your chances and losing your time
        (Dead lilac) shìjiè wèilái méi nǐ de cúnzài (Yeah)

        No place where I won't go to (I won't go to)
        No limits I won't break through (I won't break through)
        No godless deed I won't do (That I won't do)
        If they got problems, then they know to

        Call the bounty hunter (Hey)
        Yeah, they got my number
        I'm the problem solver (Hey)
        Leave you six feet under
        You won't see another (Hey)
        Winter, spring or summer
        Wǒ huì ràng nǐ xiāoshī (Hey)
        No way to recover (Hey!)

        Run it up (Run it up)
        Night or day (Hey!) (Day)
        Hunt you down (Hunt you down)
        Run away

        [English/Chinese:]

        Living legend made of myths and made of stories
        But I'm present, find the mark and I'll take the glory
        Hit the dead-end, cornered like an animal, it just
        Takes one second, I'm nothing if not practical
        But I don't really care if you're right or you're wrong
        不管你的误会 or you're just a disgrace
        Get in the zone when I put on my song
        必然的现实 when I'm up in your face
        真太可情 困在笼里的你
        But you'll find in the end that I'll leave you erased
        (If they call?) I'm the one
        (Got a job?) Get it done
        All alone, you and me, I'll be leaving no trace

        No place where I won't go to
        No limits I won't break through
        No godless deed I won't do
        If they got problems, then they know to

        Call the bounty hunter
        Yeah, they got my number
        I'm the problem solver
        Leave you six feet under
        You won't see another
        Winter, spring or summer
        我会让你消失
        No way to recover (Hey!)

        Run it up
        Night or day (Hey!)
        Hunt you down
        Run away

        No mercy from this mercenary
        Take away your sanctuary
        视力就像 20/20
        Way I see you'll end up buried
        Don't deviate
        找到你 so I terminate
        无论, 我会 make you pay
        I'll double tap you like bang, bang

        (Dead lilac) A ghost you only heard in rumor
        (Dead lilac) But only if you heard me sooner
        (Dead lilac) Losing your chances and losing your time
        (Dead lilac) 世界未来没你的存在 (Yeah)

        No place where I won't go to (I won't go to)
        No limits I won't break through (I won't break through)
        No godless deed I won't do (That I won't do)
        If they got problems, then they know to

        Call the bounty hunter (Hey)
        Yeah, they got my number
        I'm the problem solver (Hey)
        Leave you six feet under
        You won't see another (Hey)
        Winter, spring or summer
        我会让你消失 (Hey)
        No way to recover (Hey!)

        Run it up (Run it up)
        Night or day (Hey!) (Day)
        Hunt you down (Hunt you down)
        Run away
        """,
        artist="Derivakat",
        track_title="Bounty Hunter",
        url_title="Derivakat - Bounty Hunter Lyrics | AZLyrics.com",
        marks=[xfail_on_ci("AZLyrics is blocked by Cloudflare")],
    ),
```

and run

```bash
$ LYRICS_UPDATED=true poe test test/plugins/test_lyrics.py -k azlyrics -vv
```

[asciicast](https://asciinema.org/a/2Qqp0Nd1imH3eemVc6GTivznc)

grillofran:
Hey folks (happy holidays)

Ah, got it. I think I had misunderstood. I added the `else: r.encoding = None` to restore the original behavior for Google/AZLyrics. Should fix the Chinese character issue in the test. Setting `r.encoding = None` explicitly should fix that while keeping the UTF-8 fix for Genius only. Thoughts?

Thanks!

beetsplug/lyrics.py
```diff
@@ -557,7 +574,10 @@ def search(self, artist: str, title: str) -> Iterable[SearchResult]:
     def scrape(cls, html: str) -> str | None:
         if m := cls.LYRICS_IN_JSON_RE.search(html):
             html_text = cls.remove_backslash(m[0]).replace(r"\n", "\n")
-            return cls.get_soup(html_text).get_text().strip()
+            lyrics = cls.get_soup(html_text).get_text().strip()
+            # Genius embeds lyrics in JSON; escape sequences remain after parsing
+            lyrics = re.sub(r'\\+"', '"', lyrics)
+            return lyrics
```
snejus: Now that we've sorted the encoding issue, I wonder whether this could possibly be solved by improving the pattern used by `cls.remove_backslash`?

snejus: @grillofran have you seen my comment by any chance?

snejus:
@semohr worth having a look for you since I made some adjustments under `beetsplug/_utils`

beetsplug/_utils/requests.py
```diff
@@ -21,10 +21,13 @@
 class BeetsHTTPError(requests.exceptions.HTTPError):
     STATUS: ClassVar[HTTPStatus]
 
-    def __init__(self, *args, message: str | None = None, **kwargs) -> None:
-        if not message:
-            message = f"HTTP Error: {self.STATUS.value} {self.STATUS.phrase}"
+    @classmethod
+    def get_status_message(cls) -> str:
```
semohr: Is this used yet? Should we postpone add this until it is actually used?

beetsplug/_utils/requests.py
```diff
@@ -71,10 +74,15 @@ class TimeoutAndRetrySession(requests.Session, metaclass=SingletonMeta):
     * raises exceptions for HTTP error status codes
     """
 
-    def __init__(self, *args, **kwargs) -> None:
+    raise_for_status: bool
```
semohr: Small comment on what the `raise_for_status` param does would be appreciated.

beetsplug/lyrics.py
```diff
@@ -662,14 +675,37 @@ class Genius(SearchBackend):
     def headers(self) -> dict[str, str]:
         return {"Authorization": f"Bearer {self.config['genius_api_key']}"}
 
+    def create_session(self) -> requests.Session:
+        """Return session for Genius HTTP requests.
+
+        HTTP errors are not raised for Genius API requests, because Genius
```
semohr: This seems a bit like a hacky workaround to me. I would prefer if we catch the `BeetsHTTPError` and get the response from the error if it is needed. That way raising from our session stays consistent.

This should work or am I missing something?
```
try:
    response = requests.get(url)
except BeetsHTTPError as e:
    data = e.response.json()
    print(data)
```

semohr (changes requested):
Thanks for pinging me! 

I don't really think we need to add the `raise_for_status` logic. I do not see the need for it right here, also adding it makes sessions a bit inconsistent imo.

beetsplug/_utils/requests.py
snejus: It is: see `UnauthorizedMBError` and `TooManyRequestsError`

beetsplug/_utils/requests.py
snejus: I thought it's pretty self-explanatory?

beetsplug/lyrics.py
snejus: This doesn't work since `raise_for_status` raises and `response` does not get returned.

semohr: response should be attached to the error tho since we extend from HTTPError.

semohr: We do something similar in tidal to get response header for the length of the timeout. https://github.com/beetbox/beets/blob/5e6d9adbe79ed8885800bee4804d2eee14f6eb48/beetsplug/tidal/session.py#L88

snejus: Very good point! Adjusted

beetsplug/_utils/requests.py
snejus: Good point, removed it!

beetsplug/lyrics.py
```diff
@@ -663,13 +676,32 @@ def headers(self) -> dict[str, str]:
         return {"Authorization": f"Bearer {self.config['genius_api_key']}"}
 
     def get_json(self, *args, **kwargs) -> GeniusAPI.Search:
-        response: GeniusAPI.Response = super().get_json(*args, **kwargs)
+        try:
```
semohr, line 679: Should we move the error handling into the `request` function? Currently I think the `get_text` function does not raise errors on http failures, right?

If it is not necessary, also fine. I see that it was like that before 🙃

snejus: > Should we move the error handling into the `request` function? Currently I think the `get_text` function does not raise errors on http failures, right?

Which one? 😅 Doesn't seem like `Genius` defines it?

semohr: I mean the logic for the  `TooManyRequestsHTTPError` and `GeniusHTTPError`. Right now the `get_text` method will never raise these. Not sure if it should but I would hope an api is pretty consistent in their error reporting.

We could likely drop the `get_json` override here in favor for a more generic handling in `request` or even in `handle_http_error`.

---

Very much possible that the code here predates the RequestHandler abstraction.

snejus: Both of these errors depend on the data present in the JSON response!

semohr: > Not sure if it should but I would hope an api is pretty consistent in their error reporting.

We depend on the **error response** containing JSON. If the API uses a consistent error format, though, it’s entirely possible for the response body to contain JSON error data even when you access it via `.text`. We raise the HTTP error before interpreting the response, i.e. before accessing `.text` or `.json()`.

snejus: Note we haven't seen the HTML responses fail like that, even the URL root is different here:

```sh
$ beet lyrics 'title:let me love you' -vv -f --no-keep-synced
...
lyrics: Genius: Fetching JSON from https://api.genius.com/search?q=Mario+Let+Me+Love+You
lyrics: Genius: Fetching HTML from https://genius.com/Mario-let-me-love-you-lyrics
...
```

semohr: Alright!

snejus:
Any blockers left? @semohr

semohr (approved):

PR #6239, opened 2025-12-25
lastgenre: Finalize type hints in plugin

JOJ0:
## Description

Add type hints to the few remaining methods and helpers that didn't have them already.

## To Do

- [x] ~Documentation.~ 
- [x] ~Changelog.~

### DIFF
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -38,6 +38,8 @@ from beets.library import Album, Item
 from beets.util import plurality, unique_list
 
 if TYPE_CHECKING:
+    import optparse
+
     from beets.library import LibModel
 
 LASTFM = pylast.LastFMNetwork(api_key=plugins.LASTFM_KEY)
@@ -52,7 +54,11 @@ PYLAST_EXCEPTIONS = (
 # Canonicalization tree processing.
 
 
-def flatten_tree(elem, path, branches):
+def flatten_tree(
+    elem: dict[Any, Any] | list[Any] | str,
+    path: list[str],
+    branches: list[list[str]],
+) -> None:
     """Flatten nested lists/dictionaries into lists of strings
     (branches).
     """
@@ -69,7 +75,7 @@ def flatten_tree(elem, path, branches):
         branches.append(path + [str(elem)])
 
 
-def find_parents(candidate, branches):
+def find_parents(candidate: str, branches: list[list[str]]) -> list[str]:
     """Find parents genre of a given genre, ordered from the closest to
     the further parent.
     """
@@ -89,7 +95,7 @@ C14N_TREE = os.path.join(os.path.dirname(__file__), "genres-tree.yaml")
 
 
 class LastGenrePlugin(plugins.BeetsPlugin):
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__()
 
         self.config.add(
@@ -111,12 +117,12 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         )
         self.setup()
 
-    def setup(self):
+    def setup(self) -> None:
         """Setup plugin from config options"""
         if self.config["auto"]:
             self.import_stages = [self.imported]
 
-        self._genre_cache = {}
+        self._genre_cache: dict[str, list[str]] = {}
         self.whitelist = self._load_whitelist()
         self.c14n_branches, self.canonicalize = self._load_c14n_tree()
 
@@ -161,7 +167,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
             flatten_tree(genres_tree, [], c14n_branches)
         return c14n_branches, canonicalize
 
-    def _tunelog(self, msg, *args, **kwargs):
+    def _tunelog(self, msg: str, *args: Any, **kwargs: Any) -> None:
         """Log tuning messages at DEBUG level when verbosity level is high enough."""
         if config["verbose"].as_number() >= 3:
             self._log.debug(msg, *args, **kwargs)
@@ -182,7 +188,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
 
     # More canonicalization and general helpers.
 
-    def _get_depth(self, tag):
+    def _get_depth(self, tag: str) -> int | None:
         """Find the depth of a tag in the genres tree."""
         depth = None
         for key, value in enumerate(self.c14n_branches):
@@ -191,7 +197,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
                 break
         return depth
 
-    def _sort_by_depth(self, tags):
+    def _sort_by_depth(self, tags: list[str]) -> list[str]:
         """Given a list of tags, sort the tags by their depths in the
         genre tree.
         """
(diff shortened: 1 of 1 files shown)


semohr (approved):
Seems like a no brainier to me! Thanks for the additions.

snejus:
Thanks for this!

There are still a couple of missing types, would you mind adding them?

```sh
$ mypy beetsplug/lastgenre --strict --allow-untyped-call --pretty
beetsplug/lastgenre/__init__.py:96: error: Function is missing a return type annotation  [no-untyped-def]
        def __init__(self):
        ^
beetsplug/lastgenre/__init__.py:96: note: Use "-> None" if function does not return a value
beetsplug/lastgenre/__init__.py:168: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
        def _tunelog(self, msg, *args, **kwargs) -> None:
        ^
beetsplug/lastgenre/__init__.py:336: error: Returning Any from function declared to return "str"  [no-any-return]
            return self.config["separator"].as_str().join(formatted)
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
beetsplug/lastgenre/__init__.py:379: error: Function is missing a return type annotation  [no-untyped-def]
            def _try_resolve_stage(stage_label: str, keep_genres, new_genres):
            ^
beetsplug/lastgenre/__init__.py:379: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
            def _try_resolve_stage(stage_label: str, keep_genres, new_genres):
            ^
beetsplug/lastgenre/__init__.py:416: error: Returning Any from function declared to return "tuple[str | None, ...]"  [no-any-return]
                        return result
                        ^~~~~~~~~~~~~
beetsplug/lastgenre/__init__.py:423: error: Returning Any from function declared to return "tuple[str | None, ...]"  [no-any-return]
                        return result
                        ^~~~~~~~~~~~~
beetsplug/lastgenre/__init__.py:474: error: Returning Any from function declared to return "tuple[str | None, ...]"  [no-any-return]
                        return result
                        ^~~~~~~~~~~~~
beetsplug/lastgenre/__init__.py:582: error: Function is missing a type annotation  [no-untyped-def]
            def lastgenre_func(lib, opts, args):
            ^
beetsplug/lastgenre/__init__.py:592: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
        def imported(self, session, task) -> None:
        ^
Found 10 errors in 1 file (checked 1 source file)
```

beetsplug/lastgenre/__init__.py
```diff
@@ -329,7 +333,8 @@ def _format_and_stringify(self, tags: list[str]) -> str:
         else:
             formatted = tags
 
-        return self.config["separator"].as_str().join(formatted)
+        joined: str = self.config["separator"].as_str().join(formatted)
+        return joined
```
JOJ0: There are several ways to solve this. confuse unfortunately returns `Any` type. Instead of the extra `joined` variable I could also use `str()` or `cast()` (`from typing import`)

What's our (Beets) preferred way of doing it (for now, until someone adds type hints to confuse ;-))

beetsplug/lastgenre/__init__.py
```diff
@@ -575,7 +582,9 @@ def commands(self):
         )
         lastgenre_cmd.parser.set_defaults(album=True)
 
-        def lastgenre_func(lib, opts, args):
+        def lastgenre_func(
+            lib: library.Library, opts: Any, args: list[str]
```
JOJ0: `opts` could be typed as `optparse.Values` but my pylint gives me a deprecation warning when I `import optparse`, thus I simply chose this easier way. We have both versions in the codebase BTW.

JOJ0 (commented):
> Thanks for this!
> 
> There are still a couple of missing types, would you mind adding them?

Many thanks for pointing that out @snejus . Added all of them and used that handy `mypy` command to check.

Do we usually add a changelog entry when we only add/finalize types in a plugin/code-part?

beetsplug/lastgenre/__init__.py
snejus: Have you updated `confuse` in your local environment? I think types have recently been added. I see them on my end where the previous `return` statement worked fine:

image

beetsplug/lastgenre/__init__.py
snejus: Ignore the `pylint` rule - we _use_ optparse in the codebase, so we should not avoid importint it :)

snejus:
> > Thanks for this!
> > There are still a couple of missing types, would you mind adding them?
> 
> Many thanks for pointing that out @snejus . Added all of them and used that handy `mypy` command to check.
> 
> Do we usually add a changelog entry when we only add/finalize types in a plugin/code-part?

No need :)

beetsplug/lastgenre/__init__.py
JOJ0: I used 2.1.0 installed from pypi so far, now tried with git-cloned version but unfortunately still getting the same mypy error:

```
$ pip list | grep confuse
confuse                       2.1.0           /Users/jojo/git/confuse
$
$
$ mypy beetsplug/lastgenre --strict --allow-untyped-call --pretty
beetsplug/lastgenre/__init__.py:337: error: Returning Any from function declared to return "str"  [no-any-return]
            return self.config["separator"].as_str().join(formatted)
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Found 1 error in 1 file (checked 1 source file)
 ```

Comment (JOJ0) [reply beetsplug/lastgenre/__init__.py]: but you are right type hints were added in 2.1.0. already it seems. it's odd...

https://github.com/beetbox/confuse/commit/f2ec2b03cca291a8457906f74e9bbb7eb81c42a5#diff-2f5e33a65702f7ec5d59376ae7256c5966c132b61a08e4c23783c2146eba7a44R346

Comment (snejus) [reply beetsplug/lastgenre/__init__.py]: Oh, sorry! I forgot that you were checking this in strict mode... You can safely ignore this warning for now. Just use the strict mode for `no-untyped-def` warnings.

I've already taken care of the `Any` issue - just need to submit a PR.

**Review by snejus — APPROVED** [review]

File: beetsplug/lastgenre/__init__.py [hunk 3]
```diff
@@ -585,10 +595,16 @@ def lastgenre_func(lib, opts, args):
         lastgenre_cmd.func = lastgenre_func
         return [lastgenre_cmd]
 
-    def imported(self, session, task):
+    def imported(
+        self, session: library.Session, task: library.ImportTask
+    ) -> None:
```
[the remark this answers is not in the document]
JOJ0: I fixed this as copilot suggests since it sounds a little more precise. Agree that this is the best way @snejus ?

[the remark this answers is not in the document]
JOJ0: Damnit, now running into a new issue. I better revert it as it was....

```
/opt/pipx_bin/poe check-types --show-column-numbers --no-error-summary .
Poe => mypy --show-column-numbers --no-error-summary .
Error: beetsplug/lastgenre/__init__.py:600:56: error: "ImportTask" has no attribute "item"  [attr-defined]
Error: The process '/opt/pipx_bin/poe' failed with exit code 1
```

PR #6241, opened 2025-12-26
Harmonic mix plugin

Exafto:
## Description

This PR adds a new plugin, harmonicmix, which calculates the harmonic compatibility between tracks. Playlist curators may use this to offer more harmonically coherent lists. DJs may use this to search for compatible songs around the same bpm as their current track.

This is my first contribution to Beets! I really enjoy using the tool and wanted to give back using my music theory knowledge. I hope this is helpful, and I am happy to make any changes needed to get this merged. My coding skills are not that sharp, so any feedback is much appreciated! I tried to follow community standards, but please let me know if I missed anything.

## Features:

- Harmonic Matching: Finds tracks compatible with the source song's key using the Circle of Fifths. For example, it is easy to move from a song in C to a song in F, G, or Am. It is really difficult to move to a song in G#.

- Enharmonic Support: Handles standard keys (e.g., C# and Db are treated as the same), as well as more nuanced options (like E#).

- BPM Filtering: Limits results to a mixable BPM range (+/- 8%).

## Future Improvements

- Make the BPM range configurable (currently fixed at +/- 8%).

- Integrate calls to autobpm and keyfinder to fill in missing metadata.

## To Do

- [x] Documentation.
- [x] Changelog.
- [x] Tests.

### DIFF
--- a/beetsplug/harmonicmix.py
+++ b/beetsplug/harmonicmix.py
@@ -16,8 +16,8 @@
 The results also have a matching BPM (+/- 8% from the source track).
 """
 
-from beets.plugins import BeetsPlugin
 from beets import ui
+from beets.plugins import BeetsPlugin
 
 
 class HarmonicLogic:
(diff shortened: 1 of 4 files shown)


Exafto:
I think I'm finished with the code and docs, but the documentation linter keeps failing on the CI server even though it passes locally on my Windows machine. I made some attempts but I cannot pinpoint what's the problem. Could you help me spot the culprit?

JOJ0:
> I think I'm finished with the code and docs, but the documentation linter keeps failing on the CI server even though it passes locally on my Windows machine. I made some attempts but I cannot pinpoint what's the problem. Could you help me spot the culprit?

You tried those `poe` commands?

https://beets.readthedocs.io/en/latest/contributing.html#style

JOJ0:
But first of all: Welcome! Thanks for the contribution. I've always wanted to add circle of fiths functionality to Beets but I was picturing it a little differently, the main point being to be able to use it as a general sorting mechanism so it's usable with the `smartplaylist` plugin too. I'll try to come up with a high level review during the next couple of days pointing out in which direction this could go and I spotted a view low level design improvement possibilities from a quick look already too. Also there is some minimal logic around the topic here already, which we should keep in mind: https://github.com/beetbox/beets/blob/master/beets/dbcore/types.py#L403 . Speak soon!

Exafto:
> But first of all: Welcome! Thanks for the contribution. I've always wanted to add circle of fiths functionality to Beets but I was picturing it a little differently, the main point being to be able to use it as a general sorting mechanism so it's usable with the `smartplaylist` plugin too. I'll try to come up with a high level review during the next couple of days pointing out in which direction this could go and I spotted a view low level design improvement possibilities from a quick look already too. Also there is some minimal logic around the topic here already, which we should keep in mind: https://github.com/beetbox/beets/blob/master/beets/dbcore/types.py#L403 . Speak soon!

Hello! Thank you for replying so quickly and for welcoming me!

- Firstly, I ran the `poe` commands mentioned in the style guide locally, and no changes took place. I could check for a Windows/Linux line-ending (CRLF) mismatch, but I will wait for the full high-level review.

- Secondly, the idea of a sorting mechanism sounds great! I'd be happy to help to the best of my abilities. The existing logic in `types.py` will most likely replace my current logic, thanks for pointing it out.

To conclude, I'll wait for your full review and then start the refactor. Thanks again and happy holidays!

JOJ0:
Hi Angelos,  
finally my review and ideas for a way forward:

I see two use cases here:

- Integrating sorting by musical key into Beets
- The actual `mix` subcommand, which is a helper to find and print all "compatible tracks"

## Sorting

- The Beets query language could be extended with new special symbols for sorting by musical key, the tricky part is to find symbols that are possibly not used by shells like `zsh`, `bash`, Windows shells? For example using symbols like `~` and `=` would have to be escaped with single quotes (or `\`) which makes usage tedious. Let's brainstorm here... 
- Sorting by bpm is already possible (`bpm+`, `bpm-`), we don't need to do anything here
- Or, and that might be easier, with a new config option in the plugin's config space, like eg. `circle_of_fifths_sort: yes/no`  the regular `+` and `-` symbols would simply change the default sorting (alphabetically) to a clockwise or counterclockwise sort in the circle. 
- 
## `mix` subcommand

- Use Beets queries to achieve the filtering by bpm range and musical key: https://beets.readthedocs.io/en/latest/api/generated/beets.dbcore.query.FieldQuery.html, you do that here already, perfect: https://github.com/beetbox/beets/pull/6241/changes#diff-c570acdaa6732f23998d861e11d8a535f2b2061c42832d30488342ec0704946dR140
- We don't have to invent a new type of output formatting (like you did with `MATCH: ...`, if we simply `print(matching_item)`, we get the output the user has configured. That output is according to how the user likes to see their tracks. If they'd desire that bpm and key is printed here, the would have configured it that way.
    - Now we could argue that for this purpose a different kind of output would be preferred, so why not implement a plugin-level custom formating option (configurable in config.yaml in the `harmonic_mix:` section, with a usable default) for the output format (see https://github.com/beetbox/beets/blob/master/beets/config_default.yaml#L146-L147)
- Last but not least, I love that a simple command name like `mix` is not occupied by Beets already :heart:, for the configuration I think it doesn't hurt if we'd be more explicit about it as I already hinted above (`harmonic_mix:`)



  

 ## Future ideas feedback
> Integrate calls to autobpm and keyfinder to fill in missing metadata.

I would not recommend this. Integrating with specific plugins is very opinionated and creates dependencies not every user might want, and it sounds like overcomplicating things anyway

> Make the BPM range configurable (currently fixed at +/- 8%).

This is easy to implement and I think would be good to include in the initial implementation. My personal preference for a default is +/- 6% since it works well on my Technics, but I'm fine with 8 if you prefer that :-)

semohr:
As a quick FYI, the [keyfinder cli](https://github.com/EvanPurkhiser/keyfinder-cli/) can output keys in open key and camelot notation as well. Since we cant be sure which software and configuration people use to create the keys, it might be worth supporting these formats too 😉

Exafto:
Hello everyone!
Thanks again for the warm welcome and the ideas for the plugin. Unfortunately, due to a tight schedule, I will not be able to see this PR through in the foreseeable future.
I’m happy for this to be closed, or for someone else to pick up the idea if they’re interested. Thanks again for the feedback, I wish I could have contributed more!

JOJ0:
It's ok, we appreciate you letting us know! I wish I had more time atm. I would love to continue work on this. Thanks again and all the best!

snejus:
I love the idea of this plugin and I hope someone can pick this up. For now I'm marking this PR as draft.

JOJ0:
@Exafto and @snejus I split out this idea of my write-up above (which IMO is a core behavior and shouldn't be part of a plugin) to a separate PR:

> Or, and that might be easier, with a new config option in the plugin's config space, like eg. `circle_of_fifths_sort: yes/no` the regular + and - symbols would simply change the default sorting (alphabetically) to a clockwise or counterclockwise sort in the circle.

See #6669 for my draft and let's move discussion around it over there. What do you think?

PR #6253, opened 2025-12-29
fetchart: Add support for configurable fallback cover art

dtrunk90:
## Description

Fixes discussion #6252.  

## To Do

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -1101,6 +1101,16 @@ class FileSystem(LocalArtSource):
                 else:
                     remaining.append(fn)
 
+            # Fall back to a configured image.
+            if plugin.fallback:
+                self._log.debug(
+                    "using fallback art file {}",
+                    util.displayable_path(plugin.fallback),
+                )
+                yield self._candidate(
+                    path=plugin.fallback, match=MetadataMatch.FALLBACK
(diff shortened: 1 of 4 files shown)


docs/plugins/fetchart.rst
```diff
@@ -121,6 +122,7 @@ others:
     fetchart:
         cautious: true
         cover_names: front back
+        fallback: /path/to/fallback.png
```
snejus: This should display the default configuration

```suggestion
        fallback:
```

beetsplug/fetchart.py
```diff
@@ -1101,6 +1101,17 @@ def get(
                 else:
                     remaining.append(fn)
 
+            # Fall back to a configured image.
+            if plugin.fallback:
```
snejus, line 1105: Would you mind adding a test?

dtrunk90: Sure. I can do that later.

docs/plugins/fetchart.rst
dtrunk90: Are you sure? It's an example. But after re-checking I should either remove it entirely from that specific example or add another separate example.

dtrunk90:
Comments and CI failures are addressed

PR #6257, opened 2026-01-01
drop extraneous dependency on old external "mock"

a-detiste:
This project uses `unittest.mock` from the standard library


snejus:
Nice spot - just use `uv tool poetry'<2' lock` to update `poetry.lock` as we use a slightly older poetry version.

a-detiste:
> $ uv tool poetry'<2' lock 
> error: unrecognized subcommand 'poetry<2'

That does not work here ...
You could also consider merging my PR with out the update of `poetry.lock`

> Nice spot

It's fully automated ;-) ... well I detect it the random day that the package is uploaded to Debian

image

jackwilsdon:
Thanks! I've gone ahead and updated the lock in d6da6cda7ebcfb8b999bf716c92bf0b5f41ad80c. It turns out the command is `uv tool run 'poetry<2' lock --no-update` to ensure nothing else changes.

PR #6259, opened 2026-01-02
fromfilename: improvements & fixes

### LINKED ISSUE #541: fromfilename: Add filename matching regex for more forms of leading numerals
At in issue #538 sampsyo suggested, I open a new ticket for asking to add a regex implementation in order for the FromFileName plugin do "understand" leading numeration in filenames.

Many users have files sorted with leading numeration, i.e.:

```
01 - Song name.mp3
01.-.Song.name.mp3
01-Song name.mp3
```

et similia.

This lowers A LOT the similarity in beets.

I think adding a regex could help this.

Something like:

```
^[0-9][0-9]*([:punct:]|[:space:])*
```

or even more simple

```
^[0-9][0-9]*[ ,\-=_.]*
```

could remove this problem.

Don't know if Python know these expressions:

```
[:punct:] = [!"#$%&'()*+, \-./:;<=>?@ [\\\]^_`{|}~]
[:space:] = [ \t\r\n\v\f]
```

I am unable to add this to the code, don't know Python :-(

Could someone think at adding something similar?

Thanks.

### LINKED ISSUE #1400: [suggest] Methods for 'manually importing' tracks (specifically those without matching sources)
My Recommendations for <i>manual editing</i> imported tracks
=

A) the use of fromfilename plugin (although I can't get it to work)
B) simple option to just 'edit'

<b>Option A</b>
Upon importing a track (with option to skip matching from sources, because there isn't any), this plugin will provide you the option to set the format-pattern to define which goes where.

Because <code>fromfilename</code> plugin doesn't have any 'configuration' option like other plugins have, there isn't a way to verify the format used within the filename to embed the tags from. 
But if you look at the <code>beetsplug/fromfilename.py</code> it has sections for <i>field 
extraction patterns</i> for which could possibly define the format-pattern in this following scenario:

File (w/out any tags):
01. Mack 187 & Akoza - Introduction.flac

Format:
$track. $artist - $title

<b>Option B</b>
same as above but by not defining the <code>fromfilename</code> this instead will allow you to edit the general tags yourself:
-artist
-title
-album
-track/total
-disc/total
-comment
...etc

as also suggested from #154 @dokterbob recommended an similar 'Apply and edit' option, except mine includes what could be edited and that it should be up to the user to define what they want to have the option to edit from.

### LINKED ISSUE #1554: Extend Fromfilename plugin: guess album name / album artist / year from parent directory name
I often find album without tags in this form:
` album artist - album (year)\track# artist - title`

Would be great if the fromfilename plugin could guess some info from the directory name the files are in.

### LINKED ISSUE #5413: fromFilename Plugin: Does not update tags with AsIs actions.
### Problem

Using the fromFilename plugin on music files with blank Title tags, and choosing "Use as-is" options, does not update the tags.

Source file:
image
```sh
$ beet import "D:\BeetsTest"
```
Imported file:
image

I need to use the "Use as-is" option because none of the MusicBrainz candidates match.
Passing -A does the same.
I want it to update the Title tag from the filename, without then choosing a candidate for incorrectly updating the tags. I'm not sure what the intended behavior of the plug-in is here.

I've managed to achieve this by adding the following to the end of `filename_task`, however this applies the tags to both the source file and the imported file.
```python
# Write the new tags now
for item in items:
    item.try_write()
```

I guess a better solution is to set the action of the `manipulate_files` task to `action.RETAG`, but I don't know how to edit subsequent tasks from the plug-in.

### Setup

* OS: Windows 11
* Python version: 3.12.3
* beets version: 2.0.0

```yaml
directory: D:\BeetsTest\Library
library: D:\BeetsTest\musiclibrary.db

ignore:
    - Library   # Being written to during import
    - Various Artists

plugins: fromfilename
```

### LINKED ISSUE #5597: FromFilename Plugin don't work at all
I have flac files without any tags but the folder and file naming are good.

The ARTIST and ALBUM is on musicbrainz and show up after manual search.

I don't know if there's need of more [PATTERNS](https://github.com/beetbox/beets/blob/bcc91ffff5a38794897e98f5284989039909717b/beetsplug/fromfilename.py#L26) or if it because the plugin don't consider the [track as empty](https://github.com/beetbox/beets/blob/bcc91ffff5a38794897e98f5284989039909717b/beetsplug/fromfilename.py#L39) ?

**Foldername:** `ARTIST - ALBUM`
**Filename:** `01 ARTIST - TRACK.flac `

**metaflac --list 01 ARTIST - TRACK.flac** 

```
METADATA block #0
  type: 0 (STREAMINFO)
  is last: false
  length: 34
  minimum blocksize: 4096 samples
  maximum blocksize: 4096 samples
  minimum framesize: 14 bytes
  maximum framesize: 10805 bytes
  sample_rate: 44100 Hz
  channels: 2
  bits-per-sample: 16
  total samples: 6920760
  MD5 signature: 
METADATA block #1
  type: 4 (VORBIS_COMMENT)
  is last: false
  length: 47
  vendor string: reference libFLAC
  comments: 0
METADATA block #2
  type: 1 (PADDING)
  is last: true
  length: 8192
```

**beet -vv import ARTIST  -  ALBUM**

```
user configuration: /home/user/.config/beets/config.yaml
data directory: /home/user/.config/beets
plugin paths: /home/user/git/beets/beetsplug
Sending event: pluginload
library database: /home/user/.config/beets/beets.db
library directory: /home/user/musik2
Sending event: library_opened
Sending event: import_begin
fromfilename: Title replaced with: ARTIST - TRACK
fromfilename: Track replaced with: 1
Looking up: /home/user/musik2/ARTIST - TRACK
Tagging  - 
No album ID found.
Search terms:  - 
Additional search terms: {'year': 0, 'label': '', 'catalognum': '', 'country': '', 'media': ''}
Album might be VA: True
Evaluating 0 candidates.

/home/user/musik2/ARTIST - TRACK (1 items)
Sending event: import_task_before_choice
Sending event: before_choose_candidate
No matching release found for 1 tracks.
For help, see: https://beets.readthedocs.org/en/latest/faq.html#nomatch
➜ [S]kip, Use as-is, as Tracks, Group albums, Enter search, enter Id, aBort? B
Sending event: import
Sending event: cli_exit
```

Important here are the two line 
```
fromfilename: Title replaced with: ARTIST - TRACK
fromfilename: Track replaced with: 1
```
ARTIST - TRACK arent't the Title(TRACK) of course so the plugin don't even interest in ARTIST!?

### Setup

beets version 1.6.1
Python version 3.13.1
plugins: fromfilename

```yaml
library: ~/.config/beets/beets.db
directory: ~/musik2/
threaded: yes
import:
    write: yes
    copy: no
    resume: no

musicbrainz:
    enabled: yes
    host: musicbrainz.org
    https: yes
    extra_tags: [year, catalognum, country, media, label]

plugins: fromfilename

pluginpath:
    - /home/user/git/beets/beetsplug/
```

Thanks !

henry-oberholtzer:
This pull request improves the usability and configurability of the `fromfilename` plugin to encompass more file name patterns and folder names, which can contain usable information that will lead to better results when put through the metadata plugins, even for files with partial tags.

These changes gives the plugin the ability, by default, to extract the following fields:

`album`, `artist`, `albumartist`, `disc`, `track`, `title`, `media`, `catalognum` , `year`

A comprehensive set of tests has been written to make sure the patterns can recognize the information in a wide
variety of layouts and file names.

For more unusual folder and filename patterns, users can now specify a pattern, using the same method as our
path format strings, e.g. "$albumartist - $album ($comments)"

## Changes & Issues Addressed:
- [x] Full typing
- [x] Increased test coverage with more edge case handling
- [x] Make sure single files can also be checked for information
- [x] Precompile regex expressions
- [x] Expanded patterns from track information #5597
- [x] Default patterns with better track index parsing, supporting disc and track combinations. #541
- [x] Default behavior extracts information from the folder name (album artist, year, album title) #1554
- [x] Allow users to contribute additional search patterns & fields #1400 
- [x] Configurable parent folder that can be ignored for imports
- [x] Improved log message with tags being updated.
- [x] Do not consider albums if `--group-album` flag is used
- [x] fromfilename ignores folder when "group by album" or "group by tracks" is used.
- [x] Write fromefilename data when apply is chosen, reflecting metadata displayed to user  #5413~
- [x] Add a menu option to re-search without from filename.

## Notes:
- [Removed the "tag" field from the plugin](https://github.com/beetbox/beets/blob/afc26fa58f15a18f1d672eb1b2432027d7b6ad35/beetsplug/fromfilename.py#L30C9-L30C34), as its use and applications were vague, undocumented, and untested.

## Housekeeping
- [x] Documentation
- [ ] Changelog
- [x] Tests


beetsplug/fromfilename.py
```diff
@@ -16,153 +16,501 @@
...
+            if m := self._check_user_matches(filename, self.file_patterns):
+                track_matches[item] = m
+            else:
+                match = self._parse_track_info(filename)
+                track_matches[item] = match
+        return track_matches
+
+    @staticmethod
+    def _parse_track_info(text: str) -> FilenameMatch:
+        match = RE_TRACK_INFO.match(text)
```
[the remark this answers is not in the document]
henry-oberholtzer: The regular expression always returns a match.

docs/plugins/fromfilename.rst
```diff
@@ -5,8 +5,82 @@ The ``fromfilename`` plugin helps to tag albums that are missing tags altogether
...
+
+    If a field is specified in a user pattern, that field does not need
+    to be present on this list to be applied.
+
+    If you only want the plugin to contribute the track title and artist,
+    you would put ``[title, artist]``.
+
+.. conf:: patterns
+
+    Users can specify patterns to improve the efficacy of the plugin. Patterns can
```
JOJ0:
```suggestion
    Users can specify patterns to improve the efficiancy of the plugin. Patterns can
```

beetsplug/fromfilename.py
```diff
@@ -16,153 +16,499 @@
...
-    return matches
+# Filename field extraction patterns
+RE_TRACK_INFO = re.compile(
+    r"""
+    (?P<disc>\d+(?=[\.\-_]\d))?
+        # a disc must be followed by punctuation and a digit
+    [\.\-]{,1}
+        # disc punctuation
+    (?P<track>\d+)?
+        # match the track number
```
JOJ0, line 38: would it make sense to also match vinyl tracknumbers like A1, B2, C, D, ... ?

I come over files named like that often.

beetsplug/fromfilename.py
```diff
@@ -16,153 +16,499 @@
...
+        # does not start with file format
+    (?P<catalognum>[\w\s]+)
+        # actual catalog number
+    (?<!flac|.mp3|.wav)
+        # does not end with file format
+    [\)\]\}]
+        # ends with a bracket
+    """,
+    re.VERBOSE | re.IGNORECASE,
+)
```
JOJ0, line 71: finally! catalog numbers 🤩 thanks!

JOJ0 (changes requested):
just some tiny things I saw during a quick scroll through. this is a lot! but worth doing! thanks! I came along a bunch of old badly tagged mp3's lately that will be perfect for testing. hope I get to it.

docs/plugins/fromfilename.rst
henry-oberholtzer: I think efficacy is the right word here - the patterns improve how effective it is at finding the right fields, not necessarily how efficient it is.

beetsplug/fromfilename.py
henry-oberholtzer: I was thinking about that - I think the difficulty might be in figuring out how to distinguish between an index and a track title. Something like `A song title.wav` could be side A, song title, or 'a song title'. Numbers at least are pretty distinct. I'll mess around to see if I can add support for it though, may need rely more on there being hyphen diving the track index and title than for the numbers.

docs/plugins/fromfilename.rst
JOJ0: Oh sorry, then my english simply was too bad. And I just learned a new word 😂

henry-oberholtzer: I simplified the language anyway since there's no reason for our docs to be too complicated.

beetsplug/fromfilename.py
henry-oberholtzer: OK, I've implemented this. It goes before user pattern checks, but only checks for patterns that are up to two capital letters and a number at the start of a string!

henry-oberholtzer:
I've been using it a good amount with my vinyl ripping workflow, and it's been working way better than the old version. If everything looks good & works well for you, I can update the changelog & get it merged in.

JOJ0:
> I've been using it a good amount with my vinyl ripping workflow, and it's been working way better than the old version. If everything looks good & works well for you, I can update the changelog & get it merged in.

I merged it into my personal master and want to check it alongside current discogs changes. I got distracted with a crash in Beets elsewhere though.... Will try to do some testing with badly tagged files as promised soonish.

One thing I wanted to ask: How did you come up with all the regex patterns? 1. By heart since I'm a regex wizard. 2. Had AI generate them. 3. Used regex101 (or similar) to develop and test them. :-)))

JOJ0 (commented):
On a very high level, the first things I notice:

Given a badly tagged collection of Alternative Rock/Pop music residing in a folder named `Alternative` we find a hand full of single files as well as subfolders with artist names or even subfolders with full albums.

This might be out of scope with your suggested changes but since you are kind of the expert in refactoring this plugin at the moment I'd like to feedback anyway:

First beets tries to process the 7 files in the root of the collection as an album, thus I tell it to "Group Albums":

```
/......./Alternative (7 items)
...
...
fromfilename: Item updated with: dict_items([])
fromfilename: Item updated with: dict_items([])
fromfilename: Item updated with: dict_items([('album', 'Alternative')])
fromfilename: Item updated with: dict_items([('album', 'Alternative')])
fromfilename: Item updated with: dict_items([])
fromfilename: Item updated with: dict_items([('album', 'Alternative')])
fromfilename: Item updated with: dict_items([('album', 'Alternative')])
...
...
Enter search, enter Id, aBort, eDit, edit Candidates, plaY? g
...
...
fromfilename: Item updated with: dict_items([])
Looking up: /...Alternative/caesars+palace+-+jerk+it+out.mp3
```

In this case what does the "Item updated with " message mean exactly? Note that this single track actually does have an artist, album and title tag set.

On a next file i see:

In this case updating the item album Alternative is a bad idea since we just asked the importer to "Group Albums" and to not consider the containing folder's name as the album title (or I am mistaken and this is how Beets works even in "Group albums" mode. Let's investigate....).

```
fromfilename: Item updated with: dict_items([('album', 'Alternative')])
Looking up: /...Alternative/Elastica - Connection.mp3
```


This next one is probably impossible for the plugin, so we might just ignore that weird namings like that can't really be considered:

```
fromfilename: Item updated with: dict_items([('album', 'Alternative')])
Looking up: /.../Alternative/Stigmata Soundtrack - 01_Mary Mary (Stigmatic Mix) - Chumbaw.mp3
```

But also here we see that the missing album is replaced with the folder name which in this case doesn't improve the metadata search request. Also note that this file had a correct artist and title tag. Album was missing though.

```
/...Alternative/Stigmata Soundtrack - 01_Mary Mary (Stigmatic Mix) - Chumbaw.mp3 (1 items)
Sending event: import_task_before_choice
Sending event: before_choose_candidate

Finding tags for album "Chumbawamba - Alternative".
  Candidates:
```

I'm stopping now since I'm running out of time but will continue with this folder. There is loads of difficult material in it ;-)

Please don't take this too critical. Just trying to give a real life example and I'm sure there might be situations were no automation on this world will prevent a manual intervention, eg. search/rename/restructure.

One thing I always wanted to have in Beets is the ability to edit what is actually sent to the metadata source withihn the importer. This way an only halfway correct suggestion from filename or wonky tags could be quickly fixed. Similar to the edit plugin, but before the search, but this is out of scope here!

So the main thing I notice and question is: Is it expected that "non-album-folders" names will be considered as album names in "Group Albums" mode, and can we do anything about it?

Anyway, hope it helps, and again: Love that you took on the task of making this plugin more useful! I'll try to do more playing around this weekend I hope!!!

henry-oberholtzer:
Woohoo! Thank you for this very thorough and constructive feedback. The rewrite was mainly done through my own lense, so there's a lot of scenarios I may have missed or things I didn't catch.

Regex was a combination of regex101 and by heart. I'm not gonna let a chatbot have all the fun writing regex!! I unfortunately(?) enjoy the process of writing out all the code myself, so unless someone else requires me to, I tend to stay away from LLMs with my code.

```
In this case what does the "Item updated with " message mean exactly? Note that this single track actually does have an artist, album and title tag set.
```
Good issue raised here - this log message could be more clear. It essentially shows the fields that it thinks are empty and it should substitute. Since all the files are lacking the album tag, it's using "Alternative" as the album tag.

```
One thing I always wanted to have in Beets is the ability to edit what is actually sent to the metadata source withihn the importer. This way an only halfway correct suggestion from filename or wonky tags could be quickly fixed. Similar to the edit plugin, but before the search, but this is out of scope here!
```
This would be so useful! Maybe it could be an expansion of the edit plugin, since it's sort of on the right track, and could stand to allow more fields to be edited too.

Here are the changes I think I 'll make to have it perform better that will hopefully

1.I think I will restructure that log message to the following:
```
fromefilename: guessing "filename.mp3" album="Alternative"
```
This should be more instructive as to what it is actually doing with each file.

2. I'll also make sure it respects the group album command or doesn't try to update the album field if something else in the group has an album tag already. I'll also see if it can check to see if the `--group-albums` flag is thrown and then disable the parent folder guess by default.

3. I think adding just a configuration toggle to not guess the parent folder would be a good option, in addition to the `ignore_folder` command.

Thanks for this great feedback - would much rather get this right the first time than have to keep patching it.

JOJ0:
> Regex was a combination of regex101 and by heart. I'm not gonna let a chatbot have all the fun writing regex!! I unfortunately(?) enjoy the process of writing out all the code myself, so unless someone else requires me to, I tend to stay away from LLMs with my code.

Haha! Yes! That's the spirit! I'm lazy often I have to admit and let AI help me, but especially with regex I had bad experiences and regex101 and human thinking is the only way! Brings back one thought I'd like to throw in in case you are not aware already: Putting more complex regex into yaml can be a PITA, which is the case here with the custom patterns you implemented. If people try to do crazy black magic with this feature we might get issues in the future. Didn't look how you are escaping these patterns. Will do in my final low level review but I'll let you continue first withe below's awesome feature ideas! ❤️ 

> One thing I always wanted to have in Beets is the ability to edit what is actually sent to the metadata source withihn the importer. This way an only halfway correct suggestion from filename or wonky tags could be quickly fixed. Similar to the edit plugin, but before the search, but this is out of scope here!
> ```
> 
> This would be so useful! Maybe it could be an expansion of the edit plugin, since it's sort of on the right track, and could stand to allow more fields to be edited too.

Let's definitely keep that in mind for the future! Yeah an edit plugin PR sounds doable!

> Here are the changes I think I 'll make to have it perform better that will hopefully
> 
> 1.I think I will restructure that log message to the following:
> 
> ```
> fromefilename: guessing "filename.mp3" album="Alternative"
> ```
> 
> This should be more instructive as to what it is actually doing with each file.

That sounds super clear, explains everything and is still quite concise! ❤️ 

> 2. I'll also make sure it respects the group album command or doesn't try to update the album field if something else in the group has an album tag already. I'll also see if it can check to see if the `--group-albums` flag is thrown and then disable the parent folder guess by default.

Please help me out here: When we set it on the CLI already it will be enabled throughout the entire import run.? No matter in which folder a file resides, the importer will always try to mere everything together right?

Now if I enable it interactively _during_ the importer run (like in my example), only for this one folder (and it's subfolders???), will the group albums feature _unset_ itself when we jump out of the folder again. Eg. in my example, once we `cd ..` from `Alternative/` again and move on with the next top level folder. 

_Or_ will it stay enabled always from that point on, in the import run? Sorry, confused how Beets actually works in that matter 🤷 

> 3. I think adding just a configuration toggle to not guess the parent folder would be a good option, in addition to the `ignore_folder` command.

Hmm if that is easy to implement, do it. Also add a CLI option for it then. I'm  picturing a workflow where you `beet import ....` and point it to a folder where you are sure is not sorted in folders by albums but only for example with artist names folders `Aritst1/`, `Artist2/`. Having to reconfigure sounds tedious, better quickly switching mode via CLI option sounds more user friendly.

What I missed in my first read-through is that you actually already thought of disabling this in a way:

https://github.com/beetbox/beets/pull/6259/changes#diff-dc32f40dc487a7e101c9189fedc8269826b4d92300b58c8646037a7be24cf51eR84-R86

So with 

> in addition to the `ignore_folder` command ...

you tried to point to this feature? Right?

This is a great addition, let's try not ot overengineer it but brainstorm a little. How would a CLI option look like that tells the importer (well actually the fromfilename plugin) that for this run the guessing by foldernames for albumnames shouldn't be done. Something like `beet import --no-fromfilename-folder-guess` sounds a bit superweird. I'll think about this more....drink more coffee first... ;-)

> Thanks for this great feedback - would much rather get this right the first time than have to keep patching it.

Dozens of thanks going out here! This will be so so super helpful!

henry-oberholtzer:
From what I can find - detecting and adjusting with "Group Albums" is really easy. It's just stored in the import config. All it has to do is check if group_albums is true and then it will skip the folder name. I'll need to test it all, but I think it should be that easy. Drafting out some tests for it. Can probably avoid a CLI option if it works.

JOJ0:
> From what I can find - detecting and adjusting with "Group Albums" is really easy. It's just stored in the import config. All it has to do is check if group_albums is true and then it will skip the folder name. I'll need to test it all, but I think it should be that easy. Drafting out some tests for it. Can probably avoid a CLI option if it works.

friendly reminder! If you have something put together locally alread, push it in an "experimental" commit or something, I can help with testing :-)

henry-oberholtzer:
OK! It's been a busy few months for me - but I'm gonna pick this up again, hopefully get it all set up & good to go.

henry-oberholtzer:
Re-evaluating my previous work, I also want to move my regex definitions from being hard coded to being part of the configuration.

This will allow users to alter them to their liking, as well as make it easy to test different patterns.

henry-oberholtzer:
@JOJ0 I've applied the changes we were talking about months(!) ago, and now it works a whole lot better. It doesn't check the folder information if you're grouping as tracks or by album, and can restore the original tags of the files when you select `toggle FromFilename` in the menu!

The plugin also writes the tags it found when you select importing As-Is, since that follows the intuitive pattern of what it shows you in the UI.

I've tried to test it pretty thoroughly, and it's working very well, but there are probably some bugs I've overlooked or haven't noticed.

snejus:
Would it be possible for you to rebase this on top of #6940 ? I will have a look at it after!

henry-oberholtzer:
> Would it be possible for you to rebase this on top of #6940 ? I will have a look at it after!

I think with this having become a 100% rewrite of the initial plugin at this point - there's barely anything left of the original, and it is fully typed in this rewrite, so I'm not sure there's a lot to be gained rebasing off of 6940 until that's merged in? Let me know if I'm not understanding your reasoning though.

snejus:
I started looking at it... Well done mister 👍🏼 however, I'm struggling to review it due to size. Would it be possible to somehow split it? A complete rewrite + 5 fixed issues means that we will have a lot of comments.

henry-oberholtzer:
I'll see what I can do!

PR #6273, opened 2026-01-07
Fix fetchart colors broken by 67e668d81ff03d7ce14671e68676a7ad9d0ed94a

aereaux:
## Description

Fetchart colors were broken by 67e668d81ff03d7ce14671e68676a7ad9d0ed94a.  Since that commit fetchart displays this, with escape codes instead of colorized output:
```
fetchart: Alasdair Fraser & Natalie Haas - Meridians: �[37mhas album art�[39;49;00m
```

This fixes it by using `ui.print_` instead of `Logger`.  This seems to be in line with other plugins such as play:
https://github.com/beetbox/beets/blob/1a899cc92af57b72a76570fb37d1c4eda0f5e842/beetsplug/play.py#L135

## To Do

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -1588,7 +1588,7 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
                     message = ui.colorize(
                         "text_highlight_minor", "has album art"
                     )
-                    self._log.info("{}: {}", album, message)
+                    ui.print_(f"{album}: {message}")
             else:
                 # In ordinary invocations, look for images on the
                 # filesystem. When forcing, however, always go to the Web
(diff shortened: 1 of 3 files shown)


beetsplug/fetchart.py
```diff
@@ -1588,7 +1588,7 @@ def batch_fetch_art(
                     message = ui.colorize(
                         "text_highlight_minor", "has album art"
                     )
-                    self._log.info("{}: {}", album, message)
+                    ui.print_(f"{album}: {message}")
```
snejus, line 1591: Would it be possible to test this?

aereaux: Sure!  How's this?

test/plugins/test_fetchart.py
```diff
@@ -98,3 +98,12 @@ def test_filesystem_picks_up_hidden_file(self):
...
         self.check_cover_is_stored()
+
+    def test_colorization(self):
+        self.config["ui"]["color"] = True
+        with capture_stdout() as logs:
+            self.run_command("fetchart")
+        assert (
+            " - the älbum: \x1b[1;31mno art found\x1b[39;49;00m\n"
+            == logs.getvalue()
+        )
```
snejus: Tip: consider using `run_with_output` method instead 😉

aereaux: That's helpful, didn't notice that before.  Changed to use that

PR #6277, opened 2026-01-09
Suggested improvements to discogs fixes

snejus:
Summary
- Refactors artist/tracklist construction into dataclasses to isolate parsing logic:
  - Introduces `ArtistState` (with `ValidArtist`, `from_plugin`, properties like `artists`, `artist_id`, and `info`) and `TracklistState` to encapsulate artist- and tracklist-building behavior in `beetsplug/discogs.py`.
  - Moves logic previously spread across `_build_artistinfo`/`_assign_anv`/`_process_clean_tracklist` into those dataclasses, and removes duplicated imperative code in the plugin class.

- Simplifies plugin flow and responsibilities:
  - `DiscogsPlugin` now delegates artist/tracklist assembly to `ArtistState.from_plugin` and `TracklistState.build`, reducing the plugin surface and concentrating transformation rules in typed data objects.
  - Combines track `artists` and `extraartists` handling into a single `ArtistState` construction so featured/credit/anv logic is applied consistently.

- Tests updated for correctness and clarity:
  - `test/plugins/test_discogs.py` adds an `_artist` factory to ensure artist dicts in tests always include required fields, and updates `test_parse_featured_artists` to include a certain edge case.
  - Tests implementation to use `ArtistState.from_plugin` where appropriate.

### DIFF
--- a/beetsplug/discogs.py
+++ b/beetsplug/discogs.py
@@ -26,9 +26,9 @@ import socket
 import time
 import traceback
 from dataclasses import asdict, dataclass, field
-from functools import cache
+from functools import cache, cached_property
 from string import ascii_lowercase
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, NamedTuple
 
 import confuse
 from discogs_client import Client, Master, Release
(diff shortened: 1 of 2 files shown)


henry-oberholtzer:
I'll take a thorough look at this later today - looks like some good simplification going on here. The discogs plugin has to do so much to make their (otherwise very useful) database compatible with beets, so I think moving parsing to another class is a good move.

test/plugins/test_discogs.py
```diff
@@ -520,28 +526,21 @@ def test_anv(
...
+                        anv="PERF",
+                    )
                 ],
             }
         ],
         "artists": [
-            {"name": "DRUMMER", "anv": "DRUM", "id": 445, "join": ", "},
-            {"name": "ARTIST (4)", "anv": "ARTY", "id": 321, "join": "&"},
-            {"name": "SOLOIST", "anv": "SOLO", "id": 445, "join": ""},
+            _artist("DRUMMER", id=445, anv="DRUM", join=","),
```
henry-oberholtzer: Small note that inclusion of the space after the comma is intentional for this test. Discogs is very bad about not removing trailing spaces from separators. For example, currently [this release](https://www.discogs.com/release/24736595-Dica-4-Wast-Beat-Jugglers-Feat-MC-Evenson-Allen-Stamina-Vibezin) has a trailing space on "Feat." that Discogs renders in some contexts, like a collection view, but not in others, but will come through on an API request. 


image

image

beetsplug/discogs.py
```diff
@@ -133,6 +125,160 @@ class TracklistInfo(TypedDict):
...
+    class ValidArtist(NamedTuple):
+        id: str
+        name: str
+        credit: str
+        join: str
+        is_feat: bool
+
+        def get_artist(self, property_name: str) -> str:
+            return getattr(self, property_name) + (
+                {",": ", ", "": ""}.get(self.join, f" {self.join} ")
```
henry-oberholtzer: I think some comments on these data classes would be great, explaining what their role is in the plugin - it'll avoid those who may follow having to sort out the layers.

henry-oberholtzer (changes requested):
Just tested this out on my fork - works really neatly! I think we do need to make sure that we're handling space padded joins from Discogs, that's one of their annoying inconsistencies we have to handle.

I also wonder if we could move more of the track logic into the track dataclass, but I don't think it's necessary at the moment.

Thanks for building on and refining my initial go at this, I think this plugin is really near matching the musicbrainz plugin in terms of features.

test/plugins/test_discogs.py
snejus: Ah yeah, I see this [here](https://api.discogs.com/releases/24736595): album artist has `join: "Feat, "` while artist has `join: "Feat,"`. 🙄 


By the way, what a coincidence: that's my favourite club FOLD (and the only one I rave at nowadays) in the video 🤯 
image

henry-oberholtzer: The more I engage with the Discogs database the more of a mess I realize it is - but it also has more releases than Musicbrainz and actually always responds to API requests for me... Happy to pick up this PR from you again and patch in the spacing and write some docstrings for the new classes.

Whenever I make it to the UK again I definitely want to go to Fold - all the Stay Up Forever acid techno DJs I'm into play it all the time!

snejus: Addressed both before I saw your message!

henry-oberholtzer (approved):
This all looks good to me now - happy to have this merged in if you're good with where it's at!

I think there isn't much left to do in order to get our discogs tagging as effective and accurate as the Musicbrainz search now!

Issue #6256, opened 2025-12-31
Import crashes with `TypeError: 'NoneType' object is not subscriptable` for Musicbrainz plugin


Nedra1998:
### Problem

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv import --move /import
```

Led to this problem:

```
Traceback (most recent call last):
  File "/usr/bin/beet", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/lib/python3.12/site-packages/beets/ui/__init__.py", line 1629, in main
    _raw_main(args)
  File "/usr/lib/python3.12/site-packages/beets/ui/__init__.py", line 1608, in _raw_main
    subcommand.func(lib, suboptions, subargs)
  File "/usr/lib/python3.12/site-packages/beets/ui/commands/import_/__init__.py", line 131, in import_func
    import_files(lib, byte_paths, query)
  File "/usr/lib/python3.12/site-packages/beets/ui/commands/import_/__init__.py", line 75, in import_files
    session.run()
  File "/usr/lib/python3.12/site-packages/beets/importer/session.py", line 236, in run
    pl.run_parallel(QUEUE_SIZE)
  File "/usr/lib/python3.12/site-packages/beets/util/pipeline.py", line 471, in run_parallel
    raise exc_info[1].with_traceback(exc_info[2])
  File "/usr/lib/python3.12/site-packages/beets/util/pipeline.py", line 336, in run
    out = self.coro.send(msg)
          ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/beets/util/pipeline.py", line 219, in coro
    func(*(args + (task,)))
  File "/usr/lib/python3.12/site-packages/beets/importer/stages.py", line 146, in lookup_candidates
    task.lookup_candidates(session.config["search_ids"].as_str_seq())
  File "/usr/lib/python3.12/site-packages/beets/importer/tasks.py", line 367, in lookup_candidates
    autotag.tag_album(self.items, search_ids=search_ids)
  File "/usr/lib/python3.12/site-packages/beets/autotag/match.py", line 319, in tag_album
    for matched_candidate in metadata_plugins.candidates(
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/beets/plugins.py", line 559, in wrapper
    for v in func(*args, **kwargs):
             ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #6279, opened 2026-01-10
Fix bug in fetching preferred release event.

djl:
With the changes to how data is fetched from MusicBrainz, empty releases are now `None` instead of an empty dict.

### DIFF
--- a/beetsplug/musicbrainz.py
+++ b/beetsplug/musicbrainz.py
@@ -247,8 +247,9 @@ def _preferred_release_event(
     for country in preferred_countries:
         for event in release.get("release-events", {}):
             try:
-                if country in event["area"]["iso-3166-1-codes"]:
-                    return country, event["date"]
+                if area := event.get("area"):
+                    if country in area["iso-3166-1-codes"]:
+                        return country, event["date"]
             except KeyError:
                 pass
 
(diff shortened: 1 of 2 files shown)


Nedra1998:
I tested the change, and it looks it solves my issue with #6256, so thanks for that.

snejus:
Nice, thanks for this! Would you mind extending the tests to make sure that this issue is caught in the future? 

It should be as simple as including these two lines:

```diff
diff --git a/test/plugins/test_musicbrainz.py b/test/plugins/test_musicbrainz.py
index 199b62ab6..733287204 100644
--- a/test/plugins/test_musicbrainz.py
+++ b/test/plugins/test_musicbrainz.py
@@ -29,6 +29,7 @@ class MusicBrainzTestCase
     def setUp(self):
         super().setUp()
         self.mb = musicbrainz.MusicBrainzPlugin()
+        self.config["match"]["preferred"]["countries"] = ["US"]
 
 
 class MBAlbumInfoTest(MusicBrainzTestCase):
@@ -80,6 +81,7 @@ def _make_release
             "country": "COUNTRY",
             "status": "STATUS",
             "barcode": "BARCODE",
+            "release-events": [{"area": None, "date": "2021-03-26"}],
         }
 
         if multi_artist_credit:
```

djl:
Not sure how I missed the test updates. Have pushed the changes now.

PR #6280, opened 2026-01-10
Enable duplicate detection for as-is (autotag: no) imports

a544jh:
## Description

  > **Note**: This fix was developed with assistance from Claude Code (AI). The problem was identified by me, and Claude helped investigate the codebase, trace the git history to find the original FIXME, implement the fix, and update the tests. All changes have been reviewed and tested.

  When importing with `autotag: no`, duplicate detection is completely bypassed. The `import_asis` stage calls `_apply_choice()` directly without first calling `_resolve_duplicates()`, meaning any configured `duplicate_keys` and `duplicate_action` settings are ignored.

  This was a known limitation. Commit 79d1203541 (Sep 2014) added a FIXME comment:

  ```python
  # FIXME We should also resolve duplicates when not
  # autotagging. This is currently handled in `user_query`
  ```

  The FIXME was removed during a comment cleanup in f145e3b18 (Jan 2015), but the underlying issue was never fixed. A test `test_no_autotag_keeps_duplicate_album` was added to document the existing behavior at the time.

  ### The Fix
  Add `_resolve_duplicates(session, task)` to the `import_asis` stage before `_apply_choice()`, matching the behavior of the `user_query` stage used when autotagging.

  ### Test Changes
  - Renamed `test_no_autotag_keeps_duplicate_album` → `test_no_autotag_removes_duplicate_album`
  - Fixed the test to use album metadata instead of item metadata for duplicate matching
  - Added missing `import_file.save()` call

  ## To Do

  - [x] ~Documentation~ (Bugfix for existing documented config options)
  - [ ] Changelog
  - [x] Tests

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -44,15 +44,18 @@ Bug fixes
 - :doc:`plugins/zero`: When the ``omit_single_disc`` option is set,
   ``disctotal`` is zeroed alongside ``disc``.
 - :doc:`plugins/fetchart`: Prevent deletion of configured fallback cover art
-- In autotagging, initialise empty multi-valued fields with ``None`` instead of
-  empty list, which caused beets to overwrite existing metadata with empty list
-  values instead of leaving them unchanged. :bug:`6403`
+- :ref:`import-cmd` When autotagging, initialise empty multi-valued fields with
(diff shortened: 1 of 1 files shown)


snejus:
Looks good to me. Would you mind adding a note to the changelog?

PR #6283, opened 2026-01-12
fix(fetchart): prevent deletion of configured fallback cover art

dtrunk90:
## Description

When `import.delete` or `import.move` is enabled, the `assign_art` method calls `task.prune(candidate.path)` unconditionally. This incorrectly deletes the configured `fetchart.fallback` file. Add explicit check to skip pruning when the candidate path matches the configured fallback.

## To Do

- [x] Documentation. (If you've added a new command-line flag, for example, find the appropriate page under `docs/` to describe it.) - not required because it's a fixup for 9ddddf4c3948d8405b7a9e4da761005d456e3aa9

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -1446,6 +1446,16 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
             "move"
         ].get(bool)
 
+    def _is_candidate_fallback(self, candidate: Candidate) -> bool:
+        try:
+            return (
+                candidate.path is not None
+                and self.fallback is not None
+                and os.path.samefile(candidate.path, self.fallback)
+            )
+        except OSError:
+            return False
+
     # Asynchronous; after music is added to the library.
     def fetch_art(self, session: ImportSession, task: ImportTask) -> None:
         """Find art for the album being imported."""
@@ -1494,7 +1504,7 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
 
             self._set_art(task.album, candidate, not removal_enabled)
 
-            if removal_enabled:
+            if removal_enabled and not self._is_candidate_fallback(candidate):
                 task.prune(candidate.path)
 
     # Manual album art fetching.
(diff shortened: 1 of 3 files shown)


dtrunk90:
@snejus Any chance to get this into the next patch release?

beetsplug/fetchart.py
```diff
@@ -1445,6 +1445,16 @@ def _is_source_file_removal_enabled() -> bool:
...
         ].get(bool)
 
+    def _is_candidate_fallback(self, candidate: Candidate) -> bool:
+        try:
+            return (
+                candidate.path is not None
+                and self.fallback is not None
+                and os.path.samefile(candidate.path, self.fallback)
+            )
+        except OSError:
```
snejus, line 1456: When would this possibly raise an `OSError?`

snejus (commented):
Apologies, only noticed your PR now. Just one question here

snejus:
Note you will want to resolve the conflict in the changelog and move your note under the Unreleased section.

beetsplug/fetchart.py
dtrunk90: `os.path.samefile()` works by calling `os.stat()` on both paths internally and comparing the results (inode number and device).
According to the [Python docs](https://docs.python.org/3/library/os.path.html#os.path.samefile), it raises an `OSError` if an `os.stat()` call on either path fails which can happen in several realistic scenarios:

1. The file no longer exists: `candidate.path` and/or `self.fallback` could have been moved/deleted between being resolved and this check. `self.fallback` could point to a misconfigured or deleted path.
2. Permission denied: The process may not have read access to the file's metadata.
3. Broken symlink: If either path is a symlink pointing to a non-existent target, `os.stat()` will fail.
4. Race condition: Even with `os.path.exists()` guards, there's a TOCTOU (time-of-check-time-of-use) window between the existence check and the actual samefile call where the file could disappear (e.g., concurrent import process, external cleanup tool).

The `try/except OSError` wrapping samefile is therefore the correct approach, as recommended by the Python docs themselves for this function.
It safely handles all the above cases by defaulting to False (i.e., don't treat it as the fallback file), which is the safe/conservative behavior here.

dtrunk90:
Done and ready to be merged :-)

snejus (approved):

PR #6292, opened 2026-01-15
Gracefully handle 404s when importing from MusicBrainz.

djl:
A 404 error can be raised when fetching from MusicBrainz in the case of re-importing an album that has since been deleted from MusicBrainz.

- [ ] Documentation. (N/A)
- [ ] Changelog. (N/A)
- [x] Tests.

### DIFF
--- a/beets/importer/tasks.py
+++ b/beets/importer/tasks.py
@@ -680,6 +680,8 @@ class SingletonImportTask(ImportTask):
         return [self.item]
 
     def apply_metadata(self):
+        if config["import"]["from_scratch"]:
+            self.item.clear()
         autotag.apply_item_metadata(self.item, self.match.info)
 
     def _emit_imported(self, lib):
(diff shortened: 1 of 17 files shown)


test/plugins/test_musicbrainz.py
```diff
@@ -1106,3 +1109,34 @@ def test_candidates(self, monkeypatch, mb):
         assert len(candidates) == 1
         assert candidates[0].tracks[0].track_id == self.RECORDING["id"]
         assert candidates[0].album == "hi"
+
+    @responses.activate
+    def test_import_handles_404_gracefully(self, mb):
```
snejus: Please use `requests_mock` fixture: you will find some examples in `test_lyrics.py` tests.

snejus (changes requested):

test/plugins/test_musicbrainz.py
djl: Done.

PR #6299, opened 2026-01-17
musicbrainz: fix release type

snejus:
- Update the MusicBrainz integration to read the release group's canonical `'primary-type'` field instead of the legacy `'type'` field when deriving `info.albumtype`.

Noticed that `albumtype` was missing from the metadata when I did some imports in my library.


snejus:
@sourcery-ai review

PR #6303, opened 2026-01-19
fix(lastgenre): Canonicalize keep_existing fallback

Nukesor:
## Description

Fixes a bug where existing tags were set to `None`, if they weren't whitelisted, but an whitelisted canonicalized parent existed up the tree.

In all other cases, the original genres are canonicalized and considered for the final genre, except in the `keep_existing` logic branch.
This PR fixes the issue and results in the expected behavior for this combination of options.

## How it works

For the bug to trigger, several conditions had to be met:

- Canonicalization is enabled and a whitelist is specified.
- `force` and `keep_existing` are set. Meaning, that Lastfm is queried for a genre, but the existing genres are still left around when none are found online.
- A release with a non-whitelisted genre exists, but that genre has a whitelisted genre parent up the tree.
- That very release has no genre on lastfm.

This is rather convoluted, but stay with me :D
What would happen is the following:

- `keep_genres` is set to the existing genres, as `force` and `keep_existing` is set.
- Genres for `track`/`album`/`artist` aren't found for this release, as they don't exist in lastfm.
- Then the `keep_existing` logic is entered.
  - The old logic only checks if the existing genres have an **exact** match for the whitelist. In contrast to all other code branches, we don't do the `_try_resolve_stage` in case there's no direct match, resulting in no match.
- We continue to the fallback logic, which returns the fallback (`None` in my case)

This patch results in one last try to resolve the existing genres when `keep_existing` is set, which includes canonicalization (if enabled).

---

Cheers

## To Do

- [x] Documentation. (Probably a bug and thereby not needed)
- [x] Changelog.
- [x] Tests.

### DIFF
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -482,6 +482,13 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         if obj.genre and self.config["keep_existing"]:
             if not self.whitelist or self._is_valid(obj.genre.lower()):
                 return obj.genre, "original fallback"
+            else:
+                # If the original genre doesn't match a whitelisted genre, check
+                # if we can canonicalize it to find a matching, whitelisted genre!
+                if result := _try_resolve_stage(
+                    "original fallback", keep_genres, []
+                ):
+                    return result
 
         # Return fallback string.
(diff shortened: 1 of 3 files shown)


JOJ0 (approved):
Great catch! Thanks! 🤩

PR #6306, opened 2026-01-19
`zero` plugin zeroes `disctotal` if single disc

fredrik:
## Description

When `omit_single_disc` is set, `disctotal` is now also zeroed alongside `disc`. These tags work together ("Disc 2 of 3") so keeping one without the other is inconsistent.

Previously, only the `disc` tag was zeroed. This follows from #6015 which made it into v2.5.1 and added the `omit_single_disc` option.

I've added tests and have used this locally for some time.

## To Do
- [x] Tests
- [x] Documentation
- [x] Changelog

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -41,6 +41,8 @@ Bug fixes
 
 - :doc:`plugins/ftintitle`: Fix handling of multiple featured artists with
   ampersand.
+- :doc:`plugins/zero`: When the ``omit_single_disc`` option is set,
+  ``disctotal`` is zeroed alongside ``disc``.
 
 For plugin developers
 ~~~~~~~~~~~~~~~~~~~~~
@@ -158,8 +160,6 @@ New features
   ``beet import``.
 - :doc:`plugins/random`: Added ``--field`` option to specify which field to use
   for equal-chance sampling (default: ``albumartist``).
-- :doc:`plugins/zero`: When the ``omit_single_disc`` option is set,
-  ``disctotal`` is zeroed alongside ``disc``.
 
 Bug fixes
 ~~~~~~~~~


beetsplug/zero.py
```diff
@@ -124,11 +124,15 @@ def set_fields(self, item, tags):
...
-            if item.disctotal == 1:
+        if self.config["omit_single_disc"].get(bool) and item.disctotal == 1:
+            if "disc" in tags:
                 fields_set = True
                 self._log.debug("disc: {.disc} -> None", item)
                 tags["disc"] = None
+            if "disctotal" in tags:
+                fields_set = True
+                self._log.debug("disctotal: {.disctotal} -> None", item)
+                tags["disctotal"] = None
```
snejus: You can simplify this a little by using `beets.ui.show_model_changes`, see:

```diff
diff --git a/beetsplug/zero.py b/beetsplug/zero.py
index 4623a0c2d..928dda60e 100644
--- a/beetsplug/zero.py
+++ b/beetsplug/zero.py
@@ -21,7 +21,7 @@
 
 from beets.importer import Action
 from beets.plugins import BeetsPlugin
-from beets.ui import Subcommand, input_yn
+from beets.ui import Subcommand, input_yn, show_model_changes
 
 __author__ = "baobab@heresiarch.info"
 
@@ -122,17 +122,9 @@ def set_fields
         Also update the `item` itself if `update_database` is set in the
         config.
         """
-        fields_set = False
-
         if self.config["omit_single_disc"].get(bool) and item.disctotal == 1:
-            if "disc" in tags:
-                fields_set = True
-                self._log.debug("disc: {.disc} -> None", item)
-                tags["disc"] = None
-            if "disctotal" in tags:
-                fields_set = True
-                self._log.debug("disctotal: {.disctotal} -> None", item)
-                tags["disctotal"] = None
+            for tag in {"disc", "disctotal"} & set(tags):
+                tags[tag] = None
 
         if not self.fields_to_progs:
             self._log.warning("no fields list to remove")
@@ -146,13 +138,12 @@ def set_fields
                 match = not progs
 
             if match:
-                fields_set = True
                 self._log.debug("{}: {} -> None", field, value)
                 tags[field] = None
                 if self.config["update_database"]:
                     item[field] = None
 
-        return fields_set
+        return show_model_changes(item)
 
     def process_item(self, item):
         tags = dict(item)
```

This also has the benefit of changes being reported using the same format and user configuration as `beet write` and `beet update`.

fredrik: That looks great, I was not aware! I'll make that change and also update the Changelog.

fredrik:
<s>Obviously, this branch needs to be squashed before merging. ⚠️ </s>

Squashed and ready to go.

beetsplug/zero.py
fredrik: I don't think `show_model_changes` works for me here. The plugin defaults to not updating the database and in that case we're not touching `item`. Please correct me if I'm wrong or if there's another way around this.

I've kept your suggestion for neater iteration and also added changelog and docs entries.

snejus:
Merged in master and moved the changelog note under the **Unreleased** section as we've had a release since the last time this was reviewed.

PR #6313, opened 2026-01-23
Added duduping on mb_trackid to the duplicates plugin

tommyschnabel:
## Description
`plugins/duplicates`: ``dedupe_mb_trackid_on_import`` option added to deduplicate during imports based on already-imported ``mb_trackid``.

## To Do

### DIFF
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -52,6 +52,7 @@ import:
         item: artist title
     duplicate_action: ask
     duplicate_verbose_prompt: no
+    duplicate_track_resolution: no
     bell: no
     set_fields: {}
     ignored_alias_types: []
--- a/beets/importer/session.py
+++ b/beets/importer/session.py
@@ -185,6 +185,13 @@ class ImportSession:
     def resolve_duplicate(self, task: ImportTask, found_duplicates):
         raise NotImplementedError
 
+    def resolve_track_duplicates(self, task: ImportTask, duplicates) -> str:
+        """Decide what to do with album tracks that already exist in the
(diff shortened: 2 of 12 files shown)


snejus:
I think this addresses a real user problem, but it lands in the wrong architectural layer.

The `duplicates` plugin is currently scoped as a command plugin (`beet duplicates`) for inspecting and acting on duplicates already in the library (`docs/plugins/duplicates.rst`, `beetsplug/duplicates.py`). This PR adds import-time policy via `import_task_created` in `beetsplug/duplicates.py`, which changes importer behavior before core duplicate resolution runs.

Duplicate policy already has a clear owner in importer core: `import.duplicate_keys`, `import.duplicate_action`, and `_resolve_duplicates()` (`docs/reference/config.rst`, `beets/ importer/stages.py`). Splitting duplicate policy between importer core and the `duplicates` plugin will make behavior harder to reason about, harder to test, and less predictable for users.

Could we move this into importer core instead?

- Add importer-level support for track-level duplicate checks during album imports (config-driven).
- Reuse the existing duplicate action flow (`skip`/`keep`/`remove`/`merge`/`ask`) so UX and semantics stay consistent.
- Keep `duplicates` focused on post-import library analysis/actions via `beet duplicates`.

tommyschnabel:
@snejus thanks for the feedback! Not super familiar with the entire project structure, so it's not surprising that this changeset doesn't fit in the `duplicates` plugin 😅 

I'll move it to the importer 😄

tommyschnabel:
Closing while I develop the new impl

PR #6317, opened 2026-01-24
Lastgenre: Cleanup existing genres

### LINKED ISSUE #6305: Allow canonicalization of existing genres
The `lastgenre` plugin handles canonicalization when specified.
It does so for new genres from lastfm, as well as existing genres (when enabling merging).

However, there's currently no way of running the canonicalization on existing tags, when `force` is disabled, without `force`.

### Proposed solution

A new config feature flag, e.g. `canonicalize_existing` to run canonicalization on existing genres in `keep_existing`/no-`force` mode.

### Objective

Enable the usecase, where users want to only pull missing genres from lastfm, but still canonicalize/whitelist all of their library.

#### Goals

The ability to canonicalize local, existing tags with the lastgenre canonicalization logic.

#### Anti-goals

Change the current behavior without that flag.

Nukesor:
## Description

Fixes #6305 

Implements a new `cleanup_existing` config flag. The added documentation should explain its behavior. If there're any unclarities, we need to adjust the docs :)

## To Do

- [x] Documentation.
- [x] Changelog.
- [x] Tests.

### DIFF
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -107,6 +107,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
                 "count": 1,
                 "fallback": None,
                 "canonical": False,
+                "cleanup_existing": False,
                 "source": "album",
                 "force": False,
                 "keep_existing": False,
@@ -395,7 +396,19 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         genres = self._get_existing_genres(obj)
 
         if genres and not self.config["force"]:
-            # Without force pre-populated tags are returned as-is.
+            # Without force, but cleanup_existing enabled, we attempt
+            # to canonicalize pre-populated tags before returning them.
+            # If none are found, we use the fallback (if set).
+            if self.config["cleanup_existing"]:
+                keep_genres = [g.lower() for g in genres]
+                if result := _try_resolve_stage("cleanup", keep_genres, []):
+                    return result
+
+                # Return fallback string (None if not set).
+                return self.config["fallback"].get(), "fallback"
+
+            # If cleanup_existing is not set, the pre-populated tags are
+            # returned as-is.
             return genres, "keep any, no-force"
 
         if self.config["force"]:
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -12,6 +12,8 @@ Unreleased
 New features
 ~~~~~~~~~~~~
 
+- :doc:`plugins/lastgenre`: Added ``cleanup_existing`` configuration flag to
+  allow whitelist canonicalization of existing genres.
 - Add native support for multiple genres per album/track. The ``genres`` field
   now stores genres as a list and is written to files as multiple individual
   genre tags (e.g., separate GENRE tags for FLAC/MP3). The
--- a/docs/plugins/lastgenre.rst
+++ b/docs/plugins/lastgenre.rst
@@ -170,6 +170,11 @@ file. The available options are:
 - **canonical**: Use a canonicalization tree. Setting this to ``yes`` will use a
   built-in tree. You can also set it to a path, like the ``whitelist`` config
   value, to use your own tree. Default: ``no`` (disabled).
+- **cleanup_existing**: This option only takes effect with ``force: no``,
+  Setting this to ``yes`` will result in cleanup of existing genres. That
+  includes canonicalization and whitelisting, if enabled. If no matching genre
+  can be determined, the ``fallback`` is used instead. Default: ``no``
+  (disabled).
 - **count**: Number of genres to fetch. Default: 1
 - **fallback**: A string to use as a fallback genre when no genre is found
   ``or`` the original genre is not desired to be kept (``keep_existing: no``).
--- a/test/plugins/test_lastgenre.py
+++ b/test/plugins/test_lastgenre.py
@@ -316,6 +316,30 @@ def config(config):
             },
             (["Jazzin"], "album, any"),
         ),
+        # Canonicalize original genre when force is **off** and
+        # whitelist, canonical and cleanup_existing are on.
+        # "Cosmic Disco" is not in the default whitelist, thus gets resolved "up" in the
+        # tree to "Disco" and "Electronic".
+        (
+            {
+                "force": False,
+                "keep_existing": False,
+                "source": "artist",
+                "whitelist": True,
+                "canonical": True,
+                "cleanup_existing": True,
+                "prefer_specific": False,
+                "count": 10,
+            },
+            ["Cosmic Disco"],
+            {
+                "artist": [],
+            },
+            (
+                ["Disco", "Electronic"],
+                "keep + cleanup, whitelist",
+            ),
+        ),
         # fallback to next stages until found
         (
             {


beetsplug/lastgenre/__init__.py
```diff
@@ -399,6 +400,24 @@ def _try_resolve_stage(
...
+            # Without force and whitelisting + canonicalize_existing, we attempt
+            # to canonicalize pre-populated tags before returning them.
+            # If none are found, we use the fallback.
+            if (
+                self.config["whitelist"]
+                and self.config["canonical"]
+                and self.config["canonicalize_existing"]
+            ):
+                keep_genres = [g.lower() for g in genres]
+                if result := _try_resolve_stage("original", keep_genres, []):
```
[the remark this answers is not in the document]
Nukesor: Nope. `_try_resolve_stage` is a local function.

docs/plugins/lastgenre.rst
```diff
@@ -171,6 +171,10 @@ file. The available options are:
 - **canonical**: Use a canonicalization tree. Setting this to ``yes`` will use a
   built-in tree. You can also set it to a path, like the ``whitelist`` config
   value, to use your own tree. Default: ``no`` (disabled).
+- **canonicalize_existing**: This option only takes effect with ``force: no``,
+  ``canonical: yes`` and ``whitelist: yes``.
+  Setting this to ``yes``, will result in attempted canonicalization of existing
+  genres. If this fails, the ``fallback`` is used. Default: ``no`` (disabled).
```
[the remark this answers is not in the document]
Nukesor: This really doesn't match the style of the remaining documentation + it's overly verbose.
For the user the inner workings aren't really important.

JOJ0:
Hm interesting feature and useful, but I'm not sure if this should maybe be called something containing the word "cleanup". It might even include whitelist filtering without canonicalization too.

I see it as a "cleanup existing only" kind of functionality.

And are you sure this can't be included in the force branch somehow? We are moving away from force:no being what one would expect from it. let's brainstorm some more here please

JOJ0:
@Nukesor I adjusted my wording above slightly. I think it was weird. Maybe still is...

Nukesor:
I didn't read your first response, but the edited doesn't seem weird at all :D

> And are you sure this can't be included in the force branch somehow?

The force-branch already canonicalizes local genres (if `keep_existing` is set), so that case is already covered. Right now, it's just not possible to canonicalize genres in non-force mode.

> I see it as a "cleanup existing only" kind of functionality.

It's more of a "cleanup stuff from lastfm, but please also cleanup local stuff that isn't force-overwritten/merged" :D 

Regarding "cleanup", I would be up for it, although I think that canonicalization better conveys what actually happens. If we were to change the wording, we should do it for the whole plugin though.

> We are moving away from force:no being what one would expect from it.

Neat. How would that look like?
The current flag handling is definitely a bit confusing. A different approach to configuration would probably be easier to understand and use. Although I'm not sure how that would look like yet. Maybe enum-based configuration, where sensible combinations of behavior are given a dedicated name?

JOJ0:
> I didn't read your first response, but the edited doesn't seem weird at all :D
> 
> > And are you sure this can't be included in the force branch somehow?
> 
> The force-branch already canonicalizes local genres (if `keep_existing` is set), so that case is already covered. Right now, it's just not possible to canonicalize genres in non-force mode.


Yes you are right, my bad, I didn't think it through.

Still: We are introducing an option that allows the user / beets to manipulate something even though they said: DO NOT FORCE, that is why I want to be supercareful to make this most obvious and understandable.


> > I see it as a "cleanup existing only" kind of functionality.
> 
> It's more of a "cleanup stuff from lastfm, but please also cleanup local stuff that isn't force-overwritten/merged" :D

You are right

> Regarding "cleanup", I would be up for it, although I think that canonicalization better conveys what actually happens. If we were to change the wording, we should do it for the whole plugin though.
> 
> > We are moving away from force:no being what one would expect from it.
> 
> Neat. How would that look like? The current flag handling is definitely a bit confusing. A different approach to configuration would probably be easier to understand and use. Although I'm not sure how that would look like yet. Maybe enum-based configuration, where sensible combinations of behavior are given a dedicated name?

I agree that it could be easier and we discussed it way back when `-k` was introduced and the initial force behaviour was fixed. An option like `--mode overwrite`, `--mode combine`, and so on was on the table but I was against it because it felt tedious from a cli usability point of view. Using `beet lastgenre -f -k term` or `beet lastgenre -f -K` seemed more practcally usable.

Each way has it's strenghtes though, I agree, but still I don't want to change the usability concept now again. It's established and well documented (I hope!) now.


What I'm trying to say with a cleanup option is something like this:

`beet lastgenre -F --cleanup-existing` -> whitelist filtering and/or canonicalization (and in the future normalizing of spelling / applying of regex aliases) will take place if configured.

A shortform `-c` would be available. 

No better idea for option naming currently, please help me out! :-)

`--help` could state something like this:

```
  -c, --cleanup-existing   even with --no-force provided, existing genres get cleaned up (whitelist, canoncialitation, ...)
```

maybe make this shorter more concise help text...

We should not restrict this no-force-cleanup to canonicalization only.

Nukesor:
Damn. I started writing an answer and then got sidetracked. Sorry for the long wait.

> beet lastgenre -F --cleanup-existing -> whitelist filtering and/or canonicalization (and in the future normalizing of spelling / applying of regex aliases) will take place if configured.

I see. In that context, `cleanup` makes a lot more sense. I like the idea where lastgenre is heading! Rather than a "Pull genres from lastfm", it feels more like a general canonicalization/cleanup/completion tool, which is pretty much what I'm using it for :D.

Tbh. now that I know the planned changes, I'm perfectly fine with `--cleanup-existing`. It'll be a bit confusing in the beginning, but it'll make sense in the long term.

So just to double-check, regarding this PR my todos would be:

- [x] Rename the config flag
- [ ] Anything else :D?

> We should not restrict this no-force-cleanup to canonicalization only.

What other cleanup would be there to perform? The current impl does `_try_resolve_stage`, which performs the other cleanup as well, right?

Nukesor:
btw: The CI is complaining about docstring formatting.
But when I'm running the lint checks and on my machine it looks perfectly fine:

```sh
docstrfmt docs README_kr.rst CONTRIBUTING.rst CODE_OF_CONDUCT.rst README.rst
121 files was checked.
```

I'm using `docstrfmt` v2 though, as that's the only thing running with python 3.14 
See: https://github.com/beetbox/beets/issues/6377

beetsplug/lastgenre/__init__.py
```diff
@@ -399,6 +400,24 @@ def _try_resolve_stage(
...
 
         if genres and not self.config["force"]:
+            # Without force and whitelisting + cleanup_existing, we attempt
+            # to canonicalize pre-populated tags before returning them.
+            # If none are found, we use the fallback.
+            if (
+                self.config["whitelist"]
+                and self.config["canonical"]
+                and self.config["cleanup_existing"]
+            ):
```
JOJ0: please double check but i think we need to change conditions here

also if --no-force (-F) and just whitelist enabled we want to enter try_resolve_stage

canonicalize enabled check and whitelist cleanup happens in there

JOJ0 (changes requested):

docs/plugins/lastgenre.rst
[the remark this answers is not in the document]
JOJ0: I agree with you @Nukesor , keep it simple

snejus:
> I'm using `docstrfmt` v2 though, as that's the only thing running with python 3.14

Beets does not support 3.14 yet - use 3.13 instead and install the dependencies with `poe install` to make sure they are in sync.

beetsplug/lastgenre/__init__.py
```diff
@@ -399,6 +400,24 @@ def _try_resolve_stage(
...
+                if result := _try_resolve_stage("original", keep_genres, []):
+                    return result
+                elif fallback := self.config["fallback"].get():
+                    # Return fallback string.
+                    return fallback, "fallback"
+                else:
+                    # No fallback configured.
+                    return None, "fallback unconfigured"
+
             # Without force pre-populated tags are returned as-is.
```
JOJ0: please adjust this comment: without force and ...

beetsplug/lastgenre/__init__.py
```diff
@@ -399,6 +400,24 @@ def _try_resolve_stage(
...
+            # If none are found, we use the fallback.
+            if (
+                self.config["whitelist"]
+                and self.config["canonical"]
+                and self.config["cleanup_existing"]
+            ):
+                keep_genres = [g.lower() for g in genres]
+                if result := _try_resolve_stage("original", keep_genres, []):
+                    return result
+                elif fallback := self.config["fallback"].get():
```
JOJ0: no elif required we return anyway -> if

beetsplug/lastgenre/__init__.py
```diff
@@ -399,6 +400,24 @@ def _try_resolve_stage(
...
+                    # Return fallback string.
+                    return fallback, "fallback"
+                else:
```
JOJ0: no else required we return anyway -> remove one indent

beetsplug/lastgenre/__init__.py
snejus: Nice work on this! One thought on the condition here - if the user has set `cleanup_existing`, they've already opted in, so silently doing nothing when `whitelist` or `canonical` aren't set seems like a footgun.

`_try_resolve_stage` already handles whether whitelist filtering or canonicalization applies based on its own config, so wouldn't the following be sufficient?

```suggestion
            if self.config["cleanup_existing"]:
```

JOJ0: ah thanks @snejus seems like our reviews are overlapping here. yeah i agree. whitelist checking should also be included when `cleanup_existing` is on. 

also note @Nukesor I changed the name of this PR to now clearly reflect that we slightly changed/extended what the original submission was about.

JOJ0:
> > We should not restrict this no-force-cleanup to canonicalization only.
> 
> What other cleanup would be there to perform? The current impl does `_try_resolve_stage`, which performs the other cleanup as well, right?

whitelist checking as mentioned in the inline comments. and in the future maybe normalizing genre names. eg. DnB -> Drum And Bass, and removing blacklisted genres.

beetsplug/lastgenre/__init__.py
JOJ0: @snejus suggested condition fix is correct here. yes!

beetsplug/lastgenre/__init__.py
JOJ0: or maybe remove it here and extend the larger comment above, right after the condition to include a concise description right at the entry point of the `if genres and force not true`) branch

Nukesor: I kept the split, but adjusted the wording to match the code flow

beetsplug/lastgenre/__init__.py
Nukesor: Fair point. I wanted to be extra safe to not cause any breaking changes, but I went through the _try_resolve_stage path and you're right. That already handles all of the whitelist/canonicalize conditions :smile:

Nukesor:
@JOJ0 @snejus I adjusted according your review :)

The tests fail now however. From what I can see, this is an issue with the way configuration is handled in beets. Test state (specifically the config values) persist once set.

It looks like each plugin has a view to the global persisted `beets.config` field? As a result, variables that aren't explicitly overwritten are persisted in that global config view. 

Should this be fixed in a specific way or should I just work around this? The tests can be "fixed" by adding a `cleanup_existing: False` to the test fixture right after the one that sets `cleanup_existing: True`.

Nukesor:
I went ahead and added a commit on top that exposes the default config as a static method and uses that default config to reset the state in between fixture calls.

There were 5 tests that depended on `count: 10` being set on previous test fixtures, which I adjusted accordingly.

Let me know what you think of that change :)

Bergbok:
Hey, thanks for your work on this @Nukesor !

One thing I'd like to see changed is it being untied from `force: no`, I feel like it's a bit confusing.  
It's changing force's behavior unnecessarily.

I'd much prefer another solution. What if running `beet lastgenre --cleanup-existing` ran through your library's genres without interacting with Last.fm and updated them based on your whitelist/canonicalization tree? This would be very useful for when you modify your whitelist or tree file and want to quickly apply those changes to your library without re-fetching from Last.fm.

You added a section in the docs for the new option but I think you should probably also update the [Handling pre-populated tags section](https://beets.readthedocs.io/en/stable/plugins/lastgenre.html#handling-pre-populated-tags) and force's description and if we're going this route.

Nukesor:
Hey @Bergbok 

If I understood you correctly, `--cleanup-existing` would then imply a `--no-force` and behave somewhat like a subcommand. 

IMO, that behavior would makes sense, if we were to have multiple subcommands that represent different "stages".

- `lastgenre fetch` -> Pull metadata from lastfm, potentially overwriting or merging existing genres, depending on `--keep-existing` and `--force`
- `lastgenre cleanup` -> Go through all genres and normalize them (capitalization, aliasing, whitelisting, canonicalization)

Right now, existing tags are also "cleaned up" in `force: yes` + `keep-existing: yes` mode, which is why I decided to go for another flag. Having `lastgenre --keep-existing` behave like a subcommand that cleans up all existing genres, but then also having local genres cleaned up in another mode is a bit confusing.

Is there a way to easily introduce subcommands to plugins?
I guess a split into such subcommands would also require a quite large refactoring, as the fetch+merge and normalization logic should then probably be separated and those two processes are currently very closely tied together.

JOJ0:
> I went ahead and added a commit on top that exposes the default config as a static method and uses that default config to reset the state in between fixture calls.
> 
> 
> 
> There were 5 tests that depended on `count: 10` being set on previous test fixtures, which I adjusted accordingly.
> 
> 
> 
> Let me know what you think of that change :)

nice catch!!! would it be too much to ask if you could split out this test refactor/fix into a separate PR. We have so much lastgenre stuff in flight currently and it would be easier for us that way! thank you very much!! 🙏

Nukesor:
@JOJ0 created https://github.com/beetbox/beets/pull/6386 :)
I left this MR on top of the other one so the tests pass :)  

(Did a bit of a oopsie on the rebase+cherrypick, but the history should be clean now again)

Nukesor:
@JOJ0 I rebased onto `master` and resolved any conflicts with the recent lastgenre merges :)
Anything else to do in here (or the other MR) or is this waiting for #6368 to land (just asking whether I should keep checking/rebasing or wait until you ping me :D )?

JOJ0:
> @JOJ0 I rebased onto `master` and resolved any conflicts with the recent lastgenre merges :) Anything else to do in here (or the other MR) or is this waiting for #6368 to land (just asking whether I should keep checking/rebasing or wait until you ping me :D )?

Thanks. Nothing to do here. LGTM but I would like to wait for the test thing yes. Best if I ping you once that is done so you don't have to keep doing boring rebase work :-) Thanks!!!

docs/changelog.rst
```diff
@@ -119,6 +119,8 @@ been dropped.
 New features
 ~~~~~~~~~~~~
 
+- :doc:`plugins/lastgenre`: Added ``cleanup_existing`` configuration flag to
```
snejus: Move the note under the **Unreleased** section at the top of the file (we've had a release since).

Nukesor: Ah right. Done and rebased :)

Thanks @snejus

snejus (approved):

snejus:
@JOJ0 back to you!

JOJ0:
```
lastgenre: Alves - Episode 05 - 01 - Dark Era Of Amnesia (Resurrection) (2016)  [] $genre
lastgenre: raw last.fm tags: []
lastgenre: existing genres taken into account: ['drum and bass']
lastgenre: Resolved (keep + original, whitelist): ['Drum And Bass']
```

"keep + original, whitelist" is the log label we get now, right?

when running without --force we usually get "keep any, no-force"

hmm,. let's review what the purpose of this logging label was as its "kind of" documented in the docstring of _get_genre


```
        A `(genres, label)` pair is returned, where `label` is a string used for
        logging. For example, "keep + artist, whitelist" indicates that existing
        genres were combined with new last.fm genres and whitelist filtering was
        applied, while "artist, any" means only new last.fm genres are included
        and the whitelist feature was disabled.
```

In the fallback stage we have a pretty similar logic and these are the possible outcomes:

original fallback
keep + original fallback, whitelist
keep + original fallback, any


This all is very confusing to me already. Haha. Hmm I'm just thinking out lout here trying to justify that the label we get now "keep + original, whitelist" is good enough. Or would it make more sense if we pass `original cleanup` to `_try_resolve_stage`, that way we would get something like this:


```
keep + original cleanup, whitelist
```

which is also kind of weird because it says: we keep original genres and combine them with original cleaned up genres.

Maybe the `keep +` logic in try_resolve_genre is just to dumb to clearly help creating a log label  that makes sense  for this case.

Help me think here @Nukesor and sorry for another "brainstorming" delay! We are almost done here! Thanks for your patience!


One thing I do like very much as this feature turned out: In the beginning I thought we should have this option as `--cleanup-existing` too, to be able to use it on a per-run basis. Now I think it is good we make it a "once and for all decision", so it doesn't further complicate the already rather complex UI option combination possibilities! And I suppose deciding this once will suffice for most users.

Nukesor:
Haha, yep. I was pretty confused when picking the label as well :sweat_smile:. It's a bit obscure.

> which is also kind of weird because it says: we keep original genres and combine them with original cleaned up genres.

Maybe a `keep + cleanup, whitelist`? IMO, That conveys the meaning pretty well. `keep` (keep the originals) `+ cleanup` (but clean them up) `, whitlist` (and apply whitelist logic)
I changed it to `"cleanup"` for now :)

> One thing I do like very much as this feature turned out: In the beginning I thought we should have this option as --cleanup-existing too, to be able to use it on a per-run basis. Now I think it is good we make it a "once and for all decision", so it doesn't further complicate the already rather complex UI option combination possibilities! And I suppose deciding this once will suffice for most users.

Yep, I agree :D

JOJ0 (approved):

Issue #6312, opened 2026-01-22
Error: MusicBrainz not reachable in get release by ID with query '295033b5-d225-4abb-8237-29811a0806f0'


catap:
### Problem

When run beet import of few albums without verbose mode it often complains as:
```
Error: MusicBrainz not reachable in get release by ID with query '295033b5-d225-4abb-8237-29811a0806f0'
```

Sometimes I can decide about one album or a few, like here I've pressed 'S' but it fails:
```
Finding tags for album "Johnny Nash - All The Best".
  Candidates:
  1. (49.1%) Johnny Nash - The Best of Johnny Nash
             ≠ tracks, album, missing tracks, ...
             MusicBrainz, Digital Media, 2009, DZ, None, None, None
  2. (47.7%) Johnny Nash - The Best of Johnny Nash
             ≠ tracks, album, missing tracks, ...
             MusicBrainz, CD, 1991, NL, Columbia, 468859 2, None
  3. (28.2%) Tina Turner - All the Best - The Hits
             ≠ tracks, artist, missing tracks, ...
             MusicBrainz, CD, 2005, US, Capitol Records, CDP 0946 3 40186 2 3, None
  4. (25.9%) Bonnie Tyler - All the Best
             ≠ missing tracks, tracks, artist, ...
             MusicBrainz, 3xCD, 1996, XE, Ariola Express, 74321 41116 2, None
➜ # selection (default 1), Skip, Use as-is, as Tracks, Group albums,
Enter search, enter Id, aBort, eDit, edit Candidates, Print tracks,
Open files with Picard? S
Error: MusicBrainz not reachable in get release by ID with query '295033b5-d225-4abb-8237-29811a0806f0'
```

If I run it in verbose mode, chances to get issue smaller, but it still may occure. Without any stacktrace.

Looks like that we hit some rate limit on MusicBrainz side.

### Setup

* OS: OpenBSD -current
* Python version: 3.13.11
* beets version: 2.5.1
* Turning off plugins made problem go away (yes/no): without MusicBrainz problem go away, indeed.

My configuration (output of `beet config`) is:

```yaml
directory: /mnt/diskstation/music
asciify_paths: yes

import:
    move: yes
    auto: no
fetchart:
    sources: filesystem coverart itunes amazon albumart lastfm
    lastfm_key: REDACTED
    auto: yes
    minwidth: 0
    maxwidth: 0
    quality: 0
    max_filesize: 0
    enforce_ratio: no
    cautious: no
[...]

PR #6322, opened 2026-01-25
Retries with 1, 2, 4, 8, 16, 32s backoff

catap:
## Description

At least it allows me to more or less use MusicBrainz, and makes issues from #6312 almost not noticible

## To Do


test/plugins/utils/test_request_handler.py
```diff
@@ -14,15 +14,26 @@
...
+                    headers=last_response.headers,
+                )
+            return last_response
+
+        def responses():
+            yield NewConnectionError(None, "Connection failed")
+            yield URLError("bad")
+            while True:
+                yield make_response()
+
```
snejus: I had to stare at this for a couple of minutes until I managed to work out how does this work 😅 I think this needs to be simplified...

However, there is a more significant issue - test durations:

```fish
$ poe test --durations=10 test/plugins/utils/test_request_handler.py
...
=============================================================================== slowest 10 durations ===============================================================================
31.04s call     test/plugins/utils/test_request_handler.py::TestRequestHandlerRetry::test_retry_exhaustion[conn_error]
31.03s call     test/plugins/utils/test_request_handler.py::TestRequestHandlerRetry::test_retry_exhaustion[server_error]
1.01s call     test/plugins/utils/test_request_handler.py::TestRequestHandlerRetry::test_retry_on_connection_error[success]
0.13s setup    test/plugins/utils/test_request_handler.py::TestRequestHandlerRetry::test_retry_on_connection_error[success]
```

Tests that take 30s are now viable here, unfortunately. I think we want to re-think how do we test this configuration.

catap: Yes, it is a price tag. Frankly, I not sure that we should test it. And yes, figure out how to test it was... quite a challenge!

But I have no idea how to test timeout based things without time penalty.

catap:
@snejus what do you think about this tests?

catap:
@snejus as requested

PR #6324, opened 2026-01-27
Add remove_missing option to convert plugin

javierbg:
## Description

This PR adds a new configuration/option to the convert plugin.

The "remove_missing" option indicates that files in the convert destination directory not present in the library should be removed (asking for confirmation).

## To Do

- [x] Documentation
- [x] Changelog
- [x] Tests

Beware that this is my first PR here, so I'm still unaware of common practices. I tried to follow the same principles that I saw in the `convert.py` file, but I'm sure a lot of them will be outdated, feel free to suggest any changes.

Also, I'm still not 100% sold on the config/option name. Let me know if you have a suggestion.

### DIFF
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -796,11 +796,11 @@ class ConvertPlugin(BeetsPlugin):
         convert = [self.convert_item(keep_new) for _ in range(self.threads)]
         pipeline.Pipeline([iter(items), convert]).run_parallel()
 
-    def remove_non_item_files(self, items: list[Item], yes: bool):
+    def remove_non_item_files(self, items: list[Item], yes: bool) -> None:
         """
         Remove all files in the destination directory ``dest`` that do not
         correspond to an item to be converted.
-        
+
         If ``pretend`` is set, keep files and report what would be removed.
         If ``yes`` is set, do not ask for confirmation.
         """
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -271,6 +271,10 @@ class ConvertPlugin(BeetsPlugin):
         return not self.hardlink and self.config["link"].get(bool)
 
     @cached_property
+    def remove_missing(self) -> bool:
+        return self.config["remove_missing"].get(bool)
+
+    @cached_property
     def command(self) -> FormatCommand:
         """Return the command template and the extension from the config."""
         fmt = ALIASES.get(self.fmt, self.fmt)
@@ -686,8 +690,10 @@ class ConvertPlugin(BeetsPlugin):
             for album in albums:
                 self.copy_album_art(album)
 
-        if remove_missing:
-            self.remove_non_item_files(items, dest, fmt, pretend, opts.yes)
+        if self.remove_missing:
+            self.remove_non_item_files(
+                items, self.dest, self.fmt, pretend, opts.yes
+            )
 
         # If the user supplied a playlist name, create a playlist for files
         # copied to the destination.
@@ -798,7 +804,7 @@ class ConvertPlugin(BeetsPlugin):
         not actually remove any files and only print a list of files to be
         deleted. If ``yes=True`` it will not ask for confirmation.
         """
-        _, ext = get_format(fmt)
+        _, ext = self.command
         item_destinations = {
             replace_ext(item.destination(basedir=dest), ext) for item in items
         }
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -108,6 +108,7 @@ class ConvertPlugin(BeetsPlugin):
                 "force": False,
                 "refresh": False,
                 "keep_new": False,
+                "remove_missing": False,
             }
         )
         self.early_import_stages = [self.auto_convert, self.auto_convert_keep]
@@ -199,6 +200,13 @@ class ConvertPlugin(BeetsPlugin):
                 " different computer, or opened from an external drive,"
                 " relative paths pointing to media files will be used."
             ),
+        )
+        cmd.parser.add_option(
+            "-r",
+            "--remove-missing",
+            action="store_true",
+            help="""remove all files in the destination directory that are not
+                    present in the library.""",
         )
         cmd.parser.add_option(
             "-F",
@@ -675,6 +683,9 @@ class ConvertPlugin(BeetsPlugin):
         if opts.album and self.config["copy_album_art"]:
             for album in albums:
                 self.copy_album_art(album)
+        
+        if remove_missing:
+            self.remove_non_item_files(items, dest, fmt, pretend, opts.yes)
 
         # If the user supplied a playlist name, create a playlist for files
         # copied to the destination.
@@ -777,3 +788,35 @@ class ConvertPlugin(BeetsPlugin):
         """
         convert = [self.convert_item(keep_new) for _ in range(self.threads)]
         pipeline.Pipeline([iter(items), convert]).run_parallel()
+
+    def remove_non_item_files(self, items, dest, fmt, pretend, yes):
+        """
+        Remove all files in the destination directory ``dest`` that do not
+        correspond to an item to be converted. If ``pretend=True`` it will
+        not actually remove any files and only print a list of files to be
+        deleted. If ``yes=True`` it will not ask for confirmation.
+        """
+        _, ext = get_format(fmt)
+        item_destinations = {
+            replace_ext(item.destination(basedir=dest), ext) for item in items
+        }
+        files_to_remove = list()
+
+        for dirpath, _, filenames in os.walk(dest):
+            for filename in filenames:
+                filepath = util.bytestring_path(os.path.join(dirpath, filename))
+                if filepath not in item_destinations:
+                    files_to_remove.append(filepath)
+
+        if not files_to_remove:
+            ui.print_("No files to be removed")
+            return
+
+        ui.print_("Files to be deleted in the destination folder:")
+        for f in files_to_remove:
+            ui.print_(util.displayable_path(f))
+
+        if (not pretend) and (yes or ui.input_yn("Delete files? (Y/n)")):
+            for file in files_to_remove:
+                util.remove(file)
+                self._log.info(f'Removed file "{util.displayable_path(file)}"')
--- a/beetsplug/convert.py
+++ b/beetsplug/convert.py
@@ -692,9 +692,7 @@ class ConvertPlugin(BeetsPlugin):
                 self.copy_album_art(album)
 
         if self.remove_missing:
-            self.remove_non_item_files(
-                items, self.dest, self.fmt, pretend, opts.yes
-            )
+            self.remove_non_item_files(items, opts.yes)
 
         # If the user supplied a playlist name, create a playlist for files
         # copied to the destination.
@@ -798,7 +796,7 @@ class ConvertPlugin(BeetsPlugin):
         convert = [self.convert_item(keep_new) for _ in range(self.threads)]
         pipeline.Pipeline([iter(items), convert]).run_parallel()
 
-    def remove_non_item_files(self, items, dest, fmt, pretend, yes):
+    def remove_non_item_files(self, items: list[Item], yes: bool):
         """
         Remove all files in the destination directory ``dest`` that do not
         correspond to an item to be converted. If ``pretend=True`` it will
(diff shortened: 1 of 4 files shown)


javierbg:
I see some failing checks on Windows, but it appears to be failing when invoking ffmpeg. I don't know if this is to be expected or not.

javierbg:
Ok, that seemed to do the trick. I made the mistake of copying the setup from another test case that must have been trying to test something about the path encodings.

javierbg:
The same Windows error happens again. Can anyone help figure it out? I really don't understand why this would fail on Windows, when the "run_convert" call works in all other tests.

beetsplug/convert.py
```diff
@@ -242,6 +243,13 @@ def commands(self):
                               drive, relative paths pointing to media files
                               will be used.""",
         )
+        cmd.parser.add_option(
+            "-r",
+            "--remove-missing",
+            action="store_true",
+            help="""remove all files in the destination directory that are not
+                    present in the library.""",
```
snejus: You want to define this without multi lines, see #6706 for context.

beetsplug/convert.py
```diff
@@ -760,7 +773,14 @@ def _get_opts_and_config(self, opts):
         else:
             hardlink = self.config["hardlink"].get(bool)
             link = self.config["link"].get(bool)
+
+        if opts.remove_missing is not None:
+            remove_missing = opts.remove_missing
+        else:
+            remove_missing = self.config["remove_missing"].get(bool)
```
snejus: This is not needed. The configuration is updated with command-line options here:
```py
    def convert_func(
        self, lib: Library, opts: optparse.Values, args: list[str]
    ) -> None:
        self.config.set(vars(opts))
```

```suggestion
```

beetsplug/convert.py
```diff
@@ -804,3 +825,35 @@ def _parallel_convert(
         ]
         pipe = util.pipeline.Pipeline([iter(items), convert])
         pipe.run_parallel()
+
+    def remove_non_item_files(self, items, dest, fmt, pretend, yes):
```
snejus: This module is fully typed - please add types to the function as well.

snejus: Also note that this file has seen significant updates since your PR submission. Once you rebase, you will find that we these function calls with very long argument lists are gone. You are expected to define a `@cached_property` with the configuration variable, and use it directly from whichever function you need it. 

See https://github.com/beetbox/beets/commit/34772d2684b9c85822193e0cea7153a2e724368e for more context

snejus (changes requested):
Hey, apologies for a very late review - I think this work now mostly need rebasing because this plugin has seen some structural updates since.

javierbg:
Hi @snejus, thanks for the review.

I rebased to the current master branch and updated the changes with your suggestions (the new configuration/parameter usage is so much nicer).

I confirmed that everything is still working and updated the tests accordingly as well.

Also, it looks like the Windows test errors are gone now, so that is sorted out.

beetsplug/convert.py
```diff
@@ -777,3 +795,36 @@ def _parallel_convert(self, items: list[Item], keep_new: bool):
         """
         convert = [self.convert_item(keep_new) for _ in range(self.threads)]
         pipeline.Pipeline([iter(items), convert]).run_parallel()
+
+    def remove_non_item_files(self, items: list[Item], yes: bool):
+        """
+        Remove all files in the destination directory ``dest`` that do not
+        correspond to an item to be converted. If ``pretend=True`` it will
+        not actually remove any files and only print a list of files to be
+        deleted. If ``yes=True`` it will not ask for confirmation.
```
snejus:
```suggestion
        correspond to an item to be converted.
        
        If ``pretend`` is set, keep files and report what would be removed.
        If ``yes`` is set, do not ask for confirmation.
```

beetsplug/convert.py
```diff
@@ -777,3 +795,36 @@ def _parallel_convert(self, items: list[Item], keep_new: bool):
         """
         convert = [self.convert_item(keep_new) for _ in range(self.threads)]
         pipeline.Pipeline([iter(items), convert]).run_parallel()
+
+    def remove_non_item_files(self, items: list[Item], yes: bool):
```
snejus: Add the return type please!

beetsplug/convert.py
```diff
@@ -777,3 +795,36 @@ def _parallel_convert(self, items: list[Item], keep_new: bool):
...
+            for filename in filenames:
+                filepath = util.bytestring_path(os.path.join(dirpath, filename))
+                if filepath not in item_destinations:
+                    files_to_remove.append(filepath)
+
+        if not files_to_remove:
+            ui.print_("No files to be removed")
+            return
+
+        ui.print_("Files to be deleted in the destination folder:")
```
snejus, line 824: What about

```suggestion
        self._log.info("Files to be deleted in the destination folder:")
```

docs/plugins/convert.rst
```diff
@@ -170,6 +173,12 @@ The available options are:
 - **refresh**: Refresh the converted files if needed by re-converting modified
   original files. This configuration is overridden by the ``-r`` (``--refresh``)
   command line option. Default: ``false``.
+- **remove_missing**: Whether or not to remove files in the destination folder
+  that are no longer present in the library. This means that if you removed an
+  item from the database that was previously converted, it will be removed in
+  the next run of the ``convert`` command (it will ask for confirmation unless
+  the ``--yes`` option is enabled). This is useful if you want to keep a
+  converted version of your library synced. Default: ``false``.
```
snejus:
```suggestion
- **remove_missing**: Whether to remove files in the destination folder that
  are no longer present in the library. This means that if you remove an item
  from the database that was previously converted, it will be removed on the
  next run of the ``convert`` command (you will be asked for confirmation
  unless the ``--yes`` option is enabled). This is useful if you want to keep a
  converted version of your library in sync. Default: ``false``.
```

docs/changelog.rst
```diff
@@ -24,6 +24,9 @@ New features
 - :doc:`plugins/tidal`: Add cover art support. Album metadata now includes
   ``cover_art_url`` from Tidal's ``coverArt`` relationship, which the
   :doc:`plugins/fetchart` plugin can retrieve.
+- :doc:`plugins/convert`: Added the new ``remove_missing`` configuration and
+  corresponding ``--remove-missing`` option to enable removing files in the
+  destination directory that were removed from the library.
```
```suggestion
- :doc:`plugins/convert`: Added a new ``remove_missing`` configuration option
  and corresponding ``-r``/``--remove-missing`` flags to remove files from the
  destination directory when they have been removed from the library.
```

test/plugins/test_convert.py
```diff
@@ -327,3 +327,78 @@ def test_no_convert_skip(self, config, config_value, should_skip):
...
+        self.run_convert("--yes")
+
+        assert not file_to_remove.exists()
+
+    def test_convert_dont_remove_present(self):
+        """Test that already converted files are not removed when the
+        remove_missing option is enabled"""
+        # This file mocks an already existing converted file that should not be
+        # removed or modified.
+        file_not_to_remove = self.create_dummy_file("artist/album/01 title.mp3")
```
snejus: It's not clear where these values come from - use `self.item` to grab the relevant data that you need. Also, best to use `Path(item.artist) / item.album ...` to make sure this works on Windows.

test/plugins/test_convert.py
```diff
@@ -327,3 +327,78 @@ def test_no_convert_skip(self, config, config_value, should_skip):
...
+
+        self.convert_dest = self.temp_dir_path / "convert_dest"
+        self.convert_dest.mkdir(parents=True)
+
+        self.config["convert"] = {
+            "dest": str(self.convert_dest),
+            "format": "mp3",
+        }
+
+    def create_dummy_file(self, path):
```
snejus: This helper needs types

test/plugins/test_convert.py
```diff
@@ -327,3 +327,78 @@ def test_no_convert_skip(self, config, config_value, should_skip):
...
+        "Creates a dummy file in the conversion directory"
+        p = self.convert_dest / path
+        p.parent.mkdir(parents=True, exist_ok=True)
+        with p.open("w") as f:
+            f.write("test")
+        return p
+
+    def test_convert_not_removemissing(self):
```
snejus:
```suggestion
    def test_default_convert(self):
```

test/plugins/test_convert.py
```diff
@@ -327,3 +327,78 @@ def test_no_convert_skip(self, config, config_value, should_skip):
...
+        remove_missing option is enabled"""
+        # This file mocks an already existing converted file that should not be
+        # removed or modified.
+        file_not_to_remove = self.create_dummy_file("artist/album/01 title.mp3")
+        original_mtime = os.path.getmtime(file_not_to_remove)
+
+        self.run_convert("--yes", "--remove-missing")
+
+        assert file_not_to_remove.exists()
+        assert os.path.getmtime(file_not_to_remove) == original_mtime
```
snejus: I think you can reduce each of these tests into a single `@pytest.mark.parametrize`:

1. It seems like you can use the following params: `plugin_config, cli_opts, should_be_removed`
1. Pull `config` fixture into the test function
1. Assert both `file_not_to_remove` and `file_to_remove` expected outcomes.

snejus (changes requested):

beetsplug/convert.py
javierbg: I'm not sure about this: confirmation is asked afterwards about whether or not to delete the files, so it is part of the user interaction. Using the logger will be dependent on the logging configuration, isn't that so?

In any case, I'm not familiar with the project conventions in this regard, so if you are sure we can go ahead with this. Should I change the rest of `ui.print_` instances in this method?

test/plugins/test_convert.py
javierbg: Rewrote it using `syspath` and `self.item.path`. Is this okay?

test/plugins/test_convert.py
javierbg: Fixed

test/plugins/test_convert.py
javierbg: Thanks for the suggestion. I have refactored it into a single test now.

I didn't feel the need to add both a `file_not_to_remove` and `file_to_remove` separate expected outcomes, as the item file definitionally never has to be removed. Instead I have added a `files_to_mark_for_removal` parameter and a `expect_removal` parameter.

Let me know what you think.

test/plugins/test_convert.py
javierbg: Kept this change in the refactor rewrite

snejus: Use `item.filepath` instead since we are trying to migrate to `pathlib`.

test/plugins/test_convert.py
```diff
@@ -321,3 +321,71 @@ def test_no_convert_skip(self, config, config_value, should_skip):
         item = Item(format="ogg", bitrate=256)
         config["convert"]["no_convert"] = config_value
         assert convert.ConvertPlugin().in_no_convert(item) == should_skip
+
+
+class TestConvertRemoveMissing(ConvertPluginHelper, ConvertCommand):
+    "Tests the effect of the `remove_missing option`"
```
snejus, line 327:
```suggestion
    """Tests the effect of the `remove_missing option`"""
```

test/plugins/test_convert.py
```diff
@@ -321,3 +321,71 @@ def test_no_convert_skip(self, config, config_value, should_skip):
...
+        return p
+
+    @pytest.mark.parametrize(
+        "plugin_config,cli_options,files_to_mark_for_removal,expect_removal",
+        [
+            ({}, [], ["to_remove.mp3"], False),
+            ({}, ["--remove-missing", "--pretend"], ["to_remove.mp3"], False),
+            ({}, ["--remove-missing"], ["to_remove.mp3"], True),
+            ({"remove_missing": True}, [], ["to_remove.mp3"], True),
+        ],
```
snejus, line 358: `files_to_mark_for_removal` does not vary across params - remove it and assert unconditionally.

test/plugins/test_convert.py
```diff
@@ -321,3 +321,71 @@ def test_no_convert_skip(self, config, config_value, should_skip):
...
+        # removed or modified.
+        file_not_to_remove = self.create_dummy_file(
+            util.syspath(self.item.path)
+        )
+        original_mtime = os.path.getmtime(file_not_to_remove)
+
+        # Create files to be marked for removal
+        paths_to_mark_for_removal = [
+            self.create_dummy_file(f) for f in files_to_mark_for_removal
+        ]
```
snejus, line 377:
```suggestion
        path_to_mark_for_removal = self.create_dummy_file("to_remove.mp3")
```

snejus (changes requested):

test/plugins/test_convert.py
javierbg: Removed

javierbg:
Again, sorry for the late response. All issues should be addressed now.

The Windows tests are failing again due to ffmpeg errors, am I missing something? They seem to come and go.

javierbg:
I have been able to run the tests on a Windows machine and I cannot reproduce the failing tests. My best guess is that ffmpeg is not being correctly installed in the Windows test environment.

PR #6325, opened 2026-01-29
Fix to properly redact Acoustid API key when dumping configuration file

### LINKED ISSUE #6276: Redact `acoustid.apikey` for `chroma` plugin
### Problem

As @arsaboo noted under https://github.com/beetbox/beets/discussions/6269#discussioncomment-15429229, the following configuration (for `chroma`) should be redacted:

```yaml
acoustid:
  apikey: ...
```

westclox-star:
## Description

Fixes #6276 

Fix to properly redact Acoustid API key when dumping configuration file

## To Do

- [ ] Documentation. 
- [ ] Changelog. 
- [X ] Tests. 

As simple as it gets.  This is my first open source PR btw which is cool (though I wish it were something more interesting

### DIFF
--- a/beetsplug/chroma.py
+++ b/beetsplug/chroma.py
@@ -169,7 +169,6 @@ def _all_releases(items):
     for item in items:
         if item.path not in _matches:
             continue
-        
         _, release_ids = _matches[item.path]
         for release_id in release_ids:
             relcounts[release_id] += 1


snejus:
Wait... Seems like this configuration had already been redacted correctly... (Your change removes the redaction, actually - it should indeed use the global (imported) config).

Seems like #6276 was a false alarm from me (I must have tested it in one of my outdated local branches, probably...). Apologies!

westclox-star:
Np, I learned a lot anyways. I'll have to look into the referenced vs the global because it seemed to fix the issue for me.  I must have a miss understanding.

PR #6327, opened 2026-02-01
Clarify TODOs in is_hidden docstring

kalel-commits:
This PR clarifies existing TODOs by documenting known limitations in the
is_hidden() docstring. No behavior is changed.

### DIFF
--- a/beets/util/hidden.py
+++ b/beets/util/hidden.py
@@ -25,6 +25,9 @@ from pathlib import Path
 def is_hidden(path: bytes | Path) -> bool:
     """
     Determine whether the given path is treated as a 'hidden file' by the OS.
+
+    Note: A platform-specific check is performed on each invocation. Bytes
+    inputs are supported for backward compatibility.
     """
 
     if isinstance(path, bytes):


beets/util/hidden.py
```diff
@@ -25,6 +25,9 @@
 def is_hidden(path: bytes | Path) -> bool:
     """
     Determine whether the given path is treated as a 'hidden file' by the OS.
+
+    Note: A platform-specific check is performed on each invocation. Bytes
+    inputs are supported for backward compatibility.
```
JOJ0, line 30: I might be slow here but what did you intend to clarify? Lines 25-27 tell exactly the same already. Maybe I'm missing something, please elaborate your idea here.

kalel-commits: Thanks for the look happy to elaborate the intention here was not to provide any new information, rather to declare the meaning of TODOs that already exists in the docstring. 

currently the docstring is an explanation of what the function does, but the TODOs are clues to a couple details that might not be obvious behind this operation:
that a platform-dependent check is performed on each call, and since bytes inputs still have support largely for backwards compatibility.

This note seemed like an attempt to bring that detail out for readers without having them know to go find or understand the TODO comments.

If this does feel a bit redundant or not useful I will drop the changes.

snejus: > that a platform-dependent check is performed on each call

This is covered by the ending of the docstring: `... by the OS`.

> since bytes inputs still have support largely for backwards compatibility.

This is clear from the input type:

```py
def is_hidden(path: bytes | Path) -> bool:
```

I agree with @JOJ0 - this addition is redundant.

snejus:
Thanks for your contribution @kalel-commits 

Thanks for the PR and for looking into these TODOs! I am closing this one as this change is likely redundant. 

On the other hand, we definitely appreciate the drive to improve our documentation! The codebase has several areas where documentation is missing entirely or needs significant updates. If you're interested, we would love your help on some open documentation issues, please see some of them here: https://github.com/beetbox/beets/issues?q=is%3Aissue%20state%3Aopen%20label%3Adocs

PR #6329, opened 2026-02-01
Type musicbrainz

snejus:
## Add full type coverage to the MusicBrainz plugin

This PR introduces complete static typing for `beetsplug/musicbrainz.py`, `beetsplug/mbpseudo.py`, and `beetsplug/_utils/musicbrainz.py`, and enforces it via `mypy` strict mode.

### Key changes

**New `TypedDict` schema in `beetsplug/_utils/musicbrainz.py`**

A comprehensive set of `TypedDict` classes now models the full MusicBrainz API response surface — `Release`, `Recording`, `Track`, `ReleaseGroup`, `Artist`, `ArtistCredit`, `Work`, and all relation types. Public API methods (`get_release`, `get_recording`, etc.) now return these typed shapes instead of the opaque `JSONDict`.

**Key normalization: dash-to-underscore**

The internal `_normalize_data` method (previously `_group_relations`) now also converts all hyphenated keys (e.g. `artist-credit`, `release-group`, `sort-name`) to underscored equivalents (e.g. `artist_credit`, `release_group`, `sort_name`) at parse time. This makes the data structure Python-idiomatic and is what allows the `TypedDict` definitions to use clean attribute names. All downstream field accesses, test fixtures, and JSON test resources are updated accordingly.

**`mypy` strict mode enabled** for the three affected modules via `setup.cfg`.

### Impact

- No behaviour change for end users — this is purely an internal refactor.
- Code reading the MusicBrainz response now has IDE completion and type-checking support.
- The normalization boundary is clearly established at `_normalize_data`, so callers never see raw hyphenated keys.
- Test fixtures and resource JSON files are updated to match the new normalized shape.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -32,9 +32,16 @@ Bug fixes
 - :doc:`plugins/listenbrainz`: Retry listenbrainz requests for temporary
   failures.
 
-..
-    For plugin developers
-    ~~~~~~~~~~~~~~~~~~~~~
+For plugin developers
+~~~~~~~~~~~~~~~~~~~~~
+
+- Consumers of :py:class:`beetsplug._utils.musicbrainz.MusicBrainzAPI` now
+  receive normalized MusicBrainz payloads with underscore-separated field names
+  (for example ``artist_credit`` and ``release_group``) and grouped relation
+  lists such as ``work_relations``, ``release_relations``, and
+  ``url_relations``. The API responses are also now fully typed with concrete
+  ``TypedDict`` models for releases, recordings, works, and relations. Update
+  direct access to raw MusicBrainz response keys if needed.
 
 ..
     Other changes


amogus07:
I also did a similar sort of refactoring recently, and this is what I ended up with: https://github.com/prTopi/beets-vocadb/tree/eeb047027a54d331cfe1d54bc1126046c5382a9b/src/beetsplug/_utils/vocadb/vocadb_api_client

Feel free to take inpiration!

Here are some suggestions:
1. `TypedDict`s are only referenced in type annotations, so they can all be put under `if TYPE_CHECKING:`
2. Using `StrEnum`s instead of Literals (or a custom `Enum` type [like this](https://github.com/prTopi/beets-vocadb/blob/65d21d3c752691f8da34e200232e3fa6919a3ddc/beetsplug/vocadb/vocadb_api_client/models/__init__.py#L22) that overrides auto()).
3. Perhaps it's worth putting the models into their own module or even package.

snejus:
>     1. `TypedDict`s are only referenced in type annotations, so they can all be put under `if TYPE_CHECKING:`

I don't see any benefit for doing so (this may actually increase verbosity because of shorter line lengths and reformatting). Sometimes they should be defined _outside_ the `TYPE_CHECKING` block, e.g. see `beets.util.color.get_color_config`.

>     2. Using `StrEnum`s instead of Literals (or a custom `Enum` type [like this](https://github.com/prTopi/beets-vocadb/blob/65d21d3c752691f8da34e200232e3fa6919a3ddc/beetsplug/vocadb/vocadb_api_client/models/__init__.py#L22) that overrides auto()).

Again, I don't see any benefit for this - these are just static definitions without any functionality attached to them.

>     3. Perhaps it's worth putting the models into their own module or even package.

These definitions _belong_ to the `MusicBrainzAPI` and I think they should be tightly coupled. I don't want to force users of this API to import the definitions form another module.

beetsplug/musicbrainz.py
```diff
@@ -200,18 +211,20 @@ def _preferred_release_event(
...
     for country in preferred_countries:
-        for event in release.get("release-events", {}):
+        for event in release.get("release_events", {}):
             try:
                 if area := event.get("area"):
-                    if country in area["iso-3166-1-codes"]:
+                    if country in area["iso_3166_1_codes"]:
                         return country, event["date"]
             except KeyError:
                 pass
```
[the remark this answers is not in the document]
snejus: out of scope

setup.cfg
```diff
@@ -50,3 +50,12 @@ check_untyped_defs = true
...
 disallow_untyped_decorators = true
 check_untyped_defs = true
+
+[[mypy-beetsplug.musicbrainz]]
+strict = true
+
+[[mypy-beetsplug.mbpseudo]]
+strict = true
+
+[[mypy-beetsplug._utils]]
```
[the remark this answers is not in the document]
snejus: That's intentional.

beetsplug/_utils/musicbrainz.py
```diff
@@ -323,13 +737,12 @@ def _(cls, data: JSONDict) -> JSONDict:
                         {k: v for k, v in item.items() if k != "target-type"}
                         for item in group
                     ]
-                    data[f"{target_type}-relations"] = cls._group_relations(
-                        relations
+                    output_data[f"{target_type}_relations"] = (
```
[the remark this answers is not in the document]
snejus: Love you, thanks

beetsplug/_utils/musicbrainz.py
```diff
@@ -120,6 +120,406 @@ class BrowseRecordingsKwargs(BrowseReleaseGroupsKwargs, total=False):
 R = TypeVar("R")
 
 
+class _Period(TypedDict):
```
semohr, line 123: Do we want to separate the types into another file? I get why you put them here but they are quite verbose. I tend/try to separate logic and "pure" types.

In the recent https://github.com/beetbox/beets/pull/6520 PR I introduced something quite similar and placed the api types in an additional file. Maybe we want to align here?

beetsplug/_utils/musicbrainz.py
```diff
@@ -120,6 +120,406 @@ class BrowseRecordingsKwargs(BrowseReleaseGroupsKwargs, total=False):
...
+    begin: str | None
+    end: str | None
+    ended: bool
+
+
+class Alias(_Period):
```
semohr, line 129: I was wondering where the types come from. Is there a musicbrainz schema or do they come from experience working with musicbrainz/ reverse engineering?

If there is a reference maybe we should put it here for future us.

amogus07:
> > ```
> > 2. Using ``StrEnum``s instead of Literals (or a custom ``Enum`` type [like this](https://github.com/prTopi/beets-vocadb/blob/65d21d3c752691f8da34e200232e3fa6919a3ddc/beetsplug/vocadb/vocadb_api_client/models/__init__.py#L22) that overrides auto()).
> > ```
> 
> Again, I don't see any benefit for this - these are just static definitions without any functionality attached to them.

Enums can be trivially enforced at runtime, which makes it easier to catch typos. But I can see how that's not relevant when we're using TypedDict which isn't enforced at runtime either.

beetsplug/_utils/musicbrainz.py
snejus: They are indeed verbose but they belong here as they are only related to the musicbrainz API client. I thought about this but it would make things a bit awkward structurally:

```
$ tree beetsplug/_utils
beetsplug/_utils
├── art.py
├── __init__.py
├── musicbrainz.py
├── playcount.py
├── requests.py
└── vfs.py
```

Creating `musicbrainz_types.py` or a new directory for the types and `musicbrainz.py` is just a bit awkward, so I decided to keep them in the same file. I also expect to import the types from the same module where `MusicBrainzAPIClient` is defined.

beetsplug/_utils/musicbrainz.py
snejus: I fetched all albums and singletons tagged by `musicbrainz` in my library using the same params that `musicbrainz` plugin does, and asked Codex to give me typed dicts. I had to make a couple of adjustments regarding some edge cases that were not present in the data but exist in the codebase.

Unfortunately, I could not find a reference online that documents the responses.

snejus:
> I was wondering if one could have omitted the normalization of the hypens by using the TypeDict functional syntax (see [here](https://docs.python.org/3/library/typing.html#typing.TypedDict)).

Back in the day when I started this work, I created these types by hand using functional syntax but found the definitions too awkward to read. Thus I decided to just normalise the dashes to underscores for consistency, hoping that this is going to be easier to maintain going forward.

beetsplug/_utils/musicbrainz.py
semohr: Maybe this is an indicator to migrate to a musicbrainz folder for the plugin + all its dependencies?  :smile:

snejus: Yeah, that's a fair point!

semohr:
> Enums can be trivially enforced at runtime, which makes it easier to catch typos. But I can see how that's not relevant when we're using TypedDict which isn't enforced at runtime either.

The added types are not used at runtime at all or am I missing something?

beetsplug/_utils/musicbrainz.py
snejus: On the other hand the reason why `musicbrainz.py` is under `beetsplug._utils` is that multiple plugins depend on it. If we move it under `musicbrainz/client.py` we will end up ~enabling `musicbrainz` plugin~ importing `musicbrainz` source attempting to import the client in, say, `mbcollection`.

semohr: Would we introduce a circular dependency if we just import from `beetplug.musicbrainz.client`? I dont really see the difference to it being in utils other than for convention.

semohr: A underlying question is how do we want to handle api layers going forward? We removed the `musicbrainzapi` dependency at some point and this kinda started this (I guess?). Should all api/client definitions go into the `_utils` folder and we try to keep the beetsplug folder flat?

In the end it might just be a decicion how we want to handle this. Alternatively, we could also just say either way is fine :man_shrugging: Not that I have a strong opinion here. I just like unification if possible.

snejus: Nope - the issue is rather structural - `mbcollection`, `missing`, `parentwork` and `listenbrainz` should not depend on the `musicbrainz` plugin for the client. Thus the client should be imported from a 'neutral' location 😅

snejus: > Should all api/client definitions go into the `_utils` folder and we try to keep the beetsplug folder flat?

Only clients used by multiple plugins. See `lyrics.py` for example, which defines its own client in the same module, because it's not used anywhere else.

beetsplug/_utils/musicbrainz.py
semohr: > Unfortunately, I could not find a reference online that documents the responses.

Too bad. I found this with a quick search, not sure if and how it is related:
https://github.com/metabrainz/mmd-schema/blob/master/schema/musicbrainz_mmd-2.0.rng

snejus: I wish they had clear JSON documentation! In any case, note these types are specific to the `include` parameters we provide:

```py
RELEASE_INCLUDES = [
    "artists",
    "media",
    "recordings",
    "release-groups",
    "labels",
    "artist-credits",
    "aliases",
    "recording-level-rels",
    "work-rels",
    "work-level-rels",
    "artist-rels",
    "isrcs",
    "url-rels",
    "release-rels",
    "genres",
    "tags",
]

RECORDING_INCLUDES = [
    "artists",
    "aliases",
    "isrcs",
    "work-level-rels",
    "artist-rels",
]
```

beetsplug/_utils/musicbrainz.py
amogus07: how about this:
```
beetsplug/_utils
├── art.py
├── __init__.py
├── musicbrainz
│   ├── __init__.py
│   └── models.py
├── playcount.py
├── requests.py
└── vfs.py
```
or is this what you already dismissed @snejus ?

snejus: Yup, this sort of structure was included in what I mentioned here:

> Creating `musicbrainz_types.py` or a new directory for the types and `musicbrainz.py` is just a bit awkward, so I decided to keep them in the same file. I also expect to import the types from the same module where `MusicBrainzAPIClient` is defined.

amogus07:
> @amogus07 Do you have any strong opinions which could block this?

no, feel free to merge it if you want

PR #6333, opened 2026-02-02
Fix dependencies

### LINKED ISSUE #6332: Install of 2.6.0 with pipx is broken
### Problem

Running this command installs beets 2.5.1 via pipx using python 3.13:

```sh
$ pipx install —python 3.13 beets==2.5.1
```

Installs without error as expected, I am able to run beets via `beet` just fine.

However, doing something similar to install beets 2.6.0:

```sh
$ pipx install —python 3.13 beets==2.6.0
```

Pipx installs it successfully, but when I then run `beet` with or without arguments I get the following error:

```
  File "/Users/boldandbrad/.local/bin/beet", line 3, in <module>
    from beets.ui import main
  File "/Users/boldandbrad/.local/pipx/venvs/beets/lib/python3.13/site-packages/beets/__init__.py", line 20, in <module>
    from .util.deprecation import deprecate_imports
  File "/Users/boldandbrad/.local/pipx/venvs/beets/lib/python3.13/site-packages/beets/util/deprecation.py", line 7, in <module>
    from packaging.version import Version
ModuleNotFoundError: No module named 'packaging'
```

`packaging` looks like a dependency in [pyproject.toml](https://github.com/beetbox/beets/blob/680473b9e5ab3c2ee6e4083467dd7dc082688799/pyproject.toml#L135). So I wonder if pipx just isn’t installing it properly.

Surely I’m dumb and I'm somehow screwing something up here. I just don’t know why installing 2.5.1 works fine and 2.6.0 doesn’t. I appreciate any help. Thank you!

### Setup

* OS: macos 26.3
* Python version: 3.13 (although I tried with 3.12 as well with the same result)
* beets version: 2.6.0
* Turning off plugins made problem go away (yes/no): never had any enabled. this was a fresh install

```yaml
default config
```

snejus:
Fixes #6332 

- Promotes `packaging` from a release-only dependency to a required runtime dependency by moving it into `pyproject.toml`'s main dependencies.

- Updates `poetry.lock` to pick up patched versions of vulnerable libraries, notably `brotli` (`1.1.0` → `1.2.0`), `urllib3` (`2.5.0` → `2.6.3`), `werkzeug` (`3.1.3` → `3.1.5`), and `filelock` (`3.20.2` → `3.20.3`).

image

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -284,136 +284,111 @@ files = [
 
 [[package]]
 name = "brotli"
-version = "1.1.0"
+version = "1.2.0"
 description = "Python bindings for the Brotli compression library"
 optional = false
 python-versions = "*"
 files = [
-    {file = "Brotli-1.1.0-cp310-cp310-macosx_10_9_universal2.whl", hash = "sha256:e1140c64812cb9b06c922e77f1c26a75ec5e3f0fb2bf92cc8c58720dec276752"},
-    {file = "Brotli-1.1.0-cp310-cp310-macosx_10_9_x86_64.whl", hash = "sha256:c8fd5270e906eef71d4a8d19b7c6a43760c6abcfcc10c9101d14eb2357418de9"},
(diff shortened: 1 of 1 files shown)


snejus:
Merging it right away for a patch release.

PR #6341, opened 2026-02-02
Improve cached_classproperty type annotations

amogus07:
## Description

Basically a followup to https://github.com/beetbox/beets/pull/6110

## To Do

- [x] ~Documentation~
- [x] ~Changelog~
- [x] ~Tests~

### DIFF
--- a/beets/autotag/distance.py
+++ b/beets/autotag/distance.py
@@ -131,6 +131,7 @@ class Distance:
         self.tracks: dict[TrackInfo, Distance] = {}
 
     @cached_classproperty
+    @classmethod
     def _weights(cls) -> dict[str, float]:
         """A dictionary from keys to floating-point weights."""
         weights_view = config["match"]["distance_weights"]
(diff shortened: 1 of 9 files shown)


snejus:
What does addition of T2 solve/improve here? I'm fine with the new `cache_clear` method.

amogus07:
T is the class the decorated method belongs to, T2 is it's return type

snejus:
Do you know if it's possible to somehow force `mypy` to think that `cached_classproperty` accepts a `cls` (type of the class) instead of `self`?

amogus07:
would someone mind having a look at the mypy error? I'm not sure what to do about it

snejus:
This is too complicated - as I said previously, at this point we only want the method to clear the cache.

We do not want to require to specify `@classmethod` everywhere this is used. Instead, we'd rather somehow incorporate it into `cached_classproperty` definition so that we only need to use a single `@cached_classproperty` decorator.

PR #6383, opened 2026-02-20
lastgenre: Apply canonicalization to genres of items with no Last.fm results

Bergbok:
I've been configuring my canonicalization tree to remove tags I don't like, essentially using it like a blacklist, and noticed some tags not being removed after re-running the lastgenre command.

The reason for this is that when `keep_existing` is enabled and no Last.fm results are found, the existing genres are returned as-is without being processed through `_resolve_genres()`. This means existing genres that should be filtered by the canonicalization tree are kept unchanged.

As an example, if I had this in my canonicalization tree:

```yaml
- '':
  - 'some tag'
```

and then ran lastgenre on a track with 'some tag', but which had no data on Last.fm, it wouldn't get removed from the track. This PR fixes that.

### DIFF
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -480,15 +480,11 @@ class LastGenrePlugin(plugins.BeetsPlugin):
 
         # Nothing found, leave original if configured and valid.
         if obj.genre and self.config["keep_existing"]:
-            if not self.whitelist or self._is_valid(obj.genre.lower()):
-                return obj.genre, "original fallback"
-            else:
-                # If the original genre doesn't match a whitelisted genre, check
-                # if we can canonicalize it to find a matching, whitelisted genre!
-                if result := _try_resolve_stage(
-                    "original fallback", keep_genres, []
-                ):
-                    return result
+            # Process existing genres to apply whitelist and canonicalization
+            if result := _try_resolve_stage(
+                "original fallback", keep_genres, []
+            ):
+                return result
 
         # Return fallback string.
         if fallback := self.config["fallback"].get():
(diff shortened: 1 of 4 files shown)


JOJ0:
> I've been configuring my canonicalization tree to remove tags I don't like, essentially using it like a blacklist, and noticed some tags not being removed after re-running the lastgenre command.

A blacklist feature is in the pipeline but it will take a while until it's ready. 2 other lastgenre PR's need to go first. You could test from the branch but I suggest with a test library setup only for now: https://github.com/beetbox/beets/pull/5744

JOJ0:
Please let's wait with this one. There is other things to be done and I'm not entirely sure how this could/should be implemented. Sorry for the inconvenience and many thanks for the efforts! Speak soon!

JOJ0:
> > I've been configuring my canonicalization tree to remove tags I don't like, essentially using it like a blacklist, and noticed some tags not being removed after re-running the lastgenre command.
> 
> A blacklist feature is in the pipeline but it will take a while until it's ready. 2 other lastgenre PR's need to go first. You could test from the branch but I suggest with a test library setup only for now: #5744

Hi again, I was working on getting the blacklist feature ready for a final review this weekeend. If you want you can test it from the `lastgenre_forbidden` feature branch: https://github.com/beetbox/beets/pull/5744

It would help if you could report if it suits your use-cases. Thanks and have a great week!

JOJ0:
@Bergbok I'm closing this one. I think using `ignorelist` and maybe even `cleanup_existing` fits your use case.

Bergbok:
> @Bergbok I'm closing this one. I think using `ignorelist` and maybe even `cleanup_existing` fits your use case.

The `ignorelist` option works great. Being able to use regex cut my blacklist down from a couple thousand lines to 500. Thanks a lot!

PR #6399, opened 2026-02-27
Clarify %if template behavior

### LINKED ISSUE #4991: Variant of `%if` template function that does not treat 0 as false
As per the [documentation](https://beets.readthedocs.io/en/stable/reference/pathformat.html#template-functions): 
> `%if{condition,text}` or `%if{condition,truetext,falsetext}`: If `condition` is nonempty (or nonzero, if it’s a number), then returns the second argument. 

Is there a way this could be context dependent and act differently for different fields? I have an album simply titled "0", which failed the check `%if{album,album/}`. I've already rewritten my path format to avoid this issue, but it took me a while to realize what was wrong. Checking for 0 makes sense when you're checking numbers, but when you're checking strings I think it's more intuitive for `0` to evaluate to `true`. 

Sorry if this has come up before, I tried searching but there's a lot to wade through searching for "if 0".

emiham:
## Description

Fixes #4991.  

I don't know if there's a single, good workaround for this, so I opted to not add an example solution. 

## To Do

### DIFF
--- a/docs/reference/pathformat.rst
+++ b/docs/reference/pathformat.rst
@@ -62,9 +62,9 @@ These functions are built in to beets:
 - ``%left{text,n}``: Return the first ``n`` characters of ``text``.
 - ``%right{text,n}``: Return the last ``n`` characters of ``text``.
 - ``%if{condition,text}`` or ``%if{condition,truetext,falsetext}``: If
-  ``condition`` is not empty, and not one of the values ``0`` or ``false``
-  (case insensitive), then returns the second argument. Otherwise, returns the
-  third argument if specified (or nothing if ``falsetext`` is left off).
+  ``condition`` is not empty, and not one of the values ``0`` or ``false`` (case
+  insensitive), then returns the second argument. Otherwise, returns the third
+  argument if specified (or nothing if ``falsetext`` is left off).
 - ``%asciify{text}``: Convert non-ASCII characters to their ASCII equivalents.
   For example, "café" becomes "cafe". Uses the mapping provided by the
   `unidecode module`_. See the :ref:`asciify-paths` configuration option.


emiham:
> This is a good clarification for the behaviour you had a problem with, but I wonder if this is something that only makes sense for programmers. Then again, this might be something that only programmers or people somewhat familiar with programming conventions understand. Might not be a problem in that case.

I agree that it might be confusing to some people not having the prerequisite programming knowledge, but I still think those people would be better off with more documentation that can at least point them in the right direction/help them ask the right questions when asking for help. 

> Perhaps make it clear that for different data types, some are always interpreted as False. 0 is that for an integer, False for a boolean, '' for a string, etc. 

This is a great point, I had not considered this at all! ~~I can't figure out where in the code the path format actually gets interpreted, do you think you could point me to that? I did try importing two albums named "False" and "false", which both evaluate as `False`, so that should definitely also be mentioned. I'm curious about what other values might run into this though.~~ Edit: I found the relevant code.

docs/reference/pathformat.rst
```diff
@@ -64,7 +64,9 @@ These functions are built in to beets:
 - ``%if{condition,text}`` or ``%if{condition,truetext,falsetext}``: If
   ``condition`` is nonempty (or nonzero, if it's a number), then returns the
   second argument. Otherwise, returns the third argument if specified (or
-  nothing if ``falsetext`` is left off).
+  nothing if ``falsetext`` is left off). Note that beets doesn't distinguish
+  between different data types, so ``%if{album,album/}`` evaluates as false for
+  albums named exactly "0".
```
snejus: I think you probably meant to use `$album` here, however, I don't think this is correct. For example, see where this is used in my configuration:

```yaml
paths:
  albumtype:broadcast: |
    sets/
    %lower{
      %if{$album,$album/}
      $track_identification$track_artist$title
    }
```

Here I append a slash to the album name **if the album name is present**. It will be false only when `$album` is empty, i.e. `""`.

emiham: Sorry yes you're right, that is what I meant. Will fix. 

> Here I append a slash to the album name if the album name is present. It will be false only when $album is empty, i.e. "". 

Or when the album name is `0` or `false`, which is the behavior I'm trying to clarify.

emiham:
I've changed the wording to explicitly mention which values are interpreted as false, so I think it should be clear for people without programming knowledge now as well. 

> Would you mind adding some examples for the `%if` function behaviour, following what we have under `%first{text,count,skip,sep,join}` below?

I wouldn't mind doing this, but I can't think of what examples to add. My personal use case was the same as yours, i.e. creating an album only if the album field is populated, but I don't think that should be recommended unless you also add a workaround for albums with falsy names.

snejus:
Sounds good to me - now just rebase it on master and run `poe format-docs` to satisfy the CI.

emiham:
That's weird, I did run `poe format-docs`. I just tried again and it doesn't report anything, even though the CI fails.

snejus:
Interesting. I have just ran it on my end and it does indeed reformat `pathformat.rst`. Maybe try running `poetry install` to make sure the dependencies are in sync and re-run the command again?

emiham:
I tried again with a fresh venv, but it still refused to format anything. 

I managed to get it to work by adding some trailing spaces before running `poe format-docs`, that made it detect that the document needed formatting.

snejus:
That's smart 👍🏼 We updated docstrfmt now, so hopefully this issue is fixed.

PR #6417, opened 2026-03-06
Improve regex detection for the drive_sep_replace default

spaceage64:
## Description

I imported an album where a track had the name `1:00 AM - Clear` and another track named `12:00 AM - Clear` (just two examples). 

See: [Animal Crossing: New Horizons OST](https://musicbrainz.org/release/263f7ed3-60c2-4251-ac7d-6da3f8691256)

After import, the former was renamed `1_00 AM - Clear`, and the latter `12;00 AM - Clear`. Notice the inconsistency of how the `:` was replaced.

I did not make use of the (hidden) `drive_sep_replace` setting. These were my `replace` settings:
```
replace:                        # prevent file name incompatibiliy
  '[\s]'                        : ' '   # standardize whitespace
  '["`‘’“”]'                    : "'"   # standardize quotes
  '[\u002D\u2010-\u2015\u2E3A]' : '-'   # standardize dashes
  '[\u2E3B\uFE58\uFE63\uFF0D]'  : '-'   # standardize dashes
  '[\xAD]'                      : '-'   # standardize dashes
  '[\\\|\/]'                    : ' '   # slashes, pipe > space
  '[:]'                         : ';'   # colon > semicolon
  '[<>]'                        : '-'   # chevrons > dashes
  '[\?\*]'                      : ''    # remove restricted characters
  '[\x00-\x1F\x7F]'             : ''    # remove basic control characters
  '[\x80-\x9F]'                 : ''    # remove extra control characters
  '^\.'                         : ''    # remove leading period
  '\.$'                         : ''    # remove trailing period
  '^\s+'                        : ''    # remove leading space
  '\s+$'                        : ''    # remove trailing space
```

I found the issue to be too generic regex for drive separator detection. I'm on macOS, so this is irrelevant to me anyway (and I got around it by adding `drive_sep_replace: ';'` to my settings), but regardless, I think this could be improved.

This PR improves the regex to detect drive separators. Instead of merely looking for any first character followed by a colon (`^\w:`), we look for a letter, followed by a colon, followed by a backslash instead (`^[a-zA-Z]:\\`).

The regex logic is solid, but I am not able to test this on a real Windows environment.

~Still have to add an entry to the changelog, will do so soon.~

# Update

Initially this commit failed the `MoveTest.test_move_file_with_colon_alt_separator` test because it checks the logic using a `C:DOS` path. So I had to make the logic less restrictive again, not checking for a backslash (`^[a-zA-Z]:`). I would argue the test itself should be amended (test with `C:\DOS` instead), but that's not up to me.

As a result, my case of "1:00 AM" being replaced incorrectly is still resolved, but other hypothetical cases like "a:b" would still not be covered due to an arguably incorrect test limiting a more precise regex.

## To Do

- [X] ~Documentation.~ (This is an existing feature, just refined logic)
- [X] Changelog. (Added an entry to the Bugs section)
- [ ] Tests. (I tested the regex, but did not run this on a real Windows setup)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -16,6 +16,9 @@ Unreleased
 Bug fixes
 ~~~~~~~~~
 
+- :ref:`replace`: Made ``drive_sep_replace`` regex logic more precise to prevent
+  edge-case mismatches (e.g., a song titled "1:00 AM" would incorrectly be
+  considered a Windows drive path).
 - :doc:`plugins/fish`: Fix AttributeError. :bug:`6340`
 
 ..
@@ -87,9 +90,6 @@ New features
 Bug fixes
 ~~~~~~~~~
 
-- :ref:`replace`: Made ``drive_sep_replace`` regex logic more precise to prevent
-  edge-case mismatches (e.g., a song titled "1:00 AM" would incorrectly be
-  considered a Windows drive path).
 - :doc:`plugins/ftintitle`: Fix handling of multiple featured artists with
   ampersand.
 - :doc:`plugins/zero`: When the ``omit_single_disc`` option is set,
@@ -109,9 +109,6 @@ Bug fixes
   ``duplicate_action`` config options were silently ignored for as-is imports.
 - :doc:`/plugins/convert`: Fix extension substitution inside path of the
   exported playlist.
-- :ref:`replace`: Made ``drive_sep_replace`` regex logic more precise to prevent
-  edge-case mismatches (e.g., a song titled "1:00 AM" would incorrectly be
-  considered a Windows drive path).
 
 For plugin developers
 ~~~~~~~~~~~~~~~~~~~~~


spaceage64:
I see I failed one formatting test for the changelog. Can someone advise on this?

snejus:
> I see I failed one formatting test for the changelog. Can someone advise on this?

Try running `poe format-docs`?

spaceage64:
> > I see I failed one formatting test for the changelog. Can someone advise on this?
> 
> Try running `poe format-docs`?

Thanks but  have and this error remains. Maybe I'm doing something wrong, but I don't know what. I only added an entry to the changelog, nothing special. No idea why the test would consider the formatting an issue now?

snejus:
Ah I see what's happening - you're using a new `docstrfmt` version.

Run `poetry install` to sync the dependencies and re-run `poe format-docs`.

snejus:
Ah one last thing - we've had a couple of releases since this PR was submitted - would you mind moving your changelog note to the top, under the **Unreleased** section?

PR #6424, opened 2026-03-08
play: Add -R/--randomize option to `play` plugin

amiv1:
## Description

I really wanted to combine `random` with `play` to shuffle tracks in a temporary playlist, but couldn't find easy way to do it. I'd like to introduce a new flag `-R / --randomize` to the `play` plugin to shuffle tracks before passing them to the player.

I couldn't really decide between `-r` and `-R`, not sure if there's any rule there. If you believe lower case would be better, just let me know.

## To Do

### DIFF
--- a/docs/plugins/play.rst
+++ b/docs/plugins/play.rst
@@ -120,6 +120,13 @@ The ``--yes`` (or ``-y``) flag to the ``play`` command will skip the warning
 message if you choose to play more items than the **warning_threshold** value
 usually allows.
 
+The ``--randomize`` (or ``-R``) flag shuffles the order of playlist entries
+before passing it to the player:
+
+::
+
+    $ beet play --randomize my query
+
 Note on the Leakage of the Generated Playlists
 ----------------------------------------------
 
--- a/test/plugins/test_play.py
+++ b/test/plugins/test_play.py
@@ -145,6 +145,38 @@ class PlayPluginTest(IOMixin, CleanupModulesMixin, PluginTestCase):
             expected_playlist=expected_playlist,
         )
 
+    def _playlist_lines(self, open_mock):
+        """Read the playlist file passed to interactive_open and return its lines."""
+        # interactive_open is called as: interactive_open([playlist_path], command)
+        playlist_path = open_mock.call_args[0][0][0]
+        with open(playlist_path, "rb") as playlist:
+            return playlist.read().decode("utf-8").splitlines()
+
+    def _add_many_ordered_items(self, *, count, album):
+        items = []
+        for track in range(1, count + 1):
+            items.append(
+                self.add_item(
+                    album=album,
+                    artist="randomize artist",
+                    title=f"randomize {track:03d}",
+                    track=track,
+                )
+            )
+        return items
+
+    def test_randomize(self, open_mock):
+        album = "randomize_test"
+        items = self._add_many_ordered_items(count=100, album=album)
+        baseline = [str(item.filepath) for item in items]
+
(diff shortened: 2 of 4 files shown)


test/plugins/test_play.py
```diff
@@ -145,6 +145,22 @@ def test_skip_warning_threshold_bypass(self, open_mock):
             expected_playlist=expected_playlist,
         )
 
+    def test_randomize(self, open_mock):
+        self.other_item = self.add_item(title="anotherTitle")
+
+        with patch("beetsplug.play.random.shuffle") as shuffle_mock:
+            self.run_command("play", "-R", "NiceTitle")
+
+        shuffle_mock.assert_called_once()
```
snejus: I'm not a fan of _implementation_ tests - I'd rather want to see an assertion that tests _behaviour_. Would it be possible to test the output here somehow? We'd be expecting the paths to be out of order.

amiv1: @snejus Hi. I totally agree with you, though the problem with testing randomization is that the result is.. random 😄 If randomized list turns out to be exactly the same as the input, which is totally possible, the test will fail.

Though I could pass a list of 100 or more entries, do multiple iterations and assert that at least one of them is different than the input array. This should reduce the chances of test flakiness. Anyway, I'll check what can be done about it.

docs/plugins/play.rst
```diff
@@ -120,6 +120,13 @@ The ``--yes`` (or ``-y``) flag to the ``play`` command will skip the warning
 message if you choose to play more items than the **warning_threshold** value
 usually allows.
 
+The ``--randomize`` (or ``-R``) flag shuffles the order of tracks in the
+playlist before passing it to the player:
```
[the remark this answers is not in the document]
amiv1: @snejus Do you think it would make any difference with the suggested text or I can just close it without a change?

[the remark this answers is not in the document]
snejus: Just adjust it to say playlist **entries**.

test/plugins/test_play.py
```diff
@@ -145,6 +145,55 @@ def test_skip_warning_threshold_bypass(self, open_mock):
...
+            )
+        return items
+
+    def test_randomize(self, open_mock):
+        album = "randomize_test"
+        items = self._add_many_ordered_items(count=100, album=album)
+        baseline = [str(item.filepath) for item in items]
+
+        changed = False
+        for _ in range(10):
```
snejus: I feel like one run should be sufficient here - the probability that all items are going to turn out in the exact same order is minimal.

test/plugins/test_play.py
```diff
@@ -145,6 +145,55 @@ def test_skip_warning_threshold_bypass(self, open_mock):
...
+    def test_no_randomize_by_default(self, open_mock):
+        album = "no_randomize_default_test"
+        items = self._add_many_ordered_items(count=100, album=album)
+        baseline = [str(item.filepath) for item in items]
+
+        for _ in range(5):
+            open_mock.reset_mock()
+            self.run_command("play", f"album:{album}")
+            assert self._playlist_lines(open_mock) == baseline
+
```
snejus: I don't think there's a need for this test. Alternatively, assert the order in one of the existing tests if you really want to but again, I think we can get by without this.

test/plugins/test_play.py
amiv1: Ok, I guess I'm too paranoid :) removed the loop

snejus (approved):
Thanks! Just need to resolve a conflict in the changelog.

amiv1:
@snejus Rebase done

PR #6433, opened 2026-03-09
Add `extra_tags` support to Discogs plugin

jdoe29103:
## Description

Fixes #6412.  

This is my first time submitting a PR for an open source project so please point out any mistakes!

## Summary

- Add `discogs.extra_tags` configuration option to narrow Discogs search queries using existing tag values.
- Map supported tags (`barcode`, `catalognum`, `country`, `label`, `media`, `year`) to corresponding Discogs search parameters.
- Update Discogs plugin documentation and tests to cover the new behavior.

## Details

The Discogs plugin now mirrors `musicbrainz.extra_tags` by allowing users to specify additional tags that should be used when building Discogs search filters.

- New config option: `discogs.extra_tags` (default: `[]`).
- Supported tags and their Discogs search parameters:
  - `barcode` → `barcode`
  - `catalognum` → `catno` (whitespace removed)
  - `country` → `country`
  - `label` → `label`
  - `media` → `format`
  - `year` → `year`
- Tags `alias` and `tracks` are recognized but intentionally ignored for Discogs, since the Discogs API does not provide direct equivalents for these MusicBrainz-specific fields.

When `extra_tags` are configured, the plugin uses `beets.util.plurality` over the items in the import session to select the most common value for each configured tag and adds the corresponding Discogs filter.

## Testing

- Added unit tests in `test/plugins/test_discogs.py` to verify:
  - Default search filters remain unchanged when `extra_tags` is not set.
  - `discogs.extra_tags: [label, catalognum]` results in `label` and `catno` filters populated from library items (with catalog number whitespace stripped).
- Ran:
  - `pytest test/plugins/test_discogs.py`
  - `pytest test/plugins/test_musicbrainz.py`

(...)

## To Do

- [X] Documentation.
- [X] Changelog.
- [X] Tests.

### DIFF
--- a/docs/plugins/discogs.rst
+++ b/docs/plugins/discogs.rst
@@ -148,7 +148,8 @@ Default
 .. conf:: extra_tags
     :default: []
 
-    By default, beets will only use the artist and album information to query Discogs. Additional tags to be queried can be supplied with the
+    By default, beets will use only the artist and album to query Discogs.
+    Additional tags to be queried can be supplied with the
     ``extra_tags`` setting.
 
     This setting should improve the autotagger results if the metadata with the


snejus:
Hey, thanks for your contribution! I would suggest having a look at `MusicBrainzPlugin` at `beetsplug/musicbrainz.py`, namely `extra_mb_field_by_tag` property and `get_album_criteria` with `get_search_query_with_filters` methods. Ideally I'd want to see the same property and methods in `DiscogsPlugin` as well, if possible!

jdoe29103:
> Hey, thanks for your contribution! I would suggest having a look at `MusicBrainzPlugin` at `beetsplug/musicbrainz.py`, namely `extra_mb_field_by_tag` property and `get_album_criteria` with `get_search_query_with_filters` methods. Ideally I'd want to see the same property and methods in `DiscogsPlugin` as well, if possible!

Thanks for taking the time @snejus!
I’ve added an `extra_discogs_field_by_tag` cached property, mirroring `extra_mb_field_by_tag`, and updated `get_search_query_with_filters` to build Discogs search filters from the configured `extra_tags`.
Because the Discogs API is query+filters search (rather than a criteria dictionary), I didn’t introduce a separate `get_album_criteria` helper, but the behavior should be aligned with the MusicBrainz plugin.

beetsplug/discogs/__init__.py
```diff
@@ -256,7 +285,31 @@ def get_search_query_with_filters(
...
+
+        if not items:
+            return query, filters
+
+        for tag, api_field in self.extra_discogs_field_by_tag.items():
+            # The Discogs search API does not provide direct equivalents for
+            # MusicBrainz "alias" or "tracks" search fields, so we ignore
+            # those tags if configured.
+            if tag in {"alias", "tracks"}:
+                continue
```
JOJ0: why is it required to remove those silently? if we remove that check, would a misconfiguration throw an error and inform the user that they should fix there config?

beetsplug/discogs/__init__.py
```diff
@@ -256,7 +285,31 @@ def get_search_query_with_filters(
...
+
+            most_common, _count = util.plurality(
+                item.get(tag) for item in items
+            )
+            if most_common is None:
+                continue
+
+            value = str(most_common)
+            if tag == "catalognum":
+                value = value.replace(" ", "")
```
JOJ0, line 302: aha interesting, so discogs api gives better results if whitespace is removed from queries on that field. did you find this out by trial and error or do discogs api docs state that?

test/plugins/test_discogs.py
```diff
@@ -466,6 +467,42 @@ def test_strip_disambiguation_false(self):
...
+            "album", items, "Artist", "Album", False
+        )
+
+        assert "Album" in query
+        assert filters == {"type": "release"}
+
+    def test_extra_tags_populate_discogs_filters(self):
+        """Configured extra_tags should populate Discogs search filters."""
+        config["discogs"]["extra_tags"] = ["label", "catalognum"]
+        plugin = DiscogsPlugin()
```
snejus: This way you won't need to clear the config in the last line of the test.

```suggestion
        plugin = DiscogsPlugin()
        plugin.config["extra_tags"] = ["label", "catalognum"]
```

jdoe29103: Applied, thanks!

beetsplug/discogs/__init__.py
jdoe29103: I don't think it's required. I went with this approach because I was attempting to mirror the musicbrainz.py implementation and just sort of left it there. I don't see why it couldn't be removed though.

beetsplug/discogs/__init__.py
jdoe29103: The catalognum whitespace stripping is copied from the MusicBrainz plugin’s `get_album_criteria` implementation, which normalizes catalognum by removing spaces before using it in searches.

I did a few manual Discogs queries while working on this and found that removing spaces is generally helpful (users often tag catalog with spaces, while Discogs catno fields are typically stored without them in my testing), but I didn’t see this explicitly documented.

beetsplug/discogs/__init__.py
snejus: Good point @JOJ0. In musicbrainz plugin, these two fields are handled separately (see `MusicBrainzPlugin.get_album_criteria`). You don't need this here, so remove this logic.

jdoe29103: @snejus done!

beetsplug/discogs/__init__.py
snejus: I do the same locally - I use `beets-importreplace` to remove spaces from catalogue numbers at import time.

docs/changelog.rst
```diff
@@ -9,9 +9,12 @@ below!
 Unreleased
 ----------
 
-..
-    New features
-    ~~~~~~~~~~~~
+New features
+~~~~~~~~~~~~
+
+- :doc:`plugins/discogs`: Add :conf:`plugins.discogs:extra_tags` option to use
```
snejus, line 15: Good to see that you're using `:conf:` here!

docs/plugins/discogs.rst
```diff
@@ -144,6 +145,31 @@ Default
 
     Configure the string used for noting featured artists. Useful if you prefer ``Featuring`` or ``ft.``.
 
+.. conf:: extra_tags
+    :default: []
+
+    By default, beets will only use the artist and album information to query Discogs. Additional tags to be queried can be supplied with the
```
snejus: Oh, weird that CI didn't pick this up: the docs should have an 80-character limit. Would you mind reformatting this paragraph?

jdoe29103: fixed!

PR #6445, opened 2026-03-16
Fix Windows fallback for beet config -e when no editor is set

### LINKED ISSUE #6436: error: Editor 'start' not found
I'm trying to setup a new beets install, when I run 'beet config -e' to create the default config file, I get an error.

### Problem

```sh
$ beet -vv config -e
```

Led to this problem:

```
error: Editor 'start' not found.
```

### Setup

* OS: Windows 11
* Python version: 3.14.2
* beets version: 2.7.1

Is the config file this command would have created the same as https://github.com/beetbox/beets/blob/master/beets/config_default.yaml ?

api2062:
## Description

Fixes #6436.

On Windows, `beet config -e` falls back to `open_anything()` when `VISUAL`/`EDITOR` are unset. That fallback returned `start`, but `start` is a `cmd.exe` builtin (not an executable), so `os.execlp()` raised `FileNotFoundError`.

This changes the Windows fallback command to `cmd /c start ""` so the builtin is invoked via the shell and file opening works as intended.

## Changes

- Update `open_anything()` Windows fallback to `cmd /c start ""`.
- Add regression test for `interactive_open()` with the Windows fallback command.
- Add regression test for `beet config -e` on simulated Windows with no editor env vars.
- Add changelog entry under Unreleased bug fixes.

## To Do

- [x] ~Documentation~
- [x] Changelog
- [x] Tests

### DIFF
--- a/test/test_util.py
+++ b/test/test_util.py
@@ -31,7 +31,7 @@ from beets.test import _common
 class UtilTest(unittest.TestCase):
     def test_open_anything(self):
         with _common.system_mock("Windows"):
-            assert util.open_anything() == "start"
+            assert util.open_anything() == 'cmd /c start ""'
 
         with _common.system_mock("Darwin"):
             assert util.open_anything() == "open"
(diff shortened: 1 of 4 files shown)


test/test_util.py
```diff
@@ -50,6 +50,13 @@ def test_interactive_open(self, mock_open, mock_execlp):
         util.interactive_open(["foo"], "bar")
         mock_execlp.assert_called_once_with("bar", "bar", "foo")
 
+    @patch("os.execlp")
```
snejus: No need for this test as it tests implementation, not the behaviour.

test/ui/commands/test_config.py
```diff
@@ -110,6 +111,22 @@ def test_edit_config_with_automatic_open(self):
             "please_open", "please_open", self.config_path
         )
 
+    def test_edit_config_with_windows_default_open(self):
```
snejus: Same as above

snejus (changes requested):
Thanks for this, just remove the two implementation tests and we're good to go!

snejus (approved):

PR #6451, opened 2026-03-20
Implemented issue 2572, Query to find albums with/without embedded cover art

### LINKED ISSUE #2572: Feature request: Query to find albums with/without embedded cover art

Wouter2devries:
Fixes #2572

Added has_images function to query items by embedded cover art presence.
User can now do beets list has_images:true or beets list has_images:false to find albums with our without cover art.

First time contributing, tried to follow the guidelines as closely as possible. If anything is not up to standard I will edit it as soon as possible.

## To Do

The following 3 are included in the pull request:

- [x] Documentation.
- [x] Changelog.
- [x] Tests.

### DIFF
--- a/beets/dbcore/types.py
+++ b/beets/dbcore/types.py
@@ -204,14 +204,6 @@ class PaddedInt(BasePaddedInt[int]):
     pass
 
 
-class NullPaddedInt(BasePaddedInt[None]):
-    """Same as `PaddedInt`, but does not normalize `None` to `0`."""
-
-    @property
-    def null(self) -> None:
-        return None
-
-
 class ScaledInt(Integer):
     """An integer whose formatting operation scales the number by a
     constant and adds a suffix. Good for units with large magnitudes.
--- a/beetsplug/ftintitle.py
+++ b/beetsplug/ftintitle.py
@@ -343,17 +343,6 @@ class FtInTitlePlugin(plugins.BeetsPlugin):
         )
         return True
 
-    @staticmethod
-    def find_bracket_position(
-        title: str, keywords: list[str] | None = None
-    ) -> int | None:
-        normalized = (
-            DEFAULT_BRACKET_KEYWORDS if keywords is None else tuple(keywords)
-        )
-        pattern = FtInTitlePlugin._bracket_position_pattern(normalized)
-        m: re.Match[str] | None = pattern.search(title)
-        return m.start() if m else None
-
     @classmethod
     def insert_ft_into_title(
         cls, title: str, feat_part: str, keywords: list[str] | None = None
--- a/beetsplug/listenbrainz.py
+++ b/beetsplug/listenbrainz.py
@@ -257,15 +257,3 @@ class ListenBrainzPlugin(MusicBrainzAPIMixin, BeetsPlugin):
         # Fetch and return tracks from the selected playlist
         playlist = self.get_playlist(selected_playlist.get("identifier"))
         return self.get_tracks_from_playlist(playlist)
-
-    def get_weekly_exploration(self):
-        return self.get_weekly_playlist("Exploration", most_recent=True)
-
-    def get_weekly_jams(self):
-        return self.get_weekly_playlist("Jams", most_recent=True)
-
-    def get_last_weekly_exploration(self):
-        return self.get_weekly_playlist("Exploration", most_recent=False)
-
-    def get_last_weekly_jams(self):
-        return self.get_weekly_playlist("Jams", most_recent=False)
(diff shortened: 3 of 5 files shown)


beets/library/models.py
```diff
@@ -805,6 +805,7 @@ def _getters(cls):
         getters = plugins.item_field_getters()
         getters["singleton"] = lambda i: i.album_id is None
         getters["filesize"] = Item.try_filesize  # In bytes.
+        getters["has_images"] = Item.has_cover_art
```
snejus: Let's use `has_cover_art` as we refer to images as `cover_art` everywhere

```suggestion
        getters["has_cover_art"] = Item.has_cover_art
```

beets/library/models.py
```diff
@@ -1098,6 +1099,18 @@ def try_filesize(self):
...
             return 0
 
+    def has_cover_art(self):
+        """Check if item has embedded cover art.
+        
+        Return True if images embedded in file, False otherwise.
+        If file unreadable or no images, return False.
+        """
+        try:
+            mediafile = MediaFile(syspath(self.path))
```
snejus: No need for `syspath` here, the value should already be standardized upon access. And I think you can simplify this to

```suggestion
            with suppress(OSError):
                return bool(MediaFile(self.path).images)
            
            return False
```

docs/reference/query.rst
```diff
@@ -402,6 +402,28 @@ ones you've already added to your beets library.
...
+Cover Art Queries
+-----------------
+
+You can query items by whether they have embedded cover art using the
+``has_images`` field. This is useful for finding tracks that are missing
+artwork without needing external tools.
+
+To find all tracks with embedded cover art:
+
+::
```
snejus: Best to use
```suggestion
.. code-block: shell
```

test/test_query.py
```diff
@@ -518,3 +518,26 @@ def lib(self, helper):
...
+        """Verify has_images in getters dict."""
+        getters = Item._getters()
+        assert "has_images" in getters
+        assert getters["has_images"] == Item.has_cover_art
+
+    def test_has_images_returns_boolean(self):
+        """Method always return boolean."""
+        item = _common.item()
+        result = item.has_cover_art()
+        assert isinstance(result, bool)
```
snejus: These are __implementation__ tests. Instead, test the behaviour:

1. Create an item _without_ an image
2. Create an item _with_ an image
3. Assert that the query `lib.query("has_cover_art:false")` returns the first item and vice versa.

Wouter2devries:
Thank you for all the suggestions. Yeah found the tests a bit lacking myself aswell, did not look at the helper.py enough the first time. Hopefully these are better!

test/test_query.py
```diff
@@ -518,3 +518,36 @@ def lib(self, helper):
...
+        item_without.store()
+
+        return helper.lib
+
+    def test_has_cover_art_getter_exists(self):
+        """Verify has_cover_art in getters dict."""
+        getters = Item._getters()
+        assert "has_cover_art" in getters
+        assert getters["has_cover_art"] == Item.has_cover_art
+
```
snejus: No need for this - the next test covers this explicitly :)

test/test_query.py
```diff
@@ -518,3 +518,36 @@ def lib(self, helper):
...
+    def test_query_true(self, lib):
+        assert {i.title for i in lib.items("has_cover_art:true")} == {"with_art"}
+
+    def test_query_false(self, lib):
+        assert {i.title for i in lib.items("has_cover_art:false")} == {"without_art"}
```
snejus: You can make use of `pytest.mark.parametrize` here, something like

```py
@pytest.mark.parametrize("query, expected_titles", [("has_cover_art:true", ("with_art")), ...])
def test_has_cover_art_query(self, query, expected_titles):
    ...
```

snejus:
And note that the changelog has a merge conflict.

Wouter2devries:
I think the test is better now. With the merge i believe i created two bugfixes sections in the docs, or someone else did i was a bit unsure. Is it currently correct? Apologies for the inconvenience, still a bit new to this.

docs/reference/query.rst
```diff
@@ -402,6 +402,30 @@ ones you've already added to your beets library.
 Path queries are case sensitive if the queried path is on a case-sensitive
 filesystem.
 
+Cover Art Queries
+-----------------
+
+You can query items by whether they have embedded cover art using the
+``has_images`` field. This is useful for finding tracks that are missing artwork
```
jackwilsdon: This refers to the field by its previous name (`has_images`).

Wouter2devries: Ah oops, missed one i guess

snejus:
I suspect you will have to move your changelog note under the **Unreleased** section now since had a release earlier today.

PR #6452, opened 2026-03-20
Add feature to exclude albums/songs during auto import

### LINKED ISSUE #3523: Add ability to exclude songs or albums from lyrics
### Use case

I have a copy of a Nine Inch Nails album called "The Fragile: Deviations 1", which is an instrumental variation of their album, "The Fragile." Since, it's instrumental, there are no lyrics*, but the lyrics plugin keeps populating the songs with the lyrics from "The Fragile." I've tried changing sources and the lyrics still populate, so maybe there's a fuzzy match going on somewhere. In any event, it would be helpful if I could exclude that one album from processing with the lyrics plugin.

\* - I think one song out of 37 might have a one word chorus somewhere.

### Solution

I think having an `exclusions` option where the user could supply a list of albums, songs, or patterns to skip fetching lyrics for would be very helpful.

### Alternatives

One alternative might be to add the lyrics as `''` on each track. In this case, the FLAC files came directly from the artist, so anyone using those files is going to have to manually fix things up before using the lyrics plugin. But, I think doing so, as long as the `force` option isn't set, would allow to "skip" over those tracks. Another is to import the album with the lyrics plugin disabled, then re-enable for all future imports. While both options work, they make it difficult to use beets to manage the library afterwards.

kelamg:
## Description

Fixes #3523.

- [ x] Documentation.
- [ x] Changelog.
- [ x] Tests.

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -1072,7 +1072,7 @@ class LyricsPlugin(LyricsRequestHandler, plugins.BeetsPlugin):
             query, _ = parse_query_string(query_str, Item)
         else:
             # matches nothing, so all items proceed normally
-            query = FalseQuery()  
+            query = FalseQuery()
 
         for item in filterfalse(query.match, task.imported_items()):
             self.add_item_lyrics(item, False)


docs/plugins/lyrics.rst
```diff
@@ -63,6 +64,21 @@ Default configuration:
...
+  during auto import. For example, to skip tracks from Bandcamp or with a Techno
+  genre:
+
+  .. code-block:: yaml
+
+      lyrics:
+        auto_ignore: |
+          data_source:bandcamp
+          ,
+          genre:techno
```
snejus: I admit my configuration was misleading, you should use `genres:techno` here instead as `genre` field is gone!

```suggestion
          genres:techno
```

test/plugins/test_lyrics.py
```diff
@@ -366,6 +367,97 @@ def test_set_additional_lyrics_info(
         with pytest.raises(AttributeError):
             item.lyrics_translation_language
 
+    @pytest.mark.parametrize(
```
snejus: I doubt we need such a detailed test for various queries since we already have `test_query.py`.

Instead, test just a single one of these queries.

beetsplug/lyrics.py
```diff
@@ -1064,8 +1067,14 @@ def func(lib: Library, opts, args) -> None:
...
 
     def imported(self, _, task: ImportTask) -> None:
-        """Import hook for fetching lyrics automatically."""
-        for item in task.imported_items():
+        if query_str := self.config["auto_ignore"].get():
+            query, _ = parse_query_string(query_str, Item)
+        else:
+            query = (
+                FalseQuery()
+            )  # matches nothing, so all items proceed normally
```
snejus: Add the comment above to remove the need to reformat.

```suggestion
            # matches nothing, so all items proceed normally
            query = FalseQuery()  
```

beetsplug/lyrics.py
```diff
@@ -1064,8 +1067,14 @@ def func(lib: Library, opts, args) -> None:
         return [cmd]
 
     def imported(self, _, task: ImportTask) -> None:
-        """Import hook for fetching lyrics automatically."""
```
snejus, line 1067: You may want to configure your AI agent to keep pre-existing docstrings in place ;)

snejus:
I see that lyrics integration test is failing. I think you have to make the following change:

```diff
diff --git a/test/plugins/lyrics_pages.py b/test/plugins/lyrics_pages.py
index 1df53f731..df57d05b2 100644
--- a/test/plugins/lyrics_pages.py
+++ b/test/plugins/lyrics_pages.py
@@ -426,6 +426,11 @@ def backend
     LyricsPage.make(
         "https://www.musica.com/letras.asp?letra=59862",
         """
+        Friday night arrives without a suitcase
+        Sunday morning creeping like a nun
+        Monday's child has learned to tie his bootlace
+        See how they run
+
         Lady Madonna, baby at your breast
         Wonders how you manage to feed the rest
```

kelamg:
@snejus Thanks for the quick review! Yes that makes sense but I wasn't sure if it was the desired fix. I will update now.

test/plugins/lyrics_pages.py
```diff
@@ -426,6 +426,11 @@ def backend(self) -> str:
     LyricsPage.make(
         "https://www.musica.com/letras.asp?letra=59862",
         """
+        Friday night arrives without a suitcase
++       Sunday morning creeping like a nun
```
snejus: Replace **+** characters with spaces and it should work.

kelamg: Oops missed that one

beetsplug/lyrics.py
kelamg: Haha classic

Issue #133, opened 2013-02-28
Use paths relative to music directory


ghost:
This issue was automatically migrated from Google Code.
_Original author: jherl...@gmail.com (August 30, 2012 00:17:11)_
_Original issue: https://github.com/google-code-export/beets/issues/432_

maxammann:
+1

barsanuphe:
Also interested in this. I recently had to dismantle my NAS, and add its hard drive (with the beets library) on my computer, ie on a different path. I had to import it all again. 
Looking back, I should have done a `beet move` before switching the hard drives, but I had assumed only relative paths were used, since beets knows the library path.

danlamanna:
Also interested in this, was surprised to find out this wasn't the default.

jrobeson:
@sampsyo : what's your opinion here?

sampsyo:
Relative paths would probably be good idea. The problem is that this would be a deceptively large change -- path handling is extremely pervasive in beets, so the changes would not be localized. It's possible the transition would be easier if we did #1409 (pathlib) first.

naggie:
This would be great, I use syncthing to share music between 4 devices, each with a different `directory`. It would be nice to be able to share the config and library database.

awenny:
While it's simple (if you understand SQL) you can change your path like this:

`update items set path = '/Newpath' || substr(path, <startposition>) where path like '/Oldpath%'` if opening your library file with the CLI `sqlite3` (or other sqlite client).
Also for the artpath: `update albums set artpath = '/Newpath' || substr(artpath, <startposition>) where artpath like '/Oldpath%'`.
First check, if what you're updating to is what you intended: `select path, substr(path, <startposition>) from items where path like '/Oldpath%'`.
The percent sign (%) acts like an asterisk (*) in the file system. Unfortunately, this is only a one-time task. Or you need to run it every time before you execute beets with a different path configuration.

But it would really be great to have relative paths stored.

GuilhermeHideki:
I think I used a docker container one time to have a "relative path", when organizing the music on a external usb drive, on two computers. The idea was having a layer between the beets and the data.
For example, on Windows, the path was `/music = D:\music`, but on Linux was `/music = /mnt/music`

billop:
Just ran into this issue after moving to a new computer. Ran `beet update` and proceeded to have my library database completely wiped as everything was removed. This would be a very welcome change that would make libraries much more portable.

dbogdanov:
[...]

PR #6460, opened 2026-03-23
Store relative paths in the DB to make beets library portable

snejus:
## Migrate Item & Album Paths to Library-Relative Storage

Fixes: #133

### Core Problem Solved

**Before**: Beets stored absolute file system paths in SQLite (e.g. `/home/user/Music/Artist/album/track.mp3`). This made library databases non-portable — moving the music directory or sharing a database across machines broke all path references.

**After**: Paths are stored relative to the music directory (e.g. `Artist/album/track.mp3`), and expanded back to absolute paths transparently on read. The database is now portable.

---

#### 1. Context Variable for Music Directory (`beets/context.py`)

A new `contextvars.ContextVar` (`_music_dir_var`) holds the active music directory, set once during `Library.__init__` via `context.set_music_dir()`. This avoids passing `library.directory` through every call stack.

```
Library.__init__()
    └─ context.set_music_dir(self.directory)   ← single write point
           ↓
    PathType.to_sql()  / PathType.from_sql()   ← read at DB layer
    pathutils.normalize_path_for_db()
    pathutils.expand_path_from_db()
```

#### 2. Path Relativization Moved to the DB Layer (`beets/dbcore/`)

Previously, path conversion lived in `Item._setitem` / `Item.__getitem__` — model-specific overrides. It is now pushed down into `PathType.to_sql` / `PathType.from_sql` in `dbcore/types.py`, through two helpers in the new `beets/dbcore/pathutils.py`:

| Helper | Direction | Behaviour |
|---|---|---|
| `normalize_path_for_db(path)` | write | Strips music dir prefix → relative path |
| `expand_path_from_db(path)` | read | Prepends music dir → absolute path |

All models using `PathType` (currently `Item` and `Album`) benefit automatically — no per-model overrides required.

#### 3. `PathQuery` Updated for Relative Storage (`beets/dbcore/query.py`)

Queries like `path:/home/user/Music/Artist` now normalize the search term to its relative form before hitting the database, so SQL comparisons match stored values correctly. Both `col_clause` (SQL path) and `match` (in-memory path) use `normalize_path_for_db`.

#### 4. One-Time Database Migration (`RelativePathMigration`)

Existing absolute paths in `path` and `artpath` columns are migrated on startup:

```
Library startup
    └─ _migrate()
           └─ RelativePathMigration._migrate_data(Item, Album)
                  └─ _migrate_field("path")
                  └─ _migrate_field("artpath")
                         ↓
                  Reads rows where field starts with b"/"
                  Writes os.path.relpath(path, music_dir) in batches
```

`self.directory` assignment was moved **before** `super().__init__()` in `Library.__init__` so the migration can access the music dir when it runs.

#### 5. Context Propagation to Background Threads

The music dir context variable must be available in worker threads (pipeline stages, replaygain pool). Two propagation points were added:

- **`beets/util/pipeline.py`**: `Pipeline.run_parallel()` snapshots the calling context with `contextvars.copy_context()` and passes a per-thread copy to each `PipelineThread`. Each stage coroutine is invoked via `ctx.run(...)`.
- **`beetsplug/replaygain.py`**: Pool workers and their callbacks are wrapped in `ctx.run(...)` so `expand_path_from_db` works correctly inside the process pool.
- **`beets/util/__init__.py`**: `par_map` similarly propagates context into its thread pool workers.

---

### Data Flow: Read & Write Path

```
item.path = "/home/user/Music/Artist/track.mp3"   ← absolute on write
      ↓
PathType.to_sql()
      ↓
normalize_path_for_db()  →  b"Artist/track.mp3"   ← stored in SQLite

item.path                                           ← absolute on read
      ↑
PathType.from_sql()
      ↑
expand_path_from_db()    ←  b"Artist/track.mp3"   ← fetched from SQLite
```

---

### Key Invariants

- **Public API unchanged**: `item.path` always returns an absolute `bytes` path.
- **Raw DB value is relative**: direct SQL reads return the relative form (tests assert both).
- **Paths outside the music dir are stored as-is** (e.g. IPFS paths — see `beetsplug/ipfs.py` fix).
- **Migration is idempotent**: rows already relative (no leading `/`) are skipped.

### DIFF
--- a/beets/dbcore/pathutils.py
+++ b/beets/dbcore/pathutils.py
@@ -4,8 +4,10 @@ import os
 from typing import TypeVar
 
 from beets import context, util
+from beets.util import path_as_posix
 
 MaybeBytes = TypeVar("MaybeBytes", bytes, None)
+DB_PATH_SEP = b"/"
 
 
 def _is_same_path_or_child(path: bytes, music_dir: bytes) -> bool:
@@ -15,6 +17,16 @@ def _is_same_path_or_child(path: bytes, music_dir: bytes) -> bool:
     return path_cmp == music_dir_cmp or path_cmp.startswith(
         os.path.join(music_dir_cmp, "")
     )
+
+
+def _to_db_path(path: bytes) -> bytes:
+    """Store relative paths with a platform-neutral separator."""
+    return path_as_posix(path)
+
+
+def _from_db_path(path: bytes) -> bytes:
+    """Convert a stored relative path to the current platform syntax."""
+    return path.replace(DB_PATH_SEP, os.fsencode(os.sep))
 
 
 def normalize_path_for_db(path: MaybeBytes) -> MaybeBytes:
@@ -27,7 +39,7 @@ def normalize_path_for_db(path: MaybeBytes) -> MaybeBytes:
         return path
 
     if _is_same_path_or_child(path, music_dir):
-        return os.path.relpath(path, music_dir)
+        return _to_db_path(os.path.relpath(path, music_dir))
 
     return path
 
@@ -36,6 +48,6 @@ def expand_path_from_db(path: bytes) -> bytes:
     """Convert a stored database path to an absolute library path."""
     music_dir = context.get_music_dir()
     if path and not os.path.isabs(path) and music_dir:
-        return util.normpath(os.path.join(music_dir, path))
+        return util.normpath(os.path.join(music_dir, _from_db_path(path)))
 
     return path
--- a/beets/library/models.py
+++ b/beets/library/models.py
@@ -112,12 +112,11 @@ class LibModel(dbcore.Model["Library"]):
         ):
             # Regex, exact, and string queries operate on the raw DB value, so
             # strip the library prefix to match the stored relative path.
-            if isinstance(pattern, bytes):
-                pattern = normalize_path_for_db(pattern)
-            else:
-                pattern = os.fsdecode(
-                    normalize_path_for_db(util.bytestring_path(pattern))
-                )
+            bytes_pattern = normalize_path_for_db(util.bytestring_path(pattern))
+            if query_cls is not dbcore.query.RegexpQuery:
+                bytes_pattern = util.path_as_posix(bytes_pattern)
+            pattern = os.fsdecode(bytes_pattern)
+
         if field in cls.shared_db_fields:
             # This field exists in both tables, so SQLite will encounter
             # an OperationalError if we try to use it in a query.
--- a/test/library/test_migrations.py
+++ b/test/library/test_migrations.py
@@ -8,6 +8,7 @@ from beets.dbcore import types
 from beets.library import migrations
 from beets.library.models import Album, Item
 from beets.test.helper import TestHelper
+from beets.util import path_as_posix
 
 
 class TestMultiGenreFieldMigration:
@@ -270,6 +271,6 @@ class TestRelativePathMigration:
             .execute("select path from items where id=?", (item.id,))
             .fetchone()[0]
         )
-        assert stored_path == os.fsencode(relative_path)
+        assert stored_path == path_as_posix(os.fsencode(relative_path))
         # and the item.path property still returns an absolute path
         assert item.path == absolute_path
--- a/test/test_library.py
+++ b/test/test_library.py
@@ -34,7 +34,13 @@ from beets.library import Album
 from beets.test import _common
 from beets.test._common import item
 from beets.test.helper import BeetsTestCase, ItemInDBTestCase, capture_log
-from beets.util import as_string, bytestring_path, normpath, syspath
+from beets.util import (
+    as_string,
+    bytestring_path,
+    normpath,
+    path_as_posix,
+    syspath,
+)
 
 # Shortcut to path normalization.
 np = util.normpath
@@ -1143,7 +1149,7 @@ class PathStringTest(BeetsTestCase):
         album = self.lib.add_album([self.i])
 
         assert self.i.path == absolute_path
-        assert stored_path == relative_path
+        assert stored_path == path_as_posix(relative_path)
         assert album.path == os.path.dirname(absolute_path)
 
 


beets/dbcore/types.py
```diff
@@ -389,9 +390,10 @@ def normalize(self, value: Any) -> bytes | N:
...
 
     def from_sql(self, sql_value):
-        return self.normalize(sql_value)
+        return pathutils.expand_path_from_db(self.normalize(sql_value))
 
     def to_sql(self, value: bytes) -> BLOB_TYPE:
+        value = pathutils.normalize_path_for_db(value)
         if isinstance(value, bytes):
             value = BLOB_TYPE(value)
         return value
```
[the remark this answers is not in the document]
snejus: No it won't crash

beets/library/migrations.py
```diff
@@ -181,3 +185,47 @@ def _migrate_data(self, model_cls: type[Model], _: set[str]) -> None:
...
+            with self.db.transaction() as tx:
+                tx.mutate_many(
+                    f"UPDATE {table} SET {field} = ? WHERE id = ?",
+                    [
+                        (os.path.relpath(r[field], self.db.directory), r["id"])
+                        for r in batch
+                    ],
+                )
+
+            migrated += len(batch)
```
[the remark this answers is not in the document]
snejus: No need

test/test_files.py
```diff
@@ -523,11 +523,6 @@ def test_removing_last_item_preserves_library_dir(self):
...
         assert self.lib_path.exists()
 
-    def test_removing_item_outside_of_library_deletes_nothing(self):
-        self.lib.directory = os.path.join(self.temp_dir, b"xxx")
-        self.i.remove(True)
-        assert self.i.filepath.parent.exists()
-
     def test_removing_last_item_in_album_with_albumart_prunes_dir(self):
         artfile = os.path.join(self.temp_dir, b"testart.jpg")
         touch(artfile)
```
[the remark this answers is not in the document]
snejus: It's impossible for an item to be outside lib root anymore.

semohr:
I would love to find some time in the following days to look into this in detail!

Have you considered moving a library between operating systems yet? I could see that as one absolutly killer side effects for relative paths. Think portable external drive.

If you havent thought about it yet, we might want to test that paths are normalized correctly between Windows and Posix.

snejus:
> Have you considered moving a library between operating systems yet? I could see that as one absolutly killer side effects for relative paths.

I hope you mean it's a _positive_ side effect! Well, this new logic consistently uses `os.path` for all operations, so it seems to me this should work fine.

Any ideas how could we possibly test it?

snejus:
I've just [asked users](https://github.com/beetbox/beets/issues/133#issuecomment-4164638696) to test this change - hopefully we have someone who can give it a go using these two filesystems.

semohr:
Yeah was meant in a positive way 😅

No idea how to tests it other than writting an example db in either operating system and trying to load it in the other.

Im pretty sure slashes will not be converted/normalized correctly by using os.path. I guess this is one of the reason Pathlib was born.

This is by no means a blocker tho but maybe a good followup enhancement. Just a thing I was wondering as it might be a personal usecase for me which required manually migrating/rewriting paths in the database beforehand.

snejus:
Someone already responded that they will give it a try because they have access to both file systems.

I do want to get it right here before merging - if that means replacing `os.path` by pathlib then so be it. The only reason why I used `os.path` was that I wanted to avoid the conversion from `bytes` to `str`.

snejus:
Now that I think about it, I think we should actually be able to test it - it's just a matter of patching `os.system` I think.

beets/dbcore/pathutils.py
```diff
@@ -0,0 +1,41 @@
...
+
+import os
+from typing import TypeVar
+
+from beets import context, util
+
+MaybeBytes = TypeVar("MaybeBytes", bytes, None)
+
+
+def _is_same_path_or_child(path: bytes, music_dir: bytes) -> bool:
```
semohr, line 13: Isn't this just a is_relative_to call? I might be missing something but for me it seems to be 
```python
Path(path).is_relative_to(music_dir)
```
I think using pathlib here might be a bit simpler.

beets/dbcore/types.py
```diff
@@ -402,9 +402,10 @@ def normalize(self, value: Any) -> bytes | N:
             return value
 
     def from_sql(self, sql_value):
-        return self.normalize(sql_value)
+        return pathutils.expand_path_from_db(self.normalize(sql_value))
```
semohr, line 405: Wouldn't it make sense to return a pathlib objects here? That way we would be able to remove all the byte handling code and normalization code that is scattered throughout the project (at least I think we have quite some of that).

I get that this comes with some tradeoffs but I think it improves DX significantly. See also https://docs.python.org/3/library/pathlib.html#comparison-to-the-os-and-os-path-modules for comparison if you want to eval this yourself.

beets/library/migrations.py
```diff
@@ -181,3 +186,44 @@ def _migrate_data(self, model_cls: type[Model], _: set[str]) -> None:
...
+
+        total = len(rows)
+        to_migrate = [r for r in rows if r[field] and os.path.isabs(r[field])]
+        if not to_migrate:
+            return
+
+        migrated = total - len(to_migrate)
+        ui.print_(f"Migrating {field} for {total} {table}...")
+        for batch in chunks(to_migrate, self.CHUNK_SIZE):
+            with self.db.transaction() as tx:
```
semohr: I think we should move the transaction outside the chunking loop. Currently, a failure mid-batch leaves a partially applied migration with no rollback of prior batches.

beets/dbcore/types.py
semohr: Maybe not for now but a future enhancement 🤔

beets/context.py
```diff
@@ -0,0 +1,25 @@
+from contextlib import contextmanager
+from contextvars import ContextVar
+
+# Holds the music dir context
+_music_dir_var: ContextVar[bytes] = ContextVar("music_dir", default=b"")
```
semohr, line 5: To be honest I'm not too sure if contextvar is the best approach here. `ContextVar` is meant for async task isolation afaik. Beets uses a thread based pipeline atm. Why not just pass `music_dir` as an explicit parameter to the functions that need it? That would be simpler to understand and would not require the patches in the pipeline layer.

There's a disconnect between `Library` and `music_dir` now which is a bit weird to work with imo. If I create a `Library` at a specific `music_dir`, then in the same process create another Library with another `music_dir`, the context gets overwritten and the first library becomes unusable for path operations.

I would assume `music_dir` to be a part of a `Library` at construction, at least that's how I'd like to work with it. The current global state makes things a bit difficult.

In fact, if `music_dir` was just an instance attribute on `Library` instead of a context var, we wouldn't need `music_dir_context()` at all. Query functions would receive the `library` (or at least `library.directory`) directly and should be able to work with that.

---

This might not be a big issue right now given beets' architecture, but I can see this becoming problematic if the assumption of one main library
per process doesn't hold anymore. Which is well in our reach with the relative paths now.

beets/dbcore/pathutils.py
snejus: That's mainly because the rest of the code uses `os.path` everywhere, and because we need a cheap prefix check here. Using pathlib would also complicate testing which currently patches `os.path` to test different file systems.

If we used pathlib everywhere else `Path.is_relative_to` would be neat solution, but now it would be a bit of an impedance mismatch.

beets/dbcore/types.py
snejus: > Maybe not for now but a future enhancement 🤔

Agreed - we currently use pathlib in fairly limited cases, so I'm wary of introducing it in the middle of core logic. Pathlib only makes sense when there's no need for back and forth conversion to bytes (the type which pathlib doesn't support). We would have to switch to having `Item.path` return a `Path` object for good in order for this to be a viable solution.

beets/library/migrations.py
snejus: That's intentional! Since the data in the database is completely predictable, a failure in reality can only come from someone hitting CTRL-C or their system shutting down, so we __do__ want to continue the migration where we left off.

beets/context.py
snejus: > That would be simpler to understand and would not require the patches in the pipeline layer.

I tried doing it and I found that the call stack is far far too deep for this to be maintainable.

> There's a disconnect between Library and music_dir now which is a bit weird to work with imo. If I create a Library at a specific music_dir, then in the same process create another Library with another music_dir, the context gets overwritten and the first library becomes unusable for path operations.

See `ipfs` plugin which has this exact use case and how this is addressed with `Library.with_music_dir` context manager.

beets/dbcore/pathutils.py
semohr: One could still use [os.path.relpath()](https://docs.python.org/3/library/os.path.html#os.path.relpath) instead of rolling the comparison manually. Makes no difference performance wise but since we use this function only once and afterwards need the realpath anyways, there might be a way to simplify this.


```
rel = os.path.relpath(os.fsdecode(path), os.fsdecode(music_dir))
return rel == "." or not rel.startswith("..")
```

beets/library/migrations.py
semohr: I see. Resumability is useful for long migrations.

That said, it breaks the usual expectation of atomic migrations. Failures may be rare, but not impossible (e.g. fs issues), and a partial migration could leave the DB in a state that older beets versions can’t handle if the user downgrades because of issues.

beets/context.py
semohr: You’ve likely tried this, but I would have assumed that simply prepending `music_dir` to any path coming from the library should work and wouldn’t require changes to the rest of the logic 🤔

It keeps the Library layer self-contained and avoids introducing global state or context machinery.

beets/dbcore/pathutils.py
snejus: Asked Codex about it:
image

beets/library/migrations.py
snejus: Agree. To be fair the chunking was introduced initially when I still tried to migrate using `item.write` when the genre migration took 11 minutes 😅 . Will remove it from here given the importance of this field.

beets/context.py
snejus: It does because we want to continue to be able to query for absolute paths, as in `path::<absolute_path>`, paths reported by `beet ls -p`.

Therefore, since I also wanted to keep responsibilities of `FieldQuery` and `Type` contained, I decided to isolate the context-dependent logic under `pathutils`. This keeps call stacks simple and maintainable, of course with the drawback of the semi-global, context machinery (but this is the use case `contextvar` is designed for to be fair).

snejus:
@semohr the library should now be portable between different file systems.

beets/context.py
semohr: Just to throw this into the room without having thought about it too much: Have we considered a database view that resolve the relative paths?

snejus: A view is only useful when it doesn't require an external input (such as the current library path), right?

semohr: Indeed, we would have to recreate the view whenever the library path changes but it should be a close to zero cost operation as it just creates a query internally :thinking:

snejus: I see what you mean - though this would have to be done on every beet invocation because there's no way to check whether the view is 'in sync'.

I see how it would simplify __reading__ the value from the database, however we would still need the logic for writing the value.

semohr: I guess whenever we open the library should work. Indeed writing would still need the normalization. One can get around this using db triggers but I'm not sure if and how sqlite supports this.

In general, how do you seen the chances, that using a database view would help to get rid of the global state machinery?

snejus: SQLite indeed supports triggers, and I imagine it should work for writing.

This makes me think that hmm maybe a view + a trigger may be a cleaner and more maintainable solution 🤔.

Though, this would make the DB initialisation depend on the library still.

On the other hand, this would not solve the issue with ipfs.

semohr: Just brainstorming here, I have no idea what ipfs does or even is :laughing:  

Maybe there is a fix for ipfs to not require the workaround in the first place? Seems like a weird pattern. Maybe we can align the exception in ipfs to make it work.

semohr: Im also not sure if a db trigger + view is a more maintainable choice tbh. No idea what it would require from a code standpoint to work in beets.

snejus: It essentially creates a separate library. Maybe it would work as well, actually.

I like this idea actually, but I'd probably want to investigate and try it out separately.

Especially now that I've got a confirmation that multiple file systems are supported!

semohr: I have tested the windows setup too and it seems to work. Makes me want to move my beets library to an attachable storage :smile: :tada: 
 
You can decide how to proceed from here. From a code standard perspective, this is a go from me. Feel free to merge if you want, though the global context might bite us down the line.

snejus: I do like the idea you suggested and like I said it sounds cleaner! I'm definitely interested and will work on it soon, but I also want to merge what we have since it finally does the job, it's also one of the oldest issues we have in the repository!

beets/dbcore/pathutils.py
semohr: image

snejus:
@semohr would you mind to approve it if you're happy with it?

PR #6461, opened 2026-03-23
Document extra_debug in the docs

snejus:
### What Changed

Two docs are updated to describe the `BeetsLogger.extra_debug` method, which was previously undocumented for external plugin authors and the changelog note is updated accordingly.


snejus:
> well written and understandable. didnt check format and if links working but we have tools for that

Codex is very good at documentation!

PR #6474, opened 2026-03-30
lastgenre: Refactor `_get_genre`

JOJ0:
## Description

The monolithic `_get_genre` method was broken down into several private instance methods and refactored for readability. The contract is kept and is already well tested (`test_get_genre`)

- **Core Helpers** - were moved from within `_get_genre` to a reusable instance method and a `cached_property`:
    - `_try_resolve_stage`: Handles the canonicalization and logging of genres for a specific stage.
    - `fallback`: Provides the configured fallback genre. Is used as a last resort in `_try_resolve_existing_genres` and when `_get_genre` couldn't find any genre in any stage at all.
 
- **Lookup Stages** - some were complex enough to deserve their own instance method for readability, some stay  inline in `_get_genre`:
    - `_try_resolve_existing_genres`: Manages the initial check for pre-existing genres and the `cleanup_existing` logic when `force` is disabled.
    - track stage: stays inline
    - album stage: indentical to track stage, but not worth moving / deduplication doesn't buy much (see subsequent PR though)
    - `_fetch_artist_stage`: Fetches and resolves artist-level genres, including multi-valued album artists and "Various Artists" logic.
    - `_fetch_va_genres`: specifically handles the plurality logic for "Various Artists" albums.
- **Fallbacks**:
    - `_try_resolve_original_fallback`: Handles the "keep_existing" logic that attempts to use/canonicalize originally present genres if no new ones are found.

Make sure to also look at subsequent PR's:

- https://github.com/beetbox/beets/pull/6890
- https://github.com/beetbox/beets/pull/6893

## To Do

- [x] ~Documentation~
- [x] Changelog. (Not required, refactor only)
- [x] ~Tests~ (_get_genre was already well covered and the signature of the method was kept)

### DIFF
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -39,8 +39,8 @@ if TYPE_CHECKING:
     """Set of valid genre names (lowercase). Empty set means all genres allowed."""
 
     CanonTree = list[list[str]]
-    """Genre hierarchy as list of paths from general to specific.
-    Example: [['electronic', 'house'], ['electronic', 'techno']]"""
+    #: Genre hierarchy as list of paths from general to specific.
+    #: Example: [['electronic', 'house'], ['electronic', 'techno']]
 
     GenresWithLabel = tuple[list[str], str]
     #: A pair of ``(genre list, label)`` returned by a genre resolution stage.
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -42,6 +42,10 @@ if TYPE_CHECKING:
     """Genre hierarchy as list of paths from general to specific.
     Example: [['electronic', 'house'], ['electronic', 'techno']]"""
 
+    GenresWithLabel = tuple[list[str], str]
+    #: A pair of ``(genre list, label)`` returned by a genre resolution stage.
+    #: The label is used for logging and describes the source and filtering applied.
+
 
 # Canonicalization tree processing.
 
@@ -442,7 +446,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         return self._resolve_genres(combined, artist=artist)
 
     @cached_property
-    def fallback(self) -> tuple[list[str], str]:
+    def fallback(self) -> GenresWithLabel:
         """Return the configured fallback genre and label."""
         if fallback := self.config["fallback"].get():
             return [fallback], "fallback"
@@ -454,7 +458,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         keep_genres: list[str],
         new_genres: list[str],
         artist: str | None = None,
-    ) -> tuple[list[str], str] | None:
+    ) -> GenresWithLabel | None:
         """Try to resolve genres for a given stage and log the result.
 
         If any newly fetched genres and/or existing genres are resolved, return
@@ -474,7 +478,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
 
     def _try_resolve_existing_genres(
         self, obj: LibModel, genres: list[str]
-    ) -> tuple[list[str], str] | None:
+    ) -> GenresWithLabel | None:
         """Handle existing genres when not forcing.
 
         Clean up existing genres if enabled, or return them unchanged. Return
@@ -487,11 +491,11 @@ class LastGenrePlugin(plugins.BeetsPlugin):
                 "cleanup", keep_genres, [], artist=self._artist_for_filter(obj)
             )
 
-        return genres, "keep any, no-force"
+        return genres, "keep any, no-force"  # type: ignore
 
     def _try_resolve_original_fallback(
         self, obj: LibModel, genres: list[str], keep_genres: list[str]
-    ) -> tuple[list[str], str] | None:
+    ) -> GenresWithLabel | None:
         """Attempt to fall back to existing original genres if configured.
 
         ``genres`` are the original unchanged values and are checked as-is
@@ -572,7 +576,7 @@ class LastGenrePlugin(plugins.BeetsPlugin):
 
         return "most popular track", [], None
 
-    def _get_genre(self, obj: LibModel) -> tuple[list[str], str]:
+    def _get_genre(self, obj: LibModel) -> GenresWithLabel:
         """Get the final genre list for an Album or Item object.
 
         `self.sources` specifies allowed genre sources. Starting with the first
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -529,17 +529,17 @@ class LastGenrePlugin(plugins.BeetsPlugin):
         # album artist, or most popular track genre.
         if isinstance(obj, library.Item) and "track" in self.sources:
             if new_genres := self.client.fetch("track", obj):
-                if result := self._try_resolve_stage(
+                if resolved := self._try_resolve_stage(
                     "track", keep_genres, new_genres, artist=obj.artist
                 ):
-                    return result
+                    return resolved
 
         if "album" in self.sources:
             if new_genres := self.client.fetch("album", obj):
-                if result := self._try_resolve_stage(
+                if resolved := self._try_resolve_stage(
                     "album", keep_genres, new_genres, artist=obj.albumartist
                 ):
-                    return result
+                    return resolved
 
         if "artist" in self.sources:
             new_genres = []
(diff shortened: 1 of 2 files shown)


beetsplug/lastgenre/__init__.py
```diff
@@ -441,6 +441,141 @@ def _combine_resolve_and_log(
         combined = old + new
         return self._resolve_genres(combined, artist=artist)
 
+    def _configured_fallback(self) -> tuple[list[str], str]:
+        """Return the configured fallback genre and label."""
+        if fallback := self.config["fallback"].get():
+            return [fallback], "fallback"
+        return [], "fallback unconfigured"
```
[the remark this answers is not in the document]
JOJ0: fixed

beetsplug/lastgenre/__init__.py
```diff
@@ -441,6 +441,141 @@ def _combine_resolve_and_log(
...
+                rank,
+            )
+            return [most_popular]
+
+        return []
+
+    def _fetch_artist_stage(
+        self, obj: LibModel
+    ) -> tuple[str, list[str], str | None]:
+        """Fetch artist genres for an Item or Album object.
```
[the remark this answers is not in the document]
JOJ0: fixed

beetsplug/lastgenre/__init__.py
```diff
@@ -461,146 +596,50 @@ def _get_genre(self, obj: LibModel) -> tuple[list[str], str]:
...
-            if self.config["cleanup_existing"]:
-                keep_genres = [g.lower() for g in genres]
-                if result := _try_resolve_stage(
-                    "cleanup",
-                    keep_genres,
-                    [],
-                    artist=self._artist_for_filter(obj),
-                ):
-                    return result
+        existing_genres = self._get_existing_genres(obj)
```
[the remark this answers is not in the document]
JOJ0: I actually like it that way. It is a thing we use throughout in the _get_genre. It should stay, even though the fallback for the conditional assignment is there right below as copilot mentions. That's ok...

JOJ0:
@semohr @snejus @henry-oberholtzer @Serene-Arc anyone care to review a no-op refactor-only PR? :-) Thanks :-)

beetsplug/lastgenre/__init__.py
```diff
@@ -441,6 +441,141 @@ def _combine_resolve_and_log(
         combined = old + new
         return self._resolve_genres(combined, artist=artist)
 
+    def _configured_fallback(self) -> tuple[list[str], str]:
```
snejus: I think we want to define an alias:

```py
GenresWithLabel = tuple[list[str], str]
```

snejus: Define this as a property:

```py
@cached_property
def fallback(...):
```

No need to include `configured_` as this is redundant.

beetsplug/lastgenre/__init__.py
```diff
@@ -441,6 +441,141 @@ def _combine_resolve_and_log(
...
+    def _try_resolve_existing_genres(
+        self, obj: LibModel, genres: list[str]
+    ) -> tuple[list[str], str] | None:
+        """Handle existing genres when not forcing.
+
+        Return ``None`` if forcing or no genres exist. Otherwise, clean up existing
+        genres if enabled, falling back if needed, or return them unchanged.
+        """
+        if not genres or self.config["force"]:
+            return None
```
snejus: Move this check to `_get_genre` since `"force"` is only handled in that function, something like:

```py
        if existing_genres and not self.config["force"]:
            return self._try_resolve_existing_genres(obj, existing_genres)
```

This will simplify `_try_resolve_existing_genres` where `None` may never be returned.

beetsplug/lastgenre/__init__.py
```diff
@@ -441,6 +441,141 @@ def _combine_resolve_and_log(
...
+
+        if self.config["cleanup_existing"]:
+            keep_genres = [g.lower() for g in genres]
+            if result := self._try_resolve_stage(
+                "cleanup", keep_genres, [], artist=self._artist_for_filter(obj)
+            ):
+                return result
+
+            return self._configured_fallback()
```
snejus: Hmm, this line means that this function should rather be called `_resolve_existing_genres` as it _always_ returns a successful output.

Handling fallback values seems out of scope of this function. Maybe this should rather return `None` here, so that fallback handling is done within `_get_genre`? What do you reckon?

beetsplug/lastgenre/__init__.py
```diff
@@ -441,6 +441,141 @@ def _combine_resolve_and_log(
...
+                new_genres += self.client.fetch(
+                    "album_artist", obj, albumartist
+                )
+            if new_genres:
+                # Already filtered per-artist in client
+                return "multi-valued album artist", new_genres, None
+            return "album artist", [], None
+
+        # For "Various Artists", pick the most popular track genre.
+        assert isinstance(obj, Album)  # Type narrowing for mypy
```
snejus, line 573: I don't think the rest of this function is tested (neither is `_fetch_va_genres`). Would you mind adding a test for this?

beetsplug/lastgenre/__init__.py
```diff
@@ -461,146 +596,50 @@ def _get_genre(self, obj: LibModel) -> tuple[list[str], str]:
...
-            )
-            if resolved_genres:
-                suffix = "whitelist" if self.whitelist else "any"
-                label = f"{stage_label}, {suffix}"
-                if keep_genres:
-                    label = f"keep + {label}"
-                return self._format_genres(resolved_genres), label
-            return None
-
         keep_genres = []
```
snejus: This line is redundant now.
```suggestion
```

snejus (changes requested):

beetsplug/lastgenre/__init__.py
JOJ0: I know, see: https://github.com/beetbox/beets/pull/6474/changes#r3681748469 :-) I wanted to make it superobvious that this _is a thing_ we are after in `get_genre()`. Readability idea. Very bad idea/practice?

snejus: It's unused, so I'd say it's rather confusing... 😅

beetsplug/lastgenre/__init__.py
JOJ0: Interesting idea. Keeping the name _try_resolve_existing_genres but moving out the fallback handling back to the main function, see here, would look like this: 6a14dd95e

What I like about it is that now to understand when a fallback is fired we dont have to switch context. it's all happening in top-level _get_genre. 👍

beetsplug/lastgenre/__init__.py
JOJ0: https://github.com/beetbox/beets/pull/6474/changes#r3791188145

beetsplug/lastgenre/__init__.py
snejus: Looks good!

beetsplug/lastgenre/__init__.py
JOJ0: Good call! Added a test and realized some shortcomings of the ever exisiting "most popular track genre" mechanics. I'd like to address that in a forthcoming PR. Here's a (generated) summary

- Uneven vote weight: each returned tag counts as a vote, so tracks with more tags outweigh tracks with fewer.
- Silent tie-breaking: Counter.most_common(1) resolves ties by insertion order, not any meaningful rule.
- Ties discard info: only one genre is returned even when several tie for the top spot.
- No debug visibility: no log signal exists to explain why a genre won on a close or tied vote.

JOJ0:
Thanks @snejus that was great feedback! 👍  I think I addressed all of it.

beetsplug/lastgenre/__init__.py
snejus: Just don't forget to define this alias!

JOJ0: Dang. Just wanted to tell you: I forgot this. Had it, removed it for rebasing's sake. In a jiff...

JOJ0: Done. Now added.

beetsplug/lastgenre/__init__.py
```diff
@@ -42,6 +42,10 @@
     """Genre hierarchy as list of paths from general to specific.
     Example: [['electronic', 'house'], ['electronic', 'techno']]"""
 
+    GenresWithLabel = tuple[list[str], str]
+    """A pair of (genre list, label) returned by a genre resolution stage.
+    The label is used for logging and describes the source and filtering applied."""
+
```
snejus, line 48: A small tip (see https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#doc-comments-and-docstrings):


```suggestion
    GenresWithLabel = tuple[list[str], str]
    #: A pair of ``(genre list, label)`` returned by a genre resolution stage.
    #: The label is used for logging and describes the source and filtering applied.

```

snejus (approved):

JOJ0:
Hi @snejus, thanks for the approval. Did two more tiny things: Fixed mypy errors, accidentally applied `GenresWithLable` alias to `_fetch_artist_stage` which is slightly different. Fixed another 2-line docstring/comment (below type alias `CanonTree`).

PR #6480, opened 2026-03-31
autobpm: Allow hiding '"BPM already exists" log with `--quiet`

JOJ0:
## Description

When batch bpm computing, the plugin's output is very noisy. This change allows hiding "BPM already exists for item" logs with a new `--quiet` option (or config setting)

- [x] Documentation. (Added a note to the plugin docs Configuration section)
- [x] Changelog.
- [x] ~Tests.~

PR #6481, opened 2026-03-31
autobpm: Add `--force` and `--quiet` options; Deprecate `overwrite`

JOJ0:
## Description

- Reveal the `overwrite` config setting to the CLI and at the same time deprecate the option name. It should be named `force` to streamline with other beets configs of the same kind.

- Add a `quiet` CLI flag and config option that leaves out output pointing to _already existing BPM values_ 

- [x] Documentation. 
- [x] Changelog.document.)
- [x] Tests.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -18,6 +18,10 @@ New features
 - Query: Add ``has_cover_art`` computed field to query items by embedded cover
   art presence. Users can now search for tracks with or without embedded artwork
   using ``beet list has_cover_art:true`` or ``beet list has_cover_art:false``.
+- :doc:`plugins/autobpm`: Add ``force`` configuration and CLI option and
+  deprecate ``overwrite``.
+- :doc:`plugins/autobpm`: The "BPM already exists for item" log message can now
+  be hidden with the ``--quiet`` flag.
 
 Bug fixes
 ~~~~~~~~~


beetsplug/autobpm.py
```diff
@@ -34,35 +35,77 @@ def __init__(self) -> None:
...
+        )
         cmd.func = self.command
         return [cmd]
 
-    def command(self, lib: Library, _, args: list[str]) -> None:
-        self.calculate_bpm(list(lib.items(args)), write=should_write())
+    def command(self, lib: Library, opts, args: list[str]) -> None:
+        force = (
+            opts.force
+            if hasattr(opts, "force")
```
jackwilsdon: `optparse` stores the default for an option even when not passed, so `opts.force` is always present (and defaults to `False`).

beetsplug/autobpm.py
```diff
@@ -34,35 +35,77 @@ def __init__(self) -> None:
...
+        quiet = self.config["quiet"].get(bool) or opts.quiet
+        self.calculate_bpm(
+            list(lib.items(args)),
+            write=should_write(),
+            force=force,
+            quiet=quiet,
+        )
 
     def imported(self, _, task: ImportTask) -> None:
         self.calculate_bpm(task.imported_items())
```
jackwilsdon: Should this also obey `force` and `quiet` from the config?

docs/plugins/autobpm.rst
```diff
@@ -20,20 +20,44 @@ Configuration
...
+    :default: no
+
+    Calculate a BPM even for files that already have a ``bpm`` value.
+
+.. conf:: overwrite
+    :default: no
+
+    .. deprecated:: 2.9 Use ``force`` instead.
+
+.. conf:: beat_track_kwargs
```
jackwilsdon, line 52: Can we document `quiet` too?

beetsplug/autobpm.py
JOJ0: on it already. thanks!

docs/plugins/autobpm.rst
JOJ0: quiet is in a different PR

jackwilsdon: #6480 (the quiet PR) got merged into this one 😖

beetsplug/autobpm.py
JOJ0: good point! will fix!

docs/plugins/autobpm.rst
JOJ0: :-( yes, fixed it so far. will address missing docs and "obey configs on import" tomorrow. Thanks for the feedback!

JOJ0: done

JOJ0:
Docs render and relevant parts look like this (first time I actually use the conf directives and the deprecations features. Very nice stuff! Love it @snejus) Hope I did it correctly since it doesnt state which version we will throw it out? (The logs say that this is with 3.0.0, so i guess when we deprecate something we will always until next major release which is allowed to include  breaking changes, right?)

Screenshot 2026-04-01 at 09 07 41


@jackwilsdon appreciate a final review :-)

docs/plugins/autobpm.rst
```diff
@@ -20,20 +20,50 @@ Configuration
...
+.. conf:: force
+    :default: no
+
+    Calculate a BPM even for files that already have a ``bpm`` value.
+
+.. conf:: overwrite
+    :default: no
+
+    .. deprecated:: 2.9 Use ``force`` instead.
+
```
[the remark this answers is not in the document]
JOJ0: I dont think this is necessary, nor helpful. User can read description for force which is right next ot it

beetsplug/autobpm.py
```diff
@@ -34,35 +35,77 @@ def __init__(self) -> None:
...
             }
         )
 
+        if self.config["overwrite"].exists():
+            deprecate_for_user(
+                self._log,
+                f"'{self.name}.overwrite' configuration option",
+                f"'{self.name}.force'",
+            )
+            self.config["force"] = self.config["overwrite"].get()
```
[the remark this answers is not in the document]
JOJ0: Way too much hassle. The user is warned with this already:

_autobpm: 'autobpm.overwrite' configuration option is deprecated and will be removed in version 3.0.0. Use 'autobpm.force' instead._

Worst thing can happen that force is not working and user is _forced_ to finally fix there config. Good this way!

docs/plugins/autobpm.rst
```diff
@@ -20,20 +20,50 @@ Configuration
...
 .. code-block:: yaml
 
     autobpm:
-      beat_track_kwargs:
-        start_bpm: 160
+        auto: yes
+        force: no
+        beat_track_kwargs: {}
+        quiet: no
+
```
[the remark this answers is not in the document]
JOJ0: Ok fair point. I added that info to the config section and refrained from an extra usage section. Think that's straight forward enough. Eg:
Screenshot 2026-04-01 at 09 42 51

snejus:
> Docs render and relevant parts look like this (first time I actually use the conf directives and the deprecations features. Very nice stuff! Love it @snejus) Hope I did it correctly since it doesnt state which version we will throw it out? (The logs say that this is with 3.0.0, so i guess when we deprecate something we will always until next major release which is allowed to include breaking changes, right?)
> Screenshot 2026-04-01 at 09 07 41
> 
> @jackwilsdon appreciate a final review :-)

Sounds good to me as well!

beetsplug/autobpm.py
```diff
@@ -34,35 +35,77 @@ def __init__(self) -> None:
...
+    ) -> None:
         for item in items:
             path = item.filepath
             if bpm := item.bpm:
-                self._log.info("BPM for {} already exists: {}", path, bpm)
-                if not self.config["overwrite"]:
+                if not quiet:
+                    self._log.info("BPM for {} already exists: {}", path, bpm)
+                if not force:
                     continue
```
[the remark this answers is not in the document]
JOJ0: I addressed that now too. Tests for all 3 things but left out extra-in-config-only tests for quiet/force. IMO testing `--force` and `--quiet` is enough and tests the code paths already.

JOJ0:
I squashed together quite some commits making this a combined feature package containing force/quiet/deprecate-overwrite. The single commits mixed stuff anyway and won't be very helpful for future git log viewers. It's cleaner now. I also added relevant tests in the end to be complete. Making the call now and merge, thanks for both of your feedback! Much appreciated!

PR #6484, opened 2026-03-31
listenbrainz: Add pagination, play count aggregation, and recording_mbid fix

### LINKED ISSUE #6469: listenbrainz: the plugin appears to be broken
### Problem

Running this command:

```sh
$ beet lbimport
```

Led to this problem:

```
  File "/home/konstantink/.local/bin/beet", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/konstantink/.local/share/uv/tools/beets/lib/python3.13/site-packages/beets/ui/__init__.py", line 1013, in main
    _raw_main(args)
    ~~~~~~~~~^^^^^^
  File "/home/konstantink/.local/share/uv/tools/beets/lib/python3.13/site-packages/beets/ui/__init__.py", line 992, in _raw_main
  File "/home/konstantink/.local/share/uv/tools/beets/lib/python3.13/site-packages/beetsplug/listenbrainz.py", line 34, in func
    self._lbupdate(lib, self._log)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/home/konstantink/.local/share/uv/tools/beets/lib/python3.13/site-packages/beetsplug/listenbrainz.py", line 47, in _lbupdate
    found, unknown = process_tracks(lib, tracks, log)
                     ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/home/konstantink/.local/share/uv/tools/beets/lib/python3.13/site-packages/beetsplug/lastimport.py", line 269, in process_tracks
    new_count = int(tracks[num]["playcount"])
                    ~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: 'playcount'
```

This specific crash seems to have been caused by 6eff9712574d6a736f4840041bfb72d557008c41. Using .get silences the error, but listenbrainz probably shouldn't call a function from lastimport that's now hardcoded to write to lastfm_play_count. Also, when I run lbimport, it only seems to import 25 listens, even though I have way more than that. 

### Setup

* OS: Fedora Linux 43 (Workstation Edition)
* Python version: 3.13.12
* beets version: 2.8.0

arsaboo:
Follow-up to #6471 — fixes three remaining issues with the `listenbrainz` plugin:
                                                                                                                                                                                                                                                                                                                                                             
  - **Aggregate listen events into actual play counts.** ListenBrainz returns individual listen events, each mapped to `playcount: 1`. Without aggregation, the final `listenbrainz_play_count` is always 1 regardless of actual listens.
  - **Paginate through all listens.** The API defaults to 25 results per request. Now fetches up to 1000 per page and loops via `max_ts` until all listens are retrieved.
  - **Use `recording_mbid` from `mbid_mapping` when present.** Previously `mbid` was left as `None` when the mapping existed, falling back to expensive MB API lookups unnecessarily.

  Fixes #6469 (remaining issues after #6471)

### DIFF
--- a/beetsplug/listenbrainz.py
+++ b/beetsplug/listenbrainz.py
@@ -42,16 +42,23 @@ class ListenBrainzPlugin(MusicBrainzAPIMixin, BeetsPlugin):
         lbupdate_cmd = ui.Subcommand(
             "lbimport", help="Import ListenBrainz history"
         )
+        lbupdate_cmd.parser.add_option(
+            "--max",
+            dest="max_listens",
+            type="int",
+            default=None,
+            help="maximum number of listens to fetch (default: all)",
+        )
 
         def func(lib, opts, args):
-            self._lbupdate(lib, self._log)
+            self._lbupdate(lib, self._log, max_listens=opts.max_listens)
 
         lbupdate_cmd.func = func
         return [lbupdate_cmd]
 
-    def _lbupdate(self, lib, log):
+    def _lbupdate(self, lib, log, max_listens=None):
         """Obtain play counts from ListenBrainz."""
-        listens = self.get_listens()
+        listens = self.get_listens(max_total=max_listens)
         if listens is None:
             log.error("Failed to fetch listens from ListenBrainz.")
             return
@@ -120,8 +127,8 @@ class ListenBrainzPlugin(MusicBrainzAPIMixin, BeetsPlugin):
             self._log.debug("Invalid Search Error: {}", e)
             return None
 
-    def get_listens(self, min_ts=None, max_ts=None, count=None):
-        """Gets the full listen history of a given user.
+    def get_listens(self, min_ts=None, max_ts=None, count=None, max_total=None):
+        """Gets the listen history of a given user.
 
         Paginates through all available listens using the max_ts parameter.
 
@@ -131,6 +138,7 @@ class ListenBrainzPlugin(MusicBrainzAPIMixin, BeetsPlugin):
             max_ts: History after this timestamp will not be returned.
                     DO NOT USE WITH min_ts.
             count: How many listens to return per page (max 1000).
+            max_total: Stop after fetching this many listens in total.
 
         Returns:
             A list of listen info dictionaries, or None on API failure.
@@ -143,7 +151,15 @@ class ListenBrainzPlugin(MusicBrainzAPIMixin, BeetsPlugin):
         all_listens = []
 
         while True:
-            params = {"count": per_page}
+            if max_total is not None:
+                remaining_needed = max_total - len(all_listens)
+                if remaining_needed <= 0:
+                    break
+                page_size = min(per_page, remaining_needed)
+            else:
+                page_size = per_page
+
+            params = {"count": page_size}
             if max_ts is not None:
                 params["max_ts"] = max_ts
             if min_ts is not None:
@@ -163,7 +179,7 @@ class ListenBrainzPlugin(MusicBrainzAPIMixin, BeetsPlugin):
             self._log.info("Fetched {} listens so far...", len(all_listens))
 
             # If we got fewer than requested, we've reached the end
-            if len(listens) < per_page:
+            if len(listens) < page_size:
                 break
 
             # Paginate using the oldest listen's timestamp.
@@ -183,9 +199,6 @@ class ListenBrainzPlugin(MusicBrainzAPIMixin, BeetsPlugin):
                 continue
             mbid_mapping = track["track_metadata"].get("mbid_mapping", {})
             mbid = mbid_mapping.get("recording_mbid")
-            if mbid is None:
-                # search for the track using title and release
-                mbid = self.get_mb_recording_id(track)
             tracks.append(
                 {
                     "album": (
--- a/beetsplug/listenbrainz.py
+++ b/beetsplug/listenbrainz.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import datetime
+from collections import Counter
 from typing import TYPE_CHECKING, ClassVar
 
 import requests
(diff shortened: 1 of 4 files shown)


beetsplug/listenbrainz.py
```diff
@@ -80,41 +104,50 @@ def _make_request(self, url, params=None):
...
+            if not listens:
+                break
+
+            all_listens.extend(listens)
+            self._log.info("Fetched {} listens so far...", len(all_listens))
+
+            # If we got fewer than requested, we've reached the end
+            if len(listens) < per_page:
+                break
+
```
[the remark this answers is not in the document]
snejus: memory boom 🤣

snejus:
@arsaboo could you resolve the Copilot comments that you've addressed?

beetsplug/listenbrainz.py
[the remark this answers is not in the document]
arsaboo: This is probably fine to skip for typical user sizes

arsaboo:
@snejus Resolved the comments and updated the branch.

beetsplug/listenbrainz.py
```diff
@@ -80,41 +108,59 @@ def _make_request(self, url, params=None):
...
+        all_listens = []
+
+        while True:
+            params = {"count": per_page}
+            if max_ts is not None:
+                params["max_ts"] = max_ts
+            if min_ts is not None:
+                params["min_ts"] = min_ts
+
+            response = self._make_request(url, params)
```
snejus, line 168: I feel like ListenBrainz has fairly strict rate limits, just like MusicBrainz. Would you mind checking this?

Have you tested fetching many tracks?

arsaboo: Good catch — ListenBrainz does enforce rate limits and returns X-RateLimit-Remaining and X-RateLimit-Reset-In headers on every response. I have updated _make_request to sleep for reset_in + 1 seconds when the remaining quota hits zero, so bulk fetches will automatically pause at the window boundary rather than getting 429s. 

1000 is the officially documented MAX_ITEMS_PER_GET constant from the ListenBrainz API (DEFAULT_ITEMS_PER_GET = 25, MAX_ITEMS_PER_GET = 1000), so we're fetching the maximum allowed per page. I haven't tested against a real account with a large history, but the pagination logic is covered by test_get_listens_paginates, which verifies that multi-page fetching works correctly.

snejus: Just tried testing it - it seems to fetch properly. However, I got tens of thousands of listens in my account. I would expect to be able to provide a max number of listens to fetch, say `beet lbimport --max=5000`.

snejus: image

Hmm, it seems to have been stuck here for an hour now 🤔

arsaboo: Found the culprit: `get_tracks_from_listens` was calling the MusicBrainz API for every listen that didn't have a `recording_mbid` in the mapping. With a large history that's potentially tens of thousands of MB API calls, hence getting stuck. Removed those lookups since `update_play_counts` already does artist+title+album fallback matching. 

Also added `beet lbimport --max=N` as requested.

snejus: 2 hours passed, I'm killing it - seems to have gotten stuck somewhere.

arsaboo: Can you try with the latest update...thanks!

snejus: OK tried again: looking better. Though, what's taking so long to process?

image

arsaboo: The fetch is complete, so the slow part now is `update_play_counts`. It runs one SQLite query per unique track in the listen history. After aggregation, if there are, say, 15,000 unique tracks, that's 15,000 individual beets DB queries, each using SubstringQuery (a LIKE match on title, artist, and album). That's inherently slow for large histories.                           
                                                                                                                                                                                                                                                                                                                                                                                      This isn't something introduced in this PR; it's in the shared utility from #6471. A proper fix would be to load the library once and do in-memory matching, but that's a bigger refactor outside the scope of this PR.      

I added a progress logic for now.

snejus: Nice, this sounds great! Just tested, it does the job

snejus: Just waiting for the command to run across my entire listenbrainz history and will report how it goes

snejus: Oh boy it's taking forever. It seems to me that every further 250 items loop is becoming slower and slower. Could it be true?

snejus: Took a while but it managed 
image

docs/changelog.rst
```diff
@@ -46,6 +46,10 @@ Bug fixes
 - :doc:`plugins/deezer`: Fix Various Artists albums being tagged with a
   localized string instead of the configured ``va_name``. Detection now uses
   Deezer's artist ID rather than the artist name string. :bug:`4956`
+- :doc:`plugins/listenbrainz`: Paginate through all ListenBrainz listens instead
+  of fetching only 25, aggregate individual listen events into correct play
+  counts, and use ``recording_mbid`` from the API response when available.
```
snejus: Are you still using `recording_mbid`?

beetsplug/listenbrainz.py
arsaboo: Yes, each song.store() was its own SQLite transaction (a separate BEGIN/COMMIT/fsync). With thousands of tracks, the write-ahead log accumulates, and each successive commit gets slower. Wrapped the entire processing loop in a single lib.transaction() so all writes commit atomically at the end. Should be dramatically faster now.

docs/changelog.rst
arsaboo: `recording_mbid` is still used. When ListenBrainz includes it in the mbid_mapping of a listen, we pass it through as the track's mbid (used by update_play_counts to match against mb_trackid in your library). What we removed was the fallback that hit the MusicBrainz search API for every listen that didn't have a mapping.

snejus (approved):
thanks! <3

Maxr1998:
@arsaboo what was the idea behind the `--max` flag? If I understand it correctly, limiting the number of fetched total listens means that the resulting number of listens per track may be incomplete. I've been looking at this because I realized that the plugin doesn't reset the count for local listens that aren't in the dump, leaving stale numbers, but the `max` option can actually have the same impact.

I'm currently contemplating removing that option again and also resetting the listen count (which is scoped to ListenBrainz now) during an import.

PR #6491, opened 2026-04-02
Fix albumartists_sort prepending joined artist credit (#6470)

### LINKED ISSUE #6470: `albumartists_sort` contains full artist credit in first element
Since beets 2.8.0, the values for `albumartist_sort` (and related ones like `artists_sort`) changed. The first value in the list is now the complete artist credit, and only after that are the single artist credits. Example list:

1. Zimmer, Hans and Powell, John
2. Zimmer, Hans
3. Powell, John

Additionally, for tracks that only have one artist, the list only contains a single entry. Thus the workaround of using `albumartists_sort[1]` is also not applicable.

### Problem

```sh
$ beet -vv move -p "Hans Zimmer"
```

Led to this problem:

```
/home/xxx/music/Zimmer, Hans [e6de1f3b]/Kung Fu Panda_ Music From the Motion Picture/01-11 Shifu Faces Tai Lung.mp3
  -> /home/xxx/music/Zimmer, Hans and Powell, John [e6de1f3b]/Kung Fu Panda_ Music From the Motion Picture/01-11 Shifu Faces Tai Lung.mp3
```
I expect the files to not be moved

### Setup

* OS: Arch Linux
* Python version: 3.14.3
* beets version: 2.8.0

```yaml
directory: /home/xxx/music/

item_fields:
    first_album_artist: |
        return albumartists_sort[0]

paths:
  default: >-
    $first_album_artist [%left{$mb_albumartistid, 8}]/
    $album%aunique{}/$disc-$track $title
```

jazib-ahmad:
## Description

Fixes #6470

Since beets 2.8.0, albumartists_sort incorrectly includes the full combined artist credit as the first element of the list. For example, for a release with artists "Hans Zimmer" and "John Powell", the field returns ["Zimmer, Hans and Powell, John", "Zimmer, Hans", "Powell, John"] instead of ["Zimmer, Hans", "Powell, John"].

The issue was  in ensure_first_value in hooks.py. The set intersection returned an empty set  for comma separated sort names, so the early return never triggered and the joined name got prepended. For the fix  a substring check was added to correctly detect when the single value is a joined version of the list values.

- [ ] Documentation
- [x] Changelog
- [x] Tests


beets/autotag/hooks.py
```diff
@@ -80,8 +80,9 @@ def ensure_first_value(single_field: str, list_field: str) -> None:
             data.get(single_field) or "",
             data.get(list_field) or [],
         )
-        if single_val not in list_val and set(single_val.lower().split()) & set(
-            map(str.lower, list_val)
+        if single_val not in list_val and (
+            set(single_val.lower().split()) & set(map(str.lower, list_val))
+            or any(v.lower() in single_val.lower() for v in list_val)
```
snejus: We probably want to define variables for `single_val.lower()` and `set(map(str.lower, list_val))` to avoid repeated calculations.

beets/autotag/hooks.py
```diff
@@ -80,8 +80,9 @@ def ensure_first_value(single_field: str, list_field: str) -> None:
             data.get(single_field) or "",
             data.get(list_field) or [],
         )
-        if single_val not in list_val and set(single_val.lower().split()) & set(
-            map(str.lower, list_val)
+        if single_val not in list_val and (
+            set(single_val.lower().split()) & set(map(str.lower, list_val))
```
snejus: Would you mind adding comments above each of these two lines to explain what they do?

docs/changelog.rst
```diff
@@ -26,10 +26,9 @@ Bug fixes
   Play-count imports now use source-specific fields so
   :doc:`plugins/listenbrainz`, :doc:`plugins/lastimport`, and
   :doc:`plugins/mpdstats` do not clash. :bug:`6469`
-
-..
-    For plugin developers
```
snejus, line 38: I don't think you intended to remove this commented out section.

test/autotag/test_hooks.py
```diff
@@ -339,6 +339,11 @@ def test_singleton(self, item, track_info, expected_item_artist):
         ("1 FT 2", ["1", "1 ft 2"], ("1 FT 2", ["1", "1 ft 2"])),
         ("a", ["b", "A"], ("a", ["b", "A"])),
         ("1 ft 2", ["2", "1"], ("1 ft 2", ["2", "1"])),
+        (
+            "Zimmer, Hans and Powell, John",
+            ["Zimmer, Hans", "Powell, John"],
+            ("Zimmer, Hans and Powell, John", ["Zimmer, Hans", "Powell, John"]),
```
snejus: Use simplified values `a, b, c` etc. for consistency with the test cases above.

beets/autotag/hooks.py
jazib-ahmad: Thanks for the feedback. I will fix all these very soon

jazib-ahmad:
@snejus  I changed the test case to simplified values, made changes to ensure_first_value by adding comments ,simplifying logic and creating a variable. Also restored changelog.rst.

jazib-ahmad:
@snejus Also created a variable for set(map(str.lower, list_val)). Missed that in the previous commit

beets/autotag/hooks.py
```diff
@@ -80,8 +80,16 @@ def ensure_first_value(single_field: str, list_field: str) -> None:
...
-            map(str.lower, list_val)
+        single_val_lower = single_val.lower()
+        list_val_lower = set(map(str.lower, list_val))
+        if single_val not in list_val and (
+            # Joined credits share words with individual list values
+            set(single_val_lower.split()) & list_val_lower
+            or (
+                # Single value is a joined credit containing all list items
+                len(list_val) > 1
+                and all(v.lower() in single_val_lower for v in list_val)
```
snejus:
```suggestion
                and all(v in single_val_lower for v in list_val)
```

beets/autotag/hooks.py
```diff
@@ -80,8 +80,16 @@ def ensure_first_value(single_field: str, list_field: str) -> None:
...
         )
-        if single_val not in list_val and set(single_val.lower().split()) & set(
-            map(str.lower, list_val)
+        single_val_lower = single_val.lower()
+        list_val_lower = set(map(str.lower, list_val))
+        if single_val not in list_val and (
+            # Joined credits share words with individual list values
+            set(single_val_lower.split()) & list_val_lower
+            or (
+                # Single value is a joined credit containing all list items
```
snejus:
```suggestion
                # Each of the credits in the list are in the joined credit
```

snejus (changes requested):
A couple of more small comments. Additionally, I think it would be a good idea to adjust the test (and expected) data in `ApplyTest.setUp` to test that this case is being handled properly. See the following lines:

```py
                    artist_sort="trackArtistSort",
                    artists_sort=["trackArtistSort"],
```

jazib-ahmad:
@snejus I've addressed all the feedback.  updated the comment, fixed the lowercase comparison bug in the join detection, updated ApplyTest with multiple artist sort data.

PR #6498, opened 2026-04-04
fix(replaygain): clear conflicting tags on write

ShimmerGlass:
## Description

There are two ways to store replay gain data on music files: RG_ and R128_ (opus). 
Before this change the beets replaygain plugin simply set the right tag based on file format and r128 config value. there are case, however, when for example an opus files comes with RG_ tags already set. After beet write its replaingain tags the files will contain both RG_ and R128_ tags with possibly conflicting values. 

For example, Navidrome currently always prefers RG_ tags over R128_ regardless of format, leading tags set by beets to be ignored. https://github.com/navidrome/navidrome/blob/23f3556371321faf199866989b906f2ef06a8034/model/metadata/map_mediafile.go#L111

As per RFC 7845: Ogg Encapsulation for the Opus Audio Codec 
https://datatracker.ietf.org/doc/html/rfc7845#section-5.2.1
> To avoid confusion with multiple normalization schemes, an Opus
> comment header SHOULD NOT contain any of the REPLAYGAIN_TRACK_GAIN,
> REPLAYGAIN_TRACK_PEAK, REPLAYGAIN_ALBUM_GAIN, or
> REPLAYGAIN_ALBUM_PEAK tags, unless they are only to be used in some
> context where there is guaranteed to be no such confusion.

Since the replaygain plugin does not support setting both RG_ and R128_ at the same time it doesn't make much sense to keep conflicting tags when writing. These tags are also easy to recalculate if needed either with beets itself (with a different configuration), or by running replaygain commands by hand. no valuable information is lost.

This change makes it so the replagain plugin now deletes conflicting tags when processing library items:
* RG_ are deleted if we set R128_
* R128_ are deleted if we set RG_

## To Do

### DIFF
--- a/beetsplug/replaygain.py
+++ b/beetsplug/replaygain.py
@@ -139,6 +139,7 @@ class RgTask:
 
     def _store_track_gain(self, item: Item, track_gain: Gain):
         """Store track gain for a single item in the database."""
+        item.r128_track_gain = None
         item.rg_track_gain = track_gain.gain
         item.rg_track_peak = track_gain.peak
         item.store()
@@ -152,6 +153,7 @@ class RgTask:
 
         The caller needs to ensure that `self.album_gain is not None`.
         """
+        item.r128_album_gain = None
         item.rg_album_gain = album_gain.gain
         item.rg_album_peak = album_gain.peak
         item.store()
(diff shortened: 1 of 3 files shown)


snejus:
Does this mean that Navidrome now reads this metadata set by beets?

ShimmerGlass:
@snejus yes, navidrome reads RG_* / R128_* tags, has been for a while

ShimmerGlass:
done :)

Issue #6336, opened 2026-02-02
fetchart: "no art found" when `filesystem` is only source


stephanmax:
I am super confused as I haven’t changed anything recently, but today beets stopped finding local cover art with fetchart. My use case is the following: I download my own cover art, put it next to the files I want to import as `cover.jpg`, and use fetchart with `sources: filesystem` to set `artpath` and then embedart to include my local file in my mp3 tags. `cover.jpg` should be moved with the music files to my library.

That has worked flawlessly for months.

### Problem

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv fetchart album:=<album_name>
```

Led to this problem:

```
user configuration: /home/stephan/.config/beets/config.yaml
data directory: /home/stephan/.config/beets
plugin paths: []
Loading plugins: convert, duplicates, embedart, fetchart, info, inline, mbsync, musicbrainz, permissions, scrub, web, zero
fetchart: lastfm: Disabling art source due to missing key
fetchart: google: Disabling art source due to missing key
artresizer: method is ImageMagick
inline: adding item field multidisc
Sending event: pluginload
library database: /home/stephan/.config/beets/library.db
library directory: /home/stephan/nas/pub/music
Sending event: library_opened
Parsed query: AndQuery([MatchQuery('album', 'Breach', fast=True)])
Parsed sort: NullSort()
fetchart: <artist_name> - <album_name>: no art found
Sending event: cli_exit
```

### Setup

* OS: Linux Arch 6.18.5
* Python version: 3.14.2
* beets version: 2.5.1
* Turning off plugins made problem go away (yes/no): no

But what made the problem go away was changing this bit of config:

```yaml
fetchart:
  sources: filesystem
  enforce_ratio: yes
  minwidth: 1000
```

into this:

```yaml
fetchart:
  sources:
    - filesystem
  enforce_ratio: yes
  minwidth: 1000
```

As if the `sources` field wasn’t parsed correctly. I looked into the code of fetchart quickly and noticed that it is using Confuse’s `as_pairs` to parse the `sources` config field, as opposed to `as_str_seq()` other plugins are using. But it has been that way for years. Are there any other side effects in recent changes or am I missing something here?

My configuration (output of `beet config`) is:

```yaml
directory: ~/nas/pub/music

# --------------- Plugins ---------------

plugins: musicbrainz fetchart embedart convert info zero convert scrub inline permissions web duplicates mbsync

# --------------- Tagging ---------------

per_disc_numbering: yes
[...]

PR #6508, opened 2026-04-06
fix(fetchart): sources definition

ShimmerGlass:
## Description

The fetchart plugin would silently drop unknown sources defined in config, leading to hard to debug problems.
The plugin now errors when an unknown source is configured, or when no sources are configured.
In addition, a single string is now a valid value for `sources` to either enable all sources with an `*`, or a single source.

Fixes: #6336

## To Do

### DIFF
--- a/beets/util/config.py
+++ b/beets/util/config.py
@@ -27,13 +27,16 @@ def sanitize_choices(
 
 
 def sanitize_pairs(
-    pairs: Sequence[tuple[str, str]], pairs_all: Sequence[tuple[str, str]]
+    pairs: Sequence[tuple[str, str]],
+    pairs_all: Sequence[tuple[str, str]],
+    raise_on_unknown: bool = False,
 ) -> list[tuple[str, str]]:
     """Clean up a single-element mapping configuration attribute as returned
     by Confuse's `Pairs` template: keep only two-element tuples present in
     pairs_all, remove duplicate elements, expand ('str', '*') and ('*', '*')
     wildcards while keeping the original order. Note that ('*', '*') and
     ('*', 'whatever') have the same effect.
+    Set raise_on_unknown to raise an error when a provided pair is not recognised
 
     For example,
 
@@ -61,6 +64,17 @@ def sanitize_pairs(
                 res.extend(new)
             elif v == "*":
                 new = [o for o in others if o not in seen and o[0] == k]
+
+                if len(new) == 0 and raise_on_unknown:
+                    raise UnknownPairError(k, v)
+
                 seen.update(new)
                 res.extend(new)
+            elif raise_on_unknown:
+                raise UnknownPairError(k, v)
     return res
+
+
+class UnknownPairError(Exception):
+    def __init__(self, k, v):
+        super().__init__(f"setting {k}={v} is not recognized")
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -37,6 +37,8 @@ New features
   :bug:`5698`
 - :doc:`plugins/replaygain`: Conflicting replay gain tags are now removed on
   write. RG_* tags are removed when setting R128_* and vice versa.
+- :doc:`plugins/fetchart`: Error when a configured source does not exist or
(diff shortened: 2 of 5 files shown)


beets/util/config.py
```diff
@@ -27,13 +27,17 @@ def sanitize_choices(
 
 
 def sanitize_pairs(
-    pairs: Sequence[tuple[str, str]], pairs_all: Sequence[tuple[str, str]]
+    pairs: Sequence[tuple[str, str]],
+    pairs_all: Sequence[tuple[str, str]],
+    raise_on_unknown=False,
```
snejus:
```suggestion
    raise_on_unknown: bool=False,
```

beets/util/config.py
```diff
@@ -27,13 +27,17 @@ def sanitize_choices(
...
 ) -> list[tuple[str, str]]:
     """Clean up a single-element mapping configuration attribute as returned
     by Confuse's `Pairs` template: keep only two-element tuples present in
     pairs_all, remove duplicate elements, expand ('str', '*') and ('*', '*')
     wildcards while keeping the original order. Note that ('*', '*') and
     ('*', 'whatever') have the same effect.
+    Giving raise_on_unknown=True raises an UnknownPairError when a value in
+    pairs does not exist in pairs_all.
```
snejus: Make this message resilient to future refactoring. Something in the lines of

```suggestion
    Set raise_on_unknown to raise an error when a provided pair is not recognised
```

ShimmerGlass:
@snejus all updated, added a changelog entry!

docs/changelog.rst
```diff
@@ -37,6 +37,7 @@ New features
   :bug:`5698`
 - :doc:`plugins/replaygain`: Conflicting replay gain tags are now removed on
   write. RG_* tags are removed when setting R128_* and vice versa.
+- :doc:`plugins/fetchart`: Error when a configured source does not exist.
```
snejus: I think this may be more accurate?

```suggestion
- :doc:`plugins/fetchart`: Error when a configured source does not exist or sources configuration is empty.
```

ShimmerGlass: done!

snejus:
@ShimmerGlass Oh no, this just broke `beetcamp` which dynamically registers its art source:

```py
    def loaded(self) -> None:
        """Add our own artsource to the fetchart plugin."""
        for plugin in plugins.find_plugins():
            if isinstance(plugin, fetchart.FetchArtPlugin):
                if isinstance(fetchart.ART_SOURCES, set):
                    fetchart.ART_SOURCES.add(BandcampAlbumArt)
                else:
                    fetchart.ART_SOURCES[self.data_source] = BandcampAlbumArt
                    fetchart.SOURCE_NAMES[BandcampAlbumArt] = self.data_source
                    fetchart.SOURCES_ALL.append(self.data_source)
                bandcamp_fetchart = BandcampAlbumArt(self._log, self.config)
                plugin.sources = [bandcamp_fetchart, *plugin.sources]
                break
```

Furthermore, I wonder why does it report `bandcamp=*` instead of simply `bandcamp`?
```
UserError: setting bandcamp=* is not recognized
```

My configuration:
```yaml
fetchart:
  lastfm_key: XXX
  auto: yes
  store_source: yes
  sources:
    - filesystem
    - cover_art_url
    - bandcamp
    - lastfm
    - "*"
  maxwidth: 700
```

snejus:
Ignore me: the issue is gone if I remove `bandcamp` from my configuration which is indeed not found since `BandcampAlbumArt` uses capitalised `Bandcamp`:

```py
class BandcampAlbumArt(BandcampRequestsHandler, fetchart.RemoteArtSource):
    NAME = ID = "Bandcamp"
```

ShimmerGlass:
no problem :)

snejus:
@ShimmerGlass Though this question still stands:

> Furthermore, I wonder why does it report `bandcamp=*` instead of simply `bandcamp`?
> ```
> UserError: setting bandcamp=* is not recognized
> ```

PR #5637, opened 2025-02-18
Add TIDAL autotagger plugin

jcjordyn130:
## Description

This is a replacement of the last TIDAL PR I opened back in 2023 (#4641), as in, it is a proper autotagger implementation using the TIDAL API.

Candidate retrieval and support for direct TIDAL album and track URLs are supported.

It also fetches lyrics (both time-synced and not), and embeds it into both the file tags and a LRC file if configured.

Adding testing is a little complicated as testing lyrics support requires a paid account, however, basic metadata queries only require an account with or without an active subscription.

I have been using this plugin for about a week now with my library and it seems to be holding up, obtaining lyrics that are otherwise unavailable through the available lyrics backends.

Poetry files were changed to add optional dependencies for this plugin, which are tidalapi, cachetools (for LFU cache in hot functions), and backoff (to slow down queries when rate limited).

Also, please do let me know if you want me to squash my commits, as I know there are quite a few of them.

## To Do

- [X] Documentation
- [X] ~Changelog~ (Will add once change is approved)
- [X] ~Tests~

PR #6520, opened 2026-04-10
Tidal Metadata Plugin

semohr:
## Description

This PR introduces tidal as metadatasource. It add both an minimal api layer and the typical metadata source plugin capabilities.

### Details

The implementation provides a small API layer consisting of `TidalAPI` for high-level album and track fetching, and `TidalSession` which extends `requests.Session` with token authentication, automatic rate limiting (~4 req/s via `RateLimitAdapter`), and pagination resolution following the JSON:API spec.

Authentication is handled through an OAuth2 PKCE flow accessible via `beet tidal --auth`, with automatic token refresh when the access token expires.

Metadata parsing handles Tidal's JSON:API response format, extracting album and track information including ISO 8601 duration conversion, artist relationships, and copyright/label data.

## Input wanted

The API layer currently lacks comprehensive test coverage. Setting up proper tests would require either mocking all outgoing requests or creating a dedicated test token (which necessitates an account and might require read/write to github secrets). 

Are we comfortable with the current approach of unit testing the plugin itself while mocking all requests?

## TODOs

- [x] Documentation
- [x] `candidate` and `item_candidates` lookup
- [x] It should be possible to optimize batched lookups
- [x] Add tests for candidates and item_candidates
- [x] Implement batching for more than 20 filters

## Refs

thanks to @jcjordyn130 for his initial implementations in #5637 and #4641


beetsplug/tidal/api.py
```diff
@@ -0,0 +1,293 @@
...
+            if res.status_code == 401:
+                self._refresh_token()
+                return self.request(method, url, *args, **kwargs)
+            elif res.status_code == 429:
+                self._handle_rate_limit(res)
+                return self.request(method, url, *args, **kwargs)
+            raise
+
+        return res
+
```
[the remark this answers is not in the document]
semohr: This should never happen realistically as refresh token or handle_rate limit should resolve the error in the next iteration.

Adding max tries here makes this a bit too verbose in my opinion since this is never gona happen anyways.

jcjordyn130:
Yes, I'm still around.

Sorry, life was pretty busy over the past year and I lost track even though I did see your original comment.

The code looks really good, and I never realized TIDAL allows for ISRC/UPC lookup (very useful).

I am wondering, why are we rolling our own API wrapper here?

semohr:
No worries. I know how it goes. Really glad you’re still around!

> I am wondering, why are we rolling our own API wrapper here?

We’ve run into issues before with API dependencies in beets not being well maintained, so this felt like a safer path. On top of that, we likely wouldn’t need most of the features a full-fledged API library would bring anyway.
Another key reason is that I want to keep the API layer decoupled from the plugin design. It does mean a bit more work on our side to maintain it, but in the long run I think the added flexibility and control will be well worth it.

I’m not completely set on this approach, but we recently did something similar with musicbrainz, so it feels like a reasonable direction for now.

---

I also noticed that your original PRs included quite a few more features (art, label regex, popularity, etc.). I’d definitely love to revisit those and would be happy to have you onboard for future enhancements/PRs build ontop of this one (ofc. only if you have the time and are interested in continuing to contribute).

arsaboo:
It may be worthwhile to at least get album art and popularity in this iteration if it is not too much of a change. I understand if we want to leave it for a subsequent PR.

jcjordyn130:
Sounds great to me.

The reason I went with an external API was to avoid having to work on oauth authentication, but it does make more sense to maintain our own wrapper.

I'd be willing to help maintain this plugin. 

The main thing I'm wondering is if we should implement album art here or in fetchart?

arsaboo:
@jcjordyn130 it should probably be part of fetchart (where other sources are identified) as long as the wrapper provides the pipeline.

semohr:
> It may be worthwhile to at least get album art and popularity in this iteration if it is not too much of a change. I understand if we want to leave it for a subsequent PR.

I think reviewing this will already be challenging because of the size :fearful: If we already want to work on these features tho, how about we create additional PRs and base them on this branch? 
I'm was also hesitant to start on those features before we’re confident about the abstraction and architectural decisions made here.

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,531 @@
...
+        AlbumAttributes,
+        ResourceIdentifier,
+        TidalAlbum,
+        TidalArtist,
+        TidalTrack,
+        TrackAttributes,
+    )
+
+
+log = getLogger("beets.tidal")
```
snejus, line 34: Careful with using a name else than `beets` - I found that log messages do not appear if I use any other name.

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,531 @@
...
+                return candidates
+
+        for query in self._item_queries(item):
+            candidates += self.search_tracks_by_query(query)
+
+        log.debug("Found {0} candidates", len(candidates))
+        return candidates
+
+    # ---------------------------------- Search ---------------------------------- #
+
```
snejus:
```suggestion
```

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,531 @@
...
+    @overload
+    def search_tracks_by_ids(
+        self, *, tidal_ids: Iterable[str]
+    ) -> Iterable[TrackInfo | None]: ...
+
+    @overload
+    def search_tracks_by_ids(
+        self, *, isrcs: Iterable[str]
+    ) -> Iterable[TrackInfo | None]: ...
```
snejus, line 234: The types are the same in each case? Why the overload? Oh, I guess that's to show that at least one of these is required?

See `beetsplug._utils.musicbrainz.require_one_of` which may be achieving the same thing? Though your version seems to be cleaner.

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,531 @@
...
+            if item["type"] == "tracks"
+        }
+        artist_lookup: dict[str, TidalArtist] = {
+            item["id"]: item
+            for item in tracks_doc.get("included", [])
+            if item["type"] == "artists"
+        }
+
+        # Yield for IDs first
+        for id in _ids:
```
snejus: There is a builtin function `id`, use `_id` instead:
```suggestion
        for _id in _ids:
```

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,531 @@
...
+                    yield self._get_album_info(
+                        album,
+                        track_lookup=track_lookup,
+                        artist_lookup=artist_lookup,
+                    )
+                else:
+                    yield None
+
+    # ---------------------------------- Parsing --------------------------------- #
+
```
snejus:
```suggestion
```

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,531 @@
...
+            title=self._extract_title(track["attributes"]),
+            isrc=track["attributes"]["isrc"],
+            artist=", ".join(artist_names),
+            artists=artist_names,
+            duration=self._duration_to_seconds(track["attributes"]["duration"]),
+            label=self._extract_label(track["attributes"]),
+        )
+
+    @staticmethod
+    def _extract_artists(
```
snejus: In musicbrainz I called these methods `_parse_...` instead of `_extract_...`. Maybe worth making this consistent?

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,531 @@
...
+            if track := track_lookup.get(track_rel["id"]):
+                track_info = self._get_track_info(track, artist_lookup)
+                track_info.index = i
+                track_infos.append(track_info)
+
+        artist_names, artist_ids = self._extract_artists(
+            album["relationships"]["artists"]["data"],
+            artist_lookup,
+        )
+        release = self._extract_release_date(album["attributes"])
```
snejus:
```suggestion
        date_parts = self._extract_release_date(album["attributes"])
```

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,531 @@
...
+        )
+
+    @staticmethod
+    def _extract_artists(
+        artist_relationships: list[ResourceIdentifier],
+        artist_lookup: dict[str, TidalArtist],
+    ) -> tuple[list[str], list[str]]:
+        """Extract artists from a relationship.
+
+        Needed because artists are ordered in the track relationship,
```
snejus: Seems like this docstring and the one below are being wrapped below max line length.

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,531 @@
...
+    @staticmethod
+    def _extract_data_url(
+        attributes: AlbumAttributes | TrackAttributes,
+    ) -> str | None:
+        if external_links := attributes.get("externalLinks"):
+            return external_links[0].get("href")
+        return None
+
+    @staticmethod
+    def _duration_to_seconds(duration: str) -> int:
```
snejus, line 462: I have a more realistic version of this in beetcamp:

```py
    NUMBER_PAT = cached_patternprop(r"\d+")

    @cached_property
    def duration(self) -> int | None:
        try:
            h, m, s = map(int, self.NUMBER_PAT.findall(self.json_item["duration"]))
        except KeyError:
            return None

        return h * 3600 + m * 60 + s
```

😆

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,531 @@
...
+            + parts["days"] * 86400
+            + parts["months"] * 30 * 86400
+            + parts["years"] * 365 * 86400
+        )
+
+    @staticmethod
+    def _extract_label(
+        attributes: AlbumAttributes | TrackAttributes,
+    ) -> str | None:
+        if copyright := attributes.get("copyright"):
```
snejus, line 474: Nice!

beetsplug/tidal/api.py
```diff
@@ -0,0 +1,319 @@
...
+            log.error(e.response.text, stack_info=False)
+            raise e
+
+        # We update and dont override as the response might not always
+        # contain all information
+        self.token.update(res.json())
+        self.token.save_to(self.token_path)
+
+
+class TidalAPI:
```
snejus: Is it not possible to make use of `RequestHandler` from utils?

beetsplug/tidal/api.py
```diff
@@ -0,0 +1,319 @@
...
+
+        doc: Document[list[Any]] = {
+            "data": [],
+            "included": [],
+            "links": {"next": url},
+        }
+
+        while next := doc.get("links", {}).get("next"):
+            res = self.request(
+                method="GET",
```
snejus: Why not `self.get`?

beetsplug/tidal/api_types.py
```diff
@@ -0,0 +1,176 @@
+from __future__ import annotations
+
+from typing import Generic, Literal, TypedDict, TypeVar
+
+from typing_extensions import NotRequired
+
+# ----------------------------------- Core ----------------------------------- #
+
```
snejus: No need for these sections as they get out of sync and are hard to maintain

beetsplug/tidal/api_types.py
```diff
@@ -0,0 +1,176 @@
...
+    data: T
+    included: NotRequired[list[TidalArtist | TidalAlbum | TidalTrack]]
+    links: NotRequired[dict[str, str]]
+
+
+AlbumDocument = Document[list[TidalAlbum]]
+TrackDocument = Document[list[TidalTrack]]
+SearchDocument = Document[TidalSearch]
+
+# ----------------------------------- Search ---------------------------------- #
```
snejus: Here we go, that's a prime example of what I said above 😆

beetsplug/tidal/authenticate.py
```diff
@@ -0,0 +1,194 @@
...
+from beets import ui
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from beetsplug._typing import JSONDict
+
+
+@dataclass(slots=True)
+class TidalToken:
```
snejus: Can we not make use of `requests-oauthlib`?

test/plugins/test_tidal.py
```diff
@@ -0,0 +1,578 @@
...
+        AlbumAttributes,
+        TidalAlbum,
+        TidalArtist,
+        TidalTrack,
+        TrackAttributes,
+    )
+
+
+# --------------------------------- Mock Data -------------------------------- #
+def _make_artist(id: str, name: str) -> TidalArtist:
```
snejus, line 24: It may be worth rebasing on top of https://github.com/beetbox/beets/pull/65302 where you can make use of data factories!

snejus (changes requested):
👏🏼 that's some good work here, added a couple of comments but nothing very significant

beetsplug/tidal/api.py
jcjordyn130: Also, wouldn't it be a good idea to use a requests session for HTTP connection reuse and easier token injection?

beetsplug/tidal/__init__.py
semohr: This seems to work as expected for me. Things get logged using the `tidal` child logger of beets. Is there an issue I'm missing?

> I found that log messages do not appear if I use any other name.

Makes sense that nothing else shows up as [we](https://github.com/beetbox/beets/blob/ff9f93e6fb27a621f7938266b603aa1c00532d25/beets/ui/__init__.py#L59) configure the logging module like that.

beetsplug/tidal/__init__.py
semohr: That stuff is mainly for my mind while scaffolding. Wanted to remove it before we merge this :+1:

beetsplug/tidal/__init__.py
semohr: Yeah that is basically the typing only way to accomplish the `require_one_of` without introducing a runtime check.

beetsplug/tidal/__init__.py
semohr: Sounds like a good idea! Will adjust it here.

beetsplug/tidal/__init__.py
semohr: Yeah the version is is quite verbose :smile: Was a bit shocked we do not have this as a util somewhere already.

beetsplug/tidal/api.py
semohr: It should be possible thinking about it but wouldn't help too much imo. We still need to create a separate session class and as we use the `pageinated` functions from that custom session quite a bit I don't really see a benefit right now.

Will have a look tho and report back :+1:

beetsplug/tidal/api.py
semohr: @jcjordyn130 We do have this
```
class TidalSession(TimeoutAndRetrySession):
```

Note that `TimeoutAndRetrySession` inherits from `requests.Session`.

beetsplug/tidal/authenticate.py
semohr: Want me to have a look and change it? Tbh I never used that specific library and rolling the oauth myself was quick enough given the pretty good [documentation](https://developer.tidal.com/documentation/api-sdk/api-sdk-authorization) from tidal.

test/plugins/test_tidal.py
semohr: The linked PR does not exist for me :D Might be [github](https://mrshu.github.io/github-statuses/) tho.

beetsplug/tidal/__init__.py
snejus: Logs didn't appear when I used 

```py
log = getLogger(__name__)
```

within `beetsplug/_utils/musicbrainz.py`.

Presumably that's because `beetsplug` is not under `beets`?

semohr: > Presumably that's because beetsplug is not under beets?

Indeed we only attach a streamhandler to the `beets` logger. We might profit from making this more configurable in general. I thought I have seen a related issue & PR some time ago  :thinking:

beetsplug/tidal/api.py
snejus: Remember that it _does_ support custom sessions!

test/plugins/test_tidal.py
snejus: Apologies, that's a mistake on my end editing the URL. It's here: https://github.com/beetbox/beets/pull/6530

semohr: Interesting!

beetsplug/tidal/authenticate.py
snejus: Check whether it reduces the amount of code we have to maintain

semohr: We now use requests-oauthlib for the UI authentication flow. It abstracts some implementation details, which reduces the amount of code we need to maintain, though it does hide a bit of the underlying behavior.

beetsplug/tidal/api.py
semohr: I just tried to use the `RequestHandler` and in its current form i cant make it work here: The session type would need to be made generic, which in turn breaks lyrics.py for reasons I haven’t fully traced yet. It’s probably possible to make it work, but it would require some refactoring in both musicbrainz and lyrics, which is out of scope here imo.

(Ofc using a type ignore would work but didnt want to do it)

beetsplug/_utils/requests.py
```diff
@@ -103,6 +104,41 @@ def request(self, *args, **kwargs):
         return r
 
 
+class RateLimitAdapter(HTTPAdapter):
```
semohr, line 107: I think we need to add retry handling logic into this adapter too. As far as I understand one can only have **one** active adapter. Meaning that the adapter from the `TimeoutAndRetrySession` will not be triggered if we use the `RateLimitAdapter`.

see https://docs.python-requests.org/en/latest/user/advanced/#transport-adapters

@snejus I do not really want to refactor other plugins here. How do we want to approach this? I would just copy it down for now. We okay with that?

snejus: Go ahead!

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,512 @@
...
+
+    def candidates(
+        self, items: Sequence[Item], artist: str, album: str, va_likely: bool
+    ) -> Iterable[AlbumInfo]:
+        candidates: list[AlbumInfo] = []
+        # Tidal allows to lookup via isrc and barcode (nice!)
+        # We just return early here as a lookup via isrc should
+        # return a 100% match
+        barcodes: list[str] = list(
+            filter(None, set(i.get("barcode") for i in items))
```
snejus, line 127: Why barcodes in `candidates` and isrcs in `item_candidates`?

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,512 @@
...
+        for album, artist in itertools.product(album_names, artist_names):
+            yield f"{artist} {album}"
+
+    def search_tracks_by_query(self, query: str) -> Iterable[TrackInfo]:
+        """Search for tracks given a string query."""
+        search_doc = self.api.search_results(
+            query,
+            include=["tracks.artists"],
+        )
+        track_lookup: dict[str, TidalTrack] = {
```
snejus:
```suggestion
        track_by_id: dict[str, TidalTrack] = {
```

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,512 @@
...
+            item["id"]: item
+            for item in search_doc.get("included", [])
+            if item["type"] == "tracks"
+        }
+        artist_lookup: dict[str, TidalArtist] = {
```
snejus:
```suggestion
        artist_by_id: dict[str, TidalArtist] = {
```

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,512 @@
...
+        isrcs = list(isrcs or [])
+        if ids:
+            _ids = list(map(self._extract_id, ids))
+
+        tracks_doc = self.api.get_tracks(
+            ids=list(filter(None, _ids)),
+            isrcs=isrcs,
+            include=["artists"],
+        )
+        track_lookup: dict[str, TidalTrack] = {
```
snejus:
```suggestion
        track_by_id: dict[str, TidalTrack] = {
```

applies to the rest of these variables

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,512 @@
...
+        if external_links := attributes.get("externalLinks"):
+            return external_links[0].get("href")
+        return None
+
+    @staticmethod
+    def _duration_to_seconds(duration: str) -> int:
+        """Convert ISO 8601 duration to seconds. E.g. 'PT15M2S' -> 902."""
+        match = ISO_8601_RE.match(duration)
+        if not match:
+            raise ValueError(f"Invalid ISO 8601 duration: {duration}")
```
snejus, line 466: I'm pretty sure you can trust the format here and assume it's correct

beetsplug/tidal/__init__.py
snejus: Are you going to keep it as it is?

beetsplug/tidal/__init__.py
```diff
@@ -0,0 +1,512 @@
...
+    def _parse_release_date(
+        attributes: AlbumAttributes,
+    ) -> tuple[int, int, int] | None:
+        """Returns year, month, day from iso YYYY-MM-DD"""
+
+        if release_date := attributes.get("releaseDate"):
+            parts = release_date.split("-")
+            if len(parts) != 3:
+                return None
+            return int(parts[0]), int(parts[1]), int(parts[2])
```
snejus: Remove the need for the double `return None`

```suggestion
            if len(parts) == 3:
                return int(parts[0]), int(parts[1]), int(parts[2])
```

snejus: Also you can make it into a single `if` conditional

beetsplug/tidal/api.py
```diff
@@ -0,0 +1,238 @@
...
+        params = {
+            "explicitFilter": explicit_filter,
+            "countryCode": country_code,
+            "include": include or [],
+        }
+
+        return self.get(
+            f"{API_BASE}/searchResults/{urllib.parse.quote(query)}",
+            params=params,
+        ).json()
```
snejus:
```suggestion
        return self.get_json(
            f"{API_BASE}/searchResults/{urllib.parse.quote(query)}",
            params=params,
        )
```

beetsplug/tidal/api.py
```diff
@@ -0,0 +1,238 @@
...
+        return doc
+
+    def ui_authenticate_flow(self) -> None:
+        """Interactive first-time authentication (PKCE flow).
+
+        1. Visit generated URL
+        2. Paste full redirect URL (with ?code=...)
+        3. Token auto-saved for future use
+        """
+        from beets import ui
```
snejus: Why not importing this at the top of the module?

test/plugins/test_tidal.py
snejus: So what about those factories? 😆

snejus (changes requested):
Looks good, I can see `requests-oauthlib` slightly simplified the auth! A couple of comments

beetsplug/tidal/api.py
semohr: I was thinking we only need the `ui` layer for the auth flow. Might be a bit of a premature optimization but e.g. for beets as library one can assume `beets.ui` is never really required. 
Not sure how other code in beets handles this but seems like the cleanest approach here to me.

beetsplug/tidal/__init__.py
semohr: Barcodes (UPC) can only identify albums. This is also reflected in the tidal api: 
- tracks can be looked up via isrc and tidal_id 
- album can be looked up via barcode and tidal_id

ref: https://www.barcode.graphics/music-industry-identifiers-upc-vs-isrc/

As far as I know `item_candidates` shouldnt look for albums, right?

beetsplug/tidal/__init__.py
semohr: Kinda forgot about that. Lets just handle hours and minutes. I don't think there is a track on tidal that has a longer runtime.

beetsplug/tidal/__init__.py
semohr: We still need to satisfy the type checking tho. I do not like to introduce `type: ignore` if handling strange edge cases just requires an additional line of code.

If we do not like the raise here we could alternatively do:
```
assert match is not None, f"Invalid ISO 8601 duration: {duration}"
```
or 
```
if match is None:
    return None
```

beetsplug/tidal/__init__.py
snejus: ah yes this makes sense

beetsplug/tidal/api.py
snejus: We only ever have such imports when we have cyclical import issues - best to keep it at the top of the module

semohr: Alright. FYI: This should be a no issue with python 3.15+ anyways as the version sets lazy imports as a default.

test/plugins/test_tidal.py
semohr: I will have a go at it once we merge #6530. 
I'm not too sure how much it will simplify for the current tests tho. In my opinion the factory functions are quite straightforward and pretty simple already.

beetsplug/tidal/__init__.py
snejus: Or just use my take on this from beetcamp 😆

semohr: You also return `None` ;) Or do you mean use the same code?

I think we cant use the [version from `beetcamp`](https://github.com/snejus/beetcamp/blob/ec4a628458c0991e89f6b48769038166750582bf/beetcamp/track.py#L210) directly here as it is not spec conformant to [ISO8601](https://en.wikipedia.org/wiki/ISO_8601#Durations). 

All parts of the duration are optional and tidal might only return seconds or minutes.
The beetcamp version would not work for `T1H` or `T23S`, right?

snejus: Nope it wouldn't. Bandcamp always returns a fixed format:

```
$ jq < jsons/https:__togetherwithukraine.bandcamp.com_album_together-with-ukraine.json | grep duration
          "duration": "P00H05M01S",
          "duration": "P00H05M50S",
          "duration": "P00H07M18S",
          "duration": "P00H03M40S",
          "duration": "P00H05M04S",
          "duration": "P00H04M42S",
          "duration": "P00H03M58S",
          "duration": "P00H03M52S",
          "duration": "P00H06M04S",
          "duration": "P00H04M06S",
          "duration": "P00H05M00S",
          "duration": "P00H05M04S",
          "duration": "P00H05M11S",
          "duration": "P00H04M10S",
          "duration": "P00H05M28S",
          "duration": "P00H06M15S",
          "duration": "P00H04M11S",
```

snejus: I thought this was a standard format and TIDAL follows it strictly too?

semohr: I’m not entirely sure how Tidal handles this internally, it’s quite possible they also rely on a fixed format under the hood. However, the Tidal api json explicitly specifies that this field is an ISO 8601 duration.

From that perspective, it seems safer and more future-proof to adhere to the documented spec rather than relying on any undocumented or implementation-specific formatting that may change over time.

---

I’ve updated the function to return `None` when parsing fails and added a warning in those cases. This should make issues easier to spot and debug going forward.

test/plugins/test_tidal.py
snejus: True!

snejus (approved):
Tests need fixing otherwise I'm happy with it

Issue #6441, opened 2026-03-14
mbpseudo: `data_source` set to `MusicBrainz` rather than `MusicBrainzPseudoRelease` during import


amogus07:
I'm not sure how it's intended to work, but when I import albums with mbpseudo, `data_source` is initially set to `MusicBrainz`. Only when I run `beet mbsync`, it's set to `MusicBrainzPseudoRelease`.

### Setup

* OS: Fedora Linux 43 (Workstation Edition)
* Python version: 3.13.5
* beets version: 2.7.1
* Turning off plugins made problem go away (yes/no): no

amogus07:
The PR doesn't seem to have solved the problem. I can still reproduce it without the chroma plugin too.

PR #6522, opened 2026-04-10
chroma: gate candidates on the musicbrainz plugin being enabled

### LINKED ISSUE #6212: musicbrainz metadata source is used when the musicbrainz plugin is not activated
The musicbrainz metadata source is used even when the musicbrainz plugin is not activated.

### Problem

```sh
beet -vv import /music/downloads/
```

Led to this problem: musicbrainz results are present.
I didn't copy the full log because it is very long, but I can add parts if needed.

```
/music/downloads/Back in Black (10 items)

Finding tags for album "AC/DC - Back in Black".
  Candidates:
  1. (64.7%) AC/DC - Back In Black
             ≠ tracks, id, label
             Spotify, None, 1980, None, Columbia, None, None
  2. (44.1%) AC/DC - 5ive of a Kind - Back In Black (Live)
             ≠ tracks, id, unmatched tracks, ...
             Spotify, None, 1980, None, 5ive of a Kind, None, None
  3. (31.3%) AC/DC - Bonfire
             ≠ missing tracks, id, mediums, ...
             MusicBrainz, 5xCD, 1997, US, EastWest, 62119-2, None
  4. (31.0%) AC/DC - Bonfire
             ≠ missing tracks, id, mediums, ...
             MusicBrainz, 5xCD, 1998, XE, EastWest, 7559-62119-2, None
  5. (30.5%) AC/DC - Bonfire
             ≠ missing tracks, id, mediums, ...
             MusicBrainz, 5xCD, 2011, XE, Columbia, 88697856822, None
  6. (19.9%) AC/DC - The Ultimate Best of AC/DC (Remastered)
             ≠ missing tracks, tracks, id, ...
             MusicBrainz, Digital Media, 2011, None, None, None, None
  7. (6.7%) AC/DC - The Complete Collection
             ≠ missing tracks, mediums, id, ...
             MusicBrainz, 27xDigital Media, 2012, None, Columbia, None, None
➜ # selection (default 1), Skip, Use as-is, as Tracks, Group albums,
Enter search, enter Id, aBort?
```

### Setup

* OS: Linux (docker)
* Python version: 3.12.12
* beets version: 2.5.1
```
$ beet --version
beets version 2.5.1
Python version 3.12.12
plugins: chroma, convert, embedart, fetchart, info, lastgenre, lyrics, permissions, replaygain, scrub, spotify, zero
```

```yaml

plugins: spotify fetchart embedart convert scrub replaygain lastgenre chroma lyrics permissions info zero
directory: /music/library

art_filename: albumart

# --------------- Performance ---------------

threaded: yes
original_date: yes

# --------------- Tagging ---------------

per_disc_numbering: no
convert:
    auto: yes
    ffmpeg: /usr/bin/ffmpeg
    max_bitrate: 320
    threads: 1
    delete_originals: no
    dest:
    pretend: no
    link: no
    hardlink: no
    format: mp3
    id3v23: inherit
    write_metadata: yes
    formats:
        aac:
            command: ffmpeg -i $source -y -vn -acodec aac -aq 1 $dest
            extension: m4a
        alac:
            command: ffmpeg -i $source -y -vn -acodec alac $dest
            extension: m4a
        flac: ffmpeg -i $source -y -vn -acodec flac $dest
        mp3: ffmpeg -i $source -y -vn -aq 2 $dest
        opus: ffmpeg -i $source -y -vn -acodec libopus -ab 96k $dest
        ogg: ffmpeg -i $source -y -vn -acodec libvorbis -aq 3 $dest
        wma: ffmpeg -i $source -y -vn -acodec wmav2 -vn $dest
    auto_keep: no
    tmpdir:
    quiet: no
    embed: yes
    paths: {}
    no_convert: ''
    copy_album_art: no
    album_art_maxwidth: 0
    playlist:
paths:
    default: $albumartist/$album%aunique{}/$track - $title
    singleton: Singles/$artist - $title
    albumtype_soundtrack: Soundtracks/$album/$track $title

import:
    autotag: yes
    write: yes
    copy: yes
    move: no
    resume: ask
    quiet: no
    quiet_fallback: asis
    timid: no
    log: /config/beet.log
spotify:
    data_source_mismatch_penalty: 0.0
    search_limit: 5
    search_query_ascii: no
    mode: list
    tiebreak: popularity
    show_failures: no
    region_filter:
    regex: []
    client_id: REDACTED
    client_secret: REDACTED
    tokenfile: spotify_token.json
zero:
    fields: albumdisambig
    auto: yes
    keep_fields: []
    update_database: no
    omit_single_disc: no

match:
    strong_rec_thresh: 0.1
    ignored_media:
    - Data CD
    - DVD
    - DVD-Video
    - Blu-ray
    - HD-DVD
    - VCD
    - SVCD
    - UMD
    - VHS
lastgenre:
    auto: yes
    source: album
    whitelist: yes
    min_weight: 10
    count: 1
    fallback:
    canonical: no
    force: no
    keep_existing: no
    separator: ', '
    prefer_specific: no
    title_case: yes
    extended_debug: no
embedart:
    auto: yes
    remove_art_file: yes
    maxwidth: 0
    compare_threshold: 0
    ifempty: no
    quality: 0
fetchart:
    auto: yes
    sources: spotify
    store_source: yes
    minwidth: 0
    maxwidth: 0
    quality: 0
    max_filesize: 0
    enforce_ratio: no
    cautious: no
    cover_names:
    - cover
    - front
    - art
    - album
    - folder
    high_resolution: no
    deinterlace: no
    cover_format:
    lastfm_key: REDACTED
    fanarttv_key: REDACTED
    google_key: REDACTED
    google_engine: REDACTED
replaygain:
    auto: no
    overwrite: no
    backend: command
    threads: 12
    parallel_on_import: no
    per_disc: no
    peak: 'true'
    targetlevel: 89
    r128: [Opus]
    r128_targetlevel: 84
    command: ''
    noclip: yes
scrub:
    auto: yes
replace:
    ^\.: _
    '[\x00-\x1f]': _
    '[<>:"\?\*\|]': _
    '[\xE8-\xEB]': e
    '[\xEC-\xEF]': i
    '[\xE2-\xE6]': a
    '[\xF2-\xF6]': o
    '[\xF8]': o
    \.$: _
    \s+$: ''
lyrics:
    auto: yes
    forced: yes
    sources: [lrclib, genius, tekstowo]
    synced: yes
    translate:
        api_key: REDACTED
        from_languages: []
        to_language:
    dist_thresh: 0.11
    google_API_key: REDACTED
    google_engine_ID: REDACTED
    genius_api_key: REDACTED
    fallback:
    force: no
    local: no
    print: no
permissions:
    file: 644
    dir: 755
disabled_plugins: []
chroma:
    search_limit: 5
    data_source_mismatch_penalty: 0.5
    auto: yes
```

AnonymousAAArdvark:
## Summary

Fixes #6212. The ``chroma`` plugin used Acoustid fingerprinting to look
up MusicBrainz release and recording IDs, then unconditionally resolved
those IDs via MusicBrainz — even when the user had not enabled the
``musicbrainz`` plugin. The result: users with (e.g.) only
``chroma`` + ``spotify`` enabled would still see MusicBrainz-sourced
album and track candidates appear in the autotagger, defeating their
choice to opt out of MusicBrainz as a metadata source.

This PR replaces the direct ``MusicBrainzPlugin()`` instantiation with a
lookup through ``metadata_plugins.get_metadata_source(\"musicbrainz\")``.
When the musicbrainz plugin is not loaded, both ``candidates`` and
``item_candidates`` short-circuit and return empty. Acoustid fingerprinting
itself is unaffected — the ``acoustid_id`` and ``acoustid_fingerprint``
item fields are still populated exactly as before.

Note that the original issue thread mentions only the album-level
``candidates`` path, but the same bug exists in ``item_candidates`` (the
singleton/track path). Both are fixed.

### Bonus fix

The previous ``cached_property mb = MusicBrainzPlugin()`` pattern created a
second ``MusicBrainzPlugin`` instance outside the plugin registry. This
silently bypassed ``beetsplug/mbpseudo.py``, which registers a replacement
instance for the musicbrainz plugin — meaning chroma-triggered lookups were
never routed through the pseudo-release-aware code. Routing through
``get_metadata_source`` fixes this as a side effect: chroma now uses the
same shared instance as the rest of beets.

### Alternative considered

The issue reporter's maintainer response suggested a richer approach:
extract cross-reference IDs (Spotify, Discogs, etc.) from the MusicBrainz
release response and route them to whichever metadata plugins are
enabled. That's a good direction, but (a) it only applies to
``candidates`` — MusicBrainz recording responses don't carry cross-source
track IDs, so ``item_candidates`` needs the gating fix regardless; and
(b) it would still want ``get_metadata_source`` as its foundation.
This PR is intentionally scoped as the minimal gating fix; I'd be happy
to follow up with the cross-reference routing as a separate PR if the
maintainers want it.

## Test plan

- [x] New unit tests in ``test/plugins/test_chroma.py`` covering all four
      cross-product cases: ``candidates`` / ``item_candidates`` × with /
      without the musicbrainz plugin loaded. The ``without musicbrainz``
      cases directly pin the issue-6212 regression.
- [x] Existing chroma tests (``chromasearch_*``) still pass.
- [x] Full focused regression subset passes (144 tests across
      ``test/plugins/test_chroma.py``, ``test/plugins/test_musicbrainz.py``,
      ``test/test_metadata_plugins.py``, ``test/autotag/test_match.py``,
      ``test/autotag/test_distance.py``).
- [x] ``ruff check`` + ``ruff format --check`` clean on changed files.
- [x] ``mypy beetsplug/chroma.py`` clean.
- [x] ``sphinx-lint docs/plugins/chroma.rst docs/changelog.rst`` clean.
- [x] Changelog entry added under Unreleased → Bug fixes.
- [x] Note added to ``docs/plugins/chroma.rst`` explaining that chroma
      requires musicbrainz to produce candidates.

Disclosure: this contribution was made while working on an open-source
assignment for EECS 481 (Software Engineering) at the University of
Michigan.

### DIFF
--- a/beetsplug/chroma.py
+++ b/beetsplug/chroma.py
@@ -21,7 +21,7 @@ from __future__ import annotations
 import heapq
 import re
 from collections import defaultdict
-from functools import cached_property, partial
+from functools import partial
 from typing import TYPE_CHECKING
 
 import acoustid
@@ -29,15 +29,15 @@ import confuse
 
 from beets import config, ui, util
 from beets.autotag.distance import Distance
-from beets.metadata_plugins import MetadataSourcePlugin
+from beets.metadata_plugins import MetadataSourcePlugin, get_metadata_source
 from beets.util.color import colorize
-from beetsplug.musicbrainz import MusicBrainzPlugin
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator
 
     from beets.autotag.hooks import TrackInfo
     from beets.library.models import Item
+    from beetsplug.musicbrainz import MusicBrainzPlugin
 
 API_KEY = "1vOwZtEn"
 SCORE_THRESH = 0.5
@@ -196,9 +196,24 @@ class AcoustidPlugin(MetadataSourcePlugin):
             self.register_listener("import_task_start", self.fingerprint_task)
         self.register_listener("import_task_apply", apply_acoustid_metadata)
 
-    @cached_property
-    def mb(self) -> MusicBrainzPlugin:
-        return MusicBrainzPlugin()
+    def _get_musicbrainz(self) -> MusicBrainzPlugin | None:
+        """Return the loaded MusicBrainz plugin instance, or ``None``.
+
+        Acoustid fingerprint lookups return MusicBrainz release and
+        recording IDs, so chroma can only produce autotagger candidates
+        by resolving those IDs through the musicbrainz plugin. When the
+        user has not enabled the ``musicbrainz`` plugin, we must not
+        return any candidates — otherwise MusicBrainz-sourced results
+        would silently appear during tagging even though the user
+        opted out of that metadata source.
+        """
+        plugin = get_metadata_source("musicbrainz")
+        if plugin is None:
+            self._log.debug(
+                "musicbrainz plugin not enabled; "
+                "acoustid matches will not produce candidates"
+            )
+        return plugin  # type: ignore[return-value]
 
     def fingerprint_task(self, task, session):
         return fingerprint_task(self._log, task, session)
@@ -214,9 +229,13 @@ class AcoustidPlugin(MetadataSourcePlugin):
         return dist
 
     def candidates(self, items, artist, album, va_likely):
+        mb = self._get_musicbrainz()
+        if mb is None:
+            return []
+
         albums = []
         for relid in prefix(_all_releases(items), MAX_RELEASES):
-            album = self.mb.album_for_id(relid)
+            album = mb.album_for_id(relid)
             if album:
                 albums.append(album)
 
@@ -227,10 +246,14 @@ class AcoustidPlugin(MetadataSourcePlugin):
         if item.path not in _matches:
             return []
 
+        mb = self._get_musicbrainz()
+        if mb is None:
+            return []
+
         recording_ids, _ = _matches[item.path]
         tracks = []
         for recording_id in prefix(recording_ids, MAX_RECORDINGS):
-            track = self.mb.track_for_id(recording_id)
+            track = mb.track_for_id(recording_id)
             if track:
                 tracks.append(track)
         self._log.debug("acoustid item candidates: {}", len(tracks))
--- a/beetsplug/chroma.py
+++ b/beetsplug/chroma.py
@@ -29,13 +29,17 @@ import confuse
 
 from beets import config, ui, util
 from beets.autotag.distance import Distance
-from beets.metadata_plugins import MetadataSourcePlugin, get_metadata_source
+from beets.metadata_plugins import (
+    MetadataSourcePlugin,
+    find_metadata_source_plugins,
+    get_metadata_source,
+)
 from beets.util.color import colorize
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator
 
-    from beets.autotag.hooks import TrackInfo
+    from beets.autotag.hooks import AlbumInfo, TrackInfo
     from beets.library.models import Item
     from beetsplug.musicbrainz import MusicBrainzPlugin
 
@@ -45,6 +49,16 @@ TRACK_ID_WEIGHT = 10.0
 COMMON_REL_THRESH = 0.6  # How many tracks must have an album in common?
 MAX_RECORDINGS = 5
 MAX_RELEASES = 5
+
+# External metadata sources that ``musicbrainz.album_info`` knows how to
+# extract cross-reference IDs for via the release's ``url-relations``.
+# Kept in sync with the default keys of ``musicbrainz.external_ids`` in
+# :class:`beetsplug.musicbrainz.MusicBrainzPlugin`. Acoustid fingerprint
+# matches can be routed to any of these sources when the corresponding
+# metadata-source plugin is enabled.
+MB_CROSS_REF_SOURCES = frozenset(
+    {"discogs", "bandcamp", "spotify", "deezer", "tidal"}
+)
 
 # Stores the Acoustid match information for each track. This is
 # populated when an import task begins and then used when searching for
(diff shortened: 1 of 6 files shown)


AnonymousAAArdvark:
Quick note on the CI state:

Check docs: real failure docstrfmt --preserve-adornments wanted my changelog entry wrapped slightly differently than I had it. Fixed in 14b26f097.
Run tests (windows-latest, 3.10–3.13): these failed at the choco install mp3gain step with an HTTP 499 from community.chocolatey.org/api/v2/ ([example job log](https://github.com/beetbox/beets/actions/runs/24268765998/job/70869321563)). Looks like a transient Chocolatey infrastructure hiccup.

beetsplug/chroma.py
```diff
@@ -196,9 +196,24 @@ def __init__(self):
...
+        Acoustid fingerprint lookups return MusicBrainz release and
+        recording IDs, so chroma can only produce autotagger candidates
+        by resolving those IDs through the musicbrainz plugin. When the
+        user has not enabled the ``musicbrainz`` plugin, we must not
+        return any candidates — otherwise MusicBrainz-sourced results
+        would silently appear during tagging even though the user
+        opted out of that metadata source.
+        """
+        plugin = get_metadata_source("musicbrainz")
+        if plugin is None:
```
semohr: > acoustid matches will not produce candidates"

This is **not** what we want. `acoustid` should still get the musicbrainz matches (even if not enabled) and extract the ids for other sources if available.

As is we now couple `acoustid` with `musicbrainz` which is not optimal in my opinion.

semohr: Please also see https://github.com/beetbox/beets/issues/6212#issuecomment-3636951429

semohr: For example if you look at https://musicbrainz.org/release/5ecd63c6-d875-46b3-9a49-12e513d41663
This musicbrainz release has external links to apple music, beatport, spotify, deezer and tidal (see on the right). 

We should be able to get this information and if a user has e.g. the spotify plugin enabled use the external link from the musicbrainz release to retrieve the spotify release.

The idea from me was that even tho we require the musicbrainz plugin here it is not necessary to use the candidates from the musicbrainz plugin.

Just disallowing the use of the `acoustid` plugin without `musicbrainz` enabled seems like the easy way out ;)

amogus07:
> ### Bonus fix
> 
> The previous `cached_property mb = MusicBrainzPlugin()` pattern created a second `MusicBrainzPlugin` instance outside the plugin registry. This silently bypassed `beetsplug/mbpseudo.py`, which registers a replacement instance for the musicbrainz plugin — meaning chroma-triggered lookups were never routed through the pseudo-release-aware code. Routing through `get_metadata_source` fixes this as a side effect: chroma now uses the same shared instance as the rest of beets.

would this fix #6441?

semohr:
As it is yes it would fix the issue with mbpseudo. 

As I said in the other comments tho it feels a bit like a hacky and lazy way out. Musibrainz does give identifiers for other metadata sources which we could use to resolve to the proper (enabled) metadata plugin. 

But maybe these are two different issues to tackle.

AnonymousAAArdvark:
Thanks for the review @semohr, I've reworked the PR to do the cross-reference routing instead of gating. Latest commit bc956dbfa:

What chroma does now:

1. Always queries MusicBrainz to resolve the acoustid release ID (prefers the loaded musicbrainz plugin from the registry so mbpseudo still takes effect; falls back to a private, transient MusicBrainzPlugin instance when the user hasn't enabled musicbrainz).
2. Passes a new extra_external_sources keyword argument to MusicBrainzPlugin.album_for_id / album_info so chroma can force-extract cross-reference IDs for exactly the external sources whose metadata plugins are loaded — without asking the user to globally opt into musicbrainz.external_ids.
3. For each resolved MB AlbumInfo, yields the MB candidate (only when the musicbrainz plugin itself is loaded) and dispatches to each loaded external metadata plugin via album_for_id(info.{source}_album_id) for the Discogs / Bandcamp / Spotify / Deezer / Tidal cross-references in url-relations.


So:

- chroma + spotify (no musicbrainz) → Spotify candidates for releases MB has a Spotify url-relation for. Addresses your AC/DC example directly.
- chroma + musicbrainz → MB candidates, unchanged from today.
- chroma + musicbrainz + spotify → both kinds of candidates.
- chroma alone (no MB, no cross-ref targets) → empty, no network call made.

Test coverage is in test/plugins/test_chroma.py. Four classes covering all four cross-product cases (no compatible sources / MB only / external only / both), plus asserts that extra_external_sources is passed correctly and that external plugins are only called when their cross-ref ID is present on the MB AlbumInfo. Full suite (2234 tests) still passes locally; ruff / docstrfmt / sphinx-lint clean.

semohr:
Neat, this is more or less how I imagined it. Thanks for the initiative :pray:  

As the changes touch musicbrainz now I would like to ping @snejus. Adding a additional argument to `album_for_id` seems like a bad idea to me. Any ideas for a less intrusive workaround?

Im gona review this properly once we are okay with the high level decicion on how to integrate musicbrainz.

beetsplug/chroma.py
```diff
@@ -192,13 +206,104 @@ def __init__(self):
...
+        self._private_mb: MusicBrainzPlugin | None = None
+
         if self.config["auto"]:
             self.register_listener("import_task_start", self.fingerprint_task)
         self.register_listener("import_task_apply", apply_acoustid_metadata)
 
-    @cached_property
-    def mb(self) -> MusicBrainzPlugin:
-        return MusicBrainzPlugin()
+    def _musicbrainz_client(self) -> MusicBrainzPlugin:
```
semohr, line 220: We should make this a [cached property](https://docs.python.org/3/library/functools.html#functools.cached_property). Should simplify this slightly.

amogus07:
also related: #6524

snejus:
> * Passes a new extra_external_sources keyword argument to MusicBrainzPlugin.album_for_id / album_info so chroma can force-extract cross-reference IDs for exactly the external sources whose metadata plugins are loaded — without asking the user to globally opt into musicbrainz.external_ids.
> 
> * For each resolved MB AlbumInfo, yields the MB candidate (only when the musicbrainz plugin itself is loaded) and dispatches to each loaded external metadata plugin via album_for_id(info.{source}_album_id) for the Discogs / Bandcamp / Spotify / Deezer / Tidal cross-references in url-relations.

This to me seems to be well outside the scope of `chroma` plugin's responsibility. These external source IDs are **always** available when we query musicbrainz, why should `chroma` be responsible for handling them?

It seems like a large additional feature that should exist somewhere else - can we keep this focused to the first point / fix related to mbpseudo?

Would this mean that `chroma` without `musicbrainz` enabled would essentially be pointless? If that's the case, I would still want to see handling of external sources in a separate PR.

semohr:
The idea here was to implement #6212 which predates mbpseudo. The mbpseudo fix is just a sideeffect.

The idea of the PR was to implement the features of forwarding the candisates lookups to only enabled datasources, not to fix the mbpseudo issue. I would like to keep it scoped like that.

semohr:
> Would this mean that chroma without musicbrainz enabled would essentially be pointless? If that's the case, I would still want to see handling of external sources in a separate PR.

Chroma (before) only ever produced musicbrainz candidates. Imo it would be consistent to only import/show candidates of the enabled data sources.

snejus:
> The idea here was to implement #6212 which predates mbpseudo. The mbpseudo fix is just a sideeffect.

Regarding the issue scope, it seems to me that the expected behaviour would be to not return anything when musicbrainz plugin is disabled? Happy that it fixes the mbpseudo issue as a side effect!

Routing lookups to the rest of data source plugins is a good idea, but best to be moved to a separate PR.

semohr:
> Regarding the issue scope, it seems to me that the expected behaviour would be to not return anything when musicbrainz plugin is disabled?

This might break some edge-case workflows (as usual 😄), but as long as we document the behavior clearly, it should be acceptable. (Imo also the expected behavior.)

---

My suggestion would be to:

- Open a separate PR to explicitly disable candidates if mb is not enabled (and fix the mbpseudo issue)
- Then rebase this PR on top of it, so we don’t lose the progress already made here (especially around the forwarding feature)

@AnonymousAAArdvark would you be up for taking this on as well? Specifically, cherry-pick the fixes for the mbpseudo issue from this PR and restrict candidates to only those coming from MusicBrainz
If not, we can always pick it up.

AnonymousAAArdvark:
Done — opened #6532 with just the gating fix (gate `candidates` / `item_candidates` on musicbrainz being loaded via the metadata-source registry + mbpseudo side-effect fix). Clean branch off current master, no changes to `musicbrainz.py` or `mbpseudo.py`.

I'll keep this PR open for the cross-reference routing follow-up. Once #6532 lands I can rebase this branch on top.

semohr:
Hey @AnonymousAAArdvark  you still interested in picking up this PR? I marked it as draft for now feel free to mark it ready to review again once you have rebased your changes.

AnonymousAAArdvark:
Thanks for checking in! Feel free to take this over, I won't have bandwidth to continue in the near term.

PR #6523, opened 2026-04-11
Retry listenbrainz requests

aereaux:
## Description

Fixes problems with not all listens being fetched when encountering temporary network failures.

I don't think we have a way to simulate network failures in tests, but I might be wrong about that.

## To Do

### DIFF
--- a/beetsplug/listenbrainz.py
+++ b/beetsplug/listenbrainz.py
@@ -15,6 +15,7 @@ from beets.plugins import BeetsPlugin
 
 from ._utils.musicbrainz import MusicBrainzAPIMixin
 from ._utils.playcount import update_play_counts
+from ._utils.requests import TimeoutAndRetrySession
 
 if TYPE_CHECKING:
     from ._utils.playcount import Track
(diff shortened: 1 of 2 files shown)


semohr (changes requested):
We already have infrastructure inplace for retry logic. Have a look at [beetsplug/_utils/request.py](https://github.com/beetbox/beets/blob/0f2bb215ecbf541720412c146df280528929ba92/beetsplug/_utils/requests.py#L65). 

It is relative new but let's try to unify :)

aereaux:
I didn't notice that, but that's perfect.  I updated this PR to use that infrastructure.

beetsplug/listenbrainz.py
```diff
@@ -15,6 +15,9 @@
 
 from ._utils.musicbrainz import MusicBrainzAPIMixin
 from ._utils.playcount import update_play_counts
+from ._utils.requests import TimeoutAndRetrySession
+
+session = TimeoutAndRetrySession()
```
semohr: I'm in a bit of a war with import sideeffect in beets. They make working with beets as a library instead of a the typical cli usecase unpredictable and hog memory.

Not the biggest issue in this case but could we please construct the session on demand or when the ListenBrainzPlugin plugin initializes?

Highly appreciated 🙏

aereaux: Sure!  I don't think the constructor for this object has many side effects, but I moved it to the LBPlugin constructor

semohr (approved):

PR #6535, opened 2026-04-14
Switch to uv

### LINKED ISSUE #5783: Upgrade poetry to v2

snejus:
I think we've had enough of poetry so I let Claude Code migrate the code base to `uv`. Let's see.

Fixes: #5783


semohr:
We might want to consider migrating to [workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces/) and separating beetsplug and beets.

snejus:
> We might want to consider migrating to [workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces/) and separating beetsplug and beets.

Hm, it may not apply to our use case?

> Inspired by the [Cargo](https://doc.rust-lang.org/cargo/reference/workspaces.html) concept of the same name, a workspace is "a collection of one or more packages, called workspace members, that are managed together."

> Workspaces organize large codebases by splitting them into multiple packages with common dependencies. Think: a FastAPI-based web application, alongside a series of libraries that are versioned and **maintained as separate Python packages**, all in the same Git repository.

snejus:
Just attempted to build the package to see whether it suffers from the same issue (#5770):

```sh
$ uv build
Building source distribution...
Building wheel from source distribution...
Successfully built dist/beets-2.9.0.tar.gz
Successfully built dist/beets-2.9.0-py3-none-any.whl
$ tar -vtf dist/beets-2.9.0.tar.gz | head -10
-rw-r--r-- 0/0            1631 2020-02-02 00:00 beets-2.9.0/beets/__init__.py
-rw-r--r-- 0/0             825 2020-02-02 00:00 beets-2.9.0/beets/__main__.py
-rw-r--r-- 0/0            4643 2020-02-02 00:00 beets-2.9.0/beets/config_default.yaml
-rw-r--r-- 0/0             646 2020-02-02 00:00 beets-2.9.0/beets/context.py
-rw-r--r-- 0/0            6766 2020-02-02 00:00 beets-2.9.0/beets/logging.py
-rw-r--r-- 0/0            1021 2020-02-02 00:00 beets-2.9.0/beets/mediafile.py
-rw-r--r-- 0/0           15166 2020-02-02 00:00 beets-2.9.0/beets/metadata_plugins.py
-rw-r--r-- 0/0           23681 2020-02-02 00:00 beets-2.9.0/beets/plugins.py
-rw-r--r-- 0/0               0 2020-02-02 00:00 beets-2.9.0/beets/py.typed
-rw-r--r-- 0/0            1540 2020-02-02 00:00 beets-2.9.0/beets/autotag/__init__.py
```

It does not use the last file modification dates but AT LEAST it doesn't go back to 1970 like utterly useless **poetry>2**:

```sh
$ uvx poetry'>2' build
Installed 45 packages in 143ms
Building beets (2.9.0)
Building sdist
  - Building sdist
  - Built beets-2.9.0.tar.gz
Building wheel
  - Building wheel
  - Built beets-2.9.0-py3-none-any.whl
$ tar -vtf dist/beets-2.9.0.tar.gz | head -10
-rw-r--r-- 0/0            1080 1970-01-01 01:00 beets-2.9.0/LICENSE
-rw-r--r-- 0/0            5316 1970-01-01 01:00 beets-2.9.0/README.rst
-rw-r--r-- 0/0            1631 1970-01-01 01:00 beets-2.9.0/beets/__init__.py
-rw-r--r-- 0/0             825 1970-01-01 01:00 beets-2.9.0/beets/__main__.py
-rw-r--r-- 0/0            1540 1970-01-01 01:00 beets-2.9.0/beets/autotag/__init__.py
-rw-r--r-- 0/0           18987 1970-01-01 01:00 beets-2.9.0/beets/autotag/distance.py
-rw-r--r-- 0/0           22492 1970-01-01 01:00 beets-2.9.0/beets/autotag/hooks.py
-rw-r--r-- 0/0           13618 1970-01-01 01:00 beets-2.9.0/beets/autotag/match.py
-rw-r--r-- 0/0            4643 1970-01-01 01:00 beets-2.9.0/beets/config_default.yaml
-rw-r--r-- 0/0             646 1970-01-01 01:00 beets-2.9.0/beets/context.py
```

The issue was that `zipfile.ZipInfo` does not support dates before 1980.

amogus07:
> [fyn](https://github.com/duriantaco/fyn) is another option, now that Astral (the company behind uv) was bought by OpenAI. It also has real-world benefits compared to uv like a task runner (which would replace poethepoet I suppose) as well as a `shell` command.

How about using that instead?

snejus:
> > [fyn](https://github.com/duriantaco/fyn) is another option, now that Astral (the company behind uv) was bought by OpenAI. It also has real-world benefits compared to uv like a task runner (which would replace poethepoet I suppose) as well as a `shell` command.
> 
> How about using that instead?

It hasn't been adopted as widely as uv. I've been resisting the migration to `uv` until now for the exact same reason.

semohr:
> > We might want to consider migrating to [workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces/) and separating beetsplug and beets.
> 
> Hm, it may not apply to our use case?
> 
> > Inspired by the [Cargo](https://doc.rust-lang.org/cargo/reference/workspaces.html) concept of the same name, a workspace is "a collection of one or more packages, called workspace members, that are managed together."
> 
> > Workspaces organize large codebases by splitting them into multiple packages with common dependencies. Think: a FastAPI-based web application, alongside a series of libraries that are versioned and **maintained as separate Python packages**, all in the same Git repository.

I think we could make it work. For example, we could split the core, database, and UI layers into separate packages. That would give us clearer boundaries and enforce a more explicit separation of responsibilities, though it might introduce some additional complexity in terms of packaging and coordination.

snejus:
I'd be interested to see how would it work and what benefits it may bring - but of course not as part of this PR but rather a separate piece of work that builds on this.

semohr:
> but of course not as part of this PR but rather a separate piece of work that builds on this.

Agreed!

snejus:
So what are we thinking regarding this? @semohr

.github/workflows/ci.yaml
```diff
@@ -28,12 +28,10 @@ jobs:
       - uses: actions/checkout@v5
       - name: Install Python tools
         uses: BrandonLWhite/pipx-install-action@v1.0.3
-      - name: Setup Python with poetry caching
-        # poetry cache requires poetry to already be installed, weirdly
-        uses: actions/setup-python@v6
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
```
semohr: Any specific reason we use v5? I think v8 is the current version

.github/workflows/ci.yaml
```diff
@@ -28,12 +28,10 @@ jobs:
       - uses: actions/checkout@v5
       - name: Install Python tools
         uses: BrandonLWhite/pipx-install-action@v1.0.3
```
semohr: Do we even need pipx anymore? I dont think it is used in the ci pipeline.

.github/workflows/integration_test.yaml
```diff
@@ -14,13 +14,13 @@ jobs:
       - uses: actions/checkout@v5
       - name: Install Python tools
         uses: BrandonLWhite/pipx-install-action@v1.0.3
-      - uses: actions/setup-python@v6
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
```
semohr: Same as above.

.github/workflows/integration_test.yaml
```diff
@@ -14,13 +14,13 @@ jobs:
...
         with:
           python-version: ${{ env.PYTHON_VERSION }}
-          cache: poetry
 
       - name: Install dependencies
-        run: poetry install
+        run: uv sync
```
semohr: I think we should always use uv sync --locked or --frozen in the CI. We can set env variables for that globally.

https://docs.astral.sh/uv/reference/cli/#uv-sync--frozen
https://docs.astral.sh/uv/reference/cli/#uv-sync--locked

CONTRIBUTING.rst
```diff
@@ -65,71 +65,66 @@ Development Tools
...
-instructions <https://pipx.pypa.io/stable/how-to/install-pipx/>`__ or
+poethepoet_ is a stand-alone CLI tool. The Python community recommends using
+pipx_ to install such tools in isolated environments so their dependencies don't
+interfere with your system. If you don't have pipx_ installed, follow the `pipx
+installation instructions <https://pipx.pypa.io/stable/how-to/install-pipx/>`__
+or run:
 
 .. code-block:: sh
 
     $ python3 -m pip install --user pipx
```
semohr: Is the pipx recommendation still valid? I think poethepoet is now just a dev dependency. I dont think it is necessary anymore. Just running `uv sync` should do the job nicely.

pyproject.toml
```diff
@@ -24,179 +22,142 @@ classifiers = [
...
-    { include = "beetsplug" },
+dependencies = [
+    "colorama ; sys_platform == 'win32'",
+    "confuse>=2.2.0",
+    "jellyfish",
+    "lap>=0.5.12",
+    "mediafile>=0.16.1",
+    "numpy>=2.0.2 ; python_version < '3.13'",
+    "numpy>=2.3.5 ; python_version >= '3.13'",
+    "packaging>=24.0",
```
semohr, line 34: General comment: I dont think we need the packaging dependency. We only use it for getting the current version. If I'm not missing anything a simple
```python
 from importlib.metadata import version 
```
Out of scope here but maybe of interest.

pyproject.toml
```diff
@@ -24,179 +22,142 @@ classifiers = [
...
+    "test/",
+]
+exclude = [
+    "docs/_build/",
+    "docs/modd.conf",
+    "docs/**/*.css",
+]
+
+[dependency-groups]
+test = [
```
semohr, line 54: Can we add a dev target? I tend to do something like this to easy the inital setup:

```
dev = [
    { include-group = "lint" },
    { include-group = "test" },
    { include-group = "docs" },
    { include-group = "typed" },
    "pre-commit>=4.5.1",
    "poethepoet>=0.26"
]
```

pyproject.toml
```diff
@@ -24,179 +22,142 @@ classifiers = [
...
+    "types-requests",
+    "types-urllib3",
+]
+release = [
+    "click>=8.1.7",
+    "tomli>=2.0.1",
+]
 
 [tool.pipx-install]
 poethepoet = ">=0.26"
```
semohr: This can go into the dev dependencies (see above). I dont think we should mix uv and pipx if not absolutely required.

semohr (changes requested):
Thanks for taking on the work of migrating! Highly appreciated.

.github/workflows/ci.yaml
snejus: We use `poe`

.github/workflows/integration_test.yaml
snejus: I see. As far as I understand `uv sync --locked` has the same behaviour as `poetry install` has by default, exiting if the lockfile has not been updated?

snejus: It's a shame that's not the default behaviour with `uv` 🙃

CONTRIBUTING.rst
snejus: Introduction of `uv` does not change this. Both `uv` and `poethepoet` should be installed using `pipx` globally, outside of `beets` virtual environment.

pyproject.toml
snejus: We use it in `deprecation.py` and `release.py` - yeah, this can be removed but as you said out of scope of this PR.

pyproject.toml
snejus: Does the `dev` group get picked up automatically for `sync`?

pyproject.toml
snejus: As I said above `poethepoet` should be installed globally by `pipx` outside `beets` venv - previously we had `poetry` here as well as it should also be installed outside.

Now, `uv` is installed by `setup-uv` action, so we don't need to provide it here any more.

CONTRIBUTING.rst
semohr: Why tho? `uv` is just a binary you drop somewhere. `uv` has no dependency on python, thus installing `uv` inside a pipx environment does not really do anything.

`uv` creates an environment for you, including poe if registered as dev dep.

I generally think pipx is not needed at all for the dev setup if we switch to uv.

pyproject.toml
semohr: Yes I think `dev` is the only dependency group that is picked automatically on sync. You can exclude it tho using `uv sync --no-dev`

We can customize this to our liking tho

https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups

pyproject.toml
semohr: > `As I said above poethepoet should be installed globally by pipx outside beets venv`

Why? Are there any dependency conflicts I'm missing?

My suggestion seems to be a supported way to install poethepoet using uv
https://poethepoet.natn.io/installation.html#with-uv

CONTRIBUTING.rst
snejus: I use _global_ `uv` and `poethepoet` binaries installed outside any specific project environment primarily because I maintain multiple projects.

Note we recommend using `pipx` to install these tools because they may not be available through the package manager of whatever distribution people use (for example, I doubt that `poethepoet` can be installed using `apt`).

semohr: I get how you currently manage your dev envs now :+1: . Still, sorry to press on here.

Essentially we should not care how people install their `uv` (via pipx or just as a binary)`, right? We can give a recommendation ofc.

I think my misunderstanding boils down to that I don't really understand why we wont want to move `poethepoet` into the dev dependencies and have it be part of our dev environment (managed by `uv`).  
What are the realistic reasons of having it externally managed? I don't really see any that's why I'm pressing here. 

Moving `poethepoet` into our dev env will completely remove our pipx dependency in the ci and for the dev setup. It also lets us create a truly reproducible dev environment as we control the poe version. It should also eases the initial setup burden as people dont have to install another tool (just run uv sync and you are ready). 

I for myself am still using `conda` for beets  and would like to drop it in favor of just a singular uv venv if possible. One can install `poe` using `uv tool install poethepoet` too but than we wont have the guarantee that e.g. the CI poe version is the same version as my local poe version.

gbcox:
As the Fedora packager for beets, the switch from Poetry to standard PEP 621 metadata / hatchling looks generally fine from a downstream packaging perspective.

The only thing I’d ask is that if workspace-style restructuring is considered later, please keep downstream/distribution builds in mind. A single normal sdist/wheel is straightforward for Fedora; multiple workspace members, editable local dependencies, or build/test assumptions that require uv sync can make isolated RPM builds much harder.

pyproject.toml
snejus: See the recommended way to install `poethepoet`: https://poethepoet.natn.io/installation.html#install-the-cli-globally-recommended

pyproject.toml
snejus: Added it, and made `docs` a dependency group instead of an extra.

pyproject.toml
semohr: We can remove this. Was only needed for the `https://github.com/BrandonLWhite/pipx-install-action`

pyproject.toml
```diff
@@ -212,6 +212,10 @@ cmd = "ruff format --config=pyproject.toml"
 help = "Format the documentation"
 cmd = "docstrfmt --preserve-adornments docs *.rst"
 
+[tool.poe.tasks.install]
```
semohr, line 212: Huh, wondrous.

docs/guides/installation.rst
```diff
@@ -1,26 +1,26 @@
...
-conflicts and keeps your system clean.
+We recommend installing with ``uv tool`` as it isolates beets and its
+dependencies from your system Python and other Python packages. This helps avoid
+dependency conflicts and keeps your system clean.
 
 .. <!-- start-quick-install -->
 
 .. tab-set::
 
-    .. tab-item:: pipx
```
semohr, line 19: Can we keep pipx installation instructions here? I would just add another tab-item for the uv tool based setup. Ultimately it does not matter, we should give users the choice.

I was just thinking we dont need pipx anymore for our dev setup and in the ci :)

CONTRIBUTING.rst
```diff
@@ -65,79 +65,52 @@ Development Tools
 
 In order to develop beets, you will need a few tools installed:
 
-- poetry_ for packaging, virtual environment and dependency management
+- uv_ for packaging, virtual environment and dependency management
 - poethepoet_ to run tasks, such as linting, formatting, testing
 
-Python community recommends using pipx_ to install stand-alone command-line
```
semohr, line 71: Nice!

semohr (changes requested):
Last short round of reviews, after that we should be good to go from my side 👍

pyproject.toml
snejus: It's been removed in the last commit!

pyproject.toml
snejus: That's to ensure we use this flag across all installs in the CI. Also makes it easier to do this as a maintainer, I think.

docs/guides/installation.rst
snejus: I definitely want to have a consistency here for both! But yeah, it makes sense to keep `pipx` instructions in the docs. :)

snejus: Hm, given that the rest of installation instructions use either one of these tools, I think it's best to keep just one of them, so I'd vote for `uv tool` instead of `pipx`.

semohr: You decide here I don't mind either too much. It is a nice to have in my opinion as it simplifies the setup for users as they can just copy paste the command for their appropriate setup.

snejus: I've included them now. I guess it's fine to have the following section describe how to use `uv tool` to manage dependencies.

image

semohr: Seems reasonable to me :+1: We can have another iteration here and make it more generic in the future if needed.

semohr (approved):

fredrik:
I am fully behind this change 👌

PR #6526, opened 2026-04-11
Remux WAVE_FORMAT_MPEGLAYER3 WAV files to MP3 during import

elainec2024:
## Description
Fixes #6455 

WAV files can contain MP3 audio streams (WAVE_FORMAT_MPEGLAYER3, wFormatTag=0x0055), but beets has no way to tag them correctly and would silently import them with wrong  duration, bitrate, and format metadata.

This PR fixes the issue by automatically detecting such files during import and using  ffmpeg to remux them into proper .mp3 files. The original WAV is deleted after successful remuxing, and the MP3 is imported normally with correct metadata.

This fix depends on beetbox/mediafile#105 which raises a clear FileTypeError for  these files, allowing beets to detect and handle them.

## To Do
- [x] ~Documentation~
- [x] Changelog
- [x] Tests

PR #6544, opened 2026-04-17
Call remux_mpeglayer3_wav proactively in read_item

elainec2024:
## Description

Follow-up to #6526. This change moves the `remux_mpeglayer3_wav` call to  before `library.Item.from_path` in `read_item`, so beets no longer depends on mediafile raising `FileTypeError` for MPEGLAYER3 WAV files. This is related to beetbox/mediafile#107.

## To Do
- [x] ~Documentation~
- [x] ~Changelog~
- [x] ~Tests~

### DIFF
--- a/beets/importer/tasks.py
+++ b/beets/importer/tasks.py
@@ -1083,27 +1083,21 @@ class ImportTaskFactory:
         if os.path.isfile(path):
             path = extension.fix_extension(path, logger=log)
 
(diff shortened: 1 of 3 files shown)


snejus (changes requested):
Now just update `mediafile` to `>= 0.16.2` (Note **>=** instead of **^**)

snejus (approved):

PR #6547, opened 2026-04-18
Remove `lib` parameter from `_raw_main` and `_setup`

semohr:
## Description

The `lib` parameter on `_raw_main` and `_setup` was only used by the test harness. Rather than keep testing-specific logic in production code, this PR removes the parameter and updates the test harness to patch `_open_library` instead.

- [x] ~~Changelog.~~ (Not needed as this is a purely internal factor imo)

### DIFF
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -126,7 +126,7 @@ NEEDS_REFLINK = unittest.skipUnless(
 
 
 class RunMixin:
-    lib: Library | None = None
+    lib: Library
 
     def run_command(self, *args, lib: Library | None = None):
         """Run a beets command with an arbitrary amount of arguments. The
@@ -135,7 +135,6 @@ class RunMixin:
         """
         sys.argv = ["beet", *args]  # avoid leakage from test suite args
         lib = lib or self.lib
-        assert lib is not None, "Lib required"
 
         with (
             patch.object(lib, "_close", Mock()),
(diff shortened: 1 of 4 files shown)


beets/test/helper.py
```diff
@@ -126,17 +126,20 @@ def config(self) -> beets.IncludeLazyConfig:
...
         the keyword argument `lib`.
         """
-        sys.argv = ["beet"]  # avoid leakage from test suite args
-        lib = None
-        if hasattr(self, "lib"):
-            lib = self.lib
-        lib = kwargs.get("lib", lib)
-        beets.ui._raw_main(list(args), lib)
+        sys.argv = ["beet", *args]  # avoid leakage from test suite args
+        lib = lib or self.lib
```
snejus, line 137: Is `self.lib` ever defined anywhere?

semohr: Yeah the TestHelper defines it. Which is used everywhere, I think

snejus: Ah, I see what you mean. Presumably, the type definition a couple of lines above should not default to None, then? Making the assert statement redundant.

semohr: Just removed the default. Not sure if it breaks inheritance somwhere. Lets see if the ci finishes

snejus: Seems to be happy

snejus (approved):

PR #6550, opened 2026-04-19
Centralise field type defs

snejus:
### Summary

This PR eliminates duplicated field-to-type mappings across `Album` and `Item` model classes. Previously, each model class maintained its own `_fields: dict[str, types.Type]` — a verbose, hand-written dictionary that encoded both _which_ fields exist and _what type_ each one has. This duplication made it easy for the two definitions to drift out of sync.

---

### Architecture Change

**Before**: Each model class owned a full inline `_fields` dict mapping field names to type instances.

```
Album._fields = {"id": types.PRIMARY_ID, "artpath": types.NullPathType(), "added": types.DATE, ...}  # ~40 entries
Item._fields  = {"id": types.PRIMARY_ID, "path": types.PathType(), "album_id": types.FOREIGN_ID, ...}  # ~80 entries
```

**After**: Field names are declared as a `set[str]`, and the type mapping is derived lazily from a single centralised `TYPE_BY_FIELD` registry in `beets.dbcore.fields`.

```
Album._field_names = {"id", "artpath", "added", ...}        # plain set of names
Item._field_names  = Album._field_names - {"artpath"} | {...} # extends Album's set

LibModel._fields (cached_classproperty) = {f: TYPE_BY_FIELD[f] for f in cls._field_names}
```

This creates a clean separation of concerns:
- **What fields belong to a model** → declared in the model (`_field_names`)
- **What type a field has** → declared once in `dbcore.fields.TYPE_BY_FIELD`

---

### Key Impacts

| Area | Before | After |
|---|---|---|
| Type definitions | Inline per model | Single registry in `dbcore.fields` |
| `Item._fields` | Fully independent dict | Derived from `Album._field_names` |
| `Album.item_keys` | Hard-coded list of 40+ strings | `_field_names - {"artpath", "id"}` |
| `_media_fields` / `_media_tag_fields` | Intersected with `_fields.keys()` | Intersected with `_field_names` directly |
| Coverage measurement | `--cov=beets --cov=beetsplug` | `--cov=.` (fixes missing `beetsplug/musicbrainz.py`) |

---

### Risk Areas to Review

- **`TYPE_BY_FIELD` completeness**: all field names in `_field_names` must have a corresponding entry in the registry — a missing key will raise at class definition time via `cached_classproperty`.
- **`Album.item_keys` semantics**: changed from `list` to `set` and is now derived automatically. Any code that relied on ordering or specific list membership should be checked.
- **`Item._field_names` inheritance expression**: `Album._field_names - {"artpath"} | {...}` — operator precedence here is `(Album._field_names - {"artpath"}) | {...}`, which is the intended behaviour, but worth a close read.


beets/library/models.py
```diff
@@ -246,107 +250,65 @@ class Album(LibModel):
...
         "releasegroupdisambig",
         "release_group_title",
         "rg_album_gain",
         "rg_album_peak",
-        "r128_album_gain",
-        "original_year",
-        "original_month",
-        "original_day",
-    ]
+        "script",
```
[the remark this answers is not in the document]
snejus: What are you on about, it didn't have it

beets/library/fields.py
```diff
@@ -0,0 +1,102 @@
+from . import types
+
+TYPE_BY_FIELD: dict[str, types.Type] = {
```
semohr, line 3: I think this file should not be in dbcore. The core database abstractions should know nothing about the specific names of the columns/fields imo.

semohr (changes requested):

beets/library/fields.py
snejus: True! This commit existed in my fork well before `library` received its own namespace :)

semohr (approved):

PR #6551, opened 2026-04-19
Remove archive on import during move mode

### LINKED ISSUE #1663: Option for moving original archive (zip, rar) to trash after successful import
... and i don't mean the temp folder after extracting ;)
The actual archive.

Further thoughts
- Delete instead of trash?
- Not only archives but also ~"source folders"?

snystrom:
## Description

Fixes #1663.

This modifies the `cleanup` method on the `ArchiveImportTask` to remove the original archive file on successful import when `move: yes`.

I decided to do this such that partial imports do not destructively remove the archive. Perhaps this would be better as a configuration option, but I think that may go beyond the scope for this initial feature.

I am unsure whether my method for "complete import" is strictly correct, I reasoned that we would check that all files had been moved out, but surely this data exists more concretely somewhere. Let me know if this should be polished, or there's more nuance I'm not seeing!

Additionally, I added unit tests for this behavior under the *old* test framework since existing helpers are present for the archive backend setup. This way we minimize the diff and maximize probability that my test is behaving correctly. I will gladly port this entire archive test section to pytest in a future PR if this is a dealbreaker.

## To Do


beets/importer/tasks.py
```diff
@@ -804,18 +804,26 @@ class ArchiveImportTask(SentinelImportTask):
     """An import task that represents the processing of an archive.
 
     `toppath` must be a `zip`, `tar`, or `rar` archive. Archive tasks
-    serve two purposes:
+    serve three purposes:
```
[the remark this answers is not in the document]
snystrom: grug have point

docs/changelog.rst
```diff
@@ -34,6 +34,11 @@ New features
 - :doc:`plugins/lyrics`: Add ``keep_synced`` config option and ``--keep-synced``
   CLI flag to skip re-fetching lyrics for tracks that already have synced
   lyrics, even when ``force`` is enabled. :bug:`5249`
+- :ref:`import-cmd`: When importing an archive (zip, tar, rar, or 7z) with
+  ``move: yes``, the source archive is now removed after a successful import.
+  Archives are preserved if any file in the archive was not imported (e.g.
+  skipped as a duplicate, or the import was aborted), and in non-move import
+  modes.
```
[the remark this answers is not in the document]
snystrom: docs updated

snystrom:
Not really sure it's worth the squeeze to add a test for the uncovered lines, really just a fallback for if we haven't extracted by the time we hit `cleanup`.... should that ever happen (if not, should I just error?)

beets/importer/tasks.py
```diff
@@ -862,13 +870,36 @@ def handlers(cls) -> list[ArchiveHandler]:
         return _handlers
 
     def cleanup(self, copy=False, delete=False, move=False):
-        """Removes the temporary directory the archive was extracted to."""
-        if self.extracted and self.toppath:
+        """Remove the temporary extraction directory. In ``move`` mode,
+        if the extraction directory is empty after the pipeline has run
```
snejus: Try to follow PEP 257 for docstrings: use the first line as the summary, then one empty line, and follow up with a more detailed description.

snejus (changes requested):
Looks good - just a small comment regarding docstring conventions.

beets/importer/tasks.py
snystrom: Polished it up a bit. Didn't document the kwargs, since no other `cleanup` method does.

beets/importer/tasks.py
```diff
@@ -870,10 +870,12 @@ def handlers(cls) -> list[ArchiveHandler]:
         return _handlers
 
     def cleanup(self, copy=False, delete=False, move=False):
-        """Remove the temporary extraction directory. In ``move`` mode,
-        if the extraction directory is empty after the pipeline has run
-        (i.e. every file in the archive was successfully imported)
-        also remove the source archive. Archives are preserved on
+        """Remove the temporary extraction directory and the archive if in
+        ``move`` mode.
```
snejus: It should fit in a single line 😅 

```suggestion
        """Remove the temporary extraction directory and the archive.
```

snystrom: I think we're interpreting PEP257 differently. In my opinion this makes the docstring less clear, but if you prefer it, happy to comply here.

beets/importer/tasks.py
snejus: Nice, thanks!

snejus (approved):
Now just move the changelog note under the **Unreleased** section since we pushed a new release earlier today.

snystrom:
Hah, nice. Here ya go!

snystrom:
oh gah my bad, messed it up

docs/changelog.rst
```diff
@@ -12,6 +12,11 @@ Unreleased
 ..
     New features
```
snejus: You want to uncomment this section here :)

snystrom: lol yup. When I run `lint-docs` or `format-docs` locally for some reason I get errors, or a huge diff, so just did this by hand without running them, hope this passes.

snystrom: ok merge conflict again, think this solves it....

Issue #6358, opened 2026-02-08
Primary aliases being ignored in MusicBrainz?


djl:
### Problem

Last week I ran an `mbsync` over my library and noticed a ton of artists no longer use their locale-specific aliases.

Two example artists and their aliases:

- [東京酒吐座](https://musicbrainz.org/artist/a26f7ca4-d14c-408b-9f9c-c8d847286bd0/aliases) which should be "Tokyo Shoegazer"
- [دمار](https://musicbrainz.org/artist/f64892d3-b4cf-4fae-9141-e39c5d968ff0/aliases) which should be "Damaar"

No config changes made on my side. I suspect this might be a side-effect of dropping the old `musicbrainzngs` library for fetching data but I've not had a chance to fully debug this yet, so pure speculation on my part.

### Setup

* OS: Linux
* Python version: 3.14.2
* beets version: latest git
* Turning off plugins made problem go away (yes/no): no

Relevant part of my configuration:

```yaml
import:
  copy: no
  move: yes
  timid: yes
  write: yes
  from_scratch: yes
  languages:
    - en
  ignored_alias_types:
    - 'legal name'
```

eramdam:
I'm seeing the same thing but I'm not aware of whether this was working properly before or not.

Morikko:
Do you mean specifically on `albumartists_credit` / `artists_credit` or also on `artists` / `albumartists` ?

I wanted to `mbsync` some songs and I realized that  `albumartists_credit` / `artists_credit` would be un-aliased while `artists` / `albumartists` kept their alias.

djl:
@Morikko All of the fields mentioned are affected.

```shell
$ beet ls -a -f '$albumartist // $albumartist_credit // $albumartists // $albumartists_credit' دمار
دمار // دمار // دمار // دمار
$ beet ls -a -f '$albumartist // $albumartist_credit // $albumartists // $albumartists_credit' 東京酒吐座
東京酒吐座 // 東京酒吐座 // 東京酒吐座 // 東京酒吐座
東京酒吐座 // 東京酒吐座 // 東京酒吐座 // 東京酒吐座
東京酒吐座 // 東京酒吐座 // 東京酒吐座 // 東京酒吐座
東京酒吐座 // 東京酒吐座 // 東京酒吐座 // 東京酒吐座
```

djl:
Something I've just noticed is that primary aliases are not ignored for releases.

Morikko:
Do you have ?

```
[...]

PR #6552, opened 2026-04-19
Use aliases for artist credit

Morikko:
## Description

The `*_credit` fields for album and track are also aliased.

Follow up of https://github.com/beetbox/beets/pull/6231. I wanted to `mbsync` my library but then I realized that `*_credit` fields were reverted to non-alias if already aliased.

I don't how I got those aliased in the first place as the code is not using them since years. Also, I had 2 albums added the same day and one was with aliased credit and the other one not.

Discussed here: https://github.com/beetbox/beets/issues/6358#issuecomment-4274747610

## To Do

- [X] Documentation (nothing to add since it is already said that artist is aliased)
- [X] Changelog.
- [X] Tests. (Somehow artist aliased assertion were missing so I add them as well)

### DIFF
--- a/beetsplug/musicbrainz.py
+++ b/beetsplug/musicbrainz.py
@@ -314,9 +314,8 @@ class MusicBrainzPlugin(
         - Parallel lists that keep the per-artist granularity for callers that
           need to reason about individual credited artists.
 
-        When available, a preferred alias is used for the canonical artist name
-        and sort name, while the credit name preserves the exact credited text
-        from the release.
+        When available, a preferred alias is used for the canonical artist name,
+        sort name and the credit name.
         """
         artist_parts: list[str] = []
         artist_sort_parts: list[str] = []
--- a/beetsplug/musicbrainz.py
+++ b/beetsplug/musicbrainz.py
@@ -330,12 +330,17 @@ class MusicBrainzPlugin(
             artists_ids.append(el["artist"]["id"])
             alias = _preferred_alias(el["artist"].get("aliases", []))
             artist_object = alias or el["artist"]
+            credit_artist_object = alias or el
 
             joinphrase = el["joinphrase"]
             for name, parts, multi in (
                 (artist_object["name"], artist_parts, artists),
                 (artist_object["sort_name"], artist_sort_parts, artists_sort),
-                (el["name"], artist_credit_parts, artists_credit),
+                (
+                    credit_artist_object["name"],
+                    artist_credit_parts,
+                    artists_credit,
+                ),
             ):
                 parts.extend([name, joinphrase])
                 multi.append(name)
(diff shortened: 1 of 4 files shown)


test/plugins/test_musicbrainz.py
```diff
@@ -653,32 +653,75 @@ def test_genres(self, mb, expected_genres):
         assert mb.album_info(release_factory()).genres == expected_genres
 
     @pytest.mark.parametrize(
-        "languages_config, expected_titles",
+        "languages_config, is_aliased",
         [
-            _p([], ("Album", "Release Group", "Recording"), id="no aliases"),
+            _p(
+                [],
+                False,
```
snejus: Try to not use conditional logic in tests.

Given how many fields are aliased now, I'd suggest to rename the test to `test_parse_aliased_titles`, remove the `parametrize` and simply test the aliases titles.

beetsplug/musicbrainz.py
```diff
@@ -330,12 +330,17 @@ def _parse_artist_credits(artist_credits: list[ArtistCredit]) -> ArtistInfo:
             artists_ids.append(el["artist"]["id"])
             alias = _preferred_alias(el["artist"].get("aliases", []))
             artist_object = alias or el["artist"]
+            credit_artist_object = alias or el
```
snejus, line 332: The docstring needs to be updated accordingly.

snejus (changes requested):

test/plugins/test_musicbrainz.py
Morikko: Do you mean you want to remove the `no aliases` case ?

snejus: indeed!

snejus (approved):
Nice, thanks!

amogus07:
Aren't the `*_credit` fields specifically intended to store the canonical way artists are credited for a specific entity?

snejus:
Good question. My understanding is that `*_credit` fields are meant to preserve release/recording-specific credited text (MusicBrainz `artist_credit[].name`), while alias preferences are usually applied to `artist` and `artist_sort`.

This PR intentionally applied aliases to `*_credit` too, mainly to avoid value churn for users running mbsync with language aliases enabled.

Your concern is valid, but since we don't have a concrete regression from the current behavior, I suggest we keep this as-is for now and revisit if someone reports a real issue with it.

djl:
This change may have a large impact on people's libraries. I ran mbsync over my library and all `*_credit` fields have been changed to use aliases, which is not correct.

One example would be [Yasiin Bey](https://musicbrainz.org/artist/f5c4c27e-5902-4fc0-adfb-7775da3d5363). The older releases are now tagged with the current alias, not what they're actually credited as.

I suspect users may complain about this when this eventually reaches a release.

snejus:
> I suspect users may complain about this when this eventually reaches a release.

This has already been released in `2.10.0`. Let's see, if that's the case I'm happy to revert it!

djl:
I can see some users actually wanting this, but definitely not all (e.g. me!). It would make sense to make this an opt-in feature, due to it completely overwriting existing data en-masse.

snejus:
@djl would you be happy to submit a PR to make this an opt-in?

djl:
Can do. Should be up tomorrow.

PR #6557, opened 2026-04-20
fromfilename: Support 'track' prefix when parsing track number

snejus:
## `fromfilename`: Support `"track"` prefix in track number parsing

The `fromfilename` plugin's filename-matching regex is extended to recognise filenames like `track01.m4a` or `track 2.m4a`, where a literal `"track"` prefix precedes the track number.


aereaux:
This is a pretty small change but it looks good and would be useful for me!

aereaux (approved):

Issue #108, opened 2013-02-27
Path format if for unset attributes returns "None"


pscn:
Sorry I don't have time to provide more details or fix it myself at the moment.

I wanted to add the disctitle to the path if disctitle is set. I thought adding something like the following to the path would be enough:  /Disc$disc%if{$disctitle,($disctitle)}/
But I ended up with "/Disc01(None)/" for releases without a disctitle.  This happens for initial as well as re-imports.

PR #6566, opened 2026-04-21
lyrics: write synced LRC lyrics to SYLT ID3 frame

### LINKED ISSUE #6541: Use SYLT tag for synchronized lyrics
### Proposed solution

Currently the lyrics plugin has the option synced:

    synced: Prefer synced lyrics over plain lyrics if a source offers them. Currently lrclib is the only source that provides them.

The results are written into the USLT tag though. While this is correct for unsynchronized lyrics it's not for synchronized as the specification tells:

[id3.org/id3v2.3.0#sec4.10](https://id3.org/id3v2.3.0#sec4.10)

### Objective

Based on the discussion: https://github.com/beetbox/beets/discussions/6315

#### Goals

It would be great if the plugin would choose the correct tag/field (SYLT) depending on the synced modifier (or finally not the modifier itself but based on what it could fetch, since the modifier isn't enforcing but only requesting).

benstev1:
When the lyrics plugin writes LRC-formatted (timestamped) lyrics to an ID3-tagged audio file, store the timing data in the SYLT (synchronized lyrics) frame and plain text in USLT, rather than writing raw LRC text verbatim to USLT.

Fixes #6541

## Description

This PR:
- Adds a `Lyrics.sylt` property that converts `[mm:ss.cc]` LRC lines to `(text, milliseconds)` pairs for use with the ID3v2 `SYLT` frame.
- After writing via `item.try_write()`, opens the file with mutagen to replace `USLT` with clean plain text and write a proper `SYLT` frame.
- Clears any stale `SYLT` frame when lyrics become unsynced.
- Non-ID3 formats (FLAC, OGG, etc.) are silently skipped.

(...)

## To Do

- [X] Documentation.
- [X] Changelog.
- [X] Tests.

### DIFF
--- a/beets/util/lyrics.py
+++ b/beets/util/lyrics.py
@@ -117,6 +117,27 @@ class Lyrics:
         """Return per-line lyric text with timestamps removed."""
         return [ln for _, ln in self._split_lines]
 
+    @cached_property
+    def sylt(self) -> list[tuple[str, int]]:
+        """Return SYLT-format (text, milliseconds) pairs from LRC timestamps.
+
+        Converts each LRC-timestamped line into a ``(text, milliseconds)``
+        tuple as expected by the ID3v2 SYLT frame.  Lines without a timestamp
+        (e.g. blank separator lines) are omitted.
+        """
+        result: list[tuple[str, int]] = []
+        for ts, text in self._split_lines:
+            if not ts:
+                continue
+            ts_m = re.match(r"\[(\d{2}):(\d{2})\.(\d{2})\]", ts)
+            if ts_m:
+                minutes = int(ts_m[1])
+                seconds = int(ts_m[2])
+                centiseconds = int(ts_m[3])
+                ms = (minutes * 60 + seconds) * 1000 + centiseconds * 10
+                result.append((text, ms))
+        return result
+
     @property
     def synced(self) -> bool:
         """Return whether the lyrics contain synced timestamp markers."""
--- a/beets/util/lyrics.py
+++ b/beets/util/lyrics.py
@@ -28,6 +28,7 @@ class Lyrics:
     ORIGINAL_PAT = re.compile(r"[^\n]+ / ")
     TRANSLATION_PAT = re.compile(r" / [^\n]+")
     LINE_PARTS_PAT = re.compile(r"^(\[\d\d:\d\d\.\d\d\]|) *(.*)$")
+    LRC_TIMESTAMP_PAT = re.compile(r"\[(\d{2}):(\d{2})\.(\d{2})\]")
 
     text: str
     backend: str | None = None
@@ -129,12 +130,10 @@ class Lyrics:
         for ts, text in self._split_lines:
             if not ts:
                 continue
-            ts_m = re.match(r"\[(\d{2}):(\d{2})\.(\d{2})\]", ts)
+            ts_m = self.LRC_TIMESTAMP_PAT.match(ts)
             if ts_m:
-                minutes = int(ts_m[1])
-                seconds = int(ts_m[2])
-                centiseconds = int(ts_m[3])
-                ms = (minutes * 60 + seconds) * 1000 + centiseconds * 10
+                m, s, cs = map(int, ts_m.groups())
+                ms = (m * 60 + s) * 1000 + cs * 10
                 result.append((text, ms))
         return result
 
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -33,7 +33,7 @@ import requests
 from bs4 import BeautifulSoup
 from unidecode import unidecode
 
-from beets import logging, plugins, ui, util
+from beets import logging, plugins, ui
 from beets.autotag.distance import string_dist
 from beets.dbcore import types
 from beets.dbcore.query import FalseQuery
@@ -42,8 +42,6 @@ from beets.util.config import sanitize_choices
 from beets.util.lyrics import INSTRUMENTAL_LYRICS, Lyrics
 
 from ._utils.requests import HTTPNotFoundError, RequestHandler
-
-log = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator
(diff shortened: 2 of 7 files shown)


beetsplug/lyrics.py
```diff
@@ -949,6 +949,56 @@ def write(self, items: list[Item]) -> None:
...
+        tags["USLT::eng"] = USLT(
+            encoding=3, lang="eng", desc="", text=plain_text
+        )
+
+    # Write (or clear) the SYLT frame.
+    tags.delall("SYLT")
+    sylt_data = lyrics.sylt
+    if sylt_data:
+        tags["SYLT::eng"] = SYLT(
+            encoding=3, lang="eng", format=2, type=1, text=sylt_data
```
ShimmerGlass: do we have another option that hardcoding english for everything?
maybe an "unknown"/"any" lang code if we don't have the info?

beetsplug/lyrics.py
```diff
@@ -949,6 +949,56 @@ def write(self, items: list[Item]) -> None:
...
+        )
+
+    try:
+        f.save()
+    except Exception:
+        pass
```
ShimmerGlass: you should log the error here, completely ignoring it will make debugging harder

beetsplug/lyrics.py
```diff
@@ -1141,6 +1191,9 @@ def add_item_lyrics(self, item: Item, write: bool) -> None:
             item.store()
             if write:
                 item.try_write()
+                # For ID3 files, replace the raw LRC stored in USLT with plain
+                # text and write a proper SYLT frame (or clear a stale one).
+                _write_sylt_to_id3(item, Lyrics(lyrics_text))
```
ShimmerGlass: would it make sense to move this into beet's main tag-writing function?
* it feels a bit weird/wasteful to write the tags twice 
* what if lyrics are written from somwhere else?

ShimmerGlass (dismissed):

snejus:
Have you considered adding support for this tag to `mediafile` first?

benstev1:
Went ahead and implemented these. Heres a breakdown:

-- do we have another option than hardcoding english for everything?
fixed in the latest commit. Changed lang="eng" to lang="XXX" (ISO 639-2 for "undetermined") for both the USLT and SYLT frames. This is actually what the upstream mediafile library itself does for USLT

-- you should log the error here, completely ignoring it will make debugging harder
Also fixed, added a module-level log = logging.getLogger(__name__) using beets' own logging wrapper and changed the bare except: pass to except Exception as exc: log.warning(...).

-- would it make sense to move this into beet's main tag-writing function? / Have you considered adding support to mediafile first?
The mediafile library has no SYLT support at all,  no storage class, no field,  so there's nothing to hook into yet. The proper long-term fix would be a separate PR to the upstream mediafile repo adding a synced_lyrics = MediaField(MP3SYLTStorageStyle(...)) field, after which the lyrics plugin could simply set item.synced_lyrics = lyrics.sylt and let the normal write path handle it. Would be a larger undertaking but something that should be implemented in the future.

snejus:
> The mediafile library has no SYLT support at all, no storage class, no field, so there's nothing to hook into yet. The proper long-term fix would be a separate PR to the upstream mediafile repo adding a synced_lyrics = MediaField(MP3SYLTStorageStyle(...)) field, after which the lyrics plugin could simply set item.synced_lyrics = lyrics.sylt and let the normal write path handle it. Would be a larger undertaking but something that should be implemented in the future.

Ultimately, _writing tags_ is `mediafile`'s responsibility, and thus I'm not comfortable with `lyrics` plugin taking some of this responsibility here, especially given how complex the plugin is already.

Would you be happy to add support for `SYLT` to `mediafile`, even if it's a larger undertaking?

benstev1:
I'll give it a go!

snejus:
Thank you so much!

benstev1:
Ok I opened a PR in mediafile with the changes. Wasn't as difficult as it looked so I got it done pretty quickly. Its issue #108 in the mediafile repo if you'd like to take a look. If everything looks good there I'll update this PR to accommodate it and we can go from there!

benstev1:
Ok it looks like everything is working end to end. Both PR's are up to date. Let me know if there are any more changes that need to be made.

snejus:
Now that mediafile is released, update the requirement in `pyproject.toml` and run `uv tool run 'poetry<2' lock`.

Resolve whichever @ShimmerGlass's and Copilot's comments you've addressed, and I will have a final look then.

beetsplug/lyrics.py
```diff
@@ -1131,16 +1131,24 @@ def add_item_lyrics(self, item: Item, write: bool) -> None:
...
                     del item[item_key]
 
-            lyrics_text = new_lyrics.full_text
+            # For synced lyrics, store plain text in USLT and pass SYLT data
+            # separately so mediafile can write the SYLT frame.
+            if new_lyrics.synced:
+                lyrics_text = "\n".join(new_lyrics.text_lines)
+                sylt_data = new_lyrics.sylt
+            else:
+                lyrics_text = new_lyrics.full_text
```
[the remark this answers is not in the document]
snejus: @benstev1 note that we **want** to retain the timestamps in the lyrics text here. Especially so for non-mp3 files that do not support SYLT frames. Many music players support them without an additional frame.

poetry.lock
```diff
@@ -572,13 +572,13 @@ redis = ["redis (>=2.10.5)"]
 
 [[package]]
 name = "certifi"
-version = "2026.2.25"
+version = "2026.4.22"
```
snejus, line 586: I asked to run `uv tool run 'poetry<2' lock` but you seem to have updated all dependencies (using `update` command, I guess?). Do you mind undoing this commit and running `lock` instead, which is supposed to update `mediafile` only, to stay in scope of the PR?

snejus:
@benstev1 are you planning to come back to this by any chance?

benstev1:
Sorry about that yeah i'll come back and tackle this. My term ended for school and I forgot about it.

snejus:
No worries!

Regarding `poetry.lock`, I would suggest

```sh
git checkout master poetry.lock
uv tool run 'poetry<2' lock
```

benstev1:
All I did was run the command you posted and I'm failing a doc check? Its not something I've seen before its asking for a 'line_length' positional arg, do you know how to fix this?

snejus:
Just fixed `poetry.lock`. Will have a look at the docs failure soon

beets/util/lyrics.py
```diff
@@ -117,6 +117,27 @@ def text_lines(self) -> list[str]:
...
+
+        Converts each LRC-timestamped line into a ``(text, milliseconds)``
+        tuple as expected by the ID3v2 SYLT frame.  Lines without a timestamp
+        (e.g. blank separator lines) are omitted.
+        """
+        result: list[tuple[str, int]] = []
+        for ts, text in self._split_lines:
+            if not ts:
+                continue
+            ts_m = re.match(r"\[(\d{2}):(\d{2})\.(\d{2})\]", ts)
```
snejus: Compile this pattern as the class attribute so that it's reused.

beets/util/lyrics.py
```diff
@@ -117,6 +117,27 @@ def text_lines(self) -> list[str]:
...
+            if ts_m:
+                minutes = int(ts_m[1])
+                seconds = int(ts_m[2])
+                centiseconds = int(ts_m[3])
```
snejus: I think you can reduce this to

```suggestion
                m, s, cs = map(int, ts_m.groups())
```

snejus (changes requested):
Looks good, just a pair of small comments

snejus (approved):
Thanks!

skiinganchor:
@snejus are we getting a release soon? Would be great to test this one.

snejus:
Soon! Try installing the `master` branch before this is released :)

skiinganchor:
All looking great @snejus!

PR #6569, opened 2026-04-22
Remove 'slow_test' unittest marker

snejus:
## Remove `slow_test` unittest marker

Removes the `slow_test` decorator and its supporting `SKIP_SLOW_TESTS` env-var mechanism entirely. All previously "slow" test classes now run unconditionally.

**Changes:**
- Deletes `slow_test()` helper from `beets/test/_common.py`
- Removes all `@_common.slow_test()` usages across test files
- Updates `CONTRIBUTING.rst` to drop the `SKIP_SLOW_TESTS` documentation

Just doing some plumbing removing some unused noise from tests.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -315,13 +315,6 @@ Use ``poe`` to run tests:
 
     $ poe test [pytest options]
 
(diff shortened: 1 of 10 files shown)


semohr (approved):
Thanks!

PR #6585, opened 2026-04-25
Fix concurrent plugin execution

### LINKED ISSUE #6583: Discogs URLs no longer recognised by importer ID lookup (numeric IDs still work)
### Problem

Discogs release URLs are not recognised when passed via `-S` or entered interactively.

Example:

```sh
beet import -S "https://www.discogs.com/release/756816-..."
```

Led to this problem:

* URL is passed to MusicBrainz as an MBID
* MusicBrainz rejects it
* No candidates are returned
* Discogs is never consulted

Logs:
```Searching for album IDs: https://www.discogs.com/release/...
musicbrainz: Requesting MusicBrainz release https://www.discogs.com/release/...
musicbrainz: Invalid MBID (...)
Evaluating 0 candidates.```
```
Expected behavior:
* Discogs plugin should recognise the URL and return the corresponding release.

Additional observations:

Using the numeric Discogs ID does work:
```sh
beet import -S 756816
```

In this case, Discogs is correctly queried and returns a candidate
This occurs both via -S and via the interactive “Enter Id” prompt

### Setup

OS: Ubuntu (22.04-based)
Python version: 3.12.10
beets version: 2.8.0
Plugins enabled: discogs (along with standard plugins)

My configuration (relevant excerpt):
```yaml
plugins: [discogs, musicbrainz, fromfilename, inline, the]
```

snejus:
## Fix lambda closure bug in concurrent plugin method execution

Fixes a classic Python closure-over-loop-variable bug in `_yield_from_plugins`, where all `ThreadPoolExecutor` futures ended up calling the **last** plugin's method instead of their respective plugin.

### Root cause

The inline `lambda` captured `plugin` by reference. By the time the executor ran the futures, the loop had finished and all lambdas resolved `getattr(plugin, method_name)` to the final loop value.

### Fix

Extracts a named `materialize(method, *args, **kwargs)` helper. The plugin method is now **passed as an argument** to `executor.submit`, so each future holds its own reference — no closure over the loop variable.

### Testing

A regression test (`test_albums_for_ids_calls_each_plugin_once`) was added first to reproduce the bug, then `xfail` removed once fixed. The test uses a `DelayedStartExecutor` to force all submissions to complete before any worker starts, reliably triggering the original race.

Fixes: #6583

### DIFF
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -78,6 +78,12 @@ def _yield_from_plugins(
 ) -> Callable[..., Iterator[Ret]]:
     method_name = func.__name__
 
+    def materialize(
+        plugin: MetadataSourcePlugin, method_name: str, *args, **kwargs
+    ) -> list[Ret]:
+        method = getattr(plugin, method_name)
+        return list(method(*args, **kwargs))
+
     @wraps(func)
     def wrapper(*args, **kwargs) -> Iterator[Ret]:
         # Run plugin methods concurrently for faster I/O-bound lookups.
@@ -86,12 +92,9 @@ def _yield_from_plugins(
                 executor.submit(
                     # Evaluate iterator with list such that results are ready when
                     # future.result() is called.
-                    lambda *args, **kwargs: list(
-                        getattr(plugin, method_name)(
-                            *args,
-                            **kwargs,
-                        )
-                    ),
+                    materialize,
+                    plugin,
+                    method_name,
                     *args,
                     **kwargs,
                 ): plugin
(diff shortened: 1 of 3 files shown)


semohr:
Interesting! Thanks for having a look. I thought closures are captured on lambda construction in python too.

Im a bit confused how this worked locally for me at all. I tested it with 3 metadata plugins enabled and got results from all three 🤔

semohr (approved):

snejus:
> got results from all three

Me too, _for the initial **candidates**_ run. However, only got one result when I attempted to search by ID.

PR #6599, opened 2026-05-02
fix(MusicBrainz): date parsing fix

ShimmerGlass:
## Description

Correctly handle release dates where leading or intermediate components are missing, e.g. 2008-??-02

Without this fix parsing of these releases fail with:
```Error in 'MusicBrainz.albums_for_ids': invalid literal for int() with base 10: '??'```

While cases like that are rare in MusicBrainz they do exist and we should support them.
example: https://musicbrainz.org/release/18295b5f-3150-4086-9095-7497c261e6ce

## To Do

### DIFF
--- a/beetsplug/musicbrainz.py
+++ b/beetsplug/musicbrainz.py
@@ -198,7 +198,13 @@ def _get_date(date_str: str) -> tuple[int | None, int | None, int | None]:
     if not date_str:
         return None, None, None
 
-    parts = list(map(int, date_str.split("-")))
+    def _parse_part(part: str) -> int | None:
+        try:
+            return int(part)
+        except ValueError:
+            return None
+
+    parts = list(map(_parse_part, date_str.split("-")))
 
     return (
         parts[0] if len(parts) > 0 else None,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: beetsplug/musicbrainz.py [hunk 1]
```diff
@@ -190,6 +190,18 @@ def _preferred_release_event(release: Release) -> tuple[str | None, str | None]:
...
+def _parse_date_part(part: str) -> int | None:
+    # MusicBrainz uses "??" as a placeholder for unknown date components
+    # in missing leading or intermediate parts (e.g. "2008-??-02").
+    if not part or part == "??":
+        return None
+
+    try:
+        return int(part)
+    except ValueError:
+        return None
```
snejus: I think we only need this block here - no need to handle anything explicitly.

And if the function is just these 4 lines, we may as well just define it inline within `_get_date`?

ShimmerGlass: agreed, done!

snejus (approved):

PR #6606, opened 2026-05-03
listenbrainz: add option to import listens from ListenBrainz data export zip

Maxr1998:
## Description

Since importing listens from the API is rather slow and puts unnecessary load on the servers, importing a dump exported from https://listenbrainz.org/settings/export/ should also be possible. This change introduces the `export-file` parameter to the `lbimport` command, which takes all listens from the zipfile to update the play counts.

I also fixed an issue where `mbid_mapping` could be `None` if explicitly set to `null` in the JSON (which happens for these exports sometimes), and changed the recording mbid logic to prefer one explicitly supplied by the player in the `additional_data`, instead of relying on the (sometimes inaccurate) mapper.

Since the `--max` parameter isn't documented either, I'm unsure whether I should add the parameter to the plugin docs. In theory, the `--help` parameter already tells you enough.

I'm also not sure if the `--max` parameter should apply to exported files, especially since the order of zip file entries isn't deterministic.

## To Do

### DIFF
--- a/beetsplug/listenbrainz.py
+++ b/beetsplug/listenbrainz.py
@@ -236,7 +236,7 @@ class ListenBrainzPlugin(MusicBrainzAPIMixin, BeetsPlugin):
         for track in listens:
             if track["track_metadata"].get("release_name") is None:
                 continue
-            mbid_mapping = track["track_metadata"].get("mbid_mapping", {})
+            mbid_mapping = track["track_metadata"].get("mbid_mapping", {}) or {}
             mbid = mbid_mapping.get("recording_mbid")
             tracks.append(
                 {
--- a/beetsplug/listenbrainz.py
+++ b/beetsplug/listenbrainz.py
@@ -21,6 +21,8 @@ from ._utils.playcount import update_play_counts
 from ._utils.requests import TimeoutAndRetrySession
 
 if TYPE_CHECKING:
+    from pathlib import Path
+
     from ._utils.playcount import Track
 
 
@@ -164,7 +166,7 @@ class ListenBrainzPlugin(MusicBrainzAPIMixin, BeetsPlugin):
             self._log.debug("Invalid Search Error: {}", e)
             return None
 
-    def import_listenbrainz_data_export(self, export_file: str):
+    def import_listenbrainz_data_export(self, export_file: str | Path):
         """Import ListenBrainz data from a .zip file."""
         export_file = syspath(normpath(export_file))
 
--- a/beetsplug/listenbrainz.py
+++ b/beetsplug/listenbrainz.py
@@ -26,6 +26,27 @@ if TYPE_CHECKING:
     from ._utils.playcount import Track
 
 
+class Listen(TypedDict):
+    listened_at: int
+    track_metadata: TrackMetadata
+
+
+class TrackMetadata(TypedDict, total=False):
+    artist_name: str
+    track_name: str
+    release_name: str | None
+    additional_info: AdditionalInfo
+    mbid_mapping: MbidMapping | None
+
+
+class AdditionalInfo(TypedDict, total=False):
+    recording_mbid: str | None
+
+
+class MbidMapping(TypedDict, total=False):
+    recording_mbid: str | None
+
+
 class ListenBrainzPlugin(MusicBrainzAPIMixin, BeetsPlugin):
     """A Beets plugin for interacting with ListenBrainz."""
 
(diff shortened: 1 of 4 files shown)


Serene-Arc:
Thanks for the PR! I'm of the opinion that more documentation is always better. If there's an option that isn't currently documented, then please do add documentation for it, either in this PR or another one.

As for the `--max` option I agree that it shouldn't apply to imported files. From the code, it seems to be an API-limiting option which doesn't really apply to when you're importing offline data. Further, I would expect someone who has downloaded that data to either filter the new dataset themselves or else want to import the whole thing, especially if the process isn't deterministic.

grug is correct that this interaction should be noted in the documentation though.

Maxr1998:
@Serene-Arc thanks, I made some further adjustments, added some documentation, tests, and a changelog entry. Should be ready to review now.

beetsplug/listenbrainz.py
```diff
@@ -44,23 +47,47 @@ def commands(self):
...
+            "(instead of fetching from the API)",
+        )
         lbupdate_cmd.parser.add_option(
             "--max",
             dest="max_listens",
             type="int",
             default=None,
-            help="maximum number of listens to fetch (default: all)",
+            help="maximum number of listens to fetch via the API (default: all). "
+            "This option does not apply when importing a file via -f/--export-file.",
```
snejus:
```suggestion
            help=(
                "maximum number of listens to fetch via the API (default: all)."
                " This option does not apply when importing a file via"
                " -f/--export-file."
            ),

```

beetsplug/listenbrainz.py
```diff
@@ -44,23 +47,47 @@ def commands(self):
...
         )
 
         def func(lib, opts, args):
-            self._lbupdate(lib, self._log, max_listens=opts.max_listens)
+            self._lbupdate(
+                lib,
+                self._log,
```
snejus: There's no need to pass `self._log` here - just use it directly within those functions.

snejus (changes requested):

beetsplug/listenbrainz.py
```diff
@@ -129,8 +156,42 @@ def _make_request(self, url, params=None):
             self._log.debug("Invalid Search Error: {}", e)
             return None
 
+    def import_listenbrainz_data_export(self, log, export_file):
```
snejus: Also add types for new functions that you introduce!

Maxr1998:
Made the requested changes in 1e089eccf1622ddd942ee62d2b3511a992793ba6.

snejus:
@Maxr1998 See test failures.

beetsplug/listenbrainz.py
```diff
@@ -126,8 +166,42 @@ def _make_request(self, url, params=None):
             self._log.debug("Invalid Search Error: {}", e)
             return None
 
+    def import_listenbrainz_data_export(self, export_file: str | Path):
```
snejus: Just add a return type here for clarity as well please!

Maxr1998: Done. Let me know if you want me to squash the commits before you merge.

snejus: No worries about that. Would you mind defining a specific `TypedDict` type for those listens?

Maxr1998: Done. I hope all the checks pass cleanly. I can also extract the types to a separate file like the Deezer plugin does, but that's definitely a separate PR.

I also feel like the `recording_mbid` behavior change should be a separate PR. Please tell me if you agree, then I will extract that part first. (Alternatively, I could reorder and squash my commits so that they stack cleanly when you do a normal merge commit or rebase merge.)

beetsplug/listenbrainz.py
```diff
@@ -416,3 +420,24 @@ def get_weekly_playlist(self, playlist_type, most_recent=True):
...
+    additional_info: AdditionalInfo
+    mbid_mapping: MbidMapping | None
+
+
+class AdditionalInfo(TypedDict, total=False):
+    recording_mbid: str | None
+
+
+class MbidMapping(TypedDict, total=False):
+    recording_mbid: str | None
```
snejus: Thanks! Just define them at the top of the module please!

Maxr1998: Done. Did you see my other comment too?

beetsplug/listenbrainz.py
snejus: Don't worry about it - it's all good :)

beetsplug/listenbrainz.py
snejus: I did. I think it's ready to be merged now.

snejus (approved):
Thanks!

beetsplug/listenbrainz.py
Maxr1998: Alright! Thanks for the review!

PR #6615, opened 2026-05-07
feat(import): add --nomove / -M option

ShimmerGlass:
## Description

Add `--nomove` / `-M` option to override the `move: yes` config option during import.

This option is especially useful for reimporting when using players / music servers like Navidrome that easily loose track of files when they change path and tags at the same time. 
Now one can retag their files without moving them `beet import -M my-files/`, wait for a rescan, and then `beet move`. 

original idea by @snejus 

## To Do

### DIFF
--- a/beets/ui/commands/import_/__init__.py
+++ b/beets/ui/commands/import_/__init__.py
@@ -182,6 +182,13 @@ import_cmd.parser.add_option(
     help="move tracks into the library (overrides -c)",
 )
 import_cmd.parser.add_option(
+    "-M",
+    "--nomove",
+    action="store_false",
+    dest="move",
+    help="don't move tracks into the library (overrides -m)",
+)
+import_cmd.parser.add_option(
     "-w",
     "--write",
     action="store_true",
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,6 +14,8 @@ New features
 
 - :doc:`plugins/convert`: The ``--force`` and ``--keep-new`` CLI flags are now
   also available as config options via ``force`` and ``keep_new``.
+- :ref:`import-cmd`: The ``--nomove`` / ``-M`` CLI flag can now be used to
+  override the ``move: yes`` config option during import.
 
 ..
     Bug fixes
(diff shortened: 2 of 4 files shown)


semohr:
Please have a look at the getting started guide too and adjust it when applicable :upside_down_face:

ShimmerGlass:
@semohr the getting started guide does not mention these more advanced options. I did however update the auto-tagger guide :)

semohr:
> @semohr the getting started guide does not mention these more advanced options. I did however update the auto-tagger guide :)

Aren't they indirectly mentioned in the `Choose your import style (pick one):` section? I assumed these config options mapped to the same cli arguments :thinking:

ShimmerGlass:
This PR does not change how the config works but allows overriding the `move` option to `no` via the CLI. 
`Choose your import style` is about the config, not CLI options that override it (-c, -C, -m are not mentioned). They are more "advanced" and should IMHO be left to more in-depth pages of the docs.
WDYT ?

semohr:
Well, I know this PR only adds the CLI arg, but the new argument currently overrides the move option. That’s actually what made me check the related docs in the first place.

I realize this is only loosely related to the PR itself, but I wanted to quickly get your opinion on the getting started guide because it seems a bit inconsistent to me.

image

The example here is missing the `move: no` option. I was mainly wondering what you think about that, and whether you could also take another look at the guide since you already just investigated the nomove option.

ShimmerGlass:
> The example here is missing the move: no option

not technically, `move` is `no` by default.

I'm usually on the side of being explicit, but to be truly explicit here would mean having this, which is perhaps a bit too much.
```
import:
    copy: no 
    move: no
    link: no
    hardlink: no
```

I don't think the docs are wrong, but they do prioritize conciseness over clarity. They also leave out the `{,hard,ref}link` options, but this may be on purpose are these are more advanced.

---

The way import behavior is configured is a bit unfortunate IMHO: we have a bunch of mutually exclusive options: `copy`, `move`, `link`, `hardlink`. Then we have a bunch of priorities on which of these options override which: https://github.com/beetbox/beets/blob/master/beets/importer/session.py#L125 
This in turn leads to unclear behavior, especially when one doesn't know the defaults by heart. 

In a different world we could perhaps have something like:
```
import:
    mode: copy / move / ...
```
and `--mode copy` instead of all these on/off flags. 
Perhaps this is something we could look into (while keeping backwards compatibility ofc).

snejus (approved):
Looks great. This is consistent with current `beet import` options and adds of the missing pieces! (I've added in my fork a while ago)

image

Issue #6553, opened 2026-04-19
Beets logging overhaul


JOJ0:
**Goal**: Replace current messy logging setup with flexible, config-driven system using Python's `logging.config.dictConfig()` or similar. Also fix a number of issues with the current logging setup.

## Idea

Create a **comprehensive, config-driven logging system** that replaces our current messy setup. My idea would be to inline the logging config into beets' main `config.yaml`, add predefined modes like `minimal` (current), `verbose` (debug+timestamps+file refs for bug reports), and support fully custom user-defined configurations via Python's dictConfig schema.

This could for example look like this:

```yaml
# --------------- Logging --------------

verbose: 0
logging:
  mode: minimal

  modes:
    minimal:
      ...
    debug:
      version: 1
      formatters:
        default:
          format: "%(asctime)s | %(levelname)-8s | %(process)d:%(threadName)s | %(name)s:%(lineno)d | %(message)s"
      handlers:
        stdout:
          class: logging.StreamHandler
          formatter: default
      loggers:
        beets:
          level: DEBUG
          handlers: [stdout]
          propagate: false
        beetsplug:
          level: DEBUG
          handlers: [stdout]
          propagate: false
      root:
        level: DEBUG
        handlers: [stdout]
```

## Other (related issues)

Maybe we can split these of and fix them first before we start with refactoring the logging system. Ordered by assumed difficulty.

- **Test helper**: Remove custom `capture_log` and use pytest `caplog` instead
- **Plugin name prefixing**: We currently use ad-hoc string manipulation for the plugin name instead of proper `logging.Formatter`. This adds quite a bit of unnecessary logic.
- **Import side-effects**: logging setup happens during imports (breaks test isolation + plugin order + package use)
- **Multiple verbose flags**: CLI + global config + per-plugin without clear precedence (`-v`). also `--quite` flag for some plugins. This needs a guide and maybe some rethinking.
- **Inconsistent UI vs logging**: beets uses both `ui.print()` for user messages and the `logging` system inconsistently for feedback. We might profit from defining guidelines when which should be used.

JOJ0:
@semohr I just dropped your comment here without putting much effort in it to be honest. I just want to make sure we dont forget about this. It's good ideas here in my opinion! 👍 . Feel free to edit and reword.

semohr:
[...]

PR #6618, opened 2026-05-07
Refactored test_hook to use pytest and removed capture_log

semohr:
This pull request refactors the `test/plugins/test_hook.py` test suite to use `pytest` fixtures and utilities rather than the previous unittest-based helpers. It also updates the way logs are captured in tests to use `pytest`'s `caplog` fixture instead of the custom `capture_log` context manager.

---

This is part of the multi-step efforts to improve logging in beets https://github.com/beetbox/beets/issues/6553

### DIFF
--- a/test/plugins/test_hook.py
+++ b/test/plugins/test_hook.py
@@ -17,18 +17,30 @@ from __future__ import annotations
 
 import os
 import sys
-import unittest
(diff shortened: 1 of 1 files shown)


snejus (approved):
Nice!

PR #6621, opened 2026-05-08
Add await_stable option

kaybarkbark:
When importing files automatically from a folder, beets can sometimes be too eager and will try to import a file that is still in the process of being copied. The await_stable config option polls the file repeatedly, waits for its filesize to be stable, and then performs the import.


snejus:
Can you describe when does this issue happen in practice? I don't think I understand the following bit:

> will try to import a file that is still in the process of being copied

henry-oberholtzer:
Just to get some background to evaluate this - what's triggering the automatic import action from beets in this instance? I could see running beets as a cron job, but in that situation I imagine the possible overlap with file transfer / downloads being minimal.

kaybarkbark:
Sorry if this wasn't clear. I have a `systemd` path unit that watches a particular folder on my media server. When a new file gets copied there, beets runs automatically and imports whatever it can find in that folder, and then deletes it after import. Without this fix, beets will try to import files as they are being copied, leading to numerous issues.

My previous fix was just to have a dumb delay that waits for the copy to complete. However, I kept finding myself increasing it depending on my network speed, size of the music I was copying to the folder, etc. Got tired of that, and figured there were probably other people in my situation, so I wrote this patch. I am happy to update the commit message if you would like this info as part of the source tree.

Admittedly, this might be too much of an edge case for a new feature. If you think that's the case, denying the PR won't hurt my feelings. :)

semohr:
Thanks for clarifying. My view is that this likely isn’t something that should be handled inside beets itself. A simple debounce layer before triggering any beets command should be sufficient to prevent the issue at its source, without adding complexity to the core system.

For context, this is also the approach we use in [beets-flask](https://github.com/pSpitzner/beets-flask), where we debounce file additions in the inbox before automatically dispatching commands.

kaybarkbark:
That's a fair argument and a valid fix. Thank you for the very quick review! Feel free to close this if you like.

semohr:
I’d like to get broader consensus before closing this PR. I’m also interested in hearing the perspective of the other @beetbox/maintainers. Personally, I’m not entirely convinced this belongs in core, but others may feel more strongly about it.

That said, we always appreciate contributions and the effort behind them. In the future, it would help to open a discussion or issue first before investing time in implementation so we can align on direction early.

ShimmerGlass:
I agree with @semohr

* `await_stable` is hard to configure for users: a download might be stuck for a while. and resume later. a higher value means less imports of incomplete files, but slower imports.
* the target file might be pre-allocated prior to copy / download. in this case the file size won't change, even though it is not complete yet.
* we might then choose to look at `mtime` instead, but programs like `rsync -a` might override this property
* we could then look into a checksum, but that may be slow/expansive on slow hardware

This is a hard problem to solve on beets' side. On the other hand, the program doing the file operation (copying, downloading, ...) has the full knowledge of when a file is ready and is best suited to trigger the import when appropriate.

snejus:
I agree with @semohr and @ShimmerGlass that this does not really belong in the core, so I am closing this. Thanks for your understanding @kaybarkbark ❤️ Note we will happily review any other contributions you may have.

kaybarkbark:
Thank you for your professionalism and quick review!

PR #6624, opened 2026-05-10
feat(duplicates): add chroma plugin integration

ShimmerGlass:
## Description

When the chroma plugin is enabled, the duplicates plugin can now make use of
its sonic fingerprinting capabilities to compare the tracks audio in addition of
their ``keys``. This is especially useful when multiple versions of a song, such
as live or remixes exist in the library.

When the option is enabled, tracks with the same ``keys`` but different audio
will be excluded from the results.

The following additional options are available when the chroma plugin is
enabled:

- **chroma**: Enable fingerprint comparison during duplicate search
- **chroma_threshold**: Threshold, from 0 to 1, to consider track audio the
  same. 1 means an exact match; 0 nothing alike. Default: ``0.9``.

## To Do

### DIFF
--- a/beetsplug/chroma.py
+++ b/beetsplug/chroma.py
@@ -369,6 +369,34 @@ class AcoustidPlugin(MetadataSourcePlugin):
 
         return cmd
 
+    def compare_items(
+        self, a: Item, b: Item, write: bool = False
+    ) -> float | None:
+        def item_fingerprint(item: Item) -> str | None:
+            fp = fingerprint_item(
+                self._log,
+                item,
+                write=write,
+                quiet=True,
+            )
+            if fp is None:
+                self._log.warning(f"{item}: could not compute fingerprint")
+
+            return fp
+
+        fp_a = item_fingerprint(a)
+        if fp_a is None:
+            return None
+
+        fp_b = item_fingerprint(b)
+        if fp_b is None:
+            return None
+
+        return acoustid.compare_fingerprints(
+            (0, fp_a.encode("utf-8")),
+            (0, fp_b.encode("utf-8")),
+        )
+
 
 # Hooks into import process.
 
--- a/beetsplug/duplicates.py
+++ b/beetsplug/duplicates.py
@@ -14,11 +14,13 @@
 
 """List duplicate tracks or albums."""
 
+import itertools
 import os
 import shlex
 
+from beets import ui
 from beets.library import Album, Item
-from beets.plugins import BeetsPlugin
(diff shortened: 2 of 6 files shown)


beetsplug/duplicates.py
```diff
@@ -57,162 +62,185 @@ def __init__(self):
...
             "-r",
             "--remove",
             dest="remove",
             action="store_true",
             help="remove items from library",
         )
-        self._command.parser.add_all_common_options()
+        command.parser.add_all_common_options()
 
-    def commands(self):
```
semohr, line 144: Can you revert the commands back to the orginial? I get that the refactor makes sense but it is not really in scope here and makes the diff way bigger than it has to be.

beetsplug/chroma.py
```diff
@@ -366,6 +369,34 @@ def search_cmd_func(lib, opts, args):
...
+                item,
+                write=write,
+                quiet=True,
+            )
+            if fp is None:
+                self._log.warning(f"{item}: could not compute fingerprint")
+
+            return fp
+
+        fp_a = item_fingerprint(a)
```
semohr, line 387: Why do we need to recompute the fingerprints? The chroma plugin writes to the `acoustid_fingerprint` item attribute 🤔

beetsplug/duplicates.py
```diff
@@ -402,12 +430,36 @@ def _merge(self, objs):
...
+
+            if merge:
+                objs = self._merge(objs)
+            yield (k, len(objs) - offset, objs[offset:])
+
+    def _chroma_filter_dups(self, items, chroma_thresh):
+        choma_plug = self._chroma_plug()
+        res = [items[0]]
+
+        for a, b in itertools.pairwise(items):
```
semohr, line 457: Why pairwise? That does not really make sense in my opinion.

https://docs.python.org/3/library/itertools.html#itertools.pairwise

semohr (changes requested):
Hi! Sorry for the delay in the review.

Thanks for taking a stab at it. As it stands I'm a bit concerned about the performance of the current implementation 🤔 I don't think the duplicates plugin should recompute fingerprints, the computed fingerprints should be available in the database already.

beetsplug/duplicates.py
```diff
@@ -27,6 +29,7 @@
     displayable_path,
     subprocess,
 )
+from beetsplug import chroma
```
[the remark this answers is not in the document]
semohr: Valid!

PR #6630, opened 2026-05-12
Add optional id extractor pattern attribute to metadata plugins

amogus07:
## Description

Allows external `MetadataSourcePlugin`s to supply a custom pattern to be used in `_extract_id`.

## To Do

### DIFF
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -35,6 +35,7 @@ QueryType = Literal["album", "track"]
 
 if TYPE_CHECKING:
     from collections.abc import Callable, Iterable, Iterator, Sequence
+    from typing import ClassVar
 
     from .autotag.hooks import AlbumInfo, Item, TrackInfo
 
@@ -178,6 +179,8 @@ class MetadataSourcePlugin(BeetsPlugin, metaclass=abc.ABCMeta):
 
     DEFAULT_DATA_SOURCE_MISMATCH_PENALTY = 0.5
 
+    id_extractor_pattern: ClassVar[re.Pattern[str] | None] = None
+
     @cached_classproperty
     def data_source(cls) -> str:
         """The data source name for this plugin.
@@ -278,6 +281,10 @@ class MetadataSourcePlugin(BeetsPlugin, metaclass=abc.ABCMeta):
         Uses the plugin's data source name to determine the ID format and
         extracts the ID from a given URL.
         """
+        if self.id_extractor_pattern:
+            if m := self.id_extractor_pattern.search(url):
+                return m[1]
+            return None
         return extract_release_id(self.data_source, url)
 
     @staticmethod
--- a/beetsplug/musicbrainz.py
+++ b/beetsplug/musicbrainz.py
@@ -30,7 +30,7 @@ from beets import config, plugins, util
 from beets.autotag.hooks import AlbumInfo, TrackInfo
 from beets.metadata_plugins import IDResponse, SearchApiMetadataSourcePlugin
 from beets.util.deprecation import deprecate_for_user
-from beets.util.id_extractors import extract_release_id
+from beets.util.id_extractors import UrlSource, extract_release_id
 
 from ._utils.musicbrainz import MusicBrainzAPIMixin
 from ._utils.requests import HTTPNotFoundError
@@ -78,11 +78,6 @@ BROWSE_INCLUDES = [
 ]
 BROWSE_CHUNKSIZE = 100
 BROWSE_MAXTRACKS = 500
-
-
-UrlSource = Literal[
-    "discogs", "bandcamp", "spotify", "deezer", "tidal", "beatport"
-]
 
 
 class ArtistInfo(TypedDict):
@@ -541,7 +536,7 @@ class MusicBrainzPlugin(
         """
         external_ids = self.config["external_ids"].get()
         wanted_sources: set[UrlSource] = {
-            site for site, wanted in external_ids.items() if wanted
+            UrlSource(site) for site, wanted in external_ids.items() if wanted
         }
         url_by_source: dict[UrlSource, str] = {}
         for source, url_relation in product(wanted_sources, url_relations):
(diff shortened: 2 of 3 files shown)


beets/metadata_plugins.py
```diff
@@ -193,8 +195,14 @@ def data_source_mismatch_penalty(self) -> float:
         except NotFoundError:
             return self.config["data_source_mismatch_penalty"].as_number()
 
-    def __init__(self, *args, **kwargs) -> None:
+    def __init__(
+        self,
+        id_extractor_pattern: re.Pattern[str] | None = None,
```
semohr: I'm not opposed to this direction in general, since it should help further decouple the core from the plugins.

As an initial thought (without having fully considered all of the implications) we could make this a `ClassVar` instead:

```python
id_extractor_pattern: ClassVar[re.Pattern[str] | None]
```

That approach should allow us to avoid the changes to the `__init__` method entirely.

beets/metadata_plugins.py
```diff
@@ -202,6 +205,12 @@ def __init__(self, *args, **kwargs) -> None:
             }
         )
 
+    def __init_subclass__(
+        cls, id_extractor_pattern: re.Pattern[str] | None = None
+    ) -> None:
+        cls.id_extractor_pattern = id_extractor_pattern
+        return super().__init_subclass__()
+
```
snejus: This isn't needed. Just set it to `None` by default!

```suggestion
```

snejus (changes requested):
Is there a realistic use case for this? Can you share a plugin that would make use of it?

semohr:
> Is there a realistic use case for this? Can you share a plugin that would make use of it?

This is also what I was thinking. We should not introduce abstraction that isn't really used yet. 

I like the idea in general tho as you are correct in that that the plugin specific extractor logic should not be contained in core.

I would suggest we continue as following here:
- Onboard a few metadataplugins and evaluate if we like how it changes plugin definitions
- Evaluate tradeoffs and decide if we want to go this direction (As this introduces an addition to our public plugin interface we should really not jump the gun here) 
- Check how to deprecate old logic and deprecate it
- Add this to the plugin guide and check how it integrates into the docs

amogus07:
> Is there a realistic use case for this? Can you share a plugin that would make use of it?

Any external metadata source plugin could make use of a standardized way to normalize a user-provided id, e.g. your own plugin beetcamp.

beets/metadata_plugins.py
```diff
@@ -278,7 +280,13 @@ def _extract_id(self, url: str) -> str | None:
...
         extracts the ID from a given URL.
         """
-        return extract_release_id(self.data_source, url)
+
+        if self.id_extractor_pattern and (
+            m := self.id_extractor_pattern.search(url)
+        ):
+            return m[1]
+
+        return None
```
amogus07: When no extractor pattern is provided, should we instead return the provided url as-is and emit some kind of warning?

beetsplug/musicbrainz.py
```diff
@@ -553,7 +559,7 @@ def _parse_external_ids(
                 )
 
         return {
-            f"{source}_album_id": extract_release_id(source, url)
+            f"{source}_album_id": get_metadata_source(source)._extract_id(url)
```
amogus07: One of the tests fails here because no bandcamp plugin is included in beets. Should we introduce a new function in metadata_plugins for this?

beets/metadata_plugins.py
```diff
@@ -178,6 +178,8 @@ class MetadataSourcePlugin(BeetsPlugin, metaclass=abc.ABCMeta):
 
     DEFAULT_DATA_SOURCE_MISMATCH_PENALTY = 0.5
 
+    id_extractor_pattern: ClassVar[re.Pattern[str] | None] = None
```
amogus07, line 182: This could alternatively be a provided as a str, possibly combined with using a cached_classproperty to compile it on demand.

snejus:
I think we've had this discussion somewhere before. The reason we have a centralised id extraction function is that `musicbrainz` plugin needs to be able to extract data source IDs **without** depending on corresponding plugins' presence. It means that it needs to be able to extract `bandcamp_album_id` without `bandcamp` plugin enabled.

I do agree that those regexes should ideally be defined on corresponding plugins, however we currently must define them centrally in order to solve the above issue. If those regexes were defined in plugin classes, `musicbrainz` plugin would have to import each plugin module in order to extract IDs.

I do not really see a way forward regarding this PR - even if we had another metadata source like `beetcamp`, we would define the corresponding pattern in `id_extractors.py`. But let's worry about this in the future when we do have such a plugin :)

amogus07:
I think the problem here is that id_extractors has multiple separate uses:
1. For internal plugins to extract their own ids from user-provided strings
2. In musicbrainz to extract ids from already (somewhat) normalized urls.

This solution keep the current behavior intact to allow musicbrainz (and possibly other plugins) to populate `*_album_id` fields independently of which plugins are available, but allow external plugins to easily override the default behavior without having to submit a pr to add their pattern to core. It also removes the need for each plugin (currently only musicbrainz) to maintain it's own list of valid sources.

snejus:
Thanks for the PR and for pushing this discussion forward.

I'm going to close this for now.

The main reason is that this introduces a new public plugin interface, but we don't yet have a concrete in-tree or real-world migration target that demonstrates the tradeoffs clearly enough.

Relatedly, we also don't currently need a refactor of `id_extractors` as part of this change. The current implementation is simple and serves today's use cases, so we should avoid introducing unneeded complexity and verbosity without a clear immediate need.

Thanks again for the thoughtful work here.

amogus07:
> The main reason is that this introduces a new public plugin interface, but we don't yet have a concrete in-tree or real-world migration target that demonstrates the tradeoffs clearly enough.

It would mostly be useful for external plugins, e. g. [beets-vocadb](github.com/prTopi/beets-vocadb/tree/experimental).

> Relatedly, we also don't currently need a refactor of `id_extractors` as part of this change. The current implementation is simple and serves today's use cases, so we should avoid introducing unneeded complexity and verbosity without a clear immediate need.


Discogs and other sources (again, VocaDB) also provide urls which could be used to populate external id fields. The refactor makes it so that adding more sources to the mapping immedeately makes them available in all plugins that implement this.

PR #6637, opened 2026-05-14
Improve SearchApiMetadataSourcePlugin interface

amogus07:
## Description

I'd like to refactor [beets-vocadb](https://github.com/prTopi/beets-vocadb/tree/experimental) to be a `SearchApiMetadataSourcePlugin`, but it serializes api responses to dataclass-like [msgspec structs](https://jcristharif.com/msgspec/structs.html) rather than TypedDicts, making it impossible to follow the class signature. Therefore, I removed the constraint from the type variable to allow plugin developers to use any type of data structure for search search results and instead have them implement custom logic to extract the id from a search result. For example, in beets-vocadb, I'd return `result.id` instead of `result["id"]`.

## To Do

### DIFF
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -390,21 +390,24 @@ class SearchApiMetadataSourcePlugin(
         """
 
     @abc.abstractmethod
-    def get_search_response(self, params: SearchParams) -> Sequence[R]:
+    def get_search_response(
+        self, params: SearchParams
+    ) -> tuple[int, Iterable[R]]:
         """Fetch raw search results for a provider request.
 
         Implementations should return records containing source IDs so shared
         candidate resolution can perform ID-based album and track lookups.
 
         :param params: :py:namedtuple:`~SearchParams` named tuple
-        :return: Sequence of IDResponse dicts containing at least an "id" key for each
+        :return: Tuple of (``total`` number of results, ``results`` Iterable of
+        IDResponse dicts containing at least an "id" key for each)
         """
 
         raise NotImplementedError
 
     def _search_api(
         self, query_type: QueryType, query: str, filters: dict[str, str]
-    ) -> Sequence[R]:
+    ) -> tuple[int, Iterable[R]]:
         """Run shared provider search orchestration and return ID-bearing results.
 
         This path applies optional query normalization and default limits, then
@@ -419,27 +422,27 @@ class SearchApiMetadataSourcePlugin(
 
         self._log.debug("Searching for '{}' with {}", query, filters)
         try:
-            response_data = self.get_search_response(params)
+            total, response_data = self.get_search_response(params)
         except Exception as e:
             if config["raise_on_error"].get(bool):
                 raise
             self._log.error(
                 "Error searching {.data_source}: {}", self, e, exc_info=True
             )
-            return ()
+            return 0, ()
 
-        self._log.debug("Found {} result(s)", len(response_data))
-        return response_data
+        self._log.debug("Found {} result(s)", total)
+        return total, response_data
 
     def _get_candidates(
         self, query_type: QueryType, *args, **kwargs
-    ) -> Sequence[R]:
+    ) -> Iterable[R]:
         """Resolve query hooks and execute one provider search request."""
 
         return self._search_api(
             query_type,
             *self.get_search_query_with_filters(query_type, *args, **kwargs),
-        )
+        )[1]
 
     def get_candidates_from_results(
         self, results: Iterable[R]
--- a/beets/metadata_plugins.py
+++ b/beets/metadata_plugins.py
@@ -12,17 +12,11 @@ import re
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from contextlib import contextmanager
 from functools import cache, cached_property, wraps
-from typing import (
-    TYPE_CHECKING,
-    Generic,
-    Literal,
-    NamedTuple,
-    TypedDict,
-    TypeVar,
-)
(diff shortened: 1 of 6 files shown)


snejus:
Would it be possible for you to extend your classes to support `__getitem__` method to allow indexing instead?

See `beets.autotag.hooks::AttrDict` where we do the opposite.

amogus07:
> Would it be possible for you to extend your classes to support `__getitem__` method to allow indexing instead?
> 
> See `beets.autotag.hooks::AttrDict` where we do the opposite.

I considered it but it wouldn't work because 1. the TypeVar in `SearchApiMetadataSourcePlugin` is bound to the `IDResponse` TypedDict (which I can't inherit from) and 2. I would have to use Any to annotate it which I currently don't use anywhere in the codebase and would like to avoid using.

amogus07:
I also considered making `IDResponse` a protocal, but I couldn't get that to work either.

snejus:
What's the reason for using `msgpack.Struct` in your case? `IDResponse` is here for a reason - we expect the results to be returned as `dict` types. I don't think that we are ready to relax this requirement in the core.

amogus07:
Update: I could get it working now with the latest stable, except for one patch: importing TypedDict from typing_extensions instead of typing to allow subclasses to use closed=True. It's not very polished as I've mostly copy-pasted from my existing candidates methods, but it does work. Still, please leave this pr open for now as at least that small patch would have to be merged into master, and possibly the interface could be improved in a different way too.
``` Python
class SongIDResponse(IDResponse, total=True, closed=True):
    data: SongForApiContract

class PluginBase(SearchApiMetadataSourcePlugin[IDResponse | SongIDResponse]):
	...
    @override
    def get_search_query_with_filters(
        self,
        query_type: QueryType,
        items: Sequence[library.Item],
        artist: str,
        name: str,
        va_likely: bool,
    ) -> tuple[str, dict[str, str]]:
        match query_type:
            case "album":
                criteria: dict[str, str] = (
                    {"barcode": barcode}
                    if (item := next(iter(items)))
                    and (barcode := item.get("barcode"))
                    else {}
                )
            case "track":
                item: library.Item
                barcode: str | None
                criteria = {}

        return name, {
            k: _v for k, v in criteria.items() if (_v := v.lower().strip())
        }

    @override
    def get_search_response(
        self, params: SearchParams
    ) -> Sequence[IDResponse | SongIDResponse]:
        match params.query_type:
            case "album":
                self._log.debug(msg=f"Searching for album {params.query}")
                remote_album_find_result: (
                    AlbumForApiContractPartialFindResult | None
                ) = self.album_api.api_albums_get(
                    query=params.query,
                    barcode=params.filters.get("barcode", ""),
                    maxResults=params.limit,
                    nameMatchMode=NameMatchMode.AUTO,
                )
                remote_album_candidates: tuple[AlbumForApiContract, ...] | None
                if not remote_album_find_result or not (
                    remote_album_candidates := remote_album_find_result.items
                ):
                    return []
                else:
                    self._log.debug(
                        msg=f"Found {len(remote_album_candidates)} result(s)"
                        + f" for '{params.query}'"
                    )
                # songFields parameter doesn't exist for album search
                # so we'll get albums by their id
                return [
                    {"id": album_id}
                    for remote_album_candidate in remote_album_candidates
                    if (
                        album_id := self._extract_id(
                            url=remote_album_candidate.id
                        )
                    )
                ]
            case "track":
                self._log.debug(msg=f"Searching for track {params.query}")
                remote_item_find_result: (
                    SongForApiContractPartialFindResult | None
                ) = self.song_api.api_songs_get(
                    query=params.query,
                    fields=SONG_FIELDS,
                    maxResults=params.limit,
                    nameMatchMode=NameMatchMode.AUTO,
                    preferAccurateMatches=True,
                    sort=SongSortRule.SONG_TYPE,
                    lang=self.language_preference,
                )
                remote_item_candidates: tuple[SongForApiContract, ...] | None
                if not remote_item_find_result or not (
                    remote_item_candidates := remote_item_find_result.items
                ):
                    self._log.debug(msg=f"Found 0 results for '{params.query}'")
                    return []
                self._log.debug(
                    msg=f"Found {len(remote_item_candidates)} result(s)"
                    + f" for '{params.query}'"
                )
                return [
                    {
                        "id": str(remote_item_candidate.id),
                        "data": remote_item_candidate,
                    }
                    for remote_item_candidate in remote_item_candidates
                ]

    @override
    def item_candidates(
        self, item: library.Item, artist: str, title: str
    ) -> Iterable[TrackInfo]:
        return filter(
            None,
            map(
                self.mapper.track_info,
                (
                    data
                    if (data := remote_song_candidate.get("data"))
                    and isinstance(data, SongForApiContract)
                    else None
                    for remote_song_candidate in self._get_candidates(
                        query_type="track",
                        items=[item],
                        artist=artist,
                        name=title,
                        va_likely=False,
                    )
                ),
            ),
        )
```

beets/metadata_plugins.py
```diff
@@ -395,21 +390,24 @@ def get_search_query_with_filters(
         """
 
     @abc.abstractmethod
-    def get_search_response(self, params: SearchParams) -> Sequence[R]:
+    def get_search_response(
+        self, params: SearchParams
+    ) -> tuple[int, Iterable[R]]:
```
semohr, line 395: As a sidenote: We wont be able to change the signature of functions in the plugin interfaces without a deprecation. 

We can't be sure how this class is used in the wild and as the plugin interface is the only official public api we as beets have to be really careful with changes here. If we follow semver, this change here would require a major version bump.

PR #6638, opened 2026-05-14
refactor: Pytest based setup for `test_art` and `test_embedart`

semohr:
This pull request refactors the image response mocking utilities used in tests, transitioning from a unittest-based mixin (`FetchImageHelper`) to a more flexible, pytest-compatible fixture approach. It introduces a new `ImageResponseMocker` class, updates test helpers to use pytest fixtures, and adapts the embedart plugin tests to leverage these changes for clearer and more maintainable test code.

This is quite a big one, sorry about that but there was not really a way to make this more digestible. The testing mixin hierachy we had before the refactor got quite convoluted over the years. It is still not perfect imo but it is a big step up.

TODOs:

---

### DIFF
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -878,11 +878,7 @@ class ImageRequestMocker:
             content = content.encode()
 
         self.mocker.get(
-            url,
-            headers={
-                "Content-Type": content_type,
-            },
-            content=content,
+            url, headers={"Content-Type": content_type}, content=content
         )
 
 
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -37,11 +37,10 @@ from enum import Enum
 from functools import cached_property
 from pathlib import Path
 from tempfile import gettempdir, mkdtemp, mkstemp
-from typing import Any, ClassVar
+from typing import TYPE_CHECKING, Any, ClassVar
 from unittest.mock import Mock, patch
 
 import pytest
-import responses
 from mediafile import Image, MediaFile
 
 import beets
@@ -58,6 +57,9 @@ from beets.util import (
     clean_module_tempdir,
     syspath,
 )
+
+if TYPE_CHECKING:
+    from requests_mock.mocker import Mocker
 
 
 class LogCapture(logging.Handler):
@@ -404,6 +406,8 @@ class BeetsTestCase(unittest.TestCase, TestHelper):
     modifications that will then be automatically removed when the test
     completes. Also provides some additional assertion methods, a
     temporary directory, and a DummyIO.
+
+    DEPRECATED: Use pytest + PytestTestHelper instead.
     """
 
     def setUp(self):
@@ -831,9 +835,10 @@ class AutotagImportTestCase(ImportTestCase):
 
 
 @dataclass(slots=True)
-class ImageResponseMocker:
-    responses_mock: responses.RequestsMock
+class ImageRequestMocker:
+    mocker: Mocker
 
+    # Image types and their file headers
     IMAGE_HEADERS: ClassVar[dict[str, bytes]] = {
         "image/jpeg": b"\xff\xd8\xff\x00\x00\x00JFIF",
         "image/png": b"\211PNG\r\n\032\n",
@@ -846,19 +851,21 @@ class ImageResponseMocker:
         ),
     }
 
-    def add(
+    def get(
         self,
         url: str,
         *,
         content_type: str = "image/jpeg",
         file_type: str | None = None,
-        body: str | bytes | None = None,
+        content: str | bytes | None = None,
     ) -> None:
         actual_file_type = file_type or content_type
 
-        if body is None:
+        if content is None:
             try:
-                body = self.IMAGE_HEADERS[actual_file_type].ljust(32, b"\x00")
+                content = self.IMAGE_HEADERS[actual_file_type].ljust(
+                    32, b"\x00"
+                )
             except KeyError as exc:
                 # If we can't return a file that looks like real file of the requested
                 # type, better fail the test than returning something else, which might
@@ -867,11 +874,15 @@ class ImageResponseMocker:
                     f"Mocking {actual_file_type!r} responses not supported"
                 ) from exc
 
-        self.responses_mock.add(
-            responses.GET,
+        if isinstance(content, str):
+            content = content.encode()
+
+        self.mocker.get(
             url,
-            content_type=content_type,
-            body=body,
+            headers={
+                "Content-Type": content_type,
+            },
+            content=content,
         )
 
 
@@ -879,11 +890,8 @@ class FetchImageHelper:
     """Pytest mixin providing image response mocking utilities."""
 
     @pytest.fixture
-    def image_response_mocker(self):
-        with responses.RequestsMock(
-            assert_all_requests_are_fired=False
-        ) as rsps:
-            yield ImageResponseMocker(rsps)
+    def image_request_mock(self, requests_mock):
+        return ImageRequestMocker(requests_mock)
 
 
 class CleanupModulesMixin:
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -413,6 +413,18 @@ class BeetsTestCase(unittest.TestCase, TestHelper):
         self.teardown_beets()
 
 
+class PytestTestHelper(TestHelper):
+    """Same as the BeetsTestCase unittest setup but for pytest."""
+
+    @pytest.fixture(autouse=True)
+    def setup(self):
+        self.setup_beets()
+        try:
+            yield
+        finally:
+            self.teardown_beets()
+
+
 class ItemInDBTestCase(BeetsTestCase):
     """A test case that includes an in-memory library object (`lib`) and
     an item added to the library (`i`).
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -818,16 +818,11 @@ class AutotagImportTestCase(ImportTestCase):
         self.addCleanup(self.matcher.restore)
 
 
-class FetchImageHelper:
-    """Helper mixin for mocking requests when fetching images
-    with remote art sources.
-    """
+@dataclass(slots=True)
+class ImageResponseMocker:
+    responses_mock: responses.RequestsMock
 
-    @responses.activate
-    def run(self, *args, **kwargs):
-        super().run(*args, **kwargs)
-
-    IMAGEHEADER: ClassVar[dict[str, bytes]] = {
+    IMAGE_HEADERS: ClassVar[dict[str, bytes]] = {
         "image/jpeg": b"\xff\xd8\xff\x00\x00\x00JFIF",
         "image/png": b"\211PNG\r\n\032\n",
         "image/gif": b"GIF89a",
@@ -839,29 +834,44 @@ class FetchImageHelper:
         ),
     }
 
-    def mock_response(
+    def add(
         self,
         url: str,
+        *,
         content_type: str = "image/jpeg",
-        file_type: None | str = None,
+        file_type: str | None = None,
+        body: str | bytes | None = None,
     ) -> None:
-        # Potentially return a file of a type that differs from the
-        # server-advertised content type to mimic misbehaving servers.
-        if file_type is None:
-            file_type = content_type
+        actual_file_type = file_type or content_type
 
-        try:
-            # imghdr reads 32 bytes
-            header = self.IMAGEHEADER[file_type].ljust(32, b"\x00")
-        except KeyError:
-            # If we can't return a file that looks like real file of the requested
-            # type, better fail the test than returning something else, which might
-            # violate assumption made when writing a test.
-            raise AssertionError(f"Mocking {file_type} responses not supported")
+        if body is None:
+            try:
+                body = self.IMAGE_HEADERS[actual_file_type].ljust(32, b"\x00")
+            except KeyError as exc:
+                # If we can't return a file that looks like real file of the requested
+                # type, better fail the test than returning something else, which might
+                # violate assumption made when writing a test.
+                raise AssertionError(
+                    f"Mocking {actual_file_type!r} responses not supported"
+                ) from exc
 
-        responses.add(
-            responses.GET, url, content_type=content_type, body=header
+        self.responses_mock.add(
+            responses.GET,
+            url,
+            content_type=content_type,
+            body=body,
         )
+
+
+class FetchImageHelper:
+    """Pytest mixin providing image response mocking utilities."""
+
+    @pytest.fixture
+    def image_response_mocker(self):
+        with responses.RequestsMock(
+            assert_all_requests_are_fired=False
+        ) as rsps:
+            yield ImageResponseMocker(rsps)
 
 
 class CleanupModulesMixin:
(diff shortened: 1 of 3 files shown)


test/plugins/test_art.py
```diff
@@ -232,39 +232,62 @@ def mock_caa_response(self, url, json):
         )
 
 
-class FetchImageTest(FetchImageTestCase):
-    URL = "http://example.com/test.jpg"
+class TestFetchImage(FetchImageTestCase):
+    URL: str = "http://example.com/test.jpg"
```
snejus: No need for the type here

test/plugins/test_art.py
```diff
@@ -232,39 +232,62 @@ def mock_caa_response(self, url, json):
...
+        return DummyRemoteArtSource(logger, self.plugin.config)
+
+    @pytest.fixture
+    def candidate(self, source: DummyRemoteArtSource) -> fetchart.Candidate:
+        return fetchart.Candidate(logger, source.ID, url=self.URL)
 
-    def test_invalid_type_returns_none(self):
+    def test_invalid_type_returns_none(
+        self,
+        source: DummyRemoteArtSource,
```
snejus: The types for the fixtures are clear in fixture definitions. I would suggest removing the types from these function definitions to reduce the verbosity and to underline the actual test logic.

beets/test/helper.py
```diff
@@ -850,34 +845,46 @@ def run(self, *args, **kwargs):
...
-            body=header,
+            body=body,
         )
 
 
+class FetchImageHelper:
+    """Pytest mixin providing image response mocking utilities."""
+
+    @pytest.fixture
+    def image_response_mocker(self):
```
snejus: Could we instead use `requests_mock` fixture here? Now that we're using pytest we should use it instead of `responses` library.

test/plugins/test_art.py
```diff
@@ -987,9 +987,9 @@ def test_fanarttv_only_other_images(
             next(source.get(album, settings, []))
 
 
-class ArtImporterTest(UseThePlugin):
-    def setUp(self):
-        super().setUp()
+class TestArtImporter(UseThePlugin):
+    @pytest.fixture(autouse=True)
+    def _setup(self, setup_plugin):
```
snejus, line 858: I think this should be `def setup` without the underscore?

test/plugins/test_embedart.py
```diff
@@ -76,9 +81,30 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
-class EmbedartCliTest(
-    ImportHelper, IOMixin, PluginMixin, FetchImageHelper, BeetsTestCase
-):
+class PytestPluginTestHelper(PluginMixin, TestHelper):
```
snejus, line 81: OK I think it's time to move this to a central place now 😆

test/plugins/test_art.py
```diff
@@ -640,7 +644,7 @@ def test_itunesstore_fallback_match(
         self,
         source: fetchart.ITunesStore,
         settings: Settings,
-        album: _common.Bag,
```
snejus: Yay we're finally getting rid of the `Bag`!!

snejus (commented):
Great work!! A couple of small comments.

My main reservation is a verbosity explosion due to types in each test function. I don't have a very strong objection here, but I think this adds a fair bit of duplication, especially in those test functions that accept `source`, `dbpath` and `settings` only.

See
image

for example.

semohr:
> My main reservation is a verbosity explosion due to types in each test function. I don't have a very strong objection here, but I think this adds a fair bit of duplication, especially in those test functions that accept `source`, `dbpath` and `settings` only.


I noticed that too. I'm also not too sure how we want to approach this. 

There is a tradeoff here: We have to decide if we want less verbose testing code or if we want type checked code. There were quite a few typing issues which I had to address here which where not catched previously (e.g. Settings not inheriting from FetchArtPlugin). 

If there is a way to tell mypy to check the testing functions and infer the fixture types properly, I'm all for removing the hints. I don't think this is possible tho. (Summoning @amogus07 ;) )

test/plugins/test_art.py
semohr: They are clear for us when reading but not enforced/checked.  A function body is not typechecked if the arguments are missing typehints with our current mypy config. Also I'm not sure how mypy handles pytest fixtures which aren't typed in general :thinking: 

Additionally noone is stopping us from doing something like this currently, which can introduce subtle bugs if we are not careful.
```
@pytest.fixture
def foo() -> str:
    return "bar"
    
def test_baz(bar:int):
    ...
```

Im not sure how we want to approach this:  See also https://github.com/beetbox/beets/pull/6638#issuecomment-4466891037.

test/plugins/test_art.py
semohr: Sadly we cant name it `setup`. We have another function called `setup` in the class hierarchy which would get overwritten. See `PytestTestHelper`

test/plugins/test_embedart.py
semohr: Yeah :laughing: Once we get rid of unittest completely we should also be able to just put the setup fixtures into the `TestHelper` class directly.

beets/test/helper.py
semohr: Kinda. We can get rid of `responses.RequestsMock` (see latest commit) but I dont see that we can get rid of the `image_response_mocker` fixture entirely without a bigger restructuring.

snejus:
> There were quite a few typing issues which I had to address here which were not caught previously (e.g. `Settings` not inheriting from `FetchArtPlugin`).

I usually define types while I’m actively developing tests, and I often keep them, especially when I have just a few tests with fat `parametrize` definitions above. In those cases, types don’t really steal the show.

On the other hand, here, they kind of do. I understand the argument that with our current `mypy` config untyped test functions lose type checking, but for this file I’d still optimize for readability. Also, these are legacy tests that we ideally want to refactor later; adding a lot of explicit types to every test signature now will increase churn and make that refactor harder.

Could we keep annotations where they add real signal (non-obvious fixtures/subtypes) and drop the repetitive ones in straightforward test signatures?

For comparison:

<details>
<summary><code>with_types.py</code></summary>

```python
class TestFSArt(UseThePlugin):
    @pytest.fixture
    def settings(self) -> Settings:
        return Settings(cautious=False, cover_names=("art",), fallback=None)

    @pytest.fixture
    def dpath(self) -> bytes:
        dpath = os.path.join(self.temp_dir, b"arttest")
        os.mkdir(syspath(dpath))
        return dpath

    @pytest.fixture
    def source(self) -> fetchart.FileSystem:
        return fetchart.FileSystem(logger, self.plugin.config)

    def test_finds_jpg_in_directory(
        self,
        source: fetchart.FileSystem,
        dpath: bytes,
        settings: Settings,
    ) -> None:
        _common.touch(os.path.join(dpath, b"a.jpg"))
        candidate = next(source.get(Album(), settings, [dpath]))
        assert candidate.path == os.path.join(dpath, b"a.jpg")

    def test_appropriately_named_file_takes_precedence(
        self,
        source: fetchart.FileSystem,
        dpath: bytes,
        settings: Settings,
    ) -> None:
        _common.touch(os.path.join(dpath, b"a.jpg"))
        _common.touch(os.path.join(dpath, b"art.jpg"))
        candidate = next(source.get(Album(), settings, [dpath]))
        assert candidate.path == os.path.join(dpath, b"art.jpg")

    def test_non_image_file_not_identified(
        self,
        source: fetchart.FileSystem,
        dpath: bytes,
        settings: Settings,
    ) -> None:
        _common.touch(os.path.join(dpath, b"a.txt"))
        with pytest.raises(StopIteration):
            next(source.get(Album(), settings, [dpath]))

    def test_cautious_skips_fallback(
        self,
        source: fetchart.FileSystem,
        dpath: bytes,
        settings: Settings,
    ) -> None:
        _common.touch(os.path.join(dpath, b"a.jpg"))
        settings.cautious = True
        with pytest.raises(StopIteration):
            next(source.get(Album(), settings, [dpath]))

    def test_configured_fallback_is_used(
        self,
        source: fetchart.FileSystem,
        dpath: bytes,
        settings: Settings,
    ) -> None:
        fallback = os.path.join(self.temp_dir, b"a.jpg")
        _common.touch(fallback)
        settings.fallback = fallback  # type: ignore
        candidate = next(source.get(Album(), settings, [dpath]))
        assert candidate.path == fallback

    def test_empty_dir(
        self,
        source: fetchart.FileSystem,
        dpath: bytes,
        settings: Settings,
    ) -> None:
        with pytest.raises(StopIteration):
            next(source.get(Album(), settings, [dpath]))

    def test_precedence_amongst_correct_files(
        self,
        source: fetchart.FileSystem,
        dpath: bytes,
        settings: Settings,
    ) -> None:
        images = [b"front-cover.jpg", b"front.jpg", b"back.jpg"]
        paths = [os.path.join(dpath, i) for i in images]
        for p in paths:
            _common.touch(p)
        settings.cover_names = [b"cover", b"front", b"back"]
        candidates = [
            candidate.path
            for candidate in source.get(Album(), settings, [dpath])
        ]
        assert candidates == paths

    @patch("os.path.samefile")
    def test_is_candidate_fallback_os_error(
        self,
        mock_samefile,
        source: fetchart.FileSystem,
    ) -> None:
        mock_samefile.side_effect = OSError("os error")
        fallback = os.path.join(self.temp_dir, b"a.jpg")
        self.plugin.fallback = str(fallback)
        candidate = fetchart.Candidate(logger, source.ID, fallback)
        result = self.plugin._is_candidate_fallback(candidate)
        mock_samefile.assert_called_once()
        assert not result
```

</details>

<details>
<summary><code>without_types.py</code></summary>

```python
class TestFSArt(UseThePlugin):
    @pytest.fixture
    def settings(self) -> Settings:
        return Settings(cautious=False, cover_names=("art",), fallback=None)

    @pytest.fixture
    def dpath(self) -> bytes:
        dpath = os.path.join(self.temp_dir, b"arttest")
        os.mkdir(syspath(dpath))
        return dpath

    @pytest.fixture
    def source(self) -> fetchart.FileSystem:
        return fetchart.FileSystem(logger, self.plugin.config)

    def test_finds_jpg_in_directory(self, source, dpath, settings):
        _common.touch(os.path.join(dpath, b"a.jpg"))
        candidate = next(source.get(Album(), settings, [dpath]))
        assert candidate.path == os.path.join(dpath, b"a.jpg")

    def test_appropriately_named_file_takes_precedence(
        self, source, dpath, settings
    ):
        _common.touch(os.path.join(dpath, b"a.jpg"))
        _common.touch(os.path.join(dpath, b"art.jpg"))
        candidate = next(source.get(Album(), settings, [dpath]))
        assert candidate.path == os.path.join(dpath, b"art.jpg")

    def test_non_image_file_not_identified(self, source, dpath, settings):
        _common.touch(os.path.join(dpath, b"a.txt"))
        with pytest.raises(StopIteration):
            next(source.get(Album(), settings, [dpath]))

    def test_cautious_skips_fallback(self, source, dpath, settings):
        _common.touch(os.path.join(dpath, b"a.jpg"))
        settings.cautious = True
        with pytest.raises(StopIteration):
            next(source.get(Album(), settings, [dpath]))

    def test_configured_fallback_is_used(self, source, dpath, settings):
        fallback = os.path.join(self.temp_dir, b"a.jpg")
        _common.touch(fallback)
        settings.fallback = fallback  # type: ignore
        candidate = next(source.get(Album(), settings, [dpath]))
        assert candidate.path == fallback

    def test_empty_dir(self, source, dpath, settings):
        with pytest.raises(StopIteration):
            next(source.get(Album(), settings, [dpath]))

    def test_precedence_amongst_correct_files(self, source, dpath, settings):
        images = [b"front-cover.jpg", b"front.jpg", b"back.jpg"]
        paths = [os.path.join(dpath, i) for i in images]
        for p in paths:
            _common.touch(p)
        settings.cover_names = [b"cover", b"front", b"back"]
        candidates = [
            candidate.path
            for candidate in source.get(Album(), settings, [dpath])
        ]
        assert candidates == paths

    @patch("os.path.samefile")
    def test_is_candidate_fallback_os_error(self, mock_samefile, source):
        mock_samefile.side_effect = OSError("os error")
        fallback = os.path.join(self.temp_dir, b"a.jpg")
        self.plugin.fallback = str(fallback)
        candidate = fetchart.Candidate(logger, source.ID, fallback)
        result = self.plugin._is_candidate_fallback(candidate)
        mock_samefile.assert_called_once()
        assert not result
```

</details>

Just attempted to remove these types and found ~150 removed lines!

image

semohr:
I’d generally lean toward correctness and catching issues early over reducing verbosity.  Typed fixtures/tests can catch unexpected API or fixture changes pretty quickly.

So it feels a little odd to remove valid type hints just because they add some noise, they *do* provide real guarantees.

That said, I don’t feel strongly enough to block on it, especially given the readability/refactor argument for this file. I’ve removed them again 🤷

amogus07:
> > My main reservation is a verbosity explosion due to types in each test function. I don't have a very strong objection here, but I think this adds a fair bit of duplication, especially in those test functions that accept `source`, `dbpath` and `settings` only.
> 
> I noticed that too. I'm also not too sure how we want to approach this.
> 
> There is a tradeoff here: We have to decide if we want less verbose testing code or if we want type checked code. There were quite a few typing issues which I had to address here which where not catched previously (e.g. Settings not inheriting from FetchArtPlugin).
> 
> If there is a way to tell mypy to check the testing functions and infer the fixture types properly, I'm all for removing the hints. I don't think this is possible tho. (Summoning @amogus07 ;) )

I haven't looked at this closely but for multiple methods with the same arguments, you can use a TypedDict in combination with Unpack

semohr:
> I haven't looked at this closely but for multiple methods with the same arguments, you can use a TypedDict in combination with Unpack

I was more thinking if there is a way to tell mypy to infer the type of the fixtures if they are not typed. In theory it should be possible to infer them (from my understanding). We can enable `check_untyped` for the test files but this will just set the types of all fixtures to `Any` :/ 

I'm pretty sure pylance is able to do that somehow.

test/plugins/test_art.py
```diff
@@ -206,262 +212,311 @@ class CAAHelper:
...
+        dpath = os.path.join(self.temp_dir, b"arttest")
+        os.mkdir(syspath(dpath))
+        return dpath
 
-    def test_main_interface_returns_amazon_art(self):
-        self.mock_response(self.AMAZON_URL)
-        album = _common.Bag(asin=self.ASIN)
+    def test_main_interface_returns_amazon_art(
+        self,
+        image_request_mock,
```
snejus: Much more readable now, and the diff is processed properly now! Just a couple of test func definitions still need to be folded, this one:

```suggestion
        image_request_mock
```

and `test_main_interface_gives_precedence_to_fs_art`, `test_main_interface_falls_back_to_amazon` and `test_main_interface_tries_amazon_before_aao`

semohr: Can we please enable a ruff format rule to make this consistent once and for all. How about we enable and reformat the repo with [skip-magic-trailing-comma](https://docs.astral.sh/ruff/settings/#format_skip-magic-trailing-comma)?

See https://github.com/beetbox/beets/pull/6657

snejus (approved):
Good stuff, let's resolve conflicts and merge this!

PR #6639, opened 2026-05-14
Test for unsafe input fields in web UI against XSS

plenaerts:
## Description

Try to prevent GHSA-3gxm-wfjx-m847 from happening again.

This test verifies that the Underscore.js templates in index.html use
the escaping syntax (<%- %) instead of the non-escaping syntax (<%= %).

## To Do

- [ ] Changelog.


test/plugins/test_web_xss.py
```diff
@@ -0,0 +1,108 @@
...
+Generated using mistral vibe, verified by Pieter Lenaerts <plenae@disroot.org>
+"""
+
+import re
+
+from beets.test.helper import ItemInDBTestCase
+from beetsplug import web
+
+
+class WebXSSTest(ItemInDBTestCase):
```
snejus: Would you mind using `pytest` for new tests and reusing classes like `PluginMixin`, and modifying `self.config` instead of global variables?

test/plugins/test_web_xss.py
```diff
@@ -0,0 +1,108 @@
...
+- <%- variable %> DOES escape HTML (safe)
+
+The test checks the index.html template file served by Flask to ensure
+all user data interpolations in the Underscore.js templates use the escaping
+syntax.
+
+This was reported in
+https://github.com/beetbox/beets/security/advisories/GHSA-3gxm-wfjx-m847
+and remediated in
+https://github.com/plenaerts/beets/commit/75f0d8f4899e61afb939adf02dcfb078aed23a6a
```
snejus:
```suggestion
https://github.com/beetbox/beets/commit/75f0d8f4899e61afb939adf02dcfb078aed23a6a
```

test/plugins/test_web_xss.py
```diff
@@ -0,0 +1,108 @@
...
+
+Generated using mistral vibe, verified by Pieter Lenaerts <plenae@disroot.org>
```
snejus: Move this block under `WebXSSTest`

test/plugins/test_web_xss.py
```diff
@@ -0,0 +1,108 @@
+"""Tests for XSS vulnerability in the web plugin templates.
```
snejus: Rename this module to `test_web.py`

snejus (changes requested):

test/plugins/test_web_xss.py
plenaerts: Done. I pulled in PytestTestHelper into my commit from https://github.com/beetbox/beets/blob/10b6e423b33fef335ea1fe1d20c3520ddfd382c8/beets/test/helper.py#L425.

Flask config is still global, but in the WebPluginMixin fixture now.

test/plugins/test_web_xss.py
plenaerts: Done.

test/plugins/test_web_xss.py
plenaerts: Done.

test/plugins/test_web_xss.py
plenaerts: I've refactored test_web.py and merged this XSS test in.

snejus (approved):

PR #6643, opened 2026-05-16
core/structure: move UserError to beets.exceptions

snejus:
Update all references in core, plugins, and tests to import UserError
from the new location. This centralizes exception handling and improves
code organization.


semohr (approved):
Neat! Do we use any other custom exceptions? I think there are some for beetsplug 🤔

snejus:
> Neat! Do we use any other custom exceptions? I think there are some for beetsplug 🤔

Yup, there are a few under `beets.dbcore` for query errors as well.

PR #6648, opened 2026-05-17
Support += and -= for multi-value fields in modify

### LINKED ISSUE #6587: Addition and subtraction operators for multi-value fields
### Objective

Previous discussion at https://github.com/beetbox/beets/discussions/6576

#### Goals

I'd like to have a syntax for `modify` that allows for adding or subtracting of parts of a multi-value fields such as `genres`, `artists`, etc.

For example if I have an album where each track has a different set of genres, and there's one that I'd like to add to or remove from all tracks without having to lose or reinstate the other genres manually. e.g.:

Track 1 -> genres="Jazz; Blues; Funk"
Track 2 -> genres="Soul; Funk; Blues"
...

I'd like a command that can scan through the whole album and remove any instances of "Blues" without changing the other genres in each track. It would need to account for surrounding spaces and separators, and automatically add or remove them as appropriate.

### Proposed solution

A potential  sensible syntax could be `genres-="unwanted_genre"` and `genres+="wanted_genre"`

#### Non-goals

This proposal does not cover bulk-editing multi-value fields across different albums in a single command, or fuzzy matching for removal (e.g., removing "Blues" should not also remove "Blues Rock").

#### Anti-goals

Currently this would result in the insertion of a new field called `genres-` or `genres+`, so the proposed solution would have the side effect of making it impossible to inset new custom fields with names ending in `+` or `-`. If that's not acceptable, a separate --add or --remove flag could be used instead.

wanxidong:
## Description

Fixes #6587.

This PR adds `+=` and `-=` support to `beet modify` for multi-value fields.

For example:

```bash
beet modify genres+=Funk
beet modify genres-=Blues
```

With this change, `genres+=Funk` appends `Funk` to the existing `genres` values if it is not already present, while `genres-=Blues` removes only the exact value `Blues` and preserves other values such as `Blues Rock`.

The existing `field=value` replacement behavior is preserved.

Main behavior changes:

- `genres=Jazz; Blues` still replaces the full field value as before.
- `genres+=Funk` appends `Funk` to the existing multi-value field.
- `genres+=Funk` does not create duplicate values.
- `genres-=Blues` removes only the exact value `Blues`.
- Removing `Blues` does not remove partial matches like `Blues Rock`.
- Existing value order is preserved.
- `+=` and `-=` are rejected for scalar fields with a clear user-facing error.

## To Do

- [ ] Documentation. I have not added documentation yet. I can add this to the `modify` command documentation if maintainers prefer.
- [ ] Changelog. I have not added a changelog entry yet and can add one after review if this change is accepted.
- [x] Tests. Added tests for multi-value assignment, append, duplicate prevention, exact removal, partial-match preservation, ordering, and scalar-field operator errors.

## Verification

Ran:

```bash
poetry run pytest test/ui/commands/test_modify.py -q
```

### DIFF
--- a/beets/ui/commands/modify.py
+++ b/beets/ui/commands/modify.py
@@ -1,11 +1,53 @@
 """The `modify` command: change metadata fields."""
 
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, NamedTuple
+
 from beets import library, ui
+from beets.dbcore import types
 from beets.exceptions import UserError
 from beets.util import functemplate
 from beets.util.deprecation import maybe_replace_legacy_field
 
 from .utils import do_query
+
+if TYPE_CHECKING:
+    from beets.library import LibModel
+
+
+class ModifyOperation(NamedTuple):
+    operator: str | None
+    value: str
+
+    def apply(
+        self, obj: LibModel, field: str, new_values: list[str]
+    ) -> list[str]:
+        if self.operator is None:
+            return new_values
+
+        current = list(obj[field])
+
+        if self.operator == "+":
+            return current + [
+                item for item in new_values if item not in current
+            ]
+
+        return [item for item in current if item not in new_values]
+
+
+def _is_multi_value_field(model_cls: type[LibModel], field: str) -> bool:
+    return isinstance(model_cls._type(field), types.DelimitedString)
+
+
+def _check_modify_operations(
+    model_cls: type[LibModel], mods: dict[str, ModifyOperation]
+) -> None:
+    for field, mod in mods.items():
+        if mod.operator and not _is_multi_value_field(model_cls, field):
+            raise UserError(
+                f"field {field!r} does not support the {mod.operator}= operator"
+            )
 
 
 def modify_items(lib, mods, dels, query, write, move, album, confirm, inherit):
@@ -17,6 +59,7 @@ def modify_items(lib, mods, dels, query, write, move, album, confirm, inherit):
     """
     # Parse key=value specifications into a dictionary.
     model_cls = library.Album if album else library.Item
+    _check_modify_operations(model_cls, mods)
 
     # Get the items to modify.
     items, albums = do_query(lib, query, album, False)
@@ -26,14 +69,16 @@ def modify_items(lib, mods, dels, query, write, move, album, confirm, inherit):
     # objects.
     ui.print_(f"Modifying {len(objs)} {'album' if album else 'item'}s.")
     changed = []
-    templates = {
-        key: functemplate.template(value) for key, value in mods.items()
-    }
+    templates = {}
+    for key, mod in mods.items():
+        templates[key] = functemplate.template(mod.value)
     for obj in objs:
-        obj_mods = {
-            key: model_cls._parse(key, obj.evaluate_template(templates[key]))
-            for key in mods.keys()
-        }
+        obj_mods = {}
+        for key, mod in mods.items():
+            parsed_value = model_cls._parse(
+                key, obj.evaluate_template(templates[key])
+            )
+            obj_mods[key] = mod.apply(obj, key, parsed_value)
         if print_and_modify(obj, obj_mods, dels) and obj not in changed:
             changed.append(obj)
 
@@ -97,8 +142,11 @@ def modify_parse_args(args, is_album: bool):
             dels.append(arg[:-1])  # Strip trailing !.
         elif "=" in arg and ":" not in arg.split("=", 1)[0]:
             key, val = arg.split("=", 1)
+            operator = None
+            if key.endswith(("+", "-")):
+                key, operator = key[:-1], key[-1]
             key = maybe_replace_legacy_field(key, is_album, modify=True)
-            mods[key] = val
+            mods[key] = ModifyOperation(operator, val)
         else:
             query.append(arg)
     return query, mods, dels
--- a/test/ui/commands/test_modify.py
+++ b/test/ui/commands/test_modify.py
@@ -2,17 +2,15 @@ import pytest
 from mediafile import MediaFile
 
 from beets import logging
-from beets.test.helper import BeetsTestCase, IOMixin
-from beets.ui.commands.modify import modify_parse_args
+from beets.exceptions import UserError
+from beets.test.helper import BeetsTestCase, IOMixin, TestHelper
+from beets.ui.commands.modify import ModifyOperation, modify_parse_args
 from beets.util import syspath
 
+_p = pytest.param
 
-class ModifyTest(IOMixin, BeetsTestCase):
-    def setUp(self):
-        super().setUp()
-        self.album = self.add_album_fixture()
-        [self.item] = self.album.items()
 
+class ModifyHelper(IOMixin):
     def modify_inp(self, inp: list[str], *args):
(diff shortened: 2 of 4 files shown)


JOJ0:
I really love the feature idea and the straightforward syntax suggestion but don't really have time to look into it these days.... 

If I'm not mistaken this feature was requested already, maybe even multiple times. Do you mind looking up those issues or did you already and couldn't find them?

beets/ui/commands/modify.py
```diff
@@ -1,13 +1,48 @@
...
+from beets.dbcore import types
 from beets.exceptions import UserError
 from beets.util import functemplate
 from beets.util.deprecation import maybe_replace_legacy_field
 
 from .utils import do_query
 
 
+@dataclass(frozen=True)
+class ModifyOperation:
```
snejus: I think this can be simplified to a `NamedTuple`.

beets/ui/commands/modify.py
```diff
@@ -1,13 +1,48 @@
...
+    for field, mod in mods.items():
+        if isinstance(mod, ModifyOperation) and not _is_multi_value_field(
+            model_cls, field
+        ):
+            raise UserError(
+                f"field {field!r} does not support the {mod.operator}= operator"
+            )
+
+
+def _apply_modify_operation(obj, field, mod, value):
```
snejus: I want this to be a method on `ModifyOperation`, e.g. `def apply`.

beets/ui/commands/modify.py
```diff
@@ -97,8 +136,11 @@ def modify_parse_args(args, is_album: bool):
             dels.append(arg[:-1])  # Strip trailing !.
         elif "=" in arg and ":" not in arg.split("=", 1)[0]:
             key, val = arg.split("=", 1)
+            operator = None
+            if key.endswith(("+", "-")):
+                key, operator = key[:-1], key[-1]
             key = maybe_replace_legacy_field(key, is_album, modify=True)
-            mods[key] = val
+            mods[key] = ModifyOperation(operator, val) if operator else val
```
snejus: Let's only have `ModifyOperation`s here! Simply set `operator=None` when we have no operator.

snejus (changes requested):
Well done. I think we'd like `ModifyOperation` class to be a bit smarter and own all relevant methods, see my comments

JOJ0:
hi @Victor0700 will you find time to continue on this PR? In case not, I might pick it up and address @snejus suggestions. I can't promise any timeframe though. So what's your status?

Anyway, I'm using these features and can't live out them already 😆 Thanks a ton for the submission!

JOJ0:
@snejus and @wanxidong I'm taking over here with addressing the suggested code changes. Those are done and ready for a review. Will provide docs later. HTH :-)

snejus:
@JOJ0 I love your keenness! I actually want these features as well.

beets/ui/commands/modify.py
```diff
@@ -1,13 +1,45 @@
...
 from beets import library, ui
+from beets.dbcore import types
 from beets.exceptions import UserError
 from beets.util import functemplate
 from beets.util.deprecation import maybe_replace_legacy_field
 
 from .utils import do_query
 
 
+class ModifyOperation(NamedTuple):
```
snejus, line 19: Each of the new functions and `ModifyOperation.apply` need types here

beets/ui/commands/modify.py
```diff
@@ -26,14 +59,20 @@ def modify_items(lib, mods, dels, query, write, move, album, confirm, inherit):
...
+        obj_mods = {}
+        for key, mod in mods.items():
+            parsed_value = model_cls._parse(
+                key, obj.evaluate_template(templates[key])
+            )
+            obj_mods[key] = (
+                mod.apply(obj[key], parsed_value)
+                if mod.operator is not None
+                else parsed_value
+            )
```
snejus: `ModifyOperation` now handles the case when `operation` is `None`:
```suggestion
            obj_mods[key] = mod.apply(obj[key], parsed_value)
```

test/ui/commands/test_modify.py
```diff
@@ -120,6 +121,55 @@ def test_modify_formatted(self):
             item.load()
             assert item.title == f"{orig_title} - append"
 
+    def test_modify_multi_value_assignment(self):
```
snejus: All these tests (except `test_modify_scalar_operator_error`) should be used with `@pytest.mark.parametrize`. 

Just define a new class `TestMultiValue(TestHelper)`,
define a `@pytest.fixture` for `item` and merge them into a single parametrized test. See `test_sort.py` for some examples.

snejus (changes requested):

beets/ui/commands/modify.py
JOJ0: ah yeah of course 🤦

test/ui/commands/test_modify.py
JOJ0: done. that's much better now. thanks. also I realize that @semohr tidied up here recently and tests are now in smaller files. thanks ❤️ and I wasn't aware that params can have an `id` arg. I might use that in `lastgenre`

beets/ui/commands/modify.py
```diff
@@ -1,12 +1,54 @@
...
+
+def _is_multi_value_field(model_cls: type[Model], field: str) -> bool:
+    return isinstance(model_cls._type(field), types.DelimitedString)
+
+
+def _check_modify_operations(
+    model_cls: type[Model], mods: dict[str, ModifyOperation]
+) -> None:
+    for field, mod in mods.items():
+        if mod.operator is not None and not _is_multi_value_field(
```
JOJ0:
```suggestion
        if mod.operator and not _is_multi_value_field(
```
?

JOJ0:
This might be good now? Also rebased to fewer commits.

beets/ui/commands/modify.py
snejus: yes, indeed!

beets/ui/commands/modify.py
```diff
@@ -1,12 +1,54 @@
...
 
+if TYPE_CHECKING:
+    from beets.dbcore.db import Model
+
+
+class ModifyOperation(NamedTuple):
+    operator: str | None
+    value: str
+
+    def apply(self, obj: Model, field: str, new_values: list[str]) -> list[str]:
```
snejus: I think you can use a higher level model here, `LibModel` from `beets.library`?

test/ui/commands/test_modify.py
```diff
@@ -228,14 +231,74 @@ def test_arg_parsing_query_with_exclaimation(self):
...
+    @pytest.fixture
+    def item(self):
+        album = self.add_album_fixture()
+        [item] = album.items()
+        return item
+
+    @pytest.mark.parametrize(
+        "initial_genres, modify_arg, expected_genres",
+        [
+            pytest.param(
```
snejus: Define `_p = pytest.param` alias at the top of the module - hopefully it reduces line lengths.

JOJ0:
all addressed and set to auto-merge @snejus  👍

snejus (approved):
Thanks <3

arsaboo:
This is brilliant... thanks for adding this feature.

PR #6649, opened 2026-05-17
Stop using `Any` in `TypeVar` bound to Database

amogus07:
## Description

This makes the type annotations in dbcore slightly stricter
https://peps.python.org/pep-0484/#covariance-and-contravariance

## To Do

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -68,7 +68,7 @@ if TYPE_CHECKING:
     from .query import FieldQueryType, Query, SQLiteType
     from .sort import FieldSort, Sort
 
-D_co = TypeVar("D_co", bound="Database", default="Database", covariant=True)
+D_co = TypeVar("D_co", bound="Database", covariant=True)
 
 FlexAttrs = dict[str, str]
 
@@ -788,7 +788,7 @@ class Model(ABC, Generic[D_co]):
 # Database controller and supporting interfaces.
 
 
-AnyModel = TypeVar("AnyModel", bound=Model)
+AnyModel = TypeVar("AnyModel", bound=Model["Database"])
 
 
 class Results(Generic[AnyModel]):
@@ -1073,12 +1073,12 @@ class Transaction:
 
 
 @dataclass
-class Migration(ABC):
+class Migration(ABC, Generic[D_co]):
     """Define a one-time data migration that runs during database startup."""
 
     CHUNK_SIZE: ClassVar[int] = 1000
 
-    db: Database
+    db: D_co
 
     @cached_classproperty
     def name(cls) -> str:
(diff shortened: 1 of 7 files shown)


snejus:
What do we gain from this in practical terms?

amogus07:
In Python, `TypeVar`s are invariant by default. In practice, this means that, with the invariant `TypeVar` `T`, `Model` subclasses like `LibModel` would only be allowed to supply `T` or `Database` itself as type parameters and not subclasses of `Database` like `Library`. Previously, this intentional restriction was circumvented using `default=Any` as a workaround, which disables typechecking entirely as far as I understand.
``` Python
def example(m: Model) -> None:
    return m.db # type: Any
```
. Now, by replacing `T` with the covariant `T_co`, subclasses of `Model` (or any other class using it) are allowed to use any subclass of `Database`, making `Libmodel(Model[Library])` valid without using `Any`. Now, m.db from the example function is inferred as being of type `Database` rather than `Any`. I could've kept the default value in `T_co` and replaced it with `"Database"`, but that would obscure the fact that Model is generic to consumers. As mentioned in one of the commit messages, I already found some incorrect type annotations in the `library` module where `Model[Any]` was (implicitly) used instead of `Model[Library]` which is more appropriate.

snejus:
The main reason why we had `Any` there is that most of these classes that receive/depend on `Model` class **do not care** that it has a generic relationship to `Database`.

This is why I'm asking whether making this typing stricter has any **practical** benefits. From what I can see, it simply makes type definitions more verbose without a clear benefit.

amogus07:
@snejus is it better now?

snejus:
It's a bit less verbose now, however, as I said all these classes need **any** `Model`, therefore we don't win anything from stricter type annotations.

snejus:
Hi @amogus07,

Looking back at both [#5224](https://github.com/beetbox/beets/pull/5224) (specifically, [this thread](https://github.com/beetbox/beets/pull/5224#pullrequestreview-2488361510)) and [#5545](https://github.com/beetbox/beets/pull/5545), my understanding is that this type parameter was introduced for one specific reason: fixing the `_check_db` typing problem caused by optional `_db` presence, so `LibModel` can be typed against `Library` instead of falling back to `Database`/`Any`.

We used `default=Any` intentionally so existing non-parameterized `Model` usages remain valid without forcing explicit type arguments everywhere in code that does not care about the concrete DB subtype.

Given that history, I don't think we should expand this into broad generic threading across dbcore - most classes touched in this PR only need "any `Model`", and the extra strictness mostly adds verbosity and maintenance cost without a concrete payoff for beets-core.

My preference is to keep this scoped to the `_check_db`/DB-association issue it was introduced for (as in [#5224](https://github.com/beetbox/beets/pull/5224) and [#5545](https://github.com/beetbox/beets/pull/5545)), and not invest further unless a patch shows direct practical wins - real `type: ignore`/`cast` removals, or an actual bug caught.

As I said, I don't see any practical benefit in making these types more strict, therefore I am closing this for now.

PR #6657, opened 2026-05-20
Add `skip-magic-trailing-comma` ruff format setting

semohr:
This PR enables Ruff’s `skip-magic-trailing-comma` setting in the formatter configuration and applies formatting across the codebase.

Specifically:

* Enables `skip-magic-trailing-comma` in Ruff format settings
* Re-formats the codebase using `ruff format`

This aligns trailing comma usage with line length. As a result:

* Trailing commas are only preserved when they meaningfully support multiline formatting
* Single-line expressions are kept clean without unnecessary trailing commas
* Formatting becomes more consistent and deterministic across the project


snejus (approved):
Yes please!!!

PR #6658, opened 2026-05-20
Moved PytestPluginTestHelper class into test_helper.

semohr:
This PR moved the `PytestPluginTestHelper` which was duplicated in quite a few test files into the `beets/test/helper.py` file.

### DIFF
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -487,7 +487,24 @@ class PluginMixin(ConfigMixin):
 
 
 class PluginTestCase(PluginMixin, BeetsTestCase):
+    """
+    DEPRECATED: Use pytest + PytestPluginTestHelper instead.
+    """
+
     pass
+
+
+class PytestPluginTestHelper(PluginMixin, PytestTestHelper):
+    """Helper mixin for pytest-based plugin tests.
+
+    This mixin provides the standard beets test setup and automatically
+    initializes and tears down plugin state for each test.
+
+    .. code-block:: python
+
+        class TestMyPlugin(PytestPluginTestHelper):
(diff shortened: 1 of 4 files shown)


beets/test/helper.py
```diff
@@ -487,9 +487,34 @@ def configure_plugin(self, config: Any):
 
 
 class PluginTestCase(PluginMixin, BeetsTestCase):
+    """
+    DEPRECATED: Use pytest + PytestPluginTestHelper instead.
```
semohr, line 491: As an fyi: We do not use deprecation warning here as this is *very* spammy when running tests.

snejus (approved):
Note that another PR that I've just merged defined a similar class in `helper.py` - you will want to remove it after rebase.

semohr:
Hmm, which class do you mean?

snejus:
> Hmm, which class do you mean?

See line 420 in `helper.py` after rebase.

semohr:
> See line 420 in `helper.py` after rebase.

That is not the same class, it does not extend from the plugin mixin.

snejus:
> > See line 420 in `helper.py` after rebase.
> 
> That is not the same class, it does not extend from the plugin mixin.

True, though I think you want to extend it here with `PluginMixin` as it has the same fixture.

semohr:
> True, though I think you want to extend it here with PluginMixin as it has the same fixture.

I tried very shortly combining the two classes but it could not figure out how to make it work with the inheritance pattern we currently follow. We are running a bit of a strange pytest setup imo. I would properly remove these classes in favor of a proper fixture and a conftest based setup at some point. 

 For now, the classes are right next to each other so a future refactor should be straight forward :laughing:

PR #6668, opened 2026-05-24
Make aliases-as-credits functionality opt-in.

djl:
## Description

This feature was introduced in 2.10 but results a lot of data being overwritten with (potentially) incorrect data the next time a user runs `mbsync` over their library.

To make this less of a "breaking" change, make this feature opt-in. The users who want this large change can enable the functionality.

### DIFF
--- a/beetsplug/musicbrainz.py
+++ b/beetsplug/musicbrainz.py
@@ -277,6 +277,10 @@ class MusicBrainzPlugin(
     def ignore_video_tracks(self) -> bool:
         return config["match"]["ignore_video_tracks"].get(bool)
 
+    @cached_property
+    def aliases_as_credits(self) -> bool:
+        return self.config["aliases_as_credits"].get(bool)
+
     def __init__(self) -> None:
         """Set up the python-musicbrainz-ngs module according to settings
         from the beets configuration. This should be called at startup.
@@ -294,6 +298,7 @@ class MusicBrainzPlugin(
                     "tidal": False,
                 },
                 "extra_tags": [],
+                "aliases_as_credits": False,
             }
         )
         # TODO: Remove in 3.0.0
@@ -307,8 +312,9 @@ class MusicBrainzPlugin(
                 "'musicbrainz.search_limit'",
             )
 
-    @staticmethod
-    def _parse_artist_credits(artist_credits: list[ArtistCredit]) -> ArtistInfo:
+    def _parse_artist_credits(
+        self, artist_credits: list[ArtistCredit]
+    ) -> ArtistInfo:
         """Normalize MusicBrainz artist-credit data into tag-friendly fields.
 
         MusicBrainz represents credits as a sequence of credited artists, each
@@ -336,7 +342,9 @@ class MusicBrainzPlugin(
             artists_ids.append(el["artist"]["id"])
             alias = _preferred_alias(el["artist"].get("aliases", []))
             artist_object = alias or el["artist"]
-            credit_artist_object = alias or el
+            credit_artist_object = (
+                alias if (alias and self.aliases_as_credits) else el
+            )
 
             joinphrase = el["joinphrase"]
             for name, parts, multi in (
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -23,6 +23,7 @@ New features
 - :doc:`plugins/listenbrainz`: Add support for importing ListenBrainz listening
   history from an export file. Use the ``-f`` / ``--export-file`` flag to
   specify the path to the ListenBrainz export file.
+- :doc:`plugins/musicbrainz`: Make aliases-as-artist-credit optional.
 
 Bug fixes
 ~~~~~~~~~
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -23,7 +23,9 @@ New features
 - :doc:`plugins/listenbrainz`: Add support for importing ListenBrainz listening
   history from an export file. Use the ``-f`` / ``--export-file`` flag to
   specify the path to the ListenBrainz export file.
-- :doc:`plugins/musicbrainz`: Make aliases-as-artist-credit optional.
+- :doc:`plugins/musicbrainz`: Introduce
+  :conf:`plugins.musicbrainz:aliases_as_credits` to make
+  aliases-as-artist-credit optional.
 
 Bug fixes
 ~~~~~~~~~
--- a/test/plugins/test_musicbrainz.py
+++ b/test/plugins/test_musicbrainz.py
@@ -90,7 +90,7 @@ class TestUtils:
     def test_single_artist(self):
         credit = [artist_credit_factory(artist__name="Artist")]
 
-        assert MusicBrainzPlugin._parse_artist_credits(credit) == {
+        assert MusicBrainzPlugin()._parse_artist_credits(credit) == {
             "artist": "Artist",
             "artist_id": "00000000-0000-0000-0000-000000000001",
             "artist_sort": "Artist, The",
@@ -107,7 +107,7 @@ class TestUtils:
             artist_credit_factory(artist__name="Other Artist", artist__index=2),
         ]
 
-        assert MusicBrainzPlugin._parse_artist_credits(credit) == {
+        assert MusicBrainzPlugin()._parse_artist_credits(credit) == {
             "artist": "Artist AND Other Artist",
             "artist_id": "00000000-0000-0000-0000-000000000001",
             "artist_sort": "Artist, The AND Other Artist, The",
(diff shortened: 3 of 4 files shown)


beets/config_default.yaml
```diff
@@ -203,3 +203,6 @@ match:
     track_length_max: 30
     album_disambig_fields: data_source media year country label catalognum albumdisambig
     singleton_disambig_fields: data_source index track_alt album
+
+musicbrainz:
+    aliases_as_credits: no
```
amogus07:
```suggestion
```
I don't think this belongs here. The default value is already specified in the musicbrainz plugin.

beetsplug/musicbrainz.py
```diff
@@ -335,7 +336,11 @@ def _parse_artist_credits(artist_credits: list[ArtistCredit]) -> ArtistInfo:
             artists_ids.append(el["artist"]["id"])
             alias = _preferred_alias(el["artist"].get("aliases", []))
             artist_object = alias or el["artist"]
-            credit_artist_object = alias or el
+            credit_artist_object = (
+                alias
+                if (alias and config["musicbrainz"]["aliases_as_credits"].get())
+                else el
+            )
```
amogus07, line 347: You should probably use `cached_property` to get the value, and pass `bool` to the get method.

amogus07 (dismissed):

amogus07:
Oh, and you need to resolve the conflict in the changelog.

beetsplug/musicbrainz.py
djl: Same as above, this is inside a class method so no access to `self`. Good shout on the `.get(bool)`.

beets/config_default.yaml
djl: Without this the `mbpseudo` plugin fails due the missing config, because `_parse_artist_credits` is a class method with no access to `self`.

amogus07: Then turn `_make_parse_artist_credit` into an instance method I suppose. It shouldn't be a problem since it's only ever called from an instance. Only some tests will have to be adjusted as far as I can tell.

djl: Didn't think of that. Done!

test/plugins/test_musicbrainz.py
```diff
@@ -605,38 +605,58 @@ def test_missing_language(self, mb):
...
+        assert first_track.artists_credit == [track_artist_credit]
 
-        track_artist_sort = "Recording Artist Alias en, The"
+        track_artist_sort = (
+            "Recording Artist Alias en, The"
+            if aliases_as_credits
+            else "Recording Artist, The"
+        )
         assert first_track.artist_sort == track_artist_sort
         assert first_track.artists_sort == [track_artist_sort]
```
amogus07, line 644: Some suggestions regarding this test:
1. Instead of adding the `aliases_as_credits` argument and manually updating the config, pass the value to `config_dict`. Then you can just use `mb.aliases_as_credits` instead of `aliases_as_credits`
2. You added a bunch of conditional logic within the test method, which is fragile and hard to extend. Instead, add `expected_*` arguments to the method and pass the corresponding values in `pytest.parametrize`.

docs/changelog.rst
```diff
@@ -19,6 +19,7 @@ New features
 - :doc:`plugins/listenbrainz`: Add support for importing ListenBrainz listening
   history from an export file. Use the ``-f`` / ``--export-file`` flag to
   specify the path to the ListenBrainz export file.
+- :doc:`plugins/musicbrainz`: Make aliases-as-artist-credit optional.
```
```suggestion
- :doc:`plugins/musicbrainz`: Introduce :conf:`plugins.musicbrainz:aliases_as_credits` to make aliases-as-artist-credit optional.
```

test/plugins/test_musicbrainz.py
```diff
@@ -605,38 +605,58 @@ def test_missing_language(self, mb):
     def test_genres(self, mb, expected_genres):
         assert mb.album_info(release_factory()).genres == expected_genres
 
-    def test_parse_aliased_titles(self, config, mb: MusicBrainzPlugin):
+    @pytest.mark.parametrize(
+        "aliases_as_credits",
+        [_p(False, id="no aliases"), _p(True, id="aliases")],
+    )
```
snejus: Use `plugin_config` fixture here to set the plugin configuration.

```suggestion
        "plugin_config",
        [_p({"aliases_as_credits": False}, id="no aliases"), _p({"aliases_as_credits": True}, id="aliases")],
    )
```

test/plugins/test_musicbrainz.py
```diff
@@ -605,38 +605,58 @@ def test_missing_language(self, mb):
...
+    def test_aliases_as_credits(
+        self, config, mb: MusicBrainzPlugin, aliases_as_credits: bool
+    ):
+        config["import"]["languages"] = ["en"] if aliases_as_credits else []
```
snejus: I'd suggest keeping the language config unconditional here like before.

test/plugins/test_musicbrainz.py
```diff
@@ -605,38 +605,58 @@ def test_missing_language(self, mb):
...
-        album_artist = "Artist Alias en"
+        album_artist = aliased_name("Artist")
         assert d.artist == album_artist
         assert d.artists == [album_artist]
+
         # There is no artist credit specific alias
-        assert d.artist_credit == album_artist
-        assert d.artists_credit == [album_artist]
+        album_artist_credit = (
+            album_artist if aliases_as_credits else "Artist Credit"
```
snejus: I don't want to see conditional logic in the tests. I would suggest keeping this existing test as it was and adding a new one to test `aliases_as_credits` specifically.

Ideally I'd like to see a `pytest.mark.parametrize` with `plugin_config` and `expected_artist_credit` parameters without any conditional logic.

snejus (changes requested):

test/plugins/test_musicbrainz.py
amogus07: Regarding the first one, wouldn't an empty config suffice? It should be False by default anyway.

amogus07:
@snejus Now that I think about it, would anyone even need an option for this? Especially since the new behavior was only introduced recently. Doesn't making `artist_credit` the same as `artist` defeat its purpose? We could just add a notice in the changelog that if anyone still wants the option, they should stay on the current version, open an issue and wait for the next one.

snejus:
@amogus07 not sure if I understood you correctly, but the fact that this behaviour was added recently means that users wanted it?

amogus07:
Right. @Morikko Do you need the option? Or, if not, do you think we should have it?

Morikko:
https://github.com/beetbox/beets/pull/6552#issue-4290428111

The description linked above is explaining my issue, I had both aliased and not in the credits. So, I wanted to uniformize it though i don't know where the divergence came from.

I am not relying on credits so I don't mind the solution. I chose to use alias because in my use case the artists would be in their original languages (japanese, arabic...) and I would never be able to read it so it has no interest at all. But for the sake of not having duplicated fields, it might be good to keep origal artists there.

amogus07:
My usecase is the same. So as long as non-credit fields use aliases, I personally don't see why credit fields should use them too.

amogus07:
Also, I'm planning to submit a pr to only use aliases for non-credit fields when their language and script are prioritized in the config over the respective release's language and script (opt-in of course). Having both options would make things a bit more complicated. But that's fine if you still think it's needed.

test/plugins/test_musicbrainz.py
snejus: We prefer making it explicit here.

snejus:
Thanks for the feedback and test suggestions.

For this PR, I'd like to keep scope focused on addressing the regression reported by users and getting maintainer guidance on the default/option decision. I'd prefer not to branch the thread into parallel product-direction discussions here.

If there's a broader proposal about alias behavior policy, please open a separate issue/PR so we can discuss that independently.

I'm going to keep this thread limited to actionable comments for the current PR scope.

test/plugins/test_musicbrainz.py
```diff
@@ -605,40 +605,90 @@ def test_missing_language(self, mb):
...
+                    "artist_credit": "Artist Alias en",
+                    "artists_credit": ["Artist Alias en"],
+                    "track_artist_credit": "Recording Artist Alias en",
+                    "track_artists_credit": ["Recording Artist Alias en"],
+                },
+                id="aliases",
+            ),
+        ],
+    )
+    def test_aliases_as_credits(
```
amogus07, line 671: You should add another test case with `{"aliases_as_credits": False}` and `langues_config` as `["en"]`.

djl: Done.

test/plugins/test_musicbrainz.py
```diff
@@ -605,40 +605,109 @@ def test_missing_language(self, mb):
     def test_genres(self, mb, expected_genres):
         assert mb.album_info(release_factory()).genres == expected_genres
 
-    def test_parse_aliased_titles(self, config, mb: MusicBrainzPlugin):
```
snejus, line 608: @djl I asked to keep the previous test as it was and to add a separate test to test the credit fields only

djl: My bad. Brought back the original test.

test/plugins/test_musicbrainz.py
```diff
@@ -640,6 +643,110 @@ def test_parse_aliased_titles(self, config, mb: MusicBrainzPlugin):
         assert first_track.artist_sort == track_artist_sort
         assert first_track.artists_sort == [track_artist_sort]
 
+    @pytest.mark.parametrize(
+        "plugin_config, languages_config, expected_artist_credit",
```
snejus: I don't think you need `languages_config` as a parameter here. You can just set it to `["en"]` in every case, no?

test/plugins/test_musicbrainz.py
```diff
@@ -640,6 +643,110 @@ def test_parse_aliased_titles(self, config, mb: MusicBrainzPlugin):
...
+                    "album_artist_sort": "Artist Alias en, The",
+                    "album_artists_sort": ["Artist Alias en, The"],
+                    "track_artist": "Recording Artist Alias en",
+                    "track_artists": ["Recording Artist Alias en"],
+                    "artist_sort": "Recording Artist Alias en, The",
+                    "artists_sort": ["Recording Artist Alias en, The"],
+                    "artist_credit": "Artist Credit",
+                    "artists_credit": ["Artist Credit"],
+                    "track_artist_credit": "Recording Artist Credit",
+                    "track_artists_credit": ["Recording Artist Credit"],
```
snejus: The other test already tests the `languages` config and aliases and here I'd like to see this test only cover the fields that `aliases_as_credits` adjusts. I think it's only related to the `*_credit` fields, right?

djl: Right. I simplified the tests to only check the `*_credit` fields.

test/plugins/test_musicbrainz.py
djl: Yep, good point. Done.

snejus (approved):

PR #6674, opened 2026-05-27
Fix: browse recordings missing aliases for releases with more than 500 tracks

MartinCa:
## Description

When a release exceeds BROWSE_MAXTRACKS tracks, _ensure_complete_recordings
fetches recordings via browse_recordings. BROWSE_INCLUDES was missing "aliases",
so the API never returned alias data, causing a KeyError in get_tracks_from_medium.

Add "aliases" to BROWSE_INCLUDES and defensively use .get("aliases", []) at the
access site, consistent with all other alias lookups in the file.

## To Do

### DIFF
--- a/test/plugins/test_musicbrainz.py
+++ b/test/plugins/test_musicbrainz.py
@@ -672,39 +672,34 @@ class TestParseRelease(MusicBrainzPluginTestMixin):
         ]
         assert new_recordings == complete_recordings
 
-    def test_album_info_browse_recordings_without_aliases(
-        self, monkeypatch, mb
+    def test_album_info_browse_recordings_prefers_alias_over_track_title(
+        self, config, monkeypatch, mb
     ):
-        """Recordings fetched via browse_recordings may lack 'aliases' if the
-        API includes were missing; album_info must not raise KeyError."""
-        initial_recordings = [recording_factory(index=idx) for idx in range(2)]
-        # Simulate browse_recordings returning recordings without 'aliases'
-        browsed_recordings = [
-            {k: v for k, v in r.items() if k != "aliases"}
-            for r in initial_recordings
-        ]
+        config["import"]["languages"] = ["en"]
+
+        recording = recording_factory(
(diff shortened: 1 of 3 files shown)


snejus:
> defensively use .get("aliases", [])

Given that you now included `aliases`, why is there a need to be defensive here? Is the keys not always returned now?

MartinCa:
> > defensively use .get("aliases", [])
> 
> Given that you now included `aliases`, why is there a need to be defensive here? Is the keys not always returned now?

Get in general should be safer as it will handle if `aliases` ever is missing for some reason. However if we do consider that an error then get would now mask that error. Should I revert it so we will get an error if aliases is missing?

snejus:
> Should I revert it so we will get an error if aliases is missing?

I'd say so. Had we had `get` here previously, we would have never found out that `aliases` were missing and that this data had not been fetched/tagged.

test/plugins/test_musicbrainz.py
```diff
@@ -672,6 +672,40 @@ def test_ensure_complete_recordings(self, monkeypatch, mb):
         ]
         assert new_recordings == complete_recordings
 
+    def test_album_info_browse_recordings_without_aliases(
```
snejus: And thus I don't think we need this test. I think it's just enough to include the missing key in includes without the need for this test.

snejus (changes requested):

test/plugins/test_musicbrainz.py
MartinCa: Good point, I have replaced this test with a test that tests that alias actually takes precedence over the title instead.

MartinCa:
> > Should I revert it so we will get an error if aliases is missing?
> 
> I'd say so. Had we had `get` here previously, we would have never found out that `aliases` were missing and that this data had not been fetched/tagged.

Yes very good point I have reverted the get.

snejus (approved):

PR #6677, opened 2026-05-29
refactor: Pytest based setup for `test_library`

semohr:
## Description

This PR is a refactor of `test/test_library.py` and replaces the unitest test cases with a pytest setup. Also replaced capture_log with caplog fixture.

Needs rebasing after https://github.com/beetbox/beets/pull/6659 is merged.

TODOs:
---

### DIFF
--- a/test/test_library.py
+++ b/test/test_library.py
@@ -193,16 +193,12 @@ class TestGetSet(PytestItemHelper):
         assert i.get("flexx") is None
 
 
-class DestinationTest(BeetsTestCase):
+class TestDestination(PytestItemInDBHelper):
     """Confirm tests handle temporary directory path containing '.'"""
 
     def create_temp_dir(self, **kwargs):
         kwargs["prefix"] = "."
         return super().create_temp_dir(**kwargs)
-
-    def setUp(self):
-        super().setUp()
-        self.i = item(self.lib)
 
     def test_directory_works_with_trailing_slash(self):
         self.lib.directory = b"one/"
--- a/test/test_library.py
+++ b/test/test_library.py
@@ -732,9 +732,9 @@ class TestDestinationFunction(PytestTestHelper, PathFormattingMixin):
         self._assert_dest(b"/base/Alice & Bob")
 
 
-class DisambiguationTest(BeetsTestCase, PathFormattingMixin):
-    def setUp(self):
-        super().setUp()
+class TestDisambiguation(PytestTestHelper, PathFormattingMixin):
+    @pytest.fixture(autouse=True)
+    def setup_lib(self, setup):
         self.lib.directory = b"/base"
         self.lib.path_formats = [("default", "path")]
 
--- a/test/test_library.py
+++ b/test/test_library.py
@@ -46,7 +46,13 @@ from beets.util import (
 np = util.normpath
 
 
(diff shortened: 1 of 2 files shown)


test/test_library.py
```diff
@@ -46,7 +47,19 @@
 np = util.normpath
 
 
-class LoadTest(ItemInDBTestCase):
+class PytestItemInDBHelper(PytestTestHelper):
+    @pytest.fixture(autouse=True)
+    def item(self, setup):
```
snejus: Use `pytest.mark.usefixtures("setup")` instead

snejus: Or better - define a `helper` fixture and use it here. There are many `pytest`-based tests that define `helper` fixture to do the setup.

test/test_library.py
```diff
@@ -46,7 +47,19 @@
...
+        self.i = _common.item(self.lib)
```
snejus: Hmm, I'm not a big fan of this - this is only a half-way migration to `pytest`. This should return an item instead of setting an attribute.

snejus (changes requested):

test/test_library.py
semohr: Yeah. I started with that too but noticed pretty quickly that there are like 300 occurrences of `self.i` in the file. I wanted to keep this as unintrusive as possible. 

That said, we can refactor this further here but for me this feels a bit like a job for another PR. There are also many places here were it would make sense to use a parameterized test.

test/test_library.py
semohr: > Or better - define a `helper` fixture and use it here. There are many `pytest`-based tests that define `helper` fixture to do the setup.

I dont think I can follow. The setup parameter is used here because we need `self.lib` set, the `item` fixture can be executed before `setup` which yield an error here. In a pytest setup without the class hierachy, I would have a fixture for `lib` which would make the dependency clear.

test/test_library.py
snejus: I'd say we should either keep it as it is or migrate it fully 🤷🏼

semohr: Fine I can refactor it.  We use quite a similar pattern all over the place tho see e.g. `PytestHelper` or `PytestPluginHelper`. 

The `PytestItemInDBHelper` class are pretty much the same pattern.

semohr: See latest commit :upside_down_face:

test/test_library.py
```diff
@@ -1330,41 +1350,41 @@ def test_write_date_field(self):
         assert MediaFile(syspath(item.path)).year == clean_year
 
 
-class ItemReadTest(unittest.TestCase):
+class TestItemRead:
     def test_unreadable_raise_read_error(self):
         unreadable = os.path.join(_common.RSRC, b"image-2x3.png")
-        item = beets.library.Item()
```
snejus, line 1336: Is there a need to adjust `item` -> `i` in a couple of these tests?

snejus (commented):
Consider adding this last commit to `.git-blame-ignores-revs`.

test/test_library.py
semohr: Not really, was just for consistency.

snejus (approved):

PR #6680, opened 2026-05-29
refactor: Removed `ImportTestCase` in favor of `ImportHelper`

semohr:
## Description

This PR removes `ImportTestCase` and replaces all occurrences with  `ImportHelper` + `pytest`.

This touches:
- `test_scrub`
- `test_filefilter`
- `test_replaygain`
- `test_chroma`
- `test_import`

TODOs:
---

### DIFF
--- a/test/test_importer.py
+++ b/test/test_importer.py
@@ -45,9 +45,10 @@ from beets.test.helper import (
     AutotagImportTestCase,
     AutotagStub,
     BeetsTestCase,
-    ImportTestCase,
+    ImportHelper,
     IOMixin,
     PluginMixin,
+    TestHelper,
     capture_log,
     has_program,
 )
@@ -71,7 +72,7 @@ class PathsMixin:
         return self.lib_path / "Tag Artist" / "Tag Album" / "Tag Track 1.mp3"
 
 
-class NonAutotaggedImportTest(PathsMixin, AsIsImporterMixin, ImportTestCase):
+class TestNonAutotaggedImport(PathsMixin, AsIsImporterMixin, ImportHelper):
     db_on_disk = True
 
     def test_album_created_with_track_artist(self):
@@ -125,7 +126,7 @@ class NonAutotaggedImportTest(PathsMixin, AsIsImporterMixin, ImportTestCase):
         album = self.lib.albums()[0]
         assert album.mb_albumartistids == album.items()[0].mb_albumartistids
 
-    @unittest.skipUnless(_common.HAVE_SYMLINK, "need symlinks")
+    @pytest.mark.skipif(not _common.HAVE_SYMLINK, reason="need symlinks")
     def test_import_link_arrives(self):
         self.run_asis_importer(link=True)
 
@@ -133,7 +134,7 @@ class NonAutotaggedImportTest(PathsMixin, AsIsImporterMixin, ImportTestCase):
         assert self.track_lib_path.is_symlink()
         assert self.track_lib_path.resolve() == self.track_import_path.resolve()
 
-    @unittest.skipUnless(_common.HAVE_HARDLINK, "need hardlinks")
+    @pytest.mark.skipif(not _common.HAVE_HARDLINK, reason="need hardlinks")
     def test_import_hardlink_arrives(self):
         self.run_asis_importer(hardlink=True)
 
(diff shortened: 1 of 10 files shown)


test/test_importer.py
```diff
@@ -143,7 +144,9 @@ def test_import_hardlink_arrives(self):
         assert media_stat[stat.ST_INO] == lib_media_stat[stat.ST_INO]
         assert media_stat[stat.ST_DEV] == lib_media_stat[stat.ST_DEV]
 
-    @NEEDS_REFLINK
```
snejus, line 146: Maybe adjust `NEEDS_REFLINK` to use `pytest.mark.skipif` instead?

test/test_importer.py
```diff
@@ -386,16 +389,17 @@ def test_set_fields(self):
...
-class ImportFormatTest:
+class TestImportFormat(ImportHelper):
     """Test fix_extension during import."""
 
     def test_recognize_format(self):
         resource_src = os.path.join(_common.RSRC, b"no_ext")
         resource_path = os.path.join(self.import_dir, b"no_ext")
         util.copy(resource_src, resource_path)
-        self.setup_importer()
+        self.setup_importer(autotag=False)
```
snejus, line 399: Why this change?

test/test_importer.py
```diff
@@ -259,19 +262,19 @@ def create_archive(self):
...
 
-class Import7zTest(ImportZipTest):
+class TestImport7z(TestImportZip):
     def create_archive(self):
         return os.path.join(_common.RSRC, b"archive.7z")
 
 
-@unittest.skip("Implement me!")
-class ImportPasswordRarTest(ImportZipTest):
+@pytest.mark.skip("Implement me!")
```
snejus:
```suggestion
@pytest.mark.skip(reason="Implement me!")
```

test/test_logging.py
```diff
@@ -149,16 +149,19 @@ def __init__(self, *_, **__):
...
+class TestLoggingLevel(AsIsImporterMixin, PluginMixin, ImportHelper):
     plugin = "dummy"
 
-    @classmethod
-    def setUpClass(cls):
-        patcher = patch.dict(sys.modules, {"beetsplug.dummy": DummyModule()})
-        patcher.start()
-        cls.addClassCleanup(patcher.stop)
+    def setup_beets(self):
+        self._dummy_patcher = patch.dict(
```
snejus: Maybe we should instead use a class-scoped autouse fixture for this, and patch it with `monkeypatch`?

test/ui/test_ui_importer.py
```diff
@@ -23,7 +23,7 @@
 
 
 class NonAutotaggedImportTest(
-    TerminalImportMixin, test_importer.NonAutotaggedImportTest
+    TerminalImportMixin, test_importer.TestNonAutotaggedImport
 ):
     pass
```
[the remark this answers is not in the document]
snejus: Good point

test/test_logging.py
semohr: Sure, we can do that :+1: Doesnt matter too much but might give us a bit of performance. Would be the more apples to apples port indeed.

test/test_importer.py
semohr: Have a look at the name and the class inheritance before and after. Am pretty sure the tests never ran.

Without the autotag=False fix we get this error:
```
FAILED test/test_importer.py::TestImportFormat::test_recognize_format - IndexError: list index out of range
```

snejus: Great point!

test/test_logging.py
```diff
@@ -149,16 +149,13 @@ def __init__(self, *_, **__):
...
-    @classmethod
-    def setUpClass(cls):
-        patcher = patch.dict(sys.modules, {"beetsplug.dummy": DummyModule()})
-        patcher.start()
-        cls.addClassCleanup(patcher.stop)
-
-        super().setUpClass()
+    @pytest.fixture(autouse=True, scope="class")
+    def _patch_dummy_module(self):
+        with patch.dict(sys.modules, {"beetsplug.dummy": DummyModule()}):
```
snejus: Use `monkeypatch`

snejus: Or is this not possible using `monkeypatch`?

semohr: It seems not possible. `monkeypatch` does not work on class scope

semohr: This works but needs function scope instead of class scope.

```python
@pytest.fixture(autouse=True)
def _patch_dummy_module(self, monkeypatch):
    monkeypatch.setitem(sys.modules, "beetsplug.dummy", DummyModule())
```

snejus: Yeah I think this is cleaner!

semohr: Adjusted in latest commit :+1:

snejus (approved):

PR #6681, opened 2026-05-29
autotagging: introduce Source class for matching

### LINKED ISSUE #6684: Introduce `Source` Abstraction for Autotagger Context
## Summary
Introduce a `Source` abstraction that represents the tagging context for either an album or a singleton track.

`Source` should centralize shared lookup inputs currently derived in multiple places:
- entity type (`album` or `track`)
- current artist/name
- current metadata snapshot used for distance calculation
- items participating in matching
- discovered canonical ID and consensus state

This work corresponds to the `introduce-source` branch.

## Why
- Remove repeated "derive current context" logic.
- Make importer and session code consume the same shape for albums and singletons.
- Provide a stable input contract for `Candidates` and `Match.try_create`.

## Scope
- Add `Source` data model with constructors for album and singleton contexts.
- Replace direct per-call derivation of likely metadata with `Source` usage.
- Route candidate lookup to consume `Source` fields (`artist`, `name`, `id`, `id_consensus`, `items`, `data`).

## Acceptance Criteria
- [ ] Album tasks build source context via `Source.from_items(...)`.
- [ ] Singleton tasks build source context via `Source.from_item(...)`.
- [ ] Candidate resolution code consumes `Source` instead of ad hoc tuple/state passing.
- [ ] Existing lookup behavior for ID-first and text-search fallback is preserved.

snejus:
## `beets.autotag` Source-Centric Refactor

Fixes: #6684 

### Core change

This PR restructures autotagging around a new `beets.autotag.source.Source` object.

Before this change, album and singleton matching pulled "current metadata" from several places:
- importer tasks
- matching code
- distance calculation
- UI display code

After this change, those flows all receive the same input shape: a single `Source` value that carries:
- the original `items`
- the current display/search fields (`artist`, `name`)
- extracted metadata in `data`
- current external ID information (`id`, `id_consensus`)

That makes `Source` the new "current thing being matched" abstraction for both albums and tracks.

---

### 1. New single input object: `Source`

`Source` centralizes metadata extraction that used to be scattered across the autotag pipeline.

```text
Before
items
├─ importer computes current artist/album
├─ match code recomputes common tags
├─ distance recomputes common tags
└─ UI receives separate current values

After
Source
├─ source.artist
├─ source.name
├─ source.data
├─ source.items
├─ source.id / source.id_consensus
└─ source.va_likely
```

This is the biggest architectural shift in the PR.

### 2. `tag_album()` and `tag_item()` now operate on `Source`

The matching entry points now take a `Source` instead of a loose mix of:
- raw `items`
- `item`
- current artist/title values
- repeated ID lookups

That gives album and track matching the same calling convention and reduces special-case plumbing.

```python
# Before
tag_album(items, ...)
tag_item(item, ...)

# After
tag_album(source, ...)
tag_item(source, ...)
```

### 3. `distance()` is now narrower and more explicit

`distance()` no longer derives "current metadata" internally from a list of items.

Instead, it receives:
- `original`: the already-extracted metadata container
- `unmatched_count`: the number of unmatched local tracks

```text
Before
distance(items, album_info, item_info_pairs)
  -> recompute common tags from items
  -> infer unmatched track count

After
distance(source.data, album_info, item_info_pairs, unmatched_count)
  -> compare against already-prepared metadata
```

This is a good separation of responsibilities:
- `Source` prepares current metadata
- `distance()` only compares metadata

### 4. Importer tasks now cache the current source

`ImportTask` and `SingletonImportTask` now expose a cached `source` property:
- album tasks use `Source.from_items(...)`
- singleton tasks use `Source.from_item(...)`

That means importer state no longer needs separate `cur_artist` and `cur_album` fields, and downstream code can read from `task.source` instead.

### 5. UI/display code now depends on the same source model

Import session and display functions now use `task.source.artist` and `task.source.name` when showing candidate changes or prompting for manual search/ID entry.

This removes another copy of "current metadata" logic from the UI layer.

### 6. `AttrDict` moves to `beets.util`

`AttrDict` was previously defined inside `beets.autotag.hooks`. It now lives in `beets.util`.

High-level effect:
- metadata-container behavior is treated as shared infrastructure
- autotag-specific modules depend on a common utility instead of owning the abstraction

### 7. `beets.autotag` public API is flatter

Related cleanup continues by exporting more autotag types directly from `beets.autotag`, including `Source`.

This reduces import-path churn for callers and makes the package boundary easier to understand.

---

#### 1. One place defines "the current metadata we are matching from"
The main benefit is conceptual clarity.

Reviewers can now think about autotagging as:

```text
local files -> Source -> candidate lookup -> distance -> proposal/UI
```

instead of following several separate paths that each rediscover the same metadata.

#### 2. Album and singleton flows now look alike
Albums and single tracks now use the same structure and similar call patterns.

That should make future autotag changes easier because new behavior can usually be added once at the `Source` boundary rather than twice in parallel code paths.

#### 3. Less coupling between importer, matcher, distance logic, and UI
Previously, each layer needed to know how to reconstruct "current artist/album/common tags".

Now:
- importer builds `Source`
- matching uses `Source`
- UI displays from `Source`
- `distance()` receives prepared metadata

Each layer does less guessing about the others.

#### 4. Simpler data flow for debugging and future refactors
When a match looks wrong, there is now a clearer place to inspect the original input: `Source`.

That should help with debugging, logging, and future changes to metadata extraction rules.

---

## Data flow at a glance

```text
ImportTask / SingletonImportTask
        ↓
     `task.source`
        ↓
`tag_album(source)` / `tag_item(source)`
        ↓
candidate search + item/track assignment
        ↓
`distance(source.data, info, pairs, unmatched_count)`
        ↓
`Proposal`
        ↓
UI display and user choice
```

---

## Notable follow-on cleanup included here

- `get_most_common_tags()` now returns only the extracted metadata dictionary, not a `(likelies, consensus)` pair
- call sites that depended on consensus now compute only the specific signals they need, such as `id_consensus` or `va_likely`
- tests were updated to reflect the new `Source`-based flow
- `.git-blame-ignore-revs` was updated for mechanical refactor commits

---

## Reviewer guide

If reviewing this PR, the easiest way to read it is:

1. Start with `beets/autotag/source.py`
2. Then look at `tag_album()` and `tag_item()` in `beets/autotag/match.py`
3. Then review the `distance()` signature change
4. Finally scan importer/UI call sites to see how they now pass `task.source`

That path shows the architectural intent without getting lost in the mechanical updates.


beets/autotag/distance.py
```diff
@@ -427,9 +428,10 @@ def track_distance(
 
 
 def distance(
-    items: Sequence[Item],
+    original: AttrDict[Any],
```
semohr: I would assume this should take a `Source` object now instead of an `AttrDict[Any]`? 

The function arguments should only be these three in my opinion:
- Candidate (AlbumInfo)
- Source
- Assignment (item_info_pairs)

See [here](https://github.com/beetbox/beets/issues/6683#issuecomment-4586487275) for more information and my understanding of it.

beets/autotag/distance.py
```diff
@@ -427,9 +428,10 @@ def track_distance(
...
     album_info: AlbumInfo,
     item_info_pairs: list[tuple[Item, TrackInfo]],
+    unmatched_count: int,
 ) -> Distance:
     """Determines how "significant" an album metadata change would be.
```
semohr: Docstring needs an update

semohr:
Am having a bit of a hard time understanding the changes here since you also moved functions around quite a bit. For reviewing purpose, would you mind splitting the refactor into another PR?

snejus:
> Am having a bit of a hard time understanding the changes here since you also moved functions around quite a bit. For reviewing purpose, would you mind splitting the refactor into another PR?

Of course - I will split the first two commits into a separate PR.

snejus:
See https://github.com/beetbox/beets/pull/6687

beets/autotag/source.py
```diff
@@ -0,0 +1,57 @@
...
+        return f"{self.artist} - {self.name}"
+
+    @property
+    def va_likely(self) -> bool:
+        return len({i.artist for i in self.items}) > 1 or any(
+            i.comp for i in self.items
+        )
+
+    @classmethod
+    def from_items(cls, items: Sequence[Item]) -> Source:
```
semohr, line 33: Can we move the items creation into here? Maybe with a `from_path` alternate constructor? 

I can see this being a bit difficulties with integrating this into current task factory but might simplify some things further and would align with my understanding of source.

beets/autotag/distance.py
snejus: See part of the docstring:

```
    `original` is an AttrDict with the original album metadata.
```

It previously needed `items` in order to calculate `likelies`:

```py
    likelies, _ = get_most_common_tags(items)
```

Now, this calculation is done by `Source` and the data is assigned to an attribute `Source.data`. It is then fed into `distance`, since it doesn't need any other data.

I made it an `AttrDict` so that keys can be accessed as attribute, to make these calls consistent:

```py
    dist.add_string("album", original.album, album_info.album)
```

beets/autotag/distance.py
snejus: Hm it had been updated earlier, not sure how did you manage to comment under an _oudated_ version 😅 

```
    """Determine how "significant" an album metadata change would be.

    Returns a Distance object.
    `original` is an AttrDict with the original album metadata.
    `album_info` is an AlbumInfo object reflecting the album to be compared.
    `item_info_pairs` is a list with matched (Item, TrackInfo) pairs.
    `unmatched_count` is the number of unmatched tracks on the release.
    """
```

semohr: Started the review from another commit. Regard as resolved.

beets/autotag/distance.py
semohr: How about we make `Source.data` a cached property and give `Source` as parameter here than. Seems like a slightly more clear abstraction for my mental model.

semohr: Seems a bit weird that we need to precompute the `get_most_common_tags` in the `from_item` methods anyways.

beets/autotag/source.py
snejus: I think I kind of see the direction here. Are you saying we could get `Source` within `ImportTaskFactory.singleton` and `ImportTaskFactory.album`? And then replace `ImportTask.items` with `ImportTask.source` in the construction?

beets/autotag/distance.py
snejus: > How about we make `Source.data` a cached property and give `Source` as parameter here than. Seems like a slightly more clear abstraction for my mental model.

Two issues with this:

1. `data` is different for each `Source` type so we compute it either in `from_item` or `from_items`
2. `data` is needed in two places: for distance and in `ImportTask.chosen_info` so it does indeed need to be pre-computed

semohr: Both of that shouldn't be an issue anymore once the `ImportTask` and `SentienlImportTask` combine. This is one of the main goals here, right?

beets/autotag/source.py
semohr: Yes that is roughtly my idea. If we dont work with `items` but work with the `source` it should allow us to abstract most operations regardless of if we look at an album or singleton.

beets/autotag/distance.py
snejus: Indeed - but those will be combined in later PRs!

beets/autotag/source.py
snejus: Yup, this will get done but in later PRs.

beets/util/__init__.py
```diff
@@ -807,18 +809,11 @@ def plurality(objs: Iterable[T]) -> tuple[T, int]:
...
 
-def get_most_common_tags(
-    items: Sequence[Item],
-) -> tuple[dict[str, Any], dict[str, Any]]:
-    """Extract the likely current metadata for an album given a list of its
-    items. Return two dictionaries:
-     - The most common value for each field.
-     - Whether each field's value was unanimous (values are booleans).
-    """
+def get_most_common_tags(items: Sequence[Item]) -> dict[str, Any]:
```
semohr: I would maybe do something like this to get typing here:

```python
class Likelies(TypedDict, total=False):
    """Most likely values for a list of items"""

    artist: str
    album: str
    albumartist: str
    year: int
    disctotal: int
    mb_albumid: str
    label: str
    barcode: str
    catalognum: str
    country: str
    media: str
    albumdisambig: str
    data_source: str


def get_most_common_tags(items: Sequence[Item]) -> Likelies:
    """Extract the most common value for each field given a list of items."""
    assert items

    likelies: Likelies = {}

    for field in tuple(Likelies.__annotations__):
        values = [item.get(field) for item in items if item]
        value, _ = plurality(values)
        if value is not None:
            likelies[field] = value

    albumartist = likelies.get("albumartist")
    if len({i.albumartist for i in items}) == 1 and albumartist:
        likelies["artist"] = albumartist

    return likelies
```

I know we are currently not using it but it would be nice to have the counts in there and if it is a consensus value. I see this as an object we are going to pass to the `metadata_plugins` at some point.

```python

class MetadataPlugin:
    def candidates(items, likelies, search_query): ...
    
    # or
    
    def candidates(source, search_query): 
        likelies = source.likelies
```

snejus: I considered this but the biggest issue is that I wanted to feed `original` as `AttrDict` into `distance` to normalize attribute access within that function:

```diff
diff --git a/beets/autotag/distance.py b/beets/autotag/distance.py
index d5821f473..05647da21 100644
--- a/beets/autotag/distance.py
+++ b/beets/autotag/distance.py
@@ -9,13 +9,14 @@
 from unidecode import unidecode
 
 from beets import config, metadata_plugins
-from beets.util import as_string, cached_classproperty, get_most_common_tags
+from beets.util import as_string, cached_classproperty
 from beets.util.color import colorize
 
 if TYPE_CHECKING:
     from collections.abc import Iterator, KeysView, Sequence
 
     from beets.library import Item
+    from beets.util import AttrDict
     from beets.util.color import ColorName
 
     from .hooks import AlbumInfo, TrackInfo
@@ -427,28 +428,27 @@ def track_distance
 
 
 def distance(
-    items: Sequence[Item],
+    original: AttrDict[Any],
     album_info: AlbumInfo,
     item_info_pairs: list[tuple[Item, TrackInfo]],
+    unmatched_count: int,
 ) -> Distance:
-    """Determines how "significant" an album metadata change would be.
-    Returns a Distance object. `album_info` is an AlbumInfo object
-    reflecting the album to be compared. `items` is a sequence of all
-    Item objects that will be matched (order is not important).
-    `mapping` is a dictionary mapping Items to TrackInfo objects; the
-    keys are a subset of `items` and the values are a subset of
-    `album_info.tracks`.
-    """
-    likelies, _ = get_most_common_tags(items)
+    """Determine how "significant" an album metadata change would be.
 
+    Returns a Distance object.
+    `original` is an AttrDict with the original album metadata.
+    `album_info` is an AlbumInfo object reflecting the album to be compared.
+    `item_info_pairs` is a list with matched (Item, TrackInfo) pairs.
+    `unmatched_count` is the number of unmatched tracks on the release.
+    """
     dist = Distance()
 
     # Artist, if not various.
     if not album_info.va:
-        dist.add_string("artist", likelies["artist"], album_info.artist)
+        dist.add_string("artist", original.artist, album_info.artist)
 
     # Album.
-    dist.add_string("album", likelies["album"], album_info.album)
+    dist.add_string("album", original.album, album_info.album)
 
     preferred_config = config["match"]["preferred"]
     # Current or preferred media.
@@ -461,29 +461,29 @@ def distance
         if options:
             dist.add_priority("media", album_info.media, options)
         # Current media.
-        elif likelies["media"]:
-            dist.add_equality("media", album_info.media, likelies["media"])
+        elif original.media:
+            dist.add_equality("media", album_info.media, original.media)
 
     # Mediums.
-    if likelies["disctotal"] and album_info.mediums:
-        dist.add_number("mediums", likelies["disctotal"], album_info.mediums)
+    if original.disctotal and album_info.mediums:
+        dist.add_number("mediums", original.disctotal, album_info.mediums)
 
     # Prefer earliest release.
     if album_info.year and preferred_config["original_year"]:
         # Assume 1889 (earliest first gramophone discs) if we don't know the
         # original year.
-        original = album_info.original_year or 1889
-        diff = abs(album_info.year - original)
-        diff_max = abs(datetime.date.today().year - original)
+        original_year = album_info.original_year or 1889
+        diff = abs(album_info.year - original_year)
+        diff_max = abs(datetime.date.today().year - original_year)
         dist.add_ratio("year", diff, diff_max)
     # Year.
-    elif likelies["year"] and album_info.year:
-        if likelies["year"] in (album_info.year, album_info.original_year):
+    elif original.year and album_info.year:
+        if original.year in (album_info.year, album_info.original_year):
             # No penalty for matching release or original year.
             dist.add("year", 0.0)
         elif album_info.original_year:
             # Prefer matchest closest to the release year.
-            diff = abs(likelies["year"] - album_info.year)
+            diff = abs(original.year - album_info.year)
             diff_max = abs(
                 datetime.date.today().year - album_info.original_year
             )
@@ -498,30 +498,28 @@ def distance
     if album_info.country and options:
         dist.add_priority("country", album_info.country, options)
     # Country.
-    elif likelies["country"] and album_info.country:
-        dist.add_string("country", likelies["country"], album_info.country)
+    elif original.country and album_info.country:
+        dist.add_string("country", original.country, album_info.country)
 
     # Label.
-    if likelies["label"] and album_info.label:
-        dist.add_string("label", likelies["label"], album_info.label)
+    if original.label and album_info.label:
+        dist.add_string("label", original.label, album_info.label)
 
     # Catalog number.
-    if likelies["catalognum"] and album_info.catalognum:
+    if original.catalognum and album_info.catalognum:
         dist.add_string(
-            "catalognum", likelies["catalognum"], album_info.catalognum
+            "catalognum", original.catalognum, album_info.catalognum
         )
 
     # Disambiguation.
-    if likelies["albumdisambig"] and album_info.albumdisambig:
+    if original.albumdisambig and album_info.albumdisambig:
         dist.add_string(
-            "albumdisambig", likelies["albumdisambig"], album_info.albumdisambig
+            "albumdisambig", original.albumdisambig, album_info.albumdisambig
         )
 
     # Album ID.
-    if likelies["mb_albumid"]:
-        dist.add_equality(
-            "album_id", likelies["mb_albumid"], album_info.album_id
-        )
+    if original.mb_albumid:
+        dist.add_equality("album_id", original.mb_albumid, album_info.album_id)
 
     # Tracks.
     dist.tracks = {}
@@ -534,9 +532,9 @@ def distance
         dist.add("missing_tracks", 1.0)
 
     # Unmatched tracks.
-    for _ in range(len(items) - len(item_info_pairs)):
+    for _ in range(unmatched_count):
         dist.add("unmatched_tracks", 1.0)
 
-    dist.add_data_source(likelies["data_source"], album_info.data_source)
+    dist.add_data_source(original.data_source, album_info.data_source)
 
     return dist
```

snejus: But it seems like we can do this? 

```diff
diff --git a/beets/autotag/source.py b/beets/autotag/source.py
index 4f1b222e2..06e89b4ee 100644
--- a/beets/autotag/source.py
+++ b/beets/autotag/source.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Literal, NamedTuple
+from typing import TYPE_CHECKING, Literal, NamedTuple
 
-from beets.util import AttrDict, get_most_common_tags
+from beets.util import Likelies, get_most_common_tags
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
@@ -14,7 +14,7 @@ class Source
     type: Literal["album", "track"]
     artist: str
     name: str
-    data: AttrDict[Any]
+    data: Likelies
     items: Sequence[Item]
     id: str
     id_consensus: bool
@@ -35,11 +35,11 @@ def from_items
         likelies = get_most_common_tags(items)
         return cls(
             type="album",
-            artist=likelies["artist"],
-            name=likelies["album"],
-            data=AttrDict(likelies),
+            artist=likelies.artist,
+            name=likelies.album,
+            data=likelies,
             items=items,
-            id=likelies["mb_albumid"],
+            id=likelies.mb_albumid,
             id_consensus=len({i.mb_albumid for i in items}) == 1,
         )
 
@@ -50,7 +50,7 @@ def from_item
             type="track",
             artist=item.artist,
             name=item.title,
-            data=AttrDict(item),
+            data=Likelies(item),  # type: ignore[arg-type]
             items=[item],
             id=item.mb_trackid,
             id_consensus=True,
diff --git a/beets/util/__init__.py b/beets/util/__init__.py
index 9af5138ea..9bd8127fb 100644
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -809,7 +809,7 @@ def plurality
     return c.most_common(1)[0]
 
 
-def get_most_common_tags(items: Sequence[Item]) -> dict[str, Any]:
+def get_most_common_tags(items: Sequence[Item]) -> Likelies:
     """Extract the most common value for each field given a list of items."""
     assert items  # Must be nonempty.
 
@@ -837,7 +837,7 @@ def get_most_common_tags
     if len({i.albumartist for i in items}) == 1 and likelies["albumartist"]:
         likelies["artist"] = likelies["albumartist"]
 
-    return likelies
+    return Likelies(likelies)
 
 
 # stdout and stderr as bytes
@@ -1238,3 +1238,21 @@ def copy
 
     def __hash__(self) -> int:  # type: ignore[override]
         return id(self)
+
+
+class Likelies(AttrDict[str | None]):
+    """A dictionary of the most common tags in a list of items."""
+
+    artist: str
+    album: str
+    albumartist: str
+    year: str
+    disctotal: str
+    mb_albumid: str
+    label: str
+    barcode: str
+    catalognum: str
+    country: str
+    media: str
+    albumdisambig: str
+    data_source: str | None
```

semohr: Can we make the `Likelies` a frozen dataclass or is there a specific reason it needs to be a mutable `AttrDict`?

snejus: A frozen dataclass does sound like a good fit for values that are calculated once and only read afterwards. I'm just a little hesitant to introduce another metadata container alongside `Item`, `AttrDict`, `Source`, `AlbumInfo`, and `TrackInfo`, particularly if we then need conversions between them.

On the other hand, I am more than happy to add `Likelies` as suggested above. That would give us typed attribute access while retaining the existing mapping behavior and avoiding a separate representation. It would also let us pass the same typed mapping directly into `beets.autotag.distance.distance`. If we later pass this data to metadata plugins, that concrete API might give us a better basis for deciding whether a different container would be worthwhile. What do you think?

semohr: I was mostly asking because I'm not a big fan of `AttrDict` in general. Since it's backed by a `dict`, it still allows arbitrary attributes to be added dynamically, which isn't really what we want here.

That said, I'm happy with introducing `Likelies`.

To me, it also makes sense to have different containers for different concepts. I see the hierarchy roughly like this:
* `Source`: wraps the raw files on disk.
* `Item`: the canonical abstraction for track metadata (whether it comes from the filesystem or the library database).
* `AlbumInfo` / `TrackInfo`: metadata returned from remote providers.

`Likelies` feels like a different kind of object. It doesn't represent metadata itself, but the aggregated "likely" values derived from a group of `Item`s (e.g., the majority value for album artist, year, etc.). Having a dedicated type for that seems reasonable to me.

beets/util/__init__.py
```diff
@@ -1199,3 +1193,66 @@ def chunks(lst: Sequence[T], n: int) -> Iterator[list[T]]:
...
+        )
+
+    def __setattr__(self, key: str, value: T) -> None:
+        self.__setitem__(key, value)
+
+    def __hash__(self) -> int:  # type: ignore[override]
+        return id(self)
+
+
+class Likelies(AttrDict[Any]):
```
semohr, line 1250: What is the reason again for using an `AttrDict` here instead of a `dataclass`? Since we are not expecting dynamic keys or values in this case, I don't see a strong reason to keep this as an `AttrDict`. A `dataclass` would make the structure clearer and avoid the overhead of dictionary-based attribute access.

snejus: Hmm `AttrDict` pre-dates this change, and I don't want to adjust it in this PR - could we discuss it separately? I think the main reason for it is that it emulates `Model`. In addition to this, we need to be able to set flexible attributes here.

snejus:
@semohr would you mind resolving your threads and approving if it looks good?

beets/util/__init__.py
semohr: I dont think we need to be able to set flexible attributes on `Likelies`. At-least on construction we only allow use fixed attributes.


```python
# somewhere above
 fields = [
        "artist",
        "album",
        "albumartist",
        "year",
        "disctotal",
        "mb_albumid",
        "label",
        "barcode",
        "catalognum",
        "country",
        "media",
        "albumdisambig",
        "data_source",
    ] 
```

We introduced `Likelies` in this PR so it would make sense for me to not us `AttrDict` if not necessary. If you are set on keeping it for now, let's create an issue so we dont forget to refactor this.

snejus: See https://github.com/beetbox/beets/issues/6920

PR #6689, opened 2026-05-30
autotagging: make Change polymorphic

### LINKED ISSUE #6688: Consolidate `show_change` and `show_item_change` into a Single Display API
## Summary
Consolidate the importer change preview entrypoints `show_change` and `show_item_change` into one unified method that renders both album and singleton match previews.

## Why
Current display flow duplicates orchestration across two public functions that perform the same high-level job: render a preview of proposed metadata changes before user confirmation. A single entrypoint will reduce duplication and keep preview behavior consistent across albums and singletons.

## Problem
- Two parallel entrypoints increase maintenance cost.
- Shared logic is split between album and singleton wrappers.
- Behavioral adjustments to preview formatting require touching both paths.
- Importer/session code must branch only to select display function.

## Scope
- Introduce one public display function for match preview rendering.
- Route album and singleton previews through that function.
- Keep type-specific rendering details encapsulated in existing representation classes (or equivalent split internals).
- Update importer/session call sites to use the unified API.
- Preserve existing output formatting unless explicitly changed.

## Non-Goals
- Redesign preview layout, colors, or prompt UX.
- Remove internal album/singleton representation classes if they still provide value.

## Acceptance Criteria
- [ ] One public preview entrypoint replaces `show_change` and `show_item_change` at call sites.
- [ ] Album and singleton flows produce equivalent preview output to current behavior.
- [ ] No importer prompt behavior regression in interactive mode.
- [ ] Tests cover both album and singleton preview paths through the unified function.

## Validation
- Targeted importer UI tests pass for album and singleton confirmation flows.
- Existing autotagger/importer suite passes without preview regressions.

snejus:
Fixes #6688

This change simplifies import preview rendering by routing both album and track matches through a single `show_change()` flow.

#### Architectural impact
- Introduces a shared `Change.from_match()` factory that selects `AlbumChange` or `TrackChange` based on the match type.
- Moves preview rendering behind polymorphic `show()` methods, so each change type owns its own display behavior.
- Removes separate track-specific preview handling like `show_item_change()` and the related branching in `session.py`.

#### High-level effect
- Centralizes change-preview logic in `beets/ui/commands/import_/display.py`.
- Makes `session.py` simpler by using one preview entry point for both albums and tracks.
- Reduces duplicate logic and makes future display changes easier to implement consistently.

In short: this refactor does not change the import flow itself, but it makes the change-display architecture more uniform, easier to follow, and easier to extend.


beets/ui/commands/import_/display.py
```diff
@@ -361,36 +376,18 @@ def show_match_tracks(self) -> None:
...
-    metadata from `match`, a TrackMatch object.
-    """
-    change = TrackChange(source.artist, source.name, match)
-    # Print the match header.
-    change.show_match_header()
-    # Print the match details.
-    change.show_match_details()
+def show_change(source: Source, match: Match) -> None:
+    """Print out a representation of the changes."""
+    Change.from_match(match, source).show()
```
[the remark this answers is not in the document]
snejus: I'm refactoring existing functionality so I'm not going to add new tests here.

beets/ui/commands/import_/display.py
```diff
@@ -36,6 +36,16 @@ class ChangeRepresentation:
     original_name: str
     match: Match
 
+    @classmethod
+    def from_match(cls, match: Match, source: Source) -> Change:
```
semohr, line 40: It still feels a bit weird architecturally that the parent class now knows about the child classes. Seems awkward to have that coupling, and it might mean touching the parent. Any other way to solve this? Maybe through a visitor pattern?

snejus: Agreed. I removed `Change.from_match()` and used `singledispatch` instead. A visitor seems like too much machinery for dispatching two match types. Does this look better?

PR #6692, opened 2026-05-30
template evaluation: feed all templates into evaluate_fmt as strings

snejus:
This change standardizes template handling around `evaluate_fmt`, so callers now pass plain format strings everywhere instead of a mix of strings and precompiled `Template` objects.

At a high level, it:
- centralizes template compilation in `get_template`
- keeps `path_formats` as simple `(query, "format string")` pairs
- removes template parsing from models, commands, plugins, and tests

Architecturally, this makes formatting flow through one consistent path: string in, `evaluate_fmt` handles compilation and evaluation. The result is a simpler API, fewer special cases at call sites, and easier reasoning about how path and display formatting works across the codebase.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -48,8 +48,9 @@ from typing_extensions import (
 from unidecode import unidecode
 
 import beets
+from beets.util import cached_classproperty
+from beets.util.functemplate import get_template
 
-from ..util import cached_classproperty, functemplate
 from . import types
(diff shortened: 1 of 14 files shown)


semohr:
What about #6737?

snejus:
Oops, I didn't realise that I opened this one earlier - I wanted to avoid breaking the compatibility with `evaluate_fmt` rename.

PR #6695, opened 2026-05-31
Return keysview from keys methods

snejus:
### PR summary

This change makes field/key handling more consistent and better typed across the model layer.

- `Model.keys()` and `all_keys()` now return `KeysView`-style set-like views instead of lists, and related code is updated to use set operations like `|` and `sorted(...)`.
- `FormattedMapping` and `FormattedItemMapping` now store included keys as sets internally, which matches the new key semantics and simplifies membership checks.
- The formatter plumbing in `beets.dbcore.db` is tightened for mypy: `_formatter` is explicitly typed, `formatted()` now forwards arguments generically, and `Album` opts into the base `FormattedMapping` explicitly.

### High-level impact

- Standardizes the architecture around "keys are a set of field names", not an ordered mutable list.
- Reduces duplicate/merge logic at call sites and makes intent clearer.
- Fixes typing issues without changing the user-facing behavior of commands like `fields`; most changes are internal API cleanup and consistency work.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -736,11 +736,12 @@ class Model(ABC, Generic[D]):
             self.store()
 
     # Formatting and templating.
-
-    _formatter = FormattedMapping
+    _formatter: type[FormattedMapping]
 
     def formatted(
-        self, included_keys: str = _formatter.ALL_KEYS, for_path: bool = False
+        self,
+        included_keys: str = FormattedMapping.ALL_KEYS,
+        for_path: bool = False,
     ) -> FormattedMapping:
         """Get a mapping containing all values on this object formatted
         as human-readable unicode strings.
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -60,6 +60,7 @@ if TYPE_CHECKING:
         Generator,
         Iterable,
         Iterator,
+        KeysView,
         Sequence,
     )
     from sqlite3 import Connection
@@ -112,22 +113,24 @@ class FormattedMapping(Mapping[str, str]):
     """
 
     model: Model
+    model_keys: set[str]
 
     ALL_KEYS = "*"
 
     def __init__(
         self,
         model: Model,
-        included_keys: str = ALL_KEYS,
+        included_keys: str | list[str] = ALL_KEYS,
         for_path: bool = False,
-    ):
+    ) -> None:
         self.for_path = for_path
         self.model = model
-        if included_keys == self.ALL_KEYS:
+        self.model_keys = set(
             # Performance note: this triggers a database query.
-            self.model_keys = self.model.keys(True)
-        else:
-            self.model_keys = included_keys
+            self.model.keys(True)
+            if included_keys == self.ALL_KEYS
+            else included_keys
+        )
 
     def __getitem__(self, key: str) -> str:
         if key in self.model_keys:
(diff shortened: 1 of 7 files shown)


beets/dbcore/db.py
```diff
@@ -112,22 +114,23 @@ class FormattedMapping(Mapping[str, str]):
     """
 
     model: Model
+    model_keys: AbstractSet[str]
```
semohr: Why not just `set`?

beets/dbcore/db.py
```diff
@@ -733,16 +736,13 @@ def add(self, db: D | None = None):
...
 
     # Formatting and templating.
+    _formatter: type[FormattedMapping]
 
-    _formatter = FormattedMapping
-
-    def formatted(
-        self, included_keys: str = _formatter.ALL_KEYS, for_path: bool = False
-    ) -> FormattedMapping:
+    def formatted(self, *args, **kwargs) -> FormattedMapping:
```
semohr: Can we keep the explicit typing for the arguments, please?

Using `*args` and `**kwargs` makes it much harder to see what the actual function parameters are. It also weakens static analysis. I'm fairly sure mypy now treats these as `Any` in many cases, which means we lose type checking and API validation in some places.

For the "beets as a library" use case, `*args` and `**kwargs` are particularly inconvenient :/  Instead of being able to inspect a function signature directly, I have to dig into the implementation to determine which arguments are supported. Explicit signatures also allow type checkers to catch API changes early, whereas variadic arguments make those changes much easier to miss.

semohr (changes requested):
Just had a short look. Could you give a more specific reasoning why we need this? Is it needed for changes we have planned or simplifies them?

I get it as a cleanup step and it makes our code more "pythonic" (am not opposed to it necessarily) but it adds a bit of complexity which I would try avoid if it is not strictly necessary.

beets/dbcore/db.py
semohr: Maybe we dont really need/use the arguments at all and we can just remove them :thinking:

snejus:
> Just had a short look. Could you give a more specific reasoning why we need this? Is it needed for changes we have planned or simplifies them?

I was adding types to `beets.importer` module and ended up having to make `FormattedMapping` inherit from `Mapping` (as the TODO above outlines as well):

```diff

 fd121d626c7a23244fd69bb177f56fe43fa30f8a 
 2026-05-30 18:02 / Šarūnas Nejus <snejus@protonmail.com>
 2026-05-31 08:23 / Šarūnas Nejus <snejus@protonmail.com>
 
    
    Add missing types to importer.tasks



diff --git a/beets/dbcore/db.py b/beets/dbcore/db.py
index cc882eae5..b988211d7 100755
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -171,23 +171,23 @@ def _get_formatted
 
 
 # NOTE: This seems like it should be a `Mapping`, i.e.
 # ```
 # class LazyConvertDict(Mapping[str, Any])
 # ```
 # but there are some conflicts with the `Mapping` protocol such that we
 # can't do this without changing behaviour: In particular, iterators returned
 # by some methods build intermediate lists, such that modification of the
 # `LazyConvertDict` becomes safe during iteration. Some code does in fact rely
 # on this.
-class LazyConvertDict:
+class LazyConvertDict(Mapping[str, Any]):
     """Set the base data that should be lazily converted"""
```

And then `mypy` started complaining:

```
$ mypy beets/dbcore/db.py
beets/dbcore/db.py:222: error: Return type "list[str]" of "keys" incompatible with return type "KeysView[str]" in supertype "typing.Mapping"  [override]
beets/dbcore/db.py:240: error: Return type "Iterable[tuple[str, Any]]" of "items" incompatible with return type "ItemsView[str, Any]" in supertype "typing.Mapping"  [override]
Found 2 errors in 1 file (checked 1 source file)
```

beets/dbcore/db.py
snejus: Good point, made it a `set` with a small change in the constructor.

beets/dbcore/db.py
snejus: The only reason I removed this is that this method depended on `_formatter.ALL_KEYS` for default values. Since `_formatter` now defines just an expected type, I couldn't use `_formatter` here.

But now I'm looking at it again and I see I just had to replace it with `FormattedMapping.ALL_KEYS` 🤦🏼

semohr (approved):
Alright! Thanks for the explanation.

PR #6702, opened 2026-06-03
importer: improve handling of duplicates

snejus:
This change centralizes duplicate handling in the import pipeline.

- `ImportTask` now carries a single `duplicate_action` state instead of separate duplicate flags, so duplicate decisions are represented in one place.
- `ImportSession.get_duplicate_action()` becomes the shared entry point for resolving the configured duplicate policy.
- `TerminalImportSession` now only prompts the user when the configured action is `ASK`; otherwise it reuses the shared session logic.
- Duplicate resolution in `beets/importer/stages.py` is simplified to call the session API instead of re-implementing config lookup and branching.

## High-level impact

- Makes duplicate behavior more consistent across importer stages and UI sessions.
- Reduces duplicated logic between core importer flow, terminal prompting, and test fixtures.
- Simplifies future changes to duplicate handling because policy resolution now lives behind one session-level interface.
- Cleans up tests to use real config-driven duplicate behavior instead of custom fixture-only overrides.

### DIFF
--- a/beets/ui/commands/import_/session.py
+++ b/beets/ui/commands/import_/session.py
@@ -12,6 +12,7 @@ from beets.autotag import (
     tag_album,
     tag_item,
 )
+from beets.importer import DuplicateAction
 from beets.util import PromptChoice, displayable_path
 from beets.util.color import colorize
 from beets.util.units import human_bytes, human_seconds_short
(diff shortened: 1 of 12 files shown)


beets/ui/commands/import_/session.py
```diff
@@ -143,66 +149,57 @@ def choose_item(self, task):
...
-    def resolve_duplicate(self, task, found_duplicates):
+    def report_item_summary(
+        self, prefix: Literal["Old", "New"], items: list[Item], is_album: bool
+    ) -> None:
+        ui.print_(f"{prefix}: {summarize_items(items, not is_album)}")
+        if self.config["duplicate_verbose_prompt"].get(bool):
+            for dup in items:
+                print(f"  {dup}")
+
+    def get_duplicate_action_from_user(
```
semohr: Can we make this a private method? Same for `report_item_summary`. They are not used outside of the `TerminalImportSession`.

semohr (changes requested):
Very nice! Please add a changelog entry 🙃

snejus:
> Please add a changelog entry

Is there anything relevant to users or maintainers?

semohr:
> > Please add a changelog entry
> 
> Is there anything relevant to users or maintainers?

Mainly for me to remember this change tbh. It will require me to fix some stuff once it hits a release.

semohr (approved):
Thanks 🙏

PR #6704, opened 2026-06-03
initial on branch

tyson312:
## Summary

Adds the **nonsense** plugin, which prints a random quote after every beets command completes. It hooks into the existing `cli_exit` event (same pattern as `hook`, `playlist`, and `mpdupdate`), so a command like `beet ls someTerm` shows query results first, then a quote on the next line.

- New plugin module: `beetsplug/nonsense.py`
- Optional dependency via `pip install "beets[nonsense]"` (`quotes-generator`)
- Configurable quote source (`nonsense.source`) or random category by default
- Docs, changelog entry, and 5 unit tests (mocked, no network)

## Test plan

- [x] `poetry install -E nonsense && poetry run pytest test/plugins/test_nonsense.py -v` — all 5 tests pass
- [ ] Enable `plugins: nonsense` in config, run `beet ls` and confirm a quote prints after output
- [ ] Set `nonsense.source: motivational` and confirm quotes come from that category
- [ ] Without `quotes-generator` installed, confirm plugin loads with a warning and no crash


tyson312:
sorry not a real thing. Please disregard

PR #6709, opened 2026-06-04
tests: replace function-scoped helper fixtures with TestHelper

snejus:
- Test setup was consolidated into shared helper classes instead of using local `helper` fixtures.
- `test/library/test_migrations.py` now uses `MigrationTestHelper` to provide one common migration harness, with each test class defining only the legacy state and migration it needs.
- Plugin tests in `test/plugins/test_fuzzy.py`, `test/plugins/test_missing.py`, and `test/plugins/utils/test_playcount.py` now rely on shared helpers like `PluginTestHelper` and `TestHelper` for common setup.

**Note**: I kept these fixtures in these files
* `test_query.py`
* `plugins/test_random.py`
* `plugins/test_aura.py`
* `plugins/test_lyrics.py`

because they intentionally use a higher `scope`, for example in `test_query.py`:

```py
@pytest.fixture(scope="class")
def helper():
    helper = TestHelper()
    helper.setup_beets()

    yield helper

    helper.teardown_beets()
```

Using this setup pattern in `test_query.py` significantly improved testing durations.

#### Architecture impact

- Test infrastructure now follows a clearer base-class pattern for shared setup.
- Migration tests cleanly separate 'prepare old state' from 'run migration and verify result'.
- Plugin tests now use the same setup path, which makes new tests easier to add and keeps test structure more consistent across files.

#### High-level outcome

- Less duplicated fixture boilerplate.
- More consistent and easier-to-read test architecture.
- Tests stay focused on behavior being verified, which should make maintenance and future extension simpler.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -6,7 +6,7 @@ import pytest
 from beets.autotag import Distance
 from beets.dbcore.query import Query
 from beets.test._common import DummyIO
-from beets.test.helper import RUNNING_IN_CI, ConfigMixin
+from beets.test.helper import RUNNING_IN_CI, ConfigMixin, TestHelper
 from beets.test.helper import is_importable as check_import
 from beets.util import cached_classproperty
 
@@ -129,3 +129,45 @@ def is_importable():
     """Fixture that provides a function to check if a module can be imported."""
 
     return check_import
+
+
+# Shared fixtures amortize the expensive TestHelper setup across multiple tests.
+#
+# TestHelper resets and reloads the beets configuration, so recreating it for
+# every test can slow large suites noticeably.
+#
+# Inheriting from TestHelper gives each test function isolated state. Use the
+# fixtures below instead when a broader scope is safe and the suite benefits
+# from reusing the same helper instance.
+@pytest.fixture(scope="session")
+def session_helper():
+    """Share beets test state across the full test session.
+
+    Use this for suites that tolerate shared library contents and global
+    configuration. Tests should target specific records rather than assume a
+    completely fresh overall state.
+    """
+    with TestHelper() as helper:
+        yield helper
+
+
+@pytest.fixture(scope="module")
+def module_helper():
+    """Share beets test state within one test module.
+
+    Use this when tests in the same file can reuse setup and side effects, but
+    later modules should still begin from a clean environment.
+    """
+    with TestHelper() as helper:
+        yield helper
+
+
+@pytest.fixture(scope="class")
+def class_helper():
+    """Share beets test state within one test class.
+
+    Use this when methods in a class can build on the same setup, while nearby
+    classes still need independent libraries, files, or configuration.
+    """
+    with TestHelper() as helper:
+        yield helper
--- a/test/library/test_migrations.py
+++ b/test/library/test_migrations.py
@@ -8,14 +8,63 @@ from beets.dbcore import types
 from beets.library import migrations
 from beets.library.models import Album, Item
 from beets.test.helper import TestHelper
-from beets.util import path_as_posix
+from beets.util import cached_classproperty, path_as_posix
+
+Migration = tuple[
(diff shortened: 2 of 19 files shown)


test/library/test_migrations.py
```diff
@@ -220,37 +244,25 @@ def test_migrate(self, helper: TestHelper, is_importable):
...
-        monkeypatch.setattr(
-            "beets.library.library.Library._migrations",
-            ((migrations.RelativePathMigration, (Item,)),),
-        )
-        yield helper
+class TestRelativePathMigration(MigrationTestHelper):
+    """Verify stored item paths are rewritten to library-relative values."""
 
-        helper.teardown_beets()
+    migration = (migrations.RelativePathMigration, (Item,))
```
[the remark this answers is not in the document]
snejus: I'm only refactoring tests and not adjusting what they do.

test/plugins/test_aura.py
```diff
@@ -7,21 +7,16 @@
...
     from flask.testing import Client
 
 
-@pytest.fixture(scope="session", autouse=True)
-def helper():
-    helper = TestHelper()
-    helper.setup_beets()
-    yield helper
-    helper.teardown_beets()
+@pytest.fixture(scope="session")
```
semohr, line 14: Why the indirection? Couldn't we use `session_helper` directly?

test/plugins/test_lyrics.py
```diff
@@ -46,11 +46,8 @@
 
 
 @pytest.fixture(scope="module")
-def helper():
-    helper = TestHelper()
-    helper.setup_beets()
-    yield helper
-    helper.teardown_beets()
+def helper(module_helper):
```
semohr, line 49: Same here again some unneeded indirection imo. Im also wondering now why we need the helper on different scopes in the first place. Have you tried session scope here? Do we have sideffect here which we dont want to propagate to the session scope?

test/conftest.py
```diff
@@ -129,3 +129,21 @@ def is_importable():
     """Fixture that provides a function to check if a module can be imported."""
 
     return check_import
+
+
+@pytest.fixture(scope="session")
```
semohr, line 142: Maybe we add a bit of doc strings here on when we expect them to be used?

semohr (commented):
I like the introduction of these `<scope>_helper` fixtures in general. I'm just not entirely sure how the scopes were chosen based on where the fixtures are actually used. To me, it looks like most of them could probably use `session` scope. 🤔

I did a quick experiment and changed the fixture in `test_random` to `session`, and I didn't run into any issues. Maybe there's a reason for the more restrictive scopes that I'm missing?

test/plugins/test_aura.py
snejus: The indirection is intentional: it keeps each test module using the stable local `helper` fixture name while delegating lifecycle/scope to the shared helper fixtures. That means the rest of the module does not need to know whether the backing helper is session-, module-, or class-scoped.

If we used `session_helper` directly, that scope detail would spread through the module and future scope changes would require broader fixture/test renames. With this shape, changing scope requires adjusting these 3 lines only:

```py
@pytest.fixture(scope="session")
def helper(session_helper):
    return session_helper
```

test/plugins/test_lyrics.py
snejus: Added docstrings to each of those

test/conftest.py
```diff
@@ -129,3 +129,45 @@ def is_importable():
...
+    """
+    with TestHelper() as helper:
+        yield helper
+
+
+@pytest.fixture(scope="module")
+def module_helper():
+    """Share beets test state within one test module.
+
+    Use this when tests in the same file can reuse setup and side effects, but
```
semohr, line 158: Lovely. Thanks!

semohr (approved):

Issue #6629, opened 2026-05-12
Proposal: better import mode configuration


ShimmerGlass:
This issue aims to discuss the improvement of the UX for configuring file operations during beets import.

# Background

Beets supports many ways of importing files:

- use the files in their current location (in place)
- copy
- move
- link (symlink)
- hardlink
- reflink

To configure them, we have as many [boolean config options](https://beets.readthedocs.io/en/stable/reference/config.html#importer-options).
Some of these modes can be overridden via the CLI with [flags](https://beets.readthedocs.io/en/stable/reference/cli.html#import).

# Problems

While modes are mutually exclusive (it does not make sense to both copy and move, or both move and hardlink), we allow enabling multiple of them at the same time.
This leads to the following issues:

## Unclear behavior

Beets does not complain when multiple modes are enabled at the same time. It simply picks one using a [priority system](https://github.com/beetbox/beets/blob/master/beets/importer/session.py#L125) without telling the user. This can lead users to believe one mode is active while beets silently applies another. This behavior is partially documented, but not for all options. For example:

> **reflink**: [...] The option is ignored if move is enabled (i.e., beets can move or copy files but it doesn’t make sense to do both).

This is not quite true, it will be also ignored if link or hardlink are enabled.

Figuring out what beets will do requires factoring in the defaults and priorities. This is even more confusing when using CLI flags as one needs to know what exactly is in their config when running the command and factor that in as well.

## Configuration bloat

Each of these modes have their own config option, making the config system and the documentation more complex and verbose.

This complexity is even more pronounced in the CLI where we need both an ON and OFF flag for each of the modes we want to support.

## In place, the forgotten child

When all modes are turned off, beets just uses the files where they are, without copying, moving, or anything.

Users may assume that one of the options needs to be enabled and not realize that turning them all off enables another, "hidden" mode: use the original files in place.

# Proposal

Replace current boolean flag with a single enum:

- `inplace`
- `copy` (default)
- `move`
- `link`
- `hardlink`
- `reflink`

Exactly one mode may be selected.

### Benefits

- **clarity**: only one mode can be configured at a time, CLI overrides are way simpler. The precedence system is removed entirely.
- **discoverability**: users can now see at a glance in the documentation or in the CLI help what are their options.
- **validation**: we now have a clear set of valid values, everything else is an error.
- **documentation**: documentation is easier to write now that everything is a single option
[...]

PR #6711, opened 2026-06-05
typing: add missing types to beets.importer

snejus:
- This change tightens typing across the importer pipeline, mainly in `beets.importer.session`, `beets.importer.stages`, `beets.importer.tasks`, `beets.importer.state`, and `beets.util.pipeline`.

- Architecturally, it makes the importer's core flow more explicit:
  - session methods now declare clear input/output types
  - pipeline stage message types are named and propagated through `Pipeline`
  - task classes expose more precise return values and argument types
  - plugin event dispatch for `import_task_created` now has a typed contract

- A small but important part of the change is switching importer config handling to work with `confuse` views more safely, using `.set(...)` and `.get(...)` instead of replacing values directly. This keeps config mutation aligned with the config system instead of treating it like a plain dict.

- There are also a few correctness-oriented fixes uncovered while adding types, such as:
  - keeping byte/string path handling consistent in importer task cleanup and archive extraction
  - making nullable values explicit before use
  - returning `None` in places that already behaved that way implicitly

- High-level impact: no feature change to importer behavior is intended. The main benefit is a clearer, more strongly typed importer architecture that is easier to reason about, safer to refactor, and better at catching edge cases during development.

### DIFF
--- a/beets/importer/session.py
+++ b/beets/importer/session.py
@@ -25,15 +25,16 @@ from .actions import Action, DuplicateAction
 from .state import ImportState
 
 if TYPE_CHECKING:
-    from collections.abc import Sequence
+    from collections.abc import Iterator, Sequence
 
     import confuse
 
     from beets import dbcore, library
+    from beets.autotag import AlbumMatch, TrackMatch
     from beets.library import AnyLibModel
     from beets.util import PathBytes
 
-    from .tasks import ImportTask
+    from .tasks import ImportTask, SingletonImportTask
 
 
 QUEUE_SIZE = 128
@@ -65,7 +66,7 @@ class ImportSession:
         loghandler: logging.Handler | None,
         paths: Sequence[PathBytes] | None,
         query: dbcore.Query | None,
-    ):
+    ) -> None:
         """Create a session.
 
         Parameters
@@ -90,7 +91,9 @@ class ImportSession:
         # Normalize the paths.
         self.paths = list(map(normpath, paths or []))
 
-    def _setup_logging(self, loghandler: logging.Handler | None):
+    def _setup_logging(
+        self, loghandler: logging.Handler | None
+    ) -> logging.BeetsLogger:
         logger = logging.getLogger(__name__)
         logger.propagate = False
         if not loghandler:
@@ -149,13 +152,13 @@ class ImportSession:
 
         self.want_resume = config["resume"].as_choice([True, False, "ask"])
 
-    def tag_log(self, status, paths: Sequence[PathBytes]):
+    def tag_log(self, status: str, paths: Sequence[PathBytes]) -> None:
         """Log a message about a given album to the importer log. The status
         should reflect the reason the album couldn't be tagged.
         """
         self.logger.info("{} {}", status, displayable_path(paths))
 
-    def log_choice(self, task: ImportTask, duplicate=False):
+    def log_choice(self, task: ImportTask, duplicate: bool = False) -> None:
         """Logs the task's current choice if it should be logged. If
         ``duplicate``, then this is a secondary choice after a duplicate was
         detected and a decision was made.
@@ -176,10 +179,10 @@ class ImportSession:
             elif task.skip:
                 self.tag_log("skip", paths)
 
-    def should_resume(self, path: PathBytes):
+    def should_resume(self, path: PathBytes) -> bool:
         raise NotImplementedError
 
-    def choose_match(self, task: ImportTask):
+    def choose_match(self, task: ImportTask) -> AlbumMatch | Action:
         raise NotImplementedError
 
     def get_duplicate_action(
@@ -192,14 +195,15 @@ class ImportSession:
         log.debug("default action for duplicates: {}", choice)
         return DuplicateAction(choice)  # type: ignore[call-arg]
 
-    def choose_item(self, task: ImportTask):
+    def choose_item(self, task: SingletonImportTask) -> TrackMatch | Action:
         raise NotImplementedError
 
-    def run(self):
+    def run(self) -> None:
         """Run the import task."""
         self.logger.info("import started {}", time.asctime())
         self.set_config(config["import"])
 
+        stages: list[Iterator[stagefuncs.StageMessage]]
         # Set up the pipeline.
         if self.query is None:
             stages = [stagefuncs.read_tasks(self)]
@@ -250,7 +254,9 @@ class ImportSession:
 
     # Incremental and resumed imports
 
-    def already_imported(self, toppath: PathBytes, paths: Sequence[PathBytes]):
+    def already_imported(
+        self, toppath: PathBytes, paths: Sequence[PathBytes]
+    ) -> bool:
         """Returns true if the files belonging to this task have already
         been imported in a previous session.
         """
@@ -272,7 +278,7 @@ class ImportSession:
             self._history_dirs = ImportState().taghistory
         return self._history_dirs
 
-    def already_merged(self, paths: Sequence[PathBytes]):
+    def already_merged(self, paths: Sequence[PathBytes]) -> bool:
         """Returns true if all the paths being imported were part of a merge
         during previous tasks.
         """
@@ -281,7 +287,7 @@ class ImportSession:
                 return False
         return True
 
-    def mark_merged(self, paths: Sequence[PathBytes]):
+    def mark_merged(self, paths: Sequence[PathBytes]) -> None:
         """Mark paths and directories as merged for future reimport tasks."""
         self._merged_items.update(paths)
         dirs = {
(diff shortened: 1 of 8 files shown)


beets/importer/session.py
```diff
@@ -87,22 +90,23 @@ def __init__(
...
-    def set_config(self, config):
+    def set_config(self, config: confuse.ConfigView) -> None:
         """Set `config` property from global import config and make
         implied changes.
         """
         # FIXME: Maybe this function should not exist and should instead
         # provide "decision wrappers" like "should_resume()", etc.
-        iconfig = dict(config)
-        self.config = iconfig
+        self.config = iconfig = config
```
semohr, line 110: The logic changed here slightly, before we did a copy of the config `dict(config)`. Now we just save a reference, right? Could this influence anything unexpectedly?

beets/importer/tasks.py
```diff
@@ -1065,16 +1086,14 @@ def singleton(self, path: util.PathBytes):
...
+    def album(
+        self, paths: Iterable[util.PathBytes], dirs: list[util.PathBytes]
+    ) -> ImportTask | None:
         """Return a `ImportTask` with all media files from paths.
 
         `dirs` is a list of parent directories used to record already
         imported albums.
         """
-
-        if dirs is None:
```
semohr, line 1062: Why is this not needed anymore?

beets/util/pipeline.py
```diff
@@ -170,7 +189,9 @@ def multiple(messages):
 R = TypeVar("R")
 
 
-def stage(func: Callable[[Unpack[A], T], R | None]):
+def stage(
+    func: Callable[[Unpack[A], T], R | None],
```
semohr: Maybe we can use ParamSpec instead? Would it be a big issue to add kwargs here?

beets/util/pipeline.py
```diff
@@ -388,57 +438,75 @@ def run(self):
...
+    def run_parallel(self, queue_size: int = DEFAULT_QUEUE_SIZE) -> None:
         """Run the pipeline in parallel using one thread per stage. The
         messages between the stages are stored in queues of the given
         size.
         """
         base_ctx = contextvars.copy_context()
-        queue_count = len(self.stages) - 1
-        queues = [CountedQueue(queue_size) for i in range(queue_count)]
-        threads = []
+        queue_count = len(self.stages)
```
semohr: I think the changes here are a bit out of scope.

beetsplug/fetchart.py
```diff
@@ -1529,8 +1529,12 @@ def assign_art(self, session: ImportSession, task: ImportTask):
 
             self._set_art(task.album, candidate, not removal_enabled)
 
-            if removal_enabled and not self._is_candidate_fallback(candidate):
-                task.prune(candidate.path)
+            if (
+                removal_enabled
+                and not self._is_candidate_fallback(candidate)
+                and (path := candidate.path)
```
semohr, line 1535: A candidate path can be none? How so?

test/test_plugins.py
```diff
@@ -143,18 +138,9 @@ def __init__(self):
...
-                    or isinstance(task, ArchiveImportTask)
-                ):
-                    return task
-
-                new_tasks = []
-                for item in task.items:
-                    new_tasks.append(SingletonImportTask(task.toppath, item))
-
-                return new_tasks
+                return [
```
semohr, line 134: Wont this break plugins that use that event? The event system is a public interface imo and this seems like a breaking change to me.

semohr (changes requested):
Left a few comments. Mainly questions for my understanding but also some things that seem a bit out of scope to me. Quite alot of things are happening here :)

We might want to add these files to the stricter opt in mypy config too now (ofc if  possible see `setup.cfg`).

beets/importer/session.py
snejus: See a2bb02e2 commit message where this originated:

> Preserve importer config views
> 
> Previously, we had
> `iconfig: dict[str, confuse.Subview] = dict(config)`
> 
> Since this was just a dictionary, explicit assignments like
> `iconfig["copy"] = False` replaced `Subview` with the value:
> 
> ``` iconfig
> {
>  'autotag': <Subview: import.autotag>,
>  'bell': <Subview: import.bell>,
>  'copy': False,
>  ...
> }
> ```
> 
> Thus, keys overriden by `set_config` lost their `Subview` type, where methods like `.get` were lost. This ultimately
> resulted in the following type:
> 
> `ImportSession.config: dict[str, confuse.Subview | str | bool]`

beets/importer/tasks.py
snejus: See 8458baa - this branch is never executed - dirs are _always_ provided in the code.

beetsplug/fetchart.py
snejus:
```py
class Candidate:
    """Holds information about a matching artwork, deals with validation of
    dimension restrictions and resizing.
    """

    def __init__(
        self,
        ...
        path: None | bytes = None,
    ):
        self.path = path
```

semohr: I was more thinking from the high level point of view. I can have a look at the code...

Having a closer look, I guess either url or path is set for a given fetchart candidate. I haven't really looked at that plugin too much yet so I was just wondering.

beets/importer/session.py
semohr: Doesn't make it clearer for me, sorry. I understand that this makes the type hint more concise, but I'm wondering about the following:

Previously, we created a copy of the `config` object. That meant updates to the copied `iconfig` object, for example:

```python
if iconfig["move"]:
    iconfig["copy"] = False
    iconfig["link"] = False
    iconfig["hardlink"] = False
    iconfig["reflink"] = False
```

did **not** propagate back to the original `config` object. As a result, if another session is created within the same runtime, it starts from the original configuration rather than inheriting these modifications.

With this change, are these updates now applied directly to the main `config` object? If so, wouldn't that mean that subsequent sessions created in the same runtime would observe the modified configuration?

beetsplug/fetchart.py
snejus: The only reason why this change is needed is that I added a typehint to `ImportTask.prune`.

test/test_plugins.py
snejus: `import_task_created` handler is expected to return a **list** of tasks.

This issue was pointed out by grug: 
image

Had a single task been returned here it would make `ImportTask.handle_created` explode. For example, I tried it on master:

```diff
diff --git a/test/test_plugins.py b/test/test_plugins.py
index 9437ba51a..a11157f71 100644
--- a/test/test_plugins.py
+++ b/test/test_plugins.py
@@ -144,18 +144,7 @@ class ToSingletonPlugin
                 )
 
             def import_task_created_event(self, session, task):
-                if (
-                    isinstance(task, SingletonImportTask)
-                    or isinstance(task, SentinelImportTask)
-                    or isinstance(task, ArchiveImportTask)
-                ):
-                    return task
-
-                new_tasks = []
-                for item in task.items:
-                    new_tasks.append(SingletonImportTask(task.toppath, item))
-
-                return new_tasks
+                return SingletonImportTask(task.toppath, task.items[0])
 
         to_singleton_plugin = ToSingletonPlugin
         self.register_plugin(to_singleton_plugin)
```

```
FAILED test/test_plugins.py::TestEvents::test_import_task_created_with_pl
ugin - TypeError: 'SingletonImportTask' object is not iterable
```

beets/util/pipeline.py
snejus: Maybe I'll create a separate PR for `pipeline` types. The only reason why I added them was that I tried to maximise type coverage.

* Without `pipeline` types:
image

* With `pipeline` types:
image

snejus:
I removed new `beets.util.pipeline` types from this PR and will open a separate one for those!

beets/importer/session.py
snejus: > this makes the type hint more concise

Not only this: the main issue is that since the copied config has `dict[str, confuse.Subview]` structure, initially we have

```py
>>> iconfig["copy"]
<Subview: import.copy>
```

but after setting the value we have
```py
>>> iconfig["copy"] = False
>>> iconfig["copy"]
False
```

And thus:

```py
>>> iconfig["copy"].get(bool)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[8], line 1
----> 1 iconfig["copy"].get(bool)

AttributeError: 'bool' object has no attribute 'get'
```

While this works for the rest of the keys that aren't overriden:

```py
>>> iconfig["incremental"].get(bool)
True
```

---

> With this change, are these updates now applied directly to the main config object? If so, wouldn't that mean that subsequent sessions created in the same runtime would observe the modified configuration?

Indeed - however we only create a single session during import. And even if we did create another one, it would be based on the same config, so it would end up with the same config.

semohr: That is a light "how the hell did this even get added in the first place" from me :rofl: 
Thanks for also having a look here.

test/test_plugins.py
semohr: Interesting!

beets/importer/session.py
semohr: This discussion is also related to https://github.com/beetbox/beets/issues/6629 imo.

semohr (approved):

PR #6712, opened 2026-06-06
chore: improve beets maintenance path

lphuc2250gma:
Summary:
- Add or tighten focused edge-case tests or type assertions in test/test_util.py, test/test_files.py related to Python typing, tests, CLI ergonomics, observability; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/test/test_files.py
+++ b/test/test_files.py
@@ -245,6 +245,12 @@ class HelperTest(unittest.TestCase):
         a = ["a", "b", "c"]
         assert util.components(p) == a
 
+    def test_ancestry_works_on_empty_string(self):
+        assert util.ancestry("") == []
+
+    def test_components_works_on_empty_string(self):
+        assert util.components("") == []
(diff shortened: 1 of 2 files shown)


test/test_files.py
```diff
@@ -245,6 +245,12 @@ def test_components_works_on_relative(self):
         a = ["a", "b", "c"]
         assert util.components(p) == a
 
+    def test_ancestry_works_on_empty_string(self):
```
snejus, line 248: Would you mind refactoring these tests to use `pytest.mark.parametrize`?

uv.lock
```diff
@@ -0,0 +1,3 @@
+version = 1
```
snejus, line 1: What is this one for?

PR #6722, opened 2026-06-08
Add %tunique{} path template function for track disambiguation within albums

arsaboo:
Adds a new `%tunique{}` path template function (analogous to `%aunique{}` for albums and `%sunique{}` for singletons) that disambiguates tracks within the same album that would otherwise produce identical filenames (e.g., two tracks with the same title on the same album or across different discs). This follows the structure and logic used in `aunique` or `sunique`.

Fixes #5950 

**Default configuration:**

```yaml
tunique:
    keys: title
    disambiguators: track disc artist
    bracket: '[]'
```
The function first finds all sibling items (same `album_id`) with matching key values (default: same `title`). If duplicates exist, it iterates through the disambiguator fields looking for one that uniquely separates all duplicates. If no field suffices, the stable database ID is used as a fallback. For items without a duplicate, no suffix is added.

This solves the problem where `beet move` would non-deterministically append .1/.2 suffixes via `unique_path()` for duplicate-titled tracks, which would change every time the command was re-run.

Example usage in path config:
```yaml
paths:
    default: $albumartist/$album/$album ($year) - $title%tunique{}
```
For an album with two identically-titled tracks:
```yaml
- Artist/Album/Album (1993) - Common Title [07].mp3
- Artist/Album/Album (1993) - Common Title [11].mp3
```
## To Do

### DIFF
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -97,6 +97,11 @@ sunique:
     disambiguators: year trackdisambig
     bracket: '[]'
 
+tunique:
+    keys: title
+    disambiguators: track disc artist
+    bracket: '[]'
+
 # --------------- Tagging ---------------
 
 per_disc_numbering: no
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -99,7 +99,7 @@ sunique:
 
 tunique:
     keys: title
-    disambiguators: track disc artist
+    disambiguators: disc track artist
     bracket: '[]'
 
 # --------------- Tagging ---------------
--- a/beets/library/models.py
+++ b/beets/library/models.py
@@ -1467,6 +1467,56 @@ class DefaultTemplateFunctions:
             lambda i: i.album_id is not None,
         )
 
+    def tmpl_tunique(
+        self,
+        keys: str | None = None,
+        disam: str | None = None,
+        bracket: str | None = None,
+    ) -> str:
+        """Generate a string that is guaranteed to be unique among all
+        items in the same album that share the same set of keys.
+
+        A field from "disam" is used in the string if one is sufficient to
+        disambiguate the tracks. Otherwise, a fallback opaque value is
+        used. Both "keys" and "disam" should be given as
+        whitespace-separated lists of field names, while "bracket" is a
+        pair of characters to be used as brackets surrounding the
+        disambiguator or empty to have no brackets.
+        """
+        if not self.item or not self.lib:
+            return ""
+
+        if isinstance(self.item, Item):
+            item_id = self.item.id
+        else:
+            raise NotImplementedError("tunique is only implemented for items")
+
+        if item_id is None:
+            return ""
+
+        # Only applies to items in an album.
+        album_id = self.item.get("album_id")
+        if album_id is None:
+            return ""
+
+        resolved_keys = keys or beets.config["tunique"]["keys"].as_str()
+        key_fields = resolved_keys.split() or ["title"]
+        query = dbcore.AndQuery(
+            [dbcore.MatchQuery(f, self.item.get(f)) for f in key_fields]
+            + [dbcore.MatchQuery("album_id", album_id)]
+        )
+
+        return self._tmpl_unique(
+            "tunique",
+            keys,
+            disam,
+            bracket,
+            item_id,
+            self.item,
+            lambda i: i.album_id is None,
+            query=query,
+        )
+
     def _tmpl_unique_memokey(
         self,
         name: str | None,
@@ -1488,6 +1538,7 @@ class DefaultTemplateFunctions:
         item_id: int,
         db_item: LibModel,
         skip_item: Callable[[LibModel], bool],
+        query: dbcore.Query | None = None,
     ) -> str:
         """Generate a string that is guaranteed to be unique among all items of
         the same type as "db_item" who share the same set of keys.
@@ -1506,8 +1557,8 @@ class DefaultTemplateFunctions:
         "skip_item" is a function that must return True when the template
         should return an empty string.
 
-        "initial_subqueries" is a list of subqueries that should be included
-        in the query to find the ambiguous items.
+        "query" is a pre-built query to use instead of building one from
+        ``db_item.duplicates_query(keys)``.
         """
         lib = self.lib
         if lib is None:
@@ -1538,26 +1589,27 @@ class DefaultTemplateFunctions:
             bracket_r = ""
 
         # Find matching items to disambiguate with.
-        query = db_item.duplicates_query(keys_list)
-        ambigous_items = (
+        if query is None:
+            query = db_item.duplicates_query(keys_list)
+        ambiguous_items = (
             lib.items(query) if isinstance(db_item, Item) else lib.albums(query)
         )
 
         # If there's only one item to matching these details, then do
         # nothing.
-        if len(ambigous_items) == 1:
+        if len(ambiguous_items) == 1:
             lib._memotable[memokey] = ""
             return ""
 
         # Find the first disambiguator that distinguishes the items.
         for disambiguator in disam_list:
             # Get the value for each item for the current field.
-            disam_values = {s.get(disambiguator, "") for s in ambigous_items}
+            disam_values = {s.get(disambiguator, "") for s in ambiguous_items}
 
             # If the set of unique values is equal to the number of
             # items in the disambiguation set, we're done -- this is
             # sufficient disambiguation.
-            if len(disam_values) == len(ambigous_items):
+            if len(disam_values) == len(ambiguous_items):
                 break
         else:
             # No disambiguator distinguished all fields.
--- a/beets/library/models.py
+++ b/beets/library/models.py
@@ -1503,7 +1503,10 @@ class DefaultTemplateFunctions:
         resolved_keys = keys or beets.config["tunique"]["keys"].as_str()
         key_fields = resolved_keys.split() or ["title"]
         query = dbcore.AndQuery(
-            [dbcore.MatchQuery(f, self.item.get(f)) for f in key_fields]
+            [
+                self.item.field_query(f, self.item.get(f), dbcore.MatchQuery)
+                for f in key_fields
+            ]
             + [dbcore.MatchQuery("album_id", album_id)]
         )
 
@@ -1603,9 +1606,15 @@ class DefaultTemplateFunctions:
             return ""
 
         # Find the first disambiguator that distinguishes the items.
+        # Use path-formatted values so separators like "/" in "AC/DC"
+        # are treated as they appear on disk (e.g., "AC_DC" after
+        # path_sep_replace).
         for disambiguator in disam_list:
             # Get the value for each item for the current field.
-            disam_values = {s.get(disambiguator, "") for s in ambiguous_items}
+            disam_values = {
+                s.formatted(for_path=True).get(disambiguator)
+                for s in ambiguous_items
+            }
 
             # If the set of unique values is equal to the number of
             # items in the disambiguation set, we're done -- this is
--- a/beets/library/models.py
+++ b/beets/library/models.py
@@ -1163,6 +1163,7 @@ class Item(LibModel):
             )
 
         self.db._memotable = {}
+        self.db._group_memotable = {}
 
     def move(
         self,
@@ -1427,6 +1428,8 @@ class DefaultTemplateFunctions:
         if album_id is None:
             return ""
 
+        if bracket is None:
+            bracket = beets.config["aunique"]["bracket"].as_str()
         memokey = self._tmpl_unique_memokey(
             "aunique", keys, disam, bracket, album_id
         )
@@ -1534,6 +1537,7 @@ class DefaultTemplateFunctions:
             self.item,
             lambda i: i.album_id is None,
             query=query,
+            scope_id=album_id,
         )
 
     def _tmpl_unique_memokey(
@@ -1547,8 +1551,6 @@ class DefaultTemplateFunctions:
         """Get the memokey for the unique template named "name" for the
         specific parameters.
         """
-        if bracket is None and name is not None:
-            bracket = beets.config[name]["bracket"].as_str()
         return (name, keys, disam, bracket, item_id)
 
     def _tmpl_unique(
@@ -1561,6 +1563,7 @@ class DefaultTemplateFunctions:
         db_item: LibModel,
         skip_item: Callable[[LibModel], bool],
         query: dbcore.Query | None = None,
+        scope_id: int | None = None,
     ) -> str:
         """Generate a string that is guaranteed to be unique among all items of
         the same type as "db_item" who share the same set of keys.
@@ -1586,19 +1589,18 @@ class DefaultTemplateFunctions:
         if lib is None:
             return ""
 
+        if skip_item(db_item):
+            return ""
+
+        if bracket is None:
+            bracket = beets.config[name]["bracket"].as_str()
         memokey = self._tmpl_unique_memokey(name, keys, disam, bracket, item_id)
         memoval = lib._memotable.get(memokey)
         if memoval is not None:
             return memoval
 
-        if skip_item(db_item):
-            lib._memotable[memokey] = ""
-            return ""
-
         keys = keys or beets.config[name]["keys"].as_str()
         disam = disam or beets.config[name]["disambiguators"].as_str()
-        if bracket is None:
-            bracket = beets.config[name]["bracket"].as_str()
         keys_list = keys.split()
         disam_list = disam.split()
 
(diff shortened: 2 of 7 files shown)


snejus:
I think it would be best to wait until https://github.com/beetbox/beets/pull/6825 is merged as I added types to these methods here.

arsaboo:
can we get another 👀 now that #6585 is merged.

arsaboo:
@snejus, can we take another look at this when you get a chance? Thanks!

beets/config_default.yaml
```diff
@@ -97,6 +97,11 @@ sunique:
     disambiguators: year trackdisambig
     bracket: '[]'
 
+tunique:
+    keys: title
+    disambiguators: track disc artist
```
snejus, line 102: What about `title`?

beets/library/models.py
```diff
@@ -1463,6 +1463,56 @@ def tmpl_sunique(
...
+
+        # Only applies to items in an album.
+        album_id = self.item.get("album_id")
+        if album_id is None:
+            return ""
+
+        resolved_keys = keys or beets.config["tunique"]["keys"].as_str()
+        key_fields = resolved_keys.split() or ["title"]
+        query = dbcore.AndQuery(
+            [dbcore.MatchQuery(f, self.item.get(f)) for f in key_fields]
```
snejus: Need to use `item.field_query` here otherwise this breaks for flexible/computed fields.

beets/library/models.py
```diff
@@ -1463,6 +1463,56 @@ def tmpl_sunique(
...
+            disam,
+            bracket,
+            item_id,
+            self.item,
+            lambda i: i.album_id is None,
+            query=query,
+        )
+
     def _tmpl_unique_memokey(
         self,
```
snejus, line 1525: Note that uniqueness in this function is checked **before** path formatting, so this is problematic for artists such as **AC/DC**.

beets/library/models.py
```diff
@@ -1463,6 +1463,56 @@ def tmpl_sunique(
             lambda i: i.album_id is not None,
         )
 
+    def tmpl_tunique(
```
snejus, line 1471: Would it make sense to keep the scope-specific query in each template entrypoint, but reuse `_tmpl_unique` for all common uniqueness behaviour?

`%aunique{}`, `%sunique{}`, and `%tunique{}` query different scopes, so I don't think we need to force their query construction into one abstraction. However, disambiguator selection, path-formatted value comparison, brackets, fallback, and memoization should remain shared in `_tmpl_unique`.

For `%tunique{}`, the query could use `Item.field_query` so flexible fields continue to work. The path-normalization collision should then be fixed in `_tmpl_unique`, which would address it consistently for all three templates.

snejus (changes requested):

beets/config_default.yaml
arsaboo: `title` not being unique is the problem we are trying to resolve, so using it as disambiguator is redundant. It was intentionally omitted, consistent with `aunique`/`sunique`.

beets/library/models.py
arsaboo: Done

beets/library/models.py
arsaboo: Fixed...also added a test

beets/library/models.py
arsaboo: Done! Moved the scope-specific query construction into each tmpl_*unique entry point and kept disambiguator selection, path-formatted comparison, brackets, fallback, and memoization shared in `_tmpl_unique`. Also fixed the path-normalization collision there so it applies consistently to `aunique`/`sunique`/`tunique`

snejus: did you push this commit? 😅

arsaboo: Yes, it's pushed — commit [c8c11b2ed](https://github.com/beetbox/beets/pull/6722/commits/c8c11b2ed2eb384d14094950b1da009474bb00f0)

snejus: ah, yeah, makes sense

beets/library/models.py
```diff
@@ -1485,6 +1485,59 @@ def tmpl_sunique(
...
+            return ""
+
+        resolved_keys = keys or beets.config["tunique"]["keys"].as_str()
+        key_fields = resolved_keys.split() or ["title"]
+        query = dbcore.AndQuery(
+            [
+                self.item.field_query(f, self.item.get(f), dbcore.MatchQuery)
+                for f in key_fields
+            ]
+            + [dbcore.MatchQuery("album_id", album_id)]
```
snejus: This should also be fed into `field_query` I think?

beets/library/models.py
```diff
@@ -1485,6 +1485,59 @@ def tmpl_sunique(
...
+        # Only applies to items in an album.
+        album_id = self.item.get("album_id")
+        if album_id is None:
+            return ""
+
+        resolved_keys = keys or beets.config["tunique"]["keys"].as_str()
+        key_fields = resolved_keys.split() or ["title"]
+        query = dbcore.AndQuery(
+            [
+                self.item.field_query(f, self.item.get(f), dbcore.MatchQuery)
```
snejus: I think there is still a related collision when the key values themselves only become equal during path formatting. The query continues to compare their raw database values, so `%tunique{}` does not consider them duplicates.

For example:

```yaml
paths:
    default: $title%tunique{title,track}

path_sep_replace: _
```

```text
title: AC/DC, track: 1
title: AC_DC, track: 2
```

The query treats `AC/DC` and `AC_DC` as different titles and returns no suffix for either track. Path formatting then produces the same destination for both:

```text
AC_DC.mp3
AC_DC.mp3
```

Would it be possible to select the album siblings first and compare their path-formatted key tuples? In this example, I would expect:

```text
AC_DC [01].mp3
AC_DC [02].mp3
```

beets/library/models.py
```diff
@@ -1556,26 +1610,33 @@ def _tmpl_unique(
...
 
         # Find the first disambiguator that distinguishes the items.
+        # Use path-formatted values so separators like "/" in "AC/DC"
+        # are treated as they appear on disk (e.g., "AC_DC" after
+        # path_sep_replace).
         for disambiguator in disam_list:
             # Get the value for each item for the current field.
-            disam_values = {s.get(disambiguator, "") for s in ambigous_items}
+            disam_values = {
+                s.formatted(for_path=True).get(disambiguator)
```
snejus: `formatted(for_path=True)` only replaces path separators. `destination()` subsequently applies `asciify_paths`, the configured `replace` rules, and path truncation, so values can still appear unique here but become identical in the final destination.

For example:

```yaml
paths:
    default: $title%tunique{title,artist}

replace:
    '[<>:\?\*\|]': _
```

```text
title: Common Title, artist: AC:DC
title: Common Title, artist: AC_DC
```

This code sees `AC:DC` and `AC_DC` as distinct and selects `artist` as the disambiguator. The configured replacement is applied afterwards, so both destinations become:

```text
Common Title [AC_DC].mp3
Common Title [AC_DC].mp3
```

Could the comparison use the same transformations as the final destination before accepting a disambiguator? If the transformed values collide, both tracks should fall back to their stable IDs instead.

The same problem can occur with `asciify_paths`, for example when `Beyoncé` and `Beyonce` are selected as otherwise distinct disambiguator values.

beets/library/models.py
```diff
@@ -1485,6 +1485,59 @@ def tmpl_sunique(
...
+                for f in key_fields
+            ]
+            + [dbcore.MatchQuery("album_id", album_id)]
+        )
+
+        return self._tmpl_unique(
+            "tunique",
+            keys,
+            disam,
+            bracket,
```
snejus, line 1532: One other issue I noticed while checking the new bracket support: could you include the resolved `bracket` value in the memoization key?

The current key contains the template name, keys, disambiguators, and item ID, but not the brackets. Two calls for the same item using different brackets therefore reuse the result of the first call.

For example:

```yaml
paths:
    default: $title%tunique{title,track,()}%tunique{title,track,[]}
```

For a duplicate-title track with track number 7, I would expect:

```text
Common Title (07) [07].mp3
```

However, after the first call caches ` (07)`, the second call returns the same cached value:

```text
Common Title (07) (07).mp3
```

Including the resolved brackets in `_tmpl_unique_memokey` should keep these calls independent. Since the helper is shared, it would also make the bracket argument behave consistently for `%aunique{}` and `%sunique{}`.

snejus (changes requested):

snejus:
See below if that helps:


```diff
diff --git a/test/test_library.py b/test/test_library.py
index b52d9d536..82e40efe1 100644
--- a/test/test_library.py
+++ b/test/test_library.py
@@ -1022,6 +1022,34 @@ def test_tunique_path_formatted_disambiguator_collision
         self._assert_dest(b"/base/Common Title [%d]" % i1.id, i1)
         self._assert_dest(b"/base/Common Title [%d]" % i2.id, i2)
 
+    def test_tunique_path_formatted_key_collision(self, items):
+        i1, i2 = items
+        i1.title = "AC/DC"
+        i2.title = "AC_DC"
+        i1.store()
+        i2.store()
+
+        self._setf("$title%tunique{title,track}")
+        self._assert_dest(b"/base/AC_DC [07]", i1)
+        self._assert_dest(b"/base/AC_DC [11]", i2)
+
+    def test_tunique_legalized_disambiguator_collision(self, items):
+        i1, i2 = items
+        i2.track = i1.track
+        i1.artist = "AC:DC"
+        i2.artist = "AC_DC"
+        i1.store()
+        i2.store()
+
+        self._setf("$title%tunique{title,artist}")
+        self._assert_dest(b"/base/Common Title [%d]" % i1.id, i1)
+        self._assert_dest(b"/base/Common Title [%d]" % i2.id, i2)
+
+    def test_tunique_memoizes_bracket_variants_separately(self, items):
+        i1, _i2 = items
+        self._setf("$title%tunique{title,track,()}%tunique{title,track,[]}")
+        self._assert_dest(b"/base/Common Title (07) [07]", i1)
+
 
 class TestPluginDestination(TestHelper):
     @pytest.fixture(autouse=True)
```

arsaboo:
@snejus Thanks! I’ve updated the album query, added normalized key comparison, accounted for replacements and asciification when selecting disambiguators, and included resolved brackets in the cache key. These changes are in [51ce376a6](https://github.com/beetbox/beets/pull/6722/commits/51ce376a66cc606bb5764513736382f87bbbc40e).

One remaining limitation is that truncation and context-sensitive replacement rules depend on the complete rendered path, so normalizing individual fields cannot guarantee final-path uniqueness. For example, if a title already fills the available filename length, `Long title… [01].mp3` and `Long title… [02].mp3` could both be truncated to the same filename because the distinguishing suffix is removed. Stable ID suffixes have the same limitation. Fully addressing this would require reserving space for the suffix when truncating the complete filename.

Would you prefer to handle these full-path cases separately, or include them in this PR?

test/test_library.py
```diff
@@ -875,6 +883,193 @@ def test_key_flexible_attribute(self, items):
...
+        # i2 has no trackdisambig, so no suffix
+        self._assert_dest(b"/base/Common Title", i2)
+
+    def test_tunique_with_flex_key_field(self, items):
+        # Flexible field as key must be queried via field_query.
+        i1, i2 = items
+        i1.title = "Diff1"
+        i2.title = "Diff2"
+        i1["flexkey"] = "same"
+        i2["flexkey"] = "same"
```
snejus: I deleted both `flexkey` assignments so the field is never set, and this still passes - both sides normalize to `""`. Could you give them different values instead? The comment is stale too, keys aren't queried at all now.

test/test_library.py
```diff
@@ -875,6 +883,193 @@ def test_key_flexible_attribute(self, items):
...
+        self._assert_dest(b"/base/Common Title [07]", i1)
+        self._assert_dest(b"/base/Common Title [11]", i2)
+
+    def test_tunique_with_default_arguments_uses_track(self, items):
+        i1, i2 = items
+        self._setf("$title%tunique{}")
+        self._assert_dest(b"/base/Common Title [07]", i1)
+        self._assert_dest(b"/base/Common Title [11]", i2)
+
+    def test_tunique_expands_to_nothing_for_unique_titles(self, items):
```
snejus: Remove `_tunique` from test names as they are redundant

test/test_library.py
```diff
@@ -875,6 +883,193 @@ def test_key_flexible_attribute(self, items):
...
+        i2.title = "AC_DC"
+        i1.store()
+        i2.store()
+
+        self._assert_dest(b"/base/AC_DC [07]", i1)
+        self._assert_dest(b"/base/AC_DC [11]", i2)
+
+    def test_tunique_replaced_disambiguator_collision(self, items):
+        i1, i2 = items
+        self.lib.replacements = [(re.compile(":"), "_")]
```
snejus: Use `~` here which is not in the default replacements.

```suggestion
        self.lib.replacements = [(re.compile("~"), "_")]
```

test/test_library.py
```diff
@@ -875,6 +883,193 @@ def test_key_flexible_attribute(self, items):
...
+    @pytest.fixture(autouse=True)
+    def items(self, setup):
+        self.lib.directory = b"/base"
+        self.lib.path_formats = [("default", "path")]
+
+        i1 = item()
+        i1.title = "Common Title"
+        i1.track = 7
+        i1.disc = 1
+        i1.artist = "Some Artist"
```
snejus: Applies to the rest of the tests too

```suggestion
        i1 = item(title="Common Title", track=7, disc=1, artist="Some Artist")
```

test/test_library.py
```diff
@@ -875,6 +883,193 @@ def test_key_flexible_attribute(self, items):
...
+        i1, i2 = items
+        i1.title = "AC/DC"
+        i2.title = "AC_DC"
+        i1.store()
+        i2.store()
+
+        self._assert_dest(b"/base/AC_DC [07]", i1)
+        self._assert_dest(b"/base/AC_DC [11]", i2)
+
+    def test_tunique_replaced_disambiguator_collision(self, items):
```
snejus: I think `test_tunique_replaced_disambiguator_collision`, `test_tunique_asciified_disambiguator_collision` and `test_tunique_path_formatted_disambiguator_collision` can be parametrized.

test/test_library.py
```diff
@@ -875,6 +883,193 @@ def test_key_flexible_attribute(self, items):
...
+
+    def test_tunique_expands_to_disambiguating_disc(self, items):
+        i1, i2 = items
+        i2.track = i1.track
+        i2.disc = 2
+        i2.store()
+
+        self._setf("$title%tunique{title,disc}")
+        self._assert_dest(b"/base/Common Title [01]", i1)
+        self._assert_dest(b"/base/Common Title [02]", i2)
```
snejus: These can be parametrized

beets/config_default.yaml
snejus: `track` first means `disc`/`artist` never fire, so `%tunique{}` is effectively `$track`. The docs example ("same title on different discs") gives `[01]`/`[08]`. Reorder, or change the example?

beets/library/models.py
```diff
@@ -1485,17 +1487,69 @@ def tmpl_sunique(
...
         disam: str | None,
+        bracket: str | None,
         item_id: int | None,
-    ) -> tuple[str | None, str | None, str | None, int | None]:
+    ) -> tuple[str | None, str | None, str | None, str | None, int | None]:
         """Get the memokey for the unique template named "name" for the
         specific parameters.
         """
-        return (name, keys, disam, item_id)
+        if bracket is None and name is not None:
```
snejus: Resolving the bracket inside the memo key moved the config read ahead of `skip_item`. With `sunique.bracket: 5`, `%sunique{}` on an item in an album:

```
b4d99a03f  X.mp3
51ce376a   X_sunique.bracket_ must be a string_.mp3
```

`functemplate` swallows the `ConfigTypeError` into the path. Could we resolve it in the caller instead, after `skip_item`?

beets/library/models.py
```diff
@@ -1485,17 +1487,69 @@ def tmpl_sunique(
...
+        used. Both "keys" and "disam" should be given as
+        whitespace-separated lists of field names, while "bracket" is a
+        pair of characters to be used as brackets surrounding the
+        disambiguator or empty to have no brackets.
+        """
+        if not self.item or not self.lib:
+            return ""
+
+        if isinstance(self.item, Item):
+            item_id = self.item.id
```
snejus, line 1518: `%tunique{}` is expensive, and it doesn't need a collision to be expensive. Ordinary albums, all titles distinct, so nothing is ever disambiguated:

```
2400 tracks (200 albums x 12)  plain= 2.21s  aunique= 2.39s  tunique= 28.08s
6000 tracks (500 albums x 12)  plain= 5.68s  aunique= 6.16s  tunique= 70.91s
```

That's ~12.5x flat, so a 50k library goes from about 45s to roughly 10 minutes for `beet move`. It's also quadratic within an album, so a single 100 track album is ~10s on its own.

It's the memo key rather than the comparison itself. It's per item, where `%aunique{}`'s is per album, so we redo the whole sibling analysis once per track for a conclusion that's identical across the group. Profiling n=100, the time is almost entirely `FormattedItemMapping._get` -> `Item._cached_album`, which calls `album.load()` on every access - 15,450 of those for one album.

The suffix is per item, but the group size and the chosen disambiguator aren't. Caching those per group gets it to:

```
n= 50  plain= 36.1ms  tunique= 109.0ms
n=100  plain=107.6ms  tunique= 227.3ms
```

11.2s -> 0.23s at n=100, and linear rather than quadratic. Something like this

```diff
--- a/beets/library/library.py
+++ b/beets/library/library.py
@@ -53,6 +53,11 @@ class Library
     _memotable: dict[
         tuple[str | None, str | None, str | None, str | None, int | None], str
     ]
+    # Caches the per-collision-group work shared by every member of a group.
+    _group_memotable: dict[
+        tuple[str | None, str | None, str | None, int | None, tuple[str, ...]],
+        tuple[int, str | None],
+    ]
     replacements: Replacements
@@ -87,6 +92,7 @@ def get_replacements
         self.replacements = self.get_replacements()
         self._memotable = {}
+        self._group_memotable = {}
@@ -104,6 +110,7 @@ def add
         obj.add(self)
         self._memotable = {}
+        self._group_memotable = {}
         return obj.id

--- a/beets/library/models.py
+++ b/beets/library/models.py
@@ -1534,6 +1534,7 @@ def tmpl_tunique
             lambda i: i.album_id is None,
             query=query,
+            scope_id=album_id,
         )
@@ -1561,6 +1562,7 @@ def _tmpl_unique
         query: dbcore.Query | None = None,
+        scope_id: int | None = None,
     ) -> str:
@@ -1610,47 +1612,60 @@ def _tmpl_unique
-        # Find matching items to disambiguate with.
-        if query is None:
-            query = db_item.duplicates_query(keys_list)
-        candidate_items = (
-            lib.items(query) if isinstance(db_item, Item) else lib.albums(query)
-        )
-
         def normalized_field(item: LibModel, field: str) -> str:
             ...

+        # Every member of a collision group reaches the same conclusion about
+        # which disambiguator to use, so compute it once per group.
         key_values = tuple(normalized_field(db_item, key) for key in keys_list)
-        ambiguous_items = [...]
-
-        if len(ambiguous_items) == 1:
+        groupkey = (name, keys, disam, scope_id, key_values)
+        if (group := lib._group_memotable.get(groupkey)) is None:
+            # Find matching items to disambiguate with.
+            scope_query = (
+                db_item.duplicates_query(keys_list) if query is None else query
+            )
+            candidate_items = (
+                lib.items(scope_query)
+                if isinstance(db_item, Item)
+                else lib.albums(scope_query)
+            )
+            ambiguous_items = [
+                item
+                for item in candidate_items
+                if tuple(normalized_field(item, key) for key in keys_list)
+                == key_values
+            ]
+
+            # Find the first disambiguator that distinguishes the items.
+            disambiguator = None
+            for candidate in disam_list:
+                disam_values = {
+                    normalized_field(item, candidate)
+                    for item in ambiguous_items
+                }
+                if len(disam_values) == len(ambiguous_items):
+                    disambiguator = candidate
+                    break
+
+            group = (len(ambiguous_items), disambiguator)
+            lib._group_memotable[groupkey] = group
+
+        ambiguous_count, disambiguator = group
+
+        # If there's only one item matching these details, then do nothing.
+        if ambiguous_count <= 1:
             lib._memotable[memokey] = ""
             return ""

-        for disambiguator in disam_list:
-            ...
-        else:
+        if disambiguator is None:
             # No disambiguator distinguished all fields.
             res = f" {bracket_l}{item_id}{bracket_r}"
```

snejus (changes requested):

snejus:
> Would you prefer to handle these full-path cases separately, or include them in this PR?

I think we can ignore this for now :)

arsaboo:
@snejus anything else pending here?

Issue #6683, opened 2026-05-29
Autotagger Refactor: Unify Album and Singleton Matching Flows


snejus:
## Summary
The autotagger currently has separate matching paths for albums and singletons, with duplicated orchestration logic spread across matching, candidate lookup, and importer session flow.

This refactor series unifies both paths under a shared model:
- `Source` encapsulates search context and current metadata origin.
- `Candidates` encapsulates candidate discovery, deduplication, and recommendation logic.
- `Match` subclasses encapsulate validation and match construction.
- `ImportTask` is split into clear album and singleton task variants with explicit responsibilities.

The goal is to make behavior consistent between album and singleton imports, reduce duplication, and simplify importer architecture with stronger typing and clearer ownership.

## Motivation
- Reduce duplicate logic between album and singleton tagging paths.
- Keep recommendation and candidate lifecycle behavior consistent.
- Make importer/session code operate on one conceptual model (`source`, `candidates`, `match`) regardless of item type.
- Replace mixed importer task responsibilities with a clean album/singleton split.

## Scope
- Introduce shared source context abstraction.
- Move candidate collection and recommendation behavior into dedicated candidate containers.
- Move match validation/construction into `Match` subclasses.
- Update importer/session flows to use the shared abstractions.
- Split importer task responsibilities into explicit album and singleton task types.

## Out of Scope
- New user-facing autotagger features.
- Plugin API redesign beyond compatibility-preserving adjustments required by refactor.

## Deliverables
- [ ] `Source` introduced and integrated in importer candidate lookup.
- [ ] Candidate orchestration moved to `Candidates` classes.
- [ ] Album/singleton importer session paths consume shared candidate interface.
- [ ] Importer task architecture split between album and singleton flows.

## Risks
- Subtle behavior drift in recommendation thresholds/order.
- Plugin hooks depending on old internals.
- Transition churn while importer task hierarchy is reshaped.

## Validation
- Existing autotagger/importer test suite passes.
- Targeted tests verify equivalent behavior for album and singleton candidate resolution.
- No regression in manual search/manual ID flow.
- Importer task tests verify album and singleton flows through distinct task types.

semohr:
I had a look at the initial draft and I'm still not entirely convinced the proposed refactor is the best way forward.

[...]

PR #6723, opened 2026-06-08
Dedupe individual tracks during import based on stable identifier

tommyschnabel:
## Description
This PR adds the `duplicate_track_resolution` (bool) and `duplicate_track_action` (`skip`, `remove`, `keep`/`merge`, `ask`) config options to `import` to enable deduplication of individual tracks when importing.

The recommended config for this is:
```
import:
    duplicate_track_resolution: yes
    duplicate_action: ask          # whole-album duplicates
    duplicate_track_action: skip   # per-track duplicates: fold new tracks into the existing album
    duplicate_keys:
        item: mb_trackid           # match on a stable id (recommended when autotagging)
```

With the above config, any duplicate imports of tracks will be dropped, and if the whole album has been imported before, the album import itself will be dropped. If part of an album is deduped, the new tracks will be folded into the existing album. This config allows for idempotency on any repeat `beet import` executions.

Copied from the new docs, here's what each option does:
>  `skip` drops the duplicate tracks and adds the remaining *new* tracks to the
  existing album they belong to, instead of importing them as a separate album.
  Use this to complete a partially-imported album. If every track is already
  present, the whole album is skipped. (If the matching tracks do not all belong
  to a single album, the new tracks are imported as their own album.)
> - `remove` removes the matching old items from the library before importing.
> - `keep` (and `merge`) import the album unchanged.
> - `ask` prompts you to choose one of the above.

The `duplicate_keys` config option is what's used to dedupe, so set it to something stable like `mb_trackid`.

In the [previous PR](https://github.com/beetbox/beets/pull/6313), @snejus suggested adding this as an addition to the album deduplication that already exists. This implementation is separate from the album dedupe logic because the use-cases seemed incompatible. 

The album dedupe skips/merges/removes based on an album already being imported, but the use-case I wanted was a deeper track-based mechanism. Bolting this onto the album dedupe produced some really weird results, and I wasn't able to achieve parity with my last impl with the available config options. This is a better, more flexible, approach.

## To Do

### DIFF
--- a/beets/importer/session.py
+++ b/beets/importer/session.py
@@ -187,9 +187,9 @@ class ImportSession:
 
     def resolve_track_duplicates(self, task: ImportTask, duplicates) -> str:
         """Decide what to do with album tracks that already exist in the
-        library. Return ``"s"`` (skip the duplicate tracks), ``"k"`` (keep
-        all), ``"r"`` (remove the old items) or ``"f"`` (fold the remaining
-        new tracks into the existing album).
+        library. Return ``"s"`` (skip the duplicate tracks and fold the
+        remaining new tracks into the existing album), ``"k"`` (keep all) or
+        ``"r"`` (remove the old items).
         """
         raise NotImplementedError
 
--- a/beets/importer/session.py
+++ b/beets/importer/session.py
@@ -188,7 +188,8 @@ class ImportSession:
     def resolve_track_duplicates(self, task: ImportTask, duplicates) -> str:
         """Decide what to do with album tracks that already exist in the
         library. Return ``"s"`` (skip the duplicate tracks), ``"k"`` (keep
-        all) or ``"r"`` (remove the old items).
+        all), ``"r"`` (remove the old items) or ``"f"`` (fold the remaining
+        new tracks into the existing album).
         """
         raise NotImplementedError
 
--- a/beets/importer/session.py
+++ b/beets/importer/session.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import os
 import time
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
 from beets import config, logging, plugins, util
 from beets.util import displayable_path, normpath, pipeline, syspath
@@ -46,7 +46,6 @@ class ImportSession:
     _is_resuming: dict[bytes, bool]
     _merged_items: set[PathBytes]
     _merged_dirs: set[PathBytes]
-    _seen_track_keys: set[tuple[Any, ...]]
 
     def __init__(
         self,
@@ -75,7 +74,6 @@ class ImportSession:
         self._is_resuming = {}
         self._merged_items = set()
         self._merged_dirs = set()
-        self._seen_track_keys = set()
 
         # Normalize the paths.
         self.paths = list(map(normpath, paths or []))
@@ -187,24 +185,62 @@ class ImportSession:
     def choose_item(self, task: SingletonImportTask) -> TrackMatch | Action:
         raise NotImplementedError
 
-    def resolve_track_duplicates(
+    def get_track_duplicate_actions(
         self,
         task: ImportTask,
-        duplicates: dict[library.Item, list[library.Item]],
-    ) -> DuplicateAction:
-        """Decide what to do with album tracks that already exist in the
-        library: :attr:`~DuplicateAction.SKIP` (drop the duplicate tracks and
-        fold the remaining new tracks into the existing album),
-        :attr:`~DuplicateAction.KEEP` (import everything) or
-        :attr:`~DuplicateAction.REMOVE` (remove the old items).
+        track_duplicates: dict[library.Item, list[library.Item]],
+    ) -> dict[library.Item, DuplicateAction]:
+        """Get the configured action for each track that duplicates items
+        already in the library.
 
-        ``duplicates`` maps each incoming :class:`~beets.library.Item` to the
-        existing library items it duplicates.
+        ``track_duplicates`` maps each :class:`~beets.library.Item` of the
+        task to the existing library items it duplicates.
         """
-        raise NotImplementedError
+        view = config["import"]["duplicate_tracks_action"]
+        if view.get():
+            # MERGE is not offered per track, so reject it loudly here.
+            choice = view.as_choice(DuplicateAction.track_choices())
+            action = DuplicateAction(choice)  # type: ignore[call-arg]
+        else:
+            choice = config["import"]["duplicate_action"].as_choice(
+                DuplicateAction.choices()
+            )
+            action = DuplicateAction(choice)  # type: ignore[call-arg]
+            if action is DuplicateAction.MERGE:
+                log.debug(
+                    "merge is not available for duplicate tracks; keeping all"
+                )
+                action = DuplicateAction.KEEP
+        log.debug("default action for duplicate tracks: {}", action.value)
+        return dict.fromkeys(track_duplicates, action)
 
-    def choose_item(self, task: ImportTask):
-        raise NotImplementedError
+    def resolve_duplicates(
+        self,
+        task: ImportTask,
+        found_duplicates: list[AnyLibModel],
+        track_duplicates: dict[library.Item, list[library.Item]],
+    ) -> None:
+        """Decide, at a single point, what to do about the album- and
+        track-level duplicates found for ``task``.
+
+        Sets ``task.duplicate_action`` from the album-level duplicates or
+        ``task.track_duplicate_actions`` from the track-level ones. When only
+        *some* tracks duplicate existing items, the import is a partial
+        overlap (e.g. completing a partially-imported album), so per-track
+        resolution applies and the whole-album action is suppressed. When
+        every track is a duplicate, the task is a whole-album duplicate and
+        the album-level action decides.
+        """
+        partial = 0 < len(track_duplicates) < len(task.imported_items())
+
+        if found_duplicates and not partial:
+            task.duplicate_action = self.get_duplicate_action(
+                task, found_duplicates
+            )
+        elif track_duplicates:
+            task.track_duplicate_actions = self.get_track_duplicate_actions(
+                task, track_duplicates
+            )
 
     def run(self):
         """Run the import task."""
@@ -225,10 +261,6 @@ class ImportSession:
             if self.config["group_albums"] and not self.config["singletons"]:
                 # Split directory tasks into one task for each album.
                 stages += [stagefuncs.group_albums(self)]
-
-            # Optionally drop or replace album tracks that already exist in
-            # the library before the autotag lookup runs.
-            stages += [stagefuncs.resolve_track_duplicates(self)]
 
             # These stages either talk to the user to get a decision or,
             # in the case of a non-autotagged import, just choose to
@@ -309,25 +341,6 @@ class ImportSession:
             for path in paths
         }
         self._merged_dirs.update(dirs)
-
-    def track_key_seen(self, key: tuple[Any, ...]) -> bool:
-        """Return whether a per-track duplicate ``key`` was already claimed by
-        an earlier task in this import run (see :ref:`duplicate_track_resolution`).
-
-        This lets the track-level duplicate check catch tracks imported earlier
-        in the *same* run before they reach the database, which matters under
-        the default threaded import where a later task may be checked before an
-        earlier one has been added.
-        """
-        return key in self._seen_track_keys
-
-    def remember_track_key(self, key: tuple[Any, ...]) -> None:
-        """Record a per-track duplicate ``key`` claimed by the current task.
-
-        Only ever called from the single-threaded ``resolve_track_duplicates``
-        pipeline stage, so the shared set needs no locking.
-        """
-        self._seen_track_keys.add(key)
 
     def is_resuming(self, toppath: PathBytes) -> bool:
         """Return `True` if user wants to resume import of this path.
--- a/beets/importer/session.py
+++ b/beets/importer/session.py
@@ -185,11 +185,19 @@ class ImportSession:
     def choose_item(self, task: SingletonImportTask) -> TrackMatch | Action:
         raise NotImplementedError
 
-    def resolve_track_duplicates(self, task: ImportTask, duplicates) -> str:
+    def resolve_track_duplicates(
+        self,
+        task: ImportTask,
+        duplicates: dict[library.Item, list[library.Item]],
+    ) -> DuplicateAction:
         """Decide what to do with album tracks that already exist in the
-        library. Return ``"s"`` (skip the duplicate tracks and fold the
-        remaining new tracks into the existing album), ``"k"`` (keep all) or
-        ``"r"`` (remove the old items).
+        library: :attr:`~DuplicateAction.SKIP` (drop the duplicate tracks and
+        fold the remaining new tracks into the existing album),
+        :attr:`~DuplicateAction.KEEP` (import everything) or
+        :attr:`~DuplicateAction.REMOVE` (remove the old items).
+
+        ``duplicates`` maps each incoming :class:`~beets.library.Item` to the
+        existing library items it duplicates.
         """
         raise NotImplementedError
 
--- a/beets/importer/session.py
+++ b/beets/importer/session.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import os
 import time
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from beets import config, logging, plugins, util
 from beets.util import displayable_path, normpath, pipeline, syspath
@@ -46,6 +46,7 @@ class ImportSession:
     _is_resuming: dict[bytes, bool]
     _merged_items: set[PathBytes]
     _merged_dirs: set[PathBytes]
+    _seen_track_keys: set[tuple[Any, ...]]
 
     def __init__(
         self,
@@ -74,6 +75,7 @@ class ImportSession:
         self._is_resuming = {}
         self._merged_items = set()
         self._merged_dirs = set()
+        self._seen_track_keys = set()
 
         # Normalize the paths.
         self.paths = list(map(normpath, paths or []))
@@ -307,6 +309,25 @@ class ImportSession:
             for path in paths
         }
         self._merged_dirs.update(dirs)
+
+    def track_key_seen(self, key: tuple[Any, ...]) -> bool:
+        """Return whether a per-track duplicate ``key`` was already claimed by
+        an earlier task in this import run (see :ref:`duplicate_track_resolution`).
+
+        This lets the track-level duplicate check catch tracks imported earlier
+        in the *same* run before they reach the database, which matters under
+        the default threaded import where a later task may be checked before an
+        earlier one has been added.
+        """
+        return key in self._seen_track_keys
+
+    def remember_track_key(self, key: tuple[Any, ...]) -> None:
+        """Record a per-track duplicate ``key`` claimed by the current task.
+
+        Only ever called from the single-threaded ``resolve_track_duplicates``
+        pipeline stage, so the shared set needs no locking.
+        """
+        self._seen_track_keys.add(key)
 
     def is_resuming(self, toppath: PathBytes) -> bool:
         """Return `True` if user wants to resume import of this path.
--- a/beets/importer/session.py
+++ b/beets/importer/session.py
@@ -185,7 +185,17 @@ class ImportSession:
     def choose_item(self, task: SingletonImportTask) -> TrackMatch | Action:
         raise NotImplementedError
 
-    def run(self) -> None:
+    def resolve_track_duplicates(self, task: ImportTask, duplicates) -> str:
+        """Decide what to do with album tracks that already exist in the
+        library. Return ``"s"`` (skip the duplicate tracks), ``"k"`` (keep
+        all) or ``"r"`` (remove the old items).
+        """
+        raise NotImplementedError
+
+    def choose_item(self, task: ImportTask):
+        raise NotImplementedError
+
+    def run(self):
         """Run the import task."""
         self.logger.info("import started {}", time.asctime())
         self.set_config(config["import"])
@@ -204,6 +214,10 @@ class ImportSession:
             if self.config["group_albums"] and not self.config["singletons"]:
                 # Split directory tasks into one task for each album.
                 stages += [stagefuncs.group_albums(self)]
+
+            # Optionally drop or replace album tracks that already exist in
+            # the library before the autotag lookup runs.
+            stages += [stagefuncs.resolve_track_duplicates(self)]
 
             # These stages either talk to the user to get a decision or,
             # in the case of a non-autotagged import, just choose to
--- a/beets/importer/stages.py
+++ b/beets/importer/stages.py
@@ -165,10 +165,10 @@ def group_albums(session: ImportSession) -> StageCoro:
         return
 
     action = _track_duplicate_action()
-    if action == "a":
+    if action is DuplicateAction.ASK:
         action = session.resolve_track_duplicates(task, duplicates)
 
-    if action == "s":
+    if action is DuplicateAction.SKIP:
         for item in duplicates:
             log.info(
                 colorize("text_warning", "Skipping duplicate track: {}"),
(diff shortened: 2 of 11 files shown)


tommyschnabel:
@semohr this is ready for review

snejus:
Hi @tommyschnabel, thanks for this work! I would suggest to rebase as there has been a considerable amount of changes and see how refactored duplicates handling by introducing `DuplicateAction`. See how you'd be able to use/reuse it here?

tommyschnabel:
@snejus got your changes integrated! Ready for a review 😄

tommyschnabel:
@semohr ready for review again 😄

beets/test/helper.py
```diff
@@ -669,6 +670,17 @@ def choose_match(self, task):
 
     choose_item = choose_match
 
+    Resolution = Enum("Resolution", "REMOVE SKIP KEEPBOTH MERGE")
```
snejus: Why are we introducing a new Enum here? Can't we reuse `DuplicateAction`?

beets/ui/commands/import_/session.py
```diff
@@ -201,6 +201,28 @@ def get_duplicate_action(
 
         return action
 
+    def resolve_track_duplicates(self, task, duplicates) -> str:
+        """Decide what to do with album tracks already in the library."""
+        log.warning("Some tracks are already in the library!")
+
+        if config["import"]["quiet"]:
```
snejus, line 218: Cast these config options to `bool`:
```suggestion
        if config["import"]["quiet"].get(bool):
```

beets/ui/commands/import_/session.py
```diff
@@ -201,6 +201,28 @@ def get_duplicate_action(
 
         return action
 
+    def resolve_track_duplicates(self, task, duplicates) -> str:
```
snejus: We want to see types here!

docs/reference/config.rst
```diff
@@ -850,6 +850,66 @@ is applied, which would, considering the default, look like this:
...
+- ``remove`` removes the matching old items from the library before importing.
+- ``keep`` (and ``merge``) import the album unchanged.
+- ``ask`` prompts you to choose one of the above.
+
+When left empty, this falls back to :ref:`duplicate_action`.
+
+A typical configuration for completing partially-imported albums while
+autotagging looks like this:
+
+::
```
snejus: Use

```suggestion
.. code-block:: yaml
```

for syntax highlighting.

docs/reference/config.rst
```diff
@@ -850,6 +850,66 @@ is applied, which would, considering the default, look like this:
...
+
+    import:
+        duplicate_track_resolution: yes
+        duplicate_action: ask          # whole-album duplicates
+        duplicate_track_action: skip   # per-track duplicates: fold new tracks into the existing album
+        duplicate_keys:
+            item: mb_trackid           # match on a stable id (recommended when autotagging)
```
snejus: Users will find it helpful to see the default values as well!

beets/importer/session.py
```diff
@@ -205,6 +213,10 @@ def run(self) -> None:
                 # Split directory tasks into one task for each album.
                 stages += [stagefuncs.group_albums(self)]
 
+            # Optionally drop or replace album tracks that already exist in
+            # the library before the autotag lookup runs.
+            stages += [stagefuncs.resolve_track_duplicates(self)]
```
snejus: How is this expected to interact with the default `threaded: yes` mode?

Looking at the pipeline ordering, I wonder whether this stage may check a second task before the first one reaches `add()`. For example, if two one-track albums with the same duplicate keys are imported in one invocation, could both pass this check in threaded mode while the second one would be skipped in sequential mode? I'm not sufficiently familiar with the pipeline internals to tell whether this is possible, so could you clarify what guarantees we have here?

Have you tested such scenario yourself?

beets/importer/stages.py
```diff
@@ -327,10 +424,54 @@ def _apply_choice(session: ImportSession, task: ImportTask) -> None:
         task.set_fields(session.lib)
 
 
+def _track_duplicate_action() -> str:
+    """Return the single-letter action for per-track duplicate resolution.
+
+    Uses ``import.duplicate_track_action`` when set, otherwise falls back to
+    ``import.duplicate_action``.
+    """
+    choices = {
```
snejus: Can we re-use `DuplicateAction` here? See how it used in such cases:

```py
    def get_duplicate_action(
        self, task: ImportTask, found_duplicates: list[AnyLibModel]
    ) -> DuplicateAction:
        """Get the configured duplicate action."""
        choice = config["import"]["duplicate_action"].as_choice(
            DuplicateAction.choices()
        )
        log.debug("default action for duplicates: {}", choice)
        return DuplicateAction(choice)  # type: ignore[call-arg]
```

Then, in the logic I want to see conditionals like `action is DuplicationAction.SKIP` instead of `action == "s"`.

beets/importer/stages.py
```diff
@@ -119,6 +120,99 @@ def group(item: library.Item) -> tuple[str | None, str | None]:
...
+                next(iter(duplicates)).album,
+            )
+            task.set_choice(Action.SKIP)
+            return
+        # Only some tracks were duplicates; we have already dropped them, so
+        # don't let the album-level check skip the rest.
+        task.duplicate_tracks_resolved = True
+        # Fold the remaining new tracks into the existing album, if the
+        # matched duplicates all belong to a single one.
+        album_ids = {
```
snejus: Could filtering out matches with `album_id is None` affect how the fold target is selected here?

For example, if one duplicate belongs to an album and another duplicate is a singleton, would `album_ids` contain only the first duplicate's album ID and cause the remaining tracks to be folded into that album? The documentation says that new tracks are imported as their own album when the matching tracks do not all belong to a single album, so I wonder whether singleton matches should prevent folding in this case. Is this scenario handled elsewhere?

I would like to see a test covering the following situation:

- Incoming album: tracks A, B, and C.
- Existing track A belongs to album X.
- Existing track B is a singleton.
- Both A and B match incoming tracks.

Would you mind adding this case and making the intended outcome for track C explicit? I think this would clarify whether it should be folded into album X or imported as its own album.

beets/test/helper.py
```diff
@@ -669,6 +670,17 @@ def choose_match(self, task):
...
+
+    default_resolution = "REMOVE"
+
+    def resolve_track_duplicates(self, task, duplicates):
```
snejus: Actually, I doubt that we need this any more. You should be able to control this using the configuration. See the commit where I removed it for good: ca3fdb7a

You can step through commits under this PR to see how I removed the need to redefine anything to do with duplicates in `helper.py`: https://github.com/beetbox/beets/pull/6702/commits

beets/importer/session.py
```diff
@@ -185,6 +185,14 @@ def get_duplicate_action(
     def choose_item(self, task: SingletonImportTask) -> TrackMatch | Action:
         raise NotImplementedError
 
+    def resolve_track_duplicates(self, task: ImportTask, duplicates) -> str:
```
snejus: Presumably this is going to be a list of `Item`s?

```suggestion
    def resolve_track_duplicates(self, task: ImportTask, duplicates: list[Item]) -> str:
```

snejus (changes requested):
Well done! The first round of reviews :)

beets/test/helper.py
tommyschnabel: Ah we totally can!

docs/reference/config.rst
tommyschnabel: Playing around with this more I think I'm going to add a note here that 
```
    duplicate_keys:
        album: albumartist album
        item: artist title
```
is also a really good config if some of the tracks you're importing are missing `mb_trackid`. I ran into this situation and had a bunch of duplicates, despite using this branch for imports because the tracks we're labeled correctly

beets/importer/session.py
tommyschnabel: Good call here, this is confusing. It's actually a `dict[Item, list[Item]]`, a map of incoming item to any found matches. I'll add the type

beets/importer/stages.py
tommyschnabel: Yep definitely can/should reuse

beets/importer/stages.py
tommyschnabel: Ah yeah this is a bug, I'll fix it so that A and B are skipped but C is imported

beets/importer/session.py
tommyschnabel: Yeah this is completely valid. I see two ways forward here:
1. We make the concession that we're only checking against already-imported tracks, or
2. We add an extra layer of key caching so that we can check against two of the same tracks being imported in the same session

2 adds more complexity, but I do like the completeness of it. Do you have any opinions on either of these? It does seem like 1. is a pretty narrow case, but it could be hit

tommyschnabel: Actually, I went ahead and added 2. It's in an isolated commit so if you want me to remove it then that's no problem with me 😄

tommyschnabel:
@snejus thanks for the review! I think I've address all your comments!

beets/importer/stages.py
```diff
@@ -119,6 +120,125 @@ def group(item: library.Item) -> tuple[str | None, str | None]:
...
+
+    This runs before :func:`lookup_candidates` so that dropped tracks are
+    excluded from the autotag match. Singleton imports are handled by the
+    regular duplicate resolution and are ignored here.
+    """
+    if (
+        task.skip
+        or not task.is_album
+        or not task.items
+        or not config["import"]["duplicate_track_resolution"].get(bool)
```
semohr: We should use `session.config` for consistency. This allows allows us to have multiple sessions with different configurations in the same runtime. Not too necessary for beets but for the beets as library usecase.

beets/importer/stages.py
```diff
@@ -119,6 +120,125 @@ def group(item: library.Item) -> tuple[str | None, str | None]:
...
+        for item in duplicates:
+            log.info(
+                colorize("text_warning", "Skipping duplicate track: {}"),
+                displayable_path(item.path),
+            )
+            task.items.remove(item)
+        if not task.items:
+            # Every track was a duplicate: skip the whole album.
+            log.info(
+                colorize(
```
semohr: Do not colorize log messages. If we need to show a warning just use `log.warning`.

beets/importer/stages.py
```diff
@@ -327,10 +450,48 @@ def _apply_choice(session: ImportSession, task: ImportTask) -> None:
...
+    return DuplicateAction(choice)  # type: ignore[call-arg]
+
+
+def _find_track_duplicates(
+    lib: library.Library, item: library.Item, keys: list[str]
+) -> list[library.Item]:
+    """Return library items matching `item` on all `keys`, excluding the
+    item itself (so re-imports do not match their own files).
+
+    Unlike :meth:`Item.duplicates_query`, this matches *every* library item,
```
semohr: I dont understand the comment. Isnt the function exactly the same code? What am I missing? 

Image

beets/ui/commands/import_/session.py
```diff
@@ -201,6 +201,31 @@ def get_duplicate_action(
 
         return action
 
+    def resolve_track_duplicates(
+        self, task: importer.ImportTask, duplicates: dict[Item, list[Item]]
+    ) -> DuplicateAction:
+        """Decide what to do with album tracks already in the library."""
+        log.warning("Some tracks are already in the library!")
+
+        if config["import"]["quiet"].get(bool):
```
semohr: Why not reuse the `get_duplicate_action` function? It already supports `AnyModel` and this function does not seem to do much more different to me.

beets/ui/commands/import_/session.py
```diff
@@ -201,6 +201,31 @@ def get_duplicate_action(
...
+            return DuplicateAction.SKIP
+
+        existing = [item for matches in duplicates.values() for item in matches]
+        ui.print_("Old: " + summarize_items(existing, True))
+        if config["import"]["duplicate_verbose_prompt"].get(bool):
+            for item in existing:
+                print(f"  {item}")
+
+        ui.print_("New: " + summarize_items(list(duplicates), True))
+        if config["import"]["duplicate_verbose_prompt"].get(bool):
```
semohr: Use `_report_item_summary` instead.

beets/importer/session.py
```diff
@@ -205,6 +223,10 @@ def run(self) -> None:
                 # Split directory tasks into one task for each album.
                 stages += [stagefuncs.group_albums(self)]
 
+            # Optionally drop or replace album tracks that already exist in
+            # the library before the autotag lookup runs.
+            stages += [stagefuncs.resolve_track_duplicates(self)]
```
semohr: Why do we run this before autotagging? My assumption is that the metadata at this stage won't necessarily match the better-normalized metadata that ends up in the database after autotagging. Am I missing something?

For example:
```
# Original metadata (non-normalized)
artist: djFresH

# Metadata after autotagging and selecting a match
artist: dj-fresh
```

beets/importer/tasks.py
```diff
@@ -153,6 +170,16 @@ def __init__(
         items: Iterable[library.Item] | None,
     ) -> None:
         super().__init__(toppath, paths, items)
+        # Existing library items to remove because individual tracks of this
+        # album duplicate them (see ``duplicate_track_resolution``).
+        self.duplicate_track_items_to_remove: list[library.Item] = []
```
semohr: Im not too keen on adding these state variables here in task. Any other ways we could solve this?

semohr (commented):
I might be misunderstanding the intended workflow here. Even after reading the documentation, the PR description, and the code itself, I'm still not entirely sure what the intended use case is.

Couldn't duplicate resolution happen as a separate step in the pipeline after a match has been chosen? That would align better with my mental model, since we'd be working with the normalized metadata rather than the "raw" tags. Track and album level de-duplication should happen depending on the chosen candidate not the audios metadata imo.

I understand that the duplicate check is currently performed before autotagging, but I'm struggling to see what advantages that provides over doing it afterward.



---

Could we clearly define the intended behavior and use case at a high level. Having an abstract description of the workflow and its goals would make it much easier to reason about whether the current implementation is the right approach.

tommyschnabel:
Hey @semohr thanks for the review! I'm pretty new to this repo still, and so I'm really relying on your and @snejus's architectural suggestions. Making this a new step really makes sense, let me look into doing that. For reference, I had an earlier PR adding this check to the `duplicates` plugin, which @snejus helpfully pointed out would belong better in the main import code, so that's how we got here.

As a general point, it's rarely wanted behavior to import the same track multiple times, and obviously deduping is the way to stop this (my original use-case was re-importing a whole downloads dir multiple times before clearing it, and ending up with duplicates because of that). In my original implementation there was no concept of deduping albums; I was specifically targeting the track level only, and only looking at the album as a whole if it becomes empty due to skipping dedupes. 

So that brings me to the high level behavior of: 
**1. when an incoming track has a matching track in the db, we apply the config setting for deduping (keep, skip, remove, ask)
2. if all tracks from an album import are skipped, the album is also skipped**

I really do appreciate the reviews and the pointers on how you guys want this organized. Again, I'm very new to the repo and am happy to change whatever to help see this feature through.

semohr:
Alright that clears things up a bit for me.

> 1. when an incoming track has a matching track in the db, we apply the config setting for deduping (keep, skip, remove, ask)

What do you mean by incoming track? In general, I would avoid using `items` for this use case and instead operate on `candidates`, since they already contain the processed and normalized metadata. This is also how album-level deduplication works: we operate on matched candidates, and if we detect duplicates for any retrieved candidate, we mark it accordingly and let the user decide how to proceed.

I would approach track-level deduplication in the same way for consistency and unification.

> 2. if all tracks from an album import are skipped, the album is also skipped

I would fold this into a single decision point for the user (config-based behavior also possible). From the UI layer perspective, I would expect something like this:

```
Duplicates detected:

Album-level duplicate:
  • Actions available: Skip | Merge | Keep old | Replace with new

Track-level duplicates (found x duplicates out of y):
  • Actions available: Skip | Merge | Keep old | Replace with new | Select per track
```

---

We haven't really but this in the docs yet but if you need a primer on the notations and different parts of the matching flows have a look here: https://github.com/beetbox/beets/issues/6683

tommyschnabel:
Sorry for the delay @semohr @snejus , ready for another round of reviews 😄

beets/importer/tasks.py
```diff
@@ -154,6 +171,11 @@ def __init__(
     ) -> None:
         super().__init__(toppath, paths, items)
         self.is_album = True
+        # Per-track duplicate state: each entry maps an item of this task
+        # to the existing library items it duplicates, and to the action
+        # chosen for it.
+        self.track_duplicates: dict[library.Item, list[library.Item]] = {}
```
semohr, line 282: Can we find a different solution here? I would try to separate the intermediate/derived state (`track_duplicates`) from the `Task`. The Task class has too many responsibilities imo already. @snejus Do you have an opinion here?

tommyschnabel:
Went ahead and updated since there were conflicts 😄

JOJ0:
> Went ahead and updated since there were conflicts 😄

Love your work in these parts of beets BTW and merging this into my personal main branch right now 🤞

JOJ0:
BTW I didn't give this a proper review but realize looking at diffs passing by that there is quite verbose commenting (possibly agent-generated). So a general note/advice: Comments in complex areas like this are useful and appreciated but try to reduce them to a minimum. For example here the comment is redundant. The simple code and log message tells that same story already:


```
+        if config["import"]["quiet"]:
+            # In quiet mode, don't prompt -- just skip the duplicate tracks.
+            log.info("Skipping duplicate tracks.")
+            return dict.fromkeys(track_duplicates, DuplicateAction.SKIP)
```

Or alternatively, which is also less distracting in my opinion: Move code explanations you find useful to the docstring.

beets/importer/session.py
```diff
@@ -185,7 +185,64 @@ def get_duplicate_action(
...
+        task: ImportTask,
+        track_duplicates: dict[library.Item, list[library.Item]],
+    ) -> dict[library.Item, DuplicateAction]:
+        """Get the configured action for each track that duplicates items
+        already in the library.
+
+        ``track_duplicates`` maps each :class:`~beets.library.Item` of the
+        task to the existing library items it duplicates.
+        """
+        view = config["import"]["duplicate_tracks_action"]
```
semohr, line 199:
```suggestion
        if view := config["import"]["duplicate_tracks_action"].get():
```

beets/importer/session.py
```diff
@@ -185,7 +185,64 @@ def get_duplicate_action(
...
+        if view.get():
+            # MERGE is not offered per track, so reject it loudly here.
+            choice = view.as_choice(DuplicateAction.track_choices())
+            action = DuplicateAction(choice)  # type: ignore[call-arg]
+        else:
+            choice = config["import"]["duplicate_action"].as_choice(
```
semohr, line 205: We should be able to use `get_duplicate_action` here. Less duplication imo.

beets/importer/stages.py
```diff
@@ -334,35 +346,125 @@ def _apply_choice(session: ImportSession, task: ImportTask) -> None:
...
+    For album tasks with ``import.duplicate_tracks`` enabled, individual
+    tracks that duplicate existing library items are detected here as well,
+    so both levels are decided at a single point before anything is added
+    to the library.
     """
-    if task.choice_flag in (Action.ASIS, Action.APPLY, Action.RETAG):
-        found_duplicates = task.find_duplicates(session.lib)
-        if found_duplicates:
-            log.debug("found duplicates: {}", [o.id for o in found_duplicates])
+    if task.choice_flag not in (Action.ASIS, Action.APPLY, Action.RETAG):
```
semohr, line 355: Lets add a debug log here when we are skipping duplicate resolution.

beets/importer/stages.py
```diff
@@ -334,35 +346,125 @@ def _apply_choice(session: ImportSession, task: ImportTask) -> None:
...
+        return
 
-            task.duplicate_action = session.get_duplicate_action(
-                task, found_duplicates
-            )
+    found_duplicates = task.find_duplicates(session.lib)
+
+    track_duplicates: dict[library.Item, list[library.Item]] = {}
+    if task.is_album and session.config["duplicate_tracks"].get(bool):
```
semohr, line 361: Is `config["duplicate_tracks"]` always set? `session.config["duplicate_tracks"].get(bool)` can error if it is not!

beets/importer/tasks.py
```diff
@@ -590,6 +604,77 @@ def find_duplicates(self, lib: library.Library) -> list[library.Album]:
...
+            # same way `apply_metadata` would modify them.
+            pairs = []
+            for item, data in self.match.merged_pairs:
+                tmp_item = library.Item(lib, **dict(item))
+                tmp_item.update(data)
+                pairs.append((item, tmp_item))
+        elif self.choice_flag in (Action.ASIS, Action.RETAG):
+            pairs = [(item, item) for item in self.items]
+        else:
+            return {}
```
semohr, line 639: I noticed that, style-wise, you tend to use `return` statements quite frequently. Having multiple returns throughout a function body can make the control flow harder to follow. Could you try for a single `return` per function, or at most one additional early return when it clearly improves readability?

docs/changelog.rst
```diff
@@ -82,9 +90,8 @@ New features
 - :doc:`plugins/lyrics`: Added a ``rest_directory`` configuration option for
   specifying a reStructuredText output directory, semantically equivalent to
   ``-r, --write-rest``. :bug:`2806`
-- A database backup is now automatically created before running schema
-  migrations. Control with the ``create_backup_before_migrations`` option
-  (default: yes).
+- A database backup is now automatically created before running schema (default:
```
semohr, line 93: Unrelated change

beets/importer/tasks.py
```diff
@@ -63,6 +63,23 @@
...
 
 
+def _remove_duplicate_item(
+    lib: library.Library, item: library.Item, with_album: bool = True
+):
+    """Remove ``item`` from ``lib`` and delete its file when it lives inside
+    the library directory, pruning any newly-empty parent directories.
+    """
+    item.remove(with_album=with_album)
+    if lib.directory in util.ancestry(item.path):
```
semohr, line 73: I would move the delete files part of this function into the `item.remove` method. Seems appropriate to me as we could remove this helper function in its entirety. Maybe something like

```python
def item.remove(self, with_album=True, remove_file=False, clutter=()):
    ...
    if remove_file and self.library.directory in util.ancestry(self.path):
        log.debug("deleting {.filepath}", item)
        util.remove(self.path)
        util.prune_dirs(
            os.path.dirname(self.path),
            self.library.directory,
            clutter=clutter,
        )
```

@snejus You agree?

semohr (changes requested):
Had another short iteration here. I’ll have another in-depth look once these comments have been addressed.

Sorry that we’re being quite rigorous here. Since this PR includes changes to the core, we want to make sure it remains stable and that we won’t have to revisit or maintain unnecessary complexity going forward.

beets/importer/session.py
tommyschnabel: Reinstated in 8deea9a. The redesign had dropped the seen-key cache; the session now keeps `seen_track_keys`, which `TrackDuplicates.find` consults and extends. Keys are claimed per task (not per item), so two copies of the same track *within* one album are still left alone, and only the single-threaded `resolve_duplicates` stage touches the set, so no locking is needed. Added `test_seen_keys_catch_duplicate_not_yet_in_library`, which drives detection over two tasks with no `add()` in between against an empty library.

beets/importer/tasks.py
tommyschnabel: Done in 8deea9a: `_apply_track_duplicate_skips` and `_fold_into_existing_album` now have a single return (plus one guard), and `_resolve_duplicates` keeps only the choice-flag guard.

beets/importer/tasks.py
tommyschnabel: Partly done: the derived state is now its own `TrackDuplicates` dataclass in `tasks.py` (detection, per-item actions, `remove_old`), so `ImportTask` no longer carries any of that logic — only a single `task.track_duplicates` attribute.

I kept that attribute on the task because the state has to survive across three pipeline stages (`resolve_duplicates` finds and resolves it, `apply_choice` applies the skips, `manipulate_files` removes the superseded old items), and the task is the only thing that travels between stages. Parking it on the session in a dict keyed by task would need locking (stages run in separate threads) and would leak entries for the life of the run. Happy to move it if you prefer a different home — what did you have in mind?

tommyschnabel: Do you have any recommendations here? It's now just one additional property that holds everything, but happy to do something else if you'd like!

tommyschnabel:
@semohr sorry for the long delay! I believe I've addressed all your comments and am ready for another review 😄

Issue #1153, opened 2014-12-19
ftintitle: Run after mbsync


donutz0:
the plugin ftintitle is mentioned in my config, but after mbsync the featuring is back in the artist fields.
for me it would be applicable, if a mbsync follows the tagging behavior of a reimport.

m-urban:
Sounds like this could be solved by `ftintitle` catching our new events `albuminfo_received` and `trackinfo_received` (discussed in #872, introduced in #1499)…

treyturner:
Fixed in https://github.com/beetbox/beets/releases/tag/v2.13.0

PR #6726, opened 2026-06-10
feat(ftintitle): hook metadata events, fixing mbsync

treyturner:
Depends on https://github.com/beetbox/beets/pull/6732.

## Summary

This PR integrates `ftintitle` with metadata fetched through beets’ metadata-received events, so commands like `mbsync` apply already-normalized metadata when `ftintitle` is enabled.

Addresses https://github.com/beetbox/beets/issues/1153.

## Changes

- Register `ftintitle` listeners for `trackinfo_received` and `albuminfo_received` when `ftintitle.auto` is enabled.
- Rewrite fetched `TrackInfo` metadata before `mbsync` applies it, avoiding unnecessary re-sync churn.
- Preserve existing manual command and import behavior while sharing the same rewrite logic.
- Handle `artist_credit`, cached `Info` properties, empty titles, and no-op configurations consistently.
- Update `ftintitle` docs for the expanded `auto` behavior.

## To Do

### DIFF
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -149,6 +149,8 @@ class Info(AttrDict[Any]):
         return set(config["overwrite_null"][cls.type.lower()].as_str_seq())
 
     def __setitem__(self, key: str, value: Any) -> None:
+        self._invalidate_cached_properties()
+
         # handle legacy info.str_field = "abc" and info["str_field"] = "abc"
         if list_field := self.LEGACY_TO_LIST_FIELD.get(key):
             self[list_field] = self._get_list_from_string_value(
@@ -156,6 +158,12 @@ class Info(AttrDict[Any]):
             )
         else:
             super().__setitem__(key, value)
+
+    def _invalidate_cached_properties(self) -> None:
+        for cls in type(self).__mro__:
+            for field, attr in vars(cls).items():
+                if isinstance(attr, cached_property):
+                    self.__dict__.pop(field, None)
 
     @property
     def id(self) -> str | None:
--- a/beets/autotag/hooks.py
+++ b/beets/autotag/hooks.py
@@ -149,8 +149,6 @@ class Info(AttrDict[Any]):
         return set(config["overwrite_null"][cls.type.lower()].as_str_seq())
 
     def __setitem__(self, key: str, value: Any) -> None:
-        self._invalidate_cached_properties()
-
         # handle legacy info.str_field = "abc" and info["str_field"] = "abc"
         if list_field := self.LEGACY_TO_LIST_FIELD.get(key):
             self[list_field] = self._get_list_from_string_value(
@@ -158,12 +156,6 @@ class Info(AttrDict[Any]):
             )
         else:
             super().__setitem__(key, value)
-
-    def _invalidate_cached_properties(self) -> None:
-        for cls in type(self).__mro__:
-            for field, attr in vars(cls).items():
-                if isinstance(attr, cached_property):
-                    self.__dict__.pop(field, None)
 
     @property
     def id(self) -> str | None:
--- a/beetsplug/ftintitle.py
+++ b/beetsplug/ftintitle.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import re
 from functools import cached_property, lru_cache
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
 
 from beets import config, plugins, ui
 
@@ -12,6 +12,8 @@ if TYPE_CHECKING:
     from beets.autotag import AlbumInfo, Info, TrackInfo
     from beets.importer import ImportSession, ImportTask
     from beets.library import Album, Item
+
+FeaturedField = Literal["artist", "artist_credit", "artist_sort"]
 
 
 DEFAULT_BRACKET_KEYWORDS: tuple[str, ...] = (
@@ -262,26 +264,22 @@ class FtInTitlePlugin(plugins.BeetsPlugin):
             _album_artist_no_feat
         )
 
-    @staticmethod
-    def _field_value(metadata: object, field: str) -> str:
-        return getattr(metadata, field, None) or ""
-
-    def _effective_artist(self, metadata: object) -> str:
+    def _effective_artist(self, metadata: Info | Item) -> str:
         if self.artist_credit:
-            return self._field_value(
-                metadata, "artist_credit"
-            ) or self._field_value(metadata, "artist")
-        return self._field_value(metadata, "artist")
+            return metadata.get("artist_credit") or metadata.get("artist") or ""
+        return metadata.get("artist") or ""
 
     def _strip_featured_from_field(
-        self, metadata: object, field: str, for_artist: bool = True
+        self,
+        metadata: Info | Item,
+        field: FeaturedField,
+        for_artist: bool = True,
     ) -> None:
-        value = self._field_value(metadata, field)
-        if value:
+        if value := metadata.get(field):
             stripped, _ = split_on_feat(
                 value, for_artist=for_artist, custom_words=self.custom_words
             )
-            setattr(metadata, field, stripped)
+            metadata[field] = stripped
 
     def commands(self) -> list[ui.Subcommand]:
         def func(lib, opts, args):
@@ -336,7 +334,7 @@ class FtInTitlePlugin(plugins.BeetsPlugin):
                 "artist: {.artist} (Not changing due to keep_in_artist)", item
             )
         else:
-            artist = self._field_value(item, "artist")
+            artist = item.get("artist") or ""
             track_artist, _ = split_on_feat(
                 artist, custom_words=self.custom_words
             )
@@ -345,14 +343,11 @@ class FtInTitlePlugin(plugins.BeetsPlugin):
             changed |= artist != track_artist
 
             if self.artist_credit:
-                artist_credit = self._field_value(item, "artist_credit")
+                artist_credit = item.get("artist_credit") or ""
                 self._strip_featured_from_field(item, "artist_credit")
-                changed |= artist_credit != self._field_value(
-                    item, "artist_credit"
-                )
+                changed |= artist_credit != (item.get("artist_credit") or "")
 
-        artist_sort = self._field_value(item, "artist_sort")
-        if artist_sort:
+        if artist_sort := item.get("artist_sort"):
             # Just strip the featured artist from the sort name.
             artist_sort_no_feat, _ = split_on_feat(
                 artist_sort, custom_words=self.custom_words
@@ -362,7 +357,7 @@ class FtInTitlePlugin(plugins.BeetsPlugin):
 
         # Only update the title if it does not already contain a featured
         # artist and if we do not drop featuring information.
-        title = self._field_value(item, "title")
+        title = item.get("title") or ""
         if not self.drop_feat and not contains_feat(title, self.custom_words):
             formatted = self.feat_format.format(feat_part)
             new_title = self.insert_ft_into_title(
@@ -379,9 +374,9 @@ class FtInTitlePlugin(plugins.BeetsPlugin):
         changed = False
         if not self.keep_in_artist_field:
             before = (
-                self._field_value(info, "artist"),
-                self._field_value(info, "artist_credit"),
-                self._field_value(info, "artist_sort"),
+                info.get("artist"),
+                info.get("artist_credit"),
+                info.get("artist_sort"),
             )
             self._strip_featured_from_field(info, "artist")
             if self.artist_credit:
@@ -390,12 +385,12 @@ class FtInTitlePlugin(plugins.BeetsPlugin):
                 )
             self._strip_featured_from_field(info, "artist_sort")
             changed |= before != (
-                self._field_value(info, "artist"),
-                self._field_value(info, "artist_credit"),
-                self._field_value(info, "artist_sort"),
+                info.get("artist"),
+                info.get("artist_credit"),
+                info.get("artist_sort"),
             )
 
-        title = self._field_value(info, "title")
+        title = info.get("title") or ""
         if not self.drop_feat and not contains_feat(title, self.custom_words):
             formatted = self.feat_format.format(feat_part)
             new_title = self.insert_ft_into_title(
@@ -429,7 +424,7 @@ class FtInTitlePlugin(plugins.BeetsPlugin):
             True if the item has been modified. False otherwise.
         """
         artist = self._effective_artist(item).strip()
-        albumartist = self._field_value(item, "albumartist")
+        albumartist = item.get("albumartist") or ""
         if not self._has_feat_candidate(artist, albumartist):
             return False
 
--- a/beetsplug/ftintitle.py
+++ b/beetsplug/ftintitle.py
@@ -267,7 +267,7 @@ class FtInTitlePlugin(plugins.BeetsPlugin):
             write = ui.should_write()
 
             for item in lib.items(args):
-                if self.ft_in_title(item):
+                if self.ft_in_title(item, item.get("albumartist") or ""):
                     item.store()
                     if write:
                         item.try_write()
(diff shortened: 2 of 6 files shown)


beetsplug/ftintitle.py
```diff
@@ -201,32 +211,65 @@ def __init__(self) -> None:
...
+        self, metadata: object, field: str, custom_words: list[str]
+    ) -> None:
+        value = self._field_value(metadata, field)
+        if value:
+            stripped, _ = split_on_feat(value, custom_words=custom_words)
+            setattr(metadata, field, stripped)
+
+    def _configured_options(self) -> FtInTitleOptions:
+        return FtInTitleOptions(
+            self.config["drop"].get(bool),
```
snejus: Define each of these options as a `@cached_property` instead.

snejus (changes requested):
Let's split the following into a separate PR:

Define each configuration option as a `@cached_property` to remove the need to send them through each function. See 34772d2 where I did the same for `convert` plugin.

This will remove a big chunk of the current PR diff and make it easier to review the functionality that you added.

treyturner:
The cached properties are now in https://github.com/beetbox/beets/pull/6732 on which this PR is based. The remaining changes can be seen in the single subsequent commit:

[`ff085e4` (this PR)](https://github.com/beetbox/beets/pull/6726/changes/ff085e43bb9e4af3ed1a69e7801d6fe8f6d76ea3)

snejus:
Now rebase it on master and I'll review it again :)

snejus:
Now just rebase it again since your fix has been merged!

beets/autotag/hooks.py
```diff
@@ -171,6 +173,12 @@ def __setitem__(self, key: str, value: Any) -> None:
         else:
             super().__setitem__(key, value)
 
+    def _invalidate_cached_properties(self) -> None:
```
snejus: Why is this needed? I don't think that `item_info` is accessed until later/display in the pipeline?

beetsplug/ftintitle.py
```diff
@@ -257,11 +269,34 @@ def __init__(self) -> None:
...
         self.import_stages = [self.imported]
+        self.register_listener("trackinfo_received", self.trackinfo_received)
+        self.register_listener("albuminfo_received", self.albuminfo_received)
 
         self.album_template_fields["album_artist_no_feat"] = (
             _album_artist_no_feat
         )
 
+    @staticmethod
+    def _field_value(metadata: object, field: str) -> str:
```
snejus:
```suggestion
    def _field_value(metadata: Info, field: str) -> str:
```

beetsplug/ftintitle.py
```diff
@@ -257,11 +269,34 @@ def __init__(self) -> None:
...
+        return getattr(metadata, field, None) or ""
```
snejus:
```suggestion
        return metadata.get(field) or ""
```

beetsplug/ftintitle.py
```diff
@@ -257,11 +269,34 @@ def __init__(self) -> None:
...
+
+    def _effective_artist(self, metadata: object) -> str:
+        if self.artist_credit:
+            return self._field_value(
```
snejus: I don't think there's a need for `_field_value` helper. Just use `metadata.get` instead.

beetsplug/ftintitle.py
```diff
@@ -257,11 +269,34 @@ def __init__(self) -> None:
...
+                metadata, "artist_credit"
+            ) or self._field_value(metadata, "artist")
+        return self._field_value(metadata, "artist")
+
+    def _strip_feature_from_field(
```
snejus:
```suggestion
    def _strip_featured_from_field(
```

beetsplug/ftintitle.py
```diff
@@ -257,11 +269,34 @@ def __init__(self) -> None:
...
+        self, metadata: object, field: str, for_artist: bool = True
+    ) -> None:
+        value = self._field_value(metadata, field)
+        if value:
```
snejus:
```suggestion
        if value := metadata.get(field):
```

beetsplug/ftintitle.py
```diff
@@ -257,11 +269,34 @@ def __init__(self) -> None:
...
+            stripped, _ = split_on_feat(
+                value, for_artist=for_artist, custom_words=self.custom_words
+            )
+            setattr(metadata, field, stripped)
```
snejus: Remember that `Info` acts like a `dict`!
```suggestion
            metadata[field] = stripped
```

beetsplug/ftintitle.py
```diff
@@ -257,11 +269,34 @@ def __init__(self) -> None:
...
+
+    def _effective_artist(self, metadata: object) -> str:
+        if self.artist_credit:
+            return self._field_value(
+                metadata, "artist_credit"
+            ) or self._field_value(metadata, "artist")
+        return self._field_value(metadata, "artist")
+
+    def _strip_feature_from_field(
+        self, metadata: object, field: str, for_artist: bool = True
```
snejus: Would it make sense to define a `Literal` for `field`? I think you're only working with a few fields here?

beetsplug/ftintitle.py
```diff
@@ -285,41 +320,120 @@ def imported(self, session: ImportSession, task: ImportTask) -> None:
...
+            if self.artist_credit:
+                artist_credit = self._field_value(item, "artist_credit")
+                self._strip_feature_from_field(item, "artist_credit")
+                changed |= artist_credit != self._field_value(
+                    item, "artist_credit"
+                )
 
-        if item.artist_sort:
+        artist_sort = self._field_value(item, "artist_sort")
+        if artist_sort:
```
snejus: Try to use walrus operator in such cases, and as I said above no need for `_field_value`:
```suggestion
        if artist_sort := item.get("artist_sort"):
```

snejus (changes requested):
Had the first review - will have another look once these comments are addressed!

beets/autotag/hooks.py
treyturner: I stumbled on it during testing but agree it's overly defensive. Removed 👍

treyturner:
I /think/ these are all addressed, but yes please have another look as updates were quite scattered.  I really appreciate the reviews and help. Every few months when I get time to pick up more beets work I have to remember everything I learned about the data model and then forgot since the last hiatus 🤤

snejus:
> I have to remember everything I learned about the data model and then forgot since the last hiatus 🤤

It's also been regularly changing as we've been refactoring it... 😆

beetsplug/ftintitle.py
```diff
@@ -257,11 +271,30 @@ def __init__(self) -> None:
...
+        self.register_listener("albuminfo_received", self.albuminfo_received)
 
         self.album_template_fields["album_artist_no_feat"] = (
             _album_artist_no_feat
         )
 
+    def _effective_artist(self, metadata: Info | Item) -> str:
+        if self.artist_credit:
+            return metadata.get("artist_credit") or metadata.get("artist") or ""
+        return metadata.get("artist") or ""
```
snejus: Extract `metadata.get("artist") or ""` to a variable to avoid doubling of the logic.

beetsplug/ftintitle.py
```diff
@@ -257,11 +271,30 @@ def __init__(self) -> None:
...
+        self.register_listener("trackinfo_received", self.trackinfo_received)
+        self.register_listener("albuminfo_received", self.albuminfo_received)
 
         self.album_template_fields["album_artist_no_feat"] = (
             _album_artist_no_feat
         )
 
+    def _effective_artist(self, metadata: Info | Item) -> str:
+        if self.artist_credit:
+            return metadata.get("artist_credit") or metadata.get("artist") or ""
```
snejus: I think including `artist_credit` is out of scope of your change. This may have unintended side effects: `artist_credit` is populated **very often**, which means that `artist` won't get cleaned up.

I'd be leaning towards cleaning both fields, but we should discuss it in a separate PR.

beetsplug/ftintitle.py
```diff
@@ -285,41 +318,117 @@ def imported(self, session: ImportSession, task: ImportTask) -> None:
             if self.ft_in_title(item):
                 item.store()
 
-    def update_metadata(self, item: Item, feat_part: str) -> None:
+    def trackinfo_received(self, info: TrackInfo) -> None:
+        """Move featuring artists in fetched singleton metadata."""
+        if not self.auto:
```
snejus: Maybe worth moving this conditional out of each of the two methods into `__init__`?

beetsplug/ftintitle.py
```diff
@@ -285,41 +318,117 @@ def imported(self, session: ImportSession, task: ImportTask) -> None:
             if self.ft_in_title(item):
                 item.store()
 
-    def update_metadata(self, item: Item, feat_part: str) -> None:
+    def trackinfo_received(self, info: TrackInfo) -> None:
```
snejus: ✍🏼 Nit:

I'm used to seeing higher level methods defined _below_ low-level methods. As in, I expected to see `ft_in_info` defined _above_ `trackinfo_received` and `album_received`. I'd say just move these two _below_ `ft_in_info`.

snejus (changes requested):
Another quick review round: as I mentioned in the comments, inclusion of `artist_credit` will have unintended side-effects and we'd rather this separately.

Will have a final review once it's updated!

beetsplug/ftintitle.py
treyturner: This method ended up going away based on the other comments. The defensive logic is still repeated several times in the file, but seems more appropriate now as its used to process different fields. Let me know if you'd like it reduced more aggressively still and I can take another look once I know everything else is settled 👍

beetsplug/ftintitle.py
treyturner: Understood; I backed it out and adjusted the tests to match.

beetsplug/ftintitle.py
```diff
@@ -257,11 +266,30 @@ def __init__(self) -> None:
...
+        self,
+        metadata: Info | Item,
+        field: FeaturedField,
+        for_artist: bool = True,
+    ) -> None:
+        if value := metadata.get(field):
+            stripped, _ = split_on_feat(
+                value, for_artist=for_artist, custom_words=self.custom_words
+            )
+            metadata[field] = stripped
```
snejus: Why was this method introduced? This is applied within `update_info_metadata`, but isn't applied within `update_item_metadata`.

Furthermore, this method edits the structure that was passed into it - not a good idea. Best to remove it

beetsplug/ftintitle.py
```diff
@@ -285,41 +313,94 @@ def imported(self, session: ImportSession, task: ImportTask) -> None:
...
-                item.title, formatted, self.bracket_keywords
+                title, formatted, self.bracket_keywords
             )
             self._log.info("title: {.title} -> {}", item, new_title)
             item.title = new_title
+            changed |= title != new_title
+
+        return changed
+
+    def update_info_metadata(self, info: Info, feat_part: str) -> bool:
```
snejus: Why are `Item` and `Info` updated using separate methods? Is there any difference in the logic?  
 
It seems like we should be able to use a single `update_metadata` method instead.

beetsplug/ftintitle.py
```diff
@@ -328,31 +409,31 @@ def ft_in_title(self, item: Item) -> bool:
         Returns:
             True if the item has been modified. False otherwise.
         """
-        artist = item.artist.strip()
```
snejus: Why has this method been adjusted?

For example, `item.get("artist") or ""` guard is a bit pointless - `Item.artist` is always a `str`.

beetsplug/ftintitle.py
```diff
@@ -285,41 +313,94 @@ def imported(self, session: ImportSession, task: ImportTask) -> None:
...
+        """Move featuring artists in fetched singleton metadata."""
+        self.ft_in_info(info)
+
+    def albuminfo_received(self, info: AlbumInfo) -> None:
+        """Move featuring artists in fetched album track metadata."""
+        albumartist = info.get("artist") or ""
+        for track_info in info.tracks:
+            self.ft_in_info(track_info, albumartist)
 
     def ft_in_title(self, item: Item) -> bool:
```
snejus: I see you've taken the right direction and defined `ft_in_title` and `ft_in_info` to be very similar to each other. We don't need a second method here - just define

```py
    def ft_in_title(self, item: Item | Info, albumartist: str) -> bool:
```

and remove `ft_in_info`.

Note you will probably need to guard this line in `ft_in_title` implementation with `hasattr("filepath")` or something, since `filepath` is not defined on `Info`.

```py
        self._log.info("{.filepath}", item)
```

test/plugins/test_ftintitle.py
```diff
@@ -246,6 +252,217 @@ def test_ftintitle_functional(
...
+            plugins.send("trackinfo_received", info=info)
+
+        assert info.artist == "Alice"
+        assert info.artist_credit == "Alice"
+        assert info.artist_sort == "Alice"
+        assert info.artists == ["Alice", "Bob"]
+        assert info.artists_credit == ["Alice"]
+        assert info.title == "Song feat. Bob"
+        assert info.item_data["artist"] == "Alice"
+        assert info.item_data["title"] == "Song feat. Bob"
```
snejus: No need to assert these two since they don't seem to be functionally related to your changes.

beetsplug/ftintitle.py
```diff
@@ -131,14 +134,20 @@ def find_feat_part(
     """Attempt to find featured artists in the item's artist fields and
     return the results. Returns None if no featured artist found.
     """
+    if not albumartist:
+        _, feat_part = split_on_feat(
+            artist, for_artist=True, custom_words=custom_words
```
snejus: Why is this method adjusted?

test/plugins/test_ftintitle.py
```diff
@@ -246,6 +252,217 @@ def test_ftintitle_functional(
...
+            title="Song",
+        )
+
+        with self.configure_plugin({"auto": True}):
+            plugins.send("trackinfo_received", info=info)
+
+        assert info.artist == "Alice"
+        assert info.artist_credit == "Alice & Bobby"
+        assert info.title == "Song feat. Bob"
+
```
snejus: This is essentially a copy of `test_trackinfo_received_preserves_artist_credit_when_disabled`.


```suggestion
```

test/plugins/test_ftintitle.py
```diff
@@ -246,6 +252,217 @@ def test_ftintitle_functional(
...
+        )
+
+        with self.configure_plugin({"auto": True}):
+            self.run_command("ftintitle")
+
+        item.load()
+        assert item.artist == "Alice"
+        assert item.artist_credit == "Alice feat. Bobby"
+        assert item.title == "Song feat. Bob"
+
```
snejus: We removed `artist_credit` handling.

```suggestion
```

snejus (changes requested):

test/plugins/test_ftintitle.py
treyturner: The tests aimed to make the current behavior explicit and support subsequent discussion or possible upcoming changes you hinted around mutating `artist_credit`, but I'll remove them 👍

beetsplug/ftintitle.py
treyturner: Removed

beetsplug/ftintitle.py
treyturner: Good call, abstracted.

test/plugins/test_ftintitle.py
treyturner: 🤔 True for `item_data["artist"]` which is sourced from `artist_credit` here. I removed that assertion, but kept `item_data["title"]` since that's part of the metadata surface later applied by `mbsync`, and the title rewrite there is directly related to this change.

beetsplug/ftintitle.py
treyturner: This allows singleton-fetched metadata to handle generic separators like `Alice & Bob` without album-artist context. With the dust around this PR having settled some though, it's more obvious this falls outside the core aim. I removed it and can/will open a super small fast-follow with it if you'd support that.

beetsplug/ftintitle.py
treyturner: Yep! Failed to notice those had converged. It's now abstracted just as you suggested, thanks 👍

test/plugins/test_ftintitle.py
treyturner: Removed

snejus (approved):
Looks great, thanks!

PR #6727, opened 2026-06-10
eDit: Enable album-level field editing during interactive import

arsaboo:
## Description

During interactive import, the `edit` plugin only exposed item-level fields — the `albumfields` configuration was completely ignored in `importer_edit()`. Users who wanted to modify album-level fields such as `album` or `year` had to complete the import, then run `beet edit -a QUERY` as a separate step.

### Solution

Updated `importer_edit()` to support editing album-level fields during import. Album fields (configured via `albumfields`) are presented as a YAML header section prepended to the per-track documents. Changes to the header are applied to all items in the album, making album-level edits work in both the `eDit` and `edit Candidates` choices.

### DIFF
--- a/beetsplug/edit.py
+++ b/beetsplug/edit.py
@@ -221,71 +221,47 @@ class EditPlugin(plugins.BeetsPlugin):
         """
         # Get the content to edit as raw data structures.
         old_data = [flatten(o, fields) for o in objs]
-
-        # Set up a temporary file with the initial data for editing.
-        new = NamedTemporaryFile(
-            mode="w", suffix=".yaml", delete=False, encoding="utf-8"
-        )
-        old_str = dump(old_data)
-        new.write(old_str)
-        new.close()
+        cur_str = dump(old_data)
 
         # Loop until we have parseable data and the user confirms.
-        try:
-            while True:
-                # Ask the user to edit the data.
-                edit(new.name, self._log)
+        while True:
+            result = self._edit_yaml(cur_str)
+            if result is None:
+                return False
+            new_data, new_str = result
 
-                # Read the data back after editing and check whether anything
-                # changed.
-                with codecs.open(new.name, encoding="utf-8") as f:
-                    new_str = f.read()
-                if new_str == old_str:
-                    ui.print_("No changes; aborting.")
-                    return False
+            # Show the changes.
+            # If the objects are not on the DB yet, we need a copy of their
+            # original state for show_model_changes.
+            objs_old = [obj.copy() if obj.id < 0 else None for obj in objs]
+            self.apply_data(objs, old_data, new_data)
+            changed = False
+            for obj, obj_old in zip(objs, objs_old):
+                changed |= ui.show_model_changes(obj, obj_old)
+            if not changed:
+                ui.print_("No changes to apply.")
+                return False
 
-                # Parse the updated data.
-                try:
-                    new_data = load(new_str)
-                except ParseError as e:
-                    ui.print_(f"Could not read data: {e}")
-                    if ui.input_yn("Edit again to fix? (Y/n)", True):
-                        continue
-                    return False
-
-                # Show the changes.
-                # If the objects are not on the DB yet, we need a copy of their
-                # original state for show_model_changes.
-                objs_old = [obj.copy() if obj.id < 0 else None for obj in objs]
-                self.apply_data(objs, old_data, new_data)
-                changed = False
-                for obj, obj_old in zip(objs, objs_old):
-                    changed |= ui.show_model_changes(obj, obj_old)
-                if not changed:
-                    ui.print_("No changes to apply.")
-                    return False
-
-                # For cancel/keep-editing, restore objects to their original
-                # in-memory state so temp edits don't leak into the session
-                choice = ui.input_options(
-                    ("continue Editing", "apply", "cancel")
-                )
-                if choice == "a":  # Apply.
-                    return True
-                if choice == "c":  # Cancel.
-                    self.apply_data(objs, new_data, old_data)
-                    return False
-                if choice == "e":  # Keep editing.
-                    self.apply_data(objs, new_data, old_data)
-                    continue
-
-        # Remove the temporary file before returning.
-        finally:
-            os.remove(new.name)
+            # For cancel/keep-editing, restore objects to their original
+            # in-memory state so temp edits don't leak into the session
+            choice = ui.input_options(("continue Editing", "apply", "cancel"))
+            if choice == "a":  # Apply.
+                return True
+            if choice == "c":  # Cancel.
+                self.apply_data(objs, new_data, old_data)
+                return False
+            if choice == "e":  # Keep editing.
+                self.apply_data(objs, new_data, old_data)
+                cur_str = new_str
+                continue
 
     def apply_data(self, objs, old_data, new_data):
         """Take potentially-updated data and apply it to a set of Model
         objects.
+
+        Documents are matched to objects by their ``id`` field rather than
+        by position, so a reordered or otherwise misaligned document list
+        cannot cause one object's data to be applied to a different object.
 
         The objects are not written back to the database, so the changes
         are temporary.
@@ -298,8 +274,16 @@ class EditPlugin(plugins.BeetsPlugin):
             )
 
         obj_by_id = {o.id: o for o in objs}
+        old_by_id = {d.get("id"): d for d in old_data}
         ignore_fields = self.config["ignore_fields"].as_str_seq()
-        for old_dict, new_dict in zip(old_data, new_data):
+        for new_dict in new_data:
+            new_id = new_dict.get("id")
+            old_dict = old_by_id.get(new_id)
+            obj = obj_by_id.get(new_id)
+            if old_dict is None or obj is None:
+                self._log.warning("ignoring object whose id changed")
+                continue
+
             # Prohibit any changes to forbidden fields to avoid
             # clobbering `id` and such by mistake.
             forbidden = False
@@ -311,8 +295,7 @@ class EditPlugin(plugins.BeetsPlugin):
             if forbidden:
                 continue
 
-            id_ = int(old_dict["id"])
-            apply_(obj_by_id[id_], new_dict)
+            apply_(obj, new_dict)
 
     def save_changes(self, objs):
         """Save a list of updated Model objects to the database."""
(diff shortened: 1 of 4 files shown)


beetsplug/edit.py
```diff
@@ -220,76 +220,80 @@ def edit(self, album, objs, fields):
...
-
-        # Set up a temporary file with the initial data for editing.
         new = NamedTemporaryFile(
             mode="w", suffix=".yaml", delete=False, encoding="utf-8"
         )
-        old_str = dump(old_data)
         new.write(old_str)
         new.close()
 
-        # Loop until we have parseable data and the user confirms.
```
snejus, line 226: Before I review, again, please undo the removal of all comments.

snejus (changes requested):

beetsplug/edit.py
arsaboo: Apologies....restored the explanatory comments and reverted any unrelated changes.

snejus:
Would you mind adding types to methods you introduced and the ones you adjusted?

arsaboo:
> Would you mind adding types to methods you introduced and the ones you adjusted?

Done!

beetsplug/edit.py
```diff
@@ -331,7 +338,107 @@ def before_choose_candidate_listener(self, session, task):
...
+        first_item = task.items[0]
+        header = flatten(first_item, album_fields)
+        return header if header else None
+
+    def _importer_edit_apply_header(
+        self, items: list[Item], header_data: dict[str, Any]
+    ) -> None:
+        """Apply album-header changes (from ``_importer_edit_album_header``)
+        to every item in the list.
+        """
```
snejus:
```suggestion
        """Apply album-header changes to every item in the list."""
```

beetsplug/edit.py
```diff
@@ -331,7 +338,107 @@ def before_choose_candidate_listener(self, session, task):
...
+    def _edit_yaml(
+        self, old_str: str
+    ) -> tuple[list[dict[str, Any]], str] | None:
+        """Open a temporary file with `old_str`, let the user edit it, and
+        return the parsed list of YAML documents.
+
+        Returns ``(parsed_data, edited_str)`` on success, or ``None`` if the
+        user aborted (no changes or unresolvable parse error).
+        """
+        new = NamedTemporaryFile(
```
snejus: Is it not possible to use it as a context manager, where it closes automatically?

snejus (changes requested):
I tried testing it, and adjusted **album** catalognum, see the resulting output:

Image

Seems like `path`, `title` and `track` for every track got reset to the first track's values?

arsaboo:
@snejus Thanks for the screenshot. `_importer_edit_apply_tracks` was matching YAML documents back to items by list position (`zip(old_track_data, new_track_data, items)`), instead of by `id` like the existing `apply_data` already does. Unfortunately, positional matching is fragile, which matches exactly what you saw.

Fixed it to mirror `apply_data's` approach. I also added two tests (`ApplyTracksMatchingTest`) that reproduce the failure mode directly — both fail against the old positional code and pass with the fix

beetsplug/edit.py
arsaboo: done

arsaboo:
@snejus I also realized there is some code duplication we can refactor, and that we can fix the position-trusting bug you highlighted earlier across the plugin. Let me know if you would prefer that. 

edit: I went ahead and refactored the code to simplify the plugin and remove the dead/duplicate code in 
314b059. I can revert that if you prefer not to.

beetsplug/edit.py
```diff
@@ -291,8 +274,16 @@ def apply_data(self, objs, old_data, new_data):
             )
 
         obj_by_id = {o.id: o for o in objs}
+        old_by_id = {d.get("id"): d for d in old_data}
```
snejus, line 284: If a user changes doc 2 id to existing id 1, code applies doc 2 data to item 1 instead of ignoring forbidden id change. I think, we need to detect duplicate ids and id edits before applying (similar to what was done previously).

beetsplug/edit.py
```diff
@@ -343,21 +396,104 @@ def importer_edit(self, session, task):
...
+                    f"Number of documents changed from {expected_total} to "
+                    f"{len(new_all_data)}."
+                )
+                if ui.input_yn("Edit again to fix? (Y/n)", True):
+                    continue
+                self._importer_edit_cleanup(task)
+                return None
+
+            # Split into album header and per-track documents.
+            new_header_data = new_all_data[:num_data_docs] if has_header else []
```
snejus: Could we make the album header unambiguous before applying it?

Right now, when a header is present, we always treat the first YAML document as the header:

```py
new_header_data = new_all_data[:num_data_docs] if has_header else []
new_track_data = new_all_data[num_data_docs:]
```

This can go wrong if the user moves a track document above the header, which is plausible now that `apply_data` explicitly supports reordered track documents. In that case, `_importer_edit_apply_header()` would apply track fields such as `id`, `track`, and `title` to every item.

I think we should either reject a header document containing `id`, or add some other explicit way to distinguish the header from track documents before applying it.

arsaboo: Went with the former.

beetsplug/edit.py
arsaboo: Good catch. I updated `apply_data` to now check for duplicate IDs across the edited document. If the two end up claiming the same, neither is applied; we log a warning and leave those items untouched rather than guessing which one was intended. I also added a test for this case.

snejus:
I've just checked out this branch to test it but I think I'm still seeing the same issue as before:

image

snejus:
Could you please test it yourself?

arsaboo:
@snejus I verified the fix on the current `edit` branch.

1. Can you please check if the edit branch is correctly installed:
```bash
python3 - <<'PY'
import inspect
import beets
import beetsplug.edit as edit

print("beets_version:", getattr(beets, "__version__", None))
print("beets_file:", beets.__file__)
print("edit_file:", edit.__file__)

apply_src = inspect.getsource(edit.EditPlugin.apply_data)
importer_src = inspect.getsource(edit.EditPlugin.importer_edit)

print("has_duplicate_guard:", "new_id_counts" in apply_src)
print("has_header_by_id:", "new_header_data = [d for d in new_all_data if" in importer_src)
print("uses_positional_header_split:", "new_all_data[:num_data_docs]" in importer_src)
PY
```
Expected result on the fixed edit branch:
```
has_duplicate_guard: True
has_header_by_id: True
uses_positional_header_split: False
```

Next, the reordered-header failure mode must preserve per-track fields and only apply the album-level change.

```python
python3 - <<'PY'
import inspect
from beetsplug.edit import EditPlugin, flatten

plugin = EditPlugin()
plugin.config["albumfields"] = "catalognum"
plugin.config["itemfields"] = "track title path"

items = [
    Item(id=1, track=1, title="Title 1", path="/tmp/track1.mp3", catalognum="OLD"),
    Item(id=2, track=2, title="Title 2", path="/tmp/track2.mp3", catalognum="OLD"),
    Item(id=3, track=3, title="Title 3", path="/tmp/track3.mp3", catalognum="OLD"),
]

task = type("Task", (), {"is_album": True, "items": items})()

album_fields = set(plugin.config["albumfields"].as_str_seq())
track_fields = set(plugin.config["itemfields"].as_str_seq()) - album_fields
track_fields.add("id")

header_data = plugin._importer_edit_album_header(task)
old_track_data = [flatten(o, track_fields) for o in task.items]

new_header = dict(header_data)
new_header["catalognum"] = "NEW"

# Reordered docs: tracks first, album header last.
new_all_data = [*old_track_data, new_header]

# Follow whatever logic the installed importer_edit() actually contains.
importer_src = inspect.getsource(plugin.importer_edit)
if 'new_header_data = [d for d in new_all_data if "id" not in d]' in importer_src:
    new_header_data = [d for d in new_all_data if "id" not in d]
    new_track_data = [d for d in new_all_data if "id" in d]
else:
    num_data_docs = 1 if header_data is not None else 0
    new_header_data = new_all_data[:num_data_docs] if header_data is not None else []
    new_track_data = new_all_data[num_data_docs:]

if new_header_data:
    plugin._importer_edit_apply_header(task.items, new_header_data[0])
plugin.apply_data(task.items, old_track_data, new_track_data)

for item in items:
    path = item.path.decode() if isinstance(item.path, bytes) else item.path
    print((item.id, item.track, item.title, path, item.catalognum))
PY
```
Expected result on the fixed edit branch:
```
(1, 1, 'Title 1', '/tmp/track1.mp3', 'NEW')
(2, 2, 'Title 2', '/tmp/track2.mp3', 'NEW')
(3, 3, 'Title 3', '/tmp/track3.mp3', 'NEW')
```
That shows the reordered album header no longer clobbers id, track, title, or path, and only the album-level field is applied. 

Can you tell me the exact command you are running and if I can reproduce what you are seeing?

snejus:
The commit I'm on:

```
*   5 hours ago               Alok Saboo f91ba3a31 Merge branch 'master' into edit (HEAD -> edit)
```

I am running `beet import -LI 200 bibi` (album that has multiple tracks). Then I edit `country` album attribute.

arsaboo:
@snejus I checked again (across multiple tests), and it is working as expected. Here's the log from a similar test:

image

I tried it again, this time changing it to XW and it still worked perfectly:

image


Can you please reinstall: '`pip install --user --force-reinstall --no-deps git+https://github.com/arsaboo/beets.git@edit` and try again.

snejus:
See a video recording: [asciicast](https://asciinema.org/a/6eApTnNlWKaxSV42)

snejus:
See my `edit` configuration:

```yaml
edit:
  ignore_fields: []
  albumfields:
    - album
    - artist
    - artists
    - artist_sort
    - artwork
    - artpath
    - title
    - label
    - year
    - month
    - day
    - track
    - catalognum
    - album
    - albumartist
    - albumstatus
    - albumtype
    - albumtypes
    - catalognum
    - comments
    - comp
    - country
    - data_source
    - data_url
    - day
    - disctitle
    - genre
    - media
    - label
    - mb_albumid
    - mb_artistid
    - mediums
    - medium
    - medium_total
    - path
    - albumtotal
    - month
    - style
    - year
  itemfields:
    - album
    - albumartist
    - album_id
    - artist
    - artists
    - artist_sort
    - artwork
    - title
    - label
    - year
    - month
    - day
    - track
    - catalognum
    - albumstatus
    - albumtype
    - albumtypes
    - artwork_url
    - bpm
    - comments
    - country
    - cuelist
    - data_source
    - data_url
    - disc
    - disctitle
    - disctotal
    - genre
    - genres
    - index
    - length
    - lyrics
    - mb_albumid
    - mb_artistid
    - mb_releasegroup_id
    - mb_releasetrack_id
    - mb_trackid
    - media
    - medium
    - medium_index
    - medium_total
    - mediums
    - path
    - play_count
    # - rg_track_gain
    # - rg_track_peak
    - rating
    - skip_count
    - style
    - track_alt
    - track_id
    - tracklist
    - tracktotal
    - visual_url
```

arsaboo:
can you please run this and confirm:

```python
python3 - <<'PY'
import inspect
import beets
import beetsplug.edit as edit

print("beets_version:", getattr(beets, "__version__", None))
print("beets_file:", beets.__file__)
print("edit_file:", edit.__file__)

apply_src = inspect.getsource(edit.EditPlugin.apply_data)
importer_src = inspect.getsource(edit.EditPlugin.importer_edit)

print("has_duplicate_guard:", "new_id_counts" in apply_src)
print("has_header_by_id:", "new_header_data = [d for d in new_all_data if" in importer_src)
print("uses_positional_header_split:", "new_all_data[:num_data_docs]" in importer_src)
PY
```

I am following the exact same process, and it is working fine at my end. I have a simpler `edit` config:
```yaml
edit:
  albumfields: album year albumartist country
```

snejus:
```
beets_version: 2.12.0
beets_file: /home/sarunas/repo/beets/beets/__init__.py
edit_file: /home/sarunas/repo/beets/beetsplug/edit.py
has_duplicate_guard: True
has_header_by_id: True
uses_positional_header_split: False
```

snejus:
Try using my configuration when you try to reproduce - maybe it has to do with that?

arsaboo:
I see it now (it is a different bug): the problem is your `albumfields` config is incorrect and includes item-only fields like track, title, path, and catalognum alongside genuine album fields (album, year, country, etc.). The plugin never validates that `albumfields` only contains genuinely album-level fields. 

I updated the plugin to restrict `_importer_edit_album_header` to a known-safe allowlist of true album-level fields (ignoring anything not actually on the `Album` model, or anything that isn't part of `AlbumMatch`). 

@snejus Please try your query now.

snejus:
Perfect, that works now. What if I want to add flexible fields? Maybe it's better to ignore item-fields specifically there?

arsaboo:
I updated the logic to now strip item-only fixed fields (`track`, `title`) but no longer discard flexible fields. We expect the user to correctly configure flexible attributes, as there is no way for us to know if they are album-level or item-level fields.

snejus:
Perfect, that seems to work for me!

snejus (approved):

Issue #6710, opened 2026-06-05
`beet tidal --auth` broken in 2.11.0 for headless server


cletusw:
### Problem

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv tidal --auth
user configuration: /config/config.yaml
data directory: /config
plugin paths: []
Loading plugins: tidal
Sending event: pluginload
tidal: 'tidal.source_weight' configuration option is deprecated and will be removed in version 3.0.0. Use 'tidal.data_source_mismatch_penalty' instead.
library database: /config/musiclibrary.blb
library directory: /music
Sending event: library_opened
Paste redirected URL:
```

Not sure what "Paste redirected URL:" means in this context. Maybe if this wasn't a headless machine accessed via SSH it would automatically open a browser?

arsaboo/beets-tidal (which I've been using successfully til now) works fine in a headless environment, logging a https://link.tidal.com/XXXXX link that I can copy into my laptop's browser.

On a related note, it's very confusing and unfortunate that the built-in `tidal` plugin overrides my existing `beets-tidal` setup on update with AFAICT no log informing me that `beets-tidal` is now being silently ignored.

### Setup

* OS: Docker lscr.io/linuxserver/beets:2.10.0 on TrueNAS Community 25.04.2.3
* Python version: 3.12.13
* beets version: 2.11.0
* Turning off plugins made problem go away (yes/no): no
* Installed arsaboo/beets-tidal via git

My configuration (output of `beet config`) is:

```yaml
# --------------- Plugins ---------------

plugins: tidal
directory: /music
# --------------- Main ---------------

library: /config/musiclibrary.blb
art_filename: albumart
original_date: no

# --------------- Tagging ---------------

per_disc_numbering: no
tidal:
    tidal_session_file: /config/tidal.json
    data_source_mismatch_penalty: 0.5
    source_weight: 0.5
    search_limit: 5
    client_id: REDACTED
    tokenfile: tidal_token.json
disabled_plugins: []
```

treyturner:
Run with `BROWSER=echo beet tidal --auth` 👍

semohr:
Interesting. In theory, the logic should work like this:

```python
try:
    webbrowser.open(auth_url)
except webbrowser.Error:
    ui.print_(f"Visit: {auth_url}")
redirect_url = ui.input_("Paste redirected URL: ")
```

[...]

PR #6730, opened 2026-06-11
feat(lyrics): Tidal as lyrics source

treyturner:
## Description

- Add TIDAL API support for fetching track lyric relationships, including typed lyric resources and configurable OAuth scopes.
- Add `tidal` as an opt-in lyrics source with plain/synced lyric selection, token scope validation, defensive JSON:API response handling, and match-debug parity with search backends.
- Document TIDAL lyrics configuration, `user.read` authentication requirements, and scope formats.

## Testing

- `poetry run ruff format --check --diff beetsplug/tidal/__init__.py beetsplug/tidal/api.py beetsplug/tidal/api_types.py beetsplug/tidal/session.py test/plugins/test_tidal.py`
- `poetry run ruff check beetsplug/tidal/__init__.py beetsplug/tidal/api.py beetsplug/tidal/api_types.py beetsplug/tidal/session.py test/plugins/test_tidal.py`
- `poetry run pytest test/plugins/test_tidal.py`
- `poetry run ruff format --check --diff beetsplug/lyrics.py test/plugins/test_lyrics.py`
- `poetry run ruff check beetsplug/lyrics.py test/plugins/test_lyrics.py`
- `poetry run pytest test/plugins/test_lyrics.py`
- `poetry run docstrfmt --preserve-adornments --check docs/plugins/lyrics.rst docs/plugins/tidal.rst docs/changelog.rst`
- `poetry run sphinx-lint --enable all --disable default-role docs/plugins/lyrics.rst docs/plugins/tidal.rst docs/changelog.rst`
- `env -u NO_COLOR poetry run pytest`

## To Do

### DIFF
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 import itertools
+import json
 import math
 import re
 import textwrap
@@ -30,6 +31,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, ClassVar, NamedTuple
 from urllib.parse import quote, quote_plus, urlencode, urlparse
 
+import confuse
 import requests
 from bs4 import BeautifulSoup
 from unidecode import unidecode
@@ -53,8 +55,6 @@ from ._utils.requests import (
 if TYPE_CHECKING:
     from collections.abc import Callable, Iterable, Iterator
 
-    import confuse
-
     from beets.importer import ImportTask
     from beets.library import Library
     from beets.logging import BeetsLogger as Logger
@@ -65,6 +65,13 @@ if TYPE_CHECKING:
         JSONDict,
         LRCLibAPI,
         TranslatorAPI,
+    )
+    from .tidal.api import TidalAPI
+    from .tidal.api_types import (
+        ResourceIdentifier,
+        TidalArtist,
+        TidalLyrics,
+        TidalTrack,
     )
 
     HtmlTransformer = Callable[[str], str]
@@ -289,6 +296,10 @@ class Backend(LyricsRequestHandler, metaclass=BackendClass):
     def __init__(self, config: confuse.Subview, log: Logger) -> None:
         self._log = log
         self.config = config
+
+    @cached_property
+    def dist_thresh(self) -> float:
+        return self.config["dist_thresh"].get(float)
 
     def fetch(
         self, artist: str, title: str, album: str, length: int
@@ -588,10 +599,6 @@ class SearchResult(NamedTuple):
 
 
 class SearchBackend(SoupMixin, Backend):
-    @cached_property
-    def dist_thresh(self) -> float:
-        return self.config["dist_thresh"].get(float)
-
     def check_match(
         self, target_artist: str, target_title: str, result: SearchResult
     ) -> bool:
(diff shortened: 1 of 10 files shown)


beetsplug/tidal/__init__.py
```diff
@@ -52,9 +56,20 @@ def __init__(self) -> None:
...
+            scope=self._scope(),
             token_path=self._tokenfile(),
         )
 
+    def _scope(self) -> str:
+        """Return the configured OAuth scope string."""
+        scope = self.config["scope"].get()
+        if isinstance(scope, str):
+            return " ".join(scope.split())
+        if isinstance(scope, list):
```
semohr: A simple `config.as_str_seq` should work instead of the isinstance checks.

docs/plugins/lyrics.rst
```diff
@@ -110,16 +117,57 @@ The available options are:
...
   and plain text (without timestamps) in the ``USLT`` (unsynchronized lyrics)
   frame, so players that support only one of the two formats can still show the
   correct lyrics.
+- **tidal**: TIDAL API settings for the ``tidal`` source.
+
+  - **client_id**: TIDAL API client ID. This must match the client ID used to
+    create the token file.
+  - **country_code**: ISO 3166-1 alpha-2 country code used for TIDAL catalog and
+    lyrics availability.
+  - **scope**: OAuth scopes to request when authenticating. The TIDAL lyrics
```
semohr: Let's hardcore the capabilities depending on the usage and not require the users to change it.

One thing I'm wondering is what errors are raised if one uses the wrong capability. Is that sufficiently clear from the error message?

semohr (changes requested):
Thanks! It's great to see more activity around the TIDAL plugin. I already did a quick scan of the changes, but I'll do a more thorough review next week.

---

I think we should keep the requested capabilities (`scope`) out of the user configuration. While I understand it's convenient to make this configurable during development, I can foresee quite a few user errors if it's exposed.

I see two possible approaches:

* Run the TIDAL plugin with `search.read user.read` by default.
* Run with only `search.read` by default, and automatically request `user.read` when the `tidal lyrics` plugin is enabled.

I'd vote for the first option. It keeps the configuration simpler and avoids potential issues for users. The `user.read` scope will be needed at some point anyways if we want e.g. to allow users to use their own music.

@snejus Any strong opinion here?

beetsplug/tidal/__init__.py
treyturner: Nice, hadn't spotted `as_str_seq`. I cleaned this up for now while I look at improving the scope UX.

docs/plugins/lyrics.rst
treyturner: Digging into this more: unfortunately TIDAL doesn't fail or even warn when the token lacks lyric access. It returns a successful track response and silently omits lyrics as an empty relationship, indistinguishable from "this track has no lyrics". Given that fact and that the code for managing scope on-demand seems unfun to write and maintain, I've gone with hardcoding the scope 🤤

beetsplug/lyrics.py
```diff
@@ -860,6 +867,260 @@ def scrape(cls, html: str) -> str | None:
...
+                "Could not decode TIDAL token file: {}. "
+                "Run `beet tidal --auth` again.",
+                self.tokenfile,
+            )
+            return set()
+
+        return self.scope_set(token.get("scope") or token.get("scopes"))
+
+    @cached_property
+    def token_has_required_scopes(self) -> bool:
```
semohr, line 927: Lets put the scope safeguards into the general tidal session. Checking that the token includes the correct scope can be done once on token loading. Inside the `load_token` function should work. This should allow users to update their tokens. 

I get where you come from as technically we don't need the `user.read` scope for the non lyrics functions (yet). Most of the logic here looks a bit overengineered to me tho, let's keep it simple. Will spare us some maintenance down the line.

beetsplug/tidal/api.py
```diff
@@ -28,6 +28,7 @@
 
 API_BASE = "https://openapi.tidal.com/v2"
 MAX_FILTER_SIZE = 20
+TIDAL_DEFAULT_SCOPE = "search.read user.read"
```
semohr, line 31: Thanks, that looks like the right place to me!

beetsplug/tidal/api.py
```diff
@@ -39,13 +40,16 @@ def _batched(iterable: Iterable[T], n: int) -> Iterable[list[T]]:
 
 
 class TidalAPI(RequestHandler):
-    def __init__(self, client_id: str, token_path: str) -> None:
+    def __init__(
+        self, client_id: str, token_path: str, scope: str = TIDAL_DEFAULT_SCOPE
```
semohr, line 44: Do we need to be able to overwrite the scopes? I dont think we currently do. Lets remove it if it is not needed :)

beetsplug/lyrics.py
```diff
@@ -860,6 +867,260 @@ def scrape(cls, html: str) -> str | None:
...
+        if text := attributes.get("text"):
+            return text.strip()
+
+        return None
+
+    def fetch_tracks_lyrics(
+        self, track_ids: list[str]
+    ) -> Iterable[tuple[str, TidalLyrics]]:
+        """Fetch lyric resources for TIDAL track IDs."""
+        tracks_data = self.api.get_tracks(
```
semohr, line 1066: Lets rename this for consistency.
```suggestion
        tracks_doc = self.api.get_tracks(
```

beetsplug/lyrics.py
```diff
@@ -860,6 +867,260 @@ def scrape(cls, html: str) -> str | None:
...
+
+        if synced and (lrc_text := attributes.get("lrcText")):
+            return lrc_text.strip()
+
+        if text := attributes.get("text"):
+            return text.strip()
+
+        return None
+
+    def fetch_tracks_lyrics(
```
semohr, line 1062: Can we change the signature to `-> Iterable[TrackInfo | None]:` and move the warnings out of this function. Have a look at `TidalApi.search_tracks_by_ids` I used that approach to keep the order of returned objs to align with the input ids.

docs/plugins/tidal.rst
```diff
@@ -52,6 +52,13 @@ This will open a browser window where you can log in to your Tidal account.
 After successful authentication, your token will be saved to the configured
 token file and you won't need to re-authenticate on subsequent runs.
 
+On headless installations, setting ``BROWSER=echo`` may print the redirect URL
```
semohr, line 55: This is a bug imo. Lets not add the workaround to the docs and fix the underlying issue see #6710.

beetsplug/lyrics.py
```diff
@@ -860,6 +867,260 @@ def scrape(cls, html: str) -> str | None:
...
+
+    def search(self, artist: str, title: str) -> Iterable[TidalTrack]:
+        """Search TIDAL tracks and yield matching track resources."""
+        search_data = self.api.search_results(
+            f"{artist} {title}",
+            include=["tracks.artists"],
+            country_code=self.country_code,
+        )
+
+        search_result = search_data.get("data")
```
semohr, line 1022: Do we need to be defensive here? We are basically doing the same in the tidal plugin without being defensive. Am just wondering.

Same for `relationships = search_result.get("relationships")`

semohr (changes requested):

semohr:
FYI: I just enabled the `user.read` for our tidal app. Im now wondering how requesting the `user.read` scope worked in the first place :thinking: 
image

semohr:
> Document TIDAL lyrics configuration, user.read authentication requirements, and scope formats.

I'm now also wondering which request does need the `user.read` scope in the first place? :smile:

treyturner:
I had initially verified lyrics retrieval via a developer app ID, but once I saw there was a public API that purported to retrieve lyrics I cut over to it. That may have been premature though, as I now can't actually get their public API to return lyrics as documented. I'm moving this back into draft while I figure out if the only path to lyrics is via developer app credentials. Sorry for the false start.

treyturner:
The `user.read` scope was my misunderstanding that we'd get the same data back from the v2 API as I got from the legacy `r_usr` scope on the v1 API, but this seems to not be the case. I can't actually get the v2 API to hand me lyrics with any scope I'm able to configure at the moment, so I'm going to cut this back over to the v1 API totally independent of the built-in `tidal` plugin and will release it as an external plugin. I was too eager to contribute to `tidal` once I saw it existed and didn't identify the risk of functional gaps in the API version change since the documentation suggests they're still retrievable.

I'm really sad it can't land in the project and want to apologize again for the hassle. If I see something change in the API I'll pick it back up; I was also hoping to add Tidal as an art source.

semohr:
> I had initially verified lyrics retrieval via a developer app ID, but once I saw there was a public API that purported to retrieve lyrics I cut over to it. That may have been premature though, as I now can't actually get their public API to return lyrics as documented. I'm moving this back into draft while I figure out if the only path to lyrics is via developer app credentials. Sorry for the false start.

No worries :/ The tidal public api can be a bit difficult to use and sometimes even buggy. If you want I can also have a look at it, maybe we can make it work. You have a curl request that should work?

semohr:
Seems like a known issue https://github.com/orgs/tidal-music/discussions/279

treyturner:
Thanks for pinging on that thread! For what it's worth, I implemented this in [beets-tidalv1](https://pypi.org/project/beets-tidalv1/) / [@treyturner/beets-plugins](https://github.com/treyturner/beets-plugins/tree/main/plugins/tidalv1) until there's news about lyrics in v2.

PR #6729, opened 2026-06-11
lyrics: add rate limiting and exponential backoff to prevent 429 errors

arsaboo:
Add a LyricsSession class that mounts RateLimitAdapter (4 req/sec) and configures urllib3 Retry with exponential backoff on 429 responses. Each backend instance gets its own session so rate limits apply independently per API source.

Fixes #6728

- [x] Changelog. (Add an entry to `docs/changelog.rst` to the bottom of one of the lists near the top of the document.)

PR #6731, opened 2026-06-12
TimeoutAndRetrySession: fold in RateLimitAdapter and 429 retry

arsaboo:
## Description

Follow up to https://github.com/beetbox/beets/pull/6729

Move shared rate-limiting and 429 retry logic from the lyrics-specific `LyricsSession` into the base `TimeoutAndRetrySession` so all plugins benefit. Simplify `lyrics.py` by removing the now-redundant `LyricsSession` subclass.

### DIFF
--- a/beetsplug/_utils/requests.py
+++ b/beetsplug/_utils/requests.py
@@ -71,7 +71,7 @@ class TimeoutAndRetrySession(requests.Session, metaclass=SingletonMeta):
     * raises exceptions for HTTP error status codes
     """
 
-    def __init__(self, rate_limit: float = 0.25, *args, **kwargs) -> None:
+    def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         self.headers["User-Agent"] = f"beets/{__version__} https://beets.io/"
 
@@ -86,7 +86,7 @@ class TimeoutAndRetrySession(requests.Session, metaclass=SingletonMeta):
                 HTTPStatus.TOO_MANY_REQUESTS,
             ],
         )
-        adapter = RateLimitAdapter(rate_limit=rate_limit, max_retries=retry)
+        adapter = RateLimitAdapter(rate_limit=0.25, max_retries=retry)
         self.mount("https://", adapter)
         self.mount("http://", adapter)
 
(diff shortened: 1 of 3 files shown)


beetsplug/_utils/requests.py
```diff
@@ -71,22 +71,22 @@ class TimeoutAndRetrySession(requests.Session, metaclass=SingletonMeta):
     * raises exceptions for HTTP error status codes
     """
 
-    def __init__(self, *args, **kwargs) -> None:
+    def __init__(self, *args, rate_limit: float = 0.25, **kwargs) -> None:
```
snejus: Do we need this argument? Is it used anywhere?

arsaboo: Good point — nothing was using it, so I've removed the parameter and hardcoded 0.25. I kept the 0.25 default in `TimeoutAndRetrySession` because that is the value used in both lyrics/tidal plugins.

semohr: Shouldnt the usage in lyrics be 0.2 (5 req per second)?

snejus: > Shouldnt the usage in lyrics be 0.2 (5 req per second)?

I doubt that any provider can serve requests that quickly anyways?

arsaboo: 0.25 (4 per second) is probably safer, and what is used in Tidal. Makes it easy. If we wanted to, we can initialize `rate_limit=0.2` in lyrics plugin

semohr: > > Shouldnt the usage in lyrics be 0.2 (5 req per second)?
> 
> I doubt that any provider can serve requests that quickly anyways?

For tidal this is a real issue and we run into an extended rate limit if we do more than 4 request per second.

Just as an fyi, modern webframeworks can handle 10s of thousands of requests per second. So 5/s should be more than doable in my opinion 😅
https://web-frameworks-benchmark.netlify.app/

arsaboo: I am fine either way. BTW, if we want to allow plugins to be able to specify rate_limit, we should bring back the `rate_limit` arg in the `__init__`.

snejus: I think for now we should be OK keeping the default `0.25` in lyrics for simplicity and reduce it in the future if we see the need for it.

snejus (approved):

PR #6732, opened 2026-06-12
refactor(ftintitle): cache config options

treyturner:
Pre-factor for https://github.com/beetbox/beets/pull/6726.

## Description

Make `ftintitle`'s config `@cached_property` attributes instead of passing them through the call stack. Modeled after similar recent changes in `convert`.

## Changes

- Add cached plugin properties for `auto`, `drop`, `format`, `keep_in_artist`, `preserve_album_artist`, and `custom_words`, keeping the existing cached `bracket_keywords` property I forgot I added a while back.
- `commands()`, `imported()`, `ft_in_title()`, and `update_metadata()` are cleaned up to read these directly.
- Don't read `auto` during plugin init; the import stage remains registered and `imported()` checks `auto` when invoked.

## To Do

### DIFF
--- a/beetsplug/ftintitle.py
+++ b/beetsplug/ftintitle.py
@@ -140,6 +140,30 @@ def _album_artist_no_feat(album: Album) -> str:
 
 class FtInTitlePlugin(plugins.BeetsPlugin):
     @cached_property
+    def auto(self) -> bool:
+        return self.config["auto"].get(bool)
+
+    @cached_property
+    def drop_feat(self) -> bool:
+        return self.config["drop"].get(bool)
+
+    @cached_property
+    def feat_format(self) -> str:
+        return self.config["format"].as_str()
+
+    @cached_property
+    def keep_in_artist_field(self) -> bool:
+        return self.config["keep_in_artist"].get(bool)
+
+    @cached_property
+    def preserve_album_artist(self) -> bool:
+        return self.config["preserve_album_artist"].get(bool)
+
+    @cached_property
+    def custom_words(self) -> list[str]:
+        return self.config["custom_words"].as_str_seq()
+
+    @cached_property
     def bracket_keywords(self) -> list[str]:
         return self.config["bracket_keywords"].as_str_seq()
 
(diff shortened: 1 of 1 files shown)


beetsplug/ftintitle.py
```diff
@@ -236,71 +247,50 @@ def func(lib, opts, args):
...
-        item: Item,
-        feat_part: str,
-        drop_feat: bool,
-        keep_in_artist_field: bool,
-        custom_words: list[str],
-    ) -> None:
+    def update_metadata(self, item: Item, feat_part: str) -> None:
         """Choose how to add new artists to the title and set the new
         metadata. Also, print out messages about any changes that are made.
-        If `drop_feat` is set, then do not add the artist to the title; just
```
snejus, line 264: Leave it as it was here to minimise the number of unrelated changes.

test/plugins/test_ftintitle.py
```diff
@@ -239,6 +243,26 @@ def test_ftintitle_functional(
         assert item["artist"] == expected_artist
         assert item["title"] == expected_title
 
+    def test_cached_auto_is_read_after_plugin_configuration(self) -> None:
```
snejus: I don't think this is needed.

snejus (changes requested):

test/plugins/test_ftintitle.py
treyturner: Fair enough, I removed it. I tripped over it during the refactor but it's correct now.

beetsplug/ftintitle.py
```diff
@@ -236,71 +247,50 @@ def func(lib, opts, args):
...
         # artist and if we do not drop featuring information.
-        if not drop_feat and not contains_feat(item.title, custom_words):
-            feat_format = self.config["format"].as_str()
-            formatted = feat_format.format(feat_part)
+        if not self.drop_feat and not contains_feat(
+            item.title, self.custom_words
+        ):
+            formatted = self.feat_format.format(feat_part)
             new_title = self.insert_ft_into_title(
                 item.title, formatted, self.bracket_keywords
```
snejus, line 288: Does `self.bracket_keyword` _need_ to be sent into this method or could it also be accessed from within `insert_ft_into_title` instead?

treyturner: It doesn't have to be, but I think it makes a cleaner unit as a `@classmethod`. As this pattern is now very limited, it seems reasonable vs. reducing its testability and refactoring tests. Happy to adjust for either route though 👍

snejus (approved):

PR #6733, opened 2026-06-13
utils: centralise path asciification

snejus:
- This change moves path Unicode normalization and separator cleanup into `beets.util.asciify_path()`, so path shaping now lives in one place instead of being split between `beets.library.models` and `util`.

- Callers in `Item.destination()`, `art_destination()`, and `tmpl_asciify()` now use the same `util.asciify_path()` flow. This makes path generation more consistent and removes duplicate platform-specific normalization logic from the library layer.

- High-level impact: when `asciify_paths` is enabled, all asciified paths now get the same normalization, transliteration, and separator-replacement behavior through one shared utility. This should make path handling easier to reason about and safer to change later.

- Test coverage follows the architecture change: focused asciify behavior tests move from `test/test_library.py` to `test/test_util.py`, so `asciify_path()` is tested directly where the behavior now lives.

### DIFF
--- a/beets/library/models.py
+++ b/beets/library/models.py
@@ -2,9 +2,7 @@ from __future__ import annotations
 
 import os
 import string
-import sys
 import time
-import unicodedata
 from contextlib import suppress
 from functools import cached_property
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


semohr (approved):
Looks fine from my perspective. I think we lost a bit of test coverage here, maybe you want to check.

snejus:
Just fixed the coverage an expanded coverage of the function.

PR #6734, opened 2026-06-13
db: simplify LazyConvertDict

snejus:
- Replaces `LazyConvertDict` with `LazyDict`, a simpler `UserDict`-based container for lazily converted model values.
- Removes the separate `_awaken` construction path and routes database-loaded values through `Model.__init__`, so both normal and database-backed objects follow the same creation flow.
- Centralizes SQL-to-Python conversion in `Model._convert`, which makes value handling easier to reason about and reduces special-case model initialization logic.

**High-level impact**

- Simplifies the model storage architecture by reducing custom dict behavior.
- Makes object construction more consistent across the codebase.
- Lowers maintenance risk by keeping lazy conversion in one place while preserving the existing lazy-loading behavior.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -24,7 +24,7 @@ import sys
 import threading
 import time
 from abc import ABC, abstractmethod
-from collections import defaultdict
+from collections import UserDict, defaultdict
 from collections.abc import Mapping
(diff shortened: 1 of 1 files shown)


semohr (approved):
Neat!

PR #6742, opened 2026-06-14
Migration backup

semohr:
## Description

Creates a timestamped `.bak` copy of the database before any pending schema migration runs, so users can recover if a migration goes wrong. Controlled by the new `create_backup_before_migrations` config option (default: `yes`).

**Changes**:
- `Migration.migrate_model`: before running migration data, copies the database file to `<library.db>.<timestamp>.bak` (once per startup, guarded by a `ClassVar` flag)
- `Database.create_copy(dest)`: copies the SQLite file to the given path

TODOs:
- [x] Documentation.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -1045,8 +1045,22 @@ class Migration(ABC):
         """Run this migration once for a model's backing table."""
         table = model_cls._table
         if not self.db.migration_exists(self.name, table):
+            self._before_migration_backup(table)
             self._migrate_data(model_cls, *args, **kwargs)
             self.db.record_migration(self.name, table)
+
+    def _before_migration_backup(self, table: str) -> None:
+        if not beets.config["create_backup_before_migrations"].get(bool):
+            return
+
+        # In-memory databases have no persistent file to back up.
+        if self.db.path in (":memory:", b":memory:"):
+            return
+
+        dest = os.fsdecode(self.db.path) + f"-before-{table}-{self.name}.bak"
+        self.db.create_backup(dest)
+
+        print(f"Created database backup at: {dest!r}.")
 
     @abstractmethod
     def _migrate_data(
@@ -1353,6 +1367,15 @@ class Database:
                 migration.migrate_model(
                     model_cls, self.db_tables[model_cls._table]["columns"]
                 )
+
+    def create_backup(self, dest: str) -> None:
+        """Create a backup of the database at `dest`."""
+        # Use the SQLite backup API so the copy is consistent even when the
+        # database is open and may have a journal/WAL.
+        dest_conn = sqlite3.connect(dest)
+        with self.transaction():
+            self._connection().backup(dest_conn)
+        dest_conn.close()
 
     def migration_exists(self, name: str, table: str) -> bool:
         """Return whether a named migration has been marked complete."""
--- a/test/library/test_migrations.py
+++ b/test/library/test_migrations.py
@@ -315,3 +315,34 @@ class TestRelativePathMigration(MigrationTestHelper):
         str_item = self.lib.get_item(2)
         assert str_item
         assert str_item.path == abs_bytes_path
+
+
+class TestMigrationBackup(MigrationTestHelper):
+    """Tests for the backup-before-migration feature."""
+
+    migration = (migrations.LyricsMetadataInFlexFieldsMigration, (Item,))
+    db_on_disk = True
+
+    @classmethod
+    def setup_previous_state(cls, monkeypatch):
+        monkeypatch.setattr(
+            "beets.library.models.Item._fields",
+            {**Item._fields, "lyrics": types.STRING},
+        )
+
+    @pytest.mark.parametrize(
+        "config_value, expected_count", [(True, 1), (False, 0)]
+    )
+    def test_backup_config(self, config_value, expected_count):
+        self.config["create_backup_before_migrations"] = config_value
+        self.add_item(lyrics="some lyrics")
+        db_path = self.lib.path
+
+        self.lib._migrate()
+
+        backups = [
+            f
+            for f in os.listdir(os.path.dirname(db_path))
+            if os.fsdecode(f).endswith(".bak")
+        ]
+        assert len(backups) == expected_count
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -3,6 +3,7 @@
 library: library.db
 directory: ~/Music
 statefile: state.pickle
+create_backup_before_migrations: yes
 
 # --------------- Plugins ---------------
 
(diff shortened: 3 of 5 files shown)


beets/dbcore/db.py
```diff
@@ -1045,6 +1047,22 @@ def migrate_model(self, model_cls: type[Model], *args, **kwargs) -> None:
...
+
+                try:
+                    self.db.create_backup(dest)
+                    ui.print_(f"Created database backup at: {dest!r}")
+                except OSError as exc:
+                    ui.print_(
+                        f"Could not create database backup at {dest!r}: {exc}"
+                    )
+                finally:
+                    Migration._backup_created = True
```
snejus: What is this attribute for? Seems like it's needed because this logic runs before **every** pending migration.

We should instead adjust `Database._migrate` where we can check whether we have **any** new migrations - and thus create the backup just once.

beets/dbcore/db.py
```diff
@@ -1354,6 +1372,17 @@ def _migrate(self) -> None:
                     model_cls, self.db_tables[model_cls._table]["columns"]
                 )
 
+    def create_backup(self, dest: str | Path) -> None:
+        """Create a backup of the database at `dest`."""
+        if not os.path.isfile(self.path):
```
snejus: What is this check for? Beets will error early if the db does not exist and won't reach migrations logic?

beets/dbcore/db.py
```diff
@@ -1354,6 +1372,17 @@ def _migrate(self) -> None:
                     model_cls, self.db_tables[model_cls._table]["columns"]
                 )
 
+    def create_backup(self, dest: str | Path) -> None:
```
snejus: Why `Path`? You're only ever sending a `str`.

beets/dbcore/db.py
```diff
@@ -1354,6 +1372,17 @@ def _migrate(self) -> None:
...
+        """Create a backup of the database at `dest`."""
+        if not os.path.isfile(self.path):
+            raise FileNotFoundError(os.fspath(self.path))
+        # Use the SQLite backup API so the copy is consistent even when the
+        # database is open and may have a journal/WAL.
+        dest_conn = sqlite3.connect(os.fsdecode(dest))
```
snejus: Doesn't seem like `dest` can be `bytes`.
```suggestion
        dest_conn = sqlite3.connect(dest)
```

beets/dbcore/db.py
snejus: Or did you intend to check `dest`?

beets/dbcore/db.py
```diff
@@ -1045,6 +1047,22 @@ def migrate_model(self, model_cls: type[Model], *args, **kwargs) -> None:
         """Run this migration once for a model's backing table."""
         table = model_cls._table
         if not self.db.migration_exists(self.name, table):
+            if not Migration._backup_created and beets.config[
+                "create_backup_before_migrations"
+            ].get(bool):
+                dest = os.fsdecode(self.db.path) + ".bak"
```
snejus: Maybe we should name it something like `...-before-migration.bak`

beets/dbcore/db.py
```diff
@@ -1045,6 +1047,22 @@ def migrate_model(self, model_cls: type[Model], *args, **kwargs) -> None:
         """Run this migration once for a model's backing table."""
         table = model_cls._table
         if not self.db.migration_exists(self.name, table):
+            if not Migration._backup_created and beets.config[
```
snejus: Create a method `Migration.backup` and move the logic there. Let's keep `migrate_model` short and responsible for orchestration only.

test/library/test_migrations.py
```diff
@@ -284,3 +285,45 @@ def test_migrate(self, helper: TestHelper):
         str_item = helper.lib.get_item(2)
         assert str_item
         assert str_item.path == abs_bytes_path
+
+
+class TestMigrationBackup:
```
snejus: Would it be possible to move this test to any of existing test classes? Ideally we'd want to reuse the setup

snejus (changes requested):

beets/dbcore/db.py
semohr: Yeah we should be able to remove it. We can't have a in-memory database without a path, right?

beets/dbcore/db.py
semohr: I wanted to omit adding another database query and reuse existing ones. I had it in `Database` initially but this requires us to do the `is there any migration pending` check twice. Once for the backup and than once for each migration.

beets/dbcore/db.py
snejus: Ah, tests.

semohr: Now thinking about it, the tests should never trigger migrations tho (other than the migration test ofc).

beets/dbcore/db.py
snejus: Within `Database._migrate`, you can compare _existing_ vs _defined_ migrations: 

1. You can get _existing_ migrations through `Database.db_tables[<db_table>]["migrations"]`. 
   - `<db-table>` is `Model._table`: `"items"`/`"albums"`
   - each migration name corresponds to `Migration.name`
2. You can get _defined_ / _to migrate_ migrations from `Database._migrations`.

Something like this:

```py
    def _migrate(self) -> None:
        """Perform any necessary migration for the database."""
        existing = {
            (t, m) for t, v in self.db_tables.items() for m in v["migrations"]
        }
        defined = {
            (m._table, mig.name)
            for mig, models in self._migrations
            for m in models
        }
        new_migrations = bool(defined - existing)
        for migration_cls, model_classes in self._migrations:
            migration = migration_cls(self)
            for model_cls in model_classes:
                migration.migrate_model(
                    model_cls, self.db_tables[model_cls._table]["columns"]
                )
```

semohr: Maybe we just create an `has_pending_migration` method?

test/library/test_migrations.py
semohr: I would wait on #6709 to takle this.

semohr: Just updated the branch

beets/dbcore/db.py
```diff
@@ -1045,9 +1045,23 @@ def migrate_model(self, model_cls: type[Model], *args, **kwargs) -> None:
         """Run this migration once for a model's backing table."""
         table = model_cls._table
         if not self.db.migration_exists(self.name, table):
+            self._before_migration_backup(table)
             self._migrate_data(model_cls, *args, **kwargs)
             self.db.record_migration(self.name, table)
 
+    def _before_migration_backup(self, table: str):
```
snejus:
```suggestion
    def _before_migration_backup(self, table: str) -> None:
```

snejus (approved):
Looks good, just one small comment

PR #6744, opened 2026-06-14
tidal: Add flexible attributes and tidalsync command for popularity data

arsaboo:
This PR adds six typed flexible attributes to the Tidal plugin that are populated during album/track imports, and introduces a new `beet tidalsync` command to refresh popularity data post-import.

cc: @semohr 

### Flexible Attributes (`item_types`)
- `tidal_track_id`, `tidal_album_id`, `tidal_artist_id` — INTEGER fields for Tidal entity IDs.
- `tidal_track_popularity`, `tidal_alb_popularity` — INTEGER (0–100) normalized popularity scores.
- `tidal_updated` — DATE timestamp of last sync.

### Import population
- `_get_album_info()` and `_get_track_info()` pass the new fields through to `AlbumInfo` / `TrackInfo`, which store them as flexattrs on the item.
- `_popularity()` helper normalizes float/int/None from the Tidal API.

### Reimport support
- All Tidal fields added to `REIMPORT_FRESH_FIELDS_ITEM` so they refresh on reimport.

### `beet tidalsync` command
- Iterates library items, looks up each track by `tidal_track_id`, and updates `tidal_track_popularity` + `tidal_updated`.
- `--force`/`-f` — re-fetch even if data already exists.
- `--write`/`-w` — write updated tags to media files.
- Accepts an optional query to filter items (e.g., `beet tidalsync artist:Miles`).

### Documentation & Tests
- Attribute reference table and command usage examples in
  `docs/plugins/tidal.rst`.
- Flexattr assertions in existing tests, new `TestPopularity` and
  `TestTidalsync` test classes.

### DIFF
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -257,9 +257,7 @@ class TidalPlugin(MetadataSourcePlugin):
 
             for isrc in isrcs:
                 if track := isrc_to_track.get(isrc):
-                    yield self._get_track_info(
-                        track, artist_by_id=artist_by_id
-                    )
+                    yield self._get_track_info(track, artist_by_id=artist_by_id)
                 else:
                     yield None
 
@@ -364,9 +362,7 @@ class TidalPlugin(MetadataSourcePlugin):
             tracks=track_infos,
             artist=", ".join(artist_names),
             artists=artist_names,
-            duration=self._duration_to_seconds(
-                album["attributes"]["duration"]
-            ),
+            duration=self._duration_to_seconds(album["attributes"]["duration"]),
             albumtype=album["attributes"]["albumType"],
             label=self._parse_label(album["attributes"]),
             year=date_parts[0] if date_parts else None,
@@ -397,9 +393,7 @@ class TidalPlugin(MetadataSourcePlugin):
             isrc=track["attributes"]["isrc"],
             artist=", ".join(artist_names),
             artists=artist_names,
-            duration=self._duration_to_seconds(
-                track["attributes"]["duration"]
-            ),
+            duration=self._duration_to_seconds(track["attributes"]["duration"]),
             label=self._parse_label(track["attributes"]),
             # Flexattrs
             tidal_track_id=int(track["id"]),
@@ -482,9 +476,7 @@ class TidalPlugin(MetadataSourcePlugin):
         return None
 
     @staticmethod
-    def _parse_popularity(
-        attributes: AlbumAttributes | TrackAttributes,
-    ) -> int:
+    def _parse_popularity(attributes: AlbumAttributes | TrackAttributes) -> int:
         return round(attributes["popularity"] * 100)
 
     def sync_item_popularity(
@@ -584,8 +576,7 @@ class TidalPlugin(MetadataSourcePlugin):
         tidalsync_cmd = ui.Subcommand(
             "tidalsync",
             help=(
-                "Synchronize Tidal popularity data for "
-                "library items and albums"
+                "Synchronize Tidal popularity data for library items and albums"
             ),
         )
         tidalsync_cmd.parser.add_option(
@@ -611,8 +602,7 @@ class TidalPlugin(MetadataSourcePlugin):
             dest="force",
             default=False,
             help=(
-                "re-fetch popularity even if already present"
-                " (default: False)"
+                "re-fetch popularity even if already present (default: False)"
             ),
         )
         tidalsync_cmd.parser.add_option(
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -43,7 +43,7 @@ class TidalPlugin(MetadataSourcePlugin):
         "tidal_album_id": types.INTEGER,
         "tidal_artist_id": types.INTEGER,
         "tidal_track_popularity": types.INTEGER,
-        "tidal_alb_popularity": types.INTEGER,
+        "tidal_album_popularity": types.INTEGER,
         "tidal_updated": types.DATE,
     }
 
@@ -368,7 +368,7 @@ class TidalPlugin(MetadataSourcePlugin):
             # Flexattrs
             tidal_album_id=int(album["id"]),
             tidal_artist_id=int(artist_ids[0]) if artist_ids else None,
-            tidal_alb_popularity=self._popularity(
+            tidal_album_popularity=self._parse_popularity(
                 album["attributes"]["popularity"]
             ),
             tidal_updated=time.time(),
@@ -397,7 +397,7 @@ class TidalPlugin(MetadataSourcePlugin):
             # Flexattrs
             tidal_track_id=int(track["id"]),
             tidal_artist_id=int(artist_ids[0]) if artist_ids else None,
-            tidal_track_popularity=self._popularity(
+            tidal_track_popularity=self._parse_popularity(
                 track["attributes"]["popularity"]
             ),
             tidal_updated=time.time(),
@@ -477,7 +477,7 @@ class TidalPlugin(MetadataSourcePlugin):
         return None
 
     @staticmethod
-    def _popularity(val: object) -> int | None:
+    def _parse_popularity(val: object) -> int | None:
         if val is None:
             return None
         if isinstance(val, int):
@@ -493,36 +493,42 @@ class TidalPlugin(MetadataSourcePlugin):
         items = list(items)
         self._log.info("Syncing popularity for {0} tracks", len(items))
 
+        id_to_items: dict[int, list[Item]] = {}
         for item in items:
             if not force and item.get("tidal_track_popularity") is not None:
                 continue
 
             tidal_track_id = item.get("tidal_track_id")
-            if tidal_track_id is None:
+            if tidal_track_id is not None:
+                id_to_items.setdefault(int(tidal_track_id), []).append(item)
+
+        if not id_to_items:
+            return
+
+        all_ids = [str(tid) for tid in id_to_items]
+        results = list(self.search_tracks_by_ids(tidal_ids=all_ids))
+
+        for tid_str, result in zip(all_ids, results):
+            tid = int(tid_str)
+            if result is None:
+                self._log.debug("Track {0} not found in Tidal", tid)
                 continue
 
-            results = list(
-                self.search_tracks_by_ids(tidal_ids=[str(tidal_track_id)])
-            )
-            if not results or results[0] is None:
-                self._log.debug("Track {0} not found in Tidal", tidal_track_id)
-                continue
+            if result.tidal_track_popularity is not None:
+                for item in id_to_items[tid]:
+                    item["tidal_track_popularity"] = (
+                        result.tidal_track_popularity
+                    )
+                    item["tidal_updated"] = time.time()
+                    item.store()
+                    if write:
+                        item.try_write()
 
-            track_info = results[0]
-            if track_info.tidal_track_popularity is not None:
-                item["tidal_track_popularity"] = (
-                    track_info.tidal_track_popularity
-                )
-            item["tidal_updated"] = time.time()
-            item.store()
-            if write:
-                item.try_write()
-
-            self._log.debug(
-                "Updated popularity for track {0}: {1}",
-                tidal_track_id,
-                track_info.tidal_track_popularity,
-            )
+                    self._log.debug(
+                        "Updated popularity for track {0}: {1}",
+                        tid,
+                        result.tidal_track_popularity,
+                    )
 
     def commands(self) -> list[ui.Subcommand]:
         tidal_cmd = ui.Subcommand(
@@ -535,19 +541,7 @@ class TidalPlugin(MetadataSourcePlugin):
             help="Authenticate and login to Tidal",
             default=False,
         )
-
-        def func(lib: Library, opts: optparse.Values, args: list[str]):
-            if opts.auth:
-                self.api.ui_authenticate_flow()
-            else:
-                tidal_cmd.print_help()
-
-        tidal_cmd.func = func
-
-        tidalsync_cmd = ui.Subcommand(
-            "tidalsync", help="sync Tidal popularity data for library items"
-        )
-        tidalsync_cmd.parser.add_option(
+        tidal_cmd.parser.add_option(
             "-f",
             "--force",
             action="store_true",
@@ -555,7 +549,7 @@ class TidalPlugin(MetadataSourcePlugin):
             default=False,
             help="re-fetch popularity even if already present",
         )
-        tidalsync_cmd.parser.add_option(
+        tidal_cmd.parser.add_option(
             "-w",
             "--write",
             action="store_true",
@@ -564,13 +558,22 @@ class TidalPlugin(MetadataSourcePlugin):
             help="write updated tags to media files",
         )
 
-        def sync_func(lib: Library, opts: optparse.Values, args: list[str]):
-            items = lib.items(args)
-            self.tidalsync(items, write=opts.write, force=opts.force)
+        def func(lib: Library, opts: optparse.Values, args: list[str]):
+            action = args[0] if args else None
+            if action == "auth" or opts.auth:
+                self.api.ui_authenticate_flow()
+            elif action == "sync":
+                query = "data_source:tidal"
+                if len(args) > 1:
+                    query = f"{query} {' '.join(args[1:])}"
+                items = lib.items(query)
+                self.tidalsync(items, write=opts.write, force=opts.force)
+            else:
+                tidal_cmd.print_help()
 
-        tidalsync_cmd.func = sync_func
+        tidal_cmd.func = func
 
-        return [tidal_cmd, tidalsync_cmd]
+        return [tidal_cmd]
 
 
 ISO_8601_RE = re.compile(
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -45,9 +45,14 @@ log = getLogger("beets.tidal")
 class TidalPlugin(MetadataSourcePlugin):
     item_types: ClassVar[dict[str, types.Type]] = {
         "tidal_track_id": types.STRING,
-        "tidal_album_id": types.STRING,
         "tidal_artist_id": types.STRING,
         "tidal_track_popularity": types.INTEGER,
+        "tidal_updated": types.DATE,
+    }
+
+    album_types: ClassVar[dict[str, types.Type]] = {
+        "tidal_album_id": types.STRING,
+        "tidal_artist_id": types.STRING,
         "tidal_album_popularity": types.INTEGER,
         "tidal_updated": types.DATE,
     }
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -524,9 +524,10 @@ class TidalPlugin(MetadataSourcePlugin):
                 all_album_ids,
                 self.search_albums_by_ids(tidal_ids=all_album_ids),
             ):
-                album_popularity[int(aid_str)] = (
-                    album_result and album_result.tidal_album_popularity
-                )
+                if album_result is not None:
+                    album_popularity[int(aid_str)] = (
+                        album_result.tidal_album_popularity
+                    )
 
         for item in items:
             updated = False
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -37,6 +37,8 @@ if TYPE_CHECKING:
         TrackAttributes,
     )
 
+    PopularityAttributes = AlbumAttributes | TrackAttributes
+
 
 log = getLogger("beets.tidal")
 
@@ -260,9 +262,10 @@ class TidalPlugin(MetadataSourcePlugin):
 
             for isrc in isrcs:
                 if track := isrc_to_track.get(isrc):
-                    yield self._get_track_info(
+                    track_info = self._get_track_info(
                         track, artist_by_id=artist_by_id
                     )
+                    yield track_info
                 else:
                     yield None
 
@@ -355,6 +358,7 @@ class TidalPlugin(MetadataSourcePlugin):
             album["relationships"]["artists"]["data"], artist_by_id
         )
         date_parts = self._parse_release_date(album["attributes"])
+        duration = self._duration_to_seconds(album["attributes"]["duration"])
         return AlbumInfo(
             # Identifier
             data_source=self.data_source,
@@ -367,9 +371,7 @@ class TidalPlugin(MetadataSourcePlugin):
             tracks=track_infos,
             artist=", ".join(artist_names),
             artists=artist_names,
-            duration=self._duration_to_seconds(
-                album["attributes"]["duration"]
-            ),
+            duration=duration,
             albumtype=album["attributes"]["albumType"],
             label=self._parse_label(album["attributes"]),
             year=date_parts[0] if date_parts else None,
@@ -388,6 +390,7 @@ class TidalPlugin(MetadataSourcePlugin):
         artist_names, artist_ids = self._parse_artists(
             track["relationships"]["artists"]["data"], artist_by_id
         )
+        duration = self._duration_to_seconds(track["attributes"]["duration"])
 
         return TrackInfo(
             # Identifier
@@ -400,9 +403,7 @@ class TidalPlugin(MetadataSourcePlugin):
             isrc=track["attributes"]["isrc"],
             artist=", ".join(artist_names),
             artists=artist_names,
-            duration=self._duration_to_seconds(
-                track["attributes"]["duration"]
-            ),
+            duration=duration,
             label=self._parse_label(track["attributes"]),
             # Flexattrs
             tidal_track_id=track["id"],
@@ -486,9 +487,7 @@ class TidalPlugin(MetadataSourcePlugin):
         return None
 
     @staticmethod
-    def _parse_popularity(
-        attributes: AlbumAttributes | TrackAttributes,
-    ) -> int:
+    def _parse_popularity(attributes: PopularityAttributes) -> int:
         return round(attributes["popularity"] * 100)
 
     def sync_item_popularity(
@@ -584,13 +583,10 @@ class TidalPlugin(MetadataSourcePlugin):
         # It would be nice to combine both auth and sync commands but
         # currently not really supported in beets. We might be able to
         # revist this if our subcommand parsing changes
-        tidalsync_cmd = ui.Subcommand(
-            "tidalsync",
-            help=(
-                "Synchronize Tidal popularity data for library "
-                "items and albums"
-            ),
+        sync_help = (
+            "Synchronize Tidal popularity data for library items and albums"
         )
+        tidalsync_cmd = ui.Subcommand("tidalsync", help=sync_help)
         tidalsync_cmd.parser.add_album_option()
         tidalsync_cmd.parser.add_option(
             "-f",
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -78,9 +78,7 @@ class TidalPlugin(MetadataSourcePlugin):
         )
 
     def _tokenfile(self) -> str:
-        """Return the configured path to the token file
-        in the app directory.
-        """
+        """Return the configured path to the token file in the app directory."""
         return self.config["tokenfile"].get(confuse.Filename(in_app_dir=True))
 
     def require_authentication(self):
@@ -267,10 +265,7 @@ class TidalPlugin(MetadataSourcePlugin):
 
             for isrc in isrcs:
                 if track := isrc_to_track.get(isrc):
-                    track_info = self._get_track_info(
-                        track, artist_by_id=artist_by_id
-                    )
-                    yield track_info
+                    yield self._get_track_info(track, artist_by_id=artist_by_id)
                 else:
                     yield None
 
@@ -350,6 +345,7 @@ class TidalPlugin(MetadataSourcePlugin):
         track_by_id: dict[str, TidalTrack],
         artist_by_id: dict[str, TidalArtist],
     ) -> AlbumInfo:
+
         track_infos: list[TrackInfo] = []
         for i, track_rel in enumerate(
             album["relationships"]["items"]["data"], start=1
@@ -363,7 +359,6 @@ class TidalPlugin(MetadataSourcePlugin):
             album["relationships"]["artists"]["data"], artist_by_id
         )
         date_parts = self._parse_release_date(album["attributes"])
-        duration = self._duration_to_seconds(album["attributes"]["duration"])
         return AlbumInfo(
             # Identifier
             data_source=self.data_source,
@@ -376,7 +371,7 @@ class TidalPlugin(MetadataSourcePlugin):
             tracks=track_infos,
             artist=", ".join(artist_names),
             artists=artist_names,
-            duration=duration,
+            duration=self._duration_to_seconds(album["attributes"]["duration"]),
             albumtype=album["attributes"]["albumType"],
             label=self._parse_label(album["attributes"]),
             year=date_parts[0] if date_parts else None,
@@ -395,7 +390,6 @@ class TidalPlugin(MetadataSourcePlugin):
         artist_names, artist_ids = self._parse_artists(
             track["relationships"]["artists"]["data"], artist_by_id
         )
-        duration = self._duration_to_seconds(track["attributes"]["duration"])
 
         return TrackInfo(
             # Identifier
@@ -408,7 +402,7 @@ class TidalPlugin(MetadataSourcePlugin):
             isrc=track["attributes"]["isrc"],
             artist=", ".join(artist_names),
             artists=artist_names,
-            duration=duration,
+            duration=self._duration_to_seconds(track["attributes"]["duration"]),
             label=self._parse_label(track["attributes"]),
             # Flexattrs
             tidal_track_id=track["id"],
@@ -424,8 +418,7 @@ class TidalPlugin(MetadataSourcePlugin):
     ) -> tuple[list[str], list[str]]:
         """Extract artists from a relationship.
 
-        Artists are sorted in the track/album response relationship but
-        not in the
+        Artists are sorted in the track/album response relationship but not in the
         track/album responses included items.
         """
         artist_names = []
@@ -444,8 +437,7 @@ class TidalPlugin(MetadataSourcePlugin):
     @staticmethod
     def _parse_title(attributes: AlbumAttributes | TrackAttributes):
         """
-        Tidal UIs append the version string at the end of the title.
-        We do the same here
+        Tidal UIs append the version string at the end of the title. We do the same here
         by formatting it as ``"{title} ({version})"`` to stay consistent.
         """
         if version := attributes.get("version"):
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -23,6 +23,7 @@ if TYPE_CHECKING:
     import optparse
     from collections.abc import Callable, Iterable, Sequence
 
+    from beets.autotag import Info
     from beets.dbcore.db import Results
     from beets.library import Library
     from beets.library.models import Album, Item
@@ -42,9 +43,9 @@ log = getLogger("beets.tidal")
 
 class TidalPlugin(MetadataSourcePlugin):
     item_types: ClassVar[dict[str, types.Type]] = {
-        "tidal_track_id": types.INTEGER,
-        "tidal_album_id": types.INTEGER,
-        "tidal_artist_id": types.INTEGER,
+        "tidal_track_id": types.STRING,
+        "tidal_album_id": types.STRING,
+        "tidal_artist_id": types.STRING,
         "tidal_track_popularity": types.INTEGER,
         "tidal_album_popularity": types.INTEGER,
         "tidal_updated": types.DATE,
@@ -70,7 +71,9 @@ class TidalPlugin(MetadataSourcePlugin):
         )
 
     def _tokenfile(self) -> str:
-        """Return the configured token file path in the app directory."""
+        """Return the configured path to the token file
+        in the app directory.
+        """
         return self.config["tokenfile"].get(confuse.Filename(in_app_dir=True))
 
     def require_authentication(self):
@@ -211,7 +214,7 @@ class TidalPlugin(MetadataSourcePlugin):
 
     @overload
     def search_tracks_by_ids(
-        self, *, tidal_ids: Iterable[str] | Iterable[int]
+        self, *, tidal_ids: Iterable[str]
     ) -> Iterable[TrackInfo | None]: ...
 
     @overload
@@ -222,10 +225,10 @@ class TidalPlugin(MetadataSourcePlugin):
     def search_tracks_by_ids(
         self,
         ids: Iterable[str] | None = None,
-        tidal_ids: Iterable[str] | Iterable[int] | None = None,
+        tidal_ids: Iterable[str] | None = None,
         isrcs: Iterable[str] | None = None,
     ) -> Iterable[TrackInfo | None]:
-        _ids: list[str | int | None] = list(tidal_ids or [])
+        _ids: list[str | None] = list(tidal_ids or [])
         isrcs = list(isrcs or [])
         if ids:
             _ids = list(map(self._extract_id, ids))
@@ -245,7 +248,7 @@ class TidalPlugin(MetadataSourcePlugin):
         }
 
         for _id in _ids:
-            if _id is not None and (track := track_by_id.get(str(_id))):
+            if _id is not None and (track := track_by_id.get(_id)):
                 yield self._get_track_info(track, artist_by_id=artist_by_id)
             else:
                 yield None
@@ -257,7 +260,9 @@ class TidalPlugin(MetadataSourcePlugin):
 
             for isrc in isrcs:
                 if track := isrc_to_track.get(isrc):
-                    yield self._get_track_info(track, artist_by_id=artist_by_id)
+                    yield self._get_track_info(
+                        track, artist_by_id=artist_by_id
+                    )
                 else:
                     yield None
 
@@ -268,7 +273,7 @@ class TidalPlugin(MetadataSourcePlugin):
 
     @overload
     def search_albums_by_ids(
-        self, *, tidal_ids: Iterable[str] | Iterable[int]
+        self, *, tidal_ids: Iterable[str]
     ) -> Iterable[AlbumInfo | None]: ...
 
     @overload
@@ -279,10 +284,10 @@ class TidalPlugin(MetadataSourcePlugin):
     def search_albums_by_ids(
         self,
         ids: Iterable[str] | None = None,
-        tidal_ids: Iterable[str] | Iterable[int] | None = None,
+        tidal_ids: Iterable[str] | None = None,
         barcode_ids: Iterable[str] | None = None,
     ) -> Iterable[AlbumInfo | None]:
-        _ids: list[str | int | None] = list(tidal_ids or [])
+        _ids: list[str | None] = list(tidal_ids or [])
         barcode_ids = list(barcode_ids or [])
         if ids:
             _ids = list(map(self._extract_id, ids))
@@ -309,7 +314,7 @@ class TidalPlugin(MetadataSourcePlugin):
         }
 
         for _id in _ids:
-            if _id is not None and (album := album_by_id.get(str(_id))):
+            if _id is not None and (album := album_by_id.get(_id)):
                 yield self._get_album_info(
                     album, track_by_id=track_by_id, artist_by_id=artist_by_id
                 )
@@ -362,15 +367,17 @@ class TidalPlugin(MetadataSourcePlugin):
             tracks=track_infos,
             artist=", ".join(artist_names),
             artists=artist_names,
-            duration=self._duration_to_seconds(album["attributes"]["duration"]),
+            duration=self._duration_to_seconds(
+                album["attributes"]["duration"]
+            ),
             albumtype=album["attributes"]["albumType"],
             label=self._parse_label(album["attributes"]),
             year=date_parts[0] if date_parts else None,
             month=date_parts[1] if date_parts else None,
             day=date_parts[2] if date_parts else None,
             # Flexattrs
-            tidal_album_id=int(album["id"]),
-            tidal_artist_id=int(artist_ids[0]) if artist_ids else None,
+            tidal_album_id=album["id"],
+            tidal_artist_id=artist_ids[0] if artist_ids else None,
             tidal_album_popularity=self._parse_popularity(album["attributes"]),
             tidal_updated=time.time(),
         )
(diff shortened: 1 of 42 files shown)


beetsplug/tidal/__init__.py
```diff
@@ -474,6 +476,102 @@ def _parse_release_date(
             return int(parts[0]), int(parts[1]), int(parts[2])
         return None
 
+    @staticmethod
+    def _popularity(val: object) -> int | None:
```
semohr: Lets name this `_parse_popularity` for consistency.

beetsplug/tidal/__init__.py
```diff
@@ -474,6 +476,102 @@ def _parse_release_date(
...
+                item.try_write()
+
+            self._log.debug(
+                "Updated popularity for track {0}: {1}",
+                tidal_track_id,
+                track_info.tidal_track_popularity,
+            )
+
+    def commands(self) -> list[ui.Subcommand]:
+        tidal_cmd = ui.Subcommand(
```
semohr, line 561: Would be nice to have `beet tidal [auth|sync]`.

docs/plugins/tidal.rst
```diff
@@ -102,4 +102,76 @@ Default
...
+    - - ``tidal_album_id``
+      - INTEGER
+      - Tidal album ID
+    - - ``tidal_artist_id``
+      - INTEGER
+      - Tidal artist ID
+    - - ``tidal_track_popularity``
+      - INTEGER
+      - Track popularity score (0-100)
+    - - ``tidal_alb_popularity``
```
semohr: Can we rename this to `tidal_album_popularity` or was this done for compatibility with `arsaboo/beets-tidal`?

beetsplug/tidal/__init__.py
```diff
@@ -474,6 +476,102 @@ def _parse_release_date(
...
+            "-w",
+            "--write",
+            action="store_true",
+            dest="write",
+            default=False,
+            help="write updated tags to media files",
+        )
+
+        def sync_func(lib: Library, opts: optparse.Values, args: list[str]):
+            items = lib.items(args)
```
semohr: I think we can be a bit smarter here. We can query only items with `tidal` as `data_source`. Also if `force` is not given we can even omit all items with `popularity` set.

beetsplug/tidal/__init__.py
```diff
@@ -474,6 +476,102 @@ def _parse_release_date(
...
+        for item in items:
+            if not force and item.get("tidal_track_popularity") is not None:
+                continue
+
+            tidal_track_id = item.get("tidal_track_id")
+            if tidal_track_id is None:
+                continue
+
+            results = list(
+                self.search_tracks_by_ids(tidal_ids=[str(tidal_track_id)])
```
semohr: It would make sense to batch the lookup imo. The most expensive thing in the popularity lookup pipeline is the network requests, not only does tidal limit our requests but also network requests are slow compared to everything else.

The `TidalApi.get_tracks` function has batching build in so we could just aggregate all ids and than pass them to the search function in one iteration. This should speedup the lookup by up to a factor 20 (max ids in request by tidal) for large syncs 🙃

beetsplug/tidal/__init__.py
```diff
@@ -399,6 +394,13 @@ def _get_track_info(
             artists=artist_names,
             duration=self._duration_to_seconds(track["attributes"]["duration"]),
             label=self._parse_label(track["attributes"]),
+            # Flexattrs
+            tidal_track_id=int(track["id"]),
```
semohr: It looks like we're now storing the same track ID twice in the database: `track_id` and `tidal_track_id`.
Summoning @snejus ;) What's our preferred approach here? Should we consolidate these?

**same for `artist_id`

semohr (changes requested):

beetsplug/tidal/__init__.py
arsaboo: I am not sure we can consolidate.  All the other plugins (bandcamp, spotify, etc. store bandcamp_album_id, spotify_album_id as flexattrs). Until we can make musicbrainz completely optional, I feel we keep them for consistency with the plugin ecosystem pattern (other metadata plugins store their own *_id flexattrs), but we should get @snejus's input.

Also makes queries like `tidal_track_id` unambiguous without needing to know which standard field maps to which source.

beetsplug/tidal/__init__.py
arsaboo: Done

beetsplug/tidal/__init__.py
```diff
@@ -474,6 +476,105 @@ def _parse_release_date(
             return int(parts[0]), int(parts[1]), int(parts[2])
         return None
 
+    @staticmethod
+    def _parse_popularity(val: object) -> int | None:
```
semohr: Can we add typing to the params here? I would do the same `attributes: AlbumAttributes | TrackAttributes` as for the other parse functions.

semohr:
Sorry about the iterative comments. Im coming back to this whenever I have time or thought about something ;)

I just noticed that albums on tidal also have `popularity` values. I would vote we make popularity work for both tracks and albums (currently it only works for tracks, right?). Maybe a general `tidal_popularity` flex attribute could work? 

---

I'm also able to make some changes if you feel like it gets too much give me a headsup :fearful:

arsaboo:
> Sorry about the iterative comments. Im coming back to this whenever I have time or thought about something ;)
> 
> I just noticed that albums on tidal also have `popularity` values. I would vote we make popularity work for both tracks and albums (currently it only works for tracks, right?). Maybe a general `tidal_popularity` flex attribute could work?
> 
> I'm also able to make some changes if you feel like it gets too much give me a headsup 😨

We have `tidal_album_popularity` to be explicit and avoid any confusion. Updated the parsing based on your comments. 

Please don't worry about it. Happy to work with you on this and get this merged. Thanks for all your guidance, and by all means, feel free to improve upon it.

beetsplug/tidal/__init__.py
```diff
@@ -477,7 +477,8 @@ def _parse_release_date(
         return None
 
     @staticmethod
-    def _parse_popularity(val: object) -> int | None:
+    def _parse_popularity(attributes: AlbumAttributes | TrackAttributes) -> int | None:
```
semohr: We dont have to be defensive here. According to the docs popularity is a required field. Marked by the red star. see https://tidal-music.github.io/tidal-api-reference or the json directly

Image

arsaboo: I agree and am happy to make it required, but I don't think there is any downside here. If the value is present, it will be fetched.

semohr:
> We have tidal_album_popularity to be explicit and avoid any confusion. Updated the parsing based on your comments.

Im not entirely sure how the best UX should look like here tbh. Both variants (with and without prefix) come with tradeoffs :thinking:

beetsplug/tidal/__init__.py
semohr: The downside is more on the code side. We don't want beets to ship dead code branches. Should also make it a bit easier to maintain and reason about the code.

From my perspective this  should be sufficent:
```python
@staticmethod
def _parse_popularity(attributes: AlbumAttributes | TrackAttributes) -> int:
    return round(attributes["popularity"] * 100)
```

arsaboo:
> > We have tidal_album_popularity to be explicit and avoid any confusion. Updated the parsing based on your comments.
> 
> Im not entirely sure how the best UX should look like here tbh. Both variants (with and without prefix) come with tradeoffs 🤔

We are not currently exposing popularity information in UI. This is mostly for the DB so that we can use it for other smart things (playlists)

semohr:
> We are not currently exposing popularity information in UI. This is mostly for the DB so that we can use it for other smart things (playlists)

I mean the query syntax with UX. The question is whether users should have separate fields per entity, such as:
```bash
beet ls tidal_album_popularity:5..20
```
or a single generic field whose meaning depends on the search scope, for example:
```
beet ls tidal_popularity:5..50 -a
```
The latter is shorter and more flexible, but the former is more explicit and consistent across entity types.

arsaboo:
> > We are not currently exposing popularity information in UI. This is mostly for the DB so that we can use it for other smart things (playlists)
> 
> I mean the query syntax with UX. The question is whether users should have separate fields per entity, such as:
> 
> ```shell
> beet ls tidal_album_popularity:5..20
> ```
> 
> or a single generic field whose meaning depends on the search scope, for example:
> 
> ```
> beet ls tidal_popularity:5..50 -a
> ```
> 
> The latter is shorter and more flexible, but the former is more explicit and consistent across entity types.

I would (strongly) prefer the former, just to be explicit.

beetsplug/tidal/__init__.py
arsaboo: Done

semohr:
@arsaboo Just enhanced the sync logic a bit. Make sure to pull ;) Please have a look! Am mostly happy with the changes but ofcourse if you see something or have a better idea, lets discuss.

arsaboo:
@semohr Excellent. Updated the other parts of the code. Everything is green.

semohr:
Neat I will have another iteration at it tomorrow. Mainly tests and docs as I have not checked these in detail yet but than we should be good to go :+1:

beetsplug/tidal/__init__.py
```diff
@@ -487,7 +495,7 @@ def sync_item_popularity(
             results=results,
             write=write,
             force=force,
-            id_field="mb_trackid",
+            id_field="tidal_track_id",
```
semohr, line 496: I used `mb_trackid` intentionally. If we want to keep it as `tidal_track_id`, we'll need a way to migrate the existing `mb_trackid` values into the `tidal_track_id` flex field for users who were using the plugin before flex fields were introduced.

Is there a command that copies `mb_trackid` values into the `tidal_track_id` flex field? If so, we can mention it in the changelog, and I'm happy with that approach. 🙂

arsaboo: `mb_trackid` is a MusicBrainz UUID, while `tidal_track_id` is an integer (type mismatch). The Tidal API cannot be called with a MusicBrainz ID. So using mb_trackid as the id_field would have other unintended consequences. I don't think that is the right path.  Users can re-import or create a custom script. 

One option is to try `tidal_track_id` first and, if missing, fall back to `mb_trackid` (I don't think this is a good idea, but I will defer to your judgment).

semohr: That's not how it currently works, unfortunately. The `mb_trackid` field stores the ID corresponding to the `data_source`, this is what I was referring to as duplication in my other comment. For example, if `data_source == "tidal"`, then `mb_trackid` will always contain the TIDAL track ID.

What I had in mind instead was a simple beets command that migrates all `mb_*id` values for models where `data_source == "tidal"` into the new flex fields. Otherwise, users who have previously used the TIDAL plugin won't be able to use sync after these changes unless they manually fix their libraries (me included :/)

beetsplug/tidal/__init__.py
snejus: I see two separate topics here:

### 1. `<source>_album_id` set by `musicbrainz`

`musicbrainz` sets `<source>_album_id` flex attr - this is reliable information, and it's a good idea to record those. **However, we aren't making any use of them currently**.

A potential use case could be something like:

- Extend autotagging candidates with a result for each of the URLs (using the corresponding plugin, if enabled).

### 2. Everything else

I only see problems with setting source-specific `<source>_{artist,label,track,album}_id` etc. fields:

1. Lack of consistency:

```sh
$ grep -ahoR 'discogs[a-z_]*id=' beets*
discogs_albumid=
discogs_labelid=
discogs_artistid=
$ grep -ahoR 'deezer[a-z_]*id=' beets*
deezer_album_id=
deezer_track_id=
$ grep -ahoR 'spotify[a-z_]*id=' beets*
spotify_album_id=
spotify_artist_id=
spotify_track_id=
spotify_artist_id=
```

2. They can stay around even after `data_source` changes. That makes the fields look authoritative when they may just be stale leftovers from a previous tag source.

3. Our current source of truth is already confusing: `mb_albumid`, `mb_trackid`, and `mb_artistid` are the fields that drive a lot of beets behavior, but in practice they can contain source-specific IDs, not only MusicBrainz IDs. Adding more provider-specific ID fields without clarifying that model risks making this ambiguity worse.

4. They become part of the user-visible data model, but without a clear contract around naming, ownership, or semantics.

5. They are easy to write and hard to make useful. If no code reads them, we are mostly creating permanent metadata surface area that has to be documented, migrated, tested, and kept consistent across plugins.

So I think recording a source ID is fine when it has a concrete consumer or when the plugin is preserving the exact source used for a match. But I would avoid adding more provider-specific ID fields as a general pattern until we have a shared abstraction and an actual feature that uses them.

semohr: > musicbrainz sets <source>_album_id flex attr - this is reliable information, and it's a good idea to record those. However, we aren't making any use of them currently.

We use the same pattern here and it looks like we actually make use of it here: We can get the popularity values (from tidal) via these ids. Even for albums/tracks that are initially tagged via musicbrainz if they include the `tidal_album_id`.

semohr: For me this sounds like we should keep the flex fields for now until we have designed a better solution.

beetsplug/tidal/__init__.py
arsaboo: Unfortunately, I don't think there's any existing beet command that can conditionally copy one field to another. `beet modify` only accepts literal values, not dynamic field references.

If re-import is acceptable, it should be a one-time process. During reimport, beets/importer/match.py calls `tracks_for_ids([item.mb_trackid])` and `albums_for_ids([item.mb_albumid])`, which pass those IDs to all configured metadata providers concurrently. Each plugin tries to parse the ID with its own regex pattern. So you can trivially reimport by enabling only the Tidal provider, and the Tidal ID stored in mb_* fields would be used by the Tidal plugin during reimport. 

Here's a standalone migration script that should work (untested):

```python
#!/usr/bin/env python3
"""One-time migration: copy mb_*id fields to tidal_*_id flex attrs.

Before the Tidal plugin had its own tidal_track_id, tidal_album_id, and
tidal_artist_id flexible attributes, it stored IDs in the standard
mb_trackid, mb_albumid, mb_artistid fields (via MEDIA_FIELD_MAP in
beets.autotag.hooks). This script copies those values into the new
plugin-specific flex attrs so that `beet tidalsync` can find them.

Run once after upgrading to the new Tidal plugin version:

    python migrate_tidal_ids.py

Only items/albums where data_source == "Tidal" are touched.
"""

from beets import config, logging
from beets.library import Library

log = logging.getLogger("beets")


def migrate_items(lib: Library) -> int:
    count = 0
    for item in lib.items("data_source:Tidal"):
        dirty = False

        if item.get("tidal_track_id") is None and item.get("mb_trackid"):
            item["tidal_track_id"] = int(item["mb_trackid"])
            dirty = True

        if item.get("tidal_artist_id") is None and item.get("mb_artistid"):
            item["tidal_artist_id"] = int(item["mb_artistid"])
            dirty = True

        if dirty:
            item.store()
            log.info(
                "Migrated item {0}: {1} - {2}",
                item.id, item.artist, item.title,
            )
            count += 1
    return count


def migrate_albums(lib: Library) -> int:
    count = 0
    for album in lib.albums("data_source:Tidal"):
        dirty = False

        if album.get("tidal_album_id") is None and album.get("mb_albumid"):
            album["tidal_album_id"] = int(album["mb_albumid"])
            dirty = True

        if album.get("tidal_artist_id") is None and album.get("mb_albumartistid"):
            album["tidal_artist_id"] = int(album["mb_albumartistid"])
            dirty = True

        if dirty:
            album.store()
            log.info(
                "Migrated album {0}: {1} - {2}",
                album.id, album.albumartist, album.album,
            )
            count += 1
    return count


def main() -> None:
    config.clear()
    config.read()
    lib = Library(config["library"].as_filename())

    item_count = migrate_items(lib)
    album_count = migrate_albums(lib)

    log.info(
        "Migration complete: {0} items and {1} albums updated",
        item_count, album_count,
    )


if __name__ == "__main__":
    main()
```

If you really want, we can add a `--migrate/-m` flag to `beet tidalsync` that does this one-time copy.

semohr: Would not include the migration flag. This will be dead code quite quickly 😆 

Maybe just put instructions on how to re-import into the changelog, seems acceptable to me. Should be pretty quick anyways.

arsaboo: Done

beetsplug/tidal/__init__.py
arsaboo: I agree, until we can reliably make MB optional, flex fields are the way to go.

snejus: I'm not suggesting we get rid of all of them here. 🙂 

It seems that populating `tidal_album_id` allows fetching track popularity _regardless_ of the current `data_source` value, for example I see two cases where this is relevant:

1. Album is tagged with data from MusicBrainz, and may include `tidal_album_id` value
2. Album is tagged using `tidal`, populating `tidal_album_id`. Even if this album is re-tagged using another data source, `tidal_album_id` persists. Question: is there a risk of `tidal_album_id` pointing to an old/incorrect album? This value doesn't get cleared as far as I know. 

---
Regarding `tidal_artist_id` and `tidal_track_id` - why are we adding them if we don't use them?

arsaboo: tidal_ids are updated whenever a user (re)imports using Tidal. This is the same for any other tag provider. If the album is updated by another tag provider, only the respective IDs should be updated. Thus, tidal_ids will still be valid. In fact, we want to retain IDs from multiple providers so we can retrieve details from all configured providers. 

tidal_track_id is used for popularity information, which is one of the key things that we are after.

semohr: Maybe we get rid of `tidal_artist_id` as it is completely unused. Am also wondering about what value the `tidal_updated` has currently 🤔 

Since we use the `tidal_track_id` we should not get rid of it. It also mirrors the album logic so it seems pretty consistent to me.

docs/changelog.rst
```diff
@@ -27,6 +27,19 @@ New features
   :conf:`plugins.musicbrainz:aliases_as_credits` to make
   aliases-as-artist-credit optional.
 - :doc:`plugins/badfiles`: Added settings for auto error and warning actions.
+- :doc:`plugins/tidal`: New ``item_types`` flexible attributes
+  ``tidal_track_id``, ``tidal_album_id``, ``tidal_artist_id``,
```
semohr: `tidal_album_id` should be a `album` level flex attribute while `tidal_track_id` should be an item level flex attribute, right?

Same for `popularity`.

beetsplug/tidal/__init__.py
arsaboo: `tidal_updated` is used to query and update popularity information for the last X days (e..g, update popularity for tracks that have not been updated since the last 60 days). For example, this is how I use `spotifysync` right now:

beet -v spotifysync -f 'spotify_updated:..-59d'

Honestly, artist_id from other providers is also unused right now, but we retain it as it can be queried. I am fine removing it if you all strongly prefer.

semohr: I'm happy to keep both, since it pretty much mirrors how the spotify plugin currently works. Thanks for your input, both of you!

docs/changelog.rst
semohr: Was just for my understanding :laughing: No need to update the changelog. I would make it smaller anyways if possible. Information like that can go into the docs imo.
Same for the specifics of the sync command, i.e. the args and usage.

beetsplug/tidal/__init__.py
snejus: Yeah, agree - for now let's just try to keep this consistent... With `spotify` only, I guess? 🤣 

> ```sh
> $ grep -ahoR 'spotify[a-z_]*id=' beets*
> spotify_album_id=
> spotify_artist_id=
> spotify_track_id=
> spotify_artist_id=
> ```

arsaboo: You can then blame me for both 🤣🙈

beetsplug/tidal/__init__.py
```diff
@@ -219,7 +211,7 @@ def search_tracks_by_ids(
 
     @overload
     def search_tracks_by_ids(
-        self, *, tidal_ids: Iterable[str]
+        self, *, tidal_ids: Iterable[str] | Iterable[int]
```
snejus: I don't think we should make these IDs type more flexible.

Instead of relaxing types in all these downstream functions, I think you should just ensure that you're submitting `str`ings instead of `int`s:

```diff
diff --git a/beetsplug/tidal/__init__.py b/beetsplug/tidal/__init__.py
index e58d8ec4f..6365bec24 100644
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -489,7 +489,7 @@ def sync_item_popularity
             force=force,
             id_field="tidal_track_id",
             popularity_field="tidal_track_popularity",
-            search=lambda ids: self.search_tracks_by_ids(tidal_ids=ids),
+            search=lambda ids: self.search_tracks_by_ids(tidal_ids=map(str, ids)),
             label="item",
         )
 ```

File: beetsplug/tidal/__init__.py [hunk 12]
```diff
@@ -377,6 +368,11 @@ def _get_album_info(
             year=date_parts[0] if date_parts else None,
             month=date_parts[1] if date_parts else None,
             day=date_parts[2] if date_parts else None,
+            # Flexattrs
+            tidal_album_id=int(album["id"]),
```
snejus: Tidal API returns `id: str` in all documents. Why storing them as `int`?

beetsplug/tidal/__init__.py
```diff
@@ -408,8 +409,8 @@ def _parse_artists(
     ) -> tuple[list[str], list[str]]:
         """Extract artists from a relationship.
 
-        Artists are sorted in the track/album response relationship but not in the
```
snejus, line 411: Undo this change

beetsplug/tidal/__init__.py
```diff
@@ -427,8 +428,8 @@ def _parse_artists(
     @staticmethod
     def _parse_title(attributes: AlbumAttributes | TrackAttributes):
         """
-        Tidal UIs append the version string at the end of the title. We do the same here
```
snejus, line 430: Undo this one as well

beetsplug/tidal/__init__.py
```diff
@@ -56,7 +70,7 @@ def api(self) -> TidalAPI:
         )
 
     def _tokenfile(self) -> str:
-        """Return the configured path to the token file in the app directory."""
+        """Return the configured token file path in the app directory."""
```
snejus: Consider updating your `AGENTS.md` with something like:

```
- Ignore pre-existing lint issues
- Only fix lint issues that are caused by your changes
```

beetsplug/tidal/__init__.py
snejus: > I don't think there's any existing beet command that can conditionally copy one field to another. only accepts literal values, not dynamic field references.

It **does** accept dynamic templates:

```go
$ beet modify data_source:deezer deezer_album_id='$mb_albumid' first_genre='%first{$genres}'
Modifying 17 items.
16719 | 0000 / Bootlegs: Alek Száhala - The Last of the Mohicans
  deezer_album_id: 
  first_genre: freeform
54920 | 2022 / V.A SILENTSØULS 002: Parapher - Alambra
  deezer_album_id: 370280807
  first_genre: acid techno
55649 | 2005 / Take It: Tom Novy - Take It (Radio Edit)
  deezer_album_id: 382439727
  first_genre: ambient
55650 | 1999 / The Gift Of Game: Crazy Town - Butterfly
  deezer_album_id: 1206154
  first_genre: rock
55664 | 2003 / No Way Out: Diddy - I'll Be Missing You (feat. Faith Evans & 112)
  deezer_album_id: 356922
  first_genre: rap
56938 | 2004 / Dancefloor Deluxe: Alcazar - Crying at the Discoteque (Radio Edit)
  deezer_album_id: 76719
  first_genre: Pop
56943 | 2004 / Aš Noriu: Džordana Butkutė - Nebenoriu Laukt
  deezer_album_id: 182066982
  first_genre: Pop
56951 | 2021 / Miss California: Dante Thomas - Miss California (Radio Version)
  deezer_album_id: 233946292
  first_genre: Pop
57656 | 2011 / Electroman: Benny Benassi - Spaceship (feat. Kelis, apl.de.ap & Jean Baptiste) (Radio Edit)
  deezer_album_id: 6792491
  first_genre: House
58229 | 2007 / 90's Pop Pre-Cleared Comp: 4 Non Blondes - What's Up?
  deezer_album_id: 120393
  first_genre: Pop
58251 | 1999 / Californication (Deluxe Edition): Red Hot Chili Peppers - Otherside
  deezer_album_id: 86511
  first_genre: Alternative
58252 | 2016 / CDW: Charlotte Day Wilson - Work
  deezer_album_id: 188498622
  first_genre: R&B
58271 | 2010 / Nose EP: Nose - Ubetinga Kjærlighet
  deezer_album_id: 706708
  first_genre: Alternative
58274 | 2005 / Rock & Roll Queen (US DMD single): The Subways - Rock & Roll Queen
  deezer_album_id: 365020
  first_genre: Alternative/Rock/Indie Rock/Rock Pop
58377 | 2004 / Pumphonia: Benassi Bros. - Illusion (Sfaction Version)
  deezer_album_id:  -> 63469192
  first_genre: electro house
58378 | 2004 / Pumphonia: Benassi Bros. - Hit My Heart (Sfaction Version)
  deezer_album_id:  -> 63469192
  first_genre: electro house
58509 | 2026 / Juda Tavo Rankos: Jovani - Juda Tavo Rankos
  deezer_album_id: 993177061
  first_genre: Dance
Really modify and move? (Yes/no/select) ^C
```

beetsplug/tidal/__init__.py
```diff
@@ -36,6 +41,15 @@
 
 
 class TidalPlugin(MetadataSourcePlugin):
+    item_types: ClassVar[dict[str, types.Type]] = {
+        "tidal_track_id": types.INTEGER,
+        "tidal_album_id": types.INTEGER,
+        "tidal_artist_id": types.INTEGER,
```
snejus: These should be strings for compatibility with
1. `mb_{track,album,artist}id` types
2. types returned by Tidal API

```suggestion
        "tidal_track_id": types.STRING,
        "tidal_album_id": types.STRING,
        "tidal_artist_id": types.STRING,
```

beetsplug/tidal/__init__.py
```diff
@@ -36,6 +41,15 @@
...
+        "tidal_track_popularity": types.INTEGER,
+        "tidal_album_popularity": types.INTEGER,
+        "tidal_updated": types.DATE,
```
snejus, line 55: I saw your discussion:

> I mean the query syntax with UX. The question is whether users should have separate fields per entity, such as:
> ```bash
> beet ls tidal_album_popularity:5..20
> ```
> or a single generic field whose meaning depends on the search scope, for example:
> ```
> beet ls tidal_popularity:5..50 -a
> ```
> The latter is shorter and more flexible, but the former is more explicit and consistent across entity types.
 
I'd prefer the latter option, because:

### 1. User-friendliness
this is easier to use:
```sh
$ beet list -f '$tidal_popularity'
$ beet list -f '$tidal_popularity' -a
```
compared to having to adjust field name depending on whether I provide `-a` flag or not
```sh
$ beet list -f '$tidal_track_popularity'
$ beet list -f '$tidal_album_popularity' -a
```

### 2. Simpler implementation

Using a fixed `tidal_popularity` field is easier to understand than following the `popularity_field` value.

### 3. Redundancy

Seeing keys like 
- `tidal_track_popularity` under `item_attributes`
- `tidal_album_popularity` under `album_attributes`

Either `album` or `track` in these key names are redundant.

beetsplug/tidal/__init__.py
```diff
@@ -474,6 +475,166 @@ def _parse_release_date(
...
+        popularity_field: str,
+        search: Callable[
+            [Iterable[int]], Iterable[TrackInfo | AlbumInfo | None]
+        ],
+        label: Literal["item", "album"],
+    ) -> None:
+        """Sync Tidal popularity for a generic model (Item or Album)."""
+        log.info("Syncing popularity for {0} {1}s", len(results), label)
+
+        to_sync: defaultdict[int, list[Item | Album]] = defaultdict(list)
```
snejus: Why do we expect to have **multiple** models under the same tidal ID?

I expected to see `model_by_tidal_id: dict[str, Item | Album]`.

beetsplug/tidal/__init__.py
```diff
@@ -474,6 +475,166 @@ def _parse_release_date(
...
+    def _sync_popularity(
+        self,
+        *,
+        results: Results[Item] | Results[Album],
+        write: bool,
+        force: bool,
+        id_field: str,
+        popularity_field: str,
+        search: Callable[
+            [Iterable[int]], Iterable[TrackInfo | AlbumInfo | None]
```
snejus:
```suggestion
            [Iterable[int]], Iterable[Info | None]
```

beetsplug/tidal/__init__.py
```diff
@@ -474,6 +475,166 @@ def _parse_release_date(
...
+            if tidal_id := model.get(id_field):
+                if force or model.get(popularity_field) is None:
+                    to_sync[tidal_id].append(model)
+        total = sum(len(v) for v in to_sync.values())
+        log.debug("{0} {1}s need updates", total, label)
+        processed = 0
+        for models, new_info in zip(to_sync.values(), search(to_sync.keys())):
+            for model in models:
+                model[popularity_field] = (
+                    getattr(new_info, popularity_field) if new_info else None
```
snejus:
```suggestion
                    new_info.get(popularity_field) if new_info else None
```

snejus: Should we act on `new_info=None` and set `tidal_updated` though?

beetsplug/tidal/__init__.py
```diff
@@ -474,6 +475,166 @@ def _parse_release_date(
...
+            help="update albums (default: True)",
+        )
+        tidalsync_cmd.parser.add_option(
+            "-i",
+            "--item",
+            action="store_true",
+            dest="items",
+            default=True,
+            help="update items/singletons (default: True)",
+        )
```
snejus: I don't think we need this. Every other command acts on `Item` by default, and switches to `Album` whenever `-a`/`--album` is given.
```suggestion
```

snejus (changes requested):

beetsplug/tidal/__init__.py
semohr: I wanted to support running on both items and albums with only one command, which is not possible with the current paradigm, right?

beetsplug/tidal/__init__.py
semohr: Yeah that is an awkward edgecase. In theory it is well possible to have duplicates in your database. Im not sure if there is a usecase for it but I dont want to assume usage patterns if it is easy enough for us to just allow here.

beetsplug/tidal/__init__.py
semohr: I dont really mind either :shrug:
@arsaboo

beetsplug/tidal/__init__.py
semohr: I'm not entirely sure how flex fields are stored.

If they're stored as strings/bytes anyway, then I'm fine with converting them here. However, if the underlying data is stored `INTEGER`.

Using strings introduces unnecessary overhead:
* Additional encoding/decoding work.
* Larger storage requirements.
* Bigger indexes

I'd rather update the type hint to reflect reality than introduce conversions that don't provide any functional benefit.

beetsplug/tidal/__init__.py
semohr: Ints were always supported; the type hint just didn't reflect that. At least at the API layer, I would argue that accepting `int` values is reasonable, as the additional flexibility can be useful.

If I remember correctly, the tidal api itself returns integer ids. We only convert them to strings because the `album_id` and `track_id` fields are defined as strings.

That said, if we decide to replace the `INTEGER` flex fields with `STRING`  altogether as discussed in https://github.com/beetbox/beets/pull/6744#discussion_r3424481061, then I agree this change shouldn't be part of this PR.

semohr: Update: Nvm, seem like the tidal api indeed returns strings even tho the ids are always ints. Im fully behind @snejus than. We should not convert the ids to integers.

beetsplug/tidal/__init__.py
```diff
@@ -474,6 +475,166 @@ def _parse_release_date(
...
+            default=False,
+            help="write updated tags to media files (default: False)",
+        )
+        tidalsync_cmd.parser.set_usage(
+            "Usage: beet tidalsync <query> [options]"
+        )
+
+        def sync_func(lib: Library, opts: optparse.Values, args: list[str]):
+            query = "data_source:tidal"
+            if len(args) > 1:
```
snejus: And final question: why are we ignoring the first arg here? I added two print statements here:

```diff
diff --git a/beetsplug/tidal/__init__.py b/beetsplug/tidal/__init__.py
index f1575b35c..84288d0c1 100644
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -625,4 +625,6 @@ def sync_func
             query = "data_source:tidal"
+            print(f"{args=}")
             if len(args) > 1:
                 query = f"{query} {' '.join(args[1:])}"
+            print(f"{query=}")
```

and see:

```sh
$ beet tidalsync path::aaa
args=['path::aaa']
query='data_source:tidal'
tidal: Syncing popularity for 0 items
tidal: Successfully synchronised popularity for 0 items
tidal: Syncing popularity for 0 albums
tidal: Successfully synchronised popularity for 0 albums
$ beet tidalsync path::aaa path::bbb
args=['path::aaa', 'path::bbb']
query='data_source:tidal path::bbb'
tidal: Syncing popularity for 0 items
tidal: Successfully synchronised popularity for 0 items
tidal: Syncing popularity for 0 albums
tidal: Successfully synchronised popularity for 0 albums
```

beetsplug/tidal/__init__.py
snejus: I think `api_types.py` document these, no? I assumed they aren't lying to us 😆 

```py
class TidalAlbum(TypedDict):
    id: str
    type: Literal["albums"]
    attributes: AlbumAttributes
    relationships: dict[str, RelationshipData]


class TidalTrack(TypedDict):
    id: str
    type: Literal["tracks"]
    attributes: TrackAttributes
    relationships: dict[str, RelationshipData]
```

beetsplug/tidal/__init__.py
snejus: They are stored as `TEXT`, see `Database._make_attribute_table`.

semohr: Than we have no reason to use `INTEGER` :+1:

snejus: Even if for some reason they were stored as any other type we should store them as strings for the two reasons I mentioned above.

beetsplug/tidal/__init__.py
arsaboo: A couple of reasons I’d prefer to keep the explicit tidal_track_popularity / tidal_album_popularity split in this PR:

1. Not all providers expose album popularity, so a generic tidal_popularity would end up meaning “track popularity” for items and “album popularity” for albums, depending on context. I think using explicit names makes that distinction clearer and avoids overloading the field semantics.
2. This is also closer to the existing plugin conventions. For example, the Spotify plugin uses the explicit spotify_track_popularity field rather than a generic overloaded spotify_popularity.

That said, if you feel strongly about switching to a generic tidal_popularity, I’m happy to make that change. My only hesitation is that it may be better handled as a broader consistency pass across similar plugins, rather than introducing a one-off naming rule here.

beetsplug/tidal/__init__.py
snejus: It is _theoretically_ possible, but extremely unlikely. I think we'd rather simplify it here and assume we have unique IDs. In the worst case scenario, if we have some edge case with duplicate IDs, the search handles it fine:

```py
>>> [a.id for a in pl.search_albums_by_ids(['444595305', '444595305'])]
['444595305', '444595305']
```

It does become a bit simpler:

```diff
diff --git a/beetsplug/tidal/__init__.py b/beetsplug/tidal/__init__.py
index f1575b35c..84288d0c1 100644
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -529,24 +529,26 @@ def _sync_popularity
 
-        to_sync: defaultdict[int, list[Item | Album]] = defaultdict(list)
-        for model in results:
-            if tidal_id := model.get(id_field):
-                if force or model.get(popularity_field) is None:
-                    to_sync[tidal_id].append(model)
-        total = sum(len(v) for v in to_sync.values())
+        model_by_id = {
+            tidal_id: m
+            for m in results
+            if (tidal_id := m.get(id_field))
+            and (force or m.get(popularity_field) is None)
+        }
+        total = len(model_by_id)
         log.debug("{0} {1}s need updates", total, label)
         processed = 0
-        for models, new_info in zip(to_sync.values(), search(to_sync.keys())):
-            for model in models:
-                model[popularity_field] = (
-                    getattr(new_info, popularity_field) if new_info else None
-                )
-                model["tidal_updated"] = time.time()
+        for model, new_info in zip(
+            model_by_id.values(), search(model_by_id.keys())
+        ):
+            model[popularity_field] = (
+                new_info.get(popularity_field) if new_info else None
+            )
+            model["tidal_updated"] = time.time()
 
-                model.store()
-                if write:
-                    model.try_write()
+            model.store()
+            if write:
+                model.try_write()
 
-                processed += 1
-                if processed % 100 == 0 or processed == total:
-                    log.debug("Synced {0}/{1} {2}s", processed, total, label)
+            processed += 1
+            if processed % 100 == 0 or processed == total:
+                log.debug("Synced {0}/{1} {2}s", processed, total, label)
```

beetsplug/tidal/__init__.py
snejus: Then let's remove both of these options and always handle both items and albums.

beetsplug/tidal/__init__.py
semohr: It depends on the use case, imo. Storing ids as integers has a significant advantage because sql indexes are much smaller and generally perform better. Queries on string columns carry considerably more overhead, both in terms of storage and comparison costs.

Hypothetically, if the ids were stored in their own database column, I would strongly recommend using an integer type. However, you're absolutely right that it doesn't make sense in this case, since the type we've defined here is only a converter and doesn't affect how the value is actually stored.

arsaboo: Changed to STRING

beetsplug/tidal/__init__.py
snejus: @arsaboo agree with you - also just remembered how `Item.get` defaults to album fields when it doesn't have that key, so let's keep them explicit.

beetsplug/tidal/__init__.py
```diff
@@ -31,11 +36,22 @@
...
+    PopularityAttributes = AlbumAttributes | TrackAttributes
+
 
 log = getLogger("beets.tidal")
 
 
 class TidalPlugin(MetadataSourcePlugin):
+    item_types: ClassVar[dict[str, types.Type]] = {
+        "tidal_track_id": types.STRING,
+        "tidal_album_id": types.STRING,
```
semohr: We need to move the tidal_album_id and popularity into the `album_types`, right?

arsaboo: Good point. While I make this change.....do we need tidal_updated at both item and album levels?

beetsplug/tidal/__init__.py
arsaboo: Fixed — switched to query = ["data_source:tidal", *args] so no arguments are dropped

arsaboo:
@semohr and @snejus, aside from the popularity issue, everything else should be addressed. I have kept both album and track popularity for now  (but can update it if it is a blocker).

beetsplug/tidal/__init__.py
```diff
@@ -359,6 +362,7 @@ def _get_album_info(
             album["relationships"]["artists"]["data"], artist_by_id
         )
         date_parts = self._parse_release_date(album["attributes"])
+        duration = self._duration_to_seconds(album["attributes"]["duration"])
```
semohr: Lets keep this inline as it was before.

beetsplug/tidal/__init__.py
```diff
@@ -385,6 +394,7 @@ def _get_track_info(
         artist_names, artist_ids = self._parse_artists(
             track["relationships"]["artists"]["data"], artist_by_id
         )
+        duration = self._duration_to_seconds(track["attributes"]["duration"])
```
semohr: Same here lets keep it inline

beetsplug/tidal/__init__.py
```diff
@@ -56,7 +77,9 @@ def api(self) -> TidalAPI:
         )
 
     def _tokenfile(self) -> str:
-        """Return the configured path to the token file in the app directory."""
+        """Return the configured path to the token file
```
semohr: Unrelated change.

beetsplug/tidal/__init__.py
```diff
@@ -265,7 +266,10 @@ def search_tracks_by_ids(
 
             for isrc in isrcs:
                 if track := isrc_to_track.get(isrc):
-                    yield self._get_track_info(track, artist_by_id=artist_by_id)
+                    track_info = self._get_track_info(
```
semohr: Unrelated change

beetsplug/tidal/__init__.py
```diff
@@ -345,7 +349,6 @@ def _get_album_info(
         track_by_id: dict[str, TidalTrack],
         artist_by_id: dict[str, TidalArtist],
     ) -> AlbumInfo:
-
```
semohr, line 348: Unrelated change :)

docs/changelog.rst
```diff
@@ -27,6 +27,18 @@ New features
...
   aliases-as-artist-credit optional.
 - :doc:`plugins/badfiles`: Added settings for auto error and warning actions.
+- :doc:`plugins/tidal`: New flexible attributes are now populated during
+  imports, including ``tidal_track_id``, ``tidal_album_id``,
+  ``tidal_artist_id``, ``tidal_track_popularity``, ``tidal_album_popularity``,
+  and ``tidal_updated``. Added a new ``beet tidalsync`` command to refresh
+  popularity data for imported items by default, or albums with ``--album``,
+  with ``--force`` to re-fetch and ``--write`` to update file tags.
+
+  **Migration**: Existing items imported with the Tidal plugin before these
```
semohr: Lets incorporate the comment: https://github.com/beetbox/beets/pull/6744/changes#r3424462146

semohr (changes requested):
Lets get rid of the unrelated changes please 😆

beetsplug/tidal/__init__.py
arsaboo: The build fails if we don't split this line due to the line-length issue.

beetsplug/tidal/__init__.py
arsaboo: The build fails if we don't split this line due to the line-length issue.

beetsplug/tidal/__init__.py
arsaboo: The build fails if we don't split this line due to the line-length issue.

beetsplug/tidal/__init__.py
arsaboo: The build fails if we don't split this line due to the line-length issue.

beetsplug/tidal/__init__.py
arsaboo: The build fails if we don't split this line due to the line-length issue.

beetsplug/tidal/__init__.py
arsaboo: The build fails if we don't split this line due to the line-length issue.

docs/changelog.rst
arsaboo: I don't see any comment on that link.

arsaboo:
> Lets get rid of the unrelated changes please 😆

Most of the changes are to prevent the build from failing. Notice they are material changes, just reorganization to ensure that line length is less than 80.

beetsplug/tidal/__init__.py
```diff
@@ -31,11 +36,27 @@
         TrackAttributes,
     )
 
+    PopularityAttributes = AlbumAttributes | TrackAttributes
```
snejus: You should create `Attributes` typed dict which has shared keys between album and track, like `title`, `duration`, `popularity` etc. Then inherit from it in `AlbumAttributes` and `TrackAttributes`.

Then use `Attributes` in these typing annotations.

beetsplug/tidal/__init__.py
snejus: Can you undo them all and see whether it fails?

snejus: It would have failed when it was first committed - but didn't.

beetsplug/tidal/api.py
```diff
@@ -72,7 +72,7 @@ def search_results(
 
     def get_tracks(
         self,
-        ids: list[str] | None = None,
+        ids: list[str | int] | None = None,
```
snejus: Neither this nor the adjustment in `get_albums` is needed anymore.

docs/changelog.rst
semohr: This should work for the migration
 ```
beet modify data_source:tidal tidal_album_id='$mb_albumid' -a
beet modify data_source:tidal tidal_track_id='$mb_trackid'
```

File: beetsplug/tidal/__init__.py (hunk 26 above)
Comment (semohr) [reply beetsplug/tidal/__init__.py]: Image
White line marks the line length

```
Poe => ruff check --config=pyproject.toml
All checks passed!
Poe => ruff format --config=pyproject.toml
302 files left unchanged
```

Comment (semohr) [reply beetsplug/tidal/__init__.py]: Maybe your local config got mixed up?

File: beetsplug/tidal/__init__.py (hunk 30 above)
Comment (arsaboo) [reply beetsplug/tidal/__init__.py]: Done

File: beetsplug/tidal/api.py (hunk 31 above)
Comment (arsaboo) [reply beetsplug/tidal/api.py]: Done

File: beetsplug/tidal/__init__.py (hunk 27 above)
Comment (semohr) [reply beetsplug/tidal/__init__.py]: It worked before and it works on my machine :fearful:

File: beetsplug/tidal/api_types.py [hunk 32]
```diff
@@ -42,38 +42,39 @@ class ArtistAttributes(TypedDict):
     spotlighted: NotRequired[bool]
 
 
-class AlbumAttributes(TypedDict):
```
semohr, line 45: Im not too sure about these changes @snejus. You just suggested them, right?

I know it simplifies things but the types here are intentionally mirroring the json:api types from tidal. See [here](https://tidal-music.github.io/tidal-api-reference/tidal-api-oas.json). I would not simplify this if just for easier maintenance down the line.

snejus: That's exactly what I wanted - I want to be able to see attributes shared between albums and tracks.

This is similar to what we have in `beetsplug._utils.musicbrainz.RelationshipBase`.

snejus:
@arsaboo are you going to undo unrelated changes or shall I commit that instead?

beetsplug/tidal/__init__.py
```diff
@@ -18,24 +20,44 @@
...
         AlbumAttributes,
+        Attributes,
         ResourceIdentifier,
         TidalAlbum,
         TidalArtist,
         TidalTrack,
         TrackAttributes,
     )
 
+    PopularityAttributes = Attributes
```
snejus: Remove this and use `Attributes` directly.

beetsplug/tidal/api_types.py
semohr: Hmm, this seems like a maintenance issue to me. Once TIDAL changes its schema, it would be more difficult than necessary to refactor our own types because they would no longer align with the JSON schema.

For mb, there aren't any strong schemas, so it makes sense for us to define a smaller, normalized contract. But here we're essentially mirroring the types tidal gives us. I deliberately chose **not** to minify them because keeping our types aligned with tidal's reduces the cognitive and maintenance overhead, and makes future changes much easier to track.

arsaboo:
> @arsaboo are you going to undo unrelated changes or shall I commit that instead?

It will be great if you can do it. I am stepping out for a few and will complete the remaining issues.

semohr (approved):

snejus (approved):

semohr:
Thanks for the discussions and progress here!

arsaboo:
Thanks, @semohr and @snejus, for patiently working on this and all the discussions.

PR #6745, opened 2026-06-15
lyrics: add rest_directory configuration option

### LINKED ISSUE #2806: Feature Request: add an option to configure default directory to write lyrics to for lyrics plugin
The title says it all. I think it could be great to have a `lyrics.default_directory` entry in the configuration file for a default directory in which the lyrics file will be written, instead of always specifying it manually with `-r` or `--write-rest`. Usually I run this command not from a specific directory so this could be useful.

hakkilab:
## Description

Fixes #2806 

- Adds a `rest_directory` configuration option to the lyrics plugin that specifies a directory for ReST output, equivalent to the `-r, --write-rest` command line argument

- [x] Documentation.
- [x] Changelog.
- [x] Tests.

### DIFF
--- a/test/plugins/test_lyrics.py
+++ b/test/plugins/test_lyrics.py
@@ -857,7 +857,6 @@ class TestLyricsRestDirectory(PluginTestHelper):
             ),
             pytest.param("test/config", None, "test/config", id="config only"),
             pytest.param(None, "test/cmd", "test/cmd", id="cmd arg only"),
-            pytest.param(None, None, None, id="no output"),
             pytest.param(
                 "~/test/config", None, "~/test/config", id="user home path"
             ),
@@ -883,11 +882,7 @@ class TestLyricsRestDirectory(PluginTestHelper):
         cmd_args = [] if arg_path is None else ["-r", arg_path]
         self.run_command("lyrics", *cmd_args, lib=lib)
 
-        test_output = test_capture.get("directory")
-        if output_path is None:
-            assert test_output is None
-        else:
-            assert test_output == Path(output_path).expanduser()
+        assert test_capture.get("directory") == Path(output_path).expanduser()
 
 
 class TestLyricsSyltProperty:
--- a/test/plugins/test_lyrics.py
+++ b/test/plugins/test_lyrics.py
@@ -20,23 +20,21 @@ import re
 import textwrap
 from functools import partial
 from http import HTTPStatus
+from pathlib import Path
 from types import SimpleNamespace
 from typing import TYPE_CHECKING
 
 import pytest
 import requests
-from confuse.exceptions import ConfigTypeError
 
 from beets.library import Item
-from beets.test.helper import PluginMixin
+from beets.test.helper import PluginMixin, PluginTestHelper
 from beets.util.lyrics import Lyrics
 from beetsplug import lyrics
 
 from .lyrics_pages import lyrics_pages
 
 if TYPE_CHECKING:
-    from pathlib import Path
-
     from .lyrics_pages import LyricsPage
 
 PHRASE_BY_TITLE = {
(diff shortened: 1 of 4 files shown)


hakkilab:
Hi @snejus,
This is my first contribution to this project. I fixed issue #2806 by adding a `rest_directory` config option to the lyrics plugin. I also added tests and documentation updates for the feature. Would appreciate a review when you have time! Let me know what fixes are needed or if there are any steps needed before review.

test/plugins/test_lyrics.py
```diff
@@ -843,6 +844,125 @@ def test_write(self, rest_dir: Path, rest_files):
...
+        "bad_config",
+        [
+            pytest.param(42),
+            pytest.param(3.14),
+            pytest.param(False),
+            pytest.param({}),
+            pytest.param([]),
+        ],
+    )
+    def test_bad_config(self, lyrics_plugin, bad_config):
```
snejus: There's no need to test `confuse` internals :)

test/plugins/test_lyrics.py
```diff
@@ -843,6 +844,125 @@ def test_write(self, rest_dir: Path, rest_files):
...
+    @pytest.fixture
+    def backend_name(self):
+        return "lrclib"
+
+    def _setup_config(self, tmp_path, rest_directory):
+        config_file = tmp_path / "config.yaml"
+        with config_file.open("w") as f:
+            f.write(f"lyrics:\n  rest_directory: {rest_directory}")
+        self.config.set_file(config_file)
+        self.config.read(False, False)
```
snejus: You should use `self.config` to adjust configuration.

test/plugins/test_lyrics.py
```diff
@@ -843,6 +844,125 @@ def test_write(self, rest_dir: Path, rest_files):
...
+        opts, args = cmd.parser.parse_args(cmd_args)
+        cmd.func(lib, opts, args)
+
+        if output_path is None:
+            for item in tmp_path.rglob("*"):
+                assert item.name != "index.rst"
+                assert item.name != "conf.py"
+        else:
+            assert (output_path / "index.rst").exists()
+            assert (output_path / "conf.py").exists()
```
snejus: And use `self.run_command` to run commands!

snejus (changes requested):
I think these tests are slightly too extensive. Your tests do something similar to `TestRestFiles` except that you override the configuration and use `self.run_command` to run lyrics command.

hakkilab:
Hi @snejus,

Thanks for the review! I have refactored the tests as follows:
- Removed `test_bad_config` tests
- Use `self.config` to set configuration instead of writing/reading config files
- Use `run_command` instead of `cmd.func`
- Added a mock `MockRestFiles` so no output is generated and only directory configuration logic is tested

When you have a chance please take a look at my changes and let me know if any other changes are needed. Thanks!

test/plugins/test_lyrics.py
hakkilab: I have resolved this by removing this test

test/plugins/test_lyrics.py
hakkilab: I have modified the tests to use `self.config` for setting configurations

test/plugins/test_lyrics.py
hakkilab: I have refactored to use `run_command` instead of `cmd.func`

test/plugins/test_lyrics.py
```diff
@@ -843,6 +842,91 @@ def test_write(self, rest_dir: Path, rest_files):
...
+                None,
+                "~/test/config",
+                id="config only, home path",
+            ),
+            pytest.param(
+                None, "~/test/cmd", "~/test/cmd", id="cmd arg only, home path"
+            ),
+            pytest.param(None, None, None, id="no output"),
+        ],
+    )
```
snejus, line 864: Since we want to test just command line + config handling, there's no need to test home/absolute directories. I'd say remove test cases that test absolute directories and simplify them focusing on `config_path` and `arg_path` behaviour.

hakkilab: @snejus Thanks for the quick review!

I added tests for relative, absolute, and home paths because I saw behavior where a user home path in the config file was not handled correctly until I changed `RestFiles(Path(opts.rest_directory)).write(items)` to `RestFiles(Path(opts.rest_directory).expanduser()).write(items)`. 

Originally, setting something like `~/test` in the config would cause files to output to `./~/test` instead of something like `/home/user/test`. This behavior was not replicated when using the command line argument. I think this was due to user home expansion being done by the shell when using the command line argument, which doesn't occur when reading from the config, so this wasn't explicitly handled previously. This made me think checking path variations in the tests could be useful.

Given this, what would be preferred?

1) Remove all the path variations and simplify it down to the four combinations of with/without config file/command line arg
2) Remove the path variations but include one test for user home expansion to prevent a possible bug if the path handling is refactored in the future
3) Keep it as is

I will implement whatever is desired. Thanks!

snejus: Thanks for expanding on the reasoning here! I'd say go ahead with Option 2!

hakkilab: Great! I've done so and pushed a new commit.

hakkilab:
@snejus Thanks for reviewing again!

Based on our discussion I've simplified the tests further by removing the path variations. The remaining tests cover the following cases:
- Setting an output path in the config file and in the command line
- Setting an output path in the config file only
- Setting an output path in the command line only
- Not setting an output path
- Using a user home path in the config file

Please review once more and let me know if any other changes are needed

test/plugins/test_lyrics.py
```diff
@@ -843,6 +842,54 @@ def test_write(self, rest_dir: Path, rest_files):
...
+
+    @pytest.mark.parametrize(
+        "config_path, arg_path, output_path",
+        [
+            pytest.param(
+                "test/config", "test/cmd", "test/cmd", id="config and cmd arg"
+            ),
+            pytest.param("test/config", None, "test/config", id="config only"),
+            pytest.param(None, "test/cmd", "test/cmd", id="cmd arg only"),
+            pytest.param(None, None, None, id="no output"),
```
snejus: Ah I forgot to mention to remove this test case as well - it's fine to not have it. It's going to remove some conditional logic in the test!

hakkilab: I've removed this test and simplified the assert logic by removing the if/else

hakkilab:
Hi @snejus,
Thanks for the review! I've removed the "no output" test case as recommended and simplified the assert logic by removing the no longer necessary if/else. Please give my changes another review and let me know if there's anything else needed.

snejus (approved):
Thanks!

PR #6753, opened 2026-06-19
Feature: waveform image plugin

Chris-repo:
## Description

Fixes #6752.  

As discussed in #6752, Please feel free to critique 

## To Do

### DIFF
--- a/beetsplug/spectrogram.py
+++ b/beetsplug/spectrogram.py
@@ -1,0 +1,109 @@
+# This file is part of beets.
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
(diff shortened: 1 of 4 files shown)


semohr:
Hi, and thanks for the PR.

In the discussion, I asked about the motivation and intended use cases for this plugin, but I don't think we've fully explored that yet.

As it stands, I'm struggling to understand why this functionality should be introduced as a dedicated plugin in beets. The proposed workflow appears to be something that can already be implemented using the existing [hooks](https://docs.beets.io/en/latest/plugins/hook.html) plugin and external tooling.

Could you elaborate on the points raised in the discussion?

Understanding the reasoning and intended workflow would help us evaluate whether this is something that belongs in beets itself, or whether it is better served through existing extension mechanisms.

---

Separately, regardless of the outcome of that discussion, the implementation will need to conform to our coding and contribution standards. Please take a look at the contribution guide here: https://docs.beets.io/en/latest/contributing.html.

Chris-repo:
Hi, 

I was unaware of the hooks plugin, my initial thoughts was to generate waveform images for audio scrubbing and extending the web plugin but if there is existing ways of generating them without further code this pull request is kind of redundant.

Thanks again for your prompt responses, keep up the good work.

PR #6755, opened 2026-06-19
Remove logging import sideeffect

semohr:
This PR refactors configuration and logging initialization in the beets UI layer.

Logging is no longer implicitly configured as a side effect of import; it is now initialized explicitly during UI startup. This allows library users to fully control logging configuration themselves, and enables more granular logging setup based on user configuration (future PR).

It also clarifies startup order by deferring heavier initialization until after logging is available for proper error reporting.

---

### DIFF
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -905,7 +905,7 @@ def _raw_main(args: list[str] | None) -> None:
 
 
 def _bootstrap_config(options: optparse.Values) -> confuse.ConfigError | None:
-    """Apply CLI to config, initialize logging, return error as value if any."""
+    """Apply CLI to config, return error as value if any."""
 
     deferred_error: confuse.ConfigError | None = None
     try:
@@ -932,7 +932,7 @@ def _bootstrap_config(options: optparse.Values) -> confuse.ConfigError | None:
     return deferred_error
 
 
-def _bootstrap_logging():
+def _bootstrap_logging() -> None:
     if not log.handlers:
         handler = logging.StreamHandler()
         handler.setFormatter(
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -882,6 +882,14 @@ def _raw_main(args: list[str] | None) -> None:
 
         return config_edit(options)
 
+    if "AppData\\Local\\Microsoft\\WindowsApps" in sys.exec_prefix:
+        raise UserError(
+            "beets is unable to use the Microsoft Store version of "
+            "Python. Please install Python from https://python.org.\n"
+            "More details can be found here "
+            "https://beets.readthedocs.io/en/stable/guides/main.html"
+        )
+
     if deferred_error:
         raise deferred_error
 
@@ -949,14 +957,6 @@ def main(args: list[str] | None = None) -> None:
     """Run the main command-line interface for beets. Includes top-level
     exception handlers that print friendly error messages.
     """
-    if "AppData\\Local\\Microsoft\\WindowsApps" in sys.exec_prefix:
-        log.error(
-            "error: beets is unable to use the Microsoft Store version of "
-            "Python. Please install Python from https://python.org.\n"
-            "error: More details can be found here "
-            "https://beets.readthedocs.io/en/stable/guides/main.html"
-        )
-        sys.exit(1)
     try:
         _raw_main(args)
     except UserError as exc:
--- a/beets/ui/__init__.py
+++ b/beets/ui/__init__.py
@@ -42,13 +42,6 @@ if sys.platform == "win32":
 
 
 log = logging.getLogger("beets")
-if not log.handlers:
-    handler = logging.StreamHandler()
-    handler.setFormatter(
-        logging.LegacyFormatter("%(legacy_prefix)s%(message)s")
-    )
-    log.addHandler(handler)
-log.propagate = False  # Don't propagate to root handler.
 
 
 # Encoding utilities.
@@ -753,14 +746,11 @@ optparse.Option.ALWAYS_TYPED_ACTIONS += ("callback",)
 # The main entry point and bootstrapping.
 
 
-def _setup(
-    options: optparse.Values,
-) -> tuple[list[Subcommand], library.Library]:
+def _setup() -> tuple[list[Subcommand], library.Library]:
     """Prepare and global state and updates it with command line options.
 
     Returns a list of subcommands, a list of plugins, and a library instance.
     """
-    config = _configure(options)
 
     plugins.load_plugins()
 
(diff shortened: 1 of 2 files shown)


beets/ui/__init__.py
```diff
@@ -952,18 +919,55 @@ def parse_csl_callback(
...
 
 
+def _bootstrap_config(options: optparse.Values) -> ConfigError | None:
+    """Apply CLI to config, initialize logging, return error as value if any."""
+
+    # Apply config overlay (deferred error to allow setting up logging)
+    deferred_error: ConfigError | None = None
+    try:
+        # For some reason we need to materialize before adding another config
+        # file otherwise the sources order is inverted...
```
semohr: I think this was a bug before but noone noticed thus far. We might want to investigate this more and maybe create a confuse issue.

I noticed this when I printed all configuration sources (`config.sources`) without calling config.read().

beets/ui/__init__.py
```diff
@@ -952,18 +918,59 @@ def parse_csl_callback(
...
+    return deferred_error
+
+
+def _bootstrap_logging():
+    if not log.handlers:
+        handler = logging.StreamHandler()
+        handler.setFormatter(
+            logging.LegacyFormatter("%(legacy_prefix)s%(message)s")
+        )
+        log.addHandler(handler)
```
snejus, line 941: It seems to me that this line may be missing? 

```py
log.propagate = False  # Don't propagate to root handler.
```

beets/ui/__init__.py
```diff
@@ -952,18 +918,59 @@ def parse_csl_callback(
...
+        if overlay_path := getattr(options, "config", None):
+            config.set_file(overlay_path)
+    except confuse.ConfigError as e:
+        deferred_error = e
+        # Ensure defaults are loaded even when user config is broken,
+        # so that logging and other subsystems can read basic settings.
+        config.read(user=False, defaults=True)
+
+    # Even if the earlier config loading fails we seperatly try to set the config
+    # values the cli options
```
snejus:
```suggestion
    # Even if the earlier config loading fails we seperately try to set the
    # cli options
```

beets/ui/__init__.py
snejus: > I printed all configuration sources (`config.sources`) without calling config.read()

```py
>>> from beets import config

>>> config.sources
[]
```

Could you describe in a bit more detail?

beets/ui/__init__.py
```diff
@@ -952,18 +918,59 @@ def parse_csl_callback(
...
+        log.set_global_level(logging.DEBUG)
+    else:
+        log.set_global_level(logging.INFO)
+
+    # List configuration sources for user convenience.
+    log.debug("configuration sources (highest → lowest priority):")
+    for source in config.sources[1:]:
+        log.debug(
+            "{} {}", type(source).__name__, getattr(source, "filename", "")
+        )
```
snejus: I think you were meant to replace this logic with `config.log_sources(log)` call.

beets/ui/__init__.py
snejus, line 941: Maybe it could be worth moving this to a method on `BeetsLogger`?

snejus (commented):
Lovely cleanup!

beets/ui/__init__.py
semohr: The order of configuration sources is important when you want a configuration file to override existing settings. In beets, this is typically done via the `--config` argument, which adds a file to confuse as source.

While creating the `log_sources` method, I noticed that when an overlay config is provided via `--config`, it is not inserted with the highest priority if `config.read()` has not been called beforehand. Instead, it ends up not in the list of sources, meaning it is ignored 😆.

My guess is that this is an artifact of how configuration sources are materialized in Confuse. Calling read before inserting via `set_file` seems to fix the issue.

This boils down to:

```python
>>> from beets import config
>>> config.set_file("./config.yaml")
>>> config.sources
[]
>>> config.read()
>>> config.sources
[ConfigSource(<super: <class 'ConfigSource'>, <YamlSource object>>, '/home/sebastian/.config/beets/config.yaml', False, False), ConfigSource(<super: <class 'ConfigSource'>, <YamlSource object>>, '/mnt/Repositories/beets/beets/config_default.yaml', True, False)]
## MISSING ./config ##
>>> config.set_file("./config.yaml")
>>> config.sources
[ConfigSource(<super: <class 'ConfigSource'>, <YamlSource object>>, './config.yaml', False, False), ConfigSource(<super: <class 'ConfigSource'>, <YamlSource object>>, '/home/sebastian/.config/beets/config.yaml', False, False), ConfigSource(<super: <class 'ConfigSource'>, <YamlSource object>>, '/mnt/Repositories/beets/beets/config_default.yaml', True, False)]
## Includes ./config ##

File: beets/ui/__init__.py (hunk 2 above)
Comment (semohr) [reply beets/ui/__init__.py:941]: Do you know what is the use for beets? We never register another handler so there realistically is never a propagation we need to disable. 

I added it back but I doubt we need it.

Comment (semohr) [reply beets/ui/__init__.py:941]: Do you mean the full bootstrap function?

I'd lean against that. `BeetsLogger` can be used independently of the UI layer in the "beets as a library" use case, so this would effectively be dead code in that scenario. It also introduces a dependency from the logger implementation back to the UI bootstrap path, which feels like the wrong direction architecturally.

Comment (semohr) [reply beets/ui/__init__.py:941]: Seems like this breaks the tests. Makes sense as `caplog` requires `propagation = True`. I would vote we just remove it as it is not needed. If you want to keep it we need to create a workaround for the tests.

Comment (snejus) [reply beets/ui/__init__.py:941]: Yeah, we don't need it then.

File: beets/ui/__init__.py (hunk 1 above)
Comment (snejus) [reply beets/ui/__init__.py]: It seems like it works fine the way it is anyways? I just copied my config and kept

```yaml
library: hello.db
```

```sh
$ beet --config config.yaml ls
```

It gives me nothing, so the overidden configuration is picked up.

semohr: Possible it is only the `sources` list which is not updated correctly :thinking: Should we still keep the read call? Makes sense from my perspective that the `_bootstrap_config` function initializes the config fully. 

I think this was not done beforehand (only implicitly via the first `config["verbose"]` call).

snejus: I think it's fine keep it to be done implicitly.

semohr: I would opt for the explicit `read` calls tbh:
- Makes it clear when and how we expect the errors to be handled: the `config.read` can trigger errors if a broken file is given e.g. in the user folder. This error will be otherwise raised on the first `config["verbose"]` call if we opt for implicit handling. 
- The logging of used configuration sources is easier and more generic that way (atleast until we fix/enhance confuse)

I think the `_bootstrap_config` function should make subsequent calls to the config safe if possible.

snejus: Agree with your reasoning - just update the comment accordingly and remove the fixme :)

beets/ui/__init__.py
```diff
@@ -952,18 +918,53 @@ def parse_csl_callback(
     return None
 
 
+def _bootstrap_config(options: optparse.Values) -> confuse.ConfigError | None:
+    """Apply CLI to config, initialize logging, return error as value if any."""
+
+    deferred_error: confuse.ConfigError | None = None
+    try:
+        config.read()
```
snejus, line 916: I'd be happy to see the reasoning you have earlier here in the source code!

semohr: Added a more detailed comment

beets/ui/__init__.py
```diff
@@ -952,18 +918,57 @@ def parse_csl_callback(
...
+    # cli options
+    try:
+        config.set_args(options)
+    except confuse.ConfigError as e:
+        deferred_error = e
+
+    return deferred_error
+
+
+def _bootstrap_logging():
```
snejus: Last thing: just add a return type here

```suggestion
def _bootstrap_logging() -> None:
```

snejus (approved):

Issue #6193, opened 2025-11-28
fetchart: permissions errors result in a crash, needs graceful fallback


henry-oberholtzer:
### Problem

I think this is caused by my foorbar2000 instance checking the folder, regardless, fetchart should have a more graceful fallback in this situation.

Led to this problem:

```
Traceback (most recent call last):
  File "C:\Python312\Lib\site-packages\beets\util\__init__.py", line 504, in move
    os.replace(syspath(path), syspath(dest))
OSError: [WinError 17] The system cannot move the file to a different disk drive: '\\\\?\\C:\\Users\\Henry\\AppData\\Local\\Temp\\beets\\beetsplug_fetchart\\2svf0w77.jpg' -> '\\\\?\\E:\\Media Library\\Music\\Digital\\e\\Ed Rush & Optical\\Travel the Galaxy (2009, Virus Recordings VRS007DD) [FLAC 16-44]\\cover.jpg'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Python312\Lib\site-packages\beets\util\__init__.py", line 537, in move
    os.replace(tmp_filename, syspath(dest))
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: '\\\\?\\E:\\Media Library\\Music\\Digital\\e\\Ed Rush & Optical\\Travel the Galaxy (2009, Virus Recordings VRS007DD) [FLAC 16-44]\\.cover.jpg.jtsvsidc.beets' -> '\\\\?\\E:\\Media Library\\Music\\Digital\\e\\Ed Rush & Optical\\Travel the Galaxy (2009, Virus Recordings VRS007DD) [FLAC 16-44]\\cover.jpg'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Python312\Lib\site-packages\beets\util\__init__.py", line 541, in move
    raise FilesystemError(
beets.util.FilesystemError: The process cannot access the file because it is being used by another process while moving C:\Users\Henry\AppData\Local\Temp\beets\beetsplug_fetchart\2svf0w77.jpg to E:\Media Library\Music\Digital\e\Ed Rush & Optical\Travel the Galaxy (2009, Virus Recordings VRS007DD) [FLAC 16-44]\cover.jpg

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
[...]

PR #6662, opened 2026-05-21
fetchart: catch OSError in _set_art for graceful permission error handling

terminalchai:
﻿## Problem

When fetchart moves album art to the destination folder, a `PermissionError`
(or other `OSError`) can crash beets with an unhandled traceback. This happens
when a file is locked by another process, such as foobar2000 scanning the folder
during import.

Example traceback from #6193:

`PermissionError: [WinError 32] The process cannot access the file because it is being used by another process`

## Fix

Wrap `album.set_art()` in a `try/except OSError` block inside `_set_art`.
On failure, log a warning and return `False` instead of crashing. The
`batch_fetch_art` CLI path now checks the return value and reports
`"error writing album art"` instead of a traceback.

## Changes

- `beetsplug/fetchart.py`: `_set_art` now returns `bool`; catches `OSError` with a warning log
- `test/plugins/test_fetchart.py`: new test `test_set_art_oserror_is_handled_gracefully` verifies graceful handling when `Album.set_art` raises `PermissionError`
- `docs/changelog.rst`: bug fix entry added

Fixes #6193

PR #6765, opened 2026-06-23
Fix importfeeds to continue on symlink creation failures

### LINKED ISSUE #840: ImportFeeds:  fix/disable/catch symlink usage on windows
... because it doesnt work?

but it seems possible: http://en.wikipedia.org/wiki/NTFS_symbolic_link
http://lifehacker.com/5496652/how-to-use-symlinks-in-windows

```
  File "C:\Python27\Scripts\beet-script.py", line 9, in <module>
    load_entry_point('beets==1.3.6', 'console_scripts', 'beet')()
  File "C:\Python27\lib\site-packages\beets\ui\__init__.py", line 967, in main
    _raw_main(args)
  File "C:\Python27\lib\site-packages\beets\ui\__init__.py", line 958, in _raw_m
ain
  File "C:\Python27\lib\site-packages\beets\ui\commands.py", line 816, in import
_func
  File "C:\Python27\lib\site-packages\beets\ui\commands.py", line 788, in import
_files
    session.run()
  File "C:\Python27\lib\site-packages\beets\importer.py", line 262, in run
    pl.run_parallel(QUEUE_SIZE)
  File "C:\Python27\lib\site-packages\beets\util\pipeline.py", line 345, in run
    self.coro.send(msg)
  File "C:\Python27\lib\site-packages\beets\util\pipeline.py", line 158, in coro

  File "C:\Python27\lib\site-packages\beets\importer.py", line 1076, in manipula
te_files
    task.finalize(session)
  File "C:\Python27\lib\site-packages\beets\importer.py", line 395, in finalize
    self._emit_imported(session.lib)
  File "C:\Python27\lib\site-packages\beets\importer.py", line 427, in _emit_imp
orted
    plugins.send('album_imported', lib=lib, album=self.album)
  File "C:\Python27\lib\site-packages\beets\plugins.py", line 379, in send
    handler(**args)
  File "C:\Python27\lib\site-packages\beetsplug\importfeeds.py", line 138, in al
bum_imported
    _record_items(lib, album.album, album.items())
  File "C:\Python27\lib\site-packages\beetsplug\importfeeds.py", line 127, in _r
ecord_items
    os.symlink(syspath(path), syspath(dest))
AttributeError: 'module' object has no attribute 'symlink'
```

sammydynamo-dev:
## Description

Fixes #840.

When `importfeeds` is configured with `formats: link`, it creates a symlink for each imported track. If that symlink can't be created — on Windows without the privilege to make symlinks (the original report), but also on any OS when the destination directory isn't writable or the filesystem doesn't support symlinks — `beets.util.link()` raises `FilesystemError`. That exception wasn't caught, so it propagated out of the import pipeline and aborted the entire `beet import` run, even though the tracks had already been imported into the library.

This wraps the `link()` call in `ImportFeedsPlugin._record_items()` in `try/except FilesystemError`, logs a per-item warning, and continues with the remaining items — so a failed symlink is skipped rather than fatal. This is the "catch and skip" behaviour suggested in the issue thread. It catches `FilesystemError` (not `OSError`, which `util.link()` wraps) and mirrors the existing handling in `beetsplug/playlist.py` and the fetchart fix (#6193).

Adds five regression tests (warn-and-continue, that remaining items are still linked, that only `FilesystemError` is caught, and that a successful link still creates the symlink). Also verified manually: with a read-only feeds directory, `beet import` now finishes with a warning (exit `0`) instead of aborting (exit `1`), and the track is still imported.

## To Do

- [x] ~Documentation~ (bugfix; no user-facing flag or config changed)
- [x] Changelog
- [x] Tests

### DIFF
--- a/test/plugins/test_importfeeds.py
+++ b/test/plugins/test_importfeeds.py
@@ -1,8 +1,12 @@
 import datetime
 import os
+from unittest import mock
+
+import pytest
 
 from beets.library import Album, Item
 from beets.test.helper import PluginTestCase
+from beets.util import FilesystemError
 from beetsplug.importfeeds import ImportFeedsPlugin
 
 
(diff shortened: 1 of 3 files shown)


sammydynamo-dev:
Hi @snejus 👋 — this is my first contribution to beets.

This PR (#6765) fixes #840: when `importfeeds` is set to `formats: link` and a
symlink can't be created (e.g., on Windows, or into a read-only directory), the Failure is now logged as a per-item warning, and the import continues, instead of
The unhandled `FilesystemError` aborted the whole `beet import` run. It follows
the same warn-and-skip pattern as the fetchart fix (#6662) and `playlist.py`.

It includes five regression tests and a changelog entry, and the full test suite
passes locally. I tagged you since you reviewed the very similar fetchart change —
No rush at all, but I'd appreciate a review whenever you have time. Thanks!

test/plugins/test_importfeeds.py
```diff
@@ -62,3 +66,94 @@ def test_playlist_per_session(self):
...
+        self.lib.add(item1)
+        self.lib.add(item2)
+
+        with mock.patch(
+            "beetsplug.importfeeds.link",
+            side_effect=[FilesystemError("x", "link", (b"a", b"b")), None],
+        ) as mock_link:
+            self.importfeeds.album_imported(self.lib, album)
+
+        assert mock_link.call_count == 2
```
snejus: You can assert all of this using a single test.

test/plugins/test_importfeeds.py
```diff
@@ -62,3 +66,94 @@ def test_playlist_per_session(self):
...
+        self.feeds_dir.mkdir(parents=True, exist_ok=True)
+        album = Album(album="album/name", id=1)
+        item_path = os.path.join("path", "to", "item")
+        item = Item(title="song", album_id=1, path=item_path)
+        self.lib.add(album)
+        self.lib.add(item)
+
+        self.importfeeds.album_imported(self.lib, album)
+
+        linked = self.feeds_dir / os.path.basename(item_path)
```
snejus:
```suggestion
        linked = self.feeds_dir / item.filepath.name
```

snejus (changes requested):
Looks good, just a pair of comments in the tests.

test/plugins/test_importfeeds.py
sammydynamo-dev: Acknowledged. I will replace the three tests with a single test that asserts all three behaviors at once: two items, the first `link` call raises `FilesystemError`, the second succeeds, and `_log` is patched. This covers: the import doesn’t abort (no exception escapes), the failure is logged as a warning (`warning.assert_called_once`), and the remaining items are still linked (`call_count == 2`).

snejus (approved):
Thanks!

PR #6766, opened 2026-06-23
Add missing types to pipeline

snejus:
- `beets.util.pipeline` is now typed more explicitly. The main architectural change is making `Pipeline` generic over the messages it pulls and the stage coroutines it runs, then threading those types through the internal queue/thread helpers.

- The importer code now declares concrete `Pipeline[...]` types at the call sites that build import pipelines. This makes the pipeline contract clearer at the edges without changing how the pipeline is assembled or executed.

- High-level impact: this improves static type checking and makes the pipeline architecture easier to reason about, especially around stage inputs/outputs and threaded execution. The change is intended to be behavior-preserving.

- The PR also adds the pipeline typing commit to `.git-blame-ignore-revs` and documents how to use that file in `CONTRIBUTING.rst`. This helps keep future `git blame` output focused on real behavior changes instead of mechanical typing updates.

### DIFF
--- a/beets/util/pipeline.py
+++ b/beets/util/pipeline.py
@@ -37,12 +37,23 @@ import contextvars
 import queue
 import sys
 from threading import Lock, Thread
-from typing import TYPE_CHECKING, TypeVar
+from typing import TYPE_CHECKING, Any, Generic, overload
 
-from typing_extensions import TypeVarTuple, Unpack
+from typing_extensions import ParamSpec, TypeVar, TypeVarTuple, Unpack
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Generator
+    from collections.abc import (
+        Callable,
+        Generator,
+        Iterable,
+        Iterator,
+        Sequence,
+    )
+    from types import TracebackType
+
+    ExcInfo = tuple[
+        type[BaseException] | None, BaseException | None, TracebackType | None
+    ]
 
 BUBBLE = "__PIPELINE_BUBBLE__"
 POISON = "__PIPELINE_POISON__"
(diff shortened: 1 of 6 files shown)


beets/util/pipeline.py
```diff
@@ -388,57 +438,75 @@ def run(self):
...
         be at least two stages.
         """
         if len(stages) < 2:
             raise ValueError("pipeline must have at least two stages")
-        self.stages = []
-        for stage in stages:
+
+        first_stage = stages[0]
+        if isinstance(first_stage, (list, tuple)):
+            self.first_stage: Sequence[Any] = first_stage
```
snejus, line 464: @semohr you asked about this in the previous PR: note I separated stages into `first_stage` and `stages` attributes so that we can type them properly.

beets/util/pipeline.py
```diff
@@ -388,57 +438,75 @@ def run(self):
...
+class Pipeline(Generic[Tpull, Tstage]):
     """Represents a staged pattern of work. Each stage in the pipeline
     is a coroutine that receives messages from the previous stage and
     yields messages to be sent to the next stage.
     """
 
-    def __init__(self, stages):
+    @overload
+    def __init__(
+        self: Pipeline[Tpull, Tstage],
```
semohr: Do we need to type self here?

beets/util/pipeline.py
```diff
@@ -388,57 +438,75 @@ def run(self):
...
+    def run_parallel(self, queue_size: int = DEFAULT_QUEUE_SIZE) -> None:
         """Run the pipeline in parallel using one thread per stage. The
         messages between the stages are stored in queues of the given
         size.
         """
         base_ctx = contextvars.copy_context()
-        queue_count = len(self.stages) - 1
-        queues = [CountedQueue(queue_size) for i in range(queue_count)]
-        threads = []
+        queue_count = len(self.stages)
```
semohr, line 489: How come we need one more queue now?

beets/util/pipeline.py
```diff
@@ -184,7 +205,7 @@ def stage(func: Callable[[Unpack[A], T], R | None]):
     [3, 4, 5]
     """
 
-    def coro(*args: Unpack[A]) -> Generator[R | T | None, T, None]:
+    def coro(*args: Unpack[A]) -> Generator[Any, T, None]:
```
semohr: How come the yield type is now `Any` ? Imo `R | T | None` was the correct type hint beforehand.

beets/util/pipeline.py
snejus: `self.stages` does not include the first stage any more.

semohr (approved):
Looks good to me 👍

PR #6768, opened 2026-06-24
tidal: replace `AlbumAttributes`/`TrackAttributes` union with `Attributes`

snejus:
@semohr, as discussed:

- Extracts the Tidal fields shared by albums and tracks into a new `Attributes` `TypedDict` in `beetsplug/tidal/api_types.py`.

- Updates parser helpers in `beetsplug/tidal/__init__.py` to depend on `Attributes` instead of the `AlbumAttributes | TrackAttributes` union when they only read common metadata like title, links, label, and popularity.

- High-level impact: this reduces duplicated type definitions, makes the shared contract clearer, and simplifies future changes to common Tidal metadata without changing album- and track-specific types.

### DIFF
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -78,7 +78,7 @@ class TidalPlugin(MetadataSourcePlugin):
         """Return the configured path to the token file in the app directory."""
         return self.config["tokenfile"].get(confuse.Filename(in_app_dir=True))
 
-    def require_authentication(self):
+    def require_authentication(self) -> None:
         if not os.path.isfile(self._tokenfile()):
             raise UserError(
                 "Please login to TIDAL"
@@ -432,7 +432,7 @@ class TidalPlugin(MetadataSourcePlugin):
         return artist_names, artist_ids
 
     @staticmethod
-    def _parse_title(attributes: MediaAttributes):
+    def _parse_title(attributes: MediaAttributes) -> str:
         """
         Tidal UIs append the version string at the end of the title. We do the same here
         by formatting it as ``"{title} ({version})"`` to stay consistent.
@@ -565,7 +565,9 @@ class TidalPlugin(MetadataSourcePlugin):
             default=False,
         )
 
-        def auth_func(lib: Library, opts: optparse.Values, args: list[str]):
+        def auth_func(
+            lib: Library, opts: optparse.Values, args: list[str]
+        ) -> None:
             if opts.auth:
                 self.api.ui_authenticate_flow()
             else:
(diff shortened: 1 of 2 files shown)


beetsplug/tidal/api_types.py
```diff
@@ -42,38 +42,39 @@ class ArtistAttributes(TypedDict):
     spotlighted: NotRequired[bool]
 
 
-class AlbumAttributes(TypedDict):
+class Attributes(TypedDict):
```
semohr: We should rename this. `Attributes` implies that it is the common base type for all `*Attributes` types, but that is not how it is currently used. In practice, it only contains fields shared between track and album attributes.

A name that reflects its actual purpose makes more sense in my opinion. Maybe somthing like `SharedTrackAndAlbumAttributes`, `CommonTrackAlbumAttributes`, or `TrackAlbumAttributesBase`.

semohr (commented):
> High-level impact: this reduces duplicated type definitions, makes the shared contract clearer, and simplifies future changes to common Tidal metadata without changing album- and track-specific types.


I disagree with this point. While this reduces duplication, I think it makes future API changes harder to reason about because the types no longer align directly with the TIDAL contract. When the API changes, it's no longer sufficient to compare a resource definition against its corresponding type; instead, you need to trace shared fields through an additional abstraction layer. That indirection makes it more difficult to validate that our types still faithfully represent the upstream API. If the TIDAL API ever diverges between albums and tracks, I'll happily dust off this comment and collect my "I told you so" points :rofl:

beetsplug/tidal/api_types.py
snejus: I guess I'll borrow the notation from `_utils/musicbrainz.py` and call it `AttributesBase` 😆

semohr: You talking about `RelationBase`, right? That one is used for all `Relations` in musicbrainz. Here it is different, `AttributesBase` is only used for `AlbumnAttributes` and `TrackAttributes` and not for `ArtistAttributes` (or `ArtworkAttributes` once the upcoming feature PR is merged).

snejus: Oh shet, didn't realise that. Had a bit of a tunnel vision there. Renamed it to `MediaAttributes`.

beetsplug/tidal/api_types.py
```diff
@@ -42,38 +42,39 @@ class ArtistAttributes(TypedDict):
     spotlighted: NotRequired[bool]
 
 
-class AlbumAttributes(TypedDict):
+class MediaAttributes(TypedDict):
```
semohr, line 45: Can we put a small comment here for orientation? Maybe something like:

```
While this type is not part of the original TIDAL json:api schema itself, we introduced it to simplify our type definitions and allow for easier reuse of shared attributes.
```

beetsplug/tidal/__init__.py
```diff
@@ -432,7 +432,7 @@ def _parse_artists(
         return artist_names, artist_ids
 
     @staticmethod
-    def _parse_title(attributes: AlbumAttributes | TrackAttributes):
+    def _parse_title(attributes: MediaAttributes):
```
semohr: nit: We might want to add a returntype here.

semohr (approved):

PR #6771, opened 2026-06-24
tidal: normalize `albumtype` and add `albumtypes`

snejus:
- Normalizes Tidal album metadata so `albumtype` is always stored in lowercase, matching the behavior of other autotagger plugins and making imported metadata more consistent across sources.
- Extends the Tidal import path to populate `albumtypes` alongside `albumtype`, so albums now carry both the primary normalized type and the list-based field expected by the wider metadata model.
- Refactors parsing coverage in `test/plugins/test_tidal.py` by replacing four narrower parsing tests with a single data-driven high-level test. This makes the expected imported album shape easier to read, verifies more fields in one place, and keeps future parser changes easier to review.

### DIFF
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -369,7 +369,7 @@ class TidalPlugin(MetadataSourcePlugin):
             artist=", ".join(artist_names),
             artists=artist_names,
             duration=self._duration_to_seconds(album["attributes"]["duration"]),
-            albumtype=album["attributes"]["albumType"].lower(),
+            albumtypes=[album["attributes"]["albumType"].lower()],
             label=self._parse_label(album["attributes"]),
             year=date_parts[0] if date_parts else None,
             month=date_parts[1] if date_parts else None,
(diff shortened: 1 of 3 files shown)


beetsplug/tidal/__init__.py
```diff
@@ -369,7 +370,8 @@ def _get_album_info(
             artist=", ".join(artist_names),
             artists=artist_names,
             duration=self._duration_to_seconds(album["attributes"]["duration"]),
-            albumtype=album["attributes"]["albumType"],
+            albumtype=album_type,
```
semohr: 🤔 Intriguing. I could look this up but it is faster to ask here:

Dont we have a fallback system for stuff like this? What I mean if `albumtypes` is empty we use `albumtype` or am i confusing that with mediafile specific logic?

snejus: You're right - forgot about this. I think we should set `albumtypes` though, instead of `albumtype`.

I think I will be removing `albumtype` field soon, feel like it has the same fate like `genre`.

semohr: Nice, thought I have seen something like that before! This makes things a bit easier I guess.

semohr (approved):

PR #6774, opened 2026-06-24
test: split `test_dbcore.py` and `test_files.py`

snejus:
#### What changed

- Reorganized tests so they live closer to the subsystem they cover:
  - query parsing tests moved from `test/test_dbcore.py` to `test/test_queryparse.py`
  - utility/path tests moved from `test/test_files.py` to `test/test_util.py`
- Added shared lightweight db test fixtures in `beets/test/fixtures.py`, and updated tests to import `ModelFixture1` and `SortFixture` from there.
- Added the related refactor commits to `.git-blame-ignore-revs` so `git blame` stays focused on real logic changes.

#### Why this matters

- The test suite is now structured by responsibility instead of growing large mixed-purpose test files.
- Shared fixtures remove duplicate dbcore test setup and make cross-file reuse simpler.
- This lowers coupling between test modules and makes future test changes easier to find and review.

#### High-level impact

- Test-only refactor; no production behavior changes.
- Improves maintainability, readability, and ownership of the test architecture.
- Keeps blame history cleaner after the test file split.


beets/test/fixtures.py
henry-oberholtzer, line 1: I wonder if a better place for this would be under `/test/dbcore/fixtures.py` (once 6775 is merged in) since the fixtures in here are so tightly coupled to dbcore and aren't used elsewhere in the tests as far as I am aware.

snejus: #6775 depends on this PR haha. I initially thought so too, however two reasons why I kept it here:

1. In #6775, `DummyIMBackend` is moved there too, used by `test/plugins/test_embedart.py` and `test/util/test_art_resize.py`. Thus it does not belong only in `dbcore` tests any more.
2. Best practices say that we should not import stuff from `test` subdirectories - and our shared infra lives under `beets.test`.

snejus:
I'm merging this one as a fair bit of tests improvements I'm working on depends on these changes.

PR #6775, opened 2026-06-24
test: move test modules under packages that they belong to

snejus:
- Move test files from the top-level `test/` directory into package-specific folders like `test/dbcore/`, `test/plugins/`, `test/util/`, and `test/extra/`.

- Keep test coverage the same, but make the test suite match the codebase structure more closely. This makes ownership and navigation clearer, and helps reviewers find related tests faster.

- Add the large file-move commit to `.git-blame-ignore-revs` so `git blame` stays useful and points at real logic changes instead of this mechanical reorganization.


semohr (approved):
LGTM

PR #6777, opened 2026-06-25
 Remove per-file MIT license headers.

semohr:
## Description

Every `.py` file in the repo carries the full 13-line MIT license text as a header. The license is already shipped in the distribution, it's in the top-level `LICENSE` file and included in our sdist/wheels. Duplicating it across 170+ source files is just noise imo: it pushes the actual code down, adds nothing for users or contributors, and creates churn whenever a new file is scaffolded.

This PR strips the headers from all Python files. No functional change, purely cosmetic cleanup.

```
wheel:  651,685 -> 615,351 bytes  (―36 KB)
sdist: 2,382,290 -> 2,369,828 bytes  (―12 KB)
────
                total: -48 KB
```

---

If anyone has strong feelings about this, let's discuss.


henry-oberholtzer (approved):
I think this is a great simplification & saves a lot of trouble with needing to adjust copyright in each file too.

arsaboo:
Thanks for this.....I never understood the value of this.

snejus (approved):
Looks good, thanks!

PR #6781, opened 2026-06-27
tidal: respect search_limit config option in candidates()

### LINKED ISSUE #6770: tidal: `search_limit` is ignored
### Problem

The Tidal plugin does not appear to honor `tidal.search_limit`. With a config such as:

```yaml
tidal:
  search_limit: 1
```

Tidal album and track searches still return (at least 20+) results.

### Expected behavior

Tidal should cap album and track search candidates using the configured `search_limit`, matching the behavior of other metadata source plugins.

### Likely cause

Tidal implements `candidates()` and `item_candidates()` directly, then calls `api.search_results()` from `search_albums_by_query()` and `search_tracks_by_query()`. That bypasses the shared search path that reads `search_limit` into `SearchParams.limit`.

### Suggested fix

Thread `search_limit` through the Tidal query search path, or truncate returned Tidal relationships before resolving album/track IDs. Add focused tests for both album and track search with `search_limit: 1`.

Tagging @semohr for visibility.

kveni12:
Fixes #6770

## Problem

`TidalPlugin` extends `MetadataSourcePlugin`, which sets a default `search_limit: 5` config value. However, `candidates()` and `item_candidates()` called `search_albums_by_query()` and `search_tracks_by_query()` without any limit, so the config option was silently ignored.

## Fix

- Add an optional `limit: int | None = None` parameter to `search_albums_by_query()` — slices `album_ids` before passing to `search_albums_by_ids()`
- Add an optional `limit: int | None = None` parameter to `search_tracks_by_query()` — slices the track relationships list before iterating
- `candidates()` reads `self.config["search_limit"].get(int)` and passes it through
- `item_candidates()` does the same

The fix is intentionally minimal — no changes to `api.py` or the Tidal HTTP request; results are simply sliced client-side, consistent with how other direct-implementation plugins cap their results.

## Tests

Added `TestSearchLimit` with two tests:
- `test_candidates_respects_search_limit` — sets `search_limit: 1`, mocks `search_results` returning 2 album IDs, asserts only 1 candidate is returned
- `test_item_candidates_respects_search_limit` — same pattern for tracks

### DIFF
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -4,8 +4,9 @@ import itertools
 import os
 import re
 import time
+from collections import deque
 from functools import cached_property
-from typing import TYPE_CHECKING, ClassVar, Literal, overload
+from typing import TYPE_CHECKING, ClassVar, Literal, TypeVar, overload
 
 import confuse
 
@@ -20,7 +21,7 @@ from .api import TidalAPI
 
 if TYPE_CHECKING:
     import optparse
-    from collections.abc import Callable, Iterable, Sequence
+    from collections.abc import Callable, Iterable, Iterator, Sequence
 
     from beets.autotag import Info
     from beets.dbcore.db import Results
@@ -39,6 +40,20 @@ if TYPE_CHECKING:
 
 
 log = getLogger("beets.tidal")
+
+T = TypeVar("T")
+
+
+def round_robin(iterables: Iterable[Iterable[T]]) -> Iterator[T]:
+    """Yield one element from each iterable in turn until all are exhausted."""
+    iterators = deque(map(iter, iterables))
+    while iterators:
+        try:
+            yield next(iterators[0])
+        except StopIteration:
+            iterators.popleft()
+        else:
+            iterators.rotate(-1)
 
 
 class TidalPlugin(MetadataSourcePlugin):
@@ -132,27 +147,9 @@ class TidalPlugin(MetadataSourcePlugin):
         ):
             return candidates
 
-        candidates = list(
-            itertools.islice(
-                (
-                    candidate
-                    for candidate in itertools.chain.from_iterable(
-                        itertools.zip_longest(
-                            *(
-                                self.search_albums_by_query(query)
-                                for query in self._album_queries(items)
-                            ),
-                            fillvalue=None,
-                        )
-                    )
-                    if candidate is not None
-                ),
-                self.search_limit,
-            )
+        return self._search_by_queries(
+            self.search_albums_by_query, self._album_queries(items)
         )
-
-        log.debug("Found {0} candidates", len(candidates))
-        return candidates
 
     def item_candidates(
         self, item: Item, artist: str, title: str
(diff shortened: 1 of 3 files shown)


semohr (changes requested):
Thanks for already taking a look here!

We should be able to resolve all search limit based logic within the `candidates` and  `item_candidates` function. I don't really see the need for the extensive drilling. 

The `search_tracks_by_query` and `search_albums_by_query` functions are returning iterators. We should be able to break them early and limit the number of requests that way. 

Additionally the `search_limit` currently does not respect the  `_album_queries` and `_item_queries` loop. We might want to have a round robin approach here.

---

I will have a more detailed look next week.

snejus:
I've added it to my fork using a slightly simpler solution https://github.com/beetbox/beets/commit/626963d77e53e1f94d20e6ebe6fcebc9131cdd44

kveni12:
hi @semohr thank you for the feedback! I completely agree, so moved the limit logic entirely into candidates() and item_candidates() using a round-robin loop that early-breaks across query iterators, so search_albums_by_query and search_tracks_by_query stay clean. Also added a test covering the multi-query interleaving behaviour.

beetsplug/tidal/__init__.py
```diff
@@ -147,8 +160,21 @@ def item_candidates(
             ):
                 return candidates
 
-        for query in self._item_queries(item):
-            candidates += self.search_tracks_by_query(query)
+        search_limit = self.config["search_limit"].get(int)
```
snejus: I'd suggest using `itertools.chain.from_iterable` and `itertools.islice` in both of these methods.

beetsplug/tidal/__init__.py
```diff
@@ -128,8 +128,21 @@ def candidates(
         ):
             return candidates
 
-        for query in self._album_queries(items):
-            candidates += self.search_albums_by_query(query)
+        search_limit = self.config["search_limit"].get(int)
```
snejus: Define this as `@cached_attribute`.

beetsplug/tidal/__init__.py
```diff
@@ -147,8 +165,24 @@ def item_candidates(
             ):
                 return candidates
 
-        for query in self._item_queries(item):
-            candidates += self.search_tracks_by_query(query)
+        candidates = list(
```
semohr, line 179: Lets make the round robin logic a helper function. Should make the flow here more readable and should allow to dedupe some code.

beetsplug/tidal/__init__.py
```diff
@@ -128,16 +131,31 @@ def candidates(
...
             return candidates
 
-        for query in self._album_queries(items):
-            candidates += self.search_albums_by_query(query)
+        candidates = list(
+            itertools.islice(
+                (
+                    candidate
+                    for candidate in itertools.chain.from_iterable(
+                        itertools.zip_longest(
```
snejus: What is `zip_longest` here for? Also, you can use `filter(None, ...` instead of checking for `None` values.

semohr:
@kveni12 Do you intend to come back to this?

kveni12:
Sorry for the delay! i've just implemented fixes based on the comments.

semohr (approved):
LGTM! Thank you. Lets also wait for approval from @snejus.

You can already rebase and solve the conflicts tho.

snejus:
@kveni12 would you mind rebasing?

PR #6788, opened 2026-06-28
Delay Dolphin metadata until thumbnails finish

### LINKED ISSUE #3086: thumbnails: Do image conversion before writing .desktop entry for Dolphin
### Problem

So I'm trying to fetch and embed album art into each song and the album folder. I am using the fetchart, thumbnails, and embedart plugins. Since I am using dolphin, I have that option set to 'yes'.

When I run an import, I am getting a .jpg file that matches the name set for 'art_filename' and that file is added to the .directory file, which is apparently how dolphin manages it's thumbnails. So this seems to work... however, when I open up a new instance of dolphin and navigate to the album folder, the thumbnail is gone. When I go into the folder settings and manually set the thumbnail, .jpg isn't in their filtered file extensions, making me think it might not be supported? Googling around hasn't given me any answers (but .pngs seem to persist so it sounds plausible).

Anyway, I tried to get some more insight, so I ran with the verbose settings and I see the below output (art had to be pulled through google).

> Sending event: art_set
> embedart: Embedding album art into The Absence - Oceans
> embedart: embedding /music/Artists/The Absence/[2013] (Single) Oceans/.albumart.jpg
> Sending event: write
> Sending event: after_write
> thumbnails: generating thumbnail for The Absence - Oceans
> thumbnails: Wrote file /music/Artists/The Absence/[2013] (Single) Oceans/.directory
> artresizer: PIL resizing /music/Artists/The Absence/[2013] (Single) Oceans/.albumart.jpg to /config/.cache/thumbnails/large/3012ea475fee1a34403bed3321f06958.png
> artresizer: PIL resizing /music/Artists/The Absence/[2013] (Single) Oceans/.albumart.jpg to /config/.cache/thumbnails/normal/3012ea475fee1a34403bed3321f06958.png
> thumbnails: wrote thumbnail for The Absence - Oceans

According to this, it is writing the .directory entry before generating the thumbnails and the thumbnails are .png files. When I run through the dev environment with the debugger, I see a call to

```python
shutil.move(resized, target)
```

where resized and target are the same path. I'm assuming the intention was to move those back to the album directory? If it correctly moved the thumbnail back there, it wouldn't match the extension from the file that was added to the .directory file either. So I guess there are 2 bugs here?

1. target is supposed to be in the album folder
2. .directory shouldn't be written to until the desired thumbnail is actually created (which will hopefully be a .png if it is successfully moved)

### Setup

* OS: Archlinux + KDE
* Python version: 2.7, 3.7
* beets version: 1.4.8

The important parts of my configuration:

```yaml
# problem still exists without the .
art_filename: .albumart

fetchart:
    auto: yes
    cautious: false
    store_source: yes
    cover_names: front cover
    sources: 
        - coverart: release 
        - amazon
        - itunes 
        - coverart: releasegroup
        - albumart 
        - wikipedia
        - google
    google_key: *****

thumbnails:
    auto: yes
    force: yes
    dolphin: yes   

embedart:
    auto: yes
    remove_art_file: no     
```

cookesan:
## Summary

- Create Dolphin `.directory` metadata after thumbnail generation has run.
- Skip thumbnail tagging and moves when resizing returns the original art path.
- Avoid moving resized output when the resizer already wrote the target file.

Fixes #3086.

## Checklist

- [x] Tests
- [x] Changelog
- [x] Documentation not needed for this bug fix

## Checks

- `poe test -- test/plugins/test_thumbnails.py test/test_art_resize.py test/plugins/test_art.py`
- `poe lint -- beetsplug/thumbnails.py test/plugins/test_thumbnails.py`
- `poe check-format -- beetsplug/thumbnails.py test/plugins/test_thumbnails.py`
- `git diff --check`

Note: GIO and ImageMagick backend tests skipped locally because those system libraries are not installed. The PIL-backed resize tests ran.

### DIFF
--- a/beetsplug/thumbnails.py
+++ b/beetsplug/thumbnails.py
@@ -115,9 +115,6 @@ class ThumbnailsPlugin(BeetsPlugin):
             self._log.info("album {} has no art", album)
             return
 
-        if self.config["dolphin"]:
-            self.make_dolphin_cover_thumbnail(album)
-
         size = ArtResizer.shared.get_size(album.artpath)
         if not size:
             self._log.warning(
@@ -129,6 +126,9 @@ class ThumbnailsPlugin(BeetsPlugin):
         if max(size) >= 256:
             wrote &= self.make_cover_thumbnail(album, 256, LARGE_DIR)
         wrote &= self.make_cover_thumbnail(album, 128, NORMAL_DIR)
+
+        if self.config["dolphin"]:
+            self.make_dolphin_cover_thumbnail(album)
 
         if wrote:
             self._log.info("wrote thumbnail for {}", album)
@@ -161,8 +161,12 @@ class ThumbnailsPlugin(BeetsPlugin):
                 )
                 return False
         resized = ArtResizer.shared.resize(size, album.artpath, target)
+        if resized == album.artpath:
+            return False
+
         self.add_tags(album, resized)
-        shutil.move(syspath(resized), syspath(target))
+        if resized != target:
+            shutil.move(syspath(resized), syspath(target))
         return True
 
     def thumbnail_file_name(self, path):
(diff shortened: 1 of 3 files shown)


snejus:
Using AI to fix issues in bulk (across many repositories) is not allowed here.

PR #6797, opened 2026-06-28
Detach stdin when running badfiles checker commands

### LINKED ISSUE #6750: Unable to see terminal input after beet bad
### Problem

```sh
$ beet -vv bad
```

Led to this problem:

```
After `beet bad` finishes, my terminal is left in a broken state where I cannot see what I type at the next shell prompt. Input still appears to be accepted, but typed characters are not echoed.

Pressing Ctrl-C shows `^C`, but it does not restore normal terminal input. The spacing between `^C` and the next prompt also looks larger than normal.

Running either of these fixes the terminal until the next time I run `beet bad`:

reset
stty sane

Redirecting stdout/stderr does not fully solve the issue:

beet bad > ~/beet-bad.log 2>&1

This makes me suspect that `beet bad`, or one of the external checker commands invoked by the `badfiles` plugin, is leaving the terminal/TTY in a modified state, possibly with echo disabled.
```

### Setup

* OS: Ubuntu Linux
* Python version: TODO: paste output of `python --version`
* beets version: 2.11.0
* Turning off plugins made problem go away (yes/no): TODO

```yaml
directory: /path/to/Music
library: ~/path/to/beets/library.db
paths:
    default: $albumartist/$album/$id-$title
    singleton: Non-Album/$artist/$id-$title
    comp: Compilations/$album/$id-$title
replace:
    '[\\/]': _
    '^\.+': _
    '[\x00-\x1f]': _
    '[<>:"\?\*\|]': _
    '\.$': _
    '\s+': _
import:
    copy: yes
    write: yes 
plugins:
    - musicbrainz
    - chroma
    - badfiles
    - info
    - unimported
badfiles:
    check_on_import: yes
    commands:
      m4a: /path/to/Scripts/music/check-m4a-ffmpeg
```

Additional notes:

`beet bad` is mainly just outputing these warnings which I am working with:

```text
WARNING, cannot check MD5 signature since it was unset in the STREAMINFO
```

I do not know whether that warning is related to the terminal state issue. The terminal echo problem happens after running `beet bad`, and `reset` or `stty sane` restores the terminal.

augboz:
Closes #6750.

The `badfiles` checker commands (`mp3val`, `flac`, custom) are non-interactive, but `run_command` launched them with the parent's stdin still attached to the controlling terminal. A checker can leave the TTY in a modified state (e.g. echo disabled), so after `beet bad` the terminal stops echoing typed input until `reset`/`stty sane`.

**Fix:** pass `stdin=DEVNULL` to `check_output` so checkers can't read from or modify the terminal.

**Note on verification:** the symptom is terminal-state-specific and I couldn't reproduce it on my (Windows) machine, but the root cause is clear from the code path. I've added a regression test asserting `run_command` calls `check_output` with `stdin=DEVNULL`, and a changelog entry.

### DIFF
--- a/beetsplug/badfiles.py
+++ b/beetsplug/badfiles.py
@@ -18,7 +18,13 @@ import errno
 import os
 import shlex
 import sys
-from subprocess import STDOUT, CalledProcessError, check_output, list2cmdline
+from subprocess import (
+    DEVNULL,
+    STDOUT,
+    CalledProcessError,
+    check_output,
+    list2cmdline,
+)
 from typing import Literal
 
 import confuse
@@ -66,7 +72,10 @@ class BadFiles(BeetsPlugin):
             "running command: {}", displayable_path(list2cmdline(cmd))
         )
         try:
-            output = check_output(cmd, stderr=STDOUT)
+            # The checker commands are non-interactive, so detach stdin from
+            # the controlling terminal. Otherwise a checker may leave the TTY
+            # in a modified state (e.g. with echo disabled). See #6750.
+            output = check_output(cmd, stderr=STDOUT, stdin=DEVNULL)
             errors = 0
             status = 0
         except CalledProcessError as e:
--- a/beetsplug/badfiles.py
+++ b/beetsplug/badfiles.py
@@ -6,7 +6,13 @@ import errno
 import os
 import shlex
 import sys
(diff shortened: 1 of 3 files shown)


snejus:
@augboz would you mind rebasing this? I added types to this module :)

PR #6798, opened 2026-06-28
tidal: honor manual search terms

### LINKED ISSUE #6795: tidal: honor manually entered artist and album search terms
### Problem

The Tidal plugin ignores the artist and album entered through the importer's `Enter search` action. Beets passes these values to `TidalPlugin.candidates()`, but the plugin builds its queries from the existing item tags via `_album_queries(items)` instead.

This makes manual search repeat the original, incorrect search when users are trying to correct bad or missing tags. The singleton path has the same problem: `item_candidates()` ignores its artist and title arguments and uses `_item_queries(item)`.

To reproduce:

1. Enable the Tidal plugin and import an album with incorrect artist or album tags.
2. Choose `Enter search`.
3. Enter a different artist and album.
4. Observe that Tidal searches using the original tags rather than the entered values.

### Expected behavior

Tidal should use the manually entered artist and album or track title when Beets supplies explicit search terms.

### Suggested fix

Build Tidal text queries from the artist and album/title arguments accepted by `candidates()` and `item_candidates()`, while retaining barcode and ISRC lookup precedence. Add focused regression tests proving that explicit search terms override item metadata for both albums and singletons.

cookesan:
## Description

Fixes #6795.

The Tidal metadata plugin already receives explicit artist, album, and track terms from the importer, including Enter search. Album and track query fallback now use those terms instead of rebuilding queries from stored item metadata, while barcode and ISRC lookups still return first.

For album searches with no resolved artist, the previous per-item-artist fallback is kept and paired with the supplied album term.

## To Do

- [x] ~Documentation~
- [x] Changelog
- [x] Tests

## Validation

- `python -m pytest -q test/plugins/test_tidal.py test/test_metadata_plugins.py test/autotag/test_match.py`
- `ruff format --check beetsplug/tidal/__init__.py test/plugins/test_tidal.py`
- `ruff check beetsplug/tidal/__init__.py test/plugins/test_tidal.py`
- `mypy --follow-imports=skip beetsplug/tidal/__init__.py test/plugins/test_tidal.py`
- `git diff --check`

### DIFF
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import itertools
 import os
 import re
 import time
@@ -128,7 +127,7 @@ class TidalPlugin(MetadataSourcePlugin):
         ):
             return candidates
 
-        for query in self._album_queries(items):
+        for query in self._album_queries(items, artist, album):
             candidates += self.search_albums_by_query(query)
 
         log.debug("Found {0} candidates", len(candidates))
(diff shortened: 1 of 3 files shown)


snejus:
Using AI to fix issues in bulk (across many repositories) is not allowed here.

PR #6800, opened 2026-06-28
 replaygain: add metaflac backend

### LINKED ISSUE #1203: replaygain: metaflac backend
I'd like to use metaflac as a backend for the replaygain plugin.  On my NAS, neither PAL nor Gstreamer are feasible, but metaflac is available through ipkg.

SamadBallaj1:
## Description

Fixes #1203.

The replaygain plugin can't compute ReplayGain when metaflac is the only tool available. The existing backends (command, gstreamer, audiotools, ffmpeg) each need something heavier installed, which doesn't work on minimal setups like a NAS where those won't install. Issue #1203 asks for a metaflac based option so FLAC users on those setups can still tag ReplayGain.

This adds a metaflac backend for that case. It runs `metaflac --add-replay-gain` to compute the gain and reads the values back with `metaflac --show-tag`. I modeled it on the existing `CommandBackend` since both wrap an external tool. It only handles FLAC, skips other formats, and shifts the gain to the configured `targetlevel` like the other backends.

One limitation: metaflac scans a whole album in one pass, so the files of an album need the same sample rate and channel layout.

Before, selecting the metaflac backend failed:

```
$ beet replaygain
UserError: Selected ReplayGain backend metaflac is not supported.
```

After, the backend works and the plugin tests pass:

```
$ python -m pytest test/plugins/test_replaygain.py
7 passed, 4 skipped
```

The 4 skipped are the Opus R128 cases, which don't apply to the metaflac path.

## To Do

- [x] Documentation.
- [x] Changelog.
- [x] Tests.

### DIFF
--- a/beetsplug/replaygain.py
+++ b/beetsplug/replaygain.py
@@ -683,12 +683,11 @@ class MetaflacBackend(Backend):
     def compute_track_gain(self, task: AnyRgTask) -> AnyRgTask:
         """Compute the track gain for each FLAC item in the task."""
         track_gains = []
-        for item in task.items:
-            if self.format_supported(item):
-                self._add_replay_gain([item])
-                track_gains.append(
-                    self._read_gain(item, "TRACK", task.target_level)
-                )
+        for item in filter(self.format_supported, task.items):
+            self._add_replay_gain([item])
+            track_gains.append(
+                self._read_gain(item, "TRACK", task.target_level)
+            )
         task.track_gains = track_gains
         return task
 
--- a/beetsplug/replaygain.py
+++ b/beetsplug/replaygain.py
@@ -653,6 +653,102 @@ class CommandBackend(Backend):
         return out
 
 
+# metaflac CLI tool backend.
+
+
+class MetaflacBackend(Backend):
+    """A replaygain backend using the ``metaflac`` command-line tool."""
+
+    NAME = "metaflac"
+    do_parallel = True
+
+    SUPPORTED_FORMATS: ClassVar[set[str]] = {"FLAC"}
+
+    def __init__(self, config: ConfigView, log: Logger):
+        super().__init__(config, log)
+        config.add({"metaflac": "metaflac"})
+
+        command = config["metaflac"].as_str()
+        path = shutil.which(command)
+        if path is None:
+            raise FatalReplayGainError(
+                f"replaygain metaflac command not found: {command!r}"
+            )
+        self.command = path
+
+    def format_supported(self, item: Item) -> bool:
+        """Return whether metaflac can process this item."""
+        return item.format in self.SUPPORTED_FORMATS
+
+    def compute_track_gain(self, task: AnyRgTask) -> AnyRgTask:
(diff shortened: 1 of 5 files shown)


SamadBallaj1:
Pushed some updates, this is ready for review whenever you have a chance. Thanks.

beetsplug/replaygain.py
```diff
@@ -653,6 +653,102 @@ def parse_tool_output(self, text: bytes, num_lines: int) -> list[Gain]:
...
+
+    def format_supported(self, item: Item) -> bool:
+        """Return whether metaflac can process this item."""
+        return item.format in self.SUPPORTED_FORMATS
+
+    def compute_track_gain(self, task: AnyRgTask) -> AnyRgTask:
+        """Compute the track gain for each FLAC item in the task."""
+        track_gains = []
+        for item in task.items:
+            if self.format_supported(item):
```
snejus:
```suggestion
        for item in filter(self.format_supported, task.items):
```

test/plugins/test_replaygain.py
```diff
@@ -112,6 +115,16 @@ class FfmpegBackendMixin(BackendMixin):
     has_r128_support = True
 
 
+class MetaflacBackendMixin(BackendMixin):
+    plugin_config: ClassVar[dict[str, Any]] = {"backend": "metaflac"}
```
snejus, line 119: Would you mind installing `flac` in this part of the workflow, in `ci.yaml`?

```yaml
      - name: Install system dependencies on Ubuntu
```

snejus (changes requested):
Looks good, have you tested it yourself?

test/plugins/test_replaygain.py
SamadBallaj1: Thanks @snejus for the review. I updated the track loop to use filter like you suggested and also added flac to the Ubuntu deps so the metaflac tests run on CI. All clear now, both pushed.

beetsplug/replaygain.py
SamadBallaj1: Yes. I ran it locally on a FLAC file with the metaflac backend and it wrote the correct ReplayGain tags (track gain got arund -11.5 dB, peak close to 1.0). The metaflac tests also pass on my machine.

docs/plugins/replaygain.rst
```diff
@@ -139,6 +140,24 @@ file.
...
 
+metaflac
+~~~~~~~~
+
+This backend uses the metaflac_ command-line tool (part of the FLAC tools) to
+compute ReplayGain values for FLAC files. It only supports FLAC; files in other
+formats are skipped. To use it, install the ``flac`` package, which provides
+``metaflac``, and select the ``metaflac`` backend in your configuration file:
+
+::
```
snejus:
```suggestion
.. code-block:: yaml
```

snejus (changes requested):
Just one small comment to add highlighting to the docs and please rebase it on master to resolve the conflict in the changelog!

snejus (approved):
Thanks!

PR #6801, opened 2026-06-29
Fix Tidal plugin to honor manual importer search overrides

### LINKED ISSUE #6795: tidal: honor manually entered artist and album search terms

arnavahire19:
### PR Title

Fix Tidal plugin to honor manual importer search overrides

#### Problem

The Tidal plugin currently ignores the `artist` and `album`/`title` arguments passed to `TidalPlugin.candidates()` and `TidalPlugin.item_candidates()` during text query fallbacks. Instead, it builds search queries strictly from existing item tags via `_album_queries(items)` and `_item_queries(item)`.

This leads to a bug where using the importer's `Enter search` action to correct bad or missing tags simply repeats the original query, ignoring the manually entered search parameters.

Fixes #6795

#### Solution

* Updated `candidates()` to build queries using the passed `artist` and `album` arguments (i.e., `album` if `va_likely` is true, otherwise `artist album`).
* Updated `item_candidates()` to use the passed `artist` and `title` arguments.
* Kept the precedence of barcode and ISRC lookups before fallback query lookups.
* Removed the now-unused helper methods `_album_queries` and `_item_queries`, along with the unused `itertools` import.
* Added regression tests in `test/plugins/test_tidal.py` to ensure that manual overrides correctly direct the Tidal API search queries.

#### Tests

Ran tests locally using `pytest`:
`pytest test/plugins/test_tidal.py` -> 49 passed.

### DIFF
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import itertools
 import os
 import re
 import time
@@ -129,8 +128,12 @@ class TidalPlugin(MetadataSourcePlugin):
         ):
             return candidates
 
-        for query in self._album_queries(items):
-            candidates += self.search_albums_by_query(query)
+        query = (
+            album.strip()
+            if va_likely
+            else f"{artist.strip()} {album.strip()}".strip()
+        )
+        candidates += self.search_albums_by_query(query)
 
         log.debug("Found {0} candidates", len(candidates))
         return candidates
@@ -148,29 +151,15 @@ class TidalPlugin(MetadataSourcePlugin):
             ):
                 return candidates
 
-        for query in self._item_queries(item):
-            candidates += self.search_tracks_by_query(query)
+        query = (
+            f"{artist.strip()} {title.strip()}"
+            if artist.strip()
+            else title.strip()
+        )
+        candidates += self.search_tracks_by_query(query)
 
         log.debug("Found {0} candidates", len(candidates))
         return candidates
-
-    @staticmethod
-    def _item_queries(item: Item) -> Iterable[str]:
-        """Search queries for items."""
-        yield item.title
-
-        if item.artist:
-            yield f"{item.artist} {item.title}"
-
-    @staticmethod
-    def _album_queries(items: Sequence[Item]) -> Iterable[str]:
-        """Search queries for albums."""
-
-        album_names = set(i.album for i in items)
-        artist_names = set(i.artist for i in items)
-
-        for album, artist in itertools.product(album_names, artist_names):
-            yield f"{artist} {album}"
 
     def search_tracks_by_query(self, query: str) -> Iterable[TrackInfo]:
         """Search for tracks given a string query."""
--- a/test/plugins/test_tidal.py
+++ b/test/plugins/test_tidal.py
@@ -672,6 +672,46 @@ class TestCandidates(TidalPluginTest):
         assert len(candidates) == 1
         assert candidates[0].album == "Query Album"
 
+    def test_candidates_with_manual_search_override(self):
+        """Test candidates uses manual search terms instead of items tags."""
+        items = [
+            Item(
+                title="Original Song",
+                artist="Original Artist",
+                album="Original Album",
+            )
+        ]
+
+        self.tidal.api.search_results = Mock(
+            return_value={
+                "data": {
+                    "relationships": {
+                        "albums": {"data": [{"id": "1", "type": "albums"}]}
+                    }
+                }
+            }
+        )
+
+        track = _make_track("101", "Album Track", "PT3M", "ISRC001", ["1001"])
+        album, track_lookup, artist_lookup = _make_album(
+            "1", "Query Album", [track], ["1001"]
+        )
+        self.tidal.api.get_albums = Mock(
+            return_value={
+                "data": [album],
+                "included": [*artist_lookup.values(), *track_lookup.values()],
+            }
+        )
+
+        list(
+            self.tidal.candidates(items, "Manual Artist", "Manual Album", False)
+        )
+
+        # Should search using manual query terms, not item tags
+        self.tidal.api.search_results.assert_called_once_with(
+            "Manual Artist Manual Album", include=["albums"]
+        )
+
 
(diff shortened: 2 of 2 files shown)


semohr:
Thanks for your interest in beets.

Could you explain, in your own words and without using AI, why you decided to remove the existing item tag paths and what the implications are?

arnavahire19:
@semohr 
So, we decided to ditch the old way of building queries by scraping every track's metadata and switched to just using the artist/album info Beets gives us. Here’s why:

 **Fixes Manual Searches:** Before, when you’d manually fix tags during an import, the plugin ignored you and just re-ran the original (wrong) search. Now, it actually uses your input.
 **Way More Efficient:** The old `_album_queries` was creating a massive mess of redundant API calls to Tidal every time tags were inconsistent. Now it’s just one clean query per search.

### The TL;DR:

**Behavioral:** Manual searches now actually work like they're suppose to.
**Performance:** Way fewer API calls, so you won't hit Tidal's rate limits as much.
 **Consistency:** Keeps the code in line with the major plugins like Beatport and Deezer.

semohr:
> Way More Efficient: The old _album_queries was creating a massive mess of redundant API calls to Tidal every time tags were inconsistent. Now it’s just one clean query per search.

We actually use multiple queries because the Tidal API doesn’t handle search terms very well. The extra overhead is intentional and improves result quality significantly. I also intentionally sidestepped the majority vote logic for artist and title here for the same reason.

> Consistency: Keeps the code in line with the major plugins like Beatport and Deezer.

Tidal is the newest plugin and in my opinion one of the cleanest metadata plugin implementations in beets. We should not try to compare it to the `Beatport` and `Deezer` plugin too much. These are probably the least maintained ones in beets. The beatport plugin is even deprecated and will likely be removed in 3.0.0.

---

@snejus Input wanted:

At the moment there’s no clean plugin-level way to distinguish whether album or artist strings originate from user input versus being inferred (see [`tag_album` function](https://github.com/beetbox/beets/blob/08081f3c440c3aff2ff66931af645c92edfceb47/beets/autotag/match.py#L383)).

That effectively forces us into using `likelies` (majority vote) for artists and titles. However, this reduces tidal match quality because it discards valid alternative interpretations too early.

I’d prefer to keep the more verbose multi-query approach since it improves search results. Any ideas on how to better surface "user intent vs inferred metadata" without breaking the existing plugins? (Was thinking e.g. changing the `candidates` signature could work)

arnavahire19:
Thanks for the explanation, @semohr! That makes complete sense. I appreciate the context on     
  Tidal's search API behavior and why the multi-query approach is necessary for match quality      
  there, as well as the heads-up regarding Beatport/Deezer plugin maintenance status.              
                                                                                                   
  Regarding the architectural challenge of distinguishing user-initiated manual search from        
  automatic/inferred tag search at the plugin level:                                               
                                                                                                   
  Here are two backwards-compatible ways we could implement this without breaking existing plugins:
                                                                                                   
  #### Option 1: Optional Keyword Argument + Introspection (Recommended)                           
                                                                                                   
  We can update the core plugin runner to pass an optional keyword argument (e.g.,  user_initiated:
  bool = False ) to  candidates  and  item_candidates .                                            
                                                                                                   
  To prevent breaking third-party or older plugins that don't accept this argument, the core hook  
  caller can inspect the signature of the plugin's  candidates  method:                            
                                                                                                   
    import inspect                                                                                 
                                                                                                   
    # Inside beets core hook dispatch:                                                             
    sig = inspect.signature(plugin.candidates)                                                     
    kwargs = {}                                                                                    
    if 'user_initiated' in sig.parameters:                                                         
        kwargs['user_initiated'] = is_manual_search  # determined by importer state                
                                                                                                   
    plugin.candidates(items, artist, album, va_likely, **kwargs)                                   
                                                                                                   
  This allows the Tidal plugin (and any future plugins) to opt-in by simply adding  user_initiated:
  bool = False  to their signatures. When  True  (e.g., user manually entered search terms), the   
  plugin can trust the input and skip redundant fallback queries. When  False  (automatic imports),
  it can execute the fallback multi-query logic.                                                   
                                                                                                   
  #### Option 2: Add a Context Object                                                              
                                                                                                   
  Instead of changing method signatures, we could pass an optional context dictionary or object    
  containing query metadata:                                                                       
                                                                                                   
    def candidates(                                                                                
        self, items: Sequence[Item], artist: str, album: str, va_likely: bool, context:            
  QueryContext | None = None                                                                       
    ) -> Iterable[AlbumInfo]:                                                                      
                                                                                                   
  This  QueryContext  could hold flags like  is_manual ,  original_tags , or search constraints,   
  providing a clean way to scale metadata search features in the future without continually        
  modifying method signatures.                                                                     
                                                                                                   
  Let me know what you and @snejus think is the best path forward here!"

semohr:
> Here are two backwards-compatible ways we could implement this without breaking existing plugins:

We can ask an LLM for ideas ourselves... Please don't paste LLM-generated design suggestions into the review and present them as your proposals. It's not particularly helpful to the discussion and doesn't move the design forward in any meaningful way unless you've evaluated and argued for the approach yourself.

More importantly, none of this should be implemented in this PR. These changes are well outside the scope of the issue, and the proper integration and design need considerably more thought and discussion first.

PR #6811, opened 2026-06-30
tests: remove unused/redundant logic from tests

snejus:
- This change simplifies the test architecture after full coverage exposed dead paths. Redundant helpers, fixtures, fallback branches, and defensive test-only code were removed across `test/*` and `beets/test/fixtures.py`.

- Biggest structural change is in `test/plugins/test_smartplaylist.py`, where repeated temporary playlist directory setup was pulled into a small shared `PlaylistDirMixin`. This keeps filesystem setup in one place and makes the affected tests more uniform.

- Several tests now rely on direct assertions instead of `try`/`except`, manual cleanup, or skipped paths. That makes failures more explicit and reduces hidden control flow in the suite.

- High-level impact: no production architecture changes, but the test suite becomes smaller, stricter, and easier to reason about. Reviewers can read this as a cleanup pass that removes unused test scaffolding while preserving existing behavior.

### DIFF
--- a/test/plugins/test_fetchart.py
+++ b/test/plugins/test_fetchart.py
@@ -25,13 +25,12 @@ class FetchartCliTest(IOMixin, PluginTestCase):
         with open(util.syspath(self.cover_path(ext))) as f:
             assert f.read() == "IMAGE"
 
-    def hide_file_windows(self, ext="jpg"):
-        hidden_mask = 2
-        success = ctypes.windll.kernel32.SetFileAttributesW(
-            self.cover_path(ext), hidden_mask
-        )
-        if not success:
-            self.skipTest("unable to set file attributes")
+    def hide_file_windows(self, path: bytes) -> None:
+        if sys.platform == "win32":
+            hidden_mask = 2
+            assert ctypes.windll.kernel32.SetFileAttributesW(
+                util.syspath(path), hidden_mask
+            )
 
     def test_set_art_from_folder(self):
         self.touch(b"c\xc3\xb6ver.jpg", dir_=self.album.path, content="IMAGE")
@@ -62,9 +61,8 @@ class FetchartCliTest(IOMixin, PluginTestCase):
         self.check_cover_is_stored()
 
     def test_filesystem_does_not_pick_up_hidden_file(self):
-        self.touch(b".cover.jpg", dir_=self.album.path, content="IMAGE")
-        if sys.platform == "win32":
-            self.hide_file_windows()
+        path = self.touch(b".cover.jpg", dir_=self.album.path, content="IMAGE")
+        self.hide_file_windows(path)
         self.config["ignore"] = []  # By default, ignore includes '.*'.
         self.config["ignore_hidden"] = True
         self.run_command("fetchart")
@@ -79,9 +77,8 @@ class FetchartCliTest(IOMixin, PluginTestCase):
         self.check_cover_is_stored()
 
     def test_filesystem_picks_up_hidden_file(self):
-        self.touch(b".cover.jpg", dir_=self.album.path, content="IMAGE")
-        if sys.platform == "win32":
-            self.hide_file_windows()
+        path = self.touch(b".cover.jpg", dir_=self.album.path, content="IMAGE")
+        self.hide_file_windows(path)
         self.config["ignore"] = []  # By default, ignore includes '.*'.
         self.config["ignore_hidden"] = False
         self.run_command("fetchart")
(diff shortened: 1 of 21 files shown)


beets/importer/tasks.py
```diff
@@ -1156,44 +1156,61 @@ def read_item(self, path: util.PathBytes) -> library.Item | None:
...
     containing any media files is an album.
     """
-    collapse_paths: list[util.PathBytes] = []
-    collapse_items: list[util.PathBytes] = []
+    collapse_paths: list[AnyStr] = []
+    collapse_items: list[AnyStr] = []
     collapse_pat = None
 
-    ignore: list[str] = config["ignore"].as_str_seq()
+    _ignore = config["ignore"].as_str_seq()
```
semohr: Couldn't we use the `confuse.Filename` template here?

beets/importer/tasks.py
```diff
@@ -1156,44 +1156,61 @@ def read_item(self, path: util.PathBytes) -> library.Item | None:
...
+    ignore: list[AnyStr]
+    if isinstance(path, str):
+        ignore = _ignore
+    else:
+        ignore = list(map(os.fsencode, _ignore))
     ignore_hidden: bool = config["ignore_hidden"].get(bool)
 
+    patterns = (
```
semohr: Im wondering if it would make sense to convert the path early in this function to only have one branch here instead of doing both `bytes` and `str` branches separately 🤔 

This function is pretty complicated as is already.

beetsplug/fetchart.py
```diff
@@ -1058,7 +1058,7 @@ def get(
 
             # Find all files that look like images in the directory.
             images = []
-            ignore = config["ignore"].as_str_seq()
+            ignore = list(map(os.fsencode, config["ignore"].as_str_seq()))
```
semohr: Maybe it would make sense to unify this behaviour witha template or a helper method on the config class?

see https://github.com/beetbox/beets/pull/6811/changes#r3528504676

test/plugins/test_fetchart.py
```diff
@@ -25,13 +25,12 @@ def check_cover_is_stored(self, ext: str = "jpg"):
...
-            self.cover_path(ext), hidden_mask
-        )
-        if not success:
-            self.skipTest("unable to set file attributes")
+    def hide_file_windows(self, path: bytes) -> None:
+        if sys.platform == "win32":
+            hidden_mask = 2
+            assert ctypes.windll.kernel32.SetFileAttributesW(
+                util.syspath(path), hidden_mask
+            )
```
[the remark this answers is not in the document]
snejus: No, we'd rather see it fail and fix it. Go hide in cave grug, useless.

beets/importer/tasks.py
snejus: Apologies, I forgot to adjust the base of this PR. This change is part of #6806.

snejus: I will respond here but make adjustments under #6806. 

> Couldn't we use the `confuse.Filename` template here?

See `ignore` docs:

> ### ignore
> A list of glob patterns specifying file and directory names to be ignored when importing. By default, this consists of ``.*``, ``*~``, ``System Volume Information``, ``lost+found`` (i.e., beets ignores Unix-style hidden files, backup files, and directories that appears at the root of some Linux and Windows filesystems).

Not sure whether we'd win anything by using `confuse.Filename` template here.

We also need to either keep these as string or convert them to `bytes`:

```py
    _ignore = config["ignore"].as_str_seq()
    ignore: list[AnyStr]
    if isinstance(path, str):
        ignore = _ignore
    else:
        ignore = list(map(os.fsencode, _ignore))
```

beets/importer/tasks.py
snejus: I agree this function is already pretty complicated.

I think the catch is that `albums_in_dir` currently preserves the input path type: callers passing `bytes` get `bytes` paths back, and callers passing `str` get `str` paths back. If we convert the path early, we either change that behavior or still need to convert the yielded roots/items back later, which may make the function less clear.

So my preference would be to keep this small type split in #6806, and avoid expanding the refactor unless we want to change the function contract separately.

Note that once we've migrated paths, we will only support `Path` here!

beetsplug/fetchart.py
snejus: There are only two places where `"ignore"` is needed. I think it's fine to have this much of duplication :)

beets/importer/tasks.py
semohr: Ah it is a glob pattern :+1:

semohr (approved):

PR #6812, opened 2026-07-01
test(convert): cover never_convert_lossy_files in playlist entry tests

Kidus5168:
## Why

The `convert` plugin's `--playlist` option generates an m3u8 playlist of the converted output files. When `never_convert_lossy_files=True`, lossy files (e.g. OGG) are skipped — they are copied rather than transcoded — so their output path keeps the original extension. However, this scenario had no test coverage: the existing `test_playlist_entry` parametrize cases only covered `no_convert` as the mechanism for skipping transcoding, leaving `never_convert_lossy_files` untested.

Without a test for this case, a regression could silently cause the playlist to list the wrong extension for files that were not transcoded due to this option.

## What

Added a `never_convert_lossy` parameter to `test_playlist_entry` and a new parametrize case (`id="never-convert-lossy"`) that:
- sets `never_convert_lossy_files=True` in config
- runs `convert --playlist` on a lossy (OGG) item targeting MP3
- asserts the playlist entry keeps `.ogg` (the original extension), not `.mp3`

## Test output

```
test/plugins/test_convert.py::TestConvertCli::test_playlist PASSED                          [1/7]
test/plugins/test_convert.py::TestConvertCli::test_playlist_pretend PASSED                  [2/7]
test/plugins/test_convert.py::TestConvertCli::test_playlist_entry[config-format] PASSED     [3/7]
test/plugins/test_convert.py::TestConvertCli::test_playlist_entry[cli-format] PASSED        [4/7]
test/plugins/test_convert.py::TestConvertCli::test_playlist_entry[not-transcoded] PASSED    [5/7]
test/plugins/test_convert.py::TestConvertCli::test_playlist_entry[keep-new] PASSED          [6/7]
test/plugins/test_convert.py::TestConvertCli::test_playlist_entry[never-convert-lossy] PASSED [7/7]
7 passed in 0.28s
```

## Checklist

- [x] Tests added
- [x] All tests passing
- [x] Follows project style guide (ruff lint + format pass)
- [x] No breaking changes


test/plugins/test_convert.py
```diff
@@ -254,17 +254,28 @@ def test_force_overrides(self, config_overrides):
...
         ],
     )
-    def test_playlist_entry(self, args, no_convert, expected_entry):
+    def test_playlist_entry(
+        self, args, no_convert, never_convert_lossy, expected_entry
+    ):
         if no_convert:
             self.config["convert"]["no_convert"] = no_convert
+        if never_convert_lossy:
+            self.config["convert"]["never_convert_lossy_files"] = True
```
snejus, line 278: Remove conditionals from both of these lines and set the values unconditionally

snejus (changes requested):

PR #6815, opened 2026-07-02
Add `fetch_for_asis` setting to FetchArt plugin

overhacked:
Add `fetch_for_asis` setting to FetchArt plugin

## Description

Enable fetching art for imports even when as-is is selected as the metadata source. Allows for the case when files with good metadata but without album art are being imported (e.g. digital download store).

## To Do

- [x] Documentation.
- [x] Changelog.
- [x] Tests.

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -1318,6 +1318,7 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
         self.config.add(
             {
                 "auto": True,
+                "fetch_for_asis": False,
                 "minwidth": 0,
                 "maxwidth": 0,
                 "quality": 0,
@@ -1464,8 +1465,9 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
                 # Album already has art (probably a re-import); skip it.
                 return
             if task.choice_flag == importer.Action.ASIS:
-                # For as-is imports, don't search Web sources for art.
-                local = True
+                # For as-is imports, don't search Web sources for art,
+                # unless fetch_for_asis is set
+                local = not self.config["fetch_for_asis"]
             elif task.choice_flag in (
                 importer.Action.APPLY,
                 importer.Action.RETAG,
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -1439,6 +1439,10 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
         except UnknownPairError as e:
             raise UserError(e)
 
+    @cached_property
+    def fetch_for_asis(self) -> bool:
+        return self.config["fetch_for_asis"].get(bool)
+
     @staticmethod
     def _is_source_file_removal_enabled() -> bool:
         return config["import"]["delete"].get(bool) or config["import"][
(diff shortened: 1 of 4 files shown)


test/plugins/test_fetchart.py
```diff
@@ -3,16 +3,63 @@
...
 
-    def setUp(self):
-        super().setUp()
+    def setup_beets(self):
+        super().setup_beets()
+
+        self.prepare_album_for_import(1)
+        self.setup_importer()
+
+        self.art_patcher = mock.patch.object(
```
snejus: I'm not a big fan of _implementation_ tests as they tend to break after refactoring. Can we test the _expected behaviour_ instead?

beetsplug/fetchart.py
```diff
@@ -1464,8 +1465,9 @@ def fetch_art(self, session: ImportSession, task: ImportTask) -> None:
                 # Album already has art (probably a re-import); skip it.
                 return
             if task.choice_flag == importer.Action.ASIS:
-                # For as-is imports, don't search Web sources for art.
-                local = True
+                # For as-is imports, don't search Web sources for art,
+                # unless fetch_for_asis is set
+                local = not self.config["fetch_for_asis"]
```
snejus:
```suggestion
                local = not self.config["fetch_for_asis"].get(bool)
```

snejus: Also it may be worth defining a `@cached_property` for this (see `beetsplug/convert.py`, for example).

test/plugins/test_fetchart.py
```diff
@@ -31,7 +78,7 @@ def hide_file_windows(self, ext="jpg"):
             self.cover_path(ext), hidden_mask
         )
         if not success:
-            self.skipTest("unable to set file attributes")
+            pytest.skip("unable to set file attributes")
```
snejus: Best to undo this and leave it for a separate, more general refactor.

snejus (changes requested):

test/plugins/test_fetchart.py
overhacked: AOK to revert this, but the motivation was to have all the test classes in `test_fetchart.py` use the same framework, since I need to use the pytest `*Helper` class hierarchy for the importer tests. Would you rather I leave `TestFetchartCli` as it is?

beetsplug/fetchart.py
overhacked: Rather than converting all the config settings of `FetchArtPlugin` to `@cached_property`, I made `fetch_for_asis` another instance variable set in `__init__` like all the other config properties currently are handled. If you'd rather I convert the whole class to use `@cached_property`, let me know. I'd be glad to, but I didn't want to land a big refactor commit in this review without checking first.

test/plugins/test_fetchart.py
overhacked: :100: I can implement this based on the expected art output.

beetsplug/fetchart.py
overhacked: Aaaand, as soon as I switch it to an instance variable, the unit test fails because it sets the config rather than the instance variable. Now I see why convert uses `@cached_property`! I'll make `fetch_for_asis` a cached property, and do let me know if you'd like me to convert all of `FetchArtPlugin`'s config values to `@cached_property`.

test/plugins/test_fetchart.py
snejus: Ah yes, that makes sense!

beetsplug/fetchart.py
snejus: No need, we can start with just this one for now.

snejus (approved):
Thanks!

snejus:
You just need to rebase now and to resolve conflicts.

overhacked:
@snejus rebased and ready!

PR #6816, opened 2026-07-02
lyrics: add lrcmux backend

f1nniboy:
## Description

This PR adds https://lrcmux.dev ([repo](https://github.com/f1nniboy/lrcmux)) as a new lyrics fetcher backend. lrcmux aggregates various different lyrics sources and exposes them via a single API, so it's perfect as a replacement for the currently defunct Musixmatch fetcher. Users can host their own instance if they prefer not to share their queries with a central API.

## To Do

### DIFF
--- a/test/plugins/test_lyrics.py
+++ b/test/plugins/test_lyrics.py
@@ -431,6 +431,11 @@ class TestLyricsSources(LyricsBackendTest):
         }
         requests_mock.get(lyrics.Google.SEARCH_URL, json=data)
 
+    @pytest.fixture(autouse=True)
+    def _set_lrcmux_sources(self, lyrics_plugin, backend_name):
+        if backend_name == "lrcmux":
+            lyrics_plugin.config["lrcmux"]["sources"] = ["ytmusic"]
+
     def test_backend_source(
         self, monkeypatch, lyrics_plugin, lyrics_page: LyricsPage
     ):
@@ -679,6 +684,58 @@ class TestLRCLibLyrics(LyricsBackendTest):
         else:
             assert lyrics
             assert lyrics.text == expected_lyrics
+
+
+class TestLRCMuxLyrics(LyricsBackendTest):
+    SYNCED = "[00:00.00] synced"
+    PLAIN = "plain"
+
+    @pytest.fixture(scope="class")
+    def backend_name(self):
+        return "lrcmux"
+
+    @pytest.mark.parametrize(
+        "plugin_config, mocked_text, expected_format, expected_level",
+        [
+            pytest.param({"synced": True}, SYNCED, "lrc", "line", id="synced"),
+            pytest.param({"synced": False}, PLAIN, "txt", "none", id="plain"),
+            pytest.param(
+                {"synced": True, "lrcmux": {"sources": ["ytmusic"]}},
+                SYNCED,
+                "lrc",
+                "line",
+                id="synced-sources",
+            ),
+        ],
+    )
+    def test_fetch_lyrics(
+        self,
+        backend,
+        requests_mock,
+        mocked_text,
+        expected_format,
+        expected_level,
+        plugin_config,
+    ):
+        requests_mock.get(backend.url, text=mocked_text)
+        result = backend.fetch("la", "la", "la", 0)
+        assert result
+        assert result.text == mocked_text
+        assert result.backend == "lrcmux"
+        assert f"format={expected_format}" in result.url
+        assert f"level={expected_level}" in result.url
+        if sources := (plugin_config.get("lrcmux") or {}).get("sources"):
+            assert f"sources={','.join(sources)}" in result.url
+
+    @pytest.mark.parametrize("plugin_config", [{}])
+    def test_not_found(self, backend, requests_mock):
+        requests_mock.get(backend.url, status_code=HTTPStatus.NOT_FOUND)
+        assert backend.fetch("la", "la", "", 0) is None
+
+    @pytest.mark.parametrize("plugin_config", [{}])
+    def test_empty_response(self, backend, requests_mock):
+        requests_mock.get(backend.url, text="")
+        assert backend.fetch("la", "la", "", 0) is None
 
 
 @pytest.mark.requires_import("langdetect")
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -897,6 +897,42 @@ class Translator(LyricsRequestHandler):
         return lyrics
 
 
+class LRCMux(Backend):
+    """Fetch lyrics from the lrcmux API."""
+
+    DEFAULT_URL = "https://api.lrcmux.dev"
+
+    @cached_property
+    def url(self) -> str:
+        base = self.config["lrcmux"]["url"].get(str)
+        return f"{base.rstrip('/')}/get"
+
+    def fetch(
+        self, artist: str, title: str, album: str, length: int
+    ) -> Lyrics | None:
+        synced = self.config["synced"].get(bool)
+        sources = self.config["lrcmux"]["sources"].as_str_seq()
+
+        params: JSONDict = {
+            "artist": artist,
+            "title": title,
+            "duration": int(length),
+            "format": "lrc" if synced else "txt",
+            "level": "line" if synced else "none",
+        }
+        if album:
+            params["album"] = album
+        if sources:
+            params["sources"] = ",".join(sources)
+
+        full_url = self.format_url(self.url, params)
+        with suppress(HTTPNotFoundError):
+            if text := self.get_text(self.url, params=params).strip():
+                return Lyrics(text, self.__class__.name, full_url)
+
+        return None
+
+
 @dataclass
 class RestFiles:
     # The content for the base index.rst generated in ReST mode.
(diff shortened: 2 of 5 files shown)


semohr:
Hi!

It would have been nice and transparent to disclose that you're also the author of `lrcmux`.

Since the project is still very new (it's only a few days old), I would prefer to postpone both this PR and any upstream integration into beets until the project has had more time to mature. At this stage, it's difficult to assess its long-term viability and whether it will see continued development and maintenance. We tend to be fairly conservative about adding new integrations because we will be responsible for maintaining them going forward.

This isn't a criticism of you or your project. I'm generally in favor of adding useful new functionality, but as things stand today, integrating `lrcmux` into upstream feels a bit premature. We wouldd be happy to revisit this once the project has had some time to settle in tho.

Serene-Arc:
I agree with @semohr and would also add that adding `lrcmux` as a source for lyrics would, IMO, only be a good decision if the project is not only mature, but popular as an aggregate service among beets users. Even if it remains updated and matures, it's still a service that adds an intermediary layer to other APIs.

I can see the usefulness as a user that doesn't want to go to a number of sites to get lyrics, but we provide a tool ourselves. I think there would have to be a compelling case to use `lrcmux` rather than just implementing the same API handling that it possesses ourselves. Otherwise it feels like adding an unnecessary intermediary to the process. There are five providers that `lrcmux` uses and we already actually integrate two of the five used, plus some others beside.

I think there would have to be compelling case to use this rather than simply implementing the Kugou, NetEase, and YouTube Music interfaces ourselves. Then at least we would own the code and be able to test and manage it thoroughly.

f1nniboy:
Fair points, and I understand the "we should just do it ourselves instead of relying on a central point of failure". But:

Of the common providers (Genius, LRCLIB, Musixmatch), only LRCLIB actually works out of the box right now. The bundled Genius API key is rate-limited and quite over its daily limit, one issue with using a central API key for such a big project. Musixmatch+Tekstowo have been off by default. And the Google search backend needs an API key supplied by the user. So realistically a regular setup fetches from only LRCLIB, which definitely has alright track coverage, but nothing can beat fetching from many lyrics providers at once (and users actually need to submit lyrics).

On the "we can just implement YT Music/Kugou/etc. ourselves": I guess thats true, but these undocumented APIs will change over time and each breakage ends up being a beets issue to fix. That's kind of why I built lrcmux, so that not every music tool/app has to independently maintain the same set of scrapers. Fix once, and every user of the API benefits. It's really the tradeoff between directly owning the scraper code and maintenance burden.

On the "unnecessary intermediary" thing, there might actually be some benefits over doing it yourselves directly: shared cache across users (for not unnecessarily querying the upstream providers and providing fast results) and the operational side of it (dealing with blocks, scraping, rotation, etc.) already being handled. Musixmatch is blocked today, and e.g. even if you switched to the "lrcmux" approach for Musixmatch (not mine, to be fair), users would still get rate-limited regularly on their own IP.

If you don't trust the public instance or just want to host it yourself, you can easily do so (albeit with the drawback of encountering provider rate limits).

The PR would add lrcmux as one optional source, not the default. Users who want it could enable it, and the others can leave it disabled. I'm happy to maintain it here and keep it working if the API changes, that's obviously on me as the PR (and project) author.

(also, sorry about not disclosing that I'm the author 🫠)

arsaboo:
I see both sides of the argument, and personally, I'm comfortable adding another lyrics provider if it helps expand what we support. At the moment, lrclib is essentially our only lyrics source. Any option that strengthens our coverage—particularly for Bollywood tracks—has my support. On top of that, it's a small pull request, so if LRCMux proves unreliable or fails to gain traction, we can drop it with little trouble.

snejus:
I agree with @arsaboo and @f1nniboy - we currently only have one reliable source left - `lrclib`. And even those requests are slow nowadays. I fetch lyrics a lot, so would be more than happy to try out a new provider, even if it's new and not yet proven.

Just rebase and fix the conflict in the changelog.

snejus:
Also add a test case to `test/plugins/lyrics_pages.py` to test fetching from `lrcmux`!

f1nniboy:
> @snejus Also add a test case to `test/plugins/lyrics_pages.py` to test fetching from `lrcmux`!

I'm not sure about this, as the lyrics results are non-deterministic (based on which provider responds first, at the same sync level) and as such the API response may change once the cached entry expires. So the test may fail in the future, even though the API still works fine. What do you think?

snejus:
Another question we had: how does your API scale? Consider that once this is integrated into beets you will get many, many more requests compared to what you get now, I think.

snejus:
> I'm not sure about this, as the lyrics results are non-deterministic (based on which provider responds first, at the same sync level) and as such the API response may change once the cached entry expires. So the test may fail in the future, even though the API still works fine. What do you think?

Have a look at `TestLyricsSources`/`LyricsBackendTest`/`LyricsPluginMixin` in detail and you will find that we parametrize `LyricsPlugin` with the expected lyrics backend and set it in the plugin configuration. Thus, only `lrcmux` will be queried for that new test case. 

Just make sure that `lrcmux` backend is resolved correctly in `LyricsPage.source` and `LyricsPage.backend`.

f1nniboy:
> Another question we had: how does your API scale? Consider that once this is integrated into beets you will get many, many more requests compared to what you get now, I think.

Well, I haven't seen heavy API load in practice yet, but in theory it should scale pretty well. The entire API is stateless (it only needs Redis, maybe a DB in the future?), so I can just add more servers if the API itself ever struggles.

Proxy costs are negligible right now, and even projecting way up (10-20 requests/s) they're still small. I'm keeping an eye on traffic levels to scale up as needed.

And as popularity grows, the most popular tracks should be able to be served straight from cache.

> Have a look at `TestLyricsSources`/`LyricsBackendTest`/`LyricsPluginMixin` in detail and you will find that we parametrize `LyricsPlugin` with the expected lyrics backend and set it in the plugin configuration. Thus, only `lrcmux` will be queried for that new test case.
> 
> Just make sure that `lrcmux` backend is resolved correctly in `LyricsPage.source` and `LyricsPage.backend`.

Yeah, but the thing is, the lrcmux API itself will return different providers' results (each provider has slightly different lyrics for that song) depending on latency if it has to query providers again (after the cache expires), so the expected .lrc response in the test may not match what the API actually returns in the future. I wanted to avoid having a hardcoded provider order to optimize for speed.

snejus:
> Yeah, but the thing is, the lrcmux API itself will return different providers' results (each provider has slightly different lyrics for that song) depending on latency if it has to query providers again (after the cache expires), so the expected .lrc response in the test may not match what the API actually returns in the future. I wanted to avoid having a hardcoded provider order to optimize for speed.

Ah, that's not ideal. Did you see the logic I added for `lrclib`?

f1nniboy:
> Ah, that's not ideal. Did you see the logic I added for `lrclib`?

You mean fetching LRCLIB lyrics by ID? That would work for the test, but in the actual search() I would still want to keep the title+artist+duration+album query parameters, which wouldn't work with the test. I could add a `sources` parameter of some sort to the API to control which providers to use, but then I would also have to add `lrcmux_sources` as a config option. Should I do that?

snejus:
> but then I would also have to add `lrcmux_sources` as a config option. Should I do that?

Maybe no need in such case. We still somehow want to prove that 'hey, this provider still works' in the test. The worry I have is more related to using the plugin. For example, if I fetch synchronised lyrics from `lrcmux`, and then refetch them later, does it mean that I may end up with different lyrics later?

f1nniboy:
Well, it depends. Right now I cache fetched lyrics for up to a year (maybe a permanent DB in the future), and after that, if you try to fetch them again, it will go through the entire provider fetching thing again, where the fastest (and highest sync level) provider wins. Providers could also be down temporarily/rate-limited/failed/whatever when you refetch. So it could be Musixmatch again, but you may also get lyrics from Kugou, etc. These lyrics shouldn't be *that* different from eachother, and you shouldn't get wrong lyrics either way. The only way to get the same lyrics again is setting the `sources` parameter to one provider only.

snejus:
Sounds good to me!

snejus:
And... I changed my mind regarding the test - just realised that `lrcmux.sources` is indeed a configuration option. Please do add an autouse fixture that sets one specific source in `lrcmux.sources` when `backend == "lrcmux"` and add a test case please! 

Note that these tests are run only when `lyrics` plugin contents have changed, so you won't get ddosed here 😅

test/plugins/test_lyrics.py
```diff
@@ -458,6 +464,13 @@ def test_backend_source(
         assert actual_lyrics.text == expected_lyrics.text
         assert actual_lyrics == expected_lyrics
 
+    def test_lrcmux_sources(self, backend_name):
+        sources = self.config[self.plugin]["lrcmux"]["sources"].as_str_seq()
+        if backend_name == "lrcmux":
+            assert sources == ["ytmusic"]
+        else:
+            assert sources == []
+
```
snejus: No need to test this

test/plugins/test_lyrics.py
```diff
@@ -431,6 +431,12 @@ def _patch_google_search(self, requests_mock, lyrics_page):
         }
         requests_mock.get(lyrics.Google.SEARCH_URL, json=data)
 
+    @pytest.fixture(autouse=True)
+    def _set_lrcmux_sources(self, lyrics_plugin, backend_name):
+        del lyrics_plugin
+        if backend_name == "lrcmux":
+            self.config[self.plugin]["lrcmux"]["sources"] = ["ytmusic"]
```
snejus:
```suggestion
        if backend_name == "lrcmux":
            lyrics_plugin.config["lrcmux"]["sources"] = ["ytmusic"]
```

snejus (changes requested):

snejus (approved):
Excited to try it out!

PR #6820, opened 2026-07-07
typing: add types to beets.dbcore

snejus:
### What changed

- This PR is mainly a typing pass over `beets.dbcore`, especially in `db.py`, `query.py`, `sort.py`, and `types.py`.
- The main architecture change is in `db.py`: model materialization is now centralized in `Database._make_model()`, while `Results` only handles lazy row iteration and uses a materializer callback.
- Reloading is also cleaner now: `Model.get_fresh_from_db()` goes through `Database._reload()`, which keeps database lookup logic in the database layer and preserves the model's concrete type.
- Along the way, a few runtime mismatches found by typing were corrected instead of just adjusting annotations.

### Why

- The goal is to make the typed API match the real behavior of `dbcore`.
- This improves separation of responsibilities: querying and result iteration stay in `Results`, while row-to-model construction and reload behavior live in `Database`.

### High-level impact

- `dbcore` has a clearer internal structure and a more explicit API surface.
- Type information is more accurate across queries, sorting, models, and field types, which should make future refactors safer.
- A few small correctness fixes are included:
  - `DurationType.format()` now always returns `str`.
  - `PathType.from_sql()` only expands paths when the normalized value is actually `bytes`.
  - Several `__eq__` implementations now use explicit `isinstance(...)` checks.
  - Reload behavior now has regression coverage in `test/dbcore/test_db.py`.
- Overall, this is a low-risk cleanup: mostly typing and API-shape improvements, with a few small behavior fixes discovered during the work.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -52,6 +52,8 @@ if TYPE_CHECKING:
     from sqlite3 import Connection
     from types import TracebackType
 
+    from beets.util import PathLike
+
     from .query import FieldQueryType, Query, SQLiteType
     from .sort import FieldSort, Sort
 
@@ -74,7 +76,7 @@ class DBAccessError(Exception):
 class DBCustomFunctionError(Exception):
     """A sqlite function registered by beets failed."""
 
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__(
             "beets defined SQLite function failed; "
             "see the other errors above for details"
@@ -183,7 +185,7 @@ class LazyDict(UserDict[str, Any]):
         if key in self._raw:
             del self._raw[key]
 
-    def __contains__(self, key: Any) -> bool:
+    def __contains__(self, key: object) -> bool:
         """Determine whether `key` is an attribute on this object."""
         return key in self.data or key in self._raw
 
@@ -294,7 +296,7 @@ class Model(ABC, Generic[D]):
     """
 
     @cached_classproperty
-    def _relation(cls):
+    def _relation(cls) -> type[Model]:
         """The model that this model is closely related to."""
         return cls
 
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yaml
```diff
@@ -55,6 +55,14 @@ jobs:
       - name: Install system dependencies on Ubuntu
         if: matrix.platform == 'ubuntu-latest'
         run: |
+          # Work around transient signature issues in preconfigured runner repos
```
semohr: Still relevant?

beets/dbcore/db.py
```diff
@@ -330,13 +334,16 @@ def db(self) -> D:
     def get_fresh_from_db(self) -> Self:
         """Load this object from the database."""
         model_cls = self.__class__
+        if self.id is None:
+            raise ValueError(f"{model_cls.__name__} has no id")
+
         if obj := self.db._get(model_cls, self.id):
```
semohr: Can we do something like this instead? I think we should raise a not found error instead of a value error.
```suggestion
        if (self.id is not None) and (obj := self.db._get(model_cls, self.id):
```

beets/dbcore/db.py
```diff
@@ -294,9 +298,9 @@ def _queries(cls) -> dict[str, FieldQueryType]:
     """
 
     @cached_classproperty
-    def _relation(cls):
+    def _relation(cls) -> type[Model]:
```
semohr, line 301: type[Self] is an issue here?

semohr (commented):
Looks good to me overall, thanks!

Could we make the mypy configuration stricter for the files touched by this PR? That would help keep us from accumulating type-checking drift again.

beets/dbcore/db.py
snejus: Mypy thinks that `cls` has the same type as `self`, so it always complains. `type[Self]` is a bit inaccurate as this can be any `Model`.

snejus:
> Could we make the mypy configuration stricter for the files touched by this PR? That would help keep us from accumulating type-checking drift again.

We already tried doing this in previous PRs 😅 I'd like to somehow make `ruff` check these files using `ANN` checker, but only fully typed ones.

.github/workflows/ci.yaml
snejus: Maybe not, will test

semohr (approved):

PR #6825, opened 2026-07-10
typing: type beets.library

snejus:
- This PR finishes a broad typing pass across `beets.library`, mainly in `beets/library/models.py` and `beets/library/library.py`, and tightens a few related plugin and dbcore interfaces.

- Architecturally, the change makes the `Library`, `Item`, `Album`, query parsing, and template helper layers agree on clearer contracts: nullable values are explicit, collection and query return types are defined, and cross-layer APIs now use more precise types.

- The main behavioral surface is still the same, but several weak spots are now modeled directly in the code, such as optional albums on items, optional art paths, library-backed template memoization, and query objects flowing through `Library._fetch()`.

- The PR also cleans up how template uniqueness helpers work by giving `_memotable` and related keys a stable shape, and by tightening the logic around album-aware formatting and disambiguation.

- A small architectural rename in `dbcore` changes `Database._fetch()` to `Database.get_results()`, which makes the lower-level database API clearer and avoids confusion with `Library._fetch()`.

- High-level impact: better static type safety, clearer boundaries between database, library, model, and templating code, and less ambiguity for future refactors without intentionally changing user-facing behavior.

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -1384,7 +1384,7 @@ class Database:
 
     # Querying.
 
-    def _fetch(
+    def _get_results(
         self,
         model_cls: type[AnyModel],
         query: Query | None = None,
@@ -1444,7 +1444,7 @@ class Database:
 
     def _get(self, model_cls: type[AnyModel], id_: int) -> AnyModel | None:
         """Get a Model object by its id or None if the id does not exist."""
-        return self._fetch(model_cls, MatchQuery("id", id_)).get()
+        return self._get_results(model_cls, MatchQuery("id", id_)).get()
 
 
 class Index(NamedTuple):
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -15,6 +15,7 @@ from collections.abc import Mapping
 from contextlib import contextmanager
 from dataclasses import dataclass
 from functools import cached_property
+from pathlib import Path
 from sqlite3 import Connection, sqlite_version_info
 from typing import (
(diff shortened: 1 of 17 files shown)


beets/plugins.py
```diff
@@ -57,6 +56,7 @@
     "album_removed",
     "albuminfo_received",
     "album_matched",
+    "art_set",
```
semohr, line 59: The `art_set` event is missing documentation in the `events.rst`. Out of scope here, you mind creating an issue for that?

beets/dbcore/db.py
```diff
@@ -1382,7 +1384,7 @@ def record_migration(self, name: str, table: str) -> None:
 
     # Querying.
 
-    def _fetch(
```
semohr, line 1385: Im not too sure if we should rename this. Currently `_fetch` is only used within the Library class. It seem like it was intended as "private" method. Do we intend to use the method externally (from outside the Library class)?

snejus: I renamed it because typing wasn't happy that it shared the same definition with another `_fetch` method. Shall I name it `_get_results` instead?

semohr: I would vote yes to keep the intend of it being a private method.

beets/plugins.py
snejus: Done: https://github.com/beetbox/beets/issues/6845

semohr (approved):
LGTM, thanks!

PR #6826, opened 2026-07-12
test: fix requests_mock setup in test_art.py

overhacked:
## Description

`requests_mock` was configured with a fallback matcher to return an exception on any request. The intent was to have tests fail if any requests are attempted in tests without matching mocks. This is the default behavior of `requests_mock`, so no fallback matcher is necessary. The side-effect of the fallback matcher, removed by this commit, was that 1) the `NoMockAddress` exception was raised to the function under test rather than to the test harness and 2) the `NoMockAddress` exception was not able to be constructed because the `exc` parameter to `requests_mock.register_uri` does not support exceptions that require arguments. This behavior is visible in the following log snippet, taken from an incorrectly-passing test:

```
DEBUG    beets.fetchart:logging.py:163 downloading image: https://images.amazon.com/images/P/xxxx.01.LZZZZZZZ.jpg
DEBUG    requests_mock.adapter:adapter.py:256 GET https://images.amazon.com/images/P/xxxx.01.LZZZZZZZ.jpg 200
DEBUG    beets.fetchart:logging.py:163 not a supported image: text/html
DEBUG    beets.fetchart:logging.py:163 downloading image: https://images.amazon.com/images/P/xxxx.02.LZZZZZZZ.jpg
DEBUG    beets.fetchart:logging.py:163 error fetching art: NoMockAddress.__init__() missing 1 required positional argument: 'request'
```

The test that emitted these logs passed because the `FetchArt` plugin is written to handle exceptions raised by art sources, skipping and logging an error. For the exception to cause a failing test, the default behavior of `requests_mock` of raising an exception to the test harness when no matching mock requests can be found must be allowed.

## To Do

- [x] ~Documentation~
- [x] ~Changelog~ (no user-facing changes)
- [x] Tests (change to existing tests)

### DIFF
--- a/test/plugins/test_art.py
+++ b/test/plugins/test_art.py
@@ -10,8 +10,6 @@ from unittest.mock import patch
 
 import confuse
 import pytest
-from requests_mock import ANY as ANYREEQUEST
-from requests_mock.exceptions import NoMockAddress
 
 from beets import config, importer, logging, util
 from beets.autotag import AlbumInfo, AlbumMatch, Distance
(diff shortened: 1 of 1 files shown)


snejus (approved):

snejus:
Thanks!

PR #6828, opened 2026-07-12
Increase test verbosity

overhacked:
## Description

Change `config["verbose"]` from `1` to `2` so that DEBUG-level logging from plugin import tasks and event hooks are emitted. (See `beets.plugins._set_log_level_and_params`.)

`config["verbose"] == 2` does not appear to affect logging anywhere else in the codebase, while `config["verbose"] == 3` would enable output of messages logged via `BeetsLogger.extra_debug()`.

## To Do

- [x] ~Documentation~
- [x] ~Changelog~
- [x] Tests

### DIFF
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -122,7 +122,7 @@ class ConfigMixin:
         config.read(user=False, defaults=True)
 
         config["plugins"] = []
-        config["verbose"] = 1
+        config["verbose"] = 2
         config["ui"]["color"] = False
         config["threaded"] = False
         config["create_backup_before_migrations"] = False


snejus:
`extra_debug` is currently primarily used in `lastgenre` plugin for development - is there any value in enabling this logging level here?

overhacked:
> `extra_debug` is currently primarily used in `lastgenre` plugin for development - is there any value in enabling this logging level here?

I was wondering that myself, if maximum debug logging is the right way to go for tests, and I figured you would have a better sense of that. IMO, if `extra_debug()` is intended to be a restriction on user-facing console log spam and not "trace-level" logging then we should enable it for tests. If you agree, I'll re-push with `verbose=3`, or go ahead and amend and merge if you like! Thx!

snejus:
Thinking about the future, this is actually a good idea - verbosity=2 is what we'd like to have indeed!

snejus (approved):

semohr:
> > `extra_debug` is currently primarily used in `lastgenre` plugin for development - is there any value in enabling this logging level here?
> 
> I was wondering that myself, if maximum debug logging is the right way to go for tests, and I figured you would have a better sense of that. IMO, if `extra_debug()` is intended to be a restriction on user-facing console log spam and not "trace-level" logging then we should enable it for tests. If you agree, I'll re-push with `verbose=3`, or go ahead and amend and merge if you like! Thx!

Im working towards removing this extra debug level at some point :rofl: Feel free to have a look at  https://github.com/beetbox/beets/issues/6553 and https://github.com/beetbox/beets/pull/6773

PR #6829, opened 2026-07-12
Add Embedded Art support to `fetchart` plugin

### LINKED ISSUE #1130: fetchart: Extract embedded art

overhacked:
## Description

Add `embedded` source to extract embedded art from files. Add `skip_embedded` option to detect embedded art and skip fetching from other sources (if extracting isn't the desired end state).

If `skip_embedded=yes`, then files with embedded art are considered "done" and skipped by FetchArt. If the `embedded` source is enabled, then embedded artwork is always extracted. The two settings are independent of each other.

Fixes #1130.

## To Do

- [x] Documentation.
- [x] Changelog.
- [x] Tests.

### DIFF
--- a/beetsplug/fetchart.py
+++ b/beetsplug/fetchart.py
@@ -21,6 +21,7 @@ from beets.util import bytestring_path, get_temp_filename, sorted_walk, syspath
 from beets.util.artresizer import ArtResizer
 from beets.util.color import colorize
 from beets.util.config import UnknownPairError, sanitize_pairs
+from beetsplug._utils import art
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator, Sequence
@@ -771,11 +772,11 @@ class FanartTV(RemoteArtSource):
 
         matches = []
         # can there be more than one releasegroupid per response?
-        for mbid, art in data.get("albums", {}).items():
+        for mbid, images in data.get("albums", {}).items():
             # there might be more art referenced, e.g. cdart, and an albumcover
             # might not be present, even if the request was successful
-            if album.mb_releasegroupid == mbid and "albumcover" in art:
-                matches.extend(art["albumcover"])
+            if album.mb_releasegroupid == mbid and "albumcover" in images:
+                matches.extend(images["albumcover"])
             # can this actually occur?
             else:
                 self._log.debug(
@@ -1285,9 +1286,49 @@ class CoverArtUrl(RemoteArtSource):
             return
 
 
+class Embedded(LocalArtSource):
+    NAME = "Embedded"
+    ID = "embedded"
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+
+    def cleanup(self, candidate: Candidate) -> None:
+        if candidate.path:
+            try:
+                util.remove(path=candidate.path)
+            except util.FilesystemError as exc:
+                self._log.debug("error cleaning up tmp art: {}", exc)
+
+    def get(
+        self,
+        album: Album,
+        plugin: FetchArtPlugin,
+        paths: None | Sequence[bytes],
+    ) -> Iterator[Candidate]:
+        filename = get_temp_filename(__name__)
+        for item in album.items():
+            if extracted_path := art.extract(self._log, filename, item):
+                yield self._candidate(
+                    path=extracted_path, match=MetadataMatch.EXACT
+                )
+                break
+        else:
+            self._log.debug("embedded: art not found for {}", album)
+
+        try:
+            # Always remove tempfile because art.extract will have created a new
+            # file with the same basename but correct file extension added
+            util.remove(path=filename)
+        except util.FilesystemError as exc:
+            self._log.debug("error cleaning up tempfile: {}", exc)
+        return
+
+
 # All art sources. The order they will be tried in is specified by the config.
 ART_SOURCES: set[type[ArtSource]] = {
     FileSystem,
+    Embedded,
     CoverArtArchive,
     ITunesStore,
     AlbumArtOrg,
@@ -1319,6 +1360,7 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
             {
                 "auto": True,
                 "fetch_for_asis": False,
+                "skip_embedded": False,
                 "minwidth": 0,
                 "maxwidth": 0,
                 "quality": 0,
@@ -1443,6 +1485,10 @@ class FetchArtPlugin(plugins.BeetsPlugin, RequestMixin):
     def fetch_for_asis(self) -> bool:
         return self.config["fetch_for_asis"].get(bool)
 
+    @cached_property
+    def skip_embedded(self) -> bool:
+        return self.config["skip_embedded"].get(bool)
+
     @staticmethod
     def _is_source_file_removal_enabled() -> bool:
         return config["import"]["delete"].get(bool) or config["import"][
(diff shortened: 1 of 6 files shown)


overhacked:
Taking this to draft to do a little work to improve the test coverage.

overhacked:
@snejus, this is ready whenever you have time to take a look. There are some test coverage places I'd like to address, but when investigating I found that more test harness refactoring around when plugins read their configuration (e.g. `@cached_property` and `PluginMixin.preload_plugin` vs `.configure_plugin()`) would be necessary to get more coverage; also, the coverage gaps--while technically part of this PR's diff--already existed in the `fetchart` CLI test suite, so I'd be adding tests that don't test the `Embedded` source directly.

beetsplug/_utils/art.py
```diff
@@ -160,7 +160,6 @@ def extract(log, outpath, item):
     art = get_art(log, item)
     outpath = bytestring_path(outpath)
     if not art:
-        log.info("No album art present in {}, skipping.", item)
```
semohr, line 163: Any specific reason for removing the logs? Lets keep the git diff minimal.

beetsplug/_utils/art.py
```diff
@@ -170,9 +169,6 @@ def extract(log, outpath, item):
         return None
     outpath += bytestring_path(f".{ext}")
 
-    log.info(
-        "Extracting album art from: {} to: {}", item, displayable_path(outpath)
```
semohr, line 174: see https://github.com/beetbox/beets/pull/6829/changes#r3672527806

beetsplug/_utils/art.py
```diff
@@ -182,7 +178,13 @@ def extract_first(log, outpath, items):
     for item in items:
         real_path = extract(log, outpath, item)
         if real_path:
+            log.info(
```
semohr: see https://github.com/beetbox/beets/pull/6829/changes#r3672527806

beetsplug/fetchart.py
```diff
@@ -771,11 +772,11 @@ def get(
 
         matches = []
         # can there be more than one releasegroupid per response?
-        for mbid, art in data.get("albums", {}).items():
+        for mbid, images in data.get("albums", {}).items():
```
semohr, line 775: Why was this renamed from `art` to `images` ?

semohr (commented):
Thanks! Looks mostly good from my side 👍

beetsplug/fetchart.py
```diff
@@ -1554,6 +1599,17 @@ def func(lib: Library, opts, args) -> None:
         cmd.func = func
         return [cmd]
 
+    def album_has_art(self, album: Album):
```
snejus: Add a return type

docs/plugins/embedart.rst
```diff
@@ -22,6 +22,11 @@ be embedded. Art will be embedded after each album has its cover art set.
 
 This behavior can be disabled with the ``auto`` config option (see below).
 
+Extracting Art Automatically
+----------------------------
+
+See the ``embedded`` source of the :doc:`/plugins/fetchart`.
```
snejus: Refer to `embedded-art-source` directly

test/plugins/test_art.py
```diff
@@ -337,6 +337,31 @@ def test_is_candidate_fallback_os_error(
...
+    def settings(self) -> Settings:
+        return Settings(fallback=None)
+
+    @pytest.fixture
+    def source(self) -> fetchart.Embedded:
+        return fetchart.Embedded(logger, self.plugin.config)
+
+    def test_extract_embedded_art(self, source, settings):
+        for album in self.albums:
+            candidate = next(source.get(album, settings, None))
```
snejus, line 358: Do we need to check every album? Would it be enough to get the _first_ candidate instead?

snejus (changes requested):

test/plugins/test_fetchart.py
```diff
@@ -14,6 +16,16 @@
...
 
 
+@pytest.fixture(autouse=True)
+def disable_http_requests(requests_mock):
+    """Disable all outgoing requests and raise an error on HTTP
+
+    For test safety to ensure no changes cause tests to make
+    real, unmocked requests for cover art.
+    """
+    return
```
snejus, line 30: Do we actually want to make real requests here?

```suggestion
    yield
```

beetsplug/_utils/art.py
overhacked: From aac06ee:

> Two messages were emitted in the lowest-level helper functions, which
> meant that callers who expected a silent pass/fail result to validate
> whether art is contained in media files would spam the CLI with
> info-level messages. This commit moves the emission-site of the log
> messages one level up the call stack so that the messages are only
> emitted by a function that is called when the user requests that art is
> extracted and needs success/skip/failure logged to the CLI.

beetsplug/fetchart.py
overhacked: `_utils.art` is now imported, so this line triggers a variable-shadowing lint. I changed the loop variable name so as not to add to the number of linter warnings.

test/plugins/test_art.py
overhacked: This does get the first *candidate* for each album (`next()`) created in `setup_beets()`, but loops over all the albums to test embedded art extraction in all of the media formats. I was going to use pytest parameterization when I first wrote this, but it ended up being less complex code to just create mock albums for each format, IIRC.

test/plugins/test_fetchart.py
overhacked: There is no setup or teardown in this fixture, so `yield` is unnecessary, and it triggers the lint `No teardown in fixture... use 'return' instead of 'yield'`.

We don't want to make make real requests in `test_fetchart.py`, because all its tests are of the import and `fetchart` CLI logic of the plugin, so we want to fail a test if we've accidentally failed to mock something. The real art fetching tests are in `test_art.py`.

beetsplug/_utils/art.py
semohr: Thanks, did not check the commits. Nonetheless, please revert the changes here again. I get that it is currently spamy but the change is not in scope for this PR. We can decrease the level to debug if you have a strong opinion but we should not remove it completely.

We have planned to increase the default log level to `Warning` with #6773 or a following PR so this will likely be a none issue or can be evaluated again afterwards.

overhacked: Got it. I'll back the logging changes out. Just rebased and dropped the logging commit. Any thoughts on the other review comments?

beetsplug/fetchart.py
```diff
@@ -1285,9 +1286,49 @@ def get(
...
+        else:
+            self._log.debug("embedded: art not found for {}", album)
+
+        try:
+            # Always remove tempfile because art.extract will have created a new
+            # file with the same basename but correct file extension added
+            util.remove(path=filename)
+        except util.FilesystemError as exc:
+            self._log.debug("error cleaning up tempfile: {}", exc)
+        return
```
Serene-Arc: This cleanup will never actually run. Because you break out of the loop when you find a valid candidate, the generator isn't actually exhausted and it exits at the `yield` statement above. This should be in a `finally` block so it runs regardless. I added a failing test that shows this.

Serene-Arc (changes requested):

beetsplug/fetchart.py
overhacked: Got it! Didn't fully understand the mechanics of `yield` and `GeneratorExit` exceptions. It's wrapped in a `try: ... finally: ...` now.

test/plugins/test_art.py
snejus: Is there any need to test *every* given format? It doesn't seem like your logic varies depending on the format.

PR #6832, opened 2026-07-13
Fix crash when sorting by a nullable field missing on some items

### LINKED ISSUE #3461: Support sorting by specific attributes
### Use case

I'm trying to use beets to sort by item rating:

```
$ beet ls -f '$path' rating-
  File "/usr/lib/python-exec/python3.7/beet", line 11, in <module>
  File "/usr/lib/python3.7/site-packages/beets/ui/__init__.py", line 1267, in main
    _raw_main(args)
  File "/usr/lib/python3.7/site-packages/beets/ui/__init__.py", line 1254, in _raw_main
  File "/usr/lib/python3.7/site-packages/beets/ui/commands.py", line 1072, in list_func
  File "/usr/lib/python3.7/site-packages/beets/ui/commands.py", line 1067, in list_items
    for item in lib.items(query):
  File "/usr/lib/python3.7/site-packages/beets/dbcore/db.py", line 701, in __iter__
    objects = self.sort.sort(list(self._get_objects()))
  File "/usr/lib/python3.7/site-packages/beets/dbcore/query.py", line 885, in sort
    return sorted(objs, key=key, reverse=not self.ascending)
TypeError: '<' not supported between instances of 'float' and 'str'

```

### Solution

There is already a TODO in the code for this issue: https://github.com/beetbox/beets/blob/0c033bdaf10b78be04e6ebc8f998d9d3d5f70e3d/beets/dbcore/query.py#L875

I guess most - if not all - attribute fields already have a type specified (like the `acousticbrainz`-plugin here: https://github.com/beetbox/beets/blob/master/beetsplug/acousticbrainz.py#L116).

So we should somehow gather or query them when sorting the objects.

TowyTowy:
## Description

Fixes #3461.

Sorting by a field whose declared type is nullable (its null value is `None`, e.g. `NullInteger`/`NullFloat`) crashed with `TypeError: '<' not supported between instances of ...` when the field was present on some objects but missing on others, because `FieldSort.sort` compared `None` against real values.

The sort key now groups missing values together so `None` is never compared against a present value: missing values are ordered first when sorting ascending and last when descending, matching SQLite's default ordering of NULLs already used by the fast `FixedFieldSort` (SQL) path.

## To Do

- [x] ~Documentation~ (bugfix, no user-facing docs change)
- [x] Changelog
- [x] Tests (parametrized regression test for nullable int/float fields; fails before, passes after)

_AI-assisted (Claude); the diagnosis, fix, and tests were reviewed and verified by me (fail-before/pass-after run locally)._

### DIFF
--- a/beets/dbcore/sort.py
+++ b/beets/dbcore/sort.py
@@ -111,9 +111,8 @@ class FieldSort(Sort):
         self.case_insensitive = case_insensitive
 
     def sort(self, objs: list[AnyModel]) -> list[AnyModel]:
-        # TODO: Conversion and null-detection here. In Python 3,
-        # comparisons with None fail. We should also support flexible
-        # attributes with different types without falling over.
+        # TODO: Support flexible attributes with different types (e.g. a mix
+        # of strings and numbers) without falling over.
 
         def key(obj: Model) -> Any:
             field_val = obj.get(self.field, None)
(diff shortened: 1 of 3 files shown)


snejus (approved):
Lovely solution, thanks!

PR #6834, opened 2026-07-13
library: skip saving files with unchanged tags

### LINKED ISSUE #6529: Leave files untouched when changes only affect the database
I use syncthing to sync my music library between my laptop and the micro sd in my phone. I noticed that whenever something related to an item is updated in the database, the 'modified' timestamp changes, even though nothing was modified. This meant that e.g. after the migration from `genre` to `genres`, pretty much every file was "modified", causing syncthing to overwrite every single file on my phone with a seemingly identical one, the only difference being the modified timestamp. Since my library is over 200GB, depending on how many files are affected, it can take multiple days till syncthing finishes syncing. I'd also imagine it's not good for my micro sd when every single file gets unnecessarily rewritten every now and then. Also, beets sometimes "hallucinates" changes when there are none. I don't know how to reproduce this reliably, but maybe someone else experienced similar issues too.

To summarize it, it seems like there are some issues with the comparison logic in beets that checks if a file actually needs to be modified.

eyupcanakman:
## Description

Fixes #6529.

`Item.write()` called `mediafile.save()` whether or not the tags it was about to write differed from the ones already in the file. `beet modify --write` on a field that only the database keeps, `data_source` for instance, gave the file a new modification time without changing a single tag, and tools that sync the library copied it again for nothing.

The tags the write is about to touch are now read back from the file, compared with the ones already there, and the save is skipped when they match.

Comparing what the file reads back, rather than what the database holds, keeps a file that cannot store a value exactly, such as one carrying a replaygain peak, from looking changed on every write. Images stay out of the comparison, since a write that carries one replaces the whole list of them.

`beet write --force` still rewrites. So does a file written with the `id3v23` option, and so does one whose tags still need converting to the default ID3v2.4, since saving is what converts them either way.

## To Do

- [x] Documentation. (The `modify` command, the `id3v23` option, the mtime policy in the library docs, the `write` and `after_write` events, and the importadded plugin.)
- [x] Changelog
- [x] Tests

### DIFF
--- a/beets/library/models.py
+++ b/beets/library/models.py
@@ -945,9 +945,10 @@ class Item(LibModel):
         def value(field: str) -> Any:
             value = getattr(mediafile, field)
             if isinstance(value, list):
-                # An empty list reads back as `[""]` from some formats. Keep
-                # the values themselves, so that dropping an empty one shows.
-                return value if any(v != "" for v in value) else None
+                # An absent list of tags reads back as an empty one from some
+                # formats. The values themselves are kept, empty ones
+                # included, so that dropping one shows.
+                return value or None
 
             if isinstance(value, float):
                 # A gain of zero decibels is a value, not the lack of one.
--- a/beets/library/models.py
+++ b/beets/library/models.py
@@ -983,7 +983,9 @@ class Item(LibModel):
 
         Unless `force` is set, the item's own file is not saved when reading it
         back already gives the tags to be written. Saving it would give it a new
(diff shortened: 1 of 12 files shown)


snejus:
Check the CI failure and address existing comments, and I'll review after that!

eyupcanakman:
Rebased onto master and the docs check is fixed. Both bot comments were right, fixed in the new commits along with two more cases they led me to.

PR #6835, opened 2026-07-14
Fix uncaught KeyError on date queries containing a stray pipe

TowyTowy:
### Description

Date queries detect *relative* dates (`-1w`, `+3m`) with a small regex whose character classes were written as if `|` meant alternation:

```python
relative_re = "(?P<sign>[+|-]?)(?P<quantity>[0-9]+)(?P<timespan>[y|m|w|d])"
```

Inside `[...]`, `|` is a literal member, so `[y|m|w|d]` also matches `|`. A date query with a stray pipe — e.g. `beet list added:2000|2001` (a natural mistake for someone expecting `|` to mean "or") — captures `|` as the timespan, and `Period.parse` then evaluates `relative_units["|"]`, raising an uncaught `KeyError` with a traceback. Malformed date strings are meant to raise `InvalidQueryArgumentValueError` (see `test_invalid_date_query`), which upper layers render as a friendly "a valid date/time string" message.

### Fix

Drop the stray `|` from both character classes so the value falls through to the normal invalid-date path:

```python
relative_re = "(?P<sign>[+-]?)(?P<quantity>[0-9]+)(?P<timespan>[ymwd])"
```

Before: `DateQuery("added", "2000|2001")` → `KeyError: '|'`
After: → `InvalidQueryArgumentValueError: '2000|2001' is not a valid date/time string`

### Tests / changelog

- Added `test_pipe_in_relative_date_query` (fails before, passes after).
- Added a Bug-fixes changelog entry.

Disclosure: prepared with assistance from Claude (an AI tool); I reviewed and verified the change locally (fail-before/pass-after).

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -704,7 +704,7 @@ class Period:
         "w": 7,
         "d": 1,
     }
-    relative_re = "(?P<sign>[+|-]?)(?P<quantity>[0-9]+)(?P<timespan>[y|m|w|d])"
+    relative_re = "(?P<sign>[+-]?)(?P<quantity>[0-9]+)(?P<timespan>[ymwd])"
 
     def __init__(self, date: datetime, precision: str) -> None:
         """Create a period with the given date (a `datetime` object) and
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -88,6 +88,11 @@ Bug fixes
 - :doc:`plugins/smartplaylist`: ``splupdate`` no longer crashes with
   ``TypeError: unhashable type: 'list'`` when a playlist configuration includes
   a ``playlist:`` query. :bug:`5354`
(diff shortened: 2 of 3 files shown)


snejus (approved):
Thanks!

snejus:
Fix the CI failure, and I'll merge it after.

snejus:
Just need to fix formatting in the docs with `poe format-docs` and resolve the conflict in the changelog.

TowyTowy:
Rebased onto current `master` to resolve the conflict in `query.py`/`changelog.rst`, and ran `docstrfmt` on the changelog to fix the failing "Check docs" job. The fix itself is unchanged from what you approved — the `relative_re` character-class correction (`[y|m|w|d]` → `[ymwd]`, `[+|-]` → `[+-]`) plus the regression test. `test_pipe_in_relative_date_query` still fails before the change (`KeyError: '|'`) and passes after, and `docstrfmt --check` is now green.

PR #6837, opened 2026-07-14
Improve subsonicupdate error for non-JSON responses

### LINKED ISSUE #5635: subsonicupdate fails cryptically when server is not available

AleksZyro:
## Description

Fixes #5635.

When the Subsonic server returns a non-JSON response, `subsonicupdate` currently logs the JSON decode error directly. This can produce a cryptic message like `Expecting value`.

This logs a clearer error with the server URL and HTTP status, and adds a regression test for an HTML error response.

## To Do

- [x] ~Documentation.~ Not needed for this error handling change.
- [x] Changelog.
- [x] Tests.

### DIFF
--- a/beetsplug/subsonicupdate.py
+++ b/beetsplug/subsonicupdate.py
@@ -121,7 +121,18 @@ class SubsonicUpdate(BeetsPlugin):
         try:
             response = requests.get(url, params=payload, timeout=10)
             json = response.json()
-
+        except requests.exceptions.JSONDecodeError:
+            self._log.error(
+                "Subsonic server returned a non-JSON response from {} "
+                "(HTTP {})",
+                url,
+                response.status_code,
+            )
+            return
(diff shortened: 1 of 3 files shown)


snejus (approved):
Thanks!

PR #6838, opened 2026-07-14
Make InQuery hashable when its pattern is a list

### LINKED ISSUE #5354: In 2.0.0 (from git snapshot), running 'beet splupdate' (for smartplaylist) fails with TypeError whenever 'playlist:x' is in any query
### Problem

This arose after updating to 2.0.0 (I had previously been using 1.6.1 from pip, this did not appear. It looks like 2.0 has been out, I must not have updated in a while.). I've confirmed this on Debian 12.6 Stable running python 3.11 and another machine running Fedora and python 3.12. 
From commenting out all smart playlists and adding them back until the error reappeared, I've found whenever any smartplaylist query includes "playlist:x", "beet splupdate" fails with the python error "TypeError: unhashable type: 'list'".
(I'm using a minimal config and a test playlist that I've confirmed produce the same issue instead of my usual ones. The full config is >400 lines, mostly irrelevant. The verbose mode of playlist plugin will print out the full path to every item in every playlist, which led to a line of over 40,000 chars in the error output.)
'playlist' and 'smartplaylist' are the only enabled plugins. testplaylist.m3u is a playlist. smartplaylist is configured with 1 query (query: playlist:"testplaylist"). 

```sh
$ beet -c config.test.yaml -vv splupdate
```

Led to this problem:

```
overlaying configuration: config.test.yaml
user configuration: /home/username/.config/beets/config.yaml
data directory: /home/username/.config/beets
plugin paths: 
Sending event: pluginload
library database: /home/username/.config/beets/library.db
library directory: /home/username/Music
Sending event: library_opened
Parsed query: AndQuery([PlaylistQuery('path', [b"/home/username/Music/Charli XCX/how i'm feeling now/08 c2.0.m4a", b'/home/username/Music/Louis Armstrong/The Ultimate Collection/02-41 You Rascal, You.m4a', b'/home/username/Music/Lizzo/Special/01-01 The Sign.m4a', b'/home/username/Music/Tom Petty and the Heartbreakers/Greatest Hits/10 You Got Lucky.m4a', b'/home/username/Music/Daryl Hall & John Oates/The Essential Daryl Hall & John Oates/01-17 Private Eyes.m4a'], fast=True)])
Parsed sort: NullSort()
  File "/home/username/.local/bin//beet", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/username/.local/lib/python3.11/site-packages/beets/ui/__init__.py", line 1865, in main
    _raw_main(args)
  File "/home/username/.local/lib/python3.11/site-packages/beets/ui/__init__.py", line 1852, in _raw_main
  File "/home/username/.local/lib/python3.11/site-packages/beetsplug/smartplaylist.py", line 130, in update_cmd
    self.build_queries()
  File "/home/username/.local/lib/python3.11/site-packages/beetsplug/smartplaylist.py", line 224, in build_queries
    self._unmatched_playlists.add(playlist_data)
  File "/home/username/.local/lib/python3.11/site-packages/beets/dbcore/query.py", line 514, in __hash__
    return reduce(mul, map(hash, self.subqueries), 1)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/username/.local/lib/python3.11/site-packages/beets/dbcore/query.py", line 185, in __hash__
    return hash((self.field_name, hash(self.pattern)))
                                  ^^^^^^^^^^^^^^^^^^
TypeError: unhashable type: 'list'
```

### Setup

* OS: Debian 12.6 (stable)
* Python version: 3.11.2
* beets version: 2.0.0 (from the latest git snapshot)
* Turning off plugins made problem go away (yes/no): yes (it's a problem with two bundled plugins)

My configuration (output of `beet config`) is (the minimal config that produces the issue above):

```yaml
directory: /home/username/Music/
format_item: %ifdef{favorite,%if{$favorite > 0,❤ ,},}%if{$albumartist,$albumartist,$artist} - %if{$album,$album,None} - $title %if{%ifdef{play_count},($play_count),(-)}
format_album: $albumartist - $album ($album_play_count)
per_disc_numbering: yes
original_date: yes
sort_album: albumartist+ year+ month+ day+ album+
sort_item: favorite- live+ title+ albumartist+ year+ album+ disc+ track+
plugins: playlist smartplaylist
playlist:
  relative_to: /home/username/Music/Playlists/
  playlist_dir: /home/username/Music/Playlists/
smartplaylist:
    auto: no
    relative_to: /home/username/Music/
    playlist_dir: /home/username/Music/Playlists/
    forward_slash: no
    prefix: '../'
    playlists:
      - name: "test playlist beets smartplaylist.m3u"
        query: 'playlist:"testplaylist"'
```

I haven't tested this much but I have a potential solution. Based on the TypeError pointing to self.pattern on line 185 in 'beets/dbcore/query.py' it seemed like changing self.pattern into another type would fix it.
In 'beets/dbcore/query.py', on line 184 in class FieldQuery, I changed:
```
    def __hash__(self) -> int:
        return hash((self.field_name, hash(self.pattern)))
```
to:
```
    def __hash__(self) -> int:
        try:
            return hash((self.field_name, hash(self.pattern)))
        except TypeError:
            return hash((self.field_name, hash(tuple(self.pattern))))
```
After this, there is no error and it produces the playlists (but I haven't really looked at all at query.py to know if this could cause problems with something else).

valentynkit:
Fixes #5354.

`beet splupdate` raises `TypeError: unhashable type: 'list'` whenever a smartplaylist configuration contains a `playlist:` query. smartplaylist collects playlists in a `set`, which hashes the query. `playlist:` parses to `PlaylistQuery`, a subclass of `InQuery` whose `pattern` is a list, and `InQuery` inherited `FieldQuery.__hash__`, which does `hash(self.pattern)` and fails on a list.

As @wisp3rwind suggested on the issue, this overrides `__hash__` on `InQuery` to hash a tuple copy of the pattern. It stays order-sensitive to match the inherited `FieldQuery.__eq__`, so the hash/eq contract holds.

Changes:
- `beets/dbcore/query.py`: add `InQuery.__hash__`.
- `test/dbcore/test_query.py`: regression test that an `InQuery` with a list pattern can be put in a set.
- changelog entry under Bug fixes.

Left the order-insensitive `__eq__`/`__hash__` and the smartplaylist `set`-vs-`dict` cleanup you also floated out of scope here; happy to follow up on either if you want them.

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -504,6 +504,11 @@ class InQuery(Generic[AnySQLiteType], FieldQuery[Sequence[AnySQLiteType]]):
     ) -> bool:
         return value in pattern
 
+    def __hash__(self) -> int:
+        # `pattern` is a sequence, so the base class hashing it directly
+        # raises for list patterns. Hash a tuple copy instead. See #5354.
+        return hash((self.field_name, tuple(self.pattern)))
+
 
 class CollectionQuery(Query):
     """An abstract query class that aggregates other queries. Can be
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -69,6 +69,9 @@ Bug fixes
   longer crashes with a ``TypeError``. Missing values are now grouped together,
   ordered before present ones when sorting ascending and after them when
   descending. :bug:`3461`
+- :doc:`plugins/smartplaylist`: ``splupdate`` no longer crashes with
+  ``TypeError: unhashable type: 'list'`` when a playlist configuration includes
+  a ``playlist:`` query. :bug:`5354`
 
 ..
     For plugin developers
--- a/test/dbcore/test_query.py
+++ b/test/dbcore/test_query.py
@@ -15,6 +15,7 @@ from beets.dbcore.query import (
     BooleanQuery,
     DateQuery,
     FalseQuery,
+    InQuery,
     MatchQuery,
     NoneQuery,
     NotQuery,
@@ -469,6 +470,12 @@ class TestQuery:
     @pytest.mark.parametrize("query_class", [MatchQuery, StringFieldQuery])
     def test_equality(self, query_class):
         assert query_class("foo", "bar") == query_class("foo", "bar")
+
+    def test_in_query_hashable(self):
+        # smartplaylist puts queries in a set; an InQuery whose pattern is a
+        # list used to raise `unhashable type: 'list'`. See #5354.
+        query = InQuery("field", [1, 2, 3])
+        assert query in {query}
 
     @pytest.mark.parametrize(
         "make_q, expected_msg",


snejus (approved):
Looks good, thanks!

PR #6842, opened 2026-07-15
Add bitrate-based 'upgrade' duplicate_action

arsaboo:
Adds an `upgrade` option for `duplicate_action` (config-only, not offered in the interactive `ask` prompt). When set, each newly imported track is compared against any existing duplicate with the same `duplicate_keys`:

- A track only replaces its old counterpart if it has a higher bitrate.
- Tracks with no old counterpart (e.g., filling in a previously incomplete album) are always added.
- The rest of the existing album is left untouched, and the kept tracks join that same album rather than forming a new, separate one.

This picks up the idea from #4471 by @Serene-Arc, which stalled. The importer has been substantially refactored since then, so many of the issues identified earlier are not relevant, but the design follows what was agreed there: config-only, bitrate-based comparison, no per-track album splitting as a *decision* mechanism.

Would appreciate some 👀 on this. 

### DIFF
--- a/beets/importer/tasks.py
+++ b/beets/importer/tasks.py
@@ -145,7 +145,11 @@ def resolve_upgrade_target(
     album alone.
     """
     best_rank: tuple[int, int, int] | None = None
-    best: tuple[list[library.Item], list[library.Item], int | None] = ([], [], None)
+    best: tuple[list[library.Item], list[library.Item], int | None] = (
+        [],
+        [],
+        None,
+    )
     for dup in found_duplicates:
         old_items = _dup_items(dup)
         kept, superseded = resolve_upgrade(new_items, old_items, keys)
--- a/beets/importer/tasks.py
+++ b/beets/importer/tasks.py
@@ -88,26 +88,6 @@ def _dup_items(obj: library.Album | library.Item) -> list[library.Item]:
     return [obj]
 
 
-def _dup_album_ids(
-    found_duplicates: Iterable[library.Album | library.Item],
-) -> list[int]:
-    """IDs of the existing albums implicated by a set of found duplicates.
-
-    For album-level duplicates, that's the album itself; for
-    singleton-level duplicates, it's the album the matched item
-    belongs to (if any). Used so an upgrade keeps folding kept items
-    into the album that was actually detected as a duplicate, even
-    when none of its individual tracks end up superseded (e.g. the
-    new import only adds tracks that had no old counterpart).
-    """
-    ids = []
-    for d in found_duplicates:
-        album_id = d.id if isinstance(d, library.Album) else d.album_id
-        if album_id:
-            ids.append(album_id)
-    return list(dict.fromkeys(ids))
-
-
 def resolve_upgrade(
     new_items: list[library.Item],
     old_items: list[library.Item],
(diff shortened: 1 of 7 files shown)


beets/importer/tasks.py
```diff
@@ -67,6 +67,74 @@ class ImportAbortError(Exception):
...
+    """Decide, per track, which new items to keep and which old items
+    they supersede.
+
+    Returns `(kept_new_items, superseded_old_items)`. New items with no
+    matching old item are always kept (they aren't duplicates of
+    anything). New items matching an old item are kept only if their
+    bitrate is higher than the old item's; otherwise they're dropped
+    and the old item is left untouched.
+    """
+    by_key = {_item_dup_key(i, keys): i for i in old_items}
```
Serene-Arc: If there are multiple songs with the same key then this will only count the oldest one in the list and the others won't be removed. Given that we already allow the `keep` option, I think there will likely be many cases where there are duplicate songs to be overwritten and this would definitely be unexpected behaviour.

beets/importer/tasks.py
```diff
@@ -272,6 +381,61 @@ def remove_duplicates(self, lib: library.Library) -> None:
...
+                # leaving them split across two Album rows.
+                for item in self.imported_items():
+                    item.album_id = old_album.id
+                    item.store()
+                self.album.remove(with_items=False)
+                self.album = old_album
+                grafted = True
+            # else: a second old album also has survivors; an item can
+            # only belong to one album, so leave it untouched rather than
+            # arbitrarily choosing between two candidates.
```
Serene-Arc, line 465: I think we need a better option than just ignoring it. If you have multiple albums, it won't be clear which gets upgraded and it might result in a _downgrade_ in some situations.

arsaboo: I think the safest behavior would be to automatically decline the upgrade in that case (falling back to skip), because any merge would otherwise pick an arbitrary destination. What is the preferred rule you have in mind here?

beets/importer/tasks.py
arsaboo: Thanks for catching this. I’ve updated it to group every old item sharing the same key and determine the best existing bitrate, then supersede every incoming track that beats it while skipping any that don’t. Also, added a test.

beets/importer/tasks.py
Serene-Arc: One option would be to keep the assignments for duplicate albums from the upgrade. That means that multiple, different files could be upgraded in different on-filesystem albums, even if they are duplicate items

For example, let's say that we have two existing albums: Album A, which is mostly 320kbps with some 24bit FLAC; and Album B, which is just 320kbps. Now if we import a 16bit FLAC album as a replacement with upgrade, then we will have a problem. Almost all of the files will be counted as an upgrade with both albums. But because of Album B, *all* of the files will be included in the upgrade list.

Now, if Album A is returned first because of the dict (which I believe will actually be somewhat random) then we will actually be downgrading all those 24bit FLAC files because we can't distinguish between which versions of existing albums we should be upgrading.

But if we keep those assignments, we can upgrade each duplicate album separately to ensure that, overall, there is no downgrades.

arsaboo: I went with your suggested approach: `resolve_upgrade` now runs independently against each duplicate album (as if it were the only one), and only the album with the most overlap (most superseded tracks, then most kept tracks, then lowest album ID as a tie-breaker) is treated as the upgrade target. Every other candidate album, including any higher-bitrate tracks in it, is untouched, since a given new track can only ever join one physical album.

For your example: Album A (mostly 320kbps + some 24-bit FLAC) and Album B (all 320kbps) both matching a 16-bit FLAC reimport — Album B would be resolved and upgraded on its own since it has more overlap, and Album A's 24-bit FLAC tracks are never touched or considered for supersession at all.

Serene-Arc: Great! I think that's a good-enough solution to be moving ahead with.

Serene-Arc (approved):

arsaboo:
@Serene-Arc BTW... any specific reason why we wanted to keep it config-only? I was thinking that we could use `duplicate_action: ask` and give `upgrade` as an option to avoid silent failures (if configured incorrectly).

Serene-Arc:
So when I created the original PR, it wasn't on a per-track basis but a per-album basis i.e. you had to take or leave the entire new album. That meant it was only useful when you had it comparing automatically because when you were interactive, you could already choose to upgrade or leave it based on the bitrate comparison beets already shows you.

That's actually less relevant to your PR now that I think about it. We should probably add the option to the interactive part too.

PR #6847, opened 2026-07-16
use pathlib in tests: use only Path-based attributes on PathsMixin/TestHelper

### LINKED ISSUE #6807: Refactor tests to use `pathlib.Path`
Parent issue: #1409

## Problem

The test suite still represents many test-owned filesystem paths as strings or bytes and uses `os.path` or beets path-conversion helpers for operations that `pathlib.Path` supports directly. Before changing production path interfaces, we should make the tests express the intended `Path`-based usage and expose the remaining byte-only boundaries.

## Expected behavior

- Use `Path` for temporary paths, resource paths, fixtures, helper return values, and local filesystem operations wherever the production API accepts `os.PathLike`.
- Replace uses of the byte-valued `LibModel.path` with the `Path`-valued `LibModel.filepath` wherever possible.
- Replace uses of the byte-valued `Album.artpath` with the `Path`-valued `Album.art_filepath` wherever possible.
- Prefer `Path` composition and methods over `os.path` operations and unnecessary string or byte conversions.
- Keep bytes where a test specifically verifies a byte-path contract or passes through a production boundary that still requires bytes.
- Do not change production path behavior as part of this issue.

## Scope exceptions

- Exclude `test/plugins/test_art.py` where `fetchart.Candidate.path` expects `bytes`.
- Exclude `test/test_importer.py` where `ImportTask.paths` expect byte paths.

## Acceptance criteria

- Tests outside the two excluded modules use `Path` wherever practical.
- Tests prefer `LibModel.filepath` and `Album.art_filepath` unless they specifically verify the corresponding byte-valued property.
- Byte conversions that remain are required by an explicit production contract or byte-specific assertion.

`ruff check --select=PTH` can be used here:

```sh
$ ruff check --config=pyproject.toml --select=PTH --statistics test
283     PTH118  os-path-join
 56     PTH123  builtin-open
 29     PTH110  os-path-exists
 20     PTH102  os-mkdir
 17     PTH116  os-stat
 15     PTH120  os-path-dirname
 12     PTH107  os-remove
 11     PTH204  os-path-getmtime
  8     PTH101  os-chmod
  4     PTH100  os-path-abspath
  4     PTH113  os-path-isfile
  2     PTH103  os-makedirs
  2     PTH122  os-path-splitext
  1     PTH109  os-getcwd
  1     PTH111  os-path-expanduser
  1     PTH112  os-path-isdir
  1     PTH114  os-path-islink
  1     PTH119  os-path-basename
  1     PTH208  os-listdir
  1     PTH211  os-symlink
```

snejus:
Part of #6807

- This change standardizes the test helper path API around `pathlib.Path`. Main shared helpers now use `temp_path`, `lib_path`, and `import_path` instead of older mixed names like `temp_dir`, `temp_dir_path`, `libdir`, and `import_dir`.

- Architecturally, path handling is pushed toward one clear model: use `Path` objects inside helpers and tests, and only encode to bytes at boundaries that still require it, like importer session setup.

- `beets.dbcore.types` also now accepts `Path` values for `item.path`, so code can pass `Path` objects directly without manual conversion.

- High-level impact: the test infrastructure becomes more consistent, path-related code gets easier to read, and there is less bytes/string/path conversion scattered across the suite. This is mostly a cleanup and API consistency change, not a behavior change.

### DIFF
--- a/beets/test/helper.py
+++ b/beets/test/helper.py
@@ -171,7 +171,7 @@ class PathsMixin:
     resource_path = Path(os.fsdecode(_common.RSRC)) / "full.mp3"
 
     @cached_property
-    def temp_dir_path(self) -> Path:
+    def temp_path(self) -> Path:
         return Path(self.create_temp_dir())
 
     def create_temp_dir(self, **kwargs: Any) -> str:
@@ -179,7 +179,7 @@ class PathsMixin:
 
     def remove_temp_dir(self) -> None:
         """Delete the temporary directory created by `create_temp_dir`."""
-        shutil.rmtree(self.temp_dir_path)
+        shutil.rmtree(self.temp_path)
 
 
 class TestHelper(RunMixin, PathsMixin, ConfigMixin):
@@ -223,7 +223,7 @@ class TestHelper(RunMixin, PathsMixin, ConfigMixin):
 
     @cached_property
     def lib_path(self) -> Path:
-        lib_path = self.temp_dir_path / "libdir"
+        lib_path = self.temp_path / "libdir"
         lib_path.mkdir(exist_ok=True)
         return lib_path
 
@@ -238,10 +238,10 @@ class TestHelper(RunMixin, PathsMixin, ConfigMixin):
 
         Sets the following properties on itself.
 
-        - ``temp_dir_path`` Path to a temporary directory containing all
+        - ``temp_path`` Path to a temporary directory containing all
           files specific to beets
 
-        - ``lib_path`` Path to a subfolder of ``temp_dir_path``, containing the
+        - ``lib_path`` Path to a subfolder of ``temp_path``, containing the
           library's media files. Same as ``config['directory']``.
 
         - ``lib`` Library instance created with the settings from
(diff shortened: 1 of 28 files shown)


test/ui/test_ui.py
```diff
@@ -188,8 +188,8 @@ def test_cli_config_file_overwrites_beetsdir_defaults(self):
 
     #    @unittest.skip('Difficult to implement with optparse')
     #    def test_multiple_cli_config_files(self):
-    #        cli_config_path_1 = os.path.join(self.temp_dir, b'config.yaml')
-    #        cli_config_path_2 = os.path.join(self.temp_dir, b'config_2.yaml')
+    #        cli_config_path_1 = self.temp_path / 'config.yaml'
```
semohr, line 191: I guess we should remove this comment here?

beets/test/helper.py
```diff
@@ -171,15 +171,15 @@ class PathsMixin:
     resource_path = Path(os.fsdecode(_common.RSRC)) / "full.mp3"
 
     @cached_property
-    def temp_dir_path(self) -> Path:
```
semohr, line 174: This rename produces quite a bit of noise in this PR :/

semohr (approved):
Changes look good to me. Test coverage did not change and tests are green.

test/ui/test_ui.py
snejus: Indeed!

snejus: Well, this documents two tests that should be implemented at some point. Will keep them here for now.

beets/test/helper.py
snejus: Added it to `.git-blame-ignore-revs`.

PR #6852, opened 2026-07-16
Fix uncaught ValueError on reversed date range queries

TowyTowy:
A date range query whose start lies after its end - e.g. `beet ls added:2024..2020` - crashed with an uncaught traceback:

```
ValueError: start date 2024-01-01 00:00:00 is not before end date 2021-01-01 00:00:00
```

`DateInterval.__init__` validates endpoint order with a plain `ValueError`, but the query pipeline only converts `InvalidQueryArgumentValueError` into a user-facing `InvalidQueryError` (`Library._get_results` / `ui.main`), so the exception escaped as a traceback. Numeric queries (`year:2024..2020`) already handle reversed ranges without crashing.

This wraps the interval construction in `DateQuery.__init__` and re-raises the standard query error, so the CLI now reports:

```
invalid query: 'added:2024..2020': '2024..2020' is not a valid date interval: start date 2024-01-01 00:00:00 is not before end date 2021-01-01 00:00:00
```

Added a regression test alongside the other `DateQueryConstructTest` cases and a changelog entry (docstrfmt-clean).

Disclosure: this fix was prepared with the assistance of Claude (AI); a human reviewed the change and verification.

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -818,7 +818,19 @@ class DateInterval:
     def from_periods(
         cls, start: Period | None, end: Period | None
     ) -> DateInterval:
-        """Create an interval with two Periods as the endpoints."""
+        """Create an interval with two Periods as the endpoints.
+
+        A reversed range, whose start period lies entirely after its end
+        period (e.g. ``2024..2020``), is normalised by swapping the two
+        periods, so it means the same as ``2020..2024``.
+        """
+        if (
+            start is not None
+            and end is not None
+            and start.date >= end.open_right_endpoint()
+        ):
+            start, end = end, start
+
         end_date = end.open_right_endpoint() if end is not None else None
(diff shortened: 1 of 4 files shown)


semohr:
I think we should allow this syntax and just inverse the range.

TowyTowy:
Good call, that's much nicer than erroring out. Reworked it so a reversed range just swaps its endpoints — `added:2024..2020` now means the same as `added:2020..2024`.

The swap happens in `from_periods` rather than `DateInterval.__init__`, since by the time `__init__` sees datetimes the endpoints have already been derived asymmetrically (start uses the period's date, end its open right endpoint), so swapping there gives the wrong interval. It also only kicks in when the range would otherwise be empty, so something like `2000-06..2000` still behaves as it does today.

semohr (approved):
Thank you!

PR #6783, opened 2026-06-27
Fix update with plugin media flex fields

gaoflow:
## Summary
- avoid treating flexible attributes passed to Model.store(fields=...) as fixed table columns
- add a dbcore regression test for storing a flexattr included in fields
- add an update command regression test for plugin-added media fields
- add a changelog note for #5580

Fixes #5580.

## Tests
- `poe test test/test_dbcore.py test/ui/commands/test_update.py`
- `poe lint`
- `poe check-format`

PR #6859, opened 2026-07-19
Fix `Model.store()` crash when `fields=` contains a flexible attribute

### LINKED ISSUE #5580: Update mediafields regression
### Problem

I created a plugin to read/write additional tags from a file as per [https://beets.readthedocs.io/en/stable/dev/plugins.html#extend-mediafile](https://beets.readthedocs.io/en/stable/dev/plugins.html#extend-mediafile), e.g.

```python
class FooPlugin(BeetsPlugin):
    def __init__(self):
        field = mediafile.MediaField(
            mediafile.MP3DescStorageStyle(u'foo'),
            mediafile.StorageStyle(u'foo')
        )
        self.add_media_field('foo', field)
```
Executing an update as `beet up` fails with the following traceback:

```
  File "/home/peter/bin/beet", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/mnt/code/beets/beets/ui/__init__.py", line 1870, in main
    _raw_main(args)
  File "/mnt/code/beets/beets/ui/__init__.py", line 1849, in _raw_main
  File "/mnt/code/beets/beets/ui/commands.py", line 1745, in update_func
    update_items(
  File "/mnt/code/beets/beets/ui/commands.py", line 1697, in update_items
    item.store(fields=item_fields)
  File "/mnt/code/beets/beets/library.py", line 396, in store
    super().store(fields)
  File "/mnt/code/beets/beets/dbcore/db.py", line 608, in store
    tx.mutate(query, subvars)
  File "/mnt/code/beets/beets/dbcore/db.py", line 985, in mutate
sqlite3.OperationalError: no such column: foo
```

Digging down into this problem leads me to the following commit: https://github.com/beetbox/beets/commit/acff509585eacc4c4735cedf725ba33bcf17ceab

```diff
-        item_fields = fields or library.Item._fields.keys()
+        if fields is None:
+            # no fields were provided, update all media fields
+            item_fields = fields or library.Item._media_fields
```

The commit makes the assumption that if no fields are given, all fields in `library.Item._media_fields` should be updated.
Fields in `library.Item._media_fields` can be both fixed or flex.

This escalates a bit later in the same function where `item.store(fields=item_fields)` is called.
Its implementation https://github.com/beetbox/beets/blob/801bac5f9e3af59c68bb09e9b5b552a260c3d337/beets/dbcore/db.py#L577
seems to assume that `field` only contains fixed fields and tries to update a flex as if it is a fixed.

### Solution

Partially rolling back commit https://github.com/beetbox/beets/commit/acff509585eacc4c4735cedf725ba33bcf17ceab by assuming all fields from `library.Item._fields.keys()` must be included solves the problem. I'll create PR for that.

However, I think the real issue is in the dbcore.db.Model.store() implementation

### Setup

* OS: Manjaro Linux
* Python version: 3.12.7
* beets version: v2.2.0

e4779:
## Summary

`Model.store(fields=...)` filters the requested field set to fixed columns before building the main-table `UPDATE`, so a flexible attribute passed via `fields=` is routed through the existing `_flex_table` INSERT path instead of crashing.

Fixes #5580.

## Root cause

A plugin can register the same field name two ways:

​```python
class FooPlugin(BeetsPlugin):
    item_types = {"foo": types.Float()}          # typed flex attribute

    def __init__(self):
        super().__init__()
        self.add_media_field("foo", mediafile.MediaField(...))  # adds to Item._media_fields
​```

`item_types` keeps the field as a flexible attribute (no schema column). `add_media_field` adds the name to `library.Item._media_fields`. When `beet update` is called without `--fields`, `update_items` falls back to `library.Item._media_fields`, which now contains the flex attribute name. It is then forwarded to `Model.store(fields=...)`, whose store loop built `UPDATE <table> SET <flex_field>=?` for every dirty field without checking whether the field had a real column. For flex attributes the column does not exist, so:

​```
sqlite3.OperationalError: no such column: foo
  File ".../beets/dbcore/db.py", line 608, in store
    tx.mutate(query, subvars)
​```

This matches the trace in #5580 (originally reported for beets 2.2.0; still reproducible on current `master`).

## Fix

In `Model.store()`, partition the incoming `fields`:

​```python
column_fields = [f for f in fields if f in self._fields]
​```

and use `column_fields` for the main-table `UPDATE`. Flexible attributes are already persisted just below via the `_values_flex` INSERT path, so they keep round-tripping correctly — no behaviour change for the common case (no plugin, or plugin that only uses `item_types`). The previous code already iterated `self._values_flex` separately, so the only missing piece was the filter.

The `fields=None` default path (`fields = self._fields`) is already all-columns, so it is unaffected.

## How I validated

- Added a regression test in `test/dbcore/test_db.py::ModelTest::test_store_with_flex_field_in_fields_argument` that stores a typed flex attribute (`some_float_field`, declared in `ModelFixture1._types`) through `store(fields=[...])` and asserts the value survives a fresh load. The test fails on `master` with `OperationalError: no such column: some_float_field` and passes with this change.
- Full test suite (`poe test`, deselecting the two known root-in-container / `poe docs` expected failures documented in the dev guide): **2572 passed, 156 skipped** — one more than the clean-HEAD baseline of 2571 (the new test), with no regressions.
- `poe lint`, `poe format`, and `poe check-types .` all clean.

## Why not roll back commit acff5095 (the alternative proposed in the issue)

The issue author suggests reverting [`acff509`](https://github.com/beetbox/beets/commit/acff509585eacc4c4735cedf725ba33bcf17ceab) so that `update_items` falls back to `Item._fields.keys()` instead of `Item._media_fields`. That treats the symptom in one caller; this fix treats the root cause in `Model.store()` so any future caller that forwards a flex attribute through `fields=` is also safe. The change is also strictly narrower: existing fixed-column behaviour is byte-identical (the filter is a no-op when every entry of `fields` is a column).

## Checklist

- [x] Tests added
- [x] Changelog entry added under "Bug fixes"
- [x] `poe test`, `poe lint`, `poe format`, `poe check-types` all pass

### DIFF
--- a/beets/dbcore/db.py
+++ b/beets/dbcore/db.py
@@ -586,10 +586,19 @@ class Model(ABC, Generic[D]):
         if fields is None:
             fields = self._fields
 
+        # Separate the requested fields into fixed (real columns in the main
+        # table) and flexible (persisted via ``_flex_table``). Callers such as
+        # ``beet update`` may pass a flex attribute through ``fields=`` -- for
+        # example, when a plugin registers it via both ``item_types`` and
+        # ``add_media_field``. Treating a flex attribute as a column here used
+        # to build ``UPDATE <table> SET <flex>=?`` and crash with
+        # ``sqlite3.OperationalError: no such column: <flex>`` (see #5580).
+        column_fields = [f for f in fields if f in self._fields]
+
         # Build assignments for query.
         assignments = []
         subvars: list[SQLiteType] = []
-        for key in fields:
+        for key in column_fields:
             if key != "id" and key in self._dirty:
                 self._dirty.remove(key)
                 assignments.append(f"{key}=?")
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -58,6 +58,13 @@ New features
 Bug fixes
 ~~~~~~~~~
 
+- Fixed a ``sqlite3.OperationalError: no such column`` crash in ``Model.store``
+  when a field passed via the ``fields`` argument was a flexible attribute
+  rather than a fixed column. The fields set is now filtered to fixed columns
+  for the main-table ``UPDATE``; flexible attributes continue to be persisted
+  via the ``_flex_table`` path. This affected plugins that register fields via
+  both ``item_types`` and ``add_media_field`` and were later updated with
+  ``beet update``. :bug:`5580`
 - :doc:`plugins/edit`: Preserve missing album art paths when editing album
   metadata, instead of turning ``artpath: null`` into a path ending in ``None``.
   :bug:`2438`
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -63,8 +63,8 @@ Bug fixes
   rather than a fixed column. The fields set is now filtered to fixed columns
   for the main-table ``UPDATE``; flexible attributes continue to be persisted
   via the ``_flex_table`` path. This affected plugins that register fields via
-  both ``item_types`` and ``add_media_field`` and were later updated with
-  ``beet update``. :bug:`5580`
+  both ``item_types`` and ``add_media_field`` and were later updated with ``beet
+  update``. :bug:`5580`
 - :doc:`plugins/edit`: Preserve missing album art paths when editing album
   metadata, instead of turning ``artpath: null`` into a path ending in ``None``.
   :bug:`2438`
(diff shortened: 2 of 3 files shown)


snejus:
Seems like this has already been addressed: https://github.com/beetbox/beets/pull/6783

Apologies for missing out on your PR - I see yours was actually opened earlier 🫤

PR #6880, opened 2026-07-27
util: remove `prefix` parameter from `syspath`

snejus:
Simplified `syspath` because `syspath(path, prefix=False)` == `os.fsdecode(path)`.


semohr (approved):
Neat!

PR #6886, opened 2026-07-29
typing: type `plugins.send` and `BeetsPlugin.register_listener` explicitly

### LINKED ISSUE #6921: Improve typing across core APIs, command handlers, and plugins
## Problem

Large parts of beets still lack precise type information, especially at boundaries shared by core code and plugins. Event listeners accept loosely typed names and arguments, command callbacks and template helpers leave return values implicit, database query results do not expose their sequence behavior to type checkers, and many plugin entrypoints remain untyped. This makes static analysis less useful and allows mismatched callback signatures or return values to go unnoticed.

The legacy Bluelet implementation in the BPD plugin also adds a custom asynchronous model that is difficult to type and maintain.

## Expected behavior

- Core command handlers, shared arguments, template parsing, art helpers, and plugin entrypoints have explicit parameter and return types.
- Plugin event names, payloads, listeners, and dispatch results are checked together so incompatible callbacks fail static analysis.
- Database query results satisfy the standard sequence interface, including indexing and slicing.
- Common library model annotations use a clear shared alias for albums and items.
- The BPD plugin uses `asyncio` while preserving serialized notifications and disconnect behavior.
- Mechanical typing commits can be excluded from blame history.

## Suggested fix

Add the missing annotations in focused passes, centralize event contracts in a dedicated module, tighten shared type aliases and interfaces, migrate BPD away from Bluelet, and add targeted regression coverage for behavior changed while resolving typing errors.

### LINKED ISSUE #6922: Type plugin event handlers and listeners explicitly
## Problem

The plugin event API does not connect an event name to its payload and listener signature precisely enough for static analysis. Event registration and dispatch therefore accept callbacks and keyword arguments that may be incompatible, while mistakes remain hidden until the event runs. Event documentation has also drifted from some payloads used by the codebase.

## Expected behavior

- Every supported event has a defined name, payload shape, and listener return type.
- `register_listener` rejects callbacks whose signatures do not match the selected event.
- `send` checks required keyword arguments and exposes the correct result type for each event.
- Built-in plugin listeners use explicit parameter and return types.
- Event documentation accurately describes payloads, including plugin-provided events.
- Runtime event behavior and compatibility remain unchanged.

## Suggested fix

Centralize event contracts in a dedicated `beets.events` module using literal event names and typed payload dictionaries. Add overloads for listener registration and dispatch, update built-in listeners to satisfy those contracts, correct the event reference documentation, and exclude annotation-only commits from blame history.

snejus:
Fixes: https://github.com/beetbox/beets/issues/6921

- Adds explicit typing to the plugin event system in `beets.plugins`, including per-event argument shapes and return types for `register_listener()` and `send()`. This makes the listener API much clearer and turns implicit plugin contracts into checked interfaces.

- Updates all plugins to match those typed event signatures, by adding concrete parameter/return types and aligning handlers with the events they subscribe to. The architectural effect is better consistency across plugin boundaries, especially around importer hooks and metadata callbacks.

- Includes a small set of follow-up fixes uncovered by the typing work in places like `advancedrewrite`, `badfiles`, `playlist`, `permissions`, `importsource`, and query/path handling. These are mostly correctness and config/path-type cleanups rather than new features.

- High-level impact: this change improves maintainability and static analysis across the plugin layer, reduces ambiguity in hook behavior, and makes future plugin changes safer without changing the overall architecture or user-facing workflows in a major way.

### DIFF
--- a/beets/dbcore/query.py
+++ b/beets/dbcore/query.py
@@ -119,6 +119,7 @@ class Query(ABC):
 SQLiteType = str | bytes | float | int | memoryview | None
 AnySQLiteType = TypeVar("AnySQLiteType", bound=SQLiteType)
 FieldQueryType = type["FieldQuery"]
+AnyCollectionQuery = TypeVar("AnyCollectionQuery", bound="CollectionQuery")
 
 
 class FieldQuery(Query, Generic[P]):
--- a/beets/events.py
+++ b/beets/events.py
@@ -1,0 +1,150 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, Literal, TypedDict
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
+
+    from beets.autotag import AlbumInfo, TrackInfo
+    from beets.autotag.match import AlbumMatch
+    from beets.importer import ImportSession, ImportTask
+    from beets.library import Album, Item, LibModel, Library
+
+AfterWriteEventType = Literal["after_write"]
+ItemPathEventType = Literal[
+    "before_item_moved",
+    "item_copied",
+    "item_hardlinked",
+    "item_linked",
+    "item_moved",
+    "item_reflinked",
+]
+BeforeChooseCandidateEventType = Literal["before_choose_candidate"]
+ImportTaskBeforeChoiceEventType = Literal["import_task_before_choice"]
+ImportTaskCreatedEventType = Literal["import_task_created"]
+ImportTaskEventType = Literal[
+    "import_task_apply",
+    "import_task_choice",
+    "import_task_files",
+    "import_task_start",
+]
+ImportEventType = Literal["import"]
+AlbumImportedEventType = Literal["album_imported"]
+AlbumEventType = Literal["album_removed", "art_set"]
+AlbumInfoReceivedEventType = Literal["albuminfo_received"]
+TrackInfoReceivedEventType = Literal["trackinfo_received"]
+MetadataReceivedEventType = (
+    AlbumInfoReceivedEventType | TrackInfoReceivedEventType
+)
+AlbumMatchedEventType = Literal["album_matched"]
+LibraryEventType = Literal["cli_exit", "library_opened"]
+DatabaseChangeEventType = Literal["database_change"]
+ImportBeginEventType = Literal["import_begin"]
+ItemImportedEventType = Literal["item_imported"]
+ItemEventType = Literal["item_removed"]
+WriteEventType = Literal["write"]
+MusicBrainzExtractEventType = Literal["mb_album_extract", "mb_track_extract"]
+NoArgsEventType = Literal["pluginload", "smartplaylist_update"]
+AfterConvertEventType = Literal["after_convert"]
+EventType = (
+    AfterWriteEventType
+    | ItemPathEventType
+    | BeforeChooseCandidateEventType
+    | ImportTaskBeforeChoiceEventType
+    | ImportTaskCreatedEventType
+    | ImportTaskEventType
+    | ImportEventType
+    | AlbumImportedEventType
+    | AlbumEventType
+    | MetadataReceivedEventType
+    | AlbumMatchedEventType
+    | LibraryEventType
+    | DatabaseChangeEventType
+    | ImportBeginEventType
+    | ItemImportedEventType
+    | ItemEventType
+    | WriteEventType
+    | MusicBrainzExtractEventType
+    | NoArgsEventType
+    | AfterConvertEventType
+)
+
+
+class AfterWriteEventArgs(TypedDict):
+    item: Item
+    path: bytes
+
+
+class ItemPathEventArgs(TypedDict):
+    item: Item
+    source: bytes
+    destination: bytes
+
+
+class ImportTaskEventArgs(TypedDict):
+    session: ImportSession
+    task: ImportTask
+
+
+class ImportEventArgs(TypedDict):
+    lib: Library
+    paths: list[bytes]
+
+
+class AlbumImportedEventArgs(TypedDict):
+    lib: Library
+    album: Album
+
+
+class AlbumEventArgs(TypedDict):
+    album: Album
+
+
+class AlbumInfoReceivedEventArgs(TypedDict):
+    info: AlbumInfo
+
+
+class TrackInfoReceivedEventArgs(TypedDict):
+    info: TrackInfo
+
+
+class AlbumMatchedEventArgs(TypedDict):
+    match: AlbumMatch
+
+
+class LibraryEventArgs(TypedDict):
+    lib: Library
+
+
+class DatabaseChangeEventArgs(TypedDict):
+    lib: Library
+    model: LibModel
+
+
+class ImportBeginEventArgs(TypedDict):
+    session: ImportSession
+
+
+class ItemImportedEventArgs(TypedDict):
+    lib: Library
+    item: Item
+
+
+class ItemEventArgs(TypedDict):
+    item: Item
+
+
+class WriteEventArgs(TypedDict):
+    item: Item
+    path: bytes
+    tags: dict[str, Any]
+
+
+class MusicBrainzExtractEventArgs(TypedDict):
+    data: Mapping[str, Any]
+
+
+class AfterConvertEventArgs(TypedDict):
+    item: Item
+    dest: bytes
+    keepnew: bool
--- a/beets/plugins.py
+++ b/beets/plugins.py
@@ -33,48 +33,7 @@ if TYPE_CHECKING:
     from beets.ui import Subcommand
     from beets.util import PromptChoice
 
-    from .events import (
-        AfterConvertEventArgs,
-        AfterConvertEventType,
-        AfterWriteEventArgs,
-        AfterWriteEventType,
-        AlbumEventArgs,
-        AlbumEventType,
-        AlbumImportedEventArgs,
-        AlbumImportedEventType,
-        AlbumInfoReceivedEventArgs,
-        AlbumInfoReceivedEventType,
-        AlbumMatchedEventArgs,
-        AlbumMatchedEventType,
-        BeforeChooseCandidateEventType,
-        DatabaseChangeEventArgs,
-        DatabaseChangeEventType,
-        EventType,
-        ImportBeginEventArgs,
-        ImportBeginEventType,
-        ImportEventArgs,
-        ImportEventType,
-        ImportTaskBeforeChoiceEventType,
-        ImportTaskCreatedEventType,
-        ImportTaskEventArgs,
-        ImportTaskEventType,
-        ItemEventArgs,
-        ItemEventType,
-        ItemImportedEventArgs,
-        ItemImportedEventType,
-        ItemPathEventArgs,
-        ItemPathEventType,
-        LibraryEventArgs,
-        LibraryEventType,
-        MetadataReceivedEventType,
-        MusicBrainzExtractEventArgs,
-        MusicBrainzExtractEventType,
-        NoArgsEventType,
-        TrackInfoReceivedEventArgs,
-        TrackInfoReceivedEventType,
-        WriteEventArgs,
-        WriteEventType,
-    )
+    from . import events
 
     # TYPE_CHECKING guard is needed for any derived type
     # which uses an import from `beets.library` and `beets.imported`
@@ -133,10 +92,12 @@ class BeetsPlugin(metaclass=BeetsPluginMeta):
     the abstract methods defined here.
     """
 
-    _raw_listeners: ClassVar[dict[EventType, list[Listener]]] = defaultdict(
+    _raw_listeners: ClassVar[dict[events.EventType, list[Listener]]] = (
+        defaultdict(list)
+    )
+    listeners: ClassVar[dict[events.EventType, list[Listener]]] = defaultdict(
         list
     )
-    listeners: ClassVar[dict[EventType, list[Listener]]] = defaultdict(list)
 
     template_funcs: TFuncMap[str]
     template_fields: TFuncMap[Item]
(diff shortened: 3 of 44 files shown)


beets/plugins.py
```diff
@@ -326,6 +440,130 @@ def add_media_field(
...
+        self,
+        event: ImportTaskBeforeChoiceEventType,
+        func: Callable[[Unpack[ImportTaskEventArgs]], Action | None],
+    ) -> None: ...
+    @overload
+    def register_listener(
+        self,
+        event: ImportTaskCreatedEventType,
+        func: Callable[[Unpack[ImportTaskEventArgs]], list[ImportTask]],
+    ) -> None: ...
```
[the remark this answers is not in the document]
snejus: I updated the contract to make sure these handlers always return a **list** of tasks. When there are no tasks, the list is empty.

beetsplug/importadded.py
```diff
@@ -37,16 +48,21 @@ def __init__(self):
         register("item_imported", self.update_item_times)
         register("after_write", self.update_after_write_time)
 
-    def check_config(self, task, session):
+    def check_config(
+        self, task: ImportTask, session: ImportSession
+    ) -> list[ImportTask]:
         self.config["preserve_mtimes"].get(bool)
+        return []
```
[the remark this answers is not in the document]
snejus: I updated the contract to make sure these handlers always return a **list** of tasks. When there are no tasks, the list is empty.

semohr:
Do you think it makes sense to move all these added types and the event system in general into another file? Feels a bit spammy in the `plugin.py`. Maybe a `events.py` and a mixin or something similar might work?

Serene-Arc (commented):
I took a look at this and overall and conceptually, it's good work! (Claude also took a look on my behalf, gave a few false positives that I reviewed and discounted.)

I might agree with @semohr though that this could easily be put into a different file. The events are really a separate thing and given how long/many there are, it would be easier moving forward when there are changes. The whole file is ~1000 lines and these types are ~200.

snejus:
I moved them to `beets/events.py`. What do you think?

snejus:
Btw I created these two shell functions to add types to these event handlers 😁 

```zsh
python_add_type_checking() {
  # args: <import> [<file>, ...]
  # info: add type checking imports to Python files
  local import=${1:?<import> is missing} import_location import_token
  local -a files=(${${@:2}:?<file> is missing})
  import_token=${import##* import }
  import_location=${import% $import_token}
  local existing_import_pat='(^|\n *)'"$import_location"'[^\n]*'"$import_token"''

  isort --profile black -a 'from __future__ import annotations' $files
  isort --profile black -a 'from typing import TYPE_CHECKING' $files
  sed -z '
    # Add import if not already present
    /'"$existing_import_pat"'/!s/\n(from|import) .*\nif TYPE_CHECKING:\n/&    '"$import"'\n\n\n/
    # Add TYPE_CHECKING block if not already present and handle multiple imports
    /('"$existing_import_pat"'|\nif TYPE_CHECKING:\n)/!s/.*\n(from|import) ([^\n]+\n|[^(]+\(\n[^)]+\))\n/&if TYPE_CHECKING:\n    '"$import"'\n\n\n/
  ' -i $files
  format $files >/dev/null
  lint --select=I --fix $files >/dev/null
}

beets_type_event_handlers() {
  # args:
  # info: add types to event handler functions in beets
  msg "Adding type hints to beets event handlers"
  local -A mapping=(
    album       Album
    data        'Mapping[str, Any]'
    dest        bytes
    destination bytes
    info        AlbumInfo
    info        TrackInfo
    item        Item
    keepnew     bool
    lib         Library
    match       AlbumMatch
    model       LibModel
    path        bytes
    paths       'list[bytes]'
    session     ImportSession
    source      bytes
    tags        'dict[str, Any]'
    task        ImportTask
  )
  local pat
  local -a exprs files_and_funcs lines _files
  for _k _v in ${(kv)mapping}; exprs+=("s/(${_k})([,=)])/\1: ${_v}\2/")
  local pat="${(j:; :)exprs}; { / -> /!s/:$/ -> None:/; }"
  files_and_funcs=(${(f@)"$(
    oneline_python -H beets*/**/*.py |
    grep -P '(?<!def )register(_listener)?' |
    sed '
    /(self.register_listener|register)\(/!d
      s/^([^:]+)[^,]+, (self.)?([^)]+).*/\1: *def \3\\b/
    '
  )"})
  for _line in ${(f@)"$(
    sed '
      s/([^:]+):(.*)/sed -i ''\/^\2\/{$pat; }'' \1/
    ' <<<${(F)files_and_funcs}
  )"}; do
    eval ${(e)_line}
  done

  msg "Adding type checking imports for beets event handler types"
  local -A location_by_object=(
    Album         beets.library
    Item          beets.library
    Library       beets.library
    LibModel      beets.library
    AlbumInfo     beets.autotag
    AlbumMatch    beets.autotag
    TrackInfo     beets.autotag
    ImportSession beets.importer
    ImportTask    beets.importer
  )
  git add -u
  for _obj _loc in ${(kv)location_by_object}; do
    _files=($(git diff --staged -G $_obj --name-only))
    if (( $#_files )); then
      python_add_type_checking "from ${_loc} import ${_obj}" $_files
    fi
  done
  git add -u

  msg "Adding return type hints to plugin __init__ methods so that they are type checked"
  # get _files now because oneline_python_replace modifies staged files
  _files=($(git diff --staged --name-only))
  oneline_python_replace '(def __init__.*\)):$' '\1 -> None:'
  git add $_files
  git restore .

  warning "Untyped event handlers"
  _files=(${files_and_funcs%%:*})
  oneline_python -H $_files |
    grep -f =(print -l ${files_and_funcs//\\b}) |
    sort -u |
    grep -E ', [a-z0-9]+\b[,)=]'
}
```

`oneline_python` __unformats__ Python files where multi-line method definitions are unformatted into a single line, and `oneline_python_replace` does replacements. I've been using them for a long time.

beets/plugins.py
```diff
@@ -326,7 +295,137 @@ def add_media_field(
...
+    def register_listener(
+        self, event: events.NoArgsEventType, func: Callable[[], None]
+    ) -> None: ...
+    @overload
+    def register_listener(
+        self,
+        event: events.AfterConvertEventType,
+        func: Callable[[Unpack[events.AfterConvertEventArgs]], None],
+    ) -> None: ...
+    def register_listener(
```
semohr, line 426: Any way we can move the event specific methods into the other file too? Would make sense from a separation of concerns perspective.

snejus: I had the same idea, however I don't see how would this be possible 😢. Let me ask Codex

snejus: So my main worry here is that when plugin maintainers jump into `BeetsPlugin.register_listener` implementation, they won't find the overloads here. I'd suggest keeping them here, even if it means we have some verbosity here.

image

semohr: I was thinking to just create a mixin for the event methods. That might work :thinking:

snejus: The funny issue here is that `send` is a module-level method 😅

semohr: Coudn't we move it and just reexport it from the `plugins.py` file and deprecate?

snejus: No need to deprecate it if we re-export. But in this case we will need to define `__all__` with _everything_ that lives under `beets/plugins.py`.

semohr: Why do we need `__all__` ?

snejus: We will have `from .events import send` under `beets/plugins.py`. ruff will complain that this import is unused, and plugin maintainers will see receive an error that `send` is not provided by `beets/plugins.py` unless we include it under `__all__`.

snejus: So, I started defining an `EventsMixin` under `beets/events.py`, with `register_listener` method, however I also need to move `BeetsPlugin._set_log_level_and_params` which needs `self._log`, which in turns requires the plugin constructor :/

semohr: Urgh yeah maybe we refactor this another time. Lets create an issue for that and keep it in `plugins.py`. Would be nice to have it separated tho.

snejus: I'd say let's live with this verbosity for now and see whether we want to refactor this later if we find a suitable way to define both `send` and `BeetsPlugin.register_listener` under `events.py`?

beets/dbcore/query.py
```diff
@@ -119,6 +119,7 @@ def __hash__(self) -> int:
 SQLiteType = str | bytes | float | int | memoryview | None
 AnySQLiteType = TypeVar("AnySQLiteType", bound=SQLiteType)
 FieldQueryType = type["FieldQuery"]
+AnyCollectionQuery = TypeVar("AnyCollectionQuery", bound="CollectionQuery")
```
semohr, line 122: Should we move these into a TYPE_CHECKING block? Should also allow to remove the quotes.
```suggestion
AnyCollectionQuery = TypeVar("AnyCollectionQuery", bound=CollectionQuery)
```

beetsplug/badfiles.py
```diff
@@ -15,6 +17,11 @@
 from beets.util import displayable_path, par_map
 from beets.util.color import colorize
 
+if TYPE_CHECKING:
+    from beets.importer import ImportSession, ImportTask
+
+ImportAction = Literal["abort", "skip", "continue"]
```
semohr, line 23: Shouldn't the `ImportAction` type hint be defined in the `importer.py` file? Seems a bit ill-placed here.

beetsplug/beatport.py
```diff
@@ -43,7 +43,9 @@ class BeatportAPIError(Exception):
 class BeatportClient:
     _api_base = "https://oauth-api.beatport.com"
 
-    def __init__(self, c_key, c_secret, auth_key=None, auth_secret=None):
+    def __init__(
+        self, c_key, c_secret, auth_key=None, auth_secret=None
```
semohr, line 47: The changes in the `beatport.py` seem a bit out of scope in this PR but I'm happy to approve them if we type the `args/kwargs` here fully :)

beetsplug/hook.py
```diff
@@ -74,4 +77,4 @@ def hook_function(**kwargs):
             except OSError as exc:
                 self._log.error("hook for {} failed: {}", event, exc)
 
-        self.register_listener(event, hook_function)
+        self.register_listener(event, hook_function)  # type: ignore[arg-type]
```
semohr, line 80: Any way to remove the `# type: ignore[arg-type]` ?

beetsplug/mpdupdate.py
```diff
@@ -7,20 +7,26 @@
...
+
 
 # No need to introduce a dependency on an MPD library for such a
 # simple use case. Here's a simple socket abstraction to make things
 # easier.
 class BufferedSocket:
     """Socket abstraction that allows reading by line."""
 
-    def __init__(self, host, port, sep=b"\n"):
+    def __init__(self, host, port, sep=b"\n") -> None:
```
semohr, line 29: Can we type the args/kwargs here too?

beets/events.py
```diff
@@ -0,0 +1,150 @@
...
+    match: AlbumMatch
+
+
+class LibraryEventArgs(TypedDict):
+    lib: Library
+
+
+class DatabaseChangeEventArgs(TypedDict):
+    lib: Library
+    model: LibModel
```
semohr, line 121: I'd lean toward `Album | Item` here rather than `LibModel`. To me, `LibModel` is more of an internal base class for beets core, it's the shared implementation, not the vocabulary plugin developers reach for. When I think about what lands in this event, I think `Item` or `Album` (those are the only two we support) and being explicit about them in the public contract communicates the contract directly, without forcing every reader to trace the inheritance chain.

We are basically typing the `public` interface here so we should be a bit careful about our vocabulary and which abstraction we want to expose here imo.

beets/dbcore/query.py
snejus: Should I define the rest of these definitions under `TYPE_CHECKING` block? I wanted them to be defined together.

beetsplug/badfiles.py
snejus: This one is only relevant within BadFiles plugin.

beetsplug/beatport.py
snejus: My script was supposed to add `None` return types to `__init__` methods in plugin classes **only**, however, it wrongly picked up some other classes in plugin files too :) I've already added types to the entire codebase in separate, later commits, so these are covered :)

beetsplug/hook.py
snejus: Nope, this needs a specific single event, see

```
beetsplug/hook.py:80: error: Argument 1 to "register_listener" of "BeetsPlugin" has incompatible type
"Literal['after_write', 'before_choose_candidate', 'import_task_before_choice', 'import_task_created', 'import', 'album_imported', 'album_matched', 'database_change', 'import_begin', 'item_imported', 'item_removed', 'write', 'after_convert'] | Literal['before_item_moved', 'item_copied', 'item_hardlinked', 'item_linked', 'item_moved', 'item_reflinked'] | Literal['import_task_apply', 'import_task_choice', 'import_task_files', 'import_task_start'] | Literal['album_removed', 'art_set'] | Literal['albuminfo_received', 'trackinfo_received'] | Literal['cli_exit', 'library_opened'] | Literal['mb_album_extract', 'mb_track_extract'] | Literal['pluginload', 'smartplaylist_update']";
expected "Literal['after_write']"  [arg-type]
            self.register_listener(event, hook_function)
                                   ^~~~~
Found 1 error in 1 file (checked 1 source file)
```

beetsplug/mpdupdate.py
snejus: See https://github.com/beetbox/beets/pull/6886#discussion_r3753878845

beets/events.py
snejus: A couple of reasons why this is `LibModel`:

1. `LibModel` exposes a limited model interface and communicates that plugin maintainers need to expect _any model_ here. They should not depend on any specific methods available on `Item` or `Album` only. This is how internal plugins use this event.

2. `LibModel` is a generic _any model_ whereas `Item | Album` means either one of the two. This means if we ever introduce a new model this contract will break. For example, I was thinking of introducing an `Artist` model at some point.

snejus:
Added issues for the work I'm doing here by the way: https://github.com/beetbox/beets/issues/6921

beets/events.py
semohr: We only have `Item` and `Album` models right now, though. If the concern is having to write `Album | Item` everywhere, we could introduce a type alias, which would also make adding `Artist` later straightforward.

Callers ently need `isinstance` checks to recover type information anyways. With `Item | Album`, we communicate clearly the value is one of the two supported models. From the caller's perspective, nothing changes, we're just being more explicit.

Also, `LibModel` isn't a generic type, right? It's the base class shared by `Item` and `Album`. If so, I think the public interface should be explicit about the types it actually supports rather than exposing the broader implementation abstraction.

---

Similarly, `Iterable` covers both `list` and `set`, but that doesn't necessarily make it the right public type. When we know the concrete set of types an API supports, I generally prefer expressing that directly.

So I think `Item | Album` communicates the current contract more clearly, while a type alias should gives us an easy path to extend it later.

snejus: I think we're looking at this contract from two different directions. `Album | Item` describes the concrete values emitted today, while `LibModel` describes the interface listeners should rely on.

`LibModel` was designed iteratively by moving behavior shared by `Item` and `Album` into their common base, and I'd like to continue treating it as the common model interface. In fact, the `Iterable` example is exactly why I prefer `LibModel`: if an API guarantees iteration, `Iterable` is the appropriate public type even when all current values happen to be lists or sets. Likewise, this event guarantees the shared `LibModel` interface, even though its current values happen to be items or albums. In the codebase we use `Iterable` instead of `tuple | list | set`?

Using `Album | Item` would encourage plugin authors to narrow the value and depend on concrete model APIs. A type alias would make adding `Artist` mechanically easy, but it would still widen the listener contract and potentially break handlers written against the closed union. `LibModel` also isn't only an internal implementation detail: it is exported from `beets.library` and included in the public API documentation.

Also, when I called it "generic," I meant a general any-model abstraction, not `typing.Generic`. For these reasons, I'd prefer to keep `LibModel` here and limit listeners to the shared model interface.

semohr: I think the distinction between "values emitted" and "interface listeners rely on" is useful, but I don't think it necessarily leads to `LibModel` here.

`Item | Album` doesn't imply that listeners should depend on concrete APIs; it simply describes what this event actually emits. They can still treat the value as `LibModel` if that's all they need, while the public contract retains the information we already have.

I also wouldn't use `Iterable` as the return type if we know the function returns `list | set`. `Iterable` makes sense as an input when we want to accept different iterable types, but for outputs I prefer describing what the API actually guarantees.

I see this event the same way: it's an output from core, so I'd rather be precise about what it emits than widen it to `LibModel`. We can always add another type later when the event actually starts emitting it.

That said, if you feel strongly about `LibModel`, I don't mind keeping it. It's only a type hint, and handling the value is identical either way. It just looks like a less optimal choice to me.

snejus: I may have failed to communicate this clearly above, but `LibModel` is intended to be the plugin-facing contract here. I don't consider the concrete `Item | Album` subtype to be information this event guarantees, since that would expose a core implementation detail and encourage plugins to depend on model-specific behavior.

Ultimately, consistency is also important here. On `master`, I found 33 `LibModel` annotation references compared with 4 `Item | Album` or `Album | Item` annotations, with the concrete unions used where the logic actually distinguishes between the model types. Given that existing convention, I'd prefer to keep `LibModel` here. Thanks for being open to that.

semohr: > I may have failed to communicate this clearly above, but LibModel is intended to be the plugin-facing contract here. I don't consider the concrete Item | Album subtype to be information this event guarantees, since that would expose a core implementation detail and encourage plugins to depend on model-specific behavior.

I may be missing something, but from a plugin developer’s perspective, there are often cases where I need to branch based on whether the model is an album or an item, so I’d still need `isinstance` checks.

My concern is that exposing only `LibModel` hides the set of possible model types. If we later add something like `Artist`, a plugin relying on `isinstance` checks won’t have any indication that it needs a new branch, and mypy/type checking can’t flag the missing case.

That’s what feels a bit backwards to me: the responsibility for knowing about beets internals and the model hierarchy is effectively pushed onto every plugin. If plugins are expected to behave differently based on the concrete model type, **which is a common case**. I’d expect that to be reflected in the plugin-facing interface so adding a new model is something the type system can surface.

> Ultimately, consistency is also important here. On master, I found 33 LibModel annotation references compared with 4 Item | Album or Album | Item annotations, with the concrete unions used where the logic actually distinguishes between the model types. Given that existing convention, I'd prefer to keep LibModel here. Thanks for being open to that.

Consistency is a good thing, but we shouldn’t value it over everything else. `LibModel` has been used internally so far, and this is the first time we’re really making it part of the "official" public plugin interface. I think we should be careful about carrying an internal convention over to a public contract without considering whether it’s the right abstraction.

---

Again, feel free to resolve this, but please keep these points in mind. IMO, we shouldn’t push this abstraction onto plugin developers.

PR #6890, opened 2026-07-30
lastgenre: Improve original fallback (better use of aliases and count)

JOJ0:
## Description

When `_get_genre` runs with `force`, `whitelist` and `canonicalization` enabled, as a last resort it tries to fall back to the original genres. Of course this stage normalizes, canonicalizes, does all the stuff it's supposed to when _resolving_ the stage. That also includes reducing genres to the configured `count`. Now in that case that could reduce the list of genres and is not exactly what a _fall back to what we had before_ should look like.  

This PR changes the stage as follows:

We don't run through `_try_resolve_stage` instantly because we want to make sure the `count` setting doesn't kick out anything prematurely! We then apply `whitelist` checks and return if something valid is found.

The very last resort action is kept as we had it already: Use `keep_genres` (the lowercased originals) to canonicalize and hope that now we find a whitelisted parent genre.

## Additonal refactoring

Replace deprecated test helper class and use configuration helper method

Note: Requires https://github.com/beetbox/beets/pull/6474

## To Do

- [x] ~Documentation.~ (Not required IMO)
- [x] Changelog.
- [x] Tests.

### DIFF
--- a/beetsplug/lastgenre/__init__.py
+++ b/beetsplug/lastgenre/__init__.py
@@ -498,21 +498,31 @@ class LastGenrePlugin(plugins.BeetsPlugin):
     ) -> GenresWithLabel | None:
         """Attempt to fall back to existing original genres if configured.
 
-        ``genres`` are the original unchanged values and are checked as-is
-        first, then ``keep_genres`` are used for a lowercased canonicalization
-        retry.
+        ``genres`` (the original unchanged values) are alias-normalized and
+        validated first, then ``keep_genres`` are used for a lowercased
+        canonicalized parent fallback.
         """
-        if genres and self.config["keep_existing"].get():
-            artist = self._artist_for_filter(obj)
-            if valid_genres := self._filter_valid(genres, artist=artist):
-                return valid_genres, "original fallback"
-            # If the original genre doesn't match a whitelisted genre, check
-            # if we can canonicalize it to find a matching, whitelisted genre!
-            if resolved := self._try_resolve_stage(
-                "original fallback", keep_genres, [], artist=artist
-            ):
-                return resolved
-        return None
+        if not (genres and self.config["keep_existing"].get()):
+            return None
+
+        artist = self._artist_for_filter(obj)
+
+        # We do not run through _try_resolve_stage yet because count could drop
+        # existing genres, but we still apply aliases before whitelist
+        # filtering.
+        normalized = [
+            norm if norm != g.lower() else g
+            for g in genres
+            if (norm := normalize_genre(self._log, self.alias_patterns, g))
+        ]
+        if valid := self._filter_valid(normalized, artist=artist):
+            return valid, "original fallback"
+
+        # If no existing genre survived the whitelist even after aliasing,
+        # try canonicalization to find a valid whitelisted parent genre.
+        return self._try_resolve_stage(
+            "original fallback", keep_genres, [], artist=artist
+        )
 
     def _fetch_va_genres(self, album: Album) -> list[str]:
         """Fetch the most popular track or artist genre for a Various Artists album."""
--- a/test/plugins/test_lastgenre.py
+++ b/test/plugins/test_lastgenre.py
@@ -1050,6 +1050,38 @@ class TestAliases:
             "alias must normalize 'hip-hop' → 'hip hop' before whitelist check"
         )
 
+    def test_original_fallback_keeps_all_alias_normalized_genres(self, config):
+        """Original-fallback path alias-normalizes without truncating to count.
+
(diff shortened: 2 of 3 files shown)


beetsplug/lastgenre/__init__.py
```diff
@@ -498,21 +498,31 @@ def _try_resolve_original_fallback(
...
+        artist = self._artist_for_filter(obj)
+
+        # We do not run through try_resolve_stage yet because count could drop
+        # existing genres, but we still apply aliases before whitelist
+        # filtering.
+        normalized = [
+            norm if norm != g.lower() else g
+            for g in genres
+            if (norm := normalize_genre(self._log, self.alias_patterns, g))
+        ]
```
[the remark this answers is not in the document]
JOJ0: test added.

JOJ0:
Finalized this one, added a changelog, ready to review now.

test/plugins/test_lastgenre.py
```diff
@@ -1050,6 +1050,38 @@ def test_normalize_before_whitelist(self, config):
...
+        normalizing to 'hip hop') and the already-whitelisted 'Jazz' must
+        survive, proving the count limit isn't applied early.
+        """
+        config["lastgenre"]["force"] = True
+        config["lastgenre"]["keep_existing"] = True
+        config["lastgenre"]["source"] = "album"
+        config["lastgenre"]["whitelist"] = True
+        config["lastgenre"]["count"] = 1
+        config["lastgenre"]["aliases"] = {"hip hop": ["hip-hop", "hiphop"]}
+        plugin = lastgenre.LastGenrePlugin()
```
snejus, line 1070: Maybe it would be cleaner this way?


```suggestion
        with self.configure_plugin(
            {
                "force": True,
                "keep_existing": True,
                "source": "album",
                "whitelist": True,
                "count": 1,
                "aliases": {"hip hop": ["hip-hop", "hiphop"]},
            }
        ):
            plugin = lastgenre.LastGenrePlugin()
```

beetsplug/lastgenre/__init__.py
```diff
@@ -498,21 +498,31 @@ def _try_resolve_original_fallback(
...
+
+        # We do not run through _try_resolve_stage yet because count could drop
+        # existing genres, but we still apply aliases before whitelist
+        # filtering.
+        normalized = [
+            norm if norm != g.lower() else g
+            for g in genres
+            if (norm := normalize_genre(self._log, self.alias_patterns, g))
+        ]
+        if valid := self._filter_valid(normalized, artist=artist):
```
snejus, line 518: Could we deduplicate the normalized genres before filtering? For example, `"hip-hop"` and `"hiphop"` can both become `"hip hop"`, so the fallback currently returns the same genre twice. `_resolve_genres` normally handles this with `unique_list`, but this new path bypasses it.

Could you also add a regression test with two aliases mapping to the same genre?

snejus (changes requested):

beetsplug/lastgenre/__init__.py
JOJ0: done. thanks!

snejus (approved):

JOJ0:
Hi @snejus I was to fast with saying "done", appologies! I added a few more test refactorings, all around where it is straight forward to use the `configure_plugin` helper you suggested. Please take another very quick lock and then I'll merge it. Thanks you!

test/plugins/test_lastgenre.py
```diff
@@ -9,14 +9,14 @@
...
-from beets.test.helper import IOMixin, PluginTestCase
+from beets.test.helper import IOMixin, PluginTestHelper
 from beetsplug import lastgenre
 from beetsplug.lastgenre.utils import is_ignored, normalize_genre
 
 _p = pytest.param
 
 
-class LastGenrePluginTest(IOMixin, PluginTestCase):
+class LastGenrePluginTest(IOMixin, PluginTestHelper):
```
snejus: Noticed that these tests didn't run any more? 😆 

```suggestion
class TestLastGenrePlugin(IOMixin, PluginTestHelper):
```

JOJ0: @snejus oh! 😅 but why? because of the class name? Because I replaced the mixin?

snejus: Previously this test was based on `unittest` - it doesn't care about class names.

Now it uses `pytest` - which only collect test classes named `class Test...` and functions named `def test_...`.

JOJ0: thanks! 🙏

Issue #4026, opened 2021-08-25
edit plugin launches editor with xdg-open but cannot wait for changes


philipsd6:
I've configured my machine to launch `emacsclient` via `xdg-open` as my default text editor. When launching an edit session, beets launches the editor, but the contents are empty and beets continues as if there were no changes, while the editor remains open.

### Problem

Running this command in verbose (`-vv`) mode:

```sh
$ beet -vv edit --all "All of Us"
```

Led to this problem:

```
user configuration: /home/philipsd6/.config/beets/config.yaml
data directory: /home/philipsd6/.config/beets
plugin paths: 
Sending event: pluginload
inline: adding item field multidisc
inline: adding album field albumformat
lastgenre: Loading canonicalization tree /home/philipsd6/.local/lib/python3.8/site-packages/beetsplug/lastgenre/genres-tree.yaml
library database: /home/philipsd6/.config/beets/library.db
library directory: /home/philipsd6/Music
Sending event: library_opened
edit: invoking editor command: ['xdg-open', '/tmp/tmp8sp2384u.yaml']
No changes; aborting.
Sending event: cli_exit
```

My editor opens `/tmp/tmp8sp2384u.yaml` but the file is empty.

If I adjust my `emacs.desktop` file's `Terminal` value from `true` to `false`, I then receive the tag content in my editor, but beets still continues and aborts with no changes, while my editor remains open with the tags. 

On the other hand, if I set `EDITOR=emacsclient` in my session, the edit plugin works as intended.

This is because `xdg-open` execs the process and returns, so falling back on `open_anything()` if `EDITOR` is not set does not work. I don't know if there's any way around this issue, but I thought I'd raise it anyway.

### Setup

* OS: Ubuntu 20.04.2 LTS
* Python version: Python 3.8.10
* beets version: 1.5.0
* Turning off plugins made problem go away (yes/no): no

My configuration (output of `beet config`) is:

n/a I think...

PR #6878, opened 2026-07-27
feat: add editor config option to editor_command()

Mo200227:
## Description

Fixes #6641

This PR adds an `editor` config option so users can permanently set their 
preferred editor in `config.yaml`, overriding `$VISUAL` and `$EDITOR` 
environment variables which may not be available in virtualenv environments 
(e.g. pipx with uv backend).

## Usage
```yaml
editor: nano
```

## Changes
- Modified `beets/util/__init__.py` — `editor_command()` now checks 
  `config["editor"]` first before falling back to `$VISUAL`/`$EDITOR`
- Added changelog entry to `docs/changelog.rst`

## Does this PR meet the acceptance criteria?
- [x] No breaking changes
- [x] Documentation updated (changelog)
- [ ] Tests added

PR #6894, opened 2026-07-30
Add a global `editor` configuration option

### LINKED ISSUE #6641: $EDITOR is not respected, likely due to uv virtualenv
(I'm deviating from the template, because this feels like a bug, but afaik it's not really beets' fault, and the solution is more like a feature request).

### Problem
My Installation does not respect `$EDITOR`:
<sub>(I modified beets to print out the environment variables)</sub>
```
$ echo $EDITOR
micro

$ beet edit test
$EDITOR: None
$VISUAL: None
No changes; aborting

$ EDITOR=nano beet edit test
$EDITOR: nano
$VISUAL: None
No changes; aborting.
```

Because of this, it defaults to xdg-open, which doesn't wait (#4026). I can fix it by passing `EDITOR=` explicitly, but i don't want to do that *every time*.

I believe this is due to pipx's uv backend, which seems to isolate virtualenvs from the shell's environment. I don't see an obvious/user-friendly way to get my `$EDITOR` into beet's virtualenv.

### Setup

* OS: Arch Linux (WSL)
* Python version: 3.14
* beets version: 2.11.0
* Turning off plugins made problem go away (yes/no): well... i guess so, yes 😅 

<sub>hopefully irrelevant?</sub>

### Proposed solution

Defaulting to `$EDITOR` et al is nice, but I would love the ability to override that behavior - permanently, not in every invocation. Absent a convenient pipx command, that means adding it to the config.

Either in the edit plugin:
```yaml
edit:
  editor: nano
```

Or, seeing how the editor is chosen in beet's utility functions, perhaps a global setting?
```yaml
editor: nano
```

tarann26:
## Description

Fixes #6641.

`editor_command()` currently resolves the text editor from `$VISUAL`, then `$EDITOR`, then the platform's generic file opener. When beets runs from an isolated virtualenv — the reporter hit this with pipx's uv backend — neither variable reaches the process, so `beet config -e` and the `edit` plugin silently fall back to `xdg-open`, which doesn't wait for the editor to close (#4026).

This adds a top-level `editor` configuration option so the editor can be set once, permanently, without depending on the environment:

```yaml
editor: nano
```

The resolution order becomes `editor` → `$VISUAL` → `$EDITOR` → platform default. The config option comes first so that setting it actually overrides the environment, which is the point of the request; an unset or empty value falls through to the existing behaviour, so nothing changes for anyone who doesn't set it.

The issue proposed two shapes — a key under the `edit:` plugin section, or a global one. I went with the global key, which is what @semohr endorsed in the thread ("Having the editor as a standalone config option a good idea imo"). It's also the only one that works for `beet config -e`, since the lookup lives in `beets/util` and is shared by the `config` command and the `edit` plugin rather than belonging to either.

The bare `editor:` line in `config_default.yaml` is required — without it `config["editor"].get()` raises `confuse.NotFoundError` instead of returning `None`. This mirrors `terminal_encoding`, the adjacent option, which is read the same way.

Also updates the `-e/--edit` help text and the "please set..." error in the `config` command to mention the new option, matching what b47635dc2 did when `$VISUAL` support was added.

## To Do

- [x] Documentation. (New `editor` section in `docs/reference/config.rst`; updated the editor-resolution description in `docs/reference/cli.rst` and `docs/plugins/edit.rst`.)
- [x] Changelog.
- [x] Tests. (Parametrised coverage of the full precedence chain in `test/test_util.py`, plus an end-to-end test through `beet config -e` in `test/ui/commands/test_config.py`.)

### DIFF
--- a/beets/config_default.yaml
+++ b/beets/config_default.yaml
@@ -118,6 +118,7 @@ timeout: 5.0
 
 verbose: 0
 terminal_encoding:
+editor:
 
 ui:
     terminal_width: 80
--- a/beets/ui/commands/config.py
+++ b/beets/ui/commands/config.py
@@ -54,8 +54,8 @@ def config_edit(cli_options):
 
     if not editor:
         raise UserError(
-            "Please set the VISUAL or EDITOR environment variable to edit"
-            " configuration."
+            "Please set the `editor` configuration option or the VISUAL or"
+            " EDITOR environment variable to edit configuration."
         )
     try:
         if not os.path.isfile(path):
@@ -78,7 +78,7 @@ config_cmd.parser.add_option(
     "-e",
     "--edit",
     action="store_true",
-    help="edit user configuration with $VISUAL (or $EDITOR)",
+    help="edit user configuration with the configured editor, $VISUAL, or $EDITOR",
 )
 config_cmd.parser.add_option(
     "-d",
--- a/beets/util/__init__.py
+++ b/beets/util/__init__.py
@@ -933,13 +933,16 @@ def open_anything() -> str:
 def editor_command() -> str:
     """Get a command for opening a text file.
 
-    First try environment variable `VISUAL` followed by `EDITOR`. As last resort
-    fall back to `open_anything()`, the platform-specific tool for opening files
-    in general.
+    First try the `editor` configuration option, then the environment variables
+    `VISUAL` and `EDITOR`. As last resort fall back to `open_anything()`, the
+    platform-specific tool for opening files in general.
 
     """
     return (
-        os.environ.get("VISUAL") or os.environ.get("EDITOR") or open_anything()
+        beets.config["editor"].get()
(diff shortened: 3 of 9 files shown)


semohr:
Thanks for your interest in beets!

As you may have noticed from the linked issue, we already have a fix in progress. In the future, please check for existing pull requests before opening a new one, as someone may already be working on the same issue.

For reference, see:

* https://github.com/beetbox/beets/pull/6878
* https://github.com/beetbox/beets/pull/6879
* https://github.com/Mo200227/beets/pull/1

I appreciate the effort you put into your PR. However, I've already reviewed the existing ones and would prefer to move forward with those, both to avoid duplicate work and to follow a first-come, first-served approach.

That said, I'd be happy to see more contributions from you in the future, there are always plenty of opportunities to help improve beets!

PR #6905, opened 2026-08-05
fix(tidal): honor search_limit when searching albums and tracks

### LINKED ISSUE #6770: tidal: `search_limit` is ignored

seedaylight:
## What changed

The Tidal plugin now honors the `search_limit` configuration option when
searching for albums and tracks. `search_albums_by_query()` and
`search_tracks_by_query()` read the configured limit and cap the candidate
lists returned by the Tidal search results endpoint, matching the behavior of
the other metadata source plugins.

## Why

`tidal.search_limit` is documented (default 5) but was previously ignored: the
plugin implements its own candidate search path that calls the Tidal API
directly and returns the full result set. Users setting `search_limit: 1` still
received 20+ candidates. See #6770.

## How

The Tidal searchResults API reference does not document a server-side `limit`
parameter, so the candidate relationship lists are truncated client-side to the
configured limit before track/album IDs are resolved. This guarantees the
option is honored regardless of API pagination defaults.

## Testing

- Added `TestSearchLimit` with focused tests for both album and track search
  using `search_limit: 1` in `test/plugins/test_tidal.py`.
- `pytest test/plugins/test_tidal.py`: 49 passed.
- `ruff check`, `ruff format --check`, and `mypy` on the changed module pass.
- Full `test/plugins/` run: 273 passed; the single failure
  (`test_fetchart.py::TestFetchartCli::test_colorization`) is a pre-existing
  environment issue (ANSI color output disabled in the local shell) and also
  fails on `master` without this change.
- Not verified against the live Tidal API (requires Tidal account
  authentication); tests use mocked API responses consistent with the existing
  suite.

Fixes #6770

### DIFF
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -174,6 +174,7 @@ class TidalPlugin(MetadataSourcePlugin):
 
     def search_tracks_by_query(self, query: str) -> Iterable[TrackInfo]:
         """Search for tracks given a string query."""
+        limit = self.config["search_limit"].get(int)
         search_doc = self.api.search_results(query, include=["tracks.artists"])
         track_by_id: dict[str, TidalTrack] = {
             item["id"]: item
@@ -185,7 +186,9 @@ class TidalPlugin(MetadataSourcePlugin):
             for item in search_doc.get("included", [])
             if item["type"] == "artists"
         }
-        for track_rel in search_doc["data"]["relationships"]["tracks"]["data"]:
+        for track_rel in search_doc["data"]["relationships"]["tracks"]["data"][
+            :limit
+        ]:
             if track := track_by_id.get(track_rel["id"]):
                 yield self._get_track_info(track, artist_by_id=artist_by_id)
             else:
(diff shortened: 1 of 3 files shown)


semohr:
Please check the existing PRs before opening new ones.

see https://github.com/beetbox/beets/pull/6781

PR #6907, opened 2026-08-05
tidal: Normalize copyright text into a concise label name

### LINKED ISSUE #6796: tidal: normalize copyright text used as label metadata
### Problem

The Tidal plugin assigns `copyright.text` directly to the `label` field. Tidal returns rights statements rather than normalized label names, so candidate output and imported metadata contain copyright markers, years, corporate relationships, and territorial boilerplate. Other autotaggers set normalized `label`.

Examples:

- `© 2011 Motown Records, a Division of UMG Recordings, Inc.`
- `© 2019 Atlantic Recording Corporation for the United States and WEA International Inc. for the world outside of the United States`
- `(P) 1992 Zomba Recording LLC`

This affects both albums and tracks because both paths use `_parse_label()`.

### Expected behavior

Store a concise label name instead:

- `Motown Records`
- Either `Atlantic Recording Corporation` or `WEA International Inc.`
- `Zomba Recording`

When the rights statement cannot be parsed safely, retain a conservatively cleaned value rather than guessing or dropping the field.

### Suggested fix

Add a reusable label normalization helper that removes leading copyright or phonographic copyright markers and years, then trims recognized corporate and territorial boilerplate at explicit phrase boundaries. Keep the transformation single-pass and conservative.

Add parametrized tests covering `©`, `℗`, `(C)`, and `(P)` prefixes, the examples above, plain label names, missing copyright data, and ambiguous rights statements.

NoDancing:
## Description

Fixes #6796

`_parse_label()` used to store Tidal's raw copyright info as the label name.

Adds `_normalize_label()`, a single-pass helper that:

- removes leading `©` / `℗` / `(C)` / `(P)` marker and year
- chops corporate-relationship clauses (`, a Division of X`, `, a BMG
  Company`)
- keeps the second half of `under exclusive license to X` statements
- splits territorial rights statements (`X for the United States and Y
  for the world outside...`), keeping the first label
- drops bare trailing legal-entity suffixes (`Inc.`, `LLC`, `Ltd.`, `Co.`)

Tests show all functionality in action, but for the curious, here's the results of running the helper function on a random selection of albums: https://gist.github.com/NoDancing/90fae264e46023de176a11cb5c065a58

## To Do

- [x] Documentation. 
- [x] Changelog.
- [x] Tests.


beetsplug/tidal/__init__.py
```diff
@@ -487,7 +530,7 @@ def _duration_to_seconds(duration: str) -> int | None:
     @staticmethod
     def _parse_label(attributes: MediaAttributes) -> str | None:
         if copyright_ := attributes.get("copyright"):
-            return copyright_["text"]
+            return TidalPlugin._normalize_label(copyright_["text"])
```
semohr, line 510: How about we skip the regex patterns entirely? I try to omit regex whenever possible for readability and performance reasons.

```python
@staticmethod
def _normalize_label(text: str) -> str:
    text = text.strip()

    # Leading copyright markers
    for marker in ("©", "℗", "(C)", "(P)", "(c)", "(p)"):
        if text.startswith(marker):
            text = text[len(marker):].strip()
            break

    # Leading year
    if len(text) >= 4 and text[:4].isdigit():
        text = text[4:].lstrip()

    # ", a "
    lower = text.lower()
    if ", a " in lower:
        idx = lower.index(", a ")
        text = text[:idx]

    # "under exclusive license to"
    phrase = "under exclusive license to"
    phrase_alt = "under exclusive licence to"

    lower = text.lower()
    for p in (phrase, phrase_alt):
        if p in lower:
            idx = lower.index(p)
            text = text[idx + len(p):]
            break

    # Trailing company suffixes
    for suffix in (" inc.", " inc", " llc", " ltd.", " ltd", " co.", " co"):
        if text.lower().endswith(suffix):
            text = text[:-len(suffix)].rstrip(", ")
            break

    # Territorial boilerplate
    phrase = " for the united states and "
    lower = text.lower()
    if phrase in lower:
        text = text[:lower.index(phrase)]

    return text.strip()
```

semohr (changes requested):

NoDancing:
Alright! 

Implemented those changes, nearly exactly as your wrote. I just added the docstring back in, changed the ordering of the corporate relationship and the license-to parts, (since one could cancel out the other), and I made sure there's a space between the digits and the next word for the year.

beetsplug/tidal/__init__.py
snejus: @semohr in my opinion regex patterns are simpler and more readable than the above.

snejus (changes requested):
You can simplify this to:

```diff
diff --git a/beetsplug/_utils/func.py b/beetsplug/_utils/func.py
new file mode 100644
index 000000000..4907702d3
--- /dev/null
+++ b/beetsplug/_utils/func.py
@@ -0,0 +1,13 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Callable, Iterable
+
+
+def apply_transforms(text: str, methods: Iterable[Callable[[str], str]]) -> str:
+    for method in methods:
+        text = method(text)
+
+    return text
diff --git a/beetsplug/lyrics.py b/beetsplug/lyrics.py
index c9e5c529d..4f10e4f70 100644
--- a/beetsplug/lyrics.py
+++ b/beetsplug/lyrics.py
@@ -28,6 +28,7 @@
 from beets.util.config import sanitize_choices
 from beets.util.lyrics import INSTRUMENTAL_LYRICS, Lyrics
 
+from ._utils.func import apply_transforms
 from ._utils.requests import (
     HTTPNotFoundError,
     RequestHandler,
@@ -35,7 +36,7 @@
 )
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Iterable, Iterator
+    from collections.abc import Iterable, Iterator
 
     import confuse
 
@@ -51,8 +52,6 @@
         TranslatorAPI,
     )
 
-    HtmlTransformer = Callable[[str], str]
-
 
 class CaptchaError(requests.exceptions.HTTPError):
     def __init__(self, *args, **kwargs) -> None:
@@ -441,13 +440,6 @@ def fetch
         return Lyrics(lyrics, self.__class__.name, url)
 
 
-def apply_transforms(html: str, methods: Iterable[HtmlTransformer]) -> str:
-    for method in methods:
-        html = method(html)
-
-    return html
-
-
 class Html:
     collapse_space = partial(re.compile(r"(^| ) +", re.M).sub, r"\1")
     expand_br = partial(re.compile(r"\s*<br[^>]*>\s*", re.I).sub, "\n")
diff --git a/beetsplug/tidal/__init__.py b/beetsplug/tidal/__init__.py
index 1a7091ef5..a2532ebff 100644
--- a/beetsplug/tidal/__init__.py
+++ b/beetsplug/tidal/__init__.py
@@ -4,7 +4,7 @@
 import os
 import re
 import time
-from functools import cached_property
+from functools import cached_property, partial
 from typing import TYPE_CHECKING, ClassVar, Literal, overload
 
 import confuse
@@ -15,6 +15,7 @@
 from beets.exceptions import UserError
 from beets.logging import getLogger
 from beets.metadata_plugins import MetadataSourcePlugin
+from beetsplug._utils.func import apply_transforms
 
 from .api import TidalAPI
 
@@ -40,6 +41,19 @@
 
 log = getLogger("beets.tidal")
 
+_remove_leading = partial(
+    re.compile(r"^(?:(?:©|℗|\([cp]\)) *)?(?:\d{4} )?", re.IGNORECASE).sub, ""
+)
+_remove_license = partial(
+    re.compile(r".* under exclusive licence to ", re.IGNORECASE).sub, ""
+)
+_remove_trailing_clause = partial(
+    re.compile(r"(?:, a| for the united states and) .*", re.IGNORECASE).sub, ""
+)
+_remove_legal_suffix = partial(
+    re.compile(r",? (?:inc|llc|ltd|co)\.?$", re.IGNORECASE).sub, ""
+)
+
 
 class TidalPlugin(MetadataSourcePlugin):
     item_types: ClassVar[dict[str, types.Type]] = {
@@ -394,50 +408,15 @@ def _get_album_info
 
     @staticmethod
     def _normalize_label(text: str) -> str:
-        """Normalize label from Tidal copyright text by stripping markers/years
-        and known corporate/licensing/territorial boilerplate."""
-        text = text.strip()
-
-        # Leading copyright markers
-        for marker in ("©", "℗", "(C)", "(P)", "(c)", "(p)"):
-            if text.startswith(marker):
-                text = text[len(marker) :].strip()
-                break
-
-        # Leading year
-        first, _, rest = text.partition(" ")
-        if len(first) == 4 and first.isdigit():
-            text = rest.lstrip()
-
-        # "under exclusive license to"
-        phrase = "under exclusive license to"
-        phrase_alt = "under exclusive licence to"
-
-        lower = text.lower()
-        for p in (phrase, phrase_alt):
-            if p in lower:
-                idx = lower.index(p)
-                text = text[idx + len(p) :]
-                break
-
-        # ", a "
-        lower = text.lower()
-        if ", a " in lower:
-            text = text[: lower.index(", a ")]
-
-        # Territorial boilerplate
-        phrase = " for the united states and "
-        lower = text.lower()
-        if phrase in lower:
-            text = text[: lower.index(phrase)]
-
-        # Trailing company suffixes
-        for suffix in (" inc.", " inc", " llc", " ltd.", " ltd", " co.", " co"):
-            if text.lower().endswith(suffix):
-                text = text[: -len(suffix)].rstrip(", ")
-                break
-
-        return text.strip()
+        return apply_transforms(
+            text,
+            [
+                _remove_leading,
+                _remove_license,
+                _remove_trailing_clause,
+                _remove_legal_suffix,
+            ],
+        )
 
     @staticmethod
     def _parse_artwork_url(
```

**Note**: you will want to move `apply_transforms` from `beetsplug/lyrics.py` to a new module `beetsplug/_utils/func.py`.

semohr (changes requested):
I would keep the non-regex version. The logic here is straightforward, and I don't think it's worth compiling four regex patterns just because we import the Tidal plugin.

For example, to understand this:

```python
partial(
    re.compile(r"^(?:(?:©|℗|\([cp]\)) *)?(?:\d{4} )?", re.IGNORECASE).sub, ""
)
```

a reader needs to:

- understand the regex itself,
- know what re.compile() is doing,
- and know why functools.partial() is being used.

That's a fair amount of indirection for logic that is otherwise very simple.

The explicit version:

```python
for marker in ("©", "℗", "(C)", "(P)", "(c)", "(p)"):
    if text.startswith(marker):
        text = text[len(marker):].strip()
        break

```
is immediately obvious to most Python developers, even those who aren't comfortable with regexes. It also avoids the overhead of compiling regex patterns, so it's slightly more efficient as well.

Unless the matching logic becomes substantially more complex, I'd favor the simpler, more explicit implementation.


---

If we combine all patterns into a **single compiled regex**, the performance argument becomes less relevant. 

For me, the stronger argument remains that the explicit implementation communicates the intent more clearly. Unless the matching rules become substantially more complex, I would favor the simpler, more direct approach.

---

Regarding this topic, I recommend  https://dl.acm.org/doi/abs/10.1109/ASE.2019.00047

NoDancing:
I'll hold off on making any more changes until you come to a consensus.

Thanks for sharing that research paper! Good perspective, there's stuff in there I genuinely didn't know.

snejus:
> I would keep the non-regex version. The logic here is straightforward, and I don't think it's worth compiling four regex patterns just because we import the Tidal plugin.
> 
> For example, to understand this:
> 
> ```python
> partial(
>     re.compile(r"^(?:(?:©|℗|\([cp]\)) *)?(?:\d{4} )?", re.IGNORECASE).sub, ""
> )
> ```
> 
> a reader needs to:
> 
>     * understand the regex itself,
> 
>     * know what re.compile() is doing,
> 
>     * and know why functools.partial() is being used.
> 
> 
> That's a fair amount of indirection for logic that is otherwise very simple.
> 
> The explicit version:
> 
> ```python
> for marker in ("©", "℗", "(C)", "(P)", "(c)", "(p)"):
>     if text.startswith(marker):
>         text = text[len(marker):].strip()
>         break
> ```
> 
> is immediately obvious to most Python developers, even those who aren't comfortable with regexes. It also avoids the overhead of compiling regex patterns, so it's slightly more efficient as well.
> 
> Unless the matching logic becomes substantially more complex, I'd favor the simpler, more explicit implementation.
> 
> If we combine all patterns into a **single compiled regex**, the performance argument becomes less relevant.
> 
> For me, the stronger argument remains that the explicit implementation communicates the intent more clearly. Unless the matching rules become substantially more complex, I would favor the simpler, more direct approach.
> 
> Regarding this topic, I recommend https://dl.acm.org/doi/abs/10.1109/ASE.2019.00047

Thanks for expanding on this. I agree that the previous combined `_remove_leading` expression was unnecessarily dense, so I'd suggest to split it into independently named `_remove_copyright` and `_remove_year` transformations:

```py
_remove_copyright = partial(
    re.compile(r"^(©|℗|\([cp]\)) *", re.IGNORECASE).sub, ""
)
_remove_year = partial(re.compile(r"^\d{4} ", re.IGNORECASE).sub, "")
```

With that change, I find the complete regex version easier to follow than the string version. The normalization policy is visible as one short pipeline:

```python
[
    _remove_copyright,
    _remove_year,
    _remove_license,
    _remove_trailing_clause,
    _remove_legal_suffix,
]
```

Each regex now has one narrowly named responsibility. They also use only basic regex features that developers should be familiar with. By comparison, the string version spreads the same policy across repeated `lower()`, loops, `index()`, slicing, temporary phrase variables, and roughly 40 lines of control flow. The isolated marker loop is straightforward, but it is only one part of the full comparison. As a maintainer, I find 5 simple regexes much easier to maintain than a long function with custom string-method-based replacement logic.

`partial(re.compile(...).sub, "")` is also an existing pattern in the lyrics plugin, and `apply_transforms` is now shared by both callers, so this is not a new local convention.

I don't think performance should decide this either way. `_normalize_label()` runs once per candidate, and both implementations complete in under 100 μs. At that call frequency and absolute cost, any difference between them is immaterial, so I think we should choose based on readability. The regexes are also compiled once rather than once per candidate.

I read the paper you linked. It supports caution around complex regexes, but it also reports that developers choose regexes for medium-complexity matching and prefer shorter expressions with fewer features. I think these small, named, tested transformations fit that use case better than one large imperative function.

semohr:
To keep this brief, I see it differently. For this transformation, I find a few explicit string operations easier to read and maintain than **five separate** regular expressions. Each operation describes exactly what it removes or changes, while the equivalent regexes require readers to decode the patterns.

I’m fine with either of these approaches:

a) Use small string-method helpers, optionally through the existing `apply_transforms` pipeline.
b) Use a single regex to parse the complete label. This is more concise, but concentrates more complexity into one pattern.

As a maintainer and the code owner of the tidal plugin, I’m happy to maintain either solution. Beets already uses regex-based transformation rules in some plugin APIs, so I’m not opposed to regex in principle, I just don’t think it is the clearest choice here, since these transformations are simple and sequential rather than pattern-heavy. The relevant formats and edge cases should be covered by tests either way.

snejus:
Thanks. I understand what both alternatives would look like mechanically, but I don't yet understand how either would improve readability over the current design. Could you compare each option against the current implementation as a whole?

For option (a), we would keep the named transformation pipeline but replace each short regex with a custom helper containing some combination of loops, `lower()`, `index()`, slicing, suffix lists, and temporary variables. The individual string operations are familiar, but the complete implementation is substantially longer and spreads each matching rule across more control flow. What makes that easier to read or maintain than the current small, named regex transformations?

For example, I find 
```py
_remove_license = partial(
    re.compile(r".* under exclusive licence to ", re.IGNORECASE).sub, ""
)
```
more readable than
```py
        # "under exclusive license to"
        phrase = "under exclusive license to"
        phrase_alt = "under exclusive licence to"

        lower = text.lower()
        for p in (phrase, phrase_alt):
            if p in lower:
                idx = lower.index(p)
                text = text[idx + len(p) :]
                break
```

For option (b), the transformations do different things: some remove a prefix, one keeps the text after a phrase, and others keep the text before a phrase or remove a suffix. Combining those behaviors into one regex appears to require optional branches, capture groups, and conditional replacement logic. That seems considerably harder to understand than keeping one basic regex per named rule.

The current design keeps the policy visible in one place:

```python
[
    _remove_copyright,
    _remove_year,
    _remove_license,
    _remove_trailing_clause,
    _remove_legal_suffix,
]
```

Each transformation has one responsibility and uses only basic regex features. Setting performance aside, could you explain which specific part of this design remains less readable, and how either (a) or (b) improves that aspect? I'm not arguing that regex is always preferable. I'm trying to understand why the middle option of several small, named regex transformations is being excluded.

semohr:
The argument I am making (and the point discussed in the paper) is that, since these individual transformations are very simple, introducing regex does not necessarily improve readability. If we consider the complete transformation as one operation, a single, explicit implementation of the transformation logic may be easier to understand than a sequence of separate regex-based rules.

For option (a), I am not primarily arguing that the helper functions are more readable line by line. My point is more about the effort required to understand what is actually happening to the string. With the multi regex implementation, the reader has to interpret several independent regex patterns and mentally reconstruct their combined effect. A direct implementation makes the string manipulation steps explicit, even if it requires more lines of code.

For option (b), I think I did not explain the intention clearly enough. The goal is not to introduce a more complex regex-based solution, but rather to reduce the current boilerplate around the transformations. The idea is to describe the complete transformation in one place and keep the implementation concise.

```python
re.compile(
    r"^(?:(?:©|℗|\([cp]\))\s*)?"          # optional copyright / phonogram marker
    r"(?:\d{4}\s+)?"                      # optional year
    r"(?:.*? under exclusive licen[sc]e to )?"  # optional license prefix
    r"(.*?)"                              # text to keep
    r"(?:, (?:a|for the united states and) .*"  # optional trailing clause
    r"|,? (?:inc|llc|ltd|co)\.?)?$",      # optional legal suffix
    re.IGNORECASE,
)
```
The advantage of option (b) is not adding another abstraction layer, but removing unnecessary repetition while keeping the transformation logic in one place.

---

In summary, option (a) favors transparency and step-by-step explainability, whereas option (b) favors simplicity and a more compact implementation.

snejus:
> For option (b), I think I did not explain the intention clearly enough. The goal is not to introduce a more complex regex-based solution, but rather to reduce the current boilerplate around the transformations. The idea is to describe the complete transformation in one place and keep the implementation concise.
> 
> ```python
> re.compile(
>     r"^(?:(?:©|℗|\([cp]\))\s*)?"          # optional copyright / phonogram marker
>     r"(?:\d{4}\s+)?"                      # optional year
>     r"(?:.*? under exclusive licen[sc]e to )?"  # optional license prefix
>     r"(.*?)"                              # text to keep
>     r"(?:, (?:a|for the united states and) .*"  # optional trailing clause
>     r"|,? (?:inc|llc|ltd|co)\.?)?$",      # optional legal suffix
>     re.IGNORECASE,
> )
> ```

I think I'd be happy with this, provided that we use a single string with `re.VERBOSE` flag.

NoDancing:
Alright I implemented this. One big string, all tests pass.

Last question, I'm guessing that we don't want to bother with `apply_transforms` now that we're just applying one expression. Do you want me to return it to `beetsplug/lyrics.py`, and remove the new `beetsplug/_utils/func.py`?

```py
_normalize_label_re = re.compile(
    r"""
    ^(?:.*(?:©|℗|\([cp]\))\s*)?          # optional text through last marker
    (?:\d{4}\s+)?                            # optional year
    (?:.*?\ under\ exclusive\ licen[sc]e\ to\ )?  # optional license prefix
    (.*?)                                    # text to keep
    (?:,?\ (?:inc|llc|ltd|co)\.?)?           # optional legal suffix
    (?:,\ a\ .*)?                            # optional corporate clause
    (?:,?\ for\ the\ united\ states\ and\ .*)?  # optional territorial clause
    $""",
    re.IGNORECASE | re.VERBOSE,
)
```

semohr (approved):
Thank you for bearing with us!

beetsplug/tidal/__init__.py
```diff
@@ -40,6 +40,19 @@
...
+_normalize_label_re = re.compile(
+    r"""
+    ^(?:.*(?:©|℗|\([cp]\))\s*)?          # optional text through last marker
+    (?:\d{4}\s+)?                            # optional year
+    (?:.*?\ under\ exclusive\ licen[sc]e\ to\ )?  # optional license prefix
+    (.*?)                                    # text to keep
+    (?:,?\ (?:inc|llc|ltd|co)\.?)?           # optional legal suffix
+    (?:,\ a\ .*)?                            # optional corporate clause
+    (?:,?\ for\ the\ united\ states\ and\ .*)?  # optional territorial clause
+    $""",
```
snejus, line 52: Just one final question, would you mind aligning comments here?

```suggestion
    r"""
    ^(?:.*(?:©|℗|\([cp]\))\s*)?                # optional copyright marker
    (?:\d{4}\s+)?                              # optional year
    (?:.*?\ under\ exclusive\ licence\ to\ )?  # optional license prefix
    (.*?)                                      # text to keep
    (?:,?\ (?:inc|llc|ltd|co)\.?)?             # optional legal suffix
    (?:,\ a\ .*)?                              # optional corporate clause
    (?:,?\ for\ the\ united\ states\ and\ .*)? # optional territorial clause
    $""",
```

I adjusted some of the comments and simplified the license prefix marker as well.

snejus (changes requested):

NoDancing:
Fixed it. Thank you two for bearing with me as well!

snejus (approved):

PR #6912, opened 2026-08-05
fix(discogs): retry malformed search responses

treyturner:
## Summary

Retry Discogs searches once when the API returns an invalid JSON response. If the retry also fails, flow through existing error handling. Includes regression tests for successful recovery and retry exhaustion.

Reference: https://github.com/joalla/discogs_client/issues/163

Originating stack trace:

```
discogs: Error searching Discogs: Expecting value: line 1 column 1 (char 0)
  File "/usr/local/lib/python3.14/site-packages/beets/metadata_plugins.py", line 420, in _search_api
    response_data = self.get_search_response(params)
  File "/usr/local/lib/python3.14/site-packages/beetsplug/discogs/__init__.py", line 297, in get_search_response
    return [r.data for r in results.page(1)]
                            ~~~~~~~~~~~~^^^
  File "/usr/local/lib/python3.14/site-packages/discogs_client/models.py", line 347, in page
    data = self.client._get(self._url_for_page(index))
  File "/usr/local/lib/python3.14/site-packages/discogs_client/client.py", line 115, in _get
    return self._request('GET', url)
           ~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/site-packages/discogs_client/client.py", line 107, in _request
    body = json.loads(content)
  File "/usr/local/lib/python3.14/json/__init__.py", line 352, in loads
    return _default_decoder.decode(s)
           ~~~~~~~~~~~~~~~~~~~~~~~^^^
  File "/usr/local/lib/python3.14/json/decoder.py", line 345, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.14/json/decoder.py", line 363, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

## Testing

- Full test suite on Python 3.14
- Ruff, mypy, and Sphinx lint

## To Do

### DIFF
--- a/beetsplug/discogs/__init__.py
+++ b/beetsplug/discogs/__init__.py
@@ -292,9 +292,19 @@ class DiscogsPlugin(SearchApiMetadataSourcePlugin[IDResponse]):
 
     def get_search_response(self, params: SearchParams) -> Sequence[IDResponse]:
         """Search Discogs releases and return raw result mappings with IDs."""
-        results = self.discogs_client.search(params.query, **params.filters)
-        results.per_page = params.limit
-        return [r.data for r in results.page(1)]
+
+        def search() -> list[IDResponse]:
+            results = self.discogs_client.search(params.query, **params.filters)
+            results.per_page = params.limit
+            return [r.data for r in results.page(1)]
+
+        try:
+            return search()
+        except json.JSONDecodeError:
+            self._log.debug(
+                "Discogs returned an invalid JSON search response; retrying"
+            )
+            return search()
 
     @cache
     def get_master_year(self, master_id: str) -> int | None:
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -54,6 +54,8 @@ Bug fixes
   copyright/rights-statement text verbatim. It's now normalized to a concise
   label name, stripping copyright markers, years, and corporate, licensing, and
   territorial boilerplate. Affects both album and track metadata. :bug:`6796`
+- :doc:`plugins/discogs`: Retry a search once when Discogs returns an invalid
+  JSON response instead of immediately discarding all Discogs candidates.
 
 ..
     For plugin developers
--- a/test/plugins/test_discogs.py
+++ b/test/plugins/test_discogs.py
@@ -2,13 +2,16 @@
 
 from __future__ import annotations
 
+import json
 from typing import TYPE_CHECKING, Any
+from unittest.mock import MagicMock
 
 import pytest
 from discogs_client import Client, Release
 
 from beets import config
 from beets.library import Item
+from beets.metadata_plugins import SearchParams
 from beets.test.helper import TestHelper
 from beetsplug.discogs import ArtistState, DiscogsPlugin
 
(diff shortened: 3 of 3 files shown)


semohr:
This seems like something to me that could be part of the `discogs_client` instead. @JOJ0 Assigning you here since this seems to be right in your expertise :upside_down_face:

henry-oberholtzer:
The code for this PR looks good to me! I think my question is whether immediately retrying a query immediately after a failure seems to actually yield results after a json.decode() error.

If it's an outage on the discogs server side, my instinct is that it won't fix it, but I could be totally wrong if anyone has anecdotal experience with this issue.

It's also hard to test if a re-search actuallr recovers errors from discogs since it seems to be a white whale issue - none of this is to say we shouldn't try to give a query another shot in a failure.

EDIT:

Well, anecdotally - it seems to resolve it according to this post: https://github.com/joalla/discogs_client/issues/163#issuecomment-3403116431

henry-oberholtzer (approved):
Looks simple enough. I'll let another maintainer just re-confirm since I've been away from the project for a minute. :)

treyturner:
> It's also hard to test if a re-search actually recovers errors

My anecdotal experience is also that a retry returns success, but that's usually after at least a few seconds following an abort and restart of the current import. We may find success to be less likely when immediately retried. I added a log line indicating the retry so that we can later evaluate if a small sleep between attempts may increase the chance of success.

treyturner:
> This seems like something to me that could be part of the `discogs_client` instead.

I can definitely appreciate this take. I considered that approach as well, but given the discussion and apparent ambiguity in https://github.com/joalla/discogs_client/issues/163, I thought this one small effort upstream could protect the user experience while `discogs_client` continues working through opportunities to improve consistency. If it required really anything else here in `beets` I'd not have been comfortable proposing it.

Still, I completely understand if this is desired downstream instead. Thanks for your consideration!

JOJ0 (approved):
Hi, do that, since python3-discogs-client will not not implement anything anytime soon.

JOJ0:
> This seems like something to me that could be part of the `discogs_client` instead. @JOJ0 Assigning you here since this seems to be right in your expertise 🙃

Not any time soon doesn't mean _never_ :-) It is out of scope to discuss this here in the beets project, I want to do it anyway since now we sit together around the camp fire and talk about it already ;-) No, seriously I was thinking about catching this error originally as the linked discussion in the upstream project suggests. I am still not sure what to do with a catch then. If we do nothing (return None?) we might run into errors elsewhere.

So I figure we must do something about it: Should we retry it once like in this PR here? That might work or might not work and doesnt really sound like a solution that will always make sense.

Should we implement some "back off for a random time and try later approach" and then do that a configurable amount of times? (similar to what rate limiting handlers do)

Are we doing the right thing already? We notify the caller that there is a serious error and they need to handle it themselves: We throw an exception. Well, the exception could tell something more useful like: _There is a serious error happening on discogs.com, probably somewhere in their backend_ - but that is heavy guesswork also and not much better, and probably just wrong.

What's your suggestions discogs_client should do when it hits such an error?

What do other libraries do when their backend has an error? Don't they just throw an exception since it is a non-predictable non-recoverable error?

And by the way, whenever I experienced these errors ,of course this website doesnt show any discogs outage 🤣 https://github.com/joalla/discogs_client/issues/163#issuecomment-2676848514

semohr:
> What do other libraries do when their backend has an error? Don't they just throw an exception since it is a non-predictable non-recoverable error?

If it is truely unpredictable there is nothing one can do as api wrapper developer really.

Generally what I would like tho, is an obvious error. We can than catch it here and handle approprietly, just relying on a json decode exception seems a bit strange and messy imo. Something like a generic `DiscogsbackendError` could work. Here in beets we can than handle that error and retry once or with backoff if appropriate.

JOJ0:
> > What do other libraries do when their backend has an error? Don't they just throw an exception since it is a non-predictable non-recoverable error?
> 
> If it is truely unpredictable there is nothing one can do as api wrapper developer really.
> 
> Generally what I would like tho, is an obvious error. We can than catch it here and handle approprietly, just relying on a json decode exception seems a bit strange and messy imo. Something like a generic `DiscogsbackendError` could work. Here in beets we can than handle that error and retry once or with backoff if appropriate.

HI @semohr, thanks so much for you opinion. Much appreciated. I agree that a "definitive error" would be a good middleground here. Since we don't _really_ know what the origin of the error is I'd word this even more general and try not to guess what or _where_ this is/was happening. We don't really know what "in the backend" means, so.... What we know is that it is not a regular result so maybe something like `DiscogsMalformedResponseError" or even `DiscogsUnexpectedError`. Anyway thanks again I'll redirect this now to the upstream project.

PR #6913, opened 2026-08-06
duplicates: improve the checksum option docs

### LINKED ISSUE #3979: Duplicates: Improve docs on using external commands
Now that the path substitution issue from #2873 is fixed, the docs should probably be amended to reflect 

---

So maybe the advice should be to create a script
```sh
#! /usr/bin/env sh

md5sum < "$1"
```
or
```sh
#! /usr/bin/env sh

md5sum "$1" | awk '{printf $1}'
```
and use it with
```sh
beet dup -C 'myscript {file}' ...
```

_Originally posted by @wisp3rwind in https://github.com/beetbox/beets/issues/2873#issuecomment-694985413_

cycsmail:
## Description

Fixes #3979.

The docs suggested `beet dup -C 'md5sum {file}'` as a checksum command, but it can never match anything: the plugin uses the command's entire output as the checksum, and md5sum prints the file's path after the hash, so two copies of the same file always look different. The docs also didn't explain how the command is run (split shlex-style with `{file}` substituted per argument, no shell, so no pipes or redirection).

Reworded the `checksum` option description to cover that, dropped the broken md5sum example and added a small wrapper-script recipe for commands like md5sum. I checked the recipe end to end on a throwaway library with duplicate files (paths with spaces included); the duplicates plugin tests pass and sphinx-lint/docstrfmt are clean on the changed files.

## To Do

- [x] Documentation.
- [x] Changelog.
- [x] ~Tests~ (docs-only change)


docs/plugins/duplicates.rst
```diff
@@ -47,8 +47,14 @@ file. The available options mirror the command-line options:
   overrides the ``keys`` option the first time it is run; however, because it
   caches the resulting checksum as ``flexattrs`` in the database, you can use
   ``--key=name_of_the_checksumming_program --key=any_other_keys`` (or set the
-  ``keys`` configuration option) the second time around. Default: ``ffmpeg -i
-  {file} -f crc -``.
+  ``keys`` configuration option) the second time around. The command is not run
+  through a shell: it is split into arguments (respecting shell-like quoting)
+  and ``{file}`` is then replaced with the item's path within each argument, so
+  paths containing spaces need no extra quoting, but shell features such as
+  pipes or redirection cannot be used. The command's entire output is used as
```
snejus, line 54: Let's highlight this bit:

```suggestion
  paths containing spaces need no extra quoting. **Note: shell features such as
  pipes or redirection cannot be used.** The command's entire output is used as
```

docs/plugins/duplicates.rst
```diff
@@ -126,13 +132,28 @@ Get tracks with the same title, artist, and album:
...
 
     beet dup -C 'ffmpeg -i {file} -f crc -'
-    beet dup -C 'md5sum {file}'
+
+Note that ``beet dup -C 'md5sum {file}'`` would not work: the command's entire
+output is used as the checksum, and ``md5sum`` prints the file's path after the
+hash, so copies of the same file at different paths would never match. Instead,
+wrap the command in a script that prints only the hash:
+
+::
```
snejus, line 147: Add syntax highlighting to each code block

snejus (changes requested):

cycsmail:
Thanks, applied the suggestion and added bash highlighting to the code blocks, also merged master in since the changelog had moved on. Should be good for another look.

snejus (approved):
Thanks!

PR #6918, opened 2026-08-09
Fix Discogs search retry on malformed JSON

KKhandelwal1733:
Closes #6912

## Problem
The Discogs plugin crashes when the Discogs API returns a malformed JSON response during a search. The `json.loads` call raises `JSONDecodeError`, which propagates up and aborts the metadata lookup.

## Solution
Wrap the search request in a `try/except` block that catches `json.JSONDecodeError`. On the first failure, retry the request once. If the retry also fails, re‑raise the original exception so existing error handling takes over.

## Changes
- `beetsplug/discogs/__init__.py` — added retry logic around the API call, imported `json` for the exception type, and documented the behavior.

## Testing
- Added regression tests that simulate a malformed JSON response on the first call and a valid response on the retry, asserting that the search succeeds.
- Added tests where both attempts return malformed JSON, asserting that the error is propagated.
- Ran the full test suite on Python 3.14 and verified that linting (ruff, mypy) and documentation builds pass.

### DIFF
--- a/beetsplug/discogs/__init__.py
+++ b/beetsplug/discogs/__init__.py
@@ -294,7 +294,10 @@ class DiscogsPlugin(SearchApiMetadataSourcePlugin[IDResponse]):
         """Search Discogs releases and return raw result mappings with IDs."""
         results = self.discogs_client.search(params.query, **params.filters)
         results.per_page = params.limit
-        return [r.data for r in results.page(1)]
+        try:
+            return [r.data for r in results.page(1)]
+        except json.JSONDecodeError:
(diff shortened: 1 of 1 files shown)


semohr:
Have you had a chance to discuss this with @treyturner? We generally follow a first-come, first-served approach, and since there’s already a PR addressing this issue, I think it makes sense to continue with the existing one and avoid duplicating the work.

I’ll close this PR for now, but you’re more than welcome to review the existing PR, share your thoughts or suggestions there, and help give it some traction.

